PR #2606, opened 2022-05-29
Adjust colors in tokyonight themes

paulgraydon:
Change `ui.virtual.ruler` color to `bg` instead of `fg`, correct syntax highlighting to more closely resemble the neovim theme.


the-mikedavis (approved):
Looks good, thanks!

PR #2607, opened 2022-05-29
Improve goto_file

### LINKED ISSUE #1701: goto_file_impl doesn't seem to use path relative to current working directory
When using the "files in selection" goto command, the selected file path is treated as relative to what helix-term determines as the "root" rather than the current working directory, even if helix is launched within the subdirectory. Judging from #1374, it looks like this method for determining the root directory is the intended behavior, so I'm not sure whether this would be classified as a bug or a feature request.

typetetris:
- Try to resolve relative paths relative to the documents
  parent path (its directory) in case there is a path for
  the document.

In case there is only one selection of a single character:

- Use `textobject_word(.., Inside, ..)` to determine the path.

  This way you can stand on the first or last character of
  the path and it will still use this path (and not
  the previous one, in case you stood on the first character
  of the path).

- If the path points to a directory, open the file picker for
  this directory. General improvements also apply, so if the
  selected path is relative, it will also be resolved relative
  to the documents directory if possible.

This might be considered a reason to close #1701

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1013,7 +1013,7 @@ fn goto_file_vsplit(cx: &mut Context) {
     goto_file_impl(cx, Action::VerticalSplit);
 }
 
-fn resolve_selected_path(
+fn trim_and_resolve_path(
     selection_text: String,
     doc_path_parent: Option<&Path>,
 ) -> Option<PathBuf> {
@@ -1052,7 +1052,7 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         let selection_text = text
             .slice(current_word.from()..current_word.to())
             .to_string();
-        if let Some(path) = resolve_selected_path(selection_text, doc_path_parent) {
+        if let Some(path) = trim_and_resolve_path(selection_text, doc_path_parent) {
             match path.metadata() {
                 Ok(metadata) => {
                     if metadata.is_dir() {
@@ -1073,7 +1073,7 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         let paths: Vec<_> = selections
             .iter()
             .map(|r| text.slice(r.from()..r.to()).to_string())
-            .filter_map(|selection_text| resolve_selected_path(selection_text, doc_path_parent))
+            .filter_map(|selection_text| trim_and_resolve_path(selection_text, doc_path_parent))
             .collect();
         for path in paths {
             if let Err(e) = cx.editor.open(path, action) {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1023,14 +1023,10 @@ fn trim_and_resolve_path(
     }
 
     let selected_path = PathBuf::from(trimmed);
-    match (selected_path.is_relative(), doc_path_parent) {
-        (true, Some(parent_path)) => {
-            let mut result = PathBuf::new();
-            result.push(parent_path);
-            result.push(selected_path);
-            Some(result)
-        }
-        _ => Some(selected_path),
+    if let (true, Some(parent_path)) = (selected_path.is_relative(), doc_path_parent) {
+        Some(parent_path.join(selected_path))
+    } else {
+        Some(selected_path)
     }
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1013,31 +1013,70 @@ fn goto_file_vsplit(cx: &mut Context) {
     goto_file_impl(cx, Action::VerticalSplit);
 }
 
+fn resolve_selected_path(
+    selection_text: String,
+    doc_path_parent: Option<&Path>,
+) -> Option<PathBuf> {
+    let trimmed = selection_text.trim();
+    if trimmed.is_empty() {
+        return None;
+    }
+
+    let selected_path = PathBuf::from(trimmed);
+    match (selected_path.is_relative(), doc_path_parent) {
+        (true, Some(parent_path)) => {
+            let mut result = PathBuf::new();
+            result.push(parent_path);
+            result.push(selected_path);
+            Some(result)
+        }
+        _ => Some(selected_path),
+    }
+}
+
 fn goto_file_impl(cx: &mut Context, action: Action) {
     let (view, doc) = current_ref!(cx.editor);
     let text = doc.text();
     let selections = doc.selection(view.id);
-    let mut paths: Vec<_> = selections
-        .iter()
-        .map(|r| text.slice(r.from()..r.to()).to_string())
-        .collect();
     let primary = selections.primary();
+    let doc_path_parent = doc.path().and_then(|path| path.parent());
+
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
-        let current_word = movement::move_next_long_word_start(
+        let current_word = textobject::textobject_word(
             text.slice(..),
-            movement::move_prev_long_word_start(text.slice(..), primary, 1),
+            primary,
+            textobject::TextObject::Inside,
             1,
+            true,
         );
-        paths.clear();
-        paths.push(
-            text.slice(current_word.from()..current_word.to())
-                .to_string(),
-        );
-    }
-    for sel in paths {
-        let p = sel.trim();
-        if !p.is_empty() {
-            if let Err(e) = cx.editor.open(&PathBuf::from(p), action) {
+        let selection_text = text
+            .slice(current_word.from()..current_word.to())
+            .to_string();
+        if let Some(path) = resolve_selected_path(selection_text, doc_path_parent) {
+            match path.metadata() {
+                Ok(metadata) => {
+                    if metadata.is_dir() {
+                        let picker = ui::file_picker(path, &cx.editor.config());
+                        cx.push_layer(Box::new(overlayed(picker)));
+                    } else if let Err(e) = cx.editor.open(path, action) {
+                        cx.editor.set_error(format!("Open file failed: {:?}", e));
+                    }
+                }
+                Err(err) => cx.editor.set_error(format!(
+                    "Couldn't get metadata for path {}: {:?}",
+                    path.to_string_lossy(),
+                    err
+                )),
+            }
+        }
+    } else {
+        let paths: Vec<_> = selections
+            .iter()
+            .map(|r| text.slice(r.from()..r.to()).to_string())
+            .filter_map(|selection_text| resolve_selected_path(selection_text, doc_path_parent))
+            .collect();
+        for path in paths {
+            if let Err(e) = cx.editor.open(path, action) {
                 cx.editor.set_error(format!("Open file failed: {:?}", e));
             }
         }


helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
...
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
-        let current_word = movement::move_next_long_word_start(
+        let current_word = textobject::textobject_word(
             text.slice(..),
-            movement::move_prev_long_word_start(text.slice(..), primary, 1),
+            primary,
+            textobject::TextObject::Inside,
             1,
+            true,
         );
```
pickfire, line 1038: This method won't work for file names that contains CJK characters but at least it is an improvement over the current one.

test.txt
```diff
@@ -0,0 +1,4 @@
+./flake.nix
+/home/typetetris/repos/conf/
+./
+./helix-term/test.txt
```
pickfire, line 4: I don't think this is what you wanted to commit.

helix-term/test.txt
```diff
@@ -0,0 +1,2 @@
+./Cargo.toml
+./
```
pickfire, line 2: I think you accidentally committed the testing file.

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
...
+        return None;
+    }
+
+    let selected_path = PathBuf::from(trimmed);
+    match (selected_path.is_relative(), doc_path_parent) {
+        (true, Some(parent_path)) => {
+            let mut result = PathBuf::new();
+            result.push(parent_path);
+            result.push(selected_path);
+            Some(result)
```
pickfire, line 1018:
```suggestion
            Some(parent_path + selected_path)
```
Would this work?

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
...
 }
 
+fn resolve_selected_path(
+    selection_text: String,
+    doc_path_parent: Option<&Path>,
+) -> Option<PathBuf> {
+    let trimmed = selection_text.trim();
+    if trimmed.is_empty() {
+        return None;
+    }
```
pickfire, line 1010: Why not just allow it to continue so it can open up current directly?

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
     goto_file_impl(cx, Action::VerticalSplit);
 }
 
+fn resolve_selected_path(
+    selection_text: String,
+    doc_path_parent: Option<&Path>,
+) -> Option<PathBuf> {
+    let trimmed = selection_text.trim();
```
pickfire, line 1007: Should this trim be done in the outer function instead?

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
...
+    } else {
+        let paths: Vec<_> = selections
+            .iter()
+            .map(|r| text.slice(r.from()..r.to()).to_string())
+            .filter_map(|selection_text| resolve_selected_path(selection_text, doc_path_parent))
+            .collect();
+        for path in paths {
+            if let Err(e) = cx.editor.open(path, action) {
                 cx.editor.set_error(format!("Open file failed: {:?}", e));
             }
```
pickfire, line 1068: Maybe allow opening the file by creating it even though it doesn't exist? IIRC in vim I have the need to open file even though they didn't exist, not sure how good is it to have it here.

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,70 @@ fn goto_file_vsplit(cx: &mut Context) {
...
+            .to_string();
+        if let Some(path) = resolve_selected_path(selection_text, doc_path_parent) {
+            match path.metadata() {
+                Ok(metadata) => {
+                    if metadata.is_dir() {
+                        let picker = ui::file_picker(path, &cx.editor.config());
+                        cx.push_layer(Box::new(overlayed(picker)));
+                    } else if let Err(e) = cx.editor.open(path, action) {
+                        cx.editor.set_error(format!("Open file failed: {:?}", e));
+                    }
```
pickfire, line 1050: Maybe allow opening the file by creating it even though it doesn't exist? IIRC in vim I have the need to open file even though they didn't exist, not sure how good is it to have it here.

helix-term/src/commands.rs
pickfire, line 1038: Looks reasonable to me, but I am surprised it even works for CJK characters.

helix-term/test.txt
typetetris: Yes, will remove it!

helix-term/src/commands.rs
typetetris: I can test it, but I would expect `PathBuf::push` to handle directory seperators correctly in all cases and I am afraid, `parent_path + selected_path` would not.

helix-term/src/commands.rs
typetetris: This is just a reproduction of the current logic. I could play around with this, as I am not sure, what will happen, if I use the empty string as file or directory name in the respective functions.

typetetris: After thinking about this a bit it is my opinion, it is good as it is.

To open the file picker, we have `<Space>+f`.

And `gf` on a selection containing whitespace opening the file picker for the current document directory is counterintuitive in my opinion. Thats what `gf` should do on `./` or some such.

helix-term/src/commands.rs
typetetris: This would duplicate it in two places.

If I tried to extract it from that function and have like a pipeline

1. trim
2. check if empty
3. resolve relative to document, if possible

that would also be duplicated in two places. Putting the three steps in a function again, would
lead to the same result I have now.

Maybe I could rename the function `trim_and_resolve` or so?

helix-term/src/commands.rs
typetetris: At least neovim also just reports "Can't find file in path" and doesn't try to create a file.

Thought experiment: If you pressed `gf` on a random string (maybe accidentally) and it would just create a file with this
name and place you in a buffer for it, I suspect, that could get kinda annoying.

helix-term/src/commands.rs
typetetris: At least for nix it would be cool to use language text objects to get the current path (that is a type in nix)/string content. But I wasn't able to digest the tree-sitter API yet.

helix-term/src/commands.rs
pickfire: How about `Some(parent_path.join(selected_path))`?

helix-term/src/commands.rs
pickfire: Not sure what is the best approach here, but the current one seemed fine to me as well.

helix-term/src/commands.rs
pickfire: > Thought experiment: If you pressed gf on a random string (maybe accidentally) and it would just create a file with this
name and place you in a buffer for it, I suspect, that could get kinda annoying.

It sounds more like a feature than a bug for me.

helix-term/src/commands.rs
typetetris: Looks good. Will do, when I am back from work.

helix-term/src/commands.rs
typetetris: That's a big thing for us to decide I think.

Should we open a github discussion about it and see what others think?

It doesn't have to be settled with this PR.

helix-term/src/commands.rs
typetetris: Sorry, I don't understand, what you want to express with the eyes emoji? Am I too stubborn?

pickfire: Ah, it means I looked at your comment, but I don't have an opinion.

> Am I too stubborn?

No

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,66 @@ fn goto_file_vsplit(cx: &mut Context) {
...
+                    "Couldn't get metadata for path {}: {:?}",
+                    path.to_string_lossy(),
+                    err
+                )),
+            }
+        }
+    } else {
+        let paths: Vec<_> = selections
+            .iter()
+            .map(|r| text.slice(r.from()..r.to()).to_string())
```
archseer, line 1058: This can use the fragment method: https://github.com/helix-editor/helix/blob/4a27e2d93846f3b66ab0dc225e65d2fb869463f3/helix-core/src/selection.rs#L226

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,66 @@ fn goto_file_vsplit(cx: &mut Context) {
     goto_file_impl(cx, Action::VerticalSplit);
 }
 
+fn trim_and_resolve_path(
+    selection_text: String,
```
archseer, line 1004: This should accept &str, then you don't need to to_string() in the call below which allocates

helix-term/src/commands.rs
```diff
@@ -1000,31 +1000,66 @@ fn goto_file_vsplit(cx: &mut Context) {
...
-            text.slice(current_word.from()..current_word.to())
-                .to_string(),
-        );
-    }
-    for sel in paths {
-        let p = sel.trim();
-        if !p.is_empty() {
-            if let Err(e) = cx.editor.open(PathBuf::from(p), action) {
+        let selection_text = text
+            .slice(current_word.from()..current_word.to())
```
archseer, line 1036: Ditto here, can just use fragment

archseer (changes requested):
Generally looks fine, just some small tweaks

helix-term/src/commands.rs
the-mikedavis: We could consider adding an extra textobject for paths. I think the word textobject is a reasonable choice for the fallback behavior here though.

helix-term/src/commands.rs
sheyll: I stubbornly agree with @typetetris :) I am a new Helix user and it really feels like a bug to me, that the relative path is not resolved relative to the current document.

helix-term/src/commands.rs
pickfire: Okay, should be fine, not in this pull request.

helix-term/src/commands.rs
pickfire: I guess should be fine for now, I was probably confused back then.

sheyll:
I really miss the `gf` behviour I am used from vim a lot I would be really happy if that patch got integrated somehow...

sheyll:
So sad this is being closed... Is there any workaround that comes even remotely close to the elegance of `gf`?

Issue #1015, opened 2021-11-08
Completion sources


QiBaobin:
#### Describe your feature request
It's sweet to use the complete when there is a lsp server, but for all other cases, there is none of useful completion. Could we have some of below vim features, like completions keywords or lines from open files, I think it's also useful when lsp enabled. Thanks!

```
Completion can be done for:

1. Whole lines						|i_CTRL-X_CTRL-L|
2. keywords in the current file				|i_CTRL-X_CTRL-N|
3. keywords in 'dictionary'				|i_CTRL-X_CTRL-K|
4. keywords in 'thesaurus', thesaurus-style		|i_CTRL-X_CTRL-T|
5. keywords in the current and included files		|i_CTRL-X_CTRL-I|
6. tags							|i_CTRL-X_CTRL-]|
7. file names						|i_CTRL-X_CTRL-F|
8. definitions or macros				|i_CTRL-X_CTRL-D|
9. Vim command-line					|i_CTRL-X_CTRL-V|
10. User defined completion				|i_CTRL-X_CTRL-U|
11. omni completion					|i_CTRL-X_CTRL-O|
12. Spelling suggestions				|i_CTRL-X_s|
13. keywords in 'complete'				|i_CTRL-N| |i_CTRL-P|

```

pickfire:
I think we should keep this issue small, maybe we can start with keyword completion and close this issue, and open as we see fit.

bram209:
I find that path completion also comes in handy (currently using nvim-cmp with https://github.com/hrsh7th/cmp-path completion source enabled)

bram209:
when having multiple completion sources:
- Results need to be prioritized (if LSP is enabled, those completions will likely be better suggestions since they are context-aware)
- It may be annoying if the simple keyword completions already show up, before the slightly delayed LSP is done processing the request
- if however, you are dealing with a big codebase and a slow language server (looking at you, tsserver), it may take a while for the LSP to come with completions, in these scenario's it would come in handy to have a fallback on the simple keyword completions in the meanwhile

Some configurable threshold(s) could work, i.e. if LSP is enabled and completion request takes longer than X, show  Y completion source(s) instead

thoughts?

bram209:
@pickfire by keeping it small, are you intending to initially start with just some fallback completions in the case LSP is not loaded, or would you prefer that the completion system is set up in a way that completion sources are pluggable and multiple sources may be allowed at the same time?

pickfire:
[...]

PR #2507, opened 2022-05-19
Add support for multiple language servers per language

Philipp-M:
Adds support for multiple language servers per language.
    
Language Servers are now configured in a separate table in `languages.toml`:

```toml
[langauge-server.mylang-lsp]
command = "mylang-lsp"
args = ["--stdio"]
config = { provideFormatter = true }

[language-server.efm-lsp-prettier]
command = "efm-langserver"

[language-server.efm-lsp-prettier.config]
documentFormatting = true
languages = { typescript = [ { formatCommand ="prettier --stdin-filepath ${INPUT}", formatStdin = true } ] }
```

The language server for a language is configured like this (`typescript-language-server` is configured by default):

```toml
[[language]]
name = "typescript"
language-servers = [ { name = "efm-lsp-prettier", only-features = [ "format" ] }, "typescript-language-server" ]
```

or equivalent:

```toml
[[language]]
name = "typescript"
language-servers = [ { name = "typescript-language-server", except-features = [ "format" ] }, "efm-lsp-prettier" ]
```

Each requested LSP feature is priorized in the order of the `language-servers` array.
For example the first `goto-definition` supported language server (in this case `typescript-language-server`) will be taken for the relevant LSP request (command `goto_definition`).

If no `except-features` or `only-features` is given all features for the language server are enabled, as long as the language server supports these. If it doesn't the next language server which supports the feature is tried.

The list of supported features are:

- `format`
- `goto-definition`
- `goto-declaration`
- `goto-type-definition`
- `goto-reference`
- `goto-implementation`
- `signature-help`
- `hover`
- `document-highlight`
- `completion`
- `code-action`
- `workspace-command`
- `document-symbols`
- `workspace-symbols`
- `diagnostics`
- `rename-symbol`
- `inlay-hints`

Another side-effect/difference that comes with this PR, is that only one language server instance is started if different languages use the same language server.

PR #2608, opened 2022-05-29
Add support for path completion

Philipp-M:
This PR adds support for path completion.

It supports the following:

* Autocompletion is triggered with `/`.
* Documentation preview (file permissions, canonicalized path).
* shell expansion (e.g. `~/path`, `$HOME/path`, `${HOME}/path`)
* Async (via `spawn_blocking` instead of tokios file accessor functions, as they IMHO make the code less readable and are quite a bit slower than just spawning a "thread")
* Configurable with `editor.path-completion` (default `true`), per-language overrideable path-completion support
* Refactors `goto_file` with a new more robust path-detection/resolution mechanism and also adds shell-expansion.

A baby-step towards #1015

### DIFF
--- a/book/src/editor.md
+++ b/book/src/editor.md
@@ -32,6 +32,7 @@
 | `cursorcolumn` | Highlight all columns with a cursor | `false` |
 | `gutters` | Gutters to display: Available are `diagnostics` and `diff` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "spacer", "line-numbers", "spacer", "diff"]` |
 | `auto-completion` | Enable automatic pop up of auto-completion | `true` |
+| `path-completion` | Enable filepath completion. Show files and directories if an existing path at the cursor was recognized, either absolute or relative to the current opened document or current working directory (if the buffer is not yet saved). Defaults to true. | `true` |
 | `auto-format` | Enable automatic formatting on save | `true` |
 | `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. | `250` |
 | `completion-timeout` | Time in milliseconds after typing a word character before completions are shown, set to 5 for instant.  | `250` |
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -6,7 +6,7 @@ pub use dap::*;
 use futures_util::FutureExt;
 use helix_event::status;
 use helix_stdx::{
-    path::expand_tilde,
+    path::{self, find_paths},
     rope::{self, RopeSliceExt},
 };
 use helix_vcs::{FileChange, Hunk};
@@ -1272,53 +1272,31 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .unwrap_or_default();
 
     let paths: Vec<_> = if selections.len() == 1 && primary.len() == 1 {
-        // Secial case: if there is only one one-width selection, try to detect the
-        // path under the cursor.
-        let is_valid_path_char = |c: &char| {
-            #[cfg(target_os = "windows")]
-            let valid_chars = &[
-                '@', '/', '\\', '.', '-', '_', '+', '#', '$', '%', '{', '}', '[', ']', ':', '!',
-                '~', '=',
-            ];
-            #[cfg(not(target_os = "windows"))]
-            let valid_chars = &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '=', ':'];
-
-            valid_chars.contains(c) || c.is_alphabetic() || c.is_numeric()
-        };
-
-        let cursor_pos = primary.cursor(text.slice(..));
-        let pre_cursor_pos = cursor_pos.saturating_sub(1);
-        let post_cursor_pos = cursor_pos + 1;
-        let start_pos = if is_valid_path_char(&text.char(cursor_pos)) {
-            cursor_pos
-        } else if is_valid_path_char(&text.char(pre_cursor_pos)) {
-            pre_cursor_pos
-        } else {
-            post_cursor_pos
-        };
-
-        let prefix_len = text
-            .chars_at(start_pos)
-            .reversed()
-            .take_while(is_valid_path_char)
-            .count();
-
-        let postfix_len = text
-            .chars_at(start_pos)
-            .take_while(is_valid_path_char)
-            .count();
-
-        let path: String = text
-            .slice((start_pos - prefix_len)..(start_pos + postfix_len))
-            .into();
-        log::debug!("goto_file auto-detected path: {}", path);
-
-        vec![path]
+        let mut pos = primary.cursor(text.slice(..));
+        pos = text.char_to_byte(pos);
+        let search_start = text
+            .line_to_byte(text.byte_to_line(pos))
+            .max(pos.saturating_sub(1000));
+        let search_end = text
+            .line_to_byte(text.byte_to_line(pos) + 1)
+            .min(pos + 1000);
+        let search_range = text.slice(search_start..search_end);
+        // we also allow paths that are next to the cursor (can be ambigous but
+        // rarely so in practice) so that gf on quoted/braced path works (not sure about this
+        // but apparently that is how gf has worked historically in helix)
+        let path = find_paths(search_range, true)
+            .inspect(|mat| println!("{mat:?} {:?}", pos - search_start))
+            .take_while(|range| search_start + range.start <= pos + 1)
+            .find(|range| pos <= search_start + range.end)
+            .map(|range| Cow::from(search_range.byte_slice(range)));
+        log::debug!("goto_file auto-detected path: {path:?}");
+        let path = path.unwrap_or_else(|| primary.fragment(text.slice(..)));
+        vec![path.into_owned()]
     } else {
         // Otherwise use each selection, trimmed.
         selections
             .fragments(text.slice(..))
-            .map(|sel| sel.trim().to_string())
+            .map(|sel| sel.trim().to_owned())
             .filter(|sel| !sel.is_empty())
             .collect()
     };
@@ -1329,7 +1307,7 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
             continue;
         }
 
-        let path = expand_tilde(Cow::from(PathBuf::from(sel)));
+        let path = path::expand(&sel);
         let path = &rel_path.join(path);
         if path.is_dir() {
             let picker = ui::file_picker(path.into(), &cx.editor.config());
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -1,9 +1,11 @@
+use std::borrow::Cow;
 use std::collections::HashSet;
-use std::ffi::OsString;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
+use ::url::Url;
 use arc_swap::ArcSwap;
 use futures_util::future::BoxFuture;
 use futures_util::stream::FuturesUnordered;
@@ -16,12 +18,11 @@ use helix_event::{
 };
 use helix_lsp::lsp;
 use helix_lsp::util::pos_to_lsp_pos;
-use helix_stdx::path::{canonicalize, fold_home_dir};
-use helix_stdx::rope::{self, RopeSliceExt};
+use helix_stdx::path::{self, canonicalize, fold_home_dir, get_path_suffix};
+use helix_stdx::rope::RopeSliceExt;
 use helix_view::document::{Mode, SavePoint};
 use helix_view::handlers::lsp::CompletionEvent;
 use helix_view::{DocumentId, Editor, ViewId};
-use once_cell::sync::Lazy;
 use tokio::sync::mpsc::Sender;
 use tokio::time::Instant;
 use tokio_stream::StreamExt;
@@ -311,75 +312,42 @@ fn path_completion(
         return None;
     }
 
-    // TODO find a good regex for most use cases (especially Windows, which is not yet covered...)
-    // currently only one path match per line is possible in unix
-    static PATH_REGEX: Lazy<rope::Regex> =
-        Lazy::new(|| rope::Regex::new(r"((?:~|\$HOME|\$\{HOME\})?(?:\.{0,2}/)+.*)$").unwrap());
-
     let cur_line = text.char_to_line(cursor);
     let start = text.line_to_char(cur_line).max(cursor.saturating_sub(1000));
-    let line_until_cursor = text.slice(start..cursor).regex_input_at(..);
+    let line_until_cursor = text.slice(start..cursor);
 
-    let (dir_path, typed_file_name) = PATH_REGEX.search(line_until_cursor).and_then(|m| {
-        let start_byte = text.char_to_byte(start);
-        let matched_path = &text
-            .byte_slice((start_byte + m.start())..(start_byte + m.end()))
-            .to_string();
-
-        // resolve home dir (~/, $HOME/, ${HOME}/) on unix
-        #[cfg(unix)]
-        let mut path = {
-            static HOME_DIR: Lazy<Option<OsString>> = Lazy::new(|| std::env::var_os("HOME"));
-
-            let home_resolved_path = if let Some(home) = &*HOME_DIR {
-                let first_separator_after_home = if matched_path.starts_with("~/") {
-                    Some(1)
-                } else if matched_path.starts_with("$HOME") {
-                    Some(5)
-                } else if matched_path.starts_with("${HOME}") {
-                    Some(7)
-                } else {
-                    None
-                };
-                if let Some(start_char) = first_separator_after_home {
-                    let mut path = home.to_owned();
-                    path.push(&matched_path[start_char..]);
-                    path
-                } else {
-                    matched_path.into()
-                }
+    let (dir_path, typed_file_name) =
+        get_path_suffix(line_until_cursor, false).and_then(|matched_path| {
+            let matched_path = Cow::from(matched_path);
+            let path: Cow<_> = if matched_path.starts_with("file://") {
+                Url::from_str(&matched_path)
+                    .ok()
+                    .and_then(|url| url.to_file_path().ok())?
+                    .into()
             } else {
-                matched_path.into()
+                Path::new(&*matched_path).into()
             };
-            PathBuf::from(home_resolved_path)
-        };
-        #[cfg(not(unix))]
-        let mut path = PathBuf::from(matched_path);
-
-        if path.is_relative() {
-            if let Some(doc_path) = doc.path().and_then(|dp| dp.parent()) {
-                path = doc_path.join(path);
-            } else if let Ok(work_dir) = std::env::current_dir() {
-                path = work_dir.join(path);
+            let path = path::expand(&path);
+            let parent_dir = doc.path().and_then(|dp| dp.parent());
+            let path = match parent_dir {
+                Some(parent_dir) if path.is_absolute() => parent_dir.join(&path),
+                _ => path.into_owned(),
+            };
+            let ends_with_slash = matches!(matched_path.as_bytes().last(), Some(b'/' | b'\\'));
+            // check if there are chars after the last slash, and if these chars represent a directory
+            match std::fs::metadata(path.clone()).ok() {
+                Some(m) if m.is_dir() && ends_with_slash => {
+                    Some((PathBuf::from(path.as_path()), None))
+                }
+                _ if !ends_with_slash => path.parent().map(|parent_path| {
+                    (
+                        PathBuf::from(parent_path),
+                        path.file_name().and_then(|f| f.to_str().map(String::from)),
+                    )
+                }),
+                _ => None,
             }
-        }
-        let ends_with_slash = match matched_path.chars().last() {
-            Some(std::path::MAIN_SEPARATOR) => true,
-            None => return None,
-            _ => false,
-        };
-        // check if there are chars after the last slash, and if these chars represent a directory
-        match std::fs::metadata(path.clone()).ok() {
-            Some(m) if m.is_dir() && ends_with_slash => Some((PathBuf::from(path.as_path()), None)),
-            _ if !ends_with_slash => path.parent().map(|parent_path| {
-                (
-                    PathBuf::from(parent_path),
-                    path.file_name().and_then(|f| f.to_str().map(String::from)),
-                )
-            }),
-            _ => None,
-        }
-    })?;
+        })?;
 
     // The async file accessor functions of tokio were considered, but they were a bit slower
     // and less ergonomic than just using the std functions in a separate "thread"
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -1,5 +1,4 @@
 use std::collections::HashSet;
-use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -8,9 +7,7 @@ use futures_util::stream::FuturesUnordered;
 use futures_util::FutureExt;
 use helix_core::chars::char_is_word;
 use helix_core::syntax::LanguageServerFeature;
-use helix_event::{
-    cancelable_future, cancelation, register_hook, send_blocking, CancelRx, CancelTx,
-};
+use helix_event::{cancelable_future, register_hook, send_blocking, TaskController, TaskHandle};
 use helix_lsp::lsp;
 use helix_lsp::util::pos_to_lsp_pos;
 use helix_stdx::rope::RopeSliceExt;
@@ -59,12 +56,8 @@ pub(super) struct CompletionHandler {
     /// currently active trigger which will cause a
     /// completion request after the timeout
     trigger: Option<Trigger>,
-    /// A handle for currently active completion request.
-    /// This can be used to determine whether the current
-    /// request is still active (and new triggers should be
-    /// ignored) and can also be used to abort the current
-    /// request (by dropping the handle)
-    request: Option<CancelTx>,
+    in_flight: Option<Trigger>,
+    task_controller: TaskController,
     config: Arc<ArcSwap<Config>>,
 }
 
@@ -72,8 +65,9 @@ impl CompletionHandler {
     pub fn new(config: Arc<ArcSwap<Config>>) -> CompletionHandler {
         Self {
             config,
-            request: None,
+            task_controller: TaskController::new(),
             trigger: None,
+            in_flight: None,
         }
     }
 }
@@ -86,6 +80,9 @@ impl helix_event::AsyncHook for CompletionHandler {
         event: Self::Event,
         _old_timeout: Option<Instant>,
     ) -> Option<Instant> {
+        if self.in_flight.is_some() && !self.task_controller.is_running() {
+            self.in_flight = None;
+        }
         match event {
             CompletionEvent::AutoTrigger {
                 cursor: trigger_pos,
@@ -96,7 +93,7 @@ impl helix_event::AsyncHook for CompletionHandler {
                 // but people may create weird keymaps/use the mouse so lets be extra careful
                 if self
                     .trigger
-                    .as_ref()
+                    .or(self.in_flight)
                     .map_or(true, |trigger| trigger.doc != doc || trigger.view != view)
                 {
                     self.trigger = Some(Trigger {
@@ -109,7 +106,7 @@ impl helix_event::AsyncHook for CompletionHandler {
             }
             CompletionEvent::TriggerChar { cursor, doc, view } => {
                 // immediately request completions and drop all auto completion requests
-                self.request = None;
+                self.task_controller.cancel();
                 self.trigger = Some(Trigger {
                     pos: cursor,
                     view,
@@ -119,7 +116,6 @@ impl helix_event::AsyncHook for CompletionHandler {
             }
             CompletionEvent::ManualTrigger { cursor, doc, view } => {
                 // immediately request completions and drop all auto completion requests
-                self.request = None;
                 self.trigger = Some(Trigger {
                     pos: cursor,
                     view,
@@ -132,21 +128,21 @@ impl helix_event::AsyncHook for CompletionHandler {
             }
             CompletionEvent::Cancel => {
                 self.trigger = None;
-                self.request = None;
+                self.task_controller.cancel();
             }
             CompletionEvent::DeleteText { cursor } => {
                 // if we deleted the original trigger, abort the completion
-                if matches!(self.trigger, Some(Trigger{ pos, .. }) if cursor < pos) {
+                if matches!(self.trigger.or(self.in_flight), Some(Trigger{ pos, .. }) if cursor < pos)
+                {
                     self.trigger = None;
-                    self.request = None;
+                    self.task_controller.cancel();
                 }
             }
         }
         self.trigger.map(|trigger| {
             // if the current request was closed forget about it
             // otherwise immediately restart the completion request
-            let cancel = self.request.take().map_or(false, |req| !req.is_closed());
-            let timeout = if trigger.kind == TriggerKind::Auto && !cancel {
+            let timeout = if trigger.kind == TriggerKind::Auto {
                 self.config.load().editor.completion_timeout
             } else {
                 // we want almost instant completions for trigger chars
@@ -161,17 +157,17 @@ impl helix_event::AsyncHook for CompletionHandler {
 
     fn finish_debounce(&mut self) {
         let trigger = self.trigger.take().expect("debounce always has a trigger");
-        let (tx, rx) = cancelation();
-        self.request = Some(tx);
+        self.in_flight = Some(trigger);
+        let handle = self.task_controller.restart();
         dispatch_blocking(move |editor, compositor| {
-            request_completion(trigger, rx, editor, compositor)
+            request_completion(trigger, handle, editor, compositor)
         });
     }
 }
 
 fn request_completion(
     mut trigger: Trigger,
-    cancel: CancelRx,
+    handle: TaskHandle,
     editor: &mut Editor,
     compositor: &mut Compositor,
 ) {
@@ -202,7 +198,6 @@ fn request_completion(
     // necessary from our side too.
     trigger.pos = cursor;
     let trigger_text = text.slice(..cursor);
-    let cancel_completion = Arc::new(AtomicBool::new(false));
 
     let mut seen_language_servers = HashSet::new();
     let mut futures: FuturesUnordered<_> = doc
@@ -270,12 +265,7 @@ fn request_completion(
             }
             .boxed()
         })
-        .chain(path_completion(
-            cursor,
-            text.clone(),
-            doc,
-            cancel_completion.clone(),
-        ))
+        .chain(path_completion(cursor, text.clone(), doc, handle.clone()))
         .collect();
 
     let future = async move {
@@ -296,18 +286,13 @@ fn request_completion(
     let ui = compositor.find::<ui::EditorView>().unwrap();
     ui.last_insert.1.push(InsertEvent::RequestCompletion);
     tokio::spawn(async move {
-        let items = match cancelable_future(future, cancel).await {
-            Some(v) => v,
-            None => {
-                cancel_completion.store(true, std::sync::atomic::Ordering::Relaxed);
-                Vec::new()
-            }
-        };
-        if items.is_empty() {
+        let items = cancelable_future(future, &handle).await;
+        let Some(items) = items.filter(|items| !items.is_empty()) else {
             return;
-        }
+        };
         dispatch(move |editor, compositor| {
-            show_completion(editor, compositor, items, trigger, savepoint)
+            show_completion(editor, compositor, items, trigger, savepoint);
+            drop(handle)
         })
         .await
     });
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -369,13 +369,16 @@ pub fn trigger_auto_completion(
                     }) if triggers.iter().any(|trigger| text.ends_with(trigger)))
         });
 
-    let trigger_path_completion = matches!(
-        text.get_bytes_at(text.len_bytes())
-            .and_then(|t| t.reversed().next()),
-        Some(b'/' | b'\\')
-    ) && doc.path_completion_enabled();
+    let cursor_char = text
+        .get_bytes_at(text.len_bytes())
+        .and_then(|t| t.reversed().next());
 
-    if is_trigger_char || trigger_path_completion {
+    #[cfg(windows)]
+    let is_path_completion_trigger = matches!(cursor_char, Some(b'/' | b'\\'));
+    #[cfg(not(windows))]
+    let is_path_completion_trigger = matches!(cursor_char, Some(b'/'));
+
+    if is_trigger_char || (is_path_completion_trigger && doc.path_completion_enabled()) {
         send_blocking(
             tx,
             CompletionEvent::TriggerChar {
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -1,5 +1,4 @@
 use std::collections::HashSet;
-use std::ffi::OsStr;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Duration;
@@ -397,20 +396,7 @@ fn path_completion(
 
         read_dir
             .filter_map(Result::ok)
-            .filter_map(|dir_entry| {
-                let path = dir_entry.path();
-                // check if <chars> in <path>/<chars><cursor> matches the start of the filename
-                let filename_starts_with_prefix =
-                    match (path.file_name().and_then(OsStr::to_str), &typed_file_name) {
-                        (Some(re_stem), Some(t)) => re_stem.starts_with(t),
-                        _ => true,
-                    };
-                if filename_starts_with_prefix {
-                    dir_entry.metadata().ok().map(|md| (path, md))
-                } else {
-                    None
-                }
-            })
+            .filter_map(|dir_entry| dir_entry.metadata().ok().map(|md| (dir_entry.path(), md)))
             .map(|(path, md)| {
                 let file_name = path
                     .file_name()
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -493,7 +493,7 @@ fn path_completion(
                     #[cfg(not(unix))]
                     {
                         format!(
-                            "type: `{path_kind}`\n\
+                            "type: `{kind}`\n\
                              {resolved}full path: `{full_path_name}`",
                         )
                     }
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -1,31 +1,26 @@
-use std::borrow::Cow;
 use std::collections::HashSet;
-use std::path::{Path, PathBuf};
-use std::str::FromStr;
+use std::sync::atomic::AtomicBool;
 use std::sync::Arc;
 use std::time::Duration;
 
-use ::url::Url;
 use arc_swap::ArcSwap;
-use futures_util::future::BoxFuture;
 use futures_util::stream::FuturesUnordered;
 use futures_util::FutureExt;
 use helix_core::chars::char_is_word;
 use helix_core::syntax::LanguageServerFeature;
-use helix_core::Transaction;
 use helix_event::{
     cancelable_future, cancelation, register_hook, send_blocking, CancelRx, CancelTx,
 };
 use helix_lsp::lsp;
 use helix_lsp::util::pos_to_lsp_pos;
-use helix_stdx::path::{self, canonicalize, fold_home_dir, get_path_suffix};
 use helix_stdx::rope::RopeSliceExt;
 use helix_view::document::{Mode, SavePoint};
 use helix_view::handlers::lsp::CompletionEvent;
 use helix_view::{DocumentId, Editor, ViewId};
+use path::path_completion;
 use tokio::sync::mpsc::Sender;
 use tokio::time::Instant;
-use tokio_stream::StreamExt;
+use tokio_stream::StreamExt as _;
 
 use crate::commands;
 use crate::compositor::Compositor;
@@ -35,10 +30,13 @@ use crate::job::{dispatch, dispatch_blocking};
 use crate::keymap::MappableCommand;
 use crate::ui::editor::InsertEvent;
 use crate::ui::lsp::SignatureHelp;
-use crate::ui::{self, CompletionItem, LspCompletionItem, PathCompletionItem, PathKind, Popup};
+use crate::ui::{self, Popup};
 
 use super::Handlers;
+pub use item::{CompletionItem, LspCompletionItem};
 pub use resolve::ResolveHandler;
+mod item;
+mod path;
 mod resolve;
 
 #[derive(Debug, PartialEq, Eq, Clone, Copy)]
@@ -204,6 +202,7 @@ fn request_completion(
     // necessary from our side too.
     trigger.pos = cursor;
     let trigger_text = text.slice(..cursor);
+    let cancel_completion = Arc::new(AtomicBool::new(false));
 
     let mut seen_language_servers = HashSet::new();
     let mut futures: FuturesUnordered<_> = doc
@@ -271,7 +270,12 @@ fn request_completion(
             }
             .boxed()
         })
-        .chain(path_completion(cursor, text.clone(), doc))
+        .chain(path_completion(
+            cursor,
+            text.clone(),
+            doc,
+            cancel_completion.clone(),
+        ))
         .collect();
 
     let future = async move {
@@ -292,7 +296,13 @@ fn request_completion(
     let ui = compositor.find::<ui::EditorView>().unwrap();
     ui.last_insert.1.push(InsertEvent::RequestCompletion);
     tokio::spawn(async move {
-        let items = cancelable_future(future, cancel).await.unwrap_or_default();
+        let items = match cancelable_future(future, cancel).await {
+            Some(v) => v,
+            None => {
+                cancel_completion.store(true, std::sync::atomic::Ordering::Relaxed);
+                Vec::new()
+            }
+        };
         if items.is_empty() {
             return;
         }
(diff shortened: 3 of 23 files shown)


kirawi:
I think it would be better to split off the changes to the LSP into a separate PR. Unless completion is dependent upon them?

Philipp-M:
You mean all the commits that are also in #2507?

In case, I will rebase this as soon as the relevant commits this PR is dependent on are on master, currently only the last commit is relevant for this PR or different to #2507.

nrdxp:
What is left to move this out of draft status? I've been using this for over a week now and it seems to work fine.

Philipp-M:
Well for one (biggest blocker): This is currently dependent on a few commits (particularly the extension/merging of the completion menu) of #2507, and I'm unsure how to progress as I'm awaiting some feedback there.

Also support for windows paths should be added, this should be simple (by extending the path regex), but I would like to first resolve the first issue.

Another (smaller, maybe not worth to fix) issue is, that currently only one path per line is possible, this could probably be tackled with a different path-regex. But I've read that every character but `\0` is allowed in a unix filepath, so I think it's difficult to find a good regex that doesn't have (realworld) limitations.

I've also experimented with async io (via `tokio::fs`), but IMHO the performance-drop and probably race-conditions (in the completion menu, if writing to fast/slow IO) wasn't worth it to continue that path (yet).

helix-term/src/commands.rs
```diff
@@ -3942,43 +3961,188 @@ pub fn completion(cx: &mut Context) {
...
+                    // we're not in insert mode anymore
+                    return;
+                }
+                let doc = doc!(editor);
+
+                // TODO async path completion (for this probably the whole completion system has to be reworked to be async without producing race conditions)
+                let items = ui::PATH_REGEX
+                    .find(&line_until_cursor)
+                    .and_then(|matched_path| {
+                        let matched_path = matched_path.as_str();
```
cole-h, line 4037: Would it be possible to also add a denylist setting? I run NixOS, so attempting to index `/nix/store` would be extremely slow with its ~265000 files / directories.

Philipp-M: Maybe we can just use the  `.(git)ignore` for that?

Actually I've used `nix/store`  for stress testing/general feeling, and I thought it was acceptable given the amount of files (but I've got a fast CPU and NVME drive for /nix ...))

Would someone actually edit something directly in `/nix/store` or in a directory with that much files/folders?

I'm not sure of a daily usecase currently where this would really be a problem, but I'm open for different solutions (my suggestion would be a global .ignore)

cole-h: I don't _edit_ things in the Nix store, no, but I do use Helix to read files in there very frequently. I'd rather not use a .ignore for that because so many other tools read that, and I'd like to set this only in Helix.

nrdxp:
Did something change? After rebasing on the last push path completion seems to have just stopped working all together. Do I need to configure something differently?

Philipp-M:
Have you tried using the HEAD of this branch?

My personal fork (rebased on this branch) is still working.

It's a little bit out of sync with master, I will try to find time soon to rebase this and the multiple-language-servers PR onto master.

nrdxp:
Oops, I figured it out. It was my own fault, I forgot to push my latest changes of my forked branch (based on this branch) to the remote so my system never pulled my rebase. Apologies :pray:

danielgomez3:
Any status on when this might be merged to main?

Philipp-M:
This is not yet ready to merge and needs a major rebase/refactor and support for windows. I'll wait for this at least until multiple language servers support is merged.
I'm also not completely sure if it makes sense to put this into core (could be a super simple language server, that can be enabled for all languages).

nrdxp:
Now that multi-language support is merged we should either close this or rebase and merge. I would vote for the later since I think a consistent path completion baked right into the editor makes sense.

Philipp-M:
I have now rebased (or more like rewritten) it onto master.

It now uses the existing `lsp::CompletionItem` and adds a new field enum field `source` to `ui::CompletionItem` (that might be extended when more [completion sources](https://github.com/helix-editor/helix/issues/1015) will be implemented, instead of previously a custom type.
I think this is a little bit more portable (could even be used with little modification for a dedicated language server) and requires fewer changes in `completion.rs`. It's generally a cleaner and improved implementation, and should be easier to maintain/review IMHO.

Following is new:
* documentation preview (file type, file permissions, full canonicalized path).
* home-path resolution (`~/path`)
* async (via `spawn_blocking` instead of tokios file accessor, as they IMHO make the code less readable and are quite a bit slower than just spawning a "thread")
* per-language overridable path-completion support

 I think it's now in a good state for review/merging for unix platforms at least, but windows support is still missing, I'm not sure if it should be merged without windows path support...?

I don't have access to a windows machine currently and would like to avoid setting up a VM, maybe someone else can help out here, via a PR to my branch `path-completion`? Maybe *if* I find time for it (I have a few other ongoing projects, and it's not a high priority for me), I'll write some integration tests for windows paths (and for unix...). Help/support here is also appreciated.

helix-term/src/commands.rs
Philipp-M: Right now it's less of a blocker (literally :)) for typing, as it's now async and results will be discarded when not relevant anymore (i.e. the user has typed something that doesn't match the completion item).

On a fast computer it's still *acceptable* IMHO, especially if the folder was indexed once (i.e. is likely cached in memory when accessing again)).

If this is still relevant, can you/someone maybe think/prototype how the UX/configuration for this look like (I'm happy to implement it, but I'm not sure how the UI for this should look like).

But the (completion-)menu is quite slow with > ~100k entries that are sorted synchronously every keystroke.
I think a more sophisticated lazier/async/"streamable" menu/picker implementation would be probably help in general (I still think that the behavior is so similar (also in the future), that the implementations should be shared, but that's a different topic...).

Philipp-M:
The tests for OSX failed previously (now they run through, pushed a fix), due to `libc::mode_t` being 16 bit on OSX (is this correct?). Since this is only tested on linux, maybe someone with OSX should test this (which I also don't have access to currently). This is relevant for the file permissions in the documentation window, which should look something like `[rwxr-xr--]`.

inalone:
Would love for this to be merged, this is the main feature I'm missing from my neovim config.

nrdxp:
just an FYI, since this is the only thing stopping me from tracking the latest master, I went ahead and worked out the merge conflict in this branch for now:
https://github.com/nrdxp/helix/tree/fork3

Feel free to steal it to quickly fix the conflicts in this PR branch

Philipp-M:
Thanks for doing this! I have rebased this on master as well (merge conflicts are fortunately maintainable).

Well as said, feedback for OSX especially would be appreciated. Whether Windows needs to be supported within this PR as well (but my feeling is, that it should be, but I'm not super keen in adding it, I'd appreciate a PR on top of this PR).

One thing I noticed after implementing this, is that there's similar code already within `helix_term::ui::completers` it's slightly different though, so it may or may not be worth it to share the code between (in-editor-view) completion sources...

Advantage for the current code is, that it's quite portable though.

FWIW I'm maintaining my own fork with the new event-system, rainbow-treesitter matches, and a few other smaller fixes (possibly necessary due to merges) in https://github.com/Philipp-M/helix/tree/personal/

nilsherzig:
Nice work :) 

Do you think there would be any problems if the autocompletion would add a '/' at the end of a directory name? This would enable a much smoother "autocompletion chain" for files deep inside some nested folders.

eliliam:
Once we get that linting error fixed, is this good to go? I'd love to see this merged in since it seems like a really great implementation of path autocomplete!

sn0n:
any updates? I think this is the functionality I'm looking for,.. new to helix amd figure out gf for goto file, but cant figure out the 'insert file name here from this list' type path completion. sorry to bother.

daedroza:
@Philipp-M : Hey, can this PR be updated with newest HEAD? The place where we hook isn't commands anymore but in separate completion handler.

Philipp-M:
Rebased onto master.

> if the autocompletion would add a '/' at the end of a directory name?

I guess that could be done, possibly configurable, I like the quick retrigger via `/` though (while not triggering a new directory while selecting the item in the menu). Not sure about it...

> any updates?

Well as described in my last comment I'm awaiting general feedback.

nrdxp:
I don't think we have proper visibility on this PR, and poor Phillip has been force pushing updates for a few years now. I've been using this patch from the close to the beginning and haven't had any issues with it. Sorry if the ping is inappropriate but @archseer or @the-mikedavis do you guys have an idea of what is missing or why this hasn't been merged in so long?

Or has it just been overlooked?

lukeflo:
Thanks for the great work. I just forked the main Helix repo and rebased your PR as well as another PR (#11164) on my forks master. Just built it from there and works great. Was one of the major things which hindered me switching completely to Helix.

Hope it will find its way into the main repo soon.

One thing I encountered is that expanding Tilde (`~`) for home dir works fine, but the environment variable `$HOME` isn't expanded at all. Is there a reason for that or am I missing something concerning the concrete use-case?

eliliam (approved):
Looks great to me, who are the main people we need to review this to get this finally merged in?

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +299,223 @@ fn request_completion(
...
+        return None;
+    }
+
+    use helix_lsp::util::range_to_lsp_range;
+    // TODO find a good regex for most use cases (especially Windows, which is not yet covered...)
+    // currently only one path match per line is possible in unix
+    static PATH_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(r"((?:~)?(?:\.{0,2}/)+.*)$").unwrap());
+
+    let cur_line = text.char_to_line(cursor);
+    let begin_line = text.line_to_char(cur_line);
```
the-mikedavis, line 317: I wonder if we should cap this to some reasonable maximum (maybe some N hundred characters) to prevent degrading the search if you have very long single lines.

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +299,223 @@ fn request_completion(
...
+    let line_until_cursor = text
+        .slice(begin_line..cursor)
+        .as_str()
+        .unwrap_or_default()
+        .to_owned();
```
the-mikedavis, line 322: With #9422 it should be possible to operate on the ropeslice directly now. The regex uses `$` though which doesn't work the same way with regex-cursor so it may not be possible to use it yet - regex-cursor and upstream regex might need the `\z` capture to detect the end of the slice

book/src/generated/typable-cmd.md
```diff
@@ -2,7 +2,7 @@
 | --- | --- |
 | `:quit`, `:q` | Close the current view. |
 | `:quit!`, `:q!` | Force close the current view, ignoring unsaved changes. |
-| `:open`, `:o` | Open a file from disk into the current view. |
+| `:open`, `:o`, `:edit`, `:e` | Open a file from disk into the current view. |
```
the-mikedavis, line 5: Looks like this snuck in from a merge/rebase

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +299,223 @@ fn request_completion(
...
+
+                if path.is_relative() {
+                    if let Some(doc_path) = doc.path().and_then(|dp| dp.parent()) {
+                        path = doc_path.join(path);
+                    } else if let Ok(work_dir) = std::env::current_dir() {
+                        path = work_dir.join(path);
+                    }
+                }
+                let ends_with_slash = match matched_path.chars().last() {
+                    Some('/') => true, // TODO support Windows
```
the-mikedavis, line 356: Can this be covered with just using the `std::path::MAIN_SEPARATOR`?

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +299,223 @@ fn request_completion(
...
+                let edit_diff = typed_file_name.as_ref().map(|f| f.len()).unwrap_or(0);
+
+                let text_edit = Some(lsp::CompletionTextEdit::Edit(lsp::TextEdit {
+                    range: range_to_lsp_range(
+                        &text,
+                        Range::new(cursor - edit_diff, cursor),
+                        OffsetEncoding::default(),
+                    ),
+                    new_text: file_name.clone(),
+                }));
```
the-mikedavis, line 493: I'd like to try to find a way to avoid digging ourselves into lsp-types here. It makes for a pretty small+focused change but I think we should be introducing core types like `helix_core::Diagnostic` where possible or at least wrappers/replacements like #11486. I suspect it's a potentially large refactor but I think this is the right feature to introduce it with since the actual searching/completion part of this is quite small. (Compared to completing words from the buffer for example)

helix-term/src/handlers/completion.rs
```diff
@@ -346,7 +571,10 @@ pub fn trigger_auto_completion(
                         ..
                     }) if triggers.iter().any(|trigger| text.ends_with(trigger)))
         });
-    if is_trigger_char {
+
+    let trigger_path_completion = text.ends_with("/") && doc.supports_path_completion();
```
the-mikedavis, line 575: Same here with maybe being able to use `std::path::MAIN_SEPARATOR`

helix-term/src/handlers/completion.rs
the-mikedavis: I haven't looked into this much but it might be pretty simple to update the `CompletionItem` struct to have an enum between the `lsp::CompletionItem` and a `helix_core::CompletionItem` struct that has a changeset or transaction instead of text_edits

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +299,223 @@ fn request_completion(
...
+                                (libc::S_IWGRP, 'w'),
+                                (libc::S_IXGRP, 'x'),
+                                (libc::S_IROTH, 'r'),
+                                (libc::S_IWOTH, 'w'),
+                                (libc::S_IXOTH, 'x'),
+                            ]
+                            .iter()
+                            .fold(String::new(), |mut acc, (p, s)| {
+                                #[allow(clippy::unnecessary_cast)]
+                                acc.push(if mode & (*p as u32) > 0 { *s } else { '-' });
```
the-mikedavis, line 461:
```suggestion
                                acc.push(if mode & p > 0 { *s } else { '-' });
```

it looks like the compiler derefs this automatically now

the-mikedavis: Actually I guess you could use `into_iter` instead of iter here and also drop the deref for `s`

helix-term/src/handlers/completion.rs
Philipp-M: Actually a previous iteration of this PR (e.g. [this](https://github.com/helix-editor/helix/tree/12c088918405b2debd04d259293489f7a1fc539d)) had an enum `CompletionItem` with separating variants `LSP` and `Path` .
I wasn't sure having a separate mechanism additional to `lsp::CompletionItem` would make sense, since there was already the variants `lsp::CompletionItemKind::{FILE, FOLDER}`.

But I guess when thinking about more completion sources it may make sense, especially when the move to more core types resembling those in `lsp` is planned anyways.

So when I'm getting you right, you mean a `CompletionItem` at a lower-level than in `ui::completion` (which I think was the previous iteration). I wonder if it makes sense to generally abstract over `lsp::CompletionItem` and wrap it in `helix_core::CompletionItem`. I should probably investigate this, as I think you're right in that this would need a larger refactor (*maybe* worth it).

I think a `Transaction` would make more sense here, being more general (and we have `impl From<ChangeSet> for Transaction` already)?

Philipp-M: Ok, I don't think abstracting over `lsp::CompletionItem` makes sense with the current architecture. So rather having a "data" container `helix_core::CompletionItem` similar to `helix_core::Diagnostic` a little bit independent of `lsp::CompletionItem` with data like documentation (possibly lazy) and `Transaction` etc. makes probably more sense for now.

Is it planned btw. to move more logic of this kind from `helix-term` to e.g. `helix-view` or even `helix-core`?
I could imagine `helix_term::ui::CompletionItem` would make more sense in `helix-view`?

the-mikedavis: Yeah generally the more we can move from -term to -view the merrier. Ideally most of what lives in -term right now would be in -view (even commands) but it will take quite a bit of work to decouple the UI elements. So -term would just be the terminal backend and then there could be a -gui crate that would also implement the UI components.

Philipp-M: Ok, good to know, I'm actually thinking about starting that refactor in this PR, that would obviously blow up this PR a little bit, and would likely produce merge-conflicts with other PRs (e.g. #9996), which is why I'm not sure about it, as I'd like to avoid resolving merge-conflicts and/or maintain this PR for a longer time (in a state where this is more interconnected with the rest of the editor-logic).

Like putting a lot of logic of `helix_term::ui::completion` into e.g. `helix-view::completion` as I don't see many dependencies to `helix-term` currently. (Same for the handlers I guess).

But maybe keep it more incremental, and leave the `CompletionItem` (and the path/lsp `CompletionItem` resolution logic) currently in `helix-term`, i.e. only `helix_term::ui::completion::CompletionItem` to an enum roughly like this:

```rust
pub enum PathKind { Folder, File, Link } // maybe not worth it...

pub enum CompletionItem {
    Lsp {
        item: lsp::CompletionItem,
        id: LanguageServerId,
        resolved: bool,
    },
    Path { // Not sure if this should be a more generic `Core`,
           // but this probably allows being more flexible in the future
        kind: PathKind,
        item: helix_core::CompletionItem,
    },
}

pub struct helix_core::CompletionItem {
    pub transaction: Transaction,
    pub label: String,
    /// Containing Markdown, this could be turned e.g. into an enum to be more structured
    pub documentation: String,
}
```

the-mikedavis: Hmm yeah, working incrementally is probably the right way to go about this. Moving more things to helix-view would mainly help the GUI effort which I don't believe anyone is working on actively anyways. I'm also worried about conflicts with #9801 (though on second look it doesn't look like it would conflict much). An enum like that looks good even if it's only a stepping stone to a bigger refactor later.

Philipp-M: Actually I have merged the snipped system with path-completion in my personal branch, I don't remember that this caused any serious merge-conflicts...

Ok, good to know, then I'll go the incremental route (when I find the time).

helix-term/src/handlers/completion.rs
Philipp-M: Good point, I made the limit now `-1000` compared to the current cursor.

helix-term/src/handlers/completion.rs
Philipp-M: Thanks for the suggestion. I'm using now the `regex-cursor`, as you say it doesn't support `$` the same way as `regex`, so I worked around that. But since the path detection is not perfect anyway (difficult, since everything apart of `\0` is allowed in unix paths). I think this is good enough for now.

helix-term/src/handlers/completion.rs
Philipp-M: Yeah `into_iter` is possible, also added `String::with_capacity(9)` to avoid unnecessary allocs...

Btw. the cast is necessary for e.g. OSX (added a comment as well) as there is `p: u16`.

Philipp-M:
Ok, did a bit of refactoring to the `CompletionItem::{Lsp, Path}` enum. I tried to keep the diff small, but it obviously grew a little bit, since the additional variant needs a little bit different handling.

I've added the basically unchanged
```rust
struct helix_core::CompletionItem {
    pub transaction: Transaction,
    pub label: String,
    /// Containing Markdown
    pub documentation: String,
}
```

I think that could be extended over time when necessary.

> One thing I encountered is that expanding Tilde (`~`) for home dir works fine, but the environment variable `$HOME` isn't expanded (means recognized as e.g. `/home/user`) at all. Is there a reason for that or am I missing something concerning the concrete use-case?

I've also added support for this along the way.

Philipp-M:
Hmm nightly rustc seems to be able to handle a few more borrow-checking cases it seems, something like:

```rust
let transaction = match item {
    Arm1 { item, .. } => &item_to_transaction(item),
    Arm2 { item, .. } => &item.transaction
};
```
doesn't seem to be possible without nightly...

Philipp-M:
Btw. I have also refactored the handling of additional text edits of lsp completion, composing it now as a single `Transaction` (as it was a simple change, and there was a `TODO` comment anyway for this).

Philipp-M:
> Btw. I have also refactored the handling of additional text edits of lsp completion, composing it now as a single `Transaction`

Well I just reverted that change again... I thought compose handles different resulting `ChangeSet::len`s, which is not the case. So that would be more complicated, and should probably be handled in a different PR.

the-mikedavis:
Interesting that nightly allows borrowing that value. An alternative could be wrapping it in a `Cow::Owned` and the path variant uses `Cow::Borrowed` but passing it to `Document::apply_temporary` in both branches is good too.

helix-term/src/handlers/completion/resolve.rs
```diff
@@ -26,6 +27,15 @@ pub struct ResolveHandler {
     resolver: Sender<ResolveRequest>,
 }
 
+macro_rules! lsp_variant {
```
the-mikedavis, line 30: I'd like to avoid the macro here if possible - I think we can be explicit in the places where we want to match the `CompletionItem::Lsp` variant instead. For `ensure_item_resolved` that means we need to move the `if self.last_request.as_deref()...` check to the top of the function because of the mutable borrow of `item` and then we can say

```rust
let CompletionItem::Lsp(lsp_item) = item else {
    return;
}
```

and we can unconditionally pass the `item` to `ensure_item_resolved` in `helix-term/src/ui/completion.rs`. In the `ResolveRequest::execute` function we can do basically the same but use `unreachable!` (the same as what you have now but inlining the macro)

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +301,226 @@ fn request_completion(
...
+                let filename_starts_with_prefix =
+                    match (path.file_name().and_then(OsStr::to_str), &typed_file_name) {
+                        (Some(re_stem), Some(t)) => re_stem.starts_with(t),
+                        _ => true,
+                    };
+                if filename_starts_with_prefix {
+                    dir_entry.metadata().ok().map(|md| (path, md))
+                } else {
+                    None
+                }
```
the-mikedavis, line 412: I believe we want to skip the prefix check here and delegate the filtering to the ui::Menu and Nucleo. This makes a difference if you trigger completion manually, for example if you're in the helix repo and type `./lsp` you'll see `helix-lsp` and `helix-lsp-types`. If you then hit `C-c` and `C-x` to manually trigger completion though you won't have any suggestions because of the prefix check.

So instead this could be `.filter_map(|dir_entry| dir_entry.metadata().ok().map(|md| (dir_entry.path(), md)))`

helix-term/src/handlers/completion/resolve.rs
Philipp-M: Before I'm repeating work again, do you mean something like [this](https://github.com/helix-editor/helix/pull/2608/commits/cc19db1e5306e6cbb396ae33c51e99927e879584#diff-3626b17eb15f4783dd83a9f709ecef3bd3f591db4b22d83ee2603ff0a09d49e5) before that micro-optimization (I wasn't sure about it anyway, as it just saves 1 clone...)

Philipp-M: Yeah an `impl PartialEq<CompletionItem> for LspCompletionItem` is much better than that macro, for whatever reason, I haven't thought of that first...

I've reverted to that previous change, and added `PartialEq` impls.

helix-term/src/handlers/completion.rs
Philipp-M: Yeah makes sense, it also allows case-insensitive completion I think.

kik4444:
Not sure if anyone's mentioned this before, but have you considered taking inspiration or logic from the [nvim cmp-path plugin](https://github.com/hrsh7th/cmp-path)? As far as I've noticed it doesn't attempt any complexities like Home-path resolution except the tilde `~`.

Philipp-M:
I have not, but since it's implemented already (and it's not really complex IMO), I guess we can keep support for resolution of `$HOME` and `${HOME}`?

kik4444:
Well if it works it works. Either way I'd just be grateful to have this as part of the main editor. This feature is on the list of things I'd like to have before giving Helix a more serious attempt.

nrdxp:
> Well if it works it works. Either way I'd just be grateful to have this as part of the main editor. This feature is on the list of things I'd like to have before giving Helix a more serious attempt.

fwiw I have been using Helix for years now and just carrying this along as a patch pretty much that whole time. It works and I've never had an issue with it. Hope it get's merged soon

RGBCube:
Seems like all reviews are passing and the code works, maybe it's time to finally merge this ~2 year old MR?

David-Else:
I am using it and have had no problems, could not live without it now.

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+            .filter_map(|dir_entry| dir_entry.metadata().ok().map(|md| (dir_entry.path(), md)))
+            .map(|(path, md)| {
+                let file_name = path
+                    .file_name()
+                    .unwrap_or_default()
+                    .to_string_lossy()
+                    .to_string();
+
+                let full_path = path.canonicalize().unwrap_or_default();
+                let full_path_name = full_path.to_string_lossy().into_owned();
```
the-mikedavis, line 408: For canonicalization we generally prefer `helix_stdx::path::canonicalize` now (it doesn't expand symlinks like the canonicalize in `std`). For the full path name that we display in the documentation let's pass that canonicalized path through `helix_stdx::path::fold_home_dir` so it displays better

the-mikedavis (commented):
I only have some minor style nits and a comment about canonicalization and display, otherwise I think this looks good :+1:

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+    static PATH_REGEX: Lazy<rope::Regex> =
+        Lazy::new(|| rope::Regex::new(r"((?:~|\$HOME|\$\{HOME\})?(?:\.{0,2}/)+.*)").unwrap());
+
+    let cur_line = text.char_to_line(cursor);
+    let start = text.line_to_char(cur_line).max(cursor.saturating_sub(1000));
+    let line_until_cursor = text.slice(..).regex_input_at(start..cursor);
+
+    let (dir_path, typed_file_name) = PATH_REGEX
+        .find_iter(line_until_cursor)
+        .last() // this is a workaround for missing "match end of slice" support in `rope::Regex`
```
pascalkuthe, line 323: regex cursor works the same way as regex automata with regards to anchors/search ranges. What you need to do is anchor you regex (put `$` at the end) and instead of using `text.slice(...).regex_input_at(start..cursor` us `text.slice(start..cursor).regex_input_at(..)`. 

The difference is that `regex_input_at(start..end)` runs a regex within a part of a string/rope but still takes the surrounding context into account for lookaround assertions (so `\b` works correctly and so `^` only matches start/end of entire string and not the start of the search range).

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+        .and_then(|matched_path| {
+            let end = text.byte_to_char(matched_path.end());
+            if end != cursor {
+                return None;
+            }
+            let start = text.byte_to_char(matched_path.start());
+            let matched_path = &text.slice(start..end).to_string();
```
pascalkuthe, line 330: nit: just do `text.byte_slice(matched_pathn.start()..matched_path.end())` instead of converting back and fourth a bunch of times

helix-term/src/handlers/completion.rs
Philipp-M: Hmm, I wonder though, if a link *should* be shown resolved then or not (because then I think we still need `std::path::canonicalize`), which IMO makes sense but I have no strong opinion here...

the-mikedavis: I think we should try to avoid expanding symlinks in the parent directories as std's canonicalize would do. For an edge case, my `~/src/` directory is a symlink to `/nix/persist/home/michael/src` so canonicalizing a path like `./helix-event` would expand all the way to `/nix/persist/home/michael/src/helix/hx/helix-event`. If the directory entry that we are looking at is a symlink though we should probably follow that link (once?).

Philipp-M: Yeah it's what I'd expect, but I see your point.

I think following a link once, would result in otherwise inconsistent behavior but I'm not sure. So I'd rather not follow it, and display it as link then, I've pushed a version without following the link for now.

Philipp-M:
Thanks for the review, I think I handled all the feedback.

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +302,197 @@ fn request_completion(
...
+    // currently only one path match per line is possible in unix
+    static PATH_REGEX: Lazy<rope::Regex> =
+        Lazy::new(|| rope::Regex::new(r"((?:~|\$HOME|\$\{HOME\})?(?:\.{0,2}/)+.*)$").unwrap());
+
+    let cur_line = text.char_to_line(cursor);
+    let start = text.line_to_char(cur_line).max(cursor.saturating_sub(1000));
+    let line_until_cursor = text.slice(start..cursor).regex_input_at(..);
+
+    let (dir_path, typed_file_name) = PATH_REGEX.search(line_until_cursor).and_then(|m| {
+        let matched_path = &text.byte_slice(m.start()..m.end()).to_string();
```
the-mikedavis, line 324: Now that we `slice` and then use `regex_input_at` we don't have the byte indices automatically adjusted for us. (They're relative to the slice `text.slice(start..cursor)` - `regex_input_at` made them absolute to `text` previously.)

To adjust this we'll need to grab the start byte (probably easiest to do `let start_byte = text.char_to_byte(start);` after `start` is bound) and then our matched path would be `text.byte_slice(start_byte + m.start()..start_byte + m.end())`

Philipp-M: Whoops, I guess I wasn't thorough enough... thanks.

the-mikedavis (dismissed):

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+                    PathKind::File
+                };
+
+                let resolved = if kind == PathKind::Link {
+                    "resolved "
+                } else {
+                    ""
+                };
+
+                let documentation = {
```
pascalkuthe, line 428: this should also just be a seperate function

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+    let future = tokio::task::spawn_blocking(move || {
+        let Some(read_dir) = std::fs::read_dir(&dir_path).ok() else {
+            return Vec::new();
+        };
+
+        read_dir
+            .filter_map(Result::ok)
+            .filter_map(|dir_entry| dir_entry.metadata().ok().map(|md| (dir_entry.path(), md)))
+            .map(|(path, md)| {
+                let file_name = path
```
pascalkuthe, line 395: this is a bit ugly, what `dir_entry.path()` does is just `dir_path.join(dir_entry.file_name())` so going trough all this to extractd the file name again doesn't seem ideal. Instead just extract the file name and use that here (and just do `dir_path.join(file_name)` below)

helix-view/src/document.rs
```diff
@@ -1713,6 +1713,12 @@ impl Document {
         self.version
     }
 
+    pub fn supports_path_completion(&self) -> bool {
```
pascalkuthe, line 1716: this seem like the wrong name, it should be `path_completions_enabled()` (or similar)

helix-term/src/ui/completion.rs
```diff
@@ -82,18 +87,101 @@ impl menu::Item for CompletionItem {
...
+    pub item: helix_core::CompletionItem,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum CompletionItem {
+    Lsp(LspCompletionItem),
+    Path(PathCompletionItem),
+}
+
+impl PartialEq<CompletionItem> for LspCompletionItem {
```
pascalkuthe, line 158: I don't think we necessarily need these

helix-term/src/ui/completion.rs
```diff
@@ -82,18 +87,101 @@ impl menu::Item for CompletionItem {
...
+#[derive(Debug, PartialEq, Clone)]
+pub struct PathCompletionItem {
+    pub kind: PathKind,
+    pub item: helix_core::CompletionItem,
+}
+
+#[derive(Debug, PartialEq, Clone)]
+pub enum CompletionItem {
+    Lsp(LspCompletionItem),
+    Path(PathCompletionItem),
```
pascalkuthe, line 155: I don't think we need a seperate `PathCompletionItem` type. Instead we can just have a `kind: Cow<str>`  on `helix_core::CompletionItem`. At that point we could also just call that variant `Other` (so it's clear there will only ever be an lsp variant and one for everything else going trough the helix-core type).

I am also not really sure about this abstraction in general. I think first starters this should be much better encapsulated. This file is certainly the wrong place for it (I don't want the handler/resolve code to depend on the UI at all). So probably at least a submodule in `handlers::completion`. I would also like to have as much information extraction as possible within methods (like `preselect`). It's ok to keep `lsp_item_to_transaction in the UI code for now since that is a larger change (altough it also doesn't belong there)

book/src/editor.md
```diff
@@ -32,6 +32,7 @@
 | `cursorcolumn` | Highlight all columns with a cursor | `false` |
 | `gutters` | Gutters to display: Available are `diagnostics` and `diff` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "spacer", "line-numbers", "spacer", "diff"]` |
 | `auto-completion` | Enable automatic pop up of auto-completion | `true` |
+| `path-completion` | Enable filepath completion. Show files and directories if an existing path at the cursor was recognized, either absolute or relative to the current opened document or current working directory (if the buffer is not yet saved). Defaults to true. | `true` |
```
Philipp-M, line 35: Maybe this makes sense to add, to avoid questions, why it's not working in windows?
```suggestion
| `path-completion` | Enable filepath completion. Show files and directories if an existing path at the cursor was recognized, either absolute or relative to the current opened document or current working directory (if the buffer is not yet saved). This is currently only supported in Linux and OSX. Defaults to true. | `true` |
```

Probably makes sense as well to open an issue, or update #2874

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +302,200 @@ fn request_completion(
...
+            .byte_slice((start_byte + m.start())..(start_byte + m.end()))
+            .to_string();
+
+        // resolve home dir (~/, $HOME/, ${HOME}/) on unix
+        #[cfg(unix)]
+        let mut path = {
+            static HOME_DIR: Lazy<Option<OsString>> = Lazy::new(|| std::env::var_os("HOME"));
+
+            let home_resolved_path = if let Some(home) = &*HOME_DIR {
+                let first_separator_after_home = if matched_path.starts_with("~/") {
```
pascalkuthe, line 335: I think for this feature in particular (but also gf) proper path expansion would be nice instead of this special case.

pascalkuthe (changes requested):

book/src/editor.md
pascalkuthe: I think the code does work on windows (or at least after you use the regex I suggested below)

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+            let mut path = {
+                static HOME_DIR: Lazy<Option<std::ffi::OsString>> =
+                    Lazy::new(|| std::env::var_os("HOME"));
+
+                let home_resolved_path = if let Some(home) = &*HOME_DIR {
+                    let first_separator_after_home = if matched_path.starts_with("~/") {
+                        Some(1)
+                    } else if matched_path.starts_with("$HOME") {
+                        Some(5)
+                    } else if matched_path.starts_with("${HOME}") {
```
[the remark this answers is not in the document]
Philipp-M: You mean implement our own (more general) kind of shell expansion in `helix-stdx`? What kind of dependency are you talking about?

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+fn path_completion(
+    cursor: usize,
+    text: helix_core::Rope,
+    doc: &helix_view::Document,
+) -> Option<BoxFuture<'static, anyhow::Result<Vec<CompletionItem>>>> {
+    if !doc.supports_path_completion() {
+        return None;
+    }
+
+    // TODO find a good regex for most use cases (especially Windows, which is not yet covered...)
```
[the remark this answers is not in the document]
Philipp-M: Nice one, I'll take a look at it.

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+                    path = work_dir.join(path);
+                }
+            }
+            let ends_with_slash = match matched_path.chars().last() {
+                Some(std::path::MAIN_SEPARATOR) => true,
+                None => return None,
+                _ => false,
+            };
+            // check if there are chars after the last slash, and if these chars represent a directory
+            match std::fs::metadata(path.clone()).ok() {
```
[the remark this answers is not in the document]
Philipp-M: It could well be that this is an artifact, after a [recent change](https://github.com/helix-editor/helix/pull/2608#discussion_r1740152557), I've overlooked, I'll take a look into it.

helix-term/src/ui/completion.rs
Philipp-M: See https://github.com/helix-editor/helix/pull/2608#discussion_r1740929468

helix-term/src/ui/completion.rs
Philipp-M: > I don't think we need a seperate `PathCompletionItem` type. Instead we can just have a kind: `Cow<str>` on `helix_core::CompletionItem`. At that point we could also just call that variant `Other` (so it's clear there will only ever be an lsp variant and one for everything else going trough the helix-core type).

I just fear that there may be more information relevant for a completion item, but that fear may well be unjustified.

> This file is certainly the wrong place for it

See [recent discussion](https://github.com/helix-editor/helix/pull/2608#discussion_r1720817397) for it, I'm aware of it, but I don't want to start an even bigger refactor, when PRs are open for so long without any kind of review to avoid resolving merge-conflicts. It shouldn't be too hard to move this somewhere else, when that kind of refactor is done.

> At that point we could also just call that variant `Other`.

helix-term/src/handlers/completion.rs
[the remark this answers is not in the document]
pascalkuthe: I meant https://docs.rs/shellexpand/latest/shellexpand/

A lot of complexity in this crate comes from the fact that in the past converting raw OsString <-> bytes was ugly but these days it's almost trivial (since we can use a regex).

Basically just iterate all matches for a regex `\$\{[\w@/\\.\-+#$%~=:]+\}||\$\w+` and try to replace them with the environment variable. I had a function written up. If you feel uncomfortable with that I can also write it

helix-term/src/ui/completion.rs
pascalkuthe: > I just fear that there may be more information relevant for a completion item, but that fear may well be unjustified.

To me if there is more relevant information is should be added to the generic `helix_core::CompletionItem`. I don't want to couple the UI code to the completion sources. Basically `helix_core::CompletionItem` shouldn't be modeled after whatever data completion sources can or can't provide but instead by the data that the UI can actually support (and I would want everything to go trough that generic interface except LSP). The only reason that lsp needs to be separate is because creating the transaction is complicated and needs to be done lazily.

pascalkuthe: > See https://github.com/helix-editor/helix/pull/2608#discussion_r1720817397 for it, I'm aware of it, but I don't want to start an even bigger refactor, when PRs are open for so long without any kind of review to avoid resolving merge-conflicts. It shouldn't be too hard to move this somewhere else, when that kind of refactor is done.

I agree that we shouldn't do a large refactor in this PR but moving this into another file inside helix term (like in the handler) is a pretty easy cosmetic change and reduces spaghetti code by a lot. I don't think that would even conflict with the snippet system PR

helix-term/src/handlers/completion.rs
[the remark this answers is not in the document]
Philipp-M: I mean I could certainly dive into this, but it seems you already have a pretty good idea how to do all of that, so you're likely faster (including much more research regarding the path recognition regex).

Maybe it makes sense for you to create a PR with `gf` + path regex + shell expansion, and I base this path-completion PR on top of that?

I can help reviewing, if that helps getting this merged faster.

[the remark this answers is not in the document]
pascalkuthe: I would prefer doing a PR against your branch/fork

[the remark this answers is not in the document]
Philipp-M: That's good for me as well.

pascalkuthe:
pushed to this branch by accident, sorry about that I reverted the change (not clear in github UI)

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+        .and_then(|matched_path| {
+            let end = text.byte_to_char(matched_path.end());
+            if end != cursor {
+                return None;
+            }
+            let start = text.byte_to_char(matched_path.start());
+            let matched_path = &text.slice(start..end).to_string();
+
+            // resolve home dir (~/, $HOME/, ${HOME}/) on unix
+            #[cfg(unix)]
```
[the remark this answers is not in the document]
Philipp-M: Should be handled by https://github.com/Philipp-M/helix/pull/4

helix-term/src/handlers/completion.rs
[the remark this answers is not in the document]
Philipp-M: Should be handled by https://github.com/Philipp-M/helix/pull/4

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
     });
 }
 
+fn path_completion(
```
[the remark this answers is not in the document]
Philipp-M: I've moved this into `handlers/completion/path.rs`

helix-term/src/handlers/completion.rs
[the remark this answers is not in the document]
Philipp-M: This was solved with https://github.com/Philipp-M/helix/pull/4

helix-term/src/handlers/completion.rs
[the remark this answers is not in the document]
Philipp-M: I removed that, is the current version better?

helix-term/src/handlers/completion.rs
```diff
@@ -291,6 +300,213 @@ fn request_completion(
...
+
+            if path.is_relative() {
+                if let Some(doc_path) = doc.path().and_then(|dp| dp.parent()) {
+                    path = doc_path.join(path);
+                } else if let Ok(work_dir) = std::env::current_dir() {
+                    path = work_dir.join(path);
+                }
+            }
+            let ends_with_slash = match matched_path.chars().last() {
+                Some(std::path::MAIN_SEPARATOR) => true,
```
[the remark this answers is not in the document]
Philipp-M: Should be handled by https://github.com/Philipp-M/helix/pull/4

helix-term/src/handlers/completion.rs
```diff
@@ -251,15 +256,19 @@ fn request_completion(
...
+                        provider: language_server_id,
+                        resolved: false,
+                    })
                 })
                 .collect();
                 anyhow::Ok(items)
             }
+            .boxed()
         })
+        .chain(path_completion(cursor, text.clone(), doc))
```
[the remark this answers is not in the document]
Philipp-M: I've added an `Arc<AtomicBool>` for this which is polled.
Alterntative would probably just be using tokio async functions, but as I said, it was significantly slower (probably around half the speed) than using everything in a blocking call.

helix-term/src/handlers/completion.rs
Philipp-M: Should be handled in the recent push.

helix-term/src/ui/completion.rs
Philipp-M: I think this should be handled.

Philipp-M:
AFAICS path detection + shell-expansion should now be quite a bit more robust with @pascalkuthe s changes, and should also support windows.
I have no (easy) way currently test this though, so real-world testing would be appreciated.

helix-term/src/handlers/completion.rs
```diff
@@ -196,6 +202,7 @@ fn request_completion(
     // necessary from our side too.
     trigger.pos = cursor;
     let trigger_text = text.slice(..cursor);
+    let cancel_completion = Arc::new(AtomicBool::new(false));
```
pascalkuthe, line 205: I don't want to build custom synchronization code here around atomic. What you did is reasonable but I want it to be part of a single cancelation token implementation. I thought our cancelation token was already clonable when I made that comment but seems I misremembered.

I already have a specific implementation in mind (a simplified variant if the tokio-utils cancelation token). I am currently on vacation but I will send a PR for that to your branch once I get back (unless you want to /get around to implementing that first)

helix-term/src/handlers/completion.rs
```diff
@@ -346,7 +368,14 @@ pub fn trigger_auto_completion(
                         ..
                     }) if triggers.iter().any(|trigger| text.ends_with(trigger)))
         });
-    if is_trigger_char {
+
+    let trigger_path_completion = matches!(
+        text.get_bytes_at(text.len_bytes())
+            .and_then(|t| t.reversed().next()),
+        Some(b'/' | b'\\')
```
pascalkuthe, line 375: I think we only want to match the backslash on windows

helix-term/src/handlers/completion/path.rs
```diff
@@ -0,0 +1,194 @@
...
+            } else {
+                Path::new(&*matched_path).into()
+            };
+            let path = path::expand(&path);
+            let parent_dir = doc.path().and_then(|dp| dp.parent());
+            let path = match parent_dir {
+                Some(parent_dir) if path.is_relative() => parent_dir.join(&path),
+                _ => path.into_owned(),
+            };
+            let ends_with_slash = matches!(matched_path.as_bytes().last(), Some(b'/' | b'\\'));
```
pascalkuthe, line 52: This may have been my mistake but I think we want to only set ends with slash to true for backslash if we are on windows (so for windows we want to check for both forward and reverse slash and for unix we only want to check for forward)

helix-term/src/handlers/completion/path.rs
```diff
@@ -0,0 +1,194 @@
...
+                    .metadata()
+                    .ok()
+                    .map(|md| (dir_entry.file_name(), md))
+            })
+            .map_while(|(file_name, md)| {
+                if cancel.load(Ordering::Relaxed) {
+                    return None;
+                }
+
+                let file_name_str = file_name.to_string_lossy().to_string();
```
pascalkuthe, line 91: This allocated, I think we can't properly compelte non-utf8 filename so let's just skip them (so us into_string() and ignore cases that return none)

pascalkuthe (commented):
This is starting to look pretty good, thanks for addressing my comments! I found a couple smaller things but hopefully should be ready soon

helix-term/src/handlers/completion.rs
RGBCube: I don't think this should be platform specific. One could be writing code for mainly-Windows software (like PowerShell) while being on a Linux machine.

Philipp-M: > I think we only want to match the backslash on windows

Yeah I thought about doing this, but seemingly forgot about it not long after...

> One could be writing code for mainly-Windows software (like PowerShell) while being on a Linux machine

But on a linux machine there's no `\` as path separator is there? So it would not make sense to have this on Linux (as this path needs to exist and be readable for path-completion to work)?

RGBCube: Treating `/` as `\` while serving completions on Linux is fine as Windows does the reverse, there's no need to complicate behavior for no benefit.

I'm sure someone out there is editing Windows specific source code on Linux, where they treat `foo/bar` as `foo\bar`

pascalkuthe: Windows path syntax and Linux path syntax are not generally equivalent. The path detection is also OS dependent, backslashes act as escape chars (like with spaces) on unix but oath separators on windows.

Windows enables forward slashes as path separators in their paths so we support that but on unix you can absolutely not treat backslash as separators.

Some Windows paths also just don't make sense on unix. If you want crossplatform path completions always use forward slash since they work both on windows and unix (including powershell)

Philipp-M: Should be implemented now.

helix-term/src/handlers/completion.rs
Philipp-M: I can wait for you PR, thanks.

helix-term/src/handlers/completion.rs
RGBCube: Fair, I forgot that stuff like `cat foo\ bar` was valid posix shell for `cat "foo bar"`. I didn't know that `/` was that portable.

helix-term/src/handlers/completion.rs
pascalkuthe: I put up a PR, ended up a bit more complex than I thaught but I am very happy with it. After looking at it I also did a lot of "quick hacks" like this that weren't really wrong but are better centralized in dedicated shared code

Philipp-M: Yeah, probably better to have this a little bit more encapsulated.

David-Else:
When using the `%` macro with this branch the home directory is not included: https://github.com/helix-editor/helix/discussions/12002.

Check this out with a file in your Linux home directory:

```toml
[keys.normal.space]
o = "@:o <C-r>%<C-w>"
```

Philipp-M:
I'm not sure what the expected outcome is, or whether this is a bug at all?

I've compared the `path-completion` branch with current `master` and it seems to have the same behavior for me.
I don't think there should any difference to `master` for this macro reasoning from the code as the new shell-expansion doesn't seem to be used in the `open` impl.

Nice macro btw.

David-Else:
@Philipp-M I don't think it is a problem or a bug, chatting about it in the linked discussion.

the-mikedavis (approved):
First thanks for working on this @Philipp-M and especially for keeping the branch alive for so long[^1]! 😀

I've been running this locally for a few weeks now and I'm pretty sure it's good to go. I'd like to merge it down and have it sit in master for a while to eek out any bugs we might've introduced in the completion changes. I might've felt a change in the ranges we replace with a completion, but this might be a rust-analyzer peculiarity or me imagining things so I don't want it to block the merge. I'd like to see more eyes from anyone running master before a possible December release.

Using this in some Nix files and, nicely, `Cargo.toml`s has been quite nice. Thanks again @Philipp-M! (And completion and regex improvements from @pascalkuthe ofc :)

This is one small step for non-LSP completions and one giant leap for moving the codebase away from its very close relationship with LSP :rocket:

[^1]: Some of the other PRs merged recently have been 10,000 + this PR's number 😅

RoloEdits:
@the-mikedavis I have also been experiencing some issues with `rust-analyzer`, without this branch's changes, which I expanded on more in #11933. So shouldn't have anything to do with this pr.

RoloEdits:
I did notice that the msrv was broken though. Some of the `OsStr` stuff needs rust 1.74.0

```shell
warning: current MSRV (Minimum Supported Rust Version) is `1.70.0` but this item is stable since `1.74.0`
  --> helix-stdx\src\env.rs:98:21
   |
98 |     let bytes = src.as_encoded_bytes();
   |                     ^^^^^^^^^^^^^^^^^^
   |
   = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
   = note: `#[warn(clippy::incompatible_msrv)]` on by default

warning: current MSRV (Minimum Supported Rust Version) is `1.70.0` but this item is stable since `1.74.0`
   --> helix-stdx\src\env.rs:117:28
    |
117 |         let var = unsafe { OsStr::from_encoded_bytes_unchecked(var) };
    |                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv

warning: current MSRV (Minimum Supported Rust Version) is `1.70.0` but this item is stable since `1.74.0`
   --> helix-stdx\src\env.rs:124:25
    |
124 |                     val.as_encoded_bytes()
    |                         ^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv

warning: current MSRV (Minimum Supported Rust Version) is `1.70.0` but this item is stable since `1.74.0`
   --> helix-stdx\src\env.rs:138:18
    |
138 |         unsafe { OsString::from_encoded_bytes_unchecked(res) }.into()
    |                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#incompatible_msrv
```

poliorcetics:
IIRC Helix uses Firefox's MRSV ? It's currently 1.76.0 (https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html) so we could just increase it

Issue #708, opened 2021-09-04
Customizable statusline


kirawi:
Following some discussion on #649, it might be a good idea to allow for a configurable statusline through templating. We can borrow some code from [here](https://github.com/mitsuhiko/indicatif/blob/27c8cc3078b65aa2573dbe673a00471d6662698c/src/style.rs#L238-L390), and make some of the changes suggested by Blaz: 'We might be able to reuse that with some tweaking (lazy_static -> once_cell, disabling styles for now).'

Omnikar:
What should be used in the template string to indicate the separation between the left and right sides of the statusline? Maybe `{|}`?

kirawi:
Maybe we could use the alignment operators: `<`, `>`? It would just be relative to the left and right sides of the status line.

Omnikar:
If we just use a "separator", such as `{|}`, like I mentioned, we can just allow `left{|}center{|}right`.

kirawi:
Sounds good.

Omnikar:
Though we'd have to decide what to do if there are less than 2 separators present in the format string. I think:
- 0: `left`
- 1: `left{|}right`
If you wanted only left and center or right and center, you could do `left{|}center{|}` or `{|}center{|}right`. And center-only can be achieved via `{|}center{|}`.

gbaranski:
Or we could do
```toml
[statusline]
left = "editing {filename}"
center = "Date: {date}"
right = "LSP: {lsp-status}"
```

What do you think about that?

kirawi:
Conditional templating would also be interesting, e.g. #447

gbaranski:
I3status-rust does something similar to what we want. https://github.com/greshake/i3status-rust/blob/master/doc/blocks.md

sudormrfbin:
> Maybe something like the C++ short hand if else statement (`(condition) ? expressionTrue : expressionFalse`) with some predefined conditions?

I think we want to avoid introducing constructs like that since it will probably spiral into injecting more and more logic handling into a plain toml file and before we know it we would have invented a mini language that's similar to most things out there but also very different from most things out there. It was also brought up in Matrix a while ago and there was a (more or less) general consensus that it's better left to when we have a proper plugin system in place which can be used to dynamically define the config file or an actual configuration _programming_ language (lisps were of interest last time IIRC).

bestouff:
I'd like to have this kind of eye candy in the status bar possible:

[...]

PR #2434, opened 2022-05-08
Customizable/configurable status line

etienne-k:
### Summary

Building upon the feedback of #2420 and the discussion in #708, I refactored the status line implementation and made it configurable.

The configuration happens in the user's configuration file and distinguishes between three areas of the status line: **left**, **center** and **right**.

```
[ ... ... LEFT ... ... | ... ... ... ... CENTER ... ... ... ... | ... ... RIGHT ... ... ]
```

Within each area, status line elements can be defined. There are no particular restrictions regarding the order, and elements can occur multiple times or be removed completely.

The following elements can currently be configured (see `helix_view::editor::StatusLineElement`):

```
mode
spinner
file-name
file-encoding
file-type
diagnostics
selections
position
```

### Example

This would move the file name to the center, and add a file type indicated to the far right):

```toml
[editor.status-line]
left = ["mode", "spinner"]
center = ["file-name"]
right = ["diagnostics", "selections", "position", "file-encoding", "file-type"]
```

image

### Additional info

The default configuration keeps the same status line layout with the same elements as before. While I do see room for improvement here, I assume it'd better be done in another PR.

On top of the elements which were already present, I added the implementation for one additional element: the file type (disabled by default).

PR #2610, opened 2022-05-30
helix-term: ui: Refactor statusline rendering, set configurable

litchipi:
In an effort to help bring custom statusline (related issue #708), I created a first refactor that changes almost nothing in the render but would allow future configuration of the statusline:
- Deleted `render_statusline` function, created a `StatuslineRender` struct instead.
- Created `StatuslineElement` enum to store every element that can be rendered, with configurable data associated.
Every new "feature" to add to the statusline will just be a new element in this enum.
- The render of the statusline is now a rendering of the left part, right part and middle part of the statusline (in this order)
- A "TODO" has been added where the `String` (user config) to `StatuslineRender` will take place
Still have to decide the format and implement it.
I was thinking of something like:
`{mode} blah blah {filename} bli blouh {diagnostic W E}`
"blah blah" being static strings added to the statusline, (spaces, dashes, dots, etc ...)
"W" and "E" being configurable data associated to the statusline element

Changes in the rendered UI:
- The filename is now rendered in the middle of the available space between the left and right text
- No leading space in the left text of the bar

Questions I have regarding the implementation:
- Are the `struct` well placed in the code ?
- Should I add more comments ?
- Are there more arguments that should be passed to the different `render` functions in order to implement any further features on the bar ?
- How the `EditorView` struct can access the user configuration in order to load the `StatuslineRender` struct from the placeholder string ?

First contribution here, please tell me if there's something wrong :-)

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -29,13 +29,14 @@ use std::borrow::Cow;
 
 use crossterm::event::{Event, MouseButton, MouseEvent, MouseEventKind};
 use tui::buffer::Buffer as Surface;
+use tui::text::{Span, Spans};
 
 pub struct EditorView {
     pub keymaps: Keymaps,
     on_next_key: Option<Box<dyn FnOnce(&mut commands::Context, KeyEvent)>>,
     last_insert: (commands::MappableCommand, Vec<InsertEvent>),
     pub(crate) completion: Option<Completion>,
-    spinners: ProgressSpinners,
+    statusline: StatusLineRender,
 }
 
 #[derive(Debug, Clone)]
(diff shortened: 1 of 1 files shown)


litchipi:
My bad, entered directly into the meat of the code and didn't check all the issue before, missed #2434 PR.

Maybe it can bring a thing or 2 to the other PR (I do think the `StatuslineElement` enum is a good idea) so I let it open, but feel free to close this PR when you want, I'll be more precautious next time :sweat_smile:

helix-term/src/ui/editor.rs
```diff
@@ -651,151 +653,6 @@ impl EditorView {
...
-            &title,
-            viewport
-                .width
-                .saturating_sub(6)
-                .saturating_sub(right_side_text.width() as u16 + 1) as usize, // "+ 1": a space between the title and the selection info
-            |_| base_style,
-            true,
-            true,
-        );
-    }
```
pickfire, line 797: I am worried if we do a custom statusline we lost the ability to do truncation.

helix-term/src/ui/editor.rs
```diff
@@ -1404,6 +1261,230 @@ impl Component for EditorView {
...
+            viewport.y,
+            &right_text,
+            right_text.width() as u16,
+        );
+
+        let mut middle_text = Spans::default();
+        for element in self.statusline_middle.iter() {
+            element.render(
+                &mut middle_text,
+                &self,
```
pickfire, line 1345: I think you might have to fix the lints.

archseer:
#2434 was merged, but thanks for looking into this!

PR #2611, opened 2022-05-30
Appen\Paste-After\Shell-Pile-After EOL special case

andreytkachenko:
In normal mode when cursor on the new line symbol (\n) actions like append or paste after inserts data after new line to the next line. But it is annoying, you have to remember to get cursor back by 1 char and then paste. The PR fixes that behavior by checking if there is '\n' on the end of the range, if it is just subtract 1.

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -83,6 +83,18 @@ impl Range {
         std::cmp::max(self.anchor, self.head)
     }
 
+    /// End of the range taking into account end of line.
+    #[inline]
+    #[must_use]
+    pub fn to_without_eol(&self, text: &RopeSlice) -> usize {
+        let to = self.to();
+        if text.char(to - 1) == '\n' {
+            to - 1
+        } else {
+            to
+        }
+    }
+
     /// Total length of the range.
     #[inline]
     #[must_use]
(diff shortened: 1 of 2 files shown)


the-mikedavis (changes requested):
The current behavior is deliberate. Vim treats newlines with special handling like this but kakoune and helix do not - newlines are treated like any other character.

andreytkachenko:
I understand that this behavior is deliberate, I am not suggesting completely exclude newline symbol from the line, I am just suggesting to add some special cases for paste, append and shell-pipe append.

andreytkachenko:
@archseer @sudormrfbin @pickfire you opinions guys? Should we just close PR in favor `newline = "all"` config option as solution?

pickfire:
I don't have opinion on this but @cessen would probably have opinion, but I think the current approach should be correct.

sudormrfbin:
I agree with @the-mikedavis here; since we handle newline like any other character the current behavior makes sense. Also setting `editor.whitespace.render.newline = "all"` should help with any confusion regarding whether the cursor is on a newline. On a tangent, you don't have to move the cursor back one character either if you want to paste on the current line (when cursor is on newline), pressing `P` instead of `p` should do the trick.

pickfire:
> pressing P instead of p should do the trick

The issue with that is that both methods requires you to delete an empty newline. It's either one of

- `Pjd`
- `pggd`

PR #2619, opened 2022-05-31
Elixir: inject Markdown into docs, remove h sigil HEEx injection

connorlay:
@the-mikedavis I figured I'd checkout Helix and see how it compares to Neovim. Consider me a convert 😆  

This PR contains two changes:
- injects Markdown, the [official documentation format](https://hexdocs.pm/elixir/writing-documentation.html#markdown), into Elixir
- Removes the non-existent `h` sigil from Elixir language injections. Phoenix Live View only supports the uppercase `H` sigil for HEEx.

Please let me know if there are any tests or documentation that need to be updated. Thanks!

### DIFF
--- a/runtime/queries/elixir/injections.scm
+++ b/runtime/queries/elixir/injections.scm
@@ -11,6 +11,15 @@
 ((sigil
   (sigil_name) @_sigil_name
   (quoted_content) @injection.content)
- (#match? @_sigil_name "^(h|H)$")
+ (#eq? @_sigil_name "H")
  (#set! injection.language "heex")
  (#set! injection.combined))
+
+(unary_operator
+  operator: "@"
+  operand: (call
+  target: ((identifier) @_identifier (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
+    (arguments [
+      (string (quoted_content) @injection.content)
+      (sigil (quoted_content) @injection.content)
+  ])) (#set! injection.language "markdown"))
--- a/runtime/queries/elixir/injections.scm
+++ b/runtime/queries/elixir/injections.scm
@@ -18,7 +18,7 @@
 (unary_operator
   operator: "@"
   operand: (call
-  target: ((identifier) @_identifier (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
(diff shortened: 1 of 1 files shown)


runtime/queries/elixir/injections.scm
```diff
@@ -11,6 +11,15 @@
...
   (quoted_content) @injection.content)
- (#match? @_sigil_name "^(h|H)$")
+ (#eq? @_sigil_name "H")
  (#set! injection.language "heex")
  (#set! injection.combined))
+
+(unary_operator
+  operator: "@"
+  operand: (call
+  target: ((identifier) @_identifier (#any-of? @_identifier "moduledoc" "typedoc" "shortdoc" "doc"))
```
the-mikedavis, line 21:
```suggestion
  target: ((identifier) @_identifier (#match? @_identifier "^(module|type|short)?doc$"))
```

Helix doesn't implement `#any-of?` yet, we've been using `#match?` to get around it

(I think we've discussed it before and the only tricky part of implementing it is that it would most likely be slower than an equivalent `#match?` regex like above to loop through the options. So we might be able to compile it down to a regex for the best of both worlds.)

the-mikedavis (commented):
Ah TIL HEEx is limited to `~H`, that makes sense 😄

Glad we could steal you from nvim - welcome! 🥳

runtime/queries/elixir/injections.scm
connorlay: Good to know! I applied the suggestion above; is there anything else you'd like to see in this PR?

the-mikedavis: Nope this looks great! The markdown highlights look really nice, especially the injection in `## Examples` blocks 🤩

the-mikedavis (approved):

the-mikedavis:
W.r.t. tests, I think it might be good in the long-run to have highlight tests in the CI and/or cargo testsuite. For now I think it would be a little heavy since we're adding new languages very often and tree-sitter queries are already a lot to learn if you've never seen tree-sitter before

archseer:
> Consider me a convert

I see our Helix evangelism strikeforce is just as strong as the Rust one :D

PR #2620, opened 2022-05-31
feat: make `move_vertically` aware of tabs and wide characters

mtoohey31:
As the `TODO` in `move_vertically` previously noted, the current behaviour leads to jerky vertical movement on lines with wide characters and tabs. This pull request refactors a few of the related functions (and resolves some of their `TODO`s in the process) so that vertical movement is always smooth.

Since the parameters to `pos_at_coords` had to be modified to accept `tab_width`, quite a few tests had to be updated. Some of these tests also needed their expected values changed since `pos_at_coords` now operates based on visual row/column instead of graphemes, as the `TODO` called for. I'm not 100% sure if I updated a few of the tests correctly though, in particular, [this assertion](https://github.com/mtoohey31/helix/blob/966a17f29fbb5d975c9c3752605d1932273f4e06/helix-core/src/position.rs#L280), and [this block](https://github.com/mtoohey31/helix/blob/966a17f29fbb5d975c9c3752605d1932273f4e06/helix-core/src/position.rs#L300-L304). If someone could offer a second opinion on those, I'd appreciate it!

### DIFF
--- a/helix-core/src/position.rs
+++ b/helix-core/src/position.rs
@@ -109,10 +109,12 @@ pub fn visual_coords_at_pos(text: RopeSlice, pos: usize, tab_width: usize) -> Po
 /// with left-side block-cursor positions, as this prevents the the block cursor
 /// from jumping to the next line.  Otherwise you typically want it to be `false`,
 /// such as when dealing with raw anchor/head positions.
-///
-/// TODO: this should be changed to work in terms of visual row/column, not
-/// graphemes.
-pub fn pos_at_coords(text: RopeSlice, coords: Position, limit_before_line_ending: bool) -> usize {
+pub fn pos_at_coords(
+    text: RopeSlice,
+    coords: Position,
+    tab_width: usize,
+    limit_before_line_ending: bool,
+) -> usize {
     let Position { mut row, col } = coords;
     if limit_before_line_ending {
         row = row.min(text.len_lines() - 1);
@@ -125,11 +127,20 @@ pub fn pos_at_coords(text: RopeSlice, coords: Position, limit_before_line_ending
     };
 
     let mut col_char_offset = 0;
-    for (i, g) in RopeGraphemes::new(text.slice(line_start..line_end)).enumerate() {
-        if i == col {
+    let mut cols_remaining = col;
+    for grapheme in RopeGraphemes::new(text.slice(line_start..line_end)) {
+        let grapheme_width = if grapheme == "\t" {
+            tab_width
+        } else {
+            let grapheme = Cow::from(grapheme);
+            grapheme_width(&grapheme)
+        };
+        if let Some(new_cols_remaining) = cols_remaining.checked_sub(grapheme_width) {
+            cols_remaining = new_cols_remaining;
+        } else {
             break;
         }
-        col_char_offset += g.chars().count();
+        col_char_offset += grapheme.chars().count();
     }
 
     line_start + col_char_offset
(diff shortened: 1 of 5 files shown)


helix-core/src/position.rs
```diff
@@ -109,10 +109,12 @@ pub fn visual_coords_at_pos(text: RopeSlice, pos: usize, tab_width: usize) -> Po
 /// with left-side block-cursor positions, as this prevents the the block cursor
 /// from jumping to the next line.  Otherwise you typically want it to be `false`,
 /// such as when dealing with raw anchor/head positions.
-///
```
archseer, line 112: This TODO seems wrong, pos_at_coords is intended for line, col calculation, and we already have pos_at_visual_coords for functions that need to deal with tab width and grapheme widths. The method should be kept unchanged, instead commands should use visual_ versions where necessary.

mtoohey31: Ok, I've refactored things in 33e8fa765b003016aa2129166b8d72f4431140ac so `pos_at_visual_coords` is a separate function. I removed the `limit_before_line_ending` parameter because I figure when we're dealing with visual coords we'll always want the behaviour corresponding to a true value for that parameter, right?

Also, I noticed [this comment](https://github.com/helix-editor/helix/blob/eba82250bb4403fcb2e3ade74ba7301a680bc561/helix-view/src/view.rs#L267) that mentions `pos_at_visual_coords`; is it fine if I update the code there to use that function now that it exists? It looks like the code there does the same thing as my implementation (at least, it does now, I found a bug in my code with the handling of tabs that don't start at a multiple of `tab_width` from looking at it).

archseer: Yeah good catch! It should use that function, similar to how the inverse does: https://github.com/helix-editor/helix/blob/eba82250bb4403fcb2e3ade74ba7301a680bc561/helix-view/src/view.rs#L226-L232=

helix-core/src/position.rs
```diff
@@ -135,6 +132,41 @@ pub fn pos_at_coords(text: RopeSlice, coords: Position, limit_before_line_ending
...
+        let grapheme_width = if grapheme == "\t" {
+            tab_width - ((col - cols_remaining) % tab_width)
+        } else {
+            let grapheme = Cow::from(grapheme);
+            grapheme_width(&grapheme)
+        };
+
+        if grapheme_width > cols_remaining {
+            break;
+        }
```
archseer, line 161: Can you include the original comment here: https://github.com/helix-editor/helix/blob/eba82250bb4403fcb2e3ade74ba7301a680bc561/helix-view/src/view.rs#L280-L281=

helix-core/src/position.rs
```diff
@@ -135,6 +132,41 @@ pub fn pos_at_coords(text: RopeSlice, coords: Position, limit_before_line_ending
...
+    let Position { mut row, col } = coords;
+    row = row.min(text.len_lines() - 1);
+    let line_start = text.line_to_char(row);
+    let line_end = line_end_char_index(&text, row);
+
+    let mut col_char_offset = 0;
+    let mut cols_remaining = col;
+    for grapheme in RopeGraphemes::new(text.slice(line_start..line_end)) {
+        let grapheme_width = if grapheme == "\t" {
+            tab_width - ((col - cols_remaining) % tab_width)
```
archseer, line 153: The logic here and in the conditional looks slightly different though

mtoohey31: Ah true, it is still slightly different. The other implementation keeps track of the current column, while mine keeps track of the columns remaining before we reach the target column, but I think the behaviour is still the same.

helix-core/src/position.rs
mtoohey31: Thanks, updated in 324a76b7873ed64a3723c4f00eb786e4c488d502.

mtoohey31:
I just realized that `copy_selection_on_line` also doesn't respect wide characters, so 274746e29bd917c50b7a02e85b834fe7b1a90a92 updates things to fix that too.

From a quick look I didn't spot any, but do you know of other commands that should be updated?

archseer (approved):
Sorry for the delay, this looks correct to me and I couldn't find any more spots that need fixing. Great work! 🎉

PR #2625, opened 2022-05-31
Fix unwrap error when undo after `shell_append_output`

andreytkachenko:
Steps to reproduce:
1. press  `Shift+|` 
2. enter `ls` in prompt
3. press `u` for undo 
4. ether nothing happened or helix closed with error


archseer (approved):
Good catch 👍🏻

PR #2630, opened 2022-05-31
Update languages.toml

V0idMatr1x:
Added Nim LSP & Grammar Support.


kirawi:
You'll also need to run `cargo xtask docgen`

V0idMatr1x:
> You'll also need to run `cargo xtask docgen`

Will do, working on that right now.

V0idMatr1x:
> You'll also need to run `cargo xtask docgen`

Ok, should be fine now; thank you for letting me know!

PR #2633, opened 2022-05-31
Added Support for Nim LSP and Grammar.

V0idMatr1x:


languages.toml
```diff
@@ -15,7 +15,7 @@ indent = { tab-width = 4, unit = "    " }
 [language.auto-pairs]
 '(' = ')'
 '{' = '}'
-'[' = ']'
+'[' = ']'
```
the-mikedavis, line 18: can you revert this change?

runtime/runtime
```diff
@@ -0,0 +1 @@
+/Users/v0id/helix/runtime
```
the-mikedavis, line 1: and this one

languages.toml
```diff
@@ -1409,3 +1409,16 @@ injection-regex = "verilog"
 [[grammar]]
 name = "verilog"
 source = { git = "https://github.com/andreytkachenko/tree-sitter-verilog", rev = "514d8d70593d29ef3ef667fa6b0e504ae7c977e3" }
+
+[[language]]
+name = "nim"
+scope = "source.nim"
+file-types = ["nim", "nimble"]
```
the-mikedavis, line 1416:
```suggestion
file-types = ["nim", "nims", "nimble"]
```

to add support for nimscript

the-mikedavis (changes requested):

runtime/runtime
V0idMatr1x: How do I remove this?

archseer:
The grammar itself isn't useful without the queries so this PR should wait until those are added.

V0idMatr1x:
> The grammar itself isn't useful without the queries so this PR should wait until those are added.

Understandable, I am working on the Highlight queries at the moment trying to figure that part out.

V0idMatr1x:
Highlight queries seem to be functional now; is it okay to merge?

Screen Shot 2022-05-31 at 8 06 18 PM

Issue #2051, opened 2022-04-09
Allow entering insert mode with sequence of commands


David-Else:
### Summary

```toml
[keys.normal]
D = ["collapse_selection", "select_mode", "goto_line_end", "delete_selection"] # << works
C = ["collapse_selection", "select_mode", "goto_line_end", "change_selection"] # << crashes with:
```

```
thread 'main' panicked at 'not implemented', helix-term/src/ui/editor.rs:1183:34
stack backtrace:
   0:     0x5621345d3b2c - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h3e2b509ce2ce6007
   1:     0x562133f27c6c - core::fmt::write::h753c7571fa063ecb
   2:     0x5621345cbf55 - std::io::Write::write_fmt::h2815c0519c99ba09
   3:     0x5621345d6cd5 - std::panicking::default_hook::{{closure}}::h78d3e6cf97fc623d
   4:     0x5621345d6953 - std::panicking::default_hook::hda898f8d3ad1a5ae
   5:     0x562134497a63 - helix_term::application::Application::run::{{closure}}::{{closure}}::h4b76c2240aae4f67
   6:     0x5621345d7381 - std::panicking::rust_panic_with_hook::h1a5ea2d6c23051aa
   7:     0x5621345d7027 - std::panicking::begin_panic_handler::{{closure}}::h07f549390938b73f
   8:     0x5621345d4054 - std::sys_common::backtrace::__rust_end_short_backtrace::h5ec3758a92cfb00d
   9:     0x5621345d6dad - rust_begin_unwind
  10:     0x562133eb0d01 - core::panicking::panic_fmt::h3a79a6a99affe1d5
  11:     0x562133eb0c4d - core::panicking::panic::h97167cd315d19cd4
  12:     0x56213423746d - <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::handle_event::h00bf68e58e63ad3c
  13:     0x56213424c89f - helix_term::compositor::Compositor::handle_event::h2c87fb0c59cda7a2
  14:     0x562134225456 - helix_term::application::Application::handle_terminal_events::hd5459c104fbdeacf
  15:     0x5621344a4778 - <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll::h3ff36697f3352ace
  16:     0x562134471ace - tokio::park::thread::CachedParkThread::block_on::hc59f23f9dca30781
[...]

PR #2634, opened 2022-06-01
Move mode transition logic to handle_keymap_event()

lucypero:
Moves mode transition logic to `handle_keymap_event()`

Fixes this: 

https://github.com/helix-editor/helix/issues/2051#issue-1198579421

Suggestion here:

https://github.com/helix-editor/helix/issues/2051#issuecomment-1140848285


icanhazcheeze:
What is the particular importance of tracking which command moved us into insert mode?

helix-term/src/ui/editor.rs
```diff
@@ -808,13 +808,35 @@ impl EditorView {
...
+                command.execute(cxt);
+                let (_, doc) = current!(cxt.editor);
+                match (last_mode, doc.mode()) {
+                    (Mode::Normal, Mode::Insert) => {
+                        self.last_insert.0 = command.clone();
+                        self.last_insert.1.clear();
+                    }
+                    (Mode::Insert, Mode::Normal) => self.completion = None,
+                    _ => (),
+                }
```
archseer, line 824: This is a lot of duplication that should be extracted into a separate function

helix-term/src/ui/editor.rs
```diff
@@ -1277,29 +1299,6 @@ impl Component for EditorView {
...
                 }
 
-                // mode transitions
-                match (mode, doc.mode()) {
-                    (Mode::Normal, Mode::Insert) => {
-                        // HAXX: if we just entered insert mode from normal, clear key buf
-                        // and record the command that got us into this mode.
-
-                        // how we entered insert mode is important, and we should track that so
-                        // we can repeat the side effect.
```
archseer, line 1287: Comments here were useful and should not be stripped

helix-term/src/ui/editor.rs
```diff
@@ -1277,29 +1299,6 @@ impl Component for EditorView {
...
-
-                        self.last_insert.0 = match self.keymaps.get(mode, key) {
-                            KeymapResult::Matched(command) => command,
-                            // FIXME: insert mode can only be entered through single KeyCodes
-                            _ => unimplemented!(),
-                        };
-                        self.last_insert.1.clear();
-                    }
-                    (Mode::Insert, Mode::Normal) => {
-                        // if exiting insert mode, remove completion
```
archseer, line 1297: ditto here

archseer (changes requested):

archseer:
> What is the particular importance of tracking which command moved us into insert mode?

This is described in the comment, there's different ways of entering insert mode. For example
`afoo`, `ifoo`, `ofoo`, `Ofoo` all insert in different positions. If we're repeating the edit via `.`, that needs to be replicated.

lucypero:
@archseer Changes made.

It still doesn't seem like the dot `.` operator works as it should. When I press `.` after executing this keymap:

```toml
[keys.normal]
C = ["collapse_selection", "extend_to_line_end", "change_selection"]
```

It doesn't quite do what I expect. it seems like it is only executing the last command, and it inserts what I typed in insert mode. Is this the expected behavior? Isn't it supposed to execute the first two commands too?

the-mikedavis:
That's expected behavior. `extend_to_line_end` is a motion (repeats with `A-.`) rather than a change (repeats with `.`)

lucypero:
@archseer Please review the PR when possible. This bug is still present in Helix and it prevents using certain mappings because it crashes the application.

archseer:
There's over a hundred PRs in the queue so it takes me a while.

The problem is that dot is supposed to only replay the last insert (and any side effect of the insert, for example inserting on the next line), not the whole sequence of movement. How should we handle this? `itest<Esc>` fits, but `itest<Esc>ww` would need to stop at the first time we leave insert mode. But then what about `itest<Esc>wwiasd<Esc>` which is technically two inserts?

The two simple solutions are:
- Make custom sequence commands unrepeatable. This would be consistent but it drops some nice usecases.
- Always repeat the whole sequence. This would produce inconsistent results if the user creates a weird mapping, but at least it would allow repetition.

We could also detect the last insert sequence and only repeat that, but it would be weird to the user if only half of the mapping is repeated.

lucypero:
@archseer 

> Always repeat the whole sequence. This would produce inconsistent results if the user creates a weird mapping, but at least it would allow repetition.

I am totally for this option. I don't see why it shouldn't be like that. It just repeats all the actions. Just like Vim. It would be really useful and intuitive.

If you are not sure about that, we could just go with the first option for now just to fix the crashes:

```
Make custom sequence commands unrepeatable. This would be consistent but it drops some nice usecases.
```

What do you say?

archseer:
> Just like Vim.

Right, but we follow Kakoune where repeat is split into repeating edits, and repeating movement.

Let's always repeat the whole sequence for now, that way we at least resolve the panic for the time being.

helix-term/src/ui/editor.rs
```diff
@@ -806,15 +806,40 @@ impl EditorView {
...
+        let mut last_mode = mode;
         let key_result = self.keymaps.get(mode, event);
         cxt.editor.autoinfo = self.keymaps.sticky().map(|node| node.infobox());
 
+        let mut execute_command = |command: &commands::MappableCommand| {
+            command.execute(cxt);
+            let (_, doc) = current!(cxt.editor);
+            match (last_mode, doc.mode()) {
+                (Mode::Normal, Mode::Insert) => {
+                    // HAXX: if we just entered insert mode from normal, clear key buf
```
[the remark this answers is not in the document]
archseer: It's a convention I've always used because it's easy to grep for

helix-term/src/ui/editor.rs
```diff
@@ -771,15 +771,55 @@ impl EditorView {
...
         let key_result = self.keymaps.get(mode, event);
         cxt.editor.autoinfo = self.keymaps.sticky().map(|node| node.infobox());
 
+        // Track the currently open doc
+        let view = view!(cxt.editor);
+        let doc_id = view.doc;
+
+        let mut execute_command = |command: &commands::MappableCommand| {
+            command.execute(cxt);
+            let doc = cxt.editor.documents.get_mut(&doc_id).unwrap();
```
erasin, line 784: I set remaping for close buffer, It back error.

```
[keys.normal.space]
x = [":buffer-close"]
```

thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-term/src/ui/editor.rs:783:61

the-mikedavis: I added a PR to fix this case: https://github.com/helix-editor/helix/pull/3613

erasin: when close the last buffer , It will take error:


```diff
thread 'main' panicked at 'invalid HopSlotMap key used', helix-view/src/tree.rs:288:20
stack backtrace:
   0: std::panicking::begin_panic
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/std/src/panicking.rs:616:12
   1: <slotmap::hop::HopSlotMap<K,V> as core::ops::index::IndexMut<K>>::index_mut
             at /Users/erasin/.cargo/registry/src/github.com-1ecc6299db9ec823/slotmap-1.0.6/src/hop.rs:948:21
   2: helix_view::tree::Tree::get_mut
-             at ./helix-view/src/tree.rs:288:20
   3: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::handle_event
-             at ./helix-term/src/ui/editor.rs:1268:28
   4: helix_term::compositor::Compositor::handle_event
             at ./helix-term/src/compositor.rs:172:19
   5: helix_term::application::Application::handle_terminal_events
             at ./helix-term/src/application.rs:453:22
   6: helix_term::application::Application::event_loop_until_idle::{{closure}}
             at ./helix-term/src/application.rs:292:21
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/future/mod.rs:91:19
   8: helix_term::application::Application::event_loop::{{closure}}
             at ./helix-term/src/application.rs:268:57
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/future/mod.rs:91:19
  10: helix_term::application::Application::run::{{closure}}
             at ./helix-term/src/application.rs:838:38
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/4b91a6ea7258a947e59c6522cd5898e7c0a6a88f/library/core/src/future/mod.rs:91:19
  12: hx::main_impl::{{closure}}
             at ./helix-term/src/main.rs:143:53
```


https://github.com/helix-editor/helix/blob/1cbf5525540531f26cd6bd765de93b20d96078f5/helix-term/src/ui/editor.rs#L1254

https://github.com/helix-editor/helix/blob/1cbf5525540531f26cd6bd765de93b20d96078f5/helix-term/src/ui/editor.rs#L1268

I don't know how it works, but it's probably caused by this.

the-mikedavis: There's an issue for this now: https://github.com/helix-editor/helix/issues/3626

PR #2637, opened 2022-06-01
feat: added line-wise select mode

lazytanuki:
Hello there !

This PR adds an equivalent to vim's V-LINE mode, here called line-select mode, that I've always been missing in Helix. 

It allows the user to select lines up and down while automatically flipping the selections when the range head crosses the anchor. It also handles count and goto start/end of file.

The commands are usable from any other mode.
line_select_demo
(_the keymap in this GIF demo is old code, don't mind it !_)

Thanks and have a good day !

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -200,6 +200,8 @@ impl MappableCommand {
         extend_char_right, "Extend right",
         extend_line_up, "Extend up",
         extend_line_down, "Extend down",
+        extend_full_line_up, "Extend full line up",
+        extend_full_line_down, "Extend full line down",
         copy_selection_on_next_line, "Copy selection on next line",
         copy_selection_on_prev_line, "Copy selection on previous line",
         move_next_word_start, "Move to beginning of next word",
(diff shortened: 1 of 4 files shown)


pickfire:
Yeah, I find it unnecessary as well, `xxxxx` should be sufficient, but instead of doing that, it may be better to navigate using treesitter (`A-o`) or paragraph (`map`).

Mathspy:
> You can extend lines down and up with `extend_line` (`x`) and `extend_line_above` (no default binding but can be mapped to something convenient like `X`). If you need more flexibility, you can enter select mode (`v`) and use `j`/`k` or the arrow keys to select up and down lines and then ensure the selection is extended to line bounds with `extend_to_line_bounds` (`X`).

Heck, I was wondering why `extend_line_above` wasn't a thing when I tried it, thanks for pointing it out! I also think it makes sense to have a default for it

pickfire:
> Heck, I was wondering why extend_line_above wasn't a thing when I tried it, thanks for pointing it out! I also think it makes sense to have a default for it

For that, you can do `vkkkkkkkX` as an alternative. I don't know what key is good to be set as default.

lazytanuki:
Alright I understand.

musjj:
Can this PR potentially be re-considered? I believe that a true line-wise mode would be incredibly useful for composability. The only way to have line-wise select keybinds that is both *symmetrical* and *composable* is to use `vkkkX` and `vjjjX`.

Any other solution requires you to have *different* keybindings for selecting lines above and below. I think it's rather awkward & inelegant for a very primitive movement.

objectiveryan:
These bindings create a poor-man's line-wise mode that you enter with `X`:
```
"C-x" = "extend_to_line_bounds"
X = { j = "@vj<esc><C-x>X", k = "@vk<esc><C-x>X" }
```
The trick is to make each command end with `X` so it puts you back in the "mode".
(This requires #4709)

PR #2640, opened 2022-06-01
add tree-sitter-edoc

the-mikedavis:
Syntax highlighting for the Erlang doc generator format: https://www.erlang.org/doc/apps/edoc/chapter.html. The XHTML part was actually quite easy to parse: it just looks for `<` and `>` and injects it as HTML.

Also included some fixes for markdown and tsq highlights.

Example:

edoc-highlight


pickfire (approved):
Looks good to me.

PR #2644, opened 2022-06-01
catpuccin ui.popup should be a different color from ui.background

papertigers:
Looks like black_1 was defined but never used.  This was likely meant to be ui.popup all along?


the-mikedavis (approved):
This looks better, without it it's hard to tell apart the popup and the background 👍

Issue #2249, opened 2022-04-24
Recursive macro stuck in loop / stack overflow


Zoybean:
### Summary

Being used to vim keybindings, I accidentally pressed `q` to end a recording instead of `Q`. Doing this a second time after finishing recording the first macro resulted in the program freezing, presumably replaying the now-recursive macro.

The first time this occurred, the program remained frozen for several minutes until I killed it with `kill <PID>`. Attempting to reproduce the issue, I made a smaller macro, which ended up immediately overflowing the stack.

### Reproduction Steps

I tried this:

1. `hx`
2. `QqQq`

I expected this to happen:

- Helix prevents me from recording a recursive macro

Instead, this happened:

- Helix crashes due to stack overflow

### Platform

Linux (Ubuntu on WSL)

### Terminal Emulator

Windows Terminal Version: 1.12.10982.0

### Helix Version

22.03-121-g6de2e763

archseer:
Closed by https://github.com/helix-editor/helix/pull/2647

PR #2647, opened 2022-06-02
Fix crash due to cycles when replaying macros

gavynriebau:
In certain circumstances it was possible to get into an infinite loop
when replaying macros such as when different macros attempt to replay
each other.

This commit adds changes to track which macros are currently being
replayed and prevent getting into infinite loops.

Fixes issue #2249


archseer (approved):
Thanks, implementation seems reasonable!

Issue #2648, opened 2022-06-02
Syntax highlighting bug in JSDoc comments


aral:
### Summary

The syntax highlighting in JSDoc comments messes up after an `@param` declaration.

![image](https://user-images.githubusercontent.com/41798/171630614-b57f05cb-a4a9-44ff-aad7-93194d9a15dd.png)

### Reproduction Steps

n/a (just create a JSDoc comment).

### Helix log

n/a

### Platform

Linux (Fedora Silverblue 36)

### Terminal Emulator

GNOME Console

### Helix Version

helix 22.05 (fc8c4883)

the-mikedavis:
Actually I'm not sure where jsdoc highlights are currently coming from in general - we might be falling back to tree-sitter-comment. We should be able to pull in https://github.com/tree-sitter/tree-sitter-jsdoc to improve them. We already have the injection for it: https://github.com/helix-editor/helix/blob/4f3d0a77064aac18cc4386df5c43c312c3b03f69/runtime/queries/javascript/injections.scm#L25-L28

sudormrfbin:
Closed by #2650.

PR #2650, opened 2022-06-02
add tree-sitter-jsdoc

the-mikedavis:
connects #2648

This is a pretty small grammar. The injection for comments is already set up in `runtime/queries/javascript/injections.scm`. Github also uses this for their syntax highlighting, like in:

```js
/**
 * Extract the string matched by the regex and return both the resulting string
 * and the matched string
 *
 * @param {string} source
 * @param {RegExp} regExp
 * @returns
 */
export function extract(source, regExp) {
  const result = regExp.exec(source)
  return result ? {
    normalizedSource: source.replace(result[0], ''),
    extracted: result[1]
  } : { normalisedSource: source, extracted: '' }
}
```

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -41,6 +41,7 @@
 | iex | ✓ |  |  |  |
 | java | ✓ |  |  | `jdtls` |
 | javascript | ✓ |  | ✓ | `typescript-language-server` |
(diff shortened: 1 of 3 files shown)


EricCrosson (approved):
Amazing! :rocket: Thanks @the-mikedavis

sudormrfbin (approved):

Issue #1125, opened 2021-11-18
automatic reload when file changes externally


dead10ck:
It would be great if it were detected when the file you are editing has been changed externally, and if possible, automatically reload the file.

I imagine it could work similarly to vim and kakoune, i.e. if there are no modifications that have yet to be written, just reload the file, and treat it as a single modification that can be undone. If there are modifications that haven't been written yet, prompt the user about what they would like to do:

* Ignore the changes, keep that's in the editor buffer, and let a write happen over the external change, whatever it was
* Reload the file anyway, essentially "adding" to the unsaved changes

archseer:
See https://github.com/helix-editor/helix/pull/588

jtrees:
Just some quick feedback:

I'm currently trying Helix as an alternative to Kakoune and the lack of file watching really hurts. I frequently have several windows open with Helix where I work on the same files. Since there is no server-client architecture, the buffers get out-of-sync and make this workflow virtually unusable.

I suppose the most viable workaround is to use splits in a single Helix instance, instead of letting my window manager handle the windows.

ryanfowler:
👋 Hi @pascalkuthe! I see you self-assigned this issue earlier in the year, but I would love to help move this feature along if possible.

I've taken a look at https://github.com/helix-editor/helix/pull/2653, but it seems like that branch has stalled (and doesn't include reloading open buffers). I'd like to propose some functionality that could be broken up into multiple steps/PRs:

**File watching and reloading open buffers**
This first step (and likely also the largest) would involve adding "file watching" functionality to the editor. I've looked into both the [`notify`](https://docs.rs/notify/latest/notify/index.html) crate, in addition to using an external tool like [`watchman`](https://facebook.github.io/watchman/). I think that using `notify` is the better approach that doesn't make Helix reliant on external tooling. I think a reasonable first implementation would watch the files in the current working directory, and take a list of excludes (probably globs?) to cut down on noise. VSCode uses some [default excludes](https://code.visualstudio.com/docs/getstarted/settings#_default-settings) (`"files.watcherExclude"`) that seem sensible. Once file watching is in place, open buffers that haven't been modified can be reloaded on change.

**Watching individual files that are opened, but not in the CWD**
For files that are opened but not watched because they are outside of the CWD, they can be separately watched via individual requests to the [`Watcher`](https://docs.rs/notify/latest/notify/trait.Watcher.html).

**Support LSP didChangeWatchedFiles**
[...]

PR #588, opened 2021-08-15
[WIP] file watching

oberblastmeister:
Currently only the file watching actor is implemented. I am not sure where to put it, is `helix-view` the right place? I also added `rustc_hash` because it is faster but less secure, but I don't think we need the security here.

TODO:
- debouncing
- handle atomic writes (from editors like vim)
- reload the file

PR #2653, opened 2022-06-03
File watcher implementation & LSP DidChangeWatchedFiles notification

### LINKED ISSUE #2479: No new dependencies detected (rust-analyzer)
**I don't know how exactly to define the title.**
When I open a rust workspace, everything goes fine, libraries and all are loaded, but when I add a new dependency inside Cargo.toml of a crate and try to use it in some file, it is not found or not available, I think that when I modify the Cargo.toml file, rust-analyzer does not verify the change and downloads the missing dependencies, the only way to make it appear is closing and reopening helix. In the case of vscode, as an example, when you add a new dependency, rust-analyzer verifies it and lets you use it in any file.

**In helix:**
image
image

**In vscode:**
image
image
image

kyrime:
Closes #2479 and is a prerequisite for #1125. Also probably supersedes #588.

I don't think this is quite ready to merge but it's definitely due for some reviewing. Functionality works fine for closing the intended issue but I've got concerns when it comes to other issues that depend on file watching.

`notify` is the main dependency added. I've reused `globset` as a dependency to parse the LSP spec's glob patterns, it's already included through `ignore`.

There are a couple different choices with regard to what directories to set `notify` watching: 

1. Recursively watch the current working directory.
2. Set watches only on the specific files that are of interest.
3. Same as (1) but mix in non-recursive watching to allow for ignoring certain directories. 

The issue with (1) is that there's no way to outright ignore directories: it seems likely that the amount of file changes `rust-analyzer` generates in the `target` directory on every write through `cargo check` could cause stuttering on lower-specced machines. The issue with (2) is that it can't detect new files as it's not watching for them, and the LSP may be interested in events from files that aren't open in the editor. So I've gone with (3) here: non-recursively watching the root directory, ignoring those matching some default list of build/cache directories ala Rust's `target` directory, and recursively watching the rest. This approach is taken from [emacs-lsp](https://github.com/emacs-lsp/lsp-mode/blob/master/docs/page/file-watchers.md), which has a seemingly pretty comprehensive list of ignored directories. The downside is that this adds some complexity with keeping track of directory creations/deletions/renames in the root directory, but it's not bad.

File watching works as follows: Watcher sets up `notify` watches on open LSP workspaces. Interested parties register a callback to be called when the watcher receives events matching a path filter in a specific workspace.

Some considerations:
- `DidChangeWatchedFiles` can only be registered dynamically, so supporting it requires handling `RegisterCapability` and `UnregisterCapability`. I haven't yet added support for unregistering the `DidChangeWatchedFiles` capability as I'm not sure where to store the required state: it seems to belong with the LSP client, but mutating the client seems to require interior mutability right now which gives me pause.
- Helix supports having multiple LSPs in use which all may have their own root workspace folders, so I've attempted to handle the case of multiple workspaces in the file watcher. I'm not really sure if that was the right move.
- Something weird here is that you might expect the watcher to default to watching everything from the current working directory down, but right now it just works with LSP workspaces. The current model also might cause issues with something like #1125 where any file could be open in the editor, even outside any workspace or the working directory, so seems like that would require mixing in some functionality to facilitate choice (2) above. Any thoughts would be welcome.
- I haven't wired up LSP clients to add/remove their workspaces to/from the watcher's list of workspaces when they're created or dropped yet, mostly because I'm not really sure if I'm taking the right approach with how workspaces are handled right now.
- The only build directory I have ignored right now is `target`. Happy to add others from `emacs-lsp`'s list in if maintainers are fine with this approach. Something nagging me is that since these directories are per-language it would also make sense to me if these were configured per-language, rather than one big list.
- I've used the pre-release version of `notify` which has been around for a while and seems pretty stable. It doesn't support debouncing yet but it seems like it's not too far off either. I haven't added debouncing to this PR yet either, it's possible it's best to just leave it as is and wait until `notify`'s debounce support hits if the lack of debouncing isn't causing any issues.

That's all for now off the top of my head, but there's probably other stuff lurking around in the PR.

### DIFF
--- a/helix-view/src/watcher.rs
+++ b/helix-view/src/watcher.rs
@@ -229,7 +229,9 @@ impl Dispatcher {
                     self.watches.remove(id);
                 }
                 WatcherEvent::AddWorkspace(path) => {
-                    self.watch_workspace(path);
+                    if let Err(e) = self.watch_workspace(path) {
+                        log::warn!("failed to watch workspace root: {}", e);
+                    }
                 }
                 WatcherEvent::RemoveWorkspace(path) => {
                     if let Some(dirs) = self.workspaces.remove(&path) {
@@ -260,8 +262,12 @@ impl Dispatcher {
             if let Some(dirs) = self.workspaces.get_mut(parent) {
                 match event.kind {
                     EventKind::Create(CreateKind::Folder) => {
-                        self.inner.watch(path, RecursiveMode::Recursive).unwrap();
-                        dirs.insert(path.to_owned());
+                        match self.inner.watch(path, RecursiveMode::Recursive) {
+                            Ok(()) => {
+                                dirs.insert(path.to_owned());
+                            }
+                            Err(e) => log::warn!("error watching new directory: {}", e),
+                        }
                     }
                     EventKind::Remove(RemoveKind::Folder) => {
                         dirs.remove(path);
@@ -269,11 +275,13 @@ impl Dispatcher {
                     EventKind::Modify(ModifyKind::Name(RenameMode::Both)) => {
                         // This event kind always involves two paths.
                         let new_path = event.paths.get(1).unwrap();
-                        self.inner
-                            .watch(new_path, RecursiveMode::Recursive)
-                            .unwrap();
-                        dirs.insert(new_path.to_owned());
-                        dirs.remove(path);
+                        match self.inner.watch(new_path, RecursiveMode::Recursive) {
+                            Ok(()) => {
+                                dirs.insert(new_path.to_owned());
+                                dirs.remove(path);
+                            }
+                            Err(e) => log::warn!("error watching renamed directory: {}", e),
+                        }
                     }
                     _ => {}
                 }
@@ -296,7 +304,7 @@ impl Dispatcher {
         }
     }
 
-    fn watch_workspace(&mut self, root: PathBuf) {
+    fn watch_workspace(&mut self, root: PathBuf) -> anyhow::Result<()> {
         // The set of individual directory watches issued, necessary for unwatching
         // the workspace.
         let mut watches = HashSet::new();
@@ -306,9 +314,7 @@ impl Dispatcher {
         // Note that if directories are created or deleted after calling
         // `add_directory`, they must have watches set/unset on them in response
         // to events from this watch call.
-        self.inner
-            .watch(&root, RecursiveMode::NonRecursive)
-            .unwrap();
+        self.inner.watch(&root, RecursiveMode::NonRecursive)?;
         watches.insert(root.clone());
 
         // Watch the root's children directories recursively, ignoring those in
@@ -325,12 +331,11 @@ impl Dispatcher {
                     .unwrap_or(true) // Allow non-utf8 paths.
             })
         {
-            self.inner
-                .watch(&entry.path(), RecursiveMode::Recursive)
-                .unwrap();
+            self.inner.watch(&entry.path(), RecursiveMode::Recursive)?;
             watches.insert(entry.path());
         }
 
         self.workspaces.insert(root, watches);
+        Ok(())
     }
 }
--- a/helix-view/src/watcher.rs
+++ b/helix-view/src/watcher.rs
@@ -153,7 +153,7 @@ impl FileWatcher {
         };
         let ev = WatcherEvent::Register(watch, tx);
         if let Err(e) = self.tx.send(ev) {
-            anyhow::bail!("file watcher channel error: {e}");
+            anyhow::bail!("file watcher channel error: {}", e);
         }
 
         Ok(rx.await?)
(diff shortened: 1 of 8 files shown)


helix-view/src/watcher.rs
```diff
@@ -0,0 +1,341 @@
...
+
+pub use notify::{event as notify_event, Event};
+
+// A list of directory names to ignore when found in the root of
+// a workspace. These are directories that would otherwise generate
+// a very large amount of events which could cause stuttering.
+// i.e. `rust-analyzer` calls `cargo check` on every write
+// which generates a large amount of fingerprint file changes in the
+// `target` directory.
+const IGNORE: &[&str] = &["target"];
```
Philipp-M, line 25: > ... ignoring those matching some default list of build/cache directories ala Rust's target directory

I think we can just use the .gitignore to filter files (or other .ignore files) to filter relevant files/folders.

They are used anyway already for the file-picker etc.

kyrime: I'm inclined to agree that `.gitignore` is the better approach now that I've thought about it some more. Originally I imagined that could lead to some surprising behaviour in edge cases depending on what the user's chosen to add to their `.gitignore`, but thinking about it now I can't really imagine many sane scenarios where something bad could happen.

For Rust and other languages that use such an integrated project/build system it should work well in the vast majority of cases, since of course `cargo new` automatically sets up a repo and gitignores `target`. Other languages I'm not so sure, but we can hope that the amount of cases where people would have an extremely active build directory that isn't present in `.gitignore` would be comparatively low.

This would also eliminate some edge cases like if a project is using a different name than `target` for its build directory.

archseer:
@kyrime Sorry for the delay on this! I've been meaning to review this for a while.

> The issue with (1) is that there's no way to outright ignore directories: it seems likely that the amount of file changes `rust-analyzer` generates in the `target` directory on every write through `cargo check` could cause stuttering on lower-specced machines.

Hmm isn't this how [watchman](https://facebook.github.io/watchman/) would handle this internally? A recursive watch, but a way to drop unwanted notifications very early to avoid sending it through the notification system.

`lsp-mode` seems to also do recursive watching: https://github.com/emacs-lsp/lsp-mode/blob/8091f5d13efbde5bfabc774970dd689487053333/lsp-mode.el#L1871-L1910

I'd start with a simple recursive watch and implement something more complicated only if it becomes necessary.

> it seems to belong with the LSP client, but mutating the client seems to require interior mutability right now which gives me pause.

We could use an `ArcSwap` on the server capabilities, similar to how we do the editor config. That way it stays immutable but we can swap it out with another updated capability struct.

> Something weird here is that you might expect the watcher to default to watching everything from the current working directory down, but right now it just works with LSP workspaces.

We should watch from the root directory down, then if a filepath is also inside the LSP workspace, we can also trigger a LSP notification.

> Happy to add others from emacs-lsp's list in if maintainers are fine with this approach. Something nagging me is that since these directories are per-language it would also make sense to me if these were configured per-language, rather than one big list.

I think we could use a global ignore list. The problem with per-language setting is that in a monorepo we might have a mix of different languages. Just because I'm editing a rust file doesn't mean there isn't a `node_modules` present in the same tree.

krobelus:
>  The issue with (1) is that there's no way to outright ignore directories
> Happy to add others from emacs-lsp's list in if maintainers are fine with this approach. Something nagging me is that since these directories are per-language it would also make sense to me if these were configured per-language, rather than one big list.
> I think we could use a global ignore list. The problem with per-language setting is that in a monorepo we might have a mix of different languages. Just because I'm editing a rust file doesn't mean there isn't a node_modules present in the same tree.

yeah this is annoying. I wish there was an editor-independent (in LSP or outside) way of marking paths as "not containing sources". Would be nice to piggy-back on gitignore but that's not exactly the same thing.

> I'd start with a simple recursive watch and implement something more complicated only if it becomes necessary.

I agree. This is also what [kak-lsp does](https://github.com/kak-lsp/kak-lsp/commit/45615dc868c8def1777202471fd35c45c7f1f41d) because I couldn't reproduce performance problems.
I still feel iffy about watching everything but maybe the OS-imposed limits take care of that

violin0622:
Any updates recently?

krobelus:
On a quick glance this seems reasonable apart from the needlessly-complex recursive watch logic.
It's probably appropriate to use `notify-debouncer-full`.

You should know that kakoune-lsp recently disabled the file watcher by default because it caused very high latency for `pyright-langserver` startup in a repo with many files.
(Didn't root cause it yet. Very odd because the file watcher runs in its own thread.)
I'm happy to test your implementation for this issue.

glehmann:
part of the work done here has already been merged with #7665

helix-view/src/watcher.rs
```diff
@@ -0,0 +1,341 @@
+use std::collections::{HashMap, HashSet};
```
nik-rev, line 1: this implementation will likely need to be moved to the event system to be merged, see also https://github.com/helix-editor/helix/pull/13020

nicosql:
> This feature is the only one preventing me from switching over to helix completely. I love helix, but being out of sync with the state of the file on the disk is a little jarring sometimes. What needs to be done for this to work?

archseer:
Replaced by #14544

PR #2654, opened 2022-06-03
ensure :quit and :quit! take no arguments

the-mikedavis:
I lost my train of thought and accidentally typed `:q! :w` 😄

Since quitting is a little destructive, I think it makes sense to fail the quit if it's not a well-formed 0-arg `:quit(!)`.

What do you think?


dylrich:
I think this is a good idea, particularly for `:q!`

archseer:
Agree, we should check `args.len()` in each command. This works as a quick fix though

Issue #2367, opened 2022-05-02
Panic on close last buffer


PushUpek:
### Summary

Bug on closing buffers. Helix crashes with panic error:
```thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-view/src/editor.rs:630:51```

### Reproduction Steps

 

I tried this:

1. `hx`
2. open file with (space + f)
3. open another file (space + f)
4. switch buffer to first file (space + b) 
5. use command `:bc` to close file
6. use command `:bc` to close last file

I expected this to happen:
Helix should remain open with an empty buffer.

Instead, this happened:
Helix crashes with panic error

### Helix log

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-view/src/editor.rs:630:51
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic
   3: helix_view::editor::Editor::replace_document_in_view
   4: helix_view::editor::Editor::close_document
   5: helix_term::commands::typed::buffer_close
   6: helix_term::commands::typed::command_mode::{{closure}}
   7: <helix_term::ui::prompt::Prompt as helix_term::compositor::Component>::handle_event
   8: helix_term::compositor::Compositor::handle_event
   9: helix_term::application::Application::handle_terminal_events
  10: helix_term::application::Application::event_loop::{{closure}}
  11: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  12: tokio::park::thread::CachedParkThread::block_on
  13: tokio::runtime::thread_pool::ThreadPool::block_on
  14: tokio::runtime::Runtime::block_on
  15: hx::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### Platform

macOS

### Terminal Emulator

iterm2 build: 3.4.15

### Helix Version

helix 22.05-dev (2c60798b)

**Comment by archseer** [comment]
Probably a result of https://github.com/helix-editor/helix/commit/52f5a4228aef2d31cf20aeee2bed317ba1767c18

\cc @matoous

**Comment by stjepangolemac** [comment]
Just updated to the latest release and this started happening.

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-view/src/editor.rs:633:51
```

the-mikedavis:
#2646 shows another way to provoke this: `:bufffer-close-all`

[...]

PR #2658, opened 2022-06-03
Fix panic on close last buffer (#2367)

gavynriebau:
In certain circumstances it was possible to cause a panic when closing
buffers due to some mishandling of view document history.

A change has been made to delete removed documents from the history of
accessed documents for each view. The ensures we don't attempt to jump
to a deleted document by mistake.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -814,8 +814,7 @@ impl Editor {
             .tree
             .views_mut()
             .filter_map(|(view, _focus)| {
-                // remove the document from jump list of all views
-                view.jumps.remove(&doc_id);
+                view.remove_document(&doc_id);
 
                 if view.doc == doc_id {
                     // something was previously open in the view, switch to previous doc
@@ -826,9 +825,6 @@ impl Editor {
                         Some(Action::Close(view.id))
                     }
                 } else {
-                    // documents also need be removed from the view "document access history"
-                    // so we don't accidentally try to jump back to them after they have been deleted
-                    view.docs_access_history.retain(|doc| doc != &doc_id);
                     None
                 }
             })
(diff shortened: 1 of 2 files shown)


helix-view/src/editor.rs
```diff
@@ -826,6 +826,9 @@ impl Editor {
                         Some(Action::Close(view.id))
                     }
                 } else {
+                    // documents also need be removed from the view "document access history"
+                    // so we don't accidentally try to jump back to them after they have been deleted
+                    view.docs_access_history.retain(|doc| doc != &doc_id);
```
archseer, line 831: Let's fix this in https://github.com/helix-editor/helix/blob/955a6cd540ca0483859b98512d7404b5f028b8d4/helix-view/src/view.rs#L64-L66=


The method is already called on line 818

archseer: Ah wait, that's not actually possible since it's a method on `view.jumps`. Perhaps we should add a method to `View` that cleans up both jumps and access history. Any place calling `jumps.remove` should then call this new method instead

gavynriebau: Ok, I'll have a go at doing it

gavynriebau: @archseer I've moved the cleanup into a `view.remove_document` function now, take a look when you have a moment.

We could also take this approach further and add an `Editor.remove_document` function which would look like:

```rust
impl Editor {
  fn remove_document(&mut self, doc_id: &DocumentId) {
    for view in self.tree.views_mut() {
      view.remove_document(&doc_id);
    }
    self.documents.remove(&doc_id);
  }
}
```

archseer:
There's one more occurrence of view.jumps.remove in the file that needs to be changed to use the new function

gavynriebau:
> There's one more occurrence of view.jumps.remove in the file that needs to be changed to use the new function 

I've now switched the last occurrence of `view.jumps.remove` over

archseer (approved):

Issue #1216, opened 2021-12-01
Add ability to use a specified config/theme


tmforshaw:
I'd like to be able to use helix as root user but still keep my config and theme.

In vim you can [pass a config file](https://unix.stackexchange.com/questions/46699/start-vim-as-my-user-with-root-privileges) using the `-u` flag

``` bash
sudo vim -u /home/user/.vimrc /path/to/file.txt
```

Maybe helix could accept a `-c` flag so that any config file can be used.

``` bash
sudo hx -c /home/user/.config/helix/config.toml /path/to/file.txt
```

Thanks for reading :)

sudormrfbin:
For your specific usecase the recommended way would be to use [`sudoedit` after setting `$EDITOR`](https://stackoverflow.com/questions/21488357/use-sudo-with-vimrc) to `hx`.

tmforshaw:
I saw that I could do that, but I don't have sudo installed on my system as I use doas.

If this is the solution then I will use it for now, but I still think this would be a nice feature in the future.

Thanks for your reply

NNBnh:
Good new! [`doasedit`](https://aur.archlinux.org/packages/doasedit) exist.

tmforshaw:
Ah wow, that is good news!

Thanks a lot

the-mikedavis:
The new configuration in https://github.com/helix-editor/helix/pull/2135 should help make this easier to implement: we can have a single switch for the config file and then the config file can point to other configuration files and directories.

yuuyins:
i don't use helix as my primary editor i can't set it to $EDITOR. so having the option --config path/to/config is pretty useful

sudormrfbin:
@yuuyins You could also set [`$SUDO_EDITOR`][1] instead of `$EDITOR`. This doesn't help if you want to use another editor and helix interchangeably with `sudo` though.

[1]: https://www.mankier.com/8/sudo#--edit

yuuyins:
@sudormrfbin that worked thank you!

PR #2135, opened 2022-04-16
Use XDG dirs per spec

mmatous:
OK, so this ended up a bit bigger than expected. I'm not exactly happy about the static but that was what HELIX_RUNTIME did and the rest of the code relied on that. Passing the paths from config struct could be another PR.

Locations implemented as outlined in https://github.com/helix-editor/helix/issues/584#issuecomment-1086768512. 

Getting paths from loader will panic if they're not loaded properly via `init_paths`. It could silently fallback to some default, but not calling init_paths is a programmer mistake and it's good feedback when dealing with tests and build.rs that would otherwise silently write outside of their intended sandbox to regular user directories.

PR #1814, opened 2022-03-15
Continue loading the config on error

antoyo:
Fix #1732

There are cases where this improve the error that is shown to the user, but in some cases, the error message gets worse, i.e. it loses its context (key name) and/or position.
I'll try to find a way to improve this, but I'm not sure how, so suggestions are welcomed! (I tried serde-path-to-error, but it does not seem like it works when called from `ok_or_default`.)

I'm submitting this PR now to gather early reviews.

PR #2666, opened 2022-06-03
add a CLI flag for specifying config file location

the-mikedavis:
Closes https://github.com/helix-editor/helix/issues/1216

I don't think there shouldn't be follow-up flags for the theme/runtime/etc dirs: https://github.com/helix-editor/helix/pull/2135 should cover those cases.

I want this mostly to make triage easier - trying out one-off keymaps - and to some extent as an alternative to https://github.com/helix-editor/helix/pull/1814 (say I try features ahead of time, I'll want to `alias hx='hx -c path/to/config/with/extra/config.toml'` so `cargo run` doesn't reject my config when trying other PRs). In the Nix world this should also make it easier to use helix with something like home-manager and not keep config in `~/.config/helix`.


the-mikedavis:
I'm not sure a `once_cell` is the right tool for this. It was the most straightforward way to not change the `helix_loader::config_file()` API but maybe I'm missing something more idiomatic? Or this shouldn't be done without shared memory at all?

PR #1802, opened 2022-03-12
visible whitespace

the-mikedavis:
Example with this rebased on top of #1796:

whitespace-with-indent-guides

Adds some configuration options to `config.toml`:

```toml
# render whitespace characters (\t, \n, ' ')
[editor.whitespace]
render = "all"
# render = "none"    to disable visible whitespace

# or control whether each character is shown:
[editor.whitespace.render]
space = "all"
tab = "all"
newline = "none"

# control the grapheme used to represent the whitespace
[editor.whitespace.characters]
space = "·"
tab = "→"
newline = "⏎"
```

And a new scope for themes `ui.virtual.whitespace`, which should be set to something slightly off-background.

This makes way for a future `"editor.whitespace.render" = "selection"` as well, which could enable showing whitespace only in a selection rather than always.

TODOs:

- [ ] non breaking space character? (U+00A0F, default: `⍽`)
- [x] update existing themes to include `ui.virtual`
- [x] add new config documentation to the book
- [x] decide whether this should be off by default
- [x] squash down commits 🔨

connects #1068 
closes #1218
closes #688

PR #2669, opened 2022-06-03
Default rulers color to red

Mathspy:
Currently if the theme a user is using doesn't have `ui.virtual.rulers` set and they set up a ruler it just fails silently making it really hard to figure out what went wrong. Did they set incorrectly set the ruler? Are they using an outdated version of Helix that doesn't support rulers?

This happened to me today, I even switched to the default theme with the assumption that maybe my theme just doesn't have the rulers setup properly and it still didn't work.

Not sure if this is a good idea or not, feel free to suggest better alternatives!


helix-term/src/ui/editor.rs
```diff
@@ -170,7 +170,20 @@ impl EditorView {
         theme: &Theme,
     ) {
         let editor_rulers = &editor.config().rulers;
-        let ruler_theme = theme.get("ui.virtual.ruler");
+        let default_style = Style {
+            bg: Some(Color::Red),
+            ..Default::default()
+        };
```
the-mikedavis, line 176:
```suggestion
        let default_style = Style::default().bg(Color::Red);
```

helix-term/src/ui/editor.rs
```diff
@@ -170,7 +170,20 @@ impl EditorView {
...
+        let ruler_theme = theme
+            .try_get("ui.virtual.ruler")
+            .map(|style| {
+                if let None = style.bg {
+                    default_style
+                } else {
+                    style
+                }
+            })
```
the-mikedavis, line 185:
```suggestion
            .filter(|style| style.bg.is_some())
```

the-mikedavis: With this map/filter though, you can't create a ruler that is `fg` only, and `fg`-only rulers are kinda nice because they aren't visible until text crosses them.

I think having a louder default is a good idea. If we remove this filter line we'll still have good fallback behavior when a theme doesn't define `ui.virtual.ruler` or `ui.virtual` (or `ui`) at all.

Mathspy: The reason I added this is because pretty much all themes define `ui.virtual` after #1802 and pretty much all of them define it as just foreground color, which makes it fairly unobvious until the text reaches it as you mention

Mathspy: What about adding a `get_exact` to theme that doesn't fallback to `ui.virtual`?

the-mikedavis: `get_exact` would make this scope special. I would prefer that all scopes have fallback behavior for consistency's sake.

Having rulers themed with `ui.ruler` might be preferrable since `ui.virtual.whitespace` and `ui.virtual.ruler` end up conflicting which one wants `fg` and which wants `bg`. This would be a breaking change though, although I suppose that could be mitigated by adding code here to fall back to `ui.virtual.ruler` and logging a warning.

Alternatively, I think it a PR that adds `ui.virtual.ruler` to all in-repo themes that don't already have it would go a long way towards alleviating this.

Mathspy: > `get_exact` would make this scope special. I would prefer that all scopes have fallback behavior for consistency's sake.

Yeah I was worried about that, makes sense to avoid it

> Having rulers themed with `ui.ruler` might be preferrable since `ui.virtual.whitespace` and `ui.virtual.ruler` end up conflicting which one wants `fg` and which wants `bg`. This would be a breaking change though, although I suppose that could be mitigated by adding code here to fall back to `ui.virtual.ruler` and logging a warning.

That makes sense, but I am not really sure if there's precedence I can look at for how to emit that warning.

> Alternatively, I think it a PR that adds `ui.virtual.ruler` to all in-repo themes that don't already have it would go a long way towards alleviating this.

Yeah I agree, I was considering doing that but realized it might still trip people adding themes in the future or people using custom themes so I tried this and it worked well enough so I wanted to see what people thought of this idea

the-mikedavis: Actually I think switching to

```rust
let ruler_theme = theme
    .try_get("ui.ruler")
    .unwrap_or_else(|| Style::default().bg(Color::Red));
```

might be the best option: it'll be very noticeable when a theme doesn't have `ui.ruler` and it won't fall back to `ui.virtual`.

Mathspy: I just thought of another alternative that lets us have a loud default for everyone without breaking changes. How about we update all current themes from `"ui.virtual"` to `"ui.virtual.whitespace"` and then we can just filter by whether there is any style or not, in most cases there won't be so it will default to red

the-mikedavis: Yeah that's a good idea 👍

Mathspy: Should be good now!

runtime/themes/onelight.toml
```diff
@@ -79,7 +79,7 @@
 "ui.selection" = { bg = "light-white" }
 "ui.selection.primary" = { bg = "light-white", modifiers = ["dim"] }
 
-"ui.virtual" = { fg = "light-white"}
+"ui.virtual.whitespace" = { fg = "light-white"}
 "ui.virtual.ruler" = { bg = "light-white" }
 "ui.virtual.whitespace" = {fg = "light-white" }
```
the-mikedavis, line 84: looks like this one already makes the distinction between ruler and whitespace

Mathspy: Ops, good catch. Should I just get rid of `"ui.virtual"` for this one?

the-mikedavis: Hmm I'd say let's leave it. It might be a good default for scopes in the future

Mathspy: All right!

Mathspy:
(Rebased to resolve a conflict)

PR #2670, opened 2022-06-04
Avoid modifying jumplist until jumping to ref

### LINKED ISSUE #2663: Avoid saving to jumplist when goto is cancelled
When invoking `goto_reference`, the current selection will always be saved to the jumplist, even if the goto command is cancelled by pressing Esc in the file picker.

When a goto command is cancelled, no effect is expected (except for querying LSP). But in current implementation, the jumplist will be tampered.

I think this issue can be fixed by the following patch, but I'm new to Rust and not sure if there are other use cases I haven't thought of.

```patch
diff --git a/helix-term/src/commands/lsp.rs b/helix-term/src/commands/lsp.rs
index bfedb209..8f139b3b 100644
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -485,8 +485,6 @@ fn goto_impl(
     locations: Vec<lsp::Location>,
     offset_encoding: OffsetEncoding,
 ) {
-    push_jump(editor);
-
     let cwdir = std::env::current_dir().expect("couldn't determine current directory");

     match locations.as_slice() {
@@ -520,6 +518,7 @@ fn goto_impl(
                     format!("{}:{}", file, line).into()
                 },
                 move |cx, location, action| {
+                    push_jump(cx.editor);
                     jump_to_location(cx.editor, location, offset_encoding, action)
                 },
                 move |_editor, location| Some(location_to_file_location(location)),
```

yzwduck:
When a goto command is cancelled, the jumplist should remain unchanged.

This commit delays saving the current selection to the jumplist until jumping to a reference.

Fixes #2663

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -485,8 +485,6 @@ fn goto_impl(
     locations: Vec<lsp::Location>,
     offset_encoding: OffsetEncoding,
 ) {
-    push_jump(editor);
-
     let cwdir = std::env::current_dir().expect("couldn't determine current directory");
 
     match locations.as_slice() {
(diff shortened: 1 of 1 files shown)


archseer (approved):
Thanks!

Issue #2645, opened 2022-06-02
Reflow and CRLF line endings don't work


fabric-and-ink:
### Summary

If line endings of a file are CRLF, the reflow command does not work correctly.

### Reproduction Steps

Have a text, e.g.

```
hello world
hello world
etc...
```

set `:line-endings crlf` and reflow the text. The result will be... strange.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

Windows

### Terminal Emulator

Terminal

### Helix Version

helix 0.6.0

PR #2672, opened 2022-06-04
Fix reflow for CRLF line ending (wip)

koiuo:
This is a fix for the #2645.

It assumes a fix for the _textwrap_ library, which is already in `master` but not yet released.

See details here https://github.com/mgeisler/textwrap/issues/453

Here's a quick demonstration of how it works
https://asciinema.org/a/KZNzsnuHYXppwzw6tvVHyfiSe

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1033,8 +1033,7 @@ dependencies = [
 [[package]]
 name = "textwrap"
 version = "0.15.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b1141d4d61095b28419e22cb0bbf02755f5e54e0526f97f1e3d1d160e60885fb"
+source = "git+https://github.com/mgeisler/textwrap?rev=9c648da#9c648da603fd99f4b37aeafbe5c617c3e1b64a1f"
 dependencies = [
  "smawk",
  "unicode-linebreak",
--- a/helix-core/Cargo.toml
+++ b/helix-core/Cargo.toml
@@ -41,7 +41,7 @@ encoding_rs = "0.8"
 chrono = { version = "0.4", default-features = false, features = ["alloc", "std"] }
 
 etcetera = "0.4"
-textwrap = "0.15.0"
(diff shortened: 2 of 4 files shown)


edio:
@kirawi , could you please take a look and let me know if overall approach looks ok?
Thanks for your time.

the-mikedavis:
This looks good besides the clippy lint. Looks like your PR was merged but isn't released in a version of textwrap yet? https://github.com/mgeisler/textwrap/pull/454

edio:
@the-mikedavis , I fixed the clippy error.

Yes, the change in textwrap is not yet released. 
Are you are fine with this PR overall? If you intend to merge some time in the future I'll ask textwrap maintainer for an ETA on their side. If you want to merge asap, I'll ask to cut a release.

the-mikedavis:
Yeah this is looking good to me 👍. No need to rush I don't think, it'd be good to have at least one other reviewer check this out before merge anyways

mgeisler:
> Yes, the change in textwrap is not yet released.

There has been a Textwrap release now, you should be able to depend on 0.16.0 and get the line-ending control you need.

pascalkuthe:
I am not sure is this is the right approach. It fixes the common case of CRLF linenedings but actually just exposes a more deeply rooted issue: text-wrap handles line breaks differently then helix. In fact this PR will cause new bugs for crlf files which also contain LF linenedings which are not detected as line breaks at all anymore with this PR. It also fails for other supported (unicode) line-ending and CR line endings. 

AFAIK the `line_ending` field on `Document` is just the line break that is used for inserting new line but the document may contain other linebreaks. I wish `text-wrap` would just let us supply our own `lines` iterator. We could use ropeys `lines` iterator here which correctly accounts for the linebreak feature flags helix uses. We could easily strip the EOL character ourselfs.

pascalkuthe:
I am closing this pr as it has gone stale (over a year without development). Like I said in a different issue we are planning to move away from textwrap and would instead reuese our own softwrap infrastructure in the long run to gain the flexibility to solve issues like this. 

Thank you for contributing!

PR #2675, opened 2022-06-04
Sort themes, language & files by score & then name

michaeljones:
*I haven't contributed before. I'm not attached to this change. I realise it might not be considered worth while. Just excited by the project and noticed this minor thing.*

Previously the themes were appearing unordered after typing `:theme `.
This sorts them first by fuzzy score and then by name so that they
generally appear in a more ordered fashion in the initial list.

The sort by name does not really pay off when there is a score so an
alternative approach would be to sort by score if there is string to
fuzzy match against and otherwise sort by name.

I've lowercased the names as that avoids lower case & upper case letters
being sorted into separate groups. There might be a preferable approach
to that though.

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -255,7 +255,9 @@ pub mod completers {
             })
             .collect();
 
-        matches.sort_unstable_by_key(|(name, score)| (Reverse(*score), name.to_lowercase()));
+        matches.sort_unstable_by(|(name1, score1), (name2, score2)| {
+            (Reverse(*score1), name1).cmp(&(Reverse(*score2), name2))
+        });
         names = matches.into_iter().map(|(name, _)| ((0..), name)).collect();
 
         names
@@ -311,7 +313,9 @@ pub mod completers {
             })
             .collect();
 
-        matches.sort_unstable_by_key(|(_language, score)| Reverse(*score));
+        matches.sort_unstable_by(|(language1, score1), (language2, score2)| {
+            (Reverse(*score1), language1).cmp(&(Reverse(*score2), language2))
+        });
 
         matches
             .into_iter()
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/mod.rs
```diff
@@ -255,7 +255,7 @@ pub mod completers {
             })
             .collect();
 
-        matches.sort_unstable_by_key(|(_file, score)| Reverse(*score));
+        matches.sort_unstable_by_key(|(name, score)| (Reverse(*score), name.to_lowercase()));
```
archseer, line 258: Seems reasonable, but let's use `name` here. `to_lowercase` causes a lot of allocations

michaeljones: I'm very happy if someone does that but as `name` is a `std::borrow::Cow<'_, str>` and it seems like we can't use it directly as a sort key otherwise we get:

```rust
   Compiling helix-term v0.6.0 (/home/michael/root/projects/helix/helix-term)
error: lifetime may not live long enough
   --> helix-term/src/ui/mod.rs:258:54
    |
258 |         matches.sort_unstable_by_key(|(name, score)| (Reverse(*score), name));
    |                                       -------------- ^^^^^^^^^^^^^^^^^^^^^^^ returning this value requires that `'1` must outlive `'2`
    |                                       |            |
    |                                       |            return type of closure is (std::cmp::Reverse<i64>, &'2 std::borrow::Cow<'_, str>)
    |                                       has type `&'1 (std::borrow::Cow<'_, str>, i64)`

error: could not compile `helix-term` due to previous error
```

and I don't understand the error messages or the best way to work around them.

michaeljones:
There are also other parts of that source code that might have the same change. We sort `settings`, `language` and `filename_impl` by reverse score and not alphabetically. I haven't looked into triggering those though so I'm not sure what the user experience is like.

helix-term/src/ui/mod.rs
archseer: No, cloning a `&str` allocates and gives you an owned type `String`. Taking a look, I think we have to use `sort_unstable_by` instead of `by_key`

archseer: https://stackoverflow.com/questions/47121985/why-cant-i-use-a-key-function-that-returns-a-reference-when-sorting-a-vector-wi

helix-term/src/ui/mod.rs
```diff
@@ -426,13 +430,18 @@ pub mod completers {
...
+            });
+
             files = matches
                 .into_iter()
                 .map(|(file, _)| (range.clone(), file))
                 .collect();
 
             // TODO: complete to longest common match
+        } else {
+            files.sort_unstable_by(|(_, path1), (_, path2)| path1.cmp(&path2));
```
the-mikedavis, line 444:
```suggestion
            files.sort_unstable_by(|(_, path1), (_, path2)| path1.cmp(path2));
```

michaeljones:
As you've noted, I've added a commit which changes to the non-allocating sorting approach and adds sorting to the language & files code paths.

archseer (approved):
Thanks!

PR #2677, opened 2022-06-05
simplify some keymap key names

### LINKED ISSUE #2369: Simplify some keymap key names
> I think removing some of the named keys like `semicolon`, `percent`, `lt` etc and letting them be parsed from their normal representation (`;`, `%`, `<`, etc) might be better altogether. We only really need them for keys with no printable ASCII representation like `backspace` (and maybe `minus` since `A--` for Alt+minus might be confusing to read).

@sudormrfbin https://github.com/helix-editor/helix/pull/2278

EricHenry:
## Summary

Closes #2369 

Removed some of the existing named keys and added tests. Please let me know if there is anything more that I can do.

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -41,12 +41,7 @@ pub(crate) mod keys {
     pub(crate) const NULL: &str = "null";
     pub(crate) const ESC: &str = "esc";
     pub(crate) const SPACE: &str = "space";
-    pub(crate) const LESS_THAN: &str = "lt";
-    pub(crate) const GREATER_THAN: &str = "gt";
-    pub(crate) const PLUS: &str = "plus";
     pub(crate) const MINUS: &str = "minus";
-    pub(crate) const SEMICOLON: &str = "semicolon";
-    pub(crate) const PERCENT: &str = "percent";
 }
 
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
Thanks!

helix-view/src/input.rs
```diff
@@ -375,6 +363,16 @@ mod test {
...
+        assert!(str::parse::<KeyEvent>("S-percent").is_err());
+    }
+
+    #[test]
+    fn parsing_unsupported_named_keys() {
+        assert!(str::parse::<KeyEvent>("lt").is_err());
+        assert!(str::parse::<KeyEvent>("gt").is_err());
+        assert!(str::parse::<KeyEvent>("plus").is_err());
+        assert!(str::parse::<KeyEvent>("percent").is_err());
+        assert!(str::parse::<KeyEvent>("semicolon").is_err());
```
archseer, line 375: Rather than just testing these keys are now unsupported, shouldn't we also test that they get parsed correctly by the new syntax?

EricHenry: @archseer Great catch! I intended to do this, but only added a test for `percent`. I created a new PR to add in the additional tests. #2694

PR #2680, opened 2022-06-06
support for openscad

### LINKED ISSUE #1415: Add support for OpenSCAD `.scad` files

I'd like to use helix to edit my scad files. I think this already kind of prepared: There already is an [in progress LSP server](https://lib.rs/crates/openscad-language-server) and grammar [definitions for tree-sitter](https://github.com/bollian/tree-sitter-openscad).

farwyler:
This adds basic support for [openscad](https://openscad.org/index.html) and highlighting.

closes #1415

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1433,3 +1433,18 @@ indent = { tab-width = 2, unit = "  " }
 [[grammar]]
 name = "jsdoc"
 source = { git = "https://github.com/tree-sitter/tree-sitter-jsdoc", rev = "189a6a4829beb9cdbe837260653b4a3dfb0cc3db" }
+
+
+[[language]]
+name = "openscad"
+scope = "source.openscad"
+injection-regex = "openscad"
+file-types = ["scad"]
+roots = []
+comment-token = "//"
+language-server = { command = "openscad-language-server" }
+indent = { tab-width = 2, unit = "  " }
+
+[[grammar]]
+name = "openscad"
+source = { git = "https://github.com/bollian/tree-sitter-openscad", rev = "5c3ce93df0ac1da7197cf6ae125aade26d6b8972" }
--- a/languages.toml
+++ b/languages.toml
@@ -1443,7 +1443,7 @@ file-types = ["scad"]
 roots = []
 comment-token = "//"
 language-server = { command = "openscad-language-server" }
-indent = { tab-width = 2, unit = "  " }
+indent = { tab-width = 2, unit = "\t" }
 
 [[grammar]]
 name = "openscad"
--- a/runtime/queries/openscad/highlights.scm
+++ b/runtime/queries/openscad/highlights.scm
@@ -1,0 +1,51 @@
+(number) @number
+(string) @string
+(boolean) @constant.builtin
+(include_path) @string
+
+(function_call function: (identifier) @function)
+(module_call name: (identifier) @function)
+
+(identifier) @variable
+(special_variable) @variable.builtin
+
+[
+  "module"
+  "function"
+  "for"
+  "intersection_for"
+  "if"
+  "let"
+  "assign"
+  "use"
+  "include"
+  "each"
+] @keyword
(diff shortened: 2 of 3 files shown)


runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,51 @@
+(number) @number
```
the-mikedavis, line 1: Some of the scopes in these queries need to be replaced with the ones helix uses: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

`@number` should become `@constant.numeric`, the `(boolean) @constant.builtin` stanza can become `@constant.builtin.boolean` and `@delimeter` should become `@punctuation.delimiter`

farwyler: thanks for the feedback. i've renamed and split up the scopes a little.
makes sense and looks better now. still learning hx 😊

runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+  "?"
+  "!"
+  ":"
+] @operator
+
+[
+  ";"
+  ","
+  "."
+] @punctuation.delimeter
```
the-mikedavis, line 58:
```suggestion
] @punctuation.delimiter
```

looks like this is a typo in the upstream repo, I make the same typo all the time 😄

runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,60 @@
+(number) @constant.number
```
the-mikedavis, line 1:
```suggestion
(number) @constant.numeric
```

runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+(string) @constant.string
```
the-mikedavis, line 2:
```suggestion
(string) @string
```

runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,60 @@
...
+(boolean) @constant.builtin.boolean
+(include_path) @string
```
the-mikedavis, line 4: depending on what this means in the language, `@string.special.path` might be a good option. If it isn't a file-system path then just `@string` is fine

languages.toml
```diff
@@ -1433,3 +1433,18 @@ indent = { tab-width = 2, unit = "  " }
...
+
+[[language]]
+name = "openscad"
+scope = "source.openscad"
+injection-regex = "openscad"
+file-types = ["scad"]
+roots = []
+comment-token = "//"
+language-server = { command = "openscad-language-server" }
+indent = { tab-width = 2, unit = "  " }
```
the-mikedavis, line 1446: Are tabs more prevelant for openscad? If so this can be

```suggestion
indent = { tab-width = 2, unit = "\t" }
```

So tab characters are used for indentation

runtime/queries/openscad/highlights.scm
```diff
@@ -0,0 +1,60 @@
+(number) @constant.numeric
+(string) @string
+(boolean) @constant.builtin.boolean
+(include_path) @string.special.path
+
+(function_call function: (identifier) @function)
```
the-mikedavis, line 6: similar to the above capture, could we add

```
(function_declaration name: (identifier) @function)
```

which marks function names in declarations as functions. Other languages have similar rules

```scad
function flatten(l) = [for (a = l) for (b = a) b];
//         ^ function
```

And along those lines:

```
(parameters_declaration (identifier) @variable.parameter)
```

anywhere above that `(identifier) @variable)` line (so it takes precedence) which would be

```scad
function flatten(l) = [for (a = l) for (b = a) b];
//               ^ variable.parameter
```

farwyler: works great! thanks

PR #2681, opened 2022-06-06
Support opening symbol definition in a separate view.

localghost:
I would like to be able to open a symbol definition in a split view rather than replacing the current one. I have not found support for this in helix. This PR is a request for navigating me to the right direction rather than a final implementation (it does not fix all possible code branches and there is some code duplication). I suspect there might be already some more general idea/design about passing a view as a destination for a command.
Much appreciate any suggestions.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -272,7 +272,7 @@ impl MappableCommand {
         normal_mode, "Enter normal mode",
         select_mode, "Enter selection extend mode",
         exit_select_mode, "Exit selection mode",
-        goto_definition, "Goto definition",
+        goto_definition_replace, "Goto definition",
         add_newline_above, "Add newline above",
         add_newline_below, "Add newline below",
         goto_type_definition, "Goto type definition",
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -4,7 +4,7 @@ use helix_lsp::{
     OffsetEncoding,
 };
 
-use super::{align_view, push_jump, Align, Context, Editor};
+use super::{align_view, job, push_jump, Align, Context, Editor};
 
 use helix_core::Selection;
 use helix_view::editor::Action;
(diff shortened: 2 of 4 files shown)


heliostatic:
I like this functionality -- I thought the picker already did have a facility for this, but it would still require refactoring the symbol picker to use it: https://github.com/helix-editor/helix/blob/10415a80690256c2ba40e2608191b8a081c7e21b/helix-term/src/ui/picker.rs#L530-L540

kostrzewa9ld:
@heliostatic Thanks for the tip.

So it seems only the case with single location does not support opening in a separate view.
So now I wonder how to smoothly merge typed command that allows picking action with the fact that when there are multiple symbols found picker pops up and user once again is able to pick the action.
For example: user types `:goto_defintion vsplit` to open the symbol in a separate vertical view (I have it mapped to `gDv`) but multiple symbols are found and so picker pops up and here user presses enter, and so should `Action::Replace` be chosen which is [defined in picker](https://github.com/helix-editor/helix/blob/10415a80690256c2ba40e2608191b8a081c7e21b/helix-term/src/ui/picker.rs#L524-L529) or should `Action::VerticalSplit` be chosen, which is what the command was invoked with.

heliostatic:
> For example: user types `:goto_defintion vsplit` to open the symbol in a separate vertical view (I have it mapped to `gDv`) but multiple symbols are found and so picker pops up and here user presses enter, and so should `Action::Replace` be chosen which is [defined in picker](https://github.com/helix-editor/helix/blob/10415a80690256c2ba40e2608191b8a081c7e21b/helix-term/src/ui/picker.rs#L524-L529) or should `Action::VerticalSplit` be chosen, which is what the command was invoked with.

Ah, tricky. My vote is to follow most recent user intent -- if the user starts with `:goto_definition vsplit` but then uses `<C-s>` to open a horizontal split in the picker, I read it as a change in user intent. If they just hit `enter`, then I would assume original intent of `vsplit`. But ultimately this is going to be a question of taste, I think.

pascalkuthe:
t his pr has gone quite stale and has not seen development in a long run. I don't see us moving in this direction (https://github.com/helix-editor/helix/pull/5383#issuecomment-1370213000). 

Thank you for contributing!

PR #2685, opened 2022-06-06
Unimplemented repeat multiple commands

lazytanuki:
Hi !
This PR _partially_ resolves a `FIXME`. It allows the editor to enter insert mode through a sequence of commands without panicking, making possible the use of user mappings such as:

```toml
a = ["append_mode", "collapse_selection"]
C = ["extend_to_line_end", "change_selection"]
# ...
```

And of course the sequence is repeatable with "."

There are still the cases of pending, not found and cancelled keys to handle though.


lazytanuki:
Oh I didn't see this one, thanks !

PR #2687, opened 2022-06-06
Improve markdown list rendering

Frojdholm:
Render line breaks parsed by the markdown parser.

Before:
python-list

After:
python-list-fixed


archseer:
Does that correctly indent multi-level lists? I tried fixing this before but the code needed significantly more changes to track indentation levels (I briefly looked at https://codeberg.org/flausch/mdcat/src/branch/main/src/render.rs)

Frojdholm:
> Does that correctly indent multi-level lists?

No, though the previous solution doesn't seem to work with it either. I'm writing some tests to investigate that further.

This doesn't actually change how lists work. Rather I think the keyword arguments are not even recognized as lists, but the line endings themselves are converted into line breaks. This change simply inserts line breaks in the same way as line breaks are inserted at the end of for example paragraphs.

Frojdholm:
@archseer now it handles nested lists, however it's very brittle since depending on the input source the parser can start inserting paragraphs into the list items.

For example this input
```md
- item 1
    - subitem 1
- item 2
    - subitem 2

text
- item 1
    - subitem 1
```
breaks this parsing.

The problem seems pretty complex with tagged events being able to nest almost arbitrarily.

helix-term/src/ui/markdown.rs
```diff
@@ -164,22 +171,51 @@ impl Markdown {
...
+
+                    // whenever top-level list closes, empty line
+                    if list_stack.is_empty() {
+                        lines.push(Spans::default());
+                    }
                 }
                 Event::Start(Tag::Item) => {
+                    // FIXME: What to do in this case? This is a bug in the markdown parser and we
+                    // cannot really continue, but obviously we shouldn't crash the editor.
+                    assert!(list_stack.len() > 0, "Cannot start an item without a list");
```
Frojdholm, line 196: This will have to be fixed before this can be merged. If this assert triggers there is a bug in the markdown parsing, but the editor shouldn't crash because of it.

helix-term/src/ui/markdown.rs
```diff
@@ -299,3 +335,60 @@ impl Component for Markdown {
         Some((width + padding, height + padding))
     }
 }
+
+#[cfg(test)]
+mod tests {
```
Frojdholm, line 340: Do you want me to include the test code?

Frojdholm (commented):
Code now handles nested, numbered and multi-paragraph lists. Hard and soft line breaks and multiple paragraphs are indented on a best effort to match the overall indentation of the list.

Two spaces are currently chosen as the default indentation per level, but could be changed.

Known bugs:
- Fenced code blocks are not working properly inside of lists.

archseer (approved):
This looks great! Thanks

PR #2689, opened 2022-06-06
Do not add extra line breaks in markdown lists

Frojdholm:
Before:
list-newlines

After:
list-newlines-fixed


archseer:
This looks fine 👍🏻

Issue #1176, opened 2021-11-26
Resize windows


bestouff:
There's currently no way (that I see) to resize a window, once split.
There probably need some actions like "window-enlarge-horizontal", by default bound to keys like "C-S-right".

Termina94:
I have a branch working for horizontal resizing using a flex-like modal as recommended here by @archseer #1185
Has an interesting effect when expanding a centre buffer of 3 however lacks the fluidity of having a float/percentage modal

Which of the following two methods has the preferred outcome? Or do we add a config option to choose between either?

flex:
![ezgif com-gif-maker](https://user-images.githubusercontent.com/47091253/171056449-ed16907b-7703-4bd5-bd56-0d0a33db5008.gif)
float/percentage:
![ezgif com-gif-maker (1)](https://user-images.githubusercontent.com/47091253/171056469-8646d4fa-e628-4b2e-b374-dbd1742020ec.gif)

archseer:
Hmm, how different is the implementation between the two?

Termina94:
I have been playing around with different solutions to this issue and the nicest way I have found is using a flex-like approach where the parents container is split between the children bounds.

i.e | 1 | 1 | or | 2 | 1 | etc.

However this will not allow for reducing the size of a container that defaults to the value 1 and therefore all resizing will have to be managed by growing other children to make the one you want smaller which is not very intuitive.

For this draft example I have decided to hard code a slot limit per child of 20 and the each child starts with a default of 10. This feels much nicer but limits the accuracy of where you can resize your buffers to with a max of 20 potential sizes.

I have also added a focus toggle that will set the current buffers bounds to double the limit, ensuring it takes the majority of screen space allowing quick toggle to focus on different files which works quite nice with the flex approach.

As it stands this will be a draft until people have a play with the branch and we decide on a solution

https://github.com/helix-editor/helix/pull/2704

cor:
id like to add that it would be really nice if you could use the mouse for resizing windows by dragging on the windows borders

brokenthorn:
[...]

PR #2704, opened 2022-06-07
Add flex resize, focus mode, hard-coded limits for now

Termina94:
I have been playing around with different solutions to this issue and the nicest way I have found is using a flex-like approach where the parents container is split between the children bounds.

i.e     |  1  |  1  |    or    |   2   |  1  |    etc.

However this will not allow for reducing the size of a container that defaults to the value 1 and therefore all resizing will have to be managed by growing other children to make the one you want smaller which is not very intuitive.

For this draft example I have decided to hard code a `slot` limit per child of 20 and the each child starts with a default of 10. This feels much nicer but limits the accuracy of where you can resize your buffers to with a max of 20 potential sizes.

I have also added a `focus` toggle that will set the current buffers bounds to double the limit, ensuring it takes the majority of screen space allowing quick toggle to focus on different files which works quite nice with the flex approach.

As it stands this will be a draft until people have a play with the branch and we decide on a solution

https://github.com/helix-editor/helix/issues/1176


David-Else:
Nice one!

Focus mode sounds great! Is it possible to have a 'zen' focus mode where the current window is centred with empty space either side, like https://github.com/folke/zen-mode.nvim ? Ideally the width would be 80 or maybe 100 characters. This is perfect for coding and writing while looking straight on at the monitor.

Termina94:
The zen mode sounds very cool and is definitely worth having it's own issue in my opinion.

The `focus mode` here was more of a temporary expansion, to allow quick shortcut access to a good size buffer in a split without needing multiple inputs to slowly increase size. 

Maybe `focus` is the wrong word and something like `expand` would be better suited for what this is doing

helix-view/src/editor.rs
```diff
@@ -905,6 +905,26 @@ impl Editor {
...
+        self.tree.shrink_buffer_width();
+    }
+
+    pub fn grow_buffer_height(&mut self) {
+        self.tree.grow_buffer_height();
+    }
+
+    pub fn shrink_buffer_height(&mut self) {
+        self.tree.shrink_buffer_height();
+    }
```
archseer, line 922: Rather than a bunch of methods, what about one method with some enums `resize_buffer(Resize::Shrink, Dimension::Height)`

helix-term/src/keymap/default.rs
```diff
@@ -26,6 +26,14 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "home" => goto_line_start,
         "end" => goto_line_end,
 
+        "A-w" => { "Alter Window"
+            "A-w"|"A-left" => shrink_buffer_width,
+            "A-l"|"A-right" => grow_buffer_width,
+            "A-j"|"A-down" => shrink_buffer_height,
+            "A-k"|"A-up" => grow_buffer_height,
+            "A-f" => buffer_expand_mode,
+        },
```
the-mikedavis, line 35: I think it would be handy to have also a sticky mode for this (see `Z`) so you don't have to hold alt the whole time

Termina94: Thanks for the tip, didn't realise that was a thing. It's now much nicer to use!

erasin:
@Termina94 
I'm a non-English speaking user and I try to be clear.

1. Less keymap memory.
Would it be better we modify KeyMap like this：

```
        "C-w" => { "Window"
            ....
             "A-h"|"A-left" => shrink_buffer_width,
            "A-l"|"A-right" => grow_buffer_width,
            "A-j"|"A-down" => shrink_buffer_height,
            "A-k"|"A-up" => grow_buffer_height,
            "A-f" => buffer_expand_mode,
        },

       "space" => {
        ....
        "W" => { "Alter Window" sticky=true
            "h"|"left" => shrink_buffer_width,
            "l"|"right" => grow_buffer_width,
            "j"|"down" => shrink_buffer_height,
            "k"|"up" => grow_buffer_height,
            "f" => buffer_expand_mode,
        },
        ....
        }
```


2. config
we need configure the size of each resize buffer.

3. resize direction
We should care about the changes in the current window. The 'h/left' should be the boundary moving to the left and the 'l/right' to the right. This does not seem to be the case at the moment.
When I'm in the second window, I press `H` and it goes right, it should go left, and it's the right in the leftmost window.

dariooddenino:
I've been trying this branch today. A few notes:

1. It's not possible to shrink a buffer all the way down by increasing the size of another one. For example with `| A | B |` I can increase `B` width only up to a certain point, and after that I have to move to `A` and shrink it.
2. I think a `reset to equal width` command would be useful
3. It would be great to be able to get into a "resize mode" where one can just old `h/j/k/l` to resize a window without having to repeat the whole `A-w A-h/j/k/l` command multiple times.
4. Minor thing, but I'm really not a fan of the alt-key combination for commands

helix-view/src/editor.rs
```diff
@@ -1069,6 +1069,14 @@ impl Editor {
         self.tree.transpose();
     }
 
+    pub fn resize_buffer(&mut self, resize_type: Resize, dimension: Dimension) {
+        self.tree.resize_buffer(resize_type, dimension);
+    }
+
+    pub fn buffer_expand_mode(&mut self) {
+        self.tree.buffer_expand_mode();
+    }
```
the-mikedavis, line 1078: I think the name and description for this command should say `toggle`. I'm also a little hesitant to call it a "mode" since that would skew the terminology for editor modes.

Maybe something like `toggle_focus_window`?

helix-view/src/tree.rs
```diff
@@ -73,8 +91,80 @@ impl Container {
...
             area: Rect::default(),
+            node_bounds: Vec::new(),
         }
     }
+
+    fn get_child_by_view_id(&mut self, node: ViewId) -> Option<&mut ContainerBounds> {
+        if let Some(index) = self.children.iter().position(|child| child == &node) {
+            return self.node_bounds.get_mut(index);
+        };
+        None
```
the-mikedavis, line 102:
```suggestion
        self.children
            .iter()
            .position(|child| child == &node)
            .and_then(|index| self.node_bounds.get_mut(index))
```

helix-view/src/tree.rs
```diff
@@ -73,8 +91,80 @@ impl Container {
...
+    fn remove_child_bounds(&mut self, index: usize) -> &mut Self {
+        self.node_bounds.remove(index);
+        self
+    }
+
+    fn calculate_slots_width(&self) -> usize {
+        self.node_bounds
+            .iter()
+            .map(|bounds| match bounds.expand {
+                true => 40,
```
the-mikedavis, line 153: 40 is hard-coded in a few places. Can we make it a pub const that gets shared between the modules that use it? It could make sense to have this be configurable in the future for those that like resizing to be more fine-grained (or not).

helix-term/src/keymap/default.rs
```diff
@@ -26,6 +26,22 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "home" => goto_line_start,
         "end" => goto_line_end,
 
+        "A-w" => { "Alter Window"
+            "A-h"|"A-left" => shrink_buffer_width,
+            "A-l"|"A-right" => grow_buffer_width,
+            "A-j"|"A-down" => shrink_buffer_height,
+            "A-k"|"A-up" => grow_buffer_height,
+            "A-f" => buffer_expand_mode,
+        },
```
the-mikedavis, line 35: Like the `<C-w>` bindings I think there should be `<A-w><A-h>` and also `<A-w>h`. That way you don't need to keep holding alt after you've hit `<A-w>`.

Jikstra:
I would love to see this merged, this is the only thing blocking me from using helix <3

dgkf:
@Termina94 this has been idle for quite a while. Are you still planning to come back to it? If not, I think it's an important capability for any split-based text editor and would be willing to pick it up.

@pascalkuthe @the-mikedavis just tagging for visibility to get your input on what all would need to be included to get this across the finish line.

claudemuller:
@Termina94 @the-mikedavis @dgkf Great work on this - I've switched over to Helix as a long time Vim user and am really enjoying it.
I'm not sure what MR etiquette is here as I haven't really contributed to OSS yet, but I don't mind picking this up. I'm still a Rust n00b but have at least gotten the above pieces of feedback implemented (I thike 😅) and built locally.

the-mikedavis:
See #8546, let's close this in favor of that new PR

PR #2707, opened 2022-06-08
Add file picker dialogue when opening a directory with :o

### LINKED ISSUE #2558: Feature request: Open file picker at typed directory
Sometimes I need a quick search outside of the current project.
A command like this is super helpful:
```
:file-picker ../other-project
```
And if it happens too often, I will:
```toml
[keys.normal]
"space" = { z = ":file-picker ~/projects/rust/library/std" }
```

weakphish:
This pull requests closes #2558. 

The behavior of `:o` is modified such that when opening a directory, the file picker pane is opened on the chosen directory rather than returning an error. 

Jun-08-2022 02-36-03


helix-term/src/commands/typed.rs
```diff
@@ -46,12 +46,24 @@ fn open(
...
-        let (view, doc) = current!(cx.editor);
-        let pos = Selection::point(pos_at_coords(doc.text().slice(..), pos, true));
-        doc.set_selection(view.id, pos);
-        // does not affect opening a buffer without pos
-        align_view(doc, view, Align::Center);
+        if std::fs::canonicalize(&path)?.is_dir() {
+            let callback = async move {
+                let call: job::Callback =
+                    Box::new(move |editor: &mut Editor, compositor: &mut Compositor| {
+                        let picker = ui::file_picker(path, &*editor.config.load());
```
the-mikedavis, line 53:
```suggestion
                        let picker = ui::file_picker(path, &editor.config());
```

weakphish:
> The behavior with opening multiple directories might need a bit of tuning. `:open .git .github` pushes a picker for `.git` and then `.github`, so pickers end up in reverse order. It could also make sense to fail the command when more than one path is a directory, although I kinda like the multiple pickers. What do others think about the behavior?

I think if I can reverse the order of the picker appearance, it's solid. I agree that multiple pickers is a good thing to have, generally.

sudormrfbin:
I suppose it can be displayed in the command line area with `editor.set_status()`.

weakphish:
I don't know if there's a way to use `editor.set_status` due to the constraints on the function. The closest I can get is something like this:
Screen Shot 2022-07-31 at 19 28 18

which I am not sure how to go about resolving @sudormrfbin @zen3ger

the-mikedavis:
I'm not sure this is a good idea allocation-wise but `set_status` can take a `String` so you can use `path.to_string_lossy().to_string()`

the-mikedavis:
It looks like the `cx.editor.set_status` statuses overwrite each other when there are multiple directories in a row. Maybe we could have the file picker write a status message when it opens saying something like "Picking files from {dir}"?

weakphish:
@the-mikedavis If I am understanding you correctly, it seems that there is a lot of constraints on the `callback` that make this difficult. Did your suggestion mean implementing the status change as part of the callback, or is there a piece I'm missing?

Thanks for the help :)

the-mikedavis:
Hmm yeah I think it's not possible currently or easy to change to have one callback create the next like I'm thinking. I think it's probably fine to ignore `set_status`ing for the pickers for now and we can revisit it later / leave it as a TODO

dead10ck:
This seems fine to me after the `set_status` is removed. :+1:

weakphish:
I removed the `editor.set_status` call.

dead10ck (approved):
Ok this looks good to me

the-mikedavis (approved):

PR #2710, opened 2022-06-08
elixirLS disable dialyzer by default

connorlay:
Not all Elixir projects use dialyzer and it can cause the editor to slow down quite a bit on large projects if the PLT is not built.

See https://github.com/elixir-lsp/elixir-ls#dialyzer-integration=

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -88,6 +88,7 @@ shebangs = ["elixir"]
 roots = []
 comment-token = "#"
 language-server = { command = "elixir-ls" }
+config = { elixirLS.dialyzerEnabled = false }
 indent = { tab-width = 2, unit = "  " }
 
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
I think this is reasonable 👍

I don't think the dialyzer diagnostics have been that helpful in elixir anyways and it takes forever to make that initial plt file.

connorlay:
> I wouldn't expect helix to slow down if a language server is busy though unless it was competing for resources. Does it get choppy on large projects?

I haven't profiled Helix to know for sure, but anecdotally I think it is probably the later that is causing the issue? This is on a fairly large monorepo Elixir project.

Disabling dialyzer seems to fix things for me locally 🤷🏻

PR #1723, opened 2022-03-01
Actually split `mi` and `ma` into several subcommands

The0x539:
Well, this was [poorly timed](https://github.com/helix-editor/helix/pull/1686).

Adds a `fallback` field to `KeyTrieNode`, used for surround pairs as discussed. For that command to actually *work*, I added a `last_key` field to `Context`.

Outstanding concerns:
- Should `last_key` be on `Editor` instead?
- Does a better approach come to mind?
- How can `select_*_pair` be used from the command palette?
- How should fallbacks be represented in `config.toml`?

PR #2715, opened 2022-06-08
Allow a custom delay before a command popup shows up

### LINKED ISSUE #1475: Allow a custom delay before a command popup shows up
For some commands Helix shows you a helpful popup list of commands to pick from (e.g. the "Space mode" when you press <kbd>Space</kbd>), however, sometimes this popup help isn't needed if I remember the command just fine, so the flash is just a bit visually annoying

I'd like to be able to add a __custom delay__[^1] before a popup is shown so that if I plan to execute a very frequent command I remember I'd never see a popup since I'll just hit the shortcut right away, but if I need to execute a more rare command I don't remember, I'd just wait an extra __0.2 sec__ (or something) and then pick the command visually from the list

[^1]: ideally different for each command, but maybe a customizable global one is just fine

lenstr:
Hi!

This pull request closes #1475.


EpocSquadron:
There's an interesting edge case with regards to the `m` key -- the intiial `m` respects the delay, but selection of `i` or `a` which brings up another menu does not. I think this is because the auto-info for inside/around is implemented separately from the normal auto-info for all keybinds. There's been talk of reworking that so the autoinfo for inside and around do use the same method of deriving from keybinds, so maybe we could leave it until then as it will resolve itself at that point. Up to you if you want to spend the time to fix it now.

EpocSquadron:
I wonder if it wouldn't be unjust to make a default value of 600 -- in my testing it feels like about the delay i would have when unsure of the keybind i want, while not showing up for times when im a little slower to press the next binding when i do know what it is.

the-mikedavis:
There's a PR reworking `mi`/`ma`: https://github.com/helix-editor/helix/pull/1723 that may help with that, although it's been quiet for a bit and could probably be superseded.

the-mikedavis (commented):
Places where we call `cx.on_next_key` don't get the effect of the delay, like `select_textobject` in `commands.rs`. We may want to introduce a `cx.on_next_key_with_autoinfo` like Kakoune has so we can close over adding fake autoinfo with delay nicely https://github.com/mawww/kakoune/blob/689553c2e9b953a9d3822528d4ad858af95fb6a2/src/input_handler.hh#L190

cd-a:
Is this PR abandoned? I love this feature

pse403:
> 

Agreed, I'd love it too.  I've become addicted to using this editor. I wish my Rust was more advanced so I could contribute. Meantime I can only echo that this feature would solve the one thing that constantly annoys me.

arexon:
I would really love to see this merged soon!

pascalkuthe:
closing this one out as stale. Thank you for contributing!

Issue #2679, opened 2022-06-05
3J doesn't join all 3 lines as tutor says


DuckDuckWhale:
### Summary

`3J` works the same as `J`.

### Reproduction Steps

I tried this:

1. `hx --tutor`
2. Press `627gg3J`

I expected this to happen:

Either the tutor don't say to use `3J` (because it's not selection & action any more) or all 3 lines are joined.

Instead, this happened:

The tutor says to use `3J` and only one line is joined.

### Platform

Ubuntu 22.04

### Terminal Emulator

GNOME Terminal 3.44.0 for GNOME 42

### Helix Version

22.05-33-g0035c29f

the-mikedavis:
I think this should just be fixed in the tutor. The current behavior of `J` with a count matches kakoune's `A-j`, plus `3xJ` is fits the select-then-act paradigm better than `3J`

Would you like to submit a PR to update the tutor? https://github.com/helix-editor/helix/blob/0035c29fc16f6c795257e241fddd9f2178bc4b4c/runtime/tutor.txt#L624

DuckDuckWhale:
Seem to be fixed in #2716

PR #2590, opened 2022-05-28
Cleanup for runtime/tutor.txt

kappq:
I have noticed some inconsistencies in the [tutor.txt](../blob/master/runtime/tutor.txt) file.

For example, there are no fixed amount of lines between each section (I opted for 3).

Also, in every numbered list, each element is never separated by a new line, except for the *COUNTS WITH MOTIONS* section:
https://github.com/helix-editor/helix/blob/efae76160d19ebba2bba31ecd5c104374c7b4483/runtime/tutor.txt#L243-L259

PR #2716, opened 2022-06-08
update tutor

n0s4:
There have been a couple of inconsistencies pointed out in #2679 & #2590 (the latter seems to have been quiet for a few days but we could still do each separately).

I also added a few more example "solutions" for the demos, some of which aren't exactly necessary since they are pretty obvious but some already have them so it can't hurt to be consistent.

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -621,7 +621,7 @@ _________________________________________________________________
  Type J to join together lines in selection.
 
  1. Move the cursor to the line below marked -->.
- 2. Type x four times or 4x to join the sentence to one line.
+ 2. Type x four times or 4x to select all four lines.
  3. Type J to join the lines together.
 
  --> This sentence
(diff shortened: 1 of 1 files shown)


n0s4:
I've noticed also that "Insert mode" and "Normal mode" are capitalized but "command mode" isn't. I think I'll fix that here as well.

runtime/tutor.txt
```diff
@@ -621,21 +621,21 @@ _________________________________________________________________
  Type J to join together lines in selection.
 
  1. Move the cursor to the line below marked -->.
- 2. Type J thrice or 3J to join the sentence to one line.
+ 2. Type x four times or 4x to join the sentence to one line.
```
the-mikedavis, line 624:
```suggestion
 2. Type x four times or 4x to select all four lines.
```

the-mikedavis (approved):
I like the solution lines, I think it's good to have an idea where you're supposed to end up 👍

n0s4:
Alright, I've resolved them now @the-mikedavis.

PR #2718, opened 2022-06-08
Better handling of symlinks

### LINKED ISSUE #1548: `:change-current-directory` does not autocomplete for symlinks to directories
### Reproduction steps

- Open `helix`
- Use `:change-current-directory` and see that completions do not show for symlinks to directories. (If you manually type the symlink (ex. `Docs` in the screencast below), Helix will change the directory correctly)

[asciicast](https://asciinema.org/a/VZUFuDVMjGxDEcnoPYc6BKE55)

### Environment

- Platform: Linux
- Terminal emulator: Kitty 
- Helix version: `helix 0.5.0`

EMPTY
</details>

### LINKED ISSUE #2246: Panic when opening a symlink
### Summary

I have a symlink to the `runtime/` in my `~/.config/helix` . I wanted to open a theme in `runtime/themes` inside Helix so I opened the file picker. The symlink showed up but not any files within the actual folder. When I put the pointer on the runtime symlink the preview box says `<File not found>`. When I press enter on the symlink Helix panics:
```
thread 'main' panicked at 'editor.open failed: Is a directory (os error 21)', helix-term/src/ui/mod.rs:185:18
```

This doesn't seem very mysterious. I guess following symlinks just isn't implemented yet in the file picker, Helix shouldn't crash because of it though.

### Reproduction Steps

 

I tried this:

1. Open helix in the config folder.
2. Open the file picker and select the runtime symlink

I expected this to happen:

Helix would show the files in the folder the symlink points to.

Instead, this happened:

Helix panics.

### Helix log

These are the latest 10 lines of `helix.log`. Nothing looks relevant.

```
2022-04-24T11:49:36.554 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:49:36.562 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { LSP not defined }
2022-04-24T11:50:09.881 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:50:09.881 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:50:09.905 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { LSP not defined }
2022-04-24T11:50:49.229 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:50:49.230 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:50:49.252 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { LSP not defined }
2022-04-24T11:58:26.687 helix_view::theme [WARN] Theme: invalid style attribute: modifers
2022-04-24T11:58:26.687 helix_view::theme [WARN] Theme: invalid style attribute: modifers

```

</details>

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

22.03-127-g2c7f770a

zen3ger:
- Add file-picker.follow-symlinks configuration option (default is true), this
  also controls if filename and directory completers follow symlinks.

- Update FilePicker to set editor error if opening a file fails, instead of
  panicing.

Fix #1548
Fix  #2246

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -118,6 +118,7 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
         .hidden(config.file_picker.hidden)
         .parents(config.file_picker.parents)
         .ignore(config.file_picker.ignore)
+        .follow_links(config.file_picker.follow_symlinks)
         .git_ignore(config.file_picker.git_ignore)
         .git_global(config.file_picker.git_global)
         .git_exclude(config.file_picker.git_exclude)
@@ -146,14 +147,13 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
         let entry = entry.ok()?;
 
         // This is faster than entry.path().is_dir() since it uses cached fs::Metadata fetched by ignore/walkdir
-        let is_dir = entry.file_type().map(|ft| ft.is_dir()).unwrap_or(false);
-
+        let is_dir = entry.file_type().map_or(false, |ft| ft.is_dir());
         if is_dir {
             // Will give a false positive if metadata cannot be read (eg. permission error)
-            return None;
+            None
+        } else {
+            Some(entry.into_path())
         }
-
-        Some(entry.into_path())
     });
 
     // Cap the number of files if we aren't in a git project, preventing
@@ -175,9 +175,14 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
             path.strip_prefix(&root).unwrap_or(path).to_string_lossy()
         },
         move |cx, path: &PathBuf, action| {
-            cx.editor
-                .open(path.into(), action)
-                .expect("editor.open failed");
+            if let Err(e) = cx.editor.open(path.into(), action) {
+                let err = if let Some(err) = e.source() {
+                    format!("{}", err)
+                } else {
+                    format!("unable to open \"{}\"", path.display())
+                };
+                cx.editor.set_error(err);
+            }
         },
         |_editor, path| Some((path.clone(), None)),
     )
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/mod.rs
```diff
@@ -145,15 +145,27 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
...
             // Will give a false positive if metadata cannot be read (eg. permission error)
             return None;
         }
 
-        Some(entry.into_path())
+        // Only list symlink if points to a file which we have permisson for
+        match std::fs::metadata(entry.path()).map(|meta| meta.is_file()) {
+            Err(_) | Ok(false) => None,
+            Ok(true) => Some(entry.into_path()),
+        }
```
zen3ger, line 168: maybe just setting `follow_links(true)` on the builder in this case would be better, would fix #2246 as well...

archseer: I think so, `fs::metadata` is way too costly to run per-entry (out of potentially thousands). I specifically removed code calling `fs::metadata` from the completion functions https://github.com/helix-editor/helix/commit/78fba8683bdce31ad2d8e710cb70fb9f05b21e8c#diff-a2f72be0dae9186774ac716ae568f5216a5223dbf91e6ac8a25f62617c4fa55cL144

Also, wouldn't this make more sense to do inside `editor.open`? Resolve the symlink, then either `set_error` "Not allowed to open the file" or open the file. I don't think files with insufficient permissions should just get hidden, it makes it seem like they're not there at all

zen3ger: I've removed `fs::metadata` calls and changed to use `follow_symlinks()` instead. Added a configuration option to `FilePicker` to enable/disable following symlinks, so if it would lead to performance issues it can at least be toggled. Also error is now reported the same way like in `:cd` and `:open` if file picker fails. Hope I covers everything you mentioned.

PR #2728, opened 2022-06-09
Add single width left margin for completion popup

sudormrfbin:
Adds a bit of padding to the left edge of the completion popup:

| Before | After |
| --- | --- |
| before | after |

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -48,6 +48,8 @@ pub struct Menu<T: Item> {
 }
 
 impl<T: Item> Menu<T> {
+    const LEFT_PADDING: usize = 1;
+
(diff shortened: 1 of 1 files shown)


sudormrfbin:
The scroll bar was actually next on my list, along with adding a scroll bar for the doc popup :)

the-mikedavis:
Hmm strangely I have a separate style for that left margin. I bet my theme is missing a scope 🤔 

before

sudormrfbin:
Fixed in 23bc196d; was supposed to clear the area with `ui.menu`.

the-mikedavis (approved):
Looks good, I like the extra space 🎨

archseer:
I'm not sure how I feel about the extra space but let's try it in master 👍🏻

PR #2731, opened 2022-06-09
Introduce storage_class highlight scope

theli-ua:
Inspired by VIM's StorageClass hi group

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -151,6 +151,7 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
+  - `storage` - Keywords that affect the storage of a variable, function or data structure `static`, `mut`, `const`, `ref`
 
 - `operator` - `||`, `+=`, `>`
 
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -156,16 +156,22 @@
   "macro_rules!"
 
   "let"
-  "ref"
-  "move"
 
-  "dyn"
-  "static"
-  "const"
   "async"
 ] @keyword
 
-(mutable_specifier) @keyword.mut
+(mutable_specifier) @keyword.storage.modifier.mut
+
+(reference_type "&" @keyword.storage.modifier.ref)
+(self_parameter "&" @keyword.storage.modifier.ref)
+
+[
+  "static"
+  "const"
+  "ref"
+  "move"
+  "dyn"
+] @keyword.storage.modifier
 
 ; TODO: variable.mut to highlight mutable identifiers via locals.scm
 
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -1,8 +1,9 @@
+(storage_class_specifier) @keyword.storage
+
 "goto" @keyword
 "register" @keyword
 "break" @keyword
 "case" @keyword
-"const" @keyword
 "continue" @keyword
 "default" @keyword
 "do" @keyword
@@ -14,7 +15,6 @@
 "inline" @keyword
 "return" @keyword
 "sizeof" @keyword
-"static" @keyword
 "struct" @keyword
 "switch" @keyword
 "typedef" @keyword


runtime/queries/rust/highlights.scm
```diff
@@ -156,16 +156,19 @@
...
 
-  "dyn"
-  "static"
-  "const"
   "async"
 ] @keyword
 
-(mutable_specifier) @keyword.mut
+[
+  ("&" (identifier))
```
the-mikedavis, line 164: Checking this out locally, I think I prefer that `&` keeps its `operator` scope when used to take a reference, like in:

```rust
foo(&bar)
```

but I think it could make sense to highlight `&` as storage-class when used as a parameter/type specification:

```rust
pub struct Context<'a> {
    pub jobs: &'a mut Jobs,
}

pub fn foo(&mut self) {}
```

```scm
(reference_type "&" @keyword.storage-class)
(self_parameter "&" @keyword.storage-class)

[
  (mutable_specifier)
  "static"
  "const"
  "ref"
  "move"
  "dyn"
] @keyword.storage-class
```

What do you think?

the-mikedavis (commented):
I think I would prefer this use a hyphen: `@keyword.storage-class`. It shouldn't be a problem to use hyphens in the query files since they're lisp or in the theme toml

the-mikedavis (approved):

archseer:
Why not `@keyword.storage`?

runtime/queries/rust/highlights.scm
```diff
@@ -156,16 +156,21 @@
...
-  "ref"
-  "move"
 
-  "dyn"
-  "static"
-  "const"
   "async"
 ] @keyword
 
-(mutable_specifier) @keyword.mut
```
archseer, line 168: You removed an existing scope that some themes might target. This should be `@keyword.storage.modifier.mut`

runtime/queries/rust/highlights.scm
```diff
@@ -156,16 +156,21 @@
...
+(self_parameter "&" @keyword.storage)
+
+[
+  (mutable_specifier)
+  "static"
+  "const"
+  "ref"
+  "move"
+  "dyn"
+] @keyword.storage
```
archseer, line 173: `@keyword.storage.modifier` as per https://www.sublimetext.com/docs/scope_naming.html#storage (we just namespace under keyword)

runtime/queries/rust/highlights.scm
```diff
@@ -156,16 +156,21 @@
...
 
-  "dyn"
-  "static"
-  "const"
   "async"
 ] @keyword
 
-(mutable_specifier) @keyword.mut
+(reference_type "&" @keyword.storage)
+(self_parameter "&" @keyword.storage)
```
archseer, line 164: `@keyword.storage.modifier.ref`

book/src/themes.md
```diff
@@ -151,6 +151,7 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
+  - `storage` - `static`, `mut`, `const`, `ref`
```
archseer, line 154:
```suggestion
  - `storage` - Keywords that affect the storage of a variable, function or data structure `static`, `mut`, `const`, `ref`
```

runtime/queries/rust/highlights.scm
theli-ua: This is undocumented, not used in any builtin themes and I couldn't find it being used in the wild either. I'm still sure it should be removed

theli-ua: I guess showing it under storage.modifier.mut is fine

runtime/queries/rust/highlights.scm
```diff
@@ -156,16 +156,23 @@
...
 ] @keyword
 
-(mutable_specifier) @keyword.mut
+(mutable_specifier) @keyword.storage.modifier.mut
+
+(reference_type "&" @keyword.storage.modifier.ref)
+(self_parameter "&" @keyword.storage.modifier.ref)
+
+[
+  (mutable_specifier)
```
kirawi, line 169: This is redundant since there's already a capture for this scope.

theli-ua: Oops, you're right. I returned mutable as an additional scope but forgot to remove it from `storage.modifier`

the-mikedavis (approved):

PR #2733, opened 2022-06-10
Add `unbind-default-keys` config option

### LINKED ISSUE #2720: Unbind all default keybindings config option

Allow all default keybindings to be set to `no-op` with a config option. While I'm very happy with Helix's design of giving commands and bindings descriptive names instead of using arcane keyboard keys (unlike vim and it's derivatives), it's currently both cumbersome and error prone to manually set every unused binding to `no-op` when remapping keys. If a precedent is needed, Zellij has a `keybindings: unbind: <bool>` config that does exactly this.

To be more specific, I propose the following configuration: `unbind-defaults: <bool>` which defaults to `false`. When set to `true` all default keybindings are not added to the keybinding trie (or set to `no-op`, whatever's easier) and only the bindings listed in the user's config file will be used.

Lindenk:
Closes #2720.

There were a few changes I made to reduce confusion with some function/method locations and naming:

- `helix_term::keymap::merge_keys` was moved to `helix_term::config::Config::merge_keys` as a method because it only operates on and changes `Config`, and was modified to take a set of bindings to merge with itself. New functions were also added to handle the new optional behavior.
- `Config::load` now is the single source of truth for loading and merging configs, which was previously also handled in `helix_term::application::new` in a slightly different but mostly redundant manner.
- `Default` is removed from `Config` and replaced with `Config::new` and `Config::with_default_keys`. `new` returns a Config with no bindings while `with_default_keys` returns a Config with the default bindings (which is the same behavior as the old `default`). This to prevent confusion between a default rust `struct` and helix's default keybindings as these terms were previously conflated.
- `Config::keys` no longer loads default bindings through serde when no keys are found as this was always redundant.
- `helix_term::keymap::Keymaps::get` no longer crashes when a mode isn't found. This will also help with future mode additions (likely through plugins down the line).

Let me know if any of these changes should be handled differently

EDIT: After updating to a much newer version of helix, this PR only adds the new option `unbind_default_keys` which does not populate the initial keymap at all. This also includes the fix where Helix would crash if a keybind for a non-existent mode is triggered

### DIFF
--- a/book/src/remapping.md
+++ b/book/src/remapping.md
@@ -75,5 +75,23 @@ Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes
 
 Keys can be disabled by binding them to the `no_op` command.
 
+To remove all default bindings, `unbind-default-keys = true` can be added to the top level configuration.
+
+```toml
+unbind-default-keys = true
+
+# Only these normal mode bindings will be used
+[keys.normal]
+n = "normal_mode"
+t = "goto_definition"
+
+# remember to add bindings to return to normal mode
+[keys.select]
+esc = "normal_mode"
+
+[keys.insert]
+esc = "normal_mode"
+```
+
 A list of commands is available in the [Keymap](https://docs.helix-editor.com/keymap.html) documentation
  and in the source code at [`helix-term/src/commands.rs`](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs) at the invocation of `static_commands!` macro and the `TypableCommandList`.
--- a/helix-term/src/config.rs
+++ b/helix-term/src/config.rs
@@ -20,6 +20,8 @@ pub struct Config {
 #[serde(deny_unknown_fields)]
 pub struct ConfigRaw {
     pub theme: Option<String>,
+    #[serde(default)]
+    pub unbind_default_keys: bool,
     pub keys: Option<HashMap<Mode, Keymap>>,
     pub editor: Option<toml::Value>,
 }
@@ -66,7 +68,11 @@ impl Config {
             local.and_then(|file| toml::from_str(&file).map_err(ConfigLoadError::BadConfig));
         let res = match (global_config, local_config) {
             (Ok(global), Ok(local)) => {
-                let mut keys = keymap::default();
+                let mut keys;
+                match local.unbind_default_keys {
+                    true => keys = HashMap::default(),
+                    false => keys = keymap::default(),
+                }
                 if let Some(global_keys) = global.keys {
                     merge_keys(&mut keys, global_keys)
                 }
@@ -96,7 +102,11 @@ impl Config {
                 return Err(ConfigLoadError::BadConfig(err))
             }
             (Ok(config), Err(_)) | (Err(_), Ok(config)) => {
-                let mut keys = keymap::default();
+                let mut keys;
+                match config.unbind_default_keys {
+                    true => keys = HashMap::default(),
+                    false => keys = keymap::default(),
+                }
                 if let Some(keymap) = config.keys {
                     merge_keys(&mut keys, keymap);
                 }
--- a/helix-term/src/keymap.rs
+++ b/helix-term/src/keymap.rs
@@ -360,7 +360,10 @@ impl Keymaps {
     pub fn get(&mut self, mode: Mode, key: KeyEvent) -> KeymapResult {
         // TODO: remove the sticky part and look up manually
         let keymaps = &*self.map();
-        let keymap = &keymaps[&mode];
+        let keymap = match keymaps.get(&mode) {
+            Some(keymap) => keymap,
+            None => return KeymapResult::NotFound,
+        };
 
         if key!(Esc) == key {
             if !self.state.is_empty() {
@@ -415,10 +418,10 @@ impl Default for Keymaps {
     }
 }
 
-/// Merge default config keys with user overwritten keys for custom user config.
+/// Merge existing config keys with user overwritten keys.
 pub fn merge_keys(dst: &mut HashMap<Mode, Keymap>, mut delta: HashMap<Mode, Keymap>) {
-    for (mode, keys) in dst {
-        keys.merge(delta.remove(mode).unwrap_or_default())
+    for (mode, keys) in delta.drain() {
+        dst.entry(mode).or_default().merge(keys);
     }
 }
 


the-mikedavis:
Could you add some notes on using this to the docs? https://github.com/helix-editor/helix/tree/f37ffaa3a1c7754238029390db00148ec8105874/book/src

Lindenk:
No problem. If you'd like me to include a new subsection or larger blurb, or if there's a better place to provide this specific documentation, let me know

helix-term/src/config.rs
```diff
@@ -11,22 +11,14 @@ use toml::de::Error as TomlError;
 #[serde(deny_unknown_fields)]
```
sudormrfbin, line 11: We use kebab case in the config:

```suggestion
#[serde(deny_unknown_fields, rename_all = "kebab-case")]
```

book/src/remapping.md
```diff
@@ -48,7 +49,10 @@ Control, Shift and Alt modifiers are encoded respectively with the prefixes
 | Null         | `"null"`       |
 | Escape       | `"esc"`        |
 
-Keys can be disabled by binding them to the `no_op` command.
+Individual keys can be disabled by binding them to the `no_op` command.
+
+To remove the default bindings for all modes defined by the user, `unbind_defaults = true` can be added to the top level configuration.
```
sudormrfbin, line 54:
```suggestion
To remove the default bindings for all modes defined by the user, `unbind-defaults = true` can be added to the top level configuration.
```

sudormrfbin (commented):
How about renaming the config option to `unbind_default_keys` ?

helix-term/src/config.rs
```diff
@@ -43,11 +35,68 @@ impl Display for ConfigLoadError {
...
+
     pub fn load(config_path: PathBuf) -> Result<Config, ConfigLoadError> {
         match std::fs::read_to_string(config_path) {
-            Ok(config) => toml::from_str(&config)
-                .map(merge_keys)
-                .map_err(ConfigLoadError::BadConfig),
+            Ok(config) => match toml::from_str::<Config>(&config) {
+                Ok(config) if config.unbind_default_keys => {
+                    Ok(config.replace_into_keys(Config::with_default_keys()))
+                }
```
sudormrfbin, line 95: This unbinds the default keys for only those modes for which there are user defined keys in the config, i.e. with this config:

```toml
unbind-default-keys = true

[keys.normal]
"t" = "goto_definition"
```

normal mode will only have the `t` keybind while all other modes will have the default bindings. This is mentioned in the docs with 

> To remove the default bindings **for all modes defined by the user**,..._

but I think it can be made clearer. A comment in the code would also be nice.

Lindenk: I can do that. Initially the plan was to have it unbind all defaults of every mode but after trying it, it wasn't intuitive. I'll add some comments and a small example for the docs

helix-term/src/application.rs
```diff
@@ -89,6 +89,10 @@ impl Application {
         use helix_view::editor::Action;
 
         let config_dir = helix_loader::config_dir();
+        if !config_dir.exists() {
+            std::fs::create_dir_all(&config_dir).ok();
+        }
+
```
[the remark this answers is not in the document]
Lindenk: Removed the check with [119c750](https://github.com/helix-editor/helix/pull/2733/commits/119c7507144517ea62c87acc469b91c0136f9708)

helix-term/src/config.rs
Lindenk: Improved comments and docs with [119c750](https://github.com/helix-editor/helix/pull/2733/commits/119c7507144517ea62c87acc469b91c0136f9708)

helix-term/src/config.rs
```diff
@@ -43,11 +35,68 @@ impl Display for ConfigLoadError {
...
 
 impl Config {
+    // avoids confusion between a default rust struct and default helix keys
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        Config {
+            theme: None,
+            unbind_default_keys: false,
+            keys: Default::default(),
+            editor: Default::default(),
```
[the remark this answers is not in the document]
Lindenk: Changed to the above explicit naming with [119c750](https://github.com/helix-editor/helix/pull/2733/commits/119c7507144517ea62c87acc469b91c0136f9708)

helix-term/src/config.rs
```diff
@@ -43,11 +35,68 @@ impl Display for ConfigLoadError {
...
+    pub fn replace_keys(mut self, other: Self) -> Self {
+        for (mode, keys) in other.keys.into_iter() {
+            self.keys.insert(mode, keys);
+        }
+
+        self
+    }
+
+    /// Take other's keys and replace our modes into them
+    pub fn replace_into_keys(mut self, mut other: Self) -> Self {
```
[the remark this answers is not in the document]
Lindenk: Renamed function and calls to it with [119c750](https://github.com/helix-editor/helix/pull/2733/commits/119c7507144517ea62c87acc469b91c0136f9708)

helix-term/src/config.rs
```diff
@@ -43,11 +35,71 @@ impl Display for ConfigLoadError {
...
+    /// Creates a new empty config
+    #[allow(clippy::new_without_default)]
+    pub fn new() -> Self {
+        Config {
+            theme: None,
+            unbind_default_keys: false,
+            keys: HashMap::new(),
+            editor: helix_view::editor::Config::default(),
+        }
+    }
```
the-mikedavis, line 48: The original Default implementation still makes sense: it uses the default keys. I think this should be renamed as `Config::empty` or `Config::without_default_keys`

helix-term/src/keymap.rs
```diff
@@ -361,7 +362,14 @@ impl Keymaps {
     pub fn get(&mut self, mode: Mode, key: KeyEvent) -> KeymapResult {
         // TODO: remove the sticky part and look up manually
         let keymaps = &*self.map();
-        let keymap = &keymaps[&mode];
+
+        // Ideally this would be:
+        // let keymap = keymaps.get(&mode).ok_or(KeymapResult::NotFound)?;
```
the-mikedavis, line 367: I think this comment can be removed - I'm not sure it makes sense to refactor `KeymapResult` into a `Result`

helix-term/src/keymap.rs
```diff
@@ -2,6 +2,7 @@ pub mod default;
 pub mod macros;
 
 pub use crate::commands::MappableCommand;
+#[allow(unused_imports)] // `Config` is used in a macro
 use crate::config::Config;
```
the-mikedavis, line 6: It ends up being used in the `tests` module below so you can remove the `use` here and add it to the `tests` module

helix-term/src/config.rs
```diff
@@ -43,11 +35,71 @@ impl Display for ConfigLoadError {
...
+            Ok(config) => match toml::from_str::<Config>(&config) {
+                Ok(config) if config.unbind_default_keys => {
+                    Ok(config.replace_into_key_modes(Config::with_default_keys()))
+                }
+                Ok(config) => Ok(config.merge_into_keys(Config::with_default_keys())),
+
+                Err(err) => Err(ConfigLoadError::BadConfig(err)),
+            },
             Err(err) => Err(ConfigLoadError::Error(err)),
         }
```
the-mikedavis, line 104: This can be refactored to reduce the nesting:

```rs
let config_string =
    std::fs::read_to_string(config_path).map_err(|err| ConfigLoadError::Error(err))?;
let config = toml::from_str::<Config>(&config_string)
    .map_err(|err| ConfigLoadError::BadConfig(err))?;

if config.unbind_default_keys {
    Ok(config.replace_into_key_modes(Config::with_default_keys()))
} else {
    Ok(config.merge_into_keys(Config::with_default_keys()))
}
```

book/src/remapping.md
```diff
@@ -48,7 +49,24 @@ Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes
...
+
+# Only these normal mode bindings will be used
+[keys.normal]
+n = "normal_mode"
+t = "goto_definition"
+
+# No select mode bindings will be used
+[keys.select]
+
+# All default insert mode bindings will be used because new bindings weren't defined
```
the-mikedavis, line 67: This behavior suprises me - I would expect `unbind-default-keys` to clear the keymap completely

helix-term/src/config.rs
Lindenk: I rewrote the feature to bring it up to the latest version of helix and no longer changed the config's `Default` implementation. The codebase has changed a bit so I don't think a `Config::empty` method is needed anymore

helix-term/src/keymap.rs
Lindenk: Noted, I removed the comment

helix-term/src/config.rs
Lindenk: These methods aren't needed anymore with the current structure of helix

book/src/remapping.md
Lindenk: I changed it to no longer include this behaviour. The main reason for this was so users might forget to bind `esc = "normal_mode"` or similar, but I added a blurb about it in the docs instead

Lindenk:
I updated the feature to the current version of helix and addressed the above changes. Now it only adds the config option `unbind_default_keys`

helix-term/src/config.rs
```diff
@@ -96,7 +102,11 @@ impl Config {
                 return Err(ConfigLoadError::BadConfig(err))
             }
             (Ok(config), Err(_)) | (Err(_), Ok(config)) => {
-                let mut keys = keymap::default();
+                let mut keys;
+                match config.unbind_default_keys {
+                    true => keys = HashMap::default(),
+                    false => keys = keymap::default(),
+                }
```
shaleh, line 109:
```suggestion
                let mut keys = match config.unbind_default_keys {
                    true => HashMap::default(),
                    false => keymap::default(),
                };
```

helix-term/src/config.rs
```diff
@@ -66,7 +68,11 @@ impl Config {
             local.and_then(|file| toml::from_str(&file).map_err(ConfigLoadError::BadConfig));
         let res = match (global_config, local_config) {
             (Ok(global), Ok(local)) => {
-                let mut keys = keymap::default();
+                let mut keys;
+                match local.unbind_default_keys {
+                    true => keys = HashMap::default(),
+                    false => keys = keymap::default(),
+                }
```
shaleh, line 75:
```suggestion
                let mut keys = match local.unbind_default_keys {
                    true => HashMap::default(),
                    false => keymap::default(),
                };
```

helix-term/src/keymap.rs
```diff
@@ -360,7 +360,10 @@ impl Keymaps {
     pub fn get(&mut self, mode: Mode, key: KeyEvent) -> KeymapResult {
         // TODO: remove the sticky part and look up manually
         let keymaps = &*self.map();
-        let keymap = &keymaps[&mode];
+        let keymap = match keymaps.get(&mode) {
+            Some(keymap) => keymap,
+            None => return KeymapResult::NotFound,
+        };
```
shaleh, line 366:
```suggestion
        let Some(keymap) = keymaps.get(&mode) else { return KeymapResult::NotFound; };
```
This is the style used elsewhere in the code.

the-mikedavis:
https://github.com/helix-editor/helix/pull/9384#issuecomment-2065534491, https://github.com/helix-editor/helix/issues/2720#issuecomment-2065533370 - we want to solve this when we switch config away from TOML rather than adding a new config option now

PR #2735, opened 2022-06-10
Clarified the text in chapter 3 of the tutor

### LINKED ISSUE #2725: Typo in the tutor?
### Summary

In the chapter 3 recap in the tutor it says `Type C to copy the selection on the line below and Alt-C for above.`

This does not sound correct to me. Which selection is it talking about? Would it be more correct to say `Type C to duplicate the cursor to the line below and Alt-C for above.`?

I would fix this myself, but just in case it actually isn't a typo I created this issue instead.

### Reproduction Steps

hx --tutor

### Helix log

_No response_

### Platform

All

### Terminal Emulator

All

### Helix Version

22.05-46-ge0532771

BearOve:
Fixes #2725

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -332,23 +332,25 @@ _________________________________________________________________
 =                     3.1 MULTIPLE CURSORS                      =
 =================================================================
 
- Type C to duplicate the cursor to the next line.
+ Type C to duplicate the cursor to the next suitable line.
 
  1. Move the cursor to the first line below marked -->.
- 2. Type C to duplicate the cursor to the next line. Keys you
-    type will now affect both cursors.
+ 2. Type C to duplicate the cursor to the next suitable line.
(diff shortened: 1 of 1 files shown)


n0s4:
You could also correct the phrasing on line 335 and in the instructions as they both refer to it copying "to the next line".

>Type C to duplicate the cursor *to the next line*.
>  1. Move the cursor to the first line below marked -->.
>  2. Type C to duplicate the cursor *to the next line*. Keys you
    type will now affect both cursors.
>  3. Use Insert mode to correct the lines. The two cursors will
    fix both lines simultaneously.
>  4. Type , to remove the second cursor

BearOve:
I was going to say: Yes, but isn't that correct when the cursor/selection is one character like here?

But then I tested it and it is not only the widt of the selection, but also the offset that needs to match. So even the cursor may skip lines.

I will revisit the text with this in mind later. Probably no time today

PR #2736, opened 2022-06-10
Update auto-pairs and idle-timeout when the config is reloaded

BearOve:
This fixes a problem where editor.auto-pairs is not reloaded with the rest of the configuration.

I also added idle-timeout, but that is likely not to matter due to it being reset between each key press.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -316,7 +316,12 @@ impl Application {
                     }),
             );
         }
+
         self.config.store(Arc::new(config));
+
+        // Update all the relevant members in the editor after replacing
+        // the configuration.
+        self.editor.refresh_config();
     }
 
     fn true_color(&self) -> bool {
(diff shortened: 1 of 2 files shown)


helix-term/src/application.rs
```diff
@@ -316,7 +316,12 @@ impl Application {
                     }),
             );
         }
+
         self.config.store(Arc::new(config));
+
+        // Update all the relevant members in the editor after replacing
+        // the configuration.
+        self.editor.refresh_config();
```
sudormrfbin, line 324: It might be better to call `editor.refresh_config()` in the `handle_config_events()` function since that will update the editor specific config on both `Refresh` and `Update` config events. The `Update` event is triggered on a runtime `:set` invocation and `Refresh` is triggered on `:config-reload`, so as of now auto-pairs is not configurable using `:set` (the PR fixes it for `:config-reload` only).

BearOve: Makes sense, I'll look into that. Thanks

sudormrfbin (approved):
Thank you for the fix!

PR #2738, opened 2022-06-10
feat: highlight / select symbol under cursor using LSP textDocument/documentHighlight

lazytanuki:
Hi !

This PR implements the LSP [textDocument/documentHighlight](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_documentHighlight) request. The command asks the LSP server for the ranges of references to current symbol, and select them (using multiple cursors).

I initially just wanted to have these symbols highlighted (like in nvim or vscode), but using multi-cursors is more convenient because the user can navigate between the references using `(` and `)`, and also because it reuses already existing components : 

Peek 2022-06-10 14-15

I have set the `Space + h` (for highlight) mapping to use this.

Have a good day !

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -6,7 +6,7 @@ use helix_lsp::{
 
 use super::{align_view, push_jump, Align, Context, Editor};
 
-use helix_core::Selection;
+use helix_core::{Selection, SmallVec};
 use helix_view::editor::Action;
 
 use crate::{
@@ -645,6 +645,7 @@ pub fn signature_help(cx: &mut Context) {
         },
     );
 }
+
 pub fn hover(cx: &mut Context) {
     let (view, doc) = current!(cx.editor);
     let language_server = language_server!(cx.editor, doc);
@@ -694,6 +695,7 @@ pub fn hover(cx: &mut Context) {
         },
     );
 }
+
 pub fn rename_symbol(cx: &mut Context) {
     ui::prompt(
         cx,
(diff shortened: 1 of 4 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -719,3 +721,50 @@ pub fn rename_symbol(cx: &mut Context) {
...
+
+    let pos = doc.position(view.id, offset_encoding);
+
+    let future = language_server.text_document_document_highlight(doc.identifier(), pos, None);
+
+    cx.callback(
+        future,
+        move |editor, _compositor, response: Option<Vec<lsp::DocumentHighlight>>| {
+            if let Some(document_highlights) = response {
+                if !document_highlights.is_empty() {
```
sudormrfbin, line 738: Using an early return here would unindent the whole block nicely:

```rust
            let document_highlights = match response {
                Some(highlights) if !highlights.is_empty() => highlights,
                _ => return,
            };
```

helix-term/src/commands/lsp.rs
```diff
@@ -719,3 +721,50 @@ pub fn rename_symbol(cx: &mut Context) {
...
+                        let range =
+                            helix_lsp::util::lsp_range_to_range(text, lsp_range, offset_encoding);
+                        if let Some(range) = range {
+                            if range.contains(pos) || range.head == pos {
+                                primary_index = i;
+                            }
+                            ranges.push(range);
+                            i += 1;
+                        }
+                    }
```
sudormrfbin, line 762:
```suggestion
            // We must find the range that contains our primary cursor to prevent our primary cursor to move
            let mut primary_index = 0;

            let ranges = document_highlights
                .iter()
                .filter_map(|highlight| lsp_range_to_range(text, highlight.range, offset_encoding))
                .enumerate()
                .map(|(i, range)| {
                    if range.contains(pos) {
                        primary_index = i;
                    }
                    range
                })
                .collect();

```

lazytanuki:
Changes made, thanks for the review !

sudormrfbin (approved):
LGTM!

n0s4:
As for the keybind, wouldn't `m-h` be better than `space-h`? This feels like a "match" command.

lazytanuki:
Now that you mention it, indeed it does look like a match command. I chose 'h' for highlight because the call to LSP is `textDocument/documentHighlight`, although maybe something else than 'h' would make more sense as well ?
What do others think ?

n0s4:
Maybe `m-S` for 'Symbol' ? Lowercase `s` is already used (although we could change that to `w` for 'wrap'..?).

lazytanuki:
I don't think changing `m-s` would be such a good idea, it might already be muscle memory for current users. `m-S` could do it.

the-mikedavis:
I didn't mind `space-h` but I don't have any problems with `mS` either. My only hold-up with `m` is that it's usually used for matches and textobjects and I don't think this fits in perfectly with that. Given that hover and rename are already in `space` I think adding more is ok

lazytanuki:
Okay, should we go back to `space-h` then ?

the-mikedavis:
Yeah let's go for `space-h` for now, we can always revisit the binding later

lazytanuki:
Alright, done !

the-mikedavis (approved):
This is awesome, thanks for working on it!

archseer (approved):
Great work!

PR #2745, opened 2022-06-11
Expand chapter one recap of tutor

jtepe:
This adds the missing variant of entering insert mode
to the chapter 1 recap section.

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -169,7 +169,11 @@ _________________________________________________________________
 
  * Type i to enter Insert mode and type text. Type <ESC> to
    return to Normal mode.
-
+   * Use a to enter Insert mode after the current selection
+     instead of before.
+   * Use I to enter Insert mode at the first non-whitespace
+     character at the start of a line.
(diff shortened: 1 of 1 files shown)


runtime/tutor.txt
```diff
@@ -169,7 +169,10 @@ _________________________________________________________________
 
  * Type i to enter Insert mode and type text. Type <ESC> to
    return to Normal mode.
-
+   * Use a to enter Insert mode after the current selection
+     instead of before.
+   * Use I or A to enter Insert mode at the start or end of
+     a line respectively.
```
the-mikedavis, line 175: <kbd>I</kbd> prepends to the line starting after any whitespace. I'm not sure it's worth mentioning that here. I suppose we could split out bullets on <kbd>I</kbd> and <kbd>A</kbd> to be a bit more explicit

jtepe: Sure, that makes sense. Added.

the-mikedavis (approved):
thanks!

Issue #1187, opened 2021-11-29
Helix in unfocused window should appear differently than in a focused window


dannasessha:
At a glance it is impossible to determine if a helix instance is in focus (actively being used) or not. Often users will have more than one instance of a text editor open right next to each other on their screen. They should be able to see from some visual cue which instance is in focus.

I imagine an inactive helix window being slightly desaturated (colors all shifted to be more grey), or, minimally, having a clear indication of which cursor(s) are in an active in-focus window.

sudormrfbin:
Do you mean split windows inside the editor itself or different terminal windows running separate `hx` instances ? For the latter it would be the responsibility of the window manager or desktop environment to dim inactive windows if required though. I remember an issue being opened about differentiating between active and inactive splits, can't find it right now.

dannasessha:
> Do you mean split windows inside the editor itself or different terminal windows running separate `hx` instances ? For the latter it would be the responsibility of the window manager or desktop environment to dim inactive windows if required though.

I meant really, both, in the sense that the active cursor could look different than an inactive one - or that the whole buffer or pane that is in focus, and, therefore, ready to be interacted with by a user, would look different than other buffers, panes, windows, and so on.

I see your point about the window manager. I don't have a completely specific vision for this yet. I think having _something_ in the helix UI that says 'ready!' would be nice, again, due to the fact that many users including myself can have a number of windows, buffers, splits, panes etc all open at once.

archseer:
For determining the active split I've been relying on the statusline (there's an inactive style). I've been missing some sort of indicator when the editor isn't in focus though, but this isn't possible since we get no such events from the terminal. It's more visual in vim because the terminal cursor will move out of the editor, but ours are "hand-drawn" due to multiple selections and are always present, Kakoune has the same behavior.

sudormrfbin:
#1154 uses the terminal cursor, so the cursor becomes hollowed (works on alacritty, kitty, wezterm, etc) which is what I rely on with vim:

![terminal-cursor-shape-change-on-focus](https://user-images.githubusercontent.com/23398472/144073718-dd50f048-0409-4be2-abab-16085c7a6295.gif)

dannasessha:
> 
This looks good! I think it would be great to use, even as an option.

If this is the best way to go, but needs something for configurability or implementation, I'd be interested in trying to make it work. I'll also keep my eye on #1154.

reissbaker:
[...]

PR #2751, opened 2022-06-11
Add basic dimming functionality

farwyler:
This adds a first iteration of color dimming to the editor.

You can:
1. Dim the colors of all unfocused views
2. Dim the colors in the background of Overlays (pickers, etc.)

The following dimming functions are implemented:
- "dim" just set the modifier flag "dim" of all cells
- "lighten" / "darken" actually change the shade of the fg/bg colors. this only works for themes applying RGB colors, not ANSI colors.

Config example:

### `[editor.dim]` Section

Enable dimming in certain areas. Disabled by default. Enable by assigning:
- 0 = set text modifier flag DIM
- negative value = darken colors (-127 = 100% darker)
- positive value = lighten colors (+127 = 100% lighter)

> darken/lighten colors only works with themes that set rgb colors, not indexed colors.

| Key | Description | Default |
|--|--|---------|
| `overlay-backdrops` | Enable dimming and set shade of content behind overlays. | `None` |
| `unfocused-views`| Enable dimming and set shade of unfocused editor views. | `None` |

```toml
[editor.dim]
# darken backdrop of overlays
overlay-backdrops = -25
# dim fg color of unfocused views
unfocused-views = 0
```

Screenshots:

Screenshot 2022-06-11 at 19 15 28
Screenshot 2022-06-11 at 19 16 09

Todo:
* [x] feedback
* [x] documentation
* [x] some optimizations

Looking for feedback.
Do we actually want this functionality in the tui?
Leave this to alternative frontends or plugins later?
Could this be used with  #1187 ?

### DIFF
--- a/helix-tui/src/buffer.rs
+++ b/helix-tui/src/buffer.rs
@@ -1,9 +1,10 @@
 use crate::text::{Span, Spans};
 use helix_core::unicode::width::UnicodeWidthStr;
+use helix_view::graphics::{Color, DimFunction, DimOperation, Modifier, Rect, Style};
+use palette::Shade;
 use std::cmp::min;
+use std::collections::HashMap;
 use unicode_segmentation::UnicodeSegmentation;
-
-use helix_view::graphics::{Color, Modifier, Rect, Style};
 
 /// A buffer cell
 #[derive(Debug, Clone, PartialEq)]
@@ -582,6 +583,58 @@ impl Buffer {
             invalidated = std::cmp::max(affected_width, invalidated).saturating_sub(1);
         }
         updates
+    }
+
+    /// Apply DimOperation to all cells in area
+    pub fn dim(&mut self, area: Rect, op: &DimOperation) {
+        use DimFunction::*;
+
+        type Srgb8 = palette::rgb::Rgb<palette::encoding::Srgb, u8>;
+
+        let mut cache: HashMap<Color, Color> = HashMap::new();
+        let intensity = op.intensity as f32 / 100.0; // TODO: clamp
+
+        let mut modify = |col| {
+            let result = cache.entry(col).or_insert_with(|| {
+                if let Color::Rgb(r, g, b) = col {
+                    let mut rgb = Srgb8::new(r, g, b).into_format().into_linear();
+                    rgb = match op.func {
+                        Lighten => rgb.lighten(intensity),
+                        Darken => rgb.darken(intensity),
+                        _ => unimplemented!(),
+                    };
+                    let rgb = Srgb8::from(rgb).into_components();
+                    Color::Rgb(rgb.0, rgb.1, rgb.2)
+                } else {
+                    col
+                }
+            });
+            *result
+        };
+        for y in area.top()..area.bottom() {
+            for x in area.left()..area.right() {
+                let cell = &mut self[(x, y)];
+
+                match op.func {
+                    Dim => {
+                        cell.modifier.insert(Modifier::DIM);
+                    }
+                    Lighten | Darken => {
+                        match cell.fg {
+                            Color::Rgb(..) => {
+                                cell.fg = modify(cell.fg);
+                            }
+                            _ => {
+                                // cell.modifier.insert(Modifier::DIM);
+                            }
+                        };
+                        if let Color::Rgb(..) = cell.bg {
+                            cell.bg = modify(cell.bg);
+                        }
+                    }
+                }
+            }
+        }
     }
 }
 
--- a/helix-tui/src/buffer.rs
+++ b/helix-tui/src/buffer.rs
@@ -585,9 +585,9 @@ impl Buffer {
 
     /// Apply shade to all cells in area.
     ///
-    /// shade = 0 -> set text modifier DIM
-    /// shade < 0 -> darken rgb color
-    /// shade > 0 -> lighten rgb color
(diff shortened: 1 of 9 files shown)


sudormrfbin:
I guess this is a nice to have, but a config for it seems overkill, especially the kind of dimming functions to use. If this does get merged, we would also want to avoid pulling in a dependency (`palette`) simply to dim colors.

---

It does feel like a good usability improvement after trying it out though, so I'm more or less on the fence now 😛

farwyler:
thanks @CptPotato  and @sudormrfbin for the feedback.
i have simplified the implementation and removed the dependency.
new configuration looks like [this](https://github.com/farwyler/helix/blob/dimming/book/src/configuration.md#editordim-section).

helix-tui/src/buffer.rs
```diff
@@ -583,6 +583,50 @@ impl Buffer {
...
+    /// shade > 0 -> lighten rgb color
+    pub fn dim(&mut self, area: Rect, shade: i8) {
+        let shade_amount = shade.unsigned_abs().saturating_mul(2);
+        let shade_fn = if shade < 0 {
+            u8::saturating_sub
+        } else {
+            u8::saturating_add
+        };
+        // TODO: replace with saturating_add_signed when mixed_integer_ops (https://github.com/rust-lang/rust/issues/87840) get stabilized.
+        let shaded = |v| shade_fn(v, shade_amount);
```
yvt, line 600: I'd like to note that saturating addition/subtraction has a very different effect than that of [`palette::Shade::{lighten, darken}`](https://docs.rs/palette/0.6.0/src/palette/rgb/rgb.rs.html#535-580), which was used before 9ad01a06232ae03ed99cbd601764aab29fe5b379, and is more likely to cause undesirable effects such as clipping, saturation boosting, and hue shifting (examples shown below).

-  `[(100, 70, 10), (10, 100, 0), (50, 80, 20)]` (various dark tones) →  `[(0, 0, 0); 3]` (black)
-  `(200, 100, 100)` (desaturated red) →  `(100, 0, 0)` (darker, but more vivid red)
-  `(200, 100, 0)` (orange) →  `(100, 0, 0)` (dark red)

Here's a possible implementation based on alpha compositing without gamma correction (warning: untested):

```suggestion
        let alpha = i32::from(shade).unsigned_abs();
        let (src_factor, dst_factor) = (alpha, 256 - alpha);
        let src = if shade > 0 { 255u32 } else { 0u32 } * src_factor;
        let shaded = |dst_color: u8| ((u32::from(dst_color) * dst_factor + src) >> 8) as u8;
```

... which should produce colors like these:

-  `[(100, 70, 10), (10, 100, 0), (50, 80, 20)]` (various dark tones) →  `[(50, 35, 5), (5, 50, 0), (25, 40, 10)]` (various darker tones)
-  `(200, 100, 100)` (desaturated red) →  `(100, 50, 50)` (darker, still desaturated red)
-  `(200, 100, 0)` (orange) →  `(100, 50, 0)` (dark orange)

helix-tui/src/buffer.rs
```diff
@@ -583,6 +583,50 @@ impl Buffer {
...
+
+        let mut cache: HashMap<Color, Color> = HashMap::new();
+        let mut modify = |col| {
+            let result = cache.entry(col).or_insert_with(|| {
```
yvt, line 604: I suspect that the color transformation is faster than SipHash 2-4 hashing + memory allocation.

helix-tui/src/buffer.rs
farwyler: This works great. Thank you 🙏

helix-tui/src/buffer.rs
farwyler: i just benched it. you are right. way too much overhead at these tiny resolutions.
thanks for catching this. removed it in 8625892af12c3f92e4f6035835169ace0f95e44c

helix-tui/src/buffer.rs
```diff
@@ -583,6 +582,35 @@ impl Buffer {
         }
         updates
     }
+
+    /// Apply shade to all cells in area.
+    ///
+    /// shade = 0 -> set text modifier DIM
+    /// shade < 0 -> darken rgb color
+    /// shade > 0 -> lighten rgb color
```
yvt, line 590: Regular line breaks in doc comments might not render as line breaks (CommonMark [§6.8](https://spec.commonmark.org/0.30/#hard-line-breaks))

```suggestion
    /// - `shade = 0`: set text modifier `DIM`
    /// - `shade < 0`: darken rgb color
    /// - `shade > 0`: lighten rgb color
```

archseer:
Doing dimming across an entire area after the fact seems a bit of a hack -- I'd rather calculate the dimming inside render_view, similar to how we treat active/unfocused views by passing a bool into that function.

farwyler:
@archseer You are right, the post processing is quite hacky. I wanted to keep the dimming out of the rendering functions if possible, as it introduces code complexity there for something that is a graphical gimmick, at best a slight UX improvement.

With bb4fec7ee8ecaf596f932bbc18806ab1439210a7 i've tried another an implementation that does the dimming ad-hoc, with just slight adjustments to the main render loop and a little state in the drawing buffer.

pascalkuthe:
clsoing this one as stale. Almost the entire rendering code has been removed since so this would need to start over from scratch. Thank you for contributing!

PR #2754, opened 2022-06-11
Right align scrollbar with track in completion popup

sudormrfbin:
Depends on #2728.

---

- Align the scollbar to the right edge of the popup rather than at a margin of one.
- Add a scrollbar track and a new scope `ui.menu.scroll`.

| Before | After |
| --- | --- |
| before | after |

A multiline scrollbar like the one mentioned in https://github.com/helix-editor/helix/pull/2728#issuecomment-1152034487 would be nicer though:

multiline scrollbar

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -310,7 +310,7 @@ impl<T: Item + 'static> Component for Menu<T> {
         use tui::widgets::TableState;
 
         table.render_table(
-            area.clip_left(Self::LEFT_PADDING as u16),
+            area.clip_left(Self::LEFT_PADDING as u16).clip_right(1),
             surface,
             &mut TableState {
                 offset: scroll,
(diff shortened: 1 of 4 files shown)


helix-term/src/ui/menu.rs
```diff
@@ -320,10 +320,8 @@ impl<T: Item + 'static> Component for Menu<T> {
             let is_marked = i >= scroll_line && i < scroll_line + scroll_height;
```
sudormrfbin, line 320: @archseer Reading the code makes me think that this is actually supposed to be a multiline scroll bar but there's a bug, is that so?

n0s4: It is multiline for me:

hx-completion-menu

`git describe` -> `22.05-49-g567e71fb`

archseer: Yeah it's multi-line depending on total list length/page size

archseer:
> Changes the scrollbar block character to be thicker.

I preferred the thinner one! But rendering block chars seems to vary a lot by terminal

sudormrfbin:
I was experimenting with adding a scroll background, which looks better with a thicker bar. We can't style cells partially with another color, hence if a half block is used the selection style will look cut off:


| Thin half bar | Thick full bar |
| --- | --- |
| halfbar | fullbar |

The thin bar looks much better but the gap is nagging. Is the background something that we want? If so which type of block character?

helix-term/src/ui/menu.rs
sudormrfbin: Ah what confused me was that a shorter scrollbar usually means longer content and a longer scrollbar means shorter content in other applications, but here it seems to be inverted. I think inverting it will be more intuitive.

sudormrfbin: Fixed in #2860, twas a typo in the floor division impl :)

archseer:
Oh yeah the alignment was an off-by-1 error for sure 😆

archseer:
Can you rebase?

sudormrfbin:
Rebased!

Issue #497, opened 2021-07-24
Add bufferline


gbaranski:
List of opened buffers on the top, that would help workinng on multiple files at the same time, without using C-o and C-i, or opening file search every time.

Neovim(https://github.com/akinsho/nvim-bufferline.lua)
![111992693-9c6a9b00-8b0d-11eb-8c39-19db58583061](https://user-images.githubusercontent.com/34353640/126870562-732323de-bba5-4421-aed4-7f6bd3566845.gif)

VSCode:
![preview-tab](https://user-images.githubusercontent.com/34353640/126878894-d0fea40a-6cd0-4dff-8e62-e5e65781d80b.png)

kirawi:
Duplicate of #475

kirawi:
On second thought, this seems more descriptive than the aforementioned issue. Reopening it for now.

aaron404:
I have an initial implementation here, looking for feedback.
https://github.com/helix-editor/helix/pull/2759

Issue #2295, opened 2022-04-27
Add vim/neovim style tabs


adsnaider:
#### Describe your feature request

I would like to have something similar to `:tabe` on vim / neovim as it happens to be an important part of my workflow.

adsnaider:
I should mention, I would be happy to contribute this feature if I can get some direction to get started.

the-mikedavis:
I haven't found myself missing tabs as I've been using Helix. Since the buffer management (especially the picker) is so good I don't find cases where I would use it. I wonder what in your workflow makes tabs so necessary?

sudormrfbin:
Tabs are useful when you want to glance at other files without changing your current layout of open splits. Atleast that's the primary use I have for them.

Mofiqul:
In Neovim we can have different working directory per tab which is excellent to working with different directory at the same time.

David-Else:
I think tabs would be bloat and should be the job of the terminal. In [Kitty](https://sw.kovidgoyal.net/kitty/) you can control them programmatically, for example, if the project has a git repo then Lazygit is opened in the next tab:
```
# if the dir is a git repo then open a tab and run lazygit
if [[ -d ".git" ]]; then
    kitty @ launch --type=tab --tab-title="lazygit: $selected_project_name" --keep-focus --cwd="$(pwd)" lazygit
fi

# when quit close the lazygit tab if it exists
kitty @ send-text --match-tab title:"lazygit: $selected_project_name" q

```

adsnaider:
@the-mikedavis, that's a fair point. I can see myself adopting a different workflow given that it's a different editor. However, I do think that there's good reasons to have tabs and some people just prefer having that workflow available.

@David-Else I disagree on tabs being a bloat. First thing is that tabs are a relatively simple concept: a container of views (splits). Tabs simply allow you to have many instead of just one. In Vim/Neovim, tabs add negligible visual clutter. Additionally, most editors and IDEs have a concept of tabs. I prefer the vim-style tabs (splits within tabs) over say the vscode-style tabs (tabs within splits), and I would imagine that if Helix were to adopt tabs, it would align with the vim way.

Nukesor:
lTabs are indeed one of the main features I'm currently missing from Helix.

I wouldn't consider in-built tabs bloat. But I would go so far as to say that not having tabs is pretty inefficient and cumbersome.

My usecase of tabs is that I usually work on multiple "scopes" in the same codebase at the same time.
[...]

PR #2759, opened 2022-06-13
initial implementation of bufferline

aaron404:
- adds new "bufferline" setting, which enables rendering of a line displaying currently open buffers
- has 3 modes `always` (always show), `never` (never show, default), `multiple` (show if num buffers > 1)

helix2

[#497](https://github.com/helix-editor/helix/issues/497)


the-mikedavis:
I think we should call this bufferline instead of tabs. Tabs as in https://github.com/helix-editor/helix/issues/2295 behave differently.

the-mikedavis (commented):
Could you add some docs on this to the book (https://github.com/helix-editor/helix/blob/master/book/src/configuration.md)?

lazytanuki:
It could be nice to have an option that, similarly to some other editors, only shows the bufferline when there is more than one buffer to display in it.

helix-term/src/ui/editor.rs
```diff
@@ -541,6 +554,43 @@ impl EditorView {
...
+                .file_name()
+                .unwrap_or_default()
+                .to_str()
+                .unwrap();
+
+            let style = if view.doc == doc.id() {
+                theme.get("ui.background")
+            } else {
+                theme.get("ui.statusline")
+            };
```
sudormrfbin, line 581: Introducing new theme scopes like `ui.bufferline` and `ui.bufferline.active` which fall back to these values would be a nice addition.

sudormrfbin (commented):
A couple of improvements that could be implemented (not necessarily in this PR, since I'd say it's at the sweet spot for ease of reviewing):

- Showing previous path components when two buffers have files with the same name.
- Taking the width of the viewport into account and aligning the tabs so that current buffer name is visible even on small terminals.

helix-term/src/ui/editor.rs
aaron404: Is there a straighforward way to have "ui.bufferline" fallback to "ui.statusline"? Reading the docs it seems that it would simply fallback to "ui".

book/src/configuration.md
```diff
@@ -45,6 +45,7 @@ hidden = false
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
+| `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `multiple`
```
zen3ger, line 48: not sure how forgiving the markdown renderer is, but there's a missing `|` at the end.

helix-term/src/ui/editor.rs
```diff
@@ -541,6 +541,44 @@ impl EditorView {
...
+            } else {
+                editor.theme.get("ui.statusline")
+            };
+
+            surface.set_string(
+                1 + viewport.x + len as u16,
+                viewport.y,
+                format!(" {} {}", fname, if doc.is_modified() { "● " } else { "" }),
+                style,
+            );
```
the-mikedavis, line 569: it looks like this block is redundant because of the call to `surface.set_string` below

helix-term/src/ui/editor.rs
```diff
@@ -541,6 +541,44 @@ impl EditorView {
...
+
+            let (text, offset) = if doc.is_modified() {
+                (format!(" {} ● ", fname), fname.len() + 4)
```
the-mikedavis, line 572:
```suggestion
                (format!(" {}[+] ", fname), fname.len() + 5)
```

Let's reuse the modified indicator from the status-line for consistency

helix-term/src/ui/editor.rs
the-mikedavis: You could do something like `theme.try_get("ui.bufferline").unwrap_or_else(|| theme.get("ui.statusline")`. That would behave differently from how other theme fallbacks work though, so I think we should just introduce a new `ui.bufferline` and copy `ui.statusline` values in the existing themes

aaron404:
> Though, I'm not sure if quick switching and adding shortcuts should be within the scope of this initial PR.

Yeah I'd be happy to do this as a follow up PR if there was interest in it, but for now I think it is out of scope

lazytanuki:
Another thing that would be nice is to make it automatically scroll to display the currently viewed buffer. At the moment, if we have too many open buffers, the bufferline will overflow at the end. A bit like this from nvim-cokeline:

this

However, implementing the "natural" behavior would require tracking the bufferline position, in order for the render to know where was the focus during the previous render.

An alternative to postpone this could be to shorten the filenames when the bufferline gets too wide.

book/src/configuration.md
```diff
@@ -45,6 +45,7 @@ hidden = false
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
+| `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `multiple` |
```
the-mikedavis, line 48: I'd prefer this default to `never`. I think it falls under the category of features that some really like and others find distracting or overly verbose like visible whitespace (also indent guides for example).

helix-term/src/ui/editor.rs
```diff
@@ -541,6 +541,45 @@ impl EditorView {
...
+                .unwrap();
+
+            let style = if view!(editor).doc == doc.id() {
+                // editor.theme.get("ui.background")
+                editor
+                    .theme
+                    .try_get("ui.bufferline.active")
+                    .unwrap_or_else(|| editor.theme.get("ui.background"))
+            } else {
+                // editor.theme.get("ui.statusline")
```
the-mikedavis, line 565: these comments can be cleaned up

sbromberger:
I might be missing something, but how do I specify a separator between the open buffers? Right now it looks like

bufferline

sbromberger:
I've been using this for the past several days and have come to rely on it. No glaring issues that I see. FWIW, I would love to see this merged soon.

book/src/configuration.md
aaron404: There was some feedback earlier that `multiple` would be a good default, but if there is agreement I am happy to change it back to never.

aaron404:
> I might be missing something, but how do I specify a separator between the open buffers?

I haven't added any way to configure this and hadn't planned on it. Perhaps someone could do a follow up to add this functionality.

sbromberger:
> > I might be missing something, but how do I specify a separator between the open buffers?
> 
> I haven't added any way to configure this and hadn't planned on it. Perhaps someone could do a follow up to add this functionality.

Yes - sorry - I absolutely do NOT want to hold this PR up. I'll take a look at what needs to be implemented once this goes in. (I'll probably have to make some changes due to #2831 if that gets merged, too.)

book/src/configuration.md
sudormrfbin: I would personally vote for having `multiple` by default since it's more discoverable that way and can be easily disabled in the config. But then again that argument could be extended to every single new feature so :)

aaron404: I am personally also of the opinion that `multiple` is the better default option. Anyone who cares that much about the wasted space of a single line will probably have a bunch of custom options set in their config anyways, and adding one more to hide it should be trivial.

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,43 @@ impl EditorView {
...
+                    .theme
+                    .try_get("ui.bufferline")
+                    .unwrap_or_else(|| editor.theme.get("ui.statusline"))
+            };
+
+            let (text, offset) = if doc.is_modified() {
+                (format!(" {}[+] ", fname), fname.len() + 5)
+            } else {
+                (format!(" {} ", fname), fname.len() + 2)
+            };
```
sudormrfbin, line 611:
```suggestion
            let text = format!(" {}{} ", fname, if doc.is_modified() { "[+]" } else { "" });
            let offset = text.len();
```

book/src/configuration.md
the-mikedavis: I'm curious what @archseer thinks about the default behavior.

I was going to say that the default keymap favors the buffer picker for switching buffers but that's not actually true: `gn`/`gp` are the same number of keystrokes. I'm definitely biased as I only ever use the buffer picker.

> Anyone who cares that much about the wasted space of a single line will probably have a bunch of custom options set in their config anyways, and adding one more to hide it should be trivial.

It's more consistent with the current configuration options to make the reverse argument: anyone who wants to enable a cosmetic feature will probably have a bunch of custom options set in their config anyways, and adding one more to show it should be trivial. Examples: visible whitespace, indent guides, cursorline, rulers, custom cursor shapes.

> it's more discoverable [enabled by default]

Hmm yeah I agree. IMO though a toggle mode like #2849 would be the ideal approach to that - enabling new features by default as a general policy might end up with a busy/overwhelming default experience.

archseer:
Bufferline is not something I'd like to enable by default, in my opinion it encourages the wrong workflow (treat views == tabs), it's just a less featureful version of space+b buffer list. But I understand people are used to sublime/vscode-like tabs.

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,40 @@ impl EditorView {
...
+        surface.clear_with(viewport, editor.theme.get("ui.statusline"));
+        let mut len = 0usize;
+        for doc in editor.documents() {
+            let fname = doc
+                .path()
+                .unwrap_or(&pb)
+                .file_name()
+                .unwrap_or_default()
+                .to_str()
+                .unwrap();
```
archseer, line 593: No unwraps, use unwrap_or_default()

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,40 @@ impl EditorView {
         }
     }
 
+    /// Render bufferline at the top
+    pub fn render_bufferline(editor: &Editor, viewport: Rect, surface: &mut Surface) {
+        let pb = PathBuf::from(SCRATCH_BUFFER_NAME); // default filename to use for scratch buffer
+        surface.clear_with(viewport, editor.theme.get("ui.statusline"));
```
lazytanuki, line 584: Could we have a `ui.bufferline.background` style here ?

```suggestion
        surface.clear_with(
            viewport,
            editor
                .theme
                .try_get("ui.bufferline.background")
                .unwrap_or_else(|| editor.theme.get("ui.statusline")),
        );
```

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,40 @@ impl EditorView {
         }
     }
 
+    /// Render bufferline at the top
+    pub fn render_bufferline(editor: &Editor, viewport: Rect, surface: &mut Surface) {
+        let pb = PathBuf::from(SCRATCH_BUFFER_NAME); // default filename to use for scratch buffer
```
the-mikedavis, line 583: this is a bit of a nit but this variable name could be more descriptive: `scratch` or maybe `scratch_path`

helix-term/src/ui/editor.rs
```diff
@@ -1346,8 +1380,27 @@ impl Component for EditorView {
         // clear with background color
         surface.set_style(area, cx.editor.theme.get("ui.background"));
         let config = cx.editor.config();
+
+        // check if bufferline should be rendered
+        use helix_view::editor::BufferLine;
+        let use_bufferline = match config.bufferline {
+            BufferLine::Always => true,
+            BufferLine::Multiple if cx.editor.documents().count() > 1 => true,
```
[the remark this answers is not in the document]
sudormrfbin: `is_empty()` would return true even when there's only one doc, whereas we want it to be true only if there are >= 2 docs. So we would need something like:

```suggestion
            BufferLine::Multiple if cx.editor.documents().nth(1).is_some() => true, 
```

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,40 @@ impl EditorView {
...
+                editor
+                    .theme
+                    .try_get("ui.bufferline")
+                    .unwrap_or_else(|| editor.theme.get("ui.statusline"))
+            };
+
+            let text = format!(" {}{} ", fname, if doc.is_modified() { "[+]" } else { "" });
+            let offset = text.len();
+
+            surface.set_string(1 + viewport.x + len as u16, viewport.y, text, style);
```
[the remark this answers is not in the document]
aaron404: my understanding of `set_string` was that it automatically clipped the string to surface bounds. Is this not the case?

helix-term/src/ui/editor.rs
```diff
@@ -1346,8 +1386,27 @@ impl Component for EditorView {
         // clear with background color
         surface.set_style(area, cx.editor.theme.get("ui.background"));
         let config = cx.editor.config();
+
+        // check if bufferline should be rendered
+        use helix_view::editor::BufferLine;
+        let use_bufferline = match config.bufferline {
+            BufferLine::Always => true,
+            BufferLine::Multiple if cx.editor.documents().count() > 1 => true,
```
aaron404, line 1394: What about this, just grab the size of the BTree directly?
```suggestion
            BufferLine::Always => true,
            BufferLine::Multiple if cx.editor.documents.len() > 1 => true,
```

book/src/configuration.md
aaron404: I'll go with Never by default, based on @archseer's comment below

helix-term/src/ui/editor.rs
```diff
@@ -1346,8 +1380,27 @@ impl Component for EditorView {
...
+            _ => false,
+        };
+
+        // -1 for commandline and -1 for bufferline
+        let editor_area = if use_bufferline {
+            area.clip_bottom(1).clip_top(1)
+        } else {
+            area.clip_bottom(1)
```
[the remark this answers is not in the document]
aaron404: is it ok to make `editor_area` `mut` then? I generally avoid mut unless it's required. If this is the preference I'll change it though

helix-term/src/ui/editor.rs
```diff
@@ -578,6 +578,50 @@ impl EditorView {
         }
     }
 
+    /// Render bufferline at the top
+    pub fn render_bufferline(editor: &Editor, viewport: Rect, surface: &mut Surface) {
+        let editor = PathBuf::from(SCRATCH_BUFFER_NAME); // default filename to use for scratch buffer
```
aaron404, line 583:
```suggestion
        let scratch = PathBuf::from(SCRATCH_BUFFER_NAME); // default filename to use for scratch buffer
```

helix-term/src/ui/editor.rs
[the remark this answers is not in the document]
archseer: Yeah it's fine

helix-term/src/ui/editor.rs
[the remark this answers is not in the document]
archseer: It doesn't no, it uses `usize::MAX`. Any rendering functions will panic if they go offscreen.

aaron404:
I am not sure what is holding this up. Github says there are changes requested but it looks like all the conversations are resolved.

the-mikedavis:
Sorry about the wait - this is just waiting on the next round of review. It can take a while since there are so many PRs in the queue

helix-term/src/ui/editor.rs
```diff
@@ -582,6 +582,57 @@ impl EditorView {
...
+                text,
+                surface.area.width as usize,
+                style,
+            );
+
+            len += offset;
+
+            if len > surface.area.width as usize {
+                break;
+            }
```
sudormrfbin, line 632: The offset calculation for `len` will have to use `text.width()` instead of `text.len()`, but we can simply use the return value of `surface.set_string()` (a tuple `(x, y)` with the end position) which does all of the offset calculation for us:

```suggestion
            let used_width = viewport.x - x; // x is initialized to viewport.x before the loop
            let rem_width = surface.area.width.saturating_sub(used_width);
            x = surface
                .set_stringn(x, viewport.y, text, rem_width as usize, style)
                .0;

            if x >= surface.area.right() {
                break;
            }
```

A better solution would be to have a version of [`Buffer::set_spans()`](https://github.com/helix-editor/helix/blob/a766b32ed197221aca7a9fec3af805b4d04a6f1d/helix-tui/src/buffer.rs#L418) that takes an iterator of `Span`s (a Span is a string + a style). Then the width and offset calculation would be managed for us if we output the buffer names from the iterator. But a plain for loop seems good enough for now 👍🏾.

helix-term/src/ui/editor.rs
```diff
@@ -582,6 +582,57 @@ impl EditorView {
...
+            let style = if view!(editor).doc == doc.id() {
+                editor
+                    .theme
+                    .try_get("ui.bufferline.active")
+                    .unwrap_or_else(|| editor.theme.get("ui.background"))
+            } else {
+                editor
+                    .theme
+                    .try_get("ui.bufferline")
+                    .unwrap_or_else(|| editor.theme.get("ui.statusline"))
```
sudormrfbin, line 614: The theme style values should be calculated once outside the loop and reused on every iteration.

sudormrfbin:
CI seems to be failing on [`SCRATCH_BUFFER_NAME` not in scope][1], but it indeed is and builds locally too.

[1]: https://github.com/helix-editor/helix/runs/7492588246?check_suite_focus=true#step:6:21

aaron404:
> CI seems to be failing on [`SCRATCH_BUFFER_NAME` not in scope](https://github.com/helix-editor/helix/runs/7492588246?check_suite_focus=true#step:6:21), but it indeed is and builds locally too.

I merged master into it, maybe that will help

sbromberger:
Any chance this might be reviewed again (and hopefully merged) soon? I've been using it for about a month and have come to depend on it. Would be nice to have one fewer thing that requires a custom build. Thanks!

cor:
This looks fantastic! Love the amount of context it provides. I hope it gets merged soon

helix-term/src/ui/editor.rs
```diff
@@ -592,6 +592,57 @@ impl EditorView {
...
+        for doc in editor.documents() {
+            let fname = doc
+                .path()
+                .unwrap_or(&scratch)
+                .file_name()
+                .unwrap_or_default()
+                .to_str()
+                .unwrap_or_default();
+
+            let style = if view!(editor).doc == doc.id() {
```
archseer, line 626: `let current_doc = view!(editor).doc` before the for loop to avoid fetching the current doc each time it loops

archseer (changes requested):

helix-term/src/ui/editor.rs
```diff
@@ -138,7 +138,7 @@ impl EditorView {
             highlights,
             &editor.config(),
         );
-        Self::render_gutter(editor, doc, view, view.area, surface, theme, is_focused);
+        Self::render_gutter(editor, doc, view, area, surface, theme, is_focused);
```
archseer, line 141: Why the change here?

helix-term/src/ui/editor.rs
```diff
@@ -592,6 +592,57 @@ impl EditorView {
...
+            editor
+                .theme
+                .try_get("ui.bufferline.background")
+                .unwrap_or_else(|| editor.theme.get("ui.statusline")),
+        );
+
+        let bufferline_active = editor
+            .theme
+            .try_get("ui.bufferline.active")
+            .unwrap_or_else(|| editor.theme.get("ui.background"));
```
archseer, line 609: I think we could use `ui.statusline` and `ui.statusline.inactive` as fallback

helix-term/src/ui/editor.rs
aaron404: I honestly cannot remember.. I have reverted it for now

archseer (approved):

hemedani:
how can I active the bufferline ?

CptPotato:
> how can I activate the bufferline ?

You can set `bufferline` to `always`, `never` or `multiple` in the editor section of your helix config, see https://docs.helix-editor.com/master/configuration.html#editor-section.

hemedani:
tanks

yashpalgoyal1304:
> > It could be nice to have an option that, similarly to some other editors, only shows the bufferline when there is more than one buffer to display in it.
> 
> That seems like good default behavior without needing a config option.

i came here to say exactly that, multiple should be default?!

yashpalgoyal1304:
> Unchanged files with check annotations (Beta)
> Check warning on line 375 in helix-term/src/ui/picker.rs
>
> GitHub Actions / Check (msrv)
> use of deprecated associated function `retain_mut::RetainMut::retain_mut`: Rust 1.61 has included retain_mut directly


What does this mean??

PR #2763, opened 2022-06-14
adds missing tree-sitter-comment injection for js/ts

farwyler:
this adds the injection for tree-sitter-comment to javascript. i am not sure if there is a reason for this missing. maybe it was supposed to interfere with the injection for "jsdoc" comments?

anyhow, it does seem to work to with both injections on "comment".
Screenshot 2022-06-14 at 06 46 27

### DIFF
--- a/runtime/queries/javascript/injections.scm
+++ b/runtime/queries/javascript/injections.scm
@@ -22,12 +22,15 @@
 ((regex_pattern) @injection.content
  (#set! injection.language "regex"))
 
-; Parse JSDoc annotations in comments
+; Parse JSDoc annotations in multiline comments
 
 ((comment) @injection.content
- (#set! injection.language "jsdoc"))
+ (#set! injection.language "jsdoc")
+ (#match? @injection.content "^/\\*+"))
 
-; Parse comment tags (TODO, FIXME, ...)
+; Parse general tags in single line comments
 
-([(comment)] @injection.content
- (#set! injection.language "comment"))
+((comment) @injection.content
+ (#set! injection.language "comment")
+ (#match? @injection.content "^//"))
+
--- a/runtime/queries/javascript/injections.scm
+++ b/runtime/queries/javascript/injections.scm
@@ -22,7 +22,12 @@
 ((regex_pattern) @injection.content
  (#set! injection.language "regex"))
 
- ; Parse JSDoc annotations in comments
+; Parse JSDoc annotations in comments
 
 ((comment) @injection.content
  (#set! injection.language "jsdoc"))
+
+; Parse comment tags (TODO, FIXME, ...)
+
+([(comment)] @injection.content
+ (#set! injection.language "comment"))
--- a/runtime/queries/jsdoc/injections.scm
+++ b/runtime/queries/jsdoc/injections.scm
@@ -1,0 +1,5 @@
+; Parse general comment tags
+
+((document) @injection.content
+ (#set! injection.include-children)
+ (#set! injection.language "comment"))


archseer:
You probably want to inject the comment grammar into the jsdoc grammar?

farwyler:
> You probably want to inject the comment grammar into the jsdoc grammar?

I am not sure how. The jsdoc injection is already attached to the (comment) node. How would i capture the root node of jsdoc then, to inject the comment parser there?

Doing this for **javascript** 

```scheme
((comment) @injection.content (#set! injection.language "comment"))
((comment) @injection.content (#set! injection.language "jsdoc"))
```

... actually works and is how the queries are defined in neovim. The ordering here is actually important.
I don't know if this is a valid approach given the current implementation though.

If it is, there seems to be a small issue here:

https://github.com/helix-editor/helix/blob/ad15e7b5e8918e1521683c7f5c63d77e00c58023/helix-core/src/syntax.rs#L625

When updating tree-sitter, the language layers for the injections are added in order as they are discovered in the document. In our case of **comment** fighting with **jsdoc** at the same node, it only produces the correct highlighting if the language layer for **comment** comes before **jsdoc**.

So is it technically and implemenation-wise valid to have two parsers capture the same node? And if so, shouldn't the ordering of layers for the injections be stable as it is defined in the queries?

Feedback much appreciated on this. I would really like to get a better grasp of the implementation here.

the-mikedavis:
The root node for jsdoc is [`document`](https://github.com/tree-sitter/tree-sitter-jsdoc/blob/189a6a4829beb9cdbe837260653b4a3dfb0cc3db/grammar.js#L13) but actually we can be more specific and inject comment into [`description`](https://github.com/tree-sitter/tree-sitter-jsdoc/blob/189a6a4829beb9cdbe837260653b4a3dfb0cc3db/grammar.js#L20-L23) which covers a bunch of text:

```scm
; runtime/queries/jsdoc/injections.scm
((description) @injection.content
```

farwyler:
> The root node for jsdoc is [`document`](https://github.com/tree-sitter/tree-sitter-jsdoc/blob/189a6a4829beb9cdbe837260653b4a3dfb0cc3db/grammar.js#L13) but actually we can be more specific and inject comment into [`description`](https://github.com/tree-sitter/tree-sitter-jsdoc/blob/189a6a4829beb9cdbe837260653b4a3dfb0cc3db/grammar.js#L20-L23) which covers a bunch of text:
> 
> ```scheme
> ; runtime/queries/jsdoc/injections.scm
> ((description) @injection.content
>  (#set! injection.language "comment"))
> ```

`jsdoc.document` only captures multiline comments, so injecting `comment` via `jsdoc` will remove the comment highlighting from all inline comments.

sadly the javascript parser does not differentiate between single and multiline comments, as far as i can tell.

the-mikedavis:
You can tell them apart using `#match?` in queries:

```scm
; runtime/queries/javascript/injections.scm

; block comment:
((comment) @injection.content
 (#set! injection.language "jsdoc")
 (#match? @injection.content "^/\\*.*\\*/$"))

; line comment:
((comment) @injection.content
 (#set! injection.language "comment")
 (#match? @injection.content "^//"))
```

farwyler:
This works now in all kind of comment scenarios.

@the-mikedavis thanks for the support! fyi, i am using `(document)` instead of `(description)` because description seems to be specific to text following jsdoc `@tags` and not the whole comment. 

```javascript
/**
 * some description
 * FIXME: this is not (description)
 * @param {number} this counts as (description)
 * @returns
 */
```

the-mikedavis:
Ah yeah good catch, let's use `document` 👍

the-mikedavis (approved):
Gave it a spin and this looks good. Thanks!

Issue #2709, opened 2022-06-08
Command line argument to open split


kaleidawave:
#### Describe your feature request

Currently running `hx file1.txt file2.txt` will open helix with file1.txt present and file2.txt in the buffers

It would be cool if there was a command line option passed in as `--split` which would open the files in a split view. 

One difficulty would be the direction to split. Should helix pick a *layout* or should the user specify how the views should be split?

snowmang1:
I would like to work on this but I am unsure of where to begin. I have been looking around in the helix-term dir where the call to "hsplit" is made on line 467 and trying to trace back to what is actually going on. I have become lost and possibly have an incorrect understanding of what is actually happening under the hood of helix.

plexom:
What about passing in either `--vsplit` or `--hsplit` as an argument, in order to tell the program in what way to open the listed files? So, if the user were to pass in `--vsplit`, for instance, all listed files would be opened vertically next to each other. Of course we would have to handle the case that the user passes in both arguments at the same time, but that wouldn't be too difficult.

kaleidawave:
Yes this sounds good. And `--vsplit` and `--hsplit` would be mutually exclusive

pascalkuthe:
Closed by  #2773

Issue #1083, opened 2021-11-12
Too many vertical splits causes a crash


cole-h:
### Reproduction steps

1. <kbd>Ctrl</kbd>+<kbd>w</kbd>
2. <kbd>Ctrl</kbd>+<kbd>v</kbd>
3. Repeat steps 1 and 2 until you crash

### Environment

- Platform: Linux
- Helix version: v0.5.0-93-gfa0cb01

<details><summary>Backtrace</summary>

```
thread 'main' panicked at 'Trying to access position outside the buffer: x=87, y=29, area=Rect { x: 0, y: 0, width: 87, height: 31 }', helix-tui/src/buffer.rs:204:9
stack backtrace:
   0:     0x55eec65498bc - std::backtrace_rs::backtrace::libunwind::trace::h3fea1eb2e0ba2ac9
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/libunwind.rs:90:5
   1:     0x55eec65498bc - std::backtrace_rs::backtrace::trace_unsynchronized::h849d83492cbc0d59
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:     0x55eec65498bc - std::sys_common::backtrace::_print_fmt::he3179d37290f23d3
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:67:5
   3:     0x55eec65498bc - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h140f6925cad14324
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/sys_common/backtrace.rs:46:22
   4:     0x55eec6570a1c - core::fmt::write::h31b9cd1bedd7ea38
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/core/src/fmt/mod.rs:1150:17
   5:     0x55eec6543e85 - std::io::Write::write_fmt::h1fdf66f83f70913e
                               at /rustc/59eed8a2aac0230a8b53e89d4e99d55912ba6b35/library/std/src/io/mod.rs:1667:15
   6:     0x55eec654bb40 - std::sys_common::backtrace::_print::he7ac492cd19c3189
[...]

PR #2773, opened 2022-06-14
adds --vsplit and --hsplit arguments

sven-hluchy:
adds the `--vsplit` and `--hsplit` features as proposed by #2709.

When multiple files are opened in the editor via the command line, i.e. `hx test1.txt test2.txt`, the user can use these new arguments to tell the program how these files should be presented.
`hx test1.txt test2.txt --vsplit` opens both files side-by-side, whereas `hx test1.txt test2.txt --hsplit` opens them on top of each other.
I also added the new arguments to the completion scripts (hopefully correctly).
If the user passes in both of these options, `--vsplit` is used as a default.

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -1,5 +1,6 @@
 use anyhow::Result;
 use helix_core::Position;
+use helix_view::tree::Layout;
 use std::path::{Path, PathBuf};
 
 #[derive(Default)]
@@ -11,8 +12,7 @@ pub struct Args {
     pub load_tutor: bool,
     pub fetch_grammars: bool,
     pub build_grammars: bool,
-    pub vsplit: bool,
-    pub hsplit: bool,
+    pub split: Option<Layout>,
     pub verbosity: u64,
     pub files: Vec<(PathBuf, Position)>,
 }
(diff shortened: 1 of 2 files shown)


geroge2809 (approved):

EpocSquadron:
What happens if the user opens so many buffers with splits that it can't fit in the editor? Is there a limit? Should there be?

the-mikedavis:
There's a practical limit that will lead to a panic with `:vsplit` (and probably `:hsplit` too) but I'm not sure if it's based on terminal size: https://github.com/helix-editor/helix/issues/1083

EpocSquadron:
That scenario is probably better resolved in a separate PR then.

plexom:
From what I have been able to gather, the limit of splits you can have does in
fact depend on the size of the window in which helix is opened. 

With a terminal of size 2576x1426, the max number of splits the editor can open
without panicking is 17, with a size of 1442x865, the max is only 13.

If one opens more buffers than the editor can "hold", the program panics and
outputs this error message:
```
thread 'main' panicked at 'attempt to subtract with overflow', helix-view\src\tree.rs:388:50
```

This line is inside a function called `recalculate`, which, as the name implies,
is tasked with recalculating the area of the opened buffers.

helix-term/src/args.rs
```diff
@@ -11,6 +11,8 @@ pub struct Args {
     pub load_tutor: bool,
     pub fetch_grammars: bool,
     pub build_grammars: bool,
+    pub vsplit: bool,
+    pub hsplit: bool,
```
sudormrfbin, line 15: You could reuse the [`Layout` enum](https://github.com/helix-editor/helix/blob/a17626a822b36d4de3146c2d410f976e19dd189c/helix-view/src/tree.rs#L49-L53) here and represent it using a single `pub split: Option<Layout>`,where `None` would mean no splitting. The binary would still have both `--vsplit` and `--hsplit`, and only the parsing would change.

sudormrfbin (approved):

the-mikedavis (approved):
This looks great :+1:

Thanks for updating the completion too 😀

Issue #2752, opened 2022-06-11
:w selects the whole file, thus losing caret placement.


LollipopFt:
### Summary

https://user-images.githubusercontent.com/62802897/173198242-64d36aa5-4fb3-4b66-b15d-7ef8999015f3.mp4

as seen in the video above, when keying in `:w`, the entire file gets selected, and my caret always returns to the start when i exit it.

settings:
```toml
theme='onedark'
[keys.normal]
't'='move_line_down'
'n'='move_line_up'
's'='move_char_right'
'j'='find_till_char'
'k'='search_next'
'l'='select_regex'
[keys.insert]
','={'.'='normal_mode'}
[editor]
scrolloff=255
mouse=false
auto-completion=true
idle-timeout=0
completion-trigger-len=2
rulers=[80,100,120]
[editor.cursor-shape]
insert='bar'
[editor.whitespace]
render='all'
[editor.lsp]
display-messages=true
```

thank you!

### Reproduction Steps

 

I tried this:

1. `:w`

I expected this to happen:
no selections made.
Instead, this happened:
above video.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
none
```

</details>

### Platform

Windows

### Terminal Emulator

Windows Terminal Preview 1.14.1451.0

### Helix Version

helix 22.05 (27609f50)

LollipopFt:
strangely enough, it only happens to one file:

https://user-images.githubusercontent.com/62802897/173198588-3916323b-26ad-4daf-8ec2-23159d3898af.mp4

any ideas what might be happening?

sudormrfbin:
Could be due to line endings? (from #2404):

> I encountered this with rust-analyzer before. The line endings were different from what the LSP expected, and so formatting it would basically delete the entire document and send the new one with the correct line endings.

RalfNorthman:
This happens to me too, after I upgraded to 22.05. Unfortunately I don't remember what my previous helix version was.

I've only seen it in rust-files when auto-format is on. I use rust-analyzer.

Running helix with the same configuration on a remote ubuntu machine works fine, so it seems to be windows-related.

### Platform
Windows 10

### Terminal Emulator
alacritty 0.10.0 (8a26dee)

### Helix Version
helix 22.05 (27609f50)

LollipopFt:
@sudormrfbin possibly, i actually tried rewriting the entire file but it still caused the same problem. like i recreated it from scratch. so i have no idea what is happening...

[...]

PR #2778, opened 2022-06-15
Redetect indent and line endings after language server replaces documents

farwyler:
As discussed here #2752

This redetects indent and line endings after a language server replaces the whole document. This case now behaves as if the document was reloaded from disk.

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -221,29 +221,6 @@ pub mod util {
             generate_transaction_from_edits(&fmt.doc, fmt.edits, fmt.offset_encoding)
         }
     }
-
-    impl LspFormatting {
-        pub fn replaces_document(&self, old: &Rope) -> Option<&String> {
-            match &self.edits[..] {
-                [lsp::TextEdit {
-                    range:
-                        lsp::Range {
-                            start:
-                                lsp::Position {
-                                    line: 0,
-                                    character: 0,
-                                },
-                            end:
-                                lsp::Position {
-                                    line: last_line, ..
-                                },
-                        },
-                    new_text,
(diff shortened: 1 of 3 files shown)


helix-lsp/src/lib.rs
```diff
@@ -221,6 +221,35 @@ pub mod util {
...
+                                    character: 0,
+                                },
+                            end:
+                                lsp::Position {
+                                    line: last_line, ..
+                                },
+                        },
+                    new_text,
+                }] => {
+                    if *last_line as usize >= old.len_lines() - 1 {
```
the-mikedavis, line 243: This condition can be a [match guard](https://doc.rust-lang.org/reference/expressions/match-expr.html#match-guards). That will remove the inner `if` nesting

helix-term/src/commands.rs
```diff
@@ -2326,10 +2326,17 @@ async fn make_format_callback(
         let view_id = view!(editor).id;
         if let Some(doc) = editor.document_mut(doc_id) {
             if doc.version() == doc_version {
-                doc.apply(&Transaction::from(format), view_id);
-                doc.append_changes_to_history(view_id);
-                if let Modified::SetUnmodified = modified {
-                    doc.reset_modified();
+                if let Some(text) = format.replaces_document(doc.text()) {
```
archseer, line 2329: I think we shouldn't do any heuristics here and instead just always re-detect line endings on formatting callbacks.

archseer: I wouldn't call the same code as `reload()` and instead just use `self.detect_indent_and_line_ending();`

farwyler: changed by 0851f74e76c9031a4f0c9458a311bc0e6b96a7b3 to do it like that

PR #2780, opened 2022-06-15
Add clojure language support

Axot017:
This adds support for clojure programming language: tree-sitter, basic highlights and language server.

Screenshot 2022-06-15 at 12 50 51


runtime/queries/clojure/highlights.scm
```diff
@@ -0,0 +1,71 @@
...
+
+(str_lit) @string
+
+(num_lit) @constant.numeric
+
+[(bool_lit) (nil_lit)] @constant.builtin
+
+(comment) @comment
+
+(regex_lit) @string.special
```
the-mikedavis, line 13: We could use `@string.regexp` for this if you like. Or we could have an injection for tree-sitter-regex:

```scm
; runtime/queries/clojure/injections.scm
((regex_lit) @injection.content
 (#set! injection.language "regex"))
```

languages.toml
```diff
@@ -1463,3 +1463,17 @@ indent = { tab-width = 2, unit = "  " }
 [[grammar]]
 name = "prisma"
 source = { git = "https://github.com/victorhqc/tree-sitter-prisma", rev = "17a59236ac25413b81b1613ea6ba5d8d52d7cd6c" }
+
+[[language]]
+name = "clojure"
+scope = "source.clojure"
+injection-regex = "clojure"
```
the-mikedavis, line 1470:
```suggestion
injection-regex = "(clojure|clj)"
```

runtime/queries/clojure/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
...ome|some\\-fn|some\\?|sort|sort\\-by|sorted\\-map|sorted\\-map\\-by|sorted\\-set|sorted\\-set\\-by|sorted\\?|special\\-symbol\\?|spit|split\\-at|split\\-with|str|string\\?|struct|struct\\-map|subs|subseq|subvec|supers|swap!|swap\\-vals!|symbol|symbol\\?|tagged\\-literal|tagged\\-literal\\?|take|take\\-last|take\\-nth|take\\-while|tap>|test|the\\-ns|thread\\-bound\\?|to\\-array|to\\-array\\-2d|trampoline|transduce|transient|tree\\-seq|true\\?|type|unchecked\\-add|unchecked\\-add\\-int|unchecked\\-byte|unchecked\\-char|unchecked\\-dec|unchecked\\-dec\\-int|unchecked\\-divide\\-int|unchecked\\-double|unchecked\\-float|unchecked\\-inc|unchecked\\-inc\\-int|unchecked\\-int|unchecked\\-long|unchecked\\-multiply|unchecked\\-multiply\\-int|unchecked\\-negate|unchecked\\-negate\\-int|unchecked\\-remainder\\-int|unchecked\\-short|unchecked\\-subtract|unchecked\\-subtract\\-int|underive|unquote|unquote\\-splicing|unreduced|unsigned\\-bit\\-shift\\-right|update|update\\-in|update\\-proxy|uri\\?|use|uuid\\?|val|vals|var\\-get|var\\-set|var\\?|vary\\-meta|vec|vector|vector\\-of|vector\\?|volatile!|volatile\\?|vreset!|with\\-bindings\\*|with\\-meta|with\\-redefs\\-fn|xml\\-seq|zero\\?|zipmap)$"))
```
the-mikedavis, line 48: For these two, we'll need to duplicate the stanzas and replace the `list_lit` with `anon_fn_lit` to highlight macros and built-ins within anonymous functions, for example:

```scm
(anon_fn_lit
 .
 (sym_lit) @function.macro
 (#match? @function.macro "^(\\.|\\.\\.|\\->|\\->>|amap|and|areduce|as\\->|assert|binding|bound\\-fn|case|catch|comment|cond|cond\\->|cond\\->>|condp|delay|do|doseq|dosync|dotimes|doto|extend-protocol|extend-type|finally|fn|fn\\*|for|future|gen-class|gen-interface|if|if\\-let|if\\-not|if\\-some|import|io!|lazy\\-cat|lazy\\-seq|let|letfn|locking|loop|memfn|monitor\\-enter|monitor\\-exit|or|proxy|proxy-super|pvalues|quote|recur|refer\\-clojure|reify|set!|some\\->|some\\->>|sync|throw|time|try|unquote|unquote\\-splicing|var|vswap!|when|when\\-first|when\\-let|when\\-not|when\\-some|while|with\\-bindings|with\\-in\\-str|with\\-loading\\-context|with\\-local\\-vars|with\\-open|with\\-out\\-str|with\\-precision|with\\-redefs)$"))
```

We can also highlight anonymous function positional arguments with something like

```scm
((sym_lit) @operator
 (#match? @operator "^%"))
```

I'm not sure what the best scope to use for those is 🤔

runtime/queries/clojure/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+
+;; interop-ish
+(list_lit
+ .
+ (sym_lit) @function.method
+ (.match? @function.method "^\\."))
+
+;; other symbols with dots
+((sym_lit) @variable
+ (.match? @variable "\\."))
```
the-mikedavis, line 63:
```suggestion
;; other symbols
(sym_lit) @variable
```

This highlight can act as a fallthrough for any identifiers not marked above

runtime/queries/clojure/highlights.scm
```diff
@@ -0,0 +1,69 @@
...
+
+;; quote
+(quoting_lit) @constant.character.escape
+
+;; syntax quote
+(syn_quoting_lit) @constant.character.escape
```
the-mikedavis, line 69: There are some extra highlights we could add like

```scm
["{" "}" "(" ")" "[" "]" "#"] @punctuation.bracket
["~" "~@" "#'" "@"] @operator
```

for the punctuation and some of the quoting stuff

Axot017:
Changes made. Thanks for review!

the-mikedavis (approved):
Looks perfect, thanks for working on this!

PR #2781, opened 2022-06-15
Add migrate from Vim wiki link

tennix:
Now helix has its own wiki for [Migrating from Vim](https://github.com/helix-editor/helix/wiki/Migrating-from-Vim), let's add the link to the book.


the-mikedavis (approved):
Good call, thanks!

PR #2790, opened 2022-06-16
Format keys identically in statusline and command palette

sudormrfbin:
The command palette previously used `+` as a delimiter for denoting a single key in a key sequence, (like `C+w`). This was at odds with how the statusline displayed them with pending keys (like `<C-w>`). This patch uses the `<C-w>` formatting everywhere.


the-mikedavis (approved):
👍 I like the consistency

Plus the doctests are lovely 😍

Issue #506, opened 2021-07-25
Create something akin to `vimtutor` for Helix


Omnikar:
This would introduce the user to modal editing, and to Helix's commands and to its multiple selections. I think it could be opened via something like `hx --tutor`, or opened from within the editor with something like `:tutor`.

python128:
VIm tutor is basically a text file which contains explanations of keys, and some examples to work on, right?
If so, can I have a go at this?

Omnikar:
Yes, `vimtutor` is just a text file separated into lessons which interactively introduce the keys and concepts, and each of which is sized to fit on an 80x24 terminal screen.

aeosynth:
Kakoune has something similar: https://github.com/mawww/kakoune/blob/master/contrib/TRAMPOLINE

Omnikar:
> Kakoune has something similar: https://github.com/mawww/kakoune/blob/master/contrib/TRAMPOLINE

I like that file header, and sorta wanna steal that idea…
```
 _   _                 __    __
| | | |         _      \ \  / /
| |_| |   ___  | | (_)  \ \/ /
|  _  |  / _ \ | |  _    )  (
| | | | |  __/ | | | |  / /\ \
|_| |_|  \___| |_| |_| /_/  \_\
```

Omnikar:
> Kakoune has something similar: https://github.com/mawww/kakoune/blob/master/contrib/TRAMPOLINE

I know this doesn't have to do with the tutorial itself, but what if we were to copy that title page idea?
```
 _   _                 __    __
| | | |         _      \ \  / /
| |_| |   ___  | | (_)  \ \/ /
|  _  |  / _ \ | |  _    )  (
| | | | |  __/ | | | |  / /\ \
|_| |_|  \___| |_| |_| /_/  \_\
```

python128:
> > Kakoune has something similar: https://github.com/mawww/kakoune/blob/master/contrib/TRAMPOLINE
> 
> I know this doesn't have to do with the tutorial itself, but what if we were to copy that title page idea?
> 
> ```
>  _   _                 __    __
> | | | |         _      \ \  / /
> | |_| |   ___  | | (_)  \ \/ /
> |  _  |  / _ \ | |  _    )  (
> | | | | |  __/ | | | |  / /\ \
> |_| |_|  \___| |_| |_| /_/  \_\
> ```

Actually a good idea, but I would suggest using some other format, rather than the exact one used by Kakoune...

[...]

PR #2791, opened 2022-06-16
Restore section spacing in tutor.

n0s4:
The amount of empty lines between sections were truncated to 3 lines in #2590, but they were already deliberately spaced out by @Omnikar such that a section would take up exactly a 80x24 terminal screen (so 22 lines per section if you account for the status line).

Apparently @Omnikar approved of that change change though they didn't make any comment the pr itself. If the decision was made to truncate the spacing intentionally then that's fine, I was just confused why it was done.

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -20,7 +20,6 @@ _________________________________________________________________
  the first lesson.
 
 
-
 =================================================================
 =                  1.1 BASIC CURSOR MOVEMENT                    =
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):

the-mikedavis:
Based on https://github.com/helix-editor/helix/issues/506#issuecomment-886439026 I think the spacing is probably Omnikar's preference as well

PR #2793, opened 2022-06-16
Configurable spinners

sbromberger:
This PR adds the ability to customize the progress spinners. It changes the way they are created, as well, but this is not a huge issue because they weren't user-customizable beforehand anyway.

In summary:

The spinner can be defined in the config file using two keys under `[editor]` - `spinner` and `spinner-interval`. The former is a `Vec<String>` whose elements represent frames. The latter is the maximum update time in ms (equivalent to the interval that already exists in the code). Both default to the existing spinner.

This required allowing `EditorView` to take a `ProgressSpinners` argument, so I've changed the calling convention for `EditorView::new`.

If the spinner is multi-frame, then the current behavior where `Instant::now()` is repeatedly called still applies. However, for lower-powered systems, if the spinner is a single frame, then we can elide all the calls (and resulting syscalls) and save some performance.

The spinner can now be styled using `ui.spinner` in the theme. This allows the use of `rapid_blink`, for example, to simulate motion. (However - If the statusbar is non-black, I don't know yet how to make the background persist when the blink state goes to black.)

Anyway - please let me know what you think. I'm still not sure why we're using ProgressSpinners as a collection when it appears there's only ever one spinner. Could we simplify this?


helix-term/src/ui/spinner.rs
```diff
@@ -21,9 +29,9 @@ impl Default for Spinner {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Spinner {
-    frames: Vec<&'static str>,
+    frames: Vec<char>,
```
BearOve, line 34: This will limit the possible characters afaik. Some unicode is more than one char. This might not a problem though.

helix-term/src/ui/spinner.rs
```diff
@@ -32,7 +40,8 @@ pub struct Spinner {
 impl Spinner {
     /// Creates a new spinner with `frames` and `interval`.
     /// Expects the frames count and interval to be greater than 0.
-    pub fn new(frames: Vec<&'static str>, interval: u64) -> Self {
+    pub fn new(framestring: &str, interval: u64) -> Self {
+        let frames: Vec<char> = framestring.chars().collect();
```
BearOve, line 44: This will split a single character into two different ones given a certain input due to graphemes. Not sure how this is normally handled by helix.

helix-term/src/ui/spinner.rs
```diff
@@ -46,22 +55,25 @@ impl Spinner {
...
+            (self
+                .start
+                .map(|time| Instant::now().duration_since(time))?
+                .as_millis()
+                / self.interval as u128) as usize
+                % self.count
+        } else {
+            self.start.and(Some(0))?
+        };
+        Some(self.frames.get(idx)?.to_string())
```
BearOve, line 76: Is creating a string every time that much cheaper than the old method? At least on windows, allocation is slow due to security features.

helix-term/src/ui/spinner.rs
```diff
@@ -46,22 +55,25 @@ impl Spinner {
     }
 
     pub fn dots(interval: u64) -> Self {
-        Self::new(vec!["⣾", "⣽", "⣻", "⢿", "⡿", "⣟", "⣯", "⣷"], interval)
+        Self::new("⣾⣽⣻⢿⡿⣟⣯⣷", interval)
```
BearOve, line 58: Maybe this should be a const on the spinner type so you don't have to duplicate it elsewhere?

BearOve (commented):
I added some comments, but they are without authority, and only my personal feedback.

helix-term/src/ui/spinner.rs
sbromberger: You may be right, but I was under the impression that a `char` was a unicode code point - so you should be able to use a single unicode code point for each frame, regardless of the number of ("traditional char") bytes.

helix-term/src/ui/spinner.rs
sbromberger: Again, I think `chars()` operates over unicode code points, not bytes. The braille characters appear to be two bytes long, in any case.

helix-term/src/ui/spinner.rs
BearOve: It is. I'm talking about grapheme clusters, that may be multiple unicode points.

Don't you love unicode sometimes 😉

helix-term/src/ui/spinner.rs
sbromberger: that's a great point. I was under the impression that `to_string()` was relatively inexpensive, but you're right; it's a lot of allocations. Maybe we should go back to `Vec<&'static str>`.

helix-term/src/ui/spinner.rs
sbromberger: That's a great idea. I guess it can live in `spinner.rs` and then be imported in the other places it's needed.

helix-term/src/ui/spinner.rs
sbromberger: LOL. I hate unicode all the time.

I'm inclined to go back to `Vec<&'static str>` and leave this as an exercise for someone else :)

helix-term/src/ui/spinner.rs
BearOve: Or instead of vec just one one String and iterate it? Then you could iterate grapheme clusters and return &str

BearOve: An indication helix deals with this elsewhere: https://github.com/helix-editor/helix/blob/master/helix-core/src/graphemes.rs

sbromberger: Actually, the original code uses `.copied()` on the string - so I think we're allocating anyway.

helix-term/src/ui/spinner.rs
BearOve: Add it in the impl for the type.


https://doc.rust-lang.org/reference/items/associated-items.html#associated-constants

helix-term/src/ui/spinner.rs
sbromberger: On second thought - I can't find a good way to get `Vec<String>` or `Vec<&'static str>` without running into borrow checker madness. I think if we say, "you can't use grapheme clusters in your custom spinner" it's reasonable, especially since right now you can't customize it at all. What do you think?

BearOve: Vec<'static str> can't be loaded at run-time without a memory-leak so that won't work.

Vec<String> should work just fine however.

helix-term/src/ui/spinner.rs
```diff
@@ -46,22 +55,25 @@ impl Spinner {
...
 
-    pub fn frame(&self) -> Option<&str> {
-        let idx = (self
-            .start
-            .map(|time| Instant::now().duration_since(time))?
-            .as_millis()
-            / self.interval as u128) as usize
-            % self.count;
-
-        self.frames.get(idx).copied()
```
BearOve, line 64: If self.frames is Vec<String> this would be something like

```
self.frames.get(idx).map(|s| s.as_str())
```

And the return value should be &str

helix-term/src/ui/spinner.rs
sbromberger: Vec of what? `String`?

BearOve: It must have disappeared somehow. Yes String

helix-term/src/ui/spinner.rs
BearOve: If it was a Vec of String, this line would be

```
self.frames.get(idx).map(|s| s.as_str())
```

helix-term/src/ui/spinner.rs
sbromberger: Made this a constant in `helix-view` so it could be used in `helix-term`.

helix-term/src/ui/spinner.rs
sbromberger: Moved to `Vec<String>`.

BearOve:
Yes you have. The only thing now is what happens if the frame specified is more than one  visible character wide. However, it might even be a feature to have a 3 character spinner?

I will leave this for someone more familiar with the inner workings of Helix however.

Great work

sbromberger:
> The only thing now is what happens if the frame specified is more than one visible character wide. However, it might even be a feature to have a 3 character spinner?

I tried this last night and it works up to 3-4 chars before it bumps into the filename.

book/src/configuration.md
```diff
@@ -45,6 +45,8 @@ hidden = false
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
+| `spinner` | List of strings comprising the frames to be used for the progress spinner[^1]. | `["⣾","⣽","⣻","⢿","⡿","⣟","⣯","⣷"]` |
```
the-mikedavis, line 48: Let's call the setting `spinner-frames` instead. I think that matches the vector value better than the singular "spinner"

helix-term/src/ui/spinner.rs
```diff
@@ -32,7 +42,7 @@ pub struct Spinner {
 impl Spinner {
     /// Creates a new spinner with `frames` and `interval`.
     /// Expects the frames count and interval to be greater than 0.
-    pub fn new(frames: Vec<&'static str>, interval: u64) -> Self {
+    pub fn new(frames: Vec<String>, interval: u64) -> Self {
         let count = frames.len();
         assert!(count > 0);
         assert!(interval > 0);
```
the-mikedavis, line 48: Now that this is configurable, we should remove these assertions and not render the spinner if either `count` or `interval` is 0.

helix-term/src/ui/spinner.rs
```diff
@@ -32,7 +42,7 @@ pub struct Spinner {
 impl Spinner {
     /// Creates a new spinner with `frames` and `interval`.
     /// Expects the frames count and interval to be greater than 0.
-    pub fn new(frames: Vec<&'static str>, interval: u64) -> Self {
+    pub fn new(frames: Vec<String>, interval: u64) -> Self {
```
the-mikedavis, line 45: In here we could compute the visible width (`helix_core::unicode::width`) of the longest frame, then use that width to tell how large the offset should be for [this call](https://github.com/helix-editor/helix/pull/2793/files#diff-408ac3559e5a464e3cce528ed57e0bf48453d28c95212437790ac5dec0f450beL693-R694) to avoid clipping. We may want to set some sane maximum like `8` so it doesn't take up the whole statusline.

the-mikedavis (commented):
I was curious what you would want to replace the current spinner with but it turns out [there isn't a shortage](https://stackoverflow.com/questions/2685435/cooler-ascii-spinners) of possible spinners 😄

Some of my favorites:

```toml
[editor]
spinner = ["▁", "▂", "▃", "▄", "▅", "▆", "▇", "█", "▇", "▆", "▅", "▄", "▃", "▁"]
spinner = ["▉", "▊", "▋", "▌", "▍", "▎", "▏", "▎", "▍", "▌", "▋", "▊", "▉"]
spinner = ["◴", "◷", "◶", "◵"]
spinner = ["◐", "◓", "◑", "◒"]
spinner = ["◡◡", "⊙⊙", "◠◠"]
```

sbromberger:
> I was curious what you would want to replace the current spinner

Selfishly, I hate things that move erratically - I find them really distracting, and the LSP loading process stops and starts the spinner a few times (there's a CPU lag as well on my lower-performance VM). I replaced it with a single frame of a yellow dot. The single-frame notification also doesn't use `Instant::now()` so I save the syscalls.

But yes - I've been playing around with lots of alternate possibilities.

(Won't somebody think of the poor syscalls?)

helix-term/src/ui/spinner.rs
sbromberger: My hasty opinion is that anyone who wants an 8-character spinner deserves what they see. I think we're good to 4 characters. :)

book/src/configuration.md
sbromberger: Done!

helix-term/src/ui/spinner.rs
sbromberger: Great idea. Done.

Woolworths:
Does it make sense for spinner-interval set to 0 to disable spinner display? We already have this functionality if spinner-frames is set to [].

Would it not make more sense to change this functionality to keep the previous logic (update when a key-press is made, instead of just removing the spinner)? I prefer this logic, because it prevents a render every time the spinner updates (which is 👍 if you're on a slow connection/ssh'd into a server).

sbromberger:
> Would it not make more sense to change this functionality to keep the previous logic (update when a key-press is made, instead of just removing the spinner)? 

I'm not sure I understand. The current code doesn't allow you to configure the spinner interval at all, and if you manage to do it by editing the code, if you set it to zero, helix will panic on `assert`. There's currently no logic that handles "update spinner on keypress".

Because the frame is empty, there will be no significant update - I run helix (and this proposed code) on a small VM over a slow SSH connection and haven't seen any issues. In fact, it's one of the primary reasons I'm submitting this PR, since the current behavior causes the spinner to move erratically.

Woolworths:
Ahhh yes, that makes sense. Thanks for the explanation 👍

sbromberger:
Changing this to WIP pending discussion of #2829

sbromberger:
@Woolworths - just out of curiosity, are you using mosh?

Woolworths:
@sbromberger I'm not, no 😄

sbromberger:
>  While testing this PR I just ran into the situation where the spinner would not disappear, even after rust-analyzer finished loading.

I've noticed this too, but I also noticed it without the PR, so I don't think it's the code here. The spinner indicator updated to the correct state (went away) when I pressed a key.

I didn't change any of the logic relating to spinner start/stop - just how it's displayed.

sbromberger:
Also, it's not consistent. Here's a screen recording I just made with a single-char (yellow dot) spinner:

https://asciinema.org/a/Tjuz8EWfpLZiDHd6A24K1l9sW

Woolworths:
This would be really awesome to see in helix!

sbromberger:
Refactored for new customizable statusline. Width of spinner is now handled appropriately by the statusline!

Ready for review.

aikomastboom:
Been using this branch for some time with a simple one character custom spinner and have seen no obvious glitches.

sbromberger:
Could I get a review on this, please? Thank you :)

helix-term/src/ui/spinner.rs
sbromberger: This should be resolved with the new statusline implementation.

sbromberger:
Merged with master - I think this is ready for review. Thank you.

PR #2798, opened 2022-06-17
Extend Tutor.

n0s4:
In accord with #2785.

If you have a better idea for the order of the blueprint you could comment on that discussion and I'd update it.

TODO:
- [X] Initial re-ordering (no new sections yet.)
- [ ] Select mode.
- [ ] Split selection on lines.
- [ ] More Selection Commands Chapter.
  - [ ] Replace (r).
  - [ ] Flip selection.
  - [ ] Repeat last motion.
  - [ ] Repeat last action.
- [ ] Replace with yanked (R).
- [ ] Increment/Decrement.
- [ ] Registers.
- [ ] More on Searching Chapter.
  - [ ] Searching for selection.
  - [ ] Adding Next Instance of Selection.
- [ ] Advanced Selection Commands Chapter.
  - [ ] Cycling and Removing Selections.
  - [ ] Keep/Remove Selections on Regex.
  - [ ] Split Selection on Regex

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -109,29 +109,7 @@ _________________________________________________________________
        Notice that when you type i, 'NOR' changes to 'INS'.
 
 =================================================================
-=                   1.5 MORE ON INSERT MODE                     =
-=================================================================
-
- As you saw, you can type i to enter Insert mode at the current
- position of the cursor. There are a few other ways you can
- enter Insert mode at different locations.
-
- Common examples of insertion commands include:
-   i - Insert before the selection.
-   a - Insert after the selection. (a means 'append')
-   I - Insert at the start of the line.
-   A - Insert at the end of the line.
-
- 1. Move to anywhere in the line below marked -->.
- 2. Type A (<SHIFT> + a), your cursor will move to the end of
-    the line and you will be able to type.
- 3. Type the text necessary to match the line below.
-
- --> This sentence is miss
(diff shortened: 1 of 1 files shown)


n0s4:
I'm trying out vale.sh as suggested by @David-Else in the [discussion](https://github.com/helix-editor/helix/discussions/2785#discussioncomment-2975864). The Microsoft style package that was suggested is not exactly suited to the format of the tutor, it flags the intentional mistakes in the demos which creates a lot of noise, among other things.

I'm sure this could be worked around with a config but I'd rather do it fully and write a style for the tutor conventions (line length, section sizes, etc.) to make it truly worthwhile. I might try to hack a few of the basic styles for the Tutor in this pr, perhaps a more complete one later. Thanks for the suggestion!

n0s4:
I couldn't think of a solid example/demo for flip selections (`A-;`), I took it as a sign that it probably doesn't need a whole section so I've made it a note under COLLAPSING SELECTIONS.

archseer:
Is this ready for merge? I'm okay with leaving some pending items to a follow-up. I'd like to get this into the 22.07 release at the end of this month :)

n0s4:
Yeah. I'll have a check for any mistakes but I don't think there's much left worth waiting for.

n0s4:
Should be ready now :). @archseer

archseer (approved):
Thanks for the hard work on this!

Issue #2786, opened 2022-06-16
Unable to use Sorbet LSP for Ruby (LSP invalid method call id=Num(0))


klardotsh:
### Summary

Trying to use Sorbet as a Ruby LSP so I can move over from Neovim for work as I already have with my side projects, but alas, I can't get the LSP to work. Here's my `~/.config/helix/languages.toml`:

```toml
[[language]]
name = "ruby"
language-server = { command = "srb", args = ["tc", "--lsp"] }
```

### Reproduction Steps

 

I tried this:

1. `hx -vvvvvvvvvvvvvvvv lib/example/typing_example_3.rb`

I expected this to happen:

1. See the type errors in the file light up, much as they do in eg. Rust projects on this system, and much as they do with Sorbet in my Neovim setup
2. Be able to Space-K to hover a variable and see its type, much as I can do in Rust projects in Helix, and with Sorbet in Neovim

Instead, this happened:

- None of the above
- I'm editing plain old Ruby with syntax highlighting but no LSP integration
- `Language server not active for current buffer` when attempting to Space-K for hover

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-06-15T20:25:55.860 helix_loader [DEBUG] Located configuration folders: ["/home/j/src/redacted/redacted-repo/.helix"]
2022-06-15T20:25:55.937 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2022-06-15T20:25:55.937 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2022-06-15T20:25:55.937 mio::poll [TRACE] registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
2022-06-15T20:25:55.937 helix_lsp::client [INFO] Using custom LSP config: {}
2022-06-15T20:25:55.937 mio::poll [TRACE] registering event source with poller: token=Token(4), interests=READABLE | WRITABLE
2022-06-15T20:25:55.937 mio::poll [TRACE] registering event source with poller: token=Token(5), interests=READABLE | WRITABLE
[...]

PR #2801, opened 2022-06-18
embed jsonrpc types from jsonrpc-core crate

the-mikedavis:
We should not depend on jsonrpc-core anymore:

* The project just announced it's no longer actively maintained[^1],
  preferring their new implementation in `jsonrpsee`.
* The types are too strict: we would benefit from removing some
  `#[serde(deny_unknown_fields)]` annotations to allow language
  servers that disrespect the spec[^2].
* We don't use much of the project. Just the types out of core.
  These are easy to embed directly into the `helix-lsp` crate.

[^1]: https://github.com/paritytech/jsonrpc/pull/674
[^2]: https://github.com/helix-editor/helix/issues/2786

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -421,7 +421,6 @@ dependencies = [
  "futures-executor",
  "futures-util",
  "helix-core",
- "jsonrpc-core",
  "log",
  "lsp-types",
  "serde",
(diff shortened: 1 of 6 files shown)


archseer (approved):
Agreed! 👍

Issue #2721, opened 2022-06-09
Suggestion: use `<space>` to scroll in view mode


benjaminrich:
One thing that's always bugged me is that while `h`, `j`, `k`, `l` are so ergonomic, `<c-f>`, `<c-b>` and `<c-d>`, `<c-u>` are just the opposite; they feel characteristically "un-modal", and I've never understood why I should want to use them.

It's pretty ubiquitous in pagers, viewers and even web browsers to use `<space>` and `<s-space>` to scroll forwards and backwards page-wise, and this is very comfortable. Now, obviously `<space>` has a different usage in normal mode, but in view mode (and in particular 'sticky' view mode) it could be used for scrolling. That would give you the best of both worlds, it seems.

the-mikedavis:
I think this is a good idea. If you'd like to submit a PR, the default keymap for `Z` lives here: https://github.com/helix-editor/helix/blob/0b8a00ac967f4fd0608d83e2804bb6b1a8bf9c5d/helix-term/src/keymap/default.rs#L274-L285

You could bind `space` to `page_down` or maybe `half_page_down` might be smoother. I'm not sure it's possible to bind `S-space` because of either terminal or crossterm limitations. Using the `event-read` example from crossterm:

```
Event: Key(KeyEvent { code: Char(' '), modifiers: NONE })
Event: Key(KeyEvent { code: Char(' '), modifiers: NONE })
```

both space and shift+space register as just space.

benjaminrich:
I will try. Please give me a bit of time, as I have not contributed before. Thanks.

benjaminrich:
Sorry for the naive question, because I am not a real dev and have never worked with rust before, but if I make a change to the code base and want to test it, what do I do? I mean, what are the commands to compile and run my modified version of the code?

the-mikedavis:
Finally a key combination that works on Windows and not Linux 😄

For me the shift modified isn't being recognized on either Kitty or WezTerm

benjaminrich:
Yes. It would be quite a useful key combination, but you cannot rely on it.

PR #2803, opened 2022-06-18
Add new key binding to view mode

benjaminrich:
The idea is to make view mode more pager-like, and to make (half-)page-wise scrolling easier.

In particular, I suggest `space` to scroll down, like most pagers. Unfortunately, `shift-space` cannot be used to scroll up due to terminal limitations, so I suggest `backspace` as an alternative. I also suggest `J` and `K` for downwards and upwards scrolling, respectively. I also added search commands to view mode.

Addresses #2721

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -268,8 +268,13 @@ pub fn default() -> HashMap<Mode, Keymap> {
             "j" | "down" => scroll_down,
             "C-b" | "pageup" => page_up,
             "C-f" | "pagedown" => page_down,
-            "C-u" => half_page_up,
-            "C-d" => half_page_down,
+            "C-u" | "K" | "backspace" => half_page_up,
+            "C-d" | "J" | "space" => half_page_down,
+
+            "/" => search,
+            "?" => rsearch,
+            "n" => search_next,
+            "N" => search_prev,
         },
         "Z" => { "View" sticky=true
             "z" | "c" => align_view_center,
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I'm ok with `space` and `backspace` but I think the search bindings and `J` and `K` should also go. `J` and `K` are not actually improvents on `C-u`/`C-d` (shift and control both count as one keystroke) and duplicate keybindings are not good for the default keymap. You can add these to your own keymap with config:

```toml
# ~/.config/helix/config.toml
[keys.normal]
Z = { J = "half_page_down", K = "half_page_up" }
```

benjaminrich:
That's fair. It was just a suggestion. Default key bindings are obviously a tricky issue. I personally feel that `J` and `K` are logical and much more comfortable and intuitive to type because of their proximity to each other and the fact that they are on the home row, but as you point out I can always make the decision for myself it I want.

I don't see why I shouldn't be able to search in view mode though. That seems kind of unnecessarily restrictive (unless I am missing the point of view mode).

the-mikedavis:
Hmm yeah actually I agree, search bindings should be ok to make the pager-like experience better 👍

sbromberger:
FWIW, I'm against binding `space`. I use it all the time - even (especially?) in view mode, with `k`, to ensure that the thing I'm selecting is really the thing I want to be doing something with.

the-mikedavis:
How are you using space currenty? It isn't bound in view mode at the moment so it's a no-op

sbromberger:
> How are you using space currenty? It isn't bound in view mode at the moment so it's a no-op

Sorry, this may be my misunderstanding. Is "view mode" the same as select mode (`v`), or is it something else?

the-mikedavis:
View mode is a minor mode you can use with `z` or a sticky minor mode (you stay in the mode even after pressing keys) with `Z` that lets you move the view of the document.

sbromberger:
Ah, then my apologies. I withdraw my concern :)

the-mikedavis (approved):

PR #2807, opened 2022-06-18
text-objects: add test capture & elixir queries

connorlay:
See https://github.com/helix-editor/helix/discussions/2792


the-mikedavis:
This is looking good :)

I might push a commit or two with queries for more languages if that's alright

connorlay:
> I might push a commit or two with queries for more languages if that's alright

Go for it! Which languages did you have in mind?

the-mikedavis:
I was thinking at least Rust but also Erlang for my sake 😄

For Rust, we can capture the function following the `#[test]` attribute.

For a few of languages (Erlang EUnit tests and Gleam Gleeunit for example) we could have the captures be pretty much the same as `@function.*` but only for those functions with a name that have a name that ends in `_test`.

connorlay:
@the-mikedavis What do you think about queries for languages that have a number of popular test frameworks, like JavaScript?

This leads me to wonder if Helix should support some kind of opt-in queries for specific frameworks? Highlights and textobjects for the Phoenix router DSL would be handy to have available, but I don't think belongs in the core Elixir queries.

the-mikedavis:
Hmm I think we should aim to only support official language tooling in the core and leave framework-specific features to plugins (once implemented). Especially if two frameworks conflict, it's hard to be rigorous about which one should get support.

For this feature in particular though I don't see any harm in supporting javascript testing frameworks for example

archseer (approved):

the-mikedavis (approved):
Thanks for working on this @connorlay, this is great!

Issue #703, opened 2021-09-03
Marks


Omnikar:
Like Vim's `m`. You can record a mark of the current position of the primary cursor in a register, and it will remember the file, line, and column. You can then return to the mark later.

pickfire:
I wonder if this will intersect with our current `m` since vim `m` can take all the keys.

archseer:
Kakoune uses `z` but we have view mode there. https://github.com/mawww/kakoune/blob/master/doc/pages/keys.asciidoc#marks

@oberblastmeister was interested in this, we discussed about storing these as ranges (outside registers) so we can map them through text edits. That way they behave similar to selection ranges and stay in the same place as the document changes. It's a prerequisite for snippet support.

pppKin:
I really like the idea(the revised one), and I imagine once this gets implemented, it would be much easier to implement snippet support then.
Say we have a snippet:
`
Hello $1, good ${2:morning}!
`
which to give us something like 
`
Hello , good morning!
`
We can first select `$1`(maybe automatically enter edit mode if it does not have a default value). After that user can "jump to the next mark" and have `morning` selected. A really smooth and unified user experience;)

archseer:
Yeah exactly, this was the intent :) It's similar to neovim's proposed `extmark` except we're using ranges. (Neovim had trouble keeping the marks if there was any edits, which eventually led to me giving up on [snippets.nvim](https://github.com/archseer/snippets.nvim/) and starting work on Helix)

It shouldn't be too hard to do, was thinking we might be able to abstract selections into a type of mark.

pianocomposer321:
Maybe this is a stupid question (I have not been following the development of helix until recently), but using marks for snippets wouldn't collide with user-defined marks, would it? Would there be separate concepts of user-defined marks and "other" marks, which would be used for things like snippets (essentially like neovim extmarks), but they would use the same stuff under the hood? Is that an accurate understanding?

archseer:
I imagine they'd be namespaced under and `id`. "user" vs "snippet" etc

pianocomposer321:
So it's not that these will be like vim marks, but also used for snippets etc., but more that they will be like neovim extmarks, and will be used for user-defined marks (similar to vim marks) as well as other things like snippets. Is that right?

archseer:
Yep, exactly.

the-mikedavis:
[...]

PR #2808, opened 2022-06-19
[FEAT] Create a cursor on jumplist entry

carrascomj:
Related to https://github.com/helix-editor/helix/discussions/2784#discussioncomment-2967254

This PR adds a command `copy_cursor_backward` that creates a new cursor by jumping backward in the jumplist, preserving the previous selection. This adds a flexible way of editing text with multiple cursors, especially when combined with textobjects motions. 

### Demo

Save selection to jumplist `C-s`, jump to function `]f`, copy cursor backward `C-q`:
copy_back

### TODO

- [ ] Keymap, I have assigned `C-q` for no particular reason.
- [ ] Docs.
- [ ] Copy cursor forward comand? The implementation is trivial but I don't see a lot of purpose in that.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -356,8 +356,9 @@ impl MappableCommand {
         select_prev_sibling, "Select the previous sibling in the syntax tree",
         jump_forward, "Jump forward on jumplist",
         jump_backward, "Jump backward on jumplist",
-        save_selection, "Save the current selection to the jumplist",
+        copy_cursor_forward, "Place a new cursor on forward selection of the jumplist",
         copy_cursor_backward, "Place a new cursor on backward selection of the jumplist",
+        save_selection, "Save the current selection to the jumplist",
         jump_view_right, "Jump to the split to the right",
         jump_view_left, "Jump to the split to the left",
         jump_view_up, "Jump to the split above",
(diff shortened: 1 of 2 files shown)


EpocSquadron:
This is a pretty neat idea! What happens if the jumplist switches to another buffer? I would argue the forward equivalent isn't a bad idea to include, if you overshoot the jumplist and need to go forward.

carrascomj:
> What happens if the jumplist switches to another buffer?

It panics :sweat_smile:. Sorry, I hadn't checked, I will look into into it and maybe add a test if I figure out how to cover that.

Regarding the forward variant, I have tried and it is fine. I would not use it a lot but if this is merged, a user will ask for it at some point; I think I'll add it.

the-mikedavis:
Keymap-wise, I'm not a big fan of `C-q`. It looks like `C-m` is interpreted by crossterm as `Enter` so that's out. My terminal emulator is eating `C-O` altogether but maybe that's just my config. I'll think about it some more. Another option is to not add a default keybind and let people create a keybind if they want this.

For docs, I've been meaning to add a section to `book/src/usage.md` on the jumplist. This could fit there and possibly in the tutor (`runtime/tutor.md`) as well.

Forward variant: we could implement it and not bind it if we don't think it will be useful. That way someone can bind it themselves if they find a use for it without needing to change the code.

carrascomj:
I have fixed the problems with jumping to another buffer, removed the keymap (`C-O` was not working either for me) and added the forward variant without a keymap.

Anyways, this PR may not be worth it if marks are a desired feature.

helix-term/src/commands.rs
```diff
@@ -3858,32 +3860,46 @@ fn match_brackets(cx: &mut Context) {
 
 //
 
-fn jump_forward(cx: &mut Context) {
+fn jump_impl(cx: &mut Context, copy_cursor: bool, direction: Direction) {
     let count = cx.count();
-    let view = view_mut!(cx.editor);
+    let (view, doc) = current!(cx.editor);
+    let mut this_selection = doc.selection(view.id).clone();
```
carrascomj, line 3866: This clone is not required if `copy_cursor` is true, could it be avoided in some elegant way?

the-mikedavis:
> this PR may not be worth it if marks are a desired feature

yeah I was thinking similar: marks seem to be a cleaner approach to this, especially if we had the rich operations from kakoune's `A-z`/`A-Z` https://github.com/mawww/kakoune/blob/6565f6edd700a1490c670f4e60cd52f7f78dbbe5/doc/pages/keys.asciidoc#marks

It looks like @archseer is interested in tackling marks because of some prior experience with it in nvim but it might not be a priority for him at the moment #703

PR #2814, opened 2022-06-19
Reuse menu::Item trait in picker

sudormrfbin:
This opens the way for merging the menu and picker code in the future, since a picker is essentially a menu + prompt. More excitingly, this change will also allow aligning items in the picker, which would be useful (for example) in the command palette for aligning the descriptions to the left and the keybinds to the right in two separate columns. This requires more work though, and this PR only lays the foundations for it.

The item formatting of each picker has been kept as is, even though there is room for improvement now that we can format the data into columns, since that is better tackled in a separate PR.


helix-term/src/ui/menu.rs
```diff
@@ -14,22 +16,38 @@ use helix_view::{graphics::Rect, Editor};
 use tui::layout::Constraint;
 
 pub trait Item {
-    fn label(&self) -> &str;
+    /// Additional editor state that is used for label calculation.
+    type EditorData;
```
archseer, line 20: Nit: call it Data or maybe State?

helix-term/src/ui/completion.rs
```diff
@@ -15,19 +15,28 @@ use helix_lsp::{lsp, util};
...
-    fn sort_text(&self) -> &str {
-        self.filter_text.as_ref().unwrap_or(&self.label).as_str()
+    type EditorData = ();
+    fn sort_text(&self, _data: &Self::EditorData) -> Cow<str> {
+        self.filter_text
+            .as_ref()
+            .unwrap_or(&self.label)
+            .as_str()
+            .into()
     }
```
archseer, line 25: Could call the `filter_text` method and mark as inline so there's no code duplication

archseer (commented):
Great change, I attempted this before but didn't think of using an associated type for the external data.

archseer (approved):
Great work!

archseer:
Ah wait, didn't see the conflicts :/

PR #2829, opened 2022-06-20
Allow EditorView access to config, plus add file modification indicator configuration option

### LINKED ISSUE #2828: [dev] EditorView doesn't have an easy way to access Editor config elements
Consider the case where you have an editor config item that is accessed from within an `EditorView` method.

Right now we're having to extract individual config items out of the editor config prior to it being altered in https://github.com/helix-editor/helix/blob/ad15e7b5e8918e1521683c7f5c63d77e00c58023/helix-term/src/application.rs#L120, and then we pass in the extracted items to a modified `EditorView::new()` function (with the params as arguments to `new()`, which means every time we want to add a new config item that is handled by an `EditorView`, we have to change the signature of this constructor).

Is it possible to pass (a reference to) the entire editor config to an `EditorView` so that we have access to config variables within `EditorView` methods?

sbromberger:
- provides visibility of the configuration to EditorView (closes #2828)
- adds a `file-modification-indicator` config option to change the string displayed
when file is modified.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -45,6 +45,7 @@ hidden = false
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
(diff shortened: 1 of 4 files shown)


sbromberger:
I have tested the following:
- no configuration option yields the standard (existing) `[+]` indicator.
- configuration options work as expected.
- updating the configuration and reloading will automatically update any indicators on open buffers (due to the `Arc` usage).

sbromberger:
Closing in favor of #2831.

PR #2832, opened 2022-06-20
Ignore non-number suffix or prefix when in(/de)crementing

### LINKED ISSUE #2815: Increment and decrement text objects
### Summary

If a number isn't solely present then the increment & decrement operators don't even work.
```
1px 2 px
```
I was just editing a CSS file and first of all I have to place my cursor over the number and then use the increment or decrement operation. And if the word is alphanumeric then I have to place spaces around the number then place my cursor back on the number and now the increment and decrement operations work. I'm not sure if that's how it's intended to work but my complain about increment and decrement operation was quite old. In almost 90% of the cases I have a non-numeric character along with the number I'd like to increment.

Is this supposed to be fixed? I've tried it in helix from Arch as well as the one I built from source.

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.05-dev (595213ca)

A-Walrus:
Ignores suffixes to numbers such as "px" when incrementing a word like "1px".
Fixes #2815

### DIFF
--- a/helix-core/src/increment/number.rs
+++ b/helix-core/src/increment/number.rs
@@ -58,33 +58,32 @@ impl<'a> NumberIncrementor<'a> {
         let number = if prefixed {
             word[2..].replace('_', "")
         } else {
-            if !word_range.contains_range(&range) {
+            if !word_range.contains(range.head) {
                 return None;
             }
 
             let len = word_range.len();
 
             // Offsets of selection within word
-            let start_offset = range.from() - word_range.from();
-            let end_offset = range.to() - word_range.from();
+            let offset = range.head - word_range.from();
 
             // Find end of number
             let end = word
                 .chars()
                 .enumerate()
-                .skip(end_offset)
+                .skip(offset)
                 .find(|(_, c)| !(c.is_ascii_digit() || *c == '_'))
                 .map(|(i, _)| i)
                 .unwrap_or(len);
 
-            let (start_byte, _) = word.char_indices().nth(start_offset)?;
+            let (start_byte, _) = word.char_indices().nth(offset)?;
             // Find start of nubmer
             let start = word[..start_byte]
                 .chars()
                 .rev()
                 .enumerate()
                 .find(|(_, c)| !(c.is_ascii_digit() || *c == '-' || *c == '_'))
-                .map(|(i, _)| start_offset - i)
+                .map(|(i, _)| offset - i)
                 .unwrap_or(0);
 
             if end < start {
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):

A-Walrus:
I agree. Prefixes should be ignored as well. There are some strange cases, like decrementing `num1` twice, should it add a minus sign within your variable name?

sudormrfbin:
Vim

A-Walrus:
Is there any reason this isn't merged yet? Something I need to change? It works with both prefixes and suffixes (and both) now.

the-mikedavis:
It's just waiting on review. I just gave it a test with prefixes, suffixes and prefixes and suffixes and it looks good on my end 👍

sudormrfbin:
There seems to be a bug where placing the cursor _after_ the number and pressing `C-a` or `C-x` changes the number. Also if there are two numbers in inside a word (like `word66word99word` neither of the commands work on either of the numbers.

the-mikedavis (commented):
it looks like there's a little regression with the jumping behavior:

```toml
#[|]#indent = { tab-width = 4, unit = "    " }
```

jumps with `C-a` to

```toml
#[indent = { tab-width = 4|]#, unit = "    " }
```

(`#[` is anchor and `|]#` is head)

but with this change we don't increment it as we do on master

A-Walrus:
It seems like strange behavior to me to expand the selection to the next number when incrementing. I understand that this is like how it works in Vim, but it seems odd to me. I propose changing the functionality so that trying to increment something that isn't a number or a date will have no effect on the text and no effect on the selection.

the-mikedavis:
Yeah, we may want to split the functionality so we have a textobject that goes to the next number and then `C-a`/`C-x` increment and decrement the selection. But that should probably live in a separate PR from this fix PR

helix-core/src/increment/number.rs
```diff
@@ -56,9 +55,53 @@ impl<'a> NumberIncrementor<'a> {
             (10, false)
         };
 
-        let number = if prefixed { &word[2..] } else { &word };
+        let number = if prefixed {
+            word[2..].replace('_', "")
+        } else {
+            if !word_range.contains(range.head) {
```
aikomastboom, line 61: This check is false, when using single digit numbers.

aikomastboom: having a closer look.. with a debugger.. it appears the running application does not send `Range::point` ranges.. but real ranges.. a cursor on `point(0)` results in a `range(0,1)`.

rcorre:
What happened? It looks like this was approved, then closed.

PR #2836, opened 2022-06-20
fix: do not color health summary when stdout is piped

### LINKED ISSUE #2833: Health uses colors even when not outputting to tty
### Summary

When running helix with "--health" it outputs data with ANSI colors, even if piping into another process, or redirecting into a file. This makes it harder to use/parse the output. Helix should check if it is outputting to a tty, and only then use colors.

### Reproduction Steps

 

I tried this:

1. `hx --health > file.txt`

I expected this to happen:
file.txt contains the health data without ANSI colors.
Instead, this happened:
file.txt contains the health data with ansi colors.

### Helix log

```
```

</details>

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.05

lazytanuki:
Closes #2833

The [is-terminal crate](https://crates.io/crates/is-terminal) seems to be more active than the [atty crate](https://crates.io/crates/atty). A crate should be necessary for cross platform compatibility (it's a bit messier to check if stdout is piped on Windows than on Linux).

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -65,8 +65,6 @@ serde = { version = "1.0", features = ["derive"] }
 grep-regex = "0.1.9"
 grep-searcher = "0.1.8"
 
-is-terminal = "0.3.0"
-
 # Remove once retain_mut lands in stable rust
 retain_mut = "0.1.7"
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -201,27 +201,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "errno"
-version = "0.2.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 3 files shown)


helix-term/Cargo.toml
```diff
@@ -65,6 +65,8 @@ serde = { version = "1.0", features = ["derive"] }
 grep-regex = "0.1.9"
 grep-searcher = "0.1.8"
 
+is-terminal = "0.3.0"
```
archseer, line 68: Please use the method available in crossterm rather than importing another dependency https://docs.rs/crossterm/latest/crossterm/tty/trait.IsTty.html

archseer (changes requested):

helix-term/Cargo.toml
lazytanuki: I didn't realize crossterm had implemented this. Thanks !

the-mikedavis (approved):

PR #2838, opened 2022-06-20
rewrite language configuration docs

the-mikedavis:
This change moves the configuration tables from the Adding
Languages guide into the overall Languages section. It also
adds more detailed documentation on the `language-server`
configuration key and fixes a typo in the "mylang" example
(the scope was `scope.mylang` instead of `source.mylang`).


archseer (approved):
This is great 👍

Issue #395, opened 2021-06-30
snippet support


skewballfox:
not exactly a feature request, more of request to implement a feature (or whether to wait until extensions are supported). 

I've been working on what I hope will be a cross editor snippet manager ([link to project](https://github.com/skewballfox/sniper)) where the client and server are written in rust. The syntax I'm working to support is a superset of the [syntax defined in the lsp specification](https://github.com/microsoft/language-server-protocol/blob/main/snippetSyntax.md)

while I'm looking to support a few features not covered in the specification (hoping to achieve feature parity with [ultisnips](https://github.com/SirVer/ultisnips) eventually), a major usecase is reducing the amount of effort to support snippets by emerging editors (similar to the benefit of language servers)

the project is at the point to where I can get completions per keystroke and insert/navigate snippets in vscode, but I'm also depending on features specific to vscode. Working to support vscode first is also slowing me down because I really dislike/suck with javascript. 

I'm aware that [extension support is ongoing](https://github.com/helix-editor/helix/issues/122), but this also might be supportable by the editor itself (I'm trying to make it to where client implementation should be very little code) and also would make sense to a certain degree given that, while no language server (that I know of) handles snippets, snippets are a part of the LSP specification.

CBenoit:
Hi! `sniper` sounds like a very interesting project!

I assume @archseer's intention was to support LSP snippets natively so I don't know if `sniper` support will be a built-in feature (it is also obviously very young). At any rate, I hope we can at least see a plugin for that at some point hence the `A-plugin` label.

Regarding an eventual built-in support, if @archseer is favorable, you may want to get involved with helix development. Don't hesitate to drop in for a chat at [Helix community Helix Space](https://github.com/helix-editor/helix#getting-help) :smile:

EpocSquadron:
I think it's worth looking at [the-way](https://github.com/out-of-cheese-error/the-way) for inspiration.

theowenyoung:
Strongly need this feature.

Uldgiw:
There is still one things bothering me right now about the snippet support:
The snippets are previewed in the file rather than in the completion/documentation area of the editor.
While tabbing through completion candidates this is rather distracting. 

I don't know how hard this is to implement, but i at least wanted to let others know.
But all in all the snippet support is shaping up to be very good, keep up the amazing work👍

skewballfox:
> * write a simple language server: https://github.com/quantonganh/snippets-ls

[...]

PR #2852, opened 2022-06-21
Add support for digraphs

### LINKED ISSUE #1438: Digraphs and Unicode input tools
In vim you can input some unicode characters by pressing Ctrl + k in insert mode and then typing to chars like *l for λ. There are other ways to, but this is very convenient and customizable.

Lindenk:
Closes #1438

The backend is implemented as a trie and finds suggestions using a simple breadth first search starting at the node given by the user's input. I chose the default keybind of `Ctrl-K` because while vim uses `Ctrl-k`, in helix that's currently bound to `kill_to_line_end`. This implementation is to set up initial, usable support for digraphs and can later be improved with:

- Better suggestions (currently `Prompt` doesn't allow for extra text in suggestions, so that would need to exist first)
- Fuzzy find on both input sequences and descriptions
- An sane, agreed upon default set of digraphs (currently only user configured symbols are supported)
- Configurable auto-input once a digraph is matched, and there are no more options the input could represent
- `replace`, `append`, and `insert` variants for binding to other keys in other modes as needed

https://user-images.githubusercontent.com/2438655/174872719-630eaa61-118e-41e1-a27c-a336c87a8514.mp4

### DIFF
--- a/book/src/SUMMARY.md
+++ b/book/src/SUMMARY.md
@@ -13,6 +13,7 @@
   - [Key Remapping](./remapping.md)
   - [Hooks](./hooks.md)
   - [Languages](./languages.md)
+  - [Digraphs](./digraphs.md)
 - [Guides](./guides/README.md)
   - [Adding Languages](./guides/adding_languages.md)
   - [Adding Textobject Queries](./guides/textobject.md)
--- a/book/src/digraphs.md
+++ b/book/src/digraphs.md
@@ -1,0 +1,14 @@
+# Digraphs
+
+By default, special characters can be input using the `insert_digraphs` (or `C-K`) command while in insert mode. Custom digraphs can be added to the `editor.digraphs` section of the config:
+
+```toml
+[editor.digraphs]
+ka = "か"
+ku = { symbols = "く", description = "The japanese character Ku" }
+shrug = "¯\_(ツ)_/¯"
+```
+
+## Defaults
+
+TBD
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -9,38 +9,38 @@
 
 > NOTE: Unlike vim, `f`, `F`, `t` and `T` are not confined to the current line.
 
-| Key                   | Description                                        | Command                     |
-| -----                 | -----------                                        | -------                     |
-| `h`, `Left`           | Move left                                          | `move_char_left`            |
-| `j`, `Down`           | Move down                                          | `move_line_down`            |
-| `k`, `Up`             | Move up                                            | `move_line_up`              |
-| `l`, `Right`          | Move right                                         | `move_char_right`           |
-| `w`                   | Move next word start                               | `move_next_word_start`      |
-| `b`                   | Move previous word start                           | `move_prev_word_start`      |
-| `e`                   | Move next word end                                 | `move_next_word_end`        |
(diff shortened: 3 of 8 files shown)


thomkeh:
Is [digraph](https://en.wikipedia.org/wiki/Digraph_(orthography)) the right name for this feature? I guess the name comes from using, for example, ctrl-K and "ae" to get the symbol æ, which is sort of a digraph. But if you use this to write Hiragana, I'd say it's more of a general character input method.

Lindenk:
I just copied nvim on this one, and they also use it for Hiragana. There's also plugins like [better digraphs](https://github.com/protex/better-digraphs.nvim) which turns it into a general character lookup.

I would assume most vim users would look for "digraph" as the name for this feature even though it's not a very accurate description.

If it should be changed, what name should should we go with?

EpocSquadron:
We get occasional requests for this for one of the programming languages that uses mathematical symbols instead of ASCII approximations. 

I wouldn't be opposed, but perhaps this is a subset of snippet support (#395)? Could we merge this now and grow it to support snippets later when we get marks, or else would it be better to wait for an all at once implementation and ensure this is in there as a special case?

pickfire:
But it seemed weird given that it can accept more than 2 characters, digraph is supposed to accept two characters.

Lindenk:
I can rename it to something else, maybe just unicode input? I figured most people looking for this feature would search for `diagraph` though

EpocSquadron:
> But it seemed weird given that it can accept more than 2 characters, digraph is supposed to accept two characters.

That's why I say it seems like a good base for snippets, especially if it grows the ability to add custom shortcut triggers. Add on top of that marks and we have snippets.

FVellu:
Any interest in finishing this? It's an essential feature for non-english speakers.

omentic:
I really like this patch. It's easy and fast to use, and also adds support for the handful of Unicode-heavy programming languages - particularly proof assistants.

I might suggest supporting generic Unicode input via hex literals as a fallback and reading from a different file than config.toml. It might also be good to bundle default digraphs with Helix, I've been going through and making a personal file and it's turned out to look just like every other list of commonly used Unicode characters out there.

omentic:
After working with this for a while, I've found it exceedingly helpful. These bindings work well for me: and I would suggest having space function the same as enter for ease of use.

```toml
[keys.insert]
"\\" = "insert_digraph"

[editor.digraphs]
"\\" = "\\"
...
```

kirawi:
Sorry for taking **_a while_** to review. I'm confused on why you opted for a trie instead of a vector. I would probably implement this feature as a `Vec<DigraphEntry>` and implement completion through `helix_core::fuzzy::fuzzy_match`.

Lindenk:
I don't think I knew there was already a fuzzy match implementation (or if there was one a year ago). It should be pretty easy to swap over to using it instead of a handbuilt trie if that's preferable

kirawi:
So I brought this up on Matrix, and @pascalkuthe said,

>  In particular I think this should be handled by the same infrastructure that will Handel custom snippets (and abbreviations). I think just having an abbreviation which has a special flag that makes it a diagraph that would not make it showup automatically but only once you press a certain key would be enough (it would just use fuzzy filtering/the normal completion windoe but autoconfirm once there only is a single match)

omentic:
Seems this should be closed in favour of #9801?

PR #2853, opened 2022-06-21
build(nix): update flake deps, add default.nix file

yusdacra:


the-mikedavis (approved):
Thanks! ❄️

archseer:
We had a default.nix before but it got removed for some reason

the-mikedavis:
That sounds familiar but I can't find any history of the file. Maybe it was the shell.nix that got removed?

PR #2855, opened 2022-06-21
Fix backwards character deletion on other whitespaces

### LINKED ISSUE #2842: Cannot backspace on full-width space character
### Summary

The idiographic space cannot be deleted on insert mode via Backspace.
If 'Delete' is hit (in front of the idiographic space) the character is deleted.

This test has been performed with:
Google IME in IME mode as well as alphanumeric mode

Tried on WSL2 (on NixOS)

### Reproduction Steps

I tried this:

1. `hx`
2. Enter a idiographic space in insert mode "　" (U+3000)
3. Try to backspace delete the character (in insert mode)

I expected this to happen:
The character will be deleted.

Instead, this happened:
Nothing, it cannot delete the character in insert mode.

### Helix log

N/A

### Platform

Windows 11

### Terminal Emulator

Windows-Terminal

### Helix Version

helix-term 22.05 (27609f50)

s0LA1337:
This issue closes #2842

In summary:
`delete_char_backward` ignored any other whitespace other than ' ', which resulted in you not being able to backspace or `C-h` on insert mode when there's only whitespace with the character next to the cursor being a lexiographic whitespace.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2981,7 +2981,7 @@ pub mod insert {
                         for _ in 0..drop {
                             // delete up to `drop` spaces
                             match chars.next() {
-                                Some(' ') => start -= 1,
+                                Some(c) if c.is_whitespace() => start -= 1,
                                 _ => break,
                             }
                         }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2944,7 +2944,7 @@ pub mod insert {
                 let line_start_pos = text.line_to_char(range.cursor_line(text));
                 // consider to delete by indent level if all characters before `pos` are indent units.
                 let fragment = Cow::from(text.slice(line_start_pos..pos));
-                if !fragment.is_empty() && fragment.chars().all(|ch| ch.is_ascii_whitespace()) {
+                if !fragment.is_empty() && fragment.chars().all(|ch| ch == ' ' || ch == '\t') {
                     if text.get_char(pos.saturating_sub(1)) == Some('\t') {
                         // fast path, delete one char
                         (
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2981,7 +2981,7 @@ pub mod insert {
                         for _ in 0..drop {
                             // delete up to `drop` spaces
                             match chars.next() {
-                                Some(c) if c.is_ascii_whitespace() => start -= 1,
+                                Some(' ') => start -= 1,
                                 _ => break,
                             }
                         }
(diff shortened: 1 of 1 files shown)


kirawi:
We actually probably just want to change this check here:
https://github.com/helix-editor/helix/blob/5b3b6ffc9e9b34fbbb39ad33cd29c8dec78ac231/helix-term/src/commands.rs#L2947

Most indentation will be using spaces or tabs, rather than double-width characters.

s0LA1337:
> We actually probably just want to change this check here:
> 
> https://github.com/helix-editor/helix/blob/5b3b6ffc9e9b34fbbb39ad33cd29c8dec78ac231/helix-term/src/commands.rs#L2947
> 
> Most indentation will be using spaces or tabs, rather than full-width characters. I don't have a strong opinion on it though.

actually, i don't have any strong opinion either, however the change happened like that because this already looked rather wrong.

helix-term/src/commands.rs
```diff
@@ -2981,7 +2981,7 @@ pub mod insert {
                         for _ in 0..drop {
                             // delete up to `drop` spaces
                             match chars.next() {
-                                Some(' ') => start -= 1,
+                                Some(c) if c.is_whitespace() => start -= 1,
```
archseer, line 2984: I think this won't work correctly with whitespace that has a wider width than 1 (eg. space used in Japanese text).

The body of this function should be extracted into `helix-core` and properly unit tested.

s0LA1337: > I think this won't work correctly with whitespace that has a wider width than 1 (eg. space used in Japanese text).

Well the Japanese text spaces are Ideographic Spaces.
I also filed this bug because i am using a japanese ime quite frequently.

This change has been tested locally, and works.

However i see what you mean, would it maybe make sense to use the `width` variable that is used a bit higher?
```rs
                          let width: usize = fragment
                            .chars()
                            .map(|ch| {
                                if ch == '\t' {
                                    tab_size
                                } else {
                                    // it can be none if it still meet control characters other than '\t'
                                    // here just set the width to 1 (or some value better?).
                                    ch.width().unwrap_or(1)
                                }
                            })
                            .sum();
```

helix-term/src/commands.rs
```diff
@@ -2981,7 +2981,7 @@ pub mod insert {
                         for _ in 0..drop {
                             // delete up to `drop` spaces
                             match chars.next() {
-                                Some(' ') => start -= 1,
+                                Some(c) if c.is_ascii_whitespace() => start -= 1,
```
kirawi, line 2984: I think this is an unnecessary change, as aside from line endings tab and space are the only two recognized as ascii whitespace.

archseer (approved):

Issue #1252, opened 2021-12-11
Feature Request: Support stack-graphs to provide code nav in the absence of LSP, building on tree-sitter


EpocSquadron:
## What are stack-graphs

[Stack-graphs](https://github.blog/2021-12-09-introducing-stack-graphs/) were developed by Github to provide code navigation features to any of the ~500 programming languages hosted in Github repositories (although they have only defined it for python currently). Examples of such features are "go to definition" and "go to references". It builds on the information obtained from tree-sitter grammars to create incrementally-updated look-up graphs that allow sub-100ms navigation of named symbols.

## Why would we want them in Helix?

Helix at this time relies exclusively on tree-sitter to provide basic highlighting and other more advanced features (such as selection by scope, which is still being worked into the default keymap). The conceit is that sourcing or writing a tree-sitter grammar for a language should be the preferred way of implementing support for a language for Helix. If we are to keep with this preference, we should endeavour to reap the maximum benefits of having a tree-sitter parsed syntax tree available for editor functions. This seems to be part of what sets helix apart from other editors.

## Helix supports LSP for those features, why not rely on that?

There are a number of reasons:

### Resource constraints

Many LSP servers are notoriously memory and cpu intensive. Depending on the kind of development machine you are running on, that may or may not be an issue even for one running LSP server. However, in heterogenous code environments (consider a website that uses typescript on the frontend, PHP on the backend and ruby for it's build system or to provide certain endpoints), or in environments where you are working on a number of projects throughout the day, whether or not all in the same language, it can quickly become limiting to rely on LSP in every case. It will likely be a concious decision to limit to a single LSP for the most used language simply to free up resources.

### Not all languages have functional LSPs

Fairly recently, there was a lot of effort to utilize Julia's LSP server in the default configuration of Helix, but it was found that not only was the server excruciatingly slow to start up, but it wasn't entirely conformant or complete. There are certainly other examples of LSPs that only provide partial support for the standard. In those cases, it would be simpler to create a ruleset for stack-graphs to at least provide code navigation, rather than write or improve an entire LSP implementation.

### Not all languages _should_ have an LSP

[...]

Issue #1151, opened 2021-11-24
locals are not retained for the whole document


dead10ck:
I reported this in the Matrix chat room, but filing an issue to track it.

I tried to improve the C highlighting to highlight macro invocations differently than regular function calls. After reading through the tree-sitter docs, it seems like [local variables](https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables) should be able to enable this.

So far, I think I've followed the docs pretty closely and came up with [this](https://github.com/dead10ck/helix/commit/1e3813ae5211042b8efedcaef2aabd6534694b3e). However, it doesn't seem to be working. Macro invocations are still being colored like any other function call.

I figured out that locals do seem to work, but only if the definition is on screen.

[![asciicast](https://asciinema.org/a/TtsxTPTdiy7ztbOIVeUSntuKD.svg)](https://asciinema.org/a/TtsxTPTdiy7ztbOIVeUSntuKD)

As soon as you scroll past the definition, it loses it. I can see with added debug logging around [here](https://github.com/tree-sitter/tree-sitter/blob/e85a279cf29da1b08648e27214dda20a841e57c8/highlight/src/lib.rs#L791) that the `scope.local_defs` is at first populated with the macro definition, but when it goes out of sight, `scope.local_defs` becomes empty.

@archseer mentioned that it could be because highlights are done only for the range on the screen, but that locals should be kept for the whole document.

https://github.com/helix-editor/helix/blob/master/helix-term/src/ui/editor.rs#L136-L142

### Reproduction steps

Define a language locals query that finds variable definitions, and color them differently. If the definition and the reference are both on screen, it works, but as soon as the definition is off screen, the highlight reverts back to its own tree-sitter match instead of deferring to the definition's highlighter.

### Environment

- Platform: Linux
- Helix version: v0.5.0-143-g8ae6d96

dead10ck:
@archseer unfortunately it doesn't look like the recent incremental changes mentioned in #1378 have fixed this issue. [This](https://github.com/dead10ck/helix/commit/b26addbaea65ae77a5229121fc54f6781ce4b172) is the branch I am using to test locals used in highlighting.

archseer:
Yeah I haven't addressed this yet, it requires some more thought. It's possible to resolve locals for the whole document when we deal with injections (but that's slightly inefficient if we don't need the whole set), but that doesn't solve the whole problem: we'd also need to know what highlighting query matches for that `local.definition` so we can use it for the reference.

[...]

PR #2857, opened 2022-06-22
Rainbow tree-sitter matches 🌈

the-mikedavis:
Configurable, themable, nesting rainbow highlights:

demo

This approach uses the syntax tree to guide nesting levels and highlights so it ends up being much more flexible than a naive pair-matching implementation. For example, we can highlight `<`/`>` when used in Rust type parameters like `struct MyStruct<'a> { .. }` and not mistakenly highlight `<`/`>` comparison operators. We can also capture multiple nodes, for example the `#`, `[` and `]` characters in a Rust attribute like `#[cfg(windows)]` and none of these characters are required to be adjacent.

The cherry on top is that we can highlight according to the syntax tree's nesting information rather than whether characters surround each other. For example, in this TOML table, `characters` is a sub-table of `editor.whitespace` and the rainbow highlights reflect that fact.

```toml
theme = "default"

[editor.whitespace]
# <- red          ^ red
render = "all"
characters = { tab = "▸", newline = "◂" }
#            ^ yellow                   ^ yellow (comes after red in base16)
```

It also works across injected languages like javascript injected into HTML script tags.

(You can see each of these things in action in the gif above.)

<details><summary><b>How does it work?...</b></summary><br>

This implementation leverages tree-sitter queries. We use a `QueryCaptures` Iterator very similar to the existing one for syntax highlights to query the parsed syntax tree for new `rainbows.scm` patterns.

Two new captures - `@rainbow.scope` and `@rainbow.bracket` - track nesting level and capture syntax nodes for highlighting with rainbows. Specifically: `@rainbow.scope` pushes a `RainbowScope` on a stack. When `@rainbow.bracket` matches, we test that the captured node is a direct descendant of the current scope. If it is, we emit highlights for it according to the scope's level of nesting.

<hr></details>

<details><summary><b>Configuration...</b></summary><br>

In the theme we add a `rainbow` key that takes a list of `Style`s. You can have however many you want and you can re-use colors from the palette or even throw modifiers in there:

```toml
# ~/.config/helix/themes/my_theme.toml

rainbow = ["#fc6b59", { fg = "lavender", bg = "comet" }, "lilac", { modifiers = ["reversed"] }]

[palette]
lavender = "#a4a0e8"
comet = "#5a5977"
lilac = "#dbbfef"
```

A default rainbow using the red, yellow, green, blue, cyan, and magenta terminal colors is used as a fallback.

Enable the rendering of rainbow brackets in your `config.toml`

```toml
# ~/.config/helix/config.toml
[editor]
rainbow-brackets = true
```

Or enable rendering per-language in `languages.toml` (this overrides the `config.toml` setting for any languages that set this value):

```toml
[[language]]
name = "scheme"
rainbow-brackets = true

[[language]]
name = "commonlisp"
rainbow-brackets = true
```

When the `rainbow-brackets` option is disabled, rainbow brackets are not rendered or computed.

<hr></details>

<details open><summary><b>Status...</b></summary><br>

The current implementation is working well. Take it for a spin and let me know if you find any bugs!

</details>


archseer:
Surprisingly little code to get this working 👍🏻

s0LA1337:
Would it make sense to use the color palette colors as default values if the rainbow feature is enabled?
That was my idea when i tried to work on that.

the-mikedavis:
That could work but usually the palette has a bunch of different ui colors too, so some deeply nested brackets could end up using backgrounds / grays which may look bad. I think a default rainbow palette not based on the current theme could work ok

the-mikedavis:
I suspect theme colors may not be good for a theme's rainbow because they might blend in and end up looking confusing. I'll give it a try though - that would take all the guesswork out of having to come up with a default that looks good on all themes 😄

archseer:
You could get fancy and calculate the color contrast and only pick colors that are distant enough :D

atagen:
> You could get fancy and calculate the color contrast and only pick colors that are distant enough :D

well, I have a snippet to calculate the L* luminance value (as in Lab/LCH/other "professional" colour systems) from an sRGB triple. I use it for exactly that, grading and excluding colours with weak contrast in a CLI app :+1: 

if interested Helix would be most welcome to it, I can PR it to @the-mikedavis rainbow branch.

the-mikedavis:
Let's keep this branched focused just on the happy path for now (rainbows enabled and themed). If anything, we might want to merge this once cleaned up without any default behavior - i.e. the theme needs to have a `rainbow` key in order for rainbows to work - and we can find a good default palette or strategy as a follow-up.

the-mikedavis:
The reason I lean towards a hardcoded default that uses red/orange?/yellow/green/blue/purple/etc is that it's easy to get a feel for how things are nested when you can predict the ordering of the bracket colors. If I know blue comes after green, I know where a blue tuple stands inside a huge green map or list literal.

the-mikedavis:
There's some not ideal behavior where the rainbow calculation depends on how much of the syntax tree is within view: https://asciinema.org/a/q7QVMs49QH1DQIPe46tZUAEsz

I think it might work to scan back in the document until you reach some byte that belongs only to the root node of the syntax tree and start the highlight iterator from there (discarding highlights that are not in view). That would also help with https://github.com/helix-editor/helix/issues/1151 but I don't know how feasible or efficient it would be.

the-mikedavis:
I found an alternate approach for this that ends up being less complicated. Now it works similarly to the locals tracking system in the syntax highlights iterator. The queries end up being more intuitive to write and more flexible (see the updated PR description).

I think this approach is general-purpose enough: I've worked through 27 languages so far with the toughest being HTML. HTML works now so that nested XML elements behave like other nesting scopes rather than just highlighting the `<` and `>` within individual tags.

pascalkuthe:
I merged this with #4042 in 08e990c.
The merge was nontrivial becasue I had to refactor the RainbowIter to emit `Span`s instead of `HighlightEvent` (and the history is not compatible with this branch I think).
Should I open a new PR or how do you want to handles this @the-mikedavis?

the-mikedavis:
I'll rebase this once #4042 is merged. It's too soon to rebase now I think.

helix-core/src/syntax.rs
```diff
@@ -951,6 +953,86 @@ impl Syntax {
...
+                let mut cursor = PARSER.with(|ts_parser| {
+                    let highlighter = &mut ts_parser.borrow_mut();
+                    highlighter.cursors.pop().unwrap_or_else(QueryCursor::new)
+                });
+
+                // The `captures` iterator borrows the `Tree` and the `QueryCursor`, which
+                // prevents them from being moved. But both of these values are really just
+                // pointers, so it's actually ok to move them.
+                let cursor_ref =
+                    unsafe { mem::transmute::<_, &'static mut QueryCursor>(&mut cursor) };
```
pickfire, line 986: Why not instead of using unsafe transmute to extend the lifetime, we wrap this whole thing inside `PARSER.with` so that the lifetime is guaranteed?

helix-core/src/syntax.rs
```diff
@@ -1589,6 +1721,53 @@ impl<'a> HighlightIter<'a> {
             }
         }
     }
+
+    /// Scans forward to the next capture from whichever layer has the earliest highlight
+    /// boundary, returning highlight-end events or terminating the iterator if there
+    /// are no more captures or highlight-end events.
+    fn scan_to_earliest_capture(&mut self) -> Option<Option<Result<HighlightEvent, Error>>> {
```
pickfire, line 1728: Do we really need so many layers here? `Option` and `Result`

helix-view/src/editor.rs
```diff
@@ -165,6 +165,8 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    /// Whether to render rainbow highlights. Defaults to `false`.
+    pub rainbow_brackets: bool,
```
pickfire, line 169: It defaults to true not false.

pickfire:
I am also curious where do you get those `example.*`.

helix-view/src/editor.rs
Philipp-M: quoting the PR description:

> Note: while this PR is in draft state, rainbow brackets are temporarily enabled by default.

the-mikedavis:
> I am also curious where do you get those `example.*`.

Most of them are from this repo:

* `example.toml`: `/languages.toml`
* `example.rs`: `/helix-core/src/syntax.rs` (in this branch)
* `example.scm`: `/runtime/queries/elixir/highlights.scm`
* `example.html`: `/book/theme/index.hbs` generated into an HTML file

And `example.erl` is https://github.com/erlang/otp/blob/1f10f9e8b497f7efc6da74f58e12b02b8ef1a431/lib/compiler/src/beam_disasm.erl#L218-L265 if I remember correctly.

Then I made a few screenshots of the same Wayland geometry (set a `geometry` variable with 
`slurp` and then `grim -g $geometry ~/frame-n.png` for each example), and stitched them into a gif with ImageMagick: `convert -delay 300 -loop 0 frame-*.png out.gif`

helix-view/src/editor.rs
the-mikedavis: Defaulting to true is temporary - before opening un-marking this PR as a draft I will drop the commit that defaults this to true

helix-term/src/ui/editor.rs
```diff
@@ -117,8 +117,15 @@ impl EditorView {
...
 
-        let highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
-        let highlights = syntax::merge(highlights, Self::doc_diagnostics_highlights(doc, theme));
+        let mut highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
+        if editor.config().rainbow_brackets {
+            highlights = Box::new(syntax::merge_iters(
+                highlights,
+                Self::doc_rainbow_highlights(doc, view.offset, inner.height, theme),
+            ));
+        }
```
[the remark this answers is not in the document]
the-mikedavis: I think it wouldn't overlay correctly: syntax highlights for `punctuation.bracket` would overlay the rainbow colors which would remove the rainbows in most themes.

I think it could still work to have selection highlights blend colors. Once rainbow highlights are merged onto syntax highlights, you can't really tell what's a syntax highlight or a rainbow highlight anymore from the render loop

helix-view/src/editor.rs
pickfire: I wonder if it would be better to default this to true? I wonder in what case this is not a good idea? It aids readability and doesn't really have much downside beside being bad for poeple with color blindness.

helix-term/src/ui/editor.rs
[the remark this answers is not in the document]
pickfire: I think merging both together is still fine, but maybe we want to add some context as of what color is being applied?

I am thinking to add additional context, so later we have the ability style selection color differently based on the context, but I am not sure how well that is going to work.

helix-core/src/syntax.rs
the-mikedavis: It does look really badly nested 😄

It's really `Option<Option<Self::Item>>` using the Iterator implementation's item. It does need the two nested `Option`s though:

* `None` for the case where we've already scanned to the next capture and can continue
* `Some(Some(event))` for emitting HighlightEnds for captures that have ended when handling the next capture
* `Some(None)` for terminating the iterator when it's at the end of all captures

helix-core/src/syntax.rs
the-mikedavis: The borrow checker doesn't like returning the `cursor` and `cursor_ref` or `captures`. I might be misunderstanding you though

s0LA1337:
btw - would you require some assistance with the languages?
I think i would be able to do it for Zig. As i am rather familiar with the language.

the-mikedavis:
That would be great! This PR has a guide for creating the `rainbows.scm` queries: https://github.com/helix-editor/helix/pull/2857/files#diff-68d1eecf9ab4c03f71757e5839fade60b6e110bf57713e6bebb2edef6a6dd062 and I could use some feedback on that, plus better language support would be good 👍

helix-view/src/editor.rs
the-mikedavis: I think it's a somewhat controversial feature - some people really like it and some would really hate it on by default. For features like that, I think the convention is to have it disabled by default.

book/src/guides/README.md
```diff
@@ -1,4 +1,4 @@
 # Guides
 
 This section contains guides for adding new language server configurations,
-tree-sitter grammars, textobject queries, etc.
+tree-sitter grammars, textobject and rainbow bracket queries, etc.
```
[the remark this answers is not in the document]
the-mikedavis: In this case I think it's ok - the "and" is so that "queries" applies to both "textobject" and "rainbow" since both are guides about writing queries

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+
+            // Otherwise, this capture must represent a highlight.
+            // If this exact range has already been highlighted by an earlier pattern, or by
+            // a different layer, then skip over this one.
+            if let Some((last_start, last_end, last_depth)) = self.last_highlight_range {
+                if range.start == last_start && range.end == last_end && layer.depth < last_depth {
+                    self.sort_layers();
+                    continue 'main;
+                }
+            }
```
pascalkuthe, line 2169: this could be trivially moved to a shared function wtihotu further changes

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+            let range = capture.node.byte_range();
+
+            // Remove from the rainbow scope stack any rainbow scopes that have already ended.
+            while let Some(scope) = self.context.rainbow_stack.last() {
+                if range.start >= scope.range.end {
+                    self.context.rainbow_stack.pop();
+                } else {
+                    break;
+                }
+            }
```
pascalkuthe, line 2137: note: this could be fully shared if we stored the scope_stack in the layer for both iterators (could be part of the generic iterator implementation as it also appears at the same place for both functions)

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+
+            if let Some(event) = self.scan_to_earliest_capture() {
+                return event;
+            };
+
+            let layer = &mut self.layers[0];
+            let (match_, capture_index) = layer.captures.next().unwrap();
+            let capture = match_.captures[capture_index];
+            let range = capture.node.byte_range();
```
pascalkuthe, line 2128: note: this is fully duplicate but doesn't lend it self well to moving to a seperate function as it's mostly controlflow that's why I suggested merging both iterator implementations

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+                    .config
+                    .rainbow_query
+                    .property_settings(match_.pattern_index)
+                {
+                    if prop.key.as_ref() == "rainbow.include-children" {
+                        scope.node_id = usize::MAX;
+                    }
+                }
+                self.context.rainbow_stack.push(scope);
+            }
```
pascalkuthe, line 2159: note: this is also mostly duplicate and could be in a function that calls back to a trait for a couple details

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+
+            if let Some(highlight) = rainbow_highlight {
+                self.last_highlight_range = Some((range.start, range.end, layer.depth));
+                layer.highlight_end_stack.push(range.end);
+                return self
+                    .emit_event(range.start, Some(HighlightEvent::HighlightStart(highlight)));
+            }
+
+            self.sort_layers();
```
pascalkuthe, line 2206: note: this is also fully duplicate like the start of the iterator

pascalkuthe (changes requested):
I did a pretty thorough review and think that the current version works well. Nice work @the-mikedavis  (love the detailed documentation too).

I think there is still some conflict with this PR and some change in master which I can't pin down. It works well on this exact commit but if I rebase on master highlights are totally gone for rust (despite the fact that health checks claims everything is fine, I tried messing with the runtime directory/rebuilding grammars but nothing fixes it). Could you try to rebase this on master and check if you see the same thing and look into a fix @the-mikedavis. I am not sure what is causing this (or if I messed up afer all).

Furthermore I have one larger comment. I worry a bit that this PR introduces too much code duplication between `HighlighIterator` and `RainbowIterator`. I think a lot of what is solved by duplicating the highlight iterator here could be achieved using a trait. This code is a bit subtle in places (and there is quite a lot of it) so I think avoiding duplication would be good here. It takes quite a bit longer to navigate understand the code because it's split up so much. I think this could be a hassle in the future if we make further changes to the highlighting to keep these two iterators in sync.

I added a bunch of comments to highlight what I mean here. I prefixed them with `note:` to make clear that these are not normal review comments and mostly just serve to demonstrate what I mean.

I was thinking of the following abstraction:

* Keep `scope_stack` in the layer and just make the `Vec` elements generic, this would allow facotring out many shared functions which just `pop`/`compare` elements. We could maintain an offset in a separate context field instead (or just restart the rainbow highlights inside injections... I think that might be ok as-well?)
* Share the entire `Iterator` implementation by making `C` implement a trait (could `scope_stack` elements could be an associated type). which would be used for iterator specific code. That would allow reusing most of the selection.
* Factor out quite a bit of code that is either trivially duplicate or 90% duplicate right now (the last 10% could be a call to the trait) that just appears in a different place in the function.

I don't think this is necessarily a blocker. We could merge this regardless when the other comments are addressed and do such a larger refactor in a followup PR.  Not sure what your stance is on that @archseer?

the-mikedavis:
> It works well on this exact commit but if I rebase on master highlights are totally gone for rust

I can give this a rebase. IIRC the rust grammar updated and changed a node that was added in the rust rainbow queries which is probably causing the highlights to fail analysis.

> I worry a bit that this PR introduces too much code duplication between `HighlighIterator` and `RainbowIterator`. I think a lot of what is solved by duplicating the highlight iterator here could be achieved using a trait.

Yeah I agree, there's quite a lot of duplication. I'll look into combining them and using a trait 👍

helix-core/src/syntax.rs
```diff
@@ -1946,6 +2098,116 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+            };
+
+            let layer = &mut self.layers[0];
+            let (match_, capture_index) = layer.captures.next().unwrap();
+            let capture = match_.captures[capture_index];
+            let range = capture.node.byte_range();
+
+            // Remove from the rainbow scope stack any rainbow scopes that have already ended.
+            while let Some(scope) = self.context.rainbow_stack.last() {
+                if range.start >= scope.range.end {
```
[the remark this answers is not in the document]
the-mikedavis: It looks like `>=` works in the highlight iterator as well. It might be a bug that it's `>` there but I can't find any symptoms from it. Using `>` in the rainbow iterator has some weird side effects with injected content.

the-mikedavis:
Ok I was able to eliminate most of the duplicated code and shortened/simplified the approach. The new version adds an iterator over query captures across injections. Figuring out rainbow highlights then becomes a small wrapper over that. We could use the new iterator for textobject queries too and fix https://github.com/helix-editor/helix/issues/4618. I'll leave that as a follow-up PR though since it's unrelated.

helix-core/src/syntax.rs
```diff
@@ -1606,11 +1771,23 @@ impl HighlightConfiguration {
...
 
-impl<'a> HighlightIterLayer<'a> {
-    // First, sort scope boundaries by their byte offset in the document. At a
-    // given position, emit scope endings before scope beginnings. Finally, emit
-    // scope boundaries from deeper layers first.
-    fn sort_key(&mut self) -> Option<(usize, bool, isize)> {
+trait IterLayer {
+    // TODO: this can be cleaner with GATs once the MSRV is 1.65:
+    // type SortKey where Self: PartialOrd;
+    type SortKey;
```
pascalkuthe, line 1777: I don't udnerstand what GATs woul be used for here.
`type SortKey: PartialOrd` seems to be sufficient here.

pascalkuthe (commented):
I was not able to finish mz review yet but this looks much easier/readable. I like it :)

helix-core/src/syntax.rs
the-mikedavis: Oh hah actually that's just me misunderstanding traits and GATs 😄. This works great

pascalkuthe:
sorry about the delay with finishing my review, I am a bit busy in the days leading up to christmas and didn't get the chance to sit down for a longer period and do a thorough review. I will try to finish the review either in the evening today or tomorrow morning

helix-core/src/syntax.rs
```diff
@@ -712,6 +718,34 @@ thread_local! {
     })
 }
 
+/// Creates an iterator over the captures in a query within the given range,
+/// re-using a cursor from the pool if available.
+fn query_captures<'a, 'tree>(
```
pascalkuthe, line 723:
```suggestion
/// Safety: The `QueryCaptures` must bed dropped before the `QueryCusor` is dropped
unsafe fn query_captures<'a, 'tree>(
```

This function signature makes me very uncomfortable because it can easily lead to use after free, alias violations and other kinds of UB. The way it is used currently is fine but it should be marked unsafe so it doesn't get used incorrectly in the future.

Really the crux of this function is that you need an owning `QueryCaptures` iterator that owns the `QueryCursor` (and free it on `Drop`) and can be turned back into a `QueryCursor`. That kind of API could be implemented here but it would be a somewhat ugly hack. This should really be supported inside the tree sitter bindings directly. I will be submitting a breaking change to that API anyway and try to include an appropriate type with that change.

For now marking this unsafe should be good enough tough so we don't block this PR on the TS changes

pascalkuthe (dismissed):
Structuring the rainbow iterator as an eager loop instead of lazy iterator really simplified the code a lot (and made it reusable for other purposes) and you managed to remove most of the remaining duplication with the trait.

Apart from the one comment I added below I think this should be ready to merge :+1:

helix-core/src/syntax.rs
```diff
@@ -1043,6 +1079,55 @@ impl Syntax {
...
+                // If there aren't any captures for this layer, skip the layer.
+                captures.peek()?;
+
+                Some(QueryIterLayer {
+                    cursor,
+                    captures,
+                    layer,
+                })
+            })
+            .collect::<Vec<_>>();
```
pickfire, line 1115: I think doing collect with `let mut layers: Vec<_> = ` make it easier to read than doing it here.

ribosomerocker:
Just tested it out and it works very well, what's the hold-up?

the-mikedavis:
We're hoping to get #5420 in first since it's a large refactor and touches a lot of the rendering code. This will need to be rebased after that and I need to check if the queries need any fixing after recent tree-sitter update PRs.

In general though, a PR working well is a minimum requirement. We also need to check the code-quality and performance and decide whether the feature is something we want to support in core (rather than a plugin). All of that takes time, especially for a large and kinda complicated PR like this. There are a lot of PRs open that need a lot of review time and all of the maintainers review in their free time, so please be patient.

If you want to use unmerged stuff you can merge branches like this into a branch in a fork and build from that. This branch is in [my "daily driver" branch](https://github.com/the-mikedavis/helix) as an example. I'll rebase and sort out the conflicts now to make that easier.

SoraTenshi:
> If you want to use unmerged stuff you can merge branches like this into a branch in a fork and build from that. This branch is in [my "daily driver" branch](https://github.com/the-mikedavis/helix) as an example. I'll rebase and sort out the conflicts now to make that easier.

Yup, that's e.g. what i did. i have an experimental branch up which is also used by quite some people.
The code there is just an ugly merge of some prs of features i really want to use, but functional.

helix-core/src/syntax.rs
```diff
@@ -1651,6 +1842,62 @@ impl<'a> HighlightIterLayer<'a> {
...
+            break;
+        } else {
+            let layer = layers.remove(0);
+            PARSER.with(|ts_parser| {
+                let highlighter = &mut ts_parser.borrow_mut();
+                highlighter.cursors.push(layer.cursor());
+            });
+        }
+    }
 }
```
pickfire, line 1901: Is this sort based on any algorithm?

pickfire (approved):
I roughly go through the code, it looks fine to me, but maybe having another pair of eyes is better.

helix-core/src/syntax.rs
the-mikedavis: This block could probably use explanation comments: it's a quick (linear time w.r.t. number of layers) re-ordering that makes sure that the layers are sorted by the next capture but it assumes that `layers` is already almost fully sorted and will just shift a few elements towards the front if not. Plus it drops layers that are finished emitting captures. I'll add some comments to clarify what's going on in this block

helix-term/src/ui/editor.rs
```diff
@@ -136,6 +136,16 @@ impl EditorView {
 
         let mut highlights =
             Self::doc_syntax_highlights(doc, view.offset.anchor, inner.height, theme);
+        if doc
+            .language_config()
+            .and_then(|lang_config| lang_config.rainbow_brackets)
+            .unwrap_or(config.rainbow_brackets)
```
pickfire, line 142: Can we keep this as a variable instead? The condition is a bit too long.

pickfire (approved):
Other than the minor comments, the change looks good to me.

pickfire (approved):

paul-scott:
You are probably already aware of this, but some of the languages like racket and common-lisp are missing a rainbows.scm file. They can just likely inherit from scheme like they do for highlights etc. Mentioning this because these are some of the languages that would benefit most from this feature.

pickfire:
@paul-scott I think @the-mikedavis don't really have to go to do each one of them, especially those he don't even use, but other people can always add it for the other languages, just like inlay-hint, the author of inlay-hint, probably don't want to go add support for inlay-hint in each theme.

paul-scott:
Yea that makes sense

the-mikedavis:
Good call with racket & common-lisp, I added those to the PR. I will probably avoid adding more languages myself unless I use them since it adds to the already large PR size and I'd like to see if contributors can use the new guide to write some queries themselves. But racket and common-lisp can just inherit as you say so it's no issue to add them 👍

EpocSquadron:
There are a lot of conflicts with latest master. It seems like there will be more once #8021 is merged, as one of the conflicts I tried to resolve locally had to do with redraw. Perhaps the merge of that would be a good time to revisit and rebase this properly? 

I'm interested in seeing #5176 progress, and it seems like the intent there is to build on the iterator improvements that are done here. It does look like this one is intended to merge before next release if the [next milestone](https://github.com/helix-editor/helix/milestone/4) is to be believed.

Once the above cascade of merges is complete, we would have a pretty robust core tree-sitter implementation in editor, which will be a prerequisite for a lot of other cool functionality like #1252.

pascalkuthe:
The next milestone is only a set of PRs we intend to likely merge at some point. But not indicative that PRs will necessarily make it into the next release. I don't think this will conflict much with #8021, the two PRs modify completely orthogonal parts of the codebase.

the-mikedavis:
This branch gets a lot of conflicts but they're usually trivial: changes to the autogenerated [`lang-support.md`](https://github.com/helix-editor/helix/blob/master/book/src/generated/lang-support.md) file cause conflicts when new languages are merged. I'll push up the latest rebase though - it looks like there have been a bunch of smaller changes that caused true conflicts since I last pushed.

#5176 operates on trees while this branch covers queries. We'll need both to be able to work across injection layers for all tree-sitter features but their implementations are separate. It's basically [`tree_sitter::TreeCursor`](https://docs.rs/tree-sitter/0.20.10/tree_sitter/struct.TreeCursor.html) (5716) vs [`tree_sitter::QueryCaptures`](https://docs.rs/tree-sitter/0.20.10/tree_sitter/struct.QueryCaptures.html) (this branch).

I've been looking at the mechanism for running queries across injections again as a part of a new tree-sitter-based feature and I found some changes that will make it easier to use this for other tree-sitter features and remove any performance penalties when the feature is turned off, so I will set this as a draft for now. I think it will be better to land the query changes as part of a smaller feature or a feature already in-tree so I'm planning on looking at textobjects after the next release before revisiting this.

W-Lawless:
Can we bump this PR? This is a feature that is considered standard in many other editors

omentic:
Hey @the-mikedavis - any chance this could be merged soon, or put back on a list of milestones? Is there anything blocking it? I keep this merged into my soft fork and it works great. It's indispensable for writing Racket/Scheme, and pretty helpful for all languages in general. And another languishing PR (#4493) is blocked on it.

I have a rebased version over at https://github.com/omentic/helix/tree/rainbow-rebase, if needed.

JustForFun88:
I don’t understand why this PR has been pending since 2022, especially since it’s fully functional and adds valuable feature. Leaving this to plugins isn’t ideal, as it could hurt performance. VS Code faced a similar issue and ended up integrating support directly.

x10an14:
For those asking "why not merge dis??", read https://github.com/helix-editor/helix/pull/2857#issuecomment-1694728846 and explore the different iaauea/PRs related to treesitter (that are all somewhat related to one another).

yerlaser:
It's not possible to rebase and build this branch any more :(

the-mikedavis:
I have this rebased locally on #12972 - I'll push that up when I get the chance. (Tree-house covers the main part of this that was hacky since it introduces a generic query iterator across injection layers)

the-mikedavis:
Interesting... I guess fast-forward merging closes PRs targetting the merged branch. I'll reopen this in a new PR

PR #2860, opened 2022-06-22
Fix scrollbar length proportional to total menu items

sudormrfbin:
The scrollbar length used to increase with more entries in the menu, which was counter-intuitive to how scrollbars worked in most applications. Turns out there was a typo in the floor division implementation :)


archseer (approved):
good catch!

PR #2676, opened 2022-06-04
Add mode specific styles

Mathspy:
In similar vein to neovim's `lualine` and similar statusline packages this allows helix users to style their mode based on which mode it is thus making each mode more visually distinct at a glance

Screenshots of how it looks attached below
cropped-nor
cropped-ins
cropped-sel

Issue #1575, opened 2022-01-24
:wq silently exits without an error if target file is not writable


Pistahh:
### Reproduction steps
:wq should fail and indicate if it couldn't save the file but it doesn't, it just exits.

1. edit a file that you have no permissions to, e.g. /etc/hosts as non-root
2. Try to save it and exit with `:wq`
3. watch as hx exits without any indication that it couldn't save the changes
4. inspect /etc/hosts that it indeed wasn't changed

 

### Environment

- Platform: Linux
- Terminal emulator: gnome terminal
- Helix version: v0.5.0-186-g371c84f

dead10ck:
I started trying to look into this one. At first glance, it seems like this should exit with an error before quitting. However, if we look at `write_impl`, we can see that it looks like it just adds an async job and returns immediately, which would explain why it just proceeds to quit, and doesn't really care if the write succeeds or not:

https://github.com/helix-editor/helix/blob/333c2949c2b036f8aafa77be2286eb3c135ea269/helix-term/src/commands.rs#L2087-L2094

I thought it would make sense to wait on the job in between calling write and calling quit, so looking at the `Jobs` type, it looks like there is a function for waiting on all the pending jobs:

https://github.com/helix-editor/helix/blob/333c2949c2b036f8aafa77be2286eb3c135ea269/helix-term/src/job.rs#L95-L99

However, as we can see, this doesn't care about the result of the jobs either. Would it make sense to have this function return a Result that's an error if any of the pending jobs error, and Ok otherwise?

DannyJJK:
I'd love to try this using Helix as my main editor, but this single issue makes me nervous to use it. There's various reasons a file could fail to write that are not just permissions based, I wouldn't want to think I'd saved a file successfully to find that no changes have been made. Has there been any further progress on this?

dead10ck:
I agree, this seems rather important from a reliability perspective. I'll take a crack at the approach I mentioned above.

PR #2862, opened 2022-06-22
add customizable read-only indicator to statusbar 

### LINKED ISSUE #2627: No warning when editing a readonly file
When opening a file you are not allowed to edit, no indication is given that this is the case. Also, when you start editing, there is no warning either. And then if you do `:wq`, Helix closes without issues. This means that you can think you have edited a file, without ever realizing that the file is readonly and nothing has changed.

Vim on the other hand has `[readonly]` in the bottom bar when opening the file, gives a red colored warning when starting editing, and when `:wq` is tried an error is shown and the editor remains open.

sbromberger:
closes #2627.


helix-term/src/ui/editor.rs
```diff
@@ -808,13 +809,19 @@ impl EditorView {
...
+            String::new()
+        };
         let title = {
             let rel_path = doc.relative_path();
             let path = rel_path
                 .as_ref()
                 .map(|p| p.to_string_lossy())
                 .unwrap_or_else(|| SCRATCH_BUFFER_NAME.into());
-            format!("{}{}", path, if doc.is_modified() { "[+]" } else { "" })
+            format!("{}{}{}", readonly, path, modified)
```
the-mikedavis, line 824:
```suggestion
            format!("{}{}{}", path, readonly, modified)
```

It looks like it's part of the path if it comes before without any spacing

sbromberger: Yeah, that was me playing around with the position of the unicode lock icon. I prefer it before the filename, but it's fine afterwards. I also debated inserting spaces between the path and the indicators, but decided against - it's easy enough to insert a space in the indicator string if you want it.

sbromberger:
Would it be possible to get this reviewed soon? Merging will require me to modify #2831 but this lays the groundwork for these sorts of changes.

the-mikedavis:
If anything, we may want to wait on this for #2434 and then split out the `helix_view::editor::StatusLineElement::FileName` component into separate elements for the modification and read-only indicators.

Whether to default this to being enabled or disabled is worth some discussion: kakoune does not show a read-only indicator but vim does.

helix-view/src/document.rs
```diff
@@ -38,6 +38,13 @@ pub enum Mode {
     Insert = 2,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize)]
+pub enum FileStat {
```
the-mikedavis, line 42: I think it would be more accurate to call this `WritePermissions` (and same for the field in `Document`): stat covers a lot of metadata and this is pretty specific.

sbromberger:
> kakoune does not show a read-only indicator but vim does.

The problem we currently have with helix is that it's possible to lose work because of the lack of realization that you're editing a RO file.

the-mikedavis:
That particular issue was fixed recently: https://github.com/helix-editor/helix/issues/1575

sbromberger:
> That particular issue was fixed recently: https://github.com/helix-editor/helix/issues/1575

Wow, I didn't even see that merged. Thanks!

etienne-k:
> If anything, we may want to wait on this for #2434 and then split out the `helix_view::editor::StatusLineElement::FileName` component into separate elements for the modification and read-only indicators.
> 
> Whether to default this to being enabled or disabled is worth some discussion: kakoune does not show a read-only indicator but vim does.

#2434 was merged a few days ago! :)
I hope this PR is still relevant, because this feature sounds very useful.

Regarding the default behavior: I believe it should be _enabled_ by default.

I can't think of many reasons why somebody would _not_ want to see such an indicator.
The only reason could be having a feature like (neo)vim which - in addition to the indicator (`[-]`) - displays a warning in the commandline:
image
But at least for the time being, we have no information about the read-only nature of the edited file at all.

sbromberger:
I am planning on revisiting this soon but am on travel for a few days. I might have some time but do t want to promise anything specific :)

sbromberger:
Turns out this was more straightforward than I had expected (thanks to the great new statusline implementation). This works for me locally but more review would be appreciated.

book/src/configuration.md
```diff
@@ -46,6 +46,7 @@ hidden = false
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
+| `read-only-indicator` | String to display next to filename in statusbar when configured and if file is read-only. Set to `""` to disable. | `"[RO]"` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
```
etienne-k, line 50: A tiny improvement idea: I think it might be better to have both `read-only-indicator` _and_ `color-modes` under `[editor.statusline]`, because they're specific to the statusline (and I can't imagine that changing in the future).

Now, I know that in master, `color_modes` was there already. Mostly because when it was introduced by #2676, the statusline PR #2434 wasn't merged yet.

But I don't think it's too late for a little improvement which could make the configuration a bit more intuitive.

sbromberger: I was also hoping to add the RO indicator to the bufferline when it gets merged :)

etienne-k: Right, didn't think about that one! In that case, you can disregard my suggestion. :)

helix-term/src/ui/statusline.rs
```diff
@@ -354,6 +355,19 @@ where
...
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let readonly = if context.doc.write_permission == WritePermission::ReadOnly {
+        context.editor.config().read_only_indicator.to_string()
+    } else {
+        String::new()
+    };
+
+    write(context, readonly, None);
```
the-mikedavis, line 368: It's a nit but I think the `readonly` variable should be `indicator` instead. The `readonly` name makes me automatically think it's a bool 🙃

helix-view/src/document.rs
```diff
@@ -143,6 +152,7 @@ impl fmt::Debug for Document {
             .field("version", &self.version)
             .field("modified_since_accessed", &self.modified_since_accessed)
             .field("diagnostics", &self.diagnostics)
+            .field("file mode", &self.write_permission)
```
the-mikedavis, line 155:
```suggestion
            .field("write_permission", &self.write_permission)
```

helix-view/src/document.rs
```diff
@@ -121,9 +128,11 @@ pub struct Document {
 
     diagnostics: Vec<Diagnostic>,
     language_server: Option<Arc<helix_lsp::Client>>,
+    /// Mode of file - currently read-only, writable, or unknown.
```
the-mikedavis, line 131:
```suggestion
    /// Write permissions of the file - read-only, writable, or unknown.
```

the-mikedavis (commented):
This is looking good to me. I think enabling this by default is a good choice 👍

helix-term/src/ui/statusline.rs
sbromberger: how about `readonly_indicator`? There may be other indicators later.

the-mikedavis (approved):

dead10ck:
@the-mikedavis is there anything else this PR needs?

the-mikedavis:
The bufferline commits should probably be dropped out of the changes but otherwise this looks good to me

PR #2875, opened 2022-06-25
languages: Fix ruby rakefile and gemfile file type

kihaya:
I found default listed ruby file type (`rakefile` and `gemfile`) does not match Rakefile and Gemfile, so fix them.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -446,7 +446,7 @@ source = { git = "https://github.com/cstrahan/tree-sitter-nix", rev = "6b71a810c
 name = "ruby"
 scope = "source.ruby"
 injection-regex = "ruby"
-file-types = ["rb", "rake", "Rakefile", "irb", "Gemfile", "gemspec"]
+file-types = ["rb", "rake", "rakefile", "irb", "gemfile", "gemspec", "Rakefile", "Gemfile"]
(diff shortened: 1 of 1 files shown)


languages.toml
```diff
@@ -446,7 +446,7 @@ source = { git = "https://github.com/cstrahan/tree-sitter-nix", rev = "6b71a810c
 name = "ruby"
 scope = "source.ruby"
 injection-regex = "ruby"
-file-types = ["rb", "rake", "rakefile", "irb", "gemfile", "gemspec"]
+file-types = ["rb", "rake", "Rakefile", "irb", "Gemfile", "gemspec"]
```
the-mikedavis, line 449: I've seen some `foo.rakefile` files in the wild. I'm not sure it's technically valid of if it's just allowed by case-insensitive file systems. Let's have both `rakefile` and `Rakefile` for anyone using the lowercase version (might as well do the same for the gemfile too)

kihaya: @the-mikedavis 
Thank you for your review.
I changed to both file type is available.

https://github.com/helix-editor/helix/commit/ea7a1604bb305499df66078112ba1ea805d6804f

the-mikedavis (approved):
Thank you!

PR #2877, opened 2022-06-25
Fix edge-case in tree-sitter expand_selection selection command

Triton171:
If a tree-sitter grammar outputs multiple nodes with the same range that contain each other (I noticed this with the LaTex grammar and the `text` and `word` nodes), the `expand_selection` command currently "gets stuck" at them. This happens because the `descendant_for_byte_range` function always returns the lowest of these nodes, so changing the selection to the range of its parent doesn't do anything.
I changed it, so that now it skips nodes with identical ranges.

### DIFF
--- a/helix-core/src/object.rs
+++ b/helix-core/src/object.rs
@@ -2,12 +2,11 @@ use crate::{Range, RopeSlice, Selection, Syntax};
 use tree_sitter::Node;
 
 pub fn expand_selection(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
-    select_node_impl(syntax, text, selection, |descendant, from, to| {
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Great catch! I was wondering why `A-o` was getting 'stuck' on some nodes, I thought it might've had something to do with named vs unnamed but this makes sense 👍

(One example of this:

```erlang
%% f.erl
foo(Bar) ->
    Bar#{a => b}.
```

with a selection over `a => b` will get stuck on master).

Triton171:
Yeah, it wasn't a big deal but just irritating enough when it happened that I wanted to fix it. Thanks for the super quick review :)

Issue #2868, opened 2022-06-23
cannot use `.` to repeat a `replace with char (r)` action


evanrichter:
### Summary

If I replace a char with `r`, move to another place and hit `.` the replace-char does not happen. The action _before_ replace-char is done.

### Reproduction Steps

https://asciinema.org/a/Kxn1SWvwfw58rMGk4j4SSdOzm

I tried this:

1. `hx`
2. `iaaaa<ESC>` <= this repeats
3. `hhh`
4. `rB`       <= but this should repeat
5. `ll`
6. `.`

I expected this to happen:

step 4 to repeat. buffer should be `aBaB`

Instead, this happened:

step 2 was repeated, 4 a's were inserted

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.05

amitbeka:
Hey, a new user here trying to contribute 👋🏽 

Sharing some investigation I did in case I can't pick it up again soon (and for future me):

- Initially thought the problem is the lack of a call for `doc.append_changes_to_history` which I see in many edit functions but not in `replace`. However, that didn't solve the problem
- I also noticed `switch_case` action has the same issue, also lacking `doc.apply_changes_to_history`
- What puzzled me is that undo-ing a replace works, which AFAIK come from the same `history` as redo-ing
- Now I've found that repeating the last action is actually handled separately from undo/redo, in [editor.rs](https://github.com/helix-editor/helix/blob/33e6df870710a48e89875b6f7471b9ee5f54af50/helix-term/src/ui/editor.rs#L896) so I should look into that instead

archseer:
I think this is by design, the dot is meant to redo the last insert. A replace is not an insert. The same issue occurs in kakoune

amitbeka:
Interesting, in neovim for example, the dot redo the last replace also.

Another change in behaviour I see is around changing the selection - in neovim (as an example) it repeats the whole sequence (let's assume you do `cw` then insert a different word) and in helix it just does the insert again, without selecting the word and replacing it.

All in all I think it's sane to do just the insert again, since it doesn't save a lot of keystrokes to repeat the last non-insert change.
With multiple selections you usually don't need to repeat actions too much.

Unless @evanrichter has other use-cases to consider, I can update the docs (they say last _change_) and we'll call it a day.

the-mikedavis:
Yeah with neovim it's a bit different because of helix (and kakoune's) select-then-act paradigm.

[...]

PR #2878, opened 2022-06-25
book: fix the description of dot repeat

amitbeka:
It should only repeat the last insert by design, see #2868


the-mikedavis (approved):
Thank you!

PR #2267, opened 2022-04-25
Write path fixes, integration tests

dead10ck:
Well, I did not expect this PR to get this large, I'm so sorry about that. This PR started as an attempt to fix #1575, and along the way, I discovered that there was a larger problem: Document writes are submitted to a single job pool that all run unordered. I suspected that this could lead to file write inconsistencies if the user submitted multiple writes on the same document at the same time. I took this opportunity to continue work on the integration testing harness that @archseer started a while back. I added an integration test (see below) that confirms that indeed the file writes happen not only out of order, but simtultaneously, leading to the file being clobbered. In a test where we issue a thousand writes that each change the entire file to consist only of the iteration number (i.e., 1, 2, 3, ..., 1000), we can see that the file ends up with a seemingly random jumble:

```
failures:

---- integration::write::test_write_concurrent stdout ----
thread 'integration::write::test_write_concurrent' panicked at 'assertion failed: `(left == right)`
  left: `"1000"`,
 right: `"4010"`', helix-term/tests/integration/write.rs:66:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace


failures:
    integration::write::test_write_concurrent
```

This test is now passing with the changes to the write path.

Below is an explanation of some of my decisions in this PR. Opinions and suggestions are welcome!

## Document writes

The approach taken was:

1. Rather than submitting a future to the general job pool, each Document manages its own queue of writes with an mpsc channel. When a user submits a write, this merely queues up a write request on the internal channel.
1. Document writes emit a new special event type that communicates the revision number that was written and the path. The main event loop is responsible for polling the futures, which prompt the runtime to carry out the actual writes. When writes finish, the main event loop will update the document's last saved revision and the path if the write was successful.
1. All events that originate from the Editor were consolidated into one enum type. This was discussed in the chat room, and it was originally expressed that we could do this incrementally over time, but the borrow checker intervened with this plan: since the event polls require a mutable borrow, and we can only do this once within the event loop, we have to do this all at once.

Incidentally, this allowed fixes for two unrelated bugs that presently exist:

* When a document write fails, the modified flag is still reset. By sending the revision number in the write event, the current revision number and the last saved revision number need not be the same at any given point in time. This would, e.g., allow large writes to carry out in the background while the user makes other edits at the same time, and correctly keep the modified flag even after the background write succeeds.
* In a write command that sets a new path, if the write fails, the path is still changed to the  argument path. Now, it will not update the path unless the write succeeds.
  
[...]

PR #2883, opened 2022-06-26
Fix :w in insert mode

xDarksome:
Currently, when you bind `"C-s" =  ":w"` in insert mode and try to use it, the file is being written, but changes are considered not committed yet. Because of that helix complains on `:q` and still shows `[+]` in the status bar.

Tell me if there is better way to fix this.

### DIFF
--- a/runtime/themes/ayu_dark.toml
+++ b/runtime/themes/ayu_dark.toml
@@ -47,6 +47,7 @@
 "ui.text.info" = "foreground"
 "ui.virtual.whitespace" = "dark_gray"
 "ui.virtual.ruler" = { bg = "black" }
+"ui.virtual.inlay-hint" = { fg = "#e6b450", bg = "#302a20" }  # original bg #e6b45033
 "ui.menu" = { fg = "foreground", bg = "black" }
 "ui.menu.selected" = { bg = "gray", fg = "background" }
 "ui.selection" = { bg = "dark_gray" }
--- a/runtime/themes/ayu_light.toml
+++ b/runtime/themes/ayu_light.toml
@@ -47,6 +47,7 @@
 "ui.text.info" = "foreground"
 "ui.virtual.whitespace" = "dark_gray"
 "ui.virtual.ruler" = { bg = "black" }
+"ui.virtual.inlay-hint" = { fg = "#f4a028", bg = "#fcf2e3" }  # bg original #ffaa3333
 "ui.menu" = { fg = "foreground", bg = "black" }
 "ui.menu.selected" = { bg = "gray", fg = "background" }
 "ui.selection" = { bg = "dark_gray" }
--- a/runtime/themes/ayu_mirage.toml
+++ b/runtime/themes/ayu_mirage.toml
@@ -47,6 +47,7 @@
 "ui.text.info" = "foreground"
 "ui.virtual.whitespace" = "dark_gray"
 "ui.virtual.ruler" = { bg = "black" }
+"ui.virtual.inlay-hint" = { fg = "#ffcc66", bg = "#47433d" }  # original bg #ffcc6633
 "ui.menu" = { fg = "foreground", bg = "black" }
 "ui.menu.selected" = { bg = "gray", fg = "background" }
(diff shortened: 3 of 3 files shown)


feep:
Nice fix, thanks.

I was unable to bind anything to ":w" in insert, was unsure if I was doing it wrong or if it was not currently possible with keybinds in `config.toml`.

The following `config.toml` binding does not work in 22.05, does work with this PR: 

```
[keys.insert]
C-s = ":w"
```

the-mikedavis:
I can't reproduce that - running exactly that configuration file on 22.05 works and `C-s` successfully saves a file in insert mode (although it does not clear the modified indicator). You may have other bindings or configuration options that are invalid in your `config.toml`.

feep:
> running exactly that configuration file on 22.05 works and `C-s` successfully saves a file in insert mode (although it does not clear the modified indicator).

That is correct.

I think we are all describing the same behavior. Maybe I wasn't clear.

As described by you and xDarksome. v22.05 does not clear the dirty bit (It does save the file).

Didn't actually check to see if it saved the file, because that wasn't the bug.

...checking now.

With my config and v22.05, file is saved, but file is still shown as dirty.

With minimal config `[keys.insert]\nC-s = ":w"`, same result.

I'm pretty sure that is the same behavior the-mikedavis is describing.


If we still conflict, more info about my build:

```
❯ rustc --version
rustc 1.61.0 (Arch Linux rust 1:1.61.0-1)

❯ ./target/release/hx --version # PR version, resets dirty flag properly
helix 22.05 (a7f41ac5)

❯ uname -a
Linux [hostname] 5.15.48-1-lts #1 SMP Thu, 16 Jun 2022 13:32:35 +0000 x86_64 GNU/Linux
```

the-mikedavis:
Oh I see, I mis-parsed your comment as saying that helix was rejecting that keybinding config 😅

helix-view/src/document.rs
```diff
@@ -839,8 +845,10 @@ impl Document {
         let changes = std::mem::replace(&mut self.changes, new_changeset);
         // Instead of doing this messy merge we could always commit, and based on transaction
         // annotations either add a new layer or compose into the previous one.
-        let transaction =
-            Transaction::from(changes).with_selection(self.selection(view_id).clone());
+        let mut transaction = Transaction::from(changes);
+        if let Some(view_id) = view_id {
+            transaction = transaction.with_selection(self.selection(view_id).clone());
+        }
```
archseer, line 851: I think this isn't correct, the code assumes there's always a selection so we have the previous selection stored when undoing. The editor will probably panic otherwise.

archseer (changes requested):
This will quite likely be fixed by https://github.com/helix-editor/helix/pull/2267

\cc @dead10ck ?

dead10ck (commented):
I agree with @archseer. I assume this was done because in the save function call, you don't have a view associated with it. The solution isn't to just do whatever you can without it and hope it works out. This will leave the last edit in the history without a selection, which at best will make a very confusing and invisible undo operation, and at worst make the editor crash.

#2267 will not fix this either. The reason this is happening is because the modified flag is controlled by the revision number, which gets changed every time we commit changes to the history of edits for the purposes of tracking undos, and then make further changes. When the document's revision number is equal to the last revision number we saved to disk, the state is unmodified.

The way helix works as a modal editor is that you make your edits by going into insert mode, making your edits, and then leaving insert mode to go back to normal mode. This cycle of normal mode → insert mode → normal mode is the unit of discrete state changes to the text, and it's during the state transition from insert mode to normal mode that changes are recorded.

By trying to commit the changes while still in insert mode, you are breaking some key assumptions about how state changes are tracked. You shouldn't do it.

Your life would be made easier if you actually adopted the modal editing paradigm, instead of trying to make a modal editor into a non-modal editor.

danielrode:
Workaround:

```
[keys.insert]
C-s = ["commit_undo_checkpoint", ":write"]
```

omentic:
Fixed by #11062

PR #2893, opened 2022-06-27
Refactor handling of mouse events

sudormrfbin:
- Simplified match statements by destructuring `MouseEvent` struct at the top and then matching on `event.kind`.
- Extracted out closures for calculating (1) position and view of mouse click and (2) gutter coordinates and view of mouse click.


archseer (approved):
Much better 👍🏻

Issue #2856, opened 2022-06-22
Italics not reset for diagnostics text


mtoohey31:
### Summary

When language server diagnostic summary text is rendered on the screen in a spot where the file's text is italicized due to syntax highlighting, the diagnostic text is also italicized.

This screenshot demonstrates the issue:

![2022-06-21T22:07:28,186693692-04:00](https://user-images.githubusercontent.com/36740602/174928360-7399cf97-113f-450a-ac55-2eb630869391.png)

The words "not enough" in the diagnostics text are italicized, but the rest of the line isn't.

### Reproduction Steps

 

I tried this:

1. `hx <filename>` (with language server active)
2. Position window/cursor so diagnostics appear overlapping text that is italicized.

`config.toml`:

```toml
theme = "base16_terminal_kitty"
```

<details>
<summary><code>themes/base16_terminal_kitty.toml</code>:</summary>

```toml
# Authors: NNB <nnbnh@protonmail.com>, Matthew Toohey <contact@mtoohey.com>

"ui.menu" = { fg = "light-gray", bg = "opaque-black" }
"ui.menu.selected" = { modifiers = ["reversed"] }
"ui.linenr" = { fg = "gray", bg = "black" }
"ui.popup" = { bg = "opaque-black" }
"ui.window" = { bg = "opaque-black" }
"ui.linenr.selected" = { fg = "white", bg = "black", modifiers = ["bold"] }
"ui.selection" = { fg = "black", bg = "blue" }
"ui.selection.primary" = { fg = "white", bg = "blue" }
"comment" = { fg = "gray", modifiers = ["italic"] }
"ui.statusline" = { fg = "white", bg = "opaque-black" }
"ui.statusline.inactive" = { fg = "gray", bg = "opaque-black" }
"ui.help" = { fg = "white", bg = "opaque-black" }
"ui.cursor" = { fg = "light-gray", modifiers = ["reversed"] }
"ui.cursor.primary" = { fg = "light-gray", modifiers = ["reversed"] }
"ui.virtual" = "light-gray"
"variable" = "light-red"
"constant.numeric" = "yellow"
"constant" = "yellow"
"attributes" = "yellow"
"type" = "light-yellow"
"ui.cursor.match" = { fg = "light-yellow", modifiers = ["underlined"] }
"string"  = "light-green"
"variable.other.member" = "light-green"
"constant.character.escape" = "light-cyan"
"function" = "light-blue"
"constructor" = "light-blue"
[...]

PR #2900, opened 2022-06-27
fixes background reset

sbromberger:
Fixes https://github.com/helix-editor/helix/issues/2856#issuecomment-1166656545. Another set of eyes on this would be good, but here's how it looks (before / after; backgrounds deliberately garish):

Before:
IMG_0606

After:
IMG_0605

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -669,12 +669,14 @@ impl EditorView {
 
         let mut lines = Vec::new();
         for diagnostic in diagnostics {
-            let style = Style::reset().patch(match diagnostic.severity {
-                Some(Severity::Error) => error,
-                Some(Severity::Warning) | None => warning,
-                Some(Severity::Info) => info,
-                Some(Severity::Hint) => hint,
-            });
+            let style = Style::default().remove_modifier(Modifier::all()).patch(
+                match diagnostic.severity {
+                    Some(Severity::Error) => error,
+                    Some(Severity::Warning) | None => warning,
+                    Some(Severity::Info) => info,
+                    Some(Severity::Hint) => hint,
+                },
+            );
             let text = Text::styled(&diagnostic.message, style);
             lines.extend(text.lines);
         }


helix-term/src/ui/editor.rs
```diff
@@ -669,12 +669,14 @@ impl EditorView {
 
         let mut lines = Vec::new();
         for diagnostic in diagnostics {
-            let style = Style::reset().patch(match diagnostic.severity {
-                Some(Severity::Error) => error,
-                Some(Severity::Warning) | None => warning,
-                Some(Severity::Info) => info,
-                Some(Severity::Hint) => hint,
-            });
+            let style = Style::default().remove_modifier(Modifier::all()).patch(
```
pickfire, line 672: Would it be better if we patch it twice after reset? First with normal text background, then with diagnostic style.

sbromberger: I'm not sure I understand. This is intended to fix the reported problem by removing the modifiers for the diagnostics (prior to applying the diagnostic style, which can include modifiers) while leaving the other style elements untouched.

What would patching twice buy us here?

the-mikedavis: Yeah we could do

```rust
Style::reset().patch(background_style).patch(
  match {
    // ... that existing match expression ...
  },
)
```

I'm pretty sure that would be correct - I don't think there are any odd scenarios where we would want inherit a background that isn't the regular background, right 🤔?

sbromberger: Ah, I understand. OK. One sec.

the-mikedavis (approved):
I think this should do the trick 👍

helix-term/src/ui/editor.rs
archseer: `reset.patch(background)` should be identical to just using `background` though.

pickfire (approved):
Looks good to me.

helix-term/src/ui/editor.rs
the-mikedavis: Looking at `Style::reset`, it looks like it'll start out with a `sub_modifier` of `Modifier::all()` which I think is what we need to clear the italics. We could probably do:

```rust
background_style.remove_modifier(Modifier::all())
```

to be more clear that we only want to drop the modifiers from the background

sbromberger: Did you want me to follow up with this change, or are we all good at this point?

the-mikedavis: This looks like it's working well for me now so I say no need for any more changes 👍

sbromberger: Copy. Thanks, and sorry for the original oversight.

PR #561, opened 2021-08-08
Add command to set tab display width.

cessen:

PR #2905, opened 2022-06-28
Numeric window selection

freqmod:
These patches makes it possible to move focus to a window by pressing c+w and then a number like you do with ctrl+<digit> in sublime text, vscode etc. This is particularly useful if you have a wide/tall monitor and thus using the arrow keys would make you have to repeat the command to get to the desired window. 

Currently the numbering is based on the window vector list, which may not match the physical layout on the screen. However the displayed id and the window jumped to does match.

This is mostly a starting point for discussion. Especially the size of the window menu may be a bit verbose with this change.

I did look for an integration test, but i could not find any tests for other window commands to be inspired by and this is quite straightforward to i skipped it, at least for now until I know that the current implementation approach is desirable.

bilde


the-mikedavis:
Instead of separate commands for each index, this should be a command that takes the next character as input to determine which index to use. Examples: `mi`, `ma`, `f` and `t`.

sudormrfbin:
I think taking the count before would be better suited here, similar to goto line number `<n>gg`.

archseer:
Also for the number overlay, here's an idea from tmux: https://asciinema.org/a/KZAbKGDHxitvEmv0TJ90xD5Pm

freqmod:
I will look into how on_next_key works. Not sure about how to render number superimposed over the windows though (as the tmux image macro), i guess that may be a next step afterwords.

freqmod:
I rebased the PR on master, and added a commit for taking the index for the window as a parameter (now 1 indexed).

I am not sure what is the best approach to fix the clippy lints complaining about too many parameters to functions in editor.rs. Should i make a struct for passing View, the view index and if indices are visible as one struct, or is it better to just disable the lint cause for those function parameters?

archseer:
You can just tag that function with `#[allow(clippy::too_many_parameters)]` or whatever the lint name is.

freqmod:
> `#[allow(clippy::too_many_parameters)`

I guess, thanks for the pointer... I was mostly wondering what the policy of this project is, i.e. how many hoops to jump around to avoid too many arguments in the first place.

helix-term/src/commands.rs
```diff
@@ -3949,6 +3950,25 @@ fn transpose_view(cx: &mut Context) {
...
 
+fn goto_window_idx(cx: &mut Context) {
+    cx.editor.show_window_ids = true;
+
+    cx.on_next_key(move |cx, event| {
+        if let KeyEvent {
+            code: KeyCode::Char(ch),
+            ..
+        } = event
+        {
```
the-mikedavis, line 3961:
```suggestion
        if let Some(ch) = event.char() {
```

helix-term/src/commands.rs
```diff
@@ -3949,6 +3950,25 @@ fn transpose_view(cx: &mut Context) {
...
+            let idx_int = ch.to_digit(10).unwrap();
```
the-mikedavis, line 3962: This will panic if you hit something outside of 0-9. Instead we should say `cx.editor.set_error("Window index must be 1-9")` if the character isn't in 1-9.

the-mikedavis (commented):
This may need some refactoring in the statusline part since the merge of #2434. The window indicator can become a statusline element.

Overall, I think we should use "index" and "indices" over "idx" and "ids." I'm not a big abbreviation fan :P

helix-term/src/commands.rs
the-mikedavis: There's some slightly similar prior work around parsing usizes you can use as a reference https://github.com/helix-editor/helix/pull/561/files#diff-d782c07f22492bcae17333aba5dc5b840fb30317727e71267ebf4863f84787fdR1458-R1462

freqmod:
Thanks for the comments, and the hints on where to look for refactoring. I added my own conversion functions, to support 1-9 a-z for more views. I put them in helix/helix-view/src/lib.rs but i feel they are probably a bit out of place there too.

helix-view/src/lib.rs
```diff
@@ -68,3 +93,26 @@ pub use document::Document;
...
+        assert_eq!(view_index_to_identifier(9), Some('a'));
+        assert_eq!(view_index_to_identifier(34), Some('z'));
+        assert_eq!(view_index_to_identifier(35), None);
+        assert_eq!(view_identifier_to_index('0'), None);
+        assert_eq!(view_identifier_to_index('1'), Some(0));
+        assert_eq!(view_identifier_to_index('9'), Some(8));
+        assert_eq!(view_identifier_to_index('a'), Some(9));
+        assert_eq!(view_identifier_to_index('z'), Some(34));
+        assert_eq!(view_identifier_to_index('A'), Some(9));
+        assert_eq!(view_identifier_to_index('Z'), Some(34));
```
the-mikedavis, line 113: Lowercase and uppercase map to the same view index? I think that's surprising behavior

helix-view/src/tree.rs
```diff
@@ -529,6 +529,18 @@ impl Tree {
         }
     }
 
+    pub fn focus_view_idx(&mut self, idx: usize) {
+        // If the view doesn't exist, do nothing
+
+        let mut views = self.views().skip(idx);
+
+        let focus_view_id = views.next().map(|view| view.0.id);
```
the-mikedavis, line 537:
```suggestion
        let focus_view_id = self.views().nth(idx).map(|view| view.0.id);
```

book/src/keymap.md
```diff
@@ -231,23 +231,24 @@ TODO: Mappings for selecting syntax nodes (a superset of `[`).
...
+| `j`, `Ctrl-j`, `Down`  | Move to split below                                  | `jump_view_down`    |
+| `k`, `Ctrl-k`, `Up`    | Move to split above                                  | `jump_view_up`      |
+| `l`, `Ctrl-l`, `Right` | Move to right split                                  | `jump_view_right`   |
+| `q`, `Ctrl-q`          | Close current window                                 | `wclose`            |
+| `o`, `Ctrl-o`          | Only keep the current window, closing all the others | `wonly`             |
+| `H`                    | Swap window to the left                              | `swap_view_left`    |
+| `J`                    | Swap window downwards                                | `swap_view_down`    |
+| `K`                    | Swap window upwards                                  | `swap_view_up`      |
+| `L`                    | Swap window to the right                             | `swap_view_right`   |
+| `i` `<char>`           | Switch to window with index char (1-9) as shown      | `goto_window_idx` |
```
the-mikedavis, line 251: This should probably mention a-z now too

helix-term/src/commands.rs
```diff
@@ -4115,6 +4116,19 @@ fn transpose_view(cx: &mut Context) {
     cx.editor.transpose_view()
 }
 
+fn goto_window_idx(cx: &mut Context) {
```
the-mikedavis, line 4119:
```suggestion
fn goto_window_by_index(cx: &mut Context) {
```

helix-term/src/commands.rs
```diff
@@ -373,6 +373,7 @@ impl MappableCommand {
         swap_view_down, "Swap with split below",
         transpose_view, "Transpose splits",
         rotate_view, "Goto next window",
+        goto_window_idx, "Goto window with index 0",
```
the-mikedavis, line 376: This description should be updated

helix-view/src/lib.rs
frederik-vestre-pexip: i would not think that any user with have more than 34 windows up in practice, so therefore i thought there would be no need to have different windows for 0-9, a-z and then A-Z. Therefore I thought that it would be nice to make this case insensitive. However it may be cleaner to just handle lower case keys (and ignore capital letters). I can refactor the code to act this way.

helix-term/src/commands.rs
freqmod: I have tried to apply this

helix-view/src/tree.rs
freqmod: [ddress review comments and suggestions](https://github.com/helix-editor/helix/pull/2905/commits/ba97a7d301c31efb626bff23296202edc7a24f1f)

helix-term/src/commands.rs
freqmod: Updated

freqmod:
I see lint is failing. I will try fixing that tomorrow. If anyone wants to merge before that just run cargo fmt first.

freqmod:
I will be abandoning this PR. 

The current window numbering scheme is a bit lacking when moving windows around anyhow. I may be trying to get similar functionality working in helicone at some point in the future. Currently i am keeping this patch rebased in https://github.com/freqmod/helix as long as i am using the helix editor

PR #2906, opened 2022-06-28
nix: specify C compiler in Cargo metadata

the-mikedavis:
For me this fixes the `aarch64-darwin` build (I finally have one to test on 😀) via the Nix flake ❄️

I'm not totally sure why this was necessary. We have https://github.com/helix-editor/helix/blob/07e7a13a9e6b0a6c0bb456504b62572ec6be9eb2/flake.nix#L32-L36

that should say which compiler to use, but on master we end up using GCC to compile the tree-sitter dependency which fails. I poked around a little with a bisect but I haven't found a good build without this change yet.

This comes from https://github.com/yusdacra/nix-cargo-integration/issues/65#issuecomment-1077292510

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -14,6 +14,7 @@ default-run = "hx"
 [package.metadata.nix]
 build = true
 app = true
+cCompiler = "clang"
 
 [features]
 unicode-lines = ["helix-core/unicode-lines"]


the-mikedavis:
cc @yusdacra

I think we might have something wrong in our flake so we don't end up overriding the C compiler?

yusdacra:
I think this was `nix-cargo-integration`'s fault. I pushed a fix there, so could you update NCI and try again without your changes? Also FYI the NCI settings won't affect what compiler the grammars will use, so we need to change the stdenv for building grammars seperately in it's `mkDerivation` usage.

the-mikedavis:
Ah that does it! Thanks!

PR #2908, opened 2022-06-28
[Theme] Nord Light

two-six:
[Nord](https://www.nordtheme.com/) inspired light theme

Screenshot_20220629_012123

### DIFF
--- a/runtime/themes/nord_light.toml
+++ b/runtime/themes/nord_light.toml
@@ -1,0 +1,65 @@
+# Author: Two-Six<twopsix@duck.com>
+
+"ui.background" = {bg="bg"}
+"ui.text" = "black"
+"ui.selection" = {bg="bold_blue"}
+"ui.statusline" = {bg="bold_white"}
+"ui.statusline.inactive" = {bg="white"}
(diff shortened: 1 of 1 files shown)


archseer:
Does this follow the palette guidelines? https://www.nordtheme.com/docs/colors-and-palettes

From what I can tell red & orange are intended to only be used for UI elements and warnings. Compare this to the VSCode version: https://vscode.dev/theme/huytd.nord-light/Nord%20Light

two-six:
> Does this follow the palette guidelines? https://www.nordtheme.com/docs/colors-and-palettes
> 
> From what I can tell red & orange are intended to only be used for UI elements and warnings. Compare this to the VSCode version: https://vscode.dev/theme/huytd.nord-light/Nord%20Light

Haven't seen that one. I actually used `Doom Nord Light` as a reference:
image

I'll try to work with these palette guidelines and show how it looks like then.

two-six:
There's updated version(cursorline doesn't work in the screenshot, but it's probably my fault). It's based on the guidelines mentioned above:
image
It definitely has way less contrast than the previous one

Issue #2882, opened 2022-06-26
Git commit message length based highlighting


A-Walrus:
#### Describe your feature request
When using helix as the editor for git I want the commit message to have highlighting which indicates whether it is too long (over 50 characters).

This is available in Neovim like this: 
![image](https://user-images.githubusercontent.com/58790821/175819771-b32a7e75-42a6-40a2-87f6-3f0c3636a122.png)

I think this is a useful feature for making sure your messages are not too long.

archseer:
It's not the same but you could use the ruler feature to indicate a max line length

the-mikedavis:
I tried this out initially in tree-sitter-git-commit but I found that tree-sitter's compilation time for the grammar shot way up when using regular expressions with count quantifiers like

```js
choice(alias(/.{1,50}/, $.short_message), alias(/.{51,}/, $.long_message))
```

I do want this kind of highlight though. I'll take a look at how bad that performance really gets and if it affects runtime performance or just compile-time. We can also try doing this by changing the queries instead of the grammar.

the-mikedavis:
It's possible to do this if you'd like to override your `git-commit` `highlights.scm`: https://github.com/the-mikedavis/helix/commit/e2b8506cca5cf6957cfa24b5179a070ee19ae37d

In order for this to be a good core feature we'd like to have it be configurable:

- enabled / disabled
- the number of characters it takes to go from one mode to the next

even if it's enabled by default (which is not a given).

Once we change configuration languages from TOML to something more general purpose it should be easier to support this[^1]. In the meantime, the rulers feature is a reasonable way to guide commit message lengths (though not perfect because of the different good-practice lengths of messages and bodies).

[^1]: I'm thinking that along with a plugin system and/or more general purpose config language that it should be easier to control which tree-sitter queries are used via configuration.

PR #2910, opened 2022-06-29
improve highlights across languages

the-mikedavis:
This is a party pack of small highlights improvements for:

- erlang
- markdown
- bash
- rust
- html
- heex
- make
- tsq (tree-sitter-query)
- git-commit

Included is a fix for #2882. It's a bit different from the neovim version: we just switch scopes when the commit message gets too long - first to `@warning` and then `@error`.

The only change I think is controversial is 83d2c1e302b7ce4243bd9b5d66d967fece94e87f - people might prefer we always use `@markup.raw.block` as a background for code-fences even when we inject other languages. Let me know what you think!


poliorcetics:
About https://github.com/helix-editor/helix/pull/2910/commits/e2b8506cca5cf6957cfa24b5179a070ee19ae37d, I don't agree at all with this. Personally I limit my commit messages length to 80 or 100, never to 50 characters and being told this in an error by the editor will make it so helix is forcing git oudated (to me) recommendations. I can understand other peoples wanting it but this should be configurable, not absolute

archseer:
Also, isn't configuring a `ruler` enough? It won't color your text when you go over, but it should still work all the same.

> to 80 or 100, never to 50 characters

Well, it's 50 for the heading, then the body is traditionally formatted to 80 to fit on terminal screens.

> people might prefer we always use `@markup.raw.block` as a background for code-fences even when we inject other languages.

Yeah, I think we should keep that scope? That way backgrounds can be styled. Why was it removed, did you have clashing styling?

poliorcetics:
> Well, it's 50 for the heading, then the body is traditionally formatted to 80 to fit on terminal screens.

I'm speaking for the heading too

the-mikedavis:
Hmm yeah it doesn't feel great to hard-code that behavior into the query. Ideally once we have a better config language we can make it a configurable feature (enabled/disabled plus the widths to use). But also a ruler does feel like the appropriate tool, at least for now. I'll leave a comment on #2882 and drop that out of this PR.

> That way backgrounds can be styled. Why was it removed, did you have clashing styling?

Oh I didn't think to style `markup.raw.block` only with `bg` 😅. Yeah I'll drop that change. I didn't like un-highlighted text like shell code-fences for example inheriting the `fg` of `markup.raw.block` but just using `bg` instead solves that.

```bash
git clone git@github.com:helix-editor/helix.git
# <- function
#     ^ inherited
#          ^ inherited
```

archseer (approved):

PR #2916, opened 2022-06-29
Adding a cursorline for the Nord theme

heliostatic:


two-six:
Didn't actually know it was there, I think [themes section](https://docs.helix-editor.com/themes.html) needs to be updated

heliostatic:
> Didn't actually know it was there, I think [themes section](https://docs.helix-editor.com/themes.html) needs to be updated

It is updated in the docs (https://github.com/helix-editor/helix/blob/master/book/src/themes.md?plain=1#L237-L238) but not the webpage. I believe that's because the webpage reflects the latest release, not master.

the-mikedavis:
Master docs are published to a different URL: https://docs.helix-editor.com/master/themes.html

(`master/` precedes the rest of the url)

PR #2921, opened 2022-06-30
feat(line-numbers): add total line numbers to position content

joshpauline:
I have been using helix as my main power house for work recently, however there are a couple things that would really like to integrate. The first of which is having a total line number indicator in the editor. 

Some may disagree but I like having it there.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -834,7 +834,12 @@ impl EditorView {
                 .cursor(doc.text().slice(..)),
         );
         right_side_text.0.push(Span::styled(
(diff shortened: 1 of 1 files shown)


EpocSquadron:
Perhaps this could be rebased on to #2434 as I could see people wanting this to be configurable.

heliostatic:
Adding a screenshot for more context.

Existing: 
2022-06-30 at 09 15 35 - Alacritty - CleanShot

Proposed:
2022-06-30 at 09 09 51 - Alacritty - CleanShot

the-mikedavis:
Yeah I think it would be best for this to await #2434 so it can be configurable.

joshpauline:
yea ok the configurable status line would be great.

PR #2935, opened 2022-07-01
Only highlight inner area of view with cursorline

### LINKED ISSUE #2912: Restrict cursorline to editing area
The newly merged cursorline currently is drawn across the whole window, including the gutter. I think it would look better if the cursorline is only drawn in the editing area, that's how it's done in VSCode, Vim, etc. We also already can set the background (for the numbers) with `ui.linenr.selected`, so it's a bit redundant.

A-Walrus:
Fixes #2912

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -717,9 +717,9 @@ impl EditorView {
 
         for line in view.offset.row..(last_line + 1) {
             let area = Rect::new(
-                view.area.x,
-                view.area.y + (line - view.offset.row) as u16,
-                view.area.width,
+                view.inner_area().x,
+                view.inner_area().y + (line - view.offset.row) as u16,
+                view.inner_area().width,
                 1,
             );
             if primary_line == line {
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -715,11 +715,12 @@ impl EditorView {
         let primary_style = theme.get("ui.cursorline.primary");
         let secondary_style = theme.get("ui.cursorline.secondary");
 
+        let inner_area = view.inner_area();
         for line in view.offset.row..(last_line + 1) {
             let area = Rect::new(
-                view.inner_area().x,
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -717,9 +717,9 @@ impl EditorView {
 
         for line in view.offset.row..(last_line + 1) {
             let area = Rect::new(
-                view.area.x,
-                view.area.y + (line - view.offset.row) as u16,
-                view.area.width,
+                view.inner_area().x,
+                view.inner_area().y + (line - view.offset.row) as u16,
+                view.inner_area().width,
```
archseer, line 722: `let inner_area = view.inner_area()` before this line. Otherwise you're calculating it three times.

archseer (changes requested):

sudormrfbin (approved):

archseer:
I actually like the current behavior though.

ChrHorn:
> I actually like the current behavior though.

With this PR it is still possible to set `ui.linenr.selected` background color same as cursorline to achieve the current behavior (minus the diagnostics and the gap).

archseer:
Yeah it's not the same, it won't highlight all the gutters.

ChrHorn:
> Yeah it's not the same, it won't highlight all the gutters.

I see. What if we just use `ui.linenr(.selected)` for the whole gutter and rename it to `ui.gutter(.selected)` (`ui.gutter` is already used but not documented). Think this looks better in general and would allow both styles for the cursorline.

Could also use a more fine-grained approach where every selected component can be themed, but I'm not sure if this is really needed.

archseer:
`ui.gutter.selected` is a good idea, but I'd also keep `ui.linenr.selected` so that only the line number foreground can be changed.

pascalkuthe:
closing this one out as stale, the entire rendering code has been replaced (and will see further large changes) so it would be easier to start over with this one. Thank ou for contributing!

PR #2939, opened 2022-07-01
Highlight whole row in picker menus

ChrHorn:
This is for themes that use `bg` to highlight the selection in picker menus. 
Makes it easier to navigate, looks cleaner, and is in line with the style of the autocomplete menu.

before:
image

after:
image

Not 100% sure if this is still needed:
https://github.com/helix-editor/helix/blob/fbb200b7ddd3f70bdc8f76b6c53f030d3990dc62/helix-term/src/ui/picker.rs#L638-L639


sudormrfbin (approved):
Thanks!

Issue #1673, opened 2022-02-16
autopair: delete pair on backspace


tgirod:
When in insert mode, autopair gets triggered when a pairable character is entered. But if I press a pairable character by mistake and delete it using backspace, the closing pair is not deleted.

I think it would make more sense to automatically delete the closing character in that situation.

sudormrfbin:
There's an open PR: https://github.com/helix-editor/helix/pull/1379.

tgirod:
good to know, thanks!

MonAaraj:
This should not be closed, should it? Since the issue hasn't been solved yet. It's good to keep it open for people who run into this issue and see no open issues for it, then attempting to make one for it (like me just now). Sorry for the off-topic comment. And also, +1, I guess, since I am commenting on this anyways.

PR #1379, opened 2021-12-26
Auto pairs delete

dead10ck:
I began trying to implement deleting pairs on backspace, and I ran into an issue that I think is best demonstrated by showing what I've tried first. This PR is mainly just context for a discussion at this stage; it made me think a more invasive change may be necessary to support it.

Looking over the current code for `delete_char_backward`, I can see that there is code there for handling dedenting as well. In order to preserve this behavior, we need to basically check for both conditions on a backspace: if the line is all whitespace to dedent, and if not, then check if both sides of the cursor are a pair, and finally proceeding to a regular single character deletion.

Now, the current code for auto pairs works by putting together an entire transaction, assuming its logic will either create the entire change or none of it. However, with multiple ranges in the selection, one cursor could end up being in position for an auto pair deletion while another is on a line that only contains whitespace, and thus should dedent. In order to support this, we need to produce changes for *individual ranges*, not on the whole selection.

So the PR here is what I tried to get it working. And it does work, but only for one-width cursors. If it's in a selection, the head ends up moving back two characters, when it should only move back once, e.g. `[word(]) -> [wor]d`. I think this is the same reason that the current auto pairs code works with entire `Transaction`s, since by doing this, we can calculate the correct resulting selection as well as the textual changes to the document.

So we have a conflict here: auto pairs needs to change selections as well as text, but we also cannot control the entire transaction, because auto pairs is not responsible for all of the changes that could happen.

So I had an idea that maybe [`Change`](https://github.com/helix-editor/helix/blob/a306a1052a51c686b24a6f339190878b8029a894/helix-core/src/transaction.rs#L5) could become a 4-tuple, with the last being an `Option<Range>` that explicitly sets what the new range should be after the textual change. Then we could have [`Transaction::change`](https://github.com/helix-editor/helix/blob/a306a1052a51c686b24a6f339190878b8029a894/helix-core/src/transaction.rs#L474) build up a selection as it iterates through the `Change`s. Then we could build the transactions the same way, but we'd get selection manipulation too.

So the hook functions would return `Option<Change>`, and operate only on a single range of text. I believe this would also, by extension, take care of the old todo that the auto pairs insert hook should just fall back on the default insert hook when it won't insert a pair.

What do you think?

PR #2940, opened 2022-07-02
Auto pair-removal

Houkime:
Fixes https://github.com/helix-editor/helix/issues/1673
Supersedes https://github.com/helix-editor/helix/pull/1379
No problems with multiple selections observed.
Putting match inside tuple would hurt readabilty - these tuples already take a while to comprehend.

Simple logic inspired by Godot 3.x stable branch:
https://github.com/godotengine/godot/blob/14f69acaa4ef80b235ef78f9caac1072daf8008b/scene/gui/text_edit.cpp#L2101
https://github.com/godotengine/godot/blob/14f69acaa4ef80b235ef78f9caac1072daf8008b/scene/gui/text_edit.cpp#L2054

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2937,10 +2937,11 @@ pub mod insert {
 
     pub fn delete_char_backward(cx: &mut Context) {
         let count = cx.count();
-        let (view, doc) = current!(cx.editor);
+        let (view, doc) = current_ref!(cx.editor);
         let text = doc.text().slice(..);
         let indent_unit = doc.indent_unit();
         let tab_size = doc.tab_width();
+        let auto_pairs = doc.auto_pairs(cx.editor);
 
         let transaction =
             Transaction::change_by_selection(doc.text(), doc.selection(view.id), |range| {
@@ -2992,9 +2993,14 @@ pub mod insert {
                         (start, pos, None) // delete!
                     }
                 } else {
-                    match (text.get_char(pos.saturating_sub(1)), text.get_char(pos)) {
-                        (Some(_x), Some(_y)) if auto_pairs::DEFAULT_PAIRS.contains(&(_x, _y)) =>
-                        // delete both
+                    match (
+                        text.get_char(pos.saturating_sub(1)),
+                        text.get_char(pos),
+                        auto_pairs,
+                    ) {
+                        (Some(_x), Some(_y), Some(ap))
+                            if ap.get(_x).is_some() && (ap.get(_x).unwrap().close == _y) =>
+                        // delete both autopaired characters
                         {
                             (
                                 graphemes::nth_prev_grapheme_boundary(text, pos, count),
@@ -3014,6 +3020,7 @@ pub mod insert {
                     }
                 }
             });
+        let (view, doc) = current!(cx.editor);
         doc.apply(&transaction, view.id);
     }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2999,7 +2999,7 @@ pub mod insert {
                         auto_pairs,
                     ) {
                         (Some(_x), Some(_y), Some(ap))
-                            if range.len() == 1
+                            if range.is_single_grapheme(text)
                                 && ap.get(_x).is_some()
                                 && ap.get(_x).unwrap().close == _y =>
                         // delete both autopaired characters
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2999,7 +2999,9 @@ pub mod insert {
                         auto_pairs,
                     ) {
                         (Some(_x), Some(_y), Some(ap))
-                            if ap.get(_x).is_some() && (ap.get(_x).unwrap().close == _y) =>
+                            if range.len() == 1
+                                && ap.get(_x).is_some()
+                                && ap.get(_x).unwrap().close == _y =>
                         // delete both autopaired characters
                         {
                             (
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2992,12 +2992,26 @@ pub mod insert {
                         (start, pos, None) // delete!
                     }
                 } else {
-                    // delete char
-                    (
-                        graphemes::nth_prev_grapheme_boundary(text, pos, count),
-                        pos,
-                        None,
-                    )
+                    match (text.get_char(pos.saturating_sub(1)), text.get_char(pos)) {
+                        (Some(_x), Some(_y)) if auto_pairs::DEFAULT_PAIRS.contains(&(_x, _y)) =>
+                        // delete both
+                        {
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -2992,12 +2992,26 @@ pub mod insert {
...
                 } else {
-                    // delete char
-                    (
-                        graphemes::nth_prev_grapheme_boundary(text, pos, count),
-                        pos,
-                        None,
-                    )
+                    match (text.get_char(pos.saturating_sub(1)), text.get_char(pos)) {
+                        (Some(_x), Some(_y)) if auto_pairs::DEFAULT_PAIRS.contains(&(_x, _y)) =>
+                        // delete both
```
archseer, line 2997: You should note here that it's related to the autopairs functionality

helix-term/src/commands.rs
```diff
@@ -2992,12 +2992,26 @@ pub mod insert {
...
                     }
                 } else {
-                    // delete char
-                    (
-                        graphemes::nth_prev_grapheme_boundary(text, pos, count),
-                        pos,
-                        None,
-                    )
+                    match (text.get_char(pos.saturating_sub(1)), text.get_char(pos)) {
+                        (Some(_x), Some(_y)) if auto_pairs::DEFAULT_PAIRS.contains(&(_x, _y)) =>
```
archseer, line 2996: Use autopairs provided by the document/language rather than the default set:

https://github.com/helix-editor/helix/blob/290b3ebbbe0c365eee436b9de9d6d6fc2b4339e9/helix-term/src/commands.rs#L2816

archseer (changes requested):

archseer:
\cc @dead10ck

dead10ck:
This implementation is pretty much the same exact mechanism as #1379 (i.e. look ahead and behind one char and return a delete, see [here](https://github.com/helix-editor/helix/pull/1379/files#diff-50a1c8d85a9ae0f5840c9cf42a8de23831e9fa1c8454573136cdeb2deb0f48ddR223)) and so has exactly the same problem:

>it does work, but only for one-width cursors. If it's in a selection, the head ends up moving back two characters, when it should only move back once, e.g. `[word(]) -> [wor]d`

The implementation of this should live in the `auto_pairs` module so they it can also reuse all the logic that ended up being necessary to handle computing the right selection. See [here](https://github.com/helix-editor/helix/blob/9f43dbc45d254ff860f7d33db201b2f0a6073195/helix-core/src/auto_pairs.rs#L150). The problem with this is as I described in #1379.

I would also want to see test coverage to verify that the intersection of all text edit cases works, as I described in [this comment](https://github.com/helix-editor/helix/pull/1379#issuecomment-1003724997). I'm afraid the solution won't be this simple. It's why I've delayed trying until after we have an integration testing suite, which we have now. I plan on working on this after I finish up the write path fixes in #2267, if no one gets to it before me.

archseer:
Yeah it doesn't fully fix it but the plan was to at least implement this for 1-width selections which should solve this for majority editing of cases.

dead10ck:
> Yeah it doesn't fully fix it but the plan was to at least implement this for 1-width selections which should solve this for majority editing of cases.

With respect, I don't think this is as uncommon as you may realize. The test case above is going to happen any time the user enters append mode, or does a select all -> search for text, and then deletes chars and runs into a pair, e.g. empty strings, function calls with no arguments, slices, etc, etc. And it may not even be visible to them if it happens off screen.

dead10ck:
On the other hand, I saw it suggested in the chat room that we could make this behavior only happen specifically when the range is single width, otherwise fall back to the regular single char delete. If that constraint were added, this would indeed be a good workaround until it can be fixed in a more general way for selections too.

helix-term/src/commands.rs
Houkime: done

helix-term/src/commands.rs
Houkime: done in https://github.com/helix-editor/helix/pull/2940/commits/e7e591e44095e3e4a898ed1885df04b6dcc2b3c3

Houkime:
> On the other hand, I saw it suggested in the chat room that we could make this behavior only happen specifically when the range is single width, otherwise fall back to the regular single char delete. If that constraint were added, this would indeed be a good workaround until it can be fixed in a more general way for selections too.

done in https://github.com/helix-editor/helix/pull/2940/commits/005f4c6b484aa47459b42f8421cd4b3cfada2fa8

helix-term/src/commands.rs
```diff
@@ -2992,14 +2993,36 @@ pub mod insert {
...
+                        auto_pairs,
+                    ) {
+                        (Some(_x), Some(_y), Some(ap))
+                            if range.len() == 1
+                                && ap.get(_x).is_some()
+                                && ap.get(_x).unwrap().close == _y =>
+                        // delete both autopaired characters
+                        {
+                            (
+                                graphemes::nth_prev_grapheme_boundary(text, pos, count),
```
dead10ck, line 3008: This might get a bit tricky with the count. If we just delete n chars in both directions, this will be wrong in some cases. e.g. presently with this case:

```
hello, "["], goodbye("")
```

If you hit backspace 3 times, you get

```
hello[,] goodbye("")
```

like you'd expect. But if you use the command with a count, e.g. if you bound backspace to this in normal mode and then hit `3<backspace>`, you end up with:

```
hello[g]oodbye("")
```

Though, I don't know how many people bind this in normal mode and use the count with it, so I don't know if I'd consider this a blocker. It might be tricky to get the logic right.

helix-term/src/commands.rs
```diff
@@ -2992,14 +2993,36 @@ pub mod insert {
...
-                        pos,
-                        None,
-                    )
+                    match (
+                        text.get_char(pos.saturating_sub(1)),
+                        text.get_char(pos),
+                        auto_pairs,
+                    ) {
+                        (Some(_x), Some(_y), Some(ap))
+                            if range.len() == 1
```
dead10ck, line 3002: Might be better to use [`Range::is_single_grapheme`](https://github.com/helix-editor/helix/blob/9f43dbc45d254ff860f7d33db201b2f0a6073195/helix-core/src/selection.rs#L344) so that it also works when the cursor is over a non-ASCII grapheme.

dead10ck (approved):
This looks fine to me. A couple of minor comments, neither of which I'd consider blockers, but would be good changes to make. Integration tests would be even better.

Thanks for your contribution!

dead10ck:
One minor issue I just noticed: if a pair is at the very beginning of the file, and the cursor is on the first char, it will delete the first char instead of doing nothing. This might be fixable by checking if the cursor is the first char in the document.

dead10ck (approved):
Ok, tested locally, this looks good to me! Thanks for this fix! Can't wait to use it.

Issue #2362, opened 2022-05-01
Change default formatter for any language


frexsdev:
# Idea

Adding an option like this:

```toml
[[lang]]
name="yeah"
formatter="youhou" // command to execute when formatting instead of the lsp' default 
```

# Why ?

Because, for example, whe **CAN'T** format any Zig code with **ZLS** (just an example 😡)

the-mikedavis:
This needs a bit of design because currently auto-formatting is only done through the language server. We can't just add a configuration for a formatter binary because it would fight with the language server for any language that has a formatter binary and LSP format.

This probably depends on https://github.com/helix-editor/helix/issues/1125

kirawi:
Could we disable the formatting request if a binary is defined for formatting?

colaxdev:
Yes I think

thomas-profitt:
As a workaround, to manually run an external program to reformat the currently-open file in Helix:
1. `%` to select everything (or only the part you want to reformat)
2. `|` to pipe the selection to an external program's stdin and replace it with the program's stdout,
3. `mix format -` (if it's an elixir file, for example)

kirawi:
I think that what we can do is:
- Add a config option in `languages.toml` to define a formatter command, and disable LSP formatting requests in `Document::format()` if a formatter is defined.
- Send the contents of the buffer to the formatter, and use `helix-core::diff::compare_ropes()` so we don't get a problem like #2404.

PR #2942, opened 2022-07-02
Change default formatter for any language

PiergiorgioZagaria:
I have never worked on a big project in rust, so I don't know if the `Result`/`Option` handling is ok.

This PR aims to allow the user to choose which formatter to use instead of always relying on lsp. Discussion about this in issue  #2362.

Currently the only problem is that write_all_impl saves only the viewed buffers, this is because the formatter needs to reload or apply the transaction to the document. How do we get around this?

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -136,7 +136,8 @@ pub struct FormatterConfiguration {
     #[serde(default)]
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub args: Vec<String>,
-    pub r#type: FormatterType,
+    #[serde(rename = "type")]
+    pub type_: FormatterType,
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -138,13 +138,6 @@ pub struct FormatterConfiguration {
     pub args: Vec<String>,
 }
 
-#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
-#[serde(rename_all = "kebab-case")]
-pub enum FormatterType {
-    Stdio,
-    File,
-}
-
 #[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct AdvancedCompletion {
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -136,8 +136,6 @@ pub struct FormatterConfiguration {
     #[serde(default)]
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub args: Vec<String>,
-    #[serde(rename = "type")]
-    pub type_: FormatterType,
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -141,6 +141,7 @@ pub struct FormatterConfiguration {
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
 pub enum FormatterType {
     Stdio,
     File,
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -136,7 +136,7 @@ pub struct FormatterConfiguration {
     #[serde(default)]
     #[serde(skip_serializing_if = "Vec::is_empty")]
     pub args: Vec<String>,
-    pub formatter_type: FormatterType,
+    pub r#type: FormatterType,
 }
 
 #[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -129,7 +129,7 @@ pub struct LanguageServerConfiguration {
     pub language_id: Option<String>,
 }
 
-#[derive(Debug, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct FormatterConfiguration {
     pub command: String,
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -79,6 +79,9 @@ pub struct LanguageConfiguration {
     #[serde(default)]
     pub auto_format: bool,
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub formatter: Option<FormatterConfiguration>,
+
     #[serde(default)]
     pub diagnostic_severity: Severity,
 
@@ -124,6 +127,22 @@ pub struct LanguageServerConfiguration {
     #[serde(default = "default_timeout")]
     pub timeout: u64,
     pub language_id: Option<String>,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct FormatterConfiguration {
+    pub command: String,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub args: Vec<String>,
+    pub formatter_type: FormatterType,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
+pub enum FormatterType {
+    Stdio,
+    File,
 }
 
 #[derive(Debug, PartialEq, Clone, Deserialize, Serialize)]
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -268,8 +268,11 @@ fn new_file(
 fn format(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
-    _event: PromptEvent,
+    event: PromptEvent,
 ) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
     let (view, doc) = current!(cx.editor);
     if let Some(format) = doc.format(view.id) {
         let callback =
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2,10 +2,7 @@ use std::ops::Deref;
 
 use super::*;
 
-use helix_view::{
-    document::FormatterError,
-    editor::{Action, ConfigEvent},
-};
+use helix_view::editor::{Action, ConfigEvent};
 use ui::completers::{self, Completer};
 
 #[derive(Clone)]
@@ -216,23 +213,17 @@ fn write_impl(
         bail!("cannot write a buffer without a filename");
     }
     let fmt = if auto_format {
-        match doc.auto_format(view.id) {
-            Ok(fmt) => {
-                let shared = fmt.shared();
-                let callback = make_format_callback(
-                    doc.id(),
-                    doc.version(),
-                    Modified::SetUnmodified,
-                    shared.clone(),
-                );
-                jobs.callback(callback);
-                Some(shared)
-            }
-            Err(error) => match error.downcast_ref::<FormatterError>() {
-                Some(_) => None,
-                None => return Err(error),
-            },
-        }
+        doc.auto_format(view.id).map(|fmt| {
+            let shared = fmt.shared();
+            let callback = make_format_callback(
+                doc.id(),
+                doc.version(),
+                Modified::SetUnmodified,
+                shared.clone(),
+            );
+            jobs.callback(callback);
+            shared
+        })
     } else {
         None
     };
@@ -280,16 +271,10 @@ fn format(
     _event: PromptEvent,
 ) -> anyhow::Result<()> {
     let (view, doc) = current!(cx.editor);
-    match doc.format(view.id) {
-        Ok(format) => {
-            let callback =
-                make_format_callback(doc.id(), doc.version(), Modified::LeaveModified, format);
-            cx.jobs.callback(callback);
-        }
-        Err(error) => match error.downcast_ref::<FormatterError>() {
-            Some(_) => {}
-            None => return Err(error),
-        },
+    if let Some(format) = doc.format(view.id) {
+        let callback =
+            make_format_callback(doc.id(), doc.version(), Modified::LeaveModified, format);
+        cx.jobs.callback(callback);
     }
 
     Ok(())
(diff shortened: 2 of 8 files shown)


PiergiorgioZagaria:
Another bug I found using `zig fmt --stdin`, the editor hangs because the formatter doesn't understand that stdin has wrote everything. I tried pushing a 0 to the collected vector and flushing stdin, but neither seems to work.

The expected usage of `zig fmt --stdin` is: `$ cat sometext.zig | zig fmt --stdin`

archseer:
You probably need to close stdin when you're done.

PiergiorgioZagaria:
> You probably need to close stdin when you're done.

Yeah that was it, thanks

helix-view/src/document.rs
```diff
@@ -397,17 +398,103 @@ impl Document {
...
         }
     }
 
     /// If supported, returns the changes that should be applied to this document in order
     /// to format it nicely.
-    pub fn format(&self) -> Option<impl Future<Output = LspFormatting> + 'static> {
+    pub fn format(
+        &mut self,
+        view_id: ViewId,
+    ) -> Option<impl Future<Output = LspFormatting> + 'static> {
```
sudormrfbin, line 417: Changing the return type from `Option` to `Result` would be more appropriate here now, since multiple things could go wrong here and the there would be no feedback on the screen it anything errors out. Any error can be displayed in the editor by the calling functions instead of being simply logged.

helix-core/src/syntax.rs
```diff
@@ -126,6 +129,22 @@ pub struct LanguageServerConfiguration {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct FormatterConfiguration {
+    pub command: String,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub args: Vec<String>,
+    pub formatter_type: FormatterType,
```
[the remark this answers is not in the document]
PiergiorgioZagaria: I tried that before and since `type` is a keyword I would have to use `r#type` which I think looks uglier.  If I were to use the `r#type`, would I access the field with `formatter.type` or would I still need r#?

helix-view/src/document.rs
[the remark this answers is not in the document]
PiergiorgioZagaria: I have never used async in rust so I am a bit cautious in touching it, but wouldn't moving from `LspFormatting` to `Into<Transaction>` break the lsp callback?

helix-core/src/syntax.rs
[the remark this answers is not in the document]
PiergiorgioZagaria: I guess since it's used once in the whole project we can afford to use `r#type`

helix-view/src/document.rs
PiergiorgioZagaria: Would I just remove the `.ok()` and return the errors directly like that? Or should I make errors specific like when using `log::error!` ?

PiergiorgioZagaria:
Thanks for the style guides btw.

helix-core/src/syntax.rs
```diff
@@ -126,6 +129,22 @@ pub struct LanguageServerConfiguration {
...
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct FormatterConfiguration {
+    pub command: String,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub args: Vec<String>,
+    pub r#type: FormatterType,
```
sudormrfbin, line 139: You can also name the field something like `type_` and then use [serde's rename feature](https://serde.rs/field-attrs.html#rename):

```suggestion
	#[serde(rename = "type")]
    pub type_: FormatterType,
```

helix-view/src/document.rs
sudormrfbin: We definitely need the context added by the log commands, so you could use `anyhow::Result` for the return type and `result.map_err(|e| anyhow!("previous log description"))?` where ever `option.ok()` was being used.

PiergiorgioZagaria:
I found another bug. Basically, when using Stdio as the formatter type, when saving, the document is formatted and saved, but the editor shows that the file is still unsaved

helix-view/src/document.rs
```diff
@@ -397,30 +398,111 @@ impl Document {
...
     }
 
     /// If supported, returns the changes that should be applied to this document in order
     /// to format it nicely.
-    pub fn format(&self) -> Option<impl Future<Output = LspFormatting> + 'static> {
-        let language_server = self.language_server()?;
+    pub fn format(
+        &mut self,
+        view_id: ViewId,
+    ) -> anyhow::Result<Option<impl Future<Output = LspFormatting> + 'static>> {
```
sudormrfbin, line 422: Oh earlier when I mentioned using Result I meant _replacing_ Option with Result, not wrapping it with Result.

helix-view/src/document.rs
```diff
@@ -397,30 +398,111 @@ impl Document {
...
+                        .spawn()
+                        .map_err(|e| {
+                            anyhow!("Failed to format with {}. {}", formatter.command, e)
+                        })?;
+
+                    {
+                        let mut stdin = process.stdin.take().unwrap();
+                        stdin
+                            .write_all(&self.text().bytes().collect::<Vec<u8>>())
+                            .unwrap();
```
sudormrfbin, line 442: Unwraps should be avoided where possible, and since we have Result as return type `?` can be used instead.

helix-view/src/document.rs
PiergiorgioZagaria: Yeah, but then how do I distinguish between, don't return a future because the function failed vs don't return a future because I have a custom formatter? Would I return an error for executing the formatter correctly? What if there is no formatter at all?

kirawi: We can check what type of error is being returned from the result, and propagate it if it's an actual error.

PiergiorgioZagaria:
I don't know anything about async, ideas on how to make this work?
```
// let fut = async move {
//     helix_core::diff::compare_ropes(self.text(), &Rope::from_str(str))
// }
// return Ok(fut);
```
line 458 of document.rs.
(Btw how do I put the reference to the code in the comment?)

frexsdev (approved):

sudormrfbin:
> I don't know anything about async, ideas on how to make this work?
> 
> ```
> // let fut = async move {
> //     helix_core::diff::compare_ropes(self.text(), &Rope::from_str(str))
> // }
> // return Ok(fut);
> ```
> 
> line 458 of document.rs.

`Document::format()` originally returned an `Option<impl Future<...>>` which used the [`impl Trait`](https://stackoverflow.com/questions/48545035/what-does-impl-mean-when-used-as-the-argument-type-or-return-type-of-a-functio) syntax in the return type of the function. This means that the function will return some type implementing `Future`, but it also enforces that we can only really return a _single_ concrete type (see this SO post about [using `impl Trait` syntax for returning multiple types](https://stackoverflow.com/questions/52001592/why-can-impl-trait-not-be-used-to-return-multiple-conditional-types)).

Now, the actual problem is that [different async blocks generate different types](https://stackoverflow.com/questions/71070434/expected-async-block-found-a-different-async-block) that implement `Future` and therefore we can't use `impl Future` in our return type. We can use a trait object to solve this, and a [`BoxFuture`](https://docs.rs/futures-util/latest/futures_util/future/type.BoxFuture.html) is simply a type alias for `Pin<Box<dyn Future<Output = ...>>>` provided by the `futures_util` crate (which is what I ended up using).

> (Btw how do I put the reference to the code in the comment?)

https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-a-permanent-link-to-a-code-snippet

PiergiorgioZagaria:
Thanks for all the help and the explanation.

I guess the only problem now is the fact that write_all only writes to viewed buffers instead of all buffers, because we need the view for reloading the document if we are using the file formatter.

Maybe we could remove the file formatter type entirely? It depends on whether there are other known formatters that only work like that.

PiergiorgioZagaria:
Since I haven't found any formatter that doesn't support the `stdio` type, I suggest removing formatter type completely and use only `stdio`.
If there is a need for a `file` formatter type it should be implemented in a separate PR.

helix-core/src/syntax.rs
```diff
@@ -126,6 +129,22 @@ pub struct LanguageServerConfiguration {
...
+pub struct FormatterConfiguration {
+    pub command: String,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub args: Vec<String>,
+}
+
+#[derive(Debug, Clone, Copy, Eq, PartialEq, PartialOrd, Ord, Deserialize, Serialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum FormatterType {
```
kirawi, line 143: This enum should also be removed.

helix-term/src/commands/typed.rs
```diff
@@ -551,6 +551,7 @@ fn write_all_impl(
     let auto_format = cx.editor.config().auto_format;
     let jobs = &mut cx.jobs;
     // save all documents
+    // Saves only visible buffers? How do we reload the not visible ones?
```
sudormrfbin, line 554: This does save all documents, and on testing the non-visible ones are updated correctly too.

sudormrfbin (commented):
Looking good to merge 🎉

helix-term/src/commands/typed.rs
PiergiorgioZagaria: Ah yeah that was from before we used transactions, will remove the comment in a bit

samholmes:
This PR should include changes to documentation?

PiergiorgioZagaria:
Probably we should mention it in the languages configuration section

samholmes:
Can we access the current file name for the formatter config? This is needed for `eslint`:

```toml
formatter = { command = "eslint", args = ["--fix", "$HELIX_FILE"] }
```

Is there something like this?

PiergiorgioZagaria:
Helix sends the content of the file through stdin to the formatter. I think eslint has a --stdin flag that should work

kirawi:
> Probably we should mention it in the languages configuration section

You could add it here: https://docs.helix-editor.com/guides/adding_languages.html

samholmes:
The problem with `eslint` is that it wont return the file if it's able to fix. Instead, it returns a formatted response possibly containing the file output. For example, specifying the `json` format:

**/usr/local/bin/eslintfmt**
```
#!/bin/zsh

eslint --stdin --fix-dry-run --format=json | jq '.[0].output' <&0
```

This is close, except if the `output` in the JSON is not define because there was an error, then the file is emptied of contents. This script would need to parse the output and return the formatted output if there are no issues, otherwise return the `stdin`. I'm guessing this has already been written somewhere.

helix-view/src/document.rs
```diff
@@ -407,7 +409,53 @@ impl Document {
...
+                let output = process
+                    .wait_with_output()
+                    .await
+                    .map_err(|_| FormatterError::WaitForOutputFailed)?;
+
+                if !output.stderr.is_empty() {
+                    return Err(FormatterError::Stderr(
+                        String::from_utf8_lossy(&output.stderr).to_string(),
+                    ));
+                }
```
sudormrfbin, line 449: We should also check `output.status.success()` after checking stderr (so that the stderr message is always shown if it is printed) and return a `FormatterError::NonZeroExitStatus`. This way the eslint formatting script can simply return a non zero exit status and the file would be unchanged.

helix-view/src/document.rs
```diff
@@ -1034,6 +1087,39 @@ impl Default for Document {
...
+                format!("Failed to spawn formatter {}: {:?}", command, error)
+            }
+            Self::BrokenStdin => "Could not write to formatter stdin".to_string(),
+            Self::WaitForOutputFailed => "Waiting for formatter output failed".to_string(),
+            Self::Stderr(output) => format!("Formatter error: {}", output),
+            Self::InvalidUtf8Output => "Invalid UTF-8 formatter output".to_string(),
+            Self::DiskReloadError(error) => format!("Error reloading file from disk: {}", error),
+            Self::NonZeroExitStatus => "Formatter exited with non zero exit status:".to_string(),
+        };
+        f.write_str(&s)
```
archseer, line 1119: Rather than using `format!` and `to_string()` + `write_str`, just call `write!(f, "Could not write to formatter stdin")`, `write!(f, "Formatter error: {}", output)` etc. to avoid allocating

helix-view/src/document.rs
```diff
@@ -407,7 +409,57 @@ impl Document {
...
+                    .map_err(|e| FormatterError::SpawningFailed {
+                        command: formatter.command.clone(),
+                        error: e.kind(),
+                    })?;
+                {
+                    let mut stdin = process.stdin.take().ok_or(FormatterError::BrokenStdin)?;
+                    stdin
+                        .write_all(&text.bytes().collect::<Vec<u8>>())
+                        .await
+                        .map_err(|_| FormatterError::BrokenStdin)?;
```
archseer, line 437: `.bytes().collect()` is very inefficient since it reallocates a new temporary vector to hold the entire file (which could be very large).

Let's copy the approach from #3180 and use `helix_view::document::to_writer` which will stream the contents:

```rust
helix_view::document::to_writer(&mut stdin, encoding::UTF_8, &text).await?;
```

archseer:
Let's tackle the write_all problem in a separate PR, the fix there should be to just pick any view_id that already has a selection for that document. I had to do that somewhere else in commands.rs if I remember correctly

archseer (approved):
Great work on this!

PiergiorgioZagaria:
I think I forgot to mention it, but when we removed the file formatter in favour of just using stdio formatting, the problem with write_all was also solved. That is because with the file formatter we needed the view_id to reload the window, but now that we just use transactions the problem doesn't exist

David-Else:
Brilliant! Could anyone provide an example for prettier to format markdown files?

PiergiorgioZagaria:
I don't know the cli flags for prettier, assuming that the default behaviour is that it takes from stdin and prints to stdout, so for example you would call it like
`cat file.md | prettier`. Then you would just need to put 
```
[[language]]
name = "markdown"
formatter = { command = "prettier" }
```
In your languages.toml

David-Else:
Cheers! I tested it with `cat file.md |`  It should be:
```
[[language]]
name = "markdown"
formatter = { command = "prettier --parser markdown" }
```
Do I need any escape charcters for the command? In Neovim it was `--parser\ markdown`

PiergiorgioZagaria:
Then it's
`formatter = { command = "prettier", args = ["--parser", "markdown"] }`

kirawi:
>  Also, if there is an error does it abort? In Neovim the buffer just became the error message.

It'll return an error if it exits:
https://github.com/helix-editor/helix/blob/0ee20611022b5a7bec727d2159ec7c6b36e956b6/helix-view/src/document.rs#L449-L451

dam5h:
@PiergiorgioZagaria , thank you for this!  It's great to be able to use `black` as a python formater but `pyright` for LSP.  `pyright` doesn't support formatting.  

One question is related to the `auto-format` option in the editor config, as the default is `true`, shouldn't this happen automatically on saving a file, or was that circumvented when we gave this formatter priority over lsp?

helix-view/src/document.rs
```diff
@@ -407,7 +408,56 @@ impl Document {
...
+                }
+
+                let output = process
+                    .wait_with_output()
+                    .await
+                    .map_err(|_| FormatterError::WaitForOutputFailed)?;
+
+                if !output.stderr.is_empty() {
```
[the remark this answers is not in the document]
PiergiorgioZagaria: Ok, I'll open a new PR to fix it.
Btw the function `shell_impl` in `helix-term/src/commands.rs`, which is used when calling the shell with ! Or | doesn't implement these checks, should we add them there too?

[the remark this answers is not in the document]
dead10ck: I see, in that case, yeah, those need to be fixed too. The presence of output on stderr does not actually mean there was an error.

helix-view/src/document.rs
```diff
@@ -407,7 +408,56 @@ impl Document {
...
+                    return Err(FormatterError::Stderr(
+                        String::from_utf8_lossy(&output.stderr).to_string(),
+                    ));
+                }
+
+                if !output.status.success() {
+                    return Err(FormatterError::NonZeroExitStatus);
+                }
+
+                let str = String::from_utf8(output.stdout)
```
kirawi, line 453: You can use `from_reader` here instead. I think it is more efficient.

theherk:
> It's great to be able to use `black` as a python formater but `pyright` for LSP. `pyright` doesn't support formatting.

I case anybody stumbles in here searching for how to configure black, this is the configuration that does the trick.

```toml
﻿﻿﻿﻿﻿﻿﻿﻿[[language]]
name = "python"
auto-format = true
formatter = { command = "black", args = ["-", "-q"] }
```

helix-view/src/document.rs
PiergiorgioZagaria: `std::str::from_utf8` works, but `str::from_utf8` doesn't. Btw for further fixes to this part you can comment on #3349

PR #2947, opened 2022-07-02
Show clipboard info in --health output

sudormrfbin:
Screenshot_2022-07-03_01-50-07

### DIFF
--- a/helix-term/src/main.rs
+++ b/helix-term/src/main.rs
@@ -62,7 +62,8 @@ FLAGS:
     -h, --help                     Prints help information
     --tutor                        Loads the tutorial
     --health [CATEGORY]            Checks for potential errors in editor setup
-                                   CATEGORY can be a language or one of 'clipboard' or 'all'
+                                   CATEGORY can be a language or one of 'clipboard', 'languages'
+                                   or 'all'. 'all' is the default if not specified.
     -g, --grammar {{fetch|build}}    Fetches or builds tree-sitter grammars listed in languages.toml
     -c, --config <file>            Specifies a file to use for configuration
     -v                             Increases logging verbosity each use for up to 3 times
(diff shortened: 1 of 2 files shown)


helix-term/src/main.rs
```diff
@@ -61,9 +61,9 @@ ARGS:
 FLAGS:
     -h, --help                     Prints help information
     --tutor                        Loads the tutorial
-    --health [LANG]                Checks for potential errors in editor setup
-                                   If given, checks for config errors in language LANG
-    -g, --grammar {{fetch|build}}    Fetches or builds tree-sitter grammars listed in languages.toml
+    --health [CATEGORY]            Checks for potential errors in editor setup
+                                   CATEGORY can be a language or one of 'clipboard' or 'all'
+    -g, --grammar {{fetch|build}}  Fetches or builds tree-sitter grammars listed in languages.toml
```
the-mikedavis, line 66: The double space here was intentional actually:

hxhoffset

The double braces become single braces because it's interpreted as an escaped formatting argument so we end up needing those two extra spaces

helix-term/src/main.rs
```diff
@@ -61,9 +61,9 @@ ARGS:
 FLAGS:
     -h, --help                     Prints help information
     --tutor                        Loads the tutorial
-    --health [LANG]                Checks for potential errors in editor setup
-                                   If given, checks for config errors in language LANG
-    -g, --grammar {{fetch|build}}    Fetches or builds tree-sitter grammars listed in languages.toml
+    --health [CATEGORY]            Checks for potential errors in editor setup
+                                   CATEGORY can be a language or one of 'clipboard' or 'all'
```
the-mikedavis, line 65: We could add a line saying "when omitted, CATEGORY is 'all'" just to make it extra clear. What do you think?

AlexanderBrevig: Maybe it feels a bit weird to do `hx --health all` and not see the clipboard category?

the-mikedavis: Hmm actually yeah, I think "all" might be better as "languages." I would expect "all" to act the same as not passing a category

PR #2948, opened 2022-07-02
languages: added elvish

rsteube:
Added [elvish](https://elv.sh/).  Can't see the grammar taking any effect so there is likely still sth. wrong with the config.

### DIFF
--- a/runtime/queries/elvish/highlights.scm
+++ b/runtime/queries/elvish/highlights.scm
@@ -1,0 +1,76 @@
+;; SPDX-License-Identifier: 0BSD
+;; SPDX-FileCopyrightText: 2022 Tobias Frilling
+
+(comment) @comment
+
+(if "if" @keyword.control.conditional)
+(if (elif "elif" @keyword.control.conditional))
+(if (else "else" @keyword.control.conditional))
+
+(while "while" @keyword.control.repeat)
+(while (else "else" @keyword.control.repeat))
+(for "for" @keyword.control.repeat)
+(for (else "else" @keyword.control.repeat))
+
+(try "try" @keyword.control.exception)
+(try (catch "catch" @keyword.control.exception))
+(try (else "else" @keyword.control.exception))
+(try (finally "finally" @keyword.control.exception))
+
+(import "use" @keyword.control.import)
+(import (bareword) @string.special)
+
+(wildcard ["*" "**" "?"] @string.special)
+
+(command argument: (bareword) @variable.parameter)
+(command head: (identifier) @function)
+((command head: (identifier) @keyword.control.return)
+ (#eq? @keyword.control.return "return"))
+((command (identifier) @keyword.operator)
+ (#match? @keyword.operator "(and|or|coalesce)"))
+((command head: _ @function)
+ (#match? @function "([+]|[-]|[*]|[/]|[%]|[<]|[<][=]|[=][=]|[!][=]|[>]|[>][=]|[<][s]|[<][=][s]|[=][=][s]|[!][=][s]|[>][s]|[>][=][s])"))
+
+(pipeline "|" @operator)
+(redirection [">" "<" ">>" "<>"] @operator)
+
+(io_port) @constant.numeric
+
+(function_definition
+  "fn" @keyword.function
+  (identifier) @function)
+
+(parameter_list) @variable.parameter
+(parameter_list "|" @punctuation.bracket)
+
+(variable_declaration
+  "var" @keyword
+  (lhs (identifier) @variable))
+
+(variable_assignment
+  "set" @keyword
+  (lhs (identifier) @variable))
+
+(temporary_assignment
+  "tmp" @keyword
+  (lhs (identifier) @variable))
+
+(variable_deletion
+  "del" @keyword
+  (identifier) @variable)
+
+
+(number) @constant.numeric
+(string) @string
+
+((variable (identifier) @function)
+  (#match? @function ".+\\~$"))
+((variable (identifier) @constant.builtin.boolean)
+ (#match? @constant.builtin.boolean "(true|false)"))
+((variable (identifier) @constant.builtin)
+ (#match? @constant.builtin "(_|after-chdir|args|before-chdir|buildinfo|nil|notify-bg-job-success|num-bg-jobs|ok|paths|pid|pwd|value-out-indicator|version)"))
+(variable (identifier) @variable)
+
+["$" "@"] @punctuation.special
+["(" ")" "[" "]" "{" "}"] @punctuation.bracket
+";" @punctuation.delimiter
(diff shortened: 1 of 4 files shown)


the-mikedavis:
For the syntax highlighting you'll want to add `runtime/queries/elvish/highlights.scm` and `runtime/queries/elvish/injections.scm` following the queries in tree-sitter-elvish: https://github.com/ckafi/tree-sitter-elvish/tree/main/queries

The captures need to be adapted to the ones helix uses though: https://docs.helix-editor.com/master/themes.html#syntax-highlighting, for example `@parameter` should become `@variable.parameter`

And the `injections.scm` in tree-sitter-elvish looks like it's in neovim format. In helix format that injection would be

```scm
((comment) @injection.content
```

the-mikedavis:
Also to clear up the Docs CI's ❌, you'll want to run `cargo xtask docgen` and commit the changes

rsteube:
> For the syntax highlighting you'll want to add `runtime/queries/elvish/highlights.scm` and `runtime/queries/elvish/injections.scm` following the queries in tree-sitter-elvish: https://github.com/ckafi/tree-sitter-elvish/tree/main/queries
> 
> The captures need to be adapted to the ones helix uses though: https://docs.helix-editor.com/master/themes.html#syntax-highlighting, for example `@parameter` should become `@variable.parameter`
> 
> And the `injections.scm` in tree-sitter-elvish looks like it's in neovim format. In helix format that injection would be
> 
> ```scheme
> ((comment) @injection.content
>  (#set! injection.language "comment"))
> ```

So basically that probably needs to be forked to support neovim and helix (different format)?

the-mikedavis:
We host the queries in this repo (https://github.com/helix-editor/helix/tree/master/runtime/queries) since each editor tends to use their own queries. Technically the injections in that repo are wrong according to `tree-sitter-cli` and could be updated. But for getting syntax highlighting working for elvish in helix we just need the queries to be adapted in-repo here

rsteube:
Not sure if i got them all, but syntax highlighting is working now and i don't see an error.

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+(while (else "else" @keyword.control.repeat))
+(for "for" @keyword.control.repeat)
+(for (else "else" @keyword.control.repeat))
+
+(try "try" @keyword.control.exception)
+(try (catch "catch" @keyword.control.exception))
+(try (else "else" @keyword.control.exception))
+(try (finally "finally" @keyword.control.exception))
+
+(import "use" @include)
```
the-mikedavis, line 20:
```suggestion
(import "use" @keyword.control.import)
```

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+(import (bareword) @string.special)
+
+(wildcard ["*" "**" "?"] @string.special)
+
+(command argument: (bareword) @variable.parameter)
+(command head: (identifier) @function)
+((command head: (identifier) @keyword.return)
+ (#eq? @keyword.return "return"))
```
the-mikedavis, line 28:
```suggestion
((command head: (identifier) @keyword.return)
 (#eq? @keyword.return "return"))
```
```suggestion
((command head: (identifier) @keyword.control.return)
 (#eq? @keyword.control.return "return"))
```

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+((command head: _ @function)
+ (#any-of? @function
+  "+" "-" "*" "/" "%" "<" "<=""==" "!=" ">"
+  ">=" "<s" "<=s" "==s" "!=s" ">s" ">=s"
+))
+
+(pipeline "|" @operator)
+(redirection [">" "<" ">>" "<>"] @operator)
+
+(io_port) @number
```
the-mikedavis, line 40:
```suggestion
(io_port) @constant.numeric
```

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+(temporary_assignment
+  "tmp" @keyword
+  (lhs (identifier) @variable))
+
+(variable_deletion
+  "del" @keyword
+  (identifier) @variable)
+
+
+(number) @number
```
the-mikedavis, line 66:
```suggestion
(number) @constant.numeric
```

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+(string) @string
+
+(variable (identifier) @variable)
+((variable (identifier) @function)
+  (#match? @function ".+\\~$"))
+((variable (identifier) @boolean)
+ (#any-of? @boolean "true" "false"))
```
the-mikedavis, line 73:
```suggestion
((variable (identifier) @constant.builtin.boolean)
 (#match? @constant.builtin.boolean "(true|false)"))
```

We don't have `#any-of?` implemented at the moment so we use equivalent `#match?`s

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+((variable (identifier) @constant.builtin)
+ (#any-of? @constant.builtin
```
the-mikedavis, line 75: same here about `#any-of?`

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+(command head: (identifier) @function)
+((command head: (identifier) @keyword.return)
+ (#eq? @keyword.return "return"))
+((command (identifier) @keyword.operator)
+ (#any-of? @keyword.operator "and" "or" "coalesce"))
+((command head: _ @function)
+ (#any-of? @function
+  "+" "-" "*" "/" "%" "<" "<=""==" "!=" ">"
+  ">=" "<s" "<=s" "==s" "!=s" ">s" ">=s"
+))
```
the-mikedavis, line 35: and here with `#any-of?`

runtime/queries/elvish/highlights.scm
```diff
@@ -0,0 +1,82 @@
...
+
+(variable_deletion
+  "del" @keyword
+  (identifier) @variable)
+
+
+(number) @number
+(string) @string
+
+(variable (identifier) @variable)
```
the-mikedavis, line 69: There's a snag here where this stanza is higher precedence than the ones below it so the ones below will never match. If you move this down below the other `(variable (identifier) ...)` stanzas you'll see highlights for booleans for example

the-mikedavis: Oh actually I meant the `(variable (identifier) @variable)` stanza. The GitHub UI makes it look like I'm talking about the number/string stanzas 😅

frexsdev (approved):

the-mikedavis (commented):
Could you re-run `cargo xtask docgen`? It should add a checkmark to the highlight column for elvish. After that I think this should be good to go 👍

rsteube:
Sure. Will do when i'm back home.

the-mikedavis (approved):
Thanks for working on this, this looks great!

PR #2958, opened 2022-07-04
Add runtime `xcopy` command on powershell in docs

megumish:
The Windows command used to set up the runtime directory in the installation documentation is valid when run with cmd.exe, but when run with Powershell, I need to use another command.

My Powershell version

```
Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      19041  1682
```


the-mikedavis (approved):
Thanks!

PR #2961, opened 2022-07-04
Introduce storage highlighting for typescript/javascript

JakeHL:
Inspired by and following the patterns of #2731.
Adds `keyword.storage.type` and `keyword.storage.modifier` scopes for syntax highlighting.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -151,7 +151,9 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
-  - `storage` - Keywords that affect the storage of a variable, function or data structure `static`, `mut`, `const`, `ref`
+  - `storage` - Keywords that affect the storage of variables, functions, data structures etc. 
+    - `type` - Keywords for declaring the storage of variables, functions, data structures  etc. eg `let`, `trait`, `struct` in rust 
+    - `modifier` - Keywords that modify a declaration. eg `static`, `mut`, `const`, `ref` in rust
 
 - `operator` - `||`, `+=`, `>`
 
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -151,9 +151,9 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
-  - `storage` - Keywords that affect the storage of variables, functions, data structures etc. 
-    - `type` - Keywords for declaring the storage of variables, functions, data structures  etc. eg `let`, `trait`, `struct` in rust 
-    - `modifier` - Keywords that modify a declaration. eg `static`, `mut`, `const`, `ref` in rust
+  - `storage` - Keywords describing how things are stored
+    - `type` - The type of something, `class`, `function`, `var`, `let`, etc. 
+    - `modifier` - Storage modifiers like `static`, `mut`, `const`, `ref`, etc.
 
 - `operator` - `||`, `+=`, `>`
 
--- a/runtime/queries/javascript/highlights.scm
+++ b/runtime/queries/javascript/highlights.scm
@@ -163,19 +163,14 @@
 [
   "as"
   "async"
-  "class"
-  "const"
   "debugger"
   "delete"
-  "export"
   "extends"
   "from"
   "function"
   "get"
-  "import"
   "in"
   "instanceof"
-  "let"
   "new"
   "of"
   "set"
@@ -183,10 +178,16 @@
   "target"
   "try"
   "typeof"
-  "var"
   "void"
   "with"
 ] @keyword
+
+[
+  "class"
+  "let"
+  "const"
+  "var"
+] @keyword.storage.type
 
 [
   "switch"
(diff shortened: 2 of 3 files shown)


the-mikedavis:
Could you add some docs on the new keys? https://github.com/helix-editor/helix/blob/e58d28a9728fd5d451c98c48bd20c57fb9eec7dc/book/src/themes.md?plain=1#L154

JakeHL:
> Could you add some docs on the new keys?

All done. Feedback on the changes are welcome. I've slightly moved around the existing storage description as it was more specific to the modifier sub-scope.

I've added example keywords in rust as that was what was already there, however my personal opinion is that we shouldn't provide language specific examples. For a start I think those keywords are probably not implemented in the rust highlighting query.

I'm also not a rust dev so please let me know if I've misunderstood some of the keywords and they don't fall into the categories I've described.

runtime/queries/typescript/highlights.scm
```diff
@@ -22,15 +22,21 @@
...
+
+[
+  "public"
   "private"
   "protected"
-  "public"
-  "type"
   "readonly"
-] @keyword
+] @keyword.storage.modifier
```
the-mikedavis, line 42: These would probably be more appropriate as a new `keyword.storage.visibility` scope

runtime/queries/javascript/highlights.scm
```diff
@@ -163,31 +163,32 @@
...
   "void"
   "with"
 ] @keyword
 
+[
+  "class"
+  "let"
+  "const"
+  "var"
+] @keyword.storage.type
```
the-mikedavis, line 190: `const` should be `keyword.storage.modifier`

JakeHL: Not in javascript/typescript. It's used in place of let, var etc. It's a variable that cannot be reassigned.

```ts
/** can be re-assigned but not re-declared */
let someVar = 'hello world';
someVar = 'new string'; //OK!
let someVar = 'goodbye world' //ERROR!

/** can be re-assigned and re-declared */
var someVar = 'hello world;
someVar = 'new string'; //OK!
var someVar = 'goodbye world'; //OK!

/** cannot be re-assigned or re-declared */
const someVar = 'hello world';
someVar = 'new string'; //ERROR!
const someVar = 'goodbye world'; //ERROR!
```
There's a little more nuance to the differences between var and let but that should outline it.

There is another use for it in typescript when doing type assertions:

```ts
let someVar = someOtherVar as const
```

but I'm not familiar enough with the query syntax to approach that.

runtime/queries/typescript/highlights.scm
JakeHL: I'm not opposed to that for sure. I was just trying to follow the textmate grammar scopes as that's what's mentioned in the docs.

Are we happy to deviate from those?

book/src/themes.md
```diff
@@ -151,7 +151,9 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
-  - `storage` - Keywords that affect the storage of a variable, function or data structure `static`, `mut`, `const`, `ref`
+  - `storage` - Keywords that affect the storage of variables, functions, data structures etc. 
+    - `type` - Keywords for declaring the storage of variables, functions, data structures  etc. eg `let`, `trait`, `struct` in rust
```
the-mikedavis, line 155: Let's not duplicate the description here - the fallback behavior implies that `keyword.store.type` is a subset of `keyword.storage`. Textmate's description is good:

> the type of something, `class`, `function`, `int`, `var`, etc.

Examples are useful but we shouldn't use an example that isn't actually captured with the scope like the Rust ones listed here. Ones that are common between languages like the ones textmate lists are good, or the new ones from this change

book/src/themes.md
```diff
@@ -151,7 +151,9 @@ We use a similar set of scopes as
   - `operator` - `or`, `in`
   - `directive` - Preprocessor directives (`#if` in C) 
   - `function` - `fn`, `func`
-  - `storage` - Keywords that affect the storage of a variable, function or data structure `static`, `mut`, `const`, `ref`
+  - `storage` - Keywords that affect the storage of variables, functions, data structures etc. 
+    - `type` - Keywords for declaring the storage of variables, functions, data structures  etc. eg `let`, `trait`, `struct` in rust 
+    - `modifier` - Keywords that modify a declaration. eg `static`, `mut`, `const`, `ref` in rust
```
the-mikedavis, line 156: Textmate's examples are good here as well but I'd definitely keep `ref` and `mut` as examples

runtime/queries/typescript/highlights.scm
the-mikedavis: > Are we happy to deviate from those?

In general yes - we only use textmate (and sublime and vim) as a rough guide.

Textmate and sublime both have `public` as examples for `modifier` though so let's keep it under `keyword.storage.modifier` for now. We could introduce a `keyword.storage.modifier.visibility` later if it's needed.

the-mikedavis (commented):
`keyword.storage` was originally based on Vim's `StorageClass` highlight group (#2731). I think textmate's use of `storage.{type,modifier}` is a bit clearer though so let's lean towards those. For consistency though, Rust's `const` should become `keyword.storage.type`.

PR #2980, opened 2022-07-05
Sort the buffer picker by most recent access

estin:
Hi!

This is naive implementation of most-recent-used buffers picker.
As vim user in past this feature is very needs for me )

Sorry for my poor English.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2473,7 +2473,7 @@ fn buffer_picker(cx: &mut Context) {
         path: Option<PathBuf>,
         is_modified: bool,
         is_current: bool,
-        used_at: std::time::Instant,
+        focused_at: std::time::Instant,
     }
 
     impl ui::menu::Item for BufferMeta {
@@ -2506,7 +2506,7 @@ fn buffer_picker(cx: &mut Context) {
         path: doc.path().cloned(),
         is_modified: doc.is_modified(),
         is_current: doc.id() == current,
-        used_at: doc.focused_at,
+        focused_at: doc.focused_at,
     };
 
     let mut items = cx
@@ -2517,7 +2517,7 @@ fn buffer_picker(cx: &mut Context) {
         .collect::<Vec<BufferMeta>>();
 
     // mru
-    items.sort_by(|a, b| b.used_at.cmp(&a.used_at));
+    items.sort_unstable_by_key(|item| std::cmp::Reverse(item.focused_at));
 
     let picker = FilePicker::new(
         items,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2473,6 +2473,7 @@ fn buffer_picker(cx: &mut Context) {
         path: Option<PathBuf>,
         is_modified: bool,
         is_current: bool,
+        used_at: std::time::Instant,
     }
 
     impl ui::menu::Item for BufferMeta {
@@ -2505,14 +2506,21 @@ fn buffer_picker(cx: &mut Context) {
         path: doc.path().cloned(),
         is_modified: doc.is_modified(),
         is_current: doc.id() == current,
+        used_at: doc.used_at,
     };
 
+    let mut items = cx
+        .editor
+        .documents
+        .values()
+        .map(|doc| new_meta(doc))
+        .collect();
+
+    // mru
+    items.sort_by(|a, b| b.used_at.cmp(&a.used_at));
+
     let picker = FilePicker::new(
-        cx.editor
-            .documents
-            .values()
-            .map(|doc| new_meta(doc))
-            .collect(),
+        items,
         (),
         |cx, meta, action| {
             cx.editor.switch(meta.id, action);
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2506,7 +2506,7 @@ fn buffer_picker(cx: &mut Context) {
         path: doc.path().cloned(),
         is_modified: doc.is_modified(),
         is_current: doc.id() == current,
-        used_at: doc.used_at,
+        used_at: doc.focused_at,
     };
 
     let mut items = cx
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Should this be the default behavior of the buffer picker?

estin:
imho yes. this behavior must be default for buffer picker
may be some part of users prefer without any sorting.
It is must be decided by helix core team )

helix-view/src/document.rs
```diff
@@ -121,6 +121,9 @@ pub struct Document {
 
     diagnostics: Vec<Diagnostic>,
     language_server: Option<Arc<helix_lsp::Client>>,
+
+    // when document was used for most-recent-used buffer picker
+    pub used_at: std::time::SystemTime,
```
archseer, line 126: SystemTime is slower and not monotonic. https://doc.rust-lang.org/std/time/struct.Instant.html

helix-view/src/document.rs
```diff
@@ -662,6 +666,7 @@ impl Document {
         if self.selections.get(&view_id).is_none() {
             self.reset_selection(view_id);
         }
+        self.used_at = std::time::SystemTime::now();
```
archseer, line 669: This seems to only be tracked when the document is first created?

helix-view/src/document.rs
estin: done. `Instant` used instead of `SystemTime`

helix-view/src/document.rs
estin: no. works on each buffer switch. currently I use this feature right now )
And yes it was not clear. 
Now `doc.used_at` updated only on pick in buffer picker.

helix-term/src/keymap/default.rs
```diff
@@ -205,6 +205,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
             "f" => file_picker,
             "F" => file_picker_in_current_directory,
             "b" => buffer_picker,
+            "m" => mru_buffer_picker,
```
sudormrfbin, line 208: I agree with @the-mikedavis here, sorting the buffer picker by MRU sounds like good default behavior.

helix-term/src/commands.rs
```diff
@@ -2235,16 +2245,28 @@ fn buffer_picker(cx: &mut Context) {
...
-            .iter()
-            .map(|(_, doc)| new_meta(doc))
-            .collect(),
+        items,
         (),
         |cx, meta, action| {
+            // update used_at for mru sorting
+            if let Some(doc) = cx.editor.documents.get_mut(&meta.id) {
+                doc.used_at = std::time::Instant::now();
+            }
```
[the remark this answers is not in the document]
estin: done. set `doc.used_at` in `Editor::switch()`

estin:
@archseer hi!
Please check this PR again as you can.
Commits are squashed

helix-term/src/commands.rs
```diff
@@ -2216,6 +2216,7 @@ fn buffer_picker(cx: &mut Context) {
         path: Option<PathBuf>,
         is_modified: bool,
         is_current: bool,
+        used_at: std::time::Instant,
```
the-mikedavis, line 2219: Could we call this "focused_at" or "accessed_at"? "used_at" lacks some context about what in the document is used

helix-term/src/commands.rs
```diff
@@ -2253,14 +2254,21 @@ fn buffer_picker(cx: &mut Context) {
...
 
+    let mut items = cx
+        .editor
+        .documents
+        .iter()
+        .map(|(_, doc)| new_meta(doc))
+        .collect::<Vec<BufferMeta>>();
+
+    // mru
+    items.sort_by(|a, b| b.used_at.cmp(&a.used_at));
```
the-mikedavis, line 2268:
```suggestion
    // Sort so that most recently focused is on top
    items.sort_unstable_by_key(|item| std::cmp::Reverse(item.used_at));
```

In general I think sorting unstable is preferrable - it's unlikely that two items will have the same exact focused-at instant and even if they did, the ordering between them wouldn't be a big deal, right?

helix-view/src/document.rs
```diff
@@ -664,6 +668,11 @@ impl Document {
         }
     }
 
+    /// Mark document as recent used for MRU sorting
+    pub fn mark_as_used(&mut self) {
```
the-mikedavis, line 672: Should this also be called in `Editor::focus` and `Editor::focus_next` so that the focused-at time is updated when switching between splits/windows?

the-mikedavis (commented):
I've been testing this for a while and I like the behavior 👍

I left some minor comments

helix-term/src/commands.rs
estin: done

helix-term/src/commands.rs
estin: done

helix-view/src/document.rs
estin: I'm don't know how it must in common case. In my personal usage of vim each "focus on buffer" will mark that buffer as "used". 

Added set focused-at time on `Editor::focus` and `Editor::focus_next`

helix-view/src/editor.rs
```diff
@@ -1229,6 +1232,10 @@ impl Editor {
             self.mode = Mode::Normal;
             self.ensure_cursor_in_view(view_id);
         }
+
+        let view = self.tree.get_mut(view_id);
```
the-mikedavis, line 1236:
```suggestion
        let view = view!(self, view_id);
```

And for the one below

estin: done

pascalkuthe (dismissed):
The implementation LGTM. I like that you kept it simple 👍

the-mikedavis (approved):

pascalkuthe:
(rebased on master, there were some small conflicts which were not obvious from the github UI so it looked like a CI problem instead)

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #445, opened 2021-07-14
Allow specifying file start position

pickfire:
Like helix-term/src/commands.rs:3426:15

PR #2983, opened 2022-07-06
fix: restore start position functionality for the first file

mtoohey31:
This pull request fixes start position functionality (introduced in #445) for the first file argument.

This was broken by #2773 since the changes to `Application::new` now avoid reopening the first file (which is good), but it seems that also means the first file's position never gets set.


archseer:
Fixed in https://github.com/helix-editor/helix/pull/3073

PR #2984, opened 2022-07-06
Improve diagnostics picker formatting

sudormrfbin:
- Remove source (usually the name of the LSP) from diagnostic picker display
- Display diagnostic text before code in picker
- Sub sort diagnostics by line number
- Display error code only if not none
- Show file path only in workspace diagnostic picker


archseer (approved):
Looks nice 👍 we'll also align the filenames once Table support is present

PR #2988, opened 2022-07-06
Remove broken ledger tag highlight

pickfire:
> thread 'main' panicked at 'Could not parse queries for language "ledger". Are your grammars out of sync? Try running 'hx --grammar fetch' and 'hx --grammar build'. This query could not be parsed: QueryError { row: 13, column: 2, offset: 176, message: "tag", kind: NodeType }', helix-core/src/syntax.rs:379:43

I forgot that it was removed in tree-sitter-ledger even though the query is still there.


the-mikedavis (approved):
Ah whoops I forgot as well 😅

PR #2993, opened 2022-07-07
[book] Add toml theme template

d12bb:
To make it easier creating a new theme, I typed out all the scopes into a toml template file, roughly in order of the descriptions above.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -91,7 +91,7 @@ Less common modifiers might not be supported by your terminal emulator.
 
 ### Scopes
 
-The following is a list of scopes available to use for styling.
+The following is a list of scopes available to use for styling. It's also available as toml template [below](#template), so you don't have to type in all the scopes yourself.
 
 #### Syntax highlighting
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
If anything I think we should suggest starting with an existing theme in the documentation. Themes may not want to define all possible scopes and existing themes tend to have a palette set up which is easier to get started with than the full set of scopes with no palette. Plus the syntax here kind-of implies that you can have a scope key without a right-hand-side style which is invalid TOML.

d12bb:
Fair enough. I basically made this for myself to visualize what I already did set and what I'm still missing when translating a Neovim theme for Helix. So adding this here was just the afterthought that maybe it will be helpful to others, but you're probably right that there are better ways to start a theme.

the-mikedavis:
We try to keep this info up to date here as well: https://github.com/helix-editor/helix/blob/230ba264bf78d9b4ecd42440f0cbb20529f9c235/book/src/themes.md#syntax-highlighting (and those docs could use fleshing out - some keys could use docs and examples like in the textmates docs https://macromates.com/manual/en/language_grammars#naming_conventions)

I think this could make a good gist. Being in the repo though would just be extra upkeep work when adding new scopes so I'll close this out

PR #2996, opened 2022-07-07
Make gutters padding optional

pickfire:
If all gutters are removed, there are still an extra one padding, would be nice
to remove that to save some space.

[asciicast](https://asciinema.org/a/aSUgtapPdQoZs2vZC14V1nGt8)

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -38,7 +38,7 @@ hidden = false
 | `shell` | Shell to use when running external commands. | Unix: `["sh", "-c"]`<br/>Windows: `["cmd", "/C"]` |
 | `line-number` | Line number display: `absolute` simply shows each line's number, while `relative` shows the distance from the current line. When unfocused or in insert mode, `relative` will still show absolute line numbers. | `absolute` |
 | `cursorline` | Highlight all lines with a cursor. | `false` |
-| `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers`, note that `diagnostics` also includes other features like breakpoints | `["diagnostics", "line-numbers"]` |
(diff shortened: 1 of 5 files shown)


helix-view/src/view.rs
```diff
@@ -91,12 +95,23 @@ impl fmt::Debug for View {
...
+                GutterType::LineNumbers => 5,
+                GutterType::Padding => 1,
+            };
+            gutter_offset += width;
+            gutters.push((
+                match gutter_type {
+                    GutterType::Diagnostics => gutter::diagnostics_or_breakpoints,
+                    GutterType::LineNumbers => gutter::line_numbers,
+                    GutterType::Padding => gutter::padding,
+                },
```
pickfire, line 112: Not sure why but I can't seemed to keep these as a variable `gutter`, maybe function name is different?

book/src/configuration.md
```diff
@@ -38,7 +38,7 @@ hidden = false
 | `shell` | Shell to use when running external commands. | Unix: `["sh", "-c"]`<br/>Windows: `["cmd", "/C"]` |
 | `line-number` | Line number display: `absolute` simply shows each line's number, while `relative` shows the distance from the current line. When unfocused or in insert mode, `relative` will still show absolute line numbers. | `absolute` |
 | `cursorline` | Highlight all lines with a cursor. | `false` |
-| `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers`, note that `diagnostics` also includes other features like breakpoints | `["diagnostics", "line-numbers"]` |
+| `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `padding`, note that `diagnostics` also includes other features like breakpoints | `["diagnostics", "line-numbers", "padding"]` |
```
pickfire, line 41: `:set gutters diagnostics` or `:set gutters ["diagnostics"]` doesn't seemed to work for me, the completion for it is also weird.

archseer (approved):

PR #2998, opened 2022-07-07
Fix Document::language_id returning None if there is no lang server

Mathspy:
As mentioned in https://github.com/helix-editor/helix/pull/2434#discussion_r915483996 this was always returning `None` if no language server was configured because of early returning with `?` operator.
Also switched to relying on `config.language_id` instead of the last part of the scope because not all languages have `scope` configured and sometimes scope is configured as `"source.md"` so we would end up with `"md"` but returning `"markdown"` would make more sense, at least for the purposes of #2434

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -99,7 +99,7 @@ pub struct Document {
     pub line_ending: LineEnding,
 
     syntax: Option<Syntax>,
-    /// Corresponding language scope name. Usually `source.<lang>`.
+    /// Corresponding language configuration as deserialized from languages.toml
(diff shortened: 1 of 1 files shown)


archseer:
> at least for the purposes of #2434

This is used primarily to signal the document type to the LSP, I'm worried that such a change would completely break some languages.

AlexanderBrevig:
Aslo fixes an issue with #3273 where `markdown` (and more) would display as `text`

archseer:
Merged https://github.com/helix-editor/helix/pull/3338

PR #2999, opened 2022-07-07
nix flake: make the binary cache "just work"

nrdxp:
For those not familiar, Nix will ask on first use if you would like to adopt these settings from the flake, if accepted then users will pull artifacts from the official cache when available instead of needlessly building them whenever running a Nix flake command such as `nix build`.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -22,8 +22,12 @@ the project root. The flake can also be used to spin up a reproducible developme
 shell for working on Helix with `nix develop`.
 
 Flake outputs are cached for each push to master using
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Oh sweet this is handy. With this change we can probably clean up this sentence in the installation documentation: https://github.com/helix-editor/helix/blob/230ba264bf78d9b4ecd42440f0cbb20529f9c235/book/src/install.md?plain=1#L25-L27

nrdxp:
good point, I pushed a commit with an updated doc

the-mikedavis (approved):
Looks good, thanks! ❄️

PR #3003, opened 2022-07-07
nix: pass makeWrapperArgs to wrapProgram

nrdxp:
Even though in the default package this list is empty, it is still useful to apply extra wrapper args downstream. 

For example, one may wish to wrap Helix with the `PATH` set to include various language servers by default if not installed with the `--suffix` arg.

Full list of possible args:
https://github.com/NixOS/nixpkgs/blob/1c70b694fe0fece5ae74beb747a40f1b7237d251/pkgs/build-support/setup-hooks/make-wrapper.sh#L14-L35

One can see a real world example in this commit:
https://github.com/nrdxp/nrdos/commit/d8bf68e639343566d33545f189616bba57e0ed3f

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -63,7 +63,7 @@
 
             postFixup = ''
               if [ -f "$out/bin/hx" ]; then
-                wrapProgram "$out/bin/hx" --set HELIX_RUNTIME "${runtimeDir}"
+                wrapProgram "$out/bin/hx" ''${makeWrapperArgs[@]} --set HELIX_RUNTIME "${runtimeDir}"
               fi
             '';
           };


nrdxp:
looks like I can't request reviews so cc @yusdacra in case there is a better way to do this

yusdacra:
Looks fine to me. One can just use `overrideAttrs` to override the package downstream so this will work fine if you pass `makeWrapperArgs`.

nrdxp:
Originally I wanted to add an option to `override` so one could use `helix.override { makeWrapperArgs = []; }` but I realized there is no `override` function on Helix in the first place, so I had to go with the more verbose `overrideAttrs` route.

yusdacra:
> Originally I wanted to add an option to `override` so one could use `helix.override { makeWrapperArgs = []; }` but I realized there is no `override` function on Helix in the first place, so I had to go with the more verbose `overrideAttrs` route.

Yeah, the Rust builders on `dream2nix` do not expose an `override`, which is my fault I believe :sweat_smile: I didn't really expect that anyone would want to use `override` since there is really nothing special being passed to the derivation for any "configuration" of sorts. Mostly I expected people would just write their own wrappers around the packages since modifying the package in this way will trigger rebuilds, which is not something that would be wanted I think...

the-mikedavis (approved):

PR #3008, opened 2022-07-08
use % as file name register

### LINKED ISSUE #2985: Add a command to copy the filename and cursor line number in the clipboard.

QiBaobin:
fix #2985


QiBaobin:
close per #3393

PR #3015, opened 2022-07-08
Support different kinds of underline rendering

sudormrfbin:
Adds four new  modifiers that can be used in themes:

- `undercurled`
- `underdashed`
- `underdotted`
- `double-underline`

Screenshot_2022-07-09_03-42-49

### TODO

- [x] Specify color for the underline (currently automatically inherits the style of the cell).

### DIFF
--- a/helix-tui/src/backend/crossterm.rs
+++ b/helix-tui/src/backend/crossterm.rs
@@ -153,7 +153,7 @@ impl ModifierDiff {
         if removed.contains(Modifier::ITALIC) {
             map_error(queue!(w, SetAttribute(CAttribute::NoItalic)))?;
         }
-        if removed.contains(Modifier::UNDERLINED) {
+        if removed.intersects(Modifier::ANY_UNDERLINE) {
             map_error(queue!(w, SetAttribute(CAttribute::NoUnderline)))?;
         }
         if removed.contains(Modifier::DIM) {
@@ -178,6 +178,18 @@ impl ModifierDiff {
         }
         if added.contains(Modifier::UNDERLINED) {
             map_error(queue!(w, SetAttribute(CAttribute::Underlined)))?;
+        }
+        if added.contains(Modifier::UNDERCURLED) {
+            map_error(queue!(w, SetAttribute(CAttribute::Undercurled)))?;
+        }
+        if added.contains(Modifier::UNDERDOTTED) {
+            map_error(queue!(w, SetAttribute(CAttribute::Underdotted)))?;
+        }
+        if added.contains(Modifier::UNDERDASHED) {
+            map_error(queue!(w, SetAttribute(CAttribute::Underdashed)))?;
+        }
+        if added.contains(Modifier::DOUBLE_UNDERLINED) {
+            map_error(queue!(w, SetAttribute(CAttribute::DoubleUnderlined)))?;
         }
         if added.contains(Modifier::DIM) {
             map_error(queue!(w, SetAttribute(CAttribute::Dim)))?;
--- a/helix-tui/src/backend/crossterm.rs
+++ b/helix-tui/src/backend/crossterm.rs
@@ -11,8 +11,38 @@ use crossterm::{
 use helix_view::graphics::{Color, CursorKind, Modifier, Rect};
 use std::io::{self, Write};
 
+fn vte_version() -> Option<usize> {
+    std::env::var("VTE_VERSION").ok()?.parse().ok()
+}
+
+/// Describes terminal capabilities like extended underline, truecolor, etc.
+#[derive(Copy, Clone, Debug, Default)]
+struct Capabilities {
+    /// Support for undercurled, underdashed, etc.
+    has_extended_underlines: bool,
+}
+
+impl Capabilities {
+    /// Detect capabilities from the terminfo database located based
+    /// on the $TERM environment variable. If detection fails, returns
+    /// a default value where no capability is supported.
+    pub fn from_env_or_default() -> Self {
+        match cxterminfo::terminfo::TermInfo::from_env() {
+            Err(_) => Capabilities::default(),
+            Ok(t) => Capabilities {
+                // Smulx, VTE: https://unix.stackexchange.com/a/696253/246284
+                // Su (used by kitty): https://sw.kovidgoyal.net/kitty/underlines
+                has_extended_underlines: t.get_ext_string("Smulx").is_some()
+                    || *t.get_ext_bool("Su").unwrap_or(&false)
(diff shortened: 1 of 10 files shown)


runtime/themes/onedark.toml
```diff
@@ -40,7 +40,7 @@
 "diff.delta" = "gold"
 "diff.minus" = "red"
 
-diagnostic = { modifiers = ["underlined"] }
+diagnostic = { modifiers = ["undercurled"] }
```
archseer, line 43: What happens if the user's terminal doesn't support undercurl?

archseer: We should look up `Smulx` in terminfo then fallback all of these to underline if not present: https://github.com/wez/wezterm/issues/415#issuecomment-755442430=

sudormrfbin: Crossterm [doesn't support querying terminfo](https://github.com/crossterm-rs/crossterm/issues/495), so we would have to pull in something like the [`terminfo`](https://docs.rs/terminfo/latest/terminfo/index.html) crate to do this. Unless there's a better (no-dependency) way to this?

archseer: I was thinking maybe the `term` crate since it has less dependencies (no nom or phf). We could also inline a crate, it would be useful for detecting other features more reliably (true color)

sudormrfbin: The `term` crate cannot parse extended capabilities like `Smulx`, so we would have to contribute it upstream first. Another option is to use the [`cxterminfo`][1] crate which has support for [extended capabilities][2] and is quite well documented too. Another advantage of `cxterminfo` is that it's only about 500 lines long and has no dependencies so it would be easy to inline into helix.

[1]: https://github.com/BxNiom/cx-terminfo
[2]: https://github.com/BxNiom/cx-terminfo/blob/9fee0c06544dea9c5b5b25224132b51fdf3df548/src/terminfo.rs#L293-L297
[3]: https://github.com/BxNiom/cx-terminfo/blob/9fee0c06544dea9c5b5b25224132b51fdf3df548/src/terminfo.rs

archseer: That seems like a good option 👍🏻

A-Walrus:
Wrote a PR to add color to the underlines: [PR](https://github.com/sudormrfbin/helix/pull/3)

archseer:
> Wrote a PR to add color to the underlines: [PR](https://github.com/sudormrfbin/helix/pull/3)

Looks great! I wanted support for that too :)

sudormrfbin:
Added undercurl support detection based on terminfo. Thanks to @A-Walrus for the color customization code :)

sudormrfbin:
This only seems to affect them when the underlines are different, for example when both the error and cursor modifiers are `undercurl`, there is no glitch. Also the glitching behavior is also slightly different for me; putting the cursor in the middle of the error underlined word cuts off the rest of the word underline rather than removing the underline only on pressing the left key:

helix-underline-bug

The modifiers used:

```toml
"diagnostic.error" = { underline = "red", modifiers = ["underlined"] }
"ui.cursor.primary" = { fg = "white", modifiers = ["reversed", "undercurled"] }
```

archseer:
There seems to be no reset call for `SetUnderlineColor` in this block: https://github.com/helix-editor/helix/blob/da52d20a69dad2e72131055a5b99a5260f2ae848/helix-tui/src/backend/crossterm.rs#L119-L124

helix-tui/src/backend/crossterm.rs
```diff
@@ -153,7 +192,7 @@ impl ModifierDiff {
         if removed.contains(Modifier::ITALIC) {
             map_error(queue!(w, SetAttribute(CAttribute::NoItalic)))?;
         }
-        if removed.contains(Modifier::UNDERLINED) {
+        if removed.intersects(Modifier::ANY_UNDERLINE) {
```
archseer, line 195: Could it be that we have to emit `NoUnderline` followed by underline if the cell changes from one underline to a different type in the next one?

archseer:
I tried this on wezterm: the underline color changes, but it's using the same underline (not undercurl). I did confirm that undercurl works normally via

```
printf "\x1b[58:2::255:0:0m\x1b[4:1msingle\x1b[4:2mdouble\x1b[4:3mcurly\x1b[4:4mdotted\x1b[4:5mdashed\x1b[0m\n"
```

sudormrfbin:
> `has_extended_underlines` is false for me. Manually overriding it to always be true works and I don't have the same rendering issue as above

Seems like wezterm's terminfo file has to [installed separately][1]:

```bash
curl https://raw.githubusercontent.com/wez/wezterm/master/termwiz/data/wezterm.terminfo | tic -x -
```

There's also one more problem at play here even if you have the terminfo file but when it's installed in `~/.terminfo/`: `tic` attempts to install the terminfo entry in `/usr/share/terminfo` first and if denied permission installs it in `~/.terminfo`, so by default the above command will install it in your HOME. But there seems to be a bug with the [path added in `cxterminfo`][2], adding `~/` instead of `~/.terminfo`. It's a separate bug but overlaps with the wezterm issue.

So to check if wezterm works with undercurl:

```bash
curl https://raw.githubusercontent.com/wez/wezterm/master/termwiz/data/wezterm.terminfo | tic -x -
export TERM=wezterm
export TERMINFO=~/.terminfo
hx # or cargo run, etc
```

Fixing the cxterminfo bug should remove the `export TERMINFO` step, but the terminfo file will still have to be installed for wezterm (unless we also add a version check similar to checking `$VTE_VERSION`).

[1]: https://github.com/wez/wezterm/issues/415#issuecomment-755849623
[2]: https://docs.rs/cxterminfo/0.2.0/src/cxterminfo/terminfo.rs.html#212-215

EpocSquadron:
Tested this out and noticed sometimes the way underlines work in wezterm it's going to be clipped (in other words, there's not enough room for the underlines so it looks funky). I brought it up in the wezterm matrix and there may be upcoming fixes to make it look better in more cases, but I think we should include up front in the documentation that font adjustments may be needed with any given terminal emulator for the fancy underlines to look good. For example, wezterm provides a way to increase the line height and adjust the positioning of the underline to be closer to the text. Any combination of that may be enough to resolve the issue.

See matrix discussion in wezterm room for more information: https://matrix.to/#/!PirwUBcuIlTXwNveYz:matrix.org/$muIHcpgbicXxPXLFuxvfXEMmPg-oqmGu0drVRc_UnA4?via=matrix.org&via=kde.org&via=mozilla.org

sudormrfbin:
Closing in favor of #4061.

PR #3021, opened 2022-07-09
support toggling pickers' preview panel

QiBaobin:
Add a key to toggle preview panel so that we can get more info when searching files with long names.


archseer:
Nice, can you also document this in the docs?

QiBaobin:
> Nice, can you also document this in the docs?

Added now, thanks for reviewing.

the-mikedavis (approved):
This is pretty handy for the new diagnostics picker too

Thanks for working on this!

PR #3023, opened 2022-07-09
feat: document abbreviations

### LINKED ISSUE #1471: Abbreviation

Add support for vim like `abbr` where as you type you can get the abbreviation. Also, it needs a mechanism to cancel out the abbreviation. For example, `:abbr ctx context`, then when you type `ctx` it will be changed to `context`.

lazytanuki:
Peek 2022-07-10 00-26

This feature introduces four new typed commands to handle vim-like abbreviations:

- `insert-abbreviation` (alias `abbr`): adds a new abbreviation, or overwrites an existing one (with completion, local to current document only)
- `delete-abbreviation` (alias `rabbr`): removes an abbreviation (with completion)
- `load-abbreviations-from-file `(alias `labbr`): loads abbreviations from a file. Each line is an abbreviation, for instance:

    ```    
    nvm nevermind
    brb be right back
    ```
    
- `write-abbreviations-to-file` (alias `wabbr`): writes to file with the same format as above

I have tested and solved edge cases (beginning of the file, of a line, etc), but would be happy to have someone test this.

Closes #1471.

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -73,3 +73,7 @@
 | `:pipe` | Pipe each selection to the shell command. |
 | `:pipe-to` | Pipe each selection to the shell command, ignoring output. |
 | `:run-shell-command`, `:sh` | Run a shell command |
+| `:insert-abbreviation`, `:abbr` | Insert a new abbreviation |
+| `:delete-abbreviation`, `:rabbr` | Delete an abbreviation |
+| `:load-abbreviation-from-file`, `:labbr` | Load abbreviations from a file that contains abbreviations such as:<br><br>nvm nevermind<br>brb be right back<br>... |
+| `:write-abbreviation-to-file`, `:wabbr` | Write abbreviations to a file using the following format:<br><br>nvm nevermind<br>brb be right back<br>... |
--- a/helix-core/src/abbreviations.rs
+++ b/helix-core/src/abbreviations.rs
@@ -3,9 +3,10 @@ use std::collections::HashMap;
 use ropey::Rope;
 
 use crate::{movement, Change, Range, Selection, Tendril, Transaction};
+use serde::{Deserialize, Serialize};
 
 /// The type that represents the collection of abbreviations,
-#[derive(Debug, Clone)]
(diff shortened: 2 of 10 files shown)


archseer:
Rather than special files, can we make this static and a part of the current toml config? In the future this can be more scriptable then

lazytanuki:
> Rather than special files, can we make this static and a part of the current toml config? In the future this can be more scriptable then

I think we could have both ! I could add editor-wide abbreviations sourced from the configuration file, but I'd like to keep the ability to load from file because sometimes you have abbreviations specific to one document but don't want them all the time, you just want to reload them until you finish the document and then get rid of them.

EpocSquadron:
How does this relate to #2852 ? They seem to be the same thing with sightly different angles of attack to me. Also wondering how both intersect with the idea of snippets.

lazytanuki:
> How does this relate to #2852 ? They seem to be the same thing with sightly different angles of attack to me. Also wondering how both intersect with the idea of snippets.

This was not made with #2852 in mind at all, but I guess people could use abbreviations to be able to insert digraphs. I'm French, so I wouldn't be able to help a lot with that as I don't really know how one uses digraphs.

Snippets are different, they're closer to the LSP aspect of the editor and they are not something one can define "on-the-fly", unlike abbreviations.

lazytanuki:
I just rebased this on the current master. Been using it since July without any issue, would someone care to review ?
Thanks !

lazytanuki:
Rebased on master.

lazytanuki:
Rebased and resolved conflicts. Any hope for a review ?
Thanks !

lazytanuki:
Rebased on master.

lazytanuki:
Following a discussion on Matrix with @pascalkuthe, we came to the conclusion that abbreviations should be implemented as special snippets that trigger instantaneously. It would also require to implement a command to define them on-the-fly.
We'll have to wait for the snippet engine (for anyone wondering about it, there's a snippet branch where it is currently being developed).
For the impatient: feel free to merge this PR into you own custom forks until we have this in the snippet engine.

Closing.

Issue #2981, opened 2022-07-06
panicked after adding cursor on EOL


TyswSh:
### Summary

Create a blank page, remove the newline, and leave only the EOL.
Then adding a new cursor causes an error.
Not an error in Helix, but I opened it to check.

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 18446744073709551615, Rope/RopeSlice char length 0'

### Reproduction Steps

 

I tried this:

1. `hx`
make a blank page
2. `d`
Remove newline using `d`
3. `C(Shift-c)` or `Alt-C(Alt-Shift-C)`
add new cursor then crashes

### Helix log

helix_loader [DEBUG] Located configuration folders: []

### Platform

windows10

### Terminal Emulator

windows terminal

### Helix Version

helix 22.05 (c88d736d)

the-mikedavis:
Can you provide a full backtrace by setting the `RUST_BACKTRACE` environment variable to `full`? I can't reproduce this on linux even when the line-endings are set to crlf

TyswSh:
thread 'main' panicked at 'attempt to subtract with overflow', helix-term\src\commands.rs:1417:26
stack backtrace:
   0:     0x7ff7ad07104f - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff7ad07104f - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff7ad07104f - std::sys_common::backtrace::_print_fmt
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\sys_common\backtrace.rs:66
   3:     0x7ff7ad07104f - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\sys_common\backtrace.rs:45
   4:     0x7ff7ad0929fa - core::fmt::write
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\fmt\mod.rs:1194
[...]

PR #3024, opened 2022-07-10
fix: error that caused usize to overflow

GreasySlug:
fix this bug #2981

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1414,7 +1414,7 @@ fn copy_selection_on_line(cx: &mut Context, direction: Direction) {
         let (head, anchor) = if range.anchor < range.head {
             (range.head - 1, range.anchor)
         } else {
-            (range.head, range.anchor - 1)
+            (range.head, range.anchor.checked_sub(1).unwrap_or_default())
         };
 
         let tab_width = doc.tab_width();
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -1414,7 +1414,7 @@ fn copy_selection_on_line(cx: &mut Context, direction: Direction) {
         let (head, anchor) = if range.anchor < range.head {
             (range.head - 1, range.anchor)
         } else {
-            (range.head, range.anchor - 1)
+            (range.head, range.anchor.checked_sub(1).unwrap_or_default())
```
the-mikedavis, line 1417: You can use `.saturating_sub(1)` here for the same effect

the-mikedavis (approved):

GreasySlug:
Thank you reviewing😊

archseer (approved):

helix-term/src/commands.rs
```diff
@@ -1414,7 +1414,7 @@ fn copy_selection_on_line(cx: &mut Context, direction: Direction) {
         let (head, anchor) = if range.anchor < range.head {
             (range.head - 1, range.anchor)
```
EpocSquadron, line 1415: Do we need this here too?

the-mikedavis: I don't think it's possible for `range.head` to be `0` in this branch because of the if condition: `range.anchor < range.head` implies `range.head` is `>= 1` unless it's a zero-width selection which should not be possible

PR #3025, opened 2022-07-10
add fortran language

mcbunkus:
This PR adds syntax highlighting and LSP support for Fortran. It doesn't include textobjects but `fortls` seems to work without any issues.

Let me know if I made any mistakes, or if there's anything I need to change. 😄

### DIFF
--- a/runtime/queries/fortran/highlights.scm
+++ b/runtime/queries/fortran/highlights.scm
@@ -1,9 +1,3 @@
-(identifier) @variable
-(string_literal) @string
-(number_literal) @number
-(boolean_literal) @boolean
-(comment) @comment
-
 [
   (intrinsic_type)
   "dimension"
@@ -82,14 +76,14 @@
   "endif"
   "where"
   "endwhere"
-] @conditional
+] @keyword.control.conditional
 
 [
   "do"
   "enddo"
   "while"
   "forall"
-] @repeat
+] @keyword.control.repeat
 
 [
   "*"
@@ -167,4 +161,11 @@
   name: (identifier) @keyword)
 
 (derived_type_member_expression
-  (type_member) @property)
+  (type_member) @variable.other.member)
+
+(identifier) @variable
+(string_literal) @string
+(number_literal) @constant.numeric
+(boolean_literal) @constant.builtin.boolean
+(comment) @comment
+
--- a/runtime/queries/fortran/highlights.scm
+++ b/runtime/queries/fortran/highlights.scm
@@ -1,0 +1,170 @@
+(identifier) @variable
+(string_literal) @string
+(number_literal) @number
+(boolean_literal) @boolean
+(comment) @comment
+
+[
+  (intrinsic_type)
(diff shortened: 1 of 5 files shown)


runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
+(identifier) @variable
+(string_literal) @string
+(number_literal) @number
```
the-mikedavis, line 3:
```suggestion
(number_literal) @constant.numeric
```

runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
...
+(boolean_literal) @boolean
```
the-mikedavis, line 4:
```suggestion
(boolean_literal) @constant.builtin.boolean
```

runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
+(identifier) @variable
```
the-mikedavis, line 1: It looks like these queries were written for nvim-treesitter. nvim-treesitter has reversed precedence so this has the lowest priority for nvim but everything else that uses tree-sitter uses the reverse: this is the highest precedence stanza in helix, so the `@variable.parameter` capture on L136 can't match. To fix it you can move this line down below any other line that captures the `(identifier)` node.

runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
...
+
+[
+  "if" 
+  "then"
+  "else"
+  "elseif"
+  "endif"
+  "where"
+  "endwhere"
+] @conditional
```
the-mikedavis, line 85:
```suggestion
] @keyword.control.conditional
```

runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
...
+
+[
+  "do"
+  "enddo"
+  "while"
+  "forall"
+] @repeat
```
the-mikedavis, line 92:
```suggestion
] @keyword.control.repeat
```

runtime/queries/fortran/highlights.scm
```diff
@@ -0,0 +1,170 @@
...
+  (name) @function)
+
+(subroutine_call
+	(name) @function)
+
+(keyword_argument
+  name: (identifier) @keyword)
+
+(derived_type_member_expression
+  (type_member) @property)
```
the-mikedavis, line 170:
```suggestion
  (type_member) @variable.other.member)
```

mcbunkus:
I've fixed the queries, thanks for the help. I'm a complete noob to treesitter.

the-mikedavis (approved):
Looks great, thanks!

David-Else:
@mcbunkus Could you add install instructions at https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers ?

PR #3027, opened 2022-07-10
Fix locals tracking in Rust closures

the-mikedavis:
The fix comes from the rewriting of the `closure_parameters` stanza:
it was capturing the entire `closure_parameters` node including
`|`s, whitespace and commas. Capturing the identifiers within
fixes the tracking.

In order to make sure locals definitions from closure parameters don't
leak out of the body of the closure, though, we should also mark the
closure itself as a locals scope.

| Before | After |
|---|---|
| before | after |

Where `variable.parameter` is underlined white in my theme (notice `layer` and `r` above).

### DIFF
--- a/runtime/queries/rust/locals.scm
+++ b/runtime/queries/rust/locals.scm
@@ -1,6 +1,7 @@
 ; Scopes
 
 (block) @local.scope
+(closure_expression) @local.scope
 
 ; Definitions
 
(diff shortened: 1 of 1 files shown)


pickfire (approved):
I don't quite know what you did there but it is interesting. Can merge if you want I guess.

PR #3030, opened 2022-07-11
Preventing `usize` overflow

sabify:


the-mikedavis:
Are these related to panics you've seen? For a few of these branches I don't think overflow is possible

sabify:
No, they are not related to panics, and you are right, some of them were non-sense and were removed.

helix-term/src/commands.rs
```diff
@@ -781,7 +781,7 @@ fn trim_selections(cx: &mut Context) {
         let idx = ranges
             .iter()
             .position(|range| range.overlaps(&primary))
-            .unwrap_or(ranges.len() - 1);
+            .unwrap_or_else(|| ranges.len() - 1);
```
sbromberger, line 784: len() should be an O(1) lookup - is there an advantage to using a closure here?

sabify: It was part of the first commit to resolve `cargo clippy` warning related to lazily evaluation of `saturating_sub` method. I will revert it.

helix-term/src/commands.rs
```diff
@@ -1138,7 +1138,7 @@ fn find_char_impl<F, M: CharMatcher + Clone + Copy>(
         // graphemes, whereas this function doesn't yet.  So we're doing the same logic
         // here, but just in terms of chars instead.
         let search_start_pos = if range.anchor < range.head {
-            range.head - 1
+            range.head.saturating_sub(1)
```
archseer, line 1141: If anchor is less than head then this subtraction is guaranteed not to fail.

helix-term/src/commands.rs
```diff
@@ -1412,7 +1412,7 @@ fn copy_selection_on_line(cx: &mut Context, direction: Direction) {
 
         // The range is always head exclusive
         let (head, anchor) = if range.anchor < range.head {
-            (range.head - 1, range.anchor)
+            (range.head.saturating_sub(1), range.anchor)
```
archseer, line 1415: ditto here, this is not necessary because `anchor < head`

helix-term/src/commands.rs
```diff
@@ -1844,7 +1844,10 @@ fn global_search(cx: &mut Context) {
                                 entry.path(),
                                 sinks::UTF8(|line_num, _| {
                                     all_matches_sx
-                                        .send(FileResult::new(entry.path(), line_num as usize - 1))
+                                        .send(FileResult::new(
+                                            entry.path(),
+                                            (line_num as usize).saturating_sub(1),
```
archseer, line 1849: I think line_num is guaranteed to be non-zero here too

helix-term/src/commands.rs
```diff
@@ -1327,7 +1327,7 @@ pub fn scroll(cx: &mut Context, offset: usize, direction: Direction) {
     let text = doc.text().slice(..);
 
     let cursor = coords_at_pos(text, range.cursor(text));
-    let doc_last_line = doc.text().len_lines().saturating_sub(1);
+    let doc_last_line = doc.text().len_lines() - 1;
```
archseer, line 1330: On the other hand this might actually crash now since `len_lines()` can return 0

archseer (changes requested):
I think none of these changes are necessary

helix-term/src/commands.rs
sabify: `len_lines()` returns `self.root.line_break_count() + 1` and subtracting 1 can't overflow the `usize`.

helix-term/src/commands.rs
sabify: If this is also guaranteed to be non-zero here, I close this pull request.

PR #3031, opened 2022-07-11
keep jump/file history when using :split

robinvd:
i was missing the beheviour from vim

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3991,24 +3991,11 @@ fn split(cx: &mut Context, action: Action) {
     let (view, doc) = current!(cx.editor);
     let id = doc.id();
     let selection = doc.selection(view.id).clone();
-    let offset = view.offset;
-    let jumps = view.jumps.clone();
-    let docs_access_history = view.docs_access_history.clone();
-    let last_modified_docs = view.last_modified_docs;
-    let object_selections = view.object_selections.clone();
-    let gutters = view.gutters.clone();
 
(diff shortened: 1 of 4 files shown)


helix-term/src/commands.rs
```diff
@@ -3992,12 +3992,23 @@ fn split(cx: &mut Context, action: Action) {
...
+    let gutters = view.gutters.clone();
 
     cx.editor.switch(id, action);
 
     // match the selection in the previous view
     let (view, doc) = current!(cx.editor);
     view.offset = offset;
+    view.jumps = jumps;
+    view.docs_access_history = docs_access_history;
+    view.last_modified_docs = last_modified_docs;
```
pickfire, line 4008: Why do we need to change these?

helix-term/src/commands.rs
```diff
@@ -3992,12 +3992,23 @@ fn split(cx: &mut Context, action: Action) {
...
+    view.object_selections = object_selections;
+    view.gutters = gutters;
```
archseer, line 4010: How about doing this inside `switch`, under the split case? You should be able to just `view.clone()` the current view

archseer (changes requested):

helix-term/src/commands.rs
robinvd: i thought that as well, but the offset was already here, so i just included it haha

robinvd: should be fixed!

archseer (approved):
Thanks!

PR #3033, opened 2022-07-11
jumplist picker

### LINKED ISSUE #2664: Feature Request: Add a picker for the jumplist
Add a visual picker mode for the jumplist, as we have for symbols, files, and buffers.

QiBaobin:
Fixes #2664

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2293,7 +2293,7 @@ fn jumplist_picker(cx: &mut Context) {
             } else {
                 format!(" ({})", flags.join(""))
             };
-            format!("{} {}{} {}", self.id, path, flag, &self.text).into()
+            format!("{} {}{} {}", self.id, path, flag, self.text).into()
         }
     }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -264,6 +264,7 @@ impl MappableCommand {
         file_picker_in_current_directory, "Open file picker at current working directory",
         code_action, "Perform code action",
         buffer_picker, "Open buffer picker",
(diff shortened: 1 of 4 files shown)


helix-term/src/commands.rs
```diff
@@ -2260,6 +2261,87 @@ fn buffer_picker(cx: &mut Context) {
...
+        type Data = ();
+
+        fn label(&self, _data: &Self::Data) -> Spans {
+            let path = self
+                .path
+                .as_deref()
+                .map(helix_core::path::get_relative_path);
+            let path = match path.as_deref().and_then(Path::to_str) {
+                Some(path) => path,
+                None => SCRATCH_BUFFER_NAME,
```
CBenoit, line 2283: I’m not sure `[scratch]` is a good fallback if path is (partially) not valid UTF-8. Did you consider [to_string_lossy](https://doc.rust-lang.org/std/path/struct.Path.html#method.to_string_lossy) instead? This does not incur allocation when data is entirely valid UTF-8.

CBenoit (commented):
Nice!

helix-term/src/commands.rs
QiBaobin: Thanks for the view, Learned from your suggestions. Regarding this comment, I copied the logic from buffer picker, I am not sure if we shall use to_string_lossy, I think we can change it in both places if this is better.

CC: @cossonleo , @the-mikedavis

m-sz: I think that for the buffer it makes perfect sense to fall-back to [scratch] whereas here it seems an alternative is needed, of which to_string_lossy is one of.

PR #3046, opened 2022-07-12
let extend-line respect range direction

QiBaobin:


the-mikedavis:
I think this would be good as a new command but it shouldn't change the existing `x` binding. We could rename the old implementation `extend_line_below` and use that in the default binding for `x`

QiBaobin:
> I think this would be good as a new command but it shouldn't change the existing `x` binding. We could rename the old implementation `extend_line_below` and use that in the default binding for `x`

Good point, did the change based on your suggestion. I keep the new command unbind as I couldn't find a good key for it, I think we can bind it in the config for now.

book/src/keymap.md
```diff
@@ -105,7 +105,7 @@
 | `Alt-(`               | Rotate selection contents backward                                | `rotate_selection_contents_backward` |
 | `Alt-)`               | Rotate selection contents forward                                 | `rotate_selection_contents_forward`  |
 | `%`                   | Select entire file                                                | `select_all`                         |
-| `x`                   | Select current line, if already selected, extend to next line     | `extend_line`                        |
+| `x`                   | Select current line, if already selected, extend to next line     | `extend_line_below`                        |
```
pickfire, line 108:
```suggestion
| `x`                   | Select current line, if already selected, extend to next line     | `extend_line_below`                  |
```

helix-term/src/commands.rs
```diff
@@ -1922,6 +1923,15 @@ enum Extend {
 }
 
 fn extend_line(cx: &mut Context) {
+    let (view, doc) = current_ref!(cx.editor);
+    let extend = match doc.selection(view.id).primary().direction() {
+        Direction::Forward => Extend::Below,
+        Direction::Backward => Extend::Above,
+    };
```
pickfire, line 1930: This would be surprising for me.

QiBaobin: I use extend_above a lot but there is no enough key space to bind it so that I was thing we can provide both functions in one command.
This is not bound to `x` as default, can we keep it so that we can rebind it to this function if need?

dead10ck: I think this is actually very clever. I'm def going to rebind x to this myself.

pickfire: I tried it out again, doesn't seemed like an issue to me.

pickfire (approved):
I tried it out again and doesn't seemed to notice any special differences so I guess should be good. Thanks.

But since this is a breaking change (if user currently specify this themselves), I think we may need to put it in the changelog or so.

the-mikedavis:
Ah sorry, idk how I missed the ping 😅

I'll note the breaking change in the changelog 👍. I'm not sure there's anything we can do to tag them to make it easier - I look at the compare links which don't have labels. A new label could still be useful though since I do watch the PRs.

PR #3047, opened 2022-07-12
Add :show-language command

A-Walrus:
New command that shows the language of the current buffer.


the-mikedavis:
I think this is better off in the statusline which will be added in https://github.com/helix-editor/helix/pull/2434

Either that or a register like #3008, a typable command for this seems like not the right fit

A-Walrus:
Makes sense

PR #3049, opened 2022-07-12
Show "Invalid regex" message on enter (Validate)

### LINKED ISSUE #3034: "Invalid regex" message not shown on enter
### Summary

When selecting/searching with an invalid regex pattern, the "Invalid regex" prompt is only shown once you hit `n` (next), and not on enter.

### Reproduction Steps

 

I tried this:

1. type some text
2. select the text
3. use regex select command with `s`
4. type an invalid regex pattern (ie `[`)
5. hit enter

I expected this to happen:
I see the message `Invalid regex: [` at the bottom

Instead, this happened:
I don't see the message until I hit `n` to go to next selection

### Helix log

```
```

</details>

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.05

A-Walrus:
Fixes #3034

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -117,7 +117,15 @@ pub fn regex_prompt(
 
                             view.ensure_cursor_in_view(doc, config.scrolloff);
                         }
-                        Err(_err) => (), // TODO: mark command line as error
+                        Err(_err) => {
+                            if event == PromptEvent::Validate {
+                                let error = format!("Invalid regex: {}", input);
+                                cx.editor.set_error(error);
+                            } else {
+                                // Update
+                                // TODO: mark command line as error
+                            }
+                        }
                     }
                 }
             }
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -134,7 +134,8 @@ pub fn regex_prompt(
                                                 .position(Some(helix_core::Position::new(
                                                     size.height as usize - 2, // 2 = statusline + commandline
                                                     0,
-                                                )));
+                                                )))
+                                                .auto_close(true);
                                             popup.required_size((size.width, size.height));
 
                                             compositor.replace_or_push("invalid-regex", popup);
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/mod.rs
```diff
@@ -95,7 +95,15 @@ pub fn regex_prompt(
 
                             view.ensure_cursor_in_view(doc, config.scrolloff);
                         }
-                        Err(_err) => (), // TODO: mark command line as error
+                        Err(_err) => {
+                            if event == PromptEvent::Validate {
+                                let error = format!("Invalid regex: {}", input);
+                                cx.editor.set_error(error);
```
sudormrfbin, line 101: If it errors out we should also restore the initial cursor position, identical to what happens in the `PrompEvent::Abort` block a few lines up.

helix-term/src/ui/mod.rs
```diff
@@ -95,7 +95,15 @@ pub fn regex_prompt(
 
                             view.ensure_cursor_in_view(doc, config.scrolloff);
                         }
-                        Err(_err) => (), // TODO: mark command line as error
+                        Err(_err) => {
+                            if event == PromptEvent::Validate {
+                                let error = format!("Invalid regex: {}", input);
```
sudormrfbin, line 100: We should show the exact error returned by the regex library (formatting with `err` should work) since "Invalid input" isn't very useful for fixing the invalid regex.

A-Walrus: The regex errors are multiple lines, for example:
```
regex parse error:
    q(?!u)
     ^^^
error: look-around, including look-ahead and look-behind, is not supported
```
How should they be displayed? Through some pop up window? Or maybe shortened to fit on the one line of the status message?

kirawi: You could try creating a popup above the regex input and see how it looks.

A-Walrus: Added the popup, this is how it looks:
image

archseer: Popups are going to be annoying to clear, what about only taking the last line of the error for now? It contains a description

sudormrfbin: How about making them auto close? That way they get cleared on the next keypress.

PR #3053, opened 2022-07-12
Reuse table widget in picker

sudormrfbin:
Continues #2814. This change reuses the table widget directly in the picker, finally allowing items to be formatted as columns 🎉. Column formatting can be done by implementing `menu::Item::row()` on the picker items.

### Behavioral Changes

- There was a subtle bug in rendering the highlights for matching text which produced incorrect highlighting with multi byte characters. The algorithm to calculate highlights had to be re-written anyway to be injectable into a row of the table, and the issue was fixed along with this:

```
    The fuzzy matcher returns char indices of the choice that matched
    the pattern, while highlighting can only be done on whole graphemes.
    The earlier code assumed char indices == grapheme indices which
    broke highlighting with characters like å.
    
    This commit solves this issue by converting highlighting char indices
    to byte ranges, finding the grapheme which has this byte range as
    it's subrange, and then highlighting the whole grapheme.
```

| Before | After |
| --- | --- |
| picker-unicode-highlighting-incorrect-before | picker-unicode-highlighting-correct-after |

- Pickers that previously aligned items to the right (like the file picker which shows the end of the filename if the window is too narrow) cannot do so anymore; this will require a new `align` property on table cells.

### Bugs 

- [x] I noticed problems with some columns not being displayed randomly when moving up and down the picker (with a modified column formatted version of the diagnostics picker). Requires proper investigation. 

### Future Work
- Use column formatting in buffer picker, symbol picker, etc.
- Add align property to `table::Cell`

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -6,7 +6,9 @@ use crate::{
 use futures_util::future::BoxFuture;
 use tui::{
     buffer::Buffer as Surface,
-    widgets::{Block, BorderType, Borders},
+    layout::Constraint,
+    text::{Span, Spans},
+    widgets::{Block, BorderType, Borders, Cell, Table},
 };
 
 use fuzzy_matcher::skim::SkimMatcherV2 as Matcher;
@@ -19,10 +21,11 @@ use std::{
 use std::{collections::HashMap, io::Read, path::PathBuf};
 
 use crate::ui::{Prompt, PromptEvent};
-use helix_core::{movement::Direction, Position};
+use helix_core::{movement::Direction, unicode::segmentation::UnicodeSegmentation, Position};
 use helix_view::{
     editor::Action,
     graphics::{CursorKind, Margin, Modifier, Rect},
+    theme::Style,
     Document, DocumentId, Editor,
 };
 
@@ -388,6 +391,8 @@ pub struct Picker<T: Item> {
     pub truncate_start: bool,
     /// Whether to show the preview panel (default true)
     show_preview: bool,
+    /// Constraints for tabular formatting
+    widths: Vec<Constraint>,
 
     callback_fn: Box<dyn Fn(&mut Context, &T, Action)>,
 }
@@ -405,6 +410,26 @@ impl<T: Item> Picker<T> {
             |_editor: &mut Context, _pattern: &str, _event: PromptEvent| {},
         );
 
+        let n = options
+            .first()
+            .map(|option| option.row(&editor_data).cells.len())
+            .unwrap_or_default();
+        let max_lens = options.iter().fold(vec![0; n], |mut acc, option| {
+            let row = option.row(&editor_data);
+            // maintain max for each column
+            for (acc, cell) in acc.iter_mut().zip(row.cells.iter()) {
+                let width = cell.content.width();
+                if width > *acc {
+                    *acc = width;
+                }
+            }
+            acc
+        });
+        let widths = max_lens
+            .into_iter()
+            .map(|len| Constraint::Length(len as u16))
+            .collect();
+
         let mut picker = Self {
             options,
             editor_data,
@@ -417,6 +442,7 @@ impl<T: Item> Picker<T> {
             show_preview: true,
             callback_fn: Box::new(callback_fn),
             completion_height: 0,
+            widths,
         };
 
         // scoring on empty input:
@@ -436,8 +462,6 @@ impl<T: Item> Picker<T> {
     }
 
     pub fn score(&mut self) {
-        let now = Instant::now();
-
         let pattern = self.prompt.line();
 
         if pattern == &self.previous_pattern {
@@ -478,8 +502,6 @@ impl<T: Item> Picker<T> {
         } else {
             self.force_score();
         }
-
-        log::debug!("picker score {:?}", Instant::now().duration_since(now));
 
         // reset cursor position
         self.cursor = 0;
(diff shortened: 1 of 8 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -317,6 +321,26 @@ impl<T: Item> Picker<T> {
             |_editor: &mut Context, _pattern: &str, _event: PromptEvent| {},
         );
 
+        let n = options
+            .first()
+            .map(|option| option.row(&editor_data).cells.len())
+            .unwrap_or_default();
```
pickfire, line 327: I don't quite get what is this `n` for, is it the number of cells in the first row? Why do we only get the first row cells count?

sudormrfbin: Yep `n` is the number of cells in the row, and we only take the first row on the assumption that all rows will have the same number of cells. What this block of the code does is calculate the maximum width that each column would need based on the longest item in that particular column. I copied the code from the implementation of [`Menu::recalculate_size()`](https://github.com/helix-editor/helix/blob/62fdcc8cbf7fd7496de2d17b6540b072f41c3344/helix-term/src/ui/menu.rs#L151-L187).

helix-term/src/ui/picker.rs
```diff
@@ -602,62 +623,120 @@ impl<T: Item + 'static> Component for Picker<T> {
...
-                                highlighted.patch(span.style)
-                            } else if is_active {
-                                selected.patch(span.style)
+            .take(rows as usize)
+            .map(|(index, _score)| &self.options[*index])
+            .map(|option| option.row(&self.editor_data))
+            .map(|mut row| {
+                const TEMP_CELL_SEP: &str = " ";
+
+                let line = row.cell_text().join(TEMP_CELL_SEP);
```
pickfire, line 643: This will allocate at least twice, once on `Vec` and once on `join`, I wonder if it would be better if we just output to `String` directly.

helix-term/src/ui/picker.rs
```diff
@@ -602,62 +623,120 @@ impl<T: Item + 'static> Component for Picker<T> {
...
+
+                    let mut span_list: Vec<(String, Style)> = Vec::new();
+                    for (grapheme, style) in graphemes_with_style {
+                        if span_list.last().map(|(_, sty)| sty) == Some(&style) {
+                            let (string, _) = span_list.last_mut().unwrap();
+                            string.push_str(grapheme);
+                        } else {
+                            span_list.push((String::from(grapheme), style))
+                        }
+                    }
```
pickfire, line 708: Hmm, do we really have to do all these here?

helix-term/src/ui/picker.rs
```diff
@@ -602,62 +623,120 @@ impl<T: Item + 'static> Component for Picker<T> {
...
+
+                let highlight_byte_ranges: Vec<_> = line
+                    .char_indices()
+                    .enumerate()
+                    .filter_map(|(char_idx, (byte_offset, ch))| {
+                        highlights
+                            .contains(&char_idx)
+                            .then(|| byte_offset..byte_offset + ch.len_utf8())
+                    })
+                    .collect();
```
pickfire, line 662: I noticed this function have a lot of collects, and this `highlight_byte_ranges` is even doing the work for those that isn't rendered? Do we need all these middleman doing allocations, should we delay the highlight to the painting part and only do it for those that are necessary (here it seemed to even highlight those that are not for displayed)?

helix-tui/src/text.rs
```diff
@@ -428,6 +428,19 @@ impl<'a> From<Vec<Spans<'a>>> for Text<'a> {
     }
 }
 
+impl<'a> From<Text<'a>> for String {
+    fn from(text: Text<'a>) -> String {
+        let lines: Vec<String> = text.lines.iter().map(String::from).collect();
+        lines.join("\n")
```
pickfire, line 434: Would collecting into `String` directly be faster?

sudormrfbin:
The table column width calculation code has been updated to match the upstream implementation in the `tui` crate (specifically https://github.com/fdehau/tui-rs/commit/a68e38e59e6735c0a99139303b1609669d2c38da), and the random swallowing of columns have been fixed.

helix-term/src/ui/picker.rs
sudormrfbin: We do if we want accurate highlighting on the matched text -- see the `Behavioral Changes` note in my initial PR comment for an explanation. That said this area can definitely be refactored, but I reckon the logic would be more or less the same.

helix-term/src/ui/picker.rs
sudormrfbin: Seems like a good idea 👍🏾

helix-tui/src/text.rs
sudormrfbin: Done in b28f684d1d.

gibbz00:
@sudormrfbin I'm working on a branch that aims to refine the keymap presentation in both the info boxes and the command palette. Wanted to see how the command palette could look with column formatting, so I've already started on an implementation for that hehe.
 
https://github.com/helix-editor/helix/commit/5f8ba8cb0d0857e942182ae8c440a869745a6ce9

I will post the draft during the week, only marking it as a PR once this one gets merged. Feel free to use the code in the commit in the meanwhile in case mine doesn't go through, I'll just make sure to omit the commit from my draft if you do :)

archseer (changes requested):
Great work as always!

archseer (approved):
(oops)

PR #3057, opened 2022-07-13
respect count for repeating motion

QiBaobin:


pickfire (approved):
Looks good to me. Thanks.

Although I am not sure if `2alt-.` should select two selection altogether but this behavior is fine as well.

PR #3058, opened 2022-07-13
respect count in treesitter movement

QiBaobin:


pickfire (approved):
I am okay with this, but count does not seemed that useful for me in treesitter even though this makes it more consistent. I wonder what @the-mikedavis would say.

the-mikedavis (approved):
I think this would be useful 👍

You might want to jump two functions away or to the third argument/parameter. I think it's certainly better than having the count do nothing :smile:

PR #1352, opened 2021-12-24
Open the File Picker in the folder of the active buffer

bbodi:
In Programming languages with deeply nested directory hierarchies (e.g. Java or TypeScript in an Angular Project), files in the same folder/package level often needs to be easily accessed or explored.
The deep level of directory hierarchies and the long file names these projects usually use make it difficult to find those files via the default File Picker.

In IntelliJ, one can press `alt+home` to get a navigatable dialog widget to explore/access the files in the currently editing file's folder.
image

A similar functionality could be achieved if there was a way to open the File Picker from the active file's folder.

PR #3063, opened 2022-07-14
Add File sub keymap

### LINKED ISSUE #3052: Feature Requst: File picker based on current buffer's directory
I will often "go to definition" for library code and then want to explore that library, but have no good way of doing that.

Being able to open a picker based on the current buffer's file location, and having a way to "go up a level" in the picker (like emacs' helm-mode's `ctrl-l`) would solve this.

I am also interested in if anyone has a workaround for exploring library code without re-opening helix in the library directory.

QiBaobin:
Fixes #3052


poliorcetics:
Sub keycaps should not use uppercase keys IMO, it makes them at least 4 chars to type and typing `Shift+<key>` is not always practical depending on key (it never is in my opinion, but that's more drastic)

It's more breaking but maybe `space+f` is ok ? The current behaviour would be recuperated by doing `space+f+f`, which is easy to type quickly

book/src/keymap.md
```diff
@@ -257,6 +258,19 @@ This layer is a kludge of mappings, mostly pickers.
 
 > TIP: Global search displays results in a fuzzy picker, use `space + '` to bring it back up after opening a file.
 
+#### File mode
+
+This layer is a kludge of mappings, mostly pickers.
```
the-mikedavis, line 263: This could probably be something like "a collection of file-opening commands" or something similar

the-mikedavis:
I like this but I agree about `F` being tough because of the extra modifier. We tend to prefer sequences to modifiers (for example `v` instead of shifted motions like in Kakoune). As much as I like the very short `<space>f` for the main file picker, maybe it's ok to add an extra `f` in there, eh? I suppose doom emacs / spacemacs already have this same binding. What do others think?

sudormrfbin:
`<space>f` is rather short and sweet, and I'm not sure how frequently the other newly introduced pickers will be used. How much would `<space>fd` be used compared to `<space>ff` ? In projects where it makes sense to [open the current-buffer-directory picker more often](https://github.com/helix-editor/helix/pull/1352#issue-1088323293), the default `<space>f` could be binded to the buffer-directory picker. I personally have `backspace` in normal mode mapped to the file picker so that it's takes the most minimal number of keystrokes :P

----

If we do decide to go with `<space>f` as a submode, `<space>fF` (open picker at CWD) should be changed to something easier to type.

poliorcetics:
> the default <space>f could be binded to the buffer-directory picker

Having to change learned behavior and remember to make a new local config depending on the project seems wobbly, I would hate that. Configuring my language server to pick up project specific details is one thing, keybindings are supposed to be more "computer wide" IMO

archseer:
We could turn space+F into a submode instead and keep space+f as a faster shortcut for the default?

poliorcetics:
> We could turn space+F into a submode instead and keep space+f as a faster shortcut for the default?

That was the original behavior in the PR that I argued against, modes should not use uppercase, it's bothersome to type

archseer:
It's a less often used mode though. Users can always add their own keymaps to access certain pickers easier. I'm open to other suggestions but I don't want to remap `space+f`

book/src/keymap.md
sudormrfbin: This got reverted too by the way.

QiBaobin: Thanks, added it back now.

mangas:
Hey guys, I just found this issue while searching for something similar in terms of remapping space keybinds. For my own use I have swapped f and F because F is trickier to use because of shift, and because most of the times they do the same thing. When I do select a directory then I don't need to change the binding. I also think most of the other keybinds use only one key so adding space+ff seems weird to me (unless we make it a sub-command and in that case it would be what? space+ff and space+fc?) I think swapping f and F is prolly a good default as it behaves the same in most cases (I would think). 

Is there a reason we can't add the config to rebind space keybinds? If not then maybe that could be a starter?

book/src/keymap.md
```diff
@@ -257,6 +258,19 @@ This layer is a kludge of mappings, mostly pickers.
...
+#### File mode
+
+This layer is a collection of file-opening commands.
+
+
+| Key     | Description                                                             | Command                             |
+| -----   | -----------                                                             | -------                             |
+| `f`     | Open file picker                                                        | `file_picker`                       |
+| `F`     | Open file picker at current working directory                           | `file_picker_in_current_directory`  |
+| `d`     | Open file picker at current buffer's directory                          | `file_picker_in_buffer_directory`   |
```
the-mikedavis, line 292: What do you think about `b` for this?

the-mikedavis:
> Is there a reason we can't add the config to rebind space keybinds?

This is possible:

```toml
[keys.normal]
space.F.c = ":config-open"
```

QiBaobin:
close per #4783

PR #3064, opened 2022-07-14
File keymap yank file name

### LINKED ISSUE #2985: Add a command to copy the filename and cursor line number in the clipboard.

QiBaobin:
Fixes #2985

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -270,6 +270,7 @@ This layer is a kludge of mappings, mostly pickers.
 | `d`     | Open file picker at current buffer's directory                          | `file_picker_in_buffer_directory`   |
 | `c`     | Open the helix config.toml file                                         | `config_open`                       |
 | `l`     | Open the helix log file                                                 | `log_open`                          |
+| `y`     | Yank current buffer's file name to clipboard                            | `yank_file_name_to_clipboard`       |
 
 ##### Popup
 
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -237,6 +237,7 @@ This layer is a kludge of mappings, mostly pickers.
 | Key     | Description                                                             | Command                             |
 | -----   | -----------                                                             | -------                             |
 | `f`     | Open file picker                                                        | `file_picker`                       |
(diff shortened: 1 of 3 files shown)


sudormrfbin:
Going the special register route and storing the filename on the `%` register like vim and kakoune might be better here since registers can be used in prompt areas like the command line and picker directly.

sudormrfbin:
Ah I didn't see #3008. If it adds `%` as the filename register, I don't think we need a separate keybind for copying filenames since it can be done with `"%y` or `"%<space>y"`.

QiBaobin:
> 



> Ah I didn't see #3008. If it adds `%` as the filename register, I don't think we need a separate keybind for copying filenames since it can be done with `"%y` or `"%<space>y"`.

Does `"%y` or `"%<space>y"` copy the selection to the register instead of copying from it?

sudormrfbin:
Nope, `y` always copies from the register, and `"%` selects the register to copy from.

QiBaobin:
> Nope, `y` always copies from the register, and `"%` selects the register to copy from. Since the `%` register would hold the name of the file, `"%y` would copy the the filename into the default register and `"%<space>y` would copy the filename to the system clipboard.

This is not what I understand, I think `"%y` mean copy the selection to register %, this is constant with the behavior as below:

https://user-images.githubusercontent.com/7072/179348195-337d2ba7-b27f-49b0-a4d5-14f1ed23842a.mov

To copy content from a register to another, I think we need paste from it and select it then copy to another. Let me know if I miss some features here.

sudormrfbin:
Ah you're right, I mixed up `y` and `ctrl-r` in insert mode, sorry for the confusion 😅. In that case I think it's better to have a typable `:command` instead of a keybind for copying the filename to the system clilpboard.

QiBaobin:
> You're right, I mixed up `y` and `ctrl-r` in insert mode, sorry for the confusion 😅. In that case I think it's better to have a typable `:command` instead of a keybind for copying the filename to the system clilpboard. I don't think it's a frequent enough thing to do that would require a default keybind, but if required it can be easily binded in the config file.

Sure

PR #3066, opened 2022-07-14
Update tutor.txt

rykersun:
Show the other usage of J.

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -630,8 +630,8 @@ onto other
 lines.
      This sentence is spilling over onto other lines.
 
- Note: J works on all lines in selection, so typing xxx or 3x to
-       select the lines and then a single J will work the same.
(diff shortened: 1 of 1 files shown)


runtime/tutor.txt
```diff
@@ -630,8 +630,8 @@ onto other
 lines.
      This sentence is spilling over onto other lines.
 
- Note: J works on all lines in selection, so typing xxx or 3x to
-       select the lines and then a single J will work the same.
+ Note: J also works on single lines, so typing JJJ will work the same.
```
pickfire, line 633: Showing the other usage is fine, why are we removing the old text? Are they not needed?

the-mikedavis: A prior update switched which one was the note and which one was the main tutor lesson. Currently the note has the same instructions (but with the wrong count actually) as the lesson so this is showing the other use of J

PR #3067, opened 2022-07-14
Inherit theme

### LINKED ISSUE #2454: Add ability to derive from a base theme

Currently if I want to experiment with changes to a theme, I need to copy the original theme file and modify this copy. If the original is updated from upstream, I don't get those updates. If I haven't made a copy and directly modified the original, it's either overwritten, losing my changes, or causes a merge conflict if I have it in a local branch of `helix`.

I would like the ability to have something like `[fallback] name = "gruvbox"` (or `[origin]`, `[source]`, etc, name is bikeshedable) in theme files, so that I can modify only the keys that interest me while keeping the rest.

This could work by merging the TOMLs, like for the user config and the default one. Merging the TOMLs would also let me use colors declared in the original or override them.

It could even be done recursively so that I can have `gv2 -> gv1 -> gruvbox` if necessary, but that's just a nice-to-have.

TheSamsa:
This PR provides a solution to resolve #2454

### DIFF
--- a/helix-view/src/theme.rs
+++ b/helix-view/src/theme.rs
@@ -46,11 +46,12 @@ impl Loader {
             return Ok(self.base16_default());
         }
 
-        let mut raw_theme: RawTheme = self.load_raw(name)?;
+        let path = self.path(name, false);
+        let mut raw_theme: RawTheme = self.load_raw(path)?;
 
         if let Some(parent_theme_name) = &raw_theme.inherits_from {
-            //let parent_theme_name = "bogster";
-            let parent_raw_theme = self.load_raw(parent_theme_name)?;
+            let path = self.path(parent_theme_name, parent_theme_name == name);
+            let parent_raw_theme = self.load_raw(path)?;
 
             raw_theme.inherit(parent_raw_theme);
         }
@@ -74,19 +75,18 @@ impl Loader {
     }
 
     // Loads the raw theme data first from the user_dir then in default_dir
-    fn load_raw(&self, name: &str) -> Result<RawTheme, anyhow::Error> {
-        let path = self.path(name);
+    fn load_raw(&self, path: PathBuf) -> Result<RawTheme, anyhow::Error> {
         let data = std::fs::read(&path)?;
 
         toml::from_slice(data.as_slice()).context("Faled to deserialize theme")
     }
 
     // Returns the path to the theme name
-    fn path(&self, name: &str) -> PathBuf {
+    fn path(&self, name: &str, only_default_dir: bool) -> PathBuf {
         let filename = format!("{}.toml", name);
 
         let user_path = self.user_dir.join(&filename);
-        if user_path.exists() {
+        if !only_default_dir && user_path.exists() {
             user_path
         } else {
             self.default_dir.join(filename)
--- a/helix-view/src/theme.rs
+++ b/helix-view/src/theme.rs
@@ -42,17 +42,21 @@ impl Loader {
         if name == "base16_default" {
             return Ok(self.base16_default());
         }
-        let filename = format!("{}.toml", name);
 
-        let user_path = self.user_dir.join(&filename);
-        let path = if user_path.exists() {
-            user_path
-        } else {
-            self.default_dir.join(filename)
-        };
 
+        // let mut raw_theme: RawTheme = self.load_raw(name)?;
+
+        // if let Some(parent_theme_name) = &raw_theme.inherits_from {
+        //     //let parent_theme_name = "bogster";
+        //     let parent_raw_theme = self.load_raw(parent_theme_name)?;
+
+        //     raw_theme.inherit(parent_raw_theme);
+        // }
+
+        let path = self.path(name);
         let data = std::fs::read(&path)?;
-        toml::from_slice(data.as_slice()).context("Failed to deserialize theme")
+
+        toml::from_slice(data.as_slice()).context("Faled to deserialize theme")
     }
 
     pub fn read_names(path: &Path) -> Vec<String> {
@@ -68,6 +72,28 @@ impl Loader {
                     .collect()
             })
             .unwrap_or_default()
+    }
+
+    // Loads the raw theme data first from the user_dir then in default_dir
+    fn load_raw(&self, name: &str) -> Result<RawTheme, anyhow::Error> {
+        let path = self.path(name);
+        let data = std::fs::read(&path)?;
+
+        toml::from_slice(data.as_slice()).context("Faled to deserialize theme")
+    }
+
+    // Returns the path to the theme name
+    fn path(&self, name: &str) -> PathBuf {
+        let filename = format!("{}.toml", name);
+
+        let user_path = self.user_dir.join(&filename);
+        let path = if user_path.exists() {
+            user_path
+        } else {
+            self.default_dir.join(filename)
+        };
+
+        path
     }
 
     /// Lists all theme names available in default and user directory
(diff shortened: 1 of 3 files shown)


kirawi:
I think it could be simplified further by having the deserializer for `Theme` use the deserializer for `RawTheme`.

TheSamsa:
I am not familiar with serde to know how to use one `Deserializer` inside another one, or how to invoke the deserialization itself.

Also I guess this would not work fully if the theme has a base theme, because right now the deserialization of the `RawTheme` only stores the name of the `inherit_from` theme only as string, and does not load it automatically. (I didn't want to put a file operation inside the `deserialize` method)

archseer:
This `RawTheme` seems unnecessary. I'd deserialize both files to `toml::Value`, then use [`merge_toml_values`](https://github.com/helix-editor/helix/blob/d0c63409ccb9eae332e7dea5788f54f1af412d5a/helix-loader/src/config.rs#L38) like we do for the config, then deserialize the merged value into a theme. If you merge the new theme into the parent, it'll override values as needed.

TheSamsa:
@archseer 
I tested it with `merge_toml_values` but it does not work as I thought it would.
Example: i use this to remove a modifier in the gruvbox theme, from this:
`"ui.selection" = { bg = "bg3", modifiers = ["reversed"] }` to `"ui.selection" = { bg = "bg3" }`
But with the mechanic of `merge_toml_values` (https://github.com/helix-editor/helix/blob/master/helix-loader/src/lib.rs#L169-L186) it does not remove the modifier it keeps it, no matter which theme is passed as the left or right into the function.

the-mikedavis:
You'll want to use a lower merge_depth so style maps themselves are not merged. You might need to split the palette out of the toml before merging though since it has `merge_depth + 1` compared to the styles.

TheSamsa:
@the-mikedavis thanks for the hints. I removed the RawTheme and use the `toml::Value` directly, although it did require some special handling because of the palette as you correctly stated.
But it works now, even with multiple stages of inherit_from definitions. Although it is probably possible to introduce circular loading of themes. Nonetheless I thought it might be the wish to be able to use a chain of theme inheritance.

helix-view/src/theme.rs
```diff
@@ -42,17 +45,41 @@ impl Loader {
...
+    fn load_theme(
+        &self,
+        name: &str,
+        base_them_name: &str,
+        only_default_dir: bool,
+    ) -> Result<Value, anyhow::Error> {
+        let path = self.path(name, only_default_dir);
+        let theme_toml = self.load_toml(path)?;
+
+        let inherits_from = theme_toml.get("inherits_from");
```
the-mikedavis, line 66: Let's just call this `inherits` which we use already in the tree-sitter queries

helix-view/src/theme.rs
```diff
@@ -42,17 +45,41 @@ impl Loader {
...
+
+        let theme_toml = if let Some(parent_theme_name) = inherits_from {
+            let parent_theme_name = parent_theme_name.to_string().replace('\"', "");
+
+            let parent_theme_toml = self.load_theme(
+                &parent_theme_name,
+                base_them_name,
+                base_them_name == parent_theme_name,
+            )?;
```
the-mikedavis, line 75:
```suggestion
            let parent_theme_name = parent_theme_name.as_str().ok_or(anyhow!(
                "Theme: expected 'inherits' to be a string: {}",
                parent_theme_name
            ))?;

            let parent_theme_toml = self.load_theme(
                parent_theme_name,
                base_them_name,
                base_them_name == parent_theme_name,
            )?;
```

The `"`s are because this is a `toml::Value`. We'll want to check that it's parsed as a string and fail if it isn't (for example, a theme provides a list of strings of theme names to inherit).

helix-view/src/theme.rs
```diff
@@ -70,6 +97,58 @@ impl Loader {
...
+            } else {
+                parent_palette.clone()
+            }
+        } else {
+            if let Some(palette) = palette {
+                palette.clone()
+            } else {
+                Map::new().into()
+            }
+        };
```
the-mikedavis, line 119: This is a little cleaner as a `match`

```suggestion
        let palette_values = match (parent_palette, palette) {
            (Some(parent_palette), Some(palette)) => {
                merge_toml_values(parent_palette.clone(), palette.clone(), 2)
            }
            (Some(parent_palette), None) => parent_palette.clone(),
            (None, Some(palette)) => palette.clone(),
            (None, None) => Map::new().into(),
        };
```

helix-view/src/theme.rs
```diff
@@ -70,6 +97,58 @@ impl Loader {
...
+        palette.insert(String::from("palette"), palette_values);
+
+        // merge the theme into the parent theme
+        let theme = merge_toml_values(parent_theme_toml, theme_toml, 1);
+        // merge the before specially handled palette into the theme
+        merge_toml_values(theme, palette.into(), 1)
+    }
+
+    // Loads the theme data as `toml::Value` first from the user_dir then in default_dir
+    fn load_toml(&self, path: PathBuf) -> Result<toml::Value, anyhow::Error> {
```
the-mikedavis, line 132: In a few places there's this `Result<toml::Value, anyhow::Error>` - you can `use anyhow::Result` to bring in the anyhow result which is a little cleaner: `Result<toml::Value>`

helix-view/src/theme.rs
TheSamsa: will do

helix-view/src/theme.rs
TheSamsa: Thanks I somehow missed the `as_str` method, still a far way to get Rust into my system.

But clippy suggested to use `ok_or_else(|| anyhow!{...}` instead of `ok_or` which I did, since the pipeline checks with clippy.

helix-view/src/theme.rs
TheSamsa: You are right of course, was only thinking about let chains and this easy and clean solution didn't come to mind.

helix-view/src/theme.rs
TheSamsa: changed all the occurences to the `anyhow::Result` type

Omnikar:
@the-mikedavis Anything still left to be done here before merging? I'd really like to see this feature land.

helix-view/src/theme.rs
```diff
@@ -35,24 +38,53 @@ impl Loader {
...
             return Ok(self.base16_default());
         }
-        let filename = format!("{}.toml", name);
 
-        let user_path = self.user_dir.join(&filename);
-        let path = if user_path.exists() {
-            user_path
+        let theme_toml = self.load_theme(name, name, false)?;
+
+        Ok(Theme::from(theme_toml))
```
the-mikedavis, line 51:
```suggestion
        self.load_theme(name, name, false).map(Toml::from)
```

the-mikedavis (approved):

helix-view/src/theme.rs
```diff
@@ -105,16 +182,16 @@ pub struct Theme {
...
+    fn from(value: Value) -> Self {
         let mut styles = HashMap::new();
         let mut scopes = Vec::new();
         let mut highlights = Vec::new();
 
-        if let Ok(mut colors) = HashMap::<String, Value>::deserialize(deserializer) {
+        let theme_values: Result<HashMap<String, Value>> =
+            toml::from_str(&value.to_string()).context("Failed to load theme");
+
+        if let Ok(mut colors) = theme_values {
```
archseer, line 194:
```
HashMap::<String, Value>::deserialize(value)
```

Should still work since `toml::Value` implements `Deserialize`.

helix-view/src/theme.rs
```diff
@@ -105,16 +182,16 @@ pub struct Theme {
     highlights: Vec<Style>,
 }
 
-impl<'de> Deserialize<'de> for Theme {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
+impl From<Value> for Theme {
+    fn from(value: Value) -> Self {
```
archseer, line 186: Why the change? It now takes extra steps since you first call `from_slice` then `Theme::from`

archseer (changes requested):

helix-view/src/theme.rs
TheSamsa: I implemented the `From` trait for `Theme` because I could not directly deserialize from a file. Since we would need to load the parent theme while deserializing. A pattern which I did not like.
But we could implement both and share the code, I'll provide a solution.

helix-view/src/theme.rs
TheSamsa: It does for a `Deserialize` trait implementation, see the answer below.

PR #3069, opened 2022-07-14
Add section for repeat action

stirlhoss:
This is just an addition that I made to the tutor file. I want to do more of these but I want to get one submitted so I know if there are things I need to change or do better with. 

I added repeating actions to section 4 and updated the section 4 summary.

### DIFF
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -550,7 +550,8 @@ make changes across multiple lines.
  * Type Q to record and stop recording a macro.
    * Type q to repeat the recorded macro.
 
-* Type . to repeat the last set of actions entered.
+* Type . to repeat the last insert.
+* Type A-. to repeat the last motion.
 
 
 =================================================================
--- a/runtime/tutor.txt
+++ b/runtime/tutor.txt
@@ -504,6 +504,29 @@ _________________________________________________________________
      This sentence doesn't have it's first and last word.
 
 
+=================================================================
+=                     4.4 Repeating Actions                     =
+=================================================================
+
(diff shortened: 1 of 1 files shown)


the-mikedavis:
It looks like @n0s4 is covering these in the big update in #2798: https://github.com/helix-editor/helix/pull/2798/files#diff-0751fcdc9d0b2b023e5c02ee4941114ca1706d04e1ba73c2f4c5ba287c847074R661-R674

`.` is actually to repeat the last insert rather than the last action which was a common point of confusion that has recently been fixed in the docs: https://github.com/helix-editor/helix/pull/2878

stirlhoss:
Made updates with taking your comments into consideration. It is difficult to come up with examples for this feature that aren't either too complicated or too basic. I went basic this time. I think that functions like this are things that once you get a feel for when you can use it, it sort of just clicks. Let me know what you guys think.

runtime/tutor.txt
```diff
@@ -504,6 +504,36 @@ _________________________________________________________________
...
+3. Go to the next instance and select the word again.
+4. Type .
+5. Move to the next and repeat until they all say 'Orange'.
+
+--> Banana Banana Banana Banana Banana Banana 
+
+To repeat motion:
+
+1. Move the Cursor to the first line below marked -->.
+2. Press 'f' then 'O' to find the next instatne of 'O'
```
pythoneer, line 530:
```suggestion
2. Press 'f' then 'O' to find the next instance of 'O'
```

archseer:
Ended up merging https://github.com/helix-editor/helix/pull/2798

PR #3070, opened 2022-07-14
Shorten embedded command descriptions

sudormrfbin:
This mainly gives us more compact descriptions that will fit better with the command palette and the infobox.


helix-term/src/commands.rs
```diff
@@ -225,7 +225,7 @@ impl MappableCommand {
         find_prev_char, "Move to previous occurrence of char",
         extend_till_prev_char, "Extend till previous occurrence of char",
         extend_prev_char, "Extend to previous occurrence of char",
-        repeat_last_motion, "repeat last motion(extend_next_char, extend_till_char, find_next_char, find_till_char...)",
+        repeat_last_motion, "Repeat last motion",
```
the-mikedavis, line 228: Somewhat unrelated to the change here, this tends to be a source of confusion that `A-.` doesn't work on motions like `w`, just on these find-type motions (same behavior as Kakoune). Maybe we should have a different word for those kinds of motions?

helix-term/src/commands.rs
```diff
@@ -413,21 +413,21 @@ impl MappableCommand {
...
+        shell_insert_output, "Insert shell command output before selections",
+        shell_append_output, "Append shell command output after selections",
         shell_keep_pipe, "Filter selections with shell predicate",
-        suspend, "Suspend",
+        suspend, "Suspend and return to shell",
         rename_symbol, "Rename symbol",
-        increment, "Increment",
-        decrement, "Decrement",
+        increment, "Increment item under cursor",
+        decrement, "Decremen item under cursort",
```
the-mikedavis, line 430:
```suggestion
        decrement, "Decrement item under cursort",
```

the-mikedavis (commented):
These all look great to me!

Other than that typo :P

PR #3073, opened 2022-07-15
Set the selection point of the first file correctly

shnarazk:
After f10b6f6e, `hx file:row` doesn't work, because the first file skips `doc.set_selection()` in a loop now.
So I rewrote it as just a loop without `.skip(1)`.
It seems to work well.


archseer:
There's another fix here https://github.com/helix-editor/helix/pull/2983/files

But this one might be more straightforward.

PR #3080, opened 2022-07-15
Refactor 'helix-loader::merge_toml_values' to use a 'merge-depth' instead of 'merge_toplevel_arrays'

Philipp-M:
This ensures that other values than just the top-level arrays are overridden, and makes this function generally more flexible, like merging nested objects, where it makes sense.

For the `languages.toml`, `merge_depth` is set to 3 so that top-level language features are merged (like 'scope'), but everything deeper is overridden with the user-config. With the current configuration schema the result should be equivalent to the current behavior. 

This change is a requirement for #2507 since language-servers there, are configured via a table instead of an array.

### DIFF
--- a/helix-loader/src/config.rs
+++ b/helix-loader/src/config.rs
@@ -19,7 +19,23 @@ pub fn user_lang_config() -> Result<toml::Value, toml::de::Error> {
         .into_iter()
         .chain([default_lang_config()].into_iter())
         .fold(toml::Value::Table(toml::value::Table::default()), |a, b| {
-            crate::merge_toml_values(b, a, true)
+            // combines for example
+            // b:
+            //   [[language]]
+            //   name = "toml"
+            //   language-server = { command = "taplo", args = ["lsp", "stdio"] }
+            //
+            // a:
+            //   [[language]]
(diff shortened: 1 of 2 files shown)


helix-loader/src/config.rs
```diff
@@ -19,7 +19,7 @@ pub fn user_lang_config() -> Result<toml::Value, toml::de::Error> {
         .into_iter()
         .chain([default_lang_config()].into_iter())
         .fold(toml::Value::Table(toml::value::Table::default()), |a, b| {
-            crate::merge_toml_values(b, a, true)
+            crate::merge_toml_values(b, a, 3)
```
the-mikedavis, line 22: I think it's worth a comment here explaining why it's 3 for the languages.toml (just here is probably fine, I don't think it needs to be repeated in the tests)

the-mikedavis (approved):
This looks good to me 👍

The boolean did seem a little hacky, a merge depth looks much more appropriate.

helix-loader/src/config.rs
Philipp-M: Is this better?

archseer: Yeah, great writeup! 👍

PR #3082, opened 2022-07-15
Allow adding options in menu and picker

Philipp-M:
This PR allows retaining the current selected option in either the menu or the picker, when calling `score` (as long, as it's still in there, after calling `score`).
This is useful when extending the options later (via a new function `add_options`) in the picker or menu, which is the second addition in this PR.

This change is a requirement for #2507. In itself it doesn't change the current behavior of the editor.
In #2507, this is needed, because multiple language-servers send their results async (e.g. completion, code-actions or symbol-picker)

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -57,6 +57,8 @@ pub struct Menu<T: Item> {
     /// (index, score)
     matches: Vec<(usize, i64)>,
 
+    previous_pattern: String,
+
     widths: Vec<Constraint>,
 
     callback_fn: Box<dyn Fn(&mut Editor, Option<&T>, MenuEvent)>,
@@ -83,6 +85,7 @@ impl<T: Item> Menu<T> {
             editor_data,
             matcher: Box::new(Matcher::default()),
             matches,
+            previous_pattern: String::new(),
             cursor: None,
             widths: Vec::new(),
             callback_fn: Box::new(callback_fn),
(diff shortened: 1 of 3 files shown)


helix-term/src/ui/menu.rs
```diff
@@ -92,36 +95,70 @@ impl<T: Item> Menu<T> {
...
+                        // TODO: using fuzzy_indices could give us the char idx for match highlighting
+                        self.matcher
+                            .fuzzy_match(&text, &self.previous_pattern)
+                            .map(|score| (index, score))
+                    }),
+            );
+            // matches.sort_unstable_by_key(|(_, score)| -score);
+            self.matches.sort_unstable_by_key(|(index, _score)| {
+                self.options[*index].sort_text(&self.editor_data)
+            });
```
pickfire, line 132: Is it a good idea to run `sort_text` within a sort?

helix-term/src/ui/picker.rs
```diff
@@ -331,27 +335,35 @@ impl<T: Item> Picker<T> {
...
-                .map(|(index, _option)| (index, 0)),
-        );
-
+        picker.re_score(None, true, true);
         picker
     }
 
-    pub fn score(&mut self) {
-        let now = Instant::now();
+    pub fn re_score(
```
pickfire, line 342: What does `re_score` means?

helix-term/src/ui/picker.rs
```diff
@@ -331,27 +335,35 @@ impl<T: Item> Picker<T> {
...
+            self.cursor = 0
+        }
 
-        let pattern = self.prompt.line();
+        let pattern = match pattern {
+            Some(pattern) if pattern == self.previous_pattern && !force_recalculation => return,
+            None if !force_recalculation => return,
+            None => self.previous_pattern.clone(),
+            Some(pattern) => pattern,
+        };
```
pickfire, line 358: Would it be better to remove `force_recalculation` in this function and only call this function if it needs to force recalculation?

helix-term/src/ui/menu.rs
Philipp-M: What do you mean exactly?
This is actually untouched code. Just for clarification `sort_text` doesn't actually sort the text, but returns the text to be used for sorting

helix-term/src/ui/picker.rs
Philipp-M: Yeah that might probably be better, I'll tackle this soon...

helix-term/src/ui/picker.rs
pickfire: Yeah

helix-term/src/ui/picker.rs
Philipp-M: I've replaced it with a dirty flag, is this better?

pickfire: The previous one seemed better to me, can keep it in the argument I guess, don't need to track states.

helix-term/src/ui/picker.rs
```diff
@@ -413,9 +424,33 @@ impl<T: Item> Picker<T> {
...
+
+    pub fn score(&mut self, reset_cursor: bool) {
+        let pattern = self.prompt.line().clone();
+        self.recalculate_score(Some(pattern), reset_cursor);
+    }
+
+    pub fn add_options(&mut self, mut options: Vec<T>) {
+        self.options.append(&mut options);
+        self.dirty = true;
+        self.recalculate_score(None, false);
```
pickfire, line 453: `self.dirty == true && patttern == None` here will never recalculate, is that what we want?

helix-term/src/ui/picker.rs
```diff
@@ -413,9 +424,33 @@ impl<T: Item> Picker<T> {
...
+    }
+
+    pub fn score(&mut self, reset_cursor: bool) {
+        let pattern = self.prompt.line().clone();
+        self.recalculate_score(Some(pattern), reset_cursor);
+    }
+
+    pub fn add_options(&mut self, mut options: Vec<T>) {
+        self.options.append(&mut options);
+        self.dirty = true;
```
pickfire, line 452: It doesn't seemed to reset to `false`? But yeah, passing it in `recalculate_score` is better.

helix-term/src/ui/picker.rs
Philipp-M: It should?

```rust
...
            None if !self.dirty => return,
            None => self.previous_pattern.clone(),
...
```
Second match should match.

helix-term/src/ui/picker.rs
Philipp-M: resetted it to the old version...

helix-term/src/ui/menu.rs
```diff
@@ -58,6 +58,8 @@ pub struct Menu<T: Item> {
     /// (index, score)
     matches: Vec<(usize, i64)>,
 
+    previous_pattern: String,
```
pickfire, line 61: Why do we need a previous pattern? Why can't just take from the pattern from the prompt?

Philipp-M: This is necessary for the `add_options` method, because it needs to recalculate the showed options (`recalculate_score`), without knowing the current pattern.
In the future, this could also help optimizing/filtering previous matches similar as in the picker.

Philipp-M: Well as said before, the picker already had it in there, so I just adopted this in the menu as well. 
For example `self.matches` is also dependent on the current pattern and stored inside this Component. Actually all state necessary for rendering (or managing state in) the Menu (or Picker) is stored in the Component.

It e.g. makes sense for caching, prefiltering (as done in the picker) etc. 

I think it's maybe better to completely separate data and/or state/model and the view (completion menu in particular), but this is a story(or big PR) for itself, and I'm not sure if it makes the code much better/readable/compact etc.

Philipp-M:
I'm going to close this, as this was mostly a hack to get #2507 working in an async way (i.e. adding more options of a different completion source/LSP later). In #2507 we came to the consensus to go this "sync" (i.e. all completion sources have to finish before a menu is shown). I don't think this should be used otherwise...

PR #3091, opened 2022-07-17
feat(tree-sitter): :sparkles: add go template support

dannylongeuay:
Adds Go Template highlighting support.

image

### DIFF
--- a/runtime/queries/gotmpl/highlights.scm
+++ b/runtime/queries/gotmpl/highlights.scm
@@ -1,0 +1,75 @@
+; Identifiers
+
+[
+  (field)
+  (field_identifier)
+] @property
+
+(variable) @variable
+
+; Function calls
+
+(function_call
+  function: (identifier) @function)
+
+(method_call
+  method: (selector_expression
+    field: (field_identifier) @function))
+
+; Operators
+
+"|" @operator
+":=" @operator
+
+; Builtin functions
+
+((identifier) @function.builtin
+ (#match? @function.builtin "^(and|call|html|index|slice|js|len|not|or|print|printf|println|urlquery|eq|ne|lt|ge|gt|ge)$"))
+
+; Delimiters
+
+"." @punctuation.delimiter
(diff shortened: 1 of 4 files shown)


runtime/queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,75 @@
+; Identifiers
+
+[
+  (field)
+  (field_identifier)
+] @property
```
the-mikedavis, line 6:
```suggestion
] @variable.other.member
```

runtime/queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+  (rune_literal)
+] @string
+
+(escape_sequence) @string.special
+
+[
+  (int_literal)
+  (float_literal)
+  (imaginary_literal)
+] @number
```
archseer, line 66: This should also use the proper scopes

runtime/queries/gotmpl/highlights.scm
```diff
@@ -0,0 +1,75 @@
...
+
+[
+  (true)
+  (false)
+  (nil)
+] @constant.builtin
```
archseer, line 72: Split the bools out and tag them as constant.builtin.boolean

runtime/queries/gotmpl/highlights.scm
dannylongeuay: fixed

runtime/queries/gotmpl/highlights.scm
dannylongeuay: fixed

runtime/queries/gotmpl/highlights.scm
dannylongeuay: fixed

dannylongeuay:
Thanks for the feedback. I have addressed the requested changes and comments.

archseer (approved):
Thanks!

PR #3097, opened 2022-07-18
File type name capitalized for status line

erasin:
like this: typescript-react -> TypescriptReact
WX20220718-234730@2x

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -284,9 +284,23 @@ fn render_file_type<F>(context: &mut RenderContext, write: F)
 where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
-    let file_type = context.doc.language_id().unwrap_or("text");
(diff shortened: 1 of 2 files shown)


etienne-k:
Maybe having this configurable (opt-in) might be better. Lualine, for instance, doesn't capitalize... So I could imagine there's users who want it capitalized and some who don't.

I also just realized that there's certain languages for which capitalizing (camelCase or PascalCase) wouldn't make sense, e.g. TOML:
Showing "toml" or - if caps are used - "TOML" is fine, but "Toml" would definitely be a bit weird...

Same goes with other abbreviated language names, like XML, JSON, CSS, ...

And finally, there might be some language IDs where it's not easy using different flavors for one ID: e.g. instead of "typescript-react", it should actually be "type-script-react" for the name to be correctly displayed as "TypeScriptReact", but "type-script-react" in lowercase, on the other hand, feels weird to me.

erasin:
yes, configurable is better choice.

PR #3098, opened 2022-07-18
Add mode colors (`color-modes`) to the "onedark" theme

etienne-k:
After the following themes:
```
autumn
autumn_night
meliora
night_owl
rose_pine
solarized_dark
solarized_light
```

...comes **onedark**!

Regarding the colors (blue for normal, green for insert and purple for selection), I took inspiration from lualine, which uses these colors by default.
Additionally, I find it especially intuitive to have green for insertion (which we know from diffs) and blue for the regular, default state.

I opted for a dark foreground, as a light foreground simply lacks any minimal amount of contrast. A very dark gray ("light-black") seemed to yield the best results.

_For those who didn't know yet: This wonderful coloring feature comes from #2676 and must be enabled in the configuration by setting `color-modes = true` under `[editor]`._

image
image
image


archseer:
Thanks!

PR #3103, opened 2022-07-18
Improve python highlighting

kirawi:
These changes were based on `nvim-treesitter`.

- Highlight import identifiers as `@namespace`
- Highlight boolean values as `@constant.builtin.boolean`
- Made `@keyword` captures more granular with the existing scopes. No new scopes were introduced.
- Added `@punctuation` captures.

### DIFF
--- a/runtime/queries/python/highlights.scm
+++ b/runtime/queries/python/highlights.scm
@@ -108,9 +108,11 @@
 (string) @string
 (escape_sequence) @constant.character.escape
 
+"(" ")" "[" "]" "{" "}"] @punctuation.bracket
+["," "." ":" ";" (ellipsis)] @punctuation.delimiter
 (interpolation
   "{" @punctuation.special
-  "}" @punctuation.special) @embedded
+  "}" @punctuation.special)
 
 [
   "-"
--- a/runtime/queries/python/highlights.scm
+++ b/runtime/queries/python/highlights.scm
@@ -108,11 +108,11 @@
 (string) @string
 (escape_sequence) @constant.character.escape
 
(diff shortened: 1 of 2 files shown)


runtime/queries/python/highlights.scm
```diff
@@ -49,6 +57,13 @@
 (parameters (typed_default_parameter name: (identifier) @variable.parameter))
 (keyword_argument name: (identifier) @variable.parameter)
 
+(parameters
+  (list_splat_pattern ; *args
+    (identifier) @parameter))
+(parameters
+  (dictionary_splat_pattern ; **kwargs
+    (identifier) @parameter))
```
archseer, line 65: `@variable.parameter`

ChrHorn:
Nice, I planned to go over it myself.

Could you also add the missing punctuations? 
https://github.com/nvim-treesitter/nvim-treesitter/blob/8b748a7570b89822d47ac0ed0f694efda6523c7d/queries/python/highlights.scm#L252-L258

runtime/queries/python/highlights.scm
```diff
@@ -81,22 +96,23 @@
...
 
 (integer) @constant.numeric.integer
 (float) @constant.numeric.float
 (comment) @comment
 (string) @string
 (escape_sequence) @constant.character.escape
 
+["," "." ":" ";" (ellipsis)] @punctuation.delimiter
 (interpolation
   "{" @punctuation.special
```
archseer, line 113: This seems wrong to me? We have an embedded tag specifically to match interpolation, but you removed it here.

PR #3105, opened 2022-07-19
Fix: indent guides out of bound.

erasin:
when Insufficient width of split window
WX20220719-145548@2x

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -422,12 +422,18 @@ impl EditorView {
             }
 
             for i in 0..(indent_level / tab_width as u16) {
-                surface.set_string(
-                    viewport.x + (i * tab_width as u16) - offset.col as u16,
-                    viewport.y + line,
-                    &indent_guide_char,
-                    indent_style,
-                );
+                let visual_x = i * tab_width as u16;
+                let out_of_bounds =
+                    visual_x < offset.col as u16 || visual_x >= viewport.width + offset.col as u16;
+
+                if !out_of_bounds {
+                    surface.set_string(
+                        viewport.x + visual_x - offset.col as u16,
+                        viewport.y + line,
+                        &indent_guide_char,
+                        indent_style,
+                    );
+                }
             }
         };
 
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -422,12 +422,18 @@ impl EditorView {
...
+                    visual_x < offset.col as u16 || visual_x >= viewport.width + offset.col as u16;
+
+                if !out_of_bounds {
+                    surface.set_string(
+                        viewport.x + visual_x - offset.col as u16,
+                        viewport.y + line,
+                        &indent_guide_char,
+                        indent_style,
+                    );
+                }
```
aikomastboom, line 436: I would think.. once it goes `out_of_bounds` the next indent_level can not suddenly be inside again.. or ??

```suggestion
                } else {
                  break;
                }
```

erasin: thanks , i test this, If use break, the indents of the next level will not be rendered.

helix-term/src/ui/editor.rs
```diff
@@ -416,18 +416,31 @@ impl EditorView {
...
-                    &indent_guide_char,
-                    indent_style,
-                );
+                if i < hide_indent_level {
+                    continue;
+                }
+
+                let visual_x = i * tab_width as u16;
+                let out_of_bounds =
+                    visual_x < offset.col as u16 || visual_x >= viewport.width + offset.col as u16;
```
archseer, line 432: This seems overly complicated to me, you should be able to calculate the starting indent level outside the for loop then `for i in starting_level..(indent_level / tab_width as u16)` to start rendering at the right offset.

archseer (changes requested):

helix-term/src/ui/editor.rs
aikomastboom: It's not about the starting level.. it's about stopping and not exceeding as I understand.
Possibly, once the right limit has been found, code can be further optimized to reuse this value.
As the guides are an optional feature, optimization should not bleed into normal codeflow.

archseer: It's about the starting level. The code is rendering the right split's indent guides, but since the starting column is further in the file it's rendering them off screen into the previous split.

archseer:
Addressed in https://github.com/helix-editor/helix/commit/906259cc41cc251dbbe1802e3a5e54a566a7d8d2

Issue #3072, opened 2022-07-14
Slow tree-sitter performance on certain files


kchibisov:
### Summary

Certain markdown files are very slow to edit. In particular when they have a lot of lists.
For example a `CHANGELOG.md` file from the helix repo is a enough to trigger it.

The problem is likely in grammar here, since neovim with tree-sitter works fine on it.

While this particular problem could be solved with slightly optimized, but less accurate grammar. However the issue manifests itself on any large file of any other type, just requires much larger file than ~600LoC(I've used 50K LoC C++ that I have around).

To avoid delay and input lag in those particular scenarios it would make sense to either run tree-sitter only on a part of the buffer (I'm not familiar how it works) or update it after the typing ended. Since right now you have a latency of like 2 seconds on 50K lines file of C++, which is impossible to work with.

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.05 (3cced1e3)

MDeiml:
Author of the markdown grammar here. It seems that at the moment language injection cannot be parsed incrementally. The markdown grammar uses injection a lot so it becomes quite apparent for it. This is what I get from line 846 here:
https://github.com/helix-editor/helix/blob/8681fb6d9ee06610fac2baafb4ca34f93f9c7f4d/helix-core/src/syntax.rs#L841-L848
Notice how a layer can only be reused if `ranges` did not change. But this is often the case when editing a file. The test here could probably be laxer.

I didn't look to deep, but (from my knowledge of treesitter) any layer should be fine here to "reuse" as long as none is reused twice. So something akin to
1. Choosing the layer whose ranges "match best" in some way
2. Changing those ranges to the new ones

should work.

archseer:
We actually follow tree-sitter-cli/atom logic to update all the markers according to user edits beforehand: https://github.com/helix-editor/helix/blob/8681fb6d9ee06610fac2baafb4ca34f93f9c7f4d/helix-core/src/syntax.rs#L644-L726=

So the injection layers are tracked reliably.

We also seem to only use HTML injections & code blocks: https://github.com/helix-editor/helix/blob/master/runtime/queries/markdown/injections.scm

MDeiml:
Ah ok, I guess you are still using the old version of my parser then (the `main` branch). You could try updating to the `split_parser` branch. That one has some major performance improvements, but uses injection very extensively.

archseer:
Ah, good point! I'll see if I can update us to the new branch 👍

[...]

Issue #940, opened 2021-10-31
logging: provide tree-sitter parsing and query time


matu3ba:
#### Describe your feature request

Provide very simple logging for tree-sitter on `INFO`/`-vvv`, so users can estimate expections of tree-sitter performance: 1. `parsing` and `query` time can be printed separately.
As of now, tree-sitter timings are not logged at all: `hx -vvv ir.cpp` of [ir.cpp](https://github.com/ziglang/zig/blob/master/src/stage1/ir.cpp) shows only content from `helix_lsp::transport` and `mio::poll`.

Details truncated to 125 symbols line length, because some stuff is giantic in output even on only opening files.

<details>

```txt
2021-10-30T14:04:31.296 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2021-10-30T14:04:31.296 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2021-10-30T14:04:31.296 mio::poll [TRACE] registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
2021-10-30T14:04:31.296 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"initialize","params":{"capabilities":{"text
2021-10-30T14:04:31.297 mio::poll [TRACE] registering event source with poller: token=Token(4), interests=READABLE | WRITABLE
2021-10-30T14:04:31.297 mio::poll [TRACE] registering event source with poller: token=Token(5), interests=READABLE | WRITABLE
2021-10-30T14:04:31.297 mio::poll [TRACE] registering event source with poller: token=Token(0), interests=READABLE
2021-10-30T14:04:31.297 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE
2021-10-30T14:04:31.309 helix_lsp::transport [ERROR] err <- "I[14:04:31.309] clangd version 12.0.1\n"
2021-10-30T14:04:31.309 helix_lsp::transport [ERROR] err <- "I[14:04:31.309] PID: 805639\n"
2021-10-30T14:04:31.309 helix_lsp::transport [ERROR] err <- "I[14:04:31.309] Working directory: $HOME/dev/git/zig/zig/src/sta
2021-10-30T14:04:31.309 helix_lsp::transport [ERROR] err <- "I[14:04:31.309] argv[0]: clangd\n"
[...]

PR #3108, opened 2022-07-19
Use split_parser branch for markdown grammar

MDeiml:
See #3072.

Only a draft because `markdown_inline` grammar is not injected properly, but to me it seems like this is a problem with helix and not the language config? After a few experiments I think the grammar _is_ injected, but highlighting is still not done according to the injected language. I'm not very sure though.

Another weird behavior is with fenced code blocks. They seem to work to some degree, but if you enter something like
~~~markdown
```rust
let foo = "bar";
```
~~~
then `let` is highlighted correctly while `"bar"` is not highlighted at all.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -884,7 +884,20 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "markdown"
-source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "ab15701d8f3f68aeb74e30573b7d669a6ef2a7ed" }
+source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "142a5b4a1b092b64c9f5db8f11558f9dd4009a1b", subpath = "tree-sitter-markdown" }
+
+[[language]]
+name = "markdown.inline"
+scope = "source.markdown.inline"
+injection-regex = "markdown\\.inline"
+file-types = []
+roots = []
+indent = { tab-width = 2, unit = "  " }
+grammar = "markdown_inline"
+
+[[grammar]]
+name = "markdown_inline"
+source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "142a5b4a1b092b64c9f5db8f11558f9dd4009a1b", subpath = "tree-sitter-markdown-inline" }
 
 [[language]]
 name = "dart"
--- a/languages.toml
+++ b/languages.toml
@@ -892,7 +892,6 @@ scope = "source.markdown.inline"
 injection-regex = "markdown\\.inline"
 file-types = []
 roots = []
-indent = { tab-width = 2, unit = "  " }
 grammar = "markdown_inline"
 
 [[grammar]]
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -62,6 +62,7 @@
 | lua | ✓ |  | ✓ | `lua-language-server` |
 | make | ✓ |  |  |  |
 | markdown | ✓ |  |  |  |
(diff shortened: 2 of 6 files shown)


MDeiml:
Changing this to ready since I found a solution to the problems in #3129.

the-mikedavis:
I tried this branch locally but I don't see any qualitative improvements to performance (we'll need some measurements like from #940 to be definitive). Specifically I see the some slowness with incremental updates while editing (for example typing quickly in insert mode) on a reasonably large and complicated markdown document like [the changelog](https://github.com/helix-editor/helix/blob/5551cfd0ba36b5cda7a49c78631464240213e938/CHANGELOG.md). Is this branch of the grammar(s) more correct than master? This doesn't feel like a regression so I'm certainly not against merging it but I'm curious what the improvement is of splitting the parsers?

MDeiml:
I'm no longer maintaining the master branch, I just kept it because some projects seem to depend on it.

It was not really possible to implement a correct enough markdown parser in a single grammar. The specs also recommend to parse markdown in 2 passes. Now after splitting it into 2 grammars they work closer to that recommendation.

They're more correct now and with some optimizations to incremental parsing I think they'll also be faster. (But that'll be topic of future PRs).

MDeiml:
Specifically I think that tree-sitter reparses all injected ranges if they contain "conflicts" no matter if they changed or not. That means that parsing is linear with the number of injected ranges. I think we could just not reparse ranges if they did not change at all (parsing should be deterministic with regards to the input anyways).

the-mikedavis:
I'm not very familiar with the changes but I think archseer might've covered that when implementing combined injections: https://github.com/helix-editor/helix/compare/6728e4449038e9481b72251441182d508c165a9c...7c9ebd05b83e90e55d032f65d9405ad265b82258

We don't use the stock tree-sitter-cli highlighting code since it isn't incremental

MDeiml:
Just looking at the code that doesn't seem to be the case, but I should do some more investigation.

MDeiml:
It's related to this TODO:
https://github.com/helix-editor/helix/blob/681c0a91dc36f9db1d23bd31ac1b66414d2a6099/helix-core/src/syntax.rs#L724
Just going one step further and also avoid parsing layers that are after the edits

the-mikedavis:
This works well for me locally but let's await further review/merging in #3129

archseer:
Can you resolve the conflict? Looks good to merge otherwise 👍🏻

languages.toml
```diff
@@ -868,7 +868,18 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "markdown"
-source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "ab15701d8f3f68aeb74e30573b7d669a6ef2a7ed" }
+source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "142a5b4a1b092b64c9f5db8f11558f9dd4009a1b", subpath = "tree-sitter-markdown" }
+
+[[language]]
+name = "markdown_inline"
+scope = "source.markdown_inline"
+injection-regex = "markdown_inline"
```
archseer, line 876: How about `markdown.inline`?

MDeiml: Sure. For all of those? I have to admit I don't really understand what `scope` is for.

the-mikedavis: > For all of those?

Yeah but `injection-regex` should probably be `"markdown\\.inline"` to be exact.

They're like textmate scopes. Currently we use them for some language naming things like what to send to a Language Server as `language_id`.

the-mikedavis (approved):
Works great! Thanks for working on this plus going the extra mile to add the new predicate :)

Issue #196, opened 2021-06-09
interactive global search


dofinn:
The reason I use vim is for ripgrep and fzf. This makes for extremly fast navigation. 

Request:

- Ability to index files in active directory and search via FZF
![20210609_224940_grim](https://user-images.githubusercontent.com/21171128/121357257-ff353b00-c974-11eb-88f4-2376aa6ab611.png)

- Ability to grep through fies in active directory and pipe results in fzf 
- `ripgrep file_config | fzf` <- this opens up in same preview window that FZF provides. 
![20210609_225128_grim](https://user-images.githubusercontent.com/21171128/121357583-41f71300-c975-11eb-9761-69150479e048.png)
![20210609_225136_grim](https://user-images.githubusercontent.com/21171128/121357610-46bbc700-c975-11eb-8222-d77c5003604b.png)

There are many blogs etc writing about this method in vim. EG: https://sidneyliebrand.medium.com/how-fzf-and-ripgrep-improved-my-workflow-61c7ca212861

kirawi:
Should this be a core feature of Helix or be implemented as a plugin?

pickfire:
I personally think this should be core feature. We already have like skim like thing, implementing ripgrep is just near.

dofinn:
is a dependecy on ripgrep something you want to make mandatory? I would imagine having these functions pluggable
grep=plugin(ripgrep)
find=plugin(fzf)

pickfire:
I don't think we will pull in ripgrep. We may be implementing our own which is slightly slower compared to ripgrep but work well for our usecase.

wojciechkepka:
We could also just import the `grep_searcher` crate that ripgrep uses internally. It's published to crates.io 

https://docs.rs/grep-searcher/0.1.7/grep_searcher/

dofinn:
so excuse me for being an idiot. The file picker is already aweesome! cancel the FZF request

archseer:
I do think a preview window + some way to actually search through files would be useful though :) 

I think there should be a general way to pipe data into the built-in picker.

dofinn:
I agree, after using it last night i did miss the preview window + the actual fuzzy finding where i could use multiple terms seperated by whitespace rather then the exact path/to/target/file

archseer:
The finding is fuzzy (we use `skim`'s algorithm, we just don't highlight the matched characters yet.

nrdxp:
[...]

PR #3110, opened 2022-07-19
Re-request workspace symbols on keypress in picker

### LINKED ISSUE #1437: (workspace) symbol picker is limited to the first 128 elements
After #1041 it is now possible to search for symbols in a whole workspace. Problem is, there are often a lot more symbols than the default (from where) 128 displayed symbols and the list is not updated as I enter text.

### Reproduction steps

 

1. Go to the `helix` repository
2. Open `helix-lsp/src/client.rs`
3. Open the symbol picker `Space+s`
4. There should be aroud 50 symbols (48 on dbaed0b for me)
5. Searching for `workspace_symbols` goes to line 769 for me.
3. Open the workspace symbol picker `Space+Shift+s` (`Space+S`)
4. Search for `workspace_symbols`
5. There is no result and the prompt displays `0/128`

### Environment

- Platform: Linux Ubuntu 21.10 
- Helix version: `v0.5.0-331-gdbaed0b` on `git describe` 

A llittle heavy since I used `-vvv` so I uploaded a file instead

[helix.log](https://github.com/helix-editor/helix/files/7807706/helix.log)

</details>

sudormrfbin:
Fixes #1437. Most language servers limit the number of workspace symbols that are returned with an empty query even though all symbols are supposed to be returned, according to the spec (for perfomance reasons). This patch adds a workspace symbol picker based on a dynamic picker that allows re-requesting the symbols on every keypress (i.e. when the picker query text changes). The old behavior has been completely replaced, and I have only tested with rust-analyzer so far.

| Before | After |
| ---- | ---- |
| helix-old-workspace-symbols | helix-dynamic-workspace-symbols |

Notice how in the old picker the `format_and_save` function was not being picked up. This is because the picker only filtered the list of symbols that was initially given by `rust-analyzer` which does not include all symbols. In the dynamic picker a specific query of the current prompt text is sent to the server and we get accurate results from the server.

The newly added dynamic picker can also be used for interactive global search (#196), which is next on my list :)

### TODO

- [ ] Figure out a way to do debouncing (sending a request on every keypress is expensive and might cause screen jitters now and then).
- [ ] Do not score the picker before replacing with new data, which causes old contents to be re-sorted and displayed only to be replaced by new content. Caused by reusing `Picker` inside `DynamicPicker`.


archseer:
Debounce: you could use the idle timer event that's used for completion, I can't remember if the change is there or not but I imagined it would trigger a generic event on the compositor so the components can handle idle timeout the way they want

poliorcetics:
Very much interested in this, thank you for working on it !

papertigers:
Is there anything we can do to help move this forward?
Currently I am maintaining this patch in a local branch that I rebase onto.  I have a few larger rust projects where I hit the symbol limit, making this PR really useful.

johnrichardrinehart:
@the-mikedavis @archseer @sudormrfbin Anything blocking this from a merge? cc/ @papertigers

sudormrfbin:
It requires #3172 (which has some unresolved comments) and then a proper mechanism for debouncing.

archseer:
I think just #3172 is enough on it's own for now. It'll slightly lag since we will only send the request on idle and not per keypress but that at least gets it working.

johnrichardrinehart:
I think merging this in light of #3173 makes sense. If per-key operation are excessively laggy then someone can file an issue and we can introduce a PR for debouncing separately.

This would unlock #196 🥺.

sgoodluck:
#4916 is related. 

I pulled down this PR and ran it successfully. 
Please note that nothing appears until _after_ I start typing (which may be by design?)

See asciinema below:
https://asciinema.org/a/uXMGy7yVQP4CfDgL4KTdhNCTS

PR #3119, opened 2022-07-20
helix_view::theme [WARN] Theme

aikomastboom:
Quick attempt at removing some warnings from the log.
I have no theming knowledge, so unknown colors are probably off.

### DIFF
--- a/runtime/themes/catppuccin.toml
+++ b/runtime/themes/catppuccin.toml
@@ -35,7 +35,7 @@ label = "peach"
 
 "diff.plus" = "#35bf86"
 "diff.minus" = "#f22c86"
-"diff.delta" = "#6f44f0"
+"diff.delta" = "delta"
 
 "ui.background" = { bg = "black_2" }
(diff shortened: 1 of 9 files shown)


runtime/themes/catppuccin.toml
```diff
@@ -91,3 +91,8 @@ gray_2 = "#C3BAC6"
 white = "#D9E0EE"
 lavender = "#C9CBFF"
 rosewater = "#F5E0DC"
+
+silver = "#c0c0c0"
+apricot = "#f47868"
+lightning = "#ffcd1c"
```
sudormrfbin, line 97: These three keys are from the default theme file, probably a remnant from being copied. The catppuccin [style guide](https://github.com/catppuccin/catppuccin/blob/main/docs/style-guide.md#typography) recommends using `maroon` for errors and `peach` for warnings (which is where these palette keys are used in this file). Something mild like `lavender` would be a good fit for the hint key too.

archseer (approved):
Thanks for fixing these!

PR #3129, opened 2022-07-21
Exclude only named children without injection.include-children

MDeiml:
This fixes problems arising in #3108.

The markdown parser needs to exclude children for language injection, but only named ones. This is *different* from the behavior in tree-sitter-highlight, but to me it seems more sensible. 

To give a technical example: Otherwise there is almost no way to write something like
```js
rule: $ => seq('(', repeat(choice($._content, $.excluded)), ')'),
```
where `$._content` and the parentheses should be included and `$.excluded` excluded. This is because there will always be `'('` and `')'` nodes as a child of `rule`.

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -1457,7 +1457,7 @@ fn intersect_ranges(
         };
 
         for excluded_range in node
-            .children(&mut cursor)
+            .named_children(&mut cursor)
             .filter_map(|child| {
                 if includes_children {
                     None
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -766,7 +766,7 @@ impl Syntax {
                 );
                 let mut injections = Vec::new();
                 for mat in matches {
-                    let (language_name, content_node, include_children) = injection_for_match(
+                    let (language_name, content_node, included_children) = injection_for_match(
                         &layer.config,
                         &layer.config.injections_query,
                         &mat,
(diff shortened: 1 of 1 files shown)


the-mikedavis:
If you want a node from the tree to be queryable, it should be named: that's the point of named nodes. I think you should restructure the grammar to expose the nodes you want to capture in queries.

I can't test this at the moment but I suspect this breaks injections for some templating languages or maybe Rust macros? I'll give it a look when I have a chance

MDeiml:
> If you want a node from the tree to be queryable, it should be named: that's the point of named nodes. I think you should restructure the grammar to expose the nodes you want to capture in queries.

That's my point. I expose `inline` nodes and `block_continuation` nodes. I want to capture the former while excluding the latter. I don't want to exclude unnamed nodes that I have less control over when writing the grammar.

Also notice how not excluding unnamed nodes is strictly more expressive. In a tree-sitter grammar an unnamed node can always be turned into a named one with `alias(...)`. I think this is exactly what you meant?

> I can't test this at the moment but I suspect this breaks injections for some templating languages or maybe Rust macros? I'll give it a look when I have a chance

I'm gonna have a look myself too. If that's the case and the queries can't easily be changed to work with the new behavior: how do you feel about adding a `include-unnamed` or `exclude-named` flag? I'm not sure it's possible to make my grammar work without this.

MDeiml:
It seems you're right. Many languages rely on excluding unnamed nodes for injecting the `comment` language.

the-mikedavis:
A separate property like include/exclude-unnamed sounds like a good addition to me 👍

helix-core/src/syntax.rs
```diff
@@ -1808,6 +1825,7 @@ fn injection_for_match<'a>(
     }
 
     let mut include_children = false;
+    let mut include_unnamed_children = false;
```
the-mikedavis, line 1828: Instead of two booleans, I think it would be cleaner to use an enum

MDeiml: I think a enum with tree members makes the most sense here :+1:

the-mikedavis (approved):
This looks good, thanks for working on it 👍

PR #3136, opened 2022-07-21
Avoid copying fragments

### LINKED ISSUE #3127: :pipe hangs when piping a large file through sed
### Summary

Helix freezes and becomes unresponsive when piping a large (~7600 lines) file into `sed`. Piping the same file into `sed` outside of helix doesn't produce the same behavior, and helix seems to handle piping smaller files just fine. Piping into other commands also doesn't seem to cause the same issue though I didn't try very many. 

### Reproduction Steps

 

I tried this:

1. `hx frankenstein.txt` (from [here](https://raw.githubusercontent.com/asweigart/codebreaker/master/frankenstein.txt))
2. `%`
3. `|sed s/l/w/g`

I expected this to happen:
The command to execute and all of the 'l' characters in Frankenstein to be replaced with 'w'

Instead, this happened:
Helix became unresponsive

### Helix log

```
2022-07-21T01:21:42.642 helix_view::theme [WARN] Theme: invalid style attribute: modifier
2022-07-21T01:21:42.642 helix_loader [DEBUG] Located configuration folders: []
2022-07-21T01:21:42.648 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2022-07-21T01:21:42.648 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2022-07-21T01:21:42.648 mio::poll [TRACE] registering event source with poller: token=Token(0), interests=READABLE
2022-07-21T01:21:42.648 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE

```

</details>

### Platform

Linux (Pop!_Os)

### Terminal Emulator

Alactritty v0.10.0-rc1

### Helix Version

helix 22.05 (5ed62239)

MDeiml:
~Might fix #3127.~

This changes the `Range::fragment` and `Selection::fragments` to return `RopeSlice`. This way we can avoid copying the ropes' contents in a few cases, specifically

* `shell_impl`
* Sometimes one less copy in `switch_to_*case`
* Sometimes one less copy in all commands that "yank"
* Sometimes one less copy in `sort` and `rotate_selection_contents`

Also, it will make the computational overhead of copying the fragments more apparent in places where `Range::fragment` is used (until now it almost looked like there was no copy, since a `Cow` was returned).

(I also rewrite the mouse event handling because (I think, without optimizations) until now it copied the entire file on each mouse up event)

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -222,11 +222,14 @@ impl Range {
 
     // groupAt
 
-    /// This returns a RopeSlice since returning a string might imply copying
-    /// the entire content within the range. If you really need the text in a
-    /// continuous slice, you can call Cow::<str>.from(...).
+    /// Consider using `slice` instead.
     #[inline]
-    pub fn fragment<'a, 'b: 'a>(&'a self, text: RopeSlice<'b>) -> RopeSlice<'b> {
+    pub fn fragment<'a, 'b: 'a>(&'a self, text: RopeSlice<'b>) -> Cow<'b, str> {
+        self.slice(text).into()
+    }
+
+    #[inline]
+    pub fn slice<'a, 'b: 'a>(&'a self, text: RopeSlice<'b>) -> RopeSlice<'b> {
         text.slice(self.from()..self.to())
     }
 
@@ -547,8 +550,12 @@ impl Selection {
         self.transform(|range| Range::point(range.cursor(text)))
     }
 
-    pub fn fragments<'a>(&'a self, text: RopeSlice<'a>) -> impl Iterator<Item = RopeSlice> + 'a {
+    pub fn fragments<'a>(&'a self, text: RopeSlice<'a>) -> impl Iterator<Item = Cow<str>> + 'a {
         self.ranges.iter().map(move |range| range.fragment(text))
+    }
+
+    pub fn slices<'a>(&'a self, text: RopeSlice<'a>) -> impl Iterator<Item = RopeSlice> + 'a {
+        self.ranges.iter().map(move |range| range.slice(text))
     }
 
     #[inline(always)]
@@ -614,7 +621,7 @@ pub fn keep_or_remove_matches(
 ) -> Option<Selection> {
     let result: SmallVec<_> = selection
         .iter()
-        .filter(|range| regex.is_match(&Cow::from(range.fragment(text))) ^ remove)
+        .filter(|range| regex.is_match(&range.fragment(text)) ^ remove)
         .copied()
         .collect();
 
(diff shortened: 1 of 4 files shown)


archseer:
In most cases we **do** want to use a `Cow` though. Why not add a second `slice()` method that will return a `RopeSlice` for cases where that's preferrable?

MDeiml:
It's more of an opinion, but I think it's always better to not hide computational complexity in APIs. A lot of use cases want to iterate over the fragment, which can be done without copying the contents.

Ideally we could also avoid the places where we turn the `RopeSlice` into a `Cow` since that could always end up copying the entire file. This is what happens if you press `%s` at the moment since the `regex` crates methods only take strings.

MDeiml:
Ok I'll change this to add a `chunks` method.

The regex functions are actually the only ones that need `Cow`s. Commands that yank need an owned string (which can be constructed from a `Cow`). All other commands can avoid one copy by using iterators.

helix-core/src/selection.rs
```diff
@@ -222,9 +222,15 @@ impl Range {
 
     // groupAt
 
+    /// Consider using `slice` instead.
```
the-mikedavis, line 225: This comment is a bit confusing. Could you add docs to both functions and explain the difference in return type?

helix-term/src/ui/editor.rs
```diff
@@ -1029,9 +1029,11 @@ impl EditorView {
                 if doc
                     .selection(view.id)
                     .primary()
-                    .fragment(doc.text().slice(..))
-                    .width()
-                    <= 1
+                    .slice(doc.text().slice(..))
+                    .chars()
```
archseer, line 1033: `.slice(..).len_chars()` should be enough?

MDeiml: Not really. `len_chars` doesn't take into account if the character is zero width. That's what the `width` below is for. I changed the previous implementation because it was linear with the size of the selection. I admit my change is a bit ugly though, so maybe it's worth keeping the old / slower implementation for less complexity.

helix-core/src/selection.rs
MDeiml: Added some docs, are they ok?

the-mikedavis: yep looks good to me 👍

helix-term/src/ui/editor.rs
archseer: We always display zero width characters as at least of width 1 so that they are editable: https://github.com/helix-editor/helix/blob/aa4394ce1291b0af0b6da5e54cfa60a4c13f7aab/helix-core/src/graphemes.rs#L10-L32

MDeiml: Ah, ok. The code previously used the `unicode_width` crate so I thought that was necessary. But it seems we can kill two birds with one stone here :)

archseer (approved):
Thanks for working on this! 🎉

PR #3137, opened 2022-07-21
Enable CI builds for `aarch64-macos`

### LINKED ISSUE #1647: Apple Silicon arm64 release

Please provide binary for macOS arm64, likes provided for Linux arm64.
Thanks.

yvt:
Closes #1647.

A preview CI run based on this patch and the built artifacts can be found at <https://github.com/yvt/helix/actions/runs/2709143241>.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -116,7 +116,7 @@ jobs:
           args: --release --locked --target ${{ matrix.target }}
 
       - name: Strip release binary (linux and macos)
-        if: matrix.build == 'x86_64-linux' || matrix.build == 'x86_64-macos' || matrix.build == 'aarch64-macos'
+        if: matrix.build == 'x86_64-linux' || endsWith(matrix.build, 'macos')
         run: strip "target/${{ matrix.target }}/release/hx"
 
       - name: Strip release binary (arm)
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -64,10 +64,12 @@ jobs:
           rust: stable
           target: x86_64-pc-windows-msvc
           cross: false
-        # - build: aarch64-macos
-        #   os: macos-latest
-        #   rust: stable
-        #   target: aarch64-apple-darwin
+        - build: aarch64-macos
+          os: macos-latest
+          rust: stable
+          target: aarch64-apple-darwin
+          cross: false
+          skip_tests: true  # x86_64 host can't run aarch64 code
         # - build: x86_64-win-gnu
         #   os: windows-2019
         #   rust: stable-x86_64-gnu
@@ -100,6 +102,7 @@ jobs:
 
       - name: Run cargo test
         uses: actions-rs/cargo@v1
+        if: "!matrix.skip_tests"
         with:
           use-cross: ${{ matrix.cross }}
           command: test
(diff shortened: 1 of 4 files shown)


archseer:
Do we need to do anything special to support this in the homebrew tap? https://github.com/helix-editor/homebrew-helix/blob/master/Formula/helix.rb

archseer:
Looks like we need a conditional: https://github.com/golangci/homebrew-tap/blob/bb0e7c6fa14d7fd771e1a5941f5e3ca85a15624c/Formula/golangci-lint.rb#L11-L40=

yvt:
I believe switching `url` and `sha256` by CPU architectures is all we need to do.

```ruby
on_macos do
  if Hardware::CPU.arm?
    url "https://github.com/helix-editor/helix/releases/download/#{version}/helix-#{version}-aarch64-macos.tar.xz"
    sha256 "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
  end
  if Hardware::CPU.intel?
    url "https://github.com/helix-editor/helix/releases/download/#{version}/helix-#{version}-x86_64-macos.tar.xz"
    sha256 "XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
  end
end
```

.github/workflows/release.yml
```diff
@@ -113,7 +116,7 @@ jobs:
           args: --release --locked --target ${{ matrix.target }}
 
       - name: Strip release binary (linux and macos)
-        if: matrix.build == 'x86_64-linux' || matrix.build == 'x86_64-macos'
+        if: matrix.build == 'x86_64-linux' || matrix.build == 'x86_64-macos' || matrix.build == 'aarch64-macos'
```
the-mikedavis, line 119:
```suggestion
        if: matrix.build == 'x86_64-linux' || endsWith(matrix.build, 'macos')
```

helix-loader/src/grammar.rs
```diff
@@ -311,13 +320,14 @@ fn build_tree_sitter_library(src_path: &Path, grammar: GrammarConfiguration) ->
         .opt_level(3)
         .cargo_metadata(false)
         .host(BUILD_TARGET)
-        .target(BUILD_TARGET);
+        .target(target.unwrap_or(BUILD_TARGET));
```
the-mikedavis, line 323: When would `target` be different from `BUILD_TARGET`? Both come from `TARGET` environment variables in build scripts, `BUILD_TARGET` from the `helix-loader` crate's build script and `target` from `helix-term`'s

yvt: When `helix_loader::build_grammars` is being used for cross compilation.

`BUILD_TARGET` represents the target on which the code of `helix-loader` runs. `helix-term`'s build script (which has `helix-loader` as build-dependency) needs to use `helix-loader` to cross-compile for `helix-term`'s target, which may be different from `BUILD_TARGET`.

E.g., when cross-compiling `helix-term` to an `aarch64` target on an `x86_64` host, two binaries of `helix-loader` are built:

- One binary for `aarch64` target for use by `helix-term`'s main crate (normal dependency)
    - `helix-loader`'s build script runs with `TARGET = "aarch64"`
    - Hence, `env!("BUILD_TARGET") = "aarch64"`
    - `helix-term`'s main crate code passes `target = None`  because it wants to compile grammars for the current host = `helix-loader`'s build target

- One binary for `x86_64` target for use by `helix-term`'s build script (build dependency)
    - `helix-loader`'s build script runs with `TARGET = "x86_64"`
    - Hence, `env!("BUILD_TARGET") = "x86_64"`
    - `helix-term`'s build script passes `target = "aarch64"`  because it wants to cross-compile grammars for `helix-term`'s build target, not the host on which the build script runs

the-mikedavis (approved):
Looks good to me then :)

I tried out your example run artifacts on an aarch64-macos machine and it all works great 🎉

Thanks for working on this

archseer (approved):
Thanks @yvt!

PR #3138, opened 2022-07-21
Use OR of all selections in search_selection command

### LINKED ISSUE #2312: Search for an OR of all word selections when using *
Hi, this is probably the feature I miss most from Kakoune, I think it's super useful.
I'm not sure if I've expressed it well enough in the title, so let me try to elaborate with an example.

Let's say we have the following file, with the cursor placed at the beginning (the letter `a`):
```
abc
def
```
If I press `Ce*` in Helix, the search register will be set to `def`.

If I do the same thing in Kakoune, the search register will contain `\babc\b|\bdef\b`.

I would like to have the same behavior in Helix.

MilanVasko:
Closes #2312

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1748,10 +1748,16 @@ fn extend_search_prev(cx: &mut Context) {
 fn search_selection(cx: &mut Context) {
     let (view, doc) = current!(cx.editor);
     let contents = doc.text().slice(..);
-    let query = doc.selection(view.id).primary().fragment(contents);
-    let regex = regex::escape(&query);
+
+    let regex = doc
+        .selection(view.id)
+        .iter()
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -1747,10 +1747,17 @@ fn extend_search_prev(cx: &mut Context) {
...
-    let regex = regex::escape(&query);
+
+    let mut regex = String::new();
+    for (index, selection) in doc.selection(view.id).iter().enumerate() {
+        if index > 0 {
+            regex += "|";
+        }
+        let query = selection.fragment(contents);
+        regex += &regex::escape(&query);
+    }
```
the-mikedavis, line 1758: I think something like this should work instead:

```rust
    let regex = doc.selection(view.id).iter().map(|sel| regex::escape(selection.fragment(contents))).join("|");
```

PR #3140, opened 2022-07-21
Add LSP workspace command picker

MDeiml:
LSPs can communicate that they can run additional commands via the `executeCommandProvider`. This adds a picker for these commands. To use it you need to map it to e.g. `<space>W`.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -611,34 +611,6 @@ impl ui::menu::Item for lsp::Command {
     }
 }
 
-pub fn workspace_command_picker(cx: &mut Context) {
-    let (_, doc) = current!(cx.editor);
-
-    let language_server = language_server!(cx.editor, doc);
-
-    let options = match &language_server.capabilities().execute_command_provider {
-        Some(options) => options,
-        None => return,
-    };
-    let commands = options
-        .commands
-        .iter()
-        .map(|command| lsp::Command {
-            title: command.clone(),
-            command: command.clone(),
-            arguments: None,
-        })
-        .collect::<Vec<_>>();
-    cx.callback = Some(Box::new(
-        |compositor: &mut Compositor, _cx: &mut compositor::Context| {
-            let picker = ui::Picker::new(commands, (), |cx, command, _action| {
-                execute_lsp_command(cx.editor, command.clone());
-            });
-            compositor.push(Box::new(overlayed(picker)))
-        },
-    ));
-}
-
 pub fn execute_lsp_command(editor: &mut Editor, cmd: lsp::Command) {
     let doc = doc!(editor);
     let language_server = language_server!(editor, doc);
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -603,6 +603,42 @@ pub fn code_action(cx: &mut Context) {
         },
     )
 }
+
+impl ui::menu::Item for lsp::Command {
+    type Data = ();
+    fn label(&self, _data: &Self::Data) -> Spans {
+        self.title.as_str().into()
+    }
+}
+
+pub fn workspace_command_picker(cx: &mut Context) {
+    let (_, doc) = current!(cx.editor);
+
+    let language_server = language_server!(cx.editor, doc);
+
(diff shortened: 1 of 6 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -497,6 +497,44 @@ pub fn code_action(cx: &mut Context) {
...
+    let commands = execute_command_provider
+        .commands
+        .iter()
+        .map(|command| lsp::Command {
+            title: command.clone(),
+            command: command.clone(),
+            arguments: None,
+        })
+        .collect::<Vec<_>>();
+    log::debug!("outside {:?}", commands);
```
archseer, line 526: Can you get rid of the debug logs?

MDeiml: Oops sry

helix-term/src/commands/lsp.rs
```diff
@@ -496,6 +496,42 @@ pub fn code_action(cx: &mut Context) {
...
+    }
+}
+
+pub fn workspace_command_picker(cx: &mut Context) {
+    let (_, doc) = current!(cx.editor);
+
+    let language_server = language_server!(cx.editor, doc);
+
+    let execute_command_provider = match &language_server.capabilities().execute_command_provider {
+        Some(p) => p,
```
the-mikedavis, line 513: Very minor nit but I think this would be clearer as `options` since the type is `ExecuteCommandOptions`

helix-term/src/commands/lsp.rs
```diff
@@ -603,6 +603,42 @@ pub fn code_action(cx: &mut Context) {
...
+        .iter()
+        .map(|command| lsp::Command {
+            title: command.clone(),
+            command: command.clone(),
+            arguments: None,
+        })
+        .collect::<Vec<_>>();
+    cx.callback = Some(Box::new(
+        move |compositor: &mut Compositor, _cx: &mut compositor::Context| {
+            let picker = ui::Picker::new(commands, (), move |cx, command, _action| {
```
the-mikedavis, line 634: I think both of these `move`s can be removed

the-mikedavis:
For a default keybind for this, I think `<space>c` could make sense.

We should find a default keybind for this before merging - it looks like there's a bug with the command palette where it can't create any new pickers (for example running the jumplist picker command from the command palette)

MDeiml:
I agree with `<space>c` as a mnemonic for "command". I'll add the keybinding. But ultimately I'm not 100% sure this should have a default keybinding, as only very few LSP actually use this, so most users won't need it. I guess it can't hurt though.

sudormrfbin:
I think a typable command would be a better fit here. It's not as widely used as the other LSP features like `goto-*`, and I would consider the space layer of keybinds to be precious real estate that we will run out off too fast if we're not careful. Making it a typable commands also makes it easy to remap in the config.

the-mikedavis (approved):

dariooddenino:
It would be cool if the command could accept the command as an optional argument without opening the picker.

MDeiml:
As suggested I added the command name as an optional argument. This should make it really useful for adding a custom keybinding to run a specific command.

hovsater:
I just tried this out using the `typescript-language-server` and it looks like the workspace actions are NOOP? For instance, running `Organize Imports` does nothing in Helix while it does correctly sort the imports when I run the same command in Sublime Text or VSCode.

lazytanuki:
> I just tried this out using the `typescript-language-server` and it looks like some workspace actions are NOOP. For instance, running `Organize Imports` does nothing in Helix while it does correctly sort the imports when I run the same command in Sublime Text or VSCode.

Maybe this LSP server modifies the file outside of the editor. What happens when you reload the file in helix ?

hovsater:
> > I just tried this out using the `typescript-language-server` and it looks like some workspace actions are NOOP. For instance, running `Organize Imports` does nothing in Helix while it does correctly sort the imports when I run the same command in Sublime Text or VSCode.
> 
> Maybe this LSP server modifies the file outside of the editor. What happens when you reload the file in helix ?

Yep. That seem to be case. Reloading the file manually works. Not ideal. But it works.

PR #3141, opened 2022-07-22
build(nix): add a way to override what grammars get built

yusdacra:
Adds a way change which grammars get built and included by adding an `override` function to the helix package that takes an `includeGrammarIf` attribute which is a function that decided whether the grammar should be included (`true`) or not (`false`). This also avoid fetching all of the grammars if you don't need them (like me) which speeds up fetching stuff when you are fetching when the sources aren't cached already. I also added `makeWrapperArgs` to the `override` function so it can be set there, I think @nrdxp asked about this.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -18,83 +18,115 @@
     nixpkgs,
     nixCargoIntegration,
     ...
-  }:
-    nixCargoIntegration.lib.makeOutputs {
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
This is great, thanks!

Issue #3148, opened 2022-07-22
Panic when running `vsplit: filename`


Woolworths:
### Summary

`cd` into my dotfiles directory.

Running `:vs README.md` panics and crashes helix.

### Reproduction Steps

 

I tried this:

1. `cd dotfiles`
2. `hx`
3. `:vs README.md`

I expected this to happen:
File opens in a vertical split.

Instead, this happened:
Helix panics.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

2022-07-22T22:59:31.931 helix_view::editor [ERROR] Failed to initialize the LSP for `source.md` { LSP not defined }
2022-07-22T23:00:08.416 helix_view::editor [ERROR] Failed to initialize the LSP for `source.md` { LSP not defined }
2022-07-22T23:02:42.602 helix_view::editor [ERROR] Failed to initialize the LSP for `source.md` { LSP not defined }

</details>

`RUST_BACKTRACE=1 hx`:
```
thread 'main' panicked at 'no entry found for key', helix-view/src/document.rs:956:10
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::panicking::panic_display
   3: core::panicking::panic_str
   4: core::option::expect_failed
   5: helix_view::view::View::offset_coords_to_in_view
   6: helix_view::editor::Editor::_refresh
   7: helix_view::editor::Editor::open
   8: helix_term::commands::typed::vsplit
   9: helix_term::commands::typed::command_mode::{{closure}}
  10: <helix_term::ui::prompt::Prompt as helix_term::compositor::Component>::handle_event
  11: helix_term::compositor::Compositor::handle_event
  12: helix_term::application::Application::handle_terminal_events
  13: helix_term::application::Application::event_loop_until_idle::{{closure}}
  14: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  15: std::thread::local::LocalKey<T>::with
  16: tokio::park::thread::CachedParkThread::block_on
  17: tokio::runtime::thread_pool::ThreadPool::block_on
  18: tokio::runtime::Runtime::block_on
  19: hx::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

### Platform

macOS

### Terminal Emulator

Alacritty

### Helix Version

[...]

PR #3160, opened 2022-07-23
Fix different document panic

AceofSpades5757:
Issue #3148 

Would panic when given the view for the current document for a different document. This should fix the panic while keeping the jump history when splitting for the same file.


archseer (approved):
Thanks for the quick fix!

PR #3163, opened 2022-07-23
Make gutters padding automatic

### LINKED ISSUE #3111: Padding is now using gutter's color
### Summary

After e8214fb1e6009d44a4b7094c6e5297a19f18f549 the padding is now colored, which looks visually awful, since it's not really a padding anymore given that it closely visually attaches to line number.

An example of how it looks.

picture

The solution would be to use either a background color or provide a distinct color for it(using background color likely makes the most sense).

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.05 (791bf7e5)

pickfire:
Remove padding gutter type, and automatically add 1 padding if gutters is
non-empty.

Fix #3111

cc @kchibisov hopefully this solves your issue

Sad to see a breaking change but it should be fine given that we haven't do a release yet

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -38,7 +38,7 @@ hidden = false
 | `shell` | Shell to use when running external commands. | Unix: `["sh", "-c"]`<br/>Windows: `["cmd", "/C"]` |
 | `line-number` | Line number display: `absolute` simply shows each line's number, while `relative` shows the distance from the current line. When unfocused or in insert mode, `relative` will still show absolute line numbers. | `absolute` |
 | `cursorline` | Highlight all lines with a cursor. | `false` |
-| `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `padding`, note that `diagnostics` also includes other features like breakpoints | `["diagnostics", "line-numbers", "padding"]` |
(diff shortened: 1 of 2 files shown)


helix-view/src/editor.rs
```diff
@@ -332,8 +332,6 @@ pub enum GutterType {
     Diagnostics,
     /// Show line numbers
     LineNumbers,
-    /// Show one blank space
```
archseer, line 335: Please keep the element because it's still useful if you want to add padding between diagnostics and line number for example.

pickfire: If I add it back then it no longer fix the issue, I think probably a separate fix is needed, but for now I think it is fine.

pickfire:
> Like the problem that it's not like padding isn't there, but that it's using gutters color, but before it was using a backround color.

Yes, for the extra `"padding"` now it uses gutter color which is correct (in case you pad it between line-numbers it won't look weird), but the 1-width padding we added by default does not use that gutter style, so it should be fine.

archseer:
You forgot to remove the padding component from the default: https://github.com/helix-editor/helix/blob/dfc31e74af1b96d9097496f1faa38710ecd30580/helix-view/src/editor.rs#L476=

archseer:
Fixed in https://github.com/helix-editor/helix/commit/255c1734cfc46964e8faa21756ede2badd7b3813

Issue #3106, opened 2022-07-19
Elements in the status line are not separated, leading to a very cluttered result


poliorcetics:
### Summary

![Screenshot 2022-07-19 at 12 17 44](https://user-images.githubusercontent.com/7951708/179727672-fcc0b4ab-21d1-45f2-a47f-9a03d8cadc05.png)

My status line config to reproduce is:

```
[editor.statusline]
left = ["mode", "spinner", "diagnostics", "file-name"]
```

### Platform

macOS

### Terminal Emulator

kitty 0.25.2

### Helix Version

https://github.com/helix-editor/helix/commit/791bf7e50a19bcf7612788deb7514847089cb976

archseer:
I think we could use a custom spacer element maybe, or some type of config? With a simple, single background color statusline the extra padding introduces too much space.

the-mikedavis:
Fixed by https://github.com/helix-editor/helix/pull/3165

PR #3165, opened 2022-07-23
add spacer element to statusline

sbromberger:
Addresses #3106 and associated other comments. Multiple `"spacer"` elements may be specified in each statusline section.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -78,6 +78,7 @@ The following elements can be configured:
 | `diagnostics` | The number of warnings and/or errors |
 | `selections` | The number of active selections |
 | `position` | The cursor position |
+| `spacer` | Inserts a space between elements (multiple/contiguous spacers may be specified) |
 
 ### `[editor.lsp]` Section
 
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -246,6 +246,9 @@ pub enum StatusLineElement {
 
     /// The cursor position
     Position,
+
+    /// A single space
+    Spacer,
 }
 
 // Cursor shape is read and used on every rendered frame and so needs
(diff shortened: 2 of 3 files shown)


A-Walrus:
This would probably need to be added to the documentation (`book/src/configuration.md:80`)

sbromberger:
> This would probably need to be added to the documentation (`book/src/configuration.md:80`)

Of course. Sorry for the oversight. Will update.

sbromberger:
Ready for review.

AceofSpades5757:
Currently, gutters uses `"padding"`, would that not be more appropriate terminology compared to `"spacing"`?

[Helix Master Docs](https://docs.helix-editor.com/master/configuration.html?highlight=padding#editor-section)

helix-view/src/editor.rs
```diff
@@ -246,6 +246,9 @@ pub enum StatusLineElement {
 
     /// The cursor position
     Position,
+
+    /// A single space
+    Spacer,
```
AceofSpades5757, line 251: `GutterType` uses `Padding` terminology.
`helix-view/src/editor.rs` on line 330.

book/src/configuration.md
```diff
@@ -78,6 +78,7 @@ The following elements can be configured:
 | `diagnostics` | The number of warnings and/or errors |
 | `selections` | The number of active selections |
 | `position` | The cursor position |
+| `spacer` | Inserts a space between elements (multiple/contiguous spacers may be specified) |
```
AceofSpades5757, line 81: Gutters uses `"padding"`.

`/book/src/configuration.md` on line 41.

```
| `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `padding`, note that `diagnostics` also includes other features like breakpoints | `["diagnostics", "line-numbers", "padding"]` |
```

sbromberger:
To me, "padding" implies inserting spaces at the front (or back) of strings to align columns. That's not what these do, but I'm happy to change the name if there's a consensus.

helix-view/src/editor.rs
archseer: Gutters should also use the spacer terminology, I'll rename there

PR #3169, opened 2022-07-23
Add custom event type replacing crossterm's Event

sudormrfbin:
Add custom event type replacing crossterm's Event

Ported over from 61365dfbf3 in the `gui` branch. This will allow adding our own events, most notably an idle timer event (useful for adding debounced input in [dynamic pickers][1] used by interactive global search and workspace symbols).

[1]: https://github.com/helix-editor/helix/pull/3110

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -12,7 +12,7 @@ use serde_json::json;
 use crate::{
     args::Args,
     commands::apply_workspace_edit,
-    compositor::Compositor,
+    compositor::{Compositor, Event},
     config::Config,
     job::Jobs,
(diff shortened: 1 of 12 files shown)


groves:
This could be nice for https://github.com/helix-editor/helix/pull/3233. I'm doing some gyrations in https://github.com/crossterm-rs/crossterm/pull/693 to keep Copy on its event type. That change could be simpler if Copy could be dropped in crossterm, and having a custom event type in Helix would make that easier.

archseer (approved):

archseer:
Can you resolve the merge conflict here?

PR #3172, opened 2022-07-24
Propagate idle timeout event to components

sudormrfbin:
Depends on https://github.com/helix-editor/helix/pull/3169.

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -14,6 +14,7 @@ pub enum Event {
     Mouse(MouseEvent),
     Paste(String),
     Resize(u16, u16),
+    IdleTimeout,
 }
 
 #[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -435,18 +435,13 @@ impl Application {
     }
 
     pub fn handle_idle_timeout(&mut self) {
-        use crate::compositor::EventResult;
-        let editor_view = self
-            .compositor
-            .find::<ui::EditorView>()
-            .expect("expected at least one EditorView");
-
         let mut cx = crate::compositor::Context {
             editor: &mut self.editor,
             jobs: &mut self.jobs,
             scroll: None,
         };
-        if let EventResult::Consumed(_) = editor_view.handle_idle_timeout(&mut cx) {
+        let should_render = self.compositor.handle_event(&Event::IdleTimeout, &mut cx);
+        if should_render {
             self.render();
         }
     }
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1016,7 +1016,7 @@ impl EditorView {
         editor.clear_idle_timer(); // don't retrigger
     }
 
-    pub fn handle_idle_timeout(&mut self, cx: &mut crate::compositor::Context) -> EventResult {
+    pub fn handle_idle_timeout(&mut self, cx: &mut commands::Context) -> EventResult {
         if self.completion.is_some()
             || cx.editor.mode != Mode::Insert
             || !cx.editor.config().auto_completion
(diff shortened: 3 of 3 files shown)


helix-view/src/input.rs
```diff
@@ -4,8 +4,54 @@ use helix_core::unicode::{segmentation::UnicodeSegmentation, width::UnicodeWidth
...
-use crate::keyboard::{KeyCode, KeyModifiers};
+pub use crate::keyboard::{KeyCode, KeyModifiers};
+
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum Event {
+    Key(KeyEvent),
+    Mouse(MouseEvent),
+    Resize(u16, u16),
+    IdleTimeout,
+}
```
archseer, line 15: I think this would make more sense as a nested type:

```rust
enum Event {
   Input(InputEvent),
   IdleTimeout,
}
```

helix-view/src/input.rs
```diff
@@ -4,8 +4,54 @@ use helix_core::unicode::{segmentation::UnicodeSegmentation, width::UnicodeWidth
...
+
+/// Represents a mouse button.
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum MouseButton {
+    /// Left mouse button.
+    Left,
+    /// Right mouse button.
+    Right,
+    /// Middle mouse button.
+    Middle,
```
pickfire, line 53: Should we also support the 4th and 5th button which is quite common in gaming mouse too?

helix-view/src/input.rs
```diff
@@ -4,8 +4,54 @@ use helix_core::unicode::{segmentation::UnicodeSegmentation, width::UnicodeWidth
...
 
-use crate::keyboard::{KeyCode, KeyModifiers};
+pub use crate::keyboard::{KeyCode, KeyModifiers};
+
+#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum Event {
+    Key(KeyEvent),
+    Mouse(MouseEvent),
+    Resize(u16, u16),
+    IdleTimeout,
```
pickfire, line 14: Is it good to only have a single idle timeout? Not sure if it would be useful to have two types, one long (~1s) and one short (~100ms).

helix-view/src/input.rs
groves: Crossterm only supports the same three buttons: https://github.com/crossterm-rs/crossterm/blob/master/src/event.rs#L578. Can't add them here without support in crossterm.

It does sound like recent versions of XTerm support more mouse buttons: https://terminalguide.namepad.de/mouse/ Might be something that could be added to crossterm from that.

helix-view/src/input.rs
sudormrfbin: I didn't nest the type to avoid noise in the change set, and seems like `FocusGained` and `FocusLost` have also been introduced as new Event types. Should we still nest, and if so, should the focus events be moved out of `Event::Input` and into a top level `Event::Focus(FocusEvent)` or simply leave them in `InputEvent` and only factor out `IdleTimeout`?

helix-view/src/input.rs
sudormrfbin: We definitely do need multiple idle timers, for example one for completion and other for debouncing inputs. I was hoping to tackle it in a separate PR.

pickfire (approved):
Looks good to me.

johnrichardrinehart:
@sudormrfbin @pickfire just an onlooker, here: Can we merge?

archseer:
@johnrichardrinehart Please no pings :) There is still an unresolved comment of mine, I'll take care of it then we can merge.

helix-view/src/input.rs
archseer: Hmm okay, let's leave it as is for now

johnrichardrinehart:
Thanks a lot!

PR #3175, opened 2022-07-24
add configurable / theme-able statusline separator string

sbromberger:
This PR adds a configurable and themable statusline element called "separator" that can be used to provide visual separation of elements in the statusline. Like the spacer element in #3165, multiple / consecutive separators may be specified.

The PR adds a "separator" configuration key to `[editor.statusline]`, a "separator" element to be used in `left`, `right`, and `center`, and a `ui.statusline.separator` key for use by themes.


helix-view/src/editor.rs
```diff
@@ -213,6 +214,7 @@ impl Default for StatusLineConfig {
             left: vec![E::Mode, E::Spinner, E::FileName],
             center: vec![],
             right: vec![E::Diagnostics, E::Selections, E::Position, E::FileEncoding],
+            separator: String::new(),
```
archseer, line 217: Should default to `|` or better yet `│` (the box drawing character)

sbromberger: Can do.

luetage:
@sbromberger I have a hard time understanding why the separator is theme‐able.

* I have found no themes which are applying special colors to the separator
* While the separator foreground/background colors can be adjusted, this only seems to work for the active statusline. The inactive statusline will show the separator colors of the active statusline, which looks buggy.

It only adds complication.

sbromberger:
@luetage this PR was merged ~5 months ago. I would suggest that any objection/reversion be proposed in a separate issue/PR.

luetage:
Alright, fair enough. Gonna take a look at it.

Issue #1633, opened 2022-02-08
Adding confirmation message on save


premell:
When saving there is no confirmation that it actually succeeded: 
Helix: 
![image](https://user-images.githubusercontent.com/65544203/153033770-60f4ff4e-2702-43ca-9fbf-18dd84df0186.png)
nvim:
![image](https://user-images.githubusercontent.com/65544203/153034270-29e77c4c-063e-4ba5-8ada-b446c86703e3.png)

PS: This is such a small thing that I wasnt sure if I should even make an issue. However it might be a good issue, also there might be more places where messages would be nice

EricCrosson:
I ran into this a while back, was confused when I tried saving a read-only file. Everything looked good (or at least the same), there was no error (which probably deserves a second issue) but it was confusing when I checked `git status` and found no modifications.

I only learned that I had no write permissions when I tried making the same edit in vim and saw the error on save.

sudormrfbin:
> I ran into this a while back, was confused when I tried saving a read-only file. Everything looked good (or at least the same), there was no error (which probably deserves a separate issue) but it was confusing when I checked `git status` and found no modifications.
> 
> I only learned that I had no write permissions when I tried making the same edit in vim and saw the error on save.

Tracked in #1575.

PR #3178, opened 2022-07-24
Autosave all when the terminal loses focus

groves:
This listens to focus change events from the terminal and writes all files when focus is lost. It's not ready to merge, but I had a couple questions that it'd be good to answer before proceeding.

1. Should this be behind an option? I'd certainly always use it, but I could see others who only want to save when they explicitly request it.
  a. If it should, what should the option be? After this, I was thinking about adding behavior to autosave on buffers being hidden. If we add that second autosave behavior, I think an option set like the `gutters` option could make sense. For this, we'd have an option `autosave` with a default value of `["focus-lost"]` and an eventual default of `["focus-lost", "buffer-hidden"]`.
2. Do we need to do something about terminals that don't support this? It's supported by most modern terminals (Kitty, Alacritty, wezterm, iTerm 2, etc), tmux, [all the Linux terminals in this guide](https://terminalguide.namepad.de/mode/p1004/), and should be good in Windows Terminal once I finish up with the crossterm PR. I believe some older terms don't support it though. I don't think there's a way to detect support. Should I do anything other than document known working terminals?

## Before merging
- [x] Merge https://github.com/crossterm-rs/crossterm/pull/689 and depend on a released crossterm
- [x] Document behavior and known supported terminals

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -49,6 +49,7 @@ on unix operating systems.
 | `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "line-numbers"]` |
 | `auto-completion` | Enable automatic pop up of auto-completion. | `true` |
 | `auto-format` | Enable automatic formatting on save. | `true` |
+| `auto-save` | Enable automatic saving on focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal. | `false` |
 | `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. Used for autocompletion, set to 0 for instant. | `400` |
 | `completion-trigger-len` | The min-length of word under cursor to trigger autocompletion | `2` |
 | `auto-info` | Whether to display infoboxes | `true` |
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -559,17 +559,11 @@ pub(super) fn buffers_remaining_impl(editor: &mut Editor) -> anyhow::Result<()> 
     Ok(())
 }
 
-fn write_all_impl(
+pub fn write_all_impl(
     cx: &mut compositor::Context,
-    _args: &[Cow<str>],
-    event: PromptEvent,
-    quit: bool,
     force: bool,
+    write_scratch: bool,
 ) -> anyhow::Result<()> {
-    if event != PromptEvent::Validate {
-        return Ok(());
-    }
-
     let mut errors: Vec<&'static str> = Vec::new();
     let auto_format = cx.editor.config().auto_format;
     let jobs = &mut cx.jobs;
@@ -580,12 +574,13 @@ fn write_all_impl(
         .documents
         .values()
         .filter_map(|doc| {
-            if doc.path().is_none() {
-                errors.push("cannot write a buffer without a filename\n");
+            if !doc.is_modified() {
                 return None;
             }
-
-            if !doc.is_modified() {
+            if doc.path().is_none() {
+                if write_scratch {
+                    errors.push("cannot write a buffer without a filename\n");
+                }
                 return None;
             }
 
@@ -611,20 +606,6 @@ fn write_all_impl(
         cx.editor.save::<PathBuf>(id, None, force)?;
     }
 
-    if quit {
-        cx.block_try_flush_writes()?;
-
-        if !force {
-            buffers_remaining_impl(cx.editor)?;
-        }
-
-        // close all views
-        let views: Vec<_> = cx.editor.tree.views().map(|(view, _)| view.id).collect();
-        for view_id in views {
-            cx.editor.close(view_id);
-        }
-    }
-
     if !errors.is_empty() && !force {
         bail!("{:?}", errors);
     }
@@ -634,38 +615,38 @@ fn write_all_impl(
 
 fn write_all(
     cx: &mut compositor::Context,
-    args: &[Cow<str>],
+    _args: &[Cow<str>],
     event: PromptEvent,
 ) -> anyhow::Result<()> {
     if event != PromptEvent::Validate {
         return Ok(());
     }
 
-    write_all_impl(cx, args, event, false, false)
+    write_all_impl(cx, false, true)
 }
 
 fn write_all_quit(
     cx: &mut compositor::Context,
-    args: &[Cow<str>],
+    _args: &[Cow<str>],
     event: PromptEvent,
 ) -> anyhow::Result<()> {
     if event != PromptEvent::Validate {
         return Ok(());
     }
-
-    write_all_impl(cx, args, event, true, false)
+    write_all_impl(cx, false, true)?;
+    quit_all_impl(cx.editor, false)
 }
 
 fn force_write_all_quit(
     cx: &mut compositor::Context,
-    args: &[Cow<str>],
+    _args: &[Cow<str>],
     event: PromptEvent,
 ) -> anyhow::Result<()> {
     if event != PromptEvent::Validate {
         return Ok(());
     }
-
-    write_all_impl(cx, args, event, true, true)
+    let _ = write_all_impl(cx, true, true);
+    quit_all_impl(cx.editor, true)
 }
 
 fn quit_all_impl(editor: &mut Editor, force: bool) -> anyhow::Result<()> {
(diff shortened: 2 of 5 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -1233,6 +1234,13 @@ impl Component for EditorView {
             }
 
             Event::Mouse(event) => self.handle_mouse_event(event, &mut cx),
+            Event::Focus(FocusEventKind::Gained) => EventResult::Ignored(None),
+            Event::Focus(FocusEventKind::Lost) => {
+                if let Err(e) = commands::typed::write_all(context, &[], PromptEvent::Validate) {
```
groves, line 1239: It seems like I should call `MappableCommand.execute` instead of recreating a bit of its implementation here and making `write_all` public. I didn't see a good way to do that for `TypableCommand` like there is for static commands. Am I missing something?

sbromberger:
> Should this be behind an option?

Absolutely yes. I don't want any file writes in the event I get a focus change event. What if I'm in the middle of a refactor and have to switch away to look at documentation? Now my code is in a half-baked / unknown state.

Also: how does this work on terminals on mobile devices (I'm currently using blink.sh on ipad)? Does it change if I'm running inside tmux or using mosh?

groves:
> Also: how does this work on terminals on mobile devices (I'm currently using blink.sh on ipad)? Does it change if I'm running inside tmux or using mosh?

I believe for that to work you'd need all of Blink, tmux, and mosh to support focus events. Blink because it's the one that can tell if you've changed focus at the OS level, tmux because it can tell if you've changed focus between its panes and relay the focus events from the surrounding terminal emulator, and mosh because it'd need to convey it over the wire.

tmux has a [focus-events option](https://github.com/tmux/tmux/blob/06869ff22fa9891c9633ce3e3efa77cac758b520/options-table.c#L308). Looks like if you have that on, it supports these.

[mosh parses the focus event mode](https://github.com/mobile-shell/mosh/blob/0982212ca77cf068725a0a352e50fc017a768a85/src/terminal/terminalfunctions.cc#L298), so I'm guessing it supports them.

I can't see anything in Blink's repo mentioning focus events, so I'm guessing it wouldn't work there. If you checkout [my emit_focus_events](https://github.com/groves/crossterm/tree/emit_focus_events) branch for crossterm and run `cargo run --example event-read` on a server you're SSH'd into in Blink, it'll print out if it's getting focus events or not.

book/src/configuration.md
```diff
@@ -44,6 +44,7 @@ You may also specify a file to use for configuration with the `-c` or
 | `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "line-numbers"]` |
 | `auto-completion` | Enable automatic pop up of auto-completion. | `true` |
 | `auto-format` | Enable automatic formatting on save. | `true` |
+| `auto-save` | Enable automatic saving on focus moving away from Helix. | `true` |
```
the-mikedavis, line 47: Other discussion: https://github.com/helix-editor/helix/commit/c84f3e539eedfa3f72b4ea948cc21cad942e634d#r80351369

Yes, please default this to `false`. Auto-save is contentious. I know people that love it. Personally I don't like it, I think it violates user intentions around file operations and having unwritten changes in a buffer is useful. My opinion is not so much important as recognizing that people are opinionated about this.

Precedent says that we disable contentious features by default. In recent memory: visible whitespace, indent-guides, color modes, cursorline, rulers.

> This feature isn't discoverable without reading the docs otherwise.

This comes up with other features and I sympathize with the premise but overall I disagree, especially in this case.

* A policy of enabling new features by default leads to a bloated experience.
* Without https://github.com/helix-editor/helix/issues/1633 this is not very discoverable when enabled by default. You need to notice that the modification indicator changed when switching focus and know what the modification indicator means.

sbromberger: I can think of three use cases where auto-save is not optimal:

1) when you're editing a file over NFS, especially when it's over a high-latency / low bandwidth connection
2) when you've got a workflow that reloads / rebuilds an app on file change (think python or autobuild scripts)
3) when you're editing on a write-limited medium (flash or WORM)

I have no doubt that some people will find this feature useful, so I'm not arguing against its inclusion. I'm simply requesting that it not be the default, since no auto-save is the "option of least surprise" - I would suggest that very few users will expect it to be on by default, since I can't think of any other editor that enables it in a default config.

groves: Fair enough. The combination of being unlike other editors, hard to notice, **and** causing changes outside the editor on the filesystem are enough together to make this scary. Switched to off by default.

> 2. when you've got a workflow that reloads / rebuilds an app on file change (think python or autobuild scripts)

This is actually one of my favorite use cases for this: running a server that autoreloads on change for a webapp that autoreloads on change. With that, I can switch windows from my editor to my browser and see what I did without having to hit any other keys.

sbromberger: > This is actually one of my favorite use cases for this: running a server that autoreloads on change for a webapp that autoreloads on change. With that, I can switch windows from my editor to my browser and see what I did without having to hit any other keys.

Except this breaks for things like werkzeug because if you switch away on a file that can't be interpreted, the entire server crashes and you have to restart it.

groves:
This includes the changes from https://github.com/helix-editor/helix/pull/3390 to depend on a released crossterm.

groves:
OK, just added a link to https://github.com/helix-editor/helix/wiki/Terminal-Support to the book so people can fill in if their terminal supports this. I figured it was better to link out to the wiki for feature support by terminal. That way it's easier for someone using a particular terminal to update the supported features.

With that doc update and the crossterm release from this morning, I think this is ready to merge. Let me know if there's anything else you'd like me to do!

aikomastboom:
I think this is a great feature and should be part any IDE system, when developing I like my files to be saved so I can be certain I am working with a build of the latest edits (broken or not).
However (as mentioned above) there are lots of other situations where this behavior is quite unwelcome or even harmful.

I would prefer this to be a startup argument, project setting or :action to toggle on the go, combined with a UI indicator of its activation state "off"/"on current"/"on all" (similar to audio repeat setting).

Hope this makes sense.

helix-view/src/editor.rs
```diff
@@ -136,6 +136,8 @@ pub struct Config {
     pub auto_completion: bool,
     /// Automatic formatting on save. Defaults to true.
     pub auto_format: bool,
+    /// Automatic save on focus lost. Defaults to true.
```
archseer, line 139:
```suggestion
    /// Automatic save on focus lost. Defaults to false.
```

helix-term/src/commands/typed.rs
```diff
@@ -2008,7 +1995,7 @@ pub static TYPABLE_COMMAND_MAP: Lazy<HashMap<&'static str, &'static TypableComma
             .collect()
     });
 
-pub fn command_mode(cx: &mut Context) {
+pub(super) fn command_mode(cx: &mut Context) {
```
archseer, line 1998: Why the change here?

groves: Before my making `write_all_impl` pub, everything else exposed for import in commands.rs was done so through `pub(super)`. I assumed this was supposed to as well and was trying to maintain consistency.

dead10ck (approved):
This looks good to me!

book/src/configuration.md
```diff
@@ -49,6 +49,7 @@ on unix operating systems.
 | `gutters` | Gutters to display: Available are `diagnostics` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "line-numbers"]` |
 | `auto-completion` | Enable automatic pop up of auto-completion. | `true` |
 | `auto-format` | Enable automatic formatting on save. | `true` |
+| `auto-save` | Enable automatic saving on focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal. | `false` |
```
sudormrfbin, line 52: I think it would be better to have this as an array to prevent breaking config breakage if new auto save events are added, like you mentioned in the opening comment:

>  ...I think an option set like the gutters option could make sense. For this, we'd have an option autosave with a default value of ["focus-lost"]...

dead10ck: This doesn't have to be a breaking change. The type could just become an enum where the value is allowed to be either a boolean or a list of strings. This is what the auto pairs config option does. True could just translate to the default list internally.

groves: Yeah, having it function like `auto-pairs` for expansion was my rough thinking. Happy to change to a list now if that's the preferred way.

archseer:
Looks good to me but there's one conflict now. Can you rebase?

the-mikedavis (approved):

Issue #2772, opened 2022-06-14
Improve help/completion info for set-option to suggest valid values


aaron404:
Currently set-option conveniently displays a list of available options to set, but upon selecting an option, you cannot see what the valid values are without travelling to the documentation.

It would be nice if the same menu also displayed valid values, for example cursor-shape could suggest `.normal .insert .select` for keys and then subsequently `block bar underline none` for values.

PR #3183, opened 2022-07-25
Add completion for nested settings

### LINKED ISSUE #3181: No autocomplete for nested settings in get/set-option
### Summary

When using the `:set-option` (or `:get-option`) there is no completion for nested options, such as `search.wrap-around`

### Reproduction Steps

 

I tried this:

1. `hx`
2. `:set-option search.`

I expected this to happen:
See completions for search settings: (`wrap-around`, `smart-case`)

Instead, this happened:
No completions appear

### Helix log

```
```

</details>

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.05, also 22.05-275-gdad6d0f3

A-Walrus:
Nested settings such as `search.wrap-around` are now also completed.
Addresses part of #2772 (show suggestion for sub-keys such as `cursor-shape.normal` but still no suggestions for valid values for that option)

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -289,7 +289,7 @@ pub mod completers {
                 .as_object()
                 .unwrap()
                 .iter()
-                .map(|(key, value)| {
+                .flat_map(|(key, value)| {
                     if let Some(map) = value.as_object() {
                         let mut v: Vec<String> = map
                             .keys()
@@ -301,7 +301,6 @@ pub mod completers {
                         vec![key.clone()]
                     }
                 })
-                .flatten()
                 .collect()
         });
 
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -288,8 +288,20 @@ pub mod completers {
             serde_json::json!(Config::default())
                 .as_object()
                 .unwrap()
-                .keys()
-                .cloned()
+                .iter()
+                .map(|(key, value)| {
+                    if let Some(map) = value.as_object() {
+                        let mut v: Vec<String> = map
+                            .keys()
+                            .map(|sub_key| format!("{}.{}", key, sub_key))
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/mod.rs
```diff
@@ -288,8 +288,19 @@ pub mod completers {
...
                 .as_object()
                 .unwrap()
-                .keys()
-                .cloned()
+                .iter()
+                .flat_map(|(key, value)| {
+                    if let Some(map) = value.as_object() {
+                        let mut v: Vec<String> = map
+                            .keys()
+                            .map(|sub_key| format!("{}.{}", key, sub_key))
```
archseer, line 296: This only works one level deep though, right?

A-Walrus: That's correct. Fixed it now to work recursively for all keys no matter how deep.

helix-term/src/ui/mod.rs
```diff
@@ -283,14 +283,26 @@ pub mod completers {
...
+    /// Recursive function to get all keys from this value and add them to vec
+    fn get_keys(value: &serde_json::Value, vec: &mut Vec<String>, scope: Option<&str>) {
+        if let Some(map) = value.as_object() {
+            for (key, value) in map.iter() {
+                let key = match scope {
+                    Some(scope) => format!("{}.{}", scope, key),
+                    None => key.clone(),
+                };
+                get_keys(value, vec, Some(&key));
+                vec.push(key);
```
the-mikedavis, line 295:
```suggestion
                if !value.is_object() {
                    vec.push(key);
                }
```

I think we should only show completions for the leaves of the object trees here. For example we should show `whitespace.render` but not `whitespace`

the-mikedavis (approved):
Works great, thanks for the fix!

AceofSpades5757:
Thank you!!!

PR #3193, opened 2022-07-25
fix: Recalculate completion when going through prompt history

Frojdholm:
Fixes a panic related to completion when using the arrow keys to navigate through the prompt history.

Minimal reproduction steps for the panic
```
<launch helix>
:asdf<Enter> <- Anything that is shorter than what you will be trying to complete next works.
:theme d <- Note that triggering completion is not necessary since it's recalculated on each character entry.
<up arrow>
<tab>
```

It should not be possible to modify the prompt `line` without also updating completion list. Most functions modifying the `line` update the completions. `Prompt::insert_str` and `Prompt::change_history` does not which allows this panic to happen. This means that the `<c-s>` shortcut probably also can create a panic and there are also possible panics in the DAP prompt as well (`helix-term/src/commands/dap.rs`).

This PR fixes the immediate problem, but I think I real solution should force update the completion list when updating the prompt line in some centralized way. Another solution might be to tie the completion list more tightly to the prompt line itself so that it can be recalculated if needed when completing.

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -466,12 +466,12 @@ impl<T: Item> Picker<T> {
             .map(|(index, _score)| &self.options[*index])
     }
 
-    pub fn save_filter(&mut self, cx: &Context) {
+    pub fn save_filter(&mut self, cx: &mut Context) {
         self.filters.clear();
         self.filters
             .extend(self.matches.iter().map(|(index, _)| *index));
         self.filters.sort_unstable(); // used for binary search later
-        self.prompt.clear(cx);
+        self.prompt.clear(cx.editor);
     }
 
     pub fn toggle_preview(&mut self) {
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -466,7 +466,7 @@ impl<T: Item> Picker<T> {
             .map(|(index, _score)| &self.options[*index])
     }
 
-    pub fn save_filter(&mut self, cx: &mut Context) {
+    pub fn save_filter(&mut self, cx: &Context) {
         self.filters.clear();
         self.filters
             .extend(self.matches.iter().map(|(index, _)| *index));
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -88,7 +88,6 @@ impl Prompt {
         let cursor = line.len();
         self.line = line;
         self.cursor = cursor;
-        self.exit_selection();
         self.recalculate_completion(editor);
         self
     }
@@ -98,6 +97,7 @@ impl Prompt {
     }
 
     pub fn recalculate_completion(&mut self, editor: &Editor) {
+        self.exit_selection();
         self.completion = (self.completion_fn)(editor, &self.line);
     }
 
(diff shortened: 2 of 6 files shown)


helix-term/src/ui/prompt.rs
```diff
@@ -550,13 +550,17 @@ impl Component for Prompt {
                 if let Some(register) = self.history_register {
                     let register = cx.editor.registers.get_mut(register);
                     self.change_history(register.read(), CompletionDirection::Backward);
+                    self.recalculate_completion(cx.editor);
+                    self.exit_selection();
                     (self.callback_fn)(cx, &self.line, PromptEvent::Update);
```
Frojdholm, line 555: I'm unsure about this `self.exit_selection()`, but it was always called together with `recalculate_completion` so I decided to add it here as well.

Frojdholm:
With the last two commits it's no longer possible to call a function that modifies the prompt line without also updating the completion.

The only direct calls to recalculate the completion now happen when you want to have a completion with an empty line. I think this makes sense since you might not always want to have the completion list.

aikomastboom (approved):
@Frojdholm Thank you for this solution.

I've done some manual testing locally and it has not crashed on me yet.

I am new to Rust.. not to programming.. and these changes look solid to me.

I also noticed the pairing of `exit_selection` and `recalculate_completion` and am wondering if an extra (private) function to join them is warranted to keep the code short. But as the order may or may not be of importance (reversed in clear function) this may just add noise, especially since this is local to this module.

great start on testing (for regression ;))

my 2cnts.

Frojdholm:
> I've done some manual testing locally and it has not crashed on me yet.

Great!

> I also noticed the pairing of exit_selection and recalculate_completion and am wondering if an extra (private) function to join them is warranted to keep the code short. But as the order may or may not be of importance (reversed in clear function) this may just add noise, especially since this is local to this module.

I looked into it. `exit_selection` only clears an index into the completion list so order here shouldn't matter. I've added a call to `exit_selection` in `recalculate_completion` since it doesn't make sense to keep the old index when the list has changed.

In the future you could try to match the previous selection into the new list if possible, but I don't think the extra complexity would be worth it.

helix-term/src/ui/prompt.rs
```diff
@@ -84,10 +84,11 @@ impl Prompt {
         }
     }
 
-    pub fn with_line(mut self, line: String) -> Self {
+    pub fn with_line(mut self, line: String, editor: &Editor) -> Self {
         let cursor = line.len();
         self.line = line;
         self.cursor = cursor;
+        self.recalculate_completion(editor);
```
Frojdholm, line 91: Another reasonable approach here would be to clear the completion list instead of recalculating it, since this function is likely to be used initializing a new prompt and the consumer might not want completions to be calculated instantly.

Frojdholm (commented):
Since the whole prompt state is very interconnected (line, completion list and selection for example), care must be taken when modifying it so as not to make it go out of sync.

With these changes I think the public API of `Prompt` should be safe such that the internal state can't go out of sync by calling any public methods.

helix-term/src/ui/picker.rs
```diff
@@ -466,12 +466,12 @@ impl<T: Item> Picker<T> {
             .map(|(index, _score)| &self.options[*index])
     }
 
-    pub fn save_filter(&mut self, cx: &Context) {
+    pub fn save_filter(&mut self, cx: &mut Context) {
```
archseer, line 469: This shouldn't need `&mut`?

PR #3202, opened 2022-07-26
fix: gruvbox color modes

AlexanderBrevig:
Add color-modes to gruvbox and gruvbox_light themes


archseer:
@jbaa As the theme creator, do these look okay?

Issue #2697, opened 2022-06-07
[Security] workspace trust: language servers & tree sitter


DuckDuckWhale:
## As a question

Is it safe to automatically run languages servers and tree sitter on arbitrary files? Sometimes I just want to view files without executing code in any way, especially when viewing other people's code.

## As a feature request

Can something like [VS Code's workspace trust](https://code.visualstudio.com/docs/editor/workspace-trust) be implemented?

## As a security vulnerability bug report

If a language server is allowed to run automatically upon the opening of a file and is allowed to execute arbitrary code (e.g. to run build scripts for Rust) when doing its job, users thinking that this is just a text editor and view arbitrary code or users who inadvertently view untrusted code by mistake have the risk of having their machines compromised.

This was [CVE-2021-34529](https://msrc.microsoft.com/update-guide/vulnerability/CVE-2021-34529) for VS Code. The relevant release notes can be found [here](https://code.visualstudio.com/updates/v1_57). I couldn't find any vulnerability reporting guidelines for this project, so here it is.

willparsons:
I think for the most part helix is not affected by this since 'Tasks', 'Debugging' and 'Extensions' tools don't exist (currently). 

As for the LSP it could be as simple as checking if the user wants to spawn an LSP then saving that setting for the workspace. The only issue is that there may be things down the line that require the 'workspace trust'.

There could be a wrapper that lets you call actions and it would determine if they were available based on the current trust setting for the workspace, this avoids the need to wrap everything in `if trusted...`.
E.g.
```rust
action!(Action::LSP::Spawn, ...)
```

I haven't looked into the helix codebase that much so this is just a idea, I'm not sure how effectively it would work. What do you think?

sgued:
As a general idea I think it could be good to not have the language server always running. An option to have off by default and a command to start it when needed would be welcome.

crabdancing:
Any progress on this?

the-dipsy:
[...]

PR #3207, opened 2022-07-27
feat: load local config

AlexanderBrevig:
This will enable having a `.helix/config.toml` in a project repo and have it override default and `~/.config/helix/config.toml` settings.

This behaviour is controlled with `config.toml`:
```toml
[editor.security]
load-local-config = false
confirm-local-config = true
```

The default keeps Helix working as it is. 
If you only set `load-local-config = true` then you'll be promted to answer `yes` before loading. 
This prompt can be removed by setting `confirm-local-config = false`.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -28,7 +28,7 @@ hidden = false
 You may also specify a file to use for configuration with the `-c` or
 `--config` CLI argument: `hx -c path/to/custom-config.toml`. 
 
-Finally, you can have a `config.toml` local to a project by it under a `.helix` directory in your repository.
+Finally, you can have a `config.toml` local to a project by putting it under a `.helix` directory in your repository.
 Its settings will be merged with the configuration directory `config.toml` and the built-in configuration,
 if you have enabled the feature under `[editor.security]` in your global configuration.
 
@@ -240,4 +240,4 @@ Features that the paranoid among us may choose to disable.
 
 | Key                    | Description                                                                             | Default |
 | ---                    | ---                                                                                     | ---     |
-| `load-local-config`    | Load `config.yaml` from `$PWD/.helix` that will merge with your global configuration.   | `true` |
+| `load-local-config`    | Load `config.toml` from `$PWD/.helix` that will merge with your global configuration.   | `true` |
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -25,15 +25,12 @@ select = "underline"
 hidden = false
 ```
 
-<<<<<<< HEAD
 You may also specify a file to use for configuration with the `-c` or
-`--config` CLI argument: `hx -c path/to/custom-config.toml`.
-=======
-Editor configuration may also be overridden local to a project by creating
-a `config.toml` file under a `.helix` directory. Its settings will be merged
-with the configuration directory `config.toml` and the built-in configuration.
+`--config` CLI argument: `hx -c path/to/custom-config.toml`. 
 
->>>>>>> a4a8e122 (docs: add section about local config, written like the one for languages)
+Finally, you can have a `config.toml` local to a project by it under a `.helix` directory in your repository.
+Its settings will be merged with the configuration directory `config.toml` and the built-in configuration,
+if you have enabled the feature under `[editor.security]` in your global configuration.
 
 ## Editor
 
@@ -236,3 +233,13 @@ Example:
 render = true
 character = "╎"
 ```
+
+### `[editor.security]` Section
+
+Opt in to features that may put you at risk. 
+It is possible to write malicious TOML files, so we suggest you keep the `confirm_*` options to their default value of `true`.
+
+| Key                    | Description                                                                             | Default |
+| ---                    | ---                                                                                     | ---     |
+| `load_local_config`    | Load a `config.yaml` from `$PWD/.helix` that will merge with your global configuration. | `false` |
+| `confirm_local_config` | Prompt the user at launch in order to accept the loading of a found local `$PWD/.helix` | `true`  |
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -25,8 +25,15 @@ select = "underline"
 hidden = false
 ```
 
+<<<<<<< HEAD
 You may also specify a file to use for configuration with the `-c` or
 `--config` CLI argument: `hx -c path/to/custom-config.toml`.
+=======
+Editor configuration may also be overridden local to a project by creating
+a `config.toml` file under a `.helix` directory. Its settings will be merged
+with the configuration directory `config.toml` and the built-in configuration.
+
+>>>>>>> a4a8e122 (docs: add section about local config, written like the one for languages)
 
 ## Editor
 
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -237,7 +237,6 @@ character = "╎"
 ### `[editor.security]` Section
 
 Opt in to features that may put you at risk. 
-It is possible to write malicious TOML files, so we suggest you keep the `confirm_*` options to their default value of `true`.
 
 | Key                    | Description                                                                             | Default |
 | ---                    | ---                                                                                     | ---     |
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -236,8 +236,8 @@ character = "╎"
 
 ### `[editor.security]` Section
 
-Opt in to features that may put you at risk. 
+Features that the paranoid among us may choose to disable.
 
 | Key                    | Description                                                                             | Default |
 | ---                    | ---                                                                                     | ---     |
-| `load-local-config`    | Load `config.yaml` from `$PWD/.helix` that will merge with your global configuration.   | `false` |
+| `load-local-config`    | Load `config.yaml` from `$PWD/.helix` that will merge with your global configuration.   | `true` |
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -241,5 +241,4 @@ It is possible to write malicious TOML files, so we suggest you keep the `confir
 
 | Key                    | Description                                                                             | Default |
 | ---                    | ---                                                                                     | ---     |
-| `load_local_config`    | Load a `config.yaml` from `$PWD/.helix` that will merge with your global configuration. | `false` |
-| `confirm_local_config` | Prompt the user at launch in order to accept the loading of a found local `$PWD/.helix` | `true`  |
+| `load-local-config`    | Load `config.yaml` from `$PWD/.helix` that will merge with your global configuration.   | `false` |
--- a/helix-term/src/config.rs
+++ b/helix-term/src/config.rs
@@ -88,7 +88,7 @@ impl Config {
         config_path: std::path::PathBuf,
     ) -> Option<toml::Value> {
         if !config_path.exists() {
-            return None;
+            None
         } else {
             let mut confirmed = true;
             if config_path != helix_loader::config_file() {
@@ -128,9 +128,9 @@ impl Config {
                         None
                     }
                 };
-                return cfg;
+                cfg
             } else {
-                return None;
+                None
             }
         }
     }
--- a/helix-term/src/config.rs
+++ b/helix-term/src/config.rs
@@ -90,24 +90,9 @@ impl Config {
         if !config_path.exists() {
             return None;
         }
-        let mut confirmed = true;
-        if config_path != helix_loader::config_file() {
-            if root_config.editor.security.load_local_config {
-                if root_config.editor.security.confirm_local_config {
-                    eprintln!(
-                        "Type yes<ENTER> to continue with loading config: {:#?}",
-                        config_path
-                    );
-                    let mut input = String::new();
-                    std::io::stdin().read_line(&mut input).unwrap_or_default();
-                    confirmed = input.contains("yes");
-                } //else we still presume confirmed = true
-            } else {
-                // User denies local configs, do not load
-                confirmed = false;
-            }
-        }
-        if confirmed {
+        if config_path == helix_loader::config_file()
+            || root_config.editor.security.load_local_config
+        {
             log::debug!("Load config: {:?}", config_path);
             let bytes = std::fs::read(&config_path);
             let cfg: Option<toml::Value> = match bytes {
--- a/helix-term/src/config.rs
+++ b/helix-term/src/config.rs
@@ -55,6 +55,95 @@ impl Config {
     pub fn load_default() -> Result<Config, ConfigLoadError> {
         Config::load(helix_loader::config_file())
     }
+
+    // Load a merged config from configuration and $PWD/.helix/config.toml
+    pub fn load_merged_config() -> Config {
+        let root_config: Config = std::fs::read_to_string(helix_loader::config_file())
+            .ok()
+            .and_then(|config| toml::from_str(&config).ok())
+            .unwrap_or_else(|| {
+                eprintln!("Bad config: {:?}", helix_loader::config_file());
+                Config::halt_and_confirm("default");
+                Config::default()
+            });
+
+        // Load each config file
+        let local_config_values = helix_loader::local_config_dirs()
+            .into_iter()
+            .map(|path| path.join("config.toml"))
+            .chain([helix_loader::config_file()])
+            .filter_map(|file| Config::load_config_toml_values(&root_config, file));
+
+        // Merge configs and return, or alert user of error and load default
+        match local_config_values
+            .fold(toml::Value::Table(toml::value::Table::default()), |a, b| {
+                helix_loader::merge_toml_values(b, a, 3)
+            })
+            .try_into()
+        {
+            Ok(conf) => conf,
+            Err(_) => root_config,
+        }
+    }
+
+    // Load a specific config file if allowed by config
+    // Stay with toml::Values as they can be merged
+    pub fn load_config_toml_values(
+        root_config: &Config,
+        config_path: std::path::PathBuf,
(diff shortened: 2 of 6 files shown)


helix-loader/src/main.rs
```diff
@@ -5,5 +5,6 @@ use helix_loader::grammar::fetch_grammars;
 // compilation time. This is not meant to be run manually.
 
 fn main() -> Result<()> {
+    println!("uo");
```
kirawi, line 8: You should delete this.

helix-term/src/main.rs
```diff
@@ -114,24 +116,40 @@ FLAGS:
...
-                let _ = std::io::stdin().read(&mut []);
-                Config::default()
-            }),
-        Err(err) if err.kind() == std::io::ErrorKind::NotFound => Config::default(),
-        Err(err) => return Err(Error::new(err)),
+    let config_value = local_config_dirs()
+        .into_iter()
+        .chain([config_dir()].into_iter())
+        .map(|path| path.join("config.toml"))
+        .rev()
```
kirawi, line 123: Why call `rev()`? I think that would make the local configs be overridden by the upper configs.

helix-loader/src/main.rs
AlexanderBrevig: Wow. I have no idea how or why that's there. I'll go through and fix lint issues and of course remove this. Thanks!

AlexanderBrevig:
Thanks for the review and input @kirawi! I've fixed the requested changes (an embarrassing reminder that late night sessions may lead to weird things...) and fixed clippy lint errors locally.

AlexanderBrevig (commented):
Maybe it should use `eprintln` and inform the user on startup (file and error message) much like the current behaviour.

Probably also should update the docs. Will fix this tomorrow (no computer today).

:heavy_check_mark: This is now implemented.

d12bb:
This should be optional, as one may not want to use custom config coming with someone else's repo and, more important, this could harm the user's computer, for example by resetting a common key bind to some harmful shell command.

the-mikedavis:
That probably falls under something like https://github.com/helix-editor/helix/issues/2697 (or is similar in spirit). It's notable that we already merge in `.helix/languages.toml` and I think you could easily craft a malicious `languages.toml` like so:

```toml
[[language]]
name = "toml"
language-server = { command = "rm", args = ["-rf", "/"] }
```

AlexanderBrevig:
I've added this to `config.toml`

```toml
[editor.security]
load-local-config = false
confirm-local-config = true
```

(commented out, you will also see `-languages` versions for a future PR)

That's the default. If you toggle both, then Helix will load `$PWD/.helix/config.toml` without asking to do so.

What do you think?
I very much like the feature, but I am not married to the implementation details. I'll fix anything you see fit.

AlexanderBrevig:
Updated the docs and merged with master. `hx -c ~/team/hx/config.toml` will now load the "team" config, and then ask to load `$PWD/.helix/config.toml` if `[editor.security] load-local-config = true`.

Anything I can do to help this PR forward? (I understand the team is under a lot of load)

AlexanderBrevig:
I've fixed your suggestions @kirawi, do you have some more feedback or questions? :)

helix-term/src/main.rs
```diff
@@ -115,25 +115,72 @@ FLAGS:
...
-            }),
-        Err(err) if err.kind() == std::io::ErrorKind::NotFound => Config::default(),
-        Err(err) => return Err(Error::new(err)),
+    // We need the config, to get the editor.security.[load|confirm]_local_config flags
+    let root_config = match std::fs::read_to_string(helix_loader::config_file())
+        .map(|config| toml::from_str(&config))?
+    {
+        Ok(config) => config,
+        Err(_) => Config::default(),
+    };
```
pickfire, line 126:
```suggestion
    let root_config = match std::fs::read_to_string(helix_loader::config_file())
        .map(|config| toml::from_str(&config))?
        .unwrap_or_default();
```
I think this should work.

AlexanderBrevig: I will try

helix-term/src/main.rs
```diff
@@ -115,25 +115,69 @@ FLAGS:
...
+            .try_into()?,
+        Err(err) => {
+            eprintln!("Bad configs: {:#?}", err);
+            eprintln!("Press <ENTER> to continue with default config");
+            use std::io::Read;
+            // This waits for an enter press.
+            let _ = std::io::stdin().read(&mut []);
+            Config::default()
+        }
     };
```
pickfire, line 181: Please move these into a function, it became quite huge.

helix-term/src/main.rs
```diff
@@ -115,25 +115,69 @@ FLAGS:
...
-                eprintln!("Bad config: {}", err);
-                eprintln!("Press <ENTER> to continue with default config");
-                use std::io::Read;
-                let _ = std::io::stdin().read(&mut []);
-                Config::default()
-            }),
-        Err(err) if err.kind() == std::io::ErrorKind::NotFound => Config::default(),
-        Err(err) => return Err(Error::new(err)),
+    // We need the config, to get the editor.security.[load|confirm]_local_config flags
+    let root_config = std::fs::read_to_string(helix_loader::config_file())
```
pickfire, line 121:
```suggestion
    let root_config: Config = std::fs::read_to_string(helix_loader::config_file())
```

helix-term/src/main.rs
```diff
@@ -115,25 +115,69 @@ FLAGS:
...
+        .map(|config| toml::from_str(&config))?
+        .unwrap_or_default();
+
+    // Now, let's load config based on a PathBuf
+    // We will keep it toml::Value so we can use helix_loader::merge_toml_values
+    // This will respect editor.security settings
+    let load_config = |config_path: std::path::PathBuf| {
+        if std::path::Path::new(&config_path).exists() {
```
pickfire, line 129:
```suggestion
        if config_path.exists() {
```
This auto deref to `Path`, so don't need to manually create new `Path`.

helix-term/src/main.rs
```diff
@@ -115,25 +115,69 @@ FLAGS:
...
+                    if root_config.editor.security.confirm_local_config {
+                        eprintln!(
+                            "Type yes to continue with loading config: {:#?}",
+                            config_path
+                        );
+                        use std::io::Read;
+                        confirmed = match std::io::stdin().read(&mut [1, 2, 3, 4]) {
+                            Ok(n) => n == "yesE".len(),
+                            Err(_) => false,
+                        }
```
pickfire, line 142: What is this doing?

AlexanderBrevig: That whole thing is embarrasing. It should rather match on the actual input of `yes`. I thought I had fixed that to be honest. I'll make a pass on it now and refactor out that function you requested.

AlexanderBrevig:
Huge thank you @pickfire I can see that I maybe wrote this code a bit prematurely in terms of how well (let's be honest, bad) I knew Rust.

However, the strange thing now (before the refactor too) is that I seem to lose ability to input text after the editor is loaded. It did not use to be that way.

Reproduce:
Add `[editor.security] load-local-config = true`  and maybe some config file in `./helix/config.toml`.
Then `cargo run .` will first ask you to type `yes<RET>` if you have a file, then open the file picker - here you can type. When you open a file, suddenly I cannot type any more.
I have very little idea where to check, and somehow doubt that its this code that introduces this error. Unless `std::io::stdin().read_line` somehow blocks the events coming in after...?

I did some cleanup in addition to yours. It is better now, at least. A bit unsire if the functions should be in config.rs or simply at the bottom of main.rs.

helix-term/src/main.rs
AlexanderBrevig: I have refactored the entire thing. Let me know if I should do another pass at the Rust, but to me it is a lot better now. Thanks again for your time and input!

(I don't know what will generate notifications but I will presume my comment and the "resolve conversation" will suffice).

book/src/configuration.md
```diff
@@ -229,3 +233,13 @@ Example:
...
+
+### `[editor.security]` Section
+
+Opt in to features that may put you at risk. 
+It is possible to write malicious TOML files, so we suggest you keep the `confirm_*` options to their default value of `true`.
+
+| Key                    | Description                                                                             | Default |
+| ---                    | ---                                                                                     | ---     |
+| `load_local_config`    | Load a `config.yaml` from `$PWD/.helix` that will merge with your global configuration. | `false` |
+| `confirm_local_config` | Prompt the user at launch in order to accept the loading of a found local `$PWD/.helix` | `true`  |
```
pickfire, line 245: Not sure if this is just me but I don't find this quite important.

helix-term/src/config.rs
```diff
@@ -55,6 +55,95 @@ impl Config {
...
+            .into_iter()
+            .map(|path| path.join("config.toml"))
+            .chain([helix_loader::config_file()])
+            .filter_map(|file| Config::load_config_toml_values(&root_config, file));
+
+        // Merge configs and return, or alert user of error and load default
+        match local_config_values
+            .fold(toml::Value::Table(toml::value::Table::default()), |a, b| {
+                helix_loader::merge_toml_values(b, a, 3)
+            })
```
pickfire, line 81: I believe we can use `reduce` instead.

helix-term/src/config.rs
```diff
@@ -55,6 +55,95 @@ impl Config {
...
+            let mut confirmed = true;
+            if config_path != helix_loader::config_file() {
+                if root_config.editor.security.load_local_config {
+                    if root_config.editor.security.confirm_local_config {
+                        eprintln!(
+                            "Type yes<ENTER> to continue with loading config: {:#?}",
+                            config_path
+                        );
+                        let mut input = String::new();
+                        std::io::stdin().read_line(&mut input).unwrap_or_default();
```
pickfire, line 105: >  Then cargo run . will first ask you to type yes<RET> if you have a file, then open the file picker - here you can type. When you open a file, suddenly I cannot type any more.

Most likely what you mentioned is due to this line, `read_line` probably interfere with raw input mode or something, I am not sure.

helix-term/src/config.rs
```diff
@@ -55,6 +55,95 @@ impl Config {
...
+        }
+    }
+
+    // Load a specific config file if allowed by config
+    // Stay with toml::Values as they can be merged
+    pub fn load_config_toml_values(
+        root_config: &Config,
+        config_path: std::path::PathBuf,
+    ) -> Option<toml::Value> {
+        if config_path.exists() {
```
pickfire, line 95: Doing fast exit on huge nested function is good, like
```rust
        if !config_path.exists() {
            return None;
        }

File: helix-term/src/config.rs (hunk 10 above)
Comment (AlexanderBrevig) [reply helix-term/src/config.rs:105]: It turned out to be some setting in my config.toml
I should narrow it down (but I suspect it was my escape rebind)

**Review by pickfire — COMMENTED** [review]
`load_local_config` part seemed okay to me but I am not sure about the `confirm_local_config` part.

**Comment by AlexanderBrevig** [comment]
> `load_local_config` part seemed okay to me but I am not sure about the `confirm_local_config` part.

I am in favor of removing the confirmation prompt. Should I do it?

**Comment by pickfire** [comment]
> I am in favor of removing the confirmation prompt. Should I do it?

Yes, I think for now we can keep it with a small change to only have a config to allow loading local config. That would be easier review and try it out.

File: book/src/configuration.md [hunk 12]
```diff
@@ -229,3 +233,12 @@ Example:
 render = true
 character = "╎"
 ```
+
+### `[editor.security]` Section
+
+Opt in to features that may put you at risk. 
+It is possible to write malicious TOML files, so we suggest you keep the `confirm_*` options to their default value of `true`.
```
pickfire, line 240: I think the confirm part is outdated.

helix-term/src/config.rs
```diff
@@ -87,11 +87,9 @@ impl Config {
...
         config_path: std::path::PathBuf,
     ) -> Option<toml::Value> {
-        if !config_path.exists() {
-            return None;
-        }
-        if config_path == helix_loader::config_file()
-            || root_config.editor.security.load_local_config
+        if config_path.exists()
+            && (config_path == helix_loader::config_file()
+                || root_config.editor.security.load_local_config)
```
pickfire, line 92: Still same thing, this 3 can just `return None`; to reduce one indentation level.

AlexanderBrevig: ...embarrasing, thanks!

the-mikedavis:
The security aspect of this still bothers me a bit - it ends up really nerfing the feature. I think it's silly in the same way that git's [cve-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765) (the safe.directory one) is silly, and to my knowledge no other terminal based editor is careful about this. I would argue at least that this should be enabled by default.

AlexanderBrevig:
> The security aspect of this still bothers me a bit - it ends up really nerfing the feature. I think it's silly in the same way that git's [cve-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765) (the safe.directory one) is silly, and to my knowledge no other terminal based editor is careful about this. I would argue at least that this should be enabled by default.

Since three of four voices in this PR seems to agree, I'll change the default.

pickfire (approved):
Looks good to me. Thanks for working on this.

book/src/configuration.md
```diff
@@ -229,3 +233,11 @@ Example:
...
 character = "╎"
 ```
+
+### `[editor.security]` Section
+
+Opt in to features that may put you at risk. 
+
+| Key                    | Description                                                                             | Default |
+| ---                    | ---                                                                                     | ---     |
+| `load-local-config`    | Load `config.yaml` from `$PWD/.helix` that will merge with your global configuration.   | `false` |
```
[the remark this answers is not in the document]
AlexanderBrevig: Absolutely! Thanks!

book/src/configuration.md
```diff
@@ -26,7 +26,11 @@ hidden = false
 ```
 
 You may also specify a file to use for configuration with the `-c` or
-`--config` CLI argument: `hx -c path/to/custom-config.toml`.
+`--config` CLI argument: `hx -c path/to/custom-config.toml`. 
+
+Finally, you can have a `config.toml` local to a project by it under a `.helix` directory in your repository.
```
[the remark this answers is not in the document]
AlexanderBrevig: Thanks! Will fix tonight 👍

AlexanderBrevig:
Somehow I totally forgot this. Sorry.

Nazariglez:
Hello fellows! This feature looks great. What needs to be done to be able to merge this? Thanks

gibbz00:
> Hello fellows! This feature looks great. What needs to be done to be able to merge this? Thanks

I did a refactor on config loading, https://github.com/helix-editor/helix/pull/5636. I'm in the process of merging this PR with that branch, hoping that it can help getting this one merged too.

Nazariglez:
That's awesome @gibbz00, thank you!

pascalkuthe:
> I Hope I'm not too late, but I would like helix to also support `(config|languages).local.toml`, that way I can easily have a global gitignore for it and avoid polluting repos with my configs and/or ignores.
> 
> The exact behavior can either be local replaces the non-local file or is merged with it, favoring the local one (I prefer the second option but both can be argued for)

What exactly is the advantage of `config.local.toml` over `.helix/config.toml` in that regard?
I think adding `.helix/**` to the global gitignore would be enough (that would also take care of `.helix/languages.toml`)

poliorcetics:
As helix becomes more used, I expect some repos will start having a config in their git, like many already do for VSCode, jetbrains IDEs or vim/neovim, in this case `.helix/**` won't be enough

the-mikedavis:
`.helix/languages.toml` and `.helix/config.toml` also work in the global gitignore. I also don't see the advantage to using `config.local.toml` and `languages.local.toml`: it should be equivalent to `.helix/config.toml` and `.helix/languages.toml` respectively, no?

poliorcetics:
No, what I mean is, in the project repo, people will start committing `<my repo>/.helix/{config,languages}.toml` (and ideally it should be `.config/helix/...` even locally, like `cargo-nextest` does), and having the hability to override it locally without having to play with the Git index would be nice.

As an example, search [`path:/.vscode/settings.json`](https://cs.github.com/?scopeName=All+repos&scope=&q=path%3A.vscode%2Fsettings.json). The results are murkier for Vim and Jetbrains IDEs but there are examples.

the-mikedavis:
Oh I see, `config.local.toml` would be in addition to `.helix/config.toml`, right?

For scenarios like this I think it would be better to make use of the `-c <config-file>` CLI option and have a wrapper around `hx -c` to control which config is used based on the directory it's called in

poliorcetics:
> For scenarios like this I think it would be better to make use of the `-c <config-file>` CLI option and have a wrapper around `hx -c` to control which config is used based on the directory it's called in

That seems like a very fast path to the disaster that is using `$CARGO_TARGET_DIR` globally and doesn't allow for merging configs in the local -> project -> user order, because projects may be settings configs to help users go in more smoothly and having to pass `-c` is simply a workaround

charles-dyfis-net:
> The security aspect of this still bothers me a bit - it ends up really nerfing the feature. I think it's silly in the same way that git's [cve-2022-24765](https://github.blog/2022-04-12-git-security-vulnerability-announced/#cve-2022-24765) (the safe.directory one) is silly, and to my knowledge no other terminal based editor is careful about this. I would argue at least that this should be enabled by default.

Both vim and emacs disallow modelines from configuring settings that can result in executables being run by default / without user approval.

PR #3212, opened 2022-07-27
Remove let_declaration from Rust locals tracking

the-mikedavis:
This fixes some edge-cases like [here](https://github.com/helix-editor/helix/blob/9a496237215dc09ef8f639fc79f7e63e2309c080/helix-term/src/commands.rs#L1752-L1757):

| before | after |
|---|---|
| before | after |

I don't think it ever makes sense to have a let-declaration define a local, right? We want to tag parameters as `local.definition` so that you can identify function/closure parameters but let-declarations always create variable bindings even when pattern matching so we can use normal highlighting for those.

### DIFF
--- a/runtime/queries/rust/locals.scm
+++ b/runtime/queries/rust/locals.scm
@@ -8,9 +8,6 @@
 (parameter
   (identifier) @local.definition)
 
-(let_declaration
-  pattern: (identifier) @local.definition)
-
 (closure_parameters (identifier) @local.definition)
 
 ; References


poliorcetics:
For those looking at these and searching for the difference, look at the `regex::` on the 4th line, I spent way too much time searching for it 😅

poliorcetics:
What is the result for later uses of the `regex` variable ? Does it remove some highlighting ? Could this be fixed by detecting modules first when seeing something like `name::` ?

the-mikedavis:
> What is the result for later uses of the `regex` variable?

Later uses have the `@variable` highlighting as expected. This change only removes the `@variable` highlighting on the `regex::` instance.

> Could this be fixed by detecting modules first when seeing something like `name::`?

Not without some changes to the syntax highlighting code. Local reference highlights are higher precedence than anything highlights from `highlights.scm` and you can't currently 'block' a `local.reference` with something like

```scm
(scoped_identifier path: (identifier) @ignore)
(identifier) @local.reference
```

poliorcetics:
Thanks for the explanation !

archseer (approved):
Seems reasonable 👍

PR #3213, opened 2022-07-27
Add textobject for javascript

### LINKED ISSUE #3177: How to enable Textobject queries for TS and JS?
How to enable Textobject queries for TS and JS?
image

erasin:
Fix #3177

### DIFF
--- a/runtime/queries/javascript/textobjects.scm
+++ b/runtime/queries/javascript/textobjects.scm
@@ -1,14 +1,14 @@
 (function_declaration
-  body: (_)) @function.around
+  body: (_) @function.inside) @function.around
 
 (function
-  body: (_)) @function.around
+  body: (_) @function.inside) @function.around
 
 (arrow_function
-  body: (_)) @function.around
+  body: (_) @function.inside) @function.around
 
 (method_definition
-  body: (_)) @function.around
+  body: (_) @function.inside) @function.around
 
 (generator_function_declaration
   body: (_) @function.inside) @function.around
--- a/runtime/queries/javascript/textobjects.scm
+++ b/runtime/queries/javascript/textobjects.scm
@@ -1,0 +1,35 @@
+(function_declaration
+  body: (statement_block)) @function.around
+
+(function
+  body: (statement_block)) @function.around
+
+(export_statement
+  (function_declaration) @function.around) @function.around.start
+
+(arrow_function
+  body: (statement_block)) @function.around
+
+(method_definition
+  body: (statement_block)) @function.around
+
+(class_declaration
+  body: (class_body) @class.inside) @class.around
+
+(export_statement
+  (class_declaration) @class.around) @class.around.start
+
+(export_statement
+  (class) @class.around) @class.around.start
+
+(_ (statement_block) @block.inside) @block.around
+
+(formal_parameters
+  ((_) @parameter.inside . ","? @parameter.around) @parameter.around)
+
+(arguments
+  ((_) @parameter.inside . ","? @parameter.around) @parameter.around)
+
+(comment) @comment.inside
+
+(comment)+ @comment.around
--- a/runtime/queries/javascript/textobjects.scm
+++ b/runtime/queries/javascript/textobjects.scm
@@ -13,16 +13,14 @@
 (method_definition
   body: (statement_block)) @function.around
 
+(generator_function_declaration
+  body: (_) @function.inside) @function.around
+
 (class_declaration
   body: (class_body) @class.inside) @class.around
 
-(export_statement
-  (class_declaration) @class.around) @class.around.start
-
-(export_statement
-  (class) @class.around) @class.around.start
-
-(_ (statement_block) @block.inside) @block.around
+(class
+  (class_body) @class.inside) @class.around
 
 (formal_parameters
   ((_) @parameter.inside . ","? @parameter.around) @parameter.around)
--- a/runtime/queries/javascript/textobjects.scm
+++ b/runtime/queries/javascript/textobjects.scm
@@ -5,7 +5,7 @@
   body: (statement_block)) @function.around
 
 (export_statement
-  (function_declaration) @function.around) @function.around.start
+  declaration: (function_declaration) @function.around) 
 
 (arrow_function
   body: (statement_block)) @function.around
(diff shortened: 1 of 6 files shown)


runtime/queries/javascript/textobjects.scm
```diff
@@ -0,0 +1,35 @@
...
+  body: (statement_block)) @function.around
+
+(class_declaration
+  body: (class_body) @class.inside) @class.around
+
+(export_statement
+  (class_declaration) @class.around) @class.around.start
+
+(export_statement
+  (class) @class.around) @class.around.start
```
the-mikedavis, line 23: what are the `.start` captures here? I think you should be able to remove them

the-mikedavis:
To fix the docs CI, you'll need to run `cargo xtask docgen` and commit the changes

axin:
You have forgotten about generator functions and class expressions:
```
(generator_function_declaration
  body: (_) @function.inside) @function.around

(class
  (class_body) @class.inside) @class.around
```

David-Else:
Thanks for this! Can we get all of: https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/javascript/textobjects.scm ? (Sorry I don't really know the technical details of this matter)

runtime/queries/javascript/textobjects.scm
erasin: thanks , i  fix it

runtime/queries/javascript/textobjects.scm
```diff
@@ -0,0 +1,33 @@
+(function_declaration
+  body: (statement_block)) @function.around
+
+(function
+  body: (statement_block)) @function.around
+
+(export_statement
+  (function_declaration) @function.around) @function.around.start
```
the-mikedavis, line 8:
```suggestion
  (function_declaration) @function.around)
```

runtime/queries/javascript/textobjects.scm
```diff
@@ -0,0 +1,36 @@
+(function_declaration
+  body: (statement_block)) @function.around
+
+(function
+  body: (statement_block)) @function.around
+
+(export_statement
+  declaration: (function_declaration) @function.around)
```
the-mikedavis, line 8: Can we move this down to the other `export_statement`? You could even combine the stanzas:

```scm
(export_statement
  declaration: [
    (function_declaration) @function.around
    (class_declaration) @class.around
  ])
```

It's a little out of place here since it's separating three function-specific queries

runtime/queries/javascript/textobjects.scm
```diff
@@ -0,0 +1,36 @@
...
+  body: (statement_block)) @function.around
+
+(function
+  body: (statement_block)) @function.around
+
+(export_statement
+  declaration: (function_declaration) @function.around) 
+
+(arrow_function
+  body: (statement_block)) @function.around
```
the-mikedavis, line 11: Instead of `(statement_block)` let's match on `(_)` in these rules. This allows capturing single-expression arrow functions like `x => x + 1`. That node can also me barked as `function.inside` which is usually used for function bodies:

```scm
(function body: (_) @function.inside) @function.around
```

the-mikedavis:
It looks like this already covers the textobjects from nvim-treesitter which are implemented in helix (i.e. not ones like `conditional` or `loop`)

erasin:
this is my test file,
- comment
- function
- arguments
- class
    - interface (ts)
    - type (ts)

```typescript 
// js

// function
function a() { }
function b(b1) { }
export function c() { }
const d = function() { }
const e = {
  f1: function(c, d) { },
  f2: () => { }
}
const f1 = (a) => { }
const f2 = () => { () => { }; }
const f3 = f1(() => { })
const f4 = x => x + 1

// class
class A {
  constructor() { }
  tes1(a, b = 1) { }
}

export class B { }
const C = class { }

// ts 

interface I1 {
  s: string
  b: boolean
  (b: boolean): string
}

type T2 = {
  s: string
  b: boolean
  f: () => void
}

class D implements I1 {

  constructor(s: string, b: boolean) {
    this.s = s
    this.b = b
  }

  s: string
  b: boolean

  f(b: boolean): string {
    return b ? "ok": ""
  }
}
```

runtime/queries/javascript/textobjects.scm
```diff
@@ -0,0 +1,36 @@
...
+  body: (_)) @function.around
+
+(function
+  body: (_)) @function.around
+
+(arrow_function
+  body: (_)) @function.around
+
+(method_definition
+  body: (_)) @function.around
```
the-mikedavis, line 11: for all of these we can mark the body node as `function.inside`, for example:

```
(function
   body: (_) @function.inside) @function.around
```

runtime/queries/typescript/textobjects.scm
```diff
@@ -0,0 +1,7 @@
+; inherits: javascript
+
+(interface_declaration
+  name: (_) ) @class.around
+  
+(type_alias_declaration
+  name: (_) ) @class.around
```
the-mikedavis, line 7: we don't use the name node here so the pattern match can be removed:

```
(interface_declaration) @class.around
(type_alias_declaration) @class.around
```

Or they could be combined

```
[
  (interface_description)
  (type_alias_declaration)
] @class.around
```

runtime/queries/javascript/textobjects.scm
erasin: I don't quite understand mark `function.inside` .What is the difference between inside and not marking it?

the-mikedavis: You can use `mif` (inside) to select inside the function. For most (maybe all?) languages this selects the body of the function while `maf` (around) also selects things like the function name and any function keywords

erasin: thanks, i just test like `maf` and `]f` Navigating...

runtime/queries/typescript/textobjects.scm
erasin: have change like this 
```
[
  (interface_declaration 
    body:(_) @class.inside)
  (type_alias_declaration 
    value: (_) @class.inside)
] @class.around
```

the-mikedavis (approved):
These look great, thanks for working on this!

Issue #2508, opened 2022-05-19
Sometimes wrong order in autocomplete menu


ChrHorn:
### Summary

There seems to be an issue with the sorting of the autocompletion menu. The fuzzy filtering is correct but the order of the displayed items is a bit strange.

### Reproduction Steps

 

I tried this:

1. Empty Rust project
2. Type `use std::con`

I expected this to happen:

Items that start with `con` should be first in the list

Instead, this happened:

![image](https://user-images.githubusercontent.com/17070041/169383449-f1b3df8f-7b00-4f5a-997d-96faa2b3ff3d.png)

### Platform

WSL2

### Terminal Emulator

Windows Terminal

### Helix Version

22.05-dev (8493b5fc)

Issue #3084, opened 2022-07-15
fuzzy search for autocomplete


rgasper:
#### Describe your feature request

Adding a fuzzy-find capability to autocomplete would be awesome! It's probably my favorite feature from VSCode that I miss in Helix. In particular, the VSCode autocomplete provides the following types of suggestions that I'm missing:
- will show suggestions where the currently typed stub is in the middle
  - so if I've typed `back`, the autocomplete will suggest options like `my_data_backup`
- will show suggestions where letters in the currently typed stub show up spaced out throughout the suggestion in order
  - so if I've typed `bgvn`, the autocomplete will suggest options like `_big_global_variable_name`
- autocomplete can be configured via settings to be case-insensitive

kirawi:
Duplicate of #1015

archseer:
Not a duplicate.

archseer:
However this should already work out of the box? We fuzzy filter the completion.

sudormrfbin:
> * will show suggestions where the currently typed stub is in the middle
>   
>   * so if I've typed `back`, the autocomplete will suggest options like `my_data_backup`

Suggestions in the middle don't seem to work (or they work but randomly in at least rust-analyzer).

> * will show suggestions where letters in the currently typed stub show up spaced out throughout the suggestion in order
>   
>   * so if I've typed `bgvn`, the autocomplete will suggest options like `_big_global_variable_name`

This seems to work with rust-analyzer.

> * autocomplete can be configured via settings to be case-insensitive

Case insensitive autocomplete doesn't work (tested with rust-analyzer and the Dart LSP).

Maybe there's LSP specific settings that have to be modified?

sudormrfbin:
So this seems to be due to filtering the completion results and removing completions that don't _start_ with the current word:

https://github.com/helix-editor/helix/blob/fb625697854e3a9d99aa4cfe952f2050b7e82ce0/helix-term/src/commands.rs#L3712-L3722

ChrHorn:
After https://github.com/helix-editor/helix/pull/3215 gets merged we could try to simply remove the prefix filerting to get full fuzzy matching. This would also fix https://github.com/helix-editor/helix/issues/1807.

PR #3215, opened 2022-07-27
feat: menu sort strategy for fuzzy match score sorting in completion …

lazytanuki:
This adds a sorting strategy to the menu.

As of now, all completions are sorted alphabetically after passing the fuzzy match filter. With this PR they are sorted by their fuzzy match score after passing the filter.
Code actions menu's behavior remains the same.

This comes from #2705, where it is needed for LSP preselect.

The diff may seem big because of the rustmft formatting that changed the indentation of a whole closure, but the only changes are:

- a new enum
- a new argument to the `Menu` constructor
- a match on the enum in the `score()` method


ChrHorn:
This would fix https://github.com/helix-editor/helix/issues/2508

the-mikedavis:
I don't think this fixes #2508. I tried it against the reproduction case there and the results are the same.

lazytanuki:
Are you sure ? I just tried the reproduction case in #2508 and this seems to solve the issue for me.
image

the-mikedavis:
Whoops I was on the wrong branch 😅

Yep, looks like this does fix #2508 👍

lazytanuki:
Just rebased this on master, could someone review this pretty please ? It would unblock some PRs, including the LSP preselect feature that I implemented a couple months ago.

As a reminder: the diff may seem big because of the rustmft formatting that changed the indentation of a whole closure, but the only changes are:

- a new enum
- a new argument to the `Menu` constructor
- a match on the enum in the `score()` method

archseer:
Do we need the two strategies though? Why not just change the behaviour to sort based on the score?

lazytanuki:
Some pickers such as code actions do not use a score AFAIK, so it makes sense for them to be sorted alphabetically (which they are on master).

lukepighetti:
Looks like this is blocking https://github.com/helix-editor/helix/issues/3084#issuecomment-1200871015

lazytanuki:
image

archseer:
To review I've used `?w=1` on the URL which ignores whitespace changes.

I still think this could be a method on the `Item` trait? Something that returns `Ord` and `pattern` and `score` would be one of the inputs. That way different menus could have their own sort strategies rather than hardcoding it to two.


> Some pickers such as code actions do not use a score AFAIK, so it makes sense for them to be sorted alphabetically (which they are on master).

#4134 does change the strategy to sort by score too. So maybe #4134 is enough, then #2705 can be implemented on top of https://github.com/helix-editor/helix/pull/4134/commits/63a54ee19b522cb0b085f1da0f3dfe07174179d5

lazytanuki:
Superseeded by #4480.

Issue #3059, opened 2022-07-13
Background Image | Wallpaper


RqndomHax:
#### Describe your feature request

Hey !
It'd be cool to have the ability to set a background-image on supported terminals (like kitty).

Like on IntelliJ:
![intellij_example](https://user-images.githubusercontent.com/24386960/178700897-1462e9d6-0d51-4d93-8d41-47f04b582a84.png)

Here is an example of what it'd be on helix:
![helix_example](https://user-images.githubusercontent.com/24386960/178712970-e370ece6-59ec-443c-b779-1e02957d7f1b.png)

RqndomHax:
Kitty can draw images in the terminal,

Yeah I guess I'll use transparent background in that case,

Thanks;

bluebrown:
@CptPotato, commenting out doesn't pair well when trying to overwrite a pallet. There should be a color named `none` or `transparent` that we can use for overwrites.

CptPotato:
> @CptPotato, commenting out doesn't pair well when trying to overwrite a pallet. Imo, there should be a color named `none` or `transparent` that we can use for overwrites.

That's not a bad idea. Feel free to create an issue for this feature if there doesn't exist one already.

gabydd:
With the inherits you can set ``ui.background = {}`` and then there won't be a background, this could probably be documented somewhere

freepicheep:
Thanks, @gabydd! The background was so annoying since I use a translucent terminal.

Hachikoi-the-creator:
sorry if I sound ignorant, but couldn't make it work.
How would the configuration be in my `config.toml` to have by background transparent?
currently have onedark as a theme

Hachikoi-the-creator:
nvm found it! (leaving instructions in case someone needs them)

inside `.config/helix `
- `$ mkdir themes`
- then there either copy the code from a theme you like : https://github.com/helix-editor/helix/wiki/Themes#available-themes
- copy the code in another file ex: `touch themes/oneDarker.toml`
- change `"ui.background": { }`
- select it like any other theme in your `config.toml` ex: `theme = "oneDarker"`

melMass:
You actually don't need to copy the theme, you can create a simple override theme that `inherit` the main theme: https://docs.helix-editor.com/themes.html#inheritance

PR #2170, opened 2022-04-19
Implement cursorline

TobTobXX:
Highlight all lines where cursers are on, imitating vim's `cursorline` option.

See #1761.

### Todo:
 - [x] implement it
 - [x] config option
 - [x] documentation
 - [ ] update (a few) themes

### Improvements:
- [x] Use `ui.cursorline[.primary | .secondary]`
- [x] Discuss whether to highlight secondary cursors.
- [ ] TBD: Option to disable on inactive window
- [ ] TBD: Syncing?

PR #3216, opened 2022-07-27
Add transparent theme based on base16_default

GreasySlug:
related #3059 
discussions #3142 
It was created based on the base16_default theme, and matches especially dark images.
The drawback is that the background is all transparent, making it hard to see pop-ups.
The pop-up background can be black for better visibility, but I don't do that.

### DIFF
--- a/runtime/themes/base16_transparent.toml
+++ b/runtime/themes/base16_transparent.toml
@@ -18,7 +18,9 @@
 "ui.statusline.select" = { fg = "magenta", modifiers = ["reversed"] }
 "ui.help" = { fg = "light-gray" }
 "ui.cursor" = { modifiers = ["reversed"] }
+"ui.cursor.match" = { fg = "light-yellow", modifiers = ["underlined"] }
 "ui.cursor.primary" = { modifiers = ["reversed", "slow_blink"] }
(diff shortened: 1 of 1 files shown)


the-mikedavis:
You may want to add a `ui.cursorline.primary` and maybe a `ui.cursorline.secondary` theme for [cursorline](https://github.com/helix-editor/helix/pull/2170). Otherwise this looks good to me 👍

the-mikedavis:
Did you mean to update the `ui.cursor` and not `ui.cursorline` scopes in that commit? Not defining cursorline is ok if it wouldn't look good in the theme anyways but it's worth giving it a try even if you won't use the feature

GreasySlug:
Oh, sorry, I took mistook cursor for cursorline.
However, I tried to add them, but the default colors were inevitably too intense, so I gave up adding them.
I would like to add them in the future when I can turn them on and off or add transparency shading.

the-mikedavis:
I don't think real transparency is possible because of terminal limitations, although there is similar work like dimming https://github.com/helix-editor/helix/pull/2751

the-mikedavis (approved):
👍 🎨

Could you add a screenshot to the themes wiki? https://github.com/helix-editor/helix/wiki/Themes

PR #3219, opened 2022-07-27
JS/TS highlighting improvements

Philipp-M:
This PR improves highlighting in JS and TS, like shorthand identifiers, additional operators or namespaces in TS


the-mikedavis (approved):
These look great, thanks!

Issue #2239, opened 2022-04-23
Bracketed paste


the-mikedavis:
Pasting with your operating system's paste functionality (Ctrl-v or Cmd-v) is wonky:

- it's slow (https://github.com/helix-editor/helix/issues/1021)
- it inserts extra brackets if you have `auto-pair` on (https://github.com/helix-editor/helix/issues/2015)
- indentation is wrong (https://github.com/helix-editor/helix/issues/2268)

See also

- https://github.com/helix-editor/helix/issues/1484
- https://github.com/helix-editor/helix/issues/1652
- https://github.com/helix-editor/helix/issues/1963
- https://github.com/helix-editor/helix/issues/2236#issuecomment-1107454850

This happens because crossterm does not yet support bracketed paste, so when you paste with OS paste, helix receives each character as if you had typed it yourself. We await a PR upstream in crossterm: https://github.com/crossterm-rs/crossterm/pull/557.

Until bracketed paste is supported, use `Space-p` to paste from the system clipboard.

theowenyoung:
When I use a remote server, then I can only use ctrl-v to paste the local clipboard, how to handle this case?

the-mikedavis:
Yanking to system clipboard across SSH is covered by: https://github.com/helix-editor/helix/issues/247 rather than bracketed paste

mbratch:
It's more than just wonky. It just doesn't work. I can copy and paste using the typical, basic ways to and from any other application... except helix. Whether helix is in insert mode or not, it doesn't work.

I do most of my work over an ssh connection to my Linux box. Copy/paste doesn't work there.

the-mikedavis:
That's covered by #247 and not bracketed paste. Bracketed paste support is implemented, merged, and released in v22.08.1.

PR #1343, opened 2021-12-23
Sync tmux clipboard and system clipboard

Flakebi:
When copying, also write the content into the system clipboard.
When pasting, request tmux to refresh its paste buffer from the system
clipboard, wait for the update to propagate and the request the
clipboard content from tmux.

With `set -g set-clipboard on` in tmux, this enables sharing the
system clipboard with a helix running in tmux, running in an ssh
session, running in alacritty.

The need for a wait is unfortunate, but I didn't find a better way.
Tmux asks the terminal for the clipboard content and updates its buffer
after it got an answer. The answer may come or may not. It may take
long, e.g. when running through a slow ssh connection.

The feature works well on alacritty. On konsole, it doesn't do
anything, but doesn't harm either. On xterm, `tmux refresh-client -l`
prints gibberish for me, also without using helix, so I added an
option to disable this feature.

PR #3220, opened 2022-07-27
Use OSC 52 as a fallback for setting the system clipboard

groves:
This emits an [OSC 52 terminal code](https://terminalguide.namepad.de/seq/osc-52/) for copying into the system terminal if we don't find a local provider. That gets setting the local clipboard to work through SSH.

It doesn't do the get clipboard side of OSC 52 for a few reasons:
* OSC 52 get isn't often enabled as it's a security hole that lets remote terminals read your local clipboard
* It'd require getting the clipboard to be async since we'd need to wait for the terminal response
* Getting the clipboard is way less useful since you can go into insert and use your OS-level paste command to simulate it

vim-oscyank has [this nice list of supported terminals](https://github.com/ojroques/vim-oscyank#vim-oscyank). I've tried it on Kitty on Mac and it seems to work well. I believe the primary vs clipboard type selection codes are correct, but I'd need somebody on Linux to let me know if it's working as expected.

oscyank emits a [tmux-specific code](https://github.com/ojroques/vim-oscyank/blob/360ccdc01b18cd434588c34e15e5ea382b436de8/plugin/oscyank.vim#L102) to get it to work in tmux. Since we already have a tmux specific implementation, I didn't attempt to add that. I don't use tmux enough to know if OSC 52 would be better for it than the `tmux load-buffer` call we're currently making.

### DIFF
--- a/helix-view/src/base64.rs
+++ b/helix-view/src/base64.rs
@@ -1,0 +1,163 @@
+// A minimal base64 implementation to keep from pulling in a crate for just that. It's based on
+// https://github.com/marshallpierce/rust-base64 but without all the customization options.
+// The biggest portion comes from
+// https://github.com/marshallpierce/rust-base64/blob/a675443d327e175f735a37f574de803d6a332591/src/engine/naive.rs#L42
+// Thanks, rust-base64!
+
+// The MIT License (MIT)
+
+// Copyright (c) 2015 Alice Maz
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+use std::ops::{BitAnd, BitOr, Shl, Shr};
+
+const PAD_BYTE: u8 = b'=';
+const ENCODE_TABLE: &[u8] =
+    "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789+/".as_bytes();
+const LOW_SIX_BITS: u32 = 0x3F;
+
+pub fn encode(input: &[u8]) -> String {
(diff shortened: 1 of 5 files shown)


the-mikedavis:
> Getting the clipboard is way less useful since you can go into insert and use your OS-level paste command to simulate it

#2239

groves:
Heh, I guess I hadn't done much pasting of large text. It had been fine with the smaller things I pasted, but yeah, that's painful.

Your suggestion of bracketed paste does seem like a better way to solve slow pasting than OSC 52. Just opened https://github.com/helix-editor/helix/pull/3233 to enable bracketed paste.

groves:
I've opened https://github.com/crossterm-rs/crossterm/pull/697 to move the OSC 52 command into crossterm. Once that's released, I'll switch this code to use it.

In the meantime, I think this is ready for review and possible merging.

the-mikedavis:
Given that we're only using the `base64` dependency for base64-encoding a `&str`, I don't think it's worth pulling in a whole crate. A base64 encoding function can be added in-line in probably less than 100 lines of code.

groves:
> Given that we're only using the `base64` dependency for base64-encoding a `&str`, I don't think it's worth pulling in a whole crate. A base64 encoding function can be added in-line in probably less than 100 lines of code.

Just did this. I got the implementation in in 60 lines, though it's over double that with tests.

dead10ck (approved):
I'm not familiar with this level of terminal interaction, nor OSC 52, so my review isn't worth much, but I don't see any issues with this code. I do question the reluctance to bring in a base64 crate, but it looks like the work is already done.

archseer:
Would this also obsolete https://github.com/helix-editor/helix/pull/1343 ?

groves:
> Would this also obsolete #1343 ?

I think maybe? It's complicated and I'm not a `tmux` user, so I'm just basing this answer on reading their docs and may be getting it wrong.

From [tmux's clipboard docs](https://github.com/tmux/tmux/wiki/Clipboard), it sounds like if `set-clipboard on` is set for tmux, tmux would pass the OSC 52 Helix emits through to the surrounding terminal and set the system clipboard.

As of tmux 2.6 from a few years ago, the default for `set-clipboard` is `external`. That means that tmux itself can emit OSC 52 codes but that it eats them from applications to keep them from modifying the system clipboard. That means tmux users would need to configure `set-clipboard` to use this.

In this PR, we only use OSC 52 as a fallback if we don't find a "native" clipboard provider. We look for the `TMUX` env var and `tmux` executable and use that as a native provider if we find them. That means we wouldn't emit OSC 52 in that case.

Putting it all together, I think that means this could handle the clipboard setting part of #1343 if the user has set `set-clipboard on` in their tmux and we use OSC 52 in preference to setting tmux's internal paste buffers.

I think it might make sense to add an option to use OSC 52 regardless of what's available locally. That way a tmux user that wants to configure that can, and people who have X or Wayland installed on a server they're SSHing into can choose to use OSC 52 if they're not forwarding X. I was loathe to add configuration, but this seems like something that's impossible to guess from just what's on the system.

helix-view/src/base64/mod.rs
```diff
@@ -0,0 +1,135 @@
+use std::ops::{BitAnd, BitOr, Shl, Shr};
+
+const PAD_BYTE: u8 = b'=';
```
the-mikedavis, line 3: It looks like there are some compilation warnings since this ends up not being used on Windows

helix-view/src/base64/LICENSE-MIT
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
the-mikedavis, line 1: Rather than separate files for the note and license, can we inline this and `helix-view/src/base64/README.md` into the top of the base64 module and move that into `helix-view/src/base64.rs`?

helix-view/src/base64/mod.rs
the-mikedavis, line 3: It looks like there are a few compliation warnings for Windows since base64 isn't called when compiled on Windows. Making this module `pub` is ok I think

the-mikedavis (approved):
I think this looks great 😀

Sorry for the delay (and the base64 side-quest 😛)

groves:
Was fun to learn more about base64!

PR #3222, opened 2022-07-28
Hide cursorline when into SelectMode

erasin:
I  think we don't need double background colors when we select some  text .  Because it hinders the focus of reading.
Jul-28-2022 14-12-03

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -114,7 +114,7 @@ impl EditorView {
             }
         }
 
-        if is_focused && editor.config().cursorline {
+        if is_focused && doc.mode != Mode::Select && editor.config().cursorline {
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I prefer the original behavior. I think this should be fixed in your theme to use different colors for selections and cursorline.

erasin:
It can be solved by modifying the color of the theme. I see that other themes will also have such superimposed backgrounds, such as Nord , solarized_ dark.

 I use GitHub web editor and local editor vscode, both of which are the hide cursorline with select text. 
 Less bg focus, may be is better

the-mikedavis:
There is always a selection in helix though and it's easy to create a large selection without being in select-mode (`X` for example).

This behavior may make sense for VSCode but selections are much more core to Helix.

erasin:
thanks , I will change my theme.

PR #3226, opened 2022-07-28
Fix cursorline and selection themes for onelight

erasin:


archseer:
Can you show a preview of the differences?

erasin:
before
before

after

after

the-mikedavis (approved):

PR #3233, opened 2022-07-28
Add bracketed paste

groves:
This adds bracketed paste support to fix https://github.com/helix-editor/helix/issues/2239. It's based on [this PR on crossterm](https://github.com/crossterm-rs/crossterm/pull/693), so that will need to land and be released before this can be merged.

I haven't done any serious benchmarking, but I tried the [Moby Dick Workout](https://www.hogbaysoftware.com/posts/moby-dick-workout/). Pasting with this enabled took maybe a tenth of a second to get the whole meg of book in there. Without this, I think it would've taken a day or so.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -29,7 +29,10 @@ use std::{
 use anyhow::{Context, Error};
 
 use crossterm::{
-    event::{DisableMouseCapture, EnableMouseCapture, Event as CrosstermEvent},
+    event::{
+        DisableBracketedPaste, DisableMouseCapture, EnableBracketedPaste, EnableMouseCapture,
+        Event as CrosstermEvent,
+    },
     execute, terminal,
     tty::IsTty,
 };
@@ -425,14 +428,13 @@ impl Application {
             scroll: None,
         };
         // Handle key events
-        let should_redraw = match event {
-            Ok(CrosstermEvent::Resize(width, height)) => {
+        let should_redraw = match event.unwrap() {
+            CrosstermEvent::Resize(width, height) => {
                 self.compositor.resize(width, height);
                 self.compositor
-                    .handle_event(Event::Resize(width, height), &mut cx)
+                    .handle_event(&Event::Resize(width, height), &mut cx)
             }
-            Ok(event) => self.compositor.handle_event(event.into(), &mut cx),
-            Err(x) => panic!("{}", x),
+            event => self.compositor.handle_event(&event.into(), &mut cx),
         };
 
         if should_redraw && !self.editor.should_close() {
@@ -788,7 +790,7 @@ impl Application {
     async fn claim_term(&mut self) -> Result<(), Error> {
         terminal::enable_raw_mode()?;
         let mut stdout = stdout();
-        execute!(stdout, terminal::EnterAlternateScreen)?;
+        execute!(stdout, terminal::EnterAlternateScreen, EnableBracketedPaste)?;
         execute!(stdout, terminal::Clear(terminal::ClearType::All))?;
         if self.config.load().editor.mouse {
             execute!(stdout, EnableMouseCapture)?;
@@ -821,7 +823,11 @@ impl Application {
             // probably not a good idea to `unwrap()` inside a panic handler.
             // So we just ignore the `Result`s.
             let _ = execute!(std::io::stdout(), DisableMouseCapture);
-            let _ = execute!(std::io::stdout(), terminal::LeaveAlternateScreen);
+            let _ = execute!(
+                std::io::stdout(),
+                terminal::LeaveAlternateScreen,
+                DisableBracketedPaste
+            );
             let _ = terminal::disable_raw_mode();
             hook(info);
         }));
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3369,13 +3369,7 @@ enum Paste {
     Cursor,
 }
 
-fn paste_impl(
-    values: &[String],
-    doc: &mut Document,
-    view: &View,
-    action: Paste,
-    count: usize,
-) -> Option<Transaction> {
+fn paste_impl(values: &[String], doc: &mut Document, view: &View, action: Paste, count: usize) {
     let repeat = std::iter::repeat(
         values
             .last()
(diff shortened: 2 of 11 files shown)


helix-term/src/application.rs
```diff
@@ -419,21 +421,27 @@ impl Application {
...
+            Ok(input) => match input {
+                Input::Event(Event::Resize(width, height)) => {
+                    self.compositor.resize(width, height);
+                    self.compositor
+                        .handle_event(Event::Resize(width, height), &mut cx)
+                }
+                Input::Event(event) => self.compositor.handle_event(event, &mut cx),
+                Input::Paste(data) => {
+                    let (view, doc) = current!(self.editor);
+                    crate::commands::paste_bracketed_value(data, doc, view);
```
groves, line 441: Not sure where best to stick this call. Doesn't seem like something that should be on compositor?

helix-term/src/ui/editor.rs
```diff
@@ -1109,6 +1109,7 @@ impl Component for EditorView {
         };
 
         match event {
+            Event::FocusGained | Event::FocusLost => todo!(),
```
groves, line 1112: This is a new event on crossterm main that's handled in https://github.com/helix-editor/helix/pull/3178.

helix-term/src/commands.rs
```diff
@@ -3417,8 +3411,17 @@ fn paste_impl(
         };
         (pos, pos, values.next())
     });
+    doc.apply(&transaction, view.id);
+}
 
-    Some(transaction)
+pub(crate) fn paste_bracketed_value(cx: &mut Context, contents: String) {
```
groves, line 3417: This doesn't function as an actual command as I didn't see a good way to pass the contents through the Context. I think it's nicer to have it as a standalone function, but maybe I'm missing a way to make it work more like a regular command?

Because it's not a command, it doesn't set editor.last_insert and won't be repeated with `.`. Since you can spam your paste key as easily as `.`, I didn't think that was a big deal.

helix-term/src/commands.rs
```diff
@@ -3428,18 +3431,11 @@ fn paste_clipboard_impl(
...
     let (view, doc) = current!(editor);
-
-    match editor
-        .clipboard_provider
-        .get_contents(clipboard_type)
-        .map(|contents| paste_impl(&[contents], doc, view, action, count))
-    {
-        Ok(Some(transaction)) => {
-            doc.apply(&transaction, view.id);
-            doc.append_changes_to_history(view.id);
```
groves, line 3439: This isn't needed here as it's done in editor's event handling.

helix-term/src/ui/editor.rs
```diff
@@ -1126,6 +1138,13 @@ impl Component for EditorView {
         };
 
         match event {
+            Event::Paste(contents) => {
+                cx.count = cx.editor.count;
```
groves, line 1142: I felt like it made more sense to use a pending count here, though I could also see leaving it for an upcoming command.

helix-view/src/input.rs
```diff
@@ -6,10 +6,13 @@ use std::fmt;
 
 pub use crate::keyboard::{KeyCode, KeyModifiers};
 
-#[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
```
groves, line 9: Because the paste data is on the heap, we have to lose this Copy. That's why there's all this introduction of references through the rest of the PR.

sfarq:
Is there a need for `Space-p` after this is merged?

archseer:
Yes, not all terminals support bracketed paste (I also prefer `space+p` over `ctrl+v`)

archseer:
Looks like Windows sends paste events regardless, so we need to get this merged before `22.08` is released (`ctrl+v` panics on windows)

groves:
That's frustrating that Windows sends bracketed pastes regardless. This is ready to go as far as I know.

If this PR needs additional work, we could also disable the `bracketed-paste` feature in crossterm. That would turn off the parsing entirely. I'd prefer getting the PR in, naturally 😅

helix-term/src/ui/editor.rs
```diff
@@ -1205,16 +1224,9 @@ impl Component for EditorView {
...
-                let config = cx.editor.config();
-                let (view, doc) = current!(cx.editor);
-                view.ensure_cursor_in_view(doc, config.scrolloff);
-
-                // Store a history state if not in insert mode. This also takes care of
-                // committing changes when leaving insert mode.
-                if doc.mode() != Mode::Insert {
-                    doc.append_changes_to_history(view.id);
-                }
+                self.complete_change(&mut cx);
```
the-mikedavis, line 1227: I think this code should not be abstracted between key events and paste: I think it would be better if the code were duplicated. We want to ensure the cursor is in view and save the pasted change to history when pasting in any mode, and we shouldn't need to execute any mode transition hooks. But this block is currently wrong for key events and needs some modification since it may use the wrong document and view when switching between views (https://github.com/helix-editor/helix/pull/3508).

groves: Makes sense to me on the separation between key events and paste. I've changed this to duplicate the code instead of sharing it.

I'm not sure I follow on "we want to [...] save the pasted change to history when pasting in any mode" though. To me, if I'm in insert and I hit my OS paste key, I expect that paste to be part of the ongoing insert. If I edit the inserted text from pasting and then exit insert mode, I expect undo to get the whole insert, not the paste and edit separately. Maybe I'm missing something though?

the-mikedavis: Oh hmm yeah I'm not sure what I was thinking. Just when mode isn't insert we should be saving history 👍

helix-term/src/ui/editor.rs
```diff
@@ -976,15 +976,16 @@ impl EditorView {
                         doc.set_selection(view_id, Selection::point(pos));
                     }
 
-                    editor.focus(view_id);
+                    editor.tree.focus = view_id;
```
the-mikedavis, line 979: It looks like the changes of https://github.com/helix-editor/helix/commit/e4c9d4082a139aac3aea4506918171b96e81f5b9 are reverted here. Is that a mistake from a merge conflict or intentional?

groves: That's really odd... Definitely not intentional, but I can't imagine how I would've blown away the changes in a conflict.

In any case, I've restored those changes and I don't see anything else unintentional in the remaining changes. Thanks for catching that!

archseer (approved):
Thanks for working on this! It fixes a longstanding papercut for new users that didn't know about `space+p`.

PR #3234, opened 2022-07-28
feat: xtask themelint

AlexanderBrevig:
> This is meant as a basis of discussion, and a WIP / POC implementation.

## TL;DR - content of PR
1. Refactor `./xtask/src/main.rs` into separate modules
2. Add `themelint` task 

`themelint` will check to see that every "required"* key is present for a theme.

*required is not really defined, I've just used the `book/src/themes.md` as a guide.

## Try it
`cargo xtask themelint` will lint every theme, you can process the output with `jq` by `cargo xtask themelint 2> /dev/null | jq 'keys'` for instance, to get a list of all themes with issues.

Drill down to one theme with `cargo xtask themelint onedark` or `cargo xtask themelint 2> /dev/null | jq '.gruvbox'`

## Status
There's only a few themes that has no "errors". 40 of 45 themes has missing fields. This may be because I am too eagerly checking. 

I hope you want to help and try it out. It's already been helpful for me for playing with custom themes.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -252,4 +252,16 @@ These scopes are used for theming the editor interface.
 | `diagnostic.warning`      | Diagnostics warning (editing area)             |
 | `diagnostic.error`        | Diagnostics error (editing area)               |
 
+You can check compliance to spec with
+
+```shell
+cargo xtask themelint onedark # replace onedark with <name>
+```
+
+Find all themes with issues with
+
+```shell
+cargo xtask themelint 2> /dev/null | jq 'keys'
+```
+
 [editor-section]: ./configuration.md#editor-section
--- a/xtask/src/helpers.rs
+++ b/xtask/src/helpers.rs
@@ -1,0 +1,58 @@
+use std::path::{Path, PathBuf};
+
+use crate::paths;
+use helix_core::syntax::Configuration as LangConfig;
+use helix_term::health::TsFeature;
+
+/// Get the list of languages that support a particular tree-sitter
+/// based feature.
+pub fn ts_lang_support(feat: TsFeature) -> Vec<String> {
+    let queries_dir = paths::ts_queries();
+
+    find_files(&queries_dir, feat.runtime_filename())
+        .iter()
+        .map(|f| {
+            // .../helix/runtime/queries/python/highlights.scm
+            let tail = f.strip_prefix(&queries_dir).unwrap(); // python/highlights.scm
+            let lang = tail.components().next().unwrap(); // python
+            lang.as_os_str().to_string_lossy().to_string()
+        })
+        .collect()
+}
+
+/// Get the list of languages that have any form of tree-sitter
+/// queries defined in the runtime directory.
+pub fn langs_with_ts_queries() -> Vec<String> {
+    std::fs::read_dir(paths::ts_queries())
+        .unwrap()
+        .filter_map(|entry| {
+            let entry = entry.ok()?;
+            entry
+                .file_type()
+                .ok()?
+                .is_dir()
+                .then(|| entry.file_name().to_string_lossy().to_string())
+        })
+        .collect()
+}
+
+// naive implementation, but suffices for our needs
+pub fn find_files(dir: &Path, filename: &str) -> Vec<PathBuf> {
+    std::fs::read_dir(dir)
+        .unwrap()
+        .filter_map(|entry| {
+            let path = entry.ok()?.path();
+            if path.is_dir() {
+                Some(find_files(&path, filename))
+            } else {
+                (path.file_name()?.to_string_lossy() == filename).then(|| vec![path])
+            }
+        })
+        .flatten()
+        .collect()
+}
+
+pub fn lang_config() -> LangConfig {
+    let bytes = std::fs::read(paths::lang_config()).unwrap();
+    toml::from_slice(&bytes).unwrap()
+}
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -1,220 +1,28 @@
+mod docgen;
+mod helpers;
+mod paths;
+mod themelint;
+
 use std::{env, error::Error};
 
 type DynError = Box<dyn Error>;
 
-pub mod helpers {
-    use std::path::{Path, PathBuf};
-
-    use crate::path;
-    use helix_core::syntax::Configuration as LangConfig;
-    use helix_term::health::TsFeature;
-
-    /// Get the list of languages that support a particular tree-sitter
-    /// based feature.
-    pub fn ts_lang_support(feat: TsFeature) -> Vec<String> {
-        let queries_dir = path::ts_queries();
-
-        find_files(&queries_dir, feat.runtime_filename())
-            .iter()
-            .map(|f| {
-                // .../helix/runtime/queries/python/highlights.scm
-                let tail = f.strip_prefix(&queries_dir).unwrap(); // python/highlights.scm
-                let lang = tail.components().next().unwrap(); // python
-                lang.as_os_str().to_string_lossy().to_string()
-            })
-            .collect()
-    }
-
-    /// Get the list of languages that have any form of tree-sitter
-    /// queries defined in the runtime directory.
-    pub fn langs_with_ts_queries() -> Vec<String> {
-        std::fs::read_dir(path::ts_queries())
-            .unwrap()
-            .filter_map(|entry| {
-                let entry = entry.ok()?;
-                entry
-                    .file_type()
-                    .ok()?
-                    .is_dir()
-                    .then(|| entry.file_name().to_string_lossy().to_string())
-            })
-            .collect()
-    }
-
-    // naive implementation, but suffices for our needs
-    pub fn find_files(dir: &Path, filename: &str) -> Vec<PathBuf> {
-        std::fs::read_dir(dir)
-            .unwrap()
-            .filter_map(|entry| {
-                let path = entry.ok()?.path();
-                if path.is_dir() {
-                    Some(find_files(&path, filename))
-                } else {
-                    (path.file_name()?.to_string_lossy() == filename).then(|| vec![path])
-                }
-            })
-            .flatten()
-            .collect()
-    }
-
-    pub fn lang_config() -> LangConfig {
-        let bytes = std::fs::read(path::lang_config()).unwrap();
-        toml::from_slice(&bytes).unwrap()
-    }
-}
-
-pub mod md_gen {
-    use crate::DynError;
-
-    use crate::helpers;
-    use crate::path;
-    use helix_term::commands::TYPABLE_COMMAND_LIST;
-    use helix_term::health::TsFeature;
-    use std::fs;
-
-    pub const TYPABLE_COMMANDS_MD_OUTPUT: &str = "typable-cmd.md";
-    pub const LANG_SUPPORT_MD_OUTPUT: &str = "lang-support.md";
-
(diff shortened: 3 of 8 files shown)


AlexanderBrevig:
Will need to handle the `base16_transparent` case somehow, as it relies on things not being set / falling back on terminal.

Must be done before an eventual PR-flow inclusion.

xtask/src/themelint.rs
```diff
@@ -0,0 +1,172 @@
...
+    fg: Option<String>,
+    bg: Option<String>,
+    check_both: bool,
+}
+
+// Placed in an fn here, so it's the first thing you see
+fn get_rules() -> Vec<Rule> {
+    vec![
+        // Check for ui.selection, which is required
+        Rule::has_either("ui.selection".into()),
```
pickfire, line 16: I believe we can just do `&'static str` instead of `String` given that all of these we have them when we compile?

xtask/src/themelint.rs
```diff
@@ -0,0 +1,172 @@
...
+            .map(|m| {
+                let mut msg = file.clone();
+                msg.push_str(".");
+                msg.push_str(m);
+                msg
+            })
+            .collect::<Vec<String>>()
+            .join(" ")
+            .into();
+        println!("{}", message.replace(" ", "\n"));
```
pickfire, line 135: If I am a new contributor to theme, just seeing
```
everforest_light.ui.cursorline.primary.bg
everforest_light.ui.virtual.ruler
```
I have no idea what I did wrong there. At least some explanation on why is that error there is better, a better version would be like rust sort of error where it point you out the line and span, but showing at least the error is a good start.

xtask/src/themelint.rs
```diff
@@ -0,0 +1,172 @@
...
+    let files = std::fs::read_dir(path::themes())
+        .unwrap()
+        .filter_map(|entry| {
+            let path = entry.ok()?.path();
+            let name = path.file_name()?.to_string_lossy().to_string();
+            if path.is_dir() || name.contains("README") {
+                None
+            } else {
+                Some(name)
+            }
```
pickfire, line 152: Why not just filter for `.toml`?

xtask/src/themelint.rs
```diff
@@ -0,0 +1,172 @@
...
+        })
+        .collect::<Vec<String>>();
```
pickfire, line 154: Rather than writing these here, we can probably reuse helix-loader helper functions. As seen in `helix-view/src/theme.rs`
```rust
    pub fn read_names(path: &Path) -> Vec<String> {
        std::fs::read_dir(path)
            .map(|entries| {
                entries
                    .filter_map(|entry| {
                        let entry = entry.ok()?;
                        let path = entry.path();
                        (path.extension()? == "toml")
                            .then(|| path.file_stem().unwrap().to_string_lossy().into_owned())
                    })
                    .collect()
            })
            .unwrap_or_default()
    }
```

xtask/src/helpers.rs
```diff
@@ -0,0 +1,42 @@
...
+pub fn find_files(dir: &Path, filename: &str) -> Vec<PathBuf> {
+    std::fs::read_dir(dir)
+        .unwrap()
+        .filter_map(|entry| {
+            let path = entry.ok()?.path();
+            if path.is_dir() {
+                Some(find_files(&path, filename))
+            } else {
+                (path.file_name()?.to_string_lossy() == filename).then(|| vec![path])
+            }
```
pickfire, line 33: `bool.then` makes it less readable, maybe if would make it easier to read

```suggestion
            if path.is_dir() {
                Some(find_files(&path, filename))
            } else if path.file_name()?.to_string_lossy() == filename {
                Some(vec![path])
            } else {
                None
            }
```

xtask/src/themelint.rs
AlexanderBrevig:
```sh
gruvbox:"ui.cursorline.primary"={"bg"="missing"}
gruvbox:"ui.virtual.ruler"="missing"
```
This is better, yes? Now you get copy pastable "code" after the theme key, which says missing.
If you've got ideas to improve it further (maybe adding a text to each rule with an actual feedback?) I'm happy to implement!

xtask/src/helpers.rs
AlexanderBrevig: This is just a copy paste, but I wholeheartedly agree that it's better. I'll include it in this PR.

AlexanderBrevig:
thanks again @pickfire for your work on this :)

xtask/src/themelint.rs
```diff
@@ -0,0 +1,156 @@
...
+        // Check for editor.rulers
+        Rule::has_either("ui.virtual.ruler".into()),
+        // Check for menus and prompts
+        Rule::has_both("ui.menu".into()),
+        Rule::has_both("ui.help".into()),
+        Rule::has_bg("ui.popup".into()),
+        Rule::has_either("ui.window".into()),
+        // Check for visible cursor
+        Rule::has_bg("ui.cursor.primary".into()),
+        Rule::has_either("ui.cursor.match".into()),
```
pickfire, line 39: `.into()` is now unnecessary

xtask/src/themelint.rs
```diff
@@ -0,0 +1,156 @@
...
+            if !found_fg {
+                messages.push(format!("\"{}\"={{\"fg\"=\"missing\"}}", fg_name.clone()));
+            }
+            if !found_bg {
+                messages.push(format!("\"{}\"={{\"bg\"=\"missing\"}}", bg_name.clone()));
+            }
+        } else {
+            if !found_fg && !found_bg {
+                messages.push(format!("\"{}\"=\"missing\"", fg_name))
+            }
```
pickfire, line 99: All the `.clone()` is now unnecessary since `&'static str` should have `Copy`

book/src/themes.md
```diff
@@ -254,4 +254,10 @@ These scopes are used for theming the editor interface.
 | `diagnostic.warning`      | Diagnostics warning (editing area)             |
 | `diagnostic.error`        | Diagnostics error (editing area)               |
 
+You can check compliance to spec with
+
+```shell
+cargo xtask themelint onedark # replace onedark with <name>
```
pickfire, line 260:
```suggestion
cargo xtask themelint onedark  # replace onedark with <name>
```

xtask/src/main.rs
```diff
@@ -1,229 +1,38 @@
...
-
-    pub fn lang_config() -> PathBuf {
-        project_root().join("languages.toml")
-    }
-}
-
 pub mod tasks {
-    use crate::md_gen;
+    use crate::docgen::*;
+    use crate::themelint::*;
```
pickfire, line 12: Can we use explicit import here? It's just few import.

xtask/src/themelint.rs
```diff
@@ -0,0 +1,156 @@
...
+        let mut found_fg = true;
+        let mut found_bg = true;
+        let mut fg_name = "";
+        let mut bg_name = "";
+        if let Some(fg) = &self.fg {
+            fg_name = fg;
+            if theme.get(fg).fg.is_none() && theme.get(fg).add_modifier == Modifier::empty() {
+                found_fg = false;
+            }
+        }
```
pickfire, line 82: I think a better way instead of doing this logic (which seemed weird to me), is to move this part out into a function, same goes for bg, then do
```
if self.check_both {
    if self.found_fg(theme, messages) { ... }
    if self.found_bg(theme, messages) { ... }
} else if !self.found_fg(theme, messages) && !self.found_bg(theme, messages) {
    ...
}
```

xtask/src/themelint.rs
pickfire: Better slightly but still can be improved, I imagine something like
```
Missing `bg` for gruvbox `ui.cursorline.primary`
```
Also, not sure why we want to split `fg` and `bg` reporting when they can be in the same line, like `Missing bg and fg for ...`.

AlexanderBrevig: Marking this as resolved for now```shell
gruvbox: missing `fg` and `bg` for `ui.statusline`
gruvbox: missing entry for `ui.virtual.ruler`
```
Marking this s resolved for now.

File: xtask/src/themelint.rs (hunk 10 above)
Comment (AlexanderBrevig) [reply xtask/src/themelint.rs]: I've refactored it a bit, but I've kept the handling of messages to `validate` so `found_Xg` only deals with the checking of existence.

File: xtask/src/themelint.rs [hunk 11]
```diff
@@ -69,33 +69,56 @@ impl Rule {
...
-                found_bg = false;
-            }
+        return true;
+    }
+    fn found_fg(&self, theme: &Theme) -> bool {
+        return Rule::found_impl(theme, self.fg);
+    }
+    fn found_bg(&self, theme: &Theme) -> bool {
+        return Rule::found_impl(theme, self.bg);
+    }
```
pickfire, line 98: Please check what clippy say, it don't need `return` here.

runtime/queries/pascal/highlights.scm
```diff
@@ -0,0 +1,382 @@
...
+	(kWhile)
+	(kRepeat)
+	(kUntil)
+] @keyword.control.repeat
+
+[
+	(kIf)
+	(kThen)
+	(kElse)
+] @keyword.control.conditional
```
pickfire, line 382: I don't think you want to have the pascal file checked in with the same PR as adding themelint.

xtask/src/themelint.rs
AlexanderBrevig: did you accidentally review and old revision? this got refactored with https://github.com/helix-editor/helix/commit/464d7c541823cda2cc3016e9f3edfa84c3a5dca2 (because it was a bug as fg and bg also need to be retrieved from the underlying Style).

Latest commit should be ok for review now. I moved `validate` to static scope in order to have a similar style between `Existence` and `Difference` rules. The validations are now `check_difference` and `check_existence`.

AlexanderBrevig:
Ready for review again :)

xtask/src/themelint.rs
```diff
@@ -0,0 +1,194 @@
...
+    let files_count = files.len();
+    files.into_iter().for_each(|path| {
+        if let Err(err) = lint(path.replace(".toml", "")) {
+            let errs: i32 = err.to_string().parse().expect("Errors must be integral");
+            errors.push(errs)
+        }
+    });
+    println!("{} of {} themes had issues", errors.len(), files_count);
+    if !errors.is_empty() {
+        Err(errors.iter().sum::<i32>().to_string().into())
```
pickfire, line 190: This seemed pretty much useless to me such that it returns some numbers which may confuse users.
```
...
varua: missing `fg` or `bg` for `ui.virtual.ruler`
41 of 60 themes had issues
Error: "95"
```

I have no idea what `95` means when I am looking at the output, if anything `41 of 60 themes had issues` or `lint issues found in theme` may be a better error, in any case I prefer there isn't the `Error: ` line but it's fine to have one as long as it is somewhat useful.

Also, the count doesn't seemed necessary as well in any place, I don't recall seeing rustc or similar linters count how many lint errors are there.

AlexanderBrevig: The printout of error is because xtask main is or Result return type I think? I totally agree that the total count printout makes little sense. I've moved the `x of y themes had issues` printout to be the message of the error.

pickfire (approved):
Looks good to me! Thanks for working on this, this will be very helpful for people creating new themes and keeping themes up to date.

I don't think we need to put in CI so fast yet but at least other people can run it, and we can keep it in xtask for a while and see how it work out.

PR #3239, opened 2022-07-29
Add centered view mode

### LINKED ISSUE #3203: Feature Request: Centered code mode (like intellij's distraction free mode)
Hi, I find Intellij's "distaction free mode" really interesting, especially the fact of centering the code on the screen. I find it much less tiring than having to constantly look at the left side of the screen, which happens quite often with helix when you have only one window opened, because there is no sidebar which in an editor like VS Code center the code even if it's not the intention at the beginning.

So it could be cool to having an option in config file to allow the user to choose to center his code in the window, maybe with exception like when we are in split-screen.

an example of the feature (in vscode):
image

ldubos:
Resolves #3203

Add a new `toggle_centered_view` command in Window mode and a `centered-views` to the editor config, change the behaviour of the `recalculate` function in `Tree` to center a `View` when it has the `is_centered` set to `true` to center it in the available space if the `View` isn't a child of a vertical container with more than 1 children.

I found that is the "standard" behaviour of the "centered mode" in most editors I use.

image

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -274,6 +274,8 @@ pub struct Config {
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
     pub soft_wrap: SoftWrap,
+    /// Whether to center views by default, Defaults to `false`.
+    pub centered_views: bool,
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
@@ -772,6 +774,7 @@ impl Default for Config {
             indent_guides: IndentGuidesConfig::default(),
             color_modes: false,
             soft_wrap: SoftWrap::default(),
+            centered_views: false,
         }
     }
 }
@@ -1241,7 +1244,13 @@ impl Editor {
                     .try_get(self.tree.focus)
                     .filter(|v| id == v.doc) // Different Document
                     .cloned()
-                    .unwrap_or_else(|| View::new(id, self.config().gutters.clone()));
+                    .unwrap_or_else(|| {
+                        View::new(
+                            id,
+                            self.config().gutters.clone(),
+                            self.config().centered_views,
+                        )
+                    });
                 let view_id = self.tree.split(
                     view,
                     match action {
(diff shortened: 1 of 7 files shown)


ldubos:
@CptPotato for the moment we can't define the width unfortunately I don't really know how I can make it configurable and updatable during a session, it seems really complicated, also I don't take the gutters_offset so the width = the total width not the inner width.

and I calculated the centered width like this: `centered_width = min(max(available_width / 2, 120), available_width)` that means if the available_width / 2 < 120 we set it to 120 and if the result is greater than the available_width we set it to available_width.

So you always have at least 120 columns (with gutters) in centered mode or the available_width / 2.

ShalokShalom:
@ldubos I love this PR, and see that it is currently prevented from being merged by conflicts. 

Do you see any chance to resolve them?

helix-view/src/editor.rs
```diff
@@ -274,6 +274,8 @@ pub struct Config {
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
     pub soft_wrap: SoftWrap,
+    /// Whether to center views by default, Defaults to `false`.
+    pub centered_views: bool,
```
archseer, line 278: I'd remove the config option since it seems unlikely someone would want this always on.

helix-view/src/tree.rs
```diff
@@ -342,16 +342,35 @@ impl Tree {
...
+                        && (parent_child_count <= 1 || parent_layout != Layout::Vertical)
+                    {
+                        let width = std::cmp::min(std::cmp::max(area.width / 2, 120), area.width);
+                        let area =
+                            Rect::new(area.width / 2 - width / 2, area.y, width, area.height);
+
+                        view.area = area;
+                    } else {
+                        view.area = area;
+                    }
```
archseer, line 373: This gives the view the entire area but doesn't affect any of the other views. This means that you're relying on view ordering (this view should be the last one rendered) otherwise other views might get rendered over it. We also pay the performance cost of rendering all other views even though they're not on screen. (Also what happens if I change focus to a different view but this one is still centered?)

I'd rather see this done via a `zoom` bool flag on the tree. If set to `true` then the editor should just render the currently focused node and ignore the rest of the tree. The reason we haven't done this yet is because this will likely impact a bunch of other viewport calculations.

archseer (changes requested):

helix-view/src/tree.rs
ldubos: Hi, thanks for the feedback, but... to be honest I made this PR 7 months ago, so I don't remember at all why I made things in this way, and I don't have sufficent spare time to invest in this PR anymore :/ if someone is motivated enough to take over my work (or redoing it from scratch ^^') it would be better

David-Else:
It would be great if someone could take over this pull request, it is so needed!

pascalkuthe:
closing in favor of https://github.com/helix-editor/helix/pull/9838

PR #3240, opened 2022-07-29
Add cwd parameter which is not optional anymore

danyspin97:
cwd field has been changed in #2959 by making it a required field.
However, when starting the debugging session, cwd is not set, causing
an error in the deserialization part. Manually set cwd to the current
directory as a quick fix.

### DIFF
--- a/helix-term/src/commands/dap.rs
+++ b/helix-term/src/commands/dap.rs
@@ -216,6 +216,8 @@ pub fn dap_start_impl(
         }
     }
 
+    args.insert("cwd", to_value(std::env::current_dir().unwrap())?);
+
     let args = to_value(args).unwrap();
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Trying this locally it works for me when using DAP with Erlang. I don't see the "error in the deserialization part" you refer to on master though. Is that coming from debugging Rust?

danyspin97:
Yea, I am debugging a Rust program. The error is [here](https://github.com/helix-editor/helix/blob/master/helix-view/src/handlers/dap.rs#L286):

```rust
 let arguments: dap::requests::RunInTerminalArguments =
                        serde_json::from_value(request.arguments.unwrap_or_default()).unwrap();
```

`serde_json` is deserializing `request.arguments` to `RunInTerminalArguments`, but `cwd` is not set and it's required.

the-mikedavis (approved):
Makes sense, thanks for the fix!

PR #3242, opened 2022-07-29
Make matching brace higlighting configurable

### LINKED ISSUE #3228: Matching bracket only highlighted for primary with multi cursor
When using multiple cursors (like with `C`) with the cursor over brackets, the matching bracket is only highlighted for the primary selection.
image

Not sure if this is intended behavior or a bug...

A-Walrus:
Resolves  #3228 by making matching brace highlighting configurable:
`none`, `all`, `primary`, defaulting to `all`.

New default behavior:
image

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -208,48 +208,53 @@ These scopes are used for theming the editor interface.
       - `hover` - for hover popup ui
 
 
-| Key                       | Notes                                          |
-| ---                       | ---                                            |
-| `ui.background`           |                                                |
-| `ui.background.separator` | Picker separator below input line              |
-| `ui.cursor`               |                                                |
-| `ui.cursor.insert`        |                                                |
-| `ui.cursor.select`        |                                                |
-| `ui.cursor.match`         | Matching bracket etc.                          |
-| `ui.cursor.primary`       | Cursor with primary selection                  |
-| `ui.linenr`               | Line numbers                                   |
-| `ui.linenr.selected`      | Line number for the line the cursor is on      |
-| `ui.statusline`           | Statusline                                     |
-| `ui.statusline.inactive`  | Statusline (unfocused document)                |
-| `ui.statusline.normal`    | Statusline mode during normal mode ([only if `editor.color-modes` is enabled][editor-section]) |
-| `ui.statusline.insert`    | Statusline mode during insert mode ([only if `editor.color-modes` is enabled][editor-section]) |
-| `ui.statusline.select`    | Statusline mode during select mode ([only if `editor.color-modes` is enabled][editor-section]) |
-| `ui.statusline.separator` | Separator character in statusline              |
-| `ui.popup`                | Documentation popups (e.g space-k)             |
-| `ui.popup.info`           | Prompt for multiple key options                |
-| `ui.window`               | Border lines separating splits                 |
-| `ui.help`                 | Description box for commands                   |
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Instead of having this configurable with a config option it could just use theme rules: `ui.cursor.match.primary` themes the primary match and `ui.cursor.match.secondary` the secondary cursor matches

book/src/themes.md
```diff
@@ -208,48 +208,53 @@ These scopes are used for theming the editor interface.
...
+| `ui.cursor.select`          |                                                |
+| `ui.cursor.match`           | Matching bracket etc.                          |
+| `ui.cursor.match.primary`   | Matching bracket for primary selection.        |
+| `ui.cursor.match.secondary` | Matching bracket for secondary selections.     |
+| `ui.cursor.primary`         | Cursor with primary selection                  |
+| `ui.linenr`                 | Line numbers                                   |
+| `ui.linenr.selected`        | Line number for the line the cursor is on      |
+| `ui.statusline`             | Statusline                                     |
+| `ui.statusline.inactive`    | Statusline (unfocused document)                |
+| `ui.statusline.normal`      | Statusline mode during normal mode ([only if `editor.color-modes` is enabled][editor-section]) |
```
the-mikedavis, line 229: if we're reformatting this table, might as well extend the ending `|` to the longest entry

helix-term/src/ui/editor.rs
```diff
@@ -569,24 +569,43 @@ impl EditorView {
...
-                            .add_modifier(Modifier::REVERSED)
-                            .add_modifier(Modifier::DIM)
-                    });
+            let style_primary = theme
+                .try_get("ui.cursor.match.primary")
+                .unwrap_or(default_style);
+
+            let style_secondary = theme
+                .try_get("ui.cursor.match.secondary")
+                .unwrap_or(default_style);
```
the-mikedavis, line 583: Instead of falling back to the default_style I think we should not have any fallback behavior for `ui.cursor.match.secondary` so that you can _not_ highlight secondary cursor matches

A-Walrus: Not highlighting the secondary cursor matches is still possible with `"ui.cursor.match.secondary" = {}`. So either way both are possible, the question is what should be the default. Personally I think highlighting secondary matches is a sensible default behavior, but it's up to you :)

the-mikedavis:
I think secondary cursor match highlighting is pretty noisy. I only use the primary match in cases where I'm not sure what nesting level I'm on and if it were toggle-able I would probably only enable showing the primary match sparingly. I don't run into the same use-case for secondary matches so I most likely would not enable this in my theme.

Maybe this _should_ be configurable in editor config rather than theme? It could be set to `all`, `primary` or `none` for showing these and still have these `ui.cursor.match.{primary,secondary}` scopes for theming. What do others think?

Sorry for the bike-shedding 😅

AlexanderBrevig:
As a new user (not that many days ago) my knee jerk reaction to not seeing color modes after enabling it made me think it was a bug with Helix. In fact is was just my theme not being up to date, and I would call that a bug too (fixed in a PR of course :) ).

I feel (rather strongly) that a Helix theme should accommodate all Helix features. Opting in/out of feature x should not imply editing a theme or changing a theme imho.

My `0b10` 😊

pascalkuthe:
clsoing this PR as stale. I also dont think we will be going forward with this (see mikes comment above).

Thank you for contributing!

PR #3244, opened 2022-07-29
Add rust-version (MSRV) to helix-term package

danyspin97:
Adding rust-version ensure a check at compile that the compiler version can build helix. I have used [`cargo-msrv`](https://github.com/foresterre/cargo-msrv) to retrieve this value and it can be checked using `cargo msrv verify`.

This value is also useful to contributors, because they know which features they can use.

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -10,6 +10,7 @@ repository = "https://github.com/helix-editor/helix"
 homepage = "https://helix-editor.com"
 include = ["src/**/*", "README.md"]
 default-run = "hx"
+rust-version = "1.57"
 
 [package.metadata.nix]
 build = true


the-mikedavis:
Let's use the `package.metadata.msrv` key instead of `package.rust-version` to be more accurate. 1.57 is the MSRV but we use latest stable Rust to build as well and the releases are cut using latest stable Rust.

danyspin97:
`rust-version` is included in the manifest, it has the same meaning as `metadata.msrv`  and cargo will check it:

> The rust-version field is an optional key that tells cargo what version of the Rust language and compiler your package can be compiled with. If the currently selected version of the Rust compiler is older than the stated version, cargo will exit with an error, telling the user what version is required.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-rust-version-field

`cargo-msrv` also suggests using `rust-version` for Rust >= 1.56.

the-mikedavis (approved):
Oh I see, let's use `rust-version` then 👍

PR #3245, opened 2022-07-29
dap: Improve error message when starting tmux

danyspin97:
I

### DIFF
--- a/helix-view/src/handlers/dap.rs
+++ b/helix-view/src/handlers/dap.rs
@@ -307,11 +307,18 @@ impl Editor {
                                 e => panic!("Error to start debug console: {:?}", e),
                             })
                     } else {
-                        std::process::Command::new("tmux")
+                        let res = std::process::Command::new("tmux")
                             .arg("split-window")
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I think this is a good usability improvement but this whole function really needs a refactor to use a `Result` and bubble up errors into `Editor::set_error` rather than panics

helix-view/src/handlers/dap.rs
```diff
@@ -307,11 +307,18 @@ impl Editor {
...
-                            .spawn()
-                            .unwrap()
+                            .spawn();
+                        match res {
+                            Ok(child) => child,
+                            Err(e) if e.kind() == ErrorKind::NotFound => {
+                                // TODO: Gracefully handle this case without panicing
+                                panic!("Cannot find tmux in PATH");
+                            }
+                            Err(e) => panic!("Error when starting tmux: {:?}", e),
```
archseer, line 320: Rather than panic, can you call `editor.set_error()` ?

archseer:
Fixed in cb7615e0

PR #3247, opened 2022-07-29
Adds a run-shell-command-no-output command

dariooddenino:
Hi!
This PR adds an additional command to run a shell command ignoring the output.

The command is `:run-shell-command-no-output` (alias `:sh!`).

I'm a bit uncertain about the usage of `ShellBehavior::Ignore` and `Append`, but I didn't want to use a boolean.

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -69,3 +69,4 @@
 | `:append-output` | Run shell command, appending output after each selection. |
 | `:pipe` | Pipe each selection to the shell command. |
 | `:run-shell-command`, `:sh` | Run a shell command |
+| `:run-shell-command-no-output`, `:sh!` | Run a shell command ignoring output |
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Two things:

* I don't think this is an appropriate use of `!` in the alias name. Elsewhere `!` always signals "force" behavior like `:q!`. Here I'm kinda confused by its meaning.
* This should be async - `:run-shell-command-no-output sleep 5` for example blocks the whole editor for a while. Without output I can't think of any reason to block. We should probably not be blocking in `:sh` either.

dariooddenino:
I agree completely! What would a more appropriate name be?
`:sh-no-output` is not that much of a shorter alias.
I think the corresponding bash ways to do so don't translate that well to a short name.

the-mikedavis:
I think it should be `:shush` :P

the-mikedavis:
Looks like this needs a run of `cargo fmt` but otherwise this looks good

pascalkuthe:
Closing because this has gone pretty stale. We would still like to see this but as a flag for sh instead of a new command once we have a command flag parswe

PR #3248, opened 2022-07-29
Avoid setting stdin handle when not necessary

dariooddenino:
Hi!

Another small PR which avoids adding the stdin handler in `shell_impl` when not necessary.
This allows to run commands that don't expect any stdin whtout getting a `ENOTTY` error.

While this works, I have no experience with terminals (and rust), so it might be the completely wrong approach.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4541,18 +4541,18 @@ fn shell_impl(
     use std::process::{Command, Stdio};
     ensure!(!shell.is_empty(), "No shell set");
 
-    let mut process_ = Command::new(&shell[0]);
-    process_
+    let mut process = Command::new(&shell[0]);
+    process
         .args(&shell[1..])
         .arg(cmd)
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
 
     if input.is_some() {
-        process_.stdin(Stdio::piped());
+        process.stdin(Stdio::piped());
     }
 
-    let mut process = match process_.spawn() {
+    let mut process = match process.spawn() {
         Ok(process) => process,
         Err(e) => {
             log::error!("Failed to start shell: {}", e);
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4542,14 +4542,13 @@ fn shell_impl(
     ensure!(!shell.is_empty(), "No shell set");
 
     let mut process_ = Command::new(&shell[0]);
-
     process_
         .args(&shell[1..])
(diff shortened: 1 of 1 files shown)


AceofSpades5757 (approved):

AceofSpades5757:
I think this can cause problems with some programs. I'll usually let high-level functions handle anything too fancy and go with a heuristic approach to get simple scripts (like build scripts) to work.

helix-term/src/commands.rs
```diff
@@ -4541,13 +4541,24 @@ fn shell_impl(
...
+                     .stdout(Stdio::piped())
+                     .stderr(Stdio::piped())
+                     .spawn(),
+        None => Command::new(&shell[0])
+                     .args(&shell[1..])
+                     .arg(cmd)
+                     .stdout(Stdio::piped())
+                     .stderr(Stdio::piped())
+                     .spawn()
+    };
```
the-mikedavis, line 4558: This ends up being a lot of duplication. Instead you should bind the `Command::new` to a variable and only chain `.stdin` if `input.is_some()` and then call `.spawn()` at the end

dariooddenino: > This ends up being a lot of duplication. Instead you should bind the `Command::new` to a variable and only chain `.stdin` if `input.is_some()` and then call `.spawn()` at the end

I will try again. That was my first approach, but I was getting some errors and from a google search I mistakenly thought this was the only solution. Thanks!

helix-term/src/commands.rs
```diff
@@ -4541,14 +4541,18 @@ fn shell_impl(
     use std::process::{Command, Stdio};
     ensure!(!shell.is_empty(), "No shell set");
 
-    let mut process = match Command::new(&shell[0])
+    let mut process_ = Command::new(&shell[0]);
+    process_
```
the-mikedavis, line 4545: No need for the trailing `_` in `process_`. The `process` binding below will [shadow](https://doc.rust-lang.org/rust-by-example/variable_bindings/scope.html) this one

Elsewhere we've used `variable_` to avoid variable names that are keywords like [`match_`](https://github.com/helix-editor/helix/blob/0c08ff1596b398d53c8c48cc96cfbd24ddec303b/helix-core/src/syntax.rs#L1623)

dariooddenino: Thanks. Sorry for the mess, still in the process of learning rust!

AceofSpades5757:
That looks infinitely cleaner.

PR #2455, opened 2022-05-10
Allow using path suffixes to associate language file-types

midnightexigent:
Related: 

- #2443 
- #1426

After playing around with `shellcheck`, I think it's not the way to go mostly because of the path separator between linux and windows (\ vs /). Hence this PR uses globbing instead 

Now it's possible to associate for example `.git/config` to the `git-config` grammar by adding the following

```toml
[[language]]
name = "git-config"
file-globs = ["*.git/config"] 
```

image

PR #3249, opened 2022-07-29
Add tree-sitter-htmldjango

pickfire:
Great thanks to @interdependence for building [tree-sitter-django](https://github.com/interdependence/tree-sitter-htmldjango/).

I didn't check why the variables doesn't seem to work but other than that it is already good enough for now.

cc @WindSoilder I added `tag` and `keyword.function` to monokai spectrum since I was using it.

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -2,7 +2,7 @@ pub mod config;
 pub mod grammar;
 
 use etcetera::base_strategy::{choose_base_strategy, BaseStrategy};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 pub static RUNTIME_DIR: once_cell::sync::Lazy<PathBuf> = once_cell::sync::Lazy::new(runtime_dir);
 
@@ -88,21 +88,25 @@ pub fn log_file() -> PathBuf {
     cache_dir().join("helix.log")
 }
 
-pub fn find_root_impl(root: Option<&str>, root_markers: &[String]) -> Vec<PathBuf> {
-    let current_dir = std::env::current_dir().expect("unable to determine current directory");
-    let mut directories = Vec::new();
-
-    let root = match root {
+/// Get root to search until.
+pub fn search_root(root: Option<&str>) -> PathBuf {
+    let current_dir = || std::env::current_dir().expect("unable to determine current directory");
+    match root {
         Some(root) => {
-            let root = std::path::Path::new(root);
+            let root = Path::new(root);
             if root.is_absolute() {
                 root.to_path_buf()
             } else {
-                current_dir.join(root)
+                current_dir().join(root)
             }
         }
-        None => current_dir,
-    };
+        None => current_dir(),
+    }
+}
+
+pub fn find_root_impl(root: Option<&str>, root_markers: &[String]) -> Vec<PathBuf> {
+    let root = search_root(root);
+    let mut directories = Vec::new();
 
     for ancestor in root.ancestors() {
         // don't go higher than repo
--- a/languages.toml
+++ b/languages.toml
@@ -409,22 +409,10 @@ name = "scss"
 source = { git = "https://github.com/serenadeai/tree-sitter-scss", rev = "c478c6868648eff49eb04a4df90d703dc45b312a" }
 
 [[language]]
-name = "htmldjango"
-scope = "source.htmldjango"
-injection-regex = "htmldjango"
-file-types = ["**/templates/**.html"]
-roots = ["manage.py"]
-indent = { tab-width = 2, unit = "  " }
-
-[[grammar]]
-name = "htmldjango"
-source = { git = "https://github.com/interdependence/tree-sitter-htmldjango", rev = "184a50456186c2ff49b9b410f7060a176e2a3080" }
-
-[[language]]
 name = "html"
 scope = "text.html.basic"
 injection-regex = "html"
-file-types = ["*.html"]
+file-types = ["html"]
 roots = []
 language-server = { command = "vscode-html-language-server", args = ["--stdio"] }
 auto-format = true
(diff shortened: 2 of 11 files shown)


AceofSpades5757 (dismissed):

WindSoilder:
Ok, I'll checkout and update other monokai pro theme as soon as possible

languages.toml
```diff
@@ -423,6 +423,18 @@ indent = { tab-width = 2, unit = "  " }
 name = "html"
 source = { git = "https://github.com/tree-sitter/tree-sitter-html", rev = "d93af487cc75120c89257195e6be46c999c6ba18" }
 
+[[language]]
+name = "htmldjango"
+scope = "source.htmldjango"
+injection-regex = "htmldjango"
+file-types = ["html"]
```
the-mikedavis, line 430: It looks like this entry beats regular html for `.html` files. We have another conflict like this between verilog and V and I think the only way around it for now is to have one of them where you have to override it in a `.helix/languages.toml` or `~/.config/helix/languages.toml`:

```toml
[[language]]
name = "html"
file-types = []

[[language]]
name = "htmldjango"
file-types = ["html"]
```

Since HTML is more popular than django I think we should disable this language by default

```suggestion
file-types = []
```

runtime/queries/htmldjango/indents.scm
```diff
@@ -0,0 +1,3 @@
+(paired_statement) @indent
+(end_paired_statement) @indent_end
+(branch_statement) @branch
```
the-mikedavis, line 3: These look like they're in nvim style (we don't use `branch` or `indent_end` iirc). Could you adjust them to the helix ones? https://docs.helix-editor.com/guides/indent.html

I'm not caught up yet on the indentation mechanism so actually I'm not too sure how to these should look myself 😅

The default indent mechanism (removing this file) might work well. It seems to work pretty well for HTML for me

languages.toml
pickfire: Yes, it beats `html`, but doesn't it only apply if say the root file `manage.py` is found?

pickfire: Oh wait, it doesn't. Maybe we should apply it conditionally based on the `roots`?

AceofSpades5757: I was also under the impression that `file-types` would only be superceded if `manage.py` was found based on these settings, a file I've never heard of or seen people using until I did some web searching based on this PR's options. I'll need to re-read those options.

pickfire: This should be fixed now, but I am still a bit not quite happy with the result that the ordering is a bit messy (it will take the first glob that matches the path and have the root within the path). Some parts of it was taken from https://github.com/helix-editor/helix/pull/2455, but I haven't updated the docs yet.

pickfire: This should be fixed now, but I am still a bit not quite happy with the result that the ordering is a bit messy (it will take the first glob that matches the path and have the root within the path). Some parts of it was taken from https://github.com/helix-editor/helix/pull/2455, but I haven't updated the docs yet.

runtime/queries/htmldjango/indents.scm
pickfire: I changed it but didn't go and test it out.

runtime/queries/htmldjango/indents.scm
```diff
@@ -0,0 +1,2 @@
+(paired_statement) @indent
+(end_paired_statement) @outdent
```
pickfire, line 2: Thinking about this again, not sure if we want this since vim does not indent django paired statements.

languages.toml
```diff
@@ -408,11 +408,23 @@ indent = { tab-width = 2, unit = "  " }
...
+[[grammar]]
+name = "htmldjango"
+source = { git = "https://github.com/interdependence/tree-sitter-htmldjango", rev = "184a50456186c2ff49b9b410f7060a176e2a3080" }
+
 [[language]]
 name = "html"
 scope = "text.html.basic"
 injection-regex = "html"
-file-types = ["html"]
+file-types = ["*.html"]
```
the-mikedavis, line 427: One of the downsides of introducing globbing is that it becomes ambiguous whether all `file-type`s should be `*.extension` or just `extension`. In this case, this has to be `*.html` so the globs for htmldjango and html are both checked, right? I'm worried that behavior is too nuanced and gets confusing

I think it would be more predictable to have `language_config_for_file_name` allow ties and then decide between the ties based on any `roots` as you were saying above (https://github.com/helix-editor/helix/pull/3249#discussion_r934109113)

runtime/queries/htmldjango/textobjects.scm
```diff
@@ -0,0 +1,7 @@
+[
+  "{{"
+  "}}"
+  "{%"
+  "%}"
+  (end_paired_statement)
+] @tag
```
the-mikedavis, line 7: I'm not quite sure which textobject this should be. Maybe `@function.around`? Or it could use the ones from https://github.com/helix-editor/helix/pull/3282?

pickfire: I am not sure either, but since I probably won't use textobjects.scm anytime soon so I guess I will just remove it.

languages.toml
pickfire: I added hueristic and won't rely on glob now.

helix-loader/src/lib.rs
```diff
@@ -88,21 +88,25 @@ pub fn log_file() -> PathBuf {
...
+                current_dir().join(root)
             }
         }
-        None => current_dir,
-    };
+        None => current_dir(),
+    }
+}
+
+pub fn find_root_impl(root: Option<&str>, root_markers: &[String]) -> Vec<PathBuf> {
```
pickfire, line 107: I don't quite know why are we not using `Path` here, but I don't feel like changing this here.

pickfire:
@the-mikedavis What do you think about the glob changes? If you are okay then I will fix the conflicts.

the-mikedavis:
I like the new version where languages can share file-types 👍. I think that will help with v and verilog which share a file-type too.

It could probably use a line in the documentation somewhere (maybe `book/src/languages.md`?) to make it explicit what the behavior is

MrConorAE:
Hi all, any update on this? I'm currently bodging my way through this by copying the .scm files, but it isn't working well. Happy to test with my codebase if that's needed.

pascalkuthe:
closing this one as stale, thank you for contributing!

PR #3252, opened 2022-07-29
Only add cargo-tarpaulin on x86_64

groves:
It only supports that platform: https://github.com/NixOS/nixpkgs/blob/433bd5e89cbb5fec3d16e90a9df4734ef2fe9fcb/pkgs/development/tools/analysis/cargo-tarpaulin/default.nix#L29

Without this `nix develop` fails with this error on my system:

```
error: Package ‘cargo-tarpaulin-0.20.1’ in /nix/store/.../cargo-tarpaulin/default.nix:25
is not supported on ‘aarch64-linux’, refusing to evaluate.
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -82,7 +82,7 @@
               ++ (
                 with common.pkgs;
                 [lld_13 lldb cargo-flamegraph rust-analyzer] ++
-                (lib.optionals stdenv.isx86_64 [cargo-tarpaulin])
+                (lib.optional stdenv.isx86_64 cargo-tarpaulin)
               );
             env =
               prev.env
--- a/flake.nix
+++ b/flake.nix
@@ -82,7 +82,7 @@
               ++ (
                 with common.pkgs;
                 [lld_13 lldb cargo-flamegraph rust-analyzer] ++
-                (lib.optional stdenv.isx86_64 cargo-tarpaulin)
+                (lib.optional (stdenv.isx86_64 && stdenv.isLinux) cargo-tarpaulin)
               );
             env =
               prev.env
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -80,7 +80,9 @@
             packages =
               prev.packages
               ++ (
-                with common.pkgs; [lld_13 lldb cargo-tarpaulin cargo-flamegraph rust-analyzer]
+                with common.pkgs;
+                [lld_13 lldb cargo-flamegraph rust-analyzer] ++
+                (lib.optionals stdenv.isx86_64 [cargo-tarpaulin])
```
the-mikedavis, line 85:
```suggestion
                (lib.optional stdenv.isx86_64 cargo-tarpaulin)
```

groves: Nice! Hadn't seen lib.optional before.

flake.nix
```diff
@@ -80,7 +80,9 @@
             packages =
               prev.packages
               ++ (
-                with common.pkgs; [lld_13 lldb cargo-tarpaulin cargo-flamegraph rust-analyzer]
+                with common.pkgs;
+                [lld_13 lldb cargo-flamegraph rust-analyzer] ++
+                (lib.optional stdenv.isx86_64 cargo-tarpaulin)
```
the-mikedavis, line 85:
```suggestion
                (lib.optional (stdenv.isx86_64 && stdenv.isLinux) cargo-tarpaulin)
```

Based on

> Tarpaulin only supports x86_64 processors running Linux

from cargo-tarpaulin's [readme](https://github.com/xd009642/tarpaulin/blob/79c113698f2960b47496c4953ef33f7d8c8b96c1/README.md?plain=1#L17)

the-mikedavis (approved):
thanks!

Issue #1210, opened 2021-12-01
Add `:default` command that generate and open a default config file


NNBnh:
Many newcomer feel lost when trying to find a "setting panel" on a minimal software then realize that they have to open and manually edit a config file. Although me and other minimalist user prefer this config method, other may wish to have all setting option layout in front of them so they could just change the options values. That why many recent `cli` rust tool auto generate (or have an option flag) a config file template will all option inside it. This will be very useful not just to newcomer but also to a pro who is interest and want to start try out quickly. Here is my attempt:

| TODO | Later on we could add explanation comment to every option. |
|-|-|

<details>
<summary>=== Read full config ===</summary>

```toml
[editor]

"scrolloff" = 3
"mouse" = true
"middle-click-paste" = true
"scroll-lines" = 3
"shell" = ["sh", "-c"]
"line-number" = "absolute"
"smart-case" = true
"auto-pairs" = true
"auto-completion" = true
"idle-timeout" = 400
"completion-trigger-len" = 2
"auto-info" = true

[editor.filepicker]

"hidden" = true
"parents" = true
"ignore" = true
"git-ignore" = true
"git-global" = true
"git-exclude" = true


[keys.normal]

# Movement
"left" = "move_char_left"
"down" = "move_line_down"
"up" = "move_line_up"
"right" = "move_char_right"
"h" = "move_char_left"
"j" = "move_line_down"
"k" = "move_line_up"
"l" = "move_char_right"
"w" = "move_next_word_start"
"b" = "move_prev_word_start"
"e" = "move_next_word_end"
"W" = "move_next_long_word_start"
"B" = "move_prev_long_word_start"
"E" = "move_next_long_word_end"
"t" = "find_till_char"
"f" = "find_next_char"
"T" = "till_prev_char"
"F" = "find_prev_char"
"A-." = "repeat_last_motion"
"home" = "goto_line_start"
"end" = "goto_line_end"
"pageup" = "page_up"
"pagedown" = "page_down"
"C-u" = "half_page_up"
"C-d" = "half_page_down"
"C-i" = "jump_forward"
"C-o" = "jump_backward"
"v" = "select_mode"
":" = "command_mode"

# Changes
"r" = "replace"
"R" = "replace_with_yanked"
"~" = "switch_case"
"`" = "switch_to_lowercase"
"A-`" = "switch_to_uppercase"
"i" = "insert_mode"
"a" = "append_mode"
"I" = "prepend_to_line"
"A" = "append_to_line"
[...]

PR #3253, opened 2022-07-29
feat: `default-keymap-open` and `current-keymap-open` typeable commands

AlexanderBrevig:
`:current-keymap-open` will open a new buffer with the currently loaded config. 
`:default-keymap-open` will open a new buffer with the default config.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1471,9 +1471,10 @@ fn current_keymap(
                     })
                     .collect::<Result<String, toml::ser::Error>>();
 
-                let str = match serialized_str {
-                    Ok(str) => str,
-                    Err(_) => return (),
+                let str = if let Ok(str) = serialized_str {
+                    str
+                } else {
+                    return;
                 };
 
                 let doc = Document::from(Rope::from_str(&str), None);
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1,8 +1,11 @@
 use std::ops::Deref;
 
+use crate::keymap;
+
 use super::*;
 
 use helix_view::editor::{Action, ConfigEvent};
+
 use ui::completers::{self, Completer};
 
 #[derive(Clone)]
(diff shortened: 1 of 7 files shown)


AceofSpades5757:
When are you getting this `ValueAfterTable` error? I'm not getting this error.

AlexanderBrevig:
> When are you getting this `ValueAfterTable` error? I'm not getting this error.

We very quickly sorted it out. I'll update the description.

AceofSpades5757:
It's cool, to be sure, but I'm not sure what the purpose would be used for. Would I want to fill a buffer full of all the key mappings? If I change a mapping, and save it, will it change (it won't but this was something I tried)?

I think it would be cool if I could specify a key map, and it would place that configuration in my primary register. That might be a bit out of scope though for what you're trying to accomplish.

AceofSpades5757:
Adding the different modes was a nice touch. Seeing all the different keys while in the different mode is wild to see.

s0LA1337:
> It's cool, to be sure, but I'm not sure what the purpose would be used for. Would I want to fill a buffer full of all the key mappings?

Let me explain how at least I had the idea of this feature.
I am working at home with QWERTY and at work with Colemak, changing every setting here to Colemak is kinda a pain. 
The idea was, that this will make a few things easier. 
- when you want to remap keys
- debugging as in e.g. To see whether you have keys which overlap the default configuration 
- generally, I also think it's a good idea to have some way to access the default configuration

AlexanderBrevig:
Maybe it should be `open-current-keymap` that shows the active working set, and then a `open-default-keymap` as well? 
We should set the buffer to read-only.

AlexanderBrevig:
There may be a bug with it printing duplicate config. Will investigate.

AlexanderBrevig:
Since select mode inherits normal mode, we should remove duplicates before serializing to keep output more similar with `keymap/default.rs`

helix-term/src/commands/typed.rs
```diff
@@ -1424,6 +1427,87 @@ fn open_log(
...
+        // there must be a way to make this easier i think
+        .map(|couple| match couple.0 {
+            Mode::Insert => toml::to_string(&Keys {
+                keys: ModeInsert { insert: couple.1 },
+            }),
+            Mode::Normal => toml::to_string(&Keys {
+                keys: ModeNormal { normal: couple.1 },
+            }),
+            Mode::Select => toml::to_string(&Keys {
+                keys: ModeSelect { select: couple.1 },
```
AlexanderBrevig, line 1460: We should remove duplicates from normal mode here

s0LA1337:
related issue: #1210

PR #3262, opened 2022-07-30
feat: add cue support

b4nst:
This PR add support for [CUE](https://github.com/cue-lang/cue) using [dagger/cuelsp](https://github.com/dagger/cuelsp) and [eonpatapon/tree-sitter-cue](https://github.com/eonpatapon/tree-sitter-cue).

I'm wondering if it's there's a way to give the `format` command another command than the lsp (`cue fmt`)?
It's not a big caveheat since https://github.com/dagger/cuelsp/issues/44 should resolve that anyway.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -10,6 +10,7 @@
 | cpon | ✓ |  | ✓ |  |
 | cpp | ✓ | ✓ | ✓ | `clangd` |
 | css | ✓ |  |  | `vscode-css-language-server` |
+| cue | ✓ |  |  | `cuelsp` |
 | dart | ✓ |  | ✓ | `dart` |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Can you adjust the scopes here to match the ones we use in helix? https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example, we don't use `@repeat` or `@number`

b4nst:
@the-mikedavis of course! Will give it a go tomorrow

b4nst:
@the-mikedavis sorry I didn't realise it was not neovim compatible scopes. Should be better now (and indeed that looks way better 🤦)

the-mikedavis (approved):
Looks good, thank you!

PR #3266, opened 2022-07-30
Add extend variants for tree-sitter textobjects

### LINKED ISSUE #3263: Extend variants for tree-sitter textobject motions
Tree-sitter based textobject motions (for example `]f`/`[f`) do not currently have different implementations between normal mode and select mode (`v`).

The behavior of these motions in select mode is worth some discussion. I see two options:

1. Like `]p`/`[p` or `w`/`b` in select mode, they should keep the anchor and move the head. This "extend" behavior is the normal behavior of motions in select mode. There is prior implementation-work for this behavior: https://github.com/helix-editor/helix/pull/2802. (I'm looking at pulling this across the finish line now.)
2. Like `n`/`N`, they should add new selections containing the next/previous textobject.

Option 2 is a bit unconventional (`n`/`N` is the only other that uses this behavior off the top of my head) but I think it would be more useful. Tree-sitter textobjects may not be contiguous in a file in the same way paragraphs are (by `]p`/`[p`'s definitions), so I think `v]f` would select other text between two function definitions I would not want. Even if this is not selected as the correct behavior, it might be good to adding separate commands that implement this behavior.

the-mikedavis:
https://asciinema.org/a/523187

This is a somewhat significant behavior change for tree-sitter textobject motion. Prior to this change you could select all functions in a file with repeated `]f`. With this change, nested functions are not selected with repeated `]f` but I think the new behavior makes sense.

The new behavior is something like

* `]f` selects the function textobject with a `start_byte()` after the current selection's cursor
* `[f` selects the function textobject with an `end_byte()` before the current selection's cursor

After a bit of adjustment, I think the new behavior 'feels' right.

Enabled with that change, I've added `doc.mode == Mode::Select` behavior for `goto_treesitter_object` which behaves as option 1 of #3263.

~Based on #3264 for simplicity's sake~
Supersedes #2802
Closes #3263

### DIFF
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -389,6 +389,8 @@ fn reached_target(target: WordMotionTarget, prev_ch: char, next_ch: char) -> boo
     }
 }
 
+/// Finds the range of the next or previous textobject in the syntax sub-tree of `node`.
+/// Returns the range in the forwards direction.
 pub fn goto_treesitter_object(
     slice: RopeSlice,
     range: Range,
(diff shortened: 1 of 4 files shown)


dead10ck:
Is this ready to be reviewed?

the-mikedavis:
It needs a rebase

dead10ck (approved):
I agree, this behavior makes more sense. This all looks good to me, thanks!

Issue #3167, opened 2022-07-23
"Other window" concept for goto and search selections from Picker


dam5h:
I am enjoying test driving helix and think it could be my daily driver (using spacemacs now) once a few more things land.  One that I have not seen in the issue history yet is the idea of moving the "found" items from the `Picker` into the "other" split window.  This is a pattern I use all the time, where I am editting one file, but then I want to see the definion of something in another window, without replacing the current window with the search result.  I use this "other window" workflow for goto definition (1) as well as local file (2) and project wide (3) searching.

In emacs I can use `g + d` to goto definition in the current window and `g + D` to goto definition in the "other" window.  It's quite handy!

Awesome project, thank you for all the work. Looking forward to using helix more in the future!

dam5h:
I made a fork and added a helper function that takes a bool for `rotate` as to whether or not to stay in the current view or call `editor.focus_next()` in the closure given to `Context.callback()` in order to bring focus to the "other" node prior to loading the resulting locations from the lsp query.  I implemented for both lsp goto references and definition, using capital "R" and "D" respectively in the default keymap.

See the commit [here](https://github.com/dam5h/helix/commit/a89eb64d25b95e348d6d1a61d3c1d2cf81e043bd), let me know if a PR should be created.  I am new to the code base so I am not convinced this is the most idiomatic way to approach it, but its working well enough that I can keep kicking the tires on helix!

EpocSquadron:
With regards to commands that call up a picker, there is a keybind to open the result in a new split already. I see your PR figures out whatever the next split is and reuses it instead. That would be cool to also have a keybind for in pickers and should be fairly easy to add by copying and modifying, no?

dam5h:
Oh nice, I did not realize there was as keybind in the picker already, I will take a look for that.  I do think that would be better in that case -- allowing the user to defer the decision of which window to use at the end of the selection process.  For `goto_definition` since that only has a single result and no picker the user can decide at the onset to use the other window.

dam5h:
@EpocSquadron , Thank you for pointing out the keybinds on `Picker`!  I have updated the PR to use this instead for `Picker` selections.

dam5h:
Here is the [PR](https://github.com/helix-editor/helix/pull/3652) for this issue.

PR #3269, opened 2022-07-31
Lsp goto definition and references "other window"

dam5h:
This would address #3167 that I made last week.

I think it may also be nice to be able to do the same thing generically from the picker menu, whereas this approach is explicitly defined only for goto definition and goto references.  Goto definition it a direct command so knowing at the onset of the command makes sense, but for references or anything where multiple results are possible (which would give rise to the `Picker`) a keymap command at the picker stage which would rotate the window could be handy.  That would be more involved.


dam5h:
I did not realize that the picker had it's own keybinds until I saw this [comment](https://github.com/helix-editor/helix/issues/3167#issuecomment-1201017024).  Glad to now know that!  I have now added a `ctrl!('o')` keybind in the picker that will load into the "other" window, via adding a new `Action` for `editor.switch` to `match` onto, which just calls `focus_next()` and then `switch()` recursively with `Action::Replace`.

Issue #3164, opened 2022-07-23
allow customization of file type in statusbar


sbromberger:
#### Describe your feature request

Now that #2434 has been merged and we have this great new capability, I'd like to revisit the discussion started [ in this comment](https://github.com/helix-editor/helix/pull/2434#issuecomment-1163230782). Specifically, I'd like to get feedback on my idea to allow customization of file type to user-settable strings. "typescript" takes 9 characters in the statusbar, but with the properly installed glyphs can be reduced to a single "TS" icon. This gives more room for other statusbar plugins that will be coming in future updates.

As commented, this shouldn't be a heavy lift in terms of code - it looks to be a fairly straightforward PR - but it's probably worth some discussion to see whether this is a feature that others (besides myself) would use.

sbromberger:
cc @etienne-k

etienne-k:
I see a number of different approaches. Here's two of them which I believe offer the most advantages:

#### Simple but flexible configuration

The idea would be to be able to configure how the file type is displayed using two configuration options:

```toml
[editor.statusline.file-type]
name = "short" # long/short/none
icon = true
```

This is probably the highest level of language-independent configurability.

It remains to be verified if `icon` should really be a local property here, or if we shouldn't simply offer a global opt-in option like `devicons = true` which would then also apply icons to the future file tree and maybe to some VCS/Git statusline element.

The downside of this approach is that for each file type, we'd have to define a "short" name.

Nonetheless, whether or not we use short names, and whether the devicons property is defined locally or globally, I prefer this solution because it provides quite some customizability without requiring "too much configuration effort" from the user.

#### User-defined mappings

Here, we'd let the user define how each file type would be represented:

```toml
[editor.statusline.file-type]
"name.rust" = "rs" # Could also use a devicon here
"name.typescript" = "ts"
```

This would be the easiest to implement and fastest to achieve, but we're basically shifting work to the user. It remains to be seen if this would be an approach we really want.

sbromberger:
Honestly, I'm in strong favor of the second option (User-defined mappings). It allows customization beyond just "devicons" - if I wanted just `ts` or `rs` for "typescript" or "rust", I could do it - but if we use the first option that no longer becomes possible.

My feeling is that if users are going out of their way to change the defaults they should be offered the widest possible latitude for customization.

sbromberger:
[...]

PR #3273, opened 2022-07-31
customizable file type mapping

sbromberger:
ref: #3164 

This PR introduces a new `[editor.file-types]` config section that will allow custom mappings between the default file type strings and user-defined custom strings. An example:
```toml
[editor.file-types]
"rust" = "rs"
```

will remap the "rust" file type string to "rs" (currently used by the statusline). Glyphs/icons should be supported, if available in the user's font definition.

cc @etienne-k

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -330,6 +330,12 @@ where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
     let file_type = context.doc.language_id().unwrap_or("text");
+    let editor_config = context.editor.config();
+    let file_type = editor_config
+        .file_types
+        .get(file_type)
+        .map(|x| x.as_str())
+        .unwrap_or(file_type);
 
     write(context, format!(" {} ", file_type), None);
 }
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -332,7 +332,7 @@ where
     let file_type = context.doc.language_id().unwrap_or("text");
     let editor_config = context.editor.config();
     let file_type = editor_config
-        .file_types
+        .file_type_indicators
         .get(file_type)
         .map(|x| x.as_str())
         .unwrap_or(file_type);
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -334,8 +334,7 @@ where
     let file_type = editor_config
         .file_type_indicators
         .get(file_type)
-        .map(|x| x.as_str())
-        .unwrap_or(file_type);
+        .map_or(file_type, String::as_str);
 
     write(context, format!(" {} ", file_type), None);
 }
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -166,7 +166,7 @@ pub struct Config {
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
     /// File type icon/string map for overriding file type.
-    pub file_types: FileTypeConfig,
+    pub file_type_indicators: FileTypeConfig,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
@@ -532,7 +532,7 @@ impl Default for Config {
             whitespace: WhitespaceConfig::default(),
             indent_guides: IndentGuidesConfig::default(),
             color_modes: false,
-            file_types: FileTypeConfig::default(),
+            file_type_indicators: FileTypeConfig::default(),
         }
     }
 }
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -165,6 +165,8 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    /// File type icon/string map for overriding file type.
+    pub file_types: FileTypeConfig,
 }
 
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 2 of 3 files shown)


AlexanderBrevig:
Meybe this should be under the `statusline` key? As I understand it, this only affects the rendered output in the statusline. Right?

sbromberger:
Right now that's the case, but it doesn't have to be (see [disc](https://github.com/helix-editor/helix/issues/3164#issuecomment-1200386217)).

the-mikedavis:
This key name is confusing: looking just at the toml I would think this makes helix think any `.rs` files are rust (i.e. the `file-types` key in languages.toml)

sbromberger:
The key name is derived from the default strings ("rust", "typescript", "javascript", etc.) returned from `context.doc.language_id()`. Is there a better way of doing this map without having to redefine a bunch of language-specific identifiers?

the-mikedavis:
I mean the `editor.file-type` key, `editor.file-type-indicator` seems like a good choice

sbromberger:
I kept it plural (`file-type-indicators`) since you might have more than one defined in this section.

helix-view/src/editor.rs
```diff
@@ -165,6 +165,8 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    /// File type icon/string map for overriding file type.
+    pub file_type_indicators: FileTypeConfig,
```
the-mikedavis, line 169: if `FileTypeConfig` isn't doing any validation while deserializing, why not just use the HashMap directly?

sbromberger: I could do that but this seems a bit more flexible. Does it come with any performance impact?

etienne-k (approved):

sbromberger:
Just in case anyone wants to try this with nerdfonts, here's a quick file I put together that has all the filetypes I and Helix know about: https://gist.github.com/e8443841c06f938045ffc57069a281fa

helix-view/src/editor.rs
the-mikedavis: I don't think it there would be a noticeable performance change. It's more about simplicity - using a custom struct adds a bunch of extra code for (de)serializing and deref that doesn't seem to add any value. If anything, you might be able to use a [type alias](https://doc.rust-lang.org/reference/items/type-aliases.html) to "rename" `HashMap<String, String>` but IMO it's still preferrable to just use the hashmap

AlexanderBrevig:
Here's a version that falls back to use the scope identifier if no LSP is provided.
Since Rust has an LSP then `rust` will stay as `rust`. Since `markdown` has no LSP then it will use `md` as that is its scope ident.

```
fn render_file_type<F>(context: &mut RenderContext, write: F)
where
{
    let lsp_language_id = context.doc.language_id().unwrap_or("text");
    let mut lsp_language_fallback = String::from("lsp.");
    lsp_language_fallback.push_str(lsp_language_id);
    let (_, file_ext) = context
        .doc
        .language()
        .unwrap_or(lsp_language_fallback.as_str())
        .rsplit_once('.')
        .expect("Language scope should have ident.ext format");
    let editor_config = context.editor.config();
    let file_type = editor_config
        .file_type_indicators
        .get(file_ext)
        .map(|x| x.as_str())
        .unwrap_or(lsp_language_id);

    write(context, format!(" {} ", file_type), None);
}
```

Not overly testet but what I did test seemed to work. 
May serve as a point of discussion, I don't expect the code to be used.

Nice feature BTW :+1:

sbromberger:
I think this is ready for review. The bug in language identification is unrelated to this PR but fixing it will definitely make this feature more accurate.

helix-term/src/ui/statusline.rs
```diff
@@ -330,6 +330,12 @@ where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
     let file_type = context.doc.language_id().unwrap_or("text");
+    let editor_config = context.editor.config();
+    let file_type = editor_config
+        .file_type_indicators
+        .get(file_type)
+        .map(|x| x.as_str())
+        .unwrap_or(file_type);
```
the-mikedavis, line 338:
```suggestion
        .map_or(file_type, |x| x.as_str());
```

helix-term/src/ui/statusline.rs
```diff
@@ -330,6 +330,12 @@ where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
     let file_type = context.doc.language_id().unwrap_or("text");
```
the-mikedavis, line 332: Somewhat unrelated to this change, I think this might be better off as the `scope` field from `languages.toml` which is `context.doc.language().unwrap_or("text")`. That way Rust is "source.rust", toml is "source.toml", html is "text.html.basic", etc.

sbromberger: Do you suggest any changes to this PR? My opinion is that we should have a function whose sole purpose is to return the language. Right now there are several possible ways to do it, each with their own issues.

the-mikedavis: > we should have a function whose sole purpose is to return the language

The trouble is: what is the language? We could say it's the `scope` or we could say it's the `name` (language_id but not `language_id()`, see also https://github.com/helix-editor/helix/pull/3338). It depends on what we want the public API of language names to be: is the `LanguageConfiguration.language_id` an implementation detail or is it ok for configurations to use them?

I'd be hesitant to merge this PR until we figure that out since this PR introduces a way for configurations to depend on language names/ids/scopes.

@archseer what do you think?

sbromberger: This PR merely provides a mapping from the current language reporting functionality so that we can use custom strings that don't take up 1/4 of the status line. I'd be really disappointed if this PR was held up as a result of a feature that is still unimplemented.

PR #3275, opened 2022-07-31
Fix nondeterministic highlighting

### LINKED ISSUE #2484: Sometimes when hx is started, syntax highlighting is incomplete
### Summary

When running `hx` repeatedly to open the same file, ~4 out of every 10 sessions will start with incorrect syntax highlighting:
Screenshot_2022-05-16_14-29-53

I've seen this happen with both `.rb` and `.ex` files. I expect it happens with other file types too, but don't know if I've noticed it.

In a `hx` session that started with this bug, `:open`-ing a new file, even a file of a different language, does not fix the highlighting. Nor does running `:set-language`, `:theme`, or anything else I tried. `hx` must be restarted.

### Reproduction Steps

I tried this:

`hx whatever.rb`,  `hx whatever.rb`,  `hx whatever.rb`,  `hx whatever.rb`,  `hx whatever.rb`, either in sequence or in separate terminal emulator windows

I expected this to happen:

Each of the 5 `hx` invocations starts Helix, with the same color theme, and they all look the same.

Instead, this happened:

One or more of the running `hx` sessions doesn't correctly highlight these specific parts of the code. Each time the bug happens, it's these same parts that aren't properly highlighted.

### Helix log

(whether a `hx` session starts with this bug or not, log output is the same for me.)

### Platform

Linux (amd64)

### Terminal Emulator

Affects both gnome-terminal and cool-retro-term

### Helix Version

helix 22.05-dev (cc68fa85)

### LINKED ISSUE #2986: Syntax highlighting conflict of `fn` keyword in rust
### Summary

As of recently the `fn` keyword in rust sometimes has wrong colors.

This probably has a similar cause as #1597 (the third issue highlighted).
The two theme keys that match the `fn` keyword seem to be `keyword` and `function`.

### Reproduction Steps

Edit a basic rust file like this and reload the theme a few times:

```rs
fn main() {}
```

### Helix log

_No response_

### Platform

Windows

### Terminal Emulator

wezterm (nightly)

### Helix Version

helix 22.05 (1378b911)

### LINKED ISSUE #3230: Syntax highlighting sometimes broken when started
### Summary

Sometimes the highlight queries seem to be broken (tested with typescript):

Broken (note the `@type` highlighting for `const` or `type`, although, I've checked none of the `@type` queries should match `const` or `type`):

syntax_highlighting_issue

Correct:

syntax_highlighting_issue_correct

Seems to happen roughly 50% of the time.

This may be related or even a duplicate to/of https://github.com/helix-editor/helix/issues/2484

Tested with this script:

```typescript
const constant = "const";

namespace some.namespace {
  function something() {}
}

declare namespace GreetingLib.Options {
  // Refer to via GreetingLib.Options.Log
  interface Log {
    verbose?: boolean;
  }
  interface Alert {
    modal: boolean;
    title?: string;
    color?: string;
  }
}
type MyGreeter = () => number;

type GreetingLike = string | (() => string) | MyGreeter;
declare function greet(g: GreetingLike): void;
``` 

### Reproduction Steps

Start helix with the script above

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

22.05-324-g681c0a91

A-Walrus:
Resolves #2986.
Resolves #3230.
Probably resolves #2484 as well, although I didn't have the code to copy paste in order to check :P

This is done by preferring matches in the beginning, i.e. for `keyword.function`, `keyword` is a better match than `function`.

### Explanation of original problem:
Source for non-determinism: `HashMap` traverse order is different each time.

According to comments: 
> Consumers of these queries can choose to recognize highlights with different levels of specificity.
For example, the string `function.builtin` will match against `function.method.builtin` 
and `function.builtin.constructor`, but will not match `function.method`.

`keyword.function `can therefore match both `keyword` and `function`.
 - Both of these matches have the same `match_len` which is what is used to rank them
 - The list of scopes is traversed in order to find the match with the best `match_len`, in the 
   *random* order it came from the HashMap.

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -1358,11 +1358,12 @@ impl HighlightConfiguration {
 
                 let mut best_index = None;
                 let mut best_match_len = 0;
-                let mut best_leftmost = usize::MAX;
+                let mut best_positions = Vec::new();
+                let mut positions = Vec::new();
                 for (i, recognized_name) in recognized_names.iter().enumerate() {
                     let recognized_name = recognized_name;
                     let mut match_len = 0;
-                    let mut leftmost = usize::MAX;
+                    positions.clear();
                     let mut matches = true;
                     for part in recognized_name.split('.') {
                         let pos = capture_parts.iter().position(|a| *a == part);
@@ -1371,20 +1372,21 @@ impl HighlightConfiguration {
                                 matches = false;
                                 break;
                             }
-                            Some(p) => {
-                                if p < leftmost {
-                                    leftmost = p;
-                                }
+                            Some(pos) => {
+                                positions.push(pos);
                                 match_len += 1;
                             }
                         }
                     }
+                    positions.sort_unstable();
 
                     let better_len = match_len > best_match_len;
-                    let better_leftmost = match_len == best_match_len && leftmost < best_leftmost;
-                    if matches && (better_len || better_leftmost) {
+                    let better_positions =
+                        match_len == best_match_len && positions < best_positions; // Vec comparison is lexicographic
+                    if matches && (better_len || better_positions) {
                         best_index = Some(i);
-                        best_leftmost = leftmost;
+                        best_positions.clear();
+                        best_positions.extend_from_slice(positions.as_slice());
                         best_match_len = match_len;
                     }
                 }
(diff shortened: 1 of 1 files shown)


A-Walrus:
Oops found a problem. If there are queries with the same leftmost but different rightmost it will still be random. An example of this would be `function.method.builtin` with the queries `function.builtin` and `function.method`. Will fix soon

archseer:
There's an easier fix here: I think the original intends for matches to be from the start of the string. So `keyword.function` matches `keyword`, `keyword.function`, but not `function`.

the-mikedavis (approved):
Nice, thanks for the fix!

I think the old behavior may have been intentional but I think the new way is more straightforward and intuitive (plus less edge-case-y)

And thanks for updating the docs too :)

archseer:
Thanks for fixing this! I'm really glad this was resolved 🎉

PR #3280, opened 2022-08-01
javascript queries: fix parameters

ChrHorn:
Mostly taken from https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/javascript/highlights.scm

This should enable parameter highlighting, including locals.


the-mikedavis (approved):
Works like a charm, thanks!

PR #3282, opened 2022-08-01
textobjects for html/xml tags

ff2400t:
I bundled the html and jsx queries with this pr to show how it work in practice

`mae` - will select the element and all it's child element, this is can also be used to select self closing element like we see in jsx
`mie` - will select the all the child nodes of an element


AlexanderBrevig:
Please do not change the behavior of `t` as it's probably part of muscle memory for multiple people already (myself included).

If `e` is available, maybe use that instead as `element`?

the-mikedavis:
Yeah I don't see a compelling reason for a breaking change here: more languages have queryable test conventions than XML tags. We could use `T` for xml tags?

the-mikedavis:
There are also the navigation commands `]e`/`[e` to add: https://github.com/helix-editor/helix/blob/0c08ff1596b398d53c8c48cc96cfbd24ddec303b/helix-term/src/keymap/default.rs#L100-L121

runtime/queries/html/textobjects.scm
```diff
@@ -0,0 +1,5 @@
+(element) @xml_element.around
+
+((start_tag) 
+	[(text) (element) ]* @xml_element.inside
+(end_tag))
```
the-mikedavis, line 5: let's use something like

```scm
(element
  (start_tag)
  (_)* @xml_element.inside
  (end_tag)) @xml_element.around

(start_tag) @xml_element.around

(comment) @comment.around
```

for html. That handles self-closing elements like `<link ... />`

Plus might as well throw the comment textobject in there while we're making textobjects 😛

ff2400t: added the queries for self_closing_tags,
also sorry but i don't understand why do you have that anchor at the end of the first query
is it some case that i am missing to see?

the-mikedavis: Ah actually I forgot about `self_closing_tag`. The anchor there is so that the stanza matches only elements with both start and end tag. The second stanza then matches elements with just a `start_tag` like how peolpe sometimes write `<br>`

the-mikedavis: I suppose that the grammar rules ensure that an element can only have one trailing `end_tag` though, right? So I suppose there's no need for the anchor

runtime/queries/html/textobjects.scm
```diff
@@ -0,0 +1,11 @@
+(element
+  (start_tag)
+  (_)* @xml_element.inside
+  (end_tag)
+. ) @xml_element.around
```
the-mikedavis, line 5: We may want to duplicate this stanza for `style_element` and `script_element` since those are separate rules in the grammar. They both need `end_tag`s though so they won't need the `?` suggestion above

ff2400t: Added these

runtime/queries/html/textobjects.scm
```diff
@@ -0,0 +1,11 @@
+(element
+  (start_tag)
+  (_)* @xml_element.inside
+  (end_tag)
```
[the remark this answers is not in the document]
ff2400t: Added this

runtime/queries/html/textobjects.scm
```diff
@@ -0,0 +1,23 @@
+(element
+  (start_tag)
+  (_)* @xml_element.inside
+  (end_tag)?
+. ) @xml_element.around
```
the-mikedavis, line 5: Looks like there's some odd interaction here with the `*` vs the `?` actually. I think we may need to split these into two separate stanzas so that the version with the `end_tag` takes precedence:

```suggestion
(element
  (start_tag)
  (_)* @xml_element.inside
  (end_tag)
. ) @xml_element.around

(element
  (start_tag)
  (_)* @xml_element.inside
. ) @xml_element.around
```

Otherwise a snippet like:

```html
    <a href="print.html" title="Print this book" aria-label="Print this book">
        <i id="print-button" class="fa fa-print"></i>
    </a>
```

If the cursor is on the whitespace at the start of line 2, `mie` will select the `</a>`

the-mikedavis (commented):
XML has been added so we could add queries for that as well (#4518), or leave that for a separate PR

Ansimorph:
Thanks so much for this PR!

I tested the changes (rebased the current main) and it works reliably for around-queries.

For inside-queries it works when the cursor is already on the content nested in a tag. 
When the cursor is resting on an html attribute such as `class="sr-only"`, `mie` selects nothing.
When the cursor rests on the tag name of a closing tag, `mie` selects the tag name itself.

I would expect the element the cursor is in to be the context the selection is applied to:

```html
<section>
  <p [CURSOR] class="sr-only"> text </p>
</section>
```

`mie` selects `text`
`mae` selects `<p class="sr-only"> text </p>`

Currently only `mae` works that way

ff2400t:
The `mei` queries still don't work as I want them to. 
```html
<section>
  <p [CURSOR] class="sr-only"> text </p>
</section>
```
It would be ideal in the above case for `mei` to select the text inside the paragraph element. But for now it selects the whole paragraph element.
Hopefully someone who understand treesitter queries can help in that.

Furthermore, i think it would be great if the Text selection is trimmed of whitespace by default but that should be part of another PR

ff2400t:
So in the end i had to modify how textobjects are captured from rust to capture the `element.inner` as intended. This is also how nvim-treesitter-textobjects does the capture for these.

helix-core/src/textobject.rs
```diff
@@ -260,18 +260,47 @@ pub fn textobject_treesitter(
...
+                _ => outer_node,
+            }
+        } else {
+            let capture_name = format!("{}.{}", object_name, textobject); // eg. function.inner
+            lang_config
+                .textobject_query()?
+                .capture_nodes(&capture_name, slice_tree, slice, &mut cursor)?
+                .filter(|node| node.byte_range().contains(&byte_pos))
+                .min_by_key(|node| node.byte_range().len())?
+        };
```
AlexanderBrevig, line 303: I like the block comment explanation, however: There's now three almost identical `format!` calls and three chains of `lang_config.textobject_query()?.capture_nodes`. The `// et. function.inner` does not feel correct and in general the intent is kind of hard to get a grasp on (reading the code only).

Maybe you can refactor it a bit DRYer?

helix-core/src/textobject.rs
```diff
@@ -260,18 +260,47 @@ pub fn textobject_treesitter(
     object_name: &str,
     slice_tree: Node,
     lang_config: &LanguageConfiguration,
+    better_capture: bool,
```
AlexanderBrevig, line 263: I think this should more closely express what's going on. Why would you ever not use the `better_capture`? Maybe even use an enum here, to make the call site more explicit/readable?

AlexanderBrevig (changes requested):

helix-core/src/textobject.rs
ff2400t: ohh i would love make it DRY but this my first time writing rust so I don't know if making this a Closure is the rust way of Doing.

helix-core/src/textobject.rs
ff2400t: Yeah that's Temporary name, I thought someone reviewing the PR could suggest a Better name

helix-core/src/textobject.rs
ff2400t: Okay, I tried to extract that into a Closure, but i had issues returning an Iterators from the closure and then i had some lifetimes issues on top of that.
Extracting that to a separate function resulted in a similar lifetimes issue, which i don't understand

pascalkuthe:

PR #3284, opened 2022-08-01
Improve Python queries, add locals

ChrHorn:
Just improved some things I noticed.


the-mikedavis (approved):
These look good, thanks for working on them!

PR #3285, opened 2022-08-01
theme: Add gruvbox alternate contrast modes

kevinastock:
These are the same as the existing themes, except bg0 is changed as shown here: https://github.com/morhetz/gruvbox


kevinastock:
I wasn't sure how I should set/modify the author tag, so I copied them from existing files since I haven't done any of the hard work here. Let me know if that needs changed.

kevinastock:
I hadn't seen that one, waiting for that seems great. I'll close this and send a new PR once that merges.

PR #1623, opened 2022-02-06
Show git diff signs in gutter

sudormrfbin:
Shows the changes to the current file symbolically in the gutter:

| Without diagnostics | With diagnostics |
| --- | --- |
| Screenshot_2022-02-07_00-06-27 | Screenshot_2022-02-07_00-09-14 |

It looks a bit too cramped with diagnostics, so maybe the reserved space should be increased but that will make the gutter permanently too big (there's currently no way to "toggle" a gutter component). Changes are reflected only when the file is saved. I would like to show an accurate diff even if the file is unsaved, but this should be a good starting point.

The symbols are hard coded for now to keep the PR simple, and a follow up PR will make it configurable. The default might also need changing (though I would like to avoid too much bike shedding ;)

Supersedes #467.

PR #3288, opened 2022-08-01
undo change, no magically added space please, just add Spacer by default

aikomastboom:
I believe the idea for https://github.com/helix-editor/helix/pull/2996 was to be able to add a spacer if desired.

Looking ahead to https://github.com/helix-editor/helix/pull/1623 I prefer to have that space filled by git status.

This PR reverts functionality to give this choice back to the user.

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -101,6 +101,7 @@ impl View {
     pub fn new(doc: DocumentId, gutter_types: Vec<crate::editor::GutterType>) -> Self {
         let mut gutters: Vec<(Gutter, usize)> = vec![];
         let mut gutter_offset = 0;
+        let mut last_gutter_is_spacer = false;
         use crate::editor::GutterType;
         for gutter_type in &gutter_types {
             let width = match gutter_type {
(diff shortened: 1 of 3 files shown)


helix-view/src/view.rs
```diff
@@ -118,9 +118,6 @@ impl View {
                 width as usize,
             ));
         }
-        if !gutter_types.is_empty() {
-            gutter_offset += 1;
-        }
```
the-mikedavis, line 123: This is a margin that's meant to separate the gutters and the code, themed as background

See #3111

the-mikedavis (commented):
> Looking ahead to #1623 I prefer to have that space filled by git status.

#1623 is a gutter component like diagnostics or line-numbers, not part of the margin

archseer:
See the discussion here https://github.com/helix-editor/helix/issues/3111

aikomastboom:
Ah, I see.. so now we need an extra option to set the width of the margin between gutter and tekst. As the automatic adding of 1 is still reverting #2996 :)

Issue #2684, opened 2022-06-06
Helix changes terminal cursor style


GandelXIV:
### Summary

After quitting helix my terminal cursor changes from `Beam` (my default) to `Block` (helix default). I had the same issue of vim but i fixed it by adding `set guicursor=` to my `.vimrc`. However, i could not find a way to do that in helix.

### Reproduction Steps

1. Open helix
2. Exit
3. Cursor style has changed

### Platform

Linux

### Terminal Emulator

Alacritty

### Helix Version

helix 22.05

DivergentClouds:
I found the [problem line](https://github.com/helix-editor/helix/blob/55b45ec4a4cb958b241a93cc7c3f4e499379890e/helix-term/src/application.rs#L791).

The escape code written to the terminal to reset the cursor (`\x1B[2 q` ) instead sets the cursor to be a static block. 

[The specification](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_) defines the cursor set sequence as follows:

```
CSI Ps SP q
          Set cursor style (DECSCUSR), VT520.
            Ps = 0  ⇒  blinking block.
            Ps = 1  ⇒  blinking block (default).
            Ps = 2  ⇒  steady block.
            Ps = 3  ⇒  blinking underline.
            Ps = 4  ⇒  steady underline.
            Ps = 5  ⇒  blinking bar, xterm.
            Ps = 6  ⇒  steady bar, xterm.
```

`CSI` is `ESC [`.
`SP` is space.

The simplest solution is probably to get the terminfo on program start and restore it on program end. I'd submit a PR but I don't know Rust.

archseer:
Leaving this open since we'll probably be adding terminfo soon (@sudormrfbin needs it in another PR)

PR #3289, opened 2022-08-01
Fix terminal cursor shape reset

ericpko:
I changed [this line](https://github.com/helix-editor/helix/blob/55b45ec4a4cb958b241a93cc7c3f4e499379890e/helix-term/src/application.rs#L791) by changing the 2 to a 0.

The previous escape code `\x1B[2 q` sets the cursor to be a static block according to [this specification](https://invisible-island.net/xterm/ctlseqs/ctlseqs.html#h3-Functions-using-CSI-_-ordered-by-the-final-character_s_). The new escape code `\x1B[0 q` restores the cursor to the user's default.

To be honest, I'm not sure if this will fix the issue for everyone. It fixes it for me on arch linux using gnome and with the alacritty terminal. With this change, when I exit out of helix, my terminal cursor is restored to a blinking underscore.

@DivergentClouds found the solution, which he shared in issue #2684. 

There are also several other terminal-based editors that have had this same issue of not restoring the user's terminal cursor style, such as [here](https://github.com/alacritty/alacritty/issues/2839#issuecomment-537619961) and [here](https://github.com/microsoft/terminal/issues/1604).

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -801,7 +801,7 @@ impl Application {
     fn restore_term(&mut self) -> Result<(), Error> {
         let mut stdout = stdout();
         // reset cursor shape
-        write!(stdout, "\x1B[2 q")?;
+        write!(stdout, "\x1B[0 q")?;
         // Ignore errors on disabling, this might trigger on windows if we call
         // disable without calling enable previously
         let _ = execute!(stdout, DisableMouseCapture);


archseer:
This is probably a better default to reset back to, but we do want a better solution in the future that reads `terminfo` at startup. I'll merge this for now and we'll fix it more properly once terminfo is added

DivergentClouds:
Just want to give you a heads up that I use they/them and not he/him, it'd mean a lot if you edited your message accordingly, thank you

PR #3295, opened 2022-08-01
Languages: add LISP (lisp, cl, mud, el, asd) language

AlexanderBrevig:
This adds support for lisp languages, using [cl-lsp](https://github.com/cxxxr/cl-lsp) and [tree-sitter-lisp](https://github.com/alexanderbrevig/tree-sitter-lisp) which is mine (PRs welcome)

Sample with LSP response
image


the-mikedavis:
Why not use https://github.com/theHamsta/tree-sitter-commonlisp? I'm not super familiar with the differences but I also know of https://github.com/Wilfred/tree-sitter-elisp for emacs lisp. Both are quite well established, for example [used in difftastic](https://difftastic.wilfred.me.uk/languages_supported.html) and nvim

archseer:
We also already include the scheme grammar, it may support Common Lisp

AlexanderBrevig:
> Why not use https://github.com/theHamsta/tree-sitter-commonlisp? I'm not super familiar with the differences but I also know of https://github.com/Wilfred/tree-sitter-elisp for emacs lisp. Both are quite well established, for example [used in difftastic](https://difftastic.wilfred.me.uk/languages_supported.html) and nvim

Seems an eternity ago, but I tried both.
Somewhere along the line I got confused and thought I had to make a repo have queries and the grammar so I took the opportunity to learn a bit too soon.

I'll rework to use that commonlisp grammar. 
I'd like to have lisp configured with `cl-lsp` in hopes that it gets better some day.

Or do you want me to close this? And people can use `scheme` with their own LSP override? (When I think about it, that works for me)

AlexanderBrevig:
```toml
[[language]]
name = "scheme"
file-types = [ "lisp", "cl", "el", "mud", "asd" ]
language-server = { command = "cl-lsp", args = [ "stdio" ] }

This is just as good as what I started on. Thanks for the pointers!

David-Else:
@AlexanderBrevig Could you add install instructions to https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers ?

PR #3301, opened 2022-08-02
Inherit javascript/typescript from common 'ecma' queries

the-mikedavis:
Looks like https://github.com/helix-editor/helix/pull/3280 inadvertently ended up breaking typescript highlights because of typescript's `; inherits: javascript` line. ~Here I've replaced that `; inherits javascript` with as much of the javascript highlights as the query analysis will accept~

This should fix typescript and tsx query analysis (which panics on open).


EricCrosson (approved):
Yep, I'm hitting this on my side too. Quick on the draw as usual, @the-mikedavis ! Many thanks :bow:

ChrHorn:
Ah, now I understand why they did it in that way at the nvim-tresitter repo.

Both [javascript](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/javascript/highlights.scm) and [typescript](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/typescript/highlights.scm) inherit from [ecma](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/ecma/highlights.scm) to avoid code duplication. 

Should we do the same?

the-mikedavis:
Yeah good call, let's keep sharing as much as we can under ecma 👍

PR #3306, opened 2022-08-02
For removing empty scratch buffers, check that a buffer is empty rather than unmodified

### LINKED ISSUE #3305: Initial scratch buffer is discarded when non-empty and unmodified
### Summary

The initial scratch buffer created when opening `hx` without a filename may be erroneously closed when navigated away from. This is expected behavior when the scratch buffer is empty and unmodified but it can occur when the scratch buffer is non-empty and unmodified by piping in initial text.

https://github.com/helix-editor/helix/discussions/3118

### Reproduction Steps

Fill the initial scratch buffer with files for example

```
find . -name \*.md | hx
```

Then use `gf` to open up one of those buffers. The buffer picker (`Space-b`) then shows

If you leave some modifications in the scratch buffer before using `gf` then the scratch buffer stays open.

(This is not reproducible on macos because piping in is not currently supported https://github.com/helix-editor/helix/blob/69f84e25165d62d0f38bcc80bf9108f9c0802cad/helix-term/src/application.rs#L197)

### Helix log

_No response_

### Platform

Linux (NixOS 21.11)

### Terminal Emulator

Kitty 0.25.2

### Helix Version

22.05-351-g69f84e25

Omnikar:
This should fix #3305, although I am on a Mac, so I cannot test it myself.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -831,8 +831,8 @@ impl Editor {
                 // If the current view is an empty scratch buffer and is not displayed in any other views, delete it.
                 // Boolean value is determined before the call to `view_mut` because the operation requires a borrow
                 // of `self.tree`, which is mutably borrowed when `view_mut` is called.
-                let remove_empty_scratch = !doc.is_modified()
-                    // If the buffer has no path and is not modified, it is an empty scratch buffer.
+                let remove_empty_scratch = doc.text() == "\n"
+                    // If the buffer has no path and contains only the default newline, it is an empty scratch buffer.
                     && doc.path().is_none()
                     // If the buffer we are changing to is not this buffer
                     && id != doc.id


Omnikar:
Actually, I feel like we should instead be marking buffers created from stdin as modified, since they contain unsaved content and should be protected from being lost, right?

the-mikedavis:
I'm not sure how useful it would be (for example for undoing) but it's intuitive to me that the text from stdin would be a revision in the history as well

Omnikar:
> I'm not sure how useful it would be (for example for undoing) but it's intuitive to me that the text from stdin would be a revision in the history as well

There's of course also the aspect of preventing exiting the editor if the content isn't saved somewhere.

Omnikar:
How should I go about marking stdin scratch buffers as modified? Create them as an empty scratch buffer then add the text given by stdin as a change?

Omnikar:
In the `new_file_from_stdin` function in `helix_view::editor`, creating an empty `Document` using `Document::default()` rather than the current document creation via `Document::from` does not take the `encoding` value, and the `.set_encoding` method seems to take a different type.

the-mikedavis:
I think you should just be able to use `doc.line_ending = encoding`, no?

Omnikar:
> I think you should just be able to use `doc.line_ending = encoding`, no?

`encoding` is of type `&encoding::Encoding`.
`Document::set_encoding` takes type `&str`.
`Document::line_ending` is of type `helix_core::line_ending::LineEnding`.

the-mikedavis:
Oh whoops I meant `doc.encoding` 😅. Looks like that's a private field though and `set_encoding` doesn't take an encoding type as you say. Given that `set_encoding` is only used by `helix_term::commands::typed::set_encoding`, I think it's best to refactor that function to do the `encoding::Encoding::for_label` call and have `helix_view::document::Document::set_encoding` just take an `Encoding` (and then it doesn't need to return a result).

IMO it's really `helix_term::commands::typed::set_encoding`'s responsibility to parse the encoding anyways.

PR #3307, opened 2022-08-02
Resolve clippy lints

Omnikar:
Clippy was complaining about using `format!` to append to an existing `String`.


archseer:
You're probably on a nightly version right?

Omnikar:
> You're probably on a nightly version right?

No, I just have stable cargo v1.62.1.

PR #3308, opened 2022-08-02
Add Doom's Acario dark theme.

Rinnray:
This theme is ported from [acario themes](https://github.com/doomemacs/themes/blob/master/themes/doom-acario-dark-theme.el) from Doom Emacs.

I've however only focused on the dark one as that's the one I'm using.

### DIFF
--- a/runtime/themes/doom_acario_dark.toml
+++ b/runtime/themes/doom_acario_dark.toml
@@ -42,9 +42,13 @@ namespace = { fg = 'red' }
 'ui.background'= { bg = 'bg' }
 'ui.cursor' = { bg = 'orange', fg = 'bg-alt' }
 'ui.cursor.match' = { fg = 'red', modifiers = ['bold'] }
+'ui.cursorline' = { bg = 'bg-alt' }
 'ui.linenr' = { fg = 'base4' }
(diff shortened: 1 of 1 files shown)


AlexanderBrevig:
Nice!

I ran it through https://github.com/helix-editor/helix/pull/3234 and got 
```json
"acario":[
	"ui.cursorline"
]
```

It may be argued that `ui.statusline.normal|insert|select` should be sepecified as well, in order to adhere to the `editor.color-modes` setting?

Rinnray:
I didn't see anything different when I defined those, but I added them!
Statusline is the same color either way.

AlexanderBrevig:
> I didn't see anything different when I defined those, but I added them! Statusline is the same color either way.

Both of them are behind setting. You must enable color-modes and cursorline (I just realized this may not be available if you use an official release. Thanks for adding anyway)

the-mikedavis (approved):
Looks nice, thanks! 🎨

Could you add a screenshot to the themes wiki? https://github.com/helix-editor/helix/wiki/Themes

Rinnray:
> Looks nice, thanks! art
> 
> Could you add a screenshot to the themes wiki? https://github.com/helix-editor/helix/wiki/Themes

Will do!

PR #2208, opened 2022-04-21
Visible whitespace only for selections

tomKPZ:

PR #3309, opened 2022-08-02
Add keybinds for toggling common settings

### LINKED ISSUE #2849: Toggle features
This may be more useful when certain feature requests currently being worked on for core, become available. But I'd think it would be nice to have the ability to toggle certain UI elements, such as white-space indicators, file/symbol tree, a future terminal buffer?, etc.

Each toggle could either be a key bind set by the user, or set to a mode mapping like `space + t`, to get a list of potential toggle options.

Open to knowing other people's solutions to quickly performing these tasks.

the-mikedavis:
https://user-images.githubusercontent.com/21230295/182493839-da8e7a34-e3b0-4633-bca1-a6c3e67995b5.mp4

This adds a `<space>t` binding that can toggle a few select settings:

* visible whitespace
* indent guides
* cursorline
* a ruler at 80

It's has limitations:

* it's not configurable what is toggled[^1]
* the value of the setting you had before toggling is not possible to restore in all cases without using `:set`

[^1]: IMO this should wait for a more general purpose config language. It's possible to do it in TOML but it wouldn't be particularly elegant. It could be pretty elegant to do it if keymaps were written in something like lisp :P

Closes #2849

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4872,9 +4872,10 @@ fn toggle_setting(cx: &mut Context) {
                 'i' => config.indent_guides.render = !config.indent_guides.render,
                 'c' => config.cursorline = !config.cursorline,
                 'r' => {
-                    config.rulers = match config.rulers.as_slice() {
-                        [80] => vec![],
-                        _ => vec![80],
+                    config.rulers = if config.rulers.is_empty() {
+                        vec![80]
+                    } else {
+                        vec![]
                     };
                 }
                 'p' => {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4886,6 +4886,14 @@ fn toggle_setting(cx: &mut Context) {
                         _ => Enable(false),
                     }
                 }
+                'l' => {
+                    use helix_view::editor::LineNumber::{Absolute, Relative};
+
+                    config.line_number = match config.line_number {
+                        Absolute => Relative,
+                        Relative => Absolute,
+                    };
+                }
                 _ => cx.editor.set_error(format!("Unknown setting '{}'", ch)),
             }
 
@@ -4908,6 +4916,8 @@ fn toggle_setting(cx: &mut Context) {
         ("c", "Cursorline"),
         ("r", "Ruler at 80 columns"),
         ("p", "Auto-pairs"),
+        ("l", "Line-number (absolute/relative)"),
+        ("l", "Line-number (absolute, relative)"),
     ];
 
     cx.editor.autoinfo = Some(Info::new(
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -432,6 +432,7 @@ impl MappableCommand {
         record_macro, "Record macro",
         replay_macro, "Replay macro",
         command_palette, "Open command palette",
+        toggle_setting, "Toggle setting",
     );
 }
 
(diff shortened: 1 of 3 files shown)


AlexanderBrevig:
Cool! :+1: 
I would personally love line number relative toggles.

BTW I restored my config with a simple `:config-reload`. At first, I thought the toggle would modify my config. Glad that's not the case.


(a lisp for configuration would be awesome)

helix-term/src/commands.rs
```diff
@@ -4845,3 +4846,75 @@ fn replay_macro(cx: &mut Context) {
...
+                    config.whitespace.render = match config.whitespace.render {
+                        Basic(None) => Basic(All),
+                        _ => Basic(None),
+                    };
+                }
+                'i' => config.indent_guides.render = !config.indent_guides.render,
+                'c' => config.cursorline = !config.cursorline,
+                'r' => {
+                    config.rulers = match config.rulers.as_slice() {
+                        [80] => vec![],
```
AlexanderBrevig, line 4876: If the user has a ruler defined, then the toggle will not hide it but rather move it to 80.

AlexanderBrevig (approved):
Very cool! I don't know if it's OK for me to do the review, but here I am anyway :)

David-Else:
@the-mikedavis  Please can we toggle the line numbers on and off? I see this as much more important than line number relative toggles. A perfect solution would be to cycle through all three states with `l`, is this possible ?

If you are writing docs in markdown you never need line numbers, and even when coding I personally only use line numbers when some legacy linter only gives line numbers of errors rather than an LSP that can take you right to the problem.

AlexanderBrevig:
> @the-mikedavis Please can we toggle the line numbers on and off? I see this as much more important than line number relative toggles. A perfect solution would be to cycle through all three states with `l`, is this possible ?

Kind of feels weird with a ternary toggle. Maybe `L` for on/off and `l` for absolute/relative?

PS: relative line numbers will make `n<operation-that-shifts-vertically>` much easier.
You can immediately see that you need to go `nC` to create `n` carets for modifying a config file or some such. We all do things differently of course, but this has been a productivity boost for me (yeah, we're measuring sub seconds here 😉)

David-Else:
> Kind of feels weird with a ternary toggle. Maybe `L` for on/off and `l` for absolute/relative?

I think it would feel cool and easier to use than `L` and `l`, do you think it would be weird as it would be the only ternary and doesn't fit in, or just that toggles should be binary?

AlexanderBrevig:
> > Kind of feels weird with a ternary toggle. Maybe `L` for on/off and `l` for absolute/relative?
> 
> I think it would feel cool and easier to use than `L` and `l`, do you think it would be weird as it would be the only ternary and doesn't fit in, or just that toggles should be binary?

After thinking about it more, I think you're right. Probably cool and natural. I'll edit my comment with a strike through.

the-mikedavis:
I don't have any qualms about this command cycling between a list of options. IMO, if it were configurable then the options would be a list. If visible whitespace had the Selection variant implemented https://github.com/helix-editor/helix/pull/2208, the toggle would include that. Similar with https://github.com/helix-editor/helix/pull/3242.

Implementation-wise it's not very clean to disable/enable the line-number gutter because it's controlled by the gutters configuration and not `helix_view::editor::LineNumber`, so I'd prefer to not add a toggle for it. The use-case of "If you are writing docs in markdown you never need line numbers" sounds like it's somewhat better suited by #3207.

the-mikedavis:
Despite this working as a proof-of-concept I think it would be better to be able to accomplish this with custom keybinds once there is scripting support / better config.

In the meantime people can use this PR as a guide if they want to build something similar locally

Issue #3217, opened 2022-07-27
Accessibility Improvement - High Contrast Theme


jquesada2016:
#### Describe your feature request

As a legally blind developer, one most often needs to create a custom theme in order to see everything. I believe it would be a really good idea to include a default high contrast theme for both dark and light modes out of the box. For dark mode, VS Code's default high contrast theme is very good, as is Android Studio/IDEA's. For light mode, Windows ships with a light high contrast theme which is also excellent.

GreasySlug:
I created #3238, and could anyone help me, please?

PR #3312, opened 2022-08-03
Add dark high contrast theme refer to vscode

GreasySlug:
Issues: #3217 
Discussions: #3238 
I created a dark high contrast theme refer to vscode dark high contrast theme.

It is hard for me to create a light theme and I did not get a feedback for it as mentioned in the discussion.
Therefore, the it is not included in this PR.
For now, I've already created it.


d4h0:
Thanks for creating this theme, @GreasySlug!

It is much better for my crappy eyes than the other themes.

One problem I see, are comments. The contrast is way too low (I almost can't read comments).

Couldn't comments be white or almost white?

To me, comments are a very important part of source code, and I don't really understand why they are almost always treated as unimportant (cursive, and hard-to-read gray color).

GreasySlug:
You are certainly right.
No other high contrast theme used gray and italics.
I have a feeling that most of them use green, is green correct?

fixed_comment_color

d4h0:
> I have a feeling that most of them use green, is green correct?

I'm not sure if green is often used for high-contrast themes, but there seem to be already two green shades used in the theme. And while the darker green shade from your screenshot above is certainly a bit easier to ready than the current dark gray, I believe white (or a very light gray) would be easier to read (and isn't used for anything besides punctuation characters).

d4h0:
Thanks, @GreasySlug, this looks great!

Btw., I think italic text is more difficult to read than non-italic text, but I'm not completely against using italic for comments. 

Italic has the advantage that it adds some "visual contrast" (not sure how to call it), and non-italic text has the advantage that it is a bit easier to read.

I'm not sure which one is better.

That being said, I kind of like the non-italic comments of this theme (this is the first time that I have tried a theme with comments that are not italic).

GreasySlug:
Thank you for your feedback, @d4h0.
The only reason Italic was used was that it was left in place when the template theme was copied.
I removed it because it is basically unused in other editors.

the-mikedavis (approved):

the-mikedavis:
cc: @jquesada2016

PR #3313, opened 2022-08-03
Fix tab highlight when tab is partially visible

### LINKED ISSUE #2992: Background of selected, partially visible tabs is un-highlighted
### Summary

The background highlighting of the selected range doesn't extend all the way to the left as it should when tabs are partially displayed as the leftmost visible character.

screenshot

Notice how, in the screenshot above, the background of the leftmost character is not highlighted, even though the partially visible tab that occupies that space is part of the selection.

It's definitely an edge case, but it might be more noticeable with especially wide tab widths and small windows. 

### Reproduction Steps

 

I tried this:

1. `hx`
2. Position the view so a selected tab is partially visible.

I expected this to happen:

The background of the partially visible tab is highlighted, just like the rest of the selection.

Instead, this happened:

The background isn't highlighted.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

kitty 0.25.2

### Helix Version

b61221107068160c241858a530f6fa4144566122

A-Walrus:
Resolves #2992.

Works for tabs, as well as other wide graphemes such as `Ideographic Space "　"`

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -473,10 +473,10 @@ impl EditorView {
                     use helix_core::graphemes::{grapheme_width, RopeGraphemes};
 
                     for grapheme in RopeGraphemes::new(text) {
-                        if LineEnding::from_rope_slice(&grapheme).is_some() {
-                            let out_of_bounds = visual_x < offset.col as u16
-                                || visual_x >= viewport.width + offset.col as u16;
+                        let out_of_bounds = visual_x < offset.col as u16
+                            || visual_x >= viewport.width + offset.col as u16;
 
+                        if LineEnding::from_rope_slice(&grapheme).is_some() {
                             if !out_of_bounds {
                                 // we still want to render an empty cell with the style
                                 surface.set_string(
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -530,19 +530,33 @@ impl EditorView {
...
+                                    style,
                                 );
+                            } else if cut_off_start != 0 && cut_off_start < width {
+                                // partially on screen
+                                let rect = Rect::new(
+                                    viewport.x as u16,
+                                    viewport.y + line,
+                                    (width - cut_off_start) as u16,
+                                    1,
+                                );
```
the-mikedavis, line 556: Testing this out with something non-whitespace-like but multi-width like "😀", we render an empty space instead of any character (because of the `rect` here). I think that's a good choice but I wonder if there's anything better to signal that a multi-width character is getting chopped off here?

I would say `…` but then we're rendering something completely different than the actual source. Maybe the plain rect is the best choice? ¯\\\_(ツ)_/¯

the-mikedavis (approved):

helix-term/src/ui/editor.rs
A-Walrus: I considered rendering a character to indicate that a wide grapheme has been cut off, but I think keeping it simple and showing nothing is the right approach.
If we were to render a character it should probably be configurable which char to use, (like how it is with whitespace) but I think that's overkill.

helix-term/src/ui/editor.rs
```diff
@@ -530,19 +530,33 @@ impl EditorView {
                                 (grapheme.as_ref(), width)
                             };
 
+                            let style = if is_whitespace {
+                                style.patch(whitespace_style)
+                            } else {
+                                style
+                            };
```
archseer, line 537: Can we inline this into both branches? I don't mind the duplication.

This is a very hot loop that currently gets ran for *all* characters on the line, including the ones past the right edge of the screen. We'd waste time calculating the style there.

PR #3315, opened 2022-08-03
Add a .ignore file to make ripgrep more useful

groves:
I've been using this .ignore file while working on Helix to let me use ripgrep without getting a big block of spurious result in highlight.js. I thought it might be useful for others. No worries if it isn't, I can just add it to my .git/info/exclude.

### DIFF
--- a/.ignore
+++ b/.ignore
@@ -1,0 +1,5 @@
+# Things that we don't want ripgrep to search that we do want in git
+# https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering
+
+# This is some minified JS that's straight from mdbook
+book/theme/highlight.js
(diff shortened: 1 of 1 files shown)


.ignore
```diff
@@ -0,0 +1,5 @@
+# Things that we don't want ripgrep to search that we do want in git
+# https://github.com/BurntSushi/ripgrep/blob/master/GUIDE.md#automatic-filtering
+
+# This is some minified JS that's straight from mdbook
```
the-mikedavis, line 4:
```suggestion
# Minified JS vendored from mdbook
```

but of a nit sorry 🙃

the-mikedavis:
Helix also takes `.ignore` files into account for things like global search (`Space-/`): https://docs.helix-editor.com/master/configuration.html#editorfile-picker-section

the-mikedavis (approved):

PR #3316, opened 2022-08-03
Fix LF line-endings

groves:
These are the only two files in the repo with carriage returns:

```
cgroves@jojr ~/d/helix (master)> rg '\r' --count
helix-view/src/keyboard.rs:153
helix-view/src/graphics.rs:665
```

I'm assuming that means it's not on purpose?

They kept showing up in my commits to keyboard.rs, so I figured it was worth aligning with everything else.


AlexanderBrevig:
I fixed this in https://github.com/helix-editor/helix/pull/3318 with `.gitattributes` so that it probably will not appear again.

the-mikedavis:
Groves did beat you by ~20min so I'll take this one first :P

But yeah, in the long-term this should be solved by gitattributes

groves:
Hah, that's awesome that it was in there for over a year and was targeted by two PRs within 30 minutes of each other.

PR #3317, opened 2022-08-03
Fix php hightlight

erasin:
namespace
fucntion
a
b

### DIFF
--- a/runtime/queries/php/highlights.scm
+++ b/runtime/queries/php/highlights.scm
@@ -8,6 +8,14 @@
 (named_type (name) @type) @type
 (named_type (qualified_name) @type) @type
 
+(namespace_definition
+  name: (namespace_name(name)) @constructor
+)
+
+(subscript_expression
+  (variable_name(name) @constant)
+)
+
 ; Functions
 
 (array_creation_expression "array" @function.builtin)
@@ -17,7 +25,7 @@
   name: (name) @function.method)
 
 (function_call_expression
-  function: (qualified_name (name)) @function)
+  function: (_) @function)
 
 (scoped_call_expression
   name: (name) @function)
@@ -27,6 +35,7 @@
 
 (function_definition
   name: (name) @function)
+
 
 ; Member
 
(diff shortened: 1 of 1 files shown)


runtime/queries/php/highlights.scm
```diff
@@ -8,6 +8,14 @@
 (named_type (name) @type) @type
 (named_type (qualified_name) @type) @type
 
+(namespace_definition
+  name: (namespace_name(name)) @constructor
+)
```
the-mikedavis, line 13: I'm not very familiar with php but wouldn't `@namespace` be a more appropriate capture here?

runtime/queries/php/highlights.scm
```diff
@@ -8,6 +8,14 @@
 (named_type (name) @type) @type
 (named_type (qualified_name) @type) @type
 
+(namespace_definition
+  name: (namespace_name(name)) @namespace
+)
+
+(subscript_expression
+  (variable_name(name) @constant)
```
archseer, line 16: Is this really a constant?

erasin: it's not constant, it is name Superglobals in php, like constant ,is global . after php 8.1 ,  some become read-only like $GLOBALS.  Superglobals is diffent with variable , In practical applications, most of them are read-only, except magic.  It can be regarded as special constant。

may be mark this with `variable.builtin`

erasin: PHP extensions will also append some Superglobals.
I mark `constant.builtin`.

runtime/queries/php/highlights.scm
```diff
@@ -8,6 +8,14 @@
 (named_type (name) @type) @type
 (named_type (qualified_name) @type) @type
 
+(namespace_definition
+  name: (namespace_name(name)) @namespace
+)
```
the-mikedavis, line 13:
```suggestion
(namespace_definition
  name: (namespace_name (name) @namespace))
```

The current pattern is ignored because the `((name) @constructor ...` stanza on L56 is more specific.

runtime/queries/php/highlights.scm
```diff
@@ -8,6 +8,14 @@
...
+
+(subscript_expression
+  (variable_name(name) @constant.builtin)
+)
```
the-mikedavis, line 17: for code like

```php
$array = ['One', 'Two', 'Three'];
echo $array[0]; // => "One"
```

This now highlights `arrays` in L2 as `constant.builtin` but I think the `variable` capture from before was more accurate

erasin:
```
(subscript_expression
  (variable_name(name) @constant.builtin
    (#match? @constant.builtin "^_?[A-Z][A-Z\\d_]+$")
  )
)
```
it‘s seem ok

the-mikedavis (approved):
Looks great, thanks!

PR #3318, opened 2022-08-03
Add .gitattributes

AlexanderBrevig:
Based on https://github.com/alexkaratarakis/gitattributes/blob/master/Rust.gitattributes

This also changed CRLF => LF for `helix-view/src/graphics.rs` and `helix-view/src/keyboard.rs`


the-mikedavis:
Let's add

```
book/theme/highlight.js linguist-vendored
```

to tell github about the vendored javascript from mdbook

AlexanderBrevig:
Added linguist-vendored and removed the CRLF => LF conversion

the-mikedavis (approved):

PR #3322, opened 2022-08-03
Respect SIG_IGN for SIGTSTP

### LINKED ISSUE #3321: Respect SIG_IGN for SIGTSTP (Terminal SToP)

When using shells without job control (such as https://elv.sh), respect the `SIG_IGN` that's set for `SIGTSTP` so `C-z` does not suspend Helix. 

When I C-z now, the only way out (that I've found) is to do a hard shutdown of the terminal. 

I have lost quite a lot of work because of this. (Yes, I know I _could_ use a different shell but I'm weird like that).

https://github.com/elves/elvish/pull/1017

AlexanderBrevig:
This implementation is probably ugly.

Any pointers to make it cleaner is much appreciated!

It works, tested with `elvish`, `bash`, `zsh` and `sh`. It correctly disables `C-z` for `elvish` but not the others.

Fixes #3321

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -43,7 +43,6 @@ signal-hook = "0.3"
 tokio-stream = "0.1"
 futures-util = { version = "0.3", features = ["std", "async-await"], default-features = false }
 arc-swap = { version = "1.5.1" }
-libc = "0.2"
 
 # Logging
 fern = "0.6"
@@ -73,6 +72,7 @@ retain_mut = "0.1.7"
 
 [target.'cfg(not(windows))'.dependencies]  # https://github.com/vorner/signal-hook/issues/100
 signal-hook-tokio = { version = "0.3", features = ["futures-v0_3"] }
+libc = "0.2"
 
 [build-dependencies]
 helix-loader = { version = "0.6", path = "../helix-loader" }
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -43,6 +43,7 @@ signal-hook = "0.3"
 tokio-stream = "0.1"
 futures-util = { version = "0.3", features = ["std", "async-await"], default-features = false }
 arc-swap = { version = "1.5.1" }
+libc = "0.2"
 
 # Logging
 fern = "0.6"
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -209,6 +209,7 @@ impl Application {
         #[cfg(windows)]
         let signals = futures_util::stream::empty();
 
+        #[cfg(not(windows))]
         let mut app_signals = vec![signal::SIGTSTP, signal::SIGCONT];
         unsafe {
             // X platform constants
@@ -223,7 +224,6 @@ impl Application {
                 ENABLE_SIGTSTP.set(true).unwrap();
             }
         }
-        #[cfg(not(windows))]
         let signals = Signals::new(app_signals).context("build signal handler")?;
 
         let app = Self {
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -43,12 +43,6 @@ type Signals = futures_util::stream::Empty<()>;
 
 const LSP_DEADLINE: Duration = Duration::from_millis(16);
 
-// Allow compilation on Windows by forwarding these
-pub const HX_SIGTSTP: libc::c_int = 20;
-const HX_SIGCONT: libc::c_int = 18;
-const HX_SIG_IGN: libc::sighandler_t = 1;
-const HX_SIG_ERR: libc::sighandler_t = 0;
-
 pub struct Application {
     compositor: Compositor,
     pub editor: Editor,
@@ -131,10 +125,13 @@ impl Application {
         let config = Arc::new(ArcSwap::from_pointee(config));
 
         // Handle signals now, so we can pass flag to the Editor
-        let enable_suspend = check_enable_suspend();
+        #[cfg(windows)]
+        let enable_suspend = false;
         #[cfg(windows)]
         let signals = futures_util::stream::empty();
 
+        #[cfg(not(windows))]
+        let enable_suspend = check_enable_suspend();
         #[cfg(not(windows))]
         let signals = get_available_signals(enable_suspend)?;
 
@@ -856,15 +853,16 @@ impl Application {
     }
 }
 
+#[cfg(not(windows))]
 fn check_enable_suspend() -> bool {
-    unsafe { libc::signal(HX_SIGTSTP, HX_SIG_IGN) == HX_SIG_ERR }
+    unsafe { libc::signal(libc::SIGTSTP, libc::SIG_IGN) == 0 }
 }
 
 #[cfg(not(windows))]
 fn get_available_signals(enable_suspend: bool) -> Result<Signals, std::io::Error> {
-    let mut app_signals = vec![HX_SIGCONT];
+    let mut app_signals = vec![libc::SIGCONT];
     if enable_suspend {
-        app_signals.push(HX_SIGTSTP);
+        app_signals.push(libc::SIGTSTP);
     }
     Signals::new(app_signals)
 }
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -859,15 +859,14 @@ fn get_signals() -> Signals {
     const HX_SIG_IGN: libc::sighandler_t = 1;
     const HX_SIG_ERR: libc::sighandler_t = 0;
     let mut app_signals = vec![HX_SIGTSTP, HX_SIGCONT];
-    let mut enable_sigtstp = true;
-    unsafe {
-        // If SIGTSTP is SIG_IGN, then do not listen for it
-        if libc::signal(HX_SIGTSTP, HX_SIG_IGN) != HX_SIG_ERR {
-            log::debug!("Disabling SIGTSTP, C-z will not suspend Helix");
-            app_signals.remove(0);
-            enable_sigtstp = false;
-        }
-    }
+    // If SIGTSTP is SIG_IGN, then do not listen for it
+    let enable_sigtstp = if unsafe { libc::signal(HX_SIGTSTP, HX_SIG_IGN) } != HX_SIG_ERR {
+        log::debug!("Disabling SIGTSTP, C-z will not suspend Helix");
+        app_signals.remove(0);
+        false
+    } else {
+        true
+    };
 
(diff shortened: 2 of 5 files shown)


helix-term/src/application.rs
```diff
@@ -43,6 +43,8 @@ type Signals = futures_util::stream::Empty<()>;
 
 const LSP_DEADLINE: Duration = Duration::from_millis(16);
 
+pub static mut ENABLE_SIGTSTP: bool = true;
```
the-mikedavis, line 46: you can use a Lazy once_cell to make mutating this safe

AlexanderBrevig: Fixed. Thanks! I learned something 😄

AlexanderBrevig:
Ready for your wise eyes again @the-mikedavis (tests and lints pass now)

helix-term/src/commands.rs
```diff
@@ -4647,8 +4647,13 @@ fn shell_prompt(cx: &mut Context, prompt: Cow<'static, str>, behavior: ShellBeha
...
 fn suspend(_cx: &mut Context) {
-    #[cfg(not(windows))]
-    signal_hook::low_level::raise(signal_hook::consts::signal::SIGTSTP).unwrap();
+    if *crate::application::ENABLE_SIGTSTP
+        .get()
+        .expect("ENABLE_SIGTSTP flag must be set")
+    {
+        #[cfg(not(windows))]
+        signal_hook::low_level::raise(signal_hook::consts::signal::SIGTSTP).unwrap();
+    }
```
the-mikedavis, line 4656: A few things here:

Since the inside of the if statement is not run on windows, no need to run this whole expression on windows. There isn't so much benefit in this case of using `#[cfg(not(windows))]` over `cfg!(not(windows))`, it's better to be idiomatic and have it in the if condition IMO.

The `expect` will panic on `C-z` if `ENABLE_SIGSTP` isn't set. There's no need to fail so loudly - we can default to doing nothing. In fact, we can also remove the

```rs
#[cfg(not(windows))]
ENABLE_SIGSTP.set(false) // note: this also needs an .unwrap() or .ok()
```

in `application.rs` (comments below this one).

Overall, this can just be:

```suggestion
    if cfg!(not(windows)) && *crate::application::ENABLE_SIGTSTP.get() == Some(true) {
        signal_hook::low_level::raise(signal_hook::consts::signal::SIGTSTP).unwrap();
    }
```

helix-term/src/application.rs
```diff
@@ -205,9 +208,12 @@ impl Application {
...
         #[cfg(windows)]
         let signals = futures_util::stream::empty();
+
         #[cfg(not(windows))]
-        let signals =
-            Signals::new(&[signal::SIGTSTP, signal::SIGCONT]).context("build signal handler")?;
+        let signals = get_signals();
+
+        #[cfg(windows)]
+        ENABLE_SIGTSTP.set(false);
```
the-mikedavis, line 216: This can be removed. (See above comment.)

helix-term/src/application.rs
```diff
@@ -845,3 +851,26 @@ impl Application {
...
+            log::debug!("Disabling SIGTSTP, C-z will not suspend Helix");
+            app_signals.remove(0);
+            enable_sigtstp = false;
+        }
+    }
+
+    if let Err(e) = ENABLE_SIGTSTP.set(enable_sigtstp) {
+        eprintln!("ENABLE_SIGTSTP error: {}", e);
+    }
+    Signals::new(app_signals).expect("build signal handler")
```
the-mikedavis, line 875: Like `Application::new`, this function should return a `Result` (a `Result<Signals>` I think) and then this line becomes

```suggestion
    Signals::new(app_signals).context("build signal handler")
```

To prevent panicing on building `Signals` (master avoids panicing using this same strategy).

helix-term/src/application.rs
```diff
@@ -845,3 +851,26 @@ impl Application {
...
+        if libc::signal(HX_SIGTSTP, HX_SIG_IGN) != HX_SIG_ERR {
+            log::debug!("Disabling SIGTSTP, C-z will not suspend Helix");
+            app_signals.remove(0);
+            enable_sigtstp = false;
+        }
+    }
+
+    if let Err(e) = ENABLE_SIGTSTP.set(enable_sigtstp) {
+        eprintln!("ENABLE_SIGTSTP error: {}", e);
+    }
```
the-mikedavis, line 874: In this function we don't really care if `ENABLE_SIGTSTP` has been set before (see up two comments), we just want to set it to the new value. So we can discard the `Result`.

```suggestion
    ENABLE_SIGTSTP.set(enable_sigtstp).ok();
```

To put everything above together:

```rs
#[cfg(not(windows))]
fn get_signals() -> Result<Signals, Error> {
    const HX_SIGTSTP: libc::c_int = 20;
    const HX_SIGCONT: libc::c_int = 18;
    const HX_SIG_IGN: libc::sighandler_t = 1;
    const HX_SIG_ERR: libc::sighandler_t = 0;

    let app_signals = if unsafe { libc::signal(HX_SIGTSTP, HX_SIG_IGN) } != HX_SIG_ERR {
        log::debug!("Disabling SIGTSTP, C-z will not suspend Helix");
        vec![HX_SIGCONT]
    } else {
        ENABLE_SIGTSTP.set(true);
        vec![HX_SIGTSTP, HX_SIGCONT]
    };

    Signals::new(app_signals).context("build signal handler")
}
```

And the line in `Application::new` becomes

```rs
        #[cfg(not(windows))]
        let signals = get_signals()?;
```

the-mikedavis (changes requested):

the-mikedavis:
Looking at this now, I think an AtomicBool would be simpler than a OnceCell. You can use `Ordering::Relaxed` everywhere since we don't care about race conditions of changes between threads.

the-mikedavis:
Oh actually, since `Editor` is created before the `Application`, you can figure out whether sigtstp is enabled higher up in `Application::new` before `Editor::new` is called and pass the bool to `Editor::new`. No need to use an AtomicBool with this approach, you can just put a new `pub suspend_enabled: bool` on the `helix_view::editor::Editor` type.

AlexanderBrevig:
Thanks for all the amazing feedback @the-mikedavis!
I think all of them are in now.

Looking forward to hopefully get this merged so I can get back to `master` :)

AceofSpades5757:
@AlexanderBrevig It looks like there may need to be some conditional compilation. I don't think Windows has typical signals.

the-mikedavis (commented):
Looks like there's a bit more to do with conditional compiles for windows' sake but this is looking good

helix-term/src/commands.rs
```diff
@@ -4647,8 +4647,10 @@ fn shell_prompt(cx: &mut Context, prompt: Cow<'static, str>, behavior: ShellBeha
 }
 
 fn suspend(_cx: &mut Context) {
-    #[cfg(not(windows))]
-    signal_hook::low_level::raise(signal_hook::consts::signal::SIGTSTP).unwrap();
+    if _cx.editor.suspend_enabled {
```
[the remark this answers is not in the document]
AlexanderBrevig: Thanks!

helix-term/src/application.rs
```diff
@@ -845,3 +852,17 @@ impl Application {
...
+    unsafe { libc::signal(libc::SIGTSTP, libc::SIG_IGN) == 0 }
+}
+
+#[cfg(not(windows))]
+fn get_available_signals(enable_suspend: bool) -> Result<Signals, std::io::Error> {
+    let mut app_signals = vec![libc::SIGCONT];
+    if enable_suspend {
+        app_signals.push(libc::SIGTSTP);
+    }
+    Signals::new(app_signals)
```
the-mikedavis, line 867: just to make it a little closer to the original implementation

```suggestion
fn get_available_signals(enable_suspend: bool) -> Result<Signals, Error> {
    let mut app_signals = vec![libc::SIGCONT];
    if enable_suspend {
        app_signals.push(libc::SIGTSTP);
    }
    Signals::new(app_signals).context("build signal handler")
```

helix-term/src/commands.rs
```diff
@@ -4646,9 +4646,11 @@ fn shell_prompt(cx: &mut Context, prompt: Cow<'static, str>, behavior: ShellBeha
...
 }
 
-fn suspend(_cx: &mut Context) {
-    #[cfg(not(windows))]
-    signal_hook::low_level::raise(signal_hook::consts::signal::SIGTSTP).unwrap();
+fn suspend(cx: &mut Context) {
+    if cx.editor.suspend_enabled {
+        #[cfg(not(windows))]
+        signal_hook::low_level::raise(libc::SIGTSTP).unwrap();
+    }
```
the-mikedavis, line 4653:
```suggestion
    if cfg!(not(windows)) && cx.editor.suspend_enabled {
        #[cfg(not(windows))]
        signal_hook::low_level::raise(libc::SIGTSTP).unwrap();
    } else {
        cx.editor.set_status("Suspend is not supported by this terminal").
    }
```

So people know what's up if C-z isn't working

the-mikedavis:
Alright, one last nit from me 😅

In some places we have `enable_suspend` and in other `suspend_enabled`. Since there isn't anything to do in order to enable suspending I think we should just use `suspend_enabled`

AlexanderBrevig:
> Alright, one last nit from me sweat_smile
> 
> In some places we have `enable_suspend` and in other `suspend_enabled`. Since there isn't anything to do in order to enable suspending I think we should just use `suspend_enabled`

I love your feedback and nits :) 
Absolutely prefer that name myself, I think it's a remnant from the private AtomicBool with get/set-like public API.

It's all much cleaner now, thanks to you 👍🏻

the-mikedavis (approved):

helix-term/Cargo.toml
```diff
@@ -43,6 +43,7 @@ signal-hook = "0.3"
 tokio-stream = "0.1"
 futures-util = { version = "0.3", features = ["std", "async-await"], default-features = false }
 arc-swap = { version = "1.5.1" }
+libc = "0.2"
```
archseer, line 46: Move this to the `[target.'cfg(not(windows))'.dependencies]` block below

archseer (commented):
Isn't this something better fixed in the `signal-hook` library? I do see they have a check for `SIG_IGN`: https://github.com/vorner/signal-hook/blob/1236c57dc8a319a410d0c3a68dcd1fc0aa8488a2/signal-hook-registry/src/lib.rs#L236

but I guess tokio crate's implementation sidesteps that: https://github.com/vorner/signal-hook/blob/32721961967fc739118007022b17461dd6507d77/signal-hook-tokio/src/lib.rs#L134-L142

PR #3323, opened 2022-08-04
Add language: jinja2

AceofSpades5757:
Not sure how this would work as it can be embedded in any language, or text file.

- [ ] Set precise settings (e.g. there probably shouldn't be a file type since it should just be a text file, html file, etc.)
- [ ] highlights.scm
- [ ] indents.scm
- [ ] Other scm

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -48,6 +48,7 @@
 | iex | ✓ |  |  |  |
 | java | ✓ |  |  | `jdtls` |
 | javascript | ✓ | ✓ | ✓ | `typescript-language-server` |
+| jinja2 |  |  |  |  |
 | jsdoc | ✓ |  |  |  |
 | json | ✓ |  | ✓ | `vscode-json-language-server` |
 | jsx | ✓ | ✓ | ✓ | `typescript-language-server` |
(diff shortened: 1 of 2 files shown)


the-mikedavis:
> Not sure how this would work as it can be embedded in any language

Tree-sitter doesn't have a built-in way to make the injection language configurable like a templating language would need. This could be handled in a follow-up PR and would benefit other templating languages like EEx. It needs some design but we could have some sort of variable in the injections.scm queries that is derived from the file-type, like `.html.j2` files would have the variable set to `html`.

```scm
((some_node) @injection.content
 (#set! injection.combined)
 (#set! injection.language $some_magic_language_variable))
```

unrelentingtech:
This should also apply to other template languages with equivalent syntax such as [nunjucks](https://mozilla.github.io/nunjucks/) (`.njk`)

pascalkuthe:
clsoing this one as stale. Thank you for contributing!

valpackett:
This cannot actually go "stale", these template languages aren't just going away.

pascalkuthe:
this PR was a draft for nearly two years. It wasn't finished, had conflicts and wasn't going to be merged in this form. This is not an issue but a PR after all. If somebody wants to see jinja2 support they should make a new PR.

PR #3327, opened 2022-08-04
feat(cargo): strip release binaries

kskarthik:
ref https://doc.rust-lang.org/cargo/reference/profiles.html#strip

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,7 @@ split-debuginfo = "unpacked"
 
 [profile.release]
 lto = "thin"
+strip = true
 # debug = true
 
 [profile.opt]


the-mikedavis:
Looks like strip was added in cargo 1.59? https://github.com/rust-lang/cargo/blob/7259757dc618479363aca6176ee5e76fd534e68c/CHANGELOG.md#cargo-159-2022-02-24

We target an older rust version to make packaging easier for some distributions, currently 1.57. We could probably switch to this once we move the MSRV up to 1.59+ but we do already strip release binaries for macOS and linux in the CI: https://github.com/helix-editor/helix/blob/0ee20611022b5a7bec727d2159ec7c6b36e956b6/.github/workflows/release.yml#L118-L120

kchibisov:
I don't think that it makes sense to do so. Distro packages do run strip anyway and if you want to run strip for yourself you're free to do so as-well. So this change doesn't provide any benefit, since if you build it yourself in release mode you're interested in providing a backtrace so something can be done about it.

Disabling stripping that way would be annoying for package managers like `portage` to deal with, since it has an option to control stripping behavior and if the application force strip you're out of luck.

PR #3330, opened 2022-08-04
Detect indent-style in `:set-language`

the-mikedavis:
Indent style may change when choosing a language with `:set-language`.
Line-endings most likely will not change, but `:set-language` should
have a similar effect as reloading a file (`:reload`), plus the two
are currently grouped in the implementation and line-ending detection
is not particularly expensive.

That this didn't happen (prior to this change) is more noticeable because of https://github.com/helix-editor/helix/pull/3095


AceofSpades5757:
I'm actually surprised that this doesn't update the settings, like a `:config-reload` or something to that effect.

AceofSpades5757 (approved):
Tested indentation level using text to html. Looks good!

PR #3331, opened 2022-08-04
fix: add elvish completions

AlexanderBrevig:
# Completions for the _awesome_ [Elvish](elv.sh) shell!

It has the expected basics:

image

But then `-g` and `--grammar` leads to only two suggestions

image

`--config` will only show files in it's completions.

Finally, flags that will not open the editor will stop further suggestions.


AlexanderBrevig:
There's also this: https://github.com/rsteube/carapace-bin, which has support for Helix I just learned. Still think it's nice to include it for people who don't want (or can't) install carapece.

PR #3334, opened 2022-08-04
Fix comment

AceofSpades5757:
literally just fixing a comment that become incorrect due to a change made for pr #3095

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -576,7 +576,7 @@ impl Document {
     }
 
     /// Detect the indentation used in the file, or otherwise defaults to the language indentation
-    /// configured in `languages.toml`, with a fallback to 4 space indentation if it isn't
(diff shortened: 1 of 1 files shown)


helix-view/src/document.rs
```diff
@@ -576,7 +576,7 @@ impl Document {
     }
 
     /// Detect the indentation used in the file, or otherwise defaults to the language indentation
-    /// configured in `languages.toml`, with a fallback to 4 space indentation if it isn't
+    /// configured in `languages.toml`, with a fallback to the default indentation if it isn't
```
the-mikedavis, line 579: This should either say what the default value is ("tabs") or link to `DEFAULT_INDENT`

AceofSpades5757: Hmm. Had issues linking the variable, so went with "tabs". It was complaining that it was private, but then still complained when I ran with the respective flag.

the-mikedavis (approved):

PR #3338, opened 2022-08-05
Derive Document language name from languages.toml name key

the-mikedavis:
This changes switches from deriving the language name from the
`languages.toml` `scope` key to `name` (`language_id` in the
`LanguageConfiguration` type). For the most part it works to derive
the language name from scope by chopping off `source.` or `rsplit_once`
on `.` but for some languages we have now like html (`text.html.basic`),
it doesn't. This should be a more accurate fallback for the `language_id`
method which is used in LSP and currently uses the `rsplit_once`
strategy.

Here we expose the language's name as `language_name` on `Document` and
replace ad-hoc calculations of the language name with the new method.

This is most impactful for the `file-type` statusline element which is
currently using `language_id`.

It should also improve highlighting in popups for languages like html, erb, ejs, maybe solidity. For most languages there should be no functional changes from this PR.

Supersedes https://github.com/helix-editor/helix/pull/2998
Should make https://github.com/helix-editor/helix/pull/3273 more accurate

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -863,10 +863,7 @@ pub fn signature_help_impl(cx: &mut Context, invoked: SignatureHelpInvoked) {
                 }
             };
             let doc = doc!(editor);
-            let language = doc
-                .language()
-                .and_then(|scope| scope.strip_prefix("source."))
-                .unwrap_or("");
+            let language = doc.language_name().unwrap_or("");
 
             let signature = match response
                 .signatures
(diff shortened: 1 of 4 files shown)


getreu:
With this commit my `languages.toml` configuration does not work any more: 

```toml
[[language]]
name = "markdown"
language-server = { command = "/opt/ltex-ls/bin/ltex-ls" }
file-types = ["md"]
scope = "source.markdown"
roots = []
```

How can I fix it?

PR #3342, opened 2022-08-06
Attempt to bring spacebones light up to speed with recent features

EpocSquadron:
* Adds cursorline (primary and secondary)
* Reworks colors slightly to try and match upstream theme better
* Distinguishes primary from regular selections
* Implements some of the markup specific styles
* Implements mode colors support

The upstream doesn't have much in the way of guides, so I did my best. Not expecting everyone to agree on the color choices but those can be further adjusted before or after merge.

### DIFF
--- a/runtime/themes/spacebones_light.toml
+++ b/runtime/themes/spacebones_light.toml
@@ -70,6 +70,7 @@
 "ui.menu" = { fg = "fg1", bg = "bg2" }
 "ui.menu.selected" = { fg = "base", bg = "bg2", modifiers = ["bold"] }
 "ui.virtual" = "base-dim"
+"ui.virtual.ruler" = { bg = "bg1" }
 
 "diagnostic" = { modifiers = ["underlined"] }
 
--- a/runtime/themes/spacebones_light.toml
+++ b/runtime/themes/spacebones_light.toml
@@ -24,57 +24,77 @@
 "comment" = { fg = "#a49da5", modifiers = ["italic"]  }
 "constant" = { fg = "#6c3163" }
 "constant.builtin" = { fg = "#6c3163", modifiers = ["bold"] }
-"string" = "#2d9574"
+"string" = "theme_aqua"
 "number" = "#6c3163"
 "escape" = { fg = "fg2", modifiers = ["bold"] }
 "label" = "#b1951d"
 "module" = "#b1951d"
 
 # TODO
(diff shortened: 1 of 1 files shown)


runtime/themes/spacebones_light.toml
```diff
@@ -24,57 +24,77 @@
...
+"ui.selection" = { bg = "hl2" }
+"ui.selection.primary" = { bg = "hl1" }
 "ui.cursor.primary" = { modifiers = ["reversed"] }
 "ui.cursor.match" = { bg = "bg3" }
+"ui.cursorline.primary" = { bg = "bg1" }
 "ui.menu" = { fg = "fg1", bg = "bg2" }
-"ui.menu.selected" = { fg = "#655370", bg = "#d1dcdf", modifiers = ["bold"] }
-"ui.virtual.whitespace" = "bg2"
+"ui.menu.selected" = { fg = "base", bg = "bg2", modifiers = ["bold"] }
+"ui.virtual" = "base-dim"
```
the-mikedavis, line 72: The ruler is a little faint. Should it be something like

```toml
"ui.virtual.ruler" = { bg = "base-dim" }
```

so you can see it without having text cross it?

the-mikedavis:
There's also `markup.heading.1`, `markup.heading.2` etc if you'd like to add those. It can be a bit loud to have so many colors for headings though so up to you

runtime/themes/spacebones_light.toml
EpocSquadron: I don't use the rulers feature, but trying it out I found `base-dim` to be too dark for my liking. I set it to `bg1` which is the same color as the cursorline, which makes it almost look like a crosshairs type thing. Not sure if that's acceptable, I can change it agian.

EpocSquadron:
> There's also `markup.heading.1`, `markup.heading.2` etc if you'd like to add those. It can be a bit loud to have so many colors for headings though so up to you

Yeah I didn't feel the need to distinguish those colors and lacked enough guidance on what the colors should be -- the upstream theme has a fairly limited palette. I could look at the full palette of the spacemacs theme as that seems to be ultimately where this came from, but I'm not sure I want to put in that time when I can work towards some tree-sitter stuff I've been sitting on a long time.

the-mikedavis (approved):
Looks good, thanks for the updates! 🎨

Could you update the screenshot in the themes wiki?

PR #3343, opened 2022-08-06
Adjust `m` textobject description and minor code clarification

EpocSquadron:
Minor cleanups for easier understanding of functionality and code.

### DIFF
--- a/helix-core/src/textobject.rs
+++ b/helix-core/src/textobject.rs
@@ -198,26 +198,26 @@ pub fn textobject_paragraph(
     Range::new(anchor, head)
 }
 
-pub fn textobject_surround(
+pub fn pair_surround(
     slice: RopeSlice,
     range: Range,
     textobject: TextObject,
     ch: char,
     count: usize,
 ) -> Range {
-    textobject_surround_impl(slice, range, textobject, Some(ch), count)
+    pair_surround_impl(slice, range, textobject, Some(ch), count)
 }
 
-pub fn textobject_surround_closest(
+pub fn pair_surround_closest(
     slice: RopeSlice,
     range: Range,
     textobject: TextObject,
     count: usize,
 ) -> Range {
-    textobject_surround_impl(slice, range, textobject, None, count)
+    pair_surround_impl(slice, range, textobject, None, count)
 }
 
-fn textobject_surround_impl(
+fn pair_surround_impl(
     slice: RopeSlice,
     range: Range,
     textobject: TextObject,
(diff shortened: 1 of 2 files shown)


helix-core/src/textobject.rs
```diff
@@ -198,26 +198,26 @@ pub fn textobject_paragraph(
...
     textobject: TextObject,
     ch: char,
     count: usize,
 ) -> Range {
-    textobject_surround_impl(slice, range, textobject, Some(ch), count)
+    pair_surround_impl(slice, range, textobject, Some(ch), count)
 }
 
-pub fn textobject_surround_closest(
+pub fn pair_surround_closest(
```
archseer, line 211: Shouldn't this be textobject_pair_ to be consistent with the rest? Also is it worth renaming this since it'll break user mappings

helix-term/src/commands.rs
```diff
@@ -4361,7 +4361,7 @@ fn select_textobject(cx: &mut Context, objtype: textobject::TextObject) {
             ("a", "Argument/parameter (tree-sitter)"),
             ("o", "Comment (tree-sitter)"),
             ("t", "Test (tree-sitter)"),
-            ("m", "Matching delimiter under cursor"),
+            ("m", "Closest pair to cursor"),
```
the-mikedavis, line 4364:
```suggestion
            ("m", "Closest surrounding pair to cursor"),
```

helix-core/src/textobject.rs
the-mikedavis: I agree about the `textobject_pair_`

This shouldn't affect user mappings though, currently you can't bind `mim`/`mam` without macro bindings because of the `cx.on_next_key` in the `select_textobject` function from commands

helix-term/src/commands.rs
```diff
@@ -4361,7 +4361,7 @@ fn select_textobject(cx: &mut Context, objtype: textobject::TextObject) {
...
             (" ", "... or any character acting as a pair"),
```
the-mikedavis, line 4365: Oddly the "..." kinda confuses me. Along with the space on the left-hand-side it makes me think I can press any character and have that be matched. I think we could just remove the "..."?

EpocSquadron: You can press any character and it will be matched! Even `b`. This is helpful when you have a regex in PHP that you've delimited with `#`, which is not a usual pair character, for instance.

helix-core/src/textobject.rs
EpocSquadron: Prefix added. I can see the consistency argument. I just thought the original `textobject_surround` was misleading when reading the code, as my mind went to it being generic over any textobject.

helix-term/src/commands.rs
the-mikedavis: Ah, sweet! Let's keep this text then 👍

the-mikedavis (approved):

EpocSquadron:
My understanding is that the feedback has been addressed, is there anything else needed here?

PR #3345, opened 2022-08-06
Rust queries: add `(function_item)` as a local scope

ChrHorn:
Function parameter highlighting was leaking into other scopes.

before:
image

after:
image


the-mikedavis (approved):
Good catch, thanks!

PR #3347, opened 2022-08-06
feat: Find diagnostics for both `C:` and `c:` forms of a path

### LINKED ISSUE #3267: Looking up diagnostics by URL is ambigous
### Summary

Errors do not show up in the local diagnostic picker (Shift g) while writing rust code with the rust analyzer. But they do show up in the global diagnostic picker (Shift G). Screenshots below as an example.

Note that issues are shown in the top right of the screen if my cursor is on the error. As well as indicating the error presence on the left diagnostic gutter.

Code with obvious issues
image

Empty local diagnostic picker
image

Global picker has all the errors and warnings.
image

I found this issue only with rust and rust analyzer. Other languages work well, with file issues in local and workspace issues in the global diagnostic picker.

Helix :
- built from source ( dfc31e74 )
- cargo version  : cargo 1.62.0 (a748cf5a3 2022-06-08)
- rustc version : rustc 1.62.0 (a8314ef7d 2022-06-27)

Rust Analyzer: 
- built from source ( 897a7ec4b )

I have attached helix logs for today and the last time I used helix while omitting the rest as those would be too much. If you need more comprehensive date ranged logs, I'd be happy to provide them.

I don't have much experience with rust so I am not sure if this is a rust issue, a rust analyzer issue, or a helix issue.

### Reproduction Steps

 

I tried this:

1. `hx`
2. Open a rust file.
3. Add some code with errors (any kind which rust analyzer will notify)
4.  Use local diagnostic picker (Shift g)

I expected this to happen:

The issues in the currently open file to show up there

Instead, this happened:

no errors listed in the local diagnostic picker.

Opening global diagnostic picker (Shift G) shows the errors as expected.

### Helix log

```
2022-07-24T00:50:06.819 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:53:07.436 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:53:33.455 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:54:19.304 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T00:54:19.305 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T00:54:27.368 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:57:34.012 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:57:52.634 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T00:58:32.992 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T00:59:50.604 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T00:59:50.605 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T01:01:26.954 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-07-24T01:05:15.300 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-24T01:05:15.302 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-07-31T01:16:44.031 helix_lsp::transport [ERROR] <- ServerError(-32801): waiting for cargo metadata or cargo check
2022-07-31T01:16:45.551 helix_lsp::transport [ERROR] <- ServerError(-32801): waiting for cargo metadata or cargo check
2022-07-31T01:16:46.149 helix_lsp::transport [ERROR] <- ServerError(-32801): waiting for cargo metadata or cargo check
2022-07-31T01:16:47.801 helix_lsp::transport [ERROR] err <- "[ERROR project_model::workspace] cyclic deps: helix_view(CrateId(66)) -> helix_tui(CrateId(60)), alternative path: helix_tui(CrateId(60)) -> helix_view(CrateId(66))\n"
2022-07-31T01:17:10.680 helix_lsp::transport [ERROR] err <- "[ERROR project_model::workspace] cyclic deps: helix_view(CrateId(66)) -> helix_tui(CrateId(60)), alternative path: helix_tui(CrateId(60)) -> helix_view(CrateId(66))\n"
2022-07-31T01:18:54.795 helix_term::application [WARN] unhandled window/showMessage: ShowMessageParams { typ: Error, message: "overly long loop turn: 137.2607ms" }
```

</details>

### Platform

Windows

### Terminal Emulator

windows terminal 1.14.1963.0

### Helix Version

22.05-272-gdfc31e74

poliorcetics:
`Document::url()` uses `Document::path()` which is just a reference to `Document.path`. Since LSPs
can send either a path with an uppercase or lowercase drive letter, we need to check for the other
option when the first fails.

We do this throughout Helix by introducing `NormalizedUrl` which takes care of the conversion and
delegates to its base `Url` when needed. If future conversions are needed, they can be added here
instead of having to repeat them all over the place.

Closes #3267

### DIFF
--- a/helix-core/src/normalized_url.rs
+++ b/helix-core/src/normalized_url.rs
@@ -47,8 +47,11 @@ impl NormalizedUrl {
         Some(Self::new(base))
     }
 
-    /// Access the `Url` stored as the base of the normalized one
-    pub fn base(&self) -> &Url {
+    /// Access the `Url` stored as the base of the normalized one.
+    ///
+    /// If you have the possibility to consume `self` because it's not used later, consider using
+    /// [`Self::into_base()`], especially if you are cloning after the current `.base()` call.
+    pub fn as_base(&self) -> &Url {
         &self.base
     }
 
@@ -66,6 +69,13 @@ impl NormalizedUrl {
     /// See [`Url::path_segments()`]
     pub fn path_segments(&self) -> Option<std::str::Split<'_, char>> {
         self.base.path_segments()
+    }
+
+    /// Consumes the normalized URL to return the base url given on construction.
+    ///
+    /// If only a reference is needed, use [`Self::as_base()`].
+    pub fn into_base(self) -> Url {
+        self.base
     }
 }
 
--- a/helix-core/src/normalized_url.rs
+++ b/helix-core/src/normalized_url.rs
@@ -1,0 +1,175 @@
+//! [`NormalizedUrl`]s are used to make comparisons more logical for URLs, notably those holding
+//! Windows paths.
+use std::borrow::Cow;
+use std::cmp::Ordering;
+use std::fmt;
+#[cfg(windows)]
+use std::path::{Component, Prefix};
+use std::path::{Path, PathBuf};
+
+use url::Url;
+
+/// Normalize a [`Url`] to make it comparable across idiosyncracies from different tools.
+///
+/// # Example case
+///
+/// Helix gets paths from other sources, like LSPs. In such a case, Helix can't know in advance for
+/// every source what the case will be for the prefix component on Windows (`C:` or `c:`). Since
+/// this can hinder comparisons, especially in [`Url`]s, this structure gives access to the opposite
+/// casing of the `Url` on Windows and uses it to compare.
+#[derive(Debug, Clone)]
+pub struct NormalizedUrl {
+    base: Url,
+    /// This field is only available on Windows since differences in casing on other OSes are not
+    /// limited to the prefix and should be dealt with differently.
+    #[cfg(windows)]
+    opposite_case: Option<Url>,
+}
+
+impl NormalizedUrl {
+    pub fn from_file_path(path: &Path) -> Option<Self> {
+        let path: Cow<'_, Path> = if path.is_absolute() {
+            Cow::Borrowed(path)
+        } else {
+            Cow::Owned(crate::path::get_canonicalized_path(path).ok()?)
+        };
+        // If we managed to canonicalize, this should never fail else its a bug
+        let base = Url::from_file_path(&path).unwrap();
+        #[cfg(windows)]
+        let opposite_case =
+            path_prefix_opposite_casing(&path).map(|x| Url::from_file_path(&x).unwrap());
+
+        Some(Self {
+            base,
+            #[cfg(windows)]
+            opposite_case,
+        })
+    }
+
+    /// Access the `Url` stored as the base of the normalized one
+    pub fn base(&self) -> &Url {
+        &self.base
+    }
+
+    /// See [`Url::path()`]
+    pub fn path(&self) -> &str {
+        self.base.path()
+    }
+
+    /// See [`Url::to_file_path()`]
+    #[allow(clippy::result_unit_err)]
+    pub fn to_file_path(&self) -> Result<PathBuf, ()> {
+        self.base.to_file_path()
+    }
+
+    /// See [`Url::path_segments()`]
+    pub fn path_segments(&self) -> Option<std::str::Split<'_, char>> {
+        self.base.path_segments()
+    }
+}
+
+impl PartialEq<Self> for NormalizedUrl {
+    fn eq(&self, other: &Self) -> bool {
+        // Since the other fields are derived from this one deterministically, we can simply
+        // compare `.base` to know if two `NormalizedUrl`s are equal.
+        self.base.eq(&other.base)
+    }
+}
+
+impl Eq for NormalizedUrl {}
+
+impl PartialOrd<Self> for NormalizedUrl {
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
+        self.base.partial_cmp(&other.base)
+    }
+}
+
+impl Ord for NormalizedUrl {
+    fn cmp(&self, other: &Self) -> Ordering {
+        self.base.cmp(&other.base)
+    }
+}
+
+impl fmt::Display for NormalizedUrl {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        self.base.fmt(f)
+    }
+}
+
+// Interactions with the original Url type are below
+
+impl From<Url> for NormalizedUrl {
+    fn from(u: Url) -> Self {
+        let path = Path::new(u.path());
+        Self::from_file_path(path).expect("URL failed to convert to absolute path")
+    }
+}
+
+impl PartialEq<Url> for NormalizedUrl {
+    #[cfg(windows)]
+    fn eq(&self, other: &Url) -> bool {
+        self.base.eq(other) || self.opposite_case.as_ref().map_or(false, |oc| oc.eq(other))
+    }
+
+    #[cfg(not(windows))]
+    fn eq(&self, other: &Url) -> bool {
+        self.base.eq(other)
+    }
+}
+
+impl PartialOrd<Url> for NormalizedUrl {
+    fn partial_cmp(&self, other: &Url) -> Option<Ordering> {
+        self.base.partial_cmp(other)
+    }
+}
+
+/// Utility function to get the opposite casing of a Windows path, if any
+#[cfg(windows)]
+fn path_prefix_opposite_casing(default_case_path: &Path) -> Option<PathBuf> {
+    let mut components = default_case_path.components();
+    let first_comp = components.next()?;
+
+    match first_comp {
+        Component::Prefix(prefix)
+        // Anything else than `Disk` variants is not subject to this casing issue
+        if matches!(prefix.kind(), Prefix::VerbatimDisk(_) | Prefix::Disk(_)) => {
+            let cased_prefix = {
+                // Hard to get data on this, but most drive letter should be uppercased by
+                // default so lowercasing first means not having to then uppercase ?
+                let mut prefix_str = prefix.as_os_str().to_ascii_lowercase();
+                if prefix_str == prefix.as_os_str() {
+                    prefix_str.make_ascii_uppercase();
+                }
+                prefix_str
+            };
+
+            // The `push`s should be free thanks to this
+            let mut updated_case_path = PathBuf::with_capacity(default_case_path.as_os_str().len());
+            updated_case_path.push(&cased_prefix);
+            for cp in components {
+                updated_case_path.push(cp.as_os_str());
+            }
+            Some(updated_case_path)
+        },
+        _ => None,
+    }
+}
+
+#[cfg(test)]
+#[cfg(windows)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn different_drives_case_are_equal() {
+        let p1 = Path::new("C:\\one\\path\\Here");
+        let p2 = Path::new("c:\\one\\path\\Here");
+
+        assert_eq!(p1, p2);
+
+        let norm = NormalizedUrl::new(p1).unwrap();
+        let u = Url::from_file_path(p2).unwrap();
+
+        assert_eq!(norm, u);
+    }
+}
--- a/helix-core/src/normalized_url.rs
+++ b/helix-core/src/normalized_url.rs
@@ -3,8 +3,6 @@
 use std::borrow::Cow;
 use std::cmp::Ordering;
 use std::fmt;
-#[cfg(windows)]
-use std::path::{Component, Prefix};
 use std::path::{Path, PathBuf};
 
 use url::Url;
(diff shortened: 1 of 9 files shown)


poliorcetics:
I checked all uses of `Document::url()`, the main other case is for symbols, and I don't think it will be affected then ? Or at least, it won't lose symbols, only show a path when it could avoid it (because it's for the current file), but I'm not certain.

Frojdholm:
I think this is a more general problem than just this instance. Anytime URLs are compared you risk issues, for example in `diag_picker` the `current_path` is compared again with the URLs from the LSP server.

My idea would be to create a newtype `NormalizedUrl` (naming to be determined) that wraps the URL and does the normalisation. This type will uphold the invariant that the drive letter is upper case, but will otherwise be a pretty thin wrapper around `Url`. A bonus is that if any new normalisations are discovered they can done here as well.

We would then replace all usages of `Url` in helix with `NormalizedUrl`. Code using for example `Eq` will then just work basically unmodified.

Frojdholm:
@poliorcetics This is a draft of what I've been working on. It's quite clunky since URLs are often used to interface with other types in the `lsp_types` crate. The normalizing function is also a bit of a hack at the moment, but it works on my machine. Either way we will need a newtype if we want to store the URL in a hash map since the hash and eq methods are used for finding the keys.

https://github.com/helix-editor/helix/compare/23f4a40a7a037948935aea7f24e20319029d174e...Frojdholm:helix:draft-3267-normalized-url

poliorcetics:
@Frojdholm I had already started on your suggestion on my side, I looked at yours in diagonal, it seems like we got similar solutions, except for the `C:`/`c:` check.

helix-core/src/normalized_url.rs
```diff
@@ -0,0 +1,175 @@
...
+
+#[cfg(test)]
+#[cfg(windows)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn different_drives_case_are_equal() {
+        let p1 = Path::new("C:\\one\\path\\Here");
+        let p2 = Path::new("c:\\one\\path\\Here");
```
Frojdholm, line 166: Nice with the tests 🚀

Some more test cases that I can think of off the top of my head:
- Testing `p2` against `p1`
- Testing unix paths
- Testing non-file scheme URLs

helix-core/src/normalized_url.rs
```diff
@@ -0,0 +1,175 @@
...
+    }
+}
+
+// Interactions with the original Url type are below
+
+impl From<Url> for NormalizedUrl {
+    fn from(u: Url) -> Self {
+        let path = Path::new(u.path());
+        Self::from_file_path(path).expect("URL failed to convert to absolute path")
+    }
```
Frojdholm, line 105: This will panic on non-file scheme URLs, is that intended?

Frojdholm (commented):
Great work, I haven't had the opportunity to try it out yet, but it seems fairly similar to what I've implemented :+1:

I have two design related comments:
- I don't think this problem is restricted to Windows. Imagine for example accessing files on a Windows server through a Linux host.
- I think having the extra field in `NormalizedUrl` complicates the design somewhat. I guess you did it for performance to avoid having to always calculate the new path? Have you measured the performance differences between the two approaches? I don't know how slow/performance critical the URL handling is, but if it's not critical I would personally prefer a simpler implementation.

poliorcetics:
I have not yet pushed my changes, I was benchmarking both approach (normalizing at init and storing the result vs normalizing during comparison).

The worst case in both approach is a Windows path URL of the form `file://C:/my/neat/path` or `file:///<same here>`, because this is the form that has to be normalized for later comparison. All other forms of comparison, notably those with different schemes for the URLs, are **very** fast, less than a hundred (100) nanoseconds at worst, in picoseconds for the best case on my machine (when the first letter of the scheme is different, e.g, `https` and `file`).

So, worst case: `file://C:/my/neat/path`. This is still *very* fast, despite needing to `.clone()` for the benchmarks because `Url`s are not `Copy`.

- When **normalizing during construction**, cloning+normalizing (which needs to generate a new URL with `c:`, different from the base one), takes **~315ns**.
- When **normalizing during comparison**, the same cloning and normalizing takes **~350ns**, and that's because my test urls are small, if I add a few levels of depth to my test paths, it gets worse (still in the hundreds of nanoseconds, but worse all the same).

I expect performances to be worse on Windows because the allocator is (reputedly) not as good and most machines are not a Macbook Pro with the biggest M1 Pro, but even at twice the time, this is still good.

I'll continue on the road of normalizing at construction because overall all the comparisons are faster with it and the code is clearer IMO (e.g the comparison function for the normalize-during-compare case needs a little help to be faster, and it still does not reach the pre-normalized case by ~10%). On a macOS install, with local paths (so no drive letter `C:`), constructing, even with normalizing takes about 40ns and so shouldn't affect helix's performance at all, especially compared to reading the file itself.

Of course my benchmarks are probably not that representative, but they at least show both approach are sufficiently fast and the first one will do much less small and repeated allocations.

helix-core/src/normalized_url.rs
poliorcetics: Done

helix-core/src/normalized_url.rs
poliorcetics: Fixed

helix-core/src/normalized_url.rs
```diff
@@ -0,0 +1,320 @@
...
+/// every source what the case will be for the prefix component on Windows (`C:` or `c:`). Since
+/// this can hinder comparisons, especially in [`Url`]s, this structure gives access to the opposite
+/// casing of the `Url` and uses it to compare.
+#[derive(Debug, Clone)]
+pub struct NormalizedUrl {
+    base: Url,
+    /// When appropriate, this contains the base URL with the drive letter in the opposite case.
+    ///
+    /// `file://C:/my/windows/path` in `base` means this is `Some(file://c:/my/windows/path)`.
+    windows_normalized: Option<Url>,
```
Frojdholm, line 24: I still think that this should just be
```rust
pub struct NormalizedUrl {
    base: Url
}
```
Then all the important comparison traits should be deriveable. Downside is that you will always have to convert both URLs to `NormalizedUrl`s when comparing.

I can see pros and cons for both. Doing what is done here is a bit easier at the comparison site, with the tradeoff of the struct being twice the size and complicating the implementation slightly. The other way is less ergonomic, but the implementation is simpler and the size of the struct will be the same as `Url`.

Frojdholm (approved):
Very thorough benchmarking :clap:

helix-core/src/normalized_url.rs
poliorcetics: I don't think size is much of a concern right now, here are the point of use of `NormalizedUrl`:

- `helix_lsp::client::Client`, `root_uri: Option<NormalizedUrl>`. I removed it because it was only used to be sent immediately to the LSP server, no need to construct a normalization only to fail to use it after that. *bonus, it also allowed me to remove `clone` calls*
- `helix_lsp::client::Client::text_document_did_open`, `uri` parameter: I removed this one since it's directly sent to the server, made the helix code pass the owned URL instead of a reference to the NormalizedUrl. With `NormalizedUrl::into_base()` that allowed me to remove a useless clone too :tada:
- `helix_term::commands::lsp`, `impl ui::menu::Item for lsp::SymbolInformation`: here it's needed since it's comparing a Helix URL to an LSP one, precisely the use case it was written for. This function is called once for each time a menu is opened with a symbol to show, which means the comparison can easily be done hundreds of time. Speeding it up means more UI responsiveness.
- `helix_term::commands::lsp::PickerDiagnostic.url`, used for comparison but also stored, measuring would be needed to know which costs more, see `sym_picker()` and `diag_picker()` below
- `helix_term::commands::lsp::sym_picker()`, comparison for the path of each symbol, here it's very probably more useful to store the normalization than recompute it for each symbol
- `helix_term::commands::lsp::diag_picker()` (twice !) In the `BTreeMap` it will inflate the size but also allow much more efficient comparison when looking for an item. Hard to say what's more costly. The other use is the same as `sym_picker()`
- `helix_view::document::Document::url()`, allocates each time but it's not called that much, mostly on opening a file picker (where comparison is done way more) and communicating using `Document::(versioned_)identifier()`, where allocating is done only to cast back into base immediately so initialization is done more. Launching the language server too, but that's a one time cost so one way or another it won't affect performances.
- `helix_view::editor::Editor.diagnostics`: another `BTreeMap`, where I think storage is done more than comparison, but really that would depend on the use of the diagnostic picker. I use it very often so a single batch of diagnostics will usually end up being compared 5 to 10 times even if there are only a few of them, but other people may have other usage pattern. Hard to tell really, without backing data.

Overall I think the current way is good, but the 'normalize during comparison only' way would be good too. Without backing data, I will say that allocating once and then never again, especially when the cost windows normalization won't happen at all for the people on Unix system, is better for allocators than maybe allocating at each comparison.

At least reviewing the usage points helped me find a few improvements ! Thanks for sticking with me on this PR :)

poliorcetics:
Rebased on recent master, it was conflicting

archseer:
\cc @pascalkuthe 

Sorry for the delay on this PR, I've been hoping to find a way to do this without an additional wrapper type.

pascalkuthe:
The official URI reference implementation for the LSP spec is aware of this problem and even has tests for it.
They seem to simply provide an option to normalize the drive letter when converting a `URI` to a path.

More importantly when an URI that is encoded to a string (which is done when it's sent server -> client/client -> serve) the **driver litter is always converted to lower case**:

 https://github.com/microsoft/vscode-uri/blob/1af8b2576342c7e6deff4d4a2c2e6cdfda170ff2/src/uri.ts#L637

Sadly the rust URL crate exposed by LSP types doesn't seem to perform any kind of normalization like this at any point.
That is probably because this specific property is specific to LSP URIs but `lsp_types` just uses the `URL` crate regardless.
There seem to be mysterious historic reasons for this: https://github.com/microsoft/vscode/issues/42159#issuecomment-360791384

Given that the driver letters sent by any LSP conformant server (or at-least those that work with VSCode?) should always be lowercase it should hopefully be enough for us to be careful to always lowercase the driver letters of any paths before turning them into an `URL`. 

I sincerely hope that all LS actually stick to this send everything lowercase rule (but atleast all LS that are written in nodejs should as they would be using the official LSP uri library). Otherwise we need to somhow intercept messages somewhere early in the LSP client and lowercase the driver letter.

Regardless I am not a fan of intruducing a new normalized path type that stores two versions of the URI. We should just normalize all our URIs to be lowercased to match VSCode and the spec here

poliorcetics:
> I sincerely hope that all LS actually stick to this send everything lowercase rule (but at-least all LS that are written in nodejs should as they would be using the official LSP uri library). Otherwise we need to somehow intercept messages somewhere early in the LSP client and lowercase the driver letter.

Yeah that won't happen. Relying on that when the rest of the computing world uses uppercase letters was a mistake from the LSP spec IMO. We can avoid the problem if we lowercase drive letters everywhere but we absolutely shouldn't rely on LSPs doing the right thing

pascalkuthe:
Actually an even easier solution: Do we really need to store an URI instead of just a `Path` rust `PathBuf` and `Path` correctly handles this case. Is there any reason to store an `URL` over just converting everything to `PathBuf` (we still need to normalize before sending ofcourse)?

poliorcetics:
Does helix intend to support remote editing one day (like `hx scp://user@remote[:port]//path/to/file.txt`) ? If so, using `PathBuf` would be detrimental

pascalkuthe:
> Does helix intend to support remote editing one day (like `hx scp://user@remote[:port]//path/to/file.txt`) ? If so, using `PathBuf` would be detrimental

I think @archseer sentiment on the topic (https://github.com/helix-editor/helix/issues/3721#issuecomment-1237605939) was mostly to use  `sshfs` or similar so I am inclined to say that is a problem we can putpoff until helix actually gains that capability

poliorcetics:
> > Does helix intend to support remote editing one day (like `hx scp://user@remote[:port]//path/to/file.txt`) ? If so, using `PathBuf` would be detrimental
> 
> 
> 
> I think @archseer sentiment on the topic (https://github.com/helix-editor/helix/issues/3721#issuecomment-1237605939) was mostly to use  `sshfs` or similar so I am inclined to say that is a problem we can putoff until helix actually gains that capability

The arguments on the issue are pretty much what I would have said about sshfs, the idea of it is nice, in practice it's not that useful because of its inherent limitations.

Lowercasing drive letters all the time seems more forward-compatible than using `PathBuf` here, I can change `NormalizedUrl` to do that instead (and so not allocate twice)

pascalkuthe:
> > > Does helix intend to support remote editing one day (like `hx scp://user@remote[:port]//path/to/file.txt`) ? If so, using `PathBuf` would be detrimental
> > 
> > 
> > I think @archseer sentiment on the topic ([#3721 (comment)](https://github.com/helix-editor/helix/issues/3721#issuecomment-1237605939)) was mostly to use  `sshfs` or similar so I am inclined to say that is a problem we can putoff until helix actually gains that capability
> 
> The arguments on the issue are pretty much what I would have said about sshfs, the idea of it is nice, in practice it's not that useful because of its inherent limitations.
> 
> Lowercasing drive letters all the time seems more forward-compatible than using `PathBuf` here, I can change `NormalizedUrl` to do that instead (and so not allocate twice)


I am not convinced that remote editing belongs in the core editor (that issue is labeled with `A-plugin`) and even if it does supporting that will require much larger changes to the way we represent buffer paths anyway. I don't think it makes sense to introduce a bunch of code for a new wrapper type (https://github.com/helix-editor/helix/pull/3347#issuecomment-1434155542) for a hypothetical future feature right now.

The future proofing argument doesn't really make sense here as helix fundamentally only deals in `Path`s anyway and it's (in my opinion) a bug rather than a feature that we compared URIs at all.  If you look at how we actually use `URIs` in the codebase there are currently exactly three operations that we perform on URIs:

* send them to the LS server, `URI`s are always generated from `Document::path` (which is a `PathBuf`). This usecase is what `Docuemnt::uri()` was intended for. 
* Convert a `URI` from the server to a `Path` with `to_file_path` for local use.
* Comparison  with another `URI`. However that other `URI` is always generated from `Document::path` and hence a `PathBuf`/`file://` URI

All usecases only work with `file:://` URIs anyway so we might aswell switch all occurrences of the third case to the second case to be consistent and to fix problems like https://github.com/helix-editor/helix/issues/3267 (in general I think a simple URL comparison will fail). I think there will actually be more edgecases like this as file comparisons are tricky (pretty sure

PR #3349, opened 2022-08-07
Fix process spawning error handling

PiergiorgioZagaria:
As mentioned by @dead10ck in https://github.com/helix-editor/helix/pull/2942#discussion_r939606221 if something is printed to stderr it doesn't mean that the process failed, so we need to check the exit code and then print stderr if there is anything. This should also be valid when calling the shell from helix

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -875,8 +875,8 @@ fn goto_window(cx: &mut Context, align: Align) {
     let last_line = view.last_line(doc);
 
     let line = match align {
-        Align::Top => (view.offset.row + scrolloff + count),
-        Align::Center => (view.offset.row + ((last_line - view.offset.row) / 2)),
+        Align::Top => view.offset.row + scrolloff + count,
+        Align::Center => view.offset.row + ((last_line - view.offset.row) / 2),
         Align::Bottom => last_line.saturating_sub(scrolloff + count),
     }
     .max(view.offset.row + scrolloff)
@@ -4578,8 +4578,12 @@ fn shell_impl(
     }
     let output = process.wait_with_output()?;
 
-    if !output.stderr.is_empty() {
-        log::error!("Shell error: {}", String::from_utf8_lossy(&output.stderr));
+    if !output.status.success() {
+        if !output.stderr.is_empty() {
+            let err = String::from_utf8_lossy(&output.stderr);
+            bail!("Shell error: {}", err.to_string());
+        }
+        bail!("Shell command failed");
     }
 
     let str = std::str::from_utf8(&output.stdout)
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4580,10 +4580,16 @@ fn shell_impl(
 
     if !output.status.success() {
         if !output.stderr.is_empty() {
-            let err = String::from_utf8_lossy(&output.stderr);
-            bail!("Shell error: {}", err.to_string());
+            let err = String::from_utf8_lossy(&output.stderr).to_string();
+            log::error!("Shell error: {}", err);
+            bail!("Shell error: {}", err);
         }
         bail!("Shell command failed");
+    } else if !output.stderr.is_empty() {
+        log::debug!(
+            "Command printed to stderr: {}",
+            String::from_utf8_lossy(&output.stderr).to_string()
+        );
     }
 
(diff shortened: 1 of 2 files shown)


dead10ck (approved):
Looks good to me, thanks! It looks like a lot of this is duplicated, but I think it can be refactored later.

helix-term/src/commands.rs
```diff
@@ -4578,8 +4578,12 @@ fn shell_impl(
...
 
-    if !output.stderr.is_empty() {
-        log::error!("Shell error: {}", String::from_utf8_lossy(&output.stderr));
+    if !output.status.success() {
+        if !output.stderr.is_empty() {
+            let err = String::from_utf8_lossy(&output.stderr);
+            bail!("Shell error: {}", err.to_string());
+        }
+        bail!("Shell command failed");
     }
```
the-mikedavis, line 4588: With this change we loose the stderr going to `log::error!` if `output.status.success() == true`. We should always send stderr to the log:

- if the command succeeded, the stderr would otherwise be discarded
- if the command failed, the stderr may be longer than the editor's status element and the user may want to see the full stderr in the log

helix-view/src/document.rs
```diff
@@ -440,14 +440,14 @@ impl Document {
...
                 if !output.status.success() {
-                    return Err(FormatterError::NonZeroExitStatus);
+                    if !output.stderr.is_empty() {
+                        return Err(FormatterError::NonZeroExitStatus(Some(
+                            String::from_utf8_lossy(&output.stderr).to_string(),
+                        )));
+                    }
+
+                    return Err(FormatterError::NonZeroExitStatus(None));
                 }
```
the-mikedavis, line 451: Same here, let's always kick the stderr lines to the log

helix-view/src/document.rs
```diff
@@ -1108,10 +1107,12 @@ impl Display for FormatterError {
             }
             Self::BrokenStdin => write!(f, "Could not write to formatter stdin"),
             Self::WaitForOutputFailed => write!(f, "Waiting for formatter output failed"),
-            Self::Stderr(output) => write!(f, "Formatter error: {}", output),
             Self::InvalidUtf8Output => write!(f, "Invalid UTF-8 formatter output"),
             Self::DiskReloadError(error) => write!(f, "Error reloading file from disk: {}", error),
-            Self::NonZeroExitStatus => write!(f, "Formatter exited with non zero exit status:"),
+            Self::NonZeroExitStatus(Some(output)) => write!(f, "Formatter error: {}", output),
+            Self::NonZeroExitStatus(None) => {
+                write!(f, "Formatter exited with non zero exit status:")
```
the-mikedavis, line 1114: I think we should either chop off the trailing `:` here or record and show the status code. IIRC that can be None so we may want `NonZeroExitStatus` to instead be

```rs
NoneZeroExitStatus { code: Option<NonZeroI32>, stderr: Option<String> }
```

helix-term/src/commands.rs
dead10ck: >if the command succeeded, the stderr would otherwise be discarded

If the command succeeded, then is it necessary? If there was no error code, then it isn't necessarily an error. Some programs simply use it for international output, e.g. `xargs -t` will print the command it's running to stderr. `ssh -v` prints debug info to stderr. Despite its name, stderr is used for a lot more things than error messages. Maybe this could be useful as a `log::debug!` though.

> if the command failed, the stderr may be longer than the editor's status element and the user may want to see the full stderr in the log

Yeah it might be useful to add a `log::error!` at the [call site](https://github.com/helix-editor/helix/blob/4ce5a94552d887b9b14000dfdd1abe76a6d5f315/helix-term/src/commands.rs#L4610)

helix-view/src/document.rs
PiergiorgioZagaria: I don't think showing the status code is really useful, there are very few cases in which it matters since most of the time stderr is used to explain the error. 
Will remove the trailing ':'.

PiergiorgioZagaria:
Now stderr is always logged, (if the status code is zero `log::debug` is used else if there was an error `log::error` is used). In case of errors stderr still returns the error, this is because for small error messages it's more convenient to read them in the editor without opening the logs

dead10ck (approved):

PR #3359, opened 2022-08-08
Improve `:set-language` command

A-Walrus:
Add `text` to completion for possible languages, and report error if user inputs invalid language.

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -671,13 +671,12 @@ impl Document {
         &mut self,
         language_id: &str,
         config_loader: Arc<syntax::Loader>,
-    ) -> bool {
-        let language_config = config_loader.language_config_for_language_id(language_id);
-        if language_config.is_none() {
-            return false;
-        }
-        self.set_language(language_config, Some(config_loader));
-        true
+    ) -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


helix-view/src/document.rs
```diff
@@ -671,9 +671,13 @@ impl Document {
         &mut self,
         language_id: &str,
         config_loader: Arc<syntax::Loader>,
-    ) {
+    ) -> bool {
```
the-mikedavis, line 674: I think it makes sense for this function to return a Result with the context that the language ID is invalid here rather than in `language`.

the-mikedavis:
Does this cause any trouble with the language server? Do we shut down the language server if we say `:lang text`?

A-Walrus:
> Does this cause any trouble with the language server? Do we shut down the language server if we say `:lang text`?

This seems to be an already existing bug. Currently on `22.05` when changing a language the language server is not killed. In fact if I change it to a language that I don't have the language server for (`go` in my case), not only is the language server not shut-down, but it is also still responding to requests: [asciinema](https://asciinema.org/a/dHcT58nGFH69zBXX3H0Z1GLv0)

the-mikedavis (approved):

PR #3360, opened 2022-08-08
Step to previous and next item from the last picker

groves:
You can do this with `space ' down ret`, but this gets it down to `] '`. Feels nicer to me when going through a big set of global search matches.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -271,6 +271,8 @@ impl MappableCommand {
         diagnostics_picker, "Open diagnostic picker",
         workspace_diagnostics_picker, "Open workspace diagnostic picker",
         last_picker, "Open last picker",
+        goto_next_from_last_picker, "Goto next from last picker",
+        goto_prev_from_last_picker, "Goto previous from last picker",
         prepend_to_line, "Insert at start of line",
         append_to_line, "Append to end of line",
         open_below, "Open new line below selection",
@@ -2436,6 +2438,26 @@ fn last_picker(cx: &mut Context) {
         // XXX: figure out how to show error when no last picker lifetime
         // cx.editor.set_error("no last picker")
     }));
+}
+
+fn goto_next_from_last_picker(cx: &mut Context) {
+    cx.callback = Some(Box::new(
+        move |compositor: &mut Compositor, cx: &mut compositor::Context| {
+            if let Some(picker) = &mut compositor.last_picker {
+                picker.as_pickstepper().unwrap().next(cx);
+            }
+        },
+    ));
+}
+
+fn goto_prev_from_last_picker(cx: &mut Context) {
+    cx.callback = Some(Box::new(
+        move |compositor: &mut Compositor, cx: &mut compositor::Context| {
+            if let Some(picker) = &mut compositor.last_picker {
+                picker.as_pickstepper().unwrap().prev(cx);
+            }
+        },
+    ));
 }
 
 // I inserts at the first nonwhitespace character of each line with a selection
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -106,6 +106,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
             "o" => goto_prev_comment,
             "t" => goto_prev_test,
             "p" => goto_prev_paragraph,
+            "'" => goto_prev_from_last_picker,
             "space" => add_newline_above,
         },
         "]" => { "Right bracket"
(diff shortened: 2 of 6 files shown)


helix-term/src/commands.rs
```diff
@@ -2438,6 +2440,26 @@ fn last_picker(cx: &mut Context) {
     }));
 }
 
+fn goto_next_from_last_picker(cx: &mut Context) {
+    cx.callback = Some(Box::new(
+        move |compositor: &mut Compositor, cx: &mut compositor::Context| {
+            if let Some(picker) = &mut compositor.last_picker {
+                picker.as_pickstepper().unwrap().prev(cx);
```
the-mikedavis, line 2447:
```suggestion
                picker.as_pickstepper().unwrap().next(cx);
```

helix-term/src/commands.rs
```diff
@@ -2438,6 +2440,26 @@ fn last_picker(cx: &mut Context) {
     }));
 }
 
+fn goto_next_from_last_picker(cx: &mut Context) {
+    cx.callback = Some(Box::new(
+        move |compositor: &mut Compositor, cx: &mut compositor::Context| {
+            if let Some(picker) = &mut compositor.last_picker {
+                picker.as_pickstepper().unwrap().next(cx);
```
groves, line 2447: I'm sad about adding a method to Component for this, but I couldn't see a better way to do it.

I was able to get it to work without adding a method with this replacing `as_pickstepper`:

```
                if let Some(fp) = picker
                    .as_any_mut()
                    .downcast_mut::<Overlay<FilePicker<PathBuf>>>()
                {
                    fp.content.next(cx);
                }
```

but saying that "works" is stretching it, especially since we'd need to add another block for each possible Picker type.

Maybe someone who knows more Rust has a better idea?

helix-term/src/keymap/default.rs
```diff
@@ -106,6 +106,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
             "o" => goto_prev_comment,
             "t" => goto_prev_test,
             "p" => goto_prev_paragraph,
+            "'" => goto_prev_from_last_picker,
```
groves, line 109: I went with `'` to match `space '` but I'm not tied to it.

helix-term/src/commands.rs
groves: Doh!

the-mikedavis:
I'm not necessarily against these commands/bindings but I don't really understand the use-case. Why not use the picker? The preview pane is there to view the code-context of a search-result/diagnostic/jump/etc so that you don't have to exit the picker.

With these commands you discard the context that the picker gives: you have no idea where you'll jump next or how many results are left in the picker.

sudormrfbin:
This would enable an ad-hoc quickfix list behaviour, carried over from vim. In vim a list of file locations can be sent to the quickfix list, and commands exist that take us to the next/previous item in the list (usually bound to `]q` and `[q`). But notably a small window appears at the bottom with the list, and the list itself is editable (delete and reorder items).

Even without the quickfix _window_, the keybinds themselves are useful; imagine navigating between workspace diagnostics. I can open it once, close it and then jump between all the errors with `]'` and `['`. `]d` would only navigate between diagnostics in the file, and navigation that way is possible only because there's a dedicated keybind for diagnostics navigation and can't be generalized.

groves:
@sudormrfbin, thanks! That's a lot of the inspiration of this for me.

In particular I see my self using it with global search and goto references. With both of those, I'll find something that matches a smallish number of things I want to step through. I don't find the preview pane useful for that as I can't move in the source when I'm on a match and I often want to do that.

The context would be nice to keep like with Vim's quickfix window. I'm hoping to add the ability to put picker results in a split window in a later PR. Like @sudormrfbin says though, this is already useful.

helix-term/src/ui/picker.rs
```diff
@@ -34,6 +34,11 @@ pub const MIN_AREA_WIDTH_FOR_PREVIEW: u16 = 72;
 /// Biggest file size to preview in bytes
 pub const MAX_FILE_SIZE_FOR_PREVIEW: u64 = 10 * 1024 * 1024;
 
+pub trait PickStepper {
+    fn next(&mut self, cx: &mut Context);
+    fn prev(&mut self, cx: &mut Context);
+}
```
the-mikedavis, line 40: Can we do this without introducing a new trait for it? A trait feels like the wrong tool.

Maybe we can somehow downcast the `last_picker` Component to a `Picker` and add next/prev there?

the-mikedavis:
This could actually be implemented with a macro. On #4709:

```toml
[keys.normal]
"]"."'" = "@<space>'<C-n><ret>"
"["."'" = "@<space>'<C-p><ret>"
```

fastfading:
➜  bin hx ~/.config/helix/config.toml
Bad config: TOML parse error at line 17, column 11
   |
17 | "]"."'" = "@<space>'<C-n><ret>"
   |           ^^^^^^^^^^^^^^^^^^^^^
No command named '@<space>'<C-n><ret>'

the-mikedavis:
That keybinding needs the #4709 branch

Issue #2409, opened 2022-05-05
Undoing change in insert mode does nothing when first opening the file


joezak11:
### Summary

using the following configuration
```toml
[keys.insert]
C-z = "undo"
```
editing a file without going to normal mode prints `Already at oldest change` if I go to normal mode and go back to insert mode I can undo

### Reproduction Steps

 

I tried this:

1. `hx`
2. enter insert mode
3. type anything
4. press `Ctrl-z`

I expected this to happen:
to undo the changes
Instead, this happened:
I get `Already at oldest change`

### Platform

Linux

### Terminal Emulator

alacritty 0.8.0-dev (3ed0430)

### Helix Version

helix 22.05-dev (fc617968)

joezak11:
I see. Are there any plans for changing that behavior? neovim works as expected, I'm not sure how they implement it, though. could be time based or event based with a debounse or something.

the-mikedavis:
There's been some discussion about more granular or automatic undo checkpoints in insert mode based on a timeout or a number of words typed but many of us agree that it's better to be explicit about the checkpoints (i.e. use `C-s` or switch to normal mode) or else you end up having to hit `u` (or `C-z` in this case) an unpredictable number of times to go back to the last explicit checkpoint.

See https://github.com/helix-editor/helix/issues/1583#issuecomment-1022812303

PR #3366, opened 2022-08-09
Make initial editing mode configurable

p-e-w:
This simple change makes it possible to set the default editing mode for newly opened editors to something other than normal mode. For example, to start in insert mode, just add

```toml
[editor]
initial-mode = "insert"
```

to the config.

I've always wished that Vim provided this option, since I find starting in insert mode more useful when I launch the editor to quickly write something down.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -49,6 +49,7 @@ You may also specify a file to use for configuration with the `-c` or
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
+| `initial-mode` | The initial mode for newly opened editors. | `"normal"` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
 
 ### `[editor.statusline]` Section
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -163,6 +163,8 @@ pub struct Config {
     pub whitespace: WhitespaceConfig,
     /// Vertical indent width guides.
     pub indent_guides: IndentGuidesConfig,
+    /// The initial mode for newly opened editors. Defaults to `"normal"`.
+    pub initial_mode: Mode,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
 }
@@ -498,6 +500,7 @@ impl Default for Config {
             rulers: Vec::new(),
             whitespace: WhitespaceConfig::default(),
             indent_guides: IndentGuidesConfig::default(),
+            initial_mode: Mode::Normal,
             color_modes: false,
         }
     }
@@ -906,6 +909,7 @@ impl Editor {
 
     /// Generate an id for a new document and register it.
     fn new_document(&mut self, mut doc: Document) -> DocumentId {
+        doc.mode = self.config().initial_mode;
         let id = self.next_document_id;
         // Safety: adding 1 from 1 is fine, probably impossible to reach usize max
         self.next_document_id =


CBenoit (approved):
The change is simple enough and looks good to me
I just want to wait for another maintainer to approve the new option before merging

AlexanderBrevig:
Seems to me this would also set any opened documents to change the mode?
Say if you `gd` to another file?

Apologies if I did not read it correctly on my phone. I'll test it later today :)

p-e-w:
@AlexanderBrevig

I can't reproduce that (or perhaps I'm misunderstanding you?). My steps are:

1. Set `initial-mode = "insert"` in the config.
2. Open two files.
3. Both are opened in insert mode, as desired.
4. Switch to normal mode in both files.
5. Switch back and forth between the files (I use `gn`, since I don't have LSP set up at the moment).
6. Both files remain in normal mode, as desired.

AlexanderBrevig:
Try navigating to a new file. Use `<spc>f` or `gd` on a symbol. The new file is also insert mode.

p-e-w:
I think I understand now. You would like *only the first document opened in a Helix session* to be opened in the specified mode.

That's not what I want, so as far as I'm concerned, this PR is working as intended. I'm also still happy with the name `initial-mode`, since it is the initial editing mode for documents. `new-document-mode` could be misleading, because it might wrongly suggest the mode applies only to newly created documents, rather than also documents opened from files.

If you are indeed making a PR for the alternate configuration you have proposed, may I suggest the configuration name `launch-mode` instead? I think `launch-mode` and `initial-mode` are simpler and more accurate than `initial-mode` and `new-document-mode`.

AlexanderBrevig:
> This simple change makes it possible to set the default editing mode for newly opened editors to something other than normal mode. 

I really don't understand the use case for opening every document in INS mode.
But "newly opened _editors_" makes a whole lot of sense. 

I've spoken my mind (a few times already, sorry) and will wait with `launch-mode` till after this is merged. 

>initial
/ɪˈnɪʃ(ə)l/
adjective
existing or occurring at the beginning.

p-e-w:
> I really don't understand the use case for opening every document in INS mode.

Long term, I want to make it possible to configure Helix as a *modeless* editor (just like some modeless editors can be configured for modal editing). Starting in insert mode (and never leaving it afterwards) is one of many puzzle pieces required for that to work.

As for the config name, I'm definitely still open to suggestions. "Initial" here refers to the initial state of each *document,* which IMO is logical since editing modes are defined on a per-document basis. I've also considered `default-mode` and `standard-mode`, but dislike them because the "default mode" is still normal mode, which we return to when `Esc` is pressed.

dead10ck:
This is a bad idea in my opinion. Helix is *not* a modeless editor. If you want one, there are plenty of other options out there. We should not be encouraging more users to try to use helix as not-modal.

the-mikedavis:
IMO this is too niche to warrant its own config option, especially since Helix is inherently modal as dead10ck says. I think this would be more elegant when scripting support comes - probably with some sort of hook mechanism that enters insert mode when on buffer creation. Until scripting support comes, this is a one-liner to maintain in a fork and merge conflicts should be trivial.

CBenoit (commented):
> I think this would be more elegant when scripting support comes - probably with some sort of hook mechanism that enters insert mode when on buffer creation.

Fair point

p-e-w:
Since Helix [suffers from the same fundamental bug as Vim and Kakoune](https://github.com/helix-editor/helix/issues/133), its default keymap is essentially unusable with many non-English keyboard layouts. So the user has the option to [invent their own layout-independent modal bindings](https://github.com/p-e-w/envy), or try to configure standard modeless bindings.

If you don't want to come up with your own bindings, and modeless configuration isn't supported, Helix effectively becomes a (keyboard) hardware-dependent editor, which would be an extremely sad limitation considering Helix's many other strengths.

dead10ck:
I feel like I am saying this way more often than it should be said, but if you try to shoehorn a modal editor into a non-modal editor, you're gonna have a bad time. A non-exhaustive list of issues people have had that are a direct result of trying to make helix work in insert mode 100% of the time:

* https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126
* https://github.com/helix-editor/helix/issues/1381
* https://github.com/helix-editor/helix/issues/2409
* https://github.com/helix-editor/helix/issues/1583#issuecomment-1022812303
* surely others I do not have the time to comb through
* others reported in the chat room

I won't repeat myself about the reasoning again, feel free to read the issues. Helix is a modal editor, and we should not be encouraging its use as a non-modal editor.

The history of the world's software being heavily biased toward English and QWERTY is indeed unfortunate, but much much larger than this text editor, and we are not on a mission to solve the world's incompatibilities between locale, hardware, and software.

p-e-w:
Almost all source code in existence consists exclusively of ASCII symbols, yet Helix goes to *enormous* lengths to support Unicode, sacrificing simplicity and performance in the process. But the huge number of people who don't/can't use en-US keyboards get told to either figure out their own config or pound sand?

That makes no sense whatsoever.

dead10ck:
Sure it does. Most of the heavy lifting with Unicode is done for us already in the form of well tested libraries. Editing and displaying text in a single well defined international standard format is a much different beast than coming up with a key binding layout that works for the literal hundreds of physical keyboard layouts in existence. I'm not sure you're taking the time to appreciate the scope of that task before making demands of a loosely organized group of volunteers who work on this piece of software in their free time for no compensation.

This is also getting very off topic for this PR.

p-e-w:
> I'm not sure you're taking the time to appreciate the scope of that task before making demands of a loosely organized group of volunteers who work on this piece of software in their free time for no compensation.

Umm what? **I** created this PR. It's **me** who's working towards implementing a solution here. I make no demands from anybody. You commented here dropping a bunch of vaguely related issues that supposedly demonstrate why making the initial mode a config option is wrong, while offering *nothing* regarding the actual problem, besides "bad luck, that's how it is".

But whatever. If the attitude demonstrated in your comments is in any way representative of the Helix community as a whole, that's all I need to know.

dead10ck:
>It's *me* who's working towards implementing a solution here.

That's the thing, you're really not. You've come to a modal editor, expecting it to be made into a non-modal editor. You've come here with a one line change that saves you one key stroke, and does nothing to address the problem that the code base is designed around the modal editing paradigm and has problems when you try to use it otherwise. The issues I've linked are only a few of the problems, and yet they are not easy to fix.

There would be a lot of work that needs doing to support a non-modal editing paradigm, and all this PR has done is make it slightly easier to find out the hard way that this doesn't work well.

You're not the one who is doing the work to make non-modal editing work well, nor would you be the one who sticks around long-term to support any issues that come up around the non-modal paradigm; and more importantly, coming into a project and expecting a totally different editing paradigm to be supported is demanding, whether you intended it that way or not.

Helix has always advertised itself as a modal editor. If you don't want a modal editor, there are plenty of other great options out there. And there is nothing we can do about the history of bias in technological choices.

I suggest taking some time to reflect on the sense of entitlement you have about other people supporting the use cases you need. No one who works on this project owes you anything.

archseer:
Apologies, I've only seen this thread now. @p-e-w I'd like to apologize because this is certainly not the tone I'd like to set for the community. While I don't merge every PR, I think it's at least worth discussing proposals.

>  I'm not sure you're taking the time to appreciate the scope of that task before making demands of a loosely organized group of volunteers who work on this piece of software in their free time for no compensation. Kindly take your snark elsewhere.

I don't like the tone here. @p-e-w was just genuinely trying to fill a need they had, and none of the comments were snarky. We should strive to have a civilized discussion here rather than scaring away developers that are willingly trying to contribute.

While this specific PR is a one-line change, I, for one appreciated @p-e-w's input in https://github.com/helix-editor/helix/pull/1623#issuecomment-1042612811 for example, given the past experience with the Micro editor.

> I suggest taking some time to reflect on the sense of entitlement you have about other people supporting the use cases you need. No one who works on this project owes you anything.

I also didn't see this implied anywhere. Your reply was far sharper.

---

Now, regarding the PR itself:


There is some precedent in `set insertmode` in Vim. Kakoune on the other hand supports it with a script: https://github.com/mawww/kakoune/issues/4123

> I want to make it possible to configure Helix as a modeless editor (just like some modeless editors can be configured for modal editing). Starting in insert mode (and never leaving it afterwards) is one of many puzzle pieces required for that to work.

I don't hold this as a design goal myself and wouldn't ship modeless mappings out of the box, but we should at least strive to make commands flexible enough to make this work with a custom config. We are already mostly there because commands aren't mode dependent, select mode is just simply a second set of mappings overlayed. It gets a bit trickier with undo state handling though.

p-e-w:
Thanks for the feedback, @archseer. FWIW, the roadmap I had in mind towards configuring Helix as a modeless editor is something like:

1. Reset all default keybindings since we want to pretty much start from scratch (made possible by #2733)
2. Allow starting in insert mode (this PR)
3. Add an option for time-batched undo handling
4. Add an option for allowing selections in insert mode

I don't think there is much more to it than those items. While 3 and 4 are by no means trivial, they don't seem like insurmountable obstacles either.

A "modeless" editor is just a modal editor with a single mode. The reason why Helix can not yet be used modelessly isn't that Helix is a modal editor, but that Helix *hardcodes* certain behaviors into specific modes, rather than making them options for configuration. With enough options, it might even be possible to remove all mentions of "insert mode", "normal mode" etc. from the code itself and make modes (including the number of modes!) configuration-only, which would enable radical experimentation with new editing models that are modal, but not based on Vim.

It's also worth noting that there are currently **zero** modeless TUI editors that support LSP, DAP, Tree-Sitter, and many of the other great features Helix has. Making it possible to use Helix modelessly could therefore unlock a substantial market of users who work in the terminal and are currently stuck with Nano or Micro, neither of which provide advanced language features.

CBenoit:
> With enough options, it might even be possible to remove all mentions of "insert mode", "normal mode" etc. from the code itself and make modes (including the number of modes!) configuration-only, which would enable radical experimentation with new editing models that are modal, but not based on Vim.

This is something I was thinking about too!
Similar to how i3-wm allow the user to define his very own modes besides the "default" one.

j-james:
I would greatly appreciate this PR.

Helix isn't just a good modal editor, it's also just a good editor regardless - and as @p-e-w went over, the additional changes needed to make Helix work in a non-modal fashion are rather small.

ShalokShalom:
As someone brand new to this community, I am quite surprised by Helix being less hostile to newbies than Vim and Emacs traditionally happen to be, in my experience. 

Helix is the first modal editor, that makes sense for me. Out of the box, and even after configuring it a bit for myself. I must say, in parts, that had been easier than with VSCode. 

That is also, why I think its very sad, and potentially extremely damaging to the project, if hostile behavior is encouraged by simply watching it happening. Just saying

---

Now about the issue: I like to fire up my editor, and start writing. Has nothing to do with using the editor in a non-modal fashion. I like the modes, they make sense here. I just like to sit down, and start typing. 

The mode in which the editor starts for me, is preferably insert. Can we make this happen?

ShalokShalom:
@p-e-w Could you open this once again, please 😃

PR #3370, opened 2022-08-09
Update tree-sitter-latex and highlights

erasin:
update tree-sitter-latex 

update highlights from https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/latex/highlights.scm
add textobject from https://github.com/nvim-treesitter/nvim-treesitter-textobjects/blob/master/queries/latex/textobjects.scm

About highlights

I'm not sure about the markup for the Math section， It wasn't marked before. I left it as it was.

### DIFF
--- a/runtime/queries/latex/highlights.scm
+++ b/runtime/queries/latex/highlights.scm
@@ -103,23 +103,22 @@
   name: (curly_group_text (_) @string))
 
 ;; Math
-[
-  (displayed_equation)
-  (inline_formula)
-] @text.math
+
+(displayed_equation) @markup.raw.block
+(inline_formula) @markup.raw.inline
 
 (math_environment
   (begin
-    command: _ @text.math
-    name: (curly_group_text (text) @text.math)))
+    command: _ @function.builtin
+    name: (curly_group_text (text) @markup.raw)))
 
 (math_environment
-  (text) @text.math)
+  (text) @markup.raw)
 
 (math_environment
   (end
-    command: _ @text.math
-    name: (curly_group_text (text) @text.math)))
+    command: _ @function.builtin
+    name: (curly_group_text (text) @markup.raw)))
 
 ;; Sectioning
 (title_declaration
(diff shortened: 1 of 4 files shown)


runtime/queries/latex/highlights.scm
```diff
@@ -1,410 +1,236 @@
...
-  "\\textcolor"
-  "\\pagecolor"
-  "\\definecolor"
-  "\\definecolorset"
-  "\\newtheorem"
-  "\\declaretheorem"
-  "\\newacronym"
-] @function.macro
+(key_value_pair
+  key: (_) @parameter
```
the-mikedavis, line 9:
```suggestion
  key: (_) @variable.parameter
```

runtime/queries/latex/highlights.scm
```diff
@@ -1,410 +1,236 @@
...
+  command: _ @function.macro
+  from: (curly_group_text (_) @label)
+  to: (curly_group_text (_) @label))
+(label_reference
+  command: _ @function.macro
+  names: (curly_group_text_list (_) @label))
+(label_number
+  command: _ @function.macro
+  name: (curly_group_text (_) @label)
+  number: (_) @text.reference)
```
the-mikedavis, line 78:
```suggestion
  number: (_) @markup.link.label)
```

I'm not really sure about this one but it seems to be the closest.

runtime/queries/latex/highlights.scm
```diff
@@ -1,410 +1,236 @@
...
-  "\\subsection"
-  "\\subsubsection"
-  "\\paragraph"
-  "\\subparagraph"
-] @type
-
-"\\item" @punctuation.special
+  (displayed_equation)
+  (inline_formula)
+] @text.math
```
the-mikedavis, line 109: Let's split these into

```scm
(displayed_equation) @markup.raw.block
(inline_formula) @markup.raw.inline
```

the-mikedavis (commented):
Instead of `text.math` let's use `markup.raw`

runtime/queries/latex/highlights.scm
erasin: ok

the-mikedavis (approved):
Looks good, thanks for updating these!

PR #3378, opened 2022-08-09
adds xcopy param to recursively copy runtime folder

jdrst:
xcopy doesn't copy recursively, so subfolders (themes, grammars etc.) aren't getting copied atm (see [here](https://docs.microsoft.com/de-de/windows-server/administration/windows-commands/xcopy))
adding `/e` fixes that.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #3379, opened 2022-08-10
Revert conditional setting of stdin handle

### LINKED ISSUE #3374: Running a shell command breaks terminal sequences
### Summary

After running a shell command some inputs don't work anymore.

I ran into these two:

- mouse input has no effect
- <kbd>Ctrl</kbd><kbd>C</kbd> sends a keyboard interrupt and kills helix, instead of toggling a comment

It's worth noting that `:pipe` does not cause this broken behavior.

### Reproduction Steps

- run helix
- `:sh echo hi`
- try using the mouse or toggling a comment (<kbd>Ctrl</kbd><kbd>C</kbd>)

### Helix log

nothing of interest

### Platform

Windows 10 x64

### Terminal Emulator

wezterm (nightly)

### Helix Version

helix 22.05 (3e11017a)

AceofSpades5757:
fixes #3374

I still like the change to avoid setting stdio if necessary, but it looks like this causes an issue on Windows.

After attempting other fixes, reverting the changes was the only thing that worked.

P.S. Conditional compilation could be added for non-Windows targets, but I think this would not add value to the codebase or Helix.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4552,14 +4552,23 @@ fn shell_impl(
     use std::process::{Command, Stdio};
     ensure!(!shell.is_empty(), "No shell set");
 
-    let mut process = match Command::new(&shell[0])
+let mut process = Command::new(&shell[0]);
+    process
         .args(&shell[1..])
         .arg(cmd)
         .stdout(Stdio::piped())
-        .stderr(Stdio::piped())
-        .stdin(Stdio::piped())
-        .spawn()
+        .stderr(Stdio::piped());
+
+    #[cfg(not(target_family = "windows"))]
+    if input.is_some() {
+        process.stdin(Stdio::piped());
+    }
+    #[cfg(target_family = "windows")]
     {
+        process.stdin(Stdio::piped());
+    }
+
+    let mut process = match process.spawn() {
         Ok(process) => process,
         Err(e) => {
             log::error!("Failed to start shell: {}", e);
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4552,18 +4552,14 @@ fn shell_impl(
     use std::process::{Command, Stdio};
     ensure!(!shell.is_empty(), "No shell set");
 
-    let mut process = Command::new(&shell[0]);
-    process
+    let mut process = match Command::new(&shell[0])
         .args(&shell[1..])
         .arg(cmd)
         .stdout(Stdio::piped())
-        .stderr(Stdio::piped());
-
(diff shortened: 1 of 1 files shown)


AceofSpades5757:
Switched to conditional compilation and tested on Windows and Ubuntu (WSL). Everything seemed to work as intended.

helix-term/src/commands.rs
```diff
@@ -4559,9 +4559,14 @@ fn shell_impl(
...
         .stderr(Stdio::piped());
 
+    #[cfg(not(target_family = "windows"))]
     if input.is_some() {
         process.stdin(Stdio::piped());
     }
+    #[cfg(target_family = "windows")]
+    {
+        process.stdin(Stdio::piped());
+    }
```
the-mikedavis, line 4569: This doesn't need conditional compiling since it compiles ok on and off windows. The runtime behavior should be influenced by it though and that can be checked with the [`cfg!`](https://doc.rust-lang.org/rust-by-example/attribute/cfg.html) macro

```rs
if input.is_some() || cfg!(windows) {
    precss.stdin(Stdio::piped());
}
```

AceofSpades5757: Is this usually more preferable?

archseer (approved):

helix-term/src/commands.rs
the-mikedavis: From what I understand, it's best to only use what conditional compilation you really need.  That way you don't mistakenly end up with wildly different code-paths on different operating systems. At the very least it makes rust-analyzer a little quieter to keep conditional compiles to a minimum :P

AceofSpades5757: Gotcha, gotcha. Thanks!

I did have an issue with my first attempt, even though I was pretty sure it wasn't going to work. It _did_ compile, but the binary would just hang there. The most surprising thing is that this was legal, and it compiled.

```rust
#[cfg(target_family = "windows")]
let a = "hello";
```

zonkki:
Couldn’t reproduce #3374 on Win11 with the following setups:

- Windows Terminal > nu > hx > nu

- Windows Terminal > cmd > hx > cmd

- WezTerm > cmd > hx > cmd

However, as a side effect, the current `cfg!(windows)` logic causes :sh tools like `rg` to bail with `Shell command failed: status 1` on Windows, probably because they detect an open stdin and expect Some input (which is empty).

If cfg!(windows) is removed, it falls back to Stdio::null() and commands like `:sh rg hello` work again, since stdin is closed rather than empty-but-open.

Just wanted to share in case it’s helpful for future decisions.

PR #3380, opened 2022-08-10
treesitter object pickers

QiBaobin:
Add pickers so that we can search beside moving.

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -11,7 +11,6 @@
     - [Goto mode](#goto-mode)
     - [Match mode](#match-mode)
     - [Window mode](#window-mode)
-    - [Treesitter mode](#treesitter-mode)
     - [Space mode](#space-mode)
       - [Popup](#popup)
     - [Unimpaired](#unimpaired)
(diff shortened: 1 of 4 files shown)


EpocSquadron:
I like this in spirit, but the functionality overlaps a bit with the `<space>s` symbol picker. I'd rather see the symbol picker learn to fall back to tree-sitter objects when lsp is not present. It may also be interesting to teach the workspace picker to filter by object type, rather than having separate pickers for each.

heliostatic:
Echoing @EpocSquadron -- I like having a single place to go for symbols, and falling back to treesitter when there's no lsp would be amazing.

QiBaobin:
> I like this in spirit, but the functionality overlaps a bit with the `<space>s` symbol picker. I'd rather see the symbol picker learn to fall back to tree-sitter objects when lsp is not present. It may also be interesting to teach the workspace picker to filter by object type, rather than having separate pickers for each.

done.

heliostatic:
@QiBaobin I just tried this PR, and it's great. One possible improvement: today, if there's no lsp for a language, you get an error on the symbol picker. But with this PR, there's no error, but if the language doesn't have `textobjects` there are no symbols, either (e.g. `.nix` files). If there's no lsp and also no `textobjects` in the tree-sitter, it would be nice to have some sort of message to that effect.

QiBaobin:
> @QiBaobin I just tried this PR, and it's great. One possible improvement: today, if there's no lsp for a language, you get an error on the symbol picker. But with this PR, there's no error, but if the language doesn't have `textobjects` there are no symbols, either (e.g. `.nix` files). If there's no lsp and also no `textobjects` in the tree-sitter, it would be nice to have some sort of message to that effect.

Good point, the logic is added now.

Issue #3223, opened 2022-07-28
Show pseudo-pending text for `ms`/`mr`/`md` commands


AceofSpades5757:
#### Describe your feature request

Add contextual pop-ups, just like when the `Match` box pops up after pressing `m` for when Helix is waiting for the input to the next command, like `Surround delete`.

So, after a user presses `m` and then `d` to do a surround delete, it would be nice to have another box either show their available options, or at least give some instructions on what to do next. This could also help with more advanced subcontexts, such as if there was a regex surround delete.

![image](https://user-images.githubusercontent.com/10341888/181428593-28f6edcf-0feb-40c6-bc84-ebbcbe5560b3.png)

the-mikedavis:
Surround-delete works on the character pressed after `md`. There isn't any need to add future expressive power to surround-delete; you can use sub-selections (`s`) to achieve anything you would want with a regex or special characters.

AceofSpades5757:
@the-mikedavis Thanks. Instead of a pop-up, some sort of indicator would be nice to show that Helix is waiting for additional input to finish the command.

AceofSpades5757:
What about the keys, on the bottom-right of the window, staying until either an input, or canceling?

This is what it looks like after hitting `m` to open the Match context menu. 

![image](https://user-images.githubusercontent.com/10341888/183778865-453fe3fb-d624-4c26-b347-baa8c91875cc.png)

This character(s) is removed when pressing things like `d` for surround delete, or `r` for replace. It would be nice to leave these keys there to indicate that the user is still in the middle of a key sequence.

![image](https://user-images.githubusercontent.com/10341888/183779279-9cd20184-ae47-4767-880b-259a1ab9e1e3.png)

the-mikedavis:
Pseudo-pending seems appropriate for this, especially since we use it elsewhere with `m` commands 👍

If you'd like to take a shot at this, the field is `pseudo_pending: Option<String>` and you can find an example usage in `helix_term::commands::select_textobject`.

AceofSpades5757:
I'll take a crack at it.

PR #3381, opened 2022-08-10
Add pseudo_pending for ms/mr/md

AceofSpades5757:
implements #3223

Adds pseudo pending text for `ms`/`mr`/`md` commands (match surround, replace, and delete).

Currently hard-coded as there appears to be an issue with appending to the current `editor.pseudo_pending` value.

* Can leave it open to not be cleared on the next key press. Leaves lingering pseudo text.
* Can append it, only to have it immediately cleared (appending in the target locations would be cleared immediately before appending).

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4377,7 +4377,9 @@ fn select_textobject(cx: &mut Context, objtype: textobject::TextObject) {
 }
 
 fn surround_add(cx: &mut Context) {
+    cx.editor.pseudo_pending = Some("ma".to_string());
     cx.on_next_key(move |cx, event| {
+        cx.editor.pseudo_pending = None;
         let ch = match event.char() {
             Some(ch) => ch,
             None => return,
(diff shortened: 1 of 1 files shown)


AceofSpades5757:
It looks like `pseudo_pending` is affected by initial key maps such as `<space>`, or anything else opening an auto_info windows. I was unable to find where this is located, or how it is handled. If possible, it would have been nice to implement this same functionality for similar situations that wait on a key press, but that may be out of scope for this pr and issue.

helix-term/src/commands.rs
```diff
@@ -4403,7 +4405,9 @@ fn surround_add(cx: &mut Context) {
 
 fn surround_replace(cx: &mut Context) {
```
the-mikedavis, line 4406: Surround-replace has two `cx.on_next_key`s, one for the starting character and one for the replacement. Nested in the first `cx.on_next_key`, we add the character from the event to the pseudo-pending and then clear it within the second

AceofSpades5757: Thanks, I'll take a look at that.

Issue #763, opened 2021-09-18
Python file indentation is wrong


sekirocc:
### Reproduction steps

1. download latest macOS release 0.4.1
2. open a python file
3. move cursor to some line
4. press `o` to insert newline
5. the cursor is at the beginning of line, without any indentation.

### Environment

- Platform: macOS
- Helix version:  0.4.1

zero config.

sudormrfbin:
Indent support requires an `indents.toml` file describing the tree-sitter syntax nodes where it should be automatically inserted, for example see https://github.com/helix-editor/helix/blob/master/runtime/queries/rust/indents.toml. Python doesn't yet have this file, so contributions are welcome !

lisael:
I think it's not easy (if feasible at all) with the current Tree-sitter only indentation logic. In most editors when I type:

```
if True:
    blah
    else:<ret>
```
I expect the whole `else:` line to be outdented and a new indented line added.

In helix, TreeSitter fails to parse `else:` as an else statement precisely because it's not outdented (another effect is that `else` is not highlighted as a keyword in my example). Therefore

```toml
outdent = [
    "else_clause"
]
```
doesn't work. I can't see another way than parsing the current line ourselves at each return

It's a chicken-and-egg problem.

Note nvimTS has the same issue, which is really annoying, but easier to deal with because of smarttabs, a single <backspace> outdents correctly.

lisael:
That said, a single
```toml
indent = [ ":" ]
```
does a good job at indenting.

willparsons:
I've only had helix for a few days so I may not be understanding this correctly. 

Is this not the indents config for Python?  https://github.com/helix-editor/helix/blob/master/runtime/queries/python/indents.scm

Yet I still don't get auto-indent.

thomasaarholt:
I get autoindentation if I press Enter when the marker is at | in the following:
```python
a = list(|)
->
a = list(
    |
)
```

but I do not get it with the if statement as hinted should happen in the the file @willparsons refers to above.

```python
if a == 1:|
->
if a == 1:
|
```

zSnails:
any news on this? I'm looking forward to switching to helix

BonfireAtNight:
[...]

PR #3382, opened 2022-08-10
Fix python indentation

Triton171:
This introduces 2 new capture types for indent queries. With this, the python indent queries have been improved and re-enabled. It should fix most of the issues from #763. We still don't update the indentation retroactively, so there is still some room for improvements:
```python
if condition:
    do_smth()
    else:
```
Here the `else` should ideally be outdented automatically, once one has finished typing it. This is tricky for 2 reasons:
- We have to figure out when and for which lines to run indentation queries. For performance reasons, we don't want to do it too often and it also shouldn't "correct" the indentation after the user has manually changed it.
- With python, the tree-sitter grammar doesn't even recognize the `else` statement since it is indented (see the discussion in #763). Either the tree-sitter grammar needs to improve here or we need a special case to recognize this and similar situations.

If you have ideas for this, please let me know. Also, if there's other languages than python that can be fixed with the new captures (i.e. languages where functions, classes, etc. end simply where the indentation level decreases), I'm happy to update them as well

### DIFF
--- a/book/src/guides/indent.md
+++ b/book/src/guides/indent.md
@@ -46,6 +46,19 @@ capture on the same line, the indent level isn't changed at all.
 - `@outdent` (default scope `all`):
 Decrease the indent level by 1. The same rules as for `@indent` apply.
 
+- `@extend-indented`:
+Extend the range of this node to the end of the line and to lines that
+are indented more than the line that this node starts on. This is useful
+for languages like Python, where for the purpose of indentation some nodes
+(like functions or classes) should also contain indented lines that follow them.
+
+- `@stop-extend`:
+Prevents the first extension of an ancestor of this node. For example, in Python
+a return expression always ends the block that it is in. Note that this only prevents
+the next extension of one ancestor: If multiple ancestors can be extended (for example
+multiple nested conditional blocks in python), only the extension of the innermost
+ancestor is prevented.
+
 ## Predicates
 
 In some cases, an S-expression cannot express exactly what pattern should be matched.
--- a/book/src/guides/indent.md
+++ b/book/src/guides/indent.md
@@ -46,16 +46,16 @@ capture on the same line, the indent level isn't changed at all.
 - `@outdent` (default scope `all`):
 Decrease the indent level by 1. The same rules as for `@indent` apply.
 
-- `@extend-indented`:
+- `@extend`:
 Extend the range of this node to the end of the line and to lines that
 are indented more than the line that this node starts on. This is useful
 for languages like Python, where for the purpose of indentation some nodes
 (like functions or classes) should also contain indented lines that follow them.
 
-- `@stop-extend`:
+- `@extend.prevent-once`:
 Prevents the first extension of an ancestor of this node. For example, in Python
 a return expression always ends the block that it is in. Note that this only stops the
-extension of the next `@extend-indented` capture. If multiple ancestors are captured,
+extension of the next `@extend` capture. If multiple ancestors are captured,
 only the extension of the innermost one is prevented. All other ancestors are unaffected
 (regardless of whether the innermost ancestor would actually have been extended).
 
--- a/book/src/guides/indent.md
+++ b/book/src/guides/indent.md
@@ -54,10 +54,11 @@ for languages like Python, where for the purpose of indentation some nodes
 
 - `@stop-extend`:
 Prevents the first extension of an ancestor of this node. For example, in Python
-a return expression always ends the block that it is in. Note that this only prevents
-the next extension of one ancestor: If multiple ancestors can be extended (for example
-multiple nested conditional blocks in python), only the extension of the innermost
-ancestor is prevented.
+a return expression always ends the block that it is in. Note that this only stops the
+extension of the next `@extend-indented` capture. If multiple ancestors are captured,
+only the extension of the innermost one is prevented. All other ancestors are unaffected
+(regardless of whether the innermost ancestor would actually have been extended).
+
 
 ## Predicates
 
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -303,8 +303,8 @@ enum IndentScope {
 /// A capture from the indent query which does not define an indent but extends
 /// the range of a node. This is used before the indent is calculated.
 enum ExtendCapture {
-    ExtendIndented,
-    StopExtend,
+    Extend,
+    PreventOnce,
 }
 
 /// The result of running a tree-sitter indent query. This stores for
@@ -394,18 +394,18 @@ fn query_indents(
             let capture_type = match capture_name {
                 "indent" => IndentCaptureType::Indent,
                 "outdent" => IndentCaptureType::Outdent,
-                "extend-indented" => {
+                "extend" => {
                     extend_captures
                         .entry(capture.node.id())
                         .or_insert_with(|| Vec::with_capacity(1))
-                        .push(ExtendCapture::ExtendIndented);
+                        .push(ExtendCapture::Extend);
                     continue;
                 }
-                "stop-extend" => {
+                "extend.prevent-once" => {
                     extend_captures
                         .entry(capture.node.id())
                         .or_insert_with(|| Vec::with_capacity(1))
-                        .push(ExtendCapture::StopExtend);
+                        .push(ExtendCapture::PreventOnce);
                     continue;
                 }
                 _ => {
@@ -478,10 +478,10 @@ fn extend_nodes<'a>(
             if let Some(captures) = extend_captures.get(&deepest_preceding.id()) {
                 for capture in captures {
                     match capture {
-                        ExtendCapture::StopExtend => {
+                        ExtendCapture::PreventOnce => {
                             stop_extend = true;
                         }
-                        ExtendCapture::ExtendIndented => {
+                        ExtendCapture::Extend => {
                             node_captured = true;
                             // We extend the node if
                             // - the cursor is on the same line as the end of the node OR
@@ -574,8 +574,8 @@ pub fn treesitter_indent_for_pos(
         .descendant_for_byte_range(byte_pos, byte_pos)?;
     let (query_result, deepest_preceding) = {
         // The query range should intersect with all nodes directly preceding
-        // the cursor in case one of them is extended.
-        let mut deepest_preceding = None; // The deepest node preceding the cursor
+        // the position of the indent query in case one of them is extended.
+        let mut deepest_preceding = None; // The deepest node preceding the indent query position
         let mut tree_cursor = node.walk();
         for child in node.children(&mut tree_cursor) {
             if child.byte_range().end <= byte_pos {
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -471,6 +471,10 @@ fn extend_nodes<'a>(
         let mut stop_extend = false;
         while deepest_preceding != *node {
             let mut extend_node = false;
+            // This will be set to true if this node is captured, regardless of whether
+            // it actually will be extended (e.g. because the cursor isn't indented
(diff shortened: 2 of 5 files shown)


AceofSpades5757:
It's also causing an issue where it'll automatically update at the line before indentation, such as if you put your cursor at the beginning of `if condition:` in your example and press enter to move it down a line.

P.S. This may have also been in the current implementation. I don't usually code Python with Helix due to the really annoying indentation issues. And thanks for your work!

Triton171:
@AceofSpades5757 good find! The issue already existed in the old version but it's a somewhat rare edge case, so that's probably why it hasn't popped up before. The latest commit should fix it.

AceofSpades5757:
No problem. I've had this version installed while working with Python to test it out. Python is a real sore spot for Helix. It feels like I'm constantly fighting it.

Hopefully this PR will be able to change that 😄

AceofSpades5757:
This has been working pretty well, except for an issue where dedenting with < will be multiple indents rather than a single one. It sounds like some other people have this happen with other languages as well.

Triton171:
That's interesting, the code for dedenting is pretty simple and doesn't use the indentation heuristic. If you can reproduce this, it'd probably be best to create a new issue for this.

AceofSpades5757:
@Triton171 I haven't been able to reproduce it and I've been using it pretty extensively. I may have accidentally re-installed master as my default Helix. Everything has been going swimmingly recently but if something reproducible comes up, I'll comment here (or make a new issue).

Dispersia:
Scala has the same issue, and this appears to work mostly correct for it too, though I am getting a strange behavior

 When trying to add the indent functionality to scala's indent.scm, i'm currently adding

```
[
    (function_definition)
    (val_definition)
] @extend-indented
```

with just this, it seems to work correctly, however when trying to add stop-end

```
[
    (identifier)
] @stop-extend
```
it seems to not work at all and was like its previous behavior. Even having an empty @stop-extend like ` [ ] @stop-extend` does this, which I would expect does nothing

Triton171:
Could you provide an example where the current indentation for Scala is incorrect? I'm not familiar with the language but from a quick look it seems like Scala uses blocks (delimited by curly braces) to define the range of functions. Because of this, I don't think we need `@extend-indented` here. Adding `(function_definition) @extend-indented` would cause indented lines that come after the function definition to be considered part of the function, which is incorrect in this case.

Triton171:
The weird behavior with `@stop-extend` happens because you're writing an invalid indent query: Tree-Sitter doesn't allow empty alternations (for a description of the query syntax, see [their documentation](https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries)). Unfortunately, this currently fails silently and simply uses the fallback method of copying the indentation of the current line.

For getting the indentation right, you should definitely try to improve the Tree-Sitter grammar before tinkering with it in Helix. Once multiline functions are parsed correctly, you could also use something like `(function_definition (identifier) @stop-extend)`, which would only match an identifier node nested directly inside a function. This prevents e.g. an identifier that's part of an assignment expression to be matched. Basically, you have to think about what exactly makes an expression a "return expression" after which the function ends. Tree-Sitter queries are surprisingly powerful, so once you have an idea of what exactly you want in your head, it often isn't too difficult to write a query for it.

If you manage to fix the grammar and need help with indent queries in Helix, feel free to ping me.

the-mikedavis:
> Unfortunately, this currently fails silently and simply uses the fallback method of copying the indentation of the current line.

Recently we added an error log message for this. There isn't a visual indication as there is for syntax highlighting (i.e. no highlights) but you should be able to find out if your indent queries are invalid either by checking the log file or running `cargo xtask query-check` if you are building from source and the modified injections are under this repo's `runtime/` dir

helix-core/src/indent.rs
```diff
@@ -459,40 +492,119 @@ fn query_indents(
...
+                        prev_child = prev_child.child(prev_child.child_count() - 1).unwrap();
+                    }
+                    (
+                        prev_child.byte_range().end - 1..byte_pos + 1,
+                        Some(prev_child),
+                    )
+                }
+                None => (byte_pos..byte_pos + 1, None),
+            }
+        };
```
the-mikedavis, line 541: I think this block can be simplified a bit - `query_range` can be calculated after `prev_child` is determined with a map+unwrap_or. So the whole `let (query_range, prev_child)` is not needed

book/src/guides/indent.md
```diff
@@ -46,6 +46,19 @@ capture on the same line, the indent level isn't changed at all.
...
+are indented more than the line that this node starts on. This is useful
+for languages like Python, where for the purpose of indentation some nodes
+(like functions or classes) should also contain indented lines that follow them.
+
+- `@stop-extend`:
+Prevents the first extension of an ancestor of this node. For example, in Python
+a return expression always ends the block that it is in. Note that this only prevents
+the next extension of one ancestor: If multiple ancestors can be extended (for example
+multiple nested conditional blocks in python), only the extension of the innermost
+ancestor is prevented.
```
the-mikedavis, line 60: I think the names could use some workshopping - I like "extend" but `@stop-extend` doesn't convey well that it only stops the extension once

helix-core/src/indent.rs
```diff
@@ -459,40 +492,119 @@ fn query_indents(
...
         ts_parser.cursors.push(cursor);
-        query_result
+        (query_result, prev_child)
     });
+    let indent_captures = query_result.indent_captures;
+    let extend_captures = query_result.extend_captures;
+
+    // Check for extend captures (starting with the deepest
+    // candidate node and then going up the syntax tree).
+    if let Some(mut prev_child) = prev_child {
```
the-mikedavis, line 558: This block is a bit long and seems to be somewhat well-contained. It may clean this function up a bit to split this into its own function. This function is already quite large and hard to follow

book/src/guides/indent.md
the-mikedavis: Maybe `@indent.extend` and then `@indent.extend.pause`? Or `@extend` and `@extend.pause-once`? I also liked "after" from the discussion in the issue

book/src/guides/indent.md
```diff
@@ -46,6 +46,19 @@ capture on the same line, the indent level isn't changed at all.
 - `@outdent` (default scope `all`):
 Decrease the indent level by 1. The same rules as for `@indent` apply.
 
+- `@extend-indented`:
+Extend the range of this node to the end of the line and to lines that
+are indented more than the line that this node starts on. This is useful
```
the-mikedavis, line 51: I'm curious if we can fork tree-sitter-python and have the function_definition and other rules that end in `:` consume the newline. That way the cursor would belong to the indented node and I think we might not need the extend system. I'll try it out

book/src/guides/indent.md
Triton171: `@extend` sounds good. In case we add more "extend" captures types, we can still rename it to something more specific later. Maybe `@extend.prevent-once` is a more accurate description of what the other capture does? Or `@extend.end-once`?

book/src/guides/indent.md
Triton171: I think the behavior would need to be a bit more complicated: It doesn't just have to consume the newline but also any following indented lines. Additionally, this should not happen when the function ends with for example a `return` expression.

If all this was implemented in the grammar, it would make this PR redundant but I'm not sure if it's worth the effort of forking the tree-sitter grammars for all languages that are affected by this.

Triton171:
I've found an additional issue when there are multiple nested `@extend-indented` nodes:
```python
def func():
    if condition:
        return
    | # <--- Cursor
```
In this situation, pressing enter would result in a line that's not indented at all. This happens since we're not inside the `if_statement`, so the extension prevented by the `return_statement` is the extension of the function. The latest commit slightly changes the behavior of the `@stop-extend` capture to fix this.

helix-core/src/indent.rs
Triton171: Agreed. I also noticed that the treesitter parser was borrowed unnecessarily early, so I fixed that too.

helix-core/src/indent.rs
```diff
@@ -459,40 +554,73 @@ fn query_indents(
...
-    let new_line_break = if new_line {
-        Some((line, byte_pos))
-    } else {
-        None
+    let (query_result, deepest_preceding) = {
+        // The query range should intersect with all nodes directly preceding
+        // the cursor in case one of them is extended.
+        let mut deepest_preceding = None; // The deepest node preceding the cursor
+        let mut tree_cursor = node.walk();
+        for child in node.children(&mut tree_cursor) {
```
dead10ck, line 580: Sorry, this is probably just due to unfamiliarity, but I'm not understanding how you can have the cursor on a node with children that happen *before* the cursor. Could you please give an example of this?

Triton171: I think my comments are somewhat misleading: By "cursor" I mean the position of the indent query (not the tree-sitter cursor). Usually, this will just be the position of the text cursor when starting a new line. I'll change the comments to clarify that. 

An example then looks like this:
```python
def func(): # <-- The function_definition node ends here
    
    # <--- The indent is queried for this position
```
The `function_definition` node ends before the position of the indent query (but since it directly precedes the query position, it is considered for extension).

archseer (approved):
Thanks for working on this! There's been a few testers that are using this branch already so I think this is fine to merge 🎉

book/src/guides/indent.md
zyklotomic: Just a question as a reader, why is it called `@extend.prevent-once`? More specifically, I'm not sure I understand how it is semantically "once". If I understand correctly, the capture type is supposed to convey that the extended range of the most immediate ancestor ends. Would `@extend.end-immediate-ancestor` be more apt albeit verbose?

Triton171: I think the wording in the docs might be a bit confusing: It prevents the first extension of an ancestor, but this ancestor doesn't necessarily have to be immediate. Rather, this refers to the first ancestor _that is captured by an `@extend` capture_.

The last sentence in the docs ("All other ancestors are unaffected (regardless of whether the innermost ancestor would actually have been extended)") could probably be explained a bit better. What it means is that the first ancestor that is captured by an `@extend` may not actually cause any extension (for example because the lines after it are not indented enough). In this case, `@extend.prevent-once` "prevents" this extension, but since the extension wouldn't have happened anyways, it has no effect. In particular, if some more distant ancestor is actually extended, this extension is _not_ prevented.

For python, these subtleties don't really make a difference but I still think it's important to document them, since for other languages they might be important. If you find a better name than `@extend.prevent-once` or a better wording for the docs, feel free to open a PR.

zyklotomic: Ah thank you for the explanation! So technically, we can say the most immediate ancestor captured by `@extend` capture.

Would it be easier to simply think of `@extend` and `@extend.prevent-once` as opposite operations, like opening and closing parentheses? We could then rename them into `@extend.begin-once` and `@extend.end-once`, or `@extend.open` and `@extend.close` to emphasize this; something with symmetry in the naming. Is this a correct characterization?

Triton171: `@extend` and `@extend.prevent-once` are sort of opposites but it's not completely symmetric. With parentheses, there's usually exactly one closing parenthesis for each opening parenthesis. Here, this is not the case: There can be multiple `@extend` captures without any `@extend.prevent-once` (in which case the innermost one takes precedence) and there might be multiple `@extend.prevent-once` captures without any `@extend` (although they have no effect in this situation). Also, the `@extend.prevent-once` doesn't appear _after_ the `@extend` but rather as a descendant of it, in other words, it is _contained_ in it.

I'm not sure if there is a better name for this, but the documentation could definitely be improved. Maybe a link to the python indent queries could also be helpful as an example for how to use the captures in practice.

PR #3387, opened 2022-08-10
Update Clojure language definition

IGJoshua:
Hello,

I'd like some feedback on this PR since I don't fully understand all the fields that are being set in this table.

Does this edit look complete, or does the injection regex need to be updated in order to match some of the other file types, like `edn` and `boot`?

Otherwise, this is a simple update to add support for Clojure files supported by clojure-lsp but which aren't reflected in helix's language definition, and to correct the comment token.

File types
- `cljs` for ClojureScript
- `cljc` for cross-compiled Clojure files
- `clje` for Clojerl
- `cljr` for ClojureCLR
- `cljx` for legacy cross-compiled Clojure
- `edn` for Extensible Data Notation files, they are to Clojure what JSON is to JS
- `boot` for build files using the boot build tool

The last thing I'd like confirmation on is that `roots` defines a set of alternate files to detect a project root, and not that all must be present to detect a project, as the files here are for the major build tools, and it's quite common to have only one of these and none of the others.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1526,9 +1526,9 @@ source = { git = "https://github.com/victorhqc/tree-sitter-prisma", rev = "17a59
 name = "clojure"
 scope = "source.clojure"
 injection-regex = "(clojure|clj)"
-file-types = ["clj"]
-roots = ["project.clj"]
-comment-token = ";;"
+file-types = ["clj" "cljs" "cljc" "clje" "cljr" "cljx" "edn" "boot"]
+roots = ["project.clj" "build.boot" "deps.edn" "shadow-cljs.edn"]
+comment-token = ";"
 language-server = { command = "clojure-lsp" }
 indent = { tab-width = 2, unit = "  " }
 
--- a/languages.toml
+++ b/languages.toml
@@ -1525,7 +1525,7 @@ source = { git = "https://github.com/victorhqc/tree-sitter-prisma", rev = "17a59
 [[language]]
 name = "clojure"
 scope = "source.clojure"
-injection-regex = "(clojure|clj)"
+injection-regex = "(clojure|clj|edn|boot)"
 file-types = ["clj", "cljs", "cljc", "clje", "cljr", "cljx", "edn", "boot"]
 roots = ["project.clj", "build.boot", "deps.edn", "shadow-cljs.edn"]
 comment-token = ";"
--- a/languages.toml
+++ b/languages.toml
@@ -1526,8 +1526,8 @@ source = { git = "https://github.com/victorhqc/tree-sitter-prisma", rev = "17a59
 name = "clojure"
 scope = "source.clojure"
 injection-regex = "(clojure|clj)"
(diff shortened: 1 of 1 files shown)


languages.toml
```diff
@@ -1526,9 +1526,9 @@ source = { git = "https://github.com/victorhqc/tree-sitter-prisma", rev = "17a59
 name = "clojure"
 scope = "source.clojure"
 injection-regex = "(clojure|clj)"
-file-types = ["clj"]
-roots = ["project.clj"]
-comment-token = ";;"
+file-types = ["clj" "cljs" "cljc" "clje" "cljr" "cljx" "edn" "boot"]
+roots = ["project.clj" "build.boot" "deps.edn" "shadow-cljs.edn"]
```
the-mikedavis, line 1530: for TOML arrays you'll need commas between the elements

the-mikedavis:
This will update the auto-generated docs as well so you'll need to run `cargo xtask docgen` and commit the results

the-mikedavis:
I think changing the regex to include boot and edn makes sense 👍

> The last thing I'd like confirmation on is that `roots` defines a set of alternate files to detect a project root, and not that all must be present to detect a project

Yep, this is the case. The relevant code is [here](https://github.com/helix-editor/helix/blob/921027fb32018546fa459ffaf4afa4960f56d1b6/helix-loader/src/lib.rs#L107-L119): we find the workspace root (so we can tell the language server about it) by checking for any of the root markers or a `.git` directory

languages.toml
IGJoshua: Ah, thanks for the catch! It's been a while since I manually edited a toml file.

IGJoshua:
I've fixed the missing commas, updated the regex, and I ran `cargo xtask docgen` but it produced no diffs, so no commit was added for that.

the-mikedavis:
Ah yeah sorry I read the diff wrong 😅, there shouldn't be any changes

the-mikedavis (approved):

the-mikedavis:
Thanks!

sogaiu:
@IGJoshua and any others concerned, as far as I'm aware, ClojureCLR uses `clj` just like the JVM Clojure.

See the following for some evidence:

* https://github.com/clojure/clojure-clr/tree/master/Clojure/Clojure.Source/clojure
* https://github.com/arcadia-unity/Arcadia/tree/master/Source/arcadia

If you know otherwise, please let me know as that might be relevant for tree-sitter-clojure.

IGJoshua:
For clarity, the ClojureCLR port of Clojure can and will use `.clj` filetypes, however it will first look for `.cljr` filetypes to allow a single library to be created with host-specific namespaces. This is a part of the support for the `#?(:clj ... :cljr ... :cljs ...)` syntax for reader conditionals in `.cljc` files.

https://github.com/clojure/clojure-clr/blob/master/Clojure/Clojure/Lib/RT.cs#L3229

sogaiu:
@IGJoshua Thanks a lot for the explanation!

Issue #1085, opened 2021-11-12
Some configs do not work


python128:
I am not able to allow certain keymaps to work.
Here is my `config.toml`
```toml
theme = "bogster"

[lsp]
display-messages = true

[keys.normal]
S-left = "extend_char_left"
S-right = "extend_char_right"
S-up = "extend_line_up"
S-down = "extend_line_down"

"#" = "toggle_comments"
C-c = "no_op"
q = { q = "wclose" }
"space" = { "space" = "keep_primary_selection" }

A-o = "add_newline_below"
A-O = [ "open_above", "normal_mode" ]

[keys.insert]
C-backspace = "delete_word_backward"
```

- Here, `C-O` and `C-backspace` doesn't work.
- I have tried `C-O` with `add_newline_above` and the given config both.
- `A-o` seems to work perfectly fine.

### Environment

- Platform: Linux(WSL)
- Helix version: v0.5.0-94-gd3def16 (from `git describe`)

[Link](https://python128.github.io/gifs/issue1.gif) to gif.

LollipopFt:
just a suggestion (it may not work) but i think special keys like left and right should be enclosed in ```"```?

LollipopFt:
[see here](https://docs.helix-editor.com/remapping.html), the entire command is encapsulated in double quotes.

Omnikar:
> just a suggestion (it may not work) but i think special keys like left and right should be enclosed in `"`?

Quotes are only required by special TOML characters like `.`, `#`, `{`, etc.

sudormrfbin:
> `C-backspace` doesn't work

Backspace with modifiers have been problematic for terminal emulators to handle in general (some represent it as `control+h`, some send different keycodes, etc). There's also an open ticket in the crossterm repo: https://github.com/crossterm-rs/crossterm/issues/504.

pickfire:
Yeah, some keys are special, for example `C-i` is `tab` so I think only one of them works.

python128:
Is `A-O` also an ambiguous key?
If so, why does it work in Kak? (Is it because it is not crossterm or something related?)

David-Else:
Hi, the `C-backspace = "delete_word_backward"` not working is killing my typing. I think it is very important for touch typing to be able to easily delete the last word in insert mode, it is nearly always better to delete the misspelled word entirely and start again for the sake of muscle memory. `ctrl-backspace` is the standard in most software (for example, right here in editing this message on github it works)

[...]

PR #3390, opened 2022-08-11
Update to crossterm-0.25

groves:
I've got a couple PRs that depend on this update. I thought it'd be nice to get the library bump in separately to keep from conflating any issues.

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -8,6 +8,8 @@ pub use crate::keyboard::{KeyCode, KeyModifiers};
 
 #[derive(Debug, PartialOrd, PartialEq, Eq, Clone, Copy, Hash)]
 pub enum Event {
+    FocusGained,
+    FocusLost,
     Key(KeyEvent),
     Mouse(MouseEvent),
     Resize(u16, u16),
@@ -57,6 +59,7 @@ pub enum MouseButton {
 pub struct KeyEvent {
     pub code: KeyCode,
     pub modifiers: KeyModifiers,
+    // TODO: crossterm now supports kind & state if terminal supports kitty's extended protocol
 }
 
 impl KeyEvent {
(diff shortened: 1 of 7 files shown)


helix-view/src/input.rs
```diff
@@ -271,6 +273,11 @@ impl From<crossterm::event::Event> for Event {
             crossterm::event::Event::Key(key) => Self::Key(key.into()),
             crossterm::event::Event::Mouse(mouse) => Self::Mouse(mouse.into()),
             crossterm::event::Event::Resize(w, h) => Self::Resize(w, h),
+            crossterm::event::Event::FocusGained => Self::FocusGained,
```
groves, line 276: Focus events are emitted on Windows regardless of any flags being set in crossterm. Don't want to use `unreachable` like in the Paste arm as a result.

helix-view/src/input.rs
```diff
@@ -321,7 +328,14 @@ impl From<crossterm::event::MouseButton> for MouseButton {
 
 #[cfg(feature = "term")]
 impl From<crossterm::event::KeyEvent> for KeyEvent {
-    fn from(crossterm::event::KeyEvent { code, modifiers }: crossterm::event::KeyEvent) -> Self {
+    fn from(
+        crossterm::event::KeyEvent {
+            code,
+            modifiers,
+            kind: _,
+            state: _,
```
the-mikedavis, line 336: these can be discarded a bit more tersely with

```rs
crossterm::event::KeyEvent { code, modifiers, .. }
```

archseer:
Resolved the comment and added a note about the extra state now offered by crossterm

sfarq:
I am using Kitty, does this update mean I can now use some variation of `C-backspace = "delete_word_backward"` and solve https://github.com/helix-editor/helix/issues/1085 ?

the-mikedavis:
I see `C-backspace` being parsed correctly by crossterm with this update on kitty so it should work on kitty at least. I'm not knowledgable about what changed in crossterm in this update so I'm not sure the fix is expected to work on all terminal emulators.

groves:
We might need to enable some of the [keyboard enhancement flags](https://github.com/crossterm-rs/crossterm/blob/10d1dc246dcd708b4902d53a542f732cba32ce99/src/event.rs#L317) to pick up on previously unsupported key combos. I haven't looked into what happens on other terms when emitting those flags, so I didn't want to just enable it here.

David-Else:
@groves Would you be able to add them? It would be much appreciated!

Issue #3134, opened 2022-07-21
Simple :sh command substitutions


aikomastboom:
It would be nice to be able to reference editor information on the `:sh` line.

Things like for **example**:

|subs|description|
|----:|:---|
| $fn | filename (= $bn.$ex) |
| $bn | basename |
| $ex | extension |
| $fp | full path |
| $rp | relative path (to .git root?) |
| $ln | line number |

to be able to execute things like: 

`:sh gh browse $fn:$ln`

Of course above example is just a starting point to get the idea across.

the-mikedavis:
It's a bit of a nit but I'd prefer more explicit names like `$file`, `$line` rather than two-character shortcuts. They're not very discoverable and are only marginally easier to type

aikomastboom:
Indeed, whole words as tokens is indeed nicer, especially when combined with completion to speed up the typing.

Also it occurred to me the `$` prefix may not be such a great idea, as it commonly used by `sh` itself. What about `#` as a tokenizer?
- `#line`
- `#filename`
Etc. (TBD)

the-mikedavis:
I imagine this would be implemented by setting environment variables for the shell command and then the shell command would interpret them - so `$var` for unix-like and `%var%` or whatever it is for windows - rather than a search-and-replace or templating language rendered ahead of time.

aikomastboom:
Ah, yes, of course we have full control over the env space at execution time.

Still would like to have some help typing the longer vars though ;)

AceofSpades5757:
I think longer variable names also follows Helix' way of doing things. I imagine that aliases would be implemented in the future with more configuration options.

Instead of environment variables, configuration variables could be used as well, for the `project/.helix` directory.

aikomastboom:
Revised table:

|env | description|
|----:|:---|
| $HX_FILENAME | filename |
| $HX_BASENAME | basename |
| $HX_FILE_EXT | file extension |
| $HX_PATH | full absolute path |
| $HX_RELATIVE_PATH | relative path (to .git root? or $CWD) |
| $HX_LINE| line number |
| $HX_COLUMN | cursor pos (on line) |

the-mikedavis:
I think we should start small: we probably only need file, line and column. The others can be derived in the shell with `basename`/`extname`/`dirname`/`pwd`/etc (I'm sure windows has equivalents, right? 🙃)

aikomastboom:
[...]

PR #3393, opened 2022-08-11
command expansions

QiBaobin:
#3134


QiBaobin:
only support filename for now:
`bc $val{filename}`

sbromberger:
Just curious: does this quote the filename so that if you pass it to `sh` it doesn't fall victim to the IFS boundary?

QiBaobin:
> Just curious: does this quote the filename so that if you pass it to `sh` it doesn't fall victim to the IFS boundary?

I think we can use `sh echo "%val{filename}"`

trink:
FYI: This doesn't handle substitution in the keymap arguments.
```toml
[keys.normal]
"C-s" = [":sh hx_rcs_write.sh '%val{filename}'", ":w"] # version control the previous and write out the new
```

helix-term/src/commands/typed.rs
```diff
@@ -2075,8 +2075,9 @@ pub fn command_mode(cx: &mut Context) {
             // Handle typable commands
             if let Some(cmd) = typed::TYPABLE_COMMAND_MAP.get(parts[0]) {
                 let args = shellwords::shellwords(input);
+                let args = expand_args(cx.editor, &args[1..]);
```
the-mikedavis, line 2078: Should we expand arguments before determining shell words? That would eliminate the need to escaping spaces in args like `%sh{echo\ foo}`. What does kakoune do about this?

QiBaobin: done

QiBaobin:
> ```toml
> "C-s" = [":sh hx_rcs_write.sh '%val{filename}'", ":w"] 
> ```

thanks, keymap uses a different function, will look into it later.

QiBaobin:
support %val {filename} %val {dirname} and %sh {cmd} now, we can also nest them like %sh { echo %val {dirname} }, also addressed the 'have to escape space' issue as @the-mikedavis  suggested.

book/src/generated/typable-cmd.md
```diff
@@ -71,3 +71,4 @@
 | `:append-output` | Run shell command, appending output after each selection. |
 | `:pipe` | Pipe each selection to the shell command. |
 | `:run-shell-command`, `:sh` | Run a shell command |
+| `:commands`, `:cmds` | Run commands together, use && to sepearte them |
```
the-mikedavis, line 74:
```suggestion
| `:commands`, `:cmds` | Run commands together, use && to separate them |
```

the-mikedavis: I like the variable substitution part of this PR but I think the command chaining part here should be split to a separate PR. I'm not sure it makes sense to add since you can already chain multiple typable commands in config. Also, `&&` would imply to me that one typable command must succeed in order for the next to be run so we would probably also need `||` and `;` separators for different behaviors matching shell. And if we're matching shell then I think this should wait for scripting support anyways.

QiBaobin: done

helix-term/src/commands/typed.rs
```diff
@@ -1822,6 +1822,42 @@ fn run_shell_command(
...
+        let shellwords = shellwords::Shellwords::from(input.as_ref());
+        let args = shellwords.words();
+
+        if let Err(e) = (cmd.fun)(cx, &args[1..], event) {
+            cx.editor.set_error(format!("{}", e));
+            return Err(e);
+        }
+    } else if event == PromptEvent::Validate {
+        cx.editor
+            .set_error(format!("no such command: '{}'", parts[0]));
```
pickfire, line 1856: Shouldn't this return an error too?

helix-term/src/commands/typed.rs
```diff
@@ -1822,6 +1822,42 @@ fn run_shell_command(
...
+    }
```
pickfire, line 1857: Shouldn't this return an error too?

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,58 @@ pub(super) fn command_mode(cx: &mut Context) {
     prompt.recalculate_completion(cx.editor);
     cx.push_layer(Box::new(prompt));
 }
+
+fn expand_args<'a>(editor: &mut Editor, args: &'a str) -> Cow<'a, str> {
+    let reg = Regex::new(r"%(\w+)\s*\{(.*)").unwrap();
+    reg.replace(args, |caps: &regex::Captures| {
```
pickfire, line 2478: Shouldn't `replace_all` be used instead? Should we only replace one variable?

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,58 @@ pub(super) fn command_mode(cx: &mut Context) {
     prompt.recalculate_completion(cx.editor);
     cx.push_layer(Box::new(prompt));
 }
+
+fn expand_args<'a>(editor: &mut Editor, args: &'a str) -> Cow<'a, str> {
+    let reg = Regex::new(r"%(\w+)\s*\{(.*)").unwrap();
```
pickfire, line 2477: Don't we need to check for the last `}` too? I think this function should have some unit tests to make sure stuff works as expected.

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,58 @@ pub(super) fn command_mode(cx: &mut Context) {
...
+            },
+            next
+        )
+    })
+}
+
+fn find_first_open_right_braces(str: &str) -> usize {
+    let mut left_count = 1;
+    for (i, &b) in str.as_bytes().iter().enumerate() {
+        match char::from_u32(b as u32) {
```
pickfire, line 2515: Shouldn't we iterate over char directly instead? I think unicode might crash in this case when we are reading bytes within multiple bytes char.

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,58 @@ pub(super) fn command_mode(cx: &mut Context) {
     prompt.recalculate_completion(cx.editor);
     cx.push_layer(Box::new(prompt));
 }
+
+fn expand_args<'a>(editor: &mut Editor, args: &'a str) -> Cow<'a, str> {
+    let reg = Regex::new(r"%(\w+)\s*\{(.*)").unwrap();
+    reg.replace(args, |caps: &regex::Captures| {
+        let remaining = &caps[2];
+        let end = find_first_open_right_braces(remaining);
```
pickfire, line 2480: Should we auto-complete for users? I think `%sh{echo hello` shouldn't work.

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,58 @@ pub(super) fn command_mode(cx: &mut Context) {
...
+                    if let Ok((output, _)) = shell_impl(shell, &exp, None) {
+                        output.trim().into()
+                    } else {
+                        "".into()
+                    }
+                }
+                _ => "".into(),
+            },
+            next
+        )
```
pickfire, line 2508:
```suggestion
        let doc = doc!(editor);
        let rep = match &caps[1] {
            "val" => match exp.trim() {
                "filename" => doc.path().and_then(|p| p.to_str()).unwrap_or("").to_owned(),
                "dirname" => doc
                    .path()
                    .and_then(|p| p.parent())
                    .and_then(|p| p.to_str())
                    .unwrap_or("")
                    .to_owned(),
                _ => "".into(),
            },
            "sh" => {
                let shell = &editor.config().shell;
                if let Ok((output, _)) = shell_impl(shell, &exp, None) {
                    output.trim().into()
                } else {
                    "".into()
                }
            }
            _ => "".into(),
        };
        let next = expand_args(editor, remaining.get(end + 1..).unwrap_or(""));
        format!("{rep} {next}")
```
I think we can rearrange this a little.

helix-term/src/commands/typed.rs
QiBaobin: the `b` is a byte so that this won't crash.As we only use ascii chars in our patterns below, I didn't treat them as Unicode char, this shall work until we introduce other  Unicode pattern beside `%val` etc.

helix-term/src/commands/typed.rs
QiBaobin: yes, above command won't expand and will be treated literately . The auto-complete is a very nice to have feature, how about we adding it after this PR.

helix-term/src/commands/typed.rs
```diff
@@ -2460,3 +2472,55 @@ pub(super) fn command_mode(cx: &mut Context) {
...
+                if let Ok((output, _)) = shell_impl(shell, &exp, None) {
+                    output.trim().into()
+                } else {
+                    "".into()
+                }
+            }
+            _ => "".into(),
+        };
+        let next = expand_args(editor, remaining.get(end + 1..).unwrap_or(""));
+        format!("{rep} {next}")
```
trink, line 2505: The added space between rep and next is a problem when you want to compose a single argument with the substitution. e.g.
`:sh echo %val{filename},v`
This currently produces two arguments "*filename*", ",v" instead of the desired "*filename*,v"

helix-term/src/commands/typed.rs
trink: The returned index will always be in bounds (the enumerator and len are both byte based).  Trying something like the following doesn't cause any issues
`sh echo %val{𒀀𒀀𒀀𒀀𒀀𒀀𒀀}` // 4 byte char f0928080

I think this should remain as is.

However, downstream I am seeing something off in shellwords (some multi-byte characters  are lost in the generated argument list, I am investigating it now but it is unrelated to this PR)

trink (changes requested):

helix-term/src/commands/typed.rs
QiBaobin: Good point, will fix it once I have time.

danillos:
Thanks, @QiBaobin I really liked this feature, with that I could build:

```
F8 = ':open %sh { ~/.dotfiles/support/switch_to_spec.sh %val{filename} }'
F10 = ':sh ~/.dotfiles/support/run_current_file_spec.sh %val{filename}'
```

These scripts that I created are not perfect because I had to hard-code how to get the project dirname. 

Could we have something like `%val{project_dirname}` expand to the absolute folder path used in `hx .` ?

cammm:
As someone who works with perforce which requires you to checkout files to make them writable before you can save them, this feature (which I am testing out locally) is great for making a shortcut that handles that on save. Thanks for putting this together.

danillos:
Hey @QiBaobin, I did a change of your code fixing some bugs and adding more variables.
It adds `%val{root_dirname}`, `%val{relative_filename}` and `%val{line_number}`

I have a question do you have time to keep working on this feature? I saw that it wasn't rebased for a while.

```rust
fn expand_args<'a>(editor: &mut Editor, args: &'a str) -> Cow<'a, str> {

    let re_val = Regex::new(r"%val\s*?\{+(?P<var_name>\w+)\}").unwrap();
    let re_sh = Regex::new(r"%sh+\s*?\{+(.*+)\}").unwrap();

    let cmd = re_val.replace_all(args, |caps: &regex::Captures| {
        let root_dirname = PathBuf::from(".")
            .canonicalize()
            .unwrap()
            .to_string_lossy()
            .to_string()
            .to_owned();

        let file_name = doc.path().and_then(|p| p.to_str()).unwrap_or("").to_owned();

        let mut pp: String = root_dirname.clone();
        pp.push('/');
        let relative_file_name = file_name.replace(&pp, "");

        let line_number = doc.selection(view.id).primary().cursor_line(text) + 1;

        match caps[1].trim() {
            // ex. /Users/{user}/my_project/src/file.rs
            "filename" => file_name,
            // ex. /Users/{user}/my_project
            "root_dirname" => root_dirname,
            // ex. src/file.rs
            "relative_filename" => relative_file_name,
            // ex. 1
            "line_number" => line_number.to_string().to_owned(),
            // ex. /Users/{user}/my_project/src
            "dirname" => doc
                .path()
                .and_then(|p| p.parent())
                .and_then(|p| p.to_str())
                .unwrap_or("")
                .to_owned(),
            _ => caps[0].to_owned(),
        }
    });

    let cmdd = re_sh.replace_all(&cmd, |caps: &regex::Captures| -> String {
        let shell_command = caps[1].trim();
        let shell = &editor.config().shell;
        if let Ok((output, _)) = shell_impl(shell, shell_command, None) {
            output.trim().into()
        } else {
            "".into()
        }
    });

    Cow::Owned(cmdd.to_string())
}
```

QiBaobin:
> Hey @QiBaobin, I did a change of your code fixing some bugs and adding more variables. It adds `%val{root_dirname}`, `%val{relative_filename}` and `%val{line_number}`
> 
> I have a question do you have time to keep working on this feature? I saw that it wasn't rebased for a while.
> 
> ```rust
> fn expand_args<'a>(editor: &mut Editor, args: &'a str) -> Cow<'a, str> {
>     let (view, doc) = current!(editor);
>     let text = doc.text().slice(..);
> 
>     let re_val = Regex::new(r"%val\s*?\{+(?P<var_name>\w+)\}").unwrap();
>     let re_sh = Regex::new(r"%sh+\s*?\{+(.*+)\}").unwrap();
> 
>     let cmd = re_val.replace_all(args, |caps: &regex::Captures| {
>         let root_dirname = PathBuf::from(".")
>             .canonicalize()
>             .unwrap()
>             .to_string_lossy()
>             .to_string()
>             .to_owned();
> 
>         let file_name = doc.path().and_then(|p| p.to_str()).unwrap_or("").to_owned();
> 
>         let mut pp: String = root_dirname.clone();
>         pp.push('/');
>         let relative_file_name = file_name.replace(&pp, "");
> 
>         let line_number = doc.selection(view.id).primary().cursor_line(text) + 1;
> 
>         match caps[1].trim() {
>             // ex. /Users/{user}/my_project/src/file.rs
>             "filename" => file_name,
>             // ex. /Users/{user}/my_project
>             "root_dirname" => root_dirname,
>             // ex. src/file.rs
>             "relative_filename" => relative_file_name,
>             // ex. 1
>             "line_number" => line_number.to_string().to_owned(),
>             // ex. /Users/{user}/my_project/src
>             "dirname" => doc
>                 .path()
>                 .and_then(|p| p.parent())
>                 .and_then(|p| p.to_str())
>                 .unwrap_or("")
>                 .to_owned(),
>             _ => caps[0].to_owned(),
>         }
>     });
> 
>     let cmdd = re_sh.replace_all(&cmd, |caps: &regex::Captures| -> String {
>         let shell_command = caps[1].trim();
>         let shell = &editor.config().shell;
>         if let Ok((output, _)) = shell_impl(shell, shell_command, None) {
>             output.trim().into()
>         } else {
>             "".into()
>         }
>     });
> 
>     Cow::Owned(cmdd.to_string())
> }
> ```
> 
> _I'm not a Rust dev, but I'm trying to learn it._

Actually, I don't. It would be great if someone can pick up from here. The reason I use replace instead of replace_all is that I wan to support nested expression, not sure if we shall do that.

ksdrar:
Hello, I changed the variables and command syntaxes to `#{name}` for variables and `#name [body]` for commands. It makes parser easier and I think writing them is easier.

Some simple tests:
```
Testing: #{filename} -----> filename
Testing: #{filename} #{parent}\new_filename -----> filename parent\new_filename
Testing: #sh [echo #{filename}] -----> sh echo filename
Testing: #sh [echo #{filename},#{filename}] -----> sh echo filename,filename
Testing: #sh [echo #{filename},#{filename},(#sh [echo #{filename}])] -----> sh echo filename,filename,(sh echo filename
```

Here's a short video of them in action:

https://user-images.githubusercontent.com/22417151/230752179-818ca2aa-1123-4b34-8eb0-05de9020e4c7.mp4

trink:
> Hello, I changed the variables and command syntaxes to `#{name}` for variables and `#name [body]` for commands. It makes parser easier and I think writing them is easier.

The intent of the %nnn{} type prefix is to serve as a namespace for things like arguments, registers, configuration options and values to provide a bit more structure, communicate intent, and reduce potential name collisions.  The original design is more extensible.

edrex:
> `%val{line_number}`

If this were done, it would be nice if it could be generalized somehow to support selections and multiple selections. It doesn't do much for a shell command though if the buffer is dirty. Thinking about use-cases, I think the existing `shell_pipe` etc commands, powered up by command expansions, will serve many. Also, I'm not an LSP expert (yet) but I wonder what could be done with some kind of user-configurable LSP combinator proxy that does things with selections. This is all out of scope for this issue, but relevant to thinking about whether to include `line_number`.

> `%val{root_dirname}`

I'm for orthogonality. If two vars are equivalent, or one can be derived from the other, I'd prefer only one was included. Functions (either implemented as shell commands, or somehow within helix) from a file path to project path and to project-relative file path seem more generally useful than putting pre-computed derived vals in the list. But maybe I'm not looking at it right.

danillos:
@edrex about `%val{line_number}` I'll give some examples on how I'm using it.

1. To git blame the current line. So I can see who did the changed and when.
```
g = [':sh ~/.dotfiles/commands.sh git_blame %val{relative_filename}:%val{linenumber}']
```

CleanShot 2023-04-12 at 20 04 57@2x


2. Open current line in Github
```
G = [':sh ~/.dotfiles/commands.sh open_file_in_github %val{relative_filename}:%val{linenumber}']
```

3. Run Rspec Test on a specific line number
```
B = ':sh ~/.dotfiles/commands.sh run_rspec_on_line %val{relative_filename}:%val{linenumber}'
```

Without it I would not be able to daily drive in Helix.

edrex:
@danillos are you volunteering to pick this up and take it over the finish line? It seems like @QiBaobin is open to it. If so, you could push up a branch on top of this one (but please preserve existing history for clarity) and open a new PR, and then you'll have an extra expert review in @QiBaobin. Please correct me if I'm misreading intent here.

edrex:
> It doesn't do much for a shell command though if the buffer is dirty

I guess we can trust users to make sure write out first in their `%val{linenumber}`-using commands

danillos:
> @danillos are you volunteering to pick this up and take it over the finish line? It seems like @QiBaobin is open to it. If so, you could push up a branch on top of this one (but please preserve existing history for clarity) and open a new PR, and then you'll have an extra expert review in @QiBaobin. Please correct me if I'm misreading intent here.

Certainly, I was planning to do it. This pull request is essential for my development flow.

ksdrar:
Hey guys, I got the nested commands working with %key{body}

You can try the changes in the branch cmd-expansions of my fork.

The output of the tests:
```
❯ cargo run --release
   Compiling pattern_matching v0.1.0 (/home/jesus/Development/rust/pattern_matching)
    Finished release [optimized] target(s) in 0.54s
     Running `target/release/pattern_matching`
[TEST] %val {filedir} => filedir
[PASSED 821.124µs] filedir == filedir

[TEST] %val {filedir%val {aasdfasdf}} => filediraasdfasdf
[PASSED 738.772µs] filediraasdfasdf == filediraasdfasdf

[TEST] %sh {rm %val{filedir},%val{filename}} => rm filedir,filename
[PASSED 647.805µs] rm filedir,filename == rm filedir,filename

[TEST] %sh {git blame %val{filename}:%val{linenumber}} => git blame filename:linenumber
[PASSED 666.177µs] git blame filename:linenumber == git blame filename:linenumber

[TEST] :sh (%sh {cat %val{filename} | grep rust}) > %val{filedir}/test.txt => :sh (cat filename | grep rust) > filedir/test.txt
[PASSED 660.74µs] :sh (cat filename | grep rust) > filedir/test.txt == :sh (cat filename | grep rust) > filedir/test.txt

[TEST] :sh bash -c (%sh {(%sh {cat %val{filedir}/test.txt}) > %val{filedir}/../%val{filename}_test.txt}) => :sh bash -c ((cat filedir/test.txt) > filedir/../filename_test.txt)
[PASSED 736.72µs] :sh bash -c ((cat filedir/test.txt) > filedir/../filename_test.txt) == :sh bash -c ((cat filedir/test.txt) > filedir/../filename_test.txt)
```

The bare-bones code:
```rust
fn test(to_test: &str, expected: &str) {
    let start = std::time::Instant::now();
    println!("\x1b[33m[TEST]\x1b[0m {to_test} => {expected}");
    let result = find_and_replace(to_test).unwrap_or_default();
    assert_eq!(result, expected);
    println!(
        "\x1b[32m[PASSED {:2?}]\x1b[0m {result} == {expected}\n",
        start.elapsed()
    );
}

fn main() {
    test("%val {filedir}", "filedir");

    test("%val {filedir%val {aasdfasdf}}", "filediraasdfasdf");

    test(
        "%sh {rm %val{filedir},%val{filename}}",
        "rm filedir,filename",
    );

    test(
        "%sh {git blame %val{filename}:%val{linenumber}}",
        "git blame filename:linenumber",
    );

    test(
        ":sh (%sh {cat %val{filename} | grep rust}) > %val{filedir}/test.txt",
        ":sh (cat filename | grep rust) > filedir/test.txt",
    );

    test(
        ":sh bash -c (%sh {(%sh {cat %val{filedir}/test.txt}) > %val{filedir}/../%val{filename}_test.txt})",
        ":sh bash -c ((cat filedir/test.txt) > filedir/../filename_test.txt)",
    );
}

fn find_and_replace(input: &str) -> anyhow::Result<String> {
    let regexp = regex::Regex::new(r"%(\w+)\s*\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap();

    replace_all(&regexp, input, |captures| {
        let keyword = captures.get(1).unwrap().as_str();
        let body = captures.get(2).unwrap().as_str();

        match keyword {
            "val" => Ok(String::from(body)),
            "sh" => Ok(String::from(body)),
            _ => anyhow::bail!("Unknown keyword {keyword}"),
        }
    })
}

fn replace_all(
    regex: &regex::Regex,
    text: &str,
    matcher: impl Fn(&regex::Captures) -> anyhow::Result<String>,
) -> anyhow::Result<String> {
    let mut it = regex.captures_iter(text).peekable();

    if it.peek().is_none() {
        return Ok(String::from(text));
    }

    let mut new = String::with_capacity(text.len());
    let mut last_match = 0;

    for cap in it {
        let m = cap.get(0).unwrap();
        new.push_str(&text[last_match..m.start()]);

        let replace = matcher(&cap)?;

        new.push_str(&replace);

        last_match = m.end();
    }

    new.push_str(&text[last_match..]);

    replace_all(regex, &new, matcher)
}
```

danillos:
I’ve reviewed your [branch](https://github.com/helix-editor/helix/compare/master...ksdrar:helix:cmd-expansions), @ksdrar, and it worked perfectly. In my opinion, you can submit a Pull Request to replace the current one. I believe @QiBaobin will close this current PR once yours is opened.

Nice work.

danillos:
@ksdrar I found an issue. I was migrating all keybindings that I did using my version to your version and this one didn't work.

```
# before
s = ':open %sh { ~/.dotfiles/helix/command.rb switch_to_spec %val{projectdir} %val{relative_filename}}'
```

```
# after
s = ':open %sh {~/.dotfiles/helix/command.rb switch_to_spec %sh{pwd} %val{filename}}'
```

It is not expanding `%sh{pwd}` and `%val{filename}`

ksdrar:
@danillos Can you try this:

```diff
fn find_and_replace(input: &str) -> anyhow::Result<String> {
    let regexp = regex::Regex::new(r"%(\w+)\s*\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap();

    replace_all(&regexp, input, |captures| {
        let keyword = captures.get(1).unwrap().as_str();
        let body = captures.get(2).unwrap().as_str();

        match keyword {
            "val" => Ok(String::from(body)),
-            "sh" => Ok(String::from(body)),
+            "sh" => find_and_replace(body),
            _ => anyhow::bail!("Unknown keyword {keyword}"),
        }
    })
}
```

danillos:
I think it should expand the most inner expressions first, and move to the next one only when there is nothing left to expand. For example:

```
filename = test.rs
line_number = 123
```

**Command:**
`%sh { add_sufix.sh %sh { add_line_number.sh %sh { add_prefix.sh %val{filename} } } }`

**Expansion flow:**
```
0. %sh { add_sufix.sh %sh { add_line_number.sh %sh { add_prefix.sh %val{filename} } } }
1. %sh { add_sufix.sh %sh { add_line_number.sh %sh { add_prefix.sh test.rs } } }
2. %sh { add_sufix.sh %sh { add_line_number.sh prefix_test.rs } }
3. %sh { add_sufix.sh prefix_test.rs:%val{line_number} }
4. %sh { add_sufix.sh prefix_test.rs:123 }
5. prefix_test_sufix.rs:123
```


**Scripts:**
```
add_prefix.sh: adds `prefix_{string}` to a string
add_sufix.sh: adds `{string}_sufix` to a string
add_line_number.sh: adds `{string}:%val{line_number}` to a string
```

ksdrar:
@danillos Can you check if it's working now with the last commit?

danillos:
> @danillos Can you check if it's working now with the last commit?

Yes, this examples is now working.

```
s = ':open %sh {~/.dotfiles/helix/command.rb switch_to_spec %sh{pwd} %val{filename}}'
```

danillos:
@ksdrar could you open your changes as a new PR for replace this one? I could do that too.

robinmoussu:
@Septias https://github.com/helix-editor/helix/pull/11164

PR #3396, opened 2022-08-11
feat: shorten output for grammar fetching and building

poliorcetics:
feat: shorten output for grammar fetching and building

New look:

```
Fetching 102 grammars
98 up to date git grammars
4 updated grammars
        bash      now on 275effdfc0edce774acf7d481f9ea195c6c403cd
        beancount now on 4cbd1f09cd07c1f1fabf867c2cf354f9da53cc4c
        c         now on f05e279aedde06a25801c3f2b2cc8ac17fac52ae
        c-sharp   now on 53a65a908167d6556e1fcdb67f1ee62aac101dda
```

```
Building 102 grammars
100 grammars already built
2 grammars built now
        ["bash", "rust"]
```

### DIFF
--- a/helix-loader/src/grammar.rs
+++ b/helix-loader/src/grammar.rs
@@ -89,15 +89,102 @@ pub fn fetch_grammars() -> Result<()> {
     let mut grammars = get_grammar_configs()?;
     grammars.retain(|grammar| !matches!(grammar.source, GrammarSource::Local { .. }));
 
-    run_parallel(grammars, fetch_grammar, "fetch")
+    println!("Fetching {} grammars", grammars.len());
+    let results = run_parallel(grammars, fetch_grammar);
+
+    let mut errors = Vec::new();
+    let mut git_updated = Vec::new();
+    let mut git_up_to_date = 0;
+    let mut non_git = Vec::new();
+
+    for res in results {
+        match res {
+            Ok(FetchStatus::GitUpToDate) => git_up_to_date += 1,
+            Ok(FetchStatus::GitUpdated {
+                grammar_id,
+                revision,
+            }) => git_updated.push((grammar_id, revision)),
+            Ok(FetchStatus::NonGit { grammar_id }) => non_git.push(grammar_id),
+            Err(e) => errors.push(e),
+        }
+    }
+
+    non_git.sort_unstable();
+    git_updated.sort_unstable_by(|a, b| a.0.cmp(&b.0));
+
+    if git_up_to_date != 0 {
+        println!("{} up to date git grammars", git_up_to_date);
+    }
+
+    if !non_git.is_empty() {
+        println!("{} non git grammars", non_git.len());
+        println!("\t{:?}", non_git);
+    }
+
+    if !git_updated.is_empty() {
+        println!("{} updated grammars", git_updated.len());
+        // We checked the vec is not empty, unwrapping will not panic
+        let longest_id = git_updated.iter().map(|x| x.0.len()).max().unwrap();
+        for (id, rev) in git_updated {
+            println!(
+                "\t{id:width$} now on {rev}",
+                id = id,
+                width = longest_id,
+                rev = rev
+            );
+        }
+    }
+
+    if !errors.is_empty() {
+        let len = errors.len();
+        println!("{} grammars failed to fetch", len);
+        for (i, error) in errors.into_iter().enumerate() {
+            println!("\tFailure {}/{}: {}", i, len, error);
+        }
+    }
+
+    Ok(())
 }
 
 pub fn build_grammars(target: Option<String>) -> Result<()> {
-    run_parallel(
-        get_grammar_configs()?,
-        move |grammar| build_grammar(grammar, target.as_deref()),
-        "build",
-    )
+    let grammars = get_grammar_configs()?;
+    println!("Building {} grammars", grammars.len());
+    let results = run_parallel(grammars, move |grammar| {
+        build_grammar(grammar, target.as_deref())
+    });
+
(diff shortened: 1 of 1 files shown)


poliorcetics:
The output is entirely modifiable, I just find the current one horrid to work with: it's not sorted alphabetically, it's hard to see what was updated and was not, what's not a Git grammar and what is, errors can happen at anytime.

helix-loader/src/grammar.rs
```diff
@@ -189,16 +288,27 @@ fn fetch_grammar(grammar: GrammarConfiguration) -> Result<()> {
...
+            //     "Grammar '{}' checked out at '{}'.",
+            //     grammar.grammar_id, revision
+            // );
         } else {
-            println!("Grammar '{}' is already up to date.", grammar.grammar_id);
+            Ok(FetchSuccess {
+                grammar_id: grammar.grammar_id,
+                action: FetchAction::GitUpToDate,
+            })
+            // println!("Grammar '{}' is already up to date.", grammar.grammar_id);
```
the-mikedavis, line 304: These comments look like they can go 🧹

helix-loader/src/grammar.rs
```diff
@@ -89,15 +89,106 @@ pub fn fetch_grammars() -> Result<()> {
     let mut grammars = get_grammar_configs()?;
     grammars.retain(|grammar| !matches!(grammar.source, GrammarSource::Local { .. }));
 
-    run_parallel(grammars, fetch_grammar, "fetch")
+    print!("Fetching {} grammars", grammars.len());
```
the-mikedavis, line 92: `print!` doesn't flush stdout so this doesn't end up printing until the " ... done" later. I think it would be ok to turn this into a `println!` and remove the done part.

helix-loader/src/grammar.rs
```diff
@@ -89,15 +89,106 @@ pub fn fetch_grammars() -> Result<()> {
...
+    for res in results {
+        match res {
+            Ok(FetchSuccess {
+                grammar_id,
+                action: FetchAction::GitUpdated { revision },
+            }) => git_updated.push(format!("{}->{}", grammar_id, revision)),
+            Ok(FetchSuccess {
+                grammar_id,
+                action: FetchAction::GitUpToDate,
+            }) => git_up_to_date.push(grammar_id),
```
the-mikedavis, line 110: It looks like only `Vec::len` is used so this could be a usize that gets incremented instead

helix-loader/src/grammar.rs
```diff
@@ -89,15 +89,106 @@ pub fn fetch_grammars() -> Result<()> {
...
+        }
+    }
+
+    // People will easily find the grammar they want that way
+    git_updated.sort_unstable();
+    git_up_to_date.sort_unstable();
+    non_git.sort_unstable();
+
+    if !git_up_to_date.is_empty() {
+        println!("{:02} up to date git grammars", git_up_to_date.len());
```
the-mikedavis, line 125: I think I prefer it to not have leading zeros. I don't think alignment of the digits is necessary

helix-loader/src/grammar.rs
```diff
@@ -89,15 +89,106 @@ pub fn fetch_grammars() -> Result<()> {
...
+    }
+    if !non_git.is_empty() {
+        println!("{:02} non git grammars", non_git.len());
+        println!("    {:?}", non_git);
```
the-mikedavis, line 129: I find the Vec Display formatting harder to read than the current line-per-grammar. It's more compact but it's harder to pick a grammar name out if the grammar names aren't aligned. Instead I think a per-line

```
\t* <grammar-name> updated to <new-revision>
```

would look nice, especially if the grammar names and revisions were aligned

helix-loader/src/grammar.rs
```diff
@@ -146,14 +238,21 @@ where
...
-        .find(|result| result.is_err())
-        .map(|err| err.with_context(|| format!("Failed to {} some grammar(s)", action)))
-        .unwrap_or(Ok(()))
+    rx.iter().collect()
 }
 
-fn fetch_grammar(grammar: GrammarConfiguration) -> Result<()> {
+struct FetchSuccess {
+    grammar_id: String,
+    action: FetchAction,
```
the-mikedavis, line 246: This is more of a `status` than an `action` since it describes what happened in the fetch

the-mikedavis (commented):
I agree, the current fetching/building messages are super verbose. This will be a nice improvement 👍

helix-loader/src/grammar.rs
poliorcetics: Yes, thanks for the name

helix-loader/src/grammar.rs
poliorcetics: Yes, done

helix-loader/src/grammar.rs
poliorcetics: Done

helix-loader/src/grammar.rs
poliorcetics: Ah yes, I originally had them displayed but decided that if it hasn't changed, it's not very useful

helix-loader/src/grammar.rs
poliorcetics: Done

helix-loader/src/grammar.rs
poliorcetics: Done

poliorcetics:
Updated the top comment with the new look after review

archseer (approved):
Thanks! The new output looks great 👍🏻

PR #3397, opened 2022-08-11
highlight closure pipes as brackets

vlmutolo:
Currently closure pipes highlight as operators, even though semantically they should be brackets. This PR adds an explicit highlight for pipes within the `closure_parameters` scope.


the-mikedavis (approved):
Good call, these should be punctuation.bracket 👍

PR #3398, opened 2022-08-12
add penumbra+ theme

vlmutolo:
From Penumbra's [README](https://github.com/nealmckee/penumbra):

> Penumbra is a mathematically balanced colour scheme constructed in a perceptually uniform colour space with base colours inspired by the shades of colour occurring in nature due to the light of the sun and the sky. It cleanly separates the perceptual properties of colours while optimally utilizing the available colour space of typical displays.

This PR adds a new built-in theme to Helix using the Penumbra+ color scheme. The "+" is for "higher-contrast". Similar versions of this theme for the "balanced" (lower-contrast) and "++" (even higher contrast) variants of Penumbra can be created just be modifying the palette at the bottom of the theme according to the values in Penumbra's README.

---

This PR is currently in draft status because there are a few questions I wanted to get feedback on.

- [x] The biggest issue is whether punctuation should remain visually de-emphasized like comments are. This was my addition to the theme and wasn't suggested by Penumbra. Arguably a theme named "Penumbra" shouldn't make too many novel decisions on top of the original colors.
- [ ] Should references be colored like the type or as a modifier on the type? Should they be visually distinct? In this version, they're distinct.
- [ ] Is the yellow for text too distracting? Does it draw too much attention to text? We could try to shuffle around colors so that text is something less obtrusive.
- [ ] I haven't put a ton of thought into languages other than Rust. I briefly considered HTML and Markdown. It would be great if someone could try out their own preferred languages and report back deficiencies. 
- [ ] Should this PR include the higher- and lower-contrast variants of Penumbra (i.e. "balanced" and "++")?

### DIFF
--- a/runtime/themes/penumbra+.toml
+++ b/runtime/themes/penumbra+.toml
@@ -1,0 +1,122 @@
+##########
+# SYNTAX #
+##########
+
+comment = "sky-"
+
+type = "green"
+"type.enum.variant" = "green"
+label = "blue"
+tag = "blue"
+
+constant = "purple"
+"constant.numeric" = "cyan"
+"constant.character" = "purple"
+
+string = "yellow"
+"string.special.symbol" = "cyan"
+"string.special.path" = "cyan"
+
+variable = "sky"
+"variable.builtin" = "red"
+
+keyword = "magenta"
+
+function = "blue"
+"function.macro" = "purple"
+
+punctuation = "sky"
+operator = "magenta"
+namespace = "sky"
+
+"diff.plus" = "green"
+"diff.minus" = "red"
+"diff.delta" = "cyan"
+
+
+##########
+#   UI   #
+##########
+
+hint = "sky-"
+info = "sky"
+warning = "yellow"
+error = "red"
+
+diagnostic = { modifiers = ["underlined"] }
+
+"ui.background" = "sky+"
+"ui.background.separator" = "sky"
+
+"ui.cursor" = { modifiers = ["reversed"] }
+"ui.cursor.match" = { bg = "shade+" }
+
+"ui.cursorline.primary" = { bg = "shade-" }
+# "ui.cursorline.secondary" = { bg = "shade+" }
+
+"ui.linenr" = "sky-"
+"ui.linenr.selected" = "sky"
+
+# Using themed status line:
+"ui.statusline" = { fg = "shade", bg = "sky" }
+"ui.statusline.inactive" = { fg = "sky", bg = "shade+" }
+"ui.statusline.normal" = { fg = "shade", bg = "blue" }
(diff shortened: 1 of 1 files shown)


A-Walrus:
As to point one I think punctuation should be more clearly visible.
One thing I dislike is how it looks with color modes, specifically the contrast between the indicator for `visual` or `insert` mode to the rest of the `statusline`.
I think the `statusline` itself should not be blue, rather only the indicator for normal mode should be blue, with the rest of the `statusline` some form of gray.

Other than that, looks good!

vlmutolo:
> I think punctuation should be more clearly visible.

Yeah I'm enjoying this experiment with deemphasized punctuation, but for the built-in theme we should probably make the conservative choice and do text-colored punctuation. 

> I think the statusline itself should not be blue, rather only the indicator for normal mode should be blue, with the rest of the statusline some form of gray.

This is a good point. I like the blue status line for when color modes are disabled, but with them enabled it definitely makes the color modes harder to see. And for a built-in theme we probably want to err on the side of supporting more configurations.

vlmutolo:
Another issue I just noticed is that I accidentally used "white" for variables instead of "sky". I'll fix it later today.

AceofSpades5757:
This looks _really_ good

vlmutolo:
I think this is ready for review. I reverted the punctuation to be the same color as variabels (white) and made the status bar various shades of grey to avoid too much color when combined with the themed status line modes. I think this is the right choice for Helix's default version of this theme, though there's also a commented-out version for those who don't enable the colored modes.

The rest of the questions can probably go unanswered. It seems good-enough as is. If people want adjustments for other languages they can always file a PR.

vlmutolo:
Here are some random screenshots of Helix code using this theme.






.

the-mikedavis (commented):
Looks like scopes for `ui.cursorline` are missing. `ui.cursorline.primary` is for the primary selection and `ui.cursorline.secondary` for all secondaries. Many themes only highlight the primary one. They're not required and are disabled by default (use `:set cursorline true` to enable) but it's nice to add them to a shared theme like this for anyone who does want to use cursorline, similar to visible whitespace, rulers and indent-guides. Would you like to add some here?

vlmutolo:
> it's nice to add them to a shared theme like this for anyone who does want to use cursorline, similar to visible whitespace, rulers and indent-guides. Would you like to add some here?

Yeah, I'd like to make this a "universal" theme as much as possible. I'll add them.

vlmutolo:
I just realized that there's no underlining for IDE warnings and errors, which I'd like to have. So I put this PR back into draft status briefly. I'll fix the underlining then mark it as ready for review.

the-mikedavis (approved):
Looks nice, thank you! 🎨

the-mikedavis:

PR #3412, opened 2022-08-13
Theme Related Fixes

n0s4:
- Replace "text" with "ui.text" in rose_pine* themes
- Change "attributes" to "attribute" in many included themes
- Document "attribute", "property" and "punctuation.special" in the theme section of the book.
- Remove "variable.function" from the documentation: it doesn't appear to exist in any highlights.scm.
- Remove "path" from snazzy; maybe it should be string.special.path?

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -125,8 +125,6 @@ We use a similar set of scopes as
     - `url`
     - `symbol` - Erlang/Elixir atoms, Ruby symbols, Clojure keywords
 
-- `property` - Regex group names
-
 - `comment` - Code comments
   - `line` - Single line comments (`//`)
   - `block` - Block comments (e.g. (`/*     */`)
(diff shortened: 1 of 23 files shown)


the-mikedavis:
In most cases we translate `property`/`field` to `variable.other.member`. I think the current `@property` capture for regex is an oversight - looking at that now I think it should be `@label`.

For snazzy, yeah I bet that was meant to be `string.special.path`. It may have successfully been matching that scope before #3275 and should be updated now

n0s4:
Ok. I changed `@property` to `@label` in `regex/highlights.scm`. Since that was the only instance of `@property` and every theme that had `property` also included `label`, I've removed `property` from each theme and let `label` cover it.

Also added `string.special.path` back to snazzy with the same colour as `path` had before.

the-mikedavis (approved):

the-mikedavis:
Thanks @n0s4!

PR #3419, opened 2022-08-13
Improve dart indent

seb-bl:
This PR improves the indentation for dart. It is based on a more recent version of the grammar, so it updates it in addition to changing `runtime/queries/dart/indents.scm`

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -923,7 +923,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "dart"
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
This is looking good. I think the "this one is not actually used" commented out nodes can be removed, right?

seb-bl:
Yes, they can be removed (I did in the next commit). They corresponds to declaration in the syntax with braces, but are actually not contained by a node in the program syntax. For example, here <https://github.com/seb-bl/tree-sitter-dart/blob/455556f37ca9f208bdbeb76fffb4e3fd331ebf08/grammar.js#L726> `inferred_parameters` is not used anywhere

the-mikedavis (approved):

the-mikedavis:
Thank you!

PR #3425, opened 2022-08-14
Render html `<code>` tags as code in markdown

### LINKED ISSUE #3411: Improve rendering of `<code>` tags in documentation
Documentation, at least for rust, can contain `<code>` tags, for example the documentation for `into()` from the standard library. The tags themselves should not be shown, and the text within should be shown the same as text within "\`":
image
The `<code>` and `</code>` tags should not be shown and the text within them should be styled the same as the `U::from(self)` from above.

A-Walrus:
Resolves #3411.
Before:
image
After:
image

### DIFF
--- a/helix-term/src/ui/markdown.rs
+++ b/helix-term/src/ui/markdown.rs
@@ -178,6 +178,21 @@ impl Markdown {
             .map(|key| get_theme(key))
             .collect();
 
+        // Transform text in `<code>` blocks into `Event::Code`
+        let mut in_code = false;
+        let parser = parser.filter_map(|event| match event {
+            Event::Html(tag) if *tag == *"<code>" => {
+                in_code = true;
+                None
+            }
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
Hi, thank you!
This change looks quite good to me and seems pretty uncontroversial.
I'll merge it.

A-Walrus:
> EDIT: for the record, we could improve performance by checking for the `<code>` tags directly in the main loop

It's all done with lazy iterator methods, so it's not actually running through everything twice, or collecting anything into a `Vec`. It should be pretty much equivalent to putting it in the main loop.

CBenoit:
> > EDIT: for the record, we could improve performance by checking for the `<code>` tags directly in the main loop
> 
> It's all done with lazy iterator methods, so it's not actually running through everything twice, or collecting anything into a `Vec`. It should be pretty much equivalent to putting it in the main loop.

My bad, you’re right! It’s all good 💯

PR #3444, opened 2022-08-15
Remove border from code actions popup

### LINKED ISSUE #3438: Code action popup uses `ui.popup` and `ui.menu`
This is just a minor nit, but I tried to have a different color for the doc popup (`ui.popup`) next to the completion list (`ui.menu`):

image

This works nicely but then the code action popup looks odd because it nests a menu inside a popup:

image

I think the code action popup could use the same layout as the completion list (i.e. removing the border).

A-Walrus:
Resolves #3438


CBenoit (approved):
Thank you! Looks good to me.

cc @CptPotato

PR #3454, opened 2022-08-16
Add preview for scratch buffers in buffer picker

### LINKED ISSUE #3420: No content preview for unsaved scratch buffers in buffer picker
Hi, 

There's no content preview for unsaved scratch buffers in buffer picker. Is this a bug?

Screen Shot 2022-08-14 at 00 55 52

A-Walrus:
Resolves #3420.

The diff is quite big mostly because a big block in `picker.rs` is now indented more...


the-mikedavis (commented):
I think this approach makes sense. I'm hung up on the new enum's name though - would it be too on-the-nose to call it a `File` instead of `PathOrId`?

A-Walrus:
> would it be too on-the-nose to call it a `File` instead of `PathOrId`?

I feel like `File` is less descriptive, and also more confusing as people might assume it is `std::fs::File` or `tokio::fs::File`.

the-mikedavis (approved):

archseer:
Can you resolve the conflict here? Looks good to merge otherwise

the-mikedavis:
Thanks for working on this!

PR #3455, opened 2022-08-16
Fix failing test with unicode-lines feature

### LINKED ISSUE #3453: unicode-lines tests fail on the master branch
### Summary

The tests on the master branch of helix fail if they are run with the `unicode-lines` feature enabled:

```
failures:

---- line_ending::line_ending_tests::line_end_char_index_rope_slice stdout ----
thread 'line_ending::line_ending_tests::line_end_char_index_rope_slice' panicked at 'assertion failed: `(left == right)`
  left: `5`,
 right: `11`', helix-core/src/line_ending.rs:308:9


failures:
    line_ending::line_ending_tests::line_end_char_index_rope_slice

test result: FAILED. 125 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.20s
```

### Reproduction Steps

```
# clone helix
git clone https://github.com/helix-editor/helix.git
cd helix

# run the tests
cargo test -p helix-core --tests --features unicode-lines
```

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

Alacritty  

### Helix Version

master branch

A-Walrus:
Closes #3453

### DIFF
--- a/helix-core/src/line_ending.rs
+++ b/helix-core/src/line_ending.rs
@@ -305,8 +305,17 @@ mod line_ending_tests {
     fn line_end_char_index_rope_slice() {
         let r = Rope::from_str("Hello\rworld\nhow\r\nare you?");
         let s = &r.slice(..);
-        assert_eq!(line_end_char_index(s, 0), 11);
-        assert_eq!(line_end_char_index(s, 1), 15);
-        assert_eq!(line_end_char_index(s, 2), 25);
+        #[cfg(not(feature = "unicode-lines"))]
+        {
+            assert_eq!(line_end_char_index(s, 0), 11);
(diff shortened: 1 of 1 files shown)


jcornaz:
Would be nice to also update the CI to run tests with all features, so that this kind of problem does not come back.

A-Walrus:
> Would be nice to also update the CI to run tests with all features, so that this kind of problem does not come back.

This is a good idea, but running with all features enabled doesn't solve the problem. Maybe there is a bug when some feature(s) are disabled. The more thorough solution would be to try all possible combinations of feature flags, maybe with something like [cargo-all-features](https://crates.io/crates/cargo-all-features).

PR #3457, opened 2022-08-16
Update gutters on config-reload and :set gutters

### LINKED ISSUE #3445: `config-reload` doesn't apply changes in `editor.gutters` settings
### Summary

I tried to remove "diagnostics" from "gutters" and was surprised that "config-reload" doesn't work for this particular setting.

### Reproduction Steps

[asciicast](https://asciinema.org/a/hNT1TCTC1a1fUZNgpsXI69XoO)

### Helix log

```
2022-08-15T23:44:08.558 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
```

</details>

### Platform

MacOs

### Terminal Emulator

iTerm2

### Helix Version

22.05-369-gcc4fdd54

A-Walrus:
Resolves #3445

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -660,6 +660,9 @@ impl Editor {
     pub fn refresh_config(&mut self) {
         let config = self.config();
         self.auto_pairs = (&config.auto_pairs).into();
+        for view in self.tree.views_mut() {
+            view.0.update_config(&config.gutters);
+        }
         self.reset_idle_timer();
     }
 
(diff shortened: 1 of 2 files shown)


archseer:
Hmm, since each view has it's own set of gutters the idea was you could use `:set` per view to change gutters. Do we have access to the old config value? Then we could only update buffers that have the old default set of gutters.

A-Walrus:
> the idea was you could use `:set` per view to change gutters.

This feels kind of counter intuitive to me. If the `:set` command is defined as setting a config option, it should be equivalent to reloading with a changed config. Having it act on a specific view for only this specific command seems like inconsistent behavior. Maybe there should be a separate command for per-view changes?

pascalkuthe:
closing this one as stale, this will be covered by #8853. Thank you for contributing!

PR #3458, opened 2022-08-16
Allow tab virtual whitespace padding

### LINKED ISSUE #3427: Tab padding character?
Is it possible to set tab padding character? https://docs.helix-editor.com/configuration.html#editorwhitespace-section does not mention it.
By the way what is `nbsp`? Maybe it would be nice to add some comment in the line to explain it.

A-Walrus:
Resolves #3427.
Added new `whitespace.characters.tabpad` option which sets the padding for tabs
Example:
image

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -192,7 +192,7 @@ Options for rendering whitespace with visible characters. Use `:set whitespace.r
 | Key | Description | Default |
 |-----|-------------|---------|
 | `render` | Whether to render whitespace. May either be `"all"` or `"none"`, or a table with sub-keys `space`, `tab`, and `newline`. | `"none"` |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
I think this is a good angle implementation-wise but it may be preferrable to instead have the config break down like

```toml
[editor.whitespace]
render = "all"
characters.tab = { character = "→", padding = " " } # default values
# also accept the old syntax for backwards compatibility and simplicity:
characters.tab = "→"

# or alternatively
[editor.whitespace.characters]
tab = "→"
tabpad = "·" # using kakoune's name but we should probably use something more clear
```

I think this would be preferrable (albeit a larger change) because the characters could be represented by `char`s instead of a `String` - so no allocations - and there isn't a chance of misconfiguring the padding in a confusing way. With the current approach, you could use not enough padding in your configuration and end up being confused why a file with 8-column-width tabs only shows interpuncts for a few of the columns.

the-mikedavis (approved):

PR #3462, opened 2022-08-17
line_search: search lines in current buffer

QiBaobin:


EpocSquadron:
What does this do?

QiBaobin:
> What does this do?

This will search in current buffer, then open a picker to select or query from matched lines.
Screen Shot 2022-08-18 at 09 19 49
Screen Shot 2022-08-18 at 09 20 01

n0s4:
So just global search but in the current buffer?

I guess it has the advantage of giving you a list of results, but this overlaps with the current `/` search which is the same but without the menu and only selects the search match - which is usually more useful, and you could always just hit `x` after.

QiBaobin:
> So just global search but in the current buffer?
> 
> I guess it has the advantage of giving you a list of results, but this overlaps with the current `/` search which is the same but without the menu and only selects the search match - which is usually more useful, and you could always just hit `x` after.

No, the major difference is that global search pickers query file name, while the line search pickers query line content.

n0s4:
By global search I meant `space-/`.

PR #3465, opened 2022-08-17
remove duplicated shell calls

QiBaobin:


archseer:
This seems stop piping to shell after the first successful call? We always want to pipe for each selection.

QiBaobin:
> This seems stop piping to shell after the first successful call? We always want to pipe for each selection.

No, we only cache the result if we don't need pipe the selections as below:

if !pipe {
                        shell_output = Some(result.0.clone());
                    }

the-mikedavis (approved):
Oh I see, when `pipe == false` there is no input to the command so we can assume the output will be the same.

The only downside to this is if the command has random output or output based on the number of times called, like `head -c 1 | base64`. It's a pretty contrived example though, I can't imagine there's a real use-case like that

QiBaobin:
> Oh I see, when `pipe == false` there is no input to the command so we can assume the output will be the same.
> 
> The only downside to this is if the command has random output or output based on the number of times called, like `head -c 1 | base64`. It's a pretty contrived example though, I can't imagine there's a real use-case like that

Yes, you're correct. This only make sense while those command don't have side effect. Let me know if this break any use case, I can close this then. Thanks.

PR #3468, opened 2022-08-17
feat: add taskwarrior to languages so `task N edit` is nice

AlexanderBrevig:
`task N edit`

image

### DIFF
--- a/runtime/queries/task/highlights.scm
+++ b/runtime/queries/task/highlights.scm
@@ -1,4 +1,4 @@
 (comment) @comment
 (datetime) @type
-(identifier) @identifier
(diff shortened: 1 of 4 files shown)


runtime/queries/task/highlights.scm
```diff
@@ -0,0 +1,4 @@
+(comment) @comment
+(datetime) @type
+(identifier) @identifier
```
the-mikedavis, line 3:
```suggestion
(identifier) @variable
```

the-mikedavis (approved):

the-mikedavis:
Thank you!

Issue #2583, opened 2022-05-27
Greater than 99999 line file doesn't resize line number width


dalanmiller:
### Summary

When viewing a file that has more than 99,999 lines, anything more than the left five digits is squelched.

E.g. my `/etc/hosts` file.
<img width="336" alt="Screen Shot 2022-05-27 at 11 19 55 pm" src="https://user-images.githubusercontent.com/863286/170707280-d546035c-a4e8-4111-b2fe-3259601284c6.png">
 
This gets quite confusing when viewing lines much higher:

<img width="260" alt="Screen Shot 2022-05-27 at 9 35 48 pm" src="https://user-images.githubusercontent.com/863286/170708183-92181b4c-9580-4e02-87e1-fa12fd47d0c5.png">

### Reproduction Steps

Download this file:

[100001.txt](https://github.com/helix-editor/helix/files/8786922/100001.txt)

`:100001` and observe the line numbers being truncated. 

### Helix log

Probably not necessary? I didn't see any issues related.

### Platform

macOS

### Terminal Emulator

iterm2

### Helix Version

helix 22.03 (d4e45fd4)

the-mikedavis:
The line number gutter has a fixed, constant width currently: https://github.com/helix-editor/helix/blob/efae76160d19ebba2bba31ecd5c104374c7b4483/helix-view/src/view.rs#L104

Some ways to go about fixing this:

* make the fixed-width configurable instead of a constant
* make the width dynamic based on the line numbers within view or the maximum line number of the buffer

EpocSquadron:
I vote for the second, within buffer. Would save some space in smaller files. However I wonder if we keep the limit but cut off the beginning digits with an ellipsis character to avoid going nuts with space usage in massive files.

archseer:
Resolved by https://github.com/helix-editor/helix/issues/2583

PR #3469, opened 2022-08-18
Dynamically resize line number gutter width

dgkf:
Addressing #2583, this PR implements the second proposed option.

### Behavior

- The line number gutter will resize based on the number of digits in the line count for the Document
- The line number gutter will have a minimum width of 2 characters. This is so that configurations using `line-number: "relative"` aren't adversely affected by this change. In the future, the min/max width could perhaps be configurable and this hard-coded minimum could be removed.
- Enabled the `GutterType::Spacer` option. (I think the default `gutter` config should probably become `["diagnostics", "spacer", "line-number"]` now numbers will butt right up against diagnostics past line 100)

_behaviors that have been modified after review comments_

- ~~If the number of digits on screen exceeds 5 digits, only the last 4 digits are displayed (preceded by `…`)~~ no longer truncates (as suggested in https://github.com/helix-editor/helix/pull/3469#discussion_r948628101)
- ~~Line number gutter width based on _view_ bottom, not _document_ bottom.~~ Fixed

# Screenshots

image
> A very short window to highlight shrinking width (`gutters: ["diagnostics", "spacer", "line-number"]`)

image
> With default config, diagnostic symbols look a bit cramped


helix-term/src/ui/editor.rs
```diff
@@ -638,22 +638,23 @@ impl EditorView {
         // avoid lots of small allocations by reusing a text buffer for each line
         let mut text = String::with_capacity(8);
 
-        for (constructor, width) in view.gutters() {
-            let gutter = constructor(editor, doc, view, theme, is_focused, *width);
-            text.reserve(*width); // ensure there's enough space for the gutter
+        for gutter_type in view.gutters() {
+            let gutter = gutter_type.row_styler(editor, doc, view, theme, is_focused);
+            let width = gutter_type.width(view);
```
dgkf, line 643: `width` is no longer passed to `gutter_type.row_styler` (formerly `constructor`), though I could see the argument for calculating `width` first and providing it to the `row_styler`.

helix-view/src/gutter.rs
```diff
@@ -91,24 +126,41 @@ pub fn line_numbers<'doc>(
...
         }
     })
 }
 
+pub fn line_numbers_width(view: &View) -> usize {
+    // TODO: allow gutter widths to be dependent on Document. Currently the
+    // width is based on full View height, not visible line numbers.
+    let last_view_line = view.offset.row + view.area.bottom() as usize;
+    std::cmp::max(std::cmp::min(count_digits(last_view_line), 5), 3)
+}
```
dgkf, line 156: It would be great to have `Document` available here to base this line number width on `min(<last_document_line>, <last_visible_line>)`, but I quickly ended up with cyclic data flows. Total gutter width is needed for `view.inner_area` is needed for last line is needed for total gutter width.

archseer:
> * The line number gutter will resize based on the number of digits for the line _at the bottom of the view*_

Bottom of the document is what we want though, otherwise the document will jitter sideways as you scroll.

helix-view/src/gutter.rs
```diff
@@ -56,10 +88,13 @@ pub fn line_numbers<'doc>(
     view: &View,
     theme: &Theme,
     is_focused: bool,
-    width: usize,
 ) -> GutterFn<'doc> {
+    const ELLIPSIS: char = '\u{2026}';
+
     let text = doc.text().slice(..);
     let last_line = view.last_line(doc);
+    let width = line_numbers_width(view);
```
archseer, line 96: Wouldn't it be sufficient to call this with `text.len_lines()`, without having to refactor?

helix-view/src/gutter.rs
```diff
@@ -1,21 +1,53 @@
...
 
 use crate::{
+    editor::GutterType,
     graphics::{Color, Modifier, Style},
     Document, Editor, Theme, View,
 };
 
+fn count_digits(n: usize) -> usize {
+    successors(Some(n), |&n| (n >= 10).then(|| n / 10)).count()
+}
```
archseer, line 12: Fun snippet :) I'd fully qualify `std::iter::successors` rather than importing since it's only used in one location

archseer: I'd also add a note that this can be simplified to `n.checked_log10().unwrap_or(0) + 1` if `int_log` gets standardized in stdlib

helix-view/src/gutter.rs
```diff
@@ -91,24 +126,41 @@ pub fn line_numbers<'doc>(
...
             };
-            write!(out, "{:>1$}", display_num, width).unwrap();
+
+            // if line number overflows maximum alotted width, truncate start
+            if n_digits > width {
+                let display_trailing = (display_num as u32) % 10_u32.pow((width - 1) as u32);
+                write!(out, "{}{:0>2$}", ELLIPSIS, display_trailing, width - 1).unwrap();
+            } else {
+                write!(out, "{:>1$}", display_num, width).unwrap();
+            }
```
archseer, line 144: I don't think we need truncation, line numbers should always be fully visible. I can't think of any other editor truncating, past 8 digits the file will be enormously large anyway.

helix-view/src/view.rs
```diff
@@ -144,13 +115,24 @@ impl View {
 
     pub fn inner_area(&self) -> Rect {
         // TODO add abilty to not use cached offset for runtime configurable gutter
-        self.area.clip_left(self.gutter_offset).clip_bottom(1) // -1 for statusline
+        self.area.clip_left(self.gutter_offset()).clip_bottom(1) // -1 for statusline
```
archseer, line 118: This resolves the TODO on the line above

helix-view/src/gutter.rs
dgkf: Wow! You all work fast. Stepped away to pet my cat and had a full review when I got back on a PR in draft. Kudos to you!

I'll qualify the use and add a note that this can be done less opaquely once `int_log` is standardized.

helix-view/src/gutter.rs
dgkf: Alright, that's easy enough. The design was taken from the discussion in #2583 - happy to change it if it's preferred that it never truncates.

helix-view/src/gutter.rs
dgkf: Sure, that works. I was basing it off of the first suggestion here ([2438](https://github.com/helix-editor/helix/issues/2583#issuecomment-1139625428)), but I can do the latter suggestion instead. 

> make the width dynamic based on the line numbers within view or the maximum line number of the buffer

dgkf: I'm having trouble making this fix. 

It's important that the width here is the same as how it's calculated for `GutterType.width` used to calculate `gutter_offset`. If these provide different widths, the gutter and numbers don't stay aligned.

But I couldn't find an elegant way to provide the `Document` to `GutterType::LineNumber.width` in a way that doesn't require some major changes. But I'm still quite new to the code base, so perhaps there's some obvious solution I'm missing.

dgkf: Revisited this and it ended up being less problematic than I first thought. I needed to pass the `Document` to `View.inner_area`. This seemed fine in all but one place, where I think the total `View` area is a fine substitution. I'll highlight it in a separate thread.

helix-view/src/tree.rs
```diff
@@ -482,7 +482,7 @@ impl Tree {
                     // in a vertical container (and already correct based on previous search)
                     child_id = *container.children.iter().min_by_key(|id| {
                         let x = match &self.nodes[**id].content {
-                            Content::View(view) => view.inner_area().left(),
+                            Content::View(view) => view.area.left(),
```
dgkf, line 485: This is the only place where the Document isn't easily available to calculate the inner area (now that the gutter width is based on the Document contents). A simple solution is to use the entire view area instead. 

I think this would affect things like moving to a split when the screen location that you would move to would be in the gutter of a neighboring split. Previously, if you were to try to find a neighboring split at a gutter, you would get the window to the left, now you would get the window that contains the gutter.

helix-term/src/ui/editor.rs
```diff
@@ -736,9 +736,10 @@ impl EditorView {
         // avoid lots of small allocations by reusing a text buffer for each line
         let mut text = String::with_capacity(8);
 
-        for (constructor, width) in view.gutters() {
-            let gutter = constructor(editor, doc, view, theme, is_focused, *width);
-            text.reserve(*width); // ensure there's enough space for the gutter
+        for gutter_type in view.gutters() {
+            let gutter = gutter_type.row_styler(editor, doc, view, theme, is_focused);
```
the-mikedavis, line 740: I don't think this is a general rule in the codebase but for me "row_styler" comes across as https://wiki.c2.com/?DontCreateVerbClasses

I prefer the old name of "constructor" but maybe there's a better word for it: "style" or something similar

helix-view/src/gutter.rs
```diff
@@ -1,21 +1,53 @@
 use std::fmt::Write;
 
 use crate::{
+    editor::GutterType,
     graphics::{Color, Style, UnderlineStyle},
     Document, Editor, Theme, View,
 };
 
+fn count_digits(n: usize) -> usize {
+    // NOTE: if int_log gets standardized in stdlib, can use checked_log10
```
the-mikedavis, line 10: Looks like this entered the FCP the other day https://github.com/rust-lang/rust/issues/70887#issuecomment-1289028216

Could you link the issue in this comment too to raise the chances of it being refactored later?

helix-term/src/ui/editor.rs
dgkf: Thanks for the tip! I agree, this name never felt quite right to me. 

When entering into the code base, I found "constructor" to be a bit of a misnomer. I went with the suggested "style", which I quite like.

the-mikedavis:
I've been testing this for a bit and I think you're right about the diagnostics being too cramped when there aren't many lines in a file. I think changing the default gutters to include a spacer between the diagnostics and line-numbers elements looks much better 👍

dgkf:
Thanks @the-mikedavis - I forgot that was even non-standard. I've been using this with the spacer for so long now. I'll update with the change to the default.

dgkf:
@the-mikedavis, the gutter `spacer` has been added as default (separating the diagnostics from the line numbers). Let me know if you want anything else to change before it's in an acceptable place.

helix-term/src/commands.rs
```diff
@@ -870,7 +870,7 @@ fn goto_window(cx: &mut Context, align: Align) {
     let config = cx.editor.config();
     let (view, doc) = current!(cx.editor);
 
-    let height = view.inner_area().height as usize;
+    let height = view.inner_area(doc).height as usize;
```
archseer, line 873: In almost every circumstance we're only interested in the height which is unaffected by gutters. Let's add a view.height() method that calculates it without having to pass in a doc. Additionally the method should return a usize since we're casting to usize everywhere

helix-view/src/view.rs
```diff
@@ -160,15 +130,25 @@ impl View {
...
     }
 
-    pub fn gutters(&self) -> &[(Gutter, usize)] {
+    pub fn gutters(&self) -> &[GutterType] {
         &self.gutters
     }
 
+    pub fn gutter_offset(&self, doc: &Document) -> u16 {
+        let mut offset = 0;
+        for gutter in self.gutters.iter() {
```
archseer, line 143: This for loop could be replaced with an iter.map().sum()

dgkf: swapped for iterator equivalent

helix-term/src/commands.rs
dgkf: Sounds good - I added a `view.inner_height()` which is just slightly different than a `view.height()` method because it subtracts the statusbar height allowing it to be dropped in for all the places where `inner_area().height` was being used previously.

helix-term/src/commands.rs
```diff
@@ -1399,26 +1399,26 @@ pub fn scroll(cx: &mut Context, offset: usize, direction: Direction) {
...
+    let (view, _doc) = current!(cx.editor);
+    let offset = view.inner_height() / 2;
     scroll(cx, offset, Direction::Backward);
 }
 
 fn half_page_down(cx: &mut Context) {
-    let view = view!(cx.editor);
-    let offset = view.inner_area().height as usize / 2;
+    let (view, _doc) = current!(cx.editor);
+    let offset = view.inner_height() / 2;
```
the-mikedavis, line 1421: These can all be reset to `let view = view!(cx.editor)`

dgkf: Thanks - good catch. Apologies for the sloppiness here. It's been a while since I kicked off the work and the details escape me now.

the-mikedavis (approved):

dgkf:
@archseer - Any chance I could grab your attention on this one? 

Along those lines, I can't tell if I'm being annoying or helpful when I ping you all or re-request review. I know you all have an absolutely enormous PR backlog (IMO a reflection on the community's enthusiasm and not the pace of the maintainers :tada:), and I can't tell what is the best way to grab your attention.

After this PR, would it be helpful to submit a "PR Etiquette" section to the CONTRIBUTING guide to make this relationship less ambiguous? I can draft something up, but you all would probably need to weigh in on how you want things to flow.

archseer:
I think in this case a ping is fine because I was going to merge this once the last change was addressed :)

archseer (approved):
Thanks for working on this!

PR #3486, opened 2022-08-20
add `<` and  `>` to rust auto-pairs 

jcornaz:
The `<` and `>` were missing in the auto-pairs for rust.


jcornaz:
Ah sorry. I realize now why is was not set already defined. Because `<` is also used a lot for comparisons, therefore we cannot always assume we should insert the pair.

Sorry for the noise :-/

PR #3491, opened 2022-08-20
feat(rust): highlight try operator (?) as operator

### LINKED ISSUE #3490: Rust's try operator doesn't highlight as operator
The try operator (`?`) doesn't highlight as an operator. I tried to fix this myself by adding it to the end of the `@operator` list in `runtime/queries/rust/highlights.scm`, but it didn't work. I guess it's a little more complicated than that.

I also saw the `"?" @special` highlight near the end of that file, but I'm not sure what piece of syntax that refers to.

vlmutolo:
Fixes #3490.

I restricted this highlight to the `try_expression` tree-sitter scope. I think this makes sense if we don't want to highlight `?` in other places as an operator, for example in `?Sized`.

### DIFF
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -320,6 +320,8 @@
   "'"
 ] @operator
 
+(try_expression "?" @operator)
+(removed_trait_bound "?" @operator)
 
 
 ; -------
@@ -357,8 +359,6 @@
 ; -------
 ; Remaining Identifiers
 ; -------
-
-"?" @special
 
 (type_identifier) @type
 (identifier) @variable


vlmutolo:
Should I remove `"?" @special` in this PR? Seems like it may not be used if we're classifying `?` in `try_expressions` as an operator.

runtime/queries/rust/highlights.scm
```diff
@@ -320,6 +320,7 @@
   "'"
 ] @operator
 
+((try_expression "?") @operator)
```
the-mikedavis, line 323:
```suggestion
(try_expression "?" @operator)
```

so that the capture applies only to the `?` and not the overall try_expression node

the-mikedavis (commented):
I think the remaining `"?" @special` should be modified: if we remove it we won't have highlighting on `?` in `?Sized`. I'm not sure what scope to use for that `?`

vlmutolo:
> I'm not sure what scope to use for that ?

I think it's called `removed_trait_bound`. So could we just change `"?" @special ` to something like:

```
(removed_trait_bound "?" @operator)
```

I think "operator" is kind of a strange designation for the `?` in `?Sized`, but I also can't find anything better among the preexisting Helix highlights.

the-mikedavis:
Yeah I was thinking that operator was a bit awkward for it. Maybe `@keyword` or `@type` could fit better? We already use `@keyword` for `&` which is sort-of similar in nature

archseer:
I styled `?` as special since it was also colored that way in vim. It is an operator, but it's much different from other operators and it's also very easy to miss. At the very least I'd like us to style this as `@operator.special` and the `special` scope in the default theme should be updated to reflect that. Otherwise this will confuse me when coding :)

archseer:
I'd also style `?` in all of these cases including `?Sized` since it's styled differently in the official rust docs: https://doc.rust-lang.org/std/marker/trait.Sized.html

It's a modifier, not a part of the type name.

pascalkuthe:
closing this as stale, I can't link the exact pr that fixed this but the try operator is definitely highlighted these days. Thank you for contributing!

PR #3492, opened 2022-08-21
Add Nim language support

j-james:


j-james:
Ah, I now see that there have been quite a few previous PRs. Seems I've got to write some queries.

zetashift:
There is also https://github.com/nim-lang/langserver which one might want to use instead of `nimlsp`

mtoohey31:
Would it be possible to merge these changes minus the grammar for now, so we at least have comment, indentation, and language server support out of the box?

j-james:
Thanks for the bump. Rebased on `master` and switched to `nimlangserver`. 

Currently playing around with writing basic highlight / indentation queries based on `alaviss/tree-sitter-nim` - I think eventually `aMOPel/tree-sitter-nim` will be better as it exposes much more of Nim, but it's also more complicated, and I don't want to mess around with it right now while the rule structure is in flux.

mtoohey31:
On `nimlangserver` vs `nimlsp`, it looks like the `nimlsp` project is more active, but that's just based on commits/issues/stars, so take that with a grain of salt. There's a comparison here: https://github.com/nim-lang/langserver/issues/2, but it's not too in-depth. Also, I wrote my first line of nim today, so I'm not really qualified to comment on which works better. However, I was using `nimlsp` earlier today and it looks like it crashes if the file doesn't exist on disk yet...

@zetashift have you used both? Would you be able to offer your opinion on which we should include?

j-james:
I switched to `nimlangserver` after getting fed up with `nimlsp`'s constant crashes and have found it much more mature and stable (fewer crashes, better recovery after crashes, equivalent support via `nimsuggest`, no bug with orphan processes pegging your CPU at 100%)

bfg-coding:
Is there anything that I can help with to get this moving forward?   I'd like to have a good DX with helix and nim

Desdaemon:
Been experimenting with this on my end, some small thoughts:
- The grammar from `aMOPel/tree-sitter-nim` seems to work fine, however type annotations are a bit hard to write highlight queries for.
- Agreed with `nimlangserver` working much better, though at least on my Mac M1 it has a bug where child processes are not cleaned up properly, though luckily was easy to fix/patch.

j-james:
Hi, apologies, I've been pretty busy and will continue to be busy for at least another week or so. I ran into issues getting helix to recognize my `.scm` files whatsoever, that I still need to debug, and left off there.

If any of you would like to work on it in the meantime, feel free - I was modelling the `.scm` files off of the corresponding Rust and Python files, and adding in keywords taken from the manual as necessary - but getting this up and running is definitely on my todo list (just not a high priority right now).

PR #3494, opened 2022-08-21
Add c-sharp textobjects

votagus:
This PR adds basic textobjext based navigation for c-sharp.

### DIFF
--- a/runtime/queries/c-sharp/highlights.scm
+++ b/runtime/queries/c-sharp/highlights.scm
@@ -1,35 +1,5 @@
 ;; Methods
-(method_declaration
-  name: (identifier) @function)
-
-(method_declaration
-  type: [(identifier) (qualified_name)] @type)
-
-(invocation_expression
-  (member_access_expression
-    name: (identifier) @function))
-
-(invocation_expression
-  (member_access_expression
-    expression: (identifier) @variable))
-
-(invocation_expression
-  function: (conditional_access_expression
-    (member_binding_expression
-      name: (identifier) @function)))
-
-(invocation_expression
-      [(identifier) (qualified_name)] @function)
-
-; Generic Method invocation with generic type
-(invocation_expression
-  function: (generic_name
-              . (identifier) @function))
-
-;; Namespaces
-
-(namespace_declaration
-  name: [(identifier) (qualified_name)] @namespace)
+(method_declaration (identifier) @type (identifier) @function)
 
 ;; Types
 (interface_declaration name: (identifier) @type)
@@ -37,11 +7,11 @@
 (enum_declaration name: (identifier) @type)
 (struct_declaration (identifier) @type)
 (record_declaration (identifier) @type)
+(record_struct_declaration (identifier) @type)
 (namespace_declaration name: (identifier) @type)
-(using_directive (_) @namespace)
+
 (constructor_declaration name: (identifier) @type)
-(object_creation_expression [(identifier) (qualified_name)] @type)
-(type_parameter_list (type_parameter) @type)
+(destructor_declaration name: (identifier) @type)
 
 [
   (implicit_type)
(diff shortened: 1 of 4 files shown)


the-mikedavis (commented):
Jumping around in some C# files I have locally, I think we should also have function textobjects for property definitions like:

```csharp
public override bool CanRead {
	get { return true; }
}
```

But I'm not sure if we should tag the entire property as `function.around` (the `property_declaration` node) or just the accessor declaration (`accessor_declaration` node)

Also constructor definitions:

```csharp
public BufferPoolStream(BufferPool bufferPool) {
	if (bufferPool == null)
		throw new ArgumentNullException("bufferPool");
	_bufferPool = bufferPool;
}
```

```scm
(constructor_declaration body: (_) @function.inside) @function.around
```

votagus:
Do you think it would be appropriate to map it like this until we have more precise navigation commands?
```
([
  (class_declaration body: (_) @class.inside)
  (struct_declaration body: (_) @class.inside)
  (interface_declaration body: (_) @class.inside)
  (enum_declaration body: (_) @class.inside)
  (delegate_declaration)
  (record_declaration body: (_) @class.inside)
  (record_struct_declaration body: (_) @class.inside)
  ]) @class.around
```
**Note: To make `record_struct_declaration` work the grammar revision has to be updated along with its highlights.scm.**

About `property_declaration` vs `accessor_declaration`, playing around, `(property_declaration (_) @function.inside) @function.around` _feels_ more consistent with the rest of the commands, specially when using it with match inside/around function (`mif`, `maf`).

Class for testing.
```csharp
namespace NavigationTest {
	public class ClassTest {
		public int field = 0;
		
		public int Field {
			get { return field; }
			set { field = value; }
		}

		public NavigationTest(string param1, int param2) {
			// Set field.
			field = param2;
		}

		~NavigationTest() {
			// Set field.
			field = param2;
		}

		/// Documentation.
		///
		/// Documentation.
		public int Method(string param1, int param2) {
			/* Multiline
				comment.
			*/
			return 0;
		}
	}
	
	public enum EnumTest {
		Item1,
		Item2
	}
	
	public interface InterfaceTest {
		public void Method();
	}
	
	public struct StructTest {
		public int field = 0;
	}

	public record RecordTest {
		public int field = 0;
	}

	public record struct RecordStructTest {
		public int field = 0;
	}
}
```

Let me know what you think.

runtime/queries/c-sharp/textobjects.scm
```diff
@@ -0,0 +1,21 @@
+([
+  (class_declaration body: (_) @class.inside)
+  (struct_declaration body: (_) @class.inside)
+  (interface_declaration body: (_) @class.inside)
+  (enum_declaration body: (_) @class.inside)
+  (delegate_declaration)
+  (record_declaration body: (_) @class.inside)
+  (record_struct_declaration body: (_) @class.inside)
+  ]) @class.around
```
the-mikedavis, line 9:
```suggestion
[
  (class_declaration body: (_) @class.inside)
  (struct_declaration body: (_) @class.inside)
  (interface_declaration body: (_) @class.inside)
  (enum_declaration body: (_) @class.inside)
  (delegate_declaration)
  (record_declaration body: (_) @class.inside)
  (record_struct_declaration body: (_) @class.inside)
] @class.around
```

runtime/queries/c-sharp/highlights.scm
```diff
@@ -1,47 +1,17 @@
 ;; Methods
-(method_declaration
```
the-mikedavis, line 2: Most of the changes that we have in highlights.scm are improvements that haven't been upstreamed. Can you revert the changes to that file? It looks like the only changes necessary for updating the grammar come from the preprocessor change in https://github.com/tree-sitter/tree-sitter-c-sharp/pull/191:

```diff
diff --git a/runtime/queries/c-sharp/highlights.scm b/runtime/queries/c-sharp/highlights.scm
index 6b94dd8c..05eda6b9 100644
--- a/runtime/queries/c-sharp/highlights.scm
+++ b/runtime/queries/c-sharp/highlights.scm
@@ -228,8 +228,21 @@
   "let"
 ] @keyword
 
-(preprocessor_call) @keyword.directive
-(preprocessor_call (_) @keyword.directive)
+[
+  (if_directive (_) @keyword.directive)
+  (elif_directive (_) @keyword.directive)
+  (else_directive)
+  (endif_directive)
+  (warning_directive (_) @keyword.directive)
+  (error_directive (_) @keyword.directive)
+  (line_directive (_)? @keyword.directive)
+  (nullable_directive)
+  (define_directive (_) @keyword.directive)
+  (undef_directive (_) @keyword.directive)
+  (region_directive)
+  (endregion_directive)
+  (pragma_directive)
+] @keyword.directive
 
 ;; Linq
 (from_clause (identifier) @variable)
```

votagus: Yeah, I just found out that by looking at the commits, I wrongly assumed that this file was in sync with the main repo, I'll try to add those directives as well.

the-mikedavis (approved):

the-mikedavis:
Looks great, thank you!

PR #3503, opened 2022-08-22
Fix rendering of indented code blocks in markdown

connorlay:
Relates to https://github.com/helix-editor/helix/pull/3425 and https://github.com/elixir-lsp/elixir-ls/issues/694

Properly handles the rendering of indented code blocks in markdown. Given that we do not have a language identifier available, we style these blocks as generic code (visually distinct from markdown text, but no highlighting via tree-sitter).

This improves the experience of writing Elixir in Helix, as the majority of documentation favors indentation over fences.

**Before** ☹️ 
image

**After** 😁 
image

### DIFF
--- a/helix-term/src/ui/markdown.rs
+++ b/helix-term/src/ui/markdown.rs
@@ -229,10 +229,7 @@ impl Markdown {
                 Event::End(tag) => {
                     tags.pop();
                     match tag {
-                        Tag::Heading(_, _, _)
-                        | Tag::Paragraph
-                        | Tag::CodeBlock(CodeBlockKind::Fenced(_))
-                        | Tag::Item => {
+                        Tag::Heading(_, _, _) | Tag::Paragraph | Tag::CodeBlock(_) | Tag::Item => {
                             push_line(&mut spans, &mut lines);
                         }
                         _ => (),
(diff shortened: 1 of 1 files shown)


archseer:
Do you think it would be worth highlighting these blocks using the parent document's language? If the file is elixir it's also very likely the code block is elixir (or iex)

the-mikedavis:
I'm not sure there's a good strategy for picking languages in indented code blocks in markdown: it could either be Elixir, IEx or HEEx (I think I've also seen some EEx and Erlang too) when injecting or using markdown from Elixir. We could try parsing the code block with each of them and use the first that doesn't have any errors, falling back to no syntax highlighting when all error out.

archseer:
I think fixing this upstream is still the best way to get it fully correct, but falling back to the main language should work for most languages. Elixir is an outlier where code examples can also consist of a mix of `IEx` snippets

connorlay:
@archseer @the-mikedavis I agree the long-term solution for Elixir is to handle language detection within the language server. Detecting language syntax in Helix via tree-sitter sounds like a viable solution too.

With that in mind, are you open to merging this as-is? Even without syntax highlighting I think this is a significant QOL improvement 🙂

PR #3505, opened 2022-08-22
Reduce health tick width

pickfire:
Use the same tick as book to reduce margin of whitespace errors.

Before
Screenshot_20220822_214053

After
Screenshot_20220822_214126

This may be a terminal issue, but at the same time I just make it consistent with the docs. Also, I noticed this because I saw the same symbol in https://rsdlt.github.io/posts/rust-nvim-ide-guide-walkthrough-development-debug/.


the-mikedavis (approved):
The shorter width is nice 👍

PR #3507, opened 2022-08-22
Julia queries: update grammar, major refactoring

ChrHorn:
This PR updates the grammar to the newest version and does a major refactoring of the `highlights.scm`.

I reorganized the file to make it more managable and fixed the broken, obsolete and missing queries (there was a lot). This will probably need another polishing pass in the future but this should be enough for now.

I also added regex injections and some locals.

### DIFF
--- a/runtime/queries/julia/injections.scm
+++ b/runtime/queries/julia/injections.scm
@@ -1,5 +1,8 @@
-((triple_string) @injection.content
- (#set! injection.language "markdown"))
+; ((string_literal) @injection.content
+;  (#set! injection.language "markdown"))
 
-((comment) @injection.content
- (#set! injection.language "comment"))
+; ((line_comment) @injection.content
+;  (#set! injection.language "comment"))
+
+; ((block_comment) @injection.content
+;  (#set! injection.language "comment"))
--- a/runtime/queries/julia/injections.scm
+++ b/runtime/queries/julia/injections.scm
@@ -1,8 +1,16 @@
-; ((string_literal) @injection.content
-;  (#set! injection.language "markdown"))
+(
+  (string_literal) @injection.content
+  (#set! injection.language "markdown"))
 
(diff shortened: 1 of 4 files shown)


runtime/queries/julia/injections.scm
```diff
@@ -1,5 +1,16 @@
-((triple_string) @injection.content
- (#set! injection.language "markdown"))
+(
+  (string_literal) @injection.content
+  (#set! injection.language "markdown"))
```
the-mikedavis, line 3: Won't this try to inject markdown into all strings?

ChrHorn: This was one of the few parts that I didn't change. I modified it, so that now only docstrings will be captured here.

runtime/queries/julia/injections.scm
```diff
@@ -1,5 +1,29 @@
...
+        (function_definition)
+        (macro_definition)
+        (primitive_definition)
+        (abstract_definition)
+        (struct_definition)
+        (assignment_expression)
+        (const_statement)
+      ]
+    ))
+  (#set! injection.language "markdown"))
```
the-mikedavis, line 16: For this we'll want to use the anchor operator `.` to only match immediate siblings, or else the doc-string will apply to a function for example even if there's a for loop between the doc-string and function definition

```scm
(((string_literal) @injection.content
  .
  [
    (module_definition)
    (function_definition)
    (macro_definition)
    (primitive_definition)
    (abstract_definition)
    (struct_definition)
    (assignment_expression)
    (const_statement)
  ])
  (#set! injection.language "markdown"))
```

ChrHorn: Yeah, read this one wrong. Should be fixed now.

the-mikedavis (approved):

the-mikedavis:
Looks great, thank you!

PR #3508, opened 2022-08-22
Use the current document and view for mode-change hooks

the-mikedavis:
When changing the focused view with C-w for example, the w event was
being handled in the EditorView event handler as if it belonged to
the newly focused window.

~Instead of handling the event for the newly focused window, we should
return early. We don't want the change in focus to trigger a history
checkpoint (following block) and we don't want to execute the mode
transition hooks (the block after). With some configs this can cause
a panic, for example~

    keys.normal.w = ["no_op"]
    # or
    keys.normal.C-w.m = "rotate_view"

~since these fall into the `unimplemented!` in the mode transition block.~

Instead of handling the event in the newly focused view, we should handle
the event in the original view.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1140,7 +1140,8 @@ impl Component for EditorView {
                 // clear status
                 cx.editor.status_msg = None;
 
-                let doc = doc!(cx.editor);
+                let (view, doc) = current!(cx.editor);
+                let focus = view.id;
                 let mode = doc.mode();
 
                 if let Some(on_next_key) = self.on_next_key.take() {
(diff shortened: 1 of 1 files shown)


archseer:
I've pushed https://github.com/helix-editor/helix/commit/e4c9d4082a139aac3aea4506918171b96e81f5b9 but I think this PR should be applied as well.

helix-term/src/ui/editor.rs
```diff
@@ -1209,6 +1210,12 @@ impl Component for EditorView {
                 let (view, doc) = current!(cx.editor);
                 view.ensure_cursor_in_view(doc, config.scrolloff);
 
+                // The focused view has changed. Consume the event and do not execute mode
+                // transition hooks.
+                if view.id != focus {
+                    return EventResult::Consumed(None);
+                }
```
archseer, line 1217: Actually, with the changes in master, these hooks should still run (if the doc exited from insert mode, we *do* want a history state).

We just need to fetch the doc matching the original `focus` rather than `current!`

PR #3513, opened 2022-08-23
packaging: Point to homebrew-core instead of tap

the-mikedavis:
Now that helix is packaged in homebrew-core, I don't think it's necessary to run the tap anymore.

It's difficult to install a formula from a tap because core has higher precedence than taps: you need to first tap helix `brew tap helix-editor/homebrew-helix` and then `brew install helix-editor/homebrew-helix/helix` (https://docs.brew.sh/Taps#formula-with-duplicate-names).

### DIFF
--- a/README.md
+++ b/README.md
@@ -70,10 +70,9 @@ for a language.
 
 ## MacOS
 
-Helix can be installed on MacOS through homebrew via:
+Helix can be installed on MacOS through homebrew:
 
 ```
-brew tap helix-editor/helix
 brew install helix
 ```
 
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -6,10 +6,9 @@ We provide pre-built binaries on the [GitHub Releases page](https://github.com/h
 
 ## OSX
 
-A Homebrew tap is available:
+Helix is available in homebrew-core:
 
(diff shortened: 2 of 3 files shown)


archseer:
We should also deprecate the tap by opening a PR here: https://github.com/brewsci/homebrew-science/blob/master/tap_migrations.json

archseer:
The homebrew package should also remove grammar sources.

the-mikedavis:
> The homebrew package should also remove grammar sources.

Just upstreamed this in https://github.com/Homebrew/homebrew-core/pull/108657 👍

For deprecating the tap, do you mean to create a `tap_migrations.json` in the homebrew-helix repo? Looks like we should delete the formula there too to truly deprecate the tap: https://docs.brew.sh/Migrating-A-Formula-To-A-Tap. I'll make a PR

archseer:
No, looks like taps use a centralized `tap_migrations.json` in that repo

archseer:
e.g. https://github.com/brewsci/homebrew-science/pull/296

the-mikedavis:
I think those are only for the formulas that were packaged in that tap

PR #3516, opened 2022-08-23
Add row text for global search

ycycyyc:
When doing a global query, add row text, then we can do fuzzy matching.


QiBaobin:
like this, that's why I opened #3462 . Just wondering would if it is readable when file path is very long.

the-mikedavis:
A change along these lines will eventually be part of #196. Currently the fuzzy match operates on the entire row with no way to specify whether you want to search the filename or the line matched by the regex. That will come with #7265

PR #3521, opened 2022-08-23
add syntax highlights for xit

ChemicalXandco:
[xit](https://xit.jotaen.net/) ([specification](https://github.com/jotaen/xit/blob/main/Specification.md)) is a plain-text file format for todos and check lists.

Currently tags and dates are not supported in the tree sitter grammar implementation.

Default theme:
image

`dark_plus`:
image


runtime/queries/xit/highlights.scm
```diff
@@ -0,0 +1,11 @@
+(headline) @markup.heading
+
+(open_checkbox) @markup.list
+(ongoing_checkbox) @keyword.control
+(checked_checkbox) @diff.plus
+(obsolete_checkbox) @hint
+
+(checked_task) @hint
+(obsolete_task) @hint
```
the-mikedavis, line 9: I'm not very familiar with xit but `@comment.unused` might be a good scope for these. That scope is usually used for languages with pattern matching where a variable is underscored when it's unused:

```elixir
{_x, y} = {100, 50}
y * 2
```

What do you think?

ChemicalXandco: yes, you are right, I have amended the changes

Issue #3519, opened 2022-08-23
There is no way to `cd` to the directory of the current file


cole-h:
#### Describe your feature request

I'd like to be able to to `cd` to the directory the current file is in.

cole-h:
The discussion in https://github.com/helix-editor/helix/pull/3523 has somewhat changed this ticket. I've modified the title and body to reflect that.

pascalkuthe:
`:cd <c-r>%<c-w>` makes this very convenience. Can be bound as a keymap as a macro or when command expansoon lands so closing as duplicate of either those two

cole-h:
Now that https://github.com/helix-editor/helix/pull/4709 has merged, here's a macro you can use to get this functionality!

```toml
[keys.normal.space]
"c" = "@:cd <C-r>%<C-w><ret>"
```

Pressing <kbd>space</kbd><kbd>c</kbd> will `cd` helix to the directory containing the current file!

ramdomPTM:
Sorry for replying to such an old thread. Thanks for providing a solution to a function I have needed. It worked well until there was whitespace in the path.

Now that command line expansions are available from 25.07, I think `":cd %sh{dirname '%{buffer_name}'}"` is a more robust solution for Linux/macOS users.

Cheers.

Rudxain:
> ```toml
> [keys.normal.space]
> "c" = "@:cd <C-r>%<C-w><ret>"
> ```

I felt so dumb until I realized there are problems with that:
- spaces, as already mentioned
- multiple "words" per path component, which `<C-w>` won't delete

The `dirname` alternative is prone to code-injection if the path contains single-quotes: #14149

My solution is a register-buffer "dance" with proper parsing:
```toml
[keys.normal.backspace]
# clobbers reg-p and `repeat_last_motion` internal state
c = """@\
<C-s>\
:set-register p %{file_path_absolute}<ret>\
:n<ret>\
\"pp\
F/\
<A-d>\
x\
:cd %{selection}<ret>\
:bc!<ret>\
<C-o>\
"""
```
> [!note]
> set-reg is master-only (for now) #14779

I've now realized this won't work if the path has `'\n'` 🥲

I've wrote a more robust version (assuming [this](https://github.com/helix-editor/helix/pull/4709#issuecomment-1323713162)), and it no longer clobbers registers!
```toml
[keys.normal.backspace]
c = [
	# remember target buffer index
	"save_selection",
	# make temporary-memory / scratch-space
	":n",
[...]

PR #3523, opened 2022-08-23
helix-term: introduce cdf / change-directory-to-current-file

cole-h:
Fixes: https://github.com/helix-editor/helix/issues/3519.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1081,6 +1081,43 @@ fn show_clipboard_provider(
     Ok(())
 }
 
+fn change_directory_impl(cx: &mut compositor::Context, dir: &Path) -> anyhow::Result<()> {
+    if let Err(e) = std::env::set_current_dir(dir) {
+        bail!("Couldn't change the current working directory: {}", e);
+    }
+
+    let cwd = std::env::current_dir().context("Couldn't get the new working directory")?;
+    cx.editor.set_status(format!(
+        "Current working directory is now {}",
+        cwd.display()
+    ));
+
+    Ok(())
+}
+
+fn change_directory_to_current_file(
+    cx: &mut compositor::Context,
+    _args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    let (_, doc) = current_ref!(cx.editor);
+    let dir = match doc.path() {
+        Some(path) => path
+            .parent()
+            .map(ToOwned::to_owned)
+            .context("Couldn't get parent of current document")?,
+        None => {
+            bail!("Current document has no path");
+        }
+    };
+
+    change_directory_impl(cx, &dir)
+}
+
 fn change_current_directory(
     cx: &mut compositor::Context,
     args: &[Cow<str>],
@@ -1097,13 +1134,7 @@ fn change_current_directory(
             .as_ref(),
     );
 
-    helix_loader::set_current_working_dir(dir)?;
-
-    cx.editor.set_status(format!(
-        "Current working directory is now {}",
-        helix_loader::current_working_dir().display()
-    ));
-    Ok(())
+    change_directory_impl(cx, &dir)
 }
 
 fn show_current_directory(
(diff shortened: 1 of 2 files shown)


cole-h:
It's been a while since I last contributed here, so please don't be shy on telling me how I can improve / make this better :)

AceofSpades5757:
IMHO
Due to how Helix likes to handle things, wouldn't it make more sense to go to the project root directory, like a "home" equivalent, with the fallback being the parent of the current file? There's another pr that's working on shell expansion that I think would solve this problem a little better.

cole-h:
I don't mind if there's a command for going to the project root, but I would _really_ like a command like this that allows me to go to the directory a file is in, and _not_ the project root. I usually start helix from the project root anyways, so that would be a completely useless no-op in almost every case of mine. 

Here's an example where I would find this supremely useful: in Nix, where relative paths are relative to the file referencing the path, it's nice to be able to make a new file just by `:cd` and then `:o newfile.nix` instead of having to type out the entire directory tree to where it should be located. `:o some/long/list/of/directories/before/we/get/to/where/we/need/to/place/the/newfile.nix`. Maybe slightly contrived, but hopefully communicates one of my usecases.

I'd be glad to introduce another command with this ability (`cd-to-file`, and `cdf`, maybe?) so that this other PR can do it's thing with `cd`, if that's deemed a better idea.

If there is no interest in this functionality whatsoever, please let me know and I'll move on.

poliorcetics:
I think `cd` alone should go to the "home" of the project for easier mental mapping: real `cd` goes to `$HOME`, helix's `cd` goes to project home.

Adding a `cdf` is a good idea though, I have often missed it. But maybe what you're missing is more of a new command to create a file next to another, like `ont`, `open-next-to`, where the default argument would be the current file but could be any file path

CBenoit:
> I think `cd` alone should go to the "home" of the project for easier mental mapping: real `cd` goes to `$HOME`, helix's `cd` goes to project home.

I agree with that, it feels more natural to me. I also expect this behavior to come in handy once a command such as `cd-to-file` is added.

> Adding a `cdf` is a good idea though, I have often missed it.

I think it would be a useful command too, but I’m a bit unsure about the naming. What about `cd-here`?

> But maybe what you're missing is more of a new command to create a file next to another, like `ont`, `open-next-to`, where the default argument would be the current file but could be any file path

I really like the `open-next-to` command idea.

cole-h:
I've rebased on the latest `master` and regenerated the docs. I reverted my changes to `cd` and instead introduce a new `cdf` / `change-directory-to-current-file` that implements my desired functionality. I'll leave the "`cd` without args should go to the project's root" functionality to another PR.

helix-term/src/commands/typed.rs
```diff
@@ -1897,6 +1925,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             fun: change_current_directory,
             completer: Some(completers::directory),
         },
+        TypableCommand {
+            name: "change-directory-to-current-file",
+            aliases: &["cdf"],
+            doc: "Change the current working directory to the parent of the current document.",
```
cole-h, line 1931: I wonder if this should become:
```suggestion
            doc: "Change the current working directory to the parent of the current file.",
```

or if everything else referencing "file" should become "document"...

the-mikedavis:
I think the change here looks good. I wonder if it's necessary to optimize this use-case so much though? There are a few other ways that I think could be fast enough:

* Having a register for the current filename, using `:cd`, inserting the filename, using `C-w` to delete back to the parent directory of that path, and then `<ret>`
* Having a variable for the current file's parent directory and `:cd $parent`
* Having a variable only for the current file and using `dirname` and some shell expansion to `cd` to that directory

cole-h:
Hi there! Now that 23.10 has released, is there anything I can do to help get this merged?

pascalkuthe:
> Having a register for the current filename, using :cd, inserting the filename, using C-w to delete back to the parent directory of that path, and then <ret>

this has been added in the latest release, with the command expansion PR we want to implement soon it will be possible to add a keymap for this without a dedicated builtin command so I would lean towards not merging this

cole-h:
Good to know, thanks! I'll close this and carry the patch locally until that comes about :)

cole-h:
Now that https://github.com/helix-editor/helix/pull/4709 has merged, here's a macro you can use to get this functionality!

```toml
[keys.normal.space]
"c" = "@:cd <C-r>%<C-w><ret>"
```

Pressing <kbd>space</kbd><kbd>c</kbd> will `cd` helix to the directory containing the current file!

PR #3526, opened 2022-08-24
feat(lang): Add esdl grammar

greym0uth:
Adds grammar for EdgeDB's SDL as well as highlight queries.

ESDL: https://www.edgedb.com/docs/datamodel/index
Grammar: https://github.com/greym0uth/tree-sitter-esdl

### DIFF
--- a/runtime/queries/esdl/highlights.scm
+++ b/runtime/queries/esdl/highlights.scm
@@ -1,0 +1,78 @@
+; Keywords
+[
+  "module"
+  "using"
+  "single"
+  "multi"
+  "link"
+  "property"
+  "constraint"
+  "tuple"
+  "annotation"
+  "abstract"
+  "scalar"
+  "type"
+  "required"
+  "optional"
+  "extension"
+  "function"
+] @keyword
+
+(modifier) @keyword
+(extending) @keyword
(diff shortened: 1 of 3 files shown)


runtime/queries/esdl/highlights.scm
```diff
@@ -0,0 +1,78 @@
...
+(extending) @keyword
+
+(module name: (identifier) @namespace)
+(object_type) @type
+
+(comment) @comment
+
+; Properties
+(property) @property
+(link) @property
```
the-mikedavis, line 31: this might also want to be `@variable.other.member` but if it's a link like a URL it might make sense to use `@markup.link.url`

the-mikedavis:
To fix the docs CI you'll need to run `cargo xtask docgen` and commit the changes

runtime/queries/esdl/highlights.scm
greym0uth: The link is essentially a property identifier so I updated to use `@variable.other.member`. Thanks for recommendation!

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #3535, opened 2022-08-24
AWK Regex injection

ath3:
Before:
awk_regex_before
After:
awk_regex_after


the-mikedavis (approved):

the-mikedavis:
Good call, that does look a lot nicer 👍

PR #3536, opened 2022-08-24
fix: unreachable panic because of crossterms default-enabled brackete…

ahkrr:
…d-paste

When trying to paste using `ctrl-shift-v`, `helix` would `panic!`.
This is the case because in `crossterm 0.25`
```
[features]
default = ["bracketed-paste"]
```
is enabled, which would lead to a panic in `helix-view/src/input.rs`
` unreachable!("crossterm shouldn't emit Paste events without them being enabled")`
After setting `default-features = false` for `crossterm 0.25`, the use of `crossterm::event::Event::Paste` now leads to a compiler error because it is no longer included.
After removing the single use of Event::Paste, helix successfully compiles, doesn't panic, the tests pass and pasting using ctrl-shift-v also works.

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -38,7 +38,7 @@ which = "4.2"
 
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "io-util", "io-std", "time", "process", "macros", "fs", "parking_lot"] }
 tui = { path = "../helix-tui", package = "helix-tui", default-features = false, features = ["crossterm"] }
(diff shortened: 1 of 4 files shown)


the-mikedavis:
See also #3233 which fixes the panic by implementing bracketed paste.

This branch could be preferrable if #3233 needs any extra work though as a way to unblock the 22.08 release.

ahkrr:
Ok. I will close this PR if the feature lands or it is no longer needed for other reasons.

archseer:
Merged https://github.com/helix-editor/helix/pull/3233

PR #3537, opened 2022-08-24
Match smallest selected node

garlic0x1:
I have added `mn` to the match commands to match the smallest nodes in each selection.


the-mikedavis:
It seems like this is effectively the same as `A-o` (`expand_selection`)?

garlic0x1:
> It seems like this is effectively the same as `A-o` (`expand_selection`)?

I was not aware of this, thank you :)

Issue #3533, opened 2022-08-24
Update doom_acario_dark theme for Python


AceofSpades5757:
Not too familiar with themes and how they tie into tree-sitter/etc., but I noticed a couple that will make import blocks in Python look pretty wild.

This is `doom_acario_dark`, which looks really clean, until you scroll to the top.

![image](https://user-images.githubusercontent.com/10341888/186480786-37a753f2-bd13-4c34-9353-bcbc784cc386.png)

the-mikedavis:
`doom_acario_dark` highlights the `namespace` and `keyword` both as "red." `from` and `import` are captured as `keyword`s and `datetime` on the first line for example is tagged as `namespace` in both usages.

https://github.com/helix-editor/helix/blob/45add73fb1d2c4f2b0c544117c2d83b7081a2a4a/runtime/queries/python/highlights.scm#L3-L4

```scm
; from datetime import datetime
(import_from_statement
  module_name: (dotted_name (identifier))
  module_name: (dotted_name (identifier))
```

IMO the `keyword` and `namespace` captures for python are good choices. `doom_acario_dark` would probably benefit from changing the `namespace` style.

archseer:
\cc @Rinnray

Rinnray:
I'll get to it right away.

Rinnray:
I'm not really sure what's going on 😓 When I open up a python file it looks like this
![image](https://user-images.githubusercontent.com/47090855/186551664-0edfe2ac-7cd8-4e77-baf3-10a14c734d42.png)

Changing `namespace` doesn't change anything visible at least from that little snippet. From what I can see here at least, `datetime` is captured as a variable, so when I change `variable` to e.g green
![image](https://user-images.githubusercontent.com/47090855/186552009-82f6613f-a365-4d45-aa8d-44d9003a528b.png)

Don't know if it's something I'm missing or what (probably is).
I could definitely still change the `namespace` to be another color; it's just that I can't check if it'll fix it or not myself.

Rinnray:
Nevermind! I built Helix from source (had it from the AUR) and it is `namespace`. Will be setting it to correct color and make a pr with the fix 😄

the-mikedavis:
Fixed by https://github.com/helix-editor/helix/pull/3539

PR #3539, opened 2022-08-25
Fixed Doom acario theme

Rinnray:
Changed the `namespace` style to fix the issue (#3533).

I also made the theme look a little closer to how it looks in Emacs, I did however opt to still have it slightly different as I found it easier to read with my port than on the original in Emacs.
I also sorted most keys (mainly from line 8 to 28) for the theme to be in alphabetical order, so it's easier to have a quick glance where they are.


pickfire:
Can you please update the wiki screenshots as needed?

pickfire (approved):
Looks good to me.

PR #3541, opened 2022-08-25
LSP match on TextDocumentSyncKind capability

dariooddenino:
Hi!

`text_document_did_save` at the moment is ignoring some `TextDocumentSyncCapability` values. 
This results in Purescript LSP server not working properly, as it's never getting the save message.

From the code documentation it looks like `Kind(FULL)` should send the full content of the document, so I'm returning `true` here for `include_text`, but it's a bit of a shot in the dark.
It works, but I fear it might not be the proper solution.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -611,6 +611,7 @@ impl Client {
                 // Supported(false)
                 _ => return None,
             },
+            Some(lsp::TextDocumentSyncCapability::Kind(lsp::TextDocumentSyncKind::FULL)) => true,
             // unsupported
             _ => return None,
         };


the-mikedavis:
From the LSP spec, `textDocumentSync` appears to be allowed to be a `TextDocumentSyncKind` for backwards compatibility:

> ```typescript
> interface ServerCapabilities {
> 	/**
> 	 * Defines how text documents are synced.
> 	 * Is either a detailed structure defining each notification
> 	 * or for backwards compatibility, the TextDocumentSyncKind number.
> 	 * If omitted, it defaults to `TextDocumentSyncKind.None`.
> 	 */
> 	textDocumentSync?: TextDocumentSyncOptions | number;
>   // ...
> ```

Paging through the versions, it looks like `TextDocumentSyncOptions` was introduced in the 3.0.0 version of the spec being a `TextDocumentSyncKind` beforehand. In version 2.x, the spec only sent the text document identifier rather than the full text on `textDocument/didSave`. So it isn't explicit in the spec whether `TextDocumentSyncKind.Full` implies sending the entire document in the save notification.

The current version of the spec does say though:

> ```typescript
> export interface SaveOptions {
> 	/**
> 	 * The client is supposed to include the content on save.
> 	 */
> 	includeText?: boolean;
> }
> 
> interface DidSaveTextDocumentParams {
> 	/**
> 	 * The document that was saved.
> 	 */
> 	textDocument: TextDocumentIdentifier;
> 
> 	/**
> 	 * Optional the content when saved. Depends on the includeText value
> 	 * when the save notification was requested.
> 	 */
> 	text?: string;
> }
> ```

So it's ambiguous but I think the current state and history of the spec do not imply that we should send the full document on the `textDocument/didSave` when `textDocumentSync` is a `TextDocumentSyncKind.Full`.

IMO the purescript LS should not be sending a `TextDocumentSyncKind` for its `textDocumentSync` capability as that type is relatively ancient in the spec; it should be sending a `TextDocumentSyncOptions` to be unambiguous.

dariooddenino:
Thanks for the detailed reply! I'll see if they have plans to update this on their side

archseer:
lsp-mode logic for comparison: https://github.com/emacs-lsp/lsp-mode/blob/a0e1210f626cb7b5db16a9454d3bf61322d299df/lsp-mode.el#L3714-L3718

It seems to send save notifications if sync kind != none, and it does send full docs each time.

dariooddenino:
I think this can be closed

Issue #3522, opened 2022-08-23
To suspend helix while in the `git commit` editor, you have to press `^Z` twice


cole-h:
### Summary

When suspending `hx` with <kbd>Ctrl</kbd><kbd>z</kbd> after running `git commit`, helix goes away, but the terminal is still unusable. Pressing <kbd>Ctrl</kbd><kbd>z</kbd> again leads to the expected result of regaining control of the shell.

### Reproduction Steps

 

I tried this:

1. `cd $(mktemp -d)`
2. `git init`
3. `git commit --allow-empty`
4. <kbd>Ctrl</kbd><kbd>z</kbd>
5. Notice you can't do anything
6. <kbd>Ctrl</kbd><kbd>z</kbd>
7. Notice you're back in the shell

I expected this to happen:

Regain control of the shell immediately.

Instead, this happened:

I had to <kbd>Ctrl</kbd><kbd>z</kbd> again to regain control of the shell.

### Helix log

N/A.

### Platform

Linux

### Terminal Emulator

foot version: 1.13.0 +pgo +ime +graphemes -assertions

### Helix Version

helix 22.05 (18909aaf)

PR #1843, opened 2022-03-19
Fix suspend when running hx within git

pickfire:
When editing git commit message using helix, when ctrl-z is pressed
the terminal will get stuck since helix does not suspend correctly,
as helix is currently using tokio with multi-threads, with signal-hook
default low_level::emulate_default_handler it only send the SIGSTOP to
the current process but tokio uses multi-process so not everything gets
suspended and the terminal also gets stuck.

Now switched to use libc kill function and send it to pid 0 so that all
processes gets suspended correctly.

Maybe I need to create an issue in signal-hook for this to check why they didn't kill all processes.

PR #3546, opened 2022-08-25
helix-term: send the STOP signal to all processes in the process group

cole-h:
From kill(3p):

    If pid is 0, sig shall be sent to all processes (excluding an unspecified set
    of  system processes) whose process group ID is equal to the process group ID
    of the sender, and for which the process has permission to send a signal.

This fixes the issue of running `git commit`, attempting to suspend
helix with ^Z, and then not regaining control over the terminal and
having to press ^Z again.

---

Fixes https://github.com/helix-editor/helix/issues/3522.

---

I opted to add the `nix` crate because it has nice Rust wrappers around e.g. `kill`, handling error cases and other stuff for us. However, I could easily port it to the `libc` crate directly, if desired.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -40,10 +40,7 @@ use anyhow::{Context, Error};
 
 use crossterm::{event::Event as CrosstermEvent, tty::IsTty};
 #[cfg(not(windows))]
-use {
-    signal_hook::{consts::signal, low_level},
-    signal_hook_tokio::Signals,
-};
+use {signal_hook::consts::signal, signal_hook_tokio::Signals};
 #[cfg(windows)]
 type Signals = futures_util::stream::Empty<()>;
 
@@ -447,7 +444,15 @@ impl Application {
         match signal {
             signal::SIGTSTP => {
                 self.restore_term().unwrap();
-                low_level::emulate_default_handler(signal::SIGTSTP).unwrap();
+
+                // A pid of 0 sends the signal to the entire process group, allowing the user to
+                // regain control of their terminal if the editor was spawned under another process
+                // (e.g. when running `git commit`).
+                nix::sys::signal::kill(
+                    nix::unistd::Pid::from_raw(0),
+                    Some(nix::sys::signal::SIGSTOP),
+                )
+                .unwrap();
             }
             signal::SIGCONT => {
                 self.claim_term().await.unwrap();
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -458,7 +458,9 @@ impl Application {
                 };
 
                 if res != 0 {
-                    eprintln!("{}", std::io::Error::from_raw_os_error(res));
+                    let err = std::io::Error::last_os_error();
+                    eprintln!("{}", err);
+                    let res = err.raw_os_error().unwrap_or(1);
                     std::process::exit(res);
                 }
             }
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -454,6 +454,13 @@ impl Application {
                     // A pid of 0 sends the signal to the entire process group, allowing the user to
                     // regain control of their terminal if the editor was spawned under another process
                     // (e.g. when running `git commit`).
+                    //
+                    // We have to send SIGSTOP (not SIGTSTP) to the entire process group, because,
+                    // as mentioned above, the terminal will get stuck if `helix` was spawned from
+                    // an external process and that process waits for `helix` to complete. This may
+                    // be an issue with signal-hook-tokio, but the author of signal-hook believes it
+                    // could be a tokio issue instead:
+                    // https://github.com/vorner/signal-hook/issues/132
                     libc::kill(0, signal::SIGSTOP)
                 };
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -445,14 +445,17 @@ impl Application {
             signal::SIGTSTP => {
                 self.restore_term().unwrap();
 
-                // A pid of 0 sends the signal to the entire process group, allowing the user to
-                // regain control of their terminal if the editor was spawned under another process
-                // (e.g. when running `git commit`).
-                nix::sys::signal::kill(
-                    nix::unistd::Pid::from_raw(0),
-                    Some(nix::sys::signal::SIGSTOP),
-                )
-                .unwrap();
+                let res = unsafe {
+                    // A pid of 0 sends the signal to the entire process group, allowing the user to
+                    // regain control of their terminal if the editor was spawned under another process
+                    // (e.g. when running `git commit`).
+                    libc::kill(0, signal::SIGSTOP)
+                };
+
+                if res != 0 {
+                    eprintln!("{}", std::io::Error::from_raw_os_error(res));
+                    std::process::exit(res);
+                }
             }
             signal::SIGCONT => {
                 self.claim_term().await.unwrap();
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -445,6 +445,11 @@ impl Application {
             signal::SIGTSTP => {
                 self.restore_term().unwrap();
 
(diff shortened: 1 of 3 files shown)


cole-h:
I tested this with:

```
$ cargo b
$ VISUAL=$PWD/target/debug/hx git commit --allow-empty
# ^Z
$ echo "tada I'm back at my terminal"
tada I'm back at my terminal
```

the-mikedavis:
Yeah I think it would be preferrable to use libc directly. Pulling in nix for a handful of small functions doesn't seem worth it.

cole-h:
Latest commit switches from `nix` to `libc`.

kirawi:
It might be worth adding a comment explaining why it's safe.

archseer:
See https://github.com/helix-editor/helix/pull/1843 and https://github.com/vorner/signal-hook/issues/132 we were hoping to fix this in signal-hook instead

cole-h:
I'll gladly revert this once that does get handled upstream, but seeing the upstream issue closed with no further activity does not inspire confidence it will be fixed anytime soon. This also resolves one of my longstanding annoyances when using helix is my general purpose editor, so I would really love if we could accept this as "good for now" until `signal-hook` decides to fix this case.

archseer:
Maybe you could just upstream the fix? @pickfire closed the issue since he couldn't track it down.

cole-h:
Unfortunately, I don't have the time or energy to attempt to design this in a way upstream might consider palatable at the moment. Sorry.

pickfire:
I closed the issue because I feel like it's a hack. I was thinking of checking this with tokio since I think the issue originates there but I don't have the energy to check with them back then.

helix-term/src/application.rs
```diff
@@ -387,7 +384,25 @@ impl Application {
...
+                    // (e.g. when running `git commit`).
+                    libc::kill(0, signal::SIGSTOP)
+                };
+
+                if res != 0 {
+                    let err = std::io::Error::last_os_error();
+                    eprintln!("{}", err);
+                    let res = err.raw_os_error().unwrap_or(1);
+                    std::process::exit(res);
+                }
```
pickfire, line 405: I didn't do this previously because I don't know what case it will happen, so I just drop in an `unwrap`.

cole-h:
Rebased.

pickfire:
Before merging this, I think we need an investigation in https://github.com/vorner/signal-hook/issues/132#issuecomment-1079897156, as well as looking to see if the issue lies somewhere within tokio. Both @cole-h and I probably won't be able to it at this moment of time.

helix-term/src/application.rs
```diff
@@ -470,7 +467,25 @@ impl Application {
...
+                // SAFETY:
+                //
+                // - helix must have permissions to send signals to all processes in its signal
+                //   group, either by already having the requisite permission, or by having the
+                //   user's UID / EUID / SUID match that of the receiving process(es).
+                let res = unsafe {
+                    // A pid of 0 sends the signal to the entire process group, allowing the user to
+                    // regain control of their terminal if the editor was spawned under another process
+                    // (e.g. when running `git commit`).
+                    libc::kill(0, signal::SIGSTOP)
```
pickfire, line 480: Can you please leave a comment saying why we need this? And at the same time also mention that this could be an upstream issue. I am thinking that we can just fix it in helix for now until we figure out why this happens.

pickfire:
Not sure why but when I tested with crossterm example again, it seemed to work.

I cloned crossterm and modified the example event-stream-tokio.

```rust
//! Demonstrates how to read events asynchronously with tokio.
//!
//! cargo run --features="event-stream" --example event-stream-tokio
//!
//! Add to dev-dependencies
//!
//! signal-hook-tokio = { version = "0.3", features = ["futures-v0_3"] }

use libc::c_int;
use signal_hook::consts::signal::*;
use signal_hook::low_level;
use signal_hook_tokio::Signals;
use std::{io::stdout, time::Duration};

use futures::{future::FutureExt, select, StreamExt};
use futures_timer::Delay;

use crossterm::{
    cursor::position,
    event::{
        DisableMouseCapture, EnableMouseCapture, Event, EventStream, KeyCode, KeyEvent,
        KeyModifiers,
    },
    execute,
    terminal::{disable_raw_mode, enable_raw_mode},
    Result,
};

const HELP: &str = r#"EventStream based on futures_util::Stream with tokio
 - Keyboard, mouse and terminal resize events enabled
 - Prints "." every second if there's no event
 - Hit "c" to print current cursor position
 - Use Esc to quit
"#;

async fn print_events() {
    const SIGNALS: &[c_int] = &[
        SIGTERM, SIGQUIT, SIGINT, SIGTSTP, SIGWINCH, SIGHUP, SIGCHLD, SIGCONT,
    ];
    let mut reader = EventStream::new();
    let mut sigs = Signals::new(SIGNALS).unwrap();

    loop {
        let mut delay = Delay::new(Duration::from_millis(1_000)).fuse();
        let mut event = reader.next().fuse();

        select! {
            _ = delay => { println!(".\r"); },
            maybe_event = event => {
                match maybe_event {
                    Some(Ok(event)) => {
                        println!("Event::{:?}\r", event);

                        if event == Event::Key(KeyCode::Char('c').into()) {
                            println!("Cursor position: {:?}\r", position());
                        }

                        if event == Event::Key(KeyCode::Esc.into()) {
                            break;
                        }

                        if matches!(event, Event::Key(KeyEvent {
                            code: KeyCode::Char('z'),
                            modifiers: KeyModifiers::CONTROL,
                            ..
                        })) {
                            low_level::raise(SIGTSTP).unwrap();
                        }
                    }
                    Some(Err(e)) => println!("Error: {:?}\r", e),
                    None => break,
                }
            }
            maybe_signal = sigs.next().fuse() => {
                if let Some(signal) = maybe_signal {
                    println!("Received signal {:?}\r", signal);
                    // After printing it, do whatever the signal was supposed to do in the first place
                    if let SIGTSTP = signal {
                        disable_raw_mode().unwrap();
                        low_level::emulate_default_handler(signal).unwrap();
                        enable_raw_mode().unwrap();
                    } else {
                        low_level::emulate_default_handler(signal).unwrap();
                    }
                }
            }
        };
    }
}

#[tokio::main]
async fn main() -> Result<()> {
    println!("{}", HELP);

    enable_raw_mode()?;

    let mut stdout = stdout();
    execute!(stdout, EnableMouseCapture)?;

    print_events().await;

    execute!(stdout, DisableMouseCapture)?;

    disable_raw_mode()
}
```

helix-term/src/application.rs
cole-h: I can add a comment that it may be an upstream issue and link to the previous issue, but I'm not sure what more you want about "why we need this" (I thought that "allowing the user to regain control of their terminal if the editor was spawned under another process (e.g. when running `git commit`)." was pretty good :sweat_smile: ). Could you give a suggestion on what's missing?

pickfire: Why we need this as in

> Currently SIGSTOP only stops main parent process (or thread, I am not sure) of helix but subsequent processes (or threads?) does not receive the signal from tokio, and the terminal will get stuck in raw mode rendering it unusable when it was stopped from a 3rd party process like git as subsequent processes are still running, we believe this could be an issue in signal-hook-tokio but the author of signal-hook says it could be a tokio issue instead. So we have to send a SIGSTOP to the entire process group to make sure it is fully stopped.

I think the explanation I have isn't clear and can be improved, but that's what I have in mind.

cole-h: Ah, I see. I'll noodle and see if I can't make it more clear. Thanks for the tip!

cole-h: OK, I've added a comment hopefully explaining more why this is necessary.

pickfire (approved):
Looks good to me and well documented. Thanks for working on this and keeping it up to date.

Since we don't really know for a good non-hacky way to fix this just yet, we can go with this for now since it solves the issue too.

PR #3547, opened 2022-08-25
feat: ESC will escape out of selections in normal mode

AlexanderBrevig:
This will enable you to back out of normal mode selections with ESC.

`xxxESC` will put you right back to where you started
`vxxxESC` will place you in normal mode with everything selected, ESC again will get you back to start
`waESC` still works and corrects the cursor placement

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -483,6 +483,11 @@ impl Selection {
         Self::single(pos, pos)
     }
 
+    /// Converts selection to point
+    pub fn into_point(self) -> Self {
+        Self::single(self.ranges[0].anchor, self.ranges[0].anchor)
+    }
+
     /// Normalizes a `Selection`.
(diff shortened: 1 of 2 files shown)


the-mikedavis:
You can emulate this behavior currently with `<A-;>;,` or make a binding for it

```toml
[keys.normal]
esc = ["flip_selections", "collapse_selection", "keep_primary_selection"]
```

I don't think this should be default behavior of escape since selections aren't a mode.

AlexanderBrevig:
Thanks @the-mikedavis.

I'll add this in my config then.

BTW: `A-;` is not possible to type on my keyboard, or at least Helix does not react to it. `A-S-,` is what I'd have to do.

AlexanderBrevig:
Good arguments to avoid this, and a suitable workaround provided. Thanks again @the-mikedavis

PR #3549, opened 2022-08-26
Select inserted space after join

pickfire:
[asciicast](https://asciinema.org/a/Y5Jq6mJHg41UUuZeInU4jfJGw)

Will be useful to remove those inserted space if I want.

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -129,6 +129,7 @@
 | `X`                   | Extend selection to line bounds (line-wise selection)             | `extend_to_line_bounds`              |
 | `Alt-x`               | Shrink selection to line bounds (line-wise selection)             | `shrink_to_line_bounds`              |
 | `J`                   | Join lines inside selection                                       | `join_selections`                    |
+| `A-J`                 | Join lines inside selection and select space                      | `join_selections_space`              |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
I think in kakoune this ends up being a different command - `A-J` if I remember correctly?

EpocSquadron:
Yes and it should remain a separate command as you don't want to modify the users selection except as opt-in, say if you have a range inclusive of many newlines and want to join them internally, you may not want to lose the overall selection.

book/src/keymap.md
```diff
@@ -129,6 +129,7 @@
 | `X`                   | Extend selection to line bounds (line-wise selection)             | `extend_to_line_bounds`              |
 | `Alt-x`               | Shrink selection to line bounds (line-wise selection)             | `shrink_to_line_bounds`              |
 | `J`                   | Join lines inside selection                                       | `join_selections`                    |
+| `A-J`                 | Join lines inside selection and select space                      | `join_selections_space`              |
```
pickfire, line 132: I am not sure if we want to use `A-j` or `A-J` so I just go with `A-J`.

Omnikar: `A-J` makes more sense since it's an alternate behaviour of `J`.

PR #3553, opened 2022-08-26
Find workspace from document path

ath3:
When starting helix outside of a workspace/git repository, the workspace discovery fails, because he only goes upwards from cwd.
This patch makes the workspace discovery start from the directory of the opened document.


archseer:
Nice, I think this fixes issues I've been having with gopls under a subdirectory of the main workspace 👍🏻

ath3:
I rebased the pr, unless there are some changes required i think its ready for a review/merge.

Issue #2052, opened 2022-04-09
resolve completion item documentation on completion-menu selection


the-mikedavis:
The LSP [docs on completion request](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_completion) have a nice blurb about this:

> If computing full completion items is expensive, servers can additionally provide a handler for the completion item resolve request (`completionItem/resolve`). This request is sent when a completion item is selected in the user interface. A typical use case is for example: the `textDocument/completion` request doesn’t fill in the documentation property for returned completion items since it is expensive to compute.

The Erlang language server only sends documentation for completion items on `completionItem/resolve` (I assume because it's non-trivial to lookup documentation).

Upon a `Menu`'s `PromptEvent::Update` event, we should `resolve_completion_item` (implemented in #1315) if there is not existing documentation and then merge in the response to the currently selected `item`. I believe the request can be done in this block: https://github.com/helix-editor/helix/blob/78b16009433041059597cc6ccea28c17254483b7/helix-term/src/ui/completion.rs#L140-L160

We currently trigger the resolve on the `PromptEvent::Validate` event in order to resolve additional text edit actions. If I hit enter (triggers validate) on a completion item in the menu, the Erlang LS already sends:

```
2022-04-08T23:19:19.351 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"completionItem/resolve","params":{"data":{"arity":2,"function":"get","module":"maps"},"insertText":"get","insertTextFormat":1,"kind":3,"label":"get/2"},"id":9}
[...]

PR #3554, opened 2022-08-26
lsp: Check server provider capabilities

the-mikedavis:
Language Servers may signal that they do not support a method in the initialization result (server capabilities). We can check these when making LSP requests and hint in the status line when a method is not supported by the server. This can also prevent crashes in servers which assume that clients do not send requests for methods which are disabled in the server capabilities.

We already check server capabilities for document notifications and signature-help. This change covers manually invoked LSP commands like hover, documentHighlight, codeActions, etc.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -333,7 +333,14 @@ pub fn symbol_picker(cx: &mut Context) {
     let current_url = doc.url();
     let offset_encoding = language_server.offset_encoding();
 
-    let future = language_server.document_symbols(doc.identifier());
+    let future = match language_server.document_symbols(doc.identifier()) {
+        Some(future) => future,
+        None => {
+            cx.editor
+                .set_error("Language server does not support document symbols");
+            return;
+        }
+    };
 
     cx.callback(
         future,
@@ -365,7 +372,14 @@ pub fn workspace_symbol_picker(cx: &mut Context) {
     let current_url = doc.url();
     let language_server = language_server!(cx.editor, doc);
     let offset_encoding = language_server.offset_encoding();
-    let future = language_server.workspace_symbols("".to_string());
+    let future = match language_server.workspace_symbols("".to_string()) {
+        Some(future) => future,
+        None => {
+            cx.editor
+                .set_error("Language server does not support workspace symbols");
+            return;
+        }
+    };
 
     cx.callback(
         future,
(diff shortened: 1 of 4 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -325,6 +325,16 @@ pub fn symbol_picker(cx: &mut Context) {
...
     let language_server = language_server!(cx.editor, doc);
+
+    if matches!(
+        language_server.capabilities().workspace_symbol_provider,
+        None | Some(lsp::OneOf::Left(false))
+    ) {
+        cx.editor
+            .set_status("Document symbols are not enabled by the Language Server");
+        return;
+    }
```
archseer, line 336: I'd rather match on the success case, than negative since it's easy to miss a case. We should also decide if we want to handle this inside commands, or inside the LSP client: 

https://github.com/helix-editor/helix/blob/83f177d2701b0ec58abcb47c2a1ee133e72342e4/helix-lsp/src/client.rs#L760-L766

 I was thinking these calls could return a `Result<Future, Error>` so `Err::Unavailable`/`Unsupported` could be a generic way of reporting this back to the caller

the-mikedavis:
Signature help and some other requests in the LSP client module have a nice convention for this: return an Option over the future.

https://github.com/helix-editor/helix/blob/2d958d6c50b10318e29f6dacc291d8fcafd66de9/helix-lsp/src/client.rs#L657-L666

Having `Err::Unsupported` is tricky because the return type is currently `impl Future<Output = Result<_>>`, so if the method is unsupported, we need to wrap that error variant in a future instead of returning it immediately, or wrap the future in a Result itself which feels to me like too much. I'll refactor this to try out the `Option<impl Future>` pattern.

the-mikedavis:
I refactored the LSP method functions to all return `Option<impl Future>`s instead of use `async fn`s and checked the server capabilities within those functions.

I think there's a cleaner way where we could check within `Client::call` and have `Client::call` return an `Option<impl Future>` but in order to make it really nice we would want to add a member to the `Request` trait in the `lsp_types` crate:

```rs
pub trait Request {
    // ...

    fn is_supported(&self, server_capabilties: &ServerCapabilities) -> bool;
}
```

We might also want to make modifications for https://github.com/helix-editor/helix/issues/2052 (adding a `resolved: bool` field to `CompletionItem`) so maybe we should vendor `lsp_types`? OTOH it's a 5k SLOC library which is pretty big.

PR #3556, opened 2022-08-27
Allow less than and greater than in macros

### LINKED ISSUE #3538: Macros that do line indenting with < or > fails with 'Invalid macro'
### Summary

Recording a macro that contains a < or > indenting action fails with `Invalid macro: Unmatched '>'` or `Invalid macro: '>' expected`

### Reproduction Steps

I tried this:

1. `hx`
2. Add some text
3. `Q>Q`, e.g. record a macro that indents
4. Try running it with `q`, you'll get the above error

[asciicast](https://asciinema.org/a/GNUjqO4fzZEgy7uADk9G6Urkz)

I expected this to happen: The macro should run. In the example case, this means indenting the line again when pressing `q`

Instead, this happened: The aforementioned error occurred.

### Helix log

There wasn't any interesting output in the logs

### Platform

Linux

### Terminal Emulator

st 0.8.4

### Helix Version

helix 22.05 (0a221485) (e.g. basically HEAD)

A-Walrus:
Resolves #3538

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -555,8 +555,6 @@ mod test {
 
     #[test]
     fn parsing_unsupported_named_keys() {
-        assert!(str::parse::<KeyEvent>("lt").is_err());
-        assert!(str::parse::<KeyEvent>("gt").is_err());
         assert!(str::parse::<KeyEvent>("plus").is_err());
         assert!(str::parse::<KeyEvent>("percent").is_err());
         assert!(str::parse::<KeyEvent>("semicolon").is_err());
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Makes sense 👍. These were removed with https://github.com/helix-editor/helix/issues/2369 but we do end up needing `lt` and `gt` because the key-sequence DSL uses `<*>` for multi-character keys and key-combinations. With this change, a macro that uses `<`/`>` for indents are recorded as `<lt>`/`<gt>`.

archseer:
Great work as always @A-Walrus! 🎉

PR #3569, opened 2022-08-27
lint: catppuccin

AlexanderBrevig:
passes linter from https://github.com/helix-editor/helix/pull/3234
image


AlexanderBrevig:
@IsotoxalDev please sanity check my choices :)

ChrHorn:
Only setting `ui.cursorline.primary` and keeping `ui.cursorline` empty was an intentional choice in this case. I found multiple cursorlines a bit distracting. This way the single cursorline adds another indicator for the position of the primary cursor.

AlexanderBrevig:
This was changed before the linter allowed only `ui.cursorline.primary` closing as this is already in its intended state.

PR #3581, opened 2022-08-27
lint: penumbra+

AlexanderBrevig:
passes linter from https://github.com/helix-editor/helix/pull/3234
image

### DIFF
--- a/runtime/themes/penumbra+.toml
+++ b/runtime/themes/penumbra+.toml
@@ -45,7 +45,7 @@ error = "red"
 
 diagnostic = { modifiers = ["underlined"] }
 
-"ui.background" = "sky+"
+"ui.background" = { bg = "shade" }
 "ui.background.separator" = "sky"
 
 "ui.cursor" = { modifiers = ["reversed"] }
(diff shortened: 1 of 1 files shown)


AlexanderBrevig:
@vlmutolo please sanity check my choices :)

vlmutolo:
Hi @AlexanderBrevig. What's the goal of this commit? Knowing that would help me to comment on the screenshot. 

At first glance, judging only the screenshot you included, it seems like the documentation box is pretty bright for a dark theme. It's also kind of low-contrast.

AlexanderBrevig:
This is the current look on master when using a bright theme in my terminal.

image

The goal is to set what's needed to make sure that the user experience is in line with what I presume you intended :)

AlexanderBrevig:
Moved to bulk PR https://github.com/helix-editor/helix/pull/3587

vlmutolo:
Seems like the main issue here is that the background is the wrong color. I wonder if this is an issue with how Helix is rendering to the terminal.

AlexanderBrevig:
> Seems like the main issue here is that the background is the wrong color. I wonder if this is an issue with how Helix is rendering to the terminal.

`ui.background` and `ui.help` were not defined correctly, the fix is shown in the diff for this PR (and included in the PR linked above) 
When a theme does not set values, helix will default to the terminal for some things. 
Which looks weird with some configurations. In this case, a bright theme in the terminal, since bg were not set and thus defaulting to terminal settings. 
:)

AlexanderBrevig:
Sorry to be chatting in two places. I forwarded the use of sky, it should be shade. :) I'll fix my bug later today.

PR #3585, opened 2022-08-28
update colors for papercolor-light

uncomfyhalomacro:
image
now looks like this


AlexanderBrevig:
Nice!

uncomfyhalomacro:
ready for merging

pickfire (approved):
Thanks, but please update the wiki as necessary. I wonder if we should have a script to automatically update the wiki based so that contributors don't have to manually do that.

AlexanderBrevig:
> Thanks, but please update the wiki screenshots if necessary. I wonder if we should have a script to automatically update the wiki based so that contributors don't have to manually do that.

I definitely think we should do that. Do you know if anyone is working on it?

AlexanderBrevig:
I will give it a try!

PR #3588, opened 2022-08-28
Add wezterm to get_terminal_provider

unrelentingtech:
https://github.com/wez/wezterm is a terminal emulator with its own built-in multiplexer

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -211,6 +211,13 @@ pub fn get_terminal_provider() -> Option<TerminalConfig> {
         });
     }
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Since https://github.com/helix-editor/helix/commit/cb7615e0ed119f292eabdce784e4635247ed1eb1 this can be configured:

```config
# ~/.config/helix/config.toml
[editor.terminal]
command = "wezterm"
args = ["cli", "split-pane"]
```

I'm not sure we should add any more defaults. What do you think @archseer?

archseer:
I'm open to a few more common defaults, similar to what we do for the clipboard.

Issue #3589, opened 2022-08-28
Bracket goto skipping `function` `struct` `enum` `union` `trait` `impl` items and wrongly selecting comments in Rust


David-Else:
### Summary

```
// functions1.rs
// Execute `rustlings hint functions1` or use the `hint` watch subcommand for a hint.

// I AM NOT DONE

fn main() {
    call_me(10);
}

fn call_me(x: i32) -> i32 {
    x
}

fn call_me1(x: i32) -> i32 {
    x
}
```
- `gg]f` takes you to the second function, it should be `main`, the next one down
- `ge[f` 3 times results in the first function AND the comments being selected (as shown below):
![Screenshot from 2022-08-28 17-31-19](https://user-images.githubusercontent.com/12832280/187084640-55e9dc0d-a419-435d-aae1-87af97b5ea42.png)

The exact same thing happens for `struct` `enum` `union` `trait` and `impl` items as well.

### Reproduction Steps

In summary above

### Helix log

n/a

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.05 (e0667827)

the-mikedavis:
I'm not sure why comments are being matched on in the function textobject for rust: https://github.com/helix-editor/helix/blob/e06678278267a76a1d9e29b36d4734bcb39eabec/runtime/queries/rust/textobjects.scm#L2-L6

I think it should fix it to just have the textobject be

```scm
(function_item
  body: (_) @function.inside) @function.around
```

Would you like to make a PR?

David-Else:
I tried changing:
```
  (function_item
    body: (_) @function.inside)) @function.around
```
to
```
(function_item
  body: (_) @function.inside) @function.around
```
as you suggested, but it doesn't work. Is this what you meant? Sorry, this is the first time I have looked at a text object file.

the-mikedavis:
I mean that we can remove the whole outer part and only use the `function_item` pattern:

```diff
-(
-  [
-    (attribute_item)+
-    (line_comment)+
-  ]*
-  .
-  (function_item
-    body: (_) @function.inside)) @function.around
+(function_item
+  body: (_) @function.inside) @function.around
```

I think we should be able to do that for all of those patterns that start with `([(attribute_item)+ (line_comment)+]* .`

David-Else:
@the-mikedavis I have a PR https://github.com/helix-editor/helix/pull/3590. Please could you have a look? Thanks.

PR #3590, opened 2022-08-28
Fix rust text objects

David-Else:
Fixes: https://github.com/helix-editor/helix/issues/3589

I have tested them all, but am completely new to text objects, so please have a careful look at what I have done.


the-mikedavis (approved):
Works great, thanks!

Issue #3550, opened 2022-08-26
panicked at 'Attempt to slice past end of RopeSlice`


AndyJado:
0. changed my branch from another terminal window
1. I'm in a multi-window(3
2. `space + j` to jump list
3. panic!

---

```
thread 'main' panicked at 
'Attempt to slice past end of RopeSlice: slice end 2800,RopeSlice length 2179',
~/crate/src/ropey-1.5.0/src/slice.rs:645:9
```

the-mikedavis:
I can't reproduce this. Can you create an asciinema recording of the panic or write out more concrete reproduction steps? Also, what version are you running?

AceofSpades5757:
Using the latest version (86a8ea57bb3cb651cf9bee3fd1ae37f7e134c526), wasn't able to reproduce when changing the other file over switching branches.

n0s4:
I can reproduce in `8deaebde`, just open a new buffer and enter the following:

<kbd>]</kbd><kbd>space</kbd>  <kbd>geggd</kbd>  <kbd>space</kbd><kbd>j</kbd>

I'm guessing it's something to do with a previous jumplist location being beyond the length of the file.

the-mikedavis:
Thanks @n0s4! That's a really nice reproduction case 😀

That case itself is a rather small change to fix: https://github.com/the-mikedavis/helix/commit/7b691ab0d33b2992635e104247c6e67b2dba97da. The jumplist picker isn't checking the invariants of the selections in the jumplist which can lead to a panic when the selection rests on the EOF character.

Even with that fix though, we can cause further panics if we avoid the EOF character with a slightly more drastic edit: `2]<space>geggdd<space>j`. Crashes from that are also reproducible with `C-o`: `2]<space>geggdd<C-o>` which doesn't happen with your case. (The difference is that `C-o` checks selection invariants, like we do in the linked commit.)

The issue is that jumplist selections do not track document changes. You can try deleting some characters from a file and then use `C-o`: the jump point will move forward by the number of characters you deleted. I think I have a decent fix for this though, I'll make a PR.

AndyJado:
Now I've developed a habbit of `C-s` after doing anything and do the picking in git. Next time I'll show up with a easy reproductable issued, go helix!

Issue #2489, opened 2022-05-17
Helix crash on `Ctrl-o`


Zoybean:
### Summary

When pressing `ctrl-o` after deleting selection contents, helix crashes.
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 1794223, Rope/RopeSlice char length 1776214', C:\Users\runneradmin\.cargo\registry\src\github.com-1ecc6299db9ec823\ropey-1.4.1\src\slice.rs:349:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Reproduction Steps

 

I tried this:

1. select the whole file
2. select some subset of the file, and delete that subset
3. press `ctrl-o`

I expected this to happen:

- Either an error message, or the previously selected region of the file is reselected

Instead, this happened:

- helix crashed

### Helix log

No log available

### Platform

Windows 10

### Terminal Emulator

ConEmu 160914 [64] {Stable}

### Helix Version

helix 22.03 (d4e45fd4)

pickfire:
I noticed it too but I don't have exact step to reproduce, basically I just press `ctrl-o` many times when editing halfway then it crashed.

the-mikedavis:
There was a similar panic fixed in #1397 around jumping to a buffer that was closed with `:buffer-close` (https://github.com/helix-editor/helix/issues/2107). For this one

> Char index out of bounds: char index 1794223, Rope/RopeSlice char length 1776214

we might be able to reproduce by saving to the jumplist at the end of a file and then shortening the file by deleting some lines and jumping backwards.

archseer:
This indeed looks like https://github.com/helix-editor/helix/issues/2107 so it should be addressed on latest master (but not 22.03).

pickfire:
> This indeed looks like https://github.com/helix-editor/helix/issues/2107 so it should be addressed on latest master (but not 22.03).

I just faced it last week but latest master, but the issue I faced with `ctrl-o` could be a different issue.

David-Else:
I just had this on pressing `ctrl-o`:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 4081, Rope/RopeSlice char length 4079', /home/david/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.0/src/slice.rs:349:41
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
[...]

PR #3592, opened 2022-08-28
Track document changes in the jumplist

the-mikedavis:
Previously, the jumplist was not updated by changes to the document. This meant that the jumplist entries were prone to being incorrect after some edits and that usages of the jumplist were prone to panics if, for example, some edits truncated the document.

This change applies `Transaction`s (document edits) to the selections stored in any jumplists.

Implementation-wise I used a `Rc<RefCell<Selection>>` so that the `Document` could edit the selections when applying transactions and then the `View` could use the updated selections for jumping. Kakoune does this by waiting to apply transactions until `C-o`/`C-i` are used. I would prefer to do all of that work incrementally though since we have the jumplist picker (we would need to update all selections with all changes since the jumps were saved which seems like it could be a lot of work).

Closes https://github.com/helix-editor/helix/issues/3550
Closes https://github.com/helix-editor/helix/issues/2489

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2346,7 +2346,7 @@ fn jumplist_picker(cx: &mut Context) {
                 view.jumps
                     .get()
                     .iter()
-                    .map(|(doc_id, selection)| new_meta(view, *doc_id, selection.clone()))
+                    .map(|(doc_id, selection)| new_meta(view, *doc_id, selection.borrow().clone()))
             })
             .collect(),
         (),
@@ -2644,9 +2644,8 @@ fn try_restore_indent(doc: &mut Document, view_id: ViewId) {
 }
 
 // Store a jump on the jumplist.
-fn push_jump(view: &mut View, doc: &Document) {
-    let jump = (doc.id(), doc.selection(view.id).clone());
-    view.jumps.push(jump);
+fn push_jump(view: &mut View, doc: &mut Document) {
+    view.jumps.push(doc, doc.selection(view.id).clone());
 }
 
 fn goto_line(cx: &mut Context) {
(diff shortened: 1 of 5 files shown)


helix-term/src/commands.rs
```diff
@@ -4050,7 +4049,7 @@ fn jump_forward(cx: &mut Context) {
         view.doc = *id;
         let selection = selection.clone();
         let (view, doc) = current!(cx.editor); // refetch doc
-        doc.set_selection(view.id, selection);
+        doc.set_selection(view.id, selection.borrow().clone());
```
AlexanderBrevig, line 4052: I'm trying to learn so excuse me if this is obvious. But what effect does `clone().borrow().clone()` give us?

the-mikedavis: The first `clone()` is a clone of the [`Rc`](https://doc.rust-lang.org/std/rc/struct.Rc.html) and is needed to avoid borrowing conflict with `current!`. `Rc` clones are very cheap (clone of the reference, not the contained data) and I believe that one expires at the end of the block.

The call to `borrow()` is [`RefCell::borrow`](https://doc.rust-lang.org/std/cell/struct.RefCell.html#method.borrow) (note: the `Rc` is automatically dereferenced into the contained `RefCell`) which gives a reference to the data contained in the `RefCell`.

Then we need to `clone()` that since we want to pass a new `Selection` to `Document::set_selection` (more borrowing conflict with `current!`).

the-mikedavis: I suppose we can get rid of that Rc clone by moving the borrow part up

Jummit:
What's the status on this? I've face this issue and lost a bunch of work, and it was reported by three people, so there are probably lots of affected users. I think it would be cool to see this merged sooner than later.

the-mikedavis:
This approach works but the `Rc<RefCell<Selection>>` is kinda weird: it's odd having the selections on the Document and the View. Talking with archseer about this, we think we can apply changes to Views so that they can go against `View.jumps` directly. I'll open a separate PR that closes this one.

PR #3594, opened 2022-08-29
nix: add language injections

nrdxp:
By simply placing a comment with the name of the desired language above a multi-line string, that language will be injected.

Opening as a draft because I'm sure there must be a better way to write this, and hopefully someone with more experience can give me some pointers. Tried using repetition via `*` but I couldn't seem to get it right.

### DIFF
--- a/runtime/queries/nix/injections.scm
+++ b/runtime/queries/nix/injections.scm
@@ -1,0 +1,39 @@
+; mark arbitary languages with a comment
+((((comment) @injection.language) .
+  (indented_string_expression (string_fragment) @injection.content))
+  (#set! injection.combined))
+
+((binding
+   attrpath: (attrpath (identifier) @_path)
+   expression: (indented_string_expression
+     (string_fragment) @injection.content))
+ (#match? @_path "(^\\w*Phase|(pre|post)\\w*|(.*\\.)?\\w*([sS]cript|[hH]ook)|(.*\\.)?startup)$")
+ (#set! injection.language "bash")
+ (#set! injection.combined))
+
+((apply_expression
+   function: (apply_expression function: (_) @_func)
+   argument: (indented_string_expression (string_fragment) @injection.content))
+ (#match? @_func "(^|\\.)writeShellScript(Bin)?$")
+ (#set! injection.language "bash")
+ (#set! injection.combined))
+
+(apply_expression
+  (apply_expression
+    function: (apply_expression
+      function: ((_) @_func)))
+    argument: (indented_string_expression (string_fragment) @injection.content)
+  (#match? @_func "(^|\\.)runCommand(((No)?(CC))?(Local)?)?$")
+  (#set! injection.language "bash")
+  (#set! injection.combined))
+
+(apply_expression
+  function: ((_) @_func)
+  argument: (_ (_)* (_ (_)* (binding
+    attrpath: (attrpath (identifier) @_path)
+     expression: (indented_string_expression
+       (string_fragment) @injection.content))))
+  (#match? @_func "(^|\\.)writeShellApplication$")
+  (#match? @_path "^text$")
+  (#set! injection.language "bash")
+  (#set! injection.combined))
--- a/languages.toml
+++ b/languages.toml
@@ -331,7 +331,7 @@ source = { git = "https://github.com/omertuc/tree-sitter-go-work", rev = "6dd9dd
 [[language]]
 name = "javascript"
 scope = "source.js"
-injection-regex = "^(js|javascript)$"
+injection-regex = "(js|javascript)"
 file-types = ["js", "jsx", "mjs", "cjs"]
 shebangs = ["node"]
 roots = []
@@ -370,7 +370,7 @@ grammar = "javascript"
 [[language]]
 name = "typescript"
 scope = "source.ts"
-injection-regex = "^(ts|typescript)$"
+injection-regex = "(ts|typescript)"
 file-types = ["ts"]
 shebangs = []
 roots = []
@@ -385,7 +385,7 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-typescript", rev = 
 [[language]]
 name = "tsx"
 scope = "source.tsx"
-injection-regex = "^(tsx)$" # |typescript
+injection-regex = "(tsx)" # |typescript
 file-types = ["tsx"]
 roots = []
 # TODO: highlights-params
@@ -1183,7 +1183,7 @@ source = { git = "https://github.com/milisims/tree-sitter-org", rev = "698bb1a34
 [[language]]
 name = "solidity"
 scope = "source.sol"
-injection-regex = "^(sol|solidity)$"
+injection-regex = "(sol|solidity)"
 file-types = ["sol"]
 roots = []
 comment-token = "//"
@@ -1400,7 +1400,7 @@ source = { git = "https://github.com/archseer/tree-sitter-cairo", rev = "5155c6e
 [[language]]
 name = "cpon"
 scope = "scope.cpon"
-injection-regex = "^cpon$"
+injection-regex = "cpon"
 file-types = ["cpon", "cp"]
 roots = []
 auto-format = true


nrdxp:
> Unfortunately queries don't have a way to express arbitrary nesting

Well I guess that's why I couldn't figure it out then. I wonder then, if it might be better to try and parse the first line of an indented string and parse the shebang (if one exists). Not sure if there is a way to grab the first line easily though :thinking:

the-mikedavis:
Sub-selecting in nodes is also not really well supported by tree-sitter queries 😅

For injection in particular it's best to have the language injection token expressed by the parser - for example some node under string or comment that contains the first word. Then you could capture that as `@injection.content`. Without that we would need to add some predicates to sub-select nodes. Neovim has some of those if I remember correctly but I'm not sure how I feel about them. To me it seems better to always have the parser express things you want to capture with queries

nrdxp:
I'm moving out of draft as I have been able to accomplish injections in many common places in Nix where shellcode is typically expected. The comment based injection still works as well, and if you are okay with me modifying other language injection regex's I can do that to support more languages, but many already work as is.

nrdxp:
I opened a PR upstream to add these same queries there, so maybe we don't need to merge this, but just [uncomment](https://github.com/cstrahan/tree-sitter-nix/blob/master/bindings/rust/lib.rs#L39) the injections in the upstream binding if it is accepted?

runtime/queries/nix/injections.scm
```diff
@@ -0,0 +1,48 @@
...
+(binding
+  attrpath: (attrpath (identifier) @_path
+    (#match? @_path "(^\\w*Phase|(pre|post)\\w*|(.*\\.)?\\w*([sS]cript|[hH]ook)|(.*\\.)?startup)$"))
+  expression: [
+    (indented_string_expression (string_fragment) @injection.content
+      (#set! injection.language "bash"))
+    (_ (_)* (indented_string_expression (string_fragment) @injection.content
+      (#set! injection.language "bash")))
+    (_ (_)* (_ (_)* (indented_string_expression (string_fragment) @injection.content
+      (#set! injection.language "bash"))))])
```
the-mikedavis, line 19: Properties (`#match?`, `#set!`, etc) apply to the entire match - it doesn't matter where in the match they're specified. So style-wise, stanzas tend to put them at the bottom at the outer-most level of nesting. This one would be:

```suggestion
((binding
   attrpath: (attrpath (identifier) @_path)
   expression: [
     (indented_string_expression (string_fragment) @injection.content)
     (_ (_)* (indented_string_expression (string_fragment) @injection.content))
     (_ (_)* (_ (_)* (indented_string_expression (string_fragment) @injection.content)))
   ])
 (#match? @_path "(^\\w*Phase|(pre|post)\\w*|(.*\\.)?\\w*([sS]cript|[hH]ook)|(.*\\.)?startup)$")
 (#set! injection.language "bash"))
```

runtime/queries/nix/injections.scm
```diff
@@ -0,0 +1,48 @@
...
+
+(apply_expression
+  function: (apply_expression
+    function: ((_) @_func
+      (#match? @_func "(^|\\.)writeShellScript(Bin)?$")))
+  argument: (indented_string_expression (string_fragment) @injection.content
+    (#set! injection.language "bash")))
```
the-mikedavis, line 26:
```suggestion
((apply_expression
   function: (apply_expression function: (_) @_func)
   argument: (indented_string_expression (string_fragment) @injection.content))
 (#match? @_func "(^|\\.)writeShellScript(Bin)?$")
 (#set! injection.language "bash"))
```

And similar for the ones below

the-mikedavis:
> I opened a PR upstream to add these same queries there, so maybe we don't need to merge this, but just [uncomment](https://github.com/cstrahan/tree-sitter-nix/blob/master/bindings/rust/lib.rs?rgh-link-date=2022-09-01T17%3A28%3A04Z#L39) the injections in the upstream binding if it is accepted?

We don't use rust bindings or inherit the queries from the repo. It's good to upstream the changes so that other tree-sitter consumers can get the benefits too but for helix the queries only need to be updated in-repo

runtime/queries/nix/injections.scm
```diff
@@ -0,0 +1,46 @@
...
+  argument: (_ (_)* (_ (_)* (binding
+    attrpath: (attrpath (identifier) @_path
+      (#match? @_path "^text$"))
+    expression: [
+      (indented_string_expression (string_fragment) @injection.content
+        (#set! injection.language "bash"))
+      (_ (_)* (indented_string_expression (string_fragment) @injection.content
+        (#set! injection.language "bash")))
+      (_ (_)* (_ (_)* (indented_string_expression (string_fragment) @injection.content
+        (#set! injection.language "bash"))))]))))
```
the-mikedavis, line 46: These two can have their `#match?` and `#set!` predicates moved to the outside of the match too, like the above clauses

nrdxp:
So I've identified the reason for the bash quoting issue described above, and it should have been obvious but I guess I wasn't thinking correctly :sweat_smile: . 

Essentially the two strings on either side of an interpolation are highlighted seperately, so instead of having a balanced quote count, the string on the right side thinks it starts with an opening `"`, instead of matching with the `"` in the previous fragment. Not sure if there is a way to tell tree-sitter to concatenate these strings together before applying highlighting, but that'd be the best solution for that case. The other option is to just not enclose an interpolation with quotes, but there is already a lot of code in nixpkgs which does this so it doesn't help while reading, and also doesn't help if we want to avoid worksplitting in the actual script.

Also found another issue outline in cstrahan/tree-sitter-nix#32 but that will have to be worked out upstream.

runtime/queries/nix/injections.scm
```diff
@@ -0,0 +1,45 @@
+; mark arbitary languages with a comment
+(
+  ((comment) @injection.language) . [
+    (indented_string_expression (_) @injection.content)
+    (_ (_)* (indented_string_expression (_) @injection.content))
+    (_ (_)* (_ (_)* (indented_string_expression (_) @injection.content)))
+    (_ (_)* (_ (_)* (_ (_)* (indented_string_expression (_) @injection.content))))
+    (_ (_)* (_ (_)* (_ (_)* (_ (_)* (indented_string_expression (_) @injection.content)))))])
```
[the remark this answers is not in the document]
nrdxp: yeah I noticed this as well, I suppose being more explicit is good in this case, perhaps we should add a comment _after_ the string as well, for where it might look better stylistically.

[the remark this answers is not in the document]
nrdxp: done

nrdxp:
Turns out `injection.combined` is what I need to avoid the quoting imbalance, and I simplified the queries to remove any possibility for nesting. If users want to put arbitrary nested expressions in `buildPhase` and the like, they can just use the comment style to mark where the bash script starts. I think this is fine since 90% of the time, you just immediatley define the string.

the-mikedavis (approved):
These look super sweet, thanks for working on this!

mmlb:
This is pretty nice, thanks for adding it in. I'm just messing around with this and it seems a little clunky because it seems like the multi-line string needs to be the first non-whitespace chars on the line right?

I've tried the following and both work:

```nix
k = #shell
''exec ${kubectl}/bin/kubectl "$@"'';
```
and
```nix
k = 
#shell
''
exec ${kubectl}/bin/kubectl "$@"
'';
```

It'd be nice if the following worked too
```nix
#shell
k = ''exec ${kubectl}/bin/kubectl "$@"'';
```
Would that be possible with tree-sitter? (I am not versed in tree-sitter in the slightest :( ).

mmlb:
Actually I think something might be funky with the current code, I just noticed this when writing out the previous comment:

```nix
k = #shell
''exec ${kubectl}/bin/kubectl "$@"'';
kk = #shell
''exec ${kubectl}/bin/kubectl "$@"'';
```

the string for `k` highlights `exec` nicely, but `kk`'s is not. Should I open this as a bug?

nrdxp:
@mmlb I ran into this myself when writing the queries. Tree-sitter queries do not have a way to express arbitrary nesting, probably because it would lead to many false positives. I tried to work around that by simple mimicing queries multiple times inside a many nested levels and we hit some false positives ourselves that way so we decided to leave it explicit.

In order to make this as nice as possible though, there are certain names for `k` that will automatically consider the value a bash script. For example, you could use `script` or `command` and the value will highlight as a bash string without requiring the `#shell` comment marker.

nrdxp:
> Ah I see, I figured it would be something along those lines, +1. Thanks for doing this really loving this! heart

Glad you find it useful, if we can also get #3970 merged then we can also do highlights based on shebang or even file name in expressions like `writeScript "foo.sh" "echo foo"`, which would highlight as bash based on the `.sh` in the first arg.

PR #3587, opened 2022-08-28
Themes update

AlexanderBrevig:
This is a collection of changes made to force themes to handle in a predictable way. It also enforces a response to enabling features (like `cursorline`) so users who want these features will still see them even if they change themes.

I've tried my best to stick to the spirit of the theme, but some themes relied so much on terminal colors that I just had to guess and base some assumptions from other editors with similar themes.

Status from original authors (in terms of approvals)
- [x] snazzy @nrdxp 
- [ ] pop-dark @workingj 
- [x] penumbra @vlmutolo 
- [ ] noctis_bordo @jharrilim 
- [x] ingrid @intarga 
- [x] flatwhite @AlexanderBrevig 
- [x] doom_acario_dark @Rinnray 
- [x] dark_plus @kirawi 
- [x] boo_berry @bootradev 
- [x] bogster @vv9k 
- [x] everforest @CptPotato @WindSoilder 
- [ ] acme @two-six 
- [x] serika @VuiMuich 
- [ ] nord @raygervais @two-six 
- [x] dracula @samsartor 
- [ ] gruvbox @crodjer @jbaa 
- [x] autumn @getreu 
- [ ] ayu @enkodr 
- [x] rose_pine @raygervais @chunghha @atahabaki 
- [x] monokai @WindSoilder 
- [ ] solarized @kirawi @cossonleo 
- [x] spacebones @atog

PR #3599, opened 2022-08-29
fix! render highlights over rulers

AlexanderBrevig:
Bug pointed out here: https://github.com/helix-editor/helix/pull/3587#issuecomment-1230525446

This fix is so easy I'm thinking I may have missed something?
EDIT: Ref ChrHorn below; themes that use `ui.ruler.fg` will not work anymore.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -128,6 +128,7 @@ impl EditorView {
             Box::new(highlights)
         };
 
+        Self::render_rulers(editor, doc, view, inner, surface, theme);
         Self::render_text_highlights(
             doc,
             view.offset,
@@ -138,7 +139,6 @@ impl EditorView {
             &editor.config(),
         );
         Self::render_gutter(editor, doc, view, view.area, surface, theme, is_focused);
-        Self::render_rulers(editor, doc, view, inner, surface, theme);
 
         if is_focused {
             Self::render_focused_view_elements(view, doc, inner, theme, surface);


ChrHorn:
Was about to do the same.

With this fix we loose the ability to set `fg` for the ruler. Although currently only the solarized themes and base16_transparent use it explicitly. Other themes might only set `ui.virtual.fg` so they also would have to be changed, this includes the base theme.

AlexanderBrevig:
> Was about to do the same.
> 
> With this fix we loose the ability to set `fg` for the ruler. Although currently only the solarized themes and base16_transparent use it explicitly. Other themes might only set `ui.virtual.fg` so they also would have to be changed, this includes the base theme.

Spot on!
This is a regression for themes that use `fg` for rulers.
If we want to support it I guess the cleanest option is to extract cursor from general highlights, and draw it last?

luetage:
I stumbled upon the same thing earlier and found the same flawed fix. One thing to note is that with the current code, themes which set the primary cursor color modifiers to »reversed« are not affected and don’t need the fix. But the majority of themes does. I wonder whether it’s even worth supporting a foreground color for rulers, as mentioned few themes use it and traditionally a colorcolumn/ruler is just that, a change in background color. Might be easier to just fix the few themes using it than to work around it. Another thing is that cursorline and cursorcolumn “disappear” behind the ruler, don’t know whether that’s on purpose and what was intended originally.

joshbainbridge:
My two cents, this PR might well fix more themes than it breaks. Currently the character under the curser disappears entirely, or becomes hard to read, under many themes (see for example gruvbox). This is because when just the ruler background is set, it can often match or be very close to the cursor foreground colour.

luetage:
@joshbainbridge This should just be merged. It’s been 7 months now and there is no reason this bug should endure.

pascalkuthe:
clsoing this PR as stale, the rendering code has been entirely replaced. The text rendingr has been split into base and overlay highlights. The rulers should be rendered after the base highlights but before the overlay highlights.

Thank you for contributing!

PR #3607, opened 2022-08-30
Fix fleetish theme

krfl:
I made two mistakes in my previous pr. This fixes those issues and there are no more errors logged. 
1. The theme used an undefined accent color
2. string.special had a modifier that was spelled wrong and the word modifiers was not plural.


AlexanderBrevig:
Maybe the linter #3234 should handle this?

the-mikedavis (approved):

PR #3610, opened 2022-08-30
theme: Add Varua color scheme

kamek-pf:
Porting the color palette I've been using for a couple years before moving to Helix.
Cheers :)

### DIFF
--- a/runtime/themes/varua.toml
+++ b/runtime/themes/varua.toml
@@ -45,6 +45,7 @@
 "diff.minus" = "red"
 
 "ui.background" = { bg = "bg0" }
+"ui.cursorline.primary" = { bg = "bg0" }
 "ui.cursor" = { fg = "bg0", bg = "fg" }
 "ui.cursor.match" = { fg = "orange", bg = "bg_yellow" }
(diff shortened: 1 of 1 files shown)


AlexanderBrevig:
Can you please add `"ui.cursorline.primary" = { bg = "something" }`
Which is a feature available in the next version. Replace something with what you want.

kamek-pf:
Done ! I typically don't use a different color for the cursorline so I used the same color as the background. But it is defined now.

AlexanderBrevig:
It will feel like a bug for people who have enabled cursorline if it does not show up though. I think themes should provide the features Helix allows for. Cursorline is disabled by default so you won't see it unless you enable it, at which point - I think you should see it. :)

kamek-pf:
Yeah that makes sense, I picked a different color, it should look nice, I'll probably give the new feature a try :p

the-mikedavis (approved):
Looks nice, thanks! 🎨

the-mikedavis:

kamek-pf:
Screenshot was added. Thanks !

PR #3621, opened 2022-08-31
update zig grammar and highlight rules

voroskoi:


runtime/queries/zig/highlights.scm
```diff
@@ -125,44 +119,53 @@ field_constant: (IDENTIFIER) @constant
...
-  "const"
-  "var"
   "test"
+] @keyword
+
+[
   "pub"
   "usingnamespace"
-] @keyword
+] @namespace
```
the-mikedavis, line 161: typically `@namespace` is for the namespace names themselves. `pub` is `@keyword` in the rust highlights. `usingnamespace` might be better as `@keyword.control.import` although I'm not familiar with Zig so I'm not sure that's appropriate

voroskoi: I was not sure about it either, but Your suggestion makes sense. However I would leave `pub` also in `@keyword.control.import` as its a keyword for controlling what can/can not be imported.

the-mikedavis (approved):
Looks great, thank you!

PR #3625, opened 2022-09-01
fix: typo in tutor about searching compared to vim

CobaltCause:
I'm pretty sure this is what was originally intended.


the-mikedavis (approved):

the-mikedavis:
Good catch, thanks!

PR #3631, opened 2022-09-01
Add support for the Godot Language Server

Jummit:
By default Godot 3 uses port 6008, but this will be changed to 6005 in Godot 4.

Launch command copied from https://github.com/neovim/nvim-lspconfig/blob/master/lua/lspconfig/server_configurations/gdscript.lua

Users who use Godot 4 right now need to change the port in the editor settings.

This also changes the indent to tabs, based on the [official style guide](https://docs.godotengine.org/en/stable/tutorials/scripting/gdscript/gdscript_styleguide.html#encoding-and-special-characters).

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1533,6 +1533,8 @@ roots = ["project.godot"]
 auto-format = true
 comment-token = "#"
 indent = { tab-width = 4, unit = "\t" }
+# Change this to port 6005 for Godot 4.
+language-server = { command = "nc", args = ["localhost", "6008"] }
 
 [[grammar]]
 name = "gdscript"
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -38,7 +38,7 @@
 | esdl | ✓ |  |  |  |
 | fish | ✓ | ✓ | ✓ |  |
 | fortran | ✓ |  | ✓ | `fortls` |
-| gdscript | ✓ | ✓ | ✓ |  |
+| gdscript | ✓ | ✓ | ✓ | `nc` |
 | git-attributes | ✓ |  |  |  |
 | git-commit | ✓ |  |  |  |
 | git-config | ✓ |  |  |  |


languages.toml
```diff
@@ -1713,3 +1713,13 @@ language-server = { command = "pasls", args = [] }
...
+
+[[language]]
+name = "gdscript"
+scope = "source.gdscript"
+injection-regex = "gdscript"
+file-types = ["gd", "gdscript", "gdscript3", "gdscript4"]
+comment-token = "#"
+indent = { tab-width = 2, unit = "	" }
+language-server = { command = "nc", args = ["localhost", "6008"] }
+# Change this to port 6005 once Godot 4 is released.
```
archseer, line 1725: This seems to duplicate an existing grammar definition. Just edit the one on line 1321

languages.toml
```diff
@@ -1321,12 +1321,13 @@ source = { git = "https://github.com/DerekStride/tree-sitter-sql", rev = "0caa7f
...
-file-types = ["gd"]
-shebangs = []
+file-types = ["gd", "gdscript", "gdscript3", "gdscript4"]
 roots = ["project.godot"]
 auto-format = true
 comment-token = "#"
-indent = { tab-width = 4, unit = "    " }
+indent = { tab-width = 4, unit = "	" }
+# Change this to port 6005 once Godot 4 is released.
+language-server = { command = "nc", args = ["localhost", "6008"] }
```
the-mikedavis, line 1330: I'm not sure we should use `nc` for the language server in the main `languages.toml` config. `nc` is just a hack around the language server not supporting stdio and I'm not sure windows has `nc` by default. Instead I think this configuration should be listed in https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

languages.toml
```diff
@@ -1321,12 +1321,13 @@ source = { git = "https://github.com/DerekStride/tree-sitter-sql", rev = "0caa7f
...
 scope = "source.gdscript"
 injection-regex = "gdscript"
-file-types = ["gd"]
-shebangs = []
+file-types = ["gd", "gdscript", "gdscript3", "gdscript4"]
 roots = ["project.godot"]
 auto-format = true
 comment-token = "#"
-indent = { tab-width = 4, unit = "    " }
+indent = { tab-width = 4, unit = "	" }
```
the-mikedavis, line 1328:
```suggestion
indent = { tab-width = 4, unit = "\t" }
```

languages.toml
Jummit: Well, there is a note about netcat at the top of the wiki page. Should this be closed then?

the-mikedavis: The file-types and indent changes are still good so I wouldn't say this should be closed. Actually we may want to merge this as-is since - if I understand correctly - the godot LSP is built into the godot editor so there's no chance of being able to run by stdio unless godot decides to split out the LSP into its own executable/project. What do you think @archseer?

coelhucas:
Is there anything else for this to be merged rather the conflicts, @Jummit @archseer?

iver0:
Yeah, this still works and is still needed.

the-mikedavis:
Merged in 45c2b3ccbcc8ef5574f9645bb5a876140ec96a7f

PR #3636, opened 2022-09-01
Add: version command to show build info

shassard:
Add a trivial `version` command to interactively print the build version and hash to the status bar.


the-mikedavis:
You can do this with `:sh hx --version`

shassard:
That would spawn an external process and show you the version from the helix binary in your path rather than from the current instance that's currently running interactively though, right?

It felt to me like this command was missing given that it's available in all vi implementations and was pretty trivial to implement.

PR #3642, opened 2022-09-01
Ability to use `$file` and `$pwd` in shell commands and formatter args

AlexanderBrevig:
This is a work in progress, to satisfy my curiosity about a KISS way of going about it.

This will enable people to use `$file` and `$pwd` for `insert-output` and formatter command args (the latter is what prompted me to test this idea).

- [x] make use of Env in more places (all the shell interactions)
- [x] find issues and PRs to link to - I know there is pre-existing work on this
- [ ] figure out if the declarative approach is OK, or if we want something a bit more "sticky"


AlexanderBrevig:
Alternative to #3393 , the other PR probably matches better but will be harder to reuse for things like LSP config?

helix-view/src/document.rs
```diff
@@ -638,6 +639,10 @@ impl Document {
         Ok(())
     }
 
+    pub fn get_path(&self) -> Option<PathBuf> {
+        self.path.clone()
+    }
```
the-mikedavis, line 644: It's better to let the caller decide when/if to clone:

```suggestion
    pub fn path(&self) -> Option<&PathBuf> {
        self.path.as_ref()
    }
```

Then `Env` should use `doc.path().cloned()`.

helix-view/src/env.rs
```diff
@@ -0,0 +1,37 @@
...
+                    self.path
+                        .clone()
+                        .unwrap_or_default()
+                        .to_str()
+                        .unwrap_or_default(),
+                ),
+            );
+        });
+        log::debug!("injected: {:#?}", vec);
+        vec
```
the-mikedavis, line 35: The path &str should be calculated outside of the loop. It might make sense to do this when initializing the `Env` instead.

The `for_each` can be rewritten as a map+collect:

```suggestion
        let path = self
            .path
            .as_ref()
            .and_then(|p| p.to_str())
            .unwrap_or_default();

        strs.map(|s| s.to_string().replace("$file", path)).collect()
```

helix-view/src/env.rs
```diff
@@ -0,0 +1,37 @@
+use std::{fmt::Display, path::PathBuf};
+
+use crate::Document;
+pub struct Env {
```
the-mikedavis, line 4: This type can be eliminated: replacing the environment in a string can be done as a single function rather than a type+method.

helix-view/src/handlers/dap.rs
```diff
@@ -300,8 +301,9 @@ impl Editor {
                         None => return false,
                     };
 
+                    let (_, doc) = current!(self);
```
the-mikedavis, line 304:
```suggestion
                    let doc = doc!(self);
```

helix-term/src/commands/typed.rs
```diff
@@ -1484,9 +1484,10 @@ fn run_shell_command(
     if event != PromptEvent::Validate {
         return Ok(());
     }
-
     let shell = &cx.editor.config().shell;
-    let (output, success) = shell_impl(shell, &args.join(" "), None)?;
+    let (_, doc) = current!(cx.editor);
```
the-mikedavis, line 1488:
```suggestion
    let doc = doc!(cx.editor);
```

helix-term/src/commands.rs
```diff
@@ -4692,7 +4703,13 @@ fn shell_prompt(cx: &mut Context, prompt: Cow<'static, str>, behavior: ShellBeha
...
             }
 
-            shell(cx, input, &behavior);
+            let inp = input
+                .split(' ')
+                .collect::<Vec<&str>>()
+                .iter()
+                .map(|a| Cow::from(*a))
+                .collect::<Vec<Cow<str>>>();
+            shell(cx, &inp, &behavior);
```
the-mikedavis, line 4712:
```suggestion
            let input = input.split(' ').map(Cow::from).collect::<Vec<_>>();
            shell(cx, &input, &behavior);
```

AlexanderBrevig: That's amazing! TIL

helix-view/src/env.rs
AlexanderBrevig: I've converted to this style now. I opted for `inject_into_with(iterator, doc)` another way to go could be `inject_with_into(doc, iterator)` which may be better?

helix-view/src/env.rs
```diff
@@ -0,0 +1,13 @@
+use std::fmt::Display;
+
+use crate::Document;
+
+pub fn inject_into_with<T>(strs: T, doc: &Document) -> Vec<String>
```
the-mikedavis, line 5: The name is a bit awkward. I think something like `inject_envorinment` may fit better

helix-view/src/env.rs
the-mikedavis: I prefer `(iterator, doc)`

helix-view/src/env.rs
```diff
@@ -0,0 +1,13 @@
...
+
+pub fn inject_environment<T>(strs: T, doc: &Document) -> Vec<String>
+where
+    T: Iterator,
+    <T as Iterator>::Item: Display,
+{
+    let path = doc.path().and_then(|p| p.to_str()).unwrap_or_default();
+
+    strs.map(|s| s.to_string().replace("$file", path)).collect()
+}
```
the-mikedavis, line 13: I'm not sure this is possible because of lifetimes (trait items need static lifetimes when they are return values) but it would be nice if this returned something implementing `IntoIterator`, to prevent collecting since [`Command::args`](https://doc.rust-lang.org/std/process/struct.Command.html#method.args) takes an IntoIterator. It might work to return a `impl Iterator<String>` and remove the `.collect()`?

AlexanderBrevig:
```patch
diff --git a/helix-term/src/commands.rs b/helix-term/src/commands.rs
index b943d6d4..4a2a48f2 100644
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4586,12 +4586,10 @@ fn shell_impl(
     use std::process::{Command, Stdio};
     ensure!(!shell.is_empty(), "No shell set");
 
-    let cmd = inject_environment(args.iter(), doc).join(" ");
-
     let mut process = Command::new(&shell[0]);
     process
         .args(&shell[1..])
-        .arg(cmd)
+        .args(inject_environment(args.iter(), doc))
         .stdout(Stdio::piped())
         .stderr(Stdio::piped());
 
diff --git a/helix-view/src/env.rs b/helix-view/src/env.rs
index 4db32308..678e26a1 100644
--- a/helix-view/src/env.rs
+++ b/helix-view/src/env.rs
@@ -2,7 +2,10 @@
 
 use crate::Document;
 
-pub fn inject_environment<T>(strs: T, doc: &Document) -> Vec<String>
+pub fn inject_environment<'a, T: 'a>(
+    strs: T,
+    doc: &'a Document,
+) -> impl Iterator<Item = String> + 'a
 where
     T: Iterator,
     <T as Iterator>::Item: Display,
@@ -20,6 +23,6 @@ pub fn inject_environment<T>(strs: T, doc: &Document) -> Vec<String>
             "ENV:NO_PATH".to_string()
         });
 
-    strs.map(|s| s.to_string().replace("$file", path).replace("$pwd", &pwd))
-        .collect()
+    strs.into_iter()
+        .map(move |s| s.to_string().replace("$file", path).replace("$pwd", &pwd))
 }
```
The code compiles, but now the feature no longer works. Ideas? (I'm in way over my head on this one, still learning)

the-mikedavis: The switch from `.arg` to `.args` is what's doing it I believe: the command should be interpreted like `sh -c "echo foo"` but this switches it to `sh -c echo foo` since it's no longer one arg.

Looking at the callsites though I'm not sure returning an iterator saves any work. A Vec might be better here actually since Vec's have `Vec::join` and iterators don't and that's how this is usually consumed.

AlexanderBrevig: Should we close this conversation as resolved then?

helix-view/src/env.rs
```diff
@@ -0,0 +1,25 @@
...
+
+pub fn inject_environment<T>(strs: T, doc: &Document) -> Vec<String>
+where
+    T: Iterator,
+    <T as Iterator>::Item: Display,
+{
+    let path = doc.path().and_then(|p| p.to_str()).unwrap_or_else(|| {
+        log::error!("No $path found for document: {:?}", doc.url());
+        "ENV:NO_PATH"
+    });
```
the-mikedavis, line 13: For an example, scratch buffers won't have paths. In cases where `doc.path().is_none()`, I think we should just skip setting the `$pwd`

the-mikedavis (commented):
Also, instead of string replacement, it may be more elegant to set the env on the command https://doc.rust-lang.org/std/process/struct.Command.html#method.env although I'm not sure if that might make configs non-cross-platform

helix-view/src/env.rs
AlexanderBrevig: I just accidentally did `:sh echo $file` and it felt weird when it did not work (`$pwd` does work in scratch). I also think logs from failing formatters/lsps will be easier to debug if it actually says something rather than nothing.

All that said, I will defer to you.

I agree with moving the function, on it.

the-mikedavis: I think this would be ok at debug level. We probably want to use `unwrap_or_default` though to have it default to the empty string like what happens when an environment variable is unset

AlexanderBrevig: `unwrap_or_default` will not let us log the error though? Maybe simply return a default string is better? At least then we got something to go after when some user runs `hx -v` when we ask for logs. Will be a lot easier to spot than having to know if some argument takes a file or not (in the logs from lsp/formatter/whatever).

the-mikedavis: Oh yeah good point, an `unwrap_or_else` that gives logs and gives an empty string would be good I think 👍

the-mikedavis:
I'm not sure how I feel about this vs #3393-style (kakoune-like) variables. The kakoune-like variables don't look like unix environment variables which is a double edged sword - nice for Windows users but maybe less intuitive for unix users

kirawi:
Addresses https://github.com/helix-editor/helix/pull/3642

PR #3645, opened 2022-09-02
Elixir: update tree-sitter queries

connorlay:
This PR makes the following changes to the Elixir tree-sitter queries:
* tree-sitter text object comments are now grouped together
* markdown is injected into Phoenix Live View declarative attribute `:doc` options

See the following examples:
- https://github.com/phoenixframework/phoenix_live_view/blob/c1c4754e772de529529a046b42f5fdf02c66dede/lib/phoenix_component.ex#L1452
- https://github.com/phoenixframework/phoenix_live_view/blob/c1c4754e772de529529a046b42f5fdf02c66dede/lib/phoenix_component.ex#L1418
- https://github.com/phoenixframework/phoenix_live_view/blob/c1c4754e772de529529a046b42f5fdf02c66dede/test/phoenix_component/declarative_assigns_test.exs#L583

### DIFF
--- a/runtime/queries/elixir/injections.scm
+++ b/runtime/queries/elixir/injections.scm
@@ -1,6 +1,8 @@
+; Elixir Comments
 ((comment) @injection.content
  (#set! injection.language "comment"))
 
+; Elixir Regular Expressions
 ((sigil
   (sigil_name) @_sigil_name
   (quoted_content) @injection.content)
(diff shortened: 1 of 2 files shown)


runtime/queries/elixir/injections.scm
```diff
@@ -23,3 +19,25 @@
...
+ (#set! injection.language "heex")
+ (#set! injection.combined))
+
+; Phoenix Live View Component Macros
+(call 
+  (identifier) @_identifier
+  (arguments
+    (atom)+
+    (keywords (pair 
+      (keyword)
```
the-mikedavis, line 37: It looks like `:doc` is the only option that can take a string for now, but for future-proofing, should we also test that this `(keyword)` node is `#eq?` to "doc"?

connorlay:
@the-mikedavis Thanks for the review! I pushed changes addressing your feedback, plus additional sigil injections for Jason and Zigler.

the-mikedavis (approved):

the-mikedavis:
TIL about Jason's sigil J, that's handy!

PR #3652, opened 2022-09-02
Lsp goto other window (fixes 3167)

dam5h:
I inadvertently closed the [prior PR](https://github.com/helix-editor/helix/pull/3269) with these same commits via some failed git-fu.  

Anyhow, its back up now.

This addresses the [#3167 issue](https://github.com/helix-editor/helix/issues/3167) I made.

For picker related selections I have added a ctrl!('o') keybind in the picker that will load into the "other" window, via adding a new `Action` variant for `editor.switch` to match onto, which just calls `focus_next()` and then `switch()` recursively with `Action::Replace`.

For Goto defintion, since it has a unique destination and no `Picker` interaction, I just add a command that can be accessed directly if someone adds the command to their keymap.  In my personal keymap I bind that to `D` whereas lowercase `d` is goto definition in the same view.


helix-term/src/commands.rs
```diff
@@ -287,6 +287,7 @@ impl MappableCommand {
         select_mode, "Enter selection extend mode",
         exit_select_mode, "Exit selection mode",
         goto_definition, "Goto definition",
+        goto_definition_other_window, "Goto definition other window",
```
dead10ck, line 290: This could be clearer, e.g. "Goto definition in the next pane"

dead10ck (approved):
Implementation-wise, this seems fine to me. The feature may be a bit niche in my opinion, but I think it's reasonably small and wouldn't mind it being merged.

helix-term/src/commands.rs
dam5h: Thanks! Just pushed up that change.

helix-view/src/editor.rs
```diff
@@ -765,6 +765,7 @@ pub enum Action {
     Replace,
     HorizontalSplit,
     VerticalSplit,
+    Other,
```
dead10ck, line 768: Maybe this could have a more descriptive name like `ReplaceInNext`

helix-term/src/commands/lsp.rs
```diff
@@ -911,6 +911,14 @@ fn to_locations(definitions: Option<lsp::GotoDefinitionResponse>) -> Vec<lsp::Lo
 }
 
 pub fn goto_definition(cx: &mut Context) {
+    goto_definition_impl(cx, false)
+}
+
+pub fn goto_definition_other_window(cx: &mut Context) {
```
dead10ck, line 917: Same, this could be `goto_definition_next_view`

dead10ck (approved):
Just took another pass over this, and again, implementation wise, it seems fine. There are a couple of minor naming issues I'd change. Only outstanding items I think would be the binding. Ctrl-o seems weird, but I also can't think of a better one.

dam5h:
Thanks @dead10ck , made those suggested changes.   Cheers!  Dave

helix-term/src/commands/lsp.rs
```diff
@@ -1078,6 +1089,14 @@ pub fn goto_implementation(cx: &mut Context) {
...
 
 pub fn goto_reference(cx: &mut Context) {
+    goto_reference_impl(cx, false)
+}
+
+pub fn goto_reference_other_window(cx: &mut Context) {
+    goto_reference_impl(cx, true)
+}
+
+fn goto_reference_impl(cx: &mut Context, rotate: bool) {
```
bnjbvr, line 1099: nit: `rotate` is unused here (noticed thanks to the Clippy CI step which failed because of this)

bnjbvr (commented):
Just a bystander, but I'm excited about this, thanks for writing this PR!

PR #3657, opened 2022-09-02
build(nix): update nci, refactor flake, seperate wrapping, add source filtering

yusdacra:
Updates NCI, separates wrapping of hx binary to another derivation so that helix doesn't get recompiled when grammars or runtime get changed, adds filtering to source for helix-term build so that unnecessary rebuilds are reduced.


the-mikedavis (approved):

the-mikedavis:
Sweet, thanks for the updates! I built on x86_64-linux and aarch64-macos. I'm looking forward to stress-testing the rebuilding improvements 😀

PR #3661, opened 2022-09-02
nix: fix indent and use alejandra

nrdxp:
Align indents to the way alejandra formats, and format automatically on save to keep the two in sync.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -478,6 +478,8 @@ roots = []
 comment-token = "#"
 language-server = { command = "rnix-lsp" }
 indent = { tab-width = 2, unit = "  " }
+auto-format = true
+formatter = { command = "alejandra", args = ["-q"] }
 
 [[grammar]]
 name = "nix"
--- a/runtime/queries/nix/indents.scm
+++ b/runtime/queries/nix/indents.scm
@@ -1,0 +1,70 @@
+[
+  ; Bracket like
+  (let_expression)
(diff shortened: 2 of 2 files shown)


Triton171:
I've tried it out a bit and it seems to work great. You still need to run `cargo xtask docgen` to fix the failing check. Also, in case you have a test file for Nix, you could add that: Just copy the file to `helix-core/tests/data/indent`, add a `treesitter_indent_nix` function to `helix-core/tests/indent.rs` and copy the Nix language definition to `helix-core/tests/data/indent/languages.toml`. It's not needed to get this PR merged but we currently only have 2 tests for rust so it'd be nice to expand that.

nrdxp:
Thanks for testing, I've got it in draft now because there are still a few inconsistencies. I'm not sure there is an easy way to fix some of the current discrepencancies without adding extending the indent system a bit.

Triton171:
If you have an example of inconsistencies that are difficult to fix using the current system, it'd be interesting to see it. The current indentation system is far from perfect so if some problem occurs often/in multiple languages, it would be nice to add indentation captures for it. If it's a rare case, regexes can sometimes be used as a workaround even though this isn't ideal (see for example the queries for rust).

pascalkuthe:
clsoing this one as stale as its unfinished and has conflicts. Thank you for contributing!

PR #3670, opened 2022-09-03
Look for the external formatter before invoking it

the-mikedavis:
Currently it is not possible to save a file with a language that
has an external formatter configuration unless the external
formatter is installed, even if the language has a Language Server
configuration capable of auto-format. This change checks that the
external formatter exists before using it to create a formatting
callback.


the-mikedavis:
I think this change stands on its own but we may want to refactor the auto-format callback structure: a failure to format results in a failure to write currently which I think is unexpected.

sudormrfbin (approved):

Issue #1583, opened 2022-01-26
More granular undo


eugenesvk:
Currently `undo` seems to undo all the changes per an "Insert session"
So, this
`insert_mode`
type a lot of text
`normal_mode`
`undo`

deletes "type a lot of text"

This is too coarse, especially when typing a lot of text in one go.
Is there a way to increase the undo granularity to delete, say, by typed word? So it would behave more like this:
1st `undo` deletes "text"
2nd `undo` deletes "of "
3rd `undo` deletes "lot "
...

(can be a separate command or an additional argument like `:undo 1w` or something)

Jedsek:
I agree what you say. 
vim has the same behaviour(maybe a place for improvement?), but  `<n> wd` increase the undo in both hx and vim.

Jedsek:
See  #1142

Omnikar:
This issue should probably be closed as a duplicate.

Omnikar:
Oh, true.

pickfire:
What about implementing save-based undo? Like vim `:ear 1f`?

archseer:
`Ctrl-s` will be added so that you can commit undo state while in insert mode. I'm not going to add anything more granular, preciseness is a feature to me (similar to how git commit doesn't auto commit after every file change).

eugenesvk:
> What about implementing save-based undo? Like vim `:ear 1f`?

It's a good command that's worth implementing (and I like it better than undos based on time as I can have a better mental model of saving after some logical chunk of work rather than trying to guess how much time that chunk of work took), but it operates on a different "time frame": it's suitable to undo big chunks of edits you've committed to a file, but unsuitable for smaller changes of the last few words/sentences/edits.
These "undo a few latest changes" operations should require no pre-thought/planning, so you shouldn't need to save on every word or something to make them work

eugenesvk:
> similar to how git commit doesn't auto commit after every file change).

and a text editor isn't git, so not sure how that's relevant, we'd better use text editing experience from other apps for comparison.
The workflow you describe can be approximated with file saves and undos based on a number of saves, and with a "hot exit/persistent undo" feature implemented where you wouldn't need to save a file to avoid losing your changes, file saves would be an even better fit for this workflow.
But to require introducing this whole new mental model not required elsewhere just to be able to undo a few last edits is an overkill, these types of operations shouldn't require any preliminary planning

archseer:
> we'd better use text editing experience from other apps for comparison.

[...]

PR #3671, opened 2022-09-03
Remove default insert mode movement bindings

dead10ck:
Helix is first and foremost a modal editor. Willingness to support non-modal editing is there, but it is not one that should be encouraged with the default settings. There are an increasing number of users who are stumbling because they are trying to use Helix as a non-modal editor, so this is an effort to encourage new users to stop and take notice that Helix has a different paradigm than VSCode, Sublime, etc. Users can still add these bindings back to their own configs if they wish.


DannyJJK (approved):
I agree with this, this should be a user-friendly modal editor, and it's not user-friendly to allow people to do something that is counter-intuitive to how a modal editor works and conflicts with other things like autocomplete suggestions that also use the arrow keys in insert mode. I always found it strange that Vim allowed you to use the arrow keys in insert mode, I don't think that should be repeated here

eulerdisk:
I dont't agree with this.
Those keybinding are pretty basic movement and provide limited funcionality. Thery are consistent with "standard" readline keybindings found everywhere from REPL to other text editors, so most people expect to be there by default.

archseer:
Also needs to update https://github.com/helix-editor/helix/blob/master/book/src/keymap.md#insert-mode

(I'd also remove the `Ctrl-n`/`-p` doc since those two were previously removed)

dead10ck:
Done

the-mikedavis (approved):

mangas:
Why not introduce a configure people can remove instead of doing this dramatic change to how the editor behaves? This increases the barrier of entry and forces people to add a whole configuration if they like the normal behaviour (in all apps and editors) being a modal editor should be something you take advantage of and not an imposition. As a power user it is up to you to turn off what you don't like or just not use it. IMO it doesn't make any sense to remove this for everybody based on your personal view of what a modal/non-modal editor should behave like.

mangas:
As you can see in the linked issue above https://github.com/helix-editor/helix/issues/3762, people will likely not know this can be turned on and off. I would suggest reverting this change and introducing a configuration that turns off navigation for yourself rather than forcing everyone into this view of navigation

YangtseSu:
> As you can see in the linked issue above #3762, people will likely not know this can be turned on and off. I would suggest reverting this change and introducing a configuration that turns off navigation for yourself rather than forcing everyone into this opinionated view of what navigation _should_ be

Maybe turn off for default, then a configuration that turns on navigation.

dead10ck:
> As you can see in the linked issue above #3762, people will likely not know this can be turned on and off. I would suggest reverting this change and introducing a configuration that turns off navigation for yourself rather than forcing everyone into this opinionated view of what navigation _should_ be

This isn't just a case of helix being able to do both equally well and us forcing our opinions on everyone. Helix is designed to work well with modal editing and not non-modal editing. The latter causes a bad experience for many who try to force helix to be what it isn't. We get many complaints from new users trying to use helix in a way that was never intended. It's better not to mislead new users into thinking they can use helix like VSCode or Sublime when they really can't.

mangas:
> > As you can see in the linked issue above #3762, people will likely not know this can be turned on and off. I would suggest reverting this change and introducing a configuration that turns off navigation for yourself rather than forcing everyone into this opinionated view of what navigation _should_ be
> 
> This isn't just a case of helix being able to do both equally well and us forcing our opinions on everyone. Helix is designed to work well with modal editing and not non-modal editing. The latter causes a bad experience for many who try to force helix to be what it isn't. We get many complaints from new users trying to use helix in a way that was never intended. It's better not to mislead new users into thinking they can use helix like VSCode or Sublime when they really can't.

Do you have some examples of those complaints? Can you also elaborate on why it's not ready for non modal editing? Also I think we're talking about navigate in insert mode and not editing from outside insert mode right? I'm struggling to understand the underlying problem, this feels like a heavy handed solution that affects barrier to entry, that's why I would like to understand the issue better, maybe there is a more subtle solution to the problem?

dead10ck:
@mangas a non-exclusive list:

* https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126
* https://github.com/helix-editor/helix/issues/1381
* https://github.com/helix-editor/helix/issues/2409
* https://github.com/helix-editor/helix/issues/1583#issuecomment-1022812303

I've lost count of the number of users running into these issues or some variation and bringing it up in the matrix room. [This comment](https://github.com/helix-editor/helix/pull/3366#issuecomment-1213916264) lays out a reasonable roadmap to good support for modeless editing, though it is a very nontrivial effort just to figure out the UX and expected behavior, let alone to implement it.

I think I've made clear in other comments that personally I don't think it's worth the additional long term maintenance burden to try to retroactively change the internal mechanics to support both ways of editing, but if someone or some group of people put forth the effort and came up with something that worked well, didn't hinder the default kakoune-like experience, and didn't add significant maintenance burden, then the main author @archseer has expressed support for this.

But I think until that happens, we shouldn't mislead the expectations of new users, as this will just lead to more frustration for everyone involved.

mangas:
> @mangas a non-exhaustive list:
> 
> * [Fix :w in insert mode #2883 (review)](https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126)
> * [autocomplete triggers in insert mode #1381](https://github.com/helix-editor/helix/issues/1381)
> * [Undoing change in insert mode does nothing when first opening the file #2409](https://github.com/helix-editor/helix/issues/2409)
> * [More granular undo #1583 (comment)](https://github.com/helix-editor/helix/issues/1583#issuecomment-1022812303)
> 
> I've lost count of the number of users running into these issues or some variation and bringing it up in the matrix room. [This comment](https://github.com/helix-editor/helix/pull/3366#issuecomment-1213916264) lays out a reasonable roadmap to good support for modeless editing, though it is a very nontrivial effort just to figure out the UX and expected behavior, let alone to implement it.
> 
> I think I've made clear in other comments that personally I don't think it's worth the additional long term maintenance burden to try to retroactively change the internal mechanics to support both ways of editing, but if someone or some group of people put forth the effort and came up with something that worked well, didn't hinder the default kakoune-like experience, and didn't add significant maintenance burden, then the main author @archseer has expressed support for this.
> 
> But I think until that happens, we shouldn't mislead the expectations of new users, as this will just lead to more frustration for everyone involved.

Just reading through those, the only one stands out to me is the autocomplete one as being related to navigation while in insert mode. I do see the point you are making on the others as they seem mostly wrong usage (undo, save, on insert). None of that seems to really be directly related to navigation. Perhaps adjusting binds to not support commands (:w) or undo would be a solution without removing navigation by default? Thoughts?

dead10ck:
@mangas it's not necessarily that navigation in insert mode in and of itself is buggy (though that auto complete pop-up is one), but that providing users a way to move around with arrows in insert mode by default sets up the experience for new users coming from modeless editors to fall into the trap of thinking they can just do that they've done in other editors, thus users trying things like binding `Ctrl-S` to `:write` to imitate the workflow from other editors. This inevitably leads them to other issues.

I understand it's inconvenient to have this break when it worked before and to have to add these bindings back to your config (I had to add them back to my own config too), but this really is just about setting expectations for new users that are not coming from other modal editors that they will not just be able to keep doing what they're used to and still have a good experience. At least until other bigger roadblocks to that are fixed.

mangas:
I really disagree with supporting modeless editing like those global bindings as it defeats a lot of the purpose of this type of editor but I also don't think removing these bindings will do a lot for adoption. It doesn't really solve the issue, eventually someone will be sharing the old config that we've added back and people will do the same. This seems to be a technical solution to an education/people problem IMO and that's why I wanted to raise it.

dead10ck:
I fully agree with you, I don't think it will be in the best interest of the project to try to be everything for everyone; I worry doing a bunch of post-facto engineering to staple on modeless editing will make it harder to support both ways of editing. This could very well be selection bias, as this is the only editor I've ever been closely involved with in the code and the community, but it seems to me that for whatever reason, helix is drawing a lot of folks from non-modal editors who want it to support the workflow they're used to, and get surprised, frustrated, and demanding when it doesn't work well, even though the project has never advertised itself as anything but modal. I've never seen such complaints about vim or kakoune.

And I think you're absolutely right that this won't truly solve the problem of users trying to make helix what it is not, but my thinking was that for the users who just download and start trying it without reading any docs, it would at least make them aware of this right up front, instead of trying it out for a little bit with some success, but then hitting one of these other issues and going straight to the issue tracker or chat room to report a "bug". The project has quite a lot of legitimate work that needs tracking, and having users ask for the thousandth time why undo isn't working just makes it more difficult for the maintainers to keep up with triaging, which is already a tedious and thankless task. If the user wants to go ahead with using helix in a way that it was not intended, they can make a conscious effort to do so.

My expectation is that this will cause a temporary uptick of complaints from users who were used to using their arrow keys, but my hope is that it will help over time by reducing the number of repeat complaints by making it clear up front that you should not be staying in insert mode 100% of the time.

I could be wrong. But if that's the case, this change need not be permanent.

barszcz:
While I understand the desire to emphasize modal editing and discourage staying in insert mode more than is necessary, I worry that disabling arrow keys by default is extreme. Consider typing a new line containing a matched pair of delimiters, like `"key": value`. The user would begin by typing a double-quote, followed by `key`. Now the cursor is "stuck" behind the second quotation mark (which Helix matched for us, like any well-behaved modern editor); for the user to continue typing the line, they have no choice but to exit insert mode, move a single character to the right, and then reenter insert mode. Previously, this could have been a single press of the right arrow key, which is how I've dealt with this for years, in both modal and non-modal editors (or with the tab key, if the editor allowed for that). This is problematic for any other delimiter pair Helix auto-matches.

Maybe my pinkies will thank me one day, but I'm not convinced this is a better workflow. The key insight of a modal editor is that we usually spend more time reading and editing text than composing it, but the experience of composing text shouldn't be ignored either. Having to drop and reenter insert mode just to move the cursor past a delimiter creates a drag on composing at the "speed of thought." We also all make typos, and often (IMO) it's more convenient to use arrow keys to fix one or two characters than to leave insert mode. Note also that backspace still works fine in insert mode, and I hope this doesn't change either) I don't think it's a crime against modal editors to admit that, and I don't think making an insert mode as austere as humanly possible is a necessary condition or goal of modal editing. But maybe I'm misunderstanding Helix's goals here.

Finally, I know the original behavior is achievable by remapping the keys, which I've done. (I made this comment in the first place because I was annoyed enough by the behavior *while trying to navigate quotation marks editing my config*). But if the ship's really sailed on this being the default, perhaps there should be an easy opt-in option to get the old behavior back (`insert_mode_arrow_navigation = true` or something). I'd understand if people thought this would be cluttering the config API, though.

archseer:
Could we just add the full block keymaps needed to replicate this to the doc/faq? That way we don't need _yet another config option_ but the users can just copy-paste the whole block as needed.

mangas:
I think that's necessary but I think it will just bring the old problem back because once people find the keys they will go back to misusing them but I guess it's either that or adding an FAQ section of helix does not support modeless operations (saving, undo, etc)

dead10ck:
@barszcz

>Consider typing a new line containing a matched pair of delimiters, like "key": value. The user would begin by typing a double-quote, followed by key. Now the cursor is "stuck" behind the second quotation mark

You can actually just type the second quotation mark; helix will skip over it instead of inserting another one.

@archseer 

> Could we just add the full block keymaps needed to replicate this to the doc/faq? That way we don't need _yet another config option_ but the users can just copy-paste the whole block as needed.

Sure, I can put that on my to do list.

@mangas

> I think that's necessary but I think it will just bring the old problem back because once people find the keys they will go back to misusing them

That's true, but this way at least they're making a conscious decision to misuse them, and they've been made aware that this isn't how it was designed to be used. If they want to come forth and contribute to the discussion or implementation of how to better support modeless editing, this would be welcome, but my hope is that it will at least reduce the number of users who might come to the hasty but logical conclusion that "this should work, but it doesn't"

archseer:
@mangas I think it's still better than having them built-in or behind a single flag. It's a stronger statement: "we don't think you should use these by default, but if you really want to, here's the user config". Not everyone is prepared to do it "the proper way"

mangas:
> @mangas I think it's still better than having them built-in or behind a single flag. It's a stronger statement: "we don't think you should use these by default, but if you really want to, here's the user config". Not everyone is prepared to do it "the proper way"

Mostly worried about the newcomer experience although I do think it's a step in the right direction

dead10ck:
I added all the removed bindings back to the example config in the book: https://github.com/helix-editor/helix/pull/3827

jeancf:
> providing users a way to move around with arrows in insert mode by default sets up the experience for new users coming from modeless editors to fall into the trap of thinking they can just do that they've done in other editors.

I have been using helix for one week and I just want to share my opinion on this topic as a "new user coming from modeless editors".

Until recently I never succeeded to get on board modal editing. I was interested to find out what the fuss was about and I tried vi and vim a couple of times but never managed to type anything useful and ended up having to reboot my PC because I could not exit the editor -- I am barely exaggerating.

Then I experienced helix as a aha! moment. I could move around and type stuff AND I could start using powerful key combinations as well. It is precisely because helix provided navigation in edit mode out of the box that I could get on board as a newcomer. In that configuration helix is a midway point that provides a starting point for newcomers to explore modal editing. That is maybe why you are seeing so many of them around. 

I have now reached a point where I understand why edit mode should be used for editing and not navigation and I may not re-enable navigation in edit mode (and if I want to do it after all, I know how). But for those new users that will arrive after the default has changed, they will have to climb a much steeper learning curve and may never find out that the option to navigate in edit mode exists to help them get on board.

dead10ck:
@jeancf thanks for making the effort to come share your experience, that's super useful to hear. The way these things tend to go is that decisions are made based on who is the loudest, so I appreciate you coming forward to share how the arrow keys helped you transition. I, for my own part as a code contributor, will definitely keep your experience in mind as we see how this goes.

dylrich:
After rebuilding helix today I was surprised to find this had changed. I use this functionality when it's faster to hit e.g. just the down arrow instead of esc->down->i. I hope this change will be reconsidered.

For those who want the old behavior and build their own binaries from master, I will [maintain a patch](https://gist.github.com/dylrich/2bd88c35df0cfa076930fa7dbd1bd3c3) so we can avoid cluttering our config files. Here's a one-liner:

```bash
curl https://gist.githubusercontent.com/dylrich/2bd88c35df0cfa076930fa7dbd1bd3c3/raw/b8fc07d410e620f6a1a4f6fa308d5a7ec10486e8/keymap.patch | git apply
```

aral:
I just came over to file a bug because I thought navigation in insert mode had broken in the latest build (I build Helix from main).

So that should probably tell you something about expectation violation with this change . (Only later did it cross my mind that navigation in insert mode might have been removed as it’s a very drastic decision.)

Regarding the issues with navigation in insert mode, there is only one that’s specific to insert mode that is a real problem: autocompletions being triggered by vertical navigation. That’s a bug. It could be fixed. At which point, I personally do not have any other issues with navigation/delete/etc. in insert mode. It works 99%. I don’t see the logic in throwing that 99% away for a 1% issue.

The key bindings issue, on the other hand, is _not_ solved by this change. If I’m in select mode instead of normal mode, my Ctrl+S binding still doesn’t work. The way to fix that issue is to make key bindings work in all modes (or at least have the option of having them work in all modes). This change doesn’t fix that issue. 

Also (minor rant alert, shield your eyes), having a Ctrl+S binding is not an example of “misusing” Helix. I’m not holding it wrong, Mr. Jobs :) I’m trying to reduce cognitive load because I don’t live in a world where I can do 100% of what I need to do with a computer in Helix Editor. If I wanted that, I’d use Emacs :P Seriously, though, no app lives in a vacuum and it should be a design principle to adhere to common conventions (or at least make it possible to do so) so that people don’t have to keep four different keyboard shortcuts for doing the same thing in their minds when switching between apps. Am I exaggerating? I don’t think so: off the top of my head: exiting: q (lf), :q (helix), ctrl+w (my visual markdown editor), ctrl+q (most other graphical apps). So did I map Ctrl+q to quit in Helix and lf. Hell yes, I did :) Does it only work in normal mode? Hell yes, it does. Does that annoy me a microscopic amount each time I try to use it while the file picker is up or I’m in select mode… you get the idea. Those microscopic annoyances build up :)

Helix is a real breath of fresh air and is the first time I’ve even considered using a modal editor much less enjoyed using one. I absolutely love the experience 95% of the time. And I have the utmost respect for the time and effort that has gone into it and continues to go into it.

However, it is also the only editor where I’m always slightly anxious while using it. (That’s the best way I can describe it.) Because I’ve always got a tiny voice somewhere asking some variation of “am I in the right mode”, “will this command work in this mode”, “is this the best way I could be selecting/editing this?” I feel that anything that exists to reduce this, for lack of a better term, modal anxiety is a Good Thing™. Being able to navigate and do basic edits (delete/backspace/basic selection) in insert mode and having bindings work across modes provides just that.

I just hope you’ll reconsider this design decision as I feel it’s an example of throwing the baby out with the bathwater. As others have mentioned multiple times, it will also raise the barrier of entry quite considerably. And those things have a way of altering the culture of a thing as it becomes more and more exclusive. I’d hate to see Helix hosted on SourceHut one day is what I’m saying. Take of that what you will :)

Please take this in the caring but lighthearted way it’s intended. I’m off now to see how these new defaults feel (I will give them a shot) and yet I have a feeling I’ll be spending some time reconfiguring Helix so I can edit in insert mode again. That all said, I would rather like to be actually working on my project and wasn’t planning on relearning my editor today.

Right, ’nuff said. Have a lovely weekend everyone and thank you again for making an editor that makes me love my computer more. I hope we can make it even better and open up the experience to more folks without making it extremely exclusive or introducing unnecessary barriers of entry.

:two_hearts:

aral:
(Bloody hell, I didn’t realise I’d written a whole essay. Sorry. Please feel free to skim.) :)

aral:
Sorry, one more thing to add:

> I understand it's inconvenient to have this break when it worked before and to have to add these bindings back to your config (I had to add them back to my own config too)

I really don’t get this. Basically, what’s being said is that even the person who implemented this change will not be using Helix with these defaults and yet we expect others to do so (or that we expect every new person who uses Helix to go through a rite of passage to add these configurations). Neither of those make any sense to me.

dead10ck:
@aral please take the time to read through the previous discussion. Almost everything you have said had been said before and addressed in previous comments, right down to the pop-up bug example, as well as key bindings.

Which specific key bindings you use is not the issue—no one is saying you shouldn't use Ctrl-S to save. The point is it isn't a simple matter of "just making the key binds work in all modes". There are implementation details that currently make this difficult, so until and unless others step forward and put the effort into addressing this, we should not be encouraging a use pattern that doesn't work well with the default settings. And like I've said previously, the use pattern I'm talking about isn't deciding to move a few characters while still in insert mode, it's staying in insert mode all the time.

Also, let's all keep in mind here that for the arrow keys, what we're all talking about is 4 lines in your config.

YangtseSu:
> Also, let's all keep in mind here that for the arrow keys, what we're all talking about is 4 lines in your config.

So you know that we will add these 4 lines back, and you insist on deleting them.

dylrich:
It's good to understand that the persona you're targeting with these changes is someone using insert mode most of the time for movement. However, I'm not sure the change is all that effective in discouraging this behavior.

If you were using Helix mostly-in-insert-mode previously, you were going to edit your keybindings anyway because there are some things that have never worked well in insert mode, like saving or pasting. This change does not do much in the way of discouraging those folks - it's easy to copy more custom configuration to your already well-customized setup. As @mangas said I feel that the solution to this problem is education more than anything. For a user that "responsibly" uses navigation in insert mode, this change now requires a decent amount of custom keybindings where previously you didn't need any. I would think responsible use of navigation keybindings in insert mode should be encouraged - sometimes you can realize large editing speed gains this way and for me this is the great appeal of modes. Moreover, this is a rather noticeable difference in experience from both Vim and Kakoune and would certainly discourage me from getting used to Helix as a new user.

If this change guaranteed no one would misuse Helix as a modeless editor, perhaps it would make sense. But from my point of view it feels like there's no meaningful change for the already discouraged use-case, and a use-case that should be encouraged is made more annoying for new and long-term users alike.

DannyJJK:
> > Also, let's all keep in mind here that for the arrow keys, what we're all talking about is 4 lines in your config.
> 
> So you know that we will add these 4 lines back, and you insist on deleting them.

Who is "we"? I don't plan on adding those 4 lines back. Using the arrow keys in insert mode is counter-intuitive to how a modal editor works. There's the saying "just because you can do something, doesn't mean you should", but if you really want to, why not use a modeless editor that has LSP support and lots of keyboard shortcuts? What are the advantages of using arrow keys over navigation in normal mode? I think a lot of people new to modal editors who want to learn how to properly use a modal editor (like me) will appreciate this change. It's very easy to get into bad habits, especially when your editor allows you to do them by default.

If people can add these back in via the config, I fail to see the issue. The release notes will acknowledge this change so people using the release version won't be caught off guard.

dylrich:
> Using the arrow keys in insert mode is counter-intuitive to how a modal editor works

I strongly disagree. Modes are about changing keybindings to suit the task you're currently doing. Those keybindings change to support faster, more ergonomic operations than one global keybinding pool. Using e.g. arrow keys in insert mode *can* be faster and more ergonomic than dropping into normal mode, moving, then entering insert mode again. There is nothing counter intuitive here. If your movement operation is very complex, then it might make more sense to activate normal mode for the movement. But it very much depends on what you want to do.

DannyJJK:
There's also the question of consistency. If you follow the Helix :tutor, it will tell you that it's better to use the HJKL keys for movement even though the arrow keys work. But HJKL doesn't work in insert mode effectively disabling navigation, it's only the arrow key users who still have this capability. I can't see there being many people (if any?) who use HJKL and switch to the arrow keys in insert mode.

The only "bug" right now relating to using arrow keys in insert mode is the autocomplete but there could be more in the future as the devs aren't expecting people to use the arrow keys in insert mode as much as some people are. At a later point we may have plugin support where some plugins could also cause conflicts when using arrow key navigation in insert mode.

getreu:
> Helix is first and foremost a modal editor. Willingness to support non-modal
editing is there, 

Not really the topic here: I remember an external repo with a non-modal 
configuration. To make it work some patches were needed. What is the state
of this?

> but it is not one that should be encouraged with the default
settings. There are an increasing number of users who are stumbling because
they are trying to use Helix as a non-modal editor, so this is an effort to
encourage new users to stop and take notice that Helix has a different paradigm
than VSCode, Sublime, etc.

Default settings should not discourage beginners! Please keep the ability to use 
arrow keys while still in insert mode (like in Vim):

```toml
[keys.insert]
up = "move_line_up"
down = "move_line_down"
left = "move_char_left"
right = "move_char_right"
```
Being more more papist than the Pope here is not helpful for easing the first 
beginner steps to Helix. Choosing defaults is not about educating people!

DannyJJK:
> Default settings should not discourage beginners!

I don't see how this discourages beginners. People new to the editor will either go to the documentation or the in-editor tutor. The default mode is normal mode, which can be scrolled with the arrow keys, the tutor will tell them the rest. Beginners will either know that it's a modal editor and they need to learn how to use it, or once they learn it's a modal editor they will look for something else.

> Choosing defaults is not about educating people!

I'm not really sure what this means. The defaults should be logical, sensible and allow you to get the most out of the editor. There is no "education" aspect but at the same time if the user configures the application in an unusual way, it's possible they will run into issues that might not necessarily be considered bugs (e.g. like this autocomplete issue). GIven it's a modal editor with a specific mode for navigation, and that navigating via HJKL wouldn't allow you to navigate in insert mode either, I don't see how being able to navigate with the arrow keys in insert mode is a sensible default aside from matching Vim/Kakoune functionality or that up-arrow is quicker than doing `<esc>ki`, it still seems odd to me that some people would navigate using HJKL in normal mode and then start using the arrow keys in insert mode, I would find that confusing but that's just me.

archseer:
> To make it work some patches were needed.

The patch was just needed to fix a panic with the repeat operator tracking. We fixed that in the last release.


> If you follow the Helix :tutor, it will tell you that it's better to use the HJKL keys for movement even though the arrow keys work.

Well, the arrow keys are there in normal mode for the same reason they were in insert mode: some users complained/wanted them. I personally don't use these keybinds in either mode.

pickfire:
Looking at how my colleagues uses vim, pressing <kbd>i</kbd> then do everything like a normal editor <kbd>esc</kbd> makes me rethink whether to remove the arrow keys in insert mode, I think it's fine having it since advanced users won't be using it but at the very least users who don't care about modal editing can still use it too, like they can just press <kbd>i</kbd>, do anything they want and press <kbd>esc</kbd>, I personally won't use arrow keys too.

I think we should have the arrow keys back for those who need it, helix at the very least have way better defaults compared to vi and vim, for those users that don't care and just need an editor, helix could be a drop-in replacement.

DannyJJK:
> Well, the arrow keys are there in normal mode for the same reason they were in insert mode: some users complained/wanted them. I personally don't use these keybinds in either mode.

In normal mode it's fine, the arrow keys on the keyboard are primarily designed for navigation, even if they aren't the best option for a modal editor. Insert mode is for entering and removing text, but not for navigating around text. Once some users realize they can navigate around Helix in insert mode like they could do in Nano, then they will just try to use the editor like Nano. At least if the defaults stop them from doing it, they will either look to the documentation, find a way of reversing it, or simply move to a modeless editor rather than trying to shoehorn modeless functionality into a modal editor.

aral:
> I remember an external repo with a non-modal
> configuration.

Anyone know where this is? Is this what @p-e-w was/is working towards? (https://github.com/helix-editor/helix/pull/3366)

Zykino:
Challenge for the "no arrows die hard": copy this com and fix all the typos. How many tuime did you press <ESC>? (insert mode is also fix trypo/rewrite mode)

Since a PR to revert is on the way I will skip most of the rant where I planned to say how much a middle finger I got oppening helix and having arrows not work. The part where I would tell you my great mother using the mouse to move around the text would have an easier time to use helix than me or my parents. Also when I would compare you to french govenrment repriming eacefull protest. Extract from [wikipedia for those who are not aware](https://en.wikipedia.org/wiki/Yellow_vests_protests):
> International media have also reported on the disproportionate violence used by the French police response against the protestors, including the use of explosive grenades and [flashball](https://en.wikipedia.org/wiki/Flashball) weapons resulting in multiple incidences[[spelling?](https://en.wikipedia.org/wiki/Wikipedia:Manual_of_Style)] of loss of limb and sight by the protestors.[[84]](https://en.wikipedia.org/wiki/Yellow_vests_protests#cite_note-gua2019-84)
That was sad especially comming from the rust community. Imagine running `cargo >/dev/null 2>&1`, I bet you would never have learn rust.

So strainghts to the problem you have at hand: "education" of what means "X is a modal editor" and how to use one. Am I right that this PR was supposed to "fix" this?

Solutions I can see:
1) Make tutor the landing page when you start `hx`. I did it by using `hx ~/Documents/Code/Helix/runtime/tutor` every time until completing it. Only after that I learned that I could have used `hx --tutor` or `:tutor` so "discoverability even when reading the docs" could be better. Also vim has a landing page (and SO post asking how to exit it), VS (code or not) have a landing page, qtcreator have a welcome page and included docs, ... If you do not want to see the tutorial each time maybe add a parameter that has to be set to the chapter's you read/date you finished it. So it can be hidden and re-shown when the tutor is updated.
1) At school my teachers told me how to use vim at a basic level. I never got my head around vi that inserted `~A[` or whatever each time I tried to used the arrows. kakoune was next and improved so much my workflow because I could see what I selected before deleting it. And now helix, which instead of all the other required no configuration to (almost) be perfect. No need to configure keybindings, plugins to have lsp/autocomplete or whatever. Just run `hx --health`, then go to the wiki and everything is good.
1) Include a clippy! Not rust one, kakoune (or word) one. Maybe do not put it every time but make some panneal appear hwen someone is doing something "not the modal way":  
> If the user press more than 10 times the down arrow, make a message appear that tell him that to save the change in history he have to go in normal mode. Also show him that in normal mode he could have used `10j` or `10arrowDown`.
4) `hjkl` ain't the solution for everyone. Think about peoples using non standard QWERTY keyboards. I personally base [my config on this template](https://github.com/qmk/qmk_firmware/tree/master/layouts/community/ergodox/bepo) and moved the <ESC> key to the bottom right. To me it IS ergonomic to access the arrows even while editing

There is also a weak consistency/UX argument that I want to tackle:
1) `x` should not extend the selection. It should be reserved for the `v` mode.
1) `5x` does not go to the same place as `5j` so I cannot use `rnu` in the same way... I know it is a breaking change, but `x` is "select current line", and `1x` could be select "select every lines from to current to 1 down".
1) Activate relative line numbering by default! It is even better than vim `set nu set rnu` since in helix `rnu` is present in normal mode only... the only mode where you can do `5j`. This is a subtle teaching.

Do some of my points merit to have an issue hoppened?
Do you have other "problems" with bad usage from GUI users? Do you want to discuss this somehow? So the community can find ways to improve the doc or experience for new users wherever they come from.

And again sorry for all the typos. I made it a challenge not to use the arrows keys to fix them in this post (nor the mouse) just to prove my point. (By the way I am not English native so sometimes typos are just me knowing the world, not how to write it from the top of my head.)

sudormrfbin:
@Zykino Comparing an open source project that is maintained by volunteers on their free time (one that no one is forcing you to use) to a government suppressing protest through _violence_ comes across as ridiculous at best. You are welcome to start a civil discussion, but it doesn't seem like it is your intention to do so. Hurling insults and using inflammatory language will not get you anywhere, and this is the last response you will be getting on this front.

Zykino:
You are right, I should have skipped the rant entirely. I edited it out.

I apologize for what I said.

pedronasser:
I feel this kind of decision is really bad for the future of Helix, and honestly I am disappointed.
I would rather see helix remove all default keybindings instead and maybe add an argument to load the "default bindings".
By imposing these types of opinionated changes, you are clearly signaling that you don't helix to be used by anyone besides the people who already is used to the modal editing. Which is a very bad way to introducing the modal editing approach new users.
Helix should be new user friendly to a point.

But who cares? I am not a core contributor.

airtonix:
vim is such a niche editor that discussions like this are akin to a group of street thugs arguing about the finer points of a language no one alive today even remembers.

These kinds of discussions don't do anything enamour more those whore are more productive with the mouse and  `ctrl + c` + `ctrl + v` and actual left and right arrow keys.

dead10ck:
@airtonix lol. Thanks for that, that was a pretty good chuckle.

PR #3672, opened 2022-09-03
custom spinners with docs and statusline integration

sbromberger:
This is a clean implementation of #2793 applied to the latest master.

### DIFF
--- a/helix-term/src/ui/spinner.rs
+++ b/helix-term/src/ui/spinner.rs
@@ -1,7 +1,10 @@
 use std::{collections::HashMap, time::Instant};
 
+use helix_view::editor::BRAILLE_SPINNER_STRINGS;
+
 #[derive(Default, Debug)]
 pub struct ProgressSpinners {
+    default: Spinner,
     inner: HashMap<usize, Spinner>,
 }
 
@@ -11,7 +14,14 @@ impl ProgressSpinners {
     }
 
     pub fn get_or_create(&mut self, id: usize) -> &mut Spinner {
-        self.inner.entry(id).or_insert_with(Spinner::default)
+        self.inner.entry(id).or_insert_with(|| self.default.clone())
+    }
+
+    pub fn new(default: Spinner) -> Self {
+        Self {
+            default,
+            inner: HashMap::new(),
+        }
     }
 }
 
@@ -21,9 +31,9 @@ impl Default for Spinner {
     }
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Spinner {
-    frames: Vec<&'static str>,
+    frames: Vec<String>,
     count: usize,
     start: Option<Instant>,
     interval: u64,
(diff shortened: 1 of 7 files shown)


sbromberger:
This is ready for review. cc @archseer @the-mikedavis. Thanks.

helix-term/src/ui/spinner.rs
```diff
@@ -46,22 +64,31 @@ impl Spinner {
...
     pub fn frame(&self) -> Option<&str> {
-        let idx = (self
-            .start
-            .map(|time| Instant::now().duration_since(time))?
-            .as_millis()
-            / self.interval as u128) as usize
-            % self.count;
-
-        self.frames.get(idx).copied()
+        let idx = if self.count > 1 {
```
archseer, line 81: This adds complexity. Revert the changes and make the spinner in config an Option<Spinner>. If an empty set is passed just use the default.

helix-view/src/editor.rs
```diff
@@ -162,6 +164,10 @@ pub struct Config {
     pub rulers: Vec<u16>,
     #[serde(default)]
     pub whitespace: WhitespaceConfig,
+    /// Vector of Strings of custom spinner frames, one per element
+    pub spinner_frames: Vec<String>,
```
archseer, line 168: Both of these should be nested in an object: `spinner = { frames =, interval =}`

archseer (commented):
I'm still skeptical whether this is needed: it's nice to have it configurable, but will anyone actually change the default? We deliberately left this out of the original spinner PR.

helix-term/src/ui/spinner.rs
archseer: Alternatively: You could make get() always return &str and simply unwrap_or("")

sbromberger:
> I'm still skeptical whether this is needed: it's nice to have it configurable, but will anyone actually change the default? 

Objectively, yes: the original PR got several positive feedback icons, and at least a few of us have been using this for months.

Why? Well, with respect, the default spinner is suboptimal under certain circumstances (and subjectively ugly): not only is the braille hard to see, but on underpowered machines or over high-latency connections, the screen refreshes take a very long time. From an accessibility perspective, having a single frame spinner (on/off) to let you know that LSP operations are in progress is much better. Since I was adding the single frame capability, it just made sense to allow full customization.

Personal anecdote: on a recent trip, over a low-bandwidth/high-latency (satellite) connection, using helix over mosh with the default spinners was a really bad experience. Switching to single-frame spinners made the experience less painful.

sbromberger:
I'll wait to hear whether you're amenable to making this change before I address the other code comments. Thanks.

helix-term/src/ui/spinner.rs
sbromberger: The count is necessary since if we have a single frame we don't worry about the call to (and calculations with) `Instant::now()` since the spinner is either on with frame 0 or off. This is independent of the actual frames, so I'm not quite sure how checking an `Option` would work here.

PR #3678, opened 2022-09-04
Add formatter `elm-format` for Elm

heyakyra:


the-mikedavis:
I'm not sure we should add an external formatter for Elm: `elm-format` would be used instead of the LSP auto-formatting (which also uses `elm-format` but computes the diffs) when installed. There isn't currently a way to prefer the LSP formatting but fall back to the external formatter.

kirawi:
Helix will generate a diff when using an external formatter, but it may not be as efficient on larger files because it will fallback to a less efficient fast path after a 1 second timeout.

heyakyra:
Ah, thank you. I must be hitting this https://github.com/elm-tooling/elm-language-server/issues/692

michaeljones:
@heyakyra - is it working smoothly for you at the moment? I seem to be in a state where the language server works for a few minutes and then just times out on every request. I can't explain it. Probably doing some silly. I'd be interested to hear confirmation that it is working for someone else though.

PR #3684, opened 2022-09-04
Handle formatter errors, and save anyway

A-Walrus:
Adresses [comment](https://github.com/helix-editor/helix/pull/3670#issuecomment-1236150951)
If formatting fails, report error to log and save without formatting.
log entry looks something like this:
`2022-09-04T13:17:52.359 helix_view::document [ERROR] Formatter exited with non zero exit status`
image
(tested by adding formatter to `$PATH` which is just a shell script that returns exit status)
```bash
#!/usr/bin/env bash
exit 1
```

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -543,12 +543,19 @@ impl Document {
             }
 
             if let Some(fmt) = formatting {
-                let transaction = fmt.await?;
-                let success = transaction.changes().apply(&mut text);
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):

michaeljones:
Thanks for fixing this! It was bugging me in the last version and I just started to write up a report and found that it had already been fixed in the new release!

PR #3685, opened 2022-09-04
Enable CI builds for `riscv64-linux`

yvt:
AppImage build is not possible for now because [AppImageKit](https://github.com/AppImage/AppImageKit/blob/729a1a60e41cdd90554462495576cdae4a2896ed/src/appimagetool.c#L68-L73) does not recognize the architecture name.

~~A preview CI run based on this patch + `22.08.1` can be found at <https://github.com/yvt/helix/actions/runs/2987576146>.~~ (out of date)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -54,6 +54,11 @@ jobs:
           rust: stable
           target: aarch64-unknown-linux-gnu
           cross: true
+        - build: riscv64-linux
+          os: ubuntu-20.04
+          rust: stable
+          target: riscv64gc-unknown-linux-gnu
+          cross: true
         - build: x86_64-macos
           os: macos-latest
           rust: stable


.github/workflows/release.yml
```diff
@@ -128,6 +133,15 @@ jobs:
...
             /target/${{ matrix.target }}/release/hx
 
+      - name: Strip release binary (riscv64)
+        if: matrix.build == 'riscv64-linux'
+        run: |
+          docker run --rm -v \
+            "$PWD/target:/target:Z" \
+            rustembedded/cross:${{ matrix.target }} \
+            riscv64-linux-gnu-strip \
+            /target/${{ matrix.target }}/release/hx
```
the-mikedavis, line 143: I think this could be combined with the above one, right? It would just need to replace the target with `${{ matrix.target }}-gnu-strip`

yvt: Rust target names and the GCC target names selected by cross-rs don't quite match up. E.g.:

| Rust | GCC from cross-rs image |
| --- | --- |
| `aarch64-unknown-linux-gnu` | `aarch64-linux-gnu` |
| `riscv64gc-unknown-linux-gnu` | `riscv64-linux-gnu` |
| `armv7-unknown-linux-gnueabihf` | `arm-linux-gnueabihf` |

yvt: Maybe we could do #3327 in a CI step so that the change only affects the release CI run.

yvt: #3780

yvt:
Rebased on top of `master` and removed the post-build strip step (#3780 takes care of stripping).

the-mikedavis (approved):

PR #3686, opened 2022-09-04
xcopy - assume destination is a directory

NotAgentBilly:
By default xcopy asks if the destination is a file or directory. Since we know it's a directory we can add the `/i` flag and prevent an extra prompt.

Description of `/i`
```
If source is a directory or contains wildcards and destination does not exist, 
xcopy assumes destination specifies a directory name and creates a new directory. 
Then, xcopy copies all specified files into the new directory.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,8 +50,8 @@ config directory (for example `~/.config/helix/runtime` on Linux/macOS, or `%App
 
 | OS                   | Command                                      |
 | -------------------- | -------------------------------------------- |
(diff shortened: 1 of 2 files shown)


the-mikedavis:
This table is also in the book's installation section: https://github.com/helix-editor/helix/blob/44b447947c810b7fc6a13383b4379f0f48501c17/book/src/install.md?plain=1#L69-L70. Could you update the commands there too?

NotAgentBilly:
Done

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #2977, opened 2022-07-05
Add boolean toggle value with increment/decrement shortcuts

hirschenberger:
fixes #2931

PR #3688, opened 2022-09-04
Incrementer for (config driven) ordered lists.

aikomastboom:
A little spin-off from https://github.com/helix-editor/helix/pull/2977

> [the-mikedavis](https://github.com/the-mikedavis) commented https://github.com/helix-editor/helix/pull/2977#issuecomment-1210069668
> Code-wise this looks fine to me but IMO toggling a boolean is conceptually different from increment/decrement. Even as a standalone binding or feature I think this is too specific: I don't see much value over editing the boolean by hand.

> [aikomastboom](https://github.com/aikomastboom) commented https://github.com/helix-editor/helix/pull/2977#issuecomment-1216189803
> I understand your concern about it being "conceptually different", however there are only so many keybindings to remember and using C-a to scroll thru an ordered enumeration would make sense to me. Eg. [0,1,2,3,...], [true,false], [yes,no], [a,b,c,...], [red,green,blue,...16 base colors/ or colors already in use in eg. a theme file]. Agreed it will not always make sense to use this feature, however it would be up to user to choose what feature to use/apply when?

another sequence I can think of is musical notes: c, c#, d, e, f, f#, g, a, b

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -51,6 +51,7 @@ You may also specify a file to use for configuration with the `-c` or
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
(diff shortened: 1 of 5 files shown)


dead10ck (commented):
This is way too niche for core, in my opinion.

pascalkuthe:
I am closing this as stale as this PR has a lot of conflicts. I also don't see us going forward with this as it's too niche for core. Thjs can already be implemented with pipe instead and could be a simple function in your keymap once #10389 lands.

Thank you for contributing

PR #3690, opened 2022-09-04
theme keys static and documented

### LINKED ISSUE #3649: Keep all theme keys documented in one place
Currently the theme keys are all over the place in the code and keeping track of them is only possible by looking at all the commits to find when new ones are added, making it hard for themes to know what they are missing and what has been removed since.

Fo example: https://github.com/helix-editor/helix/pull/2759 added `ui.bufferline.active` but did not add it to the documentation, only the config option.

While not the only way to do so, here is what I propose:

Create a new crate only dealing with the `ThemeKey` type and its documentation.

The theme keys are a single enum like this:

```rust
#[derive(Debug, PartialEq, PartialOrd, Ord, Eq, Hash)] // PartialOrd, Ord for storage in a BTreeMap
pub enum ThemeKey {
    // Sort them alphabetically for easier maintenance
    UiBackground,
    UiOther,
    UiSomethingElse,
    // ...
}

impl FromStr for ThemeKey {
    // ...
}

impl Display for ThemeKey {
    // opposite of FromStr
}

// other impls as needed
```

Then add a binary to this crate that can be used to

1. generate the doc for all these variants
2. check the validity of a theme file (with varying levels of strictness, like "all keys and nothing more", "all keys but additional keys are okay", "not all keys but additionals are not okay", "not all keys and additionals are okay"), with checking for unused colors in the `[palette]` too

Run 1) in CI and check there are no diffs from the current docs, that way it always stays up to date for the helix doc pages, or even generate the page in CI

I'll try to work on this soon because it has really been bothering me 😅

poliorcetics:
This PR adds a `ThemeKey` type that contains as much static scopes as possible for Tree Sitter
highlight queries.

It also adds a new generated file for the book which fully documents the highlight scopes and a
new `cargo xtask` subcommand to ensure all of the default highlight queries are in the static
scopes and will not fall in `ThemeKey::OtherQuery`.

Closes #3649


poliorcetics:
@the-mikedavis I didn't add the theme checking but having a single source and always up to date source of truth for theme keys should make it much easier to know if a theme is up to date for maintainers, even with a manual workflow for now.

I added all the rest I think, and the resulting documentation is also more searchable IMO since I can now look for `keyword.<something>` whereas that was not possible before. because it used a manual tree structure.

helix-view/src/theme/theme_keys.rs
```diff
@@ -0,0 +1,533 @@
...
+    "keyword.operator"             is KeywordOperator           -> [Keyword]                                         : "Operator keywords, e.g. `not`, `or`, etc.",
+    "keyword.operator.assignment"  is KeywordOperatorAssignment -> [KeywordOperator, Keyword]                        : "Assignements keywords",
+    "keyword.other"                is KeywordOther              -> [Keyword]                                         : "Other keywords, without a clear category",
+    "keyword.storage"              is KeywordStorage            -> [Keyword]                                         : "Default style for storage keywords",
+    "keyword.storage.modifier"     is KeywordStorageModifier    -> [KeywordStorage, Keyword]                         : "Storage modifier keywords, e.g. `move`, `dyn`",
+    "keyword.storage.modifier.mut" is KeywordStorageModifierMut -> [KeywordStorageModifier, KeywordStorage, Keyword] : "Mutable storage modifier",
+    "keyword.storage.modifier.ref" is KeywordStorageModifierRef -> [KeywordStorageModifier, KeywordStorage, Keyword] : "Reference storage modifier",
```
A-Walrus, line 454: Could the macro be modified so that you only need to define the first fallback, and it will recursively take the fallback of the fallback and so on?
That way you could just say something like:
```rust
    "keyword.operator"             is KeywordOperator           -> Keyword                 : "Operator keywords, e.g. `not`, `or`, etc.",
    "keyword.operator.assignment"  is KeywordOperatorAssignment -> KeywordOperator         : "Assignements keywords",
    "keyword.other"                is KeywordOther              -> Keyword                 : "Other keywords, without a clear category",
    "keyword.storage"              is KeywordStorage            -> Keyword                 : "Default style for storage keywords",
    "keyword.storage.modifier"     is KeywordStorageModifier    -> KeywordStorage          : "Storage modifier keywords, e.g. `move`, `dyn`",
    "keyword.storage.modifier.mut" is KeywordStorageModifierMut -> KeywordStorageModifier  : "Mutable storage modifier",
    "keyword.storage.modifier.ref" is KeywordStorageModifierRef -> KeywordStorageModifier  : "Reference storage modifier",
```
BTW, big fan of the idea of making the theme keys strongly typed :+1:

poliorcetics: I could do so with a custom iterator or using [`std::iter::successors`](https://doc.rust-lang.org/std/iter/fn.successors.html) yes, seems like a good idea and avoids a `SmallVec` too. Will do this evening

helix-view/src/theme.rs
```diff
@@ -157,25 +161,38 @@ impl Theme {
...
     }
 
     /// Get the style of a scope, falling back to dot separated broader
     /// scopes. For example if `ui.text.focus` is not defined in the theme,
     /// `ui.text` is tried and then `ui` is tried.
-    pub fn try_get(&self, scope: &str) -> Option<Style> {
-        std::iter::successors(Some(scope), |s| Some(s.rsplit_once('.')?.0))
-            .find_map(|s| self.styles.get(s).copied())
+    pub fn try_get<T: AsRef<ThemeKey>>(&self, scope: T) -> Option<Style> {
+        // Remove generics to make monomorphization less costly
```
kirawi, line 172: That's a good point, using `AsRef<ThemeKey>` in so many places might be pretty costly. https://matklad.github.io/2021/09/04/fast-rust-builds.html#compilation-model-monomorphization seems to indicate it's _mainly_ a problem between multiple crates. This might need some profiling?

the-mikedavis:
I preferred seeing keys as "scope.scope.scope" rather than "ScopeScopeScope" - the fallback behavior was more intuitive. Is there a different way of approaching this, maybe by using a macro when getting a key from a theme?

poliorcetics:
> I preferred seeing keys as "scope.scope.scope" rather than "ScopeScopeScope" - the fallback behavior was more intuitive. Is there a different way of approaching this, maybe by using a macro when getting a key from a theme?

I could do `Ui_Menu` instead of `UiMenu`, that would make it much easier to read with a separator. It's not as Rust-y but it's a generated enum with dozens of variants so 🤷

kirawi:
We could make use of namespacing.

poliorcetics:
> We could make use of namespacing.

I can try generating something like `ui::Menu`, and `Ui` constants yes, with the enum as backing type, good idea. I'll see what I can do

poliorcetics:
> We could make use of namespacing.

Well it's very hard in `macro_rules` and I'm not willing to introduce such a proc macro for just this, the complexity and compile time would suffer a lot. Nested elements in macros are always hard, both with types and modules. We could maintain a manual mapping of modules to the keys used in `helix`, or I can make another `macro_rules`.

@the-mikedavis, what do you think about `Ui_Menu` ? From an implementation POV, it's very easy, and the separator is there. I can make everything lowercase too, like `ui_menu` and wrap the whole thing in something like this:

```rust
pub type tk = ThemeKeys;
```

That way the usage point becomes `tk::ui_menu`, and the import is simply `use helix_view::theme::tk;`

poliorcetics:
TODO for self:

- Fix build
- Rework from [this comment](https://github.com/helix-editor/helix/pull/3690#issuecomment-1250057151)

poliorcetics:
I moved the names to `Ui_Linenr_Selected` format, to make it easier to read and rebased on most recent master

poliorcetics:
I don't think CI failures are related to my changes

poliorcetics:
What do I have to do for this to move forward ? IMO it would greatly improve documentation for theme keys, and avoid allocating strings in several places

archseer:
> avoid allocating strings in several places

Are these allocations that costly? From what I can tell, we use static strings everywhere apart from parsing the theme file.

---

This might make sense for `ui.` but I'm not a huge fan of hardcoding the syntax scopes (even with the OtherScope escape hatch), it's a huge macro that's essentially a string interner just so we can use the syntax scopes in a couple of spots in the UI via `get/try_get`. When rendering we use raw indexes anyway.

The query validation could then be done externally: a file could essentially contain a list very similar to the macro input. We'd then use that to generate the documentation tree. A lint script could then `cut` the first column of known values, `rg @\w+` (dropping any that start with `@_`) and cross reference against that list.

poliorcetics:
The documentation benefits to know what is used in themes and what is not is nice.

For example, this PR surfaces several duplications between `sshconfig` and the rest, e.g. `constant.builtin.boolean` vs `boolean` (only used in `sshconfig`). I have also caught people unnecessarily introducing new scopes when existing ones were already used elsewhere.

Theming is very important to users, having a list of all keys, *documented* (which they would not be with a parser from the runtime queries), allows trimming down that list on our side (making theme definition easier) and correctly define new themes on the user's side.

Ideally we would even document which queries are used in which language, but that would probably make the table too big for now so I left it for later, once I have a better idea for presentation.

poliorcetics:
@archseer @the-mikedavis I completely changed the way this is done.

1) The big macro is now only compiled when compiling the `xtask` crate, which means regular `helix` compilation won't pay the cost
2) I removed the big enum, leaving the `&str` in the code as currently in `master`.
3) I kept the check and documentation to ensure languages don't introduce similar-but-not-the-same keys in their highlight queries, if you look at the first commit you'll that `master` has several such instances.

xtask/src/themekeyscheck.rs
```diff
@@ -0,0 +1,430 @@
...
+
+    table
+}
+
+/// Keys must be given in the following order:
+///
+/// - UI keys (used only in UI theming)
+/// - Mixed keys (used both for language and UI theming)
+/// - Language keys (used only in language theming)
+macro_rules! generate_theme_keys {
```
the-mikedavis, line 134: I think this would be more approachable to newer Rustaceans if it used the common `hash_map` macro (or `hashmap!` from helix_core) for associating the keys and docs. The other constants in here could just be regular constants defined outside of the hashmap macro

xtask/src/themekeyscheck.rs
```diff
@@ -0,0 +1,430 @@
...
+    write_table_header(&mut table, "UI-only Keys");
+
+    for &(key, note) in KEY_DOCS {
+        if key == FIRST_MIXED_KEY {
+            write_table_header(&mut table, "Mixed Keys (UI & languages)");
+        } else if key == FIRST_LANGUAGE_KEY {
+            write_table_header(&mut table, "Language-only Keys");
+        }
+
+        ::std::writeln!(
```
the-mikedavis, line 109: minor style nit but it's more common in the helix codebase not to use the global path, just using `writeln!`

xtask/src/themekeyscheck.rs
```diff
@@ -0,0 +1,430 @@
...
+        let language = get_language(&grammar_name);
+        let query_text = read_query(&language_name, highlight_queries_file);
+        if let (false, Ok(language)) = (query_text.is_empty(), language) {
+            match Query::new(language, &query_text) {
+                Err(reason) => {
+                    return Err(format!(
+                        "Failed to parse {} queries for {}: {}",
+                        highlight_queries_file, language_name, reason
+                    ))
+                }
```
the-mikedavis, line 26: We might want to eliminate the query-check xtask in favor of this once since we need the queries here anyways

poliorcetics: I don't think so: this only checks `highlights.scm`, not the others and conflating both is probably confusing because of that: checking themes could fail on indentation queries, that's not a good experience

xtask/src/themekeyscheck.rs
poliorcetics: Done

xtask/src/themekeyscheck.rs
poliorcetics: I don't agree, it would create a big allocation for nothing in a check that is already quite slow and would not allow iterating in insertion order, which is important if we don't want this to create big diffs on each changes.

poliorcetics:
Updated to latest master, this PR is no fun to maintain 😓

archseer:
Hmm why were these closed? We were going to merge this in the next cycle

PR #1854, opened 2022-03-22
Add syntax highlighting for Solidity

slinlee:
image

### TODO:

- [x] Add fork of `tree-sitter-solidity` that includes `src/` and the generated files.
- [x] Add queries for highlighting

Resolves #1853

PR #3692, opened 2022-09-04
Add syntax highlighting for SML

Giorbo:
Hi

This is my first PR and I hope everything is fine, I had to create a fork for the tree-sitter grammar sources like in [#1854](https://github.com/helix-editor/helix/pull/1854) but apart from that everything went smoothly thanks to the nice docs

I'm really enjoying Helix, nice work!

### DIFF
--- a/runtime/queries/sml/highlights.scm
+++ b/runtime/queries/sml/highlights.scm
@@ -79,27 +79,11 @@
 (longvid ((vid) @vid
           (#match? @vid "^[A-Z].*"))) @constructor
 
-;; "true", "false", "nil", "::", and "ref" are built-in constructors
-((vid) @constant.builtin
- (#match? @constant.builtin "true"))
-((vid) @constant.builtin
- (#match? @constant.builtin "false"))
-((vid) @constant.builtin
- (#match? @constant.builtin "nil"))
-((vid) @constant.builtin
- (#match? @constant.builtin "::"))
-((vid) @constant.builtin
- (#match? @constant.builtin "ref"))
-(longvid ((vid) @vid
-          (#match? @vid "true"))) @constant.builtin
-(longvid ((vid) @vid
-          (#match? @vid "false"))) @constant.builtin
-(longvid ((vid) @vid
-          (#match? @vid "nil"))) @constant.builtin
-(longvid ((vid) @vid
-           (#match? @vid "::"))) @constant.builtin
-(longvid ((vid) @vid
-          (#match? @vid "ref"))) @constant.builtin
+((vid) @constant.builtin (#eq? @constant.builtin "nil"))
+((vid) @constant.builtin.boolean
+ (#match? @constant.builtin.boolean "^(true|false)$"))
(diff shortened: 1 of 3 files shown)


runtime/queries/sml/highlights.scm
```diff
@@ -0,0 +1,106 @@
...
+(longvid ((vid) @vid
+          (#match? @vid "true"))) @constant.builtin
+(longvid ((vid) @vid
+          (#match? @vid "false"))) @constant.builtin
+(longvid ((vid) @vid
+          (#match? @vid "nil"))) @constant.builtin
+(longvid ((vid) @vid
+           (#match? @vid "::"))) @constant.builtin
+(longvid ((vid) @vid
+          (#match? @vid "ref"))) @constant.builtin
```
the-mikedavis, line 99: These can be simplified a bit:

```suggestion
((vid) @constructor
 (#match? @constructor "^[A-Z].*"))

((vid) @constant.builtin (#eq? @constant.builtin "nil"))
((vid) @constant.builtin.boolean
 (#match? @constant.builtin.boolean "^(true|false)$"))

((vid) @operator (#eq? @operator "::"))
((vid) @keyword.storage.modifier (#eq? @keyword.storage.modifier "ref"))
```

`@keyword.storage.modifier` is used in the rust queries for `ref`. `@operator` could also be a good scope for it.

the-mikedavis (commented):
The queries are a bit limited: there isn't any highlighting of variables or operators. I'm not very familiar with SML though - it might not be clear just from the syntax whether something is a variable or operator or function?

Giorbo:
> The queries are a bit limited: there isn't any highlighting of variables or operators. I'm not very familiar with SML though - it might not be clear just from the syntax whether something is a variable or operator or function?

I'm not sure whether you can infer that something is a variable/operator/function from just the syntax, I'm a newbie in SML as well as with tree-sitter

Your suggestions look great, in the last one though I noticed that you repeat the constructor match that is already on lines 74-75 on the original highlights, is that intentional?

runtime/queries/sml/highlights.scm
```diff
@@ -0,0 +1,93 @@
...
+
+;; *******************************************************************
+;; Constructors
+;; *******************************************************************
+
+;; Assume value identifiers starting with capital letter are constructors
+((vid) @constructor
+ (#match? @constructor "^[A-Z].*"))
+(longvid ((vid) @vid
+          (#match? @vid "^[A-Z].*"))) @constructor
```
the-mikedavis, line 80: This the stanza on L77-78 will match before this pattern so this one can be removed

the-mikedavis:
Ah actually I meant that to de-duplicate the `longvid` version rather than adding a duplicate

Giorbo:
> Ah actually I meant that to de-duplicate the `longvid` version rather than adding a duplicate

Oh ok, that makes sense

the-mikedavis (approved):

the-mikedavis:
Looks good, thank you!

PR #3694, opened 2022-09-04
optimize `split_selection_on_newline`

kirawi:
* Uses the `Lines` iterator over a regex.


kirawi:
Upon some rough testing, the performance seems to actually be the same.

Issue #2248, opened 2022-04-24
Characters rendering multiple times


maybork:
### Summary

When working on a Python file with the suggested LSP I've noticed odd behavior - sometimes (I think it's when there's a warning/error thrown by the lsp on the current line) characters show up multiple times. This behavior is inconsistent, but I've got a screen capture documenting it.

![image](https://user-images.githubusercontent.com/82605611/164979520-d137b25f-3f9c-41e3-91a1-dd68f2106863.png)
[Video of the behavior](https://youtu.be/c0E7I_4nnRg)

I used the LSP provided within the documentation ([pylsp](https://github.com/python-lsp/python-lsp-server)) with `python-lsp-black`, `pylsp-rope` and `pylsp-mypy` (none of these could cause that sort of behavior as far as I know)

### Reproduction Steps

I tried this:

Inserting text

I expected this to happen:

Text to be inserted once

Instead, this happened:

It got inserted multiple times

### Platform

Linux

### Terminal Emulator

Konsole 21.12.3

### Helix Version

helix 0.6.0

kirawi:
You should check if this was fixed in `22.03`

maybork:
> Can you check if this was fixed in `22.03`?


Tried 22.05-dev, same results.

maybork:
A fresh install of the dev branch with a fresh pylsp install produces the same results, too.

archseer:
It's only at the end of the document, right? I've seen that happen before when the diagnostics ranges were pointing past the document end

the-mikedavis:
Naively I think we might be able to handle this here: https://github.com/helix-editor/helix/blob/3cced1e3c825b0ae527f43358b29d9595c54065b/helix-term/src/ui/editor.rs#L275-L289

by doing a `min(diagnostic.range.end, document_end)`. I can give this a proper look in a while unless someone wants to beat me to it 🙂

the-mikedavis:
Digging into this, the diagnostic isn't beyond the document - we do have checks that handle that case in the rendering code. So using the `min` of the range and document-end isn't the solution (☝️).

[...]

PR #3695, opened 2022-09-04
Rewrite Merge to take two HighlightEvent Iterators

the-mikedavis:
Currently, the HighlightEvent merge function overlays spans (`Vec<(usize, std::ops::Range<usize>)>` on an Iterator of HighlightEvent. The current merge doesn't handle overlapping spans well though, specifically with LSP diagnostics:

* diagnostic ranges may be discarded if they overlap
* ghost text may appear when a language server sends overlapping diagnostics at the end of a document

This merge function takes two HighlightEvent Iterators which may themselves have overlapping highlights and produces a new Iterator with highlights from `right` overlaid on `left`. Also included is a new Iterator that properly subslices overlapping ranges in a `Vec<(usize, std::ops::Range<usize>)` which can be passed to the `merge` function and satisfies all of `merge`'s invariants and assumptions.

The new merge is more complicated and it looks like we get dinged for the extra `VecDeque`s but otherwise the performance looks comparable. I couldn't find any noticeable changes in the flamegraph except for two new tiny slivers for `VecDeque::extend`.

Fixes https://github.com/helix-editor/helix/issues/2248
I rebased #2857 on this and it works well and actually the types work out much nicer than that branch's merge function.


helix-core/src/syntax.rs
```diff
@@ -1856,137 +1856,347 @@ fn injection_for_match<'a>(
...
+/// Highlights from `right` overlay highlights from `left`.
+/// Any `Source` events in `right` which preceed or go beyond any `Source`
+/// events in `left` are truncated or discarded.
+///
+/// If all of the following invariants are satisfied for `left` and `right`
+/// then the invariants hold for the resulting merged iterator as well.
+///
+/// - `Source` events are sorted by `start`, monotonically increasing.
+///   Ignoring `HighlightStart` and `HighlightEnd` events, for all
+///   consecutive `l` and `r`, `l.start < r.start`.
```
archseer, line 1880: We could have a couple of `debug_assert!`s so that this at least gets caught on debug builds or tests

the-mikedavis: I introduced two new `Option<std::ops::Range<usize>>` fields on the `Merge` to track these two. If I understand `debug_assert_eq!` correctly, we never end up reading or writing to those fields in release builds so I think the impact should be minimal.

Pretty handy that this works, thanks for the pointer 😀

the-mikedavis:
I think I've debugged this pretty thoroughly but it's still a somewhat risky change: bugs in `SpanIter` (which _should_ all be squashed now) manifest as duplicated "ghost" text. I'll be driving this branch for the foreseeable future to see if I can find any rendering bugs. But this branch could probably use some time to "simmer".

It's also probably going to be pretty impossible to review 😅. My desk is completely covered in number-line drawings of every possible scenario for overlapping ranges. I think I have test cases for most scenarios though and the ascii art is very pretty 🙂

pascalkuthe:
I review the SpanIter locally and found/fixed some bugs and implemented some performance optimizations.
I did the changes locally because it would be really annoying to do trough the GH webui.
I opened at PR agains your fork (the-mikedavis/helix#3) that I will refer to during my review comments.

The algorithm was a bit unapproachable at first.
However it looked like the perfect candidate for property testing, because you clearly defined some properties that the resulting HighlightEvents must uphold.
I used proptest to implement this in 924e92cc42f690aed6571b0760924062dbc9de36.
I used proptest instead of quickeck, because it makes writing reduces much nicer which is relevant because the input data also needs to uphold some variants.I converted the one existing use of quickcheck to use proptest instead.

The resulting proptest almost instantly reduced some crashes. It took me quite a while to fix this because there were actually multiple bugs. The rest will be in review comments

helix-core/src/syntax/span.rs
```diff
@@ -0,0 +1,511 @@
...
+    }
+}
+
+struct FlatSpanIter<I> {
+    iter: I,
+}
+
+/// Converts a Vec of spans into an [Iterator] over [HighlightEvent]s
+///
+/// This implementation does not resolve overlapping spans. Zero-width spans are
```
pascalkuthe, line 213: Zero-width spans were actually not eliminated but was trivial to implement. (See 36d44f54b330d81ac26cd9dda9b1713a0520d6ae)

pascalkuthe (changes requested):
I reviewed the SpanIter locally and found/fixed some bugs and implemented some performance optimizations.
I did the changes locally because it would be really annoying to do trough the GH webui.
I opened at PR against your fork (https://github.com/the-mikedavis/helix/pull/3) that I will refer to during my review comments.

The algorithm was a bit unapproachable at first.
However it looked like the perfect candidate for property testing, because you clearly defined some properties that the resulting `HighlightEvents` must uphold.
I used `proptest` to implement this in https://github.com/helix-editor/helix/commit/924e92cc42f690aed6571b0760924062dbc9de36.
I used `proptest` instead of `quickeck`, because it makes writing reduces much nicer which is relevant because the input data also needs to uphold some variants.I converted the one existing use of quickcheck to use proptest instead.

The resulting proptest almost instantly reduced some crashes. It took me quite a while to fix this because there were actually multiple bugs (see review comments)

the-mikedavis:
This approach adds a lot of complexity which ends up not being necessary: diagnostics can be handled as a special case and there isn't currently another producer of `Vec<(usize, std::ops::Range<usize>)>` that has overlapping spans or needs multiple highlights per range. (Selections are non-overlapping by nature and rainbow highlights end up not overlapping and only ever need one color each.)

Plus this branch has some rendering bugs when there are overlaps.

We can revisit the `Merge` change if it ends up being necessary later down the line but for now it doesn't look necessary to take two iterators

PR #3696, opened 2022-09-04
fix: wait for is_exit_success with a timeout limit

### LINKED ISSUE #3655: Helix fails to start in WSL with DISPLAY set, but no xorg server running
### Summary

Environment:
Wsl running debian 11 (windows 10). Installed hx from source, 22.08 tag.

I have DISPLAY set so I can run frontend e2e tests, but the xorg server in windows (vcxsrv) isn't always running, I only turn it on to run the tests. I'm sure it's incorrect to have DISPLAY set while xorg server is not running, and I'll probably comment this from my settings from now on, but it still seems wrong that helix would fail to start, and I've had this env set for months without running into any other issues.

image

The cause of the problem seems to be xsel, `xsel -o -b` hangs with the exact same behavior as hx.

Maybe the answer here is just don't have DISPLAY set if xserver isn't running, and if so feel free to close.

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:
Helix to start

Instead, this happened:
Nothing happens.

### Helix log

```
```

</details>
There is nothing in the logs, except lines from runs with DISPLAY unset.

### Platform

Linux (WSL debian 11)

### Terminal Emulator

windows terminal, alacritty, *

### Helix Version

helix 22.08 (ec28b2b5)

prscoelho:
Fixes #3655 

Not sure about the logging or the actual timeout duration, please advise.

I want to test this on my work pc tomorrow because that's where it was failing, but replacing xsel with a sleep seems to return from timeout early, so it should be good.

### DIFF
--- a/helix-view/src/clipboard.rs
+++ b/helix-view/src/clipboard.rs
@@ -94,7 +94,9 @@ pub fn get_clipboard_provider() -> Box<dyn ClipboardProvider> {
 #[cfg(not(any(windows, target_os = "wasm32", target_os = "macos")))]
 pub fn get_clipboard_provider() -> Box<dyn ClipboardProvider> {
     use crate::env::{binary_exists, env_var_is_set};
+    use helix_lsp::block_on;
     use provider::command::is_exit_success;
+    use tokio::task::block_in_place;
     // TODO: support for user-defined provider, probably when we have plugin support by setting a
     // variable?
 
@@ -114,9 +116,8 @@ pub fn get_clipboard_provider() -> Box<dyn ClipboardProvider> {
         }
     } else if env_var_is_set("DISPLAY")
         && binary_exists("xsel")
-        && is_exit_success("xsel", &["-o", "-b"])
+        && block_in_place(|| block_on(is_exit_success("xsel", &["-o", "-b"])))
     {
-        // FIXME: check performance of is_exit_success
         command_provider! {
             paste => "xsel", "-o", "-b";
             copy => "xsel", "-i", "-b";
@@ -253,13 +254,24 @@ pub mod provider {
         use anyhow::{bail, Context as _, Result};
 
         #[cfg(not(any(windows, target_os = "macos")))]
-        pub fn is_exit_success(program: &str, args: &[&str]) -> bool {
-            std::process::Command::new(program)
-                .args(args)
-                .output()
-                .ok()
-                .and_then(|out| out.status.success().then(|| ())) // TODO: use then_some when stabilized
-                .is_some()
+        pub async fn is_exit_success(program: &str, args: &[&str]) -> bool {
+            let result = tokio::time::timeout(
+                std::time::Duration::from_secs(1),
+                tokio::process::Command::new(program)
+                    .args(args)
+                    .kill_on_drop(true)
+                    .output(),
+            )
+            .await;
+
+            match result {
+                Ok(Ok(output)) => output.status.success(),
+                Err(_) => {
+                    log::info!("Timed out waiting for {}", program);
+                    false
+                }
+                _ => false,
+            }
         }
 
         #[derive(Debug)]


helix-view/src/clipboard.rs
```diff
@@ -243,12 +242,37 @@ pub mod provider {
...
+            let task = async {
+                match timeout(
+                    std::time::Duration::from_secs(1),
+                    is_exit_success_async(program, args),
+                )
+                .await
+                {
+                    Ok(success) => success,
+                    Err(_) => {
+                        log::info!("Timed out waiting for {program}");
```
kirawi, line 269: The MSRV right now is 1.57

prscoelho: Woops! Fixed.

helix-view/src/clipboard.rs
```diff
@@ -243,12 +242,37 @@ pub mod provider {
...
+                {
+                    Ok(success) => success,
+                    Err(_) => {
+                        log::info!("Timed out waiting for {}", program);
+                        false
+                    }
+                }
+            };
+
+            block_on(task)
```
archseer, line 275: Needs to be wrapped with https://docs.rs/tokio/latest/tokio/task/fn.block_in_place.html

helix-view/src/clipboard.rs
```diff
@@ -243,12 +242,37 @@ pub mod provider {
...
+                    .ok()
+                    .and_then(|out| out.status.success().then(|| ()))
+                    .is_some()
+            }
+
+            let task = async {
+                match timeout(
+                    std::time::Duration::from_secs(1),
+                    is_exit_success_async(program, args),
+                )
```
archseer, line 264: Remove the async closure, instead `let task = timeout(...)` then match on the result of `block_on`

helix-view/src/clipboard.rs
```diff
@@ -243,12 +242,37 @@ pub mod provider {
...
-                .args(args)
-                .output()
-                .ok()
-                .and_then(|out| out.status.success().then(|| ())) // TODO: use then_some when stabilized
-                .is_some()
+            use helix_lsp::block_on;
+            use tokio::process::Command;
+            use tokio::time::timeout;
+
+            async fn is_exit_success_async(program: &str, args: &[&str]) -> bool {
```
archseer, line 249: Hmm, using async might actually be problematic since tests might call this function outside of a tokio runtime.

prscoelho:
Yep, I can confirm that this fixes the problem on my work computer.

helix-view/src/clipboard.rs
prscoelho: Oh no :( well, I made the changes for the rest, but let me know if you want this changed to be sync.

helix-view/src/clipboard.rs
```diff
@@ -243,12 +242,34 @@ pub mod provider {
...
+                Command::new(program)
+                    .args(args)
+                    .kill_on_drop(true)
+                    .output()
+                    .await
+                    .ok()
+                    .and_then(|out| out.status.success().then(|| ())) // TODO: use then_some when stabilized
```
the-mikedavis, line 257:
```suggestion
                    .filter(|out| out.status.success())
```

I think this ends up being a little clearer than `then_some` would be

prscoelho: That does look better. Even nicer would be replacing the two lines with `is_some_and(|out| out.status.success())`, but who knows when that will stabilize!

helix-view/src/clipboard.rs
archseer: Maybe it would be better to expose the method as an async fn then let the caller use block_on as needed? I'm just worried about the method looking sync but internally blocking on async code

archseer:
(For visibility, replied here https://github.com/helix-editor/helix/pull/3696#discussion_r1014122780)

helix-view/src/clipboard.rs
prscoelho: Yeah, that makes sense. I think what was messing with me was the cfgs, I didn't want a linux version to be async while others sync, but I see now that this function is linux only. I pushed new changes, is this what you had in mind?

prscoelho: Also, [tokio::process::Command](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.output) mentions that it doesn't have the same exact behavior as std::process::Command concerning stdout/stderr pipes, is this something we need to worry about?

prscoelho:
@archseer I pushed some changes, I dunno if you got a notification for it. Let me know for next time if you did though, not trying to spam.

pascalkuthe:
clsoing this one  as stale. Thank you for contributing!

PR #3697, opened 2022-09-04
nix: fix highlighting rules

nrdxp:
* avoid coloring `identifier`s globally
* fix function application when not part of `select_expression`
* add `has_attribute_expression` highlighting
* fix precendence for interpolation, which should be after select
* highlight `@` as delimiter

### DIFF
--- a/runtime/queries/nix/highlights.scm
+++ b/runtime/queries/nix/highlights.scm
@@ -1,16 +1,20 @@
 (comment) @comment
 
+"assert" @keyword.control.exception
+"or" @keyword.operator
+"rec" @keyword.control.repeat
+
 [
   "if" 
   "then"
   "else"
+] @keyword.control.conditional
+
+[
   "let"
   "inherit"
   "in"
-  "rec"
   "with" 
-  "assert"
-  "or"
 ] @keyword
 
 ((identifier) @variable.builtin
@@ -42,14 +46,11 @@
 (integer_expression) @constant.numeric.integer
 (float_expression) @constant.numeric.float
 
-(interpolation
-  "${" @punctuation.special
-  "}" @punctuation.special) @embedded
-
 (escape_sequence) @constant.character.escape
 
 (function_expression
   universal: (identifier) @variable.parameter
+  "@"? @punctuation.delimiter
 )
 
 (formal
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I disagree with the change that `@variable.other.member` should be only for `select_expression`. `@variable.other.member` corresponds to attribute names in Nix syntax and the highlights in master reflect the distinction between in-scope variables and attribute names fairly well

```nix
{ attr = val; }
#   ^ variable.other.member
#         ^ variable
```

nrdxp:
Actually, now that you mention it, you are right, I think I was just biased by my personal color choices :sweat_smile:

nrdxp:
Force pushed to address your request. Also, I remove the modifications to snazzy as I have made some more drastic changes and should probably be it's own PR at this point. Also, I noticed that the `@` character, which names formals was left unhighlighted so I added a rule for that as well.

runtime/queries/nix/highlights.scm
```diff
@@ -1,16 +1,22 @@
 (comment) @comment
 
+("assert") @keyword.control.exception
+
+("or") @keyword.operator
+
+("rec") @keyword.control.repeat
```
the-mikedavis, line 7: This is a bit of a nitpick but these can be rewritten without the surrounding `()`, like `"or" @keyword.operator`

runtime/queries/nix/highlights.scm
```diff
@@ -72,14 +79,18 @@
...
-  attrpath: (attrpath (identifier)) @variable.other.member)
+  attrpath: (attrpath attr: (identifier)) @variable.other.member)
 
-(identifier) @variable.other.member
+(inherit_from attrs: (inherited_attrs attr: (identifier) @variable) )
 
-(inherit_from attrs: (inherited_attrs attr: (identifier) @variable.other.member) )
+(has_attr_expression
+  expression: (_)
+  ("?") @operator
```
the-mikedavis, line 91: Similar here

```suggestion
  "?" @operator
```

the-mikedavis (approved):

the-mikedavis:
Thanks!

nrdxp:
just a heads up, I also solved the incorrect highlighting of bash `${var}` style variables upstream cstrahan/tree-sitter-nix#33

PR #3709, opened 2022-09-05
read only indicator

sbromberger:
Rebased #2862.

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -39,6 +39,13 @@ pub enum Mode {
     Insert = 2,
 }
 
+#[derive(Debug, Copy, Clone, PartialEq, Eq, Hash, Serialize)]
+pub enum WritePermission {
+    Writable,
+    ReadOnly,
+    Unknown,
+}
+
 impl Display for Mode {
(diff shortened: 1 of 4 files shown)


sbromberger:
Do we want read-only indicators in bufferline as well? It's an easy add.

helix-view/src/document.rs
```diff
@@ -388,6 +399,15 @@ impl Document {
...
         doc.detect_indent_and_line_ending();
+        // this seems overly complex but should make it easier
+        // to extend doc.file_stat with more information in the future.
+        doc.write_permission = fs::metadata(path).map_or(WritePermission::Unknown, |f| {
+            if f.permissions().readonly() {
+                WritePermission::ReadOnly
+            } else {
+                WritePermission::Writable
+            }
+        });
```
archseer, line 410: Should probably be part of `doc.set_path` since a path can also be changed after opening?

sbromberger: What are the implications of changing a path after opening?

yashpalgoyal1304:
Why did the author close this without any info?

tjkirch:
Does anyone know if there's work ongoing for a read-only indicator, or a read-only flag/option for buffers?  I can't find any info about why this PR was closed or anything happening after it.

PR #3714, opened 2022-09-05
feat(languages): jsonnet

matoous:
Add jsonnet language and syntax highlight. The best tree-sitter grammar as of now seems to be: https://github.com/sourcegraph/tree-sitter-jsonnet (used in the PR).

### DIFF
--- a/runtime/queries/jsonnet/highlights.scm
+++ b/runtime/queries/jsonnet/highlights.scm
@@ -1,0 +1,27 @@
+"if" @conditional
+[
+  (local)
+  "function"
+] @keyword
+(comment) @comment
+
+(string) @string
+(number) @number
+[
+  (true)
+  (false)
+] @boolean
+
+(binaryop) @operator
+(unaryop) @operator
+
+(id) @variable
+(param identifier: (id) @variable.parameter)
+(bind function: (id) @function)
+(fieldname) @string.special
+[
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
--- a/runtime/queries/jsonnet/highlights.scm
+++ b/runtime/queries/jsonnet/highlights.scm
@@ -1,4 +1,4 @@
-["if" "then" "else"] @conditional
+["if" "then" "else"] @keyword.control.conditional
 [
   (local)
   "function"
@@ -6,11 +6,11 @@
 (comment) @comment
 
 (string) @string
-(number) @number
+(number) @constant.numeric
 [
   (true)
   (false)
-] @boolean
+] @constant.builtin.boolean
 
 (binaryop) @operator
 (unaryop) @operator
--- a/runtime/queries/jsonnet/highlights.scm
+++ b/runtime/queries/jsonnet/highlights.scm
@@ -1,4 +1,4 @@
-"if" @conditional
+"if" @keyword.control.conditional
 [
   (local)
   "function"
@@ -6,11 +6,11 @@
 (comment) @comment
 
 (string) @string
-(number) @number
+(number) @constant.numeric
 [
   (true)
   (false)
-] @boolean
+] @constant.builtin.boolean
 
 (binaryop) @operator
 (unaryop) @operator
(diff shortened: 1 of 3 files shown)


runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,27 @@
+"if" @conditional
+[
+  (local)
+  "function"
+] @keyword
+(comment) @comment
+
+(string) @string
+(number) @number
```
the-mikedavis, line 9:
```suggestion
(number) @constant.numeric
```

runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,27 @@
...
+[
+  (true)
+  (false)
+] @boolean
```
the-mikedavis, line 13:
```suggestion
] @constant.builtin.boolean
```

runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,27 @@
+"if" @conditional
```
the-mikedavis, line 1:
```suggestion
"if" @keyword.control.conditional
```

runtime/queries/jsonnet/highlights.scm
matoous: Thanks for suggestions!

the-mikedavis (commented):
I played around with the queries a bit and I have a few more highlights that I think look good:

```scm
"for" @keyword.control.repeat
"in" @keyword.operator
[(self) (dollar)] @variable.builtin
"assert" @keyword
(null) @constant.builtin
[
  ":"
  "::"
  ";"
  "="
] @punctuation.delimiter
```

What do you think?

runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,27 @@
+"if" @keyword.control.conditional
```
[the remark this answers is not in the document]
matoous: Nice, added, I also added `else`, does that make sense? Jsonnet uses the `if ... then ... else ...` syntax.

matoous:
Screenshot 2022-09-07 at 09 24 46
@the-mikedavis looks better now, thanks a lot for suggestions. Some highlights still seem weird though but I can't tell if it's the `highlights.scm` file or it the grammar has some space for improvement as well 🤔

runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,27 @@
...
+  (false)
+] @constant.builtin.boolean
+
+(binaryop) @operator
+(unaryop) @operator
+
+(id) @variable
+(param identifier: (id) @variable.parameter)
+(bind function: (id) @function)
+(fieldname) @string.special
```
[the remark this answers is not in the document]
matoous: done, thanks

the-mikedavis (commented):
Oh hmm looks like that last commit rolled back some of the prior changes like `conditional` being `keyword.control.conditional` and number being `constant.numeric`. I think switching back to those scopes should improve how the highlights look.

It also looks like the parser could use some improvements too though. I don't think it's possible to robustly distinguish variables from properties like you can in JavaScript (example: `self.some_property`)

matoous:
@the-mikedavis you are right, sorry, I somehow overwrote them when adjusting further stuff. Should be good now.

matoous:
Screenshot 2022-09-08 at 09 21 21

Looks good now, let me know if there's anything else I should adjust 🙏

runtime/queries/jsonnet/highlights.scm
```diff
@@ -0,0 +1,38 @@
...
+  (true)
+  (false)
+] @constant.builtin.boolean
+
+(binaryop) @operator
+(unaryop) @operator
+
+(param identifier: (id) @variable.parameter)
+(bind function: (id) @function)
+(fieldname) @string.special
```
the-mikedavis, line 20:
```suggestion
(fieldname (id) @variable.other.member)
```

the-mikedavis (approved):

the-mikedavis:
Thanks @matoous!

PR #3720, opened 2022-09-05
Marks

the-mikedavis:
Marks are a way of recording selections/ranges in a way that can be used later. It's like the ability to store jumplist entry in a registers plus commands to combine selections to/from registers. This PR implements [kakoune's marks commands](https://github.com/mawww/kakoune/blob/master/doc/pages/keys.asciidoc#marks) and also adds a function that joins the ranges of a given selection together.

I would like to explore basing LSP snippets on marks so I'll keep this as a draft for now.

Questions:

* The implementation (prior to this change) knows "multiple selections" as one `helix_core::selection::Selection` with many `helix_core::selection::Range`s. Colloquially we call this multiple selections though. Should the marks commands say "selections" or "ranges"?
* Where to put the marks commands? `z`/`Z` is taken and the normal mode keymap is pretty much full. I'm using `^` for now which is a bit hidden but it's nice that it matches the default marks register.

Closes https://github.com/helix-editor/helix/issues/703

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -713,6 +713,19 @@ impl Selection {
 
         Selection::new(ranges, self.primary_index)
     }
+
+    /// Creates a new selection in which all ranges are merged into one.
+    ///
+    /// The new range covers from the lowest [Range::from] to the highest
+    /// [Range::to].
+    pub fn join_ranges(&self) -> Selection {
+        // SAFETY: selections created with Selection::new are asserted to have
+        // non-empty range vectors.
+        let first = self.ranges.first().unwrap();
+        let last = self.ranges.last().unwrap();
+
+        Selection::new(smallvec![first.union(*last)], 0)
+    }
 }
 
 impl<'a> IntoIterator for &'a Selection {
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -211,10 +211,14 @@ impl Range {
 
     /// Returns a range that encompasses both input ranges.
     ///
-    /// This is like `extend()`, but tries to negotiate the
-    /// anchor/head ordering between the two input ranges.
+    /// This is like `extend()`, but negotiates the anchor/head ordering
+    /// between the two input ranges. The returned range covers the area
+    /// of both input ranges and any space between them.
+    ///
+    /// The range is [Direction::Backward] if both input ranges are
+    /// [Direction::Backward], [Direction::Forward] otherwise.
     #[must_use]
-    pub fn merge(&self, other: Self) -> Self {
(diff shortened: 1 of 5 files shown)


EpocSquadron:
IMHO, Range should be renamed to Selection in code for confirmation with how is referred to colloquially. Correspondingly, Selection should be renamed simply to Selections, plural. But that's probably a separate PR. Regardless I think the user facing language should follow what people are saying.

EpocSquadron:
There's some really interesting discussion about how kakoune's marks could be improved or generalized in [this post on the kakoune discourse](https://discuss.kakoune.com/t/array-iterator-and-looping-behaviour-in-kak-script/1287/18?u=epocsquadron) that I think should be thought through before we commit to exactly the current kakoune implementation here. I'd like our version to be an improvement rather than just a copy.

the-mikedavis:
The top-level folding ideas look like marks ideas applied to the ranges of the current selection (rather than the current selection and the selection in the register). `^j` in this PR implements `<A-r>u`. I'm not sure intersection makes sense since you can't have overlapping ranges in a single selection. It could be used as the space between all consecutive ranges in a selection I suppose.

Since only `s`/`S` are taken under `^` currently, we could rename `^j` as `^u` and implement the rest to act on the ranges in the current selection only.

I'm not sure I like the reduce idea with `|`, it seems too specific and I'm having a hard time imagining a useful use-case for it.

EpocSquadron:
I'm still chewing on adding operations on the current selections. I agree intersection makes little sense there, and at least left and right most are achievable with a combination of `(`/`)` and `,`. Longest and shortest could theoretically be useful after `s` with a regex giving variable length matches, but it's pretty niche.

I agree that the pipe thing as presented is too complicated and niche. There's something to being able to iterate and accumulate over contents of current selections, but that's best as a separate issue.

In the meantime I found the [original issue](https://github.com/mawww/kakoune/issues/284) where `z`/`Z` was chosen and apparently it started as `^` too. I kind of agree it's not super ergonomic. What about `$`? Looks like the s in selections, and we could use that as the default register instead of `^`. Perhaps even less enough though.

EpocSquadron:
One person missing that has a lot of [discussion in kakoune](https://github.com/mawww/kakoune/issues/3504) is adding the difference operation.

goyalyashpal:
can this one be considered closed?

VKondakoff:
So, is the `marks` feature abandoned?

the-mikedavis:
I haven't looked at this or worked on it in a while. It's still a useful feature I'd like to see implemented eventually but I don't plan to work on it any time soon. So I'll close this out.

Anyone interested can pick up this PR but it needs some work: we need to map the saved ranges through changes to their documents.

PR #2128, opened 2022-04-16
add reflow command

vlmutolo:
Users need to be able to hard-wrap text for many applications, including
comments in code, git commit messages, plaintext documentation, etc. It
often falls to the user to manually insert line breaks where appropriate
in order to hard-wrap text.

This commit introduces the "reflow" command (both in the TUI and core
library) to automatically hard-wrap selected text to a given column length.

Addresses #625.

PR #3738, opened 2022-09-07
Add good default rulers when editing git messages

CBenoit:
Around 50 columns for the summary is good because it is often used as heading or as subject in emails. 72 columns for the body is generally good because some tools do not wrap long lines (`git log` with pager `less` is a good example). Helix's `:reflow` command is really good to help with the second point.

Linux kernel documentation says:

> For these reasons, the ``summary`` must be no more than 70-75 characters, and it must describe both what the patch changes, as well as why the patch might be necessary.  It is challenging to be both succinct and descriptive, but that is what a well-written summary should do.

Source:
https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/tree/Documentation/process/submitting-patches.rst#n627

tpope:
https://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

Commit message style guide for Git:
https://commit.style/

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -972,6 +972,8 @@ roots = []
 file-types = ["COMMIT_EDITMSG"]
 comment-token = "#"
 indent = { tab-width = 2, unit = "  " }
+rulers = [50, 72]
+max-line-length = 72
 
 [[grammar]]
 name = "git-commit"


sudormrfbin (approved):

the-mikedavis:
Speaking of `:reflow`, should we also add the key `max-line-length = 72` which `:reflow` uses as a default when the number of columns is not specified? https://github.com/helix-editor/helix/pull/2128/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1028-R1036

CBenoit:
> Speaking of `:reflow`, should we also add the key `max-line-length = 72` which `:reflow` uses as a default when the number of columns is not specified? https://github.com/helix-editor/helix/pull/2128/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1028-R1036

Good idea, I updated the PR with that 👍

PR #3739, opened 2022-09-07
Add darcula theme based upon Intelij Darcula

nogden:


runtime/themes/darcula.toml
```diff
@@ -0,0 +1,85 @@
...
+"diff.plus" = "green"
+"diff.delta" = "grey"
+"diff.minus" = "red"
+
+"diagnostic" = { modifiers = ["underlined"] }
+"info" = "grey0D"
+"hint" = "grey03"
+"debug" = "grey03"
+"warning" = "grey09"
+"error" = "grey08"
```
CBenoit, line 65: Looks like these colors are missing from the palette

CBenoit (commented):
Hi, thank you!

`warning` and `error` are used for comment tags highlighting (stuff like `TODO:`…).

See when using dark_plus:
image

There is no visual difference when using this theme:
image

Indeed, I don’t see the colors `grey09` and `grey08` referenced in your theme. An oversight?

the-mikedavis (commented):
I also see some UI keys missing (the table at the bottom of this page: https://docs.helix-editor.com/master/themes.html#syntax-highlighting), `ui.virtual` or `ui.virtual.{whitespace,indent-guide}` as well as `ui.cursorline.primary` and `ui.statusline.{insert,normal,select}` for color-modes. The features that use those colors (visible whitespace, indent guides, cursorline, and color-modes, respectively) are disabled by default so you'll have to enable them to see how the colors look.

nogden:
Thanks for the feedback folks. I've updated the theme to add styling for all of these issues.

runtime/themes/darcula.toml
```diff
@@ -0,0 +1,90 @@
...
+"ui.help" = { fg = "grey04", bg = "grey01" }
+"ui.cursor" = { fg = "grey04", modifiers = ["reversed"] }
+"ui.cursor.primary" = { fg = "grey05", modifiers = ["reversed"] }
+"ui.cursor.match" = { fg = "white", modifiers = ["underlined"] }
+"ui.cursorline.primary" = { bg = "grey02" }
+"ui.cursorline.secondary" = { bg = "grey02" }
+"ui.text" = "white"
+"ui.text.focus" = "grey05"
+"ui.virtual.ruler" = { bg = "grey02" }
+"ui.virtual.indent-guide" = "grey02"
```
the-mikedavis, line 24: You may want to add a `ui.virtual.whitespace` key (`:set whitespace.render all`). Without it the whitespace looks pretty bright - I think it would look good if it were slightly brighter than the indent guides color. What do you think?

nogden: I've  now added the whitespace styling as requested. I don't really have an opinion on it as it's a feature that I've never needed myself, but I've gone for the next shade lighter than the indent guides. It looks fine to me.

the-mikedavis (approved):
Looks great, thanks!

the-mikedavis:

PR #3740, opened 2022-09-07
commands: Make no arg ':theme' show name

jamesodhunt:
Most commands that accept an argument show their current value if no argument is specified. The `:theme` command previously displayed an error message in the status bar if not provided with an argument:

```
Theme name not provided
```

It now shows the current theme name in the status bar if no argument is specified.

Signed-off-by: James O. D. Hunt <jamesodhunt@gmail.com>

### DIFF
--- a/helix-view/src/theme.rs
+++ b/helix-view/src/theme.rs
@@ -14,19 +14,14 @@ use toml::{map::Map, Value};
 use crate::graphics::UnderlineStyle;
 pub use crate::graphics::{Color, Modifier, Style};
 
-pub static DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| {
-    //    let raw_theme: Value = toml::from_slice(include_bytes!("../../theme.toml"))
-    //        .expect("Failed to parse default theme");
-    //    Theme::from(raw_theme)
-
-    toml::from_slice(include_bytes!("../../theme.toml")).expect("Failed to parse default theme")
+pub static DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| Theme {
+    name: "default".into(),
+    ..toml::from_slice(include_bytes!("../../theme.toml")).expect("Failed to parse default theme")
 });
-pub static BASE16_DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| {
-    //    let raw_theme: Value = toml::from_slice(include_bytes!("../../base16_theme.toml"))
-    //        .expect("Failed to parse base 16 default theme");
-    //    Theme::from(raw_theme)
 
-    toml::from_slice(include_bytes!("../../base16_theme.toml"))
+pub static BASE16_DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| Theme {
+    name: "base16_theme".into(),
+    ..toml::from_slice(include_bytes!("../../base16_theme.toml"))
         .expect("Failed to parse base 16 default theme")
 });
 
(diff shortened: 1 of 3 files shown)


the-mikedavis:
The description of the command should be updated to reflect that it can return the current theme when no arguments are passed

helix-view/src/theme.rs
```diff
@@ -157,6 +169,10 @@ impl Theme {
         self.highlights[index]
     }
 
+    pub fn name(&self) -> String {
+        self.name.to_string()
+    }
```
archseer, line 174: Should return `&str`

jamesodhunt: Done, but I'm still converting that value to a string to avoid rust lifetime issues wrt the context. This method is only going to be called infrequently so conversion to a string seems reasonable to me.

jamesodhunt:
@the-mikedavis - Description updated.

helix-view/src/theme.rs
the-mikedavis: It's more colloquial with the rest of the codebase. It's better to leave it to the consumer of `name` to decide when to clone. `String` is ok with the current `cx.editor.set_status` since it needs it for lifetimes as you say but if instead the caller used `format!("The theme is {}", theme.name())`, you would be re-allocating it twice rather than once.

helix-view/src/theme.rs
```diff
@@ -157,6 +170,10 @@ impl Theme {
         self.highlights[index]
     }
 
+    pub fn name(&self) -> Cow<str> {
+        Cow::Borrowed(self.name.as_ref())
+    }
```
the-mikedavis, line 175:
```suggestion
    pub fn name(&self) -> &str {
        &self.name
    }
```

jamesodhunt: @the-mikedavis - Updated.

helix-view/src/theme.rs
```diff
@@ -12,11 +12,14 @@ use toml::Value;
 
 pub use crate::graphics::{Color, Modifier, Style};
 
-pub static DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| {
-    toml::from_slice(include_bytes!("../../theme.toml")).expect("Failed to parse default theme")
+pub static DEFAULT_THEME: Lazy<Theme> = Lazy::new(|| Theme {
+    name: "theme".into(),
```
the-mikedavis, line 16:
```suggestion
    name: "default".into(),
```

jamesodhunt: Fixed.

jamesodhunt:
Anything else to do on this PR?

EpocSquadron:
LGTM

jamesodhunt:
Branch rebased to resolve the conflict.

jamesodhunt:
This has been open a month now. The checks have passed and there are no unanswered question fwics.

It's not a major feature, but I do feel it helps usability (it helped me to develop my own theme atleast).

Any chance it could land soon to potentially help others? :smile:

helix-view/src/theme.rs
```diff
@@ -265,6 +266,14 @@ impl Theme {
         self.highlights[index]
     }
 
+    pub fn name(&self) -> &str {
+        &self.name
+    }
+
+    pub fn with_name(self, name: String) -> Self {
+        Theme { name, ..self }
+    }
```
the-mikedavis, line 275: I think this helper is not necessary - I don't think it will be used elsewhere in the future. In the usage above this can be done inline

jamesodhunt: @the-mikedavis - branch updated.

the-mikedavis (approved):

PR #3743, opened 2022-09-08
fix: align view after jumplist_picker

shnarazk:
Since `jumplist_picker` doesn't update screen, we need to do one action to get the correct view.
So I added 'align_view' to the callback like `jump_backward`.

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -1,4 +1,4 @@
-use crate::{editor::GutterType, graphics::Rect, Document, DocumentId, ViewId};
+use crate::{align_view, editor::GutterType, graphics::Rect, Align, Document, DocumentId, ViewId};
 use helix_core::{
     pos_at_visual_coords, visual_coords_at_pos, Position, RopeSlice, Selection, Transaction,
 };
@@ -170,6 +170,15 @@ impl View {
         doc: &Document,
         scrolloff: usize,
     ) -> Option<(usize, usize)> {
+        self.offset_coords_to_in_view_center(doc, scrolloff, false)
+    }
+
+    pub fn offset_coords_to_in_view_center(
+        &self,
+        doc: &Document,
+        scrolloff: usize,
+        centering: bool,
+    ) -> Option<(usize, usize)> {
         let cursor = doc
             .selection(self.id)
             .primary()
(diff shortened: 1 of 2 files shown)


shnarazk:
OK. I added `View::offset_coords_to_in_view_for_search` that calculates offsets for two `scrolloff`s, based on `offset_coords_to_in_view`. And all of `search_impl` and 3 jump commands were revised to use another new method `View::place_cursor_in_view`, which tries to fix the view. 
Now, this PR is about 'aligning view after jump commands'...

archseer:
Hmm this seems to have some other commits mixed in now?

shnarazk:
@archseer 
Oh, sorry. I've just rebased and squashed. Please check this again.

helix-view/src/view.rs
```diff
@@ -201,6 +202,53 @@ impl View {
...
+                // scroll right
+                self.offset.col + col - (last_col.saturating_sub(scrolloff))
+            } else if col < self.offset.col + scrolloff {
+                // scroll left
+                col.saturating_sub(scrolloff)
+            } else {
+                self.offset.col
+            };
+            (row, col)
+        };
```
the-mikedavis, line 248: It looks like this code ends up being mostly duplicated between this and the above function. Can we factor out the shared parts into a function they both call?

the-mikedavis (commented):
This behavior seems like it might be a good default. Should we just use this behavior for all instances of `view.ensure_cursor_in_view(doc, scrolloff)`?

helix-view/src/view.rs
shnarazk: @the-mikedavis 
Hmmm, `offset_coords_to_in_view` is a simplified version of this, more generic function. So I revised `offset_coords_in_view_center`to take an extra argument, and `offset_coords_to_in_view` as a wrapper for it.
Also, I renamed new methods and changed their args.

the-mikedavis:
It looks like there's a regression with the behavior of scrolloff - if you hold `j` on a document longer than the screen, helix behaves as if it's configured to use a scrolloff of 0

shnarazk:
@the-mikedavis 
My bad. The else block of the last if statement was revised to use `scrolloff`.

the-mikedavis (approved):

PR #3749, opened 2022-09-08
Add "markdown" as a possible filetype

Akselmo:
Markdown files can also end with `.markdown`.

This fixed the highlighting not being enabled in files that end with `.markdown` for me.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #3752, opened 2022-09-08
add option to automatically `hover`

crockeo:
Hi! Thanks for this project, I've been trying it out the last week and really enjoying it 🙂 

This PR introduces the option to enable `auto-hover`, similar to `auto-completion` but for the `hover` command. This is an intrusive option so I've kept it `false` by default, but I've been using on this PR and found it quite useful!

This is my first PR to the project, so there's some design questions I had but didn't know how to answer. I'll leave them inline so it's easier to cross-reference w/ the code in question.

### DIFF
--- a/helix-term/src/compositor.rs
+++ b/helix-term/src/compositor.rs
@@ -229,6 +229,13 @@ impl Compositor {
             .any(|component| component.type_name() == type_name)
     }
 
+    pub fn has_component_by(&self, predicate: impl Fn(&str) -> bool) -> bool {
+        self.layers
+            .iter()
+            .map(|component| component.type_name())
+            .any(predicate)
+    }
+
     pub fn find<T: 'static>(&mut self) -> Option<&mut T> {
         let type_name = std::any::type_name::<T>();
         self.layers
@@ -242,10 +249,6 @@ impl Compositor {
             .iter_mut()
             .find(|component| component.id() == Some(id))
             .and_then(|component| component.as_any_mut().downcast_mut())
-    }
-
-    pub fn has_overlay(&self) -> bool {
-        self.layers.len() > 1
     }
 }
 
--- a/helix-term/src/compositor.rs
+++ b/helix-term/src/compositor.rs
@@ -243,6 +243,10 @@ impl Compositor {
             .find(|component| component.id() == Some(id))
             .and_then(|component| component.as_any_mut().downcast_mut())
     }
+
+    pub fn has_overlay(&self) -> bool {
+        self.layers.len() > 1
+    }
 }
 
 // View casting, taken straight from Cursive
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1006,13 +1006,7 @@ impl EditorView {
     }
 
     pub fn handle_idle_timeout(&mut self, cx: &mut crate::compositor::Context) -> EventResult {
-        if self.completion.is_some()
-            || cx.editor.mode != Mode::Insert
-            || !cx.editor.config().auto_completion
-        {
-            return EventResult::Ignored(None);
-        }
-
+        let mut result = EventResult::Ignored(None);
         let mut cx = commands::Context {
             register: None,
             editor: cx.editor,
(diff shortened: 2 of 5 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -1021,9 +1015,21 @@ impl EditorView {
             callback: None,
             on_next_key_callback: None,
         };
```
crockeo, line 1017: Is it ok to unconditionally construct a context here? It doesn't look like it's mutative in any way--so was this just about avoiding unnecessary stack allocations before?

helix-term/src/compositor.rs
```diff
@@ -243,6 +243,10 @@ impl Compositor {
             .find(|component| component.id() == Some(id))
             .and_then(|component| component.as_any_mut().downcast_mut())
     }
+
+    pub fn has_overlay(&self) -> bool {
+        self.layers.len() > 1
+    }
```
crockeo, line 249: This feels like code smell, but I don't know the right way to think about this. If we don't check to see if there's an overlay, then auto-hover will paint on top of any active pickers. Is there a better way to check for multiple layers?

kirawi: There's a `has_component` function. You should probably check for pickers instead of overlay though.

helix-term/src/ui/editor.rs
kirawi: I feel that this is a bit ugly and causes some code duplication.

helix-term/src/compositor.rs
sudormrfbin: IIRC you have to check for an `Overlay<Picker>` type instead of a `Picker` type with `has_component()` since pickers area always wrapped in an overlay.

crockeo: Thank you both! 🙂

crockeo: Ok, did an experiment with this. It may not be feasible to use `has_component` to check for an occluding layer without enumerating every possible `Picker`. I did a quick check on `type_name()`s to verify if you could look for `Overlay<Picker>` and it looks like it would have to be the fully-qualified, fully-specialized names, e.g.:

```
// for a file picker
helix_term::ui::overlay::Overlay<helix_term::ui::picker::FilePicker<std::path::PathBuf>>

// for a symbol picker
helix_term::ui::overlay::Overlay<helix_term::ui::picker::FilePicker<lsp_types::document_symbols::SymbolInformation>>
```

kirawi: You could implement `id()` for `Overlay` and `Picker` and write a `has_id()` function. The `Overlay` id could be `format!("Overlay<{}>", T::id())`, and the `Picker` id could just be `"Picker"`. Alternatively, you could have `Overlay` just return `"Overlay"`.

sudormrfbin: `Component::id()` is supposed to uniquely identify the component if supplied, so this might not be a good idea. Maybe there can be a variant of `has_component()` that uses a predicate:

```rust
    pub fn has_component_by(&self, predicate: impl Fn(&str) -> bool) -> bool {
        self.layers
            .iter()
            .map(|component| component.type_name())
            .any(predicate)
    }
```

and you would use it like `compositor.has_component_by(|c| c.contains("Picker"))`. Well TBH this is a bit hacky too :)

helix-term/src/ui/editor.rs
crockeo: Sorry for the delay! I've been on vacation and trying not to program as much. I've got the itch again, though, so I'm back 🙂 

Do you mind elaborating more on what part is ugly/what part causes code duplication? If you're talking about the disjoint `if` statements below: I agree, but I tend to prefer duplication over a poor abstraction. I can't think of a good way to represent this without generalizing the concept of a timer which is a much bigger task to bite off.

With all of that said: please let me know if you have another preferred method of representing this! I'm not a regular contributor here so I'm happy to defer to those who are.

helix-term/src/compositor.rs
crockeo: I do like that slightly more--it's at least more generally useful than `has_overlay`. I'll give it a shot and push it up here.

crockeo:
Replied to the comments above. Thanks y'all for taking the time to review this PR!

crockeo:
closing this because it has conflicts and i don't have the time to fix them now-a-days. thanks y'all for taking a look!

Issue #3744, opened 2022-09-08
`append` is not consistent with rest of insert mode bindings


LollipopFt:
### Summary

`i` command: 
![image](https://user-images.githubusercontent.com/62802897/189123791-ee3ae987-c7f7-4c3b-8be4-7efe938c5974.png)
(no selection)

`I` command:
![image](https://user-images.githubusercontent.com/62802897/189123882-c06d5e07-b5a9-4d7c-94fe-634c33689290.png)
(no selection)

`a` command:
![image](https://user-images.githubusercontent.com/62802897/189124001-92f66893-5e7c-47a1-8523-bc09fdfe7b67.png)
(selection of letter, in this case `t`)
this problem is worse if i go to end of line using command `gl`, press `a` then enter and continue typing. the entire text is selected.
![image](https://user-images.githubusercontent.com/62802897/189124355-ec573643-8c16-445f-b146-3d8b27c68bac.png)

meanwhile, the other `append` command `A`:
![image](https://user-images.githubusercontent.com/62802897/189124534-a50a875d-7eb6-44e2-86b1-a5850a9ee5bf.png)
(no selection)

could `a` be changed to not select any text like all the other insert commands? everytime i use `a`, i need to move my cursor with arrow keys to remove the selection. shouldn't insert mode not be selecting any text? normal mode and visual mode should be doing that. this applies across all color schemes, nothing to do with my custom color scheme.

### Platform

Windows

### Terminal Emulator

Windows Terminal Preview 1.15.2282.0

### Helix Version

helix 22.08.1 (66276ce6)

the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/issues/1411
See also https://github.com/helix-editor/helix/issues/2673 and https://github.com/helix-editor/helix/discussions/2924

Appending is different from inserting: append retains the anchor (the start of a selection) and newly added text extends the head of the selection. This is useful in multi-cursor scenarios - if `a` behaved like `i`, you would discard any selection you had when pressing `a`. `A` (`append_to_line`) is actually insertion before the end-of-line character rather than appending and is roughly equivalent to `glli`. The description and name of the command are currently wrong ([Kakoune's description](https://github.com/mawww/kakoune/blob/021da117cf90bf25b65e3344fa8e43ab4262b714/README.asciidoc#38-changes) is more accurate).

You can bind `a` to what you expect like so:

```toml
# ~/.config/helix/config.toml
[keys.normal]
[...]

PR #3753, opened 2022-09-08
Rename I/A "Insert at start/end of line"

the-mikedavis:
The language for the `A` binding is potentially confusing because `A` behaves like `i` done at the end of the line rather than `a`. This change renames the command to match Kakoune's language[^1].

This is a breaking change since the name of the command changes.

See https://github.com/helix-editor/helix/issues/3744#issuecomment-1240726094

[^1]: https://github.com/mawww/kakoune/blob/021da117cf90bf25b65e3344fa8e43ab4262b714/src/normal.cc#L2229


Omnikar:
`prepend_to_line` and `append_to_line` went together before; now it seems a bit awkward.

the-mikedavis:
Hmm yeah, it's probably best to also use Kakoune's wording for `I` as well

PR #3759, opened 2022-09-09
Add godot resource support

erasin:
godot-resource hightlight for tscn, tres

- https://github.com/PrestonKnopp/tree-sitter-godot-resource

### DIFF
--- a/runtime/queries/godot-resource/highlights.scm
+++ b/runtime/queries/godot-resource/highlights.scm
@@ -6,7 +6,7 @@
 
 (string) @string
 (integer) @constant.numeric.integer
-(float) @constant.numberic.float
+(float) @constant.numeric.float
 
 (true) @constant.builtin.boolean
 (false) @constant.builtin.boolean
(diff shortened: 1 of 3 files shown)


runtime/queries/godot-resource/highlights.scm
```diff
@@ -0,0 +1,28 @@
+(section (identifier) @type.builtin)
+
+(attribute (identifier) @attribute)
+(property (path) @attribute)
+(constructor (identifier) @constructor)
+
+(string) @string
+(integer) @constant.numeric.integer
+(float) @constant.numberic.float
```
the-mikedavis, line 9:
```suggestion
(float) @constant.numeric.float
```

runtime/queries/godot-resource/highlights.scm
```diff
@@ -0,0 +1,28 @@
+(section (identifier) @type.builtin)
+
+(attribute (identifier) @attribute)
+(property (path) @attribute)
```
the-mikedavis, line 4:
```suggestion
(property (path) @variable.other.member)
```

I think this matches a bit better based on other uses of property. What do you think?

the-mikedavis (approved):

PR #3760, opened 2022-09-09
Add textobject for gdscript

erasin:
- Add textobjects.scm
- Fix class highlight
- Fix set default indent to `\t`

### DIFF
--- a/runtime/queries/gdscript/textobjects.scm
+++ b/runtime/queries/gdscript/textobjects.scm
@@ -1,0 +1,14 @@
+
+(class_definition
+  (body) @class.inside) @class.around
+
+(function_definition
+  (body) @function.inside) @function.around
+
+(parameters
+  (typed_parameter 
+    (identifier) @parameter.inside
+  ) @parameter.around)
+
+(comment) @comment.inside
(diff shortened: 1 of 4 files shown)


runtime/queries/gdscript/textobjects.scm
```diff
@@ -0,0 +1,14 @@
...
+(class_definition
+  (body) @class.inside) @class.around
+
+(function_definition
+  (body) @function.inside) @function.around
+
+(parameters
+  (typed_parameter 
+    (identifier) @parameter.inside
+  ) @parameter.around)
```
the-mikedavis, line 11: We should probably cover untyped parameters as well

```scm
(parameters (identifier) @parameter.inside @parameter.around)
```

erasin: thanks , I will Add this ...


```diff
(parameters 
  [
    (typed_parameter)
    (identifier)    
    (default_parameter)    
    (typed_default_parameter)  
  ] @parameter.inside @parameter.around)
```

the-mikedavis (approved):

Issue #3346, opened 2022-08-06
Treat runtime files from installation as immutable


the-mikedavis:
We should treat the grammar sources, compiled parsers, and queries produced while building Helix as immutable defaults which can be overridden with the existence of files in `~/.config/helix/runtime`. Effectively, `$HELIX_RUNTIME` should act like `$PATH` in unix.

Except when installing from source, these files almost always end up in some `/usr` directory that needs sudo to edit (which you probably shouldn't do). With Nix, these end up in the store which is immutable so you need to build from source to modify queries or themes locally. Instead you should just be able to override files in your `~/.config/helix/runtime` directory.

See https://github.com/helix-editor/helix/issues/3310#issuecomment-1203875141
See https://github.com/helix-editor/helix/issues/2091#issuecomment-1097472407

amtoine:
as in #4128, i have this very issue :open_mouth: 

from the Arch main repositories, `helix` comes installed with all its grammar in `/var/lib/helix/runtime/` which needs `sudo` privileges to do anything...
thus `hx --grammar fetch/build` does not work and `sudo helix --grammar fetch/build` has to be used :confused: 

always using `~/.config/helix/runtime/` and `$HELIX_RUNTIME` would be amazing :star_struck:

amtoine:
hello there :yum: 

- **is anyone working on this somewhere?** :open_mouth: 
- **is there any chance i can help?**
i do not know where to start really, but i'd like to see such feature in `helix`.
that would be amazing to avoid messing around with `sudo` :star_struck: 

cheers :wave:

domust:
Hi there,

I would also love to have this feature in `helix`.

Currently I'm trying to add a new language to `helix` while implementing `tree-sitter` grammar for it as well,
but I can't test it on my machine since I'm using NixOS.

If anyone could give me any pointers, I would try to help implementing this feature.

paul-scott:
I had a look at this. What we could do is change `helix_loader::runtime_dir` and have it return a Vec of directories in a particular priority. I thought it was going to be an easy fix but unfortunately it is used in 10 places across 4 sub-crates so would take a little bit of tinkering. In addition / alternatively we could write a public function that, given a relative path, searches for the full path of the first file found according to the runtime directory priority (or else an error). This would make it easier for the call sites.

[...]

Issue #3310, opened 2022-08-03
`hx --grammar fetch` fails when using Helix built with Nix


AumyF:
### Summary

`helix --grammar fetch` fails when using Helix built with Nix, since its runtime directory is set in Nix store which is not writable.

### Reproduction Steps

1. `nix shell github:helix-editor/helix`
2. `hx --grammar fetch`

### Helix log

```
Error: Failed to fetch some grammar(s)

Caused by:
    0: Could not create grammar directory "/nix/store/5znz3zwidqndnqa54ibhv1yfgxd5xmip-helix-runtime/grammars/sources/protobuf"
    1: Permission denied (os error 13)
```

### Platform

Arch Linux

### Terminal Emulator

foot 1.12.1

### Helix Version

helix 22.05

archseer:
There's a previous issue for this but it was closed: https://github.com/helix-editor/helix/issues/2091#issuecomment-1097472407

the-mikedavis:
You don't need to use `hx -g fetch` when using the Nix flake, grammars are fetched and built by nix when building `hx`. The runtime for nix is in the store which is immutable, so `hx -g fetch
` and `hx -g build` fail.

I don't think there's an issue for it yet but like archseer says in the linked commit, we want to treat the runtime directory from installation (usually into `/usr` but `/nix/store` for nix) as immutable and have `~/.config/helix` or other XDG dirs act as overrides like it does currently for configuration files like `languages.toml`

tgharib:
`hx --health` to confirm highlighting (from treesitter) is working

the-mikedavis:
I'll close this out as expected behavior but I've opened up #3346 describing how we'd like to do this in the long-run. With a solution to that, `hx -g fetch`/`hx -g build` should act on `~/.config/helix/runtime` rather than the Nix store.

PR #3771, opened 2022-09-09
add `bass` language + highlighting

vito:
Hiya, I'm working on an admittedly pretty obscure language called [Bass](https://bass-lang.org). I just finished writing a tree-sitter grammar and highlighter for it, and then got the itch to try out Helix, having been a fan of Kakoune in the past. I'm really impressed with the out-of-the-box experience. :)

I initially tried configuring this in my own `~/.config/helix/languages.toml` but I'm on NixOS so the grammar install path is read-only. Adding it directly to Helix seems to be the only way at the moment. (#3310 & #3346)

My language is still very young so I'll take no offense if you prefer to just close this, but I figured I'd shoot my shot!

### DIFF
--- a/runtime/queries/bass/highlights.scm
+++ b/runtime/queries/bass/highlights.scm
@@ -1,25 +1,47 @@
 ; GENERATED VIA https://github.com/vito/tree-sitter-bass
 
+;;; comments
+
 (comment) @comment.line
+
+;;; punctuation
 
 ["(" ")" "[" "]" "{" "}"] @punctuation.bracket
 
-[(ignore) (null) (bool)] @constant.builtin
+;;; constants
 
-(int) @constant.numeric
+[(ignore) (null)] @constant.builtin
 
-(string (string_escape) @constant.character.escape)
+(bool) @constant.builtin.boolean
+
+(int) @constant.numeric.integer
+
+;;; strings
+
+;; string literals
 
 (string) @string
+(string (string_escape) @constant.character.escape)
 
-[(command) (path) (relpath)] @namespace
+;; keywords (symbol literals)
 
 (keyword) @string.special.symbol
 
+;; paths
 
-;;; specific queries take precedence
+(dot) @string.special.path
+(dotdot) @string.special.path
+(command) @string.special.path
+(subpath (symbol) @string.special.path)
 
-;; classification based highlighting
+; slashes in a path denote a combiner call
+(subpath (slash) @function)
+
+
+
+;;; specific highlighting for builtins & special forms
+
+;; symbol classification based highlighting
 
 (list . (symbol) @keyword.control.conditional (#match? @keyword.control.conditional "^(if|case|cond|when)$"))
 (cons . (symbol) @keyword.control.conditional (#match? @keyword.control.conditional "^(if|case|cond|when)$"))
@@ -67,7 +89,7 @@
   (_)
   (symbol) @function)
 
-;;; generic highlight rules
+;;; generic highlighting for all forms
 
 ; first symbol in a list form is a combiner call
 (list . (symbol) @function)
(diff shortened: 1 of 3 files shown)


the-mikedavis:
In general we try to avoid esoteric and personally-built programming languages. The compiled parser for this grammar is very small (35k on linux) though so I don't mind adding it.

runtime/queries/bass/highlights.scm
```diff
@@ -0,0 +1,98 @@
...
+
+(string) @string
+
+[(command) (path) (relpath)] @namespace
+
+(keyword) @string.special.symbol
+
+
+((list . (symbol) @keyword.control.conditional (#match? @keyword.control.conditional "^(if|case|cond|when)$"))
+ (set! "priority" 105))
```
the-mikedavis, line 21: Is the priority predicate something from nvim-treesitter? I don't think we have it implemented and I don't think it should be necessary precedence-wise

the-mikedavis (commented):
To fix the CI you will need to run `cargo xtask docgen` and commit the changes. Looks like this may need a rebase/master-merge to fix the conflicts too

runtime/queries/bass/highlights.scm
vito: Yeah, this is just needed for Neovim. It doesn't seem to cause any harm but I can adjust the template generation to only include it for Neovim, since it's definitely a bit odd to have it here (folks might look to this as an example and get the wrong idea).

Thanks for allowing this to be merged (once I tidy it up). :slightly_smiling_face:

vito:
@the-mikedavis I think this is ready now. I removed the `(set! priority ...)` calls and just redid the template to swap the order in Neovim instead. Thanks again!

runtime/queries/bass/highlights.scm
```diff
@@ -0,0 +1,79 @@
+; GENERATED VIA https://github.com/vito/tree-sitter-bass
+
+(comment) @comment.line
+
+["(" ")" "[" "]" "{" "}"] @punctuation.bracket
+
+[(ignore) (null) (bool)] @constant.builtin
```
the-mikedavis, line 7:
```suggestion
[(ignore) (null)] @constant.builtin

(bool) @constant.builtin.boolean
```

runtime/queries/bass/highlights.scm
```diff
@@ -0,0 +1,79 @@
...
+
+(int) @constant.numeric
```
the-mikedavis, line 9:
```suggestion
(int) @constant.numeric.integer
```

Unless this node also covers floats. Then `@constant.numeric` would be perfect.

runtime/queries/bass/highlights.scm
```diff
@@ -0,0 +1,79 @@
...
+
+(string (string_escape) @constant.character.escape)
+
+(string) @string
+
+[(command) (path) (relpath)] @namespace
```
the-mikedavis, line 15: For `path` and `relpath`, `string.special.path` might be more appropriate

vito:
Made the suggested changes, plus a few more revisions for more fine-grained path highlighting. Thanks for the feedback.

For some context, in Bass `foo/bar` is actually sugar for `(foo ./bar)` so I opted to highlight the `/` as a `function`. The individual path segments (`bar`) are highlighted as `strings.special.path` since those are always string-like path values.

the-mikedavis (approved):

the-mikedavis:
Thanks!

poliorcetics:
Hello @vito, I'm working on #3690 and this PR added `keyword.builtin`, can bass add keywords aside from the core language ? If not, I will change it to `keyword` only, because it's not used anywhere else, is that ok ?

vito:
@poliorcetics nope, feel free to change!

David-Else:
@vito could you add install instructions to https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers ?

vito:
@David-Else Done!

PR #3772, opened 2022-09-09
feat: rust plugins poc

AlexanderBrevig:
Playing around with an idea for writing plugins in rust and have them all be part of the binary.

Made for discussion purposes. I'll close this in a few weeks, unless we agree not to.

## High level idea

We add a plugins field to the config:

```toml
[editor.plugins.sample]
enabled = true
some-other-config = "yes"
```

Now, this will load the already compiled `SamplePlugin`.

A plugin will get its chance to override a set of callbacks. We call these for all plugins (see `Plugins` in `plugin.rs`) at the appropriate times - in order of load. This means a latter plugin takes precedence over a previous one (say if both render something custom to our view).

## Demo

The included `SamplePlugin` is a _really_ stupid ruler/watermark (as in it does not scroll) at col 20. It was just something random to have a POC going that does something graphical.

## TLDR

`Plugins::something()`  will call `do_something()` for all loaded plugins.
We call `Plugins::something()` where we want to let plugins do ... something.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -462,6 +462,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "helix-plugin"
+version = "0.6.0"
+dependencies = [
+ "helix-core",
+ "helix-tui",
+ "helix-view",
+]
+
+[[package]]
 name = "helix-term"
 version = "0.6.0"
 dependencies = [
@@ -479,6 +488,7 @@ dependencies = [
  "helix-dap",
  "helix-loader",
  "helix-lsp",
+ "helix-plugin",
  "helix-tui",
  "helix-view",
  "ignore",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,6 +7,7 @@ members = [
   "helix-lsp",
   "helix-dap",
   "helix-loader",
+  "helix-plugin",
   "xtask",
 ]
 
--- a/helix-plugin/Cargo.toml
+++ b/helix-plugin/Cargo.toml
@@ -1,0 +1,18 @@
+[package]
+name = "helix-plugin"
+version = "0.6.0"
+authors = ["Blaž Hrastnik <blaz@mxxn.io>"]
+description = """
+A library to build plugins for Helix
+"""
+edition = "2021"
+license = "MPL-2.0"
+categories = ["editor"]
+repository = "https://github.com/helix-editor/helix"
+homepage = "https://helix-editor.com"
+include = ["src/**/*", "README.md"]
+
+[dependencies]
+helix-core = { version = "0.6", path = "../helix-core" }
+helix-view = { version = "0.6", path = "../helix-view" }
+tui = { path = "../helix-tui", package = "helix-tui", default-features = false, features = ["crossterm"] }
--- a/helix-plugin/src/lib.rs
+++ b/helix-plugin/src/lib.rs
@@ -1,0 +1,2 @@
+pub mod plugin;
+pub mod sample;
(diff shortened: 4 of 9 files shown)


Triton171:
I don't think that having plugins part of the binary is a good idea. Either, we'd have to include all commonly used plugins in release builds which
- adds a lot of bloat to the binary for everybody, including people who don't need any plugins.
- still only provides a benefit to people who use the included plugins. If you want to use a new/unpopular plugin, you'd still have to compile from source.
- increases the workload of the maintainers since they now also have to approve plugins for inclusion in the main repo (this also sets an expectation that these plugins will be thoroughly reviewed which isn't really the point of a plugin system).
- wouldn't be a significant improvement over the current situation. You can already just make a PR for an optional feature which can be enabled in the configuration. Making a PR to add a plugin to the official builds would mean essentially the same.

On the other hand, we could not include any plugins, but then
- everybody who wants to use plugins has to compile Helix from source. In particular, it makes it basically impossible to package Helix (if one wants plugin support).
- You have to recompile Helix every time you want to add a new plugin to it, so quickly trying out a plugin isn't possible.

All of these issues are solved by using either WebAssembly or some scripting language for plugins.

AlexanderBrevig:
I'm not trying to be defensive or indicate I don't agree with the sentiment of your comment. I try to have a discussion not an argument (and I apologize if it feels like the latter, English is not my first language and it's not my intent to be argumentative).

> I don't think that having plugins part of the binary is a good idea. Either, we'd have to include all commonly used plugins in release builds which
> 
> * adds a lot of bloat to the binary for everybody, including people who don't need any plugins.

A WASM/plugin runtime + binding will also have an impact on size and much more so for performance.
> * still only provides a benefit to people who use the included plugins. If you want to use a new/unpopular plugin, you'd still have to compile from source.

Same with the current config options, and probably will be the same with plugins as new plugins will probably use a newer "api" version than what's currently available. With a built-in "plugin" ecosystem, there will be no version issues (for our users)
> * increases the workload of the maintainers since they now also have to approve plugins for inclusion in the main repo (this also sets an expectation that these plugins will be thoroughly reviewed which isn't really the point of a plugin system).

I think this may be a strength, the maintainers can approve the concept of a plugin (and deny others) but after inclusion, I think the maintainers of that plugin should have an easy way to update their plugin.
For instance if plugins are submodules, then the maintainers first have control over what is OK (f.eks check profanity or whatever) then after that, probably dependabot or GitHub Action could automatically merge PR's that change the submodule version, given it compiles against master.
> * wouldn't be a significant improvement over the current situation. You can already just make a PR for an optional feature which can be enabled in the configuration. Making a PR to add a plugin to the official builds would mean essentially the same.

There are PRs on hold because they are better suited as plugins (and I agree). I have a few features I want myself that are not suited to be a part of the core codebase, but would fit nicely in ~10 lines in a callback written in Rust.

> [...]
> All of these issues are solved by using either WebAssembly or some scripting language for plugins.

They do add their own challenges, especially with regards to the interface API versioning and preferrably open/closed so we never regress. Additionally the core code base will (I dont' see how not) be littered with plumbing to get whatever runtime wired in to the code-base. 

"Plugins" in this idea would be more like "modular features" than actual "dynamic features" that you can just add to Helix when you feel like it. More like suckless/DWM than ExWM.

I don't disagree with the positives for a real plugin system, but there's also negatives and this could be ready for adoption by the next Helix version.

Not at all married to this idea and ready to kill it.

AlexanderBrevig:
Dit not seem to resonate or initiate a discussion. Closing in favor of dynamic plugin approaches.

PR #3782, opened 2022-09-10
Change default TS object bindings

n0s4:
Implements https://github.com/helix-editor/helix/issues/165#issuecomment-1216566867

Changes 'match inside/around' bindings for:
- type definition from `c` to `t`
- comments from `o` to `c`
- tests from `t` to `T`

Also changes those for the `]` / `[` bindings.


the-mikedavis (approved):
I think class/type are effectively equivalent and `c` being comment makes a lot more sense to me 👍

David-Else:
This makes SO much sense! I hope it can get in before the next release so new users don't have to learn the current key bindings.

dead10ck:
This seems like a good change to me 👍

PR #3783, opened 2022-09-10
add config option to disable log

ChemicalXandco:
Appending to the log file can cause write amplification on SSDs causing lots of unnecessary wear so I propose adding a configuration entry to disable it. The proposed changes do not change the current behaviour by default however it might be beneficial to turn it off by default and only turn it on when trying to diagnose a problem.


kirawi (approved):

archseer:
> SSDs causing lots of unnecessary wear

Try looking at all the temporary files continually written by browsers :) 

Users can already stop/reduce logging by raising the log level. We currently only log errors for easier debugging (otherwise we'd have to continuously teach new users how to enable logs). The LSP was a bit noisy and would log an error if no LSP was defined but that was fixed on master.

the-mikedavis:
Now that #3807 is merged, you can use `hx --log /dev/null`

the-mikedavis:
We've further reduced logging of the really noisy stuff (https://github.com/helix-editor/helix/commit/83f09ecbff5160e4350c8099be1ad4c64513f665, https://github.com/helix-editor/helix/commit/6bfe1ddc53f542d62e242fd4aaf6748dda1b0e71) and there are workarounds now with the `--log` flag as mentioned above so I will close this out

PR #3786, opened 2022-09-10
add C-y to yank picker selection to clipboard

kirawi:
I think this is useful for easily searching diagnostics online.

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -556,6 +556,22 @@ impl<T: Item + 'static> Component for Picker<T> {
             ctrl!('t') => {
                 self.toggle_preview();
             }
+            ctrl!('y') => {
+                if let Some(contents) = self.selection() {
+                    let contents = contents.sort_text(&self.editor_data).to_string();
+                    if let Err(e) = cx
+                        .editor
+                        .clipboard_provider
+                        .set_contents(contents, helix_view::clipboard::ClipboardType::Clipboard)
+                    {
+                        cx.editor
+                            .set_status(format!("Couldn't set system clipboard content: {}", e))
+                    } else {
+                        cx.editor
+                            .set_status("yanked picker selection to system clipboard");
+                    }
+                }
+            }
             _ => {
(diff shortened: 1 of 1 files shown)


CBenoit (approved):
Looks good to me! Minor comment.

helix-term/src/ui/picker.rs
```diff
@@ -556,6 +556,22 @@ impl<T: Item + 'static> Component for Picker<T> {
             ctrl!('t') => {
                 self.toggle_preview();
             }
+            ctrl!('y') => {
+                if let Some(contents) = self.selection() {
+                    let contents = contents.sort_text(&self.editor_data).to_string();
```
CBenoit, line 561: (Oops, I forgot to add the comment. Here it is.)

Since the sort text is not always equal to the stringified label, maybe we specifically want `String::from(contents.label(&self.editor_data))` so that it always copies what is actually rendered in the picker minus the style?

crushingCodes:
i was just looking for this as i usually copy error messages often and its already a pr nice!

CBenoit:
Hi @kirawi 
Any chance you could complete this PR?

Antonio-Bennett:
@kirawi Could this be pushed ahead again? This is a super helpful PR! Just switched over from neovim and this is definitely something I miss :)

kirawi:
I will finish this PR tomorrow or the day after.

the-mikedavis (commented):
I don't think we want to add this to the picker specifically. A command to yank the diagnostic under the cursor would be fine I think, as https://github.com/helix-editor/helix/issues/9527 describes, and it would be roughly equivalent to this: just use the picker (or `]d`/`[d` etc) to jump to a diagnostic and then have a command to yank it. Alternatively we could add a special register for the diagnostic(s) under the cursor and then it would be covered by https://github.com/helix-editor/helix/issues/9552.

But as we move away from the menu::Item trait in picker and towards a table layout yanking will make less sense. We could concatenate all of the columns for the selected row but that seems hacky to me.

PR #3789, opened 2022-09-11
update C queries and grammar

kirawi:
The updated queries were pulled from nvim-treesitter. I might make another PR to make it more accurate in the future.


runtime/queries/c/highlights.scm
```diff
@@ -1,72 +1,148 @@
...
+  "struct"
+  "typedef"
+  "union"
+  "volatile"
+  "goto"
+  "register"
+] @keyword
+
+"sizeof" @keyword.operator
+"return" @keyword.return
```
archseer, line 19: @keyword.control.return

runtime/queries/c/highlights.scm
```diff
@@ -1,72 +1,148 @@
...
+(comma_expression [ "," ] @operator)
+
+[
+  (true)
+  (false)
+] @constant.builtin.boolean
+
+[
+  (string_literal)
+  (system_lib_string)
```
archseer, line 108: Second one is `@string.special.path`

runtime/queries/c/highlights.scm
```diff
@@ -76,17 +152,31 @@
...
+  "_unaligned"
+  "__unaligned"
+  "__declspec"
+  (attribute_declaration)
+] @attribute
 
-(comment) @comment
+(field_identifier) @variable.other.member
+
+(ERROR) @error
```
archseer, line 182: I've seen this added more and more, I've had this added to highlights automatically before: https://github.com/helix-editor/helix/blob/c2e41082e4b2e0e1dba6a980a7f09ccaad88cbd0/helix-core/src/syntax.rs#L359-L360

I think we should remove this everywhere, use the snippet above and make it a config option.

kirawi: Will it cause the query to fail if an `ERROR` node isn't defined for the grammar?

archseer: The `ERROR` node is a meta node automatically inserted by tree-sitter on a parsing error.

book/src/languages.md
```diff
@@ -61,6 +61,7 @@ These configuration keys are available:
 | `config`              | Language Server configuration                                 |
 | `grammar`             | The tree-sitter grammar to use (defaults to the value of `name`) |
 | `formatter`           | The formatter for the language, it will take precedence over the lsp when defined. The formatter must be able to take the original file as input from stdin and write the formatted file to stdout |
+| `append-error-node`   | Whether to append a `(ERROR) @error` capture to the language query. (defaults to `true`) |
```
the-mikedavis, line 64: For the option name, I think `highlight-errors` or something similar would be more intuitive.

I'm also curious whether this should be

* language-specific vs editor-wide
* on/off by default

IMO the error highlighting is pretty noisy when you're typing out a block but I'm not sure how others feel.

kirawi: There doesn't seem to be a straightforward way to use the editor config. I could either pass it a parameter (requires changing a lot of calls), or find a way to reference the editor config within the highlighter config.

the-mikedavis: Hmm yeah good point. Using language config is ok then I think 👍

the-mikedavis:
Now that the C/C++ queries have been updated in https://github.com/helix-editor/helix/pull/4079, this might be closable? The appending error node changes could be split off to a separate PR or this PR could pivot to focusing on that

kirawi:
I should probably make a new PR instead if I do that.

PR #3792, opened 2022-09-11
Clean Up Nord Theme

amayer5125:
- Add markup styles
- Replace custom colors with Nord colors
- Clean up code spacing

### DIFF
--- a/runtime/themes/nord.toml
+++ b/runtime/themes/nord.toml
@@ -1,11 +1,10 @@
 # Author : RayGervais<raygervais@hotmail.ca>
 
-"ui.linenr.selected" = { fg = "nord4" }
+"ui.linenr.selected" = "nord4"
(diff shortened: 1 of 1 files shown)


sudormrfbin:
Could you post a before and after screenshot?

amayer5125:
Sure thing! I used a markdown file as an example because that is where I noticed the colors were wrong in the first place.
helix-markup-nord

I matched the theme as closely as I could to the [vim](https://github.com/arcticicestudio/nord-vim) and [vscode](https://github.com/arcticicestudio/nord-visual-studio-code) variants of the theme since those are provided by the author of Nord Theme. I also changed the color of the status bar to align with the [Nord palette documentation](https://www.nordtheme.com/docs/colors-and-palettes).

archseer (approved):
Thanks for working on this!

Issue #3764, opened 2022-09-09
[Idea] Option to disable mouse in insert mode


jakobrs:
# What
There should be an option to disable the mouse in insert mode.

# Why
Accidentally clicking somewhere while typing is annoying. Additionally, it's not possible to undo only what you did after clicking somewhere else. (unlike in eg. vim)

PR #3798, opened 2022-09-11
[WIP] Add option to disable mouse support in specific modes

jakobrs:
See #3764.

A few things that need to be resolved:
- What should the config option be called?
- It might be better to merge this option with the `mouse` option. Note that the `mouse` option currently cannot be set interactively.
  - Would it be preferable to use a vim-style option format? (a sequence of characters representing which modes mouse support is enabled in)

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1336,7 +1336,17 @@ impl Component for EditorView {
                 EventResult::Consumed(callback)
             }
 
-            Event::Mouse(event) => self.handle_mouse_event(event, &mut cx),
+            Event::Mouse(event) => {
+                let config = cx.editor.config();
+                let mode = cx.editor.mode();
+
+                if config.mouse_in.contains(mode) {
+                    self.handle_mouse_event(event, &mut cx)
+                } else {
+                    EventResult::Ignored(None)
+                }
+            }
+
             Event::FocusGained | Event::FocusLost => EventResult::Ignored(None),
         }
     }
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -118,6 +118,8 @@ pub struct Config {
     pub scroll_lines: isize,
     /// Mouse support. Defaults to true.
     pub mouse: bool,
+    /// Which modes mouse support is enabled in.
+    pub mouse_in: MouseIn,
     /// Shell to use for shell commands. Defaults to ["cmd", "/C"] on Windows and ["sh", "-c"] otherwise.
     pub shell: Vec<String>,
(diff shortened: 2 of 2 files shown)


sudormrfbin:
I suppose the config option can be similar to the already present `editor.cursor-shape` option:

```toml
[editor.mouse]
insert = false
normal = true
select = true
```

With the defaults being all true. Either this has to be a breaking change (`mouse` currently only takes a single bool but now takes a table of bools) or it has to be modified to take both a simple `editor.mouse = true` and the above expanded table format. I'm leaning towards the second option.

jakobrs:
This implementation of `Deserialize` should work:

```rust
impl<'de> Deserialize<'de> for Mouse {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: Deserializer<'de>,
    {
        struct MouseVisitor;

        impl<'de> Visitor<'de> for MouseVisitor {
            type Value = Mouse;

            fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
                formatter.write_str("bool or table")
            }

            fn visit_bool<E>(self, value: bool) -> Result<Mouse, E> {
                Ok(Mouse::from(value))
            }

            fn visit_map<M>(self, mut map: M) -> Result<Mouse, M::Error>
            where
                M: MapAccess<'de>,
            {
                let mut result = Mouse::default();

                while let Some((k, v)) = map.next_entry()? {
                    match k {
                        Mode::Normal => result.normal = v,
                        Mode::Select => result.select = v,
                        Mode::Insert => result.insert = v,
                    }
                }

                Ok(result)
            }
        }

        deserializer.deserialize_any(MouseVisitor)
    }
}
```

The main issue is that I can't serialise it as a bool and a struct at the same time, so autocomplete will only list it as a bool or a struct but not both.

the-mikedavis:
The `editor.whitespace.render` option has a similar structure - you can use an enum like the one it uses to avoid making a Visitor: https://github.com/helix-editor/helix/blob/03612174ee4cef23217b5adf415ced4a851b4a44/helix-view/src/editor.rs#L453-L463

jakobrs:
I think it's preferable to "delimit" the effects of how the struct is deserialised: the struct shouldn't have multiple logically equivalent values just because it makes deserialisation a bit easier. Additionally it makes it much harder to change it interactively - `set mouse true` followed by `set mouse.insert false` will fail becaus `Mouse::Basic` has no field `insert`.

pascalkuthe:
closing this one as stale. I also don't see us goint forward with this as its a pretty niche feature and mouse support is not our primary priority. Thank you for contributing

PR #3799, opened 2022-09-11
Move default comment toggle binding to pound

EpocSquadron:
This is a common rebind in both the helix and kakoune community.
Some reasons this is desirable:

* Pound is a common symbol for commenting in many languages
* Pound is not bound to anything else, and it is not obvious what else
  it would be bound to
* `C-c` should really be reserved for cancelling subshells from `:sh`,
  `|` and friends

Most sensible alternatives are taken:

* `/` is search
* `*` adds selection to search register
* `-` is not common, and though not currently taken, may make sense
  as the opposite of `_`, for another PR to implement.
* `c` change selection
* `A-c` change selection w/o yank
* `C` copy selection down
* `A-C` copy selection up

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -192,8 +192,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
             },
         },
 
-        // move under <space>c
-        "C-c" => toggle_comments,
+        "#" => toggle_comments,
 
         // z family for save/restore/combine from/to sels from register
 


the-mikedavis:
C-c does work for closing completion menus / signature help so I think it should work to have it cancel long-running subshell commands if there is one running.

I'm split on this. I think it would be really intuitive for languages with `#` comments but less so for languages with other tokens. C-c as in <i>c</i>omment feels easy to remember across languages

EpocSquadron:
What's the appropriate way to move this forward?

archseer:
I've been testing this out but I'd like to see more opinions. `#` is a nicer key but I find it harder to type than `ctrl-c` :/

David-Else:
I think it is a far better choice to use `#` than `ctrl-c`, and here is why:

- `ctrl-c` already has a well established meaning, and that is to to interrupt a process, not make a comment. I assume using it  for comments  is some abominable left over from VS Code and Microsoft. 
- `ctrl-c` is already used inside Helix to cancel things, I use it all the time for things like closing the file picker. This links nicely mentally to 'interrupt a process' in the terminal.
- I also find `#` much easier to type, but maybe that is my keyboard.
- This PR has 11 thumbs up and no thumbs down showing the idea is popular

dimas-cyriaco:
Why not `ctrl-/` like most editors? It does not seems to be bound to anything.

eulerdisk:
I think `#` is better than `C-c`, especially for non-so-common actions like commenting.
Given that, how about `<Space>c` ? I don't think is bound.
In general I much prefer a home-row friendly approach, like `gl` and `gh` instead of `0` and `$`.

pickfire:
I think this is a good choice since `#` is used as comment for a lot of stuff ranging from programming languages to configuration, as compared to `/**` or `//` or `;`.

For context:

- vscode uses `C-/`, which we can use too
- kate + vim keybinding uses `C-/`
- online editor like play.rust-lang.org uses `C-/`
- vim-commentary uses `gc<motion>`, have more power to know what to comment but translated to helix, that is `gc` which is taken for goto window center
- kakoune uses `#`, which is proposed here
- doom emacs uses `C-x` `C-j`, this is the worse since it conflicts with IME key and I have to type `C-j` twice

So I guess for non-modal editors or vim-based keybindings (`#` was taken for backward search), `C-/` is quite popular, but for modal editors, `#` is a good choice too.

Given that this is type of comment is also quite widely used, I created a pull request to set it as default instead of `//` in https://github.com/helix-editor/helix/pull/6470.

But for helix and kakoune case, `#` is not suitable for any other keys, so might as well just use it.

Once the docs are updated, I guess this can be updated.

hilpara:
C-/ is not nice for keyboards with umlaut-keys -> have to press 3 keys (ctrl+shift+7)

PR #3807, opened 2022-09-11
Add command line parameter to specify log file

hunger:
I had the logs of my debug helix mixed in with the logs from the production helix.

Add a `--logfile` command line argument to redirect any logs to other files, making my debugging easier :-)


the-mikedavis:
We try to minimize the number of CLI flags so this should probably be controlled in config (see https://github.com/helix-editor/helix/pull/2666#issue-1260385517). A config option for the log file to use would also help https://github.com/helix-editor/helix/pull/3783 since you could set it to `/dev/null`.

hunger:
I understand that, but this makes more sense as a command line switch to me.

My use-case is that I want to spin up a `hx` instance with lots of logging and see that separate from the normal `hx` instance I edit in. `hx -vvvvv --logfile /tmp/hx.log` is _much_ nicer and more discoverable than editing a `config.toml` file somewhere and doing `hx -vvvv --configfile /path/to/config.toml`.

helix-term/src/args.rs
```diff
@@ -48,6 +49,10 @@ impl Args {
                     Some(path) => args.config_file = Some(path.into()),
                     None => anyhow::bail!("--config must specify a path to read"),
                 },
+                "--logfile" => match argv.next().as_deref() {
```
the-mikedavis, line 52: I think I would prefer this as `--log` so it lines up with `--config`. What do you think?

the-mikedavis:
Could you update the completion files in `contrib/completion/` as well? Whatever we have for the config file flag should work well for log as well I think

hunger:
I updated the completion files. I do not use any of those and never wrote some before, so no idea how well that will work!

Basically I copied what I thought is the entry for `--config` and edited that to look sensible to the uninitiated.

the-mikedavis (approved):

hunger:
@the-mikedavis Thanks for taking the time to review this. I appreciate your effort!

Anything I need to do now to get the PR merged? Github will not let me hit the button myself :-)

sudormrfbin (approved):
Tested, and LGTM!

hunger:
@sudormrfbin: Do I need to do anything now?

sudormrfbin:
For entirely new features (compared to bug fixes or enhancements) we generally wait for @archseer to do a final review and merge the PR, so nothing needs to be done from your side for now :)

hunger:
Latest version makes the usage message say `--log`, not `--logfile`. Missed that when I updated that before.

PR #3810, opened 2022-09-12
Fix Nix flake devShell on aarch64-darwin

the-mikedavis:
LLDB is marked broken on ~all arches except for x86_64-linux~ aarch64-darwin. So we can fix `nix develop` by removing lldb for it on aarch64-darwin

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -92,8 +92,9 @@
             prev.packages
             ++ (
               with common.pkgs;
-                [lld_13 lldb cargo-flamegraph rust-analyzer]
+                [lld_13 cargo-flamegraph rust-analyzer]
                 ++ (lib.optional (stdenv.isx86_64 && stdenv.isLinux) cargo-tarpaulin)
+                ++ (lib.optional stdenv.isLinux lldb)
             );
           env =
             prev.env


flake.nix
```diff
@@ -92,8 +92,8 @@
             prev.packages
             ++ (
               with common.pkgs;
-                [lld_13 lldb cargo-flamegraph rust-analyzer]
-                ++ (lib.optional (stdenv.isx86_64 && stdenv.isLinux) cargo-tarpaulin)
+                [lld_13 cargo-flamegraph rust-analyzer]
+                ++ (lib.optionals (stdenv.isx86_64 && stdenv.isLinux) [cargo-tarpaulin lldb])
```
groves, line 96: Doh! I actually had the below change locally and forgot to make a PR:

```suggestion
                ++ (lib.optional (stdenv.isx86_64 && stdenv.isLinux) cargo-tarpaulin)
                ++ (lib.optional stdenv.isLinux lldb)
```

I _think_ lldb is only busted on darwin, so we don't need the x86_64 check: https://github.com/NixOS/nixpkgs/blob/master/pkgs/development/compilers/llvm/13/lldb/default.nix#L105

Issue #3808, opened 2022-09-12
No way to move cursor in Insert mode


REALERvolker1:
### Summary

There is no way to move the cursor in Insert mode currently. Helix tutor mentions using the arrow keys, but a quick review of the Github issues reveals this is a common problem.

The Ctrl and Alt navigation does not work for me either. When in insert mode, the Control key has no effect on these keys.

![image](https://user-images.githubusercontent.com/73304952/189574365-e9a281c5-81c5-4d3b-8870-3317fd63778b.png)

![image](https://user-images.githubusercontent.com/73304952/189575611-13383705-d4e4-4aec-8e7c-f711652fd310.png)

### Reproduction Steps

 

I tried this:

*pressed Up arrow key in Insert mode*

I expected this to happen:

* cursor moves up 1 line

Instead, this happened:

*cursor stays in place*

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-09-11T23:10:08.283 helix_view::theme [WARN] Theme: malformed hexcode: highlight
2022-09-11T23:10:08.594 helix_view::theme [WARN] Theme: malformed hexcode: highlight
2022-09-11T23:20:09.397 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:20:34.776 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:22:16.383 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:28:12.746 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:28:44.418 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:29:08.213 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:32:01.375 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:34:11.454 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-09-11T23:37:59.207 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
[...]

PR #3811, opened 2022-09-12
Remove arrow key recommendation from tutor

the-mikedavis:
As a follow-up to https://github.com/helix-editor/helix/pull/3671

The tutor mentioned that arrow keys can be used in insert mode which is no longer the case

See #3808


Omnikar:
You need to replace the deleted lines with blank ones to maintain the spacing between sections.

PR #3819, opened 2022-09-12
Add option to skip the first indent guide

robinvd:
coming from my previous setup, the first indent guide are a bit distracting. Therefore the option to disable this

current:

image

enabled:
image

my neovim config for comparison (using indent-blankline with `show_first_indent_level = false`

image

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -222,7 +222,7 @@ Options for rendering vertical indent guides.
 | ---           | ---                                                     | ---     |
 | `render`      | Whether to render indent guides.                        | `false` |
 | `character`   | Literal character to use for rendering the indent guide | `│`     |
-| `skip-levels` | amount of indent levels to skip                         | `0`     |
+| `skip-levels` | Number of indent levels to skip                         | `0`     |
 
 Example:
 
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -218,10 +218,11 @@ tabpad = "·" # Tabs will look like "→···" (depending on tab width)
 
 Options for rendering vertical indent guides.
 
-| Key         | Description                                             | Default |
-| ---         | ---                                                     | ---     |
-| `render`    | Whether to render indent guides.                        | `false` |
-| `character` | Literal character to use for rendering the indent guide | `│`     |
+| Key           | Description                                             | Default |
+| ---           | ---                                                     | ---     |
+| `render`      | Whether to render indent guides.                        | `false` |
+| `character`   | Literal character to use for rendering the indent guide | `│`     |
+| `skip-levels` | amount of indent levels to skip                         | `0`     |
 
 Example:
 
@@ -229,4 +230,5 @@ Example:
 [editor.indent-guides]
 render = true
 character = "╎"
+skip-levels = 1
 ```
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -436,8 +436,8 @@ impl EditorView {
                 return;
             }
 
-            let mut starting_indent = (offset.col / tab_width) as u16;
-            starting_indent += config.indent_guides.skip_levels;
+            let starting_indent =
+                (offset.col / tab_width) as u16 + config.indent_guides.skip_levels;
             // TODO: limit to a max indent level too. It doesn't cause visual artifacts but it would avoid some
             // extra loops if the code is deeply nested.
 
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -417,6 +417,7 @@ impl EditorView {
             " ".to_string()
         };
         let indent_guide_char = config.indent_guides.character.to_string();
+        let skip_first_indent_guide = config.indent_guides.skip_first;
 
         let text_style = theme.get("ui.text");
         let whitespace_style = theme.get("ui.virtual.whitespace");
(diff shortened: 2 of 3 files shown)


EpocSquadron:
Could we have a description or even better screenshots of what this does? Is there a related issue to link?

helix-view/src/editor.rs
```diff
@@ -529,15 +529,17 @@ impl Default for WhitespaceCharacters {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(default)]
+#[serde(default, rename_all = "kebab-case")]
 pub struct IndentGuidesConfig {
+    pub skip_first: bool,
     pub render: bool,
```
AlexanderBrevig, line 535: Very nit-picky but I would like `pub render: bool` to be at the top of both the struct definition and in Default impl. Seems like the other structs typically starts with the enable/render flag so it's more consistent that way.

archseer:
Is this configurable in other editors? Why not make it an `usize` and skip that many levels: `starting_indent += skip`

helix-view/src/editor.rs
robinvd: makes sense! indeed more consistent

robinvd:
> Could we have a description or even better screenshots of what this does? Is there a related issue to link?

ah my bad! updated the discription

robinvd:
> Is this configurable in other editors? Why not make it an `usize` and skip that many levels: `starting_indent += skip`

@archseer converted it to a number

EpocSquadron:
I quite like this. Can't help but think this would be an uncontroversial default.

sudormrfbin:
> in gui editors i find the indent guide to be much more subtle so the first line isnt as distracting to me.

The color can be configured to be something less loud using the `ui.virtual.indent-guide` theme key btw, it does look quite distracting in your screenshot.

The config option will have to be updated in the book too.

helix-view/src/editor.rs
```diff
@@ -529,15 +529,17 @@ impl Default for WhitespaceCharacters {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
-#[serde(default)]
+#[serde(default, rename_all = "kebab-case")]
 pub struct IndentGuidesConfig {
     pub render: bool,
     pub character: char,
+    pub skip: u16,
```
sudormrfbin, line 536: Since it takes the number of levels to skip now, something like `skip_levels` might be more suited, or maybe `render_level`.

robinvd: renamed to skip_levels

robinvd:
> The color can be configured to be something less loud using the ui.virtual.indent-guide theme key btw, it does look quite distracting in your screenshot.

good to know!

> The config option will have to be updated in the book too.

done

book/src/configuration.md
```diff
@@ -218,15 +218,17 @@ tabpad = "·" # Tabs will look like "→···" (depending on tab width)
...
 
-| Key         | Description                                             | Default |
-| ---         | ---                                                     | ---     |
-| `render`    | Whether to render indent guides.                        | `false` |
-| `character` | Literal character to use for rendering the indent guide | `│`     |
+| Key           | Description                                             | Default |
+| ---           | ---                                                     | ---     |
+| `render`      | Whether to render indent guides.                        | `false` |
+| `character`   | Literal character to use for rendering the indent guide | `│`     |
+| `skip-levels` | amount of indent levels to skip                         | `0`     |
```
A-Walrus, line 225:
```suggestion
| `skip-levels` | Number of indent levels to skip                         | `0`     |
```

A-Walrus (changes requested):

helix-term/src/ui/editor.rs
```diff
@@ -436,7 +436,8 @@ impl EditorView {
                 return;
             }
 
-            let starting_indent = (offset.col / tab_width) as u16;
+            let mut starting_indent = (offset.col / tab_width) as u16;
+            starting_indent += config.indent_guides.skip_levels;
```
the-mikedavis, line 440:
```suggestion
            let starting_indent =
                (offset.col / tab_width) as u16 + config.indent_guides.skip_levels;
```

archseer:
Rather than making this configurable, should we just always skip the first level? Not sure what the default behavior is elsewhere

Ordoviz:
I find indent guides unusable without this, but it probably should be configurable and off by default because a new Helix user wouldn't expect the first indent level to be different. Adding `skip-levels = 1` to the example in the documentation should already ensure that users who want to try out indent guides don't miss out on this new feature.

The existence of these three Vim issues supports my conclusion: [issue 1](https://github.com/Yggdroot/indentLine/issues/21), [issue 2](https://github.com/Yggdroot/indentLine/issues/356), [issue 3](https://github.com/lukas-reineke/indent-blankline.nvim/issues/12).

EpocSquadron:
Played with this just now. I feel like it works particularly well without spaces/tabs being rendered, but feels a little empty with them present. So I can see and agree that leaving it on by default is probably better.

CptPotato:
> > in gui editors i find the indent guide to be much more subtle so the first line isnt as distracting to me.
>
> The color can be configured to be something less loud using the `ui.virtual.indent-guide` theme key btw, it does look quite distracting in your screenshot.

Side note: I have kept the above theme (everforest) up to date in my own theme repository which includes the style for indent guides among other changes. I'll make a PR to update it later this week.

robinvd:
@archseer 

I would say the pr is done, except a potential change of defaults.

PR #3824, opened 2022-09-12
added case switching under match mode

zim0369:
These bindings might be more convenient. Discussing here: #3818


zim0369:
Closing

Issue #3816, opened 2022-09-12
Syntax Highlighting related panic: IncludedRangesError(1)


sevenautumns:
### Summary

When adding a `'`-symbol to a specific line of a specific nix-file, the predefined language-server for nix files (rnix-lsp) seems to provide information which results in an error in the tree_sitter parser which is unwrapped and hence panics.  
https://github.com/helix-editor/helix/blob/03612174ee4cef23217b5adf415ced4a851b4a44/helix-core/src/syntax.rs#L988

### Reproduction Steps

 

I tried this:

1. `hx flake.nix`
2. go to line 32
3. add `'` after the `=`-symbol of that line
4. crash

<details><summary>flake.nix</summary>

```nix
{
  inputs = {
    utils.url = "github:numtide/flake-utils";
    devshell.url = "github:numtide/devshell";
    fenix.url = "github:nix-community/fenix";
    fenix.inputs.nixpkgs.follows = "nixpkgs";
  };

  outputs = { self, nixpkgs, utils, devshell, fenix, ... }@inputs:
    utils.lib.eachSystem [ "aarch64-linux" "i686-linux" "x86_64-linux" ]
    (system:
      let
        pkgs = import nixpkgs {
          inherit system;
          overlays = [ devshell.overlay ];
        };
        rust-toolchain = with fenix.packages.${system};
          combine [ stable.rustc stable.cargo stable.clippy stable.rustfmt ];
      in rec {
        devShells.default = (pkgs.devshell.mkShell {
          imports = [
            # "${devshell}/extra/language/rust.nix"
            "${devshell}/extra/git/hooks.nix"
          ];
          name = "apex-rs-vanilla-dev-shell";
          packages = with pkgs; [
            rust-toolchain
            rust-analyzer
            cargo-outdated
            cargo-udeps
          ];
          devshell.startup.bindgen-hook.text = "source ${pkgs.rustPlatform.bindgenHook}/nix-support/setup-hook && populateBindgenEnv";
          git.hooks = {
            enable = true;
            pre-commit.text = "nix flake check";
          };
          commands = [
            {
              package = "git-cliff";
              help = "Changelog generator";
            }
            {
              package = "treefmt";
              help = "Format project tree with recommended formatter";
              category = "formatter";
            }
            {
              name = "udeps";
              command = "cargo-udeps udeps";
              category = "formatter";
              help = "Find unused dependencies in Cargo.toml";
            }
            {
              name = "outdated";
[...]

PR #3826, opened 2022-09-12
fix: map_err()? instead of unwrap

AlexanderBrevig:
Remove an unwrap identified in #3816 that caused a crash in syntax.rs

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -985,7 +985,9 @@ impl LanguageLayer {
     }
 
     fn parse(&mut self, parser: &mut Parser, source: &Rope) -> Result<(), Error> {
-        parser.set_included_ranges(&self.ranges).unwrap();
+        parser
+            .set_included_ranges(&self.ranges)
+            .map_err(|_| Error::InvalidRanges)?;
 
         parser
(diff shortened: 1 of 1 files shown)


AlexanderBrevig:
Closes #3816

archseer:
I think this is a good change, but it doesn't close #3816:

- It doesn't fully fix the panic since we still unwrap `syntax.update()` everywhere (caller of this function). I'm not sure how to gracefully handle syntax parsing, perhaps we just use the old tree/render without highlights.

- Doesn't fix the source of the issue

AlexanderBrevig:
> I think this is a good change, but it doesn't close #3816:
> 
> * It doesn't fully fix the panic since we still unwrap `syntax.update()` everywhere (caller of this function). I'm not sure how to gracefully handle syntax parsing, perhaps we just use the old tree/render without highlights.
> * Doesn't fix the source of the issue

Oh, didn't think to check up the call-stack and I have no way of testing as I cannot reproduce.

Do you want me to take a pass at handling errors there, and try to actually close the issue?

archseer:
I think it's good to merge, but I'll keep the parent issue open :)

PR #3839, opened 2022-09-13
update `x_offset` calculation in `Buffer::set_string_truncated`

kristopherbullinger:
Hello. First time contributor here.

SUMMARY:
With these changes, when `truncate_start` is `true` in a call to `Buffer::set_string_truncated`, the `x_offset` is now properly updated according to the (potentially truncated) content width.

Previously, the `x_offset` was never updated in this branch of code, which caused a layout error when rendering a Picker with items whose calculated labels contained multiple Spans. Since `x_offset` was not updated between calls to `set_string_truncated`, each span's content would render starting from the leftmost edge of the item's line, causing the content items to overlap each other. For example, in a menu item with two spans with content `some_directory/` and `14.5 KB`, it would render as `14.5 KBrectory/`.

Also, I took the liberty of renaming a variable for added clarity.

### DIFF
--- a/helix-tui/src/buffer.rs
+++ b/helix-tui/src/buffer.rs
@@ -343,14 +343,14 @@ impl Buffer {
             let mut start_index = self.index_of(x, y);
             let mut index = self.index_of(max_offset as u16, y);
 
-            let total_width = string.width();
-            let truncated = total_width > width;
+            let content_width = string.width();
+            let truncated = content_width > width;
             if ellipsis && truncated {
                 self.content[start_index].set_symbol("…");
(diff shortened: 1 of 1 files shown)


AlexanderBrevig (commented):
Seems like a good thing to add a test case for. 
I like the name change, it matches:
https://github.com/helix-editor/helix/blob/b5be72bff79075ef4b5b51922da15e003c274b9b/helix-term/src/ui/text.rs#L58

archseer (approved):
Sorry for the delay on this! Looks good to me 👍🏻

PR #3851, opened 2022-09-15
feat: add Fennel language

d12bb:
Adds support for [Fennel](https://fennel-lang.org). Queries were converted from [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/fennel/highlights.scm), should I add a comment noting that?

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -124,6 +124,21 @@ name = "fish"
 source = { git = "https://github.com/ram02z/tree-sitter-fish", rev = "84436cf24c2b3176bfbb220922a0fdbd0141e406" }
 
 [[language]]
+name = "fennel"
+scope = "source.fnl"
+injection-regex = "(fennel|fnl)"
+file-types = ["fnl"]
+shebangs = ["fennel"]
+comment-token = ";"
+formatter = { command = "fnlfmt", args = ["-"] }
+indent = { tab-width = 2, unit = "  " }
+roots = []
+
+[[grammar]]
+name = "fennel"
+source = { git = "https://github.com/travonted/tree-sitter-fennel", rev = "517195970428aacca60891b050aa53eabf4ba78d"}
+
+[[language]]
 name = "mint"
 scope = "source.mint"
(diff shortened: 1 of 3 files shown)


kirawi:
You need to run `cargo xtask docgen` and commit the changes.

archseer:
Do we really need another grammar? We have `tree-sitter-scheme` already, you could reuse it with some custom highlighting queries.

the-mikedavis:
Tree-sitter-scheme is also oddly large (1.5MB), I think because of all of the regexes in the grammar definition.

We might be able to support all lisps with one tree-sitter-sexp grammar?

d12bb:
Should have thought of that earlier, I only tried to reuse the queries before. Just tried using the Scheme grammar and copying queries dir: looks great at first glance, one would just have to make sure all builtins are queried correctly I guess. I'll have a deeper look tomorrow.

As for unifying lisps, there is [AbstractMachinesLab/tree-sitter-sexp](https://github.com/AbstractMachinesLab/tree-sitter-sexp), last commit was two years ago though.

icefoxen:
I have no actual dog in this fight but in passing I'll note that there can be some fairly disparate syntax between Lisp's.  CL, Clojure, Fennel and Racket all have pretty different keywords and opinions about things like character literals and compile directives.

Jummit:
So I've been using this patch for a while now, and I think the conclusion of the Lisp-syntax discussion was that Fennel should get it's own one since there are too many discrepancies between the Lisp dialects.

Since this is marked as inactive, should I create a new pr where feedback can be addressed to get this merged?

p00f:
@archseer using scheme grammar won't work, fennel uses square and curly brackets

archseer:
That should be handled by the scheme grammar, other brackets are just syntax sugar

pascalkuthe:
clsoing this one as stale. I don't think we will add an extra grammar here so any language definition should be reusing the scheme grammar. Thank you for contributing.

Issue #3820, opened 2022-09-12
Auto-indentation for JSX is incorrect


DoctorRyner:
### Summary

Going to the next line places your pointer in the wrong place
<img width="932" alt="Bildschirmfoto 2022-09-12 um 19 22 21" src="https://user-images.githubusercontent.com/24430919/189695957-3f12ea77-8c06-4460-b91e-5a7b30a97a51.png">

### Reproduction Steps

1. Open any `{jsx,tsx}` file
2. Insert this:
```tsx
export const App: FunctionComponent = () => {
  return (
    <div>
      <h1>{greeting}</h1>
      <p>{welcomeMsg}</p>
      <div>
        <p>Test</p>
    
      </div>
    </div>
  )
}
```
3. To see that it doesn't leap to the proper spot, try hitting 'o' or pressing enter at the end of a JSX line

### Platform

macOS

### Terminal Emulator

The latest iTerm 2

### Helix Version

22.08.1

the-mikedavis:
Currently JSX uses the JavaScript indentation queries: https://github.com/helix-editor/helix/blob/master/runtime/queries/jsx/indents.scm

You can try tuning these so that the JSX tags trigger indents (see https://docs.helix-editor.com/master/guides/indent.html)

mbenja:
@DoctorRyner I can take a look at this and report back within a day or so

DoctorRyner:
Can confirm that it works. Thanks!

PR #3853, opened 2022-09-15
fix: Improving JSX and TSX indents

mbenja:
Adding JSX tag indents for JSX and TSX files, addressing issue #3820


archseer (approved):
Thanks for working on this!

PR #3872, opened 2022-09-16
Change focus to modified docs on quit

### LINKED ISSUE #3857: Switch to unsaved buffers on quit

A-Walrus:
Resolves #3857
When quitting with modified documents, automatically switch focus to one of them.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2,7 +2,7 @@ use std::ops::Deref;
 
 use super::*;
 
-use helix_view::editor::{Action, ConfigEvent};
+use helix_view::editor::{Action, CloseError, ConfigEvent};
 use ui::completers::{self, Completer};
 
 #[derive(Clone)]
@@ -71,8 +71,29 @@ fn buffer_close_by_ids_impl(
     doc_ids: &[DocumentId],
     force: bool,
 ) -> anyhow::Result<()> {
-    for &doc_id in doc_ids {
-        editor.close_document(doc_id, force)?;
+    let (modified_ids, modified_names): (Vec<_>, Vec<_>) = doc_ids
+        .iter()
+        .filter_map(|&doc_id| {
+            if let Err(CloseError::BufferModified(name)) = editor.close_document(doc_id, force) {
+                Some((doc_id, name))
+            } else {
+                None
+            }
+        })
+        .unzip();
+
+    if let Some(first) = modified_ids.first() {
+        let current = doc!(editor);
+        // If the current document is unmodified, and there are modified
+        // documents, switch focus to the first modified doc.
+        if !modified_ids.contains(&current.id()) {
+            editor.switch(*first, Action::Replace);
+        }
+        bail!(
+            "{} unsaved buffer(s) remaining: {:?}",
+            modified_names.len(),
+            modified_names
+        );
     }
 
     Ok(())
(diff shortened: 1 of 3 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -513,23 +513,33 @@ fn force_write_quit(
     force_quit(cx, &[], event)
 }
 
-/// Results an error if there are modified buffers remaining and sets editor error,
-/// otherwise returns `Ok(())`
+/// Results an error if there are modified buffers remaining and sets editor
```
poliorcetics, line 516:
```suggestion
/// Results in an error if there are modified buffers remaining and sets editor
```

poliorcetics (commented):
LGTM apart from a nit :)

EpocSquadron:
Does this work with buffer-close-all and the like as well? Seems like it might.

EpocSquadron:
Tested this out with various combinations of all of the buffer close options, cquit, etc, and with one and two buffers, and everything seems to work swimingly! Great work!

helix-view/src/editor.rs
```diff
@@ -686,6 +686,14 @@ pub enum Action {
     VerticalSplit,
 }
 
+/// Error thrown on failed document closed
+pub enum CloseError {
+    /// Document doesn't exist
+    DoesntExist,
```
dead10ck, line 692: Minor nit: could we name this `DoesNotExist` instead?

helix-view/src/document.rs
```diff
@@ -1036,6 +1036,12 @@ impl Document {
             .map(helix_core::path::get_relative_path)
     }
 
+    pub fn display_name(&self) -> String {
```
dead10ck, line 1039: Glad to see some repetition get factored out 👍

helix-term/src/commands/typed.rs
```diff
@@ -513,23 +534,26 @@ fn force_write_quit(
     force_quit(cx, &[], event)
 }
 
-/// Results an error if there are modified buffers remaining and sets editor error,
-/// otherwise returns `Ok(())`
+/// Results in an error if there are modified buffers remaining and sets editor
+/// error, otherwise returns `Ok(())`. If the current document is unmodified,
+/// and there are modified documents, switches focus to one of them.
 pub(super) fn buffers_remaining_impl(editor: &mut Editor) -> anyhow::Result<()> {
```
dead10ck, line 540: I think we might be able to share some code here if we change this function to take a `&[DocumentId]`, rather than just iterating through all of the editor's documents. Then at the call sites of `buffers_remaining_impl`, we could pass in the editor documents. Then, in `buffer_close_by_ids_impl`, we could simply call `buffers_remaining_impl` and pass the IDs right through instead of repeating the iterator chain. That might also render the new error type unnecessary.

dead10ck (changes requested):
If you have the time, an integration test would be great! Though I wouldn't consider that a blocker myself.

Thanks for your contribution!

helix-view/src/editor.rs
archseer: Or just NotExist

helix-view/src/document.rs
```diff
@@ -1036,6 +1036,12 @@ impl Document {
...
+        self.relative_path()
+            .map(|path| path.to_string_lossy().to_string())
```
archseer, line 1041: Rather than always allocating could this just return a Cow<>? Then the parent code can call to_owned() if it needs to

helix-term/src/commands/typed.rs
A-Walrus: The difference is that `buffers_remaining_impl` doesn't actually close anything, it just checks if everything is saved, and if not shows one of the unsaved documents. That means that it couldn't be used by `buffer_close_by_ids_impl`, which needs to actually close buffers.

dead10ck: Right now `buffer_close_by_ids_impl` attempts to close each buffer and fails if it finds one that is modified. What I was thinking was instead it could just call `buffers_remaining_impl` first, which would do all the checking first, and fail if there are any modified buffers. If it doesn't fail, then move on to close the buffers, knowing they are all unmodified.

Though thinking about some more, this could leave room for a race condition if the buffer somehow became modified in between the check and the close.

No big deal either way, it's good how it is now, in my opinion.

PR #2211, opened 2022-04-21
feat: show filetype and filetype icon and lsp name in statusline

ttys3:
I'd like to show filetype and filetype icon and lsp name in statusline

is there better way to do this?

maybe more flexiable way, using a language like starlark ?

image

image

image

PR #3878, opened 2022-09-17
feat: File Icons in File Picker

jcdickinson:
Introduces optional file icons for the `FilePicker`. The icons are accessed through the new `FileInfo` struct, which is a valid `Item` that includes the icon in its `Span` (if configured to do so).

The idea is that `FileInfo` could be used for additional file type metadata if it is needed in the future.

These icons assume the use of a nerd font, which is why they are disabled by default.

image

### DIFF
--- a/helix-term/src/file_info.rs
+++ b/helix-term/src/file_info.rs
@@ -1,0 +1,390 @@
+use std::path::PathBuf;
+
+use tui::text::{Span, Spans};
+
+use crate::ui::menu::Item;
+
+pub struct FileInfoData {
+    pub root_path: PathBuf,
+    pub show_icons: bool,
+}
+
+pub struct FileInfo {
+    path: PathBuf,
+    icon_character: char,
+}
+
+impl FileInfo {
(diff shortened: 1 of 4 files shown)


helix-term/src/file_info.rs
```diff
@@ -0,0 +1,390 @@
...
+    }
+}
+
+impl Item for FileInfo {
+    /// Root prefix to strip.
+    type Data = FileInfoData;
+
+    fn label(&self, data: &Self::Data) -> Spans {
+        let mut result = vec![Span::raw(format!(
+            "  {}",
```
jcdickinson, line 40: If this isn't done then the icon overwrites the first character during render, and I'm not sure why. Any ideas?

the-mikedavis:
There is prior open work on this: https://github.com/helix-editor/helix/pull/2869. Also see the discussion in https://github.com/helix-editor/helix/pull/2211

helix-term/src/file_info.rs
A-Walrus: I believe that's a bug that should be solved by #3839

PR #3881, opened 2022-09-17
Add bufferline colors to 15 themes

zim0369:
Fixes visual discrepancy with a few themes by adding bufferline colors. There's one issue in a few commits where I've added `fg` to `ui.bufferline.background` where it was unnecessary. I only realized later. It does not break anything though.

### DIFF
--- a/runtime/themes/acme.toml
+++ b/runtime/themes/acme.toml
@@ -20,6 +20,9 @@
 "diagnostic.error" = {bg="white", modifiers=["bold"]}
 "diagnostic.warning" = {bg="white", modifiers=["bold"]}
 "diagnostic.hint" = {bg="white", modifiers=["bold"]}
+"ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
+"ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
+"ui.bufferline.background" = { fg = "indent", bg = "acme_bar_bg" }
 
 [palette]
 white = "#ffffff"
--- a/runtime/themes/acme.toml
+++ b/runtime/themes/acme.toml
@@ -22,7 +22,6 @@
 "diagnostic.hint" = {bg="white", modifiers=["bold"]}
 "ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
 "ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
-"ui.bufferline.background" = { fg = "indent", bg = "acme_bar_bg" }
 
 [palette]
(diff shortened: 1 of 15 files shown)


zim0369:
There's a lot of themes btw. Helix should stop adding themes at some point. And some themes like acme, base16_light, etc. which are poorly built should be removed.

zim0369:
Then at least adding a theme picker is necessary.

archseer:
`:theme` already offers a live preview.

David-Else:
How about merging https://github.com/helix-editor/helix/pull/3234 sooner rather than later and hopefully remove the endless theme fix pull requests as a result?

archseer:
@David-Else some of these fixes are adding new styles for features that didn't exist when  the theme was initially written. We don't block merging new features on adding missing styles to all themes since this is dependent on theme authors.

runtime/themes/acme.toml
```diff
@@ -20,6 +20,9 @@
 "diagnostic.error" = {bg="white", modifiers=["bold"]}
 "diagnostic.warning" = {bg="white", modifiers=["bold"]}
 "diagnostic.hint" = {bg="white", modifiers=["bold"]}
+"ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
+"ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
+"ui.bufferline.background" = { fg = "indent", bg = "acme_bar_bg" }
```
the-mikedavis, line 25: Themes have a fallback behavior: if the theme doesn't define `ui.bufferline.background`, `ui.bufferline` is used instead. So all of these that use the same colors for `ui.bufferline` and `ui.bufferline.background` can be simplified by removing the `ui.bufferline.background` line

zim0369 (commented):
Exactly duplicates(`ui.bufferline.background`) of `ui.bufferline` removed. Unnecessary foreground property of `ui.bufferline.background` removed.

zim0369 (commented):
Improvements to catppuccin_frape and catppuccin_latte

David-Else:
> > How about merging #3234 sooner rather than later and hopefully remove the endless theme fix pull requests as a result?
> 
> The linter only finds the mistakes in themes right? So, they still have to be fixed manually?

Yes, pre-linter they currently need to be fixed manually.

The linter is merged now, people just need to make sure to use it as it is not yet in CI, see bottom of https://github.com/helix-editor/helix/blob/master/book/src/themes.md

the-mikedavis:
Thanks!

PR #3888, opened 2022-09-18
docs: Punctuation on tutor

adrian5:
Correctness aside, _“the line below marked”_ reads oddly, and makes one assume _“below --->”_ for a second.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -70,7 +70,7 @@ _________________________________________________________________
 
  Type the d key to delete the character under the cursor.
 
- 1. Move the cursor to the line below marked -->.
+ 1. Move the cursor to the line marked '-->' below.
  2. Move the cursor to each extra character, and type d to
     delete it.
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I'm not sure this is more correct and the current text reads better to me. Can you explain?

My grammar is a bit rusty but the way I see it, the sentence breaks down like so

```
Move the cursor to the line below [which/that is] marked '-->'
|--|     |----| |---------------| |--------------------------|
Verb      Noun     Preposition      Defining Relative Clause
```

And I think it's a common blunder to use a comma to connect a _defining_ relative clause - though it would be correct for a non-defining relative clause. (And it's defining because you need the information that the line is marked with the "-->" to know which line the sentence is referring to.)

David-Else:
@the-mikedavis You are correct, your grammar is good, there is no need for a defining relative clause or comma.

adrian5:
> I think it sounds correct

Well, in speech one would make a pause after _below_, to differentiate:

```
Move the cursor to the line below [which is] marked '-->'
Move the cursor to the line below [the] marked items
```

I'm not a native speaker, but to me that reads better with a comma. Feel free to close if you disagree, I may be wrong.

the-mikedavis:
It might be clearer to use "Move the cursor to the line marked '-->' below". What do you think?

adrian5:
That's clearer in my view. The `marked '-->'` is what's important (to look for), and `below` is only added to ensure the user goes looking in the right direction.

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #3852, opened 2022-09-15
Separate 'brace action' and 'match selection' functionality


asvln:
The idea of 'selection first' makes a lot of sense to me compared to vim's 'action first'.
However, I do not know if the keymap currently adheres to that philosophy the best that it could.

I realized I was using `miw` so much that I am now capable of just mashing the keyboard with my hands to hit the key sequence.
But... I feel really stupid doing this.
It makes selecting anything stressful and is very unintuitive, especially when selecting the word under the cursor is my most common selection tactic.

I then realized that the current `m` mapping really does two separate things...

1. Working with matched braces
2. Selecting textobjects

In my opinion, these functionalities should **NOT** be on the same key/related in keymapping at all.

When (1): After pressing `m`, I am then making an action.
This action has a variable number of additional keypresses to appropriately complete the action.
Additionally, each action operates differently in the context of what is selected.

This api is **complex** and at the end, my current selection is **NOT changed** at all.

When (2): After pressing `m`, I am then choosing 'inner' or 'around'.
Then I am just selecting the textobject I want with my next keypress (the selection always only requires 2 keypresses after pressing `m`).

This api is **very straight-forward** and at the end, my selection **IS changed**.

If I have an editor that is based on 'selection first', I need to be able to select a textobject within 2 keypresses, otherwise I find it inefficient and repetitive, and that takes away the fun of it all.

### Hypothetical Solution

- Remove all brace actions from `m` (lets say we nest it under the `space` mapping to get it out of the way)
- `mi` = `m`
- `ma` = `M`

This solves the issue of 3 keypresses for a very commonly used approach to selection and drastically simplifies how (at least I) understand the editor as a whole.

If this was how `m` worked when I first started using helix, I would be way more proficient at working with selections and multiple cursors than I am now.

Having direct `m` access to select a textobject makes the whole approach seem as simple, quick, and intuitive as any vim 'action first' motion.

CptPotato:
I don't have a strong opinion about the proposal, but I do use `miw` very frequently as well.

I'd be interested in trying this, is it possible to get this behavior just by remapping keys in `config.toml`?

Omnikar:
> I'd be interested in trying this, is it possible to get this behavior just by remapping keys in `config.toml`?

Yep
```toml
"m" = "select_textobject_inner"
"M" = "select_textobject_around"
```

sdemura:
[...]

PR #3893, opened 2022-09-18
Separate 'brace action' and 'match selection' in keymap

asvln:
This changes the behavior of `m` to only match selections.
`m` = `mi`
`M` = `ma`
Brace actions are relocated to `^`. 

Issue: #3852

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -89,14 +89,16 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "X" => extend_to_line_bounds,
         "A-x" => shrink_to_line_bounds,
 
-        "m" => { "Match"
-            "m" => match_brackets,
+        "m" => select_textobject_inner,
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I like the idea of easier access to `mi`/`ma`.

I don't like relocating brace actions to `^` though. `^` has been proposed and tried for a few Kakoune keymaps (marks, https://github.com/mawww/kakoune/pull/1116) but it has never stuck. It's not very discoverable. (Plus I'd like to use it for marks/selections #3720 :P)

The normal-mode keymap is already very full which is why it's hard to separate key sequences like `m` out like this.

asvln:
I like the idea of `^` being reserved for marks also.

I do think brace functionality deserves it's own key rather than being nested under another key.

dead10ck:
I am also not a fan of `^`. This also breaks the "in" / "around" mnemonic. I don't think this is worth saving a single keystroke.

asvln:
moved to #4655

PR #3890, opened 2022-09-18
(Git) diff gutter implementation

pascalkuthe:
Supercedes #1623.

I took this PR up as the original PR was tagged as waiting-for-author/inactive.

This PR is rebased on the current master to avoid conflicts.
Additionally, I have updated the PR with some improvements.
The design is now much more similar to the design suggested by @p-e-w.

The original author already did most of the work by implementing incremental diffing.
However, I have now streamlined the git interface.
The incremental diff only requires that git provides the head state of a file while the rest is handled by helix.
Therefore, I have simplified the git interface down to one function: `get_file_head`.
This interface is now provided by a trait, that is implemented for git but is trivial to implement for other diff sources as well and could be exposed to the plugin system in the future.

The original PR cached the git repositories in memory.
This has been removed because it did not handle nested repositories correctly (example: `$HOME` as git repo for dotfile management).
It's also not required anymore as the git repo is only accessed when a file is open instead of every keystroke.

Finally, I have improved the git implementation by switching to gitoxide (as suggested in the original PR).
During this switch, I found a few edge cases that were not handled correctly by the previous implementation:
* Non utf-8 encoded files were not supported (now uses document encoding)
* Directories/symlinks that were replaced by files would use the directory children/link path as the diff base (now the diff is not shown)

I have purposefully not added more features to this PR to keep it somewhat small and reviewable.
In the future, I would like to expand on this PR.
In particular, I would like to implement a mechanism to reload the diff base after a file is first opened.
This avoids having to call `:reload` helix after every commit, to continue receiving correct diffs.
However, this feature is more complex (requires watching `.git`) so I choose not to include it here.

~~Currently, all commits are separate to make review easier.
I would like to squash the commits before this is merged, as the implementation changed quite a bit.~~

~~Note that there is currently one test failing in `helix-core` but that seems completely unrelated as the PR does not touch anything in `helix-core`~~

PR #3896, opened 2022-09-18
raise msrv to 1.59

pascalkuthe:
Almost all distributions (even ubuntu) now support 1.59.0.
The original MSRV was set to 1.57 to support void linux and homebrew.
Both now ship version 1.63.

This MSRV bump is required to use gitoxide (see #3890) but many crates
have bumped their MSRV to 1.59 and therefore the same problem will likely also occur for other dependencies in the future.

### DIFF
--- a/.github/workflows/msrv-rust-toolchain.toml
+++ b/.github/workflows/msrv-rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.57.0"
+channel = "1.59.0"
 components = ["rustfmt", "rust-src"]


the-mikedavis:
Have Void and homebrew updated their build rust versions up to / past 1.59?

pascalkuthe:
Yes both are in fact now at 1.63

archseer:
[Ubuntu](https://packages.ubuntu.com/search?keywords=rust) is usually the slowest one to update, apart from [debian that still has 1.48.0 for some](https://packages.debian.org/search?keywords=rust).

Issue #3005, opened 2022-07-08
Missing Code actions and code lenses (Haskell)


0rphee:
### Summary

Many code actions, and code lenses are not available compared to other text editors. 
In this [gist](https://gist.github.com/0rphee/a76874d6b496c88f4d611fb093aa99f0) two functions with multiple code actions available in Kakoune and VS Code can't be done in Helix.

This issue might refer to the same problem as #2994

In the videos below I show the behavior in helix compared to 1. Kakoune, 2. VS Code.

HLS Code actions [(apply hlint fixes)](https://haskell-language-server.readthedocs.io/en/latest/features.html#apply-hlint-fixes)
Helix (left) - Kakoune (right)

https://user-images.githubusercontent.com/79347623/177895723-6012c895-f881-471c-a4d0-834ccbb23d2d.mov

HLS Code lens [(add type signature)](https://haskell-language-server.readthedocs.io/en/latest/features.html#add-type-signature)
Helix (left) - Kakoune (right)

https://user-images.githubusercontent.com/79347623/177896246-060107dc-0c9c-49be-b8ea-38644ba376c0.mov

HLS Code lens and Code actions in VS Code

https://user-images.githubusercontent.com/79347623/177896346-0b893164-54b4-4287-98d1-093d049ef9db.mov

### Reproduction Steps

 

1.  Download the [gist](https://gist.github.com/0rphee/a76874d6b496c88f4d611fb093aa99f0)
2.  Select the whole line of the first function, type `<space> + a`. Only some actions will be available but not all. Try with each token there.
3.  Try the same with the other function.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-07-07T19:48:23.132 helix_lsp::transport [ERROR] err <- "No 'hie.yaml' found. Try to discover the project type!\n"
2022-07-07T19:48:23.137 helix_lsp::transport [ERROR] err <- "Run entered for haskell-language-server-wrapper(haskell-language-server-wrapper) Version 1.7.0.0 aarch64 ghc-9.2.2\n"
2022-07-07T19:48:23.137 helix_lsp::transport [ERROR] err <- "Current directory: /Users/roger/Documents/Dev/Haskell-Projects/Haskell-Programming-From-First-Principles/ch22\n"
2022-07-07T19:48:23.137 helix_lsp::transport [ERROR] err <- "Operating system: darwin\n"
[...]

PR #3898, opened 2022-09-19
Track source and tags in diagnostics

lcycon:
In #3005, Haskell users have been running into an issue where many of the CodeActions supported by the `haskell-language-server` are not present in Helix.

By my read of things:

- `hls` generates diagnostics and sends them to Helix
- Helix stores these, but discards the `source` and `tags` fields
- The user triggers a codeAction query
- Helix sends `textDocument/codeAction` to `hls`, in the `context` field it includes the relevant diagnostic _but without the `source` and `tags` that have been discarded_
- `hls` generates codeActions for all diagnostics it knows about (freshly computing the diagnostics, which include the `source` and `tags` fields), [only returning them if the freshly generated diagnostic matches the diagnostic in the `textDocument/codeAction` query](https://github.com/haskell/haskell-language-server/blob/362ca34488899454f2070f098a5a2ebf25e52c5e/plugins/hls-refactor-plugin/src/Development/IDE/Plugin/CodeAction.hs#L473)
- Since Helix removed the `source` and `tags` fields, this equality match fails and no codeActons are returned (beyond a set of always-applicable ones)

This PR simply tracks these two fields so they can be passed back to the LSP server. I am not sure it solves every case of missing codeAction, but it has solved all the cases I've tried.

I'm not very familiar with Helix's codebase nor the LSP, so please do let me know if I've missed some reason these fields were intentionally left out.

### DIFF
--- a/helix-core/src/diagnostic.rs
+++ b/helix-core/src/diagnostic.rs
@@ -29,6 +29,12 @@ pub enum NumberOrString {
     String(String),
 }
 
+#[derive(Debug, Clone)]
+pub enum DiagnosticTag {
+    Unnecessary,
+    Deprecated,
+}
+
 /// Corresponds to [`lsp_types::Diagnostic`](https://docs.rs/lsp-types/0.91.0/lsp_types/struct.Diagnostic.html)
 #[derive(Debug, Clone)]
 pub struct Diagnostic {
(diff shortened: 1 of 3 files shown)


dead10ck (approved):
This looks good to me, though someone with more experience with the LSP code should take a look as well. Thanks for tracking this down and fixing it!

archseer (approved):
Looks fine, thanks! We just had our own diagnostics type that didn't fully map to the LSP diagnostics.

PR #3900, opened 2022-09-19
Xtask theme previews

### LINKED ISSUE #3608: Add xtask to generate the themes wiki page automatically
It's a little bit distracting that in the [themes wiki page](https://github.com/helix-editor/helix/wiki/Themes) all screenshots are of different size and show different code.

I think it would be possible to create an xtask that:

* iterates all themes
* creates a Document with some juicy soucecode
* renders it
* makes a screenshot
* generates the wiki md including the screenshots
* maybe a github hook to automize the whole process

hirschenberger:
Add xtask to generate html previews for all themes. Requires the `aha` tool installed to convert ANSI sequences to HTML. `aha` should be available for your linux distribution of choice.

This is still WIP and needs some discussion how and where to integrate the previews to the docs.

fixes #3608

### DIFF
--- a/xtask/Cargo.toml
+++ b/xtask/Cargo.toml
@@ -9,4 +9,9 @@ edition = "2021"
 helix-term = { version = "0.6", path = "../helix-term" }
 helix-core = { version = "0.6", path = "../helix-core" }
 helix-loader = { version = "0.6", path = "../helix-loader" }
+helix-view = { version = "0.6", path = "../helix-view" }
+tokio = { version = "1", features = ["rt", "rt-multi-thread", "io-util", "io-std", "time", "process", "macros", "fs", "parking_lot"] }
+crossterm = { version = "0.25", features = ["event-stream"] }
 toml = "0.5"
+gag = "1.0"
+temp-file = "0.1"
--- a/xtask/src/main.rs
+++ b/xtask/src/main.rs
@@ -197,6 +197,10 @@ pub mod path {
         project_root().join("book/src/generated/")
     }
 
+    pub fn book_gen_theme_previews() -> PathBuf {
+        book_gen().join("themes")
+    }
+
     pub fn ts_queries() -> PathBuf {
         project_root().join("runtime/queries")
     }
@@ -252,6 +256,69 @@ pub mod tasks {
         Ok(())
     }
 
+    pub fn snap_theme() -> Result<(), DynError> {
+        use crossterm::event::EventStream;
+        use helix_term::{application::Application, args::Args, config::Config};
+        use helix_view::theme;
+
+        let themes =
+            theme::Loader::new(&helix_loader::config_dir(), &helix_loader::runtime_dir()).names();
+
+        let rt = tokio::runtime::Runtime::new().unwrap();
+
+        for theme in themes {
+            rt.block_on(async {
+                // create an application instance with a nice soucefile and the current theme
+                let args = Args {
+                    files: vec![("helix-core/src/register.rs".into(), (16, 15).into())],
+                    ..Default::default()
+                };
+                let config = Config {
+                    theme: Some(theme.clone()),
+                    ..Default::default()
+                };
+                let mut app = Application::new(args, config).unwrap();
+
+                let tmp_file = temp_file::empty();
+
+                // redirect stdout to some temp file
+                let opts = std::fs::OpenOptions::new()
+                    .truncate(true)
+                    .write(true)
+                    .create(true)
+                    .open(tmp_file.path())
+                    .unwrap();
+                let redirect = gag::Redirect::stdout(opts).unwrap();
+
+                // render the editor and stop the application after some time
+                let mut stream = EventStream::new();
+                tokio::select! {
+                    _ = tokio::time::sleep(std::time::Duration::from_millis(10)) => {},
+                    _ = app.run(&mut stream) => {}
+                }
+
+                // convert the captured ANSI sequence to html using the external `aha` tool and
+                // capture stdout
+                let html_out = tokio::process::Command::new("aha")
+                    .arg("-l")
+                    .arg("-f")
+                    .arg(tmp_file.path())
+                    .output()
+                    .await
+                    .unwrap();
+
+                let out_path = crate::path::book_gen_theme_previews();
+                let _ = tokio::fs::create_dir(&out_path).await;
+
+                // dump `aha` output to the final html file
+                tokio::fs::write(out_path.join(format!("{}.html", theme)), html_out.stdout)
+                    .await
+                    .unwrap();
+            });
+        }
+        Ok(())
+    }
+
     pub fn print_help() {
         println!(
             "
@@ -260,6 +327,7 @@ Usage: Run with `cargo xtask <task>`, eg. `cargo xtask docgen`.
     Tasks:
         docgen: Generate files to be included in the mdbook output.
         query-check: Check that tree-sitter queries are valid.
+        snap-theme: Make a snapshot of a theme
 "
         );
     }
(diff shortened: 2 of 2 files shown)


archseer:
Do we need to have a preview doc though? The users can cycle over `:theme` completions and see how each theme would look like with their codebase

hirschenberger:
> Do we need to have a preview doc though? The users can cycle over `:theme` completions and see how each theme would look like with their codebase

see #3608 for the intention of this PR. There's a preview for all themes in the wiki, but they look different.

pickfire:
I suggested this to @AlexanderBrevig in one of the pull requests but looks like you started work on it. It's definitely good to be able to compare different themes on the same code, otherwise it looks different on different languages, one thing I think would be interesting is to be able to cover a wide range of syntax (maybe we need a language switcher for that), since later users would strive to make it look good on just the screenshots and not good on real use (missing some theming).

hirschenberger:
We can use two nested loops:  Over all languages (having a folder with some pretty code snippets) and all themes and pregenerate all previews.

Or we can create an interactive webapp, that generates the previews on-the-fly and also can provide the possibility to provide own snippets. But I assume that can't be hosted on github anymore.

AlexanderBrevig:
I've been playing with doing this from github actions and publishing a wiki page with screenshots (from a headless wayland docker monstrosity). I must say that this path feels a lot more appealing to me. Makes sense to have the themes in the docs, and when it's just HTML there's no real size issues either.

hirschenberger:
> Asciinema



> EDIT: Would Asciinema be an alternative worth considering? Instead of `aha`? Same idea just a different technology for accomplishing the same goal.

Asciinema seems to be tailored towards video creation for their own hosting platform. What would be the benefit using videos? Iterating all themes or all languages with one theme?

It seems to be easy to create asciinema from the captured ANSI sequence but I see no benefit to static HTML files.

AlexanderBrevig:
> Asciinema seems to be tailored towards video creation for their own hosting platform. What would be the benefit using videos? Iterating all themes or all languages with one theme per video?
> 
> It seems to be easy to create asciinema files from the captured ANSI sequence but I see no benefit to static HTML files.

Asciinema saves to static text files, and easily embeddeble with their player. 
The way [WezTerm](https://wezfurlong.org/wezterm/colorschemes/g/index.html#gruvbox-dark) does it has come up before.

hirschenberger:
> The way [WezTerm](https://wezfurlong.org/wezterm/colorschemes/g/index.html#gruvbox-dark) does it has come up before.

I must say I don't really like using an external player just to show a single frame. I see no benefit to the current solution.

AlexanderBrevig:
> > The way [WezTerm](https://wezfurlong.org/wezterm/colorschemes/g/index.html#gruvbox-dark) does it has come up before.
> 
> I must say I don't really like using an external player just to show a single frame. I see no benefit to the current solution.

Sure :) I guess the benefit would be that we don't need to maintain workarounds to get rendering perfect. It would also mean that we _could_ generate an actual sequence showing off menus, a few languages, and markdown.

There's also the question of whether we want to enforce an additional dependency from the environment for people running this command.

Keep in mind, I'm just thinking out loud here and have no real say in the matter.

hirschenberger:
> > > The way [WezTerm](https://wezfurlong.org/wezterm/colorschemes/g/index.html#gruvbox-dark) does it has come up before.
> > 
> > 
> > I must say I don't really like using an external player just to show a single frame. I see no benefit to the current solution.
> 
> Sure :) I guess the benefit would be that we don't need to maintain workarounds to get rendering perfect. It would also mean that we _could_ generate an actual sequence showing off menus, a few languages, and markdown.
> 
> There's also the question of whether we want to enforce an additional dependency from the environment for people running this command.
> 
> Keep in mind, I'm just thinking out loud here and have no real say in the matter.

I'm also not in the position to decide. Both approaches have their pros and cons. I try to summarize it, please correct me if I miss something.

### HTML

* 👍🏼  No player needed to view, so offline capable
* 👍🏼  Already implemented as PoC
* 👎🏼  Needs thirdparty tool `aha` to generate the HTML locally
* 👎🏼  Can only generate `screenshots`

### Asciinema

* 👍🏼  No thirdparty tool needed to generate docs locally
* 👍🏼  Can generate clips and images.
* 👎🏼  Depends on thirdparty service provider (asciinema.org)
* 👎🏼  Must be online to view docs (themes), also locally generated.

David-Else:
Having a built in command to export the current document to HTML with syntax highlighting would be fantastic.

Neovim's https://neovim.io/doc/user/syntax.html#:TOhtml command is broken for treesitter syntax highlighting. It would be great to have it in Helix, would be perfect to create syntax highlighted code blocks in blogs/documentation without using https://prismjs.com/, much cleaner.

hirschenberger:
My initial Idea was to use the [ansi-to-html](https://github.com/Aloso/to-html/tree/main/crates/ansi-to-html) crate to generate the HTML, but unfortunately it does not support the **CUP** (cursor position) code, so the output is garbage.

Another approach would be to add support for CUP to **ansi-to-html** and have a version without external dependencies.

hirschenberger:
> Having a built in command to export the current document to HTML with syntax highlighting would be fantastic.

This seems to be a different topic, as this PR does create a **screenshot** of the whole editor with code.

@David-Else Maybe it makes sense to open it's own issue for that?

AlexanderBrevig:
> * 👎🏼  Depends on thirdparty service provider (asciinema.org)
> * 👎🏼 Must be online to view docs (themes), also locally generated.

We _could_ vendor the [player](https://github.com/asciinema/asciinema-player) and host the recordings in our repo (much like the html, its just json instead). 

👎🏻 ➡️ 👍🏻

hirschenberger:
Ok, I'll create a PoC, let's see how it works with Asciinema

AlexanderBrevig:
> Ok, I'll create a PoC, let's see how it works with Asciinema

It may very well be that the maintainers like the HTTP route more. Keep in mind before spending time on this! I don't want you to feel like you've wasted time because of me. 
But if you do it for fun and profit, please let me know your experience.

I personally very much like the idea that the themes are documented - in a similar and comparable way - in the docs and not in the github wiki 👍🏻 😄

hirschenberger:
Ok, pandora's box is opened 🙈  

I'm writing a asciinema preprocessor for mdbook for a tight integration from within markdown.

the-mikedavis:
Converting from escape sequences to HTML is over-engineering it IMO. It should be possible to do this by building an SVG or HTML document manually by consuming the `helix_core::syntax::Syntax::highlight_iter` iterator. Tree-sitter-cli's highlight command does the same https://github.com/tree-sitter/tree-sitter/blob/67d13cf50b0b74f7037adec314a571bd68546ebf/highlight/src/lib.rs#L888

hirschenberger:
> Converting from escape sequences to HTML is over-engineering it IMO. It should be possible to do this by building an SVG or HTML document manually by consuming the `helix_core::syntax::Syntax::highlight_iter` iterator. Tree-sitter-cli's highlight command does the same https://github.com/tree-sitter/tree-sitter/blob/67d13cf50b0b74f7037adec314a571bd68546ebf/highlight/src/lib.rs#L888

But this would only create HTML of the code, not of the editor itself, right? I'd like it more if you also see how the editor looks like with the current theme.

IMHO it's also more KISS to use an established and well-tested approach (ANSI -> HTML) that asciinema, aha, and others already have, than fiddling around with handcrafting HTML.

archseer:
> There's a preview for all themes in the wiki, but they look different.

Yeah, but we could remove that page altogether? It served a purpose when themes didn't have a live preview, but now you can just tab through.

hirschenberger:
> > There's a preview for all themes in the wiki, but they look different.
> 
> Yeah, but we could remove that page altogether? It served a purpose when themes didn't have a live preview, but now you can just tab through dozens of themes and see how they'd look on your code.

So you want to say that PR and issue is unnecessary?

archseer:
I'm just trying to gauge if this is desirable or not. If users want a html doc with these then that's okay, I'm just saying personally: I preview these in the editor now.

pascalkuthe:
I don't see usadding this to core and this PR has gone stale so I am going to close this. Thank you for contributing!

PR #3913, opened 2022-09-20
Bump MSRV to 1.61 and document MSRV policy

pascalkuthe:
The MSRV was previously bumpbed in to 1.59 in #3896.
However the newest gitoxide version (which contains some nice API additions usable in #3890) requires 1.60, because the bstr crate bumped its MSRV.
Before #3896 the MSRV was set to 1.57 to support void linux and homebrew.
Both now ship rust 1.63.
I checked [repology](https://bugs.launchpad.net/ubuntu/+source/rustc/+bug/1986648) and all distributions that ship the newest helix version also already ship at least rust 1.61.

However setting the MSRV to some arbitrary value again seemed bad to me and so I looked into MSRV in more detail.
The (mainstream) package repository which updates rustc last are usually ubuntu, debian and enterprise distributions like REHL.
Debian and REHL ship ancient rustc versions anyway that are hard to support (and already not supported by helix).
Ubuntu bumps rustc whenever a new firefox version raises their MSRV.

Therefore I propose tying the MSRV of helix to that of firefox.
Most package managers want to ship the newest firefox releases to attain the latest security fixes and treat updating any dependencies of that as a high priority.
For example ubuntu is currently in the [process](https://bugs.launchpad.net/ubuntu/+source/rustc/+bug/1986648) of updating to rust 1.61 for firefox 105 which released today.
Firefox in turn trys to be somewhat conservative with its MSRV to remain compatible with many distributions.
Therefore firefox MSRV currently represents a lowest common denomiator among most distributions.

I apologize for the duplicate PR regarding MSRV.
However I believe that with a formal policy regarding MSRV in place such a situation can be avoided in the future.
This should make maintenance and development easier as its always clear what the current MSRV is and when it will be change.

### DIFF
--- a/.github/workflows/msrv-rust-toolchain.toml
+++ b/.github/workflows/msrv-rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.59.0"
+channel = "1.61.0"
 components = ["rustfmt", "rust-src"]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -485,7 +485,6 @@ dependencies = [
  "log",
  "once_cell",
  "pulldown-cmark",
- "retain_mut",
  "serde",
  "serde_json",
  "signal-hook",
@@ -883,12 +882,6 @@ checksum = "3acd125665422973a33ac9d3dd2df85edad0f4ae9b00dafb1a05e43a9f5ef8e7"
 dependencies = [
  "winapi",
 ]
-
-[[package]]
-name = "retain_mut"
-version = "0.1.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4389f1d5789befaf6029ebd9f7dac4af7f7e3d61b69d4f30e2ac02b57e7712b0"
 
 [[package]]
 name = "ropey"
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -37,6 +37,12 @@ contributors are strongly encouraged to write integration tests for their code.
 Existing tests can be used as examples. Helpers can be found in
 [helpers.rs][helpers.rs]
 
+## Minimum Stable Rust Version (MSRV) Policy
+
+Helix follows the MSRV of Firefox.
+The current MSRV and future changes to the MSRV are listed in the [Firefox documentation].
+
+[Firefox documentation]: https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html
 [good-first-issue]: https://github.com/helix-editor/helix/labels/E-easy
 [log-file]: https://github.com/helix-editor/helix/wiki/FAQ#access-the-log-file
 [architecture.md]: ./architecture.md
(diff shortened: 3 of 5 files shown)


pickfire (approved):
I personally would prefer 1.60 since that is what bstr follows (and it solves https://github.com/helix-editor/helix/pull/3890#discussion_r975270738) but I think this looks good too, I think it would be best for someone else to have their opinion on this given that I use a rolling distro (arch) myself and likely won't be affected.

archseer:
Ubuntu still seems to carry 1.59: https://packages.ubuntu.com/search?keywords=rustc

archseer:
With the bump to 1.61, https://github.com/helix-editor/helix/blob/4c9f144dace484ca99cc84053493ed7b5f6946ac/helix-term/src/ui/picker.rs#L368-L370 should now be removed together with the crate import

pascalkuthe:
I linked to the current development thread from ubuntu. Tey usually take a few days to bump rustc and firefox but they have it as a priority high issue right now they are working on  it.

Firefox 105 (which requires 1.61) only released today so its quite close. If that is preferred I couls ping the PR again when Ubuntu updates in a few days (and it would be merged then). Although helix is currently not packaged for Ubuntu and developers usually all use rustup anyway so it doesn't matter too much

pascalkuthe:
> With the bump to 1.61,
> 
> https://github.com/helix-editor/helix/blob/4c9f144dace484ca99cc84053493ed7b5f6946ac/helix-term/src/ui/picker.rs#L368-L370
> should now be removed together with the crate import

added this to the PR

EpocSquadron:
@pascalkuthe thanks for putting in the work to identify a reasonable policy! These kinds of efforts often go unappreciated, but are very important.

README.md
```diff
@@ -76,6 +76,11 @@ Helix can be installed on MacOS through homebrew:
 brew install helix
 ```
 
+## Minimum Stable Rust Version (MSRV) Policy
+
+Helix follows the MSRV of firefox.
+The current MSRV and future changes to the MSRV are listed in the [firefox documentation](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html).
```
the-mikedavis, line 82:
```suggestion
Helix follows the MSRV of Firefox.
The current MSRV and future changes to the MSRV are listed in the [Firefox documentation](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html).
```

the-mikedavis: I'm not sure this needs such an explicit mention in the README. I would either put it as a comment in `rust-toolchain.toml` or in `CONTRIBUTING.md` since it's an implementation detail.

pascalkuthe: I think having the MSRV policy document in a visisble place is helpful (both for package maintainers aswell as developers) but I agree that `CONTRIBUTING.md ` is probably the better place. I have moved the the section on MSRV from `README.md` to `CONTRIBUTING.md~ (and fixed the spelling)

CBenoit (commented):
Greatly appreciated :100:

kirawi:
It looks like Ubuntu updated yesterday.

PR #3915, opened 2022-09-20
Keep arrow and special keys in insert

pickfire:
Advanced users won't need it and is useful for beginners. Revert part of #3671.

Beginners who do not read documentation rely on these keys, having these keys reduce the effort for them to get started, we are turning down 3 groups of people without it:

- those that have no prior experience to modal editors
- those that does not have time/effort to read docs or dig in issues to find out how to add the worse keys
- those that even after they read docs they don't know how to change the config (well they need an editor to do it first, and helix is an editor, so this means we force them to read the docs on how to use helix better keys to edit helix config to add the worse keys if they don't use other editors?)

But if we have it, it does not affect advanced users, but it does have a chance to make those beginners lazy to learn, but I don't think we can do much about that.

# Pros

- lower barrier of entry, groups mentioned above can use helix without much guidance, steering helix towards being usable with zero-config even for beginners
- users new to modal editors does not need to read docs to get started if they know a couple of basic vi keys: <kbd>i</kbd>, <kbd>esc</kbd>, `:wq`

# Cons

- user left with less motivation to learn the better keys and stick in their comfort zone
- we may get some weird bugs which we don't cover for in insert keys due to less usage of arrow/special keys

# Alternatives

We can let users use another editor if they don't want to learn or read docs, but I don't see much harm in doing so for the existing users.

I didn't keep emacs keys because most likely they already know the better keys and they can read docs if they use emacs, so it is fine not keeping it given that the keys are similar in terms of ergonomic.

The above is just my personal opinion, I am not sure about other team members. I personally don't use these keys in both insert and normal mode.

Poll was created in https://github.com/helix-editor/helix/discussions/3865 some people like it, some don't like it but I know it broke someones' workflow that people search the issues and wrote an essay about it


book/src/keymap.md
```diff
@@ -335,32 +335,22 @@ experience.
...
-"pagedown" = "page_down"
-"home" = "goto_line_start"
-"C-a" = "goto_line_start"
-"end" = "goto_line_end_newline"
-"C-e" = "goto_line_end_newline"
-"A-left" = "goto_line_start"
-```
+### Insert Mode (beginners)
+
+These keys are not recommended, it is for those who do not read documentations.
```
pickfire, line 340: I am not sure if I even need this line given that most likely they won't even read documentation.

archseer:
You need to also revert the tutor change: https://github.com/helix-editor/helix/discussions/3865#discussioncomment-3666619

book/src/keymap.md
archseer: This should say "These keys are not recommended, but are included for new users less familiar with modal editors"

dead10ck:
I think it's too soon to be reverting this. My position isn't set in stone; I've heard these arguments, and there is validity to helping new users transition, but even so, at this point I still think this

> * user left with less motivation to learn the better keys and stick in their comfort zone
> * we may get some weird bugs which we don't cover for in insert keys due to less usage of arrow/special keys

Is presently still very much a thing. I don't believe all users who like to use the arrow keys are people who are too lazy to learn their editor, but I also don't believe in making decisions to cater to such a user base. This is a different editor, and some learning is required.

It could turn out that the goal of removing them will not be accomplished, and/or it will cause more harm than is worth it. If this turns out to be the case, I will not object to reverting. But at least initially, some push back from existing users was fully expected, but we won't know the overall effect on new users until probably the couple of weeks after the next release.

helix-term/src/keymap/default.rs
```diff
@@ -360,6 +360,17 @@ pub fn default() -> HashMap<Mode, Keymap> {
 
         "C-x" => completion,
         "C-r" => insert_register,
+
+        "up" => move_line_up,
+        "down" => move_line_down,
+        "left" => move_char_left,
+        "right" => move_char_right,
+        "C-left" => move_prev_word_end,
+        "C-right" => move_next_word_start,
```
archseer, line 369: I don't think it's right to restore `C-` mappings but not the `A-` mappings.

archseer: Though that seems to be the nvim default so maybe it's ok

archseer:
> overall effect on new users until probably the couple of weeks after the next release.

I think that's too late to decide, we should make the decision before the next release.

archseer:
I'm thinking we could match kakoune's insert mode: Only arrow keys and home/end, no ctrl or alt mappings: https://github.com/mawww/kakoune#33-insert-mode

dead10ck:
That sounds reasonable to me. Keep the basic bindings, but don't make the default overly cozy for staying in insert mode.

book/src/keymap.md
aral: Genuinely happy to see this. Again, not for me (I’ve already added the mappings back in my config and am running from source) but in terms of the culture of the project and what it says for how new folks are welcomed. This might seem like a tiny thing but it does make a world of difference in terms of how someone feels when they’re not familiar with a new tool.

This is saying “hey, we recognise this is how nearly everything out there works and that you have muscle memory you’re bringing with you… that’s ok. This tool is designed with a different workflow in mind – which you’ll learn in time unless we lose you in the first 30 seconds – but we will cater for your needs to make your introduction more seamless.”

Happy to see Helix embracing a more open/welcoming approach than “our way or the highway.” I truly believe that (whether by accident or not), this is one of those things that has brought a lot of people who would otherwise not have used a modal editor to Helix.

:two_hearts:

aral:
Sounds good to me too :)

book/src/keymap.md
```diff
@@ -335,32 +335,22 @@ experience.
...
+| Key                                         | Description                 | Command                 |
+| -----                                       | -----------                 | -------                 |
+| `Up`                                        | Move to previous line       | `move_line_up`          |
+| `Down`                                      | Move to next line           | `move_line_down`        |
+| `Left`                                      | Backward a char             | `move_char_left`        |
+| `Right`                                     | Forward a char              | `move_char_right`       |
+| `Ctrl-Left`                                 | Backward a word             | `move_prev_word_end`    |
+| `Ctrl-Right`                                | Forward a word              | `move_next_word_start`  |
+| `PageUp`                                    | Move one page up            | `page_up`               |
+| `PageDown`                                  | Move one page down          | `page_down`             |
```
sudormrfbin, line 351: Page down/up will have to be removed too if we're going with only arrow keys and home/end.

helix-term/src/keymap/default.rs
pickfire: I only added back `C-` keys because like other editors and browsers, `C-` is used to move between word, not `A-` so the familiarity comes from there.

book/src/keymap.md
pickfire: Should we also support page down/up since it is one special key keyboard have that is similar to arrow keys and home/end, and it works without pressing `C-`? If another I think page up/down is more useful than home/end.

pickfire:
> Is presently still very much a thing. I don't believe all users who like to use the arrow keys are people who are too lazy to learn their editor, but I also don't believe in making decisions to cater to such a user base. This is a different editor, and some learning is required.

What do you suggest for someone who starts learning helix as the first editor? How do you suggest them to open up helix as the first editor? Do we point them to use other editor to edit the config file and then come back to use helix?

Will work on the changes tonight.

dead10ck:
@pickfire

>What do you suggest for someone who starts learning helix as the first editor? How do you suggest them to open up helix as the first editor? Do we point them to use other editor to edit the config file and then come back to use helix?

The very first thing the helix usage docs suggest is opening the tutor with `hx --tutor`, and the very first thing in the tutor explains hjkl. The rest is there for them to read. If they can't manage that, then I don't see any hope of them ever becoming productive with helix, arrow keys or not.

helix-term/src/keymap/default.rs
archseer: Let's remove `C-left`, `C-right` and `pageup`, `pagedown` to match kakoune.

book/src/keymap.md
```diff
@@ -335,32 +335,22 @@ experience.
...
-"A->" = "goto_file_end"
-"pageup" = "page_up"
-"pagedown" = "page_down"
-"home" = "goto_line_start"
-"C-a" = "goto_line_start"
-"end" = "goto_line_end_newline"
-"C-e" = "goto_line_end_newline"
-"A-left" = "goto_line_start"
-```
+### Insert Mode (beginners)
```
archseer, line 338:
```suggestion
### Insert mode
```

NomisIV:
Instead of removing the arrow keys only to add them again, wouldn't it make more sense to add this as an option? I suggest it would be enabled for new users (so either no value = enabled, or no value = disabled, but the default config file enables it). That way new users can still use the arrow keys if they so wish, but more advanced users can disable them if they are unneeded. Perhaps, even a warning comment could be added to the line enabling it in the default config, suggesting the users to turn the feature off to fully get into the helix workflow. Would there be any downsides to this approach?

the-mikedavis:
> wouldn't it make more sense to add this as an option?

I think it's distasteful to add a config option for anything you can accomplish with keybindings.

NomisIV:
Ah, I wasn't aware it was possible to re-add the functionality by configuring the keybindings. In that case I also consider the configuration option redundant. Perhaps that option would make more sense would be if it disabled helix' modal editing completely, but that's probably outside the scope of this PR anyway.

pickfire:
> Instead of removing the arrow keys only to add them again, wouldn't it make more sense to add this as an option? I suggest it would be enabled for new users (so either no value = enabled, or no value = disabled, but the default config file enables it). That way new users can still use the arrow keys if they so wish, but more advanced users can disable them if they are unneeded. Perhaps, even a warning comment could be added to the line enabling it in the default config, suggesting the users to turn the feature off to fully get into the helix workflow. Would there be any downsides to this approach?

Existing users can always disable it by adding `"noop"` for those keys.

pickfire:
> You need to also revert the tutor change: https://github.com/helix-editor/helix/discussions/3865#discussioncomment-3666619

Regarding this, I think it is better for us to not revert the tutor change. Since we are not recommending users to use it, users probably expected that it works and they can easily just try it out.

>  Note: If you want to move the cursor while in Insert mode,
>       you may use the arrow keys instead of exiting and
>       reentering Insert mode.

dylrich:
I want to add that I think framing it as an advanced users vs. new users discussion misses that many folks who are very comfortable with modal movement sometimes use arrow keys in insert mode because it can be simply faster. On my keyboard I can reach the arrow keys easily with my pinky while on the home row, but I cannot reach the escape key without moving my left hand out of position. This means that the "activation energy" to change modes is somewhat high relative to just hitting an arrow key, so for one character movements it's a large speed gain to hit the arrow key. Granted, this greatly depends on the keyboard you have, how exactly you position your hands while typing, and even the size of your hands, but I think it's important to consider that there are legitimate reasons to want to move while in insert mode outside of the person who is simply new to the insert/normal mode split.

dead10ck:
@dylrich that's a good point, in retrospect that assumption that "advanced users" would not want to use arrow keys is not correct.

aral:
I mean, I don’t know if I qualify as an “advanced user” but I know I’ve been coding for 39 years (since I was 7) and I use the arrow keys ;)

Also, I feel old now :)

pedronasser:
I feel this kind of decision is really bad for the future of Helix, and honestly I am disappointed.
I would rather see helix remove all default keybindings instead and maybe add an argument to load the "default bindings".
By imposing these types of opinionated changes, you are clearly signaling that you don't helix to be used by anyone besides the people who already is used to the modal editing. Which is a very bad way to introducing the modal editing approach new users.
Helix should be new user friendly to a point.

But who cares? I am not a core contributor.

archseer:
From the vision doc:

> Don't try to be everything for everyone. There are many great editors out there to choose from. Let's make Helix one of those great options, with its own take on things.

pedronasser:
> From the vision doc:
> 
> > Don't try to be everything for everyone. There are many great editors out there to choose from. Let's make Helix one of those great options, with its own take on things.

There's a difference between having your own take on things, and not being perfect for everything and having the minimum requirement for a text editor for a new user.
There's also a big difference between suggesting a better approach and basically requiring the user to learn by force the new approach.
Yeah, a new user could configure it before starting to use it, but if you have any experience with software you know that new users often doesn't start by the config file. And if you want to make the user always configure before using then like I said, just remove all default bindings and be clear about that in the doc and maybe also remove the "A post-modern modal text editor" part from it.

dead10ck:
@pedronasser

>having the minimum requirement for a text editor for a new user

Big assumption on your part to presume that you know what new users need.

>Yeah, a new user could configure it before starting to use it, but if you have any experience with software you know that new users often doesn't start by the config file

Or, you could spend the time to read the docs, and practice the editing style. The very first page of the Usage docs suggest reading through the tutor. I don't know about you or anyone else, but whenever I use a new tool, reading docs is the first thing I do.  Yes, this requires more work up front. Maybe this comes down to individuals' learning styles, but I personally tend to front-load and do my learning first. I don't want to speak for anyone, but perhaps it's possible this conflict is coming from people who tend to like to learn by doing first, and only read when they get stuck.

But modal editing is different than modeless editing. It is sufficiently different that this will not work. You will not get by on previous experience, arrow keys or not.

PR #3917, opened 2022-09-20
Add mix.exs and mix.lock as Elixir root markers

rbino:
Simplify opening an Elixir project placed in a monorepo

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -99,7 +99,7 @@ scope = "source.elixir"
 injection-regex = "(elixir|ex)"
 file-types = ["ex", "exs", "mix.lock"]
 shebangs = ["elixir"]
-roots = []
+roots = ["mix.exs", "mix.lock"]
 comment-token = "#"
 language-server = { command = "elixir-ls" }
 config = { elixirLS.dialyzerEnabled = false }


the-mikedavis:
Does this work properly in umbrellas? We may want to only use `mix.lock` since each umbrella app would have its own `mix.exs`

archseer:
I wonder how we are doing the scanning: I'd scan first and look for `Cargo.lock` for example, then fall back to `Cargo.toml` in Rust

the-mikedavis:
It looks like we try each root file for each ancestor directory: https://github.com/helix-editor/helix/blob/5467c65f8224049405498d65f54c3e3f8e4ec6c3/helix-loader/src/lib.rs#L115-L117

It could make sense to try each root file separately so that when `mix.lock` is at a higher ancestor, it still beats `mix.exs`?

rbino:
So, it seems I opened a whole new can of worms :smile:
Reading [this comment](https://github.com/helix-editor/helix/blob/5467c65f8224049405498d65f54c3e3f8e4ec6c3/helix-core/src/lib.rs#L44) and the line @the-mikedavis linked I thought this should've worked correctly, because in an umbrella project there would be at least one of the markers (namely `mix.lock`) in the top-most ancestor and should've been returned by `find_root`.

Turns out the implementation didn't match what the comment was saying, so I opened https://github.com/helix-editor/helix/pull/3929. Using my branch from that PR, this actually works correctly (i.e. it uses the umbrella root even when opening Helix with `hx .` inside the `apps/app_in_umbrella` directory

rbino:
@the-mikedavis now that #3929 is merged this works correctly in umbrella apps

the-mikedavis (approved):

the-mikedavis:
Thank you!

PR #1249, opened 2021-12-11
Add support for local language configuration

kirawi:
Resolves #1217, allowing users to have configuration for local directories. It defaults to looking for every `.helix` folder until it finds a `.git` folder or is at the root, and merges all the configs in order.

PR #3929, opened 2022-09-21
Split helix_core::find_root and helix_loader::find_local_config_dirs

rbino:
The documentation of find_root described the following priority for detecting a project root:
- Top-most folder containing a root marker in current git repository
- Git repository root if no marker detected
- Top-most folder containing a root marker if not git repository detected
- Current working directory as fallback

The commit contained in https://github.com/helix-editor/helix/pull/1249 extracted and changed the implementation of find_root in find_root_impl, actually reversing its result order (since that is the order that made sense for the local configuration merge, from innermost to outermost ancestors).

Since the two uses of find_root_impl have different requirements (and it's not a matter of reversing the order of results since, e.g., the top repository dir should be used by find_root only if there's not marker in other dirs), this PR splits the two implementations in two different specialized functions.

In doing so, find_root_impl is removed and the implementation is moved back in find_root, moving it closer to the documented behaviour thus making it easier to verify it's actually correct

### DIFF
--- a/helix-core/src/lib.rs
+++ b/helix-core/src/lib.rs
@@ -46,7 +46,7 @@ pub fn find_first_non_whitespace_char(line: RopeSlice) -> Option<usize> {
 /// * Git repository root if no marker detected
 /// * Top-most folder containing a root marker if not git repository detected
 /// * Current working directory as fallback
-pub fn find_root(root: Option<&str>, root_markers: &[String]) -> Option<std::path::PathBuf> {
+pub fn find_root(root: Option<&str>, root_markers: &[String]) -> std::path::PathBuf {
     let current_dir = std::env::current_dir().expect("unable to determine current directory");
 
     let root = match root {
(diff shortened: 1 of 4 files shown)


kirawi (approved):
Nice catch.

helix-core/src/lib.rs
```diff
@@ -47,9 +47,40 @@ pub fn find_first_non_whitespace_char(line: RopeSlice) -> Option<usize> {
...
+            top_marker = Some(ancestor);
+        }
+
+        // Don't go higher than repo if we're in one
+        if ancestor.join(".git").is_dir() {
+            // Top marker is repo root if not root marker is detected
+            if top_marker.is_none() {
+                top_marker = Some(ancestor);
+            }
+        }
```
the-mikedavis, line 79: These two `if`s can be combined

rbino: Actually a `break` was missing to stop the search as soon as we hit the git repo root

helix-core/src/lib.rs
```diff
@@ -47,9 +47,41 @@ pub fn find_first_non_whitespace_char(line: RopeSlice) -> Option<usize> {
...
+            if top_marker.is_none() {
+                top_marker = Some(ancestor);
+            }
+            // Don't go higher than repo if we're in one
+            break;
+        }
+    }
+
+    // Return the found top marker or the current_dir as fallback
+    top_marker.map(|a| a.to_path_buf()).or(Some(current_dir))
```
the-mikedavis, line 84: It looks like it's not possible for this Option to be returned as None now, so let's make this function return just a `std::path::PathBuf`

the-mikedavis (approved):

archseer:
Nice catch! Thanks for a detailed writeup.

PR #3932, opened 2022-09-22
Exit select after rotate selections

pickfire:
After rotating selections users is probably done with the rotation they want, so it would be better to exit select mode rather than they have to press v again.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3946,6 +3946,7 @@ fn rotate_selection_contents(cx: &mut Context, direction: Direction) {
     );
 
     doc.apply(&transaction, view.id);
+    exit_select_mode(cx);
 }
 
 fn rotate_selection_contents_forward(cx: &mut Context) {


gavincrawford (commented):
Code LGTM, but could you clarify a little further on why exactly this makes the editor more fluent for the user?

pickfire:
After I rotate the contents, then I moved the cursor and I realized I am still in extend mode, and have to quit it, I expected that after modification extend mode should be reverted back to insert mode.

dead10ck (approved):
Am I understanding correctly that this would leave all the selections, but just go back to normal mode? So if the user wants to rotate multiple times, this does not interfere with that?

If so, looks good to me.

pickfire:
Yes, it does not touch selections, it only change back to normal mode, I believe it's unlikely user want to rotate and extend again then rotate and then extend, that would be rare I guess.

gavincrawford (approved):
LGTM

gavincrawford:
Thanks for clarifying! :-)

the-mikedavis:
I'm split on this - I think this change makes sense but overall I'm not sure any command should automatically exit select mode. It would be simpler (easier to learn / memorize) if you could only ever use escape to exit select mode. If there are many commands that automatically exit select mode, it might become really nuanced or unpredictable.

gavincrawford:
> I'm split on this - I think this change makes sense but overall I'm not sure any command should automatically exit select mode. It would be simpler (easier to learn / memorize) if you could only ever use escape to exit select mode. If there are many commands that automatically exit select mode, it might become really nuanced or unpredictable.

I see what you mean. It _is_ inconsistent, but it may be slightly more fluent.
Is it more important to keep consistency than to make this change?

dead10ck:
Consistency is a strong explicit goal, but I don't think it should necessarily come at the detriment of more practical usability. For example, I think it's safe to assume most people would prefer the delete command to exit select mode.

archseer:
Yeah deletes and changes exit select mode because they do an action. For me it makes sense to stay in extend mode since I might be extending a selection, then cycling somewhere further in the file and continuing to extend.

pickfire:
> Yeah deletes and changes exit select mode because they do an action. For me it makes sense to stay in extend mode since I might be extending a selection, then cycling somewhere further in the file and continuing to extend.

Yes, it is more consistent, but I see that as a rare case, cycling the selection is already "changing" the text to a certain extent, so it may or may not be consistent depending on whether you view cycling the selection is a change operation.

In terms of practicality, extending after rotating have less practical use compared to moving the cursors after selection, at least so far I never had that use case.

Extending after rotating I mean like changing `aaaaa bbbbb ccccc ddddd eeeee` to `abcde bcdea cdeab deabc eabcd` or something similar, I see no point doing that or something similar during coding except when doing code golf. On the other hand, rotating multiple variables and then move the cursor to do something else, I always used that.

pascalkuthe:
clsoing this PR as stale. I also think we likely won't be going forward with this one. Thank you for contributing!

Issue #396, opened 2021-06-30
Show the context of the currently visible buffer contents à la context.vim


CBenoit:
#### Describe your feature request

See [`context.vim`](https://github.com/wellle/context.vim) for yourself. I think we may achieve this using tree sitter.

May be built-in or provided by a plugin.

sudormrfbin:
This is a _very_ useful feature to have, and it'd be nice if we could add it to core. For reference there's a treesitter port of context.vim which I use - https://github.com/romgrk/nvim-treesitter-context

cessen:
This looks really cool!  I don't think it belongs in core, but definitely as a plugin.

David-Else:
I thought it was amazing in Neovim briefly until I realized it was just a workaround for the bad practice of having modules that are too big. I soon uninstalled it, it is just not needed if you stick to best practices in the first place.

I would also say it is not needed in core along with @cessen . @the-mikedavis I saw on Matrix you were talking about putting it in core, maybe have a discussion in this issue?

matoous:
Not sure what the decision will be but I wanted to play around with this a bit and got this so far: https://github.com/matoous/helix/commit/b74b298fbdbee61e6ccbb6445189b087e4738c3d. Personally, I don't see issue with this being part of the core, it's closely coupled to the tree-sitter capabilities helix already provides and it isn't a large effort to have this/maintain it.

CBenoit:
@matoous Cool! I think you should open a PR with that so we can discuss in more details 🙂

matoous:
@CBenoit very rough PR: https://github.com/helix-editor/helix/pull/3944

yashpalgoyal1304:
Search term: sticky scroll for showing the current scope

ref: 
* https://code.visualstudio.com/updates/v1_71#_sticky-scroll
* https://code.visualstudio.com/updates/v1_70#_editor-sticky-scroll

luisdavim:
I'd love to see this added to `helix`, in vim I use https://github.com/nvim-treesitter/nvim-treesitter-context and it's really useful.

PR #3944, opened 2022-09-22
feat(ui): sticky context

matoous:
*Note:* This PR is ~rough~ proof of concept. I am completely fine if the final decision is _not_ to include this in the core. (although I believe it fits well with the rest of the tree-sitter support and should be part of the core).

Add option to render the context of current line if the context is outside the view. Parent Tree-sitter node(s) outside the view are displayed at the top of the view providing additional (hopefully helpful) context for the current cursor line. The nodes are filtered based on language-specific configuration to strike the right balance between helpful and too noisy.

Inspiration taken from [context.vim](https://github.com/wellle/context.vim) and a treesitter port of context.vim - https://github.com/romgrk/nvim-treesitter-context.

- [x] render only context outside the view
- [x] per-language configurable tree-sitter nodes to display
- [x] proper line numbers - maybe in a separate PR

The desired behavior should resemble this:

context-scroll

or this

demo

Fixes: https://github.com/helix-editor/helix/issues/396

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -112,11 +112,12 @@ The following statusline elements can be configured:
 
 ### `[editor.lsp]` Section
 
-| Key                   | Description                                                 | Default |
-| ---                   | -----------                                                 | ------- |
-| `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
-| `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
-| `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
+| Key                           | Description                                                 | Default |
+| ---                           | -----------                                                 | ------- |
+| `display-messages`            | Display LSP progress messages below statusline[^1]          | `false` |
+| `auto-signature-help`         | Enable automatic popup of signature help (parameter hints)  | `true`  |
+| `display-signature-help-docs` | Display docs under signature help popup                     | `true`  |
+| `context`                     | Display context of current line if outside the view         | `false` |
 
 [^1]: By default, a progress spinner is shown in the statusline beside the file path.
 
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -57,6 +57,7 @@ on unix operating systems.
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
+| `sticky-context` | Display context of current line if outside the view | `false` |
 
 ### `[editor.statusline]` Section
 
@@ -112,12 +113,11 @@ The following statusline elements can be configured:
 
 ### `[editor.lsp]` Section
 
-| Key                           | Description                                                 | Default |
-| ---                           | -----------                                                 | ------- |
-| `display-messages`            | Display LSP progress messages below statusline[^1]          | `false` |
-| `auto-signature-help`         | Enable automatic popup of signature help (parameter hints)  | `true`  |
-| `display-signature-help-docs` | Display docs under signature help popup                     | `true`  |
-| `context`                     | Display context of current line if outside the view         | `false` |
+| Key                   | Description                                                 | Default |
+| ---                   | -----------                                                 | ------- |
+| `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
+| `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
+| `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
 
 [^1]: By default, a progress spinner is shown in the statusline beside the file path.
 
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -150,6 +150,11 @@ impl EditorView {
         };
 
         Self::render_text_highlights(doc, view.offset, inner, surface, theme, highlights, &config);
+
+        if editor.config().lsp.context {
+            Self::render_context(editor, doc, view, surface, theme);
+        }
+
         Self::render_gutter(editor, doc, view, view.area, surface, theme, is_focused);
         Self::render_rulers(editor, doc, view, inner, surface, theme);
 
@@ -401,6 +406,84 @@ impl EditorView {
         }
 
         spans
+    }
+
+    pub fn render_context(
+        editor: &Editor,
+        doc: &Document,
+        view: &View,
+        surface: &mut Surface,
+        theme: &Theme,
+    ) {
+        if let Some(syntax) = doc.syntax() {
+            let text = doc.text().slice(..);
+            let tree = syntax.tree();
+            let byte = doc.selection(view.id).primary().cursor(text);
+
+            let mut parent = tree
+                .root_node()
+                .descendant_for_byte_range(byte, byte)
+                .and_then(|n| n.parent());
+
+            // context is list of numbers of lines that should be rendered in the LSP context
+            let mut context: Vec<usize> = Vec::new();
+
+            while let Some(curr) = parent {
+                let line = text.byte_to_line(curr.start_byte());
+
+                // if parent of previous node is still on the same line, use the parent node
+                if let Some(&prev_line) = context.last() {
+                    if prev_line == line {
+                        context.pop();
+                    }
+                }
+
+                context.push(line);
+                parent = curr.parent();
+            }
+
+            // we render from top most (last in the list)
+            context.reverse();
+
+            // TODO: this probably needs it's own style, although it seems to work well even with cursorline
+            let context_style = theme.get("ui.cursorline.primary");
+            let mut context_area = view.inner_area();
+            context_area.height = 1;
+
+            let mut last_line = 0;
+            for line_num in context {
+                if line_num > view.offset.row {
+                    continue;
+                }
+                surface.clear_with(context_area, context_style);
+
+                let offset = Position::new(line_num, 0);
+                let highlights = Self::doc_syntax_highlights(doc, offset, 1, theme);
+                Self::render_text_highlights(
+                    doc,
+                    offset,
+                    context_area,
+                    surface,
+                    theme,
+                    highlights,
+                    &editor.config(),
+                );
+
+                context_area.y += 1;
+                last_line = line_num;
+            }
+
+            surface.clear_with(context_area, context_style);
+            let text_style = theme.get("ui.text");
+            let last_line = text.get_line(last_line).unwrap();
+            let padding = last_line.chars().take_while(|&c| c == ' ').count();
+            surface.set_string(
+                context_area.x,
+                context_area.y,
+                format!("{}    context ----", " ".repeat(padding)),
+                text_style,
+            );
+        }
     }
 
     pub fn render_text_highlights<H: Iterator<Item = HighlightEvent>>(
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -416,83 +416,80 @@ impl EditorView {
         surface: &mut Surface,
         theme: &Theme,
     ) -> Option<Vec<usize>> {
-        if let Some(syntax) = doc.syntax() {
-            let tree = syntax.tree();
-            let text = doc.text().slice(..);
-            let cursor_byte = doc.selection(view.id).primary().cursor(text);
-            let context_nodes = doc
-                .language_config()
-                .and_then(|lc| lc.sticky_context_nodes.as_ref());
+        let syntax = doc.syntax()?;
+        let tree = syntax.tree();
+        let text = doc.text().slice(..);
+        let cursor_byte = doc.selection(view.id).primary().cursor(text);
+        let context_nodes = doc
+            .language_config()
+            .and_then(|lc| lc.sticky_context_nodes.as_ref());
 
-            let mut parent = tree
-                .root_node()
-                .descendant_for_byte_range(cursor_byte, cursor_byte)
-                .and_then(|n| n.parent());
+        let mut parent = tree
+            .root_node()
+            .descendant_for_byte_range(cursor_byte, cursor_byte)
+            .and_then(|n| n.parent());
 
-            // context is list of numbers of lines that should be rendered in the LSP context
-            let mut context: Vec<usize> = Vec::new();
+        // context is list of numbers of lines that should be rendered in the LSP context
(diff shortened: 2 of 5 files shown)


helix-view/src/editor.rs
```diff
@@ -232,6 +232,8 @@ pub struct LspConfig {
     pub auto_signature_help: bool,
     /// Display docs under signature help popup
     pub display_signature_help_docs: bool,
+    /// Display context of current cursor line if it is outside the view.
+    pub context: bool,
```
CBenoit, line 236: Maybe move to the `editor::Config` struct and go with a new category: `sticky_context: StickyContext`.

- enable: to enable sticky context feature (default: false)
- min_window_height: disable sticky context when the window is too small (default: ?)

helix-term/src/ui/editor.rs
```diff
@@ -380,6 +385,84 @@ impl EditorView {
...
+                // if parent of previous node is still on the same line, use the parent node
+                if let Some(&prev_line) = context.last() {
+                    if prev_line == line {
+                        context.pop();
+                    }
+                }
+
+                context.push(line);
+                parent = curr.parent();
+            }
```
CBenoit, line 420: We should also apply some filtering based on queries. Taking inspiration from https://github.com/nvim-treesitter/nvim-treesitter-context: `impl_item`, `struct`, `enum`, `method`, `function` for Rust (`class`, `function`, `'method` being good defaults I guess).
This should be configurable in `grammar` sections of the languages.toml file I think.

CBenoit (commented):
Thank you!

Maybe it's better to wait for an additional maintainer before putting additional work, because I'm unsure whether everyone agrees on including this. 

That being said, here is my initial feedback =)

s0LA1337:
I also want to throw in my tiny opinion i have:

the `--- context` is imo unnecessary, it's already signaled that the context is not the code itself by the changed background color.

Also this seems to be a bit untouched for a while i guess? Is some help needed to further implement this?

matoous:
@s0LA1337 I agree, that was just for testing. No help needed, I just need to find some time 😅 sorry about that; or feel free to pick it up if you feel like.

book/src/configuration.md
```diff
@@ -89,11 +89,12 @@ The following elements can be configured:
...
-| ---                   | -----------                                                 | ------- |
-| `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
-| `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
-| `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
+| Key                           | Description                                                 | Default |
+| ---                           | -----------                                                 | ------- |
+| `display-messages`            | Display LSP progress messages below statusline[^1]          | `false` |
+| `auto-signature-help`         | Enable automatic popup of signature help (parameter hints)  | `true`  |
+| `display-signature-help-docs` | Display docs under signature help popup                     | `true`  |
+| `context`                     | Display context of current line if outside the view         | `false` |
```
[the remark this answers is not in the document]
matoous: Moved to editor config. True, for some reason I completely confused LSP and Tree-sitter.

helix-view/src/editor.rs
matoous: Moved to `editor::Config` but left it as boolean config option for now. I am not sure if `min_window_height` is necessary as the context usually spans only a few lines. What do you thing?

helix-term/src/ui/editor.rs
matoous: Added option to configure `sticky-context-nodes` per language and added example config for rust. Once we are sure about the naming/approach I will try to add config for more languages.

matoous:
@s0LA1337 I had some time and moved the PR a bit. Feel free to test it out locally and I would be happy for any further feedback.

Otherwise, I have a few open questions:

- How do we want to handle line numbers? Looking at the neovim plugins linked above for inspiration: they don't bother with line numbers and leave the line numbers as they are (e.g. of the lines covered by the context). Do we want to instead show line number of the line that is part of the context? What is preferrable?
- https://github.com/wellle/context.vim can do some additional fancy stuff - such as showing other `case` statements in a `switch` (for rust that would be for example other `arms` in `match` statement). Is that something we want to pursuade or we keep it simple for now?
- How do we want to go about the coloring? For now I use `ui.cursorline.primary`. Do we want dedicated config? Or re-use some highlighting option?
- Last but not least, could someone asses or help me asses what the performance implications of the implementation are?

book/src/configuration.md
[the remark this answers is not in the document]
sudormrfbin: Seems like you missed the docs, it still has lsp.context. Also tried it out with a rust file and it feels quite nice to use, and with better queries should be even more useful :)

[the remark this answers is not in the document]
matoous: I really like it in markdown files:
```
# Title
## Subtitle
### Sub-subtitle

Paragraph I am writing.
```

sudormrfbin:
> * How do we want to handle line numbers? Looking at the neovim plugins linked above for inspiration: they don't bother with line numbers and leave the line numbers as they are (e.g. of the lines covered by the context). Do we want to instead show line number of the line that is part of the context? What is preferrable?

It would be cool to show the line numbers of the overlayed context lines instead of the ones under them, but I'd say that's best left to a separate PR.

> * [wellle/context.vim](https://github.com/wellle/context.vim) can do some additional fancy stuff - such as showing other `case` statements in a `switch` (for rust that would be for example other `arms` in `match` statement). Is that something we want to pursuade or we keep it simple for now?

I suppose cases like that would require writing actual tree-sitter queries instead of providing a simple list of scope names, and it does provide more granularity, so I suppose it would be nice to have. But I like how simple the current PR is and how it delivers most of the functionality already, so I would shelve that for a separate PR again :)

> * How do we want to go about the coloring? For now I use `ui.cursorline.primary`. Do we want dedicated config? Or re-use some highlighting option?

We definitely need separate theme keys, and it can manually fallback to another scope like `ui.cursorline` if needed.

> Last but not least, could someone asses or help me asses what the performance implications of the implementation are?

I'm not sure how expensive it is to traverse up the tree checking every parent, so someone with more experience in that area would have to give their input. The traversal is done on every render, so it might be costly. Maybe a cache could help? But only after checking actual perf numbers. Premature optimization is indeed the root of all evil :)

helix-view/src/editor.rs
CBenoit: I’m just thinking this can be useful when the editing some file in a small window. What do you think?

helix-term/src/ui/editor.rs
CBenoit: Sweet! Thank you!

s0LA1337:
> @s0LA1337 I had some time and moved the PR a bit. Feel free to test it out locally and I would be happy for any further feedback.
> 
> Otherwise, I have a few open questions:
> 
> * How do we want to handle line numbers? Looking at the neovim plugins linked above for inspiration: they don't bother with line numbers and leave the line numbers as they are (e.g. of the lines covered by the context). Do we want to instead show line number of the line that is part of the context? What is preferrable?
> * [wellle/context.vim](https://github.com/wellle/context.vim) can do some additional fancy stuff - such as showing other `case` statements in a `switch` (for rust that would be for example other `arms` in `match` statement). Is that something we want to pursuade or we keep it simple for now?
> * How do we want to go about the coloring? For now I use `ui.cursorline.primary`. Do we want dedicated config? Or re-use some highlighting option?
> * Last but not least, could someone asses or help me asses what the performance implications of the implementation are?

i actually opened a pr for your branch to include line numbers :)
Maybe i should resolve the conflicts :D

CBenoit:
This is looking pretty good!

Agreed with @sudormrfbin.

>     * Last but not least, could someone asses or help me asses what the performance implications of the implementation are?

Unless you had performance issues when testing in a reasonable sized file, I wouldn’t dig that just yet. The feature is optional and can be disabled at any time if it is an issue. Worst case, we can optimize that in a separate PR =)

matoous:
@CBenoit oh please go ahead with the PR, would love to take a look. And sorry, I amended a few more minor changes.

s0LA1337:
> It would be cool to show the line numbers of the overlayed context lines instead of the ones under them, but I'd say that's best left to a separate PR.

since i have literally worked on that already - i am not sure what you mean with that.

What i essentially did is: replacing the current line nrs in the gutter with the absolute ones (next to the sticky context) of the sticky context. pictures:

Relative line numbers:
image


Absolute line numbers:
image

s0LA1337:
also @matoous the PR is here: https://github.com/matoous/helix/pull/2

waiting for your approval, as when you merge it (it's on your fork) it will be part of this PR.

unless anyone has against it already being in this pr.

helix-view/src/editor.rs
sudormrfbin: Maybe a keybind to toggle it? That way it can be decided per case.

sudormrfbin:
@s0LA1337 we're on the same page about the line number display, your implementation was exactly how i imagined it to be.

s0LA1337:
> @s0LA1337 we're on the same page about the line number display, your implementation was exactly how i imagined it to be.

pr is open to fix the linting issue

helix-term/src/ui/editor.rs
```diff
@@ -413,6 +419,92 @@ impl EditorView {
...
+            context.reverse();
+
+            // TODO: this probably needs it's own style, although it seems to work well even with cursorline
+            let context_style = theme.get("ui.cursorline.primary");
+            let mut context_area = view.inner_area();
+            context_area.height = 1;
+
+            let mut line_numbers = Vec::new();
+            for line_num in context {
+                if line_num >= view.offset.row {
```
s0LA1337, line 472:
```suggestion
                if line_num >= view.offset.row || text.byte_to_line(cursor_byte) == line_num {
```

this is required because the sticky context will overshadow the cursor if you're on the same line with the cursor.

matoous: Have you tried with the latest changes? I ammended it to the latest comming but for me it works just with `line_num >= view.offset.row`, previously it was `line_num > view.offset.row`.

s0LA1337: > Have you tried with the latest changes? I ammended it to the latest commit but for me it works just with `line_num >= view.offset.row`, previously it was `line_num > view.offset.row`.

actually i haven't. time to try it out when i have time :)

s0LA1337: Update: yup seems good

CBenoit (commented):
This is looking pretty good to me code-wise!

@archseer I think this change is not too complex and can be part of helix’s built-ins. Are you okay with me merging this feature once last nits are addressed if any?

matoous:
@CBenoit (and maybe @archseer for further input): what I don't like right now is the way the context lines are passed from the `render_sticky_context` to `render_gutter`. Maybe you have better/cleaner idea of how to do it? Or maybe it's fine. Either way, I would keep the config simple `bool` for now and if this gets green light I would add configuration for more languages. There are some ideas for improvements that I would prefer to address separately.

s0LA1337:
I also have a proposal if it's:
1. feasible
2. not too much additional work

Couldn't we capture via tree-sitter all the arguments and have them stick, too?
I am not too sure whether this might end up being too complex.
Maybe someone which is more knowledgeable might be able to answer this..

CBenoit:
> Or maybe it's fine. Either way, I would keep the config simple `bool` for now

For reference: https://github.com/helix-editor/helix/pull/3944#discussion_r977971333

My problem with this is that it will then be a breaking change to modify this. But then, maybe we don’t want additional configuration and the additional keybinding only as suggested by @sudormrfbin is alright (I would like both)
Not my hill to die on though

helix-term/src/ui/editor.rs
```diff
@@ -413,6 +419,92 @@ impl EditorView {
...
+            let tree = syntax.tree();
+            let text = doc.text().slice(..);
+            let cursor_byte = doc.selection(view.id).primary().cursor(text);
+            let context_nodes = doc
+                .language_config()
+                .and_then(|lc| lc.sticky_context_nodes.as_ref());
+
+            let mut parent = tree
+                .root_node()
+                .descendant_for_byte_range(cursor_byte, cursor_byte)
```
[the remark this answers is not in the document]
matoous: That would complicate it a little bit as what can happen is that for example if the context would contain 3 nodes the 3 lines that we would display would cover more than just the top line of the viewport.

SoraTenshi:
I actually have noticed a small bug with the line numbers that i introduced.
=> In insert mode, when the linenumbers are set to relative, it changes from the absolute linenr to relatives, basically my logic inverts the linenumbers. I might want to fix that before the PR gets merged.

other than that: scopes on the first line seems to be sometimes a bit... wrong?
consider such a file:
```cpp
/**
  * i am a comment
  */
...
class Abc {
...
};
```

The comment will still be put on the sticky context, even though i completely left the scope.

yashpalgoyal1304:
hi! will there be option to choose the style between: `separator` (as shown in the context.vim's screen record), `colour`, `both`?

SoraTenshi:
Hi, i also just discovered that there is still a bug.
When the sticky context is overflowing, helix crashes.

An example can be found in this file: https://github.com/ziglang/zig/blob/master/src/main.zig
going past 1030 in this file will cause the following issue:
```
thread 'main' panicked at 'index out of bounds: the len is 4352 but the index is 4359', helix-tui/src/buffer.rs:513:33
```

stack trace:
```
thread 'main' panicked at 'index out of bounds: the len is 4352 but the index is 4359', helix-tui/src/buffer.rs:513:33
stack backtrace:
   0: rust_begin_unwind
   1: core::panicking::panic_fmt
   3: helix_tui::buffer::Buffer::clear_with
   4: helix_term::ui::editor::EditorView::render_sticky_context
   5: helix_term::ui::editor::EditorView::render_view
   6: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
   7: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   8: helix_term::application::Application::run::{{closure}}
  12: hx::main
```

yashpalgoyal1304:
One more point in favour of providing separator line i stumbled upon today:

The separator line will be extremely crucial choice for the default signature purple theme:
* as we don't have very many color choices remaining for the background in that theme
* so, having one more background color being stolen away by this sticky context color will make matters worst
* We are already struggling for finding a good color for primary vs secondary selection, 
(see https://github.com/helix-editor/helix/pull/5315)

SoraTenshi:
I took the time and integrated this pr's (partial) functionality to the new text api.

@matoous should i open up a pr for you to look over it, or open an entirely new pr for it?
You can find the changes here: https://github.com/SoraTenshi/helix/tree/sticky-context

matoous:
@SoraTenshi please feel free to open new PR, I am afraid I don't have time at the moment to pick this up.

SoraTenshi:
New pr created, i think this one can be closed.

PR #3951, opened 2022-09-23
Enable incremental mode for dev profile

gavincrawford:
Incremental mode should help with compile times for everyone working on Helix. AFAIK, incremental won't create any artifacts in release builds, but I'd rather not risk slowing final executables down.


archseer:
This is already the default value for dev: https://doc.rust-lang.org/cargo/reference/profiles.html#dev

gavincrawford:
Ah, okay, I wasn't aware.
Thanks!

Issue #2158, opened 2022-04-18
Ability to define second theme to sync with OS


dippa-1:
#### Describe your feature request
My OS is in light mode during day time and dark mode at night.
I like to code in the sun but in order to see anything I have to use a light theme.

My request is to define a light theme and a dark theme and the theme chosen depends on the OS's theme.
Low priority though.

thomas-profitt:
To work around this, you could have your configured theme be a symlink to another theme, and use your OS's task scheduling utility (e.g. systemd timers) to run a task every evening to replace the symlink with one pointing to the dark theme, and another to switch it to the light theme in the morning.

johalun:
+1 for this on macOS

archseer:
This is better suited to either solving with symlinks or via scripting once we support it. Querying OS themes would introduce a bunch of platform specific code.

EpocSquadron:
Is it possible to define a theme like base16 that inherits purely from the terminal's color scheme? I know for example wezterm supports switching theme based on light/dark switch in the OS (macos and Linux gnome). It would be cool if such a theme literally just output references to the terminal's foreground, background, so it would automatically change with the terminal.

sudormrfbin:
It's possible, simply specify the colors as `red`, `yellow`, etc without defining a palette or hex colors. There's a section on it in the theme docs.

EpocSquadron:
But is "background" and "foreground" a thing? Otherwise you end up specifying "black" as the background which gains nothing in responding to terminal theme change.

aral:
Just came over to open an issue on this also. New terminal apps like [Black Box](https://gitlab.gnome.org/raggesilver/blackbox) do an excellent job of integrating with the operating system’s (ok, desktop environment’s) colour scheme. It’s a joy to see my terminal go from light mode to dark mode as my whole system does but sad to see Helix Editor stick out like a sore thumb when it launches in Dracula in the middle of a light scheme. Vampires don’t like the light and all that… :)

archseer:
I'll reiterate: there's no built-in terminal way to detect this. We'd need to do a bunch of platform-specific hacks and probably link directly against a bunch of GTK-specific libraries to be able to read this on Linux+GTK.

See here for example: a nvim plugin that has to speak the dbus protocol (plus continually poll) to detect this: https://github.com/Sewdohe/Nvim-Adapt/blob/main/lua/nvim-adapt.lua

[...]

PR #3952, opened 2022-09-23
respond to SIGUSR1 by reloading config

krgn:
Small PR to add handling for the USR1 signal on unix machines, whereby the config is reloaded from disk. This is useful in settings where automation rewrites the config (or parts thereof, such as the theme config) and we want it reflected immediately instead of manually executing `config-reload` each time.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -28,6 +28,10 @@ hidden = false
 You may also specify a file to use for configuration with the `-c` or
 `--config` CLI argument: `hx -c path/to/custom-config.toml`.
 
+It is also possible to trigger configuration file reloading by sending the `USR1`
+signal to the helix process, e.g. via `pkill -USR1 hx`. This is only supported 
+on unix operating systems.
+
 ## Editor
 
 ### `[editor]` Section
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -224,8 +224,8 @@ impl Application {
         #[cfg(windows)]
         let signals = futures_util::stream::empty();
         #[cfg(not(windows))]
-        let signals =
-            Signals::new(&[signal::SIGTSTP, signal::SIGCONT]).context("build signal handler")?;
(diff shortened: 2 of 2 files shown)


dead10ck:
Could you elaborate a bit about your use case for this? Helix is an interactive editor, so I'm a bit confused about why `config-reload` is insufficient.

krgn:
Sure! My personal use case is this: i have a mechanism whereby I generate themes for various apps I use (e.g. i3/sway, wezterm, helix). While this sounds mundane (it is somewhat 😅) it’s very useful when light conditions change such that I need to pick a theme for all the text-based apps that is appropriate, I.e. a brighter theme when there is a lot of light and darker one at night. I also tend to have a lot of different helix sessions in disparate terminals open, so it quickly becomes tedious to have to `config-reload` in all of them manually, hence I came up with this patch to give me a way to make helix reload config from the outside.
Another solution could be to implement a watcher on `config.toml`, then one could simply `touch` the file and be done. The problem with this in my view is, if there is no such mechanism yet (I have not checked tbh), it will add a new dependency and significant amount of code, while this patch is just a few lines.

EpocSquadron:
This is a partial solve for #2158. I had just suggested in matrix a file watcher, but this is an admittedly more elegant solution.

dead10ck (approved):
Ok, this looks fine to me. I think @archseer mentioned doing this before somewhere too.

krgn:
Ah, one question though - should I document this somewhere? If so, where in the docs would be most appropriate? :pray:

the-mikedavis:
This would probably be a good spot in the config docs: https://github.com/helix-editor/helix/blob/master/book/src/configuration.md?plain=1#L29

dead10ck (approved):

0ihsan:
i have both tried `kill -USR1 $(pgrep hx)` and `pkill -USR1 hx`. on macos, these kill the proccess.

the-mikedavis:
This change is not released yet so you will need to build from master to use this feature. Both of those kill commands work for me on linux building from master.

PR #3953, opened 2022-09-24
Expose line numbers width in config

tonijarjour:
Here is an image of helix running with line numbers limited to 3 digits
image

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -168,6 +168,7 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    pub line_numbers_width: u16,
 }
 
 #[derive(Debug, Default, Clone, PartialEq, Serialize, Deserialize)]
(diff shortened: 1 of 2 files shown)


dead10ck (commented):
What happens when the number of lines is greater than the number of configured digits? Ideally the width should just be adjusted dynamically based on the file in view.

kirawi:
> What happens when the number of lines is greater than the number of configured digits? Ideally the width should just be adjusted dynamically based on the file in view.

https://github.com/helix-editor/helix/pull/3469

tonijarjour:
Where can I find total number of lines in the file? And would you want the width to be on a buffer by buffer basis (might cause jarring transitions when moving between buffers)? If not there would need to be a way to look through number of lines of all open files, and then set the width based on the longest.

kirawi:
#3469 implements that feature already. To get the number of lines in a file, you can do `doc.text().len_lines()`.

dgkf:
Hi, author of #3469 here

This PR looks great! I think my PR would dovetail nicely. The maximum and minimum width aren’t yet configurable in that PR, but it lends itself easily to exposing those as options. I think I even mention this as a “next step”. Take a look if you’d like, I think this could extend the behaviors nicely.

PR #3957, opened 2022-09-24
Always send a did save notification

Techatrix:
For some reason Helix only sends the [textDocument/didSave](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_didSave) notification if the [includeText](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#saveOptions) option is set to true.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -596,15 +596,12 @@ impl Client {
 
         let include_text = match &capabilities.text_document_sync {
             Some(lsp::TextDocumentSyncCapability::Options(lsp::TextDocumentSyncOptions {
-                save: Some(options),
+                save:
+                    Some(lsp::TextDocumentSyncSaveOptions::SaveOptions(lsp_types::SaveOptions {
+                        include_text,
+                    })),
                 ..
-            })) => match options {
-                lsp::TextDocumentSyncSaveOptions::SaveOptions(lsp_types::SaveOptions {
-                    include_text,
-                }) => include_text.unwrap_or(false),
-                _ => false,
-            },
-            // unsupported
+            })) => include_text.unwrap_or(false),
             _ => false,
         };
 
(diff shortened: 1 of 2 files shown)


the-mikedavis:
The `include_text` option can be false and the didSave will still be sent; it just won't include the full document text. If I'm reading the old code right, we don't send didSave when the server capability is `None` or `Some(TextDocumentSyncSaveOptions::Supported(false))` which I think is correct given the spec's wording, no?

Techatrix:
What is `TextDocumentSyncSaveOptions::Supported(false)` trying to represent? It doesn't directly map to anything in the specs if i'm not mistaken.
After re-reading the specs i see that only the `textDocument/didOpen`, `textDocument/didChange` and `textDocument/didClose` notifications are mandatory and `textDocument/didSave` therefore isn't. The server may send a `client/registerCapability` to indicate that they are interested in the `textDocument/didSave` notification but it doesn't seem to be supported. I always see 
`Language Server: Method not found client/registerCapability` in the logs. Helix's `initialize` request also doesn't specify [TextDocumentSyncClientCapabilities](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentSyncClientCapabilities) so its not even possible to find out if helix supports sending `textDocument/didSave`.

There is nothing stopping the client from always sending this notification which is also done by other clients (VS Code & Sublime Text).

the-mikedavis:
It's the line here: https://github.com/microsoft/language-server-protocol/blob/697910c7c3c559eafdd98fa8f3c001bee1ee7fdc/_specifications/lsp/3.17/textDocument/didSave.md?plain=1#L13 when the `boolean` is false

> Helix's initialize request also doesn't specify TextDocumentSyncClientCapabilities

That should definitely be added 👍 ([here](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/helix-lsp/src/client.rs#L285-L370))

Techatrix:
I now see what my mistake was. I was looking at the incomplete version of [TextDocumentSyncOptions](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocumentSyncOptions).

PR #3830, opened 2022-09-13
Tree-Sitter Powered Spell Checking

connorlay:
WIP

This is a proof-of-concept implementation of https://github.com/helix-editor/helix/discussions/3637 using `aspell` via the `ispell` crate.

PR #3965, opened 2022-09-25
Tree-Sitter Spellchecking with Hunspell FFI

connorlay:
Supersedes https://github.com/helix-editor/helix/pull/3830

This time a proof-of-concept using hunspell via Rust bindings to the C library.


runtime/queries/markdown/spellchecks.scm
```diff
@@ -0,0 +1 @@
+(paragraph) @spell
```
joukewitteveen, line 1: neovim added these `@spell` captures to their `highlights.scm` files rather than having separate files for them (here: `spellchecks.scm`). I am new to tree-sitter highlighting, but would expect that in principle such `highlights.scm` files could be shared between projects. The only constraint I can think of is that of the "highlight names" being used, but at least `@spell` is used in both projects :-).

the-mikedavis: That strategy wouldn't work well with the way helix interprets highlight queries: we prevent future stanzas from matching, so if you had a langauge with

```scm
(comment) @spell
(comment) @comment
```

You would only see the highlight for the spell capture.

Plus, highlight queries are run for every render and I expect that spell queries would not need to be run that often.

joukewitteveen:
Ouch, sorry to hear that! Any interesting insights in case someone finds time to continue this work?

ccojocar:
Thanks for putting this together. It would be great to get this work merged, especially because ltex-ls is relatively slow as an alternative solution.

connorlay:
I agree, but I don't have OSS cycles to work on this at the moment. It has fallen behind changes in `master` so starting a new PR once the dust settles around support for multiple language servers is probably the way to go.

Issue #3796, opened 2022-09-11
Add command to read file contents into current buffer (vim's `:r`)


pinpox:
I often use `:r filename.txt` to read a file into the current buffer on vim, but I could't find this feature in helix. There is `shell_insert_output` and `shell_append_output` which can be used together with `cat` to simulate this, but that has the drawback of not supporting things like tab-completion for paths and seems too complicated for such a basic feature.

My suggestion would be to either add a `file_insert/append_contents` or even better to use the existing file-picker dialog (`<space>-f`) and expand it's functionality with a shortcut to paste in the contents of the selected file, e.g.:
1. Open picker `<space>-f`
2. Type part of filename until it's selected
3. Press `<Ctrl>-o` or `<Ctrl>-O>` to paste contents of selected file at end/beginning of current buffer (or any other choosen key)

Furthermore, the `:r` (or `:read`) commands don't exist yet and would therefore not interfere with anything existing, I think it would be good idea to add those.

gdamore:
Yes please!

zephaniahong:
I will give this issue a shot 👍

PR #3966, opened 2022-09-25
feat: implement :read typed command

idursun:
This is a naive attempt at partially implementing #3796

Adds a typed command `:read` with alias `:r` (following vim) that reads the contents of a file into the current buffer.

At this stage, I just would like to validate my approach and obviously would love to finish it off after implementing some integration tests.

I also appreciate any pointers and guidance around reading the file contents into the buffer more efficiently. I am loading the file into a String for now.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1798,28 +1798,70 @@ fn read_file_info_buffer(
 
     let filename = args.get(0).unwrap();
     let path = PathBuf::from(filename.to_string());
-    if path.exists() {
-        let file = std::fs::File::open(path)?;
-        let mut reader = BufReader::new(file);
-
-        match helix_view::document::from_reader(&mut reader, Some(doc.encoding())) {
-            Ok((rope, _)) => {
-                let rope: String = rope.into();
-                let contents = Tendril::from(rope);
-                let selection = doc.selection(view.id);
-                let transaction = Transaction::insert(doc.text(), selection, contents);
-                doc.apply(&transaction, view.id);
-            }
-            Err(error) => cx
-                .editor
-                .set_error(format!("error reading file: {}", error)),
-        }
-    } else {
-        cx.editor
-            .set_error(format!("file doesn't exist: {}", filename));
+    if !path.exists() {
+        bail!("file doesn't exist: {}", filename);
     }
 
+    let file = std::fs::File::open(path).map_err(|err| anyhow!("error reading file {}", err))?;
+    let mut reader = BufReader::new(file);
+    let contents = from_reader(&mut reader, doc.encoding())
+        .map_err(|err| anyhow!("error reading file: {}", err))?;
+    let contents = Tendril::from(contents);
+    let selection = doc.selection(view.id);
+    let transaction = Transaction::insert(doc.text(), selection, contents);
+    doc.apply(&transaction, view.id);
+
     Ok(())
+}
+
+/// Stripped down version of [`helix_view::document::from_reader`] which is adapted to use encoding_rs::Decoder::read_to_string
+fn from_reader<R: std::io::Read + ?Sized>(
+    reader: &mut R,
+    encoding: &'static helix_core::encoding::Encoding,
+) -> anyhow::Result<String> {
+    let mut buf = [0u8; 8192];
+
+    let (mut decoder, mut slice, read) = {
+        let read = reader.read(&mut buf)?;
+        let decoder = encoding.new_decoder();
+        let slice = &buf[..read];
+        (decoder, slice, read)
+    };
+
+    let mut is_empty = read == 0;
+    let mut buf_str = String::with_capacity(buf.len());
+
+    loop {
+        let mut total_read = 0usize;
+
+        loop {
+            let (result, read, ..) =
+                decoder.decode_to_string(&slice[total_read..], &mut buf_str, is_empty);
+
+            total_read += read;
+
+            match result {
+                helix_core::encoding::CoderResult::InputEmpty => {
+                    debug_assert_eq!(slice.len(), total_read);
+                    break;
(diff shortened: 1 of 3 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1559,6 +1559,42 @@ fn run_shell_command(
...
+    }
+
+    let (view, doc) = current!(cx.editor);
+
+    ensure!(!args.is_empty(), "file name is expected");
+
+    let filename = args.get(0).unwrap();
+    let path = PathBuf::from(filename.to_string());
+    if path.exists() {
+        match std::fs::read_to_string(path) {
```
kirawi, line 1578: I would use `https://docs.rs/encoding_rs/latest/encoding_rs/struct.Decoder.html#method.decode_to_string` using the `doc.encoding` instead.

archseer: I'd extract parts of the from_reader code and share it.

idursun: @kirawi Thanks for the pointers! I am having a look.

> I'd extract parts of the from_reader code and share it.

I agree. It seems like `from_reader` is not a direct fit but close enough (i.e. if it worked with a `Tendril` instead of a `Rope`).

kirawi: Using `from_reader` directly is more expensive. I can try to work out something to make the function more generic, but for now I'd recommend copying that function and adapting it for a string.

pinpox:
Thanks for working on this! Would it be possible to integrate this with the file picker? Might be a separate PR though

idursun:
@kirawi I have copied and used a trimmed down version of `helix_view::document::from_reader` that is using `encoding_rs::Decoder::decode_to_string`. I have also added an integration test that is only testing the happy path.

I appreciate it if you can have a look and let me know if you suggest any changes.

helix-term/src/commands/typed.rs
```diff
@@ -1606,6 +1606,100 @@ fn run_shell_command(
...
+                }
+            }
+            Err(error) => cx
+                .editor
+                .set_error(format!("error reading file: {}", error)),
+        }
+    } else {
+        cx.editor
+            .set_error(format!("file doesn't exist: {}", filename));
+    }
```
the-mikedavis, line 1648: The statusline will show an error if you return the `Err` variant of an `anyhow::Result`, so you can remove all of these levels of nesting by returning early:

```rs
if !path.exists() {
    bail!("file doesn't exist: {}", filename);
}

let file = std::fs::File::open(path).map_err(|err| anyhow!("error reading file: {}", err))?;

// ...
```

helix-term/src/commands/typed.rs
```diff
@@ -1606,6 +1606,100 @@ fn run_shell_command(
...
+
+    Ok(())
+}
+
+/// Stripped down version of [`helix_view::document::from_reader`] which is adapted to use encoding_rs::Decoder::read_to_string
+fn from_reader<R: std::io::Read + ?Sized>(
```
the-mikedavis, line 1654: This duplicates a lot of code between this function and `document::from_reader`. Can you factor out the shared parts and share the code?

helix-term/tests/test/commands.rs
```diff
@@ -25,6 +25,42 @@ async fn test_write_quit_fail() -> anyhow::Result<()> {
...
+    let mut command = String::new();
+    let cmd = format!(":r {:?}<ret><esc>:w<ret>", output_file.path());
+    command.push_str(&cmd);
+
+    println!(">>File: {:?}", file.path());
+    println!(">>Output: {:?}", output_file.path());
+    println!(
+        ">>input: {:?}",
+        std::fs::canonicalize(output_file.path()).unwrap()
+    );
```
the-mikedavis, line 42: Are these for debugging? These lines can be removed

sudormrfbin:
How about `:read-file-into-buffer`? `:read` isn't very descriptive of what it reads, what it does to the read file, etc.

helix-term/src/commands/typed.rs
idursun: Thanks! For some reason, I assumed these methods need to return `Ok(())` even if they errored unless it's not a catastrophic error. I ended up with a lot of nesting. I have cleaned it all now.

helix-term/tests/test/commands.rs
idursun: Oops. Never happened.

helix-term/src/commands/typed.rs
idursun: I agree. I was following @kirawi's advice on this: https://github.com/helix-editor/helix/pull/3966#discussion_r979462663

I can try working something out at the expense of dragging this PR further, or that work can be the subject of another PR. I am easy either way. It's all up to the maintainers at this point.

idursun:
> How about `:read-file-into-buffer`? `:read` isn't very descriptive of what it reads, what it does to the read file, etc. We can still have the aliases for `:read` and `:r`.

@sudormrfbin I have named it `:read` just to be aligned with [vim's naming of the command](https://vimhelp.org/insert.txt.html#%3Aread).

the-mikedavis:
Vi `:read` is more general - you can read the output of shell commands like `:read !echo "foo"`

We already have `shell_insert_output` for that case so I think we should limit this command to reading from files and have a name that reflects that

helix-term/src/commands/typed.rs
the-mikedavis: This can use `read_to_string` from document now: https://github.com/helix-editor/helix/blob/636c91c76b2855a4ac58b3a030a9e45f88eb7502/helix-view/src/document.rs#L474-L477

pascalkuthe:
closing in favor of #10447

Issue #2564, opened 2022-05-25
Trim leading space in file picker search


aral:
Leading space in the file picker search can be trimmed to improve the experience.

e.g., your project structure:

```
some-file.js
```

Search box:

```
 some
```

Will not match _some-file.js_ due to the leading space (which could have been accidentally entered). Since leading spaces are not allowed in file names, they can be trimmed to provide a better experience in this edge case.

__Priority:__ low.

archseer:
The search is fuzzy though: ` test` could mean I'm trying to match `my test.txt`.

the-mikedavis:
Hmm yeah you may want to search for a space follow by a suffix. There is related discussion in https://github.com/helix-editor/helix/issues/2290 of handling spaces in the picker.

kirawi:
https://github.com/helix-editor/helix/pull/2585#issuecomment-1146783060
> Closing since this is intentional behaviour

PR #3969, opened 2022-09-25
Adjust fuzzy matching to be more inline with other editors

### LINKED ISSUE #2290: More robust fuzzy finding
Given a piece of text

```
packages/event_service/pubspec.yaml
```

You cannot match it with the query `event service pubspec`

It would be nice if fuzzy finding was robust enough to handle this type of use case

pascalkuthe:
Closes #2290 and #2564

This PR makes helix treat spaces as seperators instead of a normal character during matching.
This means that "foo bar" also matches "bar/foo" for example.
To match spaces as characeters they must now be escaped.
For example "foo\ bar" matches "foo bar" but not "bar foo" or "foobar".

This behaviour is inline with other IDEs (VScode and Intellij) and fuzzy finders (fzf and skim).
Therefore most people will expect helix to behave the same (I personally still trip up over this
even after using helix quite a while)

While a similar effect can be achieved by pressing C-space, this is quite inconvenient.
Especially in cases where you do not exactly know what you are looking for and often have to adjust the search.
Furthermore this approach also doesn't handle scenarios where spaces are typed by accident (like #2564).

### DIFF
--- a/helix-term/src/ui/fuzzy_match.rs
+++ b/helix-term/src/ui/fuzzy_match.rs
@@ -1,0 +1,74 @@
+use fuzzy_matcher::skim::SkimMatcherV2 as Matcher;
+use fuzzy_matcher::FuzzyMatcher;
+
+#[cfg(test)]
+mod test;
+
+pub struct FuzzyQuery {
+    queries: Vec<String>,
+}
+
+impl FuzzyQuery {
+    pub fn new(query: &str) -> FuzzyQuery {
+        let mut saw_backslash = false;
+        let queries = query
+            .split(|c| {
+                saw_backslash = match c {
+                    ' ' if !saw_backslash => return true,
+                    '\\' => true,
+                    _ => false,
+                };
+                false
+            })
+            .filter_map(|query| {
+                if query.is_empty() {
+                    None
+                } else {
+                    Some(query.replace("\\ ", " "))
+                }
+            })
+            .collect();
+        FuzzyQuery { queries }
+    }
+
+    pub fn fuzzy_match(&self, item: &str, matcher: &Matcher) -> Option<i64> {
+        // use the rank of the first query for the rank, because merging ranks is not really possible
+        // this behaviour matches fzf and skim
+        let score = matcher.fuzzy_match(item, self.queries.get(0)?)?;
+        if self
+            .queries
+            .iter()
+            .any(|query| matcher.fuzzy_match(item, query).is_none())
+        {
+            return None;
+        }
+        Some(score)
+    }
+
+    pub fn fuzzy_indicies(&self, item: &str, matcher: &Matcher) -> Option<(i64, Vec<usize>)> {
+        if self.queries.len() == 1 {
+            return matcher.fuzzy_indices(item, &self.queries[0]);
+        }
+
+        // use the rank of the first query for the rank, because merging ranks is not really possible
+        // this behaviour matches fzf and skim
+        let (score, mut indicies) = matcher.fuzzy_indices(item, self.queries.get(0)?)?;
+
+        // fast path for the common case of not using a space
+        // during matching this branch should be free thanks to branch prediction
+        if self.queries.len() == 1 {
+            return Some((score, indicies));
+        }
+
+        for query in &self.queries[1..] {
+            let (_, matched_indicies) = matcher.fuzzy_indices(item, query)?;
+            indicies.extend_from_slice(&matched_indicies);
+        }
+
+        // deadup and remove duplicate matches
+        indicies.sort_unstable();
+        indicies.dedup();
+
+        Some((score, indicies))
+    }
+}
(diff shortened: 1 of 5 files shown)


helix-term/src/ui/fuzzy_match.rs
```diff
@@ -0,0 +1,74 @@
...
+            .filter_map(|query| {
+                if query.is_empty() {
+                    None
+                } else {
+                    Some(query.replace("\\ ", " "))
+                }
+            })
+            .collect();
+        FuzzyQuery { queries }
+    }
```
AlexanderBrevig, line 32:
```suggestion
    pub fn new(query: &str) -> FuzzyQuery {
        const SIGNISPACE_PLACEHOLDER: &str = "$$$$";
        let queries = query.replace("\\ ", SIGNISPACE_PLACEHOLDER)
            .split_whitespace()
            .map(|query| query.replace(SIGNISPACE_PLACEHOLDER, " "))
            .collect();
        FuzzyQuery { queries }
    }
```

AlexanderBrevig (changes requested):
I had to read the `FuzzyQuery::new` a couple of times, so I tried quickly in rust playground. No guarantees, but a lot easier to read IMHO.

Works like a charm! Looking forward to this :)

helix-term/src/ui/fuzzy_match.rs
kirawi: That looks hacky too me. It's possible that someone would want to search for `$$$$`.

AlexanderBrevig: That's why I made it a const for easy replacement. I agree that `$$$$` is a bad choice, but maybe every choice is a bad one - and an FSM is better.

I did find the original code easier to read if the check against `\\` is first like:
```rust
                saw_backslash = match c {
                    '\\' => true,
                    ' ' if !saw_backslash => return true,
                    _ => false,
                };```
It kind of tells the reader what's going on in order of observed tokens.

AlexanderBrevig: I marked the suggestion as resolved, I agree that it was a bit too hacky.

pascalkuthe: I have written a ton of parsers/lexers recently (my job is working on a compiler) so this code was very natural to me but if you find it more readable I can definitely change the order of the match.

Your suggestion is actually somewhat similar to what`skim` uses internally actually.
However they use the null terminator '\0' as a placeholder.
The reasons they do it that way is because they use the `regex` crate instead of rolling their own implementation and regex can not support this kind of escape sequence.
They use regex because they actually support many more patterns in their matcher which I did not include in this PR because they are not the same across various implementations and I wanted to keep the PR uncontroversial.

In my experience working with compilers using a custom lexer often ends up being simpler (regex has a nasty habit of growing into monstrosities) and in many cases also more performant.

Also using `split_whitespace` here is not ideal because it matches any unicode white space (for example newlines) which is a behavior other tools/editors do not share and therefore do not want to introduce here.

AlexanderBrevig: I think my confusion can be explained by the late hours and this was the first time I saw an if condition as part of the match case. 

What is the intended result of input `one two\ three four` ? I would expect it to break into `["one", "two three", "four"]` but I think it will be `["one", "two three four"]` ?

pascalkuthe: No this works just fine, I tested in on the [playground](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ed75f071aa9ef2de64a2e3eb84a14a9d)

I will add a comment explaining the parser a bit more and a few more testcases but this is a very basic FSM so I am extremely confident it works as intended.

AlexanderBrevig: seems like the early hours got me too, haha - the `_` fallback will of course toggle it back. Thanks for putting up with me.

BTW; no need to add superfluous comments on my account. Now that I look at it again with the `if` case match part of my vocabulary for Rust it is concise and readable.

AlexanderBrevig (approved):

the-mikedavis:
With this change, `C-space` doesn't really have a use-case anymore, right? Should it and the associated code be removed?

pascalkuthe:
> With this change, `C-space` doesn't really have a use-case anymore, right? Should it and the associated code be removed

C-space ranks things a bit differently but the difference is probably small enough that it can be removed

pascalkuthe:
Sorry forgot about this.
I have removed `C-space` filtering from the picker (and rebased on master)

pascalkuthe:
Not really sure what is going on with the pipeline here. The failures look unrelated to the actual changes

sudormrfbin:
@pascalkuthe Re-ran the failed jobs, seems like a spurious failure.

archseer:
Thanks! Looks good to me.

PR #3970, opened 2022-09-25
inject language based on file extension & shebang

nrdxp:
Nodes can now be captured with "injection.filename". If this capture contains a valid file extension known to Helix, then the content will be highlighted as that language. In addition, a new "injection.shebang" is added which injects based on a valid shebang.

For consistency, both implementations use the same logic as used to detect the document filetype as a whole.

~~I'm not sure if the following is the most efficient way to capture the built in configuration, or if it can be passed in from somewhere else:
https://github.com/nrdxp/helix/blob/7f61de137f07cf5897d5a3240b1ece98febc0eb0/helix-core/src/syntax.rs#L1836~~

~~Also, it would probably be nice to extend this to user defined languages as well.~~ 65bc9da1f4b0b67dc3189c4a21e5cd551a733b07

Also includes an additional query for Nix so one can test this behavior themselves. A few screens with that query:
image
image

This PR also includes a new document briefly describing language injeciton.

### DIFF
--- a/book/src/guides/injection.md
+++ b/book/src/guides/injection.md
@@ -1,0 +1,57 @@
+# Adding Injection Queries
+
+Writing language injection queries allows one to highlight a specific node as a different language.
+In addition to the [standard](upstream-docs) language injection options used by tree-sitter, there
+are a few Helix specific extensions that allow for more control.
+
+And example of a simple query that would highlight all strings as bash in Nix:
+```scm
+((string_expression (string_fragment) @injection.content)
+  (#set! injection.language "bash"))
+```
+
+## Capture Types
+
+- `@injection.language` (standard):
+The captured node may contain the language name used to highlight the node captured by
+`@injection.content`.
+
+- `@injection.content` (standard):
+Marks the content to be highlighted as the language captured with `@injection.language` _et al_.
+
+- `@injection.filename` (extension):
+The captured node may contain a filename with a file-extension known to Helix,
+highlighting `@injection.content` as that language. This uses the language extensions defined in
+both the default languages.toml distributed with Helix, as well as user defined languages.
+
+- `@injection.shebang` (extension):
+The captured node may contain a shebang used to choose a language to highlight as. This also uses
+the shebangs defined in the default and user `languages.toml`.
+
+## Settings
+
+- `injection.combined` (standard):
+Indicates that all the matching nodes in the tree should have their content parsed as one
+nested document.
+
+- `injection.language` (standard):
+Forces the captured content to be highlighted as the given language
+
+- `injection.include-children` (standard):
+Indicates that the content node’s entire text should be re-parsed, including the text of its child
+nodes. By default, child nodes’ text will be excluded from the injected document.
+
+- `injection.include-unnamed-children` (extension):
+Same as `injection.include-children` but only for unnamed child nodes.
+
+## Predicates
+
+- `#eq?` (standard):
+The first argument (a capture) must be equal to the second argument
+(a capture or a string).
+
+- `#match?` (standard):
+The first argument (a capture) must match the regex given in the
+second argument (a string).
+
+[upstream-docs]: http://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -662,9 +662,8 @@ impl Loader {
 
     pub fn language_config_for_shebang(&self, source: &Rope) -> Option<Arc<LanguageConfiguration>> {
         let line = Cow::from(source.line(0));
-        static SHEBANG_REGEX: Lazy<Regex> = Lazy::new(|| {
-            Regex::new(r"^#!\s*(?:\S*[/\\](?:env\s+(?:\-\S+\s+)*)?)?([^\s\.\d]+)").unwrap()
-        });
+        static SHEBANG_REGEX: Lazy<Regex> =
+            Lazy::new(|| Regex::new(&["^", SHEBANG].concat()).unwrap());
         let configuration_id = SHEBANG_REGEX
             .captures(&line)
             .and_then(|cap| self.language_config_ids_by_shebang.get(&cap[1]));
@@ -713,6 +712,7 @@ impl Loader {
         match capture {
             InjectionCapture::Name(string) => self.language_config_for_name(string),
             InjectionCapture::File(file) => self.language_config_for_file_name(file),
+            InjectionCapture::Shebang(shebang) => self.language_config_for_language_id(shebang),
         }
     }
 
@@ -1403,6 +1403,7 @@ pub struct HighlightConfiguration {
     injection_content_capture_index: Option<u32>,
     injection_language_capture_index: Option<u32>,
     injection_filename_capture_index: Option<u32>,
+    injection_shebang_capture_index: Option<u32>,
     local_scope_capture_index: Option<u32>,
     local_def_capture_index: Option<u32>,
     local_def_value_capture_index: Option<u32>,
@@ -1547,6 +1548,7 @@ impl HighlightConfiguration {
         let mut injection_content_capture_index = None;
         let mut injection_language_capture_index = None;
         let mut injection_filename_capture_index = None;
+        let mut injection_shebang_capture_index = None;
         let mut local_def_capture_index = None;
         let mut local_def_value_capture_index = None;
         let mut local_ref_capture_index = None;
@@ -1568,6 +1570,7 @@ impl HighlightConfiguration {
                 "injection.content" => injection_content_capture_index = i,
                 "injection.language" => injection_language_capture_index = i,
                 "injection.filename" => injection_filename_capture_index = i,
+                "injection.shebang" => injection_shebang_capture_index = i,
                 _ => {}
             }
         }
@@ -1584,6 +1587,7 @@ impl HighlightConfiguration {
             injection_content_capture_index,
             injection_language_capture_index,
             injection_filename_capture_index,
+            injection_shebang_capture_index,
             local_scope_capture_index,
             local_def_capture_index,
             local_def_value_capture_index,
@@ -2057,7 +2061,10 @@ impl<'a> Iterator for HighlightIter<'a> {
 pub enum InjectionCapture<'a> {
     Name(Cow<'a, str>),
     File(Cow<'a, Path>),
+    Shebang(String),
 }
+
+static SHEBANG: &str = r"#!\s*(?:\S*[/\\](?:env\s+(?:\-\S+\s+)*)?)?([^\s\.\d]+)";
 
 fn injection_for_match<'a>(
     config: &HighlightConfiguration,
@@ -2072,6 +2079,7 @@ fn injection_for_match<'a>(
     let content_capture_index = config.injection_content_capture_index;
     let language_capture_index = config.injection_language_capture_index;
     let language_filename_index = config.injection_filename_capture_index;
+    let language_shebang_index = config.injection_shebang_capture_index;
 
     let mut injection_capture = None;
     let mut content_node = None;
@@ -2087,6 +2095,27 @@ fn injection_for_match<'a>(
             let path = Path::new(name.as_ref()).to_path_buf();
 
             injection_capture = Some(InjectionCapture::File(path.into()));
+        } else if index == language_shebang_index {
+            let node_slice = source.slice(capture.node.byte_range());
+
+            // some languages allow space and newlines before the actual string content
+            // so a shebang could be on either the first or second line
+            let lines = if node_slice.len_lines() > 1 {
+                node_slice.slice(
+                    // start of first line
+                    node_slice.line_to_byte(0)
+                    // end of second line
+                    ..node_slice.line_to_byte(2) - 1,
+                )
+            } else {
+                node_slice.line(0)
+            };
+
+            static SHEBANG_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(SHEBANG).unwrap());
+
+            injection_capture = SHEBANG_REGEX
+                .captures(lines.as_str().unwrap_or_default())
+                .map(|cap| InjectionCapture::Shebang(cap[1].to_owned()));
         }
     }
 
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -686,15 +686,14 @@ impl Loader {
             .cloned()
     }
 
-    pub fn language_configuration_for_injection_string(
-        &self,
-        string: &str,
-    ) -> Option<Arc<LanguageConfiguration>> {
+    /// Unlike language_config_for_language_id, which only returns Some for an exact id, this
+    /// function will perform a regex match on the given string to find the closest language match.
+    pub fn language_config_for_name(&self, name: &str) -> Option<Arc<LanguageConfiguration>> {
         let mut best_match_length = 0;
         let mut best_match_position = None;
         for (i, configuration) in self.language_configs.iter().enumerate() {
             if let Some(injection_regex) = &configuration.injection_regex {
-                if let Some(mat) = injection_regex.find(string) {
+                if let Some(mat) = injection_regex.find(name) {
                     let length = mat.end() - mat.start();
                     if length > best_match_length {
                         best_match_position = Some(i);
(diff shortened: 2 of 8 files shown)


the-mikedavis:
I think I would prefer `injection.filename` for this since the language is computed by the extension but the captured value is the full filename

nrdxp:
> I think I would prefer `injection.filename` for this since the language is computed by the extension but the captured value is the full filename

good point, I'll push a fix

nrdxp:
Updated to "injection.filename" and also address the clippy lints

helix-core/src/syntax.rs
```diff
@@ -1827,6 +1833,25 @@ fn injection_for_match<'a>(
...
+                let language = languages.iter().find(|language| {
+                    language
+                        .file_types
+                        .iter()
+                        .any(|e| extension.to_str() == Some(e))
+                });
+                if let Some(language) = language {
+                    language_name = Some(Cow::from(language.language_id.clone()));
+                }
+            }
```
the-mikedavis, line 1854: I think this differs slightly from the file-type detection we use for determining the language to use. Can we re-use that code here without too much headache?

In particular there's a case where `Makefile` or `Dockerfile` won't be highlighted because they don't have an extension while they are correctly picked up by the `file-types` configuration ([this](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/languages.toml#L784) and [this](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/languages.toml#L960)).

nrdxp: Hey thanks, the code is a lot simpler to follow now as well, and I didn't realize previously that the syntax struct already had a reference to the loader :facepalm:  so I fixed that as well

nrdxp:
There appears to be one little bug I would like to get to the bottom of before merging. Mainly, if the capture happens more than once in a file, then the last value or language overrides all the other string contents highlighting. Not sure why this isn't an issue with `injection.language` since I didn't set any of the highlighting logic.

To illustrate:
```
{
  f = writeText "foo.sh"  ''
    echo $bar
  '';
  g = writeText "foo.py" ''
    def foo():
        print("hello")
  '';
  }
  ```
  
  Would highlight both strings as python for some reason. Not sure if this is just an issue with the query or the rust code yet.

**Comment by the-mikedavis** [comment]
`injection.combined` is for combining all nodes captured with `injection.content` together. There isn't currently any scoping for it - it applies for the whole document. I don't think the comment highlighting or this should be using it now that I think about it. Does something break if we remove it?

I think `injection.language` should be separate from this

**Comment by the-mikedavis** [comment]
> the downside is that highlighting after interpolation isn't correct either

Ah sorry, reading is hard 😅

To cover this you most likely want to use `injection.include-children` instead of `injection.combined`

**Comment by nrdxp** [comment]
@the-mikedavis, thanks for the feedback, I'll try to fix the queries and then it should be ready. Also, fyi I changed the implementation just a little bit. I think it's definitely better. It should also be more efficient since we don't have to traverse the languages once to get the name and then again to get the config.

Instead I introduced a new enum `LanguageConfig`, which can either be a `str` or a `Path`. This should make it easier to extend in the future if we wanted too.

**Comment by nrdxp** [comment]
My last refactor made it fairly trivial to add another commit to highlight based on a captured shebang. In addition, I wrote up a brief doc explaining language injection queries.

File: book/src/guides/injection.md [hunk 2]
```diff
@@ -0,0 +1,56 @@
...
+
+## Predicates
+
+- `#eq?`:
+The first argument (a capture) must be equal to the second argument
+(a capture or a string).
+
+- `#match?`:
+The first argument (a capture) must match the regex given in the
+second argument (a string).
```
the-mikedavis, line 54: These two are standard as well

book/src/guides/injection.md
```diff
@@ -0,0 +1,56 @@
...
+
+Writing language injection queries allows one to highlight a specific node as a different language.
+In addition to the [standard](upstream-docs) language injection options used by tree-sitter, there
+are a few Helix specific extensions that allow for more control.
+
+And example of a simple query that would highligh all strings as bash:
+```scm
+((string_expression (string_fragment) @injection.content)
+  (#set! injection.language "bash"))
+```
```
the-mikedavis, line 11: It's probably worth noting here that the language is Nix for this query. Other languages won't have the same node names so this would make an invalid query

helix-core/src/syntax.rs
```diff
@@ -1837,8 +1892,10 @@ fn injection_for_match<'a>(
             // captured node, it can also be hard-coded via a `#set!` predicate
             // that sets the injection.language key.
             "injection.language" => {
-                if language_name.is_none() {
-                    language_name = prop.value.as_ref().map(|s| s.as_ref().into())
+                if language_capture.is_none() {
+                    if let Some(lang) = prop.value.as_ref().map(|s| s.as_ref()) {
+                        language_capture = Some(InjectionCapture::Name(lang.into()));
+                    }
```
the-mikedavis, line 1898:
```suggestion
                    language_capture = prop
                        .value
                        .as_ref()
                        .map(|s| InjectionCapture::Name(s.as_ref().into()));
```

book/src/guides/injection.md
```diff
@@ -0,0 +1,56 @@
...
+nested document.
+
+- `injection.language` (standard):
+Forces the captured content to be highlighted as the given language
+
+- `injection.include-children`  (standard):
+Indicates that the content node’s entire text should be re-parsed, including the text of its child
+nodes. By default, child nodes’ text will be excluded from the injected document.
+
+- `injection.include-unamed-children`  (extension):
```
the-mikedavis, line 43:
```suggestion
- `injection.include-unnamed-children`  (extension):
```

helix-core/src/syntax.rs
```diff
@@ -1809,24 +1826,62 @@ impl<'a> Iterator for HighlightIter<'a> {
...
         } else if index == content_capture_index {
             content_node = Some(capture.node);
+        } else if index == language_filename_index {
+            let name = byte_range_to_str(capture.node.byte_range(), source);
+            let path = Path::new(name.as_ref()).to_path_buf();
+
+            language_capture = Some(InjectionCapture::File(path.into()));
+        } else if index == language_shebang_index {
+            // assume that the content with the shebang is also the content to highlight
+            content_node = Some(capture.node);
```
the-mikedavis, line 1869: A single node can have multiple captures

```scm
(keyword) @injection.shebang @injection.content
```

so assuming the content node is not necessary here

nrdxp: Indeed, this is how I first had it, but I just figured, when would we ever want to capture a shebang but not highlight the same node? I don't mind being more explicit though.

nrdxp:
@the-mikedavis, I address your comments and cleaned up the commit history.

helix-core/src/syntax.rs
```diff
@@ -519,25 +518,33 @@ impl Loader {
 
     pub fn language_configuration_for_injection_string(
```
the-mikedavis, line 519: This function can be cleaned up a bit if we move the code for `InjectionCapture::Name` to its own function `language_config_for_name`

nrdxp: That's right. I meant to do that but I guess I overlooked it. Just force pushed a fix.

nrdxp:
One last force push to make shebang injection more efficient. Instead of building an interator over the lines, we take a slice to avoid an allocation, and in additional, we extract the language name capture on our first regex, so that we don't have to do a second regex match, but instead we can just call `language_config_for_language_id` on the result.

helix-core/src/syntax.rs
```diff
@@ -1809,24 +1824,65 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+            let lines = if source.len_lines() > line_idx + 2 {
+                source.slice(
+                    // start of first line
+                    source.line_to_char(line_idx)
+                    // end of second line
+                    ..source.line_to_char(line_idx + 2) - 1,
+                )
+            } else {
+                source.line(line_idx)
+            };
```
the-mikedavis, line 1879: Is it possible for this to look outside of the text of captured node? We might want to slice `source` to the byte range of `capture.node` and work from there

helix-core/src/syntax.rs
```diff
@@ -1809,24 +1824,65 @@ impl<'a> Iterator for HighlightIter<'a> {
     }
 }
 
+#[derive(Debug, Clone)]
+pub enum InjectionCapture<'a> {
+    Name(Cow<'a, str>),
+    File(Cow<'a, Path>),
```
the-mikedavis, line 1830: What do you think about "LanguageName" and "Filename" for these two variants? I think "File" has some connotation that it's a file handle. Or maybe "Path"?

nrdxp: We could maybe also rename the enum to `InjectionLanguage`? Maybe that's more descriptive in general.

helix-core/src/syntax.rs
nrdxp: I fixed this in another [commit](https://github.com/helix-editor/helix/pull/3970/commits/1541bd683dce26fef0d4b5dc66e160a3330181cf) that refactors the shebang function into a trait method to make the flow easier to follow.

helix-core/src/syntax.rs
```diff
@@ -1809,56 +1887,44 @@ impl<'a> Iterator for HighlightIter<'a> {
...
+            node_slice.line(0)
+        };
+
+        static SHEBANG_REGEX: Lazy<Regex> = Lazy::new(|| Regex::new(SHEBANG).unwrap());
 
-    (language_name, content_node, included_children)
+        SHEBANG_REGEX
+            .captures(lines.as_str().unwrap_or_default())
+            .map(|cap| InjectionCapture::Shebang(cap[1].to_owned()))
+    }
```
the-mikedavis, line 1923: I don't think a trait and implementation are the right tool here. If anything I think this can be done in an ordinary function that takes a node and the rope as input

nrdxp: How about an associated function on `InjectionCapture`? Just trying to avoid adding special purpose functions to the top-level scope.

the-mikedavis: I don't think this function is so specific that it needs to be hidden. But the body is also pretty small so I think it would be ok to inline it into `injection_pair`.

As associated item feels awkward because it doesn't end up using `self` which I think is fine for functions like `new` or `default` but feels out-of-place in this usage

helix-core/src/syntax.rs
the-mikedavis: What do you think about `InjectionLanguageMarker`?

helix-core/src/syntax.rs
nrdxp: I opted to just inline it in that case

helix-core/src/syntax.rs
```diff
@@ -1400,6 +1410,95 @@ impl HighlightConfiguration {
...
+                let node_slice = source.slice(capture.node.byte_range());
+
+                // some languages allow space and newlines before the actual string content
+                // so a shebang could be on either the first or second line
+                let lines = if node_slice.len_lines() > 1 {
+                    node_slice.slice(
+                        // start of first line
+                        node_slice.line_to_byte(0)
+                // end of second line
+                ..node_slice.line_to_byte(2) - 1,
```
the-mikedavis, line 1441: The formatting looks a bit weird here - did `cargo fmt` do that? It could be worth separating out

```rust
let start = node_slice.line_to_byte(0);
let end = node_slice.line_to_byte(2) - 1;
node_slice.slice(start..end)
```

to avoid it. Or maybe moving the comments to outside of the call to `slice` might do it. I think that may be a bug upstream in `cargo fmt`

the-mikedavis (dismissed):
Other than that odd formatting spot, I think this looks great!

helix-core/src/syntax.rs
nrdxp: I went with:
```rust
node_slice.slice({
    let start = node_slice.line_to_byte(0);
    let end = node_slice.line_to_byte(2) - 1;
    start..end
})
```
I do think it is a bug, as the formatter basically just ingored those lines. I could indent or dedent as much as I want, but the formatter wouldn't move them.

nrdxp:
Anything I can do (besides fixing merge conflicts) to get this merged, or is it just not gonna happen?

languages.toml
```diff
@@ -481,7 +481,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "nix"
-source = { git = "https://github.com/cstrahan/tree-sitter-nix", rev = "6b71a810c0acd49b980c50fc79092561f7cee307" }
+source = { git = "https://github.com/cstrahan/tree-sitter-nix", rev = "1b69cf1fa92366eefbe6863c184e5d2ece5f187d" }
```
pascalkuthe, line 484: updating the grammar seems orthogonal to this PR. Considering there is a conflict here this has been bumped on master already anyway.

pascalkuthe (changes requested):
I think we will need to wait for 23.3.1 to release before merging this since it's not quite a trivial PR but once my comments are addressed this LGTM

languages.toml
nrdxp: Actually there was a PR that fixed an issue with Nix highlighting included here. After looking at the pinned commit in master, for some reason it was pinned before my PR was merged, so a bump is still in order or one of the queries won't even function.

nrdxp:
Thanks for the thorough review! I addressed all of your comments, and hopefully it is ready now :crossed_fingers:

pascalkuthe (approved):
LGTM now

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+The captured node may contain the language name used to highlight the node captured by
+`@injection.content`.
+
+- `@injection.content` (standard):
+Marks the content to be highlighted as the language captured with `@injection.language` _et al_.
+
+- `@injection.filename` (extension):
+The captured node may contain a filename with a file-extension known to Helix,
+highlighting `@injection.content` as that language. This uses the language extensions defined in
+both the default languages.toml destributed with Helix, as well as user defined languages.
```
David-Else, line 25: destributed = distributed

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+
+- `@injection.shebang` (extension):
+The captured node may contain a shebang used to choose a language to highlight as. This also uses
+the shebangs defined in the default and user languages.toml.
```
David-Else, line 29: languages.toml = `languages.toml`

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Settings
+
+- `injection.combined` (standard):
+Indicates that all of the matching nodes in the tree should have their content parsed as one
```
David-Else, line 34: all of the = all the

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+nested document.
+
+- `injection.language` (standard):
+Forces the captured content to be highlighted as the given language
+
+- `injection.include-children`  (standard):
```
David-Else, line 40: unwanted space before (standard)

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+Indicates that the content node’s entire text should be re-parsed, including the text of its child
+nodes. By default, child nodes’ text will be excluded from the injected document.
+
+- `injection.include-unnamed-children`  (extension):
```
David-Else, line 44: unwanted space before (extension)

book/src/guides/injection.md
```diff
@@ -0,0 +1,57 @@
...
+Same as `injection.include-children` but only for unamed child nodes.
```
David-Else, line 45: unamed = unnamed

David-Else (commented):
There are few spelling and typing errors I noticed.

nrdxp:
Thank you @David-Else. Looks like we really need a spellchecker for Helix as well :sweat_smile:

David-Else:
> Thank you @David-Else. Looks like we really need a spellchecker for Helix as well sweat_smile

You can use https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#ltex-ls , for Markdown, it is excellent!

pascalkuthe (approved):

PR #3971, opened 2022-09-26
themes: Extend snazzy

nrdxp:
In Helix, we are not restricted to just 8 colors like legacy terminal applications, and we also have _a lot_ of different highlight categories, especially in languages with a lot of features, like Rust.

It was difficult to modify the original snazzy in a way that made the highlighting of all these various captures meaningful, so I decided to just extend the palette with colors that I felt went well with the main ones.

I've been using it already for several weeks, and just wanted to make sure I was happy with it before submitting a PR. I think I've got it to a place where it is worth sharing. If there are no major objections, we might even be able to just update snazzy.toml in place, but I thought adding a variant was less offensive to start.

Snazzy extended:
image

Snazzy:
image


AlexanderBrevig (changes requested):
`cargo xtask themelint snazzy_extended` with (#3983):
```
snazzy_extended: `ui.statusline.normal` and `ui.statusline.insert` cannot be equal
snazzy_extended: `ui.statusline.normal` and `ui.statusline.select` cannot be equal
snazzy_extended: missing `fg` or `bg` for `ui.virtual.indent-guide`
snazzy_extended: missing `fg` or `bg` for `markup.link.label`
```

Please add a screenshot to your description, maybe comparing it to snazzy?

nrdxp:
@AlexanderBrevig I addressed the missing rules, but I didn't notice a difference in my build of helix, perhaps the branch is too old? Also, I added some screens.

AlexanderBrevig:
> @AlexanderBrevig I addressed the missing rules, but I didn't notice a difference in my build of helix, perhaps the branch is too old? Also, I added some screens.

If you add this to your config `:config-open` then you should see both features. Color mode will change the color of `NOR` `INS` and `SEL`, and the indent guide will draw vertical bars at each indentation level.

```toml
[editor]
color-modes = true

[editor.indent-guides]
render = true
```

To see the `markup.link.label` open a `.md` file and have a link in it `[example label](https://example.com)`

AlexanderBrevig (approved):

nrdxp:
Thanks for the info, I was able to test the changes now, and I force pushed some better choices. I think that should be everything :crossed_fingers: 

Only thing I'm wondering is, should we leave this an an extended theme, or just replace snazzy outright?

the-mikedavis:
> should we leave this an an extended theme, or just replace snazzy outright?

Hmm good question. Since you originally added snazzy I suppose it's up to you. I like your reasoning and this looks like an improvement to me. But it does introduce some new warmer colors that make the palette less cold overall (which could be good or bad depending on how diverse of a palette you like).

the-mikedavis (approved):

Issue #3482, opened 2022-08-19
Kill LSP gracefully


dead10ck:
https://github.com/helix-editor/helix/blob/master/helix-lsp/src/client.rs#L62

It seems [`kill_on_drop` sends a SIGKILL](https://docs.rs/tokio/latest/tokio/process/struct.Child.html#method.kill) which is kind of aggressive, and could cause corruption. It's occurring to me that this may be why rust-analyzer has to reindex everything every time it runs. It also looks like there isn't a way to send a SIGTERM through the built-in tokio API. It looks like [we'd have to add platform-specific code to handle this properly](https://stackoverflow.com/questions/49210815/how-do-i-send-a-signal-to-a-child-subprocess).

_Originally posted by @dead10ck in https://github.com/helix-editor/helix/pull/3435#discussion_r950349542_

mangas:
https://docs.rs/nix/latest/nix/ may have some useful stuff for this

pascalkuthe:
Pretty sure this is fixed now wooth us giving lsps a ti.eout to close before force closing them

dead10ck:
Are LS's supposed to terminate themselves? Or is it common to have to send them a SIGTERM?

pascalkuthe:
There is a multiphase shutdown but in the end they are supposed to do it kn their own https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#exit

I think we have a timeout that waits for them to shutdown if they don't we kill

PR #3972, opened 2022-09-26
[lsp-restart]: call the force_shutdown method for the old_client

### LINKED ISSUE #3968: Lsp restart command spawn multiple procces for every call
### Summary

The new lsp restart command does not shut down the current server, so one process is left hanging, and a new one is executed.
Opening the editor for the first time: 3 processes
imagen

Running lsp-restart for the first time: 6 processes
imagen

Running lsp-restart for the second time: 9 processes
imagen

### Reproduction Steps

1. Open helix
2. Open the rust project and run the lsp-restart command once.
3. run lsp-restart a second time

### Helix log

not relevant

### Platform

Windows

### Terminal Emulator

wezterm 20220908-191626-25cd05a8

### Helix Version

helix 22.08.1 (e8f0886)

axdank:
fix #3968

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -359,9 +359,7 @@ impl Registry {
                 let (_, old_client) = entry.insert((id, client.clone()));
 
                 tokio::spawn(async move {
-                    if let Err(e) = old_client.force_shutdown().await {
-                        log::error!("failed shutdown language server: {}", e);
-                    }
+                    let _ = old_client.force_shutdown().await;
                 });
 
                 Ok(Some(client))
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -356,7 +356,13 @@ impl Registry {
                 let NewClientResult(client, incoming) = start_client(id, language_config, config)?;
                 self.incoming.push(UnboundedReceiverStream::new(incoming));
 
-                entry.insert((id, client.clone()));
+                let (_, old_client) = entry.insert((id, client.clone()));
+
+                tokio::spawn(async move {
+                    if let Err(e) = old_client.shutdown_and_exit().await {
+                        log::error!("failed shutdown language server: {}", e);
+                    }
+                });
 
                 Ok(Some(client))
             }
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -359,7 +359,7 @@ impl Registry {
                 let (_, old_client) = entry.insert((id, client.clone()));
 
                 tokio::spawn(async move {
-                    if let Err(e) = old_client.shutdown_and_exit().await {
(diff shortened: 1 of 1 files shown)


helix-lsp/src/lib.rs
```diff
@@ -356,7 +356,13 @@ impl Registry {
                 let NewClientResult(client, incoming) = start_client(id, language_config, config)?;
                 self.incoming.push(UnboundedReceiverStream::new(incoming));
 
-                entry.insert((id, client.clone()));
+                let (_, old_client) = entry.insert((id, client.clone()));
+
+                tokio::spawn(async move {
```
mangas, line 361: I wonder if we could add this to the Drop implementation instead on the server I guess?

mangas: I don't have a way of replicating this unfortunately, but what I'm thinking is that if you impl Drop for the client, then you could just call shutdown_and_exist, then print something if needed. This approach may also solve https://github.com/helix-editor/helix/issues/3482

helix-lsp/src/lib.rs
```diff
@@ -356,7 +356,13 @@ impl Registry {
...
+                    if let Err(e) = old_client.shutdown_and_exit().await {
```
dead10ck, line 362: I wonder if we should also attempt to kill the child process ourselves as well in case this fails?

dead10ck (approved):
One minor comment, but not a blocker, since this is a clear improvement. Thanks!

helix-lsp/src/lib.rs
mangas: I think the main issue there it's shutdown is async, I'll take a look at thsi when I have some time. This is also not a blocker, was just an improvement for better encapsulation IMO

helix-lsp/src/lib.rs
dead10ck: It looks like that method does not attempt to do anything with the child process, so if the LS fails to kill itself after the `shutdown` and `exit` methods, the process may linger. Maybe that would be an appropriate place to try to kill the process though.

helix-lsp/src/lib.rs
```diff
@@ -359,7 +359,7 @@ impl Registry {
                 let (_, old_client) = entry.insert((id, client.clone()));
 
                 tokio::spawn(async move {
-                    if let Err(e) = old_client.shutdown_and_exit().await {
+                    if let Err(e) = old_client.force_shutdown().await {
```
dead10ck, line 362: If we're opting for this instead, this method already prints an error message on shutdown failure, so we don't need one here too. It should suffice to just call the method.

dead10ck:
I actually still think this is a good change. It shuts the LS down gracefully, using the spec's documented shutdown procedure, rather than just `kill -9`ing the process.

dead10ck (approved):
Looks good, thanks!

PR #3973, opened 2022-09-26
fix: Improve JSX and TSX tag highlighting

jchand99:
Before:
Screen Shot 2022-09-26 at 00 27 57

After:
Screen Shot 2022-09-26 at 00 28 25


the-mikedavis (approved):
This is a big improvement, thanks!

jchand99:
Glad I could be of help!

PR #3948, opened 2022-09-23
Add gruvbox dark

svenstaro:
Uses the [official gruvbox dark hard colors](https://github.com/morhetz/gruvbox/blob/master/colors/gruvbox.vim#L89).

There's already [gruvbox_light](https://github.com/helix-editor/helix/blob/master/runtime/themes/gruvbox_light.toml) so let's also have gruvbox_dark. :)

To note, it would be technically more correct to rename the original gruvbox helix theme to `gruvbox_dark` and have _this_ one be called `gruvbox_dark_hard` but eh, it would break some configs.

PR #3983, opened 2022-09-26
feat: add markup scopes to themelint

AlexanderBrevig:
Based on the conversation here https://github.com/helix-editor/helix/pull/3948#discussion_r979001898 it would be nice if `themelint` also catches missing markup scopes.

I explicitly check all the sub keys, and not the catch all as to help people who omit the catch-all in favor of specificity.

### DIFF
--- a/xtask/src/themelint.rs
+++ b/xtask/src/themelint.rs
@@ -50,6 +50,23 @@ fn get_rules() -> Vec<Require> {
         // Check for visible cursor
         Require::Existence(Rule::has_bg("ui.cursor.primary")),
         Require::Existence(Rule::has_either("ui.cursor.match")),
+        // Markup scopes
+        Require::Existence(Rule::has_either("markup.heading.1")),
+        Require::Existence(Rule::has_either("markup.heading.2")),
+        Require::Existence(Rule::has_either("markup.heading.3")),
+        Require::Existence(Rule::has_either("markup.heading.4")),
+        Require::Existence(Rule::has_either("markup.heading.5")),
+        Require::Existence(Rule::has_either("markup.heading.6")),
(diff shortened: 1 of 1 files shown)


kirawi:
We could also have `themelint` automatically run in CI.

xtask/src/themelint.rs
```diff
@@ -50,6 +50,23 @@ fn get_rules() -> Vec<Require> {
         // Check for visible cursor
         Require::Existence(Rule::has_bg("ui.cursor.primary")),
         Require::Existence(Rule::has_either("ui.cursor.match")),
+        // Markup scopes
+        Require::Existence(Rule::has_either("markup.heading.1")),
+        Require::Existence(Rule::has_either("markup.heading.2")),
+        Require::Existence(Rule::has_either("markup.heading.3")),
+        Require::Existence(Rule::has_either("markup.heading.4")),
+        Require::Existence(Rule::has_either("markup.heading.5")),
+        Require::Existence(Rule::has_either("markup.heading.6")),
```
the-mikedavis, line 59: I think it's valid for a theme to only define `markup.heading` and not `.marker` or the `.n` for headings

AlexanderBrevig: Defining `markup.heading` will make all these pass, but _not_ defining it will force you to define everything.

pickfire:
> We could also have themelint automatically run in CI.

I am thinking we keep it off for now, or maybe make it faillable until we know what are the options we want to lint.

xtask/src/themelint.rs
pickfire: If anything I think the error should be on `markup.heading` instead of `.1`... which is a bit misleading.

PR #3985, opened 2022-09-26
goto_window_* extends selection

greg-enbala:
This PR changes:
 - `goto_window_{top|center|bottom}` now extend the current selection instead of dumping it when in select mode.
 
Not sure what the desired functionality is here and if this isn't it feel free to close this PR or ask me to close this PR. Personally I expected it to extend the selection. 

Before. You can see that I enter `SEL` mode, navigate the file and the selection _is not_ extended:
https://asciinema.org/a/5cgczWE5znBuFfLVcSh47RKyk

Now. You can see I enter `SEL` mode, navigate the file and the selection _is_ extended:
https://asciinema.org/a/aBlH5PFjZXhBoiJzPQsggGqP8

Apologies if I'm not following procedure around here. Looked like a quick fix so I went for it. By all means let me know how to do this the right way if this isn't it! Cheers and thanks for a great editor!

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -890,7 +890,6 @@ fn goto_window(cx: &mut Context, align: Align) {
         .selection(view.id)
         .clone()
         .transform(|range| range.put_cursor(text, pos, cx.editor.mode == Mode::Select));
-    push_jump(view, doc);
     doc.set_selection(view.id, selection);
 }
 


helix-term/src/commands.rs
```diff
@@ -885,8 +885,13 @@ fn goto_window(cx: &mut Context, align: Align) {
...
 
     let pos = doc.text().line_to_char(line);
-
-    doc.set_selection(view.id, Selection::point(pos));
+    let text = doc.text().slice(..);
+    let selection = doc
+        .selection(view.id)
+        .clone()
+        .transform(|range| range.put_cursor(text, pos, cx.editor.mode == Mode::Select));
+    push_jump(view, doc);
```
the-mikedavis, line 893: I think adding a jump point here is too noisy

the-mikedavis (approved):
I'm not sure I would want this behavior in normal mode but I do like it in select mode 👍

Plus now `gd`/`gt` align with `gl`/`gh` in select mode.

pascalkuthe:
I have a local branch (didnt get around to opening a PR yet) that also fixes this and adds an extend_to_X command for all goto_X commands that do not have one already. 

I also remember reading on some discussion that goto commands that operate on the same line (like `goto_line_end`) might be changed to also extend the selection in normal mode to be more consistent with other movements

the-mikedavis:
Whether the goto- commands should extend is still up for debate IIRC. I think this is a good change on its own and it's better to keep changes like this small

PR #3986, opened 2022-09-26
Update tutor

thiagomajesk:
Updates the tutor introduction to make it more inclusive for people that don't use `hjkl` keys for navigation.

The objective is to make it clear from the get-go that it doesn't really matter what keys you'll be using for navigation. This should make the modal experience more approachable for beginners and other folks coming to Helix that are not part of the Vim community. 

Even though this seems a small change, I think it has a positive intent in that it breaks some "tribalistic" barriers created by the Vim community that there's a always a *right* way instead of a *preferred* way to use the tool. 

As a new community, I think it's time to abandon some baggage, especially in some trivial matters like this. Cheers! 😊

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -24,17 +24,24 @@ _________________________________________________________________
 =                  1.1 BASIC CURSOR MOVEMENT                    =
 =================================================================
 
+In other modal editors like Vim and Kakoune, the cursor is 
+traditionally moved using the h, j, k, l keys as shown bellow:
+
           ↑
           k       * h is on the left
       ← h   l →   * l is on the right
           j       * j looks like a down arrow
           ↓
 
- The cursor can be moved using the h, j, k, l keys, as shown
- above. The cursor/arrow keys will also work, but it is faster
- to use the hjkl keys as they are closer to the other keys you
- will be using. Try moving around to get a feel for hjkl.
- Once you're ready, hold j to continue to the next lesson.
+You can also use the arrow keys, but it doesn't really matter which
+one you prefer as long as you feel confortable with your choice.
+
+Some think it's faster to use the hjkl keys as they are closer to 
+the other keys you will be using. Others think the arrow keys are 
+more intuitive because they are used in other editors as well. 
+
+Try moving around to get the hang of it. Once you're ready, go down 
+to continue to the next lesson.
 
 
 
@@ -134,7 +141,7 @@ _________________________________________________________________
 =                        CHAPTER 1 RECAP                        =
 =================================================================
 
- * Use the h,j,k,l keys to move the cursor.
+ * Use the h,j,k,l keys or the arrow keys to move the cursor.
 
  * Type : to enter Command mode.
    * The q/quit and q!/quit! commands will exit Helix. The


the-mikedavis:
Home-row navigation is the recommended way to use the editor and the tutor teaches the recommended way.

Normal-mode arrow keys are [bound](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/helix-term/src/keymap/default.rs#L9-L12) and [documented](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/book/src/keymap.md?plain=1#L34-L37) and I think that's sufficient.

thiagomajesk:
> Home-row navigation is the recommended way to use the editor and the tutor teaches the recommended way.
> 
> Normal-mode arrow keys are [bound](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/helix-term/src/keymap/default.rs#L9-L12) and [documented](https://github.com/helix-editor/helix/blob/8fc493137d159c4b1fbf964ed52dc6d53f6fcaf8/book/src/keymap.md?plain=1#L34-L37) and I think that's sufficient.

Hi @the-mikedavis!

Is there an official guide other than the tutor that says that the home-row navigation is the recommended way? I don't think there's any reason why we can't present both options and leave it open for users to decide. 

Honestly, there's no loss in doing so other than making a beginner's first experience with modal editing less traumatic.

IMHO, I don't think there are any advantages to strictly enforce this style of navigation besides being Vim compatible for people with years of muscle memory - it might not be a big deal for some folks, but it certainly is for newcomers. 

BTW, since most people using the tutor will be beginners, this update aims to make both options available, instead of strictly saying one is *correct* over the other. Opinions aside, there are some advantages for people outside of Vim to use arrow navigation since it's supported by normal and insert modes.

thiagomajesk:
Hi @greg-enbala, thanks for your 2 cents on the subject... Two things that caught my attention in your comment:

1) > I guess that leads to the question of, do most people who find vim/kak/hx useful use arrows or the home row for navigation?

I also think this is a problem related to Helix position in the text editor scenario. For instance: Is Helix's usefulness scoped only to people that used Vim or Kakoune in the past? Is Helix just a different implementation of Vim and Kakoune with different keybindings? If you just want Vim with different keybindings, why not create a plugin for it?

**From the vision document**

> Don't try to be everything for everyone. There are many great editors out there to choose from. Let's make Helix one of those great options, with its own take on things.

> We aren't playing code golf. It's more important for the keymap to be consistent and easy to memorize than it is to save a key stroke or two when editing.

This seems to indicate that although Vim and Kakoune are a good idea, we have room to out grow their constraints, no!? 

2) > Also I suspect many navigational shortcuts that work with the home row don't work with the arrow keys like g,h is not the same as g,<left> and that may be another reason for folks to get used to the home row and for the documentation to promote its use.

If that's the only argument against it, I'm in favor of supporting that, why not!? It won't make a difference if you already have the muscle memory and prefers using `hjkl`. However, it makes a difference for people that don't use it but might feel pressured to use something unnatural to them because `hjkl` is the *right way* (which is a kinda ridiculous mantra to inherit from Vim tbh).

There's some discussion going on here about keybindings that touch on the subject if Helix is going to be its own thing or not. Unfortunately, most of it hasn't been officially addressed yet: https://github.com/helix-editor/helix/issues/165.

IMHO, in the end, it all boils down to an official position if Helix is going 100% the "Vim reskin" route, or if is it becoming a different modal experience altogether.

dead10ck:
Recommending home row navigation has nothing to do with in/exclusivity. Some folks use the arrow keys because they find it easier to move a few characters than to escape, move, and reenter insert mode, and this is totally fine. Some users have been trying to stay in insert mode and treat helix like a modeless editor, and this inevitably leads them to pain because helix was not designed as a modeless editor, and there are [several technical reasons](https://github.com/helix-editor/helix/pull/3671#issuecomment-1242765458) that making it function well as one is a challenge.

At this moment in time, the arrow keys are not a recommended *primary* means of navigation, so we should not be introducing language that implies that both are equally good ways of navigating, as this will just encourage new users to treat helix like it's every other editor they've used. And the reason has nothing to do with sticking our noses up at noobs or any such nonsense.

If you like arrow keys, all the more power to you. But at best, that will only get one so far in learning to become productive in helix, and at worst, one will just have a bad experience. If the docs suggest that this way of navigating is equally functional and valid, one might rightly blame the editor for this bad experience.

The docs recommend home row navigation because this is what works best *at a technical level*, not at an individual level.

David-Else:
Modal editing in Helix and Vim is all about `hjkl` to navigate, without that you constantly leave the home row and all the other carefully chosen key combinations become a waste of time. 

It does matter A LOT what keys you use for navigation, you are interacting with a holistic system that is designed for touch typists from the start. New users should be aware of this and not led into bad habits that will hold them back from getting good, and ultimately getting much more enjoyment and productivity out of using Helix.

EpocSquadron:
> Recommending home row navigation has nothing to do with in/exclusivity. Some folks use the arrow keys because they find it easier to move a few characters than to escape, move, and reenter insert mode, and this is totally fine. Some users have been trying to stay in insert mode and treat helix like a modeless editor, and this inevitably leads them to pain because helix was not designed as a modeless editor, and there are [several technical reasons](https://github.com/helix-editor/helix/pull/3671#issuecomment-1242765458) that making it function well as one is a challenge.

One nitpick here, I don't think this issue is about the insert mode bindings, but the normal mode ones. However there is the interesting side effect that if arrow keys are given equal weight, there will be a nasty surprise when they find it missing in insert mode.

On the other hand, hjkl is very qwerty centric and there are a good few of us, myself included that prefer alternative layouts. In these cases, it's common to use an alternate keyboard layer at the hardware or OS level to map arrows to the hjkl positions. Worth considering if we make clear that arrow keys are there to support these use cases but that home row navigation is still the recommended path.

thiagomajesk:
> > One nitpick here, I don't think this issue is about the insert mode bindings, but the normal mode ones.
> 
> Yes, you are right; however, I think the two issues tend to become conflated often for new users coming from modeless editors, who are used to moving around with arrow keys as the primary means of navigation. If arrow keys are encouraged, then I think this naturally leads to sticking to a workflow they used in other editors, which means staying in insert mode.

I understand from your previous comment that you have some history with this, but I think you are conflating two problems. 
The proposal has nothing to do with supporting or enforcing modeless editing. The idea is to support and recommend arrow keys as an alternative for `hjkl`. Perhaps there are some technical issues to be solved first, as you mentioned.

> Modal editing in Helix and Vim is all about hjkl to navigate, without that you constantly leave the home row and all the other carefully chosen key combinations become a waste of time.

The vision document should probably be updated then because I had the impression that consistency and simplicity were more important than saving nanoseconds of hand movement from this sentence:

> We aren't playing code golf. It's more important for the keymap to be consistent and easy to memorize than it is to save a key stroke or two when editing.

---

It's not the purpose of this PR to start a flamewar, but somehow this type of discussion always strikes a nerve for hardcore Vim users. I understand that this is a very passionate community and people are protective of their tools, but existing users are literally losing nothing. 

Perhaps I'm getting some missing signals from the vision document and Helix is actually only targeted at Vimers - which leaves me with the question of why people are using Helix, instead of Neovim with Helix's keybindings. 🤔

thiagomajesk:
Hi @archseer! 

> Well, both selection first approach and not having to configure lots of plugins appeals to some people.

I think that Helix has the same appeal as other Vim distributions like LunarVim and SpaceVim then. Except it works like Kakoune. So, it's equivalent to an opinionated Kakoune distribution.

> Helix was written out of frustration as my personal Vim replacement

This makes it crystal clear what path you're going to take with Helix, thanks for the explanation! It wasn't clear if the editor is trying to mimic Vim or actually do something new. I do like Helix, but I'm going to phase it out from my workflow since I don't intend to keep adopting Vim primitives. Good luck with the project though. Cheers!

Issue #3982, opened 2022-09-26
Broken syntax highlighting in WGSL


Chickenkeeper:
### Summary

WGSL code currently generates enormous amounts of erroneous error highlights.

![whole lotta red](https://user-images.githubusercontent.com/6290295/192326091-2f5434d0-058d-42ad-bf3f-1f5627af394c.png)

I suspect this may be because of the recent changes to the WGSL syntax. The version of tree-sitter-wgsl that the latest release of Helix uses had been updated to reflect the changes, but Helix's WGSL grammar hasn't been updated since months before the changes were introduced. Also WGSL code written with the old syntax is still highlighted as it was when wgsl support was first added (#1166).

I think it's worth noting that WGSL syntax highlighting has always been incomplete, but it was at least usable.

### Platform

Windows

### Terminal Emulator

PowerShell

### Helix Version

helix 22.08.1 (66276ce6)

dead10ck:
This should actually be pretty easy to fix, if you'd like to give it a try. You can copy the latest commit hash and paste it here:

https://github.com/helix-editor/helix/blob/98dd9c4f2bf71e9afb72f61af9d45100fda0526e/languages.toml#L851

Then `hx --grammar fetch` and `hx --grammar build`.

Chickenkeeper:
> This should actually be pretty easy to fix, if you'd like to give it a try. You can copy the latest commit hash and paste it here:
> 
> https://github.com/helix-editor/helix/blob/98dd9c4f2bf71e9afb72f61af9d45100fda0526e/languages.toml#L851
> 
> Then `hx --grammar fetch` and `hx --grammar build`.

Thanks. I've been learning more about language support in Helix and I've got a fix now. I'll post a PR soon.

Chickenkeeper:
This issue's been solved by #3996, so it can be closed now.

PR #3996, opened 2022-09-27
WGSL syntax highlighting fix

Chickenkeeper:
no more red

This PR adresses #3982. tree-sitter-wgsl is bumped to the latest commit and highlights.scm is updated accordingly.

The updated highlights file is based on the one in the tree-sitter-wgsl repository but with a few changes to make it work properly in Helix. If there are any problems with it let me know.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #3998, opened 2022-09-27
Update everforest theme

CptPotato:
This syncs the latest version of the `everforest` theme, which I've kept up to date in my [theme repo](https://github.com/CptPotato/helix-themes).

It mostly adds missing theme keys and includes some smaller adjustments.

image
*edit: for whatever reason GitHub slightly tints the image on upload, so this is not 100% accurate.*

### DIFF
--- a/runtime/themes/everforest_dark.toml
+++ b/runtime/themes/everforest_dark.toml
@@ -8,11 +8,12 @@
 # Email: sainnhe@gmail.com
 # License: MIT License
 
-"constant.character.escape" = "orange"
 "type" = "yellow"
 "constant" = "purple"
 "constant.numeric" = "purple"
+"constant.character.escape" = "orange"
 "string" = "green"
+"string.regexp" = "blue"
 "comment" = "grey0"
 "variable" = "fg"
 "variable.builtin" = "blue"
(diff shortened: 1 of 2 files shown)


CptPotato:
I've also noticed that the `sonokai` theme found its way [into the repo](https://github.com/helix-editor/helix/blob/bdc7b35214ec12a013941f7f6753331d47952bb2/runtime/themes/sonokai.toml). I'm not sure if/what significant changes @p4ymak made but I could update this one, too, if desired.

the-mikedavis:
The themelint points out that `ui.selection` and `ui.selection.primary` are the same. Would you like to separate them or leave them as-is? I don't have strong feelings either way

CptPotato:
Good point. I'm only using `ui.selection` in the theme(s) so the `primary` one falls back to this.

So far this didn't really bother me so I don't have a strong opinion on this either. When *I* use multiple selections I edit all of them at the same time, so the primary one has no special meaning (except for when I use `,` to only keep the primary selection).

But I'll see if I can use a different color for secondary selections without it looking weird. Otherwise I still think it's fine? 🤷‍♂️

I'll also check if the theme lint finds other issues.

archseer:
This is already a big improvement so I'm merging as is but feel free to open a follow-up if you decide to add some changes based on the linter :)

PR #4000, opened 2022-09-27
Add :character-info command

### LINKED ISSUE #3885: Add Vim’s “get ascii” command
Reference: https://vimhelp.org/various.txt.html#ga

wetheredge:
This is my take on Vim's get ascii (<kbd>g</kbd><kbd>a</kbd>) / get utf8 (<kbd>g</kbd><kbd>8</kbd>) command. This works on graphemes not characters, so it correctly handles things like emoji with skin tone modifiers or characters with combining diacritics.

I'm open to opinions on what the output should look like. Atm, it outputs the decimal value only for single byte characters in an ascii-compatible encoding. For UTF-8, it also prints the decoded codepoints. (The decoder is based on [fasterthanlime's article](https://fasterthanli.me/articles/working-with-strings-in-rust) since I could not find anything in std or already in helix, or on crates.io for that matter, and it's pretty concise).

| Character | Encoding | Output |
|-|-|-|
| a | UTF-8 | `"a" (U+0061) Dec 97 Hex 61` |
| a | windows-1252 / extended ascii | `"a" Dec 97 Hex 61` |
| newline | UTF-8 | `"\n" (U+000a) Dec 10 Hex 0a` |
| CRLF | UTF-8 | `"\r\n" (U+000d U+000a) Hex 0d + 0a` |
| é | UTF-8 | `"é" (U+00e9) Hex c3 a9` |
| é | windows-1252 | `"é" Hex e9` |
| ë | UTF-8 | `"ë" (U+0065 U+0308) Hex 65 + cc 88` |
| 👍🏽 | UTF-8 | `"👍🏽" (U+1f44d U+1f3fd) Hex f0 9f 91 8d + f0 9f 8f bd` |

Remaining tasks:
- [x] Improve the help text
- [x] Replace `Debug` formatting of character with something more appropriate. `Debug` handles escaping non-printable & whitespace characters, but it also escapes the combining diaeresis on the ë…
	- Now the only escapes are `\0`, `\t`, `\n`, and `\r`
- [x] Tests

I'm not familiar with the codebase, so let me know if there's anything that could be more idiomatic. It does not have any tests yet, but it looks like most commands don't?

closes #3885

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1140,9 +1140,7 @@ fn get_character_info(
             );
 
             if let encoding::EncoderResult::Unmappable(char) = result {
-                cx.editor
-                    .set_error(format!("{char:?} cannot be mapped to {}", encoding.name()));
-                return Ok(());
+                bail!("{char:?} cannot be mapped to {}", encoding.name());
             }
 
             for byte in &bytes[current_byte..] {
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -4,6 +4,7 @@ use crate::job::Job;
 
 use super::*;
 
+use helix_core::encoding;
 use helix_view::editor::{Action, CloseError, ConfigEvent};
 use ui::completers::{self, Completer};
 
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Instead of a regular command which must be bound to a key, I think this would be appropriate as a typable command (`helix-term/src/commands/typed.rs`). We could follow Vim and call it `:ascii` (aliased to `:as`) or come up with a new name

wetheredge:
Got it switched over to a typable command. I've named it `:character` (aliased to `:char`) since it should support any encoding, even if it's not ASCII compatible.

sudormrfbin:
Maybe something a bit more descriptive like `:character-info`?

wetheredge:
Rebased to fix conflicts. Let me know if I should write tests, otherwise it should be ready

ArikRahman (approved):

the-mikedavis:
I think this is looking great 😀

You can add an integration test like so:

```rs
# helix-term/tests/test/commands.rs
#[tokio::test(flavor = "multi_thread")]
async fn test_character_info() -> anyhow::Result<()> {
    test_key_sequence(
        &mut helpers::AppBuilder::new().build()?,
        Some("ih<esc>h:char<ret>"),
        Some(&|app| {
            assert_eq!(r#""h" (U+0068) Dec 104 Hex 68"#, app.editor.get_status().unwrap().0);
        }),
        false,
    )
    .await?;

    Ok(())
}
```

(you will want to rebase on latest master first since there are recent changes to the integration testing harness)

wetheredge:
Thanks for the pointer on the tests. I added that and a few more cases to fully cover the possible output formats.

helix-term/src/commands/typed.rs
```diff
@@ -1017,6 +1018,133 @@ fn set_encoding(
...
+            let (result, _input_bytes_read) = encoder.encode_from_utf8_to_vec_without_replacement(
+                &char.to_string(),
+                &mut bytes,
+                true,
+            );
+
+            if let encoding::EncoderResult::Unmappable(char) = result {
+                cx.editor
+                    .set_error(format!("{char:?} cannot be mapped to {}", encoding.name()));
+                return Ok(());
```
the-mikedavis, line 1129:
```suggestion
                bail!("{char:?} cannot be mapped to {}", encoding.name());
```

has the same effect

the-mikedavis (approved):

wetheredge:
@archseer, sorry for the ping, but is there any chance of this making it into 22.12? I can rebase if that would be helpful.

kirawi (approved):

wetheredge:
Should I resolve the conflict? Looks like it's just accepting both the test I added & a couple new ones from `master`.

archseer:
Sorry for the delay! Looks good to me, just needs a rebase :)

PR #4007, opened 2022-09-28
Fix tutor typo `favourite` to `favorite`

s0LA1337:
Fixed the mismatch of the word `favourite` and `favorite` in the tutor.
So i fixed that to be synchronous now.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -519,7 +519,7 @@ _________________________________________________________________
     like 'oranges'.
  6. Type , to remove the second cursor.
 
- --> I like to eat apples since my favorite fruit is apples.
+ --> I like to eat apples since my favourite fruit is apples.
      I like to eat oranges since my favourite fruit is oranges.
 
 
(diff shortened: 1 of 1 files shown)


runtime/tutor
```diff
@@ -519,7 +519,7 @@ _________________________________________________________________
     like 'oranges'.
  6. Type , to remove the second cursor.
 
- --> I like to eat apples since my favorite fruit is apples.
+ --> I like to eat apples since my favourite fruit is apples.
      I like to eat oranges since my favourite fruit is oranges.
```
the-mikedavis, line 523: actually I think the typo is the one with the `u` - there's another `favorite` down on L833

David-Else: https://www.grammarly.com/blog/favorite-favourite/

`favorite` is American English, it is `favourite` everywhere else.

the-mikedavis: It seems like we use American English elsewhere in the tutor, but the other "favorite" is the only example I found of that. I don't have strong feelings either way between American vs. British

s0LA1337: so i assume i will fix it up to use the AE version.

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4024, opened 2022-09-29
Fix multi byte auto pairs

### LINKED ISSUE #3946: Auto-pairs with AltGr key cause crash
### Summary

Hi all,

on German keyboards ‘typographic’ quotation marks require the use of the `AltGr` key.

Using automatic paring in combination with the `AltGr` causes a helix crash.

```
# config.toml
[editor.auto-pairs]
'„' = '“'
'‚' = '‘'
```

Thank you very much!

### Reproduction Steps

 

I tried this:

1. `hx`
2. Press `AltGr + v`

Expected behaviour: `AltGr + v` should produce a `„`; helix’s auto-pairing should extend this to `„“`.

### Helix log

```
helix_view::editor [ERROR] Failed to initialize the LSP for `source.md` { LSP not defined }
```

**Error message:**

```
> thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 3, Rope/RopeSlice char length 2', /host/cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.0/src/slice.rs:349:41
> note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

</details>

### Platform

Void with River WM

### Terminal Emulator

Fish

### Helix Version

22.08.1

dead10ck:
Currently if you configure any auto pairs with multi-byte Unicode characters, auto pairing does not produce a correct selection. This is because it is summing bytes, when it should be summing chars.

I took this opportunity to migrate the auto pairs tests to the integration tests, as this makes the tests much more readable with the new selection DSL, and will also allow testing more complicated intersections of auto pairs with other features like dedenting, as described in https://github.com/helix-editor/helix/pull/1379#issuecomment-1003724997. Ironically, the tests weren't passing, and it turns out the test selection DSL parser suffered from exactly the same bug, counting bytes and not chars. This was also fixed, and regression tests were added.

Fixes #3946.


dead10ck:
Sigh, Windows strikes again. Will look into the failure tomorrow.

dead10ck:
Ok, actually it turned out to be a legitimate preexisting bug!

Additionally, it was discovered that there was a preexisting bug where the selection was not computed correctly in the case that the cursor
was:

1. a single grapheme in width
2. this grapheme was more than one `char`
3. the direction of the cursor is backwards
4. a secondary range

In this case, the offset was not being added into the anchor. An example of a case for this would be inserting a new pair in insert mode with multiple single width cursors at the end of a line in a CRLF document. This was fixed.

helix-view/src/clipboard.rs
```diff
@@ -17,7 +17,7 @@ pub trait ClipboardProvider: std::fmt::Debug {
 #[cfg(not(windows))]
 macro_rules! command_provider {
     (paste => $get_prg:literal $( , $get_arg:literal )* ; copy => $set_prg:literal $( , $set_arg:literal )* ; ) => {{
-        log::info!(
+        log::debug!(
```
the-mikedavis, line 20: Looks like an unrelated change here

dead10ck: Yeah, I made this debug because it makes the integration tests really noisy. This line gets printed for every single test.

Even outside of that context though, it seems like info that is really only important to print if you're trying to debug a problem with the clipboard.

What do you think though?

the-mikedavis: Hmm yeah I think those could be debug: we have a `log::warn!` when there isn't a clipboard provider and this info shows up in `hx --health clipboard`. The current `log::info!` does seem a little noisy

helix-core/src/auto_pairs.rs
```diff
@@ -282,20 +274,20 @@ fn handle_open(doc: &Rope, selection: &Selection, pair: &Pair) -> Transaction {
 
         let change = match next_char {
             Some(_) if !pair.should_close(doc, start_range) => {
-                len_inserted = pair.open.len_utf8();
+                len_inserted = 1;
```
archseer, line 277: Shouldn't this count graphemes? The pairs could technically be multi chars (although the code doesn't properly handle this yet)

dead10ck: No, this is being used to compute the selection, which uses char indices. Although yes, this will have to be reworked for multi-char auto pairs.

dead10ck: No, as it turns out, [`char::len_utf8`](https://doc.rust-lang.org/std/primitive.char.html#method.len_utf8) returns the number of bytes of the `char`. But one `char` is always just one `char`, so it just so happens to be correct for ASCII, but goes too far for high Unicode characters.

dead10ck: @archseer does this make sense?

the-mikedavis: Ah, I think that makes a lot of sense. I was curious about the literal `1`s as well. This might be worth a comment somewhere? I'm not sure exactly where I would put it

helix-term/tests/test/auto_pairs.rs
```diff
@@ -1,21 +1,543 @@
...
-            editor: helix_view::editor::Config {
-                auto_pairs: AutoPairConfig::Enable(false),
-                ..Default::default()
-            },
+    Ok(())
+}
+
+#[tokio::test]
+async fn insert_configured_multi_byte_chars() -> anyhow::Result<()> {
+    // [NOTE] these are multi-byte Unicode characters
```
the-mikedavis, line 31: tree-sitter-comment would make this a little more high-profile as `NOTE:` rather than `[NOTE]`

helix-core/src/auto_pairs.rs
```diff
@@ -321,13 +315,13 @@ fn handle_close(doc: &Rope, selection: &Selection, pair: &Pair) -> Transaction {
             // return transaction that moves past close
             (cursor, cursor, None) // no-op
         } else {
-            len_inserted += pair.close.len_utf8();
+            len_inserted = 1;
```
the-mikedavis, line 318: Is this an intentional change from `+=` to `=`?

dead10ck: Yes. It was just being added to 0 before. Doesn't really make a difference if it's added or assigned.

the-mikedavis (approved):

helix-term/tests/test/auto_pairs.rs
```diff
@@ -1,21 +1,543 @@
...
-                auto_pairs: AutoPairConfig::Enable(false),
-                ..Default::default()
-            },
+    Ok(())
+}
+
+#[tokio::test]
+async fn insert_configured_multi_byte_chars() -> anyhow::Result<()> {
+    // NOTE: these are multi-byte Unicode characters
+    let pairs = hashmap!('„' => '“', '‚' => '‘');
```
pickfire, line 32: Can we add one more example with double width characters? `「」`

dead10ck: Sure, done

the-mikedavis:
This looks good to me but I'll defer the merge to archseer

archseer:
Looks good but there's one conflict to resolve now

dead10ck:
Done!

archseer (approved):

PR #4027, opened 2022-09-29
Tutor 5.2.: Missing <ESC> before removing the second cursor

joleaf:
In tutor 5.2, when the user wants to remove the second cursor, he has to exit the Insert mode first.


CBenoit (approved):
LGTM

Issue #3558, opened 2022-08-27
Add syntax highlighting for new language (WebAssembly TextFormat)


Jomy10:
### Summary

I would like to add WebAssembly TextFormat syntax highlighting to Helix. But I can't get it to work.

### Reproduction Steps

I have added the following to `languages.toml`:
```toml
[[language]]
name = "webAssembly"
scope = "source.wat"
injection-regex = "wasm"
file-types = ["wat", "wast"]
roots = []
comment-token = ";;"

[[grammer]]
name = "webAssembly"
source = { git = "https://github.com/wasm-lsp/tree-sitter-wasm" }
```

I ran `hx --health webAssembly` and got the following output:
```
Configured language server: None
Configured debug adapter: None
Highlight queries: ✘
Textobject queries: ✘
Indent queries: ✘
```

I ran `hx --grammar fetch` but got no output related to the webAssembly.

I'm not sure what's wrong.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-08-27T18:46:44.679 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-08-27T18:49:30.203 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-08-27T18:49:53.540 helix_view::editor [ERROR] Failed to initialize the LSP for `source.wat` { LSP not defined }
2022-08-27T18:50:00.601 helix_view::editor [ERROR] Failed to initialize the LSP for `source.wat` { LSP not defined }
2022-08-27T19:01:19.849 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-08-27T19:05:44.283 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-08-27T19:06:34.344 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2022-08-27T19:08:58.360 helix_view::editor [ERROR] Failed to initialize the LSP for `source.wat` { LSP not defined }
2022-08-27T19:20:56.142 helix_view::editor [ERROR] Failed to initialize the LSP for `source.wat` { LSP not defined }
2022-08-27T19:24:36.958 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-08-27T19:24:36.959 helix_lsp::transport [ERROR] err: <- StreamClosed
[...]

PR #4040, opened 2022-09-29
Add support for webassembly text format

jrvidal:
Adds support for the WebAssembly text format with some basic queries.

Screenshot from 2022-09-29 23-16-43
Screenshot from 2022-09-29 23-16-52
Screenshot from 2022-09-29 23-17-29

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1786,9 +1786,20 @@ source = { git = "https://github.com/vito/tree-sitter-bass", rev = "501133e260d7
 name = "wat"
 scope = "source.webassembly"
 comment-token = ";;"
-file-types = ["wast", "wat"]
+file-types = ["wat"]
 roots = []
 
 [[grammar]]
 name = "wat"
 source = { git = "https://github.com/wasm-lsp/tree-sitter-wasm", rev = "2ca28a9f9d709847bf7a3de0942a84e912f59088", subpath = "wat" }
+
+[[language]]
+name = "wast"
+scope = "source.webassembly"
+comment-token = ";;"
+file-types = ["wast"]
+roots = []
+
+[[grammar]]
+name = "wast"
+source = { git = "https://github.com/wasm-lsp/tree-sitter-wasm", rev = "2ca28a9f9d709847bf7a3de0942a84e912f59088", subpath = "wast" }
--- a/languages.toml
+++ b/languages.toml
@@ -1781,3 +1781,14 @@ language-server = { command = "bass", args = ["--lsp"] }
 [[grammar]]
 name = "bass"
 source = { git = "https://github.com/vito/tree-sitter-bass", rev = "501133e260d768ed4e1fd7374912ed5c86d6fd90" }
+
+[[language]]
+name = "wat"
+scope = "source.webassembly"
+comment-token = ";;"
+file-types = ["wast", "wat"]
+roots = []
+
+[[grammar]]
+name = "wat"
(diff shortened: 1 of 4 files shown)


jrvidal:
Fixes #3558 #3559

languages.toml
```diff
@@ -1781,3 +1781,14 @@ language-server = { command = "bass", args = ["--lsp"] }
 [[grammar]]
 name = "bass"
 source = { git = "https://github.com/vito/tree-sitter-bass", rev = "501133e260d768ed4e1fd7374912ed5c86d6fd90" }
+
+[[language]]
+name = "wat"
```
jrvidal, line 1786: I think it would make more sense to call this `webassembly`, but it seems that this name determines the symbol that will be imported from the grammer dynamic library, and that's fixed to `tree_sitter_wat`.

languages.toml
```diff
@@ -1781,3 +1781,14 @@ language-server = { command = "bass", args = ["--lsp"] }
...
+scope = "source.webassembly"
+comment-token = ";;"
+file-types = ["wast", "wat"]
```
the-mikedavis, line 1789: It looks like wast is a separate grammar within the tree-sitter-wasm repo. It looks like a small superset of the wat parser though so I don't have strong feelings about separating these out into separate languages. What do you think: should `wast` be separate?

runtime/queries/wat/highlights.scm
```diff
@@ -0,0 +1,15 @@
...
+
+[(name) (string)] @string
+
+(identifier) @function
+
+[(comment_block) (comment_line)] @comment
+
+[(nat) (align_offset_value)] @constant.numeric.integer
+
+(value_type) @type
```
the-mikedavis, line 15:
```
["(" ")"] @punctuation.bracket
```

languages.toml
jrvidal: I bundled them together thinking that `.wast` was historically used for the true text format that is defined in the spec, and that usage might still linger around. Splitting the two is no big deal, if you're fine with having the less commonly used `wast` language bundled by default :+1:

languages.toml
the-mikedavis: There's an issue along those lines: https://github.com/helix-editor/helix/issues/3164

It would only be a cosmetic improvement for the file-type statusline element though

languages.toml
```diff
@@ -1781,3 +1781,25 @@ language-server = { command = "bass", args = ["--lsp"] }
...
+file-types = ["wat"]
+roots = []
+
+[[grammar]]
+name = "wat"
+source = { git = "https://github.com/wasm-lsp/tree-sitter-wasm", rev = "2ca28a9f9d709847bf7a3de0942a84e912f59088", subpath = "wat" }
+
+[[language]]
+name = "wast"
+scope = "source.webassembly"
```
the-mikedavis, line 1798: The scopes are the same between these two. I think wat could be `source.webassembly.text` but I'm not sure what to use for wast

the-mikedavis:
It looks like this needs a run of `cargo xtask docgen` to clean up the Docs CI

languages.toml
jrvidal: I've tentatively set this to `source.wat`, which is the scope used by the main extensions in Sublime and VSCode. But not strong opinions here. If we were to follow your suggestion, we could set `source.webassembly.script`?

the-mikedavis: We could use `source.webassembly.text` & `source.webassembly.script`. Since Sublime and VSCode use these though, I would prefer to also use `source.wat` & `source.wast`.

the-mikedavis:
Thanks!

PR #4042, opened 2022-09-29
Make Merge behave consistently

### LINKED ISSUE #2248: Characters rendering multiple times

pascalkuthe:
Supersedes #3695.

Currently, the `HighlightEvent` merge function overlays spans `Vec<(usize, std::ops::Range<usize>)>`` on an Iterator of `HighlightEvent`. The current merge doesn't handle overlapping spans well:

* diagnostic ranges may be discarded if they overlap
* ghost text may appear when a language server sends overlapping diagnostics

After a discussion with @the-mikedavis we realized that the solution in #3695 has a problem:
The spans must remain in a certain order to allow (somewhat performant) conversion to `HighlightEvent`s.
However, with #3015 we usually want to show the highest priority diagnostic on top (so that an error is not overlapped by a hint for example). This is not possible with existing approach or the one in #3695.

However, a few key realizations lead to an alternative approach:

* Implementing an `overlay` for **non-overlappig** spans is trivial and always produces the correct result
* **overlapping** spans with the same **`Highlight` scope** can be merged together  so that the resulting spans do not overlap
* The only case where overlapping spans occur are diagnostics. However, there are only 5 different `Highlight` scopes for diagnostics: One for each severity.  Therefore, the diagnostics can be separated into 5 sets of spans that all share the same highlights. These can then be used as an `overlay` as explained above.

This PR implements that approach. 
The `overlay` I am referring to here is essentially `merge` (renamed it to make it more clear). 
I implemented it from scratch because the old implementation was more complex than it needed to be and it was easier to integrate the merging directly into the iterator instead of handeling it in a seperate function.
Compared to previous implementation it should also perform a bit better (does not use any allocations for example).
This implementation can also automatically merge overlapping spans under the assumption that they have the same highlight. Therefore, there are two functions that produce such an iterator:

* `monotonic_overlay`: accepts `Span`s (range + Highlight) that must be monotonically ascending (no overlaps and sorted by start)
* `overlapping_overlay`: accepts a single Highlight and multiple `Range`s. (sorted by start)

The `overlapping_overlay` is used to layer the diagnostic levels and the `monotonic_overlay` is used for cursor/selections (and in the future rainbow brackets).

As we are now layering up to 6 iterators on top of each other, I was a bit concerned about performance.
However, the new `overlay` implementation is very performant so the overhead is very small.
Furthermore, I refactored the calls to the iterators a bit to completely avoid dynamic dispatch.
When dynamically dispatching an iterator a v-table lookup is necessary for each call to `next` (occurs in the hot loop).
While LTO might theoretically move this out of the loop, the code is pretty complex so its probably not happening.
Benchmarking the syntax highlighting code is kind of tricky but my -not very scientific- glances at flamegraphs show that this is probably neutral or an improvement to performance.

I have not written an extensive test-tsuite as @the-mikedavis did for 3695 #because the algorithm is so simple that one can easily reason about all possible scenarios.
In particular all test-cases used there trivially work here.
The old merge implementation was more complex and didn't have any tests either.
~However, a test-case that covers #2248 would probably be good, although I struggle to come up with one.
I am certain that this PR fixes that issue but until I can test it properly I am marking this as a draft.~
This definitely fixes #2248

~It might be good to rebase #2857 on this as that will probably produce the most overlays when thats merged.
However, the conflicts looked pretty daunting so I didn't get into that yet.~
I merge #2857 with this and have not found any problems


helix-term/src/ui/editor.rs
```diff
@@ -66,6 +65,52 @@ impl EditorView {
         &mut self.spinners
     }
 
+    #[allow(clippy::too_many_arguments)]
```
pascalkuthe, line 68: I had to add this one allow attribute here to make clippy shutup,
This function should be a closure that captures all arguments except `base`.
However the function must be generic over `base` but that is currently not possible with closures.
So the reason for the large argument count is just that were working around limitations of the rust type system.

helix-core/src/syntax/overlay.rs
```diff
@@ -0,0 +1,254 @@
...
+
+/// Overlays multiple different highlights from `spans` onto the `HighlightEvent` stream `events`.
+///
+/// The [`Span`]s yielded by `spans` **must never overlap*  or the iterator will produce incorrect results.
+/// The [`Span`]s **must be sorted** in ascending order by their start.
+/// If multiple [`Span`]s  share the same start, the ordering is arbitrary.
+///
+/// Together these two properties mean that `spans` must prduce monotonically increasing [`Span`]s.
+/// That means that the next span must always start after the last span ended:
+/// `next_span.end <= span.start`
```
the-mikedavis, line 19:
```suggestion
/// `span.end <= next_span.start`
```

helix-core/src/syntax/overlay.rs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                self.current_span = self.spans.next();
+            }
+
+            if let Some(span) = &mut self.current_span {
+                // only process the span if it actually covers this source (so starts before)
+                if span.start < end {
+                    // if the span starts inside the source,
+                    // split off the start of the source that is not highlighted
+                    // and emit this sourcespan first
```
the-mikedavis, line 195:
```suggestion
                    // and emit this source span first
```

helix-core/src/syntax/overlay.rs
```diff
@@ -0,0 +1,254 @@
...
+            }
+
+            break;
+        }
+
+        match replace(&mut self.next_event, self.events.next()) {
+            Some(event) => Some(event),
+            None => {
+                // Unfinished span at EOF is allowed  to finish.
+                // Special case for cursor past EOF
```
the-mikedavis, line 243:
```suggestion
                // Unfinished span at EOF is allowed to finish.
```

helix-core/src/syntax/overlay/test.rs
```diff
@@ -0,0 +1,32 @@
+use crate::syntax::{overlapping_overlay, Highlight, HighlightEvent::*};
+use std::iter;
+
+/// this tests checks that merging two overlapping ranges onto each other
+/// correctly preveres the order of merges.
+/// that is the highlight that is merged in last, gets applied last and overwrites the other layers
+/// In this test a range of lower priority (like a hint) starts at 2
+/// and another range of a high priority range (like an error) starts earlier
+/// with the old span implementation the hint would always overwrite the error.
+/// The new implementation (tested here) forces the
```
the-mikedavis, line 10: Looks like this sentence got chopped off

helix-core/src/syntax/overlay/test.rs
```diff
@@ -0,0 +1,32 @@
+use crate::syntax::{overlapping_overlay, Highlight, HighlightEvent::*};
```
the-mikedavis, line 1: For such a small test file I think inlining the module into `helix_core::syntax::overlay`'s file is appropriate and in-line with the rest of the unit tests in the codebase.

helix-core/src/syntax/overlay/test.rs
```diff
@@ -0,0 +1,32 @@
...
+/// with the old span implementation the hint would always overwrite the error.
+/// The new implementation (tested here) forces the
+#[test]
+fn overlay_long_hint() {
+    let base = iter::once(Source { start: 0, end: 31 });
+    let highlights = overlapping_overlay(base, [2..10].into_iter(), Highlight(1));
+    let highlights = overlapping_overlay(highlights, [0..4].into_iter(), Highlight(2));
+    let res: Vec<_> = highlights.collect();
+    assert_eq!(
+        &*res,
```
the-mikedavis, line 18:
```suggestion
        res,
```

helix-term/src/ui/editor.rs
```diff
@@ -276,29 +292,32 @@ impl EditorView {
...
         };
 
-        // basically just queries the theme color defined in the config
-        let hint = get_scope_of("diagnostic.hint");
-        let info = get_scope_of("diagnostic.info");
-        let warning = get_scope_of("diagnostic.warning");
-        let error = get_scope_of("diagnostic.error");
-        let r#default = get_scope_of("diagnostic"); // this is a bit redundant but should be fine
+        let highlight = match severity {
+            Some(Severity::Error) => get_scope_of("diagnostic.error"),
```
the-mikedavis, line 296: With this refactor we don't need `get_scope_of` to be a separate closure anymore

```rust
let scope = match severity { .. };
let highlight = theme.find_scope_index(scope).or_else(// ...
```

helix-term/src/ui/editor.rs
```diff
@@ -276,29 +292,32 @@ impl EditorView {
...
-                    Some(Severity::Error) => error,
-                    _ => r#default,
-                };
-                (
-                    diagnostic_scope,
-                    diagnostic.range.start..diagnostic.range.end,
-                )
-            })
-            .collect()
+        let diagnostic_ranges = doc.diagnostics().iter().filter_map(move |diagnostic| {
```
the-mikedavis, line 303: We end up filtering `doc.diagnostics()` for each severity, right? Could we iterate over it once and separate out the diagnostics into a `HashMap<Severity, SmallVec<Range<usize>>` to save some work?

helix-term/src/ui/editor.rs
```diff
@@ -363,17 +386,33 @@ impl EditorView {
             if range.head > range.anchor {
                 // Standard case.
                 let cursor_start = prev_grapheme_boundary(text, range.head);
-                spans.push((selection_scope, range.anchor..cursor_start));
+                spans.push(syntax::Span {
+                    scope: Highlight(selection_scope),
+                    start: range.anchor,
+                    end: cursor_start,
+                });
```
the-mikedavis, line 393: Some of the spans in this function will be zero-width - for example here when `range.anchor == cursor_start` (in a single-width selection, cursor == anchor). I was thinking we might want to wrap `spans.push` in a closure that doesn't push if `range.start == range.end`. What do you think?

helix-term/src/ui/editor.rs
```diff
@@ -1461,3 +1498,8 @@ fn canonicalize_key(key: &mut KeyEvent) {
         key.modifiers.remove(KeyModifiers::SHIFT)
     }
 }
+
+pub enum DocSyntaxHighlight<'d> {
+    TreeSitter(CharacterHighlightIter<'d>),
+    Default(iter::Once<HighlightEvent>),
+}
```
the-mikedavis, line 1505: This is so that we can use static dispatch for everything, right? As opposed to a Box of a dyn trait implementation?

This enum makes the uses of `doc_syntax_highlights` awkward (looking at the usages in picker.rs and here in editor.rs) and I think what we lose in elegance of the code might not be worth the performance improvement.

helix-core/src/syntax/overlay/test.rs
pascalkuthe: fixed in 87f1a17

helix-core/src/syntax/overlay/test.rs
pascalkuthe: fixed in 1f26bca

helix-term/src/ui/editor.rs
pascalkuthe: fixed in c3f652f

helix-core/src/syntax/overlay.rs
```diff
@@ -0,0 +1,254 @@
...
+use HighlightEvent::*;
+
+#[cfg(test)]
+mod test;
+
+/// Overlays multiple different highlights from `spans` onto the `HighlightEvent` stream `events`.
+///
+/// The [`Span`]s yielded by `spans` **must never overlap*  or the iterator will produce incorrect results.
+/// The [`Span`]s **must be sorted** in ascending order by their start.
+/// If multiple [`Span`]s  share the same start, the ordering is arbitrary.
```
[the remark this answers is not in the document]
pascalkuthe: fixed in c2bdbbf

helix-core/src/syntax.rs
```diff
@@ -1184,7 +1188,36 @@ struct LocalScope<'a> {
...
+                let text = self.0.source;
+                let start = text.byte_to_char(ensure_grapheme_boundary_next_byte(text, start));
+                let end = text.byte_to_char(ensure_grapheme_boundary_next_byte(text, end));
+                HighlightEvent::Source { start, end }
+            }
+            event => event,
+        };
+
+        Some(res)
+    }
```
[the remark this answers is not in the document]
pascalkuthe: > I'm concerned that helix_core is not the appropriate place to be unwrapping. HighlightIter returns Result items so that in the future we can implement cancellation of highlighting (timer based for example). There isn't a concrete plan for that AFAIK so it's hard to say how/where the Result should be handled.

I have moved unwrapping the closure back into `helix-term` in 25427ce altough it make the code simpler if we simply returned `HighlightEvent` from the `HighlightIterator` (panicking directly in the iterator) and then switched to a result once its actually required (because an `unwrap` is required everywhere the `HighlighterIterator` is used all call-sites will have to be changed in the future anyway). What do you think is better

> @archseer does the TODO for using byte slices applies anymore? I think we can just translate bytes to chars in the Iterator implementation for HighlightIter

In `markdown.rs` the highlighting directly indexes into a string and therefore uses byte indices. Using char indices and translating back would be inefficient in that case  so I am not sure there is a better solution right now.
Rewriting the rendering to use byte indices seems to amount to translating byte indices to chars anyways so there might not be a better approach.

helix-term/src/ui/editor.rs
pascalkuthe: The `Overlay` iterator already skipts empty spans anyways so the extra check here would be kind of double. 
It might be a small performance improvement by avoiding extra allocations but I doubt that would be worth it.

I think a better improvement in the future would be to implement this as a proper lazy iterator to avoid the unneeded `Vec` allocations but that's a bit more effort and there are usually not that many cursors around so its probably not worth doing in this PR (maybe in a future PR instead)

helix-term/src/ui/editor.rs
pascalkuthe: I agree that this was a bit ugly, this is one of the rare cases where the limitations of the rust type system are kind of showing (generic assoicated types would help here or generic closures in general).

I managed to implement an alternative solution in 2c6f681 that turned out quite good.
The idea is that we never return the iterator and instead call the rendering code in the same function.
An Overlay (like diagnostics, selections) are passed as parameter that implements the `HighlightOverlay`so that they are generic over the input `Iterator`. `HighlightOverlay` are composable using tuples. For example the diagnostics Layers are implemented by creating a tuple of 5 `DiagnosticOverlay`s `(hint_overlay, info_overlay, unkown_overlay, warning_overlay, error_overlay)`.

helix-term/src/ui/editor.rs
pascalkuthe: not needed anymore after 2c6f681

the-mikedavis:
I hacked together an FPS counter for the render loop: https://github.com/the-mikedavis/helix/commit/02ae7adc8a0214b105dc441a6cb2f3ea89f1e4fb and this change looks costly based on those rough numbers. I see master at 868 FPS and this branch at 805 FPS for the same code when there are no diagnostics. It seems like there is some amount of overhead for overlaying an empty vec on the highlight stream that I think can be avoided.

I have some local changes that take care of the merging of overlapping diagnostics within `Editor::doc_diagnostics_highlights` which seems to be fast; I'll make a PR soon. I think it should be possible to solve this with minimal or no changes to Merge and maybe come out with a performance boost when there are no diagnostics.

pascalkuthe:
closed in favor of #4113

Issue #4038, opened 2022-09-29
Crash on format (unwrap() in ropey library)


zensayyy:
### Summary

Helix panics on `:format`. This is the trace:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Line index out of bounds: line index 92, Rope/RopeSlice line count 87', /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.0/src/slice.rs:424:41
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: core::result::unwrap_failed
   3: ropey::slice::RopeSlice::line_to_byte
   4: helix_term::ui::editor::EditorView::doc_syntax_highlights
   5: helix_term::ui::editor::EditorView::render_view
   6: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
   7: helix_term::compositor::Compositor::render
   8: helix_term::application::Application::event_loop_until_idle::{{closure}}
   9: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  10: std::thread::local::LocalKey<T>::with
  11: tokio::park::thread::CachedParkThread::block_on
  12: tokio::runtime::thread_pool::ThreadPool::block_on
  13: tokio::runtime::Runtime::block_on
  14: hx::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

As you can see, it was caused by the library `ropey`. I may take a look at this and create a PR :)

### Reproduction Steps

1. touch a new .cpp file ( or use an existing one)
2. insert the following code (I dont remember where I got it from. It does not really make any sense)
<details>
<summary> code</summary>

```
int main() {
std::string::const_iterator it;
      std::string::const_iterator temp;

      int state = 0;

      it = uri_string.begin();
      size_t uri_len = uri_string.length();

      if (uri_len >= 7 && std::equal(it, it + 6, "wss://"))
      {
        m_secure = true;
        m_scheme = "wss";
        it += 6;
      }
      else if (uri_len >= 6 && std::equal(it, it + 5, "ws://"))
      {
        m_secure = false;
        m_scheme = "ws";
        it += 5;
      }
      else if (uri_len >= 8 && std::equal(it, it + 7, "http://"))
      {
        m_secure = false;
        m_scheme = "http";
        it += 7;
      }
[...]

PR #4043, opened 2022-09-29
added bounds check to prevent panic

zensayyy:
addresses my own issue here #4038 

simple bounds check should do the trick

### DIFF
--- a/runtime/queries/r/injections.scm
+++ b/runtime/queries/r/injections.scm
@@ -1,0 +1,2 @@
+((comment) @injection.content
+ (#set! injection.language "comment"))


the-mikedavis (changes requested):
This prevents the panic but doesn't solve the underlying issue that causes the panic.

We panic here because the `offset` points to a part of the text that doesn't exist after `:format`ing. Instead of handling the panic here, we should add a call to `view.ensure_cursor_in_view(doc, editor.config().scrolloff)` to the `:format` implementation (in this block: https://github.com/helix-editor/helix/blob/8a7a6e4cff1db3ef65a2c20e0e90c60b44516e27/helix-term/src/commands.rs#L2510-L2515)

PR #4045, opened 2022-09-30
redetect line ending when printing them

kirawi:
Previously, the displayed line ending wouldn't be reverted when a line ending change was undone. This fixes that.


kirawi:
I realize I'll need a more elegant solution to this that will run only when that particular transaction is undone. The editor will erroneously insert the old line ending even after the change was undone.

PR #4047, opened 2022-09-30
Ensure cursor in view after format

zensayyy:
addresses #4038 and "follow up" from PR #4043

credits to https://github.com/helix-editor/helix/pull/4043#pullrequestreview-1126084661 
I did not know that resetting the head would close the PR. Hope this is ok

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2504,14 +2504,16 @@ async fn make_format_callback(
 ) -> anyhow::Result<job::Callback> {
     let format = format.await?;
     let call: job::Callback = Box::new(move |editor, _compositor| {
-        let view_id = view!(editor).id;
         let scrolloff = editor.config().scrolloff;
-        let doc = doc_mut!(editor, &doc_id);
+        let doc = match editor.documents.get_mut(&doc_id) {
+            Some(doc) => doc,
+            None => return,
+        };
         let view = view_mut!(editor);
         if doc.version() == doc_version {
-            doc.apply(&format, view_id);
+            doc.apply(&format, view.id);
             view.ensure_cursor_in_view(doc, scrolloff);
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -2505,17 +2505,19 @@ async fn make_format_callback(
...
-                doc.apply(&format, view_id);
-                doc.append_changes_to_history(view_id);
-                doc.detect_indent_and_line_ending();
-                if let Modified::SetUnmodified = modified {
-                    doc.reset_modified();
-                }
-            } else {
-                log::info!("discarded formatting changes because the document changed");
+        let scrolloff = editor.config().scrolloff;
+        let doc = doc_mut!(editor, &doc_id);
```
the-mikedavis, line 2509: This macro has an `unwrap` that might cause a panic if the document has been closed after the formatting callback completes. Instead we should return early if we see that the doc is `None`:

```suggestion
        let doc = match editor.documents.get_mut(&doc_id) {
            Some(doc) => doc,
            None => return,
        };
```

helix-term/src/commands.rs
```diff
@@ -2505,17 +2505,19 @@ async fn make_format_callback(
...
+        let view = view_mut!(editor);
+        if doc.version() == doc_version {
+            doc.apply(&format, view_id);
```
the-mikedavis, line 2512: We can use `view.id` here and remove the `let view_id = view!(editor).id;` line above

the-mikedavis:
I pushed a small change to avoid computing the scrolloff when the document has been closed. The borrow-checking rules are a little tricky in this block because of the multiple mutable borrows from `editor`

the-mikedavis (approved):

the-mikedavis:
Thanks @zensayyy!

PR #4056, opened 2022-10-01
Add rainbow indentation guides

### LINKED ISSUE #4010: Smooth & colored indent-guide
Indent guides should look smoother and highlighted depending on the currently focused paragraph.

SoraTenshi:
This PR Closes #4010 

It looks something like: 
image

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -421,17 +421,26 @@ impl EditorView {
         let text_style = theme.get("ui.text");
         let whitespace_style = theme.get("ui.virtual.whitespace");
 
+        let indent_rainbow_colors = [
+            helix_view::graphics::Color::Red,
+            helix_view::graphics::Color::Yellow,
+            helix_view::graphics::Color::Green,
+            helix_view::graphics::Color::Cyan,
+            helix_view::graphics::Color::Blue,
+            helix_view::graphics::Color::Magenta,
+        ];
+
         let mut is_in_indent_area = true;
-        let mut last_line_indent_level = 0;
+        let mut last_line_indent_level: u16 = 0;
 
         // use whitespace style as fallback for indent-guide
-        let indent_guide_style = text_style.patch(
+        let mut indent_guide_style = text_style.patch(
             theme
                 .try_get("ui.virtual.indent-guide")
                 .unwrap_or_else(|| theme.get("ui.virtual.whitespace")),
         );
 
-        let draw_indent_guides = |indent_level, line, surface: &mut Surface| {
+        let mut draw_indent_guides = |indent_level, line, surface: &mut Surface| {
             if !config.indent_guides.render {
                 return;
             }
@@ -441,6 +450,11 @@ impl EditorView {
             // extra loops if the code is deeply nested.
 
             for i in starting_indent..(indent_level / tab_width as u16) {
+                if config.indent_guides.rainbow {
+                    let color_index: usize = i as usize % indent_rainbow_colors.len();
+                    indent_guide_style.fg = Some(indent_rainbow_colors[color_index]);
+                }
+
                 surface.set_string(
                     viewport.x + (i * tab_width as u16) - offset.col as u16,
                     viewport.y + line,
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -431,7 +431,7 @@ impl EditorView {
         ];
 
         let mut is_in_indent_area = true;
-        let mut last_line_indent_level: u16 = 0;
+        let mut last_line_indent_level = 0;
 
         // use whitespace style as fallback for indent-guide
         let mut indent_guide_style = text_style.patch(
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -421,15 +421,6 @@ impl EditorView {
         let text_style = theme.get("ui.text");
         let whitespace_style = theme.get("ui.virtual.whitespace");
 
-        let indent_rainbow_colors = [
-            helix_view::graphics::Color::Red,
-            helix_view::graphics::Color::Yellow,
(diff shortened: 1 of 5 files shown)


the-mikedavis:
You can cherry-pick this commit from the rainbow brackets PR that adds the ability to theme rainbows [helix-editor/helix@`c18ff11` (#2857)](https://github.com/helix-editor/helix/pull/2857/commits/c18ff1151f36fb74c94e98b7cd282f03fe6d09e8)

helix-term/src/ui/editor.rs
```diff
@@ -441,6 +441,11 @@ impl EditorView {
             // extra loops if the code is deeply nested.
 
             for i in starting_indent..(indent_level / tab_width as u16) {
+                if config.indent_guides.rainbow {
+                    let color_index: usize = i as usize % theme.rainbow_length();
+                    indent_guide_style.fg = theme.get(&format!("rainbow.{}", color_index)).fg;
```
the-mikedavis, line 446: Instead of patching only the foreground, we most likely want to patch the entire style here. We shouldn't assume that indent guides are only ever themed using foreground

s0LA1337: > Instead of patching only the foreground, we most likely want to patch the entire style here. We shouldn't assume that indent guides are only ever themed using foreground

seems reasonable, i thought about it myself and figured that there may not be a real usecase other than patching the fg.

helix-term/src/ui/editor.rs
```diff
@@ -441,6 +441,12 @@ impl EditorView {
             // extra loops if the code is deeply nested.
 
             for i in starting_indent..(indent_level / tab_width as u16) {
+                if config.indent_guides.rainbow {
+                    let color_index: usize = i as usize % theme.rainbow_length();
```
the-mikedavis, line 445:
```suggestion
                    let color_index = i as usize % theme.rainbow_length();
```

helix-term/src/ui/editor.rs
```diff
@@ -441,6 +441,12 @@ impl EditorView {
...
+                    indent_guide_style =
+                        indent_guide_style.patch(theme.get(&format!("rainbow.{}", color_index)));
+                }
```
the-mikedavis, line 448: The new `mut`s in this change can be eliminated

```suggestion
                let style = if config.indent_guides.rainbow {
                    let color_index = i as usize % theme.rainbow_length();
                    indent_guide_style.patch(theme.get(&format!("rainbow.{}", color_index)))
                } else {
                    indent_guide_style
                };
```

Saving that patched style in `indent_guide_style` has a problem that you'll patch each color of the rainbow over the last colors (for example, the first indent guide is `red + indent_guide_style`, then the second is `yellow + red + indent_guide_style`, etc. when we only want `rainbow_color + indent_guide_style`)

helix-term/src/ui/editor.rs
```diff
@@ -441,11 +441,18 @@ impl EditorView {
             // extra loops if the code is deeply nested.
 
             for i in starting_indent..(indent_level / tab_width as u16) {
+                let style = if config.indent_guides.rainbow {
+                    let color_index = i as usize % theme.rainbow_length();
+                    indent_guide_style.patch(theme.get(&format!("rainbow.{}", color_index)))
```
the-mikedavis, line 446: I think we can speed up the calculation of the rainbow color: `Theme` can have a new associated function that returns the style for the given usize

```rust
pub fn get_rainbow(&self, index: usize) -> Style {
    self.highlights[index % self.rainbow_length]
}
```

```rust
indent_guide_style.patch(theme.get_rainbow(i));
```

Since rainbow highlights are the first `self.rainbow_length` elements of the `self.highlights` Vec

s0LA1337: > I think we can speed up the calculation of the rainbow color: `Theme` can have a new associated function that returns the style for the given usize
> 
> ```rust
> pub fn get_rainbow(&self, index: usize) -> Style {
>     self.highlights[index % self.rainbow_length]
> }
> ```
> 
> ```rust
> indent_guide_style.patch(theme.get_rainbow(i));
> ```
> 
> Since rainbow highlights are the first `self.rainbow_length` elements of the `self.highlights` Vec

This serves a lot of benefits.
1. no string comparisons
2. better readability
3. you may don't have to read the code and can rely on the functions name if you want to use a rainbow

poliorcetics:
@s0LA1337, if I can ask, why did you close ?

s0LA1337:
> @s0LA1337, if I can ask, why did you close ?

I think I did an accident here.
I had somehow a corruption of my fork, and had to close the repo.
Thanks for reminding me, will reopen it.

helix-term/src/ui/editor.rs
```diff
@@ -444,11 +444,17 @@ impl EditorView {
             // extra loops if the code is deeply nested.
 
             for i in starting_indent..(indent_level / tab_width as u16) {
+                let style = if config.indent_guides.rainbow {
+                    indent_guide_style.patch(theme.get_rainbow(i as usize))
```
the-mikedavis, line 448: This may look odd if `rainbow = true` is set as well as `skip-levels = n` for some non-zero n. Currently if you skip one level, the first level uses the second color of the rainbow but I think we should start the rainbow on the first indent guide

s0LA1337: At first glance this looks like it would be weird, but this behaviour causes an interesting side-effect together with your rainbow brackets.
It matches the color based on indentation.
consider this:
image

if i set the linenumber skip for the indent guides now to let's say 2, it would still line up
image

with the fix, this alignment would be lost.

SoraTenshi:
Since i've gotten a bit of feedback from some people saying, that it may be too bright, i added a dimming feature.

With dim:
image

Without dim:
image

SoraTenshi:
Since this pr is kind of broken for me (not updating...)
Please check out this (new) PR: #4493

PR #4062, opened 2022-10-01
Add pseudo_pending for t/T/f/F

### LINKED ISSUE #4050: Key not shown in corner for find commands
The `f`, `t`, `F`, and `T` commands do not display the key in the lower right corner when they are pending input.

mbStavola:
Closes #4050.

Adds support for the following commands:

- Find 'till next char: *t*
- Find 'till previous char: *T*
- Find next char: *f*
- Find previous char: *F*
- Extend 'till next char: *t*
- Extend 'till previous char: *T*
- Extend next char: *f*
- Extend previous char: *F*

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1091,7 +1091,7 @@ fn will_find_char<F>(
     search_fn: F,
     inclusive: bool,
     extend: bool,
-    pseudo_pending: Option<String>,
+    pseudo_pending: &str,
 ) where
     F: Fn(RopeSlice, char, usize, usize, bool) -> Option<usize> + 'static,
 {
@@ -1099,7 +1099,7 @@ fn will_find_char<F>(
     // Would be nice to carry over.
     let count = cx.count();
 
-    cx.editor.pseudo_pending = pseudo_pending;
+    cx.editor.pseudo_pending = Some(pseudo_pending.to_string());
 
     // need to wait for next key
     // TODO: should this be done by grapheme rather than char?  For example,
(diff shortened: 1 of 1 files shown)


ath3:
Its possible to remap these commands to other keys, in which case the pending key shown is no longer the actually pressed key.

the-mikedavis:
That's a known limitation of pseudo-pending. It exists for `mi`/`ma` as well for example and is out-of-scope to fix for this PR

helix-term/src/commands.rs
```diff
@@ -1200,35 +1209,35 @@ fn find_prev_char_impl(
...
 }
 
 fn find_next_char(cx: &mut Context) {
-    will_find_char(cx, find_next_char_impl, true, false)
+    will_find_char(cx, find_next_char_impl, true, false, Some("f".to_string()))
 }
 
 fn extend_till_char(cx: &mut Context) {
-    will_find_char(cx, find_next_char_impl, false, true)
+    will_find_char(cx, find_next_char_impl, false, true, Some("vt".to_string()))
```
the-mikedavis, line 1220: The extend versions should not use "v". The "v" puts you in select mode but isn't considered to be a pending key in the same way that "t" is.

helix-term/src/commands.rs
```diff
@@ -1086,18 +1086,27 @@ fn extend_next_long_word_end(cx: &mut Context) {
...
 }
 
-fn will_find_char<F>(cx: &mut Context, search_fn: F, inclusive: bool, extend: bool)
-where
+fn will_find_char<F>(
+    cx: &mut Context,
+    search_fn: F,
+    inclusive: bool,
+    extend: bool,
+    pseudo_pending: Option<String>,
```
the-mikedavis, line 1094: This can be a `&'static str` and then in the line below you can say `cx.editor.pseudo_pending = Some(key.to_string())`.

mbStavola: Might be better to avoid specifying `'static` here as there isn't a hard requirement for it, right? I imagine at some point, to @ath3 's point, we might want to support displaying the remapped keys (which wouldn't be `'static`).

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4072, opened 2022-10-02
Add support for D language

### LINKED ISSUE #3832: Add support for D
There is an LSP server called serve-d, but for whatever reason I had very little luck in getting it to work with Helix.

It would be nice if this worked more or less out of the box.

A tree-sitter grammar is here: https://github.com/CyberShadow/tree-sitter-d

ErdoganSeref:
The change resolves #3832. I need help to complete the tree-sitter configuration.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -15,6 +15,7 @@
 | cpp | ✓ | ✓ | ✓ | `clangd` |
 | css | ✓ |  |  | `vscode-css-language-server` |
 | cue | ✓ |  |  | `cuelsp` |
+| d | ✓ |  |  |  | `serve-d` | 
 | dart | ✓ |  | ✓ | `dart` |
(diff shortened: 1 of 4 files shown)


the-mikedavis:
That repository git-ignores the generated `parser.c` and `parser.h` files which we need to compile the parser. See https://github.com/CyberShadow/tree-sitter-d/issues/6

You can get around it by creating a fork of the repository, removing [these lines](https://github.com/CyberShadow/tree-sitter-d/blob/c2fbf21bd3aa45495fe13247e040ad5815250032/.gitignore#L5-L7), running `tree-sitter generate`, and committing the generated files.

the-mikedavis:
It looks like the author of #3832 is working on rewriting or improving tree-sitter-d, so we might be able to take advantage of that new grammar when they are finished

the-mikedavis:
https://github.com/helix-editor/helix/issues/3832#issuecomment-1277139601

PR #4077, opened 2022-10-02
Automatically track pseudo-pending text

the-mikedavis:
This change automatically tracks pending text for for commands which use on-next-key callbacks. For example, `t` will await the next key event and "t" will be shown in the bottom right-hand corner to show that we're in a pending state.

Previously, the text for these on-next-key commands needed to be hard-coded into the command definition which had some drawbacks:

* It was easy to forget to write and clear the pending text.
* If a command was remapped in a custom config, the pending text would still show the old key.

With this change, pending text is automatically tracked based on the key events that lead to the command being executed. This works even when the command is remapped in config and when the on-next-key callback is nested under some key sequence (for example `mi`).

Closes https://github.com/helix-editor/helix/pull/3381
Closes https://github.com/helix-editor/helix/issues/3223

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1086,27 +1086,18 @@ fn extend_next_long_word_end(cx: &mut Context) {
     extend_word_impl(cx, movement::move_next_long_word_end)
 }
 
-fn will_find_char<F>(
-    cx: &mut Context,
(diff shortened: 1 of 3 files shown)


dead10ck (approved):
Very nice, looks great to me. Thanks for cleaning this up.

Issue #3849, opened 2022-09-15
Broken syntax highlighting in Markdown


David-Else:
### Summary

@MDeiml  In the `keymap.md` file you can see the markdown highlighting is broken, is seems fine in Neovim

# Helix

![Screenshot from 2022-09-15 14-37-50](https://user-images.githubusercontent.com/12832280/190419392-caf9a111-c99f-4d0f-b7f8-f435421abb98.png)

![Screenshot from 2022-09-15 14-38-01](https://user-images.githubusercontent.com/12832280/190419727-fea1b1f9-90c2-4b58-bb84-9196af291e17.png)

# Neovim
![Screenshot from 2022-09-15 14-37-45](https://user-images.githubusercontent.com/12832280/190419532-e84149bf-61d8-45a5-9029-d7c366489331.png)

### Reproduction Steps

Load https://raw.githubusercontent.com/helix-editor/helix/master/book/src/keymap.md to see the problem

### Helix log

n/a

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.08.1 (b5be72bf) < latest today

Chickenkeeper:
Updating the version of tree-sitter-markdown in languages.toml lto the latest commit helps with this issue, so that might tide things over for now. It doesn't completely fix it though, so more work is still needed there.

David-Else:
@Chickenkeeper Thanks for your help! Updating to the latest commit does improve things, but I am not sure if the queries also need updating? I don't know much about Treesitter.

I have opened an issue https://github.com/MDeiml/tree-sitter-markdown/issues/62, please add any more `weirdness caused by escape sequences` or extra details you can think of. Cheers!

Chickenkeeper:
@David-Else Ha, will do. I've looked at the queries and nothing immediately stood out, but it might take a bit of experimenting.

MDeiml:
Updating to the latest commit helped, because I added support for tables. The queries should be fine. You would only need to change them if you wanted some special highlighting for tables.

MDeiml:
Other problems (like thos discussed in https://github.com/MDeiml/tree-sitter-markdown/issues/62) might be problems with the parser, so feel free to report them there.

MDeiml:
I added a nice new fix so most of what is talked about here works now. I'll make a PR to update the grammar.

David-Else:
Brilliant, it is fixed :)

PR #4078, opened 2022-10-02
Update treesitter markdown

David-Else:
Fixes broken table highlighting: https://github.com/helix-editor/helix/issues/3849

[MDeiml](https://github.com/MDeiml), the creator of the treesitter syntax said there was no need to update any queries:

>Updating to the latest commit helped, because I added support for tables. The queries should be fine. You would only need to change them if you wanted some special highlighting for tables.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -923,7 +923,7 @@ grammar = "markdown_inline"
 
 [[grammar]]
 name = "markdown_inline"
-source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "142a5b4a1b092b64c9f5db8f11558f9dd4009a1b", subpath = "tree-sitter-markdown-inline" }
+source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "d5740f0fe4b8e4603f2229df107c5c9ef5eec389", subpath = "tree-sitter-markdown-inline" }
 
 [[language]]
 name = "dart"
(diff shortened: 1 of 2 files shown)


languages.toml
```diff
@@ -911,7 +911,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "markdown"
-source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "142a5b4a1b092b64c9f5db8f11558f9dd4009a1b", subpath = "tree-sitter-markdown" }
+source = { git = "https://github.com/MDeiml/tree-sitter-markdown", rev = "d5740f0fe4b8e4603f2229df107c5c9ef5eec389", subpath = "tree-sitter-markdown" }
```
the-mikedavis, line 914: We probably want to keep `markdown.inline` at the same revision. It looks like the update that added table support also touched the grammar for inline markdown (the next `[[grammar]]` entry)

the-mikedavis (commented):
We might need a new stanza in `runtime/queries/markdown/injections.scm`

```scm
((pipe_table_cell) @injection.content (#set! injection.language "markdown.inline") (#set! injection.include-unnamed-children))
```

This makes `keymap.md` look perfect to me 👍

Chickenkeeper:
I'm glad I checked noticed this PR, I was literally minutes away from submitting this exact change as part of another PR! No inconvenience though, just a funny coincidence

David-Else:
@the-mikedavis I have done as you recommended, and it looks great.

archseer (approved):

PR #4079, opened 2022-10-02
Fix C++ syntax highlighting, and improve C & WGSL highlighting

### LINKED ISSUE #3788: C++ syntax highlighting incorrectly coloring std functions
### Summary

When you modify the existing "onedark" theme to highlight variables as red, it incorrectly highlighted std functions as variables as well. 

In the default "onedark" theme the code looks like this: 
Screen Shot 2022-09-10 at 7 14 29 PM

Which is not actually true to what it should look like, as in atom variables are red. When you change the variables attribute in the theme file it looks like this:
Screen Shot 2022-09-10 at 7 14 10 PM

The variables are now correctly highlighted red, but std::getline() and std::tolower() are now erroneously colored. They should be highlighted blue but they are not detected as functions. 

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

macOS

### Terminal Emulator

iTerm2

### Helix Version

Helix 22.08.1

Chickenkeeper:
image

Fixes #3788, and adds various missing operators & highlights and organises keywords better for C, C++, and WGSL. Also contains a couple of fixes to WGSL highlighting to prevent text from being highlighted as an error as it's being typed, and remove edge cases where `@variable` text modifiers could leak to other theme keys.

(theme is onedark but modified to make all variables red as shown in the issue)

### DIFF
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -105,9 +105,6 @@
 (number_literal) @constant.numeric.integer
 (char_literal) @constant.character
 
-((identifier) @constant
-  (#match? @constant "^[A-Z][A-Z\\d_]*$"))
-
 (call_expression
   function: (identifier) @function)
 (call_expression
@@ -123,36 +120,16 @@
     declarator: (identifier) @variable.parameter))
 (preproc_function_def
   name: (identifier) @function.special)
-; (preproc_arg) @error
 
 (field_identifier) @variable.other.member
 (statement_identifier) @label
-(struct_specifier) @type
-(type_definition) @type
 (type_identifier) @type
 (primitive_type) @type.builtin
 (sized_type_specifier) @type
 
-(init_declarator (identifier) @variable)
-(binary_expression left: (identifier) @variable)
-(binary_expression right: (identifier) @variable)
-(compound_statement (declaration (identifier) @variable))
-(for_statement (declaration (identifier) @variable))
-(field_expression (identifier) @variable)
-(pointer_declarator (identifier) @variable)
-(pointer_expression (identifier) @variable)
-(assignment_expression (identifier) @variable)
-(unary_expression (identifier) @variable)
-(sizeof_expression (parenthesized_expression (identifier) @type))
-(parenthesized_expression (identifier) @variable)
-(initializer_list (identifier) @variable)
-(initializer_pair (identifier) @variable)
-(return_statement (identifier) @variable)
-(subscript_expression (identifier) @variable)
-(cast_expression (identifier) @variable)
-(update_expression (identifier) @variable)
-(conditional_expression (identifier) @variable)
+((identifier) @constant
+  (#match? @constant "^[A-Z][A-Z\\d_]*$"))
 
-; (identifier) @error
+(identifier) @variable
 
 (comment) @comment
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -96,7 +96,7 @@
 
 [(true) (false)] @constant.builtin.boolean
 
-(enumerator) @type.enum.variant
+(enumerator name: (identifier) @type.enum.variant)
 
 (string_literal) @string
 (system_lib_string) @string
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -1,17 +1,22 @@
-(storage_class_specifier) @keyword.storage
+[
+  "sizeof"
+] @keyword
 
 [
-  "register"
-  "const"
   "enum"
+  "struct"
+  "union"
+] @keyword.storage.type
+
+[
+  "const"
   "extern"
   "inline"
-  "sizeof"
-  "struct"
+  "register"
   "typedef"
-  "union"
   "volatile"
-] @keyword
+  (storage_class_specifier)
+] @keyword.storage.modifier
 
 [
   "for"
@@ -46,25 +51,43 @@
 ] @keyword.directive
 
 [
-  "--"
-  "-"
-  "-="
-  "->"
-  "="
-  "!="
-  "*"
-  "&"
-  "&&"
   "+"
+  "-"
+  "*"
+  "/"
   "++"
-  "+="
-  "<"
+  "--"
+  "%"
   "=="
+  "!="
   ">"
-  "||"
+  "<"
   ">="
   "<="
+  "&&"
+  "||"
+  "!"
+  "&"
+  "|"
+  "^"
+  "~"
+  "<<"
+  ">>"
+  "="
+  "+="
+  "-="
+  "*="
+  "/="
+  "%="
+  "<<="
+  ">>="
+  "&="
+  "^="
+  "|="
+  "->"
   "::"
+  "?"
+  "..."
 ] @operator
 
 ["," "." ":" ";"] @punctuation.delimiter
@@ -87,14 +110,16 @@
 (call_expression
   function: (field_expression
     field: (field_identifier) @function))
-(function_declarator) @function
 (function_declarator
-  parameters: (parameter_list) @variable.parameter)
+  declarator: (identifier) @function)
+(parameter_declaration
+  declarator: (identifier) @variable.parameter)
+(parameter_declaration
+  (pointer_declarator
+    declarator: (identifier) @variable.parameter))
 (preproc_function_def
   name: (identifier) @function.special)
 
-(compound_statement) @variable
-(init_declarator) @variable
 (field_identifier) @variable.other.member
 (statement_identifier) @label
 (struct_specifier) @type
@@ -105,5 +130,7 @@
 
 ((identifier) @constant
  (#match? @constant "^[A-Z][A-Z\\d_]*$"))
+
+(identifier) @variable
 
 (comment) @comment
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -105,13 +105,17 @@
 (number_literal) @constant.numeric.integer
 (char_literal) @constant.character
 
+((identifier) @constant
+  (#match? @constant "^[A-Z][A-Z\\d_]*$"))
+
 (call_expression
   function: (identifier) @function)
 (call_expression
   function: (field_expression
     field: (field_identifier) @function))
+(call_expression (argument_list (identifier) @variable))
 (function_declarator
-  declarator: (identifier) @function)
+  declarator: [(identifier) (field_identifier)] @function)
 (parameter_declaration
   declarator: (identifier) @variable.parameter)
 (parameter_declaration
@@ -119,6 +123,7 @@
     declarator: (identifier) @variable.parameter))
 (preproc_function_def
   name: (identifier) @function.special)
+; (preproc_arg) @error
 
 (field_identifier) @variable.other.member
 (statement_identifier) @label
@@ -128,9 +133,26 @@
 (primitive_type) @type.builtin
 (sized_type_specifier) @type
 
-((identifier) @constant
- (#match? @constant "^[A-Z][A-Z\\d_]*$"))
+(init_declarator declarator: (identifier) @variable)
+(binary_expression left: (identifier) @variable)
+(binary_expression right: (identifier) @variable)
+(compound_statement (declaration (identifier) @variable))
+(for_statement (declaration (identifier) @variable))
+(field_expression (identifier) @variable)
+(pointer_declarator (identifier) @variable)
+(pointer_expression (identifier) @variable)
+(assignment_expression (identifier) @variable)
+(unary_expression (identifier) @variable)
+(sizeof_expression (parenthesized_expression (identifier) @type))
+(parenthesized_expression (identifier) @variable)
+(initializer_list (identifier) @variable)
+(initializer_pair (identifier) @variable)
+(return_statement (identifier) @variable)
+(subscript_expression (identifier) @variable)
+(cast_expression (identifier) @variable)
+(update_expression (identifier) @variable)
+(conditional_expression (identifier) @variable)
 
-(identifier) @variable
+; (identifier) @error
 
 (comment) @comment
--- a/runtime/queries/c/highlights.scm
+++ b/runtime/queries/c/highlights.scm
@@ -88,6 +88,8 @@
   "?"
 ] @operator
 
+(conditional_expression ":" @operator)
+
 "..." @punctuation
 
 ["," "." ":" ";" "->" "::"] @punctuation.delimiter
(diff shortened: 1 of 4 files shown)


runtime/queries/c/highlights.scm
```diff
@@ -73,20 +87,23 @@
...
     field: (field_identifier) @function))
+(function_declarator) @function
 (function_declarator
-  declarator: (identifier) @function)
+  parameters: (parameter_list) @variable.parameter)
 (preproc_function_def
   name: (identifier) @function.special)
 
+(compound_statement) @variable
+(init_declarator) @variable
```
the-mikedavis, line 97: These two are covered by the deleted

```scm
(identifier) @variable
```

stanza below. That should be restored and these two lines should be removed. With these, entire function bodies and initialization statements inherit variable theming (to see the effect, set "variable" to have `modifiers = [ "italics" ]` in your theme).

runtime/queries/cpp/highlights.scm
```diff
@@ -12,13 +12,6 @@
 (template_method
   name: (field_identifier) @function)
 
-(template_function
-  name: (identifier) @function)
```
the-mikedavis, line 16: Without this stanza, template function calls are highlighted as variables

```cpp
a<T>();
```

runtime/queries/cpp/highlights.scm
```diff
@@ -12,13 +12,6 @@
...
-
-(function_declarator
-  declarator: (qualified_identifier
-    name: (identifier) @function))
```
the-mikedavis, line 20: Without this stanza, some function declarations are highlighted as properties instead:

```cpp
class C : D {
  void efg() {
  }
}

**Review by the-mikedavis — CHANGES_REQUESTED** [review]

File: runtime/queries/c/highlights.scm [hunk 4]
```diff
@@ -73,20 +87,23 @@
...
 (field_identifier) @variable.other.member
 (statement_identifier) @label
+(struct_specifier) @type
+(type_definition) @type
 (type_identifier) @type
-(primitive_type) @type
+(primitive_type) @type.builtin
```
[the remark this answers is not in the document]
Chickenkeeper: The Helix docs list that key as being for primitive/built in types, it seemed like that was its intended use

runtime/queries/cpp/highlights.scm
Chickenkeeper: I removed this and the other stanza because they are duplicates of existing ones in the same file, and their removal didn't seem to make any difference. But if the duplicates actually do make a difference I'll put them back

runtime/queries/c/highlights.scm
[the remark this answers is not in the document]
the-mikedavis: Oh whoops, I misread the diff, this is good 👍

Chickenkeeper:
@the-mikedavis I've added most of the changes. I'm still a little unclear on the last two though, is there a reason the duplicates are necessary or is it fine to just have one instance of each?

runtime/queries/c/highlights.scm
the-mikedavis: You can move the `; inherits: c` to the bottom of `runtime/queries/cpp/highlights.scm` so that those stanzas have lower precedence than any stanzas in the cpp highligts

runtime/queries/cpp/highlights.scm
the-mikedavis: Oh whoops I didn't see the duplicates 😅. Yeah these are ok to remove 👍

runtime/queries/c/highlights.scm
the-mikedavis: Doing that ☝️ should also fix this case:

```cpp
this->b<C, D>();
//    ^ @variable.other.member but should be @function
```

the-mikedavis:
The removal of duplicates are ok 👍. It looks like they're duplicated upstream in tree-sitter-cpp but it seems like a silly mistake looking at the git history.

runtime/queries/c/highlights.scm
Chickenkeeper: Ah I didn't realise that was possible, that's much simpler. Done.

Chickenkeeper:
@the-mikedavis I've added the rest of the changes. Let me know if there's anything else

Chickenkeeper:
@Triton171 Added

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,85 @@
...
 
 (function_declarator
   declarator: (field_identifier) @function)
 
 ; Types
 
-((namespace_identifier) @type
- (#match? @type "^[A-Z]"))
-
+(namespace_identifier) @namespace
```
the-mikedavis, line 27: We can separate out namespaces from types and add some stanzas for some uses:

```scm
; Namespaces

(namespace_definition name: (identifier) @namespace)
(using_declaration (identifier) @namespace)
(namespace_identifier) @namespace

; Types

(auto) @type
```

I think we can also add

```scm
(qualified_identifier name: (identifier) @type)
```

to the types section for namespaced enum variants

runtime/queries/c/highlights.scm
```diff
@@ -1,61 +1,98 @@
...
+  "::"
+  "?"
+  "..."
+] @operator
+
+["," "." ":" ";"] @punctuation.delimiter
+
+["(" ")" "[" "]" "{" "}"] @punctuation.bracket
 
 [(true) (false)] @constant.builtin.boolean
```
the-mikedavis, line 98: On the next line there's a stanza

```
(enumerator) @type.enum.variant
```

which I think should be

```
(enumerator name: (identifier) @type)
```

to fix highlighting for enum variants like in [this block](https://github.com/mawww/kakoune/blob/24d6072353f7c7e7cac84b4eb085036a8c852f96/src/normal.cc#L32-L37)

Chickenkeeper: Also I enjoy the use of Kakoune's source code there 🤭

runtime/queries/cpp/highlights.scm
Chickenkeeper: These seem good, though I'd also suggest what I mentioned below about the key being @type.enum.variant instead of @type for the qualified_identifier stanza

sam-mccall: `(using_declaration (identifier) @namespace)` is not correct, it will match `using bar` where bar is either a typename or a value.

The following works for me:
```
(using_declaration ("using" "namespace" (identifier) @namespace))
(using_declaration ("using namespace" (qualified_identifier name: (identifier) @namespace)))
```

runtime/queries/c/highlights.scm
```diff
@@ -1,65 +1,102 @@
...
+  "typedef"
+  "volatile"
+  (storage_class_specifier)
+] @keyword.storage.modifier
+
+[
+  "for"
+  "do"
+  "while"
+  "break"
```
sam-mccall, line 25: `break` is for `switch` too.
The examples in the docs suggest only the loop itself is @keyword.control.repeat.
Maybe `break`/`continue` should be `keyword.control`?

runtime/queries/c/highlights.scm
```diff
@@ -1,65 +1,102 @@
...
- "#define"
- "#elif"
- "#else"
- "#endif"
- "#if"
- "#ifdef"
- "#ifndef"
- "#include"
- (preproc_directive)
+  "goto"
```
sam-mccall, line 30: is `goto` conditional?

runtime/queries/c/highlights.scm
```diff
@@ -1,65 +1,102 @@
...
+  "if"
+  "else"
+  "switch"
+  "case"
+  "default"
```
sam-mccall, line 35: this will result in slight mishighlighting in C++:
`struct X { X() = default; };`

You could override it with 
```
(default_method_clause "default" @keyword) ; not @keyword.control
```

runtime/queries/c/highlights.scm
```diff
@@ -1,65 +1,102 @@
...
+  "%="
+  "<<="
+  ">>="
+  "&="
+  "^="
+  "|="
+  "->"
+  "::"
+  "?"
+  "..."
```
sam-mccall, line 90: is "..." an operator? maybe it fits better in `@punctuation`?

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,90 @@
...
+  "new"
+  "requires"
+  "throw"
+  "try"
+  "typename"
+  "using"
+] @keyword
+
+"<=>" @operator
+
```
sam-mccall, line 60: would you mind adding template brackets? Distinguishing these from operators can be nice for readability.

```
(template_argument_list (["<" ">"] @punctuation.bracket))
(template_parameter_list (["<" ">"] @punctuation.bracket))
```

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,90 @@
...
+  "or"
+  "or_eq"
+  "xor"
+  "xor_eq"
+] @keyword.operator
+
+[
+  "class"
+  "namespace"
+  (auto)
```
sam-mccall, line 77: `auto` here seems surprising, as it represents a type rather than a type category. (I would have expected it as @type).

In any case, where `auto` goes, `"decltype"` should probably go too.

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,90 @@
...
+  "co_yield"
+  "concept"
+  "consteval"
+  "constinit"
+  "delete"
+  "final"
+  "noexcept"
+  "new"
+  "requires"
+  "throw"
```
sam-mccall, line 53: throw, catch, try, and noexcept can be @keyword.exception

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,90 @@
...
-"typename" @keyword
-"using" @keyword
-"virtual" @keyword
+[
+  "catch"
+  "co_await"
+  "co_return"
+  "co_yield"
+  "concept"
+  "consteval"
```
sam-mccall, line 46: consteval and constinit here prevents them from matching against the more specific @keyword.storage.modifier below

runtime/queries/cpp/highlights.scm
```diff
@@ -12,56 +10,90 @@
...
+  "constinit"
+  "delete"
+  "final"
+  "noexcept"
+  "new"
+  "requires"
+  "throw"
+  "try"
+  "typename"
```
sam-mccall, line 55: `typename` may be a better fit together with `class` in `@keyword.storage.type`.

(It seems strange for `template <typename>` vs `template <class>` to highlight differently, and ISTM the use in dependent names is kind of a wash)

sam-mccall (commented):
Thanks @Chickenkeeper for working on this!
I've been working on a version myself (should have checked PRs first), and have a few suggestions.
But I'm just "some guy" - if you'd rather get this landed and maintainers are happy then feel free to ignore them.

(I'll probably send a followup PR once this gets merged in any case: there's some stuff I'd like to add like distinguishing between operator-like and declarator-like uses of `&` - that seems outside the scope of this patch)

Chickenkeeper:
@sam-mccall Thanks for the input! I'm glad more people have been working on syntax highlighting. I'll take a look at your suggestions and see what I can do.

runtime/queries/c/highlights.scm
Chickenkeeper: I think moving them is fine for now. Ideally I'd like to keep `continue` in `@keyword.control.repeat` since it isn't used as part of `switch`'s, and handle `break` as either `@keyword.control.conditional` or`@keyword.control.repeat` depending on whether it's in a `switch` or not, but that would take a bit more care.

runtime/queries/c/highlights.scm
Chickenkeeper: Good point, it can be used in various different ways so maybe it's best in `@keyword.control`?

runtime/queries/cpp/highlights.scm
Chickenkeeper: I didn't think `auto` should be `@type` directly because to me it represents something that will figure out which type it represents, but isn't name of a type itself, like an anonymous class/struct keyword. Seeing `auto` coloured as a type occasionally makes me forget there isn't (and shouldn't be) a specific struct/class named `auto`.

sam-mccall: TL;DR: I think `@type` *is* technically correct (and it fits my mental model better) but this isn't a big deal, go with whatever you think is most useful. Unneccesary but maybe interesting details below.

---

Microsoft is wrong I'm afraid (or rather, glossing over details). [The spec says](https://eel.is/c++draft/dcl.spec.auto.general#1)

> A [placeholder-type-specifier](https://eel.is/c++draft/dcl.spec.auto.general#nt:placeholder-type-specifier) [e.g auto] designates a [placeholder type](https://eel.is/c++draft/dcl.spec.auto.general#def:type,placeholder) that will be replaced later by deduction from an initializer[.](https://eel.is/c++draft/dcl.spec.auto.general#1.sentence-1)

The reason that `auto` can't be used in casts etc [is rather](https://eel.is/c++draft/dcl.spec.auto.general#6)

> A program that uses a placeholder type in a context not explicitly allowed in [[dcl.spec.auto]](https://eel.is/c++draft/dcl.spec.auto) is ill-formed[.](https://eel.is/c++draft/dcl.spec.auto.general#6.sentence-1)

runtime/queries/c/highlights.scm
sam-mccall: Sure. I do think `@keyword.control.repeat` should mean something consistent though: of the languages that have queries for it, it looks like ~25 restrict it to the syntax of the loop itself, while ~2 also highlight loop-related control flow like break/continue.

sam-mccall (approved):
Thank you! Looks good from my side, looking forward to seeing this!

runtime/queries/cpp/highlights.scm
Chickenkeeper: That fair, I've changed it back to `@type` for now unless a better reason to change it comes up

the-mikedavis:
What do you think about adding

```
(conditional_expression ":" @operator)
```

to the C highlights? That way `?` and `:` have the same highlight in a ternary expression

Chickenkeeper:
> What do you think about adding
> 
> ```
> (conditional_expression ":" @operator)
> ```
> 
> to the C highlights? That way `?` and `:` have the same highlight in a ternary expression

That's a good catch. It looks more consistent that way

the-mikedavis (approved):
Great work! C/C++ highlights have been needing some polish for a while and this is a big improvement 🎉

Issue #3405, opened 2022-08-13
Add `goto_next_diag_error`/`goto_prev_diag_error`


AceofSpades5757:
#### Describe your feature request

Wanted to continue the conversation on going to the next error. I found this _extremely_ useful in another editor and I get really annoyed because I usually have a lot of diagnostic points while I'm working out an issue.

[Original Issue](https://github.com/helix-editor/helix/issues/1929)

For reference: [Pull Request for next diagnostic](https://github.com/helix-editor/helix/pull/142/files)

EpocSquadron:
This already exists for the current document with `]d` and `[d`. There is also document and workspace diagnostic pickers as of #2013

the-mikedavis:
If I understand correctly, this is only about showing the next error diagnostic, right? `]d`/`[d` will bring you to the next diagnostic which could be info/warn/error

EpocSquadron:
That makes a lot more sense now!

AceofSpades5757:
Yeah, the diagnostics are great and all, but I'm usually trying to get to the next error first, then I can worry about the diagnostic details.

94bryanr:
I would also like to see this. Using the diagnostic picker in a workspace with hundreds/thousands of warnings requires a lot of scrolling around to find the next error.

I think having the diagnostics sorted by severity with errors at the top would make a lot of sense, as finding the next error (arguably the most common task you will be doing when opening the picker) would only require one extra keypress. Then within the severity buckets we could just continue to use the current filename sorting.

cgahr:
There is also #6331, trying to implement the same thing.

PR #4080, opened 2022-10-02
Adding 'goto error' support for open documents

Zeke-D:
The point of this is unobtrusively add support for quickly navigating to errors detected by the LSP. This integrates error navigation into the "goto" menu with keypresses "g,".


Omnikar:
I haven't looked at the code but based on your description this sounds like `[d`, `]d`, and `<space>g`?

the-mikedavis:
I think this is part of https://github.com/helix-editor/helix/issues/3405 instead

Zeke-D:
D'OH!!

I have been using helix for the better part of 3 months and have only now from your comment discovered the diagnostics panel. That's insane, thank you!

PR #4081, opened 2022-10-03
Exit gracefully when close operation fails

dead10ck:
If the close method fails, the editor will quit before restoring the terminal. This causes the shell to break if, e.g. the LS times out shutting down.

This fixes this by always restoring the terminal after closing, and printing out a message to stderr if there is an error.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -866,8 +866,15 @@ impl Application {
         }));
 
         self.event_loop(input_stream).await;
-        self.close().await?;
+
+        let err = self.close().await.err();
+
         restore_term()?;
+
+        if let Some(err) = err {
+            self.editor.exit_code = 1;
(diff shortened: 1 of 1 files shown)


helix-term/src/application.rs
```diff
@@ -866,9 +866,16 @@ impl Application {
         }));
 
         self.event_loop(input_stream).await;
-        self.close().await?;
+
+        let err = self.close().await.err();
```
Omnikar, line 870: Small nitpick: shouldn't this be named `result` or `res`, not `err`?

dead10ck: It's not a result; see [`Result::err`](https://doc.rust-lang.org/std/result/enum.Result.html#method.err).

Omnikar: Ah, right.

PR #4084, opened 2022-10-03
Add cursorcolumn

### LINKED ISSUE #4015: Add cursorcolumn
Similar to vim's cursor column.

**EDIT:** Cursor column is similar to cursor line. The only difference is that the cursor column is vertical.
This blog describes them both https://medium.com/@venumadhav888/highlighting-cursor-line-and-column-in-vi-editor-d5584b8d2eca

A-Walrus:
Closes #4015.

Currently uses the same theme keys as cursorline, but they could be separated if that would be preferred.
image

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -13,7 +13,7 @@ use helix_core::{
     movement::Direction,
     syntax::{self, HighlightEvent},
     unicode::width::UnicodeWidthStr,
-    LineEnding, Position, Range, Selection, Transaction,
+    visual_coords_at_pos, LineEnding, Position, Range, Selection, Transaction,
 };
 use helix_view::{
     document::{Mode, SCRATCH_BUFFER_NAME},
@@ -117,6 +117,9 @@ impl EditorView {
 
         if is_focused && editor.config().cursorline {
             Self::highlight_cursorline(doc, view, surface, theme);
+        }
+        if is_focused && editor.config().cursorcolumn {
+            Self::highlight_cursorcolumn(doc, view, surface, theme);
         }
 
         let highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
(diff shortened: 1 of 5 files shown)


pickfire (approved):
Looks good to me although I don't think I will be using it very soon myself.

helix-term/src/ui/editor.rs
```diff
@@ -820,6 +823,45 @@ impl EditorView {
...
+    pub fn highlight_cursorcolumn(
+        doc: &Document,
+        view: &View,
+        surface: &mut Surface,
+        theme: &Theme,
+    ) {
+        let text = doc.text().slice(..);
+
+        let primary_style = theme.get("ui.cursorline.primary");
+        let secondary_style = theme.get("ui.cursorline.secondary");
```
the-mikedavis, line 836: I'm split on whether these should be specific to cursorcolumn too. We could make them `ui.cursorcolumn.{primary.secondary}` and have them fall back to cursorline themes?

helix-term/src/ui/editor.rs
```diff
@@ -830,6 +833,49 @@ impl EditorView {
...
+        doc: &Document,
+        view: &View,
+        surface: &mut Surface,
+        theme: &Theme,
+    ) {
+        let text = doc.text().slice(..);
+
+        let primary_style = theme
+            .try_get("ui.cursorcolumn.primary")
+            .unwrap_or_else(|| theme.get("ui.cursorline.primary"));
```
the-mikedavis, line 847: For the fallback behavior, I think we should check `ui.cursorcolumn.primary` first, then `unwrap_or_else` to check `ui.cursorcolumn`, then finally unwrap_or_else to `theme.get("ui.cursorline.primary")` (and same for secondary). That way we emulate the natural fallback behavior

helix-term/src/ui/editor.rs
```diff
@@ -830,6 +833,51 @@ impl EditorView {
...
+        let text = doc.text().slice(..);
+
+        let primary_style = theme
+            .try_get("ui.cursorcolumn.primary")
+            .or_else(|| theme.try_get("ui.cursorcolumn"))
+            .unwrap_or_else(|| theme.get("ui.cursorline.primary"));
+        let secondary_style = theme
+            .try_get("ui.cursorcolumn.secondary")
+            .or_else(|| theme.try_get("ui.cursorcolumn"))
+            .unwrap_or_else(|| theme.get("ui.cursorline.secondary"));
```
sudormrfbin, line 852: `try_get` actually does check the fallback scopes, the only difference to `get` is that `get` automatically unwraps into the default style if no style is found for any scope:

https://github.com/helix-editor/helix/blob/4d4be0e880efd136100439cc34e8c526d299ec34/helix-view/src/theme.rs#L268-L278

A-Walrus: Good catch! Fixed now

helix-term/src/ui/editor.rs
```diff
@@ -830,6 +833,53 @@ impl EditorView {
...
+        // Manual fallback behaviour:
+        // ui.cursorcolumn.{p/s} -> ui.cursorcolumn -> ui.cursorline.{p/s}
+        let primary_style = theme
+            .try_get_exact("ui.cursorcolumn.primary")
+            .or_else(|| theme.try_get_exact("ui.cursorcolumn"))
+            .unwrap_or_else(|| theme.get("ui.cursorline.primary"));
+        let secondary_style = theme
+            .try_get_exact("ui.cursorcolumn.secondary")
+            .or_else(|| theme.try_get_exact("ui.cursorcolumn"))
+            .unwrap_or_else(|| theme.get("ui.cursorline.secondary"));
```
sudormrfbin, line 854: This could actually be achieved with the existing `try_get` without introducing `try_get_exact`:

```rust
        let primary_style = theme
            // checks `ui.cursorcolumn.primary` first, then `ui.cursorcolumn`
            .try_get("ui.cursorcolumn.primary")
            // if `ui.cursorcolumn` isn't defined, falls back to 
            // `ui.cursorcolumn.primary` and then to `ui.cursorcolumn`
            .unwrap_or_else(|| theme.get("ui.cursorline.primary"));
        let secondary_style = theme
            .try_get("ui.cursorcolumn.secondary")
            .unwrap_or_else(|| theme.get("ui.cursorline.secondary"));
```

PR #4088, opened 2022-10-03
doc: add example to disable arrow keys in insert mode

poliorcetics:
Following #3915, I think reducing friction for people wanting to try out the "no arrow keys in Insert Mode" is nice


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4089, opened 2022-10-03
Update sonokai.toml

vglfr:
Adds contrast between selection and cursorline.

Before (barely distinguishable):

After (clear contrast):

@p4ymak


dead10ck:
This looks fine to me

PR #4096, opened 2022-10-04
doc: add theme inheritance example

greym0uth:
With the introduction of #3067 this adds some simple documentation, with an example, on using the `inherits` property to the theme page.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -89,6 +89,14 @@ Less common modifiers might not be supported by your terminal emulator.
 | `hidden`       |
 | `crossed_out`  |
 
+### Inheritance
+
+Extend upon other themes by setting the `inherits` property to an existing theme.
+
+```toml
+inherits = "boo_berry"
+```
+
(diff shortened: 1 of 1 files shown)


book/src/themes.md
```diff
@@ -89,6 +89,14 @@ Less common modifiers might not be supported by your terminal emulator.
 | `hidden`       |
 | `crossed_out`  |
 
+### Inheritance
+
+Extend upon other themes by setting the `inherits` property to an existing theme.
+
+```toml
+inherits = "boo_berry"
```
the-mikedavis, line 97: It would also be useful to see examples of overriding values from the inherited theme. For example

```toml
# Override the theming for "keyword"s:
"keyword" = { fg = "gold" }

# Override colors in the palette:
[palette]
berry = "#2A2A4D"
```

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4097, opened 2022-10-04
Remove duplicates from the picker

ath3:
Trying to find references for example in goto_reference in commands.rs brings up a lot of duplicate matches.
This change removes the duplicates from lsp goto_*  commands.

Before:
picker_old

After:
picker_new

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -692,7 +692,7 @@ pub fn apply_workspace_edit(
 fn goto_impl(
     editor: &mut Editor,
     compositor: &mut Compositor,
-    locations: Vec<lsp::Location>,
+    mut locations: Vec<lsp::Location>,
     offset_encoding: OffsetEncoding,
 ) {
     let cwdir = std::env::current_dir().unwrap_or_default();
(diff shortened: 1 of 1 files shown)


matoous:
How can I preview multiple references to the same symbol in a specific file with these changes?

ath3:
That works, even if the same symbol is on the same line multiple times, because the position returned from lsp is different for them.
For example line `Some(highlights) if !highlights.is_empty() => highlights,` will be listed three times if i make `goto_reference` on `highlights`.

A-Walrus (approved):

helix-term/src/commands/lsp.rs
```diff
@@ -705,6 +705,11 @@ fn goto_impl(
             editor.set_error("No definition found.");
         }
         _locations => {
+            for i in (1..locations.len()).rev() {
+                if locations.as_slice()[..i].contains(&locations[i]) {
+                    locations.remove(i);
+                }
+            }
```
archseer, line 712: This is very inefficient. Can't we sort then `dedup_by`?

ath3: We would have to sort `locations` for that, but Ord trait is not implemented for `lsp_types::Location`.

poliorcetics: Could we make it a `BTreeSet` instead ? Duplicates would simply disappear. The output order would change though, and LSPs often put the definition in the first or last place, loosing this is not a good solution :/

pascalkuthe:
I think we can close this now since RA merged an upstream fix for this last year: https://github.com/rust-lang/rust-analyzer/pull/13571

Issue #2206, opened 2022-04-21
Line numbers are now configurable via the gutters setting but apparently can't be changed with the :set command


groovy9:
I'd like to be able to toggle line numbers on and off while editing.  [#1967](https://github.com/helix-editor/helix/pull/1967) added a gutters setting that allows disabling line numbers by simply not including "line-numbers" in the gutters array, which is great.  But :set throws an error if I try to set gutters to the referenced values:

`
:set gutters ["diagnostics"]
`
`
Bad arguments. Usage: ':set key field'
`

CptPotato:
`:set gutters ["diagnostics"]` doesn't error for me. If it does for you, you might be using the latest release which doesn't include that functionality, yet.

Apart from that, there seem to be two things happening here:

1. `:set` fails to parse the list, if it contains spaces
   - i.e. `["diagnostics", "line-numbers"]` fails, but `["diagnostics","line-numbers"]` works
2. the gutter is only updated when the window is created/update
   - opening a split will update it (closing one *sometimes* updates it?)

the-mikedavis:
I can reproduce this. The error comes from this line: https://github.com/helix-editor/helix/blob/24f5a47ceca5468ef22e21cdbb53e737782ff878/helix-term/src/commands/typed.rs#L1214-L1216

It looks like we're parsing the arguments so that `:set gutters ["diagnostics"]` is parsed as `["gutters", "[", "diagnostics", "]"]`.

`:set gutters "[\"diagnostics\"]"` is parsed correctly as a workaround but ideally we should be  fixing the argument parser

the-mikedavis:
Hmm actually on further investigation this looks like the intended behavior of `helix_core::shellwords` based on the tests. You should escape quotes with `\"`. Tuning that behavior could be discussed though.

`:set gutters` on its own doesn't refresh the gutters as CptPotato says. The `helix_view::editor::Editor::refresh_config` function (or maybe something in `Application`) will need some extra handling for gutters

jcalabro:
Hi there! Sorry to dig up an old issue, but I believe this is the way to do it in 23.10 at least?

```
:set gutters.layout ["line-numbers"]
```

phlippieb:
> Hi there! Sorry to dig up an old issue, but I believe this is the way to do it in 23.10 at least? This is working for me
> 
> ```
> :set gutters.layout ["line-numbers"]
> ```

[...]

PR #4098, opened 2022-10-04
Fix shellwords delimiter handling

A-Walrus:
This allows commands such as `:set statusline.center ["file-type"]` to work. Before the quotes within the list would mess it up. Also added a test to ensure correct behavior.
Addresses the parsing part of #2206 (the not updating is solved by #3457)

### DIFF
--- a/helix-core/src/shellwords.rs
+++ b/helix-core/src/shellwords.rs
@@ -3,7 +3,7 @@ use std::borrow::Cow;
 /// Get the vec of escaped / quoted / doublequoted filenames from the input str
 pub fn shellwords(input: &str) -> Vec<Cow<'_, str>> {
     enum State {
-        Delimiter, // whitespace
+        OnWhitespace,
         Unquoted,
         UnquotedEscaped,
         Quoted,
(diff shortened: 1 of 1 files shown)


helix-core/src/shellwords.rs
```diff
@@ -3,8 +3,9 @@ use std::borrow::Cow;
 /// Get the vec of escaped / quoted / doublequoted filenames from the input str
 pub fn shellwords(input: &str) -> Vec<Cow<'_, str>> {
     enum State {
-        Normal,
-        NormalEscaped,
+        Delimiter, // whitespace
```
the-mikedavis, line 6: To align with the other states I think we should call this "Delimited" or maybe "OnWhitespace"

the-mikedavis:
I think we start to diverge from the behavior of quoting within shells here, right?

```bash
$ echo ["file-type","file-encoding"]
[file-type,file-encoding]
```

Although it might be worth the tradeoff for convenience. I suppose that the alternative is that we recommend single quotes for using `:set`, for example `:set statusline.center '["file-type","file-encoding"]'`

A-Walrus:
> I think we start to diverge from the behavior of quoting within shells here, right?

Yeah, we do. I see three options

-  Diverge from  shell behavior for convenience
- Keep current behavior and document and recommend wrapping lists in single quotes
- Special handling for lists by the actual command so that `:set statusline.center file-type file-encoding` works. It would receive them as separate args from the parser, see that the keys value is of type list, and try putting all the remaining args into the list. 

@the-mikedavis it's your call, whatever you think makes sense.

archseer (approved):
LGTM, I think it's ok to diverge from regular shell behaviour

PR #4109, opened 2022-10-05
Add colours "color-modes" to ayu_light theme

smbkr:
This PR adds statusline mode colours to the ayu light theme when the `editor.color-modes` option is true. 

I've also renamed the `dark_gray` colour to `light_gray` since it's lighter than the default `gray`, so being called "dark" didn't make much sense. 

## Screenshots

Normal:
Screenshot 2022-10-05 at 15 50 18

Insert:
Screenshot 2022-10-05 at 15 50 15

Select:
Screenshot 2022-10-05 at 15 50 10

Modes are in the right split, the left split shows the inactive colours.


the-mikedavis:
\cc @enkodr

PR #4113, opened 2022-10-05
Overlay all diagnostics with highest severity on top

the-mikedavis:
Here we separate the diagnostics by severity and then overlay the Vec of spans for each severity on top of the highlights. The error diagnostics end up overlaid on the warning diagnostics, which are overlaid on the hints, overlaid on info, overlaid on any other severity (default), then overlaid on the syntax highlights.

Kudos to @pascalkuthe for coming up with this idea 🎉

This fixes two things:

* Error diagnostics are now always visible when overlapped with other diagnostics.
    * In order to see this, you need `diagnostic.*` theming that makes the severities look visually different (#4061 / #3015 are good for this, or using `bg` highlighting).
* Ghost text is eliminated.
    * Ghost text was caused by duplicate diagnostics at the EOF: overlaps within the merged `Vec<(usize, Range<usize>)>` violate assumptions in `helix_core::syntax::Merge`.
    * When we push a new range, we check it against the last range and merge the two if they overlap. This is safe because they both have the same severity and therefore highlight.

The actual merge is skipped for any of these when they are empty, so this is very fast in practice. For some data, I threw together an FPS counter which renders as fast as possible and logs the renders per second: https://github.com/the-mikedavis/helix/commit/02ae7adc8a0214b105dc441a6cb2f3ea89f1e4fb.

With no diagnostics, I see an FPS gain from this change from 868 FPS to 878 (+1.1%) on a release build on a Rust file. This makes sense: we eliminate the work of merging an empty diagnostics Vec into the highlight stream. On an Erlang file with 12 error diagnostics and 6 warnings in view (233 errors and 66 warnings total), I see a decrease in average FPS from 795 to 790 (-0.6%) on a release build.

Closes https://github.com/helix-editor/helix/issues/2248
See prior work: https://github.com/helix-editor/helix/pull/4042, https://github.com/helix-editor/helix/pull/3695

Opening as a draft for now: I'd like to add some comments and `debug_assert!`s.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -119,8 +119,15 @@ impl EditorView {
             Self::highlight_cursorline(doc, view, surface, theme);
         }
 
-        let highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
-        let highlights = syntax::merge(highlights, Self::doc_diagnostics_highlights(doc, theme));
+        let mut highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
+        for diagnostic in Self::doc_diagnostics_highlights(doc, theme) {
+            // Most of the `diagnostic` Vecs are empty most of the time. Skipping
+            // a merge for any empty Vec saves a significant amount of work.
+            if diagnostic.is_empty() {
+                continue;
+            }
+            highlights = Box::new(syntax::merge(highlights, diagnostic));
+        }
         let highlights: Box<dyn Iterator<Item = HighlightEvent>> = if is_focused {
             Box::new(syntax::merge(
                 highlights,
@@ -264,7 +271,7 @@ impl EditorView {
     pub fn doc_diagnostics_highlights(
         doc: &Document,
         theme: &Theme,
-    ) -> Vec<(usize, std::ops::Range<usize>)> {
+    ) -> [Vec<(usize, std::ops::Range<usize>)>; 5] {
         use helix_core::diagnostic::Severity;
         let get_scope_of = |scope| {
             theme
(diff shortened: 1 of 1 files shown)


pascalkuthe:
Could you go into more detail how you measured the framerate impact? @the-mikedavis 
For me the frame-rate fluctuates so widely between runs or that measuring any changes seems hopeless (on the first run I get 8131 FPS and on the second 11443 FPS for example).
I am thinking that a "proper" benchmark using criterion might be nice to get some kind of statistically relevant result. Maybe I can throw something together

the-mikedavis:
I cherry-pick that commit to a branch to test and then `cargo run --release` in one split and tail the log file in another. Then I let it run for a while undisturbed and average together the results. I see some fluctuation if the editor is busy doing things other than rendering (handling lsp messages for example) but not by many frames. I see very stable numbers between compiles and runs.

pascalkuthe:
Ah thanks I was only looking at the first couple of number, it took a while to stabilize for me.
Have a benchmark here is very useful, I should have invested more time in that before implementing #4042.
I tested this with the `overlay` implementation from #4042 but the performance was actually slightly worse (but mostly the same). I think with the benchmark now I could have made it faster but that's probably not worth the time investment right now.

This looks much nicer :+1:, great idea merging the diagnostic spans right away.
I assumed it would be slower because the extra allocations but that probably only matters in cases with unreasonable amounts of diagnostics.
Classical case of "humans are terrible at guessing performance".
I think we should add documentation to `merge` tough that it should never be called with overlapping spans (and a `debug_assert`) but I guess you already have that planned.

Really anticlimactic how few changes this needs now considering the hours we put into this, but a simple solution is ofter the best one I guess.

the-mikedavis:
> Really anticlimactic how few changes this needs now considering the hours we put into this

I was thinking the same thing! 😆

helix-term/src/ui/editor.rs
```diff
@@ -119,8 +119,12 @@ impl EditorView {
             Self::highlight_cursorline(doc, view, surface, theme);
         }
 
-        let highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
-        let highlights = syntax::merge(highlights, Self::doc_diagnostics_highlights(doc, theme));
+        let mut highlights = Self::doc_syntax_highlights(doc, view.offset, inner.height, theme);
+        for diagnostic in Self::doc_diagnostics_highlights(doc, theme) {
+            if !diagnostic.is_empty() {
+                highlights = Box::new(syntax::merge(highlights, diagnostic));
+            }
```
archseer, line 126:
```suggestion
            if diagnostic.is_empty() {
                continue;
            }
            highlights = Box::new(syntax::merge(highlights, diagnostic));
```

helix-term/src/ui/editor.rs
```diff
@@ -285,22 +301,48 @@ impl EditorView {
...
+        let mut error_vec = Vec::new();
+
+        for diagnostic in doc.diagnostics() {
+            match diagnostic.severity {
+                Some(Severity::Info) => merge_or_push(
+                    &mut info_vec,
+                    info,
+                    diagnostic.range.start,
+                    diagnostic.range.end,
+                ),
```
archseer, line 317: Flip this so it's `merge_or_push(match diagnostic.severity, ...)`

Philipp-M:
In the latest rebase in #2507 I've noticed an issue with this assertion:

```rust
// Diagnostics must be sorted by range. Otherwise, the merge strategy
// below would not be accurate.
debug_assert!(doc
    .shown_diagnostics()
    .collect::<Vec<_>>()
    .as_slice()
    .windows(2)
    .all(|window| window[0].range.start <= window[1].range.start
        && window[0].range.end <= window[1].range.end));
```

This assertion failed with diagnostics with the ranges (in this order) `161..164, 162..162`
So they're sorted by range. Do you know if the second shorter range causes any problem? Otherwise I'd suggest checking if the whole diagnostics are actually sorted by range (i.e. lexicographic ordering of the struct).

the-mikedavis:
Ah, yeah it's not necessary that `window[0].range.end <= window[1].range.end` when `window[0].range.start <= window[1].range.start` because of the `std::cmp::max` here: https://github.com/helix-editor/helix/blob/2457111bf074d4b5c1d45f76e2c7d593ceec5895/helix-term/src/ui/editor.rs#L328

Yep, changing the `all` condition to

```rs
|window| window[0].range <= window[1].range
```

would be more correct

pascalkuthe:
> Ah, yeah it's not necessary that `window[0].range.end <= window[1].range.end` when `window[0].range.start <= window[1].range.start` because of the `std::cmp::max` here:
> 
> https://github.com/helix-editor/helix/blob/2457111bf074d4b5c1d45f76e2c7d593ceec5895/helix-term/src/ui/editor.rs#L328
> 
> Yep, changing the `all` condition to
> 
> ```rust
> |window| window[0].range <= window[1].range
> ```
> 
> would be more correct

I think it should just be `|window| window[0].range.start <= window[1].range.start`, the `PartialOrd` implementation of range takes end into account AFAIK.

the-mikedavis:
https://github.com/helix-editor/helix/pull/4319

PR #4117, opened 2022-10-05
Update C and C++ tree-sitter grammars

Triton171:
I'm mainly interested in this because of the [support for alternative operators in C++](https://github.com/tree-sitter/tree-sitter-cpp/pull/179) but there are also some more improvements, such as support for C++20 concepts and attributes in C. I've checked that all our queries still work but if there's anything obvious missing, please let me know.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -184,7 +184,7 @@ args = { console = "internalConsole", attachCommands = [ "platform select remote
 
 [[grammar]]
 name = "c"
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Part of this is being taken care of in #4079

Triton171:
Thanks, I'll close this and suggest the remaining changes in the other PR.

PR #4118, opened 2022-10-05
Improve C# syntax highlighting

Chickenkeeper:
Added some missing operators and highlights, and organised the keywords into the right colours.

Before:
c#_old

After:
c#_new

### DIFF
--- a/runtime/queries/c-sharp/highlights.scm
+++ b/runtime/queries/c-sharp/highlights.scm
@@ -11,6 +11,11 @@
 
 (invocation_expression
   (member_access_expression
+    expression: ((identifier) @type
+      (#match? @type "^[A-Z]"))))
+
+(invocation_expression
+  (member_access_expression
     expression: (identifier) @variable))
 
 (invocation_expression
@@ -19,12 +24,12 @@
       name: (identifier) @function)))
 
 (invocation_expression
-      [(identifier) (qualified_name)] @function)
+  [(identifier) (qualified_name)] @function)
 
 ; Generic Method invocation with generic type
 (invocation_expression
   function: (generic_name
-              . (identifier) @function))
+    . (identifier) @function))
 
 ;; Namespaces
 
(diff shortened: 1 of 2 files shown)


runtime/queries/c-sharp/highlights.scm
```diff
@@ -334,14 +354,26 @@
 (lock_statement (identifier) @variable)
 
 ;; Rest
-(member_access_expression) @variable
+(member_access_expression
+  expression: ((identifier) @type
+    (#match? @type "^[A-Z]")))
```
the-mikedavis, line 359: Trying to make this look nice in a file like https://github.com/EventStore/EventStore/blob/7ffc4bc215e47398f2f7126f6add55c7dba23cdd/src/EventStore.Core.Tests/ClientAPI/ExpectedVersion64Bit/append_to_stream_with_event_numbers_greater_than_2_billion.cs, I think:

```scm
(member_access_expression
  expression: (identifier) @type @_expr
  name: (identifier) @type @_name
  (#match? @_expr "^[A-Z]")
  (#match? @_name "^[A-Z]"))
(member_access_expression
  expression: (_) @_expr
  name: (identifier) @type @_name
  (#match? @_expr "^[A-Z]")
  (#match? @_name "^[A-Z]"))
(member_access_expression
  expression: ((identifier) @type
    (#match? @type "^[A-Z]")))
```

Works nicely. We might be able to simplify it a little. I don't know C# well enough to know if that highlighting is correct though 😅. What do you think?

Chickenkeeper: That is an interesting idea, but it's not quite accurate. In C# member accesses can be properties, fields, types, or constants, and all of them except fields can be PascalCase, so extra context is needed to know which colour to pick. But tree-sitter-c-sharp doesn't seem to provide that context, so it's difficult to accurately colour anything other than the first variable being accessed (which can only be a variable or a type, camelCase and PascalCase respectively).

As a result your code often ends up incorrectly flagging properties and constants as types. That's why I chose to interpret everything except the first variable as a variable, because it felt like the least confusing solution. But I'm still open to better solutions

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #2019, opened 2022-04-08
Panic when incrementing two -1 values


n0s4:
### Summary

The original summary included a lot of context that isn't related. After some discussion this is what has been found.

If you have a file with the following:
```
-1
-1
```
And you enter the following keystrokes:
`%` `s` `-` `<enter>` `C-a`

Helix panics.

This **doesn't** happen when there is a character, space or empty line **after** the two lines:
```
-1
-1

```
(Helix doesn't panic with the above input on this text)

Here is the backtrace with RUST_BACKTRACE=1:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char index out of bounds: char index 4, Rope/RopeSlice char length 3', /home/jm/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.4.1/src/slice.rs:349:41
stack backtrace:
   0: rust_begin_unwind
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/std/src/panicking.rs:498:5
   1: core::panicking::panic_fmt
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/panicking.rs:116:14
   2: core::result::unwrap_failed
             at /rustc/9d1b2106e23b1abd32fce1f17267604a5102f57a/library/core/src/result.rs:1690:5
   3: ropey::slice::RopeSlice::char_to_byte
   4: helix_core::graphemes::nth_next_grapheme_boundary
   5: helix_core::selection::Range::grapheme_aligned
   6: helix_core::selection::Selection::ensure_invariants
   7: helix_view::document::Document::apply_impl
   8: helix_view::document::Document::apply
   9: helix_term::commands::increment_impl
  10: helix_term::ui::editor::EditorView::handle_keymap_event
  11: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::handle_event
  12: helix_term::compositor::Compositor::handle_event
  13: helix_term::application::Application::handle_terminal_events
  14: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  15: tokio::park::thread::CachedParkThread::block_on
  16: tokio::runtime::thread_pool::ThreadPool::block_on
  17: tokio::runtime::Runtime::block_on
  18: hx::main
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
```

[...]

Issue #3028, opened 2022-07-10
Split goto-number functionality from increment/decrement into a textobject


the-mikedavis:
Increment (`C-a`) and decrement (`C-d`) currently move forward to the next number on their line if the cursor is not already over a number.

It would fit better with the select-then-act paradigm to split the motion and action behaviors. `C-a` and `C-d` can continue to increment and decrement but should not move the cursor. Increment/decrement should only act on the current selections. A new textobject `]n`/`[n`/`min`/`man` could replace the motion.

The new textobject could use tree-sitter queries to find number specific to the current language and/or could use a default naive implementation like what we have in increment/decrement now. (Falling back to a naive implementation would only work if we can easily detect that a query does not have any `number.inside`/`number.around` captures. I think that should be possible to figure out without running the query.)

pascalkuthe:
This was addressed by  #4123 I think?

the-mikedavis:
I was thinking at the time that we might want new textobjects to make jumping to numbers easier. But with macro keybinding and possibly also https://github.com/helix-editor/helix/issues/5672, new textobjects aren't necessary. So yep I consider this resolved :+1:

PR #4123, opened 2022-10-06
Separate jump behavior from increment/decrement

greg-enbala:
Fixes: [#2019](https://github.com/helix-editor/helix/issues/2019)

This fixes a couple situations I found that cause undesired behavior when incrementing integers:

### Trailing underscores, eg: `9_`. When incrementing can cause an infinite loop.
I fixed this by not including trailing underscores in the found character Range. This means the incrementing code can
not care about them. Alternatively we could generate the range as it currently is and make the incrementing code ignore the trailing underscores.

### A number that changes width when incrementing eg `-1` to `0` or `9` to `10` can crash.
Crux of the problem: When multiple numbers are incremented at the same time, there can be multiple shifts in the document length. For example, selecting all the negative signs in `-1 -1 -1` and incrementing will cause the document to be 3 characters shorter. If you naively kept the selection positions tied to the changes, you'd end up with invalid selections that may be entirely off the end of the resulting document. This PR attempts to solve this problem by keeping a running count of how many characters are added or subtracted per applied change and adjusting the resulting selections accordingly.

## Notable code structure changes in this PR
* `NumberIncrementor` renamed to `IntegerIncrementor` to more accurately reflect what it does.
* `Incrementor` trait removed.

## Notable behavior changes in this PR
* Incrementing only effects what is directly selected. This means the number of selections will never change and selections don't go searching to the end of the line to find something to increment.

* DateTimeIncrementor can now only ever increment the Day if what is selected is a Date, or the minutes if what is selected is a DateTime. This is a consequence of only operating directly on what is currently selected. 

* Numbers can no longer wrap on increment/decrement. The addition/subtraction is saturating. I made this change as I feel it's surprising behavior.

Happy to change this PR to work however folks want it to, add tests, rename things, change the algorithm, whatever folks think is best for helix. Hope this helps!


helix-core/src/increment/integer.rs
```diff
@@ -10,37 +10,45 @@ use crate::{
...
-        } else {
-            range
+        if range.from() > 0 && text.char(range.from() - 1) == '-' {
+            range = range.extend(range.from() - 1, range.from())
         };
 
+        // If there are trailing underscores, remove them from the range.
+        // There is a `- 1` here because range ends are exclusive.
+        while !range.is_empty() && text.char(range.to() - 1) == '_' {
+            range = range.shorten()
```
the-mikedavis, line 49: `Range::shorten` is a bit confusing: it's not clear which end is being trimmed and by how much, and it can produce a range with 0 width. I think we can get away with just tracking `to` in this loop and not creating a new range for each iteration, and returning `None` if after this loop `to <= range.from()`.

helix-term/src/commands.rs
```diff
@@ -4833,48 +4831,57 @@ fn increment_impl(cx: &mut Context, amount: i64) {
...
-            Some(change)
-        }
-    });
 
-    if changes.clone().count() > 0 {
-        let transaction = Transaction::change(doc.text(), changes);
-        let transaction = transaction.with_selection(selection.clone());
+    if !changes.is_empty() {
+        let new_selection_primary =
+            std::cmp::min(selection.primary_index(), new_selection_ranges.len() - 1);
```
the-mikedavis, line 4880:
```suggestion
            selection.primary_index().min(new_selection_ranges.len() - 1);
```

helix-term/src/commands.rs
```diff
@@ -4818,63 +4818,62 @@ fn find_next_char_until_newline<M: CharMatcher>(
...
+        false,
         char::is_ascii_digit,
         1,
     );
 
     let (view, doc) = current!(cx.editor);
     let selection = doc.selection(view.id);
     let text = doc.text().slice(..);
+    let mut new_selection_ranges = SmallVec::new();
+    let mut comulative_length_diff: i128 = 0;
```
the-mikedavis, line 4834: `cumulative_length_diff`

helix-term/tests/test/commands.rs
```diff
@@ -131,3 +131,29 @@ async fn test_selection_duplication() -> anyhow::Result<()> {
...
+    .await?;
+
+    // correct selection placement with multiple -1 to zero, incremented twice
+    // `-1-1-2-1-1` becomes `00-100` then `01-99`
+    test((
+        "#[-|]#1#(-|)#1#(-|)#2#(-|)#1#(-|)#1",
+        "<C-a><C-a>",
+        "#[0|]#1#(-|)#99",
+    ))
+    .await?;
```
the-mikedavis, line 153: I understand what's happening here with the numbers colliding but I think this behavior is confusing: I wouldn't expect incrementing or decrementing to change the number of selections. If I have 5 selections over the negatives, I would expect to be able to retain those selections as I increment/decrement and always end up with 5 selections

I'm not sure it's possible to have the behavior I expect without separating the navigation part of C-a/C-x from the incrementing/decrementing behavior (https://github.com/helix-editor/helix/issues/3028)

the-mikedavis:
Yep those are all expected. I'm not sure about how others feel about this change. The jumping behavior is convenient but I think it's a Vim relic that doesn't fit the paradigm.

the-mikedavis:
Yeah, "the exact selection must be parsable as something incrementable" is what I had in mind. Just trying each incrementor in sequence should do the trick - no two incrementors should be able to successfully handle the same input

helix-core/src/increment/date_time.rs
```diff
@@ -1,116 +1,64 @@
...
-use crate::{Range, Tendril};
-
 #[derive(Debug, PartialEq, Eq)]
-pub struct DateTimeIncrementor {
-    date_time: NaiveDateTime,
-    range: Range,
-    fmt: &'static str,
-    field: DateField,
-}
+pub struct DateTimeIncrementor;
```
the-mikedavis, line 7: I think we only had these structs before to work with the trait. Now we should be able to write increment as a regular (non-associated) function for both of these modules

helix-term/src/commands.rs
```diff
@@ -4873,86 +4873,48 @@ fn decrement(cx: &mut Context) {
...
-            let (range, new_text) = incrementor.increment(amount);
+    let mut new_selection_ranges = SmallVec::new();
+    let mut cumulative_length_diff: i128 = 0;
+    let mut changes = vec![];
 
-            Some((range.from(), range.to(), Some(new_text)))
-        })
-        .collect();
+    for range in selection {
+        let selected_text: Cow<str> = text.slice(range.from()..range.to()).into();
```
the-mikedavis, line 4888: I believe there's a helper for this: `range.fragment(text)` IIRC

helix-term/src/commands.rs
```diff
@@ -4873,86 +4873,48 @@ fn decrement(cx: &mut Context) {
...
-    // the position we send to this function so that it's relative to that line and its returned
-    // position since it's expected this function returns a global position.
-    let line_index = text.char_to_line(pos);
-    let pos_delta = text.line_to_char(line_index);
-    let pos = pos - pos_delta;
-    search::find_nth_next(text.line(line_index), char_matcher, pos, 1).map(|pos| pos + pos_delta)
-}
-
-/// Decrement object under cursor by `amount`.
+/// Increment object under cursor by `amount`.
```
the-mikedavis, line 4876: Can say "increment numbers within selections by `amount`" now

the-mikedavis (commented):
Ah sorry, this fell off of my radar for a while 😅

Since we're deleting so much of both modules here, I wonder if we should pull the increment function for datetimes and the increment function for numbers into just one `helix_core::increment` module as `datetime` and `integer` functions. What do you think? OTOH having separate modules is nice since they're small and well-contained

the-mikedavis (approved):
I pushed a merge commit to resolve conflicts and bring in the changes from #4418.

Thanks for working on this! I like how much this cleaned up and simplified the increment/decrement code 🚀

PR #4134, opened 2022-10-06
Improve sorting for inline menu (codeaction + completion)

### LINKED ISSUE #2508: Sometimes wrong order in autocomplete menu

pascalkuthe:
Code actions are currently sorted alphabetically because they simply reuse the `Menu` used for autocompletions.
This is extremely inconvenient in many cases.
For example rust-analyzer emits a ton of (sometimes useful) refactors code-actions that get sorted before
the quick fix (import ...) I am interested in.

This PR improves upon this in two ways:

The simplest fix is simply not sorting code actions in the `Menu`.
However many LSP implementations send their code actions unordered because they assume that the client will sort the code actions by the `kind` field.

VScode uses 8 cateogoris for code actions:

* quickfix
* refactor (extract)
* refactor (inline)
* refactor (rewrite)
* refactor (move)
* refactor (surround)
* source
* anything else

I have opted to match this behavior by sorting the code actions into these categories.
Note that I was careful to retain the original sorting by the server within these categories.

This is a huge usability improvement for people that frequently use codeactions.
Example from the helix sourcecode where an import is missing:

**before:**
image

**after:**
image

Edit: This PR was originally just about code-actions, however I found #3215 and noticed that the changes there are sort of the same as in this PR: both PRs needed to handle code-actions separately from auto-completion sorting, #3215 sorted the code-actions alphabetically while this PR sorts them by category. Enabling fuzzy sorting for auto completions on top this PR is a single-line change and I have therefore opted to include it in this PR. I can drop the commit again and spin that out into a separate PR if this is controversial. The change is trivial and it made sense to me to include it here.  fixes #2508

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -490,35 +490,34 @@ pub fn code_action(cx: &mut Context) {
                 _ => 7,
             });
 
-            let mut picker =
-                ui::Menu::new(actions, false, (), move |editor, code_action, event| {
-                    if event != PromptEvent::Validate {
-                        return;
+            let mut picker = ui::Menu::new(actions, (), move |editor, code_action, event| {
+                if event != PromptEvent::Validate {
+                    return;
+                }
+
+                // always present here
+                let code_action = code_action.unwrap();
+
+                match code_action {
+                    lsp::CodeActionOrCommand::Command(command) => {
+                        log::debug!("code action command: {:?}", command);
+                        execute_lsp_command(editor, command.clone());
                     }
+                    lsp::CodeActionOrCommand::CodeAction(code_action) => {
+                        log::debug!("code action: {:?}", code_action);
+                        if let Some(ref workspace_edit) = code_action.edit {
+                            log::debug!("edit: {:?}", workspace_edit);
+                            apply_workspace_edit(editor, offset_encoding, workspace_edit);
+                        }
 
-                    // always present here
-                    let code_action = code_action.unwrap();
-
-                    match code_action {
-                        lsp::CodeActionOrCommand::Command(command) => {
-                            log::debug!("code action command: {:?}", command);
+                        // if code action provides both edit and command first the edit
+                        // should be applied and then the command
+                        if let Some(command) = &code_action.command {
                             execute_lsp_command(editor, command.clone());
                         }
-                        lsp::CodeActionOrCommand::CodeAction(code_action) => {
-                            log::debug!("code action: {:?}", code_action);
-                            if let Some(ref workspace_edit) = code_action.edit {
-                                log::debug!("edit: {:?}", workspace_edit);
-                                apply_workspace_edit(editor, offset_encoding, workspace_edit);
-                            }
-
-                            // if code action provides both edit and command first the edit
-                            // should be applied and then the command
-                            if let Some(command) = &code_action.command {
-                                execute_lsp_command(editor, command.clone());
-                            }
-                        }
                     }
-                });
+                }
+            });
             picker.move_down(); // pre-select the first item
 
             let popup = Popup::new("code-action", picker);
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -467,38 +467,28 @@ pub fn code_action(cx: &mut Context) {
             // while more situational commands (like refactors) show up later
             // this behaviour is modeled after the behaviour of vscode (editor/contrib/codeAction/browser/codeActionWidget.ts)
 
-            let mut categories = vec![Vec::new(); 8];
-            for action in actions.drain(..) {
-                let category = match &action {
-                    lsp::CodeActionOrCommand::CodeAction(lsp::CodeAction {
-                        kind: Some(kind),
-                        ..
-                    }) => {
-                        let mut components = kind.as_str().split('.');
+            actions.sort_by_key(|action| match &action {
+                lsp::CodeActionOrCommand::CodeAction(lsp::CodeAction {
+                    kind: Some(kind), ..
+                }) => {
+                    let mut components = kind.as_str().split('.');
 
-                        match components.next() {
-                            Some("quickfix") => 0,
-                            Some("refactor") => match components.next() {
-                                Some("extract") => 1,
-                                Some("inline") => 2,
-                                Some("rewrite") => 3,
-                                Some("move") => 4,
-                                Some("surround") => 5,
-                                _ => 7,
-                            },
-                            Some("source") => 6,
+                    match components.next() {
+                        Some("quickfix") => 0,
+                        Some("refactor") => match components.next() {
+                            Some("extract") => 1,
+                            Some("inline") => 2,
+                            Some("rewrite") => 3,
+                            Some("move") => 4,
+                            Some("surround") => 5,
                             _ => 7,
-                        }
+                        },
+                        Some("source") => 6,
+                        _ => 7,
                     }
-                    _ => 7,
-                };
-
-                categories[category].push(action);
-            }
-
-            for category in categories {
-                actions.extend(category.into_iter())
-            }
+                }
+                _ => 7,
+            });
 
             let mut picker =
                 ui::Menu::new(actions, false, (), move |editor, code_action, event| {
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -465,7 +465,7 @@ pub fn code_action(cx: &mut Context) {
             // sort by CodeActionKind
             // this ensures that the most relevant codeactions (quickfix) show up first
             // while more situational commands (like refactors) show up later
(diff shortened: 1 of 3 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -452,43 +452,76 @@ pub fn code_action(cx: &mut Context) {
...
+            // while more situational commands (like refactors) show up later
+            // this behaviour is modeled after the behaviour of vscode (editor/contrib/codeAction/browser/codeActionWidget.ts)
+
+            let mut categories = vec![Vec::new(); 8];
+            for action in actions.drain(..) {
+                let category = match &action {
+                    lsp::CodeActionOrCommand::CodeAction(lsp::CodeAction {
+                        kind: Some(kind),
+                        ..
+                    }) => match kind.as_str() {
```
poliorcetics, line 476: Why match on the str form when you could match on `Some(Enum::Member)` and `None` ?

poliorcetics: Or, here, since you match on `Some(kind)` above, directly on `Enum::Member`

poliorcetics (commented):
That bugged me for so long, I didn't know the order was not limited to "alphabetic" and "lsp-defined" !

helix-term/src/commands/lsp.rs
A-Walrus: @poliorcetics  `CodeActionKind` isn't an Enum its:
```rust
#[derive(Debug, Eq, PartialEq, Hash, PartialOrd, Clone, Deserialize, Serialize)]
pub struct CodeActionKind(Cow<'static, str>);
```
(from `lsp-types` crate)

Although maybe we should match against the consts in `CodeActionKind` rather than on strings directly :
```rust
    pub const EMPTY: CodeActionKind = CodeActionKind::new("");
    pub const QUICKFIX: CodeActionKind = CodeActionKind::new("quickfix");
    pub const REFACTOR: CodeActionKind = CodeActionKind::new("refactor");
    pub const REFACTOR_EXTRACT: CodeActionKind = CodeActionKind::new("refactor.extract");
    pub const REFACTOR_INLINE: CodeActionKind = CodeActionKind::new("refactor.inline");
    pub const REFACTOR_REWRITE: CodeActionKind = CodeActionKind::new("refactor.rewrite");
    pub const SOURCE: CodeActionKind = CodeActionKind::new("source");
    pub const SOURCE_ORGANIZE_IMPORTS: CodeActionKind = CodeActionKind::new("source.organizeImports");
```

Also currently you're treating `source.organizeImports` and `source.fixAll`[^1] as other (7), and not as part of `source` (6).

[^1]: Not present in lsp-types crate but present in spec

pascalkuthe: Matching on the constants does not work because CodeActionKind is a `Cow` which does not implement structural Eq (so can not be used in a match) while the string can be used in the match.

Upon @A-Walrus suggestion I looked more closely at VScode and noticed that they treat the `CodeActionKind` as a path (seperated by a `.`). Therefore when they match `quickfix`, they also match `quickfix.X.Y.Z`. I have updated the PR to do the same by using `str::split('.')` and matching on the components.

poliorcetics (approved):

helix-term/src/commands/lsp.rs
```diff
@@ -452,43 +452,83 @@ pub fn code_action(cx: &mut Context) {
...
-                if event != PromptEvent::Validate {
-                    return;
-                }
+            // sort by CodeActionKind
+            // this ensures that the most relevant codeactions (quickfix) show up first
+            // while more situational commands (like refactors) show up later
+            // this behaviour is modeled after the behaviour of vscode (editor/contrib/codeAction/browser/codeActionWidget.ts)
+
+            let mut categories = vec![Vec::new(); 8];
+            for action in actions.drain(..) {
```
archseer, line 471: Couldn't this just use `sort_by`/`sort_by_key`?

pascalkuthe: Oh yeah good catch, I modeled this after the original implementation in vscode but a stable sort is a better fit here. I have updated the PR to use `sort_by_key`

helix-term/src/commands/lsp.rs
```diff
@@ -452,43 +452,73 @@ pub fn code_action(cx: &mut Context) {
...
                 return;
             }
 
-            let mut picker = ui::Menu::new(actions, (), move |editor, code_action, event| {
-                if event != PromptEvent::Validate {
-                    return;
+            // sort by CodeActionKind
+            // this ensures that the most relevant codeactions (quickfix) show up first
+            // while more situational commands (like refactors) show up later
+            // this behaviour is modeled after the behaviour of vscode (editor/contrib/codeAction/browser/codeActionWidget.ts)
```
the-mikedavis, line 468: Could this be a permalink instead? So the reference doesn't get lost if vscode reorganizes their repo

helix-term/src/commands/lsp.rs
```diff
@@ -452,43 +452,73 @@ pub fn code_action(cx: &mut Context) {
...
-                    return;
+            // sort by CodeActionKind
+            // this ensures that the most relevant codeactions (quickfix) show up first
+            // while more situational commands (like refactors) show up later
+            // this behaviour is modeled after the behaviour of vscode (https://github.com/microsoft/vscode/blob/eaec601dd69aeb4abb63b9601a6f44308c8d8c6e/src/vs/editor/contrib/codeAction/browser/codeActionWidget.ts)
+
+            actions.sort_by_key(|action| match &action {
+                lsp::CodeActionOrCommand::CodeAction(lsp::CodeAction {
+                    kind: Some(kind), ..
+                }) => {
```
the-mikedavis, line 473: CodeAction has a `is_preferred` field that I think should take some priority in this sorting: https://docs.rs/lsp-types/latest/lsp_types/struct.CodeAction.html#structfield.is_preferred

(For example, edit out the `mut` on [this line](https://github.com/helix-editor/helix/blob/e16c63276082f1cf79c57ff2ed2d301fe3333a0e/helix-term/src/keymap.rs#L186) and open up code actions on `command`: "Import `crate::commands`" and "Implement default members" are also quickfixes that will take priority over "consider changing this to be mutable" unless we take `is_preferred` into account.)

helix-term/src/ui/menu.rs
```diff
@@ -74,6 +74,7 @@ impl<T: Item> Menu<T> {
     // rendering)
     pub fn new(
         options: Vec<T>,
+        sort: bool,
```
sudormrfbin, line 77: Using an enum here would be clearer, and we follow this pattern in most places in the codebase.

pascalkuthe: I usually use an `bool` whenever its just about enableing or disabling a single option and an enum if there are two seperate orthogonal options.

However your comment drew my attention to the fact that sorting in `new` never makes sense (the fuzzy score of every item is the same and its an unstable sort so the order is prettymuch random) so I just reemoved that argument and never sort. This actually yields slightly nicer results when the autocompletions popup menu first shows up

helix-term/src/commands/lsp.rs
pascalkuthe: Ah thanks for poiting that out, I took a look at how vscode handels this.
They always sort by category but within those category the codeactions are sorted by whether they fix any diagnostics and whether they are marked as preffered. I have implemented the same behaviour here.
The details are documented in the sourcecode (including permalinks to the vscode reference)

helix-term/src/ui/menu.rs
```diff
@@ -112,10 +108,7 @@ impl<T: Item> Menu<T> {
                         .map(|score| (index, score))
                 }),
         );
-        // matches.sort_unstable_by_key(|(_, score)| -score);
-        self.matches.sort_unstable_by_key(|(index, _score)| {
-            self.options[*index].sort_text(&self.editor_data)
-        });
```
archseer, line 118: Doesn't this change impact other menus now?

helix-term/src/ui/menu.rs
```diff
@@ -77,24 +77,20 @@ impl<T: Item> Menu<T> {
...
             widths: Vec::new(),
             callback_fn: Box::new(callback_fn),
             scroll: 0,
             size: (0, 0),
             viewport: (0, 0),
             recalculate: true,
-        };
-
-        // TODO: scoring on empty input should just use a fastpath
-        menu.score("");
```
archseer, line 95: Same here, this is probably desirable outside of code actions?

archseer: Ah I see, line 80 now does this fastpath

Issue #2027, opened 2022-04-08
Moving around in insert mode triggers LSP completions; it shouldn’t


aral:
### Summary

LSP completions should only be triggered while typing, not when moving around in LSP mode. Otherwise, it is very easy to corrupt code when using the up/down arrows to move between lines.

### Reproduction Steps

For easier reproduction, set `idle-timeout` to `0` in your _config.toml_:

```toml
[editor]
idle-timeout = 0
```

Then, with the following basic JS example:

```js
// Hello, World!
console.log('Hello, world')
```

1. Go to line 2 (<kbd>g2g</kbd>)
2. Enter insert mode (<kbd>i</kbd>)
3. Press the right-arrow key three times then press the up-arrow key to get to the comment on line 1.

## What should happen

The completion dialogue should not show and you should find yourself on line 1. The code on line 2 should remain unchanged.

## What actually happens

The completion dialog shows and pressing the up arrow inserts the `continue` completion in the middle of the `console` keyword so you end up with:

```js
// Hello, World!
continuensole.log('Hello, world')
```

### Helix log

n/a

### Platform

Linux (elementary OS 6; based on Ubuntu)

### Terminal Emulator

io.elementary.t

### Helix Version

helix 22.05-dev (2d4f94eb)

sudormrfbin:
Previous discussion: https://github.com/helix-editor/helix/issues/1381.

aral:
Given https://github.com/helix-editor/helix/issues/2580 if this cannot be fixed natively within the design of Helix, perhaps the thing to do is to concede that insert mode only works on a single line and to force the editor to exit into normal mode whenever the person navigates to a different line using the up/down arrow keys.

While this will require people to constantly go back and forth between insert and normal mode, this does appear to be the expectation from what I’ve been reading from the responses and by making it automatic, Helix would at least avoid the code corruptions that can result from moving between lines using the arrow keys in insert mode.

The other way to avoid the corruptions would be to make <kbd>tab</kbd> and <kbd>shift-tab</kbd> be the only ways to navigate the autocompletion drop downs. Come to think of it, this might be the better solution.

archseer:
You shouldn't set `idle-timeout` to `0`: `idle-timeout` is used in other areas than just autocompletion, i.e. https://github.com/helix-editor/helix/commit/a85e3862988933953d1c0e7435969604d3722959

[...]

PR #4141, opened 2022-10-07
Do not accept keydown to enter in autocomplete popup

p1gp1g:
With idle_timeout = 0 it is nearly impossible to navigate on the document with the arrows.
A solution is to ignore keydown until at least one autocompletion is selected. (`<tab><keydown>[...]` or `<ctrl-n><keydown>[...]`)

It is ignored **only if idle_timeout = 0**. I a user want to have the current behavior, the user can set idle_timeout to 1.

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -306,6 +306,17 @@ impl Component for Completion {
         {
             return EventResult::Ignored(None);
         }
+        if let None = self.popup.contents().selection() {
+            if let Event::Key(KeyEvent {
+                code: KeyCode::Down,
+                ..
+            }) = event
+            {
+                if cx.editor.config().idle_timeout.as_millis() == 0 {
+                    return EventResult::Ignored(None);
+                }
+            }
+        }
         self.popup.handle_event(event, cx)
     }
 


dead10ck:
This is kind of why the idle timeout exists to begin with. Why not just set it to a very small value? Surely you would not be able to perceive the difference between 0 and e.g. 20ms.

p1gp1g:
I like to see suggestions while I'm typing and I do not always pick one. But I also move around with arrows while being in insert mode.

Possible solutions are:
A. 2 timeouts, one by default (moving around : you need to wait timeout1 to see the pop up), one while typing (first non-arrow key press : you need to wait timeout2 to see the pop up), moving after typing re-use the timeout1 : There I can set the second timeout to 0. < I think that's not very user-friendly
B. Like the previous one but timeout2 is always 0 : moving around  < That's the behavior on Android Studio for instance
C. Having another config parameter to disable keydown to enter the pop up < That adds another config.. And I though other users with timeout 0 have probably the same issue (for instance https://github.com/helix-editor/helix/issues/2027) and the ones who like it can still use timeout 1. =>
D. This PR. 

For me B. and D. are the best ones. And D it pretty simple to implement and do not add complexity to the code

dead10ck:
Why don't you try 20ms and see how it works for you?

p1gp1g:
I'll try, and I reopen if needed. Thanks !

Issue #1381, opened 2021-12-26
autocomplete triggers in insert mode


adsick:
Oftentimes autocompletion triggers in the middle of a token and starts changing .methods() which is not desired as I'm just navigating the file in insert mode.

dead10ck:
Wait, are you saying you're just moving around with the arrow keys in insert mode, and it randomly decides to insert parens? Could you provide step by step instructions to reproduce this behavior?

adsick:
> are you saying you're just moving around with the arrow keys in insert mode

Yes. 
I've figured out how to reproduce it:
for example take ```helix-term/src/main.rs:44:18```
switch to append/insert mode
now move your cursor to the letter ```i``` just after the letter ```x``` as on the screenshot
![image](https://user-images.githubusercontent.com/24528839/147409200-9de74331-a494-4216-8150-2118b5c8bdf0.png)
if your minimal number of symbols to trigger autocompletion is 2 or less you will see it appearing which is extremely annoying in case you move vertically from line to line - you might accidentally change a method this way. 

Looks like autocompletion triggers in case there are completions for symbols that are LEFT to the cursor (ignoring the fact that there is stuff RIGHT to it)
I hope the problem is clear, ask if it is not

dead10ck:
Ahh, I see; this is just the auto complete being automatically triggered because you're resting the cursor in insert mode. Typically, as a modal editor, the workflow is designed around being in normal mode for navigation and switching to insert mode only to insert text. I'd encourage you to try this kind of workflow to get the most out of the modal editing paradigm.

With that said, there are a few [config options](https://docs.helix-editor.com/configuration.html) you could try. `auto-completion = false` will disable the automatic pop-up, meaning to auto complete, you must explicitly hit Ctrl-x. Setting `idle-timeout` to a bigger value might also help.

As for completing even though there is text to the left, I think this is desired for cases like if you have a chain of method calls and you want to insert another, e.g. `foo|.bar().baz() -> foo.quu|.bar().baz()` and you want it to auto complete `quux`.

kirawi:
Closing as the above comment offers a solution.

kendfss:
> As for completing even though there is text to the right, I think this is desired for cases like if you have a chain of method calls and you want to insert another, e.g. `foo|.bar().baz() -> foo.quu|.bar().baz()` and you want it to auto complete `quux`.

[...]

PR #4144, opened 2022-10-07
Do not accept keydown to enter in autocomplete popup

p1gp1g:
I must open a new PR since I have force-pushed a change for https://github.com/helix-editor/helix/pull/4141 when it was closed and I cannot reopen it.

This is a fix for #2027. If `arrows_to_enter_completion` is set to false, user must use `tab` or `C-n` to enter the autocompletion pop and can use arrows once in the pop up.

<details><summary>Old description</summary>
So, after moving around, this is just a fix for https://github.com/helix-editor/helix/issues/2027 (and related issues) if timeout is set to 0. Maybe it is better to add a new parameter instead of timeout 0 ? For instance `arrows_to_enter_autocompletion_popup` with default to `true` (current behaviour)
</details>

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -307,7 +307,7 @@ impl Component for Completion {
             return EventResult::Ignored(None);
         }
         if self.popup.contents().selection().is_none()
-            && cx.editor.config().idle_timeout.as_millis() == 0
+            && !cx.editor.config().arrows_to_enter_completion
         {
             match event {
                 Event::Key(KeyEvent {
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -306,6 +306,20 @@ impl Component for Completion {
         {
             return EventResult::Ignored(None);
         }
+        if let None = self.popup.contents().selection() {
+            if cx.editor.config().idle_timeout.as_millis() == 0 {
+                match event {
+                    Event::Key(KeyEvent {
+                        code: KeyCode::Down,
+                        ..
+                    })
+                    | Event::Key(KeyEvent {
+                        code: KeyCode::Up, ..
+                    }) => return EventResult::Ignored(None),
+                    _ => (),
+                }
+            }
+        }
         self.popup.handle_event(event, cx)
     }
 
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -306,18 +306,18 @@ impl Component for Completion {
         {
             return EventResult::Ignored(None);
         }
-        if let None = self.popup.contents().selection() {
-            if cx.editor.config().idle_timeout.as_millis() == 0 {
-                match event {
-                    Event::Key(KeyEvent {
-                        code: KeyCode::Down,
-                        ..
-                    })
-                    | Event::Key(KeyEvent {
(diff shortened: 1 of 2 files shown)


the-mikedavis:
How do you use the completion menu? Do you ever use arrow keys for it?

I think this change is too specific - it only prevents up/down from interacting once the completion menu has already opened and when the idle-timeout is exactly 0. It would be better to be able to unbind up/down for the completion menu or to prevent the completion menu from opening at all after any navigation in insert mode

archseer:
I think this also demonstrates that `idle-timeout` should never be set to `0` but a very small number.

p1gp1g:
At this moment, I have to set `idle-timeout` to `2000` and regularly tap `<C-x>` to see if there is a completion and I still often ends up tapping `<C-c>` because I've entered in the pop while moving around.

And with this change I see when there are suggestions and I do `<tab>` then arrows.

> I think this also demonstrates that `idle-timeout` should never be set to `0` but a very small number.

It sounds like going in favor of a different behavior if `idle-timeout` is set to `0` to me :)

archseer:
`idle-timeout` is used in other areas than just autocompletion, i.e. https://github.com/helix-editor/helix/commit/a85e3862988933953d1c0e7435969604d3722959

Setting it to `0` isn't recommended since you'll keep re-triggering various behaviors _immediately_ and waste a lot of CPU cycles. Even setting it to something as low as `2` or `5` would be better.

p1gp1g:
Ok, it makes sens. I think the [documentation](https://docs.helix-editor.com/configuration.html) should be updated (it recommends 0 to instant pop ups)

What about another config parameter like `arrows_to_enter_autocompletion` default to `true` ?

archseer:
This just seems like a very specific workaround to me. What about adding support for `C-o` instead? https://github.com/helix-editor/helix/issues/2027#issuecomment-1275809112

I'd also propose removing arrow key support from the completion popup and only using `C-n`/`C-p` as intended but I'm not sure I'm prepared for the [next wave of outrage](https://github.com/helix-editor/helix/pull/3671)

p1gp1g:
I think the issue linked i(#2027) is the same than the one I have : we want to see the pop up and choose when we to use it. So `C-o` doesn't work : i hides the pop up.

Actually I had the same issue on kakoune too and I ended up mapping `tab` and `s-tab` to use the completion and unmapping it as soon as the completion pop up is hidden. I have searched how to do `tab-arrows` without success.

p1gp1g:
0509c34 adds a parameter. I can then set `idle-timeout` to 20 and `arrows_to_enter_completion` to false : I can see the completion suggestions while still being able to move around in insert mode. :)

kendfss:
How about:
- creating `prev_completion` and `next_completion` commands, ideally defaulting to `shift+tab` and `tab`, and allowing for the completions window to be cancelled by line-changes?
- or changing the behaviour of `esc` so that it closes the first overlying sub-display, or [returns to] normal mode if no overlying display is present. Could also help with #5549 
Either of the above would allow for navigation to be decoupled from insertion without compromising the user's ability to [get the most out of the modal editing paradigm](https://github.com/helix-editor/helix/issues/1381#issuecomment-1001187403).

p1gp1g:
Closing in favor of #5803

PR #4146, opened 2022-10-07
Improve performance for files with lots of injections

### LINKED ISSUE #4139: Slow Markdown Highlighting
### Summary

Since the recent update to the markdown grammer by @David-Else, I have noticed extreme slowdown on markdown files. For example if I hold down backspace and then let go, helix keeps deleting for roughly 0.5seconds.
I have a very beefy PC so this should really not be happening and do not see issues for any other language.
To isolate problems I have disabled the LSP but the problem persists until I turn of syntax highlighting (`set-language text`) completly.

Looking at the flamegraphs (with LSP disabled) I noticed that a ton of time seems to be spend on parsing TS compared to other languages (rust in this example)

opening and deleting some text in a **md** file:

image

opening and deleting some text in a **rust** file:

image

### Reproduction Steps

Open any markdown file with syntax highlighting enabled and started deleting text

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

22.08.1-165-g4d4be0e8

pascalkuthe:
Files with large amounts of injections are incredibly laggy even on beefy CPUs.
For example `CHANGELOG.md` has about 800 injections, because the markdown grammar handles many items as injections.

Helix calls the TS parser for each layer on each keystroke.
TS incremental parsing relies a lot on not parsing unchanged text.
However for injections we call the parser manually defeating the incremental parser.
This PR fixes that problem by tracking in a `HashMap` which layers are modified.
The idea for this fix comes from the original PR (#3108) by @MDeiml that added the split markdown parser:

> It's related to this TODO:
> 
> https://github.com/helix-editor/helix/blob/681c0a91dc36f9db1d23bd31ac1b66414d2a6099/helix-core/src/syntax.rs#L724
> 
> Just going one step further and also avoid parsing layers that are after the edits

An edit can affect a layer in three different way:

* The edits occur **after the end** of the layer: These changes are irrelevant and ignored (no entry in `HashMap`)
* The edits occur **before the start** of the layer: The layer needs to be moved with `tree.edit()` but not re-parsed (`false` in `HashMap`)
* Any other edit modifies the content of the layer: The layer needs to be fully reparsed (`true` in `HashMap`)

These changes are pretty naive but very effective.
This completely fixes #4139 and I have not found any problems introduced by this change.

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -8,13 +8,15 @@ use crate::{
 };
 
 use arc_swap::{ArcSwap, Guard};
+use bitflags::bitflags;
 use slotmap::{DefaultKey as LayerId, HopSlotMap};
 
 use std::{
     borrow::Cow,
     cell::RefCell,
-    collections::{HashMap, HashSet, VecDeque},
+    collections::{HashMap, VecDeque},
     fmt,
+    mem::replace,
     path::Path,
     str::FromStr,
     sync::Arc,
@@ -594,6 +596,7 @@ impl Syntax {
             tree: None,
             config,
             depth: 0,
+            flags: LayerUpdateFlags::empty(),
             ranges: vec![Range {
                 start_byte: 0,
                 end_byte: usize::MAX,
@@ -656,9 +659,10 @@ impl Syntax {
                 }
             }
 
-            for layer in &mut self.layers.values_mut() {
+            for layer in self.layers.values_mut() {
                 // The root layer always covers the whole range (0..usize::MAX)
                 if layer.depth == 0 {
+                    layer.flags = LayerUpdateFlags::MODIFIED;
                     continue;
                 }
 
@@ -689,6 +693,8 @@ impl Syntax {
                                 edit.new_end_position,
                                 point_sub(range.end_point, edit.old_end_position),
                             );
+
+                            layer.flags |= LayerUpdateFlags::MOVED;
                         }
                         // if the edit starts in the space before and extends into the range
                         else if edit.start_byte < range.start_byte {
@@ -703,11 +709,13 @@ impl Syntax {
                                 edit.new_end_position,
                                 point_sub(range.end_point, edit.old_end_position),
                             );
+                            layer.flags = LayerUpdateFlags::MODIFIED;
                         }
                         // If the edit is an insertion at the start of the tree, shift
                         else if edit.start_byte == range.start_byte && is_pure_insertion {
                             range.start_byte = edit.new_end_byte;
                             range.start_point = edit.new_end_position;
+                            layer.flags |= LayerUpdateFlags::MOVED;
                         } else {
                             range.end_byte = range
                                 .end_byte
@@ -717,6 +725,7 @@ impl Syntax {
                                 edit.new_end_position,
                                 point_sub(range.end_point, edit.old_end_position),
                             );
+                            layer.flags = LayerUpdateFlags::MODIFIED;
                         }
                     }
                 }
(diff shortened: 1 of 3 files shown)


helix-core/src/syntax.rs
```diff
@@ -639,6 +639,8 @@ impl Syntax {
         // Convert the changeset into tree sitter edits.
         let edits = generate_edits(old_source, changeset);
 
+        let mut modified = HashMap::new();
```
archseer, line 642: Might be more efficient to use a `Vec` with `contains`? We don't expect this to contain a lot of elements.

If not, would it make sense to use a `HashSet` instead of a `HashMap`?

helix-core/src/syntax.rs
```diff
@@ -689,6 +692,7 @@ impl Syntax {
                                 edit.new_end_position,
                                 point_sub(range.end_point, edit.old_end_position),
                             );
+                            modified.entry(layer_id).or_insert(false);
```
archseer, line 695: Why not just skip the `or_insert(false)`?

helix-core/src/syntax.rs
```diff
@@ -743,16 +748,22 @@ impl Syntax {
...
-                        // If we applied them in order then edit 1 would disrupt the positioning of edit 2.
-                        tree.edit(edit);
+                    if let Some(&modified) = modified.get(&layer_id) {
+                        for edit in edits.iter().rev() {
+                            // Apply the edits in reverse.
+                            // If we applied them in order then edit 1 would disrupt the positioning of edit 2.
+                            tree.edit(edit);
+                        }
+
+                        if modified {
```
archseer, line 758: I'm confused how this works actually since `modified = true` only for the root layer

helix-core/src/syntax.rs
pascalkuthe: I want to only insert `false` if `true` is not already present (and if `false` is already present no update I necessary). Thats why I use `entry(..).or_insert()`. Just using `entry` does not modify the map itself. Hower `or_insert(false)` will insert false if the key is not already present

helix-core/src/syntax.rs
pascalkuthe: Sorry about that, I previously used two `HashSets` (one for `modified=true`, one for `modified=false`) and after testing refactorered to use a single `HashMap<bool>`. During that redactor I accidently used `entry(..).or_insert()` everwehere. 

Thus works because if the modification is before the start of the layer ( modified is `false`) `tree.edit`is enough to move the injection. Reparsing is only necessary if the content inside the injection actually changed

helix-core/src/syntax.rs
pascalkuthe: This will often contain lots of elements. For example if text is inserted at the start of the document, every injection is moved back so there is an entry for every element.

A `HashSet` is not sufficient there are three possible states:
* Injection does not need to be updated (no entry)
*  Injection needs to be moved (call tree.edit: false)
* Injection needs to be reparsed (call tree.edit, parse: true)


I switched to a SecondaryMap from slot_map (basically a dense map) but I will check if there is a more efficient approach

pascalkuthe: I think it would be more efficient to store this state (and the `touched` status) directly within the layer as a bitset. The memory overhead of a single integer per layer is hardly relevant and we don't need to allocate/deallocate and index any additional collections

helix-core/src/syntax.rs
pascalkuthe: Now with the bitset based approach its much clearer what is happening here

helix-core/src/syntax.rs
```diff
@@ -968,14 +982,19 @@ impl Syntax {
...
 #[derive(Debug)]
 pub struct LanguageLayer {
     // mode
     // grammar
     pub config: Arc<HighlightConfiguration>,
     pub(crate) tree: Option<Tree>,
     pub ranges: Vec<Range>,
-    pub depth: usize,
+    pub depth: u32,
+    pub flags: u32,
```
archseer, line 997: We already use the bitflags crate in other crates, maybe use that here?

helix-core/src/syntax.rs
```diff
@@ -869,7 +882,8 @@ impl Syntax {
             ts_parser.cursors.push(cursor);
 
             // Remove all untouched layers
-            self.layers.retain(|id, _| touched.contains(&id));
+            self.layers
+                .retain(|_, layer| replace(&mut layer.flags, 0) & LAYER_TOUCHED != 0);
```
archseer, line 886: Don't we need to reset these flags at the end/start of the loop?

helix-core/src/syntax.rs
pascalkuthe: I check this with cargo tree and it somehow didn't show up but now looking at the Cargo.lock by hand its there. Will update this to use the bitflags crate

helix-core/src/syntax.rs
pascalkuthe: I am not really sure what yoy mean. The `replace(.., 0)` does reset the flags here and retain calls the closure for every layer so we do end up resisting all flags.

It can not be done earlier then the retain because we still need to read the `touched` fiags here.

helix-core/src/syntax.rs
pascalkuthe: PR has been updated to use bitflags

helix-core/src/syntax.rs
archseer: Ah sorry, I missed that 👍

archseer:
Great work! 🎉

PR #4147, opened 2022-10-08
Fix append cursor location when selection anchor is at end of document

### LINKED ISSUE #4074: Append cursor positioning is not consistent during editing
### Summary

Append fails to position the cursor at the end of the selection

### Reproduction Steps

1. open hx
3. type `itest<esc>ba` # insert test, enter normal mode, back select the word, and append
4. the cursor will be properly positioned after the last 't'
5. `<esc>xditest<esc>ba` # delete the line and do it again
6. the cursor will now be positioned between the 's' and the 't'

### Helix log

nothing reported in the log

### Platform

Linux

### Terminal Emulator

st 0.8.5, gnome-terminal (GNOME Terminal 3.44.0 using VTE 0.68.0 +BIDI +GNUTLS +ICU +SYSTEMD

### Helix Version

helix 22.08.1 (6caa7a7f)

Graicc:
Fixes #4074. Cursor position was set to the wrong location when the anchor of a selection was at the end of a file. There was already a fix for the head of a selection being at the end, it just has to apply when the anchor is at the end as well.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2224,7 +2224,7 @@ fn append_mode(cx: &mut Context) {
         .iter()
         .last()
         .expect("selection should always have at least one range");
-    if !last_range.is_empty() && last_range.head == end {
+    if !last_range.is_empty() && (last_range.head == end || last_range.anchor == end) {
         let transaction = Transaction::change(
             doc.text(),
             [(end, end, Some(doc.line_ending.as_str().into()))].into_iter(),
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2224,7 +2224,7 @@ fn append_mode(cx: &mut Context) {
         .iter()
         .last()
         .expect("selection should always have at least one range");
-    if !last_range.is_empty() && (last_range.head == end || last_range.anchor == end) {
+    if !last_range.is_empty() && last_range.to() == end {
         let transaction = Transaction::change(
             doc.text(),
             [(end, end, Some(doc.line_ending.as_str().into()))].into_iter(),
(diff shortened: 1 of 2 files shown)


dead10ck (approved):
Looks good to me, thanks for the fix!

helix-term/tests/test/movement.rs
```diff
@@ -86,6 +86,25 @@ async fn cursor_position_newly_opened_file() -> anyhow::Result<()> {
...
+        helpers::platform_line("#[foobar|]#\n").as_ref()
+    ))
+    .await?;
+    
+    test((
+        "foo#[|]#",
+        "babar<esc>",
+        helpers::platform_line("#[foobar|]#\n").as_ref()
+    ))
+    .await?;
```
the-mikedavis, line 103: I think the `e` and `b` obscure the test cases here a little bit. Instead could we do those in the setup?

```rs
// Selection is forwards
test((
    "#[foo|]#",
    "abar<esc>",
    helpers::platform_line("#[foobar|]#\n").as_ref(),
))
.await?;

// Selection is backwards
test((
    "#[|foo]#",
    "abar<esc>",
    helpers::platform_line("#[foobar|]#\n").as_ref(),
))
.await?;
```

the-mikedavis (commented):
Looks like this needs a run of `cargo fmt` too

helix-term/src/commands.rs
```diff
@@ -2224,7 +2224,7 @@ fn append_mode(cx: &mut Context) {
         .iter()
         .last()
         .expect("selection should always have at least one range");
-    if !last_range.is_empty() && last_range.head == end {
+    if !last_range.is_empty() && (last_range.head == end || last_range.anchor == end) {
```
the-mikedavis, line 2227: This can be simplified a little bit using the `Range::to` method

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4156, opened 2022-10-09
fix: terminal freezing on `shell_insert_output`

### LINKED ISSUE #3524: Helix editr gets stuck when `:sh sh`
### Summary

Helix editr get stuck when `:sh sh`

### Reproduction Steps

 

I tried this:

1. `hx`
2. `:sh sh`

 this happened:

helix editor get stuck forever like this

| image |
| :--: |

### Helix log

Nothing in log file

```
```

</details>

### Platform

Linux

### Terminal Emulator

genome-terminal 3.44.1-1

### Helix Version

22.03-687-ge4c9d408

### LINKED ISSUE #4154: Terminal freezes after `shell_insert_output` or `shell_append_output` and a command that reads from stdin
### Summary

When using the shell keymaps `|` or `Alt+|` (`shell_insert_output` and `shell_append_output`) the editor will prompt for a command to run. If that command requires stdin (like `cat -`) however, the entire shell will freeze since the child process will inherit the `stdin` from the parent which is not the expected behaviour.

The fix for this is pretty simple and I will submit a PR shortly.

### Reproduction Steps

I tried this:

1. (in normal mode) `|cat -`

I expected this to happen:

Nothing (or the selection to be fed as input to the command's stdin).

Instead, this happened:

The entire terminal froze up as I couldn't send stdin for the child process (`cat -`) that started.

### Helix log

Not applicable

### Platform

Linux (WSL)

### Terminal Emulator

Alacritty

### Helix Version

helix 22.08.1 (c15f1ea2)

GabrielDertoni:
This bug occurs on `shell_insert_output` and `shell_append_output` commands.

The previous implementation would create a child process using the Rust stdlib's `Command` builder. However, when nothing should be piped in from the editor, the default value for `stdin` would be used. According to the Rust stdlib documentation that is `Stdio::inherit` which will make the child process inherit the parent process' stdin. This would cause the terminal to freeze.

This change will set the child process' stdin to `Stdio::null` whenever it doesn't pipe it. In the `if` statement where this change was made there was an extra condition for windows that I am not sure if would require some special treatment.

Fixes #3524.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4651,6 +4651,8 @@ fn shell_impl(
 
     if input.is_some() || cfg!(windows) {
         process.stdin(Stdio::piped());
+    } else {
+        process.stdin(Stdio::null());
     }
 
     let mut process = match process.spawn() {


the-mikedavis:
There's a small typo crate=>create in the commit message

the-mikedavis:
Windows seems ok in its current branch: commands like `:sh cmd` print to the popup and don't cause any hanging.

the-mikedavis (approved):

PR #4163, opened 2022-10-09
Add some vertical characters to the editor.indent-guides documentation

sumit-sahrawat:
This PR updates the default indent-guide character from a vertically aligned pipe (`|`) to a left aligned pipe (`▏`).

This makes a big difference visually as the indent guides are now aligned precisely to the indent level instead of starting half a character later.

* Before

before

* After

after

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -249,6 +249,6 @@ Example:
 ```toml
 [editor.indent-guides]
 render = true
-character = "╎"
+character = "╎", # Some characters that work well: "▏", "┆", "┊", "⸽"
 skip-levels = 1
 ```
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -249,6 +249,6 @@ Example:
 ```toml
 [editor.indent-guides]
 render = true
-character = "╎", # Some characters that work well: "▏", "┆", "┊", "⸽"
+character = "╎" # Some characters that work well: "▏", "┆", "┊", "⸽"
 skip-levels = 1
 ```


archseer:
Can you show a screenshot of the two?

sumit-sahrawat:
> Can you show a screenshot of the two?

Hello @archseer! Thanks for taking a look. I've added screenshots to the top comment. Was struggling to capture them earlier (nvidia issues).

the-mikedavis:
I prefer the center-aligned (current) character

dead10ck:
Yeah this seems like a preference thing that can be handled by the current config.

sumit-sahrawat:
> Yeah this seems like a preference thing that can be handled by the current config.

I agree, but this looks like a more reasonable default, at least to me. Setting the center aligned pipe is easier as it's available on the keyboard. Switching to the left aligned one requires a user to search for the particular Unicode character and then copy and paste it into their config.

That said, I'm fine with closing this as I've already updated my config :D

sumit-sahrawat:
> We could add a few variants to the documentation

This sounds like a good idea. Makes it easy to switch and helps discoverability as well. I've updated the PR to add some characters to the documentation instead of changing the default. The more supported character is a good default.

book/src/configuration.md
```diff
@@ -249,6 +249,6 @@ Example:
 ```toml
 [editor.indent-guides]
 render = true
-character = "╎"
+character = "╎", # Some characters that work well: "▏", "┆", "┊", "⸽"
```
the-mikedavis, line 252:
```suggestion
character = "╎" # Some characters that work well: "▏", "┆", "┊", "⸽"
```

sumit-sahrawat: Oops, thanks for catching this :+1:

the-mikedavis (approved):

the-mikedavis:
Thanks!

sumit-sahrawat:
Thank you, everyone! First commit to helix :partying_face:

PR #3958, opened 2022-09-25
Configurable labels for custom menus and typable commands

MattCheely:
I haven't done a ton of Rust development, and only a very little bit with Serde, so consider this an initial draft to see if the approach and proposed config format are sound.  I'm happy to add docs as well if this looks reasonable.

PR #4164, opened 2022-10-09
Make stickiness configurable

### LINKED ISSUE #1487: Set command list's name and sticky bit in a config
Helix allows you to group commands together to show a dialog list of commands — very helpful for keybind chaining
It also has two nice features about such dialogs: name and a sticky bit with the latter preventing the dialog from closing after a command is executed

I understand how to set the first part in my config, but then don't know how to change a name and add a sticky bit, they seem to be hardcoded in the `keymap.rs` only without the ability to add/change them from a config

```toml
  [keys.normal."a"]  # how to assign a custom name and make it sticky
   "b"	= "move_prev_word_end"
   "c"	= "move_prev_long_word_start"
```

bbodi:
Fixes #1487 partially.

Allows config such as
```
[keys.normal.W]
sticky = true
I   = "vsplit"
E   = "hsplit"
...
```

### DIFF
--- a/book/src/remapping.md
+++ b/book/src/remapping.md
@@ -50,5 +50,7 @@ Control, Shift and Alt modifiers are encoded respectively with the prefixes
 
 Keys can be disabled by binding them to the `no_op` command.
 
+Making a mode "sticky" can be achieved by adding `sticky = true` to the mapping.
+
 Commands can be found at [Keymap](https://docs.helix-editor.com/keymap.html) Commands.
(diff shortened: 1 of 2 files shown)


poliorcetics:
This probably interacts with #3958 to fully complete #1487

bbodi:
> This probably interacts with #3958 to fully complete #1487

Nice, unfortunately I haven't noticed it when I started my version. 
As I can see it would be easier to extend  #3958 with a simple 
```
"label" => label = map.next_value::<&str>()?,
"sticky" => sticky= map.next_value::<bool>()?, // Add this
```
and then set the is_sticky field later accordingly.

So this one can be closed.

Issue #4159, opened 2022-10-09
the desktop file is not easy to find


amtoine:
hello there :wave: :yum: 

i've been looking for the `.desktop` entry for `helix` to allow a file manager to open a file with `xdg-open`
thanks to #1073 and #2089, i found the [`Helix.desktop`](https://github.com/helix-editor/helix/blob/master/contrib/Helix.desktop) file, which does the trick :star_struck: 

## the problem i had
the thing is that, to me, it was not straightforward to find that file :confused: 
and i still had to `cp /path/to/helix/contrig/Helix.desktop ~/.local/share/applications` :thinking: 

## idea of enhancement
i think a few things could help people find out easily about the `.desktop` file :yum: 
- add a wiki entry for that
- add a documentation page for that
- add the `cp` operation to `/usr/share/applications/` in the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helix-git)

hope these thoughts will be usefull to the team in some way :relieved: 
cheers, amtoine :tada: :wink:

David-Else:
I think it would be best to add clear instructions in the README on how to add it, under installation.

amtoine:
> I think it would be best to add clear instructions in the README on how to add it, under installation.
>
ooh yes, the `README` is a great idea :ok_hand: 

i think the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helix-git) would be great too for Arch users as myself :relieved: 
it's something we'd like to have automatically available with either `xdg` or `rofi` and `dmenu` :yum:

amtoine:
**would you like some help to add these instructions?**
either in the `README` or in the [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=helix-git) :yum:

David-Else:
@amtoine I have added instructions, with some more bonus info on using other terminals, please check it out: https://github.com/helix-editor/helix/pull/4165

Does that cover everything?

David-Else:
@the-mikedavis I think you can close this now with https://github.com/helix-editor/helix/commit/7af7dadd17a45a22633cad2325ca3cf9ac815ca7

PR #4165, opened 2022-10-09
Add instructions for gnome desktop support using the `.desktop` file

David-Else:
This should address: https://github.com/helix-editor/helix/issues/4159

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,15 +68,15 @@ for a language.
 
 [![Packaging status](https://repology.org/badge/vertical-allrepos/helix.svg)](https://repology.org/project/helix/versions)
 
-## Gnome `.desktop` file support
+## Adding Helix to your desktop environment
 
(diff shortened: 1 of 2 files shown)


A-Walrus:
Quick fix, `.desktop` files are not specific to Gnome, they are used by any desktop environment that implements `XDG desktop menu` including Gnome, KDE, and many others.

David-Else:
@A-Walrus Thanks! I did not know that, I have updated the text, does it sound good to you now?

A-Walrus (approved):

the-mikedavis:
Could you add this note to the `book/src/install.md` page too?

David-Else:
@the-mikedavis I have done as you said, I went for just adding it to the install from source section and missed out the heading as I thought it fitted better there.

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4171, opened 2022-10-10
feat: paste from file

equals03:
This is an attempt to implement https://github.com/helix-editor/helix/issues/3796.

It is an alternative implementation to https://github.com/helix-editor/helix/pull/3966 using  a lot of the "paste" functionality that was already available (for better or worse). I elected to read the file "raw" without going through the Document infrastructure as eventually it was being read directly into a String and, because it reuses the `paste_impl` function already, decided to skip the overhead of processing the document first. Oh and as a bonus it supports multiple files.  

The command has been implemented as:  
`file-paste-after` and `file-paste-before`  
... to keep it in line with the current `:*-paste-[before|after|replace]` commands.

It does have basic file picker integration (which I am the most iffy with - comments/fixes expected and welcome!), using what seemed to be the easiest path - through the `Actions` enum. In particular, I'm looking for any suggestions as to what should happen in `helix-view/src/editor.rs`. It's not currently hit in any path so I've left it as a noop for now. Not great, but I just wanted something up for comments.

Beginning with rust so any code suggestions/critiques/guidance welcome. Additionally, any insight as to how to get `c-O` to work in the picker would be great 😄

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -36,6 +36,8 @@
 | `:clipboard-paste-after` | Paste system clipboard after selections. |
 | `:clipboard-paste-before` | Paste system clipboard before selections. |
 | `:clipboard-paste-replace` | Replace selections with content of system clipboard. |
+| `:file-paste-after` | Read a file from disk and paste afer selections. |
+| `:file-paste-before` | Read a file from disk and paste before selections. |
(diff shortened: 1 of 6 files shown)


pascalkuthe:
I am closing this PR as stale as it has a lot of conflicts. I also don't see us going in this direction. Instead we will more liekly add a simple `:read` command instead.

Thank you for contributing

PR #4172, opened 2022-10-10
Fix rendering of lines longer than 2^16

A-Walrus:
Before things would be cast to u16 earlier than needed, which would cause problems for insanely long lines (longer than 2^16 ~ 65 thousand)
In order to reproduce:
 - Open file with _really long_ lines, for example <http://tom7.org/abc/paper.txt>
- `65500l` (go 65500 to the left)
- hold `l` for a bit

Before it would disappear when you reached 2^16 (65536), now it doesn't.

**Note:** currently moving around when the cursor is very far left is very slow, especially in debug mode, this is an unrelated issue.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -441,7 +441,8 @@ impl EditorView {
                 return;
             }
 
-            let starting_indent = (offset.col / tab_width) + config.indent_guides.skip_levels;
+            let starting_indent =
+                (offset.col / tab_width) + config.indent_guides.skip_levels as usize;
             // TODO: limit to a max indent level too. It doesn't cause visual artifacts but it would avoid some
             // extra loops if the code is deeply nested.
 
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -443,16 +443,14 @@ impl EditorView {
 
             let starting_indent =
                 (offset.col / tab_width) + config.indent_guides.skip_levels as usize;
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

helix-view/src/editor.rs
```diff
@@ -558,7 +558,7 @@ impl Default for WhitespaceCharacters {
 pub struct IndentGuidesConfig {
     pub render: bool,
     pub character: char,
-    pub skip_levels: u16,
+    pub skip_levels: usize,
```
archseer, line 561: Unlikely that anyone will specify a `skip_levels` higher than `u8`. I'd store a `u8` here and cast it to `usize`  when calculating `let starting_indent`

helix-term/src/ui/editor.rs
```diff
@@ -442,17 +442,15 @@ impl EditorView {
...
-                    viewport.y + line,
-                    &indent_guide_char,
-                    indent_guide_style,
-                );
+            for i in starting_indent..(indent_level / tab_width) {
+                let x = (viewport.x as usize + (i * tab_width) - offset.col) as u16;
+                let y = viewport.y + line;
+                if !surface.in_bounds(x, y) {
+                    break;
+                }
```
archseer, line 452: This is an improvement but we still do a bunch of calculation for each `i` that's out of bounds.

Could we precompute the `starting_indent..end` that are in bounds before the loop?

archseer: It's ok to leave this as a follow-up, just add a TODO item here

archseer (approved):
Thanks!

PR #4176, opened 2022-10-10
Change twig tree-sitter grammar

gbprod:
https://github.com/eirabben/tree-sitter-twig seems to be abandonned. I suggest to use my  [twig tree-sitter parser](https://github.com/gbprod/tree-sitter-twig) instead :)

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -537,7 +537,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "twig"
-source = { git = "https://github.com/eirabben/tree-sitter-twig", rev = "b7444181fb38e603e25ea8fcdac55f9492e49c27" }
+source = { git = "https://github.com/gbprod/tree-sitter-twig", rev = "807b293fec3fead64f54c64fdf6fb05516c032b9" }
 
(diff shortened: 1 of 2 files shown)


EpocSquadron:
Tested this out with a large twig codebase I have lying around, this is pretty close! One thing I'm seeing is that with onelight theme things like `if` and `endif` are colored differently.

image

gbprod:
It's because it's highlighted with `@tag`, maybe we should use `@keyword` instead.
I update the pull request.
Thanks for your feedback

EpocSquadron:
Took a look again, didn't help. The `if` is a `conditional` in the grammar, whereas `endif` is a `tag`. I think the solution here is to special case `endif` in the grammar itself, and restore the tag highlighting to how you had it.

You can use `:tree-sitter-scopes` with the cursor on the element in question to see its parsed type.

gbprod:
I fixed it in grammar :)
Thanks for your usefull feedback!

the-mikedavis (approved):

the-mikedavis:
Looks good, thanks for picking up this grammar!

PR #4185, opened 2022-10-10
documentation: minor improvements to tutor, normalize how keys are called

softmoth:


runtime/tutor
```diff
@@ -488,7 +491,8 @@ _________________________________________________________________
 
  Type C to duplicate the cursor to the next suitable line.
 
- 1. Move the cursor to the first line marked '-->' below.
+ 1. Move the cursor to the first line below marked '-->'. Place
+    the cursor on the first word.
```
Omnikar, line 495: This change shouldn't be necessary; there is no need for the cursor to be placed specifically on the first word in this example.

runtime/tutor
```diff
@@ -663,15 +667,16 @@ _________________________________________________________________
...
  2. Make a change, insertion or appendage and repeat it with . .
- 3. Try using A-. with f and t, to select multiple sentences for
-    instance.
+ 3. Try using Alt-. with f and t, to select multiple sentences
+    for instance.
 
  --> This is some text for you to repeat things. You can repeat
-     insertions like changing words, or repeat selections like f/t.
+     insertions like changing words, or repeat selections like
+     f / t.
```
Omnikar, line 680: Make sure to delete blank lines if you add more lines of content, to keep the section spacing constant.
```suggestion
     insertions like changing words, or repeat selections like
     f / t.
```

runtime/tutor
```diff
@@ -663,15 +667,16 @@ _________________________________________________________________
...
```
Omnikar, line 679: Make sure to delete blank line(s) if you add line(s) of content, to keep the spacing between sections constant.

Omnikar:
I am concerned about changing `C-` and `A-` to `Ctrl-` and `Alt-` because that diverges from the format of keymap configuration.

David-Else:
@Omnikar I am not sure what you mean by the `format of keymap configuration` or why it should be relevant? I have never seen any technical documentation call `Ctrl-` and `Alt-` `C-` and `A-`. It certainly confused me when I first saw it.

Omnikar:
> @Omnikar I am not sure what you mean by the `format of keymap configuration` or why it should be relevant? I have never seen any technical documentation call `Ctrl-` and `Alt-` `C-` and `A-`. It certainly confused me when I first saw it.

When configuring keybindings, ctrl and alt modifiers are represented using `C-` and `A-`, like in Vim. For instance:
```toml
[keys.normal]
"A-o" = "jump_backward"
"C-k" = "expand_selection"
```
This is also how keys are represented in macro recordings.

David-Else:
> > @Omnikar I am not sure what you mean by the `format of keymap configuration` or why it should be relevant? I have never seen any technical documentation call `Ctrl-` and `Alt-` `C-` and `A-`. It certainly confused me when I first saw it.
> 
> When configuring keybindings, ctrl and alt modifiers are represented using `C-` and `A-`, like in Vim. For instance:
> 
> ```toml
> [keys.normal]
> "A-o" = "jump_backward"
> "C-k" = "expand_selection"
> ```
> 
> This is also how keys are represented in macro recordings.

Ah I see, thanks for pointing that out.

I don't think it is a problem. It is more important the documentation is clear and easy to understand. The `.toml` file and its fields are code/variables rather than English language. At the moment the documentation is all mixed up:

https://github.com/helix-editor/helix/blob/master/book/src/keymap.md

Within the space of a couple of lines there is `A-J` and `Alt-K`, and It does not say anywhere that the information is to be used for a configuration file.

runtime/tutor
softmoth: It must be past the '-->'. This is the simplest way to make that happen.

softmoth: But I see your point that it could confuse the reader, giving the impression that it *has* to start on the first word. A better replacement might be "Place the cursor somewhere after the '-->'."

softmoth:
> So it would be Ctrl, Shift, Del, and Spacebar.

That may be, and I have no strong preference other than consistency. I just went with what is already most common in the existing docs.

> I am concerned about changing `C-` and `A-` to `Ctrl-` and `Alt-` because that diverges from the format of keymap configuration.

I gave my thinking on this in the commit message. Existing docs use the Key name pretty much everywhere unless it is specifically talking about the Representation needed in config.toml.

runtime/tutor
softmoth: I changed it to "Place the cursor somewhere past the '-->'." I hope that conveys the right message.

runtime/tutor
softmoth: Thank you! I didn't realize this restriction. I've made a number of changes to get the sections back to the correct lengths.

softmoth:
> So it would be Ctrl, Shift, Del, and Spacebar.

That may be, and I have no strong preference other than consistency. I don't think I've ever seen "Spacebar" in technical documentation.

> I am concerned about changing `C-` and `A-` to `Ctrl-` and `Alt-` because that diverges from the format of keymap configuration.

I addressed this in the commit message. I believe this is the smallest change which produces a consistent text, and follows the style generally used in the existing documentation.

softmoth:
@David-Else, I overlooked a few other items you mentioned.

For "Visual mode", I copied the style that seemed already more common in the document. I can see an argument either way. I find it easier to read when they are capitalized, since "select" and "insert" aren't adjectives.

The Chicago style recommendations mention using capitals for "basic alphabet keys", which Helix explicitly does NOT do. I think it is a reasonable compromise to use `c` and `C` instead of `C` and `Shift-C`.

Especially the `tutor` document has unique constraints. It's plain text. Each page has to fit on a small screen. It has a casual, friendly tone with very minimal markup. So it would be fine to use different formatting there if needed, as long as it's consistent within the document itself.

Switching everything to use the Representation format might work well. It would be a big change, and would need an explanation of the format at the top of the documentation. Maybe something like "In this documentation, and in the Helix configuration files and commands, key names are case sensitive and modifier keys are written as `C-`, `A-`, and `S-` for Ctrl, Alt, and Shift. For example, `C-c` means Control + lowercase c, `A-'` means Alt-single quote, `S-ret` means Shift-Enter (Shift-Return), and `J` means capital letter J. See [remapping.md](book/src/remapping.md) for a complete reference."

the-mikedavis:
I think these changes look good; can you fix the conflict?

In the long run I think we should use representation format in the tutor and in the keymap doc in the book. It will need some preamble as you say to introduce it where it's used but I think it will end up being more confusing to have two representations for keys.

softmoth:
I fixed the conflict, which was just this one deleted keymapping:

```diff
commit 1a87cbd5082f410aee362a274edcc817bfb9f34a
Author: Pascal Kuthe <pascal.kuthe@semimod.de>
Date:   Sat Oct 1 19:35:49 2022 +0200

    remove filterting with C-space from picker

diff --git a/book/src/keymap.md b/book/src/keymap.md
index 85110634..9bc1c31e 100644
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -391,7 +391,6 @@ ## Picker
 | `PageDown`, `Ctrl-d`         | Page down         |
 | `Home`                       | Go to first entry |
 | `End`                        | Go to last entry  |
-| `Ctrl-space`                 | Filter options    |
```

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4188, opened 2022-10-10
Show selection over bracket matching

pathwave:
Currently the bracket matching was covering selection edges, making it impossible to see if a bracket was selected or not.
This PR disables bracket matching when the selection edge is over the matching bracket.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -604,11 +604,11 @@ impl EditorView {
             use helix_core::match_brackets;
             let pos = doc.selection(view.id).primary().cursor(text);
 
-            let pos = match_brackets::find_matching_bracket(syntax, doc.text(), pos)
-                .and_then(|pos| view.screen_coords_at_pos(doc, text, pos));
-
-            if let Some(pos) = pos {
-                // ensure col is on screen
+            if let Some(pos) = match_brackets::find_matching_bracket(syntax, doc.text(), pos)
+                // Don't render over the primary selection.
+                .filter(|pos| !doc.selection(view.id).primary().contains(*pos))
+                .and_then(|pos| view.screen_coords_at_pos(doc, text, pos))
+            {
                 if (pos.col as u16) < viewport.width + view.offset.col as u16
                     && pos.col >= view.offset.col
                 {


archseer:
This seems more of a theme issue? It should be possible to visualize both the matching bracket and the selection cursor at the same time.

CptPotato:
> This seems more of a theme issue? It should be possible to visualize both the matching bracket and the selection cursor at the same time.

You're right. It is possible but then you can't set both `fg` and `bg` for the bracket matching:

with `fg` and `bg` (selection on `)` is not visible):
image

with `fg` only:
image

I can't comment on the implementation but I think it's a decent change.
Then again, I never found this to be an issue myself.

helix-term/src/ui/editor.rs
```diff
@@ -604,22 +604,27 @@ impl EditorView {
...
+                            let style = theme.try_get("ui.cursor.match").unwrap_or_else(|| {
+                                Style::default()
+                                    .add_modifier(Modifier::REVERSED)
+                                    .add_modifier(Modifier::DIM)
+                            });
+
+                            surface[(viewport.x + pos.col as u16, viewport.y + pos.row as u16)]
+                                .set_style(style);
+                        }
+                    }
```
the-mikedavis, line 627: There are a few nested `if let Some`s here that can be refactored to reduce the nesting:

```rs
if let Some(pos) = match_brackets::find_matching_bracket(syntax, doc.text(), pos)
    // Don't render over the primary selection.
    .filter(|pos| !doc.selection(view.id).primary().contains(*pos))
    .and_then(|pos| view.screen_coords_at_pos(doc, text, pos))
{
    if (pos.col as u16) < viewport.width + view.offset.col as u16
        && pos.col >= view.offset.col
    {
        let style = theme.try_get("ui.cursor.match").unwrap_or_else(|| {
            Style::default()
                .add_modifier(Modifier::REVERSED)
                .add_modifier(Modifier::DIM)
        });

        surface[(viewport.x + pos.col as u16, viewport.y + pos.row as u16)]
            .set_style(style);
    }
}
```

pathwave: Applied the suggested changes

pascalkuthe:
I am closing this PR as stale as it has conflicts and I don't see us going in this direction.

Thank you for contributing!

PR #4192, opened 2022-10-11
refactor search functions

thaqibm:
Refactored the search functions in ``helix-core``. 
Possible improvements: Can implement a bidirectional search function with an enum for direction.

### DIFF
--- a/helix-core/src/search.rs
+++ b/helix-core/src/search.rs
@@ -28,14 +28,14 @@ pub fn find_nth_next<M: CharMatcher>(
     }
 
     let mut chars = text.chars_at(pos);
-    let mut count = 0; 
+    let mut count = 0;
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Is there a function change other than `find_nth_prev` taking a `CharMatcher`? The original `for` + `loop` is nice because it doesn't need an extra `count` variable

thaqib0437:
I found it a bit hard to figure out the intentions of the for + loop. 
Also noticed the function signature was different while refactoring so I added it.

the-mikedavis:
I prefer the current `for` + `loop` so I will close this out. In general we avoid refactors / "clean up" changes since they confuse `git blame` and can cause unnecessary conflicts with other PRs

Issue #3116, opened 2022-07-20
Go to next LSP diagnostic in workspace


rishflab:
#### Describe your feature request/potential bug

The helix docs show there is different commands for `jump to next diagnostic in document` and  `jump to next diagnostic`.

I have only been able to `jump to next diagnostic in document`. 

Would it be possible for:
1.  `jump to next diagnostic` to become `jump to next diagnostic in workspace`
2.  `jump to next diagnostic in document` to be renamed to `jump to next diagnostic in buffer`.

sohnryang:
Where did you find the command for `jump to next diagnostic in document`? I couldn't find it in the docs.

matoous:
> Where did you find the command for jump to next diagnostic in document? I couldn't find it in the docs.

It's actually called `Goto next diagnostic` and `Goto previous diagnostic`, binded to `[d` and `]d` respectively. You can find both using `space + ?` command pallete picker.

PR #4201, opened 2022-10-11
Jump to next/prev diagnostic in workspace (#3116)

dariooddenino:
Implements functions and keybindings to move to next/previous/first/last diagnostic in the workspace as per #3116
|Key|Description|
|---|---|
| [w | go to previous diagnostic in workspace |
| ]w | go to next diagnostic in workspace |
| [W | go to first diagnostic in workspace |
| ]W | go to last diagnostic in workspace |

I've also refactored some functions a little bit to reduce code duplication.

It's not entirely clear to me what's the clear cut between a function being in `commands.rs` or in `commands/lsp.rs`. I tried following the current code, but let me know if I have to move something.

There's some code repetition going on in `get_next_diag_doc`. I tried finding ways to handle the fact that the direction branches produce an `Iter` and a `Rev`, but the only solutions I found were based on external libraries not currenlty used (itertools, either).

Please, feel free to correct anything wrong here and suggest better practices.


David-Else:
I think `]d` and `]D` would make more sense, it seems to be the convention with document vs workspace: https://github.com/helix-editor/helix/blob/master/book/src/keymap.md#space-mode ... **but** `D` it is already used for `Go to first diagnostic in document ` :( Maybe replace it as moving around the whole workspace is much more useful than just the first and last item in the document? 

Changing diagnostics to `w` does not make sense to me.

EpocSquadron:
The space menu has `g` and `G` for diagnostics picker. Seems strange to have three different letters for related functionality.

dariooddenino:
Yeah I agree, ideally the same letter for everything would be better.

I wonder if the correct thing would be to replace the `[` `]` prefixes with something else, since right now they imply navigation inside the current document.
For example having `<d` `>d` `<D` `>D`, with `<` and `>` being the equivalent to the ones above for workspace movement. But then the question is if there are other keys that would benefit from this or not.

Having `[D` as `goto previous diagnostic in workspace` and `]D` as `goto next diagnostic in workspace` is a very good idea.

sudormrfbin:
The existing `]D` binding for goto last diagnostic could be bound to `]<c-d>` (`]` followed by `ctrl-d`) and the workspace variant could occupy `]D` on the premise that the latter would be more commonly used.

georgesboris:
New user here - the `g` and `G` on the space menu for diagnostics was zero intuitive. Since everyone here is also pushing for newer `d` based diagnostic commands shouldn't the current space menu shortcuts be revised maybe? I don't know how much deep the project is into no-breaking-changes yet - just a thought :)

andreytkachenko:
That's would be awesome if it will jump to errors first and then to warnings and so on

Tudyx:
What's the current blocker for this? I can help finish the work if needed.

Maybe we can provide only the commands for now, so users can bind them to whatever they want until we find a good default.

When making a change in a Rust codebase, I often just let the compiler guide me to see what other places are affected by my change. If I could jump from diagnostic to diagnostic across files, without opening the workspace diagnostic picker each time, it will be awesome.

dariooddenino:
> What's the current blocker for this? I can help finish the work if needed.
> 
> Maybe we can provide only the commands for now, so users can bind them to whatever they want until we find a good default.
> 
> When making a change in a Rust codebase, I often just let the compiler guide me to see what other places are affected by my change. If I could jump from diagnostic to diagnostic across files, without opening the workspace diagnostic picker each time, it will be awesome.

Not sure, I think it's just a matter of no one having the time to review this since it's probably near the bottom of priorities.
You can pick it up without problems! I haven't been writing rust for a long while, and I'm not sure I want to get back into it :smile:

PR #4203, opened 2022-10-11
Dead code

Borwe:
The for loop is unecassry, as the condition is already taken care of in the while loop.

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -76,12 +76,6 @@ impl Args {
                 arg => args.files.push(parse_file(arg)),
             }
         }
-
-        // push the remaining args, if any to the files
-        for arg in argv {
-            args.files.push(parse_file(&arg));
-        }
-
         Ok(args)
     }
 }


archseer:
Needs a rebase with master

Borwe:
Okay, rebased.

the-mikedavis (changes requested):
This block isn't dead: it handles any files after the "--" in the argv: https://github.com/helix-editor/helix/blob/f4d96b2ca335977ea9b9031e35994b7f15f05c25/helix-term/src/args.rs#L31

Borwe:
> This block isn't dead: it handles any files after the "--" in the argv:
> 
> https://github.com/helix-editor/helix/blob/f4d96b2ca335977ea9b9031e35994b7f15f05c25/helix-term/src/args.rs#L31

Ooh, didn't know that, when is it used though? I think we can remove it too?

Borwe:
Nevermind, I see my error in logic. Closing.

Issue #768, opened 2021-09-19
Is there a whichwrap option?


irevoire:
Hello,

I would love to have a [whichwrap](http://vimdoc.sourceforge.net/htmldoc/options.html#%27whichwrap%27) option like in vim.
Basically, the idea would be that, when I'm at the end of a line and type `right`, I should just stay at the end of my line.
The same goes for the start of the line and `left`.

Is this something planned?

kirawi:
From the documentation, it sounds like the `Home` (next line) and `End` (prev line) keys. https://docs.helix-editor.com/keymap.html

archseer:
@kirawi you should read the linked documentation, it's about configuring this behavior, not about specific keys :P

archseer:
There's no configuration option and I'm not sure if we want this to be configurable. 

Currently everything wraps: h/l used to stop at the end of line but after a discussion with @cessen we removed this limitation and had it match kakoune. I personally don't use this wrapping behavior, but what's the rationale for explicitly disabling it?

irevoire:
I would love to see this behaviour back, it was one of the reason why I went back to vim after using kakoune for ~6 months ☹️

(But overall I'm really in love with helix and don't plan to quit soon even if this behaviour is never implemented! Thanks for all the great works!)

Omnikar:
I personally think this can be useful, if a bit niche, but I don't really think there's much of a reason *not* to make this configurable.

yowayb:
@irevoire just curious what the benefit would be?

ArijanJ:
> curious what the benefit would be?

Just preventing complexity, IMO.

We've always had 4 directional keys, and making half of them go diagonally only at certain times is plain confusing.

irevoire:
Yeah, especially when you have multiple cursors and by hitting `right` one goes to the next line :weary:

hholst80:
@pantos9000 is your code something that is useful yet for the general public, or more specifically, is it something I dare to test without having my dreams and expectations shattered? :)

pantos9000:
@hholst80 Yup, this is ready and should work as expected. Just did a small test on my side to confirm. Well, at least it works as expected by me :stuck_out_tongue_closed_eyes: 

Note that my PR #10576 introduces two different kinds of movement sets: `same_line` (solving this issue) and `anchored` movement. If you want, I can also split up the PR - I kept the logic for both in separate commits, so having two separate PRs won't be a big deal for me.

Issue #1638, opened 2022-02-09
Making "x" extend selection


premell:
In the keymap discussion thread I wrote some suggestions of how to improve the keybindings of Helix https://github.com/helix-editor/helix/issues/165#issuecomment-1033690249

Many of my proposed changes are subjective, however I think my suggestion of making x into line wise select mode doesn't have many downsides. By that I mean changing "x" into "vx".

The reason we need a visual mode at all is because the selection keys are also movement keys, which means you a way to determine if the user wants to extend the selection or simply move. But "x" isn't a movement key, so I cant think of a single use case where you would want to use "x", but also wouldn't want to extend selection. Either you use x to simply select the line for an operation like "d" or "y", in which case making "x" into extended selection doesn't matter. Or you want to select select multiple lines, in which case you need to use "vx". Also it doesnt really break consistency since "x" is a quite unique key already.

I might be wrong or missed something though, so I'd be happy to hear your thoughts

dead10ck:
I'm very confused by this suggestion. vx is not a line wise mode, at least not as that term is used in vim. vx would simply turn on select mode and then select the current line. After that, moving the cursor up and down does not select whole lines. Helix doesn't currently have a line wise select mode. If you want to select multiple lines, you would hit x repeatedly, or v, then move the cursor up or down to the desired line, then x.

premell:
Okay sorry my bad, I misunderstood how vx worked. However I don't think it changes that much, I still think x should be vx, because there is no situation that I can think of where you wouldn't want extended select when you use x.

premell:
> I've read this issue yesterday, at first I think this gonna be annoying too but then I realized that delete/copy in select mode will auto go back to normal mode so there won't different any quick linewise action like xd, xy...

Yes exactly! Either you want to do an operation (which leaves extended selection mode), or you want to move in which case you want extended select mode (otherwise there was no reason to hit x in the first place

dead10ck:
Oh I see, yeah, you're right, this wouldn't affect quick line operations. Still, I don't really understand the improvement it makes. If you want to select multiple whole lines, why bother pressing x at the beginning of the motion at all? Why not just hit v and start moving up or down to the line you want? What is the difference between vkkkx and vxkkkx?

And also, I think if you really wanted this, you could make your own key binding for it.

```toml
x = [ "select_mode", "extend_line" ]
```

premell:
[...]

Issue #3076, opened 2022-07-15
Add option to not have newlines selectable


d12bb:
There already has been some discussion on newline characters being treated like any other in #585 and #2956. As this is a very frequent stumbling stone for me as well, I'd like to propose adding an option to treat newline characters special. With this new option set, Helix would:

- not allow the cursor be on a newline character
- not select it on `x`, unless it's in between multiple lines that are selected
- select it additionally on `xd` to delete the whole line

Reasoning:

- Beyond Kakoune, I don't know any editor behaving this way, so it would be more in line with what everyone else does, making it easier to switch.
- While being more consistent per se, the current behavior is less consistent with expectations and thus less intuitive.
- Some obviously seem to like the current behavior for being consistent, so make it optional.

archseer:
This gets even more inconsistent: if you place the cursor on an empty line, you're literally placing the cursor onto the EOL character. Disabling this would mean a bunch of workarounds for empty lines.

d12bb:
If I understand #376 correctly, the cursor actually internally is between the `\n`s of the current and the line above, so this wouldn't be inconsistent at all?

sudormrfbin:
#376 allows expressing ranges as positions between characters, but cursors in helix always have a width of 1. This is why pressing `d` with a normal cursor deletes the character under it -- it is simply deleting the single width selection under it.

the-mikedavis:
The current behavior is quite elegant when it comes to selections: you can treat newline like any other character so you can do things like `t<ret>` to select up until a newline or sub-select newlines in a selection.

Normally I would agree that behavior should be customizable but in this case I think it's out of scope: as archseer says, in order to make this configurable, we'd need to add a bunch of workarounds which would be pretty hacky. Ultimately this is a smaller but comparable problem as having an evil mode (vim bindings). Maintaining competing paradigms is a pain so it's out of scope

adsick:
When you say it like so for me it sounds like helix aims to be **consistently bad**, guys in #585 correctly point out that there is no point in looking at some thing called "kakoune", the point is to make a good editor and good editor should have a good design and intuitive UX.
When good decisions are "inconsistent" you are in fact doing the whole thing wrong.
@archseer

archseer:
@adsick Thanks for the ping! You are being rude again, so I'm not going to respond. You are welcome to start a polite discussion next time.

taupiqueur:
Kakoune used to treat the EOL specially, but changed this behavior during its development.

[...]

Issue #2956, opened 2022-07-04
Unexpected behavior because of newline characters being selectable


Nughm3:
### Summary

This is a small nitpick I have with Helix, coming from Neovim. Helix counts the last newline character `\n` as a selectable character.

Helix:
![image](https://user-images.githubusercontent.com/83223650/177134267-a5fe8c18-26e2-4692-9a9a-796ba1af22e3.png)

Neovim:
![image](https://user-images.githubusercontent.com/83223650/177134315-3f91ae4a-6f3a-412a-a7f3-b763cf9adbf4.png)

As seen above, the last character you can select in Neovim is the semicolon, but in Helix you can select the newline. This causes some keys, such as `a` or `xc` to work differently than I would expect. **See "reproduction steps" below.**

I recognize that this issue is a bit opinionated. If this is behavior is intentional, I'd like to know why. Thanks!

### Case 1
![image](https://user-images.githubusercontent.com/83223650/177134267-a5fe8c18-26e2-4692-9a9a-796ba1af22e3.png)
After pressing `a`, Helix recognizes that the next character after `\n` is the `}` on the next line, and inserts there instead.
![image](https://user-images.githubusercontent.com/83223650/177135036-e722e09c-562b-449f-979b-bac084bc0eb1.png)
Normally, I would intend to append text at the end of line 2 instead of inserting at the start of line 3. This always annoys me, as I would have to go back up or just use `A` instead, whereas in other editors like Neovim I'm used to it appending text at the end of the line.

### Case 2
I often find myself having to change the whole line. In Vim this would be `cc`, and in Helix it intuitively should be `xc`. 
![image](https://user-images.githubusercontent.com/83223650/177135975-3d4642e9-9a1e-4ec9-98b9-1834938fa629.png)
However, `x` selects the whole line *including* the newline character, so when I press `c` it deletes that too and shoves the next line onto the current one:
![image](https://user-images.githubusercontent.com/83223650/177136136-762353ef-f589-45aa-be0b-f782700e4a6d.png)

By the way, `gl` *doesn't* select the newline, which is inconsistent with `x`. (Correction: see 2nd comment - there's a reason for the inconsistency)

### Suggestions:
- Don't let the user select `\n`
- Change `a` to stop at line boundaries
- Change `x` to be like `ghvgl` (go to line start, select mode, go to line end not including `\n`)

### Platform

Linux (Arch x86-64)

### Terminal Emulator

wezterm 20220624-141144-bd1b7c5d

[...]

PR #4204, opened 2022-10-11
Implement Vim like movement commands

mitsuhiko:
This adds a few new movement commands to allow `hjkl` to be rebound so that they follow more the movement of vim. left/right movement cannot move between lines and up/down either follows newline or does not accidentally move onto the newline character. See #4268 for motivation. This however does not emulate vim, it just has some vim inspired movements. In particular it retains the entire behavior of Helix with regards to how newlines are otherwise handled (they stay selectable, and these alternative movement commands to not interfere with other commands in use or change their assumptions).

vim-movement mov

The following commands are added:

* `move_line_up_anchored` (bind to `k`): implements anchored movement up
* `move_line_down_anchored` (bind to `j`): implements anchored movement down
* `move_char_left_same_line` (bind to `h`): implements same-line movement left
* `move_char_right_same_line` (bind to `l`): implements same-line movement right
* `extend_line_up_anchored` (bind to `k`): implements anchored selection extension up
* `extend_line_down_anchored` (bind to `j`): implements anchored selection extension down
* `extend_char_left_same_line` (bind to `h`): implements same-line selection extension left
* `extend_char_right_same_line` (bind to `l`): implements same-line selection extension right

## Bindings

```toml
[keys.normal]
j = ["move_line_down_anchored"]
k = ["move_line_up_anchored"]
h = ["move_char_left_same_line"]
l = ["move_char_right_same_line"]

[keys.select]
j = ["extend_line_down_anchored"]
k = ["extend_line_up_anchored"]
h = ["extend_char_left_same_line"]
l = ["extend_char_right_same_line"]
```

## Same Line Movement

The default movement behavior of helix will happily place you in the next line. Vim users are not used to that and are likely to prefer the cursor not to move into the next line (that is at least my expectation). This is what `move_char_left_same_line` and `move_char_right_same_line` do. They will however let you move to the rightmost character (the newline) which is not what vim does (except of `virtualedit=oneline` is used). However this pairs up nicely with the anchored movement (see next).

## Anchored Vertical Movements

The two commands `move_line_up_anchored` and `move_line_down_anchored` implement what I would call anchored movement. If you are on the newline character (eg: one to the right of where `goto_line_end` would place you) then as you move up and down you end up on the newline of the target line again.  This emulates the behavior of vim of what happens if you move to the line end with `$`. If however you are placed to the left of the newline, as you move up and down you will not move onto the newline.

The main difference is that in vim if you use `$` and go up/down you stay anchored to the line end. In Helix with this change you need to do `gll` to stay anchored to the newline. I do however have to say that I prefer this actually because in Vim you do not actually know if you are newline anchored visually. That is as far as I can tell an internal state.

I played around with this a bit now and I really like the combination of the two. It stays true to the helix character model, but it also feels very natural for vim users.

**Limitation:** Movements across empty lines are ambiguous. To help with this I put the horizontal column to the largest possible integer on the range to tell these cases apart. From what I can tell this does not create negative impact and avoids extra complication on the range.

## Notes on Mixed Use

I think these move/extend commands are interesting to use independent of each other. For instance I find the anchored newline movement functionality useful even if one does not like restricting the movements within a line. For instance line based selection works really nice as once the visual selection is on the newline, any further moving down will continue to capture entire lines. That means for instance `vx` will continue to move downwards which I think also addresses some points of #1638 as it basically keeps you in line mode for as long as you don't move left.

### DIFF
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -131,7 +131,7 @@ pub fn move_vertically_anchored(
     // cases we take advantage that anchored movement sets the horizontal column
     // to the largest possible integer.
     let newline_move_mode = pos == line_end_char_index(&slice, line)
-        && (horiz == !0 || !rope_is_line_ending(slice.line(line)));
+        && (horiz == u32::MAX || !rope_is_line_ending(slice.line(line)));
 
     // Compute the new position.
     let new_row = match dir {
@@ -153,7 +153,7 @@ pub fn move_vertically_anchored(
 
         // when in newline_move_mode we set the horizontal column to the largest
         // possible integer to be able to disambiugate movements across empty lines.
-        (new_pos, !0)
+        (new_pos, u32::MAX)
 
     // otherwise move up and down like normal but ensure to never place the
     // cursor on the newline character.
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -104,11 +104,13 @@ pub fn move_vertically(
 
 /// Implements anchored vertical movement.
 ///
-/// Anchored movement behaves differently depending on where the cursor is placed.  If
-/// the cursor is currently on the end of the line (the newline character) then when it
-/// is moving up and down, it stays anchored to the newline as it moves up and down.
+/// Anchored movement behaves differently depending on where the cursor is placed with
+/// regards to newlines.  If the cursor is currently on a newline, then moving up and
+/// down will place you again on the newline of that line.
 ///
-/// Otherwise when it moves up and down it will not move unto the newline character.
+/// If you are not on the newline, then moving up and down will move you to the same
+/// column except if that column is further to the left from where you are.  In that case
+/// it places you on the rightmost column that is not a newline.
 pub fn move_vertically_anchored(
     slice: RopeSlice,
     range: Range,
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -123,15 +123,16 @@ pub fn move_vertically_anchored(
     let pos = range.cursor(slice);
     let line = slice.char_to_line(pos);
 
-    // if we are resting on a newline character and we did not just move
-    // across an empty line where the newline_mode_mode flag was negativley
-    // carried we are in newline movement mode.
-    let newline_move_mode = pos == line_end_char_index(&slice, line)
-        && (range.newline_move_mode || slice.line(line).len_chars() != 1);
-
     // Compute the current position's 2d coordinates.
     let Position { row, col } = visual_coords_at_pos(slice, pos, tab_width);
     let horiz = range.horiz.unwrap_or(col as u32);
+
+    // if we are resting on a newline character and we did not just move
+    // across an empty line we are in newline movement mode.  To disambiugate these
+    // cases we take advantage that anchored movement sets the horizontal column
+    // to the largest possible integer.
+    let newline_move_mode = pos == line_end_char_index(&slice, line)
+        && (horiz == !0 || slice.line(line).len_chars() != 1);
 
     // Compute the new position.
     let new_row = match dir {
@@ -141,7 +142,7 @@ pub fn move_vertically_anchored(
 
     // if we are already on a newline character, we want to navigate to the
     // newline character on the new line we moved to.
-    let (new_pos, newline_move_mode) = if newline_move_mode {
+    let (new_pos, new_horiz) = if newline_move_mode {
         let new_pos = line_end_char_index(
             &slice,
             slice.char_to_line(pos_at_visual_coords(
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Newlines being selectable like any other character is intended behavior and I don't like adding options or alternate movement commands to skirt them. See https://github.com/helix-editor/helix/issues/2956 and https://github.com/helix-editor/helix/issues/3076.

mitsuhiko:
@the-mikedavis this does not change the behavior about select-ability. They are still selectable as before. The new commands add a way that for horizontal movements you can stay constrained within the line but you still move onto them and for vertical movement it even anchors you onto them if you already are on one.

archseer:
(Also fixes https://github.com/helix-editor/helix/issues/768)

mitsuhiko:
Curious about what the odds are of this merging. I'm happy to incorporate necessary changes if there is something I can do here.

helix-core/src/movement.rs
```diff
@@ -80,6 +102,84 @@ pub fn move_vertically(
...
+    // Compute the current position's 2d coordinates.
+    let Position { row, col } = visual_coords_at_pos(slice, pos, tab_width);
+    let horiz = range.horiz.unwrap_or(col as u32);
+
+    // if we are resting on a newline character and we did not just move
+    // across an empty line we are in newline movement mode.  To disambiugate these
+    // cases we take advantage that anchored movement sets the horizontal column
+    // to the largest possible integer.
+    let newline_move_mode = pos == line_end_char_index(&slice, line)
+        && (horiz == !0 || !rope_is_line_ending(slice.line(line)));
```
the-mikedavis, line 134:
```suggestion
        && (horiz == u32::MAX || !rope_is_line_ending(slice.line(line)));
```

helix-core/src/movement.rs
```diff
@@ -80,6 +102,84 @@ pub fn move_vertically(
...
+            slice.char_to_line(pos_at_visual_coords(
+                slice,
+                Position::new(new_row, col),
+                tab_width,
+            )),
+        );
+
+        // when in newline_move_mode we set the horizontal column to the largest
+        // possible integer to be able to disambiugate movements across empty lines.
+        (new_pos, !0)
```
the-mikedavis, line 156:
```suggestion
        (new_pos, u32::MAX)
```

the-mikedavis (commented):
Looking at your comment here (https://github.com/helix-editor/helix/pull/4204#issuecomment-1279345053), Vim whichwrap and the context in https://github.com/helix-editor/helix/issues/768, I think these motions are ok and don't introduce incompatible Vim-isms 👍. (Sorry for jumping the gun on "no Vim" here; steam preemptively comes out of my ears when I see "Add Vim-like..." 😅)

mitsuhiko:
An open question I have is how this should be documented best. Is there a place to document unmapped commands that do not have a typable alias?

the-mikedavis:
Not currently, see: https://github.com/helix-editor/helix/pull/997

mitsuhiko:
Then I suppose this is ready.

the-mikedavis (approved):

archseer:
Sorry for the delay, this looks fine to me but needs a rebase now that the soft wrap changes landed!

pascalkuthe:
The rebase will be slightly more challenging than usual because the vertical movement commands were essentially reimplemented in #5420. It would also be good to have a version that does visual line movement and document line movement like we have for the normal movement command

mitsuhiko:
I will update this. Unfortunately the changes are a bit more involved so it will take a bit until I find the time for it.

yashpalgoyal1304:
i was also thinking about this _topic_ the other day - that why don't n/vi/m stop the horizontal movement at line ends and don't wrap it around - is there any benefit to that way while editing etc...

mkblast:
I think a better way to configure it is to add all of those binding into one option.

helix-core/src/movement.rs
```diff
@@ -80,6 +102,84 @@ pub fn move_vertically(
...
+/// Implements anchored vertical movement.
+///
+/// Anchored movement behaves differently depending on where the cursor is placed with
+/// regards to newlines.  If the cursor is currently on a newline, then moving up and
+/// down will place you again on the newline of that line.
+///
+/// If you are not on the newline, then moving up and down will move you to the same
+/// column except if that column is further to the left from where you are.  In that case
+/// it places you on the rightmost column that is not a newline.
+pub fn move_vertically_anchored(
```
pickfire, line 114: Can we have some tests for both of these functions?

pascalkuthe:
I would definitely like to still see this feature. Unfortunately, this branch is quite stale as the refactor in #5420 entirely changed how movement works so somebody would need to start almost from scratch. Thank you for contributing!

Issue #1762, opened 2022-03-06
Make * also select the word under the cursor


antoyo:
#### Describe your feature request

I would find it more convenient if `*` would also select the word under the cursor before doing a search.
The search should also be on word boundary (wrapped with `\b`).
That's the behavior of vim.

We can keep the current `search_selection` as it is, perhaps under another shortcut.

antoyo:
Now that I think of it, perhaps we only need to have a command for the `<C-s>` shortcut in the prompt.
That would allow use to remap `*` to the following:
```toml
[keys.normal]
"*" = ["search", "insert_word"]
```
What do you think?

ttys3:
currently `*` will set register `/` to current char, which I think is not the most freq case.

it's good to make it default select the word.

zim0369:
This is really necessary!

mitsuhiko:
At the very least I would expect `*` to actually execute the search. Right now to do what `*` does in VIM which is a pretty common operation you need to do something like `alt+o` followed by `*` and then `n`.

mitsuhiko:
I ended up with this binding BTW that mostly emulates what I'm currently used to from vim:

```toml
[keys.normal]
"*" = ["shrink_selection", "search_selection", "search_next"]
```

ubitux:
For the record, `#` and `*` in vim (respectively backward and forward search of symbol under the cursor) is incredibly useful to read code as it helps tracking usages of a given symbol. Here is a demo of a typical symbol tracking workflow in nvim using `#` and `*`:

https://user-images.githubusercontent.com/34467/218343565-646523d4-2fe4-4416-8523-f60fae98356b.mp4

I'm guessing https://github.com/helix-editor/helix/pull/5702 will also help here.

milesfrain:
Looks like this is being tackled in #6046

Issue #1733, opened 2022-03-03
Highlight search results


antoyo:
Hi.

#### Describe your feature request

I'd like an option to highlight the search results.
Thanks.

kirawi:
I think this functionality was recently merged with https://github.com/helix-editor/helix/pull/1635

antoyo:
That seems to only highlight the search results in the file picker.

I was talking about search in the buffer.

archseer:
Search is selection based though, a way to see all the results would be to use `vnnnnn` to extend the selection with all matches then `(` and `)` to cycle the primary selection

archseer:
You can also `%s` followed by search query to select all the matches.

antoyo:
That would produce a different behavior, though.

[Kakoune supports it.](https://github.com/mawww/kakoune/commit/25451ac1128feb0f1ce8c1adbf327964f6b8b086) For instance, you can create a search highlighter this way:
```
:addhl buffer/ dynregex '%reg{/}' 0:+u 1:+b
```

xJonathanLEI:
Just high-lighting all the search result is not exactly the same as selecting all of them (and hence... highlighting) though. I would like to loop through all the matches, and once I find the match I'm interested in, stop and make changes (ONLY to the current match), and then continue.

Having the next and previous matches highlighted helps me quickly navigate, as I can have a reasonable guess on where I would land before I even hit `n` (am I gonna jump to another function? or really just next line?). It saves a lot of mental effort trying to identify where I am, especially with the lack of smooth scrolling in terminal.

cemalokten:
> Just high-lighting all the search result is not exactly the same as selecting all of them (and hence... highlighting) though. I would like to loop through all the matches, and once I find the match I'm interested in, stop and make changes (ONLY to the current match), and then continue.
> 
> Having the next and previous matches highlighted helps me quickly navigate, as I can have a reasonable guess on where I would land before I even hit `n` (am I gonna jump to another function? or really just next line?). It saves a lot of mental effort trying to identify where I am, especially with the lack of smooth scrolling in terminal.

I totally agree, the ability to highlight all search results in the buffer would be incredibly useful, something I used to do in neovim all the time before switching to Helix.

eguiraud:
Maybe easier now that virtual text #411 has been merged?

PR #4222, opened 2022-10-12
Added search_selection_bounded to search with word boundaries

mitsuhiko:
This adds a new command `search_selection_bounded` which works like `search_selection` but adds `\b` word boundary markers around. This lets a user emulate the `*` behavior from vim as such:

```toml
[keys.normal]
"*" = ["move_prev_word_start", "move_next_word_end", "search_selection_bounded", "search_next"]
```

(Before I used `shrink_selection` here but that does not work as well)

I did not want to touch the behavior of `search_selection` but I think generally that bounded search is likely what users want. It's quite practical when quickly jumping between a function declaration and uses of it in the same file in many languages.

Refs https://github.com/helix-editor/helix/issues/1762 which discusses behavior of the `*` based search.

One other significant difference between how search in Vim and Helix works is that as a `*` user you are quite used to the search staying highlighted. That however is an otherwise unrelated issue that (see https://github.com/helix-editor/helix/issues/1733). Beyond this I think the behavior of `*` can be matched now though for people who want it.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -246,6 +246,7 @@ impl MappableCommand {
         extend_search_next, "Add next search match to selection",
         extend_search_prev, "Add previous search match to selection",
         search_selection, "Use current selection as search pattern",
+        search_selection_bounded, "Use current selection as search pattern with word boundaries",
         global_search, "Global search in workspace folder",
         extend_line, "Select current line, if already selected, extend to another line based on the anchor",
         extend_line_below, "Select current line, if already selected, extend to next line",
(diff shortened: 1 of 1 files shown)


mitsuhiko:
Small aside: `shrink_selection` is not ideal here. Preferrable one would be able to use what `miw` for better behavior, but that requires the current `mi`/`ma` dialog to be scriptable. See or #1723 or #4287 for PRs which expose this functionality.

the-mikedavis:
Yeah, `shrink_selection` needs there to be tree-sitter support for the language and doesn't work as you might expect within injections. I've seen `["move_prev_word_start", "move_next_word_end"]` as a workaround

mitsuhiko:
I actually tried that, but that also doesn't let you hit the selection on the first character. Personally I'm now using this in combination with the changes from #4287 and I love it:

```toml
[keys.normal]
"*" = ["select_textobject_inner_word", "search_selection_bounded", "search_next"]
```

mitsuhiko:
I have been thinking about this a bit now and an interesting other option would be to instead of adding this command, to instead have a `modify_search_add_bounds` function which reads the register and writes the same regex back, adding `\b` markers on both sides if they haven't been there yet.  The reason this could be interesting is that this would also be useful for custom searches if bound to a hot key.

That way one could search for `foobar` and if it turns out that there are just too many `foobar_X` one could hit another key to automatically resubmit the search with word boundaries.

mitsuhiko:
At this point I have a small preference of #4322 over this one here.

mitsuhiko:
Since #4322 was merged I'm going to close this PR as it is now no longer needed.

PR #4223, opened 2022-10-12
Fix some commands not showing names in palette

mitsuhiko:
This fixes the code added in #4071 to also show the names for commands without a binding. Before only commands with a binding would show the name.

image


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4238, opened 2022-10-13
Add more informational description to goto_line_end_newline

gavincrawford:
This should help distinguish `goto_line_end` and `goto_line_end_newline`.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -308,7 +308,7 @@ impl MappableCommand {
         goto_line_end, "Goto line end",
         goto_next_buffer, "Goto next buffer",
         goto_previous_buffer, "Goto previous buffer",
-        goto_line_end_newline, "Goto line end and add a new line",
+        goto_line_end_newline, "Goto newline after line end",
         goto_first_nonwhitespace, "Goto first non-blank in line",
         trim_selections, "Trim whitespace from selections",
         extend_to_line_start, "Extend to line start",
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -308,8 +308,7 @@ impl MappableCommand {
         goto_line_end, "Goto line end",
         goto_next_buffer, "Goto next buffer",
         goto_previous_buffer, "Goto previous buffer",
-        // TODO: different description ?
-        goto_line_end_newline, "Goto line end",
+        goto_line_end_newline, "Goto line end and add a new line",
```
sudormrfbin, line 311: This is not the correct behavior though -- it doesn't _add_ a newline, it simply places the cursor on the newline. `goto_line_end` places it on the last character instead.

gavincrawford: You're right, will fix.

helix-term/src/commands.rs
```diff
@@ -308,8 +308,7 @@ impl MappableCommand {
         goto_line_end, "Goto line end",
         goto_next_buffer, "Goto next buffer",
         goto_previous_buffer, "Goto previous buffer",
-        // TODO: different description ?
-        goto_line_end_newline, "Goto line end",
+        goto_line_end_newline, "Goto newline after line end",
```
the-mikedavis, line 311: I actually prefer the current description: the line end is the newline character

gavincrawford: Okay, I can see why after looking in to the command.

PR #4242, opened 2022-10-13
Add support for Purescript language

dariooddenino:
Basic highlighting based on Haskell (no treesitter grammar for Purescript yet) and LSP support

### DIFF
--- a/runtime/queries/purescript/highlights.scm
+++ b/runtime/queries/purescript/highlights.scm
@@ -1,0 +1,1 @@
+; inherits: haskell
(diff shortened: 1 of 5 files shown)


runtime/queries/purescript/highlights.scm
```diff
@@ -0,0 +1 @@
+; inherits: haskell
```
the-mikedavis, line 1: You can copy this file to `runtime/queries/purescript/injections.scm` and `runtime/queries/purescript/locals.scm` to copy over the queries from those too

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4247, opened 2022-10-13
Monokai Pro Octagon statusline modes

heliostatic:
Adding mode colors for statusline when `color-modes = true`


the-mikedavis (approved):
These look nice! 🎨

Issue #3614, opened 2022-08-31
Goto matching bracket not working in yaml


icholy:
### Summary

The Goto matching bracket not working in yaml.

### Reproduction Steps

I tried this:

1.  Place cursor on either `{` or `}` and press `mm`.

![image](https://user-images.githubusercontent.com/943597/187699050-fd3cdf69-4a05-4e98-b34c-14bab9330e7f.png)

I expected the cursor to move to the matching bracket, but it didn't move.

### Platform

Linux

### Terminal Emulator

alacritty 0.9.0

### Helix Version

helix 22.05 (83f177d2)

alevinval:
I've been looking at that issue. I'm using the following YAML document
```yaml
key: ${value}
```

I've added some debug traces, now I run `mm` on line 1, which yields:
```
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] >>>>>> FIND PAIR CALL <<<<<<
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] {Node string_scalar (0, 5) - (0, 13)}
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] pos=6
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] start_byte=5 end_byte=12
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] start_char=5 end_char=12
2022-10-13T23:28:27.687 helix_core::match_brackets [DEBUG] start=$
2022-10-13T23:28:27.688 helix_core::match_brackets [DEBUG] end=}
2022-10-13T23:28:27.688 helix_core::match_brackets [DEBUG] not valid pair...
```

It will never match because it considers the start character to be the `$` which will never match the `}`.
I also notice the `find_pair` is called up to 24 times, for that single action.

I notice that the `find_matching_bracket_fuzzy` is not running the sanity check that `find_matching_bracket` does, which first checks that the start character is a valid bracket in the first place.

alevinval:
Surely we could implement a hacky workaround like this one:

```patch
diff --git a/helix-core/src/match_brackets.rs b/helix-core/src/match_brackets.rs
index 0189dedd..a2aae514 100644
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -51,7 +51,11 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->

     loop {
[...]

PR #4259, opened 2022-10-13
match_brackets: add heuristic when matching starts with `$` char

alevinval:
When editing a variety of different types of files (eg. YAML) there's usually support for variable interpolation, this usually comes in the form of `${var}`. For the case of YAML, this is not specially recognized since it's not part of the spec. Imagine the following YAML:

```yaml
key: ${value}
```

According to the YAML spec this is simply a string scalar node, which TreeSitter correctly recognizes. Users that try going to the matching bracket may find the behaviour confusing, because nothing happens.

This patch runs a small check, in which if the syntax node starts with `$` then it skips to the next character, worst-case scenario it's not a recognized bracket, and best case scenario a supported bracket.

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -51,7 +51,11 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
 
     loop {
         let (start_byte, end_byte) = surrounding_bytes(doc, &node)?;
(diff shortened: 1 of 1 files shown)


dead10ck:
I'm not in favor of adding hardcoded character exceptions like this for the benefit of one particular markup language.

alevinval:
I totally agree with you. I'd recommend to close #3614

PR #4262, opened 2022-10-14
Update tree-sitter-cairo with support for 0.10

xJonathanLEI:
In [Cairo 0.10.0](https://github.com/starkware-libs/cairo-lang/releases/tag/v0.10.0) the comment token has been changed from the Pythonic style `#` to the more C-like `//`:

> Change comment to use `//` instead of `#`

The PR fixes the comment token used in Helix for Cairo.

There are also many other changes, including the introduction of curly brackets, but those need changes in the tree-sitter grammar, so it's probably better to deal with those in a future PR.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1413,7 +1413,7 @@ indent = { tab-width = 4, unit = "    " }
 
 [[grammar]]
 name = "cairo"
-source = { git = "https://github.com/archseer/tree-sitter-cairo", rev = "5155c6eb40db6d437f4fa41b8bcd8890a1c91716" }
+source = { git = "https://github.com/archseer/tree-sitter-cairo", rev = "e2f9b3d75f483fcfda03cf07e4176869357bfb52" }
 
 [[language]]
 name = "cpon"
--- a/languages.toml
+++ b/languages.toml
@@ -1413,7 +1413,7 @@ indent = { tab-width = 4, unit = "    " }
 
 [[grammar]]
 name = "cairo"
-source = { git = "https://github.com/archseer/tree-sitter-cairo", rev = "e2f9b3d75f483fcfda03cf07e4176869357bfb52" }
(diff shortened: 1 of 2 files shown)


archseer:
Yeah I'm the [grammar owner](https://github.com/archseer/tree-sitter-cairo/), will update according to https://github.com/starkware-libs/cairo-lang/blob/master/src/starkware/cairo/lang/compiler/cairo.ebnf

archseer:
Pushed an updated grammar into this PR, I'll leave it open for another day or two as I test this across files.

xJonathanLEI:
That was quick! I will be testing this branch on my Cairo code too :)

xJonathanLEI:
Not sure if I did something wrong, but the highlighting doesn't seem to be the same before the update:

Pre-0.10.0 highlight (with `hx` on `master`):

image

Post-0.10.0 highlight of the same snippet (but migrated to 0.10.0, of course), with `hx` on this branch:

image

Will create a minimal example to reproduce this shortly.

xJonathanLEI:
OK this code snippet reproduced just that:

```cairo
%lang starknet

namespace Space {
    func initializer{syscall_ptr: felt*, pedersen_ptr: HashBuiltin*, range_check_ptr}(
        param_1: felt, param_2: felt, param_3: felt, param_4: felt
    ) {
        Proxy.initializer(param_1);

        with_attr error_message("Zero address") {
            assert_not_zero(param_1);
        }

        if (param_2 == 0) {
            return ();
        } else {
            return ();
        }
    }
}
```

image

It started showing weird colors once I added the `if` paragraph. Am pretty sure it's the problem.

archseer:
Yeah there's a couple of problems with the grammar update still, anything highlighting red is basically an `@error`

archseer:
Should be fixed now! It was looking for a second curly bracket at the start of if statements.

xJonathanLEI:
Just tried and it works now! Will continue testing.

Issue #3967, opened 2022-09-25
data loss possible on forced kill


andreaskremser:
### Summary

when editing a huge (e.g. >80K lines) C-file, the editor sometimes gets unresponsive for extended time periods (up to several minutes). 
Killing the hanging editor has in one instance deleted the bottom half of the file, while only a few lines have actually been edited.

### Reproduction Steps

I tried this:

1. `hx huge_c_file.c` 
edit a few lines
-> editor hangs, possibly due to language server
force-kill hx

I expected this to happen:
In best case staying responsive ;) 
Losing the latest changes is OK. 

Instead, this happened:
Losing lines (about half the file), which have not been edited within the current editing session.. is unexpected.

### Platform

Linux (fedora 36)

### Terminal Emulator

konsole

### Helix Version

helix 22.08.01

dead10ck:
I actually suspected this was possible, but I wasn't sure. This confirms it for me. This is happening because every time we save, we truncate the file and write out the document text fresh:

https://github.com/helix-editor/helix/blob/e8f0886b21d279fedce53d85f53bb537c1ad808e/helix-view/src/document.rs#L562-L563

I think the solution here is to write it to a temporary file first and then move over the existing file.

4mishra:
@dead10ck Nice approach to first write to a different file instead of overwriting document directly.

dead10ck:
It's the easiest way to avoid the problem. But I actually wonder what other editors do. This would be a really inefficient way to write a very large file.

ngraham20:
I can poke at this

ngraham20:
Ok, as far as quick research goes, it seems vim handles this by using swap files that are stored either in the working directory or a specified location. This swap file contains info like undo/redo histories and the state of the file as an autosave. This allows recoveries due to crashes. Maybe helix could do something similar?

https://www.cs.longwood.edu/VimSwap.html#:~:text=The%20reason%20this%20message%20comes,your%20undo%2Fredo%20history).

the-mikedavis:
#401 might be the fix for this along those lines

ngraham20:
Should we still add this temporary workaround, as that looks like a bigger code change?

timokoesters:
I just had this problem and it wiped out the entire file. I was using rust formatter and the rust lsp and I think I triggered it by hitting ctrl-z followed by ctrl-c (because it seemed to hang).

A fix or temporary workaround would be great. Fortunately I didn't lose much data.

PR #4276, opened 2022-10-14
Workaround for Large File Corruption During Save

ngraham20:
PR as a workaround for #3967

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -13,6 +13,7 @@ use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
+use tokio::io::AsyncReadExt;
 
 use helix_core::{
     encoding,
@@ -562,14 +563,53 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
+
+            // running 5 tests
+            // test test::write::test_write_concurrent ... ignored
+            // test test::write::test_write_fail_mod_flag ... ignored
+            // test test::write::test_write_fail_new_path ... ignored
+            // 2022-10-15T23:53:23.412 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
+            // 2022-10-15T23:53:23.412 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
+            // 2022-10-15T23:53:23.673 helix_view::document [INFO] logs are getting printed
+            // 2022-10-15T23:53:23.673 helix_view::document [INFO] logs are still getting printed
+            // 2022-10-15T23:53:23.674 helix_view::document [INFO] logs are still yet getting printed
+            // 2022-10-15T23:53:23.674 helix_view::document [INFO] and yet they still be printed
+            // 2022-10-15T23:53:23.674 helix_view::document [INFO] I can't belive i'm doing this
+            // 2022-10-15T23:53:23.680 helix_view::document [INFO] logs are getting printed
+            // 2022-10-15T23:53:23.680 helix_view::document [INFO] logs are still getting printed
+            // 2022-10-15T23:53:23.680 helix_view::document [INFO] logs are still yet getting printed
+            // 2022-10-15T23:53:23.680 helix_view::document [INFO] and yet they still be printed
+            // 2022-10-15T23:53:23.680 helix_view::document [INFO] I can't belive i'm doing this
+            // test test::write::test_write_quit ... FAILED
+            // test test::write::test_write ... FAILED
             let mut tmp_filename = std::ffi::OsString::from("~");
             tmp_filename.push(path.file_name().unwrap());
 
             let tmp_path = path.parent().unwrap().join(tmp_filename);
 
             let mut tmp_file = File::create(&tmp_path).await?;
+            log::info!("logs are getting printed");
+
             to_writer(&mut tmp_file, encoding, &text).await?;
+            log::info!("logs are still getting printed");
+
+            let mut tmp_file_content = String::new();
+            log::info!("logs are still yet getting printed");
+
+            let mut actual_file_content = String::new();
+            log::info!("and yet they still be printed");
+
+            let mut actual_file = File::create(&path).await?;
+            log::info!("I can't belive i'm doing this");
+
+            tmp_file.read_to_string(&mut tmp_file_content).await?;
+            log::info!("tmp file contents: {}", tmp_file_content);
+
             tokio::fs::rename(tmp_path, path).await?;
+            log::info!("I renamed the file");
+
+            actual_file.read_to_string(&mut actual_file_content).await?;
+            log::info!("actual file contents after rename: {}", actual_file_content);
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -589,7 +589,8 @@ impl Document {
 
             let mut tmp_file = File::create(&tmp_path).await?;
             log::info!("logs are getting printed");
-
+            log::info!("temp file location: {:?}", tmp_path);
+            log::info!("actual file location: {:?}", path);
             to_writer(&mut tmp_file, encoding, &text).await?;
             log::info!("logs are still getting printed");
 
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -563,8 +563,7 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
-            let tmp_path =
-                cache_dir().join(format!("~{}", path.file_name().unwrap().to_str().unwrap()));
+            let tmp_path = cache_dir().join(format!("~{:?}", path.file_name().unwrap()));
 
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -3,7 +3,6 @@ use futures_util::future::BoxFuture;
 use futures_util::FutureExt;
 use helix_core::auto_pairs::AutoPairs;
 use helix_core::Range;
-use helix_loader::cache_dir;
 use serde::de::{self, Deserialize, Deserializer};
 use serde::Serialize;
 use std::borrow::Cow;
@@ -563,28 +562,14 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
-            let tmp_path = cache_dir().join(format!("~{:?}", path.file_name().unwrap()));
+            let mut tmp_filename = std::ffi::OsString::from("~");
+            tmp_filename.push(path.file_name().unwrap().to_os_string());
 
-            if let Some(parent) = tmp_path.parent() {
-                // TODO Temporary fix. Normally, this directory will exist by this point,
-                // but not during the integration tests.
-                if !parent.exists() {
-                    if force {
-                        std::fs::DirBuilder::new().recursive(true).create(parent)?;
-                    } else {
-                        bail!("can't save file, parent directory does not exist");
-                    }
-                }
-            }
+            let tmp_path = path.parent().unwrap().join(tmp_filename);
 
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
-
-            tokio::fs::copy(&tmp_path, path).await?;
-            tokio::fs::remove_file(tmp_path).await?;
-
-            // TODO figure out why rename makes the tests break
-            //tokio::fs::rename(tmp_path, path).await?;
+            tokio::fs::rename(tmp_path, path).await?;
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -562,13 +562,26 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
-            let mut tmp_path = path.clone();
-            tmp_path.set_file_name(format!(".~{}", path.file_name().unwrap().to_str().unwrap()));
+            let tmp_path = PathBuf::from(std::env::var("HOME").unwrap())
+                .join(".cache/")
+                .join("helix/")
+                .join(format!("~{}", path.file_name().unwrap().to_str().unwrap()));
+
+            if let Some(parent) = tmp_path.parent() {
+                // TODO: display a prompt asking the user if the directories should be created
+                if !parent.exists() {
+                    if force {
+                        std::fs::DirBuilder::new().recursive(true).create(parent)?;
+                    } else {
+                        bail!("can't save file, parent directory does not exist");
+                    }
+                }
+            }
+
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
 
-            tokio::fs::copy(&tmp_path, path).await?;
-            tokio::fs::remove_file(tmp_path).await?;
+            tokio::fs::rename(&tmp_path, path).await?;
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -3,6 +3,7 @@ use futures_util::future::BoxFuture;
 use futures_util::FutureExt;
 use helix_core::auto_pairs::AutoPairs;
 use helix_core::Range;
+use helix_loader::cache_dir;
 use serde::de::{self, Deserialize, Deserializer};
 use serde::Serialize;
 use std::borrow::Cow;
@@ -562,21 +563,8 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
-            let tmp_path = PathBuf::from(std::env::var("HOME").unwrap())
-                .join(".cache/")
-                .join("helix/")
-                .join(format!("~{}", path.file_name().unwrap().to_str().unwrap()));
-
-            if let Some(parent) = tmp_path.parent() {
-                // TODO: display a prompt asking the user if the directories should be created
-                if !parent.exists() {
-                    if force {
-                        std::fs::DirBuilder::new().recursive(true).create(parent)?;
-                    } else {
-                        bail!("can't save file, parent directory does not exist");
-                    }
-                }
-            }
+            let tmp_path =
+                cache_dir().join(format!("~{}", path.file_name().unwrap().to_str().unwrap()));
 
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -565,11 +565,26 @@ impl Document {
             // that occurs when crashing during a file save
             let tmp_path = cache_dir().join(format!("~{:?}", path.file_name().unwrap()));
 
+            if let Some(parent) = tmp_path.parent() {
+                // TODO Temporary fix. Normally, this directory will exist by this point,
+                // but not during the integration tests.
+                if !parent.exists() {
+                    if force {
+                        std::fs::DirBuilder::new().recursive(true).create(parent)?;
+                    } else {
+                        bail!("can't save file, parent directory does not exist");
+                    }
+                }
+            }
+
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
 
             tokio::fs::copy(&tmp_path, path).await?;
             tokio::fs::remove_file(tmp_path).await?;
+
+            // TODO figure out why rename makes the tests break
+            //tokio::fs::rename(tmp_path, path).await?;
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -560,8 +560,15 @@ impl Document {
                 }
             }
 
-            let mut file = File::create(path).await?;
-            to_writer(&mut file, encoding, &text).await?;
+            // TODO Temporary file creation is a workaround to solve large file corruption
+            // that occurs when crashing during a file save
+            let mut tmp_path = path.clone();
+            tmp_path.set_file_name(format!(".~{}", path.file_name().unwrap().to_str().unwrap()));
+            let mut tmp_file = File::create(&tmp_path).await?;
+            to_writer(&mut tmp_file, encoding, &text).await?;
+
+            tokio::fs::copy(&tmp_path, path).await?;
+            tokio::fs::remove_file(tmp_path).await?;
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -563,7 +563,7 @@ impl Document {
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
             let mut tmp_filename = std::ffi::OsString::from("~");
-            tmp_filename.push(path.file_name().unwrap().to_os_string());
+            tmp_filename.push(path.file_name().unwrap());
 
             let tmp_path = path.parent().unwrap().join(tmp_filename);
 
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -568,7 +568,8 @@ impl Document {
             let mut tmp_file = File::create(&tmp_path).await?;
             to_writer(&mut tmp_file, encoding, &text).await?;
 
-            tokio::fs::rename(&tmp_path, path).await?;
+            tokio::fs::copy(&tmp_path, path).await?;
+            tokio::fs::remove_file(tmp_path).await?;
 
             if let Some(language_server) = language_server {
                 if !language_server.is_initialized() {
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -40,7 +40,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       RUST_BACKTRACE: 1
-      HELIX_LOG_LEVEL: info
+      HELIX_LOG_LEVEL: debug
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3345,9 +3345,15 @@ fn yank_joined_to_clipboard_impl(
         .map(Cow::into_owned)
         .collect();
 
+    let clipboard_text = match clipboard_type {
+        ClipboardType::Clipboard => "system clipboard",
+        ClipboardType::Selection => "primary clipboard",
+    };
+
     let msg = format!(
-        "joined and yanked {} selection(s) to system clipboard",
+        "joined and yanked {} selection(s) to {}",
         values.len(),
+        clipboard_text,
     );
 
     let joined = values.join(separator);
@@ -3376,6 +3382,11 @@ fn yank_main_selection_to_clipboard_impl(
     let (view, doc) = current!(editor);
     let text = doc.text().slice(..);
 
+    let message_text = match clipboard_type {
+        ClipboardType::Clipboard => "yanked main selection to system clipboard",
+        ClipboardType::Selection => "yanked main selection to primary clipboard",
+    };
+
     let value = doc.selection(view.id).primary().fragment(text);
 
     if let Err(e) = editor
@@ -3385,7 +3396,7 @@ fn yank_main_selection_to_clipboard_impl(
         bail!("Couldn't set system clipboard content: {}", e);
     }
 
-    editor.set_status("yanked main selection to system clipboard");
+    editor.set_status(message_text);
     Ok(())
 }
 
--- a/temp.txt
+++ b/temp.txt
@@ -1,0 +1,1 @@
+asdffdsa
--- a/temp.txt
+++ b/temp.txt
@@ -1,1 +1,0 @@
-asdffdsa


ngraham20:
I could also imagine
```rust
tokio::fs::rename(&tmp_path, path).await?;
```
being used instead of a copy and remove

https://docs.rs/tokio/latest/tokio/fs/fn.rename.html

kirawi:
Using `rename` is probably better since `copy` will lead to the same issue. I think it would also be better to create the temp file in the cache directory.

ngraham20:
I went the route of making the cached file match the filename instead of having a file called "cachefile" or something just in case multiple saves and crashes occur at the same time

helix-view/src/document.rs
```diff
@@ -562,13 +562,26 @@ impl Document {
 
             // TODO Temporary file creation is a workaround to solve large file corruption
             // that occurs when crashing during a file save
-            let mut tmp_path = path.clone();
-            tmp_path.set_file_name(format!(".~{}", path.file_name().unwrap().to_str().unwrap()));
+            let tmp_path = PathBuf::from(std::env::var("HOME").unwrap())
```
kirawi, line 565: You can use this function instead: https://github.com/helix-editor/helix/blob/a3ed9169df8d0a56ab4f698d2c90e7e6b4404633/helix-loader/src/lib.rs#L70-L76

kirawi: It's already created on startup as it's the folder where the log is written.

helix-view/src/document.rs
```diff
@@ -560,8 +560,28 @@ impl Document {
...
             }
 
-            let mut file = File::create(path).await?;
-            to_writer(&mut file, encoding, &text).await?;
+            // TODO Temporary file creation is a workaround to solve large file corruption
+            // that occurs when crashing during a file save
+            let tmp_path = PathBuf::from(std::env::var("HOME").unwrap())
+                .join(".cache/")
+                .join("helix/")
+                .join(format!("~{}", path.file_name().unwrap().to_str().unwrap()));
```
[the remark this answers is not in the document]
ngraham20: sounds good to me, switched the format to use {:?} and removed to_str()

helix-view/src/document.rs
ngraham20: oooh, love it. Updated to that and removed the bit where it tests for the directory since it's already there

ngraham20:
I don't get why the builds are failing. The file saves to exactly where it's supposed to. Could it potentially be an async thing, where it's checking too quickly and the file hasn't been copied to the new location yet?

kirawi:
I'm not sure. The async operation should be completed successfully before it's checked.

kirawi:
Since it's writing to the cache, maybe the CI doesn't allow that?

ngraham20:
> Since it's writing to the cache folder, maybe the CI doesn't allow that?

Maybe? But I would assume it isn't looking there, it should be looking for the file that got created in the working directory? In which case, it saves to the temp file, then moves it to the correct location, then the assert should look there, yeah?

ngraham20:
oh you mean that it could be failing to write to the cache folder, therefore it never gets moved into the correct spot afterwards... That makes sense to me

kirawi:
If it wasn't able to write to the cache directory, then the write operation would fail. I'm assuming that's why the file is empty.

ngraham20:
> If it wasn't able to write to the cache directory, then the write operation would fail. I'm assuming that's why the file is empty.

wait, no, because I ran the integration tests locally just now and they still fail, so it can't be that

kirawi:
Could you test with reverting the `rename()` change?

ngraham20:
> Could you test with reverting the `rename()` change?

yeah that made it pass

ngraham20:
I'm going to upload this version for now. Maybe dead10ck has some idea of what's going on, I sure don't :sweat_smile:

dead10ck:
>This will not work if the new name is on a different mount point.

Interesting. That's probably what it is. The test is making a temporary file in `/tmp`, which is often on a different mount point than `/home`. That's an odd and unfortunate limitation. Copying doubles the amount of IO for each file save.

ngraham20:
hmmmm, the integration test is still failing. Works when I run them locally though, even with the copy/delete

ngraham20:
need to sign off for the night. I'll check back throughout the weekend though and keep poking around with you guys :+1:

dead10ck:
Oh I think I know what's going on: the cache directory isn't being created. That happens in the main function

https://github.com/helix-editor/helix/blob/a3ed9169df8d0a56ab4f698d2c90e7e6b4404633/helix-term/src/main.rs#L41-L47

but the integration tests don't run that code. They build an `Application` and inject right into the main event loop. So we'll either have to have the integration testing harness or the write path create the directories.

ngraham20:
Oooh, that would track for sure, since the change I added replace was the same time we moved it to the cache directory

kirawi:
As for the `rename()` debacle, I wonder if we could use that by default, and if it fails try to do the copy approach.

ngraham20:
> As for the `rename()` IO issue, I wonder if we could use that by default, and if it fails try to do the copy approach.

That would need to be directly in the integration test though, right? Since it actually does work the way it's supposed to in practice?

kirawi:
If it's failing because `tmp` is on a different mount point, then I think it would be better to have it be in `save_impl`.

helix-view/src/document.rs
```diff
@@ -560,8 +561,30 @@ impl Document {
...
-            to_writer(&mut file, encoding, &text).await?;
+            // TODO Temporary file creation is a workaround to solve large file corruption
+            // that occurs when crashing during a file save
+            let tmp_path = cache_dir().join(format!("~{:?}", path.file_name().unwrap()));
+
+            if let Some(parent) = tmp_path.parent() {
+                // TODO Temporary fix. Normally, this directory will exist by this point,
+                // but not during the integration tests.
+                if !parent.exists() {
+                    if force {
```
kirawi, line 572: I don't think the write test does `:w!` (force write).

ngraham20:
Build failed again. It's strange, because it succeeds on my machine. It's got to have something to do with the cache directory location, because when I was caching the file in the working directory, all the builds succeeded in the CI environment

helix-view/src/document.rs
kirawi: https://github.com/helix-editor/helix/blob/a3ed9169df8d0a56ab4f698d2c90e7e6b4404633/helix-term/tests/test/write.rs#L18
So it still doesn't build the directory.

ngraham20: Oh, that makes sense, hence it doesn't change how the CI works

ngraham20:
Is there maybe a way to test if the cache directory is on a different mount point than the working directory? It occurs to me that this would need to be addressed regardless

ngraham20:
That way we only do the copy in the scenario where rename wouldn't work

helix-view/src/document.rs
dead10ck: Yeah I think it's fine to create the cache directory in `save_impl` if it doesn't exist.

dead10ck:
The problem with copying is it doesn't actually solve the problem this PR is trying to address, i.e. it could fail in the middle of copying over the target file. Renaming is usually atomic, so you can be sure that if the file wrote to a temporary file successfully that it will either replace the target file whole and complete, or it will fail to rename for some reason and leave the original untouched.

It seems like we have to use rename to get what we want, but in order to do that, we have to make sure the temporary file is on the same mount point. We might be able to find a way to check if it's on the same mount point, but probably not in a trivially cross platform way. The simplest solution might be to just not use the cache directory after all. Writing the temporary file to the same directory as the target file is a really easy way to make sure the mount point is the same.

ngraham20:
How's this, I'll revert back to the working directory version of the cache file, and we could open a harder issue for finding a clever way to get it in the cache?

dead10ck:
The working directory might not be where you're writing the file though, the path of the file could be on a different mount point. I think we have to parse out the parent directory of the document path we're trying to write to.

ngraham20:
Oh, yeah. Sorry, I meant saving the temp file wherever the original file is located

archseer:
Can we verify how this is done in other editors? I imagine this is very space inefficient for large files during the save process

ngraham20:
Far as I can tell from a quick search, vim uses swap files with metadata about the session, like undo/redo and what changed were made. It autosaves every so often to that file for recovery during a crash.

Not sure about how the saving process goes to actually prevent the issue in the first place though, but we could possibly look into changing lines in-place or something?

ngraham20:
Gotta say, working with OsStr is kind of a pain XD I had similar pains before when dealing with unicode

ngraham20:
this is the code that built properly. I don't get why the new code doesn't
```rust
let mut tmp_path = path.clone();
tmp_path.set_file_name(format!(".~{}", path.file_name().unwrap().to_str().unwrap()));
let mut tmp_file = File::create(&tmp_path).await?;
to_writer(&mut tmp_file, encoding, &text).await?;

tokio::fs::copy(&tmp_path, path).await?;
tokio::fs::remove_file(tmp_path).await?;
```

ngraham20:
ok, I just did two integration tests locally. Once with copy/delete, once with rename, this time with the tempfile in the *same* directory as the edited file. The tests do not pass. The only difference is copy/delete and rename

ngraham20:
I don't know enough about how async or assert work to know why this is happening, but the test fails before the rename can happen, which is why it's failing.

ngraham20:
I have a guess though. The test function uses `flush()` and `sync_all()` which as I understand it will halt the test until the write buffer is clear. The moment it is, it reads the file contents.

If I'm right, this is problematic, since the file was written to a location the test doesn't know about, meaning the *real* file is still empty when it tries to read from it.

dead10ck:
@ngraham20 it looks like the file is actually flushed during the write process itself:

https://github.com/helix-editor/helix/blob/b07ef6bec30e1c704a63579efa5cc433c22fcc19/helix-view/src/document.rs#L314-L320

So I think actually the flushing that happens in the integration tests might be superfluous.

ngraham20:
Maybe, but how else do you explain the fact that the function doesn't even complete? I want to say we're hitting race condition, that the test reads the file before the rewrite happens, therefore the assert fails, the test terminates, and the file is never written at all

ngraham20:
oh plus, when `to_writer()` does the flush, it does it on the `tmp_file`, but when the test does the flush, it does it on the actual file (which doesn't exist yet)

kirawi:
It might help to create a minimal reproducible case of what's going on here to debug the issue.

dead10ck:
> oh plus, when `to_writer()` does the flush, it does it on the `tmp_file`, but when the test does the flush, it does it on the actual file (which doesn't exist yet)

After a bit of research, it turns out that when you rename a file, it is necessary to fsync the directory!

https://stackoverflow.com/questions/3764822/how-to-durably-rename-a-file-in-posix

Today I learned.

ngraham20:
I'm most sure what that means tbh 😅 what does that mean for what's actually going on?

dead10ck:
I'm not positive this is the problem, but it's something to try: after the rename, open a `File` with the directory the target file is in and call `sync_all` on it.

ngraham20:
Nope, that didn't do it. I did figure out that anything that occurs in the save function *on or after* it tries to `write_to_string` to check the values causes the rest of the function to terminate. We're way beyond what I understand at this point.

I think we may need to go back though and look at specifically what the difference is here.

this code does not fail
```rust
let mut tmp_filename = std::ffi::OsString::from("~");
tmp_filename.push(path.file_name().unwrap());
let tmp_path = path.parent().unwrap().join(tmp_filename);
let mut tmp_file = File::create(&tmp_path).await?;
to_writer(&mut tmp_file, encoding, &text).await?;

tokio::fs::copy(&tmp_path, path).await?;
tokio::fs::remove_file(tmp_path).await?;
```

this code does fail
```rust
let mut tmp_filename = std::ffi::OsString::from("~");
tmp_filename.push(path.file_name().unwrap());
let tmp_path = path.parent().unwrap().join(tmp_filename);
let mut tmp_file = File::create(&tmp_path).await?;
to_writer(&mut tmp_file, encoding, &text).await?;

tokio::fs::rename(tmp_path, path).await?;
```

the *only* difference is rename vs copy/remove. That's it. I personally think we're dealing with some kind of race condition or some kind of IO r/w issue is causing the read_string to fail. I've only got guesses though, and anything I can think of is to throw the noodle at the wall to see if it sticks.

kirawi:
You could try asking the tokio devs. If it's a race condition, it would be on their end.

ngraham20:
> On a typical Unix filesystem, a file may have multiple filenames at once, or even none at all. The file metadata is stored in an inode, which has a unique inode number, and this inode number can be linked from any number of directory entries. However, there are no reverse links from the inode back to the directory entries.

> Given an open File object in Rust, you can get the inode number using the [ino() method](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.ino). If you know the directory the log file is in, you can use [std::fs::read_dir()](https://doc.rust-lang.org/std/fs/fn.read_dir.html) to iterate over all entries in that directory, and each entry will also have an [ino() method](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.ino), so you can find the one(s) matching your open file object. Of course this approach is subject to race conditions – the directory entry may already be gone again once you try to do anything with it.

I wonder if this is why renaming doesn't work, since the filename changes, but maybe the inode didn't change

ngraham20:
OOOOOOOH I think I'm onto something here

### before the rename
tmpfile ino     : 3224386
actual file ino : 1714009

### after the rename, I would expect the actual file ino to be 1714009
actual file ino : 3224404

Not only is the ino not the same as the temporary file, *it's not even the same as the original file*. Therefore, I propose that what's going on is that the test tries to read the contents of 1714009, which has nothing in it.

ngraham20:
Wait maybe that was just overlap from having two async tests run at once... When I do one at a time, the rename does actually update the ino of the test's file.

dead10ck:
> OOOOOOOH I think I'm onto something here
> ### before the rename
> 
> tmpfile ino : 3224386 actual file ino : 1714009
> ### after the rename, I would expect the actual file ino to be 1714009
> 
> actual file ino : 3224404
> 
> Not only is the ino not the same as the temporary file, _it's not even the same as the original file_. Therefore, I propose that what's going on is that the test tries to read the contents of 1714009, which has nothing in it.

As far as I understand how inodes work, this actually looks correct. My understanding is that an inode is the unique id of a file *on the disk*, so e.g. when you make a hard link of a file, the inode is the same as the original because it's actually pointing to the same file on disk.

When you write to the temporary file, you're creating a brand new file on the disk, so it gets a new inode. When you rename the new file over the old file, the old file gets the new file's inode.

ngraham20:
Yeah I came to the same conclusion after more testing. Bummer haha

ngraham20:
Makes it even more confusing though why it would fail to read the contents, as clearly its getting renamed properly and absolutely has the correct string in it.

ngraham20:
*something* fundamentally has to be different between copy and rename.

kirawi:
I wrote this unit test:
```rs
#[tokio::test]
    async fn save() {
        use std::io::Read;
        let mut file = tempfile::NamedTempFile::new().unwrap();
        let mut doc = Document::open(file.path(), None, None).unwrap();
        let view = ViewId::default();
        doc.set_selection(view, Selection::single(0, 0));
        let transaction = Transaction::change(doc.text(), [(0, 0, Some("ithe gostak distims the doshes".into()))].into_iter());
        doc.apply(&transaction, view);
        doc.save(false).await.unwrap();

        let mut file_content = String::new();
        file.as_file_mut().read_to_string(&mut file_content).unwrap();
        assert_eq!(file_content, "ithe gostak distims the doshes");
    }
```
And it fails with `thread 'document::test::save' panicked at 'called Result::unwrap() on an Err value: Access is denied. (os error 5)`  originating from `tokio::fs::rename(tmp_path, path).await?;`

ngraham20:
Maybe one of the earlier processes is keeping the file open? Permission denied could occur if another of the async functions still has it open for r/w right?

kirawi:
I found this issue: https://github.com/Stebalien/tempfile/issues/131 and it might be Windows-specific. Could you see what the panic is on a unix machine?

ngraham20:
I get this

```bash
---- test::write::save stdout ----
thread 'test::write::save' panicked at 'called `Result::unwrap()` on an `Err` value: Bad file descriptor (os error 9)', helix-term/tests/test/write.rs:29:27
```

```rust
#[tokio::test]
async fn save() {
    use helix_view::document::Document;
    use helix_core::Transaction;
    use helix_view::ViewId;
    use std::io::Read;
    let file = tempfile::NamedTempFile::new().unwrap();
    // Need to close the temp file.
    let path = file.into_temp_path();
    let mut doc = Document::open(&path, None, None).unwrap();
    let view = ViewId::default();
    doc.set_selection(view, Selection::single(0, 0));
    let transaction = Transaction::insert(
        doc.text(),
        doc.selection(view),
        "ithe gostak distims the doshes".into(),
    );
    doc.apply(&transaction, view);
    doc.save(false).await.unwrap();

    let file_content = std::fs::read_to_string(path).unwrap();
    assert_eq!(file_content, "ithe gostak distims the doshes");
}
```

ngraham20:
same error

kirawi:
Could you see where it originates from in the backtrace?

ngraham20:
Here's the basic stacktrace

```bash
---- test::write::save stdout ----
thread 'test::write::save' panicked at 'called `Result::unwrap()` on an `Err` value: Bad file descriptor (os error 9)', helix-term/tests/test/write.rs:28:27
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1078:23
   4: integration::test::write::save::{{closure}}
             at ./tests/test/write.rs:28:5
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
   6: <core::pin::Pin<P> as core::future::future::Future>::poll
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/future.rs:124:9
   7: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}}
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:525:48
   8: tokio::coop::with_budget::{{closure}}
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:102:9
   9: std::thread::local::LocalKey<T>::try_with
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/thread/local.rs:442:16
  10: std::thread::local::LocalKey<T>::with
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/thread/local.rs:418:9
  11: tokio::coop::with_budget
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:95:5
  12: tokio::coop::budget
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:72:5
  13: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:525:25
  14: tokio::runtime::scheduler::current_thread::Context::enter
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:349:19
  15: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:524:36
  16: tokio::runtime::scheduler::current_thread::CoreGuard::enter::{{closure}}
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:595:57
  17: tokio::macros::scoped_tls::ScopedKey<T>::set
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/macros/scoped_tls.rs:61:9
  18: tokio::runtime::scheduler::current_thread::CoreGuard::enter
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:595:27
  19: tokio::runtime::scheduler::current_thread::CoreGuard::block_on
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:515:19
  20: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:161:24
  21: tokio::runtime::Runtime::block_on
             at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/mod.rs:490:46
  22: integration::test::write::save
             at ./tests/test/write.rs:31:5
  23: integration::test::write::save::{{closure}}
             at ./tests/test/write.rs:12:7
  24: core::ops::function::FnOnce::call_once
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
  25: core::ops::function::FnOnce::call_once
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
```

kirawi:
Could you try putting the test in `document.rs`? I don't think it's capturing the function calls there.

ngraham20:
wait hangon

kirawi:
`cargo test --workspace save`

ngraham20:
> `cargo test --workspace save`

and I put the test in the test mod in document.rs?

ngraham20:
```rust
#[cfg(test)]
mod test {
    use super::*;

    #[tokio::test]
    async fn save() {
        use std::io::Read;
        let file = tempfile::NamedTempFile::new().unwrap();
        let (_, path) = file.keep().unwrap();
        let mut doc = Document::open(&path, None, None).unwrap();
        let view = ViewId::default();
        doc.set_selection(view, Selection::single(0, 0));
        let transaction = Transaction::insert(
            doc.text(),
            doc.selection(view),
            "ithe gostak distims the doshes".into(),
        );
        doc.apply(&transaction, view);
        doc.save(false).await.unwrap();

        let file_content = std::fs::read_to_string(path).unwrap();
        assert_eq!(file_content, "ithe gostak distims the doshes");
    }
```

```bash
error[E0433]: failed to resolve: use of undeclared crate or module `tempfile`
    --> helix-view/src/document.rs:1208:20
     |
1208 |         let file = tempfile::NamedTempFile::new().unwrap();
     |                    ^^^^^^^^ use of undeclared crate or module `tempfile`

warning: unused import: `std::io::Read`
    --> helix-view/src/document.rs:1207:13
     |
1207 |         use std::io::Read;
     |             ^^^^^^^^^^^^^
     |
     = note: `#[warn(unused_imports)]` on by default

error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
    --> helix-view/src/document.rs:1209:17
     |
1209 |         let (_, path) = file.keep().unwrap();
     |                 ^^^^ doesn't have a size known at compile-time
     |
     = help: within `std::path::Path`, the trait `Sized` is not implemented for `[u8]`
     = note: required because it appears within the type `std::path::Path`
     = note: all local variables must have a statically known size
     = help: unsized locals are gated as an unstable feature

error[E0277]: the size for values of type `[u8]` cannot be known at compilation time
    --> helix-view/src/document.rs:1221:52
     |
1221 |         let file_content = std::fs::read_to_string(path).unwrap();
     |                            ----------------------- ^^^^ doesn't have a size known at compile-time
     |                            |
     |                            required by a bound introduced by this call
     |
     = help: within `std::path::Path`, the trait `Sized` is not implemented for `[u8]`
     = note: required because it appears within the type `std::path::Path`
note: required by a bound in `std::fs::read_to_string`
    --> /home/ngraham/.rustup/toolchains/1.61.0-x86_64-unknown-linux-gnu/lib/rustlib/src/rust/library/std/src/fs.rs:267:23
     |
267  | pub fn read_to_string<P: AsRef<Path>>(path: P) -> io::Result<String> {
     |                       ^ required by this bound in `std::fs::read_to_string`

Some errors have detailed explanations: E0277, E0433.
For more information about an error, try `rustc --explain E0277`.
warning: `helix-view` (lib test) generated 1 warning
error: could not compile `helix-view` due to 3 previous errors; 1 warning emitted
warning: build failed, waiting for other jobs to finish...
```

kirawi:
You also have to add `tempfile` to the `helix-view` dependencies.

alois31:
> I wrote this unit test:
> 
> ```rust
> #[tokio::test]
>     async fn save() {
>         use std::io::Read;
>         let mut file = tempfile::NamedTempFile::new().unwrap();
>         let mut doc = Document::open(file.path(), None, None).unwrap();
>         let view = ViewId::default();
>         doc.set_selection(view, Selection::single(0, 0));
>         let transaction = Transaction::change(doc.text(), [(0, 0, Some("ithe gostak distims the doshes".into()))].into_iter());
>         doc.apply(&transaction, view);
>         doc.save(false).await.unwrap();
> 
>         let mut file_content = String::new();
>         file.as_file_mut().read_to_string(&mut file_content).unwrap();
>         assert_eq!(file_content, "ithe gostak distims the doshes");
>     }
> ```
> 
> And it fails with `thread 'document::test::save' panicked at 'called Result::unwrap() on an Err value: Access is denied. (os error 5)` originating from `tokio::fs::rename(tmp_path, path).await?;`

This is on Windows, right? The unit test is keeping the original file open, and on Windows, IIRC deleting or overwriting an open file is not allowed.

alois31:
> Here's the basic stacktrace
> 
> ```shell
> ---- test::write::save stdout ----
> thread 'test::write::save' panicked at 'called `Result::unwrap()` on an `Err` value: Bad file descriptor (os error 9)', helix-term/tests/test/write.rs:28:27
> stack backtrace:
>    0: rust_begin_unwind
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
>    2: core::result::unwrap_failed
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
>    3: core::result::Result<T,E>::unwrap
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1078:23
>    4: integration::test::write::save::{{closure}}
>              at ./tests/test/write.rs:28:5
>    5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/mod.rs:91:19
>    6: <core::pin::Pin<P> as core::future::future::Future>::poll
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/future/future.rs:124:9
>    7: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}::{{closure}}
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:525:48
>    8: tokio::coop::with_budget::{{closure}}
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:102:9
>    9: std::thread::local::LocalKey<T>::try_with
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/thread/local.rs:442:16
>   10: std::thread::local::LocalKey<T>::with
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/thread/local.rs:418:9
>   11: tokio::coop::with_budget
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:95:5
>   12: tokio::coop::budget
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/coop.rs:72:5
>   13: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}::{{closure}}
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:525:25
>   14: tokio::runtime::scheduler::current_thread::Context::enter
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:349:19
>   15: tokio::runtime::scheduler::current_thread::CoreGuard::block_on::{{closure}}
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:524:36
>   16: tokio::runtime::scheduler::current_thread::CoreGuard::enter::{{closure}}
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:595:57
>   17: tokio::macros::scoped_tls::ScopedKey<T>::set
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/macros/scoped_tls.rs:61:9
>   18: tokio::runtime::scheduler::current_thread::CoreGuard::enter
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:595:27
>   19: tokio::runtime::scheduler::current_thread::CoreGuard::block_on
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:515:19
>   20: tokio::runtime::scheduler::current_thread::CurrentThread::block_on
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/scheduler/current_thread.rs:161:24
>   21: tokio::runtime::Runtime::block_on
>              at /home/ngraham/.cargo/registry/src/github.com-1ecc6299db9ec823/tokio-1.21.2/src/runtime/mod.rs:490:46
>   22: integration::test::write::save
>              at ./tests/test/write.rs:31:5
>   23: integration::test::write::save::{{closure}}
>              at ./tests/test/write.rs:12:7
>   24: core::ops::function::FnOnce::call_once
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
>   25: core::ops::function::FnOnce::call_once
>              at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
> note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
> ```

The debugging code attempts to read from a write-only file descriptor.

kirawi:
I think I figured out the issue since I dealt with the same thing for persistent undo... turns out you need to seek to the beginning of the file after any read or write operation otherwise you will read a blank string.

kirawi:
Superseded by https://github.com/helix-editor/helix/pull/9236

PR #4278, opened 2022-10-14
c/c++ highlights: *& are @type or @operator based on context

sam-mccall:
C gave us the infamous declaration-follows-use syntax for pointers, which can make it hard to tell at a glance whether a `*` is dereferencing a pointer or declaring one.
C++ copied the syntax for references, despite the logic not applying.

This patch highlights */& as (builtin) types when they're used to declare pointers/refs, and continues to treat them as operators in other contexts.

This is more formally correct, but also practically useful (at least for me), particularly with pointer-binds-right code styles. The potential downside is reduced contrast within type declarations.

Before: 
image
After:
image

### DIFF
--- a/runtime/queries/cpp/highlights.scm
+++ b/runtime/queries/cpp/highlights.scm
@@ -34,6 +34,10 @@
 (auto) @type
 "decltype" @type
 
+(ref_qualifier (["&" "&&"]) @type.builtin)
+(reference_declarator (["&" "&&"]) @type.builtin)
+(abstract_reference_declarator (["&" "&&"]) @type.builtin)
+
 ; Constants
 
 (this) @variable.builtin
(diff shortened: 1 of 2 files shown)


runtime/queries/cpp/highlights.scm
```diff
@@ -34,6 +34,10 @@
 (auto) @type
 "decltype" @type
 
+(ref_qualifier (["&" "&&"]) @type.builtin)
+(reference_declarator (["&" "&&"]) @type.builtin)
+(abstract_reference_declarator (["&" "&&"]) @type.builtin)
```
the-mikedavis, line 39: Very minor nit but the parens are not needed here around the `[..]`:

```scm
(ref_qualifier ["&" "&&"] @type.builtin)
```

is equivalent

sam-mccall: Thanks! I'm definitely having trouble internalizing exactly when I need them...

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4286, opened 2022-10-15
fix `:insert-output` doc: inserting output **before** each selection

zjp-CN:


zjp-CN:
`book/src/generated/typable-cmd.md` is automatically generated or manually updated?

the-mikedavis (approved):

the-mikedavis:
Good catch, thanks!

the-mikedavis:
Yep, `book/src/generated/*` is generated by running `cargo xtask docgen`

Issue #4013, opened 2022-09-28
Allow chaining `find_next_char`, `select_textobject_around`, etc. together with other commands in keymaps


tmke8:
There are quite a few commands that expect an argument in the form of a single character:
- `find_till_char`
- `find_next_char`
- `till_prev_char`
- `find_prev_char`
- `surround_add`
- `surround_replace` (this one even expects two arguments)
- `surround_delete`
- `select_textobject_around`
- `select_textobject_inner`

However, it doesn't seem possible to chain these together with other commands in keymaps.

For example:
```toml
[keys.normal]
f = ["find_next_char", "collapse_selection"]
```
does not work as expected. The `collapse_selection` command seems to be completely ignored.

Similarly, this
```toml
[keys.normal]
C = ["select_textobject_inner", "change_selection"]
```
seems to execute the `change_selection` command immediately, before `select_textobject_inner` has finished.

My enhancement request is that if these commands with arguments appear in a keymap, they should first finish completely and then the next command in the list should be called.

kl:
I'm taking a stab at fixing this. I got a fix that works but needs a bit more work before I open a pull request.

mitsuhiko:
The ones related to `select_textobject_inner` are basically sub menus but the commands behind the items in the popup do not have command names themselves. For instance there really should be IMO a `select_textobject_inner_word` to do `miw`. I wonder if it would be a good idea to just name those commands and expose them for custom mappings.

GenePg:
Although `w`, `e` and `b`, are great for selecting words automatically, I think `f` should change the default behavior like the key map below. In my case, It's more common for using without selecting. Is any one have the example of using `f` with selection commonly?

```
[keys.normal]
f = ["find_next_char", "collapse_selection"]
```

webdev23:
Not sure why, but to chain some commands, sometimes they requires a `:`, and sometimes not:
The following works fine for me. (Typing AltGr + d output ð on my French keyboard)
Notice the erratic usage of `:`.
It is a confusing behavior.

```
# Works fine: Duplicate line with AltGr + d
"ð" = ["extend_line_below", ":clipboard-yank", "open_below", ":clipboard-paste-before"]
```

```
# Doesn't work: Fail to load config (due to : on open_below)
"ð" = ["extend_line_below", ":clipboard-yank", ":open_below", ":clipboard-paste-before"]
```

ThinkChaos:
@archseer, I believe this isn't solved: the original examples are still not functional.

[...]

Issue #1595, opened 2022-01-28
Allow rebinding of text object selection keys (and show a dialog list)


eugenesvk:
Currently object selections like `select_textobject_inner` seem to have hardcoded keybinds for the supported objects like <kbd>w</kbd> word, <kbd>f</kbd> function, <kbd>c</kbd> class etc.

Since in my keybindings <kbd>w</kbd> isn't tied to a word, I'd also like to be able to change these internal keybindings as well to match my word-based keybinds from the Normal mode

I think it'd be great if there were a way to use Helix dialog popup list for this as well since currently `select_textobject_inner` doesn't show any hints unlike some other minor modes that do show the the full list of available keybinds

Issue #2317, opened 2022-04-29
Visual mode -Shift + V should select entire line


TheSYNcoder:
I would like helix to have the visual mode feature like in Vim where pressing shift + V selects the entire current line.
Currently, I have to manually select a line.

archseer:
You can always press `X` to expand your selection to line bounds.

PR #4287, opened 2022-10-15
Added non interactive select_textobject commands

mitsuhiko:
Hidden behind the `select_textobject_inner` and `select_textobject_around` menu are very useful selection options. However they cannot be directly bound to keys currently. You can only bring up the `select_textobject_inner` popup and then put a command into it. There are however quite a few cases where it's convenient to directly use one of the methods underneath.

This exposes most of these as commands with the exception of pair selection since there is no obvious best way to do that.

With this you can for instance do `select_textobject_inner_word` to do what `miw` does.

Loosely related issues:
* #4013 
* #1595

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -396,7 +396,25 @@ impl MappableCommand {
         surround_replace, "Surround replace",
         surround_delete, "Surround delete",
         select_textobject_around, "Select around object",
+        select_textobject_around_word, "Select word around object",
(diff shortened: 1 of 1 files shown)


thomkeh:
Related to #2317 ?

mitsuhiko:
Indeed. I did not find that PR somehow. That one goes significantly farther and also makes the keymaps configurable. I'm not sure if this PR makes sense then but I will leave it open for now. I think as a first step it might be a smaller change.

Issue #3584, opened 2022-08-28
tree-sitter independent bracket matching


sohnryang:
#### Describe your feature request
Currently helix editor depends heavily on tree-sitter to provide bracket matching. However, this makes the editor is unable to give correct results when tree-sitter does not have relevant information regarding brackets' placements. (examples are #1108, #2110, #3357 and #3487) Therefore, bracket matching which works regardless of tree-sitter availability is needed.

Ordoviz:
We need a combination of both: using tree-sitter means that unmatched brackets in comments or strings don't break bracket matching in the whole file, and tree-sitter independent matching is needed for plain text files and comments.

alevinval:
Also, if nothing was matched via tree sitter matching, we could fallback to the tree-sitter independent matching to increase chances of doing what the user expects. ( that could be helpful in cases like people editing YAML, for which there is a tree-sitter grammar, but you may still expect some plain text matching to work )

alevinval:
This issue can now be closed after #4288 was merged

Issue #3357, opened 2022-08-08
bracket matching works incrrectly sometimes


A-Walrus:
### Summary

For example in this example, with the cursor on either one of the parentheses in `Some(thing)`, the matching one is not highlighted and hitting `mm` doesn't take you to the matching one, instead it takes you to the bracket on line 6.
```rust
fn main() {
    let opt = None;
    match opt {
        Some(thing) => {},
        None
    }
}
```

### Reproduction Steps

 
[asciinema](https://asciinema.org/a/513464)

I tried this:

1. `hx`
2. open the rust file with the example code from above, make sure the language is set to rust
3. put the cursor on either one of the parentheses in `Some(thing)`
4. hit `mm`
I expected this to happen:
The matching parenthesis is highlighted in step 3, and moved to in step 4

Instead, this happened:
Nothing is highlighted in step 3, and the bracket on line 6 is gone to in step 4

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

22.05-376-g6b84344e

A-Walrus:
This appears to be because the bracket matching expects matching brackets to be the border of a tree sitter node, but in this case `Some`, `(`, `thing` and `)` are all children of the same node, with no separate parent for `(thing)`.
This is unlike the case of 
```rust
let a = Some(5);
```
where `(5)` are grouped separately under an `arguments` node.

Is this an issue/bug in tree-sitter-rust (that they should group the `(thing)` in the above example), or should we change the matching bracket code on our end?

the-mikedavis:
If matching brackets within a node is possible without behavior regressions, that'd be preferrable. tree-sitter-rust actually already exposes a node just for the `(5)` part and the parens are children of that

```scm
(call_expression
  function: (identifier)   ; "Some"
  arguments:
    (arguments             ; "(5)"
       (integer_literal))) ; "5"
```

So I'm not sure there's a change in tree-sitter-rust we can make to improve the situation.

A-Walrus:
> tree-sitter-rust actually already exposes a node just for the (5) part and the parens are children of that

[...]

PR #4288, opened 2022-10-15
feat(core): add plaintext matching fallback to tree-sitter matching

alevinval:
This patch introduces bracket matching independent of tree-sitter grammar. This matching is introduced as a fallback in cases where the tree-sitter matcher does not match any bracket. The fallback should provide a better experience in cases like the ones reported in #3614

Relates to #3584 as well

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -96,27 +96,29 @@ pub fn find_matching_bracket_current_line_plaintext(
 
     // Determine the direction of the matching.
     let is_fwd = is_forward_bracket(bracket);
-    let line = doc.byte_to_line(cursor_pos);
-    let end = doc.line_to_byte(if is_fwd { line + 1 } else { line });
-    let range = if is_fwd {
-        Range::Forward((cursor_pos + 1)..end)
+    let chars_iter = if is_fwd {
+        doc.chars_at(cursor_pos + 1)
     } else {
-        Range::Backward((end..cursor_pos).rev())
+        doc.chars_at(cursor_pos).reversed()
     };
 
     let mut open_cnt = 1;
 
-    for pos in range {
-        if doc.char(pos) == bracket {
+    for (i, candidate) in chars_iter.enumerate() {
+        if candidate == bracket {
             open_cnt += 1;
         } else if is_valid_pair(
             doc,
-            if is_fwd { cursor_pos } else { pos },
-            if is_fwd { pos } else { cursor_pos },
+            if is_fwd { cursor_pos } else { cursor_pos - i - 1 },
+            if is_fwd { cursor_pos + i + 1 } else { cursor_pos },
         ) {
             // Return when all pending brackets have been closed.
             if open_cnt == 1 {
-                return Some(pos);
+                return Some(if is_fwd {
+                    cursor_pos + i + 1
+                } else {
+                    cursor_pos - i - 1
+                });
             }
             open_cnt -= 1;
         }
@@ -148,21 +150,6 @@ fn surrounding_bytes(doc: &Rope, node: &Node) -> Option<(usize, usize)> {
     }
 
     Some((start_byte, end_byte))
-}
-
-enum Range {
-    Forward(std::ops::Range<usize>),
-    Backward(std::iter::Rev<std::ops::Range<usize>>),
-}
-
-impl Iterator for Range {
-    type Item = usize;
-    fn next(&mut self) -> Option<usize> {
-        match self {
-            Range::Forward(range) => range.next(),
-            Range::Backward(range) => range.next(),
-        }
-    }
 }
 
 #[cfg(test)]
@@ -198,11 +185,8 @@ mod tests {
         assert("(paren (paren {bracket}))", 7, 23);
         assert("(paren (paren {bracket}))", 14, 22);
 
-        // Verify that this feature only works on the current line.
-        assert_no_match("(prev line\n ) (middle) ( \n next line)", 0);
-        assert_no_match("(prev line\n ) (middle) ( \n next line)", 12);
+        assert("(prev line\n ) (middle) ( \n next line)", 0, 12);
         assert("(prev line\n ) (middle) ( \n next line)", 14, 21);
-        assert_no_match("(prev line\n ) (middle) ( \n next line)", 23);
-        assert_no_match("(prev line\n ) (middle) ( \n next line)", 36);
+        assert("(prev line\n ) (middle) ( \n next line)", 23, 36);
     }
 }
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -70,8 +70,61 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
     }
 }
 
+// Returns the position of the bracket that is closing, searching only in
+// the current line, and works on plain text, ignoring the tree-sitter grammar.
+//
+// If the cursor is on an opening or closing bracket, the function
+// behaves equivalent to [`find_matching_bracket`].
+//
+// If no matchig bracket is found on the current line, returns None.
+#[must_use]
+pub fn find_matching_bracket_current_line_plaintext(doc: &Rope, pos: usize) -> Option<usize> {
+    // Don't do anything when the cursor is not on top of a bracket.
+    let mut c = doc.char(pos);
+    if !is_valid_bracket(c) {
+        return None;
+    }
+
+    let bracket = c;
+    let bracket_pos = pos;
+
+    // Determine the direction of the matching
+    let is_fwd = is_forward_bracket(c);
+    let line = doc.byte_to_line(pos);
+    let end = doc.line_to_byte(if is_fwd { line + 1 } else { line });
+    let range = if is_fwd {
+        Range::Forward((pos + 1)..end)
+    } else {
+        Range::Backward((end..pos).rev())
+    };
+
+    let mut open_cnt = 1;
+
+    for pos in range {
+        c = doc.char(pos);
+
+        if !is_valid_bracket(c) {
+            continue;
+        } else if bracket == c {
+            open_cnt += 1;
+        } else if is_valid_pair(doc, bracket_pos, pos) || is_valid_pair(doc, pos, bracket_pos) {
+            open_cnt -= 1;
+        }
+
+        if open_cnt == 0 {
+            return Some(pos);
+        }
+    }
+
+    None
+}
+
 fn is_valid_bracket(c: char) -> bool {
     PAIRS.iter().any(|(l, r)| *l == c || *r == c)
+}
+
+fn is_forward_bracket(c: char) -> bool {
+    PAIRS.iter().any(|(l, _)| *l == c)
 }
 
 fn is_valid_pair(doc: &Rope, start_char: usize, end_char: usize) -> bool {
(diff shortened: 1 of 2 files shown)


alevinval:
Let me know your thoughts about this.

helix-term/src/commands.rs
```diff
@@ -4090,7 +4090,14 @@ fn match_brackets(cx: &mut Context) {
             {
                 range.put_cursor(text, pos, cx.editor.mode == Mode::Select)
             } else {
-                range
+                if let Some(pos) = match_brackets::find_matching_bracket_current_line_plaintext(
```
mitsuhiko, line 4093: This block can be collapsed. Pretty sure clippy complains about this.

workingj:
hi, any progress here possible?

alevinval:
@archseer thoughts on this PR?

helix-core/src/match_brackets.rs
```diff
@@ -70,10 +72,67 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
...
+) -> Option<usize> {
+    // Don't do anything when the cursor is not on top of a bracket.
+    let bracket = doc.char(cursor_pos);
+    if !is_valid_bracket(bracket) {
+        return None;
+    }
+
+    // Determine the direction of the matching.
+    let is_fwd = is_forward_bracket(bracket);
+    let line = doc.byte_to_line(cursor_pos);
```
pascalkuthe, line 99: positions are stored as char (utf-32) instead of byte (utf-8) positions in helix so `char_to_line` should be used here. 

That being said I don't actually see the point of limiting the matching to the current line. You could just search `cursor_pos + 1..` and `..cursor_pos`.rev() instead  that seems like it needs less code while being more useful.

Why did you introduce the limitation of limiting the matching to the current line?

helix-core/src/match_brackets.rs
```diff
@@ -90,3 +149,60 @@ fn surrounding_bytes(doc: &Rope, node: &Node) -> Option<(usize, usize)> {
...
+
+impl Iterator for Range {
+    type Item = usize;
+    fn next(&mut self) -> Option<usize> {
+        match self {
+            Range::Forward(range) => range.next(),
+            Range::Backward(range) => range.next(),
+        }
+    }
+}
```
pascalkuthe, line 166: I think it's better to avoid introducing a custom type here. Ropeys iterators are extremly flexible and allow both a more performant and more correct implementation.
Simply use `doc.chars_at(cursor_pos).reversed()`  and `doc.chars_at(cursor_pos + 1)`  for reverse/forward iteration respectively. These both return the same type so no need for an extra enum. You can either use `.enumerate()` to track your position in the iterator and return `char_pos - i - 1` in the reverse case and `char_pos + 1 + i` in the forward case

helix-term/src/commands.rs
```diff
@@ -4083,12 +4083,15 @@ fn match_brackets(cx: &mut Context) {
...
-        let text = doc.text().slice(..);
+        let text = doc.text();
+        let text_slice = text.slice(..);
+        let is_select = cx.editor.mode == Mode::Select;
         let selection = doc.selection(view.id).clone().transform(|range| {
-            if let Some(pos) =
-                match_brackets::find_matching_bracket_fuzzy(syntax, doc.text(), range.cursor(text))
+            let pos = range.cursor(text_slice);
+            if let Some(pos) = match_brackets::find_matching_bracket_fuzzy(syntax, text, pos)
+                .or_else(|| match_brackets::find_matching_bracket_current_line_plaintext(text, pos))
```
pascalkuthe, line 4092: This doesn't seem to be quite what we want. This always perform the plaintext matching if the TS matching doesn't find a match. Meanwhile for plaintext files this won't work as in that case `doc.syntax()` is `None` (condition above).

I think we can do better for treesitter than the current implementation (which gets easily broken by comments for example). That solution might partially reuse your solution but would be constrained to not leave the current TS node. You can decide whether you want to do that in this PR or a followup PR, that would be the right fix for https://github.com/helix-editor/helix/issues/3357

As it's implemented right now I think it's better to only us the plaintext matching for files without a TS grammar (so `doc.syntax()` would be `None`)

helix-core/src/match_brackets.rs
alevinval: I just wanted to keep things simple and avoid scanning the entire document, seems like potentially expensive, so by constraining to only a line it would ensure there's no performance hit, while providing some usefulness already. I've removed it now.

helix-core/src/match_brackets.rs
alevinval: Yeah, I did not realise how simple it is using the rope `chars_at`.

helix-term/src/commands.rs
alevinval: I see, this fallback will never run when there's no grammar. However, most of the users of this editor will be coding, and likely have an active TS grammar. My hope is that adding this fallback now already brings improvement, and eventually it will be useless if the TS logic improves.

archseer (changes requested):

helix-core/src/match_brackets.rs
```diff
@@ -70,10 +72,77 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
...
+// is limited to the current line. This function works on plain text, it
+// ignores tree-sitter grammar.
+//
+// If the cursor is on the opening bracket, the position of
+// the closing bracket is returned. If the cursor on the closing
+// bracket, the position of the opening bracket is returned.
+//
+// If the cursor is not on a bracket, `None` is returned.
+//
+// If no matching bracket is found, `None` is returned.
```
A-Walrus, line 85: This should probably be a doc-comment (`///`) as opposed to a regular comment

helix-core/src/match_brackets.rs
```diff
@@ -15,11 +15,13 @@ const PAIRS: &[(char, char)] = &[
...
 //
-// If the cursor is one the opening bracket, the position of
-// the closing bracket is returned. If the cursor in the closing
+// If the cursor is on the opening bracket, the position of
+// the closing bracket is returned. If the cursor on the closing
 // bracket, the position of the opening bracket is returned.
 //
 // If the cursor is not on a bracket, `None` is returned.
+//
+// If no matching bracket is found, `None` is returned.
```
A-Walrus, line 24: This should probably be a doc-comment (`///`) as opposed to a regular comment

A-Walrus (changes requested):
A few style nits, other than that and what has already been said by @pascalkuthe and @archseer looks good!

helix-core/src/match_brackets.rs
alevinval: https://github.com/helix-editor/helix/pull/4288/commits/f5cb382af064d6db6df2e5da89913ae24e363ee5

helix-core/src/match_brackets.rs
alevinval: https://github.com/helix-editor/helix/pull/4288/commits/f5cb382af064d6db6df2e5da89913ae24e363ee5

helix-core/src/match_brackets.rs
```diff
@@ -70,10 +72,77 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
...
+    let is_fwd = is_forward_bracket(bracket);
+    let chars_iter = if is_fwd {
+        doc.chars_at(cursor_pos + 1)
+    } else {
+        doc.chars_at(cursor_pos).reversed()
+    };
+
+    let mut open_cnt = 1;
+
+    for (i, candidate) in chars_iter.enumerate() {
```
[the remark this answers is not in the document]
alevinval: Done here:
https://github.com/helix-editor/helix/pull/4288/commits/715b928dae00d8048f32ce8c6527a9d45f7d4f70

helix-term/src/commands.rs
alevinval: That should settle it: https://github.com/helix-editor/helix/pull/4288/commits/adf0fbd3c8a663ceeaba3050f955c4bdec242381

alevinval:
Folks! I've been out of the loop for some time. I got back into this, I've rebased latest master to make sure everything kept working, the pending feedback that was requested has been applied in: https://github.com/helix-editor/helix/pull/4288/commits/adf0fbd3c8a663ceeaba3050f955c4bdec242381 https://github.com/helix-editor/helix/pull/4288/commits/715b928dae00d8048f32ce8c6527a9d45f7d4f70 and https://github.com/helix-editor/helix/pull/4288/commits/f5cb382af064d6db6df2e5da89913ae24e363ee5

pascalkuthe (commented):
One minor thing the rest lgtm

helix-term/src/commands.rs
```diff
@@ -4528,20 +4528,24 @@ fn select_prev_sibling(cx: &mut Context) {
...
-            }
-        });
-        doc.set_selection(view.id, selection);
-    }
+    let selection = doc.selection(view.id).clone().transform(|range| {
+        let pos = range.cursor(text_slice);
+        if let Some(matched_pos) = doc
+            .syntax()
+            .and_then(|syntax| match_brackets::find_matching_bracket_fuzzy(syntax, text, pos))
+            .or_else(|| match_brackets::find_matching_bracket_current_line_plaintext(text, pos))
```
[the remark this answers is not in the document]
alevinval: And you are correct, addressed: https://github.com/helix-editor/helix/pull/4288/commits/67099d1a77f6868e5774b84a3e220d719747d02c

pascalkuthe (dismissed):
LGTM 👍

dead10ck (dismissed):
Looks good to me, thanks for your contribution!

helix-core/src/match_brackets.rs
```diff
@@ -70,10 +73,77 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
...
+    let is_fwd = is_forward_bracket(bracket);
+    let chars_iter = if is_fwd {
+        doc.chars_at(cursor_pos + 1)
+    } else {
+        doc.chars_at(cursor_pos).reversed()
+    };
+
+    let mut open_cnt = 1;
+
+    for (i, candidate) in chars_iter.take(MAX_PLAINTEXT_SCAN).enumerate() {
```
dead10ck, line 108: One minor thing I'm noticing is that this doesn't limit the search to the current line only, as it advertises; it simply limits the scan to a max number of chars. I actually think this is fine, because limiting matching to the current line only seems unnecessarily restrictive to me, but we should probably update the function name and doc comment.

pascalkuthe: yeah it was changed to a fixed limit after feedback from @archseer and me. I missed that the doc comment wasn't updated good catch

alevinval: Do you folks use something to embed a constant inside a Doc? I wanted to do this:
```patch
diff --git a/helix-core/src/match_brackets.rs b/helix-core/src/match_brackets.rs
index 0ec040e5..07eb261a 100644
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -73,9 +73,9 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
     }
 }
 
-/// Returns the position of the matching bracket under cursor, the search
-/// is limited to the current line. This function works on plain text, it
-/// ignores tree-sitter grammar.
+/// Returns the position of the matching bracket under cursor.
+/// This function works on plain text and ignores tree-sitter grammar.
+/// The search is limited to [`MAX_PLAINTEXT_SCAN`] characters
 ///
 /// If the cursor is on the opening bracket, the position of
 /// the closing bracket is returned. If the cursor on the closing

```

But that would require making the constant public for the docs to link to it correctly. If you are okay with that anyway (it works, but does not link) then I can push this

dead10ck: That seems fine to me 👍

Issue #4153, opened 2022-10-09
Add numeric :sort and :rsort


rypervenche:
I don't know if this would need to be a new command or some sort of option to the existing commands, but I think it would be very useful to have these. I found a workaround to pipe into the coreutils' `sort -n`, but the speed was really bad. It would be nice to be able to do this directly in Helix, if nothing else than for the speed boost.

Thanks!

ArikRahman:
What are you looking for the command to do exactly? Sort the numbers that are entered after the command?

kirawi:
`:sort` rearranges selections based on their order. What this issue is looking for is something like https://crates.io/crates/alphanumeric-sort

ArikRahman:
So in theory if 1, 3, 5, 10, 2 were selected, it would replace the selection with 1, 2, 3, 5, 10? What sorting algorithm would be best, and what other numbers would be supported besides the standard decimal 1-9 for example?

timokoesters:
Just as another example, in vim I also like to use sort with logs to group similar items:
```
INFO sfasf
WARN aadgsljdf
INFO asdlf
WARN lasdjf
```
```
INFO asdlf
INFO sfasf
WARN aadgsljdf
WARN lasdjf
```

philipgiuliani:
This already works. Keep in mind that Helix in contrast to VIM only sorts selections. So you have to create one selection per line first (Using S to split by newline).`%S\n<ret>:sort`

the-mikedavis:
As a shortcut you can also use `A-s` (Alt s) to split a selection into lines

PR #4289, opened 2022-10-15
Add 'alphanumeric sort' commands

seb-bl:
Implements #4153 with the crate mentioned in the discussion of the issue.

This adds 2 new commands, `sortn` and `rsortn`, corresponding to the existing ones (`sort` and `rsort`), and they sort according to https://docs.rs/alphanumeric-sort/1.4.4/alphanumeric_sort/fn.compare_str.html

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1358,7 +1358,7 @@ fn sort(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> 
         return Ok(());
     }
 
-    sort_impl(cx, args, false)
+    sort_impl(cx, args, false, false)
 }
 
 fn sort_reverse(
@@ -1370,13 +1370,38 @@ fn sort_reverse(
         return Ok(());
     }
 
-    sort_impl(cx, args, true)
+    sort_impl(cx, args, true, false)
+}
+
+fn sort_numeric(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    sort_impl(cx, args, false, true)
+}
+
+fn sort_numeric_reverse(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    sort_impl(cx, args, true, true)
 }
 
 fn sort_impl(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
     reverse: bool,
+    alphanumeric: bool,
 ) -> anyhow::Result<()> {
     let (view, doc) = current!(cx.editor);
     let text = doc.text().slice(..);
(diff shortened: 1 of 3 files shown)


the-mikedavis:
It's a short (~100L) block of code within the library and the library includes a bunch of other functions I don't think will be used in the future so I think we should vendor this block https://github.com/magiclen/alphanumeric-sort/blob/bea0284409f68407dd25b432fd901b62f6e944e7/src/lib.rs#L95-L199

kirawi:
I think that https://github.com/Aloso/lexical-sort might be better since it claims to support unicode.

seb-bl:
I removed the dependency, and added a function to compare the strings.

When I copied the source, I realized the handling of long numbers was incorrect (it accumulates it in a `f64`, thus rounds it). I looked at the other crate, and it had a similar problem (accumulating the number in a `u64`).

So I implemented my own function. It orders numbers with leading zeroes like the crate `alphanumeric-sort`, but uses the slice containing the digits instead of a `f64` for the comparison

kirawi:
Just to note, I think that using a `u64` would've been okay. `len()` returns a `usize` which is equivalent to `u64` on 64bit systems.

helix-term/src/commands/typed.rs
```diff
@@ -1407,6 +1434,87 @@ fn sort_impl(
...
+                }
+            }
+            (Some(_), None) => return Ordering::Greater,
+            (None, Some(_)) => return Ordering::Less,
+            (None, None) => return Ordering::Equal,
+        }
+    }
+}
+
+#[test]
```
kirawi, line 1493: This should be moved into a
```rs
#[cfg(test)]
mod tests {
}
```
block at the bottom of the file.

kirawi:
By the way, make sure to run `cargo xtask docgen` and commit the changes to make CI pass.

seb-bl:
Thanks for the hints 🙂

I think I have not explained the problem with the `u64` correctly. When encountering a number,`lexical-sort` parses the number into a `u64`, digit by digit, it multiplies the number by 10 and add the current digit ([at this line](https://github.com/Aloso/lexical-sort/blob/main/src/cmp.rs#L14-L15)). So for example "18446744073709551616" (=2^64) would cause an overflow. Numbers this long are not very common, but may still happen

helix-term/src/commands/typed.rs
```diff
@@ -1370,13 +1370,38 @@ fn sort_reverse(
...
+
+    sort_impl(cx, args, true, true)
 }
 
 fn sort_impl(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
     reverse: bool,
+    alphanumeric: bool,
 ) -> anyhow::Result<()> {
```
leandrobbraga, line 1405: As a suggestion you remove both `reverse` and `alphanumeric` and pass a single arguments called `cmp_fn` which is a `FnMut(&Tendril, &Tendril) -> Ordering` leaving to the caller to say how it want the ordering.

See: https://github.com/helix-editor/helix/pull/8436

helix-term/src/commands/typed.rs
```diff
@@ -1388,9 +1413,11 @@ fn sort_impl(
...
 
-    fragments.sort_by(match reverse {
-        true => |a: &Tendril, b: &Tendril| b.cmp(a),
-        false => |a: &Tendril, b: &Tendril| a.cmp(b),
+    fragments.sort_by(match (reverse, alphanumeric) {
+        (true, false) => |a: &Tendril, b: &Tendril| b.cmp(a),
+        (false, false) => |a: &Tendril, b: &Tendril| a.cmp(b),
+        (true, true) => |a: &Tendril, b: &Tendril| numeric_cmp(b, a),
+        (false, true) => |a: &Tendril, b: &Tendril| numeric_cmp(a, b),
     });
```
the-mikedavis, line 1421: I agree with @leandrobbraga - this block is a little dense and we can make the call sites for `sort_impl` a little cleaner by moving these closures to parameters of `sort_impl`.

So this function would replace the `reverse` parameter with a `compare_fn: impl FnMut(&Tendril, &Tendril) -> std::cmp::Ordering,` param and then this block can be just `fragments.sort_by(compare_fn);`.

We could also use this opportunity to drop the `args` parameter which is currently unused

kirawi:
I was looking into this kind of thing for a personal project and I don't think that alphanumeric-sort provides true lexicographic ordering. For that, we'd want the Unicode Collation algorithm. I'm currently a little less than halfway into my naive implementation of the algorithm that would be fully compliant so I may post a PR for it in the future.

PR #4295, opened 2022-10-15
Pretty print `tree-sitter-subtree` expression

### LINKED ISSUE #2880: Improve output of tree-sitter-subtree
Current output of `:tree-sitter-subtree`

image

This reads fine for small code snippets, but for big blocks it gets kind of unreadable.
A way of improving this, from `the-mikedavis` from helix Matrix chat:

```
we could format these like tree-sitter parse pretty easily, i.e. break lines and do indentation
```

fisherdarling:
Closes #2880

This PR indents and line-breaks the expression displayed for `:tree-sitter-subtree`. For short sexprs it might be a little annoying. Maybe we should generate the sexpr and if it's longer than ~40 chars, manually format it?

I didn't know where to put the formatting function, so I just put it inside the command itself.

Before:
Screen Shot 2022-10-15 at 3 36 02 PM

After:
Screen Shot 2022-10-15 at 3 36 48 PM

With named fields:
Screen Shot 2022-10-15 at 3 51 13 PM

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -2024,18 +2024,18 @@ impl<I: Iterator<Item = HighlightEvent>> Iterator for Merge<I> {
     }
 }
 
-pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node<'_>) -> fmt::Result {
+pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node) -> fmt::Result {
     pretty_print_tree_impl(fmt, node, true, None, 0)
 }
 
 fn pretty_print_tree_impl<W: fmt::Write>(
     fmt: &mut W,
-    node: Node<'_>,
+    node: Node,
     is_root: bool,
     field_name: Option<&str>,
     depth: usize,
 ) -> fmt::Result {
-    fn is_visible(node: Node<'_>) -> bool {
+    fn is_visible(node: Node) -> bool {
         node.is_missing()
             || (node.is_named() && node.language().node_kind_is_visible(node.kind_id()))
     }
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -2024,6 +2024,57 @@ impl<I: Iterator<Item = HighlightEvent>> Iterator for Merge<I> {
     }
 }
 
+pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node<'_>) -> fmt::Result {
+    pretty_print_tree_impl(fmt, node, true, None, 0)
+}
+
+fn pretty_print_tree_impl<W: fmt::Write>(
+    fmt: &mut W,
+    node: Node<'_>,
+    is_root: bool,
+    field_name: Option<&str>,
+    depth: usize,
+) -> fmt::Result {
+    fn is_visible(node: Node<'_>) -> bool {
+        node.is_missing()
+            || (node.is_named() && node.language().node_kind_is_visible(node.kind_id()))
+    }
+
+    if is_visible(node) {
+        let indentation_columns = depth * 2;
+        write!(fmt, "{:indentation_columns$}", "")?;
+
+        if let Some(field_name) = field_name {
+            write!(fmt, "{}: ", field_name)?;
+        }
+
+        write!(fmt, "({}", node.kind())?;
+    } else if is_root {
+        write!(fmt, "(\"{}\")", node.kind())?;
+    }
+
+    for child_idx in 0..node.child_count() {
+        if let Some(child) = node.child(child_idx) {
+            if is_visible(child) {
+                fmt.write_char('\n')?;
+            }
+
+            pretty_print_tree_impl(
+                fmt,
+                child,
+                false,
+                node.field_name_for_child(child_idx as u32),
+                depth + 1,
+            )?;
+        }
+    }
+
+    if is_visible(node) {
+        write!(fmt, ")")?;
+    }
+
+    Ok(())
+}
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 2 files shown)


Chickenkeeper:
I frequently hit situations where a subtree is too long for one monitor and I have to stretch my terminal across two to see the end of it, so this would be great to have

helix-term/src/commands/typed.rs
```diff
@@ -1473,7 +1473,57 @@ fn tree_sitter_subtree(
...
             .descendant_for_byte_range(from, to)
         {
-            let contents = format!("```tsq\n{}\n```", selected_node.to_sexp());
+            fn pretty_print_node(
+                node: Node<'_>,
+                sexpr: &mut String,
+                is_root: bool,
+                field_name: Option<&str>,
+                depth: usize,
+            ) -> anyhow::Result<()> {
```
the-mikedavis, line 1482: I think we can move this to `helix_core::syntax` and it would be good to add a test case in there. Also, I think this should take a `std::fmt::Write` implementation rather than a String and be split into two functions so it has a friendlier function head:

```rs
pub fn pretty_print_tree<W: std::fmt::Write>(
    fmt: &mut W,
    node: Node,
) -> Result<(), std::fmt::Error> {
    pretty_print_tree_impl(fmt, node, true, None, 0)
}

pub fn pretty_print_tree_impl<W: std::fmt::Write>(
    fmt: &mut W,
    node: Node,
    is_root: bool,
    field_name: Option<&str>,
    depth: usize,
) -> Result<(), std::fmt::Error> {
  // the recursive one
}
```

helix-term/src/commands/typed.rs
```diff
@@ -1473,7 +1473,57 @@ fn tree_sitter_subtree(
...
+                        if is_visible(child) {
+                            sexpr.push('\n');
+                        }
+
+                        pretty_print_node(
+                            child,
+                            sexpr,
+                            false,
+                            node.field_name_for_child(child_idx as u32),
+                            depth + 2,
```
the-mikedavis, line 1511: Depth seems to be conflated with the indentation level. I think it's clearer to separate them: `let indentation_columns = 2 * depth;`

helix-term/src/commands/typed.rs
fisherdarling: Sounds good! Thanks for the pointer on where to put it :)

helix-term/src/commands/typed.rs
```diff
@@ -1473,7 +1473,57 @@ fn tree_sitter_subtree(
...
+                    write!(sexpr, ")")?;
+                }
+
+                Ok(())
+            }
+
+            let mut sexpr = String::new();
+            pretty_print_node(selected_node, &mut sexpr, true, None, 0)?;
+
+            let contents = format!("```tsq\n{}\n```", sexpr);
```
poliorcetics, line 1526: This will reallocate a string only to over `sexpr`, you could start `sexpr` with 

```rust
let mut sexpr = String::from("```tsq\n");
pretty_print_node(selected_node, &mut sexpr, true, None, 0)?;
sexpr.push_str("\n```");
```

helix-term/src/commands/typed.rs
fisherdarling: I moved it to `helix_core::syntax` and added some tests.

helix-term/src/commands/typed.rs
```diff
@@ -1473,7 +1474,9 @@ fn tree_sitter_subtree(
             .root_node()
             .descendant_for_byte_range(from, to)
         {
-            let contents = format!("```tsq\n{}\n```", selected_node.to_sexp());
+            let mut contents = String::from("```tsq\n");
+            pretty_print_tree(&mut contents, selected_node)?;
```
the-mikedavis, line 1478: Very tiny nit but I'd prefer this be fully qualified as `helix_core::syntax::pretty_print_tree(&mut contenst, selected_node)?` since we're unlikely to re-use the function elsewhere in this file

fisherdarling: Makes sense, just pushed a commit :)

helix-core/src/syntax.rs
```diff
@@ -2024,6 +2024,57 @@ impl<I: Iterator<Item = HighlightEvent>> Iterator for Merge<I> {
...
+        if let Some(field_name) = field_name {
+            write!(fmt, "{}: ", field_name)?;
+        }
+
+        write!(fmt, "({}", node.kind())?;
+    } else if is_root {
+        write!(fmt, "(\"{}\")", node.kind())?;
+    }
+
+    for child_idx in 0..node.child_count() {
```
the-mikedavis, line 2056: This function can be simplified a lot if we use `Node::named_children` here instead of iterating through the children by index. We can eliminate passing `is_root` and checking if nodes are visible:

```rs
pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node<'_>) -> fmt::Result {
    pretty_print_tree_impl(fmt, node, None, 0)
}

fn pretty_print_tree_impl<W: fmt::Write>(
    fmt: &mut W,
    node: Node,
    field_name: Option<&str>,
    depth: usize,
) -> fmt::Result {
    let indentation_columns = depth * 2;
    write!(fmt, "{:indentation_columns$}", "")?;

    if let Some(field_name) = field_name {
        write!(fmt, "{}: ", field_name)?;
    }

    write!(fmt, "({}", node.kind())?;

    for child in node.named_children(&mut node.walk()) {
        fmt.write_char('\n')?;

        pretty_print_tree_impl(
            fmt,
            child,
            node.field_name_for_child(child.id() as u32),
            depth + 1,
        )?;
    }

    write!(fmt, ")")?;

    Ok(())
}
```

The tree-sitter docs recommend using `children`/`named_children` with `TreeCursor`s rather than `child`/`named_child` when you expect to iterate through all of the children: https://github.com/tree-sitter/tree-sitter/blob/cabdcfb41d780c760a2d0bea32f743b7d9ac8e4e/lib/binding_rust/lib.rs#L918-L941

fisherdarling: So I think the reason I didn't go with the named_children approach is because the child's id doesn't not index correctly. I originally did something similar but decided to just plainly iterate after looking at the `to_sexp` c source in tree-sitter.

Here's the output with the suggested fn:
```
// expected
(function_item
  name: (identifier)
  parameters: (parameters)
  body: (block
    (expression_statement
      (macro_invocation
        macro: (identifier)
        (token_tree
          (string_literal))))))
// output
(function_item
  (identifier)
  (parameters)
  (block
    (expression_statement
      (macro_invocation
        (identifier)
        (token_tree
          (string_literal))))))
```
I wonder if using the iter's index will work instead of the node's id; `.iter().enumerate()`. I'll see if that works...

fisherdarling: Alright, so using this as the inner loop:
```rust
for (i, child) in node.named_children(&mut node.walk()).enumerate() {
     fmt.write_char('\n')?; 
     pretty_print_tree_impl(fmt, child, node.field_name_for_child(i as u32), depth + 1)?;
}
```
Leads to this:
```
// expected:
(function_item
  name: (identifier)
  parameters: (parameters)
  body: (block
    (expression_statement
      (macro_invocation
        macro: (identifier)
        (token_tree
          (string_literal))))))
// printed
(function_item
  (identifier)
  name: (parameters)
  parameters: (block
    (expression_statement
      (macro_invocation
        macro: (identifier)
        (token_tree
          (string_literal))))))
```
I don't know if there's a way to cleanly use the `TreeCursor` here. My original function is a modified version of the c `to_sexp` implementation, which I don't believe uses a cursor either: https://github.com/tree-sitter/tree-sitter/blob/master/lib/src/subtree.c#L894

the-mikedavis: Ah good catch with the field name. In order to use the TreeCursor, I think we would really need lean in to the TreeCursor API: use `TreeCursor::goto_first_child`, `TreeCursor::goto_next_sibling`, `TreeCursor::field_name` etc.

The implementation here I think is more readable than that would be so I think we should just use this 👍

helix-core/src/syntax.rs
```diff
@@ -2024,6 +2024,57 @@ impl<I: Iterator<Item = HighlightEvent>> Iterator for Merge<I> {
...
+}
+
+fn pretty_print_tree_impl<W: fmt::Write>(
+    fmt: &mut W,
+    node: Node<'_>,
+    is_root: bool,
+    field_name: Option<&str>,
+    depth: usize,
+) -> fmt::Result {
+    fn is_visible(node: Node<'_>) -> bool {
```
the-mikedavis, line 2038: Specifying the anonymous lifetimes here is not necessary:

```suggestion
pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node) -> fmt::Result {
    pretty_print_tree_impl(fmt, node, true, None, 0)
}

fn pretty_print_tree_impl<W: fmt::Write>(
    fmt: &mut W,
    node: Node,
    is_root: bool,
    field_name: Option<&str>,
    depth: usize,
) -> fmt::Result {
    fn is_visible(node: Node) -> bool {
```

fisherdarling: Added a fix! Thanks :)

fisherdarling:
Also @the-mikedavis if you merge this, could you add the `hacktoberfest-accepted` label for me? And I just want to add that this PR was the first time I used helix and I'm addicted. I'm trying to use it in every side-project now hoping to get productive enough to use it full-time at work! Thank you and the rest of the contributors for your work on the project :)

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4297, opened 2022-10-15
add Chinese documentation link

zjp-CN:


the-mikedavis:
I'm hesitant to link to non-official documentation from the official docs

A-Walrus:
> I'm hesitant to link to non-official documentation from the official docs

IMO it's fine as long as it as **clearly labeled as unofficial.**

zjp-CN:
We can put it in the wiki.

the-mikedavis:
I think adding it to the main wiki page would be good 👍 https://github.com/helix-editor/helix/wiki

kirawi:
We could also host the docs in the repo.

zjp-CN:
> We could also host the docs in the repo.

Thanks for the idea. But I think it will bring drawbacks more then benefits.

I've updated the Wiki. So I'm gonna close this PR. Thanks all.

PR #4299, opened 2022-10-15
Optimize `ui::menu::Item::label()` implementations

poliorcetics:
- nit: move an allocation to happen after a `continue`, making sure it's not done for nothing
- nit: Do less allocations in `ui::menu::Item::label` implementations, often dividing by two or more
  the number of allocations in a call

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -54,7 +54,7 @@ use crate::{
 
 use crate::job::{self, Jobs};
 use futures_util::StreamExt;
-use std::{collections::HashMap, fmt, future::Future};
+use std::{collections::HashMap, fmt, fmt::Write, future::Future};
 use std::{collections::HashSet, num::NonZeroUsize};
 
 use std::{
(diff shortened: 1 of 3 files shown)


helix-term/src/commands.rs
```diff
@@ -2410,17 +2410,39 @@ impl ui::menu::Item for MappableCommand {
...
-                        .collect::<Vec<String>>()
-                        .join("+")
-                })
-                .collect::<Vec<String>>()
-                .join(", ")
+            const OOM: &str = "Writing to a string can only fail on an Out-Of-Memory error";
+
+            let mut res = String::new();
+
+            fn write_binding(res: &mut String, binds: &[KeyEvent]) {
```
dead10ck, line 2422: Thanks for your optimization PRs! I'm glad to see contributors helping improve performance where possible.

Although, in my opinion, this particular change is a regression in readability. The original code is pretty clear and concise in its use of combinators, whereas this requires more than a quick glance to understand what it's doing.

I applaud optimizations in general, but personally I think readability is more important if performance is not suffering. And this code is not in any kind of hot path as far as I can tell.

poliorcetics: Yes I could remove this one, it's for the command palette, not as big of a deal as the one for LSP references. I didn't check if helix already has `itertools` in its dependencies, I may be able to avoid allocations without sacrificing readability that way

poliorcetics: Made a much clearer version using `fold`

archseer (approved):
Thanks! Looks good to me, but can you resolve the conflict?

poliorcetics:
Done :)

Issue #4211, opened 2022-10-11
Add syntax definition / highlight for Groovy, specifically Jenkinsfiles


weakphish:
Currently, no syntax highlighting exists for Jenkinsfiles, which are written in Groovy.

zephaniahong:
Hey! I'd be happy to give this issue a shot. @the-mikedavis May I get some advice as to which files/directories I should be looking at?

Thank you!

the-mikedavis:
Take a look at the adding languages guide https://docs.helix-editor.com/guides/adding_languages.html and recent PRs that added languages https://github.com/helix-editor/helix/pulls?q=is%3Apr+is%3Amerged+add+language and

weakphish:
Any updates on this?

jlebon:
If it helps anyone else, parsing it as Java works OK (definitely not perfect, but better than no highlighting at all). I've put this in my `~/.config/helix/languages.toml`:

```toml
[[language]]
name = "java"
file-types = ["java", "groovy", "Jenkinsfile", "jenkinsfile"]
```

Lonestar137:
According to the Helix documentation [Language Config](https://docs.helix-editor.com/languages.html), adding the following to the languages.toml in your helix config location should provide syntax highlighting but it does not.  
``` toml
[[language]]
name = "groovy"
scope = "source.groovy"
injection-regex = "groovy"
file-types = ["groovy", "Jenkinsfile"]
roots = []

[[grammar]]
name = "groovy"
source = { git = "https://github.com/Decodetalkers/tree-sitter-groovy", rev = "gh-pages" }
```
I checked the runtime and the groovy.so file compiles correctly but it seems like it isn't applied to the buffer.  If I change `name= "groovy"` to `name = "java"` then I get Java syntax highlighting.  I also tried replacing the groovy.so file with other working .so files and no syntax highlighting.  So, in my opinion, this is a bug.

PR #4303, opened 2022-10-16
Add LSP for groovy

zephaniahong:
#4211 
Hi! @the-mikedavis 
I'm trying to implement LSP and syntax highlighting for groovy. The LSP sorta works though it seems a little slow. But my main problem is trying to implement syntax highlighting. I tried copying the entire `highlights.scm` and `injections.scm` from java just to test things out but even that did not work. Any advice?

Thank you!


archseer:
There seems to be a separate grammar in development here? I'm not sure how different groovy is from Java https://github.com/codieboomboom/tree-sitter-groovy

languages.toml
```diff
@@ -1835,3 +1835,17 @@ roots = []
...
 source = { git = "https://github.com/wasm-lsp/tree-sitter-wasm", rev = "2ca28a9f9d709847bf7a3de0942a84e912f59088", subpath = "wast" }
+
+[[language]]
+name = "groovy"
+scope = "source.groovy"
+injection-regex = "groovy"
+file-types = ["groovy", "gvy", "gy", "gsh"]
+comment-token = "//"
+indent = { tab-width = 2, unit = "  " }
+language-server = { command = "java", args = ["-jar", "/Users/zephaniahong/Desktop/groovy-language-server/build/libs/groovy-language-server-all.jar"]}
```
archseer, line 1846: This is hard coded to your setup and shouldn't be upstreamed in this way

zephaniahong: Yup! I'm just testing things out to get a better understanding. But the main issue is that, I can't get the syntax highlighting to work

archseer:
Try using inherits instead in the same way this was done for purescript https://github.com/helix-editor/helix/commit/2d958d6c50b10318e29f6dacc291d8fcafd66de9

zephaniahong:
Tried it but still does not seem to work. Not sure if I'm missing something...

archseer:
If you use tree-sitter-groovy then you need to write queries specific to that implementation and can't just reuse the Java ones.

zephaniahong:
I see! I'll give that a shot.
I've changed the tree sitter back to Java and tried using inherits. Any advice on why that implementation does not work? I checked the logs and it did not produce any error. But there is no syntax highlighting.

The only way I've gotten syntax highlighting to work for groovy is by add the groovy file extension to the java language.

languages.toml
```diff
@@ -1835,3 +1836,23 @@ roots = []
...
+roots = []
+
+[[grammar]]
+name = "groovy"
+source = { git = "https://github.com/tree-sitter/tree-sitter-java", rev = "bd6186c24d5eb13b4623efac9d944dcc095c0dad" }
+
```
the-mikedavis, line 1854: You either need to use `grammar = "java"` in the `[[language]]` block or use a `tree-sitter-groovy` parser - this will fail to load because we're expecting the parsing to expose symbols under the name `tree_sitter_groovy` but this parser uses `tree_sitter_java`: https://github.com/helix-editor/helix/blob/2457111bf074d4b5c1d45f76e2c7d593ceec5895/helix-loader/src/grammar.rs#L76

the-mikedavis: We could probably raise this log level so it shows up in the log by default, as long as this doesn't start showing up in test output: https://github.com/helix-editor/helix/blob/2457111bf074d4b5c1d45f76e2c7d593ceec5895/helix-core/src/syntax.rs#L371

the-mikedavis:
I don't think groovy is close enough to java syntax-wise to use tree-sitter-java

languages.toml
zephaniahong: I see! I think I understand now(: Thanks!
Also, what do you mean by 'as long as this doesn't start showing up in test output'? Would changing it to `log::warning!("{}", e)` solve what you're saying?

zephaniahong:
Yupp! But I just wanted to play around to make sure I understood the mechanics of how things were working. Once I wrap my head around it, I'll swap it out with tree-sitter-groovy and write queries for it.

languages.toml
the-mikedavis: Looking at the test suite, it isn't actually a problem. The integration tests will print log lines but there isn't an integration test where this returns Error

https://github.com/helix-editor/helix/pull/4315

Issue #2719, opened 2022-06-08
It would be nice to have a `trailing` option for rendering spaces and tabs


dimus:
#### Describe your feature request

Additional setting for `[editor.whitespace.render]`

Currently spaces and tabs have 2 possible settings in config: `space = "none"` or `space = "all"` and `tab = "none"` or `tab = "all"`.
I think it woud be beneficial also to have a setting `space = "trailing"` and `tab = "trailing" to visualize only trailing white spaces/tabs

cinerea0:
If this was to be added, I think it would also be useful to have a "leading" option for people who use leading whitespace to keep track of which level of indentation they're in.

n0s4:
Also perhaps a specifier at the `[editor.whitespace.render]` level for these options:
```toml
[editor.whitespace.render]
"trailing"
# "leading"
```

dimus:
> If this was to be added, I think it would also be useful to have a "leading" option for people who use leading whitespace to keep track of which level of indentation they're in.

how to deal with different cardinality of settngs in this case? "all" and "none" can happen only by themselves
`leading` and 'trailing` can happen by themselves, or in tandem, but cannot be comboned with 'all' and 'none'?

BearOve:
Also, it would be great if the theme could have different colors for trailing and leading whitespace. For example a faded color for leading and strong for trailing.

the-mikedavis:
I think this can be solved with theming rather than adding options to render only leading/trailing whitespace. Trailing whitespace could be themed `ui.virtual.whitespace.trailing` for example and this could be set to something noticeable on a per-theme basis. With this approach you could view leading/trailing whitespace without having visible whitespace turned on.

dimus:
I do like the idea of theming, howerver would also be great to be able to toggle this setting when trailing/leading spaces do mean something, like in CSV files, or because of a user preference.

dimus:
as @archseer mentioned in chat:

> this would add some overhead to the rendering since you'd need to track exactly where the last non-whitespace char is to determine if a space is trailing or not

Would toggle help here as well, so for people who do not need the feature, or who disabled it for a time-been it does not slow down the performance, and for people who need it, it is available.

d12bb:
Indent guides are one thing, but they don't help to spot mixed indentation. If I have `listchars` set reasonably in Vim, a line starting with spaces looks differently than one starting with tabs, and it should be, as spaces have no place in indentation for me.

[...]

PR #4306, opened 2022-10-16
feat(editor): add support for displaying trailing whitespace

alevinval:
## Context

This patch implements a new feature to support highlighting of trailing whitespace as requested in #2719, a new module named `trailing_whitespace` is added, which exposes a `WhitespaceKind` enum, and a `TrailingWhitespaceTracker` struct. This tracker is provided with the actual display characters for each type of whitespace (space, nbsp, tab) but also accepts the empty_tab string, this is to avoid having to replicate the logic to construct an invisible tab string that can be indexed with the correct width. I think there's room for improvement about that, but perhaps later.

The editor code is updated to allocate a new trailing whitespace tracker, then it tracks the whitespace that are being processed (only when the feature is enabled!), the tracker accumulates all the whitespace that has been processed. Whenever a non-whitespace character is tracked, the tracker is reset. The chunk of tracked whitespace is only consumed whenever a line ending is found. When a line ending is found, force the rendering of the trailing whitespace that has been accumulated on the tracker.

There is some complexity around the fact that the user can have the trailing option enabled for different combinations of whitespace characters, that's why we need to track the setting for each type of whitespace. That's also why we need to default the character from the "visible" one to the "invisible" one in cases where the toggle is disabled (eg. if the user sets _space_ to _all_, and _nbsp_ to _trailing_, the tracker will be enabled, and when the tracked whitespace has to be written, we need to correctly fallback the space character to not be the empty one, but the displayed one).

The current implementation has little to no overhead for users that do not enable the trailing whitespace option because every operation is guarded behind the `enabled` boolean flag of the tracker. For the users that choose to enable it, there's little impact anyway. Sure, there are some extra allocations that happen, but we're running the feature in linear time since we're not doing things like looking ahead characters or looping the whole document again.

### Extra changes

- Fixed bug where _nbsp_ characters would not be rendered if they appeared before the cursor.

## Demo

screen demo


helix-term/src/ui/editor.rs
```diff
@@ -506,15 +521,15 @@ impl EditorView {
...
                     {
-                        &space
+                        space_char
                     } else {
                         " "
                     };
 
                     let nbsp = if whitespace.render.nbsp() == WhitespaceRenderValue::All
-                        && text.len_chars() < end
+                        && !is_trailing_cursor
```
alevinval, line 530: I realised that `text.len_chars() < end` was equivalent to the `is_trailing_cursor` variable, however for the `space` character it was negated, and here it wasn't. I thought this was intended, but when I was recording the demo I found that `nbsp` characters were disappearing when they were on lines previous to the cursor. I believe this was an overlook (which nobody notices because people don't go around using nbsp characters). I've inverted the condition to fix the issue.

alevinval:
I'll close PR since the codebase has been refactored and I don't see how to do it now.

alevinval:
For those interested, I've brought back the feature here:
https://github.com/helix-editor/helix/pull/7215/files

PR #4307, opened 2022-10-16
Added --print-path flag to print common paths

mitsuhiko:
I was trying to debug why stuff was not loading on my windows machine and learned in the process that the paths are quite different. Since I was trying to rewrite the config file for a test automatically I needed to add platform specific logic. That might be something people need to do more commonly so this adds a `--print-path` argument that can be used for scripting.

Example:

```
$ hx --print-path config-file
/Users/mitsuhiko/.config/helix/config.toml
```

Feel free to close this down. This seems valuable to me, but it might be too specific.

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -9,6 +9,7 @@ pub struct Args {
     pub display_version: bool,
     pub health: bool,
     pub health_arg: Option<String>,
+    pub print_path: Option<String>,
     pub load_tutor: bool,
     pub fetch_grammars: bool,
     pub build_grammars: bool,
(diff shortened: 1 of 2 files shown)


the-mikedavis:
These should be printed by `hx --health` and not add a new flag

mitsuhiko:
My use for this was external scripting. Parsing the output of `hx --health` seems brittle.

pascalkuthe:
I am closing this one out as stale. We want to keep the command line very minimal. These days health prints most of these flags. I don't see us adding a dedicated flag. Thank you for contributing!

PR #4313, opened 2022-10-16
Support drawing popup frame

ath3:
Usability improvement: `popup-frame`.

On some terminals / themes the popups are hard to differentiate from the background.
This makes (for example) LSP information hard to read, and makes the daily work using ssh sessions not as good as it could be.

With this change popups and menus can be configured have a border, and is easy to tell them apart from background with every theme and terminal.

Before:
popup_before

After:
popup_after


book/src/configuration.md
```diff
@@ -56,6 +56,7 @@ on unix operating systems.
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
+| `popup-frame` | Draw a frame around popups | `false` |
```
the-mikedavis, line 59: Let's call this "border" instead of "frame"

sudormrfbin:
Drawing borders for popups feel like a good default whereas having a config option for it feels a bit overkill.

the-mikedavis:
I think an option for it seems ok. I prefer the borderless look but I've seen some discussions and matrix messages about wanting a border around popups

helix-term/src/ui/lsp.rs
```diff
@@ -48,8 +48,9 @@ impl SignatureHelp {
 }
 
 impl Component for SignatureHelp {
-    fn render(&mut self, area: Rect, surface: &mut Buffer, cx: &mut Context) {
+    fn render(&mut self, mut area: Rect, surface: &mut Buffer, cx: &mut Context) {
         let margin = Margin::horizontal(1);
+        area.height -= 1;
```
the-mikedavis, line 53: Why the change here?

the-mikedavis: The `clip_bottom` should probably be gated on the new config option as well - there's a noticeable blank line when the option is off

ath3: I thought the lack of bottom margin was unintentional since other popups have margins around them, and here we only miss it from the bottom.
I changed this now, so without borders if should look as it originally was.

zjp-CN:
I want this to be default. I use `dracula` theme, and in order to distinguish the popup, I have to modify the theme:

image
image


And I haven't seen any popup in nvim without borders. The same codeactionin nvim with [lspsage.nvim](https://github.com/glepnir/lspsaga.nvim) plugin:
image

the-mikedavis:
This change covers popups (like signature-help) but not menus (completion, code-actions). I think an option to enable/disable borders on both would be good

ath3:
> This change covers popups (like signature-help) but not menus (completion, code-actions). I think an option to enable/disable borders on both would be good

It covers now both menus and popups, with settings to enable one of them, both, or none.

the-mikedavis (dismissed):

ath3:
Updated after the changes to popups having scrollbars.
Also fixed the unnecessary leading newlines in popups displayed for the selected menu item.

ath3:
I decided to take out the changes related to selection popup whitespace since it has nothing to do with borders.
I will make a separate pr for that.

ath3:
I think this pr is ready for merge.
The last force-push was only to fix a merge conflict with newer changes on master.

helix-term/src/ui/menu.rs
```diff
@@ -300,14 +303,24 @@ impl<T: Item + 'static> Component for Menu<T> {
...
         surface.clear_with(area, style);
 
+        let border_config = &cx.editor.config().popup_border;
+
+        let render_borders =
+            border_config == &PopupBorderConfig::All || border_config == &PopupBorderConfig::Menu;
+
+        if render_borders {
+            Widget::render(Block::default().borders(Borders::ALL), area, surface);
+            area = area.clip_top(1).clip_bottom(1);
```
archseer, line 321: area.inner doesn't work with Borders?

ath3: It does, i changed it to use area.inner now.

ath3: I removed the mut from area in function parameter and instead create a new variable "inner" to make it look more consistent with popup render code.
If you prefer the previous version i can revert this change.

SoraTenshi:
Nice job on the PR, has been a staple in my fork for like ever.
Minor bug that i noticed tho: DAP mode (list all variables - Spc g v) does not properly draw the frame.

ath3:
I cleaned up the code, added popup-borders typable command, and fixed the issues with dap variable popup (thanks @SoraTenshi), and with menu only border setting.

ath3:
I think this PR is from my perspective in a good state and unless requested i dont plan to do any changes to it.
To celebrate this PR's birthday, can i get a review please?

ath3:
@the-mikedavis @archseer

helix-term/src/ui/menu.rs
```diff
@@ -362,15 +378,19 @@ impl<T: Item + 'static> Component for Menu<T> {
...
+            if let Some(cursor) = self.cursor {
+                let offset_from_top = cursor - scroll;
+                let left = &mut surface[(area.left(), area.y + offset_from_top as u16)];
+                left.set_style(selected);
+                let right = &mut surface[(
+                    area.right().saturating_sub(1),
+                    area.y + offset_from_top as u16,
+                )];
+                right.set_style(selected);
+            }
```
dead10ck, line 393: Apologies, since I'm not super familiar with the rendering code, so I could be misunderstanding something, but this looks like if we want to draw borders around popups that we are deciding not to render the cursor? Why is that?

helix-term/src/ui/popup.rs
```diff
@@ -252,13 +255,29 @@ impl<T: Component> Component for Popup<T> {
...
+        }
+
         self.contents.render(inner, surface, cx);
 
         // render scrollbar if contents do not fit
         if self.has_scrollbar {
-            let win_height = inner.height as usize;
-            let len = self.child_size.1 as usize;
+            let win_height = inner.height as usize - 2 * border;
+            let len = self.child_size.1 as usize - 2 * border;
```
dead10ck, line 280: Should we do a `saturating_sub` here?

helix-view/src/editor.rs
```diff
@@ -950,6 +961,8 @@ pub struct Editor {
     /// field is set and any old requests are automatically
     /// canceled as a result
     pub completion_request_handle: Option<oneshot::Sender<()>>,
+    pub popup_border: bool,
+    pub menu_border: bool,
```
dead10ck, line 965: Could these just be helper methods instead of new fields on the `Editor` object?

helix-term/src/commands/typed.rs
```diff
@@ -574,6 +574,46 @@ fn set_line_ending(
...
+        cx.editor.popup_border = borders.0;
+        cx.editor.menu_border = borders.1;
+    } else {
+        cx.editor
+            .set_status("Valid options are: none, popup, menu, all");
+    };
+
+    Ok(())
+}
+
```
dead10ck, line 616: Does this need its own command? Wouldn't just doing `:set popup-border <choice>` work?

dead10ck (commented):
Thanks for your patience on this PR! This is something I'd like to use myself; we just only have so much time to review all the PRs.

helix-term/src/ui/menu.rs
ath3: This code colors the first and last cell of the selected line in menu.
With borders turned on, the colors would run through the borders.

helix-term/src/ui/popup.rs
ath3: Good catch

helix-view/src/editor.rs
ath3: Yes, i changed them

helix-term/src/commands/typed.rs
ath3: Yeah, i deleted this command now

dead10ck (approved):
Ok this looks good to me! I tried it out locally too, and it looks good. Thanks for the PR.

ath3:
Bump

helix-term/src/ui/popup.rs
```diff
@@ -252,13 +255,29 @@ impl<T: Component> Component for Popup<T> {
         let background = cx.editor.theme.get("ui.popup");
         surface.clear_with(area, background);
 
-        let inner = area.inner(&self.margin);
+        let render_borders = cx.editor.popup_border();
+
+        let inner = if self
+            .contents
+            .type_name()
+            .starts_with("helix_term::ui::menu::Menu")
```
the-mikedavis, line 263: This seems very hacky to me but I don't see a good way around it - there isn't a way to compare the type to `Menu<_>` without dealing with strings like this. I think I'm fine with this for now but it's something we should try to eliminate in the future (though it may take a large refactor to the compositor system to do it)

the-mikedavis (approved):
This looks good to me but I'll leave it up to @archseer to merge

archseer (approved):

nabinkhanal00:
How to enable popup borders on 23.04?

PR #4316, opened 2022-10-16
Correctly handle escaping in completion

### LINKED ISSUE #4266: Open auto completion does not escape spaces
### Summary

When a path contains spaces, the `:open` command completion sets it as is (without escaping the spaces), which leads to incorrect path.

### Reproduction Steps

In a shell:
```
$ mkdir -p '/tmp/a b'
$ echo hi > '/tmp/a b/c'
```

In hx:

`:open /tmp/a`\<TAB\>

`hx` autocompletes to `/tmp/a b`. 
Hitting then \<TAB\> does not autocomplete to  '/tmp/a b/c', typing `/c`\<ENTER\> makes `hx` open two files `/tmp/a` and `b/c`.

I expected this to happen:
autocomplete as `/tmp/a\ b` like bash & co do.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.08.1 (66276ce6)

mitsuhiko:
This fixes two issues with spaces handling in file names for auto completion:

1. the code so far did not use `shellworlds::shellwords` for handling completion. This is now corrected so that the logic for completion matches that of parsing
2. when auto completing it did not correctly apply escaping which is now happening.

Fixes #4266 on unix.

This however does not really fix the issue on Windows. This is because `shellwords` behaves differently on windows. I'm not sure how to properly fix this, but basically it requires double quoting there and from what I can tell it does not support escapes there. It might be that it results in a better user experience if the behavior actually changes there? I'm not sure. At the very least the code requires double quotes there and the completion code does not properly work with double quotes at all yet.

### DIFF
--- a/helix-core/src/shellwords.rs
+++ b/helix-core/src/shellwords.rs
@@ -1,5 +1,22 @@
 use std::borrow::Cow;
 
+/// Auto escape for shellwords usage.
+pub fn escape(input: &str) -> Cow<'_, str> {
+    if !input.chars().any(|x| x.is_ascii_whitespace()) {
+        Cow::Borrowed(input)
+    } else if cfg!(unix) {
+        Cow::Owned(input.chars().fold(String::new(), |mut buf, c| {
+            if c.is_ascii_whitespace() {
+                buf.push('\\');
+            }
+            buf.push(c);
+            buf
+        }))
+    } else {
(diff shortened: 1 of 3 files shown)


archseer (approved):
Thanks!

PR #4317, opened 2022-10-16
Theme: Papercolor: Fixed cursorline background

mehedir137:
Fixed cursorline background color in papercolor-dark theme. Previously the cursorline background color was `#d0d0d0` which does not match the theme, so it is changed to `#303030`


the-mikedavis:
\cc @uncomfyhalomacro

the-mikedavis (approved):

the-mikedavis:
Thanks!

uncomfyhalomacro:
Forgot to reply to this. Too busy these days. Thanks again!

Issue #1472, opened 2022-01-10
Open URL under cursor


matoous:
#### Describe your feature request

Add functionality similar to vim's `gx` that opens URL under cursor. I am still unsure whether this should be included with Helix or better left for a plugin (cc @archseer)?

This can be further extended to open files, commits, etc.

archseer:
I think we should support the hyperlink spec which makes terminal links clickable: https://gist.github.com/egmontkob/eb114294efbcd5adb1944c9f3cb5feda

crossterm doesn't seem to support this, but termwiz does ([I've been hoping to switch over](https://github.com/helix-editor/helix/tree/termwiz) since it also supports bracketed paste and better render diffing): https://docs.rs/termwiz/latest/termwiz/hyperlink/index.html

Command-wise, we already support opening the file under cursor via `gf`: https://github.com/helix-editor/helix/pull/1102

So something like `gx` could be doable if we amend the detection a bit. We'd shell out to `$BROWSER`

d12bb:
> So something like `gx` could be doable if we amend the detection a bit. We'd shell out to `$BROWSER`

Imho better than using `$BROWSER` for `gx`, which often isn't set, would be using the `open` (macOS) or `xdg-open` (Linux) commands. These are 1. using the user's default browser set in his/her desktop environment, and 2. apply to more than just websites, one could also open files, or different URL protocols (`ftp:`, `magnet:` etc). Only thing left to Helix internal would be finding begin and end of the path or URL string.

ubitux:
> > So something like `gx` could be doable if we amend the detection a bit. We'd shell out to `$BROWSER`
> 
> Imho better than using `$BROWSER` for `gx`, which often isn't set, would be using the `open` (macOS) or `xdg-open` (Linux) commands. These are 1. using the user's default browser set in his/her desktop environment, and 2. apply to more than just websites, one could also open files, or different URL protocols (`ftp:`, `mailto:` etc). Only thing left to Helix internal would be finding begin and end of the path or URL string.

Note that the protocol might not be specified and would imply `file://` (think `![img](path/to/image)` in markdown)

edrex:
As it stands this issue is about two subtly-different things, both of which seem worth doing: 
1. implement an internal open-url-under-cursor mechanism.
2. implement the hyperlink spec for detected URLs to support clickable terminal links

Since 2 depends on moving to / deciding about termwiz, and given 1 would make 2 much easier, maybe we should do an internal implementation first?

[...]

PR #4318, opened 2022-10-16
Allow the editor area to be bigger than u16 (width and height remain u16)

A-Walrus:
Now the editor can fill **very** large terminals.
Changed/removed tests which check the truncating behavior.

Closes #4283

### DIFF
--- a/helix-view/src/graphics.rs
+++ b/helix-view/src/graphics.rs
@@ -95,31 +95,19 @@ pub struct Rect {
 }
 
 impl Rect {
-    /// Creates a new rect, with width and height limited to keep the area under max u16.
-    /// If clipped, aspect ratio will be preserved.
+    /// Creates a new rect, with width and height
     pub fn new(x: u16, y: u16, width: u16, height: u16) -> Rect {
-        let max_area = u16::max_value();
-        let (clipped_width, clipped_height) =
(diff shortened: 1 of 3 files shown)


helix-view/src/graphics.rs
```diff
@@ -95,31 +95,19 @@ pub struct Rect {
 }
 
 impl Rect {
-    /// Creates a new rect, with width and height limited to keep the area under max u16.
-    /// If clipped, aspect ratio will be preserved.
+    /// Creates a new rect, with width and height
     pub fn new(x: u16, y: u16, width: u16, height: u16) -> Rect {
```
the-mikedavis, line 99: Does Rect need these fields to become `usize`s so that Rects can be drawn on very large terminal sizes?

A-Walrus: I don't think so. 65 thousand characters as a width or height is already insanely large. Also `crossterm` only works with `u16` for `x/y` positions.
Looking at `termwiz` ([which I understand we might switch to eventually](https://github.com/helix-editor/helix/issues/1472#issuecomment-1008644709)), they do use `usize`.
I just made sure that it was fine if the `area` which equals `width * height` is bigger than `2^16`, which could plausibly happen with a terminal around `256 by 256`.

the-mikedavis (approved):
I turned my font size waaaay down and this works great 👍

PR #4319, opened 2022-10-16
Fix debug assertion for diagnostic sort order

the-mikedavis:
The debug assertion that document diagnostics are sorted incorrectly panics for cases like `[161..164, 162..162]`. The merging behavior in the following lines that relies on the assertion only needs the input ranges to be sorted by `range.start`, so this change simplifies the assertion to only catch violations of that assumption.

See https://github.com/helix-editor/helix/pull/4113#issuecomment-1279997870


Philipp-M (approved):
LGTM.

I couldn't reproduce the issue anymore, I had before.

sudormrfbin (approved):

PR #4322, opened 2022-10-16
Add command to add word boundaries to search

mitsuhiko:
This adds a command to rewrite the search register to add word boundaries to the regex that's there. This is an alternative to #4222.

```toml
[keys.normal]
"*" = ["move_prev_word_start", "move_next_word_end", "search_selection", "make_search_word_bounded", "search_next"]
```

This seems more complex but has the added benefit that it can be used independently. So for instance if one is already down the search of something and discovers that it's not specific enough, a hotkey can be hit to automatically add word boundaries and resubmit the search:

```toml
[keys.normal]
A-s = ["make_search_word_bounded", "search_next"]
```

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -247,6 +247,7 @@ impl MappableCommand {
         extend_search_next, "Add next search match to selection",
         extend_search_prev, "Add previous search match to selection",
         search_selection, "Use current selection as search pattern",
+        make_search_word_bounded, "Modify current search to make it word bounded",
         global_search, "Global search in workspace folder",
         extend_line, "Select current line, if already selected, extend to another line based on the anchor",
         extend_line_below, "Select current line, if already selected, extend to next line",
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -1791,6 +1792,29 @@ fn search_selection(cx: &mut Context) {
     cx.editor.set_status(msg);
 }
 
+fn make_search_word_bounded(cx: &mut Context) {
```
the-mikedavis, line 1795: I think this should operate on the currently selected register (cx.register) and can default to `'/'`

```suggestion
fn make_register_word_bounded(cx: &mut Context) {
    let register = cx.register.unwrap_or('/');
```

helix-term/src/commands.rs
```diff
@@ -1791,6 +1792,29 @@ fn search_selection(cx: &mut Context) {
...
+    let regex = match cx.editor.registers.last('/') {
+        Some(regex) => regex,
+        None => return,
+    };
```
the-mikedavis, line 1799: I think the way that `search_selection` currently works (by pushing a new value to the end of the register) is a bit odd and might be refactored in #4504 

Instead I think this function should pull out the values of the register (`cx.editor.registers.read(register)`, add word boundaries to all of them and then write them back `cx.editor.registers.write(register, new_values)`. That will make it harder to say what the new regex is in the statusline but I think we can say something simpler, maybe `"word boundaries added to register '{}'"` is sufficient

helix-term/src/commands.rs
```diff
@@ -1791,6 +1792,29 @@ fn search_selection(cx: &mut Context) {
...
+    let mut new_regex = String::new();
```
the-mikedavis, line 1800:
```suggestion
    let mut new_regex = String::with_capacity(regex.capacity() + 2);
```

the-mikedavis (commented):
I like the idea of this command 👍: I use a lot of word boundaries when LSP references/symbols don't work (across language boundaries or in docs for example), so I will probably bind this something in my own config. ~I left some comments for making it more generic.~

the-mikedavis (approved):

PR #4324, opened 2022-10-17
Render diagnostics in the file picker preview

the-mikedavis:
This is mostly for the sake of the diagnostics pickers: without rendering the diagnostic styles, it's hard to tell where the entries in the picker are pointing to.


Philipp-M (approved):
Nice, looks good.

PR #4329, opened 2022-10-17
add LSP config for Move language

geometryolife:


languages.toml
```diff
@@ -1835,3 +1835,17 @@ roots = []
...
+injection-regex = "move"
+file-types = ["move"]
+roots = ["Move.toml"]
+comment-token = "//"
+language-server = { command = "move-analyzer" }
+indent = { tab-width = 4, unit = "    " }
+
+[[grammar]]
+name = "move"
+source = { git = "https://github.com/move-hub/tree-sitter-move", rev = "93527b2c4845293384759e8ebee8fdd8a0878ad8" }
```
archseer, line 1851: If you're adding the grammar then queries also need to be added to `runtime/`. Otherwise the grammar isn't really doing anything

geometryolife:
I have a question. I configured move-analyzer, but I can only complete the keywords and can't work well. What needs to be added to work like these language servers such as rust?
image

archseer:
We just offer whatever completion the LSP returns. It could be that the LSP needs more work since Move is still a new language.

sudormrfbin:
The completion logic [seems to be rudimentary right now][1], so I suppose this is expected.

[1]: https://github.com/move-language/move/blob/dfc34400004fa91bdd4bcaf40e3ac8156f512891/language/move-analyzer/src/completion.rs#L148

geometryolife:
@archseer Do you think I need to format the `languages.toml` file? If it is not needed, then I will go back.

the-mikedavis:
I don't think it's necessary to format the languages.toml. It will just confuse the git-blame

pascalkuthe:

PR #4334, opened 2022-10-17
Use the same `WalkBuilder` configuration for the global search file picker as the default file picker

Philipp-M:
Previously the `.git` dir was searched with the global search, I don't think that's intended (as the default file picker does ignore it), so I synced the `WalkBuilder` configs.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1865,10 +1865,15 @@ fn global_search(cx: &mut Context) {
                     .hidden(file_picker_config.hidden)
                     .parents(file_picker_config.parents)
                     .ignore(file_picker_config.ignore)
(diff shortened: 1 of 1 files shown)


pppKin:
> Previously the .git dir was searched with the global search

Cannot reproduce this. Can you open an issue and maybe provide some more details?

Philipp-M:
> > Previously the .git dir was searched with the global search
> 
> Cannot reproduce this. Can you open an issue and maybe provide some more details?

I have git rerere enabled (if you haven't really recommend, helps with rebasing/merging), it often finds stuff in the cache of it.

sudormrfbin:
I can reproduce this; seems like a good change 👍🏾.

sudormrfbin (approved):

PR #4356, opened 2022-10-18
Upgrade rescript tree sitter & highlights

jaredramirez:
Hey there, this PR updates the rescript tree sitter and it's corresponding queries!


runtime/queries/rescript/highlights.scm
```diff
@@ -10,20 +10,28 @@
...
-  "list"
 ] @type
 
+
+(list "list{" @type)
+(list_pattern "list{" @type)
+
+; To ensure that the closing curly bracket is the same color (scope) as the opening curly bracket
+(list "}" @type (#set! "priority" 105))
+(list_pattern "}" @type (#set! "priority" 105))
```
the-mikedavis, line 21: We don't implement a `priority` property so the `(#set! ..)`s here can be removed. These blocks can also be combined as long as `list` and `list_pattern` always have `}` nodes:

```scm
(list ["list{" "}"] @type)
(list_pattern ["list{" "}"] @type)
```

runtime/queries/rescript/highlights.scm
```diff
@@ -74,47 +84,65 @@
  "@"
  "@@"
  (decorator_identifier)
-] @label
+] @keyword.directive
 
 (extension_identifier) @keyword
-("%") @keyword
+("%") @keyword.
```
the-mikedavis, line 90: Looks like this may be a typo?

runtime/queries/rescript/textobjects.scm
```diff
@@ -1,16 +1,114 @@
+; Queries for nvim-treesitter/nvim-treesitter-textobjects
```
the-mikedavis, line 1: This line could be removed or could point to the upstream source for these queries if there is one

runtime/queries/rescript/textobjects.scm
```diff
@@ -1,16 +1,114 @@
...
+; Calls
+;------
+
+(call_expression arguments: ((_) @parameter.inside)) @parameter.around
+
 ; Comments
 ;---------
 
-(comment) @comment.inside
+(comment) @comment.around
```
the-mikedavis, line 27:
```suggestion
(comment) @comment.inside
(comment)+ @comment.around
```

jaredramirez:
Thanks for the comments, I've committed the updates!

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4357, opened 2022-10-19
Fix link to textobjects usage from keymap documentation

phillipspc:
Minor thing I figured I'd fix after coming across it: the anchor "#textobject" wasn't correctly linking to the section in Usage "#textobjects".


the-mikedavis (approved):
Nice catch, thanks!

PR #4362, opened 2022-10-19
Remove lower-case requirement for language/grammar

### LINKED ISSUE #4346: Inconsistent/undocumented lowercase usage for grammars/languages
### Summary

I tried to integrate a [grammar for BitBake](https://github.com/nateglims/tree-sitter-bb) files into helix. This grammar uses `BB` as its name. While a bit odd, I don't think its prohibited to use uppercase letters for the grammar name.

However this proves to be difficult to use with helix, as it will use `to_ascii_lowercase()` for some operations. Furthermore this is not even consistent, so declaring a grammar with `name = "BB"` and building it will produce a `BB.so` but loading fails because it can not find a `bb.so`. (I know that even if that would have worked it still would not find the `tree_sitter_<name>` symbol, as it is also lowercased.)

Note: when changing the tree-sitter grammar to use a lowercase name, and adjusting all my configs accordingly it works fine.

Is this forced lowercase there for a reason? Then I think this should be both documented and at least `hx --grammar build` should be fixed.

In my opinion, helix should just use the case supplied.

### Reproduction Steps

Put the following into `languages.toml`:
```toml
[[language]]
name = "BB"
scope = "source.BB"
file-types = ["bb", "bbclass", "bbappend", "inc"]
roots = []
grammar = "BB"
comment-token = "#"

[[grammar]]
name = "BB"
source = { git = "https://github.com/nateglims/tree-sitter-bb.git", rev = "d8c986da8cd043551e3084280c2f3c1edd6e49c8" }
```

Place the [queries](https://github.com/nateglims/tree-sitter-bb/tree/main/queries) into `runtime/grammars/BB/`.

Fetch and build `hx --grammar fetch && hx --grammar build`

Open a BitBake file and be happy. (Example file to be found [here](https://github.com/openembedded/bitbake/blob/master/classes/base.bbclass)

### Platform

Linux

### Terminal Emulator

alacritty 0.10.1

### Helix Version

helix 22.08.1 (9b7f349f)

kleinweby:
As described in #4346 both language and grammar names are sometimes lower-cased. For grammars this is incompatible with ones using uppercase names. And for languages this can be a bit surprising. (At least I stumbled across helix ignoring my queries because of this).

Note that the syntax part is  not strictly necessary to get grammars as described in the issue to work.

Closes #4346

### DIFF
--- a/xtask/src/querycheck.rs
+++ b/xtask/src/querycheck.rs
@@ -14,8 +14,8 @@ pub fn query_check() -> Result<(), DynError> {
     ];
 
     for language in lang_config().language {
-        let language_name = language.language_id.to_ascii_lowercase();
-        let grammar_name = language.grammar.unwrap_or(language.language_id);
+        let language_name = &language.language_id;
+        let grammar_name = language.grammar.as_ref().unwrap_or(language_name);
         for query_file in query_files {
             let language = get_language(&grammar_name);
             let query_text = read_query(&language_name, query_file);
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -361,14 +361,12 @@ pub fn read_query(language: &str, filename: &str) -> String {
 
 impl LanguageConfiguration {
     fn initialize_highlight(&self, scopes: &[String]) -> Option<Arc<HighlightConfiguration>> {
-        let language = self.language_id.to_ascii_lowercase();
-
-        let highlights_query = read_query(&language, "highlights.scm");
+        let highlights_query = read_query(&self.language_id, "highlights.scm");
         // always highlight syntax errors
         // highlights_query += "\n(ERROR) @error";
 
-        let injections_query = read_query(&language, "injections.scm");
-        let locals_query = read_query(&language, "locals.scm");
+        let injections_query = read_query(&self.language_id, "injections.scm");
+        let locals_query = read_query(&self.language_id, "locals.scm");
 
         if highlights_query.is_empty() {
             None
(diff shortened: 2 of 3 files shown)


the-mikedavis:
One more to remove here: https://github.com/helix-editor/helix/blob/418a622db9d957d09e10ccbc3897d8ac9268dc8e/xtask/src/querycheck.rs#L17

kleinweby:
Ah thanks. I've removed the missing to lowercase.

Two of the tests actually used this lowercase conversion. I've simply changed them. I've also checked `language.toml` that every language/grammar currently used lower-case names, which is the case.

xtask/src/querycheck.rs
```diff
@@ -14,17 +14,16 @@ pub fn query_check() -> Result<(), DynError> {
     ];
 
     for language in lang_config().language {
-        let language_name = language.language_id.to_ascii_lowercase();
         let grammar_name = language.grammar.unwrap_or(language.language_id);
         for query_file in query_files {
             let language = get_language(&grammar_name);
-            let query_text = read_query(&language_name, query_file);
+            let query_text = read_query(&language.language_id, query_file);
```
the-mikedavis, line 20: `language` gets shadowed above which is causing the compilation to fail

kleinweby: Thanks, sorry I missed that :)

the-mikedavis:
I missed this in the review of the PR that added it but there's a real test failure now from the uppercasing on this line: https://github.com/helix-editor/helix/blob/e25af1f7441fd1eccae580ba2e8e0eebc2be74f8/helix-core/src/syntax.rs#L2278 - could you set that to lowercase?

kleinweby:
Sure, I rebased the PR and lowercased the failing test.

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #4342, opened 2022-10-18
 Rust project: file not included in module tree


zjp-CN:
I'm reading cross-term's example [interactive-demo](https://github.com/crossterm-rs/crossterm/tree/master/examples/interactive-demo). And I open the source file with helix (master version), and get a single error from rust-analyzer `file not included in module tree` on the first line. 
```shell
git clone https://github.com/crossterm-rs/crossterm.git
cd examples/interactive-demo
hx src/main.rs
```

Hover doc, autocompletion and goto-definition are not working. But you can see they work in neovim (under `examples/interactive-demo` too):

https://user-images.githubusercontent.com/25300418/196368208-9a35451d-b85b-4b9a-9e71-1230b69c063b.mp4

the-mikedavis:
This is the expected behavior of the current workspace detection - the crossterm repo's root is the highest directory in the git repository that has a root marker (Cargo.toml). Rust-analyzer thinks that the module isn't included because it isn't included in crossterm (which is correct) rather than the interactive-demo project. Also see https://github.com/helix-editor/helix/issues/3993, https://github.com/helix-editor/helix/issues/2732, https://github.com/helix-editor/helix/issues/2970

gabydd:
@pascalkuthe do you think this and related issues are reasonably solved https://github.com/helix-editor/helix/pull/5748

pascalkuthe:
yeah that's fixed by that PR specifying `editor.workspace-lsp-roots = ["examples/interactive-demo"]` in a `.helix/config.toml` file inside the crossterm repo works. There is no situation to autodetect stuff like this and exactly why we ended up with the implementation in https://github.com/helix-editor/helix/pull/5748

I forgot to link this issue in that PR, thanks :+1:

PR #4363, opened 2022-10-19
feat(find_root): add `$ROOT` environment variable detection

zjp-CN:
fix https://github.com/helix-editor/helix/issues/4342

You can use `ROOT=$PWD hx src/main.rs` to specify the project root if helix is failed to detect the unusual project structure.

### DIFF
--- a/helix-core/src/lib.rs
+++ b/helix-core/src/lib.rs
@@ -42,12 +42,17 @@ pub fn find_first_non_whitespace_char(line: RopeSlice) -> Option<usize> {
 /// Find project root.
 ///
 /// Order of detection:
+/// * `ROOT` environment variable
 /// * Top-most folder containing a root marker in current git repository
 /// * Git repository root if no marker detected
 /// * Top-most folder containing a root marker if not git repository detected
 /// * Current working directory as fallback
 pub fn find_root(root: Option<&str>, root_markers: &[String]) -> std::path::PathBuf {
-    let current_dir = std::env::current_dir().expect("unable to determine current directory");
+    use std::env;
+    if let Ok(path) = env::var("ROOT") {
+        return path.into();
+    }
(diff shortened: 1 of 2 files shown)


zjp-CN:
This allows helix to work under any path, like
`ROOT=$PWD/interactive-demo hx interactive-demo/src/main.rs` or
`ROOT=/rust/github/crossterm/examples/interactive-demo hx /rust/github/crossterm/examples/interactive-demo/src/main.rs`.

I think the root detection rule need to be documented in the book. I can do it.

TotalKrill:
This would also be a very nice CLI flag tbh, because I believe a common user pattern with CLI software is to run: 

<software> --help

and then understand some capabilities, I advocated earlier in element ( Matrix ) for having a Clap based CLI options, and this would be one of those usecases, where clap could give both discoverability, and some documentation of the feature as in a standardised way

pascalkuthe:
we ended up solving this a different way with https://github.com/helix-editor/helix/pull/5748 with a project local config file instead (so you don't have to specify it on each startup). The implemented solution also allow multiple different roots for a single running helix instance (so you can still open the normal crossterm files too in your example) .Forcing all lsps to the same root with a single environment variable tends to also not work for mulpi language projects so I don't think we will be going forward with this.

univerz:
> we ended up solving this a different way with #5748 with a project local config file instead

it would be great if it was possible to use it from commandline too. a typical use case is to quickly explore a few `examples/foo` in a big unknown project. i can do `hx .`, but it affects filepicker only once untill lsp starts.

pascalkuthe:
We usually don't add CLI options. There isn't any CLI flrags to set config options either. You can fairly quickly open the local config file (there is even a command for it), edit the config and then reload.

This is also not about the file picker. The time picker is not affected by LSO root detection. There is already commands to open the file picker in theCWD and current buffer dir. That should be sufficient for that usecase

Issue #1248, opened 2021-12-10
Hover page up down hint


pickfire:
Currently when hover, there is a box. It can be scrolled down but there are two issues

- it is not cleared when it can be scroll up or down
- it is not discoverable for new users that ctrl-u and ctrl-d is used to scroll the hover box

One idea I have is that we draw `ctrl-u` text if it can be scrolled up and `ctrl-d` text if it can be scrolled down along the border of the box, that way we can reuse the existing border space while making it clear to users when it can be scrolled as well as being beginner friendly. Example,

```
+--------------------+
| ...                |
+------------ ctrl-d +

archseer:
We can just render a scrollbar, like we do for the menu.

pickfire:
We can render a scrollbar but user is likely to not know how to scroll up and down. Not sure about the rest but I don't know about the key to scroll until I saw someone mentioned it.

archseer:
> We can render a scrollbar but user is likely to not know how to scroll up and down. Not sure about the rest but I don't know about the key to scroll until I saw someone mentioned it.

I think it's predictable since we also use those keys in the picker, but we don't point them out anywhere except in the docs. Or how `i` enters insert mode, but there's no documentation text on screen in normal mode to point that out :P 

A scrollbar would make it at least a bit more obvious that scrolling is possible.


> We could show a line `Ctrl-d to scroll down, Ctrl-u to scroll up` in the status message area (similar to `Recording to {register}` message shown when recording a macro).

Yeah that could work

Jlll1:
> > We could show a line `Ctrl-d to scroll down, Ctrl-u to scroll up` in the status message area (similar to `Recording to {register}` message shown when recording a macro).
> 
> Yeah that could work

So, do we want to do that? I think a scrollbar would be sufficient.

sheyll:
> > > We could show a line `Ctrl-d to scroll down, Ctrl-u to scroll up` in the status message area (similar to `Recording to {register}` message shown when recording a macro).
> > 
> > 
> > Yeah that could work
> 
> So, do we want to do that? I think a scrollbar would be sufficient.

As a new Helix user I would really prefer at least the status area message.

But broadening the scope a bit: Helix seems to have this UX idiom of showing pop-ups guiding the user.

For example when pressing `Space` a pop-up appears with what keys a user can now press to complete a command sequence. 

[...]

PR #4364, opened 2022-10-19
Ctrl-d/Ctrl-u to scroll the option's documentation on auto-complete

### LINKED ISSUE #3062: ctrl-d ctrl-u not scrolling popup documentation
### Summary

When popuping documentation in helix editor using rust plugin, the keys ctrl-d/ctrl-u doesn't scroll the popup doc window.
image

### Reproduction Steps

Type code until the popup documentation appear. If has more documentation hidden, try scroll down using ctrl-d(it will not scroll).

### Helix log

_No response_

### Platform

Linux, popos 22.04

### Terminal Emulator

alacritty default gnome term

### Helix Version

helix 22.05 (27609f50)

Manosmer:
Fixes #3062. The Markdown documentation is now treated as a separate `Box<Popup<Markdown>>`. So, a `Completion` object now has 2 `Popup`s; one for the `Menu`(as before) and a new one for the relevant option's documentation. Currently, the code creates a new `Box<Popup<Markdown>>` only when the user moves to another option. 

Let me know if there is any additional optimisation needed, like: 
- ~~running `Popup::required_size()` inside `Completion::render()` only when the option changes (currently, it runs every time scroll happens).~~


David-Else:
I imagine the same scroll keys should be used as for scrolling in https://github.com/helix-editor/helix/pull/4189

Manosmer:
> I imagine the same scroll keys should be used as for scrolling in https://github.com/helix-editor/helix/pull/4189

Currently, documentation scrolling in popups that appear by doing `<Esc><Space>k` is possible via Ctrl-d/Ctrl-u. I chose the same keybindings to comply with the above.

David-Else:
This is probably another PR, but it would be great if markdown previews of documentation could show if there was more to scroll up or down to.

At the moment, when you see the docs in a popup you don't know if you are seeing the whole thing or need to scroll down to read more. If there were a symbol like `↓` in the bottom right that showed when there was more it would be very useful. There could be an `↑` too that does the same thing in reverse.

the-mikedavis:
Yeah some sort of hint for how much documentation would be a nice feature (though separate from this PR) https://github.com/helix-editor/helix/issues/1248. Maybe a scrollbar is sufficient

RmStorm:
Is there any chance of getting this fix merged? Scrolling inside the auto-complete popup is still broken I think.

pascalkuthe:
closing this one as stale,  #8047 is a newer implementation. Thank you for contributing!

PR #4370, opened 2022-10-19
Trim quotes and braces from paths in goto_file_impl

### LINKED ISSUE #4122: [BUG] Go to file is capturing surrounding single quotes
files
```
index.js
example.js
```

consider a file `example.js` that contains the line 
`import './index.js'` or
`import 'index.js'`

You cannot navigate to `index.js` by using `gf` on either line

What's interesting is you can manually select `index.js` and hit `gf` and it works.

It looks like `gf` is capturing the surrounding single quotes. Neovim does not exhibit that behavior.

So in Helix you have to `mi'gf` and in Neovim you do `gf`

dariooddenino:
`goto_file_impl` now trims surrounding `'`, `"`, `(` and `)` before opening the path.

The `current_word` selection was inverted so that now it works from both the first and last characters of the word.

Fixes #4122

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1036,9 +1036,10 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     // Checks whether there is only one selection with a width of 1
     if selections.len() == 1 && primary.len() == 1 {
         let count = cx.count();
+        let text_slice = text.slice(..);
         // In this case it selects the WORD under the cursor
         let current_word = textobject::textobject_word(
-            text.slice(..),
+            text_slice,
             primary,
             textobject::TextObject::Inside,
             count,
@@ -1048,8 +1049,8 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         let surrounding_chars: &[_] = &['\'', '"', '(', ')'];
         paths.clear();
         paths.push(
-            text.slice(current_word.from()..current_word.to())
-                .to_string()
+            current_word
+                .fragment(text_slice)
                 .trim_matches(surrounding_chars)
                 .to_string(),
         );
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1033,16 +1033,20 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .collect();
     let primary = selections.primary();
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
-        let current_word = movement::move_prev_long_word_start(
-            text.slice(..),
-            movement::move_next_long_word_start(text.slice(..), primary, 1),
-            1,
-        );
-        paths.clear();
-        paths.push(
-            text.slice(current_word.from()..current_word.to())
-                .to_string(),
-        );
+        let count = cx.count();
+        let new_selection = selections.clone().transform(|range| {
+            textobject::textobject_word(
+                text.slice(..),
+                range,
+                textobject::TextObject::Inside,
+                count,
+                true,
+            )
+        });
+        paths = new_selection
+            .iter()
+            .map(|r| text.slice(r.from()..r.to()).to_string())
+            .collect();
     }
     for sel in paths {
         let surrounding_chars: &[_] = &['\'', '"', '(', ')'];
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1033,9 +1033,9 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .collect();
     let primary = selections.primary();
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
-        let current_word = movement::move_next_long_word_start(
+        let current_word = movement::move_prev_long_word_start(
             text.slice(..),
-            movement::move_prev_long_word_start(text.slice(..), primary, 1),
+            movement::move_next_long_word_start(text.slice(..), primary, 1),
             1,
         );
         paths.clear();
@@ -1045,7 +1045,8 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         );
     }
     for sel in paths {
-        let p = sel.trim();
+        let surrounding_chars: &[_] = &['\'', '"', '(', ')'];
+        let p = sel.trim().trim_matches(surrounding_chars);
         if !p.is_empty() {
             if let Err(e) = cx.editor.open(&PathBuf::from(p), action) {
                 cx.editor.set_error(format!("Open file failed: {:?}", e));
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1032,21 +1032,25 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .map(|r| text.slice(r.from()..r.to()).to_string())
         .collect();
     let primary = selections.primary();
-    if selections.len() == 1 && primary.to() - primary.from() == 1 {
+    if primary.len() == 1 {
         let count = cx.count();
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -1031,9 +1031,9 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .collect();
     let primary = selections.primary();
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
-        let current_word = movement::move_next_long_word_start(
+        let current_word = movement::move_prev_long_word_start(
             text.slice(..),
-            movement::move_prev_long_word_start(text.slice(..), primary, 1),
+            movement::move_next_long_word_start(text.slice(..), primary, 1),
```
the-mikedavis, line 1036: With this change I see a regression:

```
helix-term/src/main.rs
import "helix-term/src/main.rs"
```

With the cursor on the the last `s` on the first line, `gf` opens "import" and not the path

dariooddenino: > With this change I see a regression:
> 
> ```
> helix-term/src/main.rs
> import "helix-term/src/main.rs"
> ```
> 
> With the cursor on the the last `s` on the first line, `gf` opens "import" and not the path

I changed approach and it should be working correctly on all cases now.

helix-term/src/commands.rs
```diff
@@ -1031,19 +1031,24 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                range,
+                textobject::TextObject::Inside,
+                count,
+                true,
+            )
+        });
+        paths = new_selection
+            .iter()
+            .map(|r| text.slice(r.from()..r.to()).to_string())
+            .collect();
```
the-mikedavis, line 1047: Instead of iterating on the ranges in the selection, you can use `primary` since we know that it's the only range (because of the `selections.len() == 1` condition

helix-term/src/commands.rs
```diff
@@ -1031,19 +1031,24 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .collect();
     let primary = selections.primary();
     if selections.len() == 1 && primary.to() - primary.from() == 1 {
```
the-mikedavis, line 1033: Unrelated to this change but could this be `primary.len() == 1`? I don't think we had `Range::len` when this block was originally written.

It's the same implementation but it's a bit clearer - `to - from` mistakenly makes me think that we'll run into an issue if the range is reversed (which isn't true actually) 😄

helix-term/src/commands.rs
dariooddenino: Right, thanks! I've changed it, hope it's good now :)

helix-term/src/commands.rs
dariooddenino: > Unrelated to this change but could this be `primary.len() == 1`? I don't think we had `Range::len` when this block was originally written.
> 
> It's the same implementation but it's a bit clearer - `to - from` mistakenly makes me think that we'll run into an issue if the range is reversed (which isn't true actually) smile

I have replaced it with `primary.len() == 1` and it works fine

helix-term/src/commands.rs
```diff
@@ -1030,20 +1030,29 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         .map(|r| text.slice(r.from()..r.to()).to_string())
         .collect();
     let primary = selections.primary();
-    if selections.len() == 1 && primary.to() - primary.from() == 1 {
```
the-mikedavis, line 1033: The condition on `selections.len()` is important here - it's checking that there is only one range (`primary`) in the selection

helix-term/src/commands.rs
```diff
@@ -1030,20 +1030,29 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            .transform(|range| {
+                textobject::textobject_word(
+                    text.slice(..),
+                    range,
+                    textobject::TextObject::Inside,
+                    count,
+                    true,
+                )
+            })
+            .primary();
```
the-mikedavis, line 1046: You can operate on the `primary` range directly here:

```suggestion
        let current_word = textobject::textobject_word(
            text.slice(..),
            primary,
            textobject::TextObject::Inside,
            count,
            true,
        );
```

helix-term/src/commands.rs
```diff
@@ -1030,20 +1030,29 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
         paths.clear();
         paths.push(
             text.slice(current_word.from()..current_word.to())
                 .to_string(),
         );
     }
     for sel in paths {
-        let p = sel.trim();
+        let surrounding_chars: &[_] = &['\'', '"', '(', ')'];
+        let p = sel.trim().trim_matches(surrounding_chars);
```
the-mikedavis, line 1055: I don't think we want to trim in all cases: if we have multiple selections or a selection that covers a range with `range.len() > 1`, `gf` should only act on the contents of that selection and shouldn't trim.

I think we should only be trimming in the `if selection.len() == 1 && primary.len() == 1` branch

helix-term/src/commands.rs
dariooddenino: Ahhn, I understand now! Thanks

helix-term/tests/test/commands.rs
```diff
@@ -133,3 +135,65 @@ async fn test_selection_duplication() -> anyhow::Result<()> {
...
+
+    fn match_paths(app: &Application, matches: Vec<&str>) -> usize {
+        app.editor
+            .documents()
+            .map(|d| d.path())
+            .flatten()
+            .map(|p| p.file_name())
+            .flatten()
+            .filter(|n| matches.iter().any(|m| &OsStr::new(m) == n))
+            .count()
```
the-mikedavis, line 151: Clippy has some complaints about this block:

```suggestion
        app.editor
            .documents()
            .filter_map(|d| d.path()?.file_name())
            .filter(|n| matches.iter().any(|m| *m == n.to_string_lossy()))
            .count()
```

then you can also remove the import of OsStr

the-mikedavis:
Clippy and the auto-formatter (in rust-analyzer) won't work in the integration tests because they're behind a feature flag and look like they aren't included in the module tree until that feature flag is enabled. You can setup a custom language config to enable the integration feature flag and check clippy on save:

```toml
# path/to/helix/.helix/languages.toml
[[language]]
name = "rust"

[language.config]
cargo.features = [ "integration" ]
checkOnSave.command = "clippy"
```

helix-term/src/commands.rs
```diff
@@ -1032,15 +1033,24 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            textobject::TextObject::Inside,
+            count,
+            true,
         );
+        // Trims some surrounding chars so that the actual file is opened.
+        let surrounding_chars: &[_] = &['\'', '"', '(', ')'];
         paths.clear();
         paths.push(
             text.slice(current_word.from()..current_word.to())
+                .to_string()
```
the-mikedavis, line 1052: We can use `current_word.fragment(text)` and then `.trim_matches` on that - it will have a chance of not copying the range's contents until we get to the `.to_string()` at the bottom (calling `.to_string()` will always copy the contents).

(For this to work, we'll need to add a line up on L1038 saying `let text = text.slice(..)` and then we can re-use the ropeslice in the call to `textobject_word`.)

dariooddenino: Thank you very much!

the-mikedavis (approved):
Thanks for working on this!

I pushed a minor style change but this is looking good

PR #4372, opened 2022-10-19
Add support for D

gdamore:


runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+    "."
+    ":"
+    ","
+] @punctuation.delimiter
+
+[
+    (null)
+    (true)
+    (false)
+] @constant.language
```
the-mikedavis, line 168: `(true)` and `(false)` should be tagged as `@constant.builtin.boolean` and `(null)` as `@constant.builtin`

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+
+(special_keyword) @constant.language
```
the-mikedavis, line 170: This could be `@constant.builtin`, `@function.builtin`, or `@special` depending on its use in D

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+    (creal)
+    (double)
+    (cfloat)
+] @type.deprecated
+
+(label (identifier) @label)
+(goto_statement (goto) @keyword (identifier) @label)
+
+(string_literal) @string
+(int_literal) @number
```
the-mikedavis, line 210:
```suggestion
(int_literal) @constant.numeric.integer
```

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+(float_literal) @number
```
the-mikedavis, line 211:
```suggestion
(float_literal) @constant.numeric.float
```

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+(char_literal) @number
```
the-mikedavis, line 212:
```suggestion
(char_literal) @constant.character
```

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+(identifier) @variable
+(at_attribute) @property
```
the-mikedavis, line 214:
```suggestion
(at_attribute) @attribute
```

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+
+; everything after __EOF_ is plain text
+(end_file) @text
```
the-mikedavis, line 217:
```suggestion
(end_file) @ui.text
```

runtime/queries/d/highlights.scm
```diff
@@ -0,0 +1,217 @@
...
+[
+    (cent)
+    (ucent)
+    (ireal)
+    (idouble)
+    (ifloat)
+    (creal)
+    (double)
+    (cfloat)
+] @type.deprecated
```
the-mikedavis, line 204: We don't use a `@type.deprecated` scope (scopes are [here](https://docs.helix-editor.com/master/themes.html#syntax-highlighting)). I think a nice way to handle this case would be to mark the non-deprecated types above as `@type.builtin` and these as `@type`. Alternatively they could be `@warning` or `@error`

runtime/queries/d/highlights.scm
gdamore: These keywords are things like __FILE__ and __LINE__.  They are predefined constants.   Probably constant.builtin, but I'm open to thoughts.

runtime/queries/d/highlights.scm
gdamore: I'll probably mark the builtin things as type.builtin, and these as warning.  Thanks for the suggestions.

runtime/queries/d/highlights.scm
the-mikedavis: `@constant.builtin` seems like the best option then 👍

gdamore: Done.  I also added more fine grained classifications so I'd appreciate it if you could have another look.

runtime/queries/d/highlights.scm
gdamore: And now that's done too.

runtime/queries/d/highlights.scm
gdamore: Done.

gdamore:
Test failure is not in any code I changed. ;-)

the-mikedavis (approved):

the-mikedavis:
Thanks!

Could you add a pointer in the language server wiki for installing serve-d? https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

Issue #4335, opened 2022-10-17
Global search and replace


CBenoit:
Add some form of support for global search-and-replace.

### Discussed in https://github.com/helix-editor/helix/discussions/3793

<div type='discussions-op-text'>

<sup>Originally posted by **nour-dev** September 11, 2022</sup>
hey, the question is the title. 
I know how to make per file search and replace. what I want to do is to search + replace in the entire workspace files (or sub of them by ext or some sort of regex).
for example when I want to search `'test'` I do:
- click `space+/`.
- the search picker show list of files contain the regex.

then when I want to replace it I need to go to each file and replace manually.
so there is not way to replace them all with one or more commands.

is there any hack to do this ? or I need to use external shell tools like `sed` or `sd` ?

thanks.
</div>

Teiolass:
How should this behaviour interact with the undo? Add the change only to the alread open buffers? 

I think this solution is quite dirty, but it is the best that comes to my mind. I think that letting this feature to be handled by external programs would be a better choice.

helmesjo:
One approach I've seen before is a prompt where you'd say if to open all files when replacing (adding to undo buffer) or replace without opening (no undo). The prompt can be handled in many other ways of course, just that it's an explicit choice.
Personally I'd say this is an expected feature in any IDE, so I vote yay. (:

QiBaobin:
Open a buffer for the result, any changes on that buffer would effect related files? This can undo, something like https://github.com/mhayashi1120/Emacs-wgrep

CBenoit:
#4381 sounds exactly like this 🙂

wongjiahau:
What we need is something like Vim’s `cdo`  command, which allows user to apply Ex command to every entry in the quickfix list.

My suggestion, add a simple command call `run-macro-on-last-picker`.

pascalkuthe:
The plan is very much something like https://github.com/helix-editor/helix/pull/4381 that would leverage helixes existing multi-cursor support. That PR just needs a better PR and for that larger changes are likely needed

Rot127:
This is probably out of scope for Helix, but I'd like to drop it anyways here as inspiration.

[serpl](https://github.com/yassinebridi/serpl) is a really nice global search and replace tool. Also written in Rust but probably too complicated to integrate in Helix (especially since it adds yet another dependency).
But something similar fancy in helix would be awesome.

Example of serpl:

[...]

PR #4381, opened 2022-10-20
[WIP] View based refactoring in workspace

pathwave:
This implements a new command: global_refactor, where one inputs a search regex and a new window shows up with all the lines in different files matching the search.

One can then do edits on the lines and afterwards apply those changes to the documents.
If a file was not opened in the editor from before it will be opened and the change added as a transaction.
This means that the refactoring can be undone in their respective documents.
* Demo:

https://user-images.githubusercontent.com/114359560/196932067-dda9118c-dcfe-416d-971a-d8612349ef01.mp4

There is still a lot of work to be done but I opened a draft pr to get some feedback:
* Is this something you want in core?
* What should be the default keybind for this?
* Suggestions / Ideas / Something that should be done different?

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -32,7 +32,7 @@ use helix_core::{
 };
 use helix_view::{
     clipboard::ClipboardType,
-    document::{FormatterError, Mode, SCRATCH_BUFFER_NAME},
+    document::{FormatterError, Mode, REFACTOR_BUFFER_NAME, SCRATCH_BUFFER_NAME},
     editor::{Action, Motion},
     info::Info,
     input::KeyEvent,
@@ -277,6 +277,7 @@ impl MappableCommand {
         search_selection, "Use current selection as search pattern",
         make_search_word_bounded, "Modify current search to make it word bounded",
         global_search, "Global search in workspace folder",
+        global_refactor, "Global refactoring in workspace folder",
         extend_line, "Select current line, if already selected, extend to another line based on the anchor",
         extend_line_below, "Select current line, if already selected, extend to next line",
         extend_line_above, "Select current line, if already selected, extend to previous line",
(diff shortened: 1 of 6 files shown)


zjp-CN:
> Is this something you want in core?

Yes! I tried to find a plugin in neovim for global searched text replacement, but failed. I know some tricks to do it in neovim with built-in commands and popular plugins like telescope (via sending live grep to quickfix window). 

> This means that the refactoring can be undone in their respective documents.

This is the one of the key and important function in my opinion. I can share the experience in nvim:

```vim
:args *.md

:argdo s/pat/replacement/ | update
" or forbid autocmd
:silent! noautocmd argdo %s/pat/replacement/ | update 

" restore contents before substitution
:argdo undo
```

archseer:
Oh cool! This reminds me of Zed's multi-buffers: https://www.youtube.com/watch?v=egI75sLT164

CBenoit:
Very cool! Would also address #4335 beautifully :100:

dahmc:
I love this! Can't wait to have this in core.

pathwave:
I'm not quite happy with using the enum to separate from a normal file-buffer.
I think a better way would be to separate different "UI elements" into their own components and also separate commands for those components, so I might do that in other PR's before continuing on this.

pascalkuthe:
There are potentially more features that would benefit from custom views (like diff-mode) so there is some overlap there.
Also this PR could provide a provide a better UI (I think Zeds multi-buffers are a great reference).
I thank it would be better to abstract over `View` rather than `Document`.
To me a `Document` really represents a single file and a `View` is a display of that `Document` where the user can edit it.
This already applies for splits but also fits diff-mode and global refactoring.

This approach does however mean that we need commands that behave differently based upon what view is visible.
I was thinking of just having a callback called when a command is dispatched. 
`AbstractView::execute_command<T>(&mut self, name: &str, f: impl FnMut(&mut View)  ;
which would allow custom views to either just forward to their sub-views or hook for commands they want to "special case" (for example scrolling/movement in diff-mode should affect both views but all other commands are treated as normal).

The approach is kind of tricky for this PR because you would have to special case selection commands and introduce a notion of a view without selection but that's it. 

Ideally you should not have to touch the document or text rendering/positioning itself at all (especially because #5420 will almost fully replace the old rendering/positioning code). Simply calculate the height and offset of each sub-view upfront (and keep it fixed even if text is removed) and then use a simple row-based offset into the view and use a binary search of views to find which views to render (and which not).

Since #5519 is mostly done now I might work on such an abstraction next in my quest to get diff-mode working :D

pathwave:
Yes, one of the main issues here is that the text edit commands use the current document.
My thoughts was that in a more component based layout one could have a component "text buffer" that had all the commands for editing text in it. Then there could be a "file view" or something that used this component and added its own commands for file specific things. "Global refactor view" could then also use "text buffer" as a sub component and add its own specific things.

When more and more things get added to the editor, for example "file-tree view" it makes sense that those things are separate components with their own commands and can have separate key bindings and we can avoid squashing everything into EditorView

nfurfaro:
Very cool to see this WIP! I was searching to see if there were any plans for something similar to ZEDs multibuffers, and here it is.

agl-alexglopez:
@pascalkuthe Hi! I'm just a new user but I've found the cursor paradigm extremely powerful and unique so far in Helix. When you say...

> This approach does however mean that we need commands that behave differently based upon what view is visible.

and that a `View` may lose some selection capabilities, would you lose some benefit of how flexible cursors are in your standard editing session. For example suppose we open all those buffers we want to refactor as in the video example. If we are given cursors that appear in all of those views we can refactor as expected but implicitly we would also have access to a selection removal similar to (y/n/a/q)? in many text editors. We could simply cycle through primary selection with `(` or `)` and even drop selections we don't want to change with `Alt-,`. This is like your standard editor prompt replace but in the reverse order; decide if you want to drop anything then confirm the change for the rest of the selections. I'm gathering from this thread that might be a technical challenge so no worries if it's not possible. I'm trying to learn about the codebase ASAP so I can leave better feedback in the future. Keep up the great work!

verte-zerg:
Hi! Could someone describe what needs to be done to merge this PR? Maybe some of the authors can share a vision of how the feature should look. I can try to take it on and implement it.

noor-tg:
Very good demo. I hope this will be merged to the core.

noor-tg:
@pathwave, are you alive? What happened to this PR ?

noor-tg:
> Hi! Could someone describe what needs to be done to merge this PR? Maybe some of the authors can share a vision of how the feature should look. I can try to take it on and implement it.

You can check the code . And start new PR based on this PR.

It seems this one was abandoned .

nik-rev:
It seems like this has been abandoned and the author isn't responding. I'll pick it up because I think this is a really great feature

JasterV:
This is a must-have feature that most code editors have nowadays, why has this been abandoned? I love Helix but today a friend showed me how to do this with Neovim using simple quickfix lists and I was like damn... I need this in Helix

noor-tg:
> This is a must-have feature that most code editors have nowadays, why has this been abandoned? I love Helix but today a friend showed me how to do this with Neovim using simple quickfix lists and I was like damn... I need this in Helix

Try scooter

Or make it using the new plugin system 🤔

archseer:
@JasterV If you finish implementing this we'll merge it ;) I'm interested in supporting this in core

JasterV:
@noor-tg Wait, there is a plugin system already??? Is it documented anywhere? :O

noor-tg:
> @noor-tg Wait, there is a plugin system already??? Is it documented anywhere? :O

https://github.com/helix-editor/helix/pull/8675

PR #4388, opened 2022-10-20
feat(lsp): re-use vector in recv_server_message

matoous:
Resolve tiny TODO to re-use vector for recv_server_message LSP transport function.


the-mikedavis (commented):
The `content` Vec isn't used in the loop in that function so I think this has the same behavior as the old code. To clear up this TODO I think we would need to add a `content_buffer` field to `Transport`

matoous:
@the-mikedavis yeah, I am blind 🙈 let me close this one, I will take another look tomorrow if I can actually resolve the TODO.

PR #4186, opened 2022-10-10
Apply transactions to jumplist selections

the-mikedavis:
From #3592:

> Previously, the jumplist was not updated by changes to the document. This meant that the jumplist entries were prone to being incorrect after some edits and that usages of the jumplist were prone to panics if, for example, some edits truncated the document.
> This change applies `Transaction`s (document edits) to the selections stored in any jumplists.

This is an alternative to https://github.com/helix-editor/helix/pull/3592

Sharing the selections between Editor and Document (as in #3592) is convenient because there are many call-sites for applying transactions to a document. It's a bit awkward to use shared memory between the View and Document though. This is an alternative that applies transactions to both Documents and Views separately and directly. We introduce a helper function to ensure they're called together.

Closes https://github.com/helix-editor/helix/issues/2489
Closes https://github.com/helix-editor/helix/pull/3592
Closes #3763

PR #4391, opened 2022-10-20
Fix integration test flakes in CI

the-mikedavis:
Write path fixes (#2267) introduced some flaky behavior in the integration test suite that happens when running within GitHub Actions.

The Windows runner has `clangd` installed, but it sometimes takes too long to shutdown which causes false-positive test failures. All runners have the `R` binary installed but not the language server package. This causes a flake whenever `tempfile::NamedTempFile` creates a file which is mistaken as R (this happens surprisingly often in practice). The R binary starts but immediately crashes because the language server package is not installed. Then the language server cannot be shutdown in the expected timeframe, resulting in a false-positive failure.

Here I've removed all language server configuration for the integration tests by default with the ability to add back the configuration explicitly. This fixes the flakes as far as I've tested and silences some unnecessary output when running locally (from `rust-analyzer`).

I've also added some small fixes to...

* use `helix_view::apply_transaction` instead of `Document::apply` (see #4186)
* silence the dead_code warning on `AppBuilder::with_config`

### DIFF
--- a/helix-term/tests/test/helpers.rs
+++ b/helix-term/tests/test/helpers.rs
@@ -286,6 +286,8 @@ impl AppBuilder {
         self
     }
 
+    // Remove this attribute once `with_config` is used in a test:
+    #[allow(dead_code)]
     pub fn with_config(mut self, config: Config) -> Self {
         self.config = config;
         self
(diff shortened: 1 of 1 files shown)


dead10ck:
Thanks!

Issue #2729, opened 2022-06-09
`delete_word_backward` and `delete_word_forward` behave unexpectedly in some cases


lunarpeach:
### Summary

The `delete_word_backward` and `delete_word_forward` behave unexpectedly in some cases.

My impression is that there are two things going on here: my expectation that `\n` counts as a word (which might not be a bug), and unexpected behavior when there's nothing under the cursor (which probably is).

### Reproduction Steps

1. Bind `delete_word_backward` and `delete_word_forward` to a key in insert mode.
2. Use the keybinds as in the following examples.

### `delete_word_forward`

With the cursor `|` on line 1:

    1 |
    2
    3 type some words

Pressing the keybind will remove `type_`, `some_`, and then `words`, leaving the cursor on line 1 and all the newlines intact.  I'd expect line 1 (the `\n`) to be deleted, then line 2 (another `\n`), then `type` (and maybe the following space?), etc.

### `delete_word_backward`

With the cursor on line 3:

    1 type some words
    2
    3 |

Pressing the keybind will delete `words`, `some_`, and then `type_`.  As in the forward case, this will leave the newlines untouched.  I'd expect the newlines between the cursor and the words to be deleted one at a time first.

..

With the cursor at the end of line 1 and no trailing newline:

    1 type some words.|

Nothing happens.

..

With the cursor on line 1 and no trailing newline:

    1 type some words|

Pressing the keybind will give:

    1 type some s|

And then the keybind no longer does anything until I reposition the cursor.

..

There's another weird case where, upon using the keybinding, I end up with two spaces between the cursor and whatever words are left, but I can't seem to reproduce it.

### Platform

Linux (5.18.2-arch1-1)

### Terminal Emulator

foot 1.12.1

### Helix Version

helix 22.05

s0LA1337:
This bug also effect `move_next_word_start` and `move_next_word_end` as well as the prev versions.

PabloMansanet:
Hey! I'm the one who wrote the unit tests and word navigation stuff :) after having a chat about it with @s0LA1337 , we concluded a couple things:

* Delete previous word is certainly bugged, as it seems to leave one character untouched when called after the word boundary. I'll be looking into why that is.
* Word movement is not bugged, it's just implementing kakoune-style word navigation as opposed to vim. Newlines on Kakoune are not handled as words, and this behaviour is consistent with that. We can have a conversation about whether that's desirable, though.

AceofSpades5757:
[...]

PR #4392, opened 2022-10-20
Potential fix to unexpected behavior in delete_word_backward and delete_word_forward

nimr0d:
Potential fix to #2729

I am unfamiliar with the codebase, so I may be doing something wrong.

The idea is to include every character between the cursor and the start/end of the prev/next word in the selection that is to be deleted, including newlines. 

In my brief testing, this leads to much natural behavior than what it was previously. However, the behavior still differs from the suggestion in #2729 which is to consider newlines to be words.


the-mikedavis (approved):
I think this behavior feels much more natural 👍

PR #4397, opened 2022-10-21
flush writes on force quit

dead10ck:
When force quitting, we need to block on the pending writes to ensure that write commands succeed before exiting, and also to avoid a crash when all the views are gone before the auto format call returns from the LS.

This fixes an intermittent failure in the integration tests as well.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -51,6 +51,7 @@ fn force_quit(
 
     ensure!(args.is_empty(), ":quit! takes no arguments");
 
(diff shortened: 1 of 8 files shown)


dead10ck:
Lol I guess this necessitates making all integration tests multithreaded

dead10ck:
Not really sure what's going on with the cargo fmt step. It seems to be complaining about trailing whitespace that doesn't exist in my local tree.

archseer:
It's just a failure on master, LGTM

PR #4406, opened 2022-10-21
lsp: Resolve completion items missing documentation on idle

the-mikedavis:
Some language servers may not send the `documentation` field if it is expensive to compute. Clients can request the missing field with a completionItem/resolve request.

In this change we use the idle-timeout event to ensure that the current completion item is resolved.

Closes https://github.com/helix-editor/helix/issues/2052

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -295,6 +295,27 @@ impl Completion {
     pub fn is_empty(&self) -> bool {
         self.popup.contents().is_empty()
     }
+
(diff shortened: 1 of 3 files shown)


dead10ck (approved):
This looks good to me

archseer:
Nice, it's interesting to see how many different features a simple idle timeout unlocks.

the-mikedavis:
Ikr 😀

This would have been much more difficult to do within the menu on update because of mutability rules in the callback

PR #4408, opened 2022-10-21
feat(commands): add clear_registers command

### LINKED ISSUE #4256: Delete registers
Would be nice if there was a way to delete registers, mine are pretty cluttered

chorcheus:
Hi,
there has been a feature request for an option to delete registers, so I politely created this. Thank you for your review.

Closes #4256

### DIFF
--- a/runtime/queries/nix/injections.scm
+++ b/runtime/queries/nix/injections.scm
@@ -6,6 +6,8 @@
   (indented_string_expression (string_fragment) @injection.content))
   (#set! injection.combined))
 
+; Common attribute keys corresponding to scripts,
+; such as those of stdenv.mkDerivation.
 ((binding
    attrpath: (attrpath (identifier) @_path)
    expression: (indented_string_expression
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Instead of this being a new command, I think this should be implemented as a special key in the `on_next_key` callback of `select_register`. So you would clear registers by pressing `"` and then the key to clear all registers. I'm not sure what the keybind should be though

geometryolife:
> Instead of this being a new command, I think this should be implemented as a special key in the `on_next_key` callback of `select_register`. So you would clear registers by pressing `"` and then the key to clear all registers. I'm not sure what the keybind should be though

Perhaps using `”|` is a good choice.

chorcheus:
> > Instead of this being a new command, I think this should be implemented as a special key in the `on_next_key` callback of `select_register`. So you would clear registers by pressing `"` and then the key to clear all registers. I'm not sure what the keybind should be though
> 
> Perhaps using `”|` is a good choice.

This could get tricky on different keyboard layouts - for example mine has it on opposite sides of keyboard. It could work if this key mapping was overridable - but for that one still needs a separate command to be called.

chorcheus:
@the-mikedavis @geometryolife Thank you for all suggestions, I implemented a solution. I'd be thankful for the review since I am still not sure about the keymapping.

chorcheus:
The work has been moved to https://github.com/helix-editor/helix/pull/5695; this PR's source branch is master, which isn't convenient.

PR #4409, opened 2022-10-21
Add textobjects.scm for zig

SoraTenshi:
Unfortunately the way how the Zig grammar is declared, i had to do some (probably) not so decent practices when it comes to TSQ.

However, this code is tested and i am pretty certain someone smarter than me might be able to refine this code a little bit.


the-mikedavis (approved):
These look good 👍

I see what you were saying in the Matrix about "pub". That will need to be fixed in the grammar itself, IMO "pub" should be part of the declaration node (for function/struct/enum/whatever)

PR #4085, opened 2022-10-03
Add :toggle-option command

A-Walrus:
This command toggles the value of boolean options. `:toggle key`
Mostly useful for custom keybindings to toggle settings such as `smart_case`, or `cursorline`.
Closes #2849

PR #4411, opened 2022-10-21
feat(cycle-opts): allows cycling option values at runtime

alevinval:
This patch implements a new `:cycle` command that supports an option's cycling of values, see the demo:

When the config is a boolean, it flips the value, otherwise cycles the argument strings.

Screen Recording 2022-10-21 at 22 42 02


archseer:
Looks similar to https://github.com/helix-editor/helix/pull/4085

alevinval:
> Looks similar to https://github.com/helix-editor/helix/pull/4085

You are right, with the addition that cycle allows us to cycle settings that are not booleans. Let me know the best way to move forward.

alevinval:
Maybe I could remove the logic that flips booleans, so boolean settings are handled with toggle. String settings are handled with cycle.

book/src/generated/typable-cmd.md
```diff
@@ -58,6 +58,7 @@
 | `:set-language`, `:lang` | Set the language of current buffer. |
 | `:set-option`, `:set` | Set a config option at runtime.<br>For example to disable smart case search, use `:set search.smart-case false`. |
 | `:get-option`, `:get` | Get the current value of a config option. |
+| `:cycle-option`, `:cycle` | Cycles a config option at runtime.<br>For example, to cycle a boolean config like smart case search, use `:cycle search.smart-case`.<br>For other kinds of configs use: `:cycle line-number absolute,relative`. |
```
alevinval, line 61:
```suggestion
| `:cycle-option`, `:cycle` | Cycles a config option at runtime.<br>For example, to cycle boolean configs like smart case search, use `:cycle search.smart-case`.<br>For other kinds of configs, use: `:cycle line-number absolute,relative`. |
```

helix-term/src/commands/typed.rs
```diff
@@ -1357,6 +1357,70 @@ fn set_option(
     Ok(())
 }
 
+/// Cycle the setting to the next value at runtime. Access nested values by dot
+/// syntax, for example to cycle smart case search, use `:cycle search.smart-case false,true`.
```
alevinval, line 1361:
```suggestion
/// syntax, for example to cycle line number, use `:cycle line-number absolute,relative`.
```

archseer:
Ah sorry, I forgot about this PR and merged https://github.com/helix-editor/helix/pull/4085 first. Can you rebase and update the `toggle` implementation? I think your behaviour is preferrable but I'd still call the command `toggle`.

I also think it would be interesting if it allowed you to cycle between options without having to specify possible values.

David-Else:
Thanks for this, I am really hoping it will have the ability to toggle line numbers:
```
gutters = ["diagnostics", "spacer", "spacer", "diff"] # no line-numbers
```
```
gutters = ["diagnostics", "spacer", "spacer", "diff", "line-numbers"]
````
Is that possible yet?

**Comment by divarvel** [comment]
> Ah sorry, I forgot about this PR and merged #4085 first. Can you rebase and update the `toggle` implementation? I think your behaviour is preferrable but I'd still call the command `toggle`.
> 
> I also think it would be interesting if it allowed you to cycle between options without having to specify possible values.

I thought about it when independently working on a toggle command. With enumerated values we have enough information to cycle, but the way things are currently implemented rely on the toml encoding, which doesn't have this information anymore.

I think this is especially important to avoid boolean blindness and avoid nudging implementors using booleans when 2-valued enums would be better.

**Comment by alevinval** [comment]
@archseer you want to update the toggle implementation to what I propose in the cycle one?

**Comment by alevinval** [comment]
Updated toggle to support cycling of options 👍

File: helix-term/src/commands/typed.rs [hunk 3]
```diff
@@ -1801,22 +1800,48 @@ fn toggle_option(
...
+                .unwrap_or_else(|| first_arg.parse().map_err(field_error))?
+        }
     };
 
-    let new_value = !old_value;
-    *value = Value::Bool(new_value);
-    // This unwrap should never fail because we only replace one boolean value
-    // with another, maintaining a valid json config
-    let config = serde_json::from_value(config).unwrap();
+    let status = format!("Option `{}` is now set to `{}`", key, value);
```
pascalkuthe, line 1836:
```suggestion
    let status = format!("'{key}' is now set to '{value}'");
```

helix-term/src/commands/typed.rs
```diff
@@ -1801,22 +1800,48 @@ fn toggle_option(
...
+                        [first]
+                            .into_iter()
+                            .chain(remaining.split(','))
+                            .skip_while(|e| e != value)
+                            .nth(1)
+                            .unwrap_or(first)
+                            .to_string(),
+                    ))
+                })
+                .unwrap_or_else(|| first_arg.parse().map_err(field_error))?
```
pascalkuthe, line 1832: this feels odd. Instead of doing the parsing yourself why not simply require one argument per value?

```
:toggle text-width 10 80 100
```

the check above would turn into `ensure(args.len()  > 2,...)`

This feels more natural to me and works better with the existing infrastructure around commands

helix-term/src/commands/typed.rs
```diff
@@ -1801,22 +1800,48 @@ fn toggle_option(
...
+        Some(value) => serde_json::Value::Bool(!value),
+        None => {
+            if args.len() != 2 {
+                anyhow::bail!(
+                    "Bad arguments. For non-boolean configurations use: `:toggle key val1,val2`"
+                );
+            }
+            if !value.is_string() {
+                anyhow::bail!("Bad configuration. Cannot cycle non-string configurations");
+            }
```
pascalkuthe, line 1813: you can use the `ensure` macro for checks like this

helix-term/src/commands/typed.rs
```diff
@@ -1801,22 +1800,48 @@ fn toggle_option(
     let pointer = format!("/{}", key.replace('.', "/"));
     let value = config.pointer_mut(&pointer).ok_or_else(key_error)?;
 
-    let Value::Bool(old_value) = *value else {
-        anyhow::bail!("Key `{}` is not toggle-able", key)
+    *value = match value.as_bool() {
+        Some(value) => serde_json::Value::Bool(!value),
```
pascalkuthe, line 1804: for bools you should still check that there is only one argument

pascalkuthe (changes requested):
great that you are picking this back up!

helix-term/src/commands/typed.rs
alevinval: I don't know what I was thinking. Much cleaner, see https://github.com/helix-editor/helix/pull/4411/commits/998426efcf754b738914a9d92b04e1a6a5bca5aa

helix-term/src/commands/typed.rs
alevinval: Neat, see https://github.com/helix-editor/helix/pull/4411/commits/998426efcf754b738914a9d92b04e1a6a5bca5aa

helix-term/src/commands/typed.rs
alevinval: Added the check in https://github.com/helix-editor/helix/pull/4411/commits/998426efcf754b738914a9d92b04e1a6a5bca5aa

helix-term/src/commands/typed.rs
alevinval: Done in https://github.com/helix-editor/helix/pull/4411/commits/998426efcf754b738914a9d92b04e1a6a5bca5aa

pascalkuthe (approved):
LGTM this makes `:toggle` much more useful 👍

David-Else:
@alevinval Thanks, great contribution! 

I am fiddling around trying to answer my own question but can't get anywhere, could you quickly let me know if https://github.com/helix-editor/helix/pull/4411#issuecomment-1427648490 is possible?

alevinval:
To keep it short: no, the current implementation is quite limited in supporting cycling of configuration values that have one possible value. What you are suggesting would require adding/removing elements from the vector. I could look into extending the implementation to deal with arrays, such that if you do `:toggle gutters.layout line-numbers` it simply adds/removes the element from the array. I'll see when I have some time on my hands.

alevinval:
Looking back into this... I've just realised that I wrote some poor code here, I did not realise that I could pattern match on `serde_json::Value` cause it's an Enum... I've opened a tiny follow-up to this PR here:

https://github.com/helix-editor/helix/pull/7240/files

It does two things:

- Gives better error message feedback to the user, when trying to toggle something that is not supported.
- Simplifies code, as the type conversion happens within the matching
- Makes it obvious on what is the way forward: which missing enum variants are pending to be implemented

David-Else:
@alevinval With great joy I realized that a toggleable zen mode is almost within reach, but this does not work :(

```
e = ":toggle gutters.line-numbers.min-width 48,0"
```
`Configuration gutters.line-numbers.min-width does not support toggle yet`

Would it be easy to fix? Would be a massive feature!

alevinval:
I did not implement `Number` because I did not see a straightforward way to get a `Value::Number(Number)` out of a `String`. Without this, you would need to parse it into some float or something. But this feels like a rabbit-hole. I feel there should be a way to get serde numbers from strings... But I don't see how.

David-Else:
Cheers, using this for now :)

```
[keys.normal.space]

Z = ":set gutters.line-numbers.min-width 3"
z = ":set gutters.line-numbers.min-width 48"
```

gabydd:
Is there reason you can't just use parse like in set-option for everything that's not a string or a bool: https://github.com/helix-editor/helix/blob/master/helix-term/src/commands/typed.rs#L1802

alevinval:
Ah you're genius! Here's the PR to add support to toggle numbers: https://github.com/helix-editor/helix/pull/7877

PR #4412, opened 2022-10-21
feat(statusline): if color_modes is enabled, color the file type as well

etienne-k:
This is a little cosmetic improvement, which colors the file type in addition to the mode; ~~shamelessly copied from~~ inspired by [lualine](https://github.com/nvim-lualine/lualine.nvim).

In addition to the subjectively better look, I found that having only the mode colored made it *intuitively unclear* which buffer the mode belongs to, when working with a vertical split. When both the left and the right side of the focused statusline are highlighted, however, this ambiguity disappears:

image

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -340,8 +340,21 @@ where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
     let file_type = context.doc.language_name().unwrap_or("text");
+    let visible = context.focused;
 
-    write(context, format!(" {} ", file_type), None);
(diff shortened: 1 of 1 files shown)


CptPotato:
Could this be kept optional? I personally prefer the current look.

Alternatively, defining which status line elements to color could be part of the status line configuration. However, that might over-complicate things 😕.

etienne-k:
I'll close this for now, because indeed, this solution might be too invasive for some, but I'd rather avoid yet another very specific configuration option. I'd rather rethink it and allow the user to define what must be colored and when, or simply coloring elements on the left and right side (like lualine does).

Issue #4407, opened 2022-10-21
Alias `insert-output` from `io`


lukepighetti:
Would be nice to quickly fire off `:insert-output` with `:io`

Omnikar:
Forgive me if I'm wrong, but isn't `:insert-output` equivalent to `!`?

Bruce-Hopkins:
I'm looking to contribute to this project. Can I be assigned the issue?

lukepighetti:
@the-mikedavis @Omnikar makes a good point here. Should we close this?

geometryolife:
I don’t think it’s necessary to add aliases, because it is convenient to run `insert-output` and `append-output` with `!` and `A-!`.

the-mikedavis:
Hmm yeah, let's close this. In general, aliasing would be improved by https://github.com/helix-editor/helix/issues/582 or this case could be covered by https://github.com/helix-editor/helix/issues/4423. But key-rebinding or using the existing `!`/`A-!` would be more convenient here.

PR #4413, opened 2022-10-21
feat(commands): io and ao aliases

Bruce-Hopkins:
Adds aliases to `insert-output` and `append-output`. As mentioned in #4407.

Are there any integration tests that are needed for a small edition like this one?


the-mikedavis:
Closing since we closed #4407 (https://github.com/helix-editor/helix/issues/4407#issuecomment-1287827153)

the-mikedavis:
> Are there any integration tests that are needed for a small edition like this one?

For adding an alias I think an integration test would be overkill

PR #4415, opened 2022-10-21
feat(themes): add mode-specific styles to the tokyonight themes

etienne-k:
This little PR adds mode-specific colors to both tokyonight themes, using the same reasoning as in https://github.com/helix-editor/helix/pull/3098, with one exception: I went with orange for the selection like in the similar cattpuccin themes. IMO it fits well in the theme, while still distinguishing the mode well enough.

**tokyonight**

image
image
image

**tokyonight_storm**

image
image
image


SoraTenshi:
Seems like we made the same PR basically at the same time.
Would you be fine if i merge some of my change into yours?
(Also i kind of dislike the orange)

etienne-k:
Oh, now that's a coincidence! Thanks for the heads up, @SoraTenshi.
I merged your changes and took over your colors, except that I left blue instead of cyan. I was quite uncertain about it, but cyan feels a bit too... electric? aggressive? ... for a normal, initially active mode.

**tokyonight**

image
image
image

**tokyonight_storm**

image
image
image

paulgraydon:
Hey guys! Love the colors chosen here, and I do agree magenta for select mode fits the other two softer colors very nicely. Very nice of you to do this, thank you!

runtime/themes/tokyonight.toml
```diff
@@ -41,6 +41,9 @@
 "ui.selection.primary" = { bg = "background_highlight" }
 "ui.statusline" = { fg = "foreground", bg = "background_menu" }
 "ui.statusline.inactive" = { fg = "foreground_gutter", bg = "background_menu" }
+"ui.statusline.normal" = { fg = "black", bg = "blue" }
+"ui.statusline.insert" = { fg = "black", bg = "green" }
+"ui.statusline.select" = { fg = "black", bg = "magenta" }
```
SoraTenshi, line 46: Instead of 'dark' for the color, I think background_gutter makes more sense.
This is a very weak opinion tho, as it doesn't really change anything in how it looks.

SoraTenshi (approved):

runtime/themes/tokyonight.toml
etienne-k: Hmm, `background_gutter` is too light:
image

But I see you had `background_menu` in your PR, could that be the color you meant?

That said, the reason I initially went with `black` was because, if we ever create a light theme version of tokyonight (such as here: https://github.com/folke/tokyonight.nvim#day), it would be important that the foreground color remains black/dark, regardless of the gutter/menu background colors (which would most definitely become light).

What do you think?

SoraTenshi: > Hmm, `background_gutter` is too light: image
> 
> But I see you had `background_menu` in your PR, could that be the color you meant?
> 
> That said, the reason I initially went with `black` was because, if we ever create a light theme version of tokyonight (such as here: https://github.com/folke/tokyonight.nvim#day), it would be important that the foreground color remains black/dark, regardless of the gutter/menu background colors (which would most definitely become light).
> 
> What do you think?

That's a decent enough reason. Because i haven't thought about it, i just had it consistent with the inactive one.
Otherwise, if that one doesn't look good, i am fine with the changes. :) Good job! :)

the-mikedavis (approved):

the-mikedavis:
Looks good! And thanks for cleaning up the storm variant with the `inherits` 👍

PR #4416, opened 2022-10-21
Cleanup tokyonight theme

SoraTenshi:
- Use inherit for tokyonight_storm (the theme i use!)
- Use statusline background colors if `color-modes`

Normal:
image
Insert:
image
Select:
image


SoraTenshi:
Maybe it would also be a good idea to wait for the review of @paulgraydon

SoraTenshi:
Since #4415 exists, turned this to a draft now.

Issue #2254, opened 2022-04-24
Config option for single status line


dasBurdock:
Support for using a global status line for the active window. 

This[ was recently introduced](https://github.com/neovim/neovim/pull/17266) in NeoVim as `laststatus = 3` and is a great option for clearer UI. 

https://user-images.githubusercontent.com/7904185/152662205-ff259d0c-2b55-458a-8951-41b6c84eeded.mov

PR #4417, opened 2022-10-22
feat(view): global status line option

matoous:
Add option to render single global status line instead of rendering status line per view.

With
```
[editor.statusline]
render = "per_view"
```
(default)

Screenshot 2022-10-23 at 21 52 52

with

```
[editor.statusline]
render = "single"
```

Screenshot 2022-10-23 at 21 50 04

Fixes: https://github.com/helix-editor/helix/issues/2254

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -19,7 +19,7 @@ use helix_core::{
 use helix_view::{
     apply_transaction,
     document::{Mode, SCRATCH_BUFFER_NAME},
-    editor::{CompleteAction, CursorShapeConfig},
+    editor::{CompleteAction, CursorShapeConfig, StatusLineRenderConfig},
     graphics::{Color, CursorKind, Modifier, Rect, Style},
     input::{KeyEvent, MouseButton, MouseEvent, MouseEventKind},
     keyboard::{KeyCode, KeyModifiers},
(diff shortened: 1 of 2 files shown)


matoous:
@CBenoit could you adjust the label to “waiting-on-review” pleas?

dahmc:
Great work @matoous , I love using a single status line!

Don't we also need something equivalent to neovim's `winbar` to show the file name in at all times? If you have multiple buffers next to each other, you don't know which file is which otherwise when using single status line.

Here's an screenshot of what I mean, latest neovim with `laststatus=3` (global status line) and `winbar=%f` (showing file path):
image

I just feel these 2 features belong together. But maybe this should be in a separate PR? Let me know your thoughts!

matoous:
@dahmc I am sorry, I am not sure if I understand. The filename of the document in currently active window is displayed in the status line. Furthermore, there's bufferline for buffers. I am not sure how else we could display the filename withing single status line.

dahmc:
@matoous The filename for the currently active buffer is shown in the status line, correct.

What I'm referring to is that we now don't know what the filename of the other buffer is, until we switch to it.

The bufferline doesn't really help here, because there could be 10 open buffers, and it's now clear what the filename of the inactive buffer is exactly.

Does that make sense? :)

matoous:
@dahmc understood. Yet, I am not sure how that would be solved if we have single status line. Maybe a configuration option for bufferline would do? If you want to see the filename for all open views then I don’t think singular status line would work.

dahmc:
@matoous Yeah I think the only way to make this work would be how it's done in neovim what I've shown in my screenshot above, using something like a winbar, which sits below the buffer line

TornaxO7:
may I ask if I can try to take over this?

matoous:
@TornaxO7 sure, feel free to.

TornaxO7:
Ok, so would it e fine, if I *only* focus on the statusline?

matoous:
Closing in favor of @TornaxO7 's PR.

Issue #3910, opened 2022-09-20
Multiple selection increment


MGlolenstine:
### Summary

```js
data[0] = 15;
data[0] = 15;
data[0] = 15;
data[0] = 15;
data[0] = 15;
```
and I'd like to increment them as such.
```js
data[0] = 15;
data[1] = 16;
data[2] = 17;
data[3] = 18;
data[4] = 19;
```
if I select all of the 15s but the first one and press C-a, it increments them all to 16 and not like I want.

### Reproduction Steps

 

1. Open a file in helix
2. Paste the above text into it
3. Navigate to `15`, press `e` and `C` few times to select all `15`s.
4. Press `C-a`, which increments numbers.
5. This results in all numbers becoming `16`s and not being incremented from 16-19 as expected.
6. Same happens for decrements.

### Helix log

Helix log is empty.

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.08.1

CptPotato:
To my knowledge this is actually the intended behavior, not a bug. Increment doesn't produce a range but increments each selected number separately.

Though, I've been in a similar situation where I wanted to generate a range of numbers a few times. For reference, Sublime Text has plugins such as [Text Pastry](https://github.com/duydao/Text-Pastry/wiki/Number-Range) that enable this.

MGlolenstine:
Interesting. I thought it would be the same way as it is with Vim. `g <C-a>`

Omnikar:
One way to achieve the desired behaviour here would be to make sure the top selection is the primary one, and alternate `C-a` and `A-,` until there is 1 selection left. This could also be sped up using a macro.

dead10ck:
Yeah as far as I know this is expected behavior. This could definitely be a new command though.

MGlolenstine:
I'd love to work on implementing command like this, but I'll need some hand-holding, as I'm new to the codebase.

ArthurPinheiroSo:
@MGlolenstine This is a change you can make in the **increment_impl** function inside the **src/commands.rs** to make the increment function behave like vim's g+<C-a> :
![image](https://user-images.githubusercontent.com/42854638/191764107-2536e148-6dc7-4431-b886-46dd2031f555.png)
I'm not sure if there is a better way to do this since I'm new to this project and also new to rust.

If we are going to implement this there is still the need to define which shortcut we can use.

Bruce-Hopkins:
[...]

PR #4418, opened 2022-10-22
feat(commands) increment by range

Bruce-Hopkins:
Increment increases the increment amount by 1 (or negative 1) for each value based on issue #3910.

This completely changes the way increment works, so maybe this is better as another command. I don't know what that would be though.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4809,8 +4809,8 @@ fn add_newline_impl(cx: &mut Context, open: Open) {
 }
 
 enum IncrementDirection {
-    Increase = 1,
-    Decrease = -1,
+    Increase,
+    Decrease,
 }
 /// Increment object under cursor by count.
 fn increment(cx: &mut Context) {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -435,6 +435,8 @@ impl MappableCommand {
         rename_symbol, "Rename symbol",
         increment, "Increment item under cursor",
         decrement, "Decrement item under cursor",
+        increasing_increment, "Increment item under cursor. Increment amount increases by 1 each item",
+        decreasing_decrement, "Decrement item under cursor. Decreament amount decreases by 1 each item",
         record_macro, "Record macro",
         replay_macro, "Replay macro",
         command_palette, "Open command palette",
@@ -4810,12 +4812,20 @@ fn add_newline_impl(cx: &mut Context, open: Open) {
 
 /// Increment object under cursor by count.
 fn increment(cx: &mut Context) {
-    increment_impl(cx, cx.count() as i64);
+    increment_impl(cx, cx.count() as i64, 0);
 }
 
 /// Decrement object under cursor by count.
 fn decrement(cx: &mut Context) {
-    increment_impl(cx, -(cx.count() as i64));
+    increment_impl(cx, -(cx.count() as i64), 0);
+}
+/// Increment object under cursor by count. Each increment increases by 1.
+fn increasing_increment(cx: &mut Context) {
+    increment_impl(cx, cx.count() as i64, 1);
+}
+/// Decrement object under cursor by count. Each decrement decreases by -1.
+fn decreasing_decrement(cx: &mut Context) {
+    increment_impl(cx, -(cx.count() as i64), -1);
 }
 
 /// This function differs from find_next_char_impl in that it stops searching at the newline, but also
@@ -4839,7 +4849,7 @@ fn find_next_char_until_newline<M: CharMatcher>(
 }
 
 /// Decrement object under cursor by `amount`.
-fn increment_impl(cx: &mut Context, amount: i64) {
+fn increment_impl(cx: &mut Context, amount: i64, increase_by: i64) {
     // TODO: when incrementing or decrementing a number that gets a new digit or lose one, the
     // selection is updated improperly.
     find_char_impl(
@@ -4871,14 +4881,7 @@ fn increment_impl(cx: &mut Context, amount: i64) {
 
             let (range, new_text) = incrementor.increment(amount);
 
-            // If the ammount is positive, increase the increment amount by 1
-            let mut value = 1;
-
-            // If the ammount is negative, increase the increment amount by - 1
-            if amount < 0 {
-                value = -1;
-            }
-            amount = amount + value;
+            amount += increase_by;
 
             Some((range.from(), range.to(), Some(new_text)))
         })
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4855,6 +4855,7 @@ fn increment_impl(cx: &mut Context, amount: i64) {
     let selection = doc.selection(view.id);
     let text = doc.text().slice(..);
 
+    let mut amount = amount;
     let changes: Vec<_> = selection
         .ranges()
         .iter()
@@ -4869,6 +4870,15 @@ fn increment_impl(cx: &mut Context, amount: i64) {
                 };
 
             let (range, new_text) = incrementor.increment(amount);
+
+            // If the ammount is positive, increase the increment amount by 1
+            let mut value = 1;
+
+            // If the ammount is negative, increase the increment amount by - 1
+            if amount < 0 {
+                value = -1;
+            }
+            amount = amount + value;
 
             Some((range.from(), range.to(), Some(new_text)))
         })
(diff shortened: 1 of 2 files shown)


SoraTenshi:
I think it'd be better to add this functionality into a seperate command, the default behaviour shouldn't be changed for a normal increment, as there a legitimate usecases for it.

also, once the increment counter hit 10+ the results get scewed heavily.

olafmustafar:
> also, once the increment counter hit 10+ the results get scewed heavily.

I believe this issue may be corrected by #4123

Bruce-Hopkins:
> I think it'd be better to add this functionality into a seperate command, the default behaviour shouldn't be changed for a normal increment, as there a legitimate usecases for it.

I agree. I'm trying to figure out which keymap to connect both of them two. Temporarily, I set it to Alt-a. Any suggestions?

> I believe this issue may be corrected by https://github.com/helix-editor/helix/pull/4123

I hope I'm not causing bad merge conflicts.

helix-term/src/keymap/default.rs
```diff
@@ -311,6 +311,9 @@ pub fn default() -> HashMap<Mode, Keymap> {
 
         "C-a" => increment,
         "C-x" => decrement,
+
+        "A-a" => increasing_increment,
+        // "A-x" => decreasing_decrement,
```
SoraTenshi, line 316: Why is this uncommented?

SoraTenshi (changes requested):
Also please update: https://github.com/helix-editor/helix/blob/master/book/src/keymap.md

helix-term/src/keymap/default.rs
Omnikar: `A-x` is already used for `shrink_to_line_bounds`.

Bruce-Hopkins: > `A-x` is already used for `shrink_to_line_bounds`.

Do you have any suggestions for a better keymap?

Bruce-Hopkins:
@CBenoit I've changed the keymappings to be alt+ctrl

is this good to be merged?

SoraTenshi:
> I've changed the keymappings to be alt+ctrl

Is C-A (ctrl + shift) and C-X already reserved / taken? Because i think they'd make more sense...

Bruce-Hopkins:
> Is C-A (ctrl + shift) and C-X already reserved / taken? Because i think they'd make more sense...

I've tried those keymaps but for some reason, they don't work. I'm guessing you can't pair (alt + shift) in the keymaps. Try it out and let me know if it works for you.

the-mikedavis:
I don't think we implement the `#` special register from Kakoune yet (https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers) but `"#<C-a>` and `#<C-x>` could be an nice way to do this without needing to add a new command and keybind.

Bruce-Hopkins:
> I don't think we implement the `#` special register from Kakoune yet (https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers) but `"#<C-a>` and `#<C-x>` could be an nice way to do this without needing to add a new command and keybind.

So should we close the PR until the special register is implemented? Or should this just be a temporary fix until then?

the-mikedavis:
I think it would be ok to kind of hack it in this PR: in `increment_impl` check the `cx.register` to see if it's `'#'` and if it is then increment by range number

Bruce-Hopkins:
> I think it would be ok to kind of hack it in this PR: in `increment_impl` check the `cx.register` to see if it's `'#'` and if it is then increment by range number (multiplied by count?)

Oh, I didn't even know we could see which register we're using. How about this? Does this look good?

helix-term/src/commands.rs
```diff
@@ -4839,7 +4852,7 @@ fn find_next_char_until_newline<M: CharMatcher>(
 }
 
 /// Decrement object under cursor by `amount`.
-fn increment_impl(cx: &mut Context, amount: i64) {
+fn increment_impl(cx: &mut Context, amount: i64, increase_by: i64) {
```
the-mikedavis, line 4855: Instead of passing amounts to `increase_impl`, we can create an enum

```rust
enum IncrementDirection {
    Increase,
    Decrease,
}
```

And pass that to `increment_impl`, then check the `cx.count` and `cx.register` in the body of `increment_impl` and then decide whether to use a positive or negative amount based on the passed direction.

Bruce-Hopkins: Okay, won't we need a 3rd value for the enum if we don't want to increase it at all?

> And pass that to increment_impl, then check the cx.count and cx.register in the body of increment_impl and then decide whether to use a positive or negative amount based on the passed direction.

Alright, that makes sense.

Bruce-Hopkins:
@the-mikedavis How does it look now?

helix-term/src/commands.rs
the-mikedavis: Count can't be 0 so we don't need a third member

helix-term/src/commands.rs
```diff
@@ -4839,7 +4844,7 @@ fn find_next_char_until_newline<M: CharMatcher>(
 }
 
 /// Decrement object under cursor by `amount`.
-fn increment_impl(cx: &mut Context, amount: i64) {
+fn increment_impl(cx: &mut Context, amount: i64, increment_direction: IncrementDirection) {
```
the-mikedavis, line 4847: You can use the direction parameter and `cx.count()` to determine the amount in this function and then `amount` doesn't need to be a parameter.

For example, `increment` can be implemented just as

```rust
fn increment(cx: &mut Context) {
    increment_impl(cx, IncrementDirection::Increase);
}
```

helix-term/src/commands.rs
```diff
@@ -4851,10 +4856,27 @@ fn increment_impl(cx: &mut Context, amount: i64) {
...
+    let increase_by = match cx.register {
+        Some(value) => match value.eq(&'#') {
+            true => match increment_direction {
+                IncrementDirection::Increase => 1,
+                IncrementDirection::Decrease => -1,
+            },
+            false => 0,
+        },
+        None => 0,
+    };
```
the-mikedavis, line 4878: This can be simplified a bit:

```rs
let sign = match increment_direction {
    IncrementDirection::Increase => 1,
    IncrementDirection::Decrease => -1,
};
let mut amount = sign * cx.count() as i64;
let increase_by = if cx.register == Some('#') {
    sign
} else {
    0
};

Comment (the-mikedavis) [reply helix-term/src/commands.rs:4878]: It could also make sense to define IncrementDirection as

```rust
enum IncrementDirection {
    Increase = 1,
    Decrease = -1,
}
```

and then this could be:

```rust
let mut amount = increment_direction as i64 * cx.count() as i64;
let increase_by = if cx.register == Some('#') {
    increment_direction as i64
} else {
    0
};
```

but that's a bit less clear readability-wise so I don't think it's worth it

Bruce-Hopkins: > It could also make sense to define IncrementDirection as
> 
> ```rust
> enum IncrementDirection {
>     Increase = 1,
>     Decrease = -1,
> }
> ```
> 
> and then this could be:
> 
> ```rust
> let mut amount = increment_direction as i64 * cx.count() as i64;
> let increase_by = if cx.register == Some('#') {
>     increment_direction as i64
> } else {
>     0
> };
> ```
> 
> but that's a bit less clear readability-wise so I don't think it's worth it

I didn't even know you could use enums like that in Rust.

I updated it, how does it look now? I added comments to help readability. How does the wording look?

helix-term/src/commands.rs
```diff
@@ -4808,14 +4808,18 @@ fn add_newline_impl(cx: &mut Context, open: Open) {
     apply_transaction(&transaction, doc, view);
 }
 
+enum IncrementDirection {
+    Increase = 1,
+    Decrease = -1,
```
the-mikedavis, line 4813: Since we don't end up using `increment_direction as isize`, the numbers here can be removed

helix-term/src/commands.rs
the-mikedavis: Wording looks good 👍

helix-term/src/commands.rs
Bruce-Hopkins: That was an accident. My bad.

the-mikedavis:
This is looking good 👍

Just needs a run of `cargo fmt`

Bruce-Hopkins:
I just ran the formatter.

CBenoit (approved):
Thank you! 😄

Diegovsky:
Now that this uses the "#" register, how do I create a remap for this? The following registers the macro, but executing it doesn't seem to work unfortunately:
```toml
C-A = "\"#C-A" 
```

(Also, should I open a discussion for this?)

PR #4424, opened 2022-10-23
add cursor movement in insert mode

geometryolife:
I've added shortcuts for moving the cursor in insert mode for Helix, which I think will make them feel at home for users who frequently edit on **iPad/Mac/Emacs**.
`C-p` -> `Up`
`C-n` -> `Down`
`C-b` -> `Left`
`C-f` -> `Right`
`C-a` -> `Home`
`C-e` -> `End`

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -356,14 +356,14 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "C-j" | "ret" => insert_newline,
         "tab" => insert_tab,
 
-        "up" => move_line_up,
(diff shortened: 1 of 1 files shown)


archseer:
I'm against having these built in, especially because they conflict with the completion keybinds so you might accidentally scroll the popup rather than moving. I think we removed Ctrl-n/-p previously because of this exact reason.

Users that want these can define them in their own config. We recently removed all movement in insert mode but decided to match kakoune and bring back a minimal set: https://github.com/helix-editor/helix/pull/3915

geometryolife:
I understand, it really shouldn't be built-in.

PR #4425, opened 2022-10-23
Provide option to completely disable LSP

dead10ck:
I had had this code sitting around in a branch I'm working on already before #4391 became necessary in order to turn off LSP for the integration tests. It slows them down and is totally unnecessary until we figure out a good way to do integration tests with real language servers.

I'm making this PR as a proposed different way to approach this, where this is generalized into a user config option exposed to the user. I figure that while my motivation for this was reducing noise in the integration tests, this could conceivably be desirable by some users without having to e.g. compile out LSP integration with a feature flag.

Another benefit is that when we do get to integration tests around LSP, we won't need to manually specify a language config; we'll just need to set this one boolean config setting in the test invocation, which is a lot less tedious, and allows always testing the same default configs that ship to users.

With that said, I don't actually have any strong feelings about keeping this, I just thought it was worth proposing and seeing what everyone thinks since I had already done it. I'm perfectly content leaving it the way it is now.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -115,6 +115,7 @@ The following statusline elements can be configured:
 
 | Key                   | Description                                                 | Default |
 | ---                   | -----------                                                 | ------- |
+| `enable`              | Enables LSP integration. Setting to false will completely disable language servers regardless of language settings.| `true` |
 | `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
 | `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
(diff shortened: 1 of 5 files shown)


the-mikedavis (commented):
I think this looks good - I would prefer an option like this to the change in #4391 (removing LSP config by deleting the toml keys). I think this just needs a new line in the book for the new option: https://github.com/helix-editor/helix/blob/801984c7fc2cd3f4072f1bdf42c5fbfe1a634348/book/src/configuration.md?plain=1#L111

dead10ck:
Done

the-mikedavis (approved):

dead10ck:
@archseer any thoughts?

PR #4435, opened 2022-10-23
Add ability to load files in the background without closing the picker

### LINKED ISSUE #3924: S-return in file picker should open file in the background and keep the picker open with the same search query
A lot of the time I want to open multiple files at once from the file picker and I think this small qol feature could make it easier to do so.

I would like to help write a PR to close this issue myself. Could anyone direct me as to where I should be looking to see how to make this work?

Itay123TheKing:
Closes #3924. This is a pretty small change but everyone has to start somewhere :P

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -1,4 +1,5 @@
 use crate::{
+    alt,
     compositor::{Component, Compositor, Context, Event, EventResult},
     ctrl, key, shift,
     ui::{self, fuzzy_match::FuzzyQuery, EditorView},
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
I'm a fan of this - I run into this situation a lot too where you want to open a file and keep the picker open (usually to change query) but want to open the file, as a more convenient way than `<space>'`

dead10ck (approved):
Thanks!

lukepighetti:
What's stopping this from merging?

dead10ck:
A review from @archseer.

PR #4440, opened 2022-10-23
Added zoxide support

ElSargo:
Adds support for zoxide integration, a command that alows for changing directories with zoxide, rather than :cd


the-mikedavis:
IMO this is too specific: this kind of integration should wait for the plugin system

PR #4443, opened 2022-10-24
Supertab (aka tabout)

dead10ck:
This adds new commands that will traverse up to the parent tree-sitter node and move to either the beginning or the end. It also adds an optional feature I've called "supertab" (name is still open to suggestion) which will run an arbitrary command when tab is pressed, unless the cursor is preceded only by whitespace, in which case it will insert a tab as normal. This can be used to emulate a tabout behavior with the new commands.

[asciicast](https://asciinema.org/a/7fbJS9wfVi2rI7KftEUosz8F1)

Users can set the command in their config with e.g.

```toml
[editor]
supertab = "move_parent_node_end"
```

but could also just as easily set it to `extend_selection` or anything else.

Some things I'm unsure about in the approach:

* I've modified the `insert_tab` command. Alternatively, this could just be a new dedicated command.
* In the menu, the tab key is bound to cycling through menu items. This cuts the usefulness of this feature because when the auto complete menu pops up, it doesn't work any more. To get around this, I added a special case to ignore the tab key in the menu component, but this feels a little bit hacky to me. I'm not really sure what else to do though.
* Key mappings. Really not sure what to bind these to in the default mappings. Personally I've found it so useful that I've bound the new movement commands to tab in normal mode as well, although due to the terminal conflict with `C-i`, this would clearly be a controversial choice in the default bindings. 

  So far the config I've been using is:

  ```toml
  [editor]
  supertab = "move_parent_node_end"

  [keys.normal]
  tab = "move_parent_node_end"
  S-tab = "move_parent_node_start"

  [keys.normal.g]
  o = "jump_backward"
  i = "jump_forward"

  [keys.insert]
  S-tab = "move_parent_node_start"

  [keys.select]
  tab = "extend_parent_node_end"
  S-tab = "extend_parent_node_start"

  [keys.select.g]
  o = "jump_backward"
  i = "jump_forward"
  ```

  It's been working pretty well for me, though I'm open to any suggestions for default bindings.

This change is based on top of #4425.

Additional changes include:

- reorganize write command tests to their own module
- make TestCase::From more generic for ergonomics
- print doc state during tests for much better debugging
- To help with ergonomics, `test::print` was changed to allow explicit newlines in test DSL. The current test DSL currently has no way to express being at the end of a line, save for putting an explicit LF or CRLF inside the `#[|]#`. The problem with this approach is that it can add unintended extra new lines if used in conjunction with raw strings, which insert newlines for you.

  This is a simple attempt to mitigate this problem. If there is an explicit newline character at the end of the selection, and then it is immediately followed by the same newline character at the right end of the selection, this following newline is removed. This way, one can express a cursor at the end of a line explicitly. I'm including this in the PR because without it, it is difficult to express the doc text in a readable way in the integration tests
- fix a bug `test::plain`: it is currently using char indices for `String::insert_str` which is supposed to take byte indices. This causes panics on Unicode characters


helix-view/src/editor.rs
```diff
@@ -900,18 +907,25 @@ impl Editor {
...
-        let doc_url = doc.url()?;
+        let (lang, path) = {
+            let doc = self.document(doc_id)?;
+            (doc.language.clone(), doc.path().cloned())
+        };
 
         // try to find a language server based on the language name
-        let language_server = doc.language.as_ref().and_then(|language| {
-            ls.get(language, doc.path())
+        let language_server = lang.as_ref().and_then(|language| {
```
archseer, line 926: Why is `doc.language` cloned into `lang` here?

dead10ck: There is a borrowing conflict otherwise. But this is just an `Arc` so it's pretty cheap anyway.

helix-core/src/movement.rs
```diff
@@ -549,6 +549,85 @@ pub fn goto_treesitter_object(
...
+                    text.byte_to_char(node.start_byte())
+                } else {
+                    end_head
+                }
+            }
+        };
+
+        if movement == Movement::Move {
+            // preserve direction of original range
+            if range.direction() == Direction::Forward {
```
pascalkuthe, line 614: I dont think you need to so this here. We usually dont care about the direction of single width selections (I dont think there is any way those bahve differently right now).

helix-core/src/object.rs
```diff
@@ -1,4 +1,5 @@
 use crate::{Range, RopeSlice, Selection, Syntax};
+
```
pascalkuthe, line 2: Seems like an accidental formatting change slipped in

helix-term/src/commands.rs
```diff
@@ -3280,14 +3286,41 @@ pub mod insert {
...
+                let cursor = range.cursor(doc.text().slice(..));
+                let current_line_num = doc.text().char_to_line(cursor);
+                let current_line_start = doc.text().line_to_char(current_line_num);
+                let left = doc.text().slice(current_line_start..cursor);
+                log::debug!("left: {:?}", left);
+                left.chars().all(|c| c.is_whitespace())
+            });
+
+            if !cursors_after_whitespace {
+                let cmd = match MappableCommand::from_str(cmd) {
```
pascalkuthe, line 3303: U think reparsing the mappable command everytime it's used is not ideal. You can do this during desexualization instead. In fact `MappableCommand` likely already implements desieralize (I didn't check so if it doesn't you can easily add an implementation, that just deserialzes a String and then uses the existing parse implementation).

pascalkuthe (commented):
This is super useful and looks great implementation wise. Just a few smaller comments

helix-core/src/movement.rs
dead10ck: Yeah perhaps not necessary for any functional reason, but my neurotic brain prefers internal consistency. Are you concerned about a performance impact?

helix-term/src/commands.rs
dead10ck: Yeah I think I tried this first, but the problem is `MappableCommand` is in `helix-term` while this config is in `helix-view`. I don't like parsing it every time either, but I'm not really sure how else to do it. I'm open to suggestions.

pascalkuthe: ah the classic problem "commands are in helix-term but I need to use the in helix-view" :D

Hmm maybe since this is more of a special keybinding it could be a special kind of keybinding (the keybinding config lives in `helix-term`).

helix-core/src/movement.rs
pascalkuthe: no I don't think this causes any problems just makes the code a bit more complex than it needs to be but it would also be fine to leave in

helix-term/src/commands.rs
dead10ck: Hm, so I took a look at the key map config:

https://github.com/helix-editor/helix/blob/53f47bc47771c94dab51626ca025be28e62eba0c/helix-term/src/config.rs#L13-L17

As we can see, it is a hash map that maps `Mode`s onto key maps. If we want this config somewhere in there, then I think we'd have two options:

1. Put it in an actual key map, as in

    ```toml
    [keys.insert]
    tab = # do insert tab or "move_parent_node_end"
    ```

    It's not really obvious to me here how to express what we want, without a generalized mechanism to detect any generic situation where you'd want to conditionally do one command or another, depending on the state of the text and the selection.

    I had actually considered the above idea when I first started thinking about this feature, but it seemed overly generic and complicated.
2. Put it one higher up under `keys`

    ```toml
    keys.supertab = "move_parent_node_end"
    ```

    But this doesn't map very well to the current config, which maps modes onto key maps.

Did you have any particular format in mind for moving into a key map config?

Also, I was thinking, currently `MappableCommand::from_str` is doing a linear search through every command. What if instead of the above, I made a static map of the commands, similar to how a map is made for `TypableCommand`?

https://github.com/helix-editor/helix/blob/d3cca0f3571fd30f4d7067a82c02753686297a67/helix-term/src/commands/typed.rs#L2809-L2818

This would get parsing down to O(1). Still not as good as no parsing at all, but significantly better than a needles linear search.

pascalkuthe: I was actually thinking of
```toml
keymap.supertab = ...
```

When I made that comment. I agree that a tab mapping doesn't make sense since this is a very special cased feature and would feel weird to give special meaning for tab here.


However, I don't think `keymap.<mode>` colliding with `keymap.insert` is that weird. If you read a config file:


``` toml

[keymap]
supertab = "...

[keymap.normal]
...

```

It actually looks kind of.intuitive to first have some normal settings and then the various mode mappings and should be pretty easy to document too.


Improving the parsing speed with a lut would definitly be nice but I would.prefer to just avoid the parsing altogether. It's not a huge performance overhead but it just feels a bit hacky to me. And I would personally actually expect a supertab setting in `keymap` and not in `editor`

dead10ck: Ok, I think I agree with all your points. The issue then becomes how to represent the config. Currently it's just a `HashMap<Mode, Keymap>`, so if we want to be able to have regular configuration in there too, then we have two options:

1. Reorganize key mapping config under another common key, e.g.

    ```toml
    [keys]
    supertab = "..."

    [keys.bindings.insert]
    ...
    ```
2. Write a custom parser that special cases the mode strings and throws everything else into the general config.

1 would be a large breaking change, but 2 feels a bit awkward. It's probably the way to go though. @archseer do you want to weigh in before I make the change?

pascalkuthe: I would avoid a breaking change here since this would affect a lot of users. It looks nicer but we should reserve stuff like that for when we move to a script based config. 

Serde has the [flatten attribute](https://serde.rs/field-attrs.html#flatten) exactly for this usecase so it should be pretty straightforward to implement:

``` rust
#[derive(Deserialize)]
struct KeymapConfig {
   supertab: MappableCommand,
   #[serde(flatten)]
   modes: HashMap<String, MappableCommand>
}
```

dead10ck: Oh neat! I didn't know about that attribute, that's a lot simpler than I was thinking. Thanks Pascal, I'll take that route

dead10ck: Ok, I managed to switch this to put the `supertab` config into the `keys` section. It turned out to be more complicated than just using the `#[serde(flatten)]` attribute: I had to add the new keymap config object to a couple of context objects, but this may end up being useful in the future anyway, if we ever add more to this config section.

Let me know what you think.

pascalkuthe (approved):
LGTM 👍

helix-term/src/commands.rs
pascalkuthe: Ah I didn't think of that. It's not that much extra code tough and might be useful in the future too so I think that's ok 👍

dpc:
How to use it with LSP-completion enabled? I'm used to pressing tab to cycle through completion, but now it jumps out of the `parent_node_end`. Ideally I'd like to jump out only if there are no completions shown.

dead10ck:
> 

Yeah, I mentioned this in the PR description. Unfortunately, when auto complete is on, having tab do completion cycling breaks this feature, since any time you need to do tab multiple times, you'll do it once, then an auto complete pops up, and you have to press esc to keep supertabbing. Unfortunately I can't think of a way around this; the bindings simply conflict with one another.

dead10ck:
Actually I got a suggestion from a maintainer; I'll see if instead of overriding the key binding if we can just avoid triggering auto completion when we run the command.

gabydd:
That's what I was thinking as well and it seems like a good solution to me. It also made me think of the ideas of https://github.com/helix-editor/helix/pull/5803 and https://github.com/helix-editor/helix/pull/6910 which might be relevant to this.

pascalkuthe:
> That's what I was thinking as well and it seems like a good solution to me. It also made me think of the ideas of #5803 and #6910 which might be relevant to this.

as I said in both of those PRs the solutions there are kind of hacky. IMO this PR should reset the idle timeout (or some other small workaround) after jumping. I have an event handling system + completions based on it locally. I was just waiting for the multi-lsp (and other completion-related PRs) to get merged first.

I didn't have time to finish that work since but that would automatically fix the problem @dead10ck described.

dead10ck:
> Just to clarify: For me supertab works even if auto completion showed up already, so I can't use auto-completion when it's enabled.

You can, you just can't use tab. You can also use C-n/C-p and arrow keys to navigate menu items. But yeah, ideally we don't take any bindings out if we can avoid it.

> BTW. Is it possible to remap super-tab ... to something other than tab? :D

You can map `move_parent_node_end` and friends to anything you want, but for the supertab behavior where it conditionally either inserts a tab or runs another command, the current implementation is built into the `insert_tab` command, which technically need not be mapped to the tab key

>It makes me think if it could be just a modifier. So there's `tab` but there's also `L+tab` (for "leading whitespace") and `N+tab` (for "no leading whitespace"). Than any keys could be mapped to act supertab-like.

Yeah I had considered if it would be possible to make something more generic than just the `insert_tab` command, but I think this would be impossible or at least very difficult and complex to express with a TOML config. We would need a generic system that can detect potentially any arbitrary state in the text and choose to conditionally run one command or another. I think this will be much easier to do once we have script-based configs and/or a plug-in system.

dead10ck:
Ok actually, now that I turned that remapping off, I'm rediscovering that the problem is bigger than just the auto complete menu being triggered on multiple invocations. Because of https://github.com/helix-editor/helix/pull/6910, even after simply turning on insert mode, the auto completion menu pops up, which prevents even the first supertab from working.

Should I also clear the idle timer on entering insert mode as a quick workaround until we can do something like @pascalkuthe suggests in https://github.com/helix-editor/helix/pull/6910#issuecomment-1529111935, or a general purpose event hook system is in place?

pascalkuthe:
hmm that depends on whether we want to show autocmpletions immidietly when entering insert mode. I am actually leaning more towards only triggering completions on actual edits (of word + and trigger chars) which is sort of how the LSP standard is intended, If we do go with that in the future then it makes sense to slightly preemt that change here and reset the idle timeout when entering insert mode. But that might be a controversial design choice so we might want to expose some knobs in the config. Not sure I would lean towards only requesting completions on actual edits since that's that LS are designed around

dead10ck:
Ok, I removed the binding override and instead clear the idle timer on entering insert mode, and for all of the `move_parent_node_*` commands. This fixes the issue and keeps the tab binding for auto completions.

This means any users expecting completions when they first enter insert mode may be surprised by this, but hopefully it's a small number of users. I think most editors don't trigger auto complete until the user actually types something.

pascalkuthe (approved):

dpc:
Sweet. I tried it, auto-completion issues seem gone and I love it.

archseer:
Should we set 

```toml
supertab = "move_parent_node_end"
```

as default?

dead10ck:
Oh sorry I missed these comments.

> Should we set
> 
> ```toml
> supertab = "move_parent_node_end"
> ```
> 
> as default?

I was hesitant to turn this on by default, since it does mean that you can't insert a tab in the middle of text while this feature is turned on.

Binding to anything but plain tab kind of defeats the point of supertab, which is to dynamically choose what behavior the tab key has. If it's bound to `S-tab`, then you don't really need supertab at all, you could just add

```toml
[keys.insert]
C-tab = "move_parent_node_end"
```

Which users still have the option of doing, and we could certainly add that as a default binding.

On the other hand, having `C-tab` *always* insert a tab no matter what could be a useful escape hatch. Though in order to do that, it might be better to change this so that supertab has its own command, rather than putting this in the `insert_tab` command. That way, `insert_tab` doesn't change, it just always does what it says, and you'd have to add a binding for supertab. e.g. we could have a default of:

```toml
[keys.supertab]
command = "move_parent_node_end"

[keys.insert]
tab = "supertab"
C-tab = "insert_tab"
```

Only thing with this is handling the tab key supertab / menu conflict will get even weirder: it might not be the tab key that someone binds to supertab.

dead10ck:
Also, I added a config option to let the user decide if they want supertab to take precedence over menu item cycling. It changed the config format slightly; I updated the docs.

dead10ck:
Ok, I split supertab out into its own dedicated command, and rebound tab to `supertab`, and C-tab to `insert_tab`. This gives users an escape hatch to always insert a tab character even when supertab is enabled by pressing C-tab. For users that don't have supertab enabled, the `supertab` command just calls `insert_tab`, so they won't notice a difference.

pascalkuthe (dismissed):

dpc:
Rebased on top of the latest version (7407f03921), fixed my config, seems to work as it did before. Will report any issues if I find any.

David-Else:
> > Should we set
> > ```toml
> > supertab = "move_parent_node_end"
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > as default?
> 
> I was hesitant to turn this on by default, since it does mean that you can't insert a tab in the middle of text while this feature is turned on.

Now that the user can use C-tab to bypass why not turn on `supertab = "move_parent_node_end"` as the default? Don't other editor have this as a default, how about VS Code?

book/src/configuration.md
```diff
@@ -347,3 +347,58 @@ max-wrap = 25 # increase value to reduce forced mid-word wrapping
...
+
+[Key remapping]: ./keymap.md
+
+### `[keys.supertab]` Section
+
+
+| Key        | Description | Default |
+|------------|-------------|---------|
+| `command` | If set to a command name, then when the cursor is in a position with non-whitespace to its left, instead of inserting a tab, it will run the given command. If there is only whitespace to the left, then it inserts a tab as normal. One can emulate a tabout behavior with this with the `move_parent_node_end` command. See [Tabout Example] | `None` |
+| `supercede_menu` | Normally, when a menu is on screen, such as when auto complete is triggered, the tab key is bound to cycling through the items. This means when menus are on screen, one cannot use the tab key to trigger the supertab command. If this option is set to true, the supertab command always takes precedence, which means one cannot use the tab key to cycle through menu items. One of the other bindings must be used instead, such as arrow keys or `C-n`/`C-p`. | `None` |
```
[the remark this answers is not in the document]
the-mikedavis: Ideally I think this could be covered by #5505. You could having a binding for the completion menu:

```toml
[keys.menu.completion]
tab = "supertab" # default would be "move_next_item" or whatever that menu command would be called
```

the-mikedavis, line 365: Is it possible to move this under regular config, maybe as a `editor.keymap.supertab` or just `editor.supertab`? It looks like a lot of the changes in this PR are caused by the refactor for KeymapConfig. I think we should always prefer having multiple versions of a command to having config options affect a command and I'm worried that the KeymapConfig creates a precedence for adding more config options in the future.

I see it takes a MappableCommand which would be a blocker because it currently lives in helix-term (until #5581) but I think that taking a command name is unnecessary: it could be an enum of options for the different kind of motions implemented (`move_parent_node_end`). It seems like there would only be a few commands that you would ever want to configure for this option anyways.

With what I have in mind here, the `tab` in the default keymap would be `insert_tab` if we want tabout disabled by default and `supertab` if we want it enabled by default.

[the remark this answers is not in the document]
dead10ck: I agree, the ideal solution is just being able to rebind the key in the config, but this currently isn't possible. And at least for me, being able to have supertab take precedence is necessary for it to be useful at all while auto complete is on at the same time.

dead10ck: > Is it possible to move this under regular config, maybe as a `editor.keymap.supertab` or just `editor.supertab`? It looks like a lot of the changes in this PR are caused by the refactor for KeymapConfig. I think we should always prefer having multiple versions of a command to having config options affect a command and I'm worried that the KeymapConfig creates a precedence for adding more config options in the future.

Hmm, so this is actually what I started with, but through discussion, it ended up changing to what it is now.

https://github.com/helix-editor/helix/pull/4443#discussion_r1192868909

When I think about the command set we have now and the precedence of having a separate command that has slightly different behavior, like `goto_line_end` vs `goto_line_end_newline`, or `change_selection` vs `change_selection_noyank`, the differences in behavior are isolated to themselves and I think generally make sense to have independent commands to enable both forms of behavior at once in one key map.

But in this case, the behavior of the feature of itself is not changed—it's kind of an odd case where it interacts with a UI element overriding the user's ability to invoke the feature. Having a `supertab` / `supertab_supercede_menu` command split seems inappropriate for the problem we're trying to solve, and also I'm not sure how it would even be implemented, since we'd need to know what command was invoked during the key event handling in order to know what the tab key should do when a menu is up.

> I see it takes a MappableCommand which would be a blocker because it currently lives in helix-term (until #5581) but I think that taking a command name is unnecessary: it could be an enum of options for the different kind of motions implemented (`move_parent_node_end`). It seems like there would only be a few commands that you would ever want to configure for this option anyways.

Hmm. The original goal of making it an arbitrary command name was for flexibility, both from the user's perspective, as well as any future uses of supertab, such as snippet placeholder cycling. We'd have to essentially keep an enum that would always be a subset of the command list, and any other command we want it to be able to do would have to require a code change.

That said, realistically, the concrete end goal was really just tabout. Some users might rebind it to something like `expand_selection`, `search_next`, `goto_definition` or perhaps others, but I can't say I believe that will be common.

If this is what we do though, is an enum what we want now? If `move_parent_node_end` is the only option, should it just be hardcoded until we add more options?

> With what I have in mind here, the `tab` in the default keymap would be `insert_tab` if we want tabout disabled by default and `supertab` if we want it enabled by default.

Well the idea of making tab = supertab was that if users want to enable it, they don't also have to rebind the tab key to supertab. Currently, if supertab is disabled, the supertab command just immediately calls the `insert_tab` command.

dead10ck:
Considering @the-mikedavis 's comments, it seems generally speaking, there are two preexisting and independent issues that would have an effect on this PR:

* #5505 would solve the whole issue of the tab key conflict in menus. The config option to override the menu cycling would not need to exist, and the key event handling would not need to care about the supertab feature, as should be the case.
  Currently there is no PR to fix this issue that I'm aware of, so it's unclear how long the supertab config option will need to exist in the meantime, but generally speaking, it would be a temporary workaround until this work is done.
* #5581 would let us put the supertab config back in the editor config section without having to parse the command on every invocation at runtime. This PR currently exists and is I believe just awaiting review. We could probably reasonably wait on this PR if we feel strongly about moving the supertab config back into editor.

dead10ck:
Regarding default bindings, it looks like `C-tab` is the same as `tab`

https://sw.kovidgoyal.net/kitty/keyboard-protocol/#legacy-functional-keys

Considering this, we'll have to decide what makes the most sense in the defaults: the `insert_tab` escape hatch or `move_parent_node_end`. If we want supertab on by default, then using it for `insert_tab` would probably make the most sense, but otherwise using it for `move_parent_node_end` would be better I think.

Two other default bindings I had considered for `move_parent_node_end` and `start` were `A-e` and `A-b` to go with the other TS bindings. Those are currently unused in any mode. If we go with that, then the default config would become

```toml
[keys.supertab]
command = "move_parent_node_end"
supercede_menu = false

[keys.insert]
tab = "supertab"
S-tab = "insert_tab"

[keys.select]
A-e = "extend_parent_node_end"
A-b = "extend_parent_node_start"
```

Any opinions?

pascalkuthe:
I would just bind a-e and a-b everywhere by default to be consistent. People can enable the more ergonomic supertab mapping in their config if they have protocol support. It's pretty easy to configure

David-Else:
I just tried to test this and the pickers won't work. They appear, but the focus stays on the document and all keyboard input continues as if you were focused on the document.

The only difference I have in my version is this little mod:

```rs
    // pub fn inner_area(&self, doc: &Document) -> Rect {
    //     self.area.clip_left(self.gutter_offset(doc)).clip_bottom(1) // -1 for statusline
    // }

    pub fn inner_area(&self, doc: &Document) -> Rect {
        // zen mode with wrap and text_width
        let config = doc.config.load();
        let self_width = self.inner_width(doc);
        let text_width = doc
            .language_config()
            .and_then(|config| config.text_width)
            .unwrap_or(config.text_width) as u16;
        let soft_wrap_at_text_width = doc
            .language_config()
            .and_then(|config| {
                config
                    .soft_wrap
                    .as_ref()
                    .and_then(|soft_wrap| soft_wrap.wrap_at_text_width)
            })
            .or(config.soft_wrap.wrap_at_text_width)
            .unwrap_or(false);

        let space_width = if soft_wrap_at_text_width && self_width.gt(&text_width) {
            (self_width - text_width) / 2
        } else {
            0
        };

        self.area
            .clip_left(self.gutter_offset(doc) + space_width)
            .clip_right(space_width)
            .clip_bottom(1) // -1 for statusline
    }
```

dead10ck:
Ok, after some discussion, the command has been renamed to "smart tab" to avoid confusion with the super key.

The config has also moved back to the `editor` section, and until we have access to `MappableCommand` in `helix-term`, the ability to set an arbitrary command has been removed; for now, it will just hardcode running `move_parent_node_end`.

dead10ck:
I'm not sure why the lint is failing. My local `cargo fmt --all` changes nothing.

helix-view/src/editor.rs
```diff
@@ -704,7 +722,7 @@ pub struct WhitespaceCharacters {
 impl Default for WhitespaceCharacters {
     fn default() -> Self {
         Self {
-            space: '·',    // U+00B7
+            space: '·',   // U+00B7
```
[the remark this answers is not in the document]
dead10ck: `cargo fmt` actually deleted a space here to line it up with the rest of the comments. I don't know if maybe there's some version mismatch causing issues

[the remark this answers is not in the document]
Philipp-M: Yeah this happens for me all the time with a nightly rust version, now it got even "worse" since finally let-else is formatted in nightly (maybe even in the latest stable, I don't know). I wonder whether it makes sense to use a (stable) up-to-date rustfmt while otherwise keeping the MSRV version, though this may unnecessarily complexify stuff.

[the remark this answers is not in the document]
pascalkuthe: you are using rustfmt installed via rustup it should automatically use the right rustfmt version (the one from rust 1.65) already I think. Its pinned to 1.65 in the `rust-toolchain.toml` file. That's the version that will be used by CI too and should be used during development. We could pin a newer version but I don't think that's really worth the effort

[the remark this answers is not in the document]
Philipp-M: Yeah I agree, though it may make sense to reformat the code (without this particular change we are commenting on) with the new let-else formatting feature, since the MSRV version shouldn't care anyway as it doesn't format let-else code.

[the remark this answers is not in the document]
pascalkuthe: yeah let..else in particular would be reasonable, basically our policy has been: we accept PRs with fixes for clippy lints/fmt changes from nightly as long as they don't break stable CI

the-mikedavis (commented):
I've been using this locally for a while and I really like it! It was especially nice when working on #7763 since Nix has the semicolons after `}`s. I have a small nit and I think some older outstanding ones about resetting some unrelated whitespace changes but once those are resolved let's get this merged!

(The diff is mostly large because of the additions to the integration test. With the recent changes hard-coding to `move_parent_node_end`, this is a nice and relatively small change :+1:)

helix-term/src/keymap/default.rs
```diff
@@ -368,7 +368,8 @@ pub fn default() -> HashMap<Mode, KeyTrie> {
         "C-h" | "backspace" | "S-backspace" => delete_char_backward,
         "C-d" | "del" => delete_char_forward,
         "C-j" | "ret" => insert_newline,
-        "tab" => insert_tab,
+        "tab" => smart_tab,
+        "S-tab" => insert_tab,
```
pascalkuthe, line 372: one more minor ching from my side, I thaught we wanted to add `a-e` and `a-b` default bindings as well? There was a lot of back and forth so maybe I forgot (sorry in that case) . Espeically in select mode I find these bindings quite usefel

dead10ck: Oh yeah, I forgot about that. I'll add that too

dead10ck:
Ok I think I got all the comments 🙂

dpc:
For reference:

```
[keys.normal]
tab = "move_parent_node_end"
S-tab = "move_parent_node_start"

[editor.smart-tab]
enable = true

[keys.insert]
S-tab = "move_parent_node_start"

[keys.select]
tab = "extend_parent_node_end"
S-tab = "extend_parent_node_start"
```

BTW. I might have missed it, but it might be worthwhile documenting the more elaborated setup somewhere in the docs.

David-Else:
@dpc I have:
```
[keys.normal]
"tab" = "expand_selection"
"S-tab" = "shrink_selection"
```
No need for any extra select mode bindings as it selects already. I am not sure about your `editor.smart-tab.enable = "full"` idea.

dpc:
> No need for any extra select mode bindings as it selects already.

How does it work in selection mode? (I haven't tried, but I'm curious what exactly is responsible for it working, more than the effect)

I have this feature enabled and configured since it was a PR, so I might have missed something, but when I google and look through the docs I only find https://docs.helix-editor.com/master/configuration.html#editorsmart-tab-section that doesn't explain (or advertise) this feature much at all, and doesn't seem to enable any extra binding that made a big difference when I was first trying it out.

dead10ck:
> @dpc I have:
> ```
> [keys.normal]
> "tab" = "expand_selection"
> "S-tab" = "shrink_selection"
> ```
> No need for any extra select mode bindings as it selects already. I am not sure about your `editor.smart-tab.enable = "full"` idea.

Not really sure what this has to do with @dpc 's question. These bindings aren't for smart tab, and the bindings they're asking about aren't in the default key map.

@dpc the reason this isn't in the default key map is because it needs a terminal with support for the Kitty keyboard protocol to work, and we don't want to require that in the default bindings.

But smart tab is enabled by default, so you don't need that part.

dead10ck:
> `editor.smart-tab.enable = "full"`

We've also been disinclined to make configuration "kits" like this because they require special case logic and have confusing semantics with respect to overrides.

dpc:
> @dpc the reason this isn't in the default key map is because it needs a terminal with support for the Kitty keyboard protocol to work, and we don't want to require that in the default bindings.

I'm confused - seems to me like all I'm binding is `tab` and `S-tab`. Is it the `S-tab` being not a valid keycode+modifier combo or something?

Anyhow, would it make sense to expand https://docs.helix-editor.com/master/configuration.html#editorsmart-tab-section with recommended settings or something?

dead10ck:
> > @dpc the reason this isn't in the default key map is because it needs a terminal with support for the Kitty keyboard protocol to work, and we don't want to require that in the default bindings.
> 
> I'm confused - seems to me like all I'm binding is `tab` and `S-tab`. Is it the `S-tab` being not a valid keycode+modifier combo or something?

Yeah, in standard terminals, Shift + Tab does not work. Also, tab is ambiguous with Ctrl-i, which is the binding for going forward in the jump history, so binding to tab in normal mode overwrites that binding.

> Anyhow, would it make sense to expand https://docs.helix-editor.com/master/configuration.html#editorsmart-tab-section with recommended settings or something? Hmmm... are even the settings i have recommended? :D . I just copied them from some previous comment in the past and it worked great.

Sure, we'd appreciate a PR 🙂

dpc:
@dead10ck #9762 , happy to tweak to improve.

PR #4449, opened 2022-10-24
 Popup scrollbar

### LINKED ISSUE #1248: Hover page up down hint

Manosmer:
Closes #1248 by implementing a scrollbar as the required hint. As mentioned by @the-mikedavis in #4364. This is a simple copy of the scrollbar functionality from `Menu`. The scrollbar is enabled by default and `Popup::with_scrollbar()` can be used to toggle it on/off.

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -356,19 +356,12 @@ impl<T: Item + 'static> Component for Menu<T> {
         let fits = len <= win_height;
 
         let scroll_style = theme.get("ui.menu.scroll");
-        for (i, _) in (scroll..(scroll + win_height).min(len)).enumerate() {
-            let cell = &mut surface[(area.x + area.width - 1, area.y + i as u16)];
+        if !fits {
+            // Draw scroll thumb
+            for i in scroll_line..(scroll_line + scroll_height) {
+                let cell = &mut surface[(area.x + area.width - 1, area.y + i as u16)];
 
-            if !fits {
-                // Draw scroll track
                 cell.set_symbol("▐"); // right half block
-                cell.set_fg(scroll_style.bg.unwrap_or(helix_view::theme::Color::Reset));
-            }
-
-            let is_marked = i >= scroll_line && i < scroll_line + scroll_height;
-
-            if !fits && is_marked {
-                // Draw scroll thumb
                 cell.set_fg(scroll_style.fg.unwrap_or(helix_view::theme::Color::Reset));
             }
         }
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -356,13 +356,20 @@ impl<T: Item + 'static> Component for Menu<T> {
         let fits = len <= win_height;
 
         let scroll_style = theme.get("ui.menu.scroll");
+        let mut cell;
         if !fits {
-            // Draw scroll thumb
-            for i in scroll_line..(scroll_line + scroll_height) {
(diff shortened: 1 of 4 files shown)


Manosmer:
Worth mentioning: I found a bug when the popup carries a `SignatureHelp`. In cases where the viewport is small enough, the Popup renders outside of it and scrolling cannot proceed all the way down.

helix-term/src/ui/popup.rs
```diff
@@ -128,6 +130,14 @@ impl<T: Component> Popup<T> {
         }
     }
 
+    /// Enables the Popup's scrollbar.
+    /// Consider disabling the scrollbar in case the child
+    /// already has its own.
+    pub fn with_scrollbar(mut self, enable_scrollbar: bool) -> Self {
+        self.has_scrollbar = enable_scrollbar;
+        self
+    }
```
the-mikedavis, line 139: I think the completion and code-actions popups will need to have this disabled since their components are menus

helix-term/src/ui/popup.rs
```diff
@@ -228,6 +238,41 @@ impl<T: Component> Component for Popup<T> {
...
+                    cell.set_symbol("▐"); // right half block
+                    cell.set_fg(scroll_style.bg.unwrap_or(helix_view::theme::Color::Reset));
+                }
+
+                let is_marked = i >= scroll_line && i < scroll_line + scroll_height;
+
+                if !fits && is_marked {
+                    // Draw scroll thumb
+                    cell.set_fg(scroll_style.fg.unwrap_or(helix_view::theme::Color::Reset));
+                }
```
pickfire, line 273: Why not move this into `!fits` block?

Manosmer: Basically, the sumbitted PR is just a copy of the already implemented menu scrollbar. You are right, though. The code could use some optimisation. I'm on it!

helix-term/src/ui/menu.rs
```diff
@@ -356,19 +356,12 @@ impl<T: Item + 'static> Component for Menu<T> {
...
-                // Draw scroll track
                 cell.set_symbol("▐"); // right half block
-                cell.set_fg(scroll_style.bg.unwrap_or(helix_view::theme::Color::Reset));
-            }
-
-            let is_marked = i >= scroll_line && i < scroll_line + scroll_height;
-
-            if !fits && is_marked {
-                // Draw scroll thumb
                 cell.set_fg(scroll_style.fg.unwrap_or(helix_view::theme::Color::Reset));
```
pickfire, line 365: Don't we need to check for `is_marked` for this?

pickfire, line 365: Don't we need to check for `is_marked` for this?

Manosmer: Basically, I have integrated the `is_marked` expression inside the for-loop condition, so that I only draw the cells of the scrollbar. However, I did not realise that the other cells are used to render the scrollbar track, since my theme does not set the "ui.menu.scroll" key. I fixed it in the last commit

the-mikedavis (approved):
I've been using this for a bit and it works great!

PR #4450, opened 2022-10-24
fix: repeating repeat operator

### LINKED ISSUE #4359: Make "repeat last insert" respect the number prefix
With the current implementation, when you press `<n>.`, the last insert is repeated only once, completely ignoring the number `<n>`, unlike in vim where it's indeed inserted `<n>` times.

This feature is useful for inserting repeated characters. I currently work around this issue by recording the insert as a macro, and do `<n>q` instead, but this isn't very efficient.

I would like to implement this feature, but would love to get a green light from the core team first just in case the current behavior is actually intended. Thanks!

kosciej:
Fixes #4359

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -25,7 +25,7 @@ use helix_view::{
     keyboard::{KeyCode, KeyModifiers},
     Document, Editor, Theme, View,
 };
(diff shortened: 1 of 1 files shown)


dead10ck (approved):
This looks fine to me

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #433, opened 2021-07-11
Custom Diffing Implementation


kirawi:
Currently, we depend upon `similar` to provide the diffing we need for `:reload` and in the future, file comparison (#405). However, this comes with a few drawbacks: (a) it can only operate on contiguous data, and (b) it is too expensive to calculate on moderately sized, extremely different files. A big part of that is our need to temporarily convert the `Rope` representation of the text into a `String` as `Rope`s are not a contiguous data structure. A [histogram](https://stackoverflow.com/a/32367597) implementation and/or incorporating SIMD may also improve real-world performance, but it would not affect the time complexity of the algorithm which is currently `O(n + (N+M)D)`.

Our main priority should be to investigate an alternative algorithm for `:reload` that can efficiently calculate a diff without human readability in mind, `rsync`, and how we can adapt these algorithms to be able to work on a `Rope` directly. I suspect that implementing the `rsync` algorithm would be the easiest one to tackle.

This is a difficult solution to a difficult problem, but it would be nice to have.

kirawi:
https://docs.rs/xi-rope/latest/xi_rope/diff/index.html

kirawi:
Will close in favor of #4457. That implementation is significantly faster for expected workloads.

PR #4457, opened 2022-10-24
Significantly improve performance of `:reload`

### LINKED ISSUE #4228: The reload operation takes up a lot of memory until the computer freezes
### Summary

Opening a file with a size of 500M, executing the reload operation will take up much more memory than the file size and keep increasing until all the memory is used up

### Reproduction Steps

 

I tried this:

1. `hx 1.txt` 
2.  `reload`
I expected this to happen:
reload 1.txt
Instead, this happened:
hx used up all memory

### Helix log

```
```

</details>

### Platform

macOS

### Terminal Emulator

zsh 5.8.1

### Helix Version

helix 22.08.1

pascalkuthe:
Closes #4228 and #433

When `:reload` is called helix currently uses `similar` to compare
the on-disk file with the current state, to create a changeset that can be added to the history.
Similar implements a basic version of myers algorithm.
However, myers algorithm without additional enhancements quickly behaves quadratic over the filesize.
The implementation in `similar` also has many inefficiencies which lead to huge memory consumption as reported in #4228.
The situation is worsened by the fact that `similar` requires random access to the collections it is diffing.
This requires us to convert the ropes to a `String` (essentially duplicating the entire file in memory) before performing the diff.
Finally, helix used a **char** diff over the whole file.
While this produces `ChangeSet`s this leads to extremely poor performance due to the large number of tokens (minimal changes).

Other popular tools that use myers algorithm (GNU diff and git for example) remedy this problem using a couple techniques:

* perform a line diff
* intern all lines before performing the diff (makes comparisons `O(1)`` instead of `O(n)`, which are one of the main operations during myer diff)
* preprocess the file by removing lines that never occur in the other file
* abort the snake search using two heuristics (one that aborts if a "good enough" snake is found and a second that uses the best snake after a hard iteration limit)

With these optimizations myers diff shows very good performance even for large files/diffs.
I have ported gits implement of myers algorithm to rust and published it as the [`imara-diff`](https://github.com/pascalkuthe/imara-diff) crate.
Furthermore, I have also ported gits implementation of a histogram diff to this crate.
This algorithm provides more human-readable output similar to patience diff while performing on par with (or better than) myers algorithm.
Compared to myers diff, the version in git wasn't actually all that optimized.
After the initial port I further optimized my histogram implementation, the final version beats myers algorithm for every line-diff I have tested by 5%-100%.
The repo contains detailed benchmarks and comparisons between the two algorithms and `similar`.
Note that the benchmarks shown there are performed on source code (linux kernel, rustc) and not on huge (500MB+) files like #4228.

This PR changes the `compare-ropes` implementation in `helix-core` to use `imara-diff` instead of `similar`.
The implementation performs a `line-diff` using the `histogram` algorithm as it generally performs better.
A char diff with myers algorithm is then performed on each changed hunk, because the `histogram` heuristic does not work well for char-diffs.
This ensures that the generated change sets remain reasonably small while offering much better performance.

I have stress tested this PR by generating a `1 GB` file with:

``` shell
base64 /dev/urandom | head -c 1000000000 > file.tx
```

Afterwards I performed multiple tests by editing/deleting/copying in random places and then reloading.
Another test I performed was repeating the above command while helix was open and then reloading.

All tests made helix spin forever on master (and use up to 50GB of RAM before being killed).
With this PR helix only took a few seconds (up to 20 seconds for complete replacement) and 3.6GB peak RAM usage.

## Alternative Approaches

@kirawi suggested the `rsync` algorithm or the [`xi_rope` algorithm](https://docs.rs/xi-rope/latest/xi_rope/diff/index.html), which is somewhat similar.
While these algorithms are faster they do not fit within helix without significant changes, because they provide a fundamentally different output:
A sequence of insertions and **random copies** from the base document.
For example consider the following diff:

```
foo
foo
bar
```

```
foo
bar
foo
```

An (ideal) rsync algorithm would produce a diff like this:

```
Copy(0) // foo  
Copy(2) // bar
Copy(0) // foo
```

By comparison a diff algorithm will produce:

```
Retain(1) //foo
Delete(1) //remove foo
Retain(1) //bar
Insert("foo")
```

This kind of output can not be converted to the (reversible) edit-sequences helix uses.
It might be feasible to refactor helix to store the undo/redo transactions separately in a format more akin to the output of rsync.
However, this would require a large refactor of a core component (`Transaction`) that would touch many parts of the editor.

This PR only requires minimal changes while providing large performance wins

### DIFF
--- a/helix-core/Cargo.toml
+++ b/helix-core/Cargo.toml
@@ -38,7 +38,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 toml = "0.5"
 
-similar = "2.2"
+imara-diff = "0.1.0"
 
 encoding_rs = "0.8"
 
--- a/helix-core/Cargo.toml
+++ b/helix-core/Cargo.toml
@@ -17,7 +17,7 @@ integration = []
 [dependencies]
 helix-loader = { version = "0.6", path = "../helix-loader" }
 
-ropey = { version = "1.5", default-features = false, features = ["simd"] }
+ropey = { version = "1.5.1-alpha", default-features = false, features = ["simd"] }
 smallvec = "1.10"
 smartstring = "1.0.1"
 unicode-segmentation = "1.10"
--- a/helix-core/src/diff.rs
+++ b/helix-core/src/diff.rs
@@ -1,57 +1,194 @@
-use crate::{Rope, Transaction};
+use std::ops::Range;
+use std::time::Instant;
+
+use imara_diff::intern::InternedInput;
+use imara_diff::Algorithm;
+use ropey::RopeSlice;
+
+use crate::{ChangeSet, Rope, Tendril, Transaction};
+
+/// A `imara_diff::Sink` that builds a `ChangeSet` for a character diff of a hunk
+struct CharChangeSetBuilder<'a> {
+    res: &'a mut ChangeSet,
+    hunk: &'a InternedInput<char>,
+    pos: u32,
+}
+
+impl imara_diff::Sink for CharChangeSetBuilder<'_> {
+    type Out = ();
+    fn process_change(&mut self, before: Range<u32>, after: Range<u32>) {
+        self.res.retain((before.start - self.pos) as usize);
+        self.res.delete(before.len());
+        self.pos = before.end;
+
+        let res = self.hunk.after[after.start as usize..after.end as usize]
+            .iter()
+            .map(|&token| self.hunk.interner[token])
+            .collect();
+
+        self.res.insert(res);
+    }
+
+    fn finish(self) -> Self::Out {
+        self.res.retain(self.hunk.before.len() - self.pos as usize);
+    }
+}
+
+struct LineChangeSetBuilder<'a> {
+    res: ChangeSet,
+    after: RopeSlice<'a>,
+    file: &'a InternedInput<RopeSlice<'a>>,
+    current_hunk: InternedInput<char>,
+    pos: u32,
+}
+
+impl imara_diff::Sink for LineChangeSetBuilder<'_> {
+    type Out = ChangeSet;
+
+    fn process_change(&mut self, before: Range<u32>, after: Range<u32>) {
+        let len = self.file.before[self.pos as usize..before.start as usize]
+            .iter()
+            .map(|&it| self.file.interner[it].len_chars())
+            .sum();
+        self.res.retain(len);
+        self.pos = before.end;
+
+        // do not perform diffs on large hunks
+        let len_before = before.end - before.start;
+        let len_after = after.end - after.start;
+
+        // Pure insertions/removals do not require a character diff.
+        // Very large changes are ignored because their character diff is expensive to compute
+        // TODO adjust heuristic to detect large changes?
+        if len_before == 0
+            || len_after == 0
+            || len_after > 5 * len_before
+            || 5 * len_after < len_before && len_before > 10
+            || len_before + len_after > 200
+        {
+            let remove = self.file.before[before.start as usize..before.end as usize]
+                .iter()
+                .map(|&it| self.file.interner[it].len_chars())
+                .sum();
+            self.res.delete(remove);
+            let mut fragment = Tendril::new();
+            if len_after > 500 {
+                // copying a rope line by line is slower then copying the entire
+                // rope. Use to_string for very large changes instead..
+                if self.file.after.len() == after.end as usize {
+                    if after.start == 0 {
+                        fragment = self.after.to_string().into();
+                    } else {
+                        let start = self.after.line_to_char(after.start as usize);
+                        fragment = self.after.slice(start..).to_string().into();
+                    }
+                } else if after.start == 0 {
+                    let end = self.after.line_to_char(after.end as usize);
+                    fragment = self.after.slice(..end).to_string().into();
+                } else {
+                    let start = self.after.line_to_char(after.start as usize);
+                    let end = self.after.line_to_char(after.end as usize);
+                    fragment = self.after.slice(start..end).to_string().into();
+                }
+            } else {
+                for &line in &self.file.after[after.start as usize..after.end as usize] {
+                    for chunk in self.file.interner[line].chunks() {
+                        fragment.push_str(chunk)
+                    }
+                }
+            };
+            self.res.insert(fragment);
+        } else {
+            // for reasonably small hunks, generating a ChangeSet from char diff can save memory
+            // TODO use a tokenizer (word diff?) for improved performance
+            let hunk_before = self.file.before[before.start as usize..before.end as usize]
+                .iter()
+                .flat_map(|&it| self.file.interner[it].chars());
+            let hunk_after = self.file.after[after.start as usize..after.end as usize]
+                .iter()
+                .flat_map(|&it| self.file.interner[it].chars());
+            self.current_hunk.update_before(hunk_before);
+            self.current_hunk.update_after(hunk_after);
+
+            // the histogram heuristic does not work as well
+            // for characters because the same characters often reoccur
+            // use myer diff instead
+            imara_diff::diff(
+                Algorithm::Myers,
+                &self.current_hunk,
+                CharChangeSetBuilder {
+                    res: &mut self.res,
+                    hunk: &self.current_hunk,
+                    pos: 0,
+                },
+            );
+
+            self.current_hunk.clear();
+        }
+    }
+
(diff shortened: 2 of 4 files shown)


helix-core/src/diff.rs
```diff
@@ -1,58 +1,188 @@
...
+    fn estimate_tokens(&self) -> u32 {
+        // we can provide a perfect estimate which is very nice for performance
+        self.0.len_lines() as u32
+    }
+}
+
+/// Compares `old` and `new` to generate a [`Transaction`] describing
+/// the steps required to get from `old` to `new`.
+pub fn compare_ropes(before: &Rope, after: &Rope) -> Transaction {
+    log::error!("diff");
```
archseer, line 160: Stray log

helix-view/src/editor.rs
```diff
@@ -1093,7 +1093,9 @@ impl Editor {
         let id = if let Some(id) = id {
             id
         } else {
+            log::error!("open");
             let mut doc = Document::open(&path, None, Some(self.syn_loader.clone()))?;
+            log::error!("done");
```
archseer, line 1099: Here too

helix-core/src/diff.rs
```diff
@@ -1,58 +1,188 @@
...
+        self.res.retain(len);
+        self.res
+    }
+}
+
+struct RopeLines<'a>(RopeSlice<'a>);
+
+impl<'a> imara_diff::intern::TokenSource for RopeLines<'a> {
+    type Token = RopeSlice<'a>;
+    // TODO: improve performance of lines iterator
```
archseer, line 144: Maybe leave a link to the upstream issue? https://github.com/cessen/ropey/issues/25

helix-core/src/diff.rs
```diff
@@ -1,58 +1,188 @@
...
+                } else {
+                    let start = self.after.line_to_char(after.start as usize);
+                    let end = self.after.line_to_char(after.end as usize);
+                    fragment = self.after.slice(start..end).to_string().into();
                 }
-            }),
-    )
+            } else {
+                for &line in &self.file.after[after.start as usize..after.end as usize] {
+                    write!(&mut fragment, "{}", self.file.interner[line]).unwrap();
```
archseer, line 96: could just use `.push_str`?

archseer (commented):
Nice work! Do you have a flamegraph of the new algorithm available?

helix-core/src/diff.rs
pascalkuthe: The lines are a `RopeSlice` so we can't just call `push_str` directly so thats why I used format.
I have switched to using `chunks` to get string that we can push with `push_str`. This is essentially the same thing `Display` did but it might be a bit more performant by avoiding the format machinery

pascalkuthe:
> Nice work! Do you have a flamegraph of the new algorithm available?

The below flamegraph is for reloading a 1GB file with a single line removed.
As you can see most of the time is spent interning the input lines.
For non-trivial source-code diff (larger diffs, smaller files) the diff itself usually takes more time (here it takes 2% of time, while benchmarking the repo it takes around 20% - 50%). As you can see improving the `Lines` iter performance would help a lot.

flamegraph_imara_small_update

Coming up with a single representative testcase is kind of difficult, that's why I took the more quantitative approach with benchmarks in the `imara-diff` repo.

helix-core/src/diff.rs
```diff
@@ -1,58 +1,188 @@
...
 
 #[cfg(test)]
 mod tests {
     use super::*;
 
+    fn test_identity(a: &str, b: &str) {
+        let mut old = Rope::from(a);
+        let new = Rope::from(b);
+        compare_ropes(&old, &new).apply(&mut old);
+        assert_eq!(old, new)
```
pickfire, line 183:
```suggestion
        assert_eq!(old, new);
```

pickfire, line 183:
```suggestion
        assert_eq!(old, new);
```

helix-core/src/diff.rs
```diff
@@ -61,4 +191,25 @@ mod tests {
...
+    }
+
+    #[test]
+    fn new_file() {
+        test_identity("", "foo")
+    }
+
+    #[test]
+    fn deleted_file() {
+        test_identity("foo", "")
```
pickfire, line 213: Inconsistent semicolon

pickfire (approved):
Looks good to me.

pascalkuthe:
This PR is affected by two bugs in ropey:
* a rare panic in the `PartialEq` implementation when using non ascii characters: cessen/ropey#65
* an incorrect `Hash` implementation that causes the `Hasher` (ahash) used by the `imara-diff` interner to frequently detect identical lines as changed lines: cessen/ropey#66

These bugs don't really break this PR, because the panic is pretty rate and the incorrect hashes just causes `imara-diff` to detect more changes than there actually are in a file (the good news is that this PR will get even faster with that fix). While unoptimal deleting and inserting the same line again is a valid transaction and does not cause problems.

We might still want to hold off merging this until the upstream issues are fixed (I am torn on this)

pascalkuthe:
> Great work! I think it's ok to wait for the fixes to land in `ropey`, it shouldn't take too long. Maybe we can attempt to resolve [cessen/ropey#25](https://github.com/cessen/ropey/issues/25) and get it all into `1.6.0`. There's already a couple of other iterators that do similar book-keeping so we don't have to fully reinvent the wheel.

Sounds like a plan :+1: I think it would also be great to land cessen/ropey#67. The bug fixed there doesn't effect this PR (or at least not more then the rest of helix). That bug could potentially cause memory safety issues (not 100% sure on that but it does create a rope which violates its internal guarantees so all bets are off). I will try my hand at implementing a line iterator that fixes cessen/ropey#25, Oonce I have a prototype I will post the PR here.

pascalkuthe:
> I'm working on [cessen/ropey#25](https://github.com/cessen/ropey/issues/25) @pascalkuthe, I should have a solution within the month, preliminary results are good :)

That's a bit unfortunate timing we did some duplicate work here. I already had an optimized version of forward iteration ready and only had a couple bugs left to iron out in reverse iteration. Life just got in the way this weekend/this week so I didn't get around to finishing those last few edgecases and posting a PR yet. The implementation was a bit subtle, I had to use a few tricks to avoid doing duplicate work when creating the subslices. Also there was a bug in the original implementation of reverse iteration or Light rope slices that did not have a trailing newline. Perhaps it makes sense to combine efforts in some form. Could you upload your implementation somewhere and I could compare it to mine?

pascalkuthe:
My line iterator implementation (cessen/ropey#70) now works decently well (it was really tricky to get all edge cases right, luckily `ropey` has a great testsuite). That PR alone brought the **total time (exlcuding io)** to reload a 1GB file down from 14 seconds to 7 so it was really worth it to invest the time here.

Until the new ropey version releases I have pointed this PR at my fork of `ropey` that contains all relevant improvements for people that already want to test it.

archseer:
With these improvements, does it still make sense to do the git gutter diffing in an async task or could we do it sync?

pascalkuthe:
> With these improvements, does it still make sense to do the git gutter diffing in an async task or could we do it sync?

Good point, my gut feeling is the anwser is yes because even if a diff is really fast it still is an expensive full file computation and would still introduce quite a bit of latency into the core edit loop.I would likd do some testing to determine the worst case impact on latency of doing the diff sync.

helix-core/src/diff.rs
```diff
@@ -1,58 +1,195 @@
...
+        self.res.retain(len);
+        self.res
+    }
+}
+
+struct RopeLines<'a>(RopeSlice<'a>);
+
+impl<'a> imara_diff::intern::TokenSource for RopeLines<'a> {
+    type Token = RopeSlice<'a>;
+    // TODO: improve performance of lines iterator (https://github.com/cessen/ropey/issues/25)
```
kirawi, line 145: This should be removed once the upstream PR is merged.

pascalkuthe: Good point, thanks for catching that :)

helix-core/Cargo.toml
```diff
@@ -38,7 +38,7 @@ serde = { version = "1.0", features = ["derive"] }
 serde_json = "1.0"
 toml = "0.5"
 
-similar = "2.2"
+imara-diff = "0.1.0"
```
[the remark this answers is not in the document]
pascalkuthe: Those were just bugfixes that did not affect this PR. The newest version should also be in the lockfile so I doesn't make a difference anyway

PR #4467, opened 2022-10-25
chore: bump Rust to 1.64.0

kosciej:


the-mikedavis:
Firefox is still on 1.63: https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html#schedule

https://github.com/helix-editor/helix/blob/master/docs/CONTRIBUTING.md#minimum-stable-rust-version-msrv-policy

archseer:
The use of 1.61 is deliberate because it makes helix easier to package. https://github.com/helix-editor/helix/blob/d801a6693c3d475b3942f705d3ef48d7966bdf65/docs/CONTRIBUTING.md#minimum-stable-rust-version-msrv-policy

kosciej:
Ok, I understand. My reasoning was that I haven't moved MSRV check in Github Actions, just the default toolchain version. So it's still guaranteed to be compiled with version within MSRV, but with additional benefit of new optimizations and new checks in place (cargo clippy)

pascalkuthe:
I am not a fan of this. I actually really like that `rust-toolchain.toml` is set to 1.61.
It's really easy to forget about MSRV (or for a new contributor to now know about this) so always checking with the right rustc version saves a lot of headache.
This is also true in regards to clippy lints. There are warnings in new clippy versions that warn against certain patterns because there is an std function then does the same thing. However if that function is still unstable on MSRV the warning is just annoying/noise.

The performance differences between new rust versions are also not that huge, especially because our MSRV is not *that* old.

kosciej:
While I agree on performance and clippy arguments, I think the project already have GitHub Actions check for MSRV version, so anything incompatible with MSRV will be catched right away.

pascalkuthe:
> While I agree on performance and clippy arguments, I think the project already have GitHub Actions check for MSRV version, so anything incompatible with MSRV will be catched right away.

Ofcourse CI will detect the problem but this involves pushing a PR, getting approval for the CI run (if you are not preapproved) and waiting for CI to finish. It makes contributing more annoying without any real benefit

kosciej:
Ok, but this basically means that GitHub Actions step for MSRV is basically useless as it's always duplicates job done in (stable) step. If this is way to go, then we should remove one of these GitHub steps

dead10ck:
Thanks for your PR, but we're not moving forward with this right now

PR #4478, opened 2022-10-26
Include unnamed children for html injected into inline markdown

MDeiml:
Simple fix for html tags in inline markdown not working sometimes.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4480, opened 2022-10-26
feat(lsp): LSP preselected items appear first in completion menu

lazytanuki:
# Support LSP preselect completion items

In addition to suggesting completion, some **LSP servers can tell the client which suggestions would be the most probable fit**.

For instance, if you want to assign a value to a variable named `position`, rust-analyzer will preselect a method that is also named `position()` if it is part of the suggested items. The LSP server can mark multiple completion items as preselect, but in this case the order they are send back to the LSP client defines their order. This PR makes these preselected items appear at the top of the completion menu, in the same order as the LSP server intended.

This superseeds #3215 after the merge of #4134. It also partially superseeds #2705.

image

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -92,19 +92,13 @@ impl Completion {
 
     pub fn new(
         editor: &Editor,
-        items: Vec<CompletionItem>,
+        mut items: Vec<CompletionItem>,
         offset_encoding: helix_lsp::OffsetEncoding,
         start_offset: usize,
         trigger_offset: usize,
     ) -> Self {
         // Sort completion items according to their preselect status (given by the LSP server)
-        let mut items = items;
-        items.sort_by(|item1, item2| {
-            item2
-                .preselect
-                .unwrap_or(false)
-                .cmp(&item1.preselect.unwrap_or(false))
-        });
+        items.sort_by_key(|item| !item.preselect.unwrap_or(false));
 
         // Then create the menu
         let menu = Menu::new(items, (), move |editor: &mut Editor, item, event| {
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/completion.rs
```diff
@@ -97,6 +97,16 @@ impl Completion {
...
         trigger_offset: usize,
     ) -> Self {
+        // Sort completion items according to their preselect status (given by the LSP server)
+        let mut items = items;
+        items.sort_by(|item1, item2| {
+            item2
+                .preselect
+                .unwrap_or(false)
+                .cmp(&item1.preselect.unwrap_or(false))
+        });
```
archseer, line 107: Should be possible to use `sort_by_key` (or better yet `sort_unstable_by_key`)

```rust
items.sort_unstable_by_key(|item| item.preselect.unwrap_or(false) )
```

helix-term/src/ui/completion.rs
```diff
@@ -97,6 +97,16 @@ impl Completion {
         start_offset: usize,
         trigger_offset: usize,
     ) -> Self {
+        // Sort completion items according to their preselect status (given by the LSP server)
+        let mut items = items;
```
archseer, line 101: Can't you just add `mut` to `items` in the `fn` declaration?

helix-term/src/ui/completion.rs
lazytanuki: Done, although we need to reverse the conditions that that the ones that have `preselect == Some(true)` are first in the completion menu.
Using `sort_unstable_by_key` would lose the order given by the LSP server, which, with what I've tested, does have its significance.

archseer (approved):

archseer:
Thanks :) Sorry it took so long to get this merged!

lazytanuki:
No problem, thanks for the merge ! 😋

PR #4487, opened 2022-10-27
Fleetish: Adjustments to resemble official theme and reworked diagnostics to reduce subconjunctival hemorrhage

krfl:
- Small changes to more closely resemble the official fleet theme. 
- Some create freedom to fit my own preferences and to keep it interesting.
- Reworked diagnostics to reduce bleeding from the eyes :)

Themelint compliant.

Will update wiki when approved. 

fleetish

### DIFF
--- a/runtime/themes/fleetish.toml
+++ b/runtime/themes/fleetish.toml
@@ -1,11 +1,11 @@
 # Author: Kristoffer Flottorp <kr.fl@outlook.com>
-# Based on a few screenshots of Jetbrains Fleet
+# A take on the JetBrains Fleet theme sprinkled with some creative freedom
 
-"type" = { fg = "orange" } # .builtin
-"constructor" = { fg = "orange" }
-"constant" = { fg = "green" }
+"type" = { fg = "yellow" } # .builtin
+"constructor" = { fg = "yellow" }
+"constant" = { fg = "cyan" }
 # "constant.builtin" = {} # .boolean
-"constant.builtin.boolean" = { fg = "green" } # .boolean
+"constant.builtin.boolean" = { fg = "cyan" } # .boolean
 # "constant.character" = {} #.escape
 "constant.numeric" = { fg = "yellow" } # .integer / .float
 "string" = { fg = "pink" } # .regexp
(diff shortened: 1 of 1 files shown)


runtime/themes/fleetish.toml
```diff
@@ -74,11 +74,11 @@
...
-"diagnostic" = { fg = "orange", bg = "darkest" } # .hint / .info / .warning / .error
-"diagnostic.hint" = { fg = "lightest", bg = "blue_accent" }
-"diagnostic.info" = { fg = "lightest", bg = "purple_accent" }
-"diagnostic.warning" = { fg = "lightest", bg = "yellow_accent" }
-"diagnostic.error" = { fg = "lightest", bg = "orange_accent" }
+"diagnostic" = { modifiers = ["underlined"], underline = { style = "curl" } } # .hint / .info / .warning / .error
+# "diagnostic.hint" = { modifiers = ["underlined"] }
+# "diagnostic.info" = { modifiers = ["underlined"] }
+# "diagnostic.warning" = { modifiers = ["underlined"] }
+# "diagnostic.error" = { modifiers = ["underlined"] }
```
pascalkuthe, line 81:
```suggestion
"diagnostic". underline = { style = "curl"  } 
```
The `underlined` modifier is deprecated and only available for backwards comparability.
Its equivalent to `underline = { style = "line" }` so it has no effect here.

On a different note: Jetbrains fleet seems to show diagnostics as non-curled colored lines.
You could achieve the same thing in helix:
``` toml
"diagnostic.error".underline = { style = "line", color = "orange_accent"  } 
"diagnostic.warning".underline = { style = "line", color = "yellow_accent"  } 
```

This would be a lot more subtle than the old diagnostic highlight because only the underline color is changed instead of the full text and match jetbrains fleet more closely

krfl: Thank you for the review. 

1. I didn't know that underlined was depricated, so I commited your suggestion. Thank you for that :) 
2. My goal with this theme is not to perfectly mimic the fleet theme, but to use the nice colors and sprinkle some creative freedom on top for comfort and my own preferences. I think I've landed on a theme that is easier to read than the official 'fleet dark' theme which has, among other things, a lighter background and a different shade of blue. It also colors some elements differently, but I really like the varied colors the fleetish theme currently has. 
3. Prior to this update, if a file had many errors, your eyes would bleed. Bady. I gave coloring both the fg and the underlines with the accent colors a shot, but I felt that the curly lines without color overrides was very easy on the eyes. Making a theme is hard, because it's all very subjective. My eyes are also very picky with colors and contrasts for some reason.

pascalkuthe: I personally really like colored underlines/curls because it lets me know if I am looking at an error, warning or a hint (its less about aesthetics and more about the provided information). But at the end of the day its your theme and I could always overwrite the diagnostics if I decided to use it.

the-mikedavis (approved):
Looks good! 🎨

the-mikedavis:
Could you update the screenshot in the themes wiki with that new one?

Issue #4488, opened 2022-10-27
Jumplist has Duplicate Entries


ngraham20:
When using the Jumplist panel to navigate the jumplist, duplicate entries quickly emerge and are difficult to get rid of as there is no "delete selection from jumplist" command

To reproduce:
1. Open a file and put the cursor someplace on the file.
2. ctrl+s to add the character to the jumplist
3. add another character
4. and another
5. now start using the jumplist rapidly. You will quickly notice multiple duplicate entries on the jumplist

ngraham20:
OH This is not a real issue, the difference is which character(s) I have highlighted, I had not understood that it was jumping to a particular selection, not a particular line.

ngraham20:
Still though, I do have duplicate entries that build up while jumping around using the jumplist panel. I do not understand what I'm doing to cause this though

the-mikedavis:
This is the expected behavior: the jumplist is analagous to history in a browser and this case is like following a link from page A to page B and back to page A, and so on. The jumplist prevents sequential duplicate entries since they aren't useful but duplicates in overall history can be useful.

ngraham20:
Hmm, I think I'm starting to understand its usecase then. I wonder then if it would be useful to have a list of unique jumplist items to scroll through separately like the buffer list? Frequently, my usecase for the jumplist picker is to scroll through the list of unique jumps I've made recently as a way to visually jump around my documents. This breaks quickly when the jumplist doesn't dedup completely

the-mikedavis:
If the jumplist picker were deduplicated by default then it might be confusing because it would hide parts of the history. Maybe a key in the picker to deduplicate items could work but I don't know how useful that would be for pickers in general since other pickers are deduplicated by default

PR #4489, opened 2022-10-27
Improved jumplist push dedup

ngraham20:
To address #4488 

The jumplist `push()` function only prevents duplicate entries if they're at the end of the vector. While this dedup check is less efficient, even with a jumplist of 100 entries shouldn't be noticeably slower.

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -28,7 +28,7 @@ impl JumpList {
     pub fn push(&mut self, jump: Jump) {
         self.jumps.truncate(self.current);
         // don't push duplicates
-        if self.jumps.last() != Some(&jump) {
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I think its a bit problematic that editor jumps will contentiously get slower the longer a file is open (100 jumps is nothing if I am working on something complexe where I jump to a definition every couple of seconds). This could be a great place for an `index_map::IndexSet`, because it avoids the `O(N)` check on each insertion. If we really don't want to add the extra dependency here, we could implement a very basic indexset with a `RawTable` from `hashbrown instead (we depend on `hashbrown` anyways) but it would be a bit clunky.

ngraham20:
Yeah, turns out the jump list is intended to be the way it is.

PR #4490, opened 2022-10-27
Simplify change/delete selection, remove blckhole register

kosciej:
From my understanding blackhole `'_'` register is used only in `delete_selection_noyank` and `change_selection_noyank`.

Blackhole `'_'` register also significantly changes logic in otherwise very simple `Record` struct.

This PR tries to remove blackhole register altogether, also unifying `yank` logic where applicable.

### DIFF
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -23,10 +23,8 @@ If there is a selected register before invoking a change or delete command, the 
 | `/`                | Last search           |
 | `:`                | Last executed command |
 | `"`                | Last yanked text      |
-| `_`                | Black hole            |
 
 > There is no special register for copying to system clipboard, instead special commands and keybindings are provided. See the [keymap](keymap.md#space-mode) for the specifics.
-> The black hole register works as a no-op register, meaning no data will be written to / read from it.
 
 ## Surround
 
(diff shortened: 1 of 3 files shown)


the-mikedavis:
The blackhole register is a feature though, and it comes from kakoune: https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers

kosciej:
Ok, I get it. What might be possible usages of this register?

ath3:
It can also be used with search if you dont want the searched term to be remembered.

kosciej:
That sounds compelling. 
@the-mikedavis I can remove change removing blackhole register. Do you think refactor for `yank` from this commit would be useful, or should I close this PR?

the-mikedavis:
Actually I kinda like the yanking behavior of `delete_selection_impl`. We could probably remove that `if cx.register != Some('_')` check in `delete_selection_impl` since it's only preventing us from writing to the blackhole register which is a noop anyways

ath3:
> Actually I kinda like the yanking behavior of `delete_selection_impl`. We could probably remove that `if cx.register != Some('_')` check in `delete_selection_impl` since it's only preventing us from writing to the blackhole register which is a noop anyways

Its to prevent unnecessary cloning and collecting of the selected text in case we wont use it.

the-mikedavis:
Closing as the blackhole register is an intentional feature

PR #4510, opened 2022-10-28
feat: Categorize Rust's keywords using more specific scopes

poliorcetics:


poliorcetics:
Screenshot 2022-10-28 at 21 30 01

Allows differenciating `return` for example

runtime/queries/rust/highlights.scm
```diff
@@ -143,24 +151,28 @@
...
 
+[
+  "struct"
+  "enum"
+  "union"
+
+  "type"
+] @keyword.storage.type
+
+"let" @keyword.local
```
sudormrfbin, line 174: `keyword.local` doesn't seem to be a valid scope

poliorcetics: It exists as a consequence of `keyword.local.import` in `ledger`, but yes that was not correct, moved to `keyword.storage`

sudormrfbin (approved):

the-mikedavis:
Thanks!

PR #4512, opened 2022-10-29
Add "Copy selection to next/prev line" to keymap docs

s1ck:


sudormrfbin:
This is already present in the `Selection manipulation` setion of the docs: https://docs.helix-editor.com/keymap.html#selection-manipulation.

s1ck:
> This is already present in the `Selection manipulation` setion of the docs: https://docs.helix-editor.com/keymap.html#selection-manipulation.

Ah ok... I was expecting it to be present in `Normal Mode` too, but maybe that's against the idea. Will close this.

Issue #4331, opened 2022-10-17
XML Syntax highlighting not working out of the box


bravo2zero:
As title says, fresh brew installation of Helix, XML syntax highlighting is missing and XML is not showing up in "set-language" hints.

❯ hx --version
helix 22.08.1 (66276ce6)

MacOS 12.6

the-mikedavis:
XML is not yet supported. See https://github.com/helix-editor/helix/issues/3479#issuecomment-1221323100

Issue #3479, opened 2022-08-19
Xml syntax highlight


sbergot:
I was surprised to see that there was no support for xml yet. I am willing to give a shot to adding syntax highlighting. Is it something doable for a rust newbie?

ChrHorn:
You could try to just enable the html highlighting for xml files.

Alternatively, there is also https://github.com/unhammer/tree-sitter-xml, but I don't know how good that implementation is.

sbergot:
@ChrHorn yes I was about to try that. Adding xml to the html file types works. It is good enough for me but it might be an issue for those who have the html language server enabled.

Anyway I am closing this. Thanks for the help!

kirawi:
You can alternatively create a new language entry for XML that uses the syntax highlighting for HTML:

```toml
[[language]]
name = "xml"
scope = "scope.xml"
injection-regex = "xml"
file-types = ["xml"]
roots = []
indent = { tab-width = 2, unit = "  " }
grammar = "html"
```

zyfyy:
> You can alternatively create a new language entry for XML that uses the syntax highlighting for HTML:
> 
> ```toml
> [[language]]
> name = "xml"
> scope = "scope.xml"
> injection-regex = "xml"
> file-types = ["xml"]
> roots = []
> indent = { tab-width = 2, unit = "  " }
> grammar = "html"
> ```
> 
> You'll have to create a folder at `runtime/queries/xml` and create `highlights.html` and `injections.scm` files that contain:
> 
> ```scheme
> ;  inherits html
> ```

just `:set language html`

PR #4518, opened 2022-10-29
feat(lang): add xml

mtoohey31:
There have been two issues opened about this (#3479 and #4331) so I think it would be good for us to include something in the default configuration that handles XML.

The solution in this pull request adds a new grammar. This same grammar was added to lapace in https://github.com/lapce/lapce/pull/1554, and it seems to handle things well. The highlight queries were copied from the grammar's repository. Here's a screenshot with the Darcula theme:

2022-10-29T14:19:26,733205436-04:00

There was also [a suggestion](https://github.com/helix-editor/helix/issues/3479#issuecomment-1221323100) on one of the XML-related issues that we could re-use the HTML grammar. If we'd prefer to avoid adding another grammar, let me know and I can update this pull request. The XML grammar is pretty small though (only 97K), so I think the small size increase might be worth it.

Also, I haven't included a language server in the configuration, since I don't use one myself. [nvim-lspconfig](https://github.com/neovim/nvim-lspconfig/blob/master/doc/server_configurations.md) has instructions for [lemminx](https://github.com/eclipse/lemminx), but it looks like it will take a bit of work to get running on NixOS, so I haven't been able to test it yet. Would it be fine if we just left it without a language server right now? I'm thinking that it should probably be added by someone who actually uses an XML language server and has tested it with Helix for a little bit.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -127,5 +127,6 @@
 | wat | ✓ |  |  |  |
 | wgsl | ✓ |  |  | `wgsl_analyzer` |
 | xit | ✓ |  |  |  |
+| xml | ✓ |  | ✓ |  |
 | yaml | ✓ |  | ✓ | `yaml-language-server` |
 | zig | ✓ | ✓ | ✓ | `zls` |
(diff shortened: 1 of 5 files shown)


the-mikedavis (approved):

the-mikedavis:
> The XML grammar is pretty small though (only 97K), so I think the small size increase might be worth it.

I agree. The HTML parser seems to break no some XML syntax like some of the pictured doctype statements. Ideally we could cover both in one but since it's so small I think it's ok to have two in this case.

> Would it be fine if we just left it without a language server right now?

Yep I think it's ok. If someone would like to test it out and add the configuration, that can be a follow-up change.

the-mikedavis:
Thanks!

PR #4519, opened 2022-10-29
fix: make `scroll` aware of tabs and wide characters

mtoohey31:
This pull request fixes the same issue as #2620, except for `scroll` instead of `move_vertically`.

The issue is kind of complicated to explain in text, so here's a video of the before and after:

out

I believe the comment on line 1395 about things getting truncated properly should still be accurate, but I wouldn't mind a second opinion on that.

Also, do we have tests for scrolling anywhere? Is there somewhere that I could add a unit test for this behaviour?


the-mikedavis (approved):
Ah nice catch, this aligns the behaviors of navigating with `j`/`k` and scrolling with `zj`/`zk`

Issue #1593, opened 2022-01-27
Allow using `ctrl + d` to close a buffer from buffer picker by default


Zaplae:
`ctrl + d` is equals to `:buffer_close`, this is quick by input `buffer_close`
if you have unsaved buffer, you have to input `:bufer_close!` :)

Zaplae:
Ok, I just did't know it's alias and now the display of aliases was fixed.
I'll close this issue.

mangas:
@sudormrfbin might be worth renaming the ticket as it looks like the purpose has changed?

sudormrfbin:
The title for this issue is referring to closing buffers from the buffer picker using a keybind and I think you opened a ticket with the same intent? If so the title matches since it's still about adding a keybind to the buffer list.

mangas:
Yep you're right, I got a little confused with the alias related discussion about, nevermind!

pickfire:
If anything I think it is better to add shortcuts to `space + b` and have a shortcut key there to delete buffer, rather than having another popup created just to delete buffer.

kirawi:
I added a `handle_key_events` closure field to the `Picker` to have custom keys for each type of picker.

pickfire:
I think later on we need some kind of infobox mechanisms to make those keys discoverable.

PR #4522, opened 2022-10-29
Feat/close buffer from picker

EricHenry:
Closes: #1593 

Hey all, this allows the user to delete a buffer from the file picker by using `ctrl-x` to close and `ctrl-X` to force close. I didn't update any documentation yet; I first wanted to get feedback on my approach and the keybindings that I used. 

Please let me know any ideas on how to improve this implementation.


helix-term/src/ui/picker.rs
```diff
@@ -330,13 +338,15 @@ pub struct Picker<T: Item> {
     show_preview: bool,
 
     callback_fn: Box<dyn Fn(&mut Context, &T, Action)>,
+    key_event_callback_fn: KeyEventCallback<T>,
```
andreytkachenko, line 341: maybe make it optional? like `Option<KeyEventCallback<T>>`?

lukepighetti:
What about `C-d` instead of `C-x`? That will match the close buffer key in normal mode

EricHenry:
@lukepighetti I initially wanted to use `c-d`, but it is already being used by the picker to scroll the page down.

helix-term/src/ui/picker.rs
EricHenry: Moved things around, the key_event_callback is not longer part of the constructor, instead there is a separate function to set the key event callback function.

EricHenry:
I think I have landed on a good implementation, let me know your thoughts. It adds infrastructure for any specific usage of the picker to define a special handling of key events that do not conflict with the default key events for the picker.

the-mikedavis:
My hangup with this change is that I think it's better solved by #5505. I don't think the code added here will be reused in the solution for that and I'm fairly certain that #5055 will make it very easy to implement this.

EricHenry:
@the-mikedavis hey! Thanks for the feedback. I totally see where you are coming from. #5505 would be a great enhancement and allow for this exact flexibility. 

If no one is working on that issue. I’d love to pick it up since I’ve been in that code with this PR. Do you have any guidance or thoughts on the implementation. You mentioned #5055, what are your thoughts on using the dynamic picker to achieve #5505. Would you like to see this PR left open and then updated once the other change is made?

P.S. let me know if you’d like to take the conversation to that issue.

the-mikedavis:
Ah, 5055 was a typo. I meant 5505 😅

I've been thinking about #5505 and tinkering with ideas for a while and it's very tricky to solve. I wrote out some ideas in [this branch](https://github.com/helix-editor/helix/compare/md-compositor-key-remapping-idea) (see [`478fc9a`](https://github.com/helix-editor/helix/commit/478fc9aa2c27f0d41819ef4feb3b42984dd99fe1) in particular). Feel free to have a go at it but I suspect that this is something that may end up being a very large change, and we will probably want to discuss it or solve it internally since really large changes are hard to collaborate on.

gabydd:
I tried to implement the ``AnyPicker`` here https://github.com/gabydd/helix/tree/md-compositor-key-remapping-idea very interesting problem, definitely a difficult one to design well, this way(see the commit on my branch that uses an ``Enum`` and ``Any``) seems better then changing the way ``Items`` work though.

pascalkuthe:
This branch has gone quite stale. We also intend to solve this differently (see above) so I am closing this out. Thanks for contributing

Issue #4523, opened 2022-10-30
[feat] support `wit` grammar


hh9527:
[WIT](https://github.com/WebAssembly/component-model/blob/main/design/mvp/WIT.md)

An IDL for WebAssembly Component Model

grammar: https://github.com/hh9527/tree-sitter-wit

I have tested the grammar/query with helix.

PR #4525, opened 2022-10-30
Support WIT grammar

hh9527:
fix https://github.com/helix-editor/helix/issues/4523

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1917,3 +1917,25 @@ roots = []
 [[grammar]]
 name = "xml"
 source = { git = "https://github.com/RenjiSann/tree-sitter-xml", rev = "422528a43630db6dcc1e222d1c5ee3babd559473" }
+
+[[language]]
+name = "wit"
+scope = "source.wit"
+injection-regex = "wit"
+file-types = ["wit"]
+roots = []
+comment-token = "//"
+indent = { tab-width = 2, unit = "  " }
+
+[language.auto-pairs]
+'(' = ')'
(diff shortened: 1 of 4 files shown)


languages.toml
```diff
@@ -1917,3 +1917,25 @@ roots = []
...
+comment-token = "//"
+indent = { tab-width = 2, unit = "  " }
+
+[language.auto-pairs]
+'(' = ')'
+'{' = '}'
+'[' = ']'
+'"' = '"'
+"'" = "'"
+"<" = ">"
```
the-mikedavis, line 1936: These are used above in XML so that `<` gets closed by `>` but shouldn't be necessary for most other languages. (If not specified, the default pairs are used.) Is `>` or `<` never used as an operator in wit?

the-mikedavis (commented):
Could you add a license to tree-sitter-wit?

languages.toml
hh9527: `<` & `>` in WIT is for template parameters only, e.g: `option<T>`, `result<T, E>`
and in WIT, there is no expression, only type & interface declaration, so I think `<>` could be paired.

hh9527: > Could you add a license to tree-sitter-wit?

new commit update the rev of `tree-sitter-wit` which include a LICENSE file & license field of `package.json` (MIT license)

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4528, opened 2022-10-30
Update LaTex grammar

Triton171:
Update the LaTex grammar. I also noticed that the comment injection has been broken for a while and math delimiters were not highlighted, so I fixed that too.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4529, opened 2022-10-30
Update SQL grammar

seshotake:
Before: 

before-tree-sitter

After:
after-tree-sitter


the-mikedavis (approved):
This is a big improvement, thanks!

PR #4535, opened 2022-10-30
feat: add 'file-absolute-path' to statusline

dcardamo:
Adds the config option 'file-abs-path' to status line to show the absolute path for a file.


helix-view/src/document.rs
```diff
@@ -1087,6 +1087,17 @@ impl Document {
...
 
+    pub fn canonicalized_path(&self) -> Option<PathBuf> {
+        if let Some(path) = self.path.as_deref() {
+            helix_core::path::get_canonicalized_path(path).ok()
+        } else {
+            None
+        }
+        // self.path
+        //     .as_deref()
+        //     .map(helix_core::path::get_canonicalized_path)
```
the-mikedavis, line 1098: This can work:

```
self.path
    .as_ref()
    .and_then(|p| helix_core::path::get_canonicalized_path(p).ok())
```

dcardamo: Thanks @the-mikedavis, that's cleaner.

helix-view/src/editor.rs
```diff
@@ -313,6 +313,9 @@ pub enum StatusLineElement {
     /// The file nane/path, including a dirty flag if it's unsaved
     FileName,
 
+    /// The file absolute path, including a dirty flag if it's unsaved
+    FileAbsPath,
```
the-mikedavis, line 317: This is quite a small nitpick but could we call this `FileAbsolutePath` (and `file_absolute_path` for the function)? With the other elements we haven't done any abbreviation yet so I think we should keep it consistent

dcardamo: Makes sense to be consistent.  Updated.

the-mikedavis (dismissed):

the-mikedavis (approved):

pascalkuthe (approved):

PR #4537, opened 2022-10-31
build(nix): update nci, fixup flake

yusdacra:
- update nci input
- fixup flake to match new nci interface
- filter helix source to reduce unnecessary recompiles

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     nci = {
-      url = "path:/home/patriot/proj/nix-cargo-integration";
+      url = "github:yusdacra/nix-cargo-integration";
       inputs.nixpkgs.follows = "nixpkgs";
       inputs.rust-overlay.follows = "rust-overlay";
     };
--- a/flake.nix
+++ b/flake.nix
@@ -8,7 +8,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     nci = {
-      url = "github:yusdacra/nix-cargo-integration";
+      url = "path:/home/patriot/proj/nix-cargo-integration";
       inputs.nixpkgs.follows = "nixpkgs";
       inputs.rust-overlay.follows = "rust-overlay";
     };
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -8,7 +8,7 @@
       inputs.nixpkgs.follows = "nixpkgs";
     };
     nci = {
-      url = "github:yusdacra/nix-cargo-integration";
+      url = "path:/home/patriot/proj/nix-cargo-integration";
```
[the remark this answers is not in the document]
yusdacra: Oops, thanks for noticing that lol. fixed

flake.nix
```diff
@@ -21,73 +21,111 @@
...
+        "CHANGELOG.md"
+        "shell.nix"
+        "default.nix"
+        "grammars.nix"
+        "flake.nix"
+        "flake.lock"
+      ];
+      ignorePaths = path: type: let
+        split = lib.splitString "/" path;
+        actual = lib.drop 4 split;
```
the-mikedavis, line 50: This seems specific - is this assuming the depth of the helix repo on the building machine?

flake.nix
```diff
@@ -21,73 +21,111 @@
...
+                mkdir -p $out/bin
+                ln -sf ${old}/bin/* $out/bin/
+                wrapProgram "$out/bin/hx" ''${makeWrapperArgs[@]} --set HELIX_RUNTIME "${runtimeDir}"
+              '';
+          in
+            helix-wrapped
+            // {override = makeOverridableHelix old;};
+        in
+          makeOverridableHelix old {};
+        helix-term.shell = prev: {
```
the-mikedavis, line 128: Based on the docs I think this should move up to `config`, right? Without that I don't have rust-analyzer in my `nix develop` shell

yusdacra: Yes, that would be correct, I forgot to test the devshell...

flake.nix
yusdacra: That is for removing `/nix/hash-source/` part after splitting the nix store path:
```
nix-repl> lib.splitString "/" "/nix/store/m2wybrc3grfn76hxcv4gyiyncng17z30-hello-2.12.1"
[ "" "nix" "store" "m2wybrc3grfn76hxcv4gyiyncng17z30-hello-2.12.1" ]
```
I can add a comment there. I did not want to use `nix-filter` here since IMO it's to complicated to just pull another dep in, we don't need all of the stuff in there for filtering the helix source.

flake.nix
yusdacra: Updated the devshell, should be working now

the-mikedavis (approved):

the-mikedavis:
Sweet, I think I just saw the path filtering make a really fast build while testing out this branch 😀

the-mikedavis:
Thanks @yusdacra!

PR #4538, opened 2022-10-31
Update SSH client config grammar & highlight queries

sebhoss:
This updates the highlight queries for SSH client configs to be more closely aligned with what helix themes expect and it fetches the latest upstream source which fixes some grammar related issues as well.

Before:
before

After:
after

### DIFF
--- a/runtime/queries/sshclientconfig/highlights.scm
+++ b/runtime/queries/sshclientconfig/highlights.scm
@@ -1,17 +1,17 @@
-(host) @keyword
-(host_value) @identifier
+(host) @namespace
+(host_value) @string
 
-(match) @keyword
-(match_value) @identifier
+(match) @namespace
+(match_value) @string
 
 (add_keys_to_agent) @keyword
-(add_keys_to_agent_value) @boolean
+(add_keys_to_agent_value) @constant.builtin.boolean
 
 (address_family) @keyword
-(address_family_value) @type
+(address_family_value) @constant.builtin.string
 
 (batch_mode) @keyword
-(batch_mode_value) @boolean
+(batch_mode_value) @constant.builtin.boolean
 
 (bind_address) @keyword
 (bind_address_value) @string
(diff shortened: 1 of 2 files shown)


runtime/queries/sshclientconfig/highlights.scm
```diff
@@ -1,17 +1,17 @@
...
+(match) @namespace
+(match_value) @string
 
 (add_keys_to_agent) @keyword
-(add_keys_to_agent_value) @boolean
+(add_keys_to_agent_value) @constant.builtin.boolean
 
 (address_family) @keyword
-(address_family_value) @type
+(address_family_value) @constant.builtin.string
```
the-mikedavis, line 11: `constant.builtin.string` is not defined: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

this could be `constant.builtin` or `string`

runtime/queries/sshclientconfig/highlights.scm
```diff
@@ -20,165 +20,165 @@
...
-(forward_agent_value) @boolean
+(forward_agent_value) @string
 
 (forward_x11) @keyword
-(forward_x11_value) @boolean
+(forward_x11_value) @constant.builtin.boolean
 
 (forward_x11_timeout) @keyword
-(forward_x11_timeout_value) @time
+(forward_x11_timeout_value) @string.special.time
```
the-mikedavis, line 103:
```suggestion
(forward_x11_timeout_value) @constant.numeric.integer
```

runtime/queries/sshclientconfig/highlights.scm
sebhoss: ah great! I was looking for that doc but somehow missed it. I've adjusted all `constant.builtin.string` usages to either be `constant.builtin` for those config options that have a predefined set of acceptable values and `string` for everything else.

runtime/queries/sshclientconfig/highlights.scm
sebhoss: Thanks - applied in aea3411

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4542, opened 2022-10-31
Added missing keywords to wat (wasm) hightlights

Jomy10:
added "if", "then", "else", "block", "loop", "end" and "mut" to the wat keyword highlights.


the-mikedavis:
Thanks!

Issue #2245, opened 2022-04-23
Move lines or blocks up or down


gregorriegler:
IMO one of the most important features for a text editor is the ability to quickly move lines, but also blocks up and down. 

In VIM I achieve this using the https://github.com/matze/vim-move plugin. However it would be awesome if helix did something like this out of the box.

archseer:
xdp/xdP is already very short and you can always create a mapping for it if you want.

workingJ:
@gregorriegler 
```toml
# line-up
C-e = ["keep_primary_selection","extend_line","yank","move_line_up","open_above","normal_mode","replace_with_yanked", "move_line_down", "move_line_down", "extend_line", "delete_selection", "move_line_up", "move_line_up"]
# line-down
C-d = ["keep_primary_selection","extend_line","delete_selection", "paste_after", "move_line_down"]
```

may be that helps

gregorriegler:
Thank you for the tips!
That works good for a single line, yes.

I wish I could just select a complete function and move it around, maybe down below the next function.
Or lets say I have a function in a function, and I want to move it out to the top level, and it automatically unindents.
Yes I know I can cut and paste. But once the moving is possible, you don't want back.

icodeforyou-dot-net:
I agree, that is a functionality I was looking for as well and couldn't find: moving whole code blocks as described.

gregorriegler:
It's also viable for non-code. Think paragraphs, config sections, lists. So many cases, I use it regularely.

workingJ:
@icodeforyou-dot-net @gregorriegler ,well you can achive that also with these commands

zakaria-chahboun:
+ Aldo duplicate the line down

zakaria-chahboun:
Also duplicate the line up/down

hustcer:
I need this feature too

zakaria-chahboun:
You can do:
`Ctrl+Alt+Up`  / `Ctrl+Alt+Down` to duplicate lines in *insert mode*
`Alt+Up`  / `Alt+Down` to move lines in *insert mode*

By adding this to the config file:

```toml
[keys.insert]
# move line up
"A-up" = ["extend_line","yank","move_line_up","open_above","replace_with_yanked","move_line_down", "move_line_down", "extend_line", "delete_selection", "move_line_up", "move_line_up"]
# move line down
"A-down" = ["extend_line","delete_selection", "paste_after", "move_line_down"]
# duplicate line up
[...]

PR #4545, opened 2022-10-31
[WIP] Add support for moving lines or selections above and below

sireliah:
Preliminary PR implementing moving lines or selections up and down. See more in https://github.com/helix-editor/helix/issues/2245


workingj:
love it !
:smiley:

helix-term/src/commands.rs
```diff
@@ -4901,6 +4905,168 @@ fn add_newline_impl(cx: &mut Context, open: Open) {
...
+                            acc.push(last_change);
+                            acc.push(change);
+                            Continue(acc)
+                        }
+                    } else {
+                        acc.push(change);
+                        Continue(acc)
+                    }
+                })
+                .into_inner()
```
the-mikedavis, line 5015: This can be written more cleanly without itertools; no need to pull in an extra dependency.

```rust
let mut new_changes: Vec<Change> = Vec::new();

for change in changes {
    match new_changes.last() {
        Some(last_change) if last_change.0 >= change.0 || last_change.1 >= change.1 => {
            return new_changes;
        }
        _ => new_changes.push(change.clone()),
    }
}

new_changes
```

the-mikedavis (changes requested):
The behavior of this on fully selected lines is fine. Moving the full line when only a subset of the line is selected though conflicts with the select-then-act paradigm.

You end up paying for it both in terms of code (conflict resolution functions for overlapping selections; also see all the deleted code in #4123 for a similar situation) and in multi-cursor behavior (for example: `C` and then `C-j` moves the cursors oddly; `C` and then `C-k` removes one of the cursors).

To reiterate https://github.com/helix-editor/helix/issues/2245#issuecomment-1107708361 (especially since https://github.com/helix-editor/helix/pull/4458): this can be roughly implemented in terms of existing commands

```toml
[keys.normal]
C-j = ["extend_to_line_bounds", "delete_selection", "paste_after"]
C-k = ["extend_to_line_bounds", "delete_selection", "move_line_up", "paste_before"]
```

So unless this adds anything extra like handling indentation, I don't see a need to add extra commands for this.

helix-term/src/commands.rs
sireliah: Thanks for the suggestions!

> Moving the full line when only a subset of the line is selected though conflicts with the select-then-act paradigm.

Yeah, I also wondered about this being a bit inconsistent in the sense that modifications are usually applied to the selected area. The way I solved the problem is similar to what other editors do, such as sublime text. 

If we have in mind that helix always has some selection (cursor itself), then moving not fully selected line would be conflicting with select-then-act paradigm. Do you have alternative suggestion that would keep the feature usable? 

> multi-cursor behavior (for example: C and then C-j moves the cursors oddly; C and then C-k removes one of the cursors).

As I mentioned in the issue comment, I've left some edge case unresolved until I get signal that community want feature like this to be supported at all. If we want to have full multi cursor support, I can work on these cases.

> So unless this adds anything extra like handling indentation, I don't see a need to add extra commands for this.

Again, I communicated I would like to add indentation as the next step, but there's nothing preventing me from doing that in this PR (since it's just one function).

sireliah: > This can be written more cleanly without itertools; no need to pull in an extra dependency.
> 
> ```rust
> let mut new_changes: Vec<Change> = Vec::new();
> 
> for change in changes {
>     match new_changes.last() {
>         Some(last_change) if last_change.0 >= change.0 || last_change.1 >= change.1 => {
>             return new_changes;
>         }
>         _ => new_changes.push(change.clone()),
>     }
> }
> 
> new_changes
> ```

Thanks, I'll do it!

7ombie:
> To reiterate https://github.com/helix-editor/helix/issues/2245#issuecomment-1107708361 (especially since https://github.com/helix-editor/helix/pull/4458): this can be roughly implemented in terms of existing commands

This is incorrect. There's currently no way to implement this functionality in Insert Mode, and it's broken in Normal Mode (the cursor position is lost). Sorry, but "roughly implemented" here just means *partial and buggy*.

> So unless this adds anything extra like handling indentation, I don't see a need to add extra commands for this.

I like to use <kbd>Shift</kbd> with the arrow keys to move text up, down, left and right. This allows me to easily (manually) alter the indentation as I move blocks up and down. I prefer this over an editor that indents and dedents code automatically as I move it (I always found that janky and inconsistent).

Helix should implement these commands, if only for the sake of correctness.

pascalkuthe:
> > To reiterate [#2245 (comment)](https://github.com/helix-editor/helix/issues/2245#issuecomment-1107708361) (especially since #4458): this can be roughly implemented in terms of existing commands
> 
> This is incorrect. There's currently no way to implement this functionality in Insert Mode, and it's broken in Normal Mode (the cursor position is lost). Sorry, but "roughly implemented" here just means _partial and buggy_.
> 
> > So unless this adds anything extra like handling indentation, I don't see a need to add extra commands for this.

* dking aything in insert mode besides inserting text and automcompletions is an anti feature and a usecase we do not support. You should use normal mode
* Helix follows the selection -> action model so any moved lines must be first selected so what you call losing the selection is on purpose because text *must be selected to be acted upon*. This is an absolutely fundamental concept of the editor that we will not deviate from.

sireliah:
> Helix follows the selection -> action model so any moved lines must be first selected so what you call losing the selection is on purpose because text must be selected to be acted upon. This is an absolutely fundamental concept of the editor that we will not deviate from.

I'd like the helix maintainers and community to finally agree on the acceptance criteria of this feature. In the current solution the "selection -> action model" is not honored, because not selected lines are moved.

@pascalkuthe @the-mikedavis would you mind setting up some kind of poll or question that would settle this matter? At this point I'm interested in clear decision what behaviour is expected in the editor. Otherwise I'm fine with closing this PR and forgetting about this feature request.

7ombie:
Just to be sure that I read that correctly, I've paraphrased it here:

> Doing anything in Insert Mode, besides inserting text and auto-completions, is an anti-feature and a usecase we do not support. You should use Normal Mode.

Sorry, but calling a usecase an anti-feature, then just stating that it's a thing we do not support, without any elaboration, doesn't really explain anything.

I guess I now know that my Helix configuration is entirely at odds with some unspecified philosophy, and that Helix forces it's users to edit text as it's developer's intended, but don't see *why* I can't bind whatever I like to whatever I like. It's my editor after all.

> Helix follows the selection -> action model so any moved lines must be first selected so what you call losing the selection is on purpose because text must be selected to be acted upon. This is an absolutely fundamental concept of the editor that we will not deviate from.

This makes sense. Point taken.

7ombie:
Please help me understand what I'm missing here...

In Insert Mode, I use simple bindings (Top Layer and Shift Layer) for editing text. In Normal Mode (and Select Mode), I use those bindings to run commands.

The chorded bindings (using <kbd>Alt</kbd> and <kbd>Ctrl</kbd>) are always available (in Insert Mode and Normal Mode), and there's no reason for them behave inconsistently across modes. Only the simple bindings need to be modal.

I personally want Normal Mode and Select Mode to use the same bindings, and for Insert Mode to only differ in terms of what it binds to the Top Layer and Shift Layer. This seems so simple and intuitive, but it's not allowed.

I can save a file in Insert Mode, but I'm not allowed to indent code, or swap adjacent lines of code, or comment-out a selection (even though I have bindings for those operations that don't conflict with anything). I have to switch back and forth between modes for no apparent reason. They're just "anti features" otherwise.

I'd be grateful if anyone could elaborate.

the-mikedavis:
I'm strongly against this behavior since it breaks with select-then-act. I think this would be a good test when the plugin system exists that the API is flexible enough but I don't want to see it in core. I won't close it outright since others may be interested in merging your work into their own forks but I won't be reviewing or merging it.

the-mikedavis:
> I guess I now know that my Helix configuration is entirely at odds with some unspecified philosophy, and that Helix forces it's users to edit text as it's developer's intended, but don't see why I can't bind whatever I like to whatever I like.

See [vision.md](https://github.com/helix-editor/helix/blob/master/docs/vision.md). We're opinionated about modal editing and select-then-act and not interested in supporting other editing modes. Insert mode is just for basic text insertion/deletion and more complicated edits belong in normal or select mode. We also don't like the readline-like commands in insert mode (see [here](https://github.com/helix-editor/helix/blob/9690e0b733fd82e0ced65388a24ad80f769ce3b4/book/src/keymap.md#L380-L393)).

> It's my editor after all.

It can be if you fork it. I'm not trying to be snarky here: I genuinely recommend forking and making whatever changes you want. Then it's truly your editor.

> I can save a file in Insert Mode, but I'm not allowed to indent code, or swap adjacent lines of code, or comment-out a selection

FWIW saving a file in insert mode also isn't supported and we're not interested in changing that: https://github.com/helix-editor/helix/pull/2883.

7ombie:
@the-mikedavis - Thank you for taking the time to reply. I appreciate that.

> FWIW saving a file in insert mode also isn't supported and we're not interested in changing that.

Sorry. I misread the docs. `Ctrl-S` is bound to `commit_undo_checkpoint`. I just thought that meant save and update the version control or something. I had to re-read it.

> It can be if you fork it. I'm not trying to be snarky here: I genuinely recommend forking and making whatever changes you want. Then it's truly your editor.

I can't argue with this. It's your project after all. Thing is, I'm not able to type properly (literally - I can't sit at a desk), so have to configure my editor very differently to other people. For example, I rely on the edges of the keyboard to orient my hands, so the home row is actually the farthest away in some sense. I can't really use an opinionated editor without forking it.

Again, thanks for taking the time. All the best.

omentic:
This patch is very nice, thanks for the hard work. It is buggy when Unicode characters are involved however - the cursor position will skip forward or backward a few positions upon every movement until it skips itself off the line and starts moving something else. It's probably just accidentally treating Unicode characters as their size rather than as a single grapheme, I'll sit down and bughunt for it in the next few days.

workingj:
Hi, any progress in sight here?

omentic:
I have it (and a number of other patches) merged into my fork. I'll rebase on master and see about distribution tomorrow.

omentic:
Found the bug: `change.line_text.as_ref().map_or(0, |x| x.len())` messes up with Unicode because `len()` returns the length of the string in bytes. You need `.chars().count()`. 

Made a couple of trivial changes to this patch and merged into my fork at https://github.com/omentic/helix-ext.

sireliah:
Awesome, thanks for debugging that @omentic!

milesgranger:
@the-mikedavis  Referred to this PR by the issue [after this comment](https://github.com/helix-editor/helix/issues/2245#issuecomment-1819840233). 

To my understanding, the argument of select-then-act being broken b/c of moving a partial line selection up or down is inconsistent as we can currently select a partial line and de/indent the entire line.

To be clear, I'm in no way advocating de/indenting should only apply to the selection of a partially selected line, that would be maddening. :) Only that the logic doesn't hold here when talking about moving partially selected lines up or down, when it's allowed when moving side to side.

workingj:
@sireliah grate work so far, would you mind finishing it? :D
my knowledge and programing experience are quite not there yet to tackle this i think...

omentic:
@workingj It is finished. The Helix devs have expressed that they don't want it in the editor. If you want it you can try my fork.

7ombie:
The *Select Then Act Principle* evidently works very well as a design guideline, but it makes no sense to adhere to it when it doesn't work. Indentation is a perfect example. Some operations are intrinsically line-based, so having operations that implicitly extend selections to the entire line seems totally natural (and already necessary).

I'm not convinced we need the feature (in its current form), and have explained why elsewhere. I just generally disagree with dismissing popular feature requests out-of-hand, just because they conflict with a design guideline.

workingj:
> @workingj It is finished. The Helix devs have expressed that they don't want it in the editor. If you want it you can try my fork.

Thankyou!
Ill do so. 😀

PR #4554, opened 2022-11-01
Exit select mode on replace commands

xJonathanLEI:
Currently Helix doesn't exit select mode when replace commands are used, which is inconsistent with yank commands.

Confirmed in Matrix with @archseer that this is a bug, though I wasn't very clear about _which_ replace command I was talking about (there are 4). But I think it makes sense to exit select mode on them all. Please let me know if that's not the case. Thanks!

This PR makes the editor exit select mode on the follow commands:

- `replace`
- `replace_with_yanked`
- `replace_selections_with_clipboard`
- `replace_selections_with_primary_clipboard`

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1304,6 +1304,7 @@ fn replace(cx: &mut Context) {
             });
 
             apply_transaction(&transaction, doc, view);
+            exit_select_mode(cx);
         }
     })
 }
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
I'm feeling that the CI failure is irrelevant. I just tried running the workflow on the same commit on my fork and it passed:

https://github.com/xJonathanLEI/helix/actions/runs/3368408275

_(note that the commit hash is the same as the PR)_

archseer:
Restarted the CI, seems fine now. Must be another less frequent race condition...

helix-term/src/commands.rs
```diff
@@ -3612,10 +3614,12 @@ fn replace_selections_with_clipboard_impl(
 
 fn replace_selections_with_clipboard(cx: &mut Context) {
     let _ = replace_selections_with_clipboard_impl(cx.editor, ClipboardType::Clipboard, cx.count());
+    exit_select_mode(cx);
```
archseer, line 3617: This could be part of replace_selections_with_clipboard_impl?

xJonathanLEI: Yes I can change this. I was originally doing that but changed to this after checkout out the `yank` command code, which seems to do exactly this.

xJonathanLEI: Moved the `exit_select_mode(cx)` call into `replace_selections_with_clipboard_impl()`. Had to also change the parameters since in needs access to `cx`.

the-mikedavis:
Thanks!

Issue #4544, opened 2022-10-31
Removing a bracket removes more than one bracket


Jomy10:
### Summary

In this short video, I delete a closing bracket, but it also removes the closing bracket to the right of it.

https://user-images.githubusercontent.com/62475953/199091930-8ad682cc-40cd-42ee-b7e5-ed9df4ef140b.mov

### Reproduction Steps

type something like `(())` in helix. Remove the third bracket with backspace, the last one will disappear as well.

### Platform

macOS

### Terminal Emulator

iTerm2

### Helix Version

helix 22.08.1 (f41f28b6)

David-Else:
This problem trips me up often too.

ygabuev:
https://github.com/helix-editor/helix/pull/4558 is the change that fixes the issue.

However, there is still a problem with AutoPairs: they are annotated to be keyed by the opener: https://github.com/helix-editor/helix/blob/8ff92c749287fb3ac4a68661d0ee99cb56cc6295/helix-core/src/auto_pairs.rs#L19-L22
but the `new()` method also adds closers as keys: https://github.com/helix-editor/helix/blob/8ff92c749287fb3ac4a68661d0ee99cb56cc6295/helix-core/src/auto_pairs.rs#L90-L91

This might be the issue why the bug was introduced in the first place - the contributor assumed that there are no closers in the list of keys of an AutoPairs instance.

PR #4558, opened 2022-11-01
Fix `delete_char_backward` for paired characters

ygabuev:
When backward-deleting a character, if this character and the following character form a Pair, we want to delete both. However, there is a bug that deletes both characters also if both characters are closers of some Pair.

This PR fixes that by adding an additional check that the deleted character should be an opener in a Pair.

Closes https://github.com/helix-editor/helix/issues/4544.

Are any more tests needed?

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3198,6 +3198,7 @@ pub mod insert {
                         (Some(_x), Some(_y), Some(ap))
                             if range.is_single_grapheme(text)
                                 && ap.get(_x).is_some()
+                                && ap.get(_x).unwrap().open == _x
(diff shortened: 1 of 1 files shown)


dead10ck:
If you want to add test(s), you can add them in [here](https://github.com/helix-editor/helix/blob/master/helix-term/tests/test/auto_pairs.rs), but this is a trivial change, so I don't consider that a blocker.

ygabuev:
> This looks right to me, thanks for the bug fix!

:tada: 

There is also this discrepancy that is worth mentioning: https://github.com/helix-editor/helix/issues/4544#issuecomment-1298518387

David-Else:
@ygabuev Thanks for fixing this this, it was driving me crazy! :)

PR #4559, opened 2022-11-01
Improve appearance of diagnostics messages

twharmon:
A few opinions:
1. The diagnostic messages need a right margin. It looks awful with characters on the edge.
2. Distinct diagnostic messages need a margin between them to indicate the messages are separate.
3. The maximum width of the area for messages needs to be significantly smaller. Long messages are hard to read when wrapped.
4. The messages need to be left aligned.

This is better:
image

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -38,6 +38,7 @@ jsonnet-language-server = { command = "jsonnet-language-server", args= ["-t", "-
 julia = { command = "julia", timeout = 60, args = [ "--startup-file=no", "--history-file=no", "--quiet", "-e", "using LanguageServer; runserver()", ] }
 kotlin-language-server = { command = "kotlin-language-server" }
 lean = { command = "lean", args = [ "--server" ] }
+ltex-ls = { command = "ltex-ls" }
 markdoc-ls = { command = "markdoc-ls", args = ["--stdio"] }
 marksman = { command = "marksman", args = ["server"] }
 metals = { command = "metals", config = { "isHttpEnabled" = true } }


archseer:
> 4\. The messages need to be left aligned.

The design was based on lsp-mode which right aligns: https://emacs-lsp.github.io/lsp-mode/

archseer:
> * The diagnostic messages need a right margin. It looks awful with characters on the edge.
> 
> * Distinct diagnostic messages need a margin between them to indicate the messages are separate.

That consumes more space on the UI though, the diagnostics area is already way too big on a smaller terminal.

> 3\. The maximum width of the area for messages needs to be significantly smaller. Long messages are hard to read when wrapped.

I already downsized this from 120 to 100 which was the lsp-mode default. 60 is even smaller than what used to be the code wrapping standard at 80. It works well if the diagnostic is mostly text, but wrapping code samples over several lines is a lot harder for me to read.

twharmon:
1. The link to emacs screen shot at least has the right margin. Surely we can at least change that.

2. As for the space between messages, if space is more important than readability, then let's not add the additional space in there.

Which language and scenario will produce diagnostic messages with that much code? I'd expect right-aligned code to be hard to read.

Manosmer:
> This is better: <the above image>

@twharmon Excuse me for the unrelated question, but how did you set up the mode names down in the statusline to their full form?

twharmon:
> > This is better: <the above image>
> 
> @twharmon Excuse me for the unrelated question, but how did you set up the mode names down in the statusline to their full form?
```toml
[editor.statusline]
mode.normal = "NORMAL"
mode.insert = "INSERT"
mode.select = "SELECT"
```

David-Else:
I think having the mode names in full looks much better, so I started a poll to see what people think: https://github.com/helix-editor/helix/discussions/4589

blitzerr:
The left vs right alignment and the addition space for readability  vs not can all be made configurable. 

As @archseer mentioned the case of smaller displays, they can go with compact mode and people with larger monitors can go for easy fit. The Diagnostic messages today are very hard to read.

twharmon:
@archseer Any new thoughts on this? Should the right margin be added? Or, is there any part of this you would like to be configurable?

PR #4562, opened 2022-11-02
Fix D unittest injection query.

gdamore:


the-mikedavis (approved):
Ah good catch, thanks!

Issue #4540, opened 2022-10-31
Add a workspace diagnostics indicator to the status line


dariooddenino:
Status lines have yellow/red icons with numbers to indicate how many warning/errors are present in the current buffer.

I think it would be useful to display the workspace total too there to get a quick visual feedback instead of having to manually input `<space>-D` to open the workspace diagnostics picker.

Some ideas:
1. Use a number format like `1/13`, meaning that the current buffer has 1 warning/error and the workspace has 13 in total.
2. Use two completely different sets of numbers/icons for buffer diagnostics and workspace diagnostics.
3. Add a workspace status line at the bottom that spans the whole application window width and which could hold more general information (like workspace diagnostics).

dariooddenino:
I was giving a quick look at this, and I have it down more or less but I'm confused by something.

We have both `lsp::DiagnosticSeverity`, which is what's stored in `context.editor.diagnostics` and `helix_core::diagnostic::Severity` returned by `context.doc.diagnostics()`.

Is there a reason for having two different data structures (that apparently have the same purpose) returned by two similar functions?

the-mikedavis:
`lsp::DiagnosticSeverity` is from the lsp_types crate https://docs.rs/lsp-types/latest/lsp_types/struct.DiagnosticSeverity.html while `helix_core::diagnostic::Severity` is custom. The lsp_types one is a struct with constants while the custom one is an Enum

PR #4569, opened 2022-11-02
Add global error/warning count statusline element

grv07:
#4540

For <SPACE>+d
image
For <SPACE>+D
image

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -207,40 +207,23 @@ where
     );
 }
 
-/// Gets the count pair (warnings, error) of warnings, error on document.
-fn diagnostics_count(doc: &Document) -> (i32, i32) {
-    doc.diagnostics().iter().fold((0, 0), |mut counts, diag| {
-        use helix_core::diagnostic::Severity;
-        match diag.severity {
-            Some(Severity::Warning) => counts.0 += 1,
-            Some(Severity::Error) | None => counts.1 += 1,
-            _ => {}
-        }
-        counts
-    })
-}
-
-/// Gets the count pair (warnings, error) of warnings, error on workspace.
-fn workspace_diagnostic_count(editor: &Editor) -> (i32, i32) {
-    editor
-        .diagnostics
-        .values()
-        .flatten()
-        .fold((0, 0), |mut counts, diag| {
-            match diag.severity {
-                Some(DiagnosticSeverity::WARNING) => counts.0 += 1,
-                Some(DiagnosticSeverity::ERROR) | None => counts.1 += 1,
-                _ => {}
-            }
-            counts
-        })
-}
-
 fn render_diagnostics<F>(context: &mut RenderContext, write: F)
 where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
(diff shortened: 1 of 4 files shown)


archseer:
The default components shouldn't be changed, this should be an additional gutter item.

the-mikedavis:
https://github.com/helix-editor/helix/pull/4535/files is a good example of this - instead of modifying the existing diagnostics statusline element, this should be a new statusline element that you can configure

helix-view/src/editor.rs
```diff
@@ -1259,6 +1262,21 @@ impl Editor {
...
+            .flatten()
+            .fold((0, 0), |mut counts, diag| {
+                match diag.severity {
+                    Some(DiagnosticSeverity::WARNING) => counts.0 += 1,
+                    Some(DiagnosticSeverity::ERROR) | None => counts.1 += 1,
+                    _ => {}
+                }
+                counts
+            })
+    }
```
the-mikedavis, line 1278: Especially since these (`workspace_diagnostic_count` and `diagnostics_count`) filter on severity I think these seem specific to the statusline element and should be inlined there: I don't think these functions would be re-used outside of this use-case

the-mikedavis:
Should this element only contain workspace diagnostics? Currently it has `document / workspace` but I think the workspace one could be its own element and then users can organize the two however they want

the-mikedavis:
If we have case with 1 doc error, 3 doc warnings, 5 workspace errors and 7 workspace warnings, the doc element would be `● 1 ● 3 ` and the workspace element would be `● 5 ● 7 `. We could style the element a little differently so it's possible to tell apart doc and workspace diagnostics, like `W ● 5 ● 7 `. Then you could have a configuration for the statusline to separate them however you like:

```toml
[editor.statusline]
right = ["diagnostics", "separator", "workspace-diagnostics"]
separator = "/"
```

grv07:
> If we have case with 1 doc error, 3 doc warnings, 5 workspace errors and 7 workspace warnings, the doc element would be `● 1 ● 3 ` and the workspace element would be `● 5 ● 7 `. We could style the element a little differently so it's possible to tell apart doc and workspace diagnostics, like `W ● 5 ● 7 `. Then you could have a configuration for the statusline to separate them however you like:
> 
> ```toml
> [editor.statusline]
> right = ["diagnostics", "separator", "workspace-diagnostics"]
> separator = "/"
> ```

Updated as suggested
image

helix-term/src/ui/statusline.rs
```diff
@@ -205,23 +207,40 @@ where
...
+            Some(Severity::Warning) => counts.0 += 1,
+            Some(Severity::Error) | None => counts.1 += 1,
+            _ => {}
+        }
+        counts
+    })
+}
+
+/// Gets the count pair (warnings, error) of warnings, error on workspace.
+fn workspace_diagnostic_count(editor: &Editor) -> (i32, i32) {
```
the-mikedavis, line 224: These can be inlined into `render_diagnostics` and `render_workspace_diagnostics` respectively

grv07: Done

the-mikedavis (approved):

the-mikedavis:
Thank you!

PR #4570, opened 2022-11-02
Fix panic from two windows editing the same document

### LINKED ISSUE #3978: Crash on delete in vertically-stacked split-screen
### Summary

- Had same document open in top and bottom windows.
- Deleted several lines of text from the top screen.

## Error

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Line index out of bounds: line index 52, Rope/RopeSlice line count 49', /var/home/aral/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.0/src/slice.rs:424:41
```

### Reproduction Steps

Not sure how to reproduce at the moment.

### Helix log

Nothing relevant in the Helix logs.

### Platform

Linux (Fedora Silverblue 37 Beta)

### Terminal Emulator

Black Box

### Helix Version

helix 22.08.1 (385ccdfc)

the-mikedavis:
This change has two fixes for a scenario that can come up if you use multiple windows to edit/view the same document.

The first fixes the panic in #3978 by clamping the syntax highlighting range to always be within the length of the document. The second fixes a minor UX issue where if you delete enough of a document, one window may be looking past the end of the document - in this case we ensure that the cursor is within view upon switching windows.

Closes #3978

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -236,7 +236,7 @@ impl EditorView {
                 .saturating_sub(1)
                 .min(last_line);
             let start = text.line_to_byte(offset.row.min(last_line));
-            let end = text.line_to_byte(last_visible_line.min(last_line) + 1);
+            let end = text.line_to_byte(last_visible_line + 1);
 
             start..end
         };
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -227,16 +227,16 @@ impl EditorView {
         _theme: &Theme,
     ) -> Box<dyn Iterator<Item = HighlightEvent> + 'doc> {
         let text = doc.text().slice(..);
-        let last_line = std::cmp::min(
-            // Saturating subs to make it inclusive zero indexing.
-            (offset.row + height as usize).saturating_sub(1),
-            doc.text().len_lines().saturating_sub(1),
-        );
 
         let range = {
-            // calculate viewport byte ranges
-            let start = text.line_to_byte(offset.row);
-            let end = text.line_to_byte(last_line + 1);
+            // Calculate viewport byte ranges:
+            // Saturating subs to make it inclusive zero indexing.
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -227,16 +227,14 @@ impl EditorView {
...
     ) -> Box<dyn Iterator<Item = HighlightEvent> + 'doc> {
         let text = doc.text().slice(..);
-        let last_line = std::cmp::min(
-            // Saturating subs to make it inclusive zero indexing.
-            (offset.row + height as usize).saturating_sub(1),
-            doc.text().len_lines().saturating_sub(1),
-        );
+        // Saturating subs to make it inclusive zero indexing.
+        let last_line = doc.text().len_lines().saturating_sub(1);
+        let view_end_line = (offset.row + height as usize).saturating_sub(1);
```
archseer, line 232: I'd move both of these into the `let range` block, and rename `let last_visible_line = (offset.row + height as usize).saturating_sub(1).min(last_line)`

helix-term/src/ui/editor.rs
```diff
@@ -227,16 +227,16 @@ impl EditorView {
...
-            let start = text.line_to_byte(offset.row);
-            let end = text.line_to_byte(last_line + 1);
+            // Calculate viewport byte ranges:
+            // Saturating subs to make it inclusive zero indexing.
+            let last_line = doc.text().len_lines().saturating_sub(1);
+            let last_visible_line = (offset.row + height as usize)
+                .saturating_sub(1)
+                .min(last_line);
+            let start = text.line_to_byte(offset.row.min(last_line));
+            let end = text.line_to_byte(last_visible_line.min(last_line) + 1);
```
archseer, line 239:
```suggestion
            let end = text.line_to_byte(last_visible_line + 1);
```

`.min()` already happens on L237

PR #4572, opened 2022-11-02
Improvements to Meson syntax.

gdamore:
This updates the Helix tree-sitter for meson to https://github.com/staysail/meson.  I believe that this grammar fixes at least one bug, and the queries are a bit richer with more idiomatic support for Helix.

(Admittedly I didn't know about the present meson tree-sitter before I created mine -- see the discussion with the author of the other tree-sitter here: https://github.com/bearcove/tree-sitter-meson/issues/7

### DIFF
--- a/runtime/queries/meson/highlights.scm
+++ b/runtime/queries/meson/highlights.scm
@@ -1,62 +1,62 @@
-(string_literal) @string
+(comment) @comment
 
-(boolean_literal) @constant.builtin.boolean
-(integer_literal) @constant.numeric.integer
-
-(comment) @comment.line
-(function_id) @function
-(keyword_arg_key) @variable.other.member
-(id_expression) @variable
+; these are listed first, because they override keyword queries
+(function_expression (identifier) @function)
 
 [
-  "if"
-  "elif"
-  "else"
-  "endif"
-] @keyword.control.conditional
-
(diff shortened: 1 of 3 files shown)


runtime/queries/meson/highlights.scm
```diff
@@ -1,62 +1,62 @@
...
+    (endforeach)
+    (break)
+    (continue)
+] @keyword.control.repeat
+
+(boolean_literal) @constant.builtin.boolean
+(int_literal) @constant.numeric.integer
+
+(keyword_argument keyword: (identifier) @variable.parameter)
+(escape_sequence) @string.special
```
the-mikedavis, line 48:
```suggestion
(escape_sequence) @constant.character.escape
```

runtime/queries/meson/highlights.scm
```diff
@@ -1,62 +1,62 @@
...
-  "["
-  "]"
-  "{"
-  "}"
-] @punctuation.bracket
+    (string_literal)
+    (fstring_literal)
+] @string
+
+(identifier) @variable.other
```
the-mikedavis, line 62:
```suggestion
(identifier) @variable
```

the-mikedavis (approved):

PR #4573, opened 2022-11-03
Add command to select surrounding whitespace

### LINKED ISSUE #4509: mis to match inner whitespace
Given a piece of text:

```
foo: |
    █hello
```

typing `mis` could select all the whitespace, then `d` would produce the editor state

```
foo: |█hello
```

Stonks3141:
Add `mis` and `mas` commands to select the whitespace surrounding the cursor.

If the cursor is on a whitespace character, the command will select all surrounding whitespace, including newlines and carriage returns. If the cursor is not on a whitespace character but adjacent to one, it will not select anything. This PR closes #4509.

https://user-images.githubusercontent.com/82178396/199630802-85261263-8844-4074-b1aa-d15a65f6aeb2.mp4

### DIFF
--- a/helix-core/src/textobject.rs
+++ b/helix-core/src/textobject.rs
@@ -112,14 +112,17 @@ pub fn textobject_word(
     }
 }
 
+// `textobject` and `count` are unused for now, this textobject behaves the same
+// whether it is inside or around.
 pub fn textobject_whitespace(
     slice: RopeSlice,
     range: Range,
-    textobject: TextObject,
+    _textobject: TextObject,
     _count: usize,
 ) -> Range {
     let pos = range.cursor(slice);
 
+    // find the number of consecutive whitespace chars in front of the cursor
     let whitespace_start = pos
         - slice
             .chars_at(slice.len_chars())
--- a/helix-core/src/textobject.rs
+++ b/helix-core/src/textobject.rs
@@ -128,14 +128,14 @@ pub fn textobject_whitespace(
             .chars_at(slice.len_chars())
             .reversed()
             .skip(slice.len_chars() - pos - 1)
-            .take_while(|c| char_is_whitespace(*c) || *c == '\r' || *c == '\n')
+            .take_while(|c| c.is_whitespace())
             .count()
         + 1;
 
     let whitespace_end = pos
         + slice
             .chars_at(pos)
-            .take_while(|c| char_is_whitespace(*c) || *c == '\r' || *c == '\n')
+            .take_while(|c| c.is_whitespace())
             .count();
 
     Range::new(whitespace_start, whitespace_end)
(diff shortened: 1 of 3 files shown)


helix-core/src/textobject.rs
```diff
@@ -112,6 +112,35 @@ pub fn textobject_word(
...
+) -> Range {
+    let pos = range.cursor(slice);
+
+    // find the number of consecutive whitespace chars in front of the cursor
+    let whitespace_start = pos
+        - slice
+            .chars_at(slice.len_chars())
+            .reversed()
+            .skip(slice.len_chars() - pos - 1)
+            .take_while(|c| char_is_whitespace(*c) || *c == '\r' || *c == '\n')
```
SoraTenshi, line 131:
```suggestion
            .take_while(|*c| c.is_whitespace())
```

would be a bit better i think, as it:
1. in the standard library
2. exactly what you want.

helix-core/src/textobject.rs
```diff
@@ -112,6 +112,35 @@ pub fn textobject_word(
...
+            .count()
+        + 1;
+
+    let whitespace_end = pos
+        + slice
+            .chars_at(pos)
+            .take_while(|c| char_is_whitespace(*c) || *c == '\r' || *c == '\n')
```
SoraTenshi, line 138:
```suggestion
            .take_while(|*c| c.is_whitespace())
```

Same as above

helix-core/src/textobject.rs
Stonks3141: I had been following #4509’s original suggestion that the selection should span newlines, and `char_is_whitespace` returns false for newlines and carriage returns ([https://github.com/helix-editor/helix/blob/master/helix-core/src/chars.rs#L43](https://github.com/helix-editor/helix/blob/master/helix-core/src/chars.rs#L43)). Are you suggesting that it wpould be better to not capture newlines?

Stonks3141: Never mind, I didn't see that you were suggesting `char::is_whitespace` and not `char_is_whitespace`. I'll apply the changes.

Stonks3141:
There are a lot of commits for this now, a squash merge would probably be better.

SoraTenshi (approved):

asvln (approved):
This is a vital match object in my opinion.

bluthej:
I see that this PR has been closed but the issue is still open.

Has this been fixed/elsewhere? Or has @Stonks3141 just abandoned the PR?

I'm looking for an opportunity to start contributing so if the issue is up for grabs I'd like to give it a shot!

Stonks3141:
I abandoned this patch since it’s 9 months old and I no longer use Helix. Feel free to take it over! You can get the changes from this PR by running `curl -fL https://github.com/helix-editor/helix/pull/4573.patch | git am` in your clone of Helix.

Issue #4125, opened 2022-10-06
Panic on XF86 keycodes


Itay123TheKing:
### Summary

I know this is kind of an edge case, but if your system does not do anything with volume control/media playback control keys, they pass through to Helix, causing a panic with the following message:
```
thread 'main' panicked at 'internal error: entered unreachable code: Shouldn't get this key without enabling DISAMBIGUATE_ESCAPE_CODES in crossterm', helix-view/src/keyboard.rs:158:40
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Reproduction Steps

So this wouldn't be too easy if you don't have access to the guts of your operating system and desktop environment or window manager, but with i3 you can edit your config and comment any line that starts with `bindsym XF86____`, save it and reload your config, then in Helix send to your PC (presumably with a keyboard button) the keycode corresponding to what you just commented out. 

### Helix log

The log didn't contain anything in regards to this error

### Platform

EndeavourOS Arch Linux with i3-gaps

### Terminal Emulator

Kitty

### Helix Version

helix 22.08.1 (5dbca0fc)

sudormrfbin:
This is probably an upstream bug since we indeed don't have [`DISAMBIGUATE_ESCAPE_CODES`][1] enabled which seems to be the prerequisite to getting the [`Media`][2] keycodes which is where the code panics:

https://github.com/helix-editor/helix/blob/4d4be0e880efd136100439cc34e8c526d299ec34/helix-view/src/keyboard.rs#L157-L160

[1]: https://docs.rs/crossterm/latest/crossterm/event/struct.KeyboardEnhancementFlags.html#associatedconstant.DISAMBIGUATE_ESCAPE_CODES

[2]: https://docs.rs/crossterm/latest/crossterm/event/enum.KeyCode.html#variant.Media

ngraham20:
Looks like this also happens when I press the "Pause" key (different than the play/pause media key)

Itay123TheKing:
@ngraham20 yeah it's a problem with all media control keys, not just play/pause. It's also Mute, Next tack, Previous track, Volume up/down, etc.

cor:
Same issue here, specifically when I use NixOS inside of a Parallels VM with macOS as the host OS.

groves:
@ngraham20 and @cor, are you using kitty as your terminal emulator, too? Or another terminal?

groves:
[...]

PR #4576, opened 2022-11-03
Don't panic if we get an unexpected key code

groves:
It seems like we get media play/pause/mute/etc keys despite not emitting DISAMBIGUATE_ESCAPE_CODES(#4125).

Make crossterm::event::KeyCode a TryInto for KeyCode instead of Into to let us ignore keys we don't care about.

### DIFF
--- a/helix-view/src/keyboard.rs
+++ b/helix-view/src/keyboard.rs
@@ -1,3 +1,4 @@
+use anyhow::{bail, Error};
 use bitflags::bitflags;
 
 bitflags! {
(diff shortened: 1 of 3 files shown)


helix-view/src/keyboard.rs
```diff
@@ -155,9 +157,9 @@ impl From<crossterm::event::KeyCode> for KeyCode {
             | CKeyCode::Menu
             | CKeyCode::KeypadBegin
             | CKeyCode::Media(_)
-            | CKeyCode::Modifier(_) => unreachable!(
+            | CKeyCode::Modifier(_) => bail!(
                 "Shouldn't get this key without enabling DISAMBIGUATE_ESCAPE_CODES in crossterm"
```
the-mikedavis, line 161:
```suggestion
                "Shouldn't get key {:?} without enabling DISAMBIGUATE_ESCAPE_CODES in crossterm",
                val
```

Since the unreachable doesn't point at this line anymore, this should make it easier to debug if it shows up in the logfile

archseer:
Rather than turning this into a `try_from`, why not add all these events to the internal event struct and properly handle them (or ignore them inside the editor)? That way we don't have to deal with intricacies of enabling `DISAMBIGUATE_ESCAPE_CODES` in a backwards-compatible way (i.e. not calling it on terminals with no support), but if we do get these events then they're at least handled.

PR #4582, opened 2022-11-03
Use language=bash when shebang line uses zsh

throwaway-helix-zsh:
This PR makes the editor use language=bash when the shebang line uses zsh. We already use language=bash for zsh related files (~/.zshrc, ~/.zshenv etc.), so it's expected for the same to be true for shebang lines as well.


throwaway-helix-zsh:
@the-mikedavis Thanks for triggering the pipelines. This is ready to merge :)

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4584, opened 2022-11-03
Add Haxe language support

spindlebink:
There may be more `roots` indicators to add, but I'm a Haxe newbie so I only added the ones I know about for now.


kirawi:
Can you run `cargo xtask docgen` and commit the changes?

languages.toml
```diff
@@ -1964,3 +1964,17 @@ roots = []
...
 source = { git = "https://github.com/justinmk/tree-sitter-ini", rev = "4d247fb876b4ae6b347687de4a179511bf67fcbc" }
+
+[[language]]
+name = "haxe"
+scope = "source.haxe"
+injection-regex = "haxe"
+file-types = ["hx"]
+comment-token = "//"
+indent = { tab-width = 2, unit = "\t" }
+roots = ["hxml", "khafile.js"]
```
the-mikedavis, line 1975: I don't think `roots` supports file extensions. I'm not very familiar with Haxe but could we use the `haxelib.json` file for this?

runtime/queries/haxe/highlights.scm
```diff
@@ -0,0 +1,71 @@
...
+(type_params
+  "<" @punctuation.bracket
+  ">" @punctuation.bracket)
+
+; Literals
+; --------
+[(null) (keyword)] @keyword
+[(type) (literal)] @type
+[(builtin_type)] @type.builtin
+[(integer) (float)] @number
```
the-mikedavis, line 31:
```suggestion
(integer) @constant.numeric.integer
(float) @constant.numeric.float
```

runtime/queries/haxe/highlights.scm
```diff
@@ -0,0 +1,71 @@
...
+(string) @string
+(bool) @constant
```
the-mikedavis, line 33:
```suggestion
(bool) @constant.builtin.boolean
```

runtime/queries/haxe/highlights.scm
```diff
@@ -0,0 +1,71 @@
...
+(operator) @operator
+
+; Interpolation
+; -------------
+(interpolation "$" @punctuation.special)
+(interpolation
+  "${" @punctuation.special
+  "}" @punctuation.special
+) @embedded
```
the-mikedavis, line 42:
```suggestion
)
```

runtime/queries/haxe/highlights.scm
```diff
@@ -0,0 +1,71 @@
+(identifier) @type
```
the-mikedavis, line 1: This needs to be moved to the bottom of the file: when it's at the top it takes precedence over all other uses of `(identifier)` so everything looks like a type.

(This is because neovim queries are in reverse order and these queries are in neovim format)

runtime/queries/haxe/injections.scm
```diff
@@ -0,0 +1,3 @@
+(interpolation) @haxe
+(comment) @jsdoc
+(comment) @comment
```
the-mikedavis, line 3: We can use the same trick as in javascript/typescript for these

```scm
; Parse JSDoc annotations in multiline comments

((comment) @injection.content
 (#set! injection.language "jsdoc")
 (#match? @injection.content "^/\\*+"))

; Parse general tags in single line comments

((comment) @injection.content
 (#set! injection.language "comment")
 (#match? @injection.content "^//"))
```

runtime/queries/haxe/injections.scm
```diff
@@ -0,0 +1,3 @@
+(interpolation) @haxe
```
the-mikedavis, line 1:
```suggestion
```

the-mikedavis (commented):
It looks like the parser and queries for tree-sitter-haxe need some work: I see some parsing errors on files like https://github.com/HaxeFoundation/haxelib/blob/development/src/legacyhaxelib/Main.hx

languages.toml
spindlebink: I included the `hxml` root because I noticed `sln` and `csproj` in the roots for C#, but that also may be an error in the first place if indeed `roots` doesn't work like that.

spindlebink:
Neither the parser nor queries are my work: I directly copied the queries from the repo into the `queries` folder, noticed the highlighting seemed to be working, and pushed without inspecting the parsing results. I'll address the review you've put here, but also I'm just beginning to learn Haxe, so for actual iteration on the parser we'd need somebody who knows the syntax better than I do.

languages.toml
the-mikedavis: It looks like we don't support extensions for roots; it's a `Path::exists` call: https://github.com/helix-editor/helix/blob/4ec2a21c6e21ab4e515f1bd7ee1049094af2a6b2/helix-core/src/lib.rs#L68

Issue #4586, opened 2022-11-04
Panic on starting a command with space


ci4ic4:
### Summary

Starting a line command with a space leads to a panic.

### Reproduction Steps

 

I tried this:

1. `hx`
2. Enter colon to go to command line
3. Enter space

I expected this to happen:
1. nothing, I can return to normal mode.
Instead, this happened:
Panic

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
`~/.cache/helix/helix.log` remains empty. 

RUST_BACKTRACE=full output under NetBSD follows.
```

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-term/src/commands/typed.rs:2206:41
stack backtrace:
   0:        0x10c1843d3 - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h7dd69a5540e54fa1
   1:        0x10b93bdfe - core::fmt::write::hdc7684da7ffcab3f
   2:        0x10c1612b5 - std::io::Write::write_fmt::hbe98920af2ef3b1c
   3:        0x10c1703fe - std::panicking::default_hook::{{closure}}::hd00883c5895d31be
   4:        0x10c17004f - std::panicking::default_hook::h37089c0e1ca56b68
   5:        0x10c170c61 - std::panicking::rust_panic_with_hook::h758a8f675318119b
   6:        0x10c1846bc - std::panicking::begin_panic_handler::{{closure}}::hff5edfae4168c537
   7:        0x10c1844f4 - std::sys_common::backtrace::__rust_end_short_backtrace::hba703bc5f07bca51
   8:        0x10c17061d - rust_begin_unwind
   9:        0x10b8b84d3 - core::panicking::panic_fmt::he94bae07399fe725
  10:        0x10b8b839d - core::panicking::panic::hd6cc19e938b8bf80
  11:        0x10be4b361 - helix_term::commands::typed::command_mode::{{closure}}::hbbb8082591504461
  12:        0x10be53e6b - helix_term::ui::prompt::Prompt::insert_char::hf05b00dceb6751de
  13:        0x10be55c95 - <helix_term::ui::prompt::Prompt as helix_term::compositor::Component>::handle_event::h33fa4adb614ab2d4
  14:        0x10bc284a6 - helix_term::compositor::Compositor::handle_event::hd734752ff674a8bf
[...]

Issue #4563, opened 2022-11-02
Completion not working with commands if no parameter was passed to them


seshotake:
### Summary

Completion not working in command mode if there is not parameter 

Also I noticed that if use alias `:o` completion works  without parameter but instead the space between them was deleted

![using-alias](https://user-images.githubusercontent.com/116971836/199372522-37309725-e046-45c5-94ed-2ded5e89a410.png)

when selected:
![selected](https://user-images.githubusercontent.com/116971836/199372900-15c30c5b-f1af-4a8d-a939-2d3547af8c0f.png)

### Reproduction Steps

I tried this:

1. `:open` or `:theme`

I expected this to happen: show completions

Instead, this happened: nothing shown

### Platform

Linux

### Terminal Emulator

zellij

### Helix Version

helix 22.08.1 (c803ef87)

archseer:
Could be a regression from https://github.com/helix-editor/helix/pull/4316

\cc @mitsuhiko

PR #4587, opened 2022-11-04
Fix whitespace handling in command-mode completion

the-mikedavis:
A few fixes/improvements here for shellwords-based completion in command_mode (`:`):

* Fix completion when the last item in the prompt is a space
* Fix a panic for the sequence `:<space>`
* Escape values in the completion list
* Fix an edge-case in completion when the input is escaped

Closes https://github.com/helix-editor/helix/issues/4586
Closes https://github.com/helix-editor/helix/issues/4563

### DIFF
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -1,6 +1,5 @@
 use crate::compositor::{Component, Compositor, Context, Event, EventResult};
 use crate::{alt, ctrl, key, shift, ui};
-use helix_core::shellwords;
 use helix_view::input::KeyEvent;
 use helix_view::keyboard::KeyCode;
 use std::{borrow::Cow, ops::RangeFrom};
@@ -336,10 +335,7 @@ impl Prompt {
 
         let (range, item) = &self.completion[index];
 
-        // since we are using shellwords to parse arguments, make sure
-        // that whitespace in files is properly escaped.
-        let item = shellwords::escape(item);
-        self.line.replace_range(range.clone(), &item);
+        self.line.replace_range(range.clone(), item);
 
         self.move_end();
     }
--- a/helix-core/src/shellwords.rs
+++ b/helix-core/src/shellwords.rs
@@ -17,18 +17,18 @@ pub fn escape(input: &str) -> Cow<'_, str> {
     }
 }
 
+enum State {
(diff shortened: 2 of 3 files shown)


ci4ic4:
With this commit helix no loner crashes when command is started with a space.

helix-term/src/ui/prompt.rs
```diff
@@ -336,10 +335,7 @@ impl Prompt {
 
         let (range, item) = &self.completion[index];
 
-        // since we are using shellwords to parse arguments, make sure
-        // that whitespace in files is properly escaped.
-        let item = shellwords::escape(item);
-        self.line.replace_range(range.clone(), &item);
+        self.line.replace_range(range.clone(), item);
```
gavincrawford, line 338: Should we pass the entire item instead of just the reference?

the-mikedavis: `item` is already a `&Cow<str>` here. The old code was taking a reference because `shellwords::escape` returns a `Cow<str>`

gavincrawford: Got it. LGTM

gavincrawford (approved):

helix-term/src/ui/prompt.rs
```diff
@@ -336,10 +335,7 @@ impl Prompt {
 
         let (range, item) = &self.completion[index];
 
-        // since we are using shellwords to parse arguments, make sure
-        // that whitespace in files is properly escaped.
-        let item = shellwords::escape(item);
-        self.line.replace_range(range.clone(), &item);
```
archseer, line 342: Doesn't this undo the previous PR? The idea was that it would complete paths with spaces in quotes

the-mikedavis: This part moved into the flatmap in `command_mode`: https://github.com/helix-editor/helix/pull/4587/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R2227

The nice part of moving it there is that the completion results in the prompt show as escaped which matches how you have to input them (for example a file "a b.txt" would show as `a\ b.txt` in the prompt's completion items).

ci4ic4 (approved):
As far as my understanding goes, this resolves my initial issue.

PR #4595, opened 2022-11-05
Use OSC 52 for yanking on Termux

### LINKED ISSUE #4594: Do not use `termux-clipboard-set` on Termux
Currently Helix uses `termux-clipboard-set` for yanking in Termux on Android:

https://github.com/helix-editor/helix/blob/master/helix-view/src/clipboard.rs#L131-L135

However, the `termux-clipboard-set` command actually has to call another app `Termux API`, making it pretty slow:

```console
$ time termux-clipboard-set "test"

real    0m2.119s
user    0m0.000s
sys     0m0.024s
```

So every time I do `<space>y` on Android, the editor hangs for 2 seconds.

The twist is that... we don't really have to use `termux-clipboard-set`. Termux supports OSC 52 and it's actually instant. So I suggest dropping `termux-clipboard-set` and only keeping `termux-clipboard-get` (which is also slow but it is what it is). I'm working on a fix right now.

Spamming @dead10ck since he uses Termux a lot too lol.

xJonathanLEI:
Fixes #4594.

Yanking to clipboard in Termux is now instant.

### DIFF
--- a/helix-view/src/clipboard.rs
+++ b/helix-view/src/clipboard.rs
@@ -128,11 +128,8 @@ pub fn get_clipboard_provider() -> Box<dyn ClipboardProvider> {
             paste => "win32yank.exe", "-o", "--lf";
             copy => "win32yank.exe", "-i", "--crlf";
         }
-    } else if binary_exists("termux-clipboard-set") && binary_exists("termux-clipboard-get") {
-        command_provider! {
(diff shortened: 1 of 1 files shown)


xJonathanLEI:
Closing this as not needed. See original issue for more info.

Issue #4298, opened 2022-10-15
Toggle for mode dependent cursor block color / cursor block color theming rewrite


ivenw:
I really like the colored mode indicator in the statusline toggled by `editor.color-mode`. I am desiring a similar option for coloring the cursor block according to the mode (i.e. `editor.cursor-color-mode`). So I went about looking at implementing this to potentially make a PR for this down the line.

In researching if a similar effort is already worked on I stumbled upon #1337 and the tangentially related #1833. To my surprise the feature is essentially implemented but is difficult to discover or to turn into an option since many themes implement the overwriting `ui.cursor.primary` setting, as pointed out in #1337.

I propose to deprecate `ui.cursor.primary`, with `ui.cursor` taking its place for theming the primary cursor block. The current role of `ui.cursor` of theming the secondary cursor block would then be taken by the new `ui.cursor.secondary` theme parameter. (For consistency, this will also deprecate `ui.selection.primary` and introduce `ui.selection.secondary` following the same logic).

While this is a breaking change, I think it comes with a number of advantages that will pay off going forward. 

1. It makes logically more sense to declare the colors of the primary selection and cursor first. If the theme desires, it can set  colors for the less common secondary selection as well.
2. From this follows also a clearer fallback behavior in the absence of certain cursor theming parameters.
3. We can easily implement mode-dependent cursor theming as an option.

With this proposed change the cursor theming logic would as follows:
- `ui.selection` sets the primary selection color. Mandatory.
- `ui.cursor` sets the primary cursor block color. Sets the primary cursor block color in `NORMAL` mode if `editor.cursor-color-mode true`. Fallback to `ui.cursor`. Fallback to `ui.selection`.
- `ui.cursor.insert` sets the primary cursor block color in `INSERT` mode if `editor.cursor-color-mode true`. Fallback to `ui.cursor`.
- `ui.cursor.select` sets the primary cursor block color in `SELECT` mode if `editor.cursor-color-mode true`. Fallback to `ui.cursor`.
- `ui.selection.secondary` sets the secondary selection color. Fallback to `ui.selection`
- `ui.cursor.secondary` sets the secondary cursor block color. Sets the secondary cursor block color in `NORMAL` mode if `editor.cursor-color-mode true`. Fallback to `ui.cursor`.
- `ui.cursor.secondary.insert` sets the secondary cursor block color in `INSERT` mode if `editor.cursor-color-mode true`. Fallback to `ui.cursor.secondary`.
[...]

PR #2366, opened 2022-05-02
change the behavior of `ui.cursor.insert` and `ui.cursor.select` when `ui.cursor.primary` is present

oati:
Fixes #1337

Previously, `ui.cursor.insert` and `ui.cursor.select` only applied to non-primary cursors when `ui.cursor.primary` was present.
This PR changes this behavior to make `ui.cursor.insert` and `ui.cursor.select` only affect the primary cursor when `ui.cursor.primary` is set.

Issue #1833, opened 2022-03-18
Add theming for primary cursor and/or selection in themes that lack them


Omnikar:
Some of the themes in `runtime/themes` make no distinction between primary and other cursors/selections; this, I think, should be addressed to prevent confusions like #1780 and in order to let people using said themes to make full use of the multiple cursors system.

List of themes that do not distinguish between primary and nonprimary cursors/selections:

* `bogster`
* `ingrid`
* `rose_pine`
* `rose_pine_dawn`
* `serika-light` *in making this list, I've realized that these two theme names mistakenly use hyphens instead of underscores
* `serika-dark`
* `everforest_light`
* `everforest_dark`

CptPotato:
I agree with this, but maybe #1337 should be resolved first?

Since `cursor.primary` sets the color for both `insert` and `normal` I didn't use it in my themes. Distinguishing between `NOR`/`INS`/`VIS` mode was/is more important to me.

LollipopFt:
@Omnikar not sure if it's me, but `gruvbox`, `dark_plus`, `monokai`, `onedark`, `spacebones_light`, `base16_default` seem to have the same problem.

Omnikar:
I just checked through those, and they seem to fall into two categories:

`gruvbox`, `spacebones_light`  
These themes provide a theming for the primary cursor, but it is very similar to the normal cursor theme, and hard to notice.

`dark_plus`, `monokai`, `onedark`, `base16_default`

These themes don't provide a theming for the primary cursor, but do provide a theming for the primary selection, meaning you can distinguish cursors only if they have selections with them.

Issue #1337, opened 2021-12-22
`ui.cursor.primary` always overrides `ui.cursor.insert`


bbodi:
If `ui.cursor.primary` is set in a theme, it always overrides e.g. `ui.cursor.insert`.

E.g. currently in gruvbox theme one can't tell where is the cursor in a selection because it has the same color as the selection:
![image](https://user-images.githubusercontent.com/2362136/147093097-61d8bc68-da8c-4d01-8de0-1b3f441b5159.png)

If I change the `ui.cursor.primary` value of the theme file, the issue is solved:
![image](https://user-images.githubusercontent.com/2362136/147093221-74feecdb-a33d-411a-b176-e77e27541f33.png)

However, I also would like to give `ui.cursor.insert` a different color, so I know if I am in insert mode, but currently it is not possible

matoous:
I am not sure whether the behavior should be changed 🤔 if the theme specifies different cursor colors for insert/normal mode and doesn't set the primary cursor color this will work fine, maybe we should adjust the theme instead? Or am I missing something?

bbodi:
Now it is not possible to set a color to normal mode primary cursor, and a different one for insert mode cursor.

matoous:
But wouldn't that require also having `ui.cursor.primary_select` and `ui.cursor.primary_insert`? What should the precedence be, e.g. if I have `ui.cursor.primary` and `ui.cursor.insert` if we give precedence to `ui.cursor.primary` than you can't distinguish between `select`/`insert` and other modes. If we give precedence to `ui.cursor.insert` than we can't distinguish primary vs. additional selections.

oati:
One solution would be to make `ui.cursor.insert` only apply to the primary cursor when `ui.cursor.primary` is present.

oati:
Going off of that more, since `ui.cursor` is doing nothing when `ui.cursor.primary` is present, why not make `ui.cursor` apply to all non-primary cursors?

PR #4596, opened 2022-11-05
[WIP] feat: Setting toggle for mode dependent cursor block color

ivenw:
This PR addresses #4298, #1833, #1337 and offering an alternative solution to #2366.

It follows the implementation suggested #4298. As part of this implementation, this PR immediately deprecates the parent theming field `ui.cursor.primary` and its children, replacing them with the `ui.cursor.secondary` family, as discussed in #4298.

For demonstration purposes, the only internal theme updated with this change is `dracula.toml`.

TODO:
- Update the remaining in-repo themes to the new cursor theming convention after implementation is approved.


helix-term/src/ui/editor.rs
```diff
@@ -337,48 +332,71 @@ impl EditorView {
         doc: &Document,
         view: &View,
         theme: &Theme,
-        cursor_shape_config: &CursorShapeConfig,
+        config: &DynGuard<Config>,
```
the-mikedavis, line 335:
```suggestion
        config: &helix_view::editor::Config,
```

see `render_text_highlights` which also takes config. This will eliminate the new imports.

helix-view/src/editor.rs
```diff
@@ -174,6 +174,8 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    /// Whether the cursor matches the mode as specified by `ui.cursor.insert` and `ui.cursor.select`. Defaults to `false`.
```
the-mikedavis, line 177:
```suggestion
    /// Whether to theme the cursor according to the editor mode. Defaults to `false`.
```

the-mikedavis: We might want to clarify the docs for `color_modes` too: "Whether to theme the mode statusline element according to the editor mode. Defaults to `false`"

the-mikedavis (commented):
This would also need some changes added to https://github.com/helix-editor/helix/blob/master/book/src/themes.md

the-mikedavis:
I like the look of this; I think this is a sane way to tackle the cursor scopes 👍

Before we proceed - I'm curious what you think @archseer?

ivenw:
I am still interested in implementing this. Are we waiting on input from @archseer ?

gibbz00:
Maybe I'm a a tad bit too tired to fully understand this PR,  but aren't we overcomplicating things this feature?

Correct me if I'm wrong, this PR makes it possible to have specific primary colors for the different modes right?

Why can't we just implement a `ui.cursor.{normal,selected,match,instert}.primary`?

Where ui.cursor(.mode) is the fallback if primary isn't explicitly set. Only difference in behaviour would be that ui.cursor.mode.primary falls back to ui.cursor.mode and not ui.primary. 

No complex state driven theming with with the  `edotor.cursor-color-mode` boolean and inconsistent fallbacks. No `ui.cursor.secondary.mode` key added, this is what ui.cursor is. Allows for users to slowly get introduced to more complex behaviour by their own pursuit of the *primary key. Similar to how one can do `printf` in python without all the ` public static void main(string[] args)` if you get what I mean. And most importantly; no deprecation of peoples theme overrides and breaking changes to each and every theme.

archseer:
There was a similar discussion on https://github.com/helix-editor/helix/pull/2366

archseer:
I think this should be solved on the theme level for sure, no user config options

archseer:
`ui.cursor.{insert,select}.primary` seems like a good solution if we can determine what the fallback should be

gibbz00:
Oh, thanks, must have missed the PR for #1337. Replying here though to preserve the discussion context.

> ui.cursor.{insert,select}.primary seems like a good solution if we can determine what the fallback should be

Mmm as @the-mikedavis pointed out in #4298, default fallback behavior for themes is a.b.c falls back to a.b which falls back to a.

If this principle is followed, then `ui.cursor.selected.primary` should fall back to `ui.cursor`. This is on the other hand not in line with the current behavior in which `ui.cursor.primary` overrides `ui.cursor.selected`. But such a fix would be minor to the themes which choose to leverage primary cursor differentiation: If the `ui.cursor.selected` property exists, just make sure to create a `ui.cursor.selected.primary` with the same properties as `ui.cursor.priamry`. Same thing for match and insert modes.

After reading #2366 on the other hand, it's clear that it's possible to add this feature with no change to current behavior whatsoever. That is; implement `ui.cursor.primary.mode` and keep fallback behavior consistent. (Opposed to `ui.cursor.mode.primary`) Meaning; we wouldn't have to also edit each and every theme in the process.

When I come to think about it, I personally am starting to prefer this solution. It explains why `ui.cursor.primary` is overwriting things as is, in addition to believing it would be a much simpler implementation. So much so that I went a head and successfully tried this myself. It only adds 6 lines to editor.rs, and in so also making it possible to specifically set cursors for normal mode, whilst using a fallback for the rest. (This PR adds 23 lines, the other 9, just for the differentiation parts).

Anyhow, I created a PR where the changes can be viewed, thinking it would be easier than asking for changes in the respective PRs. Documentation updates have also been prepared. #5130

ivenw:
@gibbz00 made very compelling arguments both here and in #5130.

All things considered, I think #5130 is the better solution for now and I'm happy to endorse it.

When #5130 is merged I will close this PR.

archseer:
Replaced by #5130

PR #4601, opened 2022-11-05
Improve Ruby TextObjects

danillos:
This PR Refactor current TextObjects and improves it adding:

- Adds Singleton Methods
- Adds Module
- Adds `@parameter.around`

### DIFF
--- a/runtime/queries/ruby/textobjects.scm
+++ b/runtime/queries/ruby/textobjects.scm
@@ -24,6 +24,17 @@
 (method (identifier) (method_parameters)
   (_)+ @function.inside)
   
+(method (identifier) !parameters
+  (_)+  @function.inside)
+
+(singleton_method) @function.around
+
+(singleton_method (identifier) (method_parameters)
+  (_)+ @function.inside)
+  
+(singleton_method (identifier) !parameters
+  (_)+  @function.inside)
+
 (do_block !parameters
   (_)+ @function.inside)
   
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
It looks like most of these nodes have `body` fields which can be used to capture the `@textobject.inside` parts more accurately. For example:

```scm
(singleton_method
  body: (_)? @function.inside) @function.around
```

should cover the new `singleton_method` patterns.

(`:tree-sitter-subtree` currently misrepresents these body fields for some nodes, https://github.com/helix-editor/helix/pull/4606)

danillos:
Hi @the-mikedavis, I was testing using the Tree from https://tree-sitter.github.io/tree-sitter/playground that prints:

```
program [0, 0] - [3, 0]
  singleton_method [0, 0] - [2, 3]
    object: self [0, 4] - [0, 8]
    name: identifier [0, 9] - [0, 13]
    identifier [1, 2] - [1, 6]
```

for the code

```ruby
def self.test
  test
end
```

Do we have a better way to see the Tree?

the-mikedavis:
The ruby parser on the tree-sitter website is out of date. You can use tree-sitter-cli and to get the exact results that helix sees you can clone tree-sitter-ruby and check out the revision listed in `languages.toml` and run `tree-sitter parse` against that. I see:

```
(program [2, 0] - [8, 0]
  (singleton_method [2, 0] - [4, 3]
    object: (self [2, 4] - [2, 8])
    name: (identifier [2, 9] - [2, 20])
    body: (body_statement [3, 2] - [3, 6]
      (true [3, 2] - [3, 6]))))
```

danillos:
@the-mikedavis thank you. I did a small refactor on the current TextObjects based on the Tree printed by `tree-sitter parse` and improved it adding some missing cases.

PR #4614, opened 2022-11-06
Run clippy on workspace in CI

xJonathanLEI:
Currently CI runs clippy as `cargo clippy --all-targets` without the `--all` flag, causing some code to be ignored. This PR enforces the `--all` flag and resolves some warnings from the addition of the flag.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -101,7 +101,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: --all-targets -- -D warnings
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Looks like `--all` is a deprecated alias for `--workspace`: https://github.com/rust-lang/cargo/blob/df56877805e613969da837e1ede28fda3a17481d/src/doc/src/commands/cargo-check.md?plain=1#L53-L54

xJonathanLEI:
Ah never realized it's deprecated since there's no depreciation warning. Fixed now. Thanks!

PR #4615, opened 2022-11-06
add switch case by kind functions

vincascm:


A-Walrus:
LGTM, although you need to run `cargo xtask docgen` to update the auto-generated documentation to include the new commands

vincascm:
thank you for your reply. this is just a set of `MappableCommand::Static`, even if `cargo xtask docgen` executed, will not change `book/src/generated/typable-cmd.md`

helix-term/Cargo.toml
```diff
@@ -67,6 +67,9 @@ serde = { version = "1.0", features = ["derive"] }
 grep-regex = "0.1.10"
 grep-searcher = "0.1.10"
 
+# case conversion
+heck = "0.4.0"
```
archseer, line 71: Pulls in another dependency :/

the-mikedavis:
I don't think it's necessary to add commands for this. Converting case is possible to do with just regular editing primitives. The Kakoune introduction docs even use it to show off the power of selections and multiple cursors: https://kakoune.org/why-kakoune/why-kakoune.html#_multiple_selections

helix-term/Cargo.toml
vincascm: `heck` has only one dependency : [unicode-segmentation = { version = "1.2.0", optional = true }](https://github.com/withoutboats/heck/blob/dbcfc7b8db8e532d1fad44518cf73e88d5212161/Cargo.toml#L20)

vincascm:
@the-mikedavis consider any conversion between the following [7 cases](https://github.com/withoutboats/heck#cases-contained-in-this-library), if regular editing primitives is used, it is not easy to do.
but these operations can indeed be composed of regular editing primitives, probably lacking a good mapping, between key bindings and a sequence of operations.

helix-term/Cargo.toml
archseer: Yeah but heck is a new dependency that's pulled in for just a couple commands.

archseer:
If we added macro keybinds then this would be easier to map to a key. A simpler implementation that doesn't pull in heck could also just set the macro register to a sequence of these keys, then execute the macro

vincascm:
ok, thanks for your replies

peterwilli:
I made my own solution after not finding anything satisfactory... 

See demo:
[asciicast](https://asciinema.org/a/627504)

Source is here: https://github.com/peterwilli/ToCase

Issue #4625, opened 2022-11-07
Gruvbox theme has disorienting palette for selected regions


RyanGPalmer:
<img width="997" alt="Screen Shot 2022-11-06 at 8 37 22 PM" src="https://user-images.githubusercontent.com/18314366/200229543-59c46cda-7c4c-434e-b96e-39730d556882.png">

Top part of the screenshot above is the OOTB Gruvbox theme. As you can see, it's quite offensive to the eyes and makes it hard to see exactly where the selected region begins and ends at a glance. The bottom part of the screenshot is the theme after a couple tweaks to the palette.

Not sure if anyone actually prefers the way it is now, but it seems untenable to me.

Here's the modified settings from the screenshot:

```toml
"ui.selection" = { bg = "bg2" }
"ui.cursor.primary" = { bg = "fg4", fg = "bg1" }
"ui.cursor.match" = { bg = "bg3" }
```

Opened a pull request for the same here: https://github.com/helix-editor/helix/pull/4626

PR #4626, opened 2022-11-07
Fixed disorienting selection palette on Gruvbox theme

RyanGPalmer:
Details from https://github.com/helix-editor/helix/issues/4625

Screen Shot 2022-11-06 at 8 37 22 PM

Top part of the screenshot above is the OOTB Gruvbox theme. As you can see, it's quite offensive to the eyes and makes it hard to see exactly where the selected region begins and ends at a glance. The bottom part of the screenshot is the theme after a couple tweaks to the palette.

Not sure if anyone actually prefers the way it is now, but it seems untenable to me.

Here's the modified settings from the screenshot:

```toml
"ui.selection" = { bg = "bg2" }
"ui.cursor.primary" = { bg = "fg4", fg = "bg1" }
"ui.cursor.match" = { bg = "bg3" }
```

### DIFF
--- a/runtime/themes/gruvbox.toml
+++ b/runtime/themes/gruvbox.toml
@@ -52,9 +52,9 @@
 "ui.help" = { bg = "bg1", fg = "fg1" }
 "ui.text" = { fg = "fg1" }
 "ui.text.focus" = { fg = "fg1" }
-"ui.selection" = { bg = "bg3", modifiers = ["reversed"] }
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
I made a similar change in my theme (a gruvbox fork) 👍

The reversed modifier is pretty noisy. I think it might be ok when combined with the dimmed modified but otherwise it's too bright.

Issue #2811, opened 2022-06-19
Show match index and number of matches when searching


A-Walrus:
#### Describe your feature request
When searching ("/" command) it would be convenient to see how many matches there are, and where you are within them, something like this from Neovim:
![image](https://user-images.githubusercontent.com/58790821/174485881-ec4a4aba-8aae-4e9f-891a-9f19c525517a.png)
[4/6] -> 4th match out of 6.

archseer:
This was done to save on computation: finding the match index & total matches requires us to search the whole document instead of simply finding the next match. We'd have to compute these values on the first search/async and then store them internally so we don't need to recompute when using `n` and `N`.

antoyo:
> This was done to save on computation: finding the match index & total matches requires us to search the whole document instead of simply finding the next match. We'd have to compute these values on the first search/async and then store them internally so we don't need to recompute when using `n` and `N`.

Vim stop counting the matches at 99, so we could take a similar approach.

ShoyuVanilla:
I miss this feature and thinking about making a PR on my own.
How about the following implementation?

- Configurable options for this;
  - Off
  - Display next to the matching line like inlay hints
  - Display on the status line
- Do not slow down search; search goes on as before and the number of matches are displayed only after the full search have finished, in an async manner
- Cache search results while the buffer is intact
- Make a search result picker window, like the global search picker

useche:
If you are interested, I implemented this a while ago. The pull request is here: https://github.com/helix-editor/helix/pull/11700

It hasn't been reviewed again. I don't know if it is going to get merged. :(

PR #4635, opened 2022-11-07
statusline: Add search position

sigmaSd:
hx_sel

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -70,6 +70,8 @@ use grep_searcher::{sinks, BinaryDetection, SearcherBuilder};
 use ignore::{DirEntry, WalkBuilder, WalkState};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 
+const SEARCH_MAX_HIT: usize = 99;
+
 pub struct Context<'a> {
     pub register: Option<char>,
     pub count: Option<NonZeroUsize>,
@@ -1610,11 +1612,6 @@ fn search_impl(
     // it out, we need to add it back to the position of the selection.
     let mut offset = 0;
 
-    // Compare matches without taking text into account
-    // Text is different depending on if its forward or a backward search
-    // So we use start and end to tell them apart
-    let is_match_eq =
-        |m1: &regex::Match, m2: &regex::Match| m1.start() == m2.start() && m1.end() == m2.end();
     // use find_at to find the next match after the cursor, loop around the end
     // Careful, `Regex` uses `bytes` as offsets, not character indices!
     let mut mat = match direction {
@@ -1622,16 +1619,22 @@ fn search_impl(
         Direction::Backward => regex.find_iter(&contents[..start]).last(),
     };
 
-    // Find all matches in the document
-    // Then find the current match position inside all matches
-    // We then render that to the status line by setting editor.search_matches
-    let all_matches: Vec<_> = regex.find_iter(contents).collect();
+    let all_matches = if let Some(matches) = doc.search_info.all_matches.clone() {
+        matches
+    } else {
+        regex
+            .find_iter(contents)
+            .map(|c| (c.start(), c.end()))
+            .take(SEARCH_MAX_HIT)
+            .collect()
+    };
+
     if let Some(ref mat) = mat {
         let current_position = all_matches
             .iter()
-            .position(|this_m| is_match_eq(this_m, mat));
+            .position(|this_m| this_m.0 == mat.start() && this_m.1 == mat.end());
         if let Some(current_position) = current_position {
-            doc.search_position = Some(SearchPosition {
+            doc.search_info.position = Some(SearchPosition {
                 current_position: current_position + 1,
                 total_positions: all_matches.len(),
                 wrapped: false,
@@ -1644,7 +1647,7 @@ fn search_impl(
             Direction::Forward => 1,
             Direction::Backward => all_matches.len(),
         };
-        doc.search_position = Some(SearchPosition {
+        doc.search_info.position = Some(SearchPosition {
             current_position,
             total_positions: all_matches.len(),
             wrapped: true,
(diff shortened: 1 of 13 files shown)


helix-term/src/commands.rs
```diff
@@ -1610,13 +1610,47 @@ fn search_impl(
...
     // Careful, `Regex` uses `bytes` as offsets, not character indices!
     let mut mat = match direction {
         Direction::Forward => regex.find_at(contents, start),
         Direction::Backward => regex.find_iter(&contents[..start]).last(),
     };
 
+    // Find all matches in the document
+    // Then find the current match position inside all matches
+    // We then render that to the status line by setting editor.search_matches
+    let all_matches: Vec<_> = regex.find_iter(contents).collect();
```
the-mikedavis, line 1628: I'm a little reluctant about this feature in general because it requires collecting all matches in the document. `n` only needs to find the first match after the `start` point and `N` works on a subslice of the document in most cases but with this change, all invocations of `n` and `N` find all matches of the pattern across the doc

helix-view/src/editor.rs
```diff
@@ -746,6 +756,16 @@ pub enum CloseError {
...
 
+/// The current search position
+pub struct SearchPosition {
+    /// The position of the current match
+    pub current_position: usize,
+    /// The total number of matches
+    pub total_positions: usize,
+    /// True if we had wrapped
+    pub wrapped: bool,
+}
```
the-mikedavis, line 767: Search position should be stored on the Document and not the Editor: the statusline element becomes inaccurate when you switch between splits

helix-term/src/commands.rs
```diff
@@ -2206,6 +2240,7 @@ fn ensure_selections_forward(cx: &mut Context) {
 
 fn enter_insert_mode(cx: &mut Context) {
     cx.editor.mode = Mode::Insert;
+    cx.editor.search_position = None;
```
the-mikedavis, line 2243: I think insert-mode is not a good hook for clearing the search position:

* You can change the document in normal mode in a way that would affect the search result count; for example pasting or reloading the doc from disk
* This makes searching a kind of informal mode (like insert/normal/select) but it doesn't follow the normal behavior of clearing search results on Esc

Instead I think `Document::apply` would be a good place to clear this: that's when a transaction gets applied which might change the text in a document. That would also allow us to store and re-use the matches between successive calls of this function since we would be confident that if `doc.search_position.is_some()` then the document hasn't changed since the last search.

helix-term/src/commands.rs
sigmaSd: It can be cached like this https://github.com/helix-editor/helix/commit/07c5e9a9408dd556e9e23864682322a3a882a21c (I can clanup the code) between consecutive n or N usages.

What do you think? I personally find this feature very useful it feels weird when I'm searching a document and I don't know how many elements ahead there is

helix-term/src/commands.rs
sigmaSd: done

helix-view/src/editor.rs
sigmaSd: done

kirawi:
Resolves #2811

sigmaSd:
I think I need to add this https://github.com/helix-editor/helix/issues/2811#issuecomment-1170245334 but I wonder how is that implemented should I put a limit based on contents number of chars: (need to figure how to not land inside Unicode surrogate because that panics)

regex.find_iter(contents [..99999]) and then cap the max number of found items as well just to guarantee thats it's always the same

kirawi:
Maybe instead we could have it reflect `s` instead?

helix-term/src/commands.rs
the-mikedavis: Storing the matches on the doc (preferrably within SearchPosition) looks good. We should be able to avoid running any regex if we already have the set of matches, and we may want to perform the search async so that we don't block the UI when there are many matches https://github.com/helix-editor/helix/issues/2811#issuecomment-1170189784

the-mikedavis:
Limiting the number of chars found seems unnecessary. To limit to a certain number of results you can use `Iterator::take` on the iterator produced by `Regex:find_iter`

kirawi:
What if we instead added an element that showed the selection position? It could address

kirawi:
Misclick.

helix-term/src/commands.rs
sigmaSd: I merged them in one struct. I'll probably need some time to figure out how to make the search async and to use only one regex search, probably we'll need to keep a state in the search like a cursor for where we are at.

sigmaSd: @the-mikedavis can you review/test https://github.com/helix-editor/helix/commit/efdb69af9e417c7f427ede9c6888864fa1aa885a This commit only makes one regex search, if that works and I can clean it up I'll probably tackle async next

sigmaSd: I just noticed a bug when the regiester `/` changes the search needs to be also reset, I can fix that later after your next review

sigmaSd: Bug fixed https://github.com/helix-editor/helix/compare/master...sigmaSd:helix:reg?expand=1 I'll keep pushing updates to that branch

the-mikedavis:
> What if we instead added an element that showed the selection position?

Do you mean the line-number/column of the current match?

zummenix:
I would like to have this feature. Sometimes I need it really bad, so bad that I've implemented it in my fork hoping to open a PR but then noticed this :)

https://github.com/zummenix/helix/commit/e1ccd0ce56caebefbce2882deed29d0f8cb3cd0a

Differences are that in my implementation I don't keep state and use the status line to display current match and total number of matches. Works fast on sane documents, haven't checked it on big.

sigmaSd:
@zummenix feel free to open another PR, I don't think I''m going to have time to finish this

sigmaSd:
The point of keeping state is to not search the whole document

pascalkuthe:

rockboynton:
Just saw this, I have a more recent implementation that doesn't need to add a bunch of extra code:
https://github.com/helix-editor/helix/pull/12326

Issue #1084, opened 2021-11-12
Creating too many horizontal splits eventually causes no visual difference


cole-h:
### Reproduction steps

1. <kbd>Ctrl</kbd>+<kbd>w</kbd>
2. <kbd>Ctrl</kbd>+<kbd>s</kbd>
3. Repeat steps 1 and 2 until you can't tell you're creating more splits

asciinema recording:

[![asciicast](https://asciinema.org/a/2WD3lrSjhujBWZCcFwZsCARUF.svg)](https://asciinema.org/a/2WD3lrSjhujBWZCcFwZsCARUF)

### Environment

- Platform: Linux
- Helix version: v0.5.0-93-gfa0cb01

PR #4642, opened 2022-11-07
Limit the amount of horizontal splits you can do

janhrastnik:
Hello,
this PR would address #1084 by setting a limit to the number of horizontal splits you can do.
If there exists a view with a height of 1 (only the status bar is drawn) after hsplit is run, then the newly created view is closed
and we refocus the previous view.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1152,6 +1152,7 @@ fn hsplit(
     }
 
     let id = view!(cx.editor).doc;
+    let view_id = view!(cx.editor).id;
 
     if args.is_empty() {
         cx.editor.switch(id, Action::HorizontalSplit);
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1162,6 +1163,22 @@ fn hsplit(
...
 
+    // check if there are views with height equal to 1
+    // if there are, close the newly created view
+    let mut reached_limit = false;
+    cx.editor.tree.views().for_each(|v| {
+        if v.0.area.height == 1 {
+            reached_limit = true;
+        }
+    });
+    if reached_limit {
```
the-mikedavis, line 1174:
```suggestion
    if cx
        .editor
        .tree
        .views()
        .any(|(view, _focused)| view.area.height == 1)
    {
```

janhrastnik: Thanks :) . I tried to put all the code in the for_each loop at first but I got a double borrow error. I don't understand why it works with any() now.

the-mikedavis (commented):
I think we will also need this check in `split` and `hsplit_new` https://github.com/helix-editor/helix/blob/a19eee6450764ec2be0e8ad25582f33bd68a5d8c/helix-term/src/commands.rs#L4242-L4265 (for repeated splitting by `<C-w><C-s>`)

janhrastnik:
I didn't include the check in `split` in commands.rs since `vsplit` also calls that and I seperated the check into its own function. I hope that's all good.

the-mikedavis (approved):

Issue #3606, opened 2022-08-30
Bash completion ends after directory


DuckDuckWhale:
### Summary

When I press tab on a directory, the completion ends and starts the second argument. This is very annoying when navigating to find a file and I'd have to backspace and `/` and tab again after every level.

### Reproduction Steps

I tried this:

1. `hx con<tab>the<tab>`

I expected this to happen:

`hx contrib/themes` or `hx contrib/themes/`

Instead, this happened:

`hx contrib theme.toml`

### Platform

Ubuntu 22.04

### Terminal Emulator

GNOME Terminal 3.44.0 for GNOME 42

### Helix Version

helix 22.05 (8a757954)

d12bb:
> Maybe @d12bb could drop a hint or two?

Unfortunately no. I don't use Bash myself and only hacked this part together using lots of Google and Stack Overflow. Glad you've found a solution :)

PR #4648, opened 2022-11-08
bash completion: use builtin filenames option

maxhille:
Use a builtin bash option which detects filenames in completion outputs and reflects this in sensible <tab> completion behaviour.

Fixes https://github.com/helix-editor/helix/issues/3606.

Solution taken from https://stackoverflow.com/a/57385447/470509


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4649, opened 2022-11-08
[MVP] Integrated Terminal

andreytkachenko:
Minimal implementation of integrated terminal based on `alacritty_terminal` crate.
Checklist:
- [x] terminal rendering logic
- [x] minimal key handlers
- [x] async pty (used `pty_process`)
- [x] handle process terminate event
- [ ] rendering glitches (suspect tab symbol wrong render)
- [ ] close terminal and kill the process
- [ ] config section in `config.toml`
- [ ] render title
- [ ] render cursor
- [ ] mouse support
- [ ] full keyboard support
- [ ] scrolling
- [ ] clipboard support

Also nice to have later:
- split and tab layouting
- ssh tty
- text motions support

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -512,6 +512,14 @@ impl Application {
                     self.last_render = Instant::now();
                 }
             }
+
+            EditorEvent::TerminalEvent(event) => {
+                let needs_render = self.editor.handle_virtual_terminal_events(event).await;
+                if needs_render {
+                    self.render();
+                }
+            }
+
             EditorEvent::DebuggerEvent(payload) => {
                 let needs_render = self.editor.handle_debugger_message(payload).await;
                 if needs_render {
--- a/helix-term/src/ui/terminal.rs
+++ b/helix-term/src/ui/terminal.rs
@@ -1,0 +1,125 @@
+use std::cell::Cell;
+
+use helix_core::Position;
+use helix_view::{
+    graphics::{CursorKind, Rect},
+    input,
+    theme::Modifier,
+    Editor,
+};
+use tui::buffer::Buffer;
+
+use crate::compositor::{Context, EventResult};
+
+pub struct Terminal {
+    size: Cell<(u16, u16)>,
+    prev_pressed: Cell<bool>,
+}
+
+impl Terminal {
+    pub fn new() -> Self {
+        Self {
+            size: Cell::new((24, 80)),
+            prev_pressed: Cell::new(false),
+        }
+    }
+
+    pub fn render(&self, area: Rect, surface: &mut Buffer, cx: &mut Context) {
+        if self.size.get() != (area.height, area.width) && area.height > 0 && area.width > 0 {
+            self.size.set((area.height, area.width));
+
+            cx.editor
+                .terminals
+                .handle_input_event(&input::Event::Resize(self.size.get().1, self.size.get().0));
+        }
+
+        if let Some((_, term)) = cx.editor.terminals.get_active() {
+            let content = term.renderable_content();
+
+            surface.clear(area);
+
+            for cell in content.display_iter {
+                if let Some(c) = surface.get_mut(
+                    area.left() + cell.point.column.0 as u16,
+                    area.top() + cell.point.line.0 as u16,
+                ) {
+                    let style = helix_view::theme::Style::reset()
+                        .bg(cell.bg.into())
+                        .fg(cell.fg.into())
+                        .add_modifier(unsafe { Modifier::from_bits_unchecked(cell.flags.bits()) });
+
+                    let style = if let Some(col) = cell.underline_color() {
+                        style.underline_color(col.into())
+                    } else {
+                        style
+                    };
+
+                    c.reset();
+                    c.set_char(cell.c);
+                    c.set_style(style);
+                }
+            }
+        }
+    }
+
+    pub(crate) fn get_cursor(
+        &self,
+        area: Rect,
+        editor: &Editor,
+    ) -> Option<(Option<Position>, CursorKind)> {
+        editor.terminals.get_active().map(|(_, term)| {
+            let pt = term.grid().cursor.point;
+            (
+                Some(Position {
+                    row: area.y as usize + pt.line.0 as usize,
+                    col: area.x as usize + pt.column.0 as usize,
+                }),
+                term.cursor_style().shape.into(),
+            )
+        })
+    }
+
+    pub(crate) fn handle_event(
+        &self,
+        event: &input::Event,
+        context: &mut Context,
+    ) -> Option<EventResult> {
+        if context.editor.terminals.visible {
+            match event {
+                input::Event::Key(input::KeyEvent {
+                    code: input::KeyCode::Char('\\'),
+                    ..
+                }) => {
+                    if self.prev_pressed.get() {
+                        context.editor.terminals.visible = false;
+                        Some(EventResult::Consumed(None))
+                    } else {
+                        self.prev_pressed.set(true);
+
+                        if context.editor.terminals.handle_input_event(event) {
+                            Some(EventResult::Consumed(None))
+                        } else {
+                            Some(EventResult::Ignored(None))
+                        }
+                    }
+                }
+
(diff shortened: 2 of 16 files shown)


helix-view/Cargo.toml
```diff
@@ -42,6 +43,7 @@ toml = "0.5"
 log = "~0.4"
 
 which = "4.2"
+alacritty_terminal = "0.17.0"
```
archseer, line 46: Can you add this dependency only in the `helix-vte` crate, then re-export it via `pub use alacritty_terminal` there? That way it's not included several times

helix-term/src/ui/terminal.rs
```diff
@@ -0,0 +1,121 @@
...
+            NamedColor::BrightYellow => Color::LightYellow,
+            NamedColor::BrightBlue => Color::LightBlue,
+            NamedColor::BrightMagenta => Color::LightMagenta,
+            NamedColor::BrightCyan => Color::LightCyan,
+            _ => Color::Reset,
+        },
+        alacritty_terminal::ansi::Color::Spec(c) => Color::Rgb(c.r, c.g, c.b),
+        alacritty_terminal::ansi::Color::Indexed(idx) => Color::Indexed(idx),
+    }
+}
```
archseer, line 121: This should be in the `helix-view` crate, you can implement `From<alacritty_terminal::ansi::Color>` for `Color`

helix-term/src/application.rs
```diff
@@ -553,6 +559,15 @@ impl Application {
...
     }
 
+    pub fn handle_virtual_terminal_events(&mut self, event: TerminalEvent) {
+        match event {
+            TerminalEvent::TitleChange(_, _) => (),
+            TerminalEvent::Update(_) => {}
+        }
+
+        self.render();
+    }
```
archseer, line 569: Similar to `helix_dap`, can you move this to the `helix-vte` crate, then call it in the loop here (with a render after?) I'm going to extract the lsp handler into helix-lsp too

helix-view/src/handlers/vte.rs
```diff
@@ -0,0 +1,7 @@
+use crate::{terminal::TerminalEvent, Editor};
+
+impl Editor {
+    pub fn handle_virtual_terminal_events(&mut self, event: TerminalEvent) -> bool {
+        unimplemented!()
+    }
+}
```
archseer, line 7: Ah sorry yeah, put it in this file :)

helix-vte/src/pty/command.rs
```diff
@@ -0,0 +1,176 @@
...
+    pub fn new<S: AsRef<std::ffi::OsStr>>(program: S) -> Self {
+        Self {
+            inner: tokio::process::Command::new(program),
+            stdin: false,
+            stdout: false,
+            stderr: false,
+            pre_exec_set: false,
+            pre_exec: None,
+        }
+    }
```
archseer, line 21: Why the wrapper type? `Command::new` is already a builder function

andreytkachenko: this is a part of `pty_process` crate I am intended to ether move it into own crate or wait while maintainer to fix `pty_process`

sigmaSd:
Was wezterm https://github.com/wez/wezterm/tree/main/term considered ? The advantage is the number of protocols it implements It has sixel iterm2 and kitty support and even more stuff already implemented.

Nvim terminal is implementing kitty protocol progressively and helix itself supports it since its using crossterm

archseer:
Good call, it should also be a bit more lightweight since you need to bring your own `pty` provider

andreytkachenko:
> Was wezterm https://github.com/wez/wezterm/tree/main/term considered ? The advantage is the number of protocols it implements It has sixel iterm2 and kitty support and even more stuff already implemented.
> 
> Nvim terminal is implementing kitty protocol progressively and helix itself supports it since its using crossterm

Hi, yes sure, but I decided that pulling whole wezterm repo only for `term` crate is too heavy and using `alacritty_terminal` instead. Since they push the `term` crate into `crates.io` it would be nice to switch to it later.

helix-view/Cargo.toml
andreytkachenko: I really don't think `helix-vte` should know anything about `alacritty_terminal`. `helix-vte` it is about process PTY or SSH.

archseer: Hmm okay. How about splitting `helix-view/src/terminal.rs` etc into a `helix-terminal` crate? I'd like to keep the least amount of code possible in `helix-view` so it's more separated from the rest of `-view`

helix-vte/src/lib.rs
```diff
@@ -0,0 +1,180 @@
...
+    pub arguments: Option<Vec<String>>,
+    pub size: Option<(u16, u16)>,
+    pub cwd: Option<PathBuf>,
+    pub env: Option<HashMap<String, String>>,
+}
+
+impl Default for PtySpawnConfig {
+    fn default() -> Self {
+        Self {
+            command: "/usr/bin/zsh".to_string(),
```
archseer, line 53: Let's pass this in and use `editor.config.shell` https://github.com/helix-editor/helix/blob/5835d0b4601e29ed4001db0c1ad365f02d88d9e3/helix-view/src/editor.rs#L124-L125

helix-view/Cargo.toml
andreytkachenko: Only concern is - the name (`helix-terminal`) is confusing with `helix-term`

helix-vte/src/lib.rs
andreytkachenko: I think we shouldn't use that option, because by default it is just `/bin/sh` and that is good for simple cases like pipe through some program. But when we starting the terminal we usually would like to see default system shell program. For now I am setting it value from `SHELL` env variable, but later it would be better make some terminal section in main configuration file.

helix-vte/src/error.rs
```diff
@@ -0,0 +1,11 @@
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("Pty Error: {0}")]
+    PtyError(#[from] pty_process::Error),
+
+    #[error("IO Error: {0}")]
+    IoError(#[from] std::io::Error),
+
+    #[error("Termina Not Found: {0}")]
```
dgkf, line 9:
```suggestion
    #[error("Terminal Not Found: {0}")]
```

Totally benign typo - just happened to be reading through the code and spotted it.

xJonathanLEI:
FWIW [Zellij](https://github.com/zellij-org/zellij) uses `wezterm`.

archseer:
It actually also uses alacritty. https://github.com/zellij-org/zellij/search?q=alacritty https://github.com/zellij-org/zellij/search?q=wezterm

kazimir-malevich:
Is the consensus to include a built in terminal as opposed to a multiplexer? If so what needs to be done?

xprnio:
> Is the consensus to include a built in terminal as opposed to a multiplexer? If so what needs to be done?

Random question as an outsider - is there a difference to the end-user?
AFAIK if a terminal was included instead of a multiplexer, wouldn't you be able to start a multiplexer (eg. tmux) inside the terminal? And wouldn't including a multiplexer basically accomplish the same, but without requiring the user to do the multiplexer hop?

archseer:
There is no difference in our case. A built-in in terminal would automatically be a multiplexer since we'd be multiplexing it in a view split

kazimir-malevich:
There has to be way to integrate the [warp terminal](https://www.warp.dev/) with Helix. Can't we just split the windows and have keybindings to change focus?

andreytkachenko:
@kazimir-malevich I suspect it wouldn't be integrated due to license incompatibility

> Warp (https://warp.dev) is currently closed-source.

Nyabinary:
What's the status on this PR?

xJonathanLEI:
The status is `Open`.

andreytkachenko:
> What's the status on this PR?

ryanabx:
I'm unable to build this on windows, build log: https://pastebin.com/EuUa17JF

Of course, this hasn't been updated in years so I'm sure it's probably a package getting updated and wreaking havoc, or maybe this just wasn't tested on windows.

Issue #4561, opened 2022-11-01
add `]F` and friends for consistency


Itay123TheKing:
I love the bracket menus `]` and `[` for goto_next_* and goto_prev_*, but I find it annoying that only diagnostics get special treatment with `]D` and `[D` for going to the first and last diagnostic.

I've been looking around the code to see how to approach this and I may have sort of already started implementing it, so this is more of a "is anyone else working on this already?" because it feels quite obvious to me but I haven't found any open issue or PR.

Also what is a good place to ask the more experienced contributors here questions about what is the best way to implement something? The Matrix room seems to be a bit more focused on using Helix and more broadly where it's going

the-mikedavis:
I'm not sure the first/last variants are necessary: they can be emulated for the most part by `gg]f` and `ge[f` (except for when there are no functions/classes/diagnostics/etc).

It would be interesting to have a way to create keybinds that act as a transactions so that if any of the included motions "fail" (in this case: can't find a function/class/diagnostic/etc), the overall motion is discarded. That would enable emulating this behavior exactly without the need to introduce new commands and keybinds. Though I think this idea should await better scripting/configuration support

the-mikedavis:
> Also what is a good place to ask the more experienced contributors here questions about what is the best way to implement something? The Matrix room seems to be a bit more focused on using Helix and more broadly where it's going

The Matrix room can be good for development discussions - it just tends to be more usage-focused at times. @archseer has considered splitting the main room into a development room and a help/usage room - so far it hasn't been so busy that this is necessary but that can be revisited

Itay123TheKing:
@the-mikedavis 
> I'm not sure the first/last variants are necessary: they can be emulated for the most part by `gg]f` and `ge[f` (except for when there are no functions/classes/diagnostics/etc).

A lot of the keybinds are "unnecessary" but I know I've been personally missing `]F` and it's not like it makes much sense to assign any other action to this key combo — and as mentioned before we already have this with diagnostics, so I think this makes sense even if just for consistencies sake, wouldn't you agree?

David-Else:
The `space Add newline below` is the real anomaly?

Itay123TheKing:
@David-Else 
> The `space Add newline below` is the real anomaly?

Yeah, that also stands out, but it's also useful so I don't mind it. I personally also mapped `]]` and `[[` to `:bn` and `:bp` respectively,

the-mikedavis:
[...]

PR #4650, opened 2022-11-08
Add command/keybinding to jump between hunks

pascalkuthe:
~~Requires #3890~~

Adds two commands `goto_next_hunk` and `goto_prev_hunk` that allow jumping between the hunks displayed in the gutter.
I have bound these commands to ~~`]h` and `[h`~~ `]g` and `[g` by default, but I am open to suggestions for alternatives.
See it in action:

[asciicast](https://asciinema.org/a/QOkgtfoGIhfPnZeqUyjLMN0St)

Marked as draft until #3890 is merged, ~~and I add a `goto_first_hunk` and `goto_last_hunk` command (bounds to `[G` and `]G`). And some docs are missing aswell~~

~~The failing CI is unrelated (currently fails on master aswell)~~

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2992,7 +2992,7 @@ fn goto_next_change_impl(cx: &mut Context, direction: Direction) {
             let hunk_end = if hunk.after.is_empty() {
                 hunk_start + 1
             } else {
-                doc_text.line_to_char(hunk.after.end as usize) - 1
+                doc_text.line_to_char(hunk.after.end as usize)
             };
             let new_range = Range::new(hunk_start, hunk_end);
             if editor.mode == Mode::Select {


xJonathanLEI:
Wow this is amazing! Though I would suggest using `[c` and `]c` instead since those are the default bindings of the popular Vim plugin [vim-gitgutter](https://github.com/airblade/vim-gitgutter) (7.9k GitHub stars), which I have been using myself.

xJonathanLEI:
Have been trying out this PR for a few days, and I still feel like `]h` and `[h` are not very ergonomic. I have to move my right hand to press `]` and `[`, and then also use the right hand to press `h` (or move the left hand) - I have to move my hand back and forth for moving _one_ hunk. It's much slower when moving across multiple. (Maybe some chads can press `[` and `]` without moving the hand... but not me)

I feel like `[`/`]` plus something that can be pressed with the left hand without moving would be nice. The default in `vim-gitgutter` is really nice but it's already taken in Helix, unfortunately.

xJonathanLEI:
Hi @pascalkuthe, sorry for pushing on a draft PR, but would you mind rebasing this on #3890 following the latest force-push on that side? I've been testing #3890 + this PR but now it has conflicts. Thanks!

helix-vcs/src/diff.rs
```diff
@@ -0,0 +1,218 @@
...
+    /// if there is no `n`th change  `Hunk::NONE` is returned instead.
+    pub fn nth_hunk(&self, n: u32) -> Hunk {
+        match self.hunks.get(n as usize) {
+            Some(hunk) if self.inverted => hunk.invert(),
+            Some(hunk) => hunk.clone(),
+            None => Hunk::NONE,
+        }
+    }
+
+    fn next_hunk_impl<const INVERT: bool, const REVERSE: bool>(&self, line: u32) -> Option<u32> {
```
the-mikedavis, line 163: I think this would be clearer if it used a `helix_core::movement::Direction` parameter than a const generic for `REVERSE`. Similarly, can't we use `Self::is_inverted` or `self.inverted` directly instead of `INVERT`?

IIUC, the const generic is faster because the compiler knows the value at compile-time and can use that knowledge for optimization but I'm not sure this is a worthy tradeoff unless the performance improvement is noticeable

pascalkuthe: It's mostly a style choice for `REVERSE`. If you look at the actual implementation of `next_hunk_impl`, the `REVERSE` and non `REVERSE` case are basically two separate functions with a shared structure. I thought the details would be easier to grasp if they are merged but in my mind these are very much separate functions that I wouldn't merge with a bool conditions. I can imagine a couple usecases in the future that might be a bit more performance sensitive that might only want to call `next_hunk` or `prev_hunk`.

I guess I view the fact that `next_hunk` and `prev_hunk` are internally the same function mostly as an implementation detail and something I would want to expose by making it part of the API.

Once I was already using const generics I guess it didn't hurt to add one for `INVERT` aswell.
Both `INVERT` and `REVERSE` are checked as part of the binary search so I also liked the garuntee of the condition being outlined.

If you think it's really ugly I can definitely change this but I actually really like this pattern of using const generics to outline conditions out of functions.

pascalkuthe:
> Hi @pascalkuthe, sorry for pushing on a draft PR, but would you mind rebasing this on #3890 following the latest force-push on that side? I've been testing #3890 + this PR but now it has conflicts (trivial but makes an auto rebasing script choke lol). Thanks!

Apologies I forgot about this, I pushed the newest changes. Altoguh note that this PR contains all changes from #3890 right now so if you are using this PR you can just ignore #3890 and avoid any conflict that way.

pascalkuthe:
I added keybindings ( `]G` and `[G`) to jump to the first/last change and added docs

helix-vcs/src/diff.rs
the-mikedavis: I wouldn't mind it so much if it were more common (look at how [Kakoune uses a similar pattern](https://github.com/mawww/kakoune/blob/760a45556cfb8f50813ce8d1712a0a23f86dcf52/src/normal.cc#L2436-L2442) with C++ equivalents) and if we could use enums instead of bools (AFAIK this is planned for the future in the language). But this will be the only use of const generics in the codebase and it's sacrificing what can be a Forward/Backward enum for a bool which we try to push against (https://github.com/helix-editor/helix/pull/1723#discussion_r831987286, https://github.com/helix-editor/helix/pull/4134#discussion_r994061294)

helix-term/src/commands.rs
```diff
@@ -2915,6 +2920,66 @@ fn goto_prev_diag(cx: &mut Context) {
...
+            let idx = if reverse {
+                hunks.len().saturating_sub(1)
+            } else {
+                0
+            };
+            hunks.nth_hunk(idx)
+        };
+        if hunk != Hunk::NONE {
+            let pos = doc.text().line_to_char(hunk.after.start as usize);
+            goto_pos(editor, pos)
```
the-mikedavis, line 2946: Instead of selecting the point of start, I think these functions should select the whole hunk:

```rust
Selection::single(
  text.line_to_char(hunk.after.start),
  text.line_to_char(hunk.after.end),
)
```

like #4713

This lets you operate on the hunk once you jump to it (delete a hunk that's new or change some part of the hunk). Along these lines we may also want to add `mi`/`ma` textobjects so that you can select a hunk from within and operate on that.

helix-term/src/commands.rs
```diff
@@ -2915,6 +2920,66 @@ fn goto_prev_diag(cx: &mut Context) {
...
+        let hunks = diff_handle.hunks();
+        let hunk_idx = hunks.next_hunk::<REVERSE>(cursor_line as u32)?;
+        let line = hunks.nth_hunk(hunk_idx).after.start;
+        Some(line)
+    });
+
+    if let Some(next_change_line) = next_changed_line {
+        let pos = doc.text().line_to_char(next_change_line as usize);
+        goto_pos(cx.editor, pos)
+    }
```
the-mikedavis, line 2980: [`goto_ts_object_impl`](https://github.com/helix-editor/helix/blob/607c74efde40812caa6379a05fcb28f259ea2c8e/helix-term/src/commands.rs#L4404-L4443) is the 'state of the art' on these kinds of jumps:

* `cx.count()` should be taken into account so that you can jump over multiple hunks.
* setting `cx.editor.last_motion` to the motion allows you to repeat the motion with `A-.`.
* `Selection::transform` lets you operate on all cursors at once. I'm not sure how useful that would be for this motion but I don't think we should limit ourselves to just the primary cursor.
* Negotiating the head, anchor and `editor.mode` makes these extend in select mode.

book/src/keymap.md
```diff
@@ -320,6 +320,10 @@ Mappings in the style of [vim-unimpaired](https://github.com/tpope/vim-unimpaire
 | `]t`     | Go to previous test (**TS**)                 | `goto_prev_test`      |
 | `]p`     | Go to next paragraph                         | `goto_next_paragraph` |
 | `[p`     | Go to previous paragraph                     | `goto_prev_paragraph` |
+| `]g`     | Go to next change                            | `goto_next_change`    |
+| `[g`     | Go to previous change                        | `goto_prev_change`    |
+| `]G`     | Go to first change                           | `goto_first_change`   |
+| `[G`     | Go to last change                            | `goto_last_change`    |
```
the-mikedavis, line 326: I have some reservations about the need for capital versions of textobjects (like `]D`/`[D`): you could emulate their behavior with `gg]<the textobject>`/`ge[<the textobject>` or create macros for them (also see https://github.com/helix-editor/helix/issues/4561). What do you think @archseer?

pascalkuthe:
Thanks for the torough reviewe @the-mikedavis, I didn't add new motions yet so I wasn't aware of the state of the art in that regard. I am a bit busy this morning but will rebase on #3890 this afternoon and apply your comments

book/src/keymap.md
pascalkuthe: I tried to just match the existing bindings for diagnostics here to ensure the bindings were consistent. I was not aware of previous discussion in that regard. I can see that a seperate command is not really necessary here although one nice advantage of habing a seperate command for this is s that `]G` does not jump if the file is unchanged (same as with `]D` then there are no diagnostics) which can be useful. Just let me know what I should do here

helix-term/src/commands.rs
pascalkuthe: I think `mig` and `mag` would definitely be useful.

I have some reservations about making jump to hunk a select motion tough. Deletion hunks are special in that they really correspond to lines being removed BEFORE the line where the gutter is shown so they do not actually correspond to any text that is still in the document. They are shown at the first unchanged line after the removal (as bar at the top.to make that clear).

I think selecting this unchanged the line for deletions is inconsistent as it does not select any changed text. Ideally we really want a selection of width 0 at the star of the line. I don't believe that is possible tough, right?


An alternative could be to just select the first char of the unchanged line for deletions but that also feels inconsistent. 

Ideally we would habe separate motions for select changed lines and jump to hunk (the former being a selection and the latter being just a jump) but that would be yet anthor keybdinging (but potrntially a useful one) I am not sure what the best option would be . What are your thaughts here?

helix-vcs/src/diff.rs
pascalkuthe: I removed the const-generics. For `next`/`prev` I created two seperate functions and an `if` for `inverted`.
I opted for separate functions as I need to match on the direction anyway at the callsite to correctly account for `count` (and as described above), while the implementation are basically two seperate functions anyway

helix-term/src/commands.rs
pascalkuthe: Apologies I should have read into existing motions more. I updated the motion to behave correctly.

helix-term/src/commands.rs
pascalkuthe: I added a surround textobject (`mig`/`mag`) for now.
This surround object only matches the added text (so it doesn't do anything for a line where a removal is shown).

xJonathanLEI: Just giving my 2 cents...

I agree with @pascalkuthe that it feels more appropriate to not implement this as selection.

From a consistency standpoint, except _addition hunks_, you can't really accurately select what you _really_ changed anyways. So we might as well just choose the start position in all cases.

From a use case perspective, personally I do the hunk jumps when:

- I'm reviewing code - not taking action on the hunk itself
- I'm picking up my work-in-progress code from my last work session, and jumping helps me get there quickly. While I do need to take action on the code (continue working), it's unlikely I will have to action on the _whole_ hunk

In both cases, the hunk range can be easily identified through the gutters. We don't need the selection for that. And if it's unlikely that we need to act on the whole hunk, it seems unnecessary to select the thing by default.

After all, if you _really_ need to select it, you can do a `mig` afterwards.

Just my 2 cents from my experience using equivalent features on different editors (including this pending PR).

pascalkuthe: > you can't really accurately select what you really changed anyways

Note yet ;) I am planning to add high quality word diffs (similar to delta or the github diff)  to `imara-diff` in the future. Then `mig` could select the words/char that were added and `mag`  could select the entire additions... That's a bit away tough

pascalkuthe:
Git gutters have been merged (yay :)) so this PR is now unblocked. I have rebased on master

helix-term/src/commands.rs
the-mikedavis: > There is really no such thing as an outer hunk

I think this is ok, the same tends to happen for tree-sitter comment and parameter textobjects :+1:

> An alternative could be to just select the first char of the unchanged line for deletions but that also feels inconsistent.

I think this behavior would be ok. I don't think that jumping to deletion hunks would be very useful but also I think it would be surprising if the textobject totally disregarded them (skipped over them).

> Ideally we would habe separate motions for select changed lines and jump to hunk

I would be interested in exploring this idea where we have stable and consistent language for the different kinds of navigation (currently it's a bit muddled). Having extra commands that don't have keybinds is ok and then people could choose their desired behavior with just keymaps. For the default keybinds though I think it's good to be very selection-oriented: in testing this PR I found these textobjects to be very useful as ways to both jump to recently edited content (when a file is big and I don't know where I was exactly) and as a way to act on blocks that were modified/added (I didn't really think I would need this but it's actually very handy).

> In both cases, the hunk range can be easily identified through the gutters. We don't need the selection for that

Actually I disagree. Being able to see the range I would select in the gutter and then being able to easily select it is exactly the behavior I want from this textobject.

> After all, if you really need to select it, you can do a mig afterwards.

Conversely, and more consistently with other implemented textobjects, if you want the behavior where you select a point, you can bind to the textobject plus `;` (collapse_selection). Selecting as little as possible as default behavior directly fights against select-then-act IMO. Plus, using points for `]<textobject>`/`[<textobject>` makes it much much harder to extend in select mode.

helix-vcs/src/diff.rs
```diff
@@ -195,4 +195,85 @@ impl FileHunks<'_> {
...
+        let res = self
+            .hunks
+            .binary_search_by_key(&line, |hunk| hunk_range(hunk).end);
+
+        match res {
+            // Search found a hunk that ends exactly at this line (so it does not include the current line).
+            // We can usually just return that hunk, howver a specical special case for empty hunk is necessary
+            // which represents a pure removal.
+            // Removals are technically empty but are still shown as single line hunks
+            // and as such we must jump to the previus hunk (if it exists) if we are already inside the removal
```
the-mikedavis, line 239:
```suggestion
            // We can usually just return that hunk, however a special case for empty hunk is necessary
            // which represents a pure removal.
            // Removals are technically empty but are still shown as single line hunks
            // and as such we must jump to the previous hunk (if it exists) if we are already inside the removal
```

helix-term/src/commands.rs
xJonathanLEI: Yeah I'm actually indifferent - both sound reasonable to me, selection or not. I just feel like not selecting the thing is more consistent with the current diagnostics navigation `[d` and `]d`, which doesn't select the range.

the-mikedavis: Yeah it would be 👍. I would like to change that behavior though - acting on the diagnostic range I think can be useful too (#4713)

book/src/keymap.md
pascalkuthe: That sounds reasonable, I will leave them in for consistency for now and we can remove them again in the future 👍

helix-term/src/commands.rs
pascalkuthe: @the-mikedavis sorry I somehow missed the review email and won't be able to fix this today. I thaught about this some more and will change this to select the whole hunk.

I was just over thinking this. It might mean that a couple multicursor use cases don't work as expected but those cases also wouldn't work when we just jump to start (for those cases I would really want a separate motion to just select insertions but I will.leave that to the future)

pascalkuthe: changed the motion the select the hunk

pascalkuthe:
This should be ready now, PR should be squashed before merging (I didn't do it to avoid the need to re-review)

xJonathanLEI:
I like how with the latest update jumps no longer center the view. The previous behavior of always centering the view makes it hard to track movement since the whole screen always changes. Now I can tell the movement is small if nothing other than the selection changes.

pascalkuthe:
> I like how with the latest update jumps no longer center the view. The previous behavior of always centering the view makes it hard to track movement since the whole screen always changes. Now I can tell the movement is small if nothing other than the selection changes.

I mean I glad you like it but I changed nothing except making  the selection span the entire hunk instead of just the first  character so it's sort of odd that would change :D

xJonathanLEI:
Hmmm I just tested again all 5 commits do indeed have the same behavior. That's strange. Maybe it had to do with the master rebase. Anyways it's good lol. I find the centering behavior annoying.

kujohn:
Are these commands available to remap? `goto_next_hunk` command isn't available on the latest release.

gabydd:
It's ``goto_next_change`` actually that should work

kujohn:
Thank you, this works as well as `goto_prev_change`.

jonahd-g:
How exactly does diff_handle work?  I'm in a Mercurial repository with a merge conflict file open, but it says "Diff is not available in the current buffer".  Does it only work with Git with actual Git repos?

pascalkuthe:
> How exactly does diff_handle work? I'm in a Mercurial repository with a merge conflict file open, but it says "Diff is not available in the current buffer". Does it only work with Git with actual Git repos?

yes right now only git is supported, given it's dominance. Support other VCS systems could be added trough a potential future plugin system

PR #4651, opened 2022-11-08
fix: small error

0xflotus:


the-mikedavis (commented):
This line is also here: https://github.com/helix-editor/helix/blob/64359f8b484062e8c17b02fd2b2d690bb5bdfbd7/book/src/install.md?plain=1#L108-L109

Would you mind fixing that one too?

CptPotato:
Oops that was me 🙃 
You can also correct the misspelling here:

https://github.com/helix-editor/helix/blob/64359f8b484062e8c17b02fd2b2d690bb5bdfbd7/book/src/install.md?plain=1#L108-L109

the-mikedavis:
Thanks!

PR #4654, opened 2022-11-08
add `extend_prev_word_end` command

asvln:
Adds `extend_prev_word_end` to user callable commands.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -311,7 +311,7 @@ impl MappableCommand {
         goto_line_end, "Goto line end",
         goto_next_buffer, "Goto next buffer",
         goto_previous_buffer, "Goto previous buffer",
-        goto_line_end_newline, "Goto line end (newline)",
+        goto_line_end_newline, "Goto newline at line end",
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -310,8 +311,7 @@ impl MappableCommand {
         goto_line_end, "Goto line end",
         goto_next_buffer, "Goto next buffer",
         goto_previous_buffer, "Goto previous buffer",
-        // TODO: different description ?
-        goto_line_end_newline, "Goto line end",
+        goto_line_end_newline, "Goto line end (newline)",
         goto_first_nonwhitespace, "Goto first non-blank in line",
```
the-mikedavis, line 315: Looks like an unrelated change?

the-mikedavis: Might as well rename this though. I would prefer "Goto newline at line end"

asvln: Apologies, I forgot I did that when reordering. I have updated to description.

the-mikedavis (approved):

the-mikedavis:
Thank you!

Issue #4657, opened 2022-11-08
Panic with not implemented CompletionItemKind(0) in Odin project


Tefached95:
### Summary

Trying to access a function implemented in an imported package causes the editor to panic.
The project can be [found here.](https://github.com/Tefached95/odin8)
`odin version dev-2022-11:f1c24f43`
OLS at commit `1c814ac250fa96083981fecc7a5cf22e6440ed2f`
Paths to both `odin` and `ols` have been added to `PATH`.

Windows 11 Pro 22H2, build 22621.755

Language support added according to [instructions found here](https://github.com/DanielGavin/ols#helix) inside `~\AppData\Roaming\helix\languages.toml`
Here's the full backtrace:
```
PS E:\Projects\Odin\Odin8> env RUST_BACKTRACE=full hx .
thread 'main' panicked at 'not implemented: CompletionItemKind(0)', helix-term\src\ui\completion.rs:69:31
stack backtrace:
   0:     0x7ff75b65e794 - std::backtrace_rs::backtrace::dbghelp::trace
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\..\..\backtrace\src\backtrace\dbghelp.rs:98
   1:     0x7ff75b65e794 - std::backtrace_rs::backtrace::trace_unsynchronized
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\..\..\backtrace\src\backtrace\mod.rs:66
   2:     0x7ff75b65e794 - std::sys_common::backtrace::_print_fmt
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\sys_common\backtrace.rs:66
   3:     0x7ff75b65e794 - std::sys_common::backtrace::_print::impl$0::fmt
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\sys_common\backtrace.rs:45
   4:     0x7ff75ad2d80a - core::fmt::write
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\fmt\mod.rs:1194
   5:     0x7ff75b657379 - std::io::Write::write_fmt<std::sys::windows::stdio::Stderr>
                               at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\io\mod.rs:1655
[...]

PR #4658, opened 2022-11-08
`error!` on unknown `CompletionItemKind`

Tefached95:
Instead of panicking on an unknown CompletionItemKind, `error!` it instead and return `""`.

Would be a (temporary?) fix for [#4657](https://github.com/helix-editor/helix/issues/4657) as per [the-mikedavis'](https://github.com/helix-editor/helix/issues/4657#issuecomment-1307579186) instructions in the thread.


the-mikedavis (approved):

the-mikedavis:
Thank you!

PR #4659, opened 2022-11-08
[rust highlights] fix scoped attribute macro matching

theli-ua:
Without this scoped attribute macros are not matched as macros. Eg
```
  #[path::macro]
```

### DIFF
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -274,7 +274,11 @@
 (meta_item
   (identifier) @function.macro)
 (attr_item
-  (identifier) @function.macro
+  [
+    (identifier) @function.macro
+    (scoped_identifier
+      name: (identifier) @function.macro)
+  ]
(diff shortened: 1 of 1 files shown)


runtime/queries/rust/highlights.scm
```diff
@@ -273,6 +273,10 @@
 ; ---
 (meta_item
   (identifier) @function.macro)
+(attr_item
+  (scoped_identifier
+    name: (identifier) @function.macro)
+  (token_tree (identifier) @function.macro)?)
 (attr_item
   (identifier) @function.macro
   (token_tree (identifier) @function.macro)?)
```
the-mikedavis, line 282:
```suggestion
(attr_item
  [
    (identifier) @function.macro
    (scoped_identifier
      name: (identifier) @function.macro)
  ]
  (token_tree (identifier) @function.macro)?)
```

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #3014, opened 2022-07-08
File picker default highlight


mangas:
#### Describe your feature request
I think it could be very handy to highlight the current file by default on the picker so that files from the same folder are close for selection.

the-mikedavis:
To be clear, highlighting in this context is selecting the current file first and not syntax highlighting, correct?

So the behavior here would be that when the file picker's prompt is empty, the sort order should push the current file and files in its directory to the top?

mangas:
That's correct, sorting the current file and directory up when no there is no filter

pascalkuthe:
You mentinoed that #4666 was an alternative take on this issue. Does that mean that you view this feature-request as done now and this issue can be closed?

mangas:
I would rather the approach of selecting the current file by default since it's more generic and one less keybinding. I seem to remember that @the-mikedavis was not a fan of this proposal so I leave this up to you guys. Personally I would rather work on and merge https://github.com/mangas/helix/tree/select-current-file-picker which

PR #4666, opened 2022-11-08
add picker: current view dir

mangas:
Slightly different take for https://github.com/helix-editor/helix/issues/3014, looking for feedback

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -268,6 +268,7 @@ impl MappableCommand {
         append_mode, "Append after selection",
         command_mode, "Enter command mode",
         file_picker, "Open file picker",
+        file_picker_in_current_buffer_directory, "Open file picker at current buffers's directory",
         file_picker_in_current_directory, "Open file picker at current working directory",
         code_action, "Perform code action",
         buffer_picker, "Open buffer picker",
@@ -2283,6 +2284,22 @@ fn file_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlayed(picker)));
 }
 
+fn file_picker_in_current_buffer_directory(cx: &mut Context) {
+    let doc_dir = doc!(cx.editor)
+        .path()
+        .and_then(|path| path.parent().map(|path| path.to_path_buf()));
+
+    let path = match doc_dir {
+        Some(path) => path,
+        None => {
+            cx.editor.set_error("current buffer has no path or parent");
+            return;
+        }
+    };
+
+    let picker = ui::file_picker(path, &cx.editor.config());
+    cx.push_layer(Box::new(overlayed(picker)));
+}
 fn file_picker_in_current_directory(cx: &mut Context) {
     let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
     let picker = ui::file_picker(cwd, &cx.editor.config());


the-mikedavis:
Looks like https://github.com/helix-editor/helix/pull/3063

mangas:
@the-mikedavis seems the other PR is still debating the submenu and such. We could split those topics and just move ahead with a simpler option while the rest is being debated?

helix-term/src/commands.rs
```diff
@@ -2267,6 +2268,17 @@ fn file_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlayed(picker)));
 }
 
+fn file_picker_in_current_view_directory(cx: &mut Context) {
```
the-mikedavis, line 2271: We should call this `file_picker_in_current_buffer_directory` like #3063. We could call in "document" instead to match the internals but in other commands like `:buffer-close` we already call documents buffers

helix-term/src/commands.rs
```diff
@@ -2267,6 +2268,17 @@ fn file_picker(cx: &mut Context) {
...
+    let (_, doc) = current_ref!(cx.editor);
```
the-mikedavis, line 2272:
```suggestion
    let doc = doc!(cx.editor);
```

helix-term/src/commands.rs
```diff
@@ -2267,6 +2268,17 @@ fn file_picker(cx: &mut Context) {
...
+
+    let cwd = doc
```
the-mikedavis, line 2274:
```suggestion
    let document_directory = doc
```

helix-term/src/commands.rs
```diff
@@ -2267,6 +2268,17 @@ fn file_picker(cx: &mut Context) {
...
+        .path()
+        .and_then(|path| path.parent().map(|path| path.to_path_buf()))
+        .unwrap_or_else(|| std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./")));
```
the-mikedavis, line 2277: For this command, instead of falling back to the CWD, I think we should `cx.editor.set_error` a message that says that the current buffer doesn't have a directory and return early

helix-term/src/keymap/default.rs
```diff
@@ -205,6 +205,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "space" => { "Space"
             "f" => file_picker,
             "F" => file_picker_in_current_directory,
+            "v" => file_picker_in_current_view_directory,
```
the-mikedavis, line 208: I would prefer to start with no binding for now. The space keymap is already very crowded and it seems like the file submap could use some more discussion

mangas: sounds reasonable

mangas:
I think i've addressed all the comments, please let me know if anything else is missing

helix-term/src/commands.rs
```diff
@@ -266,6 +266,7 @@ impl MappableCommand {
         append_mode, "Append after selection",
         command_mode, "Enter command mode",
         file_picker, "Open file picker",
+        file_picker_in_current_buffer_directory, "Open file picker at current document's directory",
```
the-mikedavis, line 269:
```suggestion
        file_picker_in_current_buffer_directory, "Open file picker at current buffer's directory",
```

helix-term/src/commands.rs
```diff
@@ -2267,6 +2268,24 @@ fn file_picker(cx: &mut Context) {
...
+    let doc = doc!(cx.editor);
+
+    let doc_dir = match doc
+        .path()
+        .and_then(|path| path.parent().map(|path| path.to_path_buf()))
+    {
+        Some(path) => path,
+        None => {
+            cx.editor
+                .set_error("current document has no path or parent");
```
the-mikedavis, line 2281:
```suggestion
                .set_error("current buffer has no path or parent");
```

the-mikedavis (approved):

DenisKolodin:
I've checked the changes - works fine :+1: 
Also sent a short refactoring for that here: https://github.com/mangas/helix/pull/1 up to you @mangas  :wave:

mangas:
> I've checked the changes - works fine +1 Also sent a short refactoring for that here: [mangas#1](https://github.com/mangas/helix/pull/1) up to you @mangas wave

Hey :wave: thanks for the PR, I used most of the suggested change only replaced the if else with a match :)

Boscop:
When I do `hx foo/bar/baz.rs` and then inside helix I `:open asd/sdf/qwert.rs`, and then open the file picker, it's still based on `foo/bar/baz.rs`'s current dir, not on the file I was in when I summoned the file picker, is this intended?

Ideally it should have at least an option (if not the default) to use the current file's dir for the file picker.

Boscop:
@pascalkuthe Thanks. But when I do the action to open the file picker in the current buffer's directory, literally nothing happens. I tried on Win10 and in Ubuntu inside WSL2, same behavior. Any idea why?

Btw, relative to which dir should the normal file picker (`space f`) operate?
Also, `space f` and `space F` seem to do the same thing (they both open relative to the working dir that the hx process was started with, NOT relative to the path of the file that was given to `hx` when it was started).
If I open `hx /foo/bar` from being inside `/asd/qwer`, shouldn't

mlsvrts:
If anyone is new/dumb at helix like me, you can use the 'buffer directory file picker' by adding a keymap to you `config.toml`. For instance, to make `space + e` open the picker, add:

```toml
[keys.normal.space]
e = "file_picker_in_current_buffer_directory"
```

PR #4679, opened 2022-11-09
Enable elixir-ls for heex language

philipgiuliani:
Elixir-LS is also supported for `.heex`-files.
Not sure if we should also add `config = { elixirLS.dialyzerEnabled = false }` like we have in the `elixir` configuration.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -50,7 +50,7 @@
 | hare | ✓ |  |  |  |
 | haskell | ✓ |  |  | `haskell-language-server-wrapper` |
 | hcl | ✓ |  | ✓ | `terraform-ls` |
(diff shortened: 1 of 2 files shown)


archseer:
\cc @connorlay

connorlay:
I'm fine with us shipping this, but I'd like to note that HEEx support in the latest Elixir LS is still missing a lot of features (not supporting local and imported function components for hover, find references, etc.) This works well in `~H` sigils, but embedded templates need some work in the Elixir LS side to provide a complete experience.

the-mikedavis:
To fix the CI you need to run `cargo xtask docgen` and commit the changes

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #4624, opened 2022-11-07
Feature request: selected character count


xJonathanLEI:
When dealing with hashes (in hex) and big decimal numbers, I frequently find myself selecting the whole thing in vim just to check if the length is correct, thanks to the vim feature that shows how many chars you've selected.

Currently Helix is only able to show the number of _selections_. It would be nice to also have the option to display the number of characters.

xJonathanLEI:
My current workaround is to put my cursor right _before_ the text I'm interested in, then press the desired length followed by `l` to see if the cursor lands on the last character. It kind of works for my case but not when you want to _measure_ the length of some text.

Siiiimon:
Hi all!
I'd like to tackle this issue if that's fine with everyone 😄

wes-adams:
@Siiiimon - sorry
this is  my first ever attempt at contributing. i didn't realize that we should claim work before starting.

wes-adams:
the linter pointed out my poor coding style

pushed 2nd attempt

Siiiimon:
no worries!

wes-adams:
@Siiiimon - thanks for the feedback - i'll definitely be implementing your suggestions

i am stuck on one thing though:
how do i link my new `config.toml` setting to the actual logic in source?
```
config.toml
. . . 
[editor.statusline]
left = ["mode", "spinner"]
center = ["file-name", "primary-selection-length"]
right = ["diagnostics", "selections", "position", "file-encoding", "file-line-ending", "file-type"]
. . . 
```
```
> hx
Bad config: unknown variant `primary-selection-length`, expected one of `mode`, `spinner`, `file-name`, `file-encoding`, `file-line-ending`, `file-type`, `diagnostics`, `selections`, `position`, `separator`, `position-percentage`, `total-line-numbers`, `spacer` for key `editor.statusline.center` at line 16 column 1
Press <ENTER> to continue with default config
```
![Screenshot from 2022-11-10 09-41-57](https://user-images.githubusercontent.com/7883354/201121197-72220ccf-4beb-4d77-8331-76e1a7e693f2.png)

PR #4682, opened 2022-11-09
Statusline indicator to show number of selected chars

wes-adams:
Feature request: #4624

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -142,6 +142,9 @@ where
         helix_view::editor::StatusLineElement::FileType => render_file_type,
         helix_view::editor::StatusLineElement::Diagnostics => render_diagnostics,
         helix_view::editor::StatusLineElement::Selections => render_selections,
+        helix_view::editor::StatusLineElement::PrimarySelectionLength => {
+            render_primary_selection_length
+        }
         helix_view::editor::StatusLineElement::Position => render_position,
         helix_view::editor::StatusLineElement::PositionPercentage => render_position_percentage,
         helix_view::editor::StatusLineElement::TotalLineNumbers => render_total_line_numbers,
@@ -250,6 +253,18 @@ where
     write(
         context,
         format!(" {} sel{} ", count, if count == 1 { "" } else { "s" }),
+        None,
+    );
+}
+
+fn render_primary_selection_length<F>(context: &mut RenderContext, write: F)
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let tot_sel = context.doc.selection(context.view.id).primary().len();
+    write(
+        context,
+        format!(" {} char{} ", tot_sel, if tot_sel == 1 { "" } else { "s" }),
         None,
     );
 }
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -328,6 +328,9 @@ pub enum StatusLineElement {
     /// The number of selections (cursors)
     Selections,
 
+    /// The number of characters currently in primary selection
+    PrimarySelectionLength,
+
     /// The cursor position
     Position,
 
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -103,6 +103,7 @@ The following statusline elements can be configured:
 | `file-type` | The type of the opened file |
 | `diagnostics` | The number of warnings and/or errors |
 | `selections` | The number of active selections |
+| `primary-selection-length` | The number of characters currently in primary selection |
 | `position` | The cursor position |
 | `position-percentage` | The cursor position as a percentage of the total number of lines |
 | `separator` | The string defined in `editor.statusline.separator` (defaults to `"│"`) |


kirawi:
I think it would be better for this to be its own separate statusline element in https://github.com/helix-editor/helix/blob/master/helix-term/src/ui/statusline.rs.

helix-term/src/ui/statusline.rs
```diff
@@ -142,6 +142,7 @@ where
         helix_view::editor::StatusLineElement::FileType => render_file_type,
         helix_view::editor::StatusLineElement::Diagnostics => render_diagnostics,
         helix_view::editor::StatusLineElement::Selections => render_selections,
+        helix_view::editor::StatusLineElement::TotalCharsSelected => render_total_chars_selected,
```
the-mikedavis, line 145: The lingo is a little tricky for this feature: this statusline element will show the total number of characters selected _in the primary selection_. Could we call this `PrimarySelectionLength` or something along those lines so that we're specific about that?

helix-term/src/ui/statusline.rs
```diff
@@ -254,6 +255,21 @@ where
...
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let tot_sel = context
+        .doc
+        .selections()
+        .get(&context.view.id)
+        .unwrap()
+        .ranges()[0]
+        .len();
```
the-mikedavis, line 268:
```suggestion
    let length = context.doc.selection(context.view.id).primary().len();
```

helix-view/src/editor.rs
```diff
@@ -276,7 +276,13 @@ impl Default for StatusLineConfig {
...
             left: vec![E::Mode, E::Spinner, E::FileName],
             center: vec![],
-            right: vec![E::Diagnostics, E::Selections, E::Position, E::FileEncoding],
+            right: vec![
+                E::Diagnostics,
+                E::Selections,
+                E::TotalCharsSelected,
+                E::Position,
+                E::FileEncoding,
+            ],
```
the-mikedavis, line 285: I think this should be disabled by default

the-mikedavis (commented):
This will also need to be added to the documentation here: https://github.com/helix-editor/helix/blob/758bace2210e83dacadd755c3f97bb62ad259a73/book/src/configuration.md?plain=1#L93-L109

helix-view/src/editor.rs
wes-adams: does this vector contain the statusline elements that are active by default?
if so, do i need to also create a means to toggle this element?

kirawi: You can undo this particular change. The statusline is configurable: https://docs.helix-editor.com/configuration.html#editorstatusline-section

helix-term/src/ui/statusline.rs
wes-adams: thanks for the code improvement

helix-term/src/ui/statusline.rs
```diff
@@ -254,6 +257,14 @@ where
     );
 }
 
+fn render_primary_selection_length<F>(context: &mut RenderContext, write: F)
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let tot_sel = context.doc.selection(context.view.id).primary().len();
```
Siiiimon, line 264: I know this PR is only concerned with the primary selection, but IMO it might be worthwhile to sum up every range's lengths to also support multi-selection counts. This could either be a separate status line element all together or it could switch to showing the primary and total character count when a multiline selection is active.

image

That's of course just a side note and this change would 100% solve the original issue

helix-term/src/ui/statusline.rs
```diff
@@ -254,6 +257,14 @@ where
...
+    write(context, format!(" {} ", tot_sel), None);
```
Siiiimon, line 265: adding the word 'char' to give some more context and handling the plural form in a similar manner as the selection counter would be a nice addition I think.

```suggestion
    write(context, format!(" {} char{} ", tot_sel, if tot_sel == 1 { "" } else { "s" }), None);
```

Siiiimon (commented):
Because of the way range lengths are calculated, selecting a line actually shows one too many characters currently.

image
here i've just selected the line by pressing 'x' and the character count shows 5 instead of 4.

the-mikedavis:
> Because of the way range lengths are calculated, selecting a line actually shows one too many characters currently.

The count is correct: it's counting the newline character.

wes-adams:
i've tested the change, and it looks good

`primary-selection-length` is off by default, and configurable via `config.toml`. this has also been tested and validated.

image

image

helix-term/src/ui/statusline.rs
wes-adams: i'll leave this effort for another time

Siiiimon: True that sounds very reasonable 👌

Siiiimon:
> > Because of the way range lengths are calculated, selecting a line actually shows one too many characters currently.
> 
> The count is correct: it's counting the newline character.

ah you're right, my mistake.

helix-term/src/ui/statusline.rs
```diff
@@ -254,6 +257,18 @@ where
...
 }
 
+fn render_primary_selection_length<F>(context: &mut RenderContext, write: F)
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let tot_sel = context.doc.selection(context.view.id).primary().len();
+    write(
+        context,
+        format!(" {} char{}", tot_sel, if tot_sel == 1 { "" } else { "s" }),
```
the-mikedavis, line 267:
```suggestion
        format!(" {} char{} ", tot_sel, if tot_sel == 1 { "" } else { "s" }),
```

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4697, opened 2022-11-10
feat: add sc to filetypes for Scala

ckipp01:
Files ending in `.sc` are known as "Scala Scripts". The scalac compiler
knows how to handle these, popular tools like
[Ammonite](https://github.com/com-lihaoyi/Ammonite) and
[scala-cli](https://github.com/Virtuslab/scala-cli) use them, and even
some Scala build tools like [mill](https://github.com/com-lihaoyi/mill)
use them as their build files (`build.sc`). This change just makes sure
that they are treated as Scala files so Metals kicks in.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #4698, opened 2022-11-10
sort fuzzy matches with equal score by length in picker

### LINKED ISSUE #4688: Symbol picker displays confusing sort order on filter
Screen Shot 2022-11-09 at 5 31 32 PM
Fuzzy match scoring can definitely be a matter of taste, but in this case I would expect the exact match "TimeEntry" to be ranked higher than all the prefix matches.

pascalkuthe:
Fixes #4688

Currently, the fuzzy matching in the picker is only sorted by the fuzzy match score.
An unstable sort is used so the order of elements with equal score is random (still deterministic but may vary between `rustc`/`std` versions).
`fzf` and `skim` offer the `--tiebreak` flag to specify a criterion to use in that case.
Available options are:
* (byte) length
* (byte) position of the first matched char
* (byte) position of the last matched char (causes matching to be performed in reverse)
* original index in the list supplied to the fuzzy matcher

Both `fzf` and `skim` allow an arbitrary list of these tiebreakers to be specified.
`fzf` uses the `length` tiebreaker by default (skim does not).
I think this is a good default because the user can always filter for longer matches by adding more characters.
The other options seem more situational to me and I couldn't think of a use case for them in helix (except `index` but here a stable sort would make more sense).

This PR changes the picker to sort items by their **char length** if their score matches (to ensure consistent behavior with non ASCII chars).
I did not implement this behavior for auto completions, because those use the stable sort order to use the order supplied by the `lsp` (especially for preselected items).

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -11,9 +11,8 @@ use tui::{
 use fuzzy_matcher::skim::SkimMatcherV2 as Matcher;
 use tui::widgets::Widget;
 
-use std::time::Instant;
+use std::{cmp::Ordering, time::Instant};
 use std::{
-    cmp::Reverse,
     collections::HashMap,
     io::Read,
     path::{Path, PathBuf},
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -309,13 +308,35 @@ impl<T: Item + 'static> Component for FilePicker<T> {
...
+            .then_with(|| self.len.cmp(&other.len))
+    }
+}
+
 pub struct Picker<T: Item> {
     options: Vec<T>,
     editor_data: T::Data,
     // filter: String,
     matcher: Box<Matcher>,
     /// (index, score)
```
the-mikedavis, line 338: This comment can be removed now

pascalkuthe: good catch, thanks!

the-mikedavis (approved):
I've been testing this for a while and I like the new behavior 👍

It works better in languages that have namespaces that nest under directories (like in Elixir or Ruby) and in languages where the convention is to prefix a module name with the namespace name (like in Erlang).

I think that the other tie-breaking options are not necessary

Issue #4702, opened 2022-11-10
Outdated typescript grammar


gabydd:
The current typescript grammar is missing some fixes that are available on the latest grammar. Specifically template literal types had some issues being parsed that messed with highlighting that was fixed in: [https://github.com/tree-sitter/tree-sitter-typescript/pull/188](https://github.com/tree-sitter/tree-sitter-typescript/pull/188).

I will create a pr for this (already have it fixed in a local branch) but I was wondering what are the guidelines for upgrading grammar? Should I update to the latest master or only to the commit that fixes the template literal type highlighting problem I had?

kirawi:
You can update to the latest stable commit, which seems to be 082da44a5263599186dadafd2c974c19f3a73d28

gabydd:
Thanks @kirawi this issue is resolved with #4703

PR #4703, opened 2022-11-10
Update typescript grammar and queries

gabydd:
fixes: #4702
updates grammars for typescript, and tsx
as well as updates queries for template literal types

### DIFF
--- a/runtime/queries/typescript/highlights.scm
+++ b/runtime/queries/typescript/highlights.scm
@@ -34,7 +34,7 @@
   "implements"
   "keyof"
   "namespace"
-  "abstract"
+  "override"
 ] @keyword
 
 [
--- a/runtime/queries/typescript/highlights.scm
+++ b/runtime/queries/typescript/highlights.scm
@@ -34,6 +34,7 @@
   "implements"
   "keyof"
   "namespace"
+  "abstract"
 ] @keyword
 
 [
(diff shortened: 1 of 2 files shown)


runtime/queries/typescript/highlights.scm
```diff
@@ -62,3 +62,15 @@
 
 ((identifier) @type
  (#match? @type "^[A-Z]"))
+
```
archseer, line 65: `"override"` was also added to the list of keywords https://github.com/tree-sitter/tree-sitter-typescript/compare/3e897ea5925f037cfae2e551f8e6b12eec2a201a..6aac031ad88dd6317f02ac0bb27d099a553a7d8c#diff-d485a982e458ef8da2cc203585065b7542665cb80b78d230b1e8f77ea25825d4R33

gabydd: just pushed a fix, thanks Blaz!

runtime/queries/typescript/highlights.scm
```diff
@@ -34,6 +34,7 @@
   "implements"
   "keyof"
   "namespace"
+  "abstract"
```
archseer, line 37:
```suggestion
  "override"
```

gabydd: Lol sorry about that

Issue #4647, opened 2022-11-08
Add `]n` and `[n` to jump between conflict markers


ocharles:
`vim-unimpaired` adds `]n` and `[n` to [navigate SCM conflict markers](https://github.com/tpope/vim-unimpaired/blob/master/doc/unimpaired.txt#L55). It'd be great to have this navigation feature in Helix, too!

wes-adams:
i'm gonna take a stab at this one

ocharles:
I look forward to seeing what you come up with! I don't know any Rust but I'd love to see how this looks, then maybe I can contribute one day :)

wes-adams:
brace yourself....


https://user-images.githubusercontent.com/7883354/201238036-29c67838-70f7-46f8-a4ee-dea942879ff8.mp4

wes-adams:
the pull request is very much in draft status. it needs to be cleaned up. 

after more trial and error and failing than i'd like to admit, it finally worked, so i pushed what i had.

i'll get it cleaned up, and then get feedback.

ocharles:
Great work!

wes-adams:
i think this issue might be dead. seems that this feature is provided in a separate PR: https://github.com/helix-editor/helix/pull/4650

@ocharles - maybe pull #4650 and verify that it works?

fwiw my work is here: https://github.com/helix-editor/helix/pull/4708

ocharles:
@wes-adams I don't think #4650 is a satisfactory fix for me. I use three-way conflict markers, and ]g doesn't seem to take me to each individual section, whereas your PR does. I'll try and put an asciinema together, but I don't think it's that #4650 does anything wrong, it's just doing something different.

pascalkuthe:
@ocharles that is already covered by searching with `/ <<<<<<`  as mentioned in #4708. In fact that is more ore less exactly what that PR was doing. If it's something you need frequently you can already bind that search to a key. 
I think this feature is a bit too specific for core and already covered by other feature. In fact they cover it better because they support jumping to both the next change with #4708 or the next section with `/ <<<<<<` depending on your needs (I use both often)

PR #4708, opened 2022-11-11
Add new command to cycle through version control conflict markers

wes-adams:
Issue: #4647


archseer:
Does this need to be a new keybind? We could structure it as a macro that does something like `/<<<<<<<`

wes-adams:
> Does this need to be a new keybind? We could structure it as a macro that does something like `/<<<<<<<`

good idea. i'll explore that.

pascalkuthe:
With  #4650 we get this sort of for free aswell (at least for git). Git refuses to perform a merge/rebase if the work tree is not clean so conflict markers will be the only change that shows up against HEAD

wes-adams:
it sounds like this merge request might be duplicating work, but i wanted to see it through regardless to learn more about how the project works.

i've tested this feature. it seems to work well. 

https://asciinema.org/a/GXSmICT22D1OOY7VoXNQCsvAt

pascalkuthe:
closing this as stale. Both regex search and `]g` can cover this usecase so I don't see us aadding a dedicated feature.

Thank you for contributing!

Issue #39, opened 2021-06-02
Gui


Kethku:
I noticed that your website references creating a gui with skia/skulpin. That sounds very similar to my architecture for neovide which is a gui for neovim.

A not broadly stated goal for neovide is to eventually expose the graphical parts as a swappable front end for other text editors such as what helix looks like. I was curious if you are interested in such a collaboration. Getting a gui right is hard work (as I've found out), I think collaborating would be great for both use cases!

Are you planning on exposing a gui protocol ala neovim's? If so what would that look like/what features do you think would be useful for helix that are unique to helix? The editor looks very cool btw!

archseer:
Hey! The skulpin mention was specifically after I looked at neovide, it seemed like a good fit. @norcalli wanted to experiment on a new frontend as well.

I'm a bit apprehensive on exposing a GUI protocol, they're either too vague and the client needs to re-implement a lot of the handling, or they're very specific and every client ends up the same. I don't have a solid opinion yet so we'll see. Right now the idea is that helix-view would provide most of the logic, and a frontend would just be another crate wrapping everything with a UI -- that needs some work right now though because `commands` are currently sitting directly in `helix-term`.

Kethku:
That makes good sense. One thing thats been on my list for a while is to split neovide into the neovim editor, and a front end renderer so that other apps could build their own rendering. Then you could integrate it as a crate without having to do all the painful font rendering and such. Let me know if something like that would be interesting for your use cases or not. In any case, if you end up going the skia route, I'm happy to help out if you run into problems. Awesome project

Kethku:
The more I think about it, the more I am interested in trying out something like this. You mentioned that commands are currently located in the helix-term. Are you suggesting that they would need to be moved into helix-view in order for a gui crate to consume and produce them? Are you interested in a PR which attempts that kind of refactoring?

archseer:
So there's a few sets of changes necessary:

- `helix-view` needs to be refactored to have it's own generic `key`/`style`/`color` structs. https://github.com/helix-editor/helix/pull/345 is a start but it depends on crossterm. Using generic structs would let us drop the crossterm dependency from `helix-view`.

- `helix-term` commands need to be moved into `helix-view`. The problem there is that [the context depends on the compositor written for the TUI](https://github.com/helix-editor/helix/blob/fd1ae35051b57e689f6e6ef7e03c552a78f3f33a/helix-term/src/commands.rs#L52), and some commands directly manipulate it by pushing in new components.
  
[...]

PR #4709, opened 2022-11-11
Support macro-style keybindings

### LINKED ISSUE #1383: Support for macro-style keybinding
Allow binding keys to sequences of other keys, like how Vim handles keybindings. I don't propose for this to *replace* the current system of binding keys to commands, but rather add this as a keybinding option, and even allow it to be mixed with command keybindings.

In the following examples, I use a `@` prefix to distinguish keypress sequences from command names, and though it is a valid possibility, note that I am not asserting that this is the syntax we should use.

```toml
# Binds D to delete the line
"D" = "@xd"
```
This would also allow for more versatile bindings. For example, zyrafal said [on Matrix](https://matrix.to/#/!zMuVRxoqjyxyjSEBXc:matrix.org/$UWsS0ce6_4U_6yvKDVcABYmgpyNxZ2pvL6L6tpjrtcI?via=matrix.org&via=kack.it&via=mtx.liftm.de):
> Is there a way to bind a key to switching to a specific register?
> For example, if i want to bind Ctrl-D to switch to register d

With macro-style keybinding support, this could be achieved like this:
```toml
"C-d" = ["select_register", "@d"]
```
Some more examples of this versatility:
```toml
# Binds M to select inside parentheses
"M" = ["select_textobjext_inner", "@("]

# Binds Alt-b to pipe the file through xxd external program
"A-b" = ["select_all", "shell_pipe", "@xxd<ret>"]
```

I can implement this once #1253 is merged. What are you all's thoughts?

the-mikedavis:
Macro key-sequences can be used to implement custom keybinds for commands that require some sort of input like `t` or `mi`:

```toml
[keys.normal]
D = "@t<ret>d"
W = "@miw"
```

To implement them, we parse a sequence of KeyEvents the same way that we parse macro input, then execute them by setting up a callback which passes the KeyEvents through the compositor.

Closes #1383

### DIFF
--- a/book/src/remapping.md
+++ b/book/src/remapping.md
@@ -75,5 +75,20 @@ Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes
 
 Keys can be disabled by binding them to the `no_op` command.
 
-A list of commands is available in the [Keymap](https://docs.helix-editor.com/keymap.html) documentation
- and in the source code at [`helix-term/src/commands.rs`](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs) at the invocation of `static_commands!` macro and the `TypableCommandList`.
+## Commands
+
+There are three kinds of commands that can be used in keymaps:
+
+* Static commands: commands like `move_char_right` which are usually bound to
+  keys and used for movement and editing. A list of static commands is
+  available in the [Keymap](./keymap.html) documentation and in the source code
+  in [`helix-term/src/commands.rs`](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs)
+  at the invocation of `static_commands!` macro and the `TypableCommandList`.
+* Typable commands: commands that can be executed from command mode (`:`), for
+  example `:write!`. See the [Commands](./commands.html) documentation for a
+  list of available typeable commands.
+* Macros: sequences of keys that are executed in order. These keybindings
+  start with `@` and then list any number of keys to be executed. For example
+  `@miw` can be used to select the surrounding word. For now, macro keybindings
+  are not allowed in keybinding sequences due to limitations in the way that
+  command sequences are executed.
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -176,9 +176,16 @@ where
 
 use helix_view::{align_view, Align};
 
-/// A MappableCommand is either a static command like "jump_view_up" or a Typable command like
-/// :format. It causes a side-effect on the state (usually by creating and applying a transaction).
-/// Both of these types of commands can be mapped with keybindings in the config.toml.
+/// MappableCommands are commands that can be bound to keys, executable in
+/// normal, insert or select mode.
+///
+/// There are three kinds:
+///
+/// * Static: commands usually bound to keys and used for editing, movement,
+///   etc., for example `move_char_left`.
+/// * Typable: commands executable from command mode, prefixed with a `:`,
+///   for example `:write!`.
+/// * Macro: a sequence of keys to execute, for example `@miw`.
 #[derive(Clone)]
 pub enum MappableCommand {
     Typable {
@@ -190,6 +197,10 @@ pub enum MappableCommand {
         name: &'static str,
         fun: fn(cx: &mut Context),
         doc: &'static str,
+    },
+    Macro {
+        name: String,
+        keys: Vec<KeyEvent>,
     },
 }
 
@@ -227,6 +238,23 @@ impl MappableCommand {
                 }
             }
             Self::Static { fun, .. } => (fun)(cx),
+            Self::Macro { keys, .. } => {
+                // Protect against recursive macros.
+                if cx.editor.macro_replaying.contains(&'@') {
+                    cx.editor.set_error(
+                        "Cannot execute macro because the [@] register is already playing a macro",
+                    );
+                    return;
+                }
+                cx.editor.macro_replaying.push('@');
+                let keys = keys.clone();
+                cx.callback.push(Box::new(move |compositor, cx| {
+                    for key in keys.into_iter() {
+                        compositor.handle_event(&compositor::Event::Key(key), cx);
+                    }
+                    cx.editor.macro_replaying.pop();
+                }));
+            }
         }
     }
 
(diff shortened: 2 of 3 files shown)


the-mikedavis:
Note that this currently has the drawbacks described here: https://github.com/helix-editor/helix/issues/1383#issuecomment-1100377552

Omnikar:
The issue of static commands being executed immediately while macros are executed by the compositor afterwards doesn't seem to be resolved?
For example, given
```toml
"D" = ["find_till_char", "@<ret>", "delete_selection"]
```
the expected behaviour upon pressing `D` would be deleting to the end of the line (specifically selecting to the end of the line and then deleting), but the actual behaviour is deleting the current selection, then selecting to the end of the line. This is because `delete_selection` runs before the `<ret>` key is processed.

the-mikedavis:
Ah nice catch, thanks for the good example! That's a tricky case. It's partially related to https://github.com/helix-editor/helix/issues/4013.

We may want to support macro keybinds initially by disallowing them in sequences: supporting macro keybinds is straightforward if we ignore sequences.

I'll push a change to fix that case (and #4013) but I feel like it might be a bit hacky: I'm adding a fake Event for the compositor to handle so that it can handle sequences that use callbacks in the correct order.

archseer:
There's also https://github.com/helix-editor/helix/pull/4244

Omnikar:
So the expected behaviour is happening now, good 👍. However, one small bug I've found: take the `D` command example. When it's executed, we get the desired behaviour of deleting to the end of the line, but a phantom `D` is rendered in the bottom right corner until the next keypress.

the-mikedavis:
> There's also #4244

It looks like #4244 has a similar concept for pending commands. It fixes the problem for on-next-key callbacks but not the compositor callbacks used by macros (`cx.callback`).

> a phantom `D` is rendered in the bottom right corner until the next keypress

Whoops, forgot to clear the pseudo-pending 😅. Should be fixed in the latest

thomkeh:
Maybe this isn't easily doable but wouldn't it be nicer if instead of

```toml
[keys.normal]
D = "@t<ret>d"
W = "@miw"
```
you could write
```toml
[keys.normal]
D = ["@find_next_char <ret>", "delete_selection"]
W = "@select_textobject_inner w"
```
(or with some symbol other than `@`) because that way there doesn't need to be an existing keymapping for `find_next_char` and `select_textobject_inner`? (And also the resulting config would be a bit more robust to other changes in your keymap.)

the-mikedavis:
That style is also possible with this branch:

```toml
[keys.normal]
D = ["find_till_char", "@<ret>", "delete_selection"]
W = ["select_textobject_inner", "@w"]
```

thomkeh:
Do you still want to get this merged? I saw you marked it as draft.

the-mikedavis:
There's a regression for long lists of regular commands where we don't re-render after executing the commands. The changes around that are a bit hacky anyway so I need to think about how to do this nicely (and without regressions)

the-mikedavis:
For the proper fix for this branch, #4244 and https://github.com/helix-editor/helix/pull/5294, I think it would be ideal to move the commands into the helix-view crate so that they can be added to the [`Event`](https://github.com/helix-editor/helix/blob/2412f7c8374d2f6a8bfd66ecf424153a37c814bf/helix-view/src/input.rs#L10) type that the compositor handles. Then we could queue up commands for execution in the same way we currently queue up keys with regular macros: https://github.com/helix-editor/helix/blob/2412f7c8374d2f6a8bfd66ecf424153a37c814bf/helix-term/src/commands.rs#L5197-L5202

(Moving commands into helix-view is also a goal of the GUI effort (#39).)

the-mikedavis:
@gibbz00 have you already started working on that? I think it will be a very large change that should be split across multiple PRs. I will write up an issue on that change with some implementation notes

nrabulinski:
I also wanted to ask which parts and how much of it have you already refactored since (as mentioned in my hotfix PR) I'd definitely like to also contribute to the long-term solution, but I wouldn't want us stepping on each other's toes 😄

godalming123:
> There's a regression for long lists of regular commands where we don't re-render after executing the commands. The changes around that are a bit hacky anyway so I need to think about how to do this nicely (and without regressions)

@the-mikedavis is this the only known problem with this PR? From what I can see, everything else has been resolved.

the-mikedavis:
Handling sequences of commands correctly will take a very large refactor as mentioned above (#5555) so I remade this branch with just the basics, so it basically mirrors `q`/`Q`. It's suitable for macros like `@miw`, entering things into prompts and using registers and counts.

I removed the ability to use macros in sequences of commands though. Without that, this is a small and IMO uncontroversial change but the functionality is a little limited. I think it's a reasonable increment since fixing sequences properly will take a lot of effort.

pascalkuthe (approved):

book/src/remapping.md
```diff
@@ -75,5 +75,20 @@ Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes
 
 Keys can be disabled by binding them to the `no_op` command.
 
-A list of commands is available in the [Keymap](https://docs.helix-editor.com/keymap.html) documentation
- and in the source code at [`helix-term/src/commands.rs`](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs) at the invocation of `static_commands!` macro and the `TypableCommandList`.
+## Commands
+
+There are there kinds of commands that can be used in keymaps:
```
Regaez, line 80: Noticed a small typo:

```suggestion
There are three kinds of commands that can be used in keymaps:
```

archseer (approved):

llakala:
> Ah nice catch, thanks for the good example! That's a tricky case. It's partially related to #4013.
> 
> We may want to support macro keybinds initially by disallowing them in sequences: supporting macro keybinds is straightforward if we ignore sequences.
> 
> I'll push a change to fix that case (and #4013) but I feel like it might be a bit hacky: I'm adding a fake Event for the compositor to handle so that it can handle sequences that use callbacks in the correct order.

Running into this now. Any chance of a follow-up PR?

KiaraGrouwstra:
could we get a new release for this feature? 🙏

the-mikedavis:
yeah the hope is to cut a new release either later this month or in early January as time allows

ThinkChaos:
@llakala I made an issue for tracking: #12420

Issue #1720, opened 2022-02-28
nix improvements: auto-advancing, caching, and completions


colemickens:
Hi,

I noticed that some cool Rust projects  are adding flakes and I'm hoping we can get feature parity across them, and serve as a template for others. The current nix support is missing:
- auto-advancing of flake inputs (aka eventually you'll be building against old system packages potentially)
- caching of binary outputs
- installation of shell completion files (`zellij`'s recently added support does this) (at least, I don't see this in the flake, and I don't seem to have it active)

Somewhat separately...

I'm pondering a community provided build/cache service. Flakes would make this really easy, and the community provided service could even basically serve as a CI offload -- imagine that helix only runs CI every day, but nix-community runs it every hour. Unless you get really unfortunate, and if you trusted the community's cache, then it would mean that y'all nix builds would basically become no-ops.

the-mikedavis:
The caching part might be covered by https://github.com/helix-editor/helix/issues/1717 - is [cachix](https://www.cachix.org/) along the lines of what you were thinking? Also, the build/cache service sounds similar to [hydra](https://github.com/NixOS/hydra), no?

colemickens:
Yes, but I think on that point specifically my ideation is growing beyond "just another cachix" repo. There are implications to having to list+trust numerous caches - every cache can provide every store path, thus every cache can poison any path, and similarly every cache *can* need to be queried for every path (and there are potentially some client implementation details here that come into play).

I think I'm going to run with the idea of an externally-driven cache, just because I get the gut instinct from the interest in nixpkgs-wayland that there are enough Nix users wanting to be on the cutting edge that there would be some use. I can share details about that here, but again, no reason for that to block just a simple push to a helix cachix repo in the build job. (which can be setup *very* quickly with the Cachix GitHub Action)

For the sake of issue sanity, I'm fine with changing the scope of this issue to track just:
* auto-advancing flake inputs
* fleshing out the rest of the package derivation for completions
and leaving helix-specific caching to #1717. If I get anything useful done with an external cache I'll share a note on #1717.

the-mikedavis:
[...]

PR #3757, opened 2022-09-09
Support "preview" release CI runs

yvt:
Expands the trigger sources of the release CI workflow (`release.yml`), allowing the developers to test changes to `.github/workflows/release.yml` easily. The new trigger sources start the workflow in a *preview* mode, in which it publishes build outputs as a CI artifact ([example](https://github.com/yvt/helix/actions/runs/2987431541)) instead of creating a new release so that they can be manually inspected.

The following events trigger the preview mode:

- Pushing to any branch matching the glob pattern `patch/ci-release-*`.
- Opening a pull request that modifies `.github/workflows/release.yml`.
- Pushing versioning tags to a forked repository (they would create a release otherwise).

PR #4710, opened 2022-11-11
Run cachix workflow for x86_64-darwin

mtoohey31:
This merge request updates the Cachix workflow so it also runs on macOS. This will allow Nix users to avoid having to build Helix themselves on `x86_64-darwin` when they have configured the Helix Cachix as a binary cache.

**I have not tested these changes.** I don't have a Cachix cache set up myself, and I don't have permissions for the Helix one. I am as confident as I can be without having tested them that these changes are correct, since they're copied directly from [cachix-action's test workflow](https://github.com/cachix/cachix-action/blob/6a9a34cdd93d0ae4b4b59fd678660efb08109f2f/.github/workflows/test.yml#L10-L13), but someone else should definitely also take a careful look at them before this gets merged.

This partially addresses [my comment on #1720](https://github.com/helix-editor/helix/issues/1720#issuecomment-1306618514).

### DIFF
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -7,8 +7,11 @@ on:
 
 jobs:
   publish:
+    strategy:
+      matrix:
+        os: [ubuntu-latest, macos-latest]
     name: Publish Flake
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.os }}
     steps:
     - name: Checkout sources
       uses: actions/checkout@v3
--- a/.github/workflows/cachix.yml
+++ b/.github/workflows/cachix.yml
@@ -19,11 +19,16 @@ jobs:
     - name: Install nix
       uses: cachix/install-nix-action@v18
 
+    - name: Print toolchain regex
+      id: print-toolchain-regex
+      run: echo "TOOLCHAIN_REGEX=$(nix show-derivation .#default.rustToolchain.cargo < /dev/null | jq -r '.[].env.paths' | tr ' ' '|')" >> "$GITHUB_OUTPUT"
+
     - name: Authenticate with Cachix
       uses: cachix/cachix-action@v12
       with:
         name: helix
         authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+        pushFilter: ^(${{ steps.print-toolchain-regex.outputs.TOOLCHAIN_REGEX }}|.*-source|.*nixpkgs\.tar\.gz)$
 
     - name: Build nix flake
       run: nix build -L


the-mikedavis (approved):
My only hesitancy is that we only have 5GB of storage and each build takes a decent amount of space (although recent nix-cargo-integration changes have improved this). I suppose we could try it though and see if it ends up being too expensive to have multiple builds for the same cache

mtoohey31:
Ah, I didn't realize there was that limitation. Is there any chance we could cut down on what's being cached without loosing most of the benefits? In the output from [this run](https://github.com/helix-editor/helix/actions/runs/3393649168/jobs/5641231774) I see `rust-docs`, `rustc`, `rust-std`, and a couple other things contributing quite a lot of size. Is caching those necessary? It looks like cachix-action supports [includes/excludes](https://github.com/cachix/cachix-action/blob/master/action.yml#L16-L19).

I was hoping to get `aarch64-linux` cached too, but I guess that's probably pushing it if this is already borderline.

the-mikedavis:
Oh hmm yeah, trimming those out would save a good amount of space. I would imagine that those are served by the nixpkgs cache so I'm surprised to see them uploaded to the cache (unless they're created by nix-cargo-integration). Excluding them from the cache would be good 👍

mtoohey31:
Ok, I've pushed another commit that attempts to do that, but I'm much less confident about that working than the prior commit. Is there any way we can test it without merging it into main and breaking things?

the-mikedavis:
It would be cool to do something like https://github.com/helix-editor/helix/pull/3757/files#diff-87db21a973eed4fef5f32b267aa60fcee5cbdf03c67fafdc2a9b553bb0b15f34R9-R11 for this cachix file but I think it wouldn't work well because secrets (like the cachix publish key) are not given to forks.

You can set up a 5GB cache with your github account through cachix (needs a sign-up but is otherwise free) and then merge this to your fork. I would do that to test this out but the Helix cachix is tied to my account and the 5GB is account-wide 😅

mtoohey31:
Ok, I'll try and give that a go sometime soon, I'm going to mark this as a draft in the meantime though.

mtoohey31:
It turns out my assumption here: https://github.com/helix-editor/helix/pull/4710#issuecomment-1311721175 was incorrect. The rust toolchain _is_ created by nix-cargo-integration, since it's based on the rust-toolchain.toml file, meaning it's not cached in the nixpkgs cache. So excluding the toolchain from the cache isn't a good option.

I'm going to close this for now, because I haven't had any other good ideas on how to reduce the size of the things being cached.

PR #4716, opened 2022-11-11
significantly improve treesitter performance while editing large files

pascalkuthe:
While working on an unrelated PR related to treesitter performance I noticed that editing text is extremely slow for large files with loads of injections.
The testcase I was looking at during testing is the largest file in the Linux kernel: `drivers/gpu/drm/amd/include/asic_reg/dcn/dcn_3_2_0_sh_mask.h`

From the flamegraph (see below) I noticed that most of the time is actually spent inside helix comparing language layers.

image

After some digging I found that helix checks if an injection layer is already present (and hence can be incrementally updated) by iterating the entire list of injections. This has worst-case performance of `O(N^2)`.
For example the Linux kernel header I was investigating contains 27859 injections (comments) all inside the top layer, so about 776 million comparisons are performed (on every edit).

This PR fixes this problem by replacing the linear search with a Hashtable that is built at the beginning of the update function.
The Hashtable allows replacing the linear search with `O(1)` lookups, so the time complexity is decreased to `O(N)`.

I am using a `RawTable` from `hashbrown` here because I did not want to change the collection of `layers` (the closes substitute is a `IndexMap`, but that does not allow reusing indices like the `HopeSlotMap`). 

After this change there is still some delay in the huge Linux kernel header while typing but now typing a single latter takes maybe 0.4s compared to 2s previously (very unscientifically benchmarked).
Looking at the flamegraph (see below) now the bottleneck is treesitter itself (which might be potentially speedup in the future by setting a byte range on the query cursor although that will be more challenging).
While the file provided here was an extreme case, this should be a win (or not matter) for all languages/files.

image

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -14,6 +14,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "ahash"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "bf6ccdb167abbf410dcb915cabd428929d7f6a04980b54a11f26a39f1c7f7107"
+dependencies = [
+ "cfg-if",
+ "getrandom",
+ "once_cell",
+ "version_check",
+]
+
+[[package]]
 name = "aho-corasick"
 version = "0.7.18"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -400,18 +412,29 @@ version = "0.12.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8a9ee70c43aaf417c914396645a0fa852624801b24ebb7ae78fe8272889ac888"
 dependencies = [
- "ahash",
+ "ahash 0.7.6",
+]
+
+[[package]]
+name = "hashbrown"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33ff8ae62cd3a9102e5637afc8452c55acf3844001bd5374e0b0bd7b6616c038"
+dependencies = [
+ "ahash 0.8.2",
 ]
 
 [[package]]
 name = "helix-core"
 version = "0.6.0"
 dependencies = [
+ "ahash 0.8.2",
  "arc-swap",
  "bitflags",
  "chrono",
  "encoding_rs",
  "etcetera",
+ "hashbrown 0.13.1",
  "helix-loader",
  "log",
  "once_cell",
@@ -1288,7 +1311,7 @@ version = "0.1.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c5faade31a542b8b35855fff6e8def199853b2da8da256da52f52f1316ee3137"
 dependencies = [
- "hashbrown",
+ "hashbrown 0.12.3",
  "regex",
 ]
 
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -719,7 +719,7 @@ pub struct Syntax {
 
 impl fmt::Debug for Syntax {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        // skip the layer_lut and layer_lut hasher here as they
+        // skip the layer_lut and layer_lut hasher here as they do not implement Debug
         f.debug_struct("Syntax")
             .field("layers", &self.layers)
             .field("root", &self.root)
@@ -806,7 +806,7 @@ impl Syntax {
             // Ensure lut is large enough to hold all layers.
             // The lut should always be empty at this point because it is only
             // kept to avoid realloctions so rehashing is never requied (hence unreachable).
-            assert_eq!(self.layers_lut.len(), 0);
+            assert!(self.layers_lut.is_empty());
             self.layers_lut
                 .reserve(self.layers.len(), |_| unreachable!());
 
@@ -1190,7 +1190,7 @@ fn hash_injection_layer(
     let mut state = state.build_hasher();
     depth.hash(&mut state);
     // The transmute is necessary here because tree_sitter::Language does not derive Hash at the moment.
-    // However it does use #[repr] transpraerent so the transmute here is save
+    // However it does use #[repr] transparent so the transmute here is safe
     // as `Language` (which `Grammar` is an alias for) is just a newtype wrapper around a (thin) pointer.
     // This is also compatible with the PartialEq implementation of language
     // as that is just a pointer comparison.
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -18,7 +18,7 @@ use std::{
     cell::RefCell,
     collections::{HashMap, VecDeque},
     fmt,
-    hash::{BuildHasher, Hash, Hasher},
+    hash::{Hash, Hasher},
     mem::{replace, transmute},
     path::Path,
     str::FromStr,
@@ -709,23 +709,11 @@ thread_local! {
     })
 }
 
+#[derive(Debug)]
 pub struct Syntax {
     layers: HopSlotMap<LayerId, LanguageLayer>,
-    layers_lut: RawTable<LayerId>,
-    layers_lut_hasher: RandomState,
     root: LayerId,
     loader: Arc<Loader>,
-}
-
-impl fmt::Debug for Syntax {
-    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        // skip the layer_lut and layer_lut hasher here as they do not implement Debug
-        f.debug_struct("Syntax")
-            .field("layers", &self.layers)
-            .field("root", &self.root)
-            .field("loader", &self.loader)
-            .finish()
-    }
 }
 
 fn byte_range_to_str(range: std::ops::Range<usize>, source: RopeSlice) -> Cow<str> {
@@ -756,8 +744,6 @@ impl Syntax {
             root,
             layers,
             loader,
-            layers_lut: RawTable::new(),
-            layers_lut_hasher: RandomState::new(),
         };
 
         syntax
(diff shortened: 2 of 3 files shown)


helix-core/src/syntax.rs
```diff
@@ -706,13 +709,25 @@ thread_local! {
...
     layers: HopSlotMap<LayerId, LanguageLayer>,
+    layers_lut: RawTable<LayerId>,
+    layers_lut_hasher: RandomState,
     root: LayerId,
     loader: Arc<Loader>,
 }
 
+impl fmt::Debug for Syntax {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // skip the layer_lut and layer_lut hasher here as they
```
the-mikedavis, line 722:
```suggestion
        // skip the layer_lut and layer_lut hasher here as they do not implement Debug
```

helix-core/src/syntax.rs
```diff
@@ -786,7 +803,14 @@ impl Syntax {
                 }
             }
 
-            for layer in self.layers.values_mut() {
+            // Ensure lut is large enough to hold all layers.
+            // The lut should always be empty at this point because it is only
+            // kept to avoid realloctions so rehashing is never requied (hence unreachable).
+            assert_eq!(self.layers_lut.len(), 0);
```
the-mikedavis, line 809:
```suggestion
            assert!(self.layers_lut.is_empty());
```

helix-core/src/syntax.rs
```diff
@@ -1138,6 +1181,25 @@ pub struct LanguageLayer {
...
+fn hash_injection_layer(
+    state: &RandomState,
+    depth: u32,
+    config: &HighlightConfiguration,
+    ranges: &[Range],
+) -> u64 {
+    let mut state = state.build_hasher();
+    depth.hash(&mut state);
+    // The transmute is necessary here because tree_sitter::Language does not derive Hash at the moment.
+    // However it does use #[repr] transpraerent so the transmute here is save
```
the-mikedavis, line 1193:
```suggestion
    // However it does use #[repr] transparent so the transmute here is safe
```

helix-core/src/syntax.rs
```diff
@@ -706,13 +709,25 @@ thread_local! {
     })
 }
 
-#[derive(Debug)]
 pub struct Syntax {
     layers: HopSlotMap<LayerId, LanguageLayer>,
+    layers_lut: RawTable<LayerId>,
+    layers_lut_hasher: RandomState,
```
the-mikedavis, line 715: I think this may be clearer as `layer_table` and `layer_hasher`. Both may need doc comments too to point out what they're used for.

I'm afraid of abbreviations :P

the-mikedavis (commented):
This is cool, nice work diving into this!

I'm pretty well out of my depth on the content of the changes; I left some small readability and documentation nits

helix-core/src/syntax.rs
pascalkuthe: I changed the name and added some doc comments as suggested.
In the process I also found that I can use `PartialEq` and `Hash` directly instead of a freestanding function which looks more similar to a traditional `HashMap` and might be easier to understand.

pascalkuthe:
Test failure on windows looks spurious to me. Its some problems with saving files which is not touched at all by this PR

the-mikedavis:
Ah yep looks like a flake https://github.com/helix-editor/helix/actions/runs/3451158551/jobs/5760200072

I reran it and now it's green

helix-core/src/syntax.rs
```diff
@@ -706,13 +709,32 @@ thread_local! {
...
 pub struct Syntax {
     layers: HopSlotMap<LayerId, LanguageLayer>,
+    /// This table allows inverse indexing of `layers`.
+    /// That is by hashing a `Layer` you can find
+    /// the `LayerId` of an existing equivalent `Layer` in `layers`.
+    ///
+    /// This hashtable is populated inside the `update` function
+    /// and used to determine if a new layer exists for an injection
+    /// or if an existing layer needs to be updated.
+    layers_tabel: RawTable<LayerId>,
```
poliorcetics, line 721:
```suggestion
    layers_table: RawTable<LayerId>,
```

helix-core/src/syntax.rs
```diff
@@ -706,13 +709,32 @@ thread_local! {
...
+    layers_hasher: RandomState,
     root: LayerId,
     loader: Arc<Loader>,
 }
 
+impl fmt::Debug for Syntax {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // skip the layer_lut and layer_lut hasher here as they do not implement Debug
```
poliorcetics, line 729: Outdated names

helix-core/src/syntax.rs
```diff
@@ -786,7 +810,13 @@ impl Syntax {
                 }
             }
 
-            for layer in self.layers.values_mut() {
+            self.layers_tabel.clear_no_drop();
+            // Ensure table is large enough to hold all layers.
+            // Because the table is cleared first, rehashing is never requied (hence unreachable).
```
poliorcetics, line 815:
```suggestion
            // Because the table is cleared first, rehashing is never required (hence unreachable).
```

helix-core/src/syntax.rs
```diff
@@ -856,6 +886,14 @@ impl Syntax {
...
                     }
                 }
+
+                let hash = self.layers_hasher.hash_one(layer);
+                // Safety: insert_no_grow is unsafe because it assumes that the table
+                // has enough capacity to hold additional elements.
+                // This is always the case as we reserved enough capacity above.
+                unsafe {
+                    self.layers_tabel.insert_no_grow(hash, layer_id);
+                }
```
poliorcetics, line 896:
```suggestion
                unsafe { self.layers_tabel.insert_no_grow(hash, layer_id) };
```

I think the line is short enough that `rustfmt` will make it a one-liner if the `;` is after the `unsafe` block

archseer:
> For example the Linux kernel header I was investigating contains 27859 injections (comments) all inside the top layer, so about 776 million comparisons are performed (on every edit).

I wish we could reduce the number of empty injections. `tree-sitter-comment` seems like a bit of a hack and a lot of the layers will just be empty (I had to add a fast path for that during render)

pascalkuthe:
> > For example the Linux kernel header I was investigating contains 27859 injections (comments) all inside the top layer, so about 776 million comparisons are performed (on every edit).
> 
> I wish we could reduce the number of empty injections. `tree-sitter-comment` seems like a bit of a hack and a lot of the layers will just be empty (I had to add a fast path for that during render)

I think with this PR the overhead of these empty injections is pretty small.
I tried disabling the comment injections doesn't really change the performance in any noticeable way (with this PR) even for that huge header (neither while editing or just navigating). Almost all performance overhead comes from the query cursor in the root layer being slow (both during the update and during rendering).
I agree it doesn't feel great but I can't think of a better way to highlight comments and it doesn't seem to cause any problems once this PR is merged (which will also offer improvements for other languages with many injections like markdown).

helix-core/src/syntax.rs
```diff
@@ -769,30 +772,38 @@ impl Syntax {
...
         let edits = generate_edits(old_source, changeset);
 
+        // This table allows inverse indexing of `layers`.
+        // That is by hashing a `Layer` you can find
+        // the `LayerId` of an existing equivalent `Layer` in `layers`.
+        //
+        // It is used to determine if a new layer exists for an injection
+        // or if an existing layer needs to be updated.
+        let mut layers_table = RawTable::with_capacity(self.layers.len());
+        let layers_hasher = RandomState::new();
```
archseer, line 782: Why do we need to pull in `RawTable`? Why not just `HashMap::with_capacity()`?

helix-core/src/syntax.rs
```diff
@@ -857,6 +868,12 @@ impl Syntax {
                     }
                 }
             }
+
+            let hash = layers_hasher.hash_one(layer);
+            // Safety: insert_no_grow is unsafe because it assumes that the table
+            // has enough capacity to hold additional elements.
+            // This is always the case as we reserved enough capacity above.
+            unsafe { layers_table.insert_no_grow(hash, layer_id) };
```
archseer, line 876: Why not just `insert()`? It's guaranteed to avoid allocation if there's enough capacity, but it won't panic if that's not the case

helix-core/src/syntax.rs
```diff
@@ -1138,6 +1151,34 @@ pub struct LanguageLayer {
...
+/// See its documentation for details.
+impl Hash for LanguageLayer {
+    fn hash<H: Hasher>(&self, state: &mut H) {
+        self.depth.hash(state);
+        // The transmute is necessary here because tree_sitter::Language does not derive Hash at the moment.
+        // However it does use #[repr] transparent so the transmute here is safe
+        // as `Language` (which `Grammar` is an alias for) is just a newtype wrapper around a (thin) pointer.
+        // This is also compatible with the PartialEq implementation of language
+        // as that is just a pointer comparison.
+        let language: *const () = unsafe { transmute(self.config.language) };
```
archseer, line 1176: `&self.config.language as *const _ as *const ()` avoids the transmute:

```rust
fn main() {
    let  a = vec![1];
    
    let ptr = &a as *const _ as *const ();
    println!("{:?}", ptr);
}
```

archseer: Though hmm, since it's a wrapper type it might not be the same thing

pascalkuthe: Yeah its not the same, because Grammar is a Newtype wrapper around a pointer:
``` rust
struct Grammar(*const ());
```

You can definitely implement this with pointer cast but at that point you are re-implementing `transmute_copy` :
``` rust
let language_ptr: *const Grammar = &self.config.language as *const Grammar;
let raw_language_ptr: *const *const () = language as *const *const ();
let language = *raw_language_ptr; 
```

I think this is one of those rare cases where transmute is the cleanest option. I use it only to get to the inner value of the newtype wrapper here (which is `repr(transpraent)`)

helix-core/src/syntax.rs
pascalkuthe: The point of the `RawTable` is not the `capacity`/`insert_no_grow`. That mainly just a nice bonus.
A `HashMap<K,V>` is essentially a `RawTable<(K,V)>` internally. However we want a `LanguageLayer -> LayerId` map here so using a normal `HashMap<LanguageLayer, LayerId>` would require that we clone all `LanguageLayer`s (we can not store references into `Syntax::layers` because it is mutated later).
With a `RawTable` we can just store the `LayerId` as `RawTable<LayerId>` .
When we need access to the original keys to resolve hash collisions during lookup later we can just use the `LayerId` to index back into `self.layers`.

`IndexMap` does essentially the same thing  internally (it just doesn't fit well here because we want to keep using a `HopSlotMap` whereas `indexmap` forces us to use essentially a `Vec`).

I also use the exact same strategy with `RawTable` for the interning in `imara-diff` (to avoid pulling in `indexmap` which offers a lot more functionality). So we will depend on `hashbrown` anyway (in fact we already do so as a dependy of `unicode-linebreak`)

helix-core/src/syntax.rs
pascalkuthe: I usually use `insert_no_grow` with `RawTable` if the capacity is trivially large enough for two reasons:
* It's faster: Swiss tables need to check an additional byte to determine if a bucket is empty so the overhead is larger then for a `Vec` because you need to read from a random location in the table. The complexity of this bounds check also means its basically never optimized out.
* `insert` requires a closure for rehashing existing entries in case the table needs to be resized (which is impossible because we have enough capacity). That either adds a unnecessary colsure that calculates a hash or a weird looking `unrechable` as shown below so I usually prefer the `insert_no_grow` variant over the other options:
``` rust
layers_table.insert(hash, layer_id, |_| unreachable!("table should be large enough"));
layers_table.insert(hash, layer_id, |rehashed_layer| hasher.hash_one(self.layers[rehashed_layer])));
```

Cargo.lock
```diff
@@ -400,18 +412,29 @@ version = "0.12.3"
...
+]
+
+[[package]]
+name = "hashbrown"
+version = "0.13.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "33ff8ae62cd3a9102e5637afc8452c55acf3844001bd5374e0b0bd7b6616c038"
+dependencies = [
+ "ahash 0.8.2",
 ]
```
archseer, line 425: Looks like this is pulling in an older version of `hashbrown`

pascalkuthe: This PR actually pulls in the newest version of `hashbrown`.
Both `textwrap` and `unicode-linebreaks` still depend on an old version.
`imara-diff` uses the newest version aswell and I do not want to downgrade `hashbrown` in `imara-diff`, because other crates that depend on it (`gitoxide` for example) already use the newest version (and in the future crates will rather upgrade then downgrade).

It's once again one of those unfortunate ecosystem splits where `unicode-lines` and `textwrap` want to maintain a MSRV of 1.56 but  `hashbrown` had raised it MSRV to 1.61.
It's a bit unfortunate that these ecosystem splits happens but I think depending on the newest version and waiting for other dependencies to bump is the correct way to go.

Issue #4009, opened 2022-09-28
Granular C-o


zim0369:
Selections in line can't be commented. The entire line gets commented. Granular control with `C-c` would fix this.

dead10ck:
Would this just insert a comment marker to the left of the cursor?

zim0369:
There's some plugin in lunarvim that comments text using `gc<movement>` or visual selection and `gc`
So here,
```
This part|[cursor]is
```
`|`(cursor) & `[]` is the selection.
Running `C-c` here should comment the word "cursor":
```
This part #cursor is
```
I'm not saying that this is a clever binding. This is just the idea. If planned, then better bindings could be implemented.

CptPotato:
I actually like the current behavior. If this is implemented I'd prefer it as a separate command.

zim0369:
> I actually like the current behavior. If this is implemented I'd prefer it as a separate command.

Yes, the current behavior is short and fast. I don't want this feature to slow down or interfere with the current implementation.

filipdutescu:
I think that for languages which support it, an option could be added to comment only the selection (from your example it appears as though the comment is from `#cursor` to the end of the line).

So for

>```
>This part|[cursor]is
>```
> `|`(cursor) & `[]` is the selection.

you'd have:

1. Comment from selection to end of the line
```
This part #cursor is
```

2. Comment only the selection
```
This part /*cursor*/ is
```

AK-10:
How about implementing a `toggle_block_comments` without touching the `toggle_comments` like vscode?

AK-10:
I'm trying implement the toggle_block_comments command

gabydd:
I actually have mostly implemented that if you want to take a look: #4718

gabydd:
I would say it's pretty much fully implemented I just need to add tests and add it to the docs

AK-10:
@gabydd 
Oh, looks pretty good!

looks pretty good
But I think it behaves a little differently than `toggle`

I think vscode's Toggle Block Comment behaves as `toggle`

https://user-images.githubusercontent.com/27846333/204569530-bb167b9b-e314-4bc4-8ffd-6c0bb677c407.mp4

gabydd:
Yeah at this moment the surrounding tokens need to be inside the selection, but i think I should be able to get it to search past whitespace surrounding a selection for comment tokens

cotneit:
Looks like this can be closed now that #4718 is merged?

`<space>-C` - block comment selection
[...]

PR #4718, opened 2022-11-11
Toggle block comments

### LINKED ISSUE #1505: Support for block comments
Add support for block comments. (line comment support per language was added in #463)

There are multiple use-cases for this:
 - Many languages don't support line comments but do support block comments, e.g. `html`/`xml`, `css`, old `c`, `ocaml`. the `toggle_comments` command could fallback to a multi-line comment.
 - Some users / codebases prefer line comments. That preference could be set in `languages.toml`.
 - A comment textobject could be interesting too, to select whole comments or comment out selections.

gabydd:
this is an initial implementation of toggling block comments which fixes #1505 it still needs
- [ ] decision for keybindings, for now I put block comments under space-c-b and line comments to space-c-c
- [x] if we should have a single command for toggle comment that does either block or line toggle
- [x] add more block comment tokens to language.toml
- [ ] ~can we get comment tokens from treesitter~
- [x] right now toggle a block doesn't extend the selection, should it do that? extends for non linewise block comments
- [x] tests
- [x] comments
- [x] add to docs when keybindings are finalized

any other feedback appreciated, this is still a draft and nothing is set in stone

some implementation notes:
- the command normalizes the commenting across the selections so if one selection is uncommented all selections get commented, this can be changed if needed
- comments surround the content of each selection ignoring whitespace when commenting and leave any whitespace surrounding block comment tokens when uncommenting
- ~~i tried to do the same thing as `find_first_non_whitespace_character` in `find_last_non_whitespace_character` but apparently `Chars` doesn't implement `DoubleEndedIterator` so I just iterate over a reversed range of chars len instead. I don't think this should have performance implications on really long selections, but someone else might know better.~~ pascal figured this out

[video of it in action](https://asciinema.org/a/kkAUgOnELliqhspIUl4Umiy9q)

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -12,6 +12,7 @@
     - [Match mode](#match-mode)
     - [Window mode](#window-mode)
     - [Space mode](#space-mode)
+      - [Comment mode](#comment-mode)
       - [Popup](#popup)
     - [Unimpaired](#unimpaired)
 - [Insert mode](#insert-mode)
@@ -289,6 +290,9 @@ This layer is a kludge of mappings, mostly pickers.
 | `h`     | Select symbol references (**LSP**)                                      | `select_references_to_symbol_under_cursor` |
 | `'`     | Open last fuzzy picker                                                  | `last_picker`                              |
 | `w`     | Enter [window mode](#window-mode)                                       | N/A                                        |
+| `c`     | Comment/uncomment selections                                            | `toggle_comments`                          |
+| `C`     | Block comment/uncomment selections                                      | `toggle_block_comments`                    |
+| `Alt-c` | Line comment/uncomment selections                                       | `toggle_line_comments`                     |
 | `p`     | Paste system clipboard after selections                                 | `paste_clipboard_after`                    |
 | `P`     | Paste system clipboard before selections                                | `paste_clipboard_before`                   |
 | `y`     | Yank selections to clipboard                                            | `yank_to_clipboard`                        |
--- a/book/src/languages.md
+++ b/book/src/languages.md
@@ -42,7 +42,7 @@ name = "mylang"
 scope = "source.mylang"
 injection-regex = "mylang"
 file-types = ["mylang", "myl"]
-comment-token = "#"
+comment-tokens = "#"
 indent = { tab-width = 2, unit = "  " }
 formatter = { command = "mylang-formatter" , args = ["--stdin"] }
 language-servers = [ "mylang-lsp" ]
@@ -61,7 +61,8 @@ These configuration keys are available:
 | `roots`               | A set of marker files to look for when trying to find the workspace root. For example `Cargo.lock`, `yarn.lock` |
 | `auto-format`         | Whether to autoformat this language when saving               |
 | `diagnostic-severity` | Minimal severity of diagnostic for it to be displayed. (Allowed values: `Error`, `Warning`, `Info`, `Hint`) |
-| `comment-token`       | The token to use as a comment-token                           |
+| `comment-tokens`      | The tokens to use as a comment token, either a single token `"//"` or an array `["//", "///", "//!"]` (the first token will be used for commenting). Also configurable as `comment-token` for backwards compatibility|
+| `block-comment-tokens`| The start and end tokens for a multiline comment either an array or single table of `{ start = "/*", end = "*/"}`. The first set of tokens will be used for commenting, any pairs in the array can be uncommented |
 | `indent`              | The indent to use. Has sub keys `unit` (the text inserted into the document when indenting; usually set to N spaces or `"\t"` for tabs) and `tab-width` (the number of spaces rendered for a tab) |
 | `language-servers`    | The Language Servers used for this language. See below for more information in the section [Configuring Language Servers for a language](#configuring-language-servers-for-a-language)   |
 | `grammar`             | The tree-sitter grammar to use (defaults to the value of `name`) |
--- a/helix-core/src/comment.rs
+++ b/helix-core/src/comment.rs
@@ -1,9 +1,12 @@
 //! This module contains the functionality toggle comments on lines over the selection
 //! using the comment character defined in the user's `languages.toml`
 
+use smallvec::SmallVec;
+
 use crate::{
-    find_first_non_whitespace_char, Change, Rope, RopeSlice, Selection, Tendril, Transaction,
+    syntax::BlockCommentToken, Change, Range, Rope, RopeSlice, Selection, Tendril, Transaction,
 };
+use helix_stdx::rope::RopeSliceExt;
 use std::borrow::Cow;
 
 /// Given text, a comment token, and a set of line indices, returns the following:
@@ -22,12 +25,12 @@ fn find_line_comment(
 ) -> (bool, Vec<usize>, usize, usize) {
     let mut commented = true;
     let mut to_change = Vec::new();
-    let mut min = usize::MAX; // minimum col for find_first_non_whitespace_char
+    let mut min = usize::MAX; // minimum col for first_non_whitespace_char
     let mut margin = 1;
     let token_len = token.chars().count();
     for line in lines {
         let line_slice = text.line(line);
-        if let Some(pos) = find_first_non_whitespace_char(line_slice) {
+        if let Some(pos) = line_slice.first_non_whitespace_char() {
             let len = line_slice.len_chars();
 
             if pos < min {
@@ -94,6 +97,222 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
     Transaction::change(doc, changes.into_iter())
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub enum CommentChange {
+    Commented {
+        range: Range,
+        start_pos: usize,
+        end_pos: usize,
+        start_margin: bool,
+        end_margin: bool,
+        start_token: String,
+        end_token: String,
+    },
+    Uncommented {
+        range: Range,
+        start_pos: usize,
+        end_pos: usize,
+        start_token: String,
+        end_token: String,
+    },
+    Whitespace {
+        range: Range,
+    },
+}
+
+pub fn find_block_comments(
+    tokens: &[BlockCommentToken],
+    text: RopeSlice,
+    selection: &Selection,
+) -> (bool, Vec<CommentChange>) {
+    let mut commented = true;
+    let mut only_whitespace = true;
+    let mut comment_changes = Vec::with_capacity(selection.len());
+    let default_tokens = tokens.first().cloned().unwrap_or_default();
+    // TODO: check if this can be removed on MSRV bump
+    #[allow(clippy::redundant_clone)]
+    let mut start_token = default_tokens.start.clone();
+    #[allow(clippy::redundant_clone)]
+    let mut end_token = default_tokens.end.clone();
+
+    let mut tokens = tokens.to_vec();
+    // sort the tokens by length, so longer tokens will match first
+    tokens.sort_by(|a, b| {
+        if a.start.len() == b.start.len() {
+            b.end.len().cmp(&a.end.len())
+        } else {
+            b.start.len().cmp(&a.start.len())
+        }
+    });
+    for range in selection {
+        let selection_slice = range.slice(text);
+        if let (Some(start_pos), Some(end_pos)) = (
+            selection_slice.first_non_whitespace_char(),
+            selection_slice.last_non_whitespace_char(),
+        ) {
+            let mut line_commented = false;
+            let mut after_start = 0;
+            let mut before_end = 0;
+            let len = (end_pos + 1) - start_pos;
+
+            for BlockCommentToken { start, end } in &tokens {
+                let start_len = start.chars().count();
+                let end_len = end.chars().count();
+                after_start = start_pos + start_len;
+                before_end = end_pos.saturating_sub(end_len);
+
+                if len >= start_len + end_len {
+                    let start_fragment = selection_slice.slice(start_pos..after_start);
+                    let end_fragment = selection_slice.slice(before_end + 1..end_pos + 1);
+
+                    // block commented with these tokens
+                    if start_fragment == start.as_str() && end_fragment == end.as_str() {
+                        start_token = start.to_string();
+                        end_token = end.to_string();
+                        line_commented = true;
+                        break;
+                    }
+                }
+            }
+
+            if !line_commented {
+                comment_changes.push(CommentChange::Uncommented {
+                    range: *range,
+                    start_pos,
+                    end_pos,
+                    start_token: default_tokens.start.clone(),
+                    end_token: default_tokens.end.clone(),
+                });
+                commented = false;
+            } else {
+                comment_changes.push(CommentChange::Commented {
+                    range: *range,
+                    start_pos,
+                    end_pos,
+                    start_margin: selection_slice
+                        .get_char(after_start)
+                        .map_or(false, |c| c == ' '),
+                    end_margin: after_start != before_end
+                        && selection_slice
+                            .get_char(before_end)
+                            .map_or(false, |c| c == ' '),
+                    start_token: start_token.to_string(),
+                    end_token: end_token.to_string(),
+                });
+            }
+            only_whitespace = false;
+        } else {
+            comment_changes.push(CommentChange::Whitespace { range: *range });
+        }
+    }
+    if only_whitespace {
+        commented = false;
+    }
+    (commented, comment_changes)
+}
+
+#[must_use]
+pub fn create_block_comment_transaction(
+    doc: &Rope,
+    selection: &Selection,
+    commented: bool,
+    comment_changes: Vec<CommentChange>,
+) -> (Transaction, SmallVec<[Range; 1]>) {
+    let mut changes: Vec<Change> = Vec::with_capacity(selection.len() * 2);
+    let mut ranges: SmallVec<[Range; 1]> = SmallVec::with_capacity(selection.len());
+    let mut offs = 0;
+    for change in comment_changes {
+        if commented {
+            if let CommentChange::Commented {
+                range,
+                start_pos,
+                end_pos,
+                start_token,
+                end_token,
+                start_margin,
+                end_margin,
+            } = change
+            {
+                let from = range.from();
+                changes.push((
+                    from + start_pos,
+                    from + start_pos + start_token.len() + start_margin as usize,
+                    None,
+                ));
+                changes.push((
+                    from + end_pos - end_token.len() - end_margin as usize + 1,
+                    from + end_pos + 1,
+                    None,
+                ));
+            }
+        } else {
+            // uncommented so manually map ranges through changes
+            match change {
+                CommentChange::Uncommented {
+                    range,
+                    start_pos,
+                    end_pos,
+                    start_token,
+                    end_token,
+                } => {
+                    let from = range.from();
+                    changes.push((
+                        from + start_pos,
+                        from + start_pos,
+                        Some(Tendril::from(format!("{} ", start_token))),
+                    ));
+                    changes.push((
+                        from + end_pos + 1,
+                        from + end_pos + 1,
+                        Some(Tendril::from(format!(" {}", end_token))),
+                    ));
+
+                    let offset = start_token.chars().count() + end_token.chars().count() + 2;
+                    ranges.push(
+                        Range::new(from + offs, from + offs + end_pos + 1 + offset)
+                            .with_direction(range.direction()),
+                    );
+                    offs += offset;
+                }
+                CommentChange::Commented { range, .. } | CommentChange::Whitespace { range } => {
+                    ranges.push(Range::new(range.from() + offs, range.to() + offs));
+                }
+            }
+        }
+    }
+    (Transaction::change(doc, changes.into_iter()), ranges)
+}
+
+#[must_use]
+pub fn toggle_block_comments(
+    doc: &Rope,
+    selection: &Selection,
+    tokens: &[BlockCommentToken],
+) -> Transaction {
+    let text = doc.slice(..);
+    let (commented, comment_changes) = find_block_comments(tokens, text, selection);
+    let (mut transaction, ranges) =
+        create_block_comment_transaction(doc, selection, commented, comment_changes);
+    if !commented {
+        transaction = transaction.with_selection(Selection::new(ranges, selection.primary_index()));
+    }
+    transaction
+}
+
+pub fn split_lines_of_selection(text: RopeSlice, selection: &Selection) -> Selection {
+    let mut ranges = SmallVec::new();
+    for range in selection.ranges() {
+        let (line_start, line_end) = range.line_range(text.slice(..));
+        let mut pos = text.line_to_char(line_start);
+        for line in text.slice(pos..text.line_to_char(line_end + 1)).lines() {
+            let start = pos;
+            pos += line.len_chars();
+            ranges.push(Range::new(start, pos));
+        }
+    }
+    Selection::new(ranges, 0)
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
@@ -148,5 +367,50 @@ mod test {
         assert!(selection.len() == 1); // to ignore the selection unused warning
 
         // TODO: account for uncommenting with uneven comment indentation
+    }
+
+    #[test]
+    fn test_find_block_comments() {
+        // three lines 5 characters.
+        let mut doc = Rope::from("1\n2\n3");
+        // select whole document
+        let selection = Selection::single(0, doc.len_chars());
+
+        let text = doc.slice(..);
+
+        let res = find_block_comments(&[BlockCommentToken::default()], text, &selection);
+
+        assert_eq!(
+            res,
+            (
+                false,
+                vec![CommentChange::Uncommented {
+                    range: Range::new(0, 5),
+                    start_pos: 0,
+                    end_pos: 4,
+                    start_token: "/*".to_string(),
+                    end_token: "*/".to_string(),
+                }]
+            )
+        );
+
+        // comment
+        let transaction = toggle_block_comments(&doc, &selection, &[BlockCommentToken::default()]);
+        transaction.apply(&mut doc);
+
+        assert_eq!(doc, "/* 1\n2\n3 */");
+
+        // uncomment
+        let selection = Selection::single(0, doc.len_chars());
+        let transaction = toggle_block_comments(&doc, &selection, &[BlockCommentToken::default()]);
+        transaction.apply(&mut doc);
+        assert_eq!(doc, "1\n2\n3");
+
+        // don't panic when there is just a space in comment
+        doc = Rope::from("/* */");
+        let selection = Selection::single(0, doc.len_chars());
+        let transaction = toggle_block_comments(&doc, &selection, &[BlockCommentToken::default()]);
+        transaction.apply(&mut doc);
+        assert_eq!(doc, "");
     }
 }
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -99,7 +99,19 @@ pub struct LanguageConfiguration {
     pub shebangs: Vec<String>, // interpreter(s) associated with language
     #[serde(default)]
     pub roots: Vec<String>, // these indicate project roots <.git, Cargo.toml>
-    pub comment_token: Option<String>,
+    #[serde(
+        default,
+        skip_serializing,
+        deserialize_with = "from_comment_tokens",
+        alias = "comment-token"
+    )]
+    pub comment_tokens: Option<Vec<String>>,
+    #[serde(
+        default,
+        skip_serializing,
+        deserialize_with = "from_block_comment_tokens"
+    )]
+    pub block_comment_tokens: Option<Vec<BlockCommentToken>>,
     pub text_width: Option<usize>,
     pub soft_wrap: Option<SoftWrap>,
 
@@ -238,6 +250,59 @@ impl<'de> Deserialize<'de> for FileType {
 
         deserializer.deserialize_any(FileTypeVisitor)
     }
+}
+
+fn from_comment_tokens<'de, D>(deserializer: D) -> Result<Option<Vec<String>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum CommentTokens {
+        Multiple(Vec<String>),
+        Single(String),
+    }
+    Ok(
+        Option::<CommentTokens>::deserialize(deserializer)?.map(|tokens| match tokens {
+            CommentTokens::Single(val) => vec![val],
+            CommentTokens::Multiple(vals) => vals,
+        }),
+    )
+}
+
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct BlockCommentToken {
+    pub start: String,
+    pub end: String,
+}
+
+impl Default for BlockCommentToken {
+    fn default() -> Self {
+        BlockCommentToken {
+            start: "/*".to_string(),
+            end: "*/".to_string(),
+        }
+    }
+}
+
+fn from_block_comment_tokens<'de, D>(
+    deserializer: D,
+) -> Result<Option<Vec<BlockCommentToken>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    #[derive(Deserialize)]
+    #[serde(untagged)]
+    enum BlockCommentTokens {
+        Multiple(Vec<BlockCommentToken>),
+        Single(BlockCommentToken),
+    }
+    Ok(
+        Option::<BlockCommentTokens>::deserialize(deserializer)?.map(|tokens| match tokens {
+            BlockCommentTokens::Single(val) => vec![val],
+            BlockCommentTokens::Multiple(vals) => vals,
+        }),
+    )
 }
 
 #[derive(Clone, Copy, Debug, Serialize, Deserialize, PartialEq, Eq, Hash)]
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -12,7 +12,7 @@ pub use typed::*;
 use helix_core::{
     char_idx_at_visual_offset, comment,
     doc_formatter::TextFormat,
-    encoding, find_first_non_whitespace_char, find_workspace, graphemes,
+    encoding, find_workspace, graphemes,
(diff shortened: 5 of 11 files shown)


gabydd:
I think I can make it so that there is one command that can toggle block and line comments,, then seperate commands for just block and line toggling, my proposed algorithm is:

if only one of single line token or block tokens exist, toggle that, else do the following:
- check if all selections are line commented
	- if they are then line toggle
- check if all selections are block toggled
	- if they are then line toggle
- check if all selections are single line
	- if they are then line toggle
	- should this check for block comments?
- else toggle block
	- could potentially also check and skip over already line commented code, but then block toggling would also do line toggling

gabydd:
I implemented the above, and then separated line comment toggling into it's own command that is under `space+c+l`

there is definitely ways to make block commenting smarter, for example comment a block if selection is surrounded by only whitespace and then comment token or uncomment block that is surrounded by other characters then whitespace in a selection(though I think this could actually extend the comment block maybe). This could be saved for future pr if this functionality is wanted.

helix-core/src/comment.rs
```diff
@@ -94,6 +94,168 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+        find_block_comments(comment_tokens.0, comment_tokens.1, text, selection);
+
+    // line comment when only line comment token available
+    if matches!(token, Some(_)) && !matches!(tokens, Some(_)) {
+        return true;
+        // block comment when only block comment tokens available
+    }
+    if !matches!(token, Some(_)) && matches!(tokens, Some(_)) {
+        return false;
+    }
```
pascalkuthe, line 242: small nit: maybe move this in front of the comment search so that languages with just one type of comment avoid searching for the default comments. As  discussed in the matrix chat it would be nice to specical case languages with just block comments here, so returning an enum would be good. Also I think you can make this clearer as a  match:
``` rust 
let (token, tokens) = match (token, tokens){
 (Some(token), Some(tokens)) => (token, tokens),
 (None, None) => ("//", ("/*","*/")),
 (Some(_), None) => return CommentType::Line,
 (None, Some(_)) => return CommentType::BlockAsLineFallback,
};
```

pascalkuthe (changes requested):
Thanks for working on this! Comment as discussed in matrix

helix-core/src/comment.rs
gabydd: thanks for this review, I implemented it with some changes. The enum and match was a good call here. Turned out ``BlockAsLineFallback`` should only be returned if the selections are only one line long so I changed it to check for that(almost left this for tomorrow but I had an epiphany). I also didn't have the enum variants assigned a value because it wasn't needed and I didn't want to deal with lifetimes. Thank you again for this review it was very helpful.

gabydd:
This PR should now be ready for review, some things to note:
- selection isn't extended to encompass the comment, wasn't sure how or if I should do this
- keybindings should be chosen, I know some people wanted to switch to ``#``, I don't have strong opinions about this both are great to me
- this is pretty much first rust stuff I have done, so if anything seems wrong please don't assume it's not you probably know better then me.
- this is breaking change to how comment toggling works
Also want to thank everyone for all the work they have put into helix, using and contributing to it has been a joy
(the doc failing seems to be based on a failed network call, shouldn't effect this pr)

pascalkuthe:
Thanks for working on this! I think block comments are useful to have in particular for languages that do not have line comments.
I tried this out a while ago with markdown and it worked well.

I had this in my daily driver for a while and have some thoughts. I think automatically switching between line comments and block commenys based upon whether more then 1 char is selectes is nice to have but I found myself frustrated by this as a default. 

For languages that support them I almost always want line comments because of the more granular control (comment ten lines, later uncomment only two of them or move some part of the comment elsewhere without having to add new commentnstsrt/end) so I would prefer to keep that as is. 
However for languages that do not have line comments by default it should fall back to adding a block comment to every selected lime (same for `<space>cl`).

 I think the automatic option is still useful but perhaphs as a secondary command (maybe mapped to `C-C` by default and people who prefer it the other way around can swap those two?).

If other people (@archseer whatdo you think?) prefer the default as in this PR I think its fine too as the  I keybindings can be swappes in my config. I think the suggesting about using block comments as a line comment fallback should definitely be implement

gabydd:
Thanks for your input @pascalkuthe it's very helpful, and unless someone else objects I think I will implement it as the default. This seems simpler and more intuitive to the user, I don't have time today to implement this but it was in the back of my mind after I marked it for review so it shouldn't take me too long.

gabydd:
sorry for the wait on this I have been pretty busy, the new relevant changes are more inline with @pascalkuthe suggestions(if you have the time to take a look at the new behaviour that would be appreciated, I am sure you are pretty busy right now so it's totally okay if you don't do an actual review)

the key changes are summarized here:
- toggle comments always does a line comment or uses block as line fallback (this makes it much less of a breaking change)
- it only does a regular block comment when the selection is already block commented

- line comment and block comment now both fallback to each other if there token is not available for that language

whats left at this point is to add this to the docs and look into making some more meaningful tests

gibbz00:
How will this affect the default `"C-c" => toggle_comments` binding?

gabydd:
Only effects if there is no line ``comment-token`` for a language(falls back to using the block comment tokens to comment each line instead)or if the whole selection is block commented other then that ``toggle_comments`` stays the same

pascalkuthe:
@gabydd Is this ready for review?

pascalkuthe:
ok thanks for letting me know. I will think about that during review

gabydd:
just fixed the conflict and rebased on master, I realize there is probably some new languages that don't have block comment tokens so will add those

gabydd:
there is now documentation as well, I would add block-comments to the tutor but then the user would have to change the language to one that only supports block comments and that seems a little bit confusing for the tutorial

book/src/keymap.md
```diff
@@ -289,6 +290,7 @@ This layer is a kludge of mappings, mostly pickers.
 | `h`     | Select symbol references (**LSP**)                                      | `select_references_to_symbol_under_cursor` |
 | `'`     | Open last fuzzy picker                                                  | `last_picker`                              |
 | `w`     | Enter [window mode](#window-mode)                                       | N/A                                        |
+| `w`     | Enter [comment mode](#comment-mode                                      | N/A                                        |
```
pascalkuthe, line 293:
```suggestion
| `c`     | Enter [comment mode](#comment-mode                                      | N/A                                        |
```

helix-core/src/comment.rs
```diff
@@ -94,6 +98,226 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            let len = selection_slice.len_chars();
+
+            // selection can be shorter than open_pos + open len
+            let open_fragment = Cow::from(
+                selection_slice.slice(open_pos..std::cmp::min(open_pos + open.len(), len)),
+            );
+            let close_fragment = Cow::from(
+                selection_slice
+                    .slice(close_pos - std::cmp::min(close.len() - 1, close_pos)..close_pos + 1),
+            );
```
pascalkuthe, line 141: no need to create a `Cow` you can compare `RopeSlice` to `str` directly. Also instead of bounding the size, you can simply check upfront that slice is large enough.

You also mix up byte and char indices here. Ropey and helix use char indices by default while strings use byte indices. I this case you should probably just calculate the char len of `open` and `close` at the start of the function and use char indices everywhere. This won't make a difference for ASCII comments but we should correctly handle non-ascii tokens even if they are rare in practice.

helix-core/src/comment.rs
```diff
@@ -94,6 +98,226 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            let close_margin = if open_pos + open.len()
+                != close_pos - std::cmp::min(close.len(), close_pos)
+            {
+                match selection_slice.get_char(close_pos - std::cmp::min(close.len(), close_pos)) {
+                    Some(c) if c == ' ' => 1,
+                    _ => 0,
+                }
+            } else {
+                0
+            };
```
pascalkuthe, line 158: this code is hard to read. I would simply do this after you have determined that there is indeed a block comment. Then you can calculate the inner slice and use `get_char(0)` and `get_char(slice.len().satusating_sub(1))`. 

you can use `map_or` to make this more readable, also store the margins as a bools. You can then use `as usize` when generating the actual transaction

helix-core/src/comment.rs
```diff
@@ -94,6 +98,226 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+    open_margin: usize,
+    close_margin: usize,
+}
+
+fn find_block_comments(
+    open: &str,
+    close: &str,
+    text: RopeSlice,
+    selection: &Selection,
+) -> (bool, Vec<CommentChange>) {
```
pascalkuthe, line 123: this could return an iterator I think which would be more efficient/idiomatic

pascalkuthe (changes requested):
I haven't finished yet but I already have some technical comments

helix-core/src/comment.rs
```diff
@@ -94,6 +98,226 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
     Transaction::change(doc, changes.into_iter())
 }
 
+fn find_last_non_whitespace_char(text: RopeSlice) -> Option<usize> {
+    (0..text.len_chars()).rev().find(|&i| {
+        text.get_char(i)
+            .map(|ch| !ch.is_whitespace())
+            .unwrap_or(false)
+    })
+}
```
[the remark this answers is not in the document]
gabydd: Yeah I tried to do it that way but for some reason using the reversed iterator I was not able to get it to work, maybe I am doing something wrong?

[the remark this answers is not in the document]
gabydd: I think I was trying something like: ``text.chars().reversed().position(|ch| !ch.is_whitespace())`` which was always returning ``None`` and I had subtracting from the amount of chars as well.

gabydd:
Thank you for the review Pascal ❤️, especially the spliting lines part I thought that might be problematic, I'll try and get to this tonight

helix-core/src/comment.rs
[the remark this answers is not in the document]
pascalkuthe: Ah right ropey iterator are cursos not double ended like std so you need to do `chars_as(slice.len_chars())` instead of `chars`

helix-core/src/comment.rs
gabydd: I hope I addressed everything else but sadly I don't think we can use an iterator here because we need to check all the selections before we know if we are commenting or not. See how we clear the clear the vec if we find an uncommented selection so that we can normalize all the selections to all be commented.

helix-core/src/comment.rs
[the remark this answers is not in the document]
gabydd: that allowed me to get it to works thanks!

helix-core/src/comment.rs
```diff
@@ -94,6 +98,226 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            // extend each line
+            let (start_line, end_line) = range.line_range(text.slice(..));
+            let start = text.line_to_char(start_line);
+            let end = text.line_to_char((end_line + 1).min(text.len_lines()));
+
+            Range::new(start, end).with_direction(range.direction())
+        }),
+        &REGEX,
+    )
+}
```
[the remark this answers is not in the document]
gabydd: I believe I was able to implement this correctly, there is no lines iterator from a range, so I just took the a ``RopeSlice`` of the line range and iterated of those lines which worked

helix-core/src/comment.rs
gabydd: ah yes this pr was pretty much the first thing I wrote in rust so I just blindly copied this from the line commenting and modified it work for block commenting, looks like there should be a follow up pr to improve the line commenting implementation to take this into account as well.

pascalkuthe: That would be nice, also split selection at whitespace could be improved

helix-core/src/comment.rs
pascalkuthe: Ah yeah ofcourse I didn't think j about that my bad

helix-core/src/comment.rs
```diff
@@ -94,6 +96,212 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            find_first_non_whitespace_char(selection_slice),
+            find_last_non_whitespace_char(selection_slice),
+        ) {
+            let len = selection_slice.len_chars();
+            let open_len = open.chars().count();
+            let close_len = close.chars().count();
+            let after_open = open_pos + open_len;
+            let before_close = close_pos.saturating_sub(close_len);
+
+            let line_commented = if len >= after_open {
```
pascalkuthe, line 135:
```suggestion
            let line_commented = if len >= open_Len + close_len {
```

I think that should ge the right length check?

helix-core/src/comment.rs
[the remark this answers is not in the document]
pascalkuthe: yeah that's what I meant :+1:

helix-core/src/comment.rs
gabydd: Hmm yeah the current check works for all possibilities but the technically correct check I think would be: ``close_pos - open_pos >= open_len + close_len``

pascalkuthe: Ah I thaught that `len = close_pos -open_pos` but it I counts the trimmed white spaces still. Then that check makes more sense but I would prefer the variant I posted (and change the definition of lrn). It's more obvious what it does I think: Check if the slice is large enough to contain a block comment. The current condition could also panic for non-symmetric block comments where the open brace is shorter. HTML jas non-symetric block comments (altough there the open brave is longer) so it's not completely infeasiabke that we would auoprt such comments in the future (or that some user configures the editor that way)

gabydd: Change made👍

gabydd:
I think I will probably have to clean up the history of this pr because it is getting quite messy, I am thinking one commit for the actual commenting and another for all the ``languages.toml`` changes

helix-core/src/comment.rs
```diff
@@ -94,6 +96,212 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            find_block_comments(tokens.0, tokens.1, text, selection).0,
+        ),
+        (None, None) => (
+            find_block_comments("/*", "*/", text, &split_lines).0,
+            find_block_comments("/*", "*/", text, selection).0,
+        ),
+        (Some(_), None) => return CommentType::Line,
+        (None, Some(tokens)) => (
+            find_block_comments(tokens.0, tokens.1, text, &split_lines).0,
+            find_block_comments(tokens.0, tokens.1, text, selection).0,
```
pascalkuthe, line 289: you are essentially looking for block comments here and immedietly throwing away the result. I think there should we a second function for creating a transaction for block comment. Then you can have the following commands:
* toggle block comments (calls find_block_comment and create_block_comments_transaction)
* toggle comments: basically this function but instead of returning an enum immediately call `toggle_block_comments_as_line_fallback`, `toggle_line_comments` or `create_block_commen_transaction` (reusing the resunts of `find'block_comments` from inside this function)

gabydd: Ah that makes sense, I was a bit iffy about doing this as well and that works much better

gabydd: this is now implemented and I was also able to simplify the detection for the toggle_comments command, i then moved it into ``commands.rs`` I can move it back if you prefer though

pascalkuthe: i think it looks great lke this. Great split: core logging in helix-core glue code in the command implementation

pascalkuthe (approved):
From a technical point of view this LGTM now thanks!

>  or do we think that can be a separate option?

I see your point I think having the continue block comment token in the middle would indeed be nice. So `["/*" ,"  *" ,"/*"]`. On the other hand having it as a separate config works too and would avoid having to make a decision now/update all configs. I am also wondering if we somehow  want to support multiple comment tokens per language (so we can support doc comments for example, there are block doc comments in rust that would have a different continuation token). 

I am not quite sure about the config. If we can't settle on something at all now I would lean towards merging as is and then adding a separate config option for continuation later

pascalkuthe:
I discussed with @the-mikedavis how the config should look like to be future proof and we cameup with the following (example config for rust):

``` toml
comment-tokens = ["//", "///", "//!"]
block-comment-tokens = [
  { start = "/*", end = "*/" },
  { start = "/**", end = "*/" },
  { start = "/*!", end = "*/" },
]
```

So the core idea is that:
* we want to be able to uncomment multiple different kinds of comments so we want a list of token (I think for commenting we can always just use the first token for now. This is especially important for continuing comments correctly (which shouldn't be too hard to implement with this config)
* we want to be able to add somekin of config option to block comments in the future for how to continue comments. We don't really want to settle on that right now and using a dict makes it possible to add that token in the future without a breaking change. It also read a little nicer IMO

To stay backwards compatible (and make the config for block comments a little less cluttered since most languages just have a single block comment type) we want to be able to also just specify a string/a dict derctly instead of a single element list:

``` toml
comment-tokens =  "//"
block-comment-tokens = { start = "/*", end = "*/" }
```

This should also mean that you can very quickly update the languages.toml with multi cursors instead of having to go through it again by hand (altough if would be nice if you would adopt my rust example).

Implementing this kind of config option can be done with an untagged enum in serde we do this for quite a few config options so you can use that as reference

pascalkuthe:
ah sorry one thing I missed:

>  right now toggle a block doesn't extend the selection, should it do that?

Imo yes it should but not for the line comment fallback (since c-c doesn't change the selection either) but for the `<space>cb` binding it should

gabydd:
Both of that sound good to me. Just a quick clarification, this pr should also update the comment-tokens config aswell, right?

pascalkuthe:
> Both of that sound good to me. Just a quick clarification, this pr should also update the comment-tokens config aswell, right?

yeah ideally we would do it in this PR since we don't want a breaking change later

gabydd:
ok I have now implemented the new config, the way it works it will search each range to see if it is commented, if all ranges are commented it will uncomment them all, if atleast one is not commented that range will be block commented with either the first token in the array or the default tokens, next I need to resolve the conflicts with master, and try and extend the selection. the history should also probably be squashed at some point.

gabydd:
I found a way to do the extending when block commenting not sure if it is the best way just the easiest way I could figure out: [`2db9dea` (#4718)](https://github.com/helix-editor/helix/pull/4718/commits/2db9deaac8829af823161f24b05f4a550699c748), maybe if the selection should be extended should be passed as an argument to ``create_block_comment_transaction`` so the ranges are never returned?

gabydd:
ah one bug in the implementation of extending is it will panic when there is an only whitespace selection, I will have to fix that sometime in the morning though

book/src/languages.md
```diff
@@ -61,7 +61,8 @@ These configuration keys are available:
 | `roots`               | A set of marker files to look for when trying to find the workspace root. For example `Cargo.lock`, `yarn.lock` |
 | `auto-format`         | Whether to autoformat this language when saving               |
 | `diagnostic-severity` | Minimal severity of diagnostic for it to be displayed. (Allowed values: `Error`, `Warning`, `Info`, `Hint`) |
-| `comment-token`       | The token to use as a comment-token                           |
+| `comment-tokens`      | The tokens to use as a comment token, either a single token `"//"` or an array `["//", "///", "//!"]` (the first token will be used for commenting) |
```
pascalkuthe, line 64: While it doesn't sound as niceas the plural version  I would prefer we keep `comment-token` to avoid a breaking change.

pascalkuthe: Ah actually you can keep the change and use `serde(alias = 'comment-token')` yo keep legacy configs working

helix-core/src/comment.rs
```diff
@@ -94,6 +97,206 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+            find_first_non_whitespace_char(selection_slice),
+            find_last_non_whitespace_char(selection_slice),
+        ) {
+            let mut line_commented = false;
+            let mut after_start = 0;
+            let mut before_end = 0;
+
+            for BlockCommentToken { start, end } in &tokens {
+                let len = (end_pos + 1) - start_pos;
+                let start_len = start.chars().count();
```
pascalkuthe, line 151: Tiny not, move this outside if the loop.

The conouker likely dies that anyway but I personally thjbj it makes the code easier to read if inky the stuff that actually needs to be calculated in a loop is inside the loop

helix-core/src/comment.rs
```diff
@@ -94,6 +97,206 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
...
+        } else {
+            b.start.len().cmp(&a.start.len())
+        }
+    });
+    for range in selection {
+        let selection_slice = range.slice(text);
+        if let (Some(start_pos), Some(end_pos)) = (
+            find_first_non_whitespace_char(selection_slice),
+            find_last_non_whitespace_char(selection_slice),
+        ) {
```
pascalkuthe, line 144: Regarding your crash: You need to bail our here if `start_pos >= end_pos`

pascalkuthe (commented):
Thus looks very promising, just a couple smaller comnents but this LGTM for the most part 👍

helix-core/src/comment.rs
gabydd: moved len outside the loop :+1:

helix-core/src/comment.rs
gabydd: the crash was actually because I wasn't including the only whitespace ranges in the ``comment_changes`` vec so they weren't being added to the new selection that was constructed when commenting so there could be less ranges then the primary index of the selection. I changed the ``CommentChange`` struct to be an enum so all whitespace ranges could be included to be manually mapped

pascalkuthe (dismissed):

book/src/keymap.md
```diff
@@ -299,6 +301,16 @@ This layer is a kludge of mappings, mostly pickers.
...
 
+##### Comment mode
+
+Comment out code.
+
+| Key     | Description                          | Command               |
+| -----   | -----------                          | -------               |
+| `c`     | "Comment/uncomment selections"       | toggle_comments       |
+| `l`     | "Line comment/uncomment selections"  | toggle_line_comments  |
+| `b`     | "Block comment/uncomment selections" | toggle_block_comments |
```
the-mikedavis, line 312:
```suggestion
| `c`     | Comment/uncomment selections       | `toggle_comments`       |
| `l`     | Line comment/uncomment selections  | `toggle_line_comments`  |
| `b`     | Block comment/uncomment selections | `toggle_block_comments` |
```

for consistent formatting with the other keymap docs

languages.toml
the-mikedavis, line 1: To cut down on the size of the diff, what do you think about restoring changes where `comment-token` was just a single token? For example `sway`'s comment-tokens change could be reset

the-mikedavis (commented):
Just a few small comments from an initial read-through. I'll merge this to my driver branch and circle back for a proper review once I give it a try

languages.toml
gabydd: Yeah I can do that, that makes sense

helix-core/src/syntax.rs
```diff
@@ -231,6 +233,35 @@ impl<'de> Deserialize<'de> for FileType {
     }
 }
 
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(untagged, rename_all = "kebab-case", deny_unknown_fields)]
+pub enum CommentTokens {
+    Mutliple(Vec<String>),
+    Single(String),
+}
```
the-mikedavis, line 241: These untagged enums can feel a little clunky to work with. If we do a little more work during deserialization we can treat this type like a `Vec<String>` which is a nicer API when you want to iterate over the tokens or grab the first. We can also assert as deserialization-time that the Vec can't be empty:

```rust
#[derive(Debug, Serialize)]
pub struct CommentTokens(Vec<String>);

impl<'de> Deserialize<'de> for CommentTokens {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
        struct Visitor;

        impl<'de> serde::de::Visitor<'de> for Visitor {
            type Value = Vec<String>;

            fn expecting(&self, formatter: &mut fmt::Formatter) -> fmt::Result {
                write!(
                    formatter,
                    "a sequence of comment tokens or a single comment token"
                )
            }

            fn visit_str<E>(self, v: &str) -> Result<Self::Value, E>
            where
                E: serde::de::Error,
            {
                Ok(vec![v.to_string()])
            }

            fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
            where
                A: serde::de::SeqAccess<'de>,
            {
                let mut tokens = Vec::new();

                while let Some(token) = seq.next_element()? {
                    tokens.push(token);
                }

                if tokens.is_empty() {
                    Err(serde::de::Error::custom(
                        "`comment-tokens` must not be empty",
                    ))
                } else {
                    Ok(tokens)
                }
            }
        }

        deserializer.deserialize_any(Visitor).map(CommentTokens)
    }
}

impl ops::Deref for CommentTokens {
    type Target = Vec<String>;

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}
```

the-mikedavis: `BlockCommentTokens` can use this same pattern but instead of a `visit_str` it will want a `visit_map` that delegates to the MapAccess deserializer (which will automagically deserialize the BlockCommentToken struct)

```rust
fn visit_map<A>(self, map: A) -> Result<Self::Value, A::Error>
where
    A: MapAccess<'de>,
{
    let deserializer = serde::de::value::MapAccessDeserializer::new(map);
    Deserialize::deserialize(deserializer)
}
```

the-mikedavis: Ah also looks like there's a typo, mutliple => multiple :P

pascalkuthe: IMO its cleaner to implement the deserializer with an internal untagged enum:

``` rust 
impl<'de> Deserialize<'de> for CommentTokens {
    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
    where
        D: serde::Deserializer<'de>,
    {
		#[derive(Deserialize)]
		#[serde(untagged)]
	    enum CommentTokens {
		    Mutliple(Vec<String>),
		    Single(String),
		}
		match CommentTokens::deserialize(deserializer){
			Single(val) => Self(vec![val]),
			Multile(vals) => Self(vals),
		}
	}
}
```

Since we don't really need a cust type using `#[serde(from)]` is even cleaner:

``` rust
#[derive(Deserialize)]
#[serde(untagged)]
enum CommentTokens {
    Mutliple(Vec<String>),
    Single(String),
}
impl From<CommentTokens> for Vec<String> {
    fn from(value: CommentTokens) -> Self {
        match value {
            CommentTokens::Mutliple(vals) => vals,
            CommentTokens::Single(val) => vec![val],
        }
    }
}
```

then you can just have a smile vec field

``` rust
#[serde(from = "CommentTokens"))]
comment_tokens: Vec<String>
```

gabydd: turns out ``from`` can only be used for container attributes https://serde.rs/container-attrs.html so I used ``deserialize_with`` instead

pascalkuthe: Ah right I gorogt about that. I actually almost prefer this tough since it makes it cleat that the enun is just a deserlaize helper 👍

helix-term/src/commands.rs
```diff
@@ -4533,18 +4535,122 @@ pub fn completion(cx: &mut Context) {
...
     doc.apply(&transaction, view.id);
     exit_select_mode(cx);
 }
 
+fn toggle_comments(cx: &mut Context) {
+    toggle_comments_impl(cx, |token, tokens, doc, selection| {
+        let text = doc.slice(..);
+
+        // only have line comment tokens
+        if matches!((token, tokens.clone()), (Some(_), None)) {
```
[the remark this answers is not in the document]
the-mikedavis: This can also be used for the other `matches!` in this file

helix-term/src/commands.rs
```diff
@@ -4533,18 +4535,122 @@ pub fn completion(cx: &mut Context) {
 }
 
 // comments
-fn toggle_comments(cx: &mut Context) {
+type CommentTransactionFn = fn(
+    token: Option<&str>,
+    tokens: Option<Vec<BlockCommentToken>>,
```
the-mikedavis, line 4540: I think it might be clearer to call these `line_token` and `block_tokens`

helix-term/src/commands.rs
```diff
@@ -4533,18 +4535,122 @@ pub fn completion(cx: &mut Context) {
...
+        if matches!((token, tokens.clone()), (Some(_), None)) {
+            return comment::toggle_line_comments(doc, selection, token);
+        }
+
+        let split_lines = comment::split_lines_of_selection(text, selection);
+
+        let (line_commented, line_comment_changes) = comment::find_block_comments(
+            tokens
+                .clone()
+                .unwrap_or_else(|| vec![BlockCommentToken::default()]),
```
the-mikedavis, line 4577: Here and in the `comment` module we can remove most of these clones by passing around slices of the Vec, so `&[BlockCommentToken]` rather than `Vec<BlockCommentToken>`. `comment::find_block_comments` will need to switch the line `let mut tokens = tokens` to `let mut tokens = tokens.to_vec()` which will allocate a new vec but we can get away with not allocating the vec otherwise.

Instead of `unwrap_or_else` in this file we will need:

```rust
let default_block_comment_tokens = &[BlockCommentToken::default()];
let block_comment_tokens = tokens.unwrap_or(default_block_comment_tokens);
```

so that the slice lives long enough

the-mikedavis (commented):
Behavior-wise this is looking good to me. Just a few small nits and a suggestion about vecs vs slices

the-mikedavis (dismissed):

helix-term/src/keymap/default.rs
```diff
@@ -274,6 +274,11 @@ pub fn default() -> HashMap<Mode, KeyTrie> {
             "k" => hover,
             "r" => rename_symbol,
             "h" => select_references_to_symbol_under_cursor,
+            "c" => { "Toggle comments"
+                "c" => toggle_comments,
+                "l" => toggle_line_comments,
+                "b" => toggle_block_comments,
+            },
```
archseer, line 281: How about using `space c` for toggle, `space C` for block, then `space alt-C` for block? That way we make it a simple press for the default usecase, and the other ones can be accessed with modifiers

helix-term/src/commands.rs
```diff
@@ -4533,18 +4535,118 @@ pub fn completion(cx: &mut Context) {
...
+        .map(|tc| &tc[..]);
+
+    let transaction =
+        comment_transaction(line_token, block_tokens, doc.text(), doc.selection(view.id));
 
     doc.apply(&transaction, view.id);
     exit_select_mode(cx);
 }
 
+fn toggle_comments(cx: &mut Context) {
```
archseer, line 4564: This needs some documentation on how it works. If I'm commenting a fresh set of lines, does it use block or line comments? I'd prefer it to use line comments if it can, because it makes it easier to partially uncomment

archseer: Also, let's say that I'm trying to uncomment two lines with line comments, I could start my selection in the middle of the first line, then end in the middle of the second line and both lines would uncomment (in other words I don't have to do a perfect selection). Is that behaviour retained?

helix-term/src/keymap/default.rs
gabydd: that seems good could you clarify which one you would like to be line comment?

helix-term/src/commands.rs
gabydd: ah yes thats a good idea, I'll do that now, but the idea of this is it doesn't change anything at all if there is a line comment, if the whole selection is surrounded in block comment then it will uncomment that region, if the whole selection is line commented the uncomment that, if there is no comments or not all lines are commented with block comments and there is only block comment tokens it will comment each non commented line individually with those tokens, if each line is commented with line comments then it will un-comment all the lines.

gabydd: this is how I just commented it:
```rust
/// commenting behavior:
/// 1. only line comment tokens -> line comment
/// 2. each line block commented -> uncomment all lines
/// 3. whole selection block commented -> uncomment selection
/// 4. all lines not commented and block tokens -> comment uncommented lines
/// 5. no comment tokens and not block commented -> line comment
```

helix-term/src/keymap/default.rs
gabydd: I changed it to being alt-c for line comment, no worries if you want it the other way but I imagine more people would use block comment binding then the line comment one.

nkitsaini:
Just curious if/how this can handle the case where same file type requires different types of comment characters. For example `html` files can require `//` for `<script>`, `/* */` for `<style>` and `<!-- -->` for html. Currently I could not find a way to manually map keybindings to specific comment characters.

pascalkuthe:
that's out of scope for this PR, HTML is really weird in that regard, the best way would likely to handle that using injections in the future

the-mikedavis:
That's https://github.com/helix-editor/helix/issues/7364

gabydd:
Just fixed the merge conflicts I also squashed the pr to one commit so it's easier to rebase when there are conflicts if we want I can make one commit the code changes and the other the languages.toml changes

pascalkuthe:
(Also closes #4009 but I can't link the issue because the GH UI is terribble)

gabydd:
This only fixes that for block comments, I didn't want to change the functionality of line comments in this pr

gabydd:
What we could do is for languages without a block comment token, when using the toggle_block_comments it would only add or remove from the start of the selection as if you had a block comment that was the line comment on the left and an empty string on the right

gabydd:
redundant_clone seems to have some false positives: https://github.com/rust-lang/rust-clippy/issues/10577

the-mikedavis:
Let's add some `#[allow(..)]`s arround those for now with a reminder to check if we can remove them when the MSRV is bumped. Looks like the issue isn't solved yet so we might have to keep the `allow`s around for a little while

daedroza:
Using this since last week and works great, thanks for putting this up!

the-mikedavis (dismissed):

pascalkuthe (dismissed):

the-mikedavis (approved):

David-Else:
Awesome merge!

> How will this affect the default `"C-c" => toggle_comments` binding?

It still seems to live on like a Zombie refusing to die, can we bury it now?

Issue #1188, opened 2021-11-29
Make gutters more configurable


dannasessha:
Following up on #783.
 
Gutters are potentially useful for a variety of purposes, we should make them more configurable.

from [previous comments](https://github.com/helix-editor/helix/pull/783#issuecomment-977903773) : "...if a user wanted to use the column in the editor to have a `git diff` type display ala the vim git gutter plugin, or a `git annotate` type feature to familiarize themselves with the history of a file from a certain perspective, or even possibly being able to manipulate the column directly for some kind of bespoke user-annotation that could be made on the fly."

[and](https://github.com/helix-editor/helix/pull/783#issuecomment-978737241) : "...these types of gutter components give us more flexibility to allow for configuring the gutter (like disabling line numbers) or using different gutters per document (for example getting rid of diagnostics on docs with no LSP) as well as defining new types of gutters in the future (i.e. git gutter)"

archseer:
I was thinking the config could be a `gutters = ["diagnostic", "line_number"]` type array.

archseer:
Then you could have a `gutter.diagnostic` section for any additional config of each component

dannasessha:
Okay, I'll see if I can get something going, thank you for the feedback!

wurosh:
+1 for disabling line numbers as a feature (Or is this possible now? - I wasn't able to find the config option). In an editor which supports jumping to lines this easily they're kind of a waste of space (unless you're screen sharing and talking about the code)

the-mikedavis:
#1967 addressed the configuration part of this.

David-Kunz:
What's still missing in https://github.com/helix-editor/helix/pull/1967 is configuring the features of each gutter (what e.g. `gutter.diagnostic` would allow).

pascalkuthe:
Gutters a re pretty flexible now, this issue is old and too general so I am closing it. If you are missing some specific gutter config open a feature request for that

PR #4724, opened 2022-11-12
Add configuration for min width of line-numbers gutter

dgkf:
Supercedes https://github.com/helix-editor/helix/pull/3953

Now that the line-number gutter can be resized (https://github.com/helix-editor/helix/pull/3469), this exposes options for the currently hard-coded minimum  number of characters used in the line-numbers gutter.

# Architectural Changes

- `GutterConfig` & `GutterLineNumbersConfig` config structs added, allowing for the configuration style suggested in [#1188](https://github.com/helix-editor/helix/issues/1188#issuecomment-982138408).
- `View` now accepts a `gutters` parameter as a `GutterConfig` object instead of `Vec<GutterType>`
- `Editor::refresh_config()` now also refreshes `View`s to respond to configuration update events

# Design feedback needed

- [ ] Should `editor.gutters` provided as an array of strings be deprecated in favor of `editor.gutters.layout`?


helix-view/src/editor.rs
```diff
@@ -130,6 +130,10 @@ pub struct Config {
     pub cursorcolumn: bool,
     /// Gutters. Default ["diagnostics", "line-numbers"]
     pub gutters: Vec<GutterType>,
+    /// Minimum number of characters to display as part of line number gutter. Defaults to 2.
+    pub line_numbers_width_min: usize,
+    /// Maximum number of characters to display as part of line number gutter, or 0 for no maximum. Defaults to 0 (no maximum).
+    pub line_numbers_width_max: usize,
```
dgkf, line 136: Currently uses options `line-numbers-width-min` and `line-numbers-width-max`. There was a proposal [in this comment](https://github.com/helix-editor/helix/issues/1188#issuecomment-982138408) to add `gutter.<gutter name>` sections to the config file so that these options could be stored as:

```toml
[gutter.line-numbers]
width-min = 1
width-max = 4
```

@archseer How would you like these options exposed? Would you like to propose a config syntax that I should aim for?

helix-view/src/editor.rs
```diff
@@ -595,6 +599,8 @@ impl Default for Config {
                 GutterType::Spacer,
                 GutterType::LineNumbers,
             ],
+            line_numbers_width_min: 2,
+            line_numbers_width_max: 0, // unlimited
```
dgkf, line 603: Currently, setting the maximum to `0` is equivalent to "unlimited". I'm mixed about this decision. I don't particularly like the idea of hard-coding a cap, but I also don't like using `0` as a "magic number". Open to suggestions.

dgkf, line 603: Preferred sensible defaults?

helix-view/src/view.rs
```diff
@@ -109,7 +112,12 @@ impl fmt::Debug for View {
 }
 
 impl View {
-    pub fn new(doc: DocumentId, gutter_types: Vec<crate::editor::GutterType>) -> Self {
+    pub fn new(
+        doc: DocumentId,
+        gutter_types: Vec<crate::editor::GutterType>,
+        gutter_line_numbers_width_min: usize,
+        gutter_line_numbers_width_max: usize,
```
dgkf, line 119: I've added parameters for the min and max line width, but I think this will get messy quick as more configuration options are added. Instead, think it would make more sense to pass a `Config` similar to how `Editor` accepts a `Config`. Looking for feedback before making changes.

the-mikedavis:
I wonder if this is necessary after #3469? You would have to be editing a really massive file for the gutter to be noticeably large

archseer:
Agree, we can offer a configurable minimum but no maximum.

dgkf:
@the-mikedavis - I can only speak for myself, but I really try to cut down on the unused space in the gutters. In `vim` I set `numberwidth=1` and `signcolumn='number'`, which usually means I get away with only 3 columns for both diagnostics and line numbers. Especially when working on a small display, saving a collective 6-10 columns can make the difference with a vertical split or two.

archseer:
I think a min of `2` was deliberate in the previous PR:

> The line number gutter will have a minimum width of 2 characters. This is so that configurations using line-number: "relative" aren't adversely affected by this change. In the future, the min/max width could perhaps be configurable and this hard-coded minimum could be removed.

archseer:
In most cases the line number column is going to be under 5-6 columns (10_000 - 100_000 lines), and any larger than that would be a pain to load fully into memory anyway.

helix-view/src/editor.rs
```diff
@@ -67,6 +67,96 @@ where
...
+
+        // Create a mirror of a GutterConfig to let serde derive our
+        // deserialization for us.
+        // This is a hack!! Please show me a better way!!
+        #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+        #[serde(rename_all = "kebab-case", default, deny_unknown_fields)]
+        struct GutterConfigStub {
+            layout: Vec<GutterType>,
+            line_numbers: GutterLineNumbersConfig,
+        }
```
dgkf, line 114: I'm not familiar with best practices for custom deserializers. This feels really sloppy and prone to issues with mismatches between types. Is there a better way?

helix-view/src/editor.rs
```diff
@@ -67,6 +67,96 @@ where
...
+        D: Deserializer<'de>,
+    {
+        // Accept either a simple array of strings as gutter defaults or
+        // a [editor.gutters] block
+        #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+        #[serde(untagged)]
+        enum GutterConfigFormat {
+            Simple(Vec<GutterType>),
+            Detailed(GutterConfigStub),
+        }
```
dgkf, line 104: Custom deserializer added for `GutterConfig` so that it can accept the current `editor.gutters` vector of gutter names, while opening up more structured gutter config settings.

Currently, this can handle either

```toml
[editor]
gutters = ["diagnostics", "line-numbers"]
```

as well as

```toml
[editor.gutters]
layout = ["diagnostics", "line-numbers"]

[editor.gutters.line-numbers]
width = 3
```

dgkf:
> Agree, we can offer a configurable minimum but no maximum.

@archseer @the-mikedavis I've altered the implementation so that only minimum linenumber gutter width is configurable. This is ready for review. I'm happy with how it's come along, with the exception of the `GutterConfig` deserializer implementation. I'd appreciate a nudge in the right direction on how that should be handled more thoughtfully.

helix-view/src/editor.rs
dgkf: This has now been reworked and I'm much happier with how it ended up. Now the config is provided in one of two forms which has its own deserializer. When passed to a `View` it is coerced into the consistent verbose form.

helix-view/src/gutter.rs
```diff
@@ -181,20 +189,26 @@ pub fn line_numbers<'doc>(
...
         }
     })
 }
 
-pub fn line_numbers_width(_view: &View, doc: &Document) -> usize {
+/// The width of a "line-numbers" gutter
+///
+/// The width of the gutter depends on the number of lines in the document,
+/// whether there is content on the last line (the `~` line), and the `line-
+/// numbers-width-min` & `line-numbers-width-max` settings.
```
the-mikedavis, line 203: Looks like these are out of date config options

helix-view/src/editor.rs
```diff
@@ -132,8 +196,8 @@ pub struct Config {
     pub cursorline: bool,
     /// Highlight the columns cursors are currently on. Defaults to false.
     pub cursorcolumn: bool,
-    /// Gutters. Default ["diagnostics", "line-numbers"]
-    pub gutters: Vec<GutterType>,
+    /// Gutters. Default { layout: ["diagnostics", "spacer", "line-numbers"] }
```
the-mikedavis, line 199: I think we can just remove the docs here on the default since the type is getting more complicated

helix-view/src/editor.rs
```diff
@@ -71,6 +71,70 @@ where
...
 }
 
+// Accept either a simple array of strings as gutter defaults or
+// a [editor.gutters] block
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum GutterConfigOption {
+    Simple(Vec<GutterType>),
+    Detailed(GutterConfig),
+}
```
the-mikedavis, line 81: Rather than juggling two types, you could use a serde visitor to parse into just the `GutterConfig` type: https://github.com/helix-editor/helix/pull/3798#issuecomment-1242947382

Also see where we use this pattern in `keymap.rs` https://github.com/helix-editor/helix/blob/fa436fa680b5bcf6d46666a03ae23b94a74e0414/helix-term/src/keymap.rs#L154-L209 (although it's used there for better error messages https://github.com/helix-editor/helix/pull/3931)

book/src/configuration.md
```diff
@@ -255,3 +255,55 @@ render = true
...
+layout = ["diff", "diagnostics", "line-numbers", "spacer"]
+```
+
+#### `[editor.gutters.line-numbers]` Section
+
+Options for the line number gutter
+
+| Key     | Description                             | Default |
+| ---     | ---                                     | ---     |
+| `width` | The minimum number of characters to use | `3`     |
```
the-mikedavis, line 290: Though we aren't providing a max-width, I think it makes sense to call this `min-width`. I would naively assume that a `width` option controls the min and max.

the-mikedavis (commented):
This looks like a good way to configure gutters to me 👍. We will probably end up using this in the future to allow configuration of the characters used for additions/modifications/deletions in the diff gutter element and maybe the separator character (as with the statusline, although I can't imagine anything other than space looking good)

helix-view/src/editor.rs
dgkf: I had a [`Deserialize` implementation](https://github.com/helix-editor/helix/blob/ea094ed143d6d23d34d1d667650fcf57025c48b6/helix-view/src/editor.rs#L92-L145) previously (not as elegant as using a Visitor, but functionally similar), but wasn't thrilled with how it came out. I'm still grappling with the right way to do this. 

With both the custom `Deserialize` and `Visitor` patterns, I can't find a way to re-use the derived `Deserialize` - which is still what I want to use for deserializing maps, with the added custom `visit_seq`. Is it possible to still derive `visit_map` (or equivalent) while implementing a custom `visit_seq`? It _feels_ like this should be possible, but I'm not digging anything up while reading through the `serde` docs.

I'll continue to hunt around for an elegant solution here, but any nudges in the right direction would be appreciated.

pascalkuthe: The `deserialize_with` option in serdw is intended for these kinds of situations

dgkf: Cheers - thanks @pascalkuthe. I'll look into that.

dgkf: Thanks both for the pointers. I'm much happier with where it's at now.

book/src/configuration.md
dgkf: I like that suggestion. It's been updated now to use `min-width` instead.

helix-view/src/editor.rs
dgkf: This has been cleaned up following suggestions to use the `#[serde(deserialize_with)]` attribute and a Visitor implementation. I no longer have concerns about the implementation.

dgkf:
@the-mikedavis @pascalkuthe - Thank you both for reviewing and for the suggestions on how to better handle the different styles of configuration. I've incorporated your feedback. This PR is now ready for another pass of review.

book/src/configuration.md
```diff
@@ -255,3 +255,55 @@ render = true
...
+
+Currently unused
+
+#### `[editor.gutters.diff]` Section
+
+Currently unused
+
+#### `[editor.gutters.spacer]` Section
+
+Currently unused
```
phaazon, line 309: What about `[editor.gutters.breakpoints]`? I’m mentioning this because of what I’m doing in #5371 (I might eventually rebase on your work).

dgkf: Right now the sub-sections are styled after the names that can be provided to `editor.gutters`. There's no real reason it _must_ be like this, but it think there's some discoverability benefit to being consistent here. 

Following that style, both breakpoints and diagnostics would be configured in `editors.gutters.diagnostics`. I think it's best to follow this precedent, but I'd say this is a call for some of the more core contributors like @the-mikedavis 

--- 

Looking into the future a bit further, I think the underlying issue here is that diagnostics and breakpoints are a bit oddly coupled - likely a solution from before gutters were configurable. I think there's a case for decoupling diagnostics and breakpoints and trying to tackle that use case more generically. I think it would be interesting to be able to layer gutters as part of the layout. Perhaps something like:

```toml
[editor.gutters]
layout = [ [ "diagnostics", "breakpoints" ], "line-numbers" ]
```

Which would provide the current behavior. I think that needs more discussion before jumping to something like that, though.

phaazon: I’m reimplementing #5371 rebased atop your PR then.

helix-view/src/editor.rs
```diff
@@ -71,6 +72,105 @@ where
...
+where
+    T: Deserialize<'de> + From<Vec<GutterType>>,
+    D: Deserializer<'de>,
+{
+    // This is a Visitor that forwards string types to T's `FromStr` impl and
+    // forwards map types to T's `Deserialize` impl. The `PhantomData` is to
+    // keep the compiler from complaining about T being an unused generic type
+    // parameter. We need T in order to know the Value type for the Visitor
+    // impl.
+    struct GutterSeqOrStruct<T>(PhantomData<fn() -> T>);
```
phaazon, line 118: You can probably use `*const T`, which is also covariant in `T`.

dgkf: Admittedly I'm just getting exposed to this now - but I appreciate the terminology which prompted a nice reading on variance in the nomicon. 

In this case, this type signature (and the accompanying comment) is lifted straight from the example provided in the [serde docs](https://serde.rs/string-or-struct.html). If there are merits to one or the other, it's not obvious to me.

`fn() -> *const T` isn't covered in [nomicon's page on PhantomData patterns](https://doc.rust-lang.org/nomicon/phantom-data.html#table-of-phantomdata-patterns). In contrast to `PhantomData<T>` vs `PhantomData<*const T>` recommended [here](https://doc.rust-lang.org/core/marker/struct.PhantomData.html#ownership-and-the-drop-check), I'm not seeing similar discussion regarding `fn() -> T` vs `fn() -> *const T`. 

I'm happy to make the change, but honestly don't know enough about the implications of each. If they provide the same variance, does using `fn() -> *const T` imbue some other benefit over `fn() -> T`?

phaazon: The implication is the same, since both `*const T` and `fn() -> T` are covariant in `T` (`fn(T)` is contravariant in `T` though). Covariance and contravariance can be a bit overwhelming are basically used to know how subtyping can be done. IIRC rustc uses it to implement coercion rules (and maybe drop but I’m not sure?).

Either way I think it should be okay, I was just mentioning it.

dgkf: Okay cool - then I'd default to whatever is recommended in the serde docs. Let's leave it as is and if there's ever a strong rationale to switch that can always come down the road in another PR.

dgkf:
@the-mikedavis, @archseer - Any chance I can get your eyes on this PR for review? It's beginning to be used by downstream work, so I'm feeling obliged to push it forward.

phaazon:
Agreed, it would be nice to be able to merge it. Rebasing my PR (#5371) requires rebasing that one so I can’t really be up-to-date with `master` besides locally rebasing this one. Is there any missing points?

helix-view/src/editor.rs
```diff
@@ -71,6 +72,105 @@ where
...
+        where
+            M: serde::de::MapAccess<'de>,
+        {
+            let deserializer = serde::de::value::MapAccessDeserializer::new(map);
+            Deserialize::deserialize(deserializer)
+        }
+    }
+
+    deserializer.deserialize_any(GutterSeqOrStruct(PhantomData))
+}
```
the-mikedavis, line 159: The `T` generic can be eliminated:

```suggestion
fn deserialize_gutter_seq_or_struct<'de, D>(deserializer: D) -> Result<GutterConfig, D::Error>
where
    D: Deserializer<'de>,
{
    struct GutterVisitor;

    impl<'de> serde::de::Visitor<'de> for GutterVisitor {
        type Value = GutterConfig;

        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
            write!(
                formatter,
                "an array of gutter names or a detailed gutter configuration"
            )
        }

        fn visit_seq<S>(self, mut seq: S) -> Result<Self::Value, S::Error>
        where
            S: serde::de::SeqAccess<'de>,
        {
            let mut gutters = Vec::new();
            while let Some(gutter) = seq.next_element::<&str>()? {
                gutters.push(
                    gutter
                        .parse::<GutterType>()
                        .map_err(serde::de::Error::custom)?,
                )
            }

            Ok(gutters.into())
        }

        fn visit_map<M>(self, map: M) -> Result<Self::Value, M::Error>
        where
            M: serde::de::MapAccess<'de>,
        {
            let deserializer = serde::de::value::MapAccessDeserializer::new(map);
            Deserialize::deserialize(deserializer)
        }
    }

    deserializer.deserialize_any(GutterVisitor)
}
```

helix-view/src/gutter.rs
```diff
@@ -158,7 +165,8 @@ pub fn line_numbers<'doc>(
     let mode = editor.mode;
 
     Box::new(move |line: usize, selected: bool, out: &mut String| {
-        if line == last_line && !draw_last {
+        if line == last_line_in_view && !draw_last {
+            let width = max(n_last_line, n_min);
```
the-mikedavis, line 169: Looks like this is computed and used in both branches so it could probably be calculated above with `n_last_line` and `n_min`. You can also use `x.min(y)` and avoid the change to use `std::cmp::max`

helix-view/src/editor.rs
```diff
@@ -981,6 +1076,7 @@ impl Editor {
         for (view, _) in self.tree.views_mut() {
             let doc = doc_mut!(self, &view.doc);
             view.sync_changes(doc);
+            view.gutters = config.gutters.clone();
```
the-mikedavis, line 1079: Does this (and the L938) change also fix https://github.com/helix-editor/helix/issues/3445?

helix-view/src/gutter.rs
```diff
@@ -181,20 +189,26 @@ pub fn line_numbers<'doc>(
...
     })
 }
 
-pub fn line_numbers_width(_view: &View, doc: &Document) -> usize {
+/// The width of a "line-numbers" gutter
+///
+/// The width of the gutter depends on the number of lines in the document,
+/// whether there is content on the last line (the `~` line), and the
+/// `editor.gutters.line-numbers.min-width` settings.
+pub fn line_numbers_width(view: &View, doc: &Document) -> usize {
```
the-mikedavis, line 204: It looks like this might be unused. Could it be used above? Also we can probably drop the `pub` if it won't be used elsewhere

runtime/queries/r/locals.scm
```diff
@@ -3,6 +3,7 @@
 (function_definition) @local.scope
 
 (formal_parameters (identifier) @local.definition)
+(formal_parameters (default_parameter (identifier) @local.definition))
```
the-mikedavis, line 6: looks like an unrelated change snuck in here

the-mikedavis (commented):
Looking good overall, just some minor comments

helix-view/src/editor.rs
dgkf: Yes, this will also address reloading of gutter-related configs. Thanks for tagging that. I had forgotten I did this :sweat_smile:.

I'll address the rest of the changes when I next have a bit of time - most likely by the end of the day.

runtime/queries/r/locals.scm
dgkf: Thanks for spotting! Change has been reverted. It's still in the commit history. Let me know if you'd prefer that I force push changes to the commit history or if squashing on merge is fine.

helix-view/src/gutter.rs
dgkf: > It looks like this might be unused.

This is used above in `GutterType::width()`

https://github.com/helix-editor/helix/blob/097fd4f32298b5327490691dea3e8b880890d84b/helix-view/src/gutter.rs#L38-L45

> Could it be used above?

I'm not sure I follow. Would you prefer that it's inlined into `GutterType::width()`?

dgkf:
Thank you for the review today @the-mikedavis - I applied all but one change that I wasn't quite clear on.

helix-view/src/gutter.rs
the-mikedavis: Ah I see, I missed that reference.

I think that the `line_numbers_width` function could be reused in the `line_numbers` above it as well though, right?

dgkf: Yeah, that would work. At one point more of the intermediate values were needed in `line_numbers`, but I think `draw_last` is the only thing that is needed in both places.

I think it's worth the minor redundancy to keep the majority of the logic in one place.

I'll make the change so that we reuse `line_numbers_width`

phaazon:
Small nit question: have you rebased on `master`? If so, even though your PR is not merged yet, I’ll rebase #5371 on yours to follow your changes.

dgkf:
@phaazon - I rebased onto master so hopefully it's easier on you now if you want to use this as the basis for your changes.

runtime/queries/r/locals.scm
the-mikedavis: If you like to force-push and constantly keep a clean-looking branch, I don't mind the force-pushes. The squash and merge at the end is also very easy so it's up to you really

the-mikedavis (approved):

runtime/queries/r/locals.scm
dgkf: I usually squash and merge - it feels safer to me not to dabble in force territory.

PR #4729, opened 2022-11-13
Add a maximum file size for highlighting with an editor setting

xEgoist:
### Problem

syntax highlighting can take so much resources to fire up especially with large files. Trying to open a very large file (100 mb file for example) will takes very long time for helix and will be very slow to navigate through.

 
##
This PR adds a language specific settings for the users to specify their desired maximum file size in bytes.

for example, 
```toml
[[language]]
name = "c"
hl-max-size = 1000000
```
will stop the highlighting for any C files larger than 1mb. 

Another option is making this a global settings, but I feel like making it a language specific provides more flexibility.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -57,6 +57,7 @@ on unix operating systems.
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
+| `syntax-max-file-size` | Specify the maximum file size in bytes to enable syntax highlighting | `157286400` |
 
 ### `[editor.statusline]` Section
 
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -174,6 +174,8 @@ pub struct Config {
     pub indent_guides: IndentGuidesConfig,
     /// Whether to color modes with different colors. Defaults to `false`.
     pub color_modes: bool,
+    /// Maximum file size to trigger syntax highlighting
+    pub syntax_max_file_size: usize,
 }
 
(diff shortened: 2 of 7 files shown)


archseer:
This should be global (under editor config) since all languages will degrade in parsing performance as size increases. And `hl-max-size` is not very descriptive.

pascalkuthe:
I also think this should be a global option and we should also consider setting a default for this. While I really want helix to perform well for large files too, I think there is a ceiling where its just not realistic anymore. That default should probably be MUCH higher then 1MB tough. With #4716 and another PR I am working on locally right now I can still edit the largest file in the linux kernel (24mb) with treesitter enabled with only a small amount of lag and have no noticeable lag during navigation. A good default value might be 150mb to 200mb. A nice QOL feature would be displaying a warning to the user when such a large file is opened like "This file exceed the highlight-limit, treesitter related features like syntax highlighting were automatically disabled."

kirawi:
I think memory usage is a concern though.

archseer:
150~200MB is a reasonable default, that's what I was thinking of using.

helix-view/src/editor.rs
```diff
@@ -1098,8 +1101,12 @@ impl Editor {
         let id = if let Some(id) = id {
             id
         } else {
-            let mut doc = Document::open(&path, None, Some(self.syn_loader.clone()))?;
```
xEgoist, line 1101: Note: it's possible to open the file and get the metadata from it and pass `None` to the function, stopping language config all together. I personally think it's not a good idea to do so.

book/src/configuration.md
```diff
@@ -57,6 +57,7 @@ on unix operating systems.
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file. | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
+| `syntax-max-file-size` | Specify the maximum file size in bytes to enable syntax highlighting | `150000000` |
```
xEgoist, line 60: Note: Better setting name?

PR #4731, opened 2022-11-13
Split modification indicator from file statusline elements

### LINKED ISSUE #4686: It should be possible to tell there are unsaved changes to the buffer, even with long file paths.
If the file path for the current buffer is longer than will fit on the allotted space in the status line, it is truncated to fit. This makes perfect sense. However, the " [+]" that indicates the current buffer *has unsaved changes* is also lost in this truncation, making it impossible to tell whether the buffer requires saving *until you try to close it or fully quit out*.

This is particularly problematic for users who tend to open new tmux windows/panes for editing and exit out by simply killing said window/pane, in which case *you don't see there are unsaved changes and Helix doesn't get a chance to warn you about the unsaved buffer at all*.

The status line string truncation should take into account whether a " [+]" (or any other status indicator) is present, and truncate *the buffer name alone*, allowing space for the indicator to still be shown on-screen.

What I'm seeing:
- no changes:
  ` NOR  a_very_long_directory_name/some_other_directory/my_file_with_a_very_long_ 13:14  ruby `
- with changes:
  ` NOR  a_very_long_directory_name/some_other_directory/my_file_with_a_very_long_ 13:14  ruby `

What I would expect:
- no changes:
  ` NOR  a_very_long_directory_name/some_other_directory/my_file_with_a_very_long_ 13:14  ruby `
- with changes:
  ` NOR  a_very_long_directory_name/some_other_directory/my_file_with_a_very_l [+] 13:14  ruby `

Additionally, it would be nice if truncation was indicated with a `…` as opposed to the buffer name simply terminating weirdly, though I realize that's more of a nitpick than the legitimate issue above.

---

helix: 22.08.1

(Wasn't quite sure if this should be reported as a bug or as an enhancement. Please do categorize accordingly if this should've been a bug report instead.)

cbclemmer:
attempts to close #4686 

Adds a new "file-dirty" status line element that can replace the `[+]` symbol that appears after the file name.
This allows the element to be added before the "file-name" element so the dirty symbol will be visible for long file/directory names.
Adds logic to keep the dirty element after the "file-name" element if it is not in the list. This keeps current config files and the default configuration working as they were before while allowing the dirty element to be placed wherever the config file specifies.


the-mikedavis:
Let's call this "FileModificationIndicator"

cbclemmer:
@the-mikedavis Ok, I found a reference to dirty somewhere so that's why I called it that.

cbclemmer:
@the-mikedavis I changed it along with the other related variables.

helix-term/src/ui/statusline.rs
```diff
@@ -354,16 +370,21 @@ where
...
-            path,
-            if context.doc.is_modified() { "[+]" } else { "" }
-        )
+        format!(" {} ", path)
     };
 
     write(context, title, None);
 }
 
+fn render_file_dirty<F>(context: &mut RenderContext, write: F)
```
the-mikedavis, line 379: also should be renamed

helix-term/src/ui/statusline.rs
```diff
@@ -66,10 +66,27 @@ pub fn render(context: &mut RenderContext, viewport: Rect, surface: &mut Surface
...
+        let file_name_index = element_ids
+            .iter()
+            .position(|id| id == &helix_view::editor::StatusLineElement::FileName);
+        if !has_modified_element && file_name_index != None {
+            ret_val.insert(file_name_index.unwrap() + 1, *file_modified);
+        }
+        return ret_val;
+    };
+
     // Left side of the status line.
```
the-mikedavis, line 87: Determining if the statusline is missing the element is not necessary: the new element can be added to the `Default` implementation for `StatusLineConfig` for `left` (in `helix-view/src/editor.rs`) next to `FileName` to emulate the old behavior. It should be possible (although not advisable) to exclude this element in a statusline configuration

cbclemmer: Ok, I can do that, I was just trying to not break current custom configs.

cbclemmer: @the-mikedavis I've reverted my changes to the render function on statusline.rs and added the modification indicator to the default config.

helix-term/src/ui/statusline.rs
cbclemmer: Changed

helix-term/src/ui/statusline.rs
```diff
@@ -354,16 +355,21 @@ where
...
     };
 
     write(context, title, None);
 }
 
+fn render_file_modified<F>(context: &mut RenderContext, write: F)
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let title = format!(" {} ", if context.doc.is_modified() { "[+]" } else { "" });
```
the-mikedavis, line 368: The surrounding space ends up putting this element pretty far apart from the file name. I think it's ok to only write `"[+]"` here without the surrounding space given the element's intended usage

the-mikedavis: I think the `else` branch should be three spaces: that would allow this element to be used before the FileName element without causing a jump when the file becomes modified

cbclemmer: Changed

cbclemmer:
@kirawi Is there anything you are waiting for me on this PR?

kirawi:
@the-mikedavis I feel like this is a fairly uncontroversial change.

cbclemmer:
@kirawi @the-mikedavis The tests are passing now. I didn't see that before.

the-mikedavis (approved):

theteachr:
I believe this one needs an update in the configuration docs.

the-mikedavis:
Ah good catch, a new line for this should be added in this table https://github.com/helix-editor/helix/blob/0a7c697dd785111626c8f47a36e07381f55c2e83/book/src/configuration.md#L93

theteachr:
Create https://github.com/helix-editor/helix/pull/6036 to update the docs.

PR #4734, opened 2022-11-13
Protect against overwriting external changes

kov:
When saving, check that the file has not changed before our last known save before writing the file. Only overwrite external changes if asked to force.

I've almost lost a bunch of work because I did it on a separate instance of helix in another terminal window, but had the same file opened on an earlier instance and saved the file without realizing. Thankfully I had a backup (a cat in another tab xD)

### DIFF
--- a/helix-term/tests/test/helpers.rs
+++ b/helix-term/tests/test/helpers.rs
@@ -322,6 +322,12 @@ impl AppBuilder {
     }
 }
 
+pub async fn run_event_loop_to_idle(app: &mut Application) {
+    let (_, rx) = tokio::sync::mpsc::unbounded_channel();
+    let mut rx_stream = UnboundedReceiverStream::new(rx);
+    app.event_loop_until_idle(&mut rx_stream).await;
+}
+
 pub fn assert_file_has_content(file: &mut File, content: &str) -> anyhow::Result<()> {
     file.flush()?;
     file.sync_all()?;
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -13,6 +13,7 @@ use std::future::Future;
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::Arc;
+use std::time::SystemTime;
 
 use helix_core::{
     encoding,
@@ -126,6 +127,10 @@ pub struct Document {
     pub history: Cell<History>,
 
     pub savepoint: Option<Transaction>,
+
+    // Last time we wrote to the file. This will carry the time the file was last opened if there
+    // were no saves.
+    last_saved_time: SystemTime,
 
     last_saved_revision: usize,
     version: i32, // should be usize?
@@ -368,6 +373,7 @@ impl Document {
             version: 0,
             history: Cell::new(History::default()),
             savepoint: None,
+            last_saved_time: SystemTime::now(),
             last_saved_revision: 0,
             modified_since_accessed: false,
             language_server: None,
@@ -557,9 +563,11 @@ impl Document {
 
         let encoding = self.encoding;
 
+        let last_saved_time = self.last_saved_time;
+
         // We encode the file according to the `Document`'s encoding.
         let future = async move {
-            use tokio::fs::File;
+            use tokio::{fs, fs::File};
             if let Some(parent) = path.parent() {
                 // TODO: display a prompt asking the user if the directories should be created
                 if !parent.exists() {
@@ -567,6 +575,17 @@ impl Document {
                         std::fs::DirBuilder::new().recursive(true).create(parent)?;
                     } else {
                         bail!("can't save file, parent directory does not exist");
+                    }
+                }
+            }
+
+            // Protect against overwriting changes made externally
+            if !force {
+                if let Ok(metadata) = fs::metadata(&path).await {
+                    if let Ok(mtime) = metadata.modified() {
+                        if last_saved_time < mtime {
+                            bail!("file modified by an external process, use :w! to force");
+                        }
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


helix-view/src/document.rs
```diff
@@ -571,6 +579,17 @@ impl Document {
                 }
             }
 
+            // Protect against overwriting changes made externally
+            if !force {
+                if let Ok(metadata) = fs::metadata(&path).await {
+                    if let Ok(mtime) = metadata.modified() {
+                        if last_saved_time < mtime {
+                            bail!("file changed from under us, unable to save, use :w! to force");
```
kirawi, line 587:
```suggestion
                            bail!("file changed by an external process, use :w! to force");
```

helix-view/src/document.rs
```diff
@@ -979,6 +1000,7 @@ impl Document {
             rev
         );
         self.last_saved_revision = rev;
+        self.last_saved_time = SystemTime::now();
```
kirawi, line 1003: Since the file isn't saved, I don't think the last_saved_time should be set.

helix-view/src/document.rs
```diff
@@ -571,6 +579,17 @@ impl Document {
...
+                    if let Ok(mtime) = metadata.modified() {
+                        if last_saved_time < mtime {
+                            bail!("file modified by an external process, use :w! to force");
+                        }
+                    }
+                }
+            }
+
             let mut file = File::create(&path).await?;
             to_writer(&mut file, encoding, &text).await?;
```
kirawi, line 595: We should set save time here.

helix-view/src/document.rs
kov: My thinking here is to mimic the way History handles revisions and last_saved_revision here. They have an initial value as if the file had been saved at the time it was opened, this avoids adding a lot of unnecessary handling for an Option<>.

helix-view/src/document.rs
kov: I don't think this would be correct, or possible due to this happening inside the future, self would need to be captured. The state is only committed when the DocumentSavedEvent that is created here gets processed by the caller of the future. I decided to set the time in set_last_saved_revision() because there is no reason I can think of you'd want to call that without updating the time, and that ensures we do not forget to do it if more call sites are added for set_last_saved_revision().

pascalkuthe (approved):
I recently got bitten pretty badly by helix habing nor protection fkr this case and lost about a days work. I looked into a watchmen based solution (which is more accurate as it uses files to synchronize os events and also offers a lot for other features so it's something I want to look into in the future) but I concluded that we want mtime based overwrite protection for a couple of reasons:
* watchmen is an external program, this functionality is basic/important enough that it should work without watchmen
* watchmen is designed to watch a directories. We can use the root dir but this doesn't work when helix opens other random files with `:open` or with the LSP. 

For these reasons I was looking to implement mtime based overwrite protection as a fallback. I got sidetracked and just found that there is an existing PR for this so it's a good thing I didn't already get started with this.

Implementation wise this is pretty simple and LGTM. I would have implemented this the same way. I think it makes sense to treat `last_saved_time` the same as `last_saved_rev`. 

The only nitpick: `last_saved_time` should be added to the `Debug` implementation of `Document`

the-mikedavis:
I think this should be configurable: I don't quite remember the context where this has happened to me previously (maybe it involved containers, virtual machines or external formatters) but if you're in a situation where mtime isn't reliable then you have to fight the editor to save files every time.

kirawi:
Perhaps we could check if the buffer's content differs from the file's if the file was written at a later mtime?

pascalkuthe:
> I think this should be configurable: I don't quite remember the context where this has happened to me previously (maybe it involved containers, virtual machines or external formatters) but if you're in a situation where mtime isn't reliable then you have to fight the editor to save files every time.

That sounds reasonable and something I didn't think about. Adding a simple configuration flag like `editor.enable-overwrite-protection` should be pretty trivial. I think it would be good to enable that by default tough. `mtime` being unreliable is probably an exception rather than the rule.

helix-term/tests/test/helpers.rs
```diff
@@ -322,6 +322,12 @@ impl AppBuilder {
     }
 }
 
+pub async fn run_event_loop_to_idle(app: &mut Application) {
```
LeoniePhiline, line 325:
```suggestion
pub async fn run_event_loop_until_idle(app: &mut Application) {
```
matching `app.event_loop_until_idle(...)`

divarvel:
Since a lot of people are waiting on this feature, i took the liberty to apply requested feedback and merge with the latest `master`: https://github.com/divarvel/helix/tree/protect-against-overwrites

@kov I have opened a PR against your branch, but the merge from master makes it a bit hard to read. You can also cherry-pick commits if you want and rebase everything on top of a recent master (the conflicts are trivial to solve).

@pascalkuthe i can also open a new PR and rebase everything on top of the latest master, if that's simpler for maintainers.

pascalkuthe:
> Since a lot of people are waiting on this feature, i took the liberty to apply requested feedback and merge with the latest `master`: [divarvel/helix@`protect-against-overwrites`](https://github.com/divarvel/helix/tree/protect-against-overwrites?rgh-link-date=2023-02-03T14%3A32%3A38Z)
> 
> @kov I have opened a PR against your branch, but the merge from master makes it a bit hard to read. You can also cherry-pick commits if you want and rebase everything on top of a recent master (the conflicts are trivial to solve).
> 
> @pascalkuthe i can also open a new PR and rebase everything on top of the latest master, if that's simpler for maintainers.

Since there wasn't respose here for quite a while you can just open a new PR that supercedes. Aslong as you keep the original commits around to give credit and give credit in the PR description that is fine.

divarvel:
Alright, done in https://github.com/helix-editor/helix/pull/5805

kov:
I'm closing this in favour of #5805 thank you =)

PR #4750, opened 2022-11-15
Limit the number of items in the jumplist

the-mikedavis:
Previously, jumplists could grow unchecked. Every transaction is applied to jumplist selections to ensure that they are up to date and within document bounds, so this would cause every edit to become more expensive as jumplist lengths increased throughout a session. Setting a maximum number of entries limits the cost.

Vim and Neovim limit their jumplists:

* https://github.com/vim/vim/blob/b298fe6cbae3b240b10dbd55d9c38d0cc8e033d3/src/structs.h#L141
* https://github.com/neovim/neovim/blob/e8cc489accc435076afb4fdf89778b64f0a48473/src/nvim/mark_defs.h#L57

Notably, Kakoune does not. In Kakoune, changes are applied to jumplist entries lazily as you hit `<C-o>`/`<C-i>` though, so Kakoune doesn't have the same growing cost concerns. Kakoune also does not have a concept of a View which limits the cost further.

Vim and Neovim limit to 100. This seems unreasonably high to me so I've set this to 30 to start. We can increase if this is problematically low.

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -3,29 +3,35 @@ use helix_core::{
     pos_at_visual_coords, visual_coords_at_pos, Position, RopeSlice, Selection, Transaction,
 };
 
-use std::fmt;
+use std::{collections::VecDeque, fmt};
+
+const JUMP_LIST_CAPACITY: usize = 30;
 
 type Jump = (DocumentId, Selection);
 
 #[derive(Debug, Clone)]
 pub struct JumpList {
-    jumps: Vec<Jump>,
+    jumps: VecDeque<Jump>,
     current: usize,
 }
 
 impl JumpList {
     pub fn new(initial: Jump) -> Self {
-        Self {
-            jumps: vec![initial],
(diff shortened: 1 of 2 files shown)


helix-view/src/view.rs
```diff
@@ -3,29 +3,35 @@ use helix_core::{
...
     pub fn push(&mut self, jump: Jump) {
         self.jumps.truncate(self.current);
         // don't push duplicates
-        if self.jumps.last() != Some(&jump) {
-            self.jumps.push(jump);
+        if self.jumps.back() != Some(&jump) {
+            // If the jumplist is full, drop the oldest item.
+            if self.jumps.len() == JUMP_LIST_CAPACITY {
+                self.jumps.pop_front();
+            }
```
archseer, line 32: let's `while self.jumps.len() >= JUMP_LIST_CAPACITY` `pop_front()` to avoid potential weird edgecases where there could be `> CAPACITY` jumps

PR #4752, opened 2022-11-15
Select surrounding characters when using match/surround (m) mode

stonewareslord:
When doing a `surround add` operation, the original selection is kept in place, which makes it difficult to chain surrounds and does not feel idiomatic. 

It might feel more natural when chaining surround operations. For example, quoting something, then wrapping parenthesis, then prepending `Box::new` might look like a selection of `someStringContent` then `ms"ms(iBox::new`

This PR/MR changes the behavior of the editor. 

`ms"` before:

`ms"` after this commit:

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4529,9 +4529,12 @@ fn surround_add(cx: &mut Context) {
         };
         let (view, doc) = current!(cx.editor);
         let selection = doc.selection(view.id);
+        let index = selection.primary_index();
         let (open, close) = surround::get_pair(ch);
 
         let mut changes = Vec::with_capacity(selection.len() * 2);
+        let mut ranges = SmallVec::with_capacity(selection.len());
+
         for range in selection.iter() {
             let mut o = Tendril::new();
             o.push(open);
@@ -4539,10 +4542,20 @@ fn surround_add(cx: &mut Context) {
             c.push(close);
             changes.push((range.from(), range.from(), Some(o)));
             changes.push((range.to(), range.to(), Some(c)));
+
+            // Add 2 characters to the range to select them
+            ranges.push(if range.anchor < range.head {
+                Range::new(range.anchor, range.head + 2)
+            } else {
+                Range::new(range.anchor + 2, range.head)
+            });
         }
 
         let transaction = Transaction::change(doc.text(), changes.into_iter());
         apply_transaction(&transaction, doc, view);
+
+        // Extend the selections to include the surrounding character
+        doc.set_selection(view.id, Selection::new(ranges, index));
     })
 }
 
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -4529,20 +4529,33 @@ fn surround_add(cx: &mut Context) {
...
+
+            // Add 2 characters to the range to select them
+            ranges.push(if range.anchor < range.head {
+                Range::new(range.anchor, range.head + 2)
+            } else {
+                Range::new(range.anchor + 2, range.head)
+            });
         }
 
         let transaction = Transaction::change(doc.text(), changes.into_iter());
```
archseer, line 4554: Transaction takes a `.with_selection` instead of `doc.set_selection` afterwards

helix-term/src/commands.rs
```diff
@@ -4529,20 +4529,33 @@ fn surround_add(cx: &mut Context) {
...
             c.push(close);
             changes.push((range.from(), range.from(), Some(o)));
             changes.push((range.to(), range.to(), Some(c)));
+
+            // Add 2 characters to the range to select them
+            ranges.push(if range.anchor < range.head {
+                Range::new(range.anchor, range.head + 2)
+            } else {
+                Range::new(range.anchor + 2, range.head)
+            });
```
archseer, line 4551: This won't work correctly on multiple selections since selections need to be offset based on these added characters. You need to track how many characters were inserted in a separate variable, then add that to the ranges (e.g. selection after the first one needs to be offset by 2 chars forward). See around the codebase how `offs` is used: https://github.com/helix-editor/helix/blob/506e8caba7887a86bc686fc778dc7ce0f1b4b79d/helix-term/src/commands.rs#L2636

archseer: See also how shell needed to be fixed recently https://github.com/helix-editor/helix/commit/c74b97447f7a84f7728ce41c796f435371563a68

helix-term/src/commands.rs
stonewareslord: Fixed in cb0f7c907ed966961b926798e6d909e24bbb6883

helix-term/src/commands.rs
stonewareslord: I tried fixing this in b3969f2e8b4f36f36b23da07c04c7506baf70f49 and 167dbc4c9ca7bfe819d9568e2be8e9fbfdb796fa

The correct length is 2 as in 2 chars, and not char.len_utf8(), right?

I will also write tests for this new behavior similar to how `shell` fixed it in c74b974

archseer:
Thanks for working on this!

dead10ck:
This makes way more sense. I had gotten used to just doing my surround chaining in reverse order, but this is much more natural. Thanks!

Issue #1237, opened 2021-12-06
Named scratch buffers


Omnikar:
#### Describe your feature request

Allow scratch buffers to have names, rather than all buffers without paths being named `[scratch]`.

archseer:
We should probably store documents by URI rather than path, that way files are under `file://`, and we can have internal buffers under `helix://` i.e. `helix://scratch/:name`

xpe:
It might help us understand each other better if we explain our use-cases for why we might want named scratch buffers. Here are some of mine:
- I'm not ready to decide where to save a buffer yet, but I want a convenient name for it
- When I use the current buffer picker, it just shows "[scratch]"...
   - which makes it harder to quickly find at a glance
   - which makes it slower to select, because I have to select by id; e.g. with `%id 5` (see screenshot below)

<img width="682" height="454" alt="Image" src="https://github.com/user-attachments/assets/6759c59f-9a81-4006-8e58-28f8e0fb399f" />

Issue #475, opened 2021-07-20
Show open (file-)buffers as a "tabline" and allow fast buffer switching with space+[0-9]


thomkeh:
I know I can switch buffers with space+b and then searching or moving, but this still introduces some kind of mental cost for buffer switching, especially when you do it frequently. I also feel like a _modern_ editor should show a list of the open files.

With (n)vim, both of these problems are solved by [buftabline.nvim](https://github.com/jose-elias-alvarez/buftabline.nvim). It shows the open buffers in the "tabline" and allows switching to them with <leader>+buffer number. This allows fast access to the first 10 buffers, at almost no mental cost. (From experience I can tell you that once you get used to this, all other methods of buffer switching will feel excruciatingly slow.)

I know that keymappings in helix deliberately do not make use of the numbers, because they're farther away, but this isn't really an editing command, so I think it's fine in this case, though I don't know how you see this.

pickfire:
Personally I don't like this because it waste space (I am on a 80x24 terminal). May seemed useful at times but I might still use `space b` for now.

kirawi:
Could be something configurable.

antoyo:
As an additional argument, I find this feature almost required to make any use of the commands `goto_next_buffer` and `goto_previous_buffer` as we currently have no way to know what are the next and previous buffers.

kirawi:
As a bufferline implementation was merged a while ago, I'm marking this as a `good-first-issue` for the fast buffer switching shortcut.

Elias-Graf:
I have implemented the command `buffer-goto` here: https://github.com/helix-editor/helix/pull/4756. 

I've currently implemented it only as a normal, typed command. So no `+<i>`, nor <kbd>space</kbd> + <kbd>&lt;i&gt;</kbd>. 

Usage: `buffer-goto <i>` or `b <i>`. 

I've also refactored the "previous" / "next" buffer commands a bit, as there was some functional overlap, and a naming collision with the (now removed) function `goto_buffer(&Editor, Direction)`.

[...]

PR #4756, opened 2022-11-15
Goto buffer

Elias-Graf:
Implemented a command to go to a specific buffer by index.

Command: `buffer-goto <i>`
Alias: `b <i>`

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -728,17 +728,28 @@ fn goto_buffer_by_direction(editor: &mut Editor, direction: Direction) {
         Direction::Backward => buffers_len - 1, // Would be out of bounds, wrap to back.
     };
 
-    goto_buffer_by_index_impl(editor, new_index);
+    // Safety: The above logic ensures that the new index is always in bounds.
+    goto_buffer_by_index_impl(editor, new_index).unwrap();
 }
 
 /// Goto a buffer by providing it's index.
 ///
 /// Note that the index starts from 0.
-/// Out of bounds indices are ignored / a noop.
-fn goto_buffer_by_index_impl(editor: &mut Editor, i: usize) {
-    if let Some(doc_id) = editor.documents.keys().nth(i).copied() {
-        editor.switch(doc_id, Action::Replace);
-    }
+///
+/// # Errors
+///
+/// Returns an error if the index could not be found - is out of bounds.
+fn goto_buffer_by_index_impl(editor: &mut Editor, index: usize) -> anyhow::Result<()> {
+    let doc_id = editor
+        .documents
+        .keys()
+        .nth(index)
+        .copied()
+        .ok_or_else(|| anyhow!("no buffer '{}' in editor", index + 1))?;
+
+    editor.switch(doc_id, Action::Replace);
+
+    Ok(())
 }
 
 fn extend_to_line_start(cx: &mut Context) {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -735,11 +735,9 @@ fn goto_buffer_by_direction(editor: &mut Editor, direction: Direction) {
 ///
 /// Note that the index starts from 0.
 /// Out of bounds indices are ignored / a noop.
-fn goto_buffer_by_idx_impl(editor: &mut Editor, i: usize) {
-    let all_buffs: Vec<&helix_view::DocumentId> = editor.documents.keys().collect();
-
-    if let Some(&&new_buff) = all_buffs.get(i) {
-        editor.switch(new_buff, Action::Replace);
+fn goto_buffer_by_index_impl(editor: &mut Editor, i: usize) {
+    if let Some(doc_id) = editor.documents.keys().nth(i).copied() {
+        editor.switch(doc_id, Action::Replace);
     }
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -660,35 +660,54 @@ fn goto_line_start(cx: &mut Context) {
 }
 
 fn goto_next_buffer(cx: &mut Context) {
-    goto_buffer(cx.editor, Direction::Forward);
+    goto_buffer_by_direction(cx.editor, Direction::Forward)
 }
 
 fn goto_previous_buffer(cx: &mut Context) {
-    goto_buffer(cx.editor, Direction::Backward);
+    goto_buffer_by_direction(cx.editor, Direction::Backward)
 }
 
-fn goto_buffer(editor: &mut Editor, direction: Direction) {
-    let current = view!(editor).doc;
+/// Gets the amount of buffers that are currently open.
+fn get_buffers_len(editor: &mut Editor) -> usize {
+    editor.documents.len()
+}
 
-    let id = match direction {
-        Direction::Forward => {
-            let iter = editor.documents.keys();
-            let mut iter = iter.skip_while(|id| *id != &current);
-            iter.next(); // skip current item
-            iter.next().or_else(|| editor.documents.keys().next())
-        }
-        Direction::Backward => {
-            let iter = editor.documents.keys();
-            let mut iter = iter.rev().skip_while(|id| *id != &current);
-            iter.next(); // skip current item
-            iter.next().or_else(|| editor.documents.keys().rev().next())
-        }
+fn goto_buffer_by_direction(editor: &mut Editor, dir: Direction) {
+    let curr_i = get_focused_buffer_idx(editor);
+    let buffs_len = get_buffers_len(editor);
+
+    let new_i = match dir {
+        Direction::Forward if curr_i < buffs_len - 1 => curr_i + 1,
+        Direction::Forward => 0, // Would be out of bounds, wrap to front.
+        Direction::Backward if curr_i > 0 => curr_i - 1,
+        Direction::Backward => buffs_len - 1, // Would be out of bounds, wrap to back.
+    };
+
+    goto_buffer_by_idx_impl(editor, new_i);
+}
+
+/// Goto a buffer by providing it's index.
+///
+/// Note that the index starts from 0.
+/// Out of bounds indices are ignored / a noop.
+fn goto_buffer_by_idx_impl(editor: &mut Editor, i: usize) {
+    let all_buffs: Vec<&helix_view::DocumentId> = editor.documents.keys().collect();
+
+    if let Some(&&new_buff) = all_buffs.get(i) {
+        editor.switch(new_buff, Action::Replace);
     }
-    .unwrap();
+}
 
-    let id = *id;
+fn get_focused_buffer_idx(editor: &mut Editor) -> usize {
+    let curr_id = view!(editor).doc;
 
-    editor.switch(id, Action::Replace);
+    editor
+        .documents
+        .keys()
+        .enumerate()
+        .find(|(_, d_id)| *d_id == &curr_id)
+        .expect("the current buffer was not in the buffers list")
+        .0
 }
 
 fn extend_to_line_start(cx: &mut Context) {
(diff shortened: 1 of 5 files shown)


helix-term/src/commands.rs
```diff
@@ -659,36 +668,79 @@ fn goto_line_start(cx: &mut Context) {
...
-        }
+    if let Some(&&new_buff) = all_buffs.get(i) {
+        editor.switch(new_buff, Action::Replace);
     }
-    .unwrap();
-
-    let id = *id;
-
-    editor.switch(id, Action::Replace);
 }
```
the-mikedavis, line 744:
```rust
fn goto_buffer_by_index_impl(editor: &mut Editor, index: usize) {
    if let Some(doc_id) = editor.documents.keys().nth(index).copied() {
        editor.switch(doc_id, Action::Replace);
    }
}
```

helix-term/src/commands.rs
```diff
@@ -659,36 +668,79 @@ fn goto_line_start(cx: &mut Context) {
...
+    goto_buffer_by_idx_impl(cx.editor, 6);
+}
+
+fn goto_eight_buffer(cx: &mut Context) {
+    goto_buffer_by_idx_impl(cx.editor, 7);
+}
+
+fn goto_ninth_buffer(cx: &mut Context) {
+    goto_buffer_by_idx_impl(cx.editor, 8);
+}
```
the-mikedavis, line 705: If this were a command, it should be implemented as a single command that uses `cx.count()` for the buffer number. I don't think that adding another command is necessary though: you can add the bindings you're after with just the typable command like so:

```toml
[keys.normal]
"C-1" = ":b 1"
"C-2" = ":b 2"
# ...
```

helix-term/src/commands.rs
```diff
@@ -659,36 +668,79 @@ fn goto_line_start(cx: &mut Context) {
...
+        .expect("current document was not in documents");
+
+    let new_i = match dir {
+        Direction::Forward if curr_i < buffs_len - 1 => curr_i + 1,
+        Direction::Forward => 0, // Would be out of bounds, wrap to front.
+        Direction::Backward if curr_i > 0 => curr_i - 1,
+        Direction::Backward => buffs_len - 1, // Would be out of bounds, wrap to back.
+    };
+
+    goto_buffer_by_idx_impl(editor, new_i);
```
the-mikedavis, line 731: This is a nitpick but can we replace all `i`/`idx`s with `index` and not use abbreviations like `curr` and `buffs`? The space saving is marginal and it makes it less readable

the-mikedavis: Also, it's typical in the codebase to use `doc_id` for the current document ID:

```suggestion
    let doc_id = &view!(editor).doc;
    let buffs_len = editor.documents.len();
    let curr_i = editor
        .documents
        .keys()
        .position(|d| d == curr_doc)
        .expect("current document was not in documents");

    let new_i = match dir {
        Direction::Forward if curr_i < buffs_len - 1 => curr_i + 1,
        Direction::Forward => 0, // Would be out of bounds, wrap to front.
        Direction::Backward if curr_i > 0 => curr_i - 1,
        Direction::Backward => buffs_len - 1, // Would be out of bounds, wrap to back.
    };

    goto_buffer_by_idx_impl(editor, new_i);
```

helix-term/src/commands/typed.rs
```diff
@@ -1803,6 +1831,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             fun: force_buffer_close_all,
             completer: None,
         },
+        TypableCommand {
+            name: "buffer-goto",
+            aliases: &["b"],
+            doc: "Goto the buffer <i>.",
```
the-mikedavis, line 1837:
```suggestion
            doc: "Goto buffer number <n>.",
```

helix-term/src/commands/typed.rs
```diff
@@ -1803,6 +1831,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
...
+            fun: goto_buffer_by_idx,
+            completer: None,
```
the-mikedavis, line 1839: It should be possible to add a small completer that only gives the current buffer numbers

helix-term/src/commands.rs
```diff
@@ -659,36 +668,79 @@ fn goto_line_start(cx: &mut Context) {
...
 }
 
-fn goto_buffer(editor: &mut Editor, direction: Direction) {
-    let current = view!(editor).doc;
```
the-mikedavis, line 671: I don't think it's necessary to change the body of the old `goto_buffer` function - it has the same behavior before and after except that going to the previous buffer skips from the opposite direction

helix-term/src/commands/typed.rs
```diff
@@ -260,7 +261,34 @@ fn buffer_previous(
...
+    let i_str = args
+        .first()
+        .ok_or_else(|| anyhow!("requires an argument for <i>"))?
+        .as_ref();
+    let i = i_str
+        .parse::<usize>()
+        .map_err(|_| anyhow!("'{}' is not a valid index", i_str))?;
+
+    // Convert to zero based index
+    goto_buffer_by_idx_impl(cx.editor, i - 1);
```
the-mikedavis, line 290: This won't give feedback when the passed number is not a buffer ID. `goto_buffer_by_idx_impl` could return a result that says that the buffer ID does not exist

helix-term/src/commands.rs
Elias-Graf: I mainly replaced the body of the function because I saw nearly identical code in both branches of the match statement. That just triggered my refactoring sense, lol.
After looking into it, using an iterator the way it is used here may be hard to understand / overkill for such a simple task.

> except that going to the previous buffer skips from the opposite direction

I'm honestly not certain what you mean by this. `buffer-next` goes "to the right", and `buffer-previous` goes "to the left". Have I missed something that the original implementation does?

helix-term/src/commands.rs
Elias-Graf: While I do disagree that those abbreviations are hard to read, I'm perfectly happy to change them for consistency's sake / project guidelines.

helix-term/src/commands.rs
Elias-Graf: This is much cleaner. But the only location I could (relevantly) find "keys.normal" was in `book/src/remapping.md`. Does that mean there shouldn't keybindings by default? If yes, I have to disagree, as this is such an essential feature. Maybe I am using something incorrectly?

How would I use `cx.count()`, or better, how is count "provided"?.

> Returns 1 if no explicit count was provided

I seem to always get 1, probably I'm not using it correctly.

QiBaobin:
can we add this only to command so that we can take advantage of cx.count()? Regarding typable commands, I think it would be better to use name instead of id. #4771

QiBaobin:
> @QiBaobin I haven't yet figured out how `cx.count()` works. This has already been suggested in the reviews above, and I'm working on it. I don't have too much time on my hands during the day, so I will most likely get to it in the evening.
> 
> Regarding using the buffer name instead of the buffer ID (I assume you mean index, because the ID is different AFAIK). I think that doesn't really work, since two buffers can be named the same thing. For example, I was testing with multiple `[scratch]` buffers. In addition, you can already quickly switch between them by typing space+ b and then the name of your buffer (path also already works). I'm specifically implementing this, so I can use a shortcut (I'm hoping for ctrl + [1-9]) to switch between tabs (buffers). I heavily utilize this in my workflow, where for example components have multiple relevant files (e.g. HTML, CSS, TS), or I'm switching back and forth between implementation and spec / unit test.
> 
> I appreciate any suggestions and ideas relevant to this pull request :)

Good point. I recommend use name or path as I want to bind a key to do something like scatchpad, will would launch one specific buffer like scratch or terminal very quick.

Regarding cx.count(), you can get it in a function defined in src/commands.rs. It would work like that: you type number first then the key you bind to that function for example C-b. So that 1 C-b would take you to the first buffer, 2 C-b would be the second etc.

Elias-Graf:
@QiBaobin ohhh, thanks for telling me that. I bet it would have been ages until I figured that out.  Sadly, this isn't really in-line with browsers / file explorers / terminals / other editors (maybe also see: https://github.com/helix-editor/helix/issues/475#issuecomment-1315187136), which can be frustrating to use because the functionality is essentially the same :/

QiBaobin:
> @QiBaobin ohhh, thanks for telling me that. I bet it would have been ages until I figured that out. Sadly, this isn't really in-line with browsers / file explorers / terminals / other editors (maybe also see: [#475 (comment)](https://github.com/helix-editor/helix/issues/475#issuecomment-1315187136)), which can be frustrating to use because the functionality is essentially the same :/

Then we only need add a TypableCommand to accept the argument, we don't need touch commands.rs, just bind keys in the default config or our own config file to use them like `C-1 = "buffer 1"` etc.

However, I still think it's better to support name and path too, can you add them so that I can close my PR? If you don't have time, I can copy your logic to my PR.

Elias-Graf:
@QiBaobin I will place them into the default config once I find it :P.

I'll see about the buffer name / path once I'm done with the things mentioned in the review. I would also appreciate the feedback of members / maintainers on this specifically (maybe @kirawi or @the-mikedavis). I personally don't see the use case (yet? :D), but I'm happy to add it if it's generally desired.

Some questions that come into mind:
- What order would the evaluation be in?
- What order would the completions be in?
- What happens if a buffer is named `1`?

Elias-Graf:
@QiBaobin Would be so kind and tell me where to do this:

> just bind keys in the default config or our own config file to use them like C-1 = "buffer 1" etc.

I'm only able to find `helix-term/src/keymap/default.rs`, which, as far as my testing goes, does not support adding things like `"C1" => ":b1"`. I also found relevant text in the book `book/src/remapping.md`... which only talks about adjusting the local configuration.

helix-term/src/commands.rs
kirawi: We have typable variants of the `pipe` commands, so there's precedence. You can have a typable version, and another version that utilizes `cx.count()`

QiBaobin:
> 



> @QiBaobin Would be so kind and tell me where to do this:
> 
> > just bind keys in the default config or our own config file to use them like C-1 = "buffer 1" etc.
> 
> I'm only able to find `helix-term/src/keymap/default.rs`, which, as far as my testing goes, does not support adding things like `"C-1" => ":b1"`. I also found relevant text in the book `book/src/remapping.md`... which only talks about adjusting the local configuration.

After looking at the code, it turned out we don't have any global config file like languages.toml yet, maybe we can create one so that we don't need write functions for every key.

icefoxen:
>  I think that doesn't really work, since two buffers can be named the same thing.

iirc other editors suffix duplicate buffer names with a number when this happens, so if you have a buffer `foo.txt` then it's just called `foo.txt`, but if you open another buffer `foo.txt` then it is named `foo.txt <2>` or such.  To me this seems nice because it lets you just use names in the common case of no duplicate names.

pascalkuthe:
Closing this as stale. The vuffer picker and "gui like" order tabs workflow is not really a priority for us so I don't see us going forward with this.

Thank you for contributing!

xpe:
@pascalkuthe In the spirit of understanding, what aspects of the above were considered unappealing by the maintainers? Given that buffers have ids, it seems natural to have a command like `:buffer-goto <id>`, so I was trying to understand what about it felt off -- and if those concerns still apply.

I ask because I'm generally hoping to explore better ways to hop around buffers. I like the buffer picker as *one* modality. Perhaps polishing it a bit would scratch my itch. For example, if I have more than one scratch buffer, it is hard to jump to as quickly as I would like. One improvement would be [Feature Request: Named scratch buffers #1237](https://github.com/helix-editor/helix/issues/1237). Another might be [Feature Request (buffer-picker): include buffer-id in the default fuzzy-match columns #15746](https://github.com/helix-editor/helix/issues/15746).

PR #4758, opened 2022-11-15
Fix flaky tests in helix-term/tests/test/write.rs

dxtr85:


dead10ck:
What is the purpose of this change? The test here is ensuring that writes that should fail are failing and that it shows an error status. This change would defeat the purpose of the test.

dxtr85:
This change does nothing to test logic itself.
Since `get_status` returns Option this change makes sure that assertions are only being performed once we did retrieve `Some(status)` and no check is being done when there is `None` returned.
I think this is logically valid to only assert data when you have the data and not otherwise.

dxtr85:
OK then, closing this PR.
However I must say that this is pretty confusing for a newcomer to see his PR causing some testsuites in completely unrelated area to fail because you use flaki test as a gatekeeper for all incoming PRs. 
One could think: Hey I can with that and fix that in a moment.

dead10ck:
@dxtr85 maybe I missed something, but this should be a bug tracked in an issue. As the original author of these tests, I wasn't aware of these flaky failures, and now I do, so I can dedicate time to fixing it when I am able. As someone less familiar with the code base, reporting issues is at least as helpful as trying to fix them. :)

Issue #4754, opened 2022-11-15
`ayu`-Themes cursor has the same color as selection


philipgiuliani:
I've just installed the `helix` master branch and noticed that the cursor in the `ayu`-Themes have the same color as the text/background in the selection mode. This causes the cursor to get invisible.

## Master (Commit 652497bdd613d2854f63a1e99e9625a21a007389)
Cursor on `,`: 
![Screenshot 2022-11-15 at 09 28 08](https://user-images.githubusercontent.com/5446019/201868865-be9253b5-40e4-4762-bc63-498d51423613.png)

Cursor after the `,`:
![Screenshot 2022-11-15 at 09 28 15](https://user-images.githubusercontent.com/5446019/201868899-d26c627f-7231-4112-9a4e-5c328c4034f0.png)

## Version 22.8.1
In the previous version, the cursor was always visible with the orange background:
![Screenshot 2022-11-15 at 09 32 05](https://user-images.githubusercontent.com/5446019/201869321-520b8773-2440-4622-a486-d2e082bd2447.png)

philipgiuliani:
@enkodr I saw you have recently updated the ayu themes: https://github.com/helix-editor/helix/pull/4662

You have any idea how to change this? ☺️

enkodr:
I've introduced this change `"ui.cursor" = { modifiers = ["reversed"] }` that introduced this.

A new [PR](https://github.com/helix-editor/helix/pull/4764) to fix this was created.

pascalkuthe:
closing as this was fixed by #4764

PR #4764, opened 2022-11-15
Fix ayu theme cursor issue

enkodr:
Fix issue with with selection where the cursor is in an on a position with no content reported on issue #4754.


the-mikedavis:
Using orange for whitespace ends up being fairly noisy if you `:set whitespace.render all`. In general I think it is not advisable to use the reversed modifier for the cursor because of its interaction with the whitespace scope. Maybe we can avoid this problem by not using the `ui.virtual.whitspace` scope when visible whitespace is disabled?

the-mikedavis:
\cc @philipgiuliani does this look good to you?

philipgiuliani:
As you've mentioned, it doesn't look good with White spaces enabled (which I actually have). Also the documentation has changed from gray to blue, not sure if that in line with the ayu color scheme.

Before:
image

After:
Screenshot 2022-11-16 at 09 15 59

enkodr:
@philipgiuliani I don't know what might be different from my system to yours but it's rendering fine for me.
helix

enkodr:
@philipgiuliani can you check again, please?

philipgiuliani:
Hey @enkodr! It's a little visible now. Personally, I do not like the inverted cursor and the blue documentation. I've just checked out the theme on VIM and VSCode, they are also quite different one from another. All of them use gray documentation and an always-orange cursor.

### This branch
Screenshot 2022-11-18 at 09 28 48

### Helix 22.08.1
Screenshot 2022-11-18 at 09 30 35

### VSCode Ayu Dark
vscode

### VIM Ayu Dark
vim

enkodr:
To be honest, I really liked the inverted color but as there are other preferences from other people I switched back to the orange cursor and will use the inverted colors only to myself.
Regarding the blue color in selection, I also prefer the previous gray but the blue is the standard oficial suggestion as you can see in VS Code.

the-mikedavis:
What about the blue documentation? From the screenshots it looks like that uses dark gray in VSCode

enkodr:
@the-mikedavis I've reverted to the previous colors for both cursor and selection.

philipgiuliani:
In the commit https://github.com/helix-editor/helix/commit/d7136d691a32cb327c358ab9e19bd9d25d61a457 the key `comment.block.documentation` was introduced with the `fg` set to `blue`.

Looking at the vim and VSCode implementations, the documentation style should be set to match the comments.

Not sure if we should change the color to `gray` or remove this key completely, as it seemed to use the `comment` as a default.

```
"comment.block.documentation" = { fg = "gray", modifiers = ["italic"] }
```

enkodr:
You were right @philipgiuliani, I forgot that one. Should be OK now!

the-mikedavis (approved):

Issue #4539, opened 2022-10-31
Windows close in the wrong order


lukepighetti:
### Summary

If you open a bunch of windows, then start closing them, it always navigates the cursor to the first window, instead of the last selected window. Ideally the focus path travelled during window creation should be followed during window teardown

### Reproduction Steps

 

https://user-images.githubusercontent.com/666539/199011992-a1fa8309-2ed5-4802-a1be-99999412d119.mov

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

macOS

### Terminal Emulator

iTerm2

### Helix Version

helix 22.08.1 (6752c7d3)

wes-adams:
i'd like to address this issue

wes-adams:
MR: https://github.com/helix-editor/helix/pull/4766

wes-adams:
**good case:**

https://asciinema.org/a/7RuDbasCErUgVhI3G2BSc14yz

**bad case:**

split view, and then try to close original view

https://asciinema.org/a/l2A7sQfbeGpouIvt5MoNeDzfH

---

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', helix-view/src/tree.rs:286:29
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:48:5
   3: core::option::Option<T>::unwrap
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/option.rs:755:21
   4: helix_view::tree::Tree::get
             at ./helix-view/src/tree.rs:286:9
   5: helix_view::editor::Editor::cursor
             at ./helix-view/src/editor.rs:1316:27
   6: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::cursor
             at ./helix-term/src/ui/editor.rs:1553:15
   7: helix_term::compositor::Compositor::cursor
             at ./helix-term/src/compositor.rs:173:40
[...]

PR #4766, opened 2022-11-15
When buffer closes, focus on parent buffer

wes-adams:
Issue: https://github.com/helix-editor/helix/issues/4539

~~this solution adds a new field to the `Tree::Node` struct: `spawned_by`. when a new view is created, the new view's `spawned_by` field is populated with the `ViewId` of the view that was in focus at the time of creation. when a view is closed, i.e. `:q`, focus will move to the `spawned_by` view, provided it still exists. if it no longer exists, `Tree::focus_next()` is called, which moves focus to an arbitrary view that exists (as best i can tell).~~

2nd attempt:
~~i've implemented `focus_prev`. when a view closes, and it is the view in focus, `focus_prev` is called. it attempts to shift focus to the "`(current_view - 1)`" view (if it exists. if it does not, the legacy method `focus_next` is called).~~

3rd attempt:
see @the-mikedavis solution below. thanks for the help

### DIFF
--- a/helix-view/src/tree.rs
+++ b/helix-view/src/tree.rs
@@ -219,7 +219,7 @@ impl Tree {
 
         if self.focus == index {
             // focus on something else
-            self.focus = self.next();
+            self.focus = self.prev();
         }
 
         stack.push(index);
@@ -521,6 +521,26 @@ impl Tree {
         Some(child_id)
     }
 
+    pub fn prev(&self) -> ViewId {
+        // This function is very dumb, but that's because we don't store any parent links.
+        // (we'd be able to go parent.prev_sibling() recursively until we find something)
+        // For now that's okay though, since it's unlikely you'll be able to open a large enough
+        // number of splits to notice.
+
+        let mut views = self
+            .traverse()
+            .rev()
+            .skip_while(|&(id, _view)| id != self.focus)
+            .skip(1); // Skip focused value
+        if let Some((id, _)) = views.next() {
+            id
+        } else {
+            // extremely crude, take the last item
+            let (key, _) = self.traverse().rev().next().unwrap();
+            key
+        }
+    }
+
     pub fn next(&self) -> ViewId {
         // This function is very dumb, but that's because we don't store any parent links.
         // (we'd be able to go parent.next_sibling() recursively until we find something)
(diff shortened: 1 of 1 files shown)


archseer:
I'd just focus the split to the left/whatever split is in the same container as the view. If it's the only view then the next one in the parent. We shouldn't need to track this on the view itself

archseer:
Or just implementing `focus_prev` would probably be enough

wes-adams:
> Or just implementing `focus_prev` would probably be enough

thanks for the feedback. i'll rework this with this in mind.

wes-adams:
@archseer - i'd really appreciate another look when you can

helix-view/src/tree.rs
```diff
@@ -521,6 +521,24 @@ impl Tree {
...
+                    None => {
+                        let (key, _) = self.traverse().next().unwrap();
+                        self.focus = key;
+                    }
+                }
+            } else {
+                self.focus_next();
+            }
+        }
+    }
```
the-mikedavis, line 540: You could implement DoubleEndedIterator for Traverse and then this function would be almost the same as `focus_next`

the-mikedavis: i.e. it should be possible to write `focus_prev` like this:

```rust
let mut views = self
    .traverse()
    .rev()
    .skip_while(|&(id, _view)| id != self.focus)
    .skip(1);
if let Some((id, _)) = views.next() {
    self.focus = id;
} else {
    let (key, _) = self.traverse().rev().next().unwrap();
    self.focus = key;
}
```

helix-view/src/tree.rs
```diff
@@ -521,6 +521,26 @@ impl Tree {
...
+
+        let mut views = self
+            .traverse()
+            .rev()
+            .skip_while(|&(id, _view)| id != self.focus)
+            .skip(1); // Skip focused value
+        if let Some((id, _)) = views.next() {
+            id
+        } else {
+            // extremely crude, take the first item again
```
the-mikedavis, line 538:
```suggestion
            // extremely crude, take the last item
```

the-mikedavis (approved):

wes-adams:
@the-mikedavis thanks for the guidance and patience. it's much appreciated

PR #4767, opened 2022-11-15
Saving macros

jrhahn:
This adds the feature to save a macro in the current working directory.

I didn't want to always re-record macros whenever I restarted helix so I thought about adding the option to store and load macros.

The way it works:
```q``` opens the macro menu.
|-- ```r``` records a new macro.
|-- ```p``` replays a recorded or loaded macro
|-- ```s``` saves a macro (in the working directory)
|-- ```l``` loads a macro (from the working directory)

Limitations: 
- Loading currently works only for macros that were originally stored in the default register at '@'

### DIFF
--- a/helix-core/src/register.rs
+++ b/helix-core/src/register.rs
@@ -70,7 +70,7 @@ impl Registers {
             .insert(name, Register::new_with_values(name, values));
     }
 
-    pub fn load(&mut self, name: char) {
+    pub fn load(&mut self, name: char) -> Result<(), String> {
         fn _read_data(name: char) -> Result<String, io::Error> {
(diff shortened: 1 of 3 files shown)


the-mikedavis:
This can be accomplished by yanking to and pasting from the `@` register: `"@p` to paste and `"@y` to yank. Adding file operations to registers feels to specific to me.

archseer:
Yeah this will be solved by persistent state in the future.

jrhahn:
> Yeah this will be solved by persistent state in the future.

Cool, any idea when this is coming? Let me know if I can support.

PR #4771, opened 2022-11-16
add buffer command so that we can bind key to switch buffers

QiBaobin:
would like something like "space".t = ":buffer [scratch]"

this would be helpful if we support script in future


QiBaobin:
Close as #4756 supports name now.

PR #4773, opened 2022-11-16
add cmds to support command chain

QiBaobin:
For instances:
`w && bc`
`vsplit-new && insert-output git status`


QiBaobin:
close per #3393

Issue #3501, opened 2022-08-21
Mapped commands that would exit into normal mode do not propagate in history


j-james:
### Summary

I have many commands like `C-s = ":write"` set up in my configuration file, as I like having some base level of commands that work across all editors in addition to editor-specific commands. But something that bothers me a lot with Helix right now is that those commands don't increment the undo history.

If I ran `:w` normally, by exiting insert mode the undo history would be appended to. But since it's directly aliased, normal mode is never inserted and the undo history is never updated.

### Reproduction Steps

 

https://asciinema.org/a/80czKCCbVB3wCja7zQLCCVii0

With this configuration:

```toml
[keys.insert]
C-s = ":write"
C-u = "undo"
```

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

Linux

### Terminal Emulator

kitty 0.25.2

### Helix Version

helix 22.05

j-james:
Related: #1583

dead10ck:
Duplicate of #2883, and I would point you to [my comment in that thread as well](https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126)

danielrode:
Workaround for the write command:

```
[keys.insert]
C-s = ["commit_undo_checkpoint", ":write"]
```

omentic:
Fixed by #11062

PR #4780, opened 2022-11-16
Update file change history before calling MappableCommand

### LINKED ISSUE #4719: file-name element doesn't update when saving under INS mode
### Summary

Saving file while in insert mode doesn't properly update the file-name `[+]`. It seems to mostly be just a graphical error, but on Linux (at least the machine I was SSHed into), saving in insert mode saves the file, but helix doesn't let you quit.
(And the saving while in insert mode is done with key mapping)

### Reproduction Steps

 
`config.toml`:
```
[keys.insert]
 C-s = ":w"
```
I tried this:

1. Open file `hx temp.txt`
2. Enter ins mode `i`
3. Enter text "hello world"
4. While still in insert mode try to save `C-s`
5. Quit out of file `ESC`, `:q`

I expected this to happen:
File would save and [+] would go away

Instead, this happened:
On windows:
File successfully saved, but the `[+]` did not successfully update on the file-name status bar.

On Linux (at least though SSH)
`[+]` graphic did not update, but the following message appeared under the status bar: `'temp.txt' written 4L 34B`. 
After exiting INS mode and trying to `:q` I get the following message: `1 unsaved buffer(s) remaining: ["temp.txt"]`. Force quitting exits the editor and the file seems to have saved correctly.

### Helix log

```
2022-11-12T04:17:39.287 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:17:53.244 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["temp.txt"]
2022-11-12T04:18:09.911 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:20:13.897 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:20:23.360 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["temp.txt"]
2022-11-12T04:20:28.330 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:20:41.857 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["temp.txt"]
2022-11-12T04:24:58.337 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:25:09.863 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["temp.txt"]
2022-11-12T04:26:34.739 helix_view::clipboard::provider [WARN] No clipboard provider found! Yanking and pasting will be internal to Helix
2022-11-12T04:28:19.549 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: [".cache/helix/helix.log"]
```

</details>

### Platform

Windows + Linux over SSH

### Terminal Emulator

Windows Terminal & xterm-256color over SSH

### Helix Version

helix 22.08.1 (001e4e30)

swork1:
Before we run a mappable command we should update the current changes to history.

 This fixes #4719

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -172,6 +172,11 @@ impl MappableCommand {
                         jobs: cx.jobs,
                         scroll: None,
                     };
+
+                    // We update any changes to history before calling the command
+                    let (view, doc) = current!(cx.editor);
+                    doc.append_changes_to_history(view.id);
+
                     if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
                         cx.editor.set_error(format!("{}", e));
                     }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -167,16 +167,12 @@ impl MappableCommand {
             Self::Typable { name, args, doc: _ } => {
                 let args: Vec<Cow<str>> = args.iter().map(Cow::from).collect();
                 if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
+                    commit_undo_checkpoint(cx);
                     let mut cx = compositor::Context {
                         editor: cx.editor,
                         jobs: cx.jobs,
                         scroll: None,
                     };
-
-                    // We update any changes to history before calling the command
-                    let (view, doc) = current!(cx.editor);
-                    doc.append_changes_to_history(view.id);
-
                     if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
                         cx.editor.set_error(format!("{}", e));
                     }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -167,7 +167,10 @@ impl MappableCommand {
             Self::Typable { name, args, doc: _ } => {
                 let args: Vec<Cow<str>> = args.iter().map(Cow::from).collect();
                 if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
-                    commit_undo_checkpoint(cx);
+                    // #4719 - Fix for checkpoint issues when calling commands in INSERT mode
+                    if cx.editor.mode == Mode::Insert {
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Alternatively, keybinds that call typable commands from insert-mode could include `commit_undo_checkpoint` before the call to save

the-mikedavis:
I meant that it could be done in keymaps:

```toml
[keys.insert]
C-s = ["commit_undo_checkpoint", ":w"]
```

but I suppose there isn't really a downside to doing this for all typable commands. Mostly this will just be a no-op since you need to enter normal/select modes to execute typable commands and that also saves a checkpoint

archseer:
This breaks how the undo is supposed to work though, we've had this issue & PR before: https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126 https://github.com/helix-editor/helix/issues/3501

archseer:
I think this needs some discussion but maybe having the undo state broken into two undo steps is preferable to having this broken.

helix-term/src/commands.rs
```diff
@@ -167,6 +167,7 @@ impl MappableCommand {
             Self::Typable { name, args, doc: _ } => {
                 let args: Vec<Cow<str>> = args.iter().map(Cow::from).collect();
                 if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
+                    commit_undo_checkpoint(cx);
```
archseer, line 170: Should only be called if in insert mode, together with a comment explaining why this is needed in the first place

pascalkuthe:
I don't think this is the right solution to the problem. Some commands are fine to call from insertmode with creating a checkpoint. These commands should not be changed to create a checkpoint. 

Issue #4723, opened 2022-11-12
Go to next symbol


lukepighetti:
Would be nice to have a bracket pair `]s` and `[s` for "go to next symbol"

PR #4787, opened 2022-11-17
4723 go to next symbol with s

dariooddenino:
As requested by #4723, this PR adds `goto_next_symbol`, `goto_prev_symbol`, `goto_first_symbol` and `goto_last_symbol`.

There are a lot of things I'm unsure about this and that I might have gotten wrong, mainly because async in rust is still a bit of a mistery to me. Feel free to let me know if reviewing this PR is just too much work for what it's worth, I won't be offended :smile: 

Some observations/questions:

1) It would be really useful to have a coding practices/guide to know where to put what. I found that often that distinction is not completely clear to me (lsp vs commands, for example). I have written a bunch of helper functions here that I have left in `commands.rs` but that should maybe be in a different file.

2) In `helix-term/src/commands.rs`'s `get_doc_lsp_symbols` I'm using `block_in_place` to get all the symbols. I have no idea if it's the appropriate approach there.

3) In `helix-term/src/commands/lsp.rs`'s `symbol_picker` I can't use `nested_to_flat_symbols`, as I get an error about `History` that cannot be shared between threads. I'm not entirely sure how to handle this.


helix-term/src/commands.rs
```diff
@@ -2962,6 +2970,122 @@ fn goto_next_change_impl(cx: &mut Context, direction: Direction) {
...
 
+/// Gets all the lsp symbols for the current document.
+fn get_doc_lsp_symbols(cx: &mut Context) -> Option<Vec<SymbolInformation>> {
+    let doc = doc!(cx.editor);
+    let language_server = doc.language_server()?;
+    let future = match language_server.document_symbols(doc.identifier()) {
+        Some(future) => future,
+        None => {
+            cx.editor
+                .set_error("Language server does nto support document symbols");
```
dragfire, line 2981: typo `not`

pascalkuthe:
I am closing this PR as stale. I also don't see us moving forward with a feature like this. For movements we tend to use a treesitter driven approach instead. 

Thank you for contributing!

PR #4794, opened 2022-11-17
Update tutor

massimiliano-della-rovere:
fixed indentation


Omnikar:
Those indentations are intentional; they basically indicate that it's a continuation of the above, outdented bullet.

PR #4813, opened 2022-11-19
feat(themes): improve fleetish

matoous:
Updates to the `fleetish` theme. Specifically:

- use darker background for menues as it (purely subjectively) loooks better
- use different color for `constant.buildin.boolean` and `keyword` to make statements such as `return true` read better
- use different colors for different markup link parts
- destructure the config where appliable

The PR also introduces new highlight scope called `function.declaration` that can be used to style function declaration different from function calls. Specifically for fleet, function declarations are white & bold and the thing that pops-out in the code the most. I think this might be useful also for other themes but happy to leave it out of this PR.

cc @krfl as the theme author.

Inspired by: https://github.com/cloudwindy/fleet-dark and https://github.com/Michaelzhouisnotwhite/Jetbrains-Fleet-Theme plus some manual digging through the Fleet itself.

### DIFF
--- a/runtime/themes/fleetish.toml
+++ b/runtime/themes/fleetish.toml
@@ -1,101 +1,116 @@
 # Author: Kristoffer Flottorp <kr.fl@outlook.com>
 # A take on the JetBrains Fleet theme sprinkled with some creative freedom
 
-"type" = { fg = "yellow" } # .builtin
-"constructor" = { fg = "yellow" }
-"constant" = { fg = "cyan" }
+"type" = "light_blue"
+"type.builtin" = "orange"
+"constructor" = "yellow"
+"constant" = "cyan"
 # "constant.builtin" = {} # .boolean
-"constant.builtin.boolean" = { fg = "cyan" } # .boolean
-# "constant.character" = {} #.escape
-"constant.numeric" = { fg = "yellow" } # .integer / .float
-"string" = { fg = "pink" } # .regexp
-# "string.special" = {} #.path / .url / .symbol
-"string.special" = { modifiers = ["underlined"] } #.path / .url / .symbol
-"comment" = { fg = "dark_gray" } # .line
+"constant.builtin.boolean" = "yellow"
+"constant.character" = "yellow"
+"constant.characted.escape" = "light"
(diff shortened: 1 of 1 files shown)


archseer:
There's a fleetish derivative that tries to mirror the official theme as accurately as possible: https://github.com/Michaelzhouisnotwhite/Jetbrains-Fleet-Theme

Maybe we can improve further based on their scopes

matoous:
@archseer I tried to adjust the whole theme based on the linked repo but it also doesn't match Fleet on certain files. I did my best to extract parts of it but also open fleet and try and match it's highlighting where possible.

krfl:
Hello @matoous.

Late reply here, I know. Life stuff.

I totally see your point in some of these changes. This update does make the theme match the official Fleet theme in some ways, but it also makes it different in others. The Fleetish theme was never intended to be a perfect replica of the official Fleet theme. It was originally based on a screenshot that turned out to be very different from the actual fleet (beta/preview) theme. I have deliberately kept it different in some areas, because I don't like the official Fleet theme as much as I like the one they used/still use in their marketing. I also don't think it's possible to match Fleet based on the highlighting scopes available in Helix and the observation that JetBrains seem to change their highlighting scope configurations for different file types. Perhaps, in hindsight, I should have changed the name of the theme once the beta/preview hit. 

We could change the name of the theme to just "Fleet" and try to match it as closely as possible if that is what the people want. This PR at least points in that direction.

matoous:
@krfl thanks for the reply! My only intention here was to improve (which when it comes to themes is highly subjective thing) a few highligting scopes and match the colors more closely as the color scheme used in `fleetish` seems slightly more pale than the one in fleet itself (and I personally enjoy higher contrast). I personally don't see a need to match the fleet theme 1-to-1 which as you said will never happen because they have different highlighting scopes per file type.

Feel free to adjust the theme in whatever direction you want, or we could even revert this. I am still experimenting with the colors even locally as there always comes a file type where I would like it to look slightly differently.

My only wish would be for this to lend: https://github.com/helix-editor/helix/pull/4892 as I really enjoy how the function declaration names pop out.

PR #3328, opened 2022-08-04
Completion by words in all loaded documents

estin:
Hi!
Draft: completion by words in all loaded documents

Works for document
 - with language server support (merge lsp result and finded words)
 - and without it

May be such feature must be implemented via plugins in future?
 
Снимок экрана в 2022-08-04 15-07-02

PR #4816, opened 2022-11-19
support word and line completion

QiBaobin:
Screen Shot 2022-11-19 at 20 09 36
Screen Shot 2022-11-19 at 20 08 17

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4142,11 +4142,7 @@ fn global_search_collect(
                     entry.path(),
                     sinks::UTF8(|line_num, line| {
                         if all_matches_sx
-                            .send((
-                                entry.path().to_owned(),
-                                line.to_string(),
-                                line_num as usize - 1,
-                            ))
+                            .send((entry.path().to_owned(), line.to_string(), line_num as usize))
                             .is_err()
                             || (max_item_count.is_some()
                                 && count.fetch_add(1, std::sync::atomic::Ordering::SeqCst)
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -65,7 +65,7 @@ use std::{
 use once_cell::sync::Lazy;
 use serde::de::{self, Deserialize, Deserializer};
 
-use grep_regex::RegexMatcherBuilder;
+use grep_regex::{RegexMatcher, RegexMatcherBuilder};
 use grep_searcher::{sinks, BinaryDetection, SearcherBuilder};
 use ignore::{DirEntry, WalkBuilder, WalkState};
 use tokio_stream::wrappers::UnboundedReceiverStream;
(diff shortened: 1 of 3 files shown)


the-mikedavis:
It looks like this works by shelling out to `rg`? You shouldn't need an additional tool installed to do this: word completion should be fully built-in

QiBaobin:
> 

Good point, fixed.

QiBaobin:
> It looks like this works by shelling out to `rg`? You shouldn't need an additional tool installed to do this: word completion should be fully built-in

Good point, fixed.

helix-term/src/commands.rs
```diff
@@ -3993,6 +3949,226 @@ pub fn completion(cx: &mut Context) {
...
+    let trigger_offset = cursor;
+    let config = cx.editor.config().file_picker.clone();
+
+    let callback = Box::pin(async move {
+        let matches = grep_impl(
+            &query,
+            current_content,
+            if search_root { Some(config) } else { None },
+            match_whold_line,
+        )
```
archseer, line 4008: I think we either want to only use open buffers or current buffer. Scanning the entire tree can easily be very costly and tons of disk I/O.

helix-term/src/commands.rs
```diff
@@ -3993,6 +3949,226 @@ pub fn completion(cx: &mut Context) {
...
+    let start_offset = cursor.saturating_sub(offset);
+    let prefix = regex::escape(&text.slice(start_offset..cursor).to_string());
+
+    completion_grep(
+        cx,
+        format!(r"\b{}\w+", prefix,),
+        start_offset,
+        search_root,
+        false,
+    );
```
archseer, line 3972: Using a regex to extract words is a lot less efficient than just scanning the rope. Here's another approach that does that https://github.com/helix-editor/helix/pull/3328

helix-term/src/commands.rs
```diff
@@ -1970,8 +1924,10 @@ fn global_search(cx: &mut Context) {
     let current_path = doc_mut!(cx.editor).path().cloned();
 
     let show_picker = async move {
-        let all_matches: Vec<FileResult> =
-            UnboundedReceiverStream::new(all_matches_rx).collect().await;
+        let all_matches: Vec<FileResult> = UnboundedReceiverStream::new(all_matches_rx)
+            .map(|(path, _, line_num)| FileResult::new(&path, line_num as usize - 1))
```
archseer, line 1928: Seems more wasteful to do this upfront rather than calculating it for matches only (after search_path)

helix-term/src/commands.rs
QiBaobin: It's very useful for editing source code without lsp, and completing lines is also very useful for writing similar logic in other files even there is a lsp.

We don't scan the file tree for C-x or idle completions, only do it when manually trigger by C-n and C-l, but also limit the match count to a few number.

helix-term/src/commands.rs
QiBaobin: can you guiding me how to write a common function or sink that accepts a transform lambda? I am stuck there, so I send all three fields to the channel.

ahmed1jilali:
I really need word completion, does any one know when it's going to happen, thanks

PR #4818, opened 2022-11-19
fix(build): don't statically link against the "musl" C standard library

usagi-flow:
By default, rustc always statically links against musl (*), unlike on glibc systems where the resulting binary is dynamically linked against glibc.

This has a very problematic side effect: Binaries which statically link against musl do not support the `dlopen()` call (which we do through the "libloading" crate) (**). **This in turn makes it impossible to load treesitter grammars.**

The result is an error in the log and the user is left without syntax highlighting, among others:
```
helix_core::syntax [ERROR] Failed to load tree-sitter parser for language "toml": Error opening dynamic library "/opt/helix-src/runtime/grammars/toml.so"
```

This PR ensures the resulting binary is never statically linked against a C standard library. This configuration should have no impact in environments where the resulting binary would be dynamically linked by default.

---

(*) Tested with v1.65.0. See also: https://github.com/rust-lang/rust/issues/34987 / https://github.com/rust-lang/compiler-team/issues/422

(**) I found it out by testing a call to `Library::new(...).unwrap()` in a statically and dynamically linked application. I couldn't find much information regarding my claim, but I found these:
https://www.openwall.com/lists/musl/2012/12/08/4
https://www.openwall.com/lists/musl/2014/08/27/14

I tested it with the following code:
```rust
use libloading::Library;

fn main() {
    let name = "/opt/helix-src/runtime/grammars/toml.so";

    println!("Loading: {}", name);
    let lib = unsafe { Library::new(name).unwrap() };

    println!("Unloading: {}", name);
    lib.close().unwrap();

    println!("Done");
}
```
Output (when built on musl with default options):
```
Loading: /opt/helix-src/runtime/grammars/toml.so
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: DlOpen { desc: "Dynamic loading not supported" }', src/main.rs:8:4
```

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,8 +2,16 @@
 xtask = "run --package xtask --"
 integration-test = "test --features integration --profile integration --workspace --test integration"
 
-[target.aarch64-unknown-linux-musl]
 # By default, on musl linux, statically linked binaries are produced.
 # But such binaries do not support loading libraries using `dlopen()` ("libloading" crate),
 # which is required for loading treesitter grammars.
+[target.arm-unknown-linux-musleabihf]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.armv7-unknown-linux-musleabihf]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.aarch64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.i686-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.x86_64-unknown-linux-musl]
 rustflags = ["-C", "target-feature=-crt-static"]
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -2,7 +2,7 @@
 xtask = "run --package xtask --"
 integration-test = "test --features integration --profile integration --workspace --test integration"
 
-[build]
+[target.aarch64-unknown-linux-musl]
 # By default, on musl linux, statically linked binaries are produced.
 # But such binaries do not support loading libraries using `dlopen()` ("libloading" crate),
 # which is required for loading treesitter grammars.
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -6,4 +6,4 @@ integration-test = "test --features integration --profile integration --workspac
 # But such binaries do not support loading libraries using `dlopen()` ("libloading" crate),
 # which is required for loading treesitter grammars.
 [target.'cfg(target_env = "musl")']
-rustflags = ["-C", "target-feature=-crt-static"]
+rustflags = "-C target-feature=-crt-static"
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -5,13 +5,5 @@ integration-test = "test --features integration --profile integration --workspac
 # By default, on musl linux, statically linked binaries are produced.
 # But such binaries do not support loading libraries using `dlopen()` ("libloading" crate),
 # which is required for loading treesitter grammars.
-[target.arm-unknown-linux-musleabihf]
-rustflags = ["-C", "target-feature=-crt-static"]
-[target.armv7-unknown-linux-musleabihf]
-rustflags = ["-C", "target-feature=-crt-static"]
-[target.aarch64-unknown-linux-musl]
-rustflags = ["-C", "target-feature=-crt-static"]
-[target.i686-unknown-linux-musl]
-rustflags = ["-C", "target-feature=-crt-static"]
-[target.x86_64-unknown-linux-musl]
+[target.'cfg(target_env = "musl")']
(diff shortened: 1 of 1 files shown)


archseer:
This shouldn't be the default for all targets, just musl.

archseer:
We should check what the alpine package does since that does build against musl and seems to work fine

archseer:
Hmm OK https://github.com/gtk-rs/gtk-rs-core/issues/387#issuecomment-977039928

kbridge:
I met the same problem today, by (manually) building helix on alpine linux.

And I solved it using the same technique you provided here.

i.e.

```
export RUSTFLAGS='-C target-feature=-crt-static'
```

I suggest this be mentioned in the [installation chapter](https://docs.helix-editor.com/install.html) of the documentation.

kbridge:
I don't know who to ping. Guess pinging the author is not a bad idea. @archseer

pascalkuthe:
I already added a note to the docs about this a while ago. You need to look at the [master docs](https://docs.helix-editor.com/master/install.html)

etienne-k:
Sorry for the delay, I nearly forgot about the existence of this PR, until I wanted to deploy a new Helix built from source on my raspberry pi nodes...

I changed the configuration file such that the compile flags apply to musl targets only, as suggested by @archseer .
I like that much more than simply including the flags for all builds, but it comes with the downside of having to hardcode different musl targets (I limited myself to common ARM targets used by RPi's, among others, and x86 targets), but the complete list is a bit bigger and probably not constant:

```
 $ rustc --print target-list | rg musl
aarch64-unknown-linux-musl
arm-unknown-linux-musleabi
arm-unknown-linux-musleabihf
armv5te-unknown-linux-musleabi
armv7-unknown-linux-musleabi
armv7-unknown-linux-musleabihf
hexagon-unknown-linux-musl
i586-unknown-linux-musl
i686-unknown-linux-musl
mips-unknown-linux-musl
mips64-openwrt-linux-musl
mips64-unknown-linux-muslabi64
mips64el-unknown-linux-muslabi64
mipsel-unknown-linux-musl
powerpc-unknown-linux-musl
powerpc64-unknown-linux-musl
powerpc64le-unknown-linux-musl
riscv32gc-unknown-linux-musl
riscv64gc-unknown-linux-musl
s390x-unknown-linux-musl
thumbv7neon-unknown-linux-musleabihf
x86_64-unknown-linux-musl
```

If someone knows how to configure this in a more elegant way: Be my guest. :)

pascalkuthe (approved):
I think this kind of hard coding is fine. Therer are probably very few people compiling for weird architectures *and* musl and those that do probably know how to add these flags themselves. I think you could actually remove the note about musl from the dos with this PR. It has been confusing some windows people that don't know what musl or glibc are

.cargo/config.toml
```diff
@@ -1,3 +1,17 @@
...
+[target.arm-unknown-linux-musleabihf]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.armv7-unknown-linux-musleabihf]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.aarch64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.i686-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
+[target.x86_64-unknown-linux-musl]
+rustflags = ["-C", "target-feature=-crt-static"]
```
polarathene, line 17: You can [match by `target_env`](https://doc.rust-lang.org/cargo/reference/config.html#target) which should cover all of these? (_the [`musleabihf` variant will also match as `musl`](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env)_)

You can also provide a string instead of an array for `rustflags`, AFAIK there is no practical difference ([rustflags from other sources are mutually exclusive](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags), no merging_).

```suggestion
[target.'cfg(target_env = "musl")']
rustflags = "-C target-feature=-crt-static"
```

Do note the [caveat of building with/without an explicit `--target`](https://doc.rust-lang.org/cargo/reference/config.html#buildrustflags). A [related issue](https://github.com/rust-lang/rust/issues/78210#issuecomment-714600426) where a build failed (_with `+crt-static` on a host with `-gnu` as the default implicit target_).

polarathene (changes requested):
Just a suggestion for a simpler representation :)

the-mikedavis:
Those docs moved here: https://github.com/helix-editor/helix/blob/585402d9ffa1d271e7c72fbb1c90d7e9d2d51482/book/src/install.md?plain=1#L185-L189

pascalkuthe:
this pr is stale and I actually included it with #8021 as I had to touch the config anyway (which would have made it impossible to configure this manually). Thanks for contributing!

Issue #4814, opened 2022-11-19
Bad language config: invalid type: map, expected a string for key `language.file-types` 


kalidzo:
### Summary

I found a bug in the 1079th line.

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:

Instead, this happened:

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

Arch linux

### Terminal Emulator

fish

### Helix Version

22.08.1

archseer:
You're using an older version of the editor, `languages.toml` on `master`works fine with the code on `master`.

the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/issues/4516

kalidzo:
oh, sorry)

PR #4819, opened 2022-11-19
#4814 bug fixed

kalidzo:
I found 2 bugs,  causing that:"Bad language config: invalid type: map, expected a string for key language.file-types"

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1076,7 +1076,7 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-regex", rev = "e1cf
 name = "git-config"
 scope = "source.gitconfig"
 roots = []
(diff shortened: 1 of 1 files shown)


the-mikedavis:
These were introduced in https://github.com/helix-editor/helix/pull/4414 and should not be reverted.

See https://github.com/helix-editor/helix/issues/4814#issuecomment-1320845034: you should update your installed helix version in order to use this new syntax. Alternatively, you can reference the [`languages.toml` file from the last release](https://github.com/helix-editor/helix/blob/22.08.1/languages.toml), although you should not need to copy the languages.toml file from the repository. The `~/.config/helix/languages.toml` is merged onto this repository's `languages.toml`.

kalidzo:
ok, thanks a lot

PR #4821, opened 2022-11-19
Implement simple indents.scm for Elixir

SoraTenshi:
Simple implementation of elixir indentations.

I am also not sure whether or not it is possible to do some kind of auto-pre outdent for "end", as it should be outdented before "end" and not after.

### DIFF
--- a/runtime/queries/elixir/indents.scm
+++ b/runtime/queries/elixir/indents.scm
@@ -1,0 +1,13 @@
+[
+  (after_block)
+  (anonymous_function)
+  (catch_block)
+  (do_block)
+  (else_block)
+  (rescue_block)
+  "->"
+] @indent
+
+[
+  "end"
+] @outdent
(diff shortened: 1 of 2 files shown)


runtime/queries/elixir/indents.scm
```diff
@@ -0,0 +1,13 @@
+[
+  (after_block)
+  (anonymous_function)
+  (catch_block)
+  (do_block)
+  (else_block)
+  (rescue_block)
+  "->"
```
the-mikedavis, line 8:
```suggestion
  (stab_clause)
```

The `->` won't work on its own because it doesn't have any children nodes for the indentation to apply to

SoraTenshi: Interestingly, it did work when i tried it, however your suggestion is better.

the-mikedavis (approved):

the-mikedavis:
> I am also not sure whether or not it is possible to do some kind of auto-pre outdent for "end", as it should be outdented before "end" and not after.

There isn't a way to do this currently

PR #4824, opened 2022-11-20
Select new pasted text in normal mode only

the-mikedavis:
#4458 changed the behavior of paste to select the newly inserted text. This is preferrable in normal mode because it's useful to be able to act on the new text. This behavior is worse for insert or select mode though:

* In insert mode, the cursor ends up on the last character of the newly selected text, so further typing inserts text before the last character.
* In select mode, the current selection is replaced with the new text selection which doesn't extend the current selection. With this change, the selection is extended to include the new text.

This aligns the behavior more closely with Kakoune, but it's coincidental instead of intentional: Kakoune doesn't implement bracketed paste (AFAIK) which causes this behavior in insert mode, and Kakoune doesn't have a select mode.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3467,7 +3467,14 @@ enum Paste {
     Cursor,
 }
 
-fn paste_impl(values: &[String], doc: &mut Document, view: &mut View, action: Paste, count: usize) {
+fn paste_impl(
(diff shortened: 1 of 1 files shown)


dead10ck (approved):
Thanks, this was a bit annoying 🙂

PR #4829, opened 2022-11-20
Fix broken html doctype highlighting

Chickenkeeper:
Adds the missing `<!` to the html highlights.scm to fix the incorrect highlighting of `DOCTYPE` declarations.

Before:
broken doctype highlight

After:
fixed doctype highlight


the-mikedavis (approved):

the-mikedavis:
Good catch, thanks!

PR #4830, opened 2022-11-20
Bump tree-sitter pattern matching limit to 256

the-mikedavis:
The limit of 64 breaks some highlighting in Erlang files with complicated record definitions. Bumping to 256  seems to work on all files I have seen.

| Before | After |
|---|---|
| before | after |

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -354,24 +354,25 @@ impl<'a> CapturedNode<'a> {
     }
 }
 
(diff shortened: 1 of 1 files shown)


archseer:
But not 128?

the-mikedavis:
128 is an improvement but it doesn't solve it for the files I've seen

here's that same case with 128:

pascalkuthe:
Considering treesitter used to limit to 32 in the past it might be worth looking into if that is some kind of issue in the elixir grammar (not as a blocker for this PR just something that might be good to look into)

PR #4837, opened 2022-11-21
support multiple selections for pickers

QiBaobin:
Support multiple selections like skim -m, so that we can open many files at once or delete all buffers or clear the jumplist.


BearOve:
What is stopping this from happening? I have wanted this feature many times. Anything I can do to help?

daedroza:
@QiBaobin : Would it be possible to update this PR with latest changes?

PR #4854, opened 2022-11-23
Clear line on `<ret>` when line is all whitespace

the-mikedavis:
This matches the insert-mode behavior for Vim and Kakoune: if the current line is empty except for whitespace, `<ret>` should insert a line ending at the beginning of the line, moving any indentation to the next line.

| Before | After |
|---|---|
| before | after |


poliorcetics:
That's awesome, thanks for the feature !

PR #4858, opened 2022-11-23
Exit select mode on surround commands

xJonathanLEI:
Similar to #4554, this PR fixes the bug where the editor doesn't exit select mode after performing a surround command. The following commands are affected:

- `surround_add` (`ms`)
- `surround_replace` (`mr`)
- `surround_delete` (`md`)

Sending this patch with @archseer's confirmation on Matrix, quoting here:

> Anything that does an action on the selection should exit select mode

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4582,6 +4582,7 @@ fn surround_add(cx: &mut Context) {
         let transaction = Transaction::change(doc.text(), changes.into_iter())
             .with_selection(Selection::new(ranges, selection.primary_index()));
         apply_transaction(&transaction, doc, view);
+        exit_select_mode(cx);
     })
 }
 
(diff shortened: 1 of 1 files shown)


QiBaobin:
One case in my mind, is that we want to do multiple commands like `ms(ms{`, after the change, we have to select again.

What's the benefit from this? How about some commands like replace selection with register? do we do the same change for them in future?

David-Else:
We just had https://github.com/helix-editor/helix/pull/4752 added that allows chaining `surround_add`, so this pull request does not seem to make sense? Also, Helix is not keeping the selection with `mr` and `md` on the latest master for me, are you using it in some different way to normal?

archseer:
The original selection is kept, but the mode is kept to normal rather than extend/select.

xJonathanLEI:
Exactly. Only the mode is changed, and the selection is still there for you to run subsequent commands on. The current behavior is rather unexpected and inconsistent with other commands (yank, replace, etc.).

QiBaobin:
Thanks all for information. Make sense to me now.

the-mikedavis (approved):

PR #4860, opened 2022-11-23
allow shell command non-zero return code while using output

QiBaobin:
Fix two issues found:
1. `$` shell_keep_pipe doesn't work as the logic will stop while there is no-zero shell return code.
2. `|` pipe_to, `!` shell_insert_output etc behavior different with expected, for instance, `|` then `rg -o foo`, we want all selections become 'foo' or empty instead of failing.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4783,7 +4783,6 @@ async fn shell_impl_async(
             log::error!("Shell error: {}", err);
             bail!("Shell error: {}", err);
         }
-        bail!("Shell command failed");
     } else if !output.stderr.is_empty() {
         log::debug!(
             "Command printed to stderr: {}",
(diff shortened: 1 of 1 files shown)


joshmuente:
Just tried it because the `$` was not working.
working good thanks for providing the fix

trink (changes requested):
Issue 1: keep_pipe definitely has a bug but if the command returns non-zero this PR still makes the assumption the failure is an error if it wrote to stderr; however, that is not universally true.

Issue 2: is not a Helix bug and ignoring errors can produce undesired/unexpected behavior.  There is no good way for Helix to determine acceptable errors from real errors for commands using multiple return codes.  They would have to be remapped to match expectations on a case by case basis.

`| wrapper.sh rg -o foo`

```sh
#!/bin/sh

$@
if [ $? -eq 1 ]; then # map the rg no match to success so the replace will use an empty string
    exit 0
fi
exit $?
```

PR #4733, opened 2022-11-13
Apply transactions to all views

the-mikedavis:
Previously, transactions were only applied to the currently focused view. This could lead to jumplist selections not being updated or panics if a jumplist selection pointed past the end of the document.

Closes #4732
Closes #4600
Closes #4841

PR #4865, opened 2022-11-23
Fix transaction application on undo/redo/earlier/later

the-mikedavis:
#4733 left a few edge-case bugs with undo/redo: `View::apply` was still being called directly in the `Document::undo_redo_impl` and `Document::earlier_later_impl` functions. This change removes those calls and applies the necessary transactions to all views in EditorView. In order to do this for `undo` and `earlier`, `History::changes_since` now handles the case where the given revision is less than the current revision and composes all of the inverted transactions between the two history points.

### DIFF
--- a/helix-core/src/history.rs
+++ b/helix-core/src/history.rs
@@ -126,16 +126,27 @@ impl History {
 
         match revision.cmp(&self.current) {
             Equal => None,
-            Greater => self.revisions[self.current + 1..=revision]
-                .iter()
-                .map(|revision| &revision.inversion)
-                .cloned()
-                .reduce(|acc, inversion| acc.compose(inversion)),
-            Less => self.revisions[revision + 1..=self.current]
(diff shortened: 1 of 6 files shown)


philipgiuliani:
Hey @the-mikedavis , does this resolve the following crash?

```
thread 'main' panicked at 'Position 11201 is out of range for changeset len 11200!', helix-core/src/transaction.rs:397:13
```

I see this on master since today (helix 22.08.1 (590a6284))

the-mikedavis:
I'm pretty sure this does fix that case. Since 42e37a57 panics like that should not happen anymore except for the bug about undo/redo/earlier/later. If you used undo/redo before getting that panic, it is most likely solved by this change

philipgiuliani:
Thanks, yes it happened on the undo/redo actions. Was just about to open issue, but then I saw your pull request (:

PR #4866, opened 2022-11-23
Added instructions for installing via AppImage

gautampanchal94:
Instructions for Linux Systems using AppImage.


README.md
```diff
@@ -117,6 +117,34 @@ Helix can be installed on macOS through homebrew:
 brew install helix
 ```
 
+## Linux
+
+### AppImage
+
+Install Helix using [AppImage](https://appimage.org/).
+(For now, it is only available for x86-64 bit architecture.)
```
the-mikedavis, line 125:
```suggestion
```

The next release will include a aarch64 appimage

README.md
```diff
@@ -117,6 +117,34 @@ Helix can be installed on macOS through homebrew:
...
+
+Download Helix AppImage form [Releases](https://github.com/helix-editor/helix/releases) page.
```
the-mikedavis, line 127:
```suggestion
Download Helix AppImage from the [Releases](https://github.com/helix-editor/helix/releases) page.
```

README.md
```diff
@@ -117,6 +117,34 @@ Helix can be installed on macOS through homebrew:
...
+
+```sh
+./helix-22.08.1-x86_64.AppImage --appimage-extract
+./squashfs-root/AppRun --version
+
+# Make available globally
+sudo mkdir -p /squashfs-root/helix
+sudo mv squashfs-root/* /squashfs-root/helix/
+sudo ln -s /squashfs-root/helix/AppRun /usr/bin/hx
+hx
```
the-mikedavis, line 145: I don't think we should advise this: isn't it one of the points of AppImages that they don't need priviledges to install or run?

the-mikedavis (commented):
Unlike the other installation instructions, these don't have anything really Helix-specific (like handling the runtime directory). I wonder if we need explicit instructions for AppImage at all? These instructions are how you would use AppImage in general and that feels out-of-place in Helix-specific installation instructions

README.md
gautampanchal94: Yea but who like to run from anywhere in file system it would be easier. We could make optional.

gautampanchal94:
Yea It's not Helix specific but lot of people maybe not know about Helix AppImage because it is not mentioned anywhere other than release page. So, I think it would be good idea to add it.
Also, it's for who want to try out helix. So, they don't have to install rust and cargo it who much easier.

the-mikedavis:
> it is not mentioned anywhere other than release page. So, I think it would be good idea to add it.

Yeah good point, it looks like we don't mention it anywhere yet. Adding a small note to the README and the [install page in the book](https://github.com/helix-editor/helix/blob/master/book/src/install.md) would be good. Although I would like to avoid writing out shell commands for these if possible since the version will be confusing.

> Also, it's for who want to try out helix. So, they don't have to install rust and cargo it who much easier.

The other pre-built binaries also don't need rust/cargo and are similarly easy to install. Maybe a section on pre-built binaries in general would be the best option?

gautampanchal94:
> Adding a small note to the README and the [install page in the book](https://github.com/helix-editor/helix/blob/master/book/src/install.md) would be good.

Yea It will work.

> since the version will be confusing.

Right now, I don't know how to handle that. We from next we could use keyword or something so shell script would be same. Still for now we can provide instruction about installing.
 
> Maybe a section on pre-built binaries in general would be the best option?

Yea It would be great!

gautampanchal94:
I deleted my fork, so there anyway I still can contribute to this pull request? or I have to make new pull request?

the-mikedavis:
Hmm I'm not sure how that works with github. A new PR for this would be ok 👍

PR #4881, opened 2022-11-25
Bump tree-sitter-html version

Chickenkeeper:
Bumps the tree-sitter-html version to address potential issues with trailing spaces.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -454,7 +454,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "html"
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

the-mikedavis:
Oh, actually the update to tree-sitter-html looks good (there was a change that stops parsing empty space as text). `punctuation.delimiter` is more appropriate for HTML `=` though. It isn't really an operator as it is in programming languages

Chickenkeeper:
> Oh, actually the update to tree-sitter-html looks good (there was a change that stops parsing empty space as text). `punctuation.delimiter` is more appropriate for HTML `=` though. It isn't really an operator as it is in programming languages

That's fair enough, I'll revert that and change the pr title

the-mikedavis (approved):

PR #4882, opened 2022-11-25
Update CSS syntax highlighting

Chickenkeeper:
Bump the tree-sitter-css version to fix issues with spaces before colons, adds missing highlights, attempts to make other highlights more consistent, and organises the rest.

### DIFF
--- a/runtime/queries/css/highlights.scm
+++ b/runtime/queries/css/highlights.scm
@@ -1,6 +1,12 @@
 (comment) @comment
 
 [
+ (tag_name)
+ (nesting_selector)
+ (universal_selector)
+] @tag
+
+[
  "~"
  ">"
  "+"
@@ -22,24 +28,16 @@
  "or"
 ] @keyword.operator
 
-((universal_selector) "*") @tag
-
-[
- (tag_name)
- (nesting_selector)
- (universal_selector)
-] @tag
-
 ((property_name) @variable
  (#match? @variable "^--"))
 ((plain_value) @variable
  (#match? @variable "^--"))
 
 (attribute_name) @attribute
-(class_name) @type
+(class_name) @label
 (feature_name) @variable.other.member
 (function_name) @function
-(id_name) @attribute
+(id_name) @label
 (namespace_name) @namespace
 (property_name) @variable.other.member
 
--- a/runtime/queries/css/highlights.scm
+++ b/runtime/queries/css/highlights.scm
@@ -1,64 +1,88 @@
 (comment) @comment
 
-(tag_name) @tag
-(nesting_selector) @tag
-(universal_selector) @tag
+[
+ "~"
+ ">"
+ "+"
+ "-"
+ "*"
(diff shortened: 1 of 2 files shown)


runtime/queries/css/highlights.scm
```diff
@@ -1,64 +1,86 @@
...
+] @punctuation
 
 (string_value) @string
+((color_value) "#") @string.special
 (color_value) @string.special
 
 (integer_value) @constant.numeric.integer
 (float_value) @constant.numeric.float
-(unit) @type
+(unit) @constant.numeric
```
the-mikedavis, line 68: Why the change here? `unit` is for types like  "rem" in `1.6rem`

Chickenkeeper: The idea was that with the highlighting of many other languages any number type suffixes and prefixes adopt the same colour as the number they're joined to, e.g. `123u`, `0.5f`, `43m` (C#), `23.4f32`, `16u8` (Rust) etc. It's also how CSS numbers are highlighted in some other editors, like VSCode. Different colours without whitespace between them looks a bit jarring to me too, but that's subjective

the-mikedavis: Hmm yeah that does make sense. I don't feel very strongly about this. It seems like some highlighters do diverge from the normal and use a different color for units like tree-sitter-css repo's highlights (https://github.com/tree-sitter/tree-sitter-css/blob/769203d0f9abe1a9a691ac2b9fe4bb4397a73c51/queries/highlights.scm#L60) or nvim-treesitter (https://github.com/nvim-treesitter/nvim-treesitter/blob/1656f67f261ab51525288d0c4b38e1577f79eaf3/queries/css/highlights.scm#L66) or github (which is probably using tree-sitter-css) but most use the same color as the value (kakoune, vim, emacs, w3c docs, mdn docs).

Let's go with `constant.numeric` 👍

Chickenkeeper: Just for clarification, are you suggesting we should allow the unit to follow the colour of the number it's joined to, in which case the `(unit)` tag should be omitted entirely, or that it should be set to `constant.numeric` so that it could be a different colour if people want it to be?

the-mikedavis: Oh, I missed that omitting the `unit` stanza highlights it following the number. This looks good then 👍

the-mikedavis (approved):

David-Else:
I just tried CSS with the latest master and immediately the syntax highlighting was broken:
```
theme = "dark_plus"
```
```
body {
  background-color: blue;
}

h1 {
  background–color: purple;
}
```

Screenshot from 2022-12-03 12-29-37

the-mikedavis:
I think that is not related: the dash in the h1 block is an en-dash U+2013 but the other dashes are the hyphen minus U+002D. The highlighting looks slightly different before this PR but in both commits the syntax tree has an ERROR node for that character

Issue #3725, opened 2022-09-06
C-backspace = "delete_word_backward" doesn't work, should be fixed in Crossterm 0.25


David-Else:
### Summary

In my config:
```
C-backspace = "delete_word_backward"
```
Does nothing in Kitty and Gnome Terminal. Crossterm 0.25 added functional key codes from the Kitty keyboard protocol https://github.com/crossterm-rs/crossterm/releases/tag/0.25 , so it should at least work for Kitty. @groves I think they need to be enabled?

(Works in Neovim with `vim.keymap.set('i', '<C-H>', '<C-W>')`)

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

22.08.1

archseer:
Should be easy to do:
```rust
use std::io::{Write, stdout};
use crossterm::execute;
use crossterm::event::{
    KeyboardEnhancementFlags,
    PushKeyboardEnhancementFlags,
    PopKeyboardEnhancementFlags
};

let mut stdout = stdout();

execute!(
    stdout,
    PushKeyboardEnhancementFlags(
        KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES
    )
);

// ...

execute!(stdout, PopKeyboardEnhancementFlags);
```

But not sure if crossterm does all the detection needed for progressive enhancement: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#detection-of-support-for-this-protocol

David-Else:
@archseer Great news! Afraid I have only just began my Rust journey, could you implement it? :)

PR #4887, opened 2022-11-25
Handle disambiguated keycodes

### LINKED ISSUE #4125: Panic on XF86 keycodes

the-mikedavis:
Media keys are sent despite `DISAMBIGUATE_ESCAPE_CODES` being unset. Previously we panicked on these. This change translates the disambiguated keys from crossterm so that they do not cause a panic.

Closes #4125.
Supersedes #4576.
Closes #4576.

### DIFF
--- a/helix-view/src/keyboard.rs
+++ b/helix-view/src/keyboard.rs
@@ -53,6 +53,164 @@ impl From<crossterm::event::KeyModifiers> for KeyModifiers {
     }
 }
 
+/// Represents a media key (as part of [`KeyCode::Media`]).
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum MediaKeyCode {
+    /// Play media key.
+    Play,
+    /// Pause media key.
+    Pause,
+    /// Play/Pause media key.
+    PlayPause,
+    /// Reverse media key.
+    Reverse,
+    /// Stop media key.
+    Stop,
+    /// Fast-forward media key.
+    FastForward,
+    /// Rewind media key.
+    Rewind,
+    /// Next-track media key.
+    TrackNext,
+    /// Previous-track media key.
+    TrackPrevious,
+    /// Record media key.
+    Record,
+    /// Lower-volume media key.
+    LowerVolume,
+    /// Raise-volume media key.
+    RaiseVolume,
+    /// Mute media key.
+    MuteVolume,
+}
+
+#[cfg(feature = "term")]
+impl From<MediaKeyCode> for crossterm::event::MediaKeyCode {
+    fn from(media_key_code: MediaKeyCode) -> Self {
+        use crossterm::event::MediaKeyCode as CMediaKeyCode;
+
+        match media_key_code {
+            MediaKeyCode::Play => CMediaKeyCode::Play,
+            MediaKeyCode::Pause => CMediaKeyCode::Pause,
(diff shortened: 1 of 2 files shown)


helix-view/src/keyboard.rs
```diff
@@ -53,6 +53,164 @@ impl From<crossterm::event::KeyModifiers> for KeyModifiers {
...
+            CMediaKeyCode::LowerVolume => MediaKeyCode::LowerVolume,
+            CMediaKeyCode::RaiseVolume => MediaKeyCode::RaiseVolume,
+            CMediaKeyCode::MuteVolume => MediaKeyCode::MuteVolume,
+        }
+    }
+}
+
+/// Represents a media key (as part of [`KeyCode::Modifier`]).
+#[derive(Debug, PartialOrd, Ord, PartialEq, Eq, Clone, Copy, Hash)]
+pub enum ModifierKeyCode {
```
the-mikedavis, line 135: Modifier keys are controlled with a different flag (see [crossterm's event-read example](https://github.com/crossterm-rs/crossterm/blob/b13e8ef64d57a3b9a642194b6beb069b55d6ef43/examples/event-read.rs#L77-L81)) and we haven't seen reports of terminals sending these without the flag, but I have added them here for completeness.

dead10ck (approved):
I haven't exhaustively reviewed all the key codes, but this generally looks good to me! Am I understanding that this will also close #3725 as well? i.e. does this generally add the ability to use every key combo with the kitty protocol extension?

the-mikedavis:
Not quite yet - we also need to set the `DISAMBIGUATE_ESCAPE_CODES` flag to fix that. We'll need to figure out how to do that in a way that works with terminals that don't support the flag and do some testing to make sure it doesn't break compatiblity with older terminals

dead10ck:
We could always start simple and make the whole feature opt-in.

archseer:
I want to avoid temporary config flags that we later remove though. The detection is described here: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#detection-of-support-for-this-protocol

dead10ck:
For sure, but would it be temporary? I think it makes sense to keep a flag that can turn off the feature if desired.

the-mikedavis:
It took a look at the detection; I'll make a PR to crossterm. We need to solve it upstream since it needs to use some of crossterm's private functions

archseer:
> For sure, but would it be temporary? I think it makes sense to keep a flag that can turn off the feature if desired.

This transparently improves key parsing to disambiguate things like `tab` vs `ctrl-i`. I can't imagine you'd ever want to turn this off. If you'd like to map something to both keys, then you can just define it twice in your config.

PR #4888, opened 2022-11-25
Remove selections for closed views on all documents

### LINKED ISSUE #4878: Panic when calling `:reload-all` (sometimes)
### Summary

Sometimes when I run `:reload-all` I get:

```
thread 'main' panicked at 'invalid HopSlotMap key used', helix-view/src/tree.rs:300:20
```

This doesn't /always/ happen.

### Reproduction Steps

So far, all I know is that I open a bunch of files and at some point call `:reload-all`. This causes a panic.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

Konsole

### Helix Version

e6dad960cf77a3a0fae92ee216d31c9dae59b0ec

the-mikedavis:
Previously we removed selections for a closed view on only the currently focused document. A view might have selections in other documents though, so the view needs to be removed from all documents.

Closes #4878

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -1116,9 +1116,10 @@ impl Editor {
     }
 
     pub fn close(&mut self, id: ViewId) {
-        let (_view, doc) = current!(self);
-        // remove selection
-        doc.remove_view(id);
+        // Remove selections for the closed view on all documents.
+        for doc in self.documents_mut() {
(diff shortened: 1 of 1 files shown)


mangas (approved):

mangas:
This works for me as well

dead10ck (approved):
Looks good to me

PR #4889, opened 2022-11-25
avoid reloading invalid views

mangas:
Closes https://github.com/helix-editor/helix/issues/4878

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1049,20 +1049,28 @@ fn reload_all(
     let docs_view_ids: Vec<(DocumentId, Vec<ViewId>)> = cx
         .editor
         .documents_mut()
-        .map(|doc| {
-            let mut view_ids: Vec<_> = doc.selections().keys().cloned().collect();
-
-            if view_ids.is_empty() {
-                doc.ensure_view_init(view_id);
-                view_ids.push(view_id);
-            };
-
-            (doc.id(), view_ids)
-        })
+        .map(|doc| (doc.id(), doc.selections().keys().cloned().collect()))
         .collect();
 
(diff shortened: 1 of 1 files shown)


dead10ck (commented):
Would #4888 make this obsolete?

mangas:
> Would #4888 make this obsolete?

Yes and no, it would not be needed but having this effectively prevents a regression if changes are made to the close mechanism or view map assumptions.

helix-term/src/commands/typed.rs
```diff
@@ -1049,20 +1049,28 @@ fn reload_all(
...
+        // contains cant be called with a reference.
+        #[allow(clippy::unnecessary_filter_map)]
+        let mut view_ids: Vec<ViewId> = view_ids
+            .into_iter()
+            // TODO: This check can be removed once
+            // https://github.com/helix-editor/helix/pull/4888 is merged.
+            .filter_map(|id| match cx.editor.tree.contains(id) {
+                true => Some(id),
+                false => None,
+            })
```
the-mikedavis, line 1066: I think clippy is saying here that this could be `filter(|id| cx.editor.tree.contains(*id))` rather than a `filter_map`

the-mikedavis (commented):
I think we should avoid being defensive here: by being defensive we might allow broken assumptions to cause hard-to-debug quirks

helix-term/src/commands/typed.rs
mangas: > I think we should avoid being defensive here: by being defensive we might allow broken assumptions to cause hard-to-debug quirks

No worries, I wasnt sure I large a fix the close stuff would be so decided to check this route as well. I'll close the PR

PR #4892, opened 2022-11-26
[PROPOSAL] feat(theming): separate highlight for function declarations

matoous:
Add separate highlighting scope for function (and method) declarations that allows styling the declarations differently from their calls. I noticed that in some editors the function declaration are styled differently from function calls which wasn't possible with the current highlight scopes.

If this gets the green light I will take a look at more languages and if they could be adjusted as well (although the scope of languages I am familiar with is quite limited).

Open question: do we also want to be able to style function and method declarations differently? Maybe adding also `function.declaration.method` would work.

Showcase:

Screenshot 2022-11-26 at 09 12 21
Screenshot 2022-11-26 at 09 12 42


archseer:
Sublime/TextMate use `entity.name.function`:

> Function and method names should be scoped using the following, but only when they are being invoked. When defined, they should use entity.name.function.
>
>  -  variable.function

https://www.sublimetext.com/docs/scope_naming.html#entity

WDYT @the-mikedavis should we introduce the `entity` group (needs to be documented in the book too)?

archseer:
Actually, a better option would be `@function` for declarations, `@variable.function` for invocations. Would need a cleanup in themes though and it might break existing themes :/

the-mikedavis:
Yeah I think I would also prefer to change the function invocation scope instead of the definition scope. Changing the definition scope is tricky because a function definition could also be a method so you would need some way to combine those scopes

matoous:
@archseer should this maybe rather be `variable.other.function` given we currently have `variable.other.member` for fields of composite datatypes?

matoous:
So I tried to experiment with the approach above but the issue I am hitting is: I want to distinguish _function/method declarations_ and _function/method calls_. With the approach where new scope is added (e.g. `variable.other.function`, `variable.other.method`, or just `variable.method`) I can only distinguish _method calls_ but regular function calls (or built-in function calls) are highlighted the same as declarations which brings me to following:

Keeping both function declaration and calls under stingle scope makes things very verbose if we want to add the ability to distinguish between those two, e.g. given:
```
- `function` - Function and method declarations.
  - `special` (preprocessor in C)
  - `builtin` - Built-in functions (e.q. `println!` in Rust)
  - `method` - Methods
  - `macro`
```
if I wanted to distinguish declaration from call I would have to add:
```
- `function` - Function and method declarations.
  - `call` - Function call
  - `special` (preprocessor in C)
    - `call`
  - `builtin` - Built-in functions (e.q. `println!` in Rust)     <--- this is implicitly a call
  - `method` - Methods
    - `call` - Method call
  - `macro`
    - `call` - Macro call
```
which is imho sub-optimal as it mixes declarations and calls breaking the inheritence benefits.

Thus I wonder if following proposal would make sense: we add 2 scopes under function: `declaration` and `call` (the terms themselves are subject to change, _tis just a suggestion_) to distinguish declarations and calls and both can branch further to distinguish the function/method types, e.g.:

```
- `function` - Functions and methods.
  - `call` - Function and method calls
    - `special` (preprocessor in C)
    - `builtin` - Built-in functions (e.q. `println!` in Rust)
    - `method` - Methods
    - `macro`
  - `declaration` - Declaration
    - `method` - Methods
    - `macro`
```

Theoretically, this could also be extended to remove `function.call.method` and have this scope under `variable.other.function` or `variable.method` instead. This will still most likely require quite a lot of manual work to ensure that all themes keep working as intended (can do that if this gets the green light) but I believe it could serve us better long term. My only concern is that this might be overcomplicated as I am not sure how far we want to go with the complexity of the highlighting scopes that we offer.

archseer:
`variable.other.function` = I'd rather use `variable.function` to match old Sublime scopes

nvim-treesitter uses `@function` and `@function.call` but I agree these are a bit too close

I think you have a point with `function.declaration.method` etc but I think it's ok to make the theme author have to style the two scopes by hand since this is a more advanced theming that a lot of themes won't deal with in the first place.

My proposal:
- `function`/`function.{method, builtin}` should be definitions/declarations
- `function.invoke`/`.call` (or just `variable.function` to match existing Sublime/TextMate/VSCode grammars) should be any function or method invocation. It doesn't make sense to distinguish these, and we often can't on the grammar level anyway without a `locals.scm` query too which should cover this case

the-mikedavis:
> We are now in the process of (incrementally) moving things that work well into core -- which is a chance for changing things, so this is now a good time to have this discussion. (Tree-sitter integration in Neovim is still marked as experimental, so we are free to make breaking changes for good reasons -- and sharing queries would definitely be one!)

Ah excellent! We try to minimize breaking changes in general in Helix but I think we can afford to make some changes on our end for the sake of compatibility. We always end up with a handful of breaking changes per release anyways. Better compatibility of queries, themes and approaches to tree-sitter features seems to me like a win for everybody 🙂

> I haven't much used it myself, but my impression is that the system here (not in Neovim core!) is not working too well, at least for languages like Python. If Helix's works better, we should take a look at it and see if we can align when we implement it in core. If nothing else, indents and folds should be writable in an editor-agnostic way.

I think that the Python indentations have some rough edges in Helix as well although I don't typically write Python myself. There are also some other improvements we want to make to our indentation like outdenting automatically. It seems like potentially a lot of work but maybe we can collaborate on an indentation approach that we can share?

(\cc @Triton171 who authored most of the indentation code for Helix)

> we concatenate all queries on runtime path so users can override individual queries in their own config. Does Helix support extending/overriding queries like that?

We have support for `; inherits: <lang1>,<lang2>,...,<langN>` but that's mostly used for cases like C++ extending C's highlights. We don't have a way for users to override or extend the built-in queries with custom queries currently. I think a feature like that might be tricky to design because sometimes you want to disable a specific pattern or remove chunks of the query. We have https://github.com/helix-editor/helix/issues/3346 tracking this and currently we're thinking of a `$PATH`-like solution but queries or parsers in user directories would replace the ones from installation rather than merging.

> I personally don't mind switching (mostly) to Helix' scheme, which I believe aligns (more) closely with Atom/TextMate naming? If you were ready to adapt some captures to ours (for example, I prefer the way we do LaTeX `math` captures), I'm sure we could come up with a joint document that upstream parser devs could also rely on (e.g., living in the tree-sitter Wiki).

Yep we try to adhere to textmate when possible. I'm definitely open to changing up some scopes so our queries and themes can be closer, especially some of the lesser-used ones like `math`. Our captures aren't set in stone, we do like to revise them when there's an opportunity for improvements (for example https://github.com/helix-editor/helix/pull/4892). It would be really cool to be able to trivially convert Neovim queries and themes to Helix queries and themes and vice versa!

We also use the fallbacks for captures extensively. I think our systems work the same way: `@constant.numeric.integer` falls back to `@constant.numeric` and then to `@constant`.

> we could discuss this and other things on Matrix

Matrix would be perfect - we use Matrix for a lot of Helix discussion as well. If you make a room, would you mind inviting me (`@the-mikedavis:matrix.org`) and @archseer (`@speed:matrix.org`)?

> another point of divergence is predicates and directives. It's clear that other editors won't support `#lua-match?` and that tree-sitter itself will probably not add too many, but do you think at least the two of us could agree on a "greatest common denominator" list both Neovim and Helix support?

Yeah the regular expression ones seem like a hard problem to solve very consistently between editors. Our `#match?` just compiles the string with the rust-lang/regex crate. I think it might be an anti-pattern to use fancy regular expression features in queries anyways so maybe the differences in engines won't really matter in practice.

Coming up with a list we can both support sounds do-able 👍. There are some from here that we already want to support like `#any-of?` that would help a lot with compatibility. Adding new predicates and properties isn't totally trivial in Helix but it's not terribly difficult either.

One other compatibility point - does neovim support combined injections (the `(#set! injection.combined)` property)? I thought that these were disabled in nvim because of some bugs but I am probably out of date on this. We use them mostly for templating languages like erb/ejs or markup like markdown so that you can combine injected ranges into a single document. Not so many languages use this though so it's probably not a huge deal.

slugbyte:
I think it might be preferable for all function definition queries  to be given a optional `.call` extension. This would allow theme designers to have maximum control and it would not break exiting themes.

`function` - function deceleration or call
`function.call` - function call overwrite
`function.method` - method deceleration or call
`function.method.call` - method call overwrite
`function.builtin` - builtin call (builtins cannot be defined)

``` scm
(builtin_identifier) @function.builtin

(call_expression
  function: (identifier) @function.call)

(call_expression
  function: (field_expression
    member: (identifier) @function.method.call))

(method_declaration
  name: (identifier) @function.method)

(function_declaration
  name: (identifier) @function)
```

but i also like @archseer proposal. which is what i ended up adding for my personal config (but the language i am using does not have methods)

PR #4896, opened 2022-11-26
Optimizes View::screen_coords_at_pos by not computing char_to_line two times

zummenix:
This solves a todo  in `View::screen_coords_at_pos` method to not compute `char_to_line` two times. 

I've introduced a new function `core::visual_col_position` which additionally accepts a row index. Also I've made corresponding changes in `core::position` module to remove some code duplication.

### DIFF
--- a/helix-core/src/position.rs
+++ b/helix-core/src/position.rs
@@ -54,18 +54,15 @@ impl From<Position> for tree_sitter::Point {
         Self::new(pos.row, pos.col)
     }
 }
+
 /// Convert a character index to (line, column) coordinates.
 ///
 /// column in `char` count which can be used for row:column display in
 /// status line. See [`visual_coords_at_pos`] for a visual one.
 pub fn coords_at_pos(text: RopeSlice, pos: usize) -> Position {
-    let line = text.char_to_line(pos);
-
-    let line_start = text.line_to_char(line);
-    let pos = ensure_grapheme_boundary_prev(text, pos);
-    let col = RopeGraphemes::new(text.slice(line_start..pos)).count();
-
-    Position::new(line, col)
+    let row = text.char_to_line(pos);
+    let col = rope_graphemes(text, pos, row).count();
+    Position::new(row, col)
 }
 
 /// Convert a character index to (line, column) coordinates visually.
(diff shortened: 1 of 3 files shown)


helix-core/src/position.rs
```diff
@@ -169,6 +168,12 @@ pub fn pos_at_visual_coords(text: RopeSlice, coords: Position, tab_width: usize)
     line_start + col_char_offset
 }
 
+fn rope_graphemes(text: RopeSlice, pos: usize, row: usize) -> RopeGraphemes {
+    let row_start = text.line_to_char(row);
+    let pos = ensure_grapheme_boundary_prev(text, pos);
+    RopeGraphemes::new(text.slice(row_start..pos))
+}
```
the-mikedavis, line 175: This is not so much code that it needs a helper function IMO, and it's a bit confusing that there are other calls to RopeGraphemes::new that don't use rope_graphemes in this module. This can be inlined into the two callsites

helix-core/src/position.rs
```diff
@@ -74,23 +71,25 @@ pub fn coords_at_pos(text: RopeSlice, pos: usize) -> Position {
 /// not in the document in the future.
 /// See [`coords_at_pos`] for an "objective" one.
 pub fn visual_coords_at_pos(text: RopeSlice, pos: usize, tab_width: usize) -> Position {
-    let line = text.char_to_line(pos);
-
```
the-mikedavis, line 78: This is a small stylistic nit but it's more common in the codebase to see `line` when translating a position to a line number (rather than `row`)

zummenix: Maybe this is just a language barrier but for me it was confusing because there is `row` in `Position`. In any case I'm fine with any name.

helix-core/src/position.rs
zummenix: Respectfully, I disagree. This function encapsulates an important concept that `text` should be sliced taking into account grapheme boundaries. Maybe the name for the function and the place should be different but otherwise...

I don't know why in other places that's not the case, haven't looked deep.

Let's close this PR because neither I nor you don't have time to argue.

pascalkuthe:
Just a small heads: As Part of #5008 I will be replacing entire positioning code  (including this function) to take virtual text, softwrapping and other kinds of decoration into account. Even if my specific PR doesn't get merged. Any implementation of these features will have to replace this code. A  quick-fix could be nice but it's probably not worth investing a lot of time into optimizing a functiion that will be replaced in the future anyway

zummenix:
Good to know. Thanks! Closing

Issue #4884, opened 2022-11-25
Cursor jumps on `:reload-all`


mchlstckl:
### Summary

When I do `:reload-all`, the content of the window jump. Sometimes the cursor is positioned at the top of what looks like a blank buffer, but scrolling up reveals the contents of the file.

Note that the file I'm looking at has not changed.

### Reproduction Steps

 

I tried this:

1. `hx`
2. Open some file
3. Do `:reload-all`

I expected this to happen:

Contents to be reloaded from disk. Cursor remains in current position.

Instead, this happened:

Contents reloaded from disk. Cursor jumps to random(?) position.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

macOS

### Terminal Emulator

kitty

### Helix Version

helix 22.08.1 (e6dad960)

David-Else:
@mangas  it seems like `reload-all` has problems as it is reported to panic too https://github.com/helix-editor/helix/issues/4878 , could you take a look?

mangas:
I am not able to replicate this. Can you maybe create an example with a few files where this happens? Or maybe a repo that you were using?

mchlstckl:
This happens when I edit Go files. 

I use `gopls`, and `golines` for formatting. 

I use Helix at work so I don’t currently have any example files or repos to share. I’ll try putting something together that triggers this that I can share.

mangas:
awesome, thanks !

mchlstckl:
I checked out https://github.com/GoogleCloudPlatform/cloud-sql-proxy repo in order to reproduce this bug.

1. `hx`
2. Open `cmd/root.go`
3. Split window into left / right
4. Open `internal/proxy/proxy.go` in right window
5. Move cursor in right window line 35
6. `:reload-all`
7. Contents of right window shift up so that line 35 is now 6 lines from the top of the window

Where to move the cursor in step 5. and what happens at step 7. seem a bit random to me.

I tested this in kitty and iTerm2 and I get the same behaviour in both.

Let me know if you need any more info.

*config.toml*
```
theme =  "onelight"

[editor]
rulers = [100]
cursorline = true
color-modes = true

[editor.file-picker]
hidden = false

[editor.indent-guides]
render = true

[editor.cursor-shape]
insert = "bar"
normal = "block"
select = "underline"

[editor.lsp]
[...]

PR #4901, opened 2022-11-26
reload-all: Only update viewport when view focuses on the doc

mangas:
May or may not fix https://github.com/helix-editor/helix/issues/4884

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1070,8 +1070,9 @@ fn reload_all(
 
         for view_id in view_ids {
             let view = view_mut!(cx.editor, view_id);
-
-            view.ensure_cursor_in_view(doc, scrolloff);
+            if view.doc.eq(&doc_id) {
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1070,6 +1070,9 @@ fn reload_all(
 
         for view_id in view_ids {
             let view = view_mut!(cx.editor, view_id);
+            if !view.doc.eq(&doc_id) {
+                continue;
+            }
 
             view.ensure_cursor_in_view(doc, scrolloff);
```
the-mikedavis, line 1077: I think it's a little clearer to check for the positive case:

```rust
if view.doc == doc_id {
}
```

mangas: I usually prefer returning early and perform any actions afterwards but I don't mind using your suggestion instead.

mangas: done

the-mikedavis (approved):

PR #4904, opened 2022-11-26
feat(lsp): pass client_info on initialization

matoous:
Pass client name ('helix') and client version (version / git hash) to LSP server on initialization.


helix-lsp/build.rs
```diff
@@ -0,0 +1,22 @@
...
+        .args(["rev-parse", "HEAD"])
+        .output()
+        .ok()
+        .filter(|output| output.status.success())
+        .and_then(|x| String::from_utf8(x.stdout).ok());
+
+    let version: Cow<_> = match git_hash {
+        Some(git_hash) => format!("{} ({})", VERSION, &git_hash[..8]).into(),
+        None => VERSION.into(),
+    };
```
the-mikedavis, line 17: It's a small block to re-use but I wonder if it should be a constant in `helix-loader` so it can be re-used in `helix-lsp` and `helix-term`

matoous: Done, also not sure if it maybe would be better to expose `GIT_SHA` on it's own.

matoous: @the-mikedavis I tried it but it makes it very verbose in around the 3 call-sides we have. Until we have a use case for `GIT_HASH` on it's own I would keep it this way.

the-mikedavis (approved):

Issue #361, opened 2021-06-23
Doom emacs-like `space space` directory navigator


CBenoit:
> The other idea I have is sort of `space space` in doom emacs which we can navigate between directories easily and show only the current directory so that is at least easier compared to `:e xxx/<tab>`.

_Originally posted by @pickfire in https://github.com/helix-editor/helix/pull/336#issuecomment-866448160_

This is useful when the name is not known but we have an idea of it's location.

Mapping would probably be something else than `space space`.

CBenoit:
> EDIT by @pickfire: `space space` in doom emacs is `space f` in helix now. So `space f` in doom emacs we need to figure out a key for it, I wish we could use `space space` for something so common.

@pickfire I think you could remap that?

gbaranski:
I personally use space-space for switching to last file.

pickfire:
> @pickfire I think you could remap that?

Yes, I could. But I wonder if the current `space space` is very much used that we need to make it `space space` compared to the one in doom emacs, or is there any other keys that fit it better, like escape or backticks.

Cons-Cat:
Personally I find `space p f` for searching within the current project generally more useful than `space space`

PR #4905, opened 2022-11-26
Select current file in file picker

mangas:
An attempt at implementing https://github.com/helix-editor/helix/issues/3014

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -112,9 +112,11 @@ impl<T: Item> FilePicker<T> {
         editor_data: T::Data,
         callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
         preview_fn: impl Fn(&Editor, &T) -> Option<FileLocation> + 'static,
+        initial_cursor: Option<T>,
     ) -> Self {
         let truncate_start = true;
-        let mut picker = Picker::new(options, editor_data, callback_fn);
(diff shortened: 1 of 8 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -112,9 +112,11 @@ impl<T: Item> FilePicker<T> {
         editor_data: T::Data,
         callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
         preview_fn: impl Fn(&Editor, &T) -> Option<FileLocation> + 'static,
+        initial_cursor: Option<T>,
```
dead10ck, line 115: Should this be called `initial_item` instead?

dead10ck (approved):
One minor comment, but otherwise I think this looks reasonable. It is generic enough that it allows for picking a different default in the other pickers as well in the future, should that be desirable.

I considered whether the default item member of the `Picker` should be a closure, rather than a concrete value, but I think in all the contexts where pickers are made, they should be able to decide at the time of that picker's construction which actual concrete value to choose as the default.

I haven't tested this locally yet, but the code seems good.

pascalkuthe:
closing this PR as stale. Ultimaetly I don't think this is the direction we would be going with the picker. Sorting is not possible anymore with a streaming picker and imo this usecase is better covered by `<space>F` or something like #361 


Thank you for contributing!

PR #4908, opened 2022-11-27
feat(keys): allow jumping between the splits with Ctrl-arrow

sigmaSd:
I use mostly arrow keys, so I prefer feature parity with `hjkl`, I think there are some other instances where this is not the case.

If its not a deliberate decision to not add those, I can pr what I find


archseer:
This breaks existing mappings. You should add them in your user config instead

Issue #4423, opened 2022-10-22
Custom typable commands


the-mikedavis:
Regular commands can be rebound but typable commands (anything entered in command mode, `:`) can't be modified in any way.

You might use custom typable commands to implement file operations in conjunction with https://github.com/helix-editor/helix/issues/3134 for example. You might define `:mv` as `:sh mv $file $1` or `:rm` as `:sh rm $1`. (The syntax to use for variables, or whether to use variables at all should be discussed.)

Custom typable commands could also be used to create custom abbreviations for commands.

rkshthrmsh:
Hi @the-mikedavis, would this also cover command descriptions? For example, I have remapped `gs` to extend selection till the first non-blank character in line. This is in line with the discussion in #1215.

```toml
[keys.normal]
g = { s = ["select_mode", "goto_first_nonwhitespace", "exit_select_mode"] }
```

However, since there is no custom description for commands (that I'm aware of), the goto menu displays a rather unhelpful prompt `[Multiple commands]`:
<img width="346" alt="Screenshot 2023-04-14 at 4 07 27 PM" src="https://user-images.githubusercontent.com/16115390/232145054-f66f4e1d-4b38-423c-a41a-16b216caaa9a.png">

the-mikedavis:
No that's https://github.com/helix-editor/helix/issues/1752

kendfss:
Does the editor already support custom commands and just not typables?

I was thinking it would be nice to have something like:
```toml
[custom.commands]
bring_line_up = {
  cmds = [ 'push_register', 'extend_line', 'delete_selection', 'move_line_up', 'paste_before', 'pop_register' ] # semicolon separates operations' names from parameters' arguments. 
  desc = "swap line with above"
  mode = editor || prompt || both
}
```

kendfss:
or set the mode via `-` vs `_`. I think that would be a more cumbersome implementation to maintain though.

the-mikedavis:
You can already map keys to lists of commands (https://docs.helix-editor.com/master/remapping.html) to create custom non-typable commands. You can also rebind the `normal_mode`, `insert_mode` and `select_mode` commands to `-` (`minus` when configuring this in `config.toml`) and/or `_`. This issue is only about creating typable commands: ones entered in command mode with `:` that optionally take arguments.

kendfss:
Could separate func name from params with `;`
[...]

PR #1288, opened 2021-12-18
feat(commands): sort command

matoous:
Add basic implementation of sort command.

This PR adds `sort` and `sort!` (reverse sort) commands that sort _selections_. This can be used to not only sort data on separate lines (select -> split on new line `Alt-s` -> `:sort`) but also any arbitrary list (select -> select all matches, for example numbers `\d+` -> `:sort`).

Additional functionality can be added in the future such as removing duplicates, case insensitivity, etc.

---

edit: provide better description for the PR

PR #4909, opened 2022-11-27
Add :write-without-auto-format commands

### LINKED ISSUE #4853: Write without formatting
Would be nice to have a command palette item to write a file without auto formatting first.

Sometimes you need to edit a "hand crafted" file and need to skip formatting, but want auto-formatting to be the behavior 99.999% of the time.

zummenix:
This adds two commands to write and force write without auto format.

Closes #4853

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -268,13 +268,13 @@ fn write_impl(
     cx: &mut compositor::Context,
     path: Option<&Cow<str>>,
     force: bool,
+    auto_format: bool,
 ) -> anyhow::Result<()> {
-    let editor_auto_fmt = cx.editor.config().auto_format;
     let jobs = &mut cx.jobs;
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
 
-    let fmt = if editor_auto_fmt {
+    let fmt = if auto_format {
         doc.auto_format().map(|fmt| {
             let callback = make_format_callback(
                 doc.id(),
@@ -307,7 +307,19 @@ fn write(
         return Ok(());
     }
 
-    write_impl(cx, args.first(), false)
+    write_impl(cx, args.first(), false, cx.editor.config().auto_format)
+}
+
+fn write_without_auto_format(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    write_impl(cx, args.first(), false, false)
 }
 
 fn force_write(
@@ -319,7 +331,19 @@ fn force_write(
         return Ok(());
     }
 
-    write_impl(cx, args.first(), true)
+    write_impl(cx, args.first(), true, cx.editor.config().auto_format)
+}
+
+fn force_write_without_auto_format(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    write_impl(cx, args.first(), true, false)
 }
 
 fn new_file(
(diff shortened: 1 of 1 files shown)


dead10ck (commented):
Implementation wise this looks fine, though I think the command names are very long. I also wonder if it wouldn't make sense to make this an argument of the command using something like a CLI switch. I think kakoune does this if I'm not mistaken. What do you think @archseer @the-mikedavis?

the-mikedavis:
Yeah I like kakoune's flags for this. It could be done as `:write --no-fmt`. That would mean we couldn't write a file named `--no-fmt` but we could also introduce `--` to separate flags from literal arguments. (`:write --no-fmt -- --no-fmt` writes a file `--no-fmt` without formatting.)

zummenix:
Personally, I don't have a problem with the length of the commands, because the prompt autocomplete works really nice.

Are there precedents of commands that accept args/flags as proposed (I don't see any)?
What other commands that could benefit from supporting args/flags in them?
How args/flags should be displayed in prompt autocomplete (UX/UI)?
How args/flags should be documented (discoverability story)?

the-mikedavis:
> Are there precedents of commands that accept args/flags as proposed (I don't see any)?

Not yet; this would be a first. But other commands could benefit from it (existing commands and future commands probably)

> What other commands that could benefit from supporting args/flags in them?

See https://github.com/helix-editor/helix/pull/1288#discussion_r775329454, sort/rsort could be turned into one command.

> How args/flags should be displayed in prompt autocomplete (UX/UI)?

A completer could provide them as options

> How args/flags should be documented (discoverability story)?

Kakoune has a nice way of showing these:

flags

dead10ck:
It's very possible switches are out of scope for this PR (at least in a general way) so consider the following somewhat off topic, but I have thought it might make sense to pull in a CLI arg parsing lib to make use of in commands. I know there was resistance to this in the past for the main binary, but I think it would be a stronger justification to pull in a lib if it were used throughout the command palette.

zummenix:
> A completer could provide them as options

`:write` accepts optional file name. I think there should be a mechanism to show switches in completer menu so they look distinct from file names.

> Kakoune has a nice way of showing these:

I like it!

> It's very possible switches are out of scope for this PR

Definitely :)

In overall I'm still on the fence about switches in commands but I see how they could be useful. I suggest you to open separate issues/discussions to iron out details how they should look like in UI and to build a vision for the feature as a whole. Also I'm thinking if there will be some kind of scripting language in the future, will it cover this usecase? For example, we could have a method `write` with parameters: `:eval 'write(force: true, auto-format: false, method: .replace)'`

The question about this PR: do we close it or I can continue with it to fix the CI error to be ready for the merge?

dead10ck:
I'd like to see these commands merged, but I'd wait for @archseer's input for this PR and what we should do for names and/or switches.

kirawi:
By the way, be sure to run `cargo xtask docgen` and commit the changes.

the-mikedavis:
I don't think it's necessary to add commands for this since there's a workaround: you can `:set auto-format false`, `:write`, and then `:set auto-format true` when you want auto-formatting again.

zummenix:
> By the way, be sure to run cargo xtask docgen and commit the changes.

Yeah, I know, thanks! I was just waiting for the final decision.

lukepighetti:
Disappointed that this didn't get merged. Many editors have this in their command palette and the `:set auto-format false`, `:write`, `:set auto-format true` workflow is not discoverable

kirawi:
With https://github.com/helix-editor/helix/issues/4423, it could be done. However, it would depend on if the maintainers are okay with bundling some command sewuences.

andradei:
@the-mikedavis this to my config, but it doesn't work. I have to run the 3 commands manually...

```toml
[keys.select]
S = [':set auto-format false', ':write', ':set auto-format true']
```

mwyvr:
Was just looking for just this, and discovered as @andradei that the three-command approach wasn't working, auto-format continued.  I don't need it often but when adding a new package to a Go project, lsp complains and removes the import line.

I settled on another route:  set a keybinding to **":toggle-option auto-format"**.

andradei:
@mwyvr That's the workaround I'm using. I think the next version of configuring helix will solve this and much more. Until then, it is what it is.

PR #4931, opened 2022-11-29
Add `:pipe-to` typable command that ignores shell output

alxshine:
I often use an editor in conjunction with a REPL open in another terminal.
To quickly send commands from my editor to the REPL I pipe the editor selection into kitty's remote control feature: `kitty @ send-text --match recent:1 --stdin`.

In kakoune I could set up a keybinding utilizing the `pipe-to` command <Alt-|> and pass the shell command directly.
For helix I can only pass the command to typed arguments, but there is currently no typed command to pipe selections to shell commands while ignoring the output.

My suggestion is this:
Add a `:pipe-to` command which is essentially a copy of the `:pipe` command, but uses `ShellBehavior::Ignore` instead of the hard-coded `ShellBehavior::Replace`.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1739,14 +1739,14 @@ fn pipe_to(
     args: &[Cow<str>],
     event: PromptEvent,
 ) -> anyhow::Result<()> {
-    _pipe(cx, args, event, &ShellBehavior::Ignore)
+    pipe_impl(cx, args, event, &ShellBehavior::Ignore)
 }
 
 fn pipe(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> anyhow::Result<()> {
-    _pipe(cx, args, event, &ShellBehavior::Replace)
+    pipe_impl(cx, args, event, &ShellBehavior::Replace)
 }
 
-fn _pipe(
+fn pipe_impl(
     cx: &mut compositor::Context,
     args: &[Cow<str>],
     event: PromptEvent,
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1734,13 +1734,30 @@ fn insert_output(
     Ok(())
 }
 
+fn pipe_to(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1734,13 +1734,30 @@ fn insert_output(
...
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    _pipe(cx, args, event, &ShellBehavior::Ignore)
+}
+
 fn pipe(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> anyhow::Result<()> {
+    _pipe(cx, args, event, &ShellBehavior::Replace)
+}
+
+fn _pipe(
```
the-mikedavis, line 1749:
```suggestion
fn pipe_impl(
```

is the common pattern in the codebase

QiBaobin:
Shall is this better to implement in async approach as we don't need the result?

alxshine:
I can't reply for some reason, so here's my response:

I tried to keep it close to how the `shell_pipe_to` static command works.
All variants currently internally use the `shell` function to handle the subshell for the command.
The only outlier is the `run_shell_command`, which calls `shell_impl` directly.

If you want the new command to be async, then IMO so should the `shell_pipe_to` static command.
I don't know if it would be better to add an argument to the `shell` function, or add a separate `shell_async` function.
There is an underlying `shell_impl_async` function, which is currently only used wrapped in a tokio barrier, so a change would not be too complex.

If we wanted we could enable async shell commands, but I'm not sure if thats good default behavior.

QiBaobin:
> I can't reply for some reason, so here's my response:
> 
> I tried to keep it close to how the `shell_pipe_to` static command works. All variants currently internally use the `shell` function to handle the subshell for the command. The only outlier is the `run_shell_command`, which calls `shell_impl` directly.
> 
> If you want the new command to be async, then IMO so should the `shell_pipe_to` static command. I don't know if it would be better to add an argument to the `shell` function, or add a separate `shell_async` function. There is an underlying `shell_impl_async` function, which is currently only used wrapped in a tokio barrier, so a change would not be too complex.
> 
> If we wanted we could enable async shell commands, but I'm not sure if thats good default behavior.

Make sense to me, although '&' or `nohup command &` on the command don't  work as we might need wait for the shell completes.

PR #4937, opened 2022-11-29
Add shift-backspace keybind alias for backspace

lesleyrs:
Not sure if this PR works but would be nice to have, no real reason not to I guess?

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -342,7 +342,7 @@ experience.
 | `Alt-d`, `Alt-Delete`                       | Delete next word            | `delete_word_forward`    |
 | `Ctrl-u`                                    | Delete to start of line     | `kill_to_line_start`     |
 | `Ctrl-k`                                    | Delete to end of line       | `kill_to_line_end`       |
-| `Ctrl-h`, `Backspace`                       | Delete previous char        | `delete_char_backward`   |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
By default the terminal can't tell the difference between backspace and S-backspace: all of these commands already work with S-backspace. Terminals that implement the progressive keyboard enhancement protocol can tell the difference but we need to detect and enable that from within Helix. See the [`enable-keycode-disambiguation` branch](https://github.com/helix-editor/helix/tree/enable-keycode-disambiguation). We could merge this change into that branch.

lesleyrs:
Ok sounds good, I'm not sure how to have 2 separate helix installs work so I can't test if I did it correctly. And I don't wanna mess with my current install.

gabydd:
Just FYI enable key code disambiguation was just merged also you can do ``cargo build`` or ``cargo run`` to test it out

the-mikedavis:
#4887 handles the new keys but doesn't enable the disambiguation flags. See https://github.com/helix-editor/helix/pull/4887#issuecomment-1328093166

kirawi:
Misclick.

lesleyrs:
Ok I just did a cargo run and it seems to work as intended in both editor and prompt 👍. Thx for the help guys!

Issue #3210, opened 2022-07-27
Mapping C-/ fails


thor314:
### Summary

The following produces no errors, but fails:
```
[keys.normal]
"C-/" = "toggle_comments" #nope
```

The following produces an error, and fails:
```
[keys.normal]
C-/ = "toggle_comments" #nope
```

The following works as expected:
```
[keys.normal]
"/" = "toggle_comments"
```

Recommended short term action:
make a note in the key-remapping section in docs, until bug can be resolved.

### Helix log

Config file: /home/thor/.config/helix/config.toml
Language file: /home/thor/.config/helix/languages.toml
Log file: /home/thor/.cache/helix/helix.log
Runtime directory: /home/thor/.config/helix/runtime
```
Language           LSP                DAP                Highlight          Textobject         Indent             
bash               ✔ bash-language…   None               ✔                  ✘                  ✘                  
c                  ✘ clangd           ✔ lldb-vscode      ✔                  ✔                  ✔                  
c-sharp            ✘ OmniSharp        None               ✔                  ✘                  ✘                  
cairo              None               None               ✔                  ✘                  ✘                  
clojure            ✘ clojure-lsp      None               ✔                  ✘                  ✘                  
cmake              ✘ cmake-languag…   None               ✔                  ✔                  ✔                  
comment            None               None               ✔                  ✘                  ✘                  
cpon               None               None               ✔                  ✘                  ✔                  
cpp                ✘ clangd           ✔ lldb-vscode      ✔                  ✔                  ✔                  
css                ✘ vscode-css-la…   None               ✔                  ✘                  ✘                  
dart               ✘ dart             None               ✔                  ✘                  ✔                  
devicetree         None               None               ✔                  ✘                  ✔                  
dockerfile         ✘ docker-langse…   None               ✔                  ✘                  ✘                  
edoc               None               None               ✔                  ✘                  ✘                  
eex                None               None               ✔                  ✘                  ✘                  
ejs                None               None               ✔                  ✘                  ✘                  
elixir             ✘ elixir-ls        None               ✔                  ✔                  ✘                  
elm                ✘ elm-language-…   None               ✔                  ✘                  ✘                  
elvish             ✘ elvish           None               ✔                  ✘                  ✘                  
erb                None               None               ✔                  ✘                  ✘                  
erlang             ✘ erlang_ls        None               ✔                  ✔                  ✘                  
fish               None               None               ✔                  ✔                  ✔                  
[...]

Issue #1368, opened 2021-12-25
keybindings with esc do not work (for me)


adsick:
I'm started messing up with Helix config and would like to have a way of quickly saving&escaping everything, and I'd like to perform ```Shift-esc```  for that, but this doesn't work for some reason:
![image](https://user-images.githubusercontent.com/24528839/147388486-dce02b0e-0e11-4283-963b-9295c85e9ccd.png)
when I press Shift+esc it simply does nothing.

It could be Gnome intercepting certain key-combinations or it may be something with Alacritty.
Can anyone test this on your system? I've found that ```C-S-esc``` doesn't work either.

OS: Fedora 35

NNBnh:
Similar to #1369, this is a limitation of the terminal. Try `showkey -a` and press <kbd>Esc</kbd> and <kbd>Shift</kbd>+<kbd>Esc</kbd>, they can't be differentiated and both produce this keycode:

```console
$ showkey -a

Press any keys - Ctrl-D will terminate this program

^[ 	27 0033 0x1b
^[ 	27 0033 0x1b
```

Again ,unless Helix support [Kitty terminal's Keyboard handling protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol), there is nothing we can do to fix this.

adsick:
Yeah, I see. It means the better to input is listening to keyboard events like 'KeyPressed(scancode)' and 'KeyReleased(scancode)', can we do that?

PR #4939, opened 2022-11-29
Enable progressive keyboard enhancement protocol

the-mikedavis:
See the Kitty documentation on this: https://sw.kovidgoyal.net/kitty/keyboard-protocol/

Key events given by the terminal are limited by default: the terminal confuses keycodes like C-i and tab and can't tell the difference between backspace and S-backspace or ret and S-ret. We can detect support for an enhanced keyboard protocol and enable that protocol to be able to disambiguate these keys on terminals which support the protocol.

Closes https://github.com/helix-editor/helix/issues/3725
Closes https://github.com/helix-editor/helix/issues/3210
Closes https://github.com/helix-editor/helix/issues/1085
Closes https://github.com/helix-editor/helix/issues/5765

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -363,7 +363,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "A-d" | "A-del" => delete_word_forward,
         "C-u" => kill_to_line_start,
         "C-k" => kill_to_line_end,
-        "C-h" | "backspace" => delete_char_backward,
+        "C-h" | "backspace" | "S-backspace" => delete_char_backward,
         "C-d" | "del" => delete_char_forward,
         "C-j" | "ret" => insert_newline,
         "tab" => insert_tab,
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -261,9 +261,9 @@ dependencies = [
 
 [[package]]
 name = "crossterm"
-version = "0.25.0"
+version = "0.26.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e64e6c0fbe2c17357405f7c758c1ef960fce08bdfb2c03d88d2a18d7e09c4b67"
+checksum = "a84cda67535339806297f1b331d6dd6320470d2a0fe65381e79ee9e156dd3d13"
 dependencies = [
  "bitflags",
  "crossterm_winapi",
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -350,7 +350,7 @@ experience.
 | `Alt-d`, `Alt-Delete`                       | Delete next word            | `delete_word_forward`    |
 | `Ctrl-u`                                    | Delete to start of line     | `kill_to_line_start`     |
 | `Ctrl-k`                                    | Delete to end of line       | `kill_to_line_end`       |
-| `Ctrl-h`, `Backspace`                       | Delete previous char        | `delete_char_backward`   |
+| `Ctrl-h`, `Backspace`, `Shift-Backspace`    | Delete previous char        | `delete_char_backward`   |
 | `Ctrl-d`, `Delete`                          | Delete next char            | `delete_char_forward`    |
 | `Ctrl-j`, `Enter`                           | Insert new line             | `insert_newline`         |
 
@@ -431,7 +431,7 @@ Keys to use within prompt, Remapping currently not supported.
 | `Alt-d`, `Alt-Delete`, `Ctrl-Delete`        | Delete next word                                                        |
 | `Ctrl-u`                                    | Delete to start of line                                                 |
 | `Ctrl-k`                                    | Delete to end of line                                                   |
-| `Backspace`, `Ctrl-h`                       | Delete previous char                                                    |
+| `Backspace`, `Ctrl-h`, `Shift-Backspace`    | Delete previous char                                                    |
 | `Delete`, `Ctrl-d`                          | Delete next char                                                        |
 | `Ctrl-s`                                    | Insert a word under doc cursor, may be changed to Ctrl-r Ctrl-w later   |
 | `Ctrl-p`, `Up`                              | Select previous history                                                 |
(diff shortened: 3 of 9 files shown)


lesleyrs:
Oh oops I just made https://github.com/helix-editor/helix/pull/4961 that also adds C-i.

Even without progressive keyboard enhancement it's needed for it to work on Windows and it doesn't seem to affect Linux. Can you merge mine first please, as it closes some issues?

the-mikedavis:
The upstream crossterm PR was merged so we just wait on a release now. (Hopefully the next one will also include the PR for https://github.com/helix-editor/helix/pull/5468.)

There's some odd behavior with the key events you get and the shift modifier so there may need to be more changes to the default keymap: https://github.com/the-mikedavis/helix/commit/4be72f72015cfc09a126bfda1af15d82f14c835c

kirawi:
Closes https://github.com/helix-editor/helix/issues/1368

yyogo:
Hey, crossterm just released 0.26, see #5468

the-mikedavis:
I tested out 0.26 and there have been some new changes that caused a regression with the function that checks for the keyboard enhancement protocol. I'll post a PR to fix that so this will need to wait on the next crossterm release

archseer:
Since there's often significant time between releases of crossterm, I wonder if we could fork it to the `helix-editor` org and use that for the time being -- seems a shame to hold this feature back because of a one line fix. We could ask for a patch release (0.26.1) but I don't want to pester the maintainer too much.

the-mikedavis:
Yeah that seems like a nice compromise to me 👍

We could just use this `Cargo.toml` patch that uses crossterm from the master branch but if we forked we would also publish to crates?

archseer:
I think that's enough, even if we fork I'd list it as a git dependency. It's a workaround specific to our use case and I'd avoid someone else depending on it by accident

the-mikedavis:
This is now waiting on another PR to crossterm now that adds support for the "report alternate keys" part of the Kitty Keyboard Protocol: https://sw.kovidgoyal.net/kitty/keyboard-protocol/#report-alternate-keys.

Some keys arrive weirdly when the protocol is enabled: `A-(` arrives as `A-S-9`, `A-)` as `A-S-0` and `A-_` as `A-S-minus`. Kakoune gets around this by enabling the "report alternate keys" part of the spec. [Here](https://github.com/mawww/kakoune/blob/eb0e9831330d3b1e1d3ddb2bc789000706e6e445/src/terminal_ui.cc#L1418) it pushes `5` for keyboard enhancement flags which is equivalent to

```rust
KeyboardEnhancementFlags::DISAMBIGUATE_ESCAPE_CODES |
    KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS
```

in crossterm. Then [here](https://github.com/mawww/kakoune/blob/eb0e9831330d3b1e1d3ddb2bc789000706e6e445/src/terminal_ui.cc#L804-L813) it overwrites the unshifted key with the shifted key.

Cargo.toml
```diff
@@ -31,3 +31,6 @@ inherits = "test"
 package.helix-core.opt-level = 2
 package.helix-tui.opt-level = 2
 package.helix-term.opt-level = 2
+
+[patch.crates-io]
+crossterm = { git = "https://github.com/crossterm-rs/crossterm", rev = "bca71adad70c36de3a4e13d7edc31c0ef9567f47" }
```
pascalkuthe, line 36: I could image that any distro packages that include helix (and want to package dependencies separately)  would have trouble with this/reservations.

Does the regression in crossterm `0.26` affect helix even without this PR? If not then distros could just remove this patch and disable the feature.
Otherwise we might consider waiting with merging this PR until the next release cycle (at which point cross-term will hopefully update)? 

I personally think git dependencies like this are fine but it's just something I wanted to draw attention to.

helix-term/src/keymap/default.rs
```diff
@@ -363,7 +363,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
         "A-d" | "A-del" => delete_word_forward,
         "C-u" => kill_to_line_start,
         "C-k" => kill_to_line_end,
-        "C-h" | "backspace" => delete_char_backward,
+        "C-h" | "backspace" | "S-backspace" => delete_char_backward,
```
pascalkuthe, line 366: are there other keybindings that might have worked without this feature but don't work anymore now?
If that is the case should we offer a config option to dsiable this feature? Also I think this technically qualifies as a breaking change because people could have custom keybdings like this which would stop working. I added the apropriate label so we don't forget about that

pascalkuthe:
I think this closes https://github.com/helix-editor/helix/issues/5765 btw

Cargo.toml
the-mikedavis: Now this patch is used for two things:

* A regression in 0.26 fixed on master that made `terminal::supports_keyboard_enhancement()` always return `Ok(false)`
* Adding support for the `KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS` flag

It would be fine for a package manager to use 0.26 for the regression, it would just disable this feature permanently. The `KeyboardEnhancementFlags::REPORT_ALTERNATE_KEYS` constant would cause a compile error though since it's new.

If package managers don't like git dependencies then this change should wait for the next crossterm release (or we could publish a new package to crates but that has problems on its own). I wouldn't mind waiting for the next crossterm release

pascalkuthe: I think in that case its fine to merge this as is with a git dependency. Its not a problem for distros to patch out two lines if they take issue with the git dependency :+1:

helix-term/src/keymap/default.rs
the-mikedavis: Ambiguous keycodes like using `C-i` for `tab` or using `C-backspace` for `C-h` would change but I think those would be pretty rare. I'm not sure how I feel about a config option for this: if the protocol is enabled then you get more predictable and correct key events - you would only disable it if you wanted to use ambiguous keycodes which I don't think has any advantage

pascalkuthe: sounds reasonable :+1: I think we should keep the breaking change label tough so this PR is a bit easier to find for people who do experience regressions.

PR #4940, opened 2022-11-30
reset command register to update it

QiBaobin:
Fix the bug that `:` register always use the first command


ath3:
This change deletes the command history.

QiBaobin:
> This change deletes the command history.

Good catch, thanks. Let me close this and try to find out a way to use the last item in the history register.

PR #4941, opened 2022-11-30
introduce `>` register as status output

QiBaobin:
we can easily use the output of a lot of commands like `theme` `get-option` etc.


pascalkuthe:
The register should be documentes in `usage.md` in the book

QiBaobin:
> The register should be documentes in `usage.md` in the book

Thanks, I will convert this to draft now, and I will do that once reviewers say it's good to go forward in this way.

PR #4954, opened 2022-12-01
Fix nightly clippy lints

tshepang:


archseer:
I like the idea, but what about nightly lints that suggest changes that won't work on MSRV? For example lints were suggesting `retain_mut` on nightly before we were on a version recent enough to include that

the-mikedavis (approved):

Issue #4874, opened 2022-11-24
Ctrl-i doesn't do anything with clangd


creikey:
### Summary

The keymap says it should jump forward on the jumplist, but it doesn't

### Reproduction Steps

Type this into helix:
```c
int test() { return 5; }

int main() {
  test();
}
```

1. Put cursor over the call to test();
2. Type `gd`, cursor jumps to `test()`
3. Press `Ctrl-o`, cursor jumps back to the call to test correctly.
4. `Ctrl-i` now should jump forward again in the jump list but it doesn't

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-11-23T23:09:20.549 helix_lsp::transport [ERROR] err <- "I[23:09:20.549] <-- textDocument/definition(28)\r\n"
2022-11-23T23:09:20.550 helix_lsp::transport [ERROR] err <- "I[23:09:20.549] --> reply:textDocument/definition(28) 0 ms\r\n"```

</details>


### Platform

Windows 11

### Terminal Emulator

Windows Terminal

### Helix Version

7e99087fa36fd49b350311ebdbc3b9338794746b

archseer:
https://github.com/helix-editor/helix/issues/4864#issuecomment-1325598084

creikey:
> [#4864 (comment)](https://github.com/helix-editor/helix/issues/4864#issuecomment-1325598084)

This would be very strange as the same example works fine with neovim

CptPotato:
It indeed is very strange but I can confirm that <kbd>Ctrl</kbd><kbd>i</kbd> works in neovim regardless of my terminal emulator configuration. In helix I have to change the config according to the comment linked above.

It's possible that <kbd>Ctrl</kbd><kbd>i</kbd> sends a different input to the terminal application by default (I think it's `Tab`) and neovim has a mapping for that which performs the same action as <kbd>Ctrl</kbd><kbd>i</kbd>. At least on my machine <kbd>Tab</kbd> as well as <kbd>Ctrl</kbd><kbd>i</kbd> both jump forward in neovim.

creikey:
> It indeed is very strange but I can confirm that Ctrli works in neovim regardless of my terminal emulator configuration. In helix I have to change the config according to the comment linked above.
> 
[...]

Issue #4864, opened 2022-11-23
Windows dead keys problem


lesleyrs:
Some problems I came across when going through the tutorial:

Ctrl+i does not seem to work at all on windows 10 in cmd/powershell/windows terminal (Ctrl+o works), but it does work on linux.

Alt-( and Alt-) are confusing to use, maybe an example on how you would use this?

~ and Alt-` or without Alt work first try on linux, but on windows you have to press them twice for it to work most of the time.

the-mikedavis:
The tutor file is here if you would like to contribute: https://github.com/helix-editor/helix/blob/master/runtime/tutor

Whichever terminal emulator you are using on Windows is most likely stealing the `C-i`, `~` and <code>A-`</code> keypresses. You may need to configure your terminal emulator or try another one (like [WezTerm](https://github.com/wez/wezterm)).

I use `A-(`/`A-)` pretty frequently. There's a common scenario where you want to swap the order of arguments in a function call:

```rs
foo(baz, bar)
```

If your cursor is on the `a` in `baz`, you can do:

* `mi(` to select within the parentheses
* `S, <ret>` to split that selection on the `", "` part. You end up with a selection on `baz` and one on `bar`
* `A-)` to rotate the selections so you end up with `foo(bar, baz)`

lesleyrs:
Thx for the response, but regarding the terminal I believe it would be a great thing if helix just works out of the box on default windows cmd/powershell.

I believe ~ and the backticks not working is related to the fact that on linux you can just press those keys and it appears on screen. While on windows to make them appear on screen you have to press space (doesn't work on helix) or press the key twice. So this makes it actually a helix bug as it's probably checking for it to receive the ` ~ keys and not the key press itself.

Not sure why Ctrl+i doesn't work though, can't find any keybinds connected to it. Would be helpful if anyone knows.

CptPotato:
> Not sure why Ctrl+i doesn't work though, can't find any keybinds connected to it. Would be helpful if anyone knows.

Huh that's interesting, I've had the same issue on Windows with WezTerm. I think the terminal ended up eating the command. The way I fixed it is by adding this to my WezTerm keybinds:

```lua
{ key="i", mods="CTRL", action=wezterm.action.SendKey { key="i", mods="CTRL" } },
```

[...]

PR #4961, opened 2022-12-01
Add Ctrl-i alias for Windows

lesleyrs:
Closes https://github.com/helix-editor/helix/issues/4874, closes https://github.com/helix-editor/helix/issues/4864

Tested on both Linux and Windows, the aliases don't seem to conflict each other.

"tab" on Linux works as both tab and C-i and "C-i" itself doesn't do anything on gnome terminal.

While on Windows tab works as tab and "C-i" works as C-i.

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -198,7 +198,7 @@ pub fn default() -> HashMap<Mode, Keymap> {
 
         // z family for save/restore/combine from/to sels from register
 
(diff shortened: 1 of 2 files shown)


archseer:
Can you remove the unrelated tutor fixes? I think we already rejected those changes in a different PR

lesleyrs:
I don't think that's the case because I added them because of this yesterday https://github.com/helix-editor/helix/issues/4864#issuecomment-1333635640.

Removed it anyway.

the-mikedavis (approved):

PR #4966, opened 2022-12-02
nightfox theme: Use brighter colors for diff scopes

alexsunxl:
Hi, I just fix the  nightfox theme Diff sign color too dumb， so that can no see it in the code edit view.
Please take a moment to see this，Thanks！

### DIFF
--- a/runtime/themes/nightfox.toml
+++ b/runtime/themes/nightfox.toml
@@ -140,10 +140,10 @@
 # Diff ==============================
 # Version control changes.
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
\cc @jhscheer what do you think about these colors?

jhscheer:
> \cc @jhscheer what do you think about these colors?

The colors were indeed too dim. This is a good fix.

the-mikedavis (approved):

PR #4970, opened 2022-12-02
feat(docs): style readme & add badges

matoous:
Centre Helix title and logo and add badge with latest release and link to matrix chat.

Screenshot 2022-12-02 at 12 31 16


archseer:
Thanks :) I experimented with centering but it looks a lot better with badges centered too

PR #4971, opened 2022-12-02
Use the logo as the favicon for the docs book

the-mikedavis:


David-Else:
@the-mikedavis https://github.com/helix-editor/helix/blob/master/contrib/helix.png also needs updating for the `Helix.desktop`.

Issue #4972, opened 2022-12-02
diff.plus, diff.minus and diff.delta theme keys need updating for many themes for git gutters


David-Else:
Since git gutters landed https://github.com/helix-editor/helix/pull/3890 many themes need the diff colors updating, I have taken screen shots of the ones I consider in obvious need of attention, but appreciate some changes may be subjective.

*EDIT There is also `"diff.delta.moved"` to consider, I missed that, I am not sure if it is needed?
*EDIT Theme authors, a new release of Helix is planned this weekend, so please make any changes you want ASAP :)

- [ ] emacs
- [x] flatwhite
- [x] kanagawa
- [x] serika_light
- [x] acme
- [x] nord_light
- [x] hex_steel
- [x] hex_toxic
- [x] zenburn
- [x] doom_acario_dark

## Correct Example
Here is an example of `dark_plus` (VS Code theme) which has been [updated specifically](https://github.com/helix-editor/helix/pull/4661/files) for https://github.com/helix-editor/helix/pull/3890:

![dark_plus_correct](https://user-images.githubusercontent.com/12832280/205325917-725d3317-ba97-4e55-b50e-a0eeb4e54a58.png)

## Need Attention

There follows screen shots of those that seem in need of updates, there maybe more, this is just my opinion. I have tried to add a ping to the authors:

@Yevgnen 
![Screenshot from 2022-12-02 14-57-48](https://user-images.githubusercontent.com/12832280/205325955-c83d4204-43cd-4d11-bbef-9429c37afa06.png)

theli-ua:
Is there a way to style gutter symbols separately from patch/diff highlighting?

David-Else:
@theli-ua No, afraid not, these theme keys were only meant for diff gutters.

Rinnray:
There we go, Doom Acario have gotten a change.

pascalkuthe:
Thanks @David-Else for going trough the effort of putting this list together!

AlexanderBrevig:
I'll have a look soon. Should maybe be added to the linter if it's not already?

David-Else:
The linter could check the keys exists, and that they are different, but many of the problems have been caused by a bad choice of colors as there were no diff gutters for visual inspection.

VuiMuich:
Realised it took actually less then anticipated, as the diffs were there and just the color choice was bad...

AlexanderBrevig:
[...]

PR #4977, opened 2022-12-02
Update zenburn theme for git gutters

theli-ua:
image


theli-ua:
fixes #4972  for zenburn

PR #4982, opened 2022-12-03
Use logo for contrib icon

the-mikedavis:
https://github.com/helix-editor/helix/pull/4971#issuecomment-1335439461

This is the icon used in the .desktop entry


archseer:
Should we size it as 512x512? It's at an odd size right now

David-Else:
I can't seem to find a definitive guide to choosing desktop icon sizes, but I think 256 x 256 is the standard for Gnome at least: https://developer.gnome.org/documentation/tutorials/themed-icons.html

> Other sizes are optional, but a 256×256 icon is the default size used by GNOME in its application grid, so you’re **strongly** encouraged to provide one.

Reaper only provides a 256 x 256 icon, and I am sure they are trying to follow standards as closely as possible:

Screenshot from 2022-12-03 08-29-38

As the Helix desktop icon here has no installer at the moment (that I know about, but I have only installed from source) it is up to the user to deal with the icon, to be future proof you might want to provide it in the following standard config, here is the OBS studio flatpak:
```
icons
└── hicolor
    ├── scalable
    │   └── apps
    │       └── com.obsproject.Studio.svg
    ├── 512x512
    │   └── apps
    │       └── com.obsproject.Studio.png
    ├── 256x256
    │   └── apps
    │       └── com.obsproject.Studio.png
    └── 128x128
        └── apps
            └── com.obsproject.Studio.png
```
Some apps provide 16x16, 64x64 etc, but I don't think there is any need for that. There is probably no need to do anything other than resize it to 256 x 256 at this point?

the-mikedavis:
Downscaled it to 256x256 👍

PR #4983, opened 2022-12-03
Make `from_reader` and `to_writer` generic

kirawi:
`from_reader` now takes a `builder` parameter that implements `Write`, and `to_writer` will accept any `Iterator<Item = &str>`. One consideration is that using `from_reader` with `RopeWrite` means needing to manually `finish()` it. The majority of the diff is from moving these two functions into a separate file. The changes are otherwise a ~15LoC change to the implementation and changing the function calls to reflect the changes.


helix-view/src/stream.rs
```diff
@@ -0,0 +1,208 @@
...
+pub struct RopeWrite(RopeBuilder);
+impl RopeWrite {
+    pub fn finish(self) -> Rope {
+        self.0.finish()
+    }
+}
+
+impl Write for RopeWrite {
+    fn write(&mut self, buf: &[u8]) -> std::io::Result<usize> {
+        self.0.append(std::str::from_utf8(buf).unwrap());
```
kirawi, line 19: `from_reader` guarantees that the input is always valid UTF8, but I wasn't sure if it would be appropriate to use `from_utf8_unchecked` here.

helix-view/src/stream.rs
```diff
@@ -0,0 +1,208 @@
...
+            total_read += read;
+            total_written += written;
+            match result {
+                encoding::CoderResult::InputEmpty => {
+                    debug_assert_eq!(slice.len(), total_read);
+                    break;
+                }
+                encoding::CoderResult::OutputFull => {
+                    debug_assert!(slice.len() > total_read);
+                    let _ = builder.write(buf_str[..total_written].as_bytes())?;
```
kirawi, line 115: TODO: I think this needs to be properly handled.

helix-view/src/stream.rs
pascalkuthe: Using `from_utf8_unchecked` would require you to mark this function unsafe as it is public and can potentially be called anywhere.

However we really want to avoid the ovheread of utf8 validation. Considering that you convert a `str` to bytes at the callsite I think it would be better to either:

* Use a custom `Write` trait that accepts `str` instead of `[u8]`
* use `std::fmt::Write`

pascalkuthe:
Just out of curiosity: Why are you implementing this? Do we need anything else but a rope anywhere?

kirawi:
https://github.com/helix-editor/helix/pull/3966 needed to reimplement `from_reader` for a `String`.

archseer:
`from_reader` is specialized for ropes, https://github.com/helix-editor/helix/pull/3966 should just `std::fs::read` and then decode the contents

archseer:
Or `RopeBuilder` could support something like `Writer` then `finish` would be called outside that `from_reader` method

PR #4984, opened 2022-12-03
[experimental] feat(commands): git-log picker

matoous:
[experimental, wip]

Add git-log command that opens picker with git commits and their preview.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -557,7 +557,7 @@ checksum = "18d4ce09c0a6c71c044700e5932877667f427f007b77e6c39ab49aebc4719e25"
 dependencies = [
  "bstr 1.0.1",
  "btoi",
- "git-date",
+ "git-date 0.2.0",
  "itoa",
  "nom",
  "quick-error",
@@ -669,12 +669,24 @@ dependencies = [
 ]
 
 [[package]]
+name = "git-date"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "e33db9f4462b565a33507aee113f3383bf16b988d2c573f07691e34302b7aa0a"
+dependencies = [
+ "bstr 1.0.1",
+ "itoa",
+ "thiserror",
+ "time",
+]
+
+[[package]]
 name = "git-diff"
 version = "0.21.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0a88666a0ae4365b55a0cbf2efde68d2a4cff0747894ad229403bd60b0b2abc5"
 dependencies = [
- "git-hash",
+ "git-hash 0.9.11",
  "git-object",
  "imara-diff",
  "thiserror",
@@ -687,7 +699,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "881e4136d5599cfdb79d8ef60d650823d1a563589fa493d8e4961e64d78a79f2"
 dependencies = [
  "bstr 1.0.1",
- "git-hash",
+ "git-hash 0.9.11",
  "git-path",
  "git-ref",
  "git-sec",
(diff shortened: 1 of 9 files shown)


Proful:
Can I help in testing it as you progress?

matoous:
@Proful yes please, feel free to. I will mostly want/need input from @pascalkuthe (if you find some time for me please, mostly architectural questions) and also importantly, I would like to know how far we want to go with vcs/git integration and what should be left to plugins eventually (cc @archseer).

I think this will need split into 3 PRs: one for picker refactor/new kind of picker, one for VCS registry refactor, and one for this log licker itself. I would also like to keep this PR as small as possible so I think the scope of this PR would could be kept at commit preview - without picker action, etc. which could be implemented in follow-up PRs.

Proful:
Looking great!!

image

Minor observation:-
- Commit picker started as soon as `:git-log` typed. It should appear after 'enter' pressed
- Bcoz of above tab key is not going to `change-current-directory`
- Loading of all the git log message is very fast

The below suggestion is too early. But it would be great to see something like this in future

image

David-Else:
>  I would like to know how far we want to go with vcs/git integration and what should be left to plugins eventually (cc @archseer).

I think LazyGit is the ultimate UI for git in the terminal and there is no point in trying to compete with it. You can use a terminal with window splitting and open it next to Helix, job done.

Screenshot from 2022-12-04 16-26-05

There are some git integrations that would really work in Helix and can't be done in another window, for example the git gutters. I don't think there is any need to add a lot of git integration other than what can't be achieved with other tools.

There is already a discussion on this subject: https://github.com/helix-editor/helix/discussions/4193

I think these are the two most popular TUI git tools:

https://github.com/jesseduffield/lazygit
https://github.com/extrawurst/gitui

pascalkuthe:
> I think LazyGit is the ultimate UI for git in the terminal and there is no point in trying to compete with it. You can use a terminal with window splitting and open it next to Helix, job done.

As I mentioned above integrating a history just for the sake of showing the history is not really useful. However for `dif`/`diff-gutter` I want to change the commit we are comparing against with a command for which I do think a picker would be very useful. Doing this outside of helix and then copying the commit hash into a command is theoretically possible (using fzf in the terminal for fuzzy search or some gitui for looking at the history) but that argument can be made for many pickers (for the file picker you could use fzf to select a file on the terminal instead and copy the path into the `:open` command). So I do think a commit picker has  merit inside helix (and I think  a history based one could be a useful alternative to a direct search).

univerz:
> I think LazyGit is the ultimate UI for git

diff without proper syntax highlighting on that screenshot looks suboptimal to me.

David-Else:
> > I think LazyGit is the ultimate UI for git
> 
> diff without proper syntax highlighting on that screenshot looks suboptimal to me.

LazyGit optionally uses [Delta](https://github.com/dandavison/delta) which lets you choose your own theme for syntax highlighting based on [bat](https://github.com/sharkdp/bat). I am using the VS Code one, it is not very close, but others should be better.

univerz:
> based on [bat](https://github.com/sharkdp/bat)

i could probably configure it somewhere close to my custom helix theme, but the point is - we are diffing (and staging!) code, we should see it as is most natural == same as we see it in our editor of choice, ideally in the editor - this is what i call ultimate UI.

jvoisin:
I'm unsure about re-implementing [magit]( https://magit.vc/ ) into helix, it's better suited as a plugin, or we'll end up with mercurial/svn/darcs/… there in it as well.

matoous:
I am turning this into and issue: https://github.com/helix-editor/helix/issues/5292, let's continue this conversation there. This PR is as far from ready as it gets (as mentioned by @pascalkuthe) and re-opining a new PR at later point once we are closer to being able to support the commit picker seems ineviatble.

PR #4987, opened 2022-12-03
Enable auto format for css and scss files

zummenix:
Took me a while to figure out why auto format doesn't work. I think these are sensible defaults.

`provideFormatter` enables capability in LS and `auto-format` performs format on save.


David-Else:
I tested and it works as expected.

the-mikedavis (approved):

PR #4991, opened 2022-12-03
Set auto-format true for toml

David-Else:
`taplo` has formatting built in, but `auto-format` has been off until now. I don't know if this was because it was broken in the past, an oversight, or a concious decision?

I notice that the Helix `.toml` files have not been formatted. If we turn on auto-format it sends the signal to contributors that the Helix `.toml` files should be formatted... with the default settings they look quite different. I have matched them approximately in a `taplo.toml` file. It brings much needed consistency to `languages.toml`. Shall I make a pull request with the settings file in Helix's root and some formatting applied?

EDIT: You can see the results of the formatting with an options file in the reverted `[Add taplo.toml formatting options and format languages.toml](https://github.com/helix-editor/helix/pull/4991/commits/a1804d554418e9751064cd6f1aac81ed8ade9d66)` commit. Please excuse my wrong way of using GitHub, I am still working it out!

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -60,6 +60,7 @@ file-types = ["toml"]
 roots = []
 comment-token = "#"
 language-server = { command = "taplo", args = ["lsp", "stdio"] }
+auto-format = true
 indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
--- a/languages.toml
+++ b/languages.toml
@@ -27,26 +27,26 @@ command = "lldb-vscode"
 [[language.debugger.templates]]
 name = "binary"
 request = "launch"
-completion = [ { name = "binary", completion = "filename" } ]
+completion = [{ name = "binary", completion = "filename" }]
 args = { program = "{0}" }
 
 [[language.debugger.templates]]
 name = "binary (terminal)"
 request = "launch"
-completion = [ { name = "binary", completion = "filename" } ]
+completion = [{ name = "binary", completion = "filename" }]
 args = { program = "{0}", runInTerminal = true }
 
 [[language.debugger.templates]]
 name = "attach"
 request = "attach"
-completion = [ "pid" ]
+completion = ["pid"]
 args = { pid = "{0}" }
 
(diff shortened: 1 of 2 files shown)


zummenix:
I've tried this. Personally, I don't mind having default formatting options. Let's see what others think.

David-Else:
Default options would be fine with me too.

the-mikedavis:
I'm hesitant to enable auto-format for any language that doesn't have a canonical (or at least very popular, like `prettier`) formatting style and it looks like taplo's style is custom. If you're working on another project and touch its Cargo.toml and introduce unrelated formatting changes, you end up fighting with the formatter and resetting any unrelated formatting changes before submitting a PR. And if we apply its formatting changes to our toml files, we would also want to setup a lint to reject changes that aren't formatted. But then contributors would need to install the taplo language server or CLI in order to contribute. I think that's fine for rust formatting and lints but feels unnecessary for toml files

David-Else:
@the-mikedavis 

> ...and it looks like taplo's style is custom

There are only 2 formatters for TOML, taplo being one of them, I have not investigated if there are any differences in the defaults: https://github.com/toml-lang/toml/wiki#formatters

> if we apply its formatting changes to our toml files, we would also want to setup a lint to reject changes that aren't formatted. 

Why not just set the GitHub actions to format it for us?

zummenix:
As I understood @the-mikedavis voiced two main concerns:

1. `auto-format = true` for toml
2. format `languages.toml` in this PR

For the first point, my opinion is that if LS for a particular language has a formatting capability, it should be enabled by default. This corresponds to **Batteries included** item in [vision.md](https://github.com/helix-editor/helix/blob/master/docs/vision.md)
Besides, as @David-Else has mentioned there aren't any other alternatives to toml formatters.

For the second point, I feel like to force contributors to have taplo LS installed is overkill indeed. Then just formatting `languages.toml` manually once in a while should be fine. There aren't any major differences anyway and most of contributions will not require any additional formatting at all.

PR #4992, opened 2022-12-03
feat: adapt status line separator style to focused or not

poliorcetics:
Before this change:

Screenshot 2022-12-03 at 16 29 55

After this change:

Screenshot 2022-12-03 at 16 30 20

It can be a little hard to spot: look on the right of the screenshots, at the separator character `|`


archseer:
I think this could be fixed in the theme, by not setting the background color of the separator

poliorcetics:
Ah yes, you're right, sorry for the noise

PR #4995, opened 2022-12-03
fix git diff when core.autocrlf is enabled

### LINKED ISSUE #4989: new diff signs don't work if line endings are crlf
### Summary

the new diff signs feature #3890 colors the entire file using the diff.delta markers on windows where files by default will use crlf line endings.
changing the line endings to lf "fixes" the issue.

### Reproduction Steps

 

I tried this:

1. `hx`
2. open a file with code which is tracked by a git repo

I expected this to happen:
image

Instead, this happened:
image

### Helix log

_No response_

### Platform

windows

### Terminal Emulator

windows terminal

### Helix Version

helix 22.08.1 (ee010c49)

pascalkuthe:
Fixes #4989 by converting all LF line endings to CRLF line endings if `core.autocrlf` is enabled when retrieving the diff_base.
This approach does not cover all edge cases as it ignores git-attributes filter which are not quite implemented in gitoxide yet, see [comment](https://github.com/helix-editor/helix/issues/4989#issuecomment-1336174385) by @Byron.

Furthermore, we can probably never handle some edge cases. If `core.autocrlf` is set to `true` or `input` all `CRLF` are automatically converted to `LF` on checkin. A normal `git diff` actually applies this preprocessing to all worktree files before computing the diff. Therefore the correct approach would be to convert all `CRLF` to `LF` during the diff however this would require cloning the entire document and is therefore very slow (would require access to git specific information inside the diff worker) and can not be perform well. 

This is rarely an issue in practice if these setting are enabled a git checkout will either be fully `LF` or fully `CRLF`. Helix does not change line-endings automatically so the only way to run into this edgecase is to use `:line-endings` (to change to the "wrong" line ending for the platform) or if some external program changes line endings.


dead10ck (approved):
This seems reasonable to me

helix-vcs/src/git.rs
```diff
@@ -17,14 +18,17 @@ impl Git {
         // custom open options
         let mut git_open_opts_map = git::sec::trust::Mapping::<git::open::Options>::default();
 
-        // don't use the global git configs (not needed)
+        // On windows various configuration options are bundled as part of the installations
+        // This path depends on the install location of git and therefore requires some overhead to lookup
+        // This is basically only used on windows and has some overhead hence it's disabled on other platforms.
+        // `gitoxide` doesn't use this as default
```
Byron, line 24: `gix` is a good example for it as it selectively enables this option depending on which command is run. Some are known to need this extra information, like when it's related to authentication on all platforms, whereas others don't.

helix-vcs/src/git.rs
```diff
@@ -61,7 +65,30 @@ impl DiffProvider for Git {
         let file_oid = find_file_in_commit(&repo, &head, file)?;
 
         let file_object = repo.find_object(file_oid).ok()?;
-        Some(file_object.detach().data)
+        let mut data = file_object.detach().data;
+        // convert LF to CRLF if configured to avoid showing every line as changed
+        if repo
+            .config_snapshot()
```
Byron, line 71: Note that this will not update throughout the lifetime of the `Repository`. Soon there should will be a `config()` counterpart which will assure all involved configuration is also uptodate.

In this case, I don't think it's an issue as this setting is very stable and usually configured by the platform installation itself to apply for all repositories on the system.

helix-vcs/src/git.rs
```diff
@@ -61,7 +65,30 @@ impl DiffProvider for Git {
...
+            .boolean("core.autocrlf")
+            .unwrap_or(false)
+        {
+            let mut normalized_file = Vec::with_capacity(data.len());
+            let mut at_cr = false;
+            for &byte in &data {
+                if byte == b'\n' {
+                    // if this is a LF instead of a CRLF (last byte was not a CR)
+                    // insert a new CR to generate a CRLF
+                    if !take(&mut at_cr) {
```
Byron, line 81: This is the first time I have seen this being used to turn off a boolean after checking its value, and it's neat despite maybe being a little implicit.

Byron (approved):
LGTM!

I have started [a support ticket](https://github.com/Byron/gitoxide/issues/637) over at the `gitoxide` repository to keep track of requirements as they pop up.

helix-vcs/src/git.rs
pascalkuthe: Repos are not cached right now anyways (we just opened that repo a few lines before) and there are not automatic reloads. Therefore if you want to update your gutter to reflect any on-disk changes you need to call `:reload` or `:reload-all` anyways which will always fully reload the repo from disk. This will becomes more of an issue once repo caching is enabled. But then your point apllies that this setting is rarely changed. In fact it's often set for windows git installations by default

CptPotato:
Gave it a quick spin and it works for me.

Also thanks for keeping up with all the "Windows problems".

helix-vcs/src/git.rs
the-mikedavis: I agree: I think it would be clearer to set `at_cr` separately from reading it. The `!` plus the `take` makes this pretty dense.  Plus I think it's non-obvious that `false` is the `Default` of `bool` which can make this look surprising

pascalkuthe: I removed the use of `take`

PR #4997, opened 2022-12-03
Fleetish theme renamed to fleet dark and adjusted to match official theme. 

krfl:
Fleetish was originally based on an early screenshot of JetBrains Fleet. Fleet's theme changed considerably between the initial reveal and the public beta/preview and while Fleetish does implement many of these changes, it also keeps elements of the old theme used in the screenshot.

While I never intended Fleetish to be a perfect copy of Fleet's official theme, I now see that the name Fleet-ish can be a little problematic because it sets the expectation of a theme similar to that of Fleet. Thus I have changed the name of the theme to "fleet_dark" and done some work to further match it with the official dark Fleet theme. There is still nitpicking to be done, but this is a big step towards the "expected" look. I don't think it's possible to match it perfectly due to the way JetBrains seem to change their highlighting scopes from filetype to filetype, but it just needs to be _good enough_. It's currently "optimized" for rust and markdown, so there will be bigger differences for other languages. 

Lastly, this will likely break a few configurations, but I do believe these changes are for the better to avoid further confusion and disappointments. 

**Opinions and suggestions are very much welcome.** 

Below you can see some screenshots comparing Helix and Fleet. I also included some python stuff where you can clearly see differences. 

fleet_dark1

fleet_dark2

fleet_dark3

fleet_dark4

Tagging @matoous, @archseer and @pascalkuthe for comments and thoughts.


archseer:
The renaming is fine since it's a one line fix on the user's end, we'll denote this in the breaking changes section of the release notes 👍🏻

archseer:
Second image: to style `Fixed`, `Auto` inside the enum you can use `@type.enum.variant`

krfl:
> Second image: to style `Fixed`, `Auto` inside the enum you can use `@type.enum.variant`

Thank you! I pushed a fix for this.

runtime/themes/fleet_dark.toml
```diff
@@ -76,25 +82,25 @@
 "ui.text" = "light" # .focus / .info
```
matoous, line 82: **suggestion**: this could provide better highlighting for the pickers matching the picker of fleet.
```suggestion
"ui.text" = "light" # .info
"ui.text.focus" = { fg = "lightest", bg = "blue_accent" }
```

krfl: Could you provide a screenshot of what this looks like? I'm currently away from my laptop

matoous: Sure think, sorry. 

Here's how the picker looks currently:

Screenshot 2022-12-04 at 22 33 18

Here's how it would look like with the change above (which is actually incorrect):

Screenshot 2022-12-04 at 22 33 56

The correct highlight color should be `#204474` which would look like this:

Screenshot 2022-12-04 at 22 38 06

And that matches fleet:

Screenshot 2022-12-04 at 22 38 16

I guess the same should be used for `ui.menu.selected`, currently looking like this:

Screenshot 2022-12-04 at 22 44 32

versus (in Fleet):

Screenshot 2022-12-04 at 22 44 32

I.e., maybe this would work:

```toml
...
"ui.menu.selected" = { fg = "lightest", bg = "dark_blue_accent" } # .selected
"ui.text.focus" = { fg = "lightest", bg = "dark_blue_accent" }

...

```

krfl: Looks great! Good job. Can we reuse the "selected" color or is that too dark?

matoous: If we are to follow Fleet strictly it uses 2 different color, the one configured right now for `selected` for selections and the one above for the menues.

krfl: Alright. Maybe name it "focus" instead of accent, as accent colors are much brighter than anything else?

matoous: Yes, sounds good 👍

archseer:
```
"keyword.control.import" = "cyan"
```

to get `from`, `import` as cyan keywords in Python/Rust

archseer:
Actually most keywords seem cyan (`if`/`else` etc)? Maybe `keyword.control` as `cyan` is enough, and you need a special setting only for `keyword.control.exception`

krfl:
> Actually most keywords seem cyan (`if`/`else` etc)? Maybe `keyword.control` as `cyan` is enough, and you need a special setting only for `keyword.control.exception`

Fixed. Also adjusted the color of builtins. It's pretty close for Rust now.

archseer (approved):
Looks good to me! Thanks for the changes :)

PR #4998, opened 2022-12-04
when using :q with unwritten buffers, prompt for whether to save them

doy:
similar to vim's `confirm` option.

i'm not sure exactly what the ui here should look like - i implemented a modal dialog window thing, but maybe something more like a message + prompt (similar to vim) would be better? open to suggestions.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -123,6 +123,14 @@ impl<'a> Context<'a> {
     pub fn count(&self) -> usize {
         self.count.map_or(1, |v| v.get())
     }
+
+    pub fn compositor_context(&mut self) -> compositor::Context {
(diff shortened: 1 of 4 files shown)


doy:
i don't really agree? it requires a lot more typing and a lot more thinking for something that should be an extremely basic and common operation. i typically have `:q` bound to a single key because i bounce around between editing and the shell a lot, and it's very hard to replicate that ease of use when you need to think about and remember which of like three or four commands to use to exit. this also doesn't really make the existing workflow any slower.

pascalkuthe:
I am closing this as stale as this is pretty old and has conflicts. We are esstially following what vim does here so I don't see us going in this direction in the future. 

Thank you for contributing!

PR #4999, opened 2022-12-04
Update Acme and Nord Light for git gutters

two-six:
nord light diffs
acme diff
Fixes for #4972

### DIFF
--- a/runtime/themes/acme.toml
+++ b/runtime/themes/acme.toml
@@ -22,6 +22,9 @@
 "diagnostic.hint" = {bg="white", modifiers=["bold"]}
 "ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
 "ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
+"diff.plus" = {fg = "green"}
+"diff.delta" = {fg = "red"}
+"diff.minus" = {fg = "acme_bar_bg"}
 
 [palette]
 white = "#ffffff"
--- a/runtime/themes/acme.toml
+++ b/runtime/themes/acme.toml
@@ -23,8 +23,8 @@
 "ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
 "ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
 "diff.plus" = {fg = "green"}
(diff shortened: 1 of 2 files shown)


runtime/themes/nord_light.toml
```diff
@@ -58,6 +58,10 @@
 "markup.link.text" = {fg="nord12"}
 "markup.quote" = {fg="nord3", modifiers=["italic"]}
 
+"diff.plus" = {fg = "nord14"}
+"diff.delta" = {fg = "nord11"}
+"diff.minus" = {fg = "nord15"}
```
archseer, line 63:
```suggestion
"diff.plus" = {fg = "nord14"}
"diff.delta" = {fg = "nord13"}
"diff.minus" = {fg = "nord11"}
```

This is per descriptions in https://www.nordtheme.com/docs/colors-and-palettes

pascalkuthe:
It's a bit odd that modified is red and removed is blue (for the second theme). That's pretty much the opposite of most other themes/the general expectation. Would it be an option to swap those two? (same for the other theme although I think it's still ok this way)

runtime/themes/acme.toml
```diff
@@ -22,6 +22,9 @@
 "diagnostic.hint" = {bg="white", modifiers=["bold"]}
 "ui.bufferline" = { fg = "indent", bg = "acme_bar_bg" }
 "ui.bufferline.active" = { fg = "black", bg = "acme_bg" }
+"diff.plus" = {fg = "green"}
+"diff.delta" = {fg = "red"}
+"diff.minus" = {fg = "acme_bar_bg"}
```
archseer, line 27: Same here, you probably want minus as red (deletion) and delta as something else (modification, usually purple? `acme_bar_bg` could work)

archseer (changes requested):

two-six:
> It's a bit odd that modified is red and removed is blue (for the second theme). That's pretty much the opposite of most other themes/the general expectation. Would it be an option to swap those two? (same for the other theme although I think it's still ok this way)

Sorry, I was taking care of it late at night so I didn't notice. You're completely right 

@archseer Thanks for this change, it looks better :)

PR #5005, opened 2022-12-04
Fix complilation on cargo 1.56

jvoisin:
This commit fix the following error:

```
$ cargo --version
cargo 1.56.0
$ cargo install --path helix-term
error: failed to parse manifest at `/home/jvoisin/Applications/helix/Cargo.toml`

Caused by:
  feature `named-profiles` is required

  The package requires the Cargo feature called `named-profiles`, but that feature is not stabilized in this version of Cargo (1.56.0).
  Consider adding `cargo-features = ["named-profiles"]` to the top of Cargo.toml (above the [package] table) to tell Cargo you are opting in to use this unstable feature.
  See https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#custom-named-profiles for more information about the status of this feature.
[101]
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,3 +1,5 @@
+cargo-features = ["named-profiles"]
+
 [workspace]
 members = [
   "helix-core",


pascalkuthe:
Helix requires rustc `1.61`., What is the use-case of compiling with a lower cargo verison when rustc needs to be updated anyway? These two are basically always packed together as they bootstrap each other

jvoisin:
It's the case on Debian:

```
$ rustc --version
rustc 1.62.1
$ cargo --version
cargo 1.56.0
$
```

pascalkuthe:
This feature was stablized on cargo 1.57.
It seem you are using debian testing (probably the cause for the weird version mismatch) which is already updated to cargo 1.57: https://packages.debian.org/search?keywords=cargo so you just need to update

Issue #411, opened 2021-07-04
Virtual text (text annotations) support


sudormrfbin:
Similar to neovim's virtual text feature (greyed out faint text at end of lines in the image):

![Screenshot_2021-07-04_10-38-38](https://user-images.githubusercontent.com/23398472/124373902-10434480-dcb4-11eb-9649-4a74a9735d16.png)

This would enable having indent guides, inlay hints (useful with rust-analyzer), inline diagnostics messages, etc. Being able to place virtual text _in between_ normal text would also be nice.

I have a general idea as to how to do it, I'll take a stab at it.

archseer:
I think this might be a bit tricky right now, but we've had discussions with @cessen on how to support these type of inline "decorations". Vim only supports these at the end of line, because placing them inline affects on-screen positioning calculation.

The way I imagine this working is that it would be a character offset `pos`, the text and a style. The render loop would then need to handle these injections and any sort of visual positioning calc then needs to account for these.

---

Indent guides: I'm not sure if those should be virtual text or just special rendering code for certain characters.

I like this behavior in vim:

```
" show whitespace with <leader>s
set listchars=tab:——,trail:·,eol:$,space:·
"set listchars+=extends:›,precedes:‹,nbsp:␣
nnoremap <silent> <leader>s :set nolist!<CR>
```

This way I'm able to toggle a "MS Word-like" view where spaces and certain other chars are visualized.

sudormrfbin:
Cursor position calcuation and rendering seems like the only problematic part of the implementaion, but I'd say it's doable with some careful thought.

I think indent guides are better implemented as an application of virtual text rather than as a special case (plugins can then tweak it if required too). `listchars` like behavior can also be done using the same mechanism. I imagine there will two ways to display virtual text:

1. Inject text after a particular offset. This can be used for inlay hints, and the cursor will move "over" them when using motions (you can't place the cursor on these character).

2. Replace existing text. This can be used for `listchars` and indent guides. The cursor can be placed over them and editing operations will simply act on the actual text.

If we have a namespace mechanism for grouping together different decorations, then toggling indent guides or `listchars` will be easy too.

cessen:
This is definitely doable, and something I think we all want Helix to support.  I would suggest holding off on implementing right now, however, as it will likely be wasted effort.  The rendering system is likely to get an overhaul in the not-too-distant future, to support not only end-of-line annotations but also inline annotations, word wrapping, etc.

[...]

Issue #405, opened 2021-07-03
File comparison


kirawi:
I have a vague idea on how to do this, but it'll take me some time to familiarize myself with the UI code.

zbindenren:
Maybe delta can help in that case: https://github.com/dandavison/delta

It is the best terminal diff viewer I ever used  and it is also written in rust.

sudormrfbin:
Delta is a pager that acts on already diffed data (with `-` and `+` prepended to diffed lines like in the default git diff output) and simply applies syntax highlighting and better colors it. We will have to use [similar](https://github.com/mitsuhiko/similar) for this, which we already use in helix to `:reload` the file from disk.

archseer:
@pascalkuthe mentioned he'll probably pick up the work on this in the future after git gutter is merged.

pascalkuthe:
After #3890 is merged, the main blocker will be rendering the diffs side by side (as a form of virtual text).
I am working with @kirawi to land the necessary changes in the rendering system to make that happen.

LifeWorks:
Is this feature available in the current version? I see it is open, so I assume I still can not diff two buffers?

archseer:
Yes, open issues typically indicate that a feature hasn't been implemented yet.

wh75er:
Hello! Would it be possible to use inline git diff in one buffer? It would be amazing

stevenxxiu:
> Hello! Would it be possible to use inline git diff in one buffer? It would be amazing

Please open a separate issue for that.

Dmole:
There are 2 uses cases I'd want helixdiff to handle:

- simple diff of 2 config files
- git conflictstyle diff3 which can use:

    vimdiff -f -d -c '4wincmd w | wincmd J' "$LOCAL" "$BASE" "$REMOTE" "$MERGED"

DocteurFlashbios:
It would be really great to offer a more clear and advanced way to diff/merge files than vim. Either threeway merge is supported, but nothing is shown about who is who and I never managed to remember the "fingertricks"

Dmole:
> It would be really great to offer a more clear and advanced way to diff/merge files than vim. Either threeway merge is supported, but nothing is shown about who is who and I never managed to remember the "fingertricks"

Is 
    :ls
or
    :set statusline=%n\ %f
not what you want?

DocteurFlashbios:
[...]

PR #5008, opened 2022-12-05
Rework text rendering and integrate softwrap rendering

pascalkuthe:
This PR implements a general rework of the text rendering system.
The main aim is to make the text rendering more flexible to easily allow implementation of various decorations or 
even new text-based views besides editor without hard-coding them into text rendering.
It also contains an implementation of soft-wrapping **rendering** (not actually usable right now see below).

While working on this PR I had the following features in mind:

* Virtual text #411
  * Inline LSP annotations
  * Overlays for jump mode #3791
  * (EOL annotations)
* Soft wrapping #136
* Side by side diffs #405 
* (Structural) search and replace (a better UI for #4381, also [ssr.nvim](https://github.com/cshuaimin/ssr.nvim))
* Better display for diagnostics (similar to [lsp_lines.nvim](https://git.sr.ht/~whynothugo/lsp_lines.nvim))

# Text Render System Desgin

This PR completely replaces the `render_text_highlights` function (but leaves the rest of the editor mostly untouched).
Although this new implementation looks quite different as it's structured it's not a complete rewrite and I heavily leaned on the old implementation.

My goal during the implementation was primarily separating different components of the text rendering.
The text rendering is now roughly split into three basic components:

* Basic grapheme rendering and visual position tracking (`TextRender`)
* Transversing the document efficiently and accumulating graphemes for rendering (`DocumentCursor`)
* Gluecode that handles syntax highlighting, soft-wrapping and integration of inline decorations (`DocumentRender`)

The primary interaction with the text rendering happens through the `DocumentRender` struct.
It allows the caller to render a single **visual** line (which could just be a part of a full line in case of soft-wrapping)
with the `reder_line` function.
It behaves similar to a rust iterator (lazy) so it only renders a line if this function is called.
This function requires a mutable reference to `TextRender` which tracks the actual onscreen position.

Thanks to this decoupling `TextRender` can be used to render other text in between lines.
For example a second `DocumentRender` could be used to interpose the removed lines of a second document (with syntax highlighting and soft-wrapping)
at hunks for a unified diff view.

Furthermore, `DocumentRender` yields after each **visual** line (instead of only after line break) to ensure other rendering components
can take soft wrapping into amount.
As the `DocumentCursor`/`DocumentRender` tracks the position within the document (both `line`/`col` and total `char_offset`/`byte_offset`) anyway
this can be used to track the exact beginning/end of wrapped lines (so the diagnostic gutter could be shown on the correct soft-wrapped line).
This information can also be used by custom views to interrupt the rendering to render other text with `TextRender` at line bounds.
In this PR I simply used this property to ensure gutters are displayed in the correct place when soft-wrapping is enabled.

Inline annotations/overlays can easily be integrated with a trait that would be called from within the `push_grapheme` function.

# Performance

Compared to the old text rendering the document transversal works very different now.
Previously `RopeSlice`s were created for each syntax range `text.slice(highlight_start..highlight_end)`.
For these slices a `RopeGraphemes` iterator was then created and rendered in a single continuous (non-interruptible) loop.

This approach does not work well for this PR because we need the iteration to be interruptible.
Therefore, the graphemes are collected into a temporary vector instead which is rendered at:

* Word boundaries
* Line breaks
* at the end of the visual line when soft-wrapping is enabled
* after a fixed chunk size when soft-wrapping is disabled (to avoid excessive memory consumption)

The overhead should be minimal as the buffer never exceeds a fixed size (viewport width with soft-warping, 64 without).

As both the `slice()` function and the creation of `RopeGraphemes` are `log(N)` operations.
The code now uses a single `RopeGraphemes` iterator for the entire text (so no more slicing at highlight boundaries)
which should be a nice performance boost.
Another advantage of this approach is that text rendering can now easily be refactored to use byte positions instead of char
positions (which avoids the costly `NlogN` `char` -> `byte` conversion currently performed for treesitter queries).
However, I left this as a followup PR for now as it requires some code outside the rendering system.

# Soft Wrapping

The PR implements soft-wrapping this is partly inspired by led (credit to @cessen here).
The text is wraps words that would exceed the line-width at their starting boundary (spaces) when possible.
When the text that would need to be wrapped exceeds a certain (configurable) width (5 by default) the word start
is instead broken exactly at the line end.
In practice this usually wraps text well while only causing minimal overhead (there are no pathological cases with this approach).

Right now only (breaking) spaces (so tabs and space) are treated as word boundaries.
To better wrap source code I am considering treating everything that is not `char::is_alphanumeric` as a word boundary instead.
This would wrap long paths in rust at `.` for example.

Wrapped lines are indented with a configurable amount of space (2 by default).
They furthermore retain the indentation (and indent guides) from the start of the line unless
this indentation exceeds as (configurable) limit (people that don't like this feature can set the limit to 0).

# Landing this

The rendering system in this PR works quite well, although more testing is obviously needed.
Rendering soft-wrapping behaves well and has all the features I personally except.
However, soft-wrapping is currently not handled well in the rest of the editor which excepts that.
`View::offset` corresponds to document lines/columns which is not the case anymore with this PR. 
To address this we need some.way to find wrapped line starts which is why I would probably move.some.of the softwrapping code into `DocumentCursor` which could then be moved to `helic-core` to provide that functionality everywhere in helix. 

Gutters also need further improvements (for example insert and modify diff gutters should be shown for all wrapped lines).

In interest of landing this PR faster and enable work on orthogonal features (like side-by-side diffs)
that could leverage the improved rendering system I propose to land this PR as is (with some minor improvements)
and disable soft-wrapping with a (off by default) feature flag (everywhere `config.soft_wrap.enable` is used simply add `&& cfg!(feature=softwrap)`).

This PR still has some rough edges to polish:
* [ ] Clean up code documentation/comments
* [ ] Maybe some tests, although I am not sure how to actually test this yet (maybe asserting the contest of a surface after render?)
* [ ] Add the feature flag mentioned above
* [ ] Adjust user facing documentation (better documentation for the soft wrap config options with a note that this is an in-development feature that is disabled)
* [ ] undo some changes to view positioning that were added to demo text wrapping

A big shoutout to @kirawi for reaching out to me about potential overlap between virtual text.
The initial design discussions with him heavily influenced this PR.

### DIFF
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -14,7 +14,11 @@ use helix_view::Theme;
 use helix_view::{editor, Document};
 use tui::buffer::Buffer as Surface;
 
-pub struct DocumentRender<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> {
+pub struct DocumentRender<'a, A, H>
+where
+    H: Iterator<Item = HighlightEvent>,
+    A: AnnotationSource<'a, Style>,
+{
     pub config: &'a editor::Config,
     pub theme: &'a Theme,
 
@@ -28,7 +32,11 @@ pub struct DocumentRender<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationS
     is_finished: bool,
 }
 
-impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRender<'a, H, A> {
+impl<'a, H, A> DocumentRender<'a, A, H>
+where
+    H: Iterator<Item = HighlightEvent>,
+    A: AnnotationSource<'a, Style>,
+{
     pub fn new(
         config: &'a editor::Config,
         theme: &'a Theme,
@@ -90,22 +98,18 @@ impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRe
 
     /// Renders the next line of the document.
     /// If softwrapping is enabled this may only correspond to rendering a part of the line
-    ///
-    /// # Returns
-    ///
-    /// Whether the rendered line was only partially rendered because the viewport end was reached
-    pub fn render_line<const SOFTWRAP: bool>(&mut self, text_render: &mut TextRender) {
+    pub fn render_line(&mut self, text_render: &mut TextRender) {
         if self.is_finished {
             return;
         }
 
         loop {
-            while let Some(word) = self.cursor.advance::<SOFTWRAP>(self.highlight_scope) {
-                text_render.posititon = word.visual_position;
-                for grapheme in word.graphmes {
+            while let Some(mut word) = self.cursor.advance_with_highlight(self.highlight_scope) {
+                for grapheme in word.consume_graphemes(&mut self.cursor) {
                     text_render.draw_grapheme(grapheme);
                 }
 
+                // TODO refactor to use let..else once MSRV reaches 1.65
                 let line_break = if let Some(line_break) = word.terminating_linebreak {
                     line_break
                 } else {
@@ -115,18 +119,9 @@ impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRe
                 if self.config.indent_guides.render {
                     text_render.draw_indent_guides();
                 }
-                if !line_break.is_softwrap {
-                    // render EOL space
-                    text_render.draw_grapheme(StyledGrapheme {
-                        grapheme: Grapheme::Space,
-                        style: self.highlight_scope.1,
-                    });
-                }
 
-                if self.config.indent_guides.render {
-                    text_render.draw_indent_guides()
-                }
                 text_render.posititon.row += 1;
+                text_render.posititon.col = line_break.indent as usize;
                 self.is_finished = text_render.reached_viewport_end();
                 return;
             }
@@ -139,7 +134,7 @@ impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRe
                 // the last word is garunteed to fit on the last line
                 // and to not wrap (otherwise it would have been yielded before)
                 // render it
-                for grapheme in self.cursor.finish() {
+                for grapheme in self.cursor.finish().consume_graphemes(&mut self.cursor) {
                     text_render.draw_grapheme(grapheme);
                 }
 
@@ -148,6 +143,7 @@ impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRe
                     text_render.draw_grapheme(StyledGrapheme {
                         grapheme: Grapheme::Space,
                         style: self.highlight_scope.1,
+                        doc_chars: 0,
                     });
                 }
 
@@ -166,7 +162,7 @@ impl<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> DocumentRe
     }
 }
 
-pub type StyledGrapheme<'a> = helix_core::graphemes::StyledGrapheme<'a, Style>;
+pub type StyledGrapheme<'a> = helix_core::doc_cursor::StyledGraphemes<'a, Style>;
 
 /// A TextRender Basic grapheme rendering and visual position tracking
 #[derive(Debug)]
@@ -307,7 +303,7 @@ impl TextRenderConfig {
         doc: &Document,
         editor_config: &editor::Config,
         theme: &Theme,
-        offset: &Position,
+        cutoff: usize,
     ) -> TextRenderConfig {
         let WhitespaceConfig {
             render: ws_render,
@@ -349,7 +345,7 @@ impl TextRenderConfig {
             tab_width: tab_width as u16,
             tab,
             whitespace_style: theme.get("ui.virtual.whitespace"),
-            starting_indent: (offset.col / tab_width)
+            starting_indent: (cutoff / tab_width)
                 + editor_config.indent_guides.skip_levels as usize,
             indent_guide_style: text_style.patch(
                 theme
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -1,0 +1,362 @@
+use std::borrow::Cow;
+use std::cmp::min;
+
+use helix_core::doc_cursor::{AnnotationSource, DocumentCursor};
+use helix_core::graphemes::Grapheme;
+use helix_core::str_utils::char_to_byte_idx;
+use helix_core::syntax::Highlight;
+use helix_core::syntax::HighlightEvent;
+use helix_core::{Position, RopeSlice};
+use helix_view::editor::{WhitespaceConfig, WhitespaceRenderValue};
+use helix_view::graphics::Rect;
+use helix_view::theme::Style;
+use helix_view::Theme;
+use helix_view::{editor, Document};
+use tui::buffer::Buffer as Surface;
+
+pub struct DocumentRender<'a, H: Iterator<Item = HighlightEvent>, A: AnnotationSource<'a>> {
+    pub config: &'a editor::Config,
+    pub theme: &'a Theme,
+
+    text: RopeSlice<'a>,
+    pub cursor: DocumentCursor<'a, Style, A>,
+
+    highlights: H,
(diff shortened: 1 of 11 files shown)


helix-view/src/view.rs
```diff
@@ -215,9 +215,9 @@ impl View {
     }
 
     pub fn ensure_cursor_in_view(&mut self, doc: &Document, scrolloff: usize) {
-        if let Some((row, col)) = self.offset_coords_to_in_view(doc, scrolloff) {
+        if let Some((row, _col)) = self.offset_coords_to_in_view(doc, scrolloff) {
             self.offset.row = row;
-            self.offset.col = col;
+            self.offset.col = 0;
```
pascalkuthe, line 220: Only included so virtual text can be tried out, will be removed before merging (altough the basic idea of avoiding horizontal view movement when softwrapping is enabled is correct)

helix-term/src/ui/document/cursor.rs
```diff
@@ -0,0 +1,194 @@
...
+        }
+    }
+
+    // pub fn byte_pos(&self) -> usize {
+    //     self.graphemes.byte_pos()
+    // }
+
+    // pub fn char_pos(&self) -> usize {
+    //     self.char_pos
+    // }
```
pascalkuthe, line 88: I want to keep these functions for use in future work but  they are currently not used. and hence generate warnings.

archseer:
To update all the view methods with calculations that take wrapping and widgets into account we might want to design something similar to Xi's line cache:

https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/width_cache.rs
https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/line_cache_shadow.rs
https://github.com/xi-editor/xi-editor/blob/master/rust/core-lib/src/linewrap.rs

They did async line wrapping which got pretty complicated (https://raphlinus.github.io/xi/2020/06/27/xi-retrospective.html#async-is-a-complexity-multiplier) but at least it might give you some ideas. You can then query the cache for measurements to figure out visual offsets

archseer:
> Maybe some tests, although I am not sure how to actually test this yet (maybe asserting the contest of a surface after render?)

`tui` came with a TestBackend (https://github.com/helix-editor/helix/blob/master/helix-tui/src/backend/test.rs) that we should enable in e2e tests (currently we compile the editor without the render function instead). Then you'd be able to do asserts based on screen contents

helix-term/src/ui/document.rs
```diff
@@ -0,0 +1,207 @@
...
+            }
+            _ => IndentLevel::None,
+        };
+        self.advance_line(indent_level, text_render);
+        text_render.skip(self.config.soft_wrap.wrap_indent);
+    }
+
+    /// Returns whether this document renderer finished rendering
+    /// either because the viewport end or EOF was reached
+    pub fn finished(&self) -> bool {
```
archseer, line 117: Nit: let's use `is_finished`, stdlib usually uses the `is_` prefix for these types of methods

helix-term/src/ui/document.rs
```diff
@@ -0,0 +1,207 @@
...
+            return;
+        }
+
+        loop {
+            while let Some(word_boundry) = self.cursor.advance(if self.config.soft_wrap.enable {
+                text_render.space_left()
+            } else {
+                // limit word size to maintain fast performance for large lines
+                64
+            }) {
```
archseer, line 138: Let's move the `if` condition before the `while let`

helix-term/src/ui/document.rs
```diff
@@ -0,0 +1,207 @@
...
+                if self.config.soft_wrap.enable && word_boundry == WordBoundary::Wrap {
```
archseer, line 139: word_boundary

helix-term/src/ui/document/text_render.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+    }
+}
+
+#[derive(Debug)]
+pub enum Grapheme<'a> {
+    Space,
+    Nbsp,
+    Tab,
+    Other { raw: Cow<'a, str>, width: u16 },
```
kirawi, line 92: Is it worth storing the `width` as a field? As `unicode-width` uses a lookup table so it should be O(1).

archseer: It's a bit more expensive than that, it has a few layers of subtables (0,1,2) + it has to unpack the subtable

pascalkuthe: As @archseer mentioned it's not quite free and the cost of storing the `u16` here is not that large. In fact due to alignment (`Cow` is aligned to `8` bytes) it might actually be free as the size of the enum might not change at all (didn't check tought).  Note that if we cared about the size of this enum enough for that to matter using [beef](https://github.com/maciejhirsz/beef) to get a faster/smaller `cow` would be more worthwhile then saving the `width`. It also doesn't introduce any complexity in the code (it reduces it in fact) so I don't see a downside.

blitzerr:
@pascalkuthe thanks a lot for this PR. This is one of the most desirable features. I am very exited about this hope this gets merged soon. Let me know if there is something I can do to help.

xJonathanLEI:
@blitzerr For starter, help test the PR by daily driving with it? That's the easiest way to help get the PR across the finish line.

xJonathanLEI:
Not sure if this PR is ready to receive bug reports given its draft status, but here is one anyways. I'm using current PR head ([81fa643](https://github.com/helix-editor/helix/pull/5008/commits/81fa6431990a475e2643b41a1eb5a46428d9e5ff)) rebased on `master` and fixes for the build errors.

There problem is there are unexpected view shifts when the cursor is on at least the 3rd visual line of soft-wrapped lines.

To illustrate the issue, first make sure you're view width is small. I used `100` when testing this:

```console
$ tput cols
100
```

Then open the editor with this long text:

```console
$ curl -L https://github.com/helix-editor/helix/files/10274419/test.txt | hx -c /dev/null
```

With the editor open, press `gl` to go to the line end.

Expected result: the view itself is not moved and the cursor position is the only thing changed.

Actual result: the view is shifted downwards.

Perhaps this is best illustrated with a asciinema recording:

[asciicast](https://asciinema.org/a/547050)

gabydd:
was looking at your error and I found an interseting one of my own:
1. enter a scratch buffer and set the language to markdown
2. insert ``# <and a continuous amount of the letter a>``
3. at some point it will panic with:
```
thread 'main' panicked at 'assertion failed: `(left == right)`
  left: `81`,
 right: `1`: Highlight scope must be aligned to grapheme boundary', /home/gaby/dev/helix/helix-core/src/doc_cursor.rs:313:17
```
with the relevant backtrace
```
2: core::panicking::assert_failed_inner
   3: core::panicking::assert_failed
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:182:5
   4: helix_core::doc_cursor::DocumentCursor<S,A>::advance_with_highlight
             at ./helix-core/src/doc_cursor.rs:313:17
   5: helix_term::ui::document::DocumentRender<A,H>::render_line
             at ./helix-term/src/ui/document.rs:107:40
   6: helix_term::ui::editor::EditorView::render_text_highlights
             at ./helix-term/src/ui/editor.rs:478:13
   7: helix_term::ui::editor::EditorView::render_view
             at ./helix-term/src/ui/editor.rs:158:27
   8: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
             at ./helix-term/src/ui/editor.rs:1350:13
```
4. the left value doesn't seem to be random if you do ``tput cols`` it is 6 less then that
5.  it can only be reproduced when the text traversed is being highlighted
I am going to try and do some more digging tomorrow but I wanted to keep track of the issue,

(also for the above issue ensure cursor in view isn't changed to work with this pr yet so that might be part of the problem I am not sure, making this all behaves well with more then two lines of softwrap does seems important for the future though)

pascalkuthe:
Thank you for the bug reports. I am mostly aware of these limitations already, these will be fixed once I finish working on this.

This branch is not yet ready for daily driving and I know its both limited and buggy. Further changes are necessary and there is even some design work left to do.

In fact the latest push was mostly to get my local dev state into github so I can work on it on my laptop (hence the all-red CI) while I visit my family during the holidays. But I likely won't have time to finish this PR until new year.

pascalkuthe:
While further working on this PR the implementation changed enough that I felt it was more appropriate to create a new PR rather than to push to this branch. Closed in favor of #5420

PR #5019, opened 2022-12-05
Update `diagnostic.error` background for acme theme

two-six:
Previously: 
acme2

After change:
acme1


archseer:
This still seems wrong to me, errors should be red?

archseer:
`"diagnostic.warning"` and `"diagnostic.info"` are still broken and will render as white too

two-six:
> This still seems wrong to me, errors should be red?

I experimented a bit and came up with something like this:
image

the-mikedavis (approved):

PR #5021, opened 2022-12-06
Move diff before line numbers by default

maaslalani:
Hey, thanks for the awesome project. I'm really enjoying using Helix as my editor.

When using the latest git diff gutters feature I found it great but generally
noticed it looked a tiny bit strange and not what I was used to.

I believe most of the other software that implements this feature, for example:

* https://github.com/lewis6991/gitsigns.nvim
* https://github.com/airblade/vim-gitgutter

place the git diff signs behind the line numbers which (in my opinion) looks a
bit better.

This PR changes the defaults to what I would (personally) expect.

Feel free to close this PR if you believe that the defaults are correct. It's
easy enough to configure but thought it would be nice to have these as the
defaults, since I presume this is what most people will want (I may be totally
wrong though).

This is what the proposed defaults look like:

image


archseer:
We specifically chose the current design, it also works similarly to how it's rendered in VSCode. The reason for such defaults in vim is because I'm not even sure if you can configure gutters to show after the line numbers

maaslalani:
Ah I see, thank you for the response! That makes sense.

PR #2377, opened 2022-05-03
File explorer and tree helper

cossonleo:
Add a explorer  wdiget that display file under the current dir with tree style or list style.
Add two shortcuts, `Ctrl-e` and `Ctrl-E` to toggle file explorer.
features
`[ctrl]-jk/tab/arrow keys` to switch selected item
`f ` key to fitler items
`b` to back to parent dir
`z` to fold current level items
`? / ` keys to search items
`rd/rf ` to delete files/dirs but you must type `YES` to confirm.
`mf/md` to create file/dir

There are four explore style:
embed + tree:
embed_tree

embed + list: 
embed_list

overlay + tree:
overlay_tree

overlay + list:
overlayt_list

PR #5022, opened 2022-12-06
Add mode separator to statusline

neva1ack:
Add a purely stylistic and configurable Unicode separator after the mode in the statusline.

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -200,18 +200,17 @@ where
                 let mut c = s.clone();
                 c.fg = s.bg;
                 c.bg = s.fg;
-                return c;
+                c
             });
 
-            let mode_separator_string = match config.statusline.mode_separator {
+            let mode_separator = match config.statusline.mode_separator {
                 ModeSeparator::Angled => "",
                 ModeSeparator::Slanted => "",
-                ModeSeparator::Rounded => "",
+                ModeSeparator::Round => "",
                 ModeSeparator::Flat => unreachable!(),
-            }
-            .to_string();
+            };
 
-            write(context, mode_separator_string, mode_separator_style);
+            write(context, mode_separator.to_string(), mode_separator_style);
         }
     };
 }
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -9,6 +9,7 @@ use helix_view::{
 
 use crate::ui::ProgressSpinners;
 
+use helix_view::editor::ModeSeparator;
 use helix_view::editor::StatusLineElement as StatusLineElementID;
 use tui::buffer::Buffer as Surface;
 use tui::text::{Span, Spans};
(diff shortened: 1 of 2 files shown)


archseer:
Isn't this solved by the existing separator component?

neva1ack:
> Isn't this solved by the existing separator component?

No, I couldn't achieve anything similar with the existing separator component. I have attached images of what it looks like. Omitting the `mode-separator = "..."` will result in the same behaviour as `flat`.

Round:
round

Slanted:
slanted

Flat (default):
flat(default)

Angled:
angled

CptPotato:
Looks pretty neat!

I think in the long run a more sophisticated way to configure/theme the statusline would be cool to see. That being said, it's a simple addition and I'd personally use it.

Have you considered adding different powerline symbols or allowing user defined characters for the separator? The only issue I'd see with this is that some might require inverting the fg/bg colors.

helix-term/src/ui/statusline.rs
```diff
@@ -179,16 +189,30 @@ where
...
     );
+
+    match config.statusline.mode_separator {
+        ModeSeparator::Flat => {}
+        _ => {
+            // invert the mode style background and foreground
+            let mode_separator_style = mode_style.map(|s| {
+                let mut c = s.clone();
+                c.fg = s.bg;
+                c.bg = s.fg;
```
CptPotato, line 202: Instead of using the mode's `fg` this should be the `bg` of `ui.statusline` to render correctly. Currently it displays like this:

image

neva1ack: Fixed and tested with other themes, commit: cf679f0.

helix-term/src/ui/statusline.rs
```diff
@@ -179,16 +189,30 @@ where
...
+                bg: context.editor.theme.get("ui.statusline").bg,
+                ..Default::default()
+            });
+
+            let mode_separator = match config.statusline.mode_separator {
+                ModeSeparator::Angled => "",
+                ModeSeparator::Slanted => "",
+                ModeSeparator::Round => "",
+                ModeSeparator::Flat => unreachable!(),
+            };
```
the-mikedavis, line 211: Like the current separator, this can just be specified as an arbitrary string rather than presets

CptPotato:
There still seems to be a coloring issue when using splits. The inactive split uses the inactive statusline `fg` as separator color, even though the mode is not shown:

image

helix-term/src/ui/statusline.rs
neva1ack: I agree but I also think the presents are very reasonable and most 90% of people would be using one of the presents. I think it would be best to do a hybrid approach that lets arbitrary strings be used but also provides presets. See commit 1c2b828 for what this might look like.

the-mikedavis: These presets require a patched font to work (they're in the private use area of unicode) which is not acceptable for core. See https://github.com/helix-editor/helix/pull/2211#issuecomment-1106015472, https://github.com/helix-editor/helix/pull/2377#issuecomment-1128819954. A free-form string allows these without building it in to core.

helix-term/src/ui/picker.rs
```diff
@@ -583,10 +583,10 @@ impl<T: Item + 'static> Component for Picker<T> {
         cx.editor.reset_idle_timer();
 
         match key_event {
-            shift!(Tab) | key!(Up) | ctrl!('p') => {
+            shift!(Tab) | key!(Up) | ctrl!('p') | ctrl!('k') => {
                 self.move_by(1, Direction::Backward);
             }
-            key!(Tab) | key!(Down) | ctrl!('n') => {
+            key!(Tab) | key!(Down) | ctrl!('n') | ctrl!('j') => {
```
archseer, line 589: This seems unrelated to the PR and should be removed

cd-a:
Is there a reason why this was suddenly closed?

silverhairs:
@0xrsp an plan to continue working on this?

0xrsp:
> @0xrsp an plan to continue working on this?

works well enough in current state for me

PR #5027, opened 2022-12-06
Use OSC 52 for tmux copy

xJonathanLEI:
Currently the tmux clipboard provider does not make use of OSC 52 when yanking content to clipboard, which results in the content only going into the tmux buffer but not the actual system clipboard.

This PR adds `-w` to the argument list, which according to the man page:

```doc
If -w is given, the buffer is also sent to the clipboard for target-client using the xterm(1) escape sequence, if possible.
```


archseer:
I think this replaces https://github.com/helix-editor/helix/pull/1343 \cc @Flakebi

PR #5037, opened 2022-12-07
fix(themelint): add diff for git gutter

AlexanderBrevig:
Ref https://github.com/helix-editor/helix/issues/4972 this adds a check for `diff.[plus|minus|delta]` and for them to be different.

### DIFF
--- a/xtask/src/themelint.rs
+++ b/xtask/src/themelint.rs
@@ -50,6 +50,12 @@ fn get_rules() -> Vec<Require> {
         // Check for visible cursor
         Require::Existence(Rule::has_bg("ui.cursor.primary")),
         Require::Existence(Rule::has_either("ui.cursor.match")),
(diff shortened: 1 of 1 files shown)


pascalkuthe:
this pr has been stale for over a year and I don't see us moving forward with this. We have pretty good diff gutter symbols these days. We don't really want to focus on the themelint (and may remove it entirely in the future). Thanks for contributing

AlexanderBrevig:
FWIW; I'm totally fine with removing the themelint. It did not pan out as I hoped since it never got to be part of CI.

I do think it's a shame that some features feel buggy on some/most themes (the recent jump labels f.ex) but as it takes 5 minutes to find out how to fix it locally I guess it's just one of those things... :) Down the line I guess themes will be plugin, which places the blame more clearly on the theme and not on Helix 👍🏻

PR #5041, opened 2022-12-07
Buffer navigation through clicking on bufferline

### LINKED ISSUE #4942: Buffer navigation through clicking on bufferline
Navigating through buffers via the keyboard is fast accurate. However there are cases where one would like to navigate through mouse:

- when using one hand
- when the keyboard is hard to use or simply not available (e.g. touch screen devices)

Being able to click on a bufferline item to navigate to the buffer solves this. The nvim bufferline plugin has this and it's handy.

imelinonkaneko:
Closes #4942


archseer:
This is not very efficient because you're iterating over the line and calculating offsets each time. This should probably be precomputed

Sympel:
I thought that iterating and calculating it on click would be an easier option and that the slight inefficiency would be fine since the rendering code iterates over the bufferline in the same way every time the editor is rendered again and this does it only once when you actually click on the bufferline. I did not like storing data that is constantly changing and is used very rarely. If you think otherwise, you can probably just close this as #5274 seems to do it with precomputed widths.

AddictArts:
It would be nice if this could be reviewed again or looked at. The feature is useful and a direction on implementation before yet another pull request is made would be nice to know. Unless of course a pull request in the queue is sufficient.

pascalkuthe:
Closing this as its quite stale.The bufferline or mouse support are also not a priority.  Thank you for contributing!

PR #5042, opened 2022-12-07
Fix LSP completions ignoring auto-completion option

Jummit:
Previously completions would still show when `auto-completion` was set to `false`.


the-mikedavis (approved):
Oh I see, we check this flag for completion on the idle-timeout event but not for trigger-based completion. Good catch, thanks!

Issue #133, opened 2021-06-05
normal mode hardcoded with english layout


whoizit:
this bug is thousands of years old, it is present in vi / vim / nvim and in the post-modern editor it is also present. This is super sad.
It must use key codes, not key chars.

kirawi:
To be fair, won't most programmers already be typing Latin characters?

whoizit:
so is this a text editor or a code editor? Constantly switching the layout to get work in default mode is a pain for everyone in the world who speaks more than just English. And who writes more than just code.

kirawi:
I believe Helix is moreso a code editor than a text editor, but you make a valid point regardless. This should definitely be a feature in the future.

itzmjauz:
> You make a valid point, this should definitely be a feature. However, it's unclear to me why you wouldn't be able to use alternate mappings on vim/neovim/etc. I'd think this is probably the route Helix should go as well.

Alternate mappings are the sensible solution to this problem(wouldn't call it a bug), the alternative is a very bloated base configuration. Especially given the way crossterm keycode mappings work https://docs.rs/crossterm/0.19.0/crossterm/event/enum.KeyCode.html

archseer:
As someone that uses an alternate layout too (Norman) I can sympathize with the problem. This needs some changes in crossterm: We'd need to be able to receive raw position-based key codes but crossterm only offers us key chars.

pickfire:
I also use an alternative layout (dvorak) and use IME to type chinese and japanese. I don't see an issue here besides the invalid IME position. But yeah, maybe non-latin (but non-pictographs) are an issue. What do you suggest?

Kethku:
I've recently been dealing with this issue on other projects. The good news (once a gui lands) is that winit will very soon have the tools to fix this problem for real (this wasn't the case a few months ago). I'm interested in helping out in this area as its fresh in my mind at least on the gui side

pickfire:
> so is this a text editor or a code editor? Constantly switching the layout to get work in default mode is a pain for everyone in the world who speaks more than just English. And who writes more than just code.

I speak chinese and malay as well but I don't have any issue with the defaults. At least I still use a US keyboard so this must have affected those that use non-us keyboard like in europe.

I think there should be a flag. I still prefer to use key chars rather than keycode. The default keys for dvorak is still good for me to use it.

[...]

PR #5046, opened 2022-12-07
implement keyboard layout remapping

sinefabula:
This patch adds a config option to remap keys from a different keyboard layout into English so that keybindings work even with the keyboard switched into a non-English layout.

The corresponding config option is `editor.layout-remap` which is a list of dictionaries with two mandatory keys, `from` and `into`, which specify the translation mapping, e.g.:

```toml
[[editor.layout-remap]]
from = 'йцукенгшщзхъЙЦУКЕНГШЩЗХЪ'
into = 'qwertyuiop[]QWERTYUIOP{}'
```

These can be repeated multiple times to facilitate specifying mappings containing different types of quotes and other special characters which may require escaping in TOML config. Naming (`from` and `into`) is chosen to be of the same length so that the letters are aligned when using a monospace font, and the config is easier to write.

These translation mappings are applied in every mode except for (obviously) the Insert mode.

This circumvents #133 in a way that while Helix still does not recognise keypresses by their scan-codes but yet it allows for the non-English layout users to operate the editor without switching the layout which can be especially useful for writing documentation in their native language with Helix.

One caveat with this patch is that you can't remap keys already present in the English layout because Helix is not aware of the current keyboard layout and will apply translation every time, thus messing with the original mapping. This patch prevents such behaviour by disabling corresponding bindings and reporting these cases to the Helix log file. The other option would be to fail but I thought it would be safe to just filter them and log it.

A complete configuration example for macOS ISO keyboard layout looks like this. I've been using this for over a month now and it works great. Of course it is not the best solution of #133 but it does like 80% of the job, covering most used cases.

```toml
[[editor.layout-remap]]
from = '<!"№%:,.;()_+'
into = '±!@#$%^&*()_+'

[[editor.layout-remap]]
from = 'йцукенгшщзхъЙЦУКЕНГШЩЗХЪ'
into = 'qwertyuiop[]QWERTYUIOP{}'

[[editor.layout-remap]]
from = 'фывапролджёФЫВАПРОЛДЖЭЁ'
into = 'asdfghjkl;\ASDFGHJKL:"|'

[[editor.layout-remap]]
from = 'э'
into = "'"

[[editor.layout-remap]]
from = ']ячсмитьбю/[ЯЧСМИТЬБЮ?'
into = '`zxcvbnm,./~ZXCVBNM<>?'
```


SoraTenshi:
My 2 cents:
I don't really think the way this is configured is optimal, i would love to have a clear distinction between which row is being remapped.

Something like this:
```toml
[[editor.layout-remap-r1]]
from = '<!"№%:,.;()_+'
into = '±!@#$%^&*()_+'

[[editor.layout-remap-r2]]
from = 'йцукенгшщзхъЙЦУКЕНГШЩЗХЪ'
into = 'qwertyuiop[]QWERTYUIOP{}'
```
etc.

idoric:
I'm using the bépo layout (some kind of french dvorak), and where you find the y on a qwerty layout there is a dead key (which doesn't generate a character by itself) on the bépo. Am I mistaken in thinking that this PR would not solve this scenario?

Also, I don't understand what the caveat is.

AmitDIRTYC0W:
Will this be merged any time soon? I'm waiting to switch to Helix and Workman support is the dealbreaker

BreakingLead (commented):
I've tested, it's good.

sinefabula:
@SoraTenshi I'm afraid I don't get your point on explicit row mapping. Why would you want that? I just allowed the user to split the mapping into sections for easier reading and writing; these sections do not have to be rows. I'm coming from vim here where you can't do this and you end up with a really long `langmap` line which is very hard to fix if there's a mistake.

@idoric I guess you're right, dead key won't generate a keystroke therefore I don't think you'd be able to remap it this way. This whole remapping thing, it's still a hack, a good one (in my opinion at least), but a hack nevertheless!

@BreakingLead glad it worked for you too! I'm using it for several moths straight and have zero problems. It's quite a simple piece of code actually.

@soloturn yeah you're right, I'd very much prefer that too. I just didn't see that coming in any near future so I felt this workaround was essential for me (and perhaps many others who write in a language other than English). I'm not sure at the moment how hard it would be to implement these scancodes. I certainly would like to try it so thanks for pointing to the library. But I guess someone may have already tried this?

SoraTenshi:
> @SoraTenshi I'm afraid I don't get your point on explicit row mapping. Why would you want that? I just allowed the user to split the mapping into sections for easier reading and writing; these sections do not have to be rows. I'm coming from vim here where you can't do this and you end up with a really long `langmap` line which is very hard to fix if there's a mistake.

My idea here was to map it based around the actual, physical keyboard 'rows', so that there is not a real ambiguity.
However, this is just an idea I had.

soloturn:
@sinefabula often only one dead key looks a challenge, though i think the PR should be merged. 

for dead keys, i found another file with keylayouts and wonder if this would help specify a dead key in the from mapping? lets pick out one example, the diaresis dead key, on swiss german keyboard at the location of english right brace, ].

```
#define ASCII_5D	KEY_RIGHT_BRACE				// 93 ]

#ifdef LAYOUT_GERMAN_SWISS

#define DIAERESIS_BITS		0x0300
#define DEADKEY_DIAERESIS	KEY_RIGHT_BRACE
#define ISO_8859_1_EB	DIAERESIS_BITS + KEY_E			// 235 ë     e DIAERESIS  TODO: check this
```

links:

- https://github.com/PaulStoffregen/cores/blob/master/teensy3/keylayouts.h
- https://github.com/Chris-Ricketts/keyboard-layouts

soloturn:
@pascalkuthe anything missing to merge this?

pascalkuthe:
I don't have merge permissions so mostly a review from @archseer. From my side tKS We have a lot on our plate right now as we are heading towards the next release but I am sure he will get to your PR eventually.

helix-view/src/editor.rs
```diff
@@ -74,6 +75,52 @@ where
...
+        .filter_map(|(_, to)| pairs.contains_key(to).then(|| *to))
+        .collect::<Vec<_>>()
+        .into_iter()
+        .for_each(|c| {
+            warn!(
+                "Key '{}' removed from layout-remap as it would overwrite Latin layout binding",
+                c
+            );
+            pairs.remove(&c);
+        });
```
the-mikedavis, line 107: Is this part actually necessary? If you wanted to remap between keyboard layouts for example:

```toml
[[editor.layout-remap]]
# qwerty
from = "qwertyuiop[]asdfghjkl;'zxcvbnm,./"
# dvorak
into = "',.pyfgcrl/=aoeuidhtns-;qjkxbmwvz"
```

you would only rebind `[` and `]` since those are the only characters not duplicated between `from` and `into`. You only translate the incoming key event once so having duplication between `from` and `into` shouldn't matter - it won't cause some recursive loop of remapping.

helix-view/src/editor.rs
```diff
@@ -74,6 +75,52 @@ where
     )
 }
 
+#[derive(Serialize, Deserialize, Default)]
+#[serde(rename_all = "kebab-case", default, deny_unknown_fields)]
+struct LayoutMap {
+    from: String,
+    into: String,
+}
```
the-mikedavis, line 83: I'm very opposed to this representation. It's a lot of indirection two have a `Vec<LayoutMap>` turn into a `HashMap<char, char>` when we could just use a TOML table. This is less verbose than a TOML table but we don't really need to optimize for concise config - I would rather have it be explicit.

the-mikedavis (changes requested):
I left some notes on the changes themselves but IMO this shouldn't be merged. It's something I imagine would be really trivial to do once we can script config: I'm imagining mapping over the keymap data structure and switching the characters using a hashmap - that's similar to how you'd do this in an editor like Kakoune. That would also be more flexible so that you could choose to apply these in any, for example only in insert mode. Also see https://github.com/helix-editor/helix/issues/5520#issuecomment-1867060739.

pascalkuthe:
closing this as this pr has gone stale. As Mike already said I don't see us going forward with something like this in core. This should instead be covered by #10389 in the future

Issue #4914, opened 2022-11-27
Add a command to merge consecutive selections


taupiqueur:

Ordoviz:
In Kakoune, this command is bound to <kbd>alt</kbd>+<kbd>_</kbd>, but I never used it.

DylanBulfin:
https://github.com/helix-editor/helix/pull/5047 pr to add this command

DylanBulfin:
#5047 PR to add this command

the-mikedavis:
Closed by #5047

tahir-hassan:
I am a new user of Helix.

Given I have two cursors in SELECT mode, when I move the cursors and the selections overlap, the selection merges and I end up with one cursor.

What is the point of `merge_consecutive_selections`  then? What is it supposed to do?

DylanBulfin:
`merge_consecutive_selections` came directly from Kakoune and was added to Helix first, while `merge_selections` was added later (#7053) and afaik doesn't have an analogue in Kakoune. So I guess `merge_consecutive_selections` exists more for historical reasons now than practical ones.

PR #5047, opened 2022-12-07
Add command to merge consecutive ranges in selection

DylanBulfin:
My attempt at https://github.com/helix-editor/helix/issues/4914 

Adds command that will merge consecutive ranges in the selection, bound to `Alt-_` for now.

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -521,6 +521,31 @@ impl Selection {
         self
     }
 
+    // Merges all ranges that are consecutive
+    pub fn merge_consecutive_ranges(mut self) -> Self {
+        let mut primary = self.ranges[self.primary_index];
+
+        self.ranges.dedup_by(|curr_range, prev_range| {
+            if prev_range.to() == curr_range.from() {
+                if prev_range == &primary || curr_range == &primary {
+                    primary = curr_range.merge(*prev_range);
+                }
+                *prev_range = curr_range.merge(*prev_range);
+                true
+            } else {
+                false
+            }
+        });
+
+        self.primary_index = self
+            .ranges
+            .iter()
+            .position(|&range| range == primary)
+            .unwrap();
+
+        self
+    }
+
     // TODO: consume an iterator or a vec to reduce allocations?
     #[must_use]
     pub fn new(ranges: SmallVec<[Range; 1]>, primary_index: usize) -> Self {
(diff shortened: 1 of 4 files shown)


kirawi:
I think you can use [`dedup_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by) instead, assuming that ranges are in order.

helix-core/src/selection.rs
```diff
@@ -728,6 +728,29 @@ pub fn split_on_matches(
...
+        if sel.from() == prev_sel.to() {
+            result.push(prev_sel.merge(*sel));
+        } else {
+            result.push(prev_sel);
+            result.push(*sel);
+        }
+    }
+
+    // TODO: figure out a new primary index
+    Selection::new(result, 0)
```
archseer, line 751: You could use something similar to the code in normalize that also handles calculating the new primary index https://github.com/helix-editor/helix/blob/a4de86e7afc0da3d847d467a21ff4fb67a51d620/helix-core/src/selection.rs#L497-L522

archseer: I'd make it modify a selection by value, since most code clones the selection first anyway. If would be almost the same code with the first conditional slightly changed

DylanBulfin: Thanks for the suggestions. I added a calculation for the primary index, but I didn't understand the second comment very well, could you clarify what you mean?

DylanBulfin:
> I think you can use [`dedup_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by) instead, assuming that ranges are in order. I'll check if they are.

I couldn't find a way to use `dedup_by`, but @austinletson helped me with a way to do it in a more elegant way using a fold. Is this more along the lines of what you were thinking?

pascalkuthe:
> > I think you can use [`dedup_by`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.dedup_by) instead, assuming that ranges are in order. I'll check if they are.
> 
> I couldn't find a way to use `dedup_by`, but @austinletson helped me with a way to do it in a more elegant way using a fold. Is this more along the lines of what you were thinking?

The problem with this version is that you are potentially allocating a vector and copying all data.  It might not be super critical but it is pretty easy to avoid here by using `dedup_by` which also is easier to read/clearer to understand (this use-case is exactly what that function is for). `dedup_by`  iterates the vec in order and passes a pair of elements to your function. If your function returns true the first element is removed and only the second one is kept, otherwise both are kept. Furthermore you receive a mutable reference so you can modify both elements. I would use something along the following lines (it didn't test this and might have messed  something up but it should point you in the right direction).

``` rust
self.ranges.deadup_by(|prev_range, curr_range|{
      if prev_range == curr_range  {
          *curr_range = curr_range.merge(curr_range);
          // prev_range will be removed
          true
      } else {
          false
      }
})
```

DylanBulfin:
> The problem with this version is that you are potentially allocating a vector and copying all data. It might not be super critical but it is pretty easy to avoid here by using `dedup_by` which also is easier to read/clearer to understand (this use-case is exactly what that function is for). `dedup_by` iterates the vec in order and passes a pair of elements to your function. If your function returns true the first element is removed and only the second one is kept, otherwise both are kept. Furthermore you receive a mutable reference so you can modify both elements. I would use something along the following lines (it didn't test this and might have messed something up but it should point you in the right direction).
> 
> ```rust
> self.ranges.deadup_by(|prev_range, curr_range|{
>       if prev_range.to() == curr_range.from()  {
>           *curr_range = curr_range.merge(curr_range);
>           // prev_range will be removed
>           true
>       } else {
>           false
>       }
> })
> ```

Thanks for the help! I had missed the fact that `dedup_by` gives mutable refs; reimplemented the command.

archseer:
Oh this is a nice implementation and we could simplify `normalize` in a similar way. The only difference is the conditional: `if prev_range.to() == curr_range.from()` vs `if prev_range.overlaps(curr_range)`

DylanBulfin:
> Oh this is a nice implementation and we could simplify `normalize` in a similar way. The only difference is the conditional: `if prev_range.to() == curr_range.from()` vs `if prev_range.overlaps(curr_range)`

Simplified normalize the same way, happy to move it to a different PR if that would be better.

the-mikedavis:
Could you add a line for the new command to the book? Somewhere in this section would be good: https://github.com/helix-editor/helix/blob/aecb524e503363c2eed2a5a72d8fd881aae18e4b/book/src/keymap.md#L107

DylanBulfin:
Added command to book, and renamed it to `merge_consecutive_selections` to fit better with the nomenclature used in other commands.

the-mikedavis (approved):

PR #5051, opened 2022-12-07
Autumn theme: use new features

getreu:


archseer:
Thanks! Should make maintenance easier :)

PR #5059, opened 2022-12-08
TS-rust move "macro_rules!" to "keyword.directive"

2dav:
Changing queries for Rust macros.
```rust
macro_rules! name{}
name!()
```
as of now  `name` `name!()` both matched as `function.macro` and `macro_rules!` as `keyword`, however macros in Rust are direct analogue to the C's preproccessor directives, which have a dedicated scope `keyword.directive`.

These commits fixes that by making `macro_rules!` and macro-invocations to be matched to `keyword.directive`, so macro-invocations and macro-declarations can be highlighted differently.

> Update
`@keyword.directive` purpose is to match preprocessor-specific keywords like C's `#if`, `#define` 
`@function.macro` as of now serves both `macro declaration` and `macro invocation`

### DIFF
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -158,10 +158,13 @@
 
   "unsafe"
   "default"
-  "macro_rules!"
 
   "async"
 ] @keyword
+
+[
+ "macro_rules!"
+] @keyword.directive
 
 [
   "struct"


runtime/queries/rust/highlights.scm
```diff
@@ -287,11 +290,11 @@
...
 (macro_invocation
   macro: [
-    ((identifier) @function.macro)
+    ((identifier) @keyword.directive)
     (scoped_identifier
-      name: (identifier) @function.macro)
+      name: (identifier) @keyword.directive)
   ]
-  "!" @function.macro)
+  "!" @keyword.directive)
```
archseer, line 297: Invocations shouldn't be `keyword.directive`, since they're not keywords (they're macro invocations = `@function.macro`)

2dav: Yeah, i've thought about `keyword.directive` in a more broad sense, like the place of interest for the preprocessor, while it's purpose is to match preprocessor-specific keywords i.e. `#if #define`, makes sense.

2dav: i'll rebase request to include only commit about `macro_rules!` -> `@keyword.directive`

PR #5061, opened 2022-12-08
Add Haskell text objects

ocharles:


pascalkuthe:
Please regenerate the documentaiton with `cargo xtask docgen`  to make the CI green

the-mikedavis (approved):

the-mikedavis:
> The only thing I don't like about this is that selecting a function seems to select all the whitespace up to the next function. I'm not sure if anything can be done here, or if the tree-sitter grammar needs to be tweaked

Yeah this would need to be changed upstream in the tree-sitter grammar

PR #5066, opened 2022-12-08
use curl underlines for gruvbox_dark_hard

cor:
I love the new curl underlines and I think they look great here in my favorite theme!


pascalkuthe (approved):
always nice to see more themes using the new underlines!

the-mikedavis (approved):

PR #5069, opened 2022-12-08
Some Rust TS queries customization

2dav:
Some tweaks to Rust queries:
- `unsafe` the keyword   => `"keyword.special"`
- `unsafe` block braces  => matched separately with `"punctuation.bracket.special"`
- `macro_rules!`             => `"function.macro"`
- fixes prelude enum variants `Ok, Some, None, Err` matched to an invalid scope
- makes `@type.enum.variant` match enum variant in construction/matching place 
- `'label:` => `"label"`

### DIFF
--- a/runtime/themes/autumn.toml
+++ b/runtime/themes/autumn.toml
@@ -58,10 +58,11 @@
 "markup.raw" = "my_green"
 
 "diff.plus" = "my_green"
-"diff.delta" = "my_white"
+"diff.delta" = "my_gray4"
 "diff.minus" = "my_red"
 
 "diagnostic" = { modifiers = ["underlined"] }
+"diagnostic.error" = { underline = { style = "curl", color = "my_red" } }
 "ui.gutter" = { bg = "my_gray0" }
 "hint" = "my_gray5"
 "debug" = "my_yellow2"
--- a/runtime/themes/autumn_night.toml
+++ b/runtime/themes/autumn_night.toml
@@ -7,67 +7,7 @@
 # Jens Getreu ported and optimised the color theme for the Helix editor.
 # Author: Jens Getreu <getreu at web.de>
 
-"ui.background" = { bg = "my_gray0" }
-"ui.menu" = { fg = "my_white", bg = "my_gray2" }
-"ui.menu.selected" = { fg = "my_gray2", bg = "my_gray5" }
-"ui.linenr" = { fg = "my_gray3", bg = "my_gray0" }
-"ui.popup" = { bg = "my_gray2" }
-"ui.window" = { fg = "my_gray3", bg = "my_gray2" }
-"ui.linenr.selected" = { fg = "my_gray6", bg = "my_gray0"}
-"ui.selection" = { bg = "my_gray3" }
-"comment" = { fg = "my_gray4", modifiers = ["italic"] }
-"ui.cursorline" = { bg = "my_gray3" }
-"ui.statusline" = { fg = "my_gray6", bg = "my_gray2" }
-"ui.statusline.inactive" = { fg = 'my_gray4', bg = 'my_gray2' }
-"ui.statusline.insert" = {fg = "my_black", bg = "my_gray5", modifiers = ["bold"]}
-"ui.statusline.normal" = {fg = "my_gray6", bg = "my_gray2"}
-"ui.statusline.select" = {fg = "my_gray6", bg = "my_black", modifiers = ["bold"]}
-"ui.cursor" = { fg = "my_gray5", modifiers = ["reversed"] }
-"ui.cursor.primary" = { fg = "my_white", modifiers = ["reversed"] }
-"ui.cursorline.primary" = { bg = "my_black" }
-"ui.cursorline.secondary" = { bg = "my_black" }
-"ui.text" = "my_white"
-"operator" = "my_white"
-"ui.text.focus" = "my_white"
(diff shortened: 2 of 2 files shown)


pascalkuthe:
I would really like the general direction of the highlight improvements but I worry that this will look worse on almost all themes initially. With these changes `unsafe` blocks are highlighted as normal function identifiers which looks pretty strange compared to the normal keyword highlights. Perhaps scoping these as special keywords `@keyword.special` might work better? I am also not sure if we want to special highlights for brackets would interact with rainbow brackets (@the-mikedavis), perphas instead of just cloring just make them bold?

runtime/queries/rust/highlights.scm
```diff
@@ -49,6 +49,14 @@
 (loop_label
   (identifier) @type)
 
+; ---
+; `unsafe {}` braces highlighting
+; ---
+
+(unsafe_block
+  (block "{" @function.special
+         "}" @function.special))
```
archseer, line 58: A bracket is not a function: `@punctuation.bracket.special`

runtime/queries/rust/highlights.scm
```diff
@@ -175,6 +181,10 @@
 
 "fn" @keyword.function
 
+"macro_rules!" @keyword.directive
+
+"unsafe" @function.special
```
archseer, line 186: Unsafe is not a function: `keyword.special`

runtime/queries/rust/highlights.scm
```diff
@@ -175,6 +181,10 @@
 
 "fn" @keyword.function
 
+"macro_rules!" @keyword.directive
```
archseer, line 184: I also think this could use a better name, not sure if `macro_rules!` is a directive

pascalkuthe: Macrorules is really just another macro (the fact that it currently isn't is an imolementation detail of the rust compiler) so it should be highlighted the same as any other macro in my opinion.

2dav:
Alright! With introduction of new scopes the things are now looks in-place and don't break the `common theme denominator`.

I see the following change-set so far:
- [ ] change Rust query for `macro_rules!` to be matched to `@function.macro` scope
- [ ] introduce new scopes `@keyword.special` `@punctuation.bracket.special`
  - [ ] update [book section about scopes](https://docs.helix-editor.com/master/themes.html#scopes) 
  - [ ] find good examples in other languages besides Rust's `unsafe`
  - [ ] update themes to use new scopes
  - [ ] implement language-specific queries
    - [ ] Rust's `unsafe` keyword and block 

On the themes update - i'm planning to update **all** the themes to highlight these new scopes using existing palette, so this won't change the look'n'feel, but likely affect the `visual complexity` of the theme, scopes are `special` enough though.

What good use of new `special` scopes in other languages might be?

archseer:
You don't need to modify existing themes since keyword.special falls back onto keyword and so on

2dav:
Yeah, i'm thinking mostly about benefits for Rust language highlighting, introducing `keyword.special` in all the themes will make `unsafe` accentuation the default one, which is a good option imo.

archseer:
That's up to the theme author to decide, I wouldn't change anything. Personally I want to keep the default theme as is

2dav:
These commits includes all discussed so far except for the themes tweaks.

2dav:
From the `first user experience` standpoint it might be useful to get a sense of what is possible and especially what is not in terms of highlighting for a specific language. Maybe queries customization's might be summarized in some sort of the `Demo` theme, which showcase all the supported `scopes` for the language.

2dav:
Fixed enum variant matching query to cover fully qualified paths e.g. `std::io::ErrorKind::NotFound`

2dav:
Changed `@type.variant` to `@type.enum.variant`. `@type.variant` is unused across themes and probably a typo.

runtime/queries/rust/highlights.scm
```diff
@@ -49,6 +49,14 @@
 (loop_label
   (identifier) @type)
```
Chickenkeeper, line 50: While we're at it, how do we feel about changing this to
```scheme
(loop_label
  "'" @label
  (identifier) @label)
```
To match the lifetime labels? I was about to make a PR for this change but it might as well be included in this one instead if possible.

2dav:
Fix for `#![]` where `!` were matched to `"@operator"`.
Matches `#[]` to `"@attribute"` instead of `"@punctuation.bracket"`.

2dav:
Attributes needs love too
```toml
"attribute" = "red"
```
`@attribute` scope should match all attributes, but now it works only for crate-level one
attr

Next, if we add `punctuation` to the theme
```toml
"attribute" = "red"
"punctuation.bracket" = "yellow"
```
attr2
notice "!" in `#![]`. I'd like attribute's punctuation(`#![]`, `#[]`) to be distinct from other punctuation with the new sub-scope`@punctuation.bracket.attribute`.

Futher, attribute's content
```toml
"attribute" = "red"
"punctuation.bracket" = "yellow"
# new
"function.macro" = "blue"
"type" = "green"
"namespace" = "magenta"
```
attr3
one way to fix this mess would be to add `.attribute` to every scope in question e.q. `@type.builtin.attribute` `@namespace.attribute` `@function.macro.attribute` `@punctuation.attribute` and so on. Ideally it should be possible to get the result like on the first screenshot with the all scopes set.

PR #5072, opened 2022-12-08
Reset mode when changing buffers

### LINKED ISSUE #5058: Changeset panic when swapping between buffers in insertion mode
### Summary

In `helix/config.toml` I have:
```
[keys.insert]
"A-left" = "goto_previous_buffer"
"A-right" = "goto_next_buffer"
```

### Reproduction Steps

Key sequence that seems to consistently panic:
```
hx a.txt b.txt
ia<A-right>a<A-right><esc>
```
with message:
```
thread 'main' panicked at 'Position 2 is out of range for changeset len 1!', helix-core/src/transaction.rs:397:13
```

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

kitty 0.26.5

### Helix Version

22.12-3-gf8b75a24

the-mikedavis:
This is similar to the change in e4c9d4082a139aac3aea4506918171b96e81f5b9: reset the editor to normal mode when changing buffers. Usually the editor is already in normal mode but it's possible to setup insert-mode keybindings that change buffers.

Fixes #5058

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -694,7 +694,6 @@ fn goto_buffer(editor: &mut Editor, direction: Direction) {
 
     let id = *id;
 
-    normal_mode_impl(editor);
     editor.switch(id, Action::Replace);
 }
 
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -694,6 +694,7 @@ fn goto_buffer(editor: &mut Editor, direction: Direction) {
 
     let id = *id;
 
+    normal_mode_impl(editor);
```
pascalkuthe, line 697: Maybe it makes more sense to move this into `editor.switch` instead of adding this only here?  I imagine the bug could appear for other commands aswell. `editor.switch` is called many times and it seems like a huge footgut for it to cause crahses it the editor is not in normal mode. I think this bug could already be produced by other commands if you created a custom insert mode binding for them. I can't think of a situation where this wouldn't want to be a normal mode after a switch.

This will require moving `normal_mode_impl` to `helix-view` but that should be okay since the function is fairly small. 

What do you think?

pascalkuthe:
Awesome that you found this so quickly, you are really on a roll recently ironing out panics :+1:

helix-term/src/commands.rs
the-mikedavis: Yeah good call. I think we currently miss some steps when switching modes in `Editor::focus` that would be resolved by that change as well

Issue #5043, opened 2022-12-07
helix 22.12: custom theme fails to deserialize


dssgabriel:
Hi! I updated today to helix 22.12 and my custom theme fails to load, although it worked fine under  helix 22.08.1.

Looking at the log file, I get the following warning at helix's startup:
```
2022-12-07T14:05:08.163 helix_term::application [WARN] failed to load theme `chocolate` - Failed to deserialize theme
```

I rebuilt helix from source today (latest commit to `master` at the moment of writing):
```
$ hx --version
helix 22.12 (a4de86e7)
```

I have no idea where this issue can come from. Reading through the changelog, I do not see anything in the breaking changes that could have messed up my theme configuration. Any idea of what could be causing this?
Otherwise, my config works fine.

<details>
<summary>Below is my theme config file, just in case I missed something in there:</summary>

```toml
# Chocolate (dark)
# Author: Gabriel Dos Santos <dss.gabriel@protonmail.com>
# Originally made by snakedye: https://gitlab.com/snakedye/chocolate
# License: MIT License
# -----------------------------------------------------------------------------

# User interface
# -----------------------------------------------------------------------------
"markup.heading" = "yellow"
"markup.list" = "beige"
"markup.bold" = { fg = "yellow", modifiers = ["bold"] }
"markup.italic" = { fg = "yellow", modifiers = ["italic"] }
"markup.link.url" = { fg = "aqua", modifiers = ["underlined"] }
"markup.link.text" = "aqua"
"markup.quote" = "beige"
"markup.raw" = "green"

"ui.background" = { fg = "white", bg = "black" }
"ui.background.separator" = "lightgray"

"ui.cursor" = { fg = "black", bg = "white" }
"ui.cursor.select" = { fg = "white", bg = "lightgray" }
"ui.cursor.match" = { fg = "white", bg = "lightgray" }
"ui.cursor.primary" = { fg = "black", bg = "white" }

"ui.linenr" = "lightgray"
"ui.linenr.selected" = { fg = "yellow", modifiers = ["bold"] }

"ui.statusline" = { fg = "white", bg = "gray" }
"ui.statusline.insert" = { fg = "gray", bg = "aqua" }
"ui.statusline.select" = { fg = "gray", bg = "magenta" }

"ui.popup" = { fg = "white", bg = "gray" }
"ui.window" = { fg = "white", bg = "darkgray" }
"ui.help" = { fg = "white", bg = "lightgray", modifiers = ["bold", "italic"] }

[...]

PR #5073, opened 2022-12-08
Improve error message handling for theme loading failures

the-mikedavis:
The error messages for a theme that failed to be deserialized (or otherwise failed to load) were covered up by the context/with_context calls:

* The log message for a bad theme configured in config.toml would only say "Failed to deserilaize theme"
* Selecting a bad theme via :theme would show "Theme does not exist"

With these changes, we let the TOML deserializer errors bubble up, so the error messages can now say the line number of a duplicated key - and that key's name - when a theme fails to load because of a duplicated key.

Providing a theme which does not exist to :theme still gives a helpful error message: "No such file or directory."

See #5043

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -777,7 +777,7 @@ fn theme(
                     .editor
                     .theme_loader
                     .load(theme_name)
-                    .with_context(|| "Theme does not exist")?;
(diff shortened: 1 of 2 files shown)


pascalkuthe (approved):
LGTM :+1:  Sorry about closing that issue prematurely it seemed resolved to me on a technical level but a better error message is definitely the real fix to those kinds of issues

Issue #5050, opened 2022-12-07
Tutor is a Blank File


theduke:
### Summary

I started `:tutor` once, and it worked fine.

Then I closed helix and at some point tried `:tutor` again.
Now all I get is a blank file.

Installed with `cargo install --path` directly from a repo clone.

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 22.12 (96ff64a8)

theduke:
Same result when I run `hx --tutor`.

kirawi:
You also need to link your runtime folder: https://docs.helix-editor.com/install.html#build-from-source

theduke:
Ah.
But then why did it work the first time?

Probably the `:tutor` command should print out an error if the tutor file can not be found.

kendfss:
for some reason my default runtime directory points to `.cargo/bin/runtime`, which does not exist, even though the docs say to use `.config/helix/runtime`. I've overwritten the $HELIX_RUNTIME variable but am wondering if should this be raised as a separate issue?

janhrastnik:
In my case I had an obsolete `.config/helix/runtime` directory, with a `tutor.txt` file, but since commit c2e4108, helix searches for a `tutor` file instead, without the .txt suffix. My helix is built from source, if I install it in a different way, then I don't get this issue as it then finds an appropriate runtime directory to use.

Issue #3768, opened 2022-09-09
hx --tutor or :tutor don't work when hx is symlink


bnoordhuis:
### Summary

The tutor doesn't work when hx is invoked through a symlink. The real path works.

### Reproduction Steps

 

I tried this:

1. download and untar helix to $HOME/opt/helix
2. `cd $HOME/bin && ln -s ../opt/helix/hx` (~/bin is on the PATH)
3. `hx --tutor`

I expected this to happen:

See the tutorial. :-)

Instead, this happened:

Nothing. Empty buffer.

### Helix log

No log file was written.

### Platform

macOS

### Terminal Emulator

Terminal.app

### Helix Version

helix 22.08.1 (66276ce6)

the-mikedavis:
Does syntax highlighting work either when `hx` is a symlink? I think the symlink is breaking the detection of the runtime directory. You can set an explicit `$HELIX_RUNTIME` environment variable to the runtime directory path to fix it.

The tutor is loaded from the runtime directory: https://github.com/helix-editor/helix/blob/fae9203c3efd4dee4ac0234090ab9c3163c3e4f7/helix-term/src/application.rs#L160-L163

bnoordhuis:
Syntax highlighting indeed doesn't work.

I believe runtime_dir() ultimately falls back to `std::env::current_exe()`? Can I suggest running that through `std::fs::canonicalize()` to get the real path?

(I can try putting together a PR sometime next week.)

9glenda:
On my nixos install helix is also a symlink (like on every nixos install) and the tutor works
here info about my symlink:
```
[glenda@nixos:~]$ ls -lah $(which hx)
lrwxrwxrwx 1 root root 62 Jan  1  1970 /run/current-system/sw/bin/hx -> /nix/store/skpa4rd1h3s23b1h1v6y5rn58vq1zd5i-helix-22.08/bin/hx
```
If there are any issues with helix being a symlink nixos users will always suffer from them.

the-mikedavis:
Nix builds set the `HELIX_RUNTIME` environment variable ([nixpkgs](https://github.com/NixOS/nixpkgs/blob/1cea3c2362ef8af81d4e66898336094b3f9710ee/pkgs/applications/editors/helix/default.nix#L28), this repo's [flake](https://github.com/helix-editor/helix/blob/ec81ec1e8d26e3a99ed713b79381a77daf817d34/flake.nix#L142))

rayhem:
[...]

PR #997, opened 2021-11-06
`:h[elp]` command

Omnikar:
Closes #668.

Each command with documentation will have it as a file at `runtime/help/command_name.txt`. I'm going for an example-driven documentation style (though not all commands need examples).

### Todos
* [x] `:h <command>` will open documentation for a given command.
* [ ] `:h <keybind>` will open documentation for the command bound to the keybind. Keys in the keybind will be separated by spaces.
* [ ] Document commands
  * [ ] no_op
  * [ ] move_char_left
  * [ ] move_char_right
  * [ ] move_line_up
  * [ ] move_line_down
  * [ ] extend_char_left
  * [ ] extend_char_right
  * [ ] extend_line_up
  * [ ] extend_line_down
  * [x] copy_selection_on_next_line
  * [ ] copy_selection_on_prev_line
  * [ ] move_next_word_start
  * [ ] move_prev_word_start
  * [ ] move_next_word_end
  * [ ] move_next_long_word_start
  * [ ] move_prev_long_word_start
  * [ ] move_next_long_word_end
  * [ ] extend_next_word_start
  * [ ] extend_prev_word_start
  * [ ] extend_next_long_word_start
  * [ ] extend_prev_long_word_start
  * [ ] extend_next_long_word_end
  * [ ] extend_next_word_end
  * [ ] find_till_char
  * [ ] find_next_char
  * [ ] extend_till_char
  * [ ] extend_next_char
  * [ ] till_prev_char
  * [ ] find_prev_char
  * [ ] extend_till_prev_char
  * [ ] extend_prev_char
  * [ ] repeat_last_motion
  * [ ] replace
  * [ ] switch_case
  * [ ] switch_to_uppercase
  * [ ] switch_to_lowercase
  * [ ] page_up
  * [ ] page_down
  * [ ] half_page_up
  * [ ] half_page_down
  * [ ] select_all
  * [ ] select_regex
  * [ ] split_selection
  * [ ] split_selection_on_newline
  * [ ] search
  * [ ] rsearch
  * [ ] search_next
  * [ ] search_prev
  * [ ] extend_search_next
  * [ ] extend_search_prev
  * [ ] search_selection
  * [ ] global_search
  * [ ] extend_line
  * [ ] extend_to_line_bounds
  * [ ] delete_selection
  * [ ] change_selection
  * [ ] collapse_selection
  * [ ] flip_selections
  * [ ] insert_mode
  * [ ] append_mode
  * [ ] command_mode
  * [ ] file_picker
[...]

PR #5076, opened 2022-12-08
Add error upon tutor requested without runtime dir

SoraTenshi:
As mentioned in #5050 it would be great to let the user immediately know whether or not the runtime directory is linked.

Might be beneficial especially for new users.

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -72,3 +72,4 @@
 | `:append-output` | Run shell command, appending output after each selection. |
 | `:pipe` | Pipe each selection to the shell command. |
 | `:run-shell-command`, `:sh` | Run a shell command |
+| `:help`, `:h` | Get help for a specific feature. |
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -185,6 +185,11 @@ impl Application {
 
         if args.load_tutor {
             let path = helix_loader::runtime_dir().join("tutor");
+            if !path.exists() {
+                return Err(anyhow::anyhow!(
+                    "No path to the tutor found. Is the runtime directory linked?"
+                ));
+            }
             editor.open(&path, Action::VerticalSplit)?;
             // Unset path to prevent accidentally saving to the original tutor file.
             doc_mut!(editor).set_path(None)?;
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -186,9 +186,7 @@ impl Application {
         if args.load_tutor {
             let path = helix_loader::runtime_dir().join("tutor");
             if !path.exists() {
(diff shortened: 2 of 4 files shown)


the-mikedavis (commented):
I think that the "Is the runtime directory linked?" phrase would be confusing for a new user. I'm not quite sure what the right words are. Maybe we could point to the healthcheck since it prints out the runtime directory?

SoraTenshi:
> Maybe we could point to the healthcheck since it prints out the runtime directory?

Hm. Not really sure about that one, since the confidience of the tutor not being able to launch due to missing runtime dir is pretty high.

the-mikedavis:
I mean that I'm not sure what the error message can say here that would help a new user fix the problem since explaining what the runtime directory is and how to link it usually takes quite a few words (for example https://github.com/helix-editor/helix/issues/3768#issuecomment-1344821332).

We could link to a wiki page or the text in the README but hyperlinks in error messages can be a little brittle. Ideally we could have a `:help` page on the runtime directory and the error message could say "Check `:help runtime` for more information" or something like that

the-mikedavis:
For the scope of this PR though it's probably ok to word this as "Runtime directory not found" (we should avoid "linked" since distros usually setup the runtime directory by setting $HELIX_RUNTIME). Better in-editor docs for the runtime dir can be figured out in the future, this will be a nice improvement over the current blank file behavior on its own 👍

SoraTenshi:
> I mean that I'm not sure what the error message can say here that would help a new user fix the problem since explaining what the runtime directory is and how to link it usually takes quite a few words (for example [#3768 (comment)](https://github.com/helix-editor/helix/issues/3768#issuecomment-1344821332)).
> 
> We could link to a wiki page or the text in the README but hyperlinks in error messages can be a little brittle. Ideally we could have a `:help` page on the runtime directory and the error message could say "Check `:help runtime` for more information" or something like that

I think the help section kind of sounds pretty neat.
I will create a mock up of it in a few hrs.

SoraTenshi:
@the-mikedavis Alright, so i made a basic mockup implementation, however i am unsure how to formulate those message. Do you have an idea?

gabydd:
There actually is already an in progress help pr #997, I was going to mention this yesterday but I totally forgot sorry

SoraTenshi:
> There actually is already an in progress help pr #997, I was going to mention this yesterday but I totally forgot sorry

Oh wow. Didn't know this was a thing yet.
But it also seems like the PR is stale. Maybe it is a good idea to just convert it to draft for now and then merge the changes?

the-mikedavis:
I've been looking at cleaning up that branch a bit and adding a page for the runtime directory https://github.com/helix-editor/helix/compare/master...help-command

We will probably want to embed the page for the runtime within the helix binary (since it should still show when the runtime directory is not there)

PR #5085, opened 2022-12-09
Treat patches as diff files

danyspin97:


the-mikedavis (commented):
Patch syntax is different from diff syntax though, see https://github.com/the-mikedavis/tree-sitter-diff/issues/1

danyspin97:
`.patch` files are not always related to the canonical Linux patch format. In the context of Linux distributions (which is where I use helix the most), it is used for any file that is supported by [`patch`](https://man7.org/linux/man-pages/man1/patch.1.html), including _context_ diffs and _unified_ diffs.

It seems to me that treating all patches as diff would still have correct highlighting in the actual content; the only thing that wouldn't be correctly shown is the metadata, did I understood correctly?

the-mikedavis (approved):
Ah hmm, looks like tree-sitter-diff works ok for patch syntax too, the patch lines just end up as `context` nodes. In the long-run this file-type should use a separate tree-sitter-patch which parses the metadata and injects tree-sitter-diff but I think this is ok to add for now.

Issue #5040, opened 2022-12-07
Differentiate files with the same name in bufferline


norman784:
I think this will make a QoL improvement, now I updated and tried the bufferline and this is the first issue found. For example in a cargo workspace you have multiple `lib.rs`, `main.rs` or also a normal project `mod.rs` files. Maybe we can draw some inspiration from VScode in this regard, I'm not sure how the bufferline.nvim solves it.

PR #5092, opened 2022-12-09
Differentiate between duplicate names in bufferline

DylanBulfin:
Currently, if there are multiple buffers with the same name, the bufferline does not differentiate these names (https://github.com/helix-editor/helix/issues/5040). This PR should fix that; if there are two open buffers with the same name, the relative path will be displayed in the bufferline instead of just the file name.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -683,11 +683,7 @@ impl EditorView {
                 .to_str()
                 .unwrap_or_default();
 
-            if names_map.contains_key(fname) {
-                names_map.insert(fname, names_map.get(fname).unwrap() + 1);
-            } else {
-                names_map.insert(fname, 1);
-            }
+            *names_map.entry(fname).or_insert(0) += 1;
         }
 
         for doc in editor.documents() {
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -680,6 +683,28 @@ impl EditorView {
                 .to_str()
                 .unwrap_or_default();
 
+            if names_map.contains_key(fname) {
+                names_map.insert(fname, names_map.get(fname).unwrap() + 1);
+            } else {
+                names_map.insert(fname, 1);
+            }
```
the-mikedavis, line 690:
```suggestion
            *names_map.entry(fname).or_insert(0) + 1;
```

DylanBulfin: Thanks, updated.

spost-stateless:
Sorry for the bump, but is there anything holding this up from merge?  This feature would be great to have for Typescript development where I tend to end up with a dozen `index.ts` files.  I'd be happy to pick up any additional work if @DylanBulfin is busy.

Issue #983, opened 2021-11-04
Smarter detection of git trees in file picker


wezm:
Related: to:

* #52 
* #187

A real world scenario for your consideration: the repository that I work on in my job contains 36,504 files tracked by git (`git ls-files|wc -l`), 564,380 files total (including build artefacts etc., `find . -type f | wc -l`). Because it's so large I use git worktrees to have multiple worktrees but only one copy of the 3.8Gb git repo. The naïve check for `.git` done here: 

https://github.com/helix-editor/helix/blob/39584cbccdb06b528220a13b643416f3fd5dc3c8/helix-term/src/ui/mod.rs#L132-L137

does not handle this case as the work trees do not have a `.git` directory, which results in the file list being capped at 8k files. There's also the issue raised on #187 that the check does not work when you are not in the top level of the project.

In my vim config I use https://github.com/lotabout/skim.vim this readily handles extremely large file tree (such as accidentally running it in my home directory) and since the UI is responsive while results are streaming it it's easy to hit Esc and cancel it if this was a mistake.

I guess my point to all of this is in an ideal scenario the file limit would be removed (thus removing the need for the `.git` check) and the UI would support cancellation of loading large trees.

archseer:
Should be fixed by https://github.com/helix-editor/helix/pull/5097

PR #5097, opened 2022-12-09
properly handle detachted git worktrees

pascalkuthe:
Right now we check if `.git` exists and is a directory to detect a git repo root.
This does not work well for git repos with multiple/detached work-tree where `.git`
is a file which contains the location of the folder that contains files that would normally be located in `.git`.

This PR changes all checks for `.git` to use `.exists()` instead of `.is_dir()`, so it also works for files.
Since the `git` cli uses the same detection mechanism (try `tough .git && git status` in an empty directory)
this seems more inline with the intended behavior.

Furthermore, workflows with multiple worktrees become much nicer with this (I use one for each branch to quickly context switch)
as the file picker now stops at the correct location. 
Fixing this will also be important for getting repo caching to work correctly in the future.

This change is technically a breaking for repos that contain random `.git` files but as this messes with 
the `git` cli as well, so I think those cases can be ignored and I did not mark it as a breaking change.

### DIFF
--- a/helix-core/src/lib.rs
+++ b/helix-core/src/lib.rs
@@ -69,7 +69,7 @@ pub fn find_root(root: Option<&str>, root_markers: &[String]) -> std::path::Path
             top_marker = Some(ancestor);
         }
 
-        if ancestor.join(".git").is_dir() {
+        if ancestor.join(".git").exists() {
             // Top marker is repo root if not root marker was detected yet
             if top_marker.is_none() {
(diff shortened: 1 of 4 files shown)


pascalkuthe:
This should also indirectly fix #983 I believe. This does not remove the limit as requested there or make the picker streaming/async but this should cover the specific usecase of detecting git worktrees as described in that issue.

wezm:
Thanks, this does fix #983 for me.

PR #5105, opened 2022-12-10
update(theme): adjust base16_transparent and dark_high_contrast

GreasySlug:
- Use `underline` instead of `modifier = ["underlined"]`
- Add base theme author sign
- Adjust some colors

base16_transparent

### DIFF
--- a/runtime/themes/base16_transparent.toml
+++ b/runtime/themes/base16_transparent.toml
@@ -65,7 +65,7 @@
 "warning" = "light-yellow"
 "error" = "light-red"
 
-"diagnostic.info" = { underline = { color = "light-blue", style = "dot" } }
+"diagnostic.info" = { underline = { color = "light-blue", style = "dotted" } }
 "diagnostic.hint" = { underline = { color = "light-gray", style = "double_line" } }
 "diagnostic.debug" = { underline ={ color ="light-gray", style = "dashed" } }
 "diagnostic.warning" = { underline = { color = "light-yellow", style = "curl" } }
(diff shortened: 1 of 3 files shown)


runtime/themes/base16_transparent.toml
```diff
@@ -53,19 +54,19 @@
...
+"warning" = "light-yellow"
+"error" = "light-red"
 
-"diagnostic" = { modifiers = ["underlined"] }
-"diagnostic.info" = { fg = "light-blue", modifiers = ["underlined"] }
-"diagnostic.hint" = { fg = "gray", modifiers = ["underlined"] }
-"diagnostic.debug" ={ fg ="gray",  modifiers = ["underlined"] }
-"diagnostic.warning" = { fg = "yellow", modifiers = ["underlined"] }
-"diagnostic.error" = { fg ="light-red", modifiers = ["underlined"] }
+"diagnostic.info" = { underline = { color = "light-blue", style = "dot" } }
```
the-mikedavis, line 68: I think the documentation is currently incorrect about this underline style, it should be `dotted`: https://github.com/helix-editor/helix/blob/f323ffabcc5846f647e3ba1d3d2ff5dec95257bd/helix-view/src/graphics.rs#L323

pascalkuthe: Yeah I think I mixed that up :sweat_smile:  I think the documentation was what I originally planning. We use `curl` and `line` so using the passive `dotted` and `dashed` seemed inconsistent. Perhaps this should be changed to `dashes` and `dots` in both documentation and implementation?

the-mikedavis: I think they're ok as passive - it makes sense to me as "dotted line" or "dashed line". When in doubt I would say we should follow whatever kitty says they are but it looks like then we would need to change to "curly" too for consistency: https://sw.kovidgoyal.net/kitty/underlines/

I think that just updating the documentation to say "dotted" would be the best choice to minimize breaking changes

GreasySlug:
Also added dark_high_contrast

dark_high_contrast

the-mikedavis (approved):

PR #5110, opened 2022-12-10
Adds a diagnostics severity marker to diagnostics picker

### LINKED ISSUE #3543: Sort diagnostics by gravity (error -> warning -> info)

When there are lots of warnings or infos in a project, the error diagnostics can be lost and hard to find in the diagnostic pickers.

They could be sorted by gravity to help go to the errors first.

zummenix:
Diagnostics severity markers `E`, `W`, `I`, `H` in diagnostics picker.

As noted in the related issue comments, this allows filtering inside the picker by these markers.
This is also useful for people who have color blindness.

Screenshot 2022-12-10 at 17 52 04
Screenshot 2022-12-10 at 17 52 31
Screenshot 2022-12-10 at 17 53 20

I was trying different formats and styles. Format like `[E]` looks heavy and takes up a precious space. Styling markers with diagnostics color looks a little weird, especially with a file path in workspace diagnostics picker. In any case I'm open to any suggestions.

Closes #3543

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -143,6 +143,19 @@ struct PickerDiagnostic {
     offset_encoding: OffsetEncoding,
 }
 
+impl PickerDiagnostic {
+    fn severity_marker(&self) -> &'static str {
+        let empty_space = "  ";
+        self.diag.severity.map_or(empty_space, |s| match s {
+            DiagnosticSeverity::HINT => "H ",
+            DiagnosticSeverity::INFORMATION => "I ",
+            DiagnosticSeverity::WARNING => "W ",
+            DiagnosticSeverity::ERROR => "E ",
+            _ => empty_space,
+        })
+    }
+}
+
 impl ui::menu::Item for PickerDiagnostic {
     type Data = (DiagnosticStyles, DiagnosticsFormat);
 
@@ -178,6 +191,7 @@ impl ui::menu::Item for PickerDiagnostic {
         };
 
         Spans::from(vec![
+            Span::raw(self.severity_marker()),
             Span::raw(path),
             Span::styled(&self.diag.message, style),
             Span::styled(code, style),


helix-term/src/commands/lsp.rs
```diff
@@ -105,6 +105,23 @@ struct PickerDiagnostic {
...
+            .diag
+            .severity
+            .map(|s| match s {
+                DiagnosticSeverity::HINT => 'H',
+                DiagnosticSeverity::INFORMATION => 'I',
+                DiagnosticSeverity::WARNING => 'W',
+                DiagnosticSeverity::ERROR => 'E',
+                _ => ' ',
+            })
+            .unwrap_or(' ');
```
poliorcetics, line 120:
```suggestion
            .map_or(' ', |s| match s {
                DiagnosticSeverity::HINT => 'H',
                DiagnosticSeverity::INFORMATION => 'I',
                DiagnosticSeverity::WARNING => 'W',
                DiagnosticSeverity::ERROR => 'E',
                _ => ' ',
            });
```

helix-term/src/commands/lsp.rs
```diff
@@ -105,6 +105,23 @@ struct PickerDiagnostic {
...
+        format!("{} ", symbol)
```
poliorcetics, line 121: I'm a little sad we need to allocate all the time for each and every diagnostic.

If `Span::raw` accept `&'static str`, we could simply embed the space in the returned string and do something like:

```rust
            // Two characters wide, so two spaces in the default
            .map_or("  ", |s| match s {
                DiagnosticSeverity::HINT => "H ",
                DiagnosticSeverity::INFORMATION => "I ",
                DiagnosticSeverity::WARNING => "W ",
                DiagnosticSeverity::ERROR => "E ",
                _ => "  ",
            })
```

poliorcetics: As long as not config option is offered for this, it would work well

pascalkuthe: A `Span` is a `Cow` so it is possible to create one from a &'static str without allocating so I also thing we should return a string here.. Even if this was configurable it should be possible to use a reference to the config here

helix-term/src/commands/lsp.rs
zummenix: Nice, thanks!

CptPotato:
FYI, this somewhat overlaps with icon support: https://github.com/helix-editor/helix/pull/2869#issuecomment-1317793866

pascalkuthe:
> FYI, this somewhat overlaps with icon support: [#2869 (comment)](https://github.com/helix-editor/helix/pull/2869#issuecomment-1317793866)

Definitely similar thanks for providing additional context.
The critical difference between those two PRs is that the characters used in this PR can (quickly) typed and used to filter by diagnostic type (with #5114) while the icons are mostly just additional eye-candy. I think the colored text already makes it quite clear what kind of diagnostic I am looking at however I find the additional filtering offered by this PR quite useful.

For people that prefer an icon here I think a simple config option to change the displayed text could be added in the future, although that would disable filtering (there is an option to add different text for filtering then is being displayed but using this for anything but a performance optimization would not only be confusing but also seems to break highlighting of matched incidences).

pascalkuthe (approved):
LGTM :+1:  I like this change. It's a useful way to filter diagnostics that doesn't impede existing use-cases and integrates well with the picker.

poliorcetics:
This has been sitting approved for two months now and it's a faily small change, are there any blockers ?

pascalkuthe:
> This has been sitting approved for two months now and it's a fairly small change, are there any blockers ?

There are loads of PRs that sit approved for a while. The review bandwidth of @archseer and @the-mikedavis is limited. We recive and merge a huge number of PRs. Eventually this PR will be reviewed/merged by them.

the-mikedavis (changes requested):
Eventually we want this and similar info (for example symbol kind in the symbol picker) in the pickers but just concatenating it into the row seems like a bandaid to me. I'd rather see the picker organized as a table so this info can be a column - it's included in the linked table picker PR.

zummenix:
Closing in favor of #7265

PR #5114, opened 2022-12-10
Add optional substring matching for picker

pascalkuthe:
This PR implements common syntax that allows matching substrings instead of fuzzy matching.
I will quote the `fzf` README here for an explanation what exactly this does:

> | Token     | Match type                 | Description                          |
> | --------- | -------------------------- | ------------------------------------ |
> | `sbtrkt`  | fuzzy-match                | Items that match `sbtrkt`            |
> | `'wild`   | exact-match (quoted)       | Items that include `wild`            |
> | `^music`  | prefix-exact-match         | Items that start with `music`        |
> | `.mp3$`   | suffix-exact-match         | Items that end with `.mp3`           |
> | `!fire`   | inverse-exact-match        | Items that do not include `fire`     |
> | `!^music` | inverse-prefix-exact-match | Items that do not start with `music` |
> | `!.mp3$`  | inverse-suffix-exact-match | Items that do not end with `.mp3`    |

All of these can be escaped by inserting a backslash `\!`, `\^`, `\'` and `\$` (not mentioned in the README but tested in practice).
The implementation here closely matches the behavior of both `fzf` and `skim` (which behave identical)

I found myself needing these recently.
In particular, I found myself missing the inverse matches.
I work on a circuit simulator (ngspice) which has some core code and then some device models (also just c code).
When I was refactoring core parts of the simulator I wanted to use `gr` to find all references.
However, this was polluted by thousands of references in the device models (all of which are inside the `devices` folder).
So just being able to add `^devices ` in the picker was invaluable.

While looking into this I found myself warming up to the other options as well, for example with #5110 `^E` can be used to filter only errors or `!^H` can be used to ignore hints.

### DIFF
--- a/helix-term/src/ui/fuzzy_match.rs
+++ b/helix-term/src/ui/fuzzy_match.rs
@@ -4,41 +4,218 @@ use fuzzy_matcher::FuzzyMatcher;
 #[cfg(test)]
 mod test;
 
+struct QueryAtom {
+    kind: QueryAtomKind,
+    atom: String,
+    ignore_case: bool,
+    inverse: bool,
+}
+impl QueryAtom {
+    fn new(atom: &str) -> Option<QueryAtom> {
+        let mut atom = atom.to_string();
+        let inverse = atom.starts_with('!');
+        if inverse {
+            atom.remove(0);
+        }
+
+        let mut kind = match atom.chars().next() {
+            Some('^') => QueryAtomKind::Prefix,
+            Some('\'') => QueryAtomKind::Substring,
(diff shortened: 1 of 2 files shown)


archseer (approved):

LeoniePhiline:
@pascalkuthe Love your example. Wildfire by sbtrkt is truly awesome! :)

pascalkuthe:
> @pascalkuthe Love your example. Wildfire by sbtrkt is truly awesome! :)

I had no idea what that was until you brough it up. The examples are from the FZF README as I am mirroring the fzf/skim syntax here

PR #5119, opened 2022-12-11
Allow ANSI colors in themes

tomleb:
My current vim theme doesn't use rgb colors but instead uses the ANSI color (150, 174, etc). 

Is this something Helix want to be able to do? It seems it already supports some colors (black, etc) but not all the ANSI colors.

### DIFF
--- a/helix-view/src/theme.rs
+++ b/helix-view/src/theme.rs
@@ -351,8 +351,23 @@ impl ThemePalette {
         Self { palette: default }
     }
 
-    pub fn hex_string_to_rgb(s: &str) -> Result<Color, String> {
-        if s.starts_with('#') && s.len() >= 7 {
+    pub fn string_to_rgb(s: &str) -> Result<Color, String> {
+        if s.starts_with('#') {
+            Self::hex_string_to_rgb(s)
+        } else {
+            Self::ansi_string_to_rgb(s)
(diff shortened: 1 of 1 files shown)


tomleb:
Figured how to install `rustfmt` and run `cargo fmt --all --check`. Fixed the lint CI error.

the-mikedavis:
I think it makes sense to add this so that themes can work well with 256-color terminal themes. What do you think @archseer?

A note on ANSI-colors should be added to the book somewhere in the color palettes section https://github.com/helix-editor/helix/blob/ba3c24aa0268735ac57321442d458ab6a1ac662c/book/src/themes.md?plain=1#L33

tomleb:
> A note on ANSI-colors should be added to the book somewhere in the color palettes section

I added a note on this in a new `Colors` section. I'm not sure about the wording, and I'm not making a distinction between regular colors and colors palettes colors.. Maybe someone has better wording for this?

I haven't found yet how to build the book, haven't had much time to search that though.

bresilla:
Is there any timeline when this will be merged? I am just writing here cos I don't want this to be forgotten :)

It's only few lines of code change, and makes possible to use all ANSI (255) colors.

archseer (approved):

Issue #2130, opened 2022-04-16
Keymap change proposal: extending lines in both directions


EpocSquadron:
## Context and Motivation

There have been a few discussions regarding the behavior of `x` in various places (#165, #1638), but it feels like maybe the community is starting to trend in a particular, actionable direction (judging by matrix discussion, new user input, and recent change in kakoune) that I would like to propose in isolation here.

## Goals

- Add upward equivalent of `x` (commonly requested)
- Keep current behavior of `x` accessible (people seem to like it as a shortcut/cheat rather than puzzling out which textobject or tree-sitter nav would work best in a given situation)
- Obtain consistency with how other movement commands interact with extension (general trend in keymap changes historically)

## Dependencies

- (#2117)

### Terminology 

"At bounds" refers to when _all_ selections have their head/cursor/selection start and anchor/selection end on the first or last character (not including the newline) of any line.

#### NORMAL Mode (not at bounds)
- `x` = extend to line bounds, ensure cursor forward
- `X` = extend to line bounds, ensure cursor reverse
- (Optional) `Alt-x` = extend to line bounds, do not alter selection direction

#### NORMAL Mode (at bounds)
- `x` = select next full line (not extending), cursor forward
- `X` = select previous full line (not extending), cursor reverse
- (Optional) `Alt-x` = extend to line bounds, do not alter selection direction

#### SELECTION Mode (not at bounds)
Same as NORMAL (not at bounds)

#### SELECTION Mode (at bounds)
- `x` and `X` same as NORMAL (at bounds), but extending
- (Optional) `Alt-x` = add one line above and below, do not alter selection direction

### Configuration for Testing

Add this to your `config.toml` in combination with running from the #2117 PR in order to see 95% how this would work as described above. The caveat being that the intelligent behavior of only extending the current selections to line bounds on the first press doesn't work as intended for most invocations, as there is no detection possible without code alteration.

```toml
[keys.normal]
x = ["move_line_down", "extend_to_line_bounds"]
X = ["move_line_up", "extend_to_line_bounds"]
"A-x" = "extend_to_line_bounds"

[keys.select]
x = "extend_line"
X = "extend_line_above"
"A-x" = ["extend_line_above", "extend_line"]
```

#### See it in action

Here are some examples of the above config in action, keeping in mind the caveats.

`x` and `X` in NORMAL:

[...]

PR #5122, opened 2022-12-11
Enabled keys for selecting previous lines, changed existing keymap;

yudjinn:
This does some changes to the default keybind to:
A) make use of an existing command
B)  align the Shift modifier and Alt modifier to align with usage. 
Before:
`x` => select current line if unselected, else extend to next line
`X` => extend current selection to end of line
`A-x` => trim current selection to new lines
Now:
`x` => unchanged
`X` => opposite of x... select current line otherwise extend to PREVIOUS line
`A-x` => extend current selection to end of lines
`A-X` => trim current selectiono to end of lines

### DIFF
--- a/helix-term/src/keymap/default.rs
+++ b/helix-term/src/keymap/default.rs
@@ -86,8 +86,9 @@ pub fn default() -> HashMap<Mode, Keymap> {
 
         "%" => select_all,
         "x" => extend_line_below,
-        "X" => extend_to_line_bounds,
-        "A-x" => shrink_to_line_bounds,
+        "X" => extend_line_above,
+        "A-x" => extend_to_line_bounds,
+        "A-X" => shrink_to_line_bounds,
 
         "m" => { "Match"
             "m" => match_brackets,
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -339,6 +339,12 @@ _________________________________________________________________
  3. Move to the fourth line.
  4. Type x twice or type 2x to select 2 lines, and d to delete.
 
+ This also works in reverse.
+
+ 1. Move the cursor to line labeled "7".
+ 2. Type X to select the current line.
+ 3. Type X again to select the previous line, and d to delete.
+
  --> 1) Roses are red,
  --> 2) Mud is fun,
  --> 3) Violets are blue,
@@ -347,8 +353,8 @@ _________________________________________________________________
  --> 6) Sugar is sweet,
  --> 7) And so are you.
 
- Note : X works similarly to x although it doesn't extend to
-      subsequent lines. X on an empty line does nothing.
+ Note : Alt-x works similarly to x although it doesn't extend to
+      subsequent lines.
 
 =================================================================
 =                   3.7 COLLAPSING SELECTIONS                   =


helix-term/src/keymap/default.rs
```diff
@@ -86,8 +86,9 @@ pub fn default() -> HashMap<Mode, Keymap> {
 
         "%" => select_all,
         "x" => extend_line_below,
-        "X" => extend_to_line_bounds,
-        "A-x" => shrink_to_line_bounds,
+        "X" => extend_line_above,
+        "A-x" => extend_to_line_bounds,
+        "A-X" => shrink_to_line_bounds,
```
the-mikedavis, line 91: I used to use a similar configuration when I used Kakoune. Ultimately I don't think that `extend_line_above` is that useful: if you need to extend upward you can go into select mode and use `A-;` (flip_selections) and `k` to extend above.

It's nice to have `X` readily available because you can use it to blindly delete/change a line (since X behaves differently on an empty line and a line with contents).

the-mikedavis (commented):
Also see the discussion in https://github.com/helix-editor/helix/issues/2130

helix-term/src/keymap/default.rs
yudjinn: this change came specifically from a few Matrix conversations where people wanted this functionality, so there is definitely a desire for it. And you can still blindly change a line using xc/Xc? I dont see how X is useful, as it only extends what would have been mid-line selections to full lines. I dont see how this reduces functionality and actually aligns the functionality more

runtime/tutor
```diff
@@ -339,6 +339,12 @@ _________________________________________________________________
  3. Move to the fourth line.
  4. Type x twice or type 2x to select 2 lines, and d to delete.
 
+ This also works in reverse.
+
+ 1. Move the cursor to line labeled "7".
+ 2. Type X to select the current line.
+ 3. Type X again to select the previous line, and d to delete.
```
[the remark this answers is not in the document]
yudjinn: they are? 1.3 is 13 lines, 1.4 is 17, 1.5 is 19, 2.1 is 20 lines long. If there is a reason for keeping 18 length sections, it's a rule thats already broken.

helix-term/src/keymap/default.rs
schecko: > I used to use a similar configuration when I used Kakoune. Ultimately I don't think that `extend_line_above` is that useful: if you need to extend upward you can go into select mode and use `A-;` (flip_selections) and `k` to extend above.

It is apparently important enough to have extend_line_below bound to `x`, but your argument can be be applied to extend_line_below as well? just use v and k to extend down. I don't see why "below" is given importance but "above" isn't? 

`x` is a shortening of a common operation extend_line_below, and extend_line_above isn't that far behind in my mind. full line select, `S-v` in vim would be great in helix too, which might put extend_line_above and extend_line_below lower on the list of important operations. To my knowledge the closest helix has to line select mode is `S-x` although `S-x` will break some muscle memory because it needs to be applied at the end of a selection instead of at the start but I realize that is a personal bias from vim.

yudjinn:
So since you added the 'needs discussion' tag, what do I need to do? What kind of discussion needs to be had/who needs to be involved/ where is it held?

runtime/tutor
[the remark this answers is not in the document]
the-mikedavis: I mean the spacing of the sections rather than the content:

```
=================================================================
=                     X.Y SECTION TITLE                         =
=================================================================
L1
L2
L3
L4
L5
L6
L7
L8
L9
L10
L11
L12
L13
L14
L15
L16
L17
L18
L19
```

looks like it's 19 rather than 18 actually. It looks to me like all sections are the correct length on master currently, no?

helix-term/src/keymap/default.rs
the-mikedavis: `x` and `X` behave differently on empty lines:

```
aaa

bbb
```

If you have your cursor on the line between the 'aaa' and 'bbb' lines, `x` extends the selection to include 'bbb' while `X` keeps the selection only on the EOL character on L2. So if you are trying to act consistently on one line, you should use `X` rather than `x` to avoid selecting multiple lines.

> I don't see why "below" is given importance but "above" isn't?

I don't really have a preference between the two; it feels a little wasteful to me to have both though. It's worth considering using `extend_line` which extends in the direction of the selection (forward goes down, backwards goes up) which would eliminate a keybinding here.

the-mikedavis:
This can be accomplished for those who prefer this keymap with these keybindings in the config:

```toml
# ~/.config/helix/config.toml
[keys.normal]
X = "extend_line_above"
A-x = "extend_to_line_bounds"
A-X = "shrink_to_line_bounds"
```

Whether it should be the default for everyone is worth some discussion, hence the needs-discussion label. That discussion should happen here and I want others to weigh in about whether they like this change and possible alternatives.

A breaking change to something as frequently used as the x family of keys isn't something we should rush into. I personally prefer the current keymap: I think that `extend_to_line_bounds` is more useful than `extend_line_above`.

It would be good to consider the `extend_line` command: it would remove the need for binding both `extend_line_above` and `extend_line_below` but OTOH its behavior might be too complicated to be reliable.

It's also worth considering this similar mapping:

```toml
# ~/.config/helix/config.toml
[keys.normal]
x = "extend_line_below" # current
X = "extend_to_line_bounds" # current
A-x = "extend_line_above" # was shrink_to_line_bounds
A-X = "shrink_to_line_bounds" # was A-x
```

since there is some precedent for Alt keybinds going up when non-Alt keybinds go down with `A-C` and `C` respectively. OTOH there is t/T and f/F that use shift for next/previous which is similar.

yudjinn:
Sorry, OTOH?

the-mikedavis:
Ah sorry, "on the other hand"

pascalkuthe:
closing this issue since there hasn't been much discussion here and I don't think we will be going forward with this. We merged an alternative set of commands already that would be more likely candidates for a changed default (but I would perosnally prefer to either not chaneg the default or change it to our current `X` binding). Thank you for contributing

PR #5126, opened 2022-12-12
Switch Normal Mode when the hx lost focus.

erasin:
It's very useful when we use the mouse to switch windows.
base on [#5072]


archseer:
What if I switched away to copy some text? Or if I press `v` then I accidentally switch away and I'm suddenly not in selection mode anymore? I think the mode should not be changed by focus events

erasin:
If I use shortcut key to switch foucs, It need use i/v to switch old mode.
If I use mouse left key back to hx , usually , I need back to normal mode and move cursor back to old position.

Maybe I should find other way.

PR #5147, opened 2022-12-13
Add Mermaid.js for markdown support

erasin:
<https://mermaid-js.github.io/mermaid/>

### DIFF
--- a/runtime/queries/mermaid/highlights.scm
+++ b/runtime/queries/mermaid/highlights.scm
@@ -126,7 +126,7 @@
  (state_annotation_choice)
  ] @type
 
-(directive) @include
+(directive) @keyword.directive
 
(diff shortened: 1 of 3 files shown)


runtime/queries/mermaid/highlights.scm
```diff
@@ -0,0 +1,187 @@
...
+ (class_annotation_line)
+ (class_stmt_annotation)
+ (class_generics)
+
+ (state_annotation_fork)
+ (state_annotation_join)
+ (state_annotation_choice)
+ ] @type
+
+(directive) @include
```
the-mikedavis, line 129:
```suggestion
(directive) @keyword.directive
```

the-mikedavis (approved):

PR #5161, opened 2022-12-15
tutor: Add a content cycling section

devgioele:
Efforts towards [#506](https://github.com/helix-editor/helix/issues/506).

The tutor section `10.1` tells you to select `would`, which causes later cycling of content make no visible changes with `Alt-(` and `Alt-)`.

Instructions are added to the note, so that the user selects different contents before attempting to cycle them.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -986,7 +986,9 @@ lines.
  --> if a wouldchuck could chuck would?
 
  Note: Additionally, Alt-( and Alt-) cycle the *contents* of the
-       selections as well.
+       selections as well. To see this in action, first select
+       different contents, like the words "would" and "much"
+       of the first line.
 
 =================================================================
 =                     10.2 CHANGING CASE                        =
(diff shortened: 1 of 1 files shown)


runtime/tutor
```diff
@@ -986,7 +986,9 @@ lines.
  --> if a wouldchuck could chuck would?
 
  Note: Additionally, Alt-( and Alt-) cycle the *contents* of the
-       selections as well.
+       selections as well. To see this in action, first select
+       different contents, like the words "would" and "much"
+       of the first line.
```
the-mikedavis, line 991: Each section of the tutor has a fixed height so these additions won't fit. It might be worth having a separate section for `A-(`/`A-)`

devgioele: Done with commit [ee7092d](https://github.com/helix-editor/helix/pull/5161/commits/ee7092d0c4223bf52db7a38af1a1641b59f03f22)

PR #5165, opened 2022-12-15
Allow max number of files seen by file seen by file_picker to be configurable

### LINKED ISSUE #5157: Allow max number of files used by file_picker to be configurable
It is currently an undocumented hard coded value. Allowing it to be configurable has the benefit of documenting it and allowing users to set a greater value. Perhaps with a warning that bad performance could be expected?

https://github.com/helix-editor/helix/blob/012fc12f97f4e8e0fc38af351388e41e8bc142d8/helix-term/src/ui/mod.rs#L214

nstinus:
Fixes #5157

### DIFF
--- a/runtime/queries/vue/injections.scm
+++ b/runtime/queries/vue/injections.scm
@@ -8,13 +8,37 @@
   (raw_text) @injection.content)
  (#set! injection.language "javascript"))
 
+; <script>
 ((script_element
-  (raw_text) @injection.content)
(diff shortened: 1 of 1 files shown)


the-mikedavis:
This should be documented in the book [here](https://github.com/helix-editor/helix/blob/ec9aa6690244bccefac24037c9f7a659816bffdd/book/src/configuration.md#L157)

Issue #2011, opened 2022-04-07
Feature Request: 💡 Light bulb on code actions 💡 


David-Else:
At the moment there is no way to discover if a code action exists other than moving around pressing `space a`. In Neovim this simple plugin https://github.com/kosayoda/nvim-lightbulb shows a little lightbulb when there is a code action available on your cursor position.

It would be great to have this in Helix!

archseer:
We can probably do this in the idle timeout that's also used for completion. Would also be a good place to trigger documentHighlights https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentHighlight

sudormrfbin:
While i like the feature itself, it can also be distracting depending on the scope of code actions sent by the server. rust-analyzer for example has code actions for flipping a binary comparison when the cursor is on a `==`.

David-Else:
How about a `space + t` for toggle and place any options that want toggling there? Something like:

`space + t + a` toggle lightbulb on code action
`space + t + l` toggle line numbers
`space + t + d` toggle lsp diagnostics

matklad:
>While i like the feature itself, it can also be distracting depending on the scope of code actions sent by the server. rust-analyzer for example has code actions for flipping a binary comparison when the cursor is on a ==, which I wouldn't want be reminded about since it's pretty trivial.


Yeah, that's the main UX problem with :bulb: in the limit, there's an infinite amount of assists for all imaginable cases, and it's hard to teach the user what's available. 

One idea I had a while back, is to create a tutorial-like experience, with the editor maintaining the set of code actions which the user invoked at least once. That way, a :bulb: is shown only for "new" assists. This allows to both: 

* teach new users about all existing assists
* incrementally teach users about _new_ assists which get implemented by the language server

David-Else:
@matklad A long time has passed since this thread was created, now we have an event system would it be easier to implement the lightbulb, are you still using Helix and is this something you might consider doing a PR for at some point? Just a simple lightbulb with the ability to turn it on and off in the config would be amazing :)

matklad:
No, I am not planning to do any actual work in this area :)

PR #5169, opened 2022-12-15
Better sorting in picker in case of ties

matklad:
Before:
old

After:
new

Wouldn't be opposed to writing a test here, but, given that this doesn't fail any existing test, also happy to let it pass :P

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -12,7 +12,10 @@ use tui::{
 use fuzzy_matcher::skim::SkimMatcherV2 as Matcher;
 use tui::widgets::Widget;
 
-use std::{cmp::Ordering, time::Instant};
+use std::{
+    cmp::{self, Ordering},
+    time::Instant,
+};
 use std::{collections::HashMap, io::Read, path::PathBuf};
 
 use crate::ui::{Prompt, PromptEvent};
@@ -344,9 +347,15 @@ impl<T: Item + 'static> Component for FilePicker<T> {
 
 #[derive(PartialEq, Eq, Debug)]
 struct PickerMatch {
-    index: usize,
     score: i64,
+    index: usize,
     len: usize,
+}
+
+impl PickerMatch {
+    fn key(&self) -> impl Ord {
+        (cmp::Reverse(self.score), self.len, self.index)
+    }
 }
 
 impl PartialOrd for PickerMatch {
@@ -357,10 +366,7 @@ impl PartialOrd for PickerMatch {
 
 impl Ord for PickerMatch {
     fn cmp(&self, other: &Self) -> Ordering {
-        self.score
-            .cmp(&other.score)
-            .reverse()
-            .then_with(|| self.len.cmp(&other.len))
+        self.key().cmp(&other.key())
     }
 }
 
@@ -502,6 +508,7 @@ impl<T: Item> Picker<T> {
                         })
                 }),
         );
+
         self.matches.sort_unstable();
     }
 
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -207,13 +207,14 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
 
     // Cap the number of files if we aren't in a git project, preventing
     // hangs when using the picker in your home directory
-    let files: Vec<_> = if root.join(".git").exists() {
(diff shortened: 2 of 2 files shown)


pascalkuthe:
Previous work for context: #4698

Thanks for taking the time to contribute! I like the general idea of sorting file picker entries alphabetically.
However there are many pickers where the initial sorting makes much more sense compared to alphabetical sorting. For example the references in the refence picker are usually sorted by their occurrence in the file (you know that better then me tough :D) `goto_reference`  where the server often returns items in the same order they appear in the sourcecode (or the definition first for goto_reference).
Furthermore this particular implementation is not that efficient as `filter_text` actually allocates twice internally (and would be called for each picker item, on each keypress). The overhead is not huge but it would be nice to avoid nonetheless as this particular action blocks the main UI thread.

I think the same result can be better achieved as follows:
Change the `Ord` implementation of `PickerItem` to also consider `index` in case of a tie (right now it compares by fuzzy score and then by char length).  Maybe a stable sort also works but I am not sure about that since a previous query could have changed the sorting due to a difference in fuzzy score. It might be alright depending on the particulars of how the fuzzy score works but I would rather not rely on that (and I believe an unstable sort by index is actually faster anyway).

When creating the picker we could then sort all items before passing them to `Picker::new`.  
This has the advantage of allowing a per-picker tie breaker that is simply based on how the original picker items are sorted. 

Furthermore this keeps the picker order a bit more consistent.
When a picker has not query it's items are simply displayed in original order.
When you then filter that picker with a query items would not be suffeled around randomly if they have the same matching score and instead keep their original order.

matklad:
makes makes sense, ptal!

helix-term/src/ui/picker.rs
```diff
@@ -344,11 +347,17 @@ impl<T: Item + 'static> Component for FilePicker<T> {
...
 struct PickerMatch {
-    index: usize,
     score: i64,
+    index: usize,
     len: usize,
 }
 
+impl PickerMatch {
+    fn key(&self) -> impl Ord {
+        (cmp::Reverse(self.score), self.index, self.len)
```
matklad, line 357: Interestingly, this also fixes the `Eq` impl -- previously the derived impl would disagree with the `Ord`

pascalkuthe, line 357:
```suggestion
        (cmp::Reverse(self.score), self.len, self.index)
```
I believe `len` and `index` must be swapped here.
Otherwise the length tiebreaker would have no effect (as the index is always unique)

matklad: :thinking: do we need a len at all then? as in, I'd expect

```
aaaaac
bc
```

order perhaps? 

Or maybe not -- I can see the appeal of starting with shorter entitiees... Let's go for a minimal diff for the start

matklad: Fixed (good thing that rust-analyzer has "flip comma" code action, hehe)

pascalkuthe: Sorting by length is useful because it usually shows you what you want first.
For example `foo/bar` matches `foo/bar.rs` before `foo/bar/baz.rs` this way.
See #4698 and  #4688 for context. `Fzf` shares the same behavior by default and I tested this a bit and it's quite useful in practice.

Note that we only apply this sorting once something is typed into the query.
I think the following example illustrates the intended behaviour.

Consider the initial list of files (sorted alphabetically):

```
aaaaac
ac
bc
```

Once you type `ac` the following is shown:

```
ac
aaaaac
```

pascalkuthe: I suppose that was a small oversight in the previous implementation. We probably should have returned `None` in case of equality to be consistent with `PartialEq` so this is definitly more consistent :+1:

matklad:
Should be good now! I briefly entertained thee idea of adding `Picker::new where T: Ord` and `Picker::new_presorted`, but think it's better to say that all call-sites are responsible for comming up with a reasonable order.

pascalkuthe (approved):
LGTM, I think this is pretty uncontroversial because it only acts as a tiebreaker (for what was previously an unstable sort).

the-mikedavis (approved):
Thanks!

TIL about "flip comma", that's handy 😄

matklad:
>TIL about "flip comma", that's handy smile

users not knowing what assists are available is a hard problem. I myself only remeber things I've added personally :D Some thoughts on how to address that: https://github.com/helix-editor/helix/issues/2011#issuecomment-1359830479

Issue #5136, opened 2022-12-12
Completion for paths that starts with `.`  off by one


sigmaSd:
### Summary

- touch .a
- hx
- `:o .`
- hit tab

expected: `.a` bu result is `..a`

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:

Instead, this happened:

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

linux

### Terminal Emulator

wezterm

### Helix Version

helix 22.12

DylanBulfin:
Another related issue that I'll add on here (something I addressed in the above PR), if you type in `:o ~/.` and press tab, instead of tabbing through the dotfiles in your home directory, it replaces the input with something completely different (in my case, it replaces the input with `:o {username}/` but I can't say for sure whether that is always the case). In general, typing a path that ends in `/.` or a path that is only a single period creates weird and incorrect behavior when displaying potential completions or tabbing through them.

sigmaSd:
seems to be fixed in master

PR #5175, opened 2022-12-15
Fix autocompletion for paths with period 

DylanBulfin:
Fix for https://github.com/helix-editor/helix/issues/5136

Bug has to do with how the `file_name` method deals with paths ending in `.` or `/.`. I hardcoded the value for `file_name` for paths ending in `/.` and paths that only consist of a single period, and updated the logic for `path`.

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -462,14 +462,13 @@ pub mod completers {
         use ignore::WalkBuilder;
         use std::path::Path;
 
-        let is_tilde = input.starts_with('~') && input.len() == 1;
+        let is_tilde = input == "~";
         let path = helix_core::path::expand_tilde(Path::new(input));
 
         let (dir, file_name) = if input.ends_with(std::path::MAIN_SEPARATOR) {
             (path, None)
         } else {
-            let is_period = (input.ends_with("/.") && input.len() > 2)
-                || (input.starts_with('.') && input.len() == 1);
+            let is_period = (input.ends_with("/.") && input.len() > 2) || input == ".";
             let file_name = if is_period {
                 Some(String::from("."))
             } else {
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/mod.rs
```diff
@@ -468,14 +468,23 @@ pub mod completers {
...
-            let file_name = path
-                .file_name()
-                .and_then(|file| file.to_str().map(|path| path.to_owned()));
-
-            let path = match path.parent() {
-                Some(path) if !path.as_os_str().is_empty() => path.to_path_buf(),
-                // Path::new("h")'s parent is Some("")...
-                _ => std::env::current_dir().expect("couldn't determine current directory"),
+            let is_period = (input.ends_with("/.") && input.len() > 2)
+                || (input.starts_with('.') && input.len() == 1);
```
the-mikedavis, line 472: The second check `input.starts_with('.') && input.len() == 1` could be `input == "."`. The `is_tilde` check on L465 could be refactored similarly

helix-term/src/ui/mod.rs
```diff
@@ -462,20 +462,28 @@ pub mod completers {
...
         } else {
-            let file_name = path
-                .file_name()
-                .and_then(|file| file.to_str().map(|path| path.to_owned()));
-
-            let path = match path.parent() {
-                Some(path) if !path.as_os_str().is_empty() => path.to_path_buf(),
-                // Path::new("h")'s parent is Some("")...
-                _ => std::env::current_dir().expect("couldn't determine current directory"),
+            let is_period = (input.ends_with("/.") && input.len() > 2) || input == ".";
```
dead10ck, line 471: Should this use `std::path::MAIN_SEPARATOR` instead of hardcoding `/`?

pascalkuthe: It should probably work for both seperators. There are many file formats that always / and convert to \ on windows (or don't you can use / aswell on windows these days AFAIK)

dead10ck (approved):
Ok this seems fine to me :+1:

PR #5176, opened 2022-12-15
Cross injection layers in tree-sitter motions

### LINKED ISSUE #2311: Using expand to parent syntax node inside an injected context selects the whole injection
### Summary

As described in the title and repro steps.

### Reproduction Steps

For example in the following minimal vue example:

```vue
<script type="javascript">
export default {
  computed: {
    isRendered () {
      return this.$store.state.isRendered
    }
}
</script>

<style>
.isRendered {
  display: block;
}
</style>
```

If you start with a single-character selection somewhere on the `return` line, and press `Alt-k` or `Alt-<up>`, instead of selecting the inner scope of the `isRendered` method as expected, the entire contents of the` <script>` are instead.

### Helix log

Log is not applicable, but `tree-sittier-scopes` does not report the scopes in the javascript, just `["component", "script_element", "raw_text"]` and the subtree shows only `(raw_text)`. This is in opposite to the fact that the highlight scopes must be applied to the actual scopes in the javascript injection, so these commands, and the expansion keybind are somehow not getting access to the injected scopes.

### Platform

Linux

### Terminal Emulator

wezterm 20220427-144728-efcc4b48

### Helix Version

helix 22.05-dev (477b88e9)

the-mikedavis:
Tree-sitter motions didn't work well for languages with injections such as Vue or JavaScript inside HTML since only the root layer's syntax-tree was considered. This change uses the range information for each injected language layer to find the appropriate syntax tree so that tree-sitter motions (A-p, A-i, A-o, A-n) work within injected content.

Here the tree-sitter motions work on the JavaScript syntax tree even though the root layer is HTML. Previously, A-o within the script tag would expand the selection to the whole script tag element.

Closes #2311

### DIFF
--- a/helix-core/src/object.rs
+++ b/helix-core/src/object.rs
@@ -1,42 +1,52 @@
-use crate::{Range, RopeSlice, Selection, Syntax};
-use tree_sitter::Node;
+use crate::{syntax::TreeCursor, Range, RopeSlice, Selection, Syntax};
 
 pub fn expand_selection(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
-    select_node_impl(syntax, text, selection, |mut node, from, to| {
-        while node.start_byte() == from && node.end_byte() == to {
-            node = node.parent()?;
+    let cursor = &mut syntax.walk();
+
+    selection.transform(|range| {
+        let from = text.char_to_byte(range.from());
+        let to = text.char_to_byte(range.to());
+
+        let byte_range = from..to;
+        cursor.reset_to_byte_range(from, to);
+
+        while cursor.node().byte_range() == byte_range {
+            if !cursor.goto_parent() {
+                break;
+            }
         }
-        Some(node)
+
+        let node = cursor.node();
+        let from = text.byte_to_char(node.start_byte());
+        let to = text.byte_to_char(node.end_byte());
+
+        Range::new(to, from).with_direction(range.direction())
     })
 }
 
 pub fn shrink_selection(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
-    select_node_impl(syntax, text, selection, |descendant, _from, _to| {
-        descendant.child(0).or(Some(descendant))
+    select_node_impl(syntax, text, selection, |cursor| {
+        cursor.goto_first_child();
     })
 }
 
-pub fn select_sibling<F>(
-    syntax: &Syntax,
-    text: RopeSlice,
-    selection: Selection,
-    sibling_fn: &F,
-) -> Selection
-where
-    F: Fn(Node) -> Option<Node>,
-{
-    select_node_impl(syntax, text, selection, |descendant, _from, _to| {
-        find_sibling_recursive(descendant, sibling_fn)
+pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
+    select_node_impl(syntax, text, selection, |cursor| {
+        while !cursor.goto_next_sibling() {
+            if !cursor.goto_parent() {
+                break;
+            }
+        }
     })
 }
 
-fn find_sibling_recursive<F>(node: Node, sibling_fn: F) -> Option<Node>
-where
-    F: Fn(Node) -> Option<Node>,
-{
-    sibling_fn(node).or_else(|| {
-        node.parent()
-            .and_then(|node| find_sibling_recursive(node, sibling_fn))
+pub fn select_prev_sibling(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
+    select_node_impl(syntax, text, selection, |cursor| {
+        while !cursor.goto_prev_sibling() {
+            if !cursor.goto_parent() {
+                break;
+            }
+        }
     })
 }
 
@@ -44,33 +54,25 @@ fn select_node_impl<F>(
     syntax: &Syntax,
     text: RopeSlice,
     selection: Selection,
-    select_fn: F,
+    motion: F,
 ) -> Selection
 where
-    F: Fn(Node, usize, usize) -> Option<Node>,
+    F: Fn(&mut TreeCursor),
 {
-    let tree = syntax.tree();
+    let cursor = &mut syntax.walk();
 
     selection.transform(|range| {
         let from = text.char_to_byte(range.from());
         let to = text.char_to_byte(range.to());
 
-        let node = match tree
-            .root_node()
-            .descendant_for_byte_range(from, to)
-            .and_then(|node| select_fn(node, from, to))
-        {
-            Some(node) => node,
-            None => return range,
-        };
+        cursor.reset_to_byte_range(from, to);
 
+        motion(cursor);
+
+        let node = cursor.node();
         let from = text.byte_to_char(node.start_byte());
         let to = text.byte_to_char(node.end_byte());
 
-        if range.head < range.anchor {
-            Range::new(to, from)
-        } else {
-            Range::new(from, to)
-        }
+        Range::new(from, to).with_direction(range.direction())
     })
 }
--- a/helix-core/src/syntax/tree_cursor.rs
+++ b/helix-core/src/syntax/tree_cursor.rs
@@ -1,0 +1,160 @@
+use std::{cmp::Reverse, ops::Range};
+
+use super::{LanguageLayer, LayerId};
+
+use slotmap::HopSlotMap;
+use tree_sitter::Node;
+
+/// The byte range of an injection layer.
+///
+/// Injection ranges may overlap, but all overlapping parts are subsets of their parent ranges.
+/// This allows us to sort the ranges ahead of time in order to efficiently find a range that
+/// contains a point with maximum depth.
+#[derive(Debug)]
+struct InjectionRange {
+    start: usize,
+    end: usize,
+    layer_id: LayerId,
+    depth: u32,
+}
+
+pub struct TreeCursor<'a> {
+    layers: &'a HopSlotMap<LayerId, LanguageLayer>,
+    root: LayerId,
+    current: LayerId,
+    injection_ranges: Vec<InjectionRange>,
+    // TODO: Ideally this would be a `tree_sitter::TreeCursor<'a>` but
+    // that returns very surprising results in testing.
+    cursor: Node<'a>,
+}
+
+impl<'a> TreeCursor<'a> {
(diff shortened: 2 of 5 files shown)


the-mikedavis:
I think this could be better if we introduce something that wraps the layers in `helix_core::syntax::Syntax` but has the same API as [`tree_sitter::TreeCursor`](https://docs.rs/tree-sitter/0.20.8/tree_sitter/struct.TreeCursor.html).

* Currently we use `tree_sitter::Node::parent` which can't cross injection layers, so `A-n`/`A-p` get 'stuck' at the edges of a layer
* We already use TreeCursor for `:tree-sitter-subtree`, so if we have the same API, we can trivially support printing the subtree for injected content.

pascalkuthe:
> I think this could be better if we introduce something that wraps the layers in `helix_core::syntax::Syntax` but has the same API as [`tree_sitter::TreeCursor`](https://docs.rs/tree-sitter/0.20.8/tree_sitter/struct.TreeCursor.html).
> 
>     * Currently we use `tree_sitter::Node::parent` which can't cross injection layers, so `A-n`/`A-p` get 'stuck' at the edges of a layer
> 
>     * We already use TreeCursor for `:tree-sitter-subtree`, so if we have the same API, we can trivially support printing the subtree for injected content.

Sounds like such an iterator might go in the same direction as what you did in the rainbow brackets PR?
I don't have much time this weekend so I didn't get a chance to look at that in more detail so I might just have missunderstood you there.

the-mikedavis:
They would be similar but the iterator for #2857 covers queries and this would cover walking around the tree without queries. With both together, tree-sitter features are very close to working seemlessly with injections 😀

pascalkuthe:
> They would be similar but the iterator for #2857 covers queries and this would cover walking around the tree without queries. With both together, tree-sitter features are very close to working seemlessly with injections grinning

That makes sense. I guess I was thinking that the query iterator could be a special case of the general tree iterator. It all seems pretty related to me. But it probably makes sense to keep the PRs separate. I am mostly just thinking out loud here so my ideas might not be possible as I dont't have time to look at code too closely right now :D

kirawi:
I just realized this is also an issue in Rust macros since we use injections for those.

divarvel:
@the-mikedavis what's the status on this PR? i'm interested in this feature and willing to pick it up if necessary/feasible

the-mikedavis:
I'm still working on this, I have some changes locally. The current approach works ok but it can have some odd edge cases on the injection boundaries. In order to make it work robustly we need to make a tree out of the injection layers and introduce a cursor type that works on that tree similar to `tree_sitter::TreeCursor`

themixednuts:
Hey just wanted to check-in on this PR's status. Love the editor, just getting harder to keep using on my projects based on Svelte.

the-mikedavis:
I got stuck for a while trying to use the tree cursor with `:tree-sitter-subtree`. Combined injections make that impossible since multiple nodes can point to the same child layer. In the future we could still print within injections with `:tree-sitter-subtree` by selecting the layer the selection range is contained in and printing only that layer. But for now I'm only updating `A-p`/`A-o`/`A-i`/`A-n`.

helix-core/src/syntax/tree_cursor.rs
```diff
@@ -0,0 +1,114 @@
...
+use tree_sitter::Node;
+
+pub struct TreeCursor<'a> {
+    layers: &'a HopSlotMap<LayerId, LanguageLayer>,
+    root: LayerId,
+    current: LayerId,
+    injection_ranges: HashMap<Range<usize>, LayerId>,
+    // TODO: Ideally this would be a `tree_sitter::TreeCursor<'a>` but
+    // that returns very surprising results in testing.
+    cursor: Node<'a>,
```
the-mikedavis, line 15: I planned to use `tree_sitter::TreeCursor` here but it gave obviously wrong results, for example `TreeCursor::goto_parent` returning `false` (indicating there is no parent) when `cursor.node().parent().is_some()` and `TreeCursor::goto_next_sibling` returning false when there were next siblings. I might not be understanding the API fully or it may have bugs. I don't see it as a blocker though since we currently use `Node::parent`/`Node::next_sibling`/`Node::prev_sibling` in `A-poin`.

helix-core/src/syntax/tree_cursor.rs
```diff
@@ -0,0 +1,114 @@
...
+                && layer.contains_byte_range(start, end)
+            {
+                container_id = layer_id;
+            }
+        }
+
+        self.current = container_id;
+        let root = self.layers[self.current].tree().root_node();
+        self.cursor = root.descendant_for_byte_range(start, end).unwrap_or(root);
+    }
```
the-mikedavis, line 113: This function is basically what we want for https://github.com/helix-editor/helix/issues/7364 (comment tokens and indent settings). This finds the right injection layer for the given byte range, which could be where the current cursor/selection is, and then on top of this we just need a way to get the language configuration for a given layer.

helix-core/src/object.rs
```diff
@@ -1,76 +1,74 @@
...
-fn find_sibling_recursive<F>(node: Node, sibling_fn: F) -> Option<Node>
-where
-    F: Fn(Node) -> Option<Node>,
-{
-    sibling_fn(node).or_else(|| {
-        node.parent()
-            .and_then(|node| find_sibling_recursive(node, sibling_fn))
+pub fn select_prev_sibling(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
+    select_node_impl(syntax, text, selection, |cursor| {
+        let _ = cursor.goto_prev_sibling() || cursor.goto_parent() && cursor.goto_prev_sibling();
```
pascalkuthe, line 45: ths may have been lost while rebasing over time but I still think you need to handle the loop/recusing the old implementation did (same for the next sibling function). You should also return None 

For example, the following while loop works (again you could fold the goto_parent into the while condition but I find loops with empty bodies weird)

```suggestion
        while !cursor.goto_prev_sibling() {
            if !cursor.goto_parent() {
                break
            }
        }
```

helix-core/src/object.rs
```diff
@@ -1,76 +1,74 @@
...
+
+        loop {
+            if cursor.node().byte_range() != byte_range {
+                break;
+            }
+            byte_range = cursor.node().byte_range();
+            if !cursor.goto_parent() {
+                break;
+            }
         }
```
pascalkuthe, line 21:
```suggestion
        while cursor.node().byte_range() == byte_range {
        	if !cursor.goto_parent() {
        		break
        	}
        }
```

this can be simplified.

`byte_range` doesn't need to be mutable since its never read after the loop and the loop arborts if the byte range changes.

Then you can use a while loop (you could fold the goto_parent into the while condition but I find loops with empty bodies weird)

helix-core/src/object.rs
```diff
@@ -1,76 +1,74 @@
...
+        let node = cursor.node();
         let from = text.byte_to_char(node.start_byte());
         let to = text.byte_to_char(node.end_byte());
 
-        if range.head < range.anchor {
-            Range::new(to, from)
-        } else {
-            Range::new(from, to)
-        }
+        Range::new(to, from).with_direction(range.direction())
```
pascalkuthe, line 72:
```suggestion
        Range::new(from, to).with_direction(range.direction())
```

nit: its confusing this is the "wrong" way around

pascalkuthe:
lets be ambitious and shhot for the next release with this one, I think this mostly looks good, I just found one perf issue and some smaller stuff

helix-core/src/object.rs
```diff
@@ -1,76 +1,74 @@
...
-    text: RopeSlice,
-    selection: Selection,
-    sibling_fn: &F,
-) -> Selection
-where
-    F: Fn(Node) -> Option<Node>,
-{
-    select_node_impl(syntax, text, selection, |descendant, _from, _to| {
-        find_sibling_recursive(descendant, sibling_fn)
+pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
```
pascalkuthe, line 33: This function needs the same treatment as the preview dinling function (using a while loop)

helix-core/src/syntax/tree_cursor.rs
```diff
@@ -0,0 +1,166 @@
...
+
+        true
+    }
+
+    /// Finds the injection layer that has exactly the same range as the given `range`.
+    fn layer_id_of_byte_range(&self, search_range: Range<usize>) -> Option<LayerId> {
+        let start_idx = self
+            .injection_ranges
+            .partition_point(|range| range.end < search_range.end);
+        if start_idx == self.injection_ranges.len() {
```
pascalkuthe, line 98: You don't need this special case it's handele automatically by the search below

helix-core/src/syntax/tree_cursor.rs
```diff
@@ -0,0 +1,166 @@
...
+        let start_idx = self
+            .injection_ranges
+            .partition_point(|range| range.end < end);
+        if start_idx == self.injection_ranges.len() {
+            return self.root;
+        }
+
+        self.injection_ranges[start_idx..]
+            .iter()
+            .take_while(|range| range.end >= end)
```
pascalkuthe, line 156:
```suggestion
            .take_while(|range| range.start <= end)
```

Since the list is sorted by range.end every range after the parition point fulfills this requirement. You need to check the start instead (otherwise you linear search the rest of the injections which is slow).

helix-core/src/syntax/tree_cursor.rs
```diff
@@ -0,0 +1,166 @@
...
+            false
+        }
+    }
+
+    /// Finds the injection layer that contains the given start-end range.
+    fn layer_id_containing_byte_range(&self, start: usize, end: usize) -> LayerId {
+        let start_idx = self
+            .injection_ranges
+            .partition_point(|range| range.end < end);
+        if start_idx == self.injection_ranges.len() {
```
pascalkuthe, line 150: Again you don't need this special case

pascalkuthe (approved):

cgahr:
Out of curiosity: does this also work for indent queries? Using the above example, if we have an html document and write js in a `<script>` tag, does this block use the injection queries defined for js?

the-mikedavis:
No this doesn't interact with queries. #9320 has some of the foundations for running queries across injection layers but it's limited to textobjects to keep the PR small. Moving indentation queries to use that functionality would be a follow-up change

archseer (approved):

PR #5189, opened 2022-12-17
Avoid trailing `s` in message when only 1 file is opened

theteachr:
Fixes a minor grammar error with "Loaded 1 files" message. Only adds the `s` character when the number of files opened is not 1.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -230,7 +230,7 @@ impl Application {
                 editor.set_status(format!(
                     "Loaded {} file{}.",
                     nr_of_files,
-                    "s".repeat((nr_of_files != 1).into()) // avoid "Loaded 1 files." grammo
+                    if nr_of_files == 1 { "" } else { "s" } // avoid "Loaded 1 files." grammo
                 ));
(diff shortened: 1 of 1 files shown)


helix-term/src/application.rs
```diff
@@ -227,7 +227,11 @@ impl Application {
                         doc.set_selection(view_id, pos);
                     }
                 }
-                editor.set_status(format!("Loaded {} files.", nr_of_files));
+                editor.set_status(format!(
+                    "Loaded {} file{}.",
+                    nr_of_files,
+                    "s".repeat((nr_of_files != 1).into()) // avoid "Loaded 1 files." grammo
```
the-mikedavis, line 233: I think this is clever but a little implicit. A regular `if` expression is more readable IMO: https://github.com/helix-editor/helix/blob/b12c65678aacc577b070c70307ef6fce528e4d85/helix-term/src/ui/statusline.rs#L302

the-mikedavis (approved):

theteachr:
You might want to remove the S-waiting-on-author label.

PR #5195, opened 2022-12-17
Print the binary required by the debug adapter

### LINKED ISSUE #5193: In '--health <lang>', print expected cmd when missing in $PATH
I tried looking through the issues but couldn't find any mention, but basically a small but useful enhancement would be to actually list what executable/cmd `helix` expects when it says `"Not found in $PATH"`.

So here:
```
$ hx --health cpp
Configured language server: clangd
Binary for language server: /usr/bin/clangd
Configured debug adapter: lldb-vscode
Binary for debug adapter: Not found in $PATH
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
```

It would be very neat if it instead said:
`Binary for debug adapter: 'lldb-vscode' not found in $PATH`

By instead doing the full `hx --health` one can figure it out, but I guess there is no reason it couldn't be mentioned in both.

hyderix:
This commit resolves #5193, where the author requested that the name of the binary needed is printed along with the rest of the health information.

This commit adds a `format!` macro which formats in the name of the binary and then it will be printed along with the rest of the debug information. The value in `cmd` is referenced to the call to which, and then consumed upon the call to `format!`.

This is my first issue in Helix, so please give feedback if this is not the kind of solution that you find optimal.

### DIFF
--- a/helix-term/src/health.rs
+++ b/helix-term/src/health.rs
@@ -281,9 +281,9 @@ fn probe_protocol(protocol_name: &str, server_cmd: Option<String>) -> std::io::R
     writeln!(stdout, "Configured {}: {}", protocol_name, cmd_name)?;
 
     if let Some(cmd) = server_cmd {
-        let path = match which::which(cmd) {
+        let path = match which::which(&cmd) {
             Ok(path) => path.display().to_string().green(),
-            Err(_) => "Not found in $PATH".to_string().red(),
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
We do already say the binary name in the `"Configured <protocol>: <cmd>"` line above each of these but I've seen some confusion over the "not found" message like in #5913 in some matrix messages and maybe some other issues / discussions. I think this change should clear that up 👍

PR #5199, opened 2022-12-18
Add locale language support for tutor

erasin:
- Added Chinese Simplified tutor file


runtime/tutors/zh_cn.txt
```diff
@@ -0,0 +1,1090 @@
...
+
+
+
+=================================================================
+=                          2.2 新建行                           =
+=================================================================
+
+ 键入 o 会在光标所在位置的下方新建一行，并进入插入模式。
+ 键入 u 可以进行撤销操作。
+ 键入 U 可以
```
Jiuh-star, line 189: Great translation, but remain some mistakes. In `2.2 新建行` (2.2 OPENING LINES) : 
```
 键入 u 可以进行撤销操作。
 键入 U 可以
```
should not be here.

erasin: sorry, I will check again.

Omnikar:
The `.txt` extension that used to be on the original tutor file was removed because the fact that the path is unset after it is opened was causing problems for people who had a language server set up for `.txt` files. I would as such suggest removing the `.txt` extension here too.

pascalkuthe:
I don't want to be resonsible for maintaing stuff in languages I can't read. This will quickly fall out of sync as the main file changes and maintainers and contributors won't be able to update. This can be maintained out of tree since the tutor is just  a simple.

I will close this as this is alerayd ouut of date with the tutor changes of the last year and I don't see us moving forward with this

Issue #3002, opened 2022-07-07
Allow remapping rootPath and rootUri sent to an LSP server to support running LSP inside containers


EpocSquadron:
I am attempting to get a dockerized LSP server running for a PHP project with Psalm. I am able to correctly set the language server path through the local dev container orchestrator (ddev), thanks to the addition of per-directory language configuration support. Here is my example workspace specific configuration:

```toml
[[language]]
name = "php"
## This invocation is correct and attaches
language-server = { command = "ddev", args = ["exec", "php", "vendor/bin/psalm-language-server"] }
```
Under the hood, this is using the `-i` switch to attach the stdout of the container to the caller (helix in this case). I am able to see the initialization succeed in the log, and error messages when attempting to use `gd` which is implemented by this LSP. The error is essentially that it is trying to access the file by the host's path, but can't find it because it runs in the container's file namespace.

I attempted setting the rootPath and rootUri in the `[language.config]` section but this does not affect the rootPath and rootUri sent in the initialization, and is ignored by this language server in the didChangeConfiguration event and in the intializationOptions passed. Thus, I need to be able to pass a map such as this, and have the lsp intialization logic in helix apply it to the rootpath and rootUri in the initialization call.

Something like this:

```toml
[[language]]
name = "php"
language-server = { command = "ddev", args = ["exec", "php", "vendor/bin/psalm-language-server"] }
path-map = "/home/epocsquadron/path/to/dir:/var/www/html"
```

Then in the rootPath and rootUri there would be a replacement of the first part with the second, delimited by the semicolon.

mindreader:
I ran into this as well. I don't even need a path map, as for every project of this nature, the root URI never changes, and I would merely set up something in the repository directory like `.helix/languages.toml`:
```
[[language]]
name = "elixir"
language-server = { command = "docker", args = ["exec", "-i", "dev_project", "/foo/bar/elixir-ls/release/language_server.sh"] }
root-uri = "/app/project"
```

And that would be sufficient.

mongrelion:
I'm not super familiar with the internals of LSP but I think that the server needs access to the files in the project in order to be able to do things like finding definitions and such? (I'm taking inspiration on how they do it at the [lspcontainers.nvim](https://github.com/lspcontainers/lspcontainers.nvim/blob/main/lua/lspcontainers/init.lua) plugin).

With that in mind, I have created a script that invokes the LSP in a container while mounting the working directory. It looks like this (removed most of the `case` block to simplify the example:

[...]

PR #5204, opened 2022-12-18
Path remapping for LSP

UseAfterFreee:
# Path remapping for LSP

(See issue #3002 for more context)

Sometimes it is useful to have the language server running in a docker container with all the required dependencies for the project, whereas the editor is running on the developers local machine. In this case helix can use the docker executable to open a tty with the language server running in the container. However since the volume mapping of the container likely causes path mismatches the communication between the editor and the language server is not done correctly.

A simple approach to fix this would be to scan the strings going over the transport layers and map the paths accordingly, however this might mistakenly modify certain pieces of the communication that are not specifying paths to be used by the editor or the language server. One example could be a string literal in the source code containing one of the path prefixes.

## Solution implemented in this PR

In this PR an alternative approach is implemented where a remap method is added to the lsp::Url struct. Then everywhere where the editor sends something to the language server a mapping is performed to adjust the paths. Similarly whenever the editor uses a path it obtained from the LSP (for example in the diagnostics and the code actions) it maps it back to the correct paths.

The mapping is made configurable in the languages.toml file (and thus can be done on a project by project basis with different path mappings for each project in the $PROJECT_DIR/.helix/languages.toml file)

an example configuration could be as follows:
```toml
[[language]]
name="rust"
language-server = { command = "docker", args = ["start", "-i", "rust-analyzer"], path-mapping = { from = "/home/user/dev/some-project", to = "/app" } }
```

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -201,6 +201,12 @@ impl<'de> Deserialize<'de> for FileType {
 }
 
 #[derive(Debug, Serialize, Deserialize)]
+pub struct PathMappingConfiguration {
+    pub from: String,
+    pub to: String,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case")]
 pub struct LanguageServerConfiguration {
     pub command: String,
@@ -212,6 +218,8 @@ pub struct LanguageServerConfiguration {
     #[serde(default = "default_timeout")]
     pub timeout: u64,
     pub language_id: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub path_mapping: Option<PathMappingConfiguration>,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -1,5 +1,6 @@
 use crate::{
     jsonrpc,
+    remap::LSPRemap,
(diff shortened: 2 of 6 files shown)


thomasaarholt:
Is there something blocking this PR? I think this would be a great feature, perhaps particularly if it could be combined with a way to connect to running containers dynamically (see my feature request in #7472).

roqvist:
Anyway to help this forward?

natearn:
This would be a great help. I use this workflow often, and getting LSP features working with it has been a persistent challenge.

rsmirnov90:
Indeed... this would be a gamechanger.

HernandoR:
quite splanded, we could off load LSP to docker service, which would be very robust and reproductable.

sandorex:
Is there a reason this is not merged? I think this feature would be great

shaleh:
Thank you @UseAfterFreee for the original work. The LSP code has shifted enough that it is not easily rebase updated.

I did a fresh exploration of this in my fork. https://github.com/shaleh/helix/tree/lsp-path-remapping-v2
It is fresh work, I have only used it for a bit today. But I can finally get a Python LSP setup working in a container which is how we do things at $DAY_JOB. It is slower. Which is not super surprising but maybe something that can be improved.

I chose to perform the mapping at a lower level to avoid needing to change the like 25 places in the LSP path that the paths are used. Based on previous conversations this is not likely to be merged but maybe you all will find it helpful or interesting. Also, open to comments on my implementation.

PR #5205, opened 2022-12-18
Use tree-sitter-ruby for crystal

the-mikedavis:
tree-sitter-crystal has some parse failures that make it very likely to have bad highlights on regular files (for example https://github.com/Blacksmoke16/oq/blob/master/src/oq.cr). It looks like there are some parse failures for strings in particular that lead to the file mostly being highlighted as a string.

Crystal and Ruby and close enough syntax-wise that we can probably get away with using tree-sitter-ruby until tree-sitter-crystal matures.


hovsater:
Some time has passed since this was merged and I believe https://github.com/crystal-lang-tools/tree-sitter-crystal can now be used with better results than relying on the Ruby grammar.

PR #5206, opened 2022-12-18
Select change range for goto_first/last_change commands

the-mikedavis:
This matches the behavior from #4713 but for the first and last change. The selection rules are the same as for goto_next/prev_change: additions and modifications select the added and modified range while deletions are represented with a point.


pascalkuthe (approved):
Thanks for catching that! I should have included that in the original PR after your comment but forgot about `goto_first_change` / `goto_last_change`. I actually hadn't noticed this yet, apparently I don't really use those motions :D

PR #5211, opened 2022-12-19
tutor: add chapter for commenting lines

uncomfyhalomacro:
Hello. Just a small addition to the tutor. :D

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -1055,10 +1055,58 @@ letters! that is not good grammar. you can fix this.
 
 
 =================================================================
-=                                                               =
+=                     11.1 COMMENTING A LINE                    =
 =================================================================
 
+Type Ctrl-c to comment the line under your cursor.
+To uncomment the line, press Ctrl-c again.
 
+1. Move your cursor to the line marked '-->' below.
+2. Now comment the line marked with '-->'.
+3. Now try uncommenting the line.
+
+--> Comment me please
+
+
+
+
+
+
+
+
+
+
+=================================================================
+=                 11.2 COMMENTING MULTIPLE LINES                =
+=================================================================
+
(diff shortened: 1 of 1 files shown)


runtime/tutor
```diff
@@ -96,7 +96,7 @@ _________________________________________________________________
  2. Move to a place in the line which is missing text and type
     i to enter Insert mode. Keys you type will now type text.
  3. Enter the missing text.
- 4. Type Escape to exit Insert mode and return to Normal mode.
+ 4. type Escape to exit Insert mode and return to Normal mode.
```
michaeljones, line 99: I'm just a random person but I'm not sure why this has been lowercased? Seems like all the points start with a capital?

uncomfyhalomacro: Oh I didn't notice this!

uncomfyhalomacro:
Okay. Done formatting and making the body the same number of lines as other sections. Thanks for the review! :heart:

gabydd:
I have been working on #4718 it only changes the ``Ctrl-c`` binding a bit see my most recent comment but I think once this is merged we should add something about how you can block comment (``<space>cb``) is what it's binded to in the pr right now

uncomfyhalomacro:
That's awesome. I guess we can just leave this PR open until that gets merged

the-mikedavis:
We can take this as-is: it looks like the current behavior of C-c in #4718 on plaintext files is unchanged. We can revise this section with future changes once they're merged 👍

the-mikedavis (approved):
This looks great, thanks!

gabydd:
> We can take this as-is: it looks like the current behavior of C-c in #4718 on plaintext files is unchanged. We can revise this section with future changes once they're merged 👍

That's what I was thinking as well I realize that probably wasn't clear in my comment

PR #5238, opened 2022-12-21
Fix & Tweak Rust's Syntax Highlighting

Chickenkeeper:
- Loop labels are now highlighted as `@label` instead of `@type`
- `unsafe` is now `@keyword.special`
- `macro_rules!` is now `@function.macro`
- `@type.variant` is replaced with `@type.enum.variant` so enum constructors are highlighted correctly
- Some newlines were removed for consistency

### DIFF
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -169,6 +169,9 @@
 
 (mutable_specifier) @keyword.storage.modifier.mut
 
+(reference_type "&" @keyword.storage.modifier.ref)
+(self_parameter "&" @keyword.storage.modifier.ref)
+
 [
   "static"
   "const"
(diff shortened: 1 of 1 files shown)


archseer (approved):

runtime/queries/rust/highlights.scm
```diff
@@ -154,30 +151,24 @@
...
-
 "fn" @keyword.function
+"unsafe" @keyword.special
+"macro_rules!" @function.macro
 
 (mutable_specifier) @keyword.storage.modifier.mut
 
-(reference_type "&" @keyword.storage.modifier.ref)
-(self_parameter "&" @keyword.storage.modifier.ref)
-
```
the-mikedavis, line 180: These were intentional: https://github.com/helix-editor/helix/pull/2731#discussion_r893982611

Chickenkeeper: Ah that's fair enough, I've put them back. On further investigation it looks like `@keyword.storage.modifier.ref` is specific enough to not clash with other storage modifier keywords if overridden in a theme anyway.

the-mikedavis (approved):

PR #5239, opened 2022-12-21
Reload language config with `:config-reload`

willful759:
Just had a go at implementing this

Not quite sure where to make test or what tests to make, but I tried it and it seems to be able to reload auto formatting, at least

https://user-images.githubusercontent.com/30967965/208830221-40562d71-cec4-4c2d-9f31-e6a056360fa4.mp4

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -393,12 +393,29 @@ impl Application {
         self.editor.refresh_config();
     }
 
+    /// refresh language config after config change
+    fn refresh_language(&mut self) {
+        match helix_core::config::user_syntax_loader() {
+            Ok(syntax_config) => {
+                self.syn_loader = std::sync::Arc::new(syntax::Loader::new(syntax_config));
+                for document in self.editor.documents.values_mut() {
+                    document.detect_language(self.syn_loader.clone());
+                }
+            }
+            Err(err) => {
+                let err_string = format!("failed to reload language config: {}", err);
+                self.editor.set_error(err_string);
+            }
+        }
+    }
+
     /// Refresh theme after config change
     fn refresh_theme(&mut self, config: &Config) {
         if let Some(theme) = config.theme.clone() {
             let true_color = self.true_color();
             match self.theme_loader.load(&theme) {
                 Ok(theme) => {
+                    self.refresh_language();
                     if true_color || theme.is_16_color() {
                         self.editor.set_theme(theme);
                     } else {
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -418,9 +418,7 @@ impl Application {
             if true_color || theme.is_16_color() {
                 self.editor.set_theme(theme);
             } else {
-                return Err(anyhow::anyhow!(
-                    "theme requires truecolor support, which is not available"
-                ));
+                anyhow::bail!("theme requires truecolor support, which is not available")
             }
         }
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -394,7 +394,7 @@ impl Application {
     }
 
     /// refresh language config after config change
-    fn refresh_language(&mut self) {
+    fn refresh_language_config(&mut self) {
         match helix_core::config::user_syntax_loader() {
             Ok(syntax_config) => {
                 self.syn_loader = std::sync::Arc::new(syntax::Loader::new(syntax_config));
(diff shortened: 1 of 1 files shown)


helix-term/src/application.rs
```diff
@@ -393,12 +393,29 @@ impl Application {
...
+        }
+    }
+
     /// Refresh theme after config change
     fn refresh_theme(&mut self, config: &Config) {
         if let Some(theme) = config.theme.clone() {
             let true_color = self.true_color();
             match self.theme_loader.load(&theme) {
                 Ok(theme) => {
+                    self.refresh_language();
```
the-mikedavis, line 418: This should be independent of reloading the theme: it can be called from `refresh_config` after `self.refresh_theme(&config)`

helix-term/src/application.rs
```diff
@@ -393,12 +393,29 @@ impl Application {
         self.editor.refresh_config();
     }
 
+    /// refresh language config after config change
+    fn refresh_language(&mut self) {
```
the-mikedavis, line 397: Minor nit but let's call this `refresh_language_config`

dead10ck:
If the language config changes the LSP, will this handle sending the new config? Or if the command changes, will it shut down the current instance and start up the new one?

the-mikedavis:
It looks like Document::detect_language calls Document::set_language under the hood which only changes syntax parts of the doc. `:set-language`/`:lang` has the same problem (https://github.com/helix-editor/helix/pull/3359#issuecomment-1209038787)

dead10ck:
Oh interesting. Should this be a separate issue/PR?

sudormrfbin:
> If the language config changes the LSP, will this handle sending the new config?

We can send the [`didChangeConfiguration`][1] request to the LSP to signal a change, but I'm not sure how many server's support dynamic config registration. I suppose the most reliable way is to restart the server if the config or the lsp command changes.

[1]: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeConfiguration

the-mikedavis:
Ah I didn't even think about changing config. For a first iteration, stopping the old server and starting a new one would be a good improvement. That can be done in a separate PR though - I'll make an issue 👍

helix-term/src/application.rs
```diff
@@ -417,6 +433,7 @@ impl Application {
     fn refresh_config(&mut self) {
```
the-mikedavis, line 433: I think the `refresh_config`, `refresh_language_config` and `refresh_theme` commands could use some refactoring to improve the error handling. With this change, if you have an error in your language config and a theme error, you only end up seeing the theme error since it overwrites the `editor.set_error` from `refresh_language_config`

`refresh_language_config` and `refresh_theme` can return `Result<(), Error>`s instead and then use the anyhow helpers like `anyhow::anyhow!` or `anyhow::bail!` to format the error messages and return early in cases of errors. Then `refresh_config` can set the editor error if there are any failures:

```rs
fn refresh_config(&mut self) {
    let mut refresh_config = || -> Result<(), Error> {
        let config = Config::load_default()
            .map_err(|err| anyhow::anyhow!("failed to load config: {}", err))?;
        self.refresh_language_config()?;
        self.refresh_theme(&config)?;

        // Store new config
        self.config.store(Arc::new(config));

        Ok(())
    };
    if let Err(err) = refresh_config() {
        self.editor.set_error(err.to_string());
    }
}
```

And the other functions can use `?` to return errors early and remove the duplicated error handling code:

```rs
fn refresh_language_config(&mut self) -> Result<(), Error> {
    let syntax_config = helix_core::config::user_syntax_loader()?;
    self.syn_loader = // ...
    // ...
    Ok(())
}

Comment (willful759) [reply helix-term/src/application.rs:430]: Liked the suggestion, code looks a lot cleaner now.

I'm new to rust so recommendations like this are always welcomed

**Review by dead10ck — APPROVED** [review]
Haven't tested locally, but the code looks good to me 👍

File: helix-term/src/application.rs [hunk 4]
```diff
@@ -393,38 +393,51 @@ impl Application {
...
+                .theme_loader
+                .load(&theme)
+                .map_err(|err| anyhow::anyhow!("Failed to load theme `{}`: {}", theme, err))?;
+
+            if true_color || theme.is_16_color() {
+                self.editor.set_theme(theme);
+            } else {
+                return Err(anyhow::anyhow!(
+                    "theme requires truecolor support, which is not available"
+                ));
```
the-mikedavis, line 423: Very minor thing but this can be written with `anyhow::bail!`:

```rs
anyhow::bail!("theme requires truecolor support, which is not available")
```

the-mikedavis (approved):

PR #5242, opened 2022-12-21
Fix old values shown in `select_register`

### LINKED ISSUE #5241: `select_register` incorrectly shows the oldest value in registers
### Summary

Most commands overwrite the entire register with a `Vec` of size `1`, so at any given point in time there's only 1 value in the register. However, some commands like `search` maintains history and actually pushes to the register instead. Currently, the `select_register` command (`"`) incorrectly takes the oldest value for these stacked registers instead of the latest one.

Discovered this bug while working on #5112 :)

### Reproduction Steps

Simply make two different searches, then press `"` to see the list of registers.

In the list, `"` would show the earlier search pattern instead of the latest one for register `/`.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

tmux 3.3a

### Helix Version

22.12-56-g6ab85917

xJonathanLEI:
Fixes #5241.

### DIFF
--- a/helix-view/src/info.rs
+++ b/helix-view/src/info.rs
@@ -74,7 +74,7 @@ impl Info {
             .map(|(ch, reg)| {
                 let content = reg
                     .read()
-                    .get(0)
+                    .last()
                     .and_then(|s| s.lines().next())
                     .unwrap_or_default();
                 (ch.to_string(), content)


the-mikedavis (commented):
This will fix the display for registers that are being used like a stack but this doesn't makes sense for other registers. Kakoune's help for `"` shows the docs for each register and special registers which is more helpful IMO.

If I remember correctly, the current behavior matches https://github.com/Delapouite/kakoune-registers/blob/master/registers.kak

xJonathanLEI:
Hmm I wasn't aware of any register with a `len()` larger than `1` (for this change to make a difference to them) but _not_ used like a stack at the same time. I made this change because it's showing wrong value for the search (`/`) register... it shows the oldest value instead of the latest.

How would you like the issue to be fixed if the proposed changes are not desired? Thanks in advance.

the-mikedavis:
You can yank multiple selections into a register.

The current display is not really wrong: it's what you will get when you paste that register (with `"/p`). It's just that search uses the last element instead of the first.

I think it could make sense to reverse the direction of the stack so that pushing operations push to the front and we always use the front element in search. This would diverge from Kakoune and we would probably want to change the datastructure used for storing values within the Register though.

xJonathanLEI:
> I think it could make sense to reverse the direction of the stack so that pushing operations push to the front and we always use the front element in search.

Ah I agree this makes the most sense.

> change the datastructure used for storing values within the Register

Yeah inserting at the beginning of `Vec` would be extremely inefficient.

ohno418:
Hi. I'm a bit interested in this, so took a look into how registers work.

@the-mikedavis 
> The current display is not really wrong: it's what you will get when you paste that register (with "/p). It's just that search uses the last element instead of the first.

Given that the `/` register is supposed to have a *last* search value (https://docs.helix-editor.com/master/usage.html#special-registers), it feels weird that `"/p` pastes the *first* search value. (Not only the case for the `/` register, as well as the `:` register.)

So my question is why does the `/` register have multiple values like logs? If I understand correctly, in the case of the yank register, we just rewrite to the new value, instead of pushing it ([code here](https://github.com/helix-editor/helix/blob/bdcd4d9411655ab69245d803e88f88cc278127da/helix-term/src/commands.rs#L3586-L3588)). Why the `/` register doesn't do the same as that?

(I came from NeoVim and not Kakoune. Sorry if I incorrectly understand registers or selections.)

the-mikedavis:
I haven't looked through the history but I think that note in the docs is just out of date. The search commands push new searches to the end of the `/` register like a stack. The prior values in the stack are used for history completion in the prompt UI element.

The yank commands use registers differently: each range in a selection has its contents saved as an element in the register.

ohno418:
> The prior values in the stack are used for history completion in the prompt UI element.

Ah, got it.

> The search commands push new searches to the end of the / register like a stack.

> The yank commands use registers differently: each range in a selection has its contents saved as an element in the register.

Yeah. Hmm, so I'm not really sure, but it looks that the problem here is the inconsistency. Some registers like the search register have their values as history stacks, and others have them simply to refer to multiple selections.

So, changing the data structures. That makes sense. Thank you!

PR #5244, opened 2022-12-21
Make editor remember the latest search register

### LINKED ISSUE #5112: Make search commands respect the register prefix?
Not sure if this should be submitted as a bug or enhancement, so enhancement it is.

Currently serach-related commands do not respect the register prefix and are hard-coded to use `/`:

- `search`
- `rsearch`
- `search_next`
- `search_prev`
- `extend_search_next`
- `extend_search_prev`
- `search_selection`
- `make_search_word_bounded`
- `global_search`

While `search`, `rsearch`, and `global_search` use a prompt, the register `/` is still used for completion.

I wonder if this is intentional, given how consistent it is that all search commands discard the register input. Otherwise I'd love to send a PR.

xJonathanLEI:
Resolves #5112.

This PR makes changes to the behavior of search commands:

- The editor now remembers the _search register_ used to kick start a `search`/`rsearch`/`global_search` command;
- `search_next`, `search_prev`, `extend_search_next`, and `extend_search_prev` now uses the _search register_ to continue the search;
- `search_selection` now uses the selected register. It also updates the _search register_;
- `make_search_word_bounded` defaults to using the _search register_ to make the `search_selection`-`make_search_word_bounded` flow more ergonomic, while also accepting register selection.

Overall, this PR should make the whole family of search commands play more nicely with register selection.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1895,7 +1895,9 @@ fn searcher(cx: &mut Context, direction: Direction) {
                 .collect()
         },
         move |editor, regex, event| {
-            if !matches!(event, PromptEvent::Update | PromptEvent::Validate) {
+            if event == PromptEvent::Validate {
+                editor.registers.last_search_register = reg;
+            } else if event != PromptEvent::Update {
                 return;
             }
             search_impl(
@@ -1914,7 +1916,9 @@ fn searcher(cx: &mut Context, direction: Direction) {
 
 fn search_next_or_prev_impl(cx: &mut Context, movement: Movement, direction: Direction) {
     let count = cx.count();
-    let register = cx.register.unwrap_or('/');
+    let register = cx
+        .register
+        .unwrap_or(cx.editor.registers.last_search_register);
     let config = cx.editor.config();
     let scrolloff = config.scrolloff;
     if let Some(query) = cx.editor.registers.first(register, cx.editor) {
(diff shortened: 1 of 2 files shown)


xJonathanLEI:
Update:

Recent changes in the editor fixed the part where commands do not respect immediate register selection (i.e. `"bn` now uses register `b` as expected). However, the UX is still bad as this requires the user to enter the `"b` prefix every time to navigate through the changes.

That's why this PR is still relevant IMO, as the 2nd part of the PR is to make the editor remember the latest register used to kick off a search, and uses that by default unless a prefix is explicitly specified. In this case users can drop the prefix and simply press `n` to navigate.

Also update the PR title to better reflect the only part left for this PR.

pascalkuthe (approved):

pascalkuthe:
yeah I think this is a nice QOL improvement and makes sense. Implementation is preggy small/unintrusive too :+1:

xJonathanLEI:
> this change makes any register that you select the default until you explicitly select another one

Not really though? If you start a new search with just `/` (no explicit prefix), it changes back to the `/` register. See [this line](https://github.com/xJonathanLEI/helix/blob/70257ae140f6c8d62e53b891f403d971b8d66aed/helix-term/src/commands.rs#L2071-L2077), which is what starts a search. It doesn't even read from `search_register`.

I too agree that if it were _that_ sticky it's a problem, but that's not the case here. I also just tested my branch and it's indeed working as expected (i.e. reverts back to `/` when performing  a new search without prefix).

the-mikedavis:
Ah yep I see now. Let's rename it to `last_search_register` though, the current name is confusing to me.

xJonathanLEI:
Renamed the field the `last_search_register` as suggested.

pascalkuthe (commented):
Diff is small enough to review quickly so no worries 
:) Maybe one more thing: we could move `last_search_register` to `editir.registers`. This avoids adding more field to `Editor`

xJonathanLEI:
Sure thing! But do you mean making this a `pub` field of `Registers`? Looks like that struct does not have any public field yet. Should I make it the first? Or should I keep it private but expose a method with the same name?

pascalkuthe (approved):
A public field is probably fine in this case. I don't think it makes sense to encapsulate here. For me its just about bundling related fields together instead of making `Editor`  the kitchen sink for everything

the-mikedavis (approved):

PR #5247, opened 2022-12-21
Add eb word selection trick to the tutor

matklad:
Maybe this is obvious, but it took me several days to google this up :-)

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -241,7 +241,7 @@ _________________________________________________________________
 
 
 =================================================================
-=                     3.2 MORE ON MOTIONS                       =
+=                     3.2 MORE MOTIONS                          =
 =================================================================
 
  As you saw, typing w moves the cursor forward until the start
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
I like having an example for `w`/`b`/`e` for selecting words 👍

Finding a place for it will be a little tricky though: each tutor section has a constant height of 19 rows and this currently makes ch 3.2 too tall

Omnikar:
We could give the extra stuff the space of one full other page, so that the uniformity is maintained.

matklad:
What about this version? This also gives space to add an example for WORDS, which I think might be useful for people not already familiar with vim.

the-mikedavis (approved):
This looks great, thank you!

matklad:
Filed https://github.com/helix-editor/helix/issues/5284 which sounds like a good `E-good-first-issue`.

PR #5260, opened 2022-12-22
Support going to specific positions in file

### LINKED ISSUE #5259: Support file path with lines and columns in `goto_file`
Currently the `goto_file` command takes the entire WORD (or selection(s)) as the the file path. However, a common use case for this command would be to locate compiler/linter errors/warnings, which usually include line & column information.

For example, clippy reports file paths as `PATH:LINE:COL`. As of now Helix doesn't even support opening the file with path like this, leave alone going to the specific location as well.

This will be particularly useful when integrated terminal (#1976) is implemented.

xJonathanLEI:
Resolves #5259.

Now you can pipe clippy output into a file, open the file in Helix, and easily navigate to the exact locations of the warnings/errors.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1197,6 +1197,11 @@ fn goto_file_vsplit(cx: &mut Context) {
 
 /// Goto files in selection.
 fn goto_file_impl(cx: &mut Context, action: Action) {
+    struct OpenOption {
+        path: PathBuf,
+        location: Option<(usize, usize)>,
+    }
+
     let (view, doc) = current_ref!(cx.editor);
     let text = doc.text();
     let selections = doc.selection(view.id);
@@ -1258,19 +1263,88 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
             .collect()
     };
 
+    // Most compilers/linters print in this format
+    static REGEX_FILE_ROW_COL: Lazy<Regex> =
+        Lazy::new(|| Regex::new("^(?P<path>.[^:]+):(?P<row>\\d+)(:(?P<col>\\d+))?$").unwrap());
+
     for sel in paths {
         if let Ok(url) = Url::parse(&sel) {
             open_url(cx, url, action);
             continue;
         }
 
+        let p = sel.trim().to_owned();
+
         let path = expand_tilde(Cow::from(PathBuf::from(sel)));
         let path = &rel_path.join(path);
         if path.is_dir() {
             let picker = ui::file_picker(path.into(), &cx.editor.config());
             cx.push_layer(Box::new(overlaid(picker)));
-        } else if let Err(e) = cx.editor.open(path, action) {
-            cx.editor.set_error(format!("Open file failed: {:?}", e));
+        } else {
+            let open_option = match REGEX_FILE_ROW_COL.captures(&p) {
+                Some(file_row_col) => {
+                    let path = file_row_col.name("path").unwrap().as_str();
+                    let loc = match file_row_col
+                        .name("row")
+                        .unwrap()
+                        .as_str()
+                        .parse::<NonZeroUsize>()
+                    {
+                        Ok(row) => match file_row_col.name("col") {
+                            Some(col) => match col.as_str().parse::<NonZeroUsize>() {
+                                Ok(col) => Some((row.get(), col.get())),
+                                Err(_) => None,
+                            },
+                            None => Some((row.get(), 1)),
+                        },
+                        Err(_) => None,
+                    };
+
+                    OpenOption {
+                        path: PathBuf::from(path),
+                        location: loc,
+                    }
+                }
+                None => OpenOption {
+                    path: PathBuf::from(p),
+                    location: None,
+                },
+            };
+
+            match cx.editor.open(&open_option.path, action) {
+                Ok(_) => {
+                    if let Some((row, col)) = open_option.location {
+                        let (view, doc) = current!(cx.editor);
+
+                        let doc_text = doc.text();
+
+                        // Number of lines is always positive even for empty buffers
+                        let doc_lines = doc_text.len_lines();
+
+                        // Zero-based line index
+                        let ind_adjusted_line = usize::min(row, doc_lines) - 1;
+
+                        let ind_dest = if row > doc_lines {
+                            // Discard designated col and simply set to end of doc
+                            doc_text.len_chars().saturating_sub(1)
+                        } else {
+                            let line_len = doc_text.line(ind_adjusted_line).len_chars();
+
+                            let adjusted_ind_col = if line_len == 0 {
+                                0
+                            } else {
+                                usize::min(col, line_len) - 1
+                            };
+
+                            doc_text.line_to_char(ind_adjusted_line) + adjusted_ind_col
+                        };
+
+                        doc.set_selection(view.id, Selection::point(ind_dest));
+                        align_view(doc, view, Align::Center);
+                    }
+                }
+                Err(e) => cx.editor.set_error(format!("Open file failed: {:?}", e)),
+            }
         }
     }
 }


helix-term/src/commands.rs
```diff
@@ -1057,11 +1062,81 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
                 .to_string(),
         );
     }
+
+    // Clippy output format.
+    let regex_file_row_col = Regex::new("(?P<path>.[^:]+):(?P<row>\\d+):(?P<col>\\d+)").unwrap();
```
kirawi, line 1067: This is too specific to Rust, and so we probably want to use a more generic way of detecting this information. Ideally, I'd imagine that you could do something like `cargo clippy | regex_query | hx`.

xJonathanLEI: I think ideally we should support all common formats so that Helix always works out of the box.

In fact, I'd argue `PATH:ROW:COL` and `PATH:ROW` are not very Rust specific. A lot of compilers print paths in this way. For example, webpack prints in this exact format as well.

Also, one more argument for not relying on terminal piping has been specified in the linked issue (#5259): this feature would be very useful once integrated terminal is implemented, where you can't do any text transformation at all, so this workaround doesn't apply anymore.

kirawi: It can also be turned into a lazy static.

gabydd: What about ``PATH COL:ROW``? This vscode issue gives some examples:  https://github.com/microsoft/vscode/issues/140780

xJonathanLEI: > It can also be turned into a lazy static.

Fixed. Thanks for pointing that out. I wasn't aware Helix already had lazy static set up.

xJonathanLEI: > What about `PATH COL:ROW`?

What do the maintainers think about this? @kirawi

To support this we simply have to allow the first `:` to be one of `:` and ` `, a trivial change in the regex.

xJonathanLEI:
The reason I coded my own char index calculation is that the LSP impl is not forgiving enough: I want to allow "clamping" the designated position into the actual doc range in a predictable and reasonable way, whereas such a mismatch in the context of LSP would be considered a bug.

Actually, maybe I should refactor this into an option when calling `editor.open`? This would make adding a CLI option for opening a file at a specific position easier. Such a functionality would be useful with tools like `zellij`, which relies on a text editor to browser through the buffer. Right now if you use Helix for that, there's another `gegl` to press just to place the cursor at the end of the buffer as expected.

helix-term/src/commands.rs
pascalkuthe: https://github.com/microsoft/vscode/blob/1f00f5a6a62b5c860333be498b9f314b93f8afcb/src/vs/workbench/contrib/terminal/browser/links/terminalLocalLinkDetector.ts#L57

You can find a list of all formats supported by vscode (including the corresponding regex) above.

xJonathanLEI: Ah this is helpful. Lemme change to use this list of regex instead. Thanks for the pointer!

xJonathanLEI: The vscode detector is so comprehensive that I think we can drop the existing trimming as well. I will update this PR by porting the whole thing over :))

I think with this powerful detector we can even change the `goto_file` feature from the current:

"take this selection (or WORD) as file path as is except some funky chars to be trimmed"

to

- if it's a single-char selection (aka. no selection), "take the whole line and try to find a file link which contains the current cursor position and open it";
- otherwise, "try to detect a file link _within_ the selection and open it".

What do you guys think?

filipdutescu: That would be such a cool feature, personally I'd love to have your suggestion as described

pascalkuthe: I don'y think you need to consider the second case. Seems pretty unintuitive that the goto command wouldn't work if you have selected something within the link. Other commands like goto_definition only take the cursor into account so.it seems more consistent to just use the first approach you mentioned

xJonathanLEI: @pascalkuthe You're right. I proposed that because the current implementation distinguishing between the two cases. But now that I took a step back and only consider UX, doing that makes little sense. The special handling currently in place is likely to make up for the cases where auto-trimming etc. fails, which shouldn't be needed with the vscode detector port.

xJonathanLEI: I started looking into the vscode impl. The regexs make heave use of backreferences which are not supported in the `regex` crate. We likely need `fancy-regex` here.

I looked into the current deps tree and `fancy-regex` is not used. Porting the detector would therefore involve introducing new dependencies. Maybe @archseer can kindly weigh in on this? Thanks!

archseer: It looks like a `splitn(3) 
` to me, we already have this parsing implemented somewhere in app init since we parse the file argument this way

archseer: https://github.com/helix-editor/helix/blob/df1830ef28a7cb49abe31a18e4bd1bcfc7eb802a/helix-term/src/args.rs#L112-L121

xJonathanLEI: @archseer Ah I was actually referring to the file path detector used in vscode pointed out by @pascalkuthe:

https://github.com/microsoft/vscode/blob/1f00f5a6a62b5c860333be498b9f314b93f8afcb/src/vs/workbench/contrib/terminal/browser/links/terminalLocalLinkDetector.ts#L57

The simple case of `PATH:ROW:COL` already works pretty well in this PR with no additional dependency needed. The`fancy-regex` crate is needed to port the whole vscode detector over (which allows us to have to same file linking capability as vscode, which works pretty awesome imo).

xJonathanLEI: Sorry. I studied the vscode detector regexs deeper and it turns out that those are not meant to be backreferences. `\0` there actually intends to mean the unicode char `\u0000` if I'm not mistaken. In that case, we actually don't need `fancy-regex`. The existing `regex` crate should work just fine.

I'm kinda confused tho. In the JavaScript regex engine, `\0` should actually mean backtraces. I just checked in node and it seems to actually be the case:

```console
$ node
> new RegExp("\\0").exec("a")
null
> new RegExp("\\0").exec("\u0000")
[ '\x00', index: 0, input: '\x00', groups: undefined ]
```

So maybe, just maybe, the vscode detector has a bug here. They incorrectly used `\0` to mean the null terminator whereas `\0` actually means a backreference, and this bug was left undiscovered due to the rare occurrence of `\u0000`.

pascalkuthe: Yeah I don't think you need to backtrack for the so.ole cases. I wanted to take a look at the regexes to confirm but didn't get the chance. I don't think we want to lull in fancy regex

xJonathanLEI: @pascalkuthe Yeah I was also wondering why they even needed backreferences at all. I guess I just trusted regex101 too much :/

the-mikedavis:
> This would make adding a CLI option for opening a file at a specific position easier

You can provide line/cols already with CLI args: `hx README.md:19:6`. Does Zellij use a different format for this?

Also see https://github.com/helix-editor/helix/blob/8347139ff582341246975f047f04b4848f6e5af9/helix-term/src/args.rs#L90-L98 as archseer linked: can't we re-use or extend this function and use it for `goto_file_impl`?

pascalkuthe:
> > This would make adding a CLI option for opening a file at a specific position easier
> 
> You can provide line/cols already with CLI args: `hx README.md:19:6`. Does Zellij use a different format for this?
> 
> Also see
> 
> https://github.com/helix-editor/helix/blob/8347139ff582341246975f047f04b4848f6e5af9/helix-term/src/args.rs#L90-L98
> as archseer linked: can't we re-use or extend this function and use it for `goto_file_impl`?

I also ran into that function and taught the same thing. 
However as discussed above there are multiple different formats to specify file columns/rows.
The formats VSCode supports can be found here:

[microsoft/vscode@1f00f5a/src/vs/workbench/contrib/terminal/browser/links/terminalLocalLinkDetector.ts#L57](https://github.com/microsoft/vscode/blob/1f00f5a6a62b5c860333be498b9f314b93f8afcb/src/vs/workbench/contrib/terminal/browser/links/terminalLocalLinkDetector.ts#L57)

I was under the impression that @xJonathanLEI implemented support for those captures but looking at the PR it does look like currently only `<FILE>:<Line>:<COL>` is supported.

Regardless the current implementation can not just use `parse_file` because this command used to use `textobect_word` to select text (which could be passed to `parse_file`) but `textobject_word` does not select colons so just passing the selection to `parse_file` would not actually work.

I think the best solution here would be to have a regex set of all possible file regexes (taken from vscode) and select the first regex capture that contains the current cursor. We could factor that regex out into a function tough and reuse it in `parse_file` if we want to support those extra path styles in the CLI aswell.

On a sidenote: Currently `goto_file` supports opening the current selection instead of using the word motions. I think this was implemented because the current word doesn't allow colons or slash. That requirement would be obsolete if a regex is used. Other goto commands (like goto_definition) only respect the cursor and ignore the selection so it would seem more consistent to do the same here (see https://github.com/helix-editor/helix/pull/5260#discussion_r1055769702, https://github.com/helix-editor/helix/pull/5260#discussion_r1055967835).

the-mikedavis:
> I think the best solution here would be to have a regex set of all possible file regexes (taken from vscode) and select the first regex capture that contains the current cursor

This sounds like a good way to improve the smart detection in this block: https://github.com/helix-editor/helix/blob/2b58ff4d7cc09bc48bcdd79096110eeb578b509f/helix-term/src/commands.rs#L1037-L1058

But the automatic detection of whether surrounding text is a file/row/column should be separate from the parsing of the contents of the selection: when there is a selection (i.e. >1 width) goto-file should just parse the contents of the selection without trying to be smart about the surrounding text.

> On a sidenote: Currently goto_file supports opening the current selection instead of using the word motions. I think this was implemented because the current word doesn't allow colons or slash.

I believe it's the other way around: the automatic detection was implemented on top of the parsing of the current selections. Kakoune only parses the contents of the selections since that fits the select-then-act paradigm (also see the original discussion in https://github.com/helix-editor/helix/pull/1102#discussion_r749188356). The smart detection is a sort of separate feature that should only kick in if there's a single 1-width cursor.

pascalkuthe:
> > I think the best solution here would be to have a regex set of all possible file regexes (taken from vscode) and select the first regex capture that contains the current cursor
> 
> This sounds like a good way to improve the smart detection in this block:
> 
> https://github.com/helix-editor/helix/blob/2b58ff4d7cc09bc48bcdd79096110eeb578b509f/helix-term/src/commands.rs#L1037-L1058
> 
> But the automatic detection of whether surrounding text is a file/row/column should be separate from the parsing of the contents of the selection: when there is a selection (i.e. >1 width) goto-file should just parse the contents of the selection without trying to be smart about the surrounding text.
> 
> > On a sidenote: Currently goto_file supports opening the current selection instead of using the word motions. I think this was implemented because the current word doesn't allow colons or slash.
> 
> I believe it's the other way around: the automatic detection was implemented on top of the parsing of the current selections. Kakoune only parses the contents of the selections since that fits the select-then-act paradigm (also see the original discussion in [#1102 (comment)](https://github.com/helix-editor/helix/pull/1102#discussion_r749188356)). The smart detection is a sort of separate feature that should only kick in if there's a single 1-width cursor.

Thanks for the additional context.

I was sort of coming from the standpoint of looking at `goto_defintion` , `goto_reference`, ... which all only use the cursor so it always seemed kind of odd to me that `goto_file` behaves differently and just assumed that was the reason. I imagine it would be a bit odd if I have part of the text select by accident and `gf` stopped working. I guess there are cases tough where you might want to select a postfix of a file (or exclude part of the selection with `K`) to produce a new path. I personally never taught about using `gf` like that but if that behavior was intentional then it makes sense to keep that.

In that case it probably works best to separate as a a `OnceCell` static so it can be reused for both the `parse_file` and the smart detection case.

xJonathanLEI:
> I was under the impression that @xJonathanLEI implemented support for those captures but looking at the PR it does look like currently only <FILE>:<Line>:<COL> is supported.

Yep I was actually going to suggest replacing the "waiting-on-review" tag with "waiting-on-author" since I haven't implemented the vscode detector.

Regarding the behavior when a selection is used, here's my 2 cents after thinking about it more. The way I think about it is: do we lose support for any use case if we implement it one way over another? Since you can always collapse the selection into a single-char one with `;` I think we should go for the "ignore surrounding text on non-single-char selections" approach as @the-mikedavis suggested.

Here's one missing use case if `goto_file` is overly "smart" and always reach for the surrounding text.

Imagine you generated the console logs containing the relative file path from the parent folder of CWD. Say you're in `/a/b/c`, but the command was run on `/a/b`, so it shows a path to your file as `./c/d/your_file`. You know it's from a different folder, so you *select* the `d/your_file` part and do `gf`.

Oops, the editor is too _smart_, it finds out the "complete" path is `./c/d/your_file` and tries to open that instead. That doesn't work unfortunately.

This is something vscode doesn't have to even think about since it's mouse-oriented.

xJonathanLEI:
Update: I shifted focus to Git related stuff as I personally don't need the full-fledged detector and am happy with the current implementation here. Anyone interested in taking this PR across the finish line please feel free to do so :)

mmlb:
vim-fetch also has some patterns that I think aren't in vscode's matcher: https://github.com/wsdjeg/vim-fetch/blob/master/autoload/fetch.vim#L12-L75

k4lizen:
is there anything left to do here other than resolve conflicts?

Issue #5174, opened 2022-12-15
Timeout waiting for LSP after quitting and garbage output for all toml files since 22.12


illode:
### Summary

Since the update, any toml file will hang after quitting, then output `Error: Timed out waiting for language servers to shutdown`. Doesn't seem to happen for other formats.

Also, sometimes after quitting the next shell prompt will be pre-filled with a bunch of escape sequence garbage like: `35;93;71M35;109;71M35;119;71M35;129;71M35;139;71M35;148;72M35;155;73M`. The exact garbage changes. I can't remember if this bug was introduced in 22.12 or not.

Both tested on two machines with both zsh (customized) and bash (distro default) and two terminal emulators.

### Reproduction Steps

1. Open toml file. Neither the contents matter, nor whether it exists or not. `helix qwe.toml`
2. Quit with `:q` or `:q!`. No changes need to be made.

UI will freeze for a second or two, then print the error. Trying this repeatedly will eventually lead to the garbage output, but it can take a lot of attempts (>50). 

### Helix log

`helix -vvv qwe.toml` log
This is with both garbage in prompt and hang.

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-12-15T11:20:44.661 helix_loader [DEBUG] Located configuration folders: []
2022-12-15T11:20:44.669 helix_view::clipboard [INFO] Using wl-copy+wl-paste to interact with the system and selection (primary) clipboard
2022-12-15T11:20:44.671 mio::poll [TRACE] registering event source with poller: token=Token(0), interests=READABLE | WRITABLE
2022-12-15T11:20:44.671 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2022-12-15T11:20:44.671 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2022-12-15T11:20:44.671 helix_view::editor [DEBUG] editor status: Loaded 1 files.
2022-12-15T11:20:44.671 mio::poll [TRACE] registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
2022-12-15T11:20:44.671 mio::poll [TRACE] registering event source with poller: token=Token(4), interests=READABLE | WRITABLE
[...]

PR #5266, opened 2022-12-23
Configurable lsp shutdown timeout

jackos:
## What
Makes the lsp shutdown timeout configurable but retains the default of `3s`,

## Why
I use a proxy server: https://github.com/pr2502/ra-multiplex so I can have multiple instances of helix open via a multiplexer using the same instance of `rust-analyzer`, every time I want to close a window gracefully it takes 3s and is a little painful, this allows me to set it to 0 via the option:
```
[editor.lsp]
shutdown-timeout = 0
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ target
 helix-term/rustfmt.toml
 result
 runtime/grammars
+.DS_Store
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -242,6 +242,13 @@ pub struct LspConfig {
     pub auto_signature_help: bool,
     /// Display docs under signature help popup
     pub display_signature_help_docs: bool,
+    /// Time in milliseconds that the editor will wait for the lsp to shutdown before exiting
+    /// Set to 0 for instant. Defaults to 3000.
+    #[serde(
+        serialize_with = "serialize_duration_millis",
+        deserialize_with = "deserialize_duration_millis"
+    )]
+    pub shutdown_timeout: Duration,
 }
 
 impl Default for LspConfig {
@@ -250,6 +257,7 @@ impl Default for LspConfig {
             display_messages: false,
             auto_signature_help: true,
             display_signature_help_docs: true,
+            shutdown_timeout: Duration::from_millis(3000),
         }
     }
 }
@@ -1356,14 +1364,13 @@ impl Editor {
         }
     }
 
-    /// Closes language servers with timeout. The default timeout is 10000 ms, use
-    /// `timeout` parameter to override this.
+    /// Closes language servers with timeout, set to 0 to instantly shutdown without waiting for lsp
     pub async fn close_language_servers(
         &self,
-        timeout: Option<u64>,
+        timeout: Duration,
     ) -> Result<(), tokio::time::error::Elapsed> {
         tokio::time::timeout(
-            Duration::from_millis(timeout.unwrap_or(3000)),
+            timeout,
             future::join_all(
                 self.language_servers
                     .iter_clients()
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -117,6 +117,7 @@ The following statusline elements can be configured:
 | `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
 | `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
 | `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
+| `shutdown-timeout`    | Time in milliseconds to wait for lsp to shutdown before closing editor, set to 0 to close instantly. | `3000`  |
 
 [^1]: By default, a progress spinner is shown in the statusline beside the file path.
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1087,7 +1087,8 @@ impl Application {
             errs.push(err);
         }
 
-        if self.editor.close_language_servers(None).await.is_err() {
+        let timeout = self.editor.config().lsp.shutdown_timeout;
+        if self.editor.close_language_servers(timeout).await.is_err() && !timeout.is_zero() {
             log::error!("Timed out waiting for language servers to shutdown");
             errs.push(anyhow::format_err!(
                 "Timed out waiting for language servers to shutdown"


the-mikedavis:
This is an interesting use-case. How does `ra-multiplex` respond to shutdown requests and exit notifications? It ignores them?

jackos:
It just drops the connection and doesn't actually send the shutdown response, just saw that this PR fixes it: https://github.com/pr2502/ra-multiplex/pull/14/commits/3be7a2fae2b3371028296b7965ee82c68cf11626

jackos:
But even if that's fixed, I don't want any lag at all when I'm closing windows, for example if I closed down `ra-multiplex` beforehand or it has an error. I'm not sure why Helix needs to wait for shutdown confirmation from the lsp before closing, but I assumed there is a reason hence the configurable option.

the-mikedavis:
The LSP spec is not very explicit about this but I believe that waiting for the server is the proper thing to do during shutdown. If you kill the server process without warning it could hypothetically cause some bad effects. For example a server could have some database connection it needs to close during shutdown and killing the server could hypothetically cause some data loss or corruption.

On the other hand, it would be useful to be able to configure this as an escape hatch to deal with servers that ignore shutdown. Ra-multiplex for one until that PR is merged but also taplo (TOML lsp) has a bug that it ignores shutdowns and this would help with that: https://github.com/helix-editor/helix/issues/5174

So I'm not sure. What do you think @archseer?

dead10ck (approved):
This seems fine to me. I think there has been previous discussion about making this configurable. There are people that are really adamant about quick shutdown times, risks or not, so I don't see a problem with letting them make that choice.

alxpettit:
Any news on this?

alxpettit:
Bump

archseer:
What's the behavior in VSCode/elsewhere?

I'm a bit wary about adding a config option that works around a couple very specific LSPs and allows everyone else to shoot themselves in the foot. It's good practice to let the LSP clean up any temporary state (e.g. I'm sure rust-analyzer uses target/ to some extent), and if the server is correctly implemented the shutdown should still be instant (this is a local connection and the lsp responding to shutdown then exiting shouldn't take more than a millisecond if it doesn't have anything else to do)

Both taplo and ra-multiplex have been fixed by now so I'd really need to see examples of more language servers that require this.

alxpettit:
@archseer 
I think there are some, but I can't recall off the top of my head. Regardless, it doesn't seem unreasonable to have a configuration option that has a lot of warnings around it, for working around bad behavior from specific LSPs. Even if it turned out there were no LSPs with this issue -- there _will_ be again.

I think in this case the potential for footguns is meager enough to not be a concern. As long as people know that overriding the LSP timeout to be too low, or zero, can cause bad behavior, or even corruption if the LSP is caching things to disk or rewriting its own config files. (Should be noted, I haven't seen any such LSPs thus far, so this is most likely a purely speculative future compliance concern at the moment.)

It should be noted that this feature isn't _just_ useful for working around bad behavior from LSPs -- if the user is doing something niche, like having a LSP running off of the machine, for some reason, rather than connecting over loopback (e.g., on-board embedded development), network congestion or lag issues may prompt them to do this sort of thing to improve usability. Or maybe something else is slowing down the LSP -- who knows?

Configurable timeouts are a commonplace feature whenever multiple pieces of software need to interact -- and it seems like a good idea to add that here too.

.gitignore
the-mikedavis, line 1: This change could probably be reset - we don't see this file checked in very often

helix-view/src/editor.rs
```diff
@@ -242,6 +242,13 @@ pub struct LspConfig {
     pub auto_signature_help: bool,
     /// Display docs under signature help popup
     pub display_signature_help_docs: bool,
+    /// Time in milliseconds that the editor will wait for the lsp to shutdown before exiting
+    /// Set to 0 for instant. Defaults to 3000.
+    #[serde(
+        serialize_with = "serialize_duration_millis",
+        deserialize_with = "deserialize_duration_millis"
+    )]
+    pub shutdown_timeout: Duration,
```
the-mikedavis, line 251: This should be a config option per-language-server (here: https://docs.helix-editor.com/languages.html#language-server-configuration) rather than editor-wide

the-mikedavis (changes requested):
I agree with @archseer here that we don't want to introduce this option unless it has specific use-cases and I don't see one other than a language server breaking spec or having a bug that prevents fast-enough shutdown. IME whenever we introduce config options for timeout values like this they tend to be tuned arbitrarily and irresponsibly and it just makes debugging upstream issues harder, so I'm not enthusiastic about adding a config option for this.

But if we do allow configuring this, I want to see it be per-language-server instead of global and it should be documented with a warning that you almost certainly don't want to use it as it might cause bad side-effects for your language server.

alxpettit:
I agree completely with making this configuration per-LSP.

archseer:
> if the user is doing something niche, like having a LSP running off of the machine, for some reason, rather than connecting over loopback (e.g., on-board embedded development), network congestion or lag issues may prompt them to do this sort of thing to improve usability. Or maybe something else is slowing down the LSP -- who knows?

Such a case isn't actually possible since the language server inspects the code and usually needs direct access to the file tree. The server also runs over a file descriptor rather than the network.

> (Should be noted, I haven't seen any such LSPs thus far, so this is most likely a purely speculative future compliance concern at the moment.)

Right, but I also haven't seen any LSPs that would need this workaround (apart from two that are fixed for 6-12 months now) so merging this workaround based on speculation that there could be a usecase eventually doesn't seem right either. 
I'm willing to merge this if a practical need emerges, but it seems redundant to me right now. There's all sorts of timeouts that could actually use a config value (e.g. timeouts that occur when you actually use the LSP for completion etc.) and we haven't even seen demand for those.

> IME whenever we introduce config options for timeout values like this they tend to be tuned arbitrarily and irresponsibly and it just makes debugging upstream issues harder, so I'm not enthusiastic about adding a config option for this.

Exactly, now any time there's an issue we have to ask if they set the timeout to 0 by any chance. We already had enough random issues pop up from users setting the idle timeout to 0, even though the docs say not to do that in bold.

alxpettit:
> Such a case isn't actually possible since the language server inspects the code and usually needs direct access to the file tree. The server also runs over a file descriptor rather than the network.

Actually, it's totally possible! I've done setups _similar_ to this seamlessly combining remote and local development with SSHFS.

alxpettit:
> Exactly, now any time there's an issue we have to ask if they set the timeout to 0 by any chance. We already had enough random issues pop up from users setting the idle timeout to 0, even though the docs say not to do that in bold.

Honestly, IMO that's what bots & issue templating were made for. Denying the user choice because of what stupidity they _might_ get up to kind of rubs me the wrong way.

alxpettit:
>. There isn't a process you could come up with that would fix that.

Fix no. Ameliorate? Yes.

> Yeah I've tried to use that myself — it's hilariously incomplete; I'm shocked that they saw fit to "release" it in the state that it's in. 

Nushell is MVP-grade software. I don't blame them for shipping incomplete features, when they're not even at 1.0, and this feature is strictly opt-in and doesn't affect normal usage.

pascalkuthe:
Closing this as this PR has gone stale and we are not looking to add this kind of config. Thank you for contributing!

snosratiershad:
@pascalkuthe It's almost impossible to use lsps like golangci-lint-langserver with large source codes cause it will always get timed out if you don't wait after saving before closing.  It gets worth as helix is an editor that is natural with closing and re-opening. I don't see any reason to not looking to add this kind of config.

crabdancing:
Honestly I just gave up and forked Helix. @snosratiershad :P

PR #5268, opened 2022-12-23
Allow custom preprocessors for 'vue' injections

farwyler:
Changes the injections for *Vue* `script` and `style` tags to allow for custom preprocessors. Before they were hardcoded to  plain `javascript` and `css`.

*Vue* tooling can specify a preprocessor via the `lang` attribute.
This only adds injections for languages that helix already supports.

e.g.
```vue
<script lang="typescript">
```

or very commonly used
```vue
<style lang="scss">
```

### DIFF
--- a/runtime/queries/vue/injections.scm
+++ b/runtime/queries/vue/injections.scm
@@ -8,13 +8,39 @@
   (raw_text) @injection.content)
  (#set! injection.language "javascript"))
 
+; <script>
 ((script_element
-  (raw_text) @injection.content)
- (#set! injection.language "javascript"))
+    (start_tag) @_no_lang
+    (raw_text) @injection.content)
+  (#not-match? @_no_lang "lang=")
+  (#set! injection.language "javascript"))
 
+; <script lang="js|javascript">
+((script_element
+    (start_tag (attribute (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @injection.content)
+    (#match? @_lang "^(js|javascript)$")
+    (#set! injection.language "javascript"))
+
+; <script lang="ts|typescript">
+((script_element
+    (start_tag (attribute (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @injection.content)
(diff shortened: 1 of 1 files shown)


runtime/queries/vue/injections.scm
```diff
@@ -8,13 +8,39 @@
...
+    (raw_text) @injection.content)
+    (#match? @_lang "^(js|javascript)$")
+    (#set! injection.language "javascript"))
+
+; <script lang="ts|typescript">
+((script_element
+    (start_tag (attribute (quoted_attribute_value (attribute_value) @_lang)))
+    (raw_text) @injection.content)
+    (#match? @_lang "^(ts|typescript)$")
+    (#set! injection.language "typescript"))
```
the-mikedavis, line 30: I think this can be a little cleaner by matching on the `attribute_name`:

```scm
((script_element
   (start_tag
     (attribute
       (attribute_name) @_attr_name
       (quoted_attribute_value (attribute_value) @injection.language)))
   (raw_text) @injection.content)
 (#eq? @_attr_name "lang"))

((script_element (raw_text) @injection.content)
 (#set! injection.language "javascript"))
```

and similar for the style tags

lucario387: Hello, was trying to learn on how to inject typescript/javascript to vue, and the @_no_lang capture may produce false positive 
I've opened a PR on nvim-treesitter to update those queries, you can check it [here](https://github.com/lucario387/nvim-treesitter/commit/d0083cc66461383ca2d3a26eeb746a317707440d)

farwyler: @the-mikedavis thanks! that is much better. i totally forgot about *injection-regex*.

i am not sure if you intended this...
```
((script_element (raw_text) @injection.content)
 (#set! injection.language "javascript"))
```
to act as a fallback for the *no-lang* case. that doesn't seem to work. it just overwrites the language to *js* because it will match any form of *script* tag.

i left in the *no-lang* query but updated the script/style matches with the *lang* attribute.

the-mikedavis (approved):

Issue #5262, opened 2022-12-22
use q key to any quotes in Match


otavioschwanck:
`miq` would execute mi', mi", mi`, depending on where you are.  Super easier to select
`maq` same thing.

the-mikedavis:
I think it would be preferrable to fix https://github.com/helix-editor/helix/issues/4475 rather than add a textobject specific to quotes

otavioschwanck:
> I think it would be preferrable to fix #4475 rather than add a textobject specific to quotes

sometimes the cursor are inside a string, but you want a quote.  In those case, is better to have the any quotes.

PR #5269, opened 2022-12-23
feat: textobject any quotes

otavioschwanck:
Resolves https://github.com/helix-editor/helix/issues/5262

edit:

Example why is better to have in a separated key:

miq to text:
image

mim to curly brackets:
image


pascalkuthe:
Closing this as stale. I also don't see us going in this direction as this will be covered by #8294 so there is no need to have a second keymap that does the exact same thing.

Thank you for contributing!

PR #5274, opened 2022-12-24
Easier way to implement mouse action

hunterliao29:
Developed an easier way to implement mouse action.

```rust
mouse_actions: HashMap<MouseEventKind, Vec<(Rect, MappableCommand)>>

pub fn add_mouse_action(
        &mut self,
        mouse_event: MouseEventKind,
        area: Rect,
        action: MappableCommand,
    ) {
        if self.mosue_enable {
            match self.mouse_actions.get_mut(&mouse_event) {
                Some(commands) => commands.push((area, action)),
                None => {
                    self.mouse_actions.insert(mouse_event, vec![(area, action)]);
                }
            }
        }
    }

let has_click_action = |kind, row, column| match self.mouse_actions.get(kind) {
    Some(commands) => commands.iter().rev().find(|(area, _)| {
        if area.x <= column
            && area.x + area.width > column
            && area.y <= row
            && area.y + area.height > row
        {
            return true;
        }
        false
    }),
    None => None,
};

if let Some((_, cmd)) = has_click_action(&kind, row, column) {
    cmd.execute(cxt);
    return EventResult::Consumed(None);
}
```

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -9,6 +9,8 @@
 | `:buffer-close-others!`, `:bco!`, `:bcloseother!` | Force close all buffers but the currently focused one. |
 | `:buffer-close-all`, `:bca`, `:bcloseall` | Close all buffers without quitting. |
 | `:buffer-close-all!`, `:bca!`, `:bcloseall!` | Force close all buffers ignoring unsaved changes without quitting. |
+| `:buffer-close-nth`, `:bcn`, `:bclosenth` | Close nth buffer without quitting. |
+| `:buffer-nth`, `:bnth` | Goto nth buffer. |
 | `:buffer-next`, `:bn`, `:bnext` | Goto next buffer. |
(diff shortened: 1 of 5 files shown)


archseer:
Partial duplicate of https://github.com/helix-editor/helix/pull/5041 but I need to review the implementation

hunterliao29:
Screenshot 2022-12-24 at 3 46 23 PM

hunterliao29:
I am not sure if set_stringn should appends a `…` at the end of truncated lines. if this is not intended, I will change it back.

hunterliao29:
Hi. I have restructure/reimplement the implementation so that it can be reuse for adding mouse action to other elements.

@archseer 
I am thinking about something like this to imitate keymap, but not sure if this is a good idea, and if I should keep woking on this.
``` rust
enum MouseMode {
  Click=0,
  Drug=1,
  Scroll=2,
}
mousemat = HashMap<MouseMode,Vec<Rect, MappableCommand >>
```

Thank you.

PR #5279, opened 2022-12-24
Add statusline option: `filename-display-mode`

theteachr:
Allows for showing only the filename instead of the path which can sometimes be long.

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -2,6 +2,7 @@ use helix_core::{coords_at_pos, encoding, Position};
 use helix_lsp::lsp::DiagnosticSeverity;
 use helix_view::{
     document::{Mode, SCRATCH_BUFFER_NAME},
+    editor::FilenameDisplayMode,
     graphics::Rect,
     theme::Style,
     Document, Editor, View,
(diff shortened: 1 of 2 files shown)


Ordoviz:
I'm not a maintainer, but this should probably be called `file-basename`.

the-mikedavis:
Instead of an option to control the element I think this should be a new statusline element

theteachr:
> Instead of an option to control the element I think this should be a new statusline element

Could we make the `FileName` (might as well rename to `FileBaseName`?) element be just the name of the file and add another for the parent directory?

the-mikedavis:
I don't think we need to modify the current elements. Statusline elements usually have spacing around them that would make it look odd to have a separate element for the basename and the parent directory/directories

theteachr:
I tested with adding a new element which just renders the file base name. I have the following snippet in my Helix Config now.

```toml
[editor.statusline]
left = ["mode", "spinner", "file-base-name"]
```

But I'm concerned that this can lead to a configuration with both `file-name` and `file-base-name` that would be redundant.

Anyway, I can close this one and raise another PR where I have added a new `StatusLineElement` instead of introducing an option.

the-mikedavis:
I don't think that's a bad thing - it should be possible to have both `file-name` and `file-base-name`

theteachr:
Closing this in favor of https://github.com/helix-editor/helix/pull/5318.

Issue #3495, opened 2022-08-21
Command to open "languages.toml" similar to "config-open"


IvanGoncharov:
I'm messing a lot with `languages.toml` after switching to helix as my primary editor.
It would be great to have some way to open/reload it quickly.

Maybe just add it as an argument to config-open, e.g., `config-open languages`
If, in the future other configuration files (e.g. `plugins.toml`) it would be useful to have one command with different arguments.

A more radical change would be to make "config-open" open a file picker in the config directory.

the-mikedavis:
This was the idea of https://github.com/helix-editor/helix/issues/3206 when opened. A command for opening language config may be more confusing than helpful since there are three possible language configurations and not all of them are editable.

gerep:
You can create a shortcut in your `config.toml`.

```toml
[keys.normal]
"C-l" = ":open ~/.config/helix/languages.toml"
```

And pressing `ctrl+l` will open the file as a new buffer.

danriedl:
+1, also having the same thing within the "config-open-workspace" to open the language file from there.
I have to quit helix and open it directly because of course `.helix` is in `.gitignore`.

Issue #3206, opened 2022-07-27
Add lang-open command for languages.toml


AceofSpades5757:
#### Describe your feature request

Add a `lang-open` command, just like the `config-open` and `log-open` commands, to open the platform-specific config file.

If this isn't already in the works (I didn't see any related issues or pull requests), then I have the bandwidth to make a pull request.

the-mikedavis:
The language config is merged from `.helix/languages.toml`, `~/.config/helix/languages.toml` and the compiled-in `languages.toml` at the root of the repo.  Because of the merging behavior I don't think it makes sense to have a `:lang-open` command. If anything it might make sense to have it open a picker with up to all three of those files but that would most likely just be confusing and these `:*-open` commands are meant to be as simple and straightforward as possible. Plus you can't edit the compiled-in `languages.toml`

AceofSpades5757:
It would still be nice to have a nice, built-in way to open the user/local languages.toml file.

AceofSpades5757:
Those are really good points though. What about a config picker for all config files, `~/.config/helix/config.toml`, `~/.config/helix/languages.toml`, `.helix/languages.toml`, and any future config files that get added to helix? For the future configs, I was thinking there may be individual configuration files based on language, so you may just want a picker to quickly edit the right one. If needed, users can always add them to their `config.toml` files as a keymapping.

AceofSpades5757:
Could also `config-open` to default go to `~/.config/helix/config.toml`, and give it an auto-completer for any other, relevant config files as a second argument.

e.g. `config-open` does the same thing it does currently.
e.g. `config-open ~/.config/helix/languages.toml` would be an auto-complete option.

Could instead give users named files, like `User Languages Config`, `Project Languages Config` (`.helix/languages.toml`), etc.

the-mikedavis:
`:config-open` and `:log-open` are meant as very beginner-friendly ways to open up files so you don't need to read the documentation to know where they are. Language config is more complicated: you need to read the language config documentation to understand the behavior.

Optimizing how easily we can open config files seems like hair-splitting to me. Instead I think it would be better to improve the in-editor documentation of configuration. The idea is half-baked but it would be nice to have a man-page that explains what each file does and covers the merging behavior and then you could `gf` on the paths it has to open the files.

AceofSpades5757:
The new issue title and description should, hopefully, lead to something better.

willparsons:
[...]

PR #5283, opened 2022-12-24
Add command to open user's language.toml file

JorgeLDB:
Added "open-language-config". It also has an alias "config-lang-open".

image

I'd like to request suggestions in case the names are not clear enough.


the-mikedavis:
There is prior discussion about this: https://github.com/helix-editor/helix/issues/3206, https://github.com/helix-editor/helix/issues/3495

pascalkuthe:
I am going to be colsing this out as stale. Considering that we are making progress towards integrating scheme/a scheme based config (which would nit have a separate language config file). I don't think this is worth adding at this point.

Thank you for contributing!

PR #5288, opened 2022-12-25
fix(commands): extend_line to proper line when count and current line selected

### LINKED ISSUE #5270: xx works ok; x2x works err
In the doc, it said: 

> x | Select current line, if already selected, extend to next line

but
1. x + x, works ok
2. x + 2x, works err, it didn't select 3 lines, it did select 2 lines.
3. 3x, works ok.
4. 2x + 2x, works err, it didn't select 4 lines, it did select 3 lines.

### Platform
Debian 12 testing, Konsole, helix 22.12 (bdeefbfb)

gabydd:
for both extend_line_(above/below) the check for if current line is selected now works if there is a count:

this now means if you select a line with ``x`` and then do ``2x`` 3 lines will be selected whereas before only 2 lines would be, this keeps ``2x`` on a not fully selected line only selecting 2 lines which is the desired functionality of ``extend_line``

fixes #5270

### DIFF
--- a/helix-term/tests/test/commands.rs
+++ b/helix-term/tests/test/commands.rs
@@ -316,17 +316,39 @@ async fn test_undo_redo() -> anyhow::Result<()> {
 async fn test_extend_line() -> anyhow::Result<()> {
     // extend with line selected then count
     test((
-        platform_line("#[a|]#\na\na\n\n").as_str(),
+        platform_line(indoc! {"\
+            #[l|]#orem
+            ipsum
+            dolor
+            
+            "})
+        .as_str(),
         "x2x",
-        platform_line("#[a\na\na\n|]#\n").as_str(),
+        platform_line(indoc! {"\
+            #[lorem
+            ipsum
+            dolor
+            |]#
+            "})
+        .as_str(),
     ))
     .await?;
 
     // extend with count on partial selection
     test((
-        platform_line("#[a|]#\na\n\n").as_str(),
+        platform_line(indoc! {"\
+            #[l|]#orem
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
This would be nice to add an integration test for in `helix-term/tests/test/commands.rs`

gabydd:
Sorry about that, I was thinking about doing that but totally forgot. Integration tests have been added now, it needed a trailing newline to work cause the test would add an extra newline after end if it didn't exist.

helix-term/tests/test/commands.rs
```diff
@@ -311,3 +311,24 @@ async fn test_undo_redo() -> anyhow::Result<()> {
...
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_extend_line() -> anyhow::Result<()> {
+    // extend with line selected then count
+    test((
+        platform_line("#[a|]#\na\na\n\n").as_str(),
+        "x2x",
+        platform_line("#[a\na\na\n|]#\n").as_str(),
+    ))
```
the-mikedavis, line 322: I think the `indoc!` format for these would be a little more readable, for example https://github.com/helix-editor/helix/blob/63dcaae1b9083396fb3faaef9eaa2421f7e48fb9/helix-term/tests/test/commands.rs#L101-L115

The `\n` newlines make it hard to tell which lines are selected at a glance

gabydd: Fixed now

the-mikedavis (approved):

PR #5289, opened 2022-12-25
docs/book: map `esc` key in config.toml to escape multi-cursor and non-normal modes to help new users

x448:
PR was updated after reading comments by @gibbz00 and @the-mikedavis.

The example config.toml embedded in docs will be most useful to new users and reviewers who don't know how to map a key .  At that level of experience, their frustration will be from:
- not knowing how to map keys in config.toml
- not having `esc` key reset Helix to fully default state of normal mode, single cursor, etc.
- getting false positive error about lack of true color support (e.g. Linux distros in WSL2 on Windows 11)

This PR resolves all of the above in a simple update to the example config embedded in the docs.  This PR does not promote this config for use by non-beginners.  It's just an example showing how use config.toml to map a key and resolve early challenges.

Non-beginners are unlikely to even look at the example config embedded in the docs, so this config can safely be ignored by everyone already familiar with Helix and it won't affect them.

I'm not the only new user to initially wonder why `esc` key doesn't reset to fully default state in Helix.  See "Taking a look at the Helix editor" video at around [14m37s (youtube)](https://www.youtube.com/watch?v=8L308PdmhMY&t=877s).

Here's what is added to the example config in the docs:

```toml
# Uncomment true-color to fix false positive "Unsupported theme: theme requires true color support."
# true-color = true

# Mapping a key is easy.  E.g. Vim users can map 0 to go to start of line.
# Experienced Helix users prefer that collapsing and keeping the primary
# selection be explicit operations and distinct from changing modes.
# However, beginners can map esc key to reset Helix to fully default state of
# normal mode, single cursor, and not having multiple chars or words selected.
[keys.insert]
esc = ["collapse_selection", "normal_mode"]

[keys.normal]
esc = ["collapse_selection", "keep_primary_selection"]
0   = "goto_line_start"

[keys.select]
esc = ["collapse_selection", "keep_primary_selection", "normal_mode"]
0   = "goto_line_start"
```
Again, this update to docs won't affect existing users and will only reduce problems for new users and reviewers.  It should improve positive reviews and increase conversion rate of new users trying Helix for the first time by creating a better first impression.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -23,6 +23,15 @@ select = "underline"
 
 [editor.file-picker]
 hidden = false
+
+[keys.insert]
+esc = ["collapse_selection", "normal_mode"]
+
+[keys.normal]
+esc = ["collapse_selection", "keep_primary_selection"]
+
+[keys.select]
+esc = ["collapse_selection", "keep_primary_selection", "normal_mode"]
 ```
 
 You may also specify a file to use for configuration with the `-c` or
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -16,6 +16,9 @@ theme = "onedark"
 line-number = "relative"
 mouse = false
 
+# Uncomment true-color to fix false positive "Unsupported theme: theme requires true color support."
+# true-color = true
+
 [editor.cursor-shape]
 insert = "bar"
 normal = "block"
@@ -24,14 +27,21 @@ select = "underline"
 [editor.file-picker]
 hidden = false
 
+# Mapping a key is easy.  E.g. Vim users can map 0 to go to start of line.
+# Experienced Helix users prefer that collapsing and keeping the primary
+# selection be explicit operations and distinct from changing modes.
+# However, beginners can map esc key to reset Helix to fully default state of
+# normal mode, single cursor, and not having multiple chars or words selected.
 [keys.insert]
 esc = ["collapse_selection", "normal_mode"]
 
 [keys.normal]
 esc = ["collapse_selection", "keep_primary_selection"]
+0   = "goto_line_start"
 
 [keys.select]
 esc = ["collapse_selection", "keep_primary_selection", "normal_mode"]
+0   = "goto_line_start"
 ```
 
 You may also specify a file to use for configuration with the `-c` or


gibbz00:
I would advice against binding `collapse_selection` and `keep_primary_selection` to the same key as I have tried that myself. You run often into the problem of needing to collapse the selection only when working with multiple cursors. Binding `collapse_selection` to escape in normal mode suffices.

x448:
> Binding collapse_selection to escape in normal mode suffices.

That didn't work for me:

```toml
[keys.normal]
# With this, pressing esc key doesn't escape out of multi-cursor.
# Try pressing esc key after pressing Shift-c a few times.
esc = ["collapse_selection"]
```

This works for me (same mapping as PR):
```toml
[keys.normal]
# With this, pressing esc key successfully escapes out of multi-cursor.
# Try pressing esc key after pressing Shift-c a few times.
esc = ["collapse_selection", "keep_primary_selection"]
```

New users will try pressing `esc` key to escape out of multi-cursor and non-normal modes.  The example config.toml will mostly be viewed by new users and it would make onboarding smoother.

gibbz00:
Well yes, that is the point of what I was trying to say. **`collapse_selection` and `keep_primary_cursors` have different purposes that aren't always used at the same time, so they shouldn't be bound together.**

x448:
I updated title and the PR description to make it more clear about mapping `esc` to escape multi-cursor and non-normal modes.

Reviewers and other new users wonder why `esc` key doesn't already do this in Helix.  See "Taking a look at the Helix editor" video at around [14m37s (youtube)](https://www.youtube.com/watch?v=8L308PdmhMY&t=877s).

More experienced Helix users can continue using `,` and `;` because mapping `esc` doesn't take that away.

gibbz00:
I get where you're coming from, and I do agree with you that escape is a good binding for a reset behavior. But not for both at the same time. I myself have `"esc" = "collapse_selection"` to normal mode only, as I sometimes want to exit select mode without collapsing the selection too.

x448:
>  I myself have "esc" = "collapse_selection" to normal mode only, as I sometimes want to exit select mode without collapsing the selection too.

I understand your preference and reasoning for not wanting to press `esc` to get back to fully default state.

Experienced users who prefer fine-grained control of pressing multiple different keys to get back to the default state will probably not be looking at example config.toml as a quick start guide.  They'll have already moved on to other parts of the docs.

For this reason, this PR only changes the example config.toml in the docs/book (aimed primarily at new users and reviewers trying Helix for the first time).  This PR doesn't change default behavior of Helix for current users.

New users and reviewers trying Helix for the first time and reading example config.toml will prefer pressing `esc` key just once to get back to the default state of normal mode, one active cursor, and not having multiple characters or words selected.

Updating the example config.toml will make the first 5 minutes with Helix less frustrating and [improve reviews (youtube)](https://www.youtube.com/watch?v=8L308PdmhMY&t=877s) by forming a more positive [confirmation bias](https://en.wikipedia.org/wiki/Confirmation_bias).

the-mikedavis:
This config block is meant to show off an example of how a config file can be written rather than be a good example for configuration. (It's more about the format than the contents.)

I think this would fit better as a config recommendation in https://github.com/helix-editor/helix/discussions/1360.

gibbz00:
@the-mikedavis Has there been any discussion about also bindning  `collapse_selection` to escape? @x448 brought up a lot of good points regarding adoption that I haven't been able to put into words as well as he did.

the-mikedavis:
I wouldn't like to see that change to the defaults. The current keys are already very easy to hit and I would prefer that collapsing and keeping the primary selection be explicit operations and distinct from changing modes

x448:
> The current keys are already very easy to hit and I would prefer that collapsing and keeping the primary selection be explicit operations and distinct from changing modes

I understand your preference and reasoning.  For this reason, the PR only updates an example config.toml embedded in the docs that is unlikely to be read by users who already use Helix.   It is not meant to promote the example config as the preferred config.

Motivation for this PR is to help new users avoid problems I encountered as a new user.  It should increase conversion rate of new users trying Helix for the first time and improve positive reviews like this one by creating a more positive initial impression. 

Based on comments from @gibbz00 and @the-mikedavis, I updated the PR description and added a commit.

the-mikedavis:
The doc is essentially written as a recommendation for beginners though. We shouldn't recommend configs like this that 'simplify' the keymap because the changes become part of people's muscle memory and prevent them from learning how to use the command separately.

I don't forsee us accepting a change like this to the docs so I'm closing this out

PR #5290, opened 2022-12-25
Update grammar and highlight queries for slint grammar rewrite

jrmoulton:
This is just updating the grammar rev and queries to be updated to the latest slint grammar.

### DIFF
--- a/runtime/queries/slint/highlights.scm
+++ b/runtime/queries/slint/highlights.scm
@@ -1,77 +1,106 @@
-
-(user_type_identifier) @type
-
-(var_identifier) @variable
-
-(state_identifier) @variable.other.member
-
-(var_identifier
-  (post_identifier) @variable)
-
-(function_identifier) @function
-
-(reference_identifier) @keyword.storage.modifier.ref
-(visibility_modifier) @keyword.storage.modifier
-
+(identifier) @variable
+(type_identifier) @type
 (comment) @comment
+(int_literal) @number
+(float_literal) @float
+(string_literal) @string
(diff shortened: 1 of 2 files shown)


jrmoulton:
Ahh. I accidentally have an extra unrelated commit in here. I'll fix that

jrmoulton:
Fixed

runtime/queries/slint/highlights.scm
```diff
@@ -118,8 +148,7 @@
  "*="
  "/="
  "?"
-
  "=>"
  ] @operator
 
-(ternary_expression [":" "?"] @keyword.control.conditional)
\ No newline at end of file
+(ternary_expression [":" "?"] @conditional)
```
the-mikedavis: Helix uses different captures for highlighting than neovim - https://docs.helix-editor.com/master/themes.html#syntax-highlighting- so this change should be undone

There are some other non-existent scopes that should be changed too, for example `@float` should become `@constant.numeric.float`

the-mikedavis (commented):
It looks like `local.scm` will need to be updated as well since it contains a node that doesn't exist in the new version of the grammar: https://github.com/helix-editor/helix/actions/runs/3777270745/jobs/6422763958

lauglam:
Has there been any progress on this issue?

jrmoulton:
I haven’t worked on this in a while. There isn’t much that needs changed at this point but the queries need to be updated for helix

gabydd:
Closed by https://github.com/helix-editor/helix/pull/7893

Issue #3204, opened 2022-07-26
Command to open external documentation for item under cursor


antoyo:
#### Describe your feature request

I'd like a command to open the external documentation for an item under the cursor, like [this](https://www.reddit.com/r/rust/comments/w7cf7d/rustanalyzer_tip_open_docsrs_for_the_symbol_under/).
For instance, it will open docs.rs for a symbol in a Rust buffer.

AlexanderBrevig:
Seem like this is a [rust-analyzer LSP Extension](https://github.com/rust-lang/rust-analyzer/blob/master/docs/dev/lsp-extensions.md#open-external-documentation) and as such would only work for `rust` lang. I started looking into implementing this, but got a bit unsure of the best way forward as it's not part of the current LSP spec.

Ideas?

the-mikedavis:
There was some prior discussion[^1] about language-specific LSP extensions. It would be perfect for a language-specific plugin once the plugin-system is featureful enough to support it, but it's a bad fit for core since it's specific to language or small subset of languages. For now it's best to keep any necessary changes up to date in a fork

[^1]: I can't find the links to the issues/discussions/matrix-chats but I think this has come up for non-standard typescript/deno location types and some omnisharp and solargraphs messages. I'll update this comment if I can find them

PR #5295, opened 2022-12-26
Add support for clangd's switchSourceHeader

bciobanu:
Depends on https://github.com/gluon-lang/lsp-types/pull/254.

clangd supports several extensions outside the Language Server Protocol specification.
One of them is the functionality to determine the corresponding source/header file from another file: https://clangd.llvm.org/extensions#switch-between-sourceheader.

Unfortunately clangd does not explicitly mention this command in its capabilities list, i.e.
```
2022-12-26T10:43:40.798 helix_lsp::transport [INFO] <- {"id":0,"jsonrpc":"2.0","result":{"capabilities":{"astProvider":true,"callHierarchyProvider":true,"codeActionProvider":{"codeActionKinds":["quickfix","refactor","info"]},"compilationDatabase":{"automaticReload":true},"completionProvider":{"allCommitCharacters":[" ","\t","(",")","[","]","{","}","<",">",":",";",",","+","-","/","*","%","^","&","#","?",".","=","\"","'","|"],"resolveProvider":false,"triggerCharacters":[".","<",">",":","\"","/","*"]},"declarationProvider":true,"definitionProvider":true,"documentFormattingProvider":true,"documentHighlightProvider":true,"documentLinkProvider":{"resolveProvider":false},"documentOnTypeFormattingProvider":{"firstTriggerCharacter":"\n","moreTriggerCharacter":[]},"documentRangeFormattingProvider":true,"documentSymbolProvider":true,"executeCommandProvider":{"commands":["clangd.applyFix","clangd.applyTweak"]},"hoverProvider":true,"implementationProvider":true,"memoryUsageProvider":true,"referencesProvider":true,"renameProvider":true,"selectionRangeProvider":true,"semanticTokensProvider":{"full":{"delta":true},"legend":{"tokenModifiers":["declaration","deprecated","deduced","readonly","static","abstract","virtual","dependentName","defaultLibrary","usedAsMutableReference","functionScope","classScope","fileScope","globalScope"],"tokenTypes":["variable","variable","parameter","function","method","function","property","variable","class","interface","enum","enumMember","type","type","unknown","namespace","typeParameter","concept","type","macro","comment"]},"range":false},"signatureHelpProvider":{"triggerCharacters":["(",","]},"textDocumentSync":{"change":2,"openClose":true,"save":true},"typeHierarchyProvider":true,"workspaceSymbolProvider":true},"serverInfo":{"name":"clangd","version":"Apple clangd version 14.0.0 (clang-1400.0.29.202) mac+xpc x86_64-apple-darwin22.2.0; target=arm64-apple-darwin22.2.0"}}}
```

so I opted to check for the server_info name instead. We could instead send the request regardless, but from testing with some other language servers, there was no response for this request. And that meant the user never got an error message about it, which felt like a bad UX.

Tested:
- [x] cargo integration-test
- [x] Verified functionality with clangd (for missing and existing source/headers)
- [x] Verified this is a noop with rls and an error message is displayed

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -32,6 +32,7 @@ pub struct Client {
     server_tx: UnboundedSender<Payload>,
     request_counter: AtomicU64,
     pub(crate) capabilities: OnceCell<lsp::ServerCapabilities>,
+    pub(crate) server_info: OnceCell<lsp::ServerInfo>,
     offset_encoding: OffsetEncoding,
     config: Option<Value>,
     root_path: std::path::PathBuf,
@@ -104,6 +105,7 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             capabilities: OnceCell::new(),
+            server_info: OnceCell::new(),
             offset_encoding: OffsetEncoding::Utf8,
             config,
             req_timeout,
@@ -144,6 +146,10 @@ impl Client {
         self.capabilities
             .get()
             .expect("language server not yet initialized!")
+    }
+
+    pub fn server_info(&self) -> Option<&lsp::ServerInfo> {
+        self.server_info.get()
     }
 
     pub fn offset_encoding(&self) -> OffsetEncoding {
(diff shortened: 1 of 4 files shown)


the-mikedavis:
Extensions to the protocol are a better fit for plugins: https://github.com/helix-editor/helix/issues/3204#issuecomment-1196465602

Until the plugin system exists, maintaining a fork with the code-change is the best way to have this implemented. It's not a good fit for core though unless it's part of the LSP spec.

bciobanu:
> #3204 (comment)

Thanks, I'll keep an eye out for that other issue then.

PR #5296, opened 2022-12-26
Add `space-w-a` as `select_all`

geometryolife:


gibbz00:
Please don't. Space-W are for window navigation commands, and select_all is already bound to %.

geometryolife:
> Please don't. Space-W are for window navigation commands, and select_all is already bound to %.

Yes, I know `%` is bound to the `select_all` command, but I think `space-w-a` gives the user a more comfortable keystroke experience.

sudormrfbin:
`%` already exists as a keybind, and this feels too specific and conflicting with window mode to be shipped as a default. This would be better off as a part of your personal `config.toml`.

gibbz00:
> `%` already exists as a keybind, and this feels too specific and conflicting with window mode to be shipped as a default. This would be better off as a part of your personal `config.toml`.

Giving a less rude reply along with a constructive action puts my response into shame hehe. If only more of us were like @sudormrfbin ✌️

geometryolife:
Thanks for the feedback, @gibbz00 @sudormrfbin  if you don't recommend doing this, then I think it should really be put in a custom configuration.

PR #5303, opened 2022-12-26
Add save on idle 🚀

ArchUsr64:
- Implements an `auto-save` like feature on idle timer trigger events.
- Useful for immediate feedback from `rust-analyzer` as most diagnostics are only refreshed on file save.
- Configured as a boolean `idle-save` in the `[editor]` section of the config file, `false` by default. 

Video showcasing the feature with `rust-analyzer` language server **(**`idle-timeout` is set to `0`**)**

https://user-images.githubusercontent.com/83179501/209581216-c29e581b-c17f-4e63-b2b8-6228e9cdff93.mp4

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -50,7 +50,8 @@ on unix operating systems.
 | `auto-completion` | Enable automatic pop up of auto-completion. | `true` |
 | `auto-format` | Enable automatic formatting on save. | `true` |
 | `auto-save` | Enable automatic saving on focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal. | `false` |
-| `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. Used for autocompletion, set to 0 for instant. | `400` |
+| `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. Used for autocompletion and idle-save, set to 0 for instant. | `400` |
+| `idle-save` | Enable automatic saving on idle timer trigger. | `false` |
 | `completion-trigger-len` | The min-length of word under cursor to trigger autocompletion | `2` |
 | `auto-info` | Whether to display infoboxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative. | `false` |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
There was some discussion on the PR that added auto-save on focus loss: https://github.com/helix-editor/helix/pull/3178#discussion_r998569273

This could be added to the `auto-save` configuration so that you could configure either or both events for auto saving:

```toml
auto-save = ["focus-lost", "idle"]
```

I'm also thinking that the idle-timeout is too short for this - it will save very aggressively. An auto-save timer should probably have a much longer duration like 10+ seconds

ArchUsr64:
> This could be added to the `auto-save` configuration so that you could configure either or both events for auto saving:
> 
> ```toml
> auto-save = ["focus-lost", "idle"]
> ```
> 
> I'm also thinking that the idle-timeout is too short for this - it will save very aggressively. An auto-save timer should probably have a much longer duration like 10+ seconds

Yeah, I think having a separate timer and all relevant configuration being done in `auto-save` option would be preferred. 

Though I'm not sure what should be the behavior in case of a boolean value for `auto-save`, which should be handled properly in order to not break existing configurations. 

Perhaps, `true` could just be an alias for current behavior of focus-lost and have `false` disable both to avoid interference with existing configs.

Connortsui20:
What was the reason why this was dropped? This seems like a good change...

pascalkuthe:
Author probably lost interest. Either way I do think this needs a seperate timers, piling more on idle tineout is not a good idea and idle-timeout is also way to short for autosave. I agree with @the-mikedavis this needs atleast a 10s timer. Probably even longer. The config also needs to be changed as mentioned above.


For the timer this should probably be based on #8021 which makes it really easy to add new event based timers. In this case the timer should be reset whenever a document revision is created (saving in insert mode is broken so that should not.happen there).

ArchUsr64:
> What was the reason why this was dropped? This seems like a good change...

The implementation was shotty at best, I'm looking to pick this back up once I have some time on hand.

mersenne-twister:
> this needs atleast a 10s timer.
I personally feel that it would be great to have a shorter timer, of maybe a second or two, for the purpose of code diagnostics. Regardless, I think it would definitely make sense to have it be a configuration option, with a reasonable default of 10+ s.

Issue #5284, opened 2022-12-24
Add an example for WORDs to the tutor


matklad:
https://github.com/helix-editor/helix/pull/5247 split off "WORDS and words" section inn thee tutor:

https://github.com/helix-editor/helix/blob/1af76b738dd5bdae14b025d07d3001c4ad23e071/runtime/tutor#L265-L286

It would be great to add some example here!

woojiq:
@the-mikedavis maybe it needs to be closed since the pr with the changes was merged?

PR #5304, opened 2022-12-26
Add tutor example for WORDS

mrjerzy:
As described in issues https://github.com/helix-editor/helix/issues/5284, we are currently missing a good example for the usage of 'W' compared to 'w' in the tutor.

This pull request is adding an example to the existing chapter "3.3 WORDS AND words"


runtime/tutor
```diff
@@ -271,18 +271,17 @@ _________________________________________________________________
...
+ 2. Type w repeatedly to select individual words until you
+    reach the end of the line.
+ 3. Note that 'one-of-a-kind' required 7 keystrokes to be
+    traversed. '"modal"' required 3 keystrokes.
+ 4. Move the cursor back to beginning of the line marked '-->'.
+ 5. Type W repeatedly to select individual WORDS.
+ 6. Note that 'one-of-a-kind' and '"modal"' have been selected
+    both with one keystroke each.
+
+--> Helix is a one-of-a-kind "modal" text editor
```
the-mikedavis, line 284:
```suggestion
--> Helix is a one-of-a-kind "modal" text editor

```

Looks like this section needs one more line to have the same height as the rest of the sections

the-mikedavis (approved):

Issue #3842, opened 2022-09-14
many built-in themes do not have a separate highlight color for the primary selection


BrianHicks:
### Summary

I've been trying to find a theme I like for use in Helix. Not hard—there are many good options—but I've had to decide against a lot of them because they do not have a separate color for the primary selection. (I use multiple selections a lot, so this is a problem for me!) Themes that seem to have this problem:

- acme
- autumn
- autumn_night
- ayu_dark
- ayu_light
- ayu_mirage
- base16_default
- base16_default_dark
- base16_default_light
- base16_terminal
- base16_transparent
- bogster
- boo_berry
- default (oh no!)
- doom_acario_dark
- everforest_dark
- everforest_light
- flatwhite
- fleetish
- gruvbox
- gruvbox_light
- noctis_bordo
- nord_light
- papercolor-light
- penumbra+
- rose_pine
- rose_pine_dawn
- rose_pine_moon
- serika-dark
- serika-light
- sonokai
- tokyonight
- tokyonight_storm

### Reproduction Steps

Load up a big file (I used `/usr/share/dict/words`) and select some lines, then hit a-s to split into multiple lines. Cycle through the themes. I expected that all (or at least most) of the themes would show a difference between the primary and non-primary selections, but many of them do not.

https://asciinema.org/a/9gvV6uuz0QKLppxU9hA2aLLbo

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

macOS

### Terminal Emulator

wezterm 20220426-073620-4d09fe1f

### Helix Version

helix 22.08.1 (66276ce6)

Omnikar:
I had opened an issue about this a while ago: #1833

BrianHicks:
oh! Sorry, I searched around but didn't find that one. Regardless, that's the complete list of themes with this problem at the current release.

AlexanderBrevig:
I've added this as a requirement to #3234

archseer:
For me this is not a problem but a feature. Since all selections are edited at once, there is no "primary" apart from view focus. The default theme deliberately doesn't style it differently.

AlexanderBrevig:
> For me this is not a problem but a feature. Since all selections are edited at once, there is no "primary" apart from view focus. The default theme deliberately doesn't style it differently.

It's not always apparent what `,` will do then, especially for beginners. 
[...]

PR #5309, opened 2022-12-26
Add a note to section 10.1 of tutor

x448:
Add a note to the tutor in section 10.1 to
prevent new users from thinking Helix is buggy
or that the user is doing something wrong.

NOTE: If typing ) or ( appears to do nothing, try using a theme
 that uses different colors for primary selection and cursor.
 These themes work well here:
 - dark_plus
 - emacs

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -1007,7 +1007,10 @@ lines.
  --> How much would would a wouldchuck chuck
  --> if a wouldchuck could chuck would?
 
-
+ NOTE: If typing ) or ( appears to do nothing, try using a 
+ different color theme. These themes work well here:
+ - dark_plus
+ - emacs
 
 
 =================================================================


yashpalgoyal1304:
just tested on hx v22.05, and I'd say that this is a bug indeed with the default theme.

yashpalgoyal1304:
This pr is like a patch for the bahaviour mentioned below:

> Just to add to this: in section 10.1 of the `hx --tutor`, the features of cycling and removing selections are demonstrated. Since the default theme does not have a way to distinguish the primary selection, it appears that `(` and `)` don't do anything (especially to someone new to `helix`). Basically, it gives the impression that something is broken.

_Originally posted by @taranlu-houzz in https://github.com/helix-editor/helix/issues/3842#issuecomment-1270597054_

x448:
I think this note will be useful even if all the current themes are fixed because some future themes might have similar problems.

yashpalgoyal1304:
> I think this note will be useful even if all the current themes are fixed because some future themes might have similar problems.

aahw, nice. fair point. i was thinking about its future use, but I wasn't able to see how it will work together, what will be the story behind.

say a user installs helix, and changes the theme in due course, but then after some time uses the helix tutor.

so, for those cases too, i'd like to point users to default theme. 'cz :
* hey, the default theme should work out of the box for all the crucial things.
* and we shouldn't refer to different themes on a per-case basis.

runtime/tutor
```diff
@@ -1007,7 +1007,10 @@ lines.
  --> How much would would a wouldchuck chuck
  --> if a wouldchuck could chuck would?
 
-
+ NOTE: If typing ) or ( appears to do nothing, try using a 
+ different color theme. These themes work well here:
+ - dark_plus
+ - emacs
```
yashpalgoyal1304, line 1013:
```suggestion
 NOTE: If typing ) or ( appears to do nothing, try using the
 default color theme.
```

yashpalgoyal1304: ofcourse this will come after the default theme is fixed...

x448: > aahw, nice. fair point. i was thinking about its future use, but I wasn't able to see how it will work together, what will be the story behind.
>
>say a user installs helix, and changes the theme in due course, but then after some time uses the helix tutor.
>
>so, for those cases too, i'd like to point users to default theme. 'cz :
>
>    hey, the default theme should work out of the box for all the crucial things.
>    and we shouldn't refer to different themes on a per-case basis.
>
> but ofcourse, that requires fixing the default theme first. that's what that other pr is for
>
> ofcourse this will come after the default theme is fixed...

That's an interesting approach.  I'm not sure it has quantifiable advantages because this PR
- resolves a current problem without depending on a future PR fixing any theme
- mitigates a future problem if the user chooses a buggy theme we don't know about
- doesn't prevent default theme from being suggested once it is fixed (the PR fixing the default theme can include a commit to suggest using it inside the tutorial)

Why not merge this PR as-is first, then once the default theme is fixed, open a new PR to include it as a recommended workaround?  This way would be less fragile because it doesn't introduce external dependencies.

If the next release doesn't include a fix for the default theme (or other problematic themes), then this PR would still be able to help new users and reviewers trying Helix tutorial.

yashpalgoyal1304: > Why not merge this PR as-is first, then once the default theme is fixed, open a new PR to include it as a recommended workaround?

i don't have any concern with the order of it, the suggestion i gave is for how it should end eventually in my views.

To reiterate, i don't have problem with what you said.

PR #5312, opened 2022-12-27
Detect truecolor support in kmscon

rdvdev2:
A patched version of kmscon for true color compatibility reports true color support by setting `COLORTERM` to `kmscon`. This PR allows helix to recognize it.

### DIFF
--- a/helix-term/src/lib.rs
+++ b/helix-term/src/lib.rs
@@ -15,7 +15,7 @@ pub use keymap::macros::*;
 #[cfg(not(windows))]
 fn true_color() -> bool {
     std::env::var("COLORTERM")
-        .map(|v| matches!(v.as_str(), "truecolor" | "24bit"))
+        .map(|v| matches!(v.as_str(), "truecolor" | "24bit" | "kmscon"))
         .unwrap_or(false)
 }
 #[cfg(windows)]


the-mikedavis:
I think this should be changed in kmscon or in your `/etc/environment` instead of in Helix.

`COLORTERM` is not formally defined anywhere as far as I can tell but the usual scheme is to test for `truecolor`. Some tools also check `24bit`:

* https://github.com/termstandard/colors/tree/2179a60be353e8739dc8cd0afa7f153ebd7e5714#checking-for-colorterm
* Neovim: https://neovim.io/doc/user/term.html#tui-colors
* Emacs: https://github.com/emacs-mirror/emacs/blob/ad5a67996ddf23df904c09165475759e2e0a68b1/doc/misc/efaq.texi#L969
* st: https://www.mail-archive.com/hackers@suckless.org/msg11307.html
* Alacritty: https://github.com/alacritty/alacritty/pull/1527
* [S-Lang](https://www.jedsoft.org/slang/) library (used by Midnight Commander): http://lists.jedsoft.org/lists/slang-users/2016/0000014.html
* Nim language runtime library: https://github.com/nim-lang/Nim/blob/82d80e6e2cb09652e15d8ec81621bba734872aaa/lib/pure/terminal.nim#L937-L938
* https://jdebp.uk/Softwares/nosh/guide/TerminalCapabilities.html

Gnome terminal is a notable exception: it [removed the COLORTERM var](https://github.com/GNOME/gnome-terminal/commit/1d5c1b6ca6373c1301494edbc9e43c3e6a9c9aaf) and it was `gnome-terminal` beforehand. Also see https://bugs.launchpad.net/ubuntu/+source/gnome-terminal/+bug/1429584 - maybe there's a better way to do the detection with terminfo? Last I saw on that was that the terminfo maintainer was [not interested](https://lists.gnu.org/archive/html/bug-ncurses/2016-08/msg00036.html) in detecting truecolor support via terminfo but that source is a little old at this point. This [gnome-terminal bug filing](https://bugs.launchpad.net/ubuntu/+source/gnome-terminal/+bug/1429584) has some interesting discussion pointing out that it could be possible to use `xtermcontrol` or checking the `VTE_VERSION` (which [we already do](https://github.com/helix-editor/helix/blob/1f4d277013783c9c537444299ba4e558b9047fe2/helix-tui/src/backend/crossterm.rs#L21) for extended underline detection).

The issue for adding this in kmscon is here: https://github.com/dvdhrm/kmscon/issues/60. I'm not sure why they chose the `kmscon` value. Using `COLORTERM` like `TERM_PROGRAM` is not ideal as any client program (like Helix) needs to add every compatible terminal to its detection list.

rdvdev2:
I will then open a PR in kmscon for this. Thank you!

PR #5320, opened 2022-12-27
Allow max number of files seen by file_picker to be configurable

### LINKED ISSUE #5157: Allow max number of files used by file_picker to be configurable

nstinus:
Fixes #5157.
Follow up on #5165.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -138,11 +138,10 @@ Valid values for these options are `block`, `bar`, `underline`, or `hidden`.
 
 ### `[editor.file-picker]` Section
 
-Sets options for file picker and global search. All but the last key listed in
-the default file-picker configuration below are IgnoreOptions: whether hidden
-files and files listed within ignore files are ignored by (not visible in) the
-helix file picker and global search. There is also one other key, `max-depth`
-available, which is not defined by default.
+Sets options for file picker and global search. Except `max-depth` and `max-
+files`, all keys listed in the default file-picker configuration below are
+IgnoreOptions: whether hidden files and files listed within ignore files are
+ignored by (not visible in) the helix file picker and global search.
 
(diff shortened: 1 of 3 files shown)


David-Else:
Would it be a good idea to warn the user that the max number of files limit had been reached? Otherwise they might be looking for a file that exists that Helix would be hiding from them due to the `max-files` limit.

At the moment the file picker opens with `100000/10000` displayed when the limit is reached, but that does not really draw attention to the limit. One of the following options could improve things:

1. The /`max-files` part could be highlighted in the warning colour to draw attention, 100000/`10000 <- in red/warning col`
2. A string saying `< max files limit reached` could be appended, `100000/10000 < max files limit reached`

nstinus:
I think both remarks make sense.

I am also not convinced at the current state of affairs: why would a git tracked repo have no limit when a plain one has this arbitrary value. Surely, if there are too many files, the fact that we are in a git repo is irrelevant. Also, the issue with this approach is that I'd rather have a slow file_picker than not finding the file I am looking for...

Perhaps an alternative could be to get rid of this value altogether and if the file_picker becomes slow, warn the user that there are a lot of files and it may be slow to use?

nstinus:
@kirawi any thoughts on this ^

pascalkuthe:
This was adreased by #7814, there is no limit anymore, files are streamed into the picker in the background now (and the matcher is fast enough to handle millions of files).

Issue #5322, opened 2022-12-28
:open doesn't expand "~" to home directory


iobtl:
# Background
Support for triggering a file picker when trying to `:open` a directory was added in #2707. But it seems like `~` is not automatically expanded by the Helix command prompt in parsing the file path. As a result, the path `~/some_directory` proceeds into the file opening branch, which then fails with the below shown status message.

https://user-images.githubusercontent.com/59901837/209760336-c67ce142-fb05-401b-9633-13cb5c088011.mov

 ---
It would be nice to implement support for this as a small QOL improvement. I can open a fix for this, assuming that there aren't any outstanding objections in prior discussions that I'm not aware of.

PR #5329, opened 2022-12-28
Expand `~` when parsing file paths in `:open`

iobtl:
Addresses #5322.

## Expected

- `:open` should work on directories with `~` in the filepath (i.e. file picker dialog opens)

https://user-images.githubusercontent.com/59901837/209839820-7887b399-07f7-4213-b98f-9d1c575012be.mov

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -88,7 +88,12 @@ impl Args {
 
 /// Parse arg into [`PathBuf`] and position.
 pub(crate) fn parse_file(s: &str) -> (PathBuf, Position) {
-    let def = || (PathBuf::from(s), Position::default());
+    let def = || {
+        (
+            helix_core::path::expand_tilde(Path::new(s)),
+            Position::default(),
+        )
+    };
     if Path::new(s).exists() {
         return def();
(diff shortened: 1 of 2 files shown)


helix-term/src/args.rs
```diff
@@ -88,7 +88,12 @@ impl Args {
 
 /// Parse arg into [`PathBuf`] and position.
 pub(crate) fn parse_file(s: &str) -> (PathBuf, Position) {
-    let def = || (PathBuf::from(s), Position::default());
+    let def = || {
+        (
+            helix_core::path::expand_tilde(Path::new(s)),
```
the-mikedavis, line 93: I think it would be better to call this from `open` in `helix-term/src/commands/typed.rs`. We don't need to do this for files passed in the argv since the shell will do the tilde expansion and there's a small bug here where `:open` fails if you give it a directory that has a tilde and a row or row/col position (though that isn't a use-case that makes sense)

iobtl: Okay, makes sense 👍

the-mikedavis (approved):

Issue #5045, opened 2022-12-07
python wrong indentation on `try:`


rumpelsepp:
### Summary

In python, the indentation is wrong when you use `try`.

![image](https://user-images.githubusercontent.com/1961699/206213626-83d9f943-4d13-4dcc-b15a-862afa81728b.png)

Line 2 should be indented by one more level.

### Reproduction Steps

Type `try:` and press enter.

### Helix log

Not relevant.

### Platform

Linux

### Terminal Emulator

gnome-console

### Helix Version

96ff64a84a4948b0aa85a453276cb0091fb9c792

rumpelsepp:
This also applies for `if FOO:`

thomkeh:
I think this is being discussed in #763

archseer:
Fixed by #5332

PR #5332, opened 2022-12-28
Improve indent queries for python

### LINKED ISSUE #763: Python file indentation is wrong

Triton171:
This is a workaround for the python tree-sitter grammar being very sensitive to incomplete code in some cases (an `try_statement` without an except/finally clause sometimes even prevents the surrounding function from being recognized). I simply added a few queries that improve the indentation in common situations. Since they only match `ERROR` nodes, this shouldn't break anything that works correctly right now.

Should fix  #763 and #5045. At least for the common scenario of writing down a `try_statement` with a missing `except_clause`/`finally_clause`, the indentation is now correct in all cases I tested,

### DIFF
--- a/runtime/queries/python/indents.scm
+++ b/runtime/queries/python/indents.scm
@@ -29,6 +29,19 @@
   (class_definition)
 ] @indent
 
+; Workaround for the tree-sitter grammar creating large errors when a
+; try_statement is missing the except/finally clause
+(ERROR
+  "try"
+  .
+  ":" @indent @extend)
+(ERROR
+  .
+  "def") @indent @extend
+(ERROR
+  (block) @indent @extend
+  (#set! "scope" "all"))
+
 [
   (if_statement)
   (for_statement)


zyklotomic:
Looks good to me! It misses the case of nested-try statements, but I understand that the PR isn't supposed to handle every case.

Triton171:
I think there's some value in keeping the patterns as generic as possible. This way, we also handle situations which we haven't specifically thought about. Since the patterns only match `ERROR` nodes, it isn't very likely to break any cases that work correctly right now. If we really want to keep the behavior as similar as possible to the previous version, we could replace the second query by something like:
````
(ERROR
  "try"
  (ERROR
    (block) @indent @extend
    (#set! "scope" "all"))
````
This would definitely not break anything but it will probably miss some situations where the grammar produces a slightly different pattern. Also, it would be very sensitive to small changes in the grammar. For example, the version of the grammar in the [playground](https://tree-sitter.github.io/tree-sitter/playground) only creates a single `ERROR` node, while in Helix there are 2 `ERROR` nodes in every case I tested. The first pattern should be fine as is, since an `ERROR` starting with the "def" keyword cannot really be anything but a function definition.

Handling nested `try_statements` is probably a bit more difficult since there's a lot more cases to think about. It's definitely something we could do in the future but I think it's best to first see if there are any issues with the handling of a single `try_statement`.

archseer (approved):
Nice, that's a pretty cool way of working around it. Thanks!

PR #5335, opened 2022-12-29
bufferline mouse action

hunterliao29:
#5274

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -9,6 +9,8 @@
 | `:buffer-close-others!`, `:bco!`, `:bcloseother!` | Force close all buffers but the currently focused one. |
 | `:buffer-close-all`, `:bca`, `:bcloseall` | Close all buffers without quitting. |
 | `:buffer-close-all!`, `:bca!`, `:bcloseall!` | Force close all buffers ignoring unsaved changes without quitting. |
+| `:buffer-close-nth`, `:bcn`, `:bclosenth` | Close nth buffer without quitting. |
+| `:buffer-nth`, `:bnth` | Goto nth buffer. |
 | `:buffer-next`, `:bn`, `:bnext` | Goto next buffer. |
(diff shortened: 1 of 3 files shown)


gibbz00:
Maybe not the best to present two features in the PR but I'm guilty of it myself 😅

Anyhow, could the buffer-nth operations be incorporated into the default buffer-next and bugger-prev? Whereby 1 is the default unless a number is specified before the motion. Similar to how one can do `2w`.

hunterliao29:
> Anyhow, could the buffer-nth operations be incorporated into the default buffer-next and bugger-prev? Whereby 1 is the default unless a number is specified before the motion. Similar to how one can do `2w`.

I'm not sure if I understanding this correctly.  Is this about 1 based index, where ```:buffer-nth 1``` should go to the first buffer?

gibbz00:
No, meaning that if I do 3 buffer next i jump to the next buffer 3 times.

hunterliao29:
What should be the behavior of  multiple ```buffer-nth 0```s?

gibbz00:
0`buffer_next` nothing?

hunterliao29:
```:buffer-nth 0``` will switch to the first open buffer, ```:buffer-nth 2``` will switch to the second if there is one and so on. It does not consider their current position like ```buffer_next```.  I am not sure if there is a use case for repeating it.

AddictArts:
I'd like to see this feature happen, but seems to be stuck here. @hunterliao29 I think you might be missing what @gibbz00 is really asking. I think from reading, helix has a `buffer_next` code path; therefore, your `buffer_nth` is really duplicating in a sense what it does. The desire looks like just improving the existing `buffer_next` instead of adding yet another way to move between buffers. Maybe I have that wrong, but looking at code it does seem like one `buffer` motion code path could do both?

pascalkuthe:
Closing this as its quite stale.The bufferline and mouse support are also not a priority.  Thank you for contributing!

Issue #4508, opened 2022-10-28
Command palette cannot create new pickers


the-mikedavis:
### Summary

Running a command from the command palette (`<space>?`) which pushes a new picker to the compositor will not create the picker.

### Reproduction Steps

`<space>?jumplist_picker<ret>` should bring up the jumplist picker but it has no effect. Running the jumplist picker command from its keybind (`<space>j`) works.

### Platform

Linux

### Terminal Emulator

Kitty 0.26.2

### Helix Version

26f21da531179ccc77e70b73e32a1b458c4e225b

nrabulinski:
Has this been picked up yet? If not I can get on it as I identified what the issue is

woojiq:
What I dug up: 

Actually, async pickers even work from the `command palette`. The reason is that they use `callback` __method__ that internally add a picker to the application's task list:
https://github.com/helix-editor/helix/blob/5ec126d3e27dcfe2544396b664cbad04611ad7eb/helix-term/src/commands/lsp.rs#L1155

Sync pickers, on the other hand, use the `push_layer` method, which internally populates the `callback` __field__:
https://github.com/helix-editor/helix/blob/5ec126d3e27dcfe2544396b664cbad04611ad7eb/helix-term/src/commands.rs#L95-L96

The interesting part about this field is that it only resolves in one place - `EditorView::handle_event`:
https://github.com/helix-editor/helix/blob/5ec126d3e27dcfe2544396b664cbad04611ad7eb/helix-term/src/ui/editor.rs#L1332
So when you populate this field while in the `command palette`, it will be lost and incomplete just because the `Picker` has a different `handle_event` implementation:
https://github.com/helix-editor/helix/blob/5ec126d3e27dcfe2544396b664cbad04611ad7eb/helix-term/src/ui/picker.rs#L781
So far I have only one idea to solve this: inside the `push_layer` make the callback asynchronous and add it to the application's task list (same as the async pickers do). Otherwise, anything that uses `push_layer` (file, jumplist, buffer pickers...) will only work if you start them from `EditorView` (the first layer).

P.S. I wrote this so tomorrow I will recall what I was thinking, xd. Maybe someone will be interested.

the-mikedavis:
[...]

PR #5340, opened 2022-12-29
Enhanced jump mode (based on #3791)

semin-park:
## What does this PR do?
~~Took #3791, rebased on top of master and modified the jump logic for word based jumps.~~
~~Now it's based on #5420.~~
Now that #5420 is merged, this is rebased on top of master :D

Disclaimer: the logic was heavily influenced by [hop.nvim](https://github.com/phaazon/hop.nvim) and the visuals (RGB values) were directly copied.

## Improvements
The improvements are:

1. #3791 takes a set of characters (to be used in jumping), calculates an [arbitrary cutoff point](https://github.com/helix-editor/helix/pull/3791/files#diff-d782c07f22492bcae17333aba5dc5b840fb30317727e71267ebf4863f84787fdR5070-R5078) and then uses `keys[..cutoff]` for single-key-press jumps and the remaining `keys[cutoff..]` for multi-key-press jumps. This logic works, but it produces too many unnecessary 3 character jumps even when we don't need them. The formula for how many <= 2 key-press-jumps we can generate with this logic is: `keys[..cutoff].len() + keys[cutoff..].len().pow(2)`. Assuming 26 characters and cutoff of 10, the result is `10 + (26-10)^2 = 266` jump locations. Any more than this will end up with some jumps requiring three key presses or more.  
 In this PR, I'm starting with all one character jumps and expand the trie only when necessary. This means that we can always produce the most optimized character sequences, and thus the arbitrary cutoff point doesn't have to be calculated anymore. Now the number of <= 2 key-press jumps we can represent is simply `26^2 = 676`. Practically, I have never had to deal with any three character sequences, since usually the number of jump locations on my screen is less than 400.

2. The previous logic used `movement::move_prev_word_start` to get the word boundary, but it didn't check what kind of character is underneath the cursor. This meant that many jump locations were too close to each other, so there used to be a logic that checks whether the current jump location is [too close to the previous jump location](https://github.com/helix-editor/helix/pull/3791/files#diff-d782c07f22492bcae17333aba5dc5b840fb30317727e71267ebf4863f84787fdR4939-R4945), and abandons the current jump location if it is too close.
In this PR, I'm reducing the number of jump locations by only generating jump targets for words that satisfy `char_is_word` (i.e., alphanumeric + `_`). Usually the place we want to edit is an identifier (function name, parameter name etc), so I feel this approach is more ergonomic.

3. Some visual enhancements, such as dimming the screen so that the jump locations can be seen more easily, and using slightly different colors in case of a multi character sequences, so that the first key can be seen more easily. This UI part is just reproducing what hop.nvim does in helix.

## How to proceed
I know that virtual text support is still WIP and this PR is going to become stale. But I wanted to put this up here just so that it could be incorporated in the future work.

With regards to how to work further on this, I'm completely open to whatever works. I can take this further after #5008 gets merged; or if @Omnikar wants to push this through, I am happy with that too after we talk this through.

### Normal use case

https://user-images.githubusercontent.com/30904140/210132132-5d114b33-fd80-4ab4-bf23-bf5950102239.mp4

### Growing selection

https://user-images.githubusercontent.com/30904140/210132227-8007c9a1-3a3c-4087-bf75-286cc04c1b27.mp4

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -223,6 +223,8 @@ Jumps to various locations.
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
 | `j`   | Move down textual (instead of visual) line       | `move_line_down`           |
 | `k`   | Move up textual (instead of visual) line         | `move_line_up`             |
+| `w`   | Word-wise jump mode                              | `jump_to_identifier_label` |
+| `/`   | Character search jump mode                       | `jump_to_char_label`       |
 
 #### Match mode
 
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -224,7 +224,7 @@ Jumps to various locations.
 | `j`   | Move down textual (instead of visual) line       | `move_line_down`           |
 | `k`   | Move up textual (instead of visual) line         | `move_line_up`             |
 | `w`   | Word-wise jump mode                              | `jump_to_identifier_label` |
-| `/`   | Character search jump mode                       | `jump_to_char_label`       |
+| `/`   | Character or string search jump mode             | `jump_to_str_label`       |
 
 #### Match mode
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -50,8 +50,8 @@ use insert::*;
 use movement::Movement;
 
 use self::jump::{
-    cleanup, find_all_identifiers_in_view, find_all_str_occurrences_in_view,
-    show_key_annotations_with_callback, sort_jump_targets, JumpSequencer, TrieNode, JUMP_KEYS,
+    cleanup, find_all_identifiers_in_view, find_all_str_occurrences_in_view, jump_keys,
+    show_key_annotations_with_callback, sort_jump_targets, JumpSequencer, TrieNode,
 };
 use crate::{
     args,
@@ -5669,7 +5669,7 @@ fn jump_with_targets(cx: &mut Context, mut jump_targets: Vec<Range>, extend_sele
         jump_to(cx, jump_targets[0], extend_selection);
         return cleanup(cx);
     }
-    let root = TrieNode::build(JUMP_KEYS, jump_targets);
+    let root = TrieNode::build(&jump_keys(cx), jump_targets);
     show_key_annotations_with_callback(cx, root.generate(), move |cx, event| {
         handle_key_event(root, cx, event, extend_selection)
     });
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1,4 +1,5 @@
 pub(crate) mod dap;
+pub(crate) mod jump;
 pub(crate) mod lsp;
 pub(crate) mod typed;
 
@@ -48,6 +49,10 @@ use fuzzy_matcher::FuzzyMatcher;
 use insert::*;
 use movement::Movement;
 
+use self::jump::{
+    cleanup, find_all_identifiers_in_view, find_all_str_occurrences_in_view,
+    show_key_annotations_with_callback, sort_jump_targets, JumpSequencer, TrieNode, JUMP_KEYS,
+};
 use crate::{
     args,
     compositor::{self, Component, Compositor},
@@ -62,7 +67,7 @@ use crate::{
 
 use crate::job::{self, Jobs};
 use futures_util::{stream::FuturesUnordered, StreamExt, TryStreamExt};
-use std::{collections::HashMap, fmt, future::Future};
+use std::{cmp::Ordering, collections::HashMap, fmt, future::Future};
 use std::{collections::HashSet, num::NonZeroUsize};
 
 use std::{
@@ -485,6 +490,10 @@ impl MappableCommand {
         decrement, "Decrement item under cursor",
         record_macro, "Record macro",
         replay_macro, "Replay macro",
+        jump_to_identifier_label, "Jump mode: word-wise",
+        jump_to_str_label, "Jump mode: character search",
+        jump_to_identifier_label_and_extend_selection, "Jump mode: extend selection with word-wise jump",
+        jump_to_str_label_and_extend_selection, "Jump mode: extend selection with character search",
         command_palette, "Open command palette",
     );
 }
@@ -5587,4 +5596,179 @@ fn replay_macro(cx: &mut Context) {
         // replaying recursively.
         cx.editor.macro_replaying.pop();
     }));
+}
+
+fn jump_to_identifier_label(cx: &mut Context) {
+    let jump_targets = find_all_identifiers_in_view(cx);
+    jump_with_targets(cx, jump_targets, false);
+}
+
+fn jump_to_identifier_label_and_extend_selection(cx: &mut Context) {
+    let jump_targets = find_all_identifiers_in_view(cx);
+    jump_with_targets(cx, jump_targets, true);
+}
+
+fn jump_to_str_label(cx: &mut Context) {
+    find_and_jump_to_str(cx, false);
+}
+
+fn jump_to_str_label_and_extend_selection(cx: &mut Context) {
+    find_and_jump_to_str(cx, true);
+}
+
+fn find_and_jump_to_str(cx: &mut Context, extend_selection: bool) {
+    jump::setup(cx);
+    let nchars = doc!(cx.editor)
+        .config
+        .load()
+        .jump_mode
+        .num_chars_before_label as usize;
+    cx.editor.set_status(format!("Press {} key(s)", nchars));
+    cx.on_next_key(move |cx, event| {
+        jump_with_str_input(
+            cx,
+            event,
+            extend_selection,
+            String::with_capacity(nchars),
+            nchars,
+        )
+    });
+}
+
+fn jump_with_str_input(
+    cx: &mut Context,
+    event: KeyEvent,
+    extend_selection: bool,
+    mut s: String,
+    nchars: usize,
+) {
+    let Some(c) = event.char().filter(|c| c.is_ascii()) else {
+        return cleanup(cx);
+    };
+    s.push(c);
+    if s.len() == nchars {
+        let jump_targets = find_all_str_occurrences_in_view(cx, s);
+        jump_with_targets(cx, jump_targets, extend_selection);
+        return;
+    }
+    cx.editor
+        .set_status(format!("Press {} more key(s)", nchars - s.len()));
+    cx.on_next_key(move |cx, event| jump_with_str_input(cx, event, extend_selection, s, nchars));
+}
+
+fn jump_with_targets(cx: &mut Context, mut jump_targets: Vec<Range>, extend_selection: bool) {
+    if jump_targets.is_empty() {
+        return cleanup(cx);
+    }
+    // Jump targets are sorted based on their distance to the current cursor.
+    jump_targets = sort_jump_targets(cx, jump_targets);
+    if extend_selection {
+        jump_targets = extend_jump_selection(cx, jump_targets);
+    }
+    if jump_targets.len() == 1 {
+        jump_to(cx, jump_targets[0], extend_selection);
+        return cleanup(cx);
+    }
+    let root = TrieNode::build(JUMP_KEYS, jump_targets);
+    show_key_annotations_with_callback(cx, root.generate(), move |cx, event| {
+        handle_key_event(root, cx, event, extend_selection)
+    });
+}
+
+fn fix_extend_mode_off_by_one(selection: &Range, target: &mut Range) {
+    // Non-zero width ranges are always inclusive on the left and exclusive on
+    // the right. But when we're extending the selection, this often creates
+    // off-by-one behavior, where the cursor doesn't quite reach the target.
+    // Thus we need to increment the upper bound when the target head is after
+    // the current anchor.
+    if selection.anchor < target.head {
+        match target.anchor.cmp(&target.head) {
+            Ordering::Less => target.head += 1,
+            Ordering::Greater => target.anchor += 1,
+            Ordering::Equal => {}
+        };
+    }
+}
+
+fn jump_to(cx: &mut Context, mut range: Range, extend_selection: bool) {
+    let (view, doc) = current!(cx.editor);
+    push_jump(view, doc);
+    let selection = doc.selection(view.id).primary();
+    if extend_selection {
+        fix_extend_mode_off_by_one(&selection, &mut range);
+    }
+    doc.set_selection(view.id, Selection::single(range.anchor, range.head));
+}
+
+/// Handle user key press.
+/// Returns whether we are finished and should move out of jump mode.
+fn handle_key(
+    mut sequencer: impl JumpSequencer + 'static,
+    cx: &mut Context,
+    key: u8,
+    extend_selection: bool,
+) -> bool {
+    cleanup(cx);
+    match sequencer.choose(key) {
+        Some(subnode) => {
+            sequencer = *subnode;
+        }
+        // char `c` is not a valid character. Finish jump mode
+        None => return true,
+    }
+    match sequencer.try_get_range() {
+        Some(range) => {
+            jump_to(cx, range, extend_selection);
+            return true;
+        }
+        None => {
+            show_key_annotations_with_callback(cx, sequencer.generate(), move |cx, event| {
+                handle_key_event(sequencer, cx, event, extend_selection)
+            });
+        }
+    }
+    false
+}
+
+fn handle_key_event(
+    node: impl JumpSequencer + 'static,
+    cx: &mut Context,
+    event: KeyEvent,
+    extend_selection: bool,
+) {
+    let finished = match event.char() {
+        Some(key) => {
+            if event.modifiers.is_empty() && key.is_ascii() {
+                handle_key(node, cx, key as u8, extend_selection)
+            } else {
+                // Only accept ascii characters with no modifiers. Otherwise, finish jump mode.
+                true
+            }
+        }
+        // We didn't get a valid character. Finish jump mode.
+        None => true,
+    };
+    if finished {
+        cleanup(cx);
+    }
+}
+
+fn extend_jump_selection(cx: &Context, jump_targets: Vec<Range>) -> Vec<Range> {
+    let (view, doc) = current_ref!(cx.editor);
+    // We only care about the primary selection
+    let mut cur = doc.selection(view.id).primary();
+    jump_targets
+        .into_iter()
+        .map(|mut range| {
+            // We want to grow the selection, so if the new head crosses the
+            // old anchor, swap the old head and old anchor
+            let cross_fwd = cur.head < cur.anchor && cur.anchor < range.head;
+            let cross_bwd = range.head < cur.anchor && cur.anchor < cur.head;
+            if cross_fwd || cross_bwd {
+                std::mem::swap(&mut cur.head, &mut cur.anchor);
+            }
+            range.anchor = cur.anchor;
+            range
+        })
+        .collect()
 }
--- a/helix-term/src/commands/jump/annotate.rs
+++ b/helix-term/src/commands/jump/annotate.rs
@@ -1,0 +1,78 @@
+use super::JumpAnnotation;
+use crate::commands::Context;
+use helix_core::chars::char_is_line_ending;
+use helix_core::text_annotations::Overlay;
+use helix_view::{input::KeyEvent, View};
+use std::rc::Rc;
+
+pub const JUMP_KEYS: &[u8] = b"etovxqpdygfblzhckisuran";
+
+#[inline]
+pub fn setup(ctx: &mut Context) {
+    let (view, doc) = current!(ctx.editor);
+    if doc.config.load().jump_mode.dim_during_jump {
+        view.dimmed = true;
+    }
+    view.in_visual_jump_mode = true;
+}
+
+#[inline]
+fn clear_dimming(view: &mut View) {
+    view.dimmed = false;
+    view.in_visual_jump_mode = false;
+}
+
+#[inline]
+pub fn cleanup(ctx: &mut Context) {
+    let mut view = view_mut!(ctx.editor);
+    clear_dimming(view);
+    view.visual_jump_labels[0] = Rc::new([]);
+    view.visual_jump_labels[1] = Rc::new([]);
+    view.visual_jump_labels[2] = Rc::new([]);
+}
+
+/// `annotations` should already be sorted by the `loc` attribute (= char_idx)
+pub fn show_key_annotations_with_callback<F>(
+    ctx: &mut Context,
+    annotations: Vec<JumpAnnotation>,
+    on_key_press_callback: F,
+) where
+    F: FnOnce(&mut Context, KeyEvent) + 'static,
+{
+    setup(ctx);
+    let (view, doc) = current!(ctx.editor);
+    let text = doc.text().slice(..);
+    let mut overlays_single: Vec<Overlay> = Vec::new();
+    let mut overlays_multi_first: Vec<Overlay> = Vec::new();
+    let mut overlays_multi_rest: Vec<Overlay> = Vec::new();
+    for jump in annotations.into_iter() {
+        if jump.keys.len() == 1 {
+            overlays_single.push(Overlay {
+                char_idx: jump.loc,
+                grapheme: jump.keys.into(),
+            });
+            continue;
+        }
+        overlays_multi_first.push(Overlay {
+            char_idx: jump.loc,
+            grapheme: jump.keys.chars().next().unwrap().to_string().into(),
+        });
+        for (i, c) in (1..jump.keys.len()).zip(jump.keys.chars().skip(1)) {
+            let char_idx = jump.loc + i;
+            let char = text.chars_at(char_idx).next().unwrap();
+            // We shouldn't overlay anything on top of a line break. If we do, the next line will
+            // crawl up and concatenate with the current line.
+            if char_is_line_ending(char) {
+                break;
+            }
+            overlays_multi_rest.push(Overlay {
+                char_idx,
+                grapheme: c.to_string().into(),
+            });
+        }
+    }
+    view.visual_jump_labels[0] = overlays_single.into();
+    view.visual_jump_labels[1] = overlays_multi_first.into();
+    view.visual_jump_labels[2] = overlays_multi_rest.into();
+    ctx.on_next_key(on_key_press_callback);
+}
--- a/helix-term/src/commands/jump/annotate.rs
+++ b/helix-term/src/commands/jump/annotate.rs
@@ -5,7 +5,15 @@ use helix_core::text_annotations::Overlay;
 use helix_view::{input::KeyEvent, View};
 use std::rc::Rc;
 
-pub const JUMP_KEYS: &[u8] = b"etovxqpdygfblzhckisuran";
+pub fn jump_keys(ctx: &mut Context) -> Vec<u8> {
+    doc!(ctx.editor)
+        .config
+        .load()
+        .jump_mode
+        .jump_keys
+        .clone()
+        .into_bytes()
+}
 
 #[inline]
 pub fn setup(ctx: &mut Context) {
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -19,8 +19,7 @@ use helix_core::{
     syntax::{self, Highlight, HighlightEvent},
     text_annotations::TextAnnotations,
     unicode::width::UnicodeWidthStr,
-    visual_offset_from_block, Change, Position, Range, RopeSlice, Selection,
-    Transaction,
+    visual_offset_from_block, Change, Position, Range, RopeSlice, Selection, Transaction,
 };
 use helix_view::{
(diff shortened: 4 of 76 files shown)


helix-term/src/commands.rs
```diff
@@ -5209,3 +5214,124 @@ fn replay_macro(cx: &mut Context) {
...
+                    backward_jump_locations.push(pos);
+                }
+                _ => break,
+            }
+        }
+
+        let jump_locations: Vec<_> = forward_jump_locations
+            .into_iter()
+            .zip(backward_jump_locations.into_iter())
+            .flat_map(|(a, b)| vec![a, b])
```
Omnikar, line 5284: The `into_iter` and vec creation here shouldn't be necessary.
```suggestion
            .zip(backward_jump_locations)
            .flat_map(|(a, b)| [a, b])
```

Omnikar:
These are some nice improvements, I'll review when I get a chance.

helix-term/src/commands.rs
```diff
@@ -5209,3 +5214,124 @@ fn replay_macro(cx: &mut Context) {
...
+                    }
+                    backward_jump_locations.push(pos);
+                }
+                _ => break,
+            }
+        }
+
+        let jump_locations: Vec<_> = forward_jump_locations
+            .into_iter()
+            .zip(backward_jump_locations.into_iter())
```
Omnikar, line 5283: Zipping together the forwards and backwards jump locations causes truncation of whichever group is longer. I don't know if this is the most optimal way to solve it, but for reference, [here is how I combated this issue in #3791](https://github.com/helix-editor/helix/pull/3791/files#diff-d782c07f22492bcae17333aba5dc5b840fb30317727e71267ebf4863f84787fdR5048-R5061).

semin-park: Ah I see, that was a last minute change I've made, but I think I'll change this part too to scan from the beginning till the end and score it based on the distance to the current cursor. I'll ping again once I find time to fix this.

semin-park:
> These are some nice improvements, I'll do a full review when I get a chance.
> 
> From a quick test it looks like it's missing the selection behaviours from #3791; i.e. extend mode support and selecting the jumped-to word when using word jump.

Oh I totally missed this behavior, I'll add them back in. Thanks for the comment, much appreciated!

helix-term/src/commands.rs
semin-park: Should be fixed now, thanks!

semin-park:
Fixed the comments and added selection extending behavior for word based jumps as well.

I tried to reorganize the code into:

1. `sequence.rs` generates the key sequences for jump targets. It doesn't know anything about the screen or the document that we're dealing with.
2. `locations.rs` generates the jump targets by reading the buffer in the current screen.
3. `score.rs` sorts the jump targets based on some scoring function.
4. `annotate.rs` has helper functions such as generating the actual text annotations understood by `helix-view`, dimming and clearing the view.

I tried to focus on readability, hope it's not too hard to read 😆

otavioschwanck (approved):

dahmc:
Looks great. Would it be possible to have the dimming as an optional setting? It's personal preference, and I myself find that very distracting while working, and prefer just highlights with a background color.

helix-term/src/commands/jump/annotate.rs
```diff
@@ -0,0 +1,73 @@
+use super::JumpAnnotation;
+use crate::commands::Context;
+use helix_core::text_annotations::Overlay;
+use helix_view::input::KeyEvent;
+use std::rc::Rc;
+
+pub const JUMP_KEYS: &[u8] = b"etovxqpdygfblzhckisuran";
+
+pub fn apply_dimming(_ctx: &mut Context) {
+    // let (view, doc) = current!(ctx.editor);
```
semin-park: These unused comments are gone in the next commit. I was a bit lazy and didn't remove it here due to merge conflicts..

helix-term/src/commands/jump/annotate.rs
```diff
@@ -0,0 +1,73 @@
...
+    //         style: Style::default().fg(Color::Rgb(0x66, 0x66, 0x66)),
+    //         line: idx,
+    //         kind: TextAnnotationKind::Overlay(0),
+    //     })
+    //     .collect();
+    // doc.push_text_annotations("jump_mode", lines.into_iter());
+}
+
+pub fn clear_dimming(ctx: &mut Context) {
+    // doc_mut!(ctx.editor).clear_text_annotations("jump_mode");
```
semin-park: ditto

helix-term/src/commands/jump/annotate.rs
```diff
@@ -0,0 +1,73 @@
...
+
+/// `annotations` should already be sorted by the `loc` attribute (= char_idx)
+pub fn show_key_annotations_with_callback<F>(
+    ctx: &mut Context,
+    annotations: Vec<JumpAnnotation>,
+    on_key_press_callback: F,
+) where
+    F: FnOnce(&mut Context, KeyEvent) + 'static,
+{
+    // apply_dimming(ctx);
```
semin-park: ditto

semin-park:
Since virtual text support is going to come from #5420, I modified the annotation logic to use #5420 instead of #417.
The diffs are:
- Use #5420 for virtual text support (annotation logic changed)
- Penalize targets that are further away in the y-axis, in order to prefer targets closer to the current row
- Color for jump characters are now configurable. The added theme parameters are `ui.virtual.jump.single`, `ui.virtual.jump.multi.first` and `ui.virtual.jump.multi.rest`. These parameters can be assigned different colors in the theme toml file.
- Took @dahmc's suggestion and made dimming optional. Jump mode will still dim the view by default, but this behavior can be overridden with
```
// config.toml
[editor]
dim-during-jump = false
```

Jump mode seems to be a bit unstable in the presence of other virtual text, so I'll debug further and report back.

QuinnFreedman:
Is this currently working? I tried it out, but when I try to enter jump mode it hangs the whole editor. There is no error log and eventually I have to just close the terminal window.

semin-park:
Huh that's odd, could you tell me your configuration and your setup, so that I can try to reproduce?
I am using this daily and I didn't mean to break anything..

QuinnFreedman:
After doing a little debugging, it looks like the issue was from the theme I was using (the built-in monokai pro theme). I installed this fork over the top of my main branch helix install, so I'm not sure which version the theme came from. I would guess it's just missing the colors for jump mode (maybe we could fallback to defaults or something in that case)

Here is the offending theme if you are interested: [monokai_pro.zip](https://github.com/helix-editor/helix/files/10474091/monokai_pro.zip)

QuinnFreedman:
Also, a very minor bug: it seems like word-wise jump ignores the last word in the document if there is no trailing newline

pascalkuthe:
That was a known issue with #5420. Debasing on the newest changes will fix the problem

semin-park:
Ah I see, thanks for the debugging :)
I rebased this PR on top of #5420 so hopefully this issue is gone now :)
But one thing that I noticed in your theme file is that it's lacking these values:

```
"ui.virtual.jump.single" = { fg = "#ff007c", modifiers = ["bold"] }
"ui.virtual.jump.multi.first" = { fg = "#00dfff", modifiers = ["bold"] }
"ui.virtual.jump.multi.rest" = "#2b8db3"
```

I added them into `helix/runtime/themes/monokai_pro.toml` in this PR, so you might want to use that if you want your jump labels to be colored correctly

QuinnFreedman:
Thanks! that worked. Everything seems to be working as expected now

QuinnFreedman:
On the subject of themes, would you consider adding a customizable style for the background text in jump mode? (Or does that already exist?) It seems like it's mapped to the comment style which looks a little weird since my comments are italic

QuinnFreedman:
One other quick bug: in `sort_jump_targets`, `cursor` is in screen coordinates but `positions` are in document coordinates. This means that if you jump from deeper in a document, all the single-character jump tags will be at the top of the screen. 

I'm happy to make a PR to address this if you want. Should be a one-line fix

semin-park:
> One other quick bug: in `sort_jump_targets`, `cursor` is in screen coordinates but `positions` are in document coordinates. This means that if you jump from deeper in a document, all the single-character jump tags will be at the top of the screen.
> 
> I'm happy to make a PR to address this if you want. Should be a one-line fix

Oh thanks for the bug report! The bug must have sneaked in when I last pushed my commits. Fixed it and pushed again :)



> On the subject of themes, would you consider adding a customizable style for the background text in jump mode? (Or does that already exist?) It seems like it's mapped to the comment style which looks a little weird since my comments are italic

The background is currently set to the comment style, but I can make it configurable that falls back to the comment style if no applicable user config is found. I don't have time right now, but I can try changing this tonight 😄

semin-park:
Added `ui.virtual.jump.dim` and set it to `#666666` by default for all available helix themes.

QuinnFreedman:
Thanks for your work on this! I can't wait for it to get merged into core

jlebon:
This is great and works really well! I have a suggestion, but feel free to defer to later: is it possible to have Ctrl-c also work as a way to cancel jump mode after entering it (i.e. make it work just like Esc) ? This makes it similar to how Ctrl-c today can be used to exit most popup menus as an alternative to Esc. Currently, Ctrl-c is just treated as if "c" were typed which feels confusing.

thomasgoulet:
Nice PR! Would love to have this be merged

Just wanted to mention that triggering the `Jump mode` commands from inside the command palette renders the virtual text but doesn't actually trigger the command / cancels it right away. This behavior also seems to affect the `find_next_char`, `find_prev_char`, etc... commands but they don't render any virtual text.

Only way to get the correct syntax highlight back is to run any of the `Jump commands` from a keybind or close the buffer.

semin-park:
> This is great and works really well! I have a suggestion, but feel free to defer to later: is it possible to have Ctrl-c also work as a way to cancel jump mode after entering it (i.e. make it work just like Esc) ? This makes it similar to how Ctrl-c today can be used to exit most popup menus as an alternative to Esc. Currently, Ctrl-c is just treated as if "c" were typed which feels confusing.

Thanks for the suggestion! I made jump mode only recognize characters when no modifier has been added to the input keys, and I think this should be reasonable for now, given that jump mode only works with ascii lowercase characters.

I personally hope that helix could alias Ctrl-c to ESC, so filed #5739 to ask about this.

semin-park:
> Nice PR! Would love to have this be merged
> 
> Just wanted to mention that triggering the `Jump mode` commands from inside the command palette renders the virtual text but doesn't actually trigger the command / cancels it right away. This behavior also seems to affect the `find_next_char`, `find_prev_char`, etc... commands but they don't render any virtual text.
> 
> Only way to get the correct syntax highlight back is to run any of the `Jump commands` from a keybind or close the buffer.

It looks like command palette creates a temporary Context when executing commands, so the context that jump mode sees is different from when it was launched directly from key binding. Asked about this in the matrix chat and was told that there are known bugs due to this (#4508, #5294) and works to address this (#5555).

But as of now, I think you will have to use the key bindings :'(

pascalkuthe:
With #5420 merged this can ne debased on master.

On a more personal note: Have you considered replacing the single char command with a double char command. That will require loooking at labels much less often. Especially if you make it directional and immidietly jump like leap

helix-term/src/ui/editor.rs
```diff
@@ -171,8 +187,19 @@ impl EditorView {
             } else {
                 Box::new(syntax::merge(highlights, focused_view_elements))
             }
+        };
+
+        let highlights = if view.in_visual_jump_mode {
```
semin-park, line 192: We already have a concept of "jump", which is different from the jump mode that this PR is implementing. So to avoid confusion, maybe it's better to rename it to something like hop mode? (this PR is heavily influenced by hop.nvim anyways..)

Omnikar:
> On a more personal note: Have you considered replacing the single char command with a double char command. That will require loooking at labels much less often. Especially if you make it directional and immidietly jump like leap

My personal use case of jump mode is to get to where I want to get to quickly, with minimal thought. Adding directionality to jump mode adds an extra thing you have to think about. Also, directional multichar jumping is starting to sound a lot like ordinary `/` `?` searching.

pascalkuthe:
> > On a more personal note: Have you considered replacing the single char command with a double char command. That will require loooking at labels much less often. Especially if you make it directional and immidietly jump like leap
> 
> My personal use case of jump mode is to get to where I want to get to quickly, with minimal thought. Adding directionality to jump mode adds an extra thing you have to think about. Also, directional multichar jumping is starting to sound a lot like ordinary `/` `?` searching.

For the usecase of getting where I am going with minimal.thought I love the word jump. I use it to just quickly get to the rough area where I want to edit.

The single char jump is too noisy for me. It places labels everywhere because many vhars appear quite often. The usecase for single/double char jump for me are cases where zi need a more precise jump where word boundaries don't cut it.

With two chars there are much less ambiguities and so often you end up at the right place right away or only habe a few labels to chose from. The directionality isn't that critical to that usecase. I tend to think of it of a much more accurate f. The case of I want to use f but there is a char in the wayoccurs so often to me and a command like I described solves it neatly. This usecase of a precise f would work especially well if the motion extends the selection (but I can also see that annoying pwople, maybe having a separate extend and goto variant of the commands could work)

semin-park:
Rebased on top of master, and made the number of characters to type before labelling configurable. The default value is still 1 just because I think that's the base case and I don't have a strong opinion on what should be the default, but you can override it like this:

```
[editor.jump-mode]
num-chars-before-label = 2
```

pascalkuthe:
> > I think there will be many more "jump modes" in the future, other than WORD an N-CHAR, and the suggested keybinding is not fast enough for most people. Personally I will probably go for a single char activation, like the Enter key. (I have a Planck keyboard, the Enter key is under my right thumb, so for me Enter is FAST)
> > There's no a strong default solution, so maybe this feature should not deserve default keybindings.
> 
> I'm okay with removing the default key binding for jump mode, as long as it can be well communicated to the end users that such a thing exists. But I'm not quite sure what the best way is. If these motion plugins are used by the majority of people, I think having a default command in the default keymap is worth it for visibility (and they can look up what other commands / options are available).
> 
> But if not, I'm also okay with removing the default key bindings and adding a section in `:tutor` instead.

I think the default bindings you came up with are nice. I like how intutive `g/` and `gw` feel. They tie in with the existing `w` and `/` binding and are therefore easy to remeber and discover which is what we usually optimize for. If they are not short enough/not palced centrally enough for somebody then they can rebind in their personal config.

semin-park:
Oh and I think it's worth mentioning that the current m-char search algorithm is brute force. It iterates all indices within the current view and tries to match the m-char string starting at each index. So the algorithm's worst-case complexity is `O(mn)`. Under normal circumstances (= close to random characters), it should still be close to `O(n)`, and even if not, `m` should be very small anyway. So practically this is still fast enough.

But if this needs to be further optimized, we could implement the [KMP algorithm](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm), although I would prefer to leave it as a low-priority future work.

pascalkuthe:
> Oh and I think it's worth mentioning that the current m-char search algorithm is brute force. It iterates all indices within the current view and tries to match the m-char string starting at each index. So the algorithm's worst-case complexity is `O(mn)`. Under normal circumstances (= close to random characters), it should still be close to `O(n)`, and even if not, `m` should be very small anyway. So practically this is still fast enough.
> 
> But if this needs to be further optimized, we could implement the [KMP algorithm](https://en.wikipedia.org/wiki/Knuth%E2%80%93Morris%E2%80%93Pratt_algorithm), although I would prefer to leave it as a low-priority future work.

O(mn) is fine in this case. I doubt everyone sets `m` above 3 and `O(3n)` is basically `O(n)` :D

cd-a:
Thanks for adding the num-chars-before-label, since I much prefer leap I find it not usable with just 1 char. But 2 works much better.

Question about the characters, are those configurable? I would assume the nearest matches should have the easiest to reach characters.

Take this example: 
CleanShot 2023-02-02 at 18 57 36@2x

Shouldn't the first matches be the one on the homerow (asdf jkl etc)?

I find it weird to have hard to reach chars like v and x up there already. 

Then for layouts like Dvorak and Colemak those could be amended.

semin-park:
Oh yeah that makes sense. I copied the keys from hop but there could be better way to order them.
Added a `jump-keys` config parameter that you can override:

```
[editor.jump-mode]
jump-keys = "<your sequence>"
```

cd-a:
> Oh yeah that makes sense. I copied the keys from hop but there could be better ways to order them. Added a `jump-keys` config parameter that you can override:

Great, thanks! Just tested it and it works like a charm!

pascalkuthe:
> Oh yeah that makes sense. I copied the keys from hop but there could be better ways to order them. Added a `jump-keys` config parameter that you can override:
> 
> ```
> [editor.jump-mode]
> jump-keys = "<your sequence>"  // default is "jwetovxqpdygfblzhckisuranm"
> ```

Thanks for adding these new config options. You should add some documentation for them in the book (`book/src/configuration.md`)

semin-park:
Thanks for the suggestion! Added it in 😁

evanrichter:
I'm now testing this, together with #5934, in my fork here: https://github.com/evanrichter/helix

so far the two features work very well together!

evanrichter:
@semin-park I'm not able to see the theme additions to jump mode anymore, but it was working previously.

- Using your branch directly (no merges)
- build/run with `nix run .`
- activate jump mode with `gw`
- text is dimmed and jump mode is activated, but the jump hint letters are not highlighted

Even with `:set-option jump-mode.dim-during-jump false`, the "ui.virtual.jump.*" colors do not show up

semin-park:
Thanks for the bug report @evanrichter! Somehow I thought the previous conflict was a minor one and didn't recompile.. it should work again now, could you please try out again?

Sorry for the inconvenience and thanks for trying it out :)

semin-park:
Friendly ping for review :)

evanrichter:
> Friendly ping for review :)

I don't have time to review right now, but I can confirm your fork works as intended

cd-a:
Tested it again, looks and works great. Love it.

Thanks again for the hard work on this!

yerlaser:
As far as I know this PR needs an approving review by one of the core team members.
Unfortunately, they are all very busy, it seems.

quantum-booty:
It would be convenient if it could have the following optional features:
1. directional search
2. if directional search, then automatically jump to the first match
3. case insensitive search

evanrichter:
along the lines of @quantum-booty making suggestions, I also would like an option to never jump within 25 or so words. When I enter jump-mode, I'm usually intending to jump farther. For smaller jumps, `t` or something like `3jwww` is less mental overhead.

just a thought though. I think options like that could be added as separate PRs

cd-a:
Personally I don't think directional search is necessary. I remember in one of the vim plugins there was a discussion, and after testing it I had to agree, that it's much faster to do a non-directional search.

quantum-booty:
> Personally I don't think directional search is necessary. I remember in one of the vim plugins there was a discussion, and after testing it I had to agree, that it's much faster to do a non-directional search. 
> 
> 

I think directional search + autojump to first match is a powerful combination. First of all, directional search is able to half the number of candidates to jump to, as a result the prompt characters are most likely a single character from the lowercase alphabets. This reduces mental overhead of typing the prompt characters.
Second, directional search enables autojump to first matching word, which in a lot of the cases, is the word I want to jump to. This feature is not possible with bidirectional search, because the difinition of "first match" is ambiguous.

quantum-booty:
> I would be fine with any of those features being added as options, but I have been using this fork for months and it already covers most of my use cases. FWIW I would advocate for getting this merged to main ASAP and adding additional options as separate features down the line

tbh as much as I like these features to be added, I agree with this.

helix-term/src/commands.rs
```diff
@@ -5371,3 +5380,181 @@ fn replay_macro(cx: &mut Context) {
...
+    extend_selection: bool,
+    mut s: String,
+    nchars: usize,
+) {
+    let c = event.char().filter(|c| c.is_ascii());
+    if c.is_none() {
+        cleanup(cx);
+        return;
+    }
+    let c = c.unwrap();
```
casperstorm, line 5433:
```suggestion
    let Some(c) = event.char().filter(|c| c.is_ascii()) else {
        cleanup(cx);
        return
    };
```

You can use `let-else` here to simplify this a lot.

helix-term/src/ui/editor.rs
```diff
@@ -341,6 +356,35 @@ impl EditorView {
...
+        height: u16,
+        theme: &Theme,
+    ) -> Box<dyn Iterator<Item = HighlightEvent>> {
+        let highlight = theme
+            .find_scope_index("ui.virtual.jump.dim")
+            .or_else(|| theme.find_scope_index("comment")); // comment style as fallback
+        if highlight.is_none() {
+            return Box::new(::std::iter::empty());
+        }
+        let highlight = highlight.unwrap();
```
casperstorm, line 371:
```suggestion
        let Some(highlight) = theme
            .find_scope_index("ui.virtual.jump.dim")
            .or_else(|| theme.find_scope_index("comment")) else {
            // comment style as fallback
            return Box::new(::std::iter::empty());
        };
```

Same as above, `let-else` simplifies this.

semin-park: Done as well

semin-park:
Thanks for all of the suggestions!

As @QuinnFreedman mentioned, I would prefer to see this get merged first before we add more features in. This is already a large PR with ~1000 lines of code and it's getting merge conflicts every now and then. Adding more features in would likely make it harder to review and more vulnerable to merge conflicts until this gets a proper review.

book/src/keymap.md
```diff
@@ -218,6 +218,8 @@ Jumps to various locations.
 | `n`   | Go to next buffer                                | `goto_next_buffer`         |
 | `p`   | Go to previous buffer                            | `goto_previous_buffer`     |
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
+| `w`   | Word-wise jump mode                              | `jump_to_identifier_label` |
+| `/`   | Character search jump mode                       | `jump_to_char_label`       |
```
oyarsa, line 222:
```suggestion
| `/`   | Character search jump mode                       | `jump_to_str_label`       |
```

According to `helix-term/src/commands.rs`, line 480.

semin-park: Oh thanks a lot for catching this out! Fixed :)

GloopShlugger:
Thank you for rebasing so frequently i can not live without it anymore 🫡

bredfield:
Wow this is awesome, thanks for all the work!

So1aric:
Will this be merged into the main branch? Just curious about whether and when :D

semin-park:
This didn't really get any proper reviews yet.. but I hope it's coming soon.

(and I'll try to resolve those conflicts within this week)

pascalkuthe:
I can't really make you any promises right now. Up until recently my focus was the 23.5 release and the multi LSP PR (large breaking change, waiting for review for a year) since then I have been extremely busy with work so I dont really have much time to review large PRs and only review small PRs as they pop up. I will try to merge this into my daily driver and try to review but I can't give you a timeline (especially since two reviewers are needed)

semin-park:
Thanks for the comment!

Yeah I understand that, work should have higher priority after all.

As long as there aren't any major conflict that is very hard to resolve, I'm okay with resolving conflicts until reviewers have time :)

runtime/themes/everforest_dark.toml
```diff
@@ -100,6 +100,11 @@
 "ui.virtual.indent-guide" = { fg = "bg4" }
 "ui.virtual.inlay-hint" = { fg = "grey0" }
 "ui.virtual.wrap" = { fg = "grey0" }
+"ui.virtual.ruler" = { bg = "bg3" }
```
pascalkuthe, line 103: this theme key has been added on master too so you need to remove it here or the theme breaks

semin-park: Oh how did that crawl in haha, thanks!

YaLTeR:
Thanks for making this! EasyMotion-style two-character jump is one of my favorite ways of getting around, and this is close enough.

Found a bug: if I activate the actions from Space-? then they won't reset on Esc or other button press, which can cause all kinds of glitched output.

evanrichter:
> Found a bug: if I activate the actions from Space-? then they won't reset on Esc or other button press, which can cause all kinds of glitched output.

confirmed here too. looks like jump mode doesn't work at all from the Space-? menu for me. The jump targets highlight with characters, text is dimmed, but you can't actually jump anywhere. If you activate jump mode through `gw` and either jump somewhere or ESC, then normal behavior resumes.

the-mikedavis:
That behavior has already been mentioned: https://github.com/helix-editor/helix/pull/5340#issuecomment-1404576970

It's a problem with the way the command palette works rather than a problem with the feature here.

semin-park:
Sorry, fixed the lint issue just now. Could you run it again?

SecretPocketCat:
I really love this, hope this gets merged ASAP so that I don't have to build from src 🙂

One little thing I encountered is that line 26 in my config collapses when entering jump mode shifting all lines that follow which is quite disorienting
[config.zip](https://github.com/helix-editor/helix/files/12268903/config.zip)
Not sure what I did there as replacing the empty line negates the issue, but I figured I might as well share the repro.


Similar shift happens when there are too many tags (e.g. matching a very common char like double quotes in the config) but I can't really imagine a way around that.

dvic:
@semin-park this PR is awesome, thank you for your hard work 🙏! Is it a lot of work to rebase/ merge master in this pr? I'm not familiar with the code base but I could give it a try and open a pull request on your branch.

Omnikar:
> @semin-park this PR is awesome, thank you for your hard work 🙏! Is it a lot of work to rebase/ merge master in this pr? I'm not familiar with the code base but I could give it a try and open a pull request on your branch.

I have rebased it here: https://github.com/semin-park/helix/pull/1

tmpm697:
Really need this feature as I like `https://github.com/ggandor/leap.nvim` a lot, reduce so much keystrokes.

ntzm:
One minor issue I've found is that modifiers are removed from dimmed text. In your example videos the italic comments are kept as italic when they are dimmed. However testing locally this does not seem to be the case.

vaheqelyan:
This feature is so much needed. Any updates on this?

semin-park:
Oh sorry for the late reply, I'll get back to resolving the conflicts maybe this or next week.

Been quite busy lately haha

helix-term/src/ui/editor.rs
DSoliz: perhaps `hone` or `skip`? a different name would certainly make searching for the feature/related-code easier

conways-glider:
Is this still being worked on? If not, what would it take to get this across the line?

yerlaser:
> Is this still being worked on? If not, what would it take to get this across the line?

It needs to be reviewed and core contributors currently don't have time for this.
In the meantime, if you're impatient, you can try helix-ext as it incorporates many such PRs which didn't make it to the Helix proper.

SecretPocketCat:
> Is this still being worked on? If not, what would it take to get this across the line?

I'd venture a guess this's not gonna be merged since #8875 exists. A solution like this will have to wait for the plugin system whenever that drops.

Drazhar:
> > Is this still being worked on? If not, what would it take to get this across the line?
> 
> It needs to be reviewed and core contributors currently don't have time for this. In the meantime, if you're impatient, you can try helix-ext as it incorporates many such PRs which didn't make it to the Helix proper.

Can you give a link to Helix-ext? Sounds interesting, but I couldn’t find it.

icecreammatt:
> > > Is this still being worked on? If not, what would it take to get this across the line?
> > 
> > 
> > It needs to be reviewed and core contributors currently don't have time for this. In the meantime, if you're impatient, you can try helix-ext as it incorporates many such PRs which didn't make it to the Helix proper.
> 
> Can you give a link to Helix-ext? Sounds interesting, but I couldn’t find it.

https://github.com/omentic/helix-ext

yerlaser:
> > > Is this still being worked on? If not, what would it take to get this across the line?
> > 
> > 
> > It needs to be reviewed and core contributors currently don't have time for this. In the meantime, if you're impatient, you can try helix-ext as it incorporates many such PRs which didn't make it to the Helix proper.
> 
> Can you give a link to Helix-ext? Sounds interesting, but I couldn’t find it.

You already have the link :)

I tried that myself some time ago and I like the features very much, the jump-mode, color-coded brackets and indents.
But, it did crash on me a couple of times and so I switched back to the proper Helix because I didn't want to loose my data.
Your mileage may vary, though.

James-Firth:
Excited when there's time for this to be reviewed!

Not to derail or scope creep this PR, this could be a future feature on top of this jump mode? Just want to mention the idea here in case it would require a deep rework if delayed:

It would be great if you could also use jump mode to place a _new_ cursor for those tricky multi-cursor moments.

p-98:
Sad to see this pr closed, as I find the leap.nvim style more intuitive to work with. Is there a possibility we can get this kind of jumping as well?

yerlaser:
> Sad to see this pr closed, as I find the leap.nvim style more intuitive to work with. Is there a possibility we can get this kind of jumping as well?

I'm maintaining a fork with a similar feature, you can check it out here: https://github.com/Helix-Unstable/helix-unstable/blob/unstable/README_FORK.md

PR #5341, opened 2022-12-30
Clean up completion, signatureHelp, and hover popup after mouse action

hunterliao29:
Completion, signatureHelp, and hover popup should be removed after mouse cursor moved


pascalkuthe:
This PR is stale and hasn't seen development in over a year. 

Signatur help has already been covered by #8021 "by accident. 

Mouse support is not a priority for us and particularly the mouse-moved behavior is not something that I see gaining traction. Thank you for contributing

PR #5349, opened 2022-12-30
Fix Broken Attribute Highlights

Chickenkeeper:
The deleted code contained keys that aren't used by Helix's themes, caused inconsistent highlighting of punctuation, highlighted lowercase arguments as integers, and was in the wrong section of the file.

However it was also highlighting `derive` as `@special` instead of `function.macro`, so if we want to keep that behaviour then maybe something like this could be inserted at the top of the `Macros` section instead?
```scheme
(attribute
  (identifier) @special
  (#eq? @special "derive")
)
```

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -140,6 +140,8 @@ We use a similar set of scopes as
 
 - `type` - Types
   - `builtin` - Primitive types provided by the language (`int`, `usize`)
+  - `enum`
+    - `variant`
 - `constructor`
 
 - `constant` (TODO: constant.other.placeholder for %v)
@@ -201,6 +203,8 @@ We use a similar set of scopes as
 - `tag` - Tags (e.g. `<body>` in HTML)
 
 - `namespace`
+
+- `special`
 
 - `markup`
   - `heading`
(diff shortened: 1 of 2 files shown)


the-mikedavis:
`@special` is undocument but used in a bunch of themes. It's used for rust's `?` operator and in the picker for highlighted matches. `@_macro` or any other scope with an underscore is used within a match just so the capture can be referenced in the predicate: here it's used so we can assert that the attribute's name is `derive`. The `@_macro` capture can probably be replaced with `@special` so that the `@special` highlighting doesn't apply to the `#[`/`,`/`]` parts of the attribute. The only thing that looks really out-of-place to me in this stanza is the `constant.numeric.integer` capture. I didn't test this but it can most likely be removed or replaced with `@type`

Chickenkeeper:
That makes sense. I've added the stanza I suggested plus your suggestion of capturing the `arguments` as `type` instead of `constant.numeric.integer` to handle the edge case where a `derive` might need to use a trait with a lowercase name, or to stop the colour changing as a name is being typed. I've also added `special` to the docs, as well as `type.enum.variant` which is also used a lot but undocumented

PR #5353, opened 2022-12-31
Add GitHub themes

OwOSwordsman:
A GitHub color theme based on the VSCode GitHub theme. The variants are based on the Primer Primitive Colors. Templates were created for the Light and Dark variants to closely match the VSCode ones. The others were generated based on the two templates.
Screen Shot 2022-12-31 at 5 33 15 PM
Screen Shot 2022-12-31 at 5 31 57 PM


archseer:
Are just the palettes different? If so you could use inherits to only override the palette, and declare the scopes once

archseer:
Great work on this by the way! I really wanted these themes included :)

OwOSwordsman:
> Are just the palettes different? If so you could use inherits to only override the palette, and declare the scopes once

Oh right, theme inheritance was just added. I am not sure if the built-in themes should be using the inherit feature or not tho.

pascalkuthe:
> > Are just the palettes different? If so you could use inherits to only override the palette, and declare the scopes once
> 
> Oh right, theme inheritance was just added. I am not sure if the built-in themes should be using the inherit feature or not though.

Other builtin themes (like `tokio`, `autumn` and `hex`) already do and it's usually preferable to duplication

archseer (approved):
Thanks!

PR #5356, opened 2023-01-01
Add command to rotate view backward

mejo13:
Adds command 'rotate_view_backward' and changes 'rotate_view' to 'rotate_view_forward' to stay consistent.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -386,6 +386,7 @@ impl MappableCommand {
         swap_view_down, "Swap with split below",
         transpose_view, "Transpose splits",
         rotate_view, "Goto next window",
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Is there a use-case for this command or is it just to be symmetric with the current `rotate_view`?

mejo13:
Using less key strokes when switching between 3+ views. In hindsight this doesn't have to be a breaking change though. Calling it `rotate_view_reverse` and leave `rotate_view` as is would be acceptable maybe?

the-mikedavis:
Ah yeah I like that, let's call it `rotate_view_reverse` 👍

the-mikedavis (approved):

PR #5357, opened 2023-01-01
Adds optional bottom border

grarco:
Adds a config parameter to draw a horizontal border between splits. This can be helpful when the status line uses the same background color as the editor, since horizontal splits become hard to tell apart, especially when they are showing documents of the same type.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -157,15 +157,28 @@ impl EditorView {
             Self::render_focused_view_elements(view, doc, inner, theme, surface);
         }
 
+        let border_style = theme.get("ui.window");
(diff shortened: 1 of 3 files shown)


the-mikedavis:
> when the status line uses the same background color as the editor

Is there a theme that does this currently? I think this should be solved through theme changes rather than adding a bottom bar.

grarco:
Good question, tbh I'm using a custom theme and I really like my status line being the same color as the background. I haven't investigated whether some of the provided themes experience the same issue

the-mikedavis:
I would prefer this to be accomplished by fixing the colors in the theme rather than making the rendering configurable. I would really like to avoid adding config flags that affect how we render things since they usually make theming more complicated and make debugging harder

PR #5378, opened 2023-01-03
Multi cursor increment

tanj:
This feature lets you replace a multi-cursor selection with an incrementing value.

- numeric (starting with 1)
- numeric (starting with 0)
- ascii alphabet incrementing the same way excel columns do
- roman numerals

The starting value can be changed with the repeat count for the command.

https://youtu.be/-mSc2fB35Cg

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -68,6 +68,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d468802bab17cbc0cc575e9b053f41e72aa36bfa6b7f55e3529ffa43161b97fa"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


the-mikedavis (changes requested):
Replacing with selection index should be possible by adding [Kakoune's special `#` register](https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers) so that you could use `"#R`. Replacing zero indexed can then be done by decrementing those selections: `"#R<C-x>`. The rest seem too specific to me to add extra commands for and are definitely not worth pulling in extra dependencies for.

PR #5379, opened 2023-01-03
Fix Sonokai theme to better match original

NickCondron:
The Sonokai theme had some inconsistencies, so I fixed some of them.

* Copied the palette exactly from https://github.com/sainnhe/sonokai/blob/master/autoload/sonokai.vim
* Used the following two resources to pick correct colors for different scopes:

1. https://www.sainnhe.dev/post/contributing-guide/#sonokai
2. https://github.com/sainnhe/sonokai/blob/master/colors/sonokai.vim

Edit: Changes I made as listed in the git commit:
* Make constants white (1)
* Make builtin variables purple (1)
* Make members orange (1)
* Make labels red (2)
* Make operators red (1)
* Make all punctuation grey (2)
* Make builtin functions and macros green (2)
* Make diff delta blue (2)
* Make cursor match bg4 (2)
* Make visible whitespace bg4 (2)

### DIFF
--- a/runtime/themes/sonokai.toml
+++ b/runtime/themes/sonokai.toml
@@ -6,24 +6,22 @@
 # License: MIT License
 
 "type" = "blue"
-"constant" = "purple"
+"constant" = "fg"
 "constant.numeric" = "purple"
 "constant.character.escape" = "orange"
 "string" = "yellow"
 "comment" = "grey"
 "variable" = "fg"
-"variable.builtin" = "orange"
+"variable.builtin" = "purple"
 "variable.parameter" = "fg"
-"variable.other.member" = "fg"
-"label" = "orange"
+"variable.other.member" = "orange"
+"label" = "red"
 "punctuation" = "grey"
-"punctuation.delimiter" = "grey"
-"punctuation.bracket" = "fg"
 "keyword" = "red"
-"operator" = "orange"
+"operator" = "red"
 "function" = "green"
-"function.builtin" = "blue"
-"function.macro" = "purple"
+"function.builtin" = "green"
+"function.macro" = "green"
 "tag" = "yellow"
 "namespace" = "blue"
 "attribute" = "purple"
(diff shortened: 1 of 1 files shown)


the-mikedavis:
\cc @p4ymak what do you think about these colors (since you originally ported sonokai)?

NickCondron:
Copying @p4ymak comments here because his comment was made in error on another PR:

> @the-mikedavis
sorry for long respond.
>
>in my opinion, in general, the palette has become better. there are only two things I don't like: macros are no different from functions now, and parentheses and punctuation are too dim.

1. Sonokai colors constant-macros purple and function-macros green. Helix does not currently have scopes that differentiate the two kinds. Given that Helix does have a "function.macro" scope I went with green.
2. Sonokai clearly makes brackets and delimiters grey (https://github.com/sainnhe/sonokai/blob/e2e89497f178bec568a0a43cda651f0429ec8ec1/colors/sonokai.vim#L434). I understand that could be too dark for some tastes, but I think it makes sense to stick to the original as much as possible. I did realize that it makes special punctuation yellow, so I will amend the colorscheme to add that.

archseer:
Was the review fully addressed here?

the-mikedavis:
I lost track of this PR a little - if you're still interested in updating these scopes could you rebase / merge master?

NickCondron:
When I wrote this PR I was learning Helix. I no longer use the editor. Respectfully, I don't want to spend a whole lot of time on this. If it's an improvement and works I would be thrilled if it were merged. If not, hopefully someone else can expand on what I did.

I chose fidelity to the original theme where possible. However, I'm not the one who should decide on different color variations for secondary selections or other Helix-specific things. I tried to respect the more recent PRs. Also, I don't know if other scopes have been added/changed since I originally made the PR that could make this theme better.

the-mikedavis (approved):

pascalkuthe (approved):

Issue #5377, opened 2023-01-03
Regression with syntax highlighting when reloading the config


phaazon:
### Summary

When calling `:config-reload` in a buffer, it breaks syntax highlighting, turning everything white. Closing the buffer and opening a new one fixes the problem, so that’s a temporary issue.

I bisected and commit 9d15b852093480d852e47dd9f0157828c2130298 introduces the bug. You can try with its parent, 3fe3f2c4ee1c483ad2d1f31dc34e33f15fc977db; it doesn’t reproduce.

### Reproduction Steps

1. Open `hx` and any buffer (I usually do `:config-open`).
  ![image](https://user-images.githubusercontent.com/506592/210291189-9a09e3cf-b628-4eb6-8aaa-4a7bbe8457d8.png)
2. `:config-reload`:
  ![image](https://user-images.githubusercontent.com/506592/210291205-3dbee793-46e8-41c3-8c71-89a6b4e3bdc8.png)

### Helix log

Nothing interesting (I cleaned it up, it doesn’t show anything useful).

### Platform

Linux

### Terminal Emulator

kitty 0.26.5

### Helix Version

helix 22.12 (febe1a51) and higher

PR #5381, opened 2023-01-03
Fix language config reload logic

zyklotomic:
Fixes  https://github.com/helix-editor/helix/issues/5377.

I think the bug https://github.com/helix-editor/helix/issues/5377 stemmed from not updating the reference to the syntax loader in the `editor` as well.

Hence, it probably was the case that before, the old reference was in fact actually never dropped, and that the `editor` was still pointing to the old syntax loader.

Took the chance to improve some status messages too.


the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #5383, opened 2023-01-03
Add split variants for LSP goto commands.

loralb:
Supersedes #2681.

Added `hsplit` and `vsplit` variants for following commands:
- `goto_definition`
- `goto_type_definition`
- `goto_implementation`
- `goto_reference`

I also assigned default keybindings in `window` mode (following current behaviour for `goto_file` command).

First PR here, so feedback is more than welcome!


the-mikedavis:
These can be done by creating a window split at the current location and then using the goto commands: `<C-w><C-v>gd` for example to go to the definition of the symbol under the cursor in a vertical split

loralb:
Well, yes, fair enough 😅.
Though I have to admit that looking at the `goto_file` implementation, which has these split variants commands even though the same "trick" could be used (for example, `<C-w><C-v>gf`), I guessed the commands way was the prefered one.
Anyway, the functionality is there already and can also be easily mapped to a custom keybinding (for example like this: `'your_keybinding' = ["vsplit","goto_definition"]`), so I may close the PR, if no further discussion is needed.
Thank you for your time!

lazytanuki:
Hi ! Just a though here. This could still be a nice thing to have, because the `["vsplit", "goto_definition"]` binding will split the window even if the `goto_definition` failed, which can be a bit frustrating.

loralb:
Hi @lazytanuki!
It's true, when the `goto_definition` command fails, we end up with a "useless" split. To be fair though, in a real case scenario this won't happen very often and, when it does, closing the new split is a `<C-w>q` away.
In the end, this is a matter of personal preference (if you ask me, I prefer commands), but the functionalities this PR offers are already there for the most part. So this is up to which way the maintainers want the project to follow.
Unless there are drawbacks in adding new commands, which is something I honestly do not know.
Thank you for your feedback!

PR #5386, opened 2023-01-03
Find out comment style from file content as a fallback

### LINKED ISSUE #5380: Read file content to determine comment style
There are situations where we need to work with temporary files without any file extension. For example, the [repo](https://source.android.com/docs/setup/download#repo) tool asks the user to uncomment a line when multiple branches have unpushed changes:

In `/tmp/tmp4bbviluy `:
```
# Uncomment the branches to upload:
#
# ...
```

It's clear that the comment style here is `#`, but helix currently defaults to `//` [if no language has been detected](https://github.com/helix-editor/helix/blob/master/helix-core/src/comment.rs#L63).

We already have something similar to this proposal for [shebang](https://github.com/helix-editor/helix/blob/a895a4e53a81da2264fee9dd3b5eec313794ea0b/helix-view/src/document.rs#L615) lines, so we could follow this pattern and extend behavior to files without extension.

semin-park:
Usually we find out the language based on file extension and use the comment style used for the language. This PR lets helix figure out the correct comment style by peeking the first few lines of the file. This is a fallback, when helix is unable to figure out the language for any reason.

fixes #5380

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -556,6 +556,10 @@ pub struct Loader {
     language_config_ids_by_extension: HashMap<String, usize>, // Vec<usize>
     language_config_ids_by_suffix: HashMap<String, usize>,
     language_config_ids_by_shebang: HashMap<String, usize>,
+    // Refer to #5380 for context. We can't find out the correct language with
+    // the comment token, but we're choosing an arbitrary language that uses
+    // the comment token.
+    language_config_ids_by_comment_token: HashMap<String, usize>,
 
     scopes: ArcSwap<Vec<String>>,
 }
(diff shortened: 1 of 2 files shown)


the-mikedavis (changes requested):
It looks like this tries to determine the full language config from only the comment token. For common comment tokens like `//` or `#` this will most likely determine the wrong language and give confusing syntax highlighting

kirawi:
An alternative I would propose is to guess the comment style without setting a language for the file.

Schweber:
> An alternative I would propose is to guess the comment style without setting a language for the file.

This would be very helpful with .config files with arbitrary rules. I was going to suggest such a feature when i found this PR.

Could this please be implemented?

pascalkuthe:
clsoing this one out as stale as there are still unadressed review comments. Thank you for contributing!

PR #5390, opened 2023-01-04
Templatized global search as a typable command

seewishnew:
## Description
This PR adds two new typable commands: one for global search and the other for a templatized global search using the current doc selection. These are made typable for enabling key remappings to further ease the pain of globally searching with complex regex surrounding the symbols that are currently selected.

### Motivation
During visual debugging, whenever the LS is unable to go to definition, especially if the definition lies in a separate file, or if it is in a separate language like with protobuf, I often find myself repeating the same boilerplate regex to get to the definitions of these symbols. This can become painful when forced to repeat this hundreds of times each week. I believe that having a typable global search that can accept a regex template as an argument, which can be filled by the current doc selection will greatly improve the search experience. 

For example, today if a Python class cannot be found by the LS I'd have to go to `global_search` mode first and then type something like `class\s+XYZ` with no way to remap the prefix `class\s+` to a key. While this does work, it is much more convenient if we have a templatized search method in typable command mode that can then be remapped.

## Use
The templatized global search (`:tgs`) takes in up to two arguments: the first argument is a templatized regex, with empty `{}` placeholders like with the rust `format!` macro, the optional second argument is for a delimiter, and on execution, the current document selection is split with the delimiter if it is present, and the results are used as parameters to format the template into a regex expression to search in the current workspace.

The global search `:gs` is pretty much the same as the `global_search` command but it is typable, allowing for key remaps to supply a constant regex argument to use. This does not use the current document selection in any way.

## Examples
demo optimized

There could be more advanced use cases for more complex pattern matching that requires multiple inputs to be filled in. Right now this implementation uses a delimiter to split the current selection to pass in as arguments to fill the regex template. Say we have a comma-separated list of test class names that we want to search for: `TestABC, TestDEF, TestXYZ`, we can search for all these classes at once with `:tgs class\s+\({}|{}|{}\) ,`. Granted, this extended use case appears more esoteric and specialized, but implementing a generalized template model with support for an arbitrary number of arguments made more sense than just having one parameter for just the current selection. 

By adding something like the following to my config file, I can now search for classes, structs, or protobuf message definitions in a couple of keystrokes: 
```toml
[keys.normal.space]
"/" = { "/" = "global_search", "c" = ":tgs (struct|class|message)\\s+{}", "f" = ":tgs (def|fn|func)\\s+{}" }
```

## Caveats
- Right now the regex supplied to the template command cannot have whitespace in them and will instead have to use `\s` to represent whitespace pattern.
- The templates are added to the `?` register for auto-complete and the filled-in regexes are added to the `/` register as it does not make sense for the regular search functions to have the incomplete template with `{}` placeholders (which is invalid regex anyway) and further clutter the search history. As far as I am aware, the `?` register is used for auto-complete only by the template global search command right now.

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -11,6 +11,8 @@
 | `:buffer-close-all!`, `:bca!`, `:bcloseall!` | Force close all buffers ignoring unsaved changes without quitting. |
 | `:buffer-next`, `:bn`, `:bnext` | Goto next buffer. |
 | `:buffer-previous`, `:bp`, `:bprev` | Goto previous buffer. |
+| `:template_global_search_selection`, `:tgs` | Templatable global search as a typable command. Accepts one mandatory regex argument that can have `{}` placeholders and an optional delimiter argument. The current selection in the document is split by the delimiter if present, and is used to fill the placeholders. |
+| `:global_search`, `:gs` | Global search as a typable command. |
 | `:write`, `:w` | Write changes to disk. Accepts an optional path (:write some/path.txt) |
 | `:write!`, `:w!` | Force write changes to disk creating necessary subdirectories. Accepts an optional path (:write some/path.txt) |
 | `:new`, `:n` | Create a new scratch buffer. |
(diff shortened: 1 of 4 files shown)


pascalkuthe:
this will be covered by a compination of command expansion (#6979) and setting the search register with a typable command (plus maybe a way to start a picker with an active prefill). In the future more advanced cases can be handled by plugins. This pr has been stale for a year and I don't see us moving in this direction so closing this. Thank you for contributing

PR #5393, opened 2023-01-04
feat: add buffer-close-hidden commands

mtoohey31:
This pull request adds the `buffer-close-hidden` and `buffer-close-hidden!` commands, which close all buffers that are not currently visible.

I'd like to add this because I find `buffer-close-others` very useful (since I often end up with a bunch of unnecessary buffers open when I'm jumping around with a language server), but that command isn't ideal when I have other buffers in a split that I want to keep open.


mtoohey31:
Hello! I know it's been a while, but I'm still interested in this. I've rebased it and simplified things as much as possible.

Please let me know if there's something problematic with this or if I'm missing an obvious alternative. If there just hasn't been time to get to it yet though, no worries!

mtoohey31:
I'd still really like to have this... Sorry for the ping @kirawi, but is there anything I can do to move this forward? No worries if there just hasn't been time; admittedly the lack of comments by other users seems to indicate that this might not be something many people have run into, so I understand if it's not a priority. That being said, the changes are quite minimal so I don't think it introduces much extra maintenance burden.

the-mikedavis:
I think this could fit as a flag (https://github.com/helix-editor/helix/issues/5828) on `:buffer-close-others(!)`, so it would be `:bco --non-visible` or `:bco --background` (naming is hard)

mtoohey31:
Ok, I'll keep an eye on #12527 then.

mtoohey31:
Hi @the-mikedavis, I see that #12527 has been merged, so I've rebased this to make use of that. I've used `--skip-visible` as the flag name for now (with `-s` for the short version), though naming is hard as you said, so if you prefer something else lmk.

PR #5404, opened 2023-01-04
Update catppuccin themes

beeb:
According to latest changes in https://github.com/catppuccin/helix


archseer:
Can you also bring in the inherits change from https://github.com/helix-editor/helix/pull/4581

beeb:
This would have to be done in the upstream I guess. I don't feel comfortable changing the theme in this repo when the official theme repo doesn't implement those changes.

beeb:
@seshotake @archseer gotcha, I misunderstood, sorry. I can do it here, not sure if the upstream should use this as they might want to keep each theme independent of the others. For the default helix it makes sense though. Pushing in a sec.

the-mikedavis (approved):

PR #5407, opened 2023-01-04
Docs: Put default theme to a light theme

sgued:
Until this patch, the docs always loaded with the "Colibri" theme, a dark theme, even for light theme users. This PR switches it to be  the "Rust" theme, a light but not all white theme.
MdBook will still switch to the "Colibri" theme if the reader's browsers is set to prefer dark themes (`prefers-color-scheme: dark`)

### DIFF
--- a/book/book.toml
+++ b/book/book.toml
@@ -7,6 +7,6 @@ edit-url-template = "https://github.com/helix-editor/helix/tree/master/book/{pat
 
 [output.html]
 cname = "docs.helix-editor.com"
-default-theme = "colibri"
+default-theme = "rust"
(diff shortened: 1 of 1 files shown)


sgued:
While I love dark modes (sometimes), consistency is a much more important factor. Most of my apps run in a light mode, so I dim the screen brightness to compensate. This means that sites that enforce a dark mode are very hard to read.

Every one has their preference, and I think the best solution is to simply respect each user's preferred configuration. Mdbook already does that.

pascalkuthe:
This is helixes default theme and we want the documentation page to match that. We will not be changing the default. Thank you for contributing

Issue #584, opened 2021-08-14
runtime Data Location


theduke:
Currently the runtime data is supposed to live in `~/.config/helix/runtime`, which is somewhat against the XDG spec.

Files like this should normally be placed in `~/.local/share/helix/`. 

`~/.config` should only hold user provided configuration.

archseer:
So in a packaged installation the runtime folder would be installed into a system folder (for example `/usr/lib/helix/runtime` on Arch) then the binary is wrapped in a script that points HELIX_RUNTIME to the location.

I consider `~/.config/helix/runtime` the place where users will be able to provide their own queries and grammars specific to their config.

For builds from source we currently recommend symlinking to that directory more for convenience: it's a frequent source of friction because users forgot to specify a HELIX_RUNTIME pointing to their runtime dir. In the future I definitely think that's the wrong thing to do. I guess we could recommend that users use a symlink in `~/.local/share/helix/runtime` if building from source?

pickfire:
> ~/.config should only hold user provided configuration.

I think it's a gray area here. Helix `runtime` contains some user configuration as well and users are welcome to edit it although being less used compared to `config.toml`.

mmatous:
I'm in favor of changing current behavior.
Current fallback is executable's directory. I just had Helix scream at me because it wanted to put stuff in `/usr/local/bin` while running as a user. 
Proposed improvements:
- disregard XDG_DATA_DIRS (`/usr/local/share:/usr/share`) for now
- look for `config.toml` in HELIX_CONFIG directory, default to `XDG_CONFIG_HOME/helix`, create if necessary
- same dir is default location of `languages.toml`
- `config.toml` may contain additional paths (languages.toml, data_dir, runtime_dir, log_dir), I disagree with todo notes in sources wanting envvar per dir, one is enough
- grammars should be separate from config, default to `XDG_DATA_HOME/helix`, create if necessary
- `helix.log` belongs to newest XDG addition, `XDG_STATE_HOME/helix`, create if necessary
- if XDG_ vars are empty, dump it all into `HOME/.helix` (or fallback to paths XDG usually uses? does anyone still put things in HOME? i know XDG isn't used by every *nix, but I don't know how crossplatform Helix is)

Even better solution would include Helix merging config and grammars from system dirs and user dirs and having fallbacks configurable at buildtime but meh. This should be enough for most usecases.

I'll make the PR, if you want.

the-mikedavis:
[...]

PR #5411, opened 2023-01-05
Generalised to multiple runtime directories with priorities

### LINKED ISSUE #3346: Treat runtime files from installation as immutable

paul-scott:
This is an implementation for #3346 
See commit message for an overview.

Here are some thing to consider when reviewing.

Depending on how helix is packaged, it still might not be possible for some users to conveniently override runtime files. E.g. on AUR helix-git, `/usr/bin/hx` is a shell script that sets `HELIX_RUNTIME` to `/usr/lib/helix/runtime` before calling `/usr/lib/helix/hx`. Setting this env var is not necessary because helix already automatically checks for a runtime directory in the same directory as the executable. Setting it prevents `~/.config/helix/runtime` from getting priority from the default runtime installation.

[The docs suggest](https://docs.helix-editor.com/master/install.html#build-from-source) that the `HELIX_RUNTIME` is intended as an option to override a `~/.config/helix/runtime`, which I agree could be useful and why I've initially kept this behaviour. However, if it looks like packager's and build-from-source installers want to use this env var for their default runtime install, then we might want to rethink the priority order as outlined in the commit message.

For consistencies sake you might want to consider if `~/.config/helix/themes` should be deprecated since this commit makes it so that `~/.config/helix/runtime/themes` can be used to the same effect. Before and after this commit the former always has the higher priority.

It just remembered that I haven't touched the user doc / book. I'll have to have a look and see where the new behaviour can be easily explained.

Closes #3346


helix-loader/src/lib.rs
```diff
@@ -25,31 +26,83 @@ pub fn initialize_config_file(specified_file: Option<PathBuf>) {
...
+pub fn runtime_dirs() -> &'static [PathBuf] {
+    &RUNTIME_DIRS
+}
+
+/// Find file with path relative to runtime directory
+///
+/// `rel_path` should be the relative path from within the `runtime/` directory.
+/// The valid runtime directories are searched in priority order and the first
+/// file found to exist is returned, otherwise None.
+fn find_runtime_file<P: AsRef<Path>>(rel_path: P) -> Option<PathBuf> {
```
archseer, line 82: I'd avoid `AsRef`/monomorphization here since it can grow the binary by generating a lot of alternative method bodies. Just take `&Path` here

helix-loader/src/lib.rs
```diff
@@ -25,31 +26,83 @@ pub fn initialize_config_file(specified_file: Option<PathBuf>) {
...
+///
+/// The priority is:
+///
+/// 1. `HELIX_RUNTIME` (if environment variable is set)
+/// 2. sibling directory to `CARGO_MANIFEST_DIR` (if environment variable is set)
+/// 3. subdirectory of user config directory (always included)
+/// 4. subdirectory of path to helix executable (always included)
+///
+/// Postcondition: returns at least two paths (they might not exist).
+fn prioritize_runtime_dirs() -> Vec<PathBuf> {
```
archseer, line 39: It seems to me that this could be done in a `Lazy` block and stored as `static RUNTIME_DIRS` so that we only compute this vector once

archseer: Ah sorry, I missed that this was actually done above!

archseer:
Priority should be (highest to lowest):

1. `~/.config/helix/runtime` (user's configuration)
2. `HELIX_RUNTIME` (indicates the system's runtime, will be usually be under `/usr/` for package installs, not writable)
3. current executable's directory (if all else fails, it might just be someone running it from a tarball)

This means that for grammar fetch & build we probably want to write into the first dir? Except on CI \cc @the-mikedavis

helix-term/src/health.rs
```diff
@@ -66,17 +66,19 @@ pub fn general() -> std::io::Result<()> {
...
+        if let Ok(path) = std::fs::read_link(&rt_dir) {
+            let msg = format!("Runtime directory is symlinked to {}", path.display());
+            writeln!(stdout, "{}", msg.yellow())?;
+        }
+        if !rt_dir.exists() {
+            writeln!(stdout, "{}", "Runtime directory does not exist.".red())?;
+        }
+        if rt_dir.read_dir().ok().map(|it| it.count()) == Some(0) {
+            writeln!(stdout, "{}", "Runtime directory is empty.".red())?;
+        }
```
archseer, line 81: I'd just output this similar to `PATH` or `GOPATH`: a bunch of directories separated by `;`

paul-scott:
> Priority should be (highest to lowest):
> 
>     1. `~/.config/helix/runtime` (user's configuration)
> 
>     2. `HELIX_RUNTIME` (indicates the system's runtime, will be usually be under `/usr/` for package installs, not writable)
> 
>     3. current executable's directory (if all else fails, it might just be someone running it from a tarball)
> 
> 
> The docs recommend setting `HELIX_RUNTIME` just because I felt it's simpler to use an env var than to copy files over and then forget to update them in the future.
> 
> This means that for grammar fetch & build we probably want to write into the first dir? Except on CI \cc @the-mikedavis

In that case how about:

1. sibling `runtime` directory to `$CARGO_MANIFEST_DIR`
2. subdirectory of user config directory (`~/.config/helix/runtime`)
3. `$HELIX_RUNTIME`
4. subdirectory of path to helix executable

Compared to the original behaviour this is just swapping 2 and 3.

I assume my 1 here is what the CI uses. In addition it seems like `helix-git` on AUR  fetches and builds the grammars under 1, and then packages them up from there so they get installed under `/usr/lib/helix/runtime`.

I think it is best if this kind of packaging behaviour remains. It would be a pain if on updating a packager started putting (and overriding) the grammars under `~/.config/helix/runtime`. A user might intentionally be trying to keep grammars in there to override the defaults.

archseer:
Oh I agree, I forgot to list it above :) Just wanted to clarify the intended use of `HELIX_RUNTIME`

helix-term/src/health.rs
paul-scott: In addition to the runtime paths, the original behaviour conditionally reports for each runtime directory:

1. the destination if a runtime directory is symlinked
2. a red warning if the runtime directory doesn't exist
3. a red warning if the runtime directory is empty

I assume you still want this info displayed? How should we clearly display this using a `;`-separated list?  For 2 and 3, we can just filter on the non-existent / empty directories and join them on `;`. Should we just do the same thing for 1, and leave it up to the user to work out *which* runtime directory is the symlink?

gibbz00:
Should we also make use of XDG_RUNTIME_DIR? I'm personally not of fan of placing runtime files the .config folder.

archseer:
It's orthogonal to this PR. https://github.com/helix-editor/helix/pull/2135 implements that but it should be rebased after this is merged

helix-term/src/health.rs
paul-scott: I've change the implementation so that a `;`-separated list of runtime directories is presented. For the warnings / errors,  changed them all to warnings (yellow) because now it might be more common for a runtime directory to be missing or empty and it not be a problem. I didn't join these paths on `;` in the warnings, instead I just report a warning per case.

One thing is that windows / linux use `;` / `:` for this separator in `PATH` and `GOPATH`. I'm not sure we want to go to the effort to make this OS aware, so I just went with your suggestion for the semicolon. This seems like it would cause the least confusion given windows using `:` after drive letter in path `C:\\` etc.

Let me know what you think.

paul-scott:
> Oh I agree, I forgot to list it above :) Just wanted to clarify the intended use of `HELIX_RUNTIME`

I've implemented pushing `HELIX_RUNTIME` down in priority below the user config runtime.

In terms of fetch and build, the behaviour is now that a cargo build will always put grammars in the runtime directory within the helix repo being built. `HELIX_RUNTIME` can no longer be used to change where they will automatically be fetched / built.

When a user calls `hx -g fetch / build` this is now always put in `~/.config/helix/runtime/grammars/`.

None of this is hard-coded, it is just the way it works with the new priority ordering and how the implementation of fetch and build always grab the highest priority runtime directory.

I've modified the build from source instructions in the book to reflect some of this but not in that much detail.

Let me know if you would like to further tweak this.

helix-view/src/theme.rs
```diff
@@ -56,24 +58,28 @@ impl Loader {
             return Ok(self.base16_default());
         }
 
-        let theme = self.load_theme(name, name, false).map(Theme::from)?;
+        let mut visited_paths = HashSet::new();
```
gibbz00, line 61: Hi, I'm trying to merge this into #5636. What's the point of this `HashSet` introduction?

gibbz00: Nevermind, sorry. Didn't see the new `load_theme` logic.

helix-view/src/theme.rs
```diff
@@ -148,32 +151,37 @@ impl Loader {
...
-            self.default_dir.join(filename)
-        }
+        self.theme_dirs.iter().find_map(|dir| {
+            let path = dir.join(&filename);
+            if path.exists() && !visited_paths.contains(&path) {
+                visited_paths.insert(path.clone());
+                Some(path)
+            } else {
+                None
+            }
```
pascalkuthe, line 175: I think we should track whether there is a cycle here (simply set a bool to true if we return `None` because `visited_paths` already contains `path` so we can create a separate error message for that case. Having the same error message for both cases seems confusing.

pascalkuthe (changes requested):
This looks great! Just some minor UI nits

book/src/install.md
```diff
@@ -108,8 +108,9 @@ RUSTFLAGS="-C target-feature=-crt-static"
 
 
 Helix also needs its runtime files so make sure to copy/symlink the `runtime/` directory into the
-config directory (for example `~/.config/helix/runtime` on Linux/macOS). This location can be overridden
-via the `HELIX_RUNTIME` environment variable.
+config directory (for example `~/.config/helix/runtime` on Linux/macOS). An alternative runtime directory can
+be used by setting the `HELIX_RUNTIME` environment variable. Both runtime directories can be used at the same
+time, with the files residing under the config runtime directory given priority.
```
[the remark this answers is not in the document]
paul-scott: Done. I've also moved it down slightly in the document to try and make the flow make a bit more sense.

helix-term/src/health.rs
```diff
@@ -66,17 +66,27 @@ pub fn general() -> std::io::Result<()> {
...
+            .iter()
+            .map(|d| d.to_string_lossy())
+            .collect::<Vec<_>>()
+            .join(";")
+    )?;
+    for rt_dir in rt_dirs.iter() {
+        if let Ok(path) = std::fs::read_link(&rt_dir) {
+            let msg = format!("Runtime directory is symlinked to: {}", path.display());
+            writeln!(stdout, "{}", msg.yellow())?;
+        }
```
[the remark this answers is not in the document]
paul-scott: Have added both paths to output.

helix-view/src/theme.rs
```diff
@@ -148,32 +151,37 @@ impl Loader {
...
-    /// Lists all theme names available in default and user directory
+    /// Lists all theme names available in all directories
     pub fn names(&self) -> Vec<String> {
-        let mut names = Self::read_names(&self.user_dir);
-        names.extend(Self::read_names(&self.default_dir));
+        let mut names = Vec::new();
+        for dir in &self.theme_dirs {
+            names.extend(Self::read_names(dir));
+        }
         names
```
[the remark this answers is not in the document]
paul-scott: This method is actually no longer used anywhere . I'm not sure if this is due to change in master being merged or if I made the change. Anyway I've removed the method.

helix-view/src/theme.rs
paul-scott: Yep I'm doing that now. This method now returns a Result which distinguishes between the two cases.

book/src/install.md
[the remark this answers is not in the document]
paul-scott: There was a conflict with master. I've now merged in latest master and resolved the conflict.

BTW, what is the preferred process to keep these PRs up to date / conflict free with master? Should we be periodically rebasing and force pushing to our PR branch, or doing a regular merge? I was doing the former to try and keep things cleaner but then I see archseer at some point decided to do a regular merge.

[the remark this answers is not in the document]
archseer: Rebase is preferred but if it's a small PR that we'll squash then it'll result in a single commit anyway

helix-term/src/ui/mod.rs
```diff
@@ -280,10 +280,10 @@ pub mod completers {
     }
 
     pub fn theme(_editor: &Editor, input: &str) -> Vec<Completion> {
-        let mut names = theme::Loader::read_names(&helix_loader::runtime_dir().join("themes"));
-        names.extend(theme::Loader::read_names(
-            &helix_loader::config_dir().join("themes"),
-        ));
+        let mut names = theme::Loader::read_names(&helix_loader::config_dir().join("themes"));
+        for rt_dir in helix_loader::runtime_dirs() {
```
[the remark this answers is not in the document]
paul-scott: Look down a couple of lines to 289 and 290 and it looks like this being done already.

book/src/install.md
```diff
@@ -134,14 +132,21 @@ cd %appdata%\helix
...
+If you plan on keeping the repo locally, an alternative to copying/symlinking runtime files is
+to set `HELIX_RUNTIME=/path/to/helix/runtime` (`HELIX_RUNTIME=$PWD/runtime` if you're in the
+helix repo directory).
+
+When Helix finds multiple runtime directories it will search through them for files in the
+following order:
 
-> NOTE: if `HELIX_RUNTIME` is set prior to calling `cargo install --path helix-term --locked`,
-> tree-sitter grammars will be built in `$HELIX_RUNTIME/grammars`.
+1. `runtime/` sibling directory to `$CARGO_MANIFEST_DIR` directory.
```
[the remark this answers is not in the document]
paul-scott: Sounds good I'll make this change.

helix-term/src/ui/mod.rs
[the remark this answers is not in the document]
pascalkuthe: Ah you are right, sorry about that 😅

pascalkuthe (approved):
LGTM now

archseer:
Thanks! This was a much needed improvement 🎉

book/src/install.md
```diff
@@ -182,13 +184,27 @@ setx HELIX_RUNTIME "%userprofile%\source\repos\helix\runtime"
...
+> Cmd as an administrator.
+
+#### Multiple runtime directories
+
+When Helix finds multiple runtime directories it will search through them for files in the
+following order:
+
+1. `runtime/` sibling directory to `$CARGO_MANIFEST_DIR` directory (this is intended for
+  developing and testing helix only).
+2. `runtime/` subdirectory of OS-dependent helix user config directory.
```
flokli, line 202: Hmm, shouldn't this be  ~/.local/state/helix/runtime (`XDG_STATE_HOME`)? 

This is clearly not config, but state, and we don't want backup programs to back up compiled tree-sitter .so grammars…

the-mikedavis: That's https://github.com/helix-editor/helix/issues/584

This PR kept it to `~/.config` for backwards compatibility

gibbz00: https://github.com/helix-editor/helix/pull/5636 🤷

PR #5418, opened 2023-01-05
New option in language config to use nearest `roots` match

seewishnew:
## Description
This PR adds a new boolean option to language config called `match-nearest-root` to use the nearest match for one of the patterns in `roots` as the LS workspace root. 

Without such an option the LS often returns error responses in languages like Golang/Java where projects can be nested and patterns such as `go.mod` or `pom.xml` can be reused in parent directories. 

## Examples
```toml
[[language]]
name = "go"
scope = "source.go"
injection-regex = "go"
file-types = ["go"]
roots = ["go.mod"]
auto-format = true
comment-token = "//"
language-server = { command = "gopls", args = ["-vv"] }
match-nearest-root = true
```

With the above configuration the nearest directory that has `go.mod` will be picked as LS workspace root. 

Note that the file-picker workspace root is still going to use the old pattern of finding the nearest `.git` directory or else defaulting to the current directory. This option is also set to false by default which makes it backwards-compatible and an opt-in feature.

### DIFF
--- a/helix-core/src/lib.rs
+++ b/helix-core/src/lib.rs
@@ -41,11 +41,16 @@ pub fn find_first_non_whitespace_char(line: RopeSlice) -> Option<usize> {
 /// Find project root.
 ///
 /// Order of detection:
-/// * Top-most folder containing a root marker in current git repository
+/// * Top-most folder containing a root marker in current git repository unless
+///   `match_nearest_root` is true in which case the nearest match is returned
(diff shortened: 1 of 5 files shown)


archseer:
See also https://github.com/helix-editor/helix/pull/4439

seewishnew:
> See also #4439

Ahh I did not see this before--thanks! Closing this PR 👍

Issue #5424, opened 2023-01-06
Clipboard contents lost when `hx` used as scrollback buffer editor within zellij session


gavynriebau:
### Summary

I am using [zellij](https://zellij.dev/) as a terminal multiplexer/session manager (e.g. like `tmux`), below I will provide info related to [zellij](https://zellij.dev/) but please hear me out before closing this issue because I think the bug lies within `hx` itself and not `zellij` and I will explain why.

`Zellij` contains an "edit scrollback buffer" feature whereby one can press `<Ctrl-s>` and it dumps the current scrollback buffer to a temporary file before spawning whatever program you've configured as your scrollback editor (defaults to $EDITOR or $VISUAL contents).

I have found that when I use `hx` as my scrollback buffer editor the contents of the clipboard are lost as described in the bug reproduction steps.

However, when I configure `nvim` as my scrollback buffer editor the copy/paste behaviour works perfectly fine and clipboard contents are not lost.

I have done some investigation and I believe the problem, which only occurs when using `hx` as the scrollback editor, is likely due to differences in how `xclip` is spawned.

Specific differences in how `xclip` is spawned:

- `hx` does _not_ call `setsid` before spawning and calls xclip with the args `xclip -i -selection clipboard`
- `nvim` calls `setsid` before spawning and calls xclip with the args `xclip -quiet -i -selection clipboard`

This is a dump of relevant `htop` output when using `nvim` to do a copy/paste within scrollback editing:

![image](https://user-images.githubusercontent.com/11895736/210927933-54edbe05-e93e-4f27-9b96-0425b3367bc5.png)

Note that the session ID and process group ID of `xclip` is different to that of `nvim`, I believe that is because nvim calls `setsid` before exec'ing `xclip`.

Now here is a dump of relevant `htop` output when using `hx` to do a copy/paste:

![image](https://user-images.githubusercontent.com/11895736/210928142-7dbc3406-975e-453d-b10c-7efc24e7fe56.png)

Note that the session ID and process group ID of `xclip` is the same as that of the spawned `hx`.

My suspicion is that when `hx` is closed the spawned `xclip` process is also closed because they are both in the same session / process group. Pasting then fails because `xclip` has been closed and X11 no longer has a source from which to get the clipboard contents.

**Reasons I believe the bug is within `hx` and not `zellij`:**

1. The same bug does not occur if I configure `nvim` as my scrollback editor
[...]

PR #5426, opened 2023-01-06
Fix for lost clipboard contents (#5424)

gavynriebau:
Fixes the issue described in #5424 

The code has been changed so that for unix type platforms a call is made to `libc::setsid` using the unix specific CommandExt [pre_exec](https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec).

The `setsid` call happens just after `fork` has been called and before `execve` is called. This means it runs in the context of the child process.

The call is necessary to ensure that spawned child processes related to the clipboard (e.g. `xclip`) have a distinct process session ID and process group ID.

Without this change when the parent process exits, the child (i.e. `xclip`), can be erroneously killed leading to copy/paste breaking in some specific circumstances such as those detailed in #5424

### DIFF
--- a/helix-view/Cargo.toml
+++ b/helix-view/Cargo.toml
@@ -22,7 +22,6 @@ helix-lsp = { version = "0.6", path = "../helix-lsp" }
 helix-dap = { version = "0.6", path = "../helix-dap" }
 crossterm = { version = "0.25", optional = true }
 helix-vcs = { version = "0.6", path = "../helix-vcs" }
-libc = "0.2"
 
 # Conversion traits
 once_cell = "1.17"
@@ -48,6 +47,9 @@ which = "4.2"
 
 [target.'cfg(windows)'.dependencies]
 clipboard-win = { version = "4.5", features = ["std"] }
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2"
 
 [dev-dependencies]
 helix-tui = { path = "../helix-tui" }
--- a/helix-view/Cargo.toml
+++ b/helix-view/Cargo.toml
@@ -22,6 +22,7 @@ helix-lsp = { version = "0.6", path = "../helix-lsp" }
 helix-dap = { version = "0.6", path = "../helix-dap" }
 crossterm = { version = "0.25", optional = true }
 helix-vcs = { version = "0.6", path = "../helix-vcs" }
+libc = "0.2"
 
 # Conversion traits
 once_cell = "1.17"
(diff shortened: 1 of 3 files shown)


helix-view/src/clipboard.rs
```diff
@@ -276,12 +276,27 @@ pub mod provider {
...
+                if cfg!(not(any(windows, target_os = "wasm32", target_os = "macos"))) {
+                    use std::os::unix::process::CommandExt;
+
+                    unsafe {
+                        command_mut = command_mut.pre_exec(|| match libc::setsid() {
+                            -1 => Err(std::io::Error::last_os_error()),
+                            _ => Ok(()),
+                        });
+                    }
+                }
```
archseer, line 297: `setsid` should be called on all unix systems:

```suggestion
                #[cfg(unix)]
                unsafe {
                    use std::os::unix::process::CommandExt;
                    command_mut = command_mut.pre_exec(|| {
                        if libc::setsid() == -1 {
                            return Err(io::Error::last_os_error());
                        }
                    });
                }
```

For windows we might also want to pass `.creation_flags(CREATE_NEW_PROCESS_GROUP)`

archseer (changes requested):

helix-view/src/clipboard.rs
gavynriebau: I've made the following changes:
- Call `setsid` for all unix systems
- Call `.creation_flags(CREATE_NEW_PROCESS_GROUP)` for windows

After making the changes I've realised the change to set `CREATE_NEW_PROCESS_GROUP` will never be executed as it is inside a `mod provider` that uses the config `#[cfg(not(target_os = "windows"))]`.

Seeking your opinion but I am thinking of reverting back by one commit so that the PR will contain no code related to setting `CREATE_NEW_PROCESS_GROUP` for windows but keeps the change you suggested of calling `setsid` for all unix systems?

Separate to this PR but I'm also considering doing a follow-up PR to cleanup the code a little, let me know if you think this would be valuable?
With the amount of conditional compilation attributes I found it slightly hard to follow what was going on andwas thinking of splitting the code out into separate files for each environment, e.g. a provider file for windows, macos, osc52, wasm, unix etc..

helix-view/Cargo.toml
```diff
@@ -22,6 +22,7 @@ helix-lsp = { version = "0.6", path = "../helix-lsp" }
 helix-dap = { version = "0.6", path = "../helix-dap" }
 crossterm = { version = "0.25", optional = true }
 helix-vcs = { version = "0.6", path = "../helix-vcs" }
+libc = "0.2"
```
archseer, line 25: Should this be listed under `[target.'cfg(unix)'.dependencies`? It probably doesn't make a difference

gavynriebau: Thanks, yes, you're right, it's a unix only dependency. I've addressed this now.

helix-view/src/clipboard.rs
gavynriebau: I think these changes are ready for another review now.

I'm not able to change how `pre_exec` is called to exactly as you suggested because it needs to return a `Result<()>` type.

archseer (approved):

PR #5435, opened 2023-01-06
add option to enforce newline at the end of document

### LINKED ISSUE #4274: Toggle to enforce a newline at the end of a document
I suggest adding a toggle `eol` to enforce a newline at the end of a document.

Xalfer:
resolves #4274


helix-view/src/editor.rs
```diff
@@ -633,6 +633,7 @@ impl Default for Config {
             bufferline: BufferLine::default(),
             indent_guides: IndentGuidesConfig::default(),
             color_modes: false,
+            newline_at_eof: false,
```
filipdutescu, line 636: maybe you could rename it to a shorter name like?
```suggestion
            eof: false,
```

also could you add a doc comment?

filipdutescu (commented):
I have also worked on this, but you opened before me the PR, so I wanted to drop some suggestions, hope you won't mind. Also, you should update the book doc as well (you can take a look at [my change](https://github.com/filipdutescu/helix/tree/feat/enforce_newline_cfg_option) to see where).

helix-view/src/editor.rs
Xalfer: Thanks for sending your changes, I used them to clean up the code a bit since I'm not all that well familiar with the helix codebase. But I don't think that the option needs to be renamed, it's not that long and `eof` is not really descriptive for what is does.

filipdutescu: Hmm that's true, `eof` is vague at best. Glad you found what I did useful :grin:

LeoniePhiline: @Xalfer You could consider `eof-newline`, where `eof` is the general topic, and `newline` is the detail applied to that topic.

pascalkuthe (changes requested):
I think this is a nice option to have. I do find myself editing EOF chars often which show up as changes in git and mess with some tooling. This PR doesn't change the core editing model of helix (all newline remain selecatble and editable) and only acts on save so it's a small non-intrusive option that I think would be a nice addition. I added some comments about the implementation

helix-view/src/editor.rs
```diff
@@ -1328,7 +1329,29 @@ impl Editor {
...
+                .get_slice((doc_len - newline.len())..)
+                .and_then(|s| s.as_str())
+                .filter(|s| s != &newline)
+                .is_some()
+        {
+            let newline = Tendril::from(newline);
+            let selection = Selection::point(doc_len);
+            let transaction = Transaction::insert(text, &selection, newline);
+            let view = view_mut!(self);
+            doc.set_selection(view.id, selection);
```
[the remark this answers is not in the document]
Xalfer: Thanks for the heads up. I fixed this by saving the current selection applying the change and then reapplying the old selection.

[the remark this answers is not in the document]
pascalkuthe: You don't need to touch the selection here at all. Just remove all calls to `set_selection`.

helix-view/src/editor.rs
```diff
@@ -1419,7 +1421,25 @@ impl Editor {
...
+        if config.newline_at_eof && get_line_ending(&text.slice(..)).is_none() {
+            let view = view_mut!(self);
+            let old_selection = doc.selection(view.id).clone();
+            let newline = Tendril::from(newline);
+            let selection = Selection::point(doc_len);
+            let transaction = Transaction::insert(text, &selection, newline);
+            doc.set_selection(view.id, selection);
+            doc.apply(&transaction, view.id);
+            doc.append_changes_to_history(view);
+            doc.set_selection(view.id, old_selection);
```
pascalkuthe, line 1440: I guess you missed my comment https://github.com/helix-editor/helix/pull/5435/files#r1100873864 here since you resolved the conversation too early. You don't need to call `set_selection` here at all. Just remove all changes to the selection.

divarvel, line 1440:
```suggestion
            doc.apply(&transaction, view.id);
            doc.append_changes_to_history(view);
```

I have applied the same changes in https://github.com/helix-editor/helix/commit/ab0c01235ae75d612867080db26c067e9c233dcb if that's simpler to use.

dead10ck (changes requested):
I agree with @pascalkuthe here: we shouldn't mess with the selection at all. But otherwise this change seems fine.

dead10ck (approved):
This looks good to me. Thanks!

Xalfer:
Sorry for not doing it earlier I ~~was lazy~~ didn't find the time to do it. Anyway the whole thing with the selections shouldn't have happened in the first place, but I, for some reason, thought that the selection needed to be applied since Helix is so selection oriented.

helix-view/src/editor.rs
```diff
@@ -1419,7 +1421,22 @@ impl Editor {
...
         let path = path.map(|path| path.into());
+        let config = self.config();
         let doc = doc_mut!(self, &doc_id);
+
+        // handle adding newline at the end of the file if there is none
+        let newline = doc.line_ending.as_str();
+        let text = doc.text();
+        let doc_len = text.len_chars();
+        if config.newline_at_eof && get_line_ending(&text.slice(..)).is_none() {
+            let view = view_mut!(self);
```
the-mikedavis, line 1432: This will fetch the currently focused view which may not be the view that's looking at the document. This can cause a panic with this feature enabled: open a split with two different files without newline endings and `:wa`.

I think it might make sense to solve this in `helix-term::commands::typed` in `write_impl` and `write_all_impl` rather than here in Editor. We probably also do not want to add a trailing newline when documents are being auto-formatted because we may end up fighting with the auto-formatter (language server or external command)

the-mikedavis (changes requested):

PR #5420, opened 2023-01-06
rework positioning/rendering and enable softwrap/virtual text

pascalkuthe:
Supersedes #5008

This PR is a large rework of the core text positioning and rendering code to
remove the assumption that on-screen columns/lines correspond to text columns/
lines. This is required to implement both softwrap and virtual text.

Both of these features are included in this PR, so the original goals outlined
in #5008 can be achieved with this PR.

## Implementation

This PR introduces generic `DocFormatter` that positions graphemes from various
sources. This formatter is used both for rendering and for movements/scrolling.
The design is heavily inspired by led, although I streamlined/adjusted some
Details as helix needs are different (especially with the introduction of
virtual text).

The basics of the softwrap implementation from #5008 mostly remains (so indent
carry over, and word splitting are supported) but rendering and positioning are
cleanly separated now and much of the architecture explained in #5008 is not
present anymore as I had to evolve the architecture for the positioning code.

The `DocFormatter` is still a single grapheme iterator (so more efficient)
that transverses the entire document. However, it does not deal with syntax
highlighting or rendering at all. It's an iterator that is only concerned witch placing
graphemes from the document and virtual text at the correct position.
Note that the `DocFormatter` does not deal with the contents of line virtual text
as that is easily handeled trough the new LineDecoration rendering mechansim.

The `DocFormatter` is constructed from an anchor char index which it backtracks
to the last known line breaks. These "known" line breaks are called blocks and
were also described by @cessen in #136. Right now only normal line breaks are
considered as blocks, but the code is written so long lines can be chunked for
improved performance for files with very long text in the future.
I did not implement this logic in this PR as it's only an optimization and can be
implemented later as suggested by @cessen.

I added the functions `visual_offset_from_anchor`/`visual_offset_from_block`
and `char_idx_at_visual_offset` to the `position` module in helix core
that utilize the `DocFormatter`. These replace `visual_coords_at_pos` and
`pos_at_visual_cords` respectively. Their names are different to more accurately
reflect their functions. The old functions have been replaced everywhere except
for the `align_selections` command. I am not sure if virtual text and softwrap
should affect selection alignment and refactoring this function was more
involved, so I left it untouched for now.

I have also updated the vertical movement command to be softwrap aware which
feels much more intuitive than the behavior in other editor like nvim. If there
is demand for vertical movements that ignore softwrap a separate command could
be added that simply sets `softwrap = false` in the `TextFormat` before passing
it to `move vertically`. Note that many commands (like goto_line_end) are not softwrap
[...]

PR #5436, opened 2023-01-07
Make mouse click extend selection in select mode

### LINKED ISSUE #5425: Mouse click should extend selections in select mode
Currently a mouse click on buffer always sets the selection to be the single char being clicked. IMO it would be a much better UX if the selection gets extended instead when the mode is set to select mode.

xJonathanLEI:
Resolves #5425. Design rationale detailed in #5425 comments.

I'm not entirely sure whether this change is desired, but am still submitting this optimistically as I think it makes a lot of sense. I use Helix on a touch screen device (phone) sometimes and find this extremely useful for selecting code blocks as touches are easier to do than key strokes on such devices. It annoys me that currently the only way to extend a selection would be to use a keyboard.

This PR should improve the mouse UX. It's now possible to extend the primary selection with mouse clicks. The original drag-to-yank behavior is also retained.

With this PR, the editor no longer collapses all selections into a single point on mouse clicks unconditionally. Instead, it extends the primary selection to the point click when it's in select mode (secondary selections are still discarded for consistency). Normal mode behavior is left untouched.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1178,7 +1178,7 @@ impl EditorView {
 
                 let (view, doc) = current!(cxt.editor);
 
-                let should_yank = match cxt.editor.mouse_down_range {
+                let should_yank = match cxt.editor.mouse_down_range.take() {
                     Some(down_range) => doc.selection(view.id).primary() != down_range,
                     None => {
                         // This should not happen under normal cases. We fall back to the original
@@ -1190,8 +1190,6 @@ impl EditorView {
                             > 1
                     }
                 };
-
-                cxt.editor.mouse_down_range = None;
 
                 if should_yank {
                     commands::MappableCommand::yank_main_selection_to_primary_clipboard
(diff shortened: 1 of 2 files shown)


xJonathanLEI:
Conflicts with #5420. Rebased.

xJonathanLEI:
Regarding the `R-breaking-change` label, I'd actually argue that the current behavior of not extending the selection in select mode is a bug :)

helix-term/src/ui/editor.rs
```diff
@@ -1169,19 +1178,28 @@ impl EditorView {
...
-                if doc
-                    .selection(view.id)
-                    .primary()
-                    .slice(doc.text().slice(..))
-                    .len_chars()
-                    <= 1
-                {
-                    return EventResult::Ignored(None);
-                }
+                let should_yank = match cxt.editor.mouse_down_range {
```
the-mikedavis, line 1181:
```suggestion
                let should_yank = match cxt.editor.mouse_down_range.take() {
```

and then the `cxt.editor.mouse_down_range = None` can be eliminated below

the-mikedavis (commented):
Sorry for the delay - I've been meaning to look at this for a while. The change itself looks good and the behavior feels right. I just have one small nit

helix-term/src/ui/editor.rs
xJonathanLEI: Ah TIL that `Option::take()` exists. Thanks! It's been fixed.

the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

PR #5438, opened 2023-01-07
Adding '--locked'

carstein:
By default `cargo install` ignores `Cargo.lock` file.
This lead to a problem where ignore 0.4.19 didn't compile (was using unstable feature of Rust 1.61)
This change make an update in the 'Install from source'  forcing cargo to respect the aforementioned file.


the-mikedavis (commented):
There's another reference to the `cargo install` command that should be updated on L130 as well

carstein:
Right, my bad.
Fixed now.

the-mikedavis (approved):
Thanks!

PR #5440, opened 2023-01-07
Sonokai theme: style secondary selections differently

jlebon:
Without styling the primary and secondary selections differently, it's
impossible to tell them apart when cycling through selections.

Make the primary selection slightly brighter and secondary selections
slightly paler.

### DIFF
--- a/runtime/themes/sonokai.toml
+++ b/runtime/themes/sonokai.toml
@@ -55,7 +55,8 @@
 "ui.cursor.match" = { fg = "orange", bg = "diff_yellow" }
 "ui.cursor.insert" = { fg = "black", bg = "grey" }
 "ui.cursor.select" = { fg = "bg0", bg = "blue" }
-"ui.selection" = { bg = "bg4" }
+"ui.selection" = { bg = "bg5" }
+"ui.selection.primary" = { bg = "bg4" }
 "ui.linenr" = "grey"
 "ui.linenr.selected" = "fg"
 "ui.cursorline.primary" = { bg = "bg1" }
@@ -86,7 +87,8 @@ bg0 = "#2c2e34"
 bg1 = "#33353f"
 bg2 = "#363944"
 bg3 = "#3b3e48"
-bg4 = "#545862"
+bg4 = "#5C606A"
+bg5 = "#444852"
 bg_red = "#ff6077"
 diff_red = "#55393d"
 bg_green = "#a7df78"


the-mikedavis:
\cc @p4ymak what do you think of the new color?

jlebon:
I suppose I should've included a screenshot. Here's what it looks like:

image

In the above, the third `test` is the primary selection. The other ones use the paler secondary styling. It's subtle but really obvious when cycling through them.

Omnikar:
That seems like a pretty subtle difference; shouldn't it be a bit more noticeable?

jlebon:
The thing is that it's already quite pale. If we make it paler, it's harder to see against the background. It's more invasive, but we could also make the primary selection whiter to increase the contrast. E.g.:

image

This is with:

```diff
diff --git a/runtime/themes/sonokai.toml b/runtime/themes/sonokai.toml
index 18417f00..45ad6ef6 100644
--- a/runtime/themes/sonokai.toml
+++ b/runtime/themes/sonokai.toml
@@ -55,7 +55,8 @@
 "ui.cursor.match" = { fg = "orange", bg = "diff_yellow" }
 "ui.cursor.insert" = { fg = "black", bg = "grey" }
 "ui.cursor.select" = { fg = "bg0", bg = "blue" }
-"ui.selection" = { bg = "bg4" }
+"ui.selection" = { bg = "bg5" }
+"ui.selection.primary" = { bg = "bg4" }
 "ui.linenr" = "grey"
 "ui.linenr.selected" = "fg"
 "ui.cursorline.primary" = { bg = "bg1" }
@@ -86,7 +87,8 @@ bg0 = "#2c2e34"
 bg1 = "#33353f"
 bg2 = "#363944"
 bg3 = "#3b3e48"
-bg4 = "#545862"
+bg4 = "#5C606A"
+bg5 = "#444852"
 bg_red = "#ff6077"
 diff_red = "#55393d"
 bg_green = "#a7df78"
```

archseer:
> It's more invasive, but we could also make the primary selection whiter to increase the contrast.

This seems reasonable to me 👍🏻

NickCondron:
I also have a PR to change this theme (#5379) fyi. I don't think they conflict though.

p4ymak:
@the-mikedavis 
sorry for long respond. 

in my opinion, in general, the palette has become better. there are only two things I don't like: macros are no different from functions now, and parentheses and punctuation are too dim.

jlebon:
Updated this as per discussion!

> in my opinion, in general, the palette has become better. there are only two things I don't like: macros are no different from functions now, and parentheses and punctuation are too dim.

Did you mean to comment this in [the other PR](https://github.com/helix-editor/helix/pull/5379) modifying the Sonokai theme?

p4ymak:
> Did you mean to comment this in [the other PR](https://github.com/helix-editor/helix/pull/5379) modifying the Sonokai theme?

Yes! You are right.

the-mikedavis (approved):

PR #5468, opened 2023-01-09
Fix piping to Helix on macOS

### LINKED ISSUE #2111: Helix crashes when launched from `zk` application
### Summary

When [`zk`](https://github.com/mickael-menu/zk) tries to run helix, helix crashes with error status 101. 

I wanted to add helix as default editor for [`zk`](https://github.com/mickael-menu/zk). For fresh install I set editor in [`zk` config](https://github.com/mickael-menu/zk/blob/main/docs/tool-editor.md) but 

### Reproduction Steps

 

I tried this:

1. Fresh install of  [`zk`](https://github.com/mickael-menu/zk)
2. Set `editor = "hx"` in config per [config documentation](https://github.com/mickael-menu/zk/blob/main/docs/tool-editor.md)
3. Invoke any action in `zk` that should cause launch of helix, for example `zk new` 

I expected this to happen:
Helix to launch

Instead, this happened:
Helix crashes with:
`zk: error: failed to launch editor: hx /Users/<snip>/5e19.md: exit status 101`

Here is the code of `zk` that [launches editor](https://github.com/mickael-menu/zk/blob/50855154e2f2878994e7a99e2ec2619addaeee90/internal/adapter/editor/editor.go#L41). It works for me with `nano` and `vim`

Log attached that corresponds with launching running `zk new`.

### Helix log

```
2022-04-14T21:22:22.528 helix_view::editor [ERROR] Failed to initialize the LSP for `source.md` { LSP not defined }
```

</details>

### Platform

macOS

### Terminal Emulator

iTerm

### Helix Version

helix 22.03 (d4e45fd4)

### LINKED ISSUE #6734: Panic when Edits the current line in the $EDITOR program
### Summary

After I set `hx` as the default editor in `zsh`, I got panic when trying to edit the command with [ctrl-x ctrl-e](https://unix.stackexchange.com/questions/696136/what-exactly-handles-the-ctrlx-shortcut-family-e-g-ctrlx-ctrle-in-linux-te#:~:text=Ctrl%20x%20Ctrl%20e%20%3A%20Edits,or%20variable%20assignments%20found%20there.)

### Reproduction Steps

 

I tried this:

1.  export EDITOR='hx' in `~/.zshrc`
2. ctrl-x ctrl-e in my terminal

I expected this to happen:
Helix opened, put what I already typed in the terminal into buffer.

Instead, this happened:
```
➜  ~ thread 'main' panicked at 'reader source not set', /Users/brew/Library/Caches/Homebrew/cargo_cache/registry/src/github.com-1ecc6299db9ec823/crossterm-0.26.1/src/event/read.rs:38:30
```

### Helix log

```
2023-04-13T13:34:40.867 helix_vcs [ERROR] Error {
    context: "failed to open git repo",
    source: Discover(
        NoGitRepository {
            path: "/private/tmp",
        },
    ),
}
2023-04-13T13:34:40.867 helix_vcs [ERROR] failed to open diff base for /private/tmp/zshtIXKYX
2023-04-13T13:34:40.867 helix_vcs [ERROR] Error {
    context: "failed to open git repo",
    source: Discover(
        NoGitRepository {
            path: "/private/tmp",
        },
    ),
}
2023-04-13T13:34:40.867 helix_vcs [ERROR] failed to obtain current head name for /private/tmp/zshtIXKYX```

</details>


### Platform

macOS

### Terminal Emulator

wezterm

### Helix Version

helix 23.03

### PR DESCRIPTION
**Pull request by yyogo**
Crossterm will soon merge https://github.com/crossterm-rs/crossterm/pull/735 which will let it use /dev/tty on macOS. This will fix #2111 and allow piping to Helix on macOS.

Fixes #6734

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -37,7 +37,7 @@ which = "4.4"
 
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "io-util", "io-std", "time", "process", "macros", "fs", "parking_lot"] }
 tui = { path = "../helix-tui", package = "helix-tui", default-features = false, features = ["crossterm"] }
-crossterm = { version = "0.26", features = ["event-stream", "use-dev-tty"] }
+crossterm = { version = "0.26", features = ["event-stream"] }
 signal-hook = "0.3"
 tokio-stream = "0.1"
 futures-util = { version = "0.3", features = ["std", "async-await"], default-features = false }
@@ -69,6 +69,9 @@ grep-searcher = "0.1.11"
 [target.'cfg(not(windows))'.dependencies]  # https://github.com/vorner/signal-hook/issues/100
 signal-hook-tokio = { version = "0.3", features = ["futures-v0_3"] }
 libc = "0.2.146"
+
+[target.'cfg(target_os = "macos")'.dependencies]
+crossterm = { version = "0.26", features = ["event-stream", "use-dev-tty"] }
 
 [build-dependencies]
 helix-loader = { version = "0.6", path = "../helix-loader" }
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -37,7 +37,7 @@ which = "4.4"
 
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "io-util", "io-std", "time", "process", "macros", "fs", "parking_lot"] }
 tui = { path = "../helix-tui", package = "helix-tui", default-features = false, features = ["crossterm"] }
-crossterm = { version = "0.26", features = ["event-stream"] }
+crossterm = { version = "0.26", features = ["event-stream", "use-dev-tty"] }
 signal-hook = "0.3"
 tokio-stream = "0.1"
 futures-util = { version = "0.3", features = ["std", "async-await"], default-features = false }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -239,6 +239,7 @@ checksum = "a84cda67535339806297f1b331d6dd6320470d2a0fe65381e79ee9e156dd3d13"
 dependencies = [
  "bitflags 1.3.2",
  "crossterm_winapi",
+ "filedescriptor",
  "futures-core",
  "libc",
  "mio",
@@ -389,6 +390,17 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d9f0c14694cbd524c8720dd69b0e3179344f04ebb5f90f2e4a440c6ea3b2f1ee"
 dependencies = [
  "log",
+]
+
+[[package]]
+name = "filedescriptor"
+version = "0.8.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7199d965852c3bac31f779ef99cbb4537f80e952e2d6aa0ffeb30cce00f4f46e"
+dependencies = [
+ "libc",
+ "thiserror",
+ "winapi",
 ]
 
 [[package]]
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -215,11 +215,6 @@ impl Application {
             }
         } else if stdin().is_tty() || cfg!(feature = "integration") {
             editor.new_file(Action::VerticalSplit);
-        } else if cfg!(target_os = "macos") {
-            // On Linux and Windows, we allow the output of a command to be piped into the new buffer.
-            // This doesn't currently work on macOS because of the following issue:
-            //   https://github.com/crossterm-rs/crossterm/issues/500
-            anyhow::bail!("Piping into helix-term is currently not supported on macOS");
         } else {
             editor
                 .new_file_from_stdin(Action::VerticalSplit)

### CODE REVIEW

**Comment by kirawi** [comment]
Waiting on a crossterm release with the meged PR.

**Comment by archseer** [comment]
What are the performance implications here? I figured the mio one is probably better optimized

**Comment by yyogo** [comment]
> What are the performance implications here? I figured the mio one is probably better optimized


Fwiw, I haven't noticed any performance differences in Helix or other projects using crossterm when testing manually. 

I did mean to do a comparative benchmark for the crossterm PR, but gave up since I couldn't figure out how to do that without emulating the controlling terminal. If there are any existing benchmarks for helix, or if you know of a similar suite in another project, I could try to run it and report what I find.

**Comment by archseer** [comment]
A flame graph before/after the change would be useful. As a really simple test you could try opening a 10MB text file and scrolling from top to bottom. Measuring the flame graph there + time taken would be at least one data point

**Comment by danriedl** [comment]
What is left to merge this PR? The test you mentioned @archseer ?

**Comment by David-Else** [comment]
@danriedl I think there is talk of using a custom fork as a temporary measure for some essential fixes before 0.26.1 https://github.com/helix-editor/helix/pull/4939

**Comment by the-mikedavis** [comment]
I scrolled through [Moby Dick](https://www.gutenberg.org/files/2701/2701-0.txt) with touchpad scrolling as fast as possible with this and master at 1a87d14439bc940d9bf3e66359a612b345aa363f

The flamegraphs look pretty similar and they both took around 35s. Master:

master

This PR:

**Comment by pascalkuthe** [comment]
> 

Actually crossterm does look like its taking up more space in that flamegraph although not by a huge amount.

Couldn't we just turn this feature on for macos and keep it off for other platforms? Platform specific feature is possible with cargo I think

**Comment by archseer** [comment]
The one by this PR actually looks shorter though? There's some mentions about enabling this feature and dropping `mio` so it might not make sense to have it conditionally enabled in the long run

**Comment by yyogo** [comment]
@the-mikedavis sorry I couldn't get to this so far, thank you for taking the time. Please see https://github.com/crossterm-rs/crossterm/issues/755  - I think we should hold off on this pr until it is fixed (I'm working on it).

**Comment by the-mikedavis** [comment]
Looks like there's a new crossterm 0.26.1 release with your fix 🚀 https://github.com/crossterm-rs/crossterm/releases/tag/0.26.1

**Comment by pascalkuthe** [comment]
> The one by this PR actually looks shorter though? There's some mentions about enabling this feature and dropping `mio` so it might not make sense to have it conditionally enabled in the long run

Hmm yeah I saw the same difference but switched the two plots up by accident. This seems very surprising to me tough. `mio` is really well optimized (and we depend on it anyway trough tokio). If I understood the Pr that integrates tis into crossterm right there is some trickery going on to create a waker. I would have assumed that caused overhead. Just out of curiosity. On which operating system was that flamegraph recorded @the-mikedavis?

**Comment by the-mikedavis** [comment]
That was a macOS laptop. I'll take some linux ones too for comparison

**Comment by pascalkuthe** [comment]
Thanks for the benchmarks! Interesting how different linux and mac are. Is it possible you switched the flamegraphs  up in your earlier comment? It looks like this PR is a slight performance regression on both macos on linux from the new flamegraphs you posted. The new macos benchmarks look similar to your earlier benchmark but swapped and now looks roughly what I would have gussed.

It doesn't seem like crossterm will switch away from mio immedietly, the followup PR (https://github.com/crossterm-rs/crossterm/pull/743) still has everything behind a flag (and doesn't drop mio from the dependencies either). I  think for now keeping it behind a cfg would make sense as its a performance regression and needs one extra dependency. It doesn't matter much either way tough.

**Comment by the-mikedavis** [comment]
Ah yeah, they weren't recorded very scientifically either, just swipe scrolling 😅. I don't have the original files around so I can't be sure

**Comment by heliostatic** [comment]
I'm running into this crash on a Mac with crossterm 0.26.1 (so with @yyogo 's crossterm patch). How do I properly configure my build of helix to use tty? Right now I'm crashing trying to use `fzf` to find an execute helix, e.g. `fzf --bind 'enter:become(hx {})'`

**Comment by pascalkuthe** [comment]
> I'm running into this crash on a Mac with crossterm 0.26.1 (so with @yyogo 's crossterm patch). How do I properly configure my build of helix to use tty? Right now I'm crashing trying to use `fzf` to find an execute helix, e.g. `fzf --bind 'enter:become(hx {})'`

this PR wasn't merged yet. Crossterm doesn't use dev-tty unless you explicitly set a feature flag

**Comment by diegs** [comment]
@the-mikedavis as you pointed out crossterm 0.26.1 was released a few months ago with the crash fix. What are the remaining decision criteria for switching to the `dev-tty` feature, at least on MacOS? Are the performance implications too concerning? Thanks!

**Comment by yyogo** [comment]
We could also enable the feature only on macos, I would think the usability improvement is worth a slight performance regression.

File: helix-term/Cargo.toml [hunk 1]
```diff
@@ -70,6 +69,11 @@ grep-searcher = "0.1.11"
 signal-hook-tokio = { version = "0.3", features = ["futures-v0_3"] }
 libc = "0.2.146"
 
+[target.'cfg(not(target_os = "macos"))'.dependencies]
+crossterm = { version = "0.26", features = ["event-stream"] }
+[target.'cfg(target_os = "macos")'.dependencies]
+crossterm = { version = "0.26", features = ["event-stream", "use-dev-tty"] }
```
pascalkuthe, line 75: you don't need the `cfg(not)` case here cargo will automatically unify features (you ofcourse need to keep the existing dependency unchanged them

```suggestion
[target.'cfg(target_os = "macos")'.dependencies]  
crossterm = { version = "0.26", features = ["use-dev-tty"] }
```

pascalkuthe (approved):

archseer (approved):

PR #5471, opened 2023-01-09
jsonrpc: Skip serializing params if params are None

### LINKED ISSUE #5400: Quitting Helix when `ocamllsp` was active causes delay and exits with an error.
When quitting the editor from an OCaml source file with `ocamllsp` active, there is a noticeable delay before exiting with the following error:

```
Error: Timed out waiting for language servers to shutdown
```

the-mikedavis:
The JSONRPC spec says:

> If present, parameters for the rpc call MUST be provided as a Structured value

https://www.jsonrpc.org/specification#parameter_structures

(Where a "Structured value" is elsewhere defined as either a map or array.)

This change skips the serialization of the `params` field for JSONRPC method calls and notifications if the `params` field is the `None` variant. This fixes compatibility with LSP servers which adhere closely to that part of the spec: `ocamllsp` in the wild.

Fixes #5400

### DIFF
--- a/helix-lsp/src/jsonrpc.rs
+++ b/helix-lsp/src/jsonrpc.rs
@@ -170,6 +170,10 @@ impl Params {
         serde_json::from_value(value)
             .map_err(|err| Error::invalid_params(format!("Invalid params: {}.", err)))
     }
+
+    pub fn is_none(&self) -> bool {
+        self == &Params::None
+    }
 }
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Looks like this was noticed on the upstream we vendored from as well: https://github.com/paritytech/jsonrpc/issues/660

PR #5482, opened 2023-01-10
Update rust-toolchain.toml

aalhendi:
Closes #5481 
Feature `scoped_threads` was stabilized in Rust version 1.63.0. Building from source using Rust version 1.61.0 fails.
Reference: https://doc.rust-lang.org/std/thread/fn.scope.html

PR #5485, opened 2023-01-10
bumping rust compiler version because of scoped threads

mangas:


pascalkuthe:
The current rust version is on purpose. Simply compile wth `--locked` to avoid problems.
See https://github.com/helix-editor/helix/pull/5482#issuecomment-1376971973.

We will bump the rustc version once the next firefox version releases (jan 17.)

mangas:
got it, thanks! 👍

PR #5489, opened 2023-01-10
Rosé Pine theme: use official variants

mvllow:
Thanks to the lovely community, we noticed our themes were included in Helix very early on. This PR updates the variants to match our current spec and palette.

We hope contributors make a PR at https://github.com/rose-pine/helix for any theme changes but understand this may not always be the case. For that reason, our Moon and Dawn variants inherit from the Main variant to try and keep them in sync as much as possible.

Please let me know if anything else is needed, and thank you 💜


mvllow:
After running `cargo xtask themelint rose_pine` I got the following log:

```
rose_pine: `ui.selection` and `ui.selection.primary` cannot be equal
rose_pine: missing `fg` for `ui.virtual.whitespace`
rose_pine: missing `fg` and `bg` for `ui.help`
rose_pine: missing `fg` or `bg` for `ui.window`
```

Will update these asap :)

mvllow:
Lint errors have been fixed for `rose_pine` but running it on `rose_pine_moon` or `rose_pine_dawn` causes several errors due to these variants not explicitly setting any theme values but rather inheriting from `rose_pine`. Is this okay or should I go ahead and make each variant independent from one another?

the-mikedavis:
For those you can ignore the errors: I would prefer that they use `inherits` rather than pass the themelint

archseer:
Sorry, slight conflict now since a previous PR changed the diagnostics to use curly underlines.

Thank you for maintaining these now officially upstream!

cor:
I like how this PR will ensure that the new curl underlines that I applied to only the normal variant will also work for the other variants 👍🏻

the-mikedavis (approved):

PR #5491, opened 2023-01-10
Add statusline element 'file-indexer'

mejo13:
Shows number of open files and index of the one currently viewed.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -100,7 +100,7 @@ The following statusline elements can be configured:
 | `file-base-name` | The basename of the opened file |
 | `file-encoding` | The encoding of the opened file if it differs from UTF-8 |
 | `file-line-ending` | The file line endings (CRLF or LF) |
-| `file-indexer` | The number of open files and current index |
+| `file-count` | The number of open files and current index |
 | `total-line-numbers` | The total line numbers of the opened file |
 | `file-type` | The type of the opened file |
 | `diagnostics` | The number of warnings and/or errors |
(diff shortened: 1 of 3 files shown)


CBenoit (commented):
Thank you! I think you need to update the documentation as well:
https://github.com/helix-editor/helix/blob/051cd786a43a84912413b526782adffa9698fcd9/book/src/configuration.md?plain=1#L93

archseer:
Hmm I think we could come up with a better name for it. Indexer sounds like a background service that indexes files

mejo13:
Maybe file_counter would be more fitting or do you have an idea?

the-mikedavis:
FileNumber or FileCount would make sense to me. I think we should avoid naming things with [agent noun](https://en.wikipedia.org/wiki/Agent_noun)s (index<i>er</i>, count<i>er</i>) except where there's a lot of precedent for it (picker, receiver, server)

mejo13:
I'd go with FileCount then, if that's ok!?

mejo13:
What is the preferred way to do this? Change the existing pr, which would make the discussion about the name kinda pointless, or make a new one? I'm pretty new to git and github so i'm realy not sure.

CBenoit:
> What is the preferred way to do this? Change the existing pr, which would make the discussion about the name kinda pointless, or make a new one? I'm pretty new to git and github so i'm realy not sure.

Feel free to commit and push additional commits to apply the suggestions. Unless you push force, original commit history will stay available from this PR. That being said, it’s generally okay to just push force if you wish to get the commits merged as it. Maintainers will just squash and merge if necessary, so feel free to proceed as you prefer in this case.

mejo13:
Hey, I was wondering if I have to change something else, request a review or is this better suited as a plugin once the system is in place maybe?

pascalkuthe:
I am going to close this one out as stale since the statusline has been refactored. I think this one is a bit niche and better suited for a plugin. Sorry we didn't give you that feedback sooner and thank you for contributing!

Issue #4699, opened 2022-11-10
Implement `window/showMessageRequest` in LSP client


ckipp01:
Some servers will make use of this. You can see it in the spec [here](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessageRequest). An example of a server that uses this is [Metals](https://scalameta.org/metals/), which is the language server for Scala. We use this when a user first opens up a workspace to prompt them to import the workspace. A request looks like this:

```json
[Trace - 03:48:41 pm] Sending request 'window/showMessageRequest - (2)'
Params: {
  "actions": [
    {
      "title": "Import build"
    },
    {
      "title": "Not now"
    },
    {
      "title": "Don\u0027t show again"
    }
  ],
  "type": 3,
  "message": "New sbt workspace detected, would you like to import the build?"
}
```

Having support for this would be great. There's also a handful of other usecases with Metals for this requestt.

jba:
I'm having the same issue with `gopls`, which is trying to tell me about a bad config:
```
helix_term::application [WARN] unhandled window/showMessage: ShowMessageParams { typ: Error, message: "Invalid settings: unexpected gopls setting \"moose\"; unexpected gopls setting \"ui\"" }
```

abrassel:
Hey, I wanted to jump in on this issue - I would love to do the remaining work to get this PR merged.

wzzrd:
I would soooo love to get the functionality merged into helix... I'm a heavy user of ansible-language-server, which doesn't currently work with helix because of this :( 

More than happy to do some testing is that helps?

roignpar:
I am also running into an issue related to this with the F# LSP ([fsautocomplete](https://github.com/ionide/FsAutoComplete)):
```
2025-05-31T22:41:44.931 helix_term::application [ERROR] Language Server: Method window/showMessageRequest not found in request 22
2025-05-31T22:41:44.932 helix_lsp::transport [ERROR] fsharp-ls err <- "[22:41:44.931 ERR] [FsAutoComplete.Lsp.AdaptiveFSharpLspServer] HandleFormatting Request Errored\n"
2025-05-31T22:41:44.932 helix_lsp::transport [ERROR] fsharp-ls err <- "System.AggregateException: One or more errors occurred. (Method not found: window/showMessageRequest)\n"
[...]

PR #5492, opened 2023-01-10
Implement window/showMessageRequest

ayoub-benali:
Picker does already a good job I think  

image

Todo:
- [ ] Render the message on top of the input bar
- [ ] Adapt width and hight 

Any pointers regarding how to render the message on top of the input correctly prompt would be really helpful  

Fix https://github.com/helix-editor/helix/issues/4699


the-mikedavis:
A menu would probably be more appropriate for this than a picker: https://github.com/helix-editor/helix/blob/927fa112ec049e5f40309ffdd57c314897e18bbc/helix-term/src/commands/lsp.rs#L627-L654

ayoub-benali:
I think it is too small as you can see:
image

And it might look worse once I add the message coming with the request

ayoub-benali:
Here is how it looks like when the message is added in the picker

image

I still need to fix the prompt in this example

pascalkuthe:
I agree with @ayoub-benali here.
The menu seems better suited for "inline" actions and less well suited for a large popup like this.

 In fact I originally suggested the picker on matrix. 

Nvim also handles this as a larger popup:

image


I think reusing the picker here is nice so there are fewer specical cased UI elements (and the PR is easier to review).
Altough it's true that the filtering capabilities of the picker don't really fit here.
Perhaps it might make more sense to create a new view that is a copy of the picker, rip out any filtering related logic and functionality and reuse the top line to display the message?

ayoub-benali:
I just noticed that `lsp-workspace-command` uses the same picker and as you can see the list can be quite long. The filtering capabilities do help because the user can just start typing rather than having to scroll a long list. 

image

`window/showMessageRequest` could contain similar long action list

ayoub-benali:
After discussion with @pascalkuthe I moved the LSP message to the right.  This way we can better handle messages that could span over one line.

image

pascalkuthe:
Is there anything still missing? Otherwise you can mark this PR as ready for review.

ayoub-benali:
almost, I am just working on sending the response to the server when the file picker is dismissed. I had to add a function to the constructor and update all the call sites ^^

helix-term/src/ui/picker.rs
```diff
@@ -390,13 +407,15 @@ pub struct Picker<T: Item> {
...
     callback_fn: Box<dyn Fn(&mut Context, &T, Action)>,
+    on_close_fn: Box<dyn Fn(&mut Context)>,
 }
 
 impl<T: Item> Picker<T> {
     pub fn new(
         options: Vec<T>,
         editor_data: T::Data,
         callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
+        on_close_fn: impl Fn(&mut Context) + 'static,
```
ayoub-benali, line 418: I added this function so that we can run it when the picker is dismissed but since it is not needed in the other context, I don't know if there is a better way

helix-term/src/ui/picker.rs
```diff
@@ -36,6 +36,7 @@ pub const MAX_FILE_SIZE_FOR_PREVIEW: u64 = 10 * 1024 * 1024;
 pub enum PathOrId {
     Id(DocumentId),
     Path(PathBuf),
+    Text(Rope),
```
ayoub-benali, line 39: This to show the same message in the "preview" part of the file picker

helix-term/src/ui/picker.rs
pascalkuthe: I think this mostly works fine but I think you could merge these two callbacks here and pass `None` when the picker is closed:

```rust
callback_fn: impl Fn(&mut Context, &T, Option<Action>) + 'static,
```

ayoub-benali: So `callback_fn` should be invoked here with `None` just before `close_fn` ? 
https://github.com/helix-editor/helix/blob/b6331394a3f341ad21f8fad3e6e0b93becda9ce5/helix-term/src/ui/picker.rs#L619-L621

ayoub-benali:
If your favorite language server doesn't implement this request, you can still test this feature by:
1. Installing Metals https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#metals
2. Cloning [this](https://github.com/scala/scala3-example-project) example project and opening via Helix in the root
3. Opening any `.sbt` or `.scala` file, the build message should appear. 

Once the build is done or you click `Don't show again` you will have to delete the `.metal` directory and open Helix again to force the message again.

helix-term/src/ui/picker.rs
```diff
@@ -114,7 +125,7 @@ impl<T: Item> FilePicker<T> {
     pub fn new(
         options: Vec<T>,
         editor_data: T::Data,
-        callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
+        callback_fn: impl Fn(&mut Context, Option<&T>, Action) + 'static,
```
ayoub-benali, line 128: The None case allows us to represent when the picker is dismissed, otherwise we have to define a separate `on_close` method, you can see the attempt here: https://github.com/helix-editor/helix/pull/5492/commits/34f15552a086a005b3411908f6031c59b3db2d73

helix-term/src/ui/picker.rs
```diff
@@ -618,29 +634,23 @@ impl<T: Item + 'static> Component for Picker<T> {
                 self.to_end();
             }
             key!(Esc) | ctrl!('c') => {
+                // Action shouldn't matter here because no item was selected
+                (self.callback_fn)(cx, None, Action::Replace);
                 return close_fn;
             }
             alt!(Enter) => {
-                if let Some(option) = self.selection() {
```
ayoub-benali, line 624: This isn't needed anymore since the None case is handled in the callback function

ayoub-benali:
Hi @the-mikedavis, sorry to ping you but Is it possible to give this PR a review ? 
`window/showMessageRequest` is required in Metals to import the build otherwise nothing works in scala projects. 
I made sure to often keep this PR conflict free with master but it is getting harder as more core changes have been merged

helix-term/src/application.rs
```diff
@@ -974,17 +978,66 @@ impl Application {
...
+                    Ok(MethodCall::ShowMessageRequest(params)) => {
+                        if let Some(actions) = params.actions {
+                            let call_id = id.clone();
+                            let icon = match params.typ {
+                                MessageType::ERROR => "❗",
+                                MessageType::WARNING => "⚠️",
+                                MessageType::INFO => "ℹ️",
+                                _ => "📝",
+                            };
+                            let message = format!("{}: {}", icon, &params.message);
```
the-mikedavis, line 1002: Emoji icons look out of place in the UI - we use unicode icons sparingly but not emoji AFAIK.

Usually we color spans with hint/info/warning/error from the theme (for exmaple the diagnostics pickers) but that isn't possible with the file preview here. I would prefer nvim's style without an indication of the severity, or a text marker like `WARNING: <message>`/`ERROR: <message>`

helix-term/src/commands.rs
```diff
@@ -2116,31 +2116,33 @@ fn global_search(cx: &mut Context) {
...
-                        match cx.editor.open(path, action) {
-                            Ok(_) => {}
-                            Err(e) => {
-                                cx.editor.set_error(format!(
-                                    "Failed to open file '{}': {}",
-                                    path.display(),
-                                    e
-                                ));
+                    move |cx, file_result, action| {
+                        if let Some(FileResult { path, line_num }) = file_result {
```
the-mikedavis, line 2120: For picker callbacks with large bodies, you can avoid adding to the indentation and complicating the diff by using let..else to return early:

```rust
let Some(FileResult { path, line_num }) = file_result else { return; };
```

This isn't necessary for callback functions with small (1-3 line) bodies though, the `if let` is fine for those

helix-term/src/ui/picker.rs
```diff
@@ -400,7 +424,7 @@ impl Ord for PickerMatch {
     }
 }
 
-type PickerCallback<T> = Box<dyn Fn(&mut Context, &T, Action)>;
+type PickerCallback<T> = Box<dyn Fn(&mut Context, Option<&T>, Action)>;
```
the-mikedavis, line 427: I'm split on the `Option` here for cancel/close vs a separate `cancel_fn` on the `Picker`. The `Option` seems ok and makes sense for the case when you hit Enter with no selection (empty picker) but it makes the diff really noisy - all existing pickers need to change - and the None variant is only handled with the new picker. Plus the None for cancel feels a little awkward: you couldn't distinguish between cancel and hitting Enter with an empty picker in the future if that's something we would want to do.

We could minimize the diff by using the separate `cancel_fn` approach and instead of adding that to the constructor for all `Picker`s, it could be added on to just this picker with a separate `with_close_fn` function (see `Prompt::with_line` for something similar).

What do you think @pascalkuthe? I'm split on these approaches so I defer to whatever you think is more idiomatic

pascalkuthe: The problem with having two separate closures is that you can not share state between them so for example if you create some expensive allocation and then want to do something with it on the callback you need to use some awkward `Arc`.  Now that `let .. else` is stable (great suggestion) I think just adding a short `let .. else` statement at the start of every picker is not that disruptive and would make the diff reasonable to review I think.

An alternative could be to add a trait instead of a second callback:

```
trait PickerAction{
	fn cancel(..);
	fn accept(..);
}

impl<F: Fn(..)>> Picker for F{
    fn accept(&self, ..){
		self(...)
	}
}
```

This allows splitting up the functions while still making it possible to share state between them and by implementing the trait for a closure you don't need to change existing call sites. This option would work but seems a little overkill to me. Using either a `PickerResult` enum (to distinguish accepting an empty picker) or an `Option` together with `let .. else` doesn't feel to disruptive and any existing PRs that add new pickers would only need to add a single `let .. else` line. 

> you couldn't distinguish between cancel and hitting Enter with an empty picker in the future if that's something we would want to do.

Interesting I hadn't considered that. I can't really think of a use-case. Did you have something in mind? We could change this to an appropriate `enum` then. With `let .. else` you would only need to replace `let Some(` with `let PickerResult::Accept(` then (or whatever we call the enum/variant.

helix-term/src/application.rs
ayoub-benali: I went with the text marker

helix-term/src/ui/picker.rs
the-mikedavis: Good points! I agree about the trait - it does seem a little overkill. I didn't have anything specific in mind with cancel vs. close so we can probably ignore that use-case for now. Let's keep the Option change as it is currrently 👍

helix-term/src/ui/picker.rs
```diff
@@ -204,6 +219,14 @@ impl<T: Item> FilePicker<T> {
                 let doc = editor.documents.get(&id).unwrap();
                 Preview::EditorDocument(doc)
             }
+            PathOrId::Text(rope) => {
+                let doc = Document::from(
+                    rope,
+                    None,
+                    Arc::new(ArcSwap::new(Arc::new(Config::default()))),
```
ayoub-benali, line 226: I am not sure if this is the correct value for this argument.

helix-term/src/commands.rs
ayoub-benali: I missed that comment last time. It is fixed now

helix-term/src/commands.rs
```diff
@@ -2116,7 +2116,8 @@ fn global_search(cx: &mut Context) {
                 let picker = FilePicker::new(
                     all_matches,
                     current_path,
-                    move |cx, FileResult { path, line_num }, action| {
+                    move |cx, file_result, action| {
+                        let Some(FileResult { path, line_num }) = file_result else { return; };
```
the-mikedavis, line 2120: This is a minor style nit but we should follow the recently merged (https://github.com/rust-lang/rust/pull/107312) style guide guidance for formatting these `let..else`s: https://github.com/rust-lang/rust/blob/master/src/doc/style-guide/src/statements.md#single-line-let-else-statements. Rustfmt won't automatically fix or lint for this style yet but it would be good to adhere to it now so we don't have odd formatting failures when we upgrade rustfmt: `let <pattern> = <expr> else { return };`

```suggestion
                        let Some(FileResult { path, line_num }) = file_result else { return };
```

helix-term/src/application.rs
```diff
@@ -1005,20 +1009,69 @@ impl Application {
...
+                                    tokio::spawn(
+                                        language_server.reply(call_id.clone(), Ok(response)),
+                                    );
+                                },
+                                move |_editor, _value| {
+                                    let file_location: FileLocation = (rope.clone().into(), None);
+                                    Some(file_location)
+                                },
+                            );
+                            let popup_id = "show_message_request";
```
the-mikedavis, line 1067:
```suggestion
                            let popup_id = "show-message-request";
```

these are always kebab-case (signature-help, invalid-regex, code-action, ..)

helix-term/src/commands/dap.rs
```diff
@@ -76,7 +76,10 @@ fn thread_picker(
             let picker = FilePicker::new(
                 threads,
                 thread_states,
-                move |cx, thread, _action| callback_fn(cx.editor, thread),
+                move |cx, thread, _action| {
+                    let Some(t) = thread else {return;};
```
the-mikedavis, line 80:
```suggestion
                    let Some(t) = thread else { return };
```

helix-term/src/commands/dap.rs
```diff
@@ -273,7 +276,8 @@ pub fn dap_launch(cx: &mut Context) {
     cx.push_layer(Box::new(overlayed(Picker::new(
         templates,
         (),
-        |cx, template, _action| {
+        |cx, debug_template, _action| {
+            let Some(template) = debug_template else {return;};
```
the-mikedavis, line 280:
```suggestion
            let Some(template) = debug_template else { return };
```

helix-term/src/commands/dap.rs
```diff
@@ -731,7 +735,8 @@ pub fn dap_switch_stack_frame(cx: &mut Context) {
     let picker = FilePicker::new(
         frames,
         (),
-        move |cx, frame, _action| {
+        move |cx, stack_frame, _action| {
+            let Some(frame) = stack_frame else {return;};
```
the-mikedavis, line 739:
```suggestion
            let Some(frame) = stack_frame else { return };
```

helix-term/src/ui/mod.rs
```diff
@@ -220,7 +220,8 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
     FilePicker::new(
         files,
         root,
-        move |cx, path: &PathBuf, action| {
+        move |cx, path_buff, action| {
+            let Some(path) = path_buff  else { return };
```
the-mikedavis, line 224:
```suggestion
            let Some(path) = path_buff else { return };
```

helix-term/src/ui/picker.rs
the-mikedavis: This should be `editor.config.clone()` instead

the-mikedavis (commented):
I left some minor comments but otherwise I think this looks good.

In the long run I think I would prefer a custom component for this kind of UI. Using the picker's preview to show a message feels inconsistent to me: the picker preview is otherwise always used for showing supporting information about the currently selected option instead of context about why the picker is open. When I open a picker my eyes are always on the options first and then the preview if needed. This is a relatively small change though to add a feature so I think we can leave that off for future implementation/discussion

the-mikedavis (approved):

NomisIV:
Any chance we could get this merged? I would personally find this very useful!

ayoub-benali:
Hi @NomisIV, I rebased this PR multiple time but it doesn't help as long as @archseer doesn't have time to review.
This same applies to https://github.com/helix-editor/helix/pull/5535

igor-ramazanov:
I would really love to see this merged.
Thank you very much for your contribution @ayoub-benali!

I am also using Metals for working with Scala.

For now I am building and patching Helix locally like I did with https://github.com/helix-editor/helix/pull/5535, but I am struggling with rebasing and resolving conflicts in this PR with `master` branch or latest `23.05` tag to do the same.

@ayoub-benali Could you or someone else, please, resolve the conflicts, at least with the latest `23.05` branch?

igor-ramazanov:
Can we include this into the next release?

filipwiech:
Hello @the-mikedavis, @pascalkuthe  and @archseer! :slightly_smiling_face: 
Sorry to ping you like this, but this PR has been sitting reviewed and approved by 2 maintainers for months now. It would be really great to get this merged, since it's part of the LSP specification and some servers (like for example Scala's Metals mentioned in the original issue) use it to provide core features without which usability and user experience are impacted quite heavily. Any chance to see this progress? Is there anything still needed to be done here? :slightly_smiling_face: 
Thank you for your work on this amazing editor (and big thanks to the author of this PR for the contribution)! :slightly_smiling_face:

hakan-demirli:
I have tried to rebase this to current master and miserably failed. Currently metals (a scala LSP) is unusable without this PR. I am using neovim as a substitute which is pita.

the-mikedavis:
Added in 76c23b306716ebd9ec6e73384915e58fad061557

PR #5509, opened 2023-01-11
Update tree-sitter-json that supports comments, add jsonc as file-type for json format (#5452)

### LINKED ISSUE #5452: JSONC support
__EDIT: collapsible sections for better readability__

<details><summary>Problem</summary>

## Problem

Json highlighting works out of the box, but then I opened my Jsonc config file and the text was greyed out. I did some research and realized that this is the correct default behavior because there is no `file-types = [..., "jsonc"]` for json language: https://github.com/helix-editor/helix/blob/b7ca0581d81f0ac80418a3e052df2bd5d9f71ad4/languages.toml#L141

Then I added this line:
``` toml
[[language]]
name = "json"
file-types = ["jsonc", "json"]
```

After this _jsonc_ was also colored but comments was bold red (means error):
image

I looked in _vscode json server_ and found that I can pass `languageId = "jsonc"` to disable comment warnings:  
[Link to lines](https://github.com/microsoft/vscode/blob/4f69cdf95a12cef48d405b38bf7812a7f297c310/extensions/json-language-features/server/src/jsonServer.ts#L324-L326)
``` ts
const documentSettings: DocumentLanguageSettings = textDocument.languageId === 'jsonc' ? { comments: 'ignore', trailingCommas: 'warning' } : { comments: 'error', trailingCommas: 'error' };
```
</details>

<details><summary>What I tried</summary>

### Attempt 1
Pass `language-id = "jsonc"` as an argument to the _json_ language server so that it works for both formats:
``` toml
[[language]]
name = "json"
file-types = ["jsonc", "json"]
language-server = { command = "vscode-json-language-server", args = ["--stdio"], language-id = "jsonc" }
```
__Result__:  nothing changes. The colors are the same as in the picture.

### Attempt 2
Create new language _jsonc_ and also pass _lang-id_ argument to language-server:
``` toml
[[language]]
name = "jsonc"
scope = "source.jsonc"
file-types = ["jsonc"]
roots = []
language-server = { command = "vscode-json-language-server", args = ["--stdio"], language-id = "jsonc" }
config = { "provideFormatter" = true, "json" = { "keepLines" = { "enable" = true } } }
indent = { tab-width = 2, unit = "  " }
```
__Result__:  gray text without any color. Adding next doesn't help:
``` toml
[[grammar]]
name = "jsonc"
source = { git = "https://github.com/tree-sitter/tree-sitter-json", rev = "65bceef69c3b0f24c0b19ce67d79f57c96e90fcb" }
```
</details>

## Question
How to fix one of my attempts to make it work or tell another approach to disable comments (and trailingCommas) warnings in jsonc?

__P.S.__ This is my first experience with all this stuff that vscode and the IDEs do themselves, so I might be missing something.

woojiq:
Helix uses old tree-sitter-json git revision. Not too long ago new [version](https://github.com/tree-sitter/tree-sitter-json/commit/124edff33adfa209247c03119285feec8c4aca43) came out that supports comments. Also I added _jsonc_ format as alias for _json_. In todays world they're equal, all uses _json_ with comments and I very rarely see _jsonc_ format. There is no needs to distinct this formats.

Both formats look like this:
image

Previously json:
image

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -138,7 +138,7 @@ indent = { tab-width = 2, unit = "  " }
 name = "json"
 scope = "source.json"
 injection-regex = "json"
-file-types = ["json"]
+file-types = ["json", "jsonc"]
 roots = []
 language-server = { command = "vscode-json-language-server", args = ["--stdio"] }
 auto-format = true
@@ -147,7 +147,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "json"
-source = { git = "https://github.com/tree-sitter/tree-sitter-json", rev = "65bceef69c3b0f24c0b19ce67d79f57c96e90fcb" }
+source = { git = "https://github.com/tree-sitter/tree-sitter-json", rev = "73076754005a460947cafe8e03a8cf5fa4fa2938" }
 
 [[language]]
 name = "c"


woojiq:
Closes #5452

David-Else:
It is very controversial and wrong to use comments in a JSON file, hence the linter is rightly going mental when it happens.

Check https://stackoverflow.com/questions/244777/can-comments-be-used-in-json , `6856` votes saying you can not use them, and lengthy explanations why.

I am very keen to see the JSONC format being used properly, Helix should definately accept JSONC files with a `.jsonc` extension.

It looks like Max Brunsfeld has decided to ignore standards... mmm, as we should be using the latest tree-sitter-json version we could be stuck with this aberration unless someone takes action.

archseer:
Tree sitter grammars are approximations so it's OK for them to diverge slightly and be more permissive. You should always use your language server or parser for validation. So I think the change here is reasonable since some parsers do actually support comments

PR #5522, opened 2023-01-13
Language support for dhall

divarvel:
This adds language configuration for the [dhall language](https://dhall-lang.org/).

# LSP & formatting

I chose to add configuration for the command line formatter even though the LSP implementation handles formatting, since the formatter is part of the `dhall` CLI and is more widely available.

# Textobject queries

- The class queries don't apply to dhall.
- Dhall does not have unit tests per se, but does have assert expressions.
- dhall does not have named functions, but has lambdas (term and type-level)

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2092,3 +2092,18 @@ comment-token = "//"
 [[grammar]]
 name = "ponylang"
 source = { git = "https://github.com/mfelsche/tree-sitter-ponylang", rev = "ef66b151bc2604f431b5668fcec4747db4290e11" }
+
+[[language]]
+name = "dhall"
+scope = "source.dhall"
+injection-regex = "dhall"
+file-types = ["dhall"]
+roots = []
+comment-token = "--"
+indent = { tab-width = 2, unit = "  " }
+language-server = { command = "dhall-lsp-server" }
+formatter = { command = "dhall" , args = ["format"] }
+
+[[grammar]]
+name = "dhall"
+source = { git = "https://github.com/jbellerb/tree-sitter-dhall", rev = "affb6ee38d629c9296749767ab832d69bb0d9ea8" }
--- a/runtime/queries/dhall/highlights.scm
+++ b/runtime/queries/dhall/highlights.scm
@@ -1,0 +1,52 @@
+;; Literals
+
+(integer_literal) @constant.numeric.integer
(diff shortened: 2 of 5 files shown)


languages.toml
```diff
@@ -2092,3 +2092,18 @@ comment-token = "//"
 [[grammar]]
 name = "ponylang"
 source = { git = "https://github.com/mfelsche/tree-sitter-ponylang", rev = "ef66b151bc2604f431b5668fcec4747db4290e11" }
+
+[[language]]
+name = "dhall"
+scope = "source.dhall"
+injection-regex = "^dhall$"
```
the-mikedavis, line 2099:
```suggestion
injection-regex = "dhall"
```

runtime/queries/dhall/textobjects.scm
```diff
@@ -0,0 +1,36 @@
...
+;; class.inside
+;; class.around
+;; test.inside
+;; test.around
+;; parameter.inside
+;; comment.inside
+;; comment.around
+
+;; Queries for navigating using textobjects
+
```
the-mikedavis, line 13: Were these comments a reference while writing these? I think they can be cleaned up now

divarvel: sure, will do.

divarvel: (done)

languages.toml
divarvel: I should update the docs, where the example for the injection regex uses anchors (https://github.com/helix-editor/helix/blob/master/book/src/languages.md?plain=1#L38). Is it ok to do it in this PR, or in another one?

runtime/queries/dhall/highlights.scm
```diff
@@ -0,0 +1,52 @@
+;; Literals
+
+(integer_literal) @constant.numeric.integer
+(natural_literal) @constant.numeric.integer
+(double_literal) @constant.numeric.float
+(boolean_literal) @constant.numeric.bool
```
the-mikedavis, line 6:
```suggestion
(boolean_literal) @constant.builtin.boolean
```

languages.toml
divarvel: done, but i don't know how to update `book/generated`

the-mikedavis (approved):

PR #5523, opened 2023-01-13
Only render the auto-complete menu if it intersects with signature help

### LINKED ISSUE #3821: Function arguments tip covers autocompletion if tip doesn't fit in the upper screen space
### Summary

Here it's covered and I can't see the autocompletion (but it's there)
Bildschirmfoto 2022-09-12 um 19 39 21

But if I make enough space for it to fit then it works as expected
Bildschirmfoto 2022-09-12 um 19 40 05

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

macOS

### Terminal Emulator

The latest iTerm 2

### Helix Version

22.08.1

kirawi:
Supersedes #5495 because it required moving `Completion` into its own layer in the compositor. This one simply checks if the two intersect, and will always defer to the auto-complete menu if they do.

Closes #3821

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4233,7 +4233,8 @@ pub fn completion(cx: &mut Context) {
                     trigger_offset,
                     size,
                 )
-                .and_then(|area| Some(area.intersects(signature_help_area?)))
+                .zip(signature_help_area)
+                .filter(|(a, b)| a.intersects(*b))
                 .is_some()
             {
                 compositor.remove(SignatureHelp::ID);
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4225,7 +4225,6 @@ pub fn completion(cx: &mut Context) {
 
             // Delete the signature help popup if they intersect.
             if ui.set_completion(
-                // compositor,
                 signature_help_area,
                 editor,
                 items,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4220,8 +4220,7 @@ pub fn completion(cx: &mut Context) {
             let size = compositor.size();
             let signature_help_area = compositor
                 .find_id::<Popup<SignatureHelp>>(SignatureHelp::ID)
-                .and_then(|signature_help| signature_help.area(size, editor))
-                .unwrap();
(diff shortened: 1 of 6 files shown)


kirawi:
One slight issue is that if you delete the retrigger character fast enough, the auto-complete popup will show again for that field. If you delete it after a while, the signature help will pop instead.
Capture

kirawi:
Screenshot 2023-01-26 at 7 31 59 PM

Ran into this, maybe it's not a good idea to re-request the signature help popup after closing the auto-complete menu.

archseer:
Replaced by https://github.com/helix-editor/helix/pull/5842 ?

sudormrfbin:
#5842 handles overlap of the completion popup and the documentation popup of the currently selected completion (the popup that gets rendered on the right of the completion popup) whereas this PR handles overlap of signature help and completion popup so it's different in scope. Both aim to improvement the popup experience :)

helix-term/src/commands.rs
```diff
@@ -4068,15 +4071,26 @@ pub fn completion(cx: &mut Context) {
...
+            if ui
+                .set_completion(
+                    editor,
+                    items,
+                    offset_encoding,
+                    start_offset,
+                    trigger_offset,
+                    size,
+                )
+                .and_then(|area| Some(area.intersects(signature_help_area?)))
```
kirawi, line 4089: self-nit: logic bug

helix-term/src/commands/lsp.rs
```diff
@@ -1131,10 +1131,25 @@ pub fn signature_help_impl(cx: &mut Context, invoked: SignatureHelpInvoked) {
...
+            let mut popup = Popup::new(SignatureHelp::ID, contents)
                 .position(old_popup.and_then(|p| p.get_position()))
                 .position_bias(Open::Above)
                 .ignore_escape_key(true);
+
+            // Don't create a popup if it intersects the auto-complete menu.
+            let size = compositor.size();
+            if compositor
+                .find::<ui::EditorView>()
+                .unwrap()
```
Proful, line 1143: I assume compositor will always find ui::EditorView, so unwrap is safe here?

kirawi: Yeah, it's always initialized on startup and never dropped.

Issue #5448, opened 2023-01-08
Helix crashes when lacking space to show popups


pathwave:
### Summary

There is two different panics happening, not quite sure what triggers the first one but I think its related to how much content
is being shown in the popup window. I sometimes got the second panic message when doing the first reproduction steps.

### Reproduction Steps

First crash trigger:

1. Start helix and run command "Show docs under cursor"
2. Resize terminal window height smaller and smaller
3. Helix crashes with:
thread 'main' panicked at 'index out of bounds: the len is 3276 but the index is 3482', helix-tui\src\buffer.rs:513:33
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\panicking.rs:584
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\panicking.rs:143
   2: core::panicking::panic_bounds_check
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\panicking.rs:84
   3: helix_tui::buffer::Buffer::clear_with
   4: <helix_term::ui::popup::Popup<T> as helix_term::compositor::Component>::render
   5: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
   6: hx::main_impl::{{closure}}
   7: tokio::runtime::park::CachedParkThread::block_on
   8: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
   9: hx::setup_logging

Second crash trigger:

1. Start helix and make the terminal window height 4 lines or smaller.
2. Run command "Show docs under cursor"
3. Helix crashes with: 
thread 'main' panicked at 'Component needs required_size implemented in order to be embedded in a popup', helix-term\src\ui\popup.rs:209:14
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\std\src\panicking.rs:584
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\panicking.rs:143
   2: core::panicking::panic_display<str>
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e\/library\core\src\panicking.rs:72
   3: core::panicking::panic_str
[...]

PR #5529, opened 2023-01-14
Fix lack of space for popup crash

nuid32:
There are several places None can be returned from one of required_size implementations. I fixed markdown implementation by merely deleting code, that returned None. Did some other minor fixes to prevent crash.
#5448


catdevnull:
this just crashed my editor losing a bunch of changes i was making, pls merge!

helix-tui/src/buffer.rs
```diff
@@ -510,9 +510,13 @@ impl Buffer {
...
-                let cell = &mut self[(x, y)];
-                cell.reset();
-                cell.set_style(style);
+                if let Some(cell) = &mut self.get_mut(x, y) {
+                    cell.reset();
+                    cell.set_style(style);
+                } else {
+                    // Skip this render if the window size has changed
+                    return;
+                }
```
Plasma-Vortex, line 519: If I'm understanding it right, this index-out-of-bounds panic only happens because of the [underflow bug in `Rect::intersection`](https://github.com/helix-editor/helix/blob/a9973c6947382010b434564481dca9916294c84a/helix-view/src/graphics.rs#L240). Now that it's fixed, and `Rect::intersection` returns a zero-area rectangle instead of a huge rectangle outside of the viewport, this if/else check and early return will never even happen, since the loop has zero elements. Should we remove it?

nuid32: That's great. I'll close my PR

Plasma-Vortex: Oops, I meant your PR is fixing the bug in `Rect::intersection`, and we should be merging instead of closing it :sweat_smile: Just pointing out that you don't have to change anything in `buffer.rs`.

Plasma-Vortex: Not sure if `required_size` returning `None` is the cause of a separate crash, or if it's also because of `Rect::intersection` — I'd have to dig into the codebase more.

But seeing as the return type of `required_size` is an `Option<(u16, u16)>`, it's strange that returning `None` is an issue. Maybe the real bug is the calling function not handling `None` properly.

Plasma-Vortex:
@nuid32 Seems like there was a misunderstanding, I wasn't suggesting to close this. See my comment above (which is now hidden)

nuid32:
> @nuid32 Seems like there was a misunderstanding, I wasn't suggesting to close this. See my comment above (which is now hidden)

My fault, haven't enough time to sort out.

nuid32 (commented):
LGTM

cd-a:
Was this supposed to be closed?

nuid32:
> Was this supposed to be closed?

I'm sorry, no, it's not. I just deleted my fork of helix and I totally forgot about this PR. I will fix it right now

Issue #4393, opened 2022-10-20
Rename buffers


cor:
It would be very nice if you could rename buffers from Helix.

For example, I'm editting `a.txt` within a project, and i need to rename it to `b.txt`. 
Currently, you would:

- `<space>f` -> `a.txt` 
- `:w b.txt`
- quit helix
- `rm a.txt`
- open helix
- `<space>f` -> `b.txt`

It would be much nicer if we had a rename command from within helix, so that you could simply:
- `<space>f` -> `a.txt` 
- `:buffer-rename b.txt`

*the buffer is now renamed to b.txt, while it is still open in your editor*

cor:
I don't have sufficient permissions in order to add labels, but this seems like a good first issue to me :)

geometryolife:
There is also a slightly more complicated way what run shell commands in Helix:
```
:run-shell-command mv old_file_name new_file_name
```
I also agree with adding a build-in command or shortcut for Helix to change the file name.

Elias-Graf:
Does this also rename references to the given file in code? I think https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_willRenameFiles is relevant.

That is also why @geometryolife suggestion does technically achieve renaming the file, but wouldn't accomplish what I would be expecting it to do.

austinletson:
@cor @Elias-Graf there seems to be some discussion in the above linked issues as to what operations involving file management should be included in the editor as opposed to externally. What if instead of creating a new file/buffer and deleting the old file, there was a command that started a new buffer and saved the old buffers content to the new buffer but didn't delete the source file for the old buffer.

This is similar to the way `:saveas` works in vim: https://vimdoc.sourceforge.net/htmldoc/editing.html#:saveas

It seems like `:write <file-name>` already achieves this though. If I open a file and write to a different file, the buffer is renamed.

cor:
@austinletson in my opinion this really should be included in the editor. People primarily use Helix for creating code, and this process almost always involves renaming, deleting, and moving files. `:saveas` and `:write` are inferior solutions since they require the user to still manually delete the old file.

Also, like @Elias-Graf mentioned, this does not rename references to the file, which it should do.

ontley:
Well I implemented the command, now to wait for it to hopefully get merged

PR #4514, opened 2022-10-29
Add rename command.

grv07:
#4393

Add rename command  `:rename new_name.txt`

PR #5531, opened 2023-01-14
Added command to rename current document, with lsp support

ontley:
#4393 

Includes willRenameFiles and didRenameFiles requests to the language server.
Honestly not sure if didRenameFiles is even needed here

Should file changes suggested by the language server be automatically saved if the file didn't have any other changes?


the-mikedavis:
There is already https://github.com/helix-editor/helix/pull/4514 but it doesn't cover the LSP renaming

cor:
Nice one! Really happy that it includes LSP rename support 👍🏻

ontley:
I added the rnm alias for the command, referring to [#4514 (comment)](https://github.com/helix-editor/helix/pull/4514#issuecomment-1396122027)

ontley:
I wonder if it would be good to add a feature to automatically save suggested changes from the lsp if a file didn't have changes before, saves the trouble of opening every file and saving manually, or perhaps a status message saying which files changed?

devgioele:
Looking forward to this!

edrex:
I need to test this, but do suggested changes cause non-open files to be opened?

> I wonder if it would be good to add a feature to automatically save suggested changes from the lsp if a file didn't have changes before, saves the trouble of opening every file and saving manually

IMO it's not needed, as `:wa` is easy enough, and there's already `editor.auto-save`. I get wanting transactionality, since the rename is immediate, but IDT it warrants a special case.

ontley:
> I need to test this, but do suggested changes cause non-open files to be opened?
> 
> > I wonder if it would be good to add a feature to automatically save suggested changes from the lsp if a file didn't have changes before, saves the trouble of opening every file and saving manually
> 
> IMO it's not needed, as `:wa` is easy enough, and there's already `editor.auto-save`. I get wanting transactionality, since the rename is immediate, but IDT it warrants a special case.

ontley:
> I need to test this, but do suggested changes cause non-open files to be opened?
> 
> > I wonder if it would be good to add a feature to automatically save suggested changes from the lsp if a file didn't have changes before, saves the trouble of opening every file and saving manually
> 
> IMO it's not needed, as `:wa` is easy enough, and there's already `editor.auto-save`. I get wanting transactionality, since the rename is immediate, but IDT it warrants a special case.

Ah I didn't even know about `:wa`, I guess it's not needed then

edrex:
> Ah I didn't even know about :wa
It's a vimism. There's also `:wqa`.

ontley:
Should the lsp errors be displayed in the status bar, along with the message that the file was successfully renamed?

ontley:
> Why it is not merged already?

Waiting for reviews

vikaskaliramna07:
> > Why it is not merged already?
> 
> Waiting for reviews

Can anyone review or contributors only?

ontley:
> > > Why it is not merged already?
> > 
> > 
> > Waiting for reviews
> 
> Can anyone review or contributors only?

Only "reviewers with write access" and I can't even request reviews

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    ensure!(args.len() == 1, ":rename takes one argument");
+    let new_name = args.first().unwrap();
+
+    let (_, doc) = current!(cx.editor);
+    let path = doc
+        .relative_path()
```
pascalkuthe, line 2185:
```suggestion
        .path()
```

using a relative path makes no sense here

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+    if path_new.exists() {
+        bail!("This file already exists");
+    }
+
+    if let Err(e) = std::fs::rename(&path, &path_new) {
+        bail!("Could not rename file, error: {}", e);
+    }
+    let (_, doc) = current!(cx.editor);
+    doc.set_path(Some(path_new.as_path()))
+        .map_err(|_| anyhow!("File renamed, but could not set path of the document"))?;
```
pascalkuthe, line 2200: i think that this should probably happen before the rename. If the fails later we can roll this back pretty easily (ideally with an unchecked method) while rolling back IO is pretty hard

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+            let old_uri = old_uri_str.to_string();
+            if let Ok(new_uri_str) = Url::from_file_path(&path_new) {
+                let new_uri = new_uri_str.to_string();
+                let files = vec![lsp::FileRename { old_uri, new_uri }];
+                match helix_lsp::block_on(lsp_client.will_rename_files(&files)) {
+                    Ok(edit) => {
+                        if apply_workspace_edit(cx.editor, helix_lsp::OffsetEncoding::Utf8, &edit)
+                            .is_err()
+                        {
+                            log::error!(":rename command failed to apply edits")
```
pascalkuthe, line 2213: we already show an error message herer you don't need to display it. The returned result is just forcases where we need to communicate to the LSP that the edit failed (not here)

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+                        }
+                    }
+                    Err(err) => log::error!("Language server error: {}", err),
```
pascalkuthe, line 2216:
```suggestion
                    Err(err) => log::error!("willRename request failed: {err}"),
```

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+    }
+
+    if let Err(e) = std::fs::rename(&path, &path_new) {
+        bail!("Could not rename file, error: {}", e);
+    }
+    let (_, doc) = current!(cx.editor);
+    doc.set_path(Some(path_new.as_path()))
+        .map_err(|_| anyhow!("File renamed, but could not set path of the document"))?;
+
+    if let Some(lsp_client) = doc.language_server() {
```
pascalkuthe, line 2202: the wii rename request must be sent (and processed) *before* the rename happens so this needs to be moved to the start of the function:

> The will rename files request is sent from the client to the server before files are actually renamed as long as the rename is triggered from within the client either by a user action or by applying a workspace edit. The request can return a WorkspaceEdit which will be applied to workspace before the files are renamed. 

At the end of the function you must send didRename:

> The did rename files notification is sent from the client to the server when files were renamed from within the client.

you also need to support mullti LSP here. Probably ok to just send wiiRename to the first server that supports it and didRename to all of them

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,64 @@ fn reset_diff_change(
...
+    let new_name = args.first().unwrap();
+
+    let (_, doc) = current!(cx.editor);
+    let path = doc
+        .relative_path()
+        .ok_or_else(|| anyhow!("Buffer not saved, use :write"))?
+        .canonicalize()
+        .map_err(|_| anyhow!("Could not get absolute path of the document"))?;
+    let mut path_new = path.clone();
+    path_new.set_file_name(OsStr::new(new_name.as_ref()));
```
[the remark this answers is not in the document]
pascalkuthe: note that this means that the destination directory might not exist so we need `rename!` too

helix-lsp/src/client.rs
```diff
@@ -592,6 +597,25 @@ impl Client {
...
+
+        async move {
+            let json = request.await?;
+            let response: Option<lsp::WorkspaceEdit> = serde_json::from_value(json)?;
+            Ok(response.unwrap_or_default())
+        }
+    }
+
+    pub fn did_rename_files(&self, params: &[lsp::FileRename]) -> impl Future<Output = Result<()>> {
+        let files = params.to_owned();
```
pascalkuthe, line 615: this must check the server capabiliy `workspace.fileOperations.didRename`

helix-lsp/src/client.rs
```diff
@@ -592,6 +597,25 @@ impl Client {
         )
     }
 
+    pub fn will_rename_files(
+        &self,
+        params: &Vec<lsp::FileRename>,
+    ) -> impl Future<Output = Result<lsp::WorkspaceEdit>> {
+        let files = params.to_owned();
```
pascalkuthe, line 604: you need to check `workspace.fileOperations.willRename` herer (or even better at the callsite to support multiple LSPs)

pascalkuthe (changes requested):

pascalkuthe:
> > > Why it is not merged already?
> > 
> > 
> > Waiting for reviews
> 
> Can anyone review or contributors only?

you are probably new to opensource. A PR needs to be reviewed by a miantainer before its merged. Only maintainers can review PRs. We can't just trust random people with code review. There are hundreds of issues and PRs for us to get through so it can take a while.

vikaskaliramna07:
> > > > Why it is not merged already?
> > > 
> > > 
> > > Waiting for reviews
> > 
> > 
> > Can anyone review or contributors only?
> 
> you are probably new to opensource. A PR needs to be reviewed by a miantainer before its merged. Only maintainers can review PRs. We can't just trust random people with code review. There are hundreds of issues and PRs for us to get through so it can take a while.

yeah, I am new. Thanks for providing information about reviews!

helix-term/src/commands/typed.rs
```diff
@@ -2752,6 +2812,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             fun: reset_diff_change,
             signature: CommandSignature::none(),
         },
+        TypableCommand {
+            name: "rename",
+            aliases: &["rnm"],
```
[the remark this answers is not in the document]
ontley: If we do set the alias to move, we're probably better of with calling the command itself `move` or `move_file`

helix-term/src/commands/typed.rs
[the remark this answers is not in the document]
ontley: This is a sort of weird one, if we change the command name to `move` as previously said, probably a good ide, thoguh `mv` in bash takes absolute paths, not ones relative to the file but I think it's work

On the other hand I don't think a `file` renaming function should be able to move it as most file managers just let you rename it

Imo, `move` is probably better

helix-term/src/commands/typed.rs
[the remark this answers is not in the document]
pascalkuthe: we have other commands that have totally different aliases too so I think leaving it like that is fine too

helix-term/src/commands/typed.rs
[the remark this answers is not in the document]
pascalkuthe: I guess it might make mose sense to just call the entire command move and make the destination a path relative to CWD. If somebody wants to just change the filename they can use `%filname/../<new_name>` in the future.

I think that generally move is a better idea than rename because it can do more (and the LS can handle full moves  for example RA can move a module to a totally different module

helix-term/src/commands/typed.rs
```diff
@@ -2167,6 +2169,62 @@ fn reset_diff_change(
...
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    ensure!(args.len() == 1, ":rename takes one argument");
+    let new_name = args.first().unwrap();
+
+    let (_, doc) = current!(cx.editor);
+    let path = doc
+        .path()
+        .ok_or_else(|| anyhow!("Scratch buffers can not be renamed, use :write"))?
```
wetheredge, line 2186:
```suggestion
        .ok_or_else(|| anyhow!("Scratch buffers can not be renamed, use :write"))?;
```

Issue #4674, opened 2022-11-09
DAP should react to Terminated event


zobo:
When a DA (debug adapter) sends the [Terminated Event](https://microsoft.github.io/debug-adapter-protocol/specification#Events_Terminated) the client should send the [Disconnect Request](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect).

zobo:
I maintain the PHP DAP. Here is my very minimal languages.toml

```toml
[[language]]
name = "php"

[language.debugger]
name = "php-debug"
transport = "stdio"
command = "...path.to..\\node.exe"
args = ["...path.to..\\vscode-php-debug\\out\\phpDebug.js"]

[[language.debugger.templates]]
name = "Listen"
request = "launch"
completion = [ ]
args = { log = "true" }

[[language.debugger.templates]]
name = "Launch script"
request = "launch"
#completion = [ { name = "script", completion = "filename" } ]
completion = [ "filename" ]
args = { log = "true", program = "{0}", runtimeExecutable = "...path.to..\\php.exe" }
```

PR #5532, opened 2023-01-14
feat(dap): send Disconnect if Terminated event received

filipdutescu:
Send a `Disconnect` DAP request if the `Terminated` event is received. According to the specification, if the debugging session was started by as `launch`, the debuggee should be terminated alongside the session. If instead the session was started as `attach`, it should not be disposed of.

This default behaviour can be overriden if the `supportTerminateDebuggee` capability is supported by the adapter, through the `Disconnect` request `terminateDebuggee` argument, as described in
[the specification][discon-spec].

This also implies saving the starting command for a debug sessions, in order to decide which behaviour should be used, as well as validating the capabilities of the adapter, in order to decide what the disconnect should do.

An additional change made is handling of the `Exited` event, showing a message if the exit code is different than `0`, for the user to be aware off the termination failure.

[discon-spec]: https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Disconnect

Closes: #4674

### DIFF
--- a/helix-dap/src/client.rs
+++ b/helix-dap/src/client.rs
@@ -1,4 +1,5 @@
 use crate::{
+    requests::DisconnectArguments,
     transport::{Payload, Request, Response, Transport},
     types::*,
     Error, Result, ThreadId,
@@ -31,6 +32,7 @@ pub struct Client {
     _process: Option<Child>,
     server_tx: UnboundedSender<Payload>,
     request_counter: AtomicU64,
+    connection_type: Option<ConnectionType>,
     pub caps: Option<DebuggerCapabilities>,
     // thread_id -> frames
     pub stack_frames: HashMap<ThreadId, Vec<StackFrame>>,
@@ -39,6 +41,12 @@ pub struct Client {
     /// Currently active frame for the current thread.
     pub active_frame: Option<usize>,
     pub quirks: DebuggerQuirks,
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum ConnectionType {
+    Launch,
+    Attach,
 }
 
 impl Client {
@@ -78,7 +86,7 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             caps: None,
-            //
+            connection_type: None,
             stack_frames: HashMap::new(),
             thread_states: HashMap::new(),
             thread_id: None,
@@ -205,6 +213,10 @@ impl Client {
 
     pub fn id(&self) -> usize {
         self.id
+    }
+
+    pub fn connection_type(&self) -> Option<ConnectionType> {
+        self.connection_type
     }
 
     fn next_request_id(&self) -> u64 {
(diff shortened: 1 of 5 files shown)


helix-view/src/handlers/dap.rs
```diff
@@ -274,6 +275,62 @@ impl Editor {
...
+                                        debugger.attach(restart_args).await
+                                    };
+
+                                if let Err(err) = relaunch_resp {
+                                    self.set_error(format!(
+                                        "Failed to restart debugging session: {:?}",
+                                        err
+                                    ));
+                                }
+                            }
```
filipdutescu, line 313: restart should be done with the [`Restart Request`](https://microsoft.github.io/debug-adapter-protocol/specification#Requests_Restart) :thinking:

filipdutescu (commented):
just realised the restart implementation I made is not correct according to the spec. will redo it

helix-view/src/handlers/dap.rs
filipdutescu: I was incorrect in my assumption. the request should be used to _*trigger*_ a restart, rather than "create one". In any case, this should be further elaborated, most likely in a different PR, handling restarting as a whole

filipdutescu:
I have re-read the spec and I believe this part of the restart sequence to be correct. What needs to be done in addition is to expose a restart request, if supported, to the user, same as with launch, terminate etc.

Cc: @pascalkuthe could we move it to waiting on review? Many thanks!

helix-dap/src/client.rs
```diff
@@ -41,6 +43,13 @@ pub struct Client {
     pub quirks: DebuggerQuirks,
 }
 
+#[derive(Clone, Copy, Debug)]
+pub enum StartingRequest {
+    Launch,
+    Attach,
```
archseer, line 49: Remove `None` and instead use `Option<StartingRequest>`

archseer: Also this seems better named as connection type?

archseer (changes requested):

helix-dap/src/client.rs
```diff
@@ -78,7 +86,7 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             caps: None,
-            //
+            starting_request: None,
```
archseer, line 89: This should also be renamed to connection_type now

helix-dap/src/client.rs
```diff
@@ -207,6 +215,10 @@ impl Client {
         self.id
     }
 
+    pub fn starting_request(&self) -> Option<ConnectionType> {
```
archseer, line 218: Same here

helix-view/src/handlers/dap.rs
```diff
@@ -274,6 +275,65 @@ impl Editor {
...
+                        terminate_debuggee: None,
+                        suspend_debuggee: None,
+                    });
+
+                    let starting_request = debugger.starting_request();
+                    let result = debugger.disconnect(disconnect_args).await;
+                    if result.is_ok() {
+                        if restart {
+                            match starting_request {
+                                None => {
```
archseer, line 296: None should come after Some

helix-view/src/handlers/dap.rs
```diff
@@ -274,6 +275,65 @@ impl Editor {
...
+                    let restart = restart_args.is_some();
+                    let disconnect_args = Some(DisconnectArguments {
+                        restart: Some(restart),
+                        terminate_debuggee: None,
+                        suspend_debuggee: None,
+                    });
+
+                    let starting_request = debugger.starting_request();
+                    let result = debugger.disconnect(disconnect_args).await;
+                    if result.is_ok() {
```
archseer, line 293: Rewrite this so it's 

```
if let Err(ere) = result {
   set_error(...)
   return;
}
```

That way the error exits early and the regular path isn't indented inside an if

helix-view/src/handlers/dap.rs
```diff
@@ -274,6 +275,65 @@ impl Editor {
...
+                        if restart {
```
archseer, line 294: Similar to the other PR you want to match on the argument so we avoid an unwrap

archseer (changes requested):

filipdutescu:
@archseer implemented your comments, thanks very much for the patience and advice, I have learnt a lot and hopefully improved the way I write Rust. I also rebased the PR and squashed the commits, to make it a bit cleaner, since there are not that many modifications (hopefully) as to make it a chore to revisit the changes.

archseer (approved):
Sorry it took me a while to get back to this! LGTM!

filipdutescu:
No worries, thanks for reviewing and merging it!

Issue #5524, opened 2023-01-13
Implement window/showMessage in LSP client


ayoub-benali:
Currently Helix is simply logging a warning here: 
https://github.com/helix-editor/helix/blob/b6331394a3f341ad21f8fad3e6e0b93becda9ce5/helix-term/src/application.rs#L821

while Scala's language server [Metals](https://scalameta.org/metals/) make use of it. You can see it in the [spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showMessage).

I saw that there was some work by @matoous in https://github.com/helix-editor/helix/pull/1445 but it is dropped. So I want first to get some ideas on how handle this message. 
Is bellow the status line for example a good place like the progress messages? should it be configurable and off by default like `display-messages`  ?

sigmaSd:
https://github.com/v-analyzer/v-analyzer also uses it

PR #5535, opened 2023-01-14
Handle window/showMessage and display it bellow status line

ayoub-benali:
This feature is also hidden behind the `display-messages` setting which is used for displaying [Progress](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#progress).

I don't think it make sense to create separate setting for this feature as I don't expect a server to send both `Progress` and `window/showMessage` but I can add one if you prefer.

[Metals](scalameta.org/metals/) relies on this notification for providing info about the build and is configured [here]( https://scalameta.org/metals/docs/integrations/new-editor#statusbarprovider). 

Hopefully this will be useful for other language servers.

Here is an example (bottom left) for the message sent by Metals.

Untitled

Fix https://github.com/helix-editor/helix/issues/5524


pascalkuthe (dismissed):
LGTM!

It's nice that helix now covers this parts of the spec aswell

the-mikedavis (commented):
I think there should be separate config for showMessage and progressMessage: progress messages are too verbose and distracting IMO but messages sent through showMessage are usually more important and meant for the user to read.

`rust-analyzer` and `erlang-ls` are two servers in the wild that behave like this:

* rust-analyzer uses showMessage for things like saying that the config is invalid: https://github.com/rust-lang/rust-analyzer/blob/5306eb06cc2de0566ca789ee3863b78059ee2376/crates/rust-analyzer/src/bin/main.rs#L173-L183
* erlang-ls uses showMessage to say that the config file is missing and how to set one up: https://github.com/erlang-ls/erlang_ls/blob/1a66e8798bb2c33dc1fefa46f054760d8edb767d/apps/els_core/src/els_config.erl#L358-L373

Both use progress messages that are very verbose about internal indexing processes which I don't want to see.

We might also want to consider putting the showMessage contents into a Popup instead of using the statusline so that you have to explicitly clear the notification with `C-c`.

ayoub-benali:
What should be the new setting name ? 
Progress is already using a missleading name. should I introduce a breaking change and make progress be controlled with 'display-progress-message' while show message would use the current name ?

the-mikedavis:
Yeah I think you're right about the names. Ideally we should avoid breaking changes but I think progress messages would be better as `display-progress-messages` and showMessage messages as `display-messages`.

ayoub-benali:
@the-mikedavis  So I made the change but I think now a `R-breaking-change` label should be added to the PR, I couldn't do it myself.

ayoub-benali:
@the-mikedavis Is the popup support a must have for this PR or could we added it later depending on the user requests ? 
I think statusline is already better than nothing, since this LSP notification isn't even supported now

igor-ramazanov:
Can we include this into the next release?

AlexanderBrevig:
(sorry, I thought I was in another PR - still, I really do want this :) would make it a bit easier to use Helix as a client for testing new LSPs 👍🏻)

the-mikedavis:
This totally fell of my radar for a while - sorry about that! Looking at the most recent changes I think this looks good. We could experiment with a popup separately but I think the statusline is good for now. I've merged master and also added a commit to enable this config option by default. `false` seems like the right default to me for progress messages but I think language servers are not usually so verbose that this new option needs to be `false` too.

the-mikedavis (approved):

PR #5536, opened 2023-01-14
displaying diagnostic count on all open buffers in bufferline and for all buffers with diagnostics in the picker in addition with a red highlight

### LINKED ISSUE #5519: show the number of diagnostic errors and warnings on the buffer for all open buffers
Buffers, next to the file name, should show the number of errors and/or warning present in the current file. The number could be colored red. If there are more than 9 errors and/or warnings, "9+" could just be displayed, similar to on VSCode.

bindsdev:
Attempts to close #5519 

This PR introduces code that displays the diagnostic count for each open buffer in the buffers section in the bufferline. In addition, it also introduces a count for each of the diagnostic severities in the buffer picker for the buffers with diagnostics. Buffers with diagnostics also have a red highlight in the buffer line and buffer picker.

# Tasks

- [x] Add display for number of diagnostics for each open buffer in the bufferline.

	*Example Image*:
	image

- [x] Add display for number of each diagnostic severity for each buffer with diagnostics in the buffer picker.

	*Example Image*:
	image

- [x] Add red highlight for buffers with diagnostics in the bufferline.

	*Example Image*:
	image

- [x] Add red highlight for each buffer with diagnostics in the buffer picker. 

	*Example Image*:
	image


LeoniePhiline:
As a user, I would prefer the number of diagnostic types, e.g. "2 errors, 5 warnings", rather than "7 diagnostics".

Otherwise, the naming is weird, since the diagnostics are not what is counted. What is counted is the number of negative results from all diagnostics that have been run.

---

Side note: 

This would be suitable to be solved with icons, see https://github.com/helix-editor/helix/pull/2869

akabinds:
> As a user, I would prefer the number of diagnostic types, e.g. "2 errors, 5 warnings", rather than "7 diagnostics".
> 
> 
> 
> Otherwise, the naming is weird, since the diagnostics are not what is counted. What is counted is the number of negative results from all diagnostics that have been run.
> 
> 
> 
> ---
> 
> 
> 
> Side note: 
> 
> 
> 
> This would be suitable to be solved with icons, see https://github.com/helix-editor/helix/pull/2869
> 
> 

That is a fair take. It should be fairly easy to implement. 

I like the idea of the icons, but in relation to the bufferline, it could get pretty cluttered up there with all the icons and numbers. In that context, I believe that the single number or 9+ would be sufficient. Once the design of the bufferline is refined (if that is a discussion), that could be revisited and something similar to `bufferline.nvim` could result.

But, in the buffer picker, there is sufficient space so it should be fine for the icons and respective numbers to be there without causing clutter.

Once the icon support is added, the code implemented in this PR could be modified to use the icons of course.

akabinds:
I am at a bit of a roadblock with the last task for this PR. I started with directly editing the `Component` implementation for `FilePicker`, but any changes made there will apply to every one of the pickers it seems, and this issue is only pertaining to the buffer picker. I am looking for some guidance on how to highlight the buffers with diagnostics red specifically the buffer picker. I assume it might have to be something relating to one of the closures passed to `FilePicker::new`.

gabydd:
You might want to look at how it's done in https://github.com/helix-editor/helix/pull/5645

akabinds:
> You might want to look at how it's done in #5645

Looking at that PR, it got me thinking whether this would be better suited as a separate picker. Any thoughts?

akabinds:
The previous commit implements the red highlight for buffers with diagnostics. It is important to note that this does not work well at all with buffers who have long paths. This seems like something that is covered in #2377.

akabinds:
any word on this? waiting for a review.

PR #5534, opened 2023-01-14
Rewrite and refactor all documentation

David-Else:
Here is the (current state of the) rewrite of the Helix documentation following the discussion at
https://github.com/helix-editor/helix/discussions/5428

- [x] Improved readability, consistency, spelling, grammar, and best practices
      for software documentation
- [x] Added missing information and expanded / rewrote sections
- [x] Used triple tick `sh` for code blocks for text to be entered in the
      terminal
- [x] Formatted Markdown to 80 chars using Prettier.
- [x] Added some Table of Contents where I thought they were needed 

This has been a huge undertaking, it is possible that I have made errors in the
parts I have re-written due to gaps in my knowledge, so I hope as many people as
possible can read it and confirm it is correct.

Please note, you can view Markdown diffs properly by clicking a button:
https://docs.github.com/en/repositories/working-with-files/using-files/working-with-non-code-files#rendering-differences-in-prose-documents

rendered_prose_diff
rendered_prose_changes

Another way to easily read it is to go to the `Files Changed` section with the diffs and click `View File` for each section:

Screenshot from 2023-01-14 15-52-29

TODO:

- [x] Refine parts of this draft pull request, I am not 100% happy with all of it
- [x] Fix the Windows section of Installing from source (waiting for feedback from @CptPotato)

NOTE: Part 2 to follow, splitting up and reorganizing the files into a better chapter structure.

PR #5545, opened 2023-01-15
Format documentation for better diffing of #5534

David-Else:
This pull request simply formats all the current documentation using the same settings as https://github.com/helix-editor/helix/pull/5534 . There are absolutely no other changes applied.

I am waiting for a reply from @archseer https://github.com/helix-editor/helix/pull/5534#discussion_r1070638074 to confirm if he thinks doing this is the best solution to make reviewing easier by making the changes easier to see, but I put in this pull request in advance to save time if the idea is approved.

I have included the `.prettierrc` file for future reference.

@the-mikedavis I don't expect there to be any problems created by this for the mdBook generation, but maybe you could confirm that?


thomkeh:
Don't you get the best diffs with [one sentence per line](https://nick.groenen.me/notes/one-sentence-per-line/)? (Or, more generally with “semantic line breaks”, so breaking at commas or colons or the start of a new clause.)

David-Else:
It seems this fix might be preferable to GitHub prose diff mode.

the-mikedavis:
The problem is with bundling content changes with auto-formatting changes - switching to different auto-formatting settings won't really make a difference.

I could go through the #5543 changes and rewrite them on master without the formatting changes and force-push that to your branch if you'd like. (This would be my way to work through and review the changes as well.)

David-Else:
@the-mikedavis I am quite new to advanced Git and don't really understand what force pushing to my branch in GitHub would mean, would that result with a new commit from you appearing with your new version on my branch? If so, go for it!

You will have a fun job moving all the text back to the old non formatted version, a good test of your Helix keyboard skills :)

1. Do I need to do anything now (like merge in new changes from upstream master) or just wait to hear something from you over on https://github.com/helix-editor/helix/pull/5534 ?
2. Shall we get what I have done onto the master and then I make a new pull request for the other changes I detail in https://github.com/helix-editor/helix/pull/5534 ?

Cheers.

the-mikedavis:
Yep it would just be rewritten commit(s) on your branch in that PR. You would still be the author on the commits too. I'll rebase the changes on master so no need to do anything to the branch. I'll try to take a look at this in the next few days.

> Shall we get what I have done onto the master and then I make a new pull request for the other changes I detail in #5534?

Yeah we can start with the changes in that branch and make improvements from there once it's merged

pascalkuthe:
If I understood this comment https://github.com/helix-editor/helix/pull/5534#issuecomment-1407474833 right this PR can be closed now. Correct?

PR #5552, opened 2023-01-16
doc: add a note about nested bindings in key remapping

divarvel:
It was not clear (to me) that minor modes were configurable in the keymap configuration (it can be inferred from the `ga` mapping in the example, but it's easy to overlook).


divarvel:
I also spotted a small issue in the `configuration` page, fixed in a separate commit. lmk if that belongs to a separate PR

the-mikedavis (approved):

Issue #4811, opened 2022-11-19
Extend how to do key remapping in the docs, in particular how to do keybindings for the minor modes


Kenneth-Mata:
i tried a few combinations to bind  ``space space`` to 'write' without much success, also ,in some parts around the issues here there's space.ret, while in others and the docs it mentions enclosing especial keys in " ".

divarvel:
This was improved a bit in https://github.com/helix-editor/helix/pull/5552 (an explicit note in the "key remapping" page. I can also do that in the "keymap" page, in the minor modes paragraphs.

Issue #3835, opened 2022-09-13
Globally invert up and down (j and k)


jharrilim:
I personally like to use j as up and k as down. I can configure it for normal mode:

```toml
[keys.normal]
j = "move_line_up"
k = "move_line_down"
```

However, I cannot currently configure it for view mode or window mode. It would be a neat enhancement to be able to configure these, but potentially in a more global fashion. That way if any new features added movement capabilities, inversion would automatically be applied.

the-mikedavis:
> I cannot currently configure it for view mode or window mode

This should be possible, for example

```toml
[keys.normal.z]
j = "scroll_up"
k = "scroll_down"
```

I think it's distasteful to add config options for things achievable through rebinding.

jharrilim:
That's fair! Didn't know I could do that. I misunderstood the section in [this page](https://docs.helix-editor.com/remapping.html) where it says `# At most one section each of 'keys.normal', 'keys.insert' and 'keys.select'`. I can make a PR to add examples of minor mode rebindings to the docs if you'd like.

the-mikedavis:
Ah yeah, I think that's saying that you don't want to have duplicate `[keys.normal]` sections for example since the later one will override the former.

Showing an example of mapping sequences like z/Z/C-w would be great! I think that's a relatively common ask in the matrix

PR #5562, opened 2023-01-17
Doc fixes

### LINKED ISSUE #5145: small errors on Helix keymap docs
at https://docs.helix-editor.com/keymap.html#unimpaired

[d | Go to previous diagnostic (LSP) | goto_prev_diag
]d | Go to next diagnostic (LSP) | goto_next_diag

^Those are good but reversed so swiping them will help ( the "next" should be above )

.....

]T | Go to next test (TS) | goto_next_test
]T | Go to previous test (TS) | goto_prev_test
^Those are duplicated

divarvel:
For the documentation on minor modes, i think it's useful to link from `keymap` to `remapping` so that readers know those modes are configurable like other bindings.

In `remapping`, I chose to keep the nested bindings in the first example, to showcase both inline toml tables and regular tables.

This should fix #5145, #3835 and #4811

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -166,6 +166,9 @@ These sub-modes are accessible from normal mode and typically switch back to nor
 | `Ctrl-w` | Enter [window mode](#window-mode)                  | N/A            |
 | `Space`  | Enter [space mode](#space-mode)                    | N/A            |
 
+These modes (except command mode) can be configured by
+[remapping keys](https://docs.helix-editor.com/remapping.html#minor-modes).
+
 #### View mode
 
 Accessed by typing `z` in [normal mode](#normal-mode).
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -305,10 +305,10 @@ Mappings in the style of [vim-unimpaired](https://github.com/tpope/vim-unimpaire
 
 | Key      | Description                                  | Command               |
 | -----    | -----------                                  | -------               |
-| `[d`     | Go to previous diagnostic (**LSP**)          | `goto_prev_diag`      |
 | `]d`     | Go to next diagnostic (**LSP**)              | `goto_next_diag`      |
-| `[D`     | Go to first diagnostic in document (**LSP**) | `goto_first_diag`     |
+| `[d`     | Go to previous diagnostic (**LSP**)          | `goto_prev_diag`      |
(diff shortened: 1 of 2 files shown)


book/src/keymap.md
```diff
@@ -166,6 +166,9 @@ These sub-modes are accessible from normal mode and typically switch back to nor
 | `Ctrl-w` | Enter [window mode](#window-mode)                  | N/A            |
 | `Space`  | Enter [space mode](#space-mode)                    | N/A            |
 
+These modes (except `v` and `:`) can be configured by
```
the-mikedavis, line 169: Command mode `:` can't currently be remapped but you can remap select mode - both the bindings in select mode:

```toml
[keys.select]
w = "extend_next_long_word_start"
```

and the binding to switch into select-mode itself:

```toml
[keys.normal]
esc = "select_mode"
```

divarvel: right

book/src/keymap.md
```diff
@@ -325,8 +328,8 @@ Mappings in the style of [vim-unimpaired](https://github.com/tpope/vim-unimpaire
 | `[g`     | Go to previous change                        | `goto_prev_change`    |
 | `[G`     | Go to first change                           | `goto_first_change`   |
 | `]G`     | Go to last change                            | `goto_last_change`    |
```
the-mikedavis, line 330: it looks like these two could be swapped as well for consistency

```suggestion
| `]G`     | Go to last change                            | `goto_last_change`    |
| `[G`     | Go to first change                           | `goto_first_change`   |
```

book/src/keymap.md
```diff
@@ -166,6 +166,9 @@ These sub-modes are accessible from normal mode and typically switch back to nor
 | `Ctrl-w` | Enter [window mode](#window-mode)                  | N/A            |
 | `Space`  | Enter [space mode](#space-mode)                    | N/A            |
 
+These modes (except `:`) can be configured by
```
the-mikedavis, line 169: minor nit but can we call this "command mode" instead of `:`? I think using `:` might make readers think that you can't remap the binding for the `command_mode` command to something other than `:` rather than the commands within command mode

```suggestion
These modes (except command mode) can be configured by
```

divarvel: yeah, that makes more sense

the-mikedavis (approved):

PR #5567, opened 2023-01-18
Introduce a PR template based on Bevy's

### LINKED ISSUE #4002: Create a PR template to help improve ease of reviewing

kirawi:
# Objective

- Resolve #4002

## Solution

-  Copied Bevy's PR format. It's simple and can also be used to automate changelogs.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,20 @@
+# Objective
+
+- Describe the objective or issue this PR addresses.
+- If you're fixing a specific issue, say "Fixes #X".
+
+## Solution
+
+- Describe the solution used to achieve the objective above.
+
+---
+
+## Changelog
+
+> This section is optional. If this was a trivial fix, or has no externally-visible impact, you can delete this section.
+
+- What changed as a result of this PR?
+- If applicable, organize changes under "Added", "Changed", or "Fixed" sub-headings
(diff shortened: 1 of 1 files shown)


pascalkuthe (commented):
I like this template 👍 I am not a fan of most PR templates because the tend to get in the way and are too static. This just provides a bit of structure that will make PRs a bit more uniform and easier too read without being too daunting.

We might want to introsucte a second template for trivia changes:
* documentation (typo fixes)
* theme changes
* langugue support (adding/changing tree sitter crammer or quieres or an LS)

kirawi:
Perhaps that could be a follow up PR?

pascalkuthe:
discussed this with @the-mikedavis I don't think we want contributors to add changelog entries. The rest of the template doesn't quite carry its weight.

We definitely want to expand the contributing.md and potentially add maybe pr templates for common tasks/review comments we make. I think it will need to be more tailored to our project (like one template for adding a new ts grammar/updating queries).

PR #2060, opened 2022-04-10
Add rulers option

DeviousStoat:
Attempts at a solution for https://github.com/helix-editor/helix/issues/1897

This adds a `rulers` option which is a list of column positions at which a color column should be drawn.
I added a `ui.virtual.ruler` theme value to set the theme of the color column. I am not sure if this is the best solution. I think this would need to be added to every theme.

I initially wanted to add the cursor column to this option as well but I thought it would be better as a different option as this needs to be drawn only on the focused window. And I didn't really find a good way to specify it in the option. I will probably open another PR for this soon if this sounds good to you.

The cool thing using the theme system is that we can set only the foreground color for the colorcolumn and I think it is actually really cool:

foreground

With `rulers = [20]` and `"ui.virtual.ruler" = "red"`

PR #5568, opened 2023-01-18
Attempting to fix rulers/cursor highlighting

luetage:
The issue: In #2060 a rulers options was added. This allows for both setting a traditional colorcolumn (background color) or highlighting the foreground at the column instead (text color). The two themes I’m using as example are the default Helix theme (colors foreground at ruler) and catppuccin_mocha (colors background at ruler). In the current state both versions work well, with the caveat/bug that the cursor at the colorcolumn has no visible background when setting the background color of the ruler. Following screenshots compare the current state and this draft:

original | fix
---|---
default-original | default-fix
mocha-original | mocha-fix

The default theme has a working ruler in the original, the fix breaks the ruler (characters at the column are no longer themed).
Catppuccin_mocha has a visible ruler in the original, but the cursor background is the background of the ruler and the character is unreadable. The fix themes the cursor again.

All I want to do is point out that the current implementation has problems for a majority of the themes, because most themes set the background color of the virtual ruler, not the foreground color. The only thing I did was switching it up and rendering rulers first. Apparently the text highlights are then able to overrule the ruler theming—which in turn breaks themes which set the foreground color of the ruler (a minority). I wonder whether there is a way to get both working properly.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -149,9 +149,9 @@ impl EditorView {
             Box::new(highlights)
         };
 
+        Self::render_rulers(editor, doc, view, inner, surface, theme);
         Self::render_text_highlights(doc, view.offset, inner, surface, theme, highlights, &config);
         Self::render_gutter(editor, doc, view, view.area, surface, theme, is_focused);
-        Self::render_rulers(editor, doc, view, inner, surface, theme);
 
         if is_focused {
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/3599

luetage:
Ahh, I’m sorry. I searched for the issue but apparently failed :/ Closing.

PR #5570, opened 2023-01-18
bump msrv to 1.63

### LINKED ISSUE #1557: Crash if double colon is input in path when changing directory
### Reproduction steps

Open hx.
type `:cd c::`

<details><summary>The error</summary>

```
thread 'main' panicked at 'range start index 2 out of range for slice of length 0', library\std\src\path.rs:713:24
stack backtrace:
   0:     0x7ff6bee7e459 - <unknown>
   1:     0x7ff6be8958ba - <unknown>
   2:     0x7ff6bee78208 - <unknown>
   3:     0x7ff6bee81bb2 - <unknown>
   4:     0x7ff6bee81652 - <unknown>
   5:     0x7ff6bed443ab - <unknown>
   6:     0x7ff6bee822b7 - <unknown>
   7:     0x7ff6bee81de5 - <unknown>
   8:     0x7ff6bee7f687 - <unknown>
   9:     0x7ff6bee81d4f - <unknown>
  10:     0x7ff6bef01f90 - <unknown>
  11:     0x7ff6bef02127 - <unknown>
  12:     0x7ff6bee7a1fd - <unknown>
  13:     0x7ff6bee7a002 - <unknown>
  14:     0x7ff6bee7bd5a - <unknown>
  15:     0x7ff6beac3aad - <unknown>
  16:     0x7ff6bead8c6f - <unknown>
  17:     0x7ff6beb3e66c - <unknown>
  18:     0x7ff6bea62b24 - <unknown>
  19:     0x7ff6bea6ae39 - <unknown>
  20:     0x7ff6bea6ccb8 - <unknown>
  21:     0x7ff6beafba1b - <unknown>
  22:     0x7ff6bec33a1c - <unknown>
  23:     0x7ff6bed3e1b6 - <unknown>
  24:     0x7ff6bed4d468 - <unknown>
  25:     0x7ff6bed210b6 - <unknown>
  26:     0x7ff6bed37a6f - <unknown>
  27:     0x7ff6bed3c224 - <unknown>
  28:     0x7ff6bed23d5e - <unknown>
  29:     0x7ff6bed1b116 - <unknown>
  30:     0x7ff6bed38ffa - <unknown>
  31:     0x7ff6bee7debc - <unknown>
  32:     0x7ff6bed24127 - <unknown>
  33:     0x7ff6beefddac - <unknown>
  34:     0x7ff838da7034 - BaseThreadInitThunk
  35:     0x7ff839ae2651 - RtlUserThreadStart
```

</details>

### Environment

- Platform: Windows
- Terminal emulator: PowerShell
- Helix version: v0.5.0-395-g0b55b21

### LINKED ISSUE #5432: Compilation error
### Summary

Currently the project does not compile because of the 'ignore v0.4.19'  package - this package requires 'scoped_threads' - an unstable feature that only stabilized in Rust 1.63. Unfortunatelly the rust-toolchain.toml overrides the current version down to 1.61. That causes the compilation to fail.

This can be fixed by bumping version up to at least 1.63.0

### Reproduction Steps

$ git clone https://github.com/helix-editor/helix helix
$ cd helix
$ cargo install --path helix-term 

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

gnome-terminal

### Helix Version

22.12-95-gf2fff9b9

pascalkuthe:
Firefox 109 was released with a MSRV of 1.63.
In accordance with the MSRV policy this PR bumps the MSRV to 1.63.

Fixes #1557
Fixes #5432

### DIFF
--- a/.github/workflows/msrv-rust-toolchain.toml
+++ b/.github/workflows/msrv-rust-toolchain.toml
@@ -1,3 +1,0 @@
-[toolchain]
-channel = "1.61.0"
-components = ["rustfmt", "rust-src"]
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -69,13 +69,15 @@ use grep_searcher::{sinks, BinaryDetection, SearcherBuilder};
 use ignore::{DirEntry, WalkBuilder, WalkState};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 
+pub type OnKeyCallback = Box<dyn FnOnce(&mut Context, KeyEvent)>;
+
 pub struct Context<'a> {
     pub register: Option<char>,
     pub count: Option<NonZeroUsize>,
     pub editor: &'a mut Editor,
 
     pub callback: Option<crate::compositor::Callback>,
-    pub on_next_key_callback: Option<Box<dyn FnOnce(&mut Context, KeyEvent)>>,
+    pub on_next_key_callback: Option<OnKeyCallback>,
     pub jobs: &'a mut Jobs,
 }
 
--- a/helix-term/src/lib.rs
+++ b/helix-term/src/lib.rs
@@ -1,3 +1,5 @@
+#![allow(clippy::type_complexity)]
+
 #[macro_use]
 extern crate helix_view;
 
--- a/helix-term/src/lib.rs
+++ b/helix-term/src/lib.rs
@@ -1,5 +1,3 @@
-#![allow(clippy::type_complexity)]
-
 #[macro_use]
 extern crate helix_view;
 
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,19 +9,11 @@ on:
 
 jobs:
   check:
-    name: Check
+    name: Check (msrv)
     runs-on: ubuntu-latest
-    strategy:
-      matrix:
-        rust: [stable, msrv]
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-
-      - name: Use MSRV rust toolchain
-        if: matrix.rust == 'msrv'
-        run: cp .github/workflows/msrv-rust-toolchain.toml rust-toolchain.toml
-
       - name: Install stable toolchain
         uses: helix-editor/rust-toolchain@v1
         with:
@@ -44,7 +36,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.61
+        uses: dtolnay/rust-toolchain@1.63
 
       - uses: Swatinem/rust-cache@v2
 
@@ -73,7 +65,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.61
+        uses: dtolnay/rust-toolchain@1.63
         with:
           components: rustfmt, clippy
 
@@ -98,7 +90,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.61
+        uses: dtolnay/rust-toolchain@1.63
 
       - uses: Swatinem/rust-cache@v2
 
--- a/helix-term/src/compositor.rs
+++ b/helix-term/src/compositor.rs
@@ -7,6 +7,7 @@ use helix_view::graphics::{CursorKind, Rect};
 use tui::buffer::Buffer as Surface;
 
 pub type Callback = Box<dyn FnOnce(&mut Compositor, &mut Context)>;
+pub type SyncCallback = Box<dyn FnOnce(&mut Compositor, &mut Context) + Sync>;
 
 // Cursive-inspired
 pub enum EventResult {
--- a/helix-term/src/job.rs
+++ b/helix-term/src/job.rs
@@ -5,9 +5,12 @@ use crate::compositor::Compositor;
 use futures_util::future::{BoxFuture, Future, FutureExt};
 use futures_util::stream::{FuturesUnordered, StreamExt};
 
+pub type EditorCompositorCallback = Box<dyn FnOnce(&mut Editor, &mut Compositor) + Send>;
+pub type EditorCallback = Box<dyn FnOnce(&mut Editor) + Send>;
+
 pub enum Callback {
-    EditorCompositor(Box<dyn FnOnce(&mut Editor, &mut Compositor) + Send>),
-    Editor(Box<dyn FnOnce(&mut Editor) + Send>),
+    EditorCompositor(EditorCompositorCallback),
+    Editor(EditorCallback),
 }
 
 pub type JobFuture = BoxFuture<'static, anyhow::Result<Option<Callback>>>;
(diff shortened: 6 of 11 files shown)


helix-term/src/lib.rs
```diff
@@ -1,3 +1,5 @@
+#![allow(clippy::type_complexity)]
```
pascalkuthe, line 1: clippy generated a bunch of warnings here which would require introducing new types.
I held off doing that to avoid introducing too many changes here.

pascalkuthe:
There is only one related TODO in the codebase (in `view.rs`) but that code will be replaced #5420 anyway so I didn't fix the TODO to avoid merge conflicts.

.github/workflows/msrv-rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.61.0"
+channel = "1.63.0"
```
the-mikedavis, line 2: I think this file doesn't end up being used since we switched to dtolnay/rust-toolchain. This can probably be removed now

the-mikedavis: Ah actually it looks like we don't use stable rust for the stable CI check: msrv and stable checks are running the exact same thing. In the `check` job, we could use `matrix: [stable, 1.63]` and then:

```yml
- uses: dtolnay/rust-toolchain@master
  with:
    toolchain: ${{ matrix.toolchain }}
```

to replace the `helix-editor/toolchain` action.

helix-term/src/lib.rs
the-mikedavis: I think it would be ok to resolve those in this PR as well as long as it's a separate commit from the existing changes

CptPotato:
One thing to note is that I'm unable to build helix on windows with a rust toolchain lower than `1.64`.

This is due to the depenency on the `windows` crate:

```
error: failed to compile `helix-term v0.6.0 ([...]\helix\helix-term)`, intermediate artifacts can be found at `[...]\helix\target`

Caused by:
  package `windows v0.43.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.61.0
```

.github/workflows/msrv-rust-toolchain.toml
pascalkuthe: I am not sure we actually want to have a stable build.
Whenever a new rust version releases you get new clippy lints which means that PR piplines could just randomly start failing. This is especially problematic as clippy often inrdocues new lints which can only be fixed with new language features (example `then_some`). 

It would also be annoying development wise because the rust_toolchain file ist pinned to the `MSRV` so we only know about clippy lints once the CI fails. I think it honestly makes more sense to only run CI on `MSRV`.

Rust breaking backwars incomptatbility is rare enough that we don't need to worry about it and I would rather fix new clippy lints when the MSRV is bumped like in this PR so I don't see much value in running a stable CI. This would also speedup CI times which is not super important but a nice bonus

pascalkuthe:
> One thing to note is that I'm unable to build helix on windows with a rust toolchain lower than `1.64`.
> 
> This is due to the depenency on the `windows` crate:
> 
> ```
> error: failed to compile `helix-term v0.6.0 ([...]\helix\helix-term)`, intermediate artifacts can be found at `[...]\helix\target`
> 
> Caused by:
>   package `windows v0.43.0` cannot be built because it requires rustc 1.64 or newer, while the currently active rustc version is 1.61.0
> ```

That is unfortunate. Probably just another MSRV bump on a minor release. Using `cargo install --locked` will fix the problem (already part of the documentation). We stick with the MSRV of firefox so that helix can be build on all distros (except debian). As all distros usually want to package the newest firefox fore security reasons. This is documented in our MSRV policy.

Firefox will raise it's MSRV to 1.65 in mid feburary so this Problem should also not exist for long.

.github/workflows/msrv-rust-toolchain.toml
the-mikedavis: Ah good point, I agree. I can't find the history for it but if I remember correctly, we pinned to 1.61 because we had a bunch of CI failures from a new rust version release being used automatically that ❌'d a bunch of open PR's CI https://github.com/helix-editor/helix/commit/aa87adf54b0b8dec07baa257aed71d09c3efb119. I would be in favor of just removing the "stable" branch from CI

pascalkuthe:
I have changed the CI so it only runs on MSRV (the same behaviour as before but we now only have one check). However github still expects the old checks so the CI can not succed for this PR. I am not sure where to change that (or maybe it changes automatically once this PR is merged into master?).

the-mikedavis:
We'll need @archseer to adjust that: there are branch protection rules in the settings that will need to be updated.

This will be a sort of breaking change for the CI runs for each PR: each PR will need to be rebased on this to satisfy the new check requirements. I wonder if we can side-step that by renaming the new "Check" job as "Check (msrv)"? Then we could just remove the "Check (stable)" from the required passing checks and no PRs would need to be rebased. I'm not sure if it's based on the naming like that though

pascalkuthe:
> This will be a sort of breaking change for the CI runs for each PR: each PR will need to be rebased on this to satisfy the new check requirements. I wonder if we can side-step that by renaming the new "Check" job as "Check (msrv)"? Then we could just remove the "Check (stable)" from the required passing checks and no PRs would need to be rebased. I'm not sure if it's based on the naming like that though


That is something I didn't think about. I am not sure how to solve this.
I would except the checks to be ID based but if we can work around the problem with names that would nb nice.
I tried understanding the GH docs but they are not exactly written well.

It would be neat to remove the duplicate CI run but it's not big deal and probably not worth breaking PRs over.
If there is no way around that I will revert to simply running the MSRV CI twice.

What do you think @archseer?

the-mikedavis:
Hah, it works! https://github.com/the-mikedavis/ghactionstest/pull/2

I guess it must be naming-based internally. I suppose that makes sense since they're reading the workflows out of the yaml so it would be hard to have a recognizable workflow ID

pascalkuthe:
> Hah, it works! [the-mikedavis/ghactionstest#2](https://github.com/the-mikedavis/ghactionstest/pull/2)
> 
> I guess it must be naming-based internally. I suppose that makes sense since they're reading the workflows out of the yaml so it would be hard to have a recognizable workflow ID

Thanks for testing this! I have updated the PR accordingly.
I somehow taught it would be a huge pain to setup but I just realized I could have performed the same test you did by simply pushing the same changes I just did and test if the CI passes. Sorry about that and thanks for checking :)

@archseer still needs to remove the `Check (stable)` requirement (which is already fine because currently `Check (msrv)` is identical anyway) but once that is done, then this should be good to go.

helix-term/src/commands.rs
```diff
@@ -69,13 +69,15 @@ use grep_searcher::{sinks, BinaryDetection, SearcherBuilder};
 use ignore::{DirEntry, WalkBuilder, WalkState};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 
+pub type KeyCallback = Box<dyn FnOnce(&mut Context, KeyEvent)>;
```
the-mikedavis, line 72: very minor naming nit but I think this could be clearer as `OnNextKeyCallback` to match the fields it is used on

pascalkuthe: fixed

the-mikedavis (approved):

PR #5571, opened 2023-01-18
Kanagawa: theme rulers and some miscellaneous fixes

luetage:
This PR introduces following changes

* Add sumiInk2 color for theming rulers – at the moment only »virtual« is set, which themes both whitespace characters and rulers. This change themes the foreground for whitespace characters and the background for rulers.
* Dark background for menu and help – while the current version works for the command mode menu and help, these UI elements are also in use for auto-completion. Using the main background for such popups is a little confusing, because there is no border to separate menu content from main content.
* Bold modifier for insert and select mode in statusline – normal mode already uses bold and it is a matter of taste, but I’d say it looks more consistent this way.
* Dark foreground color for window separators – the original nvim theme seems to use a dark color, here the default foreground color is in use. Personally I think it looks best to set both foreground and background color of window separators to sumiInk0, so the separators meld with the statusline and there is no gap between statuslines of adjacent windows. But no other theme handles it this way so I didn’t do it either.

@zetashift and @leonqadirie please take a look. This is naturally all up for discussion and apart from theming the rulers none of it is addressing anything serious.

### DIFF
--- a/runtime/themes/kanagawa.toml
+++ b/runtime/themes/kanagawa.toml
@@ -30,7 +30,8 @@
 "ui.help" = { fg = "fujiWhite", bg = "sumiInk1" }
 "ui.text" = "fujiWhite"
 "ui.text.focus" = { fg = "fujiWhite", bg = "waveBlue1", modifiers = ["bold"] }
-"ui.virtual" = "waveBlue1"
+"ui.virtual.whitespace" = "waveBlue1"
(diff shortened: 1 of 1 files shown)


leonqadirie:
I see @zetashift as the owner here, but LGTM.
Thanks!

zetashift:
Ah! I see it as a community effort :P.

Thank you for these fixes, they look nice, it's a 👍 for me!

PR #5575, opened 2023-01-18
highlight non-bar cursors

### LINKED ISSUE #1568: Render selection 1 character wider when there is a bar or underline cursor with selection to the right
image

pascalkuthe:
Fixes #1568

The logic is a bit more complicated than originally described in the issue. I think all non-block cursors don't show well that the cursor is part of the selection. Therefore, the cursor is always highlighted as a selection now. To avoid weird looking regressions in insert mode I added two special cases so that `a` (and `i` with just a single grapheme selected) don't regress.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -386,7 +386,14 @@ impl EditorView {
             if range.head > range.anchor {
                 // Standard case.
                 let cursor_start = prev_grapheme_boundary(text, range.head);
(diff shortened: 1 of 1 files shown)


pickfire (approved):
Looks good to me. Thanks.

larsblumberg:
Thanks Pascal, this fixes a visual annoyance and had learned to live with. Looking forward to the next release as I didn't install dev builds so far. Maybe something I should practice though :-)

PR #5576, opened 2023-01-18
highlight(scala): update to fix potential crash

eed3si9n:
Ref https://github.com/tree-sitter/tree-sitter-scala/pull/164

tree-sitter-scala has recently add a fix to workaround segv crashes in other editors. Not sure if it happens to Helix as well, but it's probably a good idea to use the latest.

### DIFF
--- a/runtime/queries/scala/highlights.scm
+++ b/runtime/queries/scala/highlights.scm
@@ -112,6 +112,9 @@
 (generic_function
   function: (identifier) @function)
 
+(interpolated_string_expression
+  interpolator: (identifier) @function.call)
+
 (
   (identifier) @function.builtin
   (#match? @function.builtin "^super$")
--- a/runtime/queries/scala/highlights.scm
+++ b/runtime/queries/scala/highlights.scm
@@ -113,7 +113,7 @@
   function: (identifier) @function)
 
 (interpolated_string_expression
-  interpolator: (identifier) @function.call)
+  interpolator: (identifier) @function)
 
 (
   (identifier) @function.builtin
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
It would be good to port this change from the highlights too since we're updating: https://github.com/tree-sitter/tree-sitter-scala/commit/71b408eed0703a30dbd53a57ce52ae0631af38a4#diff-d485a982e458ef8da2cc203585065b7542665cb80b78d230b1e8f77ea25825d4R96-R97

(The capture would be `@function` instead of `@function.call` since our captures are a little different from the ones used in that repo)

archseer:
The highlights need to be updated in our local file: https://github.com/helix-editor/helix/blob/master/runtime/queries/scala/highlights.scm

eed3si9n:
> The highlights need to be updated in our local file:
> 
> https://github.com/helix-editor/helix/blob/efeec1282cd82e255f54b10e94c51f51294598eb/runtime/queries/scala/highlights.scm#L112-L114

Ah! That makes sense. Let me update the PR.

runtime/queries/scala/highlights.scm
```diff
@@ -112,6 +112,9 @@
 (generic_function
   function: (identifier) @function)
 
+(interpolated_string_expression
+  interpolator: (identifier) @function.call)
```
the-mikedavis, line 116:
```suggestion
  interpolator: (identifier) @function)
```

We use different captures: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

eed3si9n: Fixed f43f3b494d54031c44cee0165675c38ad7b721c7

the-mikedavis (approved):

PR #2412, opened 2022-05-05
Find file (space .)

pickfire:
Based on doom emacs find-file but also wanted to show an alternative to
file explorer #2377.

Screenshot_20220506_004049

@cossonleo this is what I mean, just got some motivation to do a draft to show you what I meant, but maybe this complement the tree explorer, this is more of a per directory navigation.

Still missing some basic stuff that I wanted to see

- [ ] recurse find file
- [ ] edit new file
- [ ] prompt color
- [ ] file color
- [ ] file metadata

PR #5577, opened 2023-01-18
Add % current file register

pickfire:
Useful to open files relative to current file.

[asciicast](https://asciinema.org/a/hOpVj7GcalGP6SvfRI9H9UHRn)

The experience is worse especially on files that have huge paths, like python virtual environment, this allows browsing relative files easier, while being a lot easier to implement compared to find file. https://github.com/helix-editor/helix/pull/2412

To create new file from current file directory.

    :o <C-r>%<C-w>new-file<ret>

Compared to previous method (mouse copy-paste or manually type - and I usually manually type 1-7 directories)

    :o <mouse select file name><middle click paste><manually type full path if truncated><C-w>new-file<ret>


archseer:
Duplicate of https://github.com/helix-editor/helix/pull/3008 was replaced by https://github.com/helix-editor/helix/pull/3393

pickfire:
I think #3393 is slightly different and a lot harder to use compared to this scenario, the user has to type `%val{dirname}` (16 keystrokes) as compared to `<C-r>%<C-w>` (6 keystrokes).

The one implemented here is slightly similar to what @QiBaobin implemented, except being pull based (when user select registers) rather than push based (when user change view).

the-mikedavis:
The register could be useful for prompts and pickers where variable substitution might not be done (it might not make sense to replace variables if the prompt is a regex for example).

Kakoune has a special register for this as well: https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers

helix-view/src/editor.rs
```diff
@@ -1469,6 +1469,17 @@ impl Editor {
...
     }
+
+    /// Update read-only registers.
+    pub fn update_registers(&mut self) {
+        let doc = doc!(self);
+        if let Some(path) = doc.relative_path() {
+            self.registers
+                .write('%', vec![path.to_string_lossy().to_string()]);
+        } else {
+            self.registers.clear('%');
```
the-mikedavis, line 1480: To match kakoune we could write the scratch buffer name into the register when it doesn't have a path

the-mikedavis (commented):
Kakoune has an interesting way of passing state like the current buffer name to register functions: registers are defined with an anonymous function that takes `Context` and returns the value of the register. This is helpful for other special registers like `0..9` acting as regex captures. We don't need to emulate that exactly but I'm curious if we can implement this in a way that's easier to add the other special registers later

helix-view/src/editor.rs
pickfire: Yeah, I guess that's fine too.

helix-view/src/register.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+/// Currently just wraps a `BTreeMap` of `Register`s
+#[derive(Debug)]
+pub struct Registers {
+    inner: BTreeMap<char, Register>,
+}
+
+impl Registers {
+    pub fn read(&self, doc: &Document, name: char) -> Option<Vec<String>> {
```
pickfire, line 27: @the-mikedavis I did what you mentioned but now I don't like the API to always return an owned value, we need to keep cloning.

PR #5580, opened 2023-01-18
feat(picker): add directory picker

### LINKED ISSUE #4430: Implement a fuzzy dir finder
I use extensively a fuzzy dir finder in kakoune I made with skim to quickly cd into a project directory and then fuzzy pick a file from there.

I have the option to start the finder from a given dir (by default it fuzzy finds from the current directory).

So there could be a fuzzy dir picker accessible from the space mode, and a command accepting argument with completion mode on the prompt mode, but we can rely on the `:cd` command for that and then pick in space mode in 2 steps.

It should ne reasonably easy to implement on top of the actual file picker (choose dir instead of file, cd instead of open).

What do you think ?

Lenovsky:
Add the directory picker into the space mode that changes the current working directory to the selected path.

Closes #4430

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -269,6 +269,7 @@ impl MappableCommand {
         command_mode, "Enter command mode",
         file_picker, "Open file picker",
         file_picker_in_current_directory, "Open file picker at current working directory",
+        directory_picker, "Open directory picker",
         code_action, "Perform code action",
         buffer_picker, "Open buffer picker",
         jumplist_picker, "Open jumplist picker",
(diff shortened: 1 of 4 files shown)


pascalkuthe:
This seems a bit nieche to use up another key binding. What do you think k of just opening this directory picker when using `:cwd` without an argument?

Lenovsky:
Agree, that makes more sense -- I'll work on that.

pascalkuthe:
Closing as this PR has gone stale and our previous review is still unaddressed. Thanks you for contributing

markstos:
@Lenovsky I think having ":cd" with no arg open a directory picker would be a natural feature if you want to pick this back up.

PR #5581, opened 2023-01-18
move ui and commands to helix-view

### LINKED ISSUE #5555: Move commands into the `helix-view` crate
This has been discussed before in https://github.com/helix-editor/helix/issues/39#issuecomment-866446894 but it would be a large effort, so this issue focuses specifically on the change for moving commands into `helix-view`.

Commands currently exist in `helix-term` but should be moved to `helix-view`. Moving commands to `helix-view` should make it possible to switch between a terminal backend (`helix-term`) and a GUI backend. These changes may also help the macro keybindings effort (https://github.com/helix-editor/helix/pull/4709) and fix some bugs with the command palette (https://github.com/helix-editor/helix/pull/5294).

The difficult part of this change is that the commands currently manipulate the UI directly for creating Pickers and other UI elements or setting callbacks. The Compositor may need to be refactored as a trait and then components can be implemented separately by TUI and GUI backends.

pascalkuthe:
Closes #5555

This PR is super simple. All it does is:

* Move relevant files to `helix-view`
* replace `helix_term::` with `crate::`
* replace `helix_view::` with `crate::` (unfortunate that rust doesn't allow this)
* replace `tui::` with `helix_tui::` (seems more consistent, but the main reason is that RA for some reason is unable to resolve these imports evenotugh rustc can which leads to terrible dev experience)
* move a few additional small functions from `helix-term` that are used by commands

The primary focus of this PR is to avoid git conflicts. Git should handle most of these changes fine if a PR is rebased
altough there can be some conflicts in the `use` statements. However, compared to other PRs the changes here are not
that large and should be easy to resolve.

These changes will not only allow experimentation with GUIs (#39) but also unblocks other efforts (see #5555).
However, this PR implements not abstraction over the actual terminal functionality yet.
I think the best approach would be a `cfg` based approach opposed to traits as used by @archser in the `gui` branch.
Such an approach avoids many breaking changes (and making every function/struct generic).
This abstraction can be added in the future with smaller less disruptive PRs.


helix-view/src/ui/statusline.rs
```diff
@@ -137,24 +137,22 @@ where
...
-        helix_view::editor::StatusLineElement::Selections => render_selections,
-        helix_view::editor::StatusLineElement::PrimarySelectionLength => {
-            render_primary_selection_length
-        }
-        helix_view::editor::StatusLineElement::Position => render_position,
-        helix_view::editor::StatusLineElement::PositionPercentage => render_position_percentage,
-        helix_view::editor::StatusLineElement::TotalLineNumbers => render_total_line_numbers,
-        helix_view::editor::StatusLineElement::Separator => render_separator,
-        helix_view::editor::StatusLineElement::Spacer => render_spacer,
+        crate::editor::StatusLineElement::Mode => render_mode,
```
archseer, line 140: Let's add a `use crate::editor::StatusLineElement` before the `match`

pascalkuthe: Funnily enough the use was already there it just wasn't being used, fixed

pascalkuthe:
@archseer I kept this PR minimal on purpose so right now it doesn't hide terminal specific code behind cfg attributes. I wasn't 100% sure what the best way to do that would be and wanted to leave that design space for when we actually implement guis. This PR already unblocks the non-gui related work.

Are you okay with that or do you want me to add the cfg attrigites in this PR?

dead10ck (approved):
This seems fine to me, thanks!

helix-view/src/compositor.rs
```diff
@@ -35,21 +33,27 @@ impl<'a> Context<'a> {
...
+    fn render(&mut self, area: Rect, frame: &mut Surface, ctx: &mut Context);
+
     /// Should redraw? Useful for saving redraw cycles if we know component didn't change.
     fn should_update(&self) -> bool {
         true
     }
 
-    /// Render the component onto the provided surface.
-    fn render(&mut self, area: Rect, frame: &mut Surface, ctx: &mut Context);
-
```
gabydd, line 52: is this reorder necessary

helix-view/src/ui/markdown.rs
```diff
@@ -264,14 +264,14 @@ impl Markdown {
                             CodeBlockKind::Fenced(language) => language,
                             CodeBlockKind::Indented => "",
                         };
-                        let tui_text = highlighted_code_block(
+                        let helix_tui_text = highlighted_code_block(
```
gabydd, line 267: this and the one below don't need to be changed
```suggestion
                        let tui_text = highlighted_code_block(
```

helix-view/src/ui/markdown.rs
```diff
@@ -264,14 +264,14 @@ impl Markdown {
...
                             text.to_string(),
                             language,
                             theme,
                             Arc::clone(&self.config_loader),
                             None,
                         );
-                        lines.extend(tui_text.lines.into_iter());
+                        lines.extend(helix_tui_text.lines.into_iter());
```
gabydd, line 274:
```suggestion
                        lines.extend(tui_text.lines.into_iter());
```

helix-graphics/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+
+[dependencies]
+bitflags = "1.3"
+serde = { version = "1.0", features = ["derive"] }
+
+crossterm = { version = "0.25", optional = true }
+
+# TODO: graphics.rs tests rely on this, but we should remove that
+# [target.'cfg(not(target_arch = "wasm32"))'.dev-dependencies]
+# helix-tui = { path = "../helix-tui" }
```
gabydd, line 23: I think all our cargo.tomls have newlines at the end
```suggestion
# helix-tui = { path = "../helix-tui" }

```

helix-term/Cargo.toml
```diff
@@ -28,7 +28,7 @@ path = "src/main.rs"
 
 [dependencies]
 helix-core = { version = "0.6", path = "../helix-core" }
-helix-view = { version = "0.6", path = "../helix-view" }
+helix-view = { version = "0.6", path = "../helix-view", features = ["term"]}
```
gabydd, line 31:
```suggestion
helix-view = { version = "0.6", path = "../helix-view", features = ["term"] }
```

gabydd (commented):
just a couple really small nits otherwise looks good, though I assume this would need a rebase so no worries if this review isn't that helpful

gabydd (commented):
just a couple really small nits otherwise looks good, though I assume this would need a rebase so no worries if this review isn't that helpful

pascalkuthe:
I think I want to go with a slightly different approach here, this branch also has an insane amount of conflicts so I need to start over anyway. I will close this

PR #5593, opened 2023-01-19
Add more file-types for python

nabaco:
Add more file-types for Python. Based on https://stackoverflow.com/a/18032741/14986231 and https://github.com/neovim/neovim/blob/nightly/runtime/lua/vim/filetype.lua#L816

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -473,7 +473,7 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-html", rev = "29f53
 name = "python"
 scope = "source.python"
 injection-regex = "python"
-file-types = ["py"]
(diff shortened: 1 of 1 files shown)


Eloitor:
Could `.sage` be temporally added as a python file? The syntax of a sage file is basically the same as python. The only difference is that there are more build-in functions.

gabydd:
Apparently python language servers might not work the best with sage see https://ask.sagemath.org/question/50654/lsp-language-server-protocol-support-for-sagemath/ I would suggest making a new pr spacifically for sage support instead and test out how well the grammar and language servers work

the-mikedavis:
I'm not familiar with sage but we do something similar for starlark which is python-like: it's a separate language entry but it re-uses the python tree-sitter parser and queries https://github.com/helix-editor/helix/pull/2903

the-mikedavis (approved):

PR #5598, opened 2023-01-19
Remove apply_transaction helper

pickfire:
Not sure why there is this TODO, so I cleaned it up.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -26,7 +26,6 @@ use helix_core::{
     SmallVec, Tendril, Transaction,
 };
 use helix_view::{
-    apply_transaction,
     clipboard::ClipboardType,
     document::{FormatterError, Mode, SCRATCH_BUFFER_NAME},
     editor::{Action, Motion},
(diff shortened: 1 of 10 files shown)


the-mikedavis (commented):
Ah I forgot about this TODO: we removed the need for the `apply_transaction` helper a while ago but didn't want to complicate the diff. I think we're safe to remove it now 👍

There's one more reference of it in the integration test suite helpers to be removed

pickfire:
> There's one more reference of it in the integration test suite helpers to be removed

I can't seemed to find it with `rg apply_transaction`.

the-mikedavis:
Ah actually looks like there's two of them

https://github.com/helix-editor/helix/blob/edd0ba7f19c662555128c4b112faed0120d91197/helix-term/tests/test/helpers.rs#L133 and https://github.com/helix-editor/helix/blob/edd0ba7f19c662555128c4b112faed0120d91197/helix-term/tests/test/helpers.rs#L318

pickfire:
Oh, looks like I accidentally added `test/` to my git ignore, no wonder.

the-mikedavis (approved):

the-mikedavis:
Thanks for cleaning this up!

PR #5603, opened 2023-01-20
Accept +num flag for opening at line number

### LINKED ISSUE #5437: Accept `+num` flag for opening at line number
Lots of editors (including the heavyweights Vim and Emacs) accept a `+num` flag for opening a file at the line number specified by `num`. I know Helix can do this by appending `:num` to the file name, but both muscle memory and other applications (like the `v` command in less) rely on the plus syntax, so I suggest just adding it as a second option.

nabaco:
Resolves #5437 
Added basic logic for opening files with `+N` argument, where N is the line number.
Can be mentioned after each file. If mentioned several times, latest one will be taken.

This is a very basic implementation, since in Vim/NeoVim the `+` arguments runs a command in the editor.
With some help, I can try to change the implementation. Need to understand how I can use the `execute()` method upon startup. ([helix-term/src/commands.rs:164](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs#L164))

P.S: I'm new to Rust, so any comments about my code are more than welcome :)

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -17,6 +17,7 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub line_number: usize,
 }
 
 impl Args {
@@ -73,6 +74,16 @@ impl Args {
                         }
                     }
                 }
+                arg if arg.starts_with('+') => {
+                    let arg = arg.get(1..).unwrap();
+                    args.line_number = match arg.parse() {
+                        Ok(n) => n,
+                        _ => anyhow::bail!("bad line number after +"),
+                    };
+                    if args.line_number > 0 {
+                        args.line_number -= 1;
+                    }
+                }
                 arg => args.files.push(parse_file(arg)),
             }
         }
@@ -80,6 +91,10 @@ impl Args {
         // push the remaining args, if any to the files
         for arg in argv {
             args.files.push(parse_file(&arg));
+        }
+
+        if let Some(file) = args.files.first_mut() {
+            file.1.row = args.line_number;
         }
 
         Ok(args)
--- a/helix-term/src/main.rs
+++ b/helix-term/src/main.rs
@@ -73,6 +73,7 @@ FLAGS:
     -V, --version                  Prints version information
     --vsplit                       Splits all given files vertically into different windows
     --hsplit                       Splits all given files horizontally into different windows
+    +N                             Goto line number N
 ",
         env!("CARGO_PKG_NAME"),
         VERSION_AND_GIT_HASH,


nabaco:
Need to fix Clippy warnings. I'll do that, probably after the weekend.
I've managed to do an implementation where `+` executes any command, but it doesn't work in the form of `+11`. I'm suspecting `Shellwords::from(&str)` is swallowing the numbers from some reason.

Anyone willing to help me with this?

nabaco:
I've pushed my second implementation as a second commit, for review and discussion. In the second implementation, `+11` doesn't work but `+'open README.md'`, `+'goto 11'`, and `+g 11` work like a charm.
`+q` panics 😞 but I don't think it's important right now since I don't see `+q` being a very important or common usecase.

Anyway, any comments, suggestions, reviews are more than welcome.

nabaco (commented):
Here is the issue. I'm missing the number in `+11` since it's the first element. Need to either add an if here or insert "goto" into `command`

nabaco:
Fixed the numeric command issue in the final commit. Tested with the following commands:
```sh
$ ./target/debug/hx README.md +11
$ ./target/debug/hx README.md +'g 11'
$ ./target/debug/hx README.md +'goto 11'
$ ./target/debug/hx +'open README.md' +11
$ ./target/debug/hx +'open README.md' +'theme acme'
```

helix-term/src/args.rs
```diff
@@ -17,6 +17,8 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub line_num: u32,
+    pub commands: Vec<String>,
```
the-mikedavis, line 21: Command parsing and executing should be removed or moved to a separate PR: this is something that should be discussed and designed separately. I'm not sure it should be added in general.

helix-term/src/args.rs
```diff
@@ -17,6 +17,8 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub line_num: u32,
```
the-mikedavis, line 20: This field is never accessed. A minor naming nit but it should probably be named `line_number` for clarity

helix-term/src/main.rs
```diff
@@ -73,6 +73,7 @@ FLAGS:
     -V, --version                  Prints version information
     --vsplit                       Splits all given files vertically into different windows
     --hsplit                       Splits all given files horizontally into different windows
+    +N                             Goto line number N, can be mentioned after each file
```
the-mikedavis, line 76: Some programs give the `+N` before the filename like `less`'s `v` command. Kakoune only accepts one of these arguments and it only applies to the first file argument.

pascalkuthe:
Maybe this was already discussed (altough I didn't find anything) but what about using `:open <path>:<line>:<column>` here? You could reuse the `parse_file` function and would stay more consisntent with the CLI and also `gf` if something like #5260 is merged

nabaco:
@pascalkuthe that was mentioned in the original issue: https://github.com/helix-editor/helix/issues/5437.
I'll move the execution part to a separate PR. I agree that it somewhat a hack and needs better design.

helix-term/src/main.rs
nabaco: Sure, we can do that, if you insist. But why? Given that we can give more flexible functionality here?

the-mikedavis: It may simplify the code to only parse one `+` arg especially if it is allowed before the filename. This `+` syntax is only for the sake of compatibility with programs like `less` that use that syntax exclusively so there's no need to make it very flexible: you can already specify line and column with colons and numbers after the file names (for example `hx README.md:15:10`)

nabaco: Done. I looked at how NeoVim behaves and aligned to it.
It accepts as many +N arguments as you wish, but takes the latest one and applies to the first file in the list. Implementation is indeed a bit simpler and more straight forward that way.

helix-term/src/args.rs
```diff
@@ -17,6 +17,7 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub line_number: usize,
```
the-mikedavis, line 20: It looks like this isn't used outside of the `parse_args` function so it could be a local in that function: `let mut line_number = 0;`

helix-term/src/args.rs
```diff
@@ -73,6 +74,16 @@ impl Args {
...
                 }
+                arg if arg.starts_with('+') => {
+                    let arg = arg.get(1..).unwrap();
+                    args.line_number = match arg.parse() {
+                        Ok(n) => n,
+                        _ => anyhow::bail!("bad line number after +"),
+                    };
+                    if args.line_number > 0 {
+                        args.line_number -= 1;
+                    }
```
the-mikedavis, line 85:
```suggestion
                    args.line_number = match arg.parse::<usize>() {
                        Ok(n) => n.saturating_sub(1),
                        _ => anyhow::bail!("bad line number after +"),
                    };
```

helix-term/src/main.rs
```diff
@@ -73,6 +73,7 @@ FLAGS:
     -V, --version                  Prints version information
     --vsplit                       Splits all given files vertically into different windows
     --hsplit                       Splits all given files horizontally into different windows
+    +N                             Goto line number N
```
the-mikedavis, line 76:
```suggestion
    +N                             Open the first given file at line number N
```

kirawi:
Closed by https://github.com/helix-editor/helix/pull/8521

PR #5419, opened 2023-01-05
Change built-in themes to use curly underlines

blt-r:
Fixes #5361.
I didn't touch themes that don't underline their diagnostics messages and themes that use `underline.style = "line"` instead of `modifiers = ["underlined"]`

PR #5605, opened 2023-01-20
Make `fleet-dark` accurate to the official theme

kirawi:
# Objective

The current palette for the theme didn't use the official palette, and some scopes used the wrong colors. The contrast between the cursorline and comments also caused eye strain.

## Solution

I copied the palette from `dark.json` in the themes folder, and reworked the theme to reflect the official one. I'm a bit displeased with the color for `type.enum.variant`, which the official theme doesn't highlight, as well as the contrast between `variables.builtin` and the background (for which I preferred the old color).

Fleet vs. `new fleet-dark` (top) vs. `old fleet-dark` (bottom)

Capture
Capture1

---

## Changelog

updated `fleet-dark`'s palette to be identical to the one in the official theme

### DIFF
--- a/runtime/themes/fleet_dark.toml
+++ b/runtime/themes/fleet_dark.toml
@@ -99,10 +99,10 @@
 "warning" = "#FACb66"
 "error" = "#FF5269"
 
-"diagnostic.hint" = { underline = { color = "Gray 80", style = "solid" } }
-"diagnostic.info" = { underline = { color = "#A366C4", style = "solid" } }
-"diagnostic.warning" = { underline = { color = "#FACB66", style = "solid" } }
-"diagnostic.error" = { underline = { color = "#FF5269", style = "solid" } }
+"diagnostic.hint" = { underline = { color = "Gray 80", style = "line" } }
+"diagnostic.info" = { underline = { color = "#A366C4", style = "line" } }
+"diagnostic.warning" = { underline = { color = "#FACB66", style = "line" } }
+"diagnostic.error" = { underline = { color = "#FF5269", style = "line" } }
 
 [palette]
 "White" = "#ffffff"
--- a/runtime/themes/fleet_dark.toml
+++ b/runtime/themes/fleet_dark.toml
@@ -73,7 +73,7 @@
 
 "ui.cursor" = { modifiers = ["reversed"] } # .insert / .select / .match / .primary
 "ui.cursor.match" = { bg = "Blue 30" } # .insert / .select / .match / .primary
-"ui.selection" = { bg = "Gray 80" } # .primary
+"ui.selection" = { bg = "Gray 50" } # .primary
 "ui.selection.primary" = { bg = "Blue 40" }
 
 "ui.cursorline" = { bg = "Gray 20" }
--- a/runtime/themes/fleet_dark.toml
+++ b/runtime/themes/fleet_dark.toml
@@ -54,7 +54,7 @@
 # "markup.link" = "green"
 "markup.link.url" = { fg = "Pink", modifiers = ['italic', 'underlined'] }
 "markup.link.text" = "Cyan"
-"markup.link.label" = "purple"
+"markup.link.label" = "Purple 20"
 "markup.quote" = "Pink"
 "markup.raw" = "Pink"
 "markup.raw.inline" = "Cyan" # .completion / .hover
--- a/runtime/themes/fleet_dark.toml
+++ b/runtime/themes/fleet_dark.toml
@@ -4,128 +4,161 @@
 # Original author: @krfl
 # Contributors:
 # @matoous
+# @kirawi
 
-"attribute" = "green"
-"type" = "light_blue"
-"type.enum.variant" = "purple"
-"constructor" = "yellow"
-"constant" = "cyan"
+"attribute" = "Lime"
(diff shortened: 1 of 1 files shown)


kirawi:
\cc @krfl

archseer:
Comparison screenshots?

kirawi:
Yep, I've updated the description with them.

krfl:
@kirawi Thank you for taking the time to do all this work. 

I have three comments:
1. The light grey background on the horizontal divider used by the pickers looks a little out of place compared to the rest of the theme.
Horizontal divider

2. The curly underlines were just changed in #5419. Official Fleet uses thick straight lines to underline errors. My understanding was that all themes that uses underlines should use curly underlines and I see that these were removed by your update. If we are going to show underlines, then it's going to be a compromise. It would be great to get a final confirmation on this. @pascalkuthe? 
(old `fleet_dark` on the left, new `fleet_dark` on the right)
Screenshot 2023-01-21 at 09 08 22

3. Please do add yourself to the list of contributors :)

kirawi:
Thanks for the review! I didn't notice the background nor the missing underline for diagnostics. I've also diverged from the official theme by making `ui.cursor.match` yellow because the blue was too low-contrast.

runtime/themes/fleet_dark.toml
```diff
@@ -4,128 +4,161 @@
...
-"warning" = "orange_accent"
-"error" = "red_error"
-"diagnostic" = { modifiers = [] }
-"diagnostic.hint" = { underline = { color = "light", style = "curl" } }
-"diagnostic.info" = { underline = { color = "blue_accent", style = "curl" } }
-"diagnostic.warning" = { underline = { color = "yellow_accent", style = "curl" } }
-"diagnostic.error" = { underline = { color = "red_error", style = "curl" } }
+"ui.background" = { bg = "Gray 10" } # .separator
+"ui.statusline" = { fg = "Gray 120", bg = "Gray 10" } # .inactive / .normal / .insert / .select
+# "ui.statusline.normal" = { fg = "lightest", bg = "darker"}
```
kirawi, line 70: I commented out these since I didn't know what would be good colors.

krfl:
```suggestion
"ui.statusline" = { fg = "Gray 120", bg = "Gray 20" }
"ui.statusline.insert" = { bg = "Blue 80" }
"ui.statusline.select" = { bg = "Yellow 40" }
```

I think it's important to separate `fleet` features from `helix` features and just make an educated guess at what people will prefer. Fleet relies heavily on blue, so I think it will be a safe and familiar color for insert mode. 

Yellow 40 or Green 40 might work too. The yellow is darker and warmer and reminds me of vim, but the green might be more in the style of fleet. What do you think?

krfl (commented):
I would very much to know whether all themes should use curly lines or not, because that's the understanding I got from #5419

Also, I think that `Gray 80` for hints might be too subtle? It's not that it's hard to see or read, but I feel like it should pop a lot more. 

Here i have the old theme on the left and this update on the right. Also a comparison screenshot of fleet below. 
diagnostic

Screenshot 2023-01-23 at 00 36 02

kirawi:
IIRC curly underlines were only applied to all themes that didn't explicitly specify the line type.

pascalkuthe:
I think the curly part is less important as long as the underlines are colored so its possible to tell whether we are looking at a warning or an error.

krfl:
> I checked the source again, and it looks like `Gray 80` is the color Fleet uses. Do you want to diverge from the official theme and use a higher contrast color?
> [dark.json.txt](https://github.com/helix-editor/helix/files/10476907/dark.json.txt)

I agree that `Gray 80` is correct in terms of matching what Fleet uses. I also believe that the statusbar is more important in Helix than in Fleet. The answer depends on whether you view the theme as an adaptation for Helix or as a copy of Fleet Dark.

Since I'm so torn, I'll let you decide. I will be okay either way. Promise :)

kirawi:
Personally, I feel that it should be accurate to the official theme so it's not jarring for users. Users can customize the theme if they prefer :P

kirawi:
Secondary selection highlighting is actually pretty bad, so I'll have to fix that before this could get merged.

kirawi:
Fixed. I misread 50 as 80.

runtime/themes/fleet_dark.toml
```diff
@@ -4,128 +4,161 @@
...
+"markup.heading" = { fg = "Cyan", modifiers = ["bold"] } # .marker / .1 / .2 / .3 / .4 / .5 / .6
+"markup.list" = "Pink" # .unnumbered / .numbered
+"markup.list.numbered" = "Cyan"
+"markup.list.unnumbered" = "Cyan"
 # "markup.link" = "green"
-"markup.link.url" = { fg = "pink", modifiers = ['italic', 'underlined'] }
-"markup.link.text" = "cyan"
+"markup.link.url" = { fg = "Pink", modifiers = ['italic', 'underlined'] }
+"markup.link.text" = "Cyan"
 "markup.link.label" = "purple"
```
the-mikedavis, line 57: looks like this `purple` isn't defined anymore

the-mikedavis (approved):

Issue #401, opened 2021-07-03
Persistent State


kirawi:
This feature will allow you to share the same file history between sessions, and could help in situations where the editor crashes. #290 #294

bestouff:
Now the question is: what file format ? In the ideal case, like for vim, a script which can be edited.

cole-h:
This should probably be implemented as a plugin (whenever that functionality is available).

---

That said, it would probably be prudent to use something easy to serialize. TOML is already used for the basic config, so why not stick with that? An example, totally fictional file might store information about the documents themselves like:

```toml
[[documents]]
path = "/canonical/file/path"
cursor = { line = 90, column = 1 }
alt_document = "/canonical/file/path2" # this is "last_accessed_doc"
# no focused here; would be deserialized as false
[[documents.selections]]
start = { line = 70, column = 10 }
end = { line = 100, column = 1 }
[[documents.selections]]
start = { line = 1, column = 1 }
end = { line = 2, column = 23 }

[[documents]]
path = "/canonical/file/path2"
cursor = { line = 1, column = 1 }
# no alt_document here because it doesn't have one; would be deserialized as None
focused = true # means this document is currently visible
[[documents.selections]]
start = { line = 1, column = 10 }
end = { line = 5, column = 1 }
[[documents.selections]]
start = { line = 10, column = 1 }
end = { line = 10, column = 3 }
```

I don't currently have any ideas on how we would serialize the current layout of views to TOML, though.

eugenesvk:
Sublime Text has this awesome "hot exit" feature that allows you to exit without any "Save unsaved files?" confirmation and reopen to exactly the same state, it's very useful for regular use, not just crashes.
Am I correct that you feature request is exactly the same?

(quote from ST regarding this feature)

> Exiting the application with hot_exit enabled will cause it to close immediately without prompting. Unsaved modifications and open files will be preserved and restored when next starting.
Closing a window with an associated project will also close the window without prompting, preserving unsaved changes in the workspace file alongside the project.

lytedev:
For what it's worth, I'd like to propose that this _not_ be a plugin. I know Helix calls itself a text editor, but it definitely seeks to have "MVP IDE" features, and this is something VS Code and other editors do out of the box to various degrees. Additionally, this sort of persistence is really nice to have for a number of reasons, especially if you regularly switch between projects wanting to sort of pick up where you left off. Happy to argue and be wrong, but I think this should be considered core functionality!

[...]

PR #5608, opened 2023-01-20
Persistent Undo

kirawi:
# Objective

Implement persistent undo as part of #401

## Solution

- Implemented parser combinators in `helix-core::parse`.
- Implemented a serialized file format for `History`. At the moment, `timestamp`s aren't preserved.
    - New revisions are appended, and the tree is reconstructed during deserialization.
- Added new quickcheck tests for serialization/deserialization/merging histories, and an integration test.
- `:reload` will also merge the saved history with the current one.
- Enabling `persistent-undo` at runtime will cause the histories of open docs to be merged as well.
---

## Changelog

implemented persistent undo


kirawi:
So far, I've implemented serializing/deserializing the undo history. I'll check it over for bugs, and then work on implementing the manual commands.

kirawi:
Commands have been implemented. I need to fix an issue with detecting outdated file histories (which could cause a panic), and then add tests.

pascalkuthe:
> @pascalkuthe suggested appending the absolute paths of the files to the session directory, but Rust doesn't allow this: https://github.com/rust-lang/rust/issues/16507. Implementing a function with that behavior seemed more work than an index.

That's not quite what I meant. What I meant instead is that there is some central location (notably not a session but rather somewhere central. For example nvim default to `$XDG_STATE_HOME/nvim/undo/`. The location doesn't matter much and could be configurable (nvim also allows this by setting the `undodir` option).

Ofcourse you can not joint the absolute  paths on top of a directory. However you can escape a path.
For example the undo file for `/home/pascalkuthe/Projects/helix/README.md` would be ~/$XDG_STATE_HOME/helix/undo/'%home%pascalkuthe%Projects%helix%README.md'` so the entire filepath just becomes a filename. Rust is actually very well suited for implementing something like this. You can reuse the `Path::components` iterator for that which takes care of the hard part of splitting the file into componenents for you.

I am also of the opinion that the undofile should be (if enabled) automatically written on save and automatically opened when a file is opened instead of manually opened from a session.

In general undofiles are seperate from sessions to me. Just like in nvim undofiles should be seperate because they are decidedly not session specific. Each file has one history of changes no matter what session performs these changes (which is why I think it's better to autoamtically save and load undofiles from a centralized location instead of manual load from a specific session). With such a global approach the indexfile becomes impractical. There would be laods of conflict and dataloss for that index file between different helix instances whereas for standalone escaped undofiles only have contention when two helix instances write to the same file (in which case there is already contention anyway which is a seperate problem to solve with file watching/mtime safeguard)

kirawi:
Yeah, that seems like an overall improvement and not too big of a change. One thing I liked about workspaces in VSCode (but forgot to implement here) is that it would open the last open buffers. Maybe something like that could be done by looking for undofiles beginning with the cwd path, and having `.undo.last` extension?

antoyo:
I'm also for having undo-files separate from sessions since I plan to use sessions, but not undo-files.

pascalkuthe:
> Yeah, that seems like an overall improvement and not too big of a change. One thing I liked about workspaces in VSCode (but forgot to implement here) is that it would open the last open buffers. Maybe something like that could be done by looking for undofiles beginning with the cwd path, and having `.undo.last` extension?

I think that is a session related feature. Like I said I think sessions and undofiles are pretty separate features. So the undofiles always get opened and saved when the files do (if the setting is enabled). However open buffers only get restored when a session is opened (would just be a list of paths stored in the session file)

kirawi:
> I am also of the opinion that the undofile should be (if enabled) automatically written on save and automatically opened when a file is opened instead of manually opened from a session.

I've implemented this behavior in this PR, and I'm waiting on CI to check the tests.

pascalkuthe:
> > I am also of the opinion that the undofile should be (if enabled) automatically written on save and automatically opened when a file is opened instead of manually opened from a session.
> 
> I've implemented this behavior in this PR, and I'm waiting on CI to check the tests.

Awesome! I am glad you are making progress, I think you are starting to converge on a really solid solution and it took much less changes (in terms of LOC) than I expected :+1:  I don't have time for an in-depth review right now. I will try to take a closer look when I have more time. Altough all the things I saw while scanning where more nitpicks than major concerns.

Is this at a point now where it would make sense to use this as a daily driver to shake out more bugs or is it still too early for testing?

kirawi:
Yeah, I think it's ready for daily-driving. I was also trying to get `cache_dir` to use a `HELIX_CACHE` env variable set in the integration test so it wouldn't pollute it, but `std::env::var("HELIX_CACHE")` returns `NotPresent` in `helix_loader`.

kirawi:
`:wq` won't serialize the history because jobs are executed before the writes are. It didn't fix the issue to move the serialization to `save_impl` either because the file's mtime and hash may differ from what's serialized, presumably because the write finishes after the serialization.

Philipp-M:
Works great so far.

Out of interest: Why are you using your own serialization implementation instead of something like serde?

helix-core/src/flock.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.lock()?;
+        Ok(&self.file)
+    }
+
+    pub fn get_mut(&mut self) -> Result<&mut File> {
+        self.lock()?;
+        Ok(&mut self.file)
+    }
+
+    fn lock(&self) -> Result<()> {
```
Philipp-M, line 34: Does it make sense to make this async, or will this unnecessarily complicate things?

pascalkuthe:
> Works great so far.
> 
> Out of interest: Why are you using your own serialization implementation instead of something like serde?

We eventually want append-mode serialization here. Meaning that when saving a file imstead of rewriting the entirw undo hiatory to disk we only appemd new entires. This is not possible with serde.

Right now this is not implemented but to future proof the design and avoid having to start from scratch in the future we directly went with a custom binary format.

This also has the advantage of avoiding yet another dependency and makes it easier to optimize the on-disk format (having an on-disk format that is meaningfully diffeeent from the in-memory format is possible in serde but quite painful)

helix-core/src/flock.rs
pascalkuthe: I don't actually think we need a file lock at all. 

@kirawi since #5805 has landed helix already checks that there is no external contention so if we are saving the file it is also save to write the undofile. I don't think we need a lock anymore. I think the main reason you had a file lock to begin with was to protect the shared index file if  am not mistaken but since that has been removed, I think the flock can be removed too (which would make this PR smaller)

pascalkuthe:
> Works great so far.
> 
> Out of interest: Why are you using your own serialization implementation instead of something like serde?

Hmm file race considitiosn like that are tricky I need to think about that when I have more time

kirawi:
I actually made a mistake 😅 
In the current code, the reason `:wq` won't serialize is that writes are handled through https://github.com/helix-editor/helix/blob/a4751db8c7f308494e574429fc862a8deec96a7b/helix-view/src/editor.rs#L1605-L1624 instead of `Application::handle_document_write`.

helix-core/src/flock.rs
kirawi: I was also thinking about in case multiple Helix instances would write to the same undo file at the same time. Theoretically, isn't it possible for multiple instances to pass the mtime check? Although, I can't imagine anyone writing to the same file from multiple instances in the first place.

pascalkuthe:
I am a bit confused why you can't handle this in `save_impl` AFAIK all write essentially call that function as the buffer is only encoded there

kirawi:
It used to be, but `DocumentSavedEvent` wouldn't be sent back if serialization failed.

pascalkuthe:
> It used to be, but I moved it since `DocumentSavedEvent` wouldn't be sent back if serialization failed.

That is the right location to write the file, you should just either:
* ignore failures to write undofile
* add a `undofile_failed` property to `DocumentSavedEvent` to indicate that writing the undofile failed

jlebon:
Started taking this for a spin to help test it out. One issue I noticed is that I had to create the `undo` directory manually. Likely missing a `std::fs::create_dir_all()` somewhere. Otherwise, so far so good!

helix-term/src/commands/typed.rs
```diff
@@ -1176,7 +1176,12 @@ fn reload(
     doc.reload(view, &cx.editor.diff_providers, redraw_handle)
         .map(|_| {
             view.ensure_cursor_in_view(doc, scrolloff);
-        })
+        })?;
+    if let Err(e) = doc.load_history() {
```
kirawi, line 1180: On second thought, this would delete any changes not written.

Philipp-M: Yeah this shouldn't be there IMO, the history is modified by using a reload transaction, I think that should be respected in the (serialized) undo history.

kirawi: Removed

helix-core/Cargo.toml
```diff
@@ -47,5 +49,9 @@ chrono = { version = "0.4", default-features = false, features = ["alloc", "std"
 etcetera = "0.4"
 textwrap = "0.16.0"
 
+[target.'cfg(unix)'.dependencies]
```
kirawi, line 52: Self-note: Should've been removed w/ flock

helix-term/src/commands/typed.rs
pascalkuthe: I don't think this is a simple blaxk and white either load the history or don't. 

I think the we should always load the undofile into memory. Then any new revisions (simply check any revisions are newer than the last saved revision) would be merged into memory. 

The next step is to check if the undofile revision hash and the hash of the loaded file match. If that is the case we should just switch revisions to the head revision of the undofile. All edits happened in helix so no new transaction is needed.

If the two hashes don't match we compute the diff as we currently do and add a new transaction. 

This way no part of history (either in memory or in the file) ever gets lost, if all edits happen entirely in helix we just keep the history and switch the head revision (so doing edits like A-> B in instance 1 then B -> C in instance two and C -> D in instance one would form an history like A -> B -> C -> D)

pascalkuthe: BTW. when overwriting a file with contention (meaning we write with `w!` because the file changed on disk we may also want to reload the history file and do a history merge as I described (but not update the revision ofcourse) so we don't accidentlly loose any history

kirawi:
I'm thinking about whether to give a go at implementing append in this PR. I think all it would take is to modify the `len` of the number of transactions, and then append to the end of the file.

pascalkuthe:
The process I described above (https://github.com/helix-editor/helix/pull/5608#discussion_r1104016356) might also.be relevant for that.

I am not 100% sure if aopendmode is that easy. Yes new revisions can be appended but how do you maintain the tree structure? When append a new revision you need to add it to the parent as a new child that would require copying everything afterwards backwards. You could consider splitting the undofile format in two tough: One file that stores the revision data (which would be append only) and one file that stores the tree structure (which would not be an apendmode file).

pascalkuthe:
Technically the tree structure could also be aopendmode noe that I think about it. Instead of storing a list of child revisions for each revision a list of vertices (so for each child simply store (parent_id, child_I'd).

Then you could have two aopendmode files that would be pretty neat

pascalkuthe:
Actually you don't even need that. If you store the parent of each revision instead of the children you can reconstruct the tree in memory. 

That would actually work great. Just don't store the children in the undofile and only the parent and whenever you load a new revision add it to the list of children of the parent. Then you only.need a single file.and can do append mode

kirawi:
Yeah that seems sensible to me 👍 

I think another consideration would be handling the process quitting mid-append so the already saved history isn't lost.

helix-term/src/commands/typed.rs
kirawi: > The next step is to check if the undofile revision hash and the hash of the loaded file match.
Could you clarify what you mean by that?

pascalkuthe: ah sorry forgot about that. Well the idea is the following:

If we are reloading and some random program that is not helix changed the file somehow then the right way to treat that is just a normal entry in the undo history (just like we do now).

However if the edits happened inside a second helix instance then we actually have a history of those edits and there is already a revision for them.
All we need to do is update the head revision to point to the head revision of the undofile. We don't need to generate a new transaction at all.

For example lets consider the following history:

We have an undofile:
```
A -> B (B is heaD)
```
And made some additional edits to that which we haven't saved yet:

````
A -> B -> C -> D (D is head)
````

Then we open that file in a second helix instance. This instance doesn't know about our new changes yet because we haven't saved.
We do some different edits in that second instance. It's history now looks like this.

```
A -> B -> E -> F (F is head)
```

Now we save in the second history so that history is written to the undofile.
Now if we reload in the first instance the following should happen:

* first we merge the histories

```
A -> B -> C -> D (D is head)
         \ 
           \ -> E -> F
```

* then we check if the hash stored in the undofile matches the file we just loaded (so all edits happened inside helix)
* in this case that is true (since we only edited in a second helix instance). That means we don't need to create a new revision for this edit because there is already a history for it. Simply move HEAD to `F`.
```
A -> B -> C -> D
         \ 
           \ -> E -> F (F is head)
```

That means that cooperation between two helix instances works seamlessly and two helix instances can share undo history. 

We can ofcourse only do this if the file at `F` is the same as the file on disk.
This would not be the case if another program (like `git reset --hard` ) changed the file between the save from helix and `:reload`. By comparing the SHA1 hashes we can avoid this problem.

pascalkuthe: The merging would work as follows:

The Revision `B` appears both in the undofile and the in-memory version.
However there are new children in the undofile (old children can never be removed) . In this case `E`. Simply add these to the list of children of `B`.

pascalkuthe: Or in other words: The merging would work by computing the union of child sets (if you want to get academic about it)

kirawi:
I've implemented reloading the history on `:reload` and appending the history.

SollyCB:
Idk if I am missing something, I will assume no reply means I am not in the know. And I am not sure what was decided on loading the undo file into memory. But:

I understand undos to work as operations pushed to a Vec<Operation>. 
If when a change is made, it is pushed to the list of Operations, can't this list be written to disk? Along with a hash of the file? When helix opens a file, the hashes are compared with the current file, if hashes do not match (file was edited by another program) the undo list is invalidated and is emptied. Oldest changes are then the changes made from this point on. If hashes do match and the files are the same then the undo list is loaded as the current Vec<Operation>.  

Regarding the same file being edited by different instances of helix, can't the timestamp of the most recent operation be saved also, and whichever list has the most recent timestamp be the one that is maintained? So if instance A of helix writes the file that is open and therefore writes to the undo list, when instance B goes to save it can see that it is not at the most recent version of the file and can open the most recent version as a swapfile to compare changes (which also loads the newest version of the undo list), or overwrites its version of the file with the most recent one, which also overwrites its version of the undo list that it has loaded into memory.

I think I am largely reiterating PascalKuthe's comments. But I feel the tree is unnecessary: simply load the undo list into memory upon opening a file, overwrite the list on save. 

Again apologies if this is a timewasting comment, I am a noob desperate to learn.

kirawi:
See https://github.com/helix-editor/helix/pull/5608#discussion_r1104015482

kirawi:
The test for reloading don't seem sufficient, since I ran into a case where it panics.

kirawi:
I think I traced the cause of the bug to how the history is saved in `Document::save_impl`. If another Helix instance writes, then it attempts to overwrite the entire file with that doc's history, but will pass the same range of revisions as if it was appending.

kirawi:
Will be superseded by a new PR.

fastfading:
is this issue fixed ? what is the conclusion ?  
this is the basic funtion of an edit. 
without it.  I wil still use neovim although I like the design of helix

johannesrave:
@kirawi which PR is the new one to follow for this? thanks!

cole-h:
I'd venture to guess: https://github.com/helix-editor/helix/pull/9154

PR #5609, opened 2023-01-20
tutor: Change "Type" to "Press" for specials keys and modifiers

vitorgalvao:
“Type Escape” makes it seem one has press <kbd>e</kbd><kbd>s</kbd><kbd>c</kbd><kbd>a</kbd><kbd>p</kbd><kbd>e</kbd> instead of pressing the key. This may be even better as “Press the esc key” but it’s wordier.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -48,13 +48,13 @@ _________________________________________________________________
 
  1. Type : to enter Command mode. Your cursor will
     move to the bottom of the screen.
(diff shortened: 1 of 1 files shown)


the-mikedavis:
This should also be applied for special keys and modifiers: `Return`, `Alt-<key>`, `Ctrl-<key>`

vitorgalvao:
Done. All changes performed with Helix.

the-mikedavis (approved):

Issue #3596, opened 2022-08-29
Formatter doesn't have access to filename


IvanGoncharov:
#2942 added `formatter` option, which is very useful.
I used it prettier but quickly discovered that I have conflicting formatting between CI and my editor.

This happens because prettier change formatting based on filename, e.g. `package.json` formatted with  `json-stringify` but other `*.json` files formatted with `json` parser.
https://github.com/prettier/prettier/blob/2316e2fecd171335fc00f3ec97b7e46cc8964153/src/language-js/index.js#L81-L94

To solve this use case, prettier provides [--stdin-filepath](https://prettier.io/docs/en/cli.html#--stdin-filepath), but I can't use it in helix since there is no way to pass the filename into `fomatter` option.

Having a placeholder of some kind would be great, e.g. `$filename`.

kirawi:
Previously #2942 did implement passing a file path to the formatter, but it was removed since stdin was deemed sufficient. Maybe this use case will change things.

kvrohit:
I have a similar issue too which would be solved if there was a way to pass in the filename to the formatter command.

JamesMDunn:
I also have this issue with prettier. Not sure how to get it to format correctly

Anthuang:
You can manually specify the parser for prettier to use to avoid needing the file name, see https://prettier.io/docs/en/options.html#parser.

E.g.:

```toml
[[language]]
name = "typescript"
auto-format = true
formatter = { command = "prettier", args = ["--parser", "typescript"] }
```

ttys3:
@Anthuang  this is hard to use, because the cli programmer know better than you.
for prettier, it has `flow|babel|babel-flow|babel-ts|typescript|acorn|espree|meriyah|css|less|scss|json|json5|json-stringify|graphql|markdown|mdx|vue|yaml|glimmer|html|angular|lwc` parsers

we can not count on a hard coded config to manual calculate the parser to use. this is bad.

ttys3:
I got this problem too.

some formatters need the file path to location a config in the project (`stylua`) or use it to determine the file type (`prettier`), 

in neovim we can use ` vim.api.nvim_buf_get_name(0)` to get the current buf file path, but in hexlix seems no way to do so.

for example:

`style`:

>  --stdin-filepath <stdin-filepath>    Specify the location of the file that is being passed into stdin. Ignored
                                             if not taking in input from stdin. This option is only used to help
                                             determine where to find the configuration file

```
require("formatter").setup {
[...]

PR #5626, opened 2023-01-21
feat(format): current file in formatter args

filipdutescu:
Add support for a current file placeholder in formatter arguments. This should make it possible to run custom formatters, such as `rome` (Typescript), `rustfmt` (Rust) and more.

The changes take inspiration from f7ecf9c and acdce30, but instead of adding a new field to the formatter struct, it works by looking for a predefined placeholder (`{}`), which should be replaced, if found. This should allow for better control over arguments of formatters, allowing both giving it as a standalone arg (`"{}"`) and as part of one (ie `"--src-file={}"`). The placeholder, `{}`, is used since it is common, Rust using it frequently.

Closes: #3596

### DIFF
--- a/book/src/languages.md
+++ b/book/src/languages.md
@@ -96,7 +96,7 @@ The `language-server` field takes the following keys:
 | Key           | Description                                                           |
 | ---           | -----------                                                           |
 | `command`     | The name of the language server binary to execute. Binaries must be in `$PATH` |
-| `args`        | A list of arguments to pass to the language server binary             |
+| `args`        | A list of arguments to pass to the language server binary. Those can have `"{}"` as a placeholder for the current file path. |
 | `timeout`     | The maximum time a request to the language server may take, in seconds. Defaults to `20` |
 | `language-id` | The language name to pass to the language server. Some language servers support multiple languages and use this field to determine which one is being served in a buffer |
 | `environment` | Any environment variables that will be used when starting the language server `{ "KEY1" = "Value1", "KEY2" = "Value2" }` |
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -286,21 +286,23 @@ fn write_impl(
 ) -> anyhow::Result<()> {
     let editor_auto_fmt = cx.editor.config().auto_format;
     let jobs = &mut cx.jobs;
+    let redraw_handle = cx.editor.redraw_handle.clone();
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
 
     let fmt = if editor_auto_fmt {
-        doc.auto_format().map(|fmt| {
-            let callback = make_format_callback(
-                doc.id(),
-                doc.version(),
-                view.id,
-                fmt,
-                Some((path.map(Into::into), force)),
-            );
+        doc.auto_format(view, &cx.editor.diff_providers, redraw_handle)
+            .map(|fmt| {
+                let callback = make_format_callback(
+                    doc.id(),
+                    doc.version(),
+                    view.id,
+                    fmt,
+                    Some((path.map(Into::into), force)),
+                );
 
-            jobs.add(Job::with_callback(callback).wait_before_exiting());
-        })
+                jobs.add(Job::with_callback(callback).wait_before_exiting());
+            })
     } else {
         None
     };
@@ -361,7 +363,9 @@ fn format(
     }
 
     let (view, doc) = current!(cx.editor);
-    if let Some(format) = doc.format() {
+    let redraw_handle = cx.editor.redraw_handle.clone();
+    if let Some(format) = doc.format(view, &cx.editor.diff_providers, redraw_handle) {
+        let view = view!(cx.editor);
         let callback = make_format_callback(doc.id(), doc.version(), view.id, format, None);
         cx.jobs.callback(callback);
     }
@@ -584,7 +588,7 @@ pub fn write_all_impl(
     let mut errors: Vec<&'static str> = Vec::new();
     let auto_format = cx.editor.config().auto_format;
     let jobs = &mut cx.jobs;
-    let current_view = view!(cx.editor);
+    let current_view = view_mut!(cx.editor);
 
     // save all documents
     let saves: Vec<_> = cx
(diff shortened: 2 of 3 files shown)


the-mikedavis:
The syntax we use for this should mirror the variable expansion syntax https://github.com/helix-editor/helix/pull/3393

filipdutescu:
Should I add it now or wait for that one first?

filipdutescu:
Just to make sure, that syntax is `val{filename}`, correct?

the-mikedavis:
Currently yep that's the syntax but I think we should wait on that PR incase it's revised

nestor-custodio:
The original issue this PR addresses is over 8 months old now and is a show-stopping autoformatting problem (for certain users under specific tooling conditions/needs) on a number of languages. This PR fixed the problem over 3 months ago and has been sitting idle waiting on an unrelated PR, just to make sure they use the same variable expansion syntax.

Am I crazy for thinking this should go out as-is and #3393 should get a note indicating that the relevant code here should be updated to match **if** the syntax does see a change before that PR gets final approval?

(This is not intended to sound as snippy as it *probably* does. Just wanted to summarize the background behind this before asking the question.)

the-mikedavis:
We'd like to get #3393 into the next release (the one after the imminent small bugfix-like release) and I think this PR could follow #3393 shortly. I want the variable syntax to be finalized before introducing any variables so that we don't have to backtrack and make breaking changes.

> This is not intended to sound as snippy as it _probably_ does

The first paragraph _does_ basically come across as a rant :/. If it's a show-stopper for you then you should merge the changes into a fork and use that.

n3oney:
I've just tried applying the patch from this PR to the master branch and it seems to not do anything?

image
Just seems to put a literal `{}` as $1, which makes my file end up being `Error: No parser could be inferred for file: /path/to/{}`

filipdutescu:
Once there is an an agreed upon syntax I am happy to update the PR and fix the issue, just to give an update. But since there is currently nothing set in stone I see no use updating the PR just to further change it later on.

n3oney:
> Once there is an an agreed upon syntax I am happy to update the PR and fix the issue, just to give an update. But since there is currently nothing set in stone I see no use updating the PR just to further change it later on.

Yeah, I understand. I just wanted something quick n dirty working right now, because I'm honestly only using this as a hack to get WakaTime working with Helix. See https://github.com/helix-editor/helix/discussions/3477#discussioncomment-6440699

IgnisDa:
> if anyone wants a simple patch file that does the same thing (I believe) and applies on the master version currently I wrote a simple one:
> 
> ```
> diff --git a/helix-view/src/document.rs b/helix-view/src/document.rs
> index b08370f9..bcfdafa1 100644
> --- a/helix-view/src/document.rs
> +++ b/helix-view/src/document.rs
> @@ -735,11 +735,23 @@ pub fn format(&self) -> Option<BoxFuture<'static, Result<Transaction, FormatterE
>              .and_then(|c| c.formatter.clone())
>              .filter(|formatter| which::which(&formatter.command).is_ok())
>          {
> +            let file_path = self.path()?.to_str().unwrap_or("");
> +
> +            let mut args = formatter.args.clone();
> +
> +            for i in 0..args.len() {
> +                if let Some(arg) = args.get_mut(i) {
> +                    if arg.contains("{}") {
> +                        *arg = arg.replace("{}", file_path);
> +                    }
> +                }
> +            }
> +
>              use std::process::Stdio;
>              let text = self.text().clone();
>              let mut process = tokio::process::Command::new(&formatter.command);
>              process
> -                .args(&formatter.args)
> +                .args(&args)
>                  .stdin(Stdio::piped())
>                  .stdout(Stdio::piped())
>                  .stderr(Stdio::piped());
> ```

Could you tell me how to apply these changes to the current main?

tynanbe:
I'm using this as a workaround.

```toml
formatter = { command = "sh", args = ["-c", """
  file=$(
    ps -p "${PPID}" -o "command=" \
      | awk '{ print $NF }'
  )
  exec prettier \
    --stdin-filepath="${file}" \
    --config-precedence=file-override
"""] }
```

filipdutescu:
Basically yeah. I hope to get to it in the next few weeks.

bluthej:
Hey, so I just wanted to provide some personal feedback on this one. I had a look at this patch, and I eventually ended up writing [my own](https://github.com/helix-editor/helix/compare/master...bluthej:helix:feat/substitute-filename-in-format-command) because I couldn't quite get this one to work.

The main thing that I realized is that the present implementation assumes that when the placeholder for the filename is provided we want to read/write the file from/to disk by accessing the provided path instead of reading from stdin and writing to stdout. I'm wondering if it might not be better to still use stdin and stdout and just replace the filename placeholder. Some of the formatters listed by people in #3596 do work that way, i.e. you pipe the file contents to them and then use the appropriate command line option to pass the filename:

- [prettier](https://prettier.io/docs/options#file-path) and [stylua](https://github.com/JohnnyMorganz/StyLua?tab=readme-ov-file#filtering-when-using-stdin): `--stdin-filepath`
- [ruff](https://docs.astral.sh/ruff/formatter/), [black](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#stdin-filename) and [eslint](https://eslint.org/docs/latest/use/command-line-interface#--stdin-filename): `--stdin-filename`
- [clang-format](https://clang.llvm.org/docs/ClangFormatStyleOptions.html): `--assume-filename`

Granted other formatters don't work with stdin/stdout at all, but I think adding support for the `--stdin-filename` kind of argument is the most straightforward thing to do. My main question is that I'm not sure we can really assume that as soon as there is a filename placeholder in a command then we should read/write from/to disk instead of using stdin/stdout like it's done in the present PR. I'm thinking this might actually need an explicit formatter setting to choose between stdin/stdout or the file on disk.

bluthej:
@filipdutescu I don't want to be stepping on your toes, but if your OK with it I'd like to send my own PR. Just let me know if you'd rather finish this yourself!

filipdutescu:
@bluthej Sure thing! Go ahead and submit your own! Do link it here please

bluthej:
Here's the PR I just opened: #13429

nestor-custodio:
(Heads-up: this should be closed, as #3596 was closed back in April and that change was released as part of 25.07.)

the-mikedavis:
Ah yep, superseded by https://github.com/helix-editor/helix/pull/13429

PR #5628, opened 2023-01-21
fix(dap): validate key and index exist when requesting vars

### LINKED ISSUE #5625: DAP get variables request can fail if no breakpoint is hit before it being requested
### Summary

Due to unchecked access to a `HashMap`, the `dap_variables` method (found in `helix-term/src/commands/dap.rs`) can generate a panic not handled, crashing the program.

### Reproduction Steps

 

I tried this:

1. Open helix and attach to a running process (ie Helix opened with a file)
2. Request variables through `<space>gv`, before a breakpoint is hit (I tried it in the `format` method of `Document`)

I expected this to happen:

Either show nothing or a message mentioning that no breakpoint has been hit.

Instead, this happened:

Panic and crash due to unchecked bounds.

### Helix log

N/A here

### Platform

Linux

### Terminal Emulator

alacritty 0.11.0 (8dbaa0bb)

### Helix Version

helix 22.12

filipdutescu:
Check if the stack frames contain the thread id and the frame before trying to get the frame id. If case any of the two fails to be found, provide the user with messages to inform them of the issue and gracefully return.

Closes: #5625

### DIFF
--- a/helix-term/src/commands/dap.rs
+++ b/helix-term/src/commands/dap.rs
@@ -475,19 +475,36 @@ pub fn dap_variables(cx: &mut Context) {
 
     if debugger.thread_id.is_none() {
         cx.editor
-            .set_status("Cannot access variables while target is running");
+            .set_status("Cannot access variables while target is running.");
         return;
     }
     let (frame, thread_id) = match (debugger.active_frame, debugger.thread_id) {
         (Some(frame), Some(thread_id)) => (frame, thread_id),
         _ => {
             cx.editor
-                .set_status("Cannot find current stack frame to access variables");
+                .set_status("Cannot find current stack frame to access variables.");
             return;
         }
     };
 
-    let frame_id = debugger.stack_frames[&thread_id][frame].id;
+    let thread_frame = match debugger.stack_frames.get(&thread_id) {
+        Some(thread_frame) => thread_frame,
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/dap.rs
```diff
@@ -475,19 +475,32 @@ pub fn dap_variables(cx: &mut Context) {
...
+        cx.editor
+            .set_error("Failed to get stack frame for thread: {thread_id}");
+        return;
+    }
+    let stack_frame = thread_frame.unwrap().get(frame);
+    if stack_frame.is_none() {
+        cx.editor
+            .set_error("Failed to get stack frame for thread {thread_id} and frame {frame}.");
+        return;
+    }
```
archseer, line 501: Avoid unwraps, follow the same pattern as above:

```rust
let thread_frame = debugger.stack_frames.get(&thread_id) {
  Some(thread_frame) => thread_frame,
  None => {
     cx.editor.set_error(...)    
     return;
   }
};
```

filipdutescu: And this actually returns, breaking further execution, right? That's actually really cool, didn't know you could write it as such! Thanks a lot, will implement promptly!

archseer (approved):

PR #5629, opened 2023-01-22
queries/nix: add injections for builtins and writers

pacien:
- queries/nix: document existing injections subjects
- queries/nix: add injection for writeCBin nixpkgs trivial builder
- queries/nix: add injections for nix builtins
- queries/nix: add injections for nixpkgs writers

### DIFF
--- a/runtime/queries/nix/injections.scm
+++ b/runtime/queries/nix/injections.scm
@@ -69,7 +69,7 @@
 ((apply_expression
    function: (apply_expression function: (_) @_func)
    argument: (indented_string_expression (string_fragment) @injection.content))
- (#match? @_func "(^|\\.)writeCBin$")
+ (#match? @_func "(^|\\.)writeC(Bin)?$")
  (#set! injection.language "c")
  (#set! injection.combined))
 
--- a/runtime/queries/nix/injections.scm
+++ b/runtime/queries/nix/injections.scm
@@ -46,3 +46,11 @@
   (#match? @_path "^text$")
   (#set! injection.language "bash")
   (#set! injection.combined))
+
+; trivial-builders.nix pkgs.writeCBin name content
+((apply_expression
+   function: (apply_expression function: (_) @_func)
+   argument: (indented_string_expression (string_fragment) @injection.content))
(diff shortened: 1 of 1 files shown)


pacien:
CC past file committers: @nrdxp, @the-mikedavis, @farwyler

runtime/queries/nix/injections.scm
```diff
@@ -40,3 +64,89 @@
   (#match? @_path "^text$")
   (#set! injection.language "bash")
   (#set! injection.combined))
+
+; trivial-builders.nix pkgs.writeCBin name content
+((apply_expression
+   function: (apply_expression function: (_) @_func)
+   argument: (indented_string_expression (string_fragment) @injection.content))
+ (#match? @_func "(^|\\.)writeCBin$")
```
farwyler, line 72:
```suggestion
 (#match? @_func "(^|\\.)writeC(Bin)$")
```

farwyler (changes requested):
i've tried these today and they work as expected. very helpful use of injections!
🙏🏻 thanks @pacien

runtime/queries/nix/injections.scm
pacien: Those are equivalent though?

nrdxp: As written they should be, perhaps you meant `(Bin)?`?

pacien: There's no `writeC`, only `writeCBin` in nixpkgs/.../trivial-builders.nix.
It would only makes sense for scripts, not much for compiled C.

farwyler: @pacien you are right, `writeC` has been removed [here](https://github.com/NixOS/nixpkgs/commit/982f46985e37a6488d8e904b46e0cba2060adc71) a while ago. The query works for both. I wouldn't mind keeping it anyway for backwards compat? 🤷🏻  

@nrdxp thanks, thats right.

pacien: Okay, added!

the-mikedavis (approved):

PR #5634, opened 2023-01-22
Use markdown language for hub pull-request files

EricCrosson:
Use markdown language for hub pull-request files
    
The hub[^1] command-line tool uses a file called `PULLREQ_EDITMSG`[^2].
This file is used to edit the text from of each commit being submitted
in a pull request, and the final content is rendered as markdown by
GitHub.
    
This commit adds `PULLREQ_EDITMSG` to the list of markdown file-types.
    
[^1]: https://github.com/github/hub
[^2]: https://github.com/github/hub/blob/c8e68d548a39ec0fab6f674a669c21b54d4eec61/commands/pull_request.go#L225

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1019,7 +1019,7 @@ source = { git = "https://github.com/Flakebi/tree-sitter-tablegen", rev = "568dd
 name = "markdown"
 scope = "source.md"
 injection-regex = "md|markdown"
-file-types = ["md", "markdown"]
+file-types = ["md", "markdown", "PULLREQ_EDITMSG"]
 roots = [".marksman.toml"]
 language-server = { command = "marksman", args=["server"] }
 indent = { tab-width = 2, unit = "  " }


pascalkuthe:
I admitately haven't used hub much but wouldn't markdown be the appropriate file type? Given that github PRs are written in markdown (the appropriate hub documentation eben refers to markdown)

EricCrosson:
I like it! Sure, I'll make the change

EricCrosson:
I like it! We won't enjoy the default rulers that come with git-commit, but I can live without that. I'll make the change, thanks for the suggestion

the-mikedavis:
I think that rulers might not be necessary for PR descriptions anyways, no? I
haven't used hub for writing PR descriptions but the website textboxes do not
accomodate hard wrapping. (For example this comment is hardwrapped at 80 cols.)

Maybe hub fixes this though? It would be possible to add a language entry for
PR descriptions that inherits the markdown parser and queries and sets rulers
the same way that the commit message language configuration does.

EricCrosson:
> I think that rulers might not be necessary for PR descriptions anyways, no?

This is true for the body, but the title has a maximum character count else it will be truncated with an ellipsis and continued in the body, which makes for a poor UX.

Thanks for the suggestion on creating a new language entry! In my cursory search, I actually can't find any docs from GitHub detailing _what_ the maximum PR title length is. I think this is a minor case, I'm more than happy to live without the ruler for now. 

I may feel differently if the ruler extended only over the first line of the buffer, but to have a single ruler extend all the way down the screen feels less than intuitive.

the-mikedavis (approved):

PR #5635, opened 2023-01-22
Keymap presentation refinements and underlying restructuring

gibbz00:
Implemented the feedback I got from #5565.

This PR only concerns the keymap. It cleans up most of the keymap code in order to make it more readable and maintainable. (Some bugs were discovered and fixed from this alone.) The refactor also made it possible for clean implementations of a myriad of keymap feature requests.

#### Sorting

* Rewrote the predefined infobox order such that no sort is required for each time infobox is opened.
(`KeytrieNode`s are stored in a `Vec` rather than a `HashMap`.)

* Keymap infobox can optionally be sorted with a `sorted-infobox` editor config option.
Sorting is done by modifier, then by `KeyCode` category, then by each `KeyEvent`.
Single character alphas are placed before the rest (ex. "space") in `KeyCode::Char` category.

* Key events per command are sorted by length, unless it's a "C-" key event. (Should not be a regression of #952)

Standard config
| Before | After | After with `sorted-infobox = true` |
|--- |--- |---
| image |2023-02-14-173700_screenshot| 2023-02-14-173731_screenshot |

Personal
| Before | After | After with `sorted-infobox = true` |
|--- |--- |---
|2023-02-14-173910_screenshot|2023-02-14-180244_screenshot|2023-02-14-174003_screenshot|

### Custom labels

Extends the the work of https://github.com/helix-editor/helix/pull/5203, https://github.com/helix-editor/helix/pull/3958, and https://github.com/helix-editor/helix/pull/4164.

```
[keys.normal.space.c]
# Infobox title
description = "Sort menu"
sticky = true
# Description of command sequence (multiple commands)
C-s = { description = "Sort selection", exec = ["split_selection_on_newline", ":sort", "collapse_selection", "keep_primary_selection"]}
# When no description is supplied
A-s = ["split_selection_on_newline", ":sort", "collapse_selection", "keep_primary_selection"]
```
2023-02-14-175143_screenshot

One of the limitations to the custom descriptions is that the key-value pair must be defined first thing in each TOML map.

### Command palette:

* Uses the ui table widget

* Does not show all commands bound to `no_op` in config.

* Removed a lot of character clutter for showing key-events bound to each command.

| Before | After |
|--- |--- |
|image| 2023-01-22-174716_screenshot

## Future work

It would probably be wise to visit the merged PRs to check which issues can now be closed.

### Custom user commands appear in command palette

Typeable command binding such as `C-k = ":open ~/.config/helix/config.toml"` currently loses all its context
in the command palette. There it's currently shown as:
`open | C-k | Open file"  in the command palette`
Which I believe could be improved to:
`Custom | C-k | :open ~/.config/helix/config.toml`

Furthermore, it would "be nice" for customized commands to also be included in the command palette.
Scenarios would be:
* Typable command binding with description: `Custom | C-k | Open config`
* Command Sequence with description: `Custom | . | Goto last modification and view in center`
* Command Sequence with no description: `Custom | . | goto_last_modification → align_view_center`

I began implementing this functionality but gave up as a cleaner implementation would leverage refactorings of commands.rs, which definitely drop the chances of this PR getting merged down to 0. So I'll leave that for a later date.

I can open up an issue and move this part there if you would deem it appropriate.

## Conflicting PRs (16)

#5435 #5523 #5577 #5621 #5645 #5660 #5748 #5839 #6056 #6059 #6061 #6065 #6067 #4443 #2869 #3393
\* That are mergeable with master and pass all CI checks as of 2023-02-22


helix-term/src/application.rs
```diff
@@ -15,28 +13,26 @@ use helix_view::{
...
     commands::apply_workspace_edit,
     compositor::{Compositor, Event},
     config::Config,
+    keymap::Keymap,
     job::Jobs,
-    keymap::Keymaps,
     ui::{self, overlay::overlayed},
 };
-
-use log::{debug, error, warn};
```
the-mikedavis, line 31: There are a lot of cosmetic changes like the use declaration reorderings here or fields/variable renamings like `doc`->`description` below. Unless the name/ordering is horribly misleading you should avoid making these changes because it makes the PR harder to review and more likely to incur conflicts with other open PRs

gibbz00: I gave my reason for those in the commits. It's easier to following the reasoning to inspect one at the time. There it's clearly shown how "Keymaps" doesn't exist anymore, and that I renamed it to description for it not the be confused with the :help feature that is on it's way.

the-mikedavis: Replacing `keymap::Keymaps`->`keymap::Keymap` in this block is not a problem if the type is changing. I actually didn't notice that was a real change when selecting this block because it's obscured by the unecessary changes to the other use declarations.

d5520e8 should be dropped from this PR: a renaming like that belongs on that feature branch instead and shouldn't be made preemptively

gibbz00: Another reasion that I failed to mention explicitly in the commit though, is that im using description for it to stay conistent with the naming of each KeytrieNode. Current code uses doc and desc description for both.
https://github.com/helix-editor/helix/blob/17acadb305bebfef094e4fcc9ba1e6fda6159c3b/helix-term/src/keymap.rs#L82-L99

gibbz00: Compare this to:

https://github.com/gibbz00/helix/blob/keymap-presentation/helix-term/src/keymap/keytrie.rs#L107-L130

helix-term/src/config.rs
```diff
@@ -43,31 +40,38 @@ impl Display for ConfigLoadError {
 }
 
 impl Config {
-    pub fn load(config_path: PathBuf) -> Result<Config, ConfigLoadError> {
-        match std::fs::read_to_string(config_path) {
+    // REFACTOR? code similar to config assignment in main.rs,
```
gibbz00, line 43: Done in #5636

gibbz00:
@the-mikedavis @archseer PR feels ready for a review <3

archseer:
Commits in this branch now seem duplicated. I see "Keymap infobox: Use Vec in place of BTree" appearing twice for example.

GloopShlugger:
I've been testing it and noticed setting a description/label for a menu like spacemode gets overridden by the default keymap, there have been no issues with sticky

Maybe out of scope but would it be possible to have an ordering method based on toml config order? 

Great work

gibbz00:
> Commits in this branch now seem duplicated. I see "Keymap infobox: Use Vec in place of BTree" appearing twice for example.

I'm not sure how that happened :/ only saw that now. Suspect it happened in conjunction when cherry-picking commits from last PR and rebasing with master. Anyway I can fix that or is it ok if it is let be?

gibbz00:
> I've been testing it and noticed setting a description/label for a menu like spacemode gets overridden by the default keymap, there have been no issues with sticky

Thanks for letting me now! I added some failing tests for that,, and fixed them.

Ordering method based on toml config order is unfornutely out of my reach and interest frankly. One would still have to live with them not being ordered together with the pre-defined ordering, unless one rewrites the entire default keymap in a config.toml... Hence why I just prefer the automatic sorting. Also makes it very easy to spot which keys are't mapped.

GloopShlugger:
> Ordering method based on toml config order is unfortunately out of my reach and interest frankly. One would still have to live with them not being ordered together with the pre-defined ordering, unless one rewrites the entire default keymap in a config.toml... Hence why I just prefer the automatic sorting. Also makes it very easy to spot which keys aren't mapped.

Rewriting the entire keymap is exactly my use-case and only just noticed in that case its aleady toml order (after setting `sorted-infobox` to false) except for when you override default keys 

Would it make sense for future proofing sake to change the `sorted-infobox` config key from boolean to something like `by_key`, so future PR's can implement different behaviors? I do see this as a fairly menial point and would understand to just keep it as is

Because of that behavior once something like `default-keymap = false` lands it should be solved, for all the cases where the entire keymap is replaced

gibbz00:
I can look into it when time allows!!

Curious though, how come you want to rewrite the entire keymap?

GloopShlugger:
> I can look into it when time allows!!

Thank you! no rush

eugenesvk:
> I can look into it when time allows!!

Thanks, I have the same use case as I've rewrote the entire keymap for similar reasons (keycap mnemonics aren't useful as opposed to better locations), so would be nice if manual ordering were available

HactarCE:
Any update on this? / Anything I can do to help it move faster?

gibbz00:
> Any update on this? / Anything I can do to help it move faster?

There are actually! I've recently began splitting this PR into smaller ones, and I'm currently waiting for those that I've sent in to be merged before moving onto the next. First major one being https://github.com/helix-editor/helix/pull/7216 :)

daedroza:
@gibbz00 : would it be possible to rebase this on latest head?

gibbz00:
> @gibbz00 : would it be possible to rebase this on latest head?

Possible? Always :) Worth it? Probably not, at least not for me. I simply don't have the time to put efforts into PRs that I know won't get merged. See: https://github.com/helix-editor/helix/pull/5635#issuecomment-1596027479.

pascalkuthe:
Closing as stale

PR #5638, opened 2023-01-22
Add ayu_evolve theme

gibbz00:
Add ayu_evolve theme

An even darker 😈 version of ayu that leverages many of the newer theming features.

Differentiates:

* Primary non-primary selection
* Primary cursor between modes
  - Primary: Orange, Insert: Green, Select: Purple
* Variable members
* Inline raw markdown
* Strings and constructors
* Built in types
* Storage keywords
* Diagnostic underline color by type
* `ui.text.inactive` from #5104

2023-01-22-221601_screenshot

image

image

### DIFF
--- a/runtime/themes/ayu_evolve.toml
+++ b/runtime/themes/ayu_evolve.toml
@@ -3,7 +3,7 @@ inherits = 'ayu_dark'
 "keyword.control" = "orange"
 "keyword.storage" = "yellow"
 "keyword.storage.modifier" = "magenta"
-"variable.other.member" = "#gray"
+"variable.other.member" = "gray"
 "variable" = "light_gray"
 "constructor" = "magenta"
 "type.builtin" = { fg = "blue", modifiers = ["italic"] }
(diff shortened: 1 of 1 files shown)


runtime/themes/ayu_evolve.toml
```diff
@@ -0,0 +1,33 @@
+inherits = 'ayu_dark'
+
+"keyword.control" = "orange"
+"keyword.storage" = "yellow"
+"keyword.storage.modifier" = "magenta"
+"variable.other.member" = "bbbbbb"
```
the-mikedavis, line 6: this migth be a typo, this palette key doesn't exist

gibbz00: Hey, thanks for catching that, a # was missing. Should be fixed by now.

runtime/themes/ayu_evolve.toml
```diff
@@ -0,0 +1,36 @@
+inherits = 'ayu_dark'
+
+"keyword.control" = "orange"
+"keyword.storage" = "yellow"
+"keyword.storage.modifier" = "magenta"
+"variable.other.member" = "#gray"
```
the-mikedavis, line 6:
```suggestion
"variable.other.member" = "gray"
```

to use the palette color

PR #5639, opened 2023-01-23
Fix selecting a changed file in global search

### LINKED ISSUE #5604: Helix panics if we try to jump to a non existent line
### Summary

As the title says.

### I tried this:

1. `hx .` and then select a random file.
2. In a different terminal, I ran `yarn add zod`.
3. Back in helix I did a global search on `zod` (using `<space>` + `/`). Then I highlighted the entry in the `yarn.lock`. But I did not open the file.
4. I went to the terminal, and did `yarn remove zod`
5. Went back to helix and selected the entry highlighted in step 3.

### I expected this to happen:

Helix to not panic and show an error saying `the line you jumped to does not exist`

### Instead, this happened:

Helix panicked with this error:

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Line index out of bounds: line index 4998, Rope/RopeSlice line count 4993', /build/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.1-alpha/src/rope.rs:764:41
```

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty 0.12.0-dev (ed67aa3c)

### Helix Version

helix 22.12

pascalkuthe:
fixes #5604

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2003,6 +2003,10 @@ fn global_search(cx: &mut Context) {
                         let line_num = *line_num;
                         let (view, doc) = current!(cx.editor);
                         let text = doc.text();
+                        if line_num >= text.len_lines() {
+                            cx.editor.set_error("The line you jumped to does not exist anymore because the file has changed.");
+                            return;
+                        }
                         let start = text.line_to_char(line_num);
                         let end = text.line_to_char((line_num + 1).min(text.len_lines()));
 


pascalkuthe:
See https://github.com/helix-editor/helix/issues/5604#issuecomment-1399674583 for more details

the-mikedavis (approved):
I agree, let's fix the panic for now since it's easy to trigger. It would be really nice to use the buffer contents rather than disk contents for global search for any open files as you say. I think that might be tricky to do with the search we use from the `ignore` crate though

Issue #3035, opened 2022-07-11
Git blame and URL


mangas:
#### Describe your feature request
Would be great to add a git blame hint (toggle on config) 

As part of this, being able to go to the repo url with a key bind is very useful. 

Both of these are provided by this [vscode extension](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens)

Happy to also give a hand with this if the maintainers agree with this belonging on the editor itself as opposed to a plugin later (I feel like if this is the case we could add it now and move it out later)

the-mikedavis:
Blame information I think could fit well into the core. See also https://github.com/helix-editor/helix/issues/227

For opening the repo url, if you mean just opening up the main repository page, you can do that currently with a binding like

```toml
[keys.normal]
space.o = ":sh gh repo view --web"
```

for github repos.

Provider-specific features (like in [the plugin](https://marketplace.visualstudio.com/items?itemName=eamodio.gitlens#remote-provider-integrations-) should not be in core: provider-specific logic like how to format git forge URLs is undesirable for core unless it's based on an open standard (which does not seem to exist for any of the popular git forges).

mangas:
For github i was thinking something like [this](https://github.com/gitkraken/vscode-gitlens/blob/main/src/git/remotes/github.ts#L193) where you can just get the repo open on a browser pointing to the exact file, branch and line. I think despite this being provider it's quite easily extendable and could work even before there are plugins. Later it could just be moved to plugins once they are available?

the-mikedavis:
I think that would be handy in a plugin but that's the provider-specific logic that I think is distasteful to add to core. I think it's ok in a git-lens-like plugin that focuses on it and can be added I suspect without too much effort in a fork. I'd prefer that it wouldn't be merged in here though, even temporarily.

IMO, code that belongs in plugins shouldn't be added to core unless it can help advance other core features. File tree (#200) is a good example of this which was scoped as a plugin but is making its way into core (#2377). It adds a tree UI component that could be re-used else-where like in DAP or #1075.

heliostatic:
Since `gh` accepts filename and line number as arguments for `browse`, we just need to be able to pass current filename and line number as arguments in the `:sh gh browse` command.

Is that currently possible, @the-mikedavis ? If not, that does seem like a generally useful enhancement.

mangas:
[...]

PR #5645, opened 2023-01-23
Changed file picker

xJonathanLEI:
_I just started learning Git internals so I might be doing things wrong. Please feel free to correct me :)_

_See known issues [here](https://github.com/helix-editor/helix/pull/5645#issuecomment-1482451539)_

Connects #5362.

This PR adds a "Git status" file picker that's intended to mimic the list of files from a `git status` command invocation.

I find this feature very useful for picking up previous work when launching a new editing session. In fact, it's useful even when you're in the middle of a session - as you open more buffers from LSP commands etc. it gets harder and harder to find the files you're actually editing, as the number of buffers opened grows.

I set the default key binding of this picker to `<SPACE>g` and moved the deubgging menu to `<SPACE>G` because I think it makes the most sense:

- we already have `]g` and `[g` for navigating through the VCS changes; and
- `<SPACE>f` and `<SPACE>g` are close both in location and functionality.

I added the `sha1` dependency as SHA1 performance matters a lot. This is the same crate `gitoxide` uses if we enable the `fast-sha1` feature on the crate (we didn't). I could have enabled the feature and use the `Sha1` from `gitoxide` instead of importing it myself. I didn't do it beacuse unfortuantely `gitoxide` uses a wrapper around `Sha1` and the wrapper doesn't implement `Write`, making it impossible to utilize the `std::io::copy` method.

This PR is obviously in no way complete. I'm still submitting it first as I believe this feature has a lot of value even in its current form. Missing stuffs (at least):

- ~parallelization of working tree traversal and SHA1 calculation~
- ~detection of renaming w/ minor edits (which `git status` is able to do)~
- ~handling of CRLF transformation of huge files (see code comments)~

_(edit: these are no longer the case with the latest `gix` integration; see thread for details)_

With that said, the feature itself works fine. It's been tested on Ubuntu, Windows 11 (with `autocrlf` of course, not WSL), macOS. All working perfectly.

Screenshot on macOS with stock config:

Screenshot on macOS

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -324,6 +324,7 @@ impl MappableCommand {
         buffer_picker, "Open buffer picker",
         jumplist_picker, "Open jumplist picker",
         symbol_picker, "Open symbol picker",
+        changed_file_picker, "Open changed file picker",
         select_references_to_symbol_under_cursor, "Select symbol references",
         workspace_symbol_picker, "Open workspace symbol picker",
         diagnostics_picker, "Open diagnostic picker",
@@ -2991,6 +2992,46 @@ fn jumplist_picker(cx: &mut Context) {
         let line = meta.selection.primary().cursor_line(doc.text().slice(..));
         Some((meta.id.into(), Some((line, line))))
     });
+    cx.push_layer(Box::new(overlaid(picker)));
+}
+
+fn changed_file_picker(cx: &mut Context) {
+    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
+
+    let entries = match cx.editor.diff_providers.get_changed_files(&cwd) {
+        Ok(entries) => entries,
+        Err(err) => {
+            cx.editor.set_error(format!("{err}"));
+            return;
+        }
+    };
+
+    let added = cx.editor.theme.get("diff.plus");
+    let deleted = cx.editor.theme.get("diff.minus");
+    let modified = cx.editor.theme.get("diff.delta");
+
+    let picker = Picker::new(
+        entries,
+        ui::menu::FileChangeData {
+            cwd,
+            style_untracked: added,
+            style_modified: modified,
+            style_deleted: deleted,
+            style_renamed: modified,
+        },
+        |cx, meta, action| {
+            let path_to_open = meta.path();
+            if let Err(e) = cx.editor.open(path_to_open, action) {
+                let err = if let Some(err) = e.source() {
+                    format!("{}", err)
+                } else {
+                    format!("unable to open \"{}\"", path_to_open.display())
+                };
+                cx.editor.set_error(err);
+            }
+        },
+    )
+    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3008,6 +3008,7 @@ fn changed_file_picker(cx: &mut Context) {
             cwd: cwd.clone(),
             style_untracked: added,
             style_modified: modified,
+            style_conflict: deleted,
             style_deleted: deleted,
             style_renamed: modified,
         },
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2996,42 +2996,7 @@ fn jumplist_picker(cx: &mut Context) {
 }
 
 fn changed_file_picker(cx: &mut Context) {
-    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
-
-    let entries = match cx.editor.diff_providers.get_changed_files(&cwd) {
-        Ok(entries) => entries,
-        Err(err) => {
-            cx.editor.set_error(format!("{err}"));
-            return;
-        }
-    };
-
-    let added = cx.editor.theme.get("diff.plus");
-    let deleted = cx.editor.theme.get("diff.minus");
-    let modified = cx.editor.theme.get("diff.delta");
-
-    let picker = Picker::new(
-        entries,
-        ui::menu::FileChangeData {
-            cwd,
-            style_untracked: added,
-            style_modified: modified,
-            style_deleted: deleted,
-            style_renamed: modified,
-        },
-        |cx, meta, action| {
-            let path_to_open = meta.path();
-            if let Err(e) = cx.editor.open(path_to_open, action) {
-                let err = if let Some(err) = e.source() {
-                    format!("{}", err)
-                } else {
-                    format!("unable to open \"{}\"", path_to_open.display())
-                };
-                cx.editor.set_error(err);
-            }
-        },
-    )
-    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
+    let picker = ui::changed_file_picker(cx.editor);
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2999,8 +2999,10 @@ fn changed_file_picker(cx: &mut Context) {
     let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
 
     let added = cx.editor.theme.get("diff.plus");
-    let deleted = cx.editor.theme.get("diff.minus");
     let modified = cx.editor.theme.get("diff.delta");
+    let conflict = cx.editor.theme.get("diff.delta.conflict");
+    let deleted = cx.editor.theme.get("diff.minus");
+    let renamed = cx.editor.theme.get("diff.delta.moved");
 
     let picker = Picker::new(
         Vec::new(),
@@ -3008,9 +3010,9 @@ fn changed_file_picker(cx: &mut Context) {
             cwd: cwd.clone(),
             style_untracked: added,
             style_modified: modified,
-            style_conflict: deleted,
+            style_conflict: conflict,
             style_deleted: deleted,
-            style_renamed: modified,
+            style_renamed: renamed,
         },
         |cx, meta: &FileChange, action| {
             let path_to_open = meta.path();
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3032,7 +3032,7 @@ fn changed_file_picker(cx: &mut Context) {
                 ),
             };
 
-            Row::new(vec![
+            Row::new([
                 Cell::from(Span::styled(sign, style)),
                 Cell::from(content),
             ])
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4,7 +4,7 @@ pub(crate) mod typed;
 
 pub use dap::*;
 use helix_stdx::rope::{self, RopeSliceExt};
-use helix_vcs::Hunk;
+use helix_vcs::{FileChange, Hunk};
 pub use lsp::*;
 use tui::widgets::Row;
 pub use typed::*;
@@ -2996,7 +2996,43 @@ fn jumplist_picker(cx: &mut Context) {
 }
 
 fn changed_file_picker(cx: &mut Context) {
-    let picker = ui::changed_file_picker(cx.editor);
+    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
+
+    let added = cx.editor.theme.get("diff.plus");
+    let deleted = cx.editor.theme.get("diff.minus");
+    let modified = cx.editor.theme.get("diff.delta");
+
+    let picker = Picker::new(
+        Vec::new(),
+        ui::menu::FileChangeData {
+            cwd: cwd.clone(),
+            style_untracked: added,
+            style_modified: modified,
+            style_deleted: deleted,
+            style_renamed: modified,
+        },
+        |cx, meta: &FileChange, action| {
+            let path_to_open = meta.path();
+            if let Err(e) = cx.editor.open(path_to_open, action) {
+                let err = if let Some(err) = e.source() {
+                    format!("{}", err)
+                } else {
+                    format!("unable to open \"{}\"", path_to_open.display())
+                };
+                cx.editor.set_error(err);
+            }
+        },
+    )
+    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
+    let injector = picker.injector();
+
+    cx.editor.diff_providers.clone().for_each_changed_file(
+        cwd,
+        move |change| injector.push(change).is_ok(),
+        |err| {
+            helix_event::status::report_blocking(err);
+        },
+    );
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -6,7 +6,10 @@ pub use dap::*;
 use helix_stdx::rope::{self, RopeSliceExt};
 use helix_vcs::{FileChange, Hunk};
 pub use lsp::*;
-use tui::widgets::Row;
+use tui::{
+    text::Span,
+    widgets::{Cell, Row},
+};
 pub use typed::*;
 
 use helix_core::{
@@ -39,6 +42,7 @@ use helix_view::{
     info::Info,
     input::KeyEvent,
     keyboard::KeyCode,
+    theme::Style,
     tree,
     view::View,
     Document, DocumentId, Editor, ViewId,
@@ -54,7 +58,7 @@ use crate::{
     filter_picker_entry,
     job::Callback,
     keymap::ReverseKeymap,
-    ui::{self, overlay::overlaid, Picker, Popup, Prompt, PromptEvent},
+    ui::{self, menu::Item, overlay::overlaid, Picker, Popup, Prompt, PromptEvent},
 };
 
 use crate::job::{self, Jobs};
@@ -2996,6 +3000,45 @@ fn jumplist_picker(cx: &mut Context) {
 }
 
 fn changed_file_picker(cx: &mut Context) {
+    pub struct FileChangeData {
+        cwd: PathBuf,
+        style_untracked: Style,
+        style_modified: Style,
+        style_conflict: Style,
+        style_deleted: Style,
+        style_renamed: Style,
+    }
+
+    impl Item for FileChange {
+        type Data = FileChangeData;
+
+        fn format(&self, data: &Self::Data) -> Row {
+            let process_path = |path: &PathBuf| {
+                path.strip_prefix(&data.cwd)
+                    .unwrap_or(path)
+                    .display()
+                    .to_string()
+            };
+
+            let (sign, style, content) = match self {
+                Self::Untracked { path } => ("[+]", data.style_untracked, process_path(path)),
+                Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
+                Self::Conflict { path } => ("[x]", data.style_conflict, process_path(path)),
+                Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
+                Self::Renamed { from_path, to_path } => (
+                    "[>]",
+                    data.style_renamed,
+                    format!("{} -> {}", process_path(from_path), process_path(to_path)),
+                ),
+            };
+
+            Row::new(vec![
+                Cell::from(Span::styled(sign, style)),
+                Cell::from(content),
+            ])
+        }
+    }
+
     let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
 
     let added = cx.editor.theme.get("diff.plus");
@@ -3006,7 +3049,7 @@ fn changed_file_picker(cx: &mut Context) {
 
     let picker = Picker::new(
         Vec::new(),
-        ui::menu::FileChangeData {
+        FileChangeData {
             cwd: cwd.clone(),
             style_untracked: added,
             style_modified: modified,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3,6 +3,7 @@ pub(crate) mod lsp;
 pub(crate) mod typed;
 
 pub use dap::*;
+use helix_event::status;
 use helix_stdx::rope::{self, RopeSliceExt};
 use helix_vcs::{FileChange, Hunk};
 pub use lsp::*;
@@ -3074,13 +3075,16 @@ fn changed_file_picker(cx: &mut Context) {
     .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
     let injector = picker.injector();
 
-    cx.editor.diff_providers.clone().for_each_changed_file(
-        cwd,
-        move |change| injector.push(change).is_ok(),
-        |err| {
-            helix_event::status::report_blocking(err);
-        },
-    );
+    cx.editor
+        .diff_providers
+        .clone()
+        .for_each_changed_file(cwd, move |change| match change {
+            Ok(change) => injector.push(change).is_ok(),
+            Err(err) => {
+                status::report_blocking(err);
+                true
+            }
+        });
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3032,14 +3032,16 @@ fn changed_file_picker(cx: &mut Context) {
                 ),
             };
 
-            Row::new([
-                Cell::from(Span::styled(sign, style)),
-                Cell::from(content),
-            ])
+            Row::new([Cell::from(Span::styled(sign, style)), Cell::from(content)])
         }
     }
 
-    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
+    let cwd = helix_stdx::env::current_working_dir();
+    if !cwd.exists() {
+        cx.editor
+            .set_error("Current working directory does not exist");
+        return;
+    }
 
     let added = cx.editor.theme.get("diff.plus");
     let modified = cx.editor.theme.get("diff.delta");
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -14,9 +14,11 @@ use tui::{
 
 pub use tui::widgets::{Cell, Row};
 
+use helix_vcs::FileChange;
 use helix_view::{
     editor::SmartTabConfig,
     graphics::{Margin, Rect},
+    theme::Style,
     Editor,
 };
 use tui::layout::Constraint;
@@ -51,6 +53,37 @@ impl Item for PathBuf {
 }
 
 pub type MenuCallback<T> = Box<dyn Fn(&mut Editor, Option<&T>, MenuEvent)>;
+
+pub struct FileChangeData {
+    pub cwd: PathBuf,
+    pub style_untracked: Style,
+    pub style_modified: Style,
+    pub style_deleted: Style,
+    pub style_renamed: Style,
+}
+
+impl Item for FileChange {
+    type Data = FileChangeData;
+
+    fn format(&self, data: &Self::Data) -> Row {
+        let (sign, style) = match self {
+            Self::Untracked { .. } => ("[+]", data.style_untracked),
+            Self::Modified { .. } => ("[~]", data.style_modified),
+            Self::Deleted { .. } => ("[-]", data.style_deleted),
+            Self::Renamed { .. } => ("[>]", data.style_modified),
+        };
+        let path = self.path();
+
+        Row::new(vec![
+            sign.to_owned(),
+            path.strip_prefix(&data.cwd)
+                .unwrap_or(path)
+                .to_string_lossy()
+                .to_string(),
+        ])
+        .style(style)
+    }
+}
 
 pub struct Menu<T: Item> {
     options: Vec<T>,
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -66,22 +66,25 @@ impl Item for FileChange {
     type Data = FileChangeData;
 
     fn format(&self, data: &Self::Data) -> Row {
-        let (sign, style) = match self {
-            Self::Untracked { .. } => ("[+]", data.style_untracked),
-            Self::Modified { .. } => ("[~]", data.style_modified),
-            Self::Deleted { .. } => ("[-]", data.style_deleted),
-            Self::Renamed { .. } => ("[>]", data.style_modified),
-        };
-        let path = self.path();
-
-        Row::new(vec![
-            sign.to_owned(),
+        let process_path = |path: &PathBuf| {
             path.strip_prefix(&data.cwd)
                 .unwrap_or(path)
-                .to_string_lossy()
-                .to_string(),
-        ])
-        .style(style)
+                .display()
+                .to_string()
+        };
+
+        let (sign, style, content) = match self {
+            Self::Untracked { path } => ("[+]", data.style_untracked, process_path(path)),
+            Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
+            Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
+            Self::Renamed { from_path, to_path } => (
+                "[>]",
+                data.style_modified,
+                format!("{} -> {}", process_path(from_path), process_path(to_path)),
+            ),
+        };
+
+        Row::new(vec![sign.to_owned(), content]).style(style)
     }
 }
 
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -9,6 +9,7 @@ use nucleo::pattern::{Atom, AtomKind, CaseMatching};
 use nucleo::{Config, Utf32Str};
 use tui::{
     buffer::Buffer as Surface,
+    text::Span,
     widgets::{Block, Borders, Table, Widget},
 };
 
@@ -84,7 +85,10 @@ impl Item for FileChange {
             ),
         };
 
-        Row::new(vec![sign.to_owned(), content]).style(style)
+        Row::new(vec![
+            Cell::from(Span::styled(sign, style)),
+            Cell::from(content),
+        ])
     }
 }
 
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -59,6 +59,7 @@ pub struct FileChangeData {
     pub cwd: PathBuf,
     pub style_untracked: Style,
     pub style_modified: Style,
+    pub style_conflict: Style,
     pub style_deleted: Style,
     pub style_renamed: Style,
 }
@@ -77,6 +78,7 @@ impl Item for FileChange {
         let (sign, style, content) = match self {
             Self::Untracked { path } => ("[+]", data.style_untracked, process_path(path)),
             Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
+            Self::Conflict { path } => ("[x]", data.style_conflict, process_path(path)),
             Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
             Self::Renamed { from_path, to_path } => (
                 "[>]",
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -9,17 +9,14 @@ use nucleo::pattern::{Atom, AtomKind, CaseMatching};
 use nucleo::{Config, Utf32Str};
 use tui::{
     buffer::Buffer as Surface,
-    text::Span,
     widgets::{Block, Borders, Table, Widget},
 };
 
 pub use tui::widgets::{Cell, Row};
 
-use helix_vcs::FileChange;
 use helix_view::{
     editor::SmartTabConfig,
     graphics::{Margin, Rect},
-    theme::Style,
     Editor,
 };
 use tui::layout::Constraint;
@@ -54,45 +51,6 @@ impl Item for PathBuf {
 }
 
 pub type MenuCallback<T> = Box<dyn Fn(&mut Editor, Option<&T>, MenuEvent)>;
-
-pub struct FileChangeData {
-    pub cwd: PathBuf,
-    pub style_untracked: Style,
-    pub style_modified: Style,
-    pub style_conflict: Style,
-    pub style_deleted: Style,
-    pub style_renamed: Style,
-}
-
-impl Item for FileChange {
-    type Data = FileChangeData;
-
-    fn format(&self, data: &Self::Data) -> Row {
-        let process_path = |path: &PathBuf| {
-            path.strip_prefix(&data.cwd)
-                .unwrap_or(path)
-                .display()
-                .to_string()
-        };
-
-        let (sign, style, content) = match self {
-            Self::Untracked { path } => ("[+]", data.style_untracked, process_path(path)),
-            Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
-            Self::Conflict { path } => ("[x]", data.style_conflict, process_path(path)),
-            Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
-            Self::Renamed { from_path, to_path } => (
-                "[>]",
-                data.style_modified,
-                format!("{} -> {}", process_path(from_path), process_path(to_path)),
-            ),
-        };
-
-        Row::new(vec![
-            Cell::from(Span::styled(sign, style)),
-            Cell::from(content),
-        ])
-    }
-}
 
 pub struct Menu<T: Item> {
     options: Vec<T>,
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -19,7 +19,6 @@ use crate::job::{self, Callback};
 pub use completion::{Completion, CompletionItem};
 pub use editor::EditorView;
 use helix_stdx::rope;
-use helix_vcs::FileChange;
 pub use markdown::Markdown;
 pub use menu::{FileChangeData, Menu};
 pub use picker::{DynamicPicker, FileLocation, Picker};
@@ -253,47 +252,6 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
             }
         });
     }
-    picker
-}
-
-pub fn changed_file_picker(editor: &helix_view::Editor) -> Picker<FileChange> {
-    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
-
-    let added = editor.theme.get("diff.plus");
-    let deleted = editor.theme.get("diff.minus");
-    let modified = editor.theme.get("diff.delta");
-
-    let picker = Picker::new(
-        Vec::new(),
-        menu::FileChangeData {
-            cwd: cwd.clone(),
-            style_untracked: added,
-            style_modified: modified,
-            style_deleted: deleted,
-            style_renamed: modified,
-        },
-        |cx, meta: &FileChange, action| {
-            let path_to_open = meta.path();
-            if let Err(e) = cx.editor.open(path_to_open, action) {
-                let err = if let Some(err) = e.source() {
-                    format!("{}", err)
-                } else {
-                    format!("unable to open \"{}\"", path_to_open.display())
-                };
-                cx.editor.set_error(err);
-            }
-        },
-    )
-    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
-    let injector = picker.injector();
-
-    editor.diff_providers.clone().for_each_changed_file(
-        cwd,
-        move |change| injector.push(change).is_ok(),
-        |err| {
-            helix_event::status::report_blocking(err);
-        },
-    );
     picker
 }
 
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -287,20 +287,13 @@ pub fn changed_file_picker(editor: &helix_view::Editor) -> Picker<FileChange> {
     .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
     let injector = picker.injector();
 
-    let diff_providers = editor.diff_providers.clone();
-
-    std::thread::spawn(move || {
-        // There's no way we can notify the editor since we're in a background thread. This just
-        // silently fails.
-        if let Ok(change_iter) = diff_providers.get_changed_files(&cwd) {
-            // We ignore errors as they could be caused by permission issues etc
-            for change in change_iter.flatten() {
-                if injector.push(change).is_err() {
-                    break;
-                }
-            }
-        }
-    });
+    editor.diff_providers.clone().for_each_changed_file(
+        cwd,
+        move |change| injector.push(change).is_ok(),
+        |err| {
+            helix_event::status::report_blocking(err);
+        },
+    );
     picker
 }
 
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -20,7 +20,7 @@ pub use completion::{Completion, CompletionItem};
 pub use editor::EditorView;
 use helix_stdx::rope;
 pub use markdown::Markdown;
-pub use menu::{FileChangeData, Menu};
+pub use menu::Menu;
 pub use picker::{DynamicPicker, FileLocation, Picker};
 pub use popup::Popup;
 pub use prompt::{Prompt, PromptEvent};
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -19,8 +19,9 @@ use crate::job::{self, Callback};
 pub use completion::{Completion, CompletionItem};
 pub use editor::EditorView;
 use helix_stdx::rope;
+use helix_vcs::FileChange;
 pub use markdown::Markdown;
-pub use menu::Menu;
+pub use menu::{FileChangeData, Menu};
 pub use picker::{DynamicPicker, FileLocation, Picker};
 pub use popup::Popup;
 pub use prompt::{Prompt, PromptEvent};
@@ -252,6 +253,54 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
             }
         });
     }
+    picker
+}
+
+pub fn changed_file_picker(editor: &helix_view::Editor) -> Picker<FileChange> {
+    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
+
+    let added = editor.theme.get("diff.plus");
+    let deleted = editor.theme.get("diff.minus");
+    let modified = editor.theme.get("diff.delta");
+
+    let picker = Picker::new(
+        Vec::new(),
+        menu::FileChangeData {
+            cwd: cwd.clone(),
+            style_untracked: added,
+            style_modified: modified,
+            style_deleted: deleted,
+            style_renamed: modified,
+        },
+        |cx, meta: &FileChange, action| {
+            let path_to_open = meta.path();
+            if let Err(e) = cx.editor.open(path_to_open, action) {
+                let err = if let Some(err) = e.source() {
+                    format!("{}", err)
+                } else {
+                    format!("unable to open \"{}\"", path_to_open.display())
+                };
+                cx.editor.set_error(err);
+            }
+        },
+    )
+    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
+    let injector = picker.injector();
+
+    let diff_providers = editor.diff_providers.clone();
+
+    std::thread::spawn(move || {
+        // There's no way we can notify the editor since we're in a background thread. This just
+        // silently fails.
+        if let Ok(change_iter) = diff_providers.get_changed_files(&cwd) {
+            // We ignore errors as they could be caused by permission issues etc
+            for change in change_iter.flatten() {
+                if injector.push(change).is_err() {
+                    break;
+                }
+            }
+        }
+    });
     picker
 }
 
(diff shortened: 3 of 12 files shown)


helix-term/src/commands.rs
```diff
@@ -2450,6 +2451,64 @@ fn jumplist_picker(cx: &mut Context) {
...
+fn git_status_picker(cx: &mut Context) {
+    cx.editor
+        .set_error("Helix was built without the \"git\" feature");
+}
+
+#[cfg(feature = "git")]
+fn git_status_picker(cx: &mut Context) {
+    use helix_vcs::Git;
+
+    let proj_root = find_root(None, &[]);
```
pascalkuthe, line 2464: This is unnecessary. Gitoxide handles git repo detection automatically and more accurately. Just passing the cwd to gitoxide is sufficent

helix-term/src/commands.rs
```diff
@@ -272,6 +272,7 @@ impl MappableCommand {
         buffer_picker, "Open buffer picker",
         jumplist_picker, "Open jumplist picker",
         symbol_picker, "Open symbol picker",
+        git_status_picker, "Open Git status picker",
```
pascalkuthe, line 275: I think this should be called changed_file_picker and default automatically detect the VCS. Furthermore everything git specific should got into helix-vcs and be hidden behind the `DiffProvider` trait so it's easy to add support for other VCS in the future. Probably something like `changed_files(&self, rev: Option<&str>) -> Option<(Vec<PathBuf, ChsngeStatus)>`

pascalkuthe (commented):
Thanks for working on this! I haven't looked at the implementation too  much but here are already things I noticed while scanning

helix-term/src/commands.rs
xJonathanLEI: Ah I actually looked into the diff gutter impl and was aware of of the trait, but thought that I'm not exactly "diffing" here so decided to leave the trait along. No problem will refactor to fix this.

helix-term/src/commands.rs
xJonathanLEI: Fixed in the latest push. It just uses cwd now.

helix-term/src/commands.rs
xJonathanLEI: Refactored everything into `helix-vcs` behind the `DiffProvider` trait. The code is much cleaner now indeed. Thanks!

haikyuu:
This will definitely become my most used picker right away. Thanks for working on this @xJonathanLEI 🙏

Byron:
>  I didn't do it beacuse unfortuantely gitoxide uses a wrapper around Sha1 and the wrapper doesn't implement Write, making it impossible to utilize the std::io::copy method.

Good news, what you were looking for [actually exists](https://docs.rs/git-features/0.26.1/git_features/hash/struct.Write.html) :)

> * parallelization of working tree traversal and SHA1 calculation
> * detection of renaming w/ minor edits (which git status is able to do)
> * handling of CRLF transformation of huge files (see code comments)

Even though nothing speaks against having a first version of these in this codebase, especially if you have the time and fun doing it, I truly think the right place for these capabilities is `gitoxide` itself. @pascalkuthe and I will be putting our heads together to find a good solution for this so that everybody is happy.

Please note that right now [I am super busy with the `cargo` integration](https://github.com/rust-lang/cargo/pull/11448) and thus can't jump on this myself, but I will support the efforts together with you to be part of getting the best possible solution both for `helix-editor` and for `gitoxide` :).

Exciting!

xJonathanLEI:
First off thank you for taking your time looking at this! @Byron

> Good news, what you were looking for [actually exists](https://docs.rs/git-features/0.26.1/git_features/hash/struct.Write.html) :)

Thanks for thr pointer! I somehow missed this type when looking at that file. However, it looks like a write multiplxer: it writes to an internal writer and the hasher _at the same time_. I don't really have the second writer to write to and actually only want the `Sha1` wrapper to implement `Write` (which is reasonable to do I assume given it's just a newtype wrapper). Should I submit the `Write` implementation patch to `gitoxide` and wait for a new release for this PR?

> I truly think the right place for these capabilities is `gitoxide` itself.

I agree. I started out building this PR by looking for similar features in `gitoxide` but no luck. IMO having the majority (other than the UI part of course) of this PR implemented in `gitoxide` makes the most sense simply because of specialization - Helix gets incremental improvements over time for free.

That's why I stopped after having this (reasonably) working POC - leaving the hard part (perf & edge cases) to Git masters like yourself and @pascalkuthe. I do see Helix abandoning this implementation in favor of a native out-of-the-box solution from `gitoxide` in the future.

On the other hand, I do believe this would be a highly wanted feature itself. That's why I'm still submitting it depsite things not being perfect.

> @pascalkuthe and I will be putting our heads together to find a good solution for this so that everybody is happy.

Amazing! We're in good hands!

xJonathanLEI:
While using this on one my of other repos, I realized that it's wrong that I thought a `Commit` entry cannot appear in the HEAD tree - turns out it can when it's a submodule. As a result, the diff provider chokes whenever there's a submodule. I just pushed a patch that simply ignores all submodules for now.

What do maintainers think? Should I complete the support for change detection in submodules in this PR or do we leave it for future work?

I have a feeling that the UX here is kinda hard to get right. If we just extend the current changed file list to include the submodule files, things are actually pretty messy: files are listed together but you can't really commit them together - they belong to different repositories. We probably need some kind of grouping when submodules exist.

xJonathanLEI:
Just realized there's a small bug: on Linux if a file is was committed as CRLF then it would be incorrectly detected as modified. This is actually not a new bug per ser. I followed the CRLF handling logic in #4995 by @pascalkuthe. Unsurprisingly, the file that is getting incorrectly flagged also has Git gutters on all lines.

I had a feeling that our CRLF handling was too naive and it turns out to be true. I guess we need to mimic how Git handles the option to get rid of this issue (along with the gutters).

xJonathanLEI:
Update on the CRLF issue: after investigating more I realized it's not related to the CRLF handling logic itself. The project I encountered this issue on has a `.gitattributes` file:

```
# Default.
*           text eol=lf

# Treat patch files as binaries but let diff'ing them
# as normal text.
*.diff      binary diff
*.patch     binary diff
*.patch32   binary diff
*.patch64   binary diff
*.patch.*   binary diff

# Powershell scripts.
*.ps1       text eol=crlf

# Binaries.
*.gpg       binary
*.gz        binary
*.jpg       binary
*.png       binary
*.tar       binary
*.tar.*     binary
```

To resolve this we need to be able to respect this file instead of blindly following `core.autocrlf`. I already implemented binary file detection but still need to add support for this one.

Again I'm not sure if this should be left as future work. Input from maintainers would be nice. Thanks!

helix-vcs/src/git.rs
```diff
@@ -51,6 +104,111 @@ impl Git {
...
+        let work_dir = repo
+            .work_dir()
+            .ok_or_else(|| anyhow::anyhow!("working tree not found"))?;
+
+        // TODO: allow diffing against another ref
+        let head_tree = repo.head_commit()?.tree()?;
+
+        let mut head_tree_files = vec![];
+        let mut submodule_paths = vec![PathBuf::from(".git")];
+        traverse_tree(
```
Byron, line 124: Instead of traversing the tree, using [the index](https://docs.rs/git-repository/0.33.0/git_repository/struct.Repository.html#method.index) is the better and faster way to going about these things - it can be assumed to be up-to-date as git updates it to match any checked out commit.

If the index was used, the entire tree-traversal part could be replaced with a loop over index entries which should remove some complexity.

If the `allow diffing against another ref` feature would have to be implemented, tree traversal would be a must - in that case I recommend creating [an index from a tree](https://docs.rs/git-index/0.12.2/git_index/struct.State.html#method.from_tree) so the main code-path deals with [the index](https://docs.rs/git-index/0.12.2/git_index/struct.State.html#) as primary data structure only.

pascalkuthe: Initially I wanted to implement that too bit aren't you then diffing between the index/stage instead of the work tree (so git status --cached)?

pascalkuthe:
> 

I am pretty sure the diff gutter doesn't work for submodules either so I would be fine with leaving that as future work for. Especially because git status also doesn't report submodule changes I believe



> Update on the CRLF issue: after investigating more I realized it's not related to the CRLF handling logic itself. The project I encountered this issue on has a `.gitattributes` file:
> 
> ```
> # Default.
> *           text eol=lf
> 
> # Treat patch files as binaries but let diff'ing them
> # as normal text.
> *.diff      binary diff
> *.patch     binary diff
> *.patch32   binary diff
> *.patch64   binary diff
> *.patch.*   binary diff
> 
> # Powershell scripts.
> *.ps1       text eol=crlf
> 
> # Binaries.
> *.gpg       binary
> *.gz        binary
> *.jpg       binary
> *.png       binary
> *.tar       binary
> *.tar.*     binary
> ```
> 
> To resolve this we need to be able to respect this file instead of blindly following `core.autocrlf`. I already implemented binary file detection but still need to add support for this one.
> 
> Again I'm not sure if this should be left as future work. Input from maintainers would be nice. Thanks!
> 
> _Edit: just read #4995 more carefully and turns out it's a known issue_

I believe attributes aren't fully supported yet in gitoxide. The diff gutter doesn't support then for this reason either so its fine if you ignore them to now

helix-vcs/src/git.rs
xJonathanLEI: > aren't you then diffing between the index/stage instead of the work tree

Unless I can create the index from the work tree? I'll look into it a bit. Thanks @Byron for the pointer!

Byron: That's true, sorry for that, maybe that's not even intended. In any case, it's easy to create an index from a tree, and that data structure is made for exactly this kind of work. Or in other words, with `index from tree` it's possible to compare against `HEAD` as well.

xJonathanLEI: Just pushed an update. I removed `traverse_tree` in favor of the `State` struct built from the head tree. I guess this should be resolved now?

Byron: Thanks, this gets rid of the manual traversal. 
I'd expect the `gitoxide` status to be faster than the one of `git`  when done, so it's probably better to wait for it instead of implementing everything here.

I wonder how low-hanging the 'avoidance of SHA1 computation' could be similar to how git does it, as I don't know how git efficiently compares the head tree with the index. I hope soon some more analysis will find its way [into the tracking issue](https://github.com/Byron/gitoxide/issues/301).

xJonathanLEI: @Byron Thanks for checking! I didn't like the recursion in the first place so it's good to have it removed.

> it's probably better to wait for it instead of implementing everything here

Are you suggesting we wait for `gitoxide` to implement this feature and use it in this PR?

Byron: Apologies, I guess I was rambling.

I think it's OK to leave this PR as is and look forward to the day when `gitoxide` can perform this task faster than `git` can.

xJonathanLEI:
PR conflicts with `master` now due to `Cargo.lock`. I just rebased and squashed.

poliorcetics:
> Resolves https://github.com/helix-editor/helix/issues/5362.

This PR doesn't resolve it, that's not the (only) picker I had in mind for the issue

helix-vcs/src/lib.rs
```diff
@@ -38,6 +46,13 @@ impl DiffProviderRegistry {
             .iter()
             .find_map(|provider| provider.get_diff_base(file))
     }
+
+    pub fn get_changed_files(&self, cwd: &Path) -> Result<Vec<FileChange>> {
+        self.providers
+            .iter()
+            .find_map(|provider| provider.get_changed_files(cwd).ok())
+            .ok_or_else(|| anyhow::anyhow!("no diff provider returns success"))
```
poliorcetics, line 54: I find this better than using a negation, it avoids having 'success' as the last word

```suggestion
            .ok_or_else(|| anyhow::anyhow!("all diff providers failed to list changed files"))
```

helix-vcs/src/git.rs
```diff
@@ -103,3 +340,128 @@ fn find_file_in_commit(repo: &Repository, commit: &Commit, file: &Path) -> Optio
         EntryMode::Blob | EntryMode::BlobExecutable => Some(tree_entry.object_id()),
     }
 }
+
+fn git_meta_from_path(
+    path: &Path,
+    autocrlf: bool,
+) -> Result<Option<(FileEntryMode, ObjectId)>, std::io::Error> {
```
poliorcetics, line 347:
```suggestion
) -> std::io::Result<Option<(FileEntryMode, ObjectId)>> {
```

helix-vcs/src/git.rs
```diff
@@ -103,3 +340,128 @@ fn find_file_in_commit(repo: &Repository, commit: &Commit, file: &Path) -> Optio
...
+                hasher.update(b"\0");
+                hasher.update(link_content);
+
+                let hash: [u8; 20] = hasher.finalize().into();
+
+                return Ok(Some((FileEntryMode::Link, ObjectId::from(hash))));
+            }
+        }
+        Err(_) => return Ok(None),
+    };
```
poliorcetics, line 369:
```suggestion
    // Windows doesn't support symlinks. This block runs fine but is just wasting CPU cycles.
    if cfg!(not(windows)) {
        match path.symlink_metadata() {
            Ok(meta) => {
                if meta.is_symlink() {
                    let link_content = std::fs::read_link(path)?;
                    let link_content = link_content.to_string_lossy();
                    let link_content = link_content.as_bytes();

                    let mut hasher = sha1::Sha1::default();
                    hasher.update(b"blob ");
                    hasher.update(format!("{}", link_content.len()).as_bytes());
                    hasher.update(b"\0");
                    hasher.update(link_content);

                    let hash: [u8; 20] = hasher.finalize().into();

                    return Ok(Some((FileEntryMode::Link, ObjectId::from(hash))));
                }
            }
            Err(_) => return Ok(None),
        }
    }
```

When the code compiles on all platform, I find it better to use `cfg!(...)` because it allows autocompletion everywhere, even when not compiling for the targted feature or arch, unlike `#[cfg(...)]`, so it provides a nicer contributor experience :)

poliorcetics: The indentation level can be reduced using a tuple match, but I'm not sure it's worth it here

helix-vcs/src/git.rs
```diff
@@ -103,3 +340,128 @@ fn find_file_in_commit(repo: &Repository, commit: &Commit, file: &Path) -> Optio
...
+                };
+
+                Some((entry_mode, oid))
+            } else {
+                // It's a non-symlink folder. Git doesn't track folders. Same as deletion.
+                None
+            }
+        }
+        Err(_) => None,
+    })
```
poliorcetics, line 466: the indentation levels here can be reduced by using this pattern, it would avoid the line `Err(_) => None` at the end that is difficult to link to the earlier `path.metadata()` call and the `if meta.is_file()` that is 99% of the `Ok` branch of the match

```rust
let metadata = match path.metadata() {
    Ok(m) if m.is_file() => m,
    // Ignore errors and non-files
    _ => return Ok(None),
}

/// ...
```

helix-vcs/src/git.rs
```diff
@@ -88,6 +241,90 @@ impl DiffProvider for Git {
...
+    }
+}
+
+impl From<RawChanges> for Vec<FileChange> {
+    // Unlike Git, we only look for pure renames at the moment.
+    // TODO: detect renames with minor changes
+    fn from(mut raw: RawChanges) -> Self {
+        let mut status_entries = vec![];
+
+        let additions_left = if !raw.additions.is_empty() && !raw.deletions.is_empty() {
```
poliorcetics, line 281: Inverting the condition would mean having the one-line `else` branch at the top and the big branch below, making it much easier to remember the condition when getting to the `else`

helix-vcs/src/git.rs
```diff
@@ -51,6 +99,111 @@ impl Git {
...
+            .ok_or_else(|| anyhow::anyhow!("working tree not found"))?;
+
+        // TODO: allow diffing against another ref
+        let head_tree = repo.head_commit()?.tree()?;
+        let head_state = State::from_tree(&head_tree.id, |oid, buf| {
+            repo.objects.find_tree_iter(oid, buf).ok()
+        })?;
+
+        let mut head_tree_set = HashSet::new();
+        let mut submodule_paths = vec![];
```
poliorcetics, line 121: Using the macro for an empty `Vec` seems wasteful

helix-vcs/src/git.rs
```diff
@@ -1,17 +1,65 @@
...
+}
+
+#[derive(Hash, PartialEq, Eq)]
+struct RawDeletion {
+    entry_mode: FileEntryMode,
+    oid: ObjectId,
+    path: PathBuf,
+}
+
+#[allow(unused)]
```
poliorcetics, line 52: Is this necessary :/ ?

helix-vcs/src/git.rs
```diff
@@ -1,17 +1,65 @@
...
 
-use crate::DiffProvider;
+use crate::{DiffProvider, FileChange};
 
 #[cfg(test)]
 mod test;
 
 pub struct Git;
 
+/// A subset of `git_repository::objs::tree::EntryMode` that actually makes sense for tree nodes.
```
poliorcetics, line 23: You can use `[..]` so that people building the doc locally will have a link to the enum

```suggestion
/// A subset of [`git_repository::objs::tree::EntryMode`] that actually makes sense for tree nodes.
```

helix-vcs/src/git.rs
```diff
@@ -51,6 +99,111 @@ impl Git {
...
+        let mut submodule_paths = vec![];
+
+        let mut raw_changes = RawChanges::default();
+
+        for item in head_state.entries() {
+            let full_path = work_dir.join(&PathBuf::from(item.path(&head_state).to_string()));
+
+            if item.mode == Mode::COMMIT {
+                submodule_paths.push(full_path);
+            } else {
```
poliorcetics, line 130: Removes an indentation level, but for this one I'm not sure it's better

```suggestion
            if item.mode == Mode::COMMIT {
                submodule_paths.push(full_path);
                continue;
            }
            
            // ...
```

helix-vcs/src/git.rs
```diff
@@ -51,6 +99,111 @@ impl Git {
...
+                            #[cfg(unix)]
+                            {
+                                new_entry_mode != old_entry_mode
+                            }
+
+                            #[cfg(not(unix))]
+                            {
+                                false
+                            }
+                        };
```
poliorcetics, line 152:
```suggestion
                        let entry_mode_changed = cfg!(unix) && new_entry_mode != old_entry_mode;
```

poliorcetics (changes requested):

xJonathanLEI:
Rebased on `master` due to merge conflicts.

xJonathanLEI:
Rebased on `master` due to merge conflicts.

helix-vcs/src/git.rs
```diff
@@ -87,6 +241,90 @@ impl DiffProvider for Git {
...
+            .into_iter()
+            .flat_map(|val| val.iter())
+            .for_each(|item| {
+                status_entries.push(FileChange::Deleted {
+                    path: item.path.to_owned(),
+                })
+            });
+        raw.modifications
+            .into_iter()
+            .for_each(|item| status_entries.push(FileChange::Modified { path: item.path }));
```
wongjiahau, line 324:
```suggestion
let status_entries = additions_left
    .into_iter()
    .map(|item| FileChange::Untracked { path: item.path })
    .chain(
        raw.deletions
            .values()
            .into_iter()
            .flat_map(|val| val.iter())
            .map(|item| {
                FileChange::Deleted {
                    path: item.path.to_owned(),
                }
            })
    )
    .chain(
        raw.modifications
            .into_iter()
            .map(|item| FileChange::Modified { path: item.path })
    )
     .collect::<Vec<_>>();
```

xJonathanLEI: Thanks for the suggestion but it looks like the suggested code isn't formatted? If I commit the suggestion CI will turn red I think :/

xJonathanLEI: It's fine. I will accept it on GitHub and amend locally :)

xJonathanLEI:
There are a few known issues on this PR base on my own usage:

- the diff provider fails when `HEAD` does not exist (e.g. right after `git init`)
- the picker would show the same file 3 times when resolving Git conflicts
- the diff provider fails when the process doesn't have permission to read certain folders/files

Technically I should fix these, but as confirmed by @pascalkuthe, this PR will be replaced with the upstream `gitoxide` implementation when it becomes available. So I'll simply keep the PR as is and only perform `master` rebase should conflicts arise to make it easier to those who want to daily drive with this PR to pick the branch.

Byron:
Thanks for the summary, please let me share some thoughts.

> the diff provider fails when HEAD does not exist (e.g. right after git init)

This would have to be adjusted here to work as expected. I presume not having git initialized would mark all present files as new (as there also is no `.git/index` then).

> the diff provider fails when the process doesn't have permission to read certain folders/files

@pascalkuthe already made me aware of an issue on Windows where the repository couldn't be opened without admin rights, and with admin rights it could be opened but some tests failed probably indicating that the git repository didn't actually yield any information to provide a base for diffing. That should probably be fixed upstream but I wasn't able to reproduce it.

To my mind it should be possible to fix the remaining issues here (except the permission issue which needs repro to be fixed in `gitoxide`) and make the PR usable that way, without waiting for the upstream implementation. If the `git status` implementation here is correct enough, the upstream version of it should 'just' bring performance and correctness in all cases. Thus the work done here *should* mostly be reusable.

xJonathanLEI:
Hi @Byron thanks for the input! But I was under the impression that the _whole_ thing will be replaced by an upstream impl. aka I will just call something like `repo.status()` and get the whole list of diffs that my impl here works to dig out. The PR would then only be about the UI part. Did I misunderstand something? Input from @pascalkuthe would be nice!

Byron:
Yes, this is my understanding as well. To me it felt the PR was blocked waiting for something that is potentially quite far away, and I thought that maybe a 'good enough' version could fill the gap meanwhile. Ultimately me chiming in might not have been the best idea as it caused confusion and I am sorry for that - @pascalkuthe has a plan and I don't mean to change it 😅.

xJonathanLEI:
Lol no worries. I think @pascalkuthe's plan is to leave the PR open until `gitoxide` has it. In that case, I'll just leave the PR as is cuz it's good enough for me despite those issues (those are non-issues in most cases anyways). Should the plan changes and we want to merge this one soon, I'm also happy to polish it up :)

pascalkuthe:
A little update here: An inital implementation of an index/worktree diff is about the merged upstream https://github.com/Byron/gitoxide/pull/805 (tomorrow if all goes to plan). This is still missing support for untracked files and filters. I will add those in followup PRs. Untracked files shouldn't be too hard and I will restrict myself to some basic git filters (like line endings) for now as implementing full git filter support would take quite a while. Once that is ready then this PR can start using the upstream functionality.

xJonathanLEI:
Looks like the removal of file picker from #7264 has introduced some conflicts here.

xJonathanLEI:
Studied that PR a bit and conflict resolved. Git diff _plugin™_ works again.

poliorcetics:
Most (all?) or my comments earlier in the pr are still unaddressed.

Also, this still doesn't fully resolve #5362, please don't mark the issue as closed by this MR

xJonathanLEI:
Look I'm not obligated to resolve your comments, and I don't have time nor care enough to. As already stated in the convo above, there is currently no plan to merge this PR and it's still maintained only for people (myself included) who daily drives with a "juiced up" version of Helix with pending PRs incorporated.

xJonathanLEI:
Hey it's one year anniversary of this PR! 🎉🎉

Just resolved `gix` breaking changes from `gix-v0.55.0..gix-v0.57.1`. For those (if any) that still daily drive with this branch, enjoy!

Byron:
Just chiming in to mention [the `gix status` related work](https://github.com/Byron/gitoxide/pull/1252) over at `gitoxide`  which might be relevant here when done.

daedroza:
Would something like git blame current line can also be added for juicing up further?

the-mikedavis:
That's https://github.com/helix-editor/helix/issues/3035, it's not related to this PR

pascalkuthe:
@Byron I saw you made some great progress on the gutoxide front (with more work still underway). Is gitlxide Boe at a point where we could use it to implement a prototype of this feature in helix?

Byron:
In this [this PR](https://github.com/Byron/gitoxide/pull/1306) over at `gitoxide` is a complete implementation of index-worktree changes including rename tracking. It lacks the `HEAD` - `index` diff still, but you seem to not be needing this here.

I plan to first integrate what's there into the higher-level `gix` crate which is probably what you would rather use here, and I will definitely be back here to let you know when it's available :). Definitely closing in… .

(*Note the my reply is unrelated to `git blame` which was marked off-topic*)

Byron:
I am happy to announce that `gix` v0.60 has just been released. It comes with `repo.status(…).into_index_worktree_iter(…)` which can be used to obtain all the information you would need, and even has rename/rewrite and copy tracking if you choose to enable it.

The configuration is quite ~~complex~~ powerful, and if you have questions, just ping me and I will be here to help. I'd also be happy to review the `gitoxide` part of the PR to see if the API works for you, or if it needs improvements.

Good luck :)!

xJonathanLEI:
Good news. I've got something working locally with the new `gix` API. All change detection code from this PR has been removed. I will daily drive with it for a few more days before pushing for review :)

xJonathanLEI:
@Byron The API works great and the impl is blazing fast! Amazing work!

The only issue I'm having right now is that submodule changes refuse to show up. I'm not getting any `SubmoduleModification` item from the iterator. Tried messing around with `check_dirty` and nothing changes (should be irrelevant anyways). Any quick pointers? Should I send a full reproducer?

Byron:
It's great to hear the performance is good :) - it should never be slower than what Git does, particularly on Linux. If you have the WebKit repository, I recommend taking the new version out for a spin there as well. @pascalkuthe reported incredible values for the index-modification check last time we tried (to the point where the single-threaded dirwalk would now be the slowest part of the computation, taking about 2s on my machine). On that note, a progress indicator could also be implemented to show the index-modification check progress if things take longer, particularly on MacOS/Windows.

Submodule changes are definitely expected to be working, and you can try `gix status` to see them. If they wouldn't show up there, it could be they are disabled/inactive. With `gix submodule` it's easy to understand their state and if they are enabled or not.

If you find a discrepancy between `gix submodule` and `git submodule`, it's definitely something I should know about.

Please also note that submodules are only detected during `status()` if they are in the index, and present on disk with a worktree checkout, and if they are active, *and* the `submodule.*.ignore` variable isn't set to 'all'.

xJonathanLEI:
Hmm yeah submodules are definitely checked out, otherwise I wouldn't be able to attempt to test submodule changes.

Wasn't aware that `gix status` is now functional, as the `gitoxide` `README` seems to be outdated (the checkbox next to `status` is not checked). Just tested on the same repo that had the issue of submodule diffs not showing up:

```log
$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   lib/forge-std (modified content)

no changes added to commit (use "git add" and/or "git commit -a")
$ gix status
 18:51:21 status done 35.0 files in 0.00s (17.9k files/s)

head -> index isn't implemented yet
```

So it does look like `gix` isn't capturing the diff on it? Or am I missing anything? @Byron

Byron:
It's not noted in the `README` yet as it still isn't fully functional, even though it should be [after this PR lands](https://github.com/Byron/gitoxide/pull/1317). 

Can you show the changes in `lib/forge-std`, by running `git status` in there? I would think that these changes are `head -> index` ones that aren't implemented yet. Also, what does `gix submodule` show in the superproject, in case the submodule changes are captured by `gix status`?

xJonathanLEI:
Head -> index means staged changes right? I don't think that's the thing at play here, as my test changes are not staged. I just changed one line in the `README.md` file in a submodule and nothing else.

Added `gix submodule` at repo root and `git/gix status` in the submodule root:

```
~/submodule-test$ git status
On branch master
Your branch is up to date with 'origin/master'.

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
  (commit or discard the untracked or modified content in submodules)
        modified:   lib/forge-std (modified content)

no changes added to commit (use "git add" and/or "git commit -a")
~/submodule-test$ gix status
 20:40:49 status done 35.0 files in 0.00s (14.6k files/s)

head -> index isn't implemented yet
~/submodule-test$ gix submodule
 ⅹ lib/forge-std config:yes head:1d9650e index:1d9650e (v1.6.1) [https://github.com/foundry-rs/forge-std]
 ⅹ lib/openzeppelin-contracts-upgradeable config:yes head:3d4c0d57 index:3d4c0d57 (v4.9.3) [https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable]
 ⅹ lib/openzeppelin-contracts config:yes head:932fddf6 index:932fddf6 (v5.0.0) [https://github.com/OpenZeppelin/openzeppelin-contracts]
~/submodule-test$ cd lib/forge-std/
~/submodule-test/lib/forge-std$ git status
HEAD detached at 1d9650e
Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   README.md

no changes added to commit (use "git add" and/or "git commit -a")
~/submodule-test/lib/forge-std$ gix status
 20:40:57 status done 46.0 files in 0.00s (21.5k files/s)
  M README.md

head -> index isn't implemented yet
```

Apparently `gix` is able to capture the diff _when it's in the submodule_.

Byron:
Indeed, `gix` would see these changes in submodules, and I think I know what the problem is. All of these submodules are inactive, hence the `x` mark, and `gix status`, incorrectly, excludes these inactive submodules from the status.

If you try again with `gix -c submodule.active="*" status` they will show up.

The issue is already fixed and should be part of the next release in roughly a month. In the meantime, you would be able to use `gix::open_opts` and set up options with [config-overrides](https://docs.rs/gix/latest/gix/open/struct.Options.html#method.config_overrides), one key set to `submodule.active=*` should do the trick.

pascalkuthe:
@Byron once head->Index diff lands will it be possible to differentiate it from the worktree -> index diff?

With #9647 we can have multiple different coulumns/properties to filter by so then users could either filter all staged changes to check they are happy before making a commit or filter only unstaged changes to do something similar to git add -p in the editor (eventuallyonce we have a diff mode/staging in the editor).

Byron:
> @Byron once head->Index diff lands will it be possible to differentiate it from the worktree -> index diff?

Yes, I plan to make it runnable separately, as well. The full-blown status will then have both types of changes, worktree-index and head-index. In other words, I think it will be flexible enough and running these one after another or together will make no performance difference (except that if run together, `gix` will run both kinds of changes in parallel).

The way I see the new columns from the referenced issue, you'd probably just run the `status()` (as fast as possible), and get all information streamed in.

xJonathanLEI:
Personally I think we can just live with this bug for now? I mean it's something that will literally be fixed by dependabot lol. Also it only impacts a very specific use case.

pascalkuthe:
> Personally I think we can just live with this bug for now? I mean it's something that will literally be fixed by dependabot lol. Also it only impacts a very specific use case.

I would be fine without any submodule support at all this is not a blocker for me

pascalkuthe:
BTW. this PR should be calling gitoxide in a background thread and use an injector to stream in entrues like global search/the file picker do. To avoid freezing the ui on large repo (you also get a running indicator that way).

xJonathanLEI:
> BTW. this PR should be calling gitoxide in a background thread and use an injector to stream in entrues like global search/the file picker do. To avoid freezing the ui on large repo (you also get a running indicator that way).

Right. Thanks for the reminder. This PR itself predated the streaming thing I think, so it doesn't do that right now. I will make that change before pushing the new version.

Byron:
Sounds great, all sorted then! If the injector or streaming facilities also have progress support, you can integrate that too for index-modification checks which, if you have a machine like @pascalkuthe, will probably not ever show up even on the biggest repositories 😁. On my machine it would though, at least on WebKit. So it's probably not too important to have a progress bar for this.

xJonathanLEI:
Hmm it turns out to be harder than I thought. I converted the diff provider method to return an interator adapter (wraping `gix::status::index_worktree::Iter`) which is then consumed by the picker initialization code. All works well until I need to send the iter to a background thread when it's unable to finish by the deadline. Apparently `gix::status::index_worktree::Iter` is not `Send`.

Unless there's way to make that iter `Send`, we're left with these options:

1. Just freeze the UI lol
2. Clone the diff providers and build the iter in the background thread to begin with (no optimistic sync attempt) (might need to get rid of the `dyn` boxing of diff provider to allow cloning, maybe change to use a union enum?)
3. Do a sync attempt first, but if it doesn't make it, throw away the whole list and restart again in the background thread

Notably, since we turn on file rename detection (IMO wouldn't be very useful if this is off), it could take quite a while for the iterator to even return the first item on large repositories. This means that if we do any sync attempt, the first `.next()` call might well exceed the deadline, freezing the UI anyways.

Wdyt?

Byron:
Indeed, with rename tracking streaming is only happening for modified entries, but added and removed ones, depending on the setting, will have to wait. And it's correct that calling `next()` might block for a long time in the worst case.

`gix::Repository` is always per thread and contains some per-thread state. The only way around this is to turn it into [`into_sync()`](https://docs.rs/gix/latest/gix/struct.Repository.html#method.into_sync), move that to the thread, and let it initiate the iteration.

pascalkuthe:
Cloning the diff providers is free anyway

xJonathanLEI:
Noted on the diff providers. That's what I'm doing now.

> you also get a running indicator that way

@pascalkuthe It looks like the `(running)` indicator is just for the matcher? So unless there are entries that stream in endlessly I won't get it, right? In the case of the file picker this is fine as entries _do_ constantly stream in, but in this case most of the time is spent on searching with few expected output entries.

So the only benefit, with the current indicator impl, of moving it to the background is that UI won't freeze. Users have no way to tell whether there's no diff or the searcher is running.

pascalkuthe:
K

> Noted on the diff providers. That's what I'm doing now.
> 
> > you also get a running indicator that way
> 
> @pascalkuthe It looks like the `(running)` indicator is just for the matcher? So unless there are entries that stream in endlessly I won't get it, right? In the case of the file picker this is fine as entries _do_ constantly stream in, but in this case most of the time is spent on searching with few expected output entries.
> 
> So the only benefit, with the current indicator impl, of moving it to the background is that UI won't freeze. Users have no way to tell whether there's no diff or the searcher is running.
> 
> (of course, given how fast `gitoxide` is, it's very unlikely that users will actually see this case at all)

That will be fixed in #9647

xJonathanLEI:
Btw after playing with it a bit more I will completely leave out submodule support from this PR. Turns out it's a bit more work than expected. It also helps keep this one minimal. We can always add that in once the upstream bug is resolved anyways.

xJonathanLEI:
(Somehow the macOS run is just "cancelled" giving me red CI. Re-pushing just to retry CI. Sorry for spamming guys.)

xJonathanLEI:
This PR is now ready for review again. Finally!

Thanks to the new `status` feature of `gix`, this PR has been much simplified from the original submission, now containing only the UI part and the thin glue code that handles `gix` integration.

For those who have been using this PR, nothing has really changed from a user's perspective other than better performance and better Git feature support (e.g. attributes). If you liked it before, you'll love it now :)

Can't wait for this to be reviewed & merged, and Helix users to finally be able to use this (arguably expected) feature.

helix-vcs/src/git.rs
```diff
@@ -61,6 +74,28 @@ impl Git {
...
+            .to_path_buf();
+
+        // Here we discard the `status.showUntrackedFiles` config, as it makes little sense in our
+        // case to not list new (untracked) files. We could have respected this config if the
+        // default value weren't `Collapsed` though, as this default value would render the feature
+        // unusable to many.
+        let status_platform = repo
+            .status(gix::progress::Discard)?
+            .untracked_files(UntrackedFiles::Files)
+            .index_worktree_rewrites(Some(Default::default()));
```
Byron, line 92: It's probably better to bring in the `Rewrite` structure in order to make clear what's going on. Even I don't know actually 😅 - my guess is that it's on with identity based rename tracking.

helix-vcs/src/git.rs
```diff
@@ -61,6 +74,28 @@ impl Git {
...
+
+        Ok(Box::new(StatusIter {
+            work_dir,
+            inner: status_platform.into_index_worktree_iter(Default::default())?,
```
Byron, line 96: I don't like the `Default` there as it's not clear what it stands for, but also think that `vec![]` or similar would be much more descriptive.
That's probably an area where the API can improve, even though I don't know how without making it more complicated.

With that said, maybe a variable like `let empty_pathspec = vec![]` would do a slightly better job here (even though one would still have to know that this means there is no restriction.

helix-vcs/src/git.rs
```diff
@@ -112,6 +147,69 @@ impl DiffProvider for Git {
...
+                            Change::Modification { .. } => {
+                                Some(FileChange::Modified { path: full_path })
+                            }
+                            // No submodule support for now
+                            Change::Type | Change::SubmoduleModification(_) => None,
+                        },
+                        EntryStatus::NeedsUpdate(_) | EntryStatus::IntentToAdd => None,
+                    }
+                }
+                Item::DirectoryContents { entry, .. } => Some(FileChange::Untracked {
```
Byron, line 192: `entry.status` is something you'd definitely want to check and only use `Untracked` entries, for robustness and clarity.

Byron (commented):
Edit: Great work, and I am looking forward to seeing this released :)!

A few small comments from me.

helix-vcs/src/git.rs
xJonathanLEI: Not 100% relevant, but I remember checking this part of the source and wondering why a generic `IntoIterator` param isn't used. I mean this method immediately calls `index_worktree_status`, which indeed takes a generic item impl. Here we're forced to do an allocation even though we're just passing in an empty vec. Negligible though.

helix-vcs/src/git.rs
xJonathanLEI: Now explicitly constructing the `Rewrites` struct.

helix-vcs/src/git.rs
xJonathanLEI: Now explicitly naming the argument with comment.

helix-vcs/src/git.rs
xJonathanLEI: Thanks! Totally missed this. Fixed now.

helix-vcs/src/git.rs
Byron: Thanks for the hint! I think I did it to allow passing empty items easily, but now it's `IntoIterator<Item = BString>` which at least allows to pass `None` as well, a little easier.

helix-term/src/ui/mod.rs
```diff
@@ -255,6 +256,54 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
...
+    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
+    let injector = picker.injector();
+
+    let diff_providers = editor.diff_providers.clone();
+
+    std::thread::spawn(move || {
+        // There's no way we can notify the editor since we're in a background thread. This just
+        // silently fails.
+        if let Ok(change_iter) = diff_providers.get_changed_files(&cwd) {
+            // We ignore errors as they could be caused by permission issues etc
```
pascalkuthe, line 296: You can use `report_blocking` to send an error to the main loop for display. I do Thi k reporting the error would be useful.

helix-term/src/ui/mod.rs
```diff
@@ -255,6 +256,54 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
...
+                cx.editor.set_error(err);
+            }
+        },
+    )
+    .with_preview(|_editor, meta| Some((meta.path().to_path_buf().into(), None)));
+    let injector = picker.injector();
+
+    let diff_providers = editor.diff_providers.clone();
+
+    std::thread::spawn(move || {
```
pascalkuthe, line 292: Use `tokio::spawn_blocking` here. It's usually faster since it's a threadpool/reuses threads and it will make sure we have a runtime context

helix-vcs/src/git.rs
```diff
@@ -61,6 +76,36 @@ impl Git {
 
         Ok(res)
     }
+
+    /// Emulates the result of running `git status` from the command line.
+    fn status(repo: &Repository) -> Result<Box<dyn Iterator<Item = Result<FileChange>>>> {
```
pascalkuthe, line 81: For abstracting over iteration like this I think external iteration instead of internal iteration is easier.

Instead of returning a dynamic dispatch iter here, you should pass a closure to this function. Than this function spawns the thread (with tokio), drives the iterator and calls the provided closure.

This also removes the need to implememt iterator which is always slightly messy/slow.

helix-term/src/commands.rs
```diff
@@ -2994,6 +2995,46 @@ fn jumplist_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
+fn changed_file_picker(cx: &mut Context) {
+    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
+
+    let entries = match cx.editor.diff_providers.get_changed_files(&cwd) {
```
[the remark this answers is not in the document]
xJonathanLEI: Hmm this is an outdated comment I guess? It commented on the outdated diff (also time is 2 days ago).

helix-term/src/ui/mod.rs
xJonathanLEI: I didn't know this exists. I will update the code to surface both initialization error and iteration errors.

helix-term/src/ui/mod.rs
xJonathanLEI: Oh okay I was mimicking the `file_picker` function above actually. Lemme change this.

helix-vcs/src/git.rs
xJonathanLEI: You're right. I was originally planning to mimic the optimistic sync trial pattern from `file_picker`, which I believe necessitate an iterator. Now that we don't do it we can use a closure instead. Will fix. Thanks!

xJonathanLEI:
Found a small issue while working on surfacing iteration errors from the background thread. With `report_blocking` it's quite easy to implement. However, I was unable to actually test it.

I wanted to simulate a situation where there's a file/folder in the repository with permission issues. Ideally Helix would report errors on these and move on, like `git status` does. However, to much of my surprise, it turns out that I couldn't make it to the iteration in the first place. As long as there's something in the repo which I don't have permission to access, _the call to `into_index_worktree_iter` would already fail_, before I even manage to grab an iterator.

What surprises me even more is that `gix` (also on `v0.61.0`) is able to succeed on the same function call. Unlike `git`, it _does_ early terminate as soon as it encounters an error, but only _after_ it gets an iterator from it.

I tried a few quick hacks to make the code for opening repos identical on Helix vs `gix`. Yet they still behave differently. I could spend more time digging deeper, but considering this is quite an edge case, and the lib author is literally in this convo, I decided to pause and flag the issue here to see if I can get some quick pointers :) @Byron

helix-term/src/ui/mod.rs
pascalkuthe: Yeah i want to change that one over too. It's not a huge deal but slightly nice.

helix-term/src/commands.rs
```diff
@@ -2994,6 +2995,11 @@ fn jumplist_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
+fn changed_file_picker(cx: &mut Context) {
+    let picker = ui::changed_file_picker(cx.editor);
```
the-mikedavis, line 2999: The file picker is only defined in the `ui` mod because it's reused in a few places across modules. Let's define this picker here within the command like the other pickers

helix-term/src/ui/menu.rs
```diff
@@ -52,6 +54,40 @@ impl Item for PathBuf {
...
+            Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
+            Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
+            Self::Renamed { from_path, to_path } => (
+                "[>]",
+                data.style_modified,
+                format!("{} -> {}", process_path(from_path), process_path(to_path)),
+            ),
+        };
+
+        Row::new(vec![sign.to_owned(), content]).style(style)
```
the-mikedavis, line 87: Styling the entire line seems like too much color to me. Could we color only the sign instead and leave the content as the default text color?

xJonathanLEI:
@pascalkuthe Just pushed `138846fc` that should resolve all your comments. Optimistically marking them as resolved.

helix-term/src/ui/menu.rs
xJonathanLEI: For comparison, the current implementation:

image

Changing to only style the sign:

image

Personally I find the whole line highlight to be easier to work with. I can identify the types quicker. The original idea was to mimic the `git status` command where it's always the case that the whole line, including the change type letter(s) and path, is highlighted.

But as always definitely open to changes. Please let me know if I should make the change.

poliorcetics: I prefer whole line colors too

the-mikedavis: I strongly prefer just the sign. @pascalkuthe @archseer thoughts?

poliorcetics:
Note: this doesn't resolve https://github.com/helix-editor/helix/issues/5362 fully, it should not be marked as closing it, it's missing a current file change picker (not saying you should do it in this MR!)

Byron:
> What surprises me even more is that `gix` (also on `v0.61.0`) is able to succeed on the same function call. Unlike `git`, it _does_ early terminate as soon as it encounters an error, but only _after_ it gets an iterator from it.
> 
> I tried a few quick hacks to make the code for opening repos identical on Helix vs `gix`. Yet they still behave differently. I could spend more time digging deeper, but considering this is quite an edge case, and the lib author is literally in this convo, I decided to pause and flag the issue here to see if I can get some quick pointers :) @Byron

Thanks, this looks like a shortcoming of the iterator which you are using. `gix status`, which I assume you are comparing it too, doesn't use the iterator, but the non-iterator version of the same functionality. Do you have a script to create a repository with permission issues that cause the failure?  I could probably use it to setup a test-case and make the iterator more conforming, compared to `git`.

xJonathanLEI:
> doesn't use the iterator, but the non-iterator version of the same functionality

Wait are you sure? The `main` function for `gix` calls this function:

https://github.com/Byron/gitoxide/blob/9e9c653a83df58f8cdfe3a7adb2d824c8a368e72/gitoxide-core/src/repository/status.rs#L43-L182

which does use the iterator. In fact it calls the same `into_index_worktree_iter` function we're using.

Byron:
Oh, right, I was remembering an older version then, it went through a few iterations 😁.
That’s even more puzzling then, as there is definitely no global magic going on. But that should be good, as it would mean that gix can do the desired thing already (if I understood correctly).

xJonathanLEI:
> gix can do the desired thing already

Yeah that's why I meant. While it is true that currently `gix status` early terminates as soon as an error is encountered, it doesn't have to. It's a simple fix to just gracefully handle the `Err` from iteration. I couldn't make it work from Helix for some reasons: `into_index_worktree_iter` would already return `Err`. I'd say it's quite an edge case though, as it only impacts repos with permission issues. So maybe we don't need to deal with it in this PR.

Definitely possible to narrow down the issue. Was raising there just in case you know what's going on. But all good otherwise.

helix-term/src/ui/menu.rs
pascalkuthe: For the default theme I also prefer only coloring the sign but it may be ok in another theme.

We could just use two different theme keys here for the sign and the rest of the text so it's customizable (altough I would definitely only theme the sign in the default/"offical" themes).

xJonathanLEI: Okay lemme change to only style the signs unless @archseer thinks otherwise.

helix-vcs/src/diff.rs
```diff
@@ -290,3 +291,30 @@ impl Diff<'_> {
         }
     }
 }
+
+pub enum FileChange {
```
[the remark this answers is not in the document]
xJonathanLEI: Moved.

helix-vcs/src/git.rs
```diff
@@ -61,6 +70,94 @@ impl Git {
...
+        let mapper = |item: std::result::Result<Item, gix::status::index_worktree::Error>| {
+            Result::<Option<FileChange>>::Ok(match item? {
+                Item::Modification {
+                    rela_path, status, ..
+                } => {
+                    let full_path = work_dir.join(rela_path.to_path()?);
+
+                    match status {
+                        EntryStatus::Conflict(_) => {
+                            // We treat conflict as change for now for simplicity
```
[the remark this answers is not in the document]
xJonathanLEI: Added handling for conflicts. I'm using `[x]` as the sign. But for the color, since we're piggybacking on the `diff` theme colors for this we don't have many choices, so I settled on the deletion color. Looks like this now (the `README` row):

image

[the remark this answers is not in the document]
pascalkuthe: Great! For Thedinghausen I think we can do the following. Use a new theme key `diff.delta.conflict`. This will automatically fakkbaxn to `diff.delta` if undefined and still allows themes to set a different color.

While deletions fit somewhat colorwise reusing them feels like we are misusing the existing key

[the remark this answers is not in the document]
xJonathanLEI: Just pushed a commit that changes to use `diff.delta.conflict` for conflicts and `diff.delta.moved` for renames. It's weird tho that `diff.delta.moved` was documented but doesn't seem to be used anywhere.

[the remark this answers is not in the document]
xJonathanLEI: Do you also want me to define the `diff.delta.conflict` key for the default theme to be the same as the deletion color? Or just leave it alone?

helix-term/src/ui/menu.rs
```diff
@@ -52,6 +55,45 @@ impl Item for PathBuf {
 
 pub type MenuCallback<T> = Box<dyn Fn(&mut Editor, Option<&T>, MenuEvent)>;
 
+pub struct FileChangeData {
```
the-mikedavis, line 58: Let's move this into the command definition as well, similar to how buffer_picker and jumplist_picker define their Item types within the command. #9647 will eliminate the dependency of the Picker on menu::Item so this type doesn't necessarily belong here, and I think this type will be local to the changed file picker anyways. (For an in-document picker of changes we will probably not need the untracked style for example.)

xJonathanLEI: Fixed with `7fd75b50`.

helix-term/src/commands.rs
```diff
@@ -2994,6 +2999,89 @@ fn jumplist_picker(cx: &mut Context) {
...
+                Self::Conflict { path } => ("[x]", data.style_conflict, process_path(path)),
+                Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
+                Self::Renamed { from_path, to_path } => (
+                    "[>]",
+                    data.style_renamed,
+                    format!("{} -> {}", process_path(from_path), process_path(to_path)),
+                ),
+            };
+
+            Row::new(vec![
```
the-mikedavis, line 3035:
```suggestion
            Row::new([
```

We use `vec![]` in most places for `Row::new` but it takes an `IntoIterator` so we can just use a slice instead

helix-term/src/commands.rs
```diff
@@ -2994,6 +2999,89 @@ fn jumplist_picker(cx: &mut Context) {
...
+                Cell::from(Span::styled(sign, style)),
+                Cell::from(content),
+            ])
+        }
+    }
+
+    let cwd = std::env::current_dir().unwrap_or_else(|_| PathBuf::from("./"));
```
the-mikedavis, line 3042: Usages of `std::env::current_dir()` were replaced a little while ago with a wrapper around a `RwLock` of a path (to avoid panics when the cwd is deleted). This should switch to `helix_stdx::env::current_working_dir` to reuse that value

xJonathanLEI: Ah got it. Fixed in `661d772d`.

pascalkuthe (commented):
I found a couple minor things that can be improved with regarrd readability while doing one last pas.

It took me a bit to figure out what exactly I wanted so I already made the changes locally and pushed them as seperate commits. Hope you don't mind. I think with these changes this looks good to go

pascalkuthe (approved):

helix-term/src/commands.rs
```diff
@@ -2994,6 +3000,94 @@ fn jumplist_picker(cx: &mut Context) {
...
+                Self::Untracked { path } => ("[+]", data.style_untracked, process_path(path)),
+                Self::Modified { path } => ("[~]", data.style_modified, process_path(path)),
+                Self::Conflict { path } => ("[x]", data.style_conflict, process_path(path)),
+                Self::Deleted { path } => ("[-]", data.style_deleted, process_path(path)),
+                Self::Renamed { from_path, to_path } => (
+                    "[>]",
+                    data.style_renamed,
+                    format!("{} -> {}", process_path(from_path), process_path(to_path)),
+                ),
+            };
```
the-mikedavis, line 3034: For #9647 we will want to change the UI slightly. Like the symbol pickers instead of adding icons / glyphs for each kind of change I prefer words, so the symbols don't need a legend. I have this merged into my driver branch with a UI like this:

foo3

so that the symbols stay but are accompanied by a `git status`-like description of the change. But we can hash this out on #9647 once I rebase rather than here

the-mikedavis (approved):
Nice work, I'm excited to use this!

I edited the description to remove the closing keywords for #5362. I tihnk it's worth exploring/evaluating the other pickers in that issue (I have one locally in https://github.com/the-mikedavis/helix/commit/9fd4fcfddf378aee1ebc0b4063b904c4aeb18748 that I'm trying out now) so I don't want to close it too soon

mehmetkursataydin:
Wow, I love this!

PR #5649, opened 2023-01-23
Support sagemath language

Eloitor:
Hi this is my first contribution to this project. I copied the structure from: https://github.com/helix-editor/helix/pull/2903

The goal of this PR is to add support for the [sagemath](https://www.sagemath.org/) language. 
- The syntax is based on python.
- I didn't succeed at running the python language server by default with `.sage` files.
- It would be nice if  the syntax highlighting supported sage's docstrings:  In a comment, the lines starting with `sage: ` contain sage code. It would be nice if its syntax were highlighted. For an example of a docstring, see: https://doc.sagemath.org/html/en/developer/coding_basics.html#template


the-mikedavis (approved):

the-mikedavis:
Could you run `cargo xtask docgen` and commit the results? That will add sage to the supported language documentation

Issue #4207, opened 2022-10-11
Global search should search opened documents


pathwave:
Currently global search only read files from disk, this creates two problems:
Unsaved content is not search-able and the resulting search result can end up on wrong line (and selection) because of unsaved content in-between. I think it would be better if it searched throught opened documents and excluded those opened documents from the disk search.

pppKin:
Good call. I think this would be easier to implement after (the much anticipated) #3110 is merged.

pathwave:
I implemented it for #4381, so it just needs a copy/paste ish after im done with #4381.

PR #5652, opened 2023-01-23
search buffer contents during global search

pascalkuthe:
Followup to #5639, addresses https://github.com/helix-editor/helix/issues/5604#issuecomment-1399674583

Currently, helix always queries the FS when performing global search.
However, when previewing and opening a file helix automatically uses the buffer if it's already open.
That means that the search would report incorrect results for unsaved buffers (which lead to crashes prior to #5639).

This PR creates am implementation of `std::io::Read` for `Rope`, so we can easily reuse the existing API in `grep_searcher`.

This change has the nice side effect of avoiding IO for already open buffers for global search. The performance difference will rarely matter in practice, but it's a nice boost when a huge file is already open in helix.
A similar mechanism can be used for #5645 so that files that are changed but unsaved also get reported as changed.

### DIFF
--- a/helix-core/src/rope_reader.rs
+++ b/helix-core/src/rope_reader.rs
@@ -1,0 +1,37 @@
+use std::io;
+
+use ropey::iter::Chunks;
+use ropey::RopeSlice;
+
+pub struct RopeReader<'a> {
+    current_chunk: &'a [u8],
+    chunks: Chunks<'a>,
+}
+
+impl<'a> RopeReader<'a> {
+    pub fn new(rope: RopeSlice<'a>) -> RopeReader<'a> {
+        RopeReader {
+            current_chunk: &[],
+            chunks: rope.chunks(),
+        }
+    }
+}
+
+impl io::Read for RopeReader<'_> {
+    fn read(&mut self, mut buf: &mut [u8]) -> io::Result<usize> {
+        let buf_len = buf.len();
+        loop {
+            let read_bytes = self.current_chunk.read(buf)?;
+            buf = &mut buf[read_bytes..];
+            if buf.is_empty() {
+                return Ok(buf_len);
+            }
+
+            if let Some(next_chunk) = self.chunks.next() {
+                self.current_chunk = next_chunk.as_bytes();
+            } else {
+                return Ok(buf_len - buf.len());
+            }
+        }
+    }
+}
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -29,7 +29,7 @@ use helix_core::{
     tree_sitter::Node,
     unicode::width::UnicodeWidthChar,
     visual_offset_from_block, Deletion, LineEnding, Position, Range, Rope, RopeGraphemes,
-    RopeSlice, Selection, SmallVec, Tendril, Transaction,
+    RopeReader, RopeSlice, Selection, SmallVec, Tendril, Transaction,
 };
 use helix_view::{
     clipboard::ClipboardType,
@@ -2062,10 +2062,15 @@ fn global_search(cx: &mut Context) {
                 .map(|comp| (0.., std::borrow::Cow::Owned(comp.clone())))
                 .collect()
         },
-        move |_editor, regex, event| {
+        move |editor, regex, event| {
             if event != PromptEvent::Validate {
                 return;
             }
+
+            let documents: Vec<_> = editor
+                .documents()
+                .map(|doc| (doc.path(), doc.text()))
+                .collect();
 
             if let Ok(matcher) = RegexMatcherBuilder::new()
                 .case_smart(smart_case)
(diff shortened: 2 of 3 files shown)


pascalkuthe:
Closes #4207

helix-view/src/editor.rs
```diff
@@ -1421,6 +1441,12 @@ impl Editor {
         self.documents.values_mut()
     }
 
+    /// A thread save iterator over the text of all documents
```
the-mikedavis, line 1444:
```suggestion
    /// A thread-safe iterator over the text of all documents
```

helix-view/src/editor.rs
```diff
@@ -757,6 +757,26 @@ pub struct Breakpoint {
 
 use futures_util::stream::{Flatten, Once};
 
+/// Threadsave (Sync + Send) iteration over document text (and path)
```
the-mikedavis, line 760:
```suggestion
/// Thread-safe (Sync + Send) iteration over document text (and path)
```

helix-term/src/commands.rs
```diff
@@ -1943,17 +1946,36 @@ fn global_search(cx: &mut Context) {
...
+                                Ok(true)
+                            });
+                            let doc = documents.clone().find(|(doc_path, _)| {
+                                doc_path.map_or(false, |doc_path| doc_path == entry.path())
+                            });
+
+                            let result = if let Some((_, doc)) = doc {
+                                // there is already a buffer for this file
+                                // search the buffer instead of the file because it's faster
+                                // and captures new edits without requireing a save
```
the-mikedavis, line 1963:
```suggestion
                                // and captures new edits without requiring a save
```

helix-term/src/commands.rs
```diff
@@ -1943,17 +1946,36 @@ fn global_search(cx: &mut Context) {
...
+                                if searcher.multi_line_with_matcher(&matcher) {
+                                    // in this case a continous buffer is required
```
the-mikedavis, line 1965:
```suggestion
                                    // in this case a continuous buffer is required
```

the-mikedavis (commented):
I'm not familiar enough with Rust parallelism to review most of this but I think this looks good. I left some small typo fix comments

helix-core/src/rope_reader.rs
```diff
@@ -0,0 +1,37 @@
...
+                return Ok(read_bytes);
+            }
+
+            if let Some(next_chunk) = self.chunks.next() {
+                self.current_chunk = next_chunk.as_bytes();
+            } else {
+                return Ok(read_bytes);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
pascalkuthe: While you are right that the code was incorrect the reason and the fix you provided were also wrong. I simply forgot to offset the buffer appropriately.  I did also simplfy the code a bit

helix-view/src/editor.rs
```diff
@@ -757,6 +757,26 @@ pub struct Breakpoint {
...
+/// This is necessary because the editor and the document contain
+/// interior mutability which makes it impossible to send them across threads.
+/// However the document text itself (ropes) is perfectly fine to send across
+/// threads
```
LeoniePhiline, line 764:
```suggestion
/// threads.
```

helix-term/src/commands.rs
LeoniePhiline, line 1963:
```suggestion
                                // There is already a buffer for this file.
                                // Search the buffer instead of the file because it's faster,
                                // and capture new edits without requiring a save.
```

zyklotomic:
Slightly unrelated to the PR, but wanted to check if my understanding of the UnboundedReceiverStream is correct.

In the asynchronous callback `show_picker`, do we have to worry about the `UnboundedReceiverStream` not collecting any values (because all the `all_matches_sx` from earlier have been dropped)? In WalkerBuilder.run(), inside the Box, in the definition of `sink`, there is a reference to all_matches_sx, so it won't be dropped. Is this why we don't have to worry about the UnboundedReceiverStream closing early?

helix-view/src/editor.rs
```diff
@@ -799,6 +799,26 @@ pub struct Breakpoint {
...
+unsafe impl Send for SharedDocuments<'_> {}
+
+impl<'a> Iterator for SharedDocuments<'a> {
+    type Item = (Option<&'a PathBuf>, &'a Rope);
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let doc = self.0.next()?;
+        Some((doc.path(), doc.text()))
+    }
+}
```
archseer, line 820: Could we avoid this custom type by `documents.values().map(..).collect()`, then returning the Vec? It would mean allocating an intermediary vec but it's unlikely the user will have enough documents open for it to make a difference

pascalkuthe: you are totally right. I think back then I sort of started out trying that but with the full document because that doesn't work thanks to the cell. Then I implemented this iterator (with manual sync/send) and later only shared path and rope to make the unsave code sound but somehow missed the obviour thing that I could have just collect the rope/path pairs to begin with. 

I have removed the custom type, sorry about that!

archseer (approved):

PR #5658, opened 2023-01-23
Hide duplicate symlinks from the picker

### LINKED ISSUE #5210: File picker freezes trying to read named pipes
### Summary

If the file picker tries to preview a named pipe then helix will freeze until you write data to the pipe (most of the time 2 writes are needed).

### Reproduction Steps

 

I tried this:

1. `mkfifo mypipe`
2. `hx`
3. `<space>fmypipe`

I expected this to happen:

Helix detects that the file is a named pipe and doesn't attempt to show a preview (at least not in a blocking way).

Instead, this happened:

Once the fuzzy search lands on the pipe, helix will freeze waiting for data to exist in the pipe.

### Helix log

```
The log was blank.
```

</details>

### Platform

Linux

### Terminal Emulator

foot

### Helix Version

helix 22.12

### LINKED ISSUE #5395: Add an option to ignore symbolic links when fuzzy finding
I haven't thought through what exactly needs to change, but it would be nice to add a configuration option to let the fuzzy finder ignore symbolic links.

This will be useful when the user is working in a large codebase and has a lot of symbolic links for whatever reason. For example, I was working on a codebase with a lot of symlinks generated during compilation, and helix was showing 900,000 files including symbolic links whereas there were only(?) 350,000 real files in the directory.

Since helix fuzzy finder is currently synchronous, being able to ignore symlinks will also provide a nice side effect of snappier results.

### LINKED ISSUE #5725: freeze when opening /tmp directory
### Summary

when open /tmp (or /tmp/) from command line, hx will open a directory file picker as expected but will become unresponsive to any input, preventing you from interacting with the UI, selecting a file, quitting, etc. requiring the shell instance to be terminated. 

I was thinking this might because /tmp is a symlink or dir permissions but I checked with other symlinks and root owned directories and they seem to behave as expected

### Reproduction Steps

$ hx /tmp 

file picker opens

Expected behaviour:

- hx continues operating as expected; i.e. selecting a file, opening in a buffer

Actual Behaviour:

- hx becomes unresponsive to user input, requiring the shell instance to be terminated/closed

### Helix log

I cleared the log before reproducing (multiple times) to cut back on un related log messages from days/weeks prior. The log is empty after reproducing this event.

### Platform

macOS

### Terminal Emulator

wezterm 20221119-145034-49b9839f

### Helix Version

helix 22.12 (96ff64a8)

pascalkuthe:
Fixes #5395
Fixes  #5725
Fixes #5210

This PR adds a config option to the picker to hide symlinks that point into the root directory of a picker.
These symlinks always point at files that would be included in the picker anyway and are therefore essentially duplicate.
The config option is enabled by default.

While working on this I noticed some small issues with the existing handling around symlinks.
They were included in the file picker if `follow-links` was disabled (which seems really odd, `global_search` already handled this correctly). Also added documentation for `follow-links` since that seems to have been missed.

Edit: The fix for filtering symlinks from the picker I described in the last paragraph more generally filters all kinds of "non standard" files like pipes, block devices, etc. Reading thise can lock up the edtior. By fixing that I actually accidentally fixed #5725

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -150,6 +150,8 @@ All git related options are only enabled in a git repository.
 | Key | Description | Default |
 |--|--|---------|
 |`hidden` | Enables ignoring hidden files. | true
+|`follow-links` | Follow symlinks instead of ignoring them | true
+|`deduplicate-links` | Ignore symlinks that point at files already shown in the picker | true
 |`parents` | Enables reading ignore files from parent directories. | true
 |`ignore` | Enables reading `.ignore` files. | true
 |`git-ignore` | Enables reading `.gitignore` files. | true
--- a/helix-term/src/lib.rs
+++ b/helix-term/src/lib.rs
@@ -27,18 +27,17 @@ fn true_color() -> bool {
 }
 
 /// Function used for filtering dir entries in the various file pickers.
-///
-/// We always want to ignore the .git directory, otherwise if
-/// `ignore` is turned off, we end up with a lot of noise
-/// in our picker.
-/// We also ignore symlinks that point inside the current directory
-/// if `dedup_links` is enabled.
-fn filter_entry(entry: &DirEntry, root: &Path, dedup_links: bool) -> bool {
-    if entry.file_name() != ".git" {
+fn filter_picker_entry(entry: &DirEntry, root: &Path, dedup_symlinks: bool) -> bool {
+    // We always want to ignore the .git directory, otherwise if
+    // `ignore` is turned off, we end up with a lot of noise
+    // in our picker.
+    if entry.file_name() == ".git" {
         return false;
     }
 
-    if dedup_links && entry.path_is_symlink() {
+    // We also ignore symlinks that point inside the current directory
+    // if `dedup_links` is enabled.
+    if dedup_symlinks && entry.path_is_symlink() {
         return entry
             .path()
             .canonicalize()
--- a/helix-term/src/lib.rs
+++ b/helix-term/src/lib.rs
@@ -10,6 +10,9 @@ pub mod health;
 pub mod job;
 pub mod keymap;
 pub mod ui;
+use std::path::Path;
+
+use ignore::DirEntry;
 pub use keymap::macros::*;
 
 #[cfg(not(windows))]
@@ -20,5 +23,28 @@ fn true_color() -> bool {
 }
 #[cfg(windows)]
 fn true_color() -> bool {
+    true
+}
+
+/// Function used for filtering dir entries in the various file pickers.
+///
+/// We always want to ignore the .git directory, otherwise if
+/// `ignore` is turned off, we end up with a lot of noise
+/// in our picker.
+/// We also ignore symlinks that point inside the current directory
+/// if `dedup_links` is enabled.
+fn filter_entry(entry: &DirEntry, root: &Path, dedup_links: bool) -> bool {
+    if entry.file_name() != ".git" {
+        return false;
+    }
+
+    if dedup_links && entry.path_is_symlink() {
+        return entry
+            .path()
+            .canonicalize()
+            .ok()
+            .map_or(false, |path| !path.starts_with(&root));
+    }
+
     true
 }
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -197,9 +197,10 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
         let entry = entry.ok()?;
         // This is faster than entry.path().is_dir() since it uses cached fs::Metadata fetched by ignore/walkdir
         if entry.file_type()?.is_file() {
-            return Some(entry.into_path());
+            Some(entry.into_path())
+        } else {
+            None
         }
-        None
     });
 
     // Cap the number of files if we aren't in a git project, preventing
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -14,6 +14,7 @@ mod statusline;
 mod text;
 
 use crate::compositor::{Component, Compositor};
+use crate::filter_entry;
 use crate::job::{self, Callback};
 pub use completion::Completion;
 pub use editor::EditorView;
@@ -162,6 +163,9 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
 
     let now = Instant::now();
 
+    let deadup_symlinks = config.file_picker.deduplicate_links;
+    let absolute_root = root.canonicalize().unwrap_or_else(|_| root.clone());
+
     let mut walk_builder = WalkBuilder::new(&root);
     walk_builder
         .hidden(config.file_picker.hidden)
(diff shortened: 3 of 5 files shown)


book/src/configuration.md
```diff
@@ -150,6 +150,8 @@ All git related options are only enabled in a git repository.
 | Key | Description | Default |
 |--|--|---------|
 |`hidden` | Enables ignoring hidden files. | true
+|`follow-links` | Follow symlinks instead of ignoring them | true
+|`deadup-links` | Ignore symlinks that point at files already shown in the picker | true
```
the-mikedavis, line 154: I think this should be "dedup", no? I suppose it could be "deduplicate-links" to be verbose

pascalkuthe: Sorry my spelling isn't always the best :/ I like the sound of `deduplicate-links` :+1:  although I kept `dedup` in the code. That abbreviation is common enough in code I think (rust stdlib uses it)

helix-term/src/ui/mod.rs
```diff
@@ -162,6 +163,9 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
 
     let now = Instant::now();
 
+    let deadup_symlinks = config.file_picker.deduplicate_links;
```
g-re-g, line 166:
```suggestion
    let dedup_symlinks = config.file_picker.deduplicate_links;
```

helix-term/src/ui/mod.rs
```diff
@@ -172,10 +176,7 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
         .git_global(config.file_picker.git_global)
         .git_exclude(config.file_picker.git_exclude)
         .max_depth(config.file_picker.max_depth)
-        // We always want to ignore the .git directory, otherwise if
-        // `ignore` is turned off above, we end up with a lot of noise
-        // in our picker.
-        .filter_entry(|entry| entry.file_name() != ".git");
+        .filter_entry(move |entry| filter_entry(entry, &absolute_root, deadup_symlinks));
```
g-re-g, line 179:
```suggestion
        .filter_entry(move |entry| filter_entry(entry, &absolute_root, dedup_symlinks));
```

helix-view/src/editor.rs
```diff
@@ -171,6 +171,8 @@ pub struct FilePickerConfig {
     /// Enables following symlinks.
     /// Whether to follow symbolic links in file picker and file or directory completions. Defaults to true.
     pub follow_symlinks: bool,
+    /// Hides symlinks that point into the current directory. Default to true
```
g-re-g, line 174:
```suggestion
    /// Hides symlinks that point into the current directory. Defaults to true.
```

helix-term/src/ui/mod.rs
```diff
@@ -194,15 +195,11 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
...
-        let is_dir = entry.file_type().map_or(false, |ft| ft.is_dir());
-        if is_dir {
-            // Will give a false positive if metadata cannot be read (eg. permission error)
-            None
-        } else {
-            Some(entry.into_path())
+        if !entry.file_type()?.is_file() {
+            return None;
         }
+        Some(entry.into_path())
```
g-re-g, line 202: Might be easier to read as a positively expressed if
```suggestion
        if entry.file_type()?.is_file() {
            return Some(entry.into_path());
        }
        None
```

helix-term/src/lib.rs
```diff
@@ -22,3 +25,26 @@ fn true_color() -> bool {
...
+    if dedup_links && entry.path_is_symlink() {
+        return entry
+            .path()
+            .canonicalize()
+            .ok()
+            .map_or(false, |path| !path.starts_with(&root));
+    }
+
+    true
+}
```
g-re-g, line 50: I believe the logic in here around filtering the `.git` folder is backwards. https://docs.rs/ignore/0.4.20/ignore/struct.WalkBuilder.html#method.filter_entry Might want to put the comments closer to their usage in the code. Might give the function a more expressive name.
```suggestion
/// Function used for filtering dir entries in the various file pickers.
fn ignore_git_and_simlinks(entry: &DirEntry, root: &Path, dedup_links: bool) -> bool {
    // We always want to ignore the .git directory, otherwise if
    // `ignore` is turned off, we end up with a lot of noise
    // in our picker.
    if entry.file_name() == ".git" {
        return false;
    }

    // We also ignore symlinks that point inside the current directory
    // if `dedup_links` is enabled.
    if dedup_links && entry.path_is_symlink() {
        return entry
            .path()
            .canonicalize()
            .ok()
            .map_or(false, |path| !path.starts_with(&root));
    }

    true
}

```

g-re-g (commented):
imo, everywhere `links` is used by itself `symlinks` should be used. Personally I'm not one for abbreviation and in anycase both `link` and `symlink` are used in this PR and it should probably stick to one of the two.

helix-term/src/lib.rs
pascalkuthe: Thanks for pointing out the `.git` mistake to me. I refactored this after I was done and messed up the `&&` to `if` conversion. I should probably stop writing code at 3am. This function is used both in `commands.rs` and `ui/mod.rs` so this is as close to the usage as it gets.

pascalkuthe:
> Solid feature. Duplicate entries due to symlinks is something I've run into as well. imo, everywhere `links` is used `symlinks` should be used. Personally I'm not one for abbreviation and in anycase both `link` and `symlink` are used in this PR and it should probably stick to one of the two.

The `follow-links` setting already existed and a breaking change is not worth it for such a small rename and to stay consistent I also named the setting `deduplicate-links`. In the code I mostly use symlink tough (there was once place where I used `links` but I changed that now. The only place where it's called `deduplicate-links` is in the `FilePickerConfig` struct (which corresponds to the setting)

helix-term/src/ui/mod.rs
```diff
@@ -194,15 +195,11 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
...
-        let is_dir = entry.file_type().map_or(false, |ft| ft.is_dir());
-        if is_dir {
-            // Will give a false positive if metadata cannot be read (eg. permission error)
-            None
-        } else {
-            Some(entry.into_path())
+        if entry.file_type()?.is_file() {
+            return Some(entry.into_path());
         }
+        None
```
the-mikedavis, line 202: very minor style nit, I like expressions

```suggestion
        if entry.file_type()?.is_file() {
            Some(entry.into_path())
        } else {
            None
        }
```

And a then_some will be good for this case in the future, eh?

the-mikedavis (commented):
Just a minor style nit, otherwise I think this looks good 👍

the-mikedavis:
Ah nice, this fixes themes showing up twice in the helix repo under `runtime/themes` and through the `contrib/themes` symlink 👍

helix-term/src/ui/mod.rs
pascalkuthe: good point, there was code in between there previously and I forgot to change it into something nice. I actually like this better as an if tough here. I would have to use `.then(||...)` here because I call `into_path`. If you use that with `then_some` clippy complains :D

the-mikedavis (approved):

PR #5659, opened 2023-01-23
Switch from toml::from_slice to toml::from_str

pascalkuthe:
Unblocks #5656

Just a small fix since that function was removed.
I found the usage of `std::fs::read` instead of `std::fs::read_to_string` odd anyway.
Sadly there is no `include_str!` macro so we have the conversion ourselves (not a big deal regardless).

### DIFF
--- a/helix-loader/src/config.rs
+++ b/helix-loader/src/config.rs
@@ -1,6 +1,9 @@
+use std::str::from_utf8;
+
 /// Default built-in languages.toml.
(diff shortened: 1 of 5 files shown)


the-mikedavis (approved):

poliorcetics:
no `include_str!` ? Are you talking about https://doc.rust-lang.org/std/macro.include_str.html ?

Or does `languages.toml` accept non-UTF8 ? `read_to_string` is used so I suppose not ?

PR #5677, opened 2023-01-25
Change default language server for 'v' from 'vls' to 'v ls'

antoniusnaumann:
According to the [V Language Server Repository README](https://github.com/vlang/vls#via-v-cli-recommended), the recommended way to install the V Language Server is via the `v` CLI. Therefore, the default config should expect it to be called by using `v ls` not `vls`.


the-mikedavis:
Could you run `cargo xtask docgen` and commit the results? That will clear up the CI failure

antoniusnaumann:
Done!

the-mikedavis (approved):

PR #5682, opened 2023-01-25
Add a version-control statusline element

dgyurov:
## What is this Pull Request about?
This pull request is introducing a new statusline element capable of displaying the version control `HEAD` of the current workspace.

## How can I try it?
The proposed statusline element is opt-in only and will require the user to add `"version-control"`  element in their `[editor.statusline]` section of the user configuration file.

## How does it work?
1) If the current `HEAD` is a reference, it's shortened name will be displayed in the statusline. 
2) In the case of a detached `HEAD`, the first eight characters of the commit hash will be displayed instead.
3) In the case of opening helix with a root folder without a git-repository, nothing will be displayed.

## Visual preview
preview

## UX Considerations
The current version control HEAD is displayed in the statusline for each opened document. That means that if the user opens files from different repositories in the same Helix instance, this implementation will be able to tackle that. Updates to the data feeding the statusline are only done when opening a file, or when explicitly reloading a file. The current implementation of the version control element also takes into account that in the future we will have a file watching capability in Helix, which will allow us to reactively update the data feeding the version control element.


pascalkuthe:
Thank you for working on this. I definitely want to see this feature in the future and the implementation seems decent.

Sadly this feature is still missing some prerequisites (and hence I didn't implement it myself).

Without repo caching this is sadly too slow. The status line rendering is called VERY often so not having the repo cached here causes a LOT of IO on every redraw. The problem with caching this is that a cached repo doesn't reload HEAD automatically so we need file watching too. I want to implement both of these things eventually but that will take some time. We could accept a version without file watching (but then the utility is a bit limited) but I think repo caching is a hard requirement. Performing large amounts of IO on every redraw is not something we want to ever do

dgyurov:
@pascalkuthe This is a fantastic feedback! Thanks a lot! I thought a lot about this PR in the last couple of days and I must agree that the best way to go forward will be to have a watcher in place and reactivity update the version control element. 

I had a look at how a simillar feature was implemented by the `lualine.nvim` project and they are following very similar approach to the one suggested by you. I do feel however that having a file watching functionality should be centralized in the codebase (simillar to Neovim), as there might be more use-cases in the future that can benefit from it.

I implemented an intermediate approach of having a version-control statusline element, which "piggy-backs" on the current git diff implementation. That is probably a reasonable MVP, as you mentioned, that will not have extra performance implications. Do you think that this will be an acceptable solution to the problem? The UX implications of it are that the statusline element will only be updated when a document is opened or when an explicit reload of the buffer is requested. Not ideal, but it might be enough and acceptable for most of the users. Personally, as a user, I would prefer to have a version control statusline element, even with the above mentioned imperfection.

Perhaps, it will be reasonable to complete this PR as a MVP first, and explore a centralized file watching solution separately? Will such iterative improvement be acceptable? Let me know what your thoughts on this are.

Thank you so much for the time again and have an awesome day! :)

pascalkuthe:
Good to see you are making progress. I didn't actually mean that you need to implement file watching. I am working on that locally and you are right a centralized solution is required. I think a MVP which doesn't change git branch automatically is required. 

That being said I am not quite happy with the implementation you choosen here. I don't want to couple the git diff gutter to the statuline element at all.
I actually liked that your previous implementation had a seperate function in the `DiffProvider`.
That functional should simply return an `Arc<ArcSwap<Arc<str>>>` which will be updated by the file-watcher automatically in the future. The returned value should be stored on the document just as `DiffHandle` as different files can belong to different git repositories.

pascalkuthe (dismissed):
After applying the review comments this look much better. The amaount of changes is actually quite small now, nice :+1:  Just like the diff gutters there is no more IO in the rendering loop and the implementation is also future proofed so that it's easy for a background thread to update the statusline through the `ArcSwap` :+1:.

As there is no file watching implemented the status line does not update automatically. However, I think that is actually a good think because then it is also more intuitive that the diff gutter does not update automatically either. So this status line indicator can make it more apparent what the diff is being computed against.

I think it makes sense to merge this as is and file watching in a future (separate) PR. Getting file watching right is fairly challenging, and as diff gutter where merged without that I think it's also fine to similarly accept this feature without filewatching.

dgyurov:
Thank you so much for the feedback! I really liked the proposed solution and iterated over my initial commit. 

Now each document is caching a head name internally and there is much stricter control for opening a corresponding repository. The cache is only invalidated and updated after an explicit reload by the end user. In the future when we have a centralized implementation of a file watcher, we can watch `.git/HEAD` for changes and reactively update the cache.

Assigning a head to each document was a fantastic idea! It allows us to open files from different repositories in the same session and the statusline element will still work as expected. Definitely an user experience improvement. Thanks again! :rocket:

helix-view/src/document.rs
```diff
@@ -143,6 +144,7 @@ pub struct Document {
     language_server: Option<Arc<helix_lsp::Client>>,
 
     diff_handle: Option<DiffHandle>,
+    version_control_head: Option<Arc<ArcSwap<Arc<str>>>>,
```
archseer, line 147: It seems weird to me that this is an ArcSwap wrapped in an Arc. Shouldn't we simply have an `ArcSwap<Arc<str>>`?

I think ArcSwap<T> already implies ArcSwapAny<Arc<T>> so actually just `ArcSwap<str>` might be enough.

helix-vcs/src/lib.rs
```diff
@@ -18,6 +19,7 @@ pub trait DiffProvider {
     /// The data is returned as raw byte without any decoding or encoding performed
     /// to ensure all file encodings are handled correctly.
     fn get_diff_base(&self, file: &Path) -> Option<Vec<u8>>;
+    fn get_current_head_name(&self, file: &Path) -> Option<Arc<ArcSwap<Arc<str>>>>;
```
archseer, line 22: This should deref and just return `Option<&str>` by derefing or maybe `Option<String>`. The git backend already basically allocates a `String` then wraps it in this complicated Arc type.

archseer (changes requested):

helix-vcs/src/lib.rs
pascalkuthe: I suggested this type because when filewatching is implemented we can keep a copy of the `ArcSwap` in the filwatcher and swap it out when the repo changes on disk.

helix-view/src/document.rs
pascalkuthe: Ah that is my bad I suggested that but I mixedup `ArcSwapAny` and `ArcSwap` which aliases to `ArcSwapAny<Arc<T>`

dgyurov: Thank you for the feedback! This is a great observation! I changed the type to `Option<Arc<ArcSwap<Box<str>>>>`. Unfortunately ArcSwap doesn't accept `str` directly because the size of the `str` values are not known at compilation time. We thought about choosing between `String` and `Box<str>` and went for the second, because that will be smaller.

pascalkuthe: it';s really unfortunate that it doesn't allow that, see https://docs.rs/arc-swap/latest/arc_swap/docs/limitations/index.html for more details. We could use a `ThinaArc` here but it's probably not worth the extra dependency

helix-vcs/src/lib.rs
dgyurov: That's a good point. Initially we also went for an `Option<String>`, but then we realized that once the filewatcher is in place, we will have to refactor the code and took the decision to implement it in a more future-proof way. Pascal's vision is for the diff provider itself to take care of the filewatching internally and therefore it's needed to return this complex type, to ensure that the consumers will be up to date with changes.

helix-view/src/document.rs
```diff
@@ -1093,6 +1098,19 @@ impl Document {
         }
     }
 
+    pub fn version_control_head(&self) -> Option<String> {
+        self.version_control_head
+            .clone()
+            .map(|head| (*head).load().as_ref().as_ref().to_owned())
```
Proful, line 1104: I'm afraid we can't do much. But this looks ugly.

helix-view/src/document.rs
```diff
@@ -1093,6 +1098,19 @@ impl Document {
...
+    }
```
pascalkuthe, line 1105:
```suggestion
    pub fn version_control_head(&self) -> Option<Arc<Box<str>>> {
        Some(self.version_control_head.as_ref()?.load_full())
    }
```

This is slightly more efficent by avoiding the clone in favor of an `Arc` clone and is a bit shorter.
It's a real shame that `Arc<str>` doesn't work `Option<Arc<str>>` would be much nicer :/ 

We could pull in `slice-dst` and use [`ArcSwap<StrWithHeader<()>>`](https://docs.rs/slice-dst/1.5.1/slice_dst/struct.StrWithHeader.html). I trust both the author and the implementation and it's a very small crate. Not sure if it's worth pulling in for such a smallthing,

dgyurov: Nice, that's way better, I like it! I went with an equivalent variant of your suggestion where I'm mapping on the optional instead. It felt slightly cleaner to me.
 
Probably this single use-case will not justify dragging in an extra dependency. I think we achieved a reasonable balance here. Getting rid of `(*head).load().as_ref().as_ref().to_owned()` here is already a huge win.

Great feedback! Thanks! :)

helix-view/src/document.rs
dgyurov: It was quite ugly, I must agree. But we just gave it a facelift! :lipstick: 

```rust
pub fn version_control_head(&self) -> Option<Arc<Box<str>>> {
    self.version_control_head.as_ref().map(|a| a.load_full())
}
```

Thanks for the feedback! :smile:

archseer (approved):
Looks good, just needs a conflict resolution

dgyurov:
Thanks a lot! I just resolved the conflicts with the latest commit. :)

the-mikedavis (approved):

PR #5692, opened 2023-01-26
health: add formatter binary

jzelinskie:
I noticed that some languages have external formatters that are not a part of their LSP configuration.
To surface that information, this PR adds a new output to --health that displays the formatter binary.

In the future, this could be expanded to look for other fields that contain commands from the language config.

Here's some example output.

```
$ cargo run -- --health go
    Finished dev [unoptimized + debuginfo] target(s) in 0.18s
     Running `target/debug/hx --health go`
Configured language server: gopls
Binary for language server: /Users/jzelinskie/.local/share/go/bin/gopls
Configured debug adapter: dlv
Binary for debug adapter: /Users/jzelinskie/.local/share/go/bin/dlv
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
$ cargo run -- --health cue
    Finished dev [unoptimized + debuginfo] target(s) in 0.07s
     Running `target/debug/hx --health cue`
Configured language server: cuelsp
Binary for language server: /Users/jzelinskie/.local/share/go/bin/cuelsp
Binary for formatter: /opt/homebrew/bin/cue
Highlight queries: ✓
Textobject queries: ✘
Indent queries: ✘
```

### DIFF
--- a/helix-term/src/health.rs
+++ b/helix-term/src/health.rs
@@ -262,8 +262,31 @@ pub fn language(lang_str: String) -> std::io::Result<()> {
         lang.debugger.as_ref().map(|dap| dap.command.to_string()),
     )?;
 
+    probe_commands(
+        lang.formatter
+            .as_ref()
+            .map(|fmtcfg| fmtcfg.command.to_string()),
+    )?;
+
     for ts_feat in TsFeature::all() {
         probe_treesitter_feature(&lang_str, *ts_feat)?
+    }
+
+    Ok(())
+}
+
+/// Display any additional binaries that are configured as commands for the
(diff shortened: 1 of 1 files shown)


rayanpiro:
I feel more comfortable with a solution like [this](https://github.com/rayanpiro/helix/commit/afad7bffbefe487bcd041d9b83fd720c7ab35c27).

In this case probe_commands has the same goal that probe_protocol. So just change some naming and the calls from the LSP, DAP and add the formatter one.

Tests OK and is working properly.

jzelinskie:
Fixed in #7986

PR #5695, opened 2023-01-26
feat(commands): add clear_registers command

### LINKED ISSUE #4256: Delete registers

chorcheus:
Closes #4256. This is a rework of closed https://github.com/helix-editor/helix/pull/4408. Summary of discussion from there:

 * clear registers by pressing `"` and then the key to clear the registers
 * the key to clear the registers is suggested to be `|`

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2165,6 +2165,38 @@ fn reset_diff_change(
     Ok(())
 }
 
+fn clear_register(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    ensure!(args.len() <= 1, ":clear-register takes at most 1 argument");
+    if args.is_empty() {
+        cx.editor.registers.clear();
+        cx.editor.set_status("All registers cleared");
+        return Ok(());
+    }
+
+    ensure!(
+        args[0].chars().count() == 1,
+        format!("Invalid register {}", args[0])
+    );
+    let register = args[0].chars().next().unwrap_or_default();
+    match cx.editor.registers.remove(register) {
+        Some(_) => cx
+            .editor
+            .set_status(format!("Register {} cleared", register)),
+        None => cx
+            .editor
+            .set_error(format!("Register {} not found", register)),
+    }
+    Ok(())
+}
+
 pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
         TypableCommand {
             name: "quit",
@@ -2716,6 +2748,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             aliases: &["diffget", "diffg"],
             doc: "Reset the diff change at the cursor position.",
             fun: reset_diff_change,
+            signature: CommandSignature::none(),
+        },
+        TypableCommand {
+            name: "clear-register",
+            aliases: &[],
+            doc: "Clear given register. If no argument is provided, clear all registers.",
+            fun: clear_register,
             signature: CommandSignature::none(),
         },
     ];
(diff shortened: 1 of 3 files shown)


the-mikedavis:
We could also consider not adding a keybinding for this. You would need to execute it through the command palette `<space>?` but I don't think that's a problem since this isn't something you need to do very often (if at all)

chorcheus:
@the-mikedavis Thanks, I pushed the changes. User will still be able to create custom bindings, so this might be just fine.

helix-term/src/commands.rs
```diff
@@ -4457,6 +4458,12 @@ fn insert_register(cx: &mut Context) {
     })
 }
 
+fn clear_registers(cx: &mut Context) {
+    cx.editor.autoinfo = None;
```
the-mikedavis, line 4462: This line can be removed: it's used in the other `*_register` commands because they add autoinfo and then clear it in the on-next-key callback

chorcheus: Done.

archseer:
This could be a typable command? `:clear-register @`

helix-term/src/commands.rs
```diff
@@ -4564,6 +4565,11 @@ fn insert_register(cx: &mut Context) {
     })
 }
 
+fn clear_registers(cx: &mut Context) {
+    cx.editor.registers.clear();
+    cx.editor.set_status("All registers cleared.")
+}
```
the-mikedavis, line 4571: If we have the typable command `:clear-register` then we don't need the regular command here. You can create keybindings that use `:clear-register`.

the-mikedavis: I see this one is clearing all registers while the typable command clears the given register. The typable command could clear all registers if no arguments are provided.

(Side-note: this would be a nice application for typable command flags, like `:clear-register --all`. I'll make an issue for that)

helix-term/src/commands/typed.rs
```diff
@@ -1952,6 +1952,27 @@ fn run_shell_command(
...
 
+fn clear_register(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    ensure!(args.len() == 1, ":clear_register takes exactly 1 argument");
```
the-mikedavis, line 1963:
```suggestion
    ensure!(args.len() == 1, ":clear-register takes exactly 1 argument");
```

helix-term/src/commands/typed.rs
```diff
@@ -1952,6 +1952,34 @@ fn run_shell_command(
...
+        return Ok(());
+    }
+    ensure!(args.len() <= 1, ":clear-register takes at most 1 argument");
+    if args.is_empty() {
+        cx.editor.registers.clear();
+        cx.editor.set_status("All registers cleared");
+        return Ok(());
+    }
+
+    ensure!(args[0].len() == 1, format!("Invalid register {}", args[0]));
```
sudormrfbin, line 1970: Won't this cause problems with keyboards that have non ASCII keys? A register can be any character on the keyboard and `len()` only counts the number of bytes. `args[0].chars().count()` would be more appropriate in that case.

helix-term/src/commands.rs
chorcheus: Good point. I left only the typable `:clear-register`

helix-term/src/commands/typed.rs
chorcheus: Thanks for pointing that out, I pushed the corrected version.

dead10ck (dismissed):
This seems fine to me 👍

chorcheus:
@the-mikedavis could you please look at this? Are you OK with the current implementation?

the-mikedavis (approved):

pascalkuthe (approved):

PR #5697, opened 2023-01-26
Fix clippy 1.67 warnings

mimame:
This is my first commit using Rust and to Helix, so I hope to be useful :smile:

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -77,7 +77,7 @@ impl<T: Item> Menu<T> {
         Self {
             options,
             editor_data,
-            matcher: Box::new(Matcher::default()),
+            matcher: Box::default(),
             matches,
             cursor: None,
             widths: Vec::new(),
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -431,7 +431,7 @@ impl<T: Item> Picker<T> {
         let mut picker = Self {
             options,
             editor_data,
-            matcher: Box::new(Matcher::default()),
+            matcher: Box::default(),
             matches: Vec::new(),
             cursor: 0,
             prompt,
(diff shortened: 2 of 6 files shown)


archseer:
1.67 uses features which aren't available on our MSRV. Also the formatting string lint should be disabled: https://twitter.com/adam_chal/status/1618673969590304775

archseer:
https://github.com/rust-lang/rust-clippy/issues/10087

mimame:
Hi @archseer, I've reduced the changes to only MSRV warnings, If I'm not wrong 
Thanks so much for your review!

helix-term/src/ui/menu.rs
```diff
@@ -77,7 +77,7 @@ impl<T: Item> Menu<T> {
         Self {
             options,
             editor_data,
-            matcher: Box::new(Matcher::default()),
+            matcher: Box::<fuzzy_matcher::skim::SkimMatcherV2>::default(),
```
pascalkuthe, line 80: You don't need the turbo fish (type specification) here. `Bix::default()` should be enough and is much more readable

helix-term/src/ui/picker.rs
```diff
@@ -431,7 +431,7 @@ impl<T: Item> Picker<T> {
         let mut picker = Self {
             options,
             editor_data,
-            matcher: Box::new(Matcher::default()),
+            matcher: Box::<fuzzy_matcher::skim::SkimMatcherV2>::default(),
```
pascalkuthe, line 434: Same here

helix-term/src/ui/menu.rs
mimame: Thanks so much @pascalkuthe for your review! 
Fixed

helix-term/src/ui/picker.rs
mimame: Thanks so much @pascalkuthe for your review! 
Reverted

helix-term/src/ui/menu.rs
pascalkuthe: You don't need to revert this changes entirely.
You reverted to `Box::new(Matcher::default())` but I mean that you can just use `Box::default()` here. 
With my comment I just ment that you don't need the `::<fuzzy_matcher::skim::SkimMatcherV2>` part

mimame: Thanks for the clarification.
Fixed

pascalkuthe (approved):
LGTM

the-mikedavis (approved):

Issue #4798, opened 2022-11-18
I would like to have search and global search with highlight the matching part. 


grv07:
![image](https://user-images.githubusercontent.com/11697220/202640806-afb81551-d6b7-43c8-8c60-755590d036ea.png)

We can't tell what is my match part without highlight it.

wes-adams:
i like this feature too. i'll look into this one

wes-adams:
![demo](https://user-images.githubusercontent.com/7883354/215101815-608651a6-f36e-4057-96a3-7b9e9395163f.gif)

wes-adams:
@grv07 - take a look. i'm very happy with this so far. is this what you are looking for?

wes-adams:
The state of this effort is:
* working demo
* highlight style is 'random'. it can differ with each new instance of `hx`
  * i suspect this code:
```
helix-view/src/editor.rs

1596      pub fn get_matches(&self) -> Option<Vec<(usize, std::ops::Range<usize>)>> {
1597      ╎   let mut ret = Vec::new();
1598      ╎   for (start, end) in &self.matches {
1599      ╎   ╎   ret.push((0, start.clone()..end.clone()));
                            ^
                             \____ not sure what this argument does!
```
the above `Vec` is used below. it is added to `highlights`, and then passed to `EditorView::render_text_highlights()` to be displayed.
```
helix-term/src/ui/editor.rs

151 ▔    ╎   let highlights = match editor.get_matches() {
152      ╎   ╎   Some(x) => Box::new(syntax::merge(highlights, x)),
153      ╎   ╎   None => highlights,
154      ╎   };
155      ╎
156      ╎   Self::render_text_highlights(doc, view.offset, inner, surface, theme, highlights, &config);
```

ajitid:
If we do this then we would be back the Vim problem where the background color of the active selection is different from rest of the matches. If you may remember, there were plugins made just to resolve this: like to flash the active highlight once you hit enter after typing the search term. We would essentially need a different token to discern between the active match with the rest of the matches.

Issue #185, opened 2021-06-08
Regex Automata


kirawi:
In the future, we should switch to something like [Alacritty's regex functionality](https://github.com/alacritty/alacritty/blob/96ea5c445ecc25a5b2d400b11f0ff843e9cbdd7a/alacritty_terminal/src/term/search.rs#L198-L283) so we don't waste memory in our current implementation.

archseer:
[This regex thread](https://github.com/rust-lang/regex/issues/425) is also relevant.

In particular:

> Before getting too excited, you'll want to check out the list of differences between regex and regex-automata to see whether it's possible to use it. It's sadly probably not good enough for supporting a user facing text editor, for example, likely because of poor regex compile times.

and 

> As a lower level crate, this library does not do literal optimizations. In exchange, you get predictable performance regardless of input. The philosophy here is that literal optimizations should be applied at a higher level, although there is no easy support for this in the ecosystem yet.

This last one is especially limiting, since often you're just searching for a literal.

I think we need to give it a shot, but benchmark the performance hit. An optimization: we need to compile 4 DFAs, two for each search direction. But `/` always searches forward, so we can simply compile the forward DFAs while the user is inputting the regex in the prompt, then delay compiling the reverse DFAs until `Enter` is pressed.

kirawi:
I think ultimately, this is a low priority feature for us right now.

pascalkuthe:
I don't think using `regex-automata` is the path forward here, DFA compilation fundamentally has exponential time/space complexity.
That is why the regex engines basically always compiles them lazily and intelligently fall back to other engines.


What do you guys think about using a fork of the regex crate to finally resolve this issue?
I think I could fork regex that work with ropey `chunk` iterators instead of slices.
An initial patch could be pretty easy by just transparently replacing byte slices with `ropey` chunks iterator by writing a smart indexing replacement. Basically I would check if we moved past the chunk boundary on every index and intelligently advance to the correct chunk until we hit the correct chunk (advancing either forward or backwards). 
This approach introduces some overhead but  since chunk boundaries are pretty rare I think it will be pretty small thanks to branch prediction (a couple cycels per index). I am pretty confident this performance tradeoff ispreferable (both performance/memory consumption wise) to converting the entire `Rope` to a string.
I am not 100% sure this will work but I would certainly like to give it a try.

In the future I might be able to optimize this further by actually handling chunk boundaries in the engines itself but that would be much much more effort.
[...]

PR #5702, opened 2023-01-27
Tmp - highlight working, but still needs work

wes-adams:
This is for demo and feedback purposes.
I plan to get feedback, and make improvements.

References:
https://github.com/helix-editor/helix/issues/4798


the-mikedavis:
I haven't looked at this yet but also see https://github.com/helix-editor/helix/pull/4635 which I think will have some overlap implementation-wise

wes-adams:
> I haven't looked at this yet but also see #4635 which I think will have some overlap implementation-wise

Thanks. I'll definitely look into that work.
This PR is 100% not ready. I'm still finding my way in this project and Rust in general, so any feedback and/or reviews I can get will be greatly appreciated!

wes-adams:
i think the only optimization left is to search only the visible text for matches versus the entire document
i've given this a little effort, but would welcome any feedback i can get

pascalkuthe:
Just a couple comments:
* The matches should not be stored on the editor, they should be stored on the document or (even better) on the view
* Right now the matches stick around forever which causes a host of problems if the text is edited. To fix this you either need to update the matches on every edit or hide them as soon as the document is edited. Regex matches has quite a bit of overhead because we need to convert `Rope` -> `String` (see #185) so that would not be ideal. I also dislike how long the highlights stick around in nvim (even tough I like the basic feature of search highlights) so I think it's actually a great solution to remove the search matches on edits.
* Updating the matches while scrolling is too slow. I think you could recompute the matches on idle timeout if the view position changed.
* This potentially should be rebased on #5420 too as the way the "visible area" is calculated changes there (the view offset changes), this PR should also not be merged before that because I would have to update this command as part of #5420 then (which is already very large)

wes-adams:
> Just a couple comments:
> 
> * The matches should not be stored on the editor, they should be stored on the document or (even better) on the view
> * Right now the matches stick around forever which causes a host of problems if the text is edited. To fix this you either need to update the matches on every edit or hide them as soon as the document is edited. Regex matches has quite a bit of overhead because we need to convert `Rope` -> `String` (see [Regex Automata #185](https://github.com/helix-editor/helix/issues/185)) so that would not be ideal. I also dislike how long the highlights stick around in nvim (even tough I like the basic feature of search highlights) so I think it's actually a great solution to remove the search matches on edits.
> * Updating the matches while scrolling is too slow. I think you could recompute the matches on idle timeout if the view position changed.
> * This potentially should be rebased on [rework positioning/rendering and enable softwrap/virtual text #5420](https://github.com/helix-editor/helix/pull/5420) too as the way the "visible area" is calculated changes there (the view offset changes), this PR should also not be merged before that because I would have to update this command as part of [rework positioning/rendering and enable softwrap/virtual text #5420](https://github.com/helix-editor/helix/pull/5420) then (which is already very large)

i have this PR in the following state:
* `search_matches`  highlights are updated with each new key pressed (in `search_impl`)
  * if `n`(or others) is used to traverse searches, `search_matches` are not updated
* `;` can be used to clear highlights, but highlights will resume if `n` is used to continue match traversal
* **highlights do not go away when doc updates**
  * i tried to make this happen with `Document::is_modified()` but i had troubles
  * if `is_modified() == true` and the document is never saved, the `search_matches` highlights would never be rendered
  * i think i need help with this (either updating `search_matches` on edit, or clearing `search_matches` on edit)

wes-adams:
after the changes in https://github.com/helix-editor/helix/pull/5702/commits/3bfb871ebea0b9b92f361325c3366b3ad6b8a0ef, `search_matches` highlights will go away when document is altered via `insert` characters. 

the highlights **do not** go away due to deletion or paste. still need to figure this out.

helix-term/src/ui/editor.rs
```diff
@@ -145,11 +145,20 @@ impl EditorView {
...
+                Some(match_highlights) => Box::new(syntax::merge(highlights, match_highlights)),
+                None => highlights,
+            };
+        } else {
+            // if document has changed, set bool in view
+            // view.search_matches.need_clear.replace(true);
+            view.search_matches
+                .need_clear
+                .swap(true, std::sync::atomic::Ordering::Relaxed);
+        }
```
wes-adams, line 161: this is where i 'clear' the highlights if the doc has been modified. `if doc.changes().is_empty()` then i update the highlights, otherwise, i use my insanly convoluted `AtomicBool` to tell View to clear out `search_matches` as soon as i exit the current routine, and get back to a scope where view is mutable.

helix-view/src/view.rs
```diff
@@ -102,6 +102,38 @@ pub struct ViewPosition {
...
+
+impl fmt::Debug for SearchMatches {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SearchMatches")
+            .field("matches", &self.matches)
+            .field("new", &self.new)
+            .field("old", &self.old)
+            .finish()
+    }
+}
```
wes-adams, line 123: probably unnecessary - can remove

wes-adams, line 123: probably not needed - can remove

helix-view/src/view.rs
```diff
@@ -102,6 +102,38 @@ pub struct ViewPosition {
...
+
+impl SearchMatches {
+    pub fn clear(&mut self) {
+        self.matches.clear();
+    }
+
+    pub fn has_changed(&self, regex: &Regex) -> bool {
+        regex.as_str() != self.old.as_str()
+    }
```
wes-adams, line 132: could not figure out a clean way to compare `Regex`, so this is what i came up with

helix-view/src/view.rs
```diff
@@ -154,6 +187,11 @@ impl View {
             object_selections: Vec::new(),
             gutters,
             doc_revisions: HashMap::new(),
+            search_matches: SearchMatches {
+                matches: Vec::new(),
+                new: Regex::new("").unwrap(),
+                old: Regex::new("").unwrap(),
```
wes-adams, line 196: i assume `unwrap` is safe to use in a case like this? but it is a big assumption...

helix-view/src/view.rs
```diff
@@ -576,6 +614,40 @@ impl View {
...
+            self.search_matches.update(regex);
+            let contents = doc.text().slice(..).to_string();
+            self.search_matches.clear();
+            for one_match in self.search_matches.new.find_iter(&contents) {
+                self.search_matches
+                    .matches
+                    .push((one_match.start(), one_match.end()));
+            }
+        }
+    }
```
wes-adams, line 654: the first `if` checks if the `Regex` search criteria has changed. this prevents this block from executing if the user is traversing match results using `n` or similar. if the `Regex` search criteria has indeed changed, `old` is set equal to `new`, and existing matches are cleared. new matches are calculated via expensive `Regex` operation based on `new` search criteria

helix-term/src/commands.rs
```diff
@@ -2316,6 +2321,12 @@ fn collapse_selection(cx: &mut Context) {
         Range::new(pos, pos)
     });
     doc.set_selection(view.id, selection);
+
+    // clear highlights on search matches
+    // but set search_matches.old to "" so that if 'n' is pressed
+    // to continue to traverse matches, highlights resume
+    view.search_matches.clear();
+    view.search_matches.old = Regex::new("").unwrap();
```
wes-adams, line 2329: again... is `unwrap` safe to use here?

helix-term/src/ui/editor.rs
wes-adams, line 161: this is where i 'clear' the highlights if the doc has been modified. `if doc.changes().is_empty()` then i update the highlights before rendering, otherwise, i use my insanly convoluted `AtomicBool` to tell `view` to clear out `search_matches` as soon as i exit the current routine, and get back to a scope where `view` is mutable (which happens in `render()` line 1384).

helix-term/src/commands.rs
Proful: As it is initialize code with empty string. I don't think unwrap will panic here in runtime. It's ok imo.

pathwave:
Is the color here the same as selections? Might need to difference between those.

A feature I would like is to highlight words that match the current primary selection, and that seems to overlap with this (since when you search the search term gets selected)

wes-adams:
> Is the color here the same as selections? Might need to difference between those.
> 
> A feature I would like is to highlight words that match the current primary selection, and that seems to overlap with this (since when you search the search term gets selected)

the `search_matches` highlight color is dictated by `ui-selection`:
```
    pub fn get_search_matches(
        &self,
        theme: &Theme,
    ) -> Option<Vec<(usize, std::ops::Range<usize>)>> {
        let selection_scope = theme
            .find_scope_index("ui.selection")
            .expect("could not find `ui.selection` scope in the theme!");
. . . 
```
an easy update or change to have a different highlight for `selection` and `search_matches`

helix-view/src/view.rs
```diff
@@ -576,6 +614,47 @@ impl View {
...
+        if self.search_matches.has_changed(regex) {
+            self.search_matches.update(regex);
+            self.search_matches.clear();
+            for one_match in self.search_matches.new.find_iter(contents) {
+                self.search_matches
+                    .matches
+                    .push((one_match.start(), one_match.end()));
+            }
+        }
+    }
```
wes-adams, line 657: the first `if` checks if the `Regex` search criteria has changed. this prevents this block from executing if the user is traversing match results using `n` or similar. if the `Regex` search criteria has indeed changed, `old` is set equal to `new`, and existing matches are cleared. new matches are calculated via expensive `Regex` operation based on `new` search criteria

wes-adams:
this one is ready for a closer look guys

i think the behavior is close to what @pascalkuthe outlined in the comment above

@the-mikedavis 
@pascalkuthe

wes-adams:
would either of you please take a look?
i'd be very grateful for some feedback on this one.
@the-mikedavis @pascalkuthe

tmpm697:
What's status of this PR? this is really helpful to have all matched words highlighted.

lpchaim:
I've been looking into adding a search index/total matches statusline as I was used to back in nvim (e.g., indicating you're on match 2 out of 5 with something like [2/5] on the bottom) and it does look like I should wait until this gets merged in the interest of not duplicating the work made so that the context holds search match information and whatnot, so I'll be keeping an eye on this.

wes-adams:
@the-mikedavis @archseer @pascalkuthe @sudormrfbin 
sorry to blindly tag you guys again, but i'm still hoping to get some sort of review on this PR
would anyone kindly help me gain some traction on this?

PR #5707, opened 2023-01-27
Enable Metals for java files

ayoub-benali:
It is very common in Scala projects to use java dependencies and in such context Metals should be enabled for java files.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1057,7 +1057,7 @@ source = { git = "https://github.com/UserNobody14/tree-sitter-dart", rev = "2d7f
 name = "scala"
 scope = "source.scala"
 roots = ["build.sbt", "build.sc", "build.gradle", "pom.xml", ".scala-build"]
-file-types = ["scala", "sbt", "sc"]
+file-types = ["scala", "sbt", "sc", "java"]
 comment-token = "//"
 indent = { tab-width = 2, unit = "  " }
 language-server = { command = "metals" }


languages.toml
```diff
@@ -1057,7 +1057,7 @@ source = { git = "https://github.com/UserNobody14/tree-sitter-dart", rev = "2d7f
 name = "scala"
 scope = "source.scala"
```
ayoub-benali, line 1058: I am not sure if I should add `source.java` here as well

pascalkuthe:
This messes with java developmemt. Java is it's own language with it's own languages.toml entry, tree sitter queried grammar and lsp.

Multiple languages with the same filed types causes conflicts (helix will more or less randomly decide which language to associate with a file then).  Instead you can change the lsp for java in your personal config.

the-mikedavis:
This could be a recommended configuration under https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#metals, but you would want to change the `language-server` configuration for the java language config entry rather than the file-types for scala because then the scala tree-sitter parser could be used on java files as Pascal says.

```toml
[[language]]
name = "java"
language-server = { command = "metals", language-id = "java" }
```

ayoub-benali:
Thanks for the explanation, I will update doc then https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#metals. But regarding the proposed config in `~/.config/helix/languages.toml`: would it start a new metals instance if I jump from Scala file to a Java file ? The expected behavior would be to keep the same one

ayoub-benali:
Also could the language config be done per project one https://github.com/helix-editor/helix/pull/3207 is merged ?

pascalkuthe:
#3207 is only required for config.toml.
Per project language.toml settings are already possible. Just add a `.helix/languages.toml` folder to the root of your repository. There you can overwrite settings youw would normally add to `~/.config/languages.toml`

the-mikedavis:
I'm not sure if it will spawn a new one: I think the language server registry is keyed off of language-id if I remember correctly rather than binary name so I would guess that Helix will spawn a new metals server. I don't think there's a way to tell Helix to re-use a language server across languages currently. That might be a nice improvement for front-end languages too since javascript, jsx, typescript and tsx all use the same language server configuration but for different language-id's.

PR #5711, opened 2023-01-27
properly handle LSP position encoding

### LINKED ISSUE #2547: Odd errors with JavaScript language server
### Summary

Just noticed some odd errors with the JavaScript language server. Not sure what’s causing them. If I quit Helix and restart the errors go away until I add more code to the file. Not sure how to reproduce but here’s the code that’s causing it for me.

Screenshot from 2022-05-23 16-36-23

```js
import tapMonkey from '../index.js'
import assert from 'assert'
import strip from 'strip-ansi'

// Test handler tests
tapMonkey.quiet = false
tapMonkey.testHandler({name: 'mock'})
console.log(tapMonkey.spinner)
tapMonkey.spinner.stop()

// General tests

assert(tapMonkey.spinner._spinner.interval === 300, 'spinner internal is as expected')
assert(tapMonkey.spinner._spinner.frames[0] === '  🙈 ', 'animation frame 1 is correct')
assert(tapMonkey.spinner._spinner.frames[1] === '  🙈 ', 'animation frame 2 is correct')
assert(tapMonkey.spinner._spinner.frames[2] === '  🙉 ', 'animation frame 3 is correct')
assert(tapMonkey.spinner._spinner.frames[3] === '  🙊 ', 'animation frame 4 is correct')

assert(strip(tapMonkey.spinner.text).includes('Running mock tests'), 'test name is displayed correctly')

console.log('All tests passed.')
process.exit()
```

### Reproduction Steps

_No response_

### Helix log

No pertinent details found in log.

</details>

### Platform

Linux (Fedora Silverblue 36)

### Terminal Emulator

GNOME Console

### Helix Version

helix 22.05-dev (8df8ff27)

### LINKED ISSUE #3286: Rust analyzer stops responding after adding a bunch of emojis
### Summary

```rust
fn main() {
    println!("Hello, world!");
    // 👆
}
```
Adding more of the "👆" emoji to the comment causes the language server to stop responding.

Might be related to #3283 which also has to do with emojis.

I don't think this is an upstream bug in `rust-analyzer` as trying this in vscode with `rust-analyzer` works fine

### Reproduction Steps

 

[asciinema](https://asciinema.org/a/MQMsm1bL8PNorAD4B7RUQto8F)

I tried this:

1. `hx .`
2. open the file with 
```rust
fn main() {
    println!("Hello, world!");
    // 👆
}
```
3. Wait for language server to respond (by hitting `<space> k` repeatedly
4. Copy paste the emoji a bunch of times
5. Try viewing the documentation for `println!` (by hitting `<space> k`), notice it doesn't respond

I expected this to happen:
I see the documentation for `println!`, like I did before adding the emojis.

Instead, this happened:
No documentation appeared

### Helix log

```
2022-08-01T21:04:07.872 helix_lsp::transport [ERROR] err <- "Panic context:\n"
2022-08-01T21:04:07.872 helix_lsp::transport [ERROR] err <- "> \n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "version: 0.0.0 (7e2b983fd 2022-07-24)\n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "notification: textDocument/didChange\n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "\n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "thread 'LspServer' panicked at 'assertion failed: self.is_char_boundary(n)', /rustc/1.62.1/library/alloc/src/string.rs:1800:29\n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "stack backtrace:\n"
2022-08-01T21:04:07.873 helix_lsp::transport [ERROR] err <- "note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\n"
2022-08-01T21:04:08.061 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-08-01T21:04:08.061 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-08-01T21:04:08.372 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.403 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.432 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.463 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.494 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.525 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.555 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.586 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.616 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.647 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.677 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:08.708 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:09.973 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:13.802 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2022-08-01T21:04:14.303 helix_term::application [ERROR] Timed out waiting for language servers to shutdown
```

</details>

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

helix 22.05

### LINKED ISSUE #4791: Crash on Save JSON File with Emojis
### Summary

when saving a file with the following contents:
```json
[
"🇺🇸",
"🎄"
]
```
helix crashes with the following error:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Invalid char range 14..13: start must be <= end', /Users/blah/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.0/src/rope.rs:546:37
```

### Reproduction Steps

create a file `data.json` with the following contents:
```json
[
"🇺🇸",
"🎄"
]
```
(note that the level of indentation appears to be important)

Open the file with helix `hx data.json`. 

Save the file with `:w`.

Observe the crash.

### Helix log

```
2022-11-17T15:31:01.642 helix_loader [DEBUG] Located configuration folders: ["[redacted]"]
2022-11-17T15:31:01.651 helix_view::clipboard [INFO] Using pbcopy+pbpaste to interact with the system clipboard
2022-11-17T15:31:01.653 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2022-11-17T15:31:01.653 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2022-11-17T15:31:01.653 mio::poll [TRACE] registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
2022-11-17T15:31:01.653 helix_lsp::client [INFO] Using custom LSP config: {"provideFormatter":true}
2022-11-17T15:31:01.653 mio::poll [TRACE] registering event source with poller: token=Token(4), interests=READABLE | WRITABLE
2022-11-17T15:31:01.654 mio::poll [TRACE] registering event source with poller: token=Token(5), interests=READABLE | WRITABLE
2022-11-17T15:31:01.654 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"initialize","params":{"capabilities":{"textDocument":{"codeAction":{"codeActionLiteralSupport":{"codeActionKind":{"valueSet":["","quickfix","refactor","refactor.extract","refactor.inline","refactor.rewrite","source","source.organizeImports"]}}},"completion":{"completionItem":{"resolveSupport":{"properties":["documentation","detail","additionalTextEdits"]},"snippetSupport":false},"completionItemKind":{}},"hover":{"contentFormat":["markdown"]},"publishDiagnostics":{},"rename":{"dynamicRegistration":false,"honorsChangeAnnotations":false,"prepareSupport":false},"signatureHelp":{"signatureInformation":{"activeParameterSupport":true,"documentationFormat":["markdown"],"parameterInformation":{"labelOffsetSupport":true}}}},"window":{"workDoneProgress":true},"workspace":{"applyEdit":true,"configuration":true,"didChangeConfiguration":{"dynamicRegistration":false},"symbol":{"dynamicRegistration":false},"workspaceFolders":true}},"initializationOptions":{"provideFormatter":true},"processId":25858,"rootPath":"[redacted]","rootUri":"[redacted]","workspaceFolders":[{"name":"fox-weather-service","uri":"[redacted]"}]},"id":0}
2022-11-17T15:31:01.654 mio::poll [TRACE] registering event source with poller: token=Token(0), interests=READABLE
2022-11-17T15:31:01.654 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE
2022-11-17T15:31:01.821 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","id":0,"result":{"capabilities":{"textDocumentSync":2,"hoverProvider":true,"documentSymbolProvider":true,"documentRangeFormattingProvider":true,"documentFormattingProvider":true,"colorProvider":{},"foldingRangeProvider":true,"selectionRangeProvider":true,"documentLinkProvider":{}}}}
2022-11-17T15:31:01.821 helix_lsp::transport [INFO] <- {"capabilities":{"colorProvider":{},"documentFormattingProvider":true,"documentLinkProvider":{},"documentRangeFormattingProvider":true,"documentSymbolProvider":true,"foldingRangeProvider":true,"hoverProvider":true,"selectionRangeProvider":true,"textDocumentSync":2}}
2022-11-17T15:31:01.821 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"initialized","params":{}}
2022-11-17T15:31:01.821 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"workspace/didChangeConfiguration","params":{"settings":{"provideFormatter":true}}}
2022-11-17T15:31:01.821 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didOpen","params":{"textDocument":{"languageId":"json","text":"[\n\"🇺🇸\",\n\"🎄\"\n]\n","uri":"[redacted]","version":0}}}
2022-11-17T15:31:01.823 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"[redacted]","diagnostics":[]}}
2022-11-17T15:31:02.449 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/formatting","params":{"options":{"insertSpaces":true,"tabSize":2},"textDocument":{"uri":"[redacted]"}},"id":1}
2022-11-17T15:31:02.451 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","id":1,"result":[{"range":{"start":{"line":0,"character":1},"end":{"line":1,"character":0}},"newText":"\n  "},{"range":{"start":{"line":1,"character":7},"end":{"line":2,"character":0}},"newText":"\n  "},{"range":{"start":{"line":3,"character":1},"end":{"line":4,"character":0}},"newText":""}]}
2022-11-17T15:31:02.451 helix_lsp::transport [INFO] <- [{"newText":"\n  ","range":{"end":{"character":0,"line":1},"start":{"character":1,"line":0}}},{"newText":"\n  ","range":{"end":{"character":0,"line":2},"start":{"character":7,"line":1}}},{"newText":"","range":{"end":{"character":0,"line":4},"start":{"character":1,"line":3}}}]
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller
2022-11-17T15:31:02.453 mio::poll [TRACE] deregistering event source from poller

```

</details>

### Platform

macOS

### Terminal Emulator

iTerm2 3.4.16

### Helix Version

helix 22.08.1 (a7ec4aae)

### LINKED ISSUE #5809: Incorrect underlining of errors and wrong results from code action
### Summary

I think this came from the @pascalkuthe softwrap undated as I have not seen it before, but it is possible it is the language server at fault.

### Reproduction Steps

Create this markdown file:
```markdown
> 💡 If you are using the musl-libc instead of glibc the following environment variable must be set during the build
> to ensure tree sitter grammars can be loaded correctly:
```
Use ltex-ls to spell check:
```
[[language]]
name = "markdown"
max-line-length = 80
language-server = { command = "ltex-ls" }
```
NOTE: Softwrap is turned off.

and it looks like this, the underlines are shifted ahead a character, note this never usually happens:
Screenshot from 2023-02-03 17-08-36

Make the correction:
Screenshot from 2023-02-03 17-10-52

It fails to replace the word:
Screenshot from 2023-02-03 17-11-21

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

https://github.com/helix-editor/helix/commit/30412366be411335b7e2600e9b4178355c27da15

pascalkuthe:
Fixes #4791
Fixes #3286
Fixes #2547 
Fixes #5809

I started to thoroughly investigate the char_pos <-> lsp_pos conversion in an effort to fix #4791. Initially I just started with the suggestion by @the-mikdavis of capping the `character` offset to the line length. While this is only mentioned offhand in the LSP standard, it's still part of the standard and seems a good precaution against crashing from broken LSPs.

This fixed the original panic but caused panics and weird formatting bugs for slightly different json files. So I investigated further and found a pretty mayor bug: Helix implemented the `UTF-8` encoding incorrectly:

Currently, helix treat the `Position::character` field as a char offset in `Utf-8` mode and as `utf16` offset in UTF-16 mode. This is incorrect, the `Position::character` field corresponds to `UTF-8` byte, what we had really implemented was `UTF-32`.

From the LSP spec:

> When describing positions the protocol needs to specify how offsets (specifically character offsets) should be interpreted. The corresponding PositionEncodingKind is negotiated between the client and the server during initialization.

> ```
> /**
> * A type indicating how positions are encoded,
> * specifically what column offsets mean.
> *
> * @since 3.17.0
> */
> export type PositionEncodingKind = string;
> 
> /**
> * A set of predefined position encoding kinds.
> *
> * @since 3.17.0
> */
> export namespace PositionEncodingKind {
> 
> /**
> * Character offsets count UTF-8 code units (e.g bytes).
> */
> export const UTF8: PositionEncodingKind = 'utf-8';
> 
> /**
> * Character offsets count UTF-16 code units.
> *
> * This is the default and must always be supported
> * by servers
> */
> export const UTF16: PositionEncodingKind = 'utf-16';
> 
> /**
> * Character offsets count UTF-32 code units.
> *
> * Implementation note: these are the same as Unicode code points,
> * so this `PositionEncodingKind` may also be used for an
> * encoding-agnostic representation of character offsets.
> */
> export const UTF32: PositionEncodingKind = 'utf-32';
> }
> ```

Fixing the UTF-8 encoding however only made the problems worse. The reason was simply that it was incorrect for us to use UTF-8 encoding at all. The `lsp-types` crate and helix (and many LSP servers) don't support LSP `3.17` yet so this option is not available to us. Instead we just hardcoded the encoding to `UTF-8`. However, we should be hard-coding helix to use utf-16 instead. To quote the standard:

> Prior to 3.17 the offsets were always based on a UTF-16 string representation. So in a string of the form a𐐀b the character offset of the character a is 0, the character offset of 𐐀 is 1 and the character offset of b is 3 since 𐐀 is represented using two code units in UTF-16. Since 3.17 clients and servers can agree on a different string encoding representation (e.g. UTF-8). The client announces it’s supported encoding via the client capability general.positionEncodings. The value is an array of position encodings the client supports, with decreasing preference (e.g. the encoding at index 0 is the most preferred one). To stay backwards compatible the only mandatory encoding is UTF-16 represented via the string utf-16. The server can pick one of the encodings offered by the client and signals that encoding back to the client via the initialize result’s property capabilities.positionEncoding. If the string value utf-16 is missing from the client’s capability general.positionEncodings servers can safely assume that the client supports UTF-16. If the server omits the position encoding in its initialize result the encoding defaults to the string value utf-16.

The only reason this hasn't caused much more problems yet is that these two bugs somewhat cancel out. Specifically we treated `UTF-8` like `UTF-32`. Characters that require 2 `UTF-16` codepoints instead of just 1 are rare and therefore problems were rare and these bugs essentially canceled out.

TLDR:
* Helix uses UTF-8 encoding for LSP offsets
* It should be using UTF-16 (first bug fixed here)
* The UTF-8 encoding was accidently a UTF-32 encoding (second bug fixed here), which is the same as UTF-16 for commonly used chars
* Added some minor cleanups and comments to be closer to the LSP spec

### DIFF
--- a/helix-core/src/line_ending.rs
+++ b/helix-core/src/line_ending.rs
@@ -203,6 +203,13 @@ pub fn line_end_char_index(slice: &RopeSlice, line: usize) -> usize {
             .unwrap_or(0)
 }
 
+pub fn line_end_byte_index(slice: &RopeSlice, line: usize) -> usize {
+    slice.line_to_byte(line + 1)
+        - get_line_ending(&slice.line(line))
+            .map(|le| le.as_str().len())
+            .unwrap_or(0)
+}
+
 /// Fetches line `line_idx` from the passed rope slice, sans any line ending.
 pub fn line_without_line_ending<'a>(slice: &'a RopeSlice, line_idx: usize) -> RopeSlice<'a> {
     let start = slice.line_to_char(line_idx);
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -104,7 +104,7 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             capabilities: OnceCell::new(),
-            offset_encoding: OffsetEncoding::Utf8,
+            offset_encoding: OffsetEncoding::Utf16,
             config,
             req_timeout,
 
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -147,10 +147,10 @@ pub mod util {
         // Note that the end of the line here is **before** the line terminator
         // so we must use `line_end_char_index` istead of `doc.line_to_char(pos_line + 1)`
         //
-        // FIXME: Helix does not fully compley with the LSP spec for line terminators.
+        // FIXME: Helix does not fully comply with the LSP spec for line terminators.
         // The LSP standard requires that line terminators are ['\n', '\r\n', '\r'].
-        // Without the unicode-linbreak feature disabeled, the `\r` terminator is not handeled by helix.
-        // With the unicode-linbreak feature, helix recognized multiple extra line break chars
+        // Without the unicode-linebreak feature disabled, the `\r` terminator is not handled by helix.
+        // With the unicode-linebreak feature, helix recognizes multiple extra line break chars
         // which means that positions will be decoded/encoded incorrectly in their presence
 
         let line = match offset_encoding {
@@ -177,7 +177,7 @@ pub mod util {
             .unwrap_or(line.end)
             .min(line.end);
 
-        // TODO prefer UTF32/char indecies to avoid this step
+        // TODO prefer UTF32/char indices to avoid this step
         match offset_encoding {
             OffsetEncoding::Utf8 => doc.try_byte_to_char(pos).ok(),
             OffsetEncoding::Utf16 => doc.try_utf16_cu_to_char(pos).ok(),
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -57,6 +57,7 @@ pub enum OffsetEncoding {
 
 pub mod util {
     use super::*;
+    use helix_core::line_ending::{line_end_byte_index, line_end_char_index};
     use helix_core::{diagnostic::NumberOrString, Range, Rope, Selection, Tendril, Transaction};
 
     /// Converts a diagnostic in the document to [`lsp::Diagnostic`].
@@ -117,7 +118,7 @@ pub mod util {
 
     /// Converts [`lsp::Position`] to a position in the document.
     ///
-    /// Returns `None` if position exceeds document length or an operation overflows.
+    /// Returns `None` if position.line is out of bounds or an overflow occurs
     pub fn lsp_pos_to_pos(
         doc: &Rope,
         pos: lsp::Position,
@@ -128,22 +129,58 @@ pub mod util {
             return None;
         }
 
-        match offset_encoding {
+        // We need to be careful here to fully comply ith the LSP spec.
+        // Two relevant quotes from the spec:
+        //
+        // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#position
+        // > If the character value is greater than the line length it defaults back
+        // >  to the line length.
+        //
+        // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocuments
+        // > To ensure that both client and server split the string into the same
+        // > line representation the protocol specifies the following end-of-line sequences:
+        // > ‘\n’, ‘\r\n’ and ‘\r’. Positions are line end character agnostic.
+        // > So you can not specify a position that denotes \r|\n or \n| where | represents the character offset.
+        //
+        // This means that while the line must be in bounds the `charater`
+        // must be capped to the end of the line.
+        // Note that the end of the line here is **before** the line terminator
+        // so we must use `line_end_char_index` istead of `doc.line_to_char(pos_line + 1)`
+        //
+        // FIXME: Helix does not fully compley with the LSP spec for line terminators.
+        // The LSP standard requires that line terminators are ['\n', '\r\n', '\r'].
+        // Without the unicode-linbreak feature disabeled, the `\r` terminator is not handeled by helix.
+        // With the unicode-linbreak feature, helix recognized multiple extra line break chars
+        // which means that positions will be decoded/encoded incorrectly in their presence
+
+        let line = match offset_encoding {
             OffsetEncoding::Utf8 => {
-                let line = doc.line_to_char(pos_line);
-                let pos = line.checked_add(pos.character as usize)?;
-                if pos <= doc.len_chars() {
-                    Some(pos)
-                } else {
-                    None
-                }
+                let line_start = doc.line_to_byte(pos_line);
+                let line_end = line_end_byte_index(&doc.slice(..), pos_line);
+                line_start..line_end
             }
             OffsetEncoding::Utf16 => {
-                let line = doc.line_to_char(pos_line);
-                let line_start = doc.char_to_utf16_cu(line);
-                let pos = line_start.checked_add(pos.character as usize)?;
-                doc.try_utf16_cu_to_char(pos).ok()
+                // TODO directly translate line index to char-idx
+                // ropey can do this just as easily as utf-8 byte translation
+                // but the functions are just missing.
+                // Translate to char first and then utf-16 as a workaround
+                let line_start = doc.line_to_char(pos_line);
+                let line_end = line_end_char_index(&doc.slice(..), pos_line);
+                doc.char_to_utf16_cu(line_start)..doc.char_to_utf16_cu(line_end)
             }
+        };
+
+        // The LSP spec demands that the offset is capped to the end of the line
+        let pos = line
+            .start
+            .checked_add(pos.character as usize)
+            .unwrap_or(line.end)
+            .min(line.end);
+
+        // TODO prefer UTF32/char indecies to avoid this step
+        match offset_encoding {
+            OffsetEncoding::Utf8 => doc.try_byte_to_char(pos).ok(),
+            OffsetEncoding::Utf16 => doc.try_utf16_cu_to_char(pos).ok(),
         }
     }
 
@@ -158,8 +195,8 @@ pub mod util {
         match offset_encoding {
             OffsetEncoding::Utf8 => {
                 let line = doc.char_to_line(pos);
-                let line_start = doc.line_to_char(line);
-                let col = pos - line_start;
+                let line_start = doc.line_to_byte(line);
+                let col = doc.char_to_byte(pos) - line_start;
 
                 lsp::Position::new(line as u32, col as u32)
             }
@@ -606,16 +643,55 @@ mod tests {
         }
 
         test_case!("", (0, 0) => Some(0));
-        test_case!("", (0, 1) => None);
+        test_case!("", (0, 1) => Some(0));
         test_case!("", (1, 0) => None);
         test_case!("\n\n", (0, 0) => Some(0));
         test_case!("\n\n", (1, 0) => Some(1));
-        test_case!("\n\n", (1, 1) => Some(2));
+        test_case!("\n\n", (1, 1) => Some(1));
         test_case!("\n\n", (2, 0) => Some(2));
         test_case!("\n\n", (3, 0) => None);
         test_case!("test\n\n\n\ncase", (4, 3) => Some(11));
         test_case!("test\n\n\n\ncase", (4, 4) => Some(12));
-        test_case!("test\n\n\n\ncase", (4, 5) => None);
+        test_case!("test\n\n\n\ncase", (4, 5) => Some(12));
         test_case!("", (u32::MAX, u32::MAX) => None);
+    }
+
+    #[test]
+    fn emoji_format_gh_4791() {
+        use lsp_types::{Position, Range, TextEdit};
+
+        let edits = vec![
+            TextEdit {
+                range: Range {
+                    start: Position {
+                        line: 0,
+                        character: 1,
+                    },
+                    end: Position {
+                        line: 1,
+                        character: 0,
+                    },
+                },
+                new_text: "\n  ".to_string(),
+            },
+            TextEdit {
+                range: Range {
+                    start: Position {
+                        line: 1,
+                        character: 7,
+                    },
+                    end: Position {
+                        line: 2,
+                        character: 0,
+                    },
+                },
+                new_text: "\n  ".to_string(),
+            },
+        ];
+
+        let mut source = Rope::from_str("[\n\"🇺🇸\",\n\"🎄\",\n]");
+
+        let transaction = generate_transaction_from_edits(&source, edits, OffsetEncoding::Utf8);
+        assert!(transaction.apply(&mut source));
     }
 }
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -481,6 +481,11 @@ impl Transaction {
         for (from, to, tendril) in changes {
             // Verify ranges are ordered and not overlapping
             debug_assert!(last <= from);
+            // Verify ranges are correct
+            debug_assert!(
+                from <= to,
+                "Edit end must end before it starts (should {from} <= {to})"
+            );
 
             // Retain from last "to" to current "from"
             changeset.retain(from - last);


helix-lsp/src/client.rs
```diff
@@ -104,7 +104,7 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             capabilities: OnceCell::new(),
-            offset_encoding: OffsetEncoding::Utf8,
+            offset_encoding: OffsetEncoding::Utf16,
```
pascalkuthe, line 107: this is the core fix everything else is just cleanup/future proofing. As we harcode the encoding right now so only the encoding specified here gets actually used (and utf-16 encoding was implemented correctly for the most part).

helix-lsp/src/lib.rs
```diff
@@ -606,16 +643,55 @@ mod tests {
...
         test_case!("\n\n", (0, 0) => Some(0));
         test_case!("\n\n", (1, 0) => Some(1));
-        test_case!("\n\n", (1, 1) => Some(2));
+        test_case!("\n\n", (1, 1) => Some(1));
         test_case!("\n\n", (2, 0) => Some(2));
         test_case!("\n\n", (3, 0) => None);
         test_case!("test\n\n\n\ncase", (4, 3) => Some(11));
         test_case!("test\n\n\n\ncase", (4, 4) => Some(12));
-        test_case!("test\n\n\n\ncase", (4, 5) => None);
+        test_case!("test\n\n\n\ncase", (4, 5) => Some(12));
```
pascalkuthe, line 655: changes in this test reflect that the offset by `Position::character` is now capped to the line length (before the line terminator)

pascalkuthe:
Likely fixes #3286 too but I haven't tested that yet

pascalkuthe:
Likely fixes #2547 too altough that PR is missing info to reproduce the issue reliably. But since editing text after emoojis is involved I am quite certain that the problem was also caused by UTF-32/UTF-16 missmatch

helix-lsp/src/client.rs
archseer: Wasn't the idea here that we ask for UTF-8 instead of UTF-16 if possible? Based on the clangd negotiation proposal that was merged into the spec. We should prefer UTF-8 if it can be negotiated, then UTF-16 as fallback.

pascalkuthe: Yeah exactly but that proposal is only available in LSP 3.17 which `lsp_types` doesn't support yet.
Right now we don't negotiate anything (offset_encoding is only set here and never changed) and just assume UTF-8 which is incorrect for all LSPs. 

Once `lsp_types` updates to 3.17 (there is a PR but not yet merged) I am happy to add that. For now using UTF-16 seems the right thing to me tough as almost all LSPs only support that (and it's the mandatory to use UTf-16 when not otherwise negotiated according to the standard)

pascalkuthe: Also from a performance standpoint UTF-32 is actually faster as we use chard indecies. Whether we convert UTf-8 or UTF-16 to UTF-32 (char index) with ropey makes very little difference. Ropey does almost the same thing in either case.

archseer: So actually, the spec mandates UTF-16 (thanks Microsoft..) but a lot of implementations do it wrong and just use UTF-8 offsets

pascalkuthe: Hmm that is unfortunate but switching to UTF-16 can't be causing too many problems as right now on master our UTF-8 implementation is actually a UTF-32 implementation by accident (we treat the `characters` offset as char offsets instead of byte offsets) so right now helix behaves incorrectly for either case.

I think we should probably stick to the standard. We could add an option to langauges.toml to work around incorrect LS. The `lsp_types` update isn't far away so that will be the proper solution then.

Just to clarify: This is just about how the `characters` field of the `Position` struct is treated. The actual text send back and forth is always UTF-8 (that is standard compliant and it's stupid that the offset encoding doesn't match the actual data but oh well)

pascalkuthe: I am pretty sure that most LSPs support utf-16 as a fallback tough. For example for clangd:

https://clangd.llvm.org/extensions.html#utf-8-offsets


> New client capability: offsetEncoding : string[]:
> 
> Lists the encodings the client supports, in preference order. It SHOULD include "utf-16". If not present, it is assumed to be ["utf-16"]

pascalkuthe:
Also fixes #5809

helix-core/src/line_ending.rs
```diff
@@ -203,6 +203,13 @@ pub fn line_end_char_index(slice: &RopeSlice, line: usize) -> usize {
             .unwrap_or(0)
 }
 
+pub fn line_end_byte_index(slice: &RopeSlice, line: usize) -> usize {
+    slice.line_to_byte(line + 1)
```
the-mikedavis, line 207: Is there any risk of going over the line count here with the `+ 1`?

helix-lsp/src/lib.rs
```diff
@@ -128,22 +129,58 @@ pub mod util {
...
+        // This means that while the line must be in bounds the `charater`
+        // must be capped to the end of the line.
+        // Note that the end of the line here is **before** the line terminator
+        // so we must use `line_end_char_index` istead of `doc.line_to_char(pos_line + 1)`
+        //
+        // FIXME: Helix does not fully compley with the LSP spec for line terminators.
+        // The LSP standard requires that line terminators are ['\n', '\r\n', '\r'].
+        // Without the unicode-linbreak feature disabeled, the `\r` terminator is not handeled by helix.
+        // With the unicode-linbreak feature, helix recognized multiple extra line break chars
+        // which means that positions will be decoded/encoded incorrectly in their presence
```
the-mikedavis, line 154:
```suggestion
        // FIXME: Helix does not fully comply with the LSP spec for line terminators.
        // The LSP standard requires that line terminators are ['\n', '\r\n', '\r'].
        // Without the unicode-linebreak feature disabled, the `\r` terminator is not handled by helix.
        // With the unicode-linebreak feature, helix recognizes multiple extra line break chars
        // which means that positions will be decoded/encoded incorrectly in their presence
```

helix-lsp/src/lib.rs
```diff
@@ -128,22 +129,58 @@ pub mod util {
...
+        };
+
+        // The LSP spec demands that the offset is capped to the end of the line
+        let pos = line
+            .start
+            .checked_add(pos.character as usize)
+            .unwrap_or(line.end)
+            .min(line.end);
+
+        // TODO prefer UTF32/char indecies to avoid this step
```
the-mikedavis, line 180:
```suggestion
        // TODO prefer UTF32/char indices to avoid this step
```

the-mikedavis (commented):
Some small comments but otherwise looks good to me

helix-core/src/line_ending.rs
pascalkuthe: Hmm this is something I didn't consider and a good point.

I just copied the `line_end_char_idx` function and converted it to return a byte index. For this specific usecase it's fine since ropey allows one past the end index but LSP doesn't (and that's enforced at the callsite of this function at the start of `lsp_pos_to_pos`. We should probably check all existing call sites of `line_end_char_idx` but that would be seprerste from this PR as it's an existing function.

the-mikedavis (approved):

archseer (approved):

yjhn:
Both `lsp-types` and `rust-analyzer` now support LSP 3.17's encoding negotiation ([link](https://github.com/gluon-lang/lsp-types/pull/248), [link](https://github.com/rust-lang/rust-analyzer/pull/13484)), so it would be nice for Helix to support it too since it uses UTF-8 under the hood anyway.

PR #5712, opened 2023-01-27
Esc aborts completions reverting any changes

g-re-g:
Currently if you have some completions open and you hit the `Esc` key, the completion picker will get closed but the changes will still be applied to you code. This seems incorrect for 2 reasons:

 - The docs say that `Escape` will close the completions the same way `Ctrl-c` will. https://docs.helix-editor.com/master/keymap.html#picker (Admittedly I don't know if the picker this is referring to is the completion picker)
 - The imports for the change aren't properly inserted as they would be had the user pressed `Enter`

Here's some recordings, note you can't see that I'm pressing `Esc`, you'll just have to trust me or try it out:

Before this PR:
https://asciinema.org/a/yQed8bXSB3mJOHp4IqE5sld4B

After this PR:
https://asciinema.org/a/rTj3yfQbUWmU1VjSenY1ImsJb

I do not know that this is the best way to do this, and I don't know why this code originally deliberately ignored the Esc key but looked like a quick fix.

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -363,14 +363,16 @@ impl Completion {
 
 impl Component for Completion {
     fn handle_event(&mut self, event: &Event, cx: &mut Context) -> EventResult {
-        // let the Editor handle Esc instead
-        if let Event::Key(KeyEvent {
-            code: KeyCode::Esc, ..
-        }) = event
-        {
-            return EventResult::Ignored(None);
+        // Escape needs to get handled by the popup so that completions can
+        // be properly reverted and it needs to bubble up so that helix
+        // will go into normal mode.
+        let event_result = self.popup.handle_event(event, cx);
+        match event {
+            Event::Key(KeyEvent {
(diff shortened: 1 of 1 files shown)


pascalkuthe:
The docs page you mentioned refers to the licker as in the files picker or global search which is not the same thing as the completion menu at all (I don't know if/where those keybindings are documented).

Regardless I do think that ESC should either
* either accept the completion (like hitting enter)
* or discard the completion

I think that the second option is more fitting here.
Howber this PR changes intended behavior because with this change helix does not exit insertmode anymore.

In fact the special case was introduced for just that reason in d54ae09d3b7a5eb56050631a58e75a1d1686ea70. I think the right think is to handle the event in the popuo the popup but then repot the KeyEvent as ignored regardless for esc. So instead of removing the specical case you would just move it after `self.popup.handle_event(event, cx).` (and save that functions return value and return that in an else.

archseer:
Hmm I need to double check but I think this behavior was deliberate so you can avoid <enter><esc> to confirm a selection.

PR #5720, opened 2023-01-29
Modify env language to extend bash

EricCrosson:
Additionally, add `.envrc` to the `env`-supported file types.


archseer:
Hmm I think this is still incorrect: we do want to indent the `}` and code inside each clause, but the toplevel clauses should be same level as the select:

```go
		select {
		case c <- x:
			x, y = y, x+y
		case <-quit:
			fmt.Println("quit")
			return
		}
```

\cc @Triton171 what would the correct indentation class be here?

archseer (approved):

archseer:
What about `.env`? `Would cover .env.staging` etc

EricCrosson:
I'm not sure @archseer, `.env` is listed as a file-type under the `env` language [here](https://github.com/helix-editor/helix/blob/482cc22fecc109773882d3e0fd6251c33862e89e/languages.toml#L2010)

I sniped `.envrc` because it covers the majority of direnv use-cases and didn't conflict with any other settings, as far as I can tell

archseer:
`.env` has the same usecase as `.envrc`. https://direnv.net/man/direnv.toml.1.html#codeloaddotenvcode

Both of these should be under the same filetype. We could use the bash grammar for all env types and drop the `tree-sitter-env` grammar (I'd still use a distinct language definition though).

EricCrosson:
I'm not very familiar with this code -- after reading https://docs.helix-editor.com/master/languages.html I'm still not quite sure what changes you're suggesting.

If I delete [this line](https://github.com/helix-editor/helix/blob/482cc22fecc109773882d3e0fd6251c33862e89e/languages.toml#L2010) and add `.env` to [bash.file-types](https://github.com/helix-editor/helix/blob/482cc22fecc109773882d3e0fd6251c33862e89e/languages.toml#L537), is that what you're suggesting?

the-mikedavis:
You can remove this block https://github.com/helix-editor/helix/blob/482cc22fecc109773882d3e0fd6251c33862e89e/languages.toml#L2016-L2018

and add a line `grammar = "bash"` to the `[[language]]` entry for env. Then you could change the `runtime/queries/env/*.scm` to only say `; inherits: bash`. This will make env a separate language that re-uses the bash tree-sitter parser and highlights.

EricCrosson:
Thanks for the assist @the-mikedavis :cowboy_hat_face: 

I followed your instructions and moved `.envrc` from the bash `file-types` to the env `file-types`. How does it look now?

pascalkuthe:
This LGTM, I think the testsuite failure is just a GA problem and unrelated to the actual PR. I can't rerun workflows but I am pretty sure that a rerunning CI would fix the pipeline

the-mikedavis (approved):

PR #5487, opened 2023-01-10
allow LSP insert text to replace non-matching prefixes (#5469)

Taywee:
closes #5469 

Most LSPs will complete case-insensitive matches, particularly from lowercase to uppercase.  In some cases, notably Pyright, this is given as a simple insert text instead of TextEdit.  When this happens, the prefix text was left unedited.

This change seems to fix it for me.  I'm not sure if there are other side effects to doing it this way, though.  If another LSP gave an insert text that was just the rest to be inserted, I imagine this would still cause problems.  But the previous method would also fail in that case with repeated text.  If you typed `su` and the suggestion was `surration`, it would be ambiguous in that case whether that was to be inserted (making `susurration`) or replaced entirely (making `surration`).

Issue #5676, opened 2023-01-25
Completion replaces existing word incorrectly if they end on the same letter


aral:
Token following where a completion should be inserted is erroneously replaced with the completion.

![Screenshot from 2023-01-25 13-57-45](https://user-images.githubusercontent.com/41798/214583951-582b4576-3db0-4aeb-bf23-12a5390e8edc.png)

I press <kbd>i</kbd> and start typing.

![Screenshot from 2023-01-25 13-57-50](https://user-images.githubusercontent.com/41798/214583949-20afbd40-07d0-43a1-9b1f-574c078b40d2.png)

So far, so good…

![Screenshot from 2023-01-25 13-57-53](https://user-images.githubusercontent.com/41798/214583944-34ab5eb3-7df6-48b7-a612-20156847add1.png)

I select the first completion. And it replaces “events” instead of being inserted before it.

![Screenshot from 2023-01-25 13-57-57](https://user-images.githubusercontent.com/41798/214583943-a1d1b931-1773-4dc7-8b3f-e826c23bf592.png)

helix 22.12 (08c35ab6) on Fedora Silverblue 37  in Black Box terminal under GNOME.

Possibly related to #2580

archseer:
I think this is just a missing config here: https://github.com/helix-editor/helix/blob/91dca3f667e146f93e6dca205b0a3f2a7a64b488/helix-term/src/ui/completion.rs#L114-L117

archseer:
@pascalkuthe you mentioned this on another issue somewhere ^

pascalkuthe:
Sorry forgot about this @archseer . 

This is indeed intended behavior. LSPs provide two options for handeling autocompletions: replace or insert.
Helix is currently hardcoded to always use the replace edit. Many editors default to insert instead (vscode and `nvim.cmp` do atleast). I think adding  a config option here would be an easy PR. all we need to do is to add a condition to use `item.insert` instead of `item.replace` in the snippet @archseer linked.

aral:
@pascalkuthe @archseer When would this ever be the desired behaviour? It results in corrupted syntax. I’d highly recommend making insert the default behaviour. I don’t remember once ending up with corrupted syntax while using VSCode (and I’m so glad I don’t have to use VSCode anymore for a multitude of other reasons).

Also, note that the replacement isn’t being carried out perfectly either (the selection goes up to but does not include the final “t”). That might be a separate bug, thought.

pascalkuthe:
[...]

PR #1819, opened 2022-03-15
completion fix

cossonleo:
Completion will not work correct When We type chars between  sended to completion request and recived the result.  These chars will not be replaced by completion items.

PR #5728, opened 2023-01-30
Improved Completion

### LINKED ISSUE #4380: LSP Autocompletion does not insert, it overwrites
### Summary

# Helix
[helix.webm](https://user-images.githubusercontent.com/12832280/196912431-6d97b689-a83c-4419-ad31-d04901182189.webm)

# Neovim
[neovim.webm](https://user-images.githubusercontent.com/12832280/196912448-1d1f404d-3430-460f-a8f1-000fc2a6916c.webm)

Not only does Helix wrongly overwrite `width` it is also one character short of the end so when immediately typing  `.` it becomes `rectangl.e`

### Reproduction Steps

na

### Helix log

na

### Platform

Linux

### Terminal Emulator

KItty

### Helix Version

helix 22.08.1 (418a622d)

```rust
pub struct Rectangle {
    pub width: usize,
    pub height: usize,
    pub position: Point2d,
}

fn draw_rectangle(
    rectangle: &Rectangle,
    content: &style::StyledContent<String>,
    stdout: &Stdout,
) -> Result<()> {
    for y in 0..width {
        for x in 0..rectangle.height {
            if (y == 0 || y == rectangle.width - 1) || (x == 0 || x == rectangle.height - 1) {
                draw_content(
                    stdout,
                    Point2d {
                        x: rectangle.position.x + y,
                        y: rectangle.position.y + x,
                    },
                    content,
                )?;
            }
        }
    }

    Ok(())
}
```

### LINKED ISSUE #5469: Pyright LSP completion lowercase to uppercase leaves first few characters
### Summary

When I use Pyright completion that changes case from lowercase to uppercase, the first two characters remain instead of being replaced.

The same does not happen using the same language server in neovim, so I don't think it's an issue with the language server itself.

### Reproduction Steps

Using this `languages.toml` entry for Pyright:

```toml
[[language]]
name = "python"

[language.language-server]
command = "pyright-langserver"
args = ["--stdio"]

[language.config.python.analysis]
autoSearchPaths = true
diagnosticMode = "workspace"
useLibraryCodeForTypes = true
```

If I create a simple python project with a virtualenvironment and activate the language server:

```sh
$ cd $(mktemp -d)
$ touch pyproject.toml
$ mkdir src
$ python -mvenv ./venv
$ . ./venv/bin/activate
$ pip install --force pyright
$ hx src/example.py
```
Then add a simple uppercase `Uppercase` class, and start typing `uppercase` and complete it to the class, I end up with this:

```py
class Uppercase:
    pass

upUppercase
```

If I type it in the right casing to begin with, it has no problem and does not duplicate the first two letters.

### Helix log

```
2023-01-09T10:07:44.240 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":0,"line":2},"start":{"character":0,"line":2}},"text":"\n"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":23}}}
2023-01-09T10:07:44.501 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":23,"diagnostics":[]}}
2023-01-09T10:07:44.800 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":0,"line":3},"start":{"character":0,"line":3}},"text":"u"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":24}}}
2023-01-09T10:07:44.984 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":1,"line":3},"start":{"character":1,"line":3}},"text":"p"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":25}}}
2023-01-09T10:07:44.991 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/completion","params":{"position":{"character":2,"line":3},"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py"}},"id":12}
2023-01-09T10:07:44.994 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","id":12,"result":{"items":[{"label":"Uppercase","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Uppercase"},"sortText":"09.9999.Uppercase"},{"label":"super","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"super"},"sortText":"09.9999.super"},{"label":"tuple","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"tuple"},"sortText":"09.9999.tuple"},{"label":"KeyboardInterrupt","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"KeyboardInterrupt"},"sortText":"09.9999.KeyboardInterrupt"},{"label":"LookupError","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"LookupError"},"sortText":"09.9999.LookupError"},{"label":"InterruptedError","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"InterruptedError"},"sortText":"09.9999.InterruptedError"},{"label":"ProcessLookupError","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"ProcessLookupError"},"sortText":"09.9999.ProcessLookupError"},{"label":"BaseExceptionGroup","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"BaseExceptionGroup"},"sortText":"09.9999.BaseExceptionGroup"},{"label":"ExceptionGroup","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"ExceptionGroup"},"sortText":"09.9999.ExceptionGroup"},{"label":"UnionType","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"autoImportText":"from types import UnionType","symbolLabel":"UnionType"},"sortText":"12.9999.UnionType","detail":"Auto-import","labelDetails":{"description":"types"},"documentation":{"kind":"plaintext","value":"from types import UnionType"},"textEdit":{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":2}},"newText":"UnionType"}},{"label":"Unpack","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"autoImportText":"from typing import Unpack","symbolLabel":"Unpack"},"sortText":"12.9999.Unpack","detail":"Auto-import","labelDetails":{"description":"typing"},"documentation":{"kind":"plaintext","value":"from typing import Unpack"},"textEdit":{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":2}},"newText":"Unpack"}},{"label":"Unpack","kind":7,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"autoImportText":"from typing_extensions import Unpack","symbolLabel":"Unpack"},"sortText":"12.9999.Unpack","detail":"Auto-import","labelDetails":{"description":"typing_extensions"},"documentation":{"kind":"plaintext","value":"from typing_extensions import Unpack"},"textEdit":{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":2}},"newText":"Unpack"}},{"label":"update_abstractmethods","kind":3,"data":{"workspacePath":"/tmp/tmp.ZmHxbsplID","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"autoImportText":"from abc import update_abstractmethods","symbolLabel":"update_abstractmethods"},"sortText":"12.9999.update_abstractmethods","detail":"Auto-import","labelDetails":{"description":"abc"},"documentation":{"kind":"plaintext","value":"from abc import update_abstractmethods"},"textEdit":{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":2}},"newText":"update_abstractmethods"}}],"isIncomplete":true}}
2023-01-09T10:07:44.994 helix_lsp::transport [INFO] <- {"isIncomplete":true,"items":[{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Uppercase","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"Uppercase","sortText":"09.9999.Uppercase"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"super","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"super","sortText":"09.9999.super"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"tuple","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"tuple","sortText":"09.9999.tuple"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"KeyboardInterrupt","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"KeyboardInterrupt","sortText":"09.9999.KeyboardInterrupt"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"LookupError","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"LookupError","sortText":"09.9999.LookupError"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"InterruptedError","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"InterruptedError","sortText":"09.9999.InterruptedError"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"ProcessLookupError","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"ProcessLookupError","sortText":"09.9999.ProcessLookupError"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"BaseExceptionGroup","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"BaseExceptionGroup","sortText":"09.9999.BaseExceptionGroup"},{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"ExceptionGroup","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"ExceptionGroup","sortText":"09.9999.ExceptionGroup"},{"data":{"autoImportText":"from types import UnionType","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"UnionType","workspacePath":"/tmp/tmp.ZmHxbsplID"},"detail":"Auto-import","documentation":{"kind":"plaintext","value":"from types import UnionType"},"kind":7,"label":"UnionType","labelDetails":{"description":"types"},"sortText":"12.9999.UnionType","textEdit":{"newText":"UnionType","range":{"end":{"character":2,"line":3},"start":{"character":0,"line":3}}}},{"data":{"autoImportText":"from typing import Unpack","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Unpack","workspacePath":"/tmp/tmp.ZmHxbsplID"},"detail":"Auto-import","documentation":{"kind":"plaintext","value":"from typing import Unpack"},"kind":7,"label":"Unpack","labelDetails":{"description":"typing"},"sortText":"12.9999.Unpack","textEdit":{"newText":"Unpack","range":{"end":{"character":2,"line":3},"start":{"character":0,"line":3}}}},{"data":{"autoImportText":"from typing_extensions import Unpack","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Unpack","workspacePath":"/tmp/tmp.ZmHxbsplID"},"detail":"Auto-import","documentation":{"kind":"plaintext","value":"from typing_extensions import Unpack"},"kind":7,"label":"Unpack","labelDetails":{"description":"typing_extensions"},"sortText":"12.9999.Unpack","textEdit":{"newText":"Unpack","range":{"end":{"character":2,"line":3},"start":{"character":0,"line":3}}}},{"data":{"autoImportText":"from abc import update_abstractmethods","filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"update_abstractmethods","workspacePath":"/tmp/tmp.ZmHxbsplID"},"detail":"Auto-import","documentation":{"kind":"plaintext","value":"from abc import update_abstractmethods"},"kind":3,"label":"update_abstractmethods","labelDetails":{"description":"abc"},"sortText":"12.9999.update_abstractmethods","textEdit":{"newText":"update_abstractmethods","range":{"end":{"character":2,"line":3},"start":{"character":0,"line":3}}}}]}
2023-01-09T10:07:45.136 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":2,"line":3},"start":{"character":2,"line":3}},"text":"p"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":26}}}
2023-01-09T10:07:45.352 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":3,"line":3},"start":{"character":3,"line":3}},"text":"e"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":27}}}
2023-01-09T10:07:45.472 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":4,"line":3},"start":{"character":4,"line":3}},"text":"r"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":28}}}
2023-01-09T10:07:45.730 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":28,"diagnostics":[{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":5}},"message":"\"upper\" is not defined","severity":1,"code":"reportUndefinedVariable","source":"Pyright","codeDescription":{"href":"https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUndefinedVariable"}},{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":5}},"message":"Expression value is unused","severity":2,"code":"reportUnusedExpression","source":"Pyright","codeDescription":{"href":"https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUnusedExpression"}}]}}
2023-01-09T10:07:45.992 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":5,"line":3},"start":{"character":2,"line":3}},"text":""}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":29}}}
2023-01-09T10:07:45.992 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"textDocument/didChange","params":{"contentChanges":[{"range":{"end":{"character":2,"line":3},"start":{"character":2,"line":3}},"text":"Uppercase"}],"textDocument":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":30}}}
2023-01-09T10:07:45.999 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"completionItem/resolve","params":{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Uppercase","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"Uppercase","sortText":"09.9999.Uppercase"},"id":13}
2023-01-09T10:07:46.002 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","id":13,"result":{"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Uppercase","workspacePath":"/tmp/tmp.ZmHxbsplID"},"kind":7,"label":"Uppercase","sortText":"09.9999.Uppercase","documentation":{"kind":"plaintext","value":"class Uppercase()"}}}
2023-01-09T10:07:46.002 helix_lsp::transport [INFO] <- {"data":{"filePath":"/tmp/tmp.ZmHxbsplID/src/example.py","position":{"character":2,"line":3},"symbolLabel":"Uppercase","workspacePath":"/tmp/tmp.ZmHxbsplID"},"documentation":{"kind":"plaintext","value":"class Uppercase()"},"kind":7,"label":"Uppercase","sortText":"09.9999.Uppercase"}
2023-01-09T10:07:46.250 helix_lsp::transport [INFO] <- {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"uri":"file:///tmp/tmp.ZmHxbsplID/src/example.py","version":30,"diagnostics":[{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":11}},"message":"\"upUppercase\" is not defined","severity":1,"code":"reportUndefinedVariable","source":"Pyright","codeDescription":{"href":"https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUndefinedVariable"}},{"range":{"start":{"line":3,"character":0},"end":{"line":3,"character":11}},"message":"Expression value is unused","severity":2,"code":"reportUnusedExpression","source":"Pyright","codeDescription":{"href":"https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportUnusedExpression"}}]}}

```

### Platform

Fedora Linux 37

### Terminal Emulator

gnome-terminal

### Helix Version

helix 22.12

pascalkuthe:
Fixes #4380 
Fixes #5469

This PR improves the completions in helix in multiple subtle ways. These fixes were lumped together as they all interact. 

The core change of this PR is a fix for #4380: Adding an option so that helix completions only insert text and don't replace the rest of the word (using `item.insert` instead of `item.replace`). I believe this to be the better default, so I changed the default behavior to use `item.insert`. This not only feels more appropriate for an *insert* mode action in a modal editor, but I feel that the replace mode has some issues that don't make it as suitable for a default setting (see below).

While working on this I noticed that in cases where a LSP does not supply a TextEdit, and we have to determine the range based on word boundaries ourself helix already behaves this way: We only insert the sent text but don't replace the rest of the word. I changed that behavior there to respect the new setting. Here is another reason why I think insert completions are a better default: There is a third option for sending completions (which is commonly used) where the LS does send a `TextEdit` but only a single edit (not two separate replace/insert edits). These edits are always insert edits so for many LSPs helix is already inserting by default.

To avoid conflicts with the change above, I also pulled in #5487 (commit remains to credit @Taywee). I discovered that #5487 doesn't respect multi cursors (so the edit was only applied to a single cursor), so I fixed that regression too.

In the process I noticed that completions with multiple cursors can cause overlapping ranges which can cause panics (always a panic in debug mode thanks to debug assertions). For example in helix in `document.rs` you can select `History` in the `use` statement type `s[iy]<ret>` start a completion and as soon as History is selected the editor instantly crashes in debug mode.I fixed the problem by adding a variant of `Selection::change` that ignores overlapping edits. This problem rarely occurs in practice and there are some odd edge cases that are not handled 100% right now (the cursor can be moved by the applied transactions which means the preview only applies the edit at once cursor but on hitting enter the cursor has moved and the edit is applied a both cursors) but I doubt people actually run into this problem as there was no bug report about the crash I could find. For now just fixing the crash seems good enough to me. This Problem is also more likely to occur in replace mode (as the rest of the word could contain another cursor) so yet another point for that.

### DIFF
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -1,3 +1,5 @@
+use smallvec::SmallVec;
+
 use crate::{Range, Rope, Selection, Tendril};
 use std::borrow::Cow;
 
@@ -466,6 +468,33 @@ impl Transaction {
         self
     }
 
+    /// Generate a transaction from a set of potentially overlapping changes. The `change_ranges`
+    /// iterator yield the range (of removed text) in the old document for each edit. If any change
+    /// overlaps with a range overlaps with a previous range then that range is ignored.
+    ///
+    /// The `process_change` callback is called for each edit that is not ignored (in the order
+    /// yielded by `changes`) and should return the new text that the associated range will be
+    /// replaced with.
+    ///
+    /// To make this function more flexible the iterator can yield additional data for each change
+    /// that is passed to `process_change`
+    pub fn change_ignore_overlapping<T>(
+        doc: &Rope,
+        change_ranges: impl Iterator<Item = (usize, usize, T)>,
+        mut process_change: impl FnMut(usize, usize, T) -> Option<Tendril>,
+    ) -> Self {
+        let mut last = 0;
+        let changes = change_ranges.filter_map(|(from, to, data)| {
+            if from < last {
+                return None;
+            }
+            let tendril = process_change(from, to, data);
+            last = to;
+            Some((from, to, tendril))
+        });
+        Self::change(doc, changes)
+    }
+
     /// Generate a transaction from a set of changes.
     pub fn change<I>(doc: &Rope, changes: I) -> Self
     where
@@ -511,6 +540,44 @@ impl Transaction {
         F: FnMut(&Range) -> Change,
     {
         Self::change(doc, selection.iter().map(f))
+    }
+
+    pub fn change_by_selection_ignore_overlapping(
+        doc: &Rope,
+        selection: &Selection,
+        mut change_range: impl FnMut(&Range) -> (usize, usize),
+        mut create_tendril: impl FnMut(usize, usize) -> Option<Tendril>,
+    ) -> (Transaction, Selection) {
+        let mut last_selection_idx = None;
+        let mut new_primary_idx = None;
+        let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
+        let process_change = |change_start, change_end, (idx, range): (usize, &Range)| {
+            // update the primary idx
+            if idx == selection.primary_index() {
+                new_primary_idx = Some(idx);
+            } else if new_primary_idx.is_none() {
+                if idx > selection.primary_index() {
+                    new_primary_idx = last_selection_idx;
+                } else {
+                    last_selection_idx = Some(idx);
+                }
+            }
+            ranges.push(*range);
+            create_tendril(change_start, change_end)
+        };
+        let transaction = Self::change_ignore_overlapping(
+            doc,
+            selection.iter().enumerate().map(|range| {
+                let (change_start, change_end) = change_range(range.1);
+                (change_start, change_end, range)
+            }),
+            process_change,
+        );
+
+        (
+            transaction,
+            Selection::new(ranges, new_primary_idx.unwrap_or(0)),
+        )
     }
 
     /// Insert text at each selection head.
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -59,8 +59,8 @@ pub enum OffsetEncoding {
 pub mod util {
     use super::*;
     use helix_core::line_ending::{line_end_byte_index, line_end_char_index};
+    use helix_core::{chars, RopeSlice, SmallVec};
     use helix_core::{diagnostic::NumberOrString, Range, Rope, Selection, Tendril, Transaction};
-    use helix_core::{smallvec, SmallVec};
 
     /// Converts a diagnostic in the document to [`lsp::Diagnostic`].
     ///
@@ -247,13 +247,46 @@ pub mod util {
         Some(Range::new(start, end))
     }
 
+    /// If the LS did not provide a range for the completion or the range of the
+    /// primary cursor can not be used for the secondary cursor, this function
+    /// can be used to find the completion range for a cursor
+    fn find_completion_range(text: RopeSlice, cursor: usize) -> (usize, usize) {
+        let start = cursor
+            - text
+                .chars_at(cursor)
+                .reversed()
+                .take_while(|ch| chars::char_is_word(*ch))
+                .count();
+        (start, cursor)
+    }
+    fn completion_range(
+        text: RopeSlice,
+        edit_offset: Option<(i128, i128)>,
+        cursor: usize,
+    ) -> Option<(usize, usize)> {
+        let res = match edit_offset {
+            Some((start_offset, end_offset)) => {
+                let start_offset = cursor as i128 + start_offset;
+                if start_offset < 0 {
+                    return None;
+                }
+                let end_offset = cursor as i128 + end_offset;
+                if end_offset > text.len_chars() as i128 {
+                    return None;
+                }
+                (start_offset as usize, end_offset as usize)
+            }
+            None => find_completion_range(text, cursor),
+        };
+        Some(res)
+    }
+
     /// Creates a [Transaction] from the [lsp::TextEdit] in a completion response.
     /// The transaction applies the edit to all cursors.
     pub fn generate_transaction_from_completion_edit(
         doc: &Rope,
         selection: &Selection,
-        start_offset: i128,
-        end_offset: i128,
+        edit_offset: Option<(i128, i128)>,
         new_text: String,
     ) -> Transaction {
         let replacement: Option<Tendril> = if new_text.is_empty() {
@@ -263,83 +296,163 @@ pub mod util {
         };
 
         let text = doc.slice(..);
+        let (removed_start, removed_end) =
+            completion_range(text, edit_offset, selection.primary().cursor(text))
+                .expect("transaction must be valid for primary selection");
+        let removed_text = text.slice(removed_start..removed_end);
 
-        Transaction::change_by_selection(doc, selection, |range| {
-            let cursor = range.cursor(text);
-            (
-                (cursor as i128 + start_offset) as usize,
-                (cursor as i128 + end_offset) as usize,
-                replacement.clone(),
-            )
-        })
+        let (transaction, mut selection) = Transaction::change_by_selection_ignore_overlapping(
+            doc,
+            selection,
+            |range| {
+                let cursor = range.cursor(text);
+                completion_range(text, edit_offset, cursor)
+                    .filter(|(start, end)| text.slice(start..end) == removed_text)
+                    .unwrap_or_else(|| find_completion_range(text, cursor))
+            },
+            |_, _| replacement.clone(),
+        );
+        if transaction.changes().is_empty() {
+            return transaction;
+        }
+        selection = selection.map(transaction.changes());
+        transaction.with_selection(selection)
     }
 
     /// Creates a [Transaction] from the [snippet::Snippet] in a completion response.
     /// The transaction applies the edit to all cursors.
+    #[allow(clippy::too_many_arguments)]
     pub fn generate_transaction_from_snippet(
         doc: &Rope,
         selection: &Selection,
-        start_offset: i128,
-        end_offset: i128,
+        edit_offset: Option<(i128, i128)>,
         snippet: snippet::Snippet,
         line_ending: &str,
         include_placeholder: bool,
+        tab_width: usize,
     ) -> Transaction {
         let text = doc.slice(..);
 
-        // For each cursor store offsets for the first tabstop
-        let mut cursor_tabstop_offsets = Vec::<SmallVec<[(i128, i128); 1]>>::new();
-        let transaction = Transaction::change_by_selection(doc, selection, |range| {
-            let cursor = range.cursor(text);
-            let replacement_start = (cursor as i128 + start_offset) as usize;
-            let replacement_end = (cursor as i128 + end_offset) as usize;
-            let newline_with_offset = format!(
-                "{line_ending}{blank:width$}",
-                line_ending = line_ending,
-                width = replacement_start - doc.line_to_char(doc.char_to_line(replacement_start)),
-                blank = ""
-            );
+        let mut off = 0i128;
+        let mut mapped_doc = doc.clone();
+        let mut selection_tabstops: SmallVec<[_; 1]> = SmallVec::new();
+        let (removed_start, removed_end) =
+            completion_range(text, edit_offset, selection.primary().cursor(text))
+                .expect("transaction must be valid for primary selection");
+        let removed_text = text.slice(removed_start..removed_end);
 
-            let (replacement, tabstops) =
-                snippet::render(&snippet, newline_with_offset, include_placeholder);
+        let (transaction, selection) = Transaction::change_by_selection_ignore_overlapping(
+            doc,
+            selection,
+            |range| {
+                let cursor = range.cursor(text);
+                completion_range(text, edit_offset, cursor)
+                    .filter(|(start, end)| text.slice(start..end) == removed_text)
+                    .unwrap_or_else(|| find_completion_range(text, cursor))
+            },
+            |replacement_start, replacement_end| {
+                let mapped_replacement_start = (replacement_start as i128 + off) as usize;
+                let mapped_replacement_end = (replacement_end as i128 + off) as usize;
 
-            let replacement_len = replacement.chars().count();
-            cursor_tabstop_offsets.push(
-                tabstops
-                    .first()
-                    .unwrap_or(&smallvec![(replacement_len, replacement_len)])
-                    .iter()
-                    .map(|(from, to)| -> (i128, i128) {
-                        (
-                            *from as i128 - replacement_len as i128,
-                            *to as i128 - replacement_len as i128,
-                        )
-                    })
-                    .collect(),
-            );
+                let line_idx = mapped_doc.char_to_line(mapped_replacement_start);
+                let pos_on_line = mapped_replacement_start - mapped_doc.line_to_char(line_idx);
 
-            (replacement_start, replacement_end, Some(replacement.into()))
-        });
+                // we only care about the actual offset here (not virtual text/softwrap)
+                // so it's ok to use the deprecated function here
+                #[allow(deprecated)]
+                let width = helix_core::visual_coords_at_pos(
+                    mapped_doc.line(line_idx),
+                    pos_on_line,
+                    tab_width,
+                )
+                .col;
+                let newline_with_offset = format!(
+                    "{line_ending}{blank:width$}",
+                    line_ending = line_ending,
+                    blank = ""
+                );
 
-        // Create new selection based on the cursor tabstop from above
-        let mut cursor_tabstop_offsets_iter = cursor_tabstop_offsets.iter();
-        let selection = selection
-            .clone()
-            .map(transaction.changes())
-            .transform_iter(|range| {
-                cursor_tabstop_offsets_iter
-                    .next()
-                    .unwrap()
-                    .iter()
-                    .map(move |(from, to)| {
-                        Range::new(
-                            (range.anchor as i128 + *from) as usize,
-                            (range.anchor as i128 + *to) as usize,
-                        )
-                    })
-            });
+                let (replacement, tabstops) =
+                    snippet::render(&snippet, &newline_with_offset, include_placeholder);
+                selection_tabstops.push((mapped_replacement_start, tabstops));
+                mapped_doc.remove(mapped_replacement_start..mapped_replacement_end);
+                mapped_doc.insert(mapped_replacement_start, &replacement);
+                off +=
+                    replacement_start as i128 - replacement_end as i128 + replacement.len() as i128;
 
-        transaction.with_selection(selection)
+                Some(replacement)
+            },
+        );
+
+        let changes = transaction.changes();
+        if changes.is_empty() {
+            return transaction;
+        }
+
+        let mut mapped_selection = SmallVec::with_capacity(selection.len());
+        let mut mapped_primary_idx = 0;
+        let primary_range = selection.primary();
+        for (range, (tabstop_anchor, tabstops)) in selection.into_iter().zip(selection_tabstops) {
+            if range == primary_range {
+                mapped_primary_idx = mapped_selection.len()
+            }
+
+            let range = range.map(changes);
+            let tabstops = tabstops.first().filter(|tabstops| !tabstops.is_empty());
+            let Some(tabstops) = tabstops else{
+                // no tabstop normal mapping
+                mapped_selection.push(range);
+                continue;
+            };
+
+            // expand the selection to cover the tabstop to retain the helix selection semantic
+            // the tabstop closest to the range simply replaces `head` while anchor remains in place
+            // the remaining tabstops receive their own single-width cursor
+            if range.head < range.anchor {
+                let first_tabstop = tabstop_anchor + tabstops[0].1;
+
+                // if selection is forward but was moved to the right it is
+                // contained entirely in the replacement text, just do a point
+                // selection (fallback below)
+                if range.anchor >= first_tabstop {
+                    let range = Range::new(range.anchor, first_tabstop);
+                    mapped_selection.push(range);
+                    let rem_tabstops = tabstops[1..]
+                        .iter()
+                        .map(|tabstop| Range::point(tabstop_anchor + tabstop.1));
+                    mapped_selection.extend(rem_tabstops);
+                    continue;
+                }
+            } else {
+                let last_idx = tabstops.len() - 1;
+                let last_tabstop = tabstop_anchor + tabstops[last_idx].1;
+
+                // if selection is forward but was moved to the right it is
+                // contained entirely in the replacement text, just do a point
+                // selection (fallback below)
+                if range.anchor <= last_tabstop {
+                    // we can't properly compute the the next grapheme
+                    // here because the transaction hasn't been applied yet
+                    // that is not a problem because the range gets grapheme aligned anyway
+                    // tough so just adding one will always cause head to be grapheme
+                    // aligned correctly when applied to the document
+                    let range = Range::new(range.anchor, last_tabstop + 1);
+                    mapped_selection.push(range);
+                    let rem_tabstops = tabstops[..last_idx]
+                        .iter()
+                        .map(|tabstop| Range::point(tabstop_anchor + tabstop.0));
+                    mapped_selection.extend(rem_tabstops);
+                    continue;
+                }
+            };
+
+            let tabstops = tabstops
+                .iter()
+                .map(|tabstop| Range::point(tabstop_anchor + tabstop.0));
+            mapped_selection.extend(tabstops);
+        }
+
+        transaction.with_selection(Selection::new(mapped_selection, mapped_primary_idx))
     }
 
     pub fn generate_transaction_from_edits(
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -250,18 +250,27 @@ pub mod util {
     /// If the LS did not provide a range for the completion or the range of the
     /// primary cursor can not be used for the secondary cursor, this function
     /// can be used to find the completion range for a cursor
-    fn find_completion_range(text: RopeSlice, cursor: usize) -> (usize, usize) {
+    fn find_completion_range(text: RopeSlice, replace_mode: bool, cursor: usize) -> (usize, usize) {
         let start = cursor
             - text
                 .chars_at(cursor)
                 .reversed()
                 .take_while(|ch| chars::char_is_word(*ch))
                 .count();
-        (start, cursor)
+        let mut end = cursor;
+        if replace_mode {
+            end += text
+                .chars_at(cursor)
+                .skip(1)
+                .take_while(|ch| chars::char_is_word(*ch))
+                .count();
+        }
+        (start, end)
     }
     fn completion_range(
         text: RopeSlice,
         edit_offset: Option<(i128, i128)>,
+        replace_mode: bool,
         cursor: usize,
     ) -> Option<(usize, usize)> {
         let res = match edit_offset {
@@ -276,7 +285,7 @@ pub mod util {
                 }
                 (start_offset as usize, end_offset as usize)
             }
-            None => find_completion_range(text, cursor),
+            None => find_completion_range(text, replace_mode, cursor),
         };
         Some(res)
     }
@@ -287,6 +296,7 @@ pub mod util {
         doc: &Rope,
         selection: &Selection,
         edit_offset: Option<(i128, i128)>,
+        replace_mode: bool,
         new_text: String,
     ) -> Transaction {
         let replacement: Option<Tendril> = if new_text.is_empty() {
@@ -296,9 +306,13 @@ pub mod util {
         };
 
         let text = doc.slice(..);
-        let (removed_start, removed_end) =
-            completion_range(text, edit_offset, selection.primary().cursor(text))
-                .expect("transaction must be valid for primary selection");
+        let (removed_start, removed_end) = completion_range(
+            text,
+            edit_offset,
+            replace_mode,
+            selection.primary().cursor(text),
+        )
+        .expect("transaction must be valid for primary selection");
         let removed_text = text.slice(removed_start..removed_end);
 
         let (transaction, mut selection) = Transaction::change_by_selection_ignore_overlapping(
@@ -306,9 +320,9 @@ pub mod util {
             selection,
             |range| {
                 let cursor = range.cursor(text);
-                completion_range(text, edit_offset, cursor)
+                completion_range(text, edit_offset, replace_mode, cursor)
                     .filter(|(start, end)| text.slice(start..end) == removed_text)
-                    .unwrap_or_else(|| find_completion_range(text, cursor))
+                    .unwrap_or_else(|| find_completion_range(text, replace_mode, cursor))
             },
             |_, _| replacement.clone(),
         );
@@ -326,6 +340,7 @@ pub mod util {
         doc: &Rope,
         selection: &Selection,
         edit_offset: Option<(i128, i128)>,
+        replace_mode: bool,
         snippet: snippet::Snippet,
         line_ending: &str,
         include_placeholder: bool,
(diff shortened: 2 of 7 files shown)


helix-core/src/transaction.rs
```diff
@@ -466,6 +466,42 @@ impl Transaction {
...
+
+        let (lower, upper) = changes.size_hint();
+        let size = upper.unwrap_or(lower);
+        let mut changeset = ChangeSet::with_capacity(2 * size + 1); // rough estimate
+
+        let mut last = 0;
+        for (from, to, tendril) in changes {
+            if last > from {
+                continue;
+            }
```
pascalkuthe, line 485: this is the only change from the `Transaction::change` function

helix-term/src/ui/completion.rs
```diff
@@ -107,13 +108,19 @@ impl Completion {
...
                 let transaction = if let Some(edit) = &item.text_edit {
                     let edit = match edit {
                         lsp::CompletionTextEdit::Edit(edit) => edit.clone(),
                         lsp::CompletionTextEdit::InsertAndReplace(item) => {
+                            let range = if replace_mode {
+                                item.replace
+                            } else {
+                                item.insert
+                            };
                             // TODO: support using "insert" instead of "replace" via user config
```
kirawi, line 122: Completed TODO

kirawi:
I don't think I understand the distinction between replace and insert for text edits. If, for example, the text is `Apple`, but the completion is `SomeAppleTree`, how would that be possible with an insert?

pascalkuthe:
> I don't think I understand the distinction between replace and insert for text edits. If, for example, the text is `Apple`, but the completion is `SomeAppleTree`, how would that be possible with a single insert?

The distinction is where the cursor is.
For example when editing `Rope|Graphemes` you start typing Build so the text is `RopeBuild|Grapemes` and it's auto completed to `RopeBuilder` with replace it replaces the rest of the word (or logicl identifier really. The whole point if having an LSP edit is that these replaces do not always start/end at word boundaries). So in replace mode the new text would be `RopeBuilder` and in insert mode `RopeBuilderGraphemes`.

You can look at #4380 and #5676 for more examples

StratilJakub:
I've been trying this branch for last hour or so, since autocompletion issues are quite frequent in large codebase (tried multiple `idle-timeout` values from 0 to 400).

This is helping with most of the issues resulting in wrong autocompletion, so thanks a lot for your work!

Unfortunately, issue mentioned in #1819 is still present in this branch -> Sometimes there is additional character(s) after confirming completion.

pascalkuthe:
> I've been trying this branch for last hour or so, since autocompletion issues are quite frequent in large codebase (tried multiple `idle-timeout` values from 0 to 400).
> 
> This is helping with most of the issues resulting in wrong autocompletion, so thanks a lot for your work!
> 
> Unfortunately, issue mentioned in #1819 is still present in this branch -> Sometimes there is additional character(s) after confirming completion.

Thanks for testing. This PR is not intended to fix the issues from #1819 but rather about other issues with completions. The synchronization bugs that #1819 addresses are very subtle and best addressed in a separate PR. I want to also take a look if I can either exactly understand #1819 and give it a thorough review or take that PR over so we can finally get those bugs fixed too. But any such work would be separate from this PR

StratilJakub:
Understand and I didn't necessarily expect this to fix everything, just wanted to report my experience and if maybe it helped with that also as a side effect. :)
Thanks again.

helix-view/src/editor.rs
```diff
@@ -249,6 +249,9 @@ pub struct Config {
     )]
     pub idle_timeout: Duration,
     pub completion_trigger_len: u8,
+    /// Whether to instruct the LSP to replace the entire word when applying a completion
+    /// or to only insert new text
+    pub completion_replace: bool,
```
poliorcetics, line 254: needs a book update too no ?

pascalkuthe: yeah that was an oversight, thanks

archseer:
With https://github.com/helix-editor/helix/pull/5864 merged this can now be rebased

pascalkuthe:
This has been rebased on #5864. I applied the fixes for the crashes described in the initial PR to the snippet transaction generation. In the process I have fixed multiple bugs with the way tabstops are handled. 

In particular this PR now uses a partial `ChangeSet` (temporarily created by temporarily append a `retain` to the in progress `ChangeSet`) to map the start of the replacement range to ensure tabstops always endup in the same place (otherwise the transaction may merge the replacement and the tabstops are merged incorrectly). 

This partial `ChangeSet` can also be applied to the doc if a snippet needs information about it's position in the document (which can not be determined without applying the `ChangeSet` right now). Right now this is used to determine the indentation level of a line but will also be important to correctly calculate the values of variables. This has some overhead which I want to avoid in the future (by essentially applying the transaction eagerly to the `rope`, storing the result in the transaction which then gets used in `apply` instead of recalculating) but that will introduce some complexity and the performance difference probably rarely matters in practice so I decided to leave that to future work.

Finally this PR brings tabstops inline with how selections work in helix in general. This is best demonstrated with an example.

Consider the following snippet:

```
fn ${1}(#{2}){
  ${0}
}
```

Now consider the following two selections (text between the two `|` is selected)

```
foo fn| bar|
```
```
|foo fn| bar
```

these two cases represent insert mode (`i`) and append mode (`a`).

if the snippet is applied to the two cases right now we endup with the following in both cases (just a point cursor):

``` rust
foo fn |(|){
  
} bar
```

with this PR it endsup as:

```
|foo fn |(){
  
} bar
```
and

```
foo fn| (){
  
} bar|
```

helix-core/src/transaction.rs
```diff
@@ -466,6 +484,51 @@ impl Transaction {
         self
     }
 
+    /// Generate a transaction from a set of potentially overallping changes.
```
archseer, line 487:
```suggestion
    /// Generate a transaction from a set of potentially overlapping changes.
```

helix-lsp/src/lib.rs
```diff
@@ -263,83 +307,140 @@ pub mod util {
...
+                let mapped_replacement_start = changeset.map_pos(replacement_start, Assoc::After);
+                // if a snippet needs access to it's position (line number or offset) we must apply
+                // previous edits to the rope, as this has some ovherhead we do it lazily
+                // TODO: can we avoid apply the previous edits so this becomes O(N) instead of
+                // O(NlogN) with the number of selection ranges?
+                let mapped_doc = Lazy::new(|| {
+                    let mut mapped_doc = doc.clone();
+                    changeset.apply(&mut mapped_doc);
+                    mapped_doc
+                });
```
andriigrynenko, line 363: This could probably be moved into change_by_selection_ignore_overlapping while still keeping it lazy (if the goal is to make Transaction object more powerful). Then this API can be updated to accept range and transaction_context, where transaction_context has a bunch of helper methods (e.g. getting current transformed doc, getting current changeset)

helix-lsp/src/lib.rs
```diff
@@ -247,13 +248,56 @@ pub mod util {
...
+
     /// Creates a [Transaction] from the [lsp::TextEdit] in a completion response.
     /// The transaction applies the edit to all cursors.
     pub fn generate_transaction_from_completion_edit(
         doc: &Rope,
         selection: &Selection,
-        start_offset: i128,
-        end_offset: i128,
+        start_offset: Option<i128>,
+        end_offset: Option<i128>,
```
andriigrynenko, line 300: It should be an Option of tuple? There probably won't be a case where only one is present but not the other

helix-lsp/src/lib.rs
```diff
@@ -263,83 +307,140 @@ pub mod util {
...
-                            (range.anchor as i128 + *to) as usize,
-                        )
-                    })
-            });
+                    .map(|tabstop| Range::point(tabstop_anchor + tabstop.1));
+                mapped_selection.extend(rem_tabstops);
+            } else {
+                let last_idx = tabstops.len() - 1;
+                let primary_tabstop = tabstop_anchor + tabstops[last_idx].1;
+                debug_assert!(primary_tabstop >= range.anchor);
```
Urgau, line 428: I tried this PR with rust-analyzer and no matter the snippet it crashed here!
```
thread 'main' panicked at 'assertion failed: primary_tabstop >= range.anchor', helix-lsp/src/lib.rs:428:17
```

pascalkuthe: Interesting, it's a debug assert so in a release build it wouldnt't crash but it may mess up the selection a little. Could you share what your selection looked like?

Urgau: Hope this helps.
```
2023-03-09T09:57:38.253 helix_term::ui::completion [DEBUG] snippet text: "_mm256_and_ps(${1:a}, ${2:b})$0"
2023-03-09T09:57:38.253 helix_lsp::util [DEBUG] cursor=64 replacement_start=60 replacement_end=64 Snippet {
    elements: [
        Text(
            "_mm256_and_ps(",
        ),
        Placeholder {
            tabstop: 1,
            value: [
                Text(
                    "a",
                ),
            ],
        },
        Text(
            ", ",
        ),
        Placeholder {
            tabstop: 2,
            value: [
                Text(
                    "b",
                ),
            ],
        },
        Text(
            ")",
        ),
        Tabstop {
            tabstop: 0,
        },
    ],
}
2023-03-09T09:57:38.254 helix_lsp::util [DEBUG] primary_idx=0 Transaction {
    changes: ChangeSet {
        changes: [
            Retain(
                60,
            ),
            Insert(
                "_mm256_and_ps(a, b)",
            ),
            Delete(
                4,
            ),
            Retain(
                151,
            ),
        ],
        len: 215,
        len_after: 230,
    },
    selection: None,
}
2023-03-09T09:57:38.254 helix_lsp::util [DEBUG] Range {
    anchor: 79,
    head: 80,
    old_visual_position: None,
}
```

pascalkuthe: ah interesting did you edit in append mode but without any selection? That's an edgecase I might not have considered and would fit this.

Definitly does help 👍

helix-lsp/src/lib.rs
```diff
@@ -263,83 +307,140 @@ pub mod util {
...
+                let cursor = range.cursor(text);
+                let (replacement_start, replacement_end) =
+                    completion_pos(text, start_offset, end_offset, cursor)?;
+                // Asssoc::After is used here because there should be no edits after
+                // `replacement_start` (those haven't been generated yet)
+                let mapped_replacement_start = changeset.map_pos(replacement_start, Assoc::After);
+                // if a snippet needs access to it's position (line number or offset) we must apply
+                // previous edits to the rope, as this has some ovherhead we do it lazily
+                // TODO: can we avoid apply the previous edits so this becomes O(N) instead of
+                // O(NlogN) with the number of selection ranges?
```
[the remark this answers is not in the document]
pascalkuthe: Pretty sure it doesn't, because applying changeet is O(N) but the cahgneset only has the length of the already processed changes so during the iteration you are applying an O(N) N=1, O(N), N=2,,, O(N) N=M transaction. That's a classical example of an O(MlogM) algorithm. Still not great and just head an idea to optimize this further so I might be able to make it O(N)

helix-lsp/src/lib.rs
pascalkuthe: Yeah that was mostly a lazy change, guess I could make it a tuple 👍

helix-lsp/src/lib.rs
Urgau: I'm definitely in Insert mode. And the range printed is the one just here. (the `debug!` is placed just before the `debug_assert!`).

pascalkuthe: Hmm mapping ranges should never change their direction, curious what's going on here. Will see if I can reprosucez thanks for testing!

Urgau: This should help you reproduce the issue.
Steps for reproducing:

- `cargo new --lib "hx-test"` (just a new empty project)
- `cd "hx-test"`
- `[..]/target/debug/hx -vv src/lib.rs`
- go to 2nd line and remove "right"
- then type "m256"
- then arrow down and crash!

helix-core/src/transaction.rs
```diff
@@ -513,6 +576,43 @@ impl Transaction {
...
+            |changset| {
+                selection.next().map(|range| {
+                    let (change, data) = create_change(range, changset)?;
+                    Some((change, (range, data)))
+                })
+            },
+            process_change,
+        );
+        // map the transaction trough changes
+        (transaction, new_primary_idx)
```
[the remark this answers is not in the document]
andriigrynenko: Why is it lossy?

helix-lsp/src/lib.rs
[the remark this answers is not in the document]
andriigrynenko: 1 + 2 + ... + M = M^2 / 2 + M / 2 :)

helix-lsp/src/lib.rs
andriigrynenko: > we try to keep the core transaction API as minimal as possible

This PR is already adding significant complexity to the transaction code. If you are optimizing for keeping the transaction code simpler, it may make sense to move all of this into the snippet rendering code (e.g. by creating a new transaction for each range and then just combining them).

helix-core/src/transaction.rs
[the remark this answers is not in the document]
pascalkuthe: a selection is sorted and overlapping ranges are merged on creation. In this case it's fine tough because the selection is already a valid selection so removing some ranges won't change that so I did change back to a selection afterall

helix-lsp/src/lib.rs
pascalkuthe: > > we try to keep the core transaction API as minimal as possible
> 
> This PR is already adding significant complexity to the transaction code and API. If you are optimizing for keeping the transaction code simpler, it may make sense to move all of this into the snippet rendering code (e.g. by creating a new transaction for each range and then just combining them).

especially with your suggestions above the API became quite minimal. Maybe minimal isn't even the right word. The API is supposed to be genric and not specical cased to one usecase. I don't want this general transaction logic inside the snippet/completion code either.

helix-lsp/src/lib.rs
pascalkuthe: fixed it, I didn't handle the case where the selection was fully incased inside the removed text. Thanks for the pointers

helix-core/src/transaction.rs
```diff
@@ -513,6 +576,43 @@ impl Transaction {
         Self::change(doc, selection.iter().map(f))
     }
 
+    pub fn change_by_selection_ignore_overlapping<T>(
+        doc: &Rope,
+        selection: &Selection,
+        mut create_change: impl FnMut(&Range, &ChangeSet) -> Option<(Change, T)>,
```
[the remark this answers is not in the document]
pascalkuthe: still needed context here but this did improve the API

helix-lsp/src/lib.rs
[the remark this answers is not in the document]
pascalkuthe: the new rope is generated right inside the snippets now eagerly the overhead should be quite small now. Also made the code simpler.

helix-lsp/src/lib.rs
archseer: I had similar concerns but the newest commits look a bit more straightforward 👍🏻

helix-core/src/transaction.rs
```diff
@@ -513,6 +542,44 @@ impl Transaction {
...
+            selection.iter().enumerate().map(|range| {
+                let (change_start, change_end) = change_range(range.1);
+                (change_start, change_end, range)
+            }),
+            process_change,
+        );
+
+        (
+            transaction,
+            Selection::new(ranges, new_primary_idx.unwrap_or(0)),
```
andriigrynenko, line 579: Both callers of change_by_selection_ignore_overlapping end up applying transaction to the selection and using that resulting selection (generate_transaction_from_completion_edit does that directly, generate_transaction_from_snippet applies changes to each range). Couldn't you just apply transaction to selection here and attach that resulting selection to the transaction?

pascalkuthe: The selection is changed again in both cases and there is no way to remove the selection fron a transaction (and even if there were it would require an unwrap). 

Applying the selection to the transaction would also be a footgun/semantically incorret because the returned selection is not mapped trough changes yet (and we can't do that inside this function since mapping can cause ranges to be sorted/merged which would mess up the zip for snippets), instead this selection is identical to to inout aelection with only those ranges whose edits have been dropped removed.

andriigrynenko: > and we can't do that inside this function since mapping can cause ranges to be sorted/merged which would mess up the zip for snippets

I was suggesting returning the mapped selection (sorry if I didn't use the right terminology and made it confusing). I think the sorting/merging aspect is what I missed.

Issue #3364, opened 2022-08-09
Feature request: selective undo


notusknot:
A simple but very nice to have feature would be selective undo, as found in Emacs. A user can choose to only undo the selected area, which saves a lot of time and energy. I don’t know much about the Helix internals, but if someone could point out where the undo is located, I’d be happy to give this feature a shot myself.

Issue #701, opened 2021-09-03
Typing text offsets diagnostics that are after text typed


Omnikar:
https://asciinema.org/a/8zyizvoNTUMkDhhzeaS10qw0W

I presume this is probably because of char indexing for the diagnostics.

archseer:
I did initially map diagnostics through edits to prevent this (updating the positions), but I noticed that rust-analyzer would unhelpfully re-send the wrong offsets immediately, undoing the work.

archseer:
(It'll fix the positions after you save and it re-runs the checks)

archseer:
Commented out code:

https://github.com/helix-editor/helix/blob/7e1123680f474bff5113db189f63ca7f948781a5/helix-view/src/document.rs#L638-L646

kirawi:
Is there anything that can be done about it?

archseer:
Turns out this is a known rust-analyzer issue: https://github.com/rust-analyzer/rust-analyzer/issues/8836

gabydd:
correct me if I am wrong but I think this is fixed on master

the-mikedavis:
I think this still happens unless it was fixed in a recent rust-analyzer version. From what I understand this is caused by rust-analyzer sending out-of-date/cached ranges in publishDiagnostic which undoes the work we do within helix for updating ranges https://github.com/rust-lang/rust-analyzer/issues/10628#issuecomment-950904152

PR #5736, opened 2023-01-30
Add command for resetting diff hunks

### LINKED ISSUE #4974: Add command to reset diff hunks
Suggested to me by @Byron

It would be nice to have a command/keybinding to reset hunks (like `git reset --hard` but more fine grained).
This functionality is diff provider independent and can be implemented by just removing the `after` lines in the current document and replacing the with the `before` lines in the diff base.

I would implement this as a type able command (maybe `:reset` ) that resets the hunk at the current cursor. Ideally there would be a way to quickly place multi-cursors inside hunks so that multiple hunks can be selected at once.

pascalkuthe:
Closes #4974

Adds the `:reset-diff-change` command  (aliases `:diffg` and `:diffget`) that resets a hunk back
to its original text. 

This change requires exposing the `diff_base` (the file that a diff was computed against). To accommodate that I slightly refactored the terminology the `DiffHandle` and renamed the `FileHunks` struct to `Diff` as it doesn't just contains hunks anymore.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1953,6 +1953,64 @@ fn run_shell_command(
     Ok(())
 }
 
+fn reset_diff_change(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    ensure!(args.is_empty(), ":reset-diff-change takes no arguments");
+
+    let editor = &mut cx.editor;
+    let scrolloff = editor.config().scrolloff;
+
+    let (view, doc) = current!(editor);
+    // TODO refactor to use let..else once MSRV is raised to 1.65
+    let handle = match doc.diff_handle() {
+        Some(handle) => handle,
+        None => bail!("Diff is not available in the current buffer"),
+    };
+
+    let diff = handle.load();
+    let doc_text = doc.text().slice(..);
+    let line = doc.selection(view.id).primary().cursor_line(doc_text);
+
+    // TODO refactor to use let..else once MSRV is raised to 1.65
+    let hunk_idx = match diff.hunk_at(line as u32, true) {
+        Some(hunk_idx) => hunk_idx,
+        None => bail!("There is no change at the cursor"),
+    };
+    let hunk = diff.nth_hunk(hunk_idx);
+    let diff_base = diff.diff_base();
+    let before_start = diff_base.line_to_char(hunk.before.start as usize);
+    let before_end = diff_base.line_to_char(hunk.before.end as usize);
+    let text: Tendril = diff
+        .diff_base()
+        .slice(before_start..before_end)
+        .chunks()
+        .collect();
+    let anchor = doc_text.line_to_char(hunk.after.start as usize);
+    let transaction = Transaction::change(
+        doc.text(),
+        [(
+            anchor,
+            doc_text.line_to_char(hunk.after.end as usize),
+            (!text.is_empty()).then_some(text),
+        )]
+        .into_iter(),
+    );
+    drop(diff); // make borrow check happy
+    doc.apply(&transaction, view.id);
+    // select inserted text
+    let text_len = before_end - before_start;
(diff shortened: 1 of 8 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1822,6 +1822,60 @@ fn run_shell_command(
...
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    ensure!(args.is_empty(), ":reset-change takes no arguments");
+
+    let editor = &mut cx.editor;
+    let scrolloff = editor.config().scrolloff;
+
+    let (view, doc) = current!(editor);
+    if let Some(handle) = doc.diff_handle() {
```
the-mikedavis, line 1839: This isn't a super big block but the indentation here could be reduced:

```rs
let handle = match doc.diff_handle() {
    Some(handle) => handle,
    None => bail!("Diff is not available in the current buffer"),
};

// ...
```

and the same for the next `if let Some`

pascalkuthe: that does look cleaner, thanks! This is actually an ideal case for `let ... else` but sadly we haven't reached a MSRV of 1.65 yet. Added a `TODO` for that

trink:
The diagnostics get confused by the reset. It just pushed the diagnostic down and I have to save the file to clear it. Although this isn't specific to reset, it is just easier to demonstrate here.

Peek 2023-01-30 11-39

pascalkuthe:
> The diagnostics get confused by the reset. It just pushed the diagnostic down and I have to save the file to clear it. Although this isn't specific to reset, it is just easier to demonstrate here.
> 
> Peek 2023-01-30 11-39 [ Peek 2023-01-30 11-39 ](https://user-images.githubusercontent.com/795988/215578823-d608aacb-72e0-42d6-b631-a7db017f799a.gif) [ ](https://user-images.githubusercontent.com/795988/215578823-d608aacb-72e0-42d6-b631-a7db017f799a.gif)

That's just #701 and happens for any edit not just the `:reset-change` command here (also the case on master) so unrelated to this PR.

helix-term/src/commands/typed.rs
```diff
@@ -2337,6 +2395,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             fun: run_shell_command,
             completer: Some(completers::directory),
         },
+       TypableCommand {
+            name: "reset-change",
+            aliases: &["reset"],
+            doc: "Undo the change at cursor position",
+            fun: reset_change,
+            completer: None,
+        },
```
the-mikedavis, line 2404: I'm not sure where to put this but I think either the name or description (or both) should say "VCS". Otherwise this might be confused with https://github.com/helix-editor/helix/issues/3364

pascalkuthe: The reason I shy away from using `vcs` in diff related functionality is that these commands would also be used for the diffmode where we might not actually be comparing against a VCS file but rather just a different buffer/file on disk. 

I would have used something like `undo-selection` for #3364. So any command that goes back in undo history would contain `undo` and `reset` would be for the diff. I can see that the names could be potentially confusing tough. This is sort of a bikeshed issue.


For now I just changed the description so that it uses `reset` instead of `undo` (to keep terminology straight) and explicitly refer to a`diff change`. Do you think that is enough or would you rather see the name changed.
Some options I came up wiith:

* `:reset-change`: What I choose initially for this PR.
* `:reset-diff-change`: Probably most concise but a bit wordy
* `:reset-diff`: Sounds like the entire diff is reset
* `:diffget` (alias `:diffg`): Name stolen from nvim. The nice thing is that it has a cousing `:diffput` to reset the other side of the diff if we are comparing to editable files (not currently implemented)

pascalkuthe: Thanks for your input. I renamed to command to `:reset-diff-change`.
I decided to add `:diffget` and `:diffg` as aliases. These names are quite well know from other editors, and I don't think they are not easily confused with something else.
I personally don't bind this to a keybinding so I also appreciate being able to type `:diffg` as a convenience (similar to how we have `:w` as an alias for `:write`).

I also think that it will make it clearer that this command is closely linked to a future command that changes the other side of the disk (which would alias to `:diffput`)

the-mikedavis (dismissed):

PR #5738, opened 2023-01-30
Fix utf8 length handling for shellwords

trink:
If the last argument to shellwords ends in a multibyte utf8 character the entire argument will be dropped.
e.g. `:sh echo test1 test2𒀀` will only output `test1`

### DIFF
--- a/helix-core/src/shellwords.rs
+++ b/helix-core/src/shellwords.rs
@@ -129,8 +129,9 @@ impl<'a> From<&'a str> for Shellwords<'a> {
                 DquoteEscaped => Dquoted,
             };
 
-            if i >= input.len() - 1 && end == 0 {
-                end = i + 1;
+            let c_len = c.len_utf8();
+            if i == input.len() - c_len && end == 0 {
+                end = i + c_len;
             }
 
             if end > 0 {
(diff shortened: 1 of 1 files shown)


grasshorse:
I see it... thanks.

pascalkuthe (approved):
Nice catch! Mixed up utf-8 and utf-32 again :D

helix-core/src/shellwords.rs
```diff
@@ -333,4 +334,9 @@ mod test {
         assert_eq!(Shellwords::from(":o a").parts(), &[":o", "a"]);
         assert_eq!(Shellwords::from(":o a\\ ").parts(), &[":o", "a\\"]);
     }
+
+    #[test]
+    fn test_multibyte_at_end() {
+        assert_eq!(Shellwords::from("𒀀").parts(), &["𒀀"]);
```
the-mikedavis, line 340: I liked your examples in the PR description, let's add cases for them to make sure we don't have regressions in the future 👍

```suggestion
        assert_eq!(Shellwords::from("𒀀").parts(), &["𒀀"]);
        assert_eq!(Shellwords::from(":sh echo 𒀀").parts(),  &[":sh", "echo", "𒀀"]);
        assert_eq!(Shellwords::from(":sh echo hello𒀀").parts(),  &[":sh", "echo", "hello𒀀"]);
```

trink: I added the change as a separate commit as it is generally cleaner to review (as opposed to an amend). I can squash them if you want.  What is the general preference for this project?

pascalkuthe: reviewers generally squash themselves for small PRs like this if they find it necessary.  I don't think it matters much either way,

the-mikedavis:
(Force pushed without changes to trigger CI. I'm not sure why it didn't run the first time 🤔)

the-mikedavis (approved):

Issue #2210, opened 2022-04-21
Minimap and Scrollbar


kuanyinglu:
Heard about Helix in the past, but it's been a while so today I looked at all the progress made. I'm really starting to like this editor! One thing I didn't find in my quick search is minimap and scrollbar. I don't really use minimap, but I know some people like them. 

I'm really just looking looking forward to a scroll bar with features like this plugin in Neovim. 
https://github.com/petertriho/nvim-scrollbar
It highlights a rough location of where diagnostics and searches. One feature I find myself using a lot in other editors is looking for words under cursor, and think this could be really useful.

Wondering what people think about this feature.

David-Else:
I would like a scroll bar just to see where I am in the file and the LSP errors, I would not want it to be used with a mouse like a GUI application! No need for a minimap, that is bloat.

I am a bit worried about feature creep happening in Helix, it would be really bad if it started to turn into VS Code! :) Nearly every feature I ever wanted is already a pull request, just waiting for those precious nuggets of code to drop one by one...

the-mikedavis:
Scrollbar seems doable and I would argue belongs in the core. Minimap is probably very hard to accomplish without a GUI though and I don't know how I would feel about it being in the core.

slinlee:
> Scrollbar seems doable and I would argue belongs in the core. Minimap is probably very hard to accomplish without a GUI though and I don't know how I would feel about it being in the core.

I personally don't use minimaps, but here's one I saw for vim https://reposhub.com/rust/command-line/wfxr-minimap-vim.html which in turn uses a rust proj https://github.com/wfxr/code-minimap

kirawi:
There was also a discussion on this a while ago: https://matrix.to/#/!zMuVRxoqjyxyjSEBXc:matrix.org/$_yZ9RkhI4rNEbJxsdZ9WhYTNvPiCdsUqts-rQb5yEkQ?via=matrix.org&via=tchncs.de&via=mozilla.org

pickfire:
Would be interesting to have a scrollbar + minimap (with the colored find), but I feel like it is kinda wasteful for the space on terminals like 80x24 (I still do that).

cor:
I feel like minimaps would be quite wasteful, but scrollbars (especially if they work with the mouse) would be greatly appreciated, if only to get a sense of where you are in the document.

cor:
Now that we have a scrollbar for the `<space>k` popups, it should be pretty simple to add this to the main editor right?

[...]

PR #5749, opened 2023-01-31
Adds support for right-hand gutters and Scrollbar gutter item

jakesactualface:
Relates to https://github.com/helix-editor/helix/issues/2210

First time contributor, so please let me know if I'm missing anything!

## Overview
- Adds support for gutter items rendered on the right-hand side of the view
- Adds support for a new "scrollbar" gutter item
  - Note: this item is not added to the default configuration for the left- or right-hand gutter

## Architectural Changes
- A `gutters-right` configuration option is exposed as a sibling to the existing `gutters` option
- `View` now accepts an additional `gutters_right` parameter as a `GutterConfig` object
- Gutter item functions are now called for each visible line in the view, rather than stopping at the last line of document contents
  - See `last_line` initialization within `render_gutter_item` for more context
- A `Scrollbar` variant was added to the `GutterType` enum

## Design feedback needed
This `gutters` vs. `gutters-right` implementation seemed like the easiest approach (and was also passive to the existing config file setup), but I quite like how the `editor.statusline` configuration is subdivided into `left`, `center`, and `right` partitions.

- [ ] Should we wrap gutter configuration into similar `left` and `right` partitions?

### DIFF
--- a/helix-view/src/gutter.rs
+++ b/helix-view/src/gutter.rs
@@ -32,6 +32,7 @@ impl GutterType {
             GutterType::LineNumbers => line_numbers(editor, doc, view, theme, is_focused),
             GutterType::Spacer => padding(editor, doc, view, theme, is_focused),
             GutterType::Diff => diff(editor, doc, view, theme, is_focused),
+            GutterType::Scrollbar => scrollbar(editor, doc, view, theme, is_focused),
         }
     }
 
@@ -41,6 +42,7 @@ impl GutterType {
             GutterType::LineNumbers => line_numbers_width(view, doc),
             GutterType::Spacer => 1,
             GutterType::Diff => 1,
+            GutterType::Scrollbar => 1,
         }
     }
 }
@@ -137,6 +139,44 @@ pub fn diff<'doc>(
     } else {
         Box::new(move |_, _, _, _| None)
     }
+}
+
+pub fn scrollbar<'doc>(
+    _editor: &'doc Editor,
+    doc: &'doc Document,
+    view: &View,
+    theme: &Theme,
+    is_focused: bool,
+) -> GutterFn<'doc> {
+    let total_lines = doc.text().len_lines();
+    let view_height = view.inner_height();
+    let fits = view_height > total_lines;
+
+    if !is_focused || fits {
+        return Box::new(move |_, _, _, _| None);
+    }
+
+    let scroll_height = view_height.pow(2).div_ceil(total_lines).min(view_height);
+    let visual_pos = view.pos_at_visual_coords(doc, 0, 0, false).unwrap();
+    let view_vertical_offset = doc.text().char_to_line(visual_pos);
+
+    let scroll_line = ((view_height - scroll_height) * view_vertical_offset
+        / 1.max(total_lines.saturating_sub(view_height)))
+        + view_vertical_offset;
+
+    let style = theme.get("ui.menu.scroll");
+    Box::new(
+        move |line: usize, _selected: bool, _first_visual_line: bool, out: &mut String| {
+            let icon = if line >= scroll_line && line <= scroll_line + scroll_height {
+                "▐"
+            } else {
+                ""
+            };
+
+            write!(out, "{}", icon).unwrap();
+            Some(style)
+        },
+    )
 }
 
 pub fn line_numbers<'doc>(
--- a/helix-view/src/gutter.rs
+++ b/helix-view/src/gutter.rs
@@ -338,7 +338,11 @@ mod tests {
 
     #[test]
     fn test_default_gutter_widths() {
-        let mut view = View::new(DocumentId::default(), GutterConfig::default());
+        let mut view = View::new(
+            DocumentId::default(),
+            GutterConfig::default(),
+            GutterConfig::from(Vec::new()),
+        );
         view.area = Rect::new(40, 40, 40, 40);
 
         let rope = Rope::from_str("abc\n\tdef");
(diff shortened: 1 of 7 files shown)


pascalkuthe:
This  is not really your fault but there is currently a larger change to editor rendering underway in #5420.
This PR heavily conflicts with #5420 and needs to be rebased on that once that is merged. #5420 fundamentally changes how gutters are rendered (or really all editor ui) and especially the following change will be problematical:
> Gutter item functions are now called for each visible line in the view, rather than stopping at the last line of document contents

jakesactualface:
@pascalkuthe no worries! I'm fine with rebasing onto your changes. I gave it a try preemptively based on the current contents of your PR and it seems like the right gutter is still rendering correctly, though I'll have to adjust the math for the scrollbar position.

In the meantime, should I convert this PR to a draft until #5420 is merged? Or is there a different way you'd like me to proceed?

pascalkuthe:
It would have been fine to leave the PR as is, we have a label that we put in PRs that are based on other.

Howbwr #5420 has been merged into master today so you can just rebase off master

jakesactualface:
> It would have been fine to leave the PR as is

Sorry, I meant that I performed a quick rebase _locally_, just to see how big of a conflict I was in store for.

I still have some cleanup and validation to do before I'm ready to push that rebase here, but I'll try to get that done within the next few days. Thank you for the heads-up!

jakesactualface:
I have rebased my changes onto a fresh copy of `master`, as well as divided them up between 3 commits. This was done to give some logical separation between the main ideas being attempted:
1. Add support for a gutter on the right-hand side of the view
2. Add support for a "scrollbar" gutter item
3. Enable gutter rendering for lines within the viewport but beyond EOF

The changes in Point 3 above (localized for now in the 3rd commit on this PR) could definitely benefit from some more-experienced eyes than my own, if that functionality is going to be considered for merge. I'm still fairly new to Rust and I suspect there's a better way for me to be tracking the viewport's position during the text rendering iteration.

I've included some screenshots below for an overview of how this branch looks in use, in case that makes things easier for anyone interested.

### Line numbers enabled for right-hand gutter (left-hand gutter is explicitly set to default layout):
image

### Right-hand gutter set to `"scrollbar", "spacer", "line-numbers"`:
image

### Same file and setup, with the viewport scrolled beyond EOF:
image

helix-view/src/gutter.rs
```diff
@@ -139,6 +141,47 @@ pub fn diff<'doc>(
...
+        view_vertical_offset as f32 / total_lines.saturating_sub(view_height) as f32;
+
+    let start_line = view_vertical_offset
+        .saturating_add((scroll_percentage * scrollable_view_height as f32).ceil() as usize);
+
+    let style = theme.get("ui.menu.scroll");
+    Box::new(
+        move |line: usize, _selected: bool, _first_visual_line: bool, out: &mut String| {
+            let icon = if line >= start_line && line <= start_line.saturating_add(height) {
+                "█"
```
sudormrfbin, line 174: We already use the half block character for the completion popup scrollbar, so we could use the same here for consistency:

https://github.com/helix-editor/helix/blob/15e07d4db893aec7b9e117c1f88400a68ba98ae2/helix-term/src/ui/popup.rs#L277-L281

jakesactualface: I originally shied away from that since there's nothing stopping someone from putting the scrollbar on the left, but it looks like the `diff` gutter is making a similar assumption and it looks better for it.

I'll add the change as suggested, but I'm open to discussing it if others have a different opinion.

Axlefublr:
I would love to use this PR! However, there are currently some merge conflicts with master. Can you please rebase?

jakesactualface:
> I would love to use this PR! However, there are currently some merge conflicts with master. Can you please rebase?

Rebase has been pushed. I also took the opportunity to refactor some of the scrollbar calculations to better match the popup window logic.

My current plan is to leave this PR open in case it gains interest. However, I think there's discussion to be had regarding whether something like a scrollbar should be more of a "first-class" functionality vs. packaging it into the gutters mechanism. As it stands, this seems more like a stopgap than a true enhancement.

Of course, I'll still welcome any feedback or suggestions in the meantime!

Axlefublr:
wonderful thank you! when I merge this in my fork and try it later, I'll share my thoughts on whether it feels wacky or solid! using the gutters mechanism at least doesn't *sound* like it would be wacky though?

Axlefublr:
I found a bug. If `gutters.layout` is empty, but `gutters-right.layout` isn't, `gutters-right` are still not displayed. As soon as you have at least one element in the left gutter though, the right one starts displaying

jakesactualface:
> I found a bug. If `gutters.layout` is empty, but `gutters-right.layout` isn't, `gutters-right` are still not displayed. As soon as you have at least one element in the left gutter though, the right one starts displaying

Hey, thanks for catching that! The issue should be fixed now.

lovehumans:
@pascalkuthe can we have an update on whether this PR will be merged soon? I'm doing the tutorial of helix atm and sorely missing an indication of how long that file is :)

Rudxain:
@lovehumans [this `config`](https://github.com/helix-editor/helix/blob/14eda106f0a3e6a5fc6fb5cbd96bda9774f64ae1/book/src/editor.md?plain=1#L125) is enough in the meantime:
```toml
[editor.statusline]
right = [
	"diagnostics",
	"selections",
	"register",
	"position-percentage",
	"position",
	"file-encoding",
]
```

If the Scheme-based config-system (#8853) works as I expect, it should be possible to simply insert `"position-percentage"` into the default list

PR #5751, opened 2023-01-31
Do not add intermediate lines to jumplist with :<linenum> command.

askreet:
Based on [this discussion](https://github.com/helix-editor/helix/discussions/5741). (@the-mikedavis)

This patch:
* Only appends to the jumplist when the `:<linenum>` invocation is accepted with <kbd>Enter</kbd> (`PromptEvent::Validate`).
* Fixes what appears to be a bug in the jumplist behavior where the `current` pointer points past the end of the list when it's appended to, resulting in a "dead" invocation of `C-o`. This caused issues with the previous implementation especially, as it would append to the list _twice_ implicitly during validate (since the movement logic previous lived outside the handler for the `PromptEvent::Update` event.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2715,10 +2715,15 @@ fn push_jump(view: &mut View, doc: &Document) {
 }
 
 fn goto_line(cx: &mut Context) {
-    goto_line_impl(cx.editor, cx.count)
+    if cx.count.is_some() {
+        goto_line_without_jumplist(cx.editor, cx.count);
+
+        let (view, doc) = current!(cx.editor);
+        push_jump(view, doc);
+    }
 }
 
-fn goto_line_impl(editor: &mut Editor, count: Option<NonZeroUsize>) {
+fn goto_line_without_jumplist(editor: &mut Editor, count: Option<NonZeroUsize>) {
     if let Some(count) = count {
         let (view, doc) = current!(editor);
         let text = doc.text().slice(..);
@@ -2735,7 +2740,6 @@ fn goto_line_impl(editor: &mut Editor, count: Option<NonZeroUsize>) {
             .clone()
             .transform(|range| range.put_cursor(text, pos, editor.mode == Mode::Select));
 
-        push_jump(view, doc);
         doc.set_selection(view.id, selection);
     }
 }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4313,18 +4313,15 @@ fn jump_forward(cx: &mut Context) {
 }
 
 fn jump_backward(cx: &mut Context) {
-    jump_backward_impl(cx.editor, cx.count());
-}
-
-fn jump_backward_impl(editor: &mut Editor, count: usize) {
-    let config = editor.config();
-    let (view, doc) = current!(editor);
+    let count = cx.count();
+    let config = cx.editor.config();
+    let (view, doc) = current!(cx.editor);
     let doc_id = doc.id();
 
     if let Some((id, selection)) = view.jumps.backward(view.id, doc, count) {
         view.doc = *id;
         let selection = selection.clone();
-        let (view, doc) = current!(editor); // refetch doc
+        let (view, doc) = current!(cx.editor); // refetch doc
 
         if doc.id() != doc_id {
             view.add_to_history(doc_id);
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2716,10 +2716,10 @@ fn push_jump(view: &mut View, doc: &Document) {
 
 fn goto_line(cx: &mut Context) {
     if cx.count.is_some() {
-        goto_line_without_jumplist(cx.editor, cx.count);
-
         let (view, doc) = current!(cx.editor);
         push_jump(view, doc);
+
+        goto_line_without_jumplist(cx.editor, cx.count);
     }
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4313,15 +4313,18 @@ fn jump_forward(cx: &mut Context) {
 }
 
 fn jump_backward(cx: &mut Context) {
-    let count = cx.count();
-    let config = cx.editor.config();
-    let (view, doc) = current!(cx.editor);
+    jump_backward_impl(cx.editor, cx.count());
+}
+
+fn jump_backward_impl(editor: &mut Editor, count: usize) {
+    let config = editor.config();
+    let (view, doc) = current!(editor);
     let doc_id = doc.id();
 
     if let Some((id, selection)) = view.jumps.backward(view.id, doc, count) {
         view.doc = *id;
         let selection = selection.clone();
-        let (view, doc) = current!(cx.editor); // refetch doc
+        let (view, doc) = current!(editor); // refetch doc
 
         if doc.id() != doc_id {
             view.add_to_history(doc_id);
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1412,15 +1412,9 @@ fn tutor(
 
 fn undo_goto_line_number_preview(cx: &mut compositor::Context) {
     if cx.editor.goto_line_number_preview {
-        log::info!("undoing goto_line_number preview, jumping backwards");
-        // if let Some(line_number) = cx.editor.last_line_number {
-        // goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line_number));
-
         // Remove the jump we added during the preview session.
         jump_backward_impl(cx.editor, 1);
 
-        // let (view, doc) = current!(cx.editor);
-        // view.ensure_cursor_in_view(doc, line_number);
         cx.editor.goto_line_number_preview = false;
     }
 }
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1420,6 +1420,25 @@ fn abort_goto_line_number_preview(cx: &mut compositor::Context) {
     }
 }
 
+fn update_goto_line_number_preview(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+) -> anyhow::Result<()> {
+    cx.editor.last_selection.get_or_insert_with(|| {
+        let (view, doc) = current!(cx.editor);
+        doc.selection(view.id).clone()
+    });
+
+    let scrolloff = cx.editor.config().scrolloff;
+    let line = args[0].parse::<usize>()?;
+    goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line));
+
+    let (view, doc) = current!(cx.editor);
+    view.ensure_cursor_in_view(doc, scrolloff);
+
+    Ok(())
+}
+
 pub(super) fn goto_line_number(
     cx: &mut compositor::Context,
     args: &[Cow<str>],
@@ -1430,31 +1449,24 @@ pub(super) fn goto_line_number(
         PromptEvent::Validate => {
             ensure!(!args.is_empty(), "Line number required");
 
-            if let Some(last_selection) = &cx.editor.last_selection {
+            // If we are invoked directly via a keybinding, Validate is
+            // sent without any prior Update events. Ensure the cursor
+            // is moved to the appropriate location.
+            update_goto_line_number_preview(cx, args)?;
+
+            if let Some(last_selection) = cx.editor.last_selection.take() {
                 let (view, doc) = current!(cx.editor);
                 view.jumps.push((doc.id(), last_selection.clone()));
-
-                cx.editor.last_selection = None;
             }
         }
+
         // When a user hits backspace and there are no numbers left,
         // we can bring them back to their original selection. If they
         // begin typing numbers again, we'll start a new preview session.
         PromptEvent::Update if args.is_empty() => abort_goto_line_number_preview(cx),
-        PromptEvent::Update => {
-            cx.editor.last_selection.get_or_insert_with(|| {
-                let (view, doc) = current!(cx.editor);
-                doc.selection(view.id).clone()
-            });
-
-            let scrolloff = cx.editor.config().scrolloff;
-            let line = args[0].parse::<usize>()?;
-            goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line));
-
-            let (view, doc) = current!(cx.editor);
-            view.ensure_cursor_in_view(doc, scrolloff);
-        }
+        PromptEvent::Update => update_goto_line_number_preview(cx, args)?,
     }
+
     Ok(())
 }
 
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1418,7 +1418,7 @@ pub(super) fn goto_line_number(
     match event {
         PromptEvent::Abort => {
             if let Some(line_number) = cx.editor.last_line_number {
-                goto_line_impl(cx.editor, NonZeroUsize::new(line_number));
+                goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line_number));
                 let (view, doc) = current!(cx.editor);
                 view.ensure_cursor_in_view(doc, line_number);
                 cx.editor.last_line_number = None;
@@ -1427,6 +1427,10 @@ pub(super) fn goto_line_number(
         }
         PromptEvent::Validate => {
             ensure!(!args.is_empty(), "Line number required");
+
+            let (view, doc) = current!(cx.editor);
+            push_jump(view, doc);
+
             cx.editor.last_line_number = None;
         }
         PromptEvent::Update => {
@@ -1434,7 +1438,7 @@ pub(super) fn goto_line_number(
                 if let Some(line_number) = cx.editor.last_line_number {
                     // When a user hits backspace and there are no numbers left,
                     // we can bring them back to their original line
-                    goto_line_impl(cx.editor, NonZeroUsize::new(line_number));
+                    goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line_number));
                     let (view, doc) = current!(cx.editor);
                     view.ensure_cursor_in_view(doc, line_number);
                     cx.editor.last_line_number = None;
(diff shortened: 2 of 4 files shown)


pascalkuthe (changes requested):
One small nit so this becomes easier to review.

I am not quite sure if the second fix is correct. Looking trough the code it seems intentional that `current` may point one past the end (the backward function explicitly checks for that) but I am not sure

helix-view/src/view.rs
```diff
@@ -21,18 +21,21 @@ type Jump = (DocumentId, Selection);
 #[derive(Debug, Clone)]
 pub struct JumpList {
     jumps: VecDeque<Jump>,
-    current: usize,
+    current_idx: usize,
```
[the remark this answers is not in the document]
askreet: I may just back the jumplist changes out for now, I don't feel 100% confident in them and I wouldn't mind adding some tests. Will push a commit up later today.

helix-term/src/commands/typed.rs
```diff
@@ -1445,12 +1449,13 @@ pub(super) fn goto_line_number(
             let text = doc.text().slice(..);
             let line = doc.selection(view.id).primary().cursor_line(text);
             cx.editor.last_line_number.get_or_insert(line + 1);
+
+            let line = args[0].parse::<usize>()?;
+            goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line));
+            let (view, doc) = current!(cx.editor);
```
trink, line 1455: This line can be removed: view and doc are already in scope

helix-term/src/commands/typed.rs
```diff
@@ -1445,12 +1449,13 @@ pub(super) fn goto_line_number(
             let text = doc.text().slice(..);
             let line = doc.selection(view.id).primary().cursor_line(text);
             cx.editor.last_line_number.get_or_insert(line + 1);
```
trink, line 1451: It might be a good time to refactor this too. The line number is currently computed with every update but it is only needed when last_line_number is None.  Using `get_or_insert_with(|doc, view| ...`  instead, it can be computed when needed.

trink (changes requested):

helix-term/src/commands/typed.rs
askreet: I juggled some pieces around here to address your comments, but the borrow rules here make it tricky. Any feedback is appreciated.

askreet:
I removed the jumplist changes but now it's in an interesting state where the jumplist works as expected, unless you rewind and then play back the jumplist to the end. If you do that, then doing another `:<linenum>` invocation replaces the tail of the jumplist. I'll do some more debugging later.

helix-term/src/commands/typed.rs
pascalkuthe: You usage is correct `cx.editor` is passed `goto_line_without_jumplist` mutably so any references into `cx.editor` (like `view` and `doc`) are invalidated and need to be recreated. This pattern is commonly used in other commands so you did the right thing

pascalkuthe:
> I removed the jumplist changes but now it's in an interesting state where the jumplist works as expected, unless you rewind and then play back the jumplist to the end. If you do that, then doing another `:<linenum>` invocation replaces the tail of the jumplist. I'll do some more debugging later.

You are onto something and that triggered me to take a hard look/think about the jumplist code (something I haven't looked at in too much detail so far).

It works as follows:

* push is called during a normal jump **before** the jump occurs so the location to jump pack to is saved. Current is placed one past the end here so that if you go backwards by one you end up at the the last saved jump
* If we jump backwards and are one past the end then the current location is pushed to the top of the jumplist so we van rewind back to it with forward
* You can not go one past the end with forward because what was previously one past the end is the positioning previously saved during backwards (forwarding past the end doesn't make sense since there is no jump saved)

That means you previous fix was indeed wrong and the code is working as intended. The rope as on for the odd behaviour you are seeing is that `push_junp` is being called at the wrong place here. It should be called **before** the new selection is set. If you look at the old implementation `push_jump` is called before `doc.set"selection`.So now you need to call `push_jump` before `goto_line_without_jumplist` is called.

We only ever save locations to junp back to and `current` points at the current position which we don't ever want to jump back to (unless we go forward) so it makes sense that push would overwrite that as it's essentially starying a new history from the cureent point. I think what is a bit confusing is the jumplist get initalized with a single jump. In my opinion initializing it as empty would be more correct as that single jump is always overwritten and never actually has an effext. That's only a mater if code clarity tough and does not need to be addressed here.

So for this PR you just need to move `push_jump` before `goto_line_without_jumplist` is called.

helix-term/src/commands/typed.rs
trink: > I juggled some pieces around here to address your comments, but the borrow rules here make it tricky. Any feedback is appreciated.

Sorry, I should have noticed that or tried it and the borrow checker would have complained to me instead.

askreet:
Thanks @pascalkuthe for the clarification, that does make sense. I think the code clarity could be improved and may try to rationalize about that in another PR, but will fix this one for now. Thanks again!

askreet:
Alright I had an interesting thought about this whole thing this morning. Since we want to set a jumplist entry before moving the cursor to allow the user to navigate back to where they started and we need support for aborting the command, can we not just use the jumplist to accomplish both tasks?

The updated commit does the following:
* Track whether we are previewing a goto_line_number command within Editor, instead of tracking the last line number.
* Push a jump when we start previewing the command.
* Call jump_backward if the command is aborted.

This has the added benefit of retaining not only the starting line number, but the starting selection(s) as well. As far as I can tell, it works ideally when invoked with `:goto <linenum>`. However, the [special handling](https://github.com/helix-editor/helix/blob/master/helix-term/src/commands/typed.rs#L2423-L2428) of `:<linenum>` creates some edge cases:

* If the line entry is cleared by typing `:12` followed by two backspaces, no abort event is sent to `goto_line_number` to clear the flag.
* If the line entry becomes invalid by typing, for example, `:12 abc` it stops sending events to `goto_line_number` and the flag is not cleared.

I have some ideas about how to resolve these, but am pretty new to the project so greatly interested in others' thoughts. My ideas are:

* Naïvely clear the flag when command mode is invoked (in `command_mode`) with a comment.
* If unmatched typed input (e.g. empty string or `:12 ab`) is found in `command_mode`'s closure and the flag is set, send the Abort event to `goto_line_number`.

Architecturally I have an idea that might improve this, too. What if typed commands were trait objects that has their own state and tracked in the command mode handler as an `Option<TypedCommand>`? Whenever an active typed command session was in progress but the command no longer matched its invocation pattern we could reason about having to abort that typed command in order to restore the Option to None and allow another typed command session to start. The added benefit here is there are a few flags on Editor that exist for tracking state of a typed command session that could be stored directly on the struct that implements TypedCommand. I can move this idea to a discussion if anyone thinks it has merit.

the-mikedavis:
Pushing to the jumplist is potentially destructive: if you jump backwards with C-o a few times you can hit C-i to move forward again in the jumplist. But jump backwards with C-o and then push to the jumplist, the old history going forward is lost. Ideally we should only ever use the jumplist for real jumps rather than re-using it for implementation details like this.

I think the `Option<dyn TypedCommand>` sounds needlessly more complicated. Most typable commands don't need to store any internal state anyways and we try to minimize the use of traits in general since they bring a lot of complexity.

pascalkuthe:
> Pushing to the jumplist is potentially destructive: if you jump backwards with C-o a few times you can hit C-i to move forward again in the jumplist. But jump backwards with C-o and then push to the jumplist, the old history going forward is lost. Ideally we should only ever use the jumplist for real jumps rather than re-using it for implementation details like this.
> 
> I think the `Option<dyn TypedCommand>` sounds needlessly more complicated. Most typable commands don't need to store any internal state anyways and we try to minimize the use of traits in general since they bring a lot of complexity.

I agree I preferred the previous simpler fix just with the push_jump moved to the right place as I sugessted.
I would avoids messing with the jumplist which should really only be used for jumps that actually occur and never for previews.

If you want to restore the selection when aborting why not just store the previous selection on the editor instead of the previous line number (like we do now). 
I think that might be a nice change  in general if you call it `last_selection` as that field could be reused for other typable commands that preview the location we jump to which are not line based.

askreet:
> Pushing to the jumplist is potentially destructive: if you jump backwards with C-o a few times you can hit C-i to move forward again in the jumplist. But jump backwards with C-o and then push to the jumplist, the old history going forward is lost.

Ah, good point. Oh well, it was a fun exercise to learn more of the code :-).

> I think the Option<dyn TypedCommand> sounds needlessly more complicated. Most typable commands don't need to store any internal state

Fair point that there aren't many implementations that require this, I suppose I'm imagining it being generally useful for future commands. More or less complicated might be a matter of taste, I found it surprising that state was stashed in the editor struct disjoint from context of the typed command and the typed commands source file.

> I agree I preferred the previous simpler fix just with the push_jump moved to the right place as I sugessted.

I don't remember the details but it ended up being a bit more complicated. I'll add some context when I work on the next commit if I can remember it. I think your suggestion below might address it.

> If you want to restore the selection when aborting why not just store the previous selection on the editor instead of the previous line number (like we do now).

I like this idea, I'll pursue that next.

helix-term/src/commands/typed.rs
```diff
@@ -1410,47 +1410,51 @@ fn tutor(
...
+
+            let scrolloff = cx.editor.config().scrolloff;
+            let line = args[0].parse::<usize>()?;
+            goto_line_without_jumplist(cx.editor, NonZeroUsize::new(line));
+
             let (view, doc) = current!(cx.editor);
-            let text = doc.text().slice(..);
-            let line = doc.selection(view.id).primary().cursor_line(text);
-            cx.editor.last_line_number.get_or_insert(line + 1);
+            view.ensure_cursor_in_view(doc, scrolloff);
```
askreet, line 1455: Stumbled on a neat "bug" in this last pass where we were passing `line_number` as the scrolloff value of `ensure_cursor_in_view`. Fixed in the last iteration.

pascalkuthe: that could cause very weird issues. Nice catch

askreet:
New iteration is ready for review, appreciate all the pointers on this @the-mikedavis @pascalkuthe @trink!

I still think there's some edge cases around the flow of aborting the implicit `:<linenum>` syntax, but they weren't as obvious as with the jumplist approach (since the temporary jumplist entry would be 'leaked').

helix-term/src/commands.rs
```diff
@@ -2715,10 +2715,15 @@ fn push_jump(view: &mut View, doc: &Document) {
 }
 
 fn goto_line(cx: &mut Context) {
-    goto_line_impl(cx.editor, cx.count)
+    if cx.count.is_some() {
+        goto_line_without_jumplist(cx.editor, cx.count);
+
+        let (view, doc) = current!(cx.editor);
+        push_jump(view, doc);
```
pascalkuthe, line 2722: `push_jump` is still being called before `goto_line_without_jumplist`. This will messup the jumplist. Simply changing this to:

```suggestion
        let (view, doc) = current!(cx.editor);
        push_jump(view, doc);
        
        goto_line_without_jumplist(cx.editor, cx.count);
```

should work

helix-term/src/commands/typed.rs
```diff
@@ -1410,47 +1410,51 @@ fn tutor(
...
-            return Ok(());
-        }
+        PromptEvent::Abort => abort_goto_line_number_preview(cx),
         PromptEvent::Validate => {
             ensure!(!args.is_empty(), "Line number required");
-            cx.editor.last_line_number = None;
+
+            if let Some(last_selection) = &cx.editor.last_selection {
+                let (view, doc) = current!(cx.editor);
+                view.jumps.push((doc.id(), last_selection.clone()));
```
pascalkuthe, line 1435: good idea about reusing `last_selection` here to get around the fact that we might currently be at a different preview :+1:

helix-term/src/commands/typed.rs
```diff
@@ -1410,47 +1410,51 @@ fn tutor(
...
-                let (view, doc) = current!(cx.editor);
-                view.ensure_cursor_in_view(doc, line_number);
-                cx.editor.last_line_number = None;
-            }
-            return Ok(());
-        }
+        PromptEvent::Abort => abort_goto_line_number_preview(cx),
         PromptEvent::Validate => {
             ensure!(!args.is_empty(), "Line number required");
-            cx.editor.last_line_number = None;
```
pascalkuthe, line 1430: I am pretty sure that `PromptEvent::Validate` can happen immidietly without reciving a `PromptEvent::Update`. In that case the command does nothing now wheras it worked correctly before. I think you can restructure the funciton slightly to account for that. First check if we are aborting => call `abort_goto_line_number_preview` and return.

Then run the code that is currently in the update branch.

The check if we are validating and run the code here. `last_selection` can then be obtained with `cx.editor.last_selection.clone().unwrap()` (as it's always initalized a few lines prior)

pascalkuthe (changes requested):
Going in the right direction :+1: The `last_selection` turned out to be an even better idea then I thaught :D There are still some edgecases left tough, I left comments about those.

helix-term/src/commands/typed.rs
```diff
@@ -1410,47 +1410,51 @@ fn tutor(
...
+
+            if let Some(last_selection) = &cx.editor.last_selection {
+                let (view, doc) = current!(cx.editor);
+                view.jumps.push((doc.id(), last_selection.clone()));
+
+                cx.editor.last_selection = None;
```
pascalkuthe, line 1437: One more nit: Reading and clearing an option is what `Option::take` is for. So as per my comment above you can use `cx.editor.last_selection.take().unwrap()`

helix-term/src/commands/typed.rs
askreet: > I am pretty sure that PromptEvent::Validate can happen immidietly without reciving a PromptEvent::Update.

How would that happen? Keybinding?

helix-term/src/commands.rs
askreet: Ah, good catch. I forgot I had updated this as well.

helix-term/src/commands/typed.rs
pascalkuthe: I think keybindings may do that. Haven't tested it but all other typed commands work when imminently called with ` PromptEvent::Validate` so I wouldn't want to diverge from that here.

helix-term/src/commands/typed.rs
askreet: Ah, thanks for the tip -- updated.

helix-term/src/commands/typed.rs
askreet: Re-reading your comment:

> First check if we are aborting => call abort_goto_line_number_preview and return.

What do you mean by "first check"? The function only handles one event at a time, and it does call the abort function and return then an abort event is fired.

pascalkuthe: Right now you do a match on the event:
``` rust 
match event{
PromptEvent::Abort => ..,
...
}
```

Instead use a simple if:

```
if event == PromptEvent::Abort{
// abort ...
return
}

// update selection

if event == PromptEvent::Validate{
// jumplist entry and clear last_selection
}
```

askreet: Sorry, but I don't understand the difference, both constructs result in the function returning after calling abort, no?

pascalkuthe: The difference is that you don't need to copy all the update code into the validate branch.

With a match the full thing would be:
``` rust 
match event{
PromptEvent::Abort => /* abort */,
PromptEvent::Update => /* update selection  */,
PromptEvent::Validate => /* update selection, add jumplist entry and clear last_selection */,
...
}
```

I suggested the structure above so you can share the update code between the `Update` and the `Validate` branch

askreet: Okay, I see what you're saying. I'm trying to figure out the case where validate is called with arguments without update having been called. If I can find that, I'll restructure the code accordingly.

askreet: @pascalkuthe You were, of course, exactly right -- binding a key in insert mode to `:goto 123` sends only the Validate event, which currently doesn't work. New commit addresses that. Thanks for catching that and the detailed explanation!

helix-term/src/commands/typed.rs
```diff
@@ -1410,47 +1410,63 @@ fn tutor(
...
-                    cx.editor.last_line_number = None;
-                }
-                return Ok(());
+
+            // If we are invoked directly via a keybinding, Validate is
+            // sent without any prior Update events. Ensure the cursor
+            // is moved to the appropriate location.
+            update_goto_line_number_preview(cx, args)?;
+
+            if let Some(last_selection) = cx.editor.last_selection.take() {
```
pascalkuthe, line 1457: Now that `last_selection` is always populated in `update_goto_line_number_preview` you  should call `unwrap` here instead of `if let`. If `last_selection` somehow became `None` in the meantime that is a bug and therefore correct to panic rather than silently ignoring the problem. Fail hard and fail fast...

pascalkuthe (approved):
Just one last nit, Apart from this, this LGTM now. Nice improvement both fixing the jumplist bug and making sure that we rust the full selection instead of just placing the primary cursor back on the original line.

helix-term/src/commands/typed.rs
askreet: > Fail hard and fail fast...

Interesting. I did consider this code path and had the opposite instinct. In the event of a code change that results in a bug I didn't want to crash and potentially lose a user's unsaved editor changes just because a jumplist entry couldn't be set. Is this approach to error handling standard for the project, and if so, documented for contributors?

Alternatively, we could use either a debug_assert! or else conditional with logging if you're concerned about detecting such a bug.

pascalkuthe: you can `bail` too in the else case if you are super concerned about the `else` case causing data loss but  If you  look throughout the codebase, `unwrap` and `expect` are used liberally for impossible code-paths. There was some discussion about adding a panic handler that backups  unsaved work. Rust autoamtically panics on many common programmar bugs (out of bounds indexing, etc.) so that problem is not specific to this one case.

Right now the code looks like it's perfectly valid for there to be no `last_selection` when it should really be impossible. That is an anitpattern to me.. Using unwrap for code-paths that should never happens is a pretty common idiom in rust and something that makes code easier to read (which we care about) and makes it easier to find bugs.

askreet: No problem, happy to be consistent with the rest of the project. Pushed a fix for that and the clippy warnings.

PR #1792, opened 2022-03-11
Resolve conflicts between prompt/picker bindings

crodjer:
Currently, the picker's re-using a few bindings which are also present
in the prompt. This causes some editing behaviours to not function on
the picker.

**Ctrl + k** and **Ctrl + j**
This should kill till the end of the line on prompt, but is overridden
by the picker for scrolling. Since there are redundancies (`Ctrl + p`,
`Ctrl + n`), we can remove it from picker.

**Ctrl + f** and **Ctrl + b**
This are used by the prompt for back/forward movement. We could modify
it to be Ctrl + d and Ctrl + u, to match the `vim` behaviour.

PR #5752, opened 2023-01-31
Add C-j and C-k key bindings to Picker

vapourismo:
Adds C-j and C-k bindings to the Picker components in order to align with the menu (e.g. auto completion) bindings that also let you navigate with those keys.


the-mikedavis (changes requested):
See See https://github.com/helix-editor/helix/pull/1792

vapourismo:
Ah right, I missed that C-k is already taken. Too bad, j and k feel much more natural.

PR #5318, opened 2022-12-27
Add a status line element that shows just the basename of the file

theteachr:
Makes the `FileName` status line element render just the `basename` of the open file. The existing variant that renders the relative path of the open file is renamed to `FilePath` (render functions are also renamed accordingly).

# Configuration Snippets

```toml
[editor.statusline]
left = ["mode", "spinner", "file-name"]
```

image

```toml
[editor.statusline]
left = ["mode", "spinner", "file-path"]
```

image

PR #5769, opened 2023-02-01
Dry up rendering code

theteachr:
Cleans up the duplication that was introduced by https://github.com/helix-editor/helix/pull/5318.

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -421,10 +421,7 @@ where
         .map(|p| p.to_string_lossy())
         .unwrap_or_else(|| SCRATCH_BUFFER_NAME.into());
 
-    let title = format!(
-        " {} ",
-        path,
-    );
+    let title = format!(" {} ", path,);
 
     write(context, title, None)
 }
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -13,6 +13,8 @@ use helix_view::editor::StatusLineElement as StatusLineElementID;
 use tui::buffer::Buffer as Surface;
 use tui::text::{Span, Spans};
 
+use std::path::PathBuf;
+
 pub struct RenderContext<'a> {
     pub editor: &'a Editor,
     pub doc: &'a Document,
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/statusline.rs
```diff
@@ -407,44 +409,41 @@ where
     write(context, format!(" {} ", file_type), None);
 }
 
-fn render_file_name<F>(context: &mut RenderContext, write: F)
+fn render_title<F>(context: &mut RenderContext, write: F, path: Option<PathBuf>)
```
the-mikedavis, line 412: "title" doesn't feel like the right word here since this function is about rendering paths. Maybe `render_file_name_impl` (`*_impl` is already a common pattern in the codebase) or `render_path_name` or just `render_path`?

theteachr: The second commit actually changes it from `render_path `to `render_title`. I’ll be able to revert it once I can access my system. But until then, I’ll try to think of a better name, as the function isn’t just rendering the path but also the modification status of the buffer.

theteachr:
Is `render_buffer_title` a better name?

the-mikedavis:
I still like `render_path`. I don't see the file path as the title of the buffer (I'm probably biased from the implementation 😅)

Actually we might want to take https://github.com/helix-editor/helix/pull/4731 instead of this: separating the file modification indicator into its own element reduces the duplication between these two statusline elements and fixes some UI issues

theteachr:
> separating the file modification indicator into its own element

I was planning to work on this next (thanks for the heads up about https://github.com/helix-editor/helix/pull/4731), as my first priority was to remove the duplication I'd caused.

helix-term/src/ui/statusline.rs
```diff
@@ -421,10 +421,7 @@ where
         .map(|p| p.to_string_lossy())
         .unwrap_or_else(|| SCRATCH_BUFFER_NAME.into());
 
-    let title = format!(
-        " {} ",
-        path,
-    );
+    let title = format!(" {} ", path,);
```
theteachr: The formatter should have removed the trailing comma when it decided to make this a one liner.

PR #5770, opened 2023-02-01
feature: add a `toggle-option` command

divarvel:
This commands allows to toggle boolean options, with a dedicated completer that only suggests boolean options
It does not support other enumerated types.

> :toggle softwrap.enable

Alternative options:

> :set !softwrap.enable # ok as well (as long as options cannot start with a !), consistent with vim, i think
> :set softwrap.enable ! # possibly ambiguous

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2438,7 +2438,7 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             aliases: &["toggle"],
             doc: "Toggle a boolean config option at runtime.\nFor example to toggle smart case search, use `:toggle search.smart-case`.",
             fun: toggle_option,
-            completer: Some(completers::setting),
+            completer: Some(completers::boolean_setting),
         },
         TypableCommand {
             name: "get-option",
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -312,15 +312,20 @@ pub mod completers {
     }
 
     /// Recursive function to get all keys from this value and add them to vec
-    fn get_keys(value: &serde_json::Value, vec: &mut Vec<String>, scope: Option<&str>) {
+    fn get_keys(
+        value: &serde_json::Value,
+        vec: &mut Vec<String>,
+        scope: Option<&str>,
+        only_booleans: bool,
+    ) {
         if let Some(map) = value.as_object() {
             for (key, value) in map.iter() {
                 let key = match scope {
                     Some(scope) => format!("{}.{}", scope, key),
                     None => key.clone(),
                 };
-                get_keys(value, vec, Some(&key));
-                if !value.is_object() {
+                get_keys(value, vec, Some(&key), only_booleans);
+                if !value.is_object() && (!only_booleans || value.is_boolean()) {
                     vec.push(key);
                 }
             }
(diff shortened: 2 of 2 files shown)


pascalkuthe (approved):
Implementation LGTM. I think this feature certainly makes sense so users can bind it to their keys. I originally considered a special `:set` syntax for this but I think a separate command is easier to implement, more discoverable and works well.

helix-term/src/commands/typed.rs
```diff
@@ -1479,6 +1479,42 @@ fn get_option(
...
 
+/// Change config at runtime. Access nested values by dot syntax, for
+/// example to toggle smart case search, use `:toggle search.smart-case`.
+fn toggle_option(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
```
Proful, line 1490: any reason for using ```return Ok(());``` instead of `Ok()`

divarvel: using `return` avoids wrapping the rest of the function in an `else` block.
also, it's a matter of being consistent with `set_option` which has the same structure.

helix-term/src/commands/typed.rs
```diff
@@ -2246,6 +2282,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             fun: set_option,
             completer: Some(completers::setting),
         },
+        TypableCommand {
+            name: "toggle-option",
+            aliases: &["toggle"],
+            doc: "Toggle a config option at runtime.\nFor example to toggle smart case search, use `:toggle search.smart-case`.",
+            fun: toggle_option,
+            completer: Some(completers::setting),
```
clo4, line 2290: How possible would it be to only complete _boolean_ setting names instead of all of them?
BTW, this is fantastic, I would love to see this added.

pascalkuthe: Good idea, should be pretty trivial, all you would need to do is use `value.is_boolean()` in the `get_keys` function in `helix-term/ui/mod.rs` instead of `!value.is_object()`

divarvel: yeah, i'll have a shot

divarvel: done in a new commit

helix-term/src/ui/mod.rs
```diff
@@ -314,15 +314,20 @@ pub mod completers {
     }
 
     /// Recursive function to get all keys from this value and add them to vec
-    fn get_keys(value: &serde_json::Value, vec: &mut Vec<String>, scope: Option<&str>) {
+    fn get_keys(
+        value: &serde_json::Value,
+        vec: &mut Vec<String>,
+        scope: Option<&str>,
+        only_booleans: bool,
```
divarvel, line 321: i chose to defunctionalize here for simplicity purposes, but a predicate function could be passed as well.

helix-term/src/ui/mod.rs
```diff
@@ -333,7 +338,29 @@ pub mod completers {
...
+            .collect();
+
+        matches.sort_unstable_by_key(|(_file, score)| Reverse(*score));
+        matches
+            .into_iter()
+            .map(|(name, _)| ((0..), name.into()))
+            .collect()
+    }
+
+    pub fn boolean_setting(_editor: &Editor, input: &str) -> Vec<Completion> {
```
divarvel, line 359: i would have preferred adding a helper function used by both `setting` and `boolean_setting`, but the `static KEYS` can't depend on a function parameter.

pascalkuthe: you just have two `static` in the `setting` function:

``` rust
static KEYS
```

and 

``` rust
static KEYS_BOOL
```

and dispatch which of those two to use with a simple `if`. Not a perfect solution either but would reduce code duplication slightly. I don't imagine that we would add more specialized settings keys in the future as more advanced dynamic configuration is better achieved with a future plugin system. Until then hardocind the bool keys seems like a good enough solution.

divarvel: done in a separate commit, i had to duplicate the `.iter()` call as well

poliorcetics:
#4085 has been merged, is this a duplicate ?

poliorcetics:
Ah, it's a partial duplicate: in #4085, the completer proposes all keys, not only boolean ones

divarvel:
Heh, I'll rebase and rename the PR accordingly

divarvel:
Amusing how my changes were close to #4085 without being aware of it.

I've only kept the commit adding dedicated completion.

pascalkuthe:
Actually it seems like @archseer wants to merge #4411 too which allows cycling arbitrary settings by providing a list of available values. In that case we actually want to keep the completion for any settings. For example `:toggle bufferline always multiple"`

I wasn't aware of those two PRs when you started working on this originally and somehow endedup missing them while searching. That was my bad! Sorry for wasting your time.

divarvel:
ah, support for enumerated values was next on my list. i waited before working on it, thankfully ^^

no worries, i'll close this one

PR #5773, opened 2023-02-01
Fix URL in doc chapter about Textobject queries

jdevuyst:


the-mikedavis:
The new link needs GitHub's "New Code Search and Code View (Beta)" feature preview to be enabled to work. I don't think we should merge this until that feature preview is enabled by default.

the-mikedavis (approved):

pascalkuthe (approved):

PR #5778, opened 2023-02-02
Add Darcula Solid theme

jesusmgg:
Based on the original Darcula, with darker and more contrasting colors.
image

Original repository: https://github.com/jesusmgg/darcula-solid-helix

### DIFF
--- a/runtime/themes/darcula-solid.toml
+++ b/runtime/themes/darcula-solid.toml
@@ -1,102 +1,19 @@
-# Author : Jesus Gonzalez <jesus@jesusmg.net>
 # Original source and more info: https://github.com/jesusmgg/darcula-solid-helix
 
-"ui.background" = { bg = "grey01" }
+inherits = "darcula"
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Does this overlap with Darcula much (changing the palette and a few scopes)? If so it would be good to re-use most of that theme with theme inheritance: https://docs.helix-editor.com/master/themes.html#inheritance

See the `monokai_aqua` theme for an example

jesusmgg:
> Does this overlap with Darcula much (changing the palette and a few scopes)? If so it would be good to re-use most of that theme with theme inheritance: https://docs.helix-editor.com/master/themes.html#inheritance
> 
> See the `monokai_aqua` theme for an example

Makes sense, let me look into it and update.

jesusmgg:
Now inheriting the original `darcula` theme as much as possible and apply changes over it.

PR #5787, opened 2023-02-02
pyi is a python file too (ERT=1min)

FrancescElies:
Adds `pyi` to python file types


the-mikedavis:
This and few other extensions were added already in #5593: https://github.com/helix-editor/helix/blob/62d046fa219b927c536bf6726fcae1e825346e0e/languages.toml#L472-L476

FrancescElies:
I had an old master, I should have pulled before creating a pr 😬

PR #5790, opened 2023-02-02
Ensure history is completed and cursor is in view after pasting

### LINKED ISSUE #5733: Helix crashed on "Undo change" (u) [undo]
### Summary

I had an opened JS file in helix. I accidentally pasted some text from the system clipboard several times by making a three-finger gesture swiping from right to left (the pasting action is probably due to me using tmux). I pressed 'u' to undo it and it said "Already at oldest change". After pressing Escape, I could undo the change but after pressing it a few times, Helix crashed with the error message presented in the Helix log.  I may have tried writing the file with :w (don't remember exactly).

Additionally, I have the JavaScript language server running, and auto-format using prettier.

### Reproduction Steps

 

I tried this:

1. `tmux`
2. `hx file.js`
3. paste something form the system's clipboard.
4. press `u`
5. press `<Esc>`
6. press `u` a few times
7. press `:w`

I expected this to happen:

Changes are undone and the file is saved with formatting

Instead, this happened:

The editor has crashed.

### Helix log

No related logs were recorded but the program crashed with the following error outputed to the terminal:
```
thread 'main' panicked at 'assertion failed: original_doc.len_chars() == self.len', helix-core/src/transaction.rs:265:9
```

### Platform

Linux (Fedora 37)

### Terminal Emulator

alacritty 0.10.1

### Helix Version

helix 22.12

pascalkuthe:
Fixes #5733

This PR ensures that the cursor is in view and that any changes are appended to history when using typeable commands instead of keybindings. If changes are not saved to history crashes can occur later. For example the reproduction case for #5733 uses middle mouse to trigger the `:clipboard-paste-before` command which causes the undo command to produce a crash later. 

While looking into a fix, I noticed that `ensure_cursor_in_view` was also not being called. When the cursor is at the EOF (`vgegl`) of a large file and replaced with a very short string from the clipboard the view currently becomes empty (no visible text or line numbers). That second bug is fixed by adding the `ensure_cursor_in_view` call.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3734,6 +3734,7 @@ fn paste_impl(
     }
 
     doc.apply(&transaction, view.id);
+    doc.append_changes_to_history(view);
 }
 
 pub(crate) fn paste_bracketed_value(cx: &mut Context, contents: String) {
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -913,6 +913,7 @@ fn replace_selections_with_clipboard_impl(
     cx: &mut compositor::Context,
     clipboard_type: ClipboardType,
 ) -> anyhow::Result<()> {
+    let scrolloff = cx.editor.config().scrolloff;
     let (view, doc) = current!(cx.editor);
 
     match cx.editor.clipboard_provider.get_contents(clipboard_type) {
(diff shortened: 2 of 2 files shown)


archseer (approved):

pascalkuthe:
sorry about the pipline I accidently added an uesless `&` so clippy complained. Pushed a fix for that

Issue #5754, opened 2023-01-31
Add configuration for MSBuild files


adam-becker:
Currently, opening any MSBuild project file results in the default `text` language being used. At the least, the `xml` language should be used to gain proper highlighting. The file extensions I'm aware of are `.csproj`, `.fsproj`, `.vbproj`, and `.proj`.

It is, of course possible to work around this with `:lang xml` each time you open the file, and maybe other ways. However, I believe helix should provide this out of the box.

sudormrfbin:
A PR for this would be definitely welcome, you can consult https://docs.helix-editor.com/master/languages.html on adding a new language entry. Re-using the `xml` grammar is a good starting point, and it can replaced with a dedicated tree-sitter grammar if it becomes available.

adam-becker:
Related for XML support: https://github.com/eclipse/lemminx

This could bridge the gap for MSBuild project files, since a schema exists for them. I'm not sure if the schema is meant to be inferred or how that works, but that could be helpful.

PR #5793, opened 2023-02-02
Add MSBuild language based on XML grammar

brettlyons:
Filetype reference: https://stackoverflow.com/questions/2007689/is-there-a-standard-file-extension-for-msbuild-files

#5754

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2117,3 +2117,23 @@ roots = []
 comment-token = "#"
 indent = { tab-width = 4, unit = "    " }
 grammar = "python"
+
+[[language]]
+name = "MSBuild"
+scope = "source.xml"
+injection-regex = "xml"
+file-types = [".proj", ".vbproj", ".csproj", ".fsproj", ".targets", ".props"]
+indent = { tab-width = 2, unit = "  " }
+roots = []
+
+[language.auto-pairs]
+'(' = ')'
+'{' = '}'
+'[' = ']'
+'"' = '"'
+"'" = "'"
+"<" = ">"
+
+[[grammar]]
+name = "MSBuild"
+source = { git = "https://github.com/RenjiSann/tree-sitter-xml", rev = "48a7c2b6fb9d515577e115e6788937e837815651" }
--- a/languages.toml
+++ b/languages.toml
@@ -2119,12 +2119,13 @@ indent = { tab-width = 4, unit = "    " }
 grammar = "python"
 
 [[language]]
-name = "MSBuild"
-scope = "source.xml"
-injection-regex = "xml"
-file-types = [".proj", ".vbproj", ".csproj", ".fsproj", ".targets", ".props"]
+name = "msbuild"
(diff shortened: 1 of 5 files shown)


languages.toml
```diff
@@ -2117,3 +2117,23 @@ roots = []
 comment-token = "#"
 indent = { tab-width = 4, unit = "    " }
 grammar = "python"
+
+[[language]]
+name = "MSBuild"
```
the-mikedavis, line 2122:
```suggestion
name = "msbuild"
```

languages.toml
```diff
@@ -2117,3 +2117,23 @@ roots = []
...
+scope = "source.xml"
+injection-regex = "xml"
```
the-mikedavis, line 2124:
```suggestion
scope = "source.msbuild"
injection-regex = "msbuild"
```

languages.toml
```diff
@@ -2117,3 +2117,23 @@ roots = []
...
+file-types = [".proj", ".vbproj", ".csproj", ".fsproj", ".targets", ".props"]
```
the-mikedavis, line 2125:
```suggestion
file-types = ["proj", "vbproj", "csproj", "fsproj", "targets", "props"]
```

These are parsed as file extensions so they don't need the leading `.`

languages.toml
```diff
@@ -2117,3 +2117,23 @@ roots = []
...
+'(' = ')'
+'{' = '}'
+'[' = ']'
+'"' = '"'
+"'" = "'"
+"<" = ">"
+
+[[grammar]]
+name = "MSBuild"
+source = { git = "https://github.com/RenjiSann/tree-sitter-xml", rev = "48a7c2b6fb9d515577e115e6788937e837815651" }
```
the-mikedavis, line 2139:
```suggestion
roots = []
grammar = "xml"

[language.auto-pairs]
'(' = ')'
'{' = '}'
'[' = ']'
'"' = '"'
"'" = "'"
"<" = ">"
```

This will re-use the xml tree-sitter grammar without adding another download.

the-mikedavis (changes requested):
In order to add syntax highlighting, you will need to create some files in `runtime/queries/msbuild/`: `highlights.scm`, `injections.scm` and `indents.scm`. All of these should just have the contents

```scheme
; inherits: xml
```

That will re-use the XML tree-sitter queries for syntax highlighting and indentation.

Also, you will need to run `cargo xtask docgen` and commit the changes to add msbuild support to the documentation. (This will fix the CI.)

languages.toml
brettlyons: thank you.  I have learned something.  Next commit has all the requested changes.

languages.toml
brettlyons: Awesome, fixed, thanks.

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #5796, opened 2023-02-02
tutor: Tweak Restore old ASCII art for HeliX name

yashpalgoyal1304:
This restores the old ASCII art for HeliX name (with one slight tweak) due to following reasons 

old one was:
- more aesthetically pleasing to the eye
- ... as it followed a wave like pattern:
`large small medium small large`
`capital small small small capital`
`H` `eli` `X`
- the `eli` were of comparable weights
- being small, and consistent, the `eli` looked cute
- The `H` is now widened by one character, so, now both H and X are of same width: 8, so, they too are now of comparable weights

The new ascii art for name:
- is confusing eye a lot
- though the `elix` are intended to be small, there doesn't feel any significant distinction from the capital H
- gives too much weight to the `e` 
- and e looks like having gaps in the inner curve side

the new design just feels lacking any character or charm of its own, 
and totally confuses the eye due to lack of any pattern.

I don't think that forcing the graphic proportion on it is a great idea....

But again, feel free to discard if you disagree

-----

Overlapping Comparison video:

https://user-images.githubusercontent.com/19423063/216477374-a00c5acf-4bd6-47a6-b69f-fc08e325992c.mp4


yashpalgoyal1304:
https://user-images.githubusercontent.com/19423063/216479374-4f6ad31e-1688-4369-bde6-2108d611dfdf.mp4

archseer:
If I wanted to keep the original text I would have indicated that in the original PR. The new design is a lot more suitable since the name is Helix and not HeliX.

yashpalgoyal1304:
> since the name is Helix and not HeliX.

is that the only concern? if yes, then i have almost fixed it, only one 'character' remains to be changed.

> The previous version was just quickly generated using an online tool.
\- @ archseer at https://github.com/helix-editor/helix/pull/5681#issuecomment-1414553261

link?

yashpalgoyal1304:
hey @archseer 
- can u please open this pull temporarily so that 
- at least the latest final commits get fetched and show up here for the history ...
- even if you don't want to merge, at least the final result of this pull will be preserved here?
- feel free to close it after the latest commits show up here...

yashpalgoyal1304:
the new x is:
- small case, 
- kinda matches the logo's style with right pointing arrow being slightly up, and left pointing arrow being slightly down

PR #5797, opened 2023-02-03
feat: add pem language

mtoohey31:
This merge request adds support for PEM files. I'm not sure whether this sort of less commonly editied filetype is something we'd like to support out of the box, but I've been working with PEM files pretty frequently recently, so I figured I'd suggest adding support. Here's a screenshot with the nord theme:

helix-pem

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -93,6 +93,7 @@
 | openscad | ✓ |  |  | `openscad-lsp` |
 | org | ✓ |  |  |  |
 | pascal | ✓ | ✓ |  | `pasls` |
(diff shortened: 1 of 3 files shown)


mtoohey31:
Just realized the grammar only supports one block per file. I'll update this and mark it as ready once I've fixed that.

mtoohey31:
Alright, that's been fixed, and I've rebased things and resolved the merge conflicts. Here's a screenshot with multiple blocks:

helix-pem-blocks

the-mikedavis (approved):
It's a pretty small parser (I see tens of KB on Linux) so I think it's ok to add 👍

Issue #5714, opened 2023-01-28
Workspace symbol search does not support special characters


frondeus:
### Summary

Rust analyzer has a feature where if you type `*` at the end of the workspace symbol search, it does include STDLIB.
However, the helix editor cannot display it correctly.

You can see, actually, that something is sent since the result count changes from 0/1 to 0/6, AND when you press backspace (to remove `*`), it shows the results for a fraction of a second.

### Reproduction Steps

https://asciinema.org/a/oCCBJaSXXysbECQc3nw2FlVpR

1. Open any Rust workspace;
2. Wait till the indexing finishes;
3. Press `CTRL+SHIFT+S`
4. type `Mutex` - there should be no results from stdlib.
5. add `*` at the end

Expected behavior:
Helix should display the response from the server with Mutex entry from stdlib, as well as from dependencies, etc.

Actual behavior:
An empty list with the correct result count.

Additionally:
1. Remove `*` from the input

Suddenly for a couple of milliseconds, Helix shows the results of `Mutex*` query.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-01-28T18:27:59.233 helix_loader [DEBUG] Located configuration folders: ["/tmp/helix-demo/.helix"]
2023-01-28T18:27:59.242 helix_view::clipboard [INFO] Using xclip to interact with the system and selection (primary) clipboard
[...]

PR #5801, opened 2023-02-03
refactor: Merge Picker and FilePicker

sudormrfbin:
Merges the code for the Picker and FilePicker into a single Picker that can show a file preview if a preview callback is provided. This change was mainly made to facilitate refactoring out a simple skeleton of a picker that does not do any filtering to be reused in a normal Picker and a DynamicPicker (see #5714; in particular [@the-mikedavis's comment][mikes-comment] and [@sudormrfbin's comment][gokuls-comment]).

The crux of the issue is that a picker maintains a list of predefined options (eg. list of files in the directory) and (re-)filters them every time the picker prompt changes, while a dynamic picker (eg. interactive global search, #4687) recalculates the full list of options on every prompt change. Using a filtering picker to drive a dynamic picker hence does duplicate work of filtering thousands of matches for no reason. It could also cause problems like interfering with the regex pattern in the global search.

I tried to directly extract a PickerBase to be reused in Picker, FilePicker and DynamicPicker, but the problem is that DynamicPicker is actually a DynamicFilePicker (i.e. it can preview file contents) which means we would need PickerBase, Picker, FilePicker, DynamicPicker and DynamicFilePicker and then another way of sharing the previewing code between a FilePicker and a DynamicFilePicker. By merging Picker and FilePicker into Picker, we only need PickerBase, Picker and DynamicPicker.

---

The PR has been organized into commits that can be easily reviewed. Each commit moves a set of related functionality from Picker into FilePicker and keeps a placeholder `unimplemented!()` call in the original region. When all methods have been moved into FilePicker, the original Picker struct and impl are deleted and FilePicker is renamed to Picker. There's close to no new code, save for some new methods like `Picker::with_preview` and `Picker::without_preview`.

[gokuls-comment]: https://github.com/helix-editor/helix/issues/5714#issuecomment-1410949578
[mikes-comment]: https://github.com/helix-editor/helix/issues/5714#issuecomment-1407451963

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2090,7 +2090,7 @@ fn global_search(cx: &mut Context) {
                     return;
                 }
 
-                let picker = FilePicker::new(
+                let picker = FilePicker::with_preview(
                     all_matches,
                     current_path,
                     move |cx, FileResult { path, line_num }, action| {
@@ -2466,7 +2466,7 @@ fn buffer_picker(cx: &mut Context) {
         is_current: doc.id() == current,
     };
 
-    let picker = FilePicker::new(
+    let picker = FilePicker::with_preview(
         cx.editor
             .documents
             .values()
@@ -2544,7 +2544,7 @@ fn jumplist_picker(cx: &mut Context) {
         }
     };
 
-    let picker = FilePicker::new(
+    let picker = FilePicker::with_preview(
         cx.editor
             .tree
             .views()
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -52,7 +52,7 @@ use crate::{
     filter_picker_entry,
     job::Callback,
     keymap::ReverseKeymap,
-    ui::{self, overlay::overlayed, FilePicker, Picker, Popup, Prompt, PromptEvent},
+    ui::{self, overlay::overlayed, Picker, Popup, Prompt, PromptEvent},
 };
 
 use crate::job::{self, Jobs};
@@ -2090,7 +2090,7 @@ fn global_search(cx: &mut Context) {
                     return;
                 }
 
-                let picker = FilePicker::with_preview(
+                let picker = Picker::with_preview(
                     all_matches,
                     current_path,
                     move |cx, FileResult { path, line_num }, action| {
@@ -2466,7 +2466,7 @@ fn buffer_picker(cx: &mut Context) {
         is_current: doc.id() == current,
     };
 
-    let picker = FilePicker::with_preview(
+    let picker = Picker::with_preview(
         cx.editor
             .documents
             .values()
@@ -2544,7 +2544,7 @@ fn jumplist_picker(cx: &mut Context) {
         }
     };
 
-    let picker = FilePicker::with_preview(
+    let picker = Picker::with_preview(
         cx.editor
             .tree
             .views()
@@ -2616,7 +2616,7 @@ pub fn command_palette(cx: &mut Context) {
                 }
             }));
 
-            let picker = Picker::new(commands, keymap, move |cx, command, _action| {
+            let picker = Picker::without_preview(commands, keymap, move |cx, command, _action| {
                 let mut ctx = Context {
                     register: None,
                     count: std::num::NonZeroUsize::new(1),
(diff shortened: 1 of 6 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -883,15 +863,15 @@ pub type DynQueryCallback<T> =
 /// A picker that updates its contents via a callback whenever the
 /// query string changes. Useful for live grep, workspace symbols, etc.
 pub struct DynamicPicker<T: ui::menu::Item + Send> {
-    file_picker: FilePicker<T>,
+    file_picker: Picker<T>,
```
the-mikedavis, line 866: Should this be renamed to just "picker" now?

the-mikedavis (approved):
The commit layout is awesome, it feels like I just watched you make this change live 😀

This refactor makes sense to me - I think I see where you're going for making filtering/sorting optional 👍

Proful:
> The commit layout is awesome, it feels like I just watched you make this change live 😀
any video link?

helix-term/src/ui/picker.rs
```diff
@@ -408,21 +126,27 @@ pub struct Picker<T: Item> {
...
 }
 
-impl<T: Item> Picker<T> {
-    pub fn new(
+impl<T: Item + 'static> Picker<T> {
+    fn new(
         options: Vec<T>,
         editor_data: T::Data,
-        callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
+        callback_fn: Box<dyn Fn(&mut Context, &T, Action) + 'static>,
```
Proful, line 148: Reason behind changing from static dispatch to dynamic dispatch?

`impl Fn` -> `Box<dyn Fn`

helix-term/src/ui/picker.rs
```diff
@@ -408,21 +126,27 @@ pub struct Picker<T: Item> {
...
+        preview_fn: Option<FilePreviewFn<T>>,
```
Proful, line 149: As `Picker` is generic now File prefix can be renamed from `FilePreviewFn`?

helix-term/src/ui/picker.rs
```diff
@@ -482,6 +210,28 @@ impl<T: Item> Picker<T> {
         picker
     }
 
+    pub fn with_preview(
+        options: Vec<T>,
+        editor_data: T::Data,
+        callback_fn: impl Fn(&mut Context, &T, Action) + 'static,
+        preview_fn: impl Fn(&Editor, &T) -> Option<FileLocation> + 'static,
```
Proful, line 217: It would be better Preview can be any content not only File content.
FileLocation?

helix-term/src/ui/picker.rs
the-mikedavis: Currently the preview only works for files so I think this name fits. This type also didn't exist before this change

helix-term/src/ui/picker.rs
the-mikedavis: Changing what the preview can show would be out of scope for this refactor PR I think - https://github.com/helix-editor/helix/pull/5492/files#r1070336613 is looking to change that anyways

helix-term/src/ui/picker.rs
pascalkuthe: #5492 changes that so that a preview of an arbitrary `String`/`Rope` is possible too, maybe just calling it `PreviewFn` now would be easy enough so we don't have to rename in the future?

helix-term/src/ui/picker.rs
pascalkuthe: Maybe just calling it `PreviewFn` now would be easy enough so we don't have to rename in the future?

I don't think the capability itself should be added in this PR but it would be a small change to future proof here to avoid renames (and larger diffs/merge conflicts) in the future

helix-term/src/commands.rs
```diff
@@ -2616,7 +2616,7 @@ pub fn command_palette(cx: &mut Context) {
                 }
             }));
 
-            let picker = Picker::new(commands, keymap, move |cx, command, _action| {
+            let picker = Picker::without_preview(commands, keymap, move |cx, command, _action| {
```
archseer, line 2619: I'd still name this new() we could also make the with_preview() a builder function that takes `self‘ so you'd chain these: new(..).with_preview(..)

pickfire: Yes, I think beside this, everything else looks okay.

Issue #2581, opened 2022-05-27
Make tab and shift-tab the only way to navigate autocompletion drop downs or make arrow keys exit insert mode to avoid code corruption issues


aral:
If you move between lines using the arrow keys in insert mode, you can end up with code corruption (#2580) due to LSP autocompletion kicking in (#2027)

The easiest way to avoid this problem would be to limit the keybindings for navigating the autocompletion drop-down to <kbd>tab</kbd> and <kbd>shift</kbd><kbd>tab</kbd>. In other words, to disallow the use of the up and down arrows on the autocompletion drop-downs.

## The other solution

The only other way I can see of addressing this problem is to make it so that Helix automatically enters normal mode when the person uses the up or down arrow keys in insert mode.

Given save points are only registered at the end of an insert session, this might actually be a safer mode of operation.

the-mikedavis:
You can disable the arrow keys in insert mode so they do nothing:

```toml
[keys.insert]
right = "no_op"
left = "no_op"
up = "no_op"
down = "no_op"
```

Or to switch into normal mode:

```toml
[keys.insert]
right = "normal_mode"
left = "normal_mode"
up = "normal_mode"
down = "normal_mode"
```

Though rebinding these doesn't affect their behavior in the completion menu.

I would be in favor of removing up/down keys from the completion menu's keymap. I think the keymap could do with fewer arrow-key bindings in general. `C-n` and `C-p` are also available so it wouldn't be only tab and shift+tab.

Having the default keymap automatically enter normal mode on arrow key presses could be a good way to train people to not use arrow keys in insert mode but it would be unexpected behavior I think. IIRC even vim leaves those keybindings for motion.

PR #5803, opened 2023-02-03
Prevent completion menu after navigation in insert mode

p1gp1g:
This is a replacement for #4144 , after the comment from @the-mikedavis :

> It would be better to be able [...] to prevent the completion menu from opening at all after any navigation in insert mode

Related issues: #2027 #2581.

The auto-completion menu now pops if you are in insert mode AND you have inserted something, as soon as you move again or do a ^C, you need to insert something again. This behavior is similar to what other IDE do (like Android Studio)

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -881,6 +881,8 @@ pub struct Editor {
     pub next_document_id: DocumentId,
     pub documents: BTreeMap<DocumentId, Document>,
 
+    pub last_event_is_char: bool,
+
     // We Flatten<> to resolve the inner DocumentSavedEventFuture. For that we need a stream of streams, hence the Once<>.
     // https://stackoverflow.com/a/66875668
     pub saves: HashMap<DocumentId, UnboundedSender<Once<DocumentSavedEventFuture>>>,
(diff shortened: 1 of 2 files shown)


helix-view/src/editor.rs
```diff
@@ -811,6 +811,8 @@ pub struct Editor {
     pub next_document_id: DocumentId,
     pub documents: BTreeMap<DocumentId, Document>,
 
+    pub last_event_is_char: bool,
```
poliorcetics, line 814: Some docs about why we need this (a link to this PR for example) would be nice

p1gp1g: Thanks for the comment, done

p1gp1g:
I just wanted to kindly follow up on my pull request and mention that I've been using it for the past 2 months and the modification has made the completion natural and instinctive. Is there anything else I can do to help move the process along?

7flash:
Yes please merge, it solves very annoying issue corrupting my code all the time

p1gp1g:
I have rebased the branch, for users using that patch.

I don't know if it could be merged until the revamping is done (then reverted) ?

pascalkuthe:
This is fixed by #8021 which is a pretty high priority now. I don't see us merging another solution before that PR

p1gp1g:
I see, I should test that PR then

PR #5808, opened 2023-02-03
Recognize .C and .H file types as cpp

budde25:
.C, and .H and two more valid cpp source and header types. Note the capitalization, as opposed two .c, and .h for c files.
I also found a source that mentions the extensions multiple different compilers support https://stackoverflow.com/a/3223792

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -190,7 +190,7 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-c", rev = "7175a6dd
 name = "cpp"
 scope = "source.cpp"
 injection-regex = "cpp"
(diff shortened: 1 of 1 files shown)


sudormrfbin (approved):
LGTM, thank you!

archseer:
Hmm isn't file type detection case insensitive? I think it used to be

pascalkuthe:
It doesn't appear to be. It's just a hashmap lookup (`language_config_ids_by_extension), ...) and there is no case conversion before hashmap lookup or insert. That might actually cause problems on case insensitive FS (windows)

PR #5816, opened 2023-02-03
allow automatic toc

tshepang:
toc

will add this to CI if so desired, so that those not running mdbook locally see all the pretty


archseer:
That won't work out of the box, we don't have the `mdbook-toc` binary.

tshepang:
It does not fail to serve docs if `mdbook-doc` is not found.

the-mikedavis (commented):
#5534 also adds a toc for the install page and some others. I think it's fine to do this manually since the sections don't change very often

archseer:
Replaced by https://github.com/helix-editor/helix/pull/5534

PR #5819, opened 2023-02-04
enable rendering in integration tests

dead10ck:
This will allow testing more of the code base, as well as enable UI- specific testing.

Debug mode builds are prohibitively slow for the tests, mostly because of the concurrency write tests. So there is now a profile for integration tests that sets the optimization level to 2 for a few helix crates, and lowers the number of rounds of concurrent writes to 1000.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,3 @@
 [alias]
 xtask = "run --package xtask --"
-integration-test = "test --features integration --workspace --test integration"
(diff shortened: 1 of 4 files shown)


archseer:
Test runtime seems reasonable (about the same ballpark range as before the PR) but the build time was significantly longer. Though I think that's because all the crates were rebuilt rather than just the helix- ones. Should be better on a cached run

PR #4398, opened 2022-10-21
Add goto_url (gu) command

izoslav:
This PR adds `gu` command which opens URL (either selected or hovered).

Closes #1472

PR #5820, opened 2023-02-04
feat(commands): open urls with goto_file command

matoous:
Add capability for `goto_file` command to open an URL under cursor.

This pulls in additional dependency - `open` by @Byron which seems to be very minimal and as such doesn't impact the compile time and/or the final size of the binary.

The updated implementation of `goto_file` first attempts to parse the WORD under the cursor as an URL. If it suceeds to do so it tries to open it in a browser (platform specific implementation from the `open` crate). If we fail to parse the WORD as URL the implementation tries to open it as a file in the editor.

Fixes: https://github.com/helix-editor/helix/issues/1472
Superseds: https://github.com/helix-editor/helix/pull/4398

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1191,27 +1191,39 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
                 .to_string(),
         );
     }
+
     for sel in paths {
         let p = sel.trim();
-        if !p.is_empty() {
-            if let Ok(url) = Url::parse(p) {
-                if let Err(e) = open::that(url.as_str()) {
-                    cx.editor.set_error(format!(
-                        "Open file failed for url '{}': {:?}",
-                        url.as_str(),
-                        e
-                    ));
-                }
+        if p.is_empty() {
+            continue;
+        }
+
+        if let Ok(url) = Url::parse(p) {
+            // file:// urls are opened inside helix
+            if url.scheme() != "file" {
+                let commands = open::commands(url.as_str());
+                cx.jobs.callback(async {
+                    for cmd in commands {
+                        let mut command = tokio::process::Command::new(cmd.get_program());
+                        command.args(cmd.get_args());
+                        if command.output().await.is_ok() {
+                            return Ok(job::Callback::Editor(Box::new(|_| {})));
+                        }
+                    }
+                    Ok(job::Callback::Editor(Box::new(move |editor| {
+                        editor.set_error("Open file failed: no command found")
+                    })))
+                });
                 return;
             }
+        }
 
-            let path = &rel_path.join(p);
-            if path.is_dir() {
-                let picker = ui::file_picker(path.into(), &cx.editor.config());
-                cx.push_layer(Box::new(overlaid(picker)));
-            } else if let Err(e) = cx.editor.open(path, action) {
-                cx.editor.set_error(format!("Open file failed: {:?}", e));
-            }
+        let path = &rel_path.join(p);
+        if path.is_dir() {
+            let picker = ui::file_picker(path.into(), &cx.editor.config());
+            cx.push_layer(Box::new(overlaid(picker)));
+        } else if let Err(e) = cx.editor.open(path, action) {
+            cx.editor.set_error(format!("Open file failed: {:?}", e));
         }
     }
 }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -28,6 +28,7 @@ use helix_core::{
     textobject,
     tree_sitter::Node,
     unicode::width::UnicodeWidthChar,
+    url::Url,
     visual_offset_from_block, Deletion, LineEnding, Position, Range, Rope, RopeGraphemes,
     RopeReader, RopeSlice, Selection, SmallVec, Tendril, Transaction,
 };
@@ -331,7 +332,7 @@ impl MappableCommand {
         goto_implementation, "Goto implementation",
         goto_file_start, "Goto line number <n> else file start",
         goto_file_end, "Goto file end",
-        goto_file, "Goto files in selection",
+        goto_file, "Goto files/URLs in selection",
         goto_file_hsplit, "Goto files in selection (hsplit)",
         goto_file_vsplit, "Goto files in selection (vsplit)",
         goto_reference, "Goto references",
@@ -1193,6 +1194,17 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     for sel in paths {
         let p = sel.trim();
         if !p.is_empty() {
+            if let Ok(url) = Url::parse(p) {
+                if let Err(e) = open::that(url.as_str()) {
+                    cx.editor.set_error(format!(
+                        "Open file failed for url '{}': {:?}",
+                        url.as_str(),
+                        e
+                    ));
+                }
+                return;
+            }
+
             let path = &rel_path.join(p);
             if path.is_dir() {
                 let picker = ui::file_picker(path.into(), &cx.editor.config());
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -28,7 +28,6 @@ use helix_core::{
     textobject,
     tree_sitter::Node,
     unicode::width::UnicodeWidthChar,
-    url::Url,
     visual_offset_from_block, Deletion, LineEnding, Position, Range, Rope, RopeGraphemes,
     RopeReader, RopeSlice, Selection, SmallVec, Tendril, Transaction,
 };
@@ -61,8 +60,13 @@ use crate::{
 
 use crate::job::{self, Jobs};
 use futures_util::{stream::FuturesUnordered, TryStreamExt};
-use std::{collections::HashMap, fmt, future::Future};
-use std::{collections::HashSet, num::NonZeroUsize};
+use std::{
+    collections::{HashMap, HashSet},
+    fmt,
+    future::Future,
+    io::Read,
+    num::NonZeroUsize,
+};
 
 use std::{
     borrow::Cow,
@@ -71,6 +75,7 @@ use std::{
 
 use once_cell::sync::Lazy;
 use serde::de::{self, Deserialize, Deserializer};
+use url::Url;
 
 use grep_regex::RegexMatcherBuilder;
 use grep_searcher::{sinks, BinaryDetection, SearcherBuilder};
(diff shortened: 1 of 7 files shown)


helix-term/src/commands.rs
```diff
@@ -1134,6 +1135,16 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     for sel in paths {
         let p = sel.trim();
         if !p.is_empty() {
+            if let Ok(url) = Url::parse(p) {
+                if let Err(e) = open::that(url.as_str()) {
```
pascalkuthe, line 1139: We should check for `file://` URLs and open them inside helix

matoous: I wasn't sure about this part, for example if I have a link to a pdf file opening it in browser might make more sense. Anyway, changed the behaviour based on the suggestion to open `file://` urls in helix, happy for any additional suggestions.

pascalkuthe: We already have a known list of files helix cannot open that we ignore in the filepicker. We could use that list to auto detect the most common files to open externally

Byron:
Thanks for using `open`!
I think it's worth pointing out that `open::that()` can actually block (and wait for the child process to finish) [depending on the launcher program it ends up using](https://github.com/Byron/open-rs/issues/51#issuecomment-1411006453). Thus it's platform dependent and not always guaranteed that it's non-blocking - typically on linux with `xdg-open`.

The `opener` crate seems to handle it by not checking the exit code of the launcher at all. It's unclear to me how to properly solve this except to try to detect blocking behaviour. Maybe thanks to `helix` the `open` crate can make up its mind on how to solve this and become a worthy dependency that way :). Your feedback of very welcome.

pascalkuthe:
> Thanks for using `open`! I think it's worth pointing out that `open::that()` can actually block (and wait for the child process to finish) [depending on the launcher program it ends up using](https://github.com/Byron/open-rs/issues/51#issuecomment-1411006453). Thus it's platform dependent and not always guaranteed that it's non-blocking - typically on linux with `xdg-open`.
> 
> The `opener` crate seems to handle it by not checking the exit code of the launcher at all. It's unclear to me how to properly solve this except to try to detect blocking behaviour. Maybe thanks to `helix` the `open` crate can make up its mind on how to solve this and become a worthy dependency that way :). Your feedback of very welcome.

For helix we don't actually need to wait for `open` to finish or report a status as `open` doesn't impact what we do in helix itself. We could just run open as a futures callback but the problem is that `open::that` is not async so we would create a blocking future, which is not great with tokio (and doesn't work with our jobs queue as it requires futures with a waker).

The best way to deal with blocking here for helix would be to use the [tokio `Command` replacement](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.status) which allows us to wait for the command status asynchronously.
The future could be dispatched using helixes jobs mechanism. Once the future completes we would set a message like `successfully openend ...` or `opening ... failed`.

The easiest way to implement this would be if `open` simply offered a function which only constructs the `Command` but doesn't actually execute it. (tokio commands can be constructed from std commands).

Looking at the code there are potentially multiple `Commands` being executed so this function would need to return a list.
This potentially requires different handeling on windows as you seem to be using `ShellExecuteW` instead of the std `Command` mechanisim. Why is that necessary @Byron?

Byron:
> The best way to deal with blocking here for helix would be to use the [tokio Command replacement](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.status) which allows us to wait for the command status asynchronously.
The future could be dispatched using helixes jobs mechanism. Once the future completes we would set a message like successfully openend ... or opening ... failed.

I really like that idea - that would conveniently bypass the blocking issue (in every direction). Even if a launcher blocks, that wouldn't be able to block `helix`.

Returning a list of `Command` instances should be fine as well. 

On windows, somebody tried [to remove the `ShellExecuteW` already](https://github.com/Byron/open-rs/issues/25), which itself was contributed _for some reason maybe related to terminal windows not popping up_, and had to [revert it later](https://github.com/Byron/open-rs/pull/27) unfortunately without saying what the actual problem is.

It's probably safe to assume that on windows, this new interface we are talking about can't exist, and one would have to work around it in `helix` somehow. And if that workaround exists, one might as well apply it for all blocking `open::that()` calls. I hope I am missing something.

pascalkuthe:
> > The best way to deal with blocking here for helix would be to use the [tokio Command replacement](https://docs.rs/tokio/latest/tokio/process/struct.Command.html#method.status) which allows us to wait for the command status asynchronously.
> > The future could be dispatched using helixes jobs mechanism. Once the future completes we would set a message like successfully openend ... or opening ... failed.
> 
> I really like that idea - that would conveniently bypass the blocking issue (in every direction). Even if a launcher blocks, that wouldn't be able to block `helix`.
> 
> Returning a list of `Command` instances should be fine as well.
> 
> On windows, somebody tried [to remove the `ShellExecuteW` already](https://github.com/Byron/open-rs/issues/25), which itself was contributed _for some reason maybe related to terminal windows not popping up_, and had to [revert it later](https://github.com/Byron/open-rs/pull/27) unfortunately without saying what the actual problem is.
> 
> It's probably safe to assume that on windows, this new interface we are talking about can't exist, and one would have to work around it in `helix` somehow. And if that workaround exists, one might as well apply it for all blocking `open::that()` calls. I hope I am missing something.

I looked into this quite a bit more and it seems that the user simply used the wrong program. He called `explorer` whereas the right command would be `cmd /c start`  (you can't call start directly, hence going trough `cmd`).

`start` internally calls the exact same winodws API: see https://en.wikipedia.org/wiki/Start_(command).
Other editors also seem to be using the same thing. For example a very popular filetree for nvim uses this: https://github.com/preservim/nerdtree/pull/1130

So that refactor should be possible after all.

Byron:
Thanks for the research! I think `cmd /c start` is also something that was implemented at some point, it seems oddly familiar. I think it's worth validating the 'terminal windows popup' behaviour before changing it, it should be easy enough to figure out. Right now it definitely doesn't show an extra terminal window.
It's definitely encouraging to see that `nerdtree` also uses this scheme and it seems fine.

Would @matoous have some time for the `open` crate refactoring? It's probably two pieces, first to use `Command` in windows as well, and secondly to expose `open::that_commands(…)` to provide the commands it would execute in order, instead of executing them, so they can be turned into `tokio` compatible async processes.

pascalkuthe:
> Thanks for the research! I think `cmd /c start` is also something that was implemented at some point, it seems oddly familiar. I think it's worth validating the 'terminal windows popup' behaviour before changing it, it should be easy enough to figure out. Right now it definitely doesn't show an extra terminal window. It's definitely encouraging to see that `nerdtree` also uses this scheme and it seems fine.
> 
> Would @matoous have some time for the `open` crate refactoring? It's probably two pieces, first to use `Command` in windows as well, and secondly to expose `open::that_commands(…)` to provide the commands it would execute in order, instead of executing them, so they can be turned into `tokio` compatible async processes.

Yeah it does seem that this approach was used in the past but removed: https://github.com/Byron/open-rs/pull/17

It sounds like there weren't really any issues with this tough just "performance" which I don't think matters much when opening a URL compared to being able to do it correctly asynronously. If you are really attachted to using the winapi call here we could have seperate `that_commands` variants in the varios os specific modules. For windows the `that` function would remain as is and the `that_commands` function would construct `Command`. For all other platforms the `that` function could just use the `that_commands` function internally.

Byron:
Great detective work, thank you! Who would have thought that `open` pulls in the `windows` crate on windows just to support a preview build.

It definitely sounds like it's best to drop using this function and use `Command` instead, to allow a unified integration with `tokio` and less dependencies overall.

To me the path forward seems to be clear and I am happy that thanks to `helix` the venerable `open` crate gets to be even more useful and future-proof.

Byron:
A [new release of `open`](https://github.com/Byron/open-rs/releases/tag/v3.4.0) is now available. Maybe this allows the PR to proceed.

pascalkuthe (commented):
this mostly LGTM just some minor comments, thanks for doing the work to bring this upstream!

kirawi:
Are you waiting on review or are you still waiting to address the previous review?

Agilan989:
Hi Helix-Editor/Helix

I wait for both


Thank you
Best regards
Agilanberan

matoous:
Sorry, forgot about the PR. Will take a look at what remains.

matoous:
Rebased, updated, tested. Feel free to give it a shot. @pascalkuthe would you please give this another look when you find some time?

helix-term/src/commands.rs
```diff
@@ -1150,13 +1151,36 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                        let mut command = tokio::process::Command::new(cmd.get_program());
+                        command.args(cmd.get_args());
+                        if command.output().await.is_ok() {
+                            return Ok(job::Callback::Editor(Box::new(|_| {})));
+                        }
+                    }
+                    Ok(job::Callback::Editor(Box::new(move |editor| {
+                        editor.set_error("Open file failed: no command found")
+                    })))
+                });
```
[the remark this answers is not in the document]
matoous: Extracted into separate function but wasn't sure what the best place for it would be (until we find a use-case for re-using it). So for now is still lives in `helix-term/src/commands.rs`.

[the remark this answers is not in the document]
pascalkuthe: its ok to keep there for now. I would eventually like to move this to helix-core if we endup needing that elsewhere but that is currently not possible since you can't create callbacks there but I think its fine to do that once we actually need it outside of helix-term

pascalkuthe (dismissed):
good to see this finised

helix-term/src/commands.rs
```diff
@@ -1190,10 +1196,53 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            let picker = ui::file_picker(path.into(), &cx.editor.config());
+            cx.push_layer(Box::new(overlaid(picker)));
+        } else if let Err(e) = cx.editor.open(path, action) {
+            cx.editor.set_error(format!("Open file failed: {:?}", e));
+        }
+    }
+}
+
+/// Opens url. If the URL is a valid textual file it is open in helix, other
+/// the file is open using external program.
```
the-mikedavis, line 1221:
```suggestion
/// Opens the given url. If the URL is a valid textual file it is open in helix, otherwise
/// the file is open using external program.
```

the-mikedavis (commented):
just some minor comments, otherwise I think this looks good

matoous:
@the-mikedavis thank you, addressed.

David-Else:
Works great using Firefox on Linux, thanks! For Markdown:

`[Tokio](https://tokio.rs/)`

Selecting `mim (inside the brackets) gf` works a treat, this is a massive feature, can't wait to see it merged :)

I noticed these warnings when compiling:

```
warning: package `cc v1.0.84` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
warning: package `rustix v0.38.22` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
```

I

pascalkuthe (approved):

Issue #814, opened 2021-10-05
LSP syntax highlighting


kirawi:
This will likely require us to revise how we handle syntax highlighting as a whole to acclimate to methods other than tree-sitter.

archseer:
I'm not a big fan of the LSP highlighting spec. It requires a lot of back and forth traffic with the LSP to do highlighting in real time. VSCode should adapt tree-sitter already..

kirawi:
I plan to make it optional in `config.toml`, but there are some cases in which tree-sitter can't be used easily.

JakeHL:
> I'm not a big fan of the LSP highlighting spec. It requires a lot of back and forth traffic with the LSP to do highlighting in real time. VSCode should adapt tree-sitter already..

Does this mean this won't ever be a first class feature of helix? I'm not really fussed either way. I only ask because I'm working on a tool to convert themes between several text editors formats and this would factor into the implementation longer term.

LouisGariepy:
While LSP is heavy, tree-sitter does not provide comparable highlighting. One solution could be to have LSP as an opt-in, with tree-sitter as the default.

As petty as it sounds, the lack of syntax highlighting is one of the very few things that hold me from daily driving Helix. Being able to jump in and see the same theming quality as I'm used to is a must for me, and I imagine other prospective users as well.

pickfire:
@LouisGariepy IIRC in the release notes I remember there is `space-h` select the same symbols in the function, that way it can be similar like part of the syntax highlight feature.

LouisGariepy:
Could a maintainer point out the conditions for an acceptable solution then? If you simply don't want this feature in Helix (which is totally fair!), maybe you could close or mark as "wontfix" to avoid confusion.

JakeHL:
@LouisGariepy 
> Being able to jump in and see the same theming quality as I'm used to is a must for me, and I imagine other prospective users as well.

I'd be interested to see a side by side comparison with the theme/languages you use if you don't mind? In my experience, it's the opposite. I quite often find tree sitter highlighting being better than the alternatives.

univerz:
> I'd be interested to see a side by side comparison with the theme/languages you use if you don't mind?

i miss mutable variables (in rust) with `bold` modifier.

archseer:
That's doable with a custom `locals.scm` query.

poliorcetics:
> That's doable with a custom `locals.scm` query.

[...]

PR #5826, opened 2023-02-04
Add vhdl language support

teburd:
The vhdl tree sitter grammar is great but the queries primarily focus on error highlights. This adds some highlights for keywords and such.

It also adds support for vhdl_ls as a language server.


teburd:
I'm far from a tree-sitter query expert but I've done my best here, most keywords/operators/etc are highlighted for me, with vhdl_ls I get nice linter checks showing up.

It's a reasonable experience, not quite on par with emacs or even vscode quite yet (vscode uses a textmate syntax highlighter which just seems to work better than my hacky treesitter queries)

I'm very open to suggestions on improving the highlights querying. The original author added a large number of error queries to annotate erroneous code which probably is best left for the language server to be frank. So it might be worth ripping out that stuff, and having a helix specific highlights.scm? Thoughts?

the-mikedavis:
Yeah I think we should rip out the `error.*` patterns, those seem out-of-place to me for syntax highlighting. We try to stick to the scopes documented here: https://docs.helix-editor.com/master/themes.html#syntax-highlighting and I don't think it makes sense to add a bunch under the `error` tree.

teburd:
So I'm finding that a lot of things just aren't in the grammar and thus don't get highlighted, things I'd really like to highlight.

std_logic/std_logic_vector are defined as part of the std_logic_1164 library that comes with vhdl, these aren't in the grammar.

They are effectively builtin types though for how widely they are used.

Can tree-sitter somehow highlight types/functions if a library is imported somehow? the textmate vhdl highlighter seemed to do something like that.

kraigher:
Regarding highlighting it is part of the LSP spec and probably something vhdl_ls could do without much effort.

teburd:
> Regarding highlighting it is part of the LSP spec and probably something vhdl_ls could do without much effort.

This would make more sense for vhdl probably than most, is that something helix would be open to enabling? @the-mikedavis ?

pascalkuthe:
Tree-sitter will always remain the lrumary mechanisim for syntax highlkghting as it is used for a lot of other functionality in helix like motions.

LSP syntax highlights also produce a lot of overhead and there is hesistancy about supporting it https://github.com/helix-editor/helix/discussions/5589#discussioncomment-4723965. The feature would also be challenging to add into the architecture of helix and none is working on it so I don't expect this yo ne finished anytime soon.

If it's supported it would only augment tree sitter not replace it:
https://github.com/helix-editor/helix/issues/814#issuecomment-1370078855

teburd:
@the-mikedavis @archseer been awhile since this was posted. I understand the project has a lot of churn but it would be nice to at least have the startings of vhdl support

the-mikedavis (approved):

the-mikedavis:
Could you add some info on vhdl_ls to the language server wiki? https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

Looks like they suggest installing from the GitHub releases artifacts https://github.com/VHDL-LS/rust_hdl

Issue #5505, opened 2023-01-11
Key remapping for compositor components


the-mikedavis:
There are some prior issues on this for [pickers](https://github.com/helix-editor/helix/issues/2757), [prompts](https://github.com/helix-editor/helix/issues/3205), and [popups](https://github.com/helix-editor/helix/issues/5498). This issue is meant to be a tracking issue for key remapping for any `Component`. Ideally you should be able to remap any actions for any component (pickers, prompts, menus, etc.). I think that the solution to remapping for one component should be the same or very similar to the solution for remapping any component so this issue covers all components.

This is a little tricky implementation-wise: regular commands share the same function signature that takes a `&mut Context`. Component commands are currently built straight into each component. For example see scrolling a popup: https://github.com/helix-editor/helix/blob/2412f7c8374d2f6a8bfd66ecf424153a37c814bf/helix-term/src/ui/popup.rs#L186-L193

This will also take some design work to decide how the config should look. I am imagining something like:

```toml
# ~/.config/helix/config.toml
[keys.normal]
# this section already exists for regular bindings, along with keys.insert and keys.select
[keys.popup]
C-j = "scroll_popup_down"
C-k = "scroll_popup_up"
# maybe these can be namespaced, like `C-j = "popup::scroll_down"`?
[keys.picker]
C-space = "validate_picker_selection"
```

yashpalgoyal1304:
hi! would this be applicable to command output popups https://github.com/helix-editor/helix/discussions/5923#discussioncomment-4946453 too?

the-mikedavis:
Yep, the `:run-shell-command`/`:sh` command creates a Popup element and Popups are Components

yashpalgoyal1304:
hey! 
so, this concept of being able to "remap" is there in all of the above issues, 
but the following thing seems to be not discussed.

currently 
- the keys seem to be tied to the component - so, 
    + `C-d` works in [shell output],
    + `C-n`/`<tab>` works in [picker]
    + `hjkl` works in editor (despite having other components open)

but how about having
- the ability to change the current active component 
- so that the normal keys work across any component as long as it's the active one?

[shell output]: https://github.com/helix-editor/helix/discussions/5923#discussioncomment-4946453
[picker]: https://github.com/helix-editor/helix/discussions/5299#discussioncomment-4509815

tdelabro:
[...]

PR #5831, opened 2023-02-04
added ctrl+j readline-style alias for Enter key

adriand2511:
Hi, added ctrl+j to menu, picker and prompt to be more in-line with the other readline-style keybinds (ctrl+p, ctrl+n, ctrl+h etc.)


the-mikedavis:
I don't think this is necessary. We're not aiming for full compatibility with readline and Enter should be enough

adriand2511:
Is there a way to add this keybind to config.toml?

the-mikedavis:
Not currently for components like menus, prompts and pickers. See #5505

adriand2511:
Aight, I'll close it then. I'd like to tackle that key remapping issue but I'm a total Rust beginner and don't know if I could handle it.

PR #5832, opened 2023-02-05
use editor.shell for formatter

### LINKED ISSUE #5538: expand tilde to home-directory for language formatter command args
right now in my `languages.toml` I have this:

`formatter = { command = 'deno', args = ["fmt", "-", "--ext", "md", "--config", "/home/mike/.config/deno.json" ] }`

however in order to be compatible to any environment I'd like to store this in my dotfiles repo:

`formatter = { command = 'deno', args = ["fmt", "-", "--ext", "md", "--config", "~/.config/deno.json" ] }`

## reasoning:

- default formatting options for markdown of deno are not what I want
- my current hardcoded config path argument for deno works fine on my home-pc but for work my username is not mike and I am on macOS where my home is `/Users/kmike`, I'd like to share my configs hence tilde-expansion to homedir is needed

## problem:

right now using tilde breaks calling the formatter because it tries to open `$(cwd)/~/.config/deno.json` which does not exist.

kirawi:
# Objective

- Resolves #5538

## Solution

- Copied the way that the shell is used in `helix_term::commands::shell_impl_async`: https://github.com/helix-editor/helix/blob/b2e83f81e10089a0e81ce33c4beb51aefc29a62e/helix-term/src/commands.rs#L4996-L5001

---

## Changelog

now passes external formatters to `editor.shell`

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -447,8 +447,11 @@ impl Document {
         {
             use std::process::Stdio;
             let text = self.text().clone();
-            let mut process = tokio::process::Command::new(&formatter.command);
+            let shell = &self.config.load().shell;
+            let mut process = tokio::process::Command::new(&shell[0]);
             process
+                .args(&shell[1..])
+                .arg(&formatter.command)
                 .args(&formatter.args)
                 .stdin(Stdio::piped())
                 .stdout(Stdio::piped())


archseer:
We decided against this here https://github.com/helix-editor/helix/pull/8064#issuecomment-1696823637

PR #5835, opened 2023-02-05
Added show-clipboard command

Proful:
New to Rust & Helix. Open for any suggestion.

There is no issue if this PR is not approved & merged.

`:show-clipboard`
- S system clipboard content
- other register content

image

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -970,6 +970,69 @@ fn show_clipboard_provider(
         .set_status(cx.editor.clipboard_provider.name().to_string());
     Ok(())
 }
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I don't think a new picker for this is necessary: you can view the contents of a register or the system clipboard by creating a new scratch buffer and pasting out the register or clipboard. You can also preview register contents with `select_register` (`"`). Viewing the system clipboard seems like something that should be solved outside of Helix IMO, probably with some sort of clipboard manager tool (for example [clipman](https://github.com/yory8/clipman))

Proful:
Nevermind, thanks for the feedback. I will close the PR.

PR #5837, opened 2023-02-05
Auto indent on `insert_at_line_start`

### LINKED ISSUE #4859: `insert_at_line_start` with autoindent
Looking at #4854, automatically indenting when using <kbd>I</kbd> on empty lines would be a nice addition in my opinion.

I'm not sure if it makes sense to be the default behavior, since it could be annoying when the auto indents for a language are off. Maybe a separate command is a better idea?

CptPotato:
First attempt at #4859. This makes `I` and `A` automatically indent, but **only** if the line is empty. Otherwise it behaves exactly as before. I thought about also handling indentation when the line only contains white space, but I left this out for now. Either way, this is a breaking change.

I roughly followed the implementation of `open_above`/`open_below`. Though, I'm not sure if the implementation is very good, all the text editing functions in the code base are new to me.

Here's the result of `insert_at_line_start` / `I` with this change (the only difference is on the empty line):

| before | master | this PR |
|---|---|---|
| image | image | image |

I tried to handle all edge cases (i.e. cursor is at document start/end, multiple selections) and so far it seems to work. ~~I also plan to make `insert_at_line_end` behave similarly~~.

Edit: `insert_at_line_end` now works the same.

closes #4859

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2761,6 +2761,7 @@ fn insert_at_line_start(cx: &mut Context) {
     insert_with_indent(cx, |cursor_line, cursor_line_start, text| {
         find_first_non_whitespace_char(text.line(cursor_line))
             .map(|ws_offset| ws_offset + cursor_line_start)
+            .unwrap_or(cursor_line_start)
     });
 }
 
@@ -2768,7 +2769,7 @@ fn insert_at_line_start(cx: &mut Context) {
 // If the line is empty, automatically indent
 fn insert_at_line_end(cx: &mut Context) {
     insert_with_indent(cx, |cursor_line, _, text| {
-        Some(line_end_char_index(text, cursor_line))
+        line_end_char_index(text, cursor_line)
     });
 }
 
@@ -2776,7 +2777,7 @@ fn insert_at_line_end(cx: &mut Context) {
 // If the line is not empty, move the cursor to the specified fallback position.
 fn insert_with_indent(
     cx: &mut Context,
-    cursor_fallback: impl Fn(usize, usize, &RopeSlice) -> Option<usize>,
+    cursor_fallback: impl Fn(usize, usize, &RopeSlice) -> usize,
 ) {
     enter_insert_mode(cx);
 
@@ -2821,10 +2822,8 @@ fn insert_with_indent(
             (line_end_index, line_end_index, Some(indent.into()))
         } else {
             // move cursor to the fallback position
-            let range = cursor_fallback(cursor_line, cursor_line_start, &text)
-                .map(|pos| range.put_cursor(text, pos + offs, cx.editor.mode == Mode::Select))
-                .unwrap_or(*range);
-            ranges.push(range);
+            let pos = cursor_fallback(cursor_line, cursor_line_start, &text);
+            ranges.push(range.put_cursor(text, pos + offs, cx.editor.mode == Mode::Select));
 
             (cursor_line_start, cursor_line_start, None)
         }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -16,9 +16,7 @@ use helix_core::{
     history::UndoKind,
     increment, indent,
     indent::IndentStyle,
-    line_ending::{
-        get_line_ending_of_str, line_end_char_index, str_is_line_ending,
-    },
+    line_ending::{get_line_ending_of_str, line_end_char_index, str_is_line_ending},
     match_brackets,
     movement::{self, move_vertically_visual, Direction},
     object, pos_at_coords,
@@ -2760,6 +2758,26 @@ fn last_picker(cx: &mut Context) {
 // I inserts at the first nonwhitespace character of each line with a selection
 // If the line is empty, automatically indent
 fn insert_at_line_start(cx: &mut Context) {
+    insert_with_indent(cx, |cursor_line, cursor_line_start, text| {
+        find_first_non_whitespace_char(text.line(cursor_line))
+            .map(|ws_offset| ws_offset + cursor_line_start)
+    });
+}
+
+// A inserts at the end of each line with a selection
+// If the line is empty, automatically indent
+fn insert_at_line_end(cx: &mut Context) {
+    insert_with_indent(cx, |cursor_line, _, text| {
+        Some(line_end_char_index(text, cursor_line))
+    });
+}
+
+// Enter insert mode and auto-indent the current line if it is empty.
+// If the line is not empty, move the cursor to the specified fallback position.
+fn insert_with_indent(
+    cx: &mut Context,
+    cursor_fallback: impl Fn(usize, usize, &RopeSlice) -> Option<usize>,
+) {
     enter_insert_mode(cx);
 
     let (view, doc) = current!(cx.editor);
@@ -2802,13 +2820,9 @@ fn insert_at_line_start(cx: &mut Context) {
 
             (line_end_index, line_end_index, Some(indent.into()))
         } else {
-            // move cursor to the first non-whitespace character of the current line
-            let mut pos = cursor_line_start + offs;
-            let range = find_first_non_whitespace_char(text.line(cursor_line))
-                .map(|ws_offset| {
-                    pos += ws_offset;
-                    range.put_cursor(text, pos, cx.editor.mode == Mode::Select)
-                })
+            // move cursor to the fallback position
+            let range = cursor_fallback(cursor_line, cursor_line_start, &text)
+                .map(|pos| range.put_cursor(text, pos + offs, cx.editor.mode == Mode::Select))
                 .unwrap_or(*range);
             ranges.push(range);
 
@@ -2818,20 +2832,6 @@ fn insert_at_line_start(cx: &mut Context) {
 
     transaction = transaction.with_selection(Selection::new(ranges, selection.primary_index()));
     doc.apply(&transaction, view.id);
-}
-
-// A inserts at the end of each line with a selection
-fn insert_at_line_end(cx: &mut Context) {
-    enter_insert_mode(cx);
-    let (view, doc) = current!(cx.editor);
-
-    let selection = doc.selection(view.id).clone().transform(|range| {
-        let text = doc.text().slice(..);
-        let line = range.cursor_line(text);
-        let pos = line_end_char_index(&text, line);
-        Range::new(pos, pos)
-    });
-    doc.set_selection(view.id, selection);
 }
 
 // Creates an LspCallback that waits for formatting changes to be computed. When they're done,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -16,7 +16,9 @@ use helix_core::{
     history::UndoKind,
     increment, indent,
     indent::IndentStyle,
-    line_ending::{get_line_ending_of_str, line_end_char_index, str_is_line_ending},
+    line_ending::{
+        get_line_ending_of_str, line_end_char_index, str_is_line_ending,
+    },
     match_brackets,
     movement::{self, move_vertically_visual, Direction},
     object, pos_at_coords,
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I haven't looked too closely at the implementation yet but this does seem very useful to me and like a better default too. Thanks for working on this!

That being said we might want to keep the old behaviour as a seperate command so people that don't like this change can rebind.

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+
+    let mut transaction = Transaction::change_by_selection(contents, selection, |range| {
+        let cursor = range.from();
+        let cursor_line = text.char_to_line(cursor);
+
+        let doc_end = contents.len_chars();
+        if range.from() == doc_end {
+            ranges.push(Range::point(doc_end + offs));
+            return (doc_end, doc_end, None);
+        }
```
pascalkuthe, line 2689:
```suggestion
```

I don't think this is correct, the EOF position is not always it's own line. If the file doesn't end with a newline terminator the EOF is part of a normal line and `I` should work in that case too.
This condition is not even required because all functions in helix handle both the one past the end char index and one past the end line just fine (And I think we should also auto indent in that case because everything else would be inconsistent)

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+            // move cursor to the first non-whitespace character of the current line
+            let first_non_ws = line_content
+                .chars()
+                .enumerate()
+                .find_map(|(i, c)| if !c.is_whitespace() { Some(i) } else { None })
+                .unwrap_or_default();
+
+            let pos = offs + text.line_to_char(cursor_line) + first_non_ws;
+            ranges.push(Range::point(pos));
+            (pos, pos, None)
```
pascalkuthe, line 2734:
```suggestion
	     let mut pos = text.line_to_char(cursor_line);
             if let Some(whitespace_off) = find_first_non_whitespace_char(text.line(line)) {
             	   pos += whitespace_off
	          range.put_cursor(text, pos, cx.editor.mode == Mode::Select)
	     }
	     ranges.push(range)
            (pos, pos, None)
```

This branch does not behave exactly the same as the old code and also contains quite a bit of duplication. I ported the above from `goto_first_nonwhitespace` and made some adjustements to fit your usecase here (note that you must change `|range|` to `|mut range|` for this to work

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+            ranges.push(Range::point(doc_end + offs));
+            return (doc_end, doc_end, None);
+        }
+
+        let line_content = line_without_line_ending(&text, cursor_line);
+        if line_content.len_chars() == 0 {
+            // line is empty => auto indent
+            let (line_end_index, line_end_offset_width) = if cursor_line == 0 {
+                (0, 0)
+            } else {
```
pascalkuthe, line 2696: this condition seems unnecessay. I guess you special cased this for performance reasons? I don't think thats a concern here. Especially with my other suggested changes the branch below is free anyway.

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+                (
+                    line_end_char_index(&doc.text().slice(..), cursor_line),
```
pascalkuthe, line 2698:
```suggestion
                    cursor_line_start
```
Avoid recalculating that here the line is always empty (you checked above), so the result of `line_end_char_index` is always the same as the line start

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+                tab_width,
+                text,
+                cursor_line,
+                line_end_index,
+                cursor_line,
+            );
+
+            let indent_len = indent.len();
+            let mut text = String::with_capacity(indent_len);
+            text.push_str(&indent);
```
pascalkuthe, line 2716:
```suggestion
```

This is unnecessary `indent` is already a String, just delete this and use `indent` instead of `text below

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+    let language_config = doc.language_config();
+    let syntax = doc.syntax();
+    let tab_width = doc.tab_width();
+
+    let mut ranges = SmallVec::with_capacity(selection.len());
+    let mut offs = 0;
+
+    let mut transaction = Transaction::change_by_selection(contents, selection, |range| {
+        let cursor = range.from();
+        let cursor_line = text.char_to_line(cursor);
```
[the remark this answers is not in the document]
pascalkuthe: I would also recommend calculating `cursor_line_start` as you jump trough quite a few hoops to calculate that in a more complex way then really necessary later.

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+
+        let line_content = line_without_line_ending(&text, cursor_line);
+        if line_content.len_chars() == 0 {
+            // line is empty => auto indent
+            let (line_end_index, line_end_offset_width) = if cursor_line == 0 {
+                (0, 0)
+            } else {
+                (
+                    line_end_char_index(&doc.text().slice(..), cursor_line),
+                    doc.line_ending.len_chars(),
```
pascalkuthe, line 2699: `doc.line_ending` is not always equivalent to the line_ending on a line. A document can have mixed line endings. You don't need this anyway

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+
+            let indent_len = indent.len();
+            let mut text = String::with_capacity(indent_len);
+            text.push_str(&indent);
+
+            // calculate new selection ranges
+            let pos = offs + line_end_index + line_end_offset_width;
+            ranges.push(Range::point((pos + indent_len).saturating_sub(1)));
+
+            offs += text.chars().count();
```
pascalkuthe, line 2722:
```suggestion
            let pos = offs + cursor_line_start ;
            let indent_width = indent.chars().count();
            ranges.push(pos + indent_width);
            offs += indent_width;
```

you are using byte indicies here (`indent_len`). You should be using char incidences as you do below. Also you are not considering select mode here.

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,85 @@ fn last_picker(cx: &mut Context) {
...
+            ranges.push(Range::point(pos));
+            (pos, pos, None)
+        }
+    });
+
+    transaction = transaction.with_selection(Selection::new(ranges, selection.primary_index()));
+    doc.apply(&transaction, view.id);
+
+    // goto_first_nonwhitespace(cx);
+    // enter_insert_mode(cx);
```
pascalkuthe, line 2742: these comments can be removed.

pascalkuthe (changes requested):
The idea is good I think, the implementation can be simplified (and is not quite correct, see review comments)

CptPotato:
Thanks for the detailed review and pointers @pascalkuthe. I expected that I was doing a lot of these things sub-optimally. I'll try to address these issues soon.

helix-term/src/commands.rs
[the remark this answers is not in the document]
CptPotato: Thanks for the hint, I'm using this version now.

helix-term/src/commands.rs
CptPotato: Yup, this is not required anymore now.

helix-term/src/commands.rs
CptPotato: Applied this with some slight changes (using `map` instead of `if let` and taking `offs` into account). I noticed one source of confusion was that the `Change`'s from and to position refers to positions before the transactions were applied, while the `ranges` are selections using the positions of the new resulting text.

helix-term/src/commands.rs
CptPotato: This was a left-over check to prevent a panic. Though this isn't required anymore :+1:.

helix-term/src/commands.rs
CptPotato: Done.

helix-term/src/commands.rs
CptPotato: Another left-over, removed.

helix-term/src/commands.rs
CptPotato: Removed this.

helix-term/src/commands.rs
CptPotato: Done.

helix-term/src/commands.rs
CptPotato: Done.

helix-term/src/commands.rs
pascalkuthe: good point forgot about that

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,68 @@ fn last_picker(cx: &mut Context) {
...
+            // calculate new selection ranges
+            let pos = offs + cursor_line_start;
+            let indent_width = indent.chars().count();
+            ranges.push(Range::point(pos + indent_width));
+            offs += indent_width;
+
+            (line_end_index, line_end_index, Some(indent.into()))
+        } else {
+            // move cursor to the first non-whitespace character of the current line
+            let pos_before = text.line_to_char(cursor_line);
```
pascalkuthe, line 2710: this is the same as `cursor_line_start` so no need to recompute

helix-term/src/commands.rs
```diff
@@ -2659,9 +2661,68 @@ fn last_picker(cx: &mut Context) {
...
+
+    let mut ranges = SmallVec::with_capacity(selection.len());
+    let mut offs = 0;
+
+    let mut transaction = Transaction::change_by_selection(contents, selection, |range| {
+        let cursor_line = range.cursor_line(text);
+        let cursor_line_start = text.line_to_char(cursor_line);
+
+        let line_content = line_without_line_ending(&text, cursor_line);
+        if line_content.len_chars() == 0 {
```
pascalkuthe, line 2686: You are not using `line_content` anywhere except to check if its width it 0. This can be done slightly more efficiently by doing:
```suggestion
        if line_end_char_index(text, cursor_line) == cursor_line_start {
```

pascalkuthe (changes requested):
Two minor nits otherwise the implementation now LGTM.

CptPotato:
I made the last suggested changes, thanks again.

I also created a second commit that adds the same behavior to `insert_at_line_end` by moving the indent code to a separate function.

CptPotato:
Should I look into creating a test for the new behavior?

pascalkuthe:
> Should I look into creating a test for the new behavior?

I believe there is currently no tests for the existing behaviour so I wouldn't consider it a blocking issue but more tests are always good so if you want to add them go ahead

CptPotato:
I added a test and found one remaining issue (which I fixed).

The test is a little ugly but it's the best I could do :eyes:
It checks both `insert_at_line_start`/`end` on all lines with this snippet:

```rust
fn foo() {
    if let Some(_) = None {

    }
 
}

fn bar() {

}
```
<sup>(there's one line with existing whitespace)</sup>

Is this okay or should I change the test?

CptPotato:
Looks like some recent changes are causing conflicts.

Let me know when somebody wants to review this and I'll resolve them.

pascalkuthe (dismissed):
I have been testing this for a bit and did a pretty thorough review a while ago. Once the bugfix release (23.3.1) is done we can resolve the conflicts here (should be pretty easy) I think we can land this

CptPotato:
Sorry for the long delay. I rebased onto latest master, should be good to go now.

helix-term/src/commands.rs
```diff
@@ -2756,23 +2756,81 @@ fn last_picker(cx: &mut Context) {
...
+    insert_with_indent(cx, |cursor_line, _, text| {
+        line_end_char_index(text, cursor_line)
+    });
+}
+
+// Enter insert mode and auto-indent the current line if it is empty.
+// If the line is not empty, move the cursor to the specified fallback position.
+fn insert_with_indent(
+    cx: &mut Context,
+    cursor_fallback: impl Fn(usize, usize, &RopeSlice) -> usize,
```
the-mikedavis, line 2780: Rather than a function that the callers pass in, I think it would be more idiomatic for the callers to pass an enum of `Start`/`Beginning` or `End` and then to have this function do a match:

```rust
let cursor_fallback = match insert_at {
    Start => find_first_non_whitespace_char(text.line(cursor_line)).map(|ws_offset| ws_offset + cursor_line_start).unwrap_or(cursor_line_start),
    End => line_end_char_index(text, cursor_line),
};
```

It would be a little less flexible if `insert_with_indent` were to be used in more places but I don't see other use-cases for that function.

the-mikedavis (commented):
I have a small style nit but otherwise I think this looks great (especially the integration tests!)

helix-term/src/commands.rs
Tudyx: A potential use case is if we want to have `xc` insert at indent

CptPotato: Agree that this is a bit simpler, I changed it to an enum.

the-mikedavis (approved):

pascalkuthe (approved):
thanks for working on this! This is a great QOL improvement

Issue #4740, opened 2022-11-14
Trasparency/opacity should not be a part of the  theme.


igor-petruk:
It should be a separate setting that could be applied to any theme, instead of creating transparent version like `base16_transparent`. Otherwise users have to customize every single theme if they want to try it on transparent terminal.

the-mikedavis:
Since #3067, you can inherit any theme and unset the `ui.background` scope:

```toml
# ~/.config/helix/dracula_transparent.toml
inherits = "dracula"
"ui.background" = {}
```

I don't like the idea of adding an extra config option for this.

igor-petruk:
This suggestion does not work for some reason. 

First the file is only detected if it is not ` ~/.config/helix/dracula_transparent.toml`, but in `themes` subdirectory. I assume it is a typo. Also I use `onedark`, if it makes a difference.

Then the failure is "Invalid theme: 'ui.selection` required". Probably like https://github.com/helix-editor/helix/issues/2247

OS - OpenBSD 7.2, helix is installed from a repo. These are the standard OS patches from the stock source - https://github.com/openbsd/ports/tree/master/editors/helix/patches

```
hx --health
Config file: /home/user/.config/helix/config.toml
Language file: default
Log file: /home/user/.cache/helix/helix.log
Runtime directory: /usr/local/share/helix/runtime
Clipboard provider: none
....
```

the-mikedavis:
Ah yeah, that's a typo, it should be under `~/.config/helix/themes`. I'll edit the original post.

Theme inheritance is a new feature in 22.12 but you're running on 22.08.1. It looks like OpenBSD ports hasn't updated to 22.12 yet: https://repology.org/project/helix/versions

igor-petruk:
Alright, thanks. In that case I will attempt to just copy a theme and edit
it. Thanks.

aral:
Is there a way to get the transparency value from the terminal environment and apply it automatically to the background colour of whichever theme is active?

the-mikedavis:
AFAIK there isn't a cross-terminal way to figure out how the terminal opacity is configured. Even if we could find out, there's no way to tell the terminal how opaque to draw a background color: all colors are solid. The way that transparency works for themes like `base16_transparent` is that the theme unsets the `ui.background` scope so that we don't draw a background color at all.

d249u7:
[...]

PR #5839, opened 2023-02-05
The ability to define global theme overrides

caferen:
Currently, making changes to a theme is only possible by inheritance or editing the theme directly. Both of these options require a per-theme approach. But overrides such as "using the terminal's background" usually are meant for any theme. This makes it annoying to change themes after doing some alterations to the previous one.

~~This PR add the option to create a `theme_overrides.toml` file in the config directory and apply the changes to any theme currently active.~~

This PR adds the option to use a `theme` table in `config.toml` that has `theme.overrides` option. The keys specified here will override the active theme.

- The changes are backwards compatible.
- The overrides will persist while changing themes with `:theme`. 
- `theme.overrides` table is internally the same as a theme so any subkeys a theme can have, overrides can also have.

Example:

```
[theme]
name = "github_dark"

[theme.overrides]
"ui.background" = {}
"ui.cursor" = { fg = "dark", bg = "light" }
"ui.statusline" = { bg = {} }
"ui.statusline.normal" = { fg = "dark", bg = "light" }
"ui.statusline.insert" = { fg = "dark", bg = "light" }
"ui.statusline.select" = { fg = "dark", bg = "light" }
"ui.statusline.separator" = { fg = "light", bg = "dark" }

[theme.overrides.palette]
dark = "#000000"
light = "#E9E0C9"
```

[A somewhat related issue](https://github.com/helix-editor/helix/issues/4740)


pascalkuthe:
Instead of a seperate config file could this not be subkeys in `config.toml`, for example:

``` toml
[theme.overwrite]
# Theme overwrites here
```

sudormrfbin:
Exactly my thoughts as well, but `theme` would then be a toml map and I think we lose the ability to do `theme = "onedark"`.

d249u7:
> Instead of a seperate config file could this not be subkeys in `config.toml`, for example:
> 
> ```toml
> [theme.overwrite]
> # Theme overwrites here
> ```


@pascalkuthe wouldn't it add unnecessary overhead to the configuration file if a user wants to make too many overrides? That was my thinking.

pascalkuthe:
> Exactly my thoughts as well, but `theme` would then be a toml map and I think we lose the ability to do `theme = "onedark"`.

Ad damm I always forget about that. I hate that aspect of toml even if that makes it a lot easier to deal with. You could do soo many cool things like this.
Maybe we could just allow a hybrid. Either allow the current `theme = "onedark"` or make it a map:


``` toml
[theme]
name = "onedark"
[theme.overwrite]
# Theme overwrites here
```

Might be confusing tough so perhaps just adding a separate `theme_overwrite` map might be better.
``` toml
[theme_overwrite]
# Theme overwrites here
```

Not sure which of those two options are better.


> > Instead of a seperate config file could this not be subkeys in `config.toml`, for example:
> > ```toml
> > [theme.overwrite]
> > # Theme overwrites here
> > ```
> 
> @pascalkuthe wouldn't it add unnecessary overhead to the configuration file if a user wants to make too many overrides? That was my thinking.

I think a global overwrite like this is only useful for a few themekeys like the background and maybe the cursor.
Most other overwrites would be theme specifc (it would look if all themes used the same color for certain syntax highlights) so I don't think that would be a problem. 

I would be very cautious about adding new config files.

d249u7:
> I think a global overwrite like this is only useful for a few themekeys like the background and maybe the cursor. Most other overwrites would be theme specifc (it would look if all themes used the same color for certain syntax highlights) so I don't think that would be a problem.

That is correct and my use case as well.

> I would be very cautious about adding new config files.

I think including this in the config is the better choice anyway but for my reference, what is the potential issue with adding new config files?

Regarding how to define in the config, first option seems more extendable and structured. However, wouldn't it break existing configs?

pascalkuthe:
> I think including this in the config is the better choice anyway but for my reference, what is the potential issue with adding new config files?

Primarily, If we keep adding new config files, the config becomes more spread-out and harder to find what you are looking for. 
There is also an effort to add per-workspace configs (#5748) and each new file needs to be discovered with multiple fallback paths etc.

> Regarding how to define in the config, first option seems more extendable and structured. However, wouldn't it break existing configs?

As I said we would allow both options for backwards compatibility.
We check if `theme` is a string and treat it as we do now or we check if `theme` is a table and interpret as I showed above.
We have avoided breaking changes to the config multiple times with schemes like this.

It's a bit confusing that you would need to change the way the theme name is specified if you want to overwrite something that's why I also offered the second alternative but we do stuff like this already for other settings too so it's probably fine.

sudormrfbin:
> As I said we would allow both options for backwards compatibility.
> We check if `theme` is a string and treat it as we do now or we check if `theme` is a table and interpret as I showed above.
> We have avoided breaking changes to the config multiple times with schemes like this.

This is quite common, and serde even has a doc page on the subject: https://serde.rs/string-or-struct.html.

d249u7:
Gave it a shot here. This is working when the theme is specified with either one the methods but failing when it's not specified. 

Couldn't figure out why it fails. Pointers are appreciated.

d249u7:
@pascalkuthe these changes capture what I meant to do

con-f-use:
So, what's the status on this and why was it closed unmerged?

PR #5842, opened 2023-02-05
Refactor completion doc popup and fix positioning

sudormrfbin:
Refactor the logic for rendering completion documentation popup. The changes include using an early return, renaming some variables (overshadowing `area`, `height`, etc tends to be confusing) and making some area calculations more readable. Most importantly this PR also contains a fix for preventing the rendering of the completion doc popup over the completion popup itself and to not draw over the cursor.

| Before | After |
| --- | --- |
| ![before][1] | ![after][2] |

[1]: https://user-images.githubusercontent.com/23398472/216852789-05edb90f-68c9-4b9e-9da3-c8002dc5bad0.png

[2]: https://user-images.githubusercontent.com/23398472/216852784-92bdccf7-a130-40c9-99be-021955522d64.png

---

Each commit does a single refactor and the last commit contains the bug fix (The full diff across all the commits might be hard to read since the first commit introduces an early return that dedents a whole block; stepping through the commits should be easier for reviewing).

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -381,104 +381,106 @@ impl Component for Completion {
         self.popup.render(area, surface, cx);
 
         // if we have a selection, render a markdown popup on top/below with info
-        if let Some(option) = self.popup.contents().selection() {
-            // need to render:
-            // option.detail
-            // ---
-            // option.documentation
+        let option = match self.popup.contents().selection() {
+            Some(option) => option,
+            None => return,
+        };
+        // need to render:
+        // option.detail
+        // ---
+        // option.documentation
 
-            let (view, doc) = current!(cx.editor);
-            let language = doc.language_name().unwrap_or("");
-            let text = doc.text().slice(..);
-            let cursor_pos = doc.selection(view.id).primary().cursor(text);
-            let coords = view
-                .screen_coords_at_pos(doc, text, cursor_pos)
-                .expect("cursor must be in view");
-            let cursor_pos = coords.row as u16;
+        let (view, doc) = current!(cx.editor);
+        let language = doc.language_name().unwrap_or("");
+        let text = doc.text().slice(..);
+        let cursor_pos = doc.selection(view.id).primary().cursor(text);
+        let coords = view
+            .screen_coords_at_pos(doc, text, cursor_pos)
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/completion.rs
```diff
@@ -381,104 +381,106 @@ impl Component for Completion {
...
+        // ---
+        // option.documentation
+
+        let (view, doc) = current!(cx.editor);
+        let language = doc.language_name().unwrap_or("");
+        let text = doc.text().slice(..);
+        let cursor_pos = doc.selection(view.id).primary().cursor(text);
+        let coords = view
+            .screen_coords_at_pos(doc, text, cursor_pos)
+            .expect("cursor must be in view");
```
Proful, line 399: will it panic?

pascalkuthe:
Right now we render the documentation popup on the right of the completion popup when there is enough space. Maybe instead of showing the completion popup on top when there isn't enough space, we should show it in the left instead.
I think vscore does but that way too (and actually defaults to the left but that's not that importent).

helix-term/src/ui/completion.rs
sudormrfbin: The primary cursor will always be on screen, so it shouldn't panic. Also the `expect()` call was already part of the code before the change (turning on ignore whitespace changes in the diff UI will give a more accurate view of the change set).

sudormrfbin:
I actually wanted to implement it but forgot about it in the middle of fixing the bug, thanks for the reminder :) Should be a pretty small change.

helix-term/src/ui/completion.rs
pascalkuthe: it was introduced in #5420. It originally used the `visual_offset_at_pos` function which always returned a value as that function is not aware of the view position. With #5420 all those occurances had to be changed to use view position aware. I am pretty sure that `expect` is fine here because the cursor should always be in view while editing in insert mode. 

There are situations where the cursor is not inside the view (mostly bugs/oversights but it happens on purpose too) but I am pretty sure that's never the case while editing in insert mode (barring bugs). I am not even sure what else we could o here if the cursor is not in screen? Just not draw the popup?

archseer (approved):
Great work!

Issue #5732, opened 2023-01-30
Send `exit` notification to not-yet-initialized language servers instead of `shutdown`


Seelengrab:
### Summary

Per [the spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize) for the initialization of an LS:

> Notifications should be dropped, except for the exit notification. This will allow the exit of a server without an initialize request.

So during initialization, a conforming LS _should_ still respect the `exit` notification and exit (almost) immediately. Right now we send (as far as I can tell) `shutdown` here:

https://github.com/helix-editor/helix/blob/4d548a0ee39c14ba54b672f6b57072cf306c11a9/helix-view/src/editor.rs#L1456-L1466

https://github.com/helix-editor/helix/blob/4d548a0ee39c14ba54b672f6b57072cf306c11a9/helix-lsp/src/client.rs#L409-L414

which should check whether the language server is initialized and send `exit` directly instead, if it isn't. 

### Reproduction Steps

 

I tried this:

1. `hx -vvvvv some_julia_file.jl`
2. `:q` as soon as the editor is up.

It should be reproducible with any LS that takes a long time to initialize, the julia one without a sysimage is just "perfect" for this. Takes about 6 seconds to initialize.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-01-30T12:32:36.584 helix_loader [DEBUG] Located configuration folders: []
2023-01-30T12:32:36.588 helix_view::clipboard [INFO] Using wl-copy+wl-paste to interact with the system and selection (primary) clipboard
2023-01-30T12:32:36.624 mio::poll [TRACE] registering event source with poller: token=Token(0), interests=READABLE | WRITABLE
2023-01-30T12:32:36.624 mio::poll [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
2023-01-30T12:32:36.624 mio::poll [TRACE] registering event source with poller: token=Token(2), interests=READABLE | WRITABLE
2023-01-30T12:32:36.624 helix_view::editor [DEBUG] editor status: Loaded 1 files.
2023-01-30T12:32:36.624 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"initialize","params":... 
 cut short for conciseness
2023-01-30T12:32:36.624 mio::poll [TRACE] registering event source with poller: token=Token(3), interests=READABLE | WRITABLE
[...]

PR #5850, opened 2023-02-06
Exiting during LSP initialization fixed

misiasty3:
Quick fix for issue #5732, by sending exit notification to language servers that are not yet initialized.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -407,6 +407,19 @@ impl Client {
 
     /// Forcefully shuts down the language server ignoring any errors.
     pub async fn force_shutdown(&self) -> Result<()> {
+        if !self.is_initialized() {
+            return match self.exit().await {
+                Err(e) => {
+                    log::warn!(
+                        "language server failed to exit during initialization - {}",
+                        e
+                    );
+                    return Err(e);
+                }
+                _ => Ok(()),
+            };
+        }
+
         if let Err(e) = self.shutdown().await {
             log::warn!("language server failed to terminate gracefully - {}", e);
         }


pascalkuthe:
This change is not the right fix for multiple reasons.
In the `send` function in `transport.rs` we drop any `notifications` and delay any shutdown requests as long as the server is not initalized so this new exit notifcation will never be send. This happens in #5732  as indicated by the `Language server not initialized, delaying request` log message.

The right fix is to always send exit notifcations and shutdown requests when the server is not initialized. Simply add a specical conditions in transports.rs:371` something like `&& !is_exit_or_shutdown()`.

The reason we should shutdown too is that we are actually kind of racing the server initalization here so the server might finish initalization before we know about.
To still properly shutdown the server in that case we must always send shutdown.
If the server is not initalized this is not problem as  the server will simply drop the request in that case. From the LSP standard:

> For a request the response should be an error with code: -32002. The message can be picked by the server

With that change to the `send` function the only other required function is to change the `shutdown_and_exit` so that 
``` rust
pub async fn shutdown_and_exit(&self) -> Result<()> {
    if let Err(e) = self.shutdown().await {
	if self.initalized && matches!(e, Err(Error::Rpc(jsonrpc::Error{code: jsonrpc::ErrorCode::ServerError(-32002), ..) {
             return Err(e);
        }
    }
    self.shutdown().await?;
}
````

pascalkuthe:
Ah actually reading further in the spec I am not quite sure if this is the correct solution afterall:

> Until the server has responded to the initialize request with an [InitializeResult](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initializeResult), the client must not send any additional requests or notifications to the server

That means we could send an `exit` notification before even asking the server to initialize but that seems very nieche and is not what #5732 is about (the initalize request gets send basically instantly). The initialization itself takes too long (so after `initialize` is send). We are not allowed to send anything in that case. I think in that case just timing out is actually the correct thing to do.

Nothing we can really do about this I believe. So helix is behaving correctly here, might be interesting to look into what other editors in this case tough.

pascalkuthe:
See https://github.com/Microsoft/language-server-protocol/issues/246 pretty sure this is a current limitation of the LSP spec and there is nothing we can do about this.

the-mikedavis:
This is fixed in https://github.com/helix-editor/helix/pull/7449: now we follow neovim's behavior and kill the language server when shutting down during initialization.

PR #5851, opened 2023-02-06
Short-circuit the word and treesitter object movement commands

trink:
The loop always iterates the number of times the user specified even if the beginning/end of the document is reached.

For an extreme demonstration try the following commands, Helix will hang for several seconds.
100000000w
100000000]c

### DIFF
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -227,9 +227,15 @@ fn word_move(slice: RopeSlice, range: Range, count: usize, target: WordMotionTar
     };
 
     // Do the main work.
-    (0..count).fold(start_range, |r, _| {
-        slice.chars_at(r.head).range_to_target(target, r)
-    })
+    let mut range = start_range;
+    for _ in 0..count {
+        let next_range = slice.chars_at(range.head).range_to_target(target, range);
+        if range == next_range {
+            break;
+        }
+        range = next_range;
+    }
+    range
 }
 
 pub fn move_prev_paragraph(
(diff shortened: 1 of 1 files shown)


pascalkuthe:
LGTM, small and simple fix for a pathological case :+1:   

We might want to do something similar for other movements like `move_next_paragraph` and `move_prev_paragraph` too? It's not as critical there as the loop is quite a bit cheapter but `9999999999999]p` also seems to take a lot longer than it should.

pascalkuthe (approved):
LGTM now, nice catch. It's always good to fixup things that can accidently lockup the UI thread

helix-core/src/movement.rs
```diff
@@ -227,9 +227,15 @@ fn word_move(slice: RopeSlice, range: Range, count: usize, target: WordMotionTar
...
-    })
+    let mut r = start_range;
+    for _ in 0..count {
+        let tmp = slice.chars_at(r.head).range_to_target(target, r);
+        if r == tmp {
+            break;
+        }
+        r = tmp;
+    }
+    r
```
the-mikedavis, line 238:
```suggestion
    let mut range = start_range;
    for _ in 0..count {
        let next_range = slice.chars_at(range.head).range_to_target(target, range);
        if range == next_range {
            break;
        }
        range = next_range;
    }
    range
```

the-mikedavis (commented):
Just a minor style nit, otherwise I think this looks good

the-mikedavis (approved):

PR #5853, opened 2023-02-06
Change `.git` from hard-coded into a configurable project root marker

### LINKED ISSUE #5852: Do not auto-stop root dir detection at `.git`; make `.git` just another configurable root marker

LeoniePhiline:
Fixes #5852 

While `.git` should be a root marker, it might not actually signify the project root, when nested packages are used.

This PR changes the `find_root()` logic to treat `.git` as any other root marker; searching up the file tree and returning the outermost match.

This PR migrates all existing language configurations, adding `.git` to `roots`.

### DIFF
--- a/helix-core/src/lib.rs
+++ b/helix-core/src/lib.rs
@@ -70,15 +70,6 @@ pub fn find_root(root: Option<&str>, root_markers: &[String]) -> std::path::Path
         {
             top_marker = Some(ancestor);
         }
-
-        if ancestor.join(".git").exists() {
-            // Top marker is repo root if not root marker was detected yet
-            if top_marker.is_none() {
-                top_marker = Some(ancestor);
-            }
-            // Don't go higher than repo if we're in one
-            break;
-        }
     }
 
     // Return the found top marker or the current_dir as fallback
--- a/languages.toml
+++ b/languages.toml
@@ -6,7 +6,7 @@ name = "rust"
 scope = "source.rust"
 injection-regex = "rust"
 file-types = ["rs"]
-roots = ["Cargo.toml", "Cargo.lock"]
+roots = ["Cargo.toml", "Cargo.lock", ".git"]
 auto-format = true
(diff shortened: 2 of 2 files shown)


pascalkuthe:
This is not the right fix and regresses a ton of different behaviour and blocks future changes, see https://github.com/helix-editor/helix/issues/5852#issuecomment-1419853306. We should instead resolve the workspace root from the CWD and LSP roots from should not detect `.git` and instead use the already resolved workspace dir as a ceiling

LeoniePhiline:
Thanks for having a look at this! I will close this PR.

If there is anything I can help with #5748, please let me know!

PR #5860, opened 2023-02-06
Show code point in statusline

jakoschiko:
This PR adds an option to show the code point under the cursor in the statusline. The code point is formatted using "U+" notation.
Screenshot_20230207_002602

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -108,6 +108,7 @@ The following statusline elements can be configured:
 | `primary-selection-length` | The number of characters currently in primary selection |
 | `position` | The cursor position |
 | `position-percentage` | The cursor position as a percentage of the total number of lines |
+| `code-point` | The code point under the cursor |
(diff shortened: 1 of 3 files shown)


archseer:
Do we need a persistent statusline element? We recently merged `:character-info` https://github.com/helix-editor/helix/pull/4000

jakoschiko:
My use case: Checking for unexpected characters after pasting text from the internet.

The statusline would be more convient because I can simply move the cursor while looking at the statusline. But the command would still improve my current situation.

tgulacsi:
A copy-paste sometimes ends with \x00 - Vi shows it as ^@, helix as space.

    echo -e "this is \\\\x00\x00:" | hx
    
the cursor vanishes over there, but my problem is that it is not shown that it is not a normal space.

pascalkuthe:
I am going to close this as stale. I don't see us adding a statusline element for this as its fairly niche and we already have `:character-info`.

Thank you for contributing!

PR #5862, opened 2023-02-07
docs: describe tab-width and unit subkeys

cinerea0:
This is a slight addition that explains what the `tab-width` and `unit` subkeys refer to in `languages.toml`. You can figure this out by reading the default configuration, but I thought it would be useful to make it more explicit.

I didn't see any commit style guidelines; let me know if anything needs to be changed.


pascalkuthe (changes requested):
Definitely good to add documentation here, I answered questions about these on matrix before 👍 

I think the description for `unit` can be improved. There is no description what `unit` actually does (it's the text that is inserted into the documented when indentintly). It also sounds like it isntypically either set to a single whitespace or a tab. It should be clear that it's typically set to a tab or N whitespaces (instead of just a single whitespace).

PR #5866, opened 2023-02-07
Adjust everforest to resemble original more closely

basbebe:
Adjust palettes and assignments:

- Color palettes of upstream everforest where tweaked since creation of this port:
  - https://github.com/sainnhe/everforest/pull/108
  - https://github.com/sainnhe/everforest/pull/109

- Move the helix everforest theme closer to the documented
 upstream vim theme

### DIFF
--- a/runtime/themes/everforest_dark.toml
+++ b/runtime/themes/everforest_dark.toml
@@ -1,81 +1,93 @@
-# Everforest (Dark Hard)
-# Author: CptPotato
+# Everforest (Dark Medium)
+# Authors: CptPotato, basbebe
 
 # Original Author:
 # URL: https://github.com/sainnhe/everforest
-# Filename: autoload/everforest.vim
+# Filename: colors/everforest.vim
 # Author: sainnhe
 # Email: sainnhe@gmail.com
 # License: MIT License
 
-"type" = "yellow"
-"constant" = "purple"
+"type" = { fg = "yellow", modifiers = ["italic"] }
+"constant" = "fg"
+"constant.builtin" = { fg = "purple", modifiers = ["italic"] }
+"constant.builtin.boolean" = "purple"
 "constant.numeric" = "purple"
-"constant.character.escape" = "orange"
-"string" = "green"
-"string.regexp" = "blue"
-"comment" = "grey0"
+"constant.character.escape" = "green"
+"string" = "aqua"
+"string.regexp" = "green"
+"string.special" = "yellow"
+"comment" = { fg = "grey1", modifiers = ["italic"] }
 "variable" = "fg"
-"variable.builtin" = "blue"
+"variable.builtin" = { fg = "purple", modifiers = ["italic"] }
 "variable.parameter" = "fg"
-"variable.other.member" = "fg"
-"label" = "aqua"
+"variable.other.member" = "blue"
+"label" = "orange"
 "punctuation" = "grey2"
-"punctuation.delimiter" = "grey2"
+"punctuation.delimiter" = "grey1"
 "punctuation.bracket" = "fg"
+"punctuation.special" = "blue"
 "keyword" = "red"
-"keyword.directive" = "aqua"
+"keyword.operator" = "orange"
+"keyword.directive" = "purple"
+"keyword.storage" = "orange"
 "operator" = "orange"
 "function" = "green"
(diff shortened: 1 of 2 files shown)


the-mikedavis:
\cc @CptPotato what do you think?

CptPotato:
I'm fine with this change, didn't know the colors were changed at some point :+1:

When I created the port I didn't follow the original theme super closely because I preferred this look. I'm a bit on the fence about which version I like better but for the "official" themes that are bundled with helix it makes sense to adhere more to the original, so I'd say go ahead.

CptPotato:
Here are some nits, but feel free to disregard any (especially those that would deviate from the original theme again, I just thought about sharing my perspective):

<details>
<summary><b>I think modules and types having the same color makes them harder to discern</b></summary>

old:
image

this PR:
image
</details>

<details>
<summary><b>`let` is not actually orange in the original theme</b></summary>

original:
image
[screenshot source](https://user-images.githubusercontent.com/37491630/206063921-58418bb0-7752-43f3-9f3b-f3752f8ee753.png)

this PR:
image
</details>

<details>
<summary><b>Italic may look bad with some fonts</b></summary>

This is debatable and up to personal preference but I initially didn't include italic text because it can look pretty ugly depending on the font or terminal, like with Fira Code for example:
image

It also looks like there's more italic text in this PR than in the original in general. This includes types and comments for example. I actually think with a nice font this can look pretty neat (I'm using italic comments myself) but it's different from the original.
</details>

---

Lastly, I think one difficulty is that vim just highlights code differently in some cases, sometimes it's even inconsistent within vim (one screenshot shows a `write!()` call in `green`, but an `assert_eq!()` call using `aqua` for example).

That being said, I still stand by the opinion that following the original somewhate closely is desirable (since that is what most users would expect I guess). Though I would at least change the color of `let` bindings from `orange` to `red`.

basbebe:
Thanks for the feedback.
While I used gruvbox for a long time I really like this theme now, given it is similar in its qualities but has a slightly brighter background (light version) and more vivid colors.

While the original vim theme has many small adjustments for different languages, I tried to stay with what is defined in the code for the general treesitter attributes (see [everforest/everforest.vim at master · sainnhe/everforest](https://github.com/sainnhe/everforest/blob/master/colors/everforest.vim)) since those are also close to what helix uses.

Some but not all of the code is mirrored / documented in https://github.com/sainnhe/everforest/blob/master/palette.md

The vscode theme with the same name (same author!) seems to deviate a lot from the vim one.

I made a new commit (see below) which I could further tweak or squash depending on your feedback.

> **I think modules and types having the same color and style makes them harder to discern**

I decided to stick with the original theme there at first which defines `YellowItalic` for both of those (using more nuanced definitions in some but not all languages).
My solution would be to use yellow _without_ italics for types, which seems to also have been in the original until https://github.com/sainnhe/everforest/commit/62408f6cb189606bc4b16832ada29be3ae33515f (also resembled in the screenshots).

> **`let` is not actually orange in the original theme**

Thanks, fixed that oversight

> **Italic may look bad with some fonts**

I see how this leaves a lot of room for debate.  
The original has optional support for italics (which also includes italics for comments) and I decided to go with them due to personal preference and having seen italics used in other helix themes.

Fira Code particularly will be problematic in this case since it doesn't have italics natively (?) and needs special setup in the terminal to produce "proper" italics (like derived from another font).

I don't know what your preferred solution is for italics:
- Provide separate _italics_ theme
- Include italics
- Omit italics

CptPotato:
Thanks for elaborating!

> My solution would be to use yellow without italics for types

Sounds like a good compromise to me :+1:

> I don't know what your preferred solution is for italics:
>
> - Provide separate italics theme
> - Include italics
> - Omit italics

I'd say keep the italic modifiers and don't bother creating variants. I don't see this "issue" as a reason to remove them. To enable/disable italics it's probably a better idea to provide a editor config option, similar to global theme overrides that have been proposed.

> I made a new commit (see below) which I could further tweak or squash depending on your feedback.

There aren't any issues left that stick out to me, so I think this is good to go now.
Squashing can be done by the maintainer that merges the PR if I remember correctly.

basbebe:
Rebased and added styling of inlay-hints and soft-wrap indicator

the-mikedavis (approved):

PR #5872, opened 2023-02-07
update bug report template

pascalkuthe:
Added some hints to the bug report template to (hopefully) reduce the number of inactionable bug reports.

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.yaml
+++ b/.github/ISSUE_TEMPLATE/bug_report.yaml
@@ -32,7 +32,7 @@ body:
     id: helix-log
     attributes:
       label: Helix log
-      description: See `hx -h` for log file path
+      description: See `hx -h` for log file path. If you can reproduce the issue run `RUST_BACKTRACE=1 hx -vv` to generate a more detailed log file.
       value: |
         <details><summary>~/.cache/helix/helix.log</summary>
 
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -32,7 +32,7 @@ body:
     id: helix-log
     attributes:
       label: Helix log
-      description: See `hx -h` for log file path
+      description: See `hx -h` for log file path, if you can reproduce the issue run `RUST_BACKTRACE=1 helix -v` to generate a more detailed log file.
```
the-mikedavis, line 35:
```suggestion
      description: See `hx -h` for log file path. If you can reproduce the issue run `RUST_BACKTRACE=1 hx -v` to generate a more detailed log file.
```

dead10ck: Also, I'd err on the side of more verbosity with `-vvvv`

dead10ck (approved):
Lgtm!

.github/ISSUE_TEMPLATE/bug_report.yaml
```diff
@@ -32,7 +32,7 @@ body:
     id: helix-log
     attributes:
       label: Helix log
-      description: See `hx -h` for log file path
+      description: See `hx -h` for log file path. If you can reproduce the issue run `RUST_BACKTRACE=1 hx -vvvv` to generate a more detailed log file.
```
archseer, line 35: vvvv is a bit extreme since you'll get logs for tokio internals. Maybe v or vv?

pascalkuthe: I changed it to `vv`. I originally used `v` but @dead10ck suggested the more verbose version. I agree that including tracing output is probably too much so `vv` seems more appropriate.

archseer (approved):

PR #5873, opened 2023-02-07
Add `Containerfile` file-type for dockerfile language

TheAlgorythm:
The `Containerfile` is a vendor independent name for the `Dockerfile` when using things like `podman`. It has the same syntax and therefore it is possible to extend the `file-types`.
But I am not sure whether it would be better to define a different language with the same parameters. It would better show being vendor independent but also a bit more bloat.
Further I am not sure why there is a lowercase `dockerfile` in the `file-types` and if it is also needed for the `Containerfile`.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1072,8 +1072,8 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-scala", rev = "f6bb
 name = "dockerfile"
 scope = "source.dockerfile"
 injection-regex = "docker|dockerfile"
-roots = ["Dockerfile"]
-file-types = ["Dockerfile", "dockerfile"]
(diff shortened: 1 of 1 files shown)


the-mikedavis:
`.dockerfile` is a less common file extension for dockerfiles, for example `build.dockerfile`. It looks like `.containerfile` is used by some projects (see https://github.com/search?q=path%3A*.containerfile&type=code) so it probably makes sense to add it here as well

TheAlgorythm:
In my experience this file extension always started uppercase but adding the lowercase one shouldn't hurt. So I did add it.

the-mikedavis (approved):

Issue #5758, opened 2023-01-31
Allow to position status line on the top


sporto:
When I want to duplicate a buffer I cannot see the path of the current buffer anymore:

I see the path of the current buffer in the status line

![Screenshot at Feb 01 08-36-58](https://user-images.githubusercontent.com/1005498/215889744-0618dee4-9470-4768-8597-927d41e14339.png)

Then when I type `:w` I cannot see the path of the buffer anymore as it is obscured by the popup.

![Screenshot at Feb 01 08-37-43](https://user-images.githubusercontent.com/1005498/215889882-01741c6f-4b07-4065-8951-0b635e36bfde.png)

I would like to keep the path of the buffer visible, so I can copy part of the path.

So having an option to position the status line at the top would be great for this. So it is always visible.

CptPotato:
> I would like to keep the path of the buffer visible, so I can copy part of the path.

If your goal is to copy/paste the path up to the parent directory, #3134 *might* help you once implemented. However there hasn't been much development on this issue in a while.

yashpalgoyal1304:
alternates:
1. put status line on top of the completion suggestiion popup....
2. put the command line above the status line

an example of this point 1. above can be seen from `micro` editor:


https://user-images.githubusercontent.com/19423063/216253766-95b2c3aa-8f67-4f6d-978e-61c5d64a3b31.mp4

yashpalgoyal1304:
just to confirm, the 2nd alternate suggested above (put command line above status line) would also solve the _problem_ underlined in the description without actually implementing the _solution_ given in title...

so, would u still consider this solved or is only the proposed solution your focus??

sporto:
> so, would u still consider this solved or is only the proposed solution your focus??

Yes, that is all I need, for the status bar not be obscured

Fl1tzi:
Hi. While trying to do this I think it's actually not ideal.

If you have the prompt with the suggestions, the bar will look weird when it overflows over the popup and it would also move a lot, which I think makes it harder to use. So I came up with the idea to actually put the statusline under the prompt. The statusline would stay static and will be easy to see and you also don't get distracted by the suggestions. That is at least what I think would be the best option.

Would you also consider this?

yashpalgoyal1304:
> So I came up with the idea to actually put the statusline under the prompt. The statusline would stay static

hi @Fl1tzi , yes this was already suggested (quoting below). thanks a lot for trying out and sharing the screenshot.

> 2. put the command line above the status line
[...]

PR #5882, opened 2023-02-08
Switch statusline & prompt

Fl1tzi:
#5758 

This switches the position of the statusline with the position of the prompt to improve readability of the statusline while the prompt is opened.

TODO:
- [ ] split not rendering correctly

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -224,10 +224,7 @@ impl EditorView {
 
         Self::render_diagnostics(doc, view, inner, surface, theme);
 
-        let statusline_area = view
-            .area
-            .clip_top(view.area.height.saturating_sub(1))
-            .clip_bottom(1); // -1 from bottom to remove commandline
+        let statusline_area = view.area.clip_top(view.area.height);
 
         let mut context =
             statusline::RenderContext::new(editor, doc, view, is_focused, &self.spinners);
@@ -1388,7 +1385,7 @@ impl Component for EditorView {
 
             surface.set_string(
                 area.x,
-                area.y + area.height.saturating_sub(1),
+                area.y + area.height.saturating_sub(2),
                 status_msg,
                 style,
             );
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -372,7 +372,7 @@ impl Prompt {
 
         let completion_area = Rect::new(
             area.x,
-            (area.height - height).saturating_sub(1),
+            (area.height - height).saturating_sub(2),
             area.width,
             height,
         );
@@ -447,7 +447,7 @@ impl Prompt {
             text.render(inner, surface, cx);
         }
 
-        let line = area.height - 1;
+        let line = area.height - 2;
         // render buffer text
         surface.set_string(area.x, area.y + line, &self.prompt, prompt_color);
 
@@ -639,7 +639,7 @@ impl Component for Prompt {
     }
 
     fn cursor(&self, area: Rect, _editor: &Editor) -> (Option<Position>, CursorKind) {
-        let line = area.height as usize - 1;
+        let line = area.height as usize - 2;
         (
             Some(Position::new(
                 area.y as usize + line,


CptPotato:
I think this should have a config option. I personally prefer the current layout.

yashpalgoyal1304:
> [NOT READY]

u can mark it as draft

yashpalgoyal1304:
having this as a config option will make themes inconsistent.
- do u prefer current just due to habit or is there an actual functionality that it has (which proposed one doesn't have)? 
- have u tried the proposed layout?
- the current layout comes with so much habit being used in neo/vi/m lineage. so, it's natural that a change will feel odd... 


but for the proposed layout, there's actually a functionality concern - to avoid the blockage of status bar from the popups of command line

Plasma-Vortex:
Alternatively, we could just move the popup one line up, so that it appears above both the status line and command line

Pros:
- Readability: Popup does not cover the status and command line
- Habit: Status line above command line is consistent with vim/neovim layout, so users switching to Helix will be familiar with it

Cons:
- Command popup isn't as close to the command line - related components in a UI should generally be near each other. Personally I think having an extra line in between won't change much, but it'd be best to actually test this layout first.

yashpalgoyal1304:
Just stumbled upon one more pro for having the command line on top:

- if the errors in executing a command are long, then expansion of command line _can_ be a viable option for future. 
- currently, if command line were to span multiple lines - it will again - be very unpleasant with status bar.

Example: 
- try typing `:cd C:\Prog` `<tab>` `<ret>` 
- i.e. execute command `:cd C:\"Program Files"`

It will show this 128 character long error with `helix 22.12 (96ff64a8)`

```
Couldn't change the current working directory: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

quomatz:
This is a breaking UI change. This should be escalated to a comnunity discussion.
While I see this as 100% benefitial, I understand it breaks the layout loved by many users and will probably be rejected, to my regret :( 
Having config option just for this is also confusing, it would be better to have a robust layout configuration, when you could e.g. hide the status bar completely (think zen mode) just as with the upper bufferline.

helix-term/src/ui/prompt.rs
```diff
@@ -447,7 +447,7 @@ impl Prompt {
             text.render(inner, surface, cx);
         }
 
-        let line = area.height - 1;
+        let line = area.height - 2;
```
jwidauer, line 450: This and the following change on line 642 break the file picker (and probably other prompts as well), since they can underflow.
If you substitute this and the change on line 642 with `saturating_sub(2)` this seems to fix it.

helix-term/src/ui/editor.rs
```diff
@@ -224,10 +224,7 @@ impl EditorView {
 
         Self::render_diagnostics(doc, view, inner, surface, theme);
 
-        let statusline_area = view
-            .area
-            .clip_top(view.area.height.saturating_sub(1))
-            .clip_bottom(1); // -1 from bottom to remove commandline
+        let statusline_area = view.area.clip_top(view.area.height);
```
jwidauer, line 227: This introduces the issue with split views. The easiest solution I could come up with was something like this:

```rust
// if we're not at the bottom of the screen, draw the statusline higher, otherwise
// it'll overlap with the next view
let statusline_area = if (viewport.bottom() - 1) == view.area.bottom() {
    view.area.clip_top(view.area.height)
} else {
    view.area
        .clip_top(view.area.height.saturating_sub(1))
        .clip_bottom(1)
};
```

debajyoti1990:
Convenient as it might be for some users, in my humble opinion this should not be default, rather a config option.

ghc0:
Also it should be customizable  to show the status line on the top of line or other widgets attached on screen.

Also its show or not could be switched by hot key. 

Reasons are as follows:
1. when use a laptop, the surface of bottom line on screen has a big angle to eye, it is not on surface of keyboard, it is another surface.
2. Move it to top line is more convenient since its info frequently checked. Or some other place like right position of screen, and use left shift+ right shift to display or not display them as a switch.

jhugs:
What's the status on this? New to helix, and I'd definitely like to be able to _see_ the statusline while typing in the prompt. How / where / or whether it's a config option I don't feel strongly about, I just want to be able to see it

pascalkuthe:
I don't think we are going forward with this, see discussion in #7635

Issue #5716, opened 2023-01-29
Misleading message: Shell command failed


lesleyrs:
I noticed on windows that when you run `:sh explorer . ` it shows: https://github.com/helix-editor/helix/blob/482cc22fecc109773882d3e0fd6251c33862e89e/helix-term/src/commands.rs#L4920

It is only a visual issue though and still opens up the folder.

Unsure to how many other commands this applies, but it does actually return an exit code of 1 so it does makes sense.

I looked at how nvim went about this and it only gave the message `[Process exited 1]` which is a more neutral way of doing things.

What to do about this depends on how many commands act this way I guess. Would also be fine to leave as it is and close this issue, just felt like I should mention this.

PR #5885, opened 2023-02-08
Change misleading message

lesleyrs:
Closes https://github.com/helix-editor/helix/issues/5716

Simple text change copied from nvim so running commands like `explorer` which opens up file explorer on Windows doesn't explicitly say failed when it actually works.

New:

image

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -5035,7 +5035,7 @@ async fn shell_impl_async(
             log::error!("Shell error: {}", err);
             bail!("Shell error: {}", err);
         }
-        bail!("Shell command failed");
+        bail!("[Process exited 1]");
     } else if !output.stderr.is_empty() {
         log::debug!(
             "Command printed to stderr: {}",


archseer:
It might have not exited with exit code 1 though.

archseer:
I don't see how this could be misleading, the command **did** fail.

CptPotato:
This seems to be an issue with explorer. You can also use `start` on Windows to open a directory or file by it's. This command also returns a valid exit code (non zero if the path or file wasn't found). For example `:sh start .` will open the current directory.

On the other hand, outputting the *correct* exit code would be helpful in some cases.

pascalkuthe:
> On the other hand, outputting the correct exit code would be helpful in some cases.

I agree, correlating exit code != 0 to failure is the right assumption most of the time so the message should still indicate a failure but actually displaying the statuscode code be helpful. For example to google the problem with whatever program you are calling. Ideally we would display stdout/stderr in a popup aswell but that can be a seperate PR. 

Just changing the error message to harcode `1` is actually worse than the current error since the process might not have exited with 1 but 2 instead `Command::success` just checks `status != 0`.

I think `Shell Command Failed: Status <X>` where `<X>` is the actual status code and not just hard-coded to `1` would be a good change, for the weird edgecases that a process returns a non-zero access code even without failure this seems like an improvement.

Also @CptPotato is right that you should be using `start` instead of `explorer` directly (that will also return the right exitcode).

lesleyrs:
Thanks guys `:sh start .` gives me `Command succeed` unlike `:sh explorer .`!

And yea I was assuming failed command would always return 1 my bad.

> I think Shell Command Failed: Status <X> where <X> is the actual status code and not just hard-coded to 1 would be a good change, for the weird edgecases that a process returns a non-zero access code even without failure this seems like an improvement.

I like this, idk how to go about doing it though. I'll close my stuff and leave it for somebody else to make a new pr.

CptPotato:
If you still want to take this on, it should be quite simple:

`output.status` is the program's [`ExitStatus`](https://doc.rust-lang.org/std/process/struct.ExitStatus.html) which lets you retrieve the exit code using `status.code()`.

You can display that code using string interpolation in `bail!`: `bail!("blabla {}", exit_code)`. You only have to handle the case where `status.code()` returns None (print a message without exit code).

PR #5894, opened 2023-02-09
Negotiate LSP Position Encoding

pascalkuthe:
So far LSP always required that `PositionEncoding.characters` is an UTF-16 offset. Now that LSP 3.17 is available in `lsp-types` request the server to send char offsets (UTF-32) or byte offsets (UTF-8) instead. For compatability with servers that don't support other encodings, UTF-16 is used as a fallback as required by the standard.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -6,6 +6,7 @@ use crate::{
 
 use helix_core::{find_root, ChangeSet, Rope};
 use helix_loader::{self, VERSION_AND_GIT_HASH};
+use lsp::PositionEncodingKind;
 use lsp_types as lsp;
 use serde::Deserialize;
 use serde_json::Value;
@@ -32,7 +33,6 @@ pub struct Client {
     server_tx: UnboundedSender<Payload>,
     request_counter: AtomicU64,
     pub(crate) capabilities: OnceCell<lsp::ServerCapabilities>,
-    offset_encoding: OffsetEncoding,
     config: Option<Value>,
     root_path: std::path::PathBuf,
     root_uri: Option<lsp::Url>,
@@ -104,7 +104,6 @@ impl Client {
             server_tx,
             request_counter: AtomicU64::new(0),
             capabilities: OnceCell::new(),
-            offset_encoding: OffsetEncoding::Utf16,
             config,
             req_timeout,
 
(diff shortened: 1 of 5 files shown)


helix-lsp/src/client.rs
```diff
@@ -147,7 +146,19 @@ impl Client {
...
-        self.offset_encoding
+        self.capabilities()
+            .position_encoding
+            .as_ref()
+            .and_then(|encoding| match encoding.as_str() {
+                "utf-8" => Some(OffsetEncoding::Utf8),
+                "utf-16" => Some(OffsetEncoding::Utf16),
+                "utf-32" => Some(OffsetEncoding::Utf32),
+                encoding => {
+                    log::error!("Server provided invalid position_encording {encoding}, defaulting to utf-16");
```
the-mikedavis, line 157:
```suggestion
                    log::error!("Server provided invalid position_encoding {encoding}, defaulting to utf-16");
```

helix-lsp/src/lib.rs
```diff
@@ -181,6 +186,7 @@ pub mod util {
         match offset_encoding {
             OffsetEncoding::Utf8 => doc.try_byte_to_char(pos).ok(),
             OffsetEncoding::Utf16 => doc.try_utf16_cu_to_char(pos).ok(),
+            OffsetEncoding::Utf32 => Some(pos),
```
the-mikedavis, line 189: The "TODO" above this `match` can probably be removed with this change, eh?

helix-lsp/src/lib.rs
```diff
@@ -205,6 +211,13 @@ pub mod util {
                 let line_start = doc.char_to_utf16_cu(doc.line_to_char(line));
                 let col = doc.char_to_utf16_cu(pos) - line_start;
 
+                lsp::Position::new(line as u32, col as u32)
+            }
+            OffsetEncoding::Utf32 => {
+                let line = doc.char_to_line(pos);
+                let line_start = doc.line_to_byte(line);
```
the-mikedavis, line 218: For UTF-32 shouldn't this be `doc.line_to_char(line)` instead?

pascalkuthe: ah good catch! I forgot to change one occurrence here thanks :)

helix-lsp/src/lib.rs
pascalkuthe: Good point, forgot about that :+1:

the-mikedavis (approved):

PR #5895, opened 2023-02-09
Show current language when no argument is provided

LEI:
The command `:set-language` and `:lang` now behaves like `:theme`, as mentioned [here](https://github.com/helix-editor/helix/discussions/5847#discussioncomment-4883555).

- Should the default value "text" be used, like in the statusline, maybe define a constant?
- I expected `:language` to work, is there any reason against adding it as an alias?

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -6,6 +6,7 @@ use crate::job::Job;
 use super::*;
 
 use helix_core::encoding;
+use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::editor::{Action, CloseError, ConfigEvent};
 use ui::completers::{self, Completer};
 
@@ -1656,13 +1657,20 @@ fn language(
         return Ok(());
     }
 
+    if args.is_empty() {
+        let doc = doc!(cx.editor);
+        let language = &doc.language_name().unwrap_or(DEFAULT_LANGUAGE_NAME);
+        cx.editor.set_status(language.to_string());
+        return Ok(());
+    }
+
     if args.len() != 1 {
         anyhow::bail!("Bad arguments. Usage: `:set-language language`");
     }
 
     let doc = doc_mut!(cx.editor);
 
-    if args[0] == "text" {
+    if args[0] == DEFAULT_LANGUAGE_NAME {
         doc.set_language(None, None)
     } else {
         doc.set_language_by_language_id(&args[0], cx.editor.syn_loader.clone())?;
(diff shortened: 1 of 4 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -1656,6 +1656,13 @@ fn language(
         return Ok(());
     }
 
+    if args.len() == 0 {
+        let doc = doc!(cx.editor);
+        let language = &doc.language_name().unwrap_or("");
```
the-mikedavis, line 1661:
```suggestion
        let language = &doc.language_name().unwrap_or("text");
```

Since we allow setting the language as `text` to clear the language config.

Also see here in the statusline element for the current buffer language: https://github.com/helix-editor/helix/blob/8a3ec443f176218384db8c8610bbada9c43a6ea5/helix-term/src/ui/statusline.rs#L401-L408

the-mikedavis (changes requested):

helix-term/src/commands/typed.rs
LEI: I defined the constant `DEFAULT_LANGUAGE` in `helix-core/src/lib.rs`, maybe there is a better place? I figured language is not specific to `helix-term`.

the-mikedavis: Ah nice, I like the re-use of the constant 👍

I think it might belong in `helix-view/src/document.rs` alongside the `DEFAULT_INDENT`

helix-term/src/commands/typed.rs
```diff
@@ -1656,13 +1656,20 @@ fn language(
         return Ok(());
     }
 
+    if args.len() == 0 {
+        let doc = doc!(cx.editor);
+        let language = &doc.language_name().unwrap_or(DEFAULT_LANGUAGE);
+        cx.editor.set_status(language.to_string());
+        return Ok(())
```
the-mikedavis, line 1663:
```suggestion
        return Ok(());
```

helix-term/src/commands/typed.rs
```diff
@@ -1656,13 +1656,20 @@ fn language(
...
+    }
+
     if args.len() != 1 {
         anyhow::bail!("Bad arguments. Usage: `:set-language language`");
     }
 
     let doc = doc_mut!(cx.editor);
 
-    if args[0] == "text" {
+    if args[0] == DEFAULT_LANGUAGE {
```
the-mikedavis, line 1672: small style nit but I think we should call this `DEFAULT_LANGUAGE_NAME`. "Language" is already used in core either to refer to tree_sitter::Language or to language configuration

the-mikedavis (approved):

PR #5911, opened 2023-02-11
Added force_reload and fore_reload_all

Pablu23:
Normal reload functions now bail if they arent saved. 
Reload-all still reloads all saved files.

Didnt add tests because for the already existing reload functions there werent any.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1160,8 +1160,7 @@ fn get_character_info(
     Ok(())
 }
 
-/// Reload the [`Document`] from its source file.
-fn reload(
+fn force_reload(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
     event: PromptEvent,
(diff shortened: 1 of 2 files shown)


sudormrfbin:
The commands should be renamed to `:reload!` and `:reload-all!` following the conventions of `:q` and `:q!`.

Pablu23:
I am not quite sure what you mean. In the editor itself you use :reload! and :reload-all! respectivly to force the reload without saving.

helix-term/src/commands/typed.rs
```diff
@@ -1173,12 +1191,67 @@ fn reload(
...
+            let view = view_mut!(cx.editor, view_id);
+            if view.doc.eq(&doc_id) {
+                view.ensure_cursor_in_view(doc, scrolloff);
+            }
+        }
+    }
+
+    Ok(())
+}
+
```
pascalkuthe, line 1254: instead of copying the command you should simply add ann inner function that accepts an `force: bool` argument

helix-term/src/commands/typed.rs
```diff
@@ -1160,6 +1160,24 @@ fn get_character_info(
...
+    let scrolloff = cx.editor.config().scrolloff;
+    let redraw_handle = cx.editor.redraw_handle.clone();
+    let (view, doc) = current!(cx.editor);
+    doc.reload(view, &cx.editor.diff_providers, redraw_handle)
+        .map(|_| {
+            view.ensure_cursor_in_view(doc, scrolloff);
+        })
+}
+
```
pascalkuthe, line 1180: same here as below

pascalkuthe (changes requested):

Issue #3406, opened 2022-08-13
Indent and dedent don't match


Zoybean:
### Summary

In a file with mixed indent amounts, and a language set (I tried with bash an rust, I'm not sure what the exact conditions are), indent amounts do not match dedent amounts.

### Reproduction Steps

 

With the following file open:

```
test
 file
  with
   mixed
    indents
new line
```

1. `:set-language bash`
2. `ge`
3. `<`
4. `>>`
5. `:set-language rust`
6. `<`
7. `>>>>`

I expected this to happen:

At steps 4 and 7, the indent amount matches the dedent amount in the preceding step

Instead, this happened:

At step 3. the dedent amount is 2 spaces. At step 4, the indent amount is 1 space.
At step 6, the dedent amount is 4 spaces. At step 7, the indent amount is 1 space.

### Helix log

n/a

### Platform

Windows (WSL: ubuntu)

### Terminal Emulator

Windows Terminal Version: 1.14.1962.0

### Helix Version

22.05-367-g63ec1095

AceofSpades5757:
I've had this issue as well, with different languages. Most notably on a pull request branch that's working on Python indentation. So I suspect it's an issue with tree-sitter indent.scm file, but I'm not sure.

Triton171:
I've seen this bug too but it doesn't have anything to do with the `indent.scm` queries since those are only used to determine the indentation of a new line. I guess there is a bug somewhere in the `indent`/`unindent` functions in `helix-term/src/commands.rs`. Both functions are pretty short so it might not be too difficult to find it.

pascalkuthe:
Closed by  #5918, we forgot to link

PR #5918, opened 2023-02-11
Unify usage of indent_style

### LINKED ISSUE #2236: When editing HTML automatically inserted tabs are broken.
### Summary

I have set the editor to adapt indentation, when adding new lines. Whenever I edit HTML, the amount of indents grow exponentially.

### Reproduction Steps

I am using default settings, on an M1 Mac. This happens every time I edit HTML files.

### Helix log

```
2022-04-23T10:10:19.719 helix_view::theme [WARN] Theme: malformed hexcode: foreground
2022-04-23T10:10:19.720 helix_view::theme [WARN] Theme: invalid style attribute: Modifiers
2022-04-23T10:10:19.720 helix_view::theme [WARN] Theme: malformed hexcode: pink
2022-04-23T12:39:03.949 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,948 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'autopep8': No module named 'pycodestyle'\n"
2022-04-23T12:39:03.953 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,953 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'mccabe': No module named 'mccabe'\n"
2022-04-23T12:39:03.954 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,954 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'pycodestyle': No module named 'pycodestyle'\n"
2022-04-23T12:39:03.954 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,954 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'pydocstyle': No module named 'pydocstyle'\n"
2022-04-23T12:39:03.955 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,955 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'pyflakes': No module named 'pyflakes'\n"
2022-04-23T12:39:03.955 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,955 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'pylint': No module named 'pylint'\n"
2022-04-23T12:39:03.956 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,956 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'rope_completion': No module named 'rope'\n"
2022-04-23T12:39:03.956 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,956 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'rope_rename': No module named 'rope'\n"
2022-04-23T12:39:03.956 helix_lsp::transport [ERROR] err <- "2022-04-23 12:39:03,956 CEST - WARNING - pylsp.config.config - Failed to load pylsp entry point 'yapf': No module named 'yapf'\n"
2022-04-23T13:00:24.987 helix_view::editor [ERROR] Failed to initialize the LSP for `text.html.basic` { LSP not defined }
```

</details>

### Platform

Mac

### Terminal Emulator

Warp, iTerm2

### Helix Version

helix 22.03 (d4e45fd4)

Triton171:
Previously, we used sometimes `doc.indent_style` and sometimes `doc.language_config.indent`. This caused some issues if they differ (e.g. if a non-default indent style is auto-detected in a document or manually set). For example, in a markdown file, setting `:indent-style 8` and then continuously pressing enter (starting on an indented line) increases the indent each time. This happens because the indent_style used to detect the indentation level of an existing line is different from the indent_style used to convert an indentation level to a string. The same issue occurs in any language without indent queries.
 
With these changes, `language_config.indent` is only used to initialize `doc.indent_style` and afterwards ignored. Resolves #2236 and #3406.

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -1122,11 +1122,16 @@ impl Document {
         self.syntax.as_ref()
     }
 
-    /// Tab size in columns.
+    /// The width that the tab character is rendered at
     pub fn tab_width(&self) -> usize {
         self.language_config()
             .and_then(|config| config.indent.as_ref())
             .map_or(4, |config| config.tab_width) // fallback to 4 columns
+    }
+
+    // The width (in spaces) of a level of indentation.
+    pub fn indent_width(&self) -> usize {
+        self.indent_style.indent_width(self.tab_width())
     }
 
     pub fn changes(&self) -> &ChangeSet {
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -56,6 +56,14 @@ impl IndentStyle {
             }
         }
     }
+
+    #[inline]
+    pub fn indent_width(&self, tab_width: usize) -> usize {
+        match *self {
+            IndentStyle::Tabs => tab_width,
+            IndentStyle::Spaces(width) => width as usize,
+        }
+    }
 }
 
 /// Attempts to detect the indentation style used in a document.
@@ -177,7 +185,7 @@ pub fn auto_detect_indent_style(document_text: &Rope) -> Option<IndentStyle> {
 
 /// To determine indentation of a newly inserted line, figure out the indentation at the last col
 /// of the previous line.
-pub fn indent_level_for_line(line: RopeSlice, tab_width: usize) -> usize {
+pub fn indent_level_for_line(line: RopeSlice, tab_width: usize, indent_width: usize) -> usize {
     let mut len = 0;
     for ch in line.chars() {
         match ch {
@@ -187,7 +195,7 @@ pub fn indent_level_for_line(line: RopeSlice, tab_width: usize) -> usize {
         }
     }
 
-    len / tab_width
+    len / indent_width
 }
 
 /// Computes for node and all ancestors whether they are the first node on their line.
@@ -466,6 +474,7 @@ fn extend_nodes<'a>(
     text: RopeSlice,
     line: usize,
     tab_width: usize,
+    indent_width: usize,
 ) {
     let mut stop_extend = false;
 
@@ -490,10 +499,12 @@ fn extend_nodes<'a>(
                         if deepest_preceding.end_position().row == line {
                             extend_node = true;
                         } else {
-                            let cursor_indent = indent_level_for_line(text.line(line), tab_width);
+                            let cursor_indent =
+                                indent_level_for_line(text.line(line), tab_width, indent_width);
                             let node_indent = indent_level_for_line(
                                 text.line(deepest_preceding.start_position().row),
                                 tab_width,
+                                indent_width,
                             );
                             if cursor_indent > node_indent {
                                 extend_node = true;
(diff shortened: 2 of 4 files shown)


pascalkuthe (changes requested):
If I understand this change correctly that means that the number of spaces in the indent unit now always corresponds to the tabwidth.

In other words:
* If my document contains both spaces and tabs
* I set the indent width to 4
* Then all tabs are rendered with a width of 4

This does not correspond to the current config whichs allows setting `tab-width`  separately from the `unit` field.

So for example this change would break using 4 spaces for indentation and a `tab-width` of two:

```
indent-style = { unit = "    ", tab-width=2 }
```

Is unifying these two really necessary to fix the underling issue here?

From my understanding, the correct fix would be to have to separate parameters for `indent_level_for_line`
the `tab-width` which is used when `\t` is encoutered and the `indent-width` wich is used for calulating the indent level

Triton171:
Thanks for the super-fast review! You're right, I hadn't really thought about the distinction between `tab_width` (the width the tab character is rendered at) and `indent_width` (the width of an indentation level). It appears that I'm not the only one, as a significant percentage of the usages of `tab_width` in the current master should actually be `indent_width`. I've now made the difference a bit clearer in the code and fixed these bugs.

I also removed the `doc.tab_width()` function and replaced their usages by `doc.indent_style.tab_width`. This hopefully makes it more obvious that there are 2 different widths to consider when computing indentation.

helix-view/src/document.rs
```diff
@@ -1122,13 +1130,6 @@ impl Document {
         self.syntax.as_ref()
     }
```
pascalkuthe, line 1132: With this config removed `tab-width` is not updated on `:config-reload` anymore. I actually think that we should keep this function around and not make `IndentStyle` a struct at all. 

Rather we should add a second function to the document to compute the `indent_width`:
``` suggestion
     /// Tab size in columns.
    pub fn tab_width(&self) -> usize {
        self.language_config()
            .and_then(|config| config.indent.as_ref())
            .map_or(4, |config| config.tab_width) // fallback to 4 columns
     }

    /// Indentation size in columns
    pub fn indent_width(&self) -> usize {
        match self.indent_style{
            IndentStyle::Tab => self.tab_width(),
            IndentStyle::Spaces(width) => width as usize
         }
    }
```

pascalkuthe (changes requested):
Yeah it's definitely a bit subtle :+1: People usually stick to a single indent style they tend to o think about `indent_width` and `tab_width` as the same thing. 

The changes to the autodetection now LGTM :+1: 

However this PR regresses some things like config reloading. I think a simpler fix is possible by simply adding a new `indent_width` function instead of actually changing `IndentStyle`. I added suggestion below.

pascalkuthe (approved):
Thanks for tackling this, LGTM now :+1: 

Ident width and tab width is a subtle but important distinction, great to see that cleared up

Triton171:
Thanks for all the feedback!

PR #5929, opened 2023-02-11
Remove Outdated Tutor Note About Increment and Decrement Behavior

### LINKED ISSUE #5928: Increment and Decrement Behavior When No Number Under Selection
While going through the tutor, I came across "Note: If there isn't a number under the selection the cursor will jump to the next number in the line and act on it". On master, this does not appear to be the case, so it is either a bug or the tutor text should be updated.

Aesthetikx:
Closes #5928.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -830,8 +830,8 @@ lines.
  --> 3) Another point.
  --> 6) Last point.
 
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Thank you!

PR #5930, opened 2023-02-11
keep (cursor) position when exactly replacing text

### LINKED ISSUE #5910: Cursor shifts on multiple selection letter case changes
### Summary

Hi,

I was working on hx tutor file (chapter 10.3) and ran to some issue with cursor shifting right on one of the multi-selected text.

### Reproduction Steps

I tried this:

1. `hx`
2. write: `this sentence should all be in lowercase.`
3. select whole line `x`
4. search in selection with `s`
5. input regex: `[b,e,l]`
6. switch case: `~` couple of times

I expected this to happen:
Only selected characters change case. Cursor remains on those characters.

Instead, this happened:
Notice cursors shift to right from `all be` text and starts affecting different characters.

### Helix log

No logs, as it does not cause error.

### Platform

Linux - Fedora 37 86_64

### Terminal Emulator

Alacritty 0.10.1

### Helix Version

helix 22.12-223-g9c98043c (9c98043c)

pascalkuthe:
Fixes #5910

Whenever a document is changed helix maps various positions like the
cursor or diagnostics through the `ChangeSet` applied to the document.

Currently, this mapping handles replacements as follows:

* Move position to the left for `Assoc::Before` (start of selection)
* Move position to the right for `Assoc::After` (end of selection)

However, when text is exactly replaced this can produce weird results
where the cursor is moved when it shouldn't. For example if `foo` is
selected and a separate cursor is placed on each character (`s.<ret>`)
and the text is replaced (for example `rx`) then the cursors are moved
to the side instead of remaining in place.

This change adds a special case to the mapping code of replacements:
If the deleted and inserted text have the same (char) length then
the position is returned as if the replacement doesn't exist.

### DIFF
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -29,6 +29,12 @@ pub enum Assoc {
     /// Acts like `Before` if a word character is inserted
     /// before the position, otherwise acts like `After`
     BeforeWord,
+    /// Acts like `Before` but if the position is within an exact replacement
+    /// (exact size) the offset to the start of the replacement is kept
+    BeforeSticky,
+    /// Acts like `After` but if the position is within an exact replacement
+    /// (exact size) the offset to the start of the replacement is kept
+    AfterSticky,
 }
 
 impl Assoc {
@@ -40,12 +46,16 @@ impl Assoc {
     fn insert_offset(self, s: &str) -> usize {
         let chars = s.chars().count();
         match self {
-            Assoc::After => chars,
+            Assoc::After | Assoc::AfterSticky => chars,
             Assoc::AfterWord => s.chars().take_while(|&c| char_is_word(c)).count(),
             // return position before inserted text
-            Assoc::Before => 0,
+            Assoc::Before | Assoc::BeforeSticky => 0,
             Assoc::BeforeWord => chars - s.chars().rev().take_while(|&c| char_is_word(c)).count(),
         }
+    }
+
+    pub fn sticky(self) -> bool {
+        matches!(self, Assoc::BeforeSticky | Assoc::AfterSticky)
     }
 }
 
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I was not 100% sure yet if this is the right behavior for diagnostics. Exact replacements are pretty rare to begin with and basically always intentional by the user.
However replacing one variable of equal length with another variable of equal length should instantly remove the associated diagnostic instead of keeping it. Therefore I changed the bahviour described above to be configurable with a parameter that is enabled for selections but disabled for diagnostics (the only two applications for this mapping)

archseer:
I think this should be a different assoc type, I didn't port it all from CodeMirror: https://github.com/codemirror/state/blob/2117f5b58d040284f1e04ec054a1f2c79bf86a30/src/change.ts#L5-L16

archseer:
Various usages here https://github.com/search?q=org%3Acodemirror+MapMode&type=code

pascalkuthe:
I added the `BeforeSticky` and `AfterSticky` variants to assoc.
I originally went with a separate parameter because this is an orthogonal property that only applies to a single edgecase while in most cases the behavior remained unchanged.  I added some helper functions to make checking against tese variuants slightly cleaner and now it does look nicer than before.

However none of the variants from the source you linked @archseer match the behavior I implemented here so I had to comeup with my own name. I think this usecase is probably pretty specific to a multi-cursor modal editor so other implementations don't need to handle it.

helix-core/src/transaction.rs
```diff
@@ -329,6 +345,9 @@ impl ChangeSet {
     /// position close to the character before, and will place it before insertions over that
     /// range, or at that point. `After` will move it forward, placing it at the end of such
     /// insertions.
+    ///
+    /// The stick variants of these two indicate that `pos` should remain unchaged for exact replacements. This is
+    //// desirable for cursors but undesirable for other things like diagnostics
```
the-mikedavis, line 350:
```suggestion
    /// The "sticky" variants of these two indicate that `pos` should remain unchanged for exact replacements. This is
    /// desirable for cursors but undesirable for other things like diagnostics
```

the-mikedavis (commented):
I like the look of this 👍. I left a minor typo-fix comment and looks like this will need to be rebased on top of the recent optimizations

the-mikedavis (approved):

the-mikedavis (approved):

PR #5931, opened 2023-02-11
Fix mouse configuration not getting updated after config-reload and set-option (#5648)

### LINKED ISSUE #5648: config-reload fails to change mouse configuration state
### Summary

config-reload does not change the mouse configuration state after modifying the config.toml

### Reproduction Steps

Steps to reproduce:

1. Run `helix`
2. run `:config-open`
3. modify the `editor.mouse` from `true` to `false` or `false` to `true` depending on your current configuration.
4. run `:config-reload`

You will notice that the mouse state doesn't actually change.
You need to exit helix and open helix for the configuration to take place.

### Helix log

[helix.log](https://github.com/helix-editor/helix/files/10481756/helix.log)

### Platform

Linux

### Terminal Emulator

konsole

### Helix Version

22.12

johnfrankmorgan:
Fixes #5648.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -394,11 +394,29 @@ impl Application {
         // the configuration.
         self.editor.refresh_config();
 
+        match self.refresh_mouse(self.config.load().editor.mouse) {
+            Ok(_) => {}
+            Err(err) => {
+                self.editor.set_error(err.to_string());
+            }
+        }
+
         // reset view position in case softwrap was enabled/disabled
         let scrolloff = self.editor.config().scrolloff;
         for (view, _) in self.editor.tree.views_mut() {
             let doc = &self.editor.documents[&view.doc];
             view.ensure_cursor_in_view(doc, scrolloff)
+        }
+    }
+
+    /// refresh mouse state after config change
+    fn refresh_mouse(&self, enabled: bool) -> Result<(), Error> {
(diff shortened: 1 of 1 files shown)


gabydd:
Works great! thanks for taking this on(I meant to get to it some point but I have been pretty busy with exams and university applications these past couple weeks)

gabydd (approved):

johnfrankmorgan:
> Works great! thanks for taking this on(I meant to get to it some point but I have been pretty busy with exams and university applications these past couple weeks)

No problem! Thanks for your detailed instructions in the issue, I wouldn't have know where to start without them!

the-mikedavis (commented):
This fixes the behavior for `:config-reload` but it would also be nice  to fix `:set-option` as well. This block could be moved with some small modifications to `handle_config_events` after the `self.editor.refresh_config()` line

johnfrankmorgan:
> This fixes the behavior for `:config-reload` but it would also be nice to fix `:set-option` as well. This block could be moved with some small modifications to `handle_config_events` after the `self.editor.refresh_config()` line

Thanks for the suggestion! :smiley: I wasn't entirely sure how best to handle the `execute!` result so I ended up adding a helper to refresh the mouse state. Let me know if you think the error handling should be moved out of `handle_config_events`.

PR #5932, opened 2023-02-11
feat(ui): deprecated completions

matoous:
Mark deprecated completions using strike-through (CROSSED_OUT modifier). The deprection information is taken either from the `deprecated` field of the completion item or from the completion tags.

The `deprecated` field seems to be the older way of passing the deprecated information and it was already marked as deprecated for `Symbol`. In `CompletionItem` the `deprecated` field is still valid but it seems that the LSP is moving in the general direction of using tags for this kind of information and as such relying on tags as well seems reasonable and future-proof.

Screenshot 2023-02-12 at 00 24 37

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -1,6 +1,10 @@
 use crate::compositor::{Component, Context, Event, EventResult};
-use helix_view::{editor::CompleteAction, ViewId};
-use tui::buffer::Buffer as Surface;
+use helix_view::{
+    editor::CompleteAction,
+    theme::{Modifier, Style},
+    ViewId,
+};
+use tui::{buffer::Buffer as Surface, text::Span};
 
 use std::borrow::Cow;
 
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/completion.rs
```diff
@@ -33,8 +37,20 @@ impl menu::Item for CompletionItem {
     }
 
     fn format(&self, _data: &Self::Data) -> menu::Row {
+        let deprecated = if let Some(tags) = &self.tags {
+            tags.contains(&lsp::CompletionItemTag::DEPRECATED)
+        } else {
+            self.deprecated.unwrap_or_default()
+        };
```
pascalkuthe, line 44: If we support more tags in the future it might be poasible that a server sets the dprecated field, doesn't send the deprecated tag but send other tags.
Lets be a bit more robust and check both: 
```suggestion
        let deprecated = self.deprecated.unwrap_or_default() || self.tags.map_or(false, |tags| tags.contains(&lsp::CompletionItemTag::DEPRECATED));
```

pascalkuthe (changes requested):
Nice qol additon!

A minor nit otherwise LGTM

helix-term/src/ui/completion.rs
matoous: Thanks for suggestion, I had to use
```
        let deprecated = self.deprecated.unwrap_or_default() || self.tags.as_ref().map_or(false, |tags| tags.contains(&lsp::CompletionItemTag::DEPRECATED));
```
(`self.tags.as_ref()` being the difference) instead so not applied directly from the commit suggestion.

pascalkuthe (approved):
LGTM, nice little visual hint

PR #5933, opened 2023-02-12
Prevent a panic when uncommenting a line with only a comment token

trink:
Open a new document `test.rs` and type the following: `di//<esc><C-c>`

The margin calculation pushes the range out of bounds for the comment marker when there are no characters (newline) after it.

thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char range out of bounds: char range 0..3,
Rope/RopeSlice char length 2', ropey-1.6.0/src/rope.rs:546:37

The debug build catches the error in the transaction: thread 'main' panicked at 'attempt to subtract with overflow',
helix-core/src/transaction.rs:503:26

### DIFF
--- a/helix-core/src/comment.rs
+++ b/helix-core/src/comment.rs
@@ -45,7 +45,7 @@ fn find_line_comment(
 
             // determine margin of 0 or 1 for uncommenting; if any comment token is not followed by a space,
             // a margin of 0 is used for all lines.
-            if matches!(line_slice.get_char(pos + token_len), Some(c) if c != ' ') {
+            if !matches!(line_slice.get_char(pos + token_len), Some(c) if c == ' ') {
                 margin = 0;
             }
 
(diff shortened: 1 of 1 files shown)


helix-core/src/comment.rs
```diff
@@ -45,9 +45,11 @@ fn find_line_comment(
...
             // determine margin of 0 or 1 for uncommenting; if any comment token is not followed by a space,
             // a margin of 0 is used for all lines.
-            if matches!(line_slice.get_char(pos + token_len), Some(c) if c != ' ') {
-                margin = 0;
-            }
+            margin = match line_slice.get_char(pos + token_len) {
+                Some(c) if c != ' ' => 0,
+                None => 0,
+                _ => margin,
+            };
```
pascalkuthe, line 52: I think keeping this as an if is slightly cleaner

```suggestion
            if !matches!(line_slice.get_char(pos + token_len), Some(c) if c == ' '){
                margin = 0
            }
```

pascalkuthe (changes requested):
Nice, catch! A minor cosmetic nit, otherwise this LGTM :+:

pascalkuthe (approved):
LGTM, thanks!

helix-core/src/comment.rs
```diff
@@ -45,8 +45,8 @@ fn find_line_comment(
 
             // determine margin of 0 or 1 for uncommenting; if any comment token is not followed by a space,
             // a margin of 0 is used for all lines.
-            if matches!(line_slice.get_char(pos + token_len), Some(c) if c != ' ') {
-                margin = 0;
+            if !matches!(line_slice.get_char(pos + token_len), Some(c) if c == ' ') {
+                margin = 0
```
the-mikedavis, line 49:
```suggestion
                margin = 0;
```

looks like this semicolon got chopped off on accident

the-mikedavis (approved):

PR #5934, opened 2023-02-12
Feat: LSP Type Hints

### LINKED ISSUE #2070: LSP Type Hints
I'm not sure how straightforward this would be to implement.
Type hints could be given on hover, or overlaid on the text like this: https://github.com/rust-lang/vscode-rust/issues/374 

image

poliorcetics:
Add support for LSP type hints (merged in version 3.17 of the spec, a few month old now).

For now, I only added it in the main editor views, pickers do not have type hints.
Even if we decide they need to, I would prefer doing it in another PR since this one is already big
enough as it is.

I'll join a screen in a comment below, and if people could try it out with their LSPs and configs, it
would be very useful, since I pretty much only use Rust-Analyzer so testing may be somewhat limited still.

Closes #2070

EDIT: [instructions to test](https://github.com/helix-editor/helix/pull/5934#issuecomment-1445040945)

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -118,9 +118,11 @@ The following statusline elements can be configured:
 | `enable`              | Enables LSP integration. Setting to false will completely disable language servers regardless of language settings.| `true` |
 | `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
 | `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
+| `display-inlay-hints` | Display inlay hints[^2]                                     | `true`  |
 | `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
 
 [^1]: By default, a progress spinner is shown in the statusline beside the file path.
+[^2]: You may also have to activate them in the LSP config for them to appear, not just in Helix
 
 ### `[editor.cursor-shape]` Section
 
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -118,11 +118,12 @@ The following statusline elements can be configured:
 | `enable`              | Enables LSP integration. Setting to false will completely disable language servers regardless of language settings.| `true` |
 | `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
 | `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
-| `display-inlay-hints` | Display inlay hints[^2]                                     | `true`  |
+| `display-inlay-hints` | Display inlay hints[^2]                                     | `false` |
 | `display-signature-help-docs` | Display docs under signature help popup             | `true`  |
 
 [^1]: By default, a progress spinner is shown in the statusline beside the file path.
-[^2]: You may also have to activate them in the LSP config for them to appear, not just in Helix
+[^2]: You may also have to activate them in the LSP config for them to appear, not just in Helix.
+      Inlay hints in Helix are still being improved on and may be a little bit laggy/janky under some circumstances, please report any bugs you see so we can fix them!
 
 ### `[editor.cursor-shape]` Section
 
--- a/helix-core/src/text_annotations.rs
+++ b/helix-core/src/text_annotations.rs
@@ -15,6 +15,15 @@ pub struct InlineAnnotation {
     pub char_idx: usize,
 }
 
+impl InlineAnnotation {
+    pub fn new(char_idx: usize, text: impl Into<Tendril>) -> Self {
+        Self {
+            char_idx,
+            text: text.into(),
+        }
+    }
+}
+
 /// Represents a **single Grapheme** that is part of the document
 /// that start at `char_idx` that will be replaced with
 /// a different `grapheme`.
@@ -33,22 +42,13 @@ pub struct InlineAnnotation {
 /// use helix_core::text_annotations::Overlay;
 ///
 /// // replaces a
-/// Overlay {
-///   char_idx: 0,
-///   grapheme: "X".into(),
-/// };
+/// Overlay::new(0, "X");
 ///
 /// // replaces X͎̊͢͜͝͡
-/// Overlay{
-///   char_idx: 1,
-///   grapheme: "\t".into(),
-/// };
+/// Overlay::new(1, "\t");
 ///
 /// // replaces b
-/// Overlay{
-///   char_idx: 6,
-///   grapheme: "X̢̢̟͖̲͌̋̇͑͝".into(),
-/// };
+/// Overlay::new(6, "X̢̢̟͖̲͌̋̇͑͝");
 /// ```
 ///
 /// The following examples are invalid uses
@@ -57,21 +57,24 @@ pub struct InlineAnnotation {
 /// use helix_core::text_annotations::Overlay;
 ///
 /// // overlay is not aligned at grapheme boundary
-/// Overlay{
-///   char_idx: 3,
-///   grapheme: "x".into(),
-/// };
+/// Overlay::new(3, "x");
 ///
 /// // overlay contains multiple graphemes
-/// Overlay{
-///   char_idx: 0,
-///   grapheme: "xy".into(),
-/// };
+/// Overlay::new(0, "xy");
 /// ```
 #[derive(Debug, Clone)]
 pub struct Overlay {
     pub char_idx: usize,
     pub grapheme: Tendril,
+}
+
+impl Overlay {
+    pub fn new(char_idx: usize, grapheme: impl Into<Tendril>) -> Self {
+        Self {
+            char_idx,
+            grapheme: grapheme.into(),
+        }
+    }
 }
 
 /// Line annotations allow for virtual text between normal
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1404,7 +1404,7 @@ pub fn compute_inlay_hints_for_all_views(editor: &mut Editor, jobs: &mut crate::
     }
 
     for (view, _) in editor.tree.views() {
-        let doc = match editor.documents.get_mut(&view.doc) {
+        let doc = match editor.documents.get(&view.doc) {
             Some(doc) => doc,
             None => continue,
         };
@@ -1416,11 +1416,10 @@ pub fn compute_inlay_hints_for_all_views(editor: &mut Editor, jobs: &mut crate::
 
 fn compute_inlay_hints_for_view(
     view: &View,
-    doc: &mut Document,
+    doc: &Document,
 ) -> Option<std::pin::Pin<Box<impl Future<Output = Result<crate::job::Callback, anyhow::Error>>>>> {
     let view_id = view.id;
     let doc_id = view.doc;
-    let revision = doc.get_current_revision();
 
     let language_server = doc.language_server()?;
 
@@ -1446,14 +1445,14 @@ fn compute_inlay_hints_for_view(
                 .min(len_lines);
 
             let new_doc_inlay_hint_id = DocumentInlayHintsId {
-                revision,
                 first_line,
                 last_line,
             };
             // Don't recompute the annotations in case nothing has changed about the view
-            if doc
-                .inlay_hints(view_id)
-                .map_or(false, |dih| dih.id == new_doc_inlay_hint_id)
+            if !doc.inlay_hints_oudated
+                && doc
+                    .inlay_hints(view_id)
+                    .map_or(false, |dih| dih.id == new_doc_inlay_hint_id)
             {
                 return None;
             }
@@ -1490,14 +1489,17 @@ fn compute_inlay_hints_for_view(
                 None => return,
             };
 
-            let mut hints = match response {
-                Some(h) if !h.is_empty() => h,
-                _ => return,
-            };
-
-            let offset_encoding = match doc.language_server() {
-                Some(ls) => ls.offset_encoding(),
-                None => return,
+            // If we have neither hints nor an LSP, empty the inlay hints since they're now oudated
+            let (mut hints, offset_encoding) = match (response, doc.language_server()) {
+                (Some(h), Some(ls)) if !h.is_empty() => (h, ls.offset_encoding()),
+                _ => {
+                    doc.set_inlay_hints(
+                        view_id,
+                        DocumentInlayHints::empty_with_id(new_doc_inlay_hints_id),
+                    );
+                    doc.inlay_hints_oudated = false;
+                    return;
+                }
             };
 
             // Most language servers will already send them sorted but ensure this is the case to
@@ -1560,6 +1562,7 @@ fn compute_inlay_hints_for_view(
                     padding_after_inlay_hints: padding_after_inlay_hints.into(),
                 },
             );
+            doc.inlay_hints_oudated = false;
         },
     );
 
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -15,8 +15,13 @@ use tui::{
 
 use super::{align_view, push_jump, Align, Context, Editor, Open};
 
-use helix_core::{path, Selection};
-use helix_view::{document::Mode, editor::Action, theme::Style};
+use helix_core::{path, text_annotations::InlineAnnotation, Selection};
+use helix_view::{
+    document::{DocumentInlayHints, DocumentInlayHintsId, Mode},
+    editor::Action,
+    theme::Style,
+    Document, View,
+};
 
 use crate::{
     compositor::{self, Compositor},
@@ -27,7 +32,8 @@ use crate::{
 };
 
 use std::{
-    borrow::Cow, cmp::Ordering, collections::BTreeMap, fmt::Write, path::PathBuf, sync::Arc,
+    borrow::Cow, cmp::Ordering, collections::BTreeMap, fmt::Write, future::Future, path::PathBuf,
+    sync::Arc,
 };
 
 /// Gets the language server that is attached to a document, and
@@ -1390,4 +1396,171 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
             doc.set_selection(view.id, selection);
         },
     );
+}
+
+pub fn compute_inlay_hints_for_all_views(editor: &mut Editor, jobs: &mut crate::job::Jobs) {
+    if !editor.config().lsp.display_inlay_hints {
+        for doc in editor.documents.values_mut() {
+            doc.reset_all_inlay_hints();
+        }
+        return;
+    }
+
+    for (view, _) in editor.tree.views() {
+        let doc = match editor.documents.get_mut(&view.doc) {
+            Some(doc) => doc,
+            None => continue,
+        };
+        if let Some(callback) = compute_inlay_hints_for_view(view, doc) {
+            jobs.callback(callback);
+        }
+    }
+}
+
+fn compute_inlay_hints_for_view(
+    view: &View,
+    doc: &mut Document,
+) -> Option<std::pin::Pin<Box<impl Future<Output = Result<crate::job::Callback, anyhow::Error>>>>> {
+    let view_id = view.id;
+    let doc_id = view.doc;
+    let revision = doc.get_current_revision();
+
+    let language_server = doc.language_server()?;
+
+    let capabilities = language_server.capabilities();
+
+    let (future, new_doc_inlay_hints_id) = match capabilities.inlay_hint_provider {
+        Some(
+            lsp::OneOf::Left(true)
+            | lsp::OneOf::Right(lsp::InlayHintServerCapabilities::Options(_)),
+        ) => {
+            let doc_text = doc.text();
+            let len_lines = doc_text.len_lines();
+
+            // Compute ~3 times the current view height of inlay hints, that way some scrolling
+            // will not show half the view with hints and half without while still being faster
+            // than computing all the hints for the full file (which could be dozens of time
+            // longer than the view is).
+            let view_height = view.inner_height();
+            let first_visible_line = doc_text.char_to_line(view.offset.anchor);
+            let first_line = first_visible_line.saturating_sub(view_height);
+            let last_line = first_visible_line
+                .saturating_add(view_height.saturating_mul(2))
+                .min(len_lines);
+
+            let new_doc_inlay_hint_id = DocumentInlayHintsId {
+                revision,
+                first_line,
+                last_line,
+            };
+            // Don't recompute the annotations in case nothing has changed about the view
+            if doc
+                .get_inlay_hints(view_id)
+                .map_or(false, |dih| dih.id == new_doc_inlay_hint_id)
+            {
+                return None;
+            }
+
+            let doc_slice = doc_text.slice(..);
+            let first_char_in_range = doc_slice.line_to_char(first_line);
+            let last_char_in_range = doc_slice.line_to_char(last_line);
+
+            let range = helix_lsp::util::range_to_lsp_range(
+                doc_text,
+                helix_core::Range::new(first_char_in_range, last_char_in_range),
+                language_server.offset_encoding(),
+            );
+
+            (
+                language_server.text_document_range_inlay_hints(doc.identifier(), range, None),
+                new_doc_inlay_hint_id,
+            )
+        }
+        _ => return None,
+    };
+
+    let callback = super::make_job_callback(
+        future?,
+        move |editor, _compositor, response: Option<Vec<lsp::InlayHint>>| {
+            // The config was modified or the window was closed while the request was in flight
+            if !editor.config().lsp.display_inlay_hints || editor.tree.try_get(view_id).is_none() {
+                return;
+            }
+
+            // Add annotations to relevant document, not the current one (it may have changed in between)
+            let doc = match editor.documents.get_mut(&doc_id) {
+                Some(doc) => doc,
+                None => return,
+            };
+
+            let mut hints = match response {
+                Some(h) if !h.is_empty() => h,
+                _ => return,
+            };
+
+            let offset_encoding = match doc.language_server() {
+                Some(ls) => ls.offset_encoding(),
+                None => return,
+            };
+
+            // Most language servers will already send them sorted but ensure this is the case to
+            // avoid errors on our end.
+            hints.sort_unstable_by_key(|inlay_hint| inlay_hint.position);
+
+            let mut type_inlay_hints = Vec::new();
+            let mut parameter_inlay_hints = Vec::new();
+            let mut other_inlay_hints = Vec::new();
+
+            let doc_text = doc.text();
+
+            for hint in hints {
+                let char_idx =
+                    match helix_lsp::util::lsp_pos_to_pos(doc_text, hint.position, offset_encoding)
+                    {
+                        Some(pos) => pos,
+                        // Skip inlay hints that have no "real" position
+                        None => continue,
+                    };
+
+                let label = match hint.label {
+                    lsp::InlayHintLabel::String(s) => s,
+                    lsp::InlayHintLabel::LabelParts(parts) => parts
+                        .into_iter()
+                        .map(|p| p.value)
+                        .collect::<Vec<_>>()
+                        .join(""),
+                };
+
+                let inlay_hints_vec = match hint.kind {
+                    Some(lsp::InlayHintKind::TYPE) => &mut type_inlay_hints,
+                    Some(lsp::InlayHintKind::PARAMETER) => &mut parameter_inlay_hints,
+                    // We can't warn on unknown kind here since LSPs are free to set it or not, for
+                    // example Rust Analyzer does not: every kind will be `None`.
+                    _ => &mut other_inlay_hints,
+                };
+
+                if let Some(true) = hint.padding_left {
+                    inlay_hints_vec.push(InlineAnnotation::new(char_idx, " "));
+                }
+
+                inlay_hints_vec.push(InlineAnnotation::new(char_idx, label));
+
+                if let Some(true) = hint.padding_right {
+                    inlay_hints_vec.push(InlineAnnotation::new(char_idx, " "));
+                }
+            }
+
+            doc.set_inlay_hints(
+                view_id,
+                DocumentInlayHints {
+                    id: new_doc_inlay_hints_id,
+                    type_inlay_hints: type_inlay_hints.into(),
+                    parameter_inlay_hints: parameter_inlay_hints.into(),
+                    other_inlay_hints: other_inlay_hints.into(),
+                },
+            );
+        },
+    );
+
+    Some(callback)
 }
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1504,9 +1504,11 @@ fn compute_inlay_hints_for_view(
             // avoid errors on our end.
             hints.sort_unstable_by_key(|inlay_hint| inlay_hint.position);
 
+            let mut padding_before_inlay_hints = Vec::new();
             let mut type_inlay_hints = Vec::new();
             let mut parameter_inlay_hints = Vec::new();
             let mut other_inlay_hints = Vec::new();
+            let mut padding_after_inlay_hints = Vec::new();
 
             let doc_text = doc.text();
 
@@ -1537,13 +1539,13 @@ fn compute_inlay_hints_for_view(
                 };
 
                 if let Some(true) = hint.padding_left {
-                    inlay_hints_vec.push(InlineAnnotation::new(char_idx, " "));
+                    padding_before_inlay_hints.push(InlineAnnotation::new(char_idx, " "));
                 }
 
                 inlay_hints_vec.push(InlineAnnotation::new(char_idx, label));
 
                 if let Some(true) = hint.padding_right {
-                    inlay_hints_vec.push(InlineAnnotation::new(char_idx, " "));
+                    padding_after_inlay_hints.push(InlineAnnotation::new(char_idx, " "));
                 }
             }
 
@@ -1554,6 +1556,8 @@ fn compute_inlay_hints_for_view(
                     type_inlay_hints: type_inlay_hints.into(),
                     parameter_inlay_hints: parameter_inlay_hints.into(),
                     other_inlay_hints: other_inlay_hints.into(),
+                    padding_before_inlay_hints: padding_before_inlay_hints.into(),
+                    padding_after_inlay_hints: padding_after_inlay_hints.into(),
                 },
             );
         },
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1400,9 +1400,6 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
 
 pub fn compute_inlay_hints_for_all_views(editor: &mut Editor, jobs: &mut crate::job::Jobs) {
     if !editor.config().lsp.display_inlay_hints {
-        for doc in editor.documents.values_mut() {
-            doc.reset_all_inlay_hints();
-        }
         return;
     }
 
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1455,7 +1455,7 @@ fn compute_inlay_hints_for_view(
             };
             // Don't recompute the annotations in case nothing has changed about the view
             if doc
-                .get_inlay_hints(view_id)
+                .inlay_hints(view_id)
                 .map_or(false, |dih| dih.id == new_doc_inlay_hint_id)
             {
                 return None;
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1002,6 +1002,10 @@ impl EditorView {
         event: &MouseEvent,
         cxt: &mut commands::Context,
     ) -> EventResult {
+        if event.kind != MouseEventKind::Moved {
+            cxt.editor.reset_idle_timer();
+        }
+
         let config = cxt.editor.config();
         let MouseEvent {
             kind,
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -978,6 +978,8 @@ impl EditorView {
     }
 
     pub fn handle_idle_timeout(&mut self, cx: &mut commands::Context) -> EventResult {
+        commands::compute_inlay_hints_for_all_views(cx.editor, cx.jobs);
+
         if let Some(completion) = &mut self.completion {
             return if completion.ensure_item_resolved(cx) {
                 EventResult::Consumed(None)
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -881,6 +881,8 @@ impl Document {
 
     /// Apply a [`Transaction`] to the [`Document`] to change its text.
     fn apply_impl(&mut self, transaction: &Transaction, view_id: ViewId) -> bool {
+        use helix_core::Assoc;
+
         let old_doc = self.text().clone();
 
         let success = transaction.changes().apply(&mut self.text);
@@ -929,10 +931,10 @@ impl Document {
                     .unwrap();
             }
 
+            let changes = transaction.changes();
+
             // map state.diagnostics over changes::map_pos too
             for diagnostic in &mut self.diagnostics {
-                use helix_core::Assoc;
-                let changes = transaction.changes();
                 diagnostic.range.start = changes.map_pos(diagnostic.range.start, Assoc::After);
                 diagnostic.range.end = changes.map_pos(diagnostic.range.end, Assoc::After);
                 diagnostic.line = self.text.char_to_line(diagnostic.range.start);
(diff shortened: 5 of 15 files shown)


poliorcetics:
Screenshot 2023-02-12 at 04 12 20

Proful:
Amazing!

image

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+
+            // Compute ~3 times the current view height of inlay hints, that way some scrolling
+            // will not show half the view with hints and half without while still being faster
+            // than computing all the hints for the full file (which could be dozens of time
+            // longer than the view is).
+            let view_height = view.inner_height();
+            let first_visible_line = view.estimate_last_doc_line(doc).saturating_sub(view_height);
+            let first_line = first_visible_line
+                .saturating_sub(view_height)
+                .min(len_lines);
```
pascalkuthe, line 1385: the `min` is not necessary since you are subtracting

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+            let last_line = first_visible_line
+                .saturating_add(view_height)
```
pascalkuthe, line 1387: you just calculated `last_line - height + height` here. What you want is `last_line + height`.
Also you likely want to calculate `first_line` directly to make this more readable so this should really be `first_line.saturating_add(2*height)`.

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+        ) => {
+            let doc_text = &doc.text();
+            let len_lines = doc_text.len_lines();
+
+            // Compute ~3 times the current view height of inlay hints, that way some scrolling
+            // will not show half the view with hints and half without while still being faster
+            // than computing all the hints for the full file (which could be dozens of time
+            // longer than the view is).
+            let view_height = view.inner_height();
+            let first_visible_line = view.estimate_last_doc_line(doc).saturating_sub(view_height);
```
pascalkuthe, line 1382: This is a bit backwards as `estimate_last_doc_line` interally calcualtes the first line and then adds `view_height`.
Instead just use:
```suggestion
            let first_visible_line = doc.text().char_to_line(view.offfset.anchor);
```

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+            let last_char_in_range = doc_slice.line_to_char(last_line);
+
+            let range = helix_lsp::util::range_to_lsp_range(
+                doc_text,
+                helix_core::Range::new(first_char_in_range, last_char_in_range),
+                language_server.offset_encoding(),
+            );
+
+            (
+                language_server.text_document_range_inlay_hints(doc.identifier(), range, None),
```
pascalkuthe, line 1412: I think it would be a bit cleaner if we passed a `helix_core::Range` to `text_document_range_inlay_hints` and do the lsp conversion inside `text_document_range_inlay_hints` in `helix-lsp`

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+                None => return,
+            };
+            let view = match editor.tree.try_get_mut(view_id) {
+                Some(view) => view,
+                None => {
+                    // If the view doesn't exist anymore, purge its data from the text annotations
+                    // in the document
+                    doc.reset_text_annotations(view_id);
+                    return;
+                }
```
pascalkuthe, line 1439: this should be handled in 

`Document::remove_view`.

`Document::ensure_view_init` might also be relevant to this PR

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1331,191 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+                    annotations.add_inline_annotations(
+                        Rc::new([InlineAnnotation::new(char_idx, " ")]),
+                        None,
+                    );
+                }
+
+                annotations.add_inline_annotations(
+                    Rc::new([InlineAnnotation::new(char_idx, label)]),
+                    style.map(helix_core::syntax::Highlight),
+                );
```
pascalkuthe, line 1506: layers carray a certain amount of overhead. Instead of creating a seperate layer for the each annotation create a `Vec<InlineAnnotation>` and then convert that to `Rc<[InlineAnnotation]>` (same for everwhere else you create a layer, there should be only one layer per type of virtual text, one for padding, one for each kind of label (paramters,  types and default style))

pascalkuthe (changes requested):
Nice to finally see the virtual text API put to use :tada: 

Sadly I think you missunderstood the API somewhat and therefore had to do a bunch of unnecesassary changes. I designed the virtual text API very carefully (including with this usecase in mind) and the way you changed the API here makes it less efficent and less elegant.

The API is a bit abstract without an example and you are not the first person to be initially confused.  I think the core of your misunderstanding is that you created a single layer (call to `add_inline_annotations`) per inlay a hint

That are a lot of layers (and not efficent since every layers must be checked at every singel char when transvcersing the document) and I therefore you probably thaught that you needed to store these somewhere.

Instead the idea is that one layer corresponds to a single TYPE of virtual text. 
Therefore it is possible to construct these layers on the fly in the `text_annotations` functions and only store the contents of the layers. In your case there should only be the need for 4 layers. Constructing a `Vec` of 4 elements on the fly is basically free and is how the API was supposed to be used.

I left a bunch of comments with more details.

helix-view/src/document.rs
```diff
@@ -107,6 +107,7 @@ pub struct Document {
     pub(crate) id: DocumentId,
     text: Rope,
     selections: HashMap<ViewId, Selection>,
+    text_annotations: HashMap<ViewId, TextAnnotations>,
```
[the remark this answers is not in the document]
poliorcetics: Done

helix-core/src/text_annotations.rs
```diff
@@ -154,12 +157,42 @@ pub struct TextAnnotations {
...
+            for annot in layer.annotations.iter() {
+                apply_change(&annot.char_idx);
+            }
+        }
+        for layer in line_annotations {
+            for annot in layer.annotations.iter() {
+                apply_change(&annot.anchor_char_idx);
+            }
+        }
+    }
```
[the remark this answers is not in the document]
poliorcetics: Done

helix-view/src/view.rs
```diff
@@ -125,6 +125,22 @@ pub struct View {
...
+pub struct LastInlayHintRequestId {
+    /// Document identifier, since annotations for one document are not valid for another
+    pub doc_id: DocumentId,
+    /// Revision of the document at computation time
+    pub doc_revision: usize,
+    /// First line of the range for which the annotations were computed
+    pub first_line: usize,
+    /// Last line of the range for which the annotations were computed
+    pub last_line: usize,
 }
```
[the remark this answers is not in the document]
poliorcetics: Done

helix-view/src/document.rs
```diff
@@ -1254,10 +1286,6 @@ impl Document {
                 .map(Highlight),
         }
     }
-
-    pub fn text_annotations(&self, _theme: Option<&Theme>) -> TextAnnotations {
```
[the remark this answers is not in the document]
poliorcetics: Done, I still need the `view_id` though, is that a problem ?

helix-view/src/document.rs
```diff
@@ -1154,10 +1177,19 @@ impl Document {
         &self.selections[&view_id]
     }
 
+    #[inline]
```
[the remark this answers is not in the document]
poliorcetics: Made in another commit instead, it's just a small drive-by fix

helix-view/src/view.rs
```diff
@@ -402,9 +419,9 @@ impl View {
         Some(pos)
     }
 
-    pub fn text_annotations(&self, doc: &Document, theme: Option<&Theme>) -> TextAnnotations {
+    pub fn text_annotations<'d>(&self, doc: &'d Document) -> std::borrow::Cow<'d, TextAnnotations> {
         // TODO custom annotations for custom views like side by side diffs
-        doc.text_annotations(theme)
+        doc.text_annotations(self.id)
     }
```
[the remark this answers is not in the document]
poliorcetics: Fixed

helix-term/src/commands/lsp.rs
poliorcetics: Fixed

helix-term/src/commands/lsp.rs
poliorcetics: Fixed

helix-term/src/commands/lsp.rs
poliorcetics: Fixed

helix-term/src/commands/lsp.rs
poliorcetics: Eh, I don't like that: for one, it's not the style used by the other functions in this file, and it conflates `lsp::Range` and `helix_core::Range` in place where only LSP-related stuff happens, I don't think that's good for clarity compared to the single conversion call later in the crate chain

helix-term/src/commands/lsp.rs
poliorcetics: Done

helix-term/src/commands/lsp.rs
poliorcetics: Done

poliorcetics:
> Sadly I think you missunderstood the API somewhat and therefore had to do a bunch of unnecesassary changes. I designed the virtual text API very carefully (including with this usecase in mind) and the way you changed the API here makes it less efficent and less elegant.
> 
> The API is a bit abstract without an example and you are not the first person to be initially confused. I think the core of your misunderstanding is that you created a single layer (call to add_inline_annotations) per inlay a hint
>
> That are a lot of layers (and not efficent since every layers must be checked at every singel char when transvcersing the document) and I therefore you probably thaught that you needed to store these somewhere.

Yes, this is really confusing at first because there is pretty much zero doc in Helix about virtual text and annotations (or, if there is, it's absolutely not where I expect it to be, on the types and methods) so I was really confused, thanks for clarifying !

> Instead the idea is that one layer corresponds to a single TYPE of virtual text.
> Therefore it is possible to construct these layers on the fly in the text_annotations functions and only store the contents of the layers. In your case there should only be the need for 4 layers. Constructing a Vec of 4 elements on the fly is basically free and is how the API was supposed to be used.

Yes, that approach is much better, I used it instead. I can't separate padding from the hints though, else it breaks in all kinds of ways: everything is squished together and unreadable.

> Once you apply my suggestion the diff should reduce significantly and this should turn into a smaller PR.

Not really 😅 I added quite some docs for inlay hints in documents that re-inflated the count after reducing it, but still the MR is better than it was before thanks to you

helix-view/src/document.rs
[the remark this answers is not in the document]
pascalkuthe: You can move this entire function into `helix_view::View::text_annotations`.
The intentation was:
* Virtual text that is shared among all views of a buffer is added in `Document::text_annotations`
* Virtual text that is specific to a single view is added in `View::text_annotations`

You have full access to the document in `View::text_annotations` so everything you can do here can be done there too (in fact that function just calls this function).

helix-view/src/view.rs
```diff
@@ -402,9 +402,10 @@ impl View {
         Some(pos)
     }
 
+    /// Get the text annotations to display in the current view for the given document and theme.
     pub fn text_annotations(&self, doc: &Document, theme: Option<&Theme>) -> TextAnnotations {
         // TODO custom annotations for custom views like side by side diffs
-        doc.text_annotations(theme)
+        doc.text_annotations(self.id, theme)
```
pascalkuthe, line 408: just move all your code here and replace `self` with `doc`

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1330,170 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+                    Some(lsp::InlayHintKind::TYPE) => &mut type_inlay_hints,
+                    Some(lsp::InlayHintKind::PARAMETER) => &mut parameter_inlay_hints,
+                    // We can't warn on unknown kind here since LSPs are free to set it or not, for
+                    // example Rust Analyzer does not: every kind will be `None`.
+                    _ => &mut other_inlay_hints,
+                };
+
+                if let Some(true) = hint.padding_left {
+                    inlay_hints_vec.push(InlineAnnotation::new(char_idx, " "));
+                }
```
pascalkuthe, line 1477: From the LSP spec:

> Note: Padding should use the editor's background color, not the
> background color of the hint itself. That means padding can be used
> to visually align/separate an inlay hint.

That means that you need one extra layer for the the padding which uses `ui.virtual.whitespace` (the theme key normally used for spaces)

pascalkuthe:
> Yes, that approach is much better, I used it instead. I can't separate padding from the hints though, else it breaks in all kinds of ways: everything is squished together and unreadable.
> 

Huh that is odd, it shouldn't let me check.

> Not really sweat_smile I added quite some docs for inlay hints in documents that re-inflated the count after reducing it, but still the MR is better than it was before thanks to you


Well you added quite a bit of docs and its still almost 25% smaller so I would call that a win :D

pascalkuthe:
> > Yes, that approach is much better, I used it instead. I can't separate padding from the hints though, else it breaks in all kinds of ways: everything is squished together and unreadable.
> 
> Huh that is odd, it shouldn't let me check.
> 
> > Not really sweat_smile I added quite some docs for inlay hints in documents that re-inflated the count after reducing it, but still the MR is better than it was before thanks to you
> 
> Well you added quite a bit of docs and its still almost 25% smaller so I would call that a win :D

I just implemented this myself and it worked fine for me.
However for this to work the order of layer is important, since all these annotations will occur at the same position:

* first add the left padding layer
* the the actual label layers
* then the right padding layer

helix-view/src/document.rs
```diff
@@ -150,13 +156,63 @@ pub struct Document {
...
+/// Associated with a [`Document`] and [`ViewId`], uniquely identifies the state of inlay hints for
+/// for that document and view: if this changed since the last save, the inlay hints for the view
+/// should be recomputed.
+///
+/// We can't store the `ViewOffset` instead of the first and last asked-for lines because if
+/// softwrapping changes, the `ViewOffset` may not change while the displayed lines will.
+#[derive(Copy, Clone, PartialEq, Eq)]
+pub struct DocumentInlayHintsId {
+    /// Document revision at which the inlay hints were requested.
+    pub revision: usize,
```
pascalkuthe, line 201: I tested out this PR some more and noticed that using the revision for refreshing doesn't work as well as I had hoped (it cases issues with undo for example). It seemed like a good solution at first but I cameup with something better/easier: Since all inline annotations for all views need to be requested on idle timeout when the document is modified simply store a bool on the document:
``` rust
inlayhint_outdated: bool
```

* In `apply_impl` simply set this to ` true`.
* In `compute_inlay_hints_for_view`: reset this field to `false` and rerequest the inlay hints if this is true OR the view position changed

helix-view/src/document.rs
```diff
@@ -859,24 +929,46 @@ impl Document {
...
                 diagnostic.line = self.text.char_to_line(diagnostic.range.start);
             }
             self.diagnostics
                 .sort_unstable_by_key(|diagnostic| diagnostic.range);
 
+            // Update the inlay hint annotations' positions, helping ensure they are displayed in the proper place
+            let apply_inlay_hint_changes = |annotations: &mut Rc<[InlineAnnotation]>| {
+                if let Some(data) = Rc::get_mut(annotations) {
+                    for inline in data.iter_mut() {
+                        inline.char_idx = changes.map_pos(inline.char_idx, Assoc::After);
```
pascalkuthe, line 947: Neither `Assoc::After` or `Assoc::Before` works well here that is because both of these were designed for **ranges** where the start is mapped with `Assoc::Before` and `Assoc::After`. However for inlay hints we want something different. Inlay hints are bound to a single point so when that point is removed when also want to delete the inlay hint. For that you need to add a new variant `Assoc::Point` which returns `None` in case the position which contains `char_idx` was removed/replaced.

For that to worth tough we need a way to remove `InlineAnnotations` so the equivalent for `Vec::retain` but for `Rc<[T]>` which is not possible without a custom reference counted type (which we probably don't want to add here).

I am considering just switching `TextAnnotations` back to using slices instead of `Rc`  which it originally did. I changed it to make the API easier to use/remove lifetime issues that occured in #5340 but I think those issues were actually resolved by something else so maybe a slice is a better choice afterall.

I don't have more time to look into this right now but you could start with implementing `Assoc::Delete` and I will try to comeup with something in the next couple of days.

helix-term/src/ui/editor.rs
```diff
@@ -978,6 +978,8 @@ impl EditorView {
     }
 
     pub fn handle_idle_timeout(&mut self, cx: &mut commands::Context) -> EventResult {
+        commands::compute_inlay_hints_for_all_views(cx.editor, cx.jobs);
```
pascalkuthe, line 981: I am not 100% sure why but the inlay hints don't seem to update for me while scrolling with the mouse.
It seems that we do not reset the idle timeout on mouse actions. I think that might actually be an existing bug that would be fixed by adding:
``` rust
cx.editor.reset_idle_timer();
```
at the top of `handle_mouse_event`

helix-view/src/document.rs
```diff
@@ -1255,8 +1348,58 @@ impl Document {
...
+        text_annotations
+    }
+
+    /// Set the inlay hints for this document and `view_id`.
+    pub fn set_inlay_hints(&mut self, view_id: ViewId, inlay_hints: DocumentInlayHints) {
+        self.inlay_hints.insert(view_id, inlay_hints);
+    }
+
+    /// Get the inlay hints for this document and `view_id`.
+    pub fn get_inlay_hints(&self, view_id: ViewId) -> Option<&DocumentInlayHints> {
```
archseer, line 1395: Nit: I usually use `inlay_hints()` (no `get_`) on the getters

helix-view/src/document.rs
```diff
@@ -150,13 +156,63 @@ pub struct Document {
...
+/// necessary to ensure it's placed correctly instead of after all the time (or ignored because it
+/// came in a later layer and at the same `char_idx`).
+#[derive(Debug, Clone)]
+pub struct DocumentInlayHints {
+    /// Identifier for the inlay hints stored in this structure. To be checked to know if they have
+    /// to be recomputed on idle or not.
+    pub id: DocumentInlayHintsId,
+
+    /// Inlay hints of `TYPE` kind, if any.
+    pub type_inlay_hints: Rc<[InlineAnnotation]>,
```
archseer, line 179: We should probably be using `Arc` everywhere. I've avoided `Rc`

helix-term/src/commands/lsp.rs
```diff
@@ -1324,3 +1330,170 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
     );
 }
+
+pub fn compute_inlay_hints_for_all_views(editor: &mut Editor, jobs: &mut crate::job::Jobs) {
+    if !editor.config().lsp.display_inlay_hints {
+        for doc in editor.documents.values_mut() {
+            doc.reset_all_inlay_hints();
+        }
+        return;
+    }
```
archseer, line 1340: Could we avoid doing this here and instead do the reset once when editor config is updated?

helix-view/src/document.rs
pascalkuthe: In this case the only reasons this is a `Rc` instead of a `Vec` is that the virtual text API currently expects that instead of a slice. There are some other limitations imposed by using an Rc so I might just switch backto a slice, see https://github.com/helix-editor/helix/pull/5934#discussion_r1103890555.

I think there were some lifetime confusions initially in #5340 but I think those were resolved otherwise and I just kept the Rc to make the API easier to use.

helix-term/src/commands/lsp.rs
poliorcetics: Yes, it's better, done

helix-term/src/commands/lsp.rs
poliorcetics: Fixed, though I introduced `ui.virtual.inlay-hint.padding` instead, to help people have a different style for them than regular whitespaces, else it quickly becomes confusing because there appears to be an unreachable whitespace that is not an inlay hint.

poliorcetics (commented):
Fixed some of the comments, I'm gonna work on the bigger ones now

helix-view/src/document.rs
poliorcetics: Resolving in favor of https://github.com/helix-editor/helix/pull/5934#discussion_r1103890555, easier for me to track one comment than two

helix-view/src/document.rs
poliorcetics: I'm not sure I want to do this in this PR. Nor that I am qualified to do it.

`ChangeSet::map_pos()` currently unconditionnally returns a `usize`, adding `Assoc::Delete` or `Assoc::Point` would need to change the return type to account for the deletion cleanly.

Instead I think this PR should either be blocked until `map_pos` and `Assoc::...` are better equipped or, if we don't want to block too long, make an issue for it and fix it later on since for now `Assoc::After` works well enough (it probably has strange edge cases at the end of files but apart from that it should be okay for a v1)

pascalkuthe: If you don't feel confident working on this that's fine. I will create a PR myself in the next couple days that that you can then depend on. It's not that big of a change so I dlunt it will block this PR for long.

I think it is a requirement for this because without properly handeling this some odd effects do occur. For example if you delete two lines `2xd` then all inlay hints within those lines are moved to the start of the next line

lazytanuki:
Hi and thanks for working on this!
Been testing for a couple hours and it's really nice.
I find that having the hints pop up in the function arguments while typing them can be a bit unnerving, as it shifts the cursor back and forth.
I think it would be nice to have a setting to disable inlay hints for the current function parameters, in insert mode a least. Or maybe in insert mode altogether ? What do others think ?

poliorcetics:
Those should probably be options. I agree that some of that sound nice, though I think implementing them in this MR would delay it too much: I would prefer having something working merged then refine it as user reports pop up :), that way we can both keep this MR relatively small (my time is sadly limited, the maintainers' time too) and account for lots of reports at once instead of just the people following this MR

Does that sound good to you ?

univerz:
>  I find that having the hints pop up in the function arguments while typing them can be a bit unnerving

+1. I didn't notice this problem while using vscode, so there is a way to do it right.

marcbull:
Thanks for the work on this part of Helix, this is one of the features I really want to use.

> I find that having the hints pop up in the function arguments while typing them can be a bit unnerving, as it shifts the cursor back and forth.

If I compare it with VS code, the experience there is more pleasant because there is a delay after the last change until the LSP type hint is adjusted.

I don't think it would be good to disable the LSP type hint in the current line or completely in insert mode, because I personally like to check the result value as I type and it's a welcome hint to see if the result is an option<T> or already a T, especially if you're new to a project or using a new API.

poliorcetics:
> I find that having the hints pop up in the function arguments while typing them can be a bit unnerving, as it shifts the cursor back and forth.

I think part of the issue is that the default `idle-timeout` of helix is really high, can you try with the config below to see if it improves the situation (I expect it will still happen, just less visibly) ? (Test it once the LSP is fully started, on file opening there'll always be a delay for LSP startup)

```toml
[editor]
idle-timeout = 8 # for 120 fps, use 16 for 60
```

marcbull:
> I think part of the issue is that the default `idle-timeout` of helix is really high, can you try with the config below to see if it improves the situation (I expect it will still happen, just less visibly) ? (Test it once the LSP is fully started, on file opening there'll always be a delay for LSP startup)
> 
> ```toml
> [editor]
> idle-timeout = 8 # for 120 fps, use 16 for 60
> ```

For my part, I can say that with a value of 120 (equivalent to 2 seconds) my expectations are met and it behaves as in VS Code. During the input, it then no longer changes erratically.

pascalkuthe:
> > I find that having the hints pop up in the function arguments while typing them can be a bit unnerving, as it shifts the cursor back and forth.
> 
> I think part of the issue is that the default `idle-timeout` of helix is really high, can you try with the config below to see if it improves the situation (I expect it will still happen, just less visibly) ? (Test it once the LSP is fully started, on file opening there'll always be a delay for LSP startup)
> 
> ```toml
> [editor]
> idle-timeout = 8 # for 120 fps, use 16 for 60
> ```

Please don't reduce the idle timeout to such low values. It is known to cause issues and unnecessary overhead. Nothing below 100(ms) should be necessary.

lazytanuki:
I have my `idle-timeout` at 100, I don't really want to have it at 2s just for this feature. Here's an example of what I am talking about:

https://user-images.githubusercontent.com/43273245/219120731-6eaa611b-39e3-41c2-8beb-a726d0cb4df9.mp4

> I don't think it would be good to disable the LSP type hint in the current line or completely in insert mode

I was suggesting a config option here. However, I don't know what the correct defaults should be.

marcbull:
I agree that you should not use idle time for this. It should only be a temporary change to test the effects. I think a separate timeout for the LSP type hint is a good solution.

helix-view/src/document.rs
pascalkuthe: I actually changed my mind on this. Mapping points doesn't work well in general. The right fix here would be to convert these points to ranges by expanding to the word at the specific position and then tracking that range (only removing it when empty just like we do for diagnostics).

I have a pretty good idea how to implement this but it's a bit more effort and I don't have the necessary time right now. In practice this not as large of a problem as I orginally tough because the type hints get refreshed relatively quickly (on idle timeout) so the artificats usually don't stick around for that long. So I think I can do the necessary refactors in a followup PR.

helix-term/src/ui/editor.rs
```diff
@@ -1002,6 +1004,8 @@ impl EditorView {
         event: &MouseEvent,
         cxt: &mut commands::Context,
     ) -> EventResult {
+        cxt.editor.reset_idle_timer();
```
pascalkuthe, line 1007:
```suggestion
	if event.kind != MouseEventKind::Moved{
		cxt.editor.reset_idle_timer();
	}
```

I don't think we should be triggering the idle timeout if the cursor is just moved

helix-view/src/editor.rs
```diff
@@ -386,6 +388,7 @@ impl Default for LspConfig {
             display_messages: false,
             auto_signature_help: true,
             display_signature_help_docs: true,
+            display_inlay_hints: true,
```
pascalkuthe, line 391:
```suggestion
            display_inlay_hints: false,
```

same here

book/src/configuration.md
```diff
@@ -118,9 +118,11 @@ The following statusline elements can be configured:
 | `enable`              | Enables LSP integration. Setting to false will completely disable language servers regardless of language settings.| `true` |
 | `display-messages`    | Display LSP progress messages below statusline[^1]          | `false` |
 | `auto-signature-help` | Enable automatic popup of signature help (parameter hints)  | `true`  |
+| `display-inlay-hints` | Display inlay hints[^2]                                     | `true`  |
```
pascalkuthe, line 121:
```suggestion
| `display-inlay-hints` | Display inlay hints[^2]                                     | `false`  |
```

For now I don't think we should enable this by default until the implementation is improved further from this MVP.

pascalkuthe (changes requested):
I actually think that using the idle timeout for a MVP is fine.

The correct solution (what VSCode  does) would be to have a custom background task that responds to:
* any document modified (if visible)  with a long debounce timeout (3s)
* view moved/focused event with a short debounce timeout (100ms)

This ensures that:
* responses to edits are not too fast
* changes in other splits that affect inlay hints in the current document are also picked up (type check/name resolution are global analysis after all)
* response to scrolling is very fast to reduce noticeable pop-in when possible (while still avoiding excessive requests with debouncing)


Implementing that will require some refactors and is more tricky than just using idle timeout. The implementation would be similar to the diff gutters in some ways (although quite a bit simpler).
I don't think the issues here are bad enough to block on that tough. I can handle that in a followup PR. Implementing this will require centralizing the modifications of the view positions (so we can hook into that) which is something I want to do for the diffmode anyway.

For now I think we should not enable the feature by default. I also added one small suggestion otherwise this LGTM

poliorcetics:
Fixed your comments and added a note for users so that they are warned and can report any jankiness in inlay hints, so we can improve them where needed.

Thanks for the help and the review, it was really helpful!

poliorcetics:
In the line telling a force push happened, the words 'force pushed' are a link to the diff from just before the force push to just atfer

By the way, Github tip that you may not know: to review diffs in force-push, `force pushed` is a link in the screen above

pascalkuthe (approved):
LGTM :+1: 

Thanks for tackling this! It's a cool showcase for what the new virtual text support from #5420. I wanted to add this myself eventually but as I am quite busy right now I am happy you took that over for me. Despite needing some further enhancements this MVP already works quite well.

helix-view/src/view.rs
```diff
@@ -402,9 +404,53 @@ impl View {
...
+            .map(Highlight);
+        let parameter_style = theme
+            .and_then(|t| t.find_scope_index("ui.virtual.inlay-hint.parameter"))
+            .map(Highlight);
+        let other_style = theme
+            .and_then(|t| t.find_scope_index("ui.virtual.inlay-hint"))
+            .map(Highlight);
+        let padding_style = theme
+            .and_then(|t| t.find_scope_index("ui.virtual.inlay-hint.padding"))
+            .map(Highlight);
```
pascalkuthe, line 436: Ah one small think I forgot to add a review comment for:

I don't think this highlight scope should exist.
The LSP spec specifically states that the padding should have the same background color as the rest of the editor and NOT the same background as the virtual text.

Because of the way theme scopes work this will fallback to `ui.virtual.inlay-hint` if a theme doesn't define `ui.virtual.inlay-hint.padding` and therefore would not work as the LSP spec intended. I don't think it's necessary to style this empty space at all since the LSP spec is very specific that it should have the same background color as normal text.

You can simply use `None` here (in that case `ui.text` will be automatically used) :+1: 
```suggestion
        let padding_style = None;
```

poliorcetics: I don't agree with the spec here, I find it confusing in monospaced text to have a regular space I can't reach with my cursor, but done, you're right we should follow the spec regardless

jquesada2016:
What are the steps needed to test this out in Rust? Do I need to add anything to `languages.toml`? Or is this automatic?

I tried both with and without changes, with the changes listed below, and did not see inlay hints.


```toml
# [[language]]
# name = "rust"
#
# [language.config.hints]
# assignVariableTypes = true
# compositeLiteralFields = true
# constantValues = true
# functionTypeParameters = true
# parameterNames = true
# rangeVariableTypes = true
```

I saw thins config on a different issue, but not sure if these apply.

poliorcetics:
@jquesada2016 You'll need to activate the option in the `~/.config/helix/config.toml`:

```toml
[editor.lsp]
display-inlay-hints = true
```

and then use the parameters documented by rust-analyzer to activate what you want amongst inlay hints: https://rust-analyzer.github.io/manual.html#configuration (search for `inlay`)

As an example, here's what I use:

```toml
[[rust]]

[language.config]
inlayHints.bindingModeHints.enable = false
inlayHints.closingBraceHints.minLines = 10
inlayHints.closureReturnTypeHints.enable = "with_block"
inlayHints.discriminantHints.enable = "fieldless"
inlayHints.lifetimeElisionHints.enable = "skip_trivial"
inlayHints.typeHints.hideClosureInitialization = false
```

jquesada2016:
Awesome! That worked! Thank you very much.

Would you happen to know what theme options affectt it?

poliorcetics:
I updated the comment above

jquesada2016:
Thank you very much. This has been THE feature I've been longing for in helix! I work so much faster thanks to inlay hints. I can finally switch to this editor full-time!

David-Else:
It is working great, thanks!

Is it possible to style the hints like VS Code (without the rounded edges I assume)?:

inlay-hints

Dark Plus styles them in white at the moment as it has no `"ui.virtual.inlay-hint" ` yet, and that doesn't look great as they really need to either stand out or be a lot fainter so they are obviously different from the rest of the code:

Screenshot from 2023-02-27 15-43-48

pascalkuthe:
> It is working great, thanks!
> 
> Is it possible to style the hints like VS Code (without the rounded edges I assume)?:
> 
> inlay-hints
> 
> Dark Plus styles them in white at the moment as it has no `"ui.virtual.inlay-hint" ` yet, and that doesn't look great as they really need to either stand out or be a lot fainter so they are obviously different from the rest of the code:
> 
> Screenshot from 2023-02-27 15-43-48

We don't require theme adjustements for merging, those are generally done in seperate PRs (unless @poliorcetics explicitly wants to add those themselves) just like for diff gutters

David-Else:
@pascalkuthe My question was more general, I don't know how to make the text have a different background color than the actual main background (like in the VS Code example). I would like to try and match it for Dark Plus and VS Code if it is possible.

pascalkuthe:
> @pascalkuthe My question was more general, I don't know how to make the text have a different background color than the actual main background (like in the VS Code example). I would like to try and match it for Dark Plus and VS Code if it is possible.

you should just be able to set the `bg` color so: `"ui.virtual.inlay-hint" = {fg = ..., bg = ... }`. You can in theorey do this for all test highlights but those will be overwritten by selection but that is not the case for inlay hints as they can't be selected.

Note that inlay hint background color (if set) does overwrite the cursorline color. It's a matter of taste what should happen in that case. In my opinion the current behavior is fine but it's something to be aware of while theming

evanrichter:
> When might this PR get merged?

when it's ready. if you want to try it now, you can build my fork: https://github.com/evanrichter/helix

David-Else:
@KittyBorgX https://github.com/helix-editor/helix/discussions/5883

swarnimarun:
Nice, I tried a few things and will use it for work for a bit if I find any issues I will let folks here know, in case it isn't merged until then.

What my theme and setup look like atm,

some-rust-code

davxy:
- Does this works fine for other languages as well (other than Rust). If yes, which one? Any experience to share?
- Is there an option + keybind to toggle inline hints when are enabled? Sometimes I want a quick hint but then I may want to work without the hints

paholg:
>     * Is there an option + keybind to toggle inline hints when are enabled? Sometimes I want a quick hint but then I may want to work without the hints

You can set a keybind to `:toggle-option lsp.display-inlay-hints`.

pascalkuthe:
This works for languages whose LS supports inlay hints. The feature is pretty new in the LSP standard so only to most up to date version of actively developed LS will support it

makiri1993:
Was somebody able to get it working with typescript-lsp? Rust worked pretty much without config and very nice 🙌 

My config for ts:

```
[[language]]
name = "typescript"
formatter = { command = "prettier", args = ["--parser", "typescript"] }
auto-format = true
max-line-length = 100
config = { inlayHints = { includeInlayParameterNameHints = "all", includeInlayParameterNameHintsWhenArgumentMatchesName = true, includeInlayFunctionParameterTypeHints = true, includeInlayVariableTypeHints = true, includeInlayVariableTypeHintsWhenTypeMatchesName = true, includeInlayPropertyDeclarationTypeHints = true, includeInlayFunctionLikeReturnTypeHints = true, includeInlayEnumMemberValueHints = true } }


```

erasin:
@makiri1993 https://github.com/helix-editor/helix/issues/2070#issuecomment-1433195804

The config for `golang`, `lua`, `typescript/js`.


```toml
# language.toml config for inlay hit

[[language]]
name = "go"

[language.config.hints]
assignVariableTypes = true
compositeLiteralFields = true
constantValues = true
functionTypeParameters = true
parameterNames = true
rangeVariableTypes = true

[[language]]
name = "lua"

[language.config.Lua.hint]
enable = true
arrayIndex = "Enable"
setType = true
paramName = "All"
paramType = true
await = true


[[language]]
name = "typescript"

[language.config]
hostInfo = "helix"

[language.config.typescript.inlayHints]
includeInlayEnumMemberValueHints = true
includeInlayFunctionLikeReturnTypeHints = true
includeInlayFunctionParameterTypeHints = true
includeInlayParameterNameHints = "all"                       # 'none' | 'literals' | 'all';
includeInlayParameterNameHintsWhenArgumentMatchesName = true
includeInlayPropertyDeclarationTypeHints = true
includeInlayVariableTypeHints = true

[[language]]
name = "tsx"

[language.config]
hostInfo = "helix"

[language.config.typescript.inlayHints]
includeInlayEnumMemberValueHints = true
includeInlayFunctionLikeReturnTypeHints = true
includeInlayFunctionParameterTypeHints = true
includeInlayParameterNameHints = "all"                       # 'none' | 'literals' | 'all';
includeInlayParameterNameHintsWhenArgumentMatchesName = true
includeInlayPropertyDeclarationTypeHints = true
includeInlayVariableTypeHints = true

[[language]]
name = "javascript"

[language.config]
hostInfo = "helix"

[language.config.javascript.inlayHints]
includeInlayEnumMemberValueHints = true
includeInlayFunctionLikeReturnTypeHints = true
includeInlayFunctionParameterTypeHints = true
includeInlayParameterNameHints = "all"                       # 'none' | 'literals' | 'all';
includeInlayParameterNameHintsWhenArgumentMatchesName = true
includeInlayPropertyDeclarationTypeHints = true
includeInlayVariableTypeHints = true

[[language]]
name = "jsx"

[language.config]
hostInfo = "helix"
locale = "zh-CN" 

[language.config.javascript.inlayHints]
includeInlayEnumMemberValueHints = true
includeInlayFunctionLikeReturnTypeHints = true
includeInlayFunctionParameterTypeHints = true
includeInlayParameterNameHints = "all"                       # 'none' | 'literals' | 'all';
includeInlayParameterNameHintsWhenArgumentMatchesName = true
includeInlayPropertyDeclarationTypeHints = true
includeInlayVariableTypeHints = true
```

ref:

Golang https://github.com/golang/tools/blob/master/gopls/doc/settings.md#inlayhint
Lua https://github.com/LuaLS/lua-language-server/wiki/Settings#hint
Typescript https://github.com/typescript-language-server/typescript-language-server#inlay-hints-textdocumentinlayhint

xiantang:
https://github.com/neovim/neovim/pull/20130 related

archseer:
> [neovim/neovim#20130](https://github.com/neovim/neovim/pull/20130) related

We actually already have this implemented. https://github.com/helix-editor/helix/pull/5420

archseer:
@poliorcetics can we add the Rust config you provided + @erasin's config to the defaults? Then this PR should be good to merge 👍🏻

poliorcetics:
Rebased and configs added

pascalkuthe:
Yeah bacoee behaves the same and it's very tricky to change because we do selectionswith syntax highlights. If we applied those to virtual text than virtual text would have the full style of the text (so syntax highlight color, dagnowtic underlines...)

erasin:
there has a problem that the cursor position in insert mode is after the virtual text.

helix:
hx

vscode:
vs

poliorcetics:
I can’t do anything about that, all inlay hints in Helix will behave like that due to how they’re rendered, though I agree it isn’t the most agreable.

Writing a « dumb » fix here (offsetting by one) for type hints would break for arguments and hints at the end of a line

pascalkuthe:
> I've been using this pr for a week or two, and I didn't realize the problem before. There has a problem that the cursor position in insert mode is after the virtual text.
> 
> helix:
> 
> hx
> 
> vscode:
> 
> vs

It's not perfect but yeah as @poliorcetics said that's kind of just how the cursor works in helix.  In insertmode we increase the selection size by one and insert before the cursor block. If you use block cursors its much more clear why it works the way it does: The selection extends one character after the virtual text. Maybe if we switch to zero width cursors oneday that could be solved but i am not sure about that. I could also do some evil trickery here to sepcical case bar cursors but I would rather not do that.

Lordxan:
> > I've been using this pr for a week or two, and I didn't realize the problem before. There has a problem that the cursor position in insert mode is after the virtual text.
> > helix:
> > hx
> > vscode:
> > vs
> 
> It's not perfect but yeah as @poliorcetics said that's kind of just how the cursor works in helix. In insertmode we increase the selection size by one and insert before the cursor block. If you use block cursors its much more clear why it works the way it does: The selection extends one character after the virtual text. Maybe if we switch to zero width cursors oneday that could be solved but i am not sure about that. I could also do some evil trickery here to sepcical case bar cursors but I would rather not do that.

I actually would like to see an optional patch for that (sorry can't do it myself just started learning rust)

erasin:
Since I am a non-ascii characters user, the block cursor is so unfriendly to Chinese characters that I hardly ever use it.

The fact that this problem was not realized before means that it does not affect me much. Hope to become a better experience in the future.

thanks

poliorcetics:
(I'm won't do it in this MR but here is a potential solution):

- If the cursor is next to an inlay hint where the format is `text<cursor><space>` or `<space><cursor><text>`, we could display the cursor closer next to the text, adapting the position of the hint locally to obtain: `text<cursor><hint><space>` and `<space><hint><cursor>text`
- In the case of a conflict (`text<hint>text`), we would just defaut to the current behaviour.

I'm not certain how much work it would be to do that, or if it's even possible, but if it works it would probably solve most of the problematic cases ?

archseer:
I've noticed this issue too, but given that it's a new feature that's also off by default we can resolve this in a follow up.

A quick solution we could use: only display hints outside of insert mode. That would also address issues with the hints shifting the code around while you type. Not sure how desirable this would be, the hints are helpful when typing out method parameters. Though that's also resolved by signature help. It just seems to me that hints are more beneficial during reading the code (normal mode) than writing (insert mode).

poliorcetics:
> only display hints outside of insert mode.

I don't know, things disappearing just before I start typing seems very jarring too.

> It just seems to me that hints are more beneficial during reading the code (normal mode) than writing (insert mode).

I'm also reading code when I type some more, adding code is not done in a vacuum

If we do this it should be an option, because it seems to me to be very much useless to only have type hints outside of insert mode

jquesada2016:
> only display hints outside of insert mode.

I personally think this is a really bad idea. For example, when you're chaining method calls, it's often hard to remember what types the method will return, and being able to see it while in insert mode is very helpful.

erasin:
> only display hints outside of insert mode.

Unlike #6059, type hints is very useful in insert mode. We do not need additional option.

archseer:
Agreed, it was just an idea.

archseer (approved):

evanrichter:
so hyped for all the fantastic virtual text features coming down the pipeline. the next release is going to be off the charts! thanks to the maintainers and everyone involved <3

Hywan:
Awesome!

eulerdisk:
I think inlay-hints theme text style should fallback to the comment style.
Now a lot of theme are "broken", inlay hints for types are like user written types.

cidit:
do i need to use helix manually compiled on main to use this feature ?

gabydd:
No this was released in 23.05 I believe, you just need to enable it in your config. And for some languages you might need to enable it on the actual language server.

PR #5941, opened 2023-02-12
Pretty print the logo.svg

yashpalgoyal1304:
I have not changed anything else in the logo, just 
- automatic pretty print using Notepad++ XML plugin initially, 
- then manual pretty print by hand
- then trim trailing spaces in vscodium

It's weird that 
- github doesn't provide option to "diff" the images in its "rich diff" lol 
- only "2-up", "swipe", "onion"
- gitlab has it since several months already.

image


archseer:
The browser can pretty print the representation for you, and if you're actively working on the file it's always possible to pretty-print it before-hand. We deliberately provide a minimized representation to save on filesize.

yashpalgoyal1304:
- yes i perfectly agree that users likely wont care about it
- i dont think there is any save in filesize as its just  say 29*8 roughly 240 space characters more... 
- but yeah, its a cosmetic change anyway... i felt it feels more polished

yashpalgoyal1304:
u can close this if merging is not wanted

the-mikedavis:
I agree with archseer. I don't see much benefit in formatting SVGs in general, manually reading and editing any complicated SVG is not really productive if possible at all. Plus I don't think we want to make changes the logo any time soon

PR #5943, opened 2023-02-12
feat(ui): display symbol kind in symbol picker

matoous:
Display symbol kind in symbol picker (both local
and workspace symbol picker) in similar vein to
how we do for completions.

This overlaps with: https://github.com/helix-editor/helix/pull/2869 (icon support). But even with that PR landed
it might make sense to provide non-icon support
for symbol kinds.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -10,7 +10,7 @@ use helix_lsp::{
 };
 use tui::{
     text::{Span, Spans},
-    widgets::Row,
+    widgets::{Cell, Row},
 };
 
 use super::{align_view, push_jump, Align, Context, Editor, Open};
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -84,17 +84,48 @@ impl ui::menu::Item for lsp::SymbolInformation {
...
+                lsp::SymbolKind::ARRAY => "array",
+                lsp::SymbolKind::OBJECT => "object",
+                lsp::SymbolKind::KEY => "key",
+                lsp::SymbolKind::NULL => "null",
+                lsp::SymbolKind::ENUM_MEMBER => "enum_member",
+                lsp::SymbolKind::STRUCT => "struct",
+                lsp::SymbolKind::EVENT => "event",
+                lsp::SymbolKind::OPERATOR => "operator",
+                lsp::SymbolKind::TYPE_PARAMETER => "type_parameter",
+                kind => unimplemented!("{:?}", kind),
```
poliorcetics, line 115: this will make helix crash if a new LSP spec is released and helix is updated after an LSP. Using `<unknown>` or some other place holder would be better

poliorcetics (changes requested):

helix-term/src/commands/lsp.rs
dead10ck: I agree with this, but otherwise, this change looks fine to me 👍 nice and easy quality of life improvement

matoous: Done, thanks for suggestion!

matoous:
Closing in favor of: https://github.com/helix-editor/helix/pull/7265

PR #5945, opened 2023-02-12
Add prefix position to `hx` CLI file args

bheylin:
Resolves https://github.com/helix-editor/helix/issues/5437

A cursor position can now be specified using common prefix notation used by vim, nvim, kak, emacs and nano.

`hx +10 Cargo.toml`

This prefix notation also allows the user to set the cursor at the end of the file.

`hx +: Cargo.toml`

This allows `hx` to play nicely with other tools that want to open a file at a given position using prefix notation.

The existing postfix notation can still be used (`hx Cargo.toml:10`) but has been extended to support jumping to the end of a buffer.

`hx Cargo.toml:`

The prefix notation works for each file in the list.

`hx +10 Cargo.toml +20 README`

and can be intermixed with postfix notation.

`hx +10 Cargo.toml README:20`

Prefix notation also supports column indexing.

`hx +10:20 Cargo.toml`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -62,6 +62,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "assert_matches"
+version = "1.5.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9b34d609dfbaf33d6889b2b7106d3ca345eacad44200913df5ba02bfd31d2ba9"
+
+[[package]]
 name = "atoi"
 version = "2.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/5603

bheylin:
> Duplicate of #5603

I would highlight that this PR implements file positioning just like Kakoune's CLI 
so both positioning and the ability to jump to EOF.

the-mikedavis:
This was partially superseded by https://github.com/helix-editor/helix/pull/8521

We wanted to support `+N` syntax only for tools like `less` that assume that editors support that syntax. Since we have that and the suffix way of specifying line/column (e.g. `file.txt:5:3`) I think we support enough ways to open lines/cols. I'm not sure what the use-case is for Kakoune's support of opening at the end with `:`. I think it would be ok to add that syntax (specifically `file.txt:`, not `+:`) in a separate PR but I will close this out because we want to keep the position handling relatively small/simple and supporting multiple `+N`s is not necessary

bheylin:
@the-mikedavis All good Mike, as long as it works!

> I'm not sure what the use-case is for Kakoune's support of opening at the end with :.

I'm also not sure where that came from, I presumed it was some sort of standard across editors.

Anyways my need here was to integrate `helix` into `zellij` _with_ line number support and that's taken care of now.

Issue #5952, opened 2023-02-12
Better highlight breakpoints and current line at which execution is paused at


filipdutescu:
# Description

Currently, breakpoints do not use a conventional symbol, such as the classic filled red circle. The current line at which execution is paused at is also not properly shown. This makes debugging at least somewhat more challenging than it should probably be. 

# Proposal

1. Change the current symbol to the aforementioned conventional one
2. Add new theme highlights for breakpoint and current line colors
3. Highlight the current line at which debugging is stopped at 

# UI mockup
![image](https://user-images.githubusercontent.com/46979062/218331603-f906a601-568f-41a6-8a46-67a0ec696fa7.png)

pascalkuthe:
I think this would be quite easy to accomplish by:
Right now in `helix-term/src/ui/editor.rs:114` the highlight is set to:
``` rust
let style = theme.get("ui.highlight");
```

I think all you would need to do is use a custom theme key here like `ui.highlight.breakpoint` and add it to a couple of themes. That would be a very uncontroversial change.

Regarding the gutter I think #5371 could provide a means to customize the exact icon.

In a TUI our hands are a bit bound with the styling of the icon. 
The red circle is already used for diagnostic errors and it doesn't look as nice as in your mockup because its quite a bit smaller:
![image](https://user-images.githubusercontent.com/61850714/218332638-0a1008a4-95c7-478d-bd71-8e58c408d94a.png)

You could consider experimenting whether setting this to bold or using a different icon works better

filipdutescu:
I used the NerdFonts `na-fa-circle` icon in my mockups.

![image](https://user-images.githubusercontent.com/46979062/218332783-e97cc82d-99a1-41d6-8b34-4e65d695744d.png)

We could probably use that (togglable/customizable of course, since maybe not everyone has nerdfonts). Would you say this change is dependent on #5371 then?

pascalkuthe:
If you want to use a nerdfont symbol this would depend on #2869 instead.

In general we would not be using nerdfont icons by default to ensure helix can be rendered everywhere with just default fonts. Although we could just keep the current icons for now (I think the red hihglight already goes a long way) and only change the icon for use with nerdfonts once #2869 lands

filipdutescu:
I could substitute it for a `B` in the meantime I guess. Is there any resource on how to find default icons? Or how where those already existing chosen?

filipdutescu:
[...]

PR #5957, opened 2023-02-12
feat(debug): highlight current line

filipdutescu:
Add new theme highlight keys, for setting the colour of the breakpoint
character and the current line at which execution has been paused at.
The two new keys are `ui.highlight.frameline` and `ui.debug.breakpoint`.
Highlight according to those keys, both the line at which debugging
is paused at and the breakpoint indicator.

Add an indicator for the current line at which execution is paused
at, themed by the `ui.debug.active` theme scope. Update various themes
to showcase how the new functionality works.

Better icons are dependent on #2869, and as such will be handled in the
future, once it lands.

Closes: #5952

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -276,8 +276,11 @@ These scopes are used for theming the editor interface:
 | `ui.cursor.primary.normal`        |                                                                                                |
 | `ui.cursor.primary.insert`        |                                                                                                |
 | `ui.cursor.primary.select`        |                                                                                                |
+| `ui.debug.breakpoint`             | Breakpoint indicator, found in the gutter                                                      |
+| `ui.debug.active`                 | Indicator for the line at which debugging execution is paused at, found in the gutter          |
 | `ui.gutter`                       | Gutter                                                                                         |
 | `ui.gutter.selected`              | Gutter for the line the cursor is on                                                           |
+| `ui.highlight.frameline`          | Line at which debugging execution is paused at                                                 |
 | `ui.linenr`                       | Line numbers                                                                                   |
 | `ui.linenr.selected`              | Line number for the line the cursor is on                                                      |
 | `ui.statusline`                   | Statusline                                                                                     |
--- a/helix-dap/src/client.rs
+++ b/helix-dap/src/client.rs
@@ -512,4 +512,10 @@ impl Client {
 
         self.call::<requests::SetExceptionBreakpoints>(args)
     }
+
+    pub fn current_stack_frame(&self) -> Option<&StackFrame> {
+        self.stack_frames
+            .get(&self.thread_id?)?
+            .get(self.active_frame?)
+    }
 }
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -93,46 +93,29 @@ impl EditorView {
         let mut line_decorations: Vec<Box<dyn LineDecoration>> = Vec::new();
         let mut translated_positions: Vec<TranslatedPosition> = Vec::new();
 
-        // DAP: Highlight current stack frame position
-        let stack_frame = editor.debugger.as_ref().and_then(|debugger| {
-            if let (Some(frame), Some(thread_id)) = (debugger.active_frame, debugger.thread_id) {
-                debugger
-                    .stack_frames
-                    .get(&thread_id)
-                    .and_then(|bt| bt.get(frame))
-            } else {
-                None
-            }
-        });
-        if let Some(frame) = stack_frame {
-            if doc.path().is_some()
-                && frame
-                    .source
-                    .as_ref()
-                    .and_then(|source| source.path.as_ref())
-                    == doc.path()
-            {
-                let line = frame.line - 1; // convert to 0-indexing
-                let style = theme.get("ui.highlight");
-                let line_decoration = move |renderer: &mut TextRenderer, pos: LinePos| {
-                    if pos.doc_line != line {
-                        return;
-                    }
-                    renderer
-                        .surface
-                        .set_style(Rect::new(area.x, pos.visual_line, area.width, 1), style);
-                };
-
-                line_decorations.push(Box::new(line_decoration));
-            }
-        }
-
         if is_focused && config.cursorline {
             line_decorations.push(Self::cursorline_decorator(doc, view, theme))
         }
 
         if is_focused && config.cursorcolumn {
             Self::highlight_cursorcolumn(doc, view, surface, theme, inner, &text_annotations);
+        }
+
+        // Set DAP highlights, if needed.
+        if let Some(frame) = editor.current_stack_frame() {
+            let dap_line = frame.line.saturating_sub(1) as usize;
+            let style = theme.get("ui.highlight.frameline");
+            let line_decoration = move |renderer: &mut TextRenderer, pos: LinePos| {
+                if pos.doc_line != dap_line {
+                    return;
+                }
+                renderer.surface.set_style(
+                    Rect::new(inner.x, inner.y + pos.visual_line, inner.width, 1),
+                    style,
+                );
+            };
+
+            line_decorations.push(Box::new(line_decoration));
         }
 
         let mut highlights =
@@ -422,6 +405,7 @@ impl EditorView {
         let primary_selection_scope = theme
             .find_scope_index_exact("ui.selection.primary")
             .unwrap_or(selection_scope);
+
         let base_cursor_scope = theme
             .find_scope_index_exact("ui.cursor")
             .unwrap_or(selection_scope);
--- a/helix-view/src/gutter.rs
+++ b/helix-view/src/gutter.rs
@@ -2,7 +2,7 @@ use std::fmt::Write;
 
 use crate::{
     editor::GutterType,
-    graphics::{Color, Style, UnderlineStyle},
+    graphics::{Style, UnderlineStyle},
     Document, Editor, Theme, View,
 };
 
@@ -245,9 +245,9 @@ pub fn breakpoints<'doc>(
     theme: &Theme,
     _is_focused: bool,
 ) -> GutterFn<'doc> {
-    let warning = theme.get("warning");
     let error = theme.get("error");
     let info = theme.get("info");
+    let breakpoint_style = theme.get("ui.debug.breakpoint");
 
     let breakpoints = doc.path().and_then(|path| editor.breakpoints.get(path));
 
@@ -265,30 +265,52 @@ pub fn breakpoints<'doc>(
                 .iter()
                 .find(|breakpoint| breakpoint.line == line)?;
 
-            let mut style = if breakpoint.condition.is_some() && breakpoint.log_message.is_some() {
+            let style = if breakpoint.condition.is_some() && breakpoint.log_message.is_some() {
                 error.underline_style(UnderlineStyle::Line)
             } else if breakpoint.condition.is_some() {
                 error
             } else if breakpoint.log_message.is_some() {
                 info
             } else {
-                warning
+                breakpoint_style
             };
 
-            if !breakpoint.verified {
-                // Faded colors
-                style = if let Some(Color::Rgb(r, g, b)) = style.fg {
-                    style.fg(Color::Rgb(
-                        ((r as f32) * 0.4).floor() as u8,
-                        ((g as f32) * 0.4).floor() as u8,
-                        ((b as f32) * 0.4).floor() as u8,
-                    ))
-                } else {
-                    style.fg(Color::Gray)
-                }
-            };
+            let sym = if breakpoint.verified { "●" } else { "◯" };
+            write!(out, "{}", sym).unwrap();
+            Some(style)
+        },
+    )
+}
 
-            let sym = if breakpoint.verified { "▲" } else { "⊚" };
+fn execution_pause_indicator<'doc>(
+    editor: &'doc Editor,
+    doc: &'doc Document,
+    theme: &Theme,
+    is_focused: bool,
+) -> GutterFn<'doc> {
+    let style = theme.get("ui.debug.active");
+    let current_stack_frame = editor.current_stack_frame();
+    let frame_line = current_stack_frame.map(|frame| frame.line - 1);
+    let frame_source_path = current_stack_frame.map(|frame| {
+        frame
+            .source
+            .as_ref()
+            .and_then(|source| source.path.as_ref())
+    });
+    let should_display_for_current_doc =
+        doc.path().is_some() && frame_source_path.unwrap_or(None) == doc.path();
+
+    Box::new(
+        move |line: usize, _selected: bool, first_visual_line: bool, out: &mut String| {
+            if !first_visual_line
+                || !is_focused
+                || line != frame_line?
+                || !should_display_for_current_doc
+            {
+                return None;
+            }
+
+            let sym = "▶";
             write!(out, "{}", sym).unwrap();
             Some(style)
         },
@@ -304,9 +326,11 @@ pub fn diagnostics_or_breakpoints<'doc>(
 ) -> GutterFn<'doc> {
     let mut diagnostics = diagnostic(editor, doc, view, theme, is_focused);
     let mut breakpoints = breakpoints(editor, doc, view, theme, is_focused);
+    let mut execution_pause_indicator = execution_pause_indicator(editor, doc, theme, is_focused);
 
     Box::new(move |line, selected, first_visual_line: bool, out| {
-        breakpoints(line, selected, first_visual_line, out)
+        execution_pause_indicator(line, selected, first_visual_line, out)
+            .or_else(|| breakpoints(line, selected, first_visual_line, out))
             .or_else(|| diagnostics(line, selected, first_visual_line, out))
     })
 }
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -10,6 +10,7 @@ use crate::{
     view::ViewPosition,
     Align, Document, DocumentId, View, ViewId,
 };
+use dap::StackFrame;
 use helix_vcs::DiffProviderRegistry;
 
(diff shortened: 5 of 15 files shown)


pascalkuthe (changes requested):
I am gonna test this out tomorrow. Just a couple which comments.

I think the CI failure is unrelated and a return would pass fine

helix-term/src/ui/editor.rs
```diff
@@ -110,26 +110,26 @@ impl EditorView {
...
 
-        if is_focused && config.cursorline {
+        if stack_frame.is_none() && is_focused && config.cursorline {
             line_decorations.push(Self::cursorline_decorator(doc, view, theme))
         }
 
-        if is_focused && config.cursorcolumn {
+        if stack_frame.is_none() && is_focused && config.cursorcolumn {
             Self::highlight_cursorcolumn(doc, view, surface, theme, inner, &text_annotations);
         }
```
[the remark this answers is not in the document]
filipdutescu: Didn't make much sense too me too keep them, at least cursorline, since you basically get one with a different style with this change. I was also afraid they would overrule the debug line

helix-term/src/ui/editor.rs
```diff
@@ -110,26 +110,26 @@ impl EditorView {
                     .and_then(|source| source.path.as_ref())
                     == doc.path()
             {
-                let line = frame.line - 1; // convert to 0-indexing
-                let style = theme.get("ui.highlight");
+                let line = frame.line;
```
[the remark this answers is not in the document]
filipdutescu: Interesting, since without it it highlighted the precious line, not the one execution stopped at

helix-term/src/ui/editor.rs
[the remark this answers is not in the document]
pascalkuthe: With the changes you made to the code, the cursorline and cursorcolumn are always disabled when a frame is active so also if the cursor is anywhere else (so not just on the line with the breakpoint).

To make sure that the cursorline doesn't overwrite the frame highlight, simply mode the adsition of the fram highlight after the cursor line and cursorcolumn code. The last line decoration always overwrites previous ones.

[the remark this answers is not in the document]
filipdutescu: That's what I initially did, then thought it was too easy :sweat,_smile: will fix tomorrow

helix-term/src/ui/editor.rs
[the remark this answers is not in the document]
pascalkuthe: Hmm its been a while since I used the debug mode I must admit. Which debug adapter did you test?  I tested with rust and lldb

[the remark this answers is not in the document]
filipdutescu: Well, same as you, since it was at hand at the moment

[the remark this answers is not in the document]
pascalkuthe: i just checked the DAP spec again and the comment here was correct: The `line` field of the `StsckFrame` is 1-based (so the first line is one) while helix uses 0 based indexing (so the first line is 0). 

Sometimes debuggers break at a different line than where the breakpoint is actually placed. Maybe test a bit more to see if you didn't just run into one of those cases

helix-term/src/ui/editor.rs
```diff
@@ -110,26 +110,26 @@ impl EditorView {
...
+                let style = theme.get("ui.highlight.currentline");
```
archseer, line 114: The original name was actually picked for DAP. We could maybe use `current` rather than `currentline`

filipdutescu: I was not aware of that. Maybe we could rename them entirely to `ui.debug.breakpoint` and `ui.debug.current`? Feels a bit more natural.

pascalkuthe: The reason we would want to keep the `highlight` prefix here is that helix falls back to previous scope when a theme doesn't define a scope.

So if  a theme doesn't definine `ui.highlight.current` it would fallback to `ui.highlight` (which basically all themes definine already and is also the way I would expect the fallback to work).
However I agree that the name could maybe give away that it's actually a dab specific color (so it doesn't get reused again a bunch of times like `ui.highlight` did). Maybe we could name it `ui.highlight.currentframe`?

filipdutescu: I could also just go and rename `ui.highlight` to `ui.debug` for each theme, as it is a trivial change. Since highlight is not used for anything other than this and not mentioned in the docs (at least not in `book/src/themes.md`) it should be fine to give it a more meaningful name. I would also add a new key for the triangle character which @archseer requested in the issue.

pascalkuthe: `ui.highlight` is also used for the line in the picker that is currently selected.

filipdutescu: ah I see, missed that. maybe it makes sense to separate them then? have `ui.highlight` and `ui.debug`. I'll have to update all themes, but it's not such a large number, and for the most part it's a trivial change.

pascalkuthe:
I am afraid you misunderstood my hint a bit. All the code you pushed in [b26b98e](https://github.com/helix-editor/helix/pull/5957/commits/b26b98ec952d75fe6be2560c503c5b0a014c7af4) is unnecessary.

Simply add the code below after the following line:
https://github.com/filipdutescu/helix/blob/b26b98ec952d75fe6be2560c503c5b0a014c7af4/helix-term/src/ui/editor.rs#L109

This will

``` rust
if let Some(frame_line) = current_frame{
         let doc = doc.text();
	 highlights = Box::new(syntax::merge(
	    highlights,
	    vec![(frame_highlight_scope, doc.line_to_char(frame_line), doc.line_to_char(frame_line + 1))],
	));
}
```

I just wrote this down off the top of my head so there might be a typo somewhere and the code is probably not perfect but it should set you on the right path.
`current_frame` should be the result of checking whether there is a dap line for the current view (`None` if there isn't one otherwise it should be `Some(line_number)`). 
`frame_highlight_scope` should be the highlight scope you obtained from the theme for highlighting the line.

This will remove any foreground/background colors set by syntax highlighting but leave cursors untouched (since those are merged in afterwards)

filipdutescu:
Seems like I overcomplicated the solution a bunch. I did not realise I could just use `doc.line_to_char(...)`, it was exactly what I, somehow, did not see. It will probably similarly easy to change cursor and selection highlights for the dap line.

Thank you for pointing it out!

filipdutescu:
Fixed all issues, I just need to also update basically all themes, which I'll do during this week.

filipdutescu:
Talked a bit more with  @pascalkuthe (thanks for being so helpful and patient) and decided to go for a more VSCode design, rather than VS. This simplifies the logic _**a lot**_ the logic needed for this feature and makes it so we only need to add at minimum 1 new mandatory highlight, `ui.debug`. Here is an example using my theme:

image

I would still like to allow for more customisation (read as allow VS style frame-lines), but I think it's a lot more important to get the overhaul in an MVP state, rather than focus on customisation and endless architecture discussions. This solution should be a simple one, which should not raise any/many concerns, leading to rapid inclusion.

filipdutescu:
How the default theme will look:

image

book/src/themes.md
```diff
@@ -261,58 +261,62 @@ These scopes are used for theming the editor interface:
...
+| `info`                      | Diagnostics info (gutter)                                                                                                  |
+| `hint`                      | Diagnostics hint (gutter)                                                                                                  |
+| `diagnostic`                | Diagnostics fallback style (editing area)                                                                                  |
+| `diagnostic.hint`           | Diagnostics hint (editing area)                                                                                            |
+| `diagnostic.info`           | Diagnostics info (editing area)                                                                                            |
+| `diagnostic.warning`        | Diagnostics warning (editing area)                                                                                         |
+| `diagnostic.error`          | Diagnostics error (editing area)                                                                                           |
+>>>>>>> 26847d4b (feat(debug): highlight current line)
```
archseer, line 320: conflict marker here

helix-term/src/ui/editor.rs
```diff
@@ -154,6 +120,25 @@ impl EditorView {
...
+        // Set DAP highlights, with possibility to overwrite syntax highlights, if fg set.
+        let highlights: Box<dyn Iterator<Item = HighlightEvent>> =
+            if let Some(frame) = editor.current_stack_frame() {
+                let dap_line = frame.line - 1;
+                if let Some(dap_current_index) = theme.find_scope_index("ui.highlight.frameline") {
+                    let text = doc.text();
+                    let dap_line_start = text.line_to_char(dap_line);
+                    let dap_line_end = text.line_to_char(dap_line + 1);
+                    Box::new(syntax::merge(
+                        highlights,
```
pascalkuthe, line 132: No that we only highlight the background wouldn't it make more sense to highlight the whole line instead of just the text. Otherwise an empty line might endup not being invisible (when compiling with optimizations a breakpoint sometimes endsup on an empty line). Inoeraonally prefer the look too. If you don't want to draw over the gutter that's possible that way too

book/src/themes.md
filipdutescu: Thanks! Somehow missed it

helix-term/src/ui/editor.rs
filipdutescu: I am not too confident in highlighting the whole line, since it would be a behaviour very different from how almost all IDEs work in this case. For the empty line we still have the indicator as well and a highlight if you also have the EoL character enabled. I also think it makes a bit more sense to have it as is since it clearly tells you we are at this point _in the code_, rather than you are at line 3 in you editor which happens to have this code in it.

Ideally, a followup to this PR would make the highlight scope based as well, think debugging a lambda, where only part of the code line would be highlighted, to reflect where _in the code_ execution has stopped at.

Let me know your thoughts on it, but I do think the current approach is a bit more "forgiving" with users and guides their attention in a helpful way.

archseer: I'm split on this too, I see some editors use a whole line highlight, but highlighting a section of the code might be useful in case you're stepping through something like `foo(bar(), baz())` -- you can distinguish what parameter it stopped on

filipdutescu: Exactly what I was thinking about, or for  `.map(|a_var| if var.field { ... })`

filipdutescu: That is something I don't know, but DAP supports column level precision for breakpoints, so there might be a way

filipdutescu:
After talking some more with @archseer, we landed on this colour for the debug highlight, which I find beatiful:

image

helix-term/src/ui/editor.rs
filipdutescu: I think I now understand what you mean, @pascalkuthe, after taking another look. I agree, it should look as this:

image

filipdutescu:
Ok I think I addressed basically every concern for now. I will start work on a better breakpoint experience, which will be scope-aware, allowing you to properly place breakpoints in areas such as lambda functions, which in turn will highlight only that part of the code during debug.

helix-term/src/ui/editor.rs
archseer: If the DAP supports column-level precision for breakpoints then it should be doable. The `stopped` event returns `hitBreakpointIds` which can help determine which breakpoint on that line was hit but we currently don't use that param

filipdutescu: After a bit more talks, the current implementation, meaning not the whole line is highlighted might be the better approach for now, since it allows scope highlighting further down the line (#6238). Once @pascalkuthe finishes his highlighting improvements we could make it possible to achieve this UI as well

archseer: Let's use the whole line for now since that matches current `master` behaviour. We can improve once it's possible to highlight just a part of the line

filipdutescu: ok I will use that behaviour instead and add this more detailed control in #6238. are the current colours to your liking as well? or do you want me to experiment a bit more @archseer?

helix-dap/src/client.rs
```diff
@@ -512,4 +512,14 @@ impl Client {
...
     }
+
+    pub fn current_stack_frame(&self) -> Option<&StackFrame> {
+        if let (Some(frame), Some(thread_id)) = (self.active_frame, self.thread_id) {
+            self.stack_frames
+                .get(&thread_id)
+                .and_then(|bt| bt.get(frame))
+        } else {
+            None
+        }
```
pascalkuthe, line 523:
```suggestion
        self.stack_frames
           .get(&self.thread_id?)?
           .get(self.active_frame?)
```

helix-view/src/gutter.rs
```diff
@@ -265,30 +265,55 @@ pub fn breakpoints<'doc>(
...
+        move |line: usize, _selected: bool, first_visual_line: bool, out: &mut String| {
+            if !first_visual_line || !is_focused {
+                return None;
+            }
+
+            let current_stack_frame = match editor.current_stack_frame() {
+                Some(frame) => frame,
+                None => {
+                    return None;
                 }
```
pascalkuthe, line 303:
```suggestion
            let current_stack_frame = editor.current_stack_frame()?;
```

pascalkuthe (changes requested):
some smaller nits.

There are some unnecessary change in this PR now that are a result of development but now just unecessarily make the diff larger.

The changes to `dap_pos_to_pos` are correct but should be a separate PR as the changes is orthogonal to this one. 

The relevant changes LGTM :+1:  Definitely a visual upgrade (and small fix for the regression from #5420)

helix-view/src/handlers/dap.rs
```diff
@@ -20,11 +21,11 @@ macro_rules! debugger {
...
-    let line = doc.try_line_to_char(line - 1).ok()?;
-    let pos = line + column.saturating_sub(1);
-    // TODO: this is probably utf-16 offsets
-    Some(pos)
+    lsp_pos_to_pos(
+        doc,
+        Position::new(line.saturating_sub(1) as u32, column as u32),
+        OffsetEncoding::Utf16,
+    )
 }
```
pascalkuthe, line 29: The change to `dap_pos_to_pos` is still here. Like I said it's the right change but better split into a seperate PR, potentially together with the type changes to `u32` (as it's an orthogonal change).

pascalkuthe (changes requested):
You forgot to remove one last unnecessary change otherwise this LGTM :+1:

helix-view/src/handlers/dap.rs
filipdutescu: oh yeah, sorry, forgot to commit it after I removed it :sweat:

pascalkuthe (dismissed):
LGTM, thanks for working on this :+1:

filipdutescu:
Thank you very much for guiding me through the process! Looking forward to many more contributions :grin:

helix-term/src/ui/editor.rs
```diff
@@ -134,6 +100,23 @@ impl EditorView {
...
+        // Set DAP highlights, if needed.
+        if let Some(frame) = editor.current_stack_frame() {
+            let dap_line = frame.line.saturating_sub(1) as usize;
+            let style = theme.get("ui.highlight.frameline");
+            let line_decoration = move |renderer: &mut TextRenderer, pos: LinePos| {
+                if pos.doc_line != dap_line {
+                    return;
+                }
+                renderer.surface.set_style(
+                    Rect::new(inner.x, inner.y + pos.visual_line, inner.width, 1),
```
pascalkuthe, line 112: just a reference for other reviews:

Adding `inner,.y` is correct here,  the fact that it was missing before was a regression from #5420

filipdutescu:
I forgot to set the breakpoint to the proper colour, which I did with this rebase as well:

image

helix-view/src/gutter.rs
```diff
@@ -265,30 +265,50 @@ pub fn breakpoints<'doc>(
...
+) -> GutterFn<'doc> {
+    let style = theme.get("ui.debug.active");
+
+    Box::new(
+        move |line: usize, _selected: bool, first_visual_line: bool, out: &mut String| {
+            if !first_visual_line || !is_focused {
+                return None;
+            }
+
+            let current_stack_frame = editor.current_stack_frame()?;
```
archseer, line 299: This should be moved outside of the box so that we don't look up the current stack frame per each line.

archseer: Similar to how `breakpoints` lives outside the `Box::new()` closure above

archseer (changes requested):

helix-view/src/gutter.rs
archseer: (Basically: the parent function is executed once per render, but the boxed closure is executed for each gutter line on screen)

helix-view/src/gutter.rs
```diff
@@ -265,30 +265,50 @@ pub fn breakpoints<'doc>(
...
+            if line != current_stack_frame.line - 1
+                || doc.path().is_none()
+                || current_stack_frame
+                    .source
+                    .as_ref()
+                    .and_then(|source| source.path.as_ref())
+                    != doc.path()
```
archseer, line 306: I would also extract this path check outside the closure (`let is_current = ...`), then return at the top of the closure if no match.

archseer: That way the only thing you need to save outside the closure really is `let stack_fraame_line = current_stack_frame.line - 1`

helix-view/src/gutter.rs
```diff
@@ -265,30 +265,52 @@ pub fn breakpoints<'doc>(
...
+            .and_then(|source| source.path.as_ref())
+    });
+    let should_display_for_current_doc =
+        doc.path().is_some() && frame_source_path.unwrap_or(None) == doc.path();
+
+    Box::new(
+        move |line: usize, _selected: bool, first_visual_line: bool, out: &mut String| {
+            if !first_visual_line
+                || !is_focused
+                || line != frame_line? - 1
```
archseer, line 307: You already `-1` on L293

PR #5964, opened 2023-02-13
Allow LSP server to be stopped

davxy:
Sometimes you don't require the lsp to run, especially if you want to quickly have a look at a file part of a large project.

This PR allows to stop lsp when required via `:lsp-stop` command.

If required, lsp can be restarted via `lst-restart`.

Without having separate commands I was also thinking about having subcommands part a generic `:lsp` command.
That is, remove `lsp-restart`, `:lsp-stop` in favour of a more generic and extensible `lsp`.

The usage would be `:lsp start`, `:lsp stop`, `:lsp restart`.  (restart and start may be probably merged in a single start command)

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1225,6 +1225,24 @@ fn lsp_restart(
     Ok(())
 }
 
+fn lsp_stop(
+    cx: &mut compositor::Context,
+    _args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    let (_view, doc) = current!(cx.editor);
+    let config = doc
+        .language_config()
+        .context("LSP not defined for the current document")?;
+
+    cx.editor.language_servers.stop(config);
+    Ok(())
+}
+
 fn tree_sitter_scopes(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
@@ -2164,6 +2182,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
             aliases: &[],
             doc: "Restarts the Language Server that is in use by the current doc",
             fun: lsp_restart,
+            completer: None,
+        },
+        TypableCommand {
+            name: "lsp-stop",
+            aliases: &[],
+            doc: "Stops the Language Server that is in use by the current doc",
+            fun: lsp_stop,
             completer: None,
         },
         TypableCommand {
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -427,6 +427,19 @@ impl Registry {
         }
     }
 
+    pub fn stop(
+        &mut self,
+        language_config: &LanguageConfiguration,
+    ) {
+        let scope = language_config.scope.clone();
+
+        if let Some((_, client)) = self.inner.remove(&scope) {
+            tokio::spawn(async move {
+                let _ = client.force_shutdown().await;    
+            });
+        }
+    }
+
     pub fn get(
         &mut self,
         language_config: &LanguageConfiguration,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2544,16 +2544,13 @@ fn jumplist_picker(cx: &mut Context) {
         }
     };
 
+    let view = cx.editor.tree.get(cx.editor.tree.focus);
+    let jumps: Vec<_> = view.jumps.iter()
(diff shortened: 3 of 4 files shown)


yashpalgoyal1304:
> having subcommands part a generic `:lsp` command.

+1

sudormrfbin:
An advantage of having separate commands instead of a top level command with subcommands is discoverability -- every action possible is spelled out in the completion popup when you type `:lsp`.

helix-term/src/commands/typed.rs
```diff
@@ -1225,6 +1225,24 @@ fn lsp_restart(
...
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    let (_view, doc) = current!(cx.editor);
+    let config = doc
+        .language_config()
+        .context("LSP not defined for the current document")?;
+
+    cx.editor.language_servers.stop(config);
```
archseer, line 1242: This is missing a whole bunch of cleanup: all documents that use this language server will still refer to it and keep it alive. You'd need to scan over all documents and remove the language server. (There's somewhat similar bits of code in editor.rs)

helix-term/src/commands/typed.rs
```diff
@@ -1225,6 +1225,36 @@ fn lsp_restart(
...
+    let config = doc
+        .language_config()
+        .context("LSP not defined for the current document")?;
+    cx.editor.language_servers.stop(&config);
+
+    cx.editor.documents_mut()
+        .filter(|doc| doc.language_server().map_or(false, |ls| ls.id() == ls_id))
+        .for_each(|doc| {
+            doc.set_language_server(None);
+            log::warn!("XXX STOPPING for {:?}", doc.path());
```
LeoniePhiline, line 1252: You might have forgotten to remove this line before requesting review?

davxy: https://github.com/helix-editor/helix/pull/5964/commits/b2735d1b6e7188871233c9676692f11bbcf43a20

helix-term/src/commands/typed.rs
davxy: Done, hope the cleanup strategy is correct

helix-term/src/commands/typed.rs
```diff
@@ -1225,6 +1225,35 @@ fn lsp_restart(
...
+
+    let config = doc
+        .language_config()
+        .context("LSP not defined for the current document")?;
+    cx.editor.language_servers.stop(&config);
+
+    cx.editor
+        .documents_mut()
+        .filter(|doc| doc.language_server().map_or(false, |ls| ls.id() == ls_id))
+        .for_each(|doc| doc.set_language_server(None));
```
the-mikedavis, line 1252: This is a small style nit but it's considered more idiomatic to use a `for` loop in cases like this:

```
for doc in cx.editor.documents_mut() {
    if doc.language_server().map_or(false, |ls| ls.id() == ls_id) {
        doc.set_language_server(None);
    }
}
```

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each

the-mikedavis: Each of these documents should also have their `diagnostics` cleared

davxy: https://github.com/helix-editor/helix/pull/5964/commits/19371699738cf66770c4af4d8a7fe2af708a5279

helmesjo:
> An advantage of having separate commands instead of a top level command with subcommands is discoverability -- every action possible is spelled out in the completion popup when you type `:lsp`.

Yeah I've noticed this at times, but isn't this rather a separate issue (bug) that should be solved? I mean, the subcommands are known so there is no reason `helix` can't list them. Top level command + subcommands almost lalways pays for itself in the long run.

the-mikedavis:
Looks like the CI is failing because of some trailing whitespace - you can fix it with `cargo fmt`

helix-term/src/commands/typed.rs
```diff
@@ -1225,6 +1225,37 @@ fn lsp_restart(
...
+
+    let ls_id = doc
+        .language_server()
+        .map(|ls| ls.id())
+        .context("LSP not running for the current document")?;
+
+    let config = doc
+        .language_config()
+        .context("LSP not defined for the current document")?;
+    cx.editor.language_servers.stop(&config);
```
the-mikedavis, line 1247:
```suggestion
    cx.editor.language_servers.stop(config);
```

looks like clippy complains about the redundant `&`

the-mikedavis (commented):
Just a small clippy lint but otherwise this looks good 👍

the-mikedavis (approved):

Issue #5947, opened 2023-02-12
set-option completion shown after first argument is set


askreet:
### Summary

Noticed this while looking at #5939. If you start typing `:set-option bufferline` you get completion while typing `bufferline`. If you then press space you get _new_ completion of options for the 2nd argument. Perhaps we want completion that is related to the option, but at least the completion probably should not be displayed for the 2nd argument.

### Reproduction Steps

 

I tried this:

1. Start typing `set-option mouse true`.

I expected this to happen:

No completion after `set-option mouse` while typing the 2nd argument.

Instead, this happened:

New completion showing all options is shown.

### Platform

macOS

### Terminal Emulator

iTerm 2

### Helix Version

22.12-257-ga4751db8

the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/issues/2772

Issue #5939, opened 2023-02-12
2 level documentation popup for `:set`


yashpalgoyal1304:
In helix 22.12, in the documentation popup for `:set key field` is shown only for this command.

I am not saying that it should show all the way down to `field` (3rd level), but it would be really nice if the documentation for the `key` could also be shown when it is entered.

<img src="https://user-images.githubusercontent.com/19423063/218294044-e18d0b43-7300-4304-998c-f2bf6ce76479.png" width=550px/>

askreet:
I think this is a great idea. Started on a possible implementation [here](https://github.com/helix-editor/helix/compare/master...askreet:helix:issue-5939/set-option-detailed-docs?expand=1). Hope to get back to it soon.

askreet:
Just leaving some notes here for posterity, I have that branch above that does the easy part but the hard part here is that documentation for options is not embedded into the project. The first hurdle would be, do we _want_ it to be?

I'm generally in favor of this because I like how discoverable Helix is in other areas, with default `which-key` like behaviors almost everywhere and the auto-help popups, so this feels like a natural extension to me.

Assuming we do want this level of discoverability, there's a few option (and they aren't mutually exclusive to me).

1. Start building something like `help` in vim and/or `apropo` in Emacs. `<space>?` is good for key->behavior lookups but doesn't add much detail. Being able to type `help function_name` and get longform prose is the idea here.
2. Slurp in existing documentation into the binary or runtime directory where it can be looked up. Doc comments are already parsed by rustdoc so it's a matter of putting them somewhere they can be retrieved.
3. Restructure the way options are stored internally to be more useful to discoverability and validation. At present there's a few things we can't do generically:
    * Show contextual help for an option.
    * Validate whether an option's value is correct. This is done ad-hoc in Validate event handlers, sometimes.
    * Suggest option values, particularly if they are an enum (e.g. `bufferline`).
    * Validate the number of arguments. This is done ad-hoc in Validate event handlers, sometimes.

For #3 one could imagine going in the direction of:

```rust
struct HelixOption {
    name: String,
    doc: String,
    type: HelixOptionType,
}

enum HelixOptionType {
    Bool,
    AnyString,
    RegexpString(regexp),
    AnyNumber,
    RangeNumber(range),
}
```

yashpalgoyal1304:
> 1. `<space>?`

thats already reserved for "command palette" if i'm right.

rahji:
[...]

PR #5966, opened 2023-02-13
Only complete appropriate arguments for typed commands.

### LINKED ISSUE #2772: Improve help/completion info for set-option to suggest valid values

askreet:
Edit: This doesn't _fix_ #2772, it does enable the fix. It fixes [my issue](https://github.com/helix-editor/helix/issues/5947) that was closed as duplicate, though.

I think there's a bigger change we can make here to address some other code duplication that looks something like setting argument types of commands within the TypableCommand struct and using that data both for completion _and_ for verification of the number of arguments. Currently, some commands have checks for argument length and emit errors while others silently ignore trailing arguments.

~I'm not _thrilled_ about having variadic length completers represented as different enum variants but as we statically allocate all the command configuration I couldn't use a dynamic container.~ I've also left the implementation of arity 2 completers for `set-option`, even though I haven't implemented a completer for option values (I've got another patch I'm working on that will dive into option metadata for #5939 and can tie that in).

Edit: I realized after reading my own PR that I didn't need different variants for variadic completers, but I do need lifetime annotations to use slices. I'm probably irrationally allergic to lifetime annotations, but this does seem a bit cleaner.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -5,7 +5,7 @@ use crate::job::Job;
 
 use super::*;
 
-use helix_core::encoding;
+use helix_core::{encoding, shellwords::Shellwords};
 use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::editor::{Action, CloseError, ConfigEvent};
 use serde_json::Value;
@@ -18,7 +18,49 @@ pub struct TypableCommand {
     pub doc: &'static str,
     // params, flags, helper, completer
     pub fun: fn(&mut compositor::Context, &[Cow<str>], PromptEvent) -> anyhow::Result<()>,
-    pub completer: Option<Completer>,
+    /// What completion methods, if any, does this command have?
+    pub signature: CommandSignature,
+}
+
+impl TypableCommand {
+    fn completer_for_argument_number(&self, n: usize) -> &Completer {
+        match self.signature.positional_args.get(n) {
+            Some(completer) => completer,
+            _ => &self.signature.var_args,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct CommandSignature {
+    // Arguments with specific completion methods based on their position.
+    positional_args: &'static [Completer],
+
+    // All remaining arguments will use this completion method, if set.
+    var_args: Completer,
+}
+
+impl CommandSignature {
+    const fn none() -> Self {
+        Self {
+            positional_args: &[],
+            var_args: completers::none,
+        }
+    }
+
+    const fn positional(completers: &'static [Completer]) -> Self {
+        Self {
+            positional_args: completers,
+            var_args: completers::none,
+        }
+    }
+
+    const fn all(completer: Completer) -> Self {
+        Self {
+            positional_args: &[],
+            var_args: completer,
+        }
+    }
 }
 
 fn quit(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -5,19 +5,45 @@ use crate::job::Job;
...
     pub doc: &'static str,
     // params, flags, helper, completer
     pub fun: fn(&mut compositor::Context, &[Cow<str>], PromptEvent) -> anyhow::Result<()>,
-    pub completer: Option<Completer>,
+    /// What completion methods, if any, does this command have?
+    pub completer: CompletionConfig<'a>,
+}
+
+#[derive(Clone)]
+pub enum CompletionConfig<'a> {
```
pascalkuthe, line 25: I don't think an enum is an amazing fit here. NoComoletions is just a subset of positional and this approach doesn't allow commands which have a couple positional arguments followed by varags. Instead you should simply use a list of positional arguments followed by an optional vararg completed:

``` rust
strict CommandSignature{
   postional_args: &'static [Option<Completer>],
   var_arcs: Option<Completer>
} 
```

You can get the compler for field n with:
``` rust
self.positional_arts.get(n).unwrap_or(self.var_args)
```

Note that my suggestion uses a `'ststic` slice. Tgis removes the lifetime argument you were unaopy with and is a good fit here because currently all tupable commands are static anyway and if we move away from that we need to heap allocate anyway for that cade a cow would be used)

helix-term/src/commands/typed.rs
```diff
@@ -2115,413 +2143,414 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
...
-            completer: None,
+            completer: CompletionConfig::NoCompletion,
         },
         TypableCommand {
             name: "encoding",
             aliases: &[],
             doc: "Set encoding. Based on `https://encoding.spec.whatwg.org`.",
             fun: set_encoding,
-            completer: None,
+            completer: CompletionConfig::NoCompletion,
```
pascalkuthe, line 2328: Probably not in this  this PR but in the future having an enum completer would be nice for cases like this. An autoamtic way to do that would also be useful for #4411 but I am not sure if serde allows that.

pascalkuthe (changes requested):
This is a nice UX Improvement and has potential for even netter followup improvements. I like that this is quite small in scope. The diff is only a bit larger because you and to.change the comlleteres for all typable commands 👍

helix-term/src/commands/typed.rs
askreet: Yeah I want to have more metadata about options in general, some discussion in #5939.

askreet:
Thanks for the tips @pascalkuthe, pushed up some updates. I think your suggestion is much cleaner, especially using 'static scope.

helix-term/src/commands/typed.rs
pascalkuthe: Yeah that should be left to future work, just something that came to mind while reviewing

helix-term/src/commands/typed.rs
```diff
@@ -18,7 +18,36 @@ pub struct TypableCommand {
...
+    }
+}
+
+#[derive(Clone)]
+pub struct CommandSignature {
+    // Arguments with specific completion methods based on their position.
+    positional_args: &'static [Option<Completer>],
+
+    // All remaining arguments will use this completion method, if set.
+    var_args: Option<Completer>,
```
pascalkuthe, line 41: Actually thinking about it more I don't think we need to use `Option<Completer>` here we can just use `ui::completers::none` instead.
This will allow creating some utility functions (see below) to make the API a bit cleaner.

helix-term/src/commands/typed.rs
```diff
@@ -18,7 +18,36 @@ pub struct TypableCommand {
...
+}
+
+impl CommandSignature {
+    const fn none() -> Self {
+        Self {
+            positional_args: &[],
+            var_args: None,
+        }
+    }
```
pascalkuthe, line 50: I think adding a couple more utility constructs will go a long way towards keeping the command list short and readable.

```suggestion
    const fn none() -> Self {
        Self {
            positional_args: &[],
            var_args: completers::none,
        }
    }
    const fn single(completer: &'static Completer) -> Self {
        Self {
            positional_args: slice::from_ref(completer),
            var_args: None,
        }
    }
    const fn repeat(completer: Completer) -> Self {
        Self {
            positional_args: &[],
            var_args: completer,
        }
    }
```

pascalkuthe (changes requested):
I had a couple more ideas to make the API even more compact

helix-term/src/commands/typed.rs
askreet: You know it's funny I actually wrote basically the same thing first and wasn't sure if people would be comfortable with the layers of indirection. I agree and will put it back! :-)

askreet:
@pascalkuthe Updates ready! :)

pascalkuthe:
> @pascalkuthe Updates ready! :)

Ah I actually mean that the `positional_arguments` should also be just a `Completer` instead of `Option<Completer>` would remove all the `Some(..)` wrapping (and you can always use `completers:;none` instead of `None`)

askreet:
Ooh, yeah. Will do.

pascalkuthe (dismissed):
LGTM now :+1:

PR #5967, opened 2023-02-13
Expand contributing docs

magopian:
Just a couple tips and tricks that @pascalkuthe gave on the [matrix server](https://matrix.to/#/!zMuVRxoqjyxyjSEBXc:matrix.org/$k9ILSGRe2Ohj7ZpL2FnLnd2LlllzCEEwlevB_h0k1bQ?via=matrix.org&via=mozilla.org&via=envs.net)

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -8,7 +8,7 @@ Some suggestions to get started:
 - Help with packaging on various distributions needed!
 - To use print debugging to the [Helix log file][log-file], you must:
   * Print using `log::info!`, `warn!`, or `error!`. (`log::info!("helix!")`)
-  * Pass the appropriate verbosity level option for the desired log level. (`hx -v <file>` for info, more `v`s for higher severity inclusive)
+  * Pass the appropriate verbosity level option for the desired log level. (`hx -v <file>` for info, more `v`s for higher verbosity)
   * Want to display the logs in a separate file instead of using the `:log-open` command in your compiled Helix editor? Start your debug version with `cargo run --log foo.log` and in a new terminal use `tail -f foo.log`
 - Instead of running a release version of Helix, while developping you may want to run in debug mode with `cargo run` which is way faster to compile
 - Looking for even faster compile times? Give a try to [mold](https://github.com/rui314/mold)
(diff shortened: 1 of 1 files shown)


docs/CONTRIBUTING.md
```diff
@@ -9,6 +9,9 @@ Some suggestions to get started:
 - To use print debugging to the [Helix log file][log-file], you must:
   * Print using `log::info!`, `warn!`, or `error!`. (`log::info!("helix!")`)
   * Pass the appropriate verbosity level option for the desired log level. (`hx -v <file>` for info, more `v`s for higher severity inclusive)
+  * Want to display the logs in a separate file instead of using the `:log-open` command in your compiled Helix editor? Start your debug version with `cargo run -- --log foo.log` and in a new terminal use `tail -f foo.log`
+  * Instead of running a release version of Helix, while developping you may want to run in debug mode with `cargo run` which is way faster to compile
+  * Looking for even faster compile times? Give a try to [mold](https://github.com/rui314/mold)
```
the-mikedavis, line 14: These are sub-bullets of the print debugging bullet but they should be top-level since they're unrelated

magopian: You are right, thanks for spotting this mistake! I've corrected it now, and added another wording change. What do you think?

docs/CONTRIBUTING.md
```diff
@@ -8,7 +8,10 @@ Some suggestions to get started:
 - Help with packaging on various distributions needed!
 - To use print debugging to the [Helix log file][log-file], you must:
   * Print using `log::info!`, `warn!`, or `error!`. (`log::info!("helix!")`)
-  * Pass the appropriate verbosity level option for the desired log level. (`hx -v <file>` for info, more `v`s for higher severity inclusive)
+  * Pass the appropriate verbosity level option for the desired log level. (`hx -v <file>` for info, more `v`s for higher verbosity)
+  * Want to display the logs in a separate file instead of using the `:log-open` command in your compiled Helix editor? Start your debug version with `cargo run --log foo.log` and in a new terminal use `tail -f foo.log`
```
the-mikedavis, line 12:
```suggestion
  * Want to display the logs in a separate file instead of using the `:log-open` command in your compiled Helix editor? Start your debug version with `cargo run -- --log foo.log` and in a new terminal use `tail -f foo.log`
```

very minor point but without the double-dash, `cargo` thinks that the `--log` argument belongs to it instead of to `hx` so it prints usage and exits

the-mikedavis (commented):
just some minor comments, I thinks this looks good otherwise 👍

docs/CONTRIBUTING.md
magopian: Ah, yes, indeed, i'll fix this straight away 👍

magopian:
Hopefully all fixed now, sorry for the back and forth, I should have caught those earlier. Thanks for pointing the issues/typos out!

pascalkuthe (approved):
LGTM now 👍 I have answered questions like these quite a few tines on matrix so it's good to document this somewhere.

the-mikedavis (approved):

PR #5983, opened 2023-02-14
Ken cox94 reload all

### LINKED ISSUE #5665: reload_all command should ignore scratch buffers
I often run a script to strip trailing whitespace from all changed files since the last git commit, and use `:reload-all` so helix picks up on the new changes. However, if I have any scratch buffer open it will short-circuit the reloading process, which I believe is not good UX. It would make sense if a file could not be found for normal buffers, and a `:reload-all!` command could be added to ignore such situations.

This should probably be a trivial change. I'll work on it on my own fork when I have time and then make a PR for it.

KenCox94:
should resolve #5665

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1181,12 +1181,14 @@ fn reload(
 
 fn reload_all(
     cx: &mut compositor::Context,
-    _args: &[Cow<str>],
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/5911

PR #5986, opened 2023-02-14
Jump to symbol ranges in LSP goto commands

the-mikedavis:
This follows prior changes like #4713: we select the range given by the language server rather than the starting point.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -5,7 +5,7 @@ use helix_lsp::{
         self, CodeAction, CodeActionOrCommand, CodeActionTriggerKind, DiagnosticSeverity,
         NumberOrString,
     },
-    util::{diagnostic_to_lsp_diagnostic, lsp_pos_to_pos, lsp_range_to_range, range_to_lsp_range},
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -195,15 +195,15 @@ fn jump_to_location(
...
-    };
-    doc.set_selection(view.id, Selection::point(new_pos));
+    let new_range =
+        if let Some(new_range) = lsp_range_to_range(doc.text(), location.range, offset_encoding) {
+            new_range
+        } else {
+            log::warn!("lsp position out of bounds - {:?}", location.range);
+            return;
+        };
+    doc.set_selection(view.id, Selection::single(new_range.anchor, new_range.head));
```
the-mikedavis, line 206: We might want to normalize the direction of the new range, to make behavior between language servers consistent?

pascalkuthe: LSP ranges are not allowed to be reversed so `start` should always be `<=` `end`= so I don't think the behavior would vary between servers.

pascalkuthe (approved):
LGTM, nice to be consistent here :+1:

PR #5993, opened 2023-02-14
Add test-running capability

dradtke:
This PR is an initial proof-of-concept for adding test-running support to Helix, and is heavily inspired by [vim-test](https://github.com/vim-test/vim-test). To execute a test, open its file, and then execute `<Space> t b`. (Currently, only Cargo-based tests are supported)

Running tests from Helix requires two things:

1. The ability to execute arbitrary commands.
2. Logic to determine the correct command to run for a given file.

### Executing Arbitrary Commands

To execute arbitrary commands, a new module has been added for defining command executors. Because Helix does not (yet) have an [integrated terminal](https://github.com/helix-editor/helix/pull/4649), there are two executors defined: `Tmux`, which integrates with tmux to execute test commands in another pane, and `Direct`, which executes the test command as a background process and prints the results.

Additional executors can easily be added for different multiplexers. This is equivalent to what `vim-test` calls [strategies](https://github.com/vim-test/vim-test#strategies). The `Tmux` executor implemented here is similar to the [vimux](https://github.com/preservim/vimux) strategy.

### Determining Test Commands

A new `helix-test` project has been added to hold the logic for determining what command to run. The relevant function will walk up the file tree, attempting to match against all known test runners, and returning the first one that matches. For example, as soon as `Cargo.toml` is seen, the `Cargo` test runner is returned. This test runner will then be asked to build the test command for the relevant file.

`vim-test` supports quite a few executors, all of which are good candidates for porting over to Helix as well.

### Next Steps

I'm open to feedback on both the feature as a concept, and its specific implementation here.

One option as well would be to hold off until Helix's plugin system has stabilized, and then implement this as a plugin, but given Helix's large amount of core functionality, I think this would not be out-of-place living within Helix itself.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1210,6 +1210,7 @@ dependencies = [
  "helix-dap",
  "helix-loader",
  "helix-lsp",
+ "helix-test",
  "helix-tui",
  "helix-vcs",
  "helix-view",
@@ -1228,6 +1229,14 @@ dependencies = [
  "tokio-stream",
  "toml",
  "which",
+]
+
+[[package]]
+name = "helix-test"
+version = "0.6.0"
+dependencies = [
+ "anyhow",
+ "helix-core",
 ]
 
 [[package]]
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -31,6 +31,7 @@ helix-core = { version = "0.6", path = "../helix-core" }
 helix-view = { version = "0.6", path = "../helix-view" }
 helix-lsp = { version = "0.6", path = "../helix-lsp" }
 helix-dap = { version = "0.6", path = "../helix-dap" }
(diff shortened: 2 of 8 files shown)


kirawi:
FYI Helix is supposed to be language agnostic.

the-mikedavis:
I think this belongs as a plugin and not in core. Ideally test-running would be provided by an LSP-like cross-language spec like [BSP](https://build-server-protocol.github.io/docs/specification/)

dradtke:
Interesting, I wasn't aware of BSP. I'll certainly take a look at it.

Is there any movement currently in terms of things like tmux integration for executing arbitrary commands?

archseer:
You should be able to just shell out to `tmux`  for that.

pascalkuthe:
we are not going to be going forward with this and this pr is pretty stale anyway so closing

PR #5995, opened 2023-02-15
tokyonight_transparent.toml

aniketkharel:
changed background to white to acheive transparency.


archseer:
We introduced theme inheritance so you can locally override this without having to copy the whole theme. There's also https://github.com/helix-editor/helix/pull/5839

PR #6000, opened 2023-02-15
Ignore invalid file URIs from LSP

xJonathanLEI:
This PR changes a call to `unwrap()` to `match`. Currently the editor crashes when the LSP publishes diagnostics with an unsupported URI. With this PR it simply ignores them.

Submitting this patch because the [Cairo language server](https://github.com/starkware-libs/cairo/tree/main/crates/cairo-lang-language-server) crashes Helix when it sends a file URI like `vfs://xxxxxx` where `vfs` refers to the internal in-memory file system of the Cairo language server.

I'm not sure if this is considered a bug of the Cairo lang server as I don't know for sure if schemes other than `file` is allowed. But in any case I think we shouldn't panic here.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -703,7 +703,13 @@ impl Application {
                         }
                     }
                     Notification::PublishDiagnostics(mut params) => {
-                        let path = params.uri.to_file_path().unwrap();
+                        let path = match params.uri.to_file_path() {
+                            Ok(path) => path,
+                            Err(_) => {
+                                log::error!("Unsupported file URI: {}", params.uri);
+                                return;
+                            }
+                        };
                         let doc = self.editor.document_by_path_mut(&path);
 
                         if let Some(doc) = doc {


pascalkuthe (approved):
I do think this is a bug in the Cairo LS as the `uri` is expected to refer to a valid file (otherwise the notification is useless). Howber the LSP standard is pretty vaque and only states that the uri fieldust.ne parsed as a valid URI but not as a valid FILE uri:
> Many of the interfaces contain fields that correspond to the URI of a document. For clarity, the type of such a field is declared as a [DocumentUri](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#documentUri). Over the wire, it will still be transferred as a string, but this guarantees that the contents of that string can be parsed as a valid URI.

I think being defensive is good here (easy fix for a panic) especially as buggy and non standard compliant LSPs are sadly not that rare. 

LGTM 👍

PR #6002, opened 2023-02-15
Add basic support for F# syntax highlighting.

### LINKED ISSUE #4943: F# for Helix?
I wonder why there is no support for F# in Helix until now. Is there any plan to add F# to Helix in the near future?

askreet:
This integrates the [Nsidorenco's TS grammar](https://github.com/Nsidorenco/tree-sitter-fsharp) and provides some basic mapping of elements to syntax highlighting groups.

Some notes:
* A lot of the code I looked at had some errors parsing where I'd use `:tree-sitter-scopes` and see `["ERROR", "ERROR", "l"]` in place of useful scope mappings. As a result some files I looked at were partially highlighted. I suspect these are bugs in the TS grammar that will be ironed out over time, but didn't know enough to be sure. (Screenshot below).
* I'm a _super_ novice at F#, having written only a basic CLI program, so I'm sure these mappings could be improved. I offer this PR in a "something is better than nothing" state.

Fixes #4943.

image

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2167,3 +2167,16 @@ comment-token = "#"
 [[grammar]]
 name = "hosts"
 source = { git = "https://github.com/ath3/tree-sitter-hosts", rev = "301b9379ce7dfc8bdbe2c2699a6887dcb73953f9" }
+
+[[language]]
+name = "fsharp"
+scope = "source.fsharp"
+injection-regex = "^fsharp$"
+file-types = ["fs"]
+roots = []
+comment-token = "//"
+indent = { tab-width = 4, unit = "    " }
+
+[[grammar]]
+name = "fsharp"
+source = { git = "https://github.com/Nsidorenco/tree-sitter-fsharp", rev = "700c8371016ec9dc68ac31c67328ef5c8b216cf5" }
--- a/runtime/queries/fsharp/highlights.scm
+++ b/runtime/queries/fsharp/highlights.scm
@@ -1,0 +1,33 @@
+"module" @keyword.other
+
+[
+  "let"
+  "rec"
+  "if"
+  "else"
+  "match"
+  "with"
+] @keyword.control
+
+[
+  "="
+] @keyword.type
+
+[
+  "static"
+  "member"
+] @keyword.other
+
+(identifier) @type
+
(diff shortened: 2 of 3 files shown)


languages.toml
```diff
@@ -2167,3 +2167,16 @@ comment-token = "#"
...
 source = { git = "https://github.com/ath3/tree-sitter-hosts", rev = "301b9379ce7dfc8bdbe2c2699a6887dcb73953f9" }
+
+[[language]]
+name = "fsharp"
+scope = "source.fsharp"
+injection-regex = "^fsharp$"
+file-types = ["fs"]
+roots = []
+comment-token = "//"
+indent = { tab-width = 4, unit = "    " }
```
the-mikedavis, line 2178: It would be good to add a configuration for https://github.com/fsharp/FsAutoComplete as well, it looks like that's the most popular language server for F#

languages.toml
```diff
@@ -2167,3 +2167,16 @@ comment-token = "#"
...
+
+[[grammar]]
+name = "fsharp"
+source = { git = "https://github.com/Nsidorenco/tree-sitter-fsharp", rev = "700c8371016ec9dc68ac31c67328ef5c8b216cf5" }
```
the-mikedavis, line 2182: It looks like this grammar is a work-in-progress that is being actively worked on. It might be better to wait for some of the missing points or else the highlighting will be rough in some cases as they say https://github.com/Nsidorenco/tree-sitter-fsharp#status

runtime/queries/fsharp/highlights.scm
```diff
@@ -0,0 +1,33 @@
+"module" @keyword.other
```
the-mikedavis, line 1: We use these scopes for syntax highlights queries: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

Some of these like keyword.other and keyword.type should use those scopes instead

askreet:
Thanks @the-mikedavis! After reviewing this it's clear my attempt at highlights needs another pass and can be much cleaner. Since we prefer to wait for the TS implementation to stabilize, I'll shelve this for now. I'll set a reminder to check back in their progress in a few months.

PR #1755, opened 2022-03-05
Add lsp signature help

sudormrfbin:
Implements signature help, aka parameter hints:

command

### TODO

- [x] Obstructs the completion popup
- [ ] Bar cursor in insert mode can be seen through the popup if popup overlays the cursor
- [x] Popup is not updated on backspace
- [x] `editor.lsp.signature-help` config
- [x] Keybind for manual triggering (`Ctrl-S` in insert mode)

### Future Work

1. Show all signatures for a method (useful for overloaded methods)

PR #6006, opened 2023-02-15
Use Popup::ignore_escape_key helper for completion

the-mikedavis:
The completion component has a separate branch for handling the Escape key but it can use the `ignore_escape_key` helper added for signature-help instead (#1755).

This should not cause a behavior change - it's just cleaning up the completion component.


pascalkuthe (approved):
LGTM :+1:  Considering that the `ignore_escape_key` lists the completion key as an example usecase its indeed odd that we didn't use it here yet :smile:

Issue #5958, opened 2023-02-12
Issues with pyright-langserver


Lunarmagpie:
### Summary

Using the pyright-langserver for my linter results in async timeout error whenever I do anything.

### Reproduction Steps

Install the pyright langsever as described on the wiki page.

This is my global config.
```toml
[[language]]
name = "python"
roots = ["pyproject.toml"]
language-server = { command = "/usr/local/bin/pyright-langserver", args = ["--stdio"] }
config = {} # <- this is the important line

formatter = { command = "black", args = ["--quiet", "-"]}
auto-format = true
```

### Helix log

Not sure where the exact error was so I put everything.

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-02-10T20:12:43.970 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["[scratch]"]
2023-02-10T20:13:24.302 helix_view::editor [ERROR] Failed to initialize the LSP for `source.python` { cannot find binary path }
2023-02-10T20:13:48.899 helix_view::editor [ERROR] editor error: Could not load theme: No such file or directory (os error 2)
2023-02-10T20:14:48.424 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2023-02-10T20:14:55.819 helix_term::application [WARN] failed to load theme `monokai` - No such file or directory (os error 2)
2023-02-10T20:14:55.820 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2023-02-10T20:15:17.382 helix_term::application [WARN] failed to load theme `monokai` - No such file or directory (os error 2)
2023-02-10T20:15:19.984 helix_view::editor [ERROR] Failed to initialize the LSP for `source.python` { cannot find binary path }
2023-02-10T20:19:31.972 helix_view::editor [ERROR] Failed to initialize the LSP for `source.python` { cannot find binary path }
2023-02-10T20:19:48.649 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2023-02-10T20:20:44.265 helix_view::editor [ERROR] editor error: no last accessed buffer
2023-02-10T20:21:43.280 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2023-02-10T20:29:19.745 helix_view::editor [ERROR] Failed to initialize the LSP for `source.python` { cannot find binary path }
[...]

PR #6010, opened 2023-02-15
LSP: Add request ID to request timeout message

the-mikedavis:
This improves error logging for requests - without the ID it's hard to know which request is the one that timed out.

This is helpful for debugging #5958 for example.


pascalkuthe (approved):
Nice that is defintly useful for debugging LSP requests and will also make logs submitted in issues more useful :+1:

PR #6016, opened 2023-02-16
Fix: decode lsp url for workspace_diagnostics_picker

erasin:
When I used CJK characters as the path, the lsp returned the encoded content, and the correct path would be displayed only after decoding.

dia2
WX20230216-174046


archseer:
This seems unnecessary since the `url` library already uses `percent-encoding`. It's probably `url.path()` that re-encodes the path

archseer:
Yup: "Return the path for this URL, as a percent-encoded ASCII string."

erasin:
Remove `percent-encoding`,The `url.to_file_path()` has return decode path.

thanks for review.

pascalkuthe (approved):
This LGTM now :+1: We simply used the wrong function here. We already use `to_file_path` everywhere else.

Issue #5709, opened 2023-01-27
Invalid hexcodes in several themes `emacs`, `monokai_aqua`, `papercolor-light` (Should be checked in CI at theme PRs?)


LeoniePhiline:
Reproduction:

1. Open two terminals side by side
2. In one terminal, watch the hx log: `tail -f ~/.cache/helix/helix.log`
3. In the other terminal, open `hx`
4. Type `:theme`
5. Use the `Tab` key to go through all themes and observe the log.

# Theme `emacs`

```
helix_view::theme [WARN] Theme: malformed hexcode: highlight
```

# Theme `monokai_aqua`

```
helix_view::theme [WARN] Theme: malformed hexcode: light-black
helix_view::theme [WARN] Theme: malformed hexcode: light-black
helix_view::theme [WARN] Theme: malformed hexcode: purple
```

# Theme `papercolor-light`

```
helix_view::theme [WARN] Theme: malformed hexcode: indent
```

Note: Logging the affected theme's name could be helpful! _(Edit: But looking at the code, it seems unwieldy to carry the theme name around as much as would be necessary to log it in case of parsing errors.)_

---

I assume this should be checked in CI via `xtask`.

sarsapar1lla:
Does this issue still require attention? If so, happy to take a look at it

PR #6019, opened 2023-02-16
Add themecheck to validate color codes in themes, ignore current known errors.

askreet:
Looking for some feedback on approach here while thinking about #5709.

This patch:
* Adds a first-class concept of warnings to theme parsing, which currently catches all the hexcode parse errors from #5709.
* Updates `themelint` with a known issues list, so it can be enabled in CI and prevent new errors.

If this is interesting to folks the next steps would be:
* Adding it to CI.
* Plumbing the warnings from other bits of `theme.rs` up through `Loader::load`, too, and adding any known issues.

Notes:
* I moved the "//TODO: alert the user..." comment up to a higher level in the code as I think that goal relates more to sending an event to the UI, which I haven't looked into.


the-mikedavis:
A check in the CI that fails when there are missing hexcodes would be a good addition but we don't want to add themelint to the CI. (Actually we might remove it in the long-run https://github.com/helix-editor/helix/pull/5534#discussion_r1098511929.) If the palette check can be separated from the existing themelint xtask I think this would be a good change 👍

askreet:
Great context, thanks @the-mikedavis. I'll separate these checks out.

askreet:
Pushed up some changes to isolate a new check called `themecheck` that has been added to CI. Open to suggestions on the naming/scoping of the checks. @the-mikedavis

askreet:
@archseer Can I get a quick 👍🏼  / 👎🏼 on whether this is interesting for the project? No pressure on merge timeline, but if you give me a 👍🏼 I will keep it up to date with master periodically, otherwise I'll close it out.

Issue #5272, opened 2022-12-23
Unreachable Code on Multiple Select


creikey:
### Summary

`thread 'main' panicked at 'internal error: entered unreachable code: (Some(Retain(18446744073709550874)))', helix-core\src\transaction.rs:157:46`

### Reproduction Steps

`  -I"thirdparty" -I"thirdparty\minilzo" -I"thirdparty\enet\include" -I"thirdparty\Chipmunk2D\include\chipmunk" -I"thirdparty\Chipmunk2D\include" -I"thirdparty\opus\include" -I"thirdparty\opus\src"^`

With this line in a file, I selected it with `x` then selected all spaces with `s<space><enter>`, moved the cursor right once, then moved into insert mode, then pressed backspace. Then I pressed enter, which causes this crash reliably.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-12-23T11:59:43.065 helix_lsp::transport [ERROR] err <- "I[11:59:43.065] <-- initialize(0)\r\n"
2022-12-23T11:59:43.067 helix_lsp::transport [ERROR] err <- "I[11:59:43.067] --> reply:initialize(0) 1 ms\r\n"
2022-12-23T11:59:43.068 helix_lsp::transport [ERROR] err <- "I[11:59:43.068] <-- initialized\r\n"
2022-12-23T11:59:43.093 helix_lsp::transport [ERROR] err <- "I[11:59:43.093] <-- textDocument/didOpen\r\n"
2022-12-23T11:59:43.097 helix_lsp::transport [ERROR] err <- "I[11:59:43.097] Loaded compilation database from C:\\Users\\Cameron\\Documents\\flight\\compile_commands.json\r\n"
2022-12-23T11:59:43.097 helix_lsp::transport [ERROR] err <- "I[11:59:43.097] ASTWorker building file C:\\Users\\Cameron\\Documents\\flight\\main.c version 0 with command \r\n"
2022-12-23T11:59:43.097 helix_lsp::transport [ERROR] err <- "[C:/Users/Cameron/Documents/flight/]\r\n"
[...]

PR #6024, opened 2023-02-16
Fix crash when deleting with multiple cursors

### LINKED ISSUE #6021: "char range out of bounds" panic involving `select_regex` and `change_selection`
### Summary

i don't know how to summarize this beyond the steps for reproduction. the same thing happens every time.

### Reproduction Steps

 

I tried this:

1. `hx`
2. key sequence: `ix<space>x<esc>xsx<enter>c<space><backspace>`

I expected this to happen:

the two spaces that i created by pressing space with two cursors would be erased upon pressing backspace, leaving a line with three space characters.

Instead, this happened:

helix panicked with the following message:
```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Char range out of bounds: char range 18446744073709551615..2, Rope/RopeSlice char length 3', /build/.cargo/registry/src/github.com-1ecc6299db9ec823/ropey-1.5.1-alpha/src/rope.rs:546:37
```

### Helix log

```
2023-02-16T10:48:18.232 helix_loader [DEBUG] Located configuration folders: []
2023-02-16T10:48:18.245 helix_view::clipboard [INFO] Using xclip to interact with the system and selection (primary) clipboard
2023-02-16T10:48:18.245 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-02-16T10:48:18.646 helix_term::application [DEBUG] received editor event: IdleTimer
2023-02-16T10:48:19.767 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-02-16T10:48:19.951 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-02-16T10:48:20.020 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-02-16T10:48:20.118 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-02-16T10:48:20.357 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:20.358 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:20.591 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:20.806 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:21.058 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:21.208 helix_term::application [DEBUG] received editor event: IdleTimer
2023-02-16T10:48:21.675 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:22.052 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:22.454 helix_term::application [DEBUG] received editor event: IdleTimer
2023-02-16T10:48:22.454 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
2023-02-16T10:48:22.797 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 1
```

</details>

### Platform

arch linux

### Terminal Emulator

kitty 0.26.5

### Helix Version

helix 22.12

pascalkuthe:
Fixes #6021
Fixes https://github.com/helix-editor/helix/issues/5272
Fixes https://github.com/helix-editor/helix/issues/6029

The `delete_char_backward` command delete indentation when at the start of the line. The problem is that this can cause multiple overlaing deletions with multiple cursors. For example in #6021 one of the cursors is after the 3. space and one after the 1. char. Because the indent unit is 4 spaces `delete_char_backward` deletes to the start of the line for both cursors. The generated transaction is therefore invalid which causes an integer underflow when the transaction is applied.

I fixed a similar crash in #5728. Mutlicursors potentially generating overlapping changes is a common pitfall and something we should be wary of. In that PR I simply discarded changes that overlapped a previous change. That seemed the right behaviour to me for the **replacements** created by completions. For **deletions** however we can simply merge the overlapping deletions together (otherwise the behaviour here would be incorrect). To that end I added a new `delete` and `delete_by_selection` function which does not accept a tendril for replacement and allows overlapping ranges.

In the process I noticed that the `delete_by_selection_insert_mode` function was implemented a bit weirdly:
* It requires a selection rather then a closure like `Transaction::change_by_selection` which requires cloneing the selection
* Calles generating transaction causes some weirdlooking code
* It doesn't automatically invoke `lsp::signature_help_impl(cx, SignatureHelpInvoked::Automatic);` eventough all callsites do (wihout doing that its barely wort moving the logic to a shared function)

Therefore I cleaned up that function a little in a second commit as I was touching that code anyways (and there were no open PRs that touch that code).
If we don't want these cosmetic changes I can easily drop that second commit.

### DIFF
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -5,6 +5,7 @@ use std::borrow::Cow;
 
 /// (from, to, replacement)
 pub type Change = (usize, usize, Option<Tendril>);
+pub type Deletion = (usize, usize);
 
 // TODO: pub(crate)
 #[derive(Debug, Clone, PartialEq, Eq)]
@@ -534,6 +535,41 @@ impl Transaction {
         Self::from(changeset)
     }
 
+    /// Generate a transaction from a set of potentially overlapping deletions
+    /// by merging overlapping deletions together.
+    pub fn delete<I>(doc: &Rope, deletions: I) -> Self
+    where
+        I: Iterator<Item = Deletion>,
+    {
+        let len = doc.len_chars();
+
+        let (lower, upper) = deletions.size_hint();
+        let size = upper.unwrap_or(lower);
+        let mut changeset = ChangeSet::with_capacity(2 * size + 1); // rough estimate
+
+        let mut last = 0;
+        for (mut from, to) in deletions {
+            if last > to {
+                continue;
+            }
+            if last > from {
+                from = last
+            }
+            debug_assert!(
+                from <= to,
+                "Edit end must end before it starts (should {from} <= {to})"
+            );
+            // Retain from last "to" to current "from"
+            changeset.retain(from - last);
+            changeset.delete(to - from);
+            last = to;
+        }
+
+        changeset.retain(len - last);
+
+        Self::from(changeset)
+    }
+
     /// Generate a transaction with a change per selection range.
     pub fn change_by_selection<F>(doc: &Rope, selection: &Selection, f: F) -> Self
     where
@@ -578,6 +614,16 @@ impl Transaction {
             transaction,
             Selection::new(ranges, new_primary_idx.unwrap_or(0)),
         )
+    }
+
+    /// Generate a transaction with a deletion per selection range.
+    /// Compared to using `change_by_selection` directly these ranges may overlap.
+    /// In that case they are merged
+    pub fn delete_by_selection<F>(doc: &Rope, selection: &Selection, f: F) -> Self
+    where
+        F: FnMut(&Range) -> Deletion,
+    {
+        Self::delete(doc, selection.iter().map(f))
     }
 
     /// Insert text at each selection head.
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -570,6 +570,11 @@ impl Transaction {
         Self::from(changeset)
     }
 
+    pub fn insert_at_eof(mut self, text: Tendril) -> Transaction {
+        self.changes.insert(text);
+        self
+    }
+
     /// Generate a transaction with a change per selection range.
     pub fn change_by_selection<F>(doc: &Rope, selection: &Selection, f: F) -> Self
     where
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -788,42 +788,50 @@ fn extend_to_line_start(cx: &mut Context) {
 }
 
 fn kill_to_line_start(cx: &mut Context) {
-    delete_by_selection_insert_mode(cx, move |text, range| {
-        let line = range.cursor_line(text);
-        let first_char = text.line_to_char(line);
-        let anchor = range.cursor(text);
-        let head = if anchor == first_char && line != 0 {
-            // select until previous line
-            line_end_char_index(&text, line - 1)
-        } else if let Some(pos) = find_first_non_whitespace_char(text.line(line)) {
-            if first_char + pos < anchor {
-                // select until first non-blank in line if cursor is after it
-                first_char + pos
+    delete_by_selection_insert_mode(
+        cx,
+        move |text, range| {
+            let line = range.cursor_line(text);
+            let first_char = text.line_to_char(line);
+            let anchor = range.cursor(text);
+            let head = if anchor == first_char && line != 0 {
+                // select until previous line
+                line_end_char_index(&text, line - 1)
+            } else if let Some(pos) = find_first_non_whitespace_char(text.line(line)) {
+                if first_char + pos < anchor {
+                    // select until first non-blank in line if cursor is after it
+                    first_char + pos
+                } else {
+                    // select until start of line
+                    first_char
+                }
             } else {
                 // select until start of line
                 first_char
-            }
-        } else {
-            // select until start of line
-            first_char
-        };
-        (head, anchor)
-    });
+            };
+            (head, anchor)
+        },
+        Direction::Backward,
+    );
 }
 
 fn kill_to_line_end(cx: &mut Context) {
-    delete_by_selection_insert_mode(cx, |text, range| {
-        let line = range.cursor_line(text);
-        let line_end_pos = line_end_char_index(&text, line);
-        let pos = range.cursor(text);
+    delete_by_selection_insert_mode(
+        cx,
+        |text, range| {
+            let line = range.cursor_line(text);
+            let line_end_pos = line_end_char_index(&text, line);
+            let pos = range.cursor(text);
 
-        // if the cursor is on the newline char delete that
-        if pos == line_end_pos {
-            (pos, text.line_to_char(line + 1))
-        } else {
-            (pos, line_end_pos)
-        }
-    });
+            // if the cursor is on the newline char delete that
+            if pos == line_end_pos {
+                (pos, text.line_to_char(line + 1))
+            } else {
+                (pos, line_end_pos)
+            }
+        },
+        Direction::Forward,
+    );
 }
 
 fn goto_first_nonwhitespace(cx: &mut Context) {
@@ -2315,13 +2323,44 @@ fn delete_selection_impl(cx: &mut Context, op: Operation) {
 fn delete_by_selection_insert_mode(
     cx: &mut Context,
     mut f: impl FnMut(RopeSlice, &Range) -> Deletion,
+    direction: Direction,
 ) {
     let (view, doc) = current!(cx.editor);
     let text = doc.text().slice(..);
-    let transaction =
+    let mut selection = SmallVec::new();
+    let mut insert_newline = false;
+    let text_len = text.len_chars();
+    let mut transaction =
         Transaction::delete_by_selection(doc.text(), doc.selection(view.id), |range| {
-            f(text, range)
+            let (start, end) = f(text, range);
+            if direction == Direction::Forward {
+                let mut range = *range;
+                if range.head > range.anchor {
+                    insert_newline |= end == text_len;
+                    // move the cursor to the right so that the selection
+                    // doesn't shrink when deleting forward (so the text appears to
+                    // move to  left)
+                    // += 1 is enough here as the range is normalized to grapheme boundaries
+                    // later anyway
+                    range.head += 1;
+                }
+                selection.push(range);
+            }
+            (start, end)
         });
+
+    // in case we delete the last character and the cursor would be moved to the EOF char
+    // insert a newline, just like when entering append mode
+    if insert_newline {
+        transaction = transaction.insert_at_eof(doc.line_ending.as_str().into());
+    }
+
+    if direction == Direction::Forward {
+        doc.set_selection(
+            view.id,
+            Selection::new(selection, doc.selection(view.id).primary_index()),
+        );
+    }
     doc.apply(&transaction, view.id);
     lsp::signature_help_impl(cx, SignatureHelpInvoked::Automatic);
 }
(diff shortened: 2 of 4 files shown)


gabydd:
thanks for fixing this Pascal, it also fixes https://github.com/helix-editor/helix/issues/5272

pascalkuthe:
> thanks for fixing this Pascal, it also fixes #5272

Thanks for confirming that :+1: That issue was on my todo too so I get two for one here :smile:

pascalkuthe:
I added a fix for #6029 here as the refactor in [3edcc5d](https://github.com/helix-editor/helix/pull/6024/commits/3edcc5dc3df58aeca1d760813a5052c7870e60ac) made that easier to implement.
I added a detailed explanation in the commit message of [f0f2dae](https://github.com/helix-editor/helix/pull/6024/commits/f0f2daef76bcf89e34acd035948817601bc739e5).

The TLDR is:

Deleting forward in append mode used to delete one char forward and then delete backwards, because the selection was shrunk. I have changed the command to always delete forward. The following example of pressing `<del>` repeatedly in append mode explains it well I think.

What I expect to happen (and how it behaves with this PR):
```
|abcd|ef
|abce|f
|abcf|
|abc |
```

This is inline with how other editors like kakoune work.
However, helix currently moves the selection backwards leading to the
following behavior:

```
|abcd|ef
|abc|ef
|ab|ef
|e|f
```

helix-core/src/transaction.rs
```diff
@@ -534,6 +535,46 @@ impl Transaction {
         Self::from(changeset)
     }
 
+    /// Generate a transaction from a set of potentially overallping deletions
```
the-mikedavis, line 538:
```suggestion
    /// Generate a transaction from a set of potentially overlapping deletions
```

helix-core/src/transaction.rs
```diff
@@ -534,6 +535,46 @@ impl Transaction {
...
+    /// by merging overlapping deletions together.
+    pub fn delete<I>(doc: &Rope, changes: I) -> Self
```
the-mikedavis, line 540: very minor nit but I think this would be clearer as `deletions` rather than `changes`

helix-term/src/commands.rs
```diff
@@ -2325,11 +2320,49 @@ fn delete_selection_impl(cx: &mut Context, op: Operation) {
...
+            let (start, end) = f(text, range);
+            if direction == Direction::Forward {
+                let mut range = *range;
+                if range.head > range.anchor {
+                    insert_newline |= end == text_len;
+                    // move the cursor to the right so that the selection
+                    // doesn't shrink when deleting forward (so the text appears to
+                    // move the the left)
+                    // += 1 is enoug here as the range is normalized to grapheme boundaries
+                    // later anyway
```
the-mikedavis, line 2344:
```suggestion
                    // move to the left)
                    // += 1 is enough here as the range is normalized to grapheme boundaries
                    // later anyway
```

the-mikedavis (commented):
Just a few minor comments, otherwise I think this looks good

pascalkuthe:
(force pushed to keep the history intact since I did put some effort into commit messages here :D)

helix-core/src/transaction.rs
pascalkuthe: I have fixed that in the force push

dead10ck (commented):
Ok the changes themselves look good to me. I would humbly request regression tests to codify expected behavior, but not sure if you or others feel this is worth blocking on.

I spent a little time thinking over the larger API—it doesn't quite feel right to fix the function that generates invalid transactions by adding another function that doesn't and changing specific call sites to that instead. But looking over the `change` function, I can see why we can't just move the `from` marker forward in the case of overlap, since this could just end up inserting text in the wrong place.

My first instinct was to suggest that we either

1. Just keep adding and deleting text separate, or
2. Fix this issue in `change` and just make `delete` a thin wrapper around it

to fix even the possibility of this happening through the API. But 1 would make cases like `R` very awkward, and 2 I can't think of a reasonable way to handle overlapping ranges when it handles both deletions and text replacement, if it's even possible, without making it super complex.

So at the end of the day, having another function that relaxes the overlapping ranges invariant for text deletions gets the job done.

pascalkuthe:
@dead10ck I add some integration tests that was a great suggestions. We should really make more use of the integration test framework it works really great. I was pretty much just able to use the reproduction cases straight from the issues.

Regarding the API I actually think having a second function makes sense. For normal edits (replacements) there is no unambiguous way to merge overlapping edits. Instead called must deal with that (or we panic).  There is no correct way to handle this generally in `change`. Even separating the deleting and adding doesn't really fix the problem. The callers would map multiple positions to the same location in that case but there is still no unambiguous way to now add that new text (so again we would have just turned an explicit panic into weirdly/silently incorrect changes).

I spent a lot of time thinking about a better API here when I created this PR but I came to the conclusion that there really is none. Overlapping edits are fundamentally incorrect and must be prevented in the caller. The only exceptions are:

* deletions (where you can just merge the ranges as done here)
* "foreign" edits (like completions) over which we have no control. There we simply use the first range and discard all subsequent overlapping ranges (a decent fallback strategy but this doesn't work for deletions and some other cases)

When I created this PR I actually manually went trough all callsites to the `change`/`change_by_selection` and manually checked that this can't be a problem there. As far as I found all callsites were correct (in most cases edit ranges are just completions in which case overlapping edits are impossible and in the cases where they were possible the functions had explicit checks to deal with that). Only deletions were buggy.

helix-core/src/transaction.rs
```diff
@@ -534,6 +535,41 @@ impl Transaction {
...
+        let mut last = 0;
+        for (mut from, to) in deletions {
+            if last > to {
+                continue;
+            }
+            if last > from {
+                from = last
+            }
+            debug_assert!(
+                from <= to,
```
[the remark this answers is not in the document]
pascalkuthe: this is actually a copy of an existing assertion from the `change` function (the function in general is almost fully copied). If we change this here we should change it there too. I guess the point of this is to basically get a better panic message in debug builds (and to just let it underflow in release builds)

helix-term/tests/test/commands.rs
```diff
@@ -385,3 +385,70 @@ async fn test_character_info() -> anyhow::Result<()> {
 
     Ok(())
 }
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_delete_char_backward() -> anyhow::Result<()> {
+    // don't panic when deleting overlapping ranges
+    test((
+        platform_line("#(x|)# #[x|]#").as_str(),
```
dead10ck, line 393: The `as_str` actually shouldn't be necessary any more after 5b07ce7

helix-term/tests/test/commands.rs
```diff
@@ -385,3 +385,70 @@ async fn test_character_info() -> anyhow::Result<()> {
...
+        "c<space><backspace><esc>",
+        platform_line("#[\n|]#").as_str(),
+    ))
+    .await?;
+    test((
+        platform_line("#( |)##( |)#a#( |)#axx#[x|]#a").as_str(),
+        "li<backspace><esc>",
+        platform_line("#(a|)##(|a)#xx#[|a]#").as_str(),
```
dead10ck, line 401: Why is the first range forward while the rest are backward? How does this test pass? Is there a bug here where the order of the ranges isn't being preserved?

helix-term/src/commands.rs
```diff
@@ -2325,11 +2320,49 @@ fn delete_selection_impl(cx: &mut Context, op: Operation) {
...
+    // in case we delete the last character and the cursor would be moved to the EOF char
+    // insert a newline, just like when entering append mode
+    if insert_newline {
+        transaction = transaction.insert_at_eof(doc.line_ending.as_str().into());
+    }
+
+    if direction == Direction::Forward {
+        doc.set_selection(
+            view.id,
+            Selection::new(selection, doc.selection(view.id).primary_index()),
```
dead10ck, line 2361: If ranges can potentially get merged, can we always use the original selection's primary index safely?

helix-term/tests/test/commands.rs
```diff
@@ -385,3 +385,70 @@ async fn test_character_info() -> anyhow::Result<()> {
...
+    ))
+    .await?;
+    Ok(())
+}
+
+#[tokio::test(flavor = "multi_thread")]
+async fn test_delete_word_forward() -> anyhow::Result<()> {
+    // don't panic when deleting overlapping ranges
+    test((
+        platform_line("fo#[o|]#b#(|ar)#").as_str(),
```
dead10ck, line 424: We have the first range forward and the second backward here

helix-term/tests/test/commands.rs
```diff
@@ -385,3 +385,70 @@ async fn test_character_info() -> anyhow::Result<()> {
...
+                #(abc|)#ef
+                #(abc|)#f
+                #(abc|)#
+            "})
+        .as_str(),
+        "a<del><esc>",
+        platform_line(indoc! {"\
+                #[abc|]#ef
+                #(abc|)#f
+                #(abc|)#
```
dead10ck, line 446: Why does `f` get deleted here? I would expect del to delete the character to the right of the selection. After entering append mode, the cursor should be over `f`, so I would expect the LF to get deleted.

dead10ck (changes requested):
Thanks a lot for adding tests! This will help a lot in the future for anyone who makes changes to deleting behavior.

Also, I agree with your points about the API and adding a function. It does make sense to have both.

helix-term/tests/test/commands.rs
pascalkuthe: Del deletes the char under the cursor actually (also on master). It's the same for kakune iirc.

helix-term/src/commands.rs
pascalkuthe: Its save here because we just delete the selections exactly (start/end is always just the start/end of the selection range). That means that this particular function can actually never have overlapping deletion ranges (since selections are always mon-overlapping).

helix-term/tests/test/commands.rs
pascalkuthe: This is just how range mapping in helix works. We basically delete all chars inside the selection which yields an empty selection. Helix normalizes this to a single char selection but during that normalization the selection direction is lost. For single char selections the direction doesn't really matter tough so it's not usually noticabke in the ui.

This has nothing to do withthese specific commands tough as they dont actually touch the selection at all, it's just how the automatic selection mapping works.

helix-term/tests/test/commands.rs
pascalkuthe: `i` makes both ranges forward. I can change this to make this less awkward looking but it doesn't actually affect the behaviour of the test (since all text get deleted helix always places a forward cursor on the newline regardless if what the selection was previously l

helix-term/tests/test/commands.rs
pascalkuthe: I took the chance to fix this for all tests in that file and therefore made this a separate commit [d605a34](https://github.com/helix-editor/helix/pull/6024/commits/d605a341073325b7f406db524a246e756498c51a)

helix-term/tests/test/commands.rs
dead10ck: Interesting, ok, thanks for the explanation

helix-term/tests/test/commands.rs
dead10ck: Yeah you're right, this isn't a big deal

helix-term/src/commands.rs
pascalkuthe: Ah my bad aplogoues I looked at the wrong code somehow. This is a trick. Here the selection actually is set manually. This works because we call `set_selection` *before* the transaction is applied and if you look at the code we unconditionally call push so the number of ranges doesn't change and the primary index remains valid. 

We only update the selection this way (with head += 1) so that the semantics for `delete_X_forward` work as intended. Just the normally mapping the selection trough the transaction. When you call set_selection I'd normalized si if the selections overlap (because of the head += 1) that's automatically taken care of (and the primary index is then adjusted automatically). 

Furthermore when we apply the transaction below the new selection is still mapped trough the change set so only then becomes then do "overlapping deletions" become a problem. However as long as the selection is valid here (which it is) then that mapping will work too.

dead10ck: Ahhh, I get it now, so all the overlapping selections are still actually there at the end of the function, and they get removed during the transaction application. Thanks for bearing with me.

dead10ck (approved):
Ok, this looks good to me! Thanks for adding tests and bearing with me. This will help when I get to auto pair deletions.

the-mikedavis (approved):

Issue #5903, opened 2023-02-10
Match around (`ma`) inconsistent for quotes vs. brackets.


askreet:
### Summary

This may relate to #5561. I'm not 100% sure this is a bug or intended behavior. If it's a bug, I'm willing to take a crack at fixing it. The `ma'` and `ma"` commands do not select a quoted region if you are on the quote, in contrast to bracket pairs.

### Reproduction Steps

 

File:
```
a = "string"
b = 'string'
c = (list)
d = [array]
```

I tried this:

1. Place the cursor on the first double quote on line 1, press `ma"`. No selection change occurs.
2. Place the cursor on the first single quote on line 2, press `ma'`. No selection change occurs.
3. Place the cursor on `(` on line 3, press `ma(`. The entire parenthesized section is selected.
4. Place the cursor on `[` on line 4, press `ma[`. The entire section enclosed in square brackets is selected.

I expected this to happen:

For the first two examples, I'd expect the quoted regions to be selected.

### Helix log

Nothing interesting in my log file.

### Platform

macOS

### Terminal Emulator

iTerm2

### Helix Version

helix 22.12 (96ff64a8)

the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/issues/4475

askreet:
@the-mikedavis Is this a duplicate? Would love to see quotes also work for `mam` and `mim`, too, but this bug report is about `ma'` with a matching pair character.

the-mikedavis:
Oh I see, yeah this is different 👍

askreet:
Starting to dig into this, noting the following bit of code:

```rust
        if Some(open) == text.get_char(pos) {
            // Cursor is directly on match char. We return no match
            // because there's no way to know which side of the char
            // we should be searching on.
            return Err(Error::CursorOnAmbiguousPair);
        }
```

Vim seems to handle this case by only working linewise and analyzing the entire line, but only for pairs with the same character on both sides. I'm not sure whether that's the appropriate solution for Helix.

wmstack:
I don't especially admire the fact that `mim` and `mam` if on the opening brackets seem to treat the selection of the parent element. One would think that being on the opening bracket and doing `mim` or `mam` would select inside around that same bracket instead of the one around it?

What do you think?

PR #6027, opened 2023-02-17
Add test cases for existing pair matching logic.

askreet:
While doing some research on #5903 I expanded the tests of the underlying logic to understand it a bit better. Still noodling on a solution but figured these were independently worth reviewing and landing.


archseer (approved):
Thanks!

PR #6036, opened 2023-02-18
Document the file-modification-indicator statusline element

theteachr:
Document the `file-modification-indicator` status line element

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -98,6 +98,7 @@ The following statusline elements can be configured:
 | `spinner` | A progress spinner indicating LSP activity |
 | `file-name` | The path/name of the opened file |
 | `file-base-name` | The basename of the opened file |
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/statusline.rs
```diff
@@ -426,6 +426,7 @@ where
     write(context, title, None);
 }
 
+// TODO: Make this customizable (will entail an update in the `editor.statusline` section of `book/src/configuration.md`).
```
the-mikedavis, line 429: This should be a new issue or discussion instead of a TODO comment. I don't think this necessarily needs to be configurable. The discussion of that should be separate from this doc update though

theteachr: I updated the commit to just include the documentation update.

the-mikedavis (approved):

the-mikedavis:
Thanks!

Issue #6045, opened 2023-02-18
Expand to word boundary if we are at a single char selection when pressing `*`


magopian:
When pressing `*` to use the current selection as the search pattern, it would be useful to automatically expand to the whole word if there's a single char being selected.

The logic behind this is:
- in vim, pressing `*` automatically searches for the word under the cursor (no need to first select the word)
- it doesn't seem very useful to search on a single char selection (and if it's needed, one could still press `/<the char to search>` to achieve the same result)
- if more than one character is selected, only search for this selection without expanding to the whole word: keep the way it's working actually

This has been discussed on matrix with @pascalkuthe:
- [my question](https://matrix.to/#/!zMuVRxoqjyxyjSEBXc:matrix.org/$FPSOyuDdBVPZ1tI2j2Poe4eSOM7fGfBdeh6sAUqEFh8?via=matrix.org&via=mozilla.org&via=envs.net)
- [Pascal's answer and suggestion](https://matrix.to/#/!zMuVRxoqjyxyjSEBXc:matrix.org/$6eDXEKGCZoJLn_HqRr9-DadgbzhuWmNHUS6DqJSewCk?via=matrix.org&via=mozilla.org&via=envs.net)

I'm working on a PR for this proposition ;)

archseer:
You could remap this inside your user config to do `miw` before `*`

the-mikedavis:
`miw` would need macro keybinds (#1383) but you can emulate it for the most part with `be`:

```toml
[keys.normal]
"*" = ["move_prev_word_start", "move_next_word_end", "search_selection"]
```

I don't want to see this as built-in behavior for `*`: automatic changes to the selection make it hard to reason about the behavior of these commands especially when you have multiple selections. I think it fights with the select-then-act paradgim.

pascalkuthe:
> `miw` would need macro keybinds (#1383) but you can emulate it for the most part with `be`:
> 
> ```toml
> [keys.normal]
> "*" = ["move_prev_word_start", "move_next_word_end", "search_selection"]
> ```
> 
> I don't want to see this as built-in behavior for `*`: automatic changes to the selection make it hard to reason about the behavior of these commands especially when you have multiple selections. I think it fights with the select-then-act paradgim.

The description and issue title are misleading.

I suggested that `*` could select the entire word if:
* there is only a single selection
* that single selection is only a single word long

I only sort of suggested this because `gf` already does this and a single char search is comparatively useless.
[...]

PR #6046, opened 2023-02-18
Expand selection to whole word when pressing `*`

magopian:
Refs #6045 

This is an attempt at adding the functionality discussed in [this issue](#6045) (and earlier on matrix).

I'm open to suggestions, I've barely learned Rust, and it was a long time ago and seldom used it, so i'm very... rusty.

Please let me know if this is useful, relevant, and how it could be improved!

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1906,18 +1906,37 @@ fn extend_search_prev(cx: &mut Context) {
 }
 
 fn search_selection(cx: &mut Context) {
+    let count = cx.count();
     let (view, doc) = current!(cx.editor);
     let contents = doc.text().slice(..);
 
-    let regex = doc
-        .selection(view.id)
-        .iter()
-        .map(|selection| regex::escape(&selection.fragment(contents)))
-        .collect::<HashSet<_>>() // Collect into hashset to deduplicate identical regexes
-        .into_iter()
-        .collect::<Vec<_>>()
-        .join("|");
+    let regex; // The fragment to set in the search register
 
+    // Checks whether there is only one selection with a width of 1
+    let selections = doc.selection(view.id);
+    let primary = selections.primary();
+    if selections.len() == 1 && primary.len() == 1 {
+        let text = doc.text();
+        let text_slice = text.slice(..);
+        // In this case select the WORD under the cursor
+        let current_word = textobject::textobject_word(
+            text_slice,
+            primary,
+            textobject::TextObject::Inside,
+            count,
+            false,
+        );
+        let text_to_search = current_word.fragment(text_slice).to_string();
+        regex = regex::escape(&text_to_search);
+    } else {
+        regex = selections
+            .iter()
+            .map(|selection| regex::escape(&selection.fragment(contents)))
+            .collect::<HashSet<_>>() // Collect into hashset to deduplicate identical regexes
+            .into_iter()
+            .collect::<Vec<_>>()
+            .join("|");
+    }
     let msg = format!("register '{}' set to '{}'", '/', &regex);
     cx.editor.registers.push('/', regex);
     cx.editor.set_status(msg);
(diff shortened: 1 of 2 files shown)


pascalkuthe (changes requested):
I like this change and the implementation looks reasonable to me too as its just lifted from `goto_file` :+1: 

Searching for a single char with `*` seems like a nieche use-case: `*n` it not much shorter than `/[char]<ret>` or `f[char]`. 
Automatically using the word under the cursor is consistent with `gf` and can be super useful.

However this behavior change should be documented in `keymap.md` . 
Currently it says:
> Use current selection as the search pattern 

Wit this PR something like the following would help:

> Use current selection as the search pattern, if only a single char is selected the current word (`miW`) is used instead

magopian:
Thanks for the feedback, I've updated the `keymap.md` file accordingly 👍

pascalkuthe (approved):
LGTM now :+1:

firstrow:
nice improvement. going to test it locally. currently was using this config: 
```
"*" = ["move_prev_word_start", "move_next_word_end", "search_selection", "global_search"]
```

helix-term/src/commands.rs
```diff
@@ -1906,18 +1906,35 @@ fn extend_search_prev(cx: &mut Context) {
...
+    let regex = if selections.len() == 1 && primary.len() == 1 {
+        let text = doc.text();
+        let text_slice = text.slice(..);
+        // In this case select the WORD under the cursor
+        let current_word = textobject::textobject_word(
+            text_slice,
+            primary,
+            textobject::TextObject::Inside,
+            count,
+            false,
```
the-mikedavis, line 2102: Passing `false` here makes this look for the surrounding "word" rather than the "WORD". I think that's an ok behavior but the change in `book/src/keymap.md` should mention `miw` instead of `miW`. Also see https://github.com/helix-editor/helix/blob/d570c29ce37ffbb46a9c49708c31dfd81daa27cf/helix-term/src/commands.rs#L5000-L5001

helix-term/tests/test/commands.rs
```diff
@@ -258,6 +258,56 @@ async fn test_goto_file_impl() -> anyhow::Result<()> {
...
+    )
+    .await?;
+
+    // Multiple selection of length 1 each : should still only search for the selection
+    test_key_sequence(
+        &mut helpers::AppBuilder::new().build()?,
+        Some("ifoobar::baz<ret>bar::crux<esc>k3blC*"), // k3b places the cursor on the first letter of 'foobar', then move one to the right for good measure, then adds a cursor on the line below
+        Some(&|app| {
+            assert!(
+                // The selections don't seem to be ordered, so we have to test for the two possible orders.
```
the-mikedavis, line 195: I believe what's happening here is that collecting into a hashmap might change the order. The selections do have a deterministic order

the-mikedavis (changes requested):
I was initially opposed to this behavior (along these lines: https://github.com/helix-editor/helix/issues/6045#issuecomment-1471606218) but with the special registers from https://github.com/helix-editor/helix/pull/6985 you can search for the current selection exactly with `/<C-r>.`. So I think this would be a nice improvement. I left a few small comments and pushed up a merge with master since there were some changes from the special registers PR that conflicted

the-mikedavis:
I believe Kakoune also adds `\b` in some cases with `*`. I think it would be a nice improvement to move our `*` closer to Kakoune's behavior

NikitaRevenco:
If there's anything that needs to be done to merge this, I'm happy to help!

andreytkachenko:
Suggesting add new marker `disposition-merge` to clearly let everyone (including the Author) that the PR wanted to be merged and just needs issues resolution. Kinda hard to understand what is blocking this PR. I suspect Author also would be much eager in resolution issues if know that that work will be merged eventually.

PR #6059, opened 2023-02-19
Feat: CodeLens-style diagnostics

### LINKED ISSUE #3272: Error lens style LSP diagnostics
I think it would be good to have the option to show LSP warnings/errors on the same line that they occur without the user hovering over the part that causes the issue, similar to the [vscode extension](https://github.com/phindle/error-lens).

poliorcetics:
Closes #3272

### DIFF
--- a/helix-term/src/ui/editor/diagnostics_annotations.rs
+++ b/helix-term/src/ui/editor/diagnostics_annotations.rs
@@ -1,0 +1,269 @@
+use std::borrow::Cow;
+use std::rc::Rc;
+
+use helix_core::diagnostic::Severity;
+use helix_core::text_annotations::TextAnnotations;
+use helix_core::visual_offset_from_anchor;
+use helix_core::SmallVec;
+use helix_view::graphics::Rect;
+use helix_view::theme::Style;
+use helix_view::{Document, Theme, View};
+
+use crate::ui::document::{LineDecoration, LinePos, TextRenderer};
+
+pub fn inline_diagnostics_decorator(
+    doc: &Document,
+    view: &View,
+    viewport: Rect,
+    theme: &Theme,
+    text_annotations: &TextAnnotations,
+) -> Box<dyn LineDecoration> {
+    let hint = theme.get("hint");
+    let info = theme.get("info");
+    let warning = theme.get("warning");
+    let error = theme.get("error");
+
+    let messages = doc.diagnostic_annotations_messages();
+
+    let text = doc.text().slice(..);
+    let text_fmt = doc.text_format(viewport.width, None);
+
+    let mut visual_offsets = Vec::with_capacity(messages.len());
+    for message in messages.iter() {
+        visual_offsets.push(
+            visual_offset_from_anchor(
+                text,
+                view.offset.anchor,
+                message.anchor_char_idx,
+                &text_fmt,
+                text_annotations,
+                viewport.height as usize,
+            )
+            .map(|x| x.0),
+        );
+    }
+
+    // Compute the Style for a given severity
+    let sev_style = move |sev| match sev {
+        Some(Severity::Error) => error,
+        // The same is done when highlighting gutters so we do it here too to be consistent.
+        Some(Severity::Warning) | None => warning,
+        Some(Severity::Info) => info,
+        Some(Severity::Hint) => hint,
+    };
+
+    // Vectors used when computing the items to display. We declare them here so that they're not deallocated when the
+    // closure is done, only when it is dropped, that way calls are don't have to allocate as much.
+    let mut stack = Vec::new();
+    let mut left = Vec::new();
+    let mut center = SmallVec::<[_; 2]>::new();
+
+    let line_decoration = move |renderer: &mut TextRenderer, pos: LinePos| {
+        let mut first_message_idx = usize::MAX;
+        let mut found_first = false;
+        let mut last_message_idx = usize::MAX;
+
+        for (idx, message) in messages.iter().enumerate() {
+            if message.line == pos.doc_line {
+                if !found_first {
+                    first_message_idx = idx;
+                    found_first = true;
+                }
+                last_message_idx = idx;
+            }
+        }
+
+        // If we found no diagnostic for this position, do nothing.
+        if !found_first {
+            return;
+        }
+
+        // Extract the relevant diagnostics and visual offsets.
+        let messages = match messages.get(first_message_idx..=last_message_idx) {
+            Some(m) => m,
+            None => return,
+        };
+        let visual_offsets = match visual_offsets.get(first_message_idx..=last_message_idx) {
+            Some(v) => v,
+            None => return,
+        };
+
+        // Used to build a stack of diagnostics and items to use when computing `DisplayItem`
+        #[derive(Debug)]
+        enum StackItem {
+            // Insert `n` spaces
+            Space(u16),
+            // Two diagnostics are overlapping in their rendering, we'll need to insert a vertical bar
+            Overlap,
+            // Leave a blank space that needs a style (used when a diagnostic message is empty)
+            Blank(Style),
+            // A diagnostic and its style (computed from its severity)
+            Diagnostic(Rc<String>, Style),
+        }
+
+        // Additional items to display to point the messages to the diagnostic's position in the text
+        #[derive(Debug)]
+        enum DisplayItem {
+            Space(u16),
+            Static(&'static str, Style, u16),
+            String(String, Style, u16),
+        }
+
+        stack.clear();
+        stack.reserve(
+            stack
+                .capacity()
+                .saturating_sub(messages.len().saturating_mul(2)),
+        );
+        let mut prev_col = None;
+        let mut line_count = 0_u16;
+
+        // Attribution: the algorithm to compute the layout of the symbols and columns here has been
+        // originally written by Hugo Osvaldo Barrera, for https://git.sr.ht/~whynothugo/lsp_lines.nvim.
+        // At the time of this comment's writing, the commit used is ec98b45c8280e5ef8c84028d4f38aa447276c002.
+        //
+        // We diverge from the original code in that we don't iterate in reverse since we display at the end of the
+        // loop instead of later, which means we don't have the stack problem that `lsp_lines.nvim` has.
+
+        // First we build the stack, inserting `StackItem`s as needed
+        for (message, visual_offset) in messages.iter().zip(visual_offsets.iter()) {
+            let visual_offset = match visual_offset {
+                Some(o) => *o,
+                None => continue,
+            };
+
+            let style = sev_style(message.severity);
+
+            // First the item to offset the diagnostic's text
+            stack.push(match prev_col {
+                Some(prev_col) if prev_col != visual_offset.col => StackItem::Space(
+                    visual_offset
+                        .col
+                        .abs_diff(prev_col)
+                        // Account for the vertical bars that are inserted to point diagnostics to
+                        // their position in the text
+                        .saturating_sub(1)
+                        .min(u16::MAX as _) as _,
+                ),
+                Some(_) => StackItem::Overlap,
+                None => StackItem::Space(visual_offset.col.min(u16::MAX as _) as _),
+            });
+
+            let trimmed = message.message.trim();
+
+            // Then the diagnostic's text
+            if trimmed.is_empty() {
+                stack.push(StackItem::Blank(style));
+            } else {
+                stack.push(StackItem::Diagnostic(Rc::clone(&message.message), style));
+            }
+
+            prev_col = Some(visual_offset.col);
+            line_count = line_count.saturating_add(trimmed.lines().count().min(u16::MAX as _) as _);
+        }
+
+        // When several diagnostics are present in the same virtual block, we will start by
+        // displaying the last one and go up one at a time
+        let mut pos_y = viewport
+            .y
+            .saturating_add(pos.visual_line)
+            .saturating_add(line_count);
+
+        // Then we iterate the stack we just built to find diagnostics
+        for (idx, item) in stack.iter().enumerate() {
+            let (text, style) = match item {
+                StackItem::Diagnostic(text, style) => (text.trim(), *style),
+                _ => continue,
+            };
+
+            left.clear();
+            let mut overlap = false;
+            let mut multi = 0;
+
+            // Iterate the stack for this line to find elements on the left.
+            let mut peekable = stack[..idx].iter().peekable();
+            while let Some(item2) = peekable.next() {
+                match item2 {
+                    &StackItem::Space(n) if multi == 0 => left.push(DisplayItem::Space(n)),
+                    &StackItem::Space(n) => {
+                        left.push(DisplayItem::String("─".repeat(n as usize), style, n))
+                    }
+                    StackItem::Blank(_) => {
+                        left.push(DisplayItem::Static(
+                            if multi == 0 { "└" } else { "┴" },
+                            style,
+                            1,
+                        ));
+                        multi += 1;
+                    }
+                    StackItem::Diagnostic(_, style) => {
+                        // If an overlap follows this, don't add an extra column.
+                        if !(matches!(peekable.peek(), Some(StackItem::Overlap))) {
+                            left.push(DisplayItem::Static("│", *style, 1));
+                        }
+                        overlap = false;
+                    }
+                    StackItem::Overlap => overlap = true,
+                }
+            }
+
+            let center_symbol = if overlap && multi > 0 {
+                "┼─── "
+            } else if overlap {
+                "├─── "
+            } else if multi > 0 {
+                "┴─── "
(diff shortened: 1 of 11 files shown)


poliorcetics:
This branch should work correctly but I want to refacto and split it into proper commits, I'm just pushing so that other people can try it out

poliorcetics:
image

pascalkuthe:
Thanks for working on this, I am really lookimg forward to.this! Already looks great! I haben't done a full review yet since you aren't done but I have some foundational comments that I think are good to make as early as possible to avoid duplicate work:

The way you compute the exact position where to render a message is not quite correct.

Fundementally due to softwrap and virtual text the on screen position can not ne computed from the text and can change at any time fully async.

Instead the idea is that these decorations are anchored to a char position. When we render we already know the exact onscreen position of each char anyway. The idea is basically that multiline virtual text should only be computed lazily (when visible) in the rendering code. In general the entire idea is that on screen positions are only temporary. Everything that is stored for longer periods is stored as char offsets that get translated to screen coordinates then back. I should write more documentation for that (and maybe even a design doc) but I am sadly busy at the moment.

I actually hadn't thought how to do to handle this specific usecase 100% yet. There is a second API to hook into the rendering: TranslatePosition.These are called when a specific char index is reached with the screen position of that char index. This is a lot closer to what you need. It doesn't quite fit your usecase because you still need to accumelate all positions of the current line and then finally render then (because diagnostics rendering stack right to left not left to right). 

The right way to do this is probably to unify these two APIs so you can have a single struct which accumulates positions and then renders them.

For now I think you could just start by using `helix_core::visual_offset_from_anchor` to move away from doing any rendering in `helix-view/document.rs`. I will send a PR to streamline the API which you can then base ykurs shouldn't be a large PR and it should be quite easy to replace visual_offset_from_anchor with these positions.

BTW. a small code clarity thing: I would prefer if we didn't keep adding too much code to ui/editor.rs. I purposfukly modularized the text rendering to avoid that.

Since `LineDecoration` is a trait you can create a neee file `ui/inlay diagnostics.rs` and create a strucy `struct InlayDiagnstics` which implement `LineAnnotstion` and contains any dats you need. This also allows you to split the logic into subfunctions easier.

poliorcetics:
Screenshot 2023-02-20 at 09 43 37

@pascalkuthe thanks for the tip, I tested with the inlay-hint branch to fix and now it works much better ^^(sill not ready for review, the code is a mess right now, documentation is wrong everywhere too after that fix)

poliorcetics:
This is ready for review but there are still several problems I know of:

The cursor will go out of view when diagnostics are displayed. I think the `LineAnnotation`s are not taken into account somewhere when computing the view.

In the same vein, I was forced to make commit https://github.com/helix-editor/helix/pull/6059/commits/4d3ef39ec8df220dfbf97a38d47a61f88fe75840 because when the top of the file is modified (for example by rust-analyzer inserting a missing import) while out of the view, it would panic. I think I did something wrong but it's late and that fix make it "work" (i.e. not crash) so that's good enough to ask for help and review.


Finally, when updating text out of view, because of the previous fix I think diagnostics are not moved correctly so we end up with things like the screenshot below:

Screenshot 2023-02-21 at 01 07 33

Again, not a crash so I'll take it for now. I would like to see it fixed though

pascalkuthe:
> This is ready for review but there are still several problems I know of:
> 
> The cursor will go out of view when diagnostics are displayed. I think the `LineAnnotation`s are not taken into account somewhere when computing the view.
> 
> In the same vein, I was forced to make commit [4d3ef39](https://github.com/helix-editor/helix/commit/4d3ef39ec8df220dfbf97a38d47a61f88fe75840) because when the top of the file is modified (for example by rust-analyzer inserting a missing import) while out of the view, it would panic. I think I did something wrong but it's late and that fix make it "work" (i.e. not crash) so that's good enough to ask for help and review.
> 
> Finally, when updating text out of view, because of the previous fix I think diagnostics are not moved correctly so we end up with things like the screenshot below (appears after asking rust-analyzer to insert the missing import for `use std::arc::Arc`):
> 
> Screenshot 2023-02-21 at 01 07 33
> 
> Again, not a crash so I'll take it for now. I would like to see it fixed though

After a quick scan I already have some comments but I am actually working on the improve rendering API I mentioned earlier and those comments are related to do that so I will holdoff a review until I am finished. Hopefully tomorrow if nothing comes up. 

I think https://github.com/helix-editor/helix/commit/4d3ef39ec8df220dfbf97a38d47a61f88fe75840 might be a legitimate bug although the cursor going out of view should also not happen. That should ne properly accounted for but I might habe missed something.

Howber the diagnostic mapping should not be affected by any of this so that sounds like a separate bug.

erasin:
When using 'k' At file bottom, the cursor keep at the bottom or hide when file have a lot of diagnostics message.

Whether to update `view.area.height`? I have no idea.

erasin:
Whether to update `view.offset.vertical_offset` ？
The cursor offset has error at bottom.

poliorcetics:
> The cursor offset has error at view bottom. When the file content has multiple pages ,cursor will be keep at view bottom when have multiline diagnostics message and use k moving.

Yes it's a known problem, see my previous message, I don't know where it comes from though

CptPotato:
Two small nits / ideas (might be suited for a follow-up PR):

- Is it possible to color the tree lines in a neutral color? I think it can look a little odd when different diagnostics types are present:
  image

  I had something like this in mind:
  <details>
  <summary>screenshot</summary>
  
  image
  </details>
- What do you think about adding a separate theme key for the code lens background, adding some separation to the code lens diagnostics:
  <details>
  <summary>screenshot</summary>
  
  
  image
  </details>

poliorcetics:
@CptPotato did you test on the latest version ? I fixed a bug that highlighted the lines wrong (it used the diagnostic on the same line instead of the target diagnostic), now it works like this:

Screenshot 2023-02-22 at 16 19 48

> What do you think about adding a separate theme key for the code lens background, adding some separation to the code lens diagnostics:

Yes, that's a good idea, will do

poliorcetics:
@CptPotato  Added a default style under `ui.virtual.diagnostics`, it's not necessary to use `background`: helix will use it if nothing else is set and I can see someone wanting to set `ui.virtual` to have all virtual text with the same style at once.

@pascalkuthe I discovered a bug where I trigger the assertion in `helix-term/src/ui/document.rs`, `assert_eq!(0, offset.vertical_offset);` when using `Z` mode with inline diagnostics. I'm not sure what's happening here but I suppose the text annotations I create are wrong or something else is mis-handled when computing heights

helix-term/src/ui/editor/diagnostics_annotations.rs
```diff
@@ -0,0 +1,282 @@
...
+            // d. Is not an overlap.
+
+            let lines_offset = text.lines().count() as u16;
+            pos_y -= lines_offset;
+
+            // Use `view` since it's the whole outer view instead of just the inner area so that the background
+            // is also applied to the gutters and other elements that are not in the editable part of the document
+            let diag_area = Rect::new(
+                whole_view_aera.x,
+                pos_y + 1,
```
SoraTenshi, line 239: This can go out of bounds, aka. higher than the `viewport.height` which causes a panic.
I have debugged this as i am using this in my daily-driver fork.

a simple fix would be some kind of clamp of `pos_y` (also on L#247 ) to the `viewport.height`

helix-term/src/ui/editor/diagnostics_annotations.rs
```diff
@@ -0,0 +1,282 @@
...
+use crate::ui::document::{LineDecoration, LinePos, TextRenderer};
+
+pub fn inline_diagnostics_decorator(
+    doc: &Document,
+    view: &View,
+    viewport: Rect,
+    theme: &Theme,
+    text_annotations: &TextAnnotations,
+) -> Box<dyn LineDecoration> {
+    let whole_view_aera = view.area;
```
SoraTenshi, line 21:
```suggestion
    let whole_view_area = view.area;
```

helix-term/src/ui/editor/diagnostics_annotations.rs
```diff
@@ -0,0 +1,282 @@
...
+            // c. Is just one line.
+            // d. Is not an overlap.
+
+            let lines_offset = text.lines().count() as u16;
+            pos_y -= lines_offset;
+
+            // Use `view` since it's the whole outer view instead of just the inner area so that the background
+            // is also applied to the gutters and other elements that are not in the editable part of the document
+            let diag_area = Rect::new(
+                whole_view_aera.x,
```
SoraTenshi, line 238:
```suggestion
                whole_view_area.x,
```

helix-term/src/ui/editor/diagnostics_annotations.rs
```diff
@@ -0,0 +1,282 @@
...
+                pos_y + 1,
+                whole_view_aera.width,
```
SoraTenshi, line 240:
```suggestion
                whole_view_area.width,
```

SoraTenshi (changes requested):
Awesome job! i really like the style of the diagnostics!

poliorcetics:
@SoraTenshi thanks for the find, I had planned to search for it but you made it a 100 times easier 😄  it should be fixed now

There is still the bug with the cursor going out of view and the `Z` mode panicking when a diagnostic (or more likely, any line annotation) comes into view

SoraTenshi:
> @SoraTenshi thanks for the find, I had planned to search for it but you made it a 100 times easier 😄  it should be fixed now
> 
> There is still the bug with the cursor going out of view and the `Z` mode panicking when a diagnostic (or more likely, any line annotation) comes into view

Have you figured out how to reproduce it?

erasin:
image

The length of the error message is out of range in the split window.

pascalkuthe:
> image
> 
> The length of the error message is out of range in the split window.

Yeah it needs to be softwrapped but variable height virtual text isn't currently supported by the virtual text API. I don't have much time at the moment but next week I will post a PR to expand the virtual text API to cover this usecase.

I actually have a similar need for side by side diffing so it's needed anyway. This PR also has a bunch of bugs that are partly caused by the virtual text API being limited (and partly by it not being used correctly). Once I have update the virtual text API I will do an indepth review.

poliorcetics:
> image
> 
> The length of the error message is out of range in the split window.

Should be fixed :)

erasin:
> Should be fixed :)

Thanks.

----

Also, the error message is displayed in real time when programming in insert mode, which is very distracting, we need a better way to display the new error message only when going back to normal mode from insert mode.

poliorcetics:
> example only when going back to normal mode from insert mode.

As in #5934, I think the result would be very jarring when going from one mode to another, though for diagnostics it seems more useful than for inlay hints.

It could be an option there too, but I'm wary of introducing options by the dozen

pythoneer:
> > example only when going back to normal mode from insert mode.
> 
> As in #5934, I think the result would be very jarring when going from one mode to another, though for diagnostics it seems more useful than for inlay hints.
> 
> It could be an option there too, but I'm wary of introducing options by the dozen

I can understand the original problem but this should really be an option (if at all) because i feel like that those diagnostics needs immediate attention anyway and the "correct" way to remove them is by fixing them. I think you can make the case that they should be "annoying by design" as you can't "just compile" even if there are errors. If you want to deviate from this we can think about introducing an option to turn that off while editing or having a command to toggle them and people can bind keys to it and turn it off when they feel like it.

lazytanuki:
I agree that it should be an option, but I think it really should be possible.

> I think you can make the case that they should be "annoying by design" as you can't "just compile" even if there are errors.

I have to disagree with this though. Some errors that pop up while typing are non relevant, since you've not finished typing and most of the time you weren't gonna make these mistakes. But with this PR they do make the code below constantly jump up and down, which is what I find quite jarring and disorienting.
That is why an option would be very welcome, either to disable the codelens-style diags in insert mode, or even better, disable them on the current line in insert mode. That is also the way I feel about #5934.

Anyway, thanks for your work on this @poliorcetics, this is a very nice thing to have !

poliorcetics:
I won't implement the options in those PRs 1) to keep them smaller, they're already big enough as is 2) I think just having mapping to quickly enable/disable as someone did in the inlay hints PR is enough for a first-approach.

With more usage we can see about adding them in follow-up PRs though

lazytanuki:
Sounds good!

pascalkuthe:
Just to avoid duplicate work here:

I am doing a medium sized refactor of the virtual text API from the lessons larned from using the API in practice. To test that PR I essentially had to reimplement this PR on top of the new API (the ASCII art is kept but all the infrastructure is unnecessary because the API handles it all). I got lots of things to juggle so it might take a bit. But I just wanted to let you know so you don't invest too much effort into the implementation details here since they will change anyway.

poliorcetics:
I'll be unavailable for a week starting tomorrow so no worries, I won't be making any big changes in this PR any time soon

T4varo:
I have merged your PR into the master locally. The feature is awesome! So much easier to read the LSP Hints like this.

Unfortunately it crashes on some files:
```
thread 'main' panicked at 'index out of bounds: the len is 5332 but the index is 5332', helix-tui/src/buffer.rs:476:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:64:14
             at /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74/library/core/src/panicking.rs:159:5
   3: helix_tui::buffer::Buffer::set_style
   4: <F as helix_term::ui::document::LineDecoration>::render_background
   5: helix_term::ui::document::render_document
   6: helix_term::ui::editor::EditorView::render_view
   7: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
   8: helix_term::application::Application::render::{{closure}}
   9: hx::main_impl::{{closure}}
  12: tokio::runtime::runtime::Runtime::block_on
  13: hx::main
```

This happens on this Rust code when the dependencies are not set in the Cargo.toml:
```rs
use tower_lsp::jsonrpc::Result;
use tower_lsp::lsp_types::*;
use tower_lsp::{Client, LanguageServer, LspService, Server};

#[derive(Debug)]
struct Backend {
    client: Client,
}

#[tower_lsp::async_trait]
impl LanguageServer for Backend {
    async fn initialize(&self, _: InitializeParams) -> Result<InitializeResult> {
        Ok(InitializeResult::default())
    }

    async fn initialized(&self, _: InitializedParams) {
        self.client
            .log_message(MessageType::INFO, "server initialized!")
            .await;
    }

    async fn shutdown(&self) -> Result<()> {
        Ok(())
    }
}

#[tokio::main]
async fn main() {
    let stdin = tokio::io::stdin();
    let stdout = tokio::io::stdout();

    let (service, socket) = LspService::new(|client| Backend { client });
    Server::new(stdin, stdout, socket).serve(service).await;
}
```
demo_crash

Another problem is that the view position is not properly adjusted for the size of virtual lines so that the cursor can go out of the view field.
demo_cursor

cidit:
how soon can we expect to see this? looks awesome

i3elj:
Since some stuff weren't mentioned here I think it would be nice to have them mentioned somewhere. Therefore, I'll leave it here just as a note. And thanks for this awesome PR beforehand.

1. why shift the code instead of showing the diagnostics as a contiguous line, like what happens in neovim, and toggling some pop-up with a keybinding (e.g what `gl` doest in nvim)?
2. I think would be interesting to have an option to toggle it by default. e.g
```toml
[editor.lsp]
code-lens.enabled = true # or false
```
3. having a keybinding to toggle it it's great and have already been mentioned here.
4. this will replace the old style or just add one extra? if so, it will be defined in the `config.toml` file? e.g
```toml
[editor.lsp]
code-lens.style = "default" # or "toggleable" ??
```

Thanks again for this awesome feature!

pascalkuthe:
Superseded by #6417 

That PR contains a bunch of bug fixes and architectural improvements to the virtual text API and a complete reimplementation of this PR on top of those improvements. This should fix the artifacts and crashes reported here.
 
 That branch also contains a bunch of additional improvements:
 * text is softwrapped so all diagnostics are always readable and never cutoff (although the number of visible diagnostics is limited to 10 by default to avoid filling the entire screen with diagnostics, but that limit can be configured)
 * diagnostics can be filtered by their level (warning by default)
 * diagnostics on the line that contains the cursor can be filtered by a separate level filter (shows all diagnostics by default)
 
 It would be great if that PR was tested more to shake out any remaining bugs in the virtual text rendering (although I hopefully got most of them). It's still missing some docs but apart from that it should work well.

nathan-at-least:
I filed #7015 due to a UX problem with 0.6.0 in which brainstormed a different UI. Then I discovered this, which is better than what I imagined while addressing my problem well. Closed my ticket. 👍 and Kudos!

PR #6064, opened 2023-02-20
Added yuck language support (for eww)

Philipp-M:
Adds syntax highlighting support for yuck (DSL for eww). The parser should be complete now and closely follows the original lalrpop parser in eww.

ewwyucktshighlighting

### DIFF
--- a/runtime/queries/yuck/highlights.scm
+++ b/runtime/queries/yuck/highlights.scm
@@ -15,7 +15,7 @@
 ; strings
 (string_interpolation
   (string_interpolation_start) @punctuation.special
-  (string_interpolation_end) @punctuation.special) @embedded
+  (string_interpolation_end) @punctuation.special)
 
 (escape_sequence) @constant.character.escape
 
--- a/runtime/queries/yuck/highlights.scm
+++ b/runtime/queries/yuck/highlights.scm
@@ -1,0 +1,66 @@
+(ERROR) @error
+
+(line_comment) @comment
+
+; keywords and symbols
+
+(keyword) @keyword
+(symbol) @tag
+
+; literals
+
+(bool_literal) @constant.builtin.boolean
+(num_literal) @constant.numeric
+
+; strings
+(string_interpolation
+  (string_interpolation_start) @punctuation.special
+  (string_interpolation_end) @punctuation.special) @embedded
+
+(escape_sequence) @constant.character.escape
+
+(string
+  [
+    (unescaped_single_quote_string_fragment)
+    (unescaped_double_quote_string_fragment)
+    (unescaped_backtick_string_fragment)
+    "\""
+    "'"
+    "`"
+  ]) @string
+
+; operators and general punctuation
+
+(unary_expression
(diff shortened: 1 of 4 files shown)


the-mikedavis:
Can this be covered by an lisp grammar we already include like clojure or scheme instead? Or is the syntax different enough that it needs its own parser?

Philipp-M:
Well one thing that makes it indeed different (and more difficult at that, as it's a context-sensitive grammar), is that `yuck` includes a separate DSL that eww calls "simplexpr" (e.g. `{time.date}` or stuff inside string interpolation (which is a "simplexpr" as well). "simplexpr" seems to be a mix out of simple general expressions like binary/unary expressions, function calls and json data access.

the-mikedavis:
Ah yeah, looks like there is string interpolation too which I don't think is supported in the existing lisps we have

runtime/queries/yuck/highlights.scm
```diff
@@ -0,0 +1,66 @@
...
+
+; literals
+
+(bool_literal) @constant.builtin.boolean
+(num_literal) @constant.numeric
+
+; strings
+(string_interpolation
+  (string_interpolation_start) @punctuation.special
+  (string_interpolation_end) @punctuation.special) @embedded
```
the-mikedavis, line 18: I think neovim uses `@embedded` for something but it shouldn't be necessary in the Helix queries

```suggestion
(string_interpolation
  (string_interpolation_start) @punctuation.special
  (string_interpolation_end) @punctuation.special)
```

runtime/queries/yuck/highlights.scm
```diff
@@ -0,0 +1,66 @@
+(ERROR) @error
+
+(line_comment) @comment
+
+; keywords and symbols
+
+(keyword) @keyword
+(symbol) @tag
```
the-mikedavis, line 8:
```suggestion
(symbol) @string.special.symbol
```

I'm not sure if this is the same kind of symbol though: we currently use this scope for things like Erlang/Elixir atoms, Ruby symbols and Clojure keywords

Philipp-M: Yeah (symbol) highlighting could certainly be improved still, maybe some real keywords for things that are used a lot and are builtin like `defwidget`, `label` or `box`.

Hmm I'm not sure, it seems to be something like a DOM element definition, thus IMO `@tag` makes more sense here?

runtime/queries/yuck/highlights.scm
Philipp-M: I copy/pasted this mostly from the nix highlighting queries, so it may indeed be an artifact that originated from neovim (where the nix highlights are from originally AFAIK).

runtime/queries/yuck/highlights.scm
the-mikedavis: Ah ok, if these are more DOM-like then let's keep `@tag` 👍

the-mikedavis (approved):

PR #5812, opened 2023-02-03
Alignment of the bufferline

vapourismo:
Adds the `bufferline-alignment` configuration item which allows you to choose an alignment (e.g. left, centre, right) for the bufferline contents.

PR #4726, opened 2022-11-12
Limit the amount of vertical splits

janhrastnik:
This would fix #1083.
Right now there's this error happening: https://github.com/helix-editor/helix/pull/2773#issuecomment-1162007131, when you do an appropriate amount of vertical splits.
This is the line that causes the crash: https://github.com/helix-editor/helix/blob/fd585c1ee4ae0ac541f99b5e69a40833f5e3246d/helix-view/src/tree.rs#L408
I think recalculate() here doesn't take into consideration that inner gaps decrease area width, so I changed that.
I also added a function in typed.rs that checks if a child view of the focused view has a width of 1. If it does, then we can't make any more vertical splits.

PR #6067, opened 2023-02-20
(rename) View.doc -> View.doc_id

gibbz00:
## Conflicting PRs (7)
#5812 #6056 #6061 #4642 #4726 #2869 #3239
\* That are mergeable with master and pass all CI checks as of 2023-02-22

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -138,17 +138,17 @@ impl fmt::Debug for View {
 }
 
 impl View {
-    pub fn new(doc: DocumentId, gutters: GutterConfig) -> Self {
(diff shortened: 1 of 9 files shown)


helix-view/src/view.rs
```diff
@@ -141,7 +141,7 @@ impl View {
     pub fn new(doc: DocumentId, gutters: GutterConfig) -> Self {
         Self {
             id: ViewId::default(),
-            doc,
+            doc_id: doc,
```
pickfire, line 144: Can you please change the arg to `doc_id` so we can do `doc_id,`?

dead10ck:
Sorry, I think we're just not going to merge this. This falls under "personal preferences" and breaks a lot of other PRs.

PR #6068, opened 2023-02-20
feat: add nasm language

### LINKED ISSUE #3329: Add highlighting for assembler (.s)
Would need a tree-sitter grammar for that.

mtoohey31:
This pull request adds nasm assembly as a language.

Closes #3329.

Under the issue mentioned above, there is a discussion about how to support various assembly flavours. I think the best approach is to add flavour-specific language definitions with the appropriate grammars, then users can override the `file-types` values as desired in their local `languages.toml` files so that the `.asm` extension (or whichever extension they're using) is recognized as the flavour that they're working with. Since nasm will be the only flavour supported so far after the changes in this merge request, I've included the `.asm` extension under nasm's `file-types` so that there is at least _some_ syntax highlighting out of the box, even though it might not be for the right flavour. If there's a different approach that might work better though, just let me know and I can rework things.

Also, CC @naclsn (the original author of the grammar and queries). I hope you don't mind me including your work here; I've copied the queries pretty much verbatim, only I've removed some todo-related comments that probably don't belong in the version inside this repository, and I've uncommented the `(ERROR) @error` highlight.

Edit: oh, and here's a screenshot:

helix-nasm-nord

### DIFF
--- a/runtime/queries/nasm/highlights.scm
+++ b/runtime/queries/nasm/highlights.scm
@@ -1,0 +1,130 @@
+(comment) @comment
+(ERROR) @error
+
+(label) @label
+
+[
+  (preproc_expression)
+  (line_here_token)
+  (section_here_token)
+] @variable.builtin
+(preproc_expression
+  "]" @variable.builtin)
+(preproc_expression
+  "}" @variable.builtin)
+
+(unary_expression
+  operator: _ @operator.unary)
+(binary_expression
+  operator: _ @operator.binary)
+(conditional_expression
+  "?" @operator
+  ":" @operator)
+
+[
+  ":"
+  ","
+] @punctuation.delimiter
+
+[
+  "("
+  ")"
+  "["
+  "]"
+  "{"
+  "}"
+] @punctuation.bracket
+
+(instruction_prefix) @keyword
+(actual_instruction
+  instruction: (word) @function)
+
+(call_syntax_expression
+  base: (word) @function)
+
+(size_hint) @type
+(struc_declaration
+  name: (word) @type)
+(struc_instance
+  name: (word) @type)
+
+(effective_address
+ hint: _ @type)
+(effective_address
+ segment: _ @constant.builtin)
+
+(register) @constant.builtin
+
+(number_literal) @number.integer
+(string_literal) @string
+(float_literal) @number.float
+(packed_bcd_literal) @number.integer
+
+((word) @constant
+  (#match? @constant "^[A-Z_][?A-Z_0-9]+$"))
+((word) @constant.builtin
+  (#match? @constant.builtin "^__\\?[A-Z_a-z0-9]+\\?__$"))
+(word) @variable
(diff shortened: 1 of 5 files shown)


naclsn:
Neat, thanks for using my work!

Like most language with (C-like) preprocessor, the written code can be syntactically invalid despite being correct. This is why I had the same approach as the C highlight queries to not have `@error` at all.

mtoohey31:
> Neat, thanks for using my work!
> 
> Like most language with (C-like) preprocessor, the written code can be syntactically invalid despite being correct. This is why I had the same approach as the C highlight queries to not have `@error` at all.

Ah, that makes sense! I've removed that query in eda2cc6590c783719871c95bde85e6589326bf69.

runtime/queries/nasm/highlights.scm
```diff
@@ -0,0 +1,129 @@
...
+] @variable.builtin
+(preproc_expression
+  "]" @variable.builtin)
+(preproc_expression
+  "}" @variable.builtin)
+
+(unary_expression
+  operator: _ @operator.unary)
+(binary_expression
+  operator: _ @operator.binary)
```
the-mikedavis, line 18: Currently we don't use these elsewhere

```suggestion
  operator: _ @operator)
(binary_expression
  operator: _ @operator)
```

runtime/queries/nasm/highlights.scm
```diff
@@ -0,0 +1,129 @@
...
+ hint: _ @type)
+(effective_address
+ segment: _ @constant.builtin)
+
+(register) @constant.builtin
+
+(number_literal) @number.integer
+(string_literal) @string
+(float_literal) @number.float
+(packed_bcd_literal) @number.integer
```
the-mikedavis, line 60:
```suggestion
(number_literal) @constant.numeric.integer
(string_literal) @string
(float_literal) @constant.numeric.float
(packed_bcd_literal) @constant.numeric.integer
```

runtime/queries/nasm/highlights.scm
```diff
@@ -0,0 +1,129 @@
...
+
+((word) @constant
+  (#match? @constant "^[A-Z_][?A-Z_0-9]+$"))
+((word) @constant.builtin
+  (#match? @constant.builtin "^__\\?[A-Z_a-z0-9]+\\?__$"))
+(word) @variable
+
+(preproc_arg) @keyword
```
the-mikedavis, line 68: If the preprocessor stuff is like the C preprocessor, these might fall under `@keyword.directive` https://docs.helix-editor.com/master/themes.html#syntax-highlighting

mtoohey31: I think you're right that `@keyword.directive` makes more sense. I've switched it to that in df213757f80e48498dfb5ea48de678c296af0f69 along with a few other improvements.

the-mikedavis (approved):

PR #891, opened 2021-10-22
repeat find char motion and fix a motion bug

ghost:
Implement repeat find motion (find_till_char, find_next_char, extend_till_char, extend_next_char and pre {find,extend})

Fix a bug find till motion works unexpect when the next char of cursor equals searching char.

https://user-images.githubusercontent.com/20379044/138402468-4496a2f7-f4d5-4633-966a-19c596e8b3e0.mp4

PR #6078, opened 2023-02-22
(Updated) Apply motion API refinements

gibbz00:
Same as #6061 but without relying on #6056 as per the discussion we had there.

## Summary:

`last_motion` is now a private field and can only be set by calling `editor.apply_motion()`. Thus removing the need (and possibility of forgetting) to write the following every time a motion is to be applied and saved:

```
motion(editor); 
editor.last_motion = motion
```

Now it's just: `editor.apply_motion(motion)`

New behavior that fell out from this is that repetitions of t,T,f,F motions are stored with the context of extension/move and count. (Not defaulting to extend by 1 count). So 3f -> repeat_last_motion becomes 3f and not 1f. I'm not sure the old behavior was intentional, as I saw no discussion of it when introduced in #891.

## Conflicts

Of all the open PRs that are currently mergeable with master, and pass all CI checks; only #2869 conflicts with this one. There it is simply an import statement in command.rs that needs to be changed.

```
 5 ▍<<<<<<< HEAD
 4  │   editor::Action,
 3 ▍=======
 2 ▍│   editor::{Action, Motion} 
 1 ▍│   icons::Icons,
43 ▍>>>>>>> TEMP_2869
```

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1239,10 +1239,11 @@ fn extend_next_long_word_end(cx: &mut Context) {
     extend_word_impl(cx, movement::move_next_long_word_end)
 }
 
-fn find_char<F>(cx: &mut Context, search_fn: F, inclusive: bool, extend: bool)
-where
-    F: Fn(RopeSlice, char, usize, usize, bool) -> Option<usize> + 'static,
-{
+enum SearchDirection {
+    Next,
+    Prev,
+}
+fn find_char(cx: &mut Command, search_direction: SearchDirection, inclusive: bool, extend: bool) {
     // TODO: count is reset to 1 before next key so we move it into the closure here.
     // Would be nice to carry over.
     let count = cx.count();
@@ -1274,12 +1275,18 @@ where
             } => ch,
             _ => return,
         };
+        let motion = move |editor: &mut Editor| {
+            match search_direction {
+                SearchDirection::Next => {
+                    find_char_impl(editor, &find_next_char_impl, inclusive, extend, ch, count)
+                }
+                SearchDirection::Prev => {
+                    find_char_impl(editor, &find_prev_char_impl, inclusive, extend, ch, count)
+                }
+            };
+        };
 
-        // NOTE: not using _apply_motion as the repitition isn't identical
-        find_char_impl(cx.editor, &search_fn, inclusive, extend, ch, count);
-        cx.editor.last_motion = Some(Motion(Box::new(move |editor: &mut Editor| {
-            find_char_impl(editor, &search_fn, inclusive, extend, ch, 1);
-        })));
+        _apply_motion(cx, motion);
     })
 }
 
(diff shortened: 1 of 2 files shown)


helix-view/src/editor.rs
```diff
@@ -856,8 +844,8 @@ pub struct Editor {
     pub auto_pairs: Option<AutoPairs>,
 
     pub idle_timer: Pin<Box<Sleep>>,
-    pub last_motion: Option<Motion>,
-
+    #[allow(clippy::complexity)]
+    last_motion: Option<Arc<Box<dyn Fn(&mut Editor)>>>,
```
pickfire, line 848: Why not make a `type` for it?

gibbz00: Only used once 🤷

poliorcetics: You can probably directly use `Arc<dyn Fn(&mut Editor)>` instead of wrapping a `Box` in an `Arc`

gibbz00: You're right, thanks :)

helix-term/src/commands.rs
```diff
@@ -1188,10 +1187,11 @@ fn extend_next_long_word_end(cx: &mut Context) {
     extend_word_impl(cx, movement::move_next_long_word_end)
 }
 
-fn will_find_char<F>(cx: &mut Context, search_fn: F, inclusive: bool, extend: bool)
-where
-    F: Fn(RopeSlice, char, usize, usize, bool) -> Option<usize> + 'static,
-{
+enum SearchDirection {
```
dead10ck, line 1190: Rather than introduce a new type, I'd use the existing `Direction` type

https://github.com/helix-editor/helix/blob/6494fc1daf2bb379e6a1f4ec94f3dafd99b3fb34/helix-core/src/movement.rs#L23-L26

dead10ck (changes requested):
Just one issue I can see, but this is a good change overall 👍

helix-term/src/commands.rs
gibbz00: Thanks!

gibbz00:
@dead10ck Fixed suggestions and rebased with master 😊

helix-view/src/editor.rs
the-mikedavis: I agree with pickfire here: even in cases like this where it's a private field in a struct and only used once we always make another type. See `helix-term/src/ui/picker.rs` for example.

gibbz00: Would a type alias be enough? https://github.com/helix-editor/helix/pull/6078/commits/57eab67f07169fa03361191954bf9101207404f0

helix-view/src/editor.rs
```diff
@@ -908,8 +896,7 @@ pub struct Editor {
     pub auto_pairs: Option<AutoPairs>,
 
     pub idle_timer: Pin<Box<Sleep>>,
-    pub last_motion: Option<Motion>,
-
+    last_motion: LastMotion,
```
the-mikedavis, line 899: I think this was clearer before as `Option<Motion>`. It can just be a `type Motion` or `type LastMotion` rather than the newtype it was before.

the-mikedavis (changes requested):
There's an opportunity to expand the functionality of the last motion to take a `Movement` argument so that you can move or extend based on the editor mode. Since we're already making a cosmetic change here it's a good time to refactor that function

See also https://github.com/helix-editor/helix/issues/7033#issuecomment-1546693389

helix-view/src/editor.rs
the-mikedavis: Yep I think that's fine. The newtype it was before was probably not necessary

helix-view/src/editor.rs
gibbz00: Option<Motion> it is, thanks!

gibbz00:
> There's an opportunity to expand the functionality of the last motion to take a `Movement` argument so that you can move or extend based on the editor mode. Since we're already making a cosmetic change here it's a good time to refactor that function

Is it ok if we add that in a separate PR? I worked on a more general solution to that a while a go, and I remember it being quite comprehensive before moving to other things.  There are a lot of other commands that should possibly also behave based on mode.

the-mikedavis:
Yeah, I probably don't have an appetite for a change that large. I mean just about these repeatable motions but that can be a separate PR. The changes here should make that a smaller and more centralized change so I'm ok with this. In general though I'm very hesitant about these "refinement"-style PRs. This one is small and uncontroversial so I'm ok with it but larger changes usually don't strike a good balance between time spent reading/reviewing/discussing plus the cost of changing the codebase in general versus concrete value-added.

helix-view/src/editor.rs
```diff
@@ -943,6 +930,8 @@ pub struct Editor {
     pub completion_request_handle: Option<oneshot::Sender<()>>,
 }
 
+pub type Motion = Arc<dyn Fn(&mut Editor)>;
```
the-mikedavis, line 933: I don't think the Arc is necessary here, this can just be a Box as it was

gibbz00: right 🙈

gibbz00:
> In general though I'm very hesitant about these "refinement"-style PRs. This one is small and uncontroversial so I'm ok with it but larger changes usually don't strike a good balance between time spent reading/reviewing/discussing plus the cost of changing the codebase in general versus concrete value-added.

I get where you're coming from. This PR cherry-picked out from a feature branch I worked on a while ago that tried to create a "motion pipeline" that would allow motions to be made more consistent and/or configurable with regard to expand variants, jump list pushes, replay motions, count use and guaranteed bijectivity for commands that have an inverse etc... So yeah, it might feel like a bit out of place, but that's why.

I guess I just haven't found a good balance between sending in PRs that are isolated enough to be bearable to review, yet important enough to add tangible value by themselves. I've also accepted to a greater extent that some code is just the way it is, and refrain from touching it unless it comes in direct contact with a feature I'm trying to work on. Instead, I usually try to mention such "pain points" as footnotes when wrapping the PR draft. ✌️

the-mikedavis (approved):

pascalkuthe (approved):

dead10ck (approved):
lgtm, thanks!

dead10ck (approved):
lgtm, thanks!

PR #6079, opened 2023-02-22
Implement LSP window/showDocument request

MDeiml:
Implements the LSP request for [`window/showDocument`](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument).

This does not yet implement requests with the `external` parameter set to `true`, but it would probably best to do this in the same PR (as LSP servers would probably assume that to work). I believe there is no similar functionality in helix yet, and I wanted to check if my suggested approach would be welcome.

I suggest adding a `External` variant to the `Action` enum which is passed to `Editor::open`. The actual implementation could either be done using the [`open` crate](https://docs.rs/open/latest/open/) or by reimplementing the functionality of this crate.

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -527,6 +527,7 @@ pub enum MethodCall {
     WorkspaceFolders,
     WorkspaceConfiguration(lsp::ConfigurationParams),
     RegisterCapability(lsp::RegistrationParams),
+    ShowDocument(lsp::ShowDocumentParams),
 }
 
 impl MethodCall {
@@ -549,6 +550,10 @@ impl MethodCall {
             lsp::request::RegisterCapability::METHOD => {
                 let params: lsp::RegistrationParams = params.parse()?;
                 Self::RegisterCapability(params)
+            }
+            lsp::request::ShowDocument::METHOD => {
+                let params: lsp::ShowDocumentParams = params.parse()?;
+                Self::ShowDocument(params)
             }
             _ => {
                 return Err(Error::Unhandled);
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -959,6 +959,14 @@ impl Application {
             Call::MethodCall(helix_lsp::jsonrpc::MethodCall {
                 method, params, id, ..
             }) => {
+                let offset_encoding = match self.editor.language_servers.get_by_id(server_id) {
+                    Some(language_server) => language_server.offset_encoding(),
+                    None => {
+                        warn!("can't find language server with id `{}`", server_id);
+                        return;
+                    }
+                };
+
                 let reply = match MethodCall::parse(&method, params) {
                     Err(helix_lsp::Error::Unhandled) => {
                         error!(
@@ -999,11 +1007,7 @@ impl Application {
                         Ok(serde_json::Value::Null)
                     }
                     Ok(MethodCall::ApplyWorkspaceEdit(params)) => {
-                        let res = apply_workspace_edit(
-                            &mut self.editor,
-                            helix_lsp::OffsetEncoding::Utf8,
-                            &params.edit,
-                        );
+                        apply_workspace_edit(&mut self.editor, offset_encoding, &params.edit);
 
                         Ok(json!(lsp::ApplyWorkspaceEditResponse {
                             applied: res.is_ok(),
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -5,7 +5,11 @@ use helix_core::{
     path::get_relative_path,
     pos_at_coords, syntax, Selection,
 };
-use helix_lsp::{lsp, util::lsp_pos_to_pos, LspProgressMap};
+use helix_lsp::{
+    lsp,
+    util::{lsp_pos_to_pos, lsp_range_to_range},
+    LspProgressMap,
+};
 use helix_view::{
     align_view,
     document::DocumentSavedEventResult,
@@ -1007,7 +1011,8 @@ impl Application {
                         Ok(serde_json::Value::Null)
                     }
                     Ok(MethodCall::ApplyWorkspaceEdit(params)) => {
-                        apply_workspace_edit(&mut self.editor, offset_encoding, &params.edit);
+                        let res =
+                            apply_workspace_edit(&mut self.editor, offset_encoding, &params.edit);
 
                         Ok(json!(lsp::ApplyWorkspaceEditResponse {
                             applied: res.is_ok(),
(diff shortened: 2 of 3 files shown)


helix-term/src/application.rs
```diff
@@ -1029,6 +1029,61 @@ impl Application {
...
+                                            );
+                                            self.editor.set_error(err);
+                                            Ok(json!(lsp::ShowDocumentResult { success: false }))
+                                        }
+                                        Ok(_) => {
+                                            if let Some(range) = params.selection {
+                                                let (view, doc) = current!(self.editor);
+                                                let definition_pos = range.start;
+                                                // TODO: convert inside server
+                                                if let Some(new_pos) = lsp_pos_to_pos(
```
the-mikedavis, line 1063: This should use `lsp_range_to_range` on the whole `range`

helix-term/src/application.rs
```diff
@@ -1029,6 +1029,61 @@ impl Application {
...
+                                                    doc.text(),
+                                                    definition_pos,
+                                                    helix_lsp::OffsetEncoding::Utf8,
```
the-mikedavis, line 1066: Here we should use the `language_server.offset_encoding()`: [this block](https://github.com/helix-editor/helix/blob/8043959265366ed0c945b7f868337a3f9823a747/helix-term/src/application.rs#L1034-L1040) can be moved above the `match call { .. }` and then we should also use the `language_server.offset_encoding()` [here](https://github.com/helix-editor/helix/blob/8043959265366ed0c945b7f868337a3f9823a747/helix-term/src/application.rs#L986-L990)

helix-term/src/application.rs
```diff
@@ -1029,6 +1029,61 @@ impl Application {
...
+                                                    );
+                                                    align_view(doc, view, Align::Center);
+                                                }
+                                            }
+
+                                            Ok(json!(lsp::ApplyWorkspaceEditResponse {
+                                                applied: true,
+                                                failure_reason: None,
+                                                failed_change: None,
+                                            }))
```
the-mikedavis, line 1080: Shouldn't this be `Ok(json!(lsp::ShowDocumentResult { success: true }))`?

helix-term/src/application.rs
MDeiml: I had to add a `get_by_id_arc` methods to `Registry` as the original `get_by_id` returned a immutable reference, meaning that `self.editor` could not be borrowed mutably.

MDeiml:
Thanks for the quick feedback :)

helix-term/src/application.rs
```diff
@@ -967,6 +971,14 @@ impl Application {
                     }
                 };
 
+                let language_server = match self.editor.language_servers.get_by_id_arc(server_id) {
+                    Some(language_server) => language_server,
+                    None => {
+                        warn!("can't find language server with id `{}`", server_id);
+                        return;
+                    }
+                };
```
poliorcetics, line 980:
```suggestion
                let offset_encoding = match self.editor.language_servers.get_by_id(server_id) {
                    Some(language_server) => language_server.offset_encoding(),
                    None => {
                        warn!("can't find language server with id `{}`", server_id);
                        return;
                    }
                };
```

helix-term/src/application.rs
```diff
@@ -985,7 +997,7 @@ impl Application {
                     MethodCall::ApplyWorkspaceEdit(params) => {
                         apply_workspace_edit(
                             &mut self.editor,
-                            helix_lsp::OffsetEncoding::Utf8,
+                            language_server.offset_encoding(),
```
poliorcetics, line 1000:
```suggestion
                            offset_encoding,
```

helix-term/src/application.rs
```diff
@@ -1029,13 +1041,67 @@ impl Application {
...
+                                            Ok(json!(lsp::ShowDocumentResult { success: false }))
+                                        }
+                                        Ok(_) => {
+                                            if let Some(range) = params.selection {
+                                                let (view, doc) = current!(self.editor);
+                                                // TODO: convert inside server
+                                                if let Some(new_range) = lsp_range_to_range(
+                                                    doc.text(),
+                                                    range,
+                                                    language_server.offset_encoding(),
```
poliorcetics, line 1086:
```suggestion
                                                    offset_encoding,
```

helix-lsp/src/lib.rs
```diff
@@ -439,6 +444,13 @@ impl Registry {
             .map(|(_, client)| client.as_ref())
     }
 
+    pub fn get_by_id_arc(&self, id: usize) -> Option<Arc<Client>> {
+        self.inner
+            .values()
+            .find(|(client_id, _)| client_id == &id)
+            .map(|(_, client)| client.clone())
+    }
+
```
poliorcetics, line 453:
```suggestion
```

poliorcetics (changes requested):
You can avoid adding a new method by simply storing the `offset_encoding` earlier and keeping it for the duration of the method, since that type is `Copy` it won't carry the borrow of `self.editor`

MDeiml:
> You can avoid adding a new method by simply storing the `offset_encoding` earlier and keeping it for the duration of the method, since that type is `Copy` it won't carry the borrow of `self.editor`

Hm, that would work, but I would need to call `get_by_id` again after the match block. I feel like it's a bit neater with the `Arc`. Also future contributions might also need the `Arc` in this method or other places so it probably makes sense for `Registry` to have this method.

poliorcetics:
Ah I didn't check if the language server itself was used later

helix-term/src/application.rs
```diff
@@ -1029,13 +1041,67 @@ impl Application {
                             .collect();
                         Ok(json!(result))
                     }
-                };
+                    MethodCall::ShowDocument(params) => {
+                        use helix_view::editor::Action;
+
+                        if params.external.unwrap_or(false) {
+                            match open::that(params.uri.as_str()) {
```
pascalkuthe, line 1048: `open::that` is a blocking function that is ill suited for use in async code (or really helix in general). See discussion in https://github.com/Byron/open-rs/pull/64 and https://github.com/helix-editor/helix/pull/5820

helix-term/src/application.rs
pascalkuthe: Just a small note: This is an existing bug in helix and something I missed in #5894 entirely orthogonal to this PR (that said the fix is rather simple and correct although @poliorcetics suggestions should definitely be applied)

helix-term/src/application.rs
```diff
@@ -1029,13 +1041,67 @@ impl Application {
...
+                                        Ok(_) => {
+                                            if let Some(range) = params.selection {
+                                                let (view, doc) = current!(self.editor);
+                                                // TODO: convert inside server
+                                                if let Some(new_range) = lsp_range_to_range(
+                                                    doc.text(),
+                                                    range,
+                                                    language_server.offset_encoding(),
+                                                ) {
+                                                    doc.set_selection(
```
pascalkuthe, line 1088: I think this should receive a jumplist entry so that the old position on the document isn't lost if it was already open.

```suggestion
                                                    let jump = (doc.id(), doc.selection(view.id).clone());
                                                    view.jumps.push(jump);
                                                    doc.set_selection(
```

pascalkuthe (changes requested):

helix-term/src/application.rs
MDeiml: I removed the commit. I guess the best thing is to leave this unimplemented for now or wait for Byron/open-rs#64?

helix-term/src/application.rs
```diff
@@ -1029,6 +1037,62 @@ impl Application {
...
+                                            Ok(json!(lsp::ShowDocumentResult { success: true }))
+                                        }
+                                    }
+                                }
+                            }
+                        }
+                    }
                 };
 
                 let language_server = match self.editor.language_servers.get_by_id(server_id) {
```
the-mikedavis, line 1098: This can be changed to an unwrap now that we return early if there isn't a language server for this ID above

MDeiml:
Should be ready to merge now. It would also make sense to wait on Byron/open-rs#64.

matoous:
There’s open PR for opening files externally: https://github.com/helix-editor/helix/pull/5820

MDeiml:
Oh, the `open` PR was merged in the meantime. So I guess it's worth waiting for #5820 and rebasing the changes on that to support opening externally.

matoous:
@MDeiml I am sorry, I completely forgot about existence of this PR 🙈 here's second implementation of this based on the PR for handling opening documents in external programs: https://github.com/helix-editor/helix/pull/8865

MDeiml:
@matoous haha I already saw it, no worries. Let's just use yours then.

pascalkuthe:
Closing in favor of https://github.com/helix-editor/helix/pull/8865 as discussed above. Thanks for working on this @MDeiml

PR #1340, opened 2021-12-23
feat(commands): shrink_selection

matoous:
Add `shrink_selection` command that can be used to shrink previously expanded selection.

To make `shrink_selection` work it was necessary to add selection history to the Document since we want to shrink the selection towards the syntax tree node that was initially selected.

Selection history is cleared any time the user changes selection other way than by `expand_selection`. This ensures that we don't get some funky edge cases when user calls `shrink_selection`.

Related: https://github.com/helix-editor/helix/discussions/1328

PR #6088, opened 2023-02-23
fix #6087

### LINKED ISSUE #6087: `shrink_selection` breaks when `expand_selection` is used on a different document in the same view
### Summary

Introduced in https://github.com/helix-editor/helix/pull/1340

Not linking the selection change to a document leads to broken behavior.

Say `expand_selection` is used on the current document, and then again on another document in the same view,
If `shrink_selection` is used on the original document, then the wrong selection will be popped from
`view.object_selections`. The resulting behavior is usually that `shrink_selection` stops working on the original document.

Also, not linking it to the document will in no way free the selection memory when the document is closed, which I presume is a memory leak.

### Reproduction Steps

[asciicast](https://asciinema.org/a/7Y5ILkMXFQO5Jrytqd1A7XSOG)

Jittering and weird color/characters is only on the asciinema. Not visible on my screen during recording nor normal use of helix.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

foot

### Helix Version

master

gibbz00:
Explained there: #6087

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -117,7 +117,7 @@ pub struct View {
     // two last modified docs which we need to manually keep track of
     pub last_modified_docs: [Option<DocumentId>; 2],
     /// used to store previous selections of tree-sitter objects
-    pub object_selections: Vec<Selection>,
+    pub object_selections: HashMap<DocumentId, Vec<Selection>>,
     /// all gutter-related configuration settings, used primarily for gutter rendering
     pub gutters: GutterConfig,
     /// A mapping between documents and the last history revision the view was updated at.
@@ -151,7 +151,7 @@ impl View {
             jumps: JumpList::new((doc, Selection::point(0))), // TODO: use actual sel
             docs_access_history: Vec::new(),
             last_modified_docs: [None, None],
-            object_selections: Vec::new(),
+            object_selections: HashMap::new(),
             gutters,
             doc_revisions: HashMap::new(),
         }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4310,7 +4310,12 @@ fn expand_selection(cx: &mut Context) {
             // check if selection is different from the last one
             if *current_selection != selection {
                 // save current selection so it can be restored using shrink_selection
-                view.object_selections.push(current_selection.clone());
+                if let Some(object_selections) = view.object_selections.get_mut(&doc.id()) {
+                    object_selections.push(current_selection.clone());
+                } else {
+                    view.object_selections
+                        .insert(doc.id(), vec![current_selection.clone()]);
+                }
 
                 doc.set_selection(view.id, selection);
             }
@@ -4325,7 +4330,11 @@ fn shrink_selection(cx: &mut Context) {
         let (view, doc) = current!(editor);
         let current_selection = doc.selection(view.id);
         // try to restore previous selection
-        if let Some(prev_selection) = view.object_selections.pop() {
+        if let Some(prev_selection) = view
+            .object_selections
+            .get_mut(&doc.id())
+            .and_then(|e| e.pop())
+        {
             if current_selection.contains(&prev_selection) {
                 // allow shrinking the selection only if current selection contains the previous object selection
                 doc.set_selection(view.id, prev_selection);


dead10ck (commented):
Would it be simpler just to clear the object selections any time we switch the view to a different document?

gibbz00:
Wouldn't that be a regression? The entire feature is just thrown away. And it's not like this is very complicated right?

dead10ck:
What do you mean thrown away? If I recall correctly, the original use case was to allow successive expand / shrink calls to remember the whole inner selection, rather than just the first child.

By tracking the selecting history per document like this, it would be enriching in that you could switch documents and still retain the previous selections for a document you switch back to. But the use case would be something like:

1. Expand n times
1. Switch documents, do other things
1. Go back to the first document
1. Shrink back to smaller whole subtree selections with shrinking n - m times

Is this a common use case for you?

gibbz00:
I really don't understand what you're trying say. Is there anything wrong with this PR? Why should the shrink to previous textobject selection stop working if an expand_selection is done on another document?

helix-view/src/view.rs
```diff
@@ -117,7 +117,7 @@ pub struct View {
     // two last modified docs which we need to manually keep track of
     pub last_modified_docs: [Option<DocumentId>; 2],
     /// used to store previous selections of tree-sitter objects
-    pub object_selections: Vec<Selection>,
+    pub object_selections: HashMap<DocumentId, Vec<Selection>>,
```
pascalkuthe, line 120: Storing per document data on the view is not a good idea. If the document is closed that data is essentially leaked and it also preculdes you from updating data on edits. If we are going to be storing this per-view and per-document then this should be stored on the `Document` as ` HashMap<ViewId,..>` just like selections are and inserted cleared in `remove_view` and `ensure_view_init`

pascalkuthe (changes requested):
I think you identified a larger problem with the `object_selection` than what was fixed here. It's not just that one document can contaminate another documents `object_selections` it's also that `object_selections` are not updated on edits so if you used `Alt-i` somewhereits possible to randomly get a completly nonsensical result. In general all of these bugs are just a symtom of a much larger problem the `object_selections` sticks around forever so if you are inside a a rust function in an impl block and do `alt-o` once to select the current word (or call) and then use `mat` to select the entire implt block then doing `alt-i` puts you right back at the word (rather then the first TS child). This was actually raised in the original PR: https://github.com/helix-editor/helix/pull/1340#discussion_r776108732. Even weirder the history is never clear so you get weird behaviour like the following:

``` rust
fn foo(){
   let bar = 1;
   let foo = 2;
}
```

if you expand the selection to the `foo` statement first, the do something else, comeback and expand the selection on the `bar` statement and then select the entire function and shrink_selection the cursor will select the bar statement first and the `foo` statement second. 

I am not entirely sure what the best way to solve this should be but we clearly need to clear the `object_selections` history much more agressively. The easiest solution would be to just clear the history if:

* the selection changes (and it's not caused by `expand_selection` or `shrink_selection`)
* the document changes
* the visible document changes

Alternatively we could treat `object_selections` closer to a real selection and store it on the document (and therefore loose the third rule). In theory we could even map object_selections trough edits (like we do normal selections) and therefore also loose the third rule, I added a comment with moreA details. 

The problem is that this regresses cases where we might select something else go back with `c-o` and then shrink the selection again. We might just want to store `object_selections` on the jumplist to account for those cases. 

Btw. I also think that `shrink_selection` needs a ton of improvements in case there isn't a history. It will often select the top of the file in rust for example. I think that `shrink_selection` should only ever decrease the size of the selection and never select anything outside of it. We probably need to look for the first child that is fully contained in the selection instead of just using the first child. That should be a separate PR tough

dead10ck:
>I am not entirely sure what the best way to solve this should be but we clearly need to clear the `object_selections` history much more agressively.

Yeah, I'm actually working on a feature right now that I started a while ago, but I'm picking back up, and this feature also involves [changes to the object selections](https://github.com/dead10ck/helix/commit/41b6b9839d7e33d75b0852da1ab94bc2ca9a9694#diff-18195c1eaadd7a2747b6556008c6022d79f814d64624e9e69d0735a3e0bbe296R120). I came to the same discovery about how messed up the object selections can become through writing integration tests for it. I decided to solve it by [clearing all object selection history any time a selection is set on a view](https://github.com/dead10ck/helix/commit/41b6b9839d7e33d75b0852da1ab94bc2ca9a9694#diff-36a64f6c432dc47d11c27851879a0b67459d908edfaa70262aefc21171dc62b9R783). It's aggressive, but I think it fixes the problems.

It does regress the feature, but I think the only way to solve the problems *and* retain history through more operations than exclusively `expand_history` is by making it significantly more complicated, i.e. tracking all text mutations in them, etc., like you mentioned.

PR #6093, opened 2023-02-23
Fix `shrink_selection` with multiple cursors.

### LINKED ISSUE #6092: `shrink_selection` does not properly restore cursor with multiple selections.
### Summary

Asciinema says most of it.

### Reproduction Steps

[asciicast](https://asciinema.org/a/SNtXeWHmQBFPXdEScj5rKcC4T)

### Platform

Linux

### Terminal Emulator

kitty

gibbz00:
Fixes #6092 

Caused by some incorrect assumptions that missed an edge case in the `Selection.contains()` calculation. Tests were added accordingly.

[asciicast](https://asciinema.org/a/NINyXHNywRUe6ULYZo4nGQjCi?t=23)

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -615,11 +615,6 @@ impl Selection {
 
     // returns true if self ⊇ other
     pub fn contains(&self, other: &Selection) -> bool {
-        // can't contain other if it is larger
-        if other.len() > self.len() {
-            return false;
-        }
-
         let (mut iter_self, mut iter_other) = (self.iter(), other.iter());
         let (mut ele_self, mut ele_other) = (iter_self.next(), iter_other.next());
 
@@ -635,7 +630,20 @@ impl Selection {
                     };
                 }
                 (None, Some(_)) => {
-                    // exhausted `self`, we can't match the reminder of `other`
+                    // Other might still contain all of its ranges in a single range of self,
+                    let mut still_subset: bool;
+                    for self_range in self {
+                        still_subset = true;
+                        for other_range in other {
+                            if !self_range.contains_range(other_range) {
+                                still_subset = false;
+                                break;
+                            }
+                        }
+                        if still_subset {
+                            return true;
+                        }
+                    }
                     return false;
                 }
                 (_, None) => {
@@ -1230,5 +1238,8 @@ mod test {
             vec!((3, 4), (7, 9))
         ));
         assert!(!contains(vec!((1, 1), (5, 6)), vec!((1, 6))));
+
+        // multiple ranges of other are all contained in one range of self,
+        assert!(contains(vec!((7, 13)), vec!((7, 9), (11, 12))));
     }
 }
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -630,20 +630,7 @@ impl Selection {
                     };
                 }
                 (None, Some(_)) => {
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -4327,7 +4327,6 @@ fn shrink_selection(cx: &mut Context) {
         // try to restore previous selection
         if let Some(prev_selection) = view.object_selections.pop() {
             if current_selection.contains(&prev_selection) {
-                // allow shrinking the selection only if current selection contains the previous object selection
```
dead10ck, line 4330: This is still true, so we should not remove this comment.

helix-core/src/selection.rs
```diff
@@ -615,11 +615,6 @@ impl Selection {
 
     // returns true if self ⊇ other
     pub fn contains(&self, other: &Selection) -> bool {
-        // can't contain other if it is larger
-        if other.len() > self.len() {
```
dead10ck, line 619: I am fairly certain this is the only change needed to get this working for the cases where there are big ranges in the first one that contain several smaller ranges in the second one. The algorithm is correct for these cases, this was just an incorrect attempt at a fast path.

helix-core/src/selection.rs
```diff
@@ -635,7 +630,20 @@ impl Selection {
                     };
                 }
                 (None, Some(_)) => {
-                    // exhausted `self`, we can't match the reminder of `other`
+                    // Other might still contain all of its ranges in a single range of self,
+                    let mut still_subset: bool;
+                    for self_range in self {
```
dead10ck, line 635: So I think there are several issues here:

1. This is not necessary, as mentioned above
2. It actually makes it less correct, because you could have multiple cross sections of ranges where a larger range in the first overlaps with multiple smaller ranges in the second, e.g.
    
    ```
    [a b c d e f g]  [h i j k l m n]
    [a]b c[d]e f[g]  [h]i j[k]l m[n]
    ```
    
    In this case, it will say it doesn't contain it because of `hkn`
3. It is currently an incremental algorithm linear on the number of ranges in the selections, but this makes it exponential.

dead10ck (changes requested):

helix-term/src/commands.rs
gibbz00: Yes it still holds true, but the comment is reduntand an out of place. Usually, they explain the proceeding code. But here its explaining the if statement from within.  And it is not adding any context as the code is extremely self expanatory.

```rust
if current_selection.contains(&prev_selection) {
    doc.set_selection(view.id, prev_selection);
```

helix-core/src/selection.rs
gibbz00: You're right. I had overlooked exactly what was being done in the first match arm. I also expanded the test case for the example case class you gave. Should be resolved now, thank you.

helix-term/src/commands.rs
dead10ck: The code is pretty self explanatory, but comments aren't solely to help explain what code is doing; they also help clarify what the original intention behind the code was. Anyone looking at this can see the logic is such that it will only accept a previous selection if the current one contains it, but it's not always clear *why*. Having a comment written in plain English makes it clear that this was in fact intentional.

If you take issue with placement, it would be fine to move it.

dead10ck (approved):

PR #6095, opened 2023-02-24
LSP: Separate diagnostic picker message and code

FrancescElies:
**Problem**
Currently in master lsp errors are reported as follows, as you can see the diagnostic message and code are attached to each other making it hard to read.
grafik

**This pr** solves that.
grafik

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -32,8 +32,7 @@ use crate::{
 };
 
 use std::{
-    borrow::Cow, cmp::Ordering, collections::BTreeMap, fmt::Write, future::Future, path::PathBuf,
-    sync::Arc,
+    cmp::Ordering, collections::BTreeMap, fmt::Write, future::Future, path::PathBuf, sync::Arc,
 };
 
 /// Gets the language server that is attached to a document, and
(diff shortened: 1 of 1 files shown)


dead10ck (approved):
Seems like a good change to me 👍

helix-term/src/commands/lsp.rs
```diff
@@ -154,7 +154,8 @@ impl ui::menu::Item for PickerDiagnostic {
         Spans::from(vec![
             Span::raw(path),
             Span::styled(&self.diag.message, style),
-            Span::styled(code, style),
+            Span::styled(" ", style),
+            Span::styled(format!("({code})"), style),
```
the-mikedavis, line 158: This will draw the extra space and empty parens if the diagnostic's `code` is `None`. Instead the formatting can be done at the `let code` part:

```rust
let code = self
    .diag
    .code
    .as_ref()
    .map(|c| {
        format!(
            " ({})",
            match c {
                NumberOrString::Number(n) => n.to_string(),
                NumberOrString::String(s) => s.clone(),
            }
        )
    })
    .unwrap_or_default();
```

poliorcetics:
```rust
let code = match self.diag.code.as_ref() {
    Some(NumberOrString::Number(n)) => format!(" ({n})"),
    Some(NumberOrString::String(s)) => format!(" ({s})"),
    None => String::new(),
};
```

Will allocate less (no need for `clone()` nor `to_string()`) and be shorter

the-mikedavis (approved):

PR #6102, opened 2023-02-25
Semantic tokens highlighting

poliorcetics:
This PR introduces Semantic Highlighting capabilities to Helix.

## Unresolved question

In this first version, all semantic highlighting is done via a dumb list like this:

Output of `:semantic-token` command in Helix with this PR

But VsCode does it like this (https://rust-analyzer.github.io/manual.html#semantic-style-customizations):

```jsonc
{
  "editor.semanticTokenColorCustomizations": {
    "rules": {
      "*.mutable": {
        "fontStyle": "underline", // underline is the default
      },
    }
  },
}
```

I much prefer VsCode's version since it allows applying modifiers only to certain elements, like `variable.mutable`,
without too much repetions.

In a first version I could simply concatenate everything under `semantic.<type>.<mod1>.<mod2>...` so that people can
at least narrow it like in VsCode (though it will need dozens of similar lines).

Do we want the "glob-like" patterns in this PR or in a follow-up ?

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -50,6 +50,7 @@
 | `:lsp-workspace-command` | Open workspace command picker |
 | `:lsp-restart` | Restarts the Language Server that is in use by the current doc |
 | `:tree-sitter-scopes` | Display tree sitter scopes, primarily for theming and development. |
+| `:semantic-tokens` | Display the semantic tokens, primarily for theming and development. |
 | `:debug-start`, `:dbg` | Start a debug session from a given template with given parameters. |
 | `:debug-remote`, `:dbg-tcp` | Connect to a debug adapter by TCP address and start a debugging session from a given template with given parameters. |
 | `:debug-eval` | Evaluate expression in current debug context. |
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -10,13 +10,13 @@ use lsp::PositionEncodingKind;
 use lsp_types as lsp;
 use serde::Deserialize;
 use serde_json::Value;
-use std::collections::HashMap;
 use std::future::Future;
 use std::process::Stdio;
 use std::sync::{
     atomic::{AtomicU64, Ordering},
-    Arc,
+    Arc, RwLock,
 };
+use std::{collections::HashMap, sync::RwLockReadGuard};
 use tokio::{
     io::{BufReader, BufWriter},
     process::{Child, Command},
@@ -38,6 +38,9 @@ pub struct Client {
     root_uri: Option<lsp::Url>,
     workspace_folders: Vec<lsp::WorkspaceFolder>,
     req_timeout: u64,
+
+    pub(crate) semantic_token_types_legend: RwLock<Vec<Arc<String>>>,
+    pub(crate) semantic_token_modifiers_legend: RwLock<Vec<Arc<String>>>,
 }
 
 impl Client {
@@ -110,6 +113,9 @@ impl Client {
             root_path,
             root_uri,
             workspace_folders,
+
+            semantic_token_types_legend: Default::default(),
+            semantic_token_modifiers_legend: Default::default(),
         };
 
         Ok((client, server_rx, initialize_notify))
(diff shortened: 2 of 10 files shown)


poliorcetics:
This PR is a draft: I need to write docs, cleanup the code and resolve the issue explained in the first message, but if people want to try it out, it should work nicely, though you'll need to explore with your LSPs and read the code to see what to set in themes to see it working.

LouisGariepy:
There's historically been reluctance to implement LSP highlighting in Helix (for performance reasons IIUC), but this was proposed as a potential solution https://github.com/helix-editor/helix/discussions/5589#discussioncomment-4723984

> Tree-sitter will be the default highlighter, while range-based LSP syntax highlighting will be merged asynchronously.

Does this PR implement this functionality? Or has the Helix team's stance on LSP highlighting changed?

kirawi:
This PR does not implement that feature, and AFAIK the stance hasn't changed. It's uncertain what kind of discussion might happen around this PR with respect to that.

poliorcetics:
> This PR does not implement that feature 

Yes it does, this PR used the tree sitter coloring by default, only adding the LSP colors when it got them

Anyway, I'm not interested in working on helix right, I just forgot to close this one

kirawi:
Ah, good to know if you or anyone else picks this up then.

Eliemer:
F# for example relies pretty heavily on the LSP to distinguish variables from functions as the only way to figure that out is with types. They look the same in declaration.
```fs
let x: int = 10
let y: unit -> int = fun () -> 20
let z a b = a + b // z: 'a -> 'a -> 'a
```

kirawi:
I'll take a look at picking this up.

EricHenry:
@kirawi depending on how busy you are. I’d love to help out with this one if possible.

LouisGariepy:
I've been thinking about this a bit more, and I started having some doubts about the "start with tree-sitter then merge semantic highlighting asynchronously" approach.

Instead, why not simply have an option to toggle semantic highlighting? So people who only want tree-sitter don't pay for semantic highlighting, and people who only want semantic highlighting don't pay for tree-sitter?

In addition to avoid wasting compute power, I imagine this would also be easier to implement overall.

Thoughts @kirawi ?

LouisGariepy:
@pascalkuthe 

> we don't want multiple alternative syntax highlighting implementations in core.
> 
> If we are going to support semantic highlighting doingit asynchrounsly and only for a few selected scopes is the only thing we are going to accept.

Can you expand on why that is? IMO that should be a decision left to the user based on how much they value semantic highlighting and how fast their machine is.

I think such an entrenched position on the side of the project should at least have clearly stated technical reasons for doing so.

pascalkuthe:
> @pascalkuthe
> 
> > we don't want multiple alternative syntax highlighting implementations in core.
> > If we are going to support semantic highlighting doingit asynchrounsly and only for a few selected scopes is the only thing we are going to accept.
> 
> Can you expand on why that is? IMO that should be a decision left to the user based on how much they value semantic highlighting and how fast their machine is.
> 
> I think such an entrenched position on the side of the project should at least have clearly stated technical reasons.



As @archseer said recently on matrix about opinions like this in general: just "adding an option to let the user decide" when we have made a different design decision (in this case using tree sitter) is not something we accept usually. From the user perspective it just "lets them do what they want" but for us it means reviewing and maintaining a bunch of extra code. Code that lives in helix core should be up to a certain quality (including performance) so adding a bunch of different alternatives is a lot of work.

The technical reason in this case: its an inefficient/slow protocol and requires additional regex highlighting (which we also don't want to add/maintain) to avoid UI flickering. Tree sitter is used for many additional things inside helix besides syntax highlighting so it's a core editor component. We don't want to add/maintain alternatives to core editor components. 

This is documented in the original issue for semantic highlighting, by the comment from @archseer there.

metaspace:
> The technical reason in this case: its an inefficient/slow protocol and requires additional regex highlighting (which we also don't want to add/maintain) to avoid UI flickering. Tree sitter is used for many additional things inside helix besides syntax highlighting so it's a core editor component. We don't want to add/maintain alternatives to core editor components.

I really don't get the statement about UI flickering. Surely you pull in info from other sources such as cargo check and clippy asynchronously anyway?

For me, the ability to have some kind of highlighting of the variable under the cursor is extremely useful when reading code.

I get the maintenance burden argument. Perhaps ask that the contributor of such a feature also maintain it, and if they fail to do so, yank the feature :shrug:

PR #6103, opened 2023-02-25
LSP: Support textDocument/prepareRename

askreet:
@the-mikedavis here's the updated branch with fallback behavior in all the right places, I think. I used inner functions to factor our duplicate logic and actually think it came out quite nice. Seen it elsewhere in the codebase but curious if it's considered a "good" pattern we want to embrace.

'textDocument/prepareRename' can be used by the client to ask the server the range of the symbol under the cursor which would be changed by a subsequent call to 'textDocument/rename' with that position.

We can use this information to fill the prompt with an accurate prefill which can improve the UX for renaming symbols when the symbol doesn't align with the "word" textobject. (We currently use the "word" textobject as a default value for the prompt.)

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1232,49 +1232,116 @@ pub fn hover(cx: &mut Context) {
 }
 
 pub fn rename_symbol(cx: &mut Context) {
-    let (view, doc) = current_ref!(cx.editor);
-    let text = doc.text().slice(..);
-    let primary_selection = doc.selection(view.id).primary();
-    let prefill = if primary_selection.len() > 1 {
-        primary_selection
-    } else {
-        use helix_core::textobject::{textobject_word, TextObject};
-        textobject_word(text, primary_selection, TextObject::Inside, 1, false)
+    fn get_prefill_from_word_boundary(editor: &Editor) -> String {
+        let (view, doc) = current_ref!(editor);
+        let text = doc.text().slice(..);
+        let primary_selection = doc.selection(view.id).primary();
+        if primary_selection.len() > 1 {
+            primary_selection
+        } else {
+            use helix_core::textobject::{textobject_word, TextObject};
+            textobject_word(text, primary_selection, TextObject::Inside, 1, false)
+        }
+        .fragment(text)
+        .into()
     }
-    .fragment(text)
-    .into();
-    ui::prompt_with_input(
-        cx,
-        "rename-to:".into(),
-        prefill,
-        None,
-        ui::completers::none,
-        move |cx: &mut compositor::Context, input: &str, event: PromptEvent| {
-            if event != PromptEvent::Validate {
-                return;
+
+    fn get_prefill_from_lsp_response(
+        editor: &Editor,
+        offset_encoding: OffsetEncoding,
+        response: Option<lsp::PrepareRenameResponse>,
+    ) -> Result<String, String> {
+        match response {
+            Some(lsp::PrepareRenameResponse::Range(range)) => {
+                let text = doc!(editor).text();
+
+                Ok(lsp_range_to_range(text, range, offset_encoding)
+                    .ok_or_else(|| "lsp sent invalid selection range for rename".to_string())?
+                    .fragment(text.slice(..))
+                    .into())
             }
-
-            let (view, doc) = current!(cx.editor);
-            let language_server = language_server!(cx.editor, doc);
-            let offset_encoding = language_server.offset_encoding();
-
-            let pos = doc.position(view.id, offset_encoding);
-
-            let future =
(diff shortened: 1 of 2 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -1232,49 +1232,115 @@ pub fn hover(cx: &mut Context) {
...
+                    .fragment(text.slice(..))
+                    .into())
+            }
+            Some(lsp::PrepareRenameResponse::RangeWithPlaceholder { placeholder, .. }) => {
+                Ok(placeholder)
+            }
+            Some(lsp::PrepareRenameResponse::DefaultBehavior { .. }) => {
+                Ok(get_prefill_from_word_boundary(editor))
             }
+            None => Err("lsp did not response to prepare rename request".to_string()),
```
the-mikedavis, line 1269: This can be a `Result<String, &'static str>` since we know all of the error messages ahead of time:

```suggestion
    ) -> Result<String, &'static str> {
        match response {
            Some(lsp::PrepareRenameResponse::Range(range)) => {
                let text = doc!(editor).text();

                Ok(lsp_range_to_range(text, range, offset_encoding)
                    .ok_or_else(|| "lsp sent invalid selection range for rename")?
                    .fragment(text.slice(..))
                    .into())
            }
            Some(lsp::PrepareRenameResponse::RangeWithPlaceholder { placeholder, .. }) => {
                Ok(placeholder)
            }
            Some(lsp::PrepareRenameResponse::DefaultBehavior { .. }) => {
                Ok(get_prefill_from_word_boundary(editor))
            }
            None => Err("lsp did not response to prepare rename request"),
```

and `Editor::set_error` takes a `Into<Cow<'static, str>>` which fits `&'static str`s

helix-term/src/commands/lsp.rs
```diff
@@ -1232,49 +1232,115 @@ pub fn hover(cx: &mut Context) {
...
+    }
+
+    let (view, doc) = current!(cx.editor);
+    let language_server = language_server!(cx.editor, doc);
+    let offset_encoding = language_server.offset_encoding();
+
+    let pos = doc.position(view.id, offset_encoding);
+
+    match language_server.prepare_rename(doc.identifier(), pos) {
+        // Language server support textDocument/prepareRename, use it.
```
the-mikedavis, line 1316:
```suggestion
        // Language server supports textDocument/prepareRename, use it.
```

helix-term/src/commands/lsp.rs
```diff
@@ -1232,49 +1232,115 @@ pub fn hover(cx: &mut Context) {
...
+        Some(future) => cx.callback(
+            future,
+            move |editor, compositor, response: Option<lsp::PrepareRenameResponse>| {
+                let prefill = match get_prefill_from_lsp_response(editor, offset_encoding, response)
+                {
+                    Ok(p) => p,
+                    Err(e) => {
+                        editor.set_error(e);
```
the-mikedavis, line 1324: If we return early here we won't be able to try the rename. That might be ok: if `get_prefill_from_lsp_response` returns an `Err`, the language server either doesn't support renaming at the current position or it sent an invalid range which is a problem too.

Alternatively though we could fall back to the word textobject prefill value here. I'm not sure which is better behavior. Let's leave it as-is for now?

the-mikedavis (commented):
This is looking good 👍

Splitting up the code by using functions within `rename_symbol` makes sense to me.

helix-term/src/commands/lsp.rs
pascalkuthe: I don't have the time to investigate myself right now but it might be worth looking at what vscode does here

the-mikedavis: I couldn't find exactly where this is handled in vscode but it looks like the current behavior here matches nvim's https://github.com/neovim/neovim/blob/ca3bc56a3b1b3454498a4a23c0d700486d554077/runtime/lua/vim/lsp/buf.lua#L319-L379

helix-term/src/commands/lsp.rs
```diff
@@ -1232,49 +1232,115 @@ pub fn hover(cx: &mut Context) {
...
+                    .fragment(text.slice(..))
+                    .into())
+            }
+            Some(lsp::PrepareRenameResponse::RangeWithPlaceholder { placeholder, .. }) => {
+                Ok(placeholder)
+            }
+            Some(lsp::PrepareRenameResponse::DefaultBehavior { .. }) => {
+                Ok(get_prefill_from_word_boundary(editor))
             }
+            None => Err("lsp did not response to prepare rename request"),
```
the-mikedavis, line 1269:
```suggestion
            None => Err("lsp did not respond to prepare rename request"),
```

the-mikedavis (approved):

PR #6105, opened 2023-02-25
Rename the json server executable

alpendop:
Installing the language server from NPM has the name set to `vscode-json-languageserver`

Currently, it needs to be edited manually in `languages.toml` file for it to be recognised.


kirawi:
That is a different server than the one in https://github.com/hrsh7th/vscode-langservers-extracted

alpendop:
They still seem to be updated though.

If that's the case, I'm gonna close this issue.

PR #6107, opened 2023-02-26
Fix diagnostic underline colors in catppuccin themes

adam-becker:
The current behavior sets the foreground color, and this is very difficult to read. Instead, only the underline itself should be a different color.

### DIFF
--- a/runtime/themes/catppuccin_mocha.toml
+++ b/runtime/themes/catppuccin_mocha.toml
@@ -103,10 +103,10 @@
 "ui.menu" = { fg = "overlay2", bg = "surface0" }
 "ui.menu.selected" = { fg = "text", bg = "surface1", modifiers = ["bold"] }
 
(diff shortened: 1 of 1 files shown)


leonqadirie:
Hi, I like the change.

Do you know the [official upstream repo](https://github.com/catppuccin/helix)? We should try to keep in sync.

adam-becker:
@leonqadirie It seems the files in this repo are already different (usage of inheritance over templates), is there a better way to keep them in sync? Perhaps a submodule of the official repo that can be fast-forwarded?

adam-becker:
Just had a look at the upstream repo, it seems they [already have this fix](https://github.com/catppuccin/helix/blob/main/data/template.tmpl#L102-L105) (slightly different syntax, but the same idea)

the-mikedavis (approved):

PR #6109, opened 2023-02-26
Fix lacking space panic

nuid32:
Reincarnation of #5529.
I apologize for the mess around this PR and thank you all for your patience.


helix-term/src/ui/markdown.rs
```diff
@@ -342,9 +342,6 @@ impl Component for Markdown {
 
     fn required_size(&mut self, viewport: (u16, u16)) -> Option<(u16, u16)> {
         let padding = 2;
-        if padding >= viewport.1 || padding >= viewport.0 {
-            return None;
-        }
```
pascalkuthe, line 347: This is a footgun right now. Any popup that returns `None` in it's required size implementation leads to a crash.
I wonder if we should just change the signature to return `(u16, u16)` and return `(0, 0) by default. There is no code that actually relies on `required_size` being optional and logically if `required_size` is `None` I would assume that a component requires no size. 

That being said jsut removing this early bail here is not correct as the `viewport.0 - padding` subtractiong would underflow. In my opionion the correct thing to do is change `(viewport.0 - padding)` to `viewport.0.saturation_sub(padding)`. From the `required_size` documentation:

> The returned size might be larger than the viewport if the child is too big to fit.
> In this case the parent can use the values to calculate scroll.

pascalkuthe (changes requested):

helix-view/src/graphics.rs
```diff
@@ -237,8 +237,8 @@ impl Rect {
         Rect {
             x: x1,
             y: y1,
-            width: x2 - x1,
-            height: y2 - y1,
+            width: x2.wrapping_sub(x1),
+            height: y2.wrapping_sub(y1),
```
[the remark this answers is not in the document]
nuid32: Yes, it definitely would. I confused wrapping and saturating substraction one more time.

pascalkuthe (dismissed):
I think the additional saturating subs and defensive clear are definitly correct fixes so this LGTM.

I am not entirely sure if we want to stop returning an optiom `required_size` to avoid similar crashes in the future and remove the footgun. That's mostly a code clarity issue tough and can be a separate PR

helix-tui/src/buffer.rs
```diff
@@ -510,9 +510,13 @@ impl Buffer {
     pub fn clear_with(&mut self, area: Rect, style: Style) {
         for x in area.left()..area.right() {
             for y in area.top()..area.bottom() {
-                let cell = &mut self[(x, y)];
-                cell.reset();
-                cell.set_style(style);
+                if let Some(cell) = &mut self.get_mut(x, y) {
```
the-mikedavis, line 513:
```suggestion
                if let Some(cell) = self.get_mut(x, y) {
```

I think this is redundant because of the get_mut

the-mikedavis (approved):

helix-tui/src/buffer.rs
```diff
@@ -510,9 +510,13 @@ impl Buffer {
...
-                let cell = &mut self[(x, y)];
-                cell.reset();
-                cell.set_style(style);
+                if let Some(cell) = self.get_mut(x, y) {
+                    cell.reset();
+                    cell.set_style(style);
+                } else {
+                    // Skip this render if the window size has changed
+                    continue;
+                }
```
archseer, line 519: I'm a bit wary about changing this: the window size won't change halfway through a render. If `clear_with` is called with an out of bounds rectangle then that's a bug that should be addressed and skipping it here would hide the bug.

archseer:
> I am not entirely sure if we want to stop returning an optiom `required_size` to avoid similar crashes in the future

The idea there is that `required_size` is a hint to the parent component on how much space is needed. Similar to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

helix-tui/src/buffer.rs
nuid32: I had a problem with that. Why window can not change it size during render?

pascalkuthe:
> > I am not entirely sure if we want to stop returning an optiom `required_size` to avoid similar crashes in the future
> 
> The idea there is that `required_size` is a hint to the parent component on how much space is needed. Similar to https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

Hmm but size hint also always returns a lower bound and only the upper bound is optional. The value is optional because allocating a larger value is only a performance optimization and never causes a panic.

If I read the code right `required_size` is not actually treated like the upper bound of `size_hint`. It's treated like the exact size used by the component which would be equivalent to `size_hint` returning `(size Some(size))` or the lower bound (which is also not an `Option` for `size_hint`). Right now I don't really see why a component would ever return `None` from `required_size` as all called of `required_size` unwrap the return value.

helix-tui/src/buffer.rs
nuid32: Thank you for the detailed answer. `clear_with` is not related with crashes I had indeed.

nuid32:
Well now I don't think any bound-check is redundant.
```
thread 'main' panicked at 'index out of bounds: the len is 8 but the index is 8', helix-tui/src/buffer.rs:476:17
stack backtrace:
   0: rust_begin_unwind
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
   1: core::panicking::panic_fmt
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
             at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:147:5
   3: helix_tui::buffer::Buffer::set_style
   4: <F as helix_term::ui::document::LineDecoration>::render_background
   5: helix_term::ui::document::render_document
   6: helix_term::ui::editor::EditorView::render_view
   7: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
   8: helix_term::application::Application::render::{{closure}}
   9: helix_term::application::Application::run::{{closure}}
  11: tokio::runtime::runtime::Runtime::block_on
  12: hx::main
```
Here's `set_style` function:
```
pub fn set_style(&mut self, area: Rect, style: Style) {
    for y in area.top()..area.bottom() {
        for x in area.left()..area.right() {
            self[(x, y)].set_style(style);
        }
    }
}
```

I think bound-check won't harm readability more than it's absence harms sustainability.

pascalkuthe:
> Well now I don't think any bound-check is redundant.
> 
> ```
> thread 'main' panicked at 'index out of bounds: the len is 8 but the index is 8', helix-tui/src/buffer.rs:476:17
> stack backtrace:
>    0: rust_begin_unwind
>              at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/std/src/panicking.rs:575:5
>              at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:64:14
>    2: core::panicking::panic_bounds_check
>              at /rustc/d5a82bbd26e1ad8b7401f6a718a9c57c96905483/library/core/src/panicking.rs:147:5
>    3: helix_tui::buffer::Buffer::set_style
>    4: <F as helix_term::ui::document::LineDecoration>::render_background
>    5: helix_term::ui::document::render_document
>    6: helix_term::ui::editor::EditorView::render_view
>    7: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::render
>    8: helix_term::application::Application::render::{{closure}}
>    9: helix_term::application::Application::run::{{closure}}
>   10: tokio::runtime::park::CachedParkThread::block_on
>   11: tokio::runtime::runtime::Runtime::block_on
>   12: hx::main
> ```
> 
> Here's `set_style` function:
> 
> ```
> pub fn set_style(&mut self, area: Rect, style: Style) {
>     for y in area.top()..area.bottom() {
>         for x in area.left()..area.right() {
>             self[(x, y)].set_style(style);
>         }
>     }
> }
> ```
> 
> I think bound-check won't harm readability more than it's absence harms sustainability.

This is a completly separate bug, totally unrelated to the issue this PR is adressing, It looks like it might be caused caused by a regression in #5420. Please open a separate issue for that. While it *might* be ok to ignore `clear_with` calls with an out of bounds rect, `set_style` is actually used for rendering and should under no circumstance allow any out of bounds positions. Anytime `set_style` is called with an out of bounds rect there is a bug which would manifest in rendering issues. We should always panic in that case  otherwise debugging rendering issues basically becomes extremely hard. Removing the panic here is really just a bandage and not the real fix.

Please remove the changes to `set_style` and `clear_with` so we can merge the other two (correct) fixes

nuid32:
I don't understand why `set_style` call with an out of bounds `Rect` is ok. Isn't it a bug itself?

pascalkuthe:
> I don't understand why `set_style` call with an out of bounds `Rect` is ok. Isn't it a bug itself?

No it's not ok it is definitely a bug that is why it **should** panic. The idea is to fail fast and fail hard. If `set_style` just silently ignores out of bounds positions we would never be able to find those kinds bugs. The bug is not that `set_style` panics, the bug is that `set_style` is called with an out of bounds position. To fix this problem the callsite should be fixed to not call `set_style` incorrectly instead of making `set_style` ignore an incorrect argument.

nuid32:
Maybe it caused by the window size changing during `clear_with`, but it's strange helix not crashes during rendering, so I guess you're right.

pascalkuthe:
As I said earlier it's **impossible for the buffer size to change during rendering** , the cause of the crash is a bug somewhere else in the code that should be fixed. If you can find out what caused the crash in `set_style` above I am happy to look into a fix. Panics make it a lot easier to find the actual causes of bugs.

nuid32:
People can lose unsaved changes due to panic, aren't there ways to keep helix sustainable while not hiding information for debug? Maybe debug code which will be executed in debug build?

nuid32:
> Maybe it caused by the window size changing during `clear_with`, but it's strange helix not crashes during rendering, so I guess you're right.

Wait, it does.

pascalkuthe:
> People can lose unsaved changes due to panic, aren't there ways to keep helix sustainable while not hiding information for debug? Maybe debug code which will be executed in debug build?



That's what `debug_assert` does. It still makes bug reports a lot harder to track down (often I can guess a bug and fix just from the panic back-trace without a reproduction case which most users can't/won't provide). I think it's ok to only use `debug_assert` in `clear_with` because there are no adverse effects to not clear something off screen (that's why I suggested `debug_assert` in  https://github.com/helix-editor/helix/pull/6109#discussion_r1122264604) but for `set_style` this would lead to visual artifacts so in that case a crash is preferable so we can quickly identify the problems

helix-tui/src/buffer.rs
```diff
@@ -510,9 +510,10 @@ impl Buffer {
     pub fn clear_with(&mut self, area: Rect, style: Style) {
         for x in area.left()..area.right() {
             for y in area.top()..area.bottom() {
-                let cell = &mut self[(x, y)];
-                cell.reset();
-                cell.set_style(style);
+                if let Some(cell) = self.get_mut(x, y) {
+                    cell.reset();
+                    cell.set_style(style);
+                }
```
archseer, line 516: This should still be reverted or we're still hiding the bug

archseer (changes requested):

archseer:
> That's what `debug_assert` does.

Very few people actually use debug builds outside of CI/integration tests yeah. Even locally I only run release builds

pascalkuthe:
> > That's what `debug_assert` does.
> 
> Very few people actually use debug builds outside of CI/integration tests yeah. Even locally I only run release builds

Hmm yeah it does make bug reports harder to track down (see second sentence). I suggested this because we use `debug_assert` in quite a few other places to assert an argument is correct. Maybe we want to make some of those `debug_assert` normal `assert` too.

pascalkuthe (approved):

PR #6119, opened 2023-02-27
Add Objective-C language support

steinuil:
I only pulled the tree-sitter grammar from https://github.com/jiyee/tree-sitter-objc/ and pasted the highlight.scm file from that repository, so there might be things to improve.

### DIFF
--- a/runtime/queries/objc/highlights.scm
+++ b/runtime/queries/objc/highlights.scm
@@ -1,7 +1,83 @@
 [
-  (comment)
-  (pragma)
-] @comment
+  "sizeof"
+  "typeof"
+  "__typeof"
+  "__typeof__"
+  "@defs"
+  "@selector"
+  "@encode"
+] @keyword.operator
+
+[
+  "enum"
+  "struct"
+  "typedef"
+  "union"
+  "@interface"
+  "@protocol"
+  "@property"
+  "@synthesize"
+  "@implementation"
+  "@class"
+  "@dynamic"
+  "@compatibility_alias"
+] @keyword.storage.type
+
+[
+  "const"
+  "inline"
+  "static"
+  "extern"
+  "register"
+  "volatile"
+  "_Atomic"
+  "volatile"
+  "NS_NOESCAPE"
+  "NS_ENUM"
+  "NS_ERROR_ENUM"
+  "NS_OPTIONS"
+  "NS_SWIFT_NAME"
+  "__covariant"
+  "__contravariant"
+  "__GENERICS"
+  (private)
+  (public)
+  (protected)
+  (package)
+  (optional)
+  (required)
+  (type_qualifier)
+  (storage_class_specifier)
+] @keyword.storage.modifier
+
+[
+  "goto"
+  "break"
+  "continue"
+] @keyword.control
+
+[
+  "do"
+  "for"
+  "while"
+] @keyword.control.repeat
+
+[
+  "if"
+  "else"
+  "switch"
+  "case"
+  "default"
+] @keyword.control.conditional
+ 
+[
+  "@try"
+  "@catch"
+  "@finally"
+  "@throw"
+] @keyword.control.exception
+
+(sizeof_expression value: (parenthesized_expression (identifier) @type))
 
 [
   (self)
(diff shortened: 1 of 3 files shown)


runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+  "volatile"
+  "goto"
+  "register"
+  "__covariant"
+  "__contravariant"
+  "__GENERICS"
+] @keyword
+
+"sizeof" @keyword.operator
+"return" @keyword.return
```
the-mikedavis, line 88:
```suggestion
"return" @keyword.control.return
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+
+[
+  "while"
+  "for"
+  "do"
+  "continue"
+  "break"
+] @keyword.repeat
```
the-mikedavis, line 96:
```suggestion
] @keyword.control.repeat
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+
+[
+  "#if"
+  "#ifdef"
+  "#ifndef"
+  "#else"
+  "#elif"
+  "#endif"
+  (preproc_directive)
+] @keyword
```
the-mikedavis, line 108:
```suggestion
"#define" @keyword.directive

[
  "#if"
  "#ifdef"
  "#ifndef"
  "#else"
  "#elif"
  "#endif"
  (preproc_directive)
] @keyword.directive
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+
+"#include" @include
+"#import" @include
+"@import" @include
```
the-mikedavis, line 112:
```suggestion
"#include" @keyword.include
"#import" @keyword.include
"@import" @keyword.include
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+] @operator
+
+[
+ "if"
+ "else"
+ "case"
+ "switch"
+] @keyword.conditional
+
+(conditional_expression [ "?" ":" ] @keyword.conditional)
```
the-mikedavis, line 165:
```suggestion
] @keyword.control.conditional

(conditional_expression [ "?" ":" ] @keyword.control.conditional)
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+
+[
+ (true)
+ (false)
+ (YES)
+ (NO)
+] @keyword.boolean
```
the-mikedavis, line 172:
```suggestion
] @constant.builtin.boolean
```

runtime/queries/objc/highlights.scm
```diff
@@ -0,0 +1,395 @@
...
+[ "(" ")" "[" "]" "{" "}"] @punctuation.bracket
+
+[
+  (string_literal)
+  (string_expression)
+  (system_lib_string)
+  (module_string)
+] @string
+
+(escape_sequence) @string.escape
```
the-mikedavis, line 187:
```suggestion
(escape_sequence) @constant.character.escape
```

the-mikedavis (commented):
We use some custom captures documented here: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

I've left some example changes as review comments but there are more below L190 that should be changed to use the Helix captures instead too

kasperisager:
This would be a fantastic addition! @steinuil, do you need a hand with any of this?

steinuil:
Hi @kasperisager, yes gladly! I know very little Objective-C so I had a hard time figuring out how to categorize the keywords, I'll push what I have so if you want you can start with that.

kasperisager:
@steinuil I've just committed and pushed the grammars and queries I've been using for Obj-C and Obj-C++, also based on https://github.com/jiyee/tree-sitter-objc: https://github.com/kasperisager/dotfiles/commit/0193bb537323a4fe40feb379f19e2d54beb81794.

@the-mikedavis I initially tried making the Obj-C queries inherit from the plain C queries, which unfortunately didn't work as Helix choked on certain queries, such as `(attribute name: (identifier) @attribute)`. Could this be because the Obj-C tree sitter parser doesn't recognise `attribute` nodes?

pascalkuthe:
grammer are not usually related to each other and just using queries from one grammar for another unrelated language doesn't work

pascalkuthe:
I am closing this PR PR as stale. Te captures were just copied from upstream and not adapted to helix. If somebody wants to add objective-c language support to helix they should do that and open a new PR. Thank you for contributing

PR #6126, opened 2023-02-27
feat(prql): add prql support

### LINKED ISSUE #6050: PRQL support
Support for the [PRQL](https://prql-lang.org/) language would be great!

There is an issue for a tree-sitter grammar in the PRQL repo:
https://github.com/PRQL/prql/issues/1882

with an associated start of an implementation over here:
https://github.com/matthias-Q/tree-sitter-prql

matthias-Q:
Adds support for PRQL languange. 

I hope I added everything thats necessary to add language support. 

Closes #6050

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2210,7 +2210,7 @@ source = { git = "https://github.com/Philipp-M/tree-sitter-yuck", rev = "9e97da5
 
 [[language]]
 name = "prql"
-scope = "source.yprql"
+scope = "source.prql"
 injection-regex = "prql"
 file-types = ["prql"]
 roots = []
--- a/languages.toml
+++ b/languages.toml
@@ -2209,6 +2209,19 @@ name = "yuck"
 source = { git = "https://github.com/Philipp-M/tree-sitter-yuck", rev = "9e97da5773f82123a8c8cccf8f7e795d140ed7d1" }
 
 [[language]]
+name = "prql"
+scope = "source.yprql"
+injection-regex = "prql"
+file-types = ["prql"]
+roots = []
+comment-token = "#"
+indent = { tab-width = 4, unit = "    " }
+
+[[grammar]]
+name = "prql"
+source = { git = "https://github.com/PRQL/tree-sitter-prql", rev = "3f27cac466f030ee7d985d91eba5470e01dd21ea" }
+
+[[language]]
 name = "po"
 scope = "source.po"
 file-types = ["po", "pot"]
--- a/runtime/queries/prql/highlights.scm
+++ b/runtime/queries/prql/highlights.scm
@@ -28,9 +28,9 @@
 (literal) @string
 
 (assignment
-  alias: (field) @field)
+  alias: (field) @variable.other.member)
 
-alias: (identifier) @field
+alias: (identifier) @variable.other.member
 
 (f_string) @string.special
 (s_string) @string.special
@@ -72,14 +72,14 @@ alias: (identifier) @field
 ] @punctuation.delimiter
 
 (literal
-  (integer) @number)
+  (integer) @constant.numeric.integer)
 
-(integer) @number
+(integer) @constant.numeric.integer
 
 (literal
-  (decimal_number) @float)
+  (decimal_number) @constant.numeric.float)
 
-(decimal_number) @float
+(decimal_number) @constant.numeric.float
 
 [
   (keyword_min)
--- a/runtime/queries/prql/highlights.scm
+++ b/runtime/queries/prql/highlights.scm
@@ -1,0 +1,140 @@
+[
+  (keyword_from)
+  (keyword_filter)
+  (keyword_derive)
+  (keyword_group)
+  (keyword_aggregate)
+  (keyword_sort)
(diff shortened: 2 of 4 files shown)


matthias-Q:
Sorry, I have not run `cargo xtasks docgen` I will do that and update the PR

erasin:
> The only difference is that prql should be a few lines lower. should these changes be pushed?

Since `book/src/generated` is an automatically generated file, use `cargo xtasks docgen` to handle it to avoid conflicts.

I think a normal commit will do the trick. 

Alternatively Git provides some rebase and commit merge methods, so you can use `git commit --amend` to merge some changes directly into the last commit, and then push with `--force` to overwrite remote branch.

languages.toml
```diff
@@ -2208,6 +2208,19 @@ indent = { tab-width = 2, unit = "  " }
 name = "yuck"
 source = { git = "https://github.com/Philipp-M/tree-sitter-yuck", rev = "9e97da5773f82123a8c8cccf8f7e795d140ed7d1" }
 
+[[language]]
+name = "prql"
+scope = "source.yprql"
```
the-mikedavis, line 2213: Is this a typo or is the "y" intentional?

runtime/queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+  (keyword_remove)
+  (keyword_intersect)
+] @keyword
+
+(literal) @string
+
+(assignment
+  alias: (field) @field)
+
+alias: (identifier) @field
```
the-mikedavis, line 33:
```suggestion
  alias: (field) @variable.other.member)

alias: (identifier) @variable.other.member
```

runtime/queries/prql/highlights.scm
```diff
@@ -0,0 +1,140 @@
...
+
+(literal
+  (integer) @number)
+
+(integer) @number
+
+(literal
+  (decimal_number) @float)
+
+(decimal_number) @float
```
the-mikedavis, line 82:
```suggestion
(literal
  (integer) @constant.numeric.integer)

(integer) @constant.numeric.integer

(literal
  (decimal_number) @constant.numeric.float)

(decimal_number) @constant.numeric.float
```

We use slightly different scopes than nvim. There are some below that should be switched to the Helix scopes like `@method`, `@boolean` and `@parameter`
https://docs.helix-editor.com/master/themes.html#syntax-highlighting

runtime/queries/prql/injections.scm
```diff
@@ -0,0 +1,11 @@
...
+ (s_string) @sql
+ (#offset! @sql 0 2 0 -1)
+)
+
+(from_text
+  (keyword_from_text)
+  (keyword_json)
+  (literal) @json
+  (#offset! @json 0 3 0 -3)
+)
```
the-mikedavis, line 11:
```suggestion
((s_string) @injection.content
 (#set! injection.language "sql"))

(from_text
  (keyword_from_text)
  (keyword_json)
  (literal) @injection.content
  (#set! injection.language "json"))
```

We use a slightly different syntax for injections as well: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection

languages.toml
matthias-Q: Typo, will fix that later

runtime/queries/prql/highlights.scm
matthias-Q: I guess in line 31, the last bracket should still close after `other.member` right?

runtime/queries/prql/highlights.scm
matthias-Q: Thanks, I tried to update them. I hope I catched everything

runtime/queries/prql/highlights.scm
the-mikedavis: Yep right, I deleted too much in this suggestion 😅

runtime/queries/prql/highlights.scm
```diff
@@ -0,0 +1,136 @@
...
+
+(f_string) @string.special
+(s_string) @string.special
+
+(comment) @comment
+
+(keyword_func) @keyword.function
+
+(function_call
+  (identifier) @function.call)
```
the-mikedavis, line 39:
```suggestion
  (identifier) @function)
```

runtime/queries/prql/highlights.scm
```diff
@@ -0,0 +1,136 @@
...
+  (keyword_min)
+  (keyword_max)
+  (keyword_count)
+  (keyword_count_distinct)
+  (keyword_average)
+  (keyword_avg)
+  (keyword_sum)
+  (keyword_stddev)
+  (keyword_count)
+] @function.call
```
the-mikedavis, line 90:
```suggestion
] @function
```

the-mikedavis (commented):
Otherwise this looks good

(Also see https://github.com/helix-editor/helix/pull/4892 which talks about separating the highlights for function definitions and references)

the-mikedavis (approved):

the-mikedavis:
Thanks!

PR #6142, opened 2023-02-28
Fix scrolloff at view bottom

### LINKED ISSUE #6141: String search jumps off-screen
### Summary

I just built a new version of Helix at HEAD (https://github.com/helix-editor/helix/commit/8dab8a0a039fe1f3dd98fc62ac97d2f1c089793a) and the behavior of search (i.e. `/` then `n`) has broken such that the cursor and entry are always one-line off the bottom of the screen. I don't know what the previous behavior was (e.g. center of screen, bottom of screen), but it definitely wasn't this.

As a workaround I just type `nzz` to center the (otherwise not-visible) result.

### Reproduction Steps

Should be readily reproducible, just search for a known string outside the current viewport.

### Helix log

Not including this for now, can include if deemed helpful

### Platform

Linux

### Terminal Emulator

alacritty / tmux

### Helix Version

helix 22.12 (8dab8a0a)

pascalkuthe:
Fixes #6141

Fixes a regression introduced in #5420 where a scrolloff of `x - 1` was used instead of `x` at the bottom of the screen (but not at the top). This was especially problematic if the scrolloff was set to `0` in that case the scrolloff behaved as tough set to `-1` and the cursor disappeared from the view if scrolled to the bottom (see #6141).

### DIFF
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -242,7 +242,7 @@ impl View {
                 at_top = true;
                 true
             }
-            Some((visual_pos, _)) if visual_pos.row >= vertical_viewport_end - scrolloff => {
+            Some((visual_pos, _)) if visual_pos.row + scrolloff + 1 >= vertical_viewport_end => {
                 if CENTERING && visual_pos.row >= vertical_viewport_end {
                     // cursor out of view
                     return None;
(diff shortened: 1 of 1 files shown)


gibbz00:
Possibly add a test for this?

the-mikedavis:
I think this would be more tricky to test than it is worth at the moment - we don't have tests that look at the rendering output currently

the-mikedavis (approved):

PR #6143, opened 2023-02-28
Extend the set of tags highlighted in comments

Isti115:


runtime/queries/comment/highlights.scm
```diff
@@ -5,17 +5,33 @@
...
+; Hint level tags
+((tag (name) @hint)
+ (#match? @hint "^(HINT|MARK)$"))
+
+("text" @hint
+ (#match? @hint "^(HINT|MARK)$"))
+
+; Info level tags
+((tag (name) @info)
+ (#match? @info "^(INFO|NOTE|TODO)$"))
```
the-mikedavis, line 17: Changing "TODO" from `@warning` to `@info` is potentially controversial - I know that some tools or communities lint against TODOs and treat them like warnings/errors. I suppose that you can use linters for that anyway so let's give this a try and revisit it if the change is problematic 👍

the-mikedavis (approved):

PR #6156, opened 2023-03-02
Integration test helper improvements

dead10ck:
- reorganize write command tests to their own module
- make `TestCase::From` more generic for ergonomics
- print doc state during tests for much better debugging
- To help with ergonomics, `test::print` was changed to allow explicit newlines in test DSL. The current test DSL currently has no way to express being at the end of a line, save for putting an explicit LF or CRLF inside the `#[|]#`. The problem with this approach is that it can add unintended extra new lines if used in conjunction with raw strings, which insert newlines for you.

  This is a simple attempt to mitigate this problem. If there is an explicit newline character at the end of the selection, and then it is immediately followed by the same newline character at the right end of the selection, this following newline is removed. This way, one can express a cursor at the end of a line explicitly.
- fix a bug `test::plain`: it is currently using char indices for `String::insert_str` which is supposed to take byte indices. This causes panics on Unicode characters
- use `AppBuilder` more consistently throughout tests to make use of saner defaults and improve readability


archseer (approved):
LGTM, just a Cargo.lock conflict to resolve

helix-core/src/test.rs
```diff
@@ -147,6 +178,7 @@ pub fn plain(s: &str, selection: Selection) -> String {
     out
 }
 
+#[allow(clippy::module_inception)]
 #[cfg(test)]
 #[allow(clippy::module_inception)]
```
the-mikedavis, line 183: Looks like this `allow` is duplicated

dead10ck: Oh weird, that must have been a bad rebase. Will fix. Good catch

PR #6160, opened 2023-03-02
tui: Only query keyboard enhancement support once

### LINKED ISSUE #6149: Weird behavior when forking helix to background
### Summary

Helix behaves weird when forking the process to the background (`Ctrl+Z`) and sending it to the foreground (`fg`) on Linux on latest git version.

### Reproduction Steps

 

I tried this:

1. `hx`
2. `Ctrl+Z`
3. `fg`

I expected this to happen:
It should fork the helix process to the background when I `Ctrl+Z` and resume the process when I send it to the foreground with `fg`.

Instead, this happened:
When trying to fork to the background, nothing happens the first time. The second time it works as expected.

When trying to send it to the foreground, it clears the terminal window on the first attempt. When trying again it brings the editor back, but it takes a second or two for the TUI to display.

If I quit helix after this, `^[[?6c` is present in my terminal after my terminal prompt, which seems to have been produced after the second attempt to resume the process.

### Helix log

Nothing since I updated.

### Platform

Arch Linux

### Terminal Emulator

Alacritty

### Helix Version

helix 22.12 (c082ef28)

the-mikedavis:
We query the host terminal emulator to check whether it supports the keyboard enhancement protocol after entering raw mode on startup. We only need to check this value once though, so we can store the result in a OnceCell to prevent repeated queries. The query for keyboard enhancement support times out and fails when suspending (C-z) and resuming (`fg`) which can cause delays and leaves sequences in the terminal after Helix suspends or exits. Caching in the OnceCell prevents this behavior.

Fixes #6149

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -112,12 +112,10 @@ fn restore_term() -> Result<(), Error> {
     let mut stdout = stdout();
     // reset cursor shape
     write!(stdout, "\x1B[0 q")?;
-    if matches!(terminal::supports_keyboard_enhancement(), Ok(true)) {
-        execute!(stdout, PopKeyboardEnhancementFlags)?;
-    }
     // Ignore errors on disabling, this might trigger on windows if we call
     // disable without calling enable previously
     let _ = execute!(stdout, DisableMouseCapture);
+    let _ = execute!(stdout, PopKeyboardEnhancementFlags);
     execute!(
         stdout,
         DisableBracketedPaste,
@@ -1071,7 +1069,10 @@ impl Application {
         if self.config.load().editor.mouse {
             execute!(stdout, EnableMouseCapture)?;
         }
-        if matches!(terminal::supports_keyboard_enhancement(), Ok(true)) {
+        if matches!(
+            self.terminal.supports_keyboard_enhancement_protocol(),
+            Ok(true)
+        ) {
             log::debug!("The enhanced keyboard protocol is supported on this terminal");
             execute!(
                 stdout,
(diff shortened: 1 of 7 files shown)


helix-term/src/application.rs
```diff
@@ -112,12 +112,10 @@ fn restore_term() -> Result<(), Error> {
     let mut stdout = stdout();
     // reset cursor shape
     write!(stdout, "\x1B[0 q")?;
-    if matches!(terminal::supports_keyboard_enhancement(), Ok(true)) {
-        execute!(stdout, PopKeyboardEnhancementFlags)?;
-    }
     // Ignore errors on disabling, this might trigger on windows if we call
     // disable without calling enable previously
     let _ = execute!(stdout, DisableMouseCapture);
+    let _ = execute!(stdout, PopKeyboardEnhancementFlags);
```
the-mikedavis, line 118: Ideally we would check `terminal.supports_keyboard_enhancement_protocol()` here but it's complicated to pass in values from Application because this function is set as a panic handler. We might be able to pass in just the boolean for whether keyboard enhancement is supported. I will check on some different terminals to see if it works to just send this `PopKeyboardEnhancementFlags` always though

helix-tui/src/backend/crossterm.rs
```diff
@@ -62,6 +64,7 @@ where
         CrosstermBackend {
             buffer,
             capabilities: Capabilities::from_env_or_default(),
+            supports_keyboard_enhancement_protocol: OnceCell::new(),
```
the-mikedavis, line 67: This value is not included in Capabilities because the tty needs to be in raw mode to perform the query and the backend is created before raw mode is enabled

helix-term/src/application.rs
the-mikedavis: I gave this a test with Kitty, WezTerm (with enhanced keyboard protocol disabled), Alacritty and the default login terminal for my linux machine and it seems to not make a difference - the terminal ignores the escape sequence when it doesn't recognize it

the-mikedavis:
Superseded by #6194 which should make it easier to add other queries in the future

Issue #3914, opened 2022-09-20
Keymap documentation has insufficient contrast when printed


jhbrown94:
When printing the documentation at https://docs.helix-editor.com/print.html using the printer icon in the upper right, the keymap keys have inadequate contrast.  You can see a bit of it in this screenshot of the generated PDF, but the contrast is much lower/worse when actually printed on paper using a typical laserprinter (I have tried a Canon and an HP).  It would be great to get the key names and command bindings to show up better.

<img width="166" alt="image" src="https://user-images.githubusercontent.com/7493750/191292303-cc803407-06fb-42b7-82e8-0cedc8985676.png">

jhbrown94:
Also -- some, but not all, material in the dark-background-boxes changes print-presentation depending on selected display theme.  For instance, with the "light" theme selected, printed Register information disappears entirely:

<img width="216" alt="image" src="https://user-images.githubusercontent.com/7493750/191293261-10947c72-d8a0-486a-9a56-9dfd9ae6836e.png">

PR #6168, opened 2023-03-03
book: fix code highlighting in light theme

jorpic:
ref. #3914

Inline code is not readable in nested lists and in table headers.

image

The offending CSS rule is [this one](https://github.com/helix-editor/helix/blob/ddc5bf4e606230dd8bc20c59b1cc114e93bbf1c0/book/theme/css/chrome.css#L199-L202):
```css
:not(pre):not(a):not(td):not(p) > .hljs {
    color: var(--inline-code-color);
    overflow-x: initial;
}
```

`--inline-code-color` is too dark and blends with the background. This variable was changed in https://github.com/helix-editor/helix/commit/4260b31ec059ffa2207f8f2541af1c3996b00cf9. The CSS rule was extended with `:not(td):not(p)` to fix blending in `p` and `td`.

But now inline code occurs in `li` and `th` also. It is possible to extend the with `:not(th):not(li)` but this will affect other color themes and potentially will make them less colorful. E.g. in `aui` orange here becomes grey:
image

Seems that the easiest solution that fixes readability of the light theme and does not affect other themes is to revert `--inline-code-color` to its original value.

### DIFF
--- a/book/theme/css/variables.css
+++ b/book/theme/css/variables.css
@@ -131,7 +131,7 @@
 
     --links: #20609f;
 
-    --inline-code-color: #301900;
+    --inline-code-color: #a39e9b;
 
     --theme-popup-bg: #fafafa;
     --theme-popup-border: #cccccc;


the-mikedavis:
This has already been fixed on master. https://docs.helix-editor.com shows the latest release notes while https://docs.helix-editor.com/master follows the latest master

PR #6170, opened 2023-03-03
fix: Handle signals before crossterm events

AlexanderBrevig:
Since https://github.com/helix-editor/helix/commit/79bf5e3094e16a34637703b14c7bf090d2dcf155 `C-z` and `fg` has taken 2 seconds for me. This small change fixes that.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -345,11 +345,11 @@ impl Application {
             tokio::select! {
                 biased;
 
-                Some(event) = input_stream.next() => {
-                    self.handle_terminal_events(event).await;
-                }
                 Some(signal) = self.signals.next() => {
                     self.handle_signals(signal).await;
+                }
+                Some(event) = input_stream.next() => {
+                    self.handle_terminal_events(event).await;
                 }
                 Some(callback) = self.jobs.futures.next() => {
                     self.jobs.handle_callback(&mut self.editor, &mut self.compositor, callback);


poliorcetics:
Good catch !

poliorcetics (changes requested):
A comment explaining this for future edits would be nice

AlexanderBrevig:
> A comment explaining this for future edits would be nice

Agreed, but I'm embarrassed to say I'm not quite sure. 
My uneducated guess is that when crossterm updated, it seems we got some changes to the events https://github.com/helix-editor/helix/commit/27211abf0688794f9bf7a395a8c47846b4f7fd41 and it somehow messed up the way we prioritized inputs over signals?

In any case I think it makes intuitive sense to handle the signals (kind of like a higher priority "interrupt") before the input.

If anyone has any suggestions for a comment to add, I'll gladly add it!

the-mikedavis:
#6149, #6160

It's interesting that the ordering fixes it here - I'm not totally sure why/how this branch ends up stealing the response to the keyboard enhancement check 🤔

the-mikedavis:
Right, the biased part isn't confusing. It's confusing that checking the event stream future changes the behavior of `crossterm::terminal::supports_keyboard_enhancement`. The signals and events are handled in the same order between this branch and master.

AlexanderBrevig:
This is with my patch
```
2023-03-03T21:02:47.827 helix_term::application [DEBUG] loop:handle_input_stream
2023-03-03T21:02:47.836 helix_view::document [DEBUG] id 3 modified - last saved: 0, current: 0
2023-03-03T21:02:47.839 helix_term::application [DEBUG] loop:handle_sig
2023-03-03T21:02:49.351 helix_term::application [DEBUG] loop:handle_sig
2023-03-03T21:02:49.351 helix_lsp::transport [INFO] <- ...
2023-03-03T21:02:49.353 helix_lsp::transport [INFO] <- ...
2023-03-03T21:02:49.354 helix_term::application [DEBUG] The enhanced keyboard protocol is supported on this terminal
2023-03-03T21:02:49.356 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","method":"initialized","params":{}}
```

This is without the patch:
```
2023-03-03T20:57:28.371 helix_term::application [DEBUG] loop:handle_input_stream
2023-03-03T20:57:28.386 helix_view::document [DEBUG] id 2 modified - last saved: 0, current: 0
2023-03-03T20:57:28.391 helix_term::application [DEBUG] loop:handle_sig
2023-03-03T20:57:28.451 helix_lsp::transport [INFO] <- ...
2023-03-03T20:57:28.451 helix_lsp::transport [INFO] <- ...
2023-03-03T20:57:32.019 helix_term::application [DEBUG] loop:handle_sig
2023-03-03T20:57:34.020 helix_term::application [DEBUG] The enhanced keyboard protocol is not supported on this terminal
2023-03-03T20:57:34.040 helix_view::document [DEBUG] id 2 modified - last saved: 0, current: 0
```

For me, the keyboard protocol is also correctly identified with the signals processed first.

What's the issue you're seeing?

the-mikedavis (approved):
Ok it looks like we're locking up crossterm's event polling system by trying to set up multiple concurrent readers. There's a mutex that allows only one reader to read/poll for events at a time. The `input_stream.next()` locks that mutex and then `crossterm::terminal::supports_keyboard_enhancement` tries to lock the mutex again to read the response to the query. The input event stream already holds the lock though so this times out unless a keyboard event arrives before the timeout. This has an effect you can test manually: C-z will hang the editor for two seconds on master but if you hit a button on your keyboard faster than the 2s, Helix will suspend properly. That's because the event stream future receives a KeyEvent and can drop its hold on the mutex. Then the keyboard enhancement protocol query can acquire the mutex and read the response which was pending that whole time.

I will open up an issue in the crossterm repo to talk about refactoring queries against the host terminal. This isn't just limited to `supports_keyboard_enhancement`: I believe it should also be possible to lock ourselves up with the cursor position query. We want other query capabilities in crossterm anyways like checking for synchronized output support and querying the shape of the cursor and title of the terminal on startup, so this is something I've been meaning to discuss anyways.

In the meantime I will merge this as a workaround since it's very easy to trigger #6149 and I don't see any downsides to handling signals first.

the-mikedavis:
Opened up https://github.com/crossterm-rs/crossterm/issues/763

PR #6173, opened 2023-03-03
Fix completion race conditions

pascalkuthe:
Supercedes #1819

This PR resolves multiple problems with the completions.
The root cause of all of these issues were shortcomings in the way savepoints are handled.
Savepoints are set when a completion is triggered so that the changes sent by the server
can be applied correctly (refer to the same document).

The exact explanation is a bit tricky. I will provide an overview here but looking at the commits individually also helps. Every commit has a separate explanation/justification and usually just contains a couple of LOC of changes and therefore makes the explanation more digestible.

# Overview
 
The savepoint is currently created when the completion request **response was received**. However, a (spec compliant) language server computes the completion response for the document state when it **receives the completion request**. That means that any changes between sending the completion requests to the server and receiving a response can cause all kinds of issues. For example if completion is requested at `foo::` and you keep typing `ba` before the response arrives and then select the `bar` completion the result is `foo::babar` instead of `foo::bar`.

The savepoint was moved to the completion request in 5545aad. However, this actually introduces new issues as there is only a single savepoint tracked on the document at a given time. The completion popup restores the savepoint whenever a new completion option is selected. However, if a new completion request is created while the completion popup is open (idle timeout) and a completion is selected before a response is received, then the completion popup would restore the wrong savepoint (as the savepoint was overwritten during the request). To alleviate that problem the savepoint mechanism was changed in 7e42f25 so that we can track an arbitrary number of (reference counted) snapshots on the document.

I noticed that multiple completion requests could be active at the same time (either because of low idle timeout values or because the user manually invoked `<C-x>`). Because LSP servers can arbitrarily reorder non-dependent requests (and it's random which response helix processes first if both finish at the same time) these completions are essentially racing. This is not as much of an issue anymore with the previous fixes, but it can cause an outdated completion request to replace the current completion request (which is not good). It can also cause some flickering. Most importantly it makes the association of savepoints to the completion popup essentially non-deterministic and therefore impossible to repeat correctly (see below). Therefore, I introduced a mechanism to automatically discard any running completion requests when a new one is created in 0a22fd2. Completion requests are also discarded if view/document changes.

Repeating completions also operated under the assumption that completion trigger creates savepoints. As this is not the case anymore I had to add an `InsertEvent` for requesting completions. When this event is replayed a new savepoint is created and saved. When the `TriggerCompletion` event is replayed the savepoint from the last request is set as activated (simply moved to a different variable). From that point forward any `CompletionAction` events will use that savepoint. For this to work it's critical that the `TriggerCompletion` event always belongs to the last completion request (otherwise the wrong savepoint is used). That is ensured by dropping stale completion requests (explained in the previous paragraph).

Another small fix included here: Savepoints now also store the selection. This is important since we apply multicursor completions relative to the selections. Usually this wasn't necessary because cursors tend to map back to the same position. However, this isn't always the case and can cause weird behavior in edgecase. In particular with snippet support (#5864) some completion snippets can cause more complex changes/arbitrary selection changes, and it's important that the selection is fully reset.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -820,6 +820,7 @@ impl EditorView {
                 (Mode::Insert, Mode::Normal) => {
                     // if exiting insert mode, remove completion
                     self.completion = None;
+                    cxt.editor.completion_request_handle = None;
 
                     // TODO: Use an on_mode_change hook to remove signature help
                     cxt.jobs.callback(async {
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -914,8 +914,8 @@ impl EditorView {
                                 doc.apply(&tx, view.id);
                             }
                             InsertEvent::TriggerCompletion => {
-                                let (_, doc) = current!(cxt.editor);
-                                doc.savepoint();
+                                let (view, doc) = current!(cxt.editor);
+                                doc.savepoint(view);
                             }
                         }
                     }
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -957,9 +957,6 @@ impl EditorView {
             return;
         }
 
-        // Immediately initialize a savepoint
-        doc_mut!(editor).savepoint();
-
         editor.last_completion = None;
         self.last_insert.1.push(InsertEvent::TriggerCompletion);
 
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -21,14 +21,14 @@ use helix_core::{
     visual_offset_from_block, Position, Range, Selection, Transaction,
 };
 use helix_view::{
-    document::{Mode, SCRATCH_BUFFER_NAME},
+    document::{Mode, SavePoint, SCRATCH_BUFFER_NAME},
     editor::{CompleteAction, CursorShapeConfig},
     graphics::{Color, CursorKind, Modifier, Rect, Style},
     input::{KeyEvent, MouseButton, MouseEvent, MouseEventKind},
     keyboard::{KeyCode, KeyModifiers},
     Document, Editor, Theme, View,
 };
-use std::{num::NonZeroUsize, path::PathBuf, rc::Rc};
+use std::{mem::take, num::NonZeroUsize, path::PathBuf, rc::Rc, sync::Arc};
 
 use tui::buffer::Buffer as Surface;
 
@@ -39,7 +39,7 @@ pub struct EditorView {
     pub keymaps: Keymaps,
     on_next_key: Option<OnKeyCallback>,
     pseudo_pending: Vec<KeyEvent>,
-    last_insert: (commands::MappableCommand, Vec<InsertEvent>),
+    pub(crate) last_insert: (commands::MappableCommand, Vec<InsertEvent>),
     pub(crate) completion: Option<Completion>,
     spinners: ProgressSpinners,
 }
@@ -49,6 +49,7 @@ pub enum InsertEvent {
     Key(KeyEvent),
     CompletionApply(CompleteAction),
     TriggerCompletion,
+    RequestCompletion,
 }
 
 impl Default for EditorView {
@@ -891,6 +892,8 @@ impl EditorView {
                 for _ in 0..cxt.editor.count.map_or(1, NonZeroUsize::into) {
                     // first execute whatever put us into insert mode
                     self.last_insert.0.execute(cxt);
+                    let mut last_savepoint = None;
+                    let mut last_request_savepoint = None;
                     // then replay the inputs
                     for key in self.last_insert.1.clone() {
                         match key {
@@ -898,7 +901,9 @@ impl EditorView {
                             InsertEvent::CompletionApply(compl) => {
                                 let (view, doc) = current!(cxt.editor);
 
-                                doc.restore(view);
+                                if let Some(last_savepoint) = last_savepoint.as_deref() {
+                                    doc.restore(view, last_savepoint);
+                                }
 
                                 let text = doc.text().slice(..);
                                 let cursor = doc.selection(view.id).primary().cursor(text);
@@ -915,8 +920,11 @@ impl EditorView {
                                 doc.apply(&tx, view.id);
                             }
                             InsertEvent::TriggerCompletion => {
+                                last_savepoint = take(&mut last_request_savepoint);
+                            }
+                            InsertEvent::RequestCompletion => {
                                 let (view, doc) = current!(cxt.editor);
-                                doc.savepoint(view);
+                                last_request_savepoint = Some(doc.savepoint(view));
                             }
                         }
                     }
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -940,17 +940,25 @@ impl EditorView {
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
     pub fn set_completion(
         &mut self,
         editor: &mut Editor,
+        savepoint: Arc<SavePoint>,
         items: Vec<helix_lsp::lsp::CompletionItem>,
         offset_encoding: helix_lsp::OffsetEncoding,
         start_offset: usize,
         trigger_offset: usize,
         size: Rect,
     ) {
-        let mut completion =
-            Completion::new(editor, items, offset_encoding, start_offset, trigger_offset);
+        let mut completion = Completion::new(
+            editor,
+            savepoint,
+            items,
+            offset_encoding,
+            start_offset,
+            trigger_offset,
+        );
 
         if completion.is_empty() {
             // skip if we got no completion results
@@ -969,8 +977,6 @@ impl EditorView {
         self.completion = None;
 
         // Clear any savepoints
-        let doc = doc_mut!(editor);
-        doc.savepoint = None;
         editor.clear_idle_timer(); // don't retrigger
     }
 
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -9,6 +9,7 @@ use helix_core::text_annotations::TextAnnotations;
 use helix_core::Range;
 use helix_vcs::{DiffHandle, DiffProviderRegistry};
 
+use ::parking_lot::Mutex;
 use serde::de::{self, Deserialize, Deserializer};
 use serde::Serialize;
 use std::borrow::Cow;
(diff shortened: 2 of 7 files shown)


pascalkuthe:
I put together some test infrastructure locally (which includes an in-process LSP server skeleton and a way to replace servers with a local server). This infrastructure has allowed me to test this much more thoroughly and identify edgecases missed in https://github.com/helix-editor/helix/pull/1819. However that infrastructure is quite a bit larger than the actual changes here. It's also still quite messy and therefore I have opted not to include it yet. Instead I will add that as a followup PR in the future.

Even with that testing infrastructure it was quite hard for me to reliably test the various issue that we have attributed to https://github.com/helix-editor/helix/pull/1819 in the past as none of them contain a reproduction case and even with a finegrained testing infrastructure I need to understand the nature of the bug to create a test.

I am fairly certain that this should fix the reported issues, nontheless it would be good if people could try this out locally to see if it fixes their problems as I can not reproduce most issues.

kirawi:
Would introducing an artificial delay to the LSP skeleton provide a testcase?

pascalkuthe:
> Would introducing an artificial delay to the LSP skeleton provide a testcase?

I have done this and used it for developing this prototype. Just an artifical delay is not enough as the typing and completion selection needs to be typed carefully. The problem is that this testecase is quite complicated and even just having any in process LS at all requires ~500loc of changes. I need to cleanup that code a lot more before that is ready for merging. The problem is that that case only covers the most basic case: The user types before the completion response is processed. But this PR handles many more cases (described in the overview). Creating clean testcases for all of is quite a bit of work and I didn't feel that the couple messy tests I have right now justified pulling in all that extra testing code and would have turned this reasonably sized PR is something quite large.

I found that the tests also don't really help while reviewing as they are complex themselves. Therefore I choose to delay adding itnegration tests to separate PR . Th3ere is lots of other tests we likely want to add for LSP support and I want to make sure that the in-process LS sceleton is flexible enough to allow all of them (without requiring huge amounts of boilerplate)

helix-view/src/editor.rs
```diff
@@ -881,6 +881,14 @@ pub struct Editor {
...
     /// times during rendering and should not be set by other functions.
     pub cursor_cache: Cell<Option<Option<Position>>>,
+    /// When a new completion request is sent to the server old
+    /// unifinished request must be dropped. Each completion
+    /// request is associated with a channel that cancels
+    /// when the channel is dropped. That channel is stored
+    /// here. When a new completion request is sent this
+    /// field is set and any old requests are automatically
+    /// canceled as a result
+    pub completion_request_handel: Option<oneshot::Sender<()>>,
```
rcorre, line 891:
```suggestion
    pub completion_request_handle: Option<oneshot::Sender<()>>,
```

helix-term/src/ui/editor.rs
```diff
@@ -891,14 +892,18 @@ impl EditorView {
...
                     // then replay the inputs
                     for key in self.last_insert.1.clone() {
                         match key {
                             InsertEvent::Key(key) => self.insert_mode(cxt, key),
                             InsertEvent::CompletionApply(compl) => {
                                 let (view, doc) = current!(cxt.editor);
 
-                                doc.restore(view);
+                                if let Some(last_savepoint) = last_savepoint.as_deref() {
+                                    doc.restore(view, last_savepoint);
```
dead10ck, line 905: Since this is just getting whatever the currently focused view is, that may not be the same as the view in the last save point, right? E.g. would this panic if we sent a completion request and then changed views?

dead10ck (commented):
On a first pass, this looks reasonable, though I'd like to take another look. I'm not quite sure I've grokked the save points being a list now.

helix-term/src/ui/editor.rs
pascalkuthe: Your general idea is correct, that's why I added a check to drop requests if the view or document has changed in https://github.com/helix-editor/helix/pull/6173/commits/56bf37a5b55ac46bda51f85c8840a355bb113db3

pascalkuthe:
> On a first pass, this looks reasonable, though I'd like to take another look. I'm not quite sure I've grokked the save points being a list now.

Simply think about save points being an `Arc` so the save point remains valid as long as you hold a defence to. The list on the document is only needed to keep updating the savepoints as long as anybody holds a reference to them (hence weak references)

Shirk:
It's unlikely that the issue is more frequent on 'slower CPUs' and feels a lot more related to what LSP and completion scenario you apply for testing. 

Using the rust-analyser LSP I don't experience letter doubling or any other issues with completions; using pylsp together with pylsp-mypy and the rather massive PySide6 type hints I can reproduce doubling on pretty much any system ranging from a 9th gen i7 going all the way up to a Ryzen 9 7600x as well as on an M1 and M2 max with a reasonable probability of issues happening.

poliorcetics:
> Using the rust-analyser LSP I don't experience letter doubling or any other issues with completions

As another datapoint, with helix-master (not this PR) I experience it a lot even with `rust-analyzer` on bigger projects with more than 100k lines of Rust with hundreds of dependencies

pascalkuthe:
The bugs this PR are fixed are observable if user input (or completion re-request) is faster than the response to completion requests. So a fast LSP server (RA) has the problem less often (usually only on large projects as @poliorcetics pointed out). My PC is *very* beefy (32 threads a 4.5ghz) so that is very rare for me (I never experienced this problem with rustc). I also don't tend to type super fast (and have my idle timeout set to the default) so that also contributes

Shirk:
Hence our suggestion to try something like PyLSP / a large project and not the very optimised rust LSP.

Trust me I am a very fast types as well and the hardware I listed is beefy as well but using pylsp the problem is very much testable if you want to try.

(this isn't meant as criticism it's meant to help you reproduce the issue, beefy PC is not a reason for it not to happen as multiple people pointed out)

helix-term/src/ui/editor.rs
dead10ck: I think I'm not quite understanding how this works. It looks like those additional checks would prevent a new pop-up menu of completion items if we received a response after moving views. But this code looks like it's for handling repetition of a completion application, which in my understanding, does not trigger a new pop-up, just inserts the same text again. How does that extra check come into play here?

helix-view/src/document.rs
```diff
@@ -940,14 +955,46 @@ impl Document {
...
+    /// `Document::savepoint` is set to `None`, or `savepoint` is called again
+    /// a transaction to revert the document back to this savepoint is automatically
+    /// maintained. `Document::restore` can be used to reset the document contents
+    /// and view selection back to the state when this function was called.
+    ///
+    /// # Returns
+    ///
+    /// An `id` that can uniquely identify this savepoint. A document
+    /// can only track a single savepoint at a time, to check if your savepoint still
+    /// applies simply
```
dead10ck, line 969: Looks like the sentence was never finished

dead10ck:
>The list on the document is only an implemententation detail that API consumers dont need to think about. It's only needed to keep updating the savepoints as long as anybody holds a reference to them (hence weak references)

Ok, I think I'm starting to understand. I was thinking about what happens if we have multiple save points in flight at a time, and at some point or another, we end up trying to restore them in succession? If I'm understanding this right, that would be fine since on each restore, we apply the invert transaction, and now this will compose the changes into each save point that's still valid.

helix-term/src/ui/editor.rs
pascalkuthe: The repetition code only replays events recorded while in insertmode and the appropriate event only gets generated if the completion menu is actually opened (prevent by said check). Specifically there are three events:

* Request completion: Generated immidietly when a completion is requested, generates a new savepoint when repeated
* trigger completion: Generated only if the completion menu is actually opened, active the savepoiny  described above (move it to a second variable)
* completion even: triggered when selecting something in the menu, restores the **active** savepoint

The important part is that there are always two savepoints during repeating: an active savepoint actually used for completions and the savepoint of the current request. 

So in the case you mentioned a completion request (and completion request event) would occur bit if the view/doc is switched then the request gets dropped after it finishes and no trigger completion event is ever generated. Therefore the savepoint belonging to the request is never activated and would be replaced by the next completion request event

pascalkuthe:
> > The list on the document is only an implemententation detail that API consumers dont need to think about. It's only needed to keep updating the savepoints as long as anybody holds a reference to them (hence weak references)
> 
> Ok, I think I'm starting to understand. I was thinking about what happens if we have multiple save points in flight at a time, and at some point or another, we end up trying to restore them in succession? If I'm understanding this right, that would be fine since on each restore, we apply the invert transaction, and now this will compose the changes into each save point that's still valid.

Ah I see yeah that can be a bit confusing. Your idea is exactly right the revert transaction is simply added on top so the revert is essentially undone too.

Basically savepoints are only kept on the document as a performance optimization. A simpler/slower implantation would simply store a snapshot of the document (so a clone of the `Rope`) and compute a diff of that with the document contents to generate a transaction. The list on the document only exists so that we don't have to compute a full diff and instead "remember" what changed

helix-view/src/document.rs
pascalkuthe: Ah this documentation is actually outdated and refers to an old version of this function from a previous slightly different approach that I moved awaybfrom (refcount stored on the document, and savepoints tracked using ids, was very error prone because it was easy to forgot to drop a savepoint or use a dropped savepoint for a revert, the recounting is much more robust). Will fix that comment in a bit

rcorre:
This problem is really easy to repro using the Godot LSP with helix `main`. I wasn't able to repro at all with this patch. Here's a comparison with idle_timeout=100ms (`main` first, this patch second):
https://asciinema.org/a/IJWMTBQiBFMYSwfvi9p7KQ2a2

helix-term/src/ui/editor.rs
dead10ck: Got it, this makes a lot more sense now, thanks for the write-up!

pascalkuthe:
> This problem is really easy to repro using the Godot LSP with helix `main`. I wasn't able to repro at all with this patch. Here's a comparison with idle_timeout=100ms (`main` first, this patch second): [asciinema.org/a/IJWMTBQiBFMYSwfvi9p7KQ2a2](https://asciinema.org/a/IJWMTBQiBFMYSwfvi9p7KQ2a2)

thanks for testing :)

dead10ck (approved):
Ok, I've been using it for a couple days now, and the code looks good to me too. Thanks @pascalkuthe!

StratilJakub:
Just want to say thanks a lot for fixing this. This was quite a big issue with languages like JavaScript.

Rudxain:
Does this fix microsoft/TypeScript#60606? It's been a long time since I've last experienced the bug. I'm still not sure if it was fixed on the Helix side, or the TS-LSP side, or the `tsserver` side (or multiple of them!)

I feel like I need some closure on this, because this mystery is still "haunting" me

Issue #3901, opened 2022-09-19
Support taplo with schemas


dead10ck:
A couple of issues (#3383, #3897) have been opened around integration with taplo for TOML files related to Helix itself. It seems the root of this issue is that there are no schemas published for Helix.

I myself don't know much about how taplo works, or json schemas, but perhaps we could put together and publish a scheme, or maybe there's something we could check in and a default config setting we could use for them?

ModProg:
For anyone interested in working on this, here is the documentation on how to create/publish schemas for taplo: https://taplo.tamasfe.dev/configuration/developing-schemas.html

LukeHalasy:
I would love ❤️  to run with this.
Let's auto-generate the schema for helix-related TOML files as others have mentioned.

Additionally, we should commit the generated schema files since the possible keys + values within each TOML file change depending on the commit you are on.

Will be awesome to be able to view the documentation for all the possible keys within the TOML files without having to go to the book/website.

(Would be a cool UX feature to have Taplo auto-installed along with helix so that anytime a new user goes to their config they have auto-complete enabled by default)

ModProg:
> Additionally, we should commit the generated schema files since the possible keys + values within each TOML file change depending on the commit you are on.

It should probably be possible for helix to tell taplo exactly what schema to use, so you always have the correct completions.

Publishing them to the [Json Schema Store](https://www.schemastore.org/json/) additionally makes sense, but I think it is save to assume that the config file you are editing is in most cases for the version of helix you are currently running.

pascalkuthe:
I don't really think it makes sense to invest a bunch of effort into writing and maintaining these schemas at this point since we plan to replace the toml config with a scheme based config in the future

LukeHalasy:
> I don't really think it makes sense to invest a bunch of effort into writing and maintaining these schemas at this point since we plan to replace the toml config with a scheme based config in the future

Thanks for the quick notice. I will **NOT** be running with this then. 

Should we close this issue since the original description talks about "integration with taplo for TOML files related to Helix itself" ?

andradei:
What is Helix switching to?

the-mikedavis:
The plugin system will most likely be a scheme dialect, so config will be written in scheme and be scriptable rather than TOML

PR #432, opened 2021-07-11
Dynamically load grammar libraries at runtime

archseer:
Motivated by some of the findings in #330 I went ahead and attempted to split out grammars.

The set of languages is no longer static, this way the user will be able to self-compile additional grammars and provide them under `runtime/grammars`. Languages can now be declared without a matching grammar too, which should unblock #205.

I think it would be cool to expose some of the `build.rs` code as an optional feature under the `hx` CLI, so that the user would be able to easily point to a cloned tree sitter grammar and compile it into their runtime dir.

Additionally, the grammar objects are now persisted and we only recompile when necessary (if the grammar sources changed), which should massively speed up `cargo clean && cargo build`.

With these changes, release build of `hx` is now `7.8M` after stripping.

PR #6174, opened 2023-03-03
[WIP] Add option documentation to :set-option, :get-option.

### LINKED ISSUE #5938: Improve documentation popup for `:set`
- In hx 22.12, the `:set` shows following in help popup (or what's it called?)
- and the "Usage" line is shown only when it is used wrongly. 
- I'd say that the usage line should be part of normal set popup too

What do u think?

### Current

```
Set a config option at runtime.
For example to disable smart case search, use `:set search.smart-case false`.
```
```
Bad arguments. Usage: `:set key field`
```

### Proposals

1. Put usage in next line, shorten example to e.g.
```
Set a config option at runtime.
Usage: `:set key field`. E.g. to disable smart case search, use `:set search.smart-case false`.
```

2. Put usage in title line
```
Set a config option at runtime. Usage: `:set key field`. 
For example to disable smart case search, use `:set search.smart-case false`.
```

3. Put shortened usage in title line.
```
Set a config option at runtime. `:set key field`.
For example to disable smart case search, use `:set search.smart-case false`.
```

<br/>I think, 2. makes most sense.
Ur thoughts?

### LINKED ISSUE #5939: 2 level documentation popup for `:set`

askreet:
~Closes #5938.~ Oops, I mean closes #5939.

Looking for some early feedback on approach here. The idea is that we move the options from `configuration.md` into a structured format that can be used both by docgen and the help built into the editor.

When combined with #5966 (which suppresses autocomplete for the second argument) this is the user experience:

[asciicast](https://asciinema.org/a/jUqX3taJbjl0YP7nZN9SpCk38)

If this general direction gets a few :+1:, I would do the following before merge:
- [ ] Finish importing all the options.
- [ ] Update `configuration.md` to be some sort of template that can pull sections of options from `runtime_options::Options` and build the tables dynamically.

Some open questions I have:
* The current approach is hacked in for a subset of specific commands. More generally, we want the ability to defer documentation optionally to a TypedCommand, perhaps by adding a `doc_fn: Option<Fn...>` to TypedCommand?
* Should the option-specific documentation be in _addition_ to the command documentation, perhaps separate by a `===` line, instead of replacing it?
* My `Lazy<Result<...>>` for the reference to `runtime_options::Options` is defensive against the runtime directory not having `help/options.toml`. Is this a good approach? Alternatives?
  * One option would just be baking the options into a const static instead of using toml. Thoughts?

Potential future roadmap:
* Auto-completers with more context that would allow us to auto-complete one-of validations in the editor (bufferline is a prime example).
* Color or popups when the pending value is not correct.
* More specific validators: numeric ranges, regular expressions, filesystem paths?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -322,6 +322,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "0bd4b30a6560bbd9b4620f4de34c3f14f60848e58a9b7216801afcb4c7b31c3c"
 
 [[package]]
+name = "dyn-clone"
+version = "1.0.11"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "68b0cf012f1230e43cd00ebb729c6bb58707ecfa8ad08b52ef3a4ccd2697fc30"
+
+[[package]]
 name = "either"
 version = "1.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1086,6 +1092,7 @@ dependencies = [
  "quickcheck",
  "regex",
  "ropey",
+ "schemars",
  "serde",
  "serde_json",
  "slotmap",
@@ -1181,6 +1188,7 @@ dependencies = [
  "log",
  "once_cell",
  "pulldown-cmark",
+ "schemars",
  "serde",
  "serde_json",
  "signal-hook",
@@ -1244,6 +1252,7 @@ dependencies = [
  "log",
  "once_cell",
  "parking_lot",
+ "schemars",
  "serde",
  "serde_json",
  "slotmap",
@@ -1762,6 +1771,30 @@ dependencies = [
 ]
 
 [[package]]
+name = "schemars"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "02c613288622e5f0c3fdc5dbd4db1c5fbe752746b1d1a56a0630b78fd00de44f"
+dependencies = [
+ "dyn-clone",
+ "schemars_derive",
+ "serde",
+ "serde_json",
+]
+
+[[package]]
+name = "schemars_derive"
+version = "0.8.12"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "109da1e6b197438deb6db99952990c7f959572794b80ff93707d55a232545e7c"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "serde_derive_internals",
+ "syn",
+]
+
+[[package]]
 name = "scopeguard"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -1787,6 +1820,17 @@ name = "serde_derive"
 version = "1.0.155"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d071a94a3fac4aff69d023a7f411e33f40f3483f8c5190b1953822b6b76d7630"
+dependencies = [
+ "proc-macro2",
+ "quote",
+ "syn",
+]
+
+[[package]]
+name = "serde_derive_internals"
+version = "0.26.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "85bf8229e7920a9f636479437026331ce11aa132b4dde37d121944a44d6e5f3c"
 dependencies = [
  "proc-macro2",
  "quote",
@@ -2469,5 +2513,7 @@ dependencies = [
  "helix-loader",
  "helix-term",
  "helix-view",
+ "schemars",
+ "serde_json",
  "toml",
 ]
--- a/helix-core/Cargo.toml
+++ b/helix-core/Cargo.toml
@@ -46,6 +46,7 @@ chrono = { version = "0.4", default-features = false, features = ["alloc", "std"
 
 etcetera = "0.4"
 textwrap = "0.16.0"
+schemars = { version = "0.8.12", features = ["derive_json_schema"] }
 
 [dev-dependencies]
 quickcheck = { version = "1", default-features = false }
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -26,6 +26,7 @@ use std::{
 };
 
 use once_cell::sync::{Lazy, OnceCell};
+use schemars::JsonSchema;
 use serde::{Deserialize, Serialize};
 
 use helix_loader::grammar::{get_language, load_runtime_file};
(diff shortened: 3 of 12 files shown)


the-mikedavis (commented):
Why store the options in `runtime/` as TOML? We know all options at compilation time so it should be possible to describe them using types and/or constants

askreet:
> Why store the options in runtime/ as TOML? We know all options at compilation time so it should be possible to describe them using types and/or constants

I had the same question (listed above), any tradeoffs to consider here? I look at something like `languages.toml` and some of the other files in `runtime` and wonder when each of these options is preferred? Presumably it just balloons the binary size a bit. Any other downsides?

If not, I'll move it into a constant -- seems a lot easier to work with (e.g., not having to worry about whether it loads).

the-mikedavis:
The configuration in `languages.toml` is exposed as TOML so that you can override or extend it without recompiling the editor. That wouldn't help for options though since you would need to edit/recompile the editor anyway to take advantage of a new option.

A large binary size was a problem in the past but that was mostly about tree-sitter parsers being built in to the Helix binary (see https://github.com/helix-editor/helix/pull/432). I don't think the binary size change would be very noticeable from adding option docs

pascalkuthe:
While I like the general direction of providing a better user experience I would prefer if we could auto generate the data from doc comments (and update those to match the documentation once).

I have been thinking that we could use the `schemars` crate here. It's designed around automatically providing json schemas for serde deserialized structs. This would also solve #3901 and you can extract doc comments from this quite easily. For example:

``` rust
use schemars::gen::SchemaGenerator;
use schemars::JsonSchema;

#[derive(JsonSchema)]
/// Hello Foo
struct Foo {
    /// Hello Bar
    pub bar: String,
}

#[test]
fn test_schmema() {
    let mut schema = Foo::json_schema(&mut SchemaGenerator::default()).into_object();
    assert_eq!(schema.metadata().description.as_deref(), Some("Hello Foo"));
    assert_eq!(
        schema.object().properties["bar"]
            .clone()
            .into_object()
            .metadata()
            .description
            .as_deref(),
        Some("Hello Bar")
    );
}
```

askreet:
> While I like the general direction of providing a better user experience I would prefer if we could auto generate the data from doc comments (and update those to match the documentation once).

Yeah! I was looking for something to simply give me the doc comments but it all seemed really complicated. Didn't consider extracting a schema first and traversing that. I really like this idea, let me spend some time on it over the coming days/weeks.

askreet:
@pascalkuthe I managed to get it working using the JsonSchema trait -- it's very cool. The code is _super_ rough right now, and doesn't hit a ton of cases, but works for all top level booleans and numerics, as well as enums with simple variants (e.g., `bufferline`). The `schemars` crate doesn't offer a lot of flexibility in traversing it's graph so the code is full of boilerplate and unwraps. My thoughts on moving forward:

* If we write an algorithm that handles all the current shapes of our config types, are we concerned about the maintainability of adding differently-shaped types? For example, the current implementation doesn't know what to do with `idle-timeout` which has a structure with two fields, and some custom serialization/deserialization logic. I'm pretty sure the JsonSchema for it is just _wrong_.
* Assuming we move forward, how defensive should this code be? If we run `from_config` in a test case (as I have now) it should be impossible for this to fail at runtime if that test has passed. Is that good enough, or should I put a lot of effort into converting all the unwrap cases into errors for example?

Overall I'm a bit torn, I love the code being the source of truth, but I wouldn't want it to be harder to expand the configuration structures in the future, especially if someone tries to do that and then has to look at all the struct->JsonSchema->magic->runtime_options::Options heroics I may have to do to get this to where I want it.

pascalkuthe:
> @pascalkuthe I managed to get it working using the JsonSchema trait -- it's very cool. The code is _super_ rough right now, and doesn't hit a ton of cases, but works for all top level booleans and numerics, as well as enums with simple variants (e.g., `bufferline`). The `schemars` crate doesn't offer a lot of flexibility in traversing it's graph so the code is full of boilerplate and unwraps. My thoughts on moving forward:
> 
>     * If we write an algorithm that handles all the current shapes of our config types, are we concerned about the maintainability of adding differently-shaped types? For example, the current implementation doesn't know what to do with `idle-timeout` which has a structure with two fields, and some custom serialization/deserialization logic. I'm pretty sure the JsonSchema for it is just _wrong_.
> 
>     * Assuming we move forward, how defensive should this code be? If we run `from_config` in a test case (as I have now) it should be impossible for this to fail at runtime if that test has passed. Is that good enough, or should I put a lot of effort into converting all the unwrap cases into errors for example?
> 
> 
> Overall I'm a bit torn, I love the code being the source of truth, but I wouldn't want it to be harder to expand the configuration structures in the future, especially if someone tries to do that and then has to look at all the struct->JsonSchema->magic->runtime_options::Options heroics I may have to do to get this to where I want it.

I would want to keep it simple too so we shouldn't write some complex magic to extract the description.

That being said, I think `schemars` is actually build to generate schemas for custom destabilization logic. Just like `serde` allows you to specify a custom deserialize using `#[serde(with` for a field or just implement `Deserialize` yourself you can use `#[schema(with` or just impl `JsonSchema` yourself. So you really get the best of both worlds. An Automatically generated schema for most cases avoiding duplication and in the edgecases where deriving doesn't quite cover it you can specify the schema manually (which would be required with a manual approach anyway)

> Assuming we move forward, how defensive should this code be? If we run from_config in a test case (as I have now) it should be impossible for this to fail at runtime if that test has passed. Is that good enough, or should I put a lot of effort into converting all the unwrap cases into errors for example?

If I understand what you mean here correctly than I think unwrapping is fine since the schema is essentially hardcoded.

askreet:
> ... you can use #[schema(with or just impl JsonSchema yourself.

Ah, good idea. I haven't looked to see what annotations they support but you're right I could implement JsonSchema myself rather than juggle the bits on the scanning side. I'll find some time in the coming days to keep poking at this, thanks for the thoughts. 👍🏼

askreet:
@pascalkuthe I spent a bit more time on this trying to adapt JsonSchema to some of the special-case config options that exist in the project and I'm about ready to give up and pivot to something more structured (such as a trait implementation we can create for each runtime-configurable pile of settings). These edge cases (like auto-pairs accepting boolean values, but not really being a bool) are really rough to work with within the JsonSchema hierarchies and produce piles of code as a result. I'm curious for your thoughts here on whether something simpler is a good interim step until someone with more stamina wants to chase this down?

This branch can export the full JSON schema of the config object, I'm not sure if that's useful for #3901 or if it has all the same edge cases.

askreet:
I'm going to close this out for now. IMO, the JsonSchema stuff, while interesting, is a ton of work and potentially breaks some of the config modeling. I'm going to pivot this to a statically defined approach (as @the-mikedavis mentioned) when I have some cycles to work on it.

goyalyashpal:
@askreet did u continue this in some other pr?

askreet:
@goyalyashpal I haven't. Feel free to pick it up if interested, there's some generally useful bits here for either approach.

PR #6185, opened 2023-03-04
Raise MSRV to 1.65

pascalkuthe:
Updates MSRV to 1.65 per our policy.

Ubuntu has finally followed their own policy and updated rustc to 1.65 so this should now be uncontroversial and not cause any problems.

This PR includes #6065 (with permission from @Byron) to update gix and the usual update of the CI, `rust-toolchain.toml` and resolving of MSRV related TODOs (clippy lints were resolved in #5892)

Note that the large diff is caused by `Cargo.lock` changes mostly from renaming all the `gitoxide` subcrates from `git-..` to `gix-..` and gitoxide dropping dependencies (caused by reviews from the cargo people, as gitoxide is now used in `cargo`)

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/helix-editor/helix"
 homepage = "https://helix-editor.com"
 include = ["src/**/*", "README.md"]
 default-run = "hx"
-rust-version = "1.57"
+rust-version = "1.65"
 
 [features]
 default = ["git"]
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -37,7 +37,7 @@ jobs:
         uses: actions/checkout@v3
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.63
+        uses: dtolnay/rust-toolchain@1.65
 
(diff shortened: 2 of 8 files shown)


helix-term/Cargo.toml
```diff
@@ -10,7 +10,7 @@ repository = "https://github.com/helix-editor/helix"
 homepage = "https://helix-editor.com"
 include = ["src/**/*", "README.md"]
 default-run = "hx"
-rust-version = "1.57"
+rust-version = "1.65"
```
the-mikedavis, line 13: Can we remove this key altogether? Looks like it's been out of date for some time anyways

pascalkuthe: Its not needed but I think it's nice to have as it lets cargo know what MSRV we are targeting. If somebody then tries to build helix with an out of date toolchain (om Debian or if they forgot to update for a while) they will get a helpful error message telling them that their rust toolchain is out of date and needs to be updated. 

It's not super important but nice to have and not a lot of effort to maintain so I would prefer to keep it if it's ok with you

the-mikedavis (approved):

dead10ck (approved):
Thanks!

archseer:
> and gitoxide dropping dependencies

Nice!

PR #6188, opened 2023-03-05
Enhancement: Add better default breakpoint icon

### LINKED ISSUE #5956: Add better breakpoint icons
# Description

Currently, breakpoints do not use a conventional symbol, such as the classic filled red circle, the empty circle or the circle with a warning sign.

# Proposal

Use nerdfont icons to set more conventional icons for breakpoints (dependent on #2869).

# UI mockup
image

EricHenry:
## Summary

closes #5956 

In #5952 @archseer and @filipdutescu suggested to use the same icon being used by the diagnostics for breakpoints. I also updated the default color to use the `error` color set by the UI theme.

Screenshot 2023-03-04 at 11 25 54 PM


archseer:
That just looks like a regular error now. I thought we picked a different icon? I can't find the thread right now

pascalkuthe:
Duplicate of #5957 which includes an improved icon and a new highlight scope plus theme adjuatement

filipdutescu:
I meant adding NerdFont icons such as

image
image


making use of #2869, which would look a bit better than the current ones in #5957. What you did is a duplicate of my work, as @pascalkuthe and @archseer mentioned.

EricHenry:
@archseer @filipdutescu whoops sorry about that! I definitely misunderstood. The flow of the conversation was a bit confusing between the two issues.

EricHenry:
What are your thoughts on using the bug and stackframe arrow for the icons? Also I noticed that the breakpoint struct has information about conditionals and log messages, which seem to have icons for. What are your thoughts about using them?
Screenshot 2023-03-05 at 8 48 28 PM

Here is how it looks in the code with the debug... bug.
Screenshot 2023-03-05 at 8 45 12 PM
Screenshot 2023-03-05 at 8 46 58 PM

pascalkuthe:
the bug looks a bit noisy to me.

For specical breakpoints (condition, log, data, function,...) there are already icons:

image

I am not sure what a data breakpoint is supposed to be so we could just use that icon for normal breakpoints. Alternatively just the various nerdfont cricels seem like a good option:

image

EricHenry:
Thanks for the feedback!

> For specical breakpoints (condition, log, data, function,...) 
We don't have those codified yet, I'll add those icons in and update the logic in the function that determines the icon for the breakpoint.

> the bug looks a bit noisy to me.
Yeah, I agree. I'll use the regular filled circle for the normal breakpoints.

filipdutescu:
Stack frame one is exactly the kind of icon I had in mind adding the task. Great catch!

filipdutescu:
I would say that conditional breakpoints take precedence. To me it is very important to know when/if a breakpoint will be hit, to plan how my debugging session should go.

PR #6192, opened 2023-03-05
Update onelight theme

erasin:
Reset the keyword section to complete most of the highlighting colors.

wiki has update.

20230305-225817@2x

### DIFF
--- a/runtime/themes/onelight.toml
+++ b/runtime/themes/onelight.toml
@@ -1,6 +1,6 @@
 # Author : erasin<erasinoo@gmail.com>
 
-"attribute" = { fg = "light-blue" }
+"attribute" = { fg = "yellow" }
 "constructor" = { fg = "brown" }
 "label" = { fg = "cyan" }
 "operator" = { fg = "red" }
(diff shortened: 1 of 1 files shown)


runtime/themes/onelight.toml
```diff
@@ -1,129 +1,184 @@
...
-"punctuation.bracket" = { fg = "gray" }
-
-"variable" = { fg = "black" }
-"variable.builtin" = { fg = "light-blue" }
-"variable.parameter" = { fg = "red" }
-"variable.other.member" = { fg = "red" }
+"function" = { fg = "blue" }
+"function.builtin" = { fg = "cyan" }
+"function.method" = { fg = "light-blue" }
+"function.macro" = { fg = "pink", modifiers = ["blod"] }
```
the-mikedavis, line 67:
```suggestion
"function.macro" = { fg = "pink", modifiers = ["bold"] }
```

runtime/themes/onelight.toml
```diff
@@ -1,129 +1,184 @@
...
 
 "ui.background" = { bg = "white" }
+"ui.background.separator" = { bg = "white" }
 
-"ui.cursor" = { fg = "white", bg = "gray" }
+"ui.cursor" = { fg = "white", bg = "grey" }
+"ui.cursor.normal" = { fg = "white", bg = "grey" }
+"ui.cursor.insert" = { fg = "white", bg = "grey" }
+"ui.cursor.select" = { fg = "white", bg = "grey" }
+"ui.cursor.match" = { bg = "light-grey" }
```
the-mikedavis, line 106: It looks like this isn't defined in the palette. Maybe this could use one of the `grey-*` shades?

erasin: Thanks for the review, here is a change to `light-white`, because the multi-cursor uses the grey style, here only do grayscale and multi-cursor style to distinguish.

PR #6194, opened 2023-03-05
Move terminal claim/restore to helix-tui, cache keyboard enhancement support query

the-mikedavis:
This is #6160 but in a way that is more extendable in the long-term. This will make it easier to add checks for other host terminal information like initial cursor shape, title line, whether the terminal supports synchronized output, etc. without adding a whole bunch of fields to Application or functions to the Terminal trait.

The first commit in this change moves `Application::claim_term` and `application::restore_term` to `helix-tui` into the crossterm backend. I believe this is the more proper place for this code and it makes the second change much easier: we cache whether the terminal supports the enhanced keyboard protocol, resolving #6149 (though it would stil be nice to improve this upstream).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1217,6 +1217,7 @@ dependencies = [
  "helix-core",
  "helix-view",
  "log",
+ "once_cell",
  "serde",
  "termini",
  "unicode-segmentation",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1216,6 +1216,7 @@ dependencies = [
  "crossterm",
  "helix-core",
  "helix-view",
+ "log",
  "serde",
  "termini",
  "unicode-segmentation",
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -30,22 +30,14 @@ use crate::{
 
 use log::{debug, error, warn};
 use std::{
-    io::{stdin, stdout, Write},
+    io::{stdin, stdout},
     sync::Arc,
     time::{Duration, Instant},
 };
 
 use anyhow::{Context, Error};
 
-use crossterm::{
-    event::{
-        DisableBracketedPaste, DisableFocusChange, DisableMouseCapture, EnableBracketedPaste,
-        EnableFocusChange, EnableMouseCapture, Event as CrosstermEvent, KeyboardEnhancementFlags,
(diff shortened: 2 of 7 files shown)


rcorre:
This resolves https://github.com/helix-editor/helix/issues/6149 for me with `st 0.8.5`.

I don't know if there's a good way to measure it, but it feels like there's still a small but noticeable lag when I first open a file, compared to versions before a066815833b322233dc11aeae38679bc8466ec56, where opening a file feels instantaneous.

Would it make sense to have an option similar to `true-color`, where you can just set it once in your config and bypass the live check every time?

the-mikedavis:
I'll give `st` a try. The query itself has been unnoticeably fast in the editors that I have tried so if there's a delay, I would guess it may be an issue with crossterm's event polling system. This PR executes the same query but just caches the result so the initial query shouldn't be any slower. I'll test a bunch of different terminals to see what I can find though.

> Would it make sense to have an option similar to `true-color`, where you can just set it once in your config and bypass the live check every time?

The mechanisms for detecting true-color and undercurl support (see also #6196) are not very robust (they depend on an environment variable and your installed terminfo database respectively) and the query for keyboard enhancement support here is much more reliable (a direct request/response between the application and host terminal) so I don't think a new option to force it should be necessary.

rcorre:
> so the initial query shouldn't be any slower

Sorry, to clarify, this PR isn't any slower than https://github.com/helix-editor/helix/commit/a066815833b322233dc11aeae38679bc8466ec56, and is a significant improvement in general.
I meant that helix _appears_ slightly slower in https://github.com/helix-editor/helix/commit/a066815833b322233dc11aeae38679bc8466ec56 than in prior versions.

the-mikedavis:
I added some logging to the debug line that says whether the protocol is supported or not. I checked around and the check seems pretty fast on terminals that do support the protocol (3-4ms on Kitty, WezTerm) and very fast on terminals that don't (~100μs on st, Alacritty). Let me know if you find longer measurements!

I think it might be a nice improvement to run this query and enable keyboard enhancement asynchronously after startup. It isn't safe to do it with the current way queries work in crossterm though - that would need to wait on a change in crossterm that would allow multiple concurrent readers.

rcorre:
Ugh, I was comparing a `debug` build to a `release` build. That might have been it, or I was imagining. Either way, I'm not noticing it now. Sorry for the wild goose chase, and thanks for the fix!

PR #6197, opened 2023-03-05
Feat: Highlight the cursor-word

### LINKED ISSUE #3436: Highlighting the word under the main cursor automatically.
Kakoune allows highlighting the word under the main cursor automatcially.

```kak
declare-option -hidden regex cursor_word
add-highlighter global/ dynregex '%opt{cursor_word}' 0:bright-white
```
The result is as follows:

image

Is it possible with helix?

SoraTenshi:
This Feature closes #3436

(Early) Feature preview: 
image

I know that `<Space>-h` exists, however i prefer it the way, that it automatically highlights all occurrences of that word that is under the cursor.

- [x] Base functionality
- [x] Use LSP
- [x] Configurable style

Also another question: would it be preferred if the style is either underline / bold? It depends heavily on the implementation and i am not sure if my current implementation is good.

Inspired by: https://github.com/xiyaowong/nvim-cursorword


pascalkuthe:
You seem to perform a simple text based matching. While this is a good fallback, I think we should use the LS if possible since the same textual word might not always correspond to the same identifier.  Specifically you can probably use the `text_document_document_highlight` request already used by `<space>-h`. Since you get a list of text ranges from that API it would be more appropriate to create a highlight iterator and merge it before the syntax highlights with `syntax::merge`.

SoraTenshi:
> You seem to perform a simple text based matching. While this is a good fallback, I think we should use the LS if possible since the same textual word might not always correspond to the same identifier. Specifically you can probably use the `text_document_document_highlight` request already used by `<space>-h`. Since you get a list of text ranges from that API it would be more appropriate to create a highlight iterator and merge it before the syntax highlights with `syntax::merge`.

You're correct. I would however still "finalize" the raw, non-lsp version of this feature and then probably prioritize based on the lsp.
Also; from what i know about `<Space>-h`, i would also implement boundary checks for the highlighting (Space-h behaviour is, that it select the variables)

pascalkuthe:
> You're correct. I would however still "finalize" the raw, non-lsp version of this feature and then probably prioritize based on the lsp.

Sure I think having a plaintext version is still nice. That being said I think that doing this as a line annotation is probably not quite the right way to go (although I am really happy you like the API :P). Really long words can be split across multiple lines when soft-wrapping so a line doesn't always end with a word (which can lead to wrong highlights in multiple ways). It's also slightly slower than it could be (details are a bit complex but unimportant).

The ideal way to implement this would be using a regex `\b{word}\b` but sadly right now `regex` requires us to convert a `String` which negates the performance benefits. Although that will changes [soon](https://github.com/rust-lang/regex/issues/656#issue-586996204) (super excited for that) but not yet... Instead you probably want to use `text.slice(text.line_to_char(start_line)..text.line_to_char(start_line + height).chars()` and manually start matching the search string at every search boundary, keep the found matches in a `Vec` and then use `syntax::merge` just like you would for the LSP

> Also; from what i know about <Space>-h, i would also implement boundary checks for the highlighting (Space-h behaviour is, that it select the variables)

Note quite sure what you mean with this

SoraTenshi:
> > Also; from what i know about -h, i would also implement boundary checks for the highlighting (Space-h behaviour is, that it select the variables)
> 
> Note quite sure what you mean with this

Space-h selects all word occurrences, that may also not be within the viewport.

Also, thanks for the idea. i will take a look at that!

pascalkuthe:
> > > Also; from what i know about -h, i would also implement boundary checks for the highlighting (Space-h behaviour is, that it select the variables)
> > 
> > 
> > Note quite sure what you mean with this
> 
> Space-h selects all word occurrences, that may also not be within the viewport.
> 
> Also, thanks for the idea. i will take a look at that!

That's true but not really an issue, `syntax:merge` will skip anything that's out of bound automatically

SoraTenshi:
> > > > Also; from what i know about -h, i would also implement boundary checks for the highlighting (Space-h behaviour is, that it select the variables)
> > > 
> > > 
> > > Note quite sure what you mean with this
> > 
> > 
> > Space-h selects all word occurrences, that may also not be within the viewport.
> > Also, thanks for the idea. i will take a look at that!
> 
> That's true but not really an issue, `syntax:merge` will skip anything that's out of bound automatically

oh wow, than it's a lot easier than i thought.
Thanks! Will continue working on it :)

SoraTenshi:
@pascalkuthe If you have some spare time, I would love a review :)
Anyone else is of course free to review it as well! :)

helix-term/src/ui/editor.rs
```diff
@@ -264,6 +272,121 @@ impl EditorView {
...
+
+            first_visible_line..last_visible_line
+        };
+
+        for line in line_range {
+            let current_line = text.line(line);
+            let current_line = match current_line.as_str() {
+                Some(l) => l,
+                None => continue,
+            };
```
pascalkuthe, line 337: this is ineffecit. First you are iterating a line range and obtaining a line with `text.line`  which is  `O(logN)`  so this loop is `O(MlogN)` with the `lines` iterator it would be `O(logN + M)`. However you are not actually interested in lines anyway since you are really just obtaining the string. The problem is that `as_str` will return `None` at any line that contains a ropey chunk boundary which could weridly discard lots of matches. There is actually aa way to implement this with a small temporary buffer that doesn't discard any matches at all but that complexity is probably not worth it/required here for now. Instead you can just iterate the chunks directly. That way only words that actually contain a chunk boundary are now shown (which is pretty rare):

```suggestion

        for chunk in text.lines(line_range).chunks(){

```

helix-term/src/ui/editor.rs
```diff
@@ -264,6 +272,121 @@ impl EditorView {
...
+            result.extend(
+                current_line
+                    .match_indices(&cursor_word)
+                    .map(|(i, _)| i)
+                    .filter(|i| {
+                        if *i != 0 {
+                            !is_word_char(current_line.chars().nth(i - 1))
+                        } else {
+                            true
+                        }
```
pascalkuthe, line 350: `match_indicies` return byte indices so using that as a char offset is incorrect (and slow).
A more efficient way to implement this would be 

```suggestion
	char_is_word(chunk[..i].chars().next_back())
```

(returns `false` automatically in chunk boundaries as `next_back` returns `None` in that case)

helix-term/src/ui/editor.rs
```diff
@@ -264,6 +272,121 @@ impl EditorView {
...
+    pub fn collect_cursor_word_highlights(
+        doc: &Document,
+        editor: &Editor,
+        view: &View,
+        viewport: Rect,
+        theme: &Theme,
+    ) -> Option<Vec<(usize, std::ops::Range<usize>)>> {
+        let scope_index = theme.find_scope_index("ui.wordmatch")?;
+        let mut result: Vec<(usize, std::ops::Range<usize>)> = Vec::new();
+        let is_lsp_ready = doc.language_server().is_some() && !editor.cursor_highlights.is_empty();
```
pascalkuthe, line 371: I think there should be some specical handling to indicate when the LSP is ready. It's perfectly valid to for the LSP to return an empty range and if that is really the LSPs newest response than we we should do respect that. An option would probably work well

helix-view/src/editor.rs
```diff
@@ -865,6 +867,9 @@ pub struct Editor {
     /// avoid calculating the cursor position multiple
     /// times during rendering and should not be set by other functions.
     pub cursor_cache: Cell<Option<Option<Position>>>,
+
+    /// Contains all the cursor word highlights
+    pub cursor_highlights: Arc<Vec<std::ops::Range<usize>>>,
```
pascalkuthe, line 872: this needs to be either cleared when edits are applied to the document or (even better) mapped trough the changes applied to the document which would remove the need for the fallback implementation you added entirely. You can probably build on the improvements to change mapping I did in #6447

pascalkuthe (changes requested):

helix-term/src/ui/editor.rs
SoraTenshi: I am not sure if my updated check works as expected, but i think it's as close to figuring out whether lsp positions should be used or not.

helix-term/src/ui/editor.rs
```diff
@@ -264,6 +272,121 @@ impl EditorView {
...
+            None => return result,
+        };
+
+        let is_word_char = |c: Option<char>| -> bool {
+            if let Some(c) = c {
+                c.is_ascii_alphanumeric() || c == '-' || c == '_'
+            } else {
+                false
+            }
+        };
```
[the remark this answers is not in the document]
SoraTenshi: Thanks!

helix-term/src/commands/lsp.rs
```diff
@@ -1473,6 +1476,44 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+    let language_server = language_server!(cx.editor, doc);
+    let offset_encoding = language_server.offset_encoding();
+
+    let pos = doc.position(view.id, offset_encoding);
+
+    let future = match language_server.text_document_document_highlight(doc.identifier(), pos, None)
+    {
+        Some(future) => future,
+        None => return,
+    };
```
poliorcetics, line 1490:
```suggestion
    let Some(future) =  language_server.text_document_document_highlight(doc.identifier(), pos, None) else {
        return;
    };
```

I think helix is now in 1.65+ ?

helix-term/src/commands/lsp.rs
```diff
@@ -1473,6 +1476,44 @@ pub fn select_references_to_symbol_under_cursor(cx: &mut Context) {
...
+                        .iter()
+                        .filter_map(|highlight| {
+                            lsp_range_to_ops_range(text, highlight.range, offset_encoding)
+                        })
+                        .collect(),
+                );
+            } else {
+                // Reset the cursor_word
+                editor.cursor_highlights = Arc::new(Vec::new());
+            }
```
poliorcetics, line 1512:
```suggestion
            let Some(highlights) = response else {
                // Reset the cursor_word
                editor.cursor_highlights = Arc::new(Vec::new());
                return;
            };
            let (_, doc) = current!(editor);
            let language_server = language_server!(editor, doc);
            let offset_encoding = language_server.offset_encoding();
            let text = doc.text();

            editor.cursor_highlights = Arc::new(
                highlights
                    .iter()
                    .filter_map(|highlight| {
                        lsp_range_to_ops_range(text, highlight.range, offset_encoding)
                    })
                    .collect(),
            );
```

helix-term/src/ui/editor.rs
```diff
@@ -248,6 +257,103 @@ impl EditorView {
...
+        if !char_under_cursor.map_or(false, char_is_word) {
+            return None;
+        }
+
+        let chars_at_cursor = text.chars_at(cursor);
+        let reversed_chars = chars_at_cursor.clone().reversed();
+        let start = cursor.saturating_sub(reversed_chars.take_while(|c| char_is_word(*c)).count());
+        let end = cursor + chars_at_cursor.take_while(|c| char_is_word(*c)).count();
+
+        Some(text.slice(start..end).to_string())
```
poliorcetics, line 274: Is returning a `String` necessary here ? I think it could be a `&str` reference to the document and only converted to a `String` when the lifetime become annoying

helix-term/src/ui/editor.rs
```diff
@@ -248,6 +257,103 @@ impl EditorView {
...
+        viewport: Rect,
+        scope_index: usize,
+    ) -> Vec<(usize, std::ops::Range<usize>)> {
+        let text = doc.text().slice(..);
+        let mut result = Vec::new();
+
+        let cursor_word = match Self::cursor_word(doc, view) {
+            Some(cw) => cw,
+            None => return result,
+        };
```
poliorcetics, line 290:
```suggestion
        let Some(cursor_word) = Self::cursor_word(doc, view) else {
            return result;
        };
```

helix-term/src/ui/editor.rs
```diff
@@ -248,6 +257,103 @@ impl EditorView {
...
+        let line_range = {
+            // Calculate the lines in view
+            let last_line = text.len_lines().saturating_sub(1);
+            let last_visible_line = (row + viewport.height as usize).min(last_line);
+            let first_visible_line = row;
+
+            first_visible_line..last_visible_line
+        };
+
+        let relevent_lines = text
```
poliorcetics, line 302:
```suggestion
        let relevant_lines = text
```

typo

helix-term/src/commands/lsp.rs
SoraTenshi: Ahh good catch!

helix-term/src/ui/editor.rs
SoraTenshi: returning a `String` is not necessary. Thanks for finding that one! :)

PR #6202, opened 2023-03-06
Add comment-end-token language config option

MaelCravero:
Some languages lack line comments and only have block comments, in which case an end of comment token is needed to comment single lines.


pascalkuthe:
duplicate of  #4718 which includes general block comment handeling including falling back to block comments for single line comments

MaelCravero:
Oh I had completely missed that MR! Sorry for the inconvenience

PR #6204, opened 2023-03-06
Update highlights for golang

erasin:
- update tree-sitter-go
- refine keywords
- set package as namespace
- add label
- add error

### DIFF
--- a/runtime/queries/go/highlights.scm
+++ b/runtime/queries/go/highlights.scm
@@ -19,6 +19,9 @@
 (method_declaration
   name: (field_identifier) @function.method)
 
+(method_spec 
+  name: (field_identifier) @function.method) 
+
 ; Identifiers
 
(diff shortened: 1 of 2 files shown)


runtime/queries/go/highlights.scm
```diff
@@ -157,3 +190,7 @@
 ] @constant.builtin
 
 (comment) @comment
+
+; Error
+
+(ERROR) @error
```
the-mikedavis, line 196: For this PR and #6203: I'm not sure we should add the `(ERROR) @error` scopes. We've discussed this before and it might be best to remove them from all queries and add an option that automatically adds the highlight: https://github.com/helix-editor/helix/pull/3789#discussion_r967739912

erasin: The ERROR section has been removed.

the-mikedavis (approved):

Issue #4862, opened 2022-11-23
Freezes with 100% CPU when trying to open a specific Nix file


max-privatevoid:
### Summary

Probably related to https://github.com/helix-editor/helix/issues/2997

### Reproduction Steps

- Create a file with the content listed below
- In another terminal, open the file in Helix
- Observe it getting stuck before launching
- Edit the file: Remove a single character from a string
- Try opening the file with Helix again
- Observe it now working even though there should be no change to the AST
```nix
{ config, lib, pkgs, ... }:
with lib;
let
  # most of the things here should probably be incorporated into a module
  
  # TODO: make a proper type for env vars or pathspecs
  coerceToEnv = val: let
    parsed = match "^\\$([a-zA-Z0-9_]*)(/(.*))?$" val;
    key = elemAt parsed 0;
    append = elemAt parsed 2;
  in if isString val && parsed != null then
      if append != null then
        concat (env key) (concat "/" (coerceToEnv append))
      else
        env key
    else if isAttrs val && val ? _sloth then
      val._sloth
    else
      val;

  instanceId = { type = "instanceId"; };
  concat = a: b: { type = "concat"; inherit a b; };
  env = key: { type = "env"; inherit key; };

  splitPath = path: if isList path then {
    a = elemAt path 0;
    b = elemAt path 1;
  } else {
    a = path;
    b = path;
  };

  bind' = arg: path: let
    split = splitPath path;
    coerced = mapAttrs (_: coerceToEnv) split;
  in [ arg coerced.a coerced.b ];
  bind = bind' "--bind-try";
  bindRo = bind' "--ro-bind-try";
  bindDev = bind' "--dev-bind-try";
  setEnv = key: val: [ "--setenv" key val ];
  mountTmpfs = path: [ "--tmpfs" path ];
  
  bindPaths = map bind config.bubblewrap.bind.rw;
  bindRoPaths = map bindRo config.bubblewrap.bind.ro;
  bindDevPaths = map bindDev config.bubblewrap.bind.dev;
  envVars = mapAttrsToList setEnv config.bubblewrap.env;
  tmpfs = map mountTmpfs config.bubblewrap.tmpfs;

  app = config.app.package;
  info = pkgs.closureInfo { rootPaths = app; };
  launcher = pkgs.callPackage ../launcher {};
  dbusOutsidePath = concat (env "XDG_RUNTIME_DIR") (concat "/nixpak-bus-" instanceId);
  
  bwrapArgs = flatten [
[...]

PR #6218, opened 2023-03-07
Pin tree-sitter at git master

the-mikedavis:
Tree-sitter has some unreleased improvements that can speed up small queries and prevent hangs due to error recovery in some parsers. This change pins tree-sitter to the latest master.

Neovim also pins tree-sitter to a commit on master.

(Distributions could decide to ignore this patch if they don't allow git revisions - there are no changes in the Helix codebase that require the new changes to tree-sitter.)

Fixes https://github.com/helix-editor/helix/issues/4862
See https://github.com/tree-sitter/tree-sitter/pull/2085

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2154,8 +2154,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter"
 version = "0.20.9"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I am actually recently convinced that quite a few of the: helix is slow for large file xyz issues which were TS related would be solved by this. Many grammars for nontraditional programming languages endup very flat (think markdown, or c headed with comments + defines)

PR #6253, opened 2023-03-10
Add undercurl config option

### LINKED ISSUE #6196: Add configuration option to "forcefully" toggle undercurl when not detected.
# Description

Due to the way some distributions and package managers package the `TermInfo` file of various emulators, it is possible to have a false negative reading in Helix that the used emulator does not support undercurls. Even if it actually does (for example, Alacritty in ArchLinux shows as not supporting undercurls, when it acutally does).

# Proposal

A configuration option could be added to "force" Helix to use undercurls even if its automated detection mechanism says they are not supported.

jpttrssn:
Fixes  #6196. If set to 'true' this option will force terminal undercurl support.

I don't have ArchLinux so tested this by setting `undercurl = true` in my config file and running a version of helix where `has_extended_underlines` was only set by the config option.

I didn't find any related integration tests that could help me in writing one for this. If needed (and someone can point me in the right direction) I may be able to add that as well.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -53,6 +53,7 @@ signal to the Helix process on Unix operating systems, such as by using the comm
 | `completion-replace` | Set to `true` to make completions always replace the entire word and not just the part before the cursor | `false` |
 | `auto-info` | Whether to display info boxes | `true` |
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative | `false` |
+| `undercurl` | Set to `true` to override automatic detection of terminal undercurl support in the event of a false negative | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file | `[]` |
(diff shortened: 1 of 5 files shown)


pascalkuthe (dismissed):
LGTM thanks for taking this on :+1:  This will also be useful for people using `ssh`. Sadly terminfo can be unreliable both because some distros don't package it correctly but also because there is no solution for situations like `ssh`

pascalkuthe:
Looks like you forgot to update some tests @jpttrssn  sorry I forgot about tests and didn't wait for CI to finish before posting my review

jpttrssn:
I think I've fixed the tests. I didn't know how to run them! I created a PR that updates the contributing documentation on how to run tests for Rust newbies like myself.  #6268

pascalkuthe (approved):
LGTM now :+1:

archseer (approved):
Thanks! Sadly I need this option too, despite tinkering with my tmux config for ages I'm not able to get it working without an override :/

PR #6257, opened 2023-03-10
Check language server symbol renaming support before prompting

### LINKED ISSUE #6249: Check language server rename support before showing rename-to prompt
If a language server doesn't support LSP renaming, `rename_symbol` (`<space>r`) should bail early. Currently it creates the "rename-to" prompt and then fails to rename.

We check whether the server declares rename support here when creating the rename request future: https://github.com/helix-editor/helix/blob/d63e570e0a4013f5ad703a9b1ce2d19a06630a82/helix-lsp/src/client.rs#L1115-L1120

(The rename text entered in the prompt is used in the params for that request.) We should extract that match as a function:

```rust
pub fn supports_rename(&self) -> bool { /* check rename capability */ }
```

and then check that support before creating the prompt.

`bash-language-server` is a language server in the wild that doesn't support symbol renaming and can be used to test this.

misiasty3:
Tried my best to fix #6249
Tell me if there is something to change :smile:

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -1106,7 +1106,10 @@ impl Client {
 
     pub fn supports_rename(&self) -> bool {
         let capabilities = self.capabilities.get().unwrap();
-        matches!(capabilities.rename_provider, Some(lsp::OneOf::Left(true)) | Some(lsp::OneOf::Right(_)))
+        matches!(
+            capabilities.rename_provider,
+            Some(lsp::OneOf::Left(true) | lsp::OneOf::Right(_))
+        )
     }
 
     pub fn rename_symbol(
@@ -1115,6 +1118,10 @@ impl Client {
         position: lsp::Position,
         new_name: String,
     ) -> Option<impl Future<Output = Result<lsp::WorkspaceEdit>>> {
+        if !self.supports_rename() {
+            return None;
+        }
+
         let params = lsp::RenameParams {
             text_document_position: lsp::TextDocumentPositionParams {
                 text_document,
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -1106,11 +1106,7 @@ impl Client {
 
     pub fn supports_rename(&self) -> bool {
         let capabilities = self.capabilities.get().unwrap();
-        match capabilities.rename_provider {
-            Some(lsp::OneOf::Left(true)) | Some(lsp::OneOf::Right(_)) => true,
-            // None | Some(false)
-            _ => false,
-        }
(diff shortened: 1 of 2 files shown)


helix-lsp/src/client.rs
```diff
@@ -1104,21 +1104,21 @@ impl Client {
         Some(self.call::<lsp::request::CodeActionRequest>(params))
     }
 
+    pub fn supports_rename(&self) -> bool {
+        let capabilities = self.capabilities.get().unwrap();
+        match capabilities.rename_provider {
+            Some(lsp::OneOf::Left(true)) | Some(lsp::OneOf::Right(_)) => true,
+            // None | Some(false)
+            _ => false,
+        }
```
poliorcetics, line 1113:
```suggestion
        matches!(capabilities.rename_provider, Some(lsp::OneOf::Left(true)) | Some(lsp::OneOf::Right(_)))
```

poliorcetics: Even shorter:

```rust
matches!(capabilities.rename_provider, Some(lsp::OneOf::Left(true) | lsp::OneOf::Right(_)))
```

helix-lsp/src/client.rs
```diff
@@ -1104,21 +1104,21 @@ impl Client {
...
         new_name: String,
     ) -> Option<impl Future<Output = Result<lsp::WorkspaceEdit>>> {
-        let capabilities = self.capabilities.get().unwrap();
-
-        // Return early if the language server does not support renaming.
-        match capabilities.rename_provider {
-            Some(lsp::OneOf::Left(true)) | Some(lsp::OneOf::Right(_)) => (),
-            // None | Some(false)
-            _ => return None,
-        };
```
the-mikedavis, line 1120: This can be replaced with the new supports_rename function:

```rust
if !self.supports_rename() {
    return None;
}
```

If we don't check this then we are relying on any callers to call `supports_rename` themselves which could be a little error-prone.

helix-term/src/commands/lsp.rs
```diff
@@ -1290,8 +1290,7 @@ pub fn rename_symbol(cx: &mut Context) {
                     match language_server.rename_symbol(doc.identifier(), pos, input.to_string()) {
                         Some(future) => future,
                         None => {
-                            cx.editor
-                                .set_error("Language server does not support symbol renaming");
+                            cx.editor.set_error("Failed to rename symbol");
```
the-mikedavis, line 1293: This error message is still valid here - that function should only return `None` if the server doesn't support rename requests

helix-lsp/src/client.rs
misiasty3: Good idea.

the-mikedavis:
Thanks!

Issue #6243, opened 2023-03-10
Volar vue-language-server crashed when open vue files


sun617:
### Summary

Build latest master branch of helix to open vue file then volar of vue-language-server crashed
Maybe this is related to change of https://github.com/helix-editor/helix/pull/6058 

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:
open vue file the vue-language-server volar started
Instead, this happened:
language server crashed

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-03-10T13:42:10.533 helix_term::application [ERROR] Language Server: Method client/registerCapability not found in request 0
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "/home/sun/.npm/_npx/8b20ff0d5db79130/node_modules/vscode-jsonrpc/lib/common/connection.js:565\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "                        responsePromise.reject(new messages_1.ResponseError(error.code, error.message, error.data));\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "                                               ^\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "ResponseError: Method not found: client/registerCapability\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "    at handleResponse (/home/sun/.npm/_npx/8b20ff0d5db79130/node_modules/vscode-jsonrpc/lib/common/connection.js:565:48)\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "    at handleMessage (/home/sun/.npm/_npx/8b20ff0d5db79130/node_modules/vscode-jsonrpc/lib/common/connection.js:345:13)\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "    at processMessageQueue (/home/sun/.npm/_npx/8b20ff0d5db79130/node_modules/vscode-jsonrpc/lib/common/connection.js:362:17)\n"
2023-03-10T13:42:10.535 helix_lsp::transport [ERROR] err <- "    at Immediate.<anonymous> (/home/sun/.npm/_npx/8b20ff0d5db79130/node_modules/vscode-jsonrpc/lib/common/connection.js:334:13)\n"
[...]

Issue #6230, opened 2023-03-09
`bash-language-server` crashes after update


AlexanderBrevig:
### Summary

Using latest from branch Helix and latest from AUR `bash-language-server` `4.8.2` I get no LSP for bash.

I am not sure if this is an issue with Helix or the LSP, but maybe we could be more graceful about the issue?

Let me know if this is somehow better suited as an issue against the upstream LSP.

### Reproduction Steps

 

I tried this: Updated bash-language-server to latest version

I expected this to happen: Still get LSP for bash

Instead, this happened: No LSP for bash, since it crashes

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-03-09T11:11:21.175 helix_lsp::transport [INFO] -> {"jsonrpc":"2.0","result":[null],"id":1}
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "/usr/lib/bash-language-server/node_modules/vscode-jsonrpc/lib/common/connection.js:531\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "                        responsePromise.reject(new messages_1.ResponseError(error.code, error.message, error.data));\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "                                               ^\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "ResponseError: Method not found: client/registerCapability\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "    at handleResponse (/usr/lib/bash-language-server/node_modules/vscode-jsonrpc/lib/common/connection.js:531:48)\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "    at processMessageQueue (/usr/lib/bash-language-server/node_modules/vscode-jsonrpc/lib/common/connection.js:327:17)\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "    at Immediate.<anonymous> (/usr/lib/bash-language-server/node_modules/vscode-jsonrpc/lib/common/connection.js:311:13)\n"
2023-03-09T11:11:21.178 helix_lsp::transport [ERROR] err <- "    at processImmediate (node:internal/timers:466:21) {\n"
[...]

PR #6258, opened 2023-03-10
LSP: No-op client/registerCapability requests

the-mikedavis:
Some language servers send this request even though we don't enable dynamic capability registration. In some language servers that use the vscode-languageserver-node dependency, replying with an error can cause the server to exit because an un-caught rejected promise is fatal.

Replying with a JSONRPC error that the method is not handled is the correct thing to do but given that it causes some language servers in the wild to exit and that ignoring it doesn't cause any harm, we can workaround the exits by just sending an OK and doing nothing. This is similar to a workaround done by neovim.

Recent examples: #6230, #6243

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1014,6 +1014,17 @@ impl Application {
                             .collect();
                         Ok(json!(result))
                     }
+                    Ok(MethodCall::RegisterCapability(_params)) => {
+                        log::warn!("Ignoring a client/registerCapability request because dynamic capability registration is not enabled. Please report this upstream to the language server");
+                        // Language Servers based on the `vscode-languageserver-node` library often send
+                        // client/registerCapability even though we do not enable dynamic registration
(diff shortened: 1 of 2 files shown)


helix-term/src/application.rs
```diff
@@ -1014,6 +1014,17 @@ impl Application {
                             .collect();
                         Ok(json!(result))
                     }
+                    Ok(MethodCall::RegisterCapability(_params)) => {
+                        log::warn!("Ignoring a client/registerCapability request because dynamic capability registration is not enabled");
```
archseer, line 1018: I do like how the neovim error says to report the issue upstream: https://github.com/neovim/nvim-lspconfig/pull/471/files#diff-0e1bd346624fa540fa153e867fbc2036ced5c83d5cd7d79d9308df1328b7c879R58-R59

That way if someone spots it, it does actually get fixed upstream rather than considered a helix bug

PR #6278, opened 2023-03-12
Fix indentation level calculation for lines mixing tabs and spaces

### LINKED ISSUE #6262: Indentation level is calculated incorrectly for lines mixing tabs and spaces
Indentation of existing lines is calculated with (`helix_core::indent`):

``` rust
pub fn indent_level_for_line(line: RopeSlice, tab_width: usize, indent_width: usize) -> usize {
    let mut len = 0;
    for ch in line.chars() {
        match ch {
            '\t' => len += tab_width,
            ' ' => len += 1,
            _ => break,
        }
    }

    len / indent_width
}
```

This works for lines that are purely indented with spaces or tabs (so almost always) but in the edge-case that a lines is indented with a mix of tabs and spaces (for some ungodly reason) then the result is incorrect because the width of a tab differs in that case. The fix is simple: use `helix_core::graphemes::tab_width` to calculate the tab width instead in the function mentioned above.

CC @Triton171 (in case I missed something)

stabor705:
Should be an easy fix that closes #6262

### DIFF
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -4,6 +4,7 @@ use tree_sitter::{Query, QueryCursor, QueryPredicateArg};
 
 use crate::{
     chars::{char_is_line_ending, char_is_whitespace},
+    graphemes::tab_width_at,
     syntax::{LanguageConfiguration, RopeProvider, Syntax},
     tree_sitter::Node,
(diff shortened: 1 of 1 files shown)


pascalkuthe (approved):
LGTM, thanks!

Issue #6267, opened 2023-03-11
V and Vue seem incompatible


bfg-coding:
### Summary

I've downloaded V and got it's lsp setup but for some reason when I use `:set-language` to V it seems to be using my Vue lsp as my auto completes are showing Vue things but the syntax highlighting is in V.  They have the same name in the `hx --health` vls but when I look at the language.toml I don't see anything that seems to be causing Helix to confuse the use of two.  

Is anyone else having this issue when they have both vue and v lsp installed on their machine?

### Reproduction Steps

I tried this:

1. open v file `hx .`  in a V project
2. run `:set-language` and select V as there is another bug being reported on some file name oddities
3 . Trigger code suggestion

I expected this to happen:

see V code suggestion and autocompletion

Instead, this happened:

I see Vue's suggestion and auto completion

### Helix log

2023-03-10T14:04:46.381 helix_view::editor [ERROR] Failed to initialize the LSP for `source.verilog` { cannot find binary path }
2023-03-10T14:05:21.311 helix_view::editor [ERROR] Failed to initialize the LSP for `source.verilog` { cannot find binary path }
2023-03-10T14:12:40.584 helix_view::editor [ERROR] Failed to initialize the LSP for `source.verilog` { cannot find binary path }
2023-03-10T14:13:13.834 helix_term::application [ERROR] Language Server: Method not found client/registerCapability
2023-03-10T14:13:14.068 helix_term::application [ERROR] Language Server: Method not found window/showMessageRequest
2023-03-10T14:13:14.068 helix_term::application [ERROR] Language Server: Method not found window/showMessageRequest
2023-03-10T21:50:56.799 helix_view::editor [ERROR] Failed to initialize the LSP for `source.verilog` { cannot find binary path }
2023-03-10T21:51:04.875 helix_term::application [ERROR] Language Server: Method not found client/registerCapability
2023-03-10T21:51:05.110 helix_term::application [ERROR] Language Server: Method not found window/showMessageRequest
2023-03-10T21:51:05.110 helix_term::application [ERROR] Language Server: Method not found window/showMessageRequest
2023-03-10T21:55:14.994 helix_view::editor [ERROR] Failed to initialize the LSP for `source.verilog` { cannot find binary path }
2023-03-10T21:55:22.243 helix_term::application [ERROR] Language Server: Method not found client/registerCapability
2023-03-10T21:55:22.474 helix_term::application [ERROR] Language Server: Method not found window/showMessageRequest
[...]

Issue #6164, opened 2023-03-03
Vlang syntax not work by default


linc01:
### Summary

Although there are some configurations available for Vlang in the languages.toml file by default, they do not seem to work as documented. Please advise on how to enable vlang-lsp, as it is working well with my Vim environment.

Thanks!

### Platform

Linux

### Terminal Emulator

Qt

### Helix Version

22.12

kirawi:
Thanks for opening an issue. Can you check https://github.com/helix-editor/helix/wiki/Troubleshooting#missing-syntax-highlighting?

linc01:
Thanks kirawi, here is:
```
Configured language server: vls
Binary for language server: Not found in $PATH
Configured debug adapter: None
Highlight queries: ✓
Textobject queries: ✘
Indent queries: ✘
```
and the vls $PATH(`export PATH=~/.vls/bin/vls_linux_x64:$PATH`) is there, but it's not found in the above check.

linc01:
@askreet Thank you for your response,! Unfortunately, I have already attempted using the `bin` file at first, but not work. And it's not affect the syntax highlight, right?

askreet:
> Unfortunately, I have already attempted using the bin file at first, but not work.

When you added only the `bin` directory to your path, did `hx --health v` continue to show "Not found in $PATH"? If you have a binary called `vls_linux_x64` in a directory called `bin` that will not work, as (per the output of `--health`) Helix is looking to find a binary called `vls`.

> And it's not affect the syntax highlight, right?

Correct, the language server has nothing to do with syntax highlighting. The fact that your output shows `Highlight queries: ✓` indicates the treesitter queries exist for the language and should be highlighting. Does the correct filetype show when you use `:set-language`?

I don't know anything about V or Vue, but `hx --health v` appears to want a language server called `v` while `hx --health vue` wants one called `vls` -- which type of file are you working with?

linc01:
`.v` file for Vlang

this is the current highlight:
![vlang](https://user-images.githubusercontent.com/30107747/222932758-939176fa-768f-4cf5-8d5f-4e104660d5c4.png)

seems the highlight works, but not very well.

kirawi:
That's quite different from "doesn't work by default." The highlighting queries for V result in simple highlighting: https://github.com/helix-editor/helix/blob/master/runtime/queries/v/highlights.scm

Someone would need to put in the work to get more detailed highlighting.

bfg-coding:
[...]

PR #6279, opened 2023-03-12
Improve tree-sitter queries for vlang

erasin:
- Update highlight
- add indents
- add textobject
- add injections

Issues #6164 and #6267.

vlang and verilog has file type conflicts in languages configurations.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -140,7 +140,7 @@
 | typescript | ✓ | ✓ | ✓ | `typescript-language-server` |
 | ungrammar | ✓ |  |  |  |
 | uxntal | ✓ |  |  |  |
-| v | ✓ |  |  | `v` |
+| v | ✓ | ✓ | ✓ | `v` |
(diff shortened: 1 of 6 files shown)


the-mikedavis:
https://github.com/helix-editor/helix/pull/3249 has an approach for improving language conflicts that might help with vlang/verilog

erasin:
> It looks like this entry beats regular html for .html files. We have another conflict like this between verilog and V and I think the only way around it for now is to have one of them where you have to override it in a .helix/languages.toml or ~/.config/helix/languages.toml:

Do you mean this way, I'm already using it with html/twig，#6164. 

In fact, the solution I want for conflict resolution is vscode's [files.associations](https://code.visualstudio.com/docs/languages/identifiers).

the-mikedavis (approved):

the-mikedavis:
#3249 would help even when the file-types are the same for both languages. What you recommend in #6164 looks like a good workaround for now

PR #6281, opened 2023-03-12
Test indentation level calculation

jpttrssn:
This uses @stabor705's fix for #6262 and adds tests that fail with the current implementation, but pass with this fix.


dead10ck (approved):
This looks right to me, thanks!

the-mikedavis:
(Rebased on top of #6262)

the-mikedavis (approved):

PR #6282, opened 2023-03-12
Discard LSP requests/notifcation that target outdated versions

pascalkuthe:
This is a slight followup to #6173

I noticed that similar bugs could occur basically anytime the LS sends an edit to the client. The user can always add some new text which makes the received text invalid. However, for these edits the LSP standard actually includes a mechanism to deal with the problem: Whenever we send a change a notification to the server, the client must attach a unique version to the resulting document state. 

If the LSP server sends an edit to the client, that edit includes the document version it applies to and if that version doesn't match the current version then that change can not be applied/should be discarded.

For workspace edits specifically there are actually many capabilities which include the exact behavior here, but we didn't set any of them. Even worse the client is supposed to return that applying the edits failed but we always returned a success. I have fixed that by implemented the `abort` failure mechanism for now (and setting the capabilities appropriately) as it was the easiest to implement (basically just propagating a `Result` with `?`). In the future we likely want to check whether a `WorkspaceEdit` is valid before we start apply it and switch to transactional failure handling. I think that is how VSCode handles it. This way we won't end-up with half-applied changes (a good LS would resend the missing edits, but I suspect many don't). Implementing this is a bit more work so I left that to future work. The current failure handling strategy is spec compliant and avoids apply incorrect edits/crashes so I wanted a quick fix for that first.

`PublishDiagnostic` notifications also send a document version and this PR also adds a check to drop outdated `PublishDiagnostic` notifications. This fixes diagnostics "lagging behind" while typing.

# Future Work

I noticed that we don't normalize line endings while applying LSP edits. We currently do this for snippets but not anywhere else. Furthermore, we also need to normalize indentation (replace tabs with spaces and vice versa). I will create an issue for that. Once we do that we should also set the appropriate capability to true (I set it to `false` for now to reflect what we are doing

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -387,6 +387,7 @@ impl Client {
                         ..Default::default()
                     }),
                     publish_diagnostics: Some(lsp::PublishDiagnosticsClientCapabilities {
+                        version_support: Some(true),
                         ..Default::default()
                     }),
                     inlay_hint: Some(lsp::InlayHintClientCapabilities {
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -318,6 +318,17 @@ impl Client {
                     inlay_hint: Some(lsp::InlayHintWorkspaceClientCapabilities {
                         refresh_support: Some(false),
                     }),
+                    workspace_edit: Some(lsp::WorkspaceEditClientCapabilities {
+                        document_changes: Some(true),
+                        resource_operations: Some(vec![
+                            lsp::ResourceOperationKind::Create,
+                            lsp::ResourceOperationKind::Rename,
+                            lsp::ResourceOperationKind::Delete,
+                        ]),
+                        failure_handling: Some(lsp::FailureHandlingKind::Abort),
+                        normalizes_line_endings: Some(false),
+                        change_annotation_support: None,
+                    }),
                     ..Default::default()
                 }),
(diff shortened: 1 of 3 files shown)


pascalkuthe:
(sorry messed something up while reabsing)

helix-lsp/src/client.rs
```diff
@@ -318,6 +318,17 @@ impl Client {
                     inlay_hint: Some(lsp::InlayHintWorkspaceClientCapabilities {
                         refresh_support: Some(false),
                     }),
+                    workspace_edit: Some(lsp::WorkspaceEditClientCapabilities {
+                        document_changes: Some(true),
+                        resource_operations: Some(vec![
+                            lsp::ResourceOperationKind::Create,
```
dead10ck, line 324: I don't think we actually handle these, do we? Should we not advertise these capabilities?

dead10ck (approved):
I just had one minor question, but this all looks good to me, thanks for fixing these!

helix-lsp/src/client.rs
pascalkuthe: we do actually but it seems we just forgot to add the capability (many LS were sending these regardless :D):

https://github.com/helix-editor/helix/blob/14f1a62367fe70456201c0a64174b643cf294fdd/helix-term/src/commands/lsp.rs#L860

dead10ck: Ahh ok cool

the-mikedavis (approved):

PR #6290, opened 2023-03-13
Canonicalize paths before stripping current dir as prefix

### LINKED ISSUE #6182: LSP (clangd) actions open duplicate buffers
When I perform navigational code actions through the LSP (clangd), e.g., gd (goto definition), it opens a duplicate buffer with the absolute path rather than the path relative to pwd. This buffer is separate from the original buffer even though they refer to same file, so I can unknowingly write changes to this duplicate buffer, then overwrite those changes in the original buffer, which is frustrating since I end up losing my work.

For example, within buffer "src/code.cpp" -> gd -> opens buffer "C:/full/path/to/src/code.cpp" (cursor at the definition).

I expect the correct behavior to be for it to resolve the LSP's requested file path relative to pwd such that
a) It doesn't open the buffer with an absolute path if it's under pwd and
b) subsequently, doesn't open a duplicate buffer

jazzfool:
Fixes #6182


the-mikedavis (approved):

pascalkuthe (changes requested):
This is not the right fix.

Yes this does fix the very specific instance you ran into but it doesn't adress the root cause of the problem. Right now you only canonicalize if it exists. (also you crash the editor if CWD doesn't exist, we do that elsewhere too but we **really** shouldn't do that). If `clangd` had sent a message for a file that is not yet saved to disk (but exists as a buffer in helix) the same problem would occur. Furthermore there also many other places where we canonicalize paths in helix where having the normal canonicalize function being unreliable is a big problem.

As I described here: https://github.com/helix-editor/helix/issues/6182#issuecomment-1465424919 you need to change `get_normalized_path` to canonicalize the **first parent that exists** (and normalize the rest with the existing code). You can then use that function to normalize cwd (and the argument) in `get_relative_path`

helix-core/src/path.rs
```diff
@@ -40,6 +40,24 @@ pub fn expand_tilde(path: &Path) -> PathBuf {
...
+    let (base, path) = path
+        .ancestors()
+        .find(|path| path.exists())
+        .and_then(|path| Some((path, dunce::canonicalize(path).ok()?)))
+        .map(|(base, canonicalized)| {
+            (
+                canonicalized,
+                path.strip_prefix(base).unwrap(/* base is an ancestor of path */).into(),
+            )
+        })
```
pascalkuthe, line 54:
```suggestion
        .find_map(|base| { 
            let canonicalized_base = dunce::canonicalize(base).ok()?;
            Some((canonicalized_base,  path.strip_prefix(base)))
        })
```
just because a path exists doesn't mean cannonicalization necessary successed (Sadly). Using `find_map` ensures we only stoip once canonicalization succedes. It's also a bit cleaner.

pascalkuthe (approved):
This looks great now, thanks for sticking with it! one minor nit

pascalkuthe (changes requested):
(didn't mean to approve yet until you accept my suggestion :sweat_smile: )

Test suite failures look unrelated to me. These macos failures have been popping up frequently and I don't think they are actually related to this PR

pascalkuthe (approved):
LGTM now

the-mikedavis (commented):
Ah whoops I didn't see the conflict with #6156. Could you rebase this?

the-mikedavis (approved):

Issue #6297, opened 2023-03-13
Whitespace inside LSP hints should not render


EpocSquadron:
The new inlay hints functionality introduced in #5934 interacts in an interesting way with the whitespace rendering functionality. That is to say, if you have `whitespace.render = "all"` set, you will see that spaces within the inlay hints themselves -- either spacing around the hint or between characters within it -- are displayed as any other whitespace in the document. Here's a screenshot:

![image](https://user-images.githubusercontent.com/414963/224775225-ca82b8fc-093b-4982-bdc4-c125c660c824.png)

To me this is not intuitively correct, because the inlay hints are virtual text that are not part of the real document. I rely on the whitespace rendering to provide information on the characters that are part of the document that I can't see, to help understand how motions and selections will behave before I use them. Displaying them within the inlay hint is neither informative nor helpful in that goal.

Further problematic is the spacing after the inlay hints is not rendered with a background to make it obvious it is part of the inlay hint and not a real space in the document. This makes it downright misleading to render the inlay hint whitespace. Here is the above example with inlay hints off, so you can see the difference:

![image](https://user-images.githubusercontent.com/414963/224775777-c0ae57f6-62a0-43da-8be9-2b0fa08f72c9.png)

erasin:
The theme onelight/onedark_local I'm using, whitespace is not noise to code. Most format tool can handle excess white space.


onelight
![WX20230314-155532](https://user-images.githubusercontent.com/716514/224948244-7657ba22-8bd1-475a-885a-1cd134ccd245.png)

[onedark_local](https://github.com/erasin/helix-config/blob/main/themes/onedark_local.toml)
![WX20230314-155511](https://user-images.githubusercontent.com/716514/224948257-711ee2ab-fc88-4cb3-bb19-a3a8e5f0b66a.png)

PR #6312, opened 2023-03-14
Prevent whitespace from rendering inside inlay hints

Zenthial:
Addresses most of #6297. 

Would like some advice on what should be done about tab rendering. Is it even needed to be addressed because I don't think the inlay hints would draw a tab?

### DIFF
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -413,15 +413,15 @@ impl<'a> TextRenderer<'a> {
         let width = grapheme.width();
         let space = if is_virtual { " " } else { &self.space };
         let nbsp = if is_virtual { " " } else { &self.nbsp };
+        let tab = if is_virtual {
+            &self.virtual_tab
+        } else {
+            &self.tab
+        };
         let grapheme = match grapheme {
             Grapheme::Tab { width } => {
-                if !is_virtual {
-                    let grapheme_tab_width = char_to_byte_idx(&self.tab, width);
-                    &self.tab[..grapheme_tab_width]
-                } else {
-                    let grapheme_tab_width = char_to_byte_idx(&self.virtual_tab, width);
-                    &self.virtual_tab[..grapheme_tab_width]
-                }
+                let grapheme_tab_width = char_to_byte_idx(&tab, width);
+                &tab[..grapheme_tab_width]
             }
             // TODO special rendering for other whitespaces?
             Grapheme::Other { ref g } if g == " " => space,
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -287,9 +287,11 @@ pub fn render_text<'t>(
             style_span.0
         };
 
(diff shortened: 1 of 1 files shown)


archseer:
An empty file named `origin` was pushed

helix-term/src/ui/document.rs
```diff
@@ -287,9 +287,11 @@ pub fn render_text<'t>(
             style_span.0
         };
 
+        let virt = grapheme.is_virtual().clone();
```
pascalkuthe, line 290:
```suggestion
        let virt = grapheme.is_virtual();
```

Doesn't really make a difference but cloning a `bool` is kind of pointless

helix-term/src/ui/document.rs
```diff
@@ -405,14 +408,16 @@ impl<'a> TextRenderer<'a> {
         }
 
         let width = grapheme.width();
+        let space = if is_virtual { " " } else { &self.space };
+        let nbsp = if is_virtual { " " } else { &self.nbsp };
         let grapheme = match grapheme {
             Grapheme::Tab { width } => {
                 let grapheme_tab_width = char_to_byte_idx(&self.tab, width);
```
pascalkuthe, line 415: To properly render tabs you need to add a second field `virtual_tab` to which gets filled with white spaces in `new` instead of using whitespace chars.

pascalkuthe (changes requested):

helix-term/src/ui/document.rs
Zenthial: Had this as a relic when I was working with &bool

helix-term/src/ui/document.rs
Zenthial: Gotcha. I believe its a bit verbose inside the Grapheme::Tab, but it should now handle the tab rendering

pascalkuthe: Looks good now but I would prefer if you could move the if condition outside the match and handle it just like the other variables (so `let tab = ..`).

I thought it was quite clean how you did it for the other whitespace chars so it would be great if tab was the same

Zenthial: That makes sense, done.

pascalkuthe (dismissed):
LGTM, thanks for taking this on :+1:

pascalkuthe:
ah actually you need to fix the clippy lint

pascalkuthe (approved):

helix-term/src/ui/document.rs
```diff
@@ -313,6 +315,7 @@ pub struct TextRenderer<'a> {
     pub nbsp: String,
     pub space: String,
     pub tab: String,
+    pub virtual_tab: String,
```
pickfire, line 318: Must this be `String`? Can it be integer instead?

pascalkuthe: No we need the allocation to be able to hand out `&str` to it, just like the `tab` field. The tab width is not the interesting part here, we already know that from the grapheme.

PR #6317, opened 2023-03-14
Update tree-sitter-haskell

ocharles:


ocharles:
Can anyone explain what CI is trying to tell me?

```
Error: "Failed to parse textobjects.scm queries for haskell: Query error at 5:4. Invalid node type decl_type"
```

Does this mean `decl_type` no longer exists in `tree-sitter-haskell` or something?

the-mikedavis:
Yeah that's a failure of query analysis within tree-sitter - it's saying that it isn't possible to match `decl_type`. That usually means that the node doesn't exist but it can also mean that it won't show up in the tree, for example if it is aliased to another node name. I believe that's the case here: https://github.com/tree-sitter/tree-sitter-haskell/blob/fb3c19e8e307acaf9336ab88330fd386ce731638/grammar.js#L232

the-mikedavis (approved):

PR #6320, opened 2023-03-14
feat: lint for `ui.virtual` and catch present and future virtual based features

### LINKED ISSUE #2070: LSP Type Hints

AlexanderBrevig:
Since #5934 closed #2070 I think the linter should check for these.

Maybe it's about time that we run the linter against PRs with changes to `/runtime/themes`?

EDIT: this now lints for `ui.virtual` as a catch all for present and future virtual text.

### DIFF
--- a/xtask/src/themelint.rs
+++ b/xtask/src/themelint.rs
@@ -36,15 +36,14 @@ fn get_rules() -> Vec<Require> {
         Require::Difference("ui.statusline.normal", "ui.statusline.select"),
         // Check for editor.cursorline
         Require::Existence(Rule::has_bg("ui.cursorline.primary")),
+        // Check for general ui.virtual (such as inlay-hint)
+        Require::Existence(Rule::has_fg("ui.virtual")),
         // Check for editor.whitespace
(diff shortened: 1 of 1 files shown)


archseer (changes requested):
But both of these are optional! I only intend to use `ui.virtual` on the default theme.

AlexanderBrevig:
> But both of these are optional! I only intend to use `ui.virtual` on the default theme.

Maybe you're right that this setting will not feel broken given the existence of a virtual key.

I'm just trying to help make the feeling of helix (and especially changing themes) a bit less broken.

I'll check as soon as I get on my computer today and close this if setting inlay-hints simply works across the board. I should've checked...

Thanks

AlexanderBrevig:
I agree. The themes that look off/wrong are all missing `ui.virtual` catch all, so this now lints against that instead.

PR #6324, opened 2023-03-15
adjust inlay hints, color modes and virtual for fleet dark

krfl:
As Fleet does not have these, this is an interpretation and adaptation for helix.

* Adjust inlay hints to make them blend more into the background and not stand out as much.
* Bring back color modes
* EDIT: Adjusted virtual + ruler while I was at it

No errors from themelint

Screenshot 2023-03-15 at 13 07 27


pickfire (approved):
Thanks for the quick update.

PR #6325, opened 2023-03-15
Add language support for Cap’n Proto format

angelodlfrtr:
Tree-sitter grammar and queries from https://github.com/amaanq/tree-sitter-capnp.

### DIFF
--- a/runtime/queries/capnp/highlights.scm
+++ b/runtime/queries/capnp/highlights.scm
@@ -1,0 +1,141 @@
+; Preproc
+
+(unique_id) @preproc
+(top_level_annotation_body) @preproc
+
+; Includes
+
+[
+  "import"
+  "$import"
+  "embed"
+] @include
+
+(import_path) @string
+
+; Builtins
+
+[
+  (primitive_type)
+  "List"
+] @type.builtin
+
+; Typedefs
+
+(type_definition) @type.definition
+
+; Labels (@number, @number!)
+
+(field_version) @label
+
+; Methods
+
+(annotation_definition_identifier) @method
+(method_identifier) @method
+
+; Fields
+
+(field_identifier) @field
+
+; Properties
+
+(property) @property
+
+; Parameters
+
+(param_identifier) @parameter
+(return_identifier) @parameter
+
+; Constants
+
+(const_identifier) @constant
+(local_const) @constant
+(enum_member) @constant
+
+(void) @constant.builtin
+
+; Types
+
+(enum_identifier) @type
+(extend_type) @type
+(type_identifier) @type
+
+; Attributes
+
+(annotation_identifier) @attribute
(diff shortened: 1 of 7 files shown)


runtime/queries/capnp/highlights.scm
```diff
@@ -0,0 +1,141 @@
+; Preproc
+
+(unique_id) @preproc
```
the-mikedavis, line 3: It looks like these queries use neovim captures. We use a slightly different set of captures documented here: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example, this should be `@keyword.directive`, `@include` should be `@keyword.control.include`, and `@method` should be `@function.method`

runtime/queries/capnp/injections.scm
```diff
@@ -0,0 +1 @@
+(comment) @comment
```
the-mikedavis, line 1:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
```

runtime/queries/capnp/locals.scm
the-mikedavis, line 1: For locals we use the captures `@local.scope`, `@local.definition` and `@local.reference`

runtime/queries/capnp/highlights.scm
angelodlfrtr: Updated in [ac50eb3](https://github.com/helix-editor/helix/pull/6325/commits/ac50eb3fd5ead297a5ef5c277f99b4057ed927be)

angelodlfrtr: Thanks for the review! The result is much better with the correct set of captures.

runtime/queries/capnp/indents.scm
```diff
@@ -0,0 +1,34 @@
...
+  (struct)
+  (struct_shorthand)
+  (union)
+] @indent
+
+((struct_shorthand (property)) @aligned_indent
+  (#set! "delimiter" "()"))
+
+((const_list (const_value)) @aligned_indent
+  (#set! "delimiter" "[]"))
```
pickfire, line 20: I don't think we have `@aligned_indent`?

angelodlfrtr: Thanks @pickfire 
From https://docs.helix-editor.com/guides/indent.html#adding-indent-queries, no.
I replaced `@indent_end` by `@outer`, removed `@branch` and `@aligned_indent`.
Only "scope" property is supported from https://github.com/helix-editor/helix/blob/master/helix-core/src/indent.rs#L433

angelodlfrtr:
Hi @pickfire @the-mikedavis, thanks for your reviews ! Does the changes seems good to you ?

the-mikedavis (approved):

PR #6331, opened 2023-03-16
add goto next/prev error (bound to ]e/[e)

robinvd:
some of my projects have a warnings from very verbose linters, this helps navigate the files when fixes errors.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -11,6 +11,7 @@ pub use typed::*;
 
 use helix_core::{
     char_idx_at_visual_offset, comment,
+    diagnostic::Severity,
     doc_formatter::TextFormat,
     encoding, find_first_non_whitespace_char, find_root, graphemes,
     history::UndoKind,
@@ -335,6 +336,8 @@ impl MappableCommand {
         goto_last_diag, "Goto last diagnostic",
         goto_next_diag, "Goto next diagnostic",
         goto_prev_diag, "Goto previous diagnostic",
+        goto_next_error, "Goto next error diagnostic",
+        goto_prev_error, "Goto prev error diagnostic",
         goto_next_change, "Goto next change",
         goto_prev_change, "Goto previous change",
         goto_first_change, "Goto first change",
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -3017,6 +3020,57 @@ fn goto_prev_diag(cx: &mut Context) {
     doc.set_selection(view.id, selection);
 }
 
+fn goto_next_error(cx: &mut Context) {
```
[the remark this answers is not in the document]
robinvd: done

helix-term/src/commands.rs
```diff
@@ -335,6 +336,8 @@ impl MappableCommand {
         goto_last_diag, "Goto last diagnostic",
         goto_next_diag, "Goto next diagnostic",
         goto_prev_diag, "Goto previous diagnostic",
+        goto_next_error, "Goto next error diagnostic",
+        goto_prev_error, "Goto prev error diagnostic",
```
pickfire, line 340:
```suggestion
        goto_next_error, "Goto next error",
        goto_prev_error, "Goto prev error",
```
I guess error should be enough?

helix-term/src/commands.rs
```diff
@@ -2972,19 +2975,31 @@ fn goto_last_diag(cx: &mut Context) {
...
+            .clone()
+            .find(|diag| diag.range.start > cursor_pos)
+            .or_else(|| iter.next()),
+        Direction::Backward => {
+            let mut iter = iter.rev();
+            iter.clone()
+                .find(|diag| diag.range.start < cursor_pos)
+                .or_else(|| iter.next())
+        }
+    };
```
pickfire, line 3002:
```suggestion
    let diag = match direction {
        Direction::Forward => iter
            .find(|diag| diag.range.start > cursor_pos)
            .or_else(|| iter.next()),
        Direction::Backward => {
            let mut iter = iter.rev();
            iter.find(|diag| diag.range.start < cursor_pos)
                .or_else(|| iter.next())
        }
    };
```
Unnecessary clone

helix-term/src/commands.rs
```diff
@@ -2972,19 +2975,31 @@ fn goto_last_diag(cx: &mut Context) {
...
 
     let selection = match diag {
         Some(diag) => Selection::single(diag.range.start, diag.range.end),
```
pickfire, line 3005: Can you please change this to let else?
```rust
if let Some(diag) = diag else {
    return;
};

helix-term/src/commands.rs
robinvd: The clone is necessary as find exhausts the iterator, thus in the suggestion or_else(|| iter.next()) will always return None

helix-term/src/commands.rs
robinvd: Thx! Ill update the pr with the suggestions when i have time

helix-term/src/commands.rs
pickfire: Should we leave a comment here so we know it shouldn't exhaust the iterator?

robinvd:
@pickfire thx for the review! sorry about the delay. I've applied your feedback.

cgahr:
`goto_first_error` and `goto_last_error` is also missing, for completeness they should probably also be implemented.

beeb:
This is probably superseeded by https://github.com/helix-editor/helix/pull/8717 (this should be closed)

PR #6333, opened 2023-03-16
Fix highlighting in picker with multiple columns

### LINKED ISSUE #6254: Picker fuzzy highlighting is offset when using more than one column
### Summary

In pickers that use more than one column (such as the buffer picker, then only one for now I think), the highlight of the fuzzy matching is offset from the second column.

image

Discovered this issue when working on #2869 .

In `helix-term/src/ui/picker.rs:218`, I noticed that, in the following code, for each span the `grapheme_byte_offset` is reset to 0. 

```rust
let graphemes_with_style: Vec<_> = spans
    .0
    .iter()
    .flat_map(|span| {
        span.content
            .grapheme_indices(true)
            .zip(std::iter::repeat(span.style))
    })
    .map(|((grapheme_byte_offset, grapheme), style)| {
        cell_len += grapheme.len();
        let start = cell_start_byte_offset;

        let grapheme_byte_range =
            grapheme_byte_offset..grapheme_byte_offset + grapheme.len();

        if highlight_byte_ranges.iter().any(|hl_rng| {
            hl_rng.start >= start + grapheme_byte_range.start
                && hl_rng.end <= start + grapheme_byte_range.end
        }) {
            (grapheme, style.patch(highlight_style))
        } else {
            (grapheme, style)
        }
    })
    .collect();
```

This may be the source of the offset we see.

### Reproduction Steps

1. open the buffer picker
2. type in some chars and see the highlighted portion of the line

### Helix log

This issue is not visible in the logs.

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

master build

exp80:
Fixes #6254

This could be probably done in a less hacky way, but I didn't want to overcomplicate things.

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -812,7 +812,10 @@ impl<T: Item + 'static> Component for Picker<T> {
                 for cell in row.cells.iter_mut() {
                     let spans = match cell.content.lines.get(0) {
                         Some(s) => s,
-                        None => continue,
+                        None => {
+                            cell_start_byte_offset += TEMP_CELL_SEP.len();
+                            continue;
+                        }
                     };
 
                     let mut cell_len = 0;
--- a/helix-tui/src/text.rs
+++ b/helix-tui/src/text.rs
@@ -453,10 +453,12 @@ impl<'a> From<&Text<'a>> for String {
         let mut output = String::with_capacity(size);
 
         for spans in &text.lines {
+            if !output.is_empty() {
+                output.push('\n');
+            }
             for span in &spans.0 {
                 output.push_str(&span.content);
             }
-            output.push('\n');
         }
         output
     }


the-mikedavis (changes requested):
Adding to `cell_start_byte_offset` in the `None` branch is a nice catch and that should stay 👍. The other change is not the right fix though. The offset is caused by `Row::cell_text` appending newlines to each cell's text and that should be fixed directly instead. It's caused by a bug introduced in [`b2837ff`](https://github.com/helix-editor/helix/commit/b2837ff3bea286ce7ccfba9b6fbcd861977caf83) in [`impl<'a> From<&Text<'a>> for String`](https://github.com/helix-editor/helix/blob/ac2a7731a6d04b8b913f49531058c7e2a843efd5/helix-tui/src/text.rs#L445-L463) where each line gets a newline appended. That newline should only be added between lines though. There's a small refactor of that `for` that fixes the issue:

```rust
for spans in &text.lines {
    if !output.is_empty() {
        output.push('\n');
    }
    for span in &spans.0 {
        output.push_str(&span.content);
    }
}
```

exp80:
Thanks, I didn't know if that was intentional or not. Fixed.

the-mikedavis (approved):

PR #6342, opened 2023-03-17
theme: update adwaita-dark for bufferline

ohno418:
The adwaita-dark theme has no distinction between active and inactive buffers on a bufferline. Add contrast to clarify differences.


archseer:
\cc @Mofiqul

Mofiqul:
> 

It looks good to me

the-mikedavis (approved):

PR #6345, opened 2023-03-17
Update onedark theme

erasin:
- sort config key by theme document
- Multi-cursor diff
- keyword
- add virtual text
- add makeup.heading
- ui.popup
- ...

@sudormrfbin Can I modify this theme file, if not, I will close this pr.

WX20230317-234833

after.

WX20230317-234816

### DIFF
--- a/runtime/themes/onedark.toml
+++ b/runtime/themes/onedark.toml
@@ -1,72 +1,114 @@
 # Author : Gokul Soumya <gokulps15@gmail.com>
 
 "attribute" = { fg = "yellow" }
-"comment" = { fg = "light-gray", modifiers = ["italic"] }
-"constant" = { fg = "cyan" }
-"constant.numeric" = { fg = "gold" }
-"constant.builtin" = { fg = "gold" }
-"constant.character.escape" = { fg = "gold" }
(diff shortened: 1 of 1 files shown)


sudormrfbin:
I haven't looked at the changes yet, but there seems to be changes to the colors used for keywords and some other scopes? I directly ported over the colors from joshdick/onedark.vim and I'd like to mirror the original onedark theme as close as possible. Do you have any other reference for changing the colors? It could be that I made a mistake in porting the scopes since vim uses different highlight group names.

erasin:
I compare part of [vscode's onedark Pro](https://marketplace.visualstudio.com/items?itemName=zhuangtongfa.Material-theme) theme and modify the color of keyword. I used the [Vim-one](https://github.com/rakr/vim-one) theme when I used vim earlier, and I also contributed to it. In atom, ide, vscode, vim and other editors, the onedark theme has some small difference.

I modify the color of `keyword.control` , this gives the code a layered feel.  Virtual text are grayscale processing, virtual text should not be too conspicuous, only provide auxiliary functions.

The theme provides a more comprehensive configuration, so that users can customize the modification.

PR #6369, opened 2023-03-19
Add `switch_statement` to ecma indents

thoradam:


thoradam:
Using `switch_body` also works, is that preferable?

the-mikedavis (commented):
Actually I think we want a few additions here to make the indents really correct. That `switch_statement` is good and shouldn't be a `switch_body`. `switch_statement` gives the value good indentation:

```javascript
switch (
  very_long_expression_broken_on_a_line
) {
  // ..
}
```

We should also indent and extend the `switch_case` and `switch_default` nodes though so that the bodies of those are indented as well:

```scheme
[
  (switch_case)
  (switch_default)
] @indent @extend
```

thoradam:
Ah cool, I've added that now.

the-mikedavis (approved):

PR #6370, opened 2023-03-19
Add initial Smithy support

zetashift:
Adds syntax highlighting support for https://smithy.io/2.0/index.html a meta-modeling language to describe APIs.

LSP: https://github.com/disneystreaming/smithy-language-server
Treesitter grammar: https://github.com/indoorvivants/tree-sitter-smithy

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2352,7 +2352,7 @@ roots = ["smithy-build.json"]
 auto-format = false
 comment-token = "//"
 indent = { tab-width = 4, unit = "  "  }
-language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:0.0.21", "--", "0"] }
+language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:latest.release", "--", "0"] }
 
 [[grammar]]
 name = "smithy"
--- a/languages.toml
+++ b/languages.toml
@@ -2347,11 +2347,9 @@ name = "smithy"
 scope = "source.smithy"
 injection-regex = "smithy"
 file-types = ["smithy"]
-shebangs = []
 roots = ["smithy-build.json"]
-auto-format = false
 comment-token = "//"
-indent = { tab-width = 4, unit = "  "  }
+indent = { tab-width = 4, unit = "    " }
 language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:latest.release", "--", "0"] }
 
 [[grammar]]
--- a/languages.toml
+++ b/languages.toml
@@ -2341,3 +2341,19 @@ indent = { tab-width = 2, unit = "  " }
 [[grammar]]
 name = "capnp"
 source = { git = "https://github.com/amaanq/tree-sitter-capnp", rev = "fc6e2addf103861b9b3dffb82c543eb6b71061aa" }
+
+[[language]]
+name = "smithy"
+scope = "source.smithy"
+injection-regex = "smithy"
+file-types = ["smithy"]
+shebangs = []
+roots = ["smithy-build.json"]
+auto-format = false
+comment-token = "//"
+indent = { tab-width = 4, unit = "  "  }
+language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:0.0.21", "--", "0"] }
+
+[[grammar]]
+name = "smithy"
+source = { git = "https://github.com/indoorvivants/tree-sitter-smithy", rev = "cf8c7eb9faf7c7049839585eac19c94af231e6a0" }
--- a/runtime/queries/smithy/highlights.scm
+++ b/runtime/queries/smithy/highlights.scm
@@ -1,6 +1,6 @@
 ; Queries are taken from: https://github.com/indoorvivants/tree-sitter-smithy/blob/main/queries/highlights.scm
 ; Preproc
-(control_key) @preproc
+(control_key) @keyword.directive
 
 ; Namespace
 (namespace) @namespace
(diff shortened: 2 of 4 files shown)


runtime/queries/smithy/highlights.scm
```diff
@@ -0,0 +1,102 @@
...
+; Types
+(identifier) @type
+(structure_resource
+  (shape_id) @type)
+
+; Attributes
+(mixins
+  (shape_id) @attribute)
+(trait_statement
+  (shape_id (#set! "priority" 105)) @attribute)
```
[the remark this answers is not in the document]
the-mikedavis: No I think "priority" is neovim-specific, we don't implement it

runtime/queries/smithy/highlights.scm
the-mikedavis, line 1: Some of the scopes here need to be updated to use the helix ones rather than the neovim ones, for examples `preproc` should be `keyword.directive`, `field` should be `variable.other.member`, `string.escape` should be `constant.character.escape`

https://docs.helix-editor.com/master/themes.html#syntax-highlighting

zetashift: Ah darn, most of the scopes matches when I glossed over it, so I thought it was alright :P. Thank you for the examples!

languages.toml
```diff
@@ -2341,3 +2341,19 @@ indent = { tab-width = 2, unit = "  " }
...
+name = "smithy"
+scope = "source.smithy"
+injection-regex = "smithy"
+file-types = ["smithy"]
+shebangs = []
+roots = ["smithy-build.json"]
+auto-format = false
+comment-token = "//"
+indent = { tab-width = 4, unit = "  "  }
+language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:0.0.21", "--", "0"] }
```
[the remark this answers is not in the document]
zetashift: It's a tool to grab Scala/JVM artifacts: https://get-coursier.io/

It's how the vscode smithy extensions installs the LSP as well: https://github.com/disneystreaming/vscode-smithy/blob/main/src/coursier/download-coursier.ts

There is another LS by AWS, but it's installation is a bit more involved: https://github.com/mason-org/mason-registry/pull/148/files#diff-842575865783293dbaedbb7d5de0a10794860d0a04d6e26acd437927d206baae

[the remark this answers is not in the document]
zetashift: Sorry I just realized I didn't really answer your second question, if you have coursier installed you can get the LS by just running:
`cs launch com.disneystreaming.smithy:smithy-language-server:latest -- 0`

languages.toml
```diff
@@ -2341,3 +2341,19 @@ indent = { tab-width = 2, unit = "  " }
...
 source = { git = "https://github.com/amaanq/tree-sitter-capnp", rev = "fc6e2addf103861b9b3dffb82c543eb6b71061aa" }
+
+[[language]]
+name = "smithy"
+scope = "source.smithy"
+injection-regex = "smithy"
+file-types = ["smithy"]
+shebangs = []
+roots = ["smithy-build.json"]
+auto-format = false
```
the-mikedavis, line 2352: These are the defaults for `shebangs` and `auto-format` so they can be removed

languages.toml
```diff
@@ -2341,3 +2341,19 @@ indent = { tab-width = 2, unit = "  " }
...
+comment-token = "//"
+indent = { tab-width = 4, unit = "  "  }
```
the-mikedavis, line 2354:
```suggestion
indent = { tab-width = 4, unit = "    "  }
```

From the docs it looks like this should be 4 columns, right?

runtime/queries/smithy/highlights.scm
```diff
@@ -0,0 +1,102 @@
...
+; Preproc
+(control_key) @keyword.directive
+
+; Namespace
+(namespace) @namespace
+
+; Includes
+[
+  "use"
+] @include
```
the-mikedavis, line 11:
```suggestion
] @keyword.control.import
```

the-mikedavis (approved):

the-mikedavis:
Thanks! Could you add a note about coursier to the language server wiki? https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

zetashift:
> Thanks! Could you add a note about coursier to the language server wiki? https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

Thank you! Added a small section: https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#smithy-language-server

mtdowling:
This is awesome @zetashift. Any reason you used the Disney version instead of the version maintained by AWS? The AWS version has additional fixes like not crashing if there isn't a smithy-build.json file. It can also be installed through Coursier ([via their contrib channel](https://github.com/coursier/apps/blob/main/apps-contrib/resources/smithy-language-server.json)) using:

```
cs launch --contrib smithy-language-server -- 0
```

zetashift:
> This is awesome @zetashift. Any reason you used the Disney version instead of the version maintained by AWS?

Hi @mtdowling! At first I wanted to use the version maintained by AWS, but when I was looking at what neovim's mason plugin does: https://github.com/mason-org/mason-registry/pull/148/files#diff-842575865783293dbaedbb7d5de0a10794860d0a04d6e26acd437927d206baae I kinda got scared and just went with Disney's, because that one has a simple coursier install command, and I am more familiar with Scala.

But able to use coursier for the AWS one changes things quite a lot for me... :P. 

How about I make a new PR that uses the LS of AWS and coursier?

mtdowling:
Yeah a PR to use the AWS version would be awesome!

> or if you know of better "run the LS" programs/commands I can look into that too, I just made this PR on whatever got me up and running the quickest

For now, using `cs launch --contrib smithy-language-server -- 0` seems like the right approach. We might figure out a better way in the future to do something more self contained, but don't have that yet.

zetashift:
> Yeah a PR to use the AWS version would be awesome!

Thank you for the coursier tip, the PR is up: https://github.com/helix-editor/helix/pull/6370#issuecomment-1494844702

PR #6375, opened 2023-03-20
fleet-dark improvements

krfl:
A few changes that does not comply with the original theme, but adjusts the theme to better fit the editor of choice.

* Statusline is now visible. This was an issue when using horizontal splits
* ui.virtual was adjusted to a much lighter color. This is an issue only because we are using the "reversed" modifier for the cursor, and thus this change is required to make the cursor visible again when moving through whitespace.  As a result, whitespace is now much more "in your face" when enabled. I consider this an okay price to pay. Feel free to disagree

Screenshot 2023-03-20 at 10 01 25
Screenshot 2023-03-20 at 10 02 21

### DIFF
--- a/runtime/themes/fleet_dark.toml
+++ b/runtime/themes/fleet_dark.toml
@@ -66,8 +66,9 @@
 
 # ui specific
 "ui.background" = { bg = "Gray 10" } # .separator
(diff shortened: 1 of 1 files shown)


krfl:
@matoous @kirawi any thoughts?

matoous:
@krfl looks good to me 👍

kirawi (approved):
Looks good.

the-mikedavis (approved):

PR #6378, opened 2023-03-20
Show diagnostic codes for LSP diagnostics

FrancescElies:
Similar to #6095

**Problem** currently during diagnostic rendering only the description is shown but the code is also useful when one wants to ignore that error report.

**After this pr** errors will be rendered as follows.
image

Now you know what code to use to ignore the error
image

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -31,7 +31,7 @@ use helix_view::{
 };
 use std::{mem::take, num::NonZeroUsize, path::PathBuf, rc::Rc, sync::Arc};
 
-use tui::buffer::Buffer as Surface;
+use tui::{buffer::Buffer as Surface, text::Span};
 
 use super::statusline;
 use super::{document::LineDecoration, lsp::SignatureHelp};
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -684,6 +685,14 @@ impl EditorView {
                 });
             let text = Text::styled(&diagnostic.message, style);
             lines.extend(text.lines);
+            let code = diagnostic.code.as_ref().map(|x| match x {
+                NumberOrString::Number(n) => format!("({n})"),
+                NumberOrString::String(s) => format!("({s})"),
+            });
+            if let Some(code) = code {
+                let text = Text::styled(code, style);
+                lines.extend(text.lines);
```
the-mikedavis, line 694: Using `Text::styled` is a little bit overkill here. It's used in the above block because the diagnostic message may have newlines in it but we know for sure that this span will not. We can use `Span` instead which won't try to parse any lines or look at the text:

```suggestion
                let span = Span::styled(code, style);
                lines.push(span.into());
```

the-mikedavis (approved):

PR #6391, opened 2023-03-21
book: changed displayed key binding for diagnostics picker

janhrastnik:
Currently the displayed key binding for diagnostics picker is 'd' and 'D', the right one should be 'g' and 'G'


the-mikedavis:
Actually the diagnostics pickers and the debug mode were swapped in https://github.com/helix-editor/helix/pull/4229

janhrastnik:
My bad, I did not realise I'm still on 22.08.1 :p

Issue #6346, opened 2023-03-17
Regression: truncation in file picker no longer working in master


keirlawson:
### Summary

In the case where file paths are too long for the size of the file picker they should be truncated, as they are in 22.12, however on master they are not.

### Reproduction Steps

I tried this:

1. `hx`
2. <space> <f>

I expected this to happen:

long file paths truncated

Instead, this happened:

long file paths were not truncated

### Platform

Ubuntu 22.10

### Terminal Emulator

GNOME terminal 3.46.2

### Helix Version

22.12-436-ga07655d5

PR #6410, opened 2023-03-22
Allow truncation of table row text

mWalrus:
This fixes the regression mentioned in https://github.com/helix-editor/helix/issues/6346 where paths in the file picker were no longer being truncated.
We now allow all pickers who have `self.truncate_start` set to `true` to render truncated rows.

Not sure if I'm going about this the right way, any feedback is appreciated :)

### DIFF
--- a/helix-tui/src/buffer.rs
+++ b/helix-tui/src/buffer.rs
@@ -434,6 +434,20 @@ impl Buffer {
     }
 
     pub fn set_spans_truncated(&mut self, x: u16, y: u16, spans: &Spans, width: u16) -> (u16, u16) {
+        // if the spans contains only 1 span and the
+        // content length of that span is 1, we can
+        // assume that we dont need to truncate the
+        // string.
+        if spans.0.len() == 1 {
+            match spans.0.first() {
+                Some(s) if s.content.len() == 1 => {
+                    return self.set_stringn(x, y, s.content.as_ref(), width as usize, s.style)
+                }
+                // if we pass the above checks, we continue as normal
+                _ => {}
+            }
+        }
+
         let mut remaining_width = width;
         let mut alt_x = x;
         let (text, styles) =
--- a/helix-tui/src/buffer.rs
+++ b/helix-tui/src/buffer.rs
@@ -434,20 +434,6 @@ impl Buffer {
     }
 
     pub fn set_spans_truncated(&mut self, x: u16, y: u16, spans: &Spans, width: u16) -> (u16, u16) {
-        // if the spans contains only 1 span and the
-        // content length of that span is 1, we can
-        // assume that we dont need to truncate the
-        // string.
-        if spans.0.len() == 1 {
-            match spans.0.first() {
-                Some(s) if s.content.len() == 1 => {
-                    return self.set_stringn(x, y, s.content.as_ref(), width as usize, s.style)
-                }
-                // if we pass the above checks, we continue as normal
-                _ => {}
-            }
-        }
-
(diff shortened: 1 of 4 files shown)


pascalkuthe:
I had the thaught of soft wrapping text here instead of truncation. At least for diagnostics, it would be really nice to be able the entire diagnostic instead of truncating. I am not sure if soft wrapping is always preferable to truncation (I personally am not a fan of the way paths were truncated because there was no way to horizontall scroll a picker item so it was sometimes very hard to tell certain paths apart). If this needs a fix anyway this might be a good opportunity.

Although it's probably a bit more complex so maybe we try to get this PR in as a regression fix for the next release and look into softwrapping later.

helix-tui/src/buffer.rs
```diff
@@ -433,6 +433,19 @@ impl Buffer {
         (x_offset as u16, y)
     }
 
+    pub fn set_spans_truncated(&mut self, x: u16, y: u16, spans: &Spans, width: u16) -> (u16, u16) {
+        let mut text = String::new();
+        let mut styles = vec![];
+
+        for span in &spans.0 {
+            text.push_str(span.content.as_ref());
+            span.styled_graphemes(span.style)
```
pascalkuthe, line 442: Performing grapheme segmentation here is both slow and incorrect. The `idx` used below is not a grapheme index but a byte index. Instead you should create an iterator that is transversed whenever the current style range is exhausted.

The current implementation can easily lead to crashes

helix-tui/src/widgets/table.rs
```diff
@@ -457,27 +464,32 @@ impl<'a> Table<'a> {
...
 
-fn render_cell(buf: &mut Buffer, cell: &Cell, area: Rect) {
+fn render_cell(buf: &mut Buffer, cell: &Cell, area: Rect, truncate: bool) {
     buf.set_style(area, cell.style);
     for (i, spans) in cell.content.lines.iter().enumerate() {
         if i as u16 >= area.height {
             break;
         }
-        buf.set_spans(area.x, area.y + i as u16, spans, area.width);
+        if cell.content.width() > 1 && truncate {
```
pascalkuthe, line 481: I dont think the emptyness check is required here. Why did you add it?

In any case it should be using byte Len (or `is_empty` really) instead of grapheme width since it's O(1) and any non-empty string usually has a width (the cases where it doesn't can be negelagted IMO)

pascalkuthe (changes requested):
I actually like this a lot more than the old truncation (uses ... Instead of shortening ). It keeps more screen realeaste available the shortened path components were never really that useful to me. Other tools like fzf and skim di the same thing. The implementation needs some fixes tough

helix-tui/src/buffer.rs
mWalrus: I figured this was a slow solution.
Could you elaborate a bit about what you mean and how I could go about correcting this? I'm pretty new to this kind of thing.

kirawi:
Superseded by https://github.com/helix-editor/helix/pull/6488

pascalkuthe:
Thanks for working on this. Usually I would have given you more feedback here until this was ready but we wanted to get a fix in before the release (and there is not a lit.of.time left in March :D) so I just cherry picked your commit and implemented `set_spans_truncated` in a second commjt myself. Using the truncation here is actually really good ui so it was a great idea and just nee

mWalrus:
No worries! I'll look through the implementation for learning purposes :)
Thanks for looking at this!

PR #6412, opened 2023-03-23
Puppet Support

### LINKED ISSUE #6073: Puppet and ERB language support
Puppet Tree-sitter [exists](https://github.com/neovim-puppet/tree-sitter-puppet)

I'm not exactly sure how template languages work with Helix/Tree-sitter: an ERB (or Jinja or any other) template is essentially two distinct formats in single file.

I will sooner or later probably try to create a PR for Puppet but I'm not even sure where to start with ERB...

gautamprikshit1:
Fixes #6073 
@jficz will you please test this?

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1,5 +1,4 @@
-# Language support configuration.
-# See the languages documentation: https://docs.helix-editor.com/master/languages.html
+# Language support configuration. See the languages documentation: https://docs.helix-editor.com/master/languages.html
 
 [[language]]
 name = "rust"
@@ -2369,3 +2368,16 @@ language-server = { command = "cs", args = ["launch", "com.disneystreaming.smith
 [[grammar]]
 name = "smithy"
 source = { git = "https://github.com/indoorvivants/tree-sitter-smithy", rev = "cf8c7eb9faf7c7049839585eac19c94af231e6a0" }
+
+[[language]]
+name = "puppet"
+scope = "source.puppet"
+injection-regex = "puppet"
+file-types = ["puppet"]
+roots = []
+indent = { tab_width = 2, unit = "  " }
+comment-token = "#"
+
+[[grammar]]
+name = "puppet"
+source = { git = "https://github.com/neovim-puppet/tree-sitter-puppet" }
--- a/languages.toml
+++ b/languages.toml
@@ -2380,4 +2380,4 @@ comment-token = "#"
 
 [[grammar]]
 name = "puppet"
-source = { git = "https://github.com/neovim-puppet/tree-sitter-puppet" }
+source = { git = "https://github.com/neovim-puppet/tree-sitter-puppet", rev = "bdedf4d5b13b4020745a678364e2f441cd3abf2a" }
--- a/runtime/queries/puppet/highlights.scm
+++ b/runtime/queries/puppet/highlights.scm
@@ -1,0 +1,68 @@
+(function) @function
+(variable) @variable
+[
+  (identifier)
+] @identifier
+
+[
+ "class"
+ "define"
+ "plan"
+ "node"
+ "type"
+] @keyword
+
+[
+  "contain"
+  "include"
+  "inherits"
+  "require"
+] @include
+
+[
+ "case"
+ "else"
+ "elsif"
+ "if"
+ "unless"
+] @conditional
+
+(string) @string
+(regex) @regex
+(integer) @number
+(float) @float
+(regex) @string.regex
+(comment) @comment
+[(true) (false)] @boolean
+
+(unprotected_string) @text
+
+[
+ (chaining_arrow)
+ (operator)
+] @operator
+
+
+(interpolation
+  "${" @punctuation.special
+  "}" @punctuation.special) @none
+
+[
+ "("
+ ")"
+ "["
+ "]"
+ "{"
+ "}"
+ ]@punctuation.bracket
+
+[
+ ","
+ ":"
+ "::"
+] @punctuation.delimiter
+
+[(type) (type_identifier)] @type
+
+(escape_sequence) @escape
+


runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+[
+ (chaining_arrow)
+ (operator)
+] @operator
+
+
+(interpolation
+  "${" @punctuation.special
+  "}" @punctuation.special) @none
+
```
archseer, line 49: Please follow https://docs.helix-editor.com/themes.html#syntax-highlighting

languages.toml
```diff
@@ -2369,3 +2368,16 @@ language-server = { command = "cs", args = ["launch", "com.disneystreaming.smith
 [[grammar]]
 name = "smithy"
 source = { git = "https://github.com/indoorvivants/tree-sitter-smithy", rev = "cf8c7eb9faf7c7049839585eac19c94af231e6a0" }
+
+[[language]]
+name = "puppet"
+scope = "source.puppet"
+injection-regex = "puppet"
+file-types = ["puppet"]
```
erasin, line 2376: file-types = ["pp"]

languages.toml
```diff
@@ -2369,3 +2368,16 @@ language-server = { command = "cs", args = ["launch", "com.disneystreaming.smith
...
+roots = []
+indent = { tab_width = 2, unit = "  " }
```
erasin, line 2378: indent = { tab-width = 2, unit = " " }

runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+ "unless"
+] @conditional
+
+(string) @string
+(regex) @regex
+(integer) @number
+(float) @float
+(regex) @string.regex
+(comment) @comment
+[(true) (false)] @boolean
```
erasin, line 36: (string) @string
(regex) @string.regex
(comment) @comment
[(true) (false)] @constant.boolean

runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+  "require"
+] @include
+
+[
+ "case"
+ "else"
+ "elsif"
+ "if"
+ "unless"
+] @conditional
```
erasin, line 28: ] @keyword.control.conditional

runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+  (identifier)
+] @identifier
+
+[
+ "class"
+ "define"
+ "plan"
+ "node"
+ "type"
+] @keyword
```
erasin, line 13: ] @keyword.storage.type

runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+
+[
+  "contain"
+  "include"
+  "inherits"
+  "require"
+] @include
```
erasin, line 20: ] @keyword.control.import

runtime/queries/puppet/highlights.scm
```diff
@@ -0,0 +1,68 @@
...
+
+[
+ ","
+ ":"
+ "::"
+] @punctuation.delimiter
+
+[(type) (type_identifier)] @type
+
+(escape_sequence) @escape
```
erasin, line 67: (escape_sequence) @constant.character.escape

languages.toml
```diff
@@ -2369,3 +2368,16 @@ language-server = { command = "cs", args = ["launch", "com.disneystreaming.smith
...
+comment-token = "#"
```
erasin, line 2379: file-types = ["pu"]
indent = { tab-width = 2, unit = " " }

erasin:
I sort out the hightlights, check https://github.com/neovim-puppet/tree-sitter-puppet/tree/main/test/corpus, there are still some highlights that need to be implemented.


hightlights.scm

```scm
(function) @function
(function_declaration 
  (identifier) @function)
(parameter
  (variable) @variable.parameter)
(resource_title) @label
(resource_declaration (identifier) @type.constructor)
(resource_call (identifier) @type.constructor )
(include (identifier) @type.constructor )
(require (identifier) @type.constructor )
(contain (identifier) @type.constructor )
(resource_defaults (resource_type) @type.constructor)
(argument_list (argument (argument_name) @variable.other.member))

; reference ---
; (reference (identifier))
; (reference_identifier)


(variable) @variable
[
  (identifier)
] @identifier

[
  "function"
] @keyword.function

[
 "class"
 "define"
 "plan"
 "node"
 "type"
] @keyword.storage.type

[
  "contain"
  "include"
  "inherits"
  "require"
] @keyword.control.import

[
 "case"
 "else"
 "elsif"
 "if"
 "unless"

(heredoc) @string
(string) @string
(regex) @string.regexp
(comment) @comment
[(true) (false)] @constant.boolean

(unprotected_string) @text

[
 (chaining_arrow)
 (operator)
] @operator


(interpolation
  "${" @punctuation.special
  "}" @punctuation.special) @none

[
 "("
 ")"
 "["
 "]"
 "{"
 "}"
 ]@punctuation.bracket

[
 ","
 ":"
 "::"
] @punctuation.delimiter

[(type) (type_identifier)] @type

```

Append `indents.scm`

```scm
[
(function_declaration) 
(resource_declaration)
(resource_defaults)
(if)
]@indent

[
  "case"
  "}"
]@outdent
```

languages.toml
jficz: I would actually argue that it is Pascal (or, specifically, FreePascal's need to be different) which is in conflict with Puppet here. Historically and practically, `.pas` is the recognized Pascal extension. `.pp` is just a vendor-specific variant which doesn't really have a rational reason to exist in the first place and therefore it should be the other way around.

ardrigh:
It would be great to see some Puppet `.pp` support added, as being detected as Pascal files breaks commenting (inserts `//` instead of `#`) and requires more care with formatting.

jficz:
so I tried this config using a custom [patch](https://gist.github.com/jficz/c6a7666b73883636e6441e8cf918f755) against helix release 23.03 which contains the changes to `language.toml` as are present in the PR (with fixed `tab-width` typo), the `highlights.scm` as posted by @erasin and an update to pascal files removing the `.pp` suffix from it but I'm probably doing something wrong because I don't see _any_ highlights at all. Is there a way to debug this and, preferably, to make changes without having to recompile the whole thing?

jficz:
never mind, `hx --grammar fetch, build` and now it works well!

Need to tune the theme because I don't like the color scheme but the highlighter seems to work.

There are a few small issues though:

`$vars`, `properties` and `"${interpolations}"` are not highlighted, a certain strings break string highlights:

puppet

looks like that it is the dash (`-`) character that breaks the string highlight

jficz:
@gautamprikshit1 I couldn't a way to bind my fixes to your PR so I created a new one:
https://github.com/helix-editor/helix/pull/14117

the-mikedavis:
Closing in favor of #14117

PR #6415, opened 2023-03-23
Add language support for rego (OpenPolicyAgent)

jcmuller:
Tree-sitter grammar from https://github.com/FallenAngel97/tree-sitter-rego.
Queries from https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/rego/highlights.scm

### DIFF
--- a/runtime/queries/rego/injections.scm
+++ b/runtime/queries/rego/injections.scm
@@ -1,0 +1,1 @@
+(comment) @comment
--- a/runtime/queries/rego/injections.scm
+++ b/runtime/queries/rego/injections.scm
@@ -1,1 +1,2 @@
-(comment) @comment
+((comment) @injection.content
+ (#set! injection.language "comment"))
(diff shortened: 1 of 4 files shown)


runtime/queries/rego/injections.scm
```diff
@@ -0,0 +1 @@
+(comment) @comment
```
the-mikedavis, line 1:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
```

jcmuller: Done.

the-mikedavis (approved):

the-mikedavis:
Thanks! Could you add a note for regols to the language server wiki? https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

jcmuller:
Done!

PR #6436, opened 2023-03-25
Add blocking UI feedback to `goto_definition`, `goto_references`.

askreet:
Based on some initial work by @pascalkuthe for general blocking UI components.

The main feature added here is user feedback for goto_definition and goto_references. Once initiated, they can be canceled with C-c if the user no longer wishes to wait.

I didn't want to create too much change in one PR, but if accepted this pattern should probably be applied to nearly all LSP commands and some DAP commands as well. I'd be happy to chase that down in a follow-up series of PRs.

Some thoughts on the implementation are below.

## Handling layered editor status.

The way editor status is implemented, it's easy to stomp on another async message. In fact, the handler for LSP incoming progress updates (see application.rs:909) previously cleared the status line regardless of whether the user had `lsp.display-status` enabled. This is likely just a simple bugfix (which enabled my added message not to be cleared prematurely), but it highlights the brittle nature of the current solution.

I'm wondering if longer term something like the LSP status struct that tracks all current background tasks should be leveled up to a general editor concept which can track multiple active activities and perhaps a priority rating of each, so that we can show the most urgent one (your currently requested action is blocked and waiting) vs. lower priority ones (other background LSP status update messages).

Because of all this, if you _do_ have `lsp.display-status` set, the LSP message I set gets trampled and you can just be left with a dim window for a bit while you wait (or cancel).

## Use of editor status vs. new layer.

I didn't pursue this option, but it occurred to me that since we're dimming and freezing the main UI until we can react to the user's request, we have a lot of real estate to play with. We could place a message here for example:

image

## Impact on LSP message handling.

One thing I don't understand well is how this impacts the LSP or the message handling. When the cancel message is sent, we de-allocate all the futures (I believe) waiting for the response. Any downside to that? Any LSP commands we should send to ignore the previous request?

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1,5 +1,5 @@
 use arc_swap::{access::Map, ArcSwap};
-use futures_util::Stream;
+use futures_util::{future::OptionFuture, Stream};
 use helix_core::{
     diagnostic::{DiagnosticTag, NumberOrString},
     path::get_relative_path,
@@ -11,7 +11,7 @@ use helix_view::{
     document::DocumentSavedEventResult,
     editor::{ConfigEvent, EditorEvent},
     graphics::Rect,
-    theme,
+    theme::{self, Modifier, Style},
     tree::Layout,
     Align, Editor,
 };
@@ -23,6 +23,7 @@ use crate::{
     commands::apply_workspace_edit,
     compositor::{Compositor, Event},
     config::Config,
+    ctrl,
     job::Jobs,
     keymap::Keymaps,
     ui::{self, overlay::overlayed},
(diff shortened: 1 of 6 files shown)


pascalkuthe (approved):
Sorry I kind of forgot about this. Thanks for picking up my work!

> One thing I don't understand well is how this impacts the LSP or the message handling. When the cancel message is sent, we de-allocate all the futures (I believe) waiting for the response. Any downside to that? Any LSP commands we should send to ignore the previous request?

That's not an issue. We already set timeouts to lsp commands, we just stop waiting for the servers reply. 

> I didn't pursue this option, but it occurred to me that since we're dimming and freezing the main UI until we can react to the user's request, we have a lot of real estate to play with. We could place a message here for example

The reason I used the status message is that in most cases the wait time should be pretty short and it would be distracting to have a big popup open everytime IMO. We could definitly experiment here but for an initial version reusing the statusline seems like a pretty decent option.

Multiple status messages conflicting is actually an existing problem that I noticed. I am not sure how to best address that. IMO it would be good to address that seperatly

daedroza:
Can this be rebased on latest head?

PR #6438, opened 2023-03-25
tui: Handle keyboard enhancement check failure

the-mikedavis:
If the terminal doesn't send the primary device attributes response to the query, the `terminal::supports_keyboard_enhancement` function from crossterm may timeout and return an Err.

We should interpret this error to mean that the terminal doesn't support the keyboard enhancement protocol rather than an error in claiming the terminal.

This was a bug introduced in the refactor in #6194

### DIFF
--- a/helix-tui/src/backend/crossterm.rs
+++ b/helix-tui/src/backend/crossterm.rs
@@ -78,21 +78,20 @@ where
     }
 
     #[inline]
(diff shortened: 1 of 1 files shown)


pascalkuthe (approved):
LGTM, should probably get this one in before the next release since it's a regression with a simple fix:+1:

Issue #6358, opened 2023-03-18
Git diff gutter doesn't work on Windows


Snoupix:
### Summary

Alright so I'm using Helix 22.12 (built from source) on Windows 11 with the *new* Windows Terminal.

Everything is working well except the Git diff gutter. I tried on many projects and I first thought that the problem was Windows Terminal but on my WSL distro, everything is working well (Rocky Linux 9, Helix version 22.12).

Feel free to ask me for details about my setup but I think I've mentioned everything.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>
2023-03-17T03:07:55.356 helix_loader [DEBUG] Located configuration folders: []
2023-03-17T03:07:55.373 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:55.375 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:55.782 helix_term::application [DEBUG] received editor event: IdleTimer
2023-03-17T03:07:56.830 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:57.221 helix_term::commands::typed [DEBUG] quitting...
2023-03-17T03:07:57.221 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:57.243 helix_term::application [DEBUG] received editor event: IdleTimer
2023-03-17T03:07:57.308 helix_term::commands::typed [DEBUG] quitting...
2023-03-17T03:07:57.350 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:57.351 helix_term::job [DEBUG] waiting on jobs...
2023-03-17T03:07:57.351 helix_term::job [DEBUG] waiting on jobs...
2023-03-17T03:07:59.922 helix_loader [DEBUG] Located configuration folders: []
2023-03-17T03:07:59.938 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:07:59.941 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:08:00.348 helix_term::application [DEBUG] received editor event: IdleTimer
2023-03-17T03:08:06.319 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-03-17T03:08:06.609 helix_term::commands::typed [DEBUG] quitting...
[...]

PR #6441, opened 2023-03-25
log failures in the git integration

pascalkuthe:
Issues with the diff gutter like #6358 are pretty hard to debug right now because there is essentially no logging of failures in the git integration (they all get turned into options). This PR makes the functions of the `DiffProvider` trait return a Result and log any failures.


pickfire (approved):
Looks good to me.

PR #6446, opened 2023-03-26
60 fps progress spinner

pickfire:
Previous spinner kinda look a bit weird given that it needs to wait for progress to spin, makes the editor feels laggy.

[asciicast](https://asciinema.org/a/EKxA31vqAGWls2fZMXoS3km0Z)

Now it spins at 60 fps which makes the editor look fast.

[asciicast](https://asciinema.org/a/K64QU9629b8b56EZpcRRcvFNp)

Not sure if there is a better way to do the interval timer at commit fb572fb5f604023a8f2e19b2909761ff32a197ec, not sure if there are any performance issues either.

The pause at the end to remove the progress message looks weird due to the idle timeout, I was thinking to remove it, but then I figure out if the last message is an error, user should see it, but so far I only see rust-analyzer outputting completed message, so not removing it as I am not sure.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -500,6 +500,16 @@ impl Application {
         }
     }
 
+    pub async fn handle_frame_tick(&mut self) {
+        let editor_view = self
+            .compositor
+            .find::<ui::EditorView>()
+            .expect("expected at least one EditorView");
+        if editor_view.spinners_mut().spinning() {
+            self.render().await;
+        }
+    }
+
     pub fn handle_document_write(&mut self, doc_save_event: DocumentSavedEventResult) {
         let doc_save_event = match doc_save_event {
             Ok(event) => event,
@@ -599,6 +609,9 @@ impl Application {
                 {
                     return true;
                 }
+            }
+            EditorEvent::TickFrame => {
+                self.handle_frame_tick().await;
             }
         }
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -872,7 +872,7 @@ impl Application {
                                 } else {
                                     self.lsp_progress.end_progress(server_id, &token);
                                     if !self.lsp_progress.is_progressing(server_id) {
-                                        editor_view.spinners_mut().get_or_create(server_id).stop();
+                                        editor_view.spinners_mut().stop(server_id);
                                     }
                                     self.editor.clear_status();
 
@@ -915,7 +915,7 @@ impl Application {
                         if let lsp::WorkDoneProgress::End(_) = work {
                             self.lsp_progress.end_progress(server_id, &token);
                             if !self.lsp_progress.is_progressing(server_id) {
-                                editor_view.spinners_mut().get_or_create(server_id).stop();
+                                editor_view.spinners_mut().stop(server_id);
                             }
                         } else {
                             self.lsp_progress.update(server_id, token, work);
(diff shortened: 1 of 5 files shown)


pascalkuthe:
Looks very cool a nice visual improvement 👍 However I am a bit iffy about running the rendering code this often.

Right now it's mostly fine to always redraw everything because usually only the editor is visible and that needs to be redrawn on every command. But with animations that's. a lot of wasted CPU cycles especially for larger files where TS can get slow or even just transversing the document at all for rendering could be slow (altough that will be improved with chunked softwrapping). These cass would also make the animation laggy.

This issue actually already exists in a limited capacity. Scrolling in the picker or typing in command mode can lag if a very large file is opened in the background (altough ebersince the TS query performance improvements that's less of an issue in practice). 

I think we can learn from UI libraries here: The compositor should support invalidating components so they can be redraw individually. The easiest implementation right now would simply redraw all componenents that come after the first invalidated  componenent (om top of the last render). In most cases we would redraw everything because most cases where we redraw right now invalidate the editor. For compositor events we can simply invalidate the components that consume them.

In this specific case that wouldn't quite work because the status line is not it's own componemnnt. That is kind of unfortunate but we can probably do a workaround inside the editor rendering code

CptPotato:
I like the consistent spin interval. Though I think having it refresh at a high rate could cause issues when using helix via ssh.

pickfire:
But I am surprised that even though I limited the render rate, having just sending the emitting a regular event every 16ms is causing a 2% cpu usage even though it only renders 8 frame per second. Maybe somewhere in tokio select it is using the cpu.

pickfire:
I am thinking of putting 8 fps just enough to render the progress bar and with the event being sent only when there are any progress bar. Okay, I will add region invalidation too.

sbromberger:
Please test this over a mosh connection as well - my experience with mosh is that it does some weird optimizations with character rendering over high-latency connections. Thanks,

dpc:
> This is a 👎🏻 for me: it'll waste battery life and the stuttering is actually useful to me too as a sort of a progress bar: it means the LSP is stuck on a certain step.

FWIW. If the spinner is running your LSP is probably busy at work so CPUs are not going to be idling.

But I think 60 fps in a CLI is a complete overkill. 8fps seems more reasonable, but I'd be fine even with 1fps, just as long as I can tell that LSP is still spinning instead of scratching my head.

Zoybean:
Is this only an aesthetic improvement? In my mind, the current spinner presents semantic information that is lost if it always updates:
- currently:
  - the spinner's presence indicates that I'm waiting on the LSP (or some other processing idk)
  - the movement indicates progress
  - the spinner stopping indicates either the UI is frozen or the LSP is stuck on something.
- proposed:
  - the spinner's presence still indicates that I'm waiting
  - movement indicates that the UI is responsive
  - the spinner stopping indicates that the UI is frozen. 

But I don't really need the spinner to tell me whether the UI is frozen or not, since there are already other ways to tell that (i.e. the ui doesn't respond to input).

I guess this is all to say, could this be a configurable preference, if it is implemented?

dpc:
> * the spinner stopping indicates either the UI is frozen or the LSP is stuck on something.

I just looked at the spinners behavior on our 100k LoC Rust project, and I think the main problem is that there is not distinction between spinner waiting for LSP and just being idle. When I press a key and `⣻` starts changing  for a brief moment and turns into `⣾`, I don't know if it's because it's waiting for LSP now or is it just.

IMO, the spinner should either disappear or turn into something else when there are no more outstanding LSP request. Possibly a lot of what I considered "stuck" was actually already idle.

PR #6417, opened 2023-03-24
Inline Diagnostics

pascalkuthe:
Closes https://github.com/helix-editor/helix/issues/3272
Closes #3078
closes https://github.com/helix-editor/helix/issues/1462
Closes https://github.com/helix-editor/helix/issues/1953
Supersedes #6059

This PR implements inline diagnostics just like #6059. However, it also contains a streamlined virtual text API and significantly improves on the implementation in #6059.

The various improvements and bugfixes made to the virtual text API are currently part of this PR until it's been tested more. I will spin those out into a separate PR later. These should address the various bugs and panics discovered by poeple in #6059. Only the last commit contains an implementation of inline diagnostics.

Due to the changes to the virtual text API I had to essentially rebuild #6059 as most changes in that PR became unnecessary by the improvements in the virtual text API.

I also rebuild the ASCII rendering to support soft wrapping. Diagnostics which exceed the width of the screen are now soft wrapped automatically. Additionally, there is a minimum width setting for diagnostics. Any diagnostics so far to the right that there is not this minimum width available are displayed using a backwards arching line. See screenshot below:

image

Additionally, this PR implements a severity-based filter for diagnostics. This allows users to choose one of the following:
* a minimum severity for inline diagnostics: `"warning"`
* a list of severities to allow (useful if a certain LSP spams one particular severity). `["warning", "errror"]` Using an empty list disables inline diagnostics

The diagnostics to display can be separately configured for the line that currently contains the cursor. By disabling diagnostics for all other lines users can therefore only display diagnostics on the current line too. By default, only warnings/errors are displayed, on the cursor line all diagnostics are displayed. I am open to changing the default here, but this did feel quite nice to use in my testing.

Additionally, some details of the soft wrapping can be configured. 
Currently, a draft as this needs more documentation, a separate PR for most of the virtual text changes, and more testing. That said it should work quite well already, and I invite people to try this out.

PR #6447, opened 2023-03-26
Track long lived diagnostics inside helix

### LINKED ISSUE #701: Typing text offsets diagnostics that are after text typed

pascalkuthe:
Fixes #701

This PR adds a configuration option (`persistant-diagnostic-sources`) that allows helix to ignore some diagnostics in `publishDiagnostic`. This option takes affect if all diagnostics from a single `lsp_type::Diagnostic::source` are identical to the last list of diagnostics received `publishDiagnostic` for this source. In that case (If the `source` is part of the `persistant-diagnostic-sources` list) all these diagnostics are ignored and instead, the "old" diagnostics for that source are maintained.

Helix already automatically updates the position of diagnostics based on edits so by keeping the existing diagnostics around we can use this mapping mechanism of helix if the server doesn't map long-lived diagnostics through changes itself.

`rust-analyzer` is a good example here which has many diagnostics which update instantly, but diagnostics generated by the fly-check (rustc/clippy) can only be updated when a file is saved. That means that rust-analyzer will send a publishDiagnostic notification to update it's native diagnostics. It's very hard (and inefficient) to map diagnostics trough changes server side and therefore RA has no choice but to also resend the fly check diagnostics (otherwise they are removed) which undoes the work helix does to map them.

The real root cause of this problem is that the LSP standard currently doesn't allow the server to only update some diagnostics for a file (for example by adding a `source` attribute to the `publishDiagnostic` notification itself). Therefore, I came up with this client-side workaround. This works very well in practice but is not 100% correct because in theory the diagnostics could have actually moved and incidentally all end up at the same position as previously. However, the workaround here compares the entire diagnostic (including message)so the chances of that occurring are extremely slim. Perhaps an LSP extension can be proposed for this use case in the future to handle this properly.

As some of these diagnostics now stick around longer some limitations in helixes own diagnostic mapping became apparent. In b3145f94bd9b44e7776f3b421c225dc481b4eb03 I therefore improved the handling of these cases (detailed description in commit message). The logic implemented here can also be reused for better inlay-hint tracking in the future (in fact VScodes inlay hint tracking partially inspired the implementation).

### DIFF
--- a/book/src/languages.md
+++ b/book/src/languages.md
@@ -68,6 +68,7 @@ These configuration keys are available:
 | `formatter`           | The formatter for the language, it will take precedence over the lsp when defined. The formatter must be able to take the original file as input from stdin and write the formatted file to stdout |
 | `text-width`          |  Maximum line length. Used for the `:reflow` command and soft-wrapping if `soft-wrap.wrap-at-text-width` is set, defaults to `editor.text-width`   |
 | `workspace-lsp-roots`     | Directories relative to the workspace root that are treated as LSP roots. Should only be set in `.helix/config.toml`. Overwrites the setting of the same name in `config.toml` if set. |
+| `persistent-diagnostic-sources` | An array of LSP diagnostic sources assumed unchanged when the language server resends the same set of diagnostics. Helix can track the position for these diagnostics internally instead. Useful for diagnostics that are recomputed on save.
 
 ### File-type detection and the `file-types` key
 
--- a/helix-core/src/diagnostic.rs
+++ b/helix-core/src/diagnostic.rs
@@ -39,6 +39,10 @@ pub enum DiagnosticTag {
 #[derive(Debug, Clone)]
 pub struct Diagnostic {
     pub range: Range,
+    // whether this diagnostic ends at the end of(or inside) a word
+    pub ends_at_word: bool,
+    pub starts_at_word: bool,
+    pub zero_width: bool,
     pub line: usize,
     pub message: String,
     pub severity: Option<Severity>,
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -154,6 +154,8 @@ pub struct LanguageConfiguration {
     /// Hardcoded LSP root directories relative to the workspace root, like `examples` or `tools/fuzz`.
     /// Falling back to the current working directory if none are configured.
     pub workspace_lsp_roots: Option<Vec<PathBuf>>,
+    #[serde(default)]
+    pub persistent_diagnostic_sources: Vec<String>,
 }
 
 #[derive(Debug, PartialEq, Eq, Hash)]
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -928,7 +928,9 @@ impl Application {
 
                         // Sort diagnostics first by severity and then by line numbers.
                         // Note: The `lsp::DiagnosticSeverity` enum is already defined in decreasing order
-                        diagnostics.sort_unstable_by_key(|(d, _)| (d.severity, d.range.start));
+                        diagnostics.sort_unstable_by_key(|(d, server_id)| {
+                            (d.severity, d.range.start, *server_id)
+                        });
                     }
                     Notification::ShowMessage(params) => {
                         log::warn!("unhandled window/showMessage: {:?}", params);
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1,6 +1,7 @@
 use arc_swap::{access::Map, ArcSwap};
 use futures_util::Stream;
 use helix_core::{
+    chars::char_is_word,
     diagnostic::{DiagnosticTag, NumberOrString},
     path::get_relative_path,
     pos_at_coords, syntax, Selection,
@@ -832,7 +833,6 @@ impl Application {
                                         log::warn!("lsp position out of bounds - {:?}", diagnostic);
                                         return None;
                                     };
-
                                     let severity =
                                         diagnostic.severity.map(|severity| match severity {
                                             DiagnosticSeverity::ERROR => Error,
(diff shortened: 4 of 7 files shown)


languages.toml
```diff
@@ -11,6 +11,7 @@ auto-format = true
 comment-token = "//"
 language-server = { command = "rust-analyzer" }
 indent = { tab-width = 4, unit = "    " }
+persistant-diagnostic-sources = ["rustc", "clippy"]
```
poliorcetics, line 14:
```suggestion
persistent-diagnostic-sources = ["rustc", "clippy"]
```

pascalkuthe:
rebased after mege of #2507. I added a small commit to fix some small inconsistencies around diagnostic sorting (and to make sure its always deterministic) since I was touching that code anyways

helix-term/src/application.rs
```diff
@@ -747,31 +748,63 @@ impl Application {
...
+                                                *d_server == server_id
+                                                    && d.source.as_ref() == Some(source)
+                                            })
+                                            .map(|(d, _)| d);
+                                        if new_diagnostics.eq(old_diagnostics) {
+                                            unchaged_diag_sources_.push(source.clone())
+                                        }
+                                    }
+                                }
+                            }
```
dead10ck, line 797: I wonder if all this logic could just move into `replace_diagnostics`. As it is, the late function has been changed so that it assumes repeat diagnostics have already been removed, and could potentially allow repeats to be added if this invariant is violated.

dead10ck:
Could you explain why we would want this to be configurable, rather than just doing it for all diagnostics? Especially since we're not adding any defaults.

pascalkuthe:
> Could you explain why we would want this to be configurable, rather than just doing it for all diagnostics? Especially since we're not adding any defaults.

The reason we don't do this for all diagnostics too is that we consider diagnostics unchanged if the server sends the same diagnostics (made line/col number, message,..) again. This is not really LSP compliant and can lead to situations where diagnostics are not updated when they should (text was edited and diagnostic is now somewhere else that only incidentilally corresponds to the same positions). This is a tradeoff and only with it for diagnostics that actually benefit from this. Namely diagnostics that case only updated on safe (or some other rare event) like rust-analyer `checkOnSafe` diagnostics.

I also added a default for rust-analyzer so this setting is used by default.

helix-term/src/application.rs
pascalkuthe: that's not really possible because we need access to `Editor::diagnostics` (not the same as `Document::diagnostics`)  which would be pretty awkward. I also actually prefer to keep the LSP specific logic like thous out of the document if possible.

pascalkuthe: that said this call site doesn't actually handle this correctly. i think I messed that up in the last rebase by accident. I need to rebase this again anyway so I will take a look at that then

pascalkuthe:
I improved the word tracking mapping even further the implementation basically matches vscode in behavior and could be used to fix most of the weirdness surrounding inlay hints  (except the low idle timeout)

helix-core/src/diagnostic.rs
```diff
@@ -39,6 +39,10 @@ pub enum DiagnosticTag {
 #[derive(Debug, Clone)]
 pub struct Diagnostic {
     pub range: Range,
+    // whether this diagnostic ends at the end of(or inside) a word
+    pub ends_at_word: bool,
+    pub starts_at_word: bool,
+    pub zero_width: bool,
```
the-mikedavis, line 45: Originally I thought this could be trivially computed but if I'm understanding correctly, a diagnostic may come in as zero-width and then changes may cause `range.start` and `range.end` to shift, so `range.start == range.end` may not detect this case after changes, right?

pascalkuthe: Exactly these are the properties of the diagnostic when sent from the server. The idea is that non zero width diagnostics are removed if they are empty after mapping. But at that point we don't now if they were *previously* empty

helix-core/src/syntax.rs
```diff
@@ -154,6 +154,8 @@ pub struct LanguageConfiguration {
     /// Hardcoded LSP root directories relative to the workspace root, like `examples` or `tools/fuzz`.
     /// Falling back to the current working directory if none are configured.
     pub workspace_lsp_roots: Option<Vec<PathBuf>>,
+    #[serde(default)]
+    pub persistent_diagnostic_sources: Vec<String>,
```
the-mikedavis, line 158: Could you add a blurb about this to book/src/languages.md?

the-mikedavis (commented):
I think this looks good. I've been using it for a while and I like the assoc changes, especially when stacked with #6417. Just a small comment about the docs

helix-core/src/syntax.rs
pascalkuthe: sure sorry forgot about that :smile: I added an entry to the docs

book/src/languages.md
```diff
@@ -68,6 +68,7 @@ These configuration keys are available:
 | `formatter`           | The formatter for the language, it will take precedence over the lsp when defined. The formatter must be able to take the original file as input from stdin and write the formatted file to stdout |
 | `text-width`          |  Maximum line length. Used for the `:reflow` command and soft-wrapping if `soft-wrap.wrap-at-text-width` is set, defaults to `editor.text-width`   |
 | `workspace-lsp-roots`     | Directories relative to the workspace root that are treated as LSP roots. Should only be set in `.helix/config.toml`. Overwrites the setting of the same name in `config.toml` if set. |
+| `persistent_diagnostic_sources` | An array of LSP diagnostic sources assumed unchanged when the language server resends the same set of diagnostics. Helix can track the position for these diagnostics internally instead. Useful for diagnostics that are recomputed on save.
```
the-mikedavis, line 71:
```suggestion
| `persistent-diagnostic-sources` | An array of LSP diagnostic sources assumed unchanged when the language server resends the same set of diagnostics. Helix can track the position for these diagnostics internally instead. Useful for diagnostics that are recomputed on save.
```

`#[serde(rename_all = "kebab-case")]` :P

pascalkuthe: ah damm copy paste got me :D

David-Else:
@pascalkuthe This PR has caused immediate problems. Warnings are not shown on screen unless I first save a file. They are shown in the picker immediately. They were shown on the screen immediately before this change.

If you want an example:

1. clone and switch to this branch https://github.com/David-Else/rusty-zombie/tree/crossterm
2. open main, open workspace diagnostics and see the 2 warnings.
3. go to the warnings and see nothing is displayed on the screen
4. save the file and they appear, you will then see:

Screenshot from 2024-01-06 10-41-07

pascalkuthe:
this is not caused by this PR and not a new problem either, see https://github.com/helix-editor/helix/pull/8873

David-Else:
> this is not caused by this PR and not a new problem either, see #8873

OK, but it reliably happens every time with this PR, and I have never seen the problem before. Hope to see that fix merged asap :)

PR #6453, opened 2023-03-27
Add standalone language support for `just`

VuiMuich:
While there was some basic support for `just` language derived from `make` this adds it as its own language with its dediceted treesitter.
All the queries are just copied from the [tree-sitter-just](https://github.com/IndianBoy42/tree-sitter-just) project.

Let me know if this is missing anything.


languages.toml
```diff
@@ -2421,3 +2421,16 @@ language-server = { command = "nimlangserver" }
...
 name = "nim"
 source = { git = "https://github.com/aMOPel/tree-sitter-nim", rev = "240239b232550e431d67de250d1b5856209e7f06" }
+
+[[grammar]]
+name = "just"
+source = { git = "https://github.com/IndianBoy42/tree-sitter-just", rev = "8af0aab79854aaf25b620a52c39485849922f766" }
+
+[[language]]
+name = "just"
+scope = "source.make"
```
poliorcetics, line 2431: Is this scope still the correct one ?

VuiMuich: Thanks, classic copy paste I guess :blush:

languages.toml
```diff
@@ -2421,3 +2421,17 @@ language-server = { command = "nimlangserver" }
 [[grammar]]
 name = "nim"
 source = { git = "https://github.com/aMOPel/tree-sitter-nim", rev = "240239b232550e431d67de250d1b5856209e7f06" }
+
+[[language]]
+name = "just"
+scope = "source.just"
+file-types = ["justfile", "Justfile", ".justfile", ".just"]
```
the-mikedavis, line 2428:
```suggestion
file-types = ["justfile", "Justfile", "just"]
```

I didn't catch this in the initial PR that added just but the `.`s aren't necessary: the file-types act as either extensions or whole filenames

runtime/queries/just/highlights.scm
the-mikedavis, line 1: The queries here need to use the Helix scopes which are slightly different than the neovim style ones used in the upstream repo: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example `boolean` should become `constant.builtin.boolean`, `parameter` should become `variable.parameter`, etc.

runtime/queries/just/highlights.scm
```diff
@@ -0,0 +1,35 @@
...
+
+(shebang interpreter:(TEXT) @keyword ) @comment
+
+["export" "alias" "set"] @keyword
+
+["@" "==" "!=" "+" ":="] @operator
+
+[ "(" ")" "[" "]" "{{" "}}" "{" "}"] @punctuation.bracket
+
+(ERROR) @error
```
the-mikedavis, line 35: Eventually we want to add error node highlighting with a config option. It's very noisy when writing new code for most parsers so we've been removing the highlight in general

```suggestion
```

runtime/queries/just/injections.scm
```diff
@@ -0,0 +1,14 @@
...
+        interpreter:(TEXT) @language)
+    (shebang_body) @content
+) 
+
+(source_file 
+    (item (setting lang:(NAME) @language))
+    (item (recipe (body (recipe_body) @content)))
+) 
+
+; (interpolation (expression) @just)
```
the-mikedavis, line 14:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))

(shebang_recipe 
    (shebang 
        interpreter:(TEXT) @injection.language)
    (shebang_body) @injection.content
) 

(source_file 
    (item (setting lang:(NAME) @injection.language))
    (item (recipe (body (recipe_body) @injection.content)))
) 

; ((interpolation (expression) @injection.content)
;  (#set! injection.language "just"))
```

neovim has a custom way of doing injections, we use the same way as tree-sitter's highlight crate: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection

runtime/queries/just/locals.scm
the-mikedavis, line 1: Same thing for locals as injections: we use the same way as tree-sitter https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

For these, anything that is `definition.*` should become `local.definition` and `reference.*` should become `local.reference`

runtime/queries/just/textobjects.scm
the-mikedavis, line 1: These will also need some adjusting to use Helix scopes https://docs.helix-editor.com/master/guides/textobject.html

We also don't support block, call, statement or conditional

VuiMuich:
Ok, I get it, straight out copying the queries was a bit short-sighted :joy:

I'll look into rewriting the remaining deviations today or tomorrow.
Thanks @the-mikedavis for the detailed pointers and suggestions!

runtime/queries/just/textobjects.scm
VuiMuich: Will any of these be supported in the future? Or has this been a design decision to not support those at all?

Would it be ok, to supplement these with other like `class` or `function` as I pushed with the latest commit? Or is this in any way bad and those lines should just be removed?

the-mikedavis, line 1: These should use `.around` rather than `.outer`

runtime/queries/just/locals.scm
the-mikedavis, line 1: We don't do fallback for the locals so these must either be `local.definition`, `local.reference` or `local.scope` without the `.variable`/`.function` part on the end

the-mikedavis (commented):
The other queries look good. I have just some minor comments on textobjects and locals

runtime/queries/just/textobjects.scm
```diff
@@ -0,0 +1,46 @@
...
+(parameters 
+    [(parameter) 
+    (variadic_parameters)] @parameter.inside
+    . " "? @_end
+    (#make-range! "parameter.around" @parameter.inside @_end)
+)
+
+(expression 
+    (condition) @function.inside
+) @functio.around
```
the-mikedavis, line 34:
```suggestion
) @function.around
```

runtime/queries/just/textobjects.scm
```diff
@@ -0,0 +1,46 @@
...
+    (string) @parameter.inside
+    . ","? @_end
+    (#make-range! "parameter.around" @parameter.inside @_end)
+)
+(parameters 
+    [(parameter) 
+    (variadic_parameters)] @parameter.inside
+    . " "? @_end
+    (#make-range! "parameter.around" @parameter.inside @_end)
+)
```
the-mikedavis, line 30: I think the `make-range!` property is only implemented in nvim. For now we may be only able to capture `parameter.inside` since I don't think we have anything that would work the same

the-mikedavis (approved):

pascalkuthe (approved):

VuiMuich:
Thanks for the merge and especially thanks for the patient review @the-mikedavis!

PR #6454, opened 2023-03-27
Fix clippy lints

Philipp-M:
Fix all (nightly) clippy lints, as I agree with all of them.


helix-core/src/position.rs
```diff
@@ -309,8 +309,8 @@ pub fn pos_at_visual_coords(text: RopeSlice, coords: Position, tab_width: usize)
 /// on the visual line is returned if the visual line contains any text:
 /// If the visual line at the specified offset is a virtual line generated by a `LineAnnotation`
 /// the previous char_index is returned, together with the remaining vertical offset (`virtual_lines`)
-pub fn char_idx_at_visual_offset<'a>(
-    text: RopeSlice<'a>,
+pub fn char_idx_at_visual_offset(
+    text: RopeSlice<'_>,
```
the-mikedavis, line 313: I think we can get away with not mentioning the lifetime at all here:

```suggestion
    text: RopeSlice,
```

the-mikedavis (approved):

PR #6455, opened 2023-03-27
Add changelog notes for the 23.03 release

the-mikedavis:
Release notes: https://github.com/helix-editor/website/pull/33


CHANGELOG.md
```diff
@@ -1,3 +1,271 @@
...
+- Split file modification indicator from filename statusline elements (#4731, #6036)
+- Jump to symbol ranges in LSP goto commands (#5986)
+
+Features:
+
+- Dynamic workspace symbol picker (#5055)
+- Soft-wrap (#5420, #5786, #5893, #6440)
+- Initial support for LSP snippet completions (#5864, b1f7528, #6263, bbf4800, 90348b8, f87299f, #6371)
+- Add a statusline element for showing the current version control HEAD (#5682)
+- Display LSP type hints (#5934, #6312)
```
pascalkuthe, line 22: I don't want to claim the credit here but I think mentioning #5420 would make sense here since that PR added all the virtual text API/implementation and is therefore the main reason we have this feature at all. We usually don't mention PRs twice I believe but we could make an exception for that PR since it was so large

CHANGELOG.md
```diff
@@ -1,3 +1,271 @@
...
+- `:lsp-stop` to stop the language server of the current buffer (#5964)
+- `:reset-diff-change` for resetting a diff hunk to its original text (#4974)
+
+Usability improvements and fixes:
+
+- Remove empty detail section in completion menu when LSP doesn't send details (#4902)
+- Pass client information on LSP initialization (#4904)
+- Fix behavior of `auto-completion` flag for completion-on-trigger (#5042)
+- Allow specifying environment variables for language servers in language config (#4004)
+- VCS gutter: Properly handle detached git worktrees (#5097)
```
pascalkuthe, line 43: This is actually not related to the gutter. The VCS gutter handled this case fine because we use gitoxide to load the repo.which does it's own repo.discovery. This is only related to detecting the root path (for the file picker for for example)

CHANGELOG.md
```diff
@@ -1,3 +1,271 @@
+# 23.03 (2022-03-31)
```
the-mikedavis, line 1:
```suggestion
# 23.03 (2023-03-31)
```

🙃

pascalkuthe (commented):
Thanks for putting this togetger! I noticed a few small things.


The usability improvements and bugfix section has been growing kind of large. Those are pretty hard to seperate but there is a lot of stuff in there now that I would categorize ass minor features that gets a bit lost between the bugfixes which not eberbody might read trough. For example the ability to have multiple runtime directories is a good example IMO. Not really that important and it's also good as is but might make the changes a bit more organized and break up that on really large list.

CHANGELOG.md
the-mikedavis: Whoops, good catch!

```suggestion
- Allow detached git worktrees to be recognized as root paths (#5097)
```

the-mikedavis:
Yeah I agree. I've been reserving the features part for really large stuff but I think it's actually ok to move all features under that header that aren't covered otherwise (commands for example). I think we can probably eliminate the LSP configurations section since that mostly overlaps with the new languages section

pascalkuthe:
> Yeah I agree. I've been reserving the features part for really large stuff but I think it's actually ok to move all features under that header that aren't covered otherwise (commands for example). I think we can probably eliminate the LSP configurations section since that mostly overlaps with the new languages section

Yeah both of those changes sound like good suggestions 👍

CHANGELOG.md
```diff
@@ -1,3 +1,274 @@
...
+- Avoid trailing `s` when only one file is opened (#5189)
+- Fix erroneous indent between closers of auto-pairs (#5330)
+- Expand `~` when parsing file paths in `:open` (#5329)
+- Fix theme inheritance for default themes (#5218)
+- Fix `extend_line` with a count when the current line(s) are selected (#5288)
+- Prompt: Fix autocompletion for paths containing periods (#5175)
+- Skip serializing JSONRPC params if params is null (#5471)
+- Fix interaction with the `xclip` clipboard provider (#5426)
+- Fix undo/redo execution from the command palette (#5294)
+- Fix highlighting of non-bar cursors (#5575)
```
pascalkuthe, line 95: This is actually a typo in the PR title it should be non-bkock cursors not non-bar cursors

CHANGELOG.md
```diff
@@ -1,3 +1,274 @@
...
+- Fix a panic when uncommenting a line with only a comment token (#5933)
+- Fix panic in `goto_window_center` at EOF (#5987)
+- Ignore invalid file URIs sent by a language server (#6000)
+- Decode LSP URIs for the workspace diagnostics picker (#6016)
+- Fix incorrect usages of `tab_width` with `indent_width` (#5918)
+- DAP: Send Disconnect if the Terminated event is received (#5532)
+- DAP: Validate key and index exist when requesting variables (#5628)
+- Fix indent guide rendering (#6136)
+- Fix division by zero panic (#6155)
+- Fix scrolloff at view bottom (#6142)
```
pascalkuthe, line 114: I out could include this with softwrap above since this is a fix for a regression frk #5420 and not an existing bug

CHANGELOG.md
```diff
@@ -1,3 +1,274 @@
...
+- Fix lacking space panic (#6109)
+- Send error replies for malformed and unhandled LSP requests (#6058)
+- Fix table column calculations for dynamic pickers (#5920)
+- Skip adding jumplist entries for `:<n>` line number previews (#5751)
+- Fix completion race conditions (#6173)
+- Fix `shrink_selection` with multiple cursors (#6093)
+- Fix indentation calculation for lines with mixed tabs/spaces (#6278)
+- No-op `client/registerCapability` LSP requests (#6258)
+- Send the STOP signal to all processes in the process group (#3546)
+- Fix workspace edit client capabilities declaration (7bf168d)
```
pascalkuthe, line 124: Thanks for going trough the effort of splitting changelog messages for my slightly kitchen sinkish PRs. It made sense to bundle these changes from a review perspective but probably makes generating the changelog a bit more annoying

pascalkuthe (commented):
ah just saw you akready did this after I wrote my comment. Splitting up that section looks much nicer 👍

CHANGELOG.md
```diff
@@ -1,3 +1,274 @@
...
+- Short-circuit tree-sitter and word object motions (#5851)
+- Add exit code to failed command message (#5898)
+- Make `m` textobject look for pairs enclosing selections (#3344)
+- Negotiate LSP position encoding (#5894)
+- Display deprecated LSP completions with strikethrough (#5932)
+- Add JSONRPC request ID to failed LSP/DAP request log messages (#6010, #6018)
+- Ignore case when filtering LSP completions (#6008)
+- Show current language when no arguments are passed to `:set-language` (#5895)
+- Refactor and rewrite all book documentation (#5534)
+- Check LSP renaming support before prompting for rename text (#6257)
```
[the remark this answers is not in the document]
the-mikedavis: Yeah this one was really fuzzy for me. I can see it as a bug fix though so I'll move it over 👍

CHANGELOG.md
```diff
@@ -1,3 +1,265 @@
+# 23.03 (2023-03-31)
```
pickfire, line 1: @the-mikedavis I noticed every time you had to write this by looking at the logs, even though writing it from scratch will be better for the reader, but it is a lot of work, should we add an unreleased section so each new feature change needs to be written to that so you don't have to spend so much time on this?

VERSION
```diff
@@ -1 +1 @@
-22.12
\ No newline at end of file
+23.03
```
pickfire, line 1: Should we add EOF for this?

CHANGELOG.md
pascalkuthe: see https://github.com/helix-editor/helix/pull/4150 for previous discussion. I think the points raised there were pretty good. I could comeup with a solution which would somewhat automate it but I guess Mikes point about that being an unnecessary effort still applies

Issue #6125, opened 2023-02-27
] r for Goto next reference


lukepighetti:
Right now you can show a picker containing references with `gr`

What would be nice though is to navigate to the next reference in a file with `]r`

Right now the closest approximation is doing a search for the name and iterating through

kirawi:
https://github.com/helix-editor/helix/pull/2738 should be usable for this in the meantime.

askreet:
Hey @lukepighetti, I've been working on a version of quickfix that can be populated from any file picker. The workflow would be:

* Use `gr` to find references to a symbol.
* Use `C-q` to copy the search results into a quickfix list.
* Use `]q` and `[q` to navigate between entries in the list.

This was a common workflow I'd used a lot in vim/intellij that I missed most switching to Helix. If you want to give it a try it's on [this branch](https://github.com/askreet/helix/tree/quickfix).

lukepighetti:
That sounds interesting for use in a global context, but I do wish we could do this at the local file much quicker

PR #6465, opened 2023-03-28
add goto_next_reference command

ATTron:
for #6125 

this pr add both the `] r` and `[ r` commands to goto next reference and goto previous reference respectively


helix-term/src/commands/lsp.rs
```diff
@@ -1100,6 +1100,56 @@ pub fn goto_reference(cx: &mut Context) {
     );
 }
 
+pub fn goto_reference_direction(cx: &mut Context, direction: Direction) {
+    let (view, doc) = current!(cx.editor);
+    let language_server = language_server!(cx.editor, doc);
+    let offset_encoding = language_server.offset_encoding();
+
+    let pos = doc.position(view.id, offset_encoding);
```
pickfire, line 1108: Compared to all the other `]` commands, this is the only one that relies heavily on the cursor, I am not sure if this is a good idea, I think it would be good for user to first mark and select it with `<space>r` then only allow `]r`, otherwise user move cursor around and they have to get back to the same position to do `]r`.

ATTron: that would make this very similar to the already existing cmd implemented by #2738 . i don't think we would want to duplicate the same functionality across two different cmds, but rather let this be useful for jumping without having to select first if a user chooses to and use the existing `S-h` + `)/(` if they would rather select first, then jump

SoraTenshi:
Not sure how much additional effort it would be, but having it synonymous to the current search would be beneficial i think.
Concrete: Potentially read from the configuration whether or not wrapping on searches are available and then wrap around the document if you are the the last position.

Otherwise looks very cool!

PR #6470, opened 2023-03-28
Switch default comment to #

pickfire:
Having # as comment is far more common in non-support files like random configuration files as compared to //, so it will be using the correct comment at a higher chance.

### DIFF
--- a/helix-core/src/comment.rs
+++ b/helix-core/src/comment.rs
@@ -60,7 +60,7 @@ fn find_line_comment(
 pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&str>) -> Transaction {
     let text = doc.slice(..);
 
-    let token = token.unwrap_or("//");
+    let token = token.unwrap_or("#");
     let comment = Tendril::from(format!("{} ", token));
 
     let mut lines: Vec<usize> = Vec::with_capacity(selection.len());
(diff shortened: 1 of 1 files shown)


the-mikedavis:
There is https://github.com/helix-editor/helix/pull/3799

pickfire:
Thanks for linking me to that, I think that is good, but this does a different thing, it sets the default comment character to `#` instead of `//`, it does not change the key.

the-mikedavis:
Ah whoops, I hadn't looked at the changes 😅

firstrow:
dependes. I'm perfectly fine with // for c .

pickfire:
We have no issues with c and c++ and rust or the likes with `//` because there are language servers and tree-sitter clients for those, but for those without I am not aware of anything that is using `//` as comment.

But instead, tons of random configuration that we do not support uses `#` instead of `//`, which is why I suggested this change.

helix-core/src/comment.rs
```diff
@@ -60,7 +60,7 @@ fn find_line_comment(
 pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&str>) -> Transaction {
     let text = doc.slice(..);
 
-    let token = token.unwrap_or("//");
+    let token = token.unwrap_or("#");
```
alevinval, line 63: This feels like it should be configurable. And then change the default in the configuration.

frou:
If anything is, then `#` is the one true line comment character because languages that use it get shebang support for free.

omentic:
Superceded by #12080

PR #6473, opened 2023-03-29
Add OpenCL tree-sitter and LSP config

lefp:


the-mikedavis (commented):
We could re-use C or C++ for the syntax highlighting. See https://github.com/helix-editor/helix/pull/4242 as a good example

lefp:
> We could re-use C or C++ for the syntax highlighting. See #4242 as a good example

Thanks for the example; I've done that for now.
I'd like to modify the C grammar to better match OpenCL (at least to update the keywords and built-in types), but I don't know how long that will take me. I've only started learning about treesitter yesterday.

lefp:
> I'd like to modify the C grammar to better match OpenCL.

I think I can have an acceptable version within about a week. I've set the pull request status to Draft in the meantime.

the-mikedavis (approved):

PR #6475, opened 2023-03-29
Render tab with indent-guides

### LINKED ISSUE #6051: indent-guides double character dilemma
### Steps
Assume: `:set whitespace.render all` or `editor.whitespace.render = "all"`

- open hx
- enter a new line
- press 4 spaces and write some word
- press enter to go to next line (this inserts the indent with "tab" lol)
- enable indent-guides by `:set indent-guides.render true`

**See that:**
- the indent guide vertical line is drawn
- but that seems to be drawn _in place of_ the whitespace character
- so, only 3 (not 4) spaces are now shown in first line
- and the tab indentation character is effectively eaten up entirely

### Screenrecord

https://user-images.githubusercontent.com/19423063/219942649-89e6ce5c-3cfb-42ca-bcb8-5477f0168c8f.mp4

### External References

From VSCodium:
It shows both the indentation guide, and whitespace character.

image

### Versions
- hx: 22.12 (96ff64a8)
- os: win 10 20h2

erasin:
- fix #6051
- fix when set `tab-width=0` ,will get panic `thread 'main' panicked at 'attempt to calculate the remainder with a divisor of zero', helix-core/src/graphemes.rs:20:26`

set indent guides render false

before:

after:

> this has set `tabpad = "·"`

### DIFF
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -338,16 +338,9 @@ impl<'a> TextRenderer<'a> {
 
         let tab_width = doc.tab_width();
         let tab = if ws_render.tab() == WhitespaceRenderValue::All {
-            if editor_config.indent_guides.render {
-                std::iter::once(ws_chars.tabpad)
-                    .chain(std::iter::once(ws_chars.tab))
-                    .chain(std::iter::repeat(ws_chars.tabpad).take(tab_width - 2))
-                    .collect()
-            } else {
-                std::iter::once(ws_chars.tab)
-                    .chain(std::iter::repeat(ws_chars.tabpad).take(tab_width - 1))
-                    .collect()
-            }
+            std::iter::once(ws_chars.tab)
+                .chain(std::iter::repeat(ws_chars.tabpad).take(tab_width - 1))
+                .collect()
         } else {
             " ".repeat(tab_width)
         };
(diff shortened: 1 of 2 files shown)


pascalkuthe:
This needs a smarter implementation. While tabs are most commonly used for indentation they can be used randomly in the document and with the implementation here tabs would look off in that case. Furthermore, helix allows mixing tabs and spaces for indentation (or setting a different indentation setting). In that case indent guides can dynamically appear *anywhere* inside the tab.

erasin:
Thanks for the review

I am modifying it to only handle indent guides when rendering，This will does not affect other render.

pascalkuthe:
Ah I see I missed that I guess I was tired. This should cover most things but I am kind  of concerned about just comparing the symbol here. The tab whitespace symbol is not necessarily unique. Consider the following: tabs are marked with .... And spaces with . but nbso with ,

Now if you have
`.,,,` (so a space followed by nbsp) then your implementation would do
`|.,,`.  

Also querying the buffer like this has some overhead we should add a check whether we are even rendering tabs around this so this doesn't affect people that don't use whitespace chars. It should also only work if there is a  distinct pad tab character as this workaround isn't needed otherwise. This can still lead to visual artifacts but I guess those are unlikely enough to ignore

erasin:
Sorry, I shouldn't be querying characters.

In this commit, I append `indent_tab` , when grapheme is in `indent_area`, it will be replace tab  like `virtual_tab`.

erasin:
I have tried several ways, indent_area processing does not satisfy the condition, must recalculate the rendered line content.

Issue #1565, opened 2022-01-23
Code Actions on save


archseer:
### Discussed in https://github.com/helix-editor/helix/discussions/1540

<div type='discussions-op-text'>

<sup>Originally posted by **kevinsjoberg** January 19, 2022</sup>
Helix currently supports running [Code Actions](https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_codeAction) manually by pressing `<space>+a`. Some language servers also supports running specific code actions on save, e.g., [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server#code-actions-on-save) and [gopls](https://github.com/golang/vscode-go/blob/master/docs/features.md#format-and-organize-imports).

Preferably, we extend the language server configuration to allow for specifying code actions to run on save. The actions themselves would then run asynchronously, similar to what we already do for auto formatting.</div>

archseer:
The VSCode config seems to be:

```json
"[go]": {
    "editor.codeActionsOnSave": {
        "source.organizeImports": true
    }
},
```

`source.organizeImports` and now `source.sortImports` (https://code.visualstudio.com/updates/v1_57#_sort-imports-source-action) seem to be the standardized name for this behavior and could be auto-enabled if the LSP offers them.

HoangNguyen689:
Really want to have this feature w

brielov:
Any news on this? I rely heavily on removing unused imports on save when working on typescript projects.

nrei0:
That will be great to have for Typescript as well.

rezzubs:
Is there anything left to do here now that #14481 is merged?

junglerobba:
> Is there anything left to do here now that https://github.com/helix-editor/helix/pull/14481 is merged?

There's a followup fix at https://github.com/helix-editor/helix/pull/16207, but that's only if it's relevant for your setup with multiple language servers configured, for me this has been working just fine ever since it was merged :) So I think whoever has the authority can close this.

PR #6486, opened 2023-03-30
Add code actions on save

jpttrssn:
Addresses issue #1565

* Add a per language config option for code-actions-on-save that takes a list of LSP actions to run in order during save
* Verify configured code actions against available source code actions for the entire document being saved
* Apply actions on write and write_all before auto formatting
* Refactor to reduce duplicate code
* Update languages configuration documentation

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -118,6 +118,8 @@ pub struct LanguageConfiguration {
     pub block_comment_tokens: Option<Vec<BlockCommentToken>>,
     pub text_width: Option<usize>,
     pub soft_wrap: Option<SoftWrap>,
+    #[serde(default)]
+    pub code_actions_on_save: Option<Vec<CodeActionsOnSave>>, // List of LSP code actions to be run in order upon saving
 
     #[serde(default)]
     pub auto_format: bool,
@@ -488,6 +490,13 @@ pub struct AdvancedCompletion {
     pub name: Option<String>,
     pub completion: Option<String>,
     pub default: Option<String>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub struct CodeActionsOnSave {
+    pub code_action: String,
+    pub enabled: bool,
 }
 
 #[derive(Debug, PartialEq, Eq, Clone, Deserialize, Serialize)]
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -836,7 +836,7 @@ pub fn code_actions_on_save(cx: &mut compositor::Context, doc_id: &DocumentId) {
                 // TODO: Find a better way to handle this
                 // Sleep to avoid race condition between next code action/auto-format
                 // and the textDocument/didChange notification
-                std::thread::sleep(std::time::Duration::from_millis(100));
+                std::thread::sleep(std::time::Duration::from_millis(10));
             }
         }
     }
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -551,6 +551,11 @@ struct CodeActionOrCommandItem {
     language_server_id: LanguageServerId,
 }
 
+struct CodeActionItem {
+    lsp_item: lsp::CodeAction,
+    language_server_id: LanguageServerId,
+}
+
 impl ui::menu::Item for CodeActionOrCommandItem {
     type Data = ();
     fn format(&self, _data: &Self::Data) -> Row {
@@ -712,8 +717,36 @@ pub fn code_action(cx: &mut Context) {
 
                 // always present here
                 let action = action.unwrap();
+                let Some(language_server) = editor.language_server_by_id(action.language_server_id)
+                else {
+                    editor.set_error("Language Server disappeared");
+                    return;
+                };
+                let offset_encoding = language_server.offset_encoding();
 
-                apply_code_action(editor, action);
+                match &action.lsp_item {
+                    lsp::CodeActionOrCommand::Command(command) => {
+                        log::debug!("code action command: {:?}", command);
+                        execute_lsp_command(editor, action.language_server_id, command.clone());
+                    }
+                    lsp::CodeActionOrCommand::CodeAction(code_action) => {
+                        log::debug!("code action: {:?}", code_action);
+                        let resolved_code_action =
+                            resolve_code_action(code_action, language_server);
+
+                        if let Some(ref workspace_edit) =
+                            resolved_code_action.as_ref().unwrap_or(code_action).edit
+                        {
+                            let _ = editor.apply_workspace_edit(offset_encoding, workspace_edit);
+                        }
+
+                        // if code action provides both edit and command first the edit
+                        // should be applied and then the command
+                        if let Some(command) = &code_action.command {
+                            execute_lsp_command(editor, action.language_server_id, command.clone());
+                        }
+                    }
+                }
             });
             picker.move_down(); // pre-select the first item
 
@@ -780,7 +813,7 @@ pub fn code_actions_on_save(cx: &mut compositor::Context, doc_id: &DocumentId) {
             log::debug!("Attempting code action on save {:?}", code_action_kind);
             let doc = doc!(cx.editor, doc_id);
             let full_range = Range::new(0, doc.text().len_chars());
-            let code_actions: Vec<CodeActionOrCommandItem> =
+            let code_actions: Vec<CodeActionItem> =
                 code_actions_for_range(doc, full_range, Some(vec![code_action_kind.clone()]))
                     .into_iter()
                     .filter_map(|(future, language_server_id)| {
@@ -808,10 +841,15 @@ pub fn code_actions_on_save(cx: &mut compositor::Context, doc_id: &DocumentId) {
 
                                 // Use the first matching code action
                                 if let Some(lsp_item) = actions.first() {
-                                    return Some(CodeActionOrCommandItem {
-                                        lsp_item: lsp_item.clone(),
-                                        language_server_id,
-                                    });
+                                    return match lsp_item {
+                                        CodeActionOrCommand::CodeAction(code_action) => {
+                                            Some(CodeActionItem {
+                                                lsp_item: code_action.clone(),
+                                                language_server_id,
+                                            })
+                                        }
+                                        _ => None,
+                                    };
                                 }
                             }
                         }
@@ -831,55 +869,52 @@ pub fn code_actions_on_save(cx: &mut compositor::Context, doc_id: &DocumentId) {
                     code_action.lsp_item,
                     code_action.language_server_id
                 );
-                apply_code_action(cx.editor, &code_action);
+                let Some(language_server) = cx
+                    .editor
+                    .language_server_by_id(code_action.language_server_id)
+                else {
+                    log::error!(
+                        "Language server disappeared {:?}",
+                        code_action.language_server_id
+                    );
+                    continue;
+                };
 
-                // TODO: Find a better way to handle this
-                // Sleep to avoid race condition between next code action/auto-format
-                // and the textDocument/didChange notification
-                std::thread::sleep(std::time::Duration::from_millis(10));
+                let offset_encoding = language_server.offset_encoding();
+
+                let resolved_code_action =
+                    resolve_code_action(&code_action.lsp_item, language_server);
+
+                if let Some(ref workspace_edit) = resolved_code_action
+                    .as_ref()
+                    .unwrap_or(&code_action.lsp_item)
+                    .edit
+                {
+                    let _ = cx
+                        .editor
+                        .apply_workspace_edit_sync(offset_encoding, workspace_edit);
+                }
             }
         }
     }
 }
 
-fn apply_code_action(editor: &mut Editor, action: &CodeActionOrCommandItem) {
-    let Some(language_server) = editor.language_server_by_id(action.language_server_id) else {
-        editor.set_error("Language Server disappeared");
-        return;
-    };
-    let offset_encoding = language_server.offset_encoding();
-
-    match &action.lsp_item {
-        lsp::CodeActionOrCommand::Command(command) => {
-            log::debug!("code action command: {:?}", command);
-            execute_lsp_command(editor, action.language_server_id, command.clone());
-        }
-        lsp::CodeActionOrCommand::CodeAction(code_action) => {
-            log::debug!("code action: {:?}", code_action);
-            // we support lsp "codeAction/resolve" for `edit` and `command` fields
-            let mut resolved_code_action = None;
-            if code_action.edit.is_none() || code_action.command.is_none() {
-                if let Some(future) = language_server.resolve_code_action(code_action.clone()) {
-                    if let Ok(response) = helix_lsp::block_on(future) {
-                        if let Ok(code_action) = serde_json::from_value::<CodeAction>(response) {
-                            resolved_code_action = Some(code_action);
-                        }
-                    }
+pub fn resolve_code_action(
+    code_action: &CodeAction,
+    language_server: &Client,
+) -> Option<CodeAction> {
+    // we support lsp "codeAction/resolve" for `edit` and `command` fields
+    let mut resolved_code_action = None;
+    if code_action.edit.is_none() || code_action.command.is_none() {
+        if let Some(future) = language_server.resolve_code_action(code_action.clone()) {
+            if let Ok(response) = helix_lsp::block_on(future) {
+                if let Ok(code_action) = serde_json::from_value::<CodeAction>(response) {
+                    resolved_code_action = Some(code_action);
                 }
-            }
-            let resolved_code_action = resolved_code_action.as_ref().unwrap_or(code_action);
-
-            if let Some(ref workspace_edit) = resolved_code_action.edit {
-                let _ = editor.apply_workspace_edit(offset_encoding, workspace_edit);
-            }
-
-            // if code action provides both edit and command first the edit
-            // should be applied and then the command
-            if let Some(command) = &code_action.command {
-                execute_lsp_command(editor, action.language_server_id, command.clone());
             }
         }
     }
+    resolved_code_action
 }
 
 pub fn execute_lsp_command(
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -546,9 +546,9 @@ pub fn workspace_diagnostics_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
-pub struct CodeActionOrCommandItem {
-    pub lsp_item: lsp::CodeActionOrCommand,
-    pub language_server_id: LanguageServerId,
+struct CodeActionOrCommandItem {
+    lsp_item: lsp::CodeActionOrCommand,
+    language_server_id: LanguageServerId,
 }
 
 impl ui::menu::Item for CodeActionOrCommandItem {
@@ -842,7 +842,7 @@ pub fn code_actions_on_save(cx: &mut compositor::Context, doc_id: &DocumentId) {
     }
 }
 
-pub fn apply_code_action(editor: &mut Editor, action: &CodeActionOrCommandItem) {
+fn apply_code_action(editor: &mut Editor, action: &CodeActionOrCommandItem) {
     let Some(language_server) = editor.language_server_by_id(action.language_server_id) else {
         editor.set_error("Language Server disappeared");
         return;
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1,20 +1,24 @@
-use futures_util::{stream::FuturesOrdered, FutureExt};
+use futures_util::{
+    stream::{FuturesOrdered, FuturesUnordered},
+    FutureExt,
+};
 use helix_lsp::{
     block_on,
     lsp::{
-        self, CodeAction, CodeActionOrCommand, CodeActionTriggerKind, DiagnosticSeverity,
-        NumberOrString,
+        self, CodeAction, CodeActionKind, CodeActionOrCommand, CodeActionTriggerKind,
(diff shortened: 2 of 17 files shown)


cd-a:
Nice!

Does this work with https://github.com/helix-editor/helix/pull/2507?

As far as I can see, that was the only feature that was still not working there

jpttrssn:
This would need some work to be compatible. I can work on adding my changes on top of that PR.

cd-a:
That would be amazing. That PR should be merged soon.

helix-view/src/document.rs
```diff
@@ -1331,6 +1414,11 @@ impl Document {
         &self.text
     }
 
+    #[inline]
+    pub fn full_range(&self) -> Range {
+        Range::new(0, self.text.len_chars())
+    }
+
```
pascalkuthe, line 1421: I dont think this needs a dedicated helper method. We have far more common operations that don't have helper methods. Adding too many helper methods clutters the (already very large) document impls even more

helix-view/src/document.rs
```diff
@@ -633,6 +634,88 @@ impl Document {
...
+                        },
+                        _ => false,
+                    }) {
+                        log::error!(
+                            "Configured code action on save is invalid {:?}",
+                            configured_action
+                        );
+                    }
+                })
+            }
```
pascalkuthe, line 686: I don't think these code actions are necessarily returned on every import so this would potentially generate a lot of errors. This should atlest be an info` log but this also `O(NM)` so I am not sure we really need this anyway? 

Also, I think a for loop is basically always more idiomatic than `for_each`

helix-view/src/document.rs
```diff
@@ -633,6 +634,88 @@ impl Document {
...
+            let mut code_actions = match response {
+                Some(value) => value,
+                None => helix_lsp::lsp::CodeActionResponse::default(),
+            };
+            log::debug!("Available code actions {:?}", code_actions);
+            code_actions.retain(|action| {
+                matches!(
+                    action,
+                    helix_lsp::lsp::CodeActionOrCommand::CodeAction(x) if x.disabled.is_none() &&
+                        code_actions_on_save.iter().any(|a| match &x.kind {
```
pascalkuthe, line 665: probably a good idea to make this a `HashSet` so we avoid being `O(MN)` here even if `M` and `N` is usually small frserializing to a `HashSet` just takes replacing `Vec` with `HashSet`

helix-term/src/commands/lsp.rs
```diff
@@ -544,31 +541,9 @@ fn action_fixes_diagnostics(action: &CodeActionOrCommand) -> bool {
...
-                    selection_range
-                        .overlaps(&helix_core::Range::new(diag.range.start, diag.range.end))
-                })
-                .map(|diag| diagnostic_to_lsp_diagnostic(doc.text(), diag, offset_encoding))
-                .collect(),
-            only: None,
-            trigger_kind: Some(CodeActionTriggerKind::INVOKED),
-        },
-    ) {
+    let future = match doc.code_actions(selection_range) {
```
pascalkuthe, line 546: I don't think we should move this to a function on the document. This doesn't need to be moved into helix-view for this PR to work. Just moving this to a separate function inside `commands/lsp.rs` works just as well. We generally try to keep the LSP-related stuff separate from the core document primitives. Same for the other function you added to the document (which should also just be a standalone function in this file)

pascalkuthe (commented):
Thanks for taking this on! I left some minor comments. I think we will likely land #2507 before this one since that has been waiting around for a longtime and while sometimes merging bugfixes before that PR makes sense ideally I think we should holdoff adding new LSP related features until after that lands.

helix-term/src/commands/lsp.rs
jpttrssn: Alright, cool. That makes sense.

helix-view/src/document.rs
jpttrssn: I agree this isn't great. I struggled to find a good solution for this. Ideally this would be done at the same time as the filtering with the `code_actions.retain()` above. Perhaps a `for` loop there instead? I can play around with it some more and see if I can come up with something. Otherwise I'm fine with leaving this out. It's a nice to have, but like you said perhaps not needed.

cd-a:
https://github.com/helix-editor/helix/pull/2507 has been merged now, which should enable this PR

jpttrssn:
Great! I'll update this PR as soon as I've got my changes working with master.

ravsii:
@jpttrssn this only adds pre-defined LSP actions on save, but no support for third-party commands, like vscode's [run on save](https://marketplace.visualstudio.com/items?itemName=emeraldwalk.RunOnSave)?

just asking

pascalkuthe:
No that's out of scope for this PR.  This PR should only implement support for the relevant part of the LSP spec

cd-a:
Any luck getting it rebased on latest master? Would love to test it again

jpttrssn:
I've made some progress, but haven't had much time to work on it. I'll try to make some time, though. I also want to get this done.

jpttrssn:
Finally got this rebased with master. Manually tested with the following language config:

```
[[language]]
name = "go"
code-actions-on-save = ["source.organizeImports", "source.invalid"]
```

cd-a:
Great news, thanks @jpttrssn !

Any idea how to test this with TS/TSX?

I tried this under the language config but it doesn't have any effect

```
code-actions-on-save = ["source.fixAll.eslint"]
```

helix-core/src/syntax.rs
```diff
@@ -105,6 +105,8 @@ pub struct LanguageConfiguration {
     pub comment_token: Option<String>,
     pub text_width: Option<usize>,
     pub soft_wrap: Option<SoftWrap>,
+    #[serde(default)]
+    pub code_actions_on_save: HashSet<String>, // List of LSP code actions to be run in order upon saving
```
archseer, line 109: A `HashSet` is ordered based on the item's hash value though. Let's just use a `Vec` here since I can also imagine usecases where you would want to execute action A, B then A again

helix-term/src/commands/lsp.rs
```diff
@@ -770,6 +721,131 @@ pub fn code_action(cx: &mut Context) {
...
+        return None;
+    }
+
+    let full_range = Range::new(0, doc.text().len_chars());
+
+    Some(
+        code_actions_for_range(doc, full_range)
+            .into_iter()
+            .map(|(request, ls_id)| {
+                let code_actions_on_save = code_actions_on_save_cfg.clone();
```
archseer, line 784: This assumes that we can trigger multiple code actions in parallel, which works in some cases, but one action might impact another action (e.g. if A removes a function, and B formats the file). Maybe to simplify we should limit to a single code action on save (since that's the majority usecase anyway)

jpttrssn: If I'm understanding the LSP spec as long as the requests are in the correct order, the lsp should determine how best to execute those requests.

https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#messageOrdering

archseer: Right, but if you send actions A and B in parallel, the server is computing the edits based on that initial state. Server only considers the change applied when we send back a `didChange` event. Helix does use OT internally to transform edits (so initial -> A -> B can be applied) but it's not necessarily going to produce the same result as if we requested A, applied, then requested B and applied B.

jpttrssn: But over multiple language servers we can't guarantee the order. That's what you were getting at, I assume. Simplifying to a single code action would definitely make this problem go away, though then we are not feature parity with VSCode, if that is important.

jpttrssn: > Right, but if you send actions A and B in parallel, the server is computing the edits based on that initial state. Server only considers the change applied when we send back a `didChange` event. Helix does use OT internally to transform edits (so initial state -> A -> B can be applied) but it's not necessarily going to produce the same result as if we requested A, applied, then requested B and applied B.

Ok, I see. Thanks! Let me know which direction we want to take and I can make the changes.

jpttrssn:
> 
> Any idea how to test this with TS/TSX?
> 
> I tried this under the language config but it doesn't have any effect
> 
> ```
> code-actions-on-save = ["source.fixAll.eslint"]
> ```

`source.fixAll.eslint` appears to be specific to the VSCode Eslint extention. See: https://github.com/microsoft/vscode-eslint#settings-options

Addtionally, this PR only allows executing code actions that are returned from the LSP selecting over the range of the entire document. It would be the same as if you did `%-space-a` in the editor. I based this off of the PR to add code actions on save to VSCode: https://github.com/microsoft/vscode/pull/48086

There is definitely room for adding more functionality, but perhaps out of scope for this PR?

rcorre:
This is great, I'm using it with `source.organizeImports` in Go and it seems to work well.

Out of curiosity, how do you know what the name of an "action" is, or what actions a particular LSP supports? I just copied "source.organizeImports" from the docs in the PR, but I have no idea how I'd figure out what else I could put in there, or if it would be the same for other language servers.

jpttrssn:
> Out of curiosity, how do you know what the name of an "action" is, or what actions a particular LSP supports? I just copied "source.organizeImports" from the docs in the PR, but I have no idea how I'd figure out what else I could put in there, or if it would be the same for other language servers.

There are only two `source` actions defined in the LSP spec (`source.organizeImports` and `source.fixAll`). However, the spec leaves the option for a language server or extension to provide additional `source` actions. I have myself yet to find a LSP I can use to test this PR with a `source.fixAll`.

As a side note, I'm in the process of re-working this PR to apply the configured code actions in order (applied individually) and enabling the `source.fixAll` option.

jpttrssn:
This version will now run code actions, formatting and saving in order. It was a bit tricky to figure out a way to do this. I had to solve around the following constraints:

1. Current job implementation does not allow for running jobs in order
2. Each code action requires a async function call with a callback using the non-thread safe Editor
3. Formatting requires knowledge of the current document version after any code action changes

In order to solve this I ended up needing to run everything in the main thread. A downside to this is that there is no concurrency when saving multiple files. One plus is that there is now only one code path to save a file.

Perhaps there is a better solution I'm not seeing? Making the Editor thread safe, perhaps, but that seems like a big change. @archseer @pascalkuthe

pascalkuthe:
I don't think we should ever make `Editor` thread save, that would require lots of locking and is not really necessary IMO (keeping the main editor state single threaded make it simpler to impose a chronological order on events too)

mariansimecek:
Hi,
can´t get it working with typescript LSP.
I have this in `languages.toml`

```toml
 [[language]]
 name = "typescript"
 code-actions-on-save = ["source.sortImports"]
```

logs:

```
2023-11-09T20:39:17.277 helix_term::commands [DEBUG] Attempting code action on save "source.sortImports"
2023-11-09T20:39:17.277 helix_lsp::transport [INFO] typescript-language-server -> {"jsonrpc":"2.0","method":"textDocument/codeAction","params":{"context":{"diagnostics":[{"code":6133,"message":"'NodeTracing' is declared but its value is never read.","range":{"end":{"character":40,"line":0},"start":{"character":0,"line":0}},"severity":4,"source":"typescript"},{"code":6133,"message":"'Mod' is declared but its value is never read.","range":{"end":{"character":27,"line":2},"start":{"character":24,"line":2}},"severity":4,"source":"typescript"},{"code":6133,"message":"'something' is declared but its value is never read.","range":{"end":{"character":37,"line":3},"start":{"character":0,"line":3}},"severity":4,"source":"typescript"},{"code":6133,"message":"'Agent' is declared but its value is never read.","range":{"end":{"character":29,"line":4},"start":{"character":0,"line":4}},"severity":4,"source":"typescript"},{"code":6133,"message":"'aa' is declared but its value is never read.","range":{"end":{"character":6,"line":8},"start":{"character":4,"line":8}},"severity":4,"source":"typescript"},{"code":6133,"message":"'aa2' is declared but its value is never read.","range":{"end":{"character":7,"line":9},"start":{"character":4,"line":9}},"severity":4,"source":"typescript"}],"triggerKind":1},"range":{"end":{"character":0,"line":10},"start":{"character":0,"line":0}},"textDocument":{"uri":"file:///home/marian-simecek/dev/private/base-ts-proj/index.ts"}},"id":1}
2023-11-09T20:39:17.346 helix_lsp::transport [INFO] typescript-language-server <- {"jsonrpc":"2.0","id":1,"result":[{"title":"Remove import from 'inspector'","command":{"title":"Remove import from 'inspector'","command":"_typescript.applyWorkspaceEdit","arguments":[{"documentChanges":[{"textDocument":{"uri":"file:///home/marian-simecek/dev/private/base-ts-proj/index.ts","version":0},"edits":[{"range":{"start":{"line":0,"character":0},"end":{"line":1,"character":0}},"newText":""}]}]}]},"kind":"quickfix"},{"title":"Convert default export to named export","kind":"refactor","disabled":{"reason":"This file already has a default export"}},{"title":"Convert named export to default export","kind":"refactor","disabled":{"reason":"This file already has a default export"}},{"title":"Extract to typedef","kind":"refactor","disabled":{"reason":"Selection is not a valid type node"}},{"title":"Extract to type alias","kind":"refactor","disabled":{"reason":"Selection is not a valid type node"}},{"title":"Extract to interface","kind":"refactor","disabled":{"reason":"Selection is not a valid type node"}},{"title":"Move to a new file","kind":"refactor.move","command":{"title":"Move to a new file","command":"_typescript.applyRefactoring","arguments":[{"file":"/home/marian-simecek/dev/private/base-ts-proj/index.ts","startLine":1,"startOffset":1,"endLine":11,"endOffset":1,"refactor":"Move to a new file","action":"Move to a new file"}]}},{"title":"Add braces to arrow function","kind":"refactor","disabled":{"reason":"Could not find a containing arrow function"}},{"title":"Remove braces from arrow function","kind":"refactor","disabled":{"reason":"Could not find a containing arrow function"}},{"title":"Convert to template string","kind":"refactor","disabled":{"reason":"Can only convert string concatenation"}},{"title":"Convert to optional chain expression","kind":"refactor","disabled":{"reason":"Could not find convertible access expression"}},{"title":"Extract function","kind":"refactor","disabled":{"reason":"Cannot extract import statement."}},{"title":"Extract constant","kind":"refactor","disabled":{"reason":"Cannot extract import statement."}},{"title":"Generate 'get' and 'set' accessors","kind":"refactor","disabled":{"reason":"Could not find property for which to generate accessor"}},{"title":"Infer function return type","kind":"refactor","disabled":{"reason":"Return type must be inferred from a function"}}]}
2023-11-09T20:39:17.346 helix_lsp::transport [INFO] typescript-language-server <- [{"command":{"arguments":[{"documentChanges":[{"edits":[{"newText":"","range":{"end":{"character":0,"line":1},"start":{"character":0,"line":0}}}],"textDocument":{"uri":"file:///home/marian-simecek/dev/private/base-ts-proj/index.ts","version":0}}]}],"command":"_typescript.applyWorkspaceEdit","title":"Remove import from 'inspector'"},"kind":"quickfix","title":"Remove import from 'inspector'"},{"disabled":{"reason":"This file already has a default export"},"kind":"refactor","title":"Convert default export to named export"},{"disabled":{"reason":"This file already has a default export"},"kind":"refactor","title":"Convert named export to default export"},{"disabled":{"reason":"Selection is not a valid type node"},"kind":"refactor","title":"Extract to typedef"},{"disabled":{"reason":"Selection is not a valid type node"},"kind":"refactor","title":"Extract to type alias"},{"disabled":{"reason":"Selection is not a valid type node"},"kind":"refactor","title":"Extract to interface"},{"command":{"arguments":[{"action":"Move to a new file","endLine":11,"endOffset":1,"file":"/home/marian-simecek/dev/private/base-ts-proj/index.ts","refactor":"Move to a new file","startLine":1,"startOffset":1}],"command":"_typescript.applyRefactoring","title":"Move to a new file"},"kind":"refactor.move","title":"Move to a new file"},{"disabled":{"reason":"Could not find a containing arrow function"},"kind":"refactor","title":"Add braces to arrow function"},{"disabled":{"reason":"Could not find a containing arrow function"},"kind":"refactor","title":"Remove braces from arrow function"},{"disabled":{"reason":"Can only convert string concatenation"},"kind":"refactor","title":"Convert to template string"},{"disabled":{"reason":"Could not find convertible access expression"},"kind":"refactor","title":"Convert to optional chain expression"},{"disabled":{"reason":"Cannot extract import statement."},"kind":"refactor","title":"Extract function"},{"disabled":{"reason":"Cannot extract import statement."},"kind":"refactor","title":"Extract constant"},{"disabled":{"reason":"Could not find property for which to generate accessor"},"kind":"refactor","title":"Generate 'get' and 'set' accessors"},{"disabled":{"reason":"Return type must be inferred from a function"},"kind":"refactor","title":"Infer function return type"}]
2023-11-09T20:39:17.346 helix_term::commands [DEBUG] Code action on save not found "source.sortImports"
2023-11-09T20:39:17.346 helix_view::editor [ERROR] editor error: Code Action not found: "source.sortImports"

```

I tried golang and it works as expected. 

Am I doing something wrong?

jpttrssn:
@mariansimecek Looks like for the [typescript-language-server](https://github.com/typescript-language-server/typescript-language-server#code-actions-on-save) the code action should be `source.sortImports.ts`. But regardless, none of the Typescript specific `source.*.ts` code actions are being returned as options from the `typescript-language-server`, whether during on-save or within the editor itself. I was able to reproduce this. There might be some issue when the LSP announces these actions. I can dig in a little here, but it doesn't look like this is an issue specific to this PR.

Do you get the expected code actions from within the editor when selecting over the entire file; `%-space-a`?

mariansimecek:
> Do you get the expected code actions from within the editor when selecting over the entire file; `%-space-a`?

Unfortunately, neither.

quantonganh:
> This makes working with helix on golang code a much better experience (manually doing imports stuff is a pain)

@ds-281 Have you tried this?

```toml
[[language]]
name = "go"
formatter = { command = "goimports"}
```

rcorre:
@quantonganh thank you! I thought `goimports` needed a filename, as the docs don't suggest it can use stdin/stdout, but it in fact works fine.
This would still be a nice feature to have, so you could take advantage of language server features without needing a separate binary, but that's a good work around for now.

brielov:
Any news on this?

jpttrssn:
This probably needs a revisit at this point. From the sound of it this PR might benefit from the new "event system" that came out in the latest release (24.03). I'll see if I can find some time to see what's possible there.

ds-281:
That would be much appreciated, thank you.

It's nice that the `goimports` trick works, but relying on it adds an extra dependency to the system that feels unnecessary and, to my understanding, relies on the fact that only one action is needed (as opposed to this PR, which takes a list of LSP actions to run upon save).

Philipp-M:
(Just a suggestion: I'd use additional commits instead of rebasing/ammending, so that it's easier to track what has changed since the last push)

jpttrssn:
> (Just a suggestion: I'd use additional commits instead of rebasing/ammending, so that it's easier to track what has changed since the last push)

OK! Didn't think anyone was paying attention. I'm currently working through issues getting my tests to go through on windows and didn't think anyone would be interested in seeing those commits. Switched to doing this on my fork instead.

ds-281:
Indeed, I'm also following this issue, and since I'm not overly familiar with Rust (and this project's structure) I would have had an easier time to compare the changes incrementally.

Of course, once the work is done and ready for review, squash-rebasing everything in a perfect set of commits without mistakes makes sense if that's what you prefer.

jpttrssn:
@archseer @pascalkuthe 

Some notable changes in this latest revision are the addition of integration tests, ensuring that `write_all_quit` works and running the code actions directly in the `typed::write_*` functions (instead of a job).

Note that there is a race condition between the `textDocument/didChange` notification and running the next code action and/or auto-format which I've "solved" with a sleep. Obviously not an ideal solution. This also causes the lsp integration tests on the Windows runner to fail unless the tests are not run in parallel. It looks like there is a TODO to make applying work space edits transactional, which sounds like it might help here. Other ideas and/or guidance welcome.

Also, I experimented with using a job to handle the code actions, but this causes a panic with` write_quit/write_all_quit` (i.e.,` block_try_flush_writes`) because of the need to use `helix_lsp::block_on` in the callback in order to both call async lsp functions and have access to the editor for applying workspace edits. If running this in a job is preferred, one workaround could be to run multiple jobs with a single code action (per language server) in a sequence (which I have experimented some with), but would be more complex and then there's still the race condition mentioned above.

Long story short this felt like the most straight forward approach and a good starting point with room for future improvements. Let me know what you think!

the-mikedavis:
I'm very wary of adding integration tests against a language server. As you have in the code already we need to install the language server in CI and require that you install it if you want to run the tests manually, the tests need to sleep pretty pervasively to wait for the language server to respond, and we need to work around language-server-specific oddities like the LF line-ending setting here. If we want to test different features we need to test against many servers: if we want to test pull diagnostics (#11315) for example we need to depend on a version of language servers that uses that feature. Plus I don't think that the code paths involved with this feature will change often enough that the tests would catch a breakage that woludn't be caught in review. On balance I don't think LSP integration tests are worth their added weight and I would prefer we test LSP features manually.

What do you think @archseer & @pascalkuthe?

jpttrssn:
In [c314acb](https://github.com/helix-editor/helix/pull/6486/commits/c314acb304857a8fe5032a133b7d13996cdd62f0) I resolved the need to sleep to fix the race condition by allowing the textDocument/didChange notification to be sent synchronously (only during on save). Tried to have this affect as little other code as I could.

Notarin:
Is this stale or are there roadblocks?

junglerobba:
Besides needing a rebase, is there anything left to do here? If it's just that or some minor stuff but the overall approach would be accepted and it's just a lack of time, I'd be interested in picking this up. In fact I'm already running this rebased on latest master on my own machine.

jpttrssn:
I'm still waiting on feedback from the maintainers.

@junglerobba you are more than welcome to pick this up. I don't use Go on a daily basis so keeping this up-to-date with master isn't top of mind. To summarize my previous comments I would not call this an elegant solution, but it does work. Glad you're getting some use out of it!

the-mikedavis:
We just merged in https://github.com/helix-editor/helix/pull/14481 with further edits. Thanks!

PR #6493, opened 2023-03-30
Update tree-sitter-git-commit

the-mikedavis:
The last update introduced a bug with comments where a comment would be recognized as a message if there were multiple newlines between the last message or subject and the comment, causing a noticeable change in highlighting. This change fixes that behavior.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1194,7 +1194,7 @@ text-width = 72
 
 [[grammar]]
 name = "git-commit"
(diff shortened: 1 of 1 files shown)


the-mikedavis:
(Added this to 23.03 since the bug is easy to trigger and easy to fix)

pascalkuthe (approved):
lets get this in before release :+1:  it's one of the most commonly used languages after all :P

PR #6504, opened 2023-03-31
Fix aarch64 macos release builds

### LINKED ISSUE #6495: Missing build (aarch64-macos) for 22.03 release
The new 23.03 release (https://github.com/helix-editor/helix/releases/tag/23.03) doesn't have an ARM version for OSX, whereas one exists for the previous releases.

Was expecting to find the release at https://github.com/helix-editor/helix/releases/download/23.03/helix-23.03-aarch64-macos.tar.xz

Edit: Looks like build issues mean it was disabled earlier
https://github.com/helix-editor/helix/commit/e59cb19892c7b3987d9e27ae8505a5a42bc4e07d

the-mikedavis:
Fixes #6495

See the dry-run https://github.com/helix-editor/helix/actions/runs/4576449372

This could be solved instead by using `rustup target add ${{ matrix.target }}` but that would be papering over the problem: we do install stable-aarch64-apple-darwin but that's for the wrong Rust version. `cargo` looks to use 1.65 because of the `rust-toolchain.toml` file but `dtolnay/rust-toolchain` installs it for stable which is 1.68.2 currently.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -86,12 +86,12 @@ jobs:
           target: x86_64-pc-windows-msvc
           cross: false
         # 23.03: build issues
-        # - build: aarch64-macos
-        #   os: macos-latest
-        #   rust: stable
-        #   target: aarch64-apple-darwin
-        #   cross: false
-        #   skip_tests: true  # x86_64 host can't run aarch64 code
+        - build: aarch64-macos
+          os: macos-latest
+          rust: stable
+          target: aarch64-apple-darwin
+          cross: false
+          skip_tests: true  # x86_64 host can't run aarch64 code
         # - build: x86_64-win-gnu
         #   os: windows-2019
         #   rust: stable-x86_64-gnu
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -255,7 +255,7 @@ jobs:
                   exe=".exe"
               fi
               pkgname=helix-$GITHUB_REF_NAME-$platform
-              mkdir $pkgname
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -114,6 +114,12 @@ jobs:
...
 
+      # The rust-toolchain action ignores rust-toolchain.toml files.
+      # Removing this before building with cargo ensures that the rust-toolchain
+      # is considered the same between installation and usage.
+      - name: Remove the rust-toolchain.toml file
+        run: rm rust-toolchain.toml
+
       - name: Install ${{ matrix.rust }} toolchain
         uses: dtolnay/rust-toolchain@master
         with:
```
pascalkuthe, line 125: we might want to force this to MSRV (`1.65`) instead of stable?

the-mikedavis: Since we control the runners doing the building here I think it's ok to use a newer Rust. We could definitely pin this at the MSRV though to reduce the chances that a future Rust would give compilation errors for code that is valid when compiled by the MSRV. What do you think @archseer?

archseer: Similar to ripgrep we want to build with the latest nightly to get as many optimizations as we can. The MSRV is just there for backwards compatibility

archseer: e.g. ripgrep uses nightly: https://github.com/BurntSushi/ripgrep/blob/041544853c86dde91c49983e5ddd0aa799bd2831/.github/workflows/release.yml#L69-L97

PR #6511, opened 2023-03-31
Only apply completions on confirmation

### LINKED ISSUE #6259: Snippets insert literal text
Given:

```
std::vector<int> i;
i.at|<Tab>
```

You get

```
std::vector<int> i;
i.at(size_type n|)
```

Typing inserts text after `n`, but the `size_type n` is left in place.

This started as of d2af31b916114dd7ebbc3390eb44c0d72e06ec50 and is still present on `master` as of today. 

https://asciinema.org/a/UidjhxmHTQmLsUHO2IKPawgwT

### LINKED ISSUE #6261: Svelte duplicated auto-imports on tab
### Summary

SvelteKit auto-generates types, so I had to make a few rounds over the options in the popup with tab, and it duplicated the import as many times as I tabbed.

See video below:
[Screencast from 2023-03-10 .webm](https://user-images.githubusercontent.com/51718252/224455818-3d5129e7-40a0-4ae4-a9ec-4584f42ad561.webm)

https://user-images.githubusercontent.com/51718252/224455818-3d5129e7-40a0-4ae4-a9ec-4584f42ad561.webm

### Reproduction Steps

 

1. Run `npm create svelte@latest` then select the following options:
**Which Svelte app template?**
`Skeleton project`
**Add type checking with Typescript?**
`Yes, using Typescript syntax`
Make sure to install the dependencies **AND** start the dev server.

2. Create `src/routes/+page.ts` whit the following content:
``` javascript
import type { PageLoad } from "./$types";

export const load = (() => {
	return {};
}) satisfies PageLoad;
```

3. Go to `src/routes/+page.svelte` insert `<script lang="ts"></script>`

4. While still in `src/routes/+page.svelte`, inside the script tag write `export let data: ActionData` and here cycle, with `TAB`, through auto-completion options a couple times, making sure you let the docs to popup and hit `ENTER`

### Helix log

```
```

</details>

### Platform

Fedora Linux 37 (Workstation Edition)

### Terminal Emulator

gnome-terminal

### Helix Version

helix 22.12

pascalkuthe:
Closes #6259
Closes #6261 (and a similar bug with the typesecript LSP reported on matrix)

Helix currently immediately "partly" (without additional edits) applies completion items when they are selected. This fundamentally breaks assumptions from the LSP protocol. LSP protocol relies on rerequsting completions dynamically. We already do this on idle timeout, but our support is subpar right now (we ignore `incomplete` completion responses) so rerequsting completions will become even more critical.

However, by immediately applying the completions we easily break rerequests for snippets. Snippets move the cursor (inside the applies completion) and usually are more complex syntactic items that completely discard any existing filters. 

I checked out other editors and all IDEs I tested (VSCode, Clion/Intellij, Sublime) only apply completions once explicitly selected. To make the experience smoother (and inline with other IDEs) I changed the completion menu to always automatically select the first completion option. This means you can keep typing until the desired completion is on top and the just hit `<ret>` once to confirm. 

This change also makes the completion menu work better as a sort of inline picker. The text you type is used as a fuzzy match query. You can scroll without messing up this query and confirm your pick by hitting enter.

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -220,8 +220,6 @@ impl Completion {
                 }
                 PromptEvent::Update => {}
                 PromptEvent::Validate => {
-                    // if more text was entered, remove it
-                    doc.restore(view, &savepoint);
                     // always present here
                     let item = item.unwrap();
 
@@ -235,13 +233,6 @@ impl Completion {
                         replace_mode,
                     );
 
-                    doc.apply(&transaction, view.id);
-
-                    editor.last_completion = Some(CompleteAction {
-                        trigger_offset,
-                        changes: completion_changes(&transaction, trigger_offset),
-                    });
-
                     // apply additional edits, mostly used to auto import unqualified types
                     let resolved_item = if item
                         .additional_text_edits
@@ -254,6 +245,14 @@ impl Completion {
                         Self::resolve_completion_item(doc, item.clone())
                     };
 
+                    // if more text was entered, remove it
+                    doc.restore(view, &savepoint);
+                    doc.apply(&transaction, view.id);
+
+                    editor.last_completion = Some(CompleteAction {
+                        trigger_offset,
+                        changes: completion_changes(&transaction, trigger_offset),
+                    });
                     if let Some(additional_edits) = resolved_item
                         .as_ref()
                         .and_then(|item| item.additional_text_edits.as_ref())
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -213,36 +213,15 @@ impl Completion {
 
             let (view, doc) = current!(editor);
 
-            // if more text was entered, remove it
-            doc.restore(view, &savepoint);
-
             match event {
                 PromptEvent::Abort => {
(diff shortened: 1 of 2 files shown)


Philipp-M:
I mostly agree with all the changes.

I'm not sure if selecting the first completion option is really the best choice always.
If I want to just insert a new line it can be annoying I guess (I think that might also be the main reason why it's not selected by default currently?). I think this should at least be configurable, or there should be an easy way around, so that I can just press `<ret>` for a new line.
Though I think preselecting the top completion item is probably more often useful, especially with the proposal below.

I like to get immediate feedback when selecting the completion items, maybe we can add this feature back via the virtual text API, by just displaying what might change when selecting an option (similar as virtual inlay hints)?

faldor20:
I think if the `<ret>` to complete is a problem rebinding it to `ctr+ret`  would be an easy fix. Even `tab` would be a good option if you're happy to use `ctrl+n` and `ctr+p` to navigate.

pascalkuthe:
Closing in favor of https://github.com/helix-editor/helix/pull/6594 as that contains additional fixes and handles more cases and is much closer to existing behavior (and feels better in my subjective opinion)

Issue #6022, opened 2023-02-16
Gleam LSP fails to format because the end range is out of bounds


the-mikedavis:
### Summary

The textDocument/formatting response from the gleam LSP sends a line number that is way past the end of the document:

```json
{
  "jsonrpc": "2.0",
  "id": 1,
  "result": [
    {
      "newText": "fn main() {\n  let x = 5\n}\n",
      "range": {
        "end": {
          "character": 0,
          "line": 4294967295
        },
        "start": {
          "character": 0,
          "line": 0
        }
      }
    }
  ]
}
```

That causes `lsp_pos_to_pos` to return `None` which causes an early return here: https://github.com/helix-editor/helix/blob/ce0837dbb75badf39c9b1ac251fba9c3efbc57c4/helix-lsp/src/lib.rs#L323-L327

making the Transaction a no-op.

### Reproduction Steps

 

I tried this:

1. `hx file.gleam`
2. add some contents like
    ```gleam
    fn main() {
        let x = 5
    }
    ```
3. `:format`

I expected this to happen: the formatter should turn the 4 column indent into a 2 column indent.

Instead, this happened: no change to the document.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
[...]

PR #6512, opened 2023-03-31
robustly handle invalid LSP ranges

pascalkuthe:
Fix crashes with the TS language server reported on matrix

The Problem is that the way helix deals with invalid LSP ranges is different from other editors right now. Helix currently panics if the start of a LSP range is after the end of the range. The standard doesn't actually state that this is forbidden. VSCode simply sets `start=end` in that case and there seem to be a few servers relying on this sadly. I matched that behavior but logged an error to make sure this is not missed when tracking down issues. Alternatively we could discard such ranges. We definitely shouldn't panic tough.

Similarly, we now allow ranges with line numbers past the end (simply treated as EOF). VSCode behaves the same and so does neovim. I copied their comment here:

> If it extends past the end, truncate it to the end. This is because the
> way the LSP describes the range including the last newline is by
> specifying a line number after what we would call the last line.

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -129,7 +129,11 @@ pub mod util {
     ) -> Option<usize> {
         let pos_line = pos.line as usize;
         if pos_line > doc.len_lines() - 1 {
-            return None;
+            // If it extends past the end, truncate it to the end. This is because the
+            // way the LSP describes the range including the last newline is by
+            // specifying a line number after what we would call the last line.
+            log::warn!("LSP position {pos:?} out of range assuming EOF");
+            return Some(doc.len_chars());
         }
 
         // We need to be careful here to fully comply ith the LSP spec.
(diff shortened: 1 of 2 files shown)


helix-lsp/src/lib.rs
```diff
@@ -129,7 +129,11 @@ pub mod util {
     ) -> Option<usize> {
         let pos_line = pos.line as usize;
         if pos_line > doc.len_lines() - 1 {
-            return None;
+            // If it extends past the end, truncate it to the end. This is because the
+            // way the LSP describes the range including the last newline is by
+            // specifying a line number after what we would call the last line.
+            log::warn!("LSP position {pos:?} out of range assuming EOF");
+            return Some(doc.len_chars());
```
the-mikedavis, line 136: This is https://github.com/helix-editor/helix/issues/6022, right? And this change brings us in line with VSCode's handling of this edge-case?

pascalkuthe: exactly

the-mikedavis (dismissed):

helix-lsp/src/lib.rs
```diff
@@ -239,9 +243,20 @@ pub mod util {
 
     pub fn lsp_range_to_range(
         doc: &Rope,
-        range: lsp::Range,
+        mut range: lsp::Range,
         offset_encoding: OffsetEncoding,
     ) -> Option<Range> {
+        // This is sort of an edgecase. It's not clear from the spec how to deal with
+        // ranges where end < start. They don't make much sense but vscode simply caps start to end
+        // and because it's not specified quite a few LS rely on this as a result (for example the TS server)
```
archseer, line 251: (╯°□°)╯︵ ┻━┻

archseer (approved):

PR #6515, opened 2023-03-31
dracula theme: style inlay hints as comments

loewenheim:
This styles all inlay hints in the same color as comments. Additionally, types and parameters will be italic.

Bildschirm­foto 2023-03-31 um 21 48 30

I haven't yet found any colors I really like for styling types and parameters differently, though I'm open to suggestions. 

For reference, this is what it looks like using the same colors as ordinary types and parameters, respectively, but with an additional dim modifier:
Bildschirm­foto 2023-03-31 um 21 47 29


pickfire:
Thanks.

PR #6517, opened 2023-03-31
normalize LSP workspaces

### LINKED ISSUE #6505: rust-analyzer doesn't work most of the time on 23.03
### Summary

After updating to 23.03 today, I noticed that rust-analyzer doesn't work with helix most of the time, go-to definition, doc for item under cursor etc. rarely works, then you restart helix or do `:lsp-restart`  and it works or doesn't. This is not the case with 22.12 which I have used daily since it was released. I tested again with the 22.12 and it works as expected each time.

### Reproduction Steps

```
# src/main.rs

fn main() {
    let _ = foo(32);
}

pub fn foo(n: i32) -> i32 {
    n
}
```

1. `hx src/main.rs`
2. move to `foo`
3. \<space\>k

I expected this to happen:

popup with the relevant info
```
foo

pub fn foo(n: i32) -> i32
```
Instead, this happened:

Nothing

### Helix log

broken: [helix 23.03.log](https://github.com/helix-editor/helix/files/11124287/helix.23.03.log)
working [helix 22.12.log](https://github.com/helix-editor/helix/files/11124289/helix.22.12.log)

### Platform

Windows 10 22H2 (19045.2728)

### Terminal Emulator

Windows Terminal Version: 1.16.10261.0, wezterm 20221120-093910-0bf36e5a

### Helix Version

helix 23.03 (3cf03723)

pascalkuthe:
Fixes #6505

We didn't normalize/canoncialize LSP workspace paths with can both lead to bugs in the workspace detection and leads to inconsistent paths being sent to the LS which might not handle this correctly.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -4,7 +4,7 @@ use crate::{
     Call, Error, OffsetEncoding, Result,
 };
 
-use helix_core::{find_workspace, ChangeSet, Rope};
+use helix_core::{find_workspace, path, ChangeSet, Rope};
 use helix_loader::{self, VERSION_AND_GIT_HASH};
 use lsp::{
     notification::DidChangeWorkspaceFolders, DidChangeWorkspaceFoldersParams, OneOf,
@@ -66,6 +66,7 @@ impl Client {
         may_support_workspace: bool,
     ) -> bool {
         let (workspace, workspace_is_cwd) = find_workspace();
+        let workspace = path::get_normalized_path(&workspace);
         let root = find_lsp_workspace(
             doc_path
                 .and_then(|x| x.parent().and_then(|x| x.to_str()))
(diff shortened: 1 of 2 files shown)


helix-lsp/src/client.rs
```diff
@@ -201,6 +202,7 @@ impl Client {
         let (server_rx, server_tx, initialize_notify) =
             Transport::start(reader, writer, stderr, id);
         let (workspace, workspace_is_cwd) = find_workspace();
+        let workspace = path::get_normalized_path(&workspace);
```
the-mikedavis, line 205: What do you think about canonicalizing in `find_workspace` instead?

pascalkuthe: I wanted to do that at first but the problem with that is the `normalize` is in `helix-core` and `find_workspace` is in `helix-loader` (which can't pull in `helix-core`). All other uses of `find_workspace` don't care about path equality (they just use it to do IO where it doesn't matter which path is used)

the-mikedavis (approved):

PR #6531, opened 2023-04-01
Add commands for open languages.toml

erasin:
- `:lang-config-open`
- `:lang-config-workspace-open`


pascalkuthe:
I am going to be colsing this out as stale. Considering that we are making progress towards integrating scheme/a scheme based config (which would nit have a separate language config file). I don't think this is worth adding at this point.

Thank you for contributing!

PR #6547, opened 2023-04-01
update bass grammar and queries

vito:
Hiya, this is a follow-up to #3771 that bumps to the latest grammar (now rewritten by @amaanq :pray:), updates the highlight rules, and adds a few more queries:

* `folds.scm`
* `indents.scm`
* `injections.scm`
* `locals.scm`

Tested locally on my machine and all seems well.

### DIFF
--- a/runtime/queries/bass/highlights.scm
+++ b/runtime/queries/bass/highlights.scm
@@ -1,75 +1,58 @@
-; GENERATED VIA https://github.com/vito/tree-sitter-bass
+;; Includes
 
-;;; comments
+((symbol) @keyword.control.import
+  (#match? @keyword.control.import "^(use|import|load)$"))
 
-(comment) @comment.line
+;; Keywords
 
-;;; punctuation
+((symbol) @keyword
+  (#match? @keyword "^(do|doc)$"))
 
-["(" ")" "[" "]" "{" "}"] @punctuation.bracket
-
-;;; constants
-
-[(ignore) (null)] @constant.builtin
-
-(bool) @constant.builtin.boolean
-
-(int) @constant.numeric.integer
-
-;;; strings
-
-;; string literals
-
-(string) @string
-(string (string_escape) @constant.character.escape)
-
-;; keywords (symbol literals)
+; Keywords construct a symbol
 
 (keyword) @string.special.symbol
 
-;; paths
+;; Operators
 
-(dot) @string.special.path
-(dotdot) @string.special.path
-(command) @string.special.path
-(subpath (symbol) @string.special.path)
+; TODO: classify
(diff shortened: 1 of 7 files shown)


vito:
Marking as draft; noticed a minor thing to fix in the grammar (expanding allowed characters in paths), will reopen when it's ready.

runtime/queries/bass/highlights.scm
```diff
@@ -89,13 +72,60 @@
...
+;; Literals
+
+(string) @string
+
+(escape_sequence) @constant.character.escape
+
+(path) @string.special.path
+(command) @string.special.path
+
+(number) @constant.numeric.integer
```
the-mikedavis, line 116: Does `number` capture integers and floats? `@constant.numeric` would be more appropriate if it is both but the current capture is good if it's just integers

runtime/queries/bass/highlights.scm
```diff
@@ -89,13 +72,60 @@
...
+  (null)
+] @constant.builtin
+
+[
+  "^"
+] @character.special
+
+;; Comments
+
+(comment) @comment.line @spell
```
the-mikedavis, line 131:
```suggestion
(comment) @comment
```

runtime/queries/bass/highlights.scm
```diff
@@ -89,13 +72,60 @@
...
+(boolean) @constant.builtin.boolean
+
+[
+  (ignore)
+  (null)
+] @constant.builtin
+
+[
+  "^"
+] @character.special
```
the-mikedavis, line 127:
```suggestion
] @string.special
```

runtime/queries/bass/indents.scm
the-mikedavis, line 1: It looks like these are in the neovim format. We use different captures for indentation queries: https://docs.helix-editor.com/master/guides/indent.html

runtime/queries/bass/injections.scm
```diff
@@ -0,0 +1 @@
+(comment) @comment
```
the-mikedavis, line 1: Similar here about injections

```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
```

runtime/queries/bass/locals.scm
the-mikedavis, line 1: We follow tree-sitter for locals captures but these are in nvim's format. See https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

`@scope` should become `@local.scope` while `@reference` becomes `@local.reference` and the two `@definition...` captures become `@local.definition`

vito:
Closing this for now, will reopen if/when the Bass project regains momentum. Thanks for the reviews so far!

Issue #6553, opened 2023-04-02
Themes should all use _ or all use - in their names


sourproton:
There is this inconsistency where some themes use `_` in their names, like `ayu_dark`, while others use `-`, like `adwaita-dark`.

For consistency purposes and better user experience, I think they should all use one of the two options.

Possible solution: change all `-` to `_`.

pickfire:
> This is a breaking change for any configs that use these themes. I don't think the consistency is worth breaking all of those configs -- @the-mikedavis 

Or maybe we can do like docs.rs to automatically swap to the counter part if it isn't found.

tqwewe:
I wish this could be fixed, it's a pretty simple thing to change and the inconsistencies are a little annoying for me with OCD lol

PR #6561, opened 2023-04-02
Fix inconsistency in theme names

casperstorm:
Fixes: https://github.com/helix-editor/helix/issues/6553.

This is a small PR to enforce snake case and therefor avoid the inconsistency in theme names.


the-mikedavis:
This is a breaking change for any configs that use these themes. I don't think the consistency is worth breaking all of those configs

casperstorm:
> This is a breaking change for any configs that use these themes. I don't think the consistency is worth breaking all of those configs

That is fair. I am new to contributing to this project so I dont know how you like breaking changes 😄.
Feel free to close if don't feel like its worth it.

Issue #3451, opened 2022-08-15
Auto-save


onion108:
So, hmm, I found that it's hard to write Rust on Helix, because the `rust-analyzer` won't give me the informations about borrow checking, ownership and many other things unless I save my file. In VSCode I can enable the auto-save so it's OK for me, but in helix I can't find a way to enable auto-save. It's really inconvenient to type `<ESC> :w <Return> i` after typing each line of the code. I think it's necessary for `helix-editor` to offer an option to enable/disable auto-save. Thanks a lot.

Omnikar:
In the meantime, you could bind saving the file to a keybind while in insert mode. For example, you can use `Ctrl-s` to save while in insert mode using this binding in your `config.toml`:
```toml
[keys.insert]
"C-s" = ":w"
```

AceofSpades5757:
What would you want to trigger the auto-save? Some amount of idle time?

willparsons:
iirc rust-analyzer plans to add on the fly diagnostics (no idea when)

archseer:
https://github.com/helix-editor/helix/pull/3178

onion108:
> In the meantime, you could bind saving the file to a keybind while in insert mode. For example, you can use `Ctrl-s` to save while in insert mode using this binding in your `config.toml` (accessible via `:config-open`):
> 
> ```toml
> [keys.insert]
> "C-s" = ":w"
> ```

well, it looks not working for me

onion108:
> What would you want to trigger the auto-save? Some amount of idle time?

just like what vscode does

archseer:
Some of us never used vscode so you'll have to elaborate.

kirawi:
https://code.visualstudio.com/docs/editor/codebasics#_save-auto-save

Looks like there are different auto save modes.

onion108:
> https://code.visualstudio.com/docs/editor/codebasics#_save-auto-save
> 
> Looks like there are different auto save modes.

`afterDelay`, i mean

TobTobXX:
> What would you want to trigger the auto-save? Some amount of idle time?

Something that worked well for me in nivm is on reentering normal mode. This means that it gets saved after basically every change.

AfoHT:
In addition to what @TobTobXX suggested, which also was part of how I had nvim setup, I propose that also changing to a different split or buffer might be a good time to save the buffer that was moved away from.

I guess that maps closely to the vscode `onFocusChange`

sourcefrog:
[...]

PR #6563, opened 2023-04-02
Add entering normal mode as autosave trigger

TobTobXX:
For me, autosave is an essential feature. But the 'unfocus' trigger doesn't work, since autosave also triggrers reformatting (which I think is useful), but causes a massive loss of context when triggered on unfocus. (I'd switch workspaces to copy some code and bam! autoformat moved my cursor around when I look back...)

However, when exiting edit mode, it's quite natural, as I've finished doing my edit, I actually want the formatter to clean it up. It is also a rather logical time to autosave, since I've just made an edit.

### TODOs
- [ ] Documentation: Wasn't there a command to generate the book documentations?
- [ ] Don't break compatibility: https://github.com/helix-editor/helix/pull/3178#discussion_r998702595 (how do I do that?(
- [ ] Update modification indicator in statusline after autosave.

### See also:
- Discussion: #3451 
- Conflicts with #5303 (Rebasing is easy, just inform me whether this or the other PR should be the lower commit)

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -23,7 +23,7 @@ use helix_core::{
 };
 use helix_view::{
     document::{Mode, SavePoint, SCRATCH_BUFFER_NAME},
-    editor::{CompleteAction, CursorShapeConfig},
+    editor::{CompleteAction, CursorShapeConfig, AutoSaveTrigger},
     graphics::{Color, CursorKind, Modifier, Rect, Style},
     input::{KeyEvent, MouseButton, MouseEvent, MouseEventKind},
     keyboard::{KeyCode, KeyModifiers},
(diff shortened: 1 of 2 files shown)


TobTobXX:
Note, that the last TODO (updating statusline) doesn't work with the unfocus trigger either. *shrug

TobTobXX:
Ugh, thanks to [this](https://github.com/helix-editor/helix/pull/3178#discussion_r998702595) stupid comment the config is now quite a bit uglier and more indirect.

btw, can we ignore clippy on `helix-view/src/editor.rs:1699:9`? That hint, while technically correct, is really NOT helping readability.

archseer:
Do we really need another config option? We could just trigger on both

\cc @groves

pascalkuthe:
I am not sure this even needs to be build on. I have just bound `esc` to `[normal_mode, :w<ret>]` works well enough for saving on normal.mode exit. It doesn't quite catch clal instances of exiting normal mode but it's good enough for the most part.

I think in the long run we would want something more flexible to save (after a decently long denounced timeout) after every edit so doing edits in normal.mode (replace, paste, rename) also triggers autosave. Until such a solution is.implemented I think the keybindkng above is a good enough workaround

TobTobXX:
Ok, I'll yield. Your point is valid.

PR #6564, opened 2023-04-03
match against unix file path [windows] (#5414)

### LINKED ISSUE #5414: Support forward slash '/' in fuzzy-finder etc. on Windows
Hi! I tried to look through the issues but didn't see this one mentioned.

Basically, on Windows when using the fuzzy-finder or opening a new file etc., then one must use backslashes. It would be a great convenience if it worked with forwardslash as well (Windows supports this).

Thanks!

JackMordaunt:
This commit introduces a change to the Atom, whereby '/' are replace with '\\' on Windows. This allows the user to type in unix style paths and have it successfully match against Windows file paths.

Fixes #5414

https://user-images.githubusercontent.com/6916093/229416948-6b33b395-5e76-475c-9692-a38a22973e25.mp4

### DIFF
--- a/helix-term/src/ui/fuzzy_match.rs
+++ b/helix-term/src/ui/fuzzy_match.rs
@@ -42,13 +42,21 @@ impl QueryAtom {
             }
         }
 
+        atom = atom.replace('\\', "");
+
+        if std::env::consts::OS == "windows" {
+            atom = atom.replace('/', "\\");
+        }
+
+        // not ideal but fuzzy_matches only knows ascii uppercase so more consistent
+        // to behave the same
+        let ignore_case =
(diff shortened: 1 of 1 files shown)


pascalkuthe (changes requested):
We use the fuzzy matching infrastructure for lits of different pickers. Not all of which are paths sk this should be something specific to global seadch/file picker (where we only match paths) and not added inside the core fuzzy matching code

pppKin:
Also for consistency might as well display the paths with forward slashes in file picker/search result picker.

pascalkuthe:
going to close this one out as the picker has been rewritten since this one was made. I also don't see us going forward with this approach. Thank you fro contributing!

PR #6572, opened 2023-04-03
Change smithy language server to the one from AWS

zetashift:
Changes the current smithy language server to use the one from: https://github.com/awslabs/smithy-language-server
I've used both just fine(AWS one with VSCode and Disneys with helix) but I believe the AWS one is a bit nicer for general usage.

I'll update the wiki LS install instructions if this PR gets approved.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2364,7 +2364,7 @@ file-types = ["smithy"]
 roots = ["smithy-build.json"]
 comment-token = "//"
 indent = { tab-width = 4, unit = "    " }
-language-server = { command = "cs", args = ["launch", "com.disneystreaming.smithy:smithy-language-server:latest.release", "--", "0"] }
+language-server = { command = "cs", args = ["launch", "--contrib", "smithy-language-server", "--", "0"] }
 
 [[grammar]]
 name = "smithy"


zetashift:
I am hitting a problem where the "go-to definition" selects more than it should:

- Press go to definition on a shape ( in my case `GetCustomerInput` )
- It'll select a few more lines than it should
image

(notice how `@output` is also selected, even tho it's not part of that shape).

Not entirely sure if this is a LS problem, I can also reproduce this on Disney's LSP

mtdowling (approved):
I think any generally applicable changes from the Disney fork are upstreamed, so this LGTM. Since the goto definition issue is in both LSPs, I still think we should switch, but opening an issue on https://github.com/awslabs/smithy-language-server would be great!

Updating the installation instructions would be great too, but I know you're waiting on agreement for the PR.

zetashift:
> I think any generally applicable changes from the Disney fork are upstreamed, so this LGTM. Since the goto definition issue is in both LSPs, I still think we should switch, but opening an issue on https://github.com/awslabs/smithy-language-server would be great!
> 
> Updating the installation instructions would be great too, but I know you're waiting on agreement for the PR.

Ah yes, sounds good. I'll open up an issue is that repo, thank you for your help :D.

pascalkuthe (approved):
LGTM

the-mikedavis (approved):

zetashift:
Thank you for merging! :D

I updated the wiki to clarify that we're using the one from AWS: https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers#smithy
Installation instructions remain unchanged because coursier is still being used.

PR #6587, opened 2023-04-04
Update Ruby Highlights

danillos:
- Updates Ruby `highlights.scm` with many improvements
- Updates `tree-sitter-ruby` to the last commit
- Add SQL to `injections.scm`

_`tree-sitter-ruby` update was required to be able to create highlights for some new Ruby 3 features like the new Forward Paramenter `...`_

Some screenshots:

CleanShot 2023-04-04 at 10 22 04@2x
CleanShot 2023-04-04 at 10 23 51@2x
CleanShot 2023-04-04 at 10 26 14@2x
CleanShot 2023-04-04 at 11 15 53@2x


pickfire (approved):
Thanks.

PR #6588, opened 2023-04-04
Add textobject queries for Julia

Ratfink:
This PR finishes filling out the language support for Julia by adding textobject queries.


pickfire (approved):
Thanks.

Issue #6582, opened 2023-04-04
Support for more modifiers with progressive keyboard enhancement protocol


eugenesvk:
Now that Helix has progressive keyboard enhancement protocol implemented, and that protocol supports extra modifiers beyond Control/Shift/Alt, would be great if Helix also allowed using those extra modifiers in key bindings

the-mikedavis:
https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers

IMO the only other one worth supporting is Super. It's a fairly small change to support this, I'll post a PR

eugenesvk:
Thank you
Some create an artificial 'hyper' key by using simultaneous 4 right-side modifiers. Am I correct that with the added support for Super this would also work in this protocol now that all 4 modifiers (although there is no difference between right and left at the moment per https://github.com/helix-editor/helix/discussions/6479)?

(I understand that the Hyper/Meta in the spec are separate keys)

Issue #2913, opened 2022-06-29
add info about binding to mac specific modifier keys


adsick:
[documentation](https://docs.helix-editor.com/remapping.html)
describes modifier keys in terms of classic "Control, Alt, Shift" trinity, but mac has "Command, Option, Shift" (and Control which I don't want to use), how to deal with it? maybe add a note about it in the documentation?

I want to have Command-x bindings, not Control-x.

the-mikedavis:
Command modifiers are not supported because they're not supported upstream in crossterm: https://github.com/crossterm-rs/crossterm/issues/389

Support for command could most likely be added with a GUI but otherwise awaits that issue upstream.

groves:
Command can be exposed with [Kitty's keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers). That's started being added in https://github.com/crossterm-rs/crossterm/pull/691. Once that lands, I think we could build on it for additional modifiers.

restfuladi:
When using alacritty, `command+s` shows `select:`. I'd love to know how it's doing that so I can remap it to `:write`.

kee-oth:
> Command can be exposed with [Kitty's keyboard protocol](https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers). That's started being added in [crossterm-rs/crossterm#691](https://github.com/crossterm-rs/crossterm/pull/691). Once that lands, I think we could build on it for additional modifiers.

@groves Just pinging that the linked PR has been merged as I am interested in this as well. Thanks!

peredwardsson:
Also interested. Editor looks great, but using Mac + nordic keyboard makes it a little awkward.

justaskz:
Are there any updates?

I tried to remap keys using kitty:
When `Cmd+KEY` is pressed, kitty sends `Ctrl+KEY` hex code as text. However it does not work. And there is no way to debug it.

Does anyone have any success in making Command key working?

danillos:
@justaskz one example with Kitty that I did to save with `cmd+s`

```
# Kitty config

map cmd+S send_text all \x13
```

```
# Helix Config

[keys.normal]
'C-s' = ':w!'
```


`\x13` = `control+s` you can see [more here](https://donsnotes.com/tech/charsets/ascii.html) 
Kitty accepts [unicode](https://www.rapidtables.com/code/text/unicode-characters.html) too.

the-mikedavis:
[...]

PR #6592, opened 2023-04-04
Support bindings with the Super (Meta/Cmd/Win) modifier

the-mikedavis:
Terminals which support the enhanced keyboard protocol send events for keys pressed with the Super modifier (Windows/Linux key or the Command key). We can support keybindings with the Super modifier in Helix with a few small changes: mapping the event from crossterm and parsing/representing the modifier in the KeyEvent string representation.

The binding can be done with `Meta-`, `Win-` or `Cmd-` prefixes - these are all synonymous. For example `keys.normal.Cmd-s = ":write"`.

Closes https://github.com/helix-editor/helix/issues/2913
Closes https://github.com/helix-editor/helix/issues/6582

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -162,7 +162,12 @@ pub(crate) mod keys {
 impl fmt::Display for KeyEvent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
-            "{}{}{}",
+            "{}{}{}{}",
+            if self.modifiers.contains(KeyModifiers::SUPER) {
+                "Meta-"
+            } else {
+                ""
+            },
             if self.modifiers.contains(KeyModifiers::SHIFT) {
                 "S-"
             } else {
@@ -312,6 +317,10 @@ impl UnicodeWidthStr for KeyEvent {
         if self.modifiers.contains(KeyModifiers::CONTROL) {
             width += 2;
         }
+        if self.modifiers.contains(KeyModifiers::SUPER) {
+            // "-Meta"
+            width += 5;
+        }
         width
     }
 
@@ -396,6 +405,7 @@ impl std::str::FromStr for KeyEvent {
                 "S" => KeyModifiers::SHIFT,
                 "A" => KeyModifiers::ALT,
                 "C" => KeyModifiers::CONTROL,
+                "Meta" | "Cmd" | "Win" => KeyModifiers::SUPER,
                 _ => return Err(anyhow!("Invalid key modifier '{}-'", token)),
             };
 
@@ -707,6 +717,28 @@ mod test {
             KeyEvent {
                 code: KeyCode::Char('W'),
                 modifiers: KeyModifiers::NONE
+            }
+        );
+
+        assert_eq!(
+            str::parse::<KeyEvent>("Meta-c").unwrap(),
+            KeyEvent {
+                code: KeyCode::Char('c'),
+                modifiers: KeyModifiers::SUPER
+            }
+        );
+        assert_eq!(
+            str::parse::<KeyEvent>("Win-s").unwrap(),
+            KeyEvent {
+                code: KeyCode::Char('s'),
+                modifiers: KeyModifiers::SUPER
+            }
+        );
+        assert_eq!(
+            str::parse::<KeyEvent>("Cmd-d").unwrap(),
+            KeyEvent {
+                code: KeyCode::Char('d'),
+                modifiers: KeyModifiers::SUPER
             }
         );
     }
--- a/book/src/remapping.md
+++ b/book/src/remapping.md
@@ -50,8 +50,22 @@ t = ":run-shell-command cargo test"
 
 ## Special keys and modifiers
 
-Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes
-`C-`, `S-` and `A-`. Special keys are encoded as follows:
+Ctrl, Shift and Alt modifiers are encoded respectively with the prefixes `C-`, `S-` and `A-`.
+
+The [Super key](https://en.wikipedia.org/wiki/Super_key_(keyboard_button)) - the Windows/Linux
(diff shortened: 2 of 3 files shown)


helix-view/src/input.rs
```diff
@@ -158,7 +158,12 @@ pub(crate) mod keys {
 impl fmt::Display for KeyEvent {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> std::fmt::Result {
         f.write_fmt(format_args!(
-            "{}{}{}",
+            "{}{}{}{}",
+            if self.modifiers.contains(KeyModifiers::SUPER) {
+                "Super-"
```
the-mikedavis, line 163: We could use another name/letter for this in the representation. Either [Meta](https://en.wikipedia.org/wiki/Meta_key) or Hyper would probably be close enough but would be incorrect. Those two are also possible to receive as distinct modifiers although it's unlikely (https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers)

eugenesvk: maybe `W-` as it's a single letter abbreviation that's closer to Super (Win-key) and also correct unlike Meta/Hyper?

eugenesvk: Well, `Super` is incorrect everywhere as it has a different name on each OS and given that the first letters aren't unique, first-letter OS-agnostic-but-correct abbreviations aren't an option (or maybe use two-letter one? `CW-` cmd-win)

the-mikedavis: It seems to be a common choice for free software, apparently because of trademark issues: https://en.wikipedia.org/wiki/Super_key_(keyboard_button)#Usage

`OS-` is could be a good choice for this (https://support.system76.com/articles/pop-keyboard-shortcuts/). QMK calls it GUI but I think that's a little confusing (https://docs.qmk.fm/#/keycodes_basic?id=modifiers). I don't like `CW-`: it isn't used anywhere else as far as I know.

jpsecher: So how do we reach a decision on this?  I think it would be wise to avoid inventing yet another name for the key, and if we want to avoid something that starts with `s`, then the choices seem to be

1. `win`
2. `windows`
3. `cmd`
4. `command`
5. `gui`
6. `meta`
7. `os`

realeinherjar: I like `OS-`

koalalorenzo:
Any plans for this PR? :)

helix-view/src/input.rs
matta: I did some digging on this, probably too much.

The world already uses inconsistent terms for the Windows/Command key in cross-platform contexts, but it seems to *consistently* pick between either "Meta" or "Super".

I don't think it would be good to invent a new name for the key. Helix should pick "Meta" or "Super". I'd pick "Meta" but either choice is reasonable.

From what I can suss out, "Meta" is the most widely used. It is what web standards call it: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey, and it is probably what most people will readily understand. I say "most people" because I'm guessing web developers may outnumber basically all other kinds of developers, and web apps outnumber basically all other kinds of apps (at least those with any kind of broad reach), so web standards win any familiarity context. E.g. VS Code is the most popular "free" editor in the world, probably the most popular editor full stop, and it calls it "Meta," because it is basically a browser app.

(Interestingly, I fired up RustRover, and it doesn't support binding the Windows key, at least on my Gnome box.)

Calling it "Super" would be consistent with X11, Gnome, and the Kitty protocol at https://sw.kovidgoyal.net/kitty/keyboard-protocol/#modifiers.

Calling it "Meta" would be consistent with the web, and KDE, and the Linux kernel, but may confuse die hard GNU Readline and Emacs users, and other X11-first software.

I'd pick "Meta" because the influence of web standards on developer mind share is hard to ignore.

eugenesvk: > and web apps outnumber basically all other kinds of apps (at least those with any kind of broad reach), so web standards win any familiarity context. E.g. VS Code is the most popular "free" editor in the world, probably the most popular editor full stop, and it calls it "Meta," because it is basically a browser app.

But broader user base outnumbers any specific tech-stack user base by a wide margin, so using an obscure tech-stack-based name can't be a sensible alternative to using a widely known name

And your VSCode example confirms this since VSCode doesn't use Meta on the platforms where it's unknown (Mac and Windows) in it's user-facing config (even though it does use the web-standards Meta name in debugigng output): 

> [Accepted keys](https://code.visualstudio.com/docs/getstarted/keybindings#_accepted-keys)
The key is made up of modifiers and the key itself.

>  The following modifiers are accepted:

>  Platform	Modifiers
macOS	Ctrl+, Shift+, Alt+, Cmd+
Windows	Ctrl+, Shift+, Alt+, Win+
Linux	Ctrl+, Shift+, Alt+, Meta+

By the way, why not allow any value from `Cmd`/`Win`/`Meta` and simply treat either config value as `Cmd` on a Mac,`Win` on Windows, and the Linux key alternative on Linux

the-mikedavis: I like the idea of Cmd+Win+Meta synonyms! Let's go with that

mighabana:
Hi is there any way I can help to get this merged? I'd be happy to help make any code changes or have any design descussions if that's what's blocking the PR. 😄

the-mikedavis:
It just needs a round of review from the other maintainers, codewise it should  be fine

pascalkuthe (approved):

omentic:
I have no stake in this PR, but: why all of `Meta-`, `Cmd-`, `Win-`?

It seems simpler to pick just one (or to pick `M-`, on analogy with `S-`, `A-`, `C-`)

the-mikedavis:
See the collapsed thread above - it started as "Super-" but there isn't an ideal choice for what to call the modifier

NewtonChutney:
Hello!
Have any links pointing to terminals supporting `enhanced keyboard protocol` been added in the docs?

Issue #5336, opened 2022-12-29
Allow multiple language servers for a language


arendjr:
I am not sure if what I’m asking for is already possible, but in order to configure the Rome language server (https://rome.tools) for TypeScript correctly, I think Helix needs to provide a way to configure multiple language servers to be used for a single language.

Rome actually does provide instructions on how to configure their language server with Helix: https://github.com/rome/tools/blob/main/editors/helix/manual.md

But after following those instructions, standard TypeScript functionality such as “goto definition” doesn’t work anymore, presumably because Helix doesn’t understand it should *also* still use the regular TS language server.

Related issue as reported against Rome: https://github.com/rome/tools/issues/4107#issuecomment-1367312352

archseer:
https://github.com/helix-editor/helix/pull/2507

arendjr:
Just my luck, thanks! :D

PR #6594, opened 2023-04-04
correctly handle completion rerequest

### LINKED ISSUE #6259: Snippets insert literal text

### LINKED ISSUE #6261: Svelte duplicated auto-imports on tab

pascalkuthe:
Alternative to #6511
Closes #6259
Closes #6261
Fixes a similar bug with the typescript LSP reported on matrix where imports would be duplicated. 

This PR resolved the same issue as #6511 but keeps the current behavior of previewing the current completion. This is achieved by generating "ghost" transactions. Ghost transactions are normal transactions that are never sent to the LS.

Maintaining such transactions is tricky/dangerous because we always need to undo them (with the undo transaction also not sent to the LS) before we apply any other transaction to the document. Otherwise, we would send invalid changes to LS with incremental sync and therefore break the server (until `:lsp-restart`).

To that end I used the more generic savepoints introduced in #6173. The completion menu now saves a second savepoint before a completion option is first selected with tab/s-tab. The completion preview (so what you see when hitting tab but before pressing `<ret>`) is now such a ghost transaction. To ensure these transactions are always removed there have been some behavioral changes:

* both c-c and esc now reset the document to the savepoint created before previewing a transaction (or do nothing otherwise)
* As soon as a completion is selected with "tab" (previewed) it becomes "active" as soon as any key not captured by the completion menu is pressed the completion is fully applied (same as pressing enter) automatically. So there are no "half applied" completions anymore

This behavior matches Kakoune (which also automatically applies the transaction when hitting enter as discovered by @the-mikedavis). 

This approach is safe regarding the hazards mentioned above because as soon as a ghost transaction is applied it will always be realized when pressing any key (that isn't consumed by the completion menu). The only remaining hazard are "async" edits (or savepoints like completions request) that are created without an explicit key press. These might pickup the ghost document state. I think there aren't really any of those (besides completions themselves where I habe taken great care to cover all edge-cases) because any async transaction would require a savepoint (otherwise the transaction would refer to a potentially outdated document state) and the only use of savepoint is completions/completion replay (which always require a keypress). If we ever want to add other async transaction we might think about generalizing the system a bit but that will require some more complex design. For now I simply expanded the existing `Editor::last_complete` field to cover the usecase here.

The type script import problem was actually a special case of the more general problem fixed by this PR. It was caused by resolving completions *after they were already applied*. The general problem was the same (we send didChange and then rerequested a completion) but fixing it furthermore required moving the request in a3a4645. It seems other clients are running into this issue too (for example nvim) so getting this right is difficult but vscode seems to handle this correctly (also using some ghost text it seems). See https://github.com/typescript-language-server/typescript-language-server/issues/699 (the examples there also cause bugs on master but work with this PR or #6511).

#6261 was caused by us resolving a completion multiple times. The LSP standard doesn't explicitly state it but the intention seems to be that completion items are only resolved once (since resolving is intended to delay expensive calculations). Helix now separately tracks whether a completion item has been resolved and therefore mirrors vscode in that regard. I think that this was really always the intention of the helix code (continuously querying the server is quite inefficient) and the check we did just wasn't quite correct

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4169,16 +4169,23 @@ pub fn completion(cx: &mut Context) {
 
     let (view, doc) = current!(cx.editor);
 
+    let savepoint = if let Some(CompleteAction::Selected { savepoint }) = &cx.editor.last_completion
+    {
+        savepoint.clone()
+    } else {
+        doc.savepoint(view)
+    };
+
     let language_server = match doc.language_server() {
         Some(language_server) => language_server,
         None => return,
     };
 
     let offset_encoding = language_server.offset_encoding();
-    let text = doc.text().slice(..);
-    let cursor = doc.selection(view.id).primary().cursor(text);
+    let text = savepoint.text.clone();
+    let cursor = savepoint.cursor();
 
-    let pos = pos_to_lsp_pos(doc.text(), cursor, offset_encoding);
+    let pos = pos_to_lsp_pos(&text, cursor, offset_encoding);
 
     let future = match language_server.completion(doc.identifier(), pos, None) {
         Some(future) => future,
@@ -4213,12 +4220,6 @@ pub fn completion(cx: &mut Context) {
     iter.reverse();
     let offset = iter.take_while(|ch| chars::char_is_word(*ch)).count();
     let start_offset = cursor.saturating_sub(offset);
-    let savepoint = if let Some(CompleteAction::Selected { savepoint }) = &cx.editor.last_completion
-    {
-        savepoint.clone()
-    } else {
-        doc.savepoint(view)
-    };
 
     let trigger_doc = doc.id();
     let trigger_view = view.id;
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -33,7 +33,7 @@ use helix_core::{
 use helix_view::{
     clipboard::ClipboardType,
     document::{FormatterError, Mode, SCRATCH_BUFFER_NAME},
-    editor::{Action, Motion},
+    editor::{Action, CompleteAction, Motion},
     info::Info,
     input::KeyEvent,
     keyboard::KeyCode,
@@ -4213,7 +4213,12 @@ pub fn completion(cx: &mut Context) {
     iter.reverse();
     let offset = iter.take_while(|ch| chars::char_is_word(*ch)).count();
     let start_offset = cursor.saturating_sub(offset);
-    let savepoint = doc.savepoint(view);
+    let savepoint = if let Some(CompleteAction::Selected { savepoint }) = &cx.editor.last_completion
+    {
+        savepoint.clone()
+    } else {
+        doc.savepoint(view)
+    };
 
     let trigger_doc = doc.id();
     let trigger_view = view.id;
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -209,14 +209,27 @@ impl Completion {
 
             let (view, doc) = current!(editor);
 
-            // if more text was entered, remove it
-            doc.restore(view, &savepoint);
-
             match event {
-                PromptEvent::Abort => {
-                    editor.last_completion = None;
-                }
+                PromptEvent::Abort => {}
                 PromptEvent::Update => {
+                    // Update creates "ghost" transactiosn which are not send to the
+                    // lsp server to avoid messing up rerequesting completions. Once a
+                    // completion has been selected (with) tab it's always accepted whenever anything
+                    // is typed. The only way to avoid that is to explicitly abort the completion
+                    // with esc/c-c. This will remove the "ghost" transaction.
+                    //
+                    // The ghost transaction is modeled with a transaction that is not send to the LS.
+                    // (apply_temporary) and a savepoint. It's extremly important this savepoint is restored
+                    // (also without sending the transaction to the LS) *before any further transaction is applied*.
+                    // Otherwise incremental sync breaks (since the state of the LS doesn't match the state the transaction
+                    // is applied to).
+                    if editor.last_completion.is_none() {
+                        editor.last_completion = Some(CompleteAction::Selected {
+                            savepoint: doc.savepoint(view),
+                        })
+                    }
+                    // if more text was entered, remove it
+                    doc.restore(view, &savepoint, false);
                     // always present here
                     let item = item.unwrap();
 
@@ -229,19 +242,20 @@ impl Completion {
                         true,
                         replace_mode,
                     );
-
-                    // initialize a savepoint
-                    doc.apply(&transaction, view.id);
-
-                    editor.last_completion = Some(CompleteAction {
-                        trigger_offset,
-                        changes: completion_changes(&transaction, trigger_offset),
-                    });
+                    doc.apply_temporary(&transaction, view.id);
                 }
                 PromptEvent::Validate => {
+                    if let Some(CompleteAction::Selected { savepoint }) =
+                        editor.last_completion.take()
+                    {
+                        doc.restore(view, &savepoint, false);
+                    }
                     // always present here
                     let item = item.unwrap();
 
+
+                    // if more text was entered, remove it
+                    doc.restore(view, &savepoint, true);
                     let transaction = item_to_transaction(
                         doc,
                         view.id,
@@ -251,10 +265,9 @@ impl Completion {
                         false,
                         replace_mode,
                     );
-
                     doc.apply(&transaction, view.id);
 
-                    editor.last_completion = Some(CompleteAction {
+                    editor.last_completion = Some(CompleteAction::Applied {
                         trigger_offset,
                         changes: completion_changes(&transaction, trigger_offset),
                     });
@@ -270,7 +283,6 @@ impl Completion {
                     } else {
                         Self::resolve_completion_item(doc, item.clone())
                     };
-
                     if let Some(additional_edits) = resolved_item
                         .as_ref()
                         .and_then(|item| item.additional_text_edits.as_ref())
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -226,14 +226,14 @@ impl Completion {
             match event {
                 PromptEvent::Abort => {}
                 PromptEvent::Update => {
-                    // Update creates "ghost" transactiosn which are not send to the
-                    // lsp server to avoid messing up rerequesting completions. Once a
-                    // completion has been selected (with) tab it's always accepted whenever anything
+                    // Update creates "ghost" transactions which are not sent to the
+                    // lsp server to avoid messing up re-requesting completions. Once a
+                    // completion has been selected (with tab, c-n or c-p) it's always accepted whenever anything
                     // is typed. The only way to avoid that is to explicitly abort the completion
-                    // with esc/c-c. This will remove the "ghost" transaction.
+                    // with c-c. This will remove the "ghost" transaction.
                     //
-                    // The ghost transaction is modeled with a transaction that is not send to the LS.
-                    // (apply_temporary) and a savepoint. It's extremly important this savepoint is restored
+                    // The ghost transaction is modeled with a transaction that is not sent to the LS.
+                    // (apply_temporary) and a savepoint. It's extremely important this savepoint is restored
                     // (also without sending the transaction to the LS) *before any further transaction is applied*.
                     // Otherwise incremental sync breaks (since the state of the LS doesn't match the state the transaction
                     // is applied to).
@@ -293,7 +293,7 @@ impl Completion {
                         changes: completion_changes(&transaction, trigger_offset),
                     });
 
-                    // TOOD: add additional _edits to completion_changes?
+                    // TODO: add additional _edits to completion_changes?
                     if let Some(additional_edits) = item.item.additional_text_edits {
                         if !additional_edits.is_empty() {
                             let transaction = util::generate_transaction_from_edits(
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -16,7 +16,6 @@ use crate::commands;
 use crate::ui::{menu, Markdown, Menu, Popup, PromptEvent};
 
 use helix_lsp::{lsp, util};
-use lsp::CompletionItem;
 
 impl menu::Item for CompletionItem {
     type Data = ();
@@ -26,28 +25,29 @@ impl menu::Item for CompletionItem {
 
     #[inline]
     fn filter_text(&self, _data: &Self::Data) -> Cow<str> {
-        self.filter_text
+        self.item
+            .filter_text
             .as_ref()
-            .unwrap_or(&self.label)
+            .unwrap_or(&self.item.label)
             .as_str()
             .into()
     }
 
     fn format(&self, _data: &Self::Data) -> menu::Row {
-        let deprecated = self.deprecated.unwrap_or_default()
-            || self.tags.as_ref().map_or(false, |tags| {
+        let deprecated = self.item.deprecated.unwrap_or_default()
+            || self.item.tags.as_ref().map_or(false, |tags| {
                 tags.contains(&lsp::CompletionItemTag::DEPRECATED)
             });
         menu::Row::new(vec![
             menu::Cell::from(Span::styled(
-                self.label.as_str(),
+                self.item.label.as_str(),
                 if deprecated {
                     Style::default().add_modifier(Modifier::CROSSED_OUT)
                 } else {
                     Style::default()
                 },
             )),
-            menu::Cell::from(match self.kind {
+            menu::Cell::from(match self.item.kind {
                 Some(lsp::CompletionItemKind::TEXT) => "text",
                 Some(lsp::CompletionItemKind::METHOD) => "method",
                 Some(lsp::CompletionItemKind::FUNCTION) => "function",
@@ -88,6 +88,12 @@ impl menu::Item for CompletionItem {
     }
 }
 
+#[derive(Debug, PartialEq, Default, Clone)]
+struct CompletionItem {
+    item: lsp::CompletionItem,
+    resolved: bool,
+}
+
 /// Wraps a Menu.
 pub struct Completion {
     popup: Popup<Menu<CompletionItem>>,
@@ -103,7 +109,7 @@ impl Completion {
     pub fn new(
         editor: &Editor,
         savepoint: Arc<SavePoint>,
-        mut items: Vec<CompletionItem>,
+        mut items: Vec<lsp::CompletionItem>,
         offset_encoding: helix_lsp::OffsetEncoding,
         start_offset: usize,
         trigger_offset: usize,
@@ -111,6 +117,13 @@ impl Completion {
         let replace_mode = editor.config().completion_replace;
         // Sort completion items according to their preselect status (given by the LSP server)
         items.sort_by_key(|item| !item.preselect.unwrap_or(false));
+        let items = items
+            .into_iter()
+            .map(|item| CompletionItem {
+                item,
+                resolved: false,
+            })
+            .collect();
 
         // Then create the menu
         let menu = Menu::new(items, (), move |editor: &mut Editor, item, event| {
@@ -128,7 +141,7 @@ impl Completion {
                 let text = doc.text().slice(..);
                 let primary_cursor = selection.primary().cursor(text);
 
-                let (edit_offset, new_text) = if let Some(edit) = &item.text_edit {
+                let (edit_offset, new_text) = if let Some(edit) = &item.item.text_edit {
                     let edit = match edit {
                         lsp::CompletionTextEdit::Edit(edit) => edit.clone(),
                         lsp::CompletionTextEdit::InsertAndReplace(item) => {
@@ -151,9 +164,10 @@ impl Completion {
                     (Some((start_offset, end_offset)), edit.new_text)
                 } else {
                     let new_text = item
+                        .item
                         .insert_text
                         .clone()
-                        .unwrap_or_else(|| item.label.clone());
+                        .unwrap_or_else(|| item.item.label.clone());
                     // check that we are still at the correct savepoint
                     // we can still generate a transaction regardless but if the
                     // document changed (and not just the selection) then we will
@@ -162,9 +176,9 @@ impl Completion {
                     (None, new_text)
                 };
 
-                if matches!(item.kind, Some(lsp::CompletionItemKind::SNIPPET))
+                if matches!(item.item.kind, Some(lsp::CompletionItemKind::SNIPPET))
                     || matches!(
-                        item.insert_text_format,
+                        item.item.insert_text_format,
                         Some(lsp::InsertTextFormat::SNIPPET)
                     )
                 {
@@ -251,26 +265,22 @@ impl Completion {
                         doc.restore(view, &savepoint, false);
                     }
                     // always present here
-                    let item = item.unwrap();
+                    let mut item = item.unwrap().clone();
 
-                    // apply additional edits, mostly used to auto import unqualified types
-                    let resolved_item = if item
-                        .additional_text_edits
-                        .as_ref()
-                        .map(|edits| !edits.is_empty())
-                        .unwrap_or(false)
-                    {
-                        None
-                    } else {
-                        Self::resolve_completion_item(doc, item.clone())
+                    // resolve item if not yet resolved
+                    if !item.resolved {
+                        if let Some(resolved) =
+                            Self::resolve_completion_item(doc, item.item.clone())
+                        {
+                            item.item = resolved;
+                        }
                     };
-
                     // if more text was entered, remove it
                     doc.restore(view, &savepoint, true);
                     let transaction = item_to_transaction(
                         doc,
                         view.id,
-                        item,
+                        &item,
                         offset_encoding,
                         trigger_offset,
                         false,
@@ -283,15 +293,12 @@ impl Completion {
                         changes: completion_changes(&transaction, trigger_offset),
                     });
 
-                    if let Some(additional_edits) = resolved_item
-                        .as_ref()
-                        .and_then(|item| item.additional_text_edits.as_ref())
-                        .or(item.additional_text_edits.as_ref())
-                    {
+                    // TOOD: add additional _edits to completion_changes?
+                    if let Some(additional_edits) = item.item.additional_text_edits {
                         if !additional_edits.is_empty() {
                             let transaction = util::generate_transaction_from_edits(
                                 doc.text(),
-                                additional_edits.clone(),
+                                additional_edits,
                                 offset_encoding, // TODO: should probably transcode in Client
                             );
                             doc.apply(&transaction, view.id);
@@ -318,7 +325,7 @@ impl Completion {
     fn resolve_completion_item(
         doc: &Document,
         completion_item: lsp::CompletionItem,
-    ) -> Option<CompletionItem> {
+    ) -> Option<lsp::CompletionItem> {
         let language_server = doc.language_server()?;
 
         let future = language_server.resolve_completion_item(completion_item)?;
@@ -371,7 +378,7 @@ impl Completion {
         self.popup.contents().is_empty()
     }
 
-    fn replace_item(&mut self, old_item: lsp::CompletionItem, new_item: lsp::CompletionItem) {
+    fn replace_item(&mut self, old_item: CompletionItem, new_item: CompletionItem) {
         self.popup.contents_mut().replace_option(old_item, new_item);
     }
 
@@ -387,7 +394,7 @@ impl Completion {
         // > The returned completion item should have the documentation property filled in.
         // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion
         let current_item = match self.popup.contents().selection() {
-            Some(item) if item.documentation.is_none() => item.clone(),
+            Some(item) if !item.resolved => item.clone(),
             _ => return false,
         };
 
@@ -397,7 +404,7 @@ impl Completion {
         };
 
         // This method should not block the compositor so we handle the response asynchronously.
-        let future = match language_server.resolve_completion_item(current_item.clone()) {
+        let future = match language_server.resolve_completion_item(current_item.item.clone()) {
             Some(future) => future,
             None => return false,
         };
@@ -415,7 +422,13 @@ impl Completion {
                     .unwrap()
                     .completion
                 {
-                    completion.replace_item(current_item, resolved_item);
+                    completion.replace_item(
+                        current_item,
+                        CompletionItem {
+                            item: resolved_item,
+                            resolved: true,
+                        },
+                    );
                 }
             },
         );
@@ -469,25 +482,25 @@ impl Component for Completion {
             Markdown::new(md, cx.editor.syn_loader.clone())
         };
 
-        let mut markdown_doc = match &option.documentation {
+        let mut markdown_doc = match &option.item.documentation {
             Some(lsp::Documentation::String(contents))
             | Some(lsp::Documentation::MarkupContent(lsp::MarkupContent {
                 kind: lsp::MarkupKind::PlainText,
                 value: contents,
             })) => {
                 // TODO: convert to wrapped text
-                markdowned(language, option.detail.as_deref(), Some(contents))
+                markdowned(language, option.item.detail.as_deref(), Some(contents))
             }
             Some(lsp::Documentation::MarkupContent(lsp::MarkupContent {
                 kind: lsp::MarkupKind::Markdown,
                 value: contents,
             })) => {
                 // TODO: set language based on doc scope
-                markdowned(language, option.detail.as_deref(), Some(contents))
+                markdowned(language, option.item.detail.as_deref(), Some(contents))
             }
-            None if option.detail.is_some() => {
+            None if option.item.detail.is_some() => {
                 // TODO: set language based on doc scope
-                markdowned(language, option.detail.as_deref(), None)
+                markdowned(language, option.item.detail.as_deref(), None)
             }
             None => return,
         };
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -253,6 +253,17 @@ impl Completion {
                     // always present here
                     let item = item.unwrap();
 
+                    // apply additional edits, mostly used to auto import unqualified types
+                    let resolved_item = if item
+                        .additional_text_edits
+                        .as_ref()
+                        .map(|edits| !edits.is_empty())
+                        .unwrap_or(false)
+                    {
+                        None
+                    } else {
+                        Self::resolve_completion_item(doc, item.clone())
+                    };
 
                     // if more text was entered, remove it
                     doc.restore(view, &savepoint, true);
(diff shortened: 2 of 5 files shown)


helix-term/src/ui/completion.rs
```diff
@@ -379,7 +398,7 @@ impl Completion {
         // > The returned completion item should have the documentation property filled in.
         // https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_completion
         let current_item = match self.popup.contents().selection() {
-            Some(item) if item.documentation.is_none() => item.clone(),
+            Some(item) if !item.resolved => item.clone(),
```
the-mikedavis, line 401: 🚀 I'm glad to see 25bbd13 for this case in particular: checking `documentation.is_none()` felt quite hacky

helix-term/src/ui/completion.rs
```diff
@@ -213,14 +227,27 @@ impl Completion {
...
+                    // lsp server to avoid messing up rerequesting completions. Once a
+                    // completion has been selected (with) tab it's always accepted whenever anything
+                    // is typed. The only way to avoid that is to explicitly abort the completion
+                    // with esc/c-c. This will remove the "ghost" transaction.
+                    //
+                    // The ghost transaction is modeled with a transaction that is not send to the LS.
+                    // (apply_temporary) and a savepoint. It's extremly important this savepoint is restored
+                    // (also without sending the transaction to the LS) *before any further transaction is applied*.
+                    // Otherwise incremental sync breaks (since the state of the LS doesn't match the state the transaction
+                    // is applied to).
```
the-mikedavis, line 243:
```suggestion
                    // Update creates "ghost" transactions which are not send to the
                    // lsp server to avoid messing up re-requesting completions. Once a
                    // completion has been selected (with) tab it's always accepted whenever anything
                    // is typed. The only way to avoid that is to explicitly abort the completion
                    // with esc/c-c. This will remove the "ghost" transaction.
                    //
                    // The ghost transaction is modeled with a transaction that is not sent to the LS.
                    // (apply_temporary) and a savepoint. It's extremely important this savepoint is restored
                    // (also without sending the transaction to the LS) *before any further transaction is applied*.
                    // Otherwise incremental sync breaks (since the state of the LS doesn't match the state the transaction
                    // is applied to).
```

the-mikedavis: minor nit about the "selected with tab" part - I think that could be clearer as "selected (for example with tab or C-n/C-p) since there are a few ways of selecting an item. It's possible that tab could get more complicated behavior in the future too if we use it to jump between tabstops so this might not age well

helix-view/src/document.rs
```diff
@@ -1049,12 +1073,23 @@ impl Document {
         }
         success
     }
+    /// Apply a [`Transaction`] to the [`Document`] to change its text.
+    pub fn apply(&mut self, transaction: &Transaction, view_id: ViewId) -> bool {
+        self.apply_inner(transaction, view_id, true)
+    }
+
+    /// Apply a [`Transaction`] to the [`Document`] to change its text.
+    /// without notifying the language servers. This is useful for temporary transactions
```
the-mikedavis, line 1082:
```suggestion
    /// Apply a [`Transaction`] to the [`Document`] to change its text
    /// without notifying the language servers. This is useful for temporary transactions
```

helix-term/src/ui/editor.rs
```diff
@@ -1265,12 +1286,22 @@ impl Component for EditorView {
...
 
-                                if let EventResult::Consumed(callback) = res {
-                                    consumed = true;
+                                    if let EventResult::Consumed(callback) =
+                                        completion.handle_event(event, &mut cx)
+                                    {
+                                        consumed = true;
+                                        Some(callback)
+                                    } else if let EventResult::Consumed(callback) =
+                                        completion.handle_event(&Event::Key(key!(Enter)), &mut cx)
```
the-mikedavis, line 1296: Ah this is clever. To me, this makes snippet completions feel really fluent 👍

the-mikedavis (commented):
I think this looks and feels great 👍

I only have some minor comments on the comments

helix-term/src/ui/completion.rs
```diff
@@ -88,6 +88,12 @@ impl menu::Item for CompletionItem {
     }
 }
 
+#[derive(Debug, PartialEq, Default, Clone)]
+struct CompletionItem {
+    item: lsp::CompletionItem,
+    resolved: bool,
+}
```
pickfire, line 95: Would it be better to put this as enum?

helix-term/src/ui/completion.rs
```diff
@@ -213,14 +227,27 @@ impl Completion {
...
+                    // (also without sending the transaction to the LS) *before any further transaction is applied*.
+                    // Otherwise incremental sync breaks (since the state of the LS doesn't match the state the transaction
+                    // is applied to).
+                    if editor.last_completion.is_none() {
+                        editor.last_completion = Some(CompleteAction::Selected {
+                            savepoint: doc.savepoint(view),
+                        })
+                    }
+                    // if more text was entered, remove it
+                    doc.restore(view, &savepoint, false);
```
pickfire, line 250: Is there mechanism to tell the language server to abort?

helix-view/src/editor.rs
```diff
@@ -900,9 +900,14 @@ enum ThemeAction {
...
-    pub trigger_offset: usize,
-    pub changes: Vec<Change>,
+pub enum CompleteAction {
+    Applied {
+        trigger_offset: usize,
+        changes: Vec<Change>,
+    },
+    /// A savepoint of the currently active completion. The completion
+    /// MUST be restored before sending any event to the LSP
+    Selected { savepoint: Arc<SavePoint> },
```
pickfire, line 910: I am a bit confused with reading this name, what is selected? Only a bit clearer when I read the long comments above.

helix-term/src/ui/completion.rs
pascalkuthe: yeah that's a good point the tab was always supposed to be in brackets and was means as an example just to make sure it's clear what selected meant. I updated the comment to be more clear

helix-term/src/ui/completion.rs
pascalkuthe: (and was the cause of #6261 since the server didn't return any documentation and therefore caused continuous resolve requests)

helix-term/src/ui/completion.rs
pascalkuthe: in #5336 this will become a struct anyway (the source server needs to be stored) so I don't think that's a good fit here. Just for the resolved status alone you would endup with:

``` rust
enum ResolveStatus{
	Resolved,
	Unresolved,
}
```

which is not any clearer. Also completion item and the resolved status are orthogonal properties and we only care whether an item is resolved in a few place so I don't think it makes sense right now.

helix-term/src/ui/completion.rs
pascalkuthe: as with most things completions are entirely managed by the client so no there isn't

helix-view/src/editor.rs
pascalkuthe: `CompleteAction` represents the last action in the completion menu. You could either have:
* selected a completion (c-n, c-p, tab)
* or applied a completion (ret or any other key haver a completion was selected)
that's what this enum represents. If the completion menu is closed/replaced for whatever reason we must process this last action. For selected we must restore the savepoint and for Applied we must record the change so we can repeat it (existing code)

helix-term/src/ui/completion.rs
```diff
@@ -209,14 +223,27 @@ impl Completion {
...
-            // if more text was entered, remove it
-            doc.restore(view, &savepoint);
-
             match event {
-                PromptEvent::Abort => {
-                    editor.last_completion = None;
-                }
+                PromptEvent::Abort => {}
                 PromptEvent::Update => {
+                    // Update creates "ghost" transactions which are not send to the
```
the-mikedavis, line 229:
```suggestion
                    // Update creates "ghost" transactions which are not sent to the
```

the-mikedavis (approved):

helix-term/src/commands.rs
```diff
@@ -4169,16 +4169,23 @@ pub fn completion(cx: &mut Context) {
 
     let (view, doc) = current!(cx.editor);
 
+    let savepoint = if let Some(CompleteAction::Selected { savepoint }) = &cx.editor.last_completion
+    {
+        savepoint.clone()
+    } else {
+        doc.savepoint(view)
+    };
+
```
Philipp-M, line 4178: Can you explain what this does exactly (maybe add a comment then)? I still have not really understood how this'll happen (`last_completion` is `Some`), why it's necessary, and if it's even correct (I think when an item is selected and the completion is requested again, this should be included/applied I guess, but this seems to happen somewhere else already as discovered below).

I tried to trigger this case by pressing `C-x` while already a completion menu is open with a selected item, to trigger this case, but `last_completion` was already `None`, as the selected completion is applied when the completion menu doesn't swallow (that behavior seems to be added in this PR as well below in `editor.rs`).

Philipp-M (approved):
> ... esc now reset the document to the savepoint created before previewing a transaction

Doesn't do that for me, `esc` applies the current state (selected completion) and exits to normal mode for me, but I think this behavior is fine IMHO, `C-c` works as described though (which IMHO also makes sense). The PR description might not be up to date after having a second look, as this seems to be handled in the comments.

Took some time to wrap my head around this change.

I still think that showing the completion in the document should probably not be an actual edit of the document until the completion item is really applied in a logic sense at least, but this is nitpicking, and doesn't really have anything to do with this PR (as it's existing behavior)...
 
For now I think this is a good approach to the issue(s) at hand, though it would be nice at some point in the future to untangle complexity a little bit (editor <-> completion menu <-> current doc state)

helix-term/src/commands.rs
pascalkuthe: There are basically two versions of the document:
* the version stored in the server
* the version stored in the client

In theory, these are synced but they can become out of sync. That's for example why we have to check version numbers before applying workspace edits. This PR introduced the concept of virtual edits which are not sent to the server. These are used when selecting a completion. This purposefully desynchronizes the two document versions so that the selected completion doesn't affect the server-side completion filtering (which can produce weird results or missing completion items). 

The savepoint in CompleteAction::Selected { savepoint } represents the last state send to the server.  Before we interact with the LSP in any way we must always reset to this state. This is not a big problem for async requests/notifiactions from the server as we already check the version number anyway. It's also not a problem for commands trigger by keypresses. Because every keypress will automatically apply the completion (which synchronizes the server). This is also why `c-x` doesn't trigger this branch as you are pressing a key.

However, what is a problem is places where we send requests to the server asynchronously so not directly triggered by a keypress. Completion requests are sent automatically on idle-timeout. The idle timeout does not cause the completion to be automatically accepted tough and can therefore cause invalid positions to be sent to the server. With a robust server this usually just means no/wrong completions but the problem is then that the completions it sends back will (as always) refer to the state of the document when the completion request was sent. However, as the server side state is (purposefully) out of sync that means that the server will send back potentially invalid edits which would cause similar problems as fixed in #6173. 

An example of what would trigger this branch:

Initial State

Server:
```
foo
```

Client:
```
foo
```

request completions and type `b`

```
foob
```

Client:
```
foob
```

server sends back the following edit: `foo` -> `foobar` and `foo` -> `foober` (notice that the `b` will be removed by the savepoint) as a completion items, You select the first completion (`foobar`)  with tab but don't confirm the selection.

Server (purposefully out of sync):

```
foob
```

Client:
```
foobar
```

now wait for the completion re-request on idle timeout. This does not close the completion menu (that would be weird). Without this code we would request completions at the end of `foobar` which will be somewhere else in the document for the server which could cause all of the following:
* crash the client (out of range edit)
* cause an invalid edit to be applied
* just show no completions (so randomly removing `foober` from the list which is especially annoying with a low idle timeout and slow scrolling)
* crash the server immidietly (if the new cursor position is out of bounds)
* cause a permanent server/client desync making the LS misbehave until restarted

This is definitely an edgecase but it's important to correctly to account for to avoid really weird bugs

pascalkuthe:
> I still think that showing the completion in the document should probably not be an actual edit of the document until the completion item is really applied in a logic sense at least

This would definitely be simpler but I think this preview behavior is quite useful and many people prefer that. I don't think the virtual text is really a good fit here and would be very complex to implement.

helix-term/src/commands.rs
Philipp-M: Thanks for the thorough explanation!

> now wait for the completion re-request on idle timeout

I have not yet encountered this (was a thought I also had while reviewing)

AFAIK completion requests aren't sent on idle timeout when the completion menu is open (probably worth a PR though, had this feature in mind as well already).
Have you triggered this branch while testing? I currently can't think of a case where this branch is triggered.

Still I think it's safer to keep this, shouldn't hurt at least, when for whatever reason that edge case occurs where the completion menu is still open and a new completion request is sent, probably because of a new feature (like rerequesting completion items on idle-timeout when the completion menu is still open)

Philipp-M:
> This would definitely be simpler but I think this preview behavior is quite useful and many people prefer it to the VSCode/Intelij approach. I don't think the virtual text is really a good fit here (it would need to be concealed which is not implemented yet and conceal hides the cursor) and would be very complex to implement.

Yeah we should definitely keep that behavior, thanks for the explanation regarding complexity with a potential virtual text implementation, with snippets and jumping of the cursor etc. this would certainly make things more difficult, I can imagine.

> ... If you change the document using potentially asynchronously computed edits that are purposefully unsynchronized that's fundamentally just very complex to handle.

Yeah it's definitely not trivial and may require a bigger refactor. I was thinking a little about having some kind of branching and merging behavior for all kinds of edits and likely the whole history(-tree) (maybe by using CRDTs, for generally easier async handling of edits/transactions). So it would be something like this:

* completion -> snapshot/fork of current state, the document and the language-server use this snapshot.
* when selecting completions the editor forks temporary branches from this snapshot with the application of the selected completion item, and when actually applied the "actual" state of the editor is updated to this temporary branch
* this temporary branch is used for rendering the document in the meantime (however this would be implemented)

It's already kind of going this way but I think it may be easier to comprehend, when this logic is abstracted away at a lower level (I may be wrong though).

I think this behavior could be beneficial for different kinds of applications as well (e.g. previewing code completions before applying them).

helix-term/src/commands.rs
pascalkuthe: I always assumed that was the case but that would explain some problems I have been running into... we really should be requesting completions especially to handle incomplete completion requests. I want to work on that soon anyway so at worst its a bit of future proofing

poliorcetics:
I have been testing this branch for a few weeks now and have encountered the following problem:

When editing Rust code, at some point there is an edit in Helix that is not synchronized with Rust-Analyzer, which is then crashes in two different ways (I haven't taken the time to reliably reproduce them yet):

<details><summary>Way 1, <code class="notranslate">ServerError(-32801)</code class="notranslate"></summary>
<pre class="notranslate"><code class="notranslate">2023-04-20T12:01:07.545 helix_view::editor [ERROR] editor error: Async job failed: protocol error: ServerError(-32801): content modified
2023-04-20T12:01:22.428 helix_lsp::transport [ERROR] <- ServerError(-32801): content modified
2023-04-20T12:01:22.429 helix_view::editor [ERROR] editor error: Async job failed: protocol error: ServerError(-32801): content modified
2023-04-20T12:01:22.502 helix_lsp::transport [ERROR] Tried sending response into a closed channel (id=Num(715)), original request likely timed out
2023-04-20T12:01:25.979 helix_lsp::transport [ERROR] Tried sending response into a closed channel (id=Num(728)), original request likely timed out
2023-04-20T12:01:53.574 helix_lsp::transport [ERROR] Tried sending response into a closed channel (id=Num(751)), original request likely timed out
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "Panic context:\n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "> \n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "version: 0.0.0 (2400b36a2 2023-04-19)\n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "notification: textDocument/didChange\n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "\n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "thread 'LspServer' panicked at 'assertion failed: self.is_char_boundary(n)', /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/alloc/src/string.rs:1819:29\n"
2023-04-20T12:04:23.038 helix_lsp::transport [ERROR] err <- "stack backtrace:\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "   0: rust_begin_unwind\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "   1: core::panicking::panic_fmt\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "   2: core::panicking::panic\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:114:5\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "   3: alloc::string::String::replace_range\n"
2023-04-20T12:04:23.064 helix_lsp::transport [ERROR] err <- "   4: rust_analyzer::lsp_utils::apply_document_changes\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   5: core::ops::function::FnOnce::call_once\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   6: rust_analyzer::dispatch::NotificationDispatcher::on\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   7: rust_analyzer::main_loop::<impl rust_analyzer::global_state::GlobalState>::run\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   8: rust_analyzer::main_loop::main_loop\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   9: rust_analyzer::run_server\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: \"SendError(..)\"', lib/lsp-server/src/stdio.rs:26:37\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "stack backtrace:\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   0: rust_begin_unwind\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   1: core::panicking::panic_fmt\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "   2: core::result::unwrap_failed\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/result.rs:1790:5\n"
2023-04-20T12:04:23.065 helix_lsp::transport [ERROR] err <- "note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\n"
2023-04-20T12:04:23.907 helix_lsp::transport [ERROR] err: <- StreamClosed
2023-04-20T12:04:23.907 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T12:04:23.909 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T12:04:23.910 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T12:04:24.081 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })</code class="notranslate"></pre class="notranslate">
</details> 

<details><summary>Way 2, Crash on a windows-looking path, happens while cross compiling for windows from macOS</summary>
<pre class="notranslate"><code class="notranslate">2023-04-20T17:35:33.215 helix_view::editor [ERROR] editor error: Async job failed: protocol error: ServerError(-32801): content modified
2023-04-20T17:37:00.187 helix_lsp::transport [ERROR] <- ServerError(-32801): content modified
2023-04-20T17:37:00.191 helix_view::editor [ERROR] editor error: Async job failed: protocol error: ServerError(-32801): content modified
2023-04-20T17:37:02.259 helix_lsp::transport [ERROR] <- ServerError(-32801): content modified
2023-04-20T17:37:02.261 helix_view::editor [ERROR] editor error: Async job failed: protocol error: ServerError(-32801): content modified
2023-04-20T17:37:16.224 helix_lsp::transport [ERROR] Tried sending response into a closed channel (id=Num(580)), original request likely timed out
2023-04-20T17:37:44.396 helix_lsp::transport [ERROR] err <- "[ERROR rust_analyzer::main_loop] flycheck 0: File with cargo diagnostic not found in VFS: file not found: /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74\\library\\core\\src\\result.rs\n"
2023-04-20T17:37:44.397 helix_lsp::transport [ERROR] err <- "[ERROR rust_analyzer::main_loop] flycheck 0: File with cargo diagnostic not found in VFS: file not found: /rustc/2c8cc343237b8f7d5a3c3703e3a87f2eb2c54a74\\library\\core\\src\\result.rs\n"
2023-04-20T17:38:03.396 helix_lsp::transport [ERROR] Tried sending response into a closed channel (id=Num(693)), original request likely timed out
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "Panic context:\n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "> \n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "version: 0.0.0 (2400b36a2 2023-04-19)\n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "notification: textDocument/didChange\n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "\n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "thread 'LspServer' panicked at 'assertion failed: self.is_char_boundary(n)', /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/alloc/src/string.rs:1819:29\n"
2023-04-20T17:38:14.461 helix_lsp::transport [ERROR] err <- "stack backtrace:\n"
2023-04-20T17:38:14.514 helix_lsp::transport [ERROR] err <- "   0: rust_begin_unwind\n"
2023-04-20T17:38:14.514 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   1: core::panicking::panic_fmt\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   2: core::panicking::panic\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:114:5\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   3: alloc::string::String::replace_range\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   4: rust_analyzer::lsp_utils::apply_document_changes\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   5: core::ops::function::FnOnce::call_once\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   6: rust_analyzer::dispatch::NotificationDispatcher::on\n"
2023-04-20T17:38:14.515 helix_lsp::transport [ERROR] err <- "   7: rust_analyzer::main_loop::<impl rust_analyzer::global_state::GlobalState>::run\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "   8: rust_analyzer::main_loop::main_loop\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "   9: rust_analyzer::run_server\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "thread '<unnamed>' panicked at 'called `Result::unwrap()` on an `Err` value: \"SendError(..)\"', lib/lsp-server/src/stdio.rs:26:37\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "stack backtrace:\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "   0: rust_begin_unwind\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/std/src/panicking.rs:575:5\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "   1: core::panicking::panic_fmt\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/panicking.rs:64:14\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "   2: core::result::unwrap_failed\n"
2023-04-20T17:38:14.516 helix_lsp::transport [ERROR] err <- "             at /rustc/9eb3afe9ebe9c7d2b84b71002d44f4a0edac95e0/library/core/src/result.rs:1790:5\n"
2023-04-20T17:38:14.517 helix_lsp::transport [ERROR] err <- "note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.\n"
2023-04-20T17:38:15.553 helix_lsp::transport [ERROR] err: <- StreamClosed
2023-04-20T17:38:15.553 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T17:38:15.555 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T17:38:15.556 helix_view::editor [ERROR] editor error: Async job failed: server closed the stream
2023-04-20T17:38:15.696 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:15.938 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:16.187 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:19.350 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:19.642 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:20.363 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:20.606 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:21.419 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:22.670 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:22.670 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:22.670 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:22.670 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })
2023-04-20T17:38:22.671 helix_lsp::transport [ERROR] err: <- IO(Os { code: 32, kind: BrokenPipe, message: "Broken pipe" })</code class="notranslate"></pre class="notranslate">
</details>

pascalkuthe:
> I have been testing this branch for a few weeks now and have encountered the following problem:
> 
> When editing Rust code, at some point there is an edit in Helix that is not synchronized with Rust-Analyzer, which is then crashes in two different ways (I haven't taken the time to reliably reproduce them yet):
> Way 1, `ServerError(-32801)`
> Way 2, Crash on a windows-looking path, happens while cross compiling for windows from macOS

We actually had multiple bug reports of the exact same bug report from people on matrix that were just using the latest helix release so I don't think that bug is actually related to this PR at all

pascalkuthe:
> > This would definitely be simpler but I think this preview behavior is quite useful and many people prefer it to the VSCode/Intelij approach. I don't think the virtual text is really a good fit here (it would need to be concealed which is not implemented yet and conceal hides the cursor) and would be very complex to implement.
> 
> Yeah we should definitely keep that behavior, thanks for the explanation regarding complexity with a potential virtual text implementation, with snippets and jumping of the cursor etc. this would certainly make things more difficult, I can imagine.
> 
> > ... If you change the document using potentially asynchronously computed edits that are purposefully unsynchronized that's fundamentally just very complex to handle.
> 
> Yeah it's definitely not trivial and may require a bigger refactor. I was thinking a little about having some kind of branching and merging behavior for all kinds of edits and likely the whole history(-tree) (maybe by using CRDTs, for generally easier async handling of edits/transactions). So it would be something like this:
> 
>     * completion -> snapshot/fork of current state, the document and the language-server use this snapshot.
> 
>     * when selecting completions the editor forks temporary branches from this snapshot with the application of the selected completion item, and when actually applied the "actual" state of the editor is updated to this temporary branch
> 
>     * this temporary branch is used for rendering the document in the meantime (however this would be implemented)
> 
> 
> It's already kind of going this way but I think it may be easier to comprehend, when this logic is abstracted away at a lower level (I may be wrong though).
> 
> I think this behavior could be beneficial for different kinds of applications as well (e.g. previewing code completions before applying them).

Creating a lower abstraction that we currently have is very difficult. Helix is fundamentally build around a CRDT like transactions. Every change to a document must be a transaction. This even includes any temporary changes. The entire editor is fundamentally built around that model. 


For example when a transaction is applied to a buffer:
* the diff gutter updates
* diagnostics are moved around
* tree sitter parsing rungs
* changes are sent to the LSP


This "transactional" view of the world is very core so any snapshots we may have must also be updated using this transactional model. The savepoint system is exactly this: A snapshot of a document that is internally represented as an automatically updated CRDT transaction that can be used to revert the buffer to a previous snapshot.

I guess what you are getting at is that there should be somekind of "LSP snapshot" which is automatically used. However, this doesn't really work. Basically all LSP actions require interactiing with the buffer. This can only work if the server and client buffer matches. Completions are very much special here because these "LSP snapshots" are created by the completion menu so for completion rerequest its ok to revert this LSP snapshot. Everything else should just fail/revert the completion snapshot.

pascalkuthe:
@poliorcetics that crash is an existing issue see https://github.com/helix-editor/helix/pull/6921 that is unrelated to this pr

Philipp-M:
Thanks for the explanation, though I'm aware of most of that.

What I meant with "CRDTs may be helpful" is the commutative property of transactions (which is pretty much all what CRDTs are about), so that no matter when the transaction occurred it all leads to the same (expected) result.  This is not (yet) implemented (and not a trivial task, especially with multi line/file transactions), but would be quite helpful for various different domains (including this).

Anyway all of this is orthogonal to this PR and would require quite a few changes in the core.


Is there anything blocking this PR? AFAIK it's the only missing thing for the (bugfix-)release isn't it?

pascalkuthe:
> Thanks for the explanation, though I'm aware of most of that.
> 
> What I meant with "CRDTs may be helpful" is the commutative property of transactions (which is pretty much all what CRDTs are about), so that no matter when the transaction occurred it all leads to the same (expected) result. This is not (yet) implemented (and not a trivial task, especially with multi line/file transactions), but would be quite helpful for various different domains (including this).
> 
> Anyway all of this is orthogonal to this PR and would require quite a few changes in the core.


Transaction::map is not yet implemented, it seems attractive and I thaught about implementing it before but its not actually a good fit for completions.  The important realization is that the text applied by the user must always be removed before a completion is applied even with a CRDT. The text by the user is just used to filter the completion but its not part of the transactions send by the server. So for example if you complete at `foo.` `type `bar` and selection the `foobar` completion and used a CRDT to map the completion transaction you would endup with `foo.barfoobar`.  So the snapshot based system is not a weak emulation of CRDT but is actually required to make completions work.


I think this PR is ready to merge just waiting until @archseer has time to review it.

archseer (approved):

PR #6613, opened 2023-04-06
Initial support for history search in command-line mode

LEI:
This is an attempt to implement "partial history search" allowing `:a<Up>` to show the previous entry starting with "a".

Added search for `<Up>`/`<Down>` and `<C-p>`/`<C-n>` and changed their behavior, even with no search term:
- `:<Up><Down>` now goes back to the initial empty state
- and prevent `<C-n>` or `<Up>` from cycling back the first history entry.

The prompt value at the end of the existing `history_completion` test was "asdf" before but now `<tab>` seems to have an effect, I am guessing this test should only verify the absence of errors?

Maybe this could apply to "/" and "?" as well.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1148,4 +1148,8 @@ impl Application {
 
         errs
     }
+
+    pub fn get_compositor_layers(&self) -> &Vec<Box<dyn crate::compositor::Component>> {
+        &self.compositor.layers
+    }
 }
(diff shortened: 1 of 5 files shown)


pascalkuthe:
It looks like you are doing a simple prefix match here. Using fuzzy matching would be preferable I think we use it for all completions (both in the editor and command mode)

LEI:
How would fuzzy matching work more precisely? I would not expect `:b<up>` to match `abc` for example.

I think this is the default in Vim, and a similar keybind is possible with [readline](https://www.man7.org/linux/man-pages/man3/readline.3.html):

       history-search-backward
              Search backward through the history for the string of
              characters between the start of the current line and the
              current cursor position (the point).  The search string
              must match at the beginning of a history line.  This is a
              non-incremental search.

PR #6616, opened 2023-04-06
Add command to change CWD to current buffer's directory

ksdrar:


ksdrar:
I was checking out the other pull request and thought it would be cool to have a `$parent` variable for the `:cd` command. So, I added it to the `change_current_directory` function in `typed.rs`. But, I'm not sure if this is the right way to add variables to commands. What do you think?

```rust
    let dir = {
        let arg = args
            .first()
            .context("target directory not provided")?
            .trim();

        if arg == "$parent" {
            doc!(cx.editor)
                .path()
                .and_then(|path| path.parent())
                .expect("Current buffer has no path or parent")
                .to_path_buf()
        } else {
            helix_core::path::expand_tilde(arg.as_ref())
        }
    };
```

the-mikedavis:
That seems too hacky to me. Initial support for variables is started in #3393

PR #6622, opened 2023-04-06
Ayu dark improvements

palango:
Some improvements to the `ayu_dark` theme. Happy to discuss color choices, this is just what I use in my local config.

### Before
Screenshot 2023-04-06 at 15 11 12

### After

Screenshot 2023-04-06 at 15 10 56

### DIFF
--- a/runtime/themes/ayu_dark.toml
+++ b/runtime/themes/ayu_dark.toml
@@ -39,6 +39,9 @@
 "ui.linenr" = "dark_gray"
 "ui.linenr.selected" = "gray"
 "ui.statusline" = { fg = "foreground", bg = "black" }
+"ui.statusline.normal" = { fg = "black", bg = "blue" }
+"ui.statusline.insert" = { fg = "black", bg = "orange" }
+"ui.statusline.select" = { fg = "black", bg = "magenta" }
 "ui.cursorline" = { bg = "black" }
 "ui.popup" = { fg = "#7B91b3", bg = "black" }
 "ui.window" = "dark_gray"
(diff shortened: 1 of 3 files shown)


David-Else:
Screen shots of the changes would be very helpful.

runtime/themes/ayu_dark.toml
```diff
@@ -33,11 +33,15 @@
 
 # Interface
 "ui.background"= { bg = "background" }
-"ui.cursor" = { fg = "dark_gray", bg = "orange" }
+"ui.cursor" = { fg = "dark_gray", bg = "yellow" }
+"ui.cursor.primary" = { fg = "dark_gray", bg = "orange" }
```
pickfire, line 37: Maybe do the same for ayu light and ayu mirage? I don't think this is done in other editors but maybe it is useful since we need to differentiate it.

runtime/themes/ayu_dark.toml
```diff
@@ -47,7 +51,7 @@
 "ui.text.info" = "foreground"
 "ui.virtual.whitespace" = "dark_gray"
 "ui.virtual.ruler" = { bg = "black" }
-"ui.virtual.inlay-hint" = { fg = "#e6b450", bg = "#302a20" }  # original bg #e6b45033
+"ui.virtual.inlay-hint" = { fg = "gray" }
```
pickfire, line 54: Also, I put orange previously because it looks like the rest of the ayu themes, as seen in vscode.

Yes, gray looks more like other themes but I find orange quite readable and nice too.

palango: Sure, the gray might be personal preference, I can overwrite that locally.

runtime/themes/ayu_dark.toml
palango: Done

palango:
@pickfire Please have another look.

PR #6632, opened 2023-04-06
23.03 - Allow opening new files into current helix session from the command line

michaeljones:
Using Unix socket as a communication device. Only works on unix and isn't even marked up properly at the moment so compliation on windows will fail.

I have no expectations of this making it into the project. It is just a quick hack.


michaeljones:
Sorry opened in the wrong repo. My bad.

michaeljones:
I meant to open it in my fork as a reference to link to from https://github.com/helix-editor/helix/discussions/5107

PR #6633, opened 2023-04-06
Allow filtering fetched or built languages when using --grammar

kareigu:
Similar to having `use-grammars` in `languages.toml`, this PR implements allowing users the filter which grammars are fetched or built by passing a list of languages as a parameter to `--grammar fetch` or `--grammar build`.

If no list is provided all available grammars or fetched or built as is current behaviour as well.

**Example output**
```sh
hx --grammar build
Building 143 grammars

hx --grammar build rust cpp
Building 2 grammars
```

Rationale behind this feature is allowing users to only download the few grammars they might need, without having to mess around with configuration files.

One question arises however with how this should interact with the `use-grammars` field in `languages.toml`. With the current implementation settings configured in `languages.toml` are still honored, but maybe it would make more sense for those settings to be overridden by the list supplied to the `--grammar` parameter?


helix-loader/src/grammar.rs
```diff
@@ -146,8 +152,14 @@ pub fn fetch_grammars() -> Result<()> {
...
 
-pub fn build_grammars(target: Option<String>) -> Result<()> {
-    let grammars = get_grammar_configs()?;
+pub fn build_grammars(target: Option<String>, filter: Option<Vec<String>>) -> Result<()> {
+    let mut grammars = get_grammar_configs()?;
+    if let Some(filter) = filter {
+        if filter.len() > 0 {
+            grammars.retain(|grammar| filter.contains(&grammar.grammar_id));
+        }
+    }
```
pickfire, line 161: Why not just pass vec without option?

helix-term/src/args.rs
```diff
@@ -45,8 +56,24 @@ impl Args {
...
+                        vec.reserve_exact(remaining);
+                        while let Some(arg) = argv.next() {
+                            vec.push(arg);
+                        }
+                        args.grammar_input = GrammarInput::Fetch(vec);
+                    }
+                    Some("build") => {
+                        let remaining = argv.size_hint().0;
+                        let mut vec: Vec<String> = Vec::new();
+                        vec.reserve_exact(remaining);
```
pickfire, line 71: Why not do `with_capacity`?

helix-loader/src/grammar.rs
kareigu: The original thought process was that this would be a more explicit way of saying you won't filter anything. But looking now I have to agree, it makes more sense to just pass an empty vec instead.

Rudxain:
stale?

helix-term/src/main.rs
```diff
@@ -65,7 +65,9 @@ FLAGS:
     --health [CATEGORY]            Checks for potential errors in editor setup
                                    CATEGORY can be a language or one of 'clipboard', 'languages'
                                    or 'all'. 'all' is the default if not specified.
-    -g, --grammar {{fetch|build}}    Fetches or builds tree-sitter grammars listed in languages.toml
+    -g, --grammar {{fetch|build}} [..languages]    
+                                   Fetches or builds tree-sitter grammars listed in languages.toml. 
+                                   Can be followed by a list of language names to only fetch or build those.
```
pickfire, line 70: How is this list being specified? The current way doesn't seemed quite clear, I don't get why we can't fetch and build everything, is it because it's slow or for other reasons?

Rudxain: I thought that was out-of-scope for this PR, but then I realized that duping the list for each cmd is very error-prone, even when user does it in a script. It'd be really nice to have a single `fetch && build` "alias"!

PR #6635, opened 2023-04-06
:log-open opens the current log file as specified by --log

### LINKED ISSUE #6469: :log-open does not respect --log flag
### Summary

When running `hx --log /tmp/helix.log`, the `:log-open` command opens `~/.cache/helix/helix.log`, not `/tmp/helix.log`

### Reproduction Steps

 

I tried this:

1. `hx --log /tmp/helix.log`
2. `:log-open`

I expected this to happen:

Opens /tmp/helix.log

Instead, this happened:

Opens default log location

### Helix log

```
```

</details>
[helix.log](https://github.com/helix-editor/helix/files/11088995/helix.log)

### Platform

Linux

### Terminal Emulator

st 0.8.5

### Helix Version

22.12-405-gd458ddd0

fherenius:
MR should fix the issue mentioned in https://github.com/helix-editor/helix/issues/6469.

I used a similar implementation to `CONFIG_FILE`. Tests should still be added but I'll ensure people agree on the implementation first.

Fixes #6469

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -11,6 +11,8 @@ static RUNTIME_DIRS: once_cell::sync::Lazy<Vec<PathBuf>> =
 
 static CONFIG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
 
+static LOG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
+
 pub fn initialize_config_file(specified_file: Option<PathBuf>) {
     let config_file = specified_file.unwrap_or_else(|| {
         let config_dir = config_dir();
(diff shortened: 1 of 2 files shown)


pickfire (approved):
I wonder if we really need a static variable just to do this since we can just pass the config around, but since I saw config file is like that we can keep it like this too. But I will let others to merge this.

helix-loader/src/lib.rs
```diff
@@ -26,6 +28,11 @@ pub fn initialize_config_file(specified_file: Option<PathBuf>) {
     CONFIG_FILE.set(config_file).ok();
 }
 
+pub fn initialize_log_file(specified_file: PathBuf) {
```
the-mikedavis, line 31: Like `initialize_config_file`, this can take an `Option<PathBuf>` and we can move the directory creation (from [here](https://github.com/helix-editor/helix/blob/f6d65cf089a0b845ef01b91d0424a9ffe1e9ea45/helix-term/src/main.rs#L45-L48)) into that block. That way we don't create the cache_dir unless we use it

PR #6643, opened 2023-04-07
Fix typos

wutchzone:
Hello,
I have run the spellchecker across the whole codebase and found some typos.

### DIFF
--- a/helix-core/src/position.rs
+++ b/helix-core/src/position.rs
@@ -109,7 +109,7 @@ pub fn visual_coords_at_pos(text: RopeSlice, pos: usize, tab_width: usize) -> Po
 /// softwrapping positions are estimated with an O(1) algorithm
 /// to ensure consistent performance for large lines (currently unimplemented)
 ///
-/// Usualy you want to use `visual_offset_from_anchor` instead but this function
+/// Usually you want to use `visual_offset_from_anchor` instead but this function
 /// can be useful (and faster) if
 /// * You already know the visual position of the block
 /// * You only care about the horizontal offset (column) and not the vertical offset (row)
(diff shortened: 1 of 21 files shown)


pascalkuthe:
I have been thinking about doing something similar with https://github.com/crate-ci/typos and then running it in CI but the initial typo PR would break a lot of PRs so I am not sure if it's worth it.

pickfire (approved):
Thanks.

wutchzone:
@pascalkuthe I have experimented with this idea of adding [`typos`](https://github.com/crate-ci/typos) to the CI in some of my projects. However, I found it quite frustrating because it can detect many false positive results. I think it can only work in the CI if the detected errors of the `typos` are not hard errors but rather warnings.

> ... running it in CI, but the initial typo PR would break a lot of PRs so I am not sure if it's worth it.

I think this is not a concern because it should be easily rebase-able. I think the main problem is to decide which files should be checked and which not. For example the `CHANGELOG.md`, you do not want to correct previous errors in previous releases, but only the new ones, because the changelog should be only "appendable". However, the `typos` can exclude only whole files and not the sections of the file.

However if you want I can draft a PR, where we can experiment with the `typos` in the CI.

pascalkuthe:
> @pascalkuthe I have experimented with this idea of adding [typos](https://github.com/crate-ci/typos) to the CI in some of my projects. However, I found it quite frustrating because it can detect many false positive results. I think it can only work in the CI if the detected errors of the typos are not hard errors but rather warnings.

The reason I wanted it in CI is that I want to use typos locally to check/fix my commits but if there are existing errors it will endup generating a bunch of unrelated changes to existing code which is unacceptable.  It's pretty straightforward to add exceptions to typos so it might still be worth it. 

> I think the main problem is to decide which files should be checked and which not. For example the CHANGELOG.md, you do not want to correct previous errors in previous releases, but only the new ones, because the changelog should be only "appendable". However, the typos can exclude only whole files and not the sections of the file.

I think retroactively fixing typos in the CHANGELOG is quite alright. But we have a ton of testcases in helix which consist of gibberish text (like `fooobar`). I don't want to get typo errors for those so that's a bit more challenging. 

> I think this is not a concern because it should be easily rebase-able.

Not quite if you rename a variable (by fixing a typo) and somebody uses that variable in their code then no conflict will show up and CI won't fail until we actually merge. We have almost 300 open PRs that's why we are usually pretty careful about renames for that reason.

pickfire:
I think right now typos have too many false positive in tests, maybe we might want to exclude some stuff, but more of an issue now I think is for themes and language support, that one takes more time to review and is less automated now.

We can add typos in another pull request but I will merge this for now.

PR #6644, opened 2023-04-07
Add DTD language support

KMikeeU:
- Added syntax highlighting for the DTD subset of the XML spec.
- Included .dtd and .ent as common file extensions

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2099,6 +2099,24 @@ roots = []
 name = "xml"
 source = { git = "https://github.com/RenjiSann/tree-sitter-xml", rev = "48a7c2b6fb9d515577e115e6788937e837815651" }
 
+
+[[language]]
+name = "dtd"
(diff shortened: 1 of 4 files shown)


languages.toml
```diff
@@ -2099,6 +2099,24 @@ roots = []
...
+
+[[language]]
+name = "dtd"
+scope = "source.dtd"
+injection-regex = "dtd"
+file-types = ["dtd", "ent"]
+indent = {tab-width = 2, unit = "  "}
+roots = []
+
+[language.auto-pairs]
```
the-mikedavis, line 2111: Why remove the auto-pairs for `'` and `"`?

KMikeeU: Excellent question! I don't know where those went, I'll add them back in :)

the-mikedavis (approved):

PR #6652, opened 2023-04-08
global indentation configuration

### LINKED ISSUE #3159: Allow setting `indent` globally
I currently have an incredibly long `~/.config/helix/languages.toml`, to which I append a copy of this for every language I use or encounter:

```
[[language]]
name = "..."
auto-pairs = false
```

I discovered while writing this that I don't actually have to set `auto-pairs = false` in each language anymore (I don't think the global one existed/worked when I started using Helix?), so that's neat, thanks!

I would like to be able to do the same for `indent`. The desired behavior is that I can say "always indent with tabs, for any file type", and spaces are only used if I open an existing file that is already indented with spaces.

Potential issue with this: Some languages (e.g. makefiles) require a particular type of indentation as part of their syntax. These languages should probably have some sort of "no, really, use this even if the user said something else" flag in their `indent` configuration, to avoid emitting broken makefiles when the user has set space-indentation globally. Alternatively, the language-level `indent` could be split into `indent` and `default-indent`, where the order of precedence is `language.indent > global.indent > language.default-indent`.

nnym:
Indentation can be configured globally as shown below.
```toml
[editor.indent]
tab-width = 4
unit = "t"
```
These settings are used when a buffer is loaded and an indentation style is not detected and override language-specific defaults. However, they can be overridden per language in `languages.toml` by setting `indent.required = true`.
```toml
[[language]]
name = "yaml"
indent = {tab-width = 2, unit = "  ", required = true}
```

Setting `indent.required = false` for a language in the user's `language.toml` requires setting `indent.{tab-width,unit}` too but I think that this is out of the scope of this patch given that the same happens when setting `indent.tab-width` or `indent.unit` only.
These changes should resolve #3159.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1440,7 +1440,7 @@ file-types = [{ glob = "Makefile" }, { glob = "makefile" }, "make", "mk", "mak",
 shebangs = ["make", "gmake"]
 injection-regex = "(make|makefile|Makefile|mk)"
 comment-token = "#"
-indent = { tab-width = 4, unit = "\t" }
+indent = { tab-width = 4, unit = "\t", required = true }
 
 [[grammar]]
 name = "make"
--- a/languages.toml
+++ b/languages.toml
@@ -3069,7 +3069,7 @@ file-types = ["nim", "nims", "nimble"]
 shebangs = []
 comment-token = "#"
 block-comment-tokens = { start = "#[", end = "]#" }
-indent = { tab-width = 2, unit = "  " }
+indent = { tab-width = 2, unit = "  ", required = true }
 language-servers = [ "nimlangserver" ]
 
 [language.auto-pairs]
--- a/languages.toml
+++ b/languages.toml
@@ -863,7 +863,7 @@ roots = ["pyproject.toml", "setup.py", "poetry.lock", "pyrightconfig.json"]
 comment-token = "#"
 language-servers = ["ruff", "jedi", "pylsp"]
 # TODO: pyls needs utf-8 offsets
-indent = { tab-width = 4, unit = "    " }
+indent = { tab-width = 4, unit = "    ", required = true }
 
 [[grammar]]
 name = "python"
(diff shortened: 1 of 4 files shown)


Jackojc:
Would be nice to try and merge this

XavierCS-dev:
Would really like this, annoying to have to keep setting it every time I open helix. It would be nice if there was some helix.rc or such so this wouldn't be a problem

languages.toml
```diff
@@ -695,7 +695,7 @@ roots = ["pyproject.toml", "setup.py", "poetry.lock", "pyrightconfig.json"]
 comment-token = "#"
 language-servers = [ "pylsp" ]
 # TODO: pyls needs utf-8 offsets
-indent = { tab-width = 4, unit = "    " }
+indent = { tab-width = 4, unit = "    ", required = true }
```
thecaralice, line 698:
```suggestion
indent = { tab-width = 4, unit = "    " }
```
Python allows using hard tabs for indentation (and actually any amount of spaces works too), it only disallows ambiguous mixing (i.e. when semantics depend on a specific tab width)

nnym: I didn't mark the indentation style as `required` because Python requires it. I did it in order to override the global setting for conformance with PEP 8.

thecaralice: I think `required` should only be applied to languages that do not support using tabs (like Nim) or spaces (like Makefile); everything else is basically just your stylistic choice which shouldn't affect anyone else. I mean, that's exactly the reason for using tabs instead of spaces.

nnym: I agree—I use tabs in Python and everywhere else—but this felt right to me at the time. In any case, I'll leave the decision to the maintainers.

oworope:
Any updates on this?

nnym:
@oworope https://github.com/helix-editor/helix/discussions/10363#discussioncomment-9513472

oworope:
So this won't be merged for now? We have to wait for plugin system to be implemented?

Issue #6044, opened 2023-02-18
Retroactively update indentation


Triton171:
Currently, we run indentation queries only when a new line is being inserted. However, other editors can in some cases update the indentation while typing, for example in C++ switch statements:
````cpp
switch (var) {
    case 0: // As soon as this can be recognized as a case label, it should be outdented
}
````
in python if statements:
````python
if condition:
    do_smth()
    else: # As soon as this can be recognized as an else statement, it should be outdented
````
or when inserting a closing curly brace `}` which should be outdented. CLion (basically IntelliJ for C++ development) even correctly indents multiple lines after surrounding them with curly braces (or adding something like an if statement around it):
````cpp
{ // Adding these curly braces causes the entire block to be correctly indented
do_smth();
do_smth_on_multiple_lines();
}
````

Implementing features like this in helix would require us to run some kind of indent query on every edit (or at least every time the syntax tree changes significantly). Because of that (and because it has the potential to be very annoying if the updated indentation is incorrect), this kind of auto-indent should probably off by default but I still think it would be very nice to support it in helix. As far as I see, there are a few different ways it could be implemented:

1. We could introduce a new kind of indent query "`update-indent.scm`" which just specifies situations in which the indent should be updated. This could either include the required update to the indentation (e.g. decrease the indentation by 1) or simply rerun the normal indent queries for the specified parts of the syntax trees. This is probably the easiest solution but it would require additional queries for all languages, for which we want to support this feature.

2. A different option would be to simply use the existing indent queries. The advantage to this is that we don't need additional queries. However, there are obviously some cases where the indent shouldn't be updated and we would need to exclude them: The indent computed by our heuristic may not match the indent that the user wants (either because there is a bug in the heuristic or because the file uses a non-conventional indentation style). Incorrectly updating the indentation once may be fine but if the user changes it back to what it was before, the indentation shouldn't be updated again.

    One possible heuristic for this could be to only update the indentation when a new node is inserted in the syntax tree (we may also want to require that this node doesn't contain any errors). This way, the indentation is only updated once and the user is able to correct it. Additionally, it might be a good idea to check that the indentation heuristic matches the actual indentation for surrounding code. If that's not the case, it's not very likely that the heuristic can correctly predict the indentation for the part of the code that changed.

These are only some ideas and I haven't started any implementation, so let me know your thoughts on this. Also, please tell me if this is not something that helix should include out-of-the-box.

yashpalgoyal1304:
[...]

PR #6653, opened 2023-04-08
Add python interactive else dedent

lucario387:
Python else statements were not dedented automatically, so adding a pattern to capture this case,

Example code:
```python
def test():
    if True:
        print("Ok")
        else: # After pressing ":", the line will become
    else:
```

### DIFF
--- a/runtime/queries/python/indents.scm
+++ b/runtime/queries/python/indents.scm
@@ -72,4 +72,6 @@
   "elif" @outdent)
 (else_clause
   "else" @outdent)
-
+(ERROR
+  (_) @outdent ":" . 
+  (#match? @outdent "^(else|elif)"))


the-mikedavis (commented):
I don't see any behavior change with this change to the indent query. As far as I know, automatic dedent is not currently implemented

lucario387:
> I don't see any behavior change with this change to the indent query. 

Sorry, I didn't test it enough. I was trying to mimic the behavior of `vim-python-pep8-indent`

> As far as I know, automatic dedent is not currently implemented
Does this mean indent calculation for each line is only calculated on pressing enter and not while typing?

kirawi:
It would require https://github.com/helix-editor/helix/issues/6044 to be implemented.

lucario387:
I see. Thanks for the explanation 

Is it okay for me to park this PR and wait till the PR above is merged?

pickfire:
I think we can close this for now given that when that PR is done most likely the python support have already been added.

PR #6656, opened 2023-04-08
Fix #6605: Remove soft-wrap.enable option wrapping.

### LINKED ISSUE #6605: Soft-wrap enable toggle does not work.
### Summary

`:toggle-option softwrap-enable` returns "Key `soft-wrap.enable` is not toggle-able" when soft-wrap is not set to true before. Works after a `:set-option softwrap-enable true`.

### Platform

Linux

### Terminal Emulator

foot

### Helix Version

23.03-16-g2bdb58fb

gibbz00:


archseer:
(Had to revert in master, @pascalkuthe pointed out a subtle bug with config override handling)

pascalkuthe:
This is not the right fix for the problem. Softwrap can be configured both per language and globally. The language configuration takes precedence over the global config. 

Therefore all softwrap options are options so that we can fallback to the global config.

With this PR the following language config would disable softwrap for `markdown` even if softwrap is globally enabled:

``` toml
[[language]]
name ="markdown"
softwrap.wrap-at-text-width = true
```

To fix this problem we should either add a sspegifal case to the default implementation of `edit::Config` to use `Some(false)`

PR #6666, opened 2023-04-09
Make ```cargo xtask themelint``` accepts arbitrary files

EruEri:
- Make the subsommand ```cargo xtask themelint``` accepts other files the one in the "<root_project>/runtime/themes" folder

### DIFF
--- a/xtask/src/themelint.rs
+++ b/xtask/src/themelint.rs
@@ -1,3 +1,5 @@
+use std::path::PathBuf;
+
 use crate::path;
 use crate::DynError;
 use helix_view::theme::Loader;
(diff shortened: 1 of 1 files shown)


xtask/src/themelint.rs
```diff
@@ -157,7 +158,13 @@ pub fn lint(file: String) -> Result<(), DynError> {
...
         return Ok(());
     }
-    let path = path::themes().join(file.clone() + ".toml");
+
+    let path = PathBuf::from(file.clone());
+    let path = if path.is_file() {
+        path
+    } else {
+        path::themes().join(file.clone() + ".toml")
+    };
```
pickfire, line 167: Why do you need to do this? Why not just add it to the themes repository?

EruEri: I thought that it would be easier to check a custom theme by just passing its path to the themelint subcommand.
But it's true that I haven't think about just copying the file into the themes repository

pickfire: I guess this isn't an issue, currently themelint needs more work to reduce false positive, but I think we don't really need to allow it to lint custom file, at least I think a better way for this issue is to copy the files to themes repository since sooner or later you will probably submit it to helix anyway.

pickfire:
I will be closing this issue since I don't see much benefit behind the change as the workflow to edit a theme file directly within themes/ is recommended, but if you think otherwise do re-open the issue, maybe I missed something here.

Issue #5867, opened 2023-02-07
Create a new named buffer using `:new ...`


CptPotato:
The `:new` command currently doesn't handle parameters that you pass to it. 

I think this is confusing in cases where you want to create a new document `:new foo.rs` and then save it using `:w` (helix reports an error "Can't save with no path set!" instead). In contrast to that, `:open foo.rs` actually has this behavior when the file does not exist, even though technically it's not "opening" a file.

In my opinion `:new foo.rs` should either work like `:open foo.rs` or provide an error message if parameters are passed. Another possibility is merging both into one command (implementation) where zero arguments creates a new scratch buffer and one or more would function like `:open`.

Initially discussed in #5865

pascalkuthe:
You can use :open for that see discussion in #6668

PR #6668, opened 2023-04-09
Command `new` open named buffer

sillent:
Hello

Refactor the `new` command, now it open a named buffer  that passed as arg
For resolving #5867

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -106,33 +106,7 @@ fn open(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> 
     }
 
     ensure!(!args.is_empty(), "wrong argument count");
-    for arg in args {
-        let (path, pos) = args::parse_file(arg);
-        let path = helix_core::path::expand_tilde(&path);
-        // If the path is a directory, open a file picker on that directory and update the status
-        // message
-        if let Ok(true) = std::fs::canonicalize(&path).map(|p| p.is_dir()) {
-            let callback = async move {
-                let call: job::Callback = job::Callback::EditorCompositor(Box::new(
-                    move |editor: &mut Editor, compositor: &mut Compositor| {
-                        let picker = ui::file_picker(path, &editor.config());
-                        compositor.push(Box::new(overlayed(picker)));
-                    },
-                ));
-                Ok(call)
-            };
-            cx.jobs.callback(callback);
-        } else {
-            // Otherwise, just open the file
-            let _ = cx.editor.open(&path, Action::Replace)?;
-            let (view, doc) = current!(cx.editor);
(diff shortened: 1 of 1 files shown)


the-mikedavis:
I would prefer the other idea from #5867:

> provide an error message if parameters are passed [to `:new`]

I don't like having `:new` and `:open` overlap so much in functionality.

sillent:
> I don't like having :new and :open overlap so much in functionality.

Why? If we look on the VIM, there :new can take additional argument, or :edit that take required argument, and they both open the new named buffer.

I think it's very comfortable

For quick save you can just tap :w - and work is saved, your mind is clean, no needs to write additional arguments for :w

pascalkuthe:
I think merging both commands and just making `:open` do the same and just making `:new` an alias for `:open` would be the nicest option (that was one of the options in https://github.com/helix-editor/helix/issues/5867) not sure if that is what you meant too @the-mikedavis

the-mikedavis:
`:new` should always open a fresh scratch document, so if we allow an argument passed to `:new` I think it should open an empty document even if the file exists, and it should return an error if the path passed is an existing directory. I don't want `:new` and `:open` to be essentially the same command.

My preference though would be to just to add a line to `new_file`:

```rust
ensure!(args.len() == 0, "Expected no arguments");
```

pascalkuthe:
> `:new` should always open a fresh scratch document, so if we allow an argument passed to `:new` I think it should open an empty document even if the file exists, and it should return an error if the path passed is an existing directory. I don't want `:new` and `:open` to be essentially the same command.
> 
> My preference though would be to just to add a line to `new_file`:
> 
> ```rust
> ensure!(args.len() == 0, "Expected no arguments");
> ```

Ah I see where you are coming from `:new` opening an existing file feels odd so these should probably be kept seperate. I didn't quite think that trough. If we would allow `:new` to accept a path then (which would create a scratch buffer or an error if the file exists) then to be consistent we should make `:open` also an error if the file doesn't exist yet.

This doesn't really feel that great to me though since `:open foo.rs` is usually not taught of as a scratch buffer but rather an on-disk file that just happens to be empty (and is treated that way by LSP integration and so on). The scratch buffer by comparison really is special and is not an on-disk file so `: new` not accepting any arguments also makes the most sense to me and keeps the current behaviour (just with the explicit check) :+1:

sillent:
So, it turns out these changes are not needed? 

Close the PR?

the-mikedavis:
I won't approve the current changes but I would be interested in the [other approaches](https://github.com/helix-editor/helix/pull/6668#issuecomment-1509249878) (especially the `ensure!` since the change would be so small), either in this PR or a new one

sillent:
Ok, I'm closing the PR because these changes should have changed the logic :new

Issue #6628, opened 2023-04-06
Panic when opening jump list


kirawi:
### Summary

This happened accidentally while editing, so there are no reproduction steps.

```
thread 'main' panicked at 'Attempt to slice past end of RopeSlice: slice end 213, RopeSlice length 182', /Users/kira/.cargo/registry/src/index.crates.io-6f17d22bba15001f/ropey-1.6.0/src/slice.rs:656:9
stack backtrace:
   0: _rust_begin_unwind
   1: core::panicking::panic_fmt
   2: ropey::slice::RopeSlice::slice
   3: <core::iter::adapters::map::Map<I,F> as core::iter::traits::iterator::Iterator>::fold
   4: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter
   5: core::ops::function::impls::<impl core::ops::function::FnOnce<A> for &mut F>::call_once
   6: <core::iter::adapters::flatten::FlatMap<I,U,F> as core::iter::traits::iterator::Iterator>::next
   7: alloc::vec::Vec<T,A>::extend_desugared
   8: <alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter
   9: helix_term::commands::jumplist_picker
  10: helix_term::ui::editor::EditorView::handle_keymap_event::{{closure}}
  11: helix_term::ui::editor::EditorView::handle_keymap_event
  12: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::handle_event
  13: helix_term::compositor::Compositor::handle_event
  14: hx::main_impl::{{closure}}
  15: tokio::runtime::park::CachedParkThread::block_on
  16: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
  17: hx::main

```

### Helix log

N/a

### Platform

macOS

### Terminal Emulator

Kitty 0.27

### Helix Version

Helix 23.03 (531b745c54e5d316603601cc05ade50ab02066ab)

kirawi:
\cc @pascalkuthe

pascalkuthe:
The cause for this panic is that the jumplist is updated lazily:
```
    /// A mapping between documents and the last history revision the view was updated at.
    /// Changes between documents and views are synced lazily when switching windows. This
    /// mapping keeps track of the last applied history revision so that only new changes
    /// are applied.
    doc_revisions: HashMap<DocumentId, size>,
```

[...]

PR #6672, opened 2023-04-09
Fix crash on opening jumplist

EsfoNL:
Hopefully solves #6628.
Moved the code to a separate function as was recommended in the issue.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -1491,7 +1491,13 @@ impl Editor {
         for (view, _focused) in self.tree.views_mut() {
             let doc = doc_mut!(self, &view.doc);
             view.sync_changes(doc);
-            for (doc_id, _selection) in view.jumps.clone().iter() {
+            for doc_id in view
+                .jumps
+                .iter()
+                .map(|e| e.0)
+                .collect::<Vec<DocumentId>>()
+                .iter()
+            {
                 let doc = doc_mut!(self, doc_id);
                 view.sync_changes(doc);
             }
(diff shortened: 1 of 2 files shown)


the-mikedavis (changes requested):
This change doesn't fully solve the issue: you can still cause a panic with the sequence `[<space>:new<ret><C-w>v:buffer-next<ret>d<space>j`. The current change syncs the view for each focused doc but the docs in the jumplist may not be focused. That sequence panics because the first scratch buffer isn't focused in the first view so it isn't synced.

We need to sync changes between each view and all of the docs in `view.jumps()`

EsfoNL:
Should be fixed, according to my testing

helix-view/src/editor.rs
```diff
@@ -1460,10 +1460,7 @@ impl Editor {
             self.ensure_cursor_in_view(view_id);
 
             // Update jumplist selections with new document changes.
-            for (view, _focused) in self.tree.views_mut() {
-                let doc = doc_mut!(self, &view.doc);
-                view.sync_changes(doc);
-            }
+            self.sync_views();
```
the-mikedavis, line 1463: We don't need to sync unfocused docs here so using a shared function does more work than we need. Let's eliminate the shared `sync_views` function and instead sync the views as we create the items for the file picker (currently [this block](https://github.com/helix-editor/helix/blob/e72ad1e731877570c6f728e1490485e4b31524a5/helix-term/src/commands.rs#L2570-L2578))

EsfoNL: because the `new meta` closure captures `&cx.editor`, it is impossible to also use `&mut cx.editor` to create the `&mut Doc' that is needed to sync the view with a doc so i have opted to sync the views earlier in the function.

the-mikedavis (approved):
Thanks!

PR #6675, opened 2023-04-09
tui: Allow toggling mouse capture at runtime

### LINKED ISSUE #5648: config-reload fails to change mouse configuration state

### LINKED ISSUE #6638: Turning off mouse mode leaves terminal in a broken state
### Summary

Setting `mouse = false` and reloading the config and then exiting breaks mouse handling in the terminal.  

Clicking or moving the mouse after quitting sends codes to the command line. Removing the line from config and reloading will stop this behaviour but doesn't seem to fully restore the terminals mouse handling; text selection no longer works the way it should.

This is happening with Ubuntu running in VMFusion on a M1 mac using the Tilix terminal.
I also checked this using the byobu terminal and saw the same behavior

This doesn't happen in MacOS running in the Warp terminal.

Could this be a problem with Helix or is it something with VMWare? 

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:

Instead, this happened:

### Helix log

```
```

</details>

### Platform

Linux on VMWare Fusion

### Terminal Emulator

Tilix. Byobu 

### Helix Version

helix 23.03

the-mikedavis:
This picks up changes to the `editor.mouse` option at runtime - either through `:set-option` or `:config-reload`. When the value changes, we tell the terminal to enable or disable mouse capture sequences.

Closes #6638
Closes #5648
Closes #5931

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -361,6 +361,9 @@ impl Application {
             ConfigEvent::Update(editor_config) => {
                 let mut app_config = (*self.config.load().clone()).clone();
                 app_config.editor = *editor_config;
+                if let Err(err) = self.terminal.reconfigure(app_config.editor.clone().into()) {
+                    self.editor.set_error(err.to_string());
+                };
                 self.config.store(Arc::new(app_config));
             }
         }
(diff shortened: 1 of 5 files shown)


pascalkuthe:
I think this supercedes #5931 (and therefore closes #5648)  IIRC

pascalkuthe (approved):
LGTM

PR #6677, opened 2023-04-09
Make the jumplist infinite

Philipp-M:
As I was annoyed a little bit about the short jump list at times, I made it infinite.

Maybe it makes sense to make this configurable non the less. But I think since one jump is only 64 bytes (most of the time) I don't see a reason why it shouldn't be infinite by default.


the-mikedavis:
The limit was intentional and it's for limiting the cost of updating the selections rather than space savings: https://github.com/helix-editor/helix/pull/4750

Philipp-M:
Ah makes sense, ok, maybe we should do it lazily at some time too? I'm closing this in the meantime.

PR #6694, opened 2023-04-10
Fix #6669: Theme preview doesn't return theme to normal

### LINKED ISSUE #6669: Theme preview doesn't return theme to normal when delete name with Alt-Backspace
### Summary

The theme preview when using :theme will keep the previewed theme if you delete name of theme using Alt-Backspace keys.
But if you delete name of theme by repeating Backspace key, then previewed theme return to normal.

### Reproduction Steps

 

I tried this:

1. `hx`

I expected this to happen:

Instead, this happened:

### Helix log

```
```

</details>

### Platform

macOS

### Terminal Emulator

alacritty 0.12.0 (5a72819)

### Helix Version

23.03-29-g3dd715a1

h1t:
Theme preview doesn't return theme to normal when delete name with Alt-Backspace

### DIFF
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -511,7 +511,8 @@ impl Component for Prompt {
             ctrl!('e') | key!(End) => self.move_end(),
             ctrl!('a') | key!(Home) => self.move_start(),
             ctrl!('w') | alt!(Backspace) | ctrl!(Backspace) => {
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Could you make the same change on the next three lines? We want to also call the prompt callback for other keybinds that change the prompt text

pickfire (approved):
Thanks. We have have a look of the other issues separately.

PR #6713, opened 2023-04-11
Allow overriding config and cache dir by env var

savente93:
Howdy folks. I wanted to add the ability to have my helix config behind a symlink for myself, then I saw the TODOs and figured I might as well kill two birds with one stone. 

I would have added tests, but `helix-loader` doesn't seem to have any and I wasn't quite sure how to hook it into the current framework. It's a very small change so I think it's okay, but if anyone would like to see them added and is willing to help me for a few minutes I'm happy to add them. 

If there's any additional requirements I wasn't aware of, please let me know. 

cheers :wave:

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -2,6 +2,7 @@ pub mod config;
 pub mod grammar;
 
 use etcetera::base_strategy::{choose_base_strategy, BaseStrategy};
+use std::fs::{create_dir_all, metadata};
 use std::path::{Path, PathBuf};
 
 pub const VERSION_AND_GIT_HASH: &str = env!("VERSION_AND_GIT_HASH");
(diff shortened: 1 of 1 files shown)


archseer:
Couldn't you just set the `XDG_CONFIG_HOME` and `XDG_CACHE_HOME` variables?

savente93:
I'm not familiar with those variables, though I believe you if you say Helix would obey them. I did look them up though and from what I understand that would mean that all applications would obey that which at least in my case is not what I want.

the-mikedavis:
Only if you set those variables globally. You can use them just for the helix command

```sh
XDG_CACHE_HOME=$HOME/tmp hx
```

or create a wrapper script for Helix that sets them somewhere custom:

```sh
#!/usr/bin/env bash
XDG_CACHE_HOME=/some/custom/cache/dir
/path/to/actual/binary/for/hx
```

Some packagers like nixpkgs use that strategy to adjust environment variables used by a package without polluting global environment variables

savente93:
okay that makes sense. Do you want me to close the PR? I don't really mind, but there were TODOs in the code about this so I don't know if someone else wanted this for some other reason? I still think this is a bit of a nicer solution, but I understand if you'd rather not add the extra code.

the-mikedavis:
Yeah I think this is not necessary. We also have the CLI flags for specifying the config and log files which I think would work as an alternative to the XDG environment variables too.

linsui:
I want this feature. :) It's much easier to set a env var than to add a wrapper.

linsui:
A problem mentioned in the matrix is that changing the XDG variables can affect programs invoked by helix, e.g. lsp.

PR #6721, opened 2023-04-12
Add nix language server and formatter

dlip:


cor:
`nil` is actually a lot better than `rnix-lsp` in my experience

nrdxp:
yes, please don't merge this rnix-lsp basically gets no development nowadays whereas nil is progressing very quickly and is already far superior.

dlip:
Thanks for the comments! I had no idea `nil` was an lsp. I thought it just meant `null` :facepalm: 

nixpkgs-fmt is probably not needed either since #3661 alejandra seems good. I'll close

PR #6742, opened 2023-04-13
ensure :toggle soft-wrap.enable works by default

### LINKED ISSUE #6605: Soft-wrap enable toggle does not work.

pascalkuthe:
Fixes #6605

As described in https://github.com/helix-editor/helix/pull/6656#issuecomment-1500887367 the correct fix here was simply to default to `Some(false)` for the global config (instead of `None` which lead to the same behaviour but caused issues with `:toggle`) while continuing to default to `None` for the language config so that the fallback keeps working correctly.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -749,7 +749,10 @@ impl Default for Config {
             bufferline: BufferLine::default(),
             indent_guides: IndentGuidesConfig::default(),
             color_modes: false,
-            soft_wrap: SoftWrap::default(),
+            soft_wrap: SoftWrap {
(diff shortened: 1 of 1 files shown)


dead10ck (approved):
Lgtm

the-mikedavis (approved):

PR #6743, opened 2023-04-13
emit cargo metadata duiring build scripts to avoid outdated buildscript outputs

pascalkuthe:
While triaging issues I often noticed that the revision reported by `helix --version` if often wrong. I finally found the reason for this: The `build.rs` script that extracts the HEAD branch by calling `git` is not automatically rerun so if you build helix then checkout/merge a branch/revision that doesn't touch `helix-loader` (which doesn't see that much activity compared to the rest of the editor) and rebuild then build script will not be rerun and`helix --version` will report the old (incorrect) revision.

I made the build script emit an appropriate `cargo:rerun-if-changed=` so that it is rerun if `HEAD` changes to fix that.

### DIFF
--- a/helix-loader/build.rs
+++ b/helix-loader/build.rs
@@ -1,4 +1,5 @@
 use std::borrow::Cow;
+use std::path::Path;
 use std::process::Command;
 
 const VERSION: &str = include_str!("../VERSION");
@@ -11,7 +12,7 @@ fn main() {
         .filter(|output| output.status.success())
         .and_then(|x| String::from_utf8(x.stdout).ok());
 
-    let version: Cow<_> = match git_hash {
+    let version: Cow<_> = match &git_hash {
         Some(git_hash) => format!("{} ({})", VERSION, &git_hash[..8]).into(),
         None => VERSION.into(),
     };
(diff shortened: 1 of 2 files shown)


the-mikedavis (dismissed):

dead10ck (changes requested):
Oh shoot, I had stumbled across this myself a long time ago and then totally forgot to make a PR! I think we also have to rerun if basically any file changes inside `runtime/grammars/sources`, or the `languages.toml` file too.

pascalkuthe:
> 

I looked into adding this. We are already rebuilding on `languages.toml` changes since we `include_bytes!` those which is a compiler builtin that handles invalidation correctly.

I added the metadata for `runtime/grammars/sources`. In the process I noticed that the simple approach I had taken to git metadata before didn't quite work and I didn't consider that HEAD is (usually) a git symbolic-ref and that `.git` can also be a reference (for detached work trees which I am even using myself). Considering that I heavily contribute to gitoxide I really should know better but it slipped me by. The new solution is much more robust. Sorry about that

helix-loader/build.rs
```diff
@@ -23,4 +24,40 @@ fn main() {
...
+    // from command output so for now we can't accept non-utf8 paths here
+    // probably rare enouch where it doesn't matter tough we could use gitoxide
+    // here but that would be make it a hard dependency and slow compile times
+    let Some(git_dir): Option<String> = Command::new("git")
+        .args(["rev-parse", "--git-dir"])
+        .output()
+        .ok()
+        .filter(|output| output.status.success())
+        .and_then(|x| String::from_utf8(x.stdout).ok())
+    else{ return; };
```
dead10ck, line 43: Does `cargo fmt` work on build scripts?

dead10ck (approved):
Ahh, that makes sense, thanks! This looks good to me.

helix-loader/build.rs
pascalkuthe: yes but rustfmt currently ignores `let .. else` (leaves it unchanged) altough that will hopefully be fixed in the not-so-distant future

the-mikedavis (approved):

Issue #6725, opened 2023-04-12
Helix 23.03 crashes when user tries to open it from suspended state (fg, bg)


HellConnector:
### Summary

When i need to suspend hx and go back to the terminal, i press `ctrl-z`, and helix suspended well. But if i execute `bg` and then `fg` in the terminal, helix just crashes.

<img width="912" alt="image" src="https://user-images.githubusercontent.com/50717671/231527100-0837efee-0769-471f-ba20-01716fcf9651.png">

### Reproduction Steps

 

I tried this:

1. `hx`
2. `ctrl-z`
3. `bg`
4. `fg`

I expected this to happen: `hx` opens normally at the state it was suspended

Instead, this happened:
<img width="912" alt="image" src="https://user-images.githubusercontent.com/50717671/231528082-aee1d00e-4ac9-46b2-858e-35a52855f058.png">

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Os { code: 4, kind: Interrupted, message: "Interrupted system call" }', /private/tmp/helix-20230331-6480-2cxefw/helix-term/src/application.rs:474:41
stack backtrace:
   0:        0x10ff427aa - <std::sys_common::backtrace::_print::DisplayBacktrace as core::fmt::Display>::fmt::h7ccb0a5911308a20
   1:        0x10f5bca7a - core::fmt::write::h8158d95376d3ffd4
   2:        0x10ff1d4cc - std::io::Write::write_fmt::h906cad35a9503f97
   3:        0x10ff42590 - std::sys_common::backtrace::print::h699856dda992f2e1
   4:        0x10ff244f3 - std::panicking::default_hook::{{closure}}::h66a03b3c8979fdca
   5:        0x10ff24205 - std::panicking::default_hook::hb24853726d3292d6
   6:        0x10ff24d07 - std::panicking::rust_panic_with_hook::h537956d5b9d077ec
   7:        0x10ff43975 - std::panicking::begin_panic_handler::{{closure}}::hf37433418b0d1181
   8:        0x10ff428f8 - std::sys_common::backtrace::__rust_end_short_backtrace::h56b0b14822285f07
   9:        0x10ff24862 - _rust_begin_unwind
  10:        0x1100149a3 - core::panicking::panic_fmt::h02cf2d7d99f4ebb0
  11:        0x110014e15 - core::result::unwrap_failed::h010baac391a0d5e7
[...]

PR #6764, opened 2023-04-15
Fix crash on opening from suspend state

h1t:
Fixes [#6725](https://github.com/helix-editor/helix/issues/6725)

This commit is copy/paste from same issue of neovim:
- [Issue](https://github.com/neovim/neovim/issues/12322)
 - [Fixes](https://github.com/neovim/neovim/pull/13084/commits/2e70fee1e9da6ed044b207e9ec03bb5950846ef1)

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -473,11 +473,9 @@ impl Application {
             signal::SIGCONT => {
                 // Copy/Paste from same issue from neovim:
                 // https://github.com/neovim/neovim/issues/12322
-                let mut retry_count = 10;
-                while retry_count > 0 {
-                    match self.claim_term().await {
-                        Ok(_) => break,
-                        Err(_) => retry_count -= 1,
+                for _ in 1..10 {
+                    if self.claim_term().await.is_ok() {
+                        break;
                     }
                 }
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -474,15 +474,12 @@ impl Application {
                 // Copy/Paste from same issue from neovim:
                 // https://github.com/neovim/neovim/issues/12322
                 // https://github.com/neovim/neovim/pull/13084
-                let mut retry_count = 10;
-                while retry_count > 0 {
-                    if self.claim_term().await.is_ok() {
-                        break;
+                for retries in 1..=10 {
+                    match self.claim_term().await {
+                        Ok(()) => break,
+                        Err(err) if retries == 10 => panic!("Failed to claim terminal: {}", err),
+                        Err(_) => continue,
                     }
-                    retry_count -= 1;
-                }
-                if retry_count == 0 {
-                    panic!("couldn't set terminal raw mode");
                 }
 
                 // redraw the terminal
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -471,7 +471,16 @@ impl Application {
                 }
             }
             signal::SIGCONT => {
-                self.claim_term().await.unwrap();
+                // Copy/Paste from same issue from neovim:
+                // https://github.com/neovim/neovim/issues/12322
+                let mut retry_count = 10;
+                while retry_count > 0 {
+                    match self.claim_term().await {
+                        Ok(_) => break,
+                        Err(_) => retry_count -= 1,
+                    }
+                }
+
                 // redraw the terminal
                 let area = self.terminal.size().expect("couldn't get terminal size");
                 self.compositor.resize(area);
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -440,8 +440,6 @@ impl Application {
 
     #[cfg(not(windows))]
     pub async fn handle_signals(&mut self, signal: i32) {
-        use core::panic;
-
         match signal {
             signal::SIGTSTP => {
                 self.restore_term().unwrap();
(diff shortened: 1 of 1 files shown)


helix-term/src/application.rs
```diff
@@ -471,7 +471,16 @@ impl Application {
                 }
             }
             signal::SIGCONT => {
-                self.claim_term().await.unwrap();
+                // Copy/Paste from same issue from neovim:
+                // https://github.com/neovim/neovim/issues/12322
+                let mut retry_count = 10;
+                while retry_count > 0 {
```
kareigu, line 477: Instead of copy-pasting this should probably be made more Rusty, something along the lines of
```rs
for _ in 0..10 {
    if let Ok(_) = self.claim_term().await {
        break;
    }
}
```

helix-term/src/application.rs
```diff
@@ -471,7 +471,14 @@ impl Application {
                 }
             }
             signal::SIGCONT => {
-                self.claim_term().await.unwrap();
+                // Copy/Paste from same issue from neovim:
+                // https://github.com/neovim/neovim/issues/12322
```
pascalkuthe, line 475: I 
```suggestion
                // https://github.com/neovim/neovim/pull/13084
```

we should link the PR instead of the issue here. I was confused at first because there were other patches mentioned in the issue. The PR back links to the issue so it's easy to discover too but it took a bit of digging for me to find the PR

pascalkuthe:
also shouldn't we still panic if `claim_term` fails all retries? Currently we just silently continue in that case which we really shouldn't

pascalkuthe (approved):
LGTM

helix-term/src/application.rs
```diff
@@ -471,7 +473,20 @@ impl Application {
...
+                let mut retry_count = 10;
+                while retry_count > 0 {
+                    if self.claim_term().await.is_ok() {
+                        break;
+                    }
+                    retry_count -= 1;
+                }
+                if retry_count == 0 {
+                    panic!("couldn't set terminal raw mode");
+                }
```
maxchisto, line 488: consider simplifying the whole thing to the following:
```rust
let retry_count = 10;
let _ = (0..retry_count).find(|_| self.claim_term().await.is_ok()).unwrap_or_else(|| {
    panic!("Couldn't set terminal raw mode");
});
```
`find()` returns when it encounters `true` for the first time

pascalkuthe: I don't think the functional style is clearer here. Abusing unwrap_or_else to handle the error case is really unidiomatic to me and `find` doesn't really seem intuitive for retries. The while loop is honestly quite readable here. The `rety_count == 0` check is quite intuitive. We could rename the variable `retries_left` to make it even clearer. If you desperately wanted to get rid of the while loop you could do:
``` rust
let mut retries_left = 10;
loop{
	if self.claim_term.await.is_ok(){
		break
	}
	retries_left -= 1;
	if retries_left == 0{
    	panic!("couldn't set terminal raw mode");
	}

}
```

but both versions are fine IMO

h1t: I'm newbie in Rust and I tried to find a solution using functional style. But I'm stuck trying to use async closure in iterator.

For example this code:
```rust
let retry_count = 10;
let _ = (0..retry_count).find(|_| self.claim_term().await.is_ok()).unwrap_or_else(|| {
    panic!("Couldn't set terminal raw mode");
});```

does not compile:
```
error[E0728]: `await` is only allowed inside `async` functions and blocks
   --> helix-term/src/application.rs:481:48
    |
481 |                     .find(|_| self.claim_term().await.is_ok())
    |                           ---                  ^^^^^^ only allowed inside `async` functions and blocks
    |                           |
    |                           this is not `async`
    ```

kareigu: You'd need to have an async closure there with 
```rs
async |_| self.claim_term().await.is_ok()
```
but as those are currently unstable, it would need some extra trickery to get working, which would be way more trouble than it's worth.

Even if the looping approach is a bit messy due to having the extra mutable variable, it's probably the simplest way to do this while having the panic functionality in place. 

Pascal's approach could be combined with what I originally suggested to be more concise
```rs
for retries in 1..=10 {
    if self.claim_term.await.is_ok(){
        break
    }
    if retries == 10 {
        panic!("couldn't set terminal raw mode");
    }
}
```
But with that you're going to be doing that second if check every iteration, even though it's redundant for 9 of those.

helix-term/src/application.rs
```diff
@@ -440,6 +440,8 @@ impl Application {
 
     #[cfg(not(windows))]
     pub async fn handle_signals(&mut self, signal: i32) {
+        use core::panic;
+
```
the-mikedavis, line 444:
```suggestion
```

this important isn't necessary

helix-term/src/application.rs
```diff
@@ -471,7 +473,20 @@ impl Application {
...
+                // https://github.com/neovim/neovim/pull/13084
+                let mut retry_count = 10;
+                while retry_count > 0 {
+                    if self.claim_term().await.is_ok() {
+                        break;
+                    }
+                    retry_count -= 1;
+                }
+                if retry_count == 0 {
+                    panic!("couldn't set terminal raw mode");
```
the-mikedavis, line 487: It could be useful to show the Err from claim_term. Also, claim_term can fail in other ways than not being able to enter raw mode so the message here should be a little more generic:

```rust
match self.claim_term().await {
    Ok(()) => break,
    Err(err) if retries == 10 => panic!("Failed to claim terminal: {}", err),
    Err(_) => continue,
}
```

I don't have strong opinions on `while` vs. `loop` vs. `for` here

the-mikedavis (approved):

pascalkuthe (approved):

PR #6766, opened 2023-04-15
Change Odin grammar to `ap29600/tree-sitter-odin`

ap29600:
The previously adopted grammar, `MineBill/tree-sitter-odin`, is unmaintained and mentions my repository as an alternative source.

### DIFF
--- a/runtime/queries/odin/highlights.scm
+++ b/runtime/queries/odin/highlights.scm
@@ -1,8 +1,8 @@
 (keyword) @keyword
 (operator) @operator
 
-(int_literal)   @constant.numeric
-(float_literal) @constant.numeric
+(int_literal)   @constant.numeric.integer
+(float_literal) @constant.numeric.float
 (rune_literal)  @string.character
 (bool_literal) @constant.builtin.boolean
 (nil) @constant.builtin
--- a/runtime/queries/odin/highlights.scm
+++ b/runtime/queries/odin/highlights.scm
@@ -3,7 +3,7 @@
 
 (int_literal)   @constant.numeric.integer
 (float_literal) @constant.numeric.float
-(rune_literal)  @string.character
+(rune_literal)  @constant.character
 (bool_literal) @constant.builtin.boolean
 (nil) @constant.builtin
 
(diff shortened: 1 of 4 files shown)


ap29600:
I noticed the CI was failing because I forgot to update the queries folder, sorry!

the-mikedavis (commented):
Are the highlights and indents copied over from the new repository? They are in neovim format and need the captures adjusted to the ones Helix uses (https://docs.helix-editor.com/master/themes.html#syntax-highlighting) and the indents should use the Helix style for indentation https://docs.helix-editor.com/master/guides/indent.html

ap29600:
thank you, I updated those. I didn't realize the format of the queries was editor-specific.

runtime/queries/odin/highlights.scm
```diff
@@ -1,141 +1,29 @@
...
+(float_literal) @constant.numeric
+(rune_literal)  @string.character
+(bool_literal) @constant.builtin.boolean
+(nil) @constant.builtin
 
-(call_expression
-  function: (selector_expression
-    field: (field_identifier) @function))
 
+(ERROR) @special
```
the-mikedavis, line 12: Eventually we want to make the ERROR highlight optional with config since the ERROR node exists for all languages (we could append it based on the config option). For now though we can remove this pattern

ap29600: Perfect, thank you for the help!

runtime/queries/odin/highlights.scm
```diff
@@ -1,141 +1,26 @@
...
-  "union"
-  "enum"
-  "switch"
-  "dynamic"
-] @keyword
-
-; ; Literals
+(int_literal)   @constant.numeric.integer
+(float_literal) @constant.numeric.float
+(rune_literal)  @string.character
```
the-mikedavis, line 6:
```suggestion
(rune_literal)  @constant.character
```

the-mikedavis (approved):

pascalkuthe (approved):

PR #6767, opened 2023-04-15
Another dracula refactoring

Chirikumbrah:
Sat_15 04 2023_19-35
Sat_15 04 2023_21-01

I changed the colors according to the  official Dracula specifications and vscode theme and I hope u'll like it :)
As I said earlier, I'm open to any ideas :)

### DIFF
--- a/runtime/themes/dracula.toml
+++ b/runtime/themes/dracula.toml
@@ -1,75 +1,65 @@
 # Author : Sebastian Zivota <loewenheim@mailbox.org>
 # Author : Chirikumbrah
 
-"annotation"                = { fg = "white"                           }
-"attribute"                 = { fg = "cyan"                            }
-"comment"                   = { fg = "comment"                         }
-"constant"                  = { fg = "purple"                          }
-"constant.numeric"          = { fg = "orange"                          }
-"constant.builtin"          = { fg = "orange"                          }
-"constant.character.escape" = { fg = "pink"                            }
-"constant.macro"            = { fg = "purple"                          }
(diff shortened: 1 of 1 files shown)


the-mikedavis:
cc @loewenheim what do you think?

loewenheim:
Sorry, I'm a little lost. Which colors is this changing to agree with the spec? Variable colors are still different.

Chirikumbrah:
@loewenheim As you can see, the color of the variables is now white.
image
>
Builtins (variable.builtin) are purple and italic, according to VS Code config.
image
>
Function parameters (variable.parameter) are orange and italic, according to VS Code config and dracula specs.
image
>
I made 'variable.other' cyan, but didn't see any changes and left it as is for now.
The 'variable.other.member' now purple because I couldn't find info about the color of this option, so you suggest me something about that and I'll make the change :)

loewenheim:
Right, I see. Thank you for the explanation. To be honest I would leave members as `foreground`, the same as other variables. Apart from that, this looks good :)

Chirikumbrah:
@loewenheim No problems)

loewenheim (approved):

truongvan:
I love your work, but I prefer `ui.menu` is more contrast with default background.

Screenshot 2023-04-23 at 10 30 43

Chirikumbrah:
@truongvan, thank you so much :)
U can see desired changes in my new pull request)

mo8it:
@Chirikumbrah Thank you :)

PR #6768, opened 2023-04-15
Fix YAML auto indent

### LINKED ISSUE #6661: YAML indentation
### Summary

It seems the auto-indent doesn't work for YAML.

I would expect helix to indent when a line containing `:` and keep the same indent as the previous line otherwise.

### Reproduction Steps

I tried this:

1. `hx file.yml`
2. Open a new line.

I expected this to happen:
the new line should be indented.

Instead, this happened:
the new line starts at character 0.

### Helix log

```
2023-04-08T15:51:30.050 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
```

</details>

### Platform

Linux

### Terminal Emulator

alacritty 0.12.0 (5a728195)

### Helix Version

helix 23.03

dead10ck:
This makes use of the new `extend` capture to correct YAML indentation.

Fixes #6661


pascalkuthe (approved):
I am not an expert on ident queries but this seems good to me Really awesome that you added a bunch of tests. This doesn't just test the YAML queries but also the indent system in general :+1:

dead10ck:
Yeah this actually took me a few days of just digging around in the indent code to figure out how it even works and where the captures are supposed to go, and what they are supposed to cover. I was convinced there was some kind of bug and added a whole bunch of debug prints, and then figured out the hard way how it works.

It seems needlessly complex, but I don't say that with a high level of confidence. Maybe I should go over the docs now that I'm not completely clueless about how it works.

dead10ck:
@pascalkuthe sorry, I added some more tests for inserting above, and also for flow style strings.

pascalkuthe (dismissed):

dead10ck:
Actually I just discovered another broken case: lists of maps

dead10ck:
Ok, I'm thinking actually there will need to be a code change to fix this. Consider this really simple case, where the `|` represents a cursor, and we want to insert a line below with `o`:

```yaml
- top:|
    baz: foo
    bax: foox
```

I think I have an ok enough understanding now of the indent system to say with reasonable confidence that this set of indent queries should cover what we want.

```scheme
(block_scalar) @indent @extend

(block_mapping_pair) @indent

(block_sequence_item) @indent
```

The last two cover the two indent cases we want: one level for the list, and one for the mapping pair. The problem is the current code will intentionally combine all `@indent` captures that happen on the same line into one.

https://github.com/helix-editor/helix/blob/8839eb0af47b99511956d81a6dbb32e948d2076c/helix-core/src/indent.rs#L682

I assume this was meant to cover giving the ability to have multiple overlapping patterns without them stepping on each other's toes.

But in this case, the sequence and the mapping pair simultaneously both happen on the same line, and both introduce a new level of indentation.

I think maybe what we need is a capture group that will always add a new indent level, even if multiple of them happen on the same line. Maybe it could be called `@indent.always` or something.

pascalkuthe:
The indent code is not really my forte but sounds reasonable cc @Triton171 since he did a lot of work on indent queries

Triton171:
I agree that this needs a new capture type since currently indent can only ever increase by 1 from one line to the next. And you're right, this behavior is needed since in most languages e.g. multiple nested blocks starting on the same line only increase indent by 1. `@indent.always` sounds good and it should be relatively simple to implement.

The indent code can probably also be simplified a bit: For example, we could remove the `get_first_in_line` function and replace it by just checking if a node is preceeded only by whitespace. The behavior is different in theory but I've yet to see a situation where it makes a difference in practice. Also, I'm thinking about making the documentation a bit more extensive so hopefully people writing indent queries don't have to look at the code. Is there anything that is particularly unclear in the current docs?

dead10ck:
Thanks @Triton171 , glad to see I'm on the right track. I have some feedback for the docs, I just have to write it up and/or might just update it myself, since I will need to add the new capture type to the docs anyway.

dead10ck:
Ok, I think I got the code changes done. I added the new `indent.always` capture, and I have tests passing. I did end up adding a couple of other additional changes along the way:

* Added a new `#one-line?` predicate which tests whether the capture only spans a single line. This was necessary for YAML sequences, where we don't want to add an indent for an individual list item unless it spans more than one line.
* Slightly modified the position that the indent query starts from. Currently, it starts the query from the very end of the line, including the line feed. This was problematic in YAML because e.g. if we have
  
  ```yaml
  top:
    baz: foo
    bazi:
      more: yes
      why: b#[e|]#cause
    quux:
      - 1
      - 2
  bax: foox
  ```
  
  and hit `o`, it will currently start the query from the line feed *past* the `e`, and the smallest descendant for this is
  
  ```yaml
  top:
    #[baz: foo
    bazi:
      more: yes
      why: because
    quux:
      - 1
      - 2|]#
  bax: foox
  ```
  
  so it ends up skipping a whole lot of nodes and captures. To fix this, I changed it to start the query from one character behind the line feed.

At this point, I think I just need to update the docs.

archseer:
Let's get @Triton171's review here too

helix-core/src/indent.rs
```diff
@@ -281,27 +281,32 @@ impl Indentation {
...
     Indent,
+    IndentAlways,
     Outdent,
+    OutdentAlways,
 }
 impl IndentCaptureType {
     fn default_scope(&self) -> IndentScope {
         match self {
             IndentCaptureType::Indent => IndentScope::Tail,
+            IndentCaptureType::IndentAlways => IndentScope::Tail,
```
pascalkuthe, line 300: These match arms van be merged with OR patterns as above (there is a clippynlint for this, I thought it was on by default maybe we should turn it on)

helix-term/src/commands.rs
```diff
@@ -2934,27 +2934,40 @@ fn open(cx: &mut Context, open: Open) {
...
             (0, 0)
         } else {
             (
-                line_end_char_index(&doc.text().slice(..), new_line.saturating_sub(1)),
+                line_end_char_index(&text, line_num),
                 doc.line_ending.len_chars(),
             )
         };
 
+        let line_start_index = doc.text().line_to_char(line_num);
```
pascalkuthe, line 2951:
```suggestion
        let line_start_index = doc_text.line_to_char(line_num);
```

helix-term/src/commands.rs
```diff
@@ -2934,27 +2934,40 @@ fn open(cx: &mut Context, open: Open) {
...
+
+        let indent_query_pos =
+            if !Range::new(line_start_index, line_end_index).is_single_grapheme(text) {
```
pascalkuthe, line 2954: I think this would be more readable if you removed the inversion and swapped the if/else branches. Also a small optimization/making the code a bit clearer (to me atleast). Simply always calculate the prev grapheme boundary and simply check if it's the sane as the line start (if it is the same then then the line only contains the newline grapheme)

helix-term/src/commands.rs
```diff
@@ -2934,27 +2934,40 @@ fn open(cx: &mut Context, open: Open) {
...
+                graphemes::prev_grapheme_boundary(text, line_end_index)
```
pascalkuthe, line 2955: If I understood you correctly you are purposefully starting the capture at the start of the newline char instead of the end here, right? In that case a comment would be nice.

helix-core/src/test.rs
```diff
@@ -152,7 +152,7 @@ pub fn print(s: &str) -> (String, Selection) {
 pub fn plain<R: Into<Rope>>(s: R, selection: &Selection) -> String {
     let s = s.into();
     let primary = selection.primary_index();
-    let mut out = String::with_capacity(s.len_bytes() + 5 * selection.len());
+    let mut out = String::with_capacity(s.len_chars() + 5 * selection.len());
```
pascalkuthe, line 155: This looks unrelated and also incorrect. Strings only oeprste in terms of byte indecies so you should reserve bytes (it would also be more efficent and easier to read if you just initialized out to `s.to_string()` and avoided the copy but this is test code so I would just remove that from this PR

helix-core/src/indent.rs
```diff
@@ -337,6 +344,35 @@ fn query_indents(
     let mut indent_captures: HashMap<usize, Vec<IndentCapture>> = HashMap::new();
     let mut extend_captures: HashMap<usize, Vec<ExtendCapture>> = HashMap::new();
     cursor.set_byte_range(range);
+
+    enum End {
+        Start,
```
pascalkuthe, line 349: Not a huge fan of nested items like this. Can this just be a normal enum and a normal function instead of a closure? I only noticed one capture that could easily br another argument.

I personally also think that an enum makes the whole thing a bit confusing. The function is short enough where having two functions might be more readable.

helix-core/src/indent.rs
```diff
@@ -388,6 +417,23 @@ fn query_indents(
...
                 }
+                "one-line?" | "not-one-line?" => match pred.args.get(0) {
+                    Some(QueryPredicateArg::Capture(capture_idx)) => {
+                        let node = m.nodes_for_capture_index(*capture_idx).next();
+
+                        match node {
+                            Some(node) => {
+                                let (start_line, end_line) = (get_line_num(node, End::Start), get_line_num(node, End::End));
+                                let one_line = end_line == start_line;
+                                !(one_line ^ (pred.operator.as_ref() == "one-line?"))
```
pascalkuthe, line 428: This is a bit confusing. I think the fillkwng is clearer. (For bool XOR is just neq)
```suggestion
                                one_line != (pred.operator.as_ref() == "not-one-line?")
`

**Review by pascalkuthe — COMMENTED** [review]
I jkticed that `one-line?` checks if a capture is contained on a single line after the new line break is inerted. I think that makes sense but just to make sure: Is that intentuonal? If yes then that should probably be documented

File: helix-term/src/commands.rs (hunk 4 above)
Comment (dead10ck) [reply helix-term/src/commands.rs:2955]: No, I'm starting it at the last character in the line, before the line feed. This way, it is more likely to start with a deeper node, such as an identifier. Currently, it's starting the indent query from the line feed, which is often a big scope. I will add a comment.

File: helix-core/src/test.rs (hunk 5 above)
Comment (dead10ck) [reply helix-core/src/test.rs:155]: Oh I did not mean to check this in. Will remove.

**Comment by dead10ck** [comment]
Great comments Pascal, thanks for the review!

**Comment by dead10ck** [comment]
Also, yes, it's intentional that `one-line?` is only true if it is unaffected by a new line insertion. Otherwise, if you have e.g.

```yaml
foo: #[|bar]#
```

and you hit return, it would only query the current state of the text, determine that it's one line, and not indent as it should.

I will document this for sure.

Triton171:
Most of this looks good to me, it's also very nice to get some integration testing for the indent code. There's 2 things which I'm unsure about:

(1): Changing the indent position to the character before the line feed is probably not always what we want. For example it breaks indentation in this C code:
````c
int main() {
    if (cond)
        do_smth(); // Pressing `o` here would add a line with 2 indent levels (1 too many)
}
````
This is a fundamental difference between languages where trailing whitespace should be considered to be part of a node and languages where this is not the case. Instead of changing the position of indent computation, we can instead use `@extend` captures for YAML. The only difference to your approach would be that this also extends nodes to trailing blank lines (if they have sufficient indentation). If that causes issues, we could add something like an `@extend.line` capture which only extends a node up to (and including) the next newline.


(2): I would expect the `@indent.always` and `@outdent.always` captures to simply add/subtract 1 to the total indentation (no matter where they appear & what other captures are encountered on the same line). The implementation seems to do something slightly different: `@indent.always` is equivalent to `@indent` but we pretend that there is an additional newline at its position. This probably behaves similarly to what I described above, but it's not the same.

For example, if we encounter (from innermost to outermost node) `@outdent`, `@indent` `@indent.always` captures all on the same line and no other captures:
- With the behavior that I described, `@indent` and `@outdent` cancel out, so `@indent.always` gives a total indentation of 1
- In the implementation, `@indent.always` would be equivalent to `@indent` here (because there are no other captures after it), so everything cancels out and we get a total indentation of 0.

Is there some reason why we need this more complex behavior? It seems like it makes the indentation logic significantly harder to reason about, because now the order in which captures are encountered matters, even if they all apply to the same line.

**Comment by dead10ck** [comment]
@Triton171 Thanks for the review!

> (1): Changing the indent position to the character before the line feed is probably not always what we want. For example it breaks indentation in this C code:
> ````c
> int main() {
>     if (cond)
>         do_smth(); // Pressing `o` here would add a line with 2 indent levels (1 too many)
> }
> ````

Shoot you're right. It's actually worse than this: it doesn't indent at all. Nor does it indent when inserting a new line from the function declaration. I'll look into the `@extend` capture. Thanks for finding this.

> (2): I would expect the `@indent.always` and `@outdent.always` captures to simply add/subtract 1 to the total indentation (no matter where they appear & what other captures are encountered on the same line). The implementation seems to do something slightly different: `@indent.always` is equivalent to `@indent` but we pretend that there is an additional newline at its position. This probably behaves similarly to what I described above, but it's not the same.
> 
> For example, if we encounter (from innermost to outermost node) `@outdent`, `@indent` `@indent.always` captures all on the same line and no other captures:
> - With the behavior that I described, `@indent` and `@outdent` cancel out, so `@indent.always` gives a total indentation of 1
> - In the implementation, `@indent.always` would be equivalent to `@indent` here (because there are no other captures after it), so everything cancels out and we get a total indentation of 0.
> 
> Is there some reason why we need this more complex behavior? It seems like it makes the indentation logic significantly harder to reason about, because now the order in which captures are encountered matters, even if they all apply to the same line.

Hm, I actually tried this at first, just changing the `add*` functions to sum the indent levels to greater than one for `indent.always` capture types, but it did not work because it only considers captures whose nodes have parents on the preceding line, so they get skipped altogether.

Do you have any suggestions? And also, out of genuine ignorance, is having indent and outdent captures on the same line a thing that happens?

**Comment by dead10ck** [comment]
@Triton171 you were right, simply adding `@extend` to the captures solved that problem.

**Comment by Triton171** [comment]
> @Triton171 you were right, simply adding `@extend` to the captures solved that problem.

Nice!

> Hm, I actually tried this at first, just changing the `add*` functions to sum the indent levels to greater than one for `indent.always` capture types, but it did not work because it only considers captures whose nodes have parents on the preceding line, so they get skipped altogether.

That's weird, in theory the indent query should consider the captures of all nodes that contain the query position. The only exception I can think of is if the node starts on the line of the query and the `tail` scope is used. If that's the issue, we can add `(#set! "scope" "all")` somewhere in the pattern (or change the default scope of `@indent.always` to `IndentScope::All`). Do you know in which part of the indent computation the indent was discarded?

Just modifying the add* functions might not be enough though, because the `Indentation::add_line` function needs some way to distinguish between (multiple) `@indent` captures and an `@indent.always` capture. We might also need to add a field `additional_indent` to the `Indentation` struct, that contains "number of `@indent.always` captures - number of `@outdent.always` captures". Maybe there's a better way though.


> is having indent and outdent captures on the same line a thing that happens?

It's not frequent but it definitely happens. We once had an issue with Go indentation which was caused by us incorrectly handling 2 `@indent` and 1 `@outdent` capture on the same line (#1523).

Also, thanks for all the work you've put into this. I would never have thought that YAML of all languages would be tricky for our indentation system :D

**Comment by dead10ck** [comment]
> That's weird, in theory the indent query should consider the captures of all nodes that contain the query position. The only exception I can think of is if the node starts on the line of the query and the `tail` scope is used. If that's the issue, we can add `(#set! "scope" "all")` somewhere in the pattern (or change the default scope of `@indent.always` to `IndentScope::All`). Do you know in which part of the indent computation the indent was discarded?

You can see the logic here:

https://github.com/helix-editor/helix/blob/8839eb0af47b99511956d81a6dbb32e948d2076c/helix-core/src/indent.rs#L670-L694

It does indeed gather all the captures for every node, but it does not add the indentation to the result unless the parent node is on the previous line.

> Just modifying the add* functions might not be enough though, because the `Indentation::add_line` function needs some way to distinguish between (multiple) `@indent` captures and an `@indent.always` capture. We might also need to add a field `additional_indent` to the `Indentation` struct, that contains "number of `@indent.always` captures - number of `@outdent.always` captures". Maybe there's a better way though.

The first thing I tried was to change `add_capture` so that for regular indents, it only turns a 0 into a 1; otherwise, the value is left alone. For "always" capture types, it would just increment.

And then `add_line` would sum the two indents together, rather than just add 1.

I could add those changes back; I only removed them because it seemed redundant once I figured out the parent node issue, but from the sound of it, it may not be.

> It's not frequent but it definitely happens. We once had an issue with Go indentation which was caused by us incorrectly handling 2 `@indent` and 1 `@outdent` capture on the same line (#1523).

Hmm, ok, I'll use this issue as a reference and see if I can't come up with a test case. Thanks!

**Comment by dead10ck** [comment]
@Triton171 actually it turns out you are right yet again! I made a mistake the first time I tried this approach.

https://github.com/helix-editor/helix/blob/b4fe31cad61caa64ae735db104a4d510ad14c6d7/helix-core/src/indent.rs#L250-L259

I left these if guards in place, and only changed it to sum the other indentation. But I thought about this a bit more, and I'm realizing that this is how cancellation was handled—when *both* indent and outdent are > 0, nothing happens to `self`, i.e. the indent gets dropped.

However, with this new indent capture, it suffices to just always sum the indent and outdent of the other indentation. Now, the only time these can be > 1 is when an "always" capture is encountered. For the more common case of regular indents, if you happen to get an indentation that has both an indent and an outdent, they both get summed into `self`, and will get cancelled out during final rendering of the indentation through plain arithmetic, rather than conditional logic.

I was also mistaken about the indent levels getting dropped. I can see now that the indent sums are kept across iterations of the loop, until we cross a line boundary.

Thanks so much for your help figuring this out!

**Comment by Triton171** [comment]
> But I thought about this a bit more, and I'm realizing that this is how cancellation was handled—when _both_ indent and outdent are > 0, nothing happens to `self`, i.e. the indent gets dropped.

Yeah, a comment would probably have been helpful there, it's easy to overlook.


> However, with this new indent capture, it suffices to just always sum the indent and outdent of the other indentation. Now, the only time these can be > 1 is when an "always" capture is encountered. For the more common case of regular indents, if you happen to get an indentation that has both an indent and an outdent, they both get summed into `self`, and will get cancelled out during final rendering of the indentation through plain arithmetic, rather than conditional logic.

Does this really always do the right thing? If we first add an `@indent` and then an `@indent.always` capture, it works (the indent is first set to 1 and then increased to 2). But if we encounter them in a different order, i.e. first `@indent.always` and then `@indent`, the indent is first increased to 1 and is therefore not changed by the `@indent`. This again means that the order of captures on the same line matters (which we should probably try to avoid). It seems to me like we do need at least 1 additional field in the `Indentation` struct to make this consistent (although it would be nice to avoid it, if there is some elegant way to do that).

**Comment by dead10ck** [comment]
Hmm, that's a good point, it would be best if they were consistent. In my mind at least, I'm considering `@indent.always` just "another kind of `@indent`", so I think it makes the most sense to go with the latter case you mentioned, i.e. first add all the `@indent.always`, and then only if that's 0, add the single `@indent`

 I can't really think of another way to do this besides another two fields for the "always" variants. I thought of maybe just making a single field of `Vec<IndentCaptureType>` and making logic to combine them, but this doesn't seem any better.

I can make this change.

**Comment by Triton171** [comment]
> I can't really think of another way to do this besides another two fields for the "always" variants. I thought of maybe just making a single field of `Vec<IndentCaptureType>` and making logic to combine them, but this doesn't seem any better.

I think a single additional (signed) field should also be enough, since `@indent.always` and `@outdent.always` cancel each other, so we only need to store their difference. I don't think it matters much either way though; both versions would be fine imo.

**Comment by dead10ck** [comment]
Ok this has to be the weirdest thing I've ever seen in Rust. I changed `Indentation::add_capture` to return its own `&mut self` argument, so that you could chain calls to it.

https://github.com/helix-editor/helix/pull/6768/commits/f28077da0e5d793d2da7536b02a6d5170216df8a#diff-44db3a32fdd620ed51f6953534c6ed5df3d2966010cc86aaf13448e9bbb65f07R264

But this inexplicably causes indentation to break in some cases.

The only places this function is used is within calculating the indent level here:

https://github.com/helix-editor/helix/blob/29638babea410d6d8397b02e6a826de0076a69c0/helix-core/src/indent.rs#L658

We can see the reference returned is not used. But that ... should be fine, shouldn't it? How on Earth is this changing the result?

Anyway, I added the two new fields for tracking always captures, and added unit tests to verify the associativity. The tests are uglier than I would like because of the above, but they work now.

I opted for two separate fields because in my mind it's a bit more confusing to have one field be a count and one be a signed delta.

**Comment by dead10ck** [comment]
@Triton171 if you get some time, I would appreciate a review. ❤️ I think these implementation details will affect the docs, so I want to wait on code review before I spend time updating docs.

File: helix-core/src/indent.rs [hunk 8]
```diff
@@ -360,21 +414,14 @@ fn query_indents(
...
-                                node_line
-                            };
                             let n1 = m.nodes_for_capture_index(*capt1).next();
                             let n2 = m.nodes_for_capture_index(*capt2).next();
                             match (n1, n2) {
                                 (Some(n1), Some(n2)) => {
-                                    let same_line = get_line_num(n1)==get_line_num(n2);
+                                    let n1_line = get_node_start_line(n1);
+                                    let n2_line = get_node_start_line(n2);
+                                    log::debug!("n1 line: {}, n2 line: {}", n1_line, n2_line);
```
Triton171, line 423: Did you forget to remove this? Getting some debug logging for indent queries is not a bad idea in my opinion, but the text should be a bit more descriptive (maybe also just with `log::trace!` so it doesn't clutter the log too much).

Triton171:
Sorry for taking some time with the review, I had a busy few days. Apart from the 2 small things I pointed out above, all of this looks very good. Once those are resolved, this can be merged in my opinion.

> I opted for two separate fields because in my mind it's a bit more confusing to have one field be a count and one be a signed delta.

I agree, and it also has the advantage that we can always use `usize` for counting indents instead of using signed / unsigned types for different fields.



> But this inexplicably causes indentation to break in some cases.

I looked at [the diff](https://github.com/helix-editor/helix/pull/6768/commits/9dd8a2a0eafa96a56c29f30da63717b93dc097f7) in which you reverted that change and I don't understand either how this can change the behavior of indent queries. Since it only makes the testing code slightly cleaner, I guess it's not really worth the time to debug why this happens.

helix-core/src/indent.rs
```diff
@@ -237,38 +237,58 @@ fn get_first_in_line(mut node: Node, new_line_byte_pos: Option<usize>) -> Vec<bo
...
-                self.indent = 1;
+                if self.indent_always == 0 {
+                    self.indent = 1;
+                }
+            }
+            IndentCaptureType::IndentAlways => {
+                // any time we encounter an `indent.always` on the same line, we
+                // want to cancel out all regular indents
+                self.indent_always += 1;
+                self.indent = 0;
```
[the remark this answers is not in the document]
dead10ck: Yeah, it's because of when we're rendering the indent here:

https://github.com/helix-editor/helix/pull/6768/files#diff-44db3a32fdd620ed51f6953534c6ed5df3d2966010cc86aaf13448e9bbb65f07R287-R288

The issue is that during the indent capture search, we accumulate the indents across two dimensions:

* `add_capture` accumulates *within* lines; we want to ensure that all `indent.always` captures are always additive. `indent` captures within lines are more like a boolean—any indent capture encountered, no matter which type, means any further `indent` captures have no effect
* `add_line` accumulates *across* lines. The purpose is to combine indent levels of lines higher up in the syntax tree, so indent levels are always additive

In order to ensure both of these are true, `indent` captures are still stored as a `usize`, despite their behavior when accumulating within lines, because we need to be able to sum them across lines. But if we don't reset the `indent` field here, then we can't tell the difference between `indent` captures within and across lines when we render the final indent level.

The API is a bit confusing and hard to follow because it uses the same type and object instance to compute indents within and across lines, and the only thing ensuring that everything works is that it so happens that we only use the `add_line` and `add_capture` methods where we're supposed to.

But fixing that would need a larger refactor and I wasn't quite up for that.

helix-core/src/indent.rs
dead10ck: Oh yeah, I forgot to remove it. I'll take care of that.

dead10ck:
Thanks so much for the review @Triton171 ! I'll go ahead and work on the docs then.

helix-core/src/indent.rs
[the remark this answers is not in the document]
Triton171: Yeah, the code would definitely be easier to read if we had different types for the total indent and the indent for one line.

> But if we don't reset the `indent` field here, then we can't tell the difference between `indent` captures within and across lines when we render the final indent level.

Is this really true? Even if we don't reset it, the `add_capture` never increases the `indent` field beyond 1, so multiple `@indent` captures on the same line should still cancel out. The behavior should only be different when `@indent.always` and `@indent` appear on the same line. They could either stack (i.e. produce a total indent of 2) or the `@indent` could be ignored, producing a total indent of 1 (the current behavior).

From the queries, it's not obvious to me whether this can actually happen (since the only `@indent` capture is for the `block_scalar` node). If it doesn't, I think the first behavior makes more sense (but of course this is somewhat subjective).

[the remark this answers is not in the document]
dead10ck: Yeah, this was part of the discussion starting [here](https://github.com/helix-editor/helix/pull/6768#issuecomment-1595693901) that centered around what should happen when we encounter both on the same line; the problem was it could stack in one case and not the other, depending on the order they were encountered.

For consistency, we should choose one or the other. But which one is I think the issue at hand. The way I see it, `@indent.always` is just "another kind of `@indent`", so if we should combine multiple `@indent`s on the same line, then we should include `@indent.always` in that calculation.

But I can see how the other way would make sense too.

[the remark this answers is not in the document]
Triton171: > For consistency, we should choose one or the other. But which one is I think the issue at hand.

I definitely agree.



> But which one is I think the issue at hand. The way I see it, `@indent.always` is just "another kind of `@indent`", so if we should combine multiple `@indent`s on the same line, then we should include `@indent.always` in that calculation.

Yeah that makes sense. I think that to me `@indent.always` seemed like something separate from a regular `@indent`, but (especially considering the naming) your intuition is probably better. Thanks for explaining it to me.

dead10ck:
Ok, I've finished going over the docs. I added a bunch of examples that should hopefully help newcomers understand how everything works. These are details I found myself sorely missing when I was trying to put together how to make indent queries.

I think this pretty much wraps up the PR. It is fully ready for review.

helix-core/src/indent.rs
pascalkuthe: I think this still needs to ne adressed

pascalkuthe (approved):
This looks reall great. The does are really nicht!

PR #6773, opened 2023-04-16
Add `inlay-hint` colours to Kanagawa

kareigu:
Makes the LSP hints more legible against the background colour and more in line with the normal colours in the theme.


the-mikedavis:
cc @zetashift what do you think of these colors?

zetashift:
LGTM!

I haven't used neovim for a while now so I don't know if it's the same colors as the neovim one. But the hint and type hints seem to correspond to the original theme, so I am okay with this.

the-mikedavis (approved):

pascalkuthe (approved):

PR #6774, opened 2023-04-16
make :toggle-option print the new value

zummenix:
I think some kind of feedback is useful for this command. I chose to set the status line with format `key = new_value`. For example, `:toggle lsp.enable` will set the status line to `lsp.enable = false` or `lsp.enable = true` after successfully applying updated config to the editor.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1784,7 +1784,8 @@ fn toggle_option(
         .config_events
         .0
         .send(ConfigEvent::Update(config))?;
-    cx.editor.set_status(format!("{} = {}", key, new_value));
+    cx.editor
+        .set_status(format!("Option `{}` is now set to `{}`", key, new_value));
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


pascalkuthe (commented):
This seems like a good usability improvement to me. One minor style nit: I think this code could be clear if we just extracted the value with let .. else and then just did `let new_value = !old_value`.

zummenix:
@pascalkuthe Nice! Done.

pascalkuthe (approved):
LGTM :+1:

helix-term/src/commands/typed.rs
```diff
@@ -1784,6 +1784,7 @@ fn toggle_option(
         .config_events
         .0
         .send(ConfigEvent::Update(config))?;
+    cx.editor.set_status(format!("{} = {}", key, new_value));
```
the-mikedavis, line 1787: I think this would be clearer as text:

```suggestion
    cx.editor.set_status(format!("Option `{}` is now set to `{}`", key, new_value));
```

the-mikedavis (approved):

pascalkuthe (approved):

PR #6776, opened 2023-04-16
ferra: fix ruler overriding foreground style

dylrich:
Ferra's current ruler styling overrides the foreground style, which is an odd look only for characters in the ruler column. This commit removes the foreground styling for the ruler rule. This is more in line with what other themes do for the ruler.


the-mikedavis:
cc @casperstorm what do you think?

casperstorm:
> cc @casperstorm what do you think?

Great catch. Looks good to me.

the-mikedavis (approved):

pascalkuthe (approved):

Issue #5779, opened 2023-02-02
Completion on paths containing spaces overwrites characters


ninevra:
### Summary

When autocompleting a path in a command, the last `n` characters of the previous path segment are overwritten by the next path segment, where `n` is the number of escaped spaces in the previous path segment.

This overwrites the path separator character.

Usually, the resulting path doesn't actually exist; so if the user was expecting to open a file, they instead get a blank buffer associated with the mangled path; or if they were expecting to navigate a subdirectory, autocomplete instead ceases.

### Reproduction Steps

 

I tried this:

1. `mkdir test\ dir`
2. `touch test\ dir/file`
3. `hx`
4. `:o test<tab><tab>`

I expected this to happen:

Autocomplete would construct the command `:o test\ dir/file`. Executing the command would open `test\ dir/file`.

Instead, this happened:

Autocomplete generated `:o test\ dirfile`.

Executing this command opens a blank buffer at `test dirfile`, rather than `test dir/file`.

### Helix log

No log entries at any relevant timestamp.

### Platform

linux

### Terminal Emulator

gnome terminal

### Helix Version

helix 22.12 (0f2ae35a)

adwhit:
Looked into it a little bit. 

The problem is that escape characters are stripped here: https://github.com/helix-editor/helix/blob/05e7862e8fc34e2e4dd745f9dc95c1b1ab0dc20f/helix-term/src/commands/typed.rs#L2721, then this shorter string is used to create a modification range here: https://github.com/helix-editor/helix/blob/05e7862e8fc34e2e4dd745f9dc95c1b1ab0dc20f/helix-term/src/ui/mod.rs#L548. But then this range is used to modify to original, unescaped string here: https://github.com/helix-editor/helix/blob/05e7862e8fc34e2e4dd745f9dc95c1b1ab0dc20f/helix-term/src/ui/prompt.rs#L340

This is the first time I've looked at the codebase though so I can't say how best to fix it.

cgahr:
Can this be close? #6779 fixes this issue and seems to be merged.

PR #6779, opened 2023-04-16
Fix completion on paths containing spaces

ogimgd:
It solves #5779

Before:
render1681600147669

After:

render1681600980390

P.S. I fixed this by getting word_len for a range replacement from the actual word being displayed (instead of initial representation of a path from autocomp), hope that was an initial idea here.


ogimgd:
Fixed the issues with 'cargo fmt' :)

pascalkuthe (approved):
Good catch! Thanks!

PR #6792, opened 2023-04-17
always build grammars with c++14 and c11

### LINKED ISSUE #6788: Build failure on Cabal and Just grammars
### Summary

Getting build failures with the new Cabal (#6485) grammar and Just. Appears to be C++ related?

### Reproduction Steps

 

I tried this:

1. `hx --grammar build`

I expected this to happen:

Grammars built with no errors

Instead, this happened:

Errors building `just` and `cabal` grammars.

<details><summary>grammar build log</summary>

```
Building 149 grammars
147 grammars already built
2 grammars failed to build
        Failure 0/2: Parser compilation failed.
Stdout:
Stderr: /Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:15:24: warning: default member initializer for non-static data member is a C++11 extension [-Wc++11-extensions]
  uint32_t prev_indent = 0;
                       ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:47:3: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
  auto curr = std::to_string(state->prev_indent);
  ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:64:14: error: expected expression
    *state = {};
             ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:88:3: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
  auto advance = [lexer] { lexer->advance(lexer, false); };
  ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:88:18: error: expected expression
  auto advance = [lexer] { lexer->advance(lexer, false); };
                 ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:89:3: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
  auto skip = [lexer] { lexer->advance(lexer, true); };
  ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:89:15: error: expected expression
  auto skip = [lexer] { lexer->advance(lexer, true); };
              ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:91:3: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
  auto get_column = [lexer] { return lexer->get_column(lexer); };
  ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:91:21: error: expected expression
  auto get_column = [lexer] { return lexer->get_column(lexer); };
                    ^
/Users/bc/.config/helix/runtime/grammars/sources/just/src/scanner.cc:128:5: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
    auto indent = get_column();
    ^
6 warnings and 4 errors generated.

        Failure 1/2: Parser compilation failed.
Stdout:
Stderr: /Users/bc/.config/helix/runtime/grammars/sources/cabal/src/scanner.cc:26:60: error: expected expression
        for_each(indent_lvls.cbegin(), indent_lvls.cend(), [buffer, &i](uint8_t const n) { buffer[i++] = n; });
                                                           ^
/Users/bc/.config/helix/runtime/grammars/sources/cabal/src/scanner.cc:70:9: warning: 'auto' type specifier is a C++11 extension [-Wc++11-extensions]
        auto cur_indent_lvl = indent_lvls.back();
        ^
1 warning and 1 error generated.
```
</details>

### Helix log

```
```

</details>

### Platform

macOS

### Terminal Emulator

Alacritty

### Helix Version

helix 23.03 (4cdba7cc)

pascalkuthe:
Closes #6788

Usually, modern compilers always default to c++ 14 or c++17 anyway. Clang defaults to c++14 and gcc to c++17. MSVC++ even dropped support for older standards in msvc17 (and defaults to c++14) However, it seems that apple-clang sometimes (but not always) defaults to c++98. By explicitly setting the argument we can work around such cases.

This also ensures more consistent builds across various systems (that may default to newer c++ standards in the future) and provides more easily understood error messages on ancient systems that don't support c++14 yet.

### DIFF
--- a/helix-loader/src/grammar.rs
+++ b/helix-loader/src/grammar.rs
@@ -8,6 +8,7 @@ use std::{
     process::Command,
     sync::mpsc::channel,
 };
+use tempfile::TempPath;
 use tree_sitter::Language;
 
 #[cfg(unix)]
@@ -446,13 +447,17 @@ fn build_tree_sitter_library(
         command.env(key, value);
     }
     command.args(compiler.args());
+    // used to delay dropping the temporary object file until after the compilation is complete
+    let _path_guard;
 
     if cfg!(all(windows, target_env = "msvc")) {
         command
             .args(["/nologo", "/LD", "/I"])
             .arg(header_path)
             .arg("/Od")
-            .arg("/utf-8");
+            .arg("/utf-8")
+            .arg("/std:c++14")
+            .arg("/std:c11");
         if let Some(scanner_path) = scanner_path.as_ref() {
             command.arg(scanner_path);
         }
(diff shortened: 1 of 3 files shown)


helix-loader/src/grammar.rs
```diff
@@ -471,6 +472,7 @@ fn build_tree_sitter_library(
             .arg(header_path)
             .arg("-o")
             .arg(&library_path)
+            .arg("-std=c++14")
```
the-mikedavis, line 475: Will the `.arg("-std=c99")` below conflict with this? Should that be removed?

David-Else:
Sorry if this is not strictly related to this PR, but on Linux (I have not tested on another OS), if there is no c++ compiler installed when using `cargo install --path helix-term --locked` Helix silently fails to compile the grammar  and shows no highlighting, perhaps that could also be addressed with an error message?

helix-loader/src/grammar.rs
pascalkuthe: For clang and msvbc the C standard and C++ standard can be set separately when using a c++ compiler.  This makes sense since these two are really decoupled but it seems some versions of GCC are not as flexible.

 I amended to commit to always use c11 for c files and c++14 for c++ files keeping the existing file based detection in tact (and setting the appropriate standard) to (both on Windows and UNIX) to again be consistent across various toolchains and just moved the whole thing out of the if. The special handling for file extensions is actually not possible compilers should do that automatically but I left it in for now.

David-Else:
>  Should probably be a separate PR but not sure if we want to change the formatting a bit.

> Anyway, this patch fixes what you described:

@pascalkuthe Great news you fixed it so fast! I have no opinion on the formatting, but I hope the fix can get into `23.03.1` https://github.com/helix-editor/helix/milestone/5 as it has caught me out a few times, and I have also seen people run into this problem in the issues. Silent failure is the worst kind of fail.

pascalkuthe:
@David-Else  #6795

pascalkuthe:
It turns out that the grammar build was actually failing in CI too but because of the bug pointed out by @David-Else it didn't cause a build failure. With #6795 the macros build with master fails in CI but succedes with this branch so this is actually reproducible :D

archseer:
Hmm I remember we used to build with C++14 because the haskell grammar had more complicated requirements but that failed to build on some systems. Eventually the haskell grammar was simplified and we dropped the requirement.

archseer:
That was probably because the cross images used a very old compiler, I think that has since been resolved

pascalkuthe:
If its really an issue in the future we can lower the c++ version down to something more compatible and patch the grammars. c++14 support is pretty widespread these days so I hope it won't be an issue. The only distro wehrre this may cause problems that is still supported is redhl 7 based stuff and that's nearing EOL and has an easy way to install a newer compiler trough the support packagss.

Fxing the exact c standard version probably a good idea regardless even if we would habe to lower the version again to make sure builds are consistent across different systems

PR #6793, opened 2023-04-17
Highlight sqlx's `query_scalar{,_unchecked}` macros as SQL

cole-h:
I noticed these two were missing, so I figured we should add them for completeness.

### DIFF
--- a/runtime/queries/rust/injections.scm
+++ b/runtime/queries/rust/injections.scm
@@ -41,7 +41,7 @@
 (macro_invocation
   macro: (scoped_identifier
     path: (identifier) @_sqlx (#eq? @_sqlx "sqlx")
-    name: (identifier) @_query_as (#match? @_query "^query_as(_unchecked)?$"))
+    name: (identifier) @_query_as (#match? @_query_as "^query_as(_unchecked)?$"))
   (token_tree
     ; Only the second argument is SQL
     .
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
I didn't notice the duplication before but I think we should collapse these patterns where "Only the first argument is SQL" into one pattern and change the `(#eq? @_query "query")` part to `(#match? @_query "^query(_as|_scalar|_scalar_unchecked)?$")` to match them all more concisely

cole-h:
Sounds like a good idea.

However, my lisp skills are fairly low, and I don't currently have the time (or desire) to improve them to make that change. Would you be opposed to merging this as-is and then having someone file a follow-up PR that does improve this situation?

archseer:
It seems more effort to open yet another PR than to just do the replacement @the-mikedavis gave you? Remove other queries, then replace `(#eq? @_query "query")` with `(#match? @_query "^query(_as|_scalar|_scalar_unchecked)?$")` should be possible to do without having to understand lisp :)

cole-h:
Oh, sorry! That shows my lisp knowledge -- it sounded more involved than some simple replacements 😅 

I've hopefully addressed that in 041b116cab4030ce014773c51e457000185f544e.

the-mikedavis (approved):

PR #6795, opened 2023-04-17
produce error when grammar build fails

pascalkuthe:
Fixes the issue described in https://github.com/helix-editor/helix/pull/6792#issuecomment-1512058559.

Currently, grammar builds fail without actually aborting the build (even tough that seems to be the intention since we explicitly call `except`) because we only print to stdout when the build fails but don't actually return an error. This makes diagnosing and fixing build failures harder. This PR fixes that by simply calling `bail!` instead of `println!`. As described here https://github.com/helix-editor/helix/pull/6792#issuecomment-1512080841 this slightly changes the formatting of the error message. I have therefore opted to remove the indentation from the failure message as that now looks a bit out of place

### DIFF
--- a/helix-loader/src/grammar.rs
+++ b/helix-loader/src/grammar.rs
@@ -1,4 +1,4 @@
-use anyhow::{anyhow, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use serde::{Deserialize, Serialize};
 use std::fs;
 use std::time::SystemTime;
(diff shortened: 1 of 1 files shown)


pascalkuthe:
Interesting with this patch applied the CI actually fails on macos with https://github.com/helix-editor/helix/issues/6788 so this also caused that issue to slip trough :D

archseer:
With the other PR merged, can you rebase this one?

pascalkuthe:
Will do that tomorrow to.ing was a bit unfortunate just turned off my PC for today :D 

I had the other branch merged into this branch to be able to check that the build really passes CI, I guess I should have just made this a single PR

pascalkuthe:
(done)

PR #6802, opened 2023-04-18
feat: Allow setting a menu name from config file

tanj:
While it was possible to create sub-menus or other menus from a configuration file it was not possible to give those menus a name. This change lets you specify a name for the menu.


the-mikedavis:
This is also added by https://github.com/helix-editor/helix/pull/3958

PR #6808, opened 2023-04-18
Run cargo update

### LINKED ISSUE #6806: warning: package `windows v0.47.0` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
I just got this error:

```
warning: package `windows v0.47.0` in Cargo.lock is yanked in registry `crates-io`, consider running without --locked
```

When using `cargo install --locked --path helix-term`, I am not sure if this needs some kind of attention?

pascalkuthe:
Closes #6806

The windows v0.47.0 was [yanked](https://github.com/microsoft/windows-rs/pull/2412) recently (apparently broke some builds but didn't affect us) so I ran `cargo update` to fix that. It seems that depandabot doesn't really update indirect dependencies with `cargo update`, so there are a bunch more updates here.

I have reviewed all commits in the dependencies to avoid supply chain attacks (although all these crates are published by people I trust). I extracted any changes that are interesting for helix

* `core-foundation-sys` (macos): remove build script in favor of conditional link attribute (slightly faster build times)
* `syn`: -
* `gix-hash` / `gix-hashtable`: Fix bug where hashing an `oid` instead of an `ObjectId` caused a panic on `32bit` systems. Doesn't affect helix currently but would have with future stuff I had planned.
* `gix-revision`: correctly parse revisions that contain `@`. Doesn't affect helix currently but would have with future stuff I had planned.
* `filetime` `windows-sys` bump
* `rustix`: -
* `getrandom` :
* `errno` -
* `proc-macro2`: -
* `windows`: changes to the way files are build. Apparently this is more reliable but it did mean we had to clear build caches (CI failed otherwise)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -197,9 +197,9 @@ dependencies = [
 
 [[package]]
 name = "core-foundation-sys"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5827cebf4670468b8772dd191856768aedcb1b0278a04f989f7766351917b9dc"
+checksum = "e496a50fda8aacccc86d7529e2c1e0892dbd0f898a6b5645b5561b89c3210efa"
 
 [[package]]
 name = "crc32fast"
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I had to change the build changes for windows to make the CI pass because the `windows` update changed something about the way they link to stuff I think. The error was very non-descript (documentation pretty much said "anything can cause this") but the Microsoft documentation suggested clearing caches so I did that and it worked :D

PR #6809, opened 2023-04-18
Improve and update the installation documentation

David-Else:
Before the release of 23.03.1, I wanted to address a few issues that have been bothering me and make some general improvements. 

Primarily, I corrected grammar and clarity issues, but I also added a new requirements section. Initially, we had agreed that this section was unnecessary as everyone knew that the Rust toolchain was mandatory. However, since a c++14 requirement was recently added, I included the Rust toolchain requirement in the new section as well. I think it is much better now as I also added the `If you are using the musl-libc...` section to the requirements up front, it made no sense to have it after you had tried to install Helix as it would have failed by then!

I think there was some debate about numbering `Clone the repository` and `compile from source`, but with the new requirements things seemed a bit harder to follow without them.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -41,8 +41,6 @@ line.
 
 ## Linux, macOS, Windows and OpenBSD packaging status
 
-Helix is available for Linux, macOS and Windows via the official repositories listed below.
-
 [![Packaging status](https://repology.org/badge/vertical-allrepos/helix.svg)](https://repology.org/project/helix/versions)
 
 ## Linux
@@ -51,7 +49,7 @@ The following third party repositories are available:
 
 ### Ubuntu
 
-Helix is available via [Maveonair's PPA](https://launchpad.net/~maveonair/+archive/ubuntu/helix-editor):
+Add the `PPA` for Helix:
 
 ```sh
 sudo add-apt-repository ppa:maveonair/helix-editor
@@ -61,7 +59,7 @@ sudo apt install helix
 
 ### Fedora/RHEL
 
-Helix is available via `copr`:
+Enable the `COPR` repository for Helix:
 
 ```sh
 sudo dnf copr enable varlad/helix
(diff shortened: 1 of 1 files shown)


book/src/install.md
```diff
@@ -143,33 +141,35 @@ pacman -S mingw-w64-ucrt-x86_64-helix
 
 ## Building from source
 
-Clone the repository:
+Requirements:
+
+- The [Rust toolchain](https://www.rust-lang.org/tools/install)
+- A c++14 compatible compiler to build the tree-sitter grammars, for example `gcc-c++`
```
the-mikedavis, line 147: Is `gcc-c++` a debian package? I don't think the reference makes sense for other OSs, for example Arch or macos through brew. We might be able to say something generic like "GCC or Clang" if `gcc-c++` is just g++

David-Else: Yes much better to make it generic. `gcc-c++` was just the package I used:

```
Installed Packages
Name         : gcc-c++
Version      : 11.3.1
Release      : 2.1.el9.alma
Architecture : x86_64
Size         : 32 M
Source       : gcc-11.3.1-2.1.el9.alma.src.rpm
Repository   : @System
From repo    : appstream
Summary      : C++ support for GCC
URL          : http://gcc.gnu.org
License      : GPLv3+ and GPLv3+ with exceptions and GPLv2+ with exceptions and LGPLv2+ and BSD
Description  : This package adds C++ support to the GNU Compiler Collection.
             : It includes support for most of the current C++ specification,
             : including templates and exception handling.
```

book/src/install.md
```diff
@@ -143,33 +141,35 @@ pacman -S mingw-w64-ucrt-x86_64-helix
 
 ## Building from source
 
-Clone the repository:
+Requirements:
```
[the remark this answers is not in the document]
David-Else: Good catch, done.

the-mikedavis (approved):

pascalkuthe (approved):

PR #5713, opened 2023-01-28
Fix Go select indenting

gavincrawford:
Resolves #5610
Select statements previously were incorrectly indented.
Before this PR:
https://asciinema.org/a/VnOgUThOMTeK3gP2nRr5YQmTt
After this PR:
https://asciinema.org/a/pPAl4ghHjTp6TCiIy4yLIeyUf

PR #6817, opened 2023-04-19
Avoid extra indentation on Go switches

### LINKED ISSUE #6772: Extra indentation in Go switches
### Summary

Go switches by themselves should not indent their contents; it's only the bodies of each `case` which are indented. It seems like Helix wants to indent both, meaning that the body of each `case` ends up with one extra level.

### Reproduction Steps

Use `o` to insert a new line right after a `switch` or `case` line in Go. The new line is indented one level more than desired.

https://asciinema.org/a/xklzt1tRNliVp4SM02BmuR11Z

I believe the problem is that [we indent switches like other languages](https://github.com/helix-editor/helix/blob/4cdba7cccdd6a9fe5fbb3d15c9db3edc147b7a90/runtime/queries/go/indents.scm#L18-L19), when in fact Go doesn't indent the entire switch block like other languages.

I suspect that the fix in https://github.com/helix-editor/helix/pull/5713 should be made to work for `switch` as well as `select`, and I'm happy to try to send a PR if that works.

Is there a way for me to edit and test a change to that `indents.scm` file without having to build all of Helix from source? Can I copy the file into `~/.config/helix` somewhere, make the change, and test it to send a PR?

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

foot 1.14.0

### Helix Version

helix 23.03

mvdan:
(see commit message)

Fixes #6772.

### DIFF
--- a/runtime/queries/go/indents.scm
+++ b/runtime/queries/go/indents.scm
@@ -14,8 +14,6 @@
   (argument_list)
   (field_declaration_list)
   (block)
-  (type_switch_statement)
-  (expression_switch_statement)
   (var_declaration)
 ] @indent
 
@@ -24,5 +22,19 @@
   ")"
 ] @outdent
 
-((_ "}" @outdent) @outer (#not-kind-eq? @outer "select_statement"))
-(communication_case) @extend
+; Switches and selects aren't indented, only their case bodies are.
+; Outdent all closing braces except those closing switches or selects.
+(
+    (_ "}" @outdent) @outer
+    (#not-kind-eq? @outer "select_statement")
+    (#not-kind-eq? @outer "type_switch_statement")
+    (#not-kind-eq? @outer "expression_switch_statement")
+)
+
+; Starting a line after a new case should indent.
+[
+  (communication_case)
+  (expression_case)
+  (default_case)
+  (type_case)
+] @extend


the-mikedavis (approved):

pascalkuthe:
@Triton171  I am not an expert on the indent queries but from your comment on #6772 it sounds like you had something slightly different in mind.

Triton171:
Yeah, this is not what I meant. I believe this issue needs essentially the same fix as #5713. This PR currently breaks go indentation in some ways (for example since the `@outdent` query for `}` is completely removed, instead of disabling it just for switch statements). @mvdan let me know if you have any questions on how to approach this.

mvdan:
I indeed had a different fix in mind at first, but then I realized that indenting on switches and selects is the wrong approach to begin with, since those aren't indented in Go at all - only their cases. That is what I tried to explain in the commit message :)

I have tested these changes locally and tried to find edge cases where they break (by using `o` at particular lines in files with switches, selects, etc), and it seems to me like indentation in Go works well now. @Triton171 if you found cases where this change breaks indentation, could you share details so I can reproduce?

I'm slightly confused by the need for the `@outdent` bits; I can add `}` there, or remove them all, and the indentation still works well for me. My best guess is that the `@indent` for nodes like `block` or `communication_case` already unindent when the nodes are closed, so I suspect that an extra `@unindent` is deduplicated.

Triton171:
Sorry, I should have been more specific. Removing the `@indent` for `type_switch_statement` and `expression_switch_statement` should be correct. However, the `@outdent` query for `}` is still needed e.g. for functions:
````go
func f() {
    do_smt()
} // This curly brace is part of the function block but it shouldn't be indented, so it needs an @outdent query
````
Your change causes the last curly brace to be indented one level which is incorrect. This can be fixed by keeping the query but adding more exceptions to it:
````
(
    (_ "}" @outdent) @outer
    (#not-kind-eq? @outer "select_statement")
    (#not-kind-eq? @outer "type_switch_statement")
    (#not-kind-eq? @outer "expression_switch_statement")
)
````

Also, your solution will not correctly indent incomplete case expressions:
````go
func f() {
    switch {
    case true: // Pressing <Enter> here will not increase indentation but it should
    }
}
````
This happens because the `expression_case` node is "too short" (i.e. it doesn't include the following whitespace). It can be fixed by adding `@extend` queries, like the one that was already there:
````
[
    (communication_case)
    (expression_case)
    (default_case)
    (type_case)
] @extend
````
I haven't tested all of this, so let me know if you encounter any issues.

**Comment by mvdan** [comment]
Your suggested changes do seem to do a lot better. I've updated the PR to include them.

I spotted two more bugs, starting at one level of indentation:

1) `switch {<Enter>` (or the same with select) leaves you with an indented new line, when it shouldn't be:

```
    switch {
        // cursor; too much indentation
    }
```

2) `[<Enter>`, which arguably doesn't make sense as Go syntax, leaves you with:

```
    [
        // cursor
] // too much outdentation
```

The second one is arguably not worth fixing, as I only found it by chance, but the first one we should fix. I'm not sure why I'm getting the indentation; switches and selects don't appear in any `@indent` or `@extend`.

Triton171:
Loogs good to me, thanks for working on this! I think the CI failure is unrelated (rebasing on the latest master should fix it).

1. This is due to the auto-pairing code: Inserting a new line between matching parentheses will insert an additional line with an extra level of indentation. Fixing this is not that easy (it requires significant code changes and cannot be fixed by changing the queries), so I think it's fine for now. In the future, we could maybe run 2 indent queries (one for the closing parenthesis and one for the additional new line).
2. I agree that this is not worth fixing unless this is something that frequently happens when editing code.

pascalkuthe (dismissed):
LGTM please rebase on master so the ci is fixed and we can merge

mvdan:
@Triton171 thanks again - that makes sense. I've attributed you in the commit message.

@pascalkuthe rebased.

pascalkuthe (approved):

the-mikedavis (approved):

Issue #6039, opened 2023-02-18
yank into register and system clipboard at the same time


fastfading:
when press  y   
yank text into register and system clipboard at the same time

and there seems to be a error.  
on mac 13 iterm2   
when you select a word by touchpad.   
it shows "yanked main selection to primary clipboard ",  but  actually  not in the system clip board or register

Clipboard provider: pbpaste+pbcopy
System clipboard provider: pbpaste+pbcopy

Abdillah:
This is also true when combining both yank into one shortcut,

```
C = [ "yank", "yank_main_selection_to_clipboard" ]
```

Only `yank` operation takes place although the message subsequently display both operation.

the-mikedavis:
I can't reproduce that. That binding works fine for me locally: the current primary selection ends up on the clipboard and all selections end up on the `"` register (which you can see by pressing `"` in normal mode). Does `yank_main_selection_to_clipboard` work outside of that binding for you?

Abdillah:
Oh it is! 
Sorry, I'm not recheck my recollection (or the issue is not what I mentioned above).
I hope it was pre-23.10 and now has been fixed.

If that is the case, I think there's no strong reason to implement this issue.

Issue #2038, opened 2022-04-08
Configuration: Option to use only system clipboard for all yank & paste actions


thomas-profitt:
(Related: https://github.com/helix-editor/helix/issues/1901 )

I use the following config to use the system clipboard for yank, paste, replace and delete operations:
```
[keys.normal]
p = ":clipboard-paste-after"
P = ":clipboard-paste-before"
y = ":clipboard-yank-join"
Y = ":clipboard-yank"
R = ":clipboard-paste-replace"
d = [":clipboard-yank-join", "delete_selection"]

[keys.select]
p = ":clipboard-paste-after"
P = ":clipboard-paste-before"
y = ":clipboard-yank-join"
Y = ":clipboard-yank"
R = ":clipboard-paste-replace"
d = [":clipboard-yank-join", "delete_selection"]
```

But, shouldn't using the system clipboard rather than internal registers be the default behavior?

Registers are very clean for things like when you want to do some special operation with the text in the register, but that can be implemented using the system clipboard too. It's useful to have separate registers, instead of just one, but it's much more common to want to use one register, especially when we often copy and paste text between the editor and different applications.

So, isn't it sensible for the default primary register to just be the system clipboard?

the-mikedavis:
> it's much more common to want to use one register

I disagree: I almost always use the registers instead of the system clipboard.

It probably shouldn't take so much keybinding to change the behavior if you prefer using the system clipboard all the time though. Maybe a config option that controls the default yank register would be a good middleground?

archseer:
This is deliberate: I wanted to default to the system clipboard but since it's just a single clipboard it gets tricky with multiple selections: you wouldn't be able to yank all of them, just the primary one.

thomas-profitt:
In that case, this issue should be about making it possible to set the default register to the system clipboard in one's own config, without the remapping above.

> It probably shouldn't take so much keybinding to change the behavior if you prefer using the system clipboard all the time though. Maybe a config option that controls the default yank register would be a good middleground?

Just like you said. There should be a "I only ever want to use one register, the system clipboard, for all yank & paste actions" option.


(I used Vim for more than a decade, probably almost every day, and I get when I might want multiple registers, like when recording and playing back macros, but I've never understood why I'd want the yank & paste actions to use anything but the system clipboard by default. I really don't get why some people prefer that UI -- why have another separate default clipboard just for your editor?)

glow12121:
[...]

PR #6820, opened 2023-04-19
Refactor register.rs

CcydtN:
Background
---
I am planning to work on #2038 and #6039. which are both related to registers and clipboard. However, I think the two enhancement will be hard to implement without changing the design of registers, so I work on the register first.

Changes
---
Register become a trait, so different kinds of register can store in a single table, but handle input differently.
`SimpleRegister` is same as the original `Register`
`HistoryRegister` provide the old `Register::push()` function
`BlackHoleRegister` to avoid the "hardcode if" for the underscore(_) register
> For `HistoryRegister`
Before refactoring, `"`,`/`,`:`register was overwrite when yank (e.g`"/y`). After refactoring, it will become append instead of overwrite.

### DIFF
--- a/helix-core/src/register.rs
+++ b/helix-core/src/register.rs
@@ -1,48 +1,95 @@
+use core::fmt::Debug;
 use std::collections::HashMap;
 
-#[derive(Debug)]
-pub struct Register {
-    name: char,
+pub trait Register: Debug {
+    fn read(&self) -> &[String];
+    fn write(&mut self, values: Vec<String>);
+}
+
+#[derive(Debug, Default)]
+pub struct SimpleRegister {
     values: Vec<String>,
 }
 
-impl Register {
-    pub const fn new(name: char) -> Self {
-        Self {
-            name,
-            values: Vec::new(),
-        }
-    }
-
-    pub fn new_with_values(name: char, values: Vec<String>) -> Self {
-        Self { name, values }
-    }
-
-    pub const fn name(&self) -> char {
-        self.name
-    }
-
-    pub fn read(&self) -> &[String] {
+impl Register for SimpleRegister {
+    fn read(&self) -> &[String] {
         &self.values
     }
 
-    pub fn write(&mut self, values: Vec<String>) {
+    fn write(&mut self, values: Vec<String>) {
         self.values = values;
-    }
-
-    pub fn push(&mut self, value: String) {
-        self.values.push(value);
     }
 }
 
-/// Currently just wraps a `HashMap` of `Register`s
+#[derive(Debug, Default)]
+pub struct HistoryRegister {
+    values: Vec<String>,
+}
+
+impl Register for HistoryRegister {
+    fn read(&self) -> &[String] {
+        &self.values
+    }
+
+    fn write(&mut self, values: Vec<String>) {
+        self.values.extend(values.into_iter());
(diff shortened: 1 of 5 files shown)


CcydtN:
Follow up issue: (discovered while development)
- [ ] History register can store infinite record (up to vector limit), should consider giving a fixed limit (exist in before this commit)
- [ ] `/` and `:` register, should show the latest item for history register, currently showing the oldest one (exist before this commit)

the-mikedavis (changes requested):
Normal registers shouldn't have different backends. `/`, `"` and `:` are not special registers but just default registers used by some commands. You should be able to use any other register by using `"<reg>` before that command, for example `"/y` to yank the current selections into the search register. Whether a register is "simple" or "history" is should be decided by how a command uses the register rather than which data structure the register is.

CcydtN:
Could you elaborate more on the case of searching or enter a command?

In those case, should we overwrite the register? or appending?

archseer:
I'm also not sure we want #6039 implemented, it's not an option available in vim

CcydtN:
> I'm also not sure we want #6039 implemented, it's not an option available in vim

In vim, i think it can be done by setting the default register to '+' or '*' which is the clipboard register.

the-mikedavis:
Searching and entering commands should append to whichever register is being used. That's done here in the Prompt: https://github.com/helix-editor/helix/blob/b336cd76a740eea6ea41baa1028233ccf6e12ee5/helix-term/src/ui/prompt.rs#L570-L573

System clipboard yanking is less powerful than register yanking since the clipboard can't support multiple values which is used for yanking/pasting multiple selections in Helix. So I agree with archseer, I don't think we should add #2038 or #6039 - those should be possible to do with keybinding and we don't want extra options where keybindings are sufficient, especially ones that reduce the functionality of a feature.

CcydtN:
System clipboard yanking and paste with multiple selection should be possible. Here is [a draft](https://github.com/CcydtN/helix/blob/clipboard_register/helix-view/src/clipboard_register.rs). If the hash didn't change, it means the content comes from helix, and can be pasted with multiple cursor.

For the two register `:`,`/`, I still think they are weird. Right now, they are having two different function with one data structure. The inner vector can be history records, and also selections from multiple cursor. One of the consequence is that history record gets clear when yank to register. Vim tackle the problem by making them read-only.

An other approach might be storing history in another variable. We can still store the last search in the register, but with no history.

the-mikedavis:
I think that could be a good strategy regardless of #2038 / #6039. I'm still against implementing those two. That change doesn't need this refactor or a new ClipboardRegister type though. It can be done more minimally with changes in `paste_clipboard_impl`, new fields on Editor for the value and hash, and a new function that yanks all selections to the clipboard.

CcydtN:
Alright, I am giving up this idea. I will close this pull request and will make another pull request for the multiple-selection yank function and other minor issue I motioned on top. 

Anyway, thanks for your opinion. @the-mikedavis

Abdillah:
> I don't think we should add #2038 or #6039 - those should be possible to do with keybinding and we don't want extra options where keybindings are sufficient, especially ones that reduce the functionality of a feature.

On the note of #6039, combining `yank` and "yank to system clipboard" actually failed.

PR #6822, opened 2023-04-19
New high contrast theme

nelson-111:
This is my personal high contrast theme. I want it to be part of helix so I don't have to copy paste the file every time I update helix. Why had I made my own theme?
I've made it just because I don't like the other themes.


gabydd:
It's also probably better if you have a personal theme to keep it in ``.config/helix/themes`` or whatever the config path is on your system, then you can update it without submitting a pr to helix or need to copy anything when you update helix.

nelson-111:
Thank you for your comments, I'll close the pull request because I don't want to share the theme.

PR #6834, opened 2023-04-20
fix windows build

pascalkuthe:
#6792 seems to have broken windows builds apparently `/std=c11` and `/std=c++14` can not be used together after all even tough it did sound like that in the documentation. I added the same workaround we already had for unix. I don't have a local windows environment setup and relied on CI for testing but that used a cached build, because the actual grammars didn't change. 

I had to clear windows CI caches for #6808 which meant that CI is now failing on master even tough it passed for all PRs.

### DIFF
--- a/helix-loader/src/grammar.rs
+++ b/helix-loader/src/grammar.rs
@@ -443,10 +443,42 @@ fn build_tree_sitter_library(
             .arg(header_path)
             .arg("/Od")
             .arg("/utf-8")
-            .arg("/std:c++14")
             .arg("/std:c11");
         if let Some(scanner_path) = scanner_path.as_ref() {
-            command.arg(scanner_path);
+            if scanner_path.extension() == Some("c".as_ref()) {
+                command.arg(scanner_path);
+            } else {
+                let mut cpp_command = Command::new(compiler.path());
+                cpp_command.current_dir(src_path);
+                for (key, value) in compiler.env() {
+                    cpp_command.env(key, value);
+                }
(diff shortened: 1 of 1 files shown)


helix-loader/src/grammar.rs
```diff
@@ -443,10 +443,42 @@ fn build_tree_sitter_library(
             .arg(header_path)
             .arg("/Od")
             .arg("/utf-8")
-            .arg("/std:c++14")
             .arg("/std:c11");
```
dead10ck, line 446: Won't we need to move this into the block for c files too?

gabydd:
Apparently sr.ht has an ongoing DDoS and that is why it is down: https://codeberg.org/ddevault/pages/src/commit/678f15e04e01ee4420b6140ad333cb3f3d22d7bc/affected/git.sr.ht/index.html#L461 (also the reason the build is failing for anyone checking this pr)

helix-loader/src/grammar.rs
pascalkuthe: yup that's right I fogot to commit something. I already suspected that the windows test CI did not in fact build the grammars but wanted to investigate further once the source hut DOS if over

pascalkuthe: still gonna clear the caches later just to be sure tough

dead10ck: Oh I see, the C++ command is a totally separate command; at first glance, I thought it was adding these options to the same command. This makes sense now.

PR #6835, opened 2023-04-21
make `:u` alias `:update`

Merlin1846:
The `update` command didn't have an alias despite `write` having an alias, so I gave it the alias `u`. I felt like `u` matched the same theme that `write` and `open` had so that's why I choose `u`.


the-mikedavis (approved):
This seems ok to me 👍

It mirrors `:w` and we don't have any other typable commands that start with "u" yet.

pascalkuthe (approved):

PR #6846, opened 2023-04-22
display formatting error

### LINKED ISSUE #6736: There should be an error when formating fails
### Summary

If I format a document that does not have a LSP installed for its language nothing happens

### Reproduction Steps

**I tried this:**
1. `:format`

**I expected this to happen:**
The editor tells me that their is no language server installed for this language and possibly shows me a guide to install a language server

**Instead, this happened:**
Nothing happend

### Helix log

```
2023-04-13T10:38:25.338 helix_view::editor [ERROR] Failed to initialize the LSP for `source.python` { cannot find binary path }
```

</details>

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

helix 23.03

shinhs0506:
fixes #6736


helix-term/src/commands.rs
```diff
@@ -2730,14 +2730,20 @@ async fn make_format_callback(
...
-                log::info!("discarded formatting changes because the document changed");
+        match format {
+            Ok(format) => {
+                if doc.version() == doc_version {
+                    doc.apply(&format, view.id);
+                    doc.append_changes_to_history(view);
+                    doc.detect_indent_and_line_ending();
+                    view.ensure_cursor_in_view(doc, scrolloff);
+                } else {
+                    log::info!("discarded formatting changes because the document changed");
```
pascalkuthe, line 2741: We likely also want an error message in this else branch

helix-term/src/commands.rs
```diff
@@ -2730,14 +2730,20 @@ async fn make_format_callback(
...
+                }
+            },
+            Err(error) => {
+                editor.set_error(format!("Error while formatting: {}", error));
```
pascalkuthe, line 2745: I think we can do a bit better with the error message here:

```suggestion
                editor.set_error(format!("Failed to format document: {}", error));
```

pascalkuthe (commented):
This is a good usability improvement in general 👍

helix-view/src/document.rs
```diff
@@ -605,32 +605,38 @@ impl Document {
...
-            Ok(helix_lsp::util::generate_transaction_from_edits(
-                &text,
-                edits,
-                offset_encoding,
-            ))
-        };
-        Some(fut.boxed())
+            None => {
+                Some(async move { return Err(FormatterError::LanguageServerNotInitialized) }.boxed())
+            }
```
the-mikedavis, line 638: `Document::language_server` returning None can happen if you don't have the language server installed for a language. We shouldn't show an error if we can't format because the language server isn't installed - it would be too noisy

pascalkuthe:
Ah I forgot this conflicts with #2507. We might want to hold off until that is merged since that PR probably changes how the failure condition is triggered anyway

David-Else:
https://github.com/helix-editor/helix/pull/2507 got merged, can we merge this now?

helix-view/src/document.rs
shinhs0506: The code needs more work, but that is exaclty what the issue wants resolved. Do we still not want to show errors in those cases (when language server isn't installed)?

gabydd: I think probably we should show an error in the format type-able command if the lsp is not installed instead

the-mikedavis: Showing errors in `:format` is ok but `make_format_callback` is called by `:write` and `:write-all` as well

gabydd: Oh yeah sorry that's what I meant should have been more clear

shinhs0506: sounds good, thanks

the-mikedavis:
Superseded by #12183

PR #6856, opened 2023-04-23
return early if soft wrap is disabled

ymgyt:
Change to use the result of move_vertically if soft wrap is disabled

In reading the helix code, I wondered if not using the result of move_vertically was not the intent.
If you do not need this change, please feel free to CLOSE it!


pascalkuthe (approved):
Good catch,thanks!

the-mikedavis (approved):

PR #6858, opened 2023-04-23
Use crossterm cursor when terminal loses focus

hnorkowski:
All but the documents block cursor change their shape to a box outline if the terminal loses focus. I am missing that feature because (depending on your window manager) it can be hard to tell which editor has focus if you have multiple open at the same time. Because the custom rendered block cursor is just the background color of the respective char it is no possible to achieve the same behavior (display as outline).

Therefore I created a proof of concept for the custom block cursor to change its color when the terminal window loses focus. 

If the theme does not define a color for the unfocused cursor (e.g. `ui.cursor.primary.unfocused`) it should calculate the color by faking transparency of the cursor. Therefore it tries to get the background color and the cursor color and mix them.

Currently I have 2 main problems with the color calculation: 

- I do not know how to get the RGB values of a named color (e.g. yellow) but I need those to calculate the "transparent" cursor color
- Some themes (e.g. default) do not define a cursor color (Style.bg is None). There are default colors defined but I do not know how I can query them.
- I also tried to use the DIM modifier but it did not seem to have any effect (atleast in alacritty). I forgot to create a commit for that but if desired I could reimpl. that to analyze why it did not work.

The current implementation works well with some themes, e.g. gruvbox, catpuccin_*, emacs.

Another questions that should be discussed:

- If the feature gets accepted, should it be the default? 
  - I think yes because all other cursors have this behavior

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -158,6 +158,7 @@ impl EditorView {
                     view,
                     theme,
                     &config.cursor_shape,
+                    self.terminal_focused,
                 ),
             );
             let focused_view_elements = Self::highlight_focused_view_elements(view, doc, theme);
@@ -420,6 +421,7 @@ impl EditorView {
         view: &View,
         theme: &Theme,
         cursor_shape_config: &CursorShapeConfig,
+        is_terminal_focused: bool,
     ) -> Vec<(usize, std::ops::Range<usize>)> {
         let text = doc.text().slice(..);
         let selection = doc.selection(view.id);
@@ -490,13 +492,17 @@ impl EditorView {
                         cursor_start
                     };
                 spans.push((selection_scope, range.anchor..selection_end));
-                if !selection_is_primary || cursor_is_block {
+                // add block cursors
+                // skip primary cursor if terminal is unfocused - crossterm cursor is used in that case
+                if !selection_is_primary || (cursor_is_block && is_terminal_focused) {
                     spans.push((cursor_scope, cursor_start..range.head));
                 }
             } else {
                 // Reverse case.
                 let cursor_end = next_grapheme_boundary(text, range.head);
-                if !selection_is_primary || cursor_is_block {
+                // add block cursors
+                // skip primary cursor if terminal is unfocused - crossterm cursor is used in that case
+                if !selection_is_primary || (cursor_is_block && is_terminal_focused) {
                     spans.push((cursor_scope, range.head..cursor_end));
                 }
                 // non block cursors look like they exclude the cursor
@@ -1520,8 +1526,15 @@ impl Component for EditorView {
 
     fn cursor(&self, _area: Rect, editor: &Editor) -> (Option<Position>, CursorKind) {
         match editor.cursor() {
-            // All block cursors are drawn manually
-            (pos, CursorKind::Block) => (pos, CursorKind::Hidden),
+            // all block cursors are drawn manually
+            (pos, CursorKind::Block) => {
+                if self.terminal_focused {
+                    (pos, CursorKind::Hidden)
+                } else {
+                    // use crossterm cursor when terminal loses focus
+                    (pos, CursorKind::Underline)
+                }
+            }
             cursor => cursor,
         }
     }


helix-term/src/ui/editor.rs
```diff
@@ -409,29 +414,54 @@ impl EditorView {
         let base_cursor_scope = theme
             .find_scope_index_exact("ui.cursor")
             .unwrap_or(selection_scope);
+        let base_cursor_unfocused_scope = theme
+            .find_scope_index_exact("ui.cursor_unfocused")
```
the-mikedavis, line 418: These theme keys with `_unfocused` should be `*.unfocused` instead. `ui.cursor.primary_unfocused.{insert,select,normal}` should be `ui.cursor.primary.{insert,select,normal}.unfocused`.

The new theme keys should be added to the book in `book/src/themes.md`

helix-view/src/theme.rs
```diff
@@ -288,9 +288,76 @@ fn build_theme_values(
...
+                scopes.push(unfocused_cursor_key.to_owned());
+                highlights.push(cursor_unfocused_scope);
+            }
+        }
+    }
+
     (styles, scopes, highlights)
 }
 
+fn calculate_transparent_color(
```
the-mikedavis, line 332: Automatically figuring out a default dimmed color seems unnecessary to me. If a theme wants to take advantage of focused/unfocused it should define the `*.unfocused` keys

archseer:
Hmm this is probably a better idea than what I had: I simply hid the cursor. But that makes it a bit confusing when the window is out of focus since no other editor would behave that way

helix-term/src/ui/editor.rs
hnorkowski: Changed the names and added the keys into the book. I also removed non primary keys as I only update the primary cursor.

helix-view/src/theme.rs
hnorkowski: Agreed. Removed it.

pascalkuthe (commented):
This looks good to me. It would be nice if you could define an unfocused color for the default theme tough (the rest of the themes can be changed by their author but it's nice if the default works)

book/src/themes.md
```diff
@@ -265,65 +265,69 @@ These scopes are used for theming the editor interface:
...
-| `diagnostic.info`                 | Diagnostics info (editing area)                                                                |
-| `diagnostic.warning`              | Diagnostics warning (editing area)                                                             |
-| `diagnostic.error`                | Diagnostics error (editing area)                                                               |
+| Key                                         | Notes                                                                                          |
+| ---                                         | ---                                                                                            |
+| `ui.background`                             |                                                                                                |
+| `ui.background.separator`                   | Picker separator below input line                                                              |
+| `ui.cursor`                                 |                                                                                                |
```
the-mikedavis, line 272: Should there be a `ui.cursor.unfocused` as well for themes that use the same cursor styling for all cursors?

hnorkowski: I think the question comes down to: 
Should all cursors support this behavior or only the primary?

I am not sure about the answer but I could impl it that way.

hnorkowski:
> This looks good to me. It would be nice if you could define an unfocused color for the default theme tough (the rest of the themes can be changed by their author but it's nice if the default works)

Yeah I am trying to but I cannot figure out the default theme keys. The cursor has different colors depending on if its on a char or a whitespace and cannot figure out where this behavior is defined the theme.toml. With my current impl. the cursor will always have the same color when out of focus.

book/src/themes.md
pascalkuthe: I think the behaviour should be applied to all cursors (if the theme defines the appropriate theme keys).

pascalkuthe:
> > This looks good to me. It would be nice if you could define an unfocused color for the default theme tough (the rest of the themes can be changed by their author but it's nice if the default works)
> 
> Yeah I am trying to but I cannot figure out the default theme keys. The cursor has different colors depending on if its on a char or a whitespace and cannot figure out where this behavior is defined the theme.toml. With my current impl. the cursor will always have the same color when out of focus.

That's default theme uses `"ui.cursor" = { modifiers = ["reversed"] }` which just swaps foreground and background color. Even if you keep `reversed` changing the bg color would involve overwiting the `fg` color so defining different colors for whithespaces is not possible. I think that's ok this distinction isn't that important if the editor isn't focused.

hnorkowski:
But this creates the problem that I need to chose a color that works well for all colors the cursor can have. This is kinda tricky. I will try to find a good one.

book/src/themes.md
hnorkowski: Ok I impl that and for selections as well.

pascalkuthe:
I marked this as ready for review as we are already reviewing. If you feel this is not ready yet feel free to change the status again. There seem to be some problems with the default theme and lints that need fixing (see CI)

hnorkowski:
I think this is ready. I updated the catppuccin mocha theme to support this feature aswell.

The only thing that could need work is the default theme. You guys might wanna check it out and maybe even try if you like other colors more.

helix-term/src/ui/editor.rs
```diff
@@ -43,6 +43,8 @@ pub struct EditorView {
     pub(crate) last_insert: (commands::MappableCommand, Vec<InsertEvent>),
     pub(crate) completion: Option<Completion>,
     spinners: ProgressSpinners,
+    // Tracks if the terminal window is focused by reaction to terminal focus events
```
the-mikedavis, line 46:
```suggestion
    /// Tracks if the terminal window is focused by reaction to terminal focus events
```

so it becomes a doc comment. We don't generate documentation for the helix crates but docs are still nice since they'll show up in LSP hover

the-mikedavis (commented):
This is looking good, just a minor comment. We'll also want to get @archseer's review since this touches the default theme

the-mikedavis (dismissed):

pascalkuthe (dismissed):

hnorkowski:
Would it be possible to just change the cursor type when out of focus? The insert cursor behaves like expected and changes to an empty box when helix loses focus. This would replicated the expected behavior of the box cursor.

hnorkowski:
> this is not possible because helix does not use the terminals builtin block cursor (since that doesn't work with multi cursors) tui applications can not draw a hollow cursor

Yes I know that but when using the `bar` cursor in insert mode it does use the builtin cursor (at least if helix loses focus while in insert mode with the bar cursor it renders it as an empty box). So my idea is: Why not just change the cursor type to `bar` while helix has no focus and back to block when its in focus again?

I guess the problem could be that if you use multi cursors at that point we would need to change only the cursor type of the main cursor?

archseer:
> So my idea is: Why not just change the cursor type to `bar` while helix has no focus and back to block when its in focus again?

This just might be the best solution because it'll properly disappear when changing tmux splits, but not when focus is lost

the-mikedavis (approved):

hnorkowski:
This is not ready yet. I think we should go back to draft. I want to try this idea:


> So my idea is: Why not just change the cursor type to `bar` while helix has no focus and back to block when its in focus again?

However I am a bit lost in the code base. I was yet not able to find the correct place to overwrite the cursor. Do you have a hint for me?

the-mikedavis:
https://github.com/helix-editor/helix/blob/8c68074fa6d2e3771bbc1a7f80bc9b422a787da4/helix-term/src/application.rs#L1152-L1156 is a good example of explicitly setting a block cursor

hnorkowski:
I finally managed to impl. the behavior. It seems to work flawless.

The only point left to discuss: Should be also do something about secondary cursors? Currently only the main cursor is changing when losing focus. I think for me thats ok. What is your opinion?

the-mikedavis:
I think we can leave secondary cursors as-is. We can't make them bars (we just fill in a background to make them look like blocks normally) like the primary cursor. We could add theming for them when unfocused like the original change in this PR if the bar cursor isn't clear enough. That could probably be a follow-up though

hnorkowski:
Yeah I think you are right that it could be a follow-up after using this for some time and see if it would be helpful.

With that I think the PR is ready.

helix-term/src/ui/editor.rs
```diff
@@ -1521,7 +1527,14 @@ impl Component for EditorView {
     fn cursor(&self, _area: Rect, editor: &Editor) -> (Option<Position>, CursorKind) {
         match editor.cursor() {
             // All block cursors are drawn manually
-            (pos, CursorKind::Block) => (pos, CursorKind::Hidden),
+            (pos, CursorKind::Block) => {
+                if self.terminal_focused {
+                    (pos, CursorKind::Hidden)
+                } else {
+                    // Use crossterm cursor when terminal loses focus for consistent unfocues cursor behaviour
```
the-mikedavis, line 1534:
```suggestion
                    // Use crossterm cursor when terminal loses focus
```

There's a little typo there in "unfocues" but I think we can drop the consistency part anyways

the-mikedavis (commented):
Just a minor nit about a comment.

I like this behavior - it looks nice to me on Kitty for single panes and splits

helix-term/src/ui/editor.rs
hnorkowski: Done

the-mikedavis (approved):

pascalkuthe (approved):

Issue #1752, opened 2022-03-05
Setting labels for custom menus / commands?


sudormrfbin:
### Discussed in https://github.com/helix-editor/helix/discussions/1744

<div type='discussions-op-text'>

<sup>Originally posted by **@MattCheely** March  4, 2022</sup>
I recently realized I can add my own menu entries to do things like create "sub-menus", but the labeling gets a bit awkward. As an example if I do something in my config file like:

```
[keys.normal.space]
"f" = { "f" = "file_picker", "s" = ":write", "c" = ":open ~/.config/helix/config.toml" }
```
I can use `space f` to get into file mode, but the `f` entry after pressing `space` has no label, and the `c` entry displays the open command, but it would be nice to make it say something like "Edit Helix config".  Are either of those possible? </div>

alunyaka:
Still no workaround on this?

brokenthorn:
https://github.com/TheAwiteb/helix/pull/1 has a solution...

janos-r:
After
```
space.B = ":echo %sh{git blame -L %{cursor_line},+1 %{buffer_name}}"
```
was added as a recommendation to the docs, this feels even more pressing. There has to be a way to improve this

<img width="665" height="112" alt="Image" src="https://github.com/user-attachments/assets/5e8e5e1a-1bc8-42f2-bf7e-49974cdeb854" />

gsrec:
I support request, it would be great to add your own names for custom menus, which can become a lot in a large project. This causes confusion:
<img width="1514" height="279" alt="Image" src="https://github.com/user-attachments/assets/23d2fd65-fd80-40dd-8d51-21ddbf028d21" />
Maybe it will be possible to extract comments from toml with special tags?
```toml
[keys.normal.C-g]
# <go test current buffer/method>
t = ':sh inv go-test %{buffer_name} -e "%{selection}"'
```

MarSik:
Folks, instead of inveting a new and creative way to mangle TOML format, why not simply agree on a well known TOML key suffix? Something like:

```
[keys.normal.C-g]
t = ':sh inv go-test %{buffer_name} -e "%{selection}"'
t.title = "go test current buffer/method"
```

gsrec:
[...]

PR #6862, opened 2023-04-24
add support for naming multi commands in config

robinvd:
Any keybinding with multiple commands currently show up as `Multiple commands`, this pr adds support for naming these. 

Because strings/lists/maps are already used in the keys config file, we have to use some other method to convey the name. for Example
```
[keys.normal.g]
l = {"name": "goto line end and select", "commands": ["select_mode", "goto_line_end", "exit_select_mode"]}
```

would try to map `name`/`commands` as keys (we could also special case this format i guess). I just quickly added the `?` to get it working but any other ideas welcome (maybe `:named some name`, i quickly searched the issue tracker but couldnt find a related issue.

before (see h,l):
image

after:
```
[keys.normal.g.l]
label = "select till end of line"
command = ["select_mode", "goto_line_end", "exit_select_mode"]

[keys.normal.g.h]
label = "select till start of line"
command = ["select_mode", "goto_line_start", "exit_select_mode"]

```
image

again ideas for other syntaxes welcome

edit: just found https://github.com/helix-editor/helix/issues/1752


robinvd:
> I would prefer to see this build on #3958 and use "label" and "commands" rather than the custom syntax with `?`

ah @the-mikedavis there was already a PR ill close this hope, hopefully that one gets merged :)

robinvd:
@the-mikedavis I see that mr was still lacking named multi commands support, so i've updated this to build upon that mr and pulled in the the latest master

pascalkuthe:
closing this one as stale. I think this is a feature I woudl want to see but it probably needs a fresh start and may need to wait until after we switched to scheme. Thank you for contributing!

PR #6874, opened 2023-04-25
Update Typescript, TSX and Svelte grammar, to latest tag

jeppech:
Support for the `satisfies`-keyword in Typescript
Support for the `@const`-expression in Svelte templates


jeppech:
I just saw that the `tsx`-grammar, uses the typescript tree sitter repo too. So I've updated that as well.

archseer:
Did you also check for any query differences or new nodes we can now highlight? It's usually not enough to just bump the grammar

jeppech:
@archseer Oh sorry, I did not know, I'll take a closer look at the changes, to the different repos.

jeppech:
I've added the `satisfies`-keyword in highlights.scm for typescript, and an alternative css injection, for svelte templates, that the svelte tree-sitter repo mentioned.

I'm unsure, where I can find any new node that the repos may have added.

the-mikedavis (approved):

erasin:
@jeppech need append file [indents.scm](https://github.com/Himujjal/tree-sitter-svelte/blob/master/queries/indents.scm) for svelte.

```
[
  (element)
  (if_statement)
  (each_statement)
  (await_statement)
  (script_element)
  (style_element)
] @indent

[
  (end_tag)
  (else_statement)
  (if_end_expr)
  (each_end_expr)
  (await_end_expr)
  ">"
  "/>"
] @outdent
```

PR #6881, opened 2023-04-26
Don't crash when getcwd fails

sam-mccall:
This can happen e.g. when you lose access to the filesystem

get_relative_path() already returns absolute paths when the input isn't under the working directory, so we just return absolute paths in this case too.


pascalkuthe:
In general this is a good change but we unwrap/except `env::curennt_dir()` in a bunch of other places (I found at least 4 case with a  quick grep) so those should probably receive a similar treatment.

the-mikedavis:
Superseded by https://github.com/helix-editor/helix/pull/7185

PR #6882, opened 2023-04-26
reduce stat calls during git discovery

pascalkuthe:
This should help with https://github.com/helix-editor/helix/issues/6867, I talked with @Byron and for our purposes checking `.git` should be enough and take less stat calls.

I still want to cache the repo lookup here eventually, but that needs some more considerations about resource cleanup (so we don't keep an infinite amout of repos open, probably some kind of LRU mechanism is needed). I also want to look into opening the repo asynchronously, but that's a bit awkward to synchronize, so I will look at that in a separate PR too.

### DIFF
--- a/helix-vcs/src/git.rs
+++ b/helix-vcs/src/git.rs
@@ -1,4 +1,4 @@
-use anyhow::{bail, Context, Result};
+use anyhow::{anyhow, bail, Context, Result};
 use arc_swap::ArcSwap;
 use std::path::Path;
 use std::sync::Arc;
@@ -15,7 +15,13 @@ mod test;
 pub struct Git;
 
 impl Git {
-    fn open_repo(path: &Path, ceiling_dir: Option<&Path>) -> Result<ThreadSafeRepository> {
+    fn open_repo(path: &Path) -> Result<ThreadSafeRepository> {
+        let git_dir = path
+            .ancestors()
+            .find(|ancestor| ancestor.join(".git").exists())
+            .ok_or_else(|| anyhow!("No repostiory found at {path:?}"))?
+            .join(".git");
+
         // custom open options
         let mut git_open_opts_map = gix::sec::trust::Mapping::<gix::open::Options>::default();
 
@@ -31,6 +37,7 @@ impl Git {
             includes: true,
             git_binary: cfg!(windows),
         };
+
         // change options for config permissions without touching anything else
         git_open_opts_map.reduced = git_open_opts_map.reduced.permissions(gix::Permissions {
             config,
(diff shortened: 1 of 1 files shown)


helix-vcs/src/git.rs
```diff
@@ -15,7 +15,13 @@ mod test;
 pub struct Git;
 
 impl Git {
-    fn open_repo(path: &Path, ceiling_dir: Option<&Path>) -> Result<ThreadSafeRepository> {
+    fn open_repo(path: &Path) -> Result<ThreadSafeRepository> {
+        let git_dir = path
+            .ancestors()
+            .find(|ancestor| ancestor.join(".git").exists())
+            .ok_or_else(|| anyhow!("No repostiory found at {path:?}"))?
```
Byron, line 22:
```suggestion
            .ok_or_else(|| anyhow!("No repository found at {path:?}"))?
```

Byron (commented):
I think it's worth noting that this way the git ceiling dir isn't respected, which might have been a silent benefit of the current discovery mechanism.

If it helps, in theory https://github.com/Byron/gitoxide/pull/832 should end up doing pretty much the same work that git does as alternative.

If ceilings aren't important, I think this is about as fast as it gets.

pascalkuthe:
> I think it's worth noting that this way the git ceiling dir isn't respected, which might have been a silent benefit of the current discovery mechanism.
> 
> If it helps, in theory [Byron/gitoxide#832](https://github.com/Byron/gitoxide/pull/832) should end up doing pretty much the same work that git does as alternative.
> 
> If ceilings aren't important, I think this is about as fast as it gets.

Right now we set the ceiling dir always to `None`. For helix using cwd as a ceiling directory doesn't work so I couldn't really comeup with a portable ceiling dir previously. We could think about setting it to `$HOME` (or windows equivalent) I suppose

Byron:
True! Maybe it's the best choice anyway as it puts control over how to handle ceilings and crossing of filesystems into `helix` hands.

On the other hand, maybe a future step could be to adjust `gix_discover::is_git|discover()` to only expect worktrees for optimal FS access. Obviously its relevant to editors, maybe it's an API that should be available in `gix-discover`.

helix-vcs/src/git.rs
```diff
@@ -41,16 +48,8 @@ impl Git {
             ..gix::Permissions::default_for_level(gix::sec::Trust::Full)
         });
 
-        let mut open_options = gix::discover::upwards::Options::default();
-        if let Some(ceiling_dir) = ceiling_dir {
-            open_options.ceiling_dirs = vec![ceiling_dir.to_owned()];
```
Byron, line 46: I actually recommend to not go ahead with this PR and instead use the new `dot_git_only` option. With it set to `true` the discovery can be optimal for this usecase, while making it easy to setup ceiling dirs and handle cross-filesystem checks.

pascalkuthe: thanks for  taking this further, I will update this PR to use the new gitoxide version once that is released

pascalkuthe:
Closing in favor of https://github.com/helix-editor/helix/pull/6890

PR #6884, opened 2023-04-26
add go.work and remove Gopkg.toml as Go roots

mvdan:
(see commit message)

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -292,7 +292,7 @@ name = "go"
 scope = "source.go"
 injection-regex = "go"
 file-types = ["go"]
-roots = ["Gopkg.toml", "go.mod"]
+roots = ["go.work", "go.mod"]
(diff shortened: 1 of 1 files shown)


pascalkuthe (approved):
This seems reasonable but why remove `Gopkg.toml`? It doesn't seem to cause any adverse affects and would keep backwards comparability with older projects

mvdan:
I couldn't find another major editor that still looks for the file. And dep hasn't been maintained for over four years now, so it's unlikely that it has worked for some time. Most importantly: this is to determine the workspace root for gopls, and gopls only supports Go modules, so I see even less reason for any form of optimistic backwards compatibility :)

pascalkuthe:
Ah if gopls doesn't support this then it definuly makes sense to remove this 👍

Issue #5344, opened 2022-12-30
Exclude the definition from references in LSP searches and also jump straight to reference if only one exists


effinsky:
The first ref search result is actually definition that I started the search from: 
<img width="1255" alt="Screenshot 2022-12-30 at 11 07 17" src="https://user-images.githubusercontent.com/54535543/210058750-b8ec2100-1e2e-44ef-b3ce-5a51ce1a87ab.png">

As far as I'm concerned, it should not be listed at all. Relatedly, It would be cool if a jump happened straight to reference if only one exists, with the search window never appearing.

the-mikedavis:
It's possible to configure this in the LSP references request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#referenceContext. Currently we hard-code this to `true` but it could be configurable.

> It would be cool if a jump happened straight to reference if only one exists, with the search window never appearing.

That is how goto-references should behave if the language server only sends one location: https://github.com/helix-editor/helix/blob/63dcaae1b9083396fb3faaef9eaa2421f7e48fb9/helix-term/src/commands/lsp.rs#L878-L880

pascalkuthe:
> It's possible to configure this in the LSP references request: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#referenceContext. Currently we hard-code this to true but it could be configurable.

I wanted to send a PR for this but didn't get around to this yet. Was it a concious choice to hardcode true here? It seems like `false` would be the better default here since you can always use `gd` instead. I didn't find prior discussion the last time I looked into this

the-mikedavis:
The descision/code predates me (#8) but the `true` default seems sensible to me. IMO the definition is a reference - when I want to see and especially edit all instances of some symbol, I expect to see the definition too. I'm curious what other editors use for a default here.

effinsky:
I'm cool with this being the default if most folks find that useful and/or consistent with their expectations. I'd just like to be able to exclude it in a cfg opt and jump straight to a sole reference other than the def as well.

Omnikar:
I'd say a good default behaviour is listing the definition reference at the bottom of the list, which seems to already be how it behaves for Rust.

PR #6886, opened 2023-04-26
feat: add a config option to exclude declaration from LSP references

mcdoker18:
Fixes: #5344

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1078,9 +1078,9 @@ pub fn goto_implementation(cx: &mut Context) {
 }
 
 pub fn goto_reference(cx: &mut Context) {
+    let config = cx.editor.config();
     let (view, doc) = current!(cx.editor);
     let language_server = language_server!(cx.editor, doc);
-    let config = cx.editor.config.load();
     let offset_encoding = language_server.offset_encoding();
 
     let pos = doc.position(view.id, offset_encoding);
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1080,7 +1080,7 @@ pub fn goto_implementation(cx: &mut Context) {
 pub fn goto_reference(cx: &mut Context) {
     let (view, doc) = current!(cx.editor);
     let language_server = language_server!(cx.editor, doc);
-    let editor_cfg = cx.editor.config.load();
+    let config = cx.editor.config.load();
     let offset_encoding = language_server.offset_encoding();
 
     let pos = doc.position(view.id, offset_encoding);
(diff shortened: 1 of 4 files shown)


pascalkuthe (dismissed):
LGTM, I erpsonally prefer turning this off so it's nice to have a config option for this :+1:

helix-term/src/commands/lsp.rs
```diff
@@ -1080,11 +1080,17 @@ pub fn goto_implementation(cx: &mut Context) {
 pub fn goto_reference(cx: &mut Context) {
     let (view, doc) = current!(cx.editor);
     let language_server = language_server!(cx.editor, doc);
+    let editor_cfg = cx.editor.config.load();
```
the-mikedavis, line 1083:
```suggestion
    let config = cx.editor.config();
```

mcdoker18: Fixed

pascalkuthe (dismissed):

mcdoker18:
Personally, I'd like to have this option disabled by default. Because we have the `go to declaration` action.  I have never wanted to find all usage and declarations of a function at the same time.

pascalkuthe:
> Personally, I'd like to have this option disabled by default. Because we have the `go to declaration` action. I have never wanted to find all usage and declarations of a function at the same time.

There was some discussion about this in #5344 I personally also prefer the option to be disabled but I think @the-mikedavis and @archseer prefer the current default. I think it might be a difference of people who come from vscode/nvim vs intellij. At the end of the day this is personal preference and I think both are fine defaults. It's pretty easy to change so I don't think it's that important what the default is

the-mikedavis (approved):

helix-term/src/commands/lsp.rs
```diff
@@ -1080,11 +1080,17 @@ pub fn goto_implementation(cx: &mut Context) {
 pub fn goto_reference(cx: &mut Context) {
     let (view, doc) = current!(cx.editor);
     let language_server = language_server!(cx.editor, doc);
+    let config = cx.editor.config.load();
```
the-mikedavis, line 1083:
```suggestion
    let config = cx.editor.config();
```

There will be a borrowing conflict for `cx.editor`. Moving this above the `let (view, doc)` line will resolve that

the-mikedavis (changes requested):
Ah actually one more minor thing

helix-term/src/commands/lsp.rs
mcdoker18: Sorry, I am just learning Rust. Can you please elaborate about the borrowing conflict? It works fine right now.

If I remove the call of load method and put it above then it will not be compiled.

the-mikedavis (approved):

effinsky:
thanks, all!

PR #6889, opened 2023-04-26
Allow clipboard paste with multiple selection

CcydtN:
Allow pasting content with multiple selection if two requirement is fulfill:
1. The content is yank from same instance.
2. No other application overwrite the clipboard in that period. (unless the exact same string is copied or hash collision occurs)

Tested and working:
- wl-paste+wl-copy
- clipboard-win
- termcode

Not tested:
- pbpaste+pbcopy (mac)
- xclip
- xsel
- win32yank.exe
- termux-clipboard-set
- tmux

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -61,11 +61,13 @@ use crate::{
 
 use crate::job::{self, Jobs};
 use futures_util::StreamExt;
-use std::{collections::HashMap, fmt, future::Future};
-use std::{collections::HashSet, num::NonZeroUsize};
-
 use std::{
     borrow::Cow,
+    collections::{hash_map::DefaultHasher, HashMap, HashSet},
+    fmt,
+    future::Future,
+    hash::{Hash, Hasher},
+    num::NonZeroUsize,
     path::{Path, PathBuf},
 };
 
@@ -3597,6 +3599,12 @@ fn yank(cx: &mut Context) {
     exit_select_mode(cx);
 }
 
+fn compute_hash<T: Hash>(item: &T) -> u64 {
+    let mut hasher = DefaultHasher::default();
+    item.hash(&mut hasher);
+    hasher.finish()
+}
+
 fn yank_joined_to_clipboard_impl(
     editor: &mut Editor,
     separator: &str,
@@ -3623,6 +3631,8 @@ fn yank_joined_to_clipboard_impl(
     );
 
     let joined = values.join(separator);
+    editor.last_clipboard[clipboard_type as usize].contents = values;
+    editor.last_clipboard[clipboard_type as usize].hash = Some(compute_hash(&joined));
 
     editor
         .clipboard_provider
@@ -3786,7 +3796,15 @@ fn paste_clipboard_impl(
     let (view, doc) = current!(editor);
     match editor.clipboard_provider.get_contents(clipboard_type) {
         Ok(contents) => {
-            paste_impl(&[contents], doc, view, action, count, editor.mode);
+            let hash = compute_hash(&contents);
+            let contents_slice = [contents];
+            let is_hash_equal = Some(hash) == editor.last_clipboard[clipboard_type as usize].hash;
+            let paste_contents: &[String] = if is_hash_equal {
+                &editor.last_clipboard[clipboard_type as usize].contents[..]
+            } else {
+                &contents_slice
+            };
+            paste_impl(paste_contents, doc, view, action, count, editor.mode);
             Ok(())
         }
         Err(e) => Err(e.context("Couldn't get system clipboard contents")),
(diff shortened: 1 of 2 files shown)


pascalkuthe:
The default hasher is not a cryptographically secure hash (like sha1) and collisions are quite likely. We don't need a hash here tough a simple text comparison should be just as fast.

pascalkuthe (changes requested):
The default hasher is not a cryptographically secure hash (like sha1) and collisions are quite likely. Usually that's account for by doing quadratic probing of all entries in the same hashmap bucket (basically comparing the contents) but here we don't need a hash here a simple text comparison should be just as fast.

helix-view/src/editor.rs
```diff
@@ -823,7 +823,7 @@ pub struct Editor {
     pub breakpoints: HashMap<PathBuf, Vec<Breakpoint>>,
 
     pub clipboard_provider: Box<dyn ClipboardProvider>,
-
+    pub last_clipboard: [Vec<String>; 2],
```
the-mikedavis, line 826: Very minor nit but I think this field should have docs on it: the name doesn't describe the values well (and I think it's hard to come up with a good name for this field without it being 4 or more words long) so I think it will be difficult to tell what this is used for without checking the reference in `commands.rs`.

```suggestion
    /// The last selection contents yanked to the system and primary clipboards.
    /// These values are saved so that pasting from the system clipboard may
    /// retain multiple selections.
    pub last_clipboard: [Vec<String>; 2],
```

the-mikedavis (commented):
One minor point but otherwise I think this looks good

the-mikedavis:
Ah I think I see where you were going now with #6820: was the idea to refactor the system and primary clipboards into special registers? (For context, `_` is the only special register currently implemented but there are [more](https://github.com/mawww/kakoune/blob/master/doc/pages/registers.asciidoc#special-registers) we'd like to add.)

I think it might be ok to implement #2038 if it were a config option

```rust
// helix-view/src/editor.rs
// ...
pub struct Config {
  // ...
  #[serde(default)] // <- default this to `"`
  pub default_yank_register: char,
}
// ...
```

and then commands that yank could use that value if no register were provided through `select_register` (`"`).

I played around with refactoring Register/Registers to implement the remaining special registers in a branch and it's a rather large refactor but I believe it's possible to implement the system & primary clipboards as special registers. The changes overlap with https://github.com/helix-editor/helix/pull/5577 (but need to pass the editor rather than document and some other minor changes).

Do you have an opinion on the clipboards being special registers @archseer? I think they are not registers in Kakoune only because Kakoune interacts with clipboard tools only via shell piping.

helix-view/src/editor.rs
CcydtN: Sure. I just added the commit.

CcydtN:
> Ah I think I see where you were going now with #6820: was the idea to refactor the system and primary clipboards into special registers?

yep, that's my final goal for that Pull request.
But I soon realized it is a huge refactoring and affecting other functionality, like search history. It require much more planning than I expected, so I stoped.

helix-term/src/commands.rs
```diff
@@ -3786,7 +3787,16 @@ fn paste_clipboard_impl(
     let (view, doc) = current!(editor);
     match editor.clipboard_provider.get_contents(clipboard_type) {
         Ok(contents) => {
-            paste_impl(&[contents], doc, view, action, count, editor.mode);
+            let has_equal_contents = editor.last_clipboard[clipboard_type as usize]
+                .join(doc.line_ending.as_str())
+                .eq(&contents);
```
pascalkuthe, line 3792: Instead of joining `last_clipboard` here (which allocats and therefore causes some overhead) you could iterate over it and compare `contents` on the fly. That would roughly look like the following:

```suggestion
	let mut pos = contents;
	let mut last_clipboard = editor.last_clipboard[clipboard_type as usize].iter();
	let mut has_equal_contents = true;
	if let Some(first) = last_clipboard.next() {
		if pos.starts_with(first) {
			pos = &pos.slice(first.len()];
			for selection in last_clipboard {
				if pos.starts_with(doc.line_ending.as_str()) && pos[1..].starts_with(selection) {
					pos = &pos[1 + selection.len()..];
				}else{
					has_equal_contents = false;
					break;
				}
			}
		}
	} else {
		has_equal_contents = false
	}
```

this is just something I came up quickly to illustrate what I mean so the implementation might be buggy/not compile/is not super idiomatic but I hope you get the idea

CcydtN: I did some test with your idea and you are correct, it is much faster when there is large amount of data.
So, I just push a new commit to implement the idea. It uses a custom iterator for better readability, but the overall logic should remain the same.
Please take a look.

pascalkuthe:
this was included in https://github.com/helix-editor/helix/pull/6985 (which ahs been merged) so I am closing this as stale/duplicate of that. Thank you for your contributions, your idea here definitely helped improving our register handling!

PR #6893, opened 2023-04-26
helix-core/surround: `Syntax` aware

cole-h:
Prior to this change, attempting to replace the double quotes in `" asdf "` would fail with "Cursor on ambiguous surround pair".

Now, after running e.g. `:set-language rust` in a file with the contents `" asdf "`, it is now possible to replace the double quotes with `mr` as you could with any other pair.

This change essentially makes any pair that you can use `mm` on to go to the matching pair also able to be replaced with `mr` or deleted with `md` (whether they be built-in, or provided by a tree-sitter grammar).

---

I would like to add tests for this, but I don't know how I should go about doing this (specifically, creating e.g. a Rust tree-sitter `Syntax` instance). I noticed one instance in `helix-core/tests/indent.rs`, but....  it seemed kinda heavy-weight. I'd be happy to use that approach, but I'm hoping to learn about something that's less verbose. Suggestions / examples appreciated :)

### DIFF
--- a/helix-core/src/surround.rs
+++ b/helix-core/src/surround.rs
@@ -1,6 +1,8 @@
 use std::fmt::Display;
 
-use crate::{movement::Direction, search, Range, Selection};
+use crate::{
+    match_brackets::find_matching_bracket, movement::Direction, search, Range, Selection, Syntax,
+};
 use ropey::RopeSlice;
 
 pub const PAIRS: &[(char, char)] = &[
@@ -124,6 +126,7 @@ pub fn find_nth_closest_pairs_pos(
 /// or opening pair. `n` will skip n - 1 pairs (eg. n=2 will discard (only)
 /// the first pair found and keep looking)
 pub fn find_nth_pairs_pos(
+    syntax: Option<&Syntax>,
     text: RopeSlice,
     ch: char,
     range: Range,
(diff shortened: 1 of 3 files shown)


clo4:
This is a great change, I run up against this all the time. A much more robust solution than what vim plugins tend to do!

pascalkuthe:
For testing you can use the integration test framework which runs a full editors instance. You can then simply invoke the appropriate keybindings and compare the resulting text. An easy to understand example might be `helix-terms/tests/test/auto_indent.rs`. `helix-terms/tests/test/commands.rs` contains many more examples

cole-h:
Ah, that sounds like the best place then. I was sorta-kinda hoping to _not_ need to setup docker/podman again, but I'd much rather test this than not, so I guess podman wins!

I'll try to whip something up shortly.

pascalkuthe:
you don't need podman/docker for the integration tests AFAIK

cole-h:
Yeah, you're 100% right, I was thinking of another project I had once hacked on :sweat_smile: Sorry for the confusion; I'll gladly write a test regardless!

cole-h:
OK, integration tests added in da73d79d9b1b0d800e163d583a6d90a88fefdc53! Everything passes for me.

cole-h:
I think I get what you mean. I'll do some playing, and probably end up asking some questions in the Matrix room.

My TS-fu is non-existent, so the work will be slow going, but I'll try to keep this PR apprised of my current progress. (I'll likely end up making the history ugly in the process, but I'll squash it once it appears to work.)

shaleh:
What is needed to resurrect this?

helix-core/src/surround.rs
```diff
@@ -141,15 +144,26 @@ pub fn find_nth_pairs_pos(
...
-            // Cursor is directly on match char. We return no match
-            // because there's no way to know which side of the char
-            // we should be searching on.
-            return Err(Error::CursorOnAmbiguousPair);
+            // Cursor is directly on match char and we don't have any tree-sitter syntax available
+            // to help identify the matching pair. We return no match because there's no way to know
+            // which side of the char we should be searching on.
+            let syntax = syntax.ok_or(Error::CursorOnAmbiguousPair)?;
+
+            let match_pos = find_matching_bracket(syntax, &text.into(), pos)
```
shaleh, line 152: This line is the source of the current merge conflict. The RopeSlice can be passed on through now.

```suggestion
            let match_pos = find_matching_bracket(syntax, text, pos)
```

Issue #1596, opened 2022-01-28
Soft undo/redo (de)selection


eugenesvk:
Let's say I carefuly selected a few words between semicolons with a couple of useful `find_next_char` commands, and then accidentally move a cursor.

Would be great if there were a pair of `soft_undo` and `soft_redo` (similar to Sublime Text's) that would allow me to easily revert selection changes just like regular `undo`/`redo` operate on text change

samholmes:
I with this 100%. But what would the default key binding be for soft-undo/redo? Maybe `alt+u` and `alt+U` or `cmd+u` and `cmd+alt+u`?

kirawi:
As @EpocSquadron pointed out, relevant Kakoune PR: https://github.com/mawww/kakoune/pull/4725

zim0369:
I'm looking for something similar to vim's `gv`

samholmes:
I didn't know about `gv`. That will only get you to the last selection. The jump-list feature is the closest we have to a soft undo. Maybe we can configure motions to automatically add/remove to/from the jump-list.

yashpalgoyal1304:
This will also enable more frequent use of the `/` register for search... or say, easy recovery from accidental use of it via `n` thereof.

ref: #5784

mjs:
I’d like to have this feature and am interested in implementing it myself. I fairly regularly screw up a selection due to Vim muscle memory. 

One question I have is what should happen when the buffer has been changed such that it isn’t really possible to undo a selection?

A simple example: enter some text in a buffer, then %d. What should a soft undo do at this point given that the previous selection is no longer possible?

Perhaps soft undo should only be possible back to the last modification? I’ll take a look at Kakoune and Sublime and see what they do.

kirawi:
According to the PR description, it looks like they try to map the selection to the new buffer.

samholmes:
I think in VSCode, soft-undo will do a hard-undo when necessary.

mjs:
I've taken a look at what other editors do for their soft undo/redo feature.

### VSCode

The feature is called "Cursor undo" and it only works back to the last change. This is simple, pragmatic approach and works for the common "I just accidentally messed up my complicated selection" use case.

### Kakoune

Kakoune's soft undo and redo is accessed via C-h and C-k.

[...]

PR #6910, opened 2023-04-29
Only suggest automatic autocompletion after edit

freqmod:
When i move around in the document in insert mode i sometimes contemplate and wait a bit before continuing moving around. In this case i often get caught in an autocomplete dialogue and ends up changing the document without intention. I think it would be better to limit autocomplete to when a user has just changed the document, and not moved around.

I made a patch which made moving around with autocomplete enabled much more intuitive for me.
I guess it could be made an option if you don't want to change the default helix behaviour though.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3182,6 +3182,11 @@ pub mod insert {
                 _ => return,
             }
         }
+        /* Only autocomplete on timer after the user has input character,
+        not when they are moving around the document. */
+        if doc.moved_since_changed() {
+            return;
+        }
         super::completion(cx);
     }
 
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -172,6 +172,7 @@ pub struct Document {
 
     // when document was used for most-recent-used buffer picker
     pub focused_at: std::time::Instant,
+    moved_since_changed: bool,
 }
 
 /// Inlay hints for a single `(Document, View)` combo.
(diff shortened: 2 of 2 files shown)


the-mikedavis:
This is mostly the same as #5803. Tracking this with a bool toggled on the doc or view or editor seems ad-hoc to me, I'd like to find a more robust way to know when to trigger completion in insert mode. Maybe the implementation for #1596 would be relevant for this.

Also I think there may be some missing changes for this PR? `register_insert_mode_movement` isn't called so once `move_since_changed` is set to `false` it's never `true` again

pascalkuthe:
The way vscode handles this is by requesting completions on every trigger character.

Trigger characters are those send by the LSP and any word character (`\w`). From the spec:

> The additional characters, beyond the defaults provided by the client (typically  [a-zA-Z]), that should automatically trigger a completion request


If you read the spec carefully its quite clear that completion requests are mostly only intended to be send for trigger chars or manually (pressing c-x). There is not "timeout" completion trigger. These completion triggers are of course debounced. By setting the debounce timer to something very low (like 20ms) we would attain vscode-like behavior (without lots of wasted processing/werid behavior that a low idle timeout causes) and by setting the debounce timer to something higher (500ms). 

This also solves some unfortunate interactions right now that are caused by the fact that the idle timeout was originally added for completion but is now used by lots of things (inlay hints for example). For many of these applications I want to move to a similar similar scheme that I described for completions (custom debounced triggers). For that I am working on a general event/hook system for helix. Otherwise, all these custom hooks sprinkled throughout the codebase would cause too much clutter.

freqmod:
I am sorry, it seems like a line in set_selection was lost when i ported it from my own branch with support for selection undo.

In any case i have fixed it now, even though it sounds like you want to do this in another way anyhow.

debajyoti1990:
This would a very nice addition. Currently I always try to ensure I move around in normal mode.

PR #6918, opened 2023-04-30
Update docs for `move_visual_line_*`

sscheele:
Change j/k/up/down to be `move_visual_line_*` over `move_line_*`

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -29,33 +29,33 @@
 
 > NOTE: Unlike Vim, `f`, `F`, `t` and `T` are not confined to the current line.
 
-| Key                   | Description                                        | Command                           |
-| -----                 | -----------                                        | -------                           |
-| `h`, `Left`           | Move left                                          | `move_char_left`                  |
-| `j`, `Down`           | Move down                                          | `move_visual_line_down`           |
-| `k`, `Up`             | Move up                                            | `move_visual_line_up`             |
-| `l`, `Right`          | Move right                                         | `move_char_right`                 |
-| `w`                   | Move next word start                               | `move_next_word_start`            |
-| `b`                   | Move previous word start                           | `move_prev_word_start`            |
-| `e`                   | Move next word end                                 | `move_next_word_end`              |
-| `W`                   | Move next WORD start                               | `move_next_long_word_start`       |
-| `B`                   | Move previous WORD start                           | `move_prev_long_word_start`       |
-| `E`                   | Move next WORD end                                 | `move_next_long_word_end`         |
-| `t`                   | Find 'till next char                               | `find_till_char`                  |
-| `f`                   | Find next char                                     | `find_next_char`                  |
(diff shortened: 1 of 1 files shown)


book/src/keymap.md
the-mikedavis, line 1: Could you undo the whitespace changes for the table? Formatting changes increase the chances that the changes here will cause conflicts for other PRs

sscheele: Sure - I'll just extend the two relevant lines and leave the others untouched so the right edge is ragged

pascalkuthe:
Could you also add the `gj` and `gk` binding in the relevant section?

book/src/keymap.md
the-mikedavis: Do the two lines need any extra space? I think there's enough room with the current right edge

sscheele:
> Could you also add the `gj` and `gk` binding in the relevant section?

Done!

book/src/keymap.md
```diff
@@ -218,6 +218,8 @@ Jumps to various locations.
 | `n`   | Go to next buffer                                | `goto_next_buffer`         |
 | `p`   | Go to previous buffer                            | `goto_previous_buffer`     |
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
+| `j`   | Go down file/logical (vs visual) line            | `move_line_down`           |
+| `k`   | Go up file/logical (vs visual) line              | `move_line_up`             |
```
pascalkuthe, line 222:
```suggestion
| `j`   | Move down textual (instead of  visual) line            | `move_line_down`           |
| `k`   | Move up textual (instead of visual) line              | `move_line_up`             |
```

I think we should also call it move down here to make it clear that this is the same as `j`/`k`. And I would call it `textual` line instead of `file/logical` line

sscheele: Yes, that seems better to me too

book/src/keymap.md
```diff
@@ -32,8 +32,8 @@
 | Key                   | Description                                        | Command                     |
 | -----                 | -----------                                        | -------                     |
 | `h`, `Left`           | Move left                                          | `move_char_left`            |
-| `j`, `Down`           | Move down                                          | `move_line_down`            |
-| `k`, `Up`             | Move up                                            | `move_line_up`              |
+| `j`, `Down`           | Move down                                          | `move_visual_line_down`            |
+| `k`, `Up`             | Move up                                            | `move_visual_line_up`              |
```
the-mikedavis, line 36:
```suggestion
| `j`, `Down`           | Move down                                          | `move_visual_line_down`     |
| `k`, `Up`             | Move up                                            | `move_visual_line_up`       |
```

sscheele: Oh yeah of course, not sure why I didn't see that haha

the-mikedavis (approved):

pascalkuthe (approved):

PR #6921, opened 2023-04-30
fix didChange notification offset encoding

pascalkuthe:
Fixes https://github.com/rust-lang/rust-analyzer/issues/14679

In https://github.com/helix-editor/helix/pull/5894 I overlooked that the incremental sync code manually transcodes char positions to utf16 line/column offset. This could cause server crashes for LSP servers that support utf8 or utf32 when editing on the same line after a non-ascii character.

This issue has been noticed by quite a few people both on matrix and in various github comments (for example https://github.com/helix-editor/helix/pull/6594#issuecomment-1519028627) altough we don't have an issue for it in our repo

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -645,7 +645,11 @@ impl Client {
         // Calculation is therefore a bunch trickier.
 
         use helix_core::RopeSlice;
-        fn traverse(pos: lsp::Position, text: RopeSlice) -> lsp::Position {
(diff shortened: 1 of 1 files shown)


poliorcetics:
Thanks for fixing it! If this isn't merged by Tuesday, I'll have the test to test it for a work day then

the-mikedavis (approved):

dead10ck (approved):
This is good to me!

Issue #6909, opened 2023-04-29
Utilize `.gitignore` outside of Git repositories


emilengler:
It would amazing, if the `.gitignore` file would be parsed, even if the directory is not a Git repository.

Certain projects do not use Git, but have a `.gitignore` in their source tree for exactly this reason. OpenBSD is a good for example for it.

pascalkuthe:
All that's needed to implement this is to add a `require_git` setting to the file picker settings so you can control `ignore::WalkBuilder::require_git`. That setting default to true in `ignore` which is the reason why `gitignore` file outside of git repositories are ignored. 

Note that we support global gitignore rules[^1] so setting this to false would cause those to be always active aswell so if we make this configurable in helix it should still default to `true` 

^1: well partially, ignore isn't particularly compliant with git

sscheele:
I can work on this over the next day or two!

sscheele:
I've created [a pull request](https://github.com/helix-editor/helix/pull/6923) with this feature!

janos-r:
I just want to note, that even currently, outside of a git folder, `.ignore` is correctly applied by default. Which is probably what most people want when looking for this.
Either way, this would be a simple and good addition. That whole ignore crate looks great.

tskardal:
I took [jj (jujutsu)](https://github.com/martinvonz/jj) for a spin this weekend, and noticed that Helix ignored .gitignore (🥁). Jujutzu uses .gitignore too, and even though it is git compatible, it has its own `.jj` folder instead of `.git`.

I worked around this by symlinking .gitignore as .ignore, but it would be nice if it was picked up automatically even though its not really a git repo.

Unsure about how often you have a .gitignore that you don't want to affect Helix. Maybe a better default would be to always parse .gitignore.

zschreur:
Is there any thing holding up either of the two PRs which address this being shipped? I too use `jj` and would like to not have to symlink a `.ignore` file.

Rudxain:
Some tools create repos by default, which is annoying when I just want to make a throw-away project (like a local "playground"). This is a somewhat regular occurrence to me:
```sh
cargo new tmp
cd tmp/
# forgot to pass `--vcs=none` to Cargo 🤦
nuke .git
mv .gitignore .ignore
# do stuff
hx src/main.rs
# in some rare cases, I realize I do want a repo!
git init
# rename AGAIN!
mv .ignore .gitignore
```

I even wrote this `fix-ign` script for less painful switching:
```sh
[...]

PR #6923, opened 2023-05-01
Add require-git option

sscheele:
Adds a `require-git` config option to the file picker and uses it to determine whether the `ignore::WalkBuilder` requires the current directory to be a git directory or not.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -190,6 +190,9 @@ pub struct FilePickerConfig {
     /// Enables reading `.git/info/exclude` files.
     /// Whether to hide files listed in .git/info/exclude in file picker and global search results. Defaults to true.
     pub git_exclude: bool,
+    /// Enables use of local and global .gitignore files even if not in a git directory
+    /// Whether to require that the workspace is in a git repository for .gitignore parsing. Defaults to true.
+    pub require_git: bool,
     /// WalkBuilder options
     /// Maximum Depth to recurse directories in file picker and global search. Defaults to `None`.
     pub max_depth: Option<usize>,
@@ -206,6 +209,7 @@ impl Default for FilePickerConfig {
             git_ignore: true,
             git_global: true,
             git_exclude: true,
+            require_git: true,
             max_depth: None,
         }
     }
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -177,6 +177,7 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> FilePi
         .git_global(config.file_picker.git_global)
         .git_exclude(config.file_picker.git_exclude)
         .max_depth(config.file_picker.max_depth)
+        .require_git(config.file_picker.require_git)
         .filter_entry(move |entry| filter_picker_entry(entry, &absolute_root, dedup_symlinks));
 
     // We want to exclude files that the editor can't handle yet


pascalkuthe:
The code changes loog good but this needs documentation in `book/SRC/configuration.md`

sscheele:
> The code changes loog good but this needs documentation in `book/SRC/configuration.md`

Done! I broke the convention of the book being the same as the comments, though, because this one has the less intuitive property that it enables a new feature when set to false rather than true.

book/src/configuration.md
```diff
@@ -169,6 +169,7 @@ All git related options are only enabled in a git repository.
 |`git-ignore` | Enables reading `.gitignore` files | true
 |`git-global` | Enables reading global `.gitignore`, whose path is specified in git's config: `core.excludefile` option | true
 |`git-exclude` | Enables reading `.git/info/exclude` files | true
+|`require-git` | Enables use of local and global .gitignore files even if not in a git directory (if set to `false`) | true
```
pascalkuthe, line 172: I think it's a but odd to describe a setting by it's inverse (si what disableing it changes). We usually just describe what the flag enable and then show the default aftwards. For example:


```suggestion
|`require-git` | Only use (local and global) .gitignore if inside a git repository/worktree | true
```

helix-view/src/editor.rs
```diff
@@ -190,6 +190,9 @@ pub struct FilePickerConfig {
     /// Enables reading `.git/info/exclude` files.
     /// Whether to hide files listed in .git/info/exclude in file picker and global search results. Defaults to true.
     pub git_exclude: bool,
+    /// Enables use of local and global .gitignore files even if not in a git directory
```
pascalkuthe, line 193: This should probably also not be inverse

book/src/configuration.md
sscheele: Ah OK, I was trying to keep the same phrasing of "Enables..." but this does make more sense. Will update this and comment.

helix-view/src/editor.rs
sscheele: Resolved in both files!

pascalkuthe (approved):
LGTM

archseer:
Does this need to be an option? Why not just turn this on by default?

sscheele:
> Does this need to be an option? Why not just turn this on by default?

Having an on-by-default behavior that invisibly uses the global gitignore seems pretty opaque to me. It might lead to confusion.

pascalkuthe:
> > Does this need to be an option? Why not just turn this on by default?
> 
> Having an on-by-default behavior that invisibly uses the global gitignore to decide what files to hide from you seems pretty opaque to me. It might lead to confusion.
> 
> Maybe ideally there are fully separate settings for global/local? This might be a slightly bigger change, but definitely doable.

It's not doable with ignore sadly. You would need to reimplement the git detection yourself

sscheele:
> You would need to reimplement the git detection yourself

Yes, I agree - though if we could then use ignore to actually do the parsing and filesystem walk, I think it would just be a function that steps up the directory tree looking for .git directories. Still, rolling our own git detection does feel like it's a little bit out of scope, and I think the behavior we have (whether or not to force git-like behavior regardless of git status) is pretty straightforward for users.

Niketin:
This pull request resolves the issue #6909 so it should probably be properly linked to it.

the-mikedavis (approved):

alevinval (approved):
Looks ready to ship

PR #6924, opened 2023-05-01
Adding Amberwood theme

gj1118:
Hi,  I have created this theme , called "Amberwood" . Its a subtle dark theme

image

### DIFF
--- a/runtime/themes/amberwood.toml
+++ b/runtime/themes/amberwood.toml
@@ -1,0 +1,135 @@
+
+"attribute" = { fg = "blue", modifiers = ["italic"] }
+"ui.virtual.wrap"="softwrap"
+"keyword" = "keyword"
+"keyword.control.conditional" = { fg = "conditional", modifiers = ["italic"] }
+"keyword.directive" = "magenta" # -- preprocessor comments (#if in C)
+
+"namespace" = { fg = "namespace", modifiers = ["italic"] }
+
+"punctuation" = "gray06"
+"punctuation.delimiter" = "gray06"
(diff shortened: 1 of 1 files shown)


runtime/themes/amberwood.toml
```diff
@@ -0,0 +1,135 @@
...
+
+"ui.popup" = { bg = "gray01" }
+"ui.window" = { fg = "gray02" }
+"ui.help" = { bg = "gray01", fg = "white" }
+
+"ui.text" = { fg = "fg" }
+"ui.text.focus" = { fg = "fg" }
+
+"ui.virtual" = { fg = "gray02" }
+"ui.virtual.indent-guide" = { fg = "gray02" }
```
the-mikedavis, line 79: You might want to add a `"ui.virtual.ruler"` key and style it with something like `{ bg = "gray02" }`. It inherits from `ui.virtual` by default so it will be a foreground ruler which can look odd. (Try `:set rulers [80]` to enable a ruler at 80 columns temporarily)

gj1118: Thank you @the-mikedavis for the pointer. much appreciated for the help. Please note that I have now updated the theme file with the changes and attached the updated screenshot above. Indeed without the change, it was difficult seeing where the ruler was. That was my bad. Sincere apologies, for the same. Thanks once again.

the-mikedavis (dismissed):

pascalkuthe (approved):

the-mikedavis (approved):

PR #6942, opened 2023-05-02
Rename till commands

### LINKED ISSUE #6929: Add `find_till_prev_char`
It seems there is no `find_till_prev_char` as counterpart to `find_till_[next_]char` (t).

In comparison, `find_prev_char` (F) exists as counterpart to `find_next_char`.

Also, renaming `find_till_char` to `find_till_next_char` would be more consistent to indicate its direction.

vwkd:
Fixes #6929

This renames the till commands `find_till_char`, `extend_till_char`, and `till_prev_char` to `find_till_next_char`, `extend_till_next_char`, and `find_till_prev_char`.

Currently, this keeps the existing commands intact as aliases to the renamed ones, adding "[DEPRECATED]" to the description and a TODO comments for scheduled removal after [TBD] versions (with current version being 23.03).

Opening to gather feedback.

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -43,7 +43,7 @@
 | `E`                   | Move next WORD end                                 | `move_next_long_word_end`   |
 | `t`                   | Find 'till next char                               | `find_till_char`            |
 | `f`                   | Find next char                                     | `find_next_char`            |
-| `T`                   | Find 'till previous char                           | `till_prev_char`            |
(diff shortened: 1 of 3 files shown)


the-mikedavis (changes requested):
Renaming these just for consistency isn't worth the breaking changes to configs. Eventually we want to replace the TOML configuration files with something more flexible and that would be the time to introduce breaking changes like this.

vwkd:
Please note, the PR as it stand now is not a breaking change since the old commands are kept as aliases.

PR #6947, opened 2023-05-02
Implement wbc and wbc! commands

### LINKED ISSUE #6857: Write and close buffer
I often find myself wanting to write and close a buffer, currently I have to write two commands namely :write (:w) and :buffer-close (:bc). Would it be possible to combine these into one command :write-buffer-close or :wbc for short.

Regards and thanks for the great editor.

l4l:
Resolves #6857

I'm not sure what are the policies are there wrt to new commands but this one feels useful to me as well.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -382,6 +382,36 @@ fn force_write(
     write_impl(cx, args.first(), true)
 }
 
+fn write_buffer_close(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    write_impl(cx, args.first(), false)?;
+
+    let document_ids = buffer_gather_paths_impl(cx.editor, args);
+    buffer_close_by_ids_impl(cx, &document_ids, false)
+}
+
+fn force_write_buffer_close(
+    cx: &mut compositor::Context,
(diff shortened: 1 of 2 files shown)


helix-term/src/commands/typed.rs
```diff
@@ -2291,6 +2321,20 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
...
+            name: "write-buffer-close",
+            aliases: &["wbc"],
+            doc: "Write changes to disk and closes the buffer. Accepts an optional path (:write some/path.txt)",
+            fun: write_buffer_close,
+            signature: CommandSignature::positional(&[completers::filename]),
+        },
+        TypableCommand {
+            name: "write-buffer-close!",
+            aliases: &["wbc!"],
+            doc: "Force write changes to disk creating necessary subdirectories and closes the buffer. Accepts an optional path (:write some/path.txt)",
```
the-mikedavis, line 2334: The docs here should mention ":write-close-buffer(!)" instead. It looks like force_write could also be updated to say ":write!" instead of ":write" but that isn't totally related to this PR

l4l: Right, thanks for noticing. Fixed the `write!` as well

the-mikedavis (approved):
Looks good 👍

I think this is fine to add. I don't use `:bc` myself much but I could see the use-case for these. Ideally it would be possible to build these with #4423 but that feature needs more design (and maybe the plugin system).

pascalkuthe (approved):

Issue #6878, opened 2023-04-25
Key sequence terminating character is not considered to start a new one


KnorrFG:
Hey,
first of all, let me thank you for Helix. I love it. <3

To the point: I have this in my config:

```toml
[keys.insert]
j = { k = "normal_mode"}
```

and typically this works. However, when I type the sequence `jjk` in insert mode, it will input jjk, instead of inputting j, and then exiting insert mode. This seems to happen because a character that terminates a possible key-sequence is not considered to be able to start a new one. 

This is quite annoying if you try to exit insert mode after typing a j as last character, like e.g. in `obj`. It's very minor, and you might not even consider it a bug, but I thought I'll bring it up.

ghc0:
It's not powerful for helix to handle key events. I mean `powerful`!

Go https://github.com/rvaiya/keyd .

Or more https://github.com/qmk/qmk_firmware .

kirawi:
Remapping is not relevant to this issue.

pascalkuthe:
This issue is specific to insert mode the description is kind of a bit missleading. The problem is that in insert mode inserting text usually causes it to be inserted (this is not considered a keybinding but rather just the definition of insert mode).  

To solve this issue we would need to add a special handling for the last pending key in https://github.com/helix-editor/helix/blob/096ed0ced449ff0355fed6220c762c4f7f28ec25/helix-term/src/ui/editor.rs#L857 and call `self.keymap.get` again for the last key (similar to  https://github.com/helix-editor/helix/blob/096ed0ced449ff0355fed6220c762c4f7f28ec25/helix-term/src/ui/editor.rs#L861) and only insert the char if `KeymapResult::NotFound` is returned

ghc0:
> Remapping is not relevant to this issue.

He wanna remap `jk` as `esc`.

Method 1: using helix
Method 2: using keyd/QMK
1. could use `jk` as `esc` in helix, shell, browser everywhere
2. could use `jk` as `esc` in modeNormal/Insert, modeMinor_window/space/...

pascalkuthe:
this is already existing functionality in helix and a simple bugfix for that.

Actually I think this bug doesn't just occur in insert mode but also in normal mode and visual mode. The general issue is that with `KeymapResult::Cancelled` we still need to try if any key is bound to the pressed key. so this is probably best addressed in `handle_keymap_event` directly

sscheele:
[...]

PR #6960, opened 2023-05-04
Properly handle case where last character in a cancelled command begins a new command

sscheele:
This is per issue #6878

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -853,7 +853,8 @@ impl EditorView {
                         commands::insert::insert_char(cx, ch)
                     }
                 }
-                KeymapResult::Cancelled(pending) => {
+                KeymapResult::Cancelled(mut pending) => {
+                    let last_evt = pending.pop().unwrap();
                     for ev in pending {
                         match ev.char() {
                             Some(ch) => commands::insert::insert_char(cx, ch),
@@ -865,6 +866,14 @@ impl EditorView {
                                 }
                             }
                         }
+                    }
+                    match self.handle_keymap_event(Mode::Insert, cx, last_evt) {
+                        Some(KeymapResult::NotFound) => {
+                            if let Some(ch) = last_evt.char() {
+                                commands::insert::insert_char(cx, ch)
+                            }
+                        }
+                        _ => (),
                     }
                 }
                 _ => unreachable!(),
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -830,17 +830,23 @@ impl EditorView {
             last_mode = current_mode;
         };
 
-        match &key_result {
+        match key_result {
             KeymapResult::Matched(command) => {
-                execute_command(command);
+                execute_command(&command);
             }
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -830,17 +830,23 @@ impl EditorView {
...
             KeymapResult::MatchedSequence(commands) => {
                 for command in commands {
-                    execute_command(command);
+                    execute_command(&command);
                 }
             }
-            KeymapResult::NotFound | KeymapResult::Cancelled(_) => return Some(key_result),
+            KeymapResult::NotFound => return Some(key_result),
+            KeymapResult::Cancelled(mut pending) => {
+                if !matches!(self.keymaps.get(mode, event), KeymapResult::NotFound) {
```
pascalkuthe, line 845: I meant that you should recursively call `handle_keymap_event` here. Just calling `keymapsmget` directly misses a bunch of important logic

pascalkuthe (commented):
Thanks for taking a stab at this! Looks good  mostly just one thing I noticed. However I would like to see some integration tests here. The integration testing framework is actually really week suited for something like this. Just great a test that inputs key sequences where the first key is types twice (like `]]p` and check that the command is executed directly (so the selection moves in that example).

sscheele:
Thanks! I've updated to code to do the recursive call (locally) and am working on the integration tests now.

sscheele:
I've updated the code to make the recursive call and added two test cases to `test/movement.rs`, but I think I might need some help with the tests, as the second case is failing for reasons I don't fully understand. I'm pretty sure my code is working and I've written the test itself wrong, but I'm having trouble understanding the formatting for the string that's supposed to represent the editor state. Could someone with more insight into this maybe explain where I've gone wrong?

sscheele:
I actually think that in addition to the integration test problem, there's unexpected behavior associated with my code here. I've been running this code for a bit and a weird thing is that when I type `j` and then a non-display character, they either register in reverse order (eg: `j` then `<up>` will move the cursor up and then insert a j) or the j gets ignored entirely (this happens if you switch modes, which makes sense - so `j` then `<esc>` simply exits to normal mode without moving down a line). I don't have time to work on a fix for this atm, but I didn't want this to get merged with a bug in it.

pascalkuthe:
> > I think this might be a bug with the integration testing Framework. It seems that `\n\n` gets turned into `\n` somehow.
> > CC @dead10ck
> 
> This is actually intentional, see #6156. It's a somewhat hacky way to be able to explicitly express the cursor being at the end of the line, on the EOF. I'm open to suggestions, if you can think of any ideas for a better way to do this.

Hmm I see, I will try to thinm of omething but nothing comes to mind right now. If just a single line ending si removed would just adding an extra newline `\n|]\n\n` work as a workaround?

dead10ck:
> Hmm I see, I will try to thinm of omething but nothing comes to mind right now. If just a single line ending si removed would just adding an extra newline `\n|]\n\n` work as a workaround?

Yes it would. I did that in a few tests when I made the change.

aromeronavia:
FWIW, tested this on my end, checked out @sscheele's branch and this worked 👍 
I was trying to fix this one on my end but didn't notice this already had a PR hehehe

Issue #6754, opened 2023-04-14
Helix leaves terminal in a bad state when killed


ccmtaylor:
### Summary

After killing helix because it was stuck (I inadvertently caused a global search in my entire home directory), the editor left the terminal (iTerm 2 on macOS) in a bad state: entering normal characters work, but control keys (e.g. C-a, C-e) echo escape codes instead of moving the shell cursor.

Possibly related to #2250, however mouse movement doesn't spam the terminal for me.

### Reproduction Steps

 
reproduction recording:
[![](https://asciinema.org/a/fYBfXjeU2vYkStvkZr2MsvzD8.png)](https://asciinema.org/a/fYBfXjeU2vYkStvkZr2MsvzD8)

I tried this:

1. `hx`
2. killall hx (either in a separate terminal, or within helix with `:sh`)
3. try to enter control characters in the terminal

I expected this to happen:

I expected the terminal to work normally after exiting Helix.

Instead, this happened:

the terminal echoes control characters. Trying to reset the terminal using `reset`, or starting and exiting Helix cleanly, does not return things to normal.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-04-14T14:09:58.690 helix_view::clipboard [DEBUG] Using pbcopy+pbpaste to interact with the system clipboard
2023-04-14T14:09:58.693 helix_tui::backend::crossterm [DEBUG] The keyboard enhancement protocol is supported in this terminal (checked in 2.146462ms)
2023-04-14T14:09:58.694 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-04-14T14:09:59.092 helix_term::application [DEBUG] received editor event: IdleTimer
2023-04-14T14:09:59.766 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-04-14T14:10:00.168 helix_term::application [DEBUG] received editor event: IdleTimer
2023-04-14T14:10:01.103 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-04-14T14:10:01.303 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-04-14T14:10:01.519 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
2023-04-14T14:10:01.629 helix_view::document [DEBUG] id 1 modified - last saved: 0, current: 0
[...]

PR #6962, opened 2023-05-04
Cleanup terminal state when killing program

Zenthial:
Addresses #6754, where killing the process does not reset the terminal fully.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -235,6 +235,7 @@ impl Application {
             signal::SIGCONT,
             signal::SIGUSR1,
             signal::SIGTERM,
+            signal::SIGINT,
         ])
         .context("build signal handler")?;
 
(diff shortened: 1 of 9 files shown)


helix-term/src/application.rs
```diff
@@ -446,7 +452,7 @@ impl Application {
     #[cfg(not(windows))]
     pub async fn handle_signals(&mut self, signal: i32) {
         match signal {
-            signal::SIGTSTP => {
+            signal::SIGTSTP | signal::SIGTERM | signal::SIGINT => {
```
the-mikedavis, line 455: This block handles suspending the process (with `C-z`) rather than terminating it. SIGTERM and SIGINT should be handled in a new block that calls `self.restore_term().unwrap()` and then exits the event loop. That will also need some changes in `event_loop_until_idle`

the-mikedavis (changes requested):

PR #6979, opened 2023-05-06
Add command expansions %key{body}

ksdrar:
Based on: https://github.com/helix-editor/helix/pull/3393

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -86,3 +86,4 @@
 | `:clear-register` | Clear given register. If no argument is provided, clear all registers. |
 | `:redraw` | Clear and re-render the whole UI |
 | `:move` | Move the current buffer and its corresponding file to a different path |
+| `:echo` | Print the processed input to the editor status |
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -213,8 +213,8 @@ Helix provides several variables that can be used when typing commands or creati
 | `%{filename}`   | The absolute path of the currently focused file. |
 | `%{dirname}`    | The absolute path of the parent directory of the currently focused file. |
 | `%{cwd}`        | The absolute path of the current working directory of Helix. |
-| `%{linenumber}` | The line number where the main cursor is positioned. |
-| `%{selection}`  | The text selected by the main cursor. |
+| `%{linenumber}` | The line number where the primary cursor is positioned. |
+| `%{selection}`  | The text selected by the primary cursor. |
 | `%sh{cmd}`      | Executes `cmd` with the default shell and returns the command output, if any. |
 
 ### Example
--- a/book/src/usage.md
+++ b/book/src/usage.md
@@ -8,6 +8,7 @@
 - [Selecting and manipulating text with textobjects](#selecting-and-manipulating-text-with-textobjects)
 - [Navigating using tree-sitter textobjects](#navigating-using-tree-sitter-textobjects)
 - [Moving the selection with syntax-aware motions](#moving-the-selection-with-syntax-aware-motions)
+- [Using variables in typed commands and mapped shortcuts](#using-variables-in-typed-commands-and-mapped-shortcuts)
 <!--toc:end-->
 
 For a full interactive introduction to Helix, refer to the
@@ -202,6 +203,26 @@ child node will be selected. In the event that `arg1` does not have a previous
 sibling, the selection will move up the syntax tree and select the previous
 element. As a result, using `Alt-p` with a selection on `arg1` will move the
 selection to the "func" `identifier`.
+
+## Using variables in typed commands and mapped shortcuts
+Helix provides several variables that can be used when typing commands or creating custom shortcuts. These variables are listed below:
+
+| Variable        | Description |
+| ---             | ---                      |
+| `%{basename}`   | The name and extension of the currently focused file. |
+| `%{filename}`   | The absolute path of the currently focused file. |
+| `%{dirname}`    | The absolute path of the parent directory of the currently focused file. |
+| `%{cwd}`        | The absolute path of the current working directory of Helix. |
+| `%{linenumber}` | The line number where the main cursor is positioned. |
+| `%{selection}`  | The text selected by the main cursor. |
+| `%sh{cmd}`      | Executes `cmd` with the default shell and returns the command output, if any. |
+
+### Example
+```toml
+[keys.normal]
+# Print blame info for the line where the main cursor is.
+C-b = ":echo %sh{git blame -L %{linenumber} %{filename}}"
+```
 
 [lang-support]: ./lang-support.md
 [unimpaired-keybinds]: ./keymap.md#unimpaired
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -202,13 +202,16 @@ impl MappableCommand {
                         jobs: cx.jobs,
                         scroll: None,
                     };
+
                     let args = args.join(" ");
 
                     match helix_view::editor::expand_variables(cx.editor, &args) {
                         Ok(args) => {
-                            let args = [args];
+                            let args = args.split_whitespace();
+                            let args: Vec<Cow<str>> = args.map(|it| Cow::Borrowed(it)).collect();
 
-                            if let Err(e) = (command.fun)(&mut cx, &args, PromptEvent::Validate) {
+                            if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate)
+                            {
                                 cx.editor.set_error(format!("{}", e));
                             }
                         }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -203,10 +203,10 @@ impl MappableCommand {
                 };
                 if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
                     let args = args.join(" ");
-                    let args = match typed::expand_args(cx.editor, &args) {
+                    let args = match helix_view::editor::expand_args(cx.editor, &args) {
                         Ok(a) => a,
                         Err(e) => {
-                            cx.editor.set_error(format!("{}", e));
+                            cx.editor.set_error(format!("{e}"));
                             return;
                         }
                     };
@@ -214,7 +214,7 @@ impl MappableCommand {
                     let args: Vec<Cow<str>> = args.split_whitespace().map(Cow::from).collect();
 
                     if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
-                        cx.editor.set_error(format!("{}", e));
+                        cx.editor.set_error(format!("{e}"));
                     }
                 }
             }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -201,12 +201,21 @@ impl MappableCommand {
                     jobs: cx.jobs,
                     scroll: None,
                 };
-                if let Err(e) = typed::process_cmd(
-                    &mut cx,
-                    &format!("{} {}", name, args.join(" ")),
-                    PromptEvent::Validate,
-                ) {
-                    cx.editor.set_error(format!("{}", e));
+                if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
+                    let args = args.join(" ");
+                    let args = match typed::expand_args(cx.editor, &args) {
+                        Ok(a) => a,
+                        Err(e) => {
+                            cx.editor.set_error(format!("{}", e));
+                            return;
+                        }
+                    };
+
+                    let args: Vec<Cow<str>> = args.split_whitespace().map(Cow::from).collect();
+
+                    if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
+                        cx.editor.set_error(format!("{}", e));
+                    }
                 }
             }
             Self::Static { fun, .. } => (fun)(cx),
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -208,7 +208,7 @@ impl MappableCommand {
                     match helix_view::editor::expand_variables(cx.editor, &args) {
                         Ok(args) => {
                             let args = args.split_whitespace();
-                            let args: Vec<Cow<str>> = args.map(|it| Cow::Borrowed(it)).collect();
+                            let args: Vec<Cow<str>> = args.map(Cow::Borrowed).collect();
 
                             if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate)
                             {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -196,25 +196,29 @@ impl MappableCommand {
     pub fn execute(&self, cx: &mut Context) {
         match &self {
             Self::Typable { name, args, doc: _ } => {
-                let mut cx = compositor::Context {
-                    editor: cx.editor,
-                    jobs: cx.jobs,
-                    scroll: None,
-                };
                 if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
-                    let args = args.join(" ");
-                    let args = match helix_view::editor::expand_args(cx.editor, &args) {
-                        Ok(a) => a,
-                        Err(e) => {
-                            cx.editor.set_error(format!("{e}"));
+                    let mut cx = compositor::Context {
+                        editor: cx.editor,
+                        jobs: cx.jobs,
+                        scroll: None,
+                    };
+
+                    let args: Vec<Cow<str>> = match args
+                        .iter()
+                        .map(|arg| {
+                            helix_view::editor::expand_variables(cx.editor, arg).map(Cow::from)
+                        })
+                        .collect()
+                    {
+                        Ok(expanded_args) => expanded_args,
+                        Err(err) => {
+                            cx.editor.set_error(err.to_string());
                             return;
                         }
                     };
 
-                    let args: Vec<Cow<str>> = args.split_whitespace().map(Cow::from).collect();
-
                     if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
-                        cx.editor.set_error(format!("{e}"));
+                        cx.editor.set_error(format!("{}", e));
                     }
                 }
             }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -196,16 +196,17 @@ impl MappableCommand {
     pub fn execute(&self, cx: &mut Context) {
         match &self {
             Self::Typable { name, args, doc: _ } => {
-                let args: Vec<Cow<str>> = args.iter().map(Cow::from).collect();
-                if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(name.as_str()) {
-                    let mut cx = compositor::Context {
-                        editor: cx.editor,
-                        jobs: cx.jobs,
-                        scroll: None,
-                    };
-                    if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
-                        cx.editor.set_error(format!("{}", e));
-                    }
+                let mut cx = compositor::Context {
+                    editor: cx.editor,
+                    jobs: cx.jobs,
+                    scroll: None,
+                };
+                if let Err(e) = typed::process_cmd(
+                    &mut cx,
+                    &format!("{} {}", name, args.join(" ")),
+                    PromptEvent::Validate,
+                ) {
+                    cx.editor.set_error(format!("{}", e));
                 }
             }
             Self::Static { fun, .. } => (fun)(cx),
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -202,24 +202,20 @@ impl MappableCommand {
                         jobs: cx.jobs,
                         scroll: None,
                     };
+                    let args = args.join(" ");
 
-                    let args: Vec<Cow<str>> = match args
-                        .iter()
-                        .map(|arg| {
-                            helix_view::editor::expand_variables(cx.editor, arg).map(Cow::from)
-                        })
-                        .collect()
-                    {
-                        Ok(expanded_args) => expanded_args,
+                    match helix_view::editor::expand_variables(cx.editor, &args) {
+                        Ok(args) => {
+                            let args = [args];
+
+                            if let Err(e) = (command.fun)(&mut cx, &args, PromptEvent::Validate) {
+                                cx.editor.set_error(format!("{}", e));
+                            }
+                        }
                         Err(err) => {
                             cx.editor.set_error(err.to_string());
-                            return;
                         }
                     };
-
-                    if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
-                        cx.editor.set_error(format!("{}", e));
-                    }
                 }
             }
             Self::Static { fun, .. } => (fun)(cx),
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -205,7 +205,7 @@ impl MappableCommand {
 
                     let args = args.join(" ");
 
-                    match helix_view::editor::expand_variables(cx.editor, &args) {
+                    match cx.editor.expand_variables(&args) {
                         Ok(args) => {
                             let args = args.split_whitespace();
                             let args: Vec<Cow<str>> = args.map(Cow::Borrowed).collect();
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2477,7 +2477,7 @@ pub fn process_cmd(
     event: PromptEvent,
 ) -> anyhow::Result<()> {
     let input: Cow<str> = if event == PromptEvent::Validate {
-        expand_args(cx.editor, input)?
+        helix_view::editor::expand_args(cx.editor, input)?
     } else {
         Cow::Borrowed(input)
     };
@@ -3203,107 +3203,6 @@ pub(super) fn command_mode(cx: &mut Context) {
     // Calculate initial completion
     prompt.recalculate_completion(cx.editor);
     cx.push_layer(Box::new(prompt));
-}
-
-static EXPAND_ARGS_REGEXP: Lazy<Regex> =
-    Lazy::new(|| Regex::new(r"%(\w+)\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap());
-
-pub fn expand_args<'a>(editor: &Editor, args: &'a str) -> anyhow::Result<Cow<'a, str>> {
-    let (view, doc) = current_ref!(editor);
-    let shell = &editor.config().shell;
-
-    replace_all(
-        Lazy::force(&EXPAND_ARGS_REGEXP),
-        Cow::Borrowed(args),
-        move |keyword, body| match keyword.trim() {
-            "val" => match body.trim() {
-                "filename" => Ok((match doc.path() {
-                    Some(p) => p.to_str().unwrap(),
-                    None => SCRATCH_BUFFER_NAME,
-                })
-                .to_owned()),
-                "dirname" => doc
-                    .path()
-                    .and_then(|p| p.parent())
-                    .and_then(|p| p.to_str())
-                    .map_or(
-                        Err(anyhow::anyhow!("Current buffer has no path or parent")),
-                        |v| Ok(v.to_owned()),
-                    ),
-                "line_number" => Ok((doc
-                    .selection(view.id)
-                    .primary()
-                    .cursor_line(doc.text().slice(..))
-                    + 1)
-                .to_string()),
-                _ => anyhow::bail!("Unknown variable: {body}"),
-            },
-            "sh" => {
-                let result = tokio::task::block_in_place(move || {
-                    helix_lsp::block_on(async move {
-                        let args = &expand_args(editor, body)?[..];
-
-                        let mut command = tokio::process::Command::new(&shell[0]);
-                        command.args(&shell[1..]).arg(args);
-
-                        let output = command
-                            .output()
-                            .await
-                            .map_err(|_| anyhow::anyhow!("Shell command failed: {args}"))?;
-
-                        if output.status.success() {
-                            String::from_utf8(output.stdout)
-                                .map_err(|_| anyhow::anyhow!("Process did not output valid UTF-8"))
-                        } else if output.stderr.is_empty() {
-                            Err(anyhow::anyhow!("Shell command failed: {args}"))
-                        } else {
-                            let stderr = String::from_utf8_lossy(&output.stderr);
-
-                            Err(anyhow::anyhow!("{stderr}"))
-                        }
-                    })
-                });
-
-                result.map_err(|it| {
-                    log::error!("{}", it.to_string());
-
-                    it
-                })
-            }
-            _ => anyhow::bail!("Unknown keyword {keyword}"),
-        },
-    )
-}
-
-// Copy of regex::Regex::replace_all to allow using result in the replacer function
-fn replace_all<'a>(
-    regex: &regex::Regex,
-    text: Cow<'a, str>,
-    matcher: impl Fn(&str, &str) -> anyhow::Result<String>,
-) -> anyhow::Result<Cow<'a, str>> {
-    let mut it = regex.captures_iter(&text).peekable();
-
-    if it.peek().is_none() {
-        return Ok(text);
-    }
-
-    let mut new = String::with_capacity(text.len());
-    let mut last_match = 0;
-
-    for cap in it {
-        let m = cap.get(0).unwrap();
-        new.push_str(&text[last_match..m.start()]);
-
-        let replace = matcher(cap.get(1).unwrap().as_str(), cap.get(2).unwrap().as_str())?;
-
-        new.push_str(&replace);
-
-        last_match = m.end();
-    }
-
-    new.push_str(&text[last_match..]);
-
-    replace_all(regex, Cow::Owned(new), matcher)
 }
 
 fn argument_number_of(shellwords: &Shellwords) -> usize {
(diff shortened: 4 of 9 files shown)


the-mikedavis (changes requested):
For this PR we only touch typable commands but we will probably want to re-use expansion elsewhere (for example custom commands (https://github.com/helix-editor/helix/pull/3393#issuecomment-1506091918) or exposing the file-type in formatter arguments (https://github.com/helix-editor/helix/pull/5626)). So I think we can formalize expansion a little by moving the replacement function into the editor module in helix-view. I'd also like to see a few unit tests of the regex.

We could also add some integration tests if we add an echo command:

```rust
// helix-term/src/commands/typed.rs
fn echo(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> anyhow::Result<()> {
        return Ok(());
    }

    cx.editor.set_status(format!("{}", args.join(" ")));
    Ok(())
}
```

(for example could be used like `:echo %val{filename}` or `:echo %sh{echo %{filename}}`)

helix-term/src/commands.rs
```diff
@@ -185,16 +185,17 @@ impl MappableCommand {
...
+                    editor: cx.editor,
+                    jobs: cx.jobs,
+                    scroll: None,
+                };
+                if let Err(e) = typed::process_cmd(
+                    &mut cx,
+                    &format!("{} {}", name, args.join(" ")),
+                    PromptEvent::Validate,
+                ) {
+                    cx.editor.set_error(format!("{}", e));
```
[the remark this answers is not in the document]
ksdrar: I think I got what you mean, can you check the latest commit please

helix-term/src/commands/typed.rs
```diff
@@ -2882,6 +2905,82 @@ pub(super) fn command_mode(cx: &mut Context) {
...
+
+        match keyword.trim() {
+            "val" => match body.trim() {
+                "filename" => doc
+                    .path()
+                    .and_then(|p| p.to_str())
+                    .map_or(Err(anyhow::anyhow!("Current buffer has no path")), |v| {
+                        Ok(v.to_owned())
+                    }),
+                "filedir" => doc
```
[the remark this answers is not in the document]
ksdrar: I think dirname migth be confused with cwd by users

helix-term/src/commands/typed.rs
```diff
@@ -2882,6 +2905,82 @@ pub(super) fn command_mode(cx: &mut Context) {
     cx.push_layer(Box::new(prompt));
 }
 
+fn expand_args(editor: &Editor, args: &str) -> anyhow::Result<String> {
+    let regexp = regex::Regex::new(r"%(\w+)\s*\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap();
+
+    let view = editor.tree.get(editor.tree.focus);
+    let doc = editor.documents.get(&view.doc).unwrap();
```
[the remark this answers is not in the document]
ksdrar: I change it to allow sh body to expand using recursion, but let me check if it can be done

helix-term/src/commands/typed.rs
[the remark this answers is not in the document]
ksdrar: Oh, didn't know that. In that case it's fine to change it

ksdrar:
%sh{body} uses `shell_impl` from helix-term/src/commands.rs which cannot be accessed from helix-view, what should be done in this case?

helix-term/src/commands/typed.rs
```diff
@@ -2199,6 +2199,47 @@ fn clear_register(
     Ok(())
 }
 
+pub fn process_cmd(
```
the-mikedavis, line 2202: This doesn't need to be a function on its own, it can be moved back into the prompt definition in `command_mode`

helix-view/src/editor.rs
```diff
@@ -689,7 +689,7 @@ pub struct WhitespaceCharacters {
 impl Default for WhitespaceCharacters {
     fn default() -> Self {
         Self {
-            space: '·',    // U+00B7
+            space: '·',   // U+00B7
```
the-mikedavis, line 692: This was recently addressed in #7013 so if you rebase then this line should disappear from the diff

helix-view/src/editor.rs
```diff
@@ -1715,3 +1715,106 @@ fn try_restore_indent(doc: &mut Document, view: &mut View) {
...
+                            Err(anyhow::anyhow!("{stderr}"))
+                        }
+                    })
+                });
+
+                result.map_err(|it| {
+                    log::error!("{}", it.to_string());
+
+                    it
+                })
```
the-mikedavis, line 1784: I don't think we need to log here since we're returning and handling the Err variant. `Editor::set_error` already logs each error message

helix-view/src/editor.rs
```diff
@@ -1715,3 +1715,106 @@ fn try_restore_indent(doc: &mut Document, view: &mut View) {
...
+    once_cell::sync::Lazy::new(|| {
+        helix_core::regex::Regex::new(r"%(\w+)\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap()
+    });
+
+pub fn expand_args<'a>(editor: &Editor, args: &'a str) -> anyhow::Result<Cow<'a, str>> {
+    let (view, doc) = current_ref!(editor);
+    let shell = &editor.config().shell;
+
+    replace_all(
+        once_cell::sync::Lazy::force(&EXPAND_ARGS_REGEXP),
```
the-mikedavis, line 1729:
```suggestion
        &EXPAND_ARGS_REGEXP,
```

the-mikedavis: Also, a minor nit that this constant should probably be called `REGEX` rather than `REGEXP` to be consistent with regex in the regex crate and other usages in Helix

helix-view/src/editor.rs
```diff
@@ -1715,3 +1715,106 @@ fn try_restore_indent(doc: &mut Document, view: &mut View) {
         doc.apply(&transaction, view.id);
     }
 }
+
+static EXPAND_ARGS_REGEXP: once_cell::sync::Lazy<helix_core::regex::Regex> =
+    once_cell::sync::Lazy::new(|| {
+        helix_core::regex::Regex::new(r"%(\w+)\{([^{}]*(\{[^{}]*\}[^{}]*)*)\}").unwrap()
+    });
+
+pub fn expand_args<'a>(editor: &Editor, args: &'a str) -> anyhow::Result<Cow<'a, str>> {
```
the-mikedavis, line 1724:
```suggestion
pub fn expand_variables<'a>(editor: &Editor, input: &'a str) -> anyhow::Result<Cow<'a, str>> {
```

This will eventually be used for things other than arguments

helix-view/src/editor.rs
ksdrar: I don't think it was changed, I'll save the file without formatting that line if you want

helix-term/src/commands/typed.rs
```diff
@@ -2879,6 +2879,13 @@ pub(super) fn command_mode(cx: &mut Context) {
             }
         }, // completion
         move |cx: &mut compositor::Context, input: &str, event: PromptEvent| {
+            let input: Cow<str> = if event == PromptEvent::Validate {
+                helix_view::editor::expand_variables(cx.editor, input)
```
ksdrar, line 2883: Should it print the error and return if `expand_variables` fails?

pascalkuthe:
> So I think we can formalize expansion a little by moving the replacement function into the editor module in helix-view. 

I personally would prefer this as a separate module so `helix-view/src/editor/substitute.rs` I think `editors.rs` is already a bit too much of  a kitchen sink and this is pretty orthogonal so it makes more sense as its own module IMO

ksdrar:
Like this?

```rust
// helix-view/src/editor.rs
mod variable_expansion;
pub use variable_expansion::expand_variables;
...
```

```rust
// helix-view/src/editor/variable_expansion.rs
pub fn expand_variables<'a>(...
```

pascalkuthe:
yeah exactly :+1:

pascalkuthe (approved):
LGTM :+1:  Just missing some tests

jlebon:
Awesome, looking forward to this! Can we document this somewhere, esp. the exact variable names supported?

Also, should this be marked as closing https://github.com/helix-editor/helix/issues/3134?

ksdrar:
Checking out https://github.com/helix-editor/helix/pull/7090 I think it's a good idea for users to be able to expand the variables while in `:command` mode and maybe even add the variables to the list of suggestions for autocompletion. What do you think?

ksdrar:
I just pushed this feature into a new branch of my fork (ce-prompt).

Input inside a Prompt (command_mode, file_picker, etc) can be expanded from variables to values using Ctrl + .

https://github.com/helix-editor/helix/assets/22417151/25befd25-8e7f-4195-b458-3f7b9c4d6c6a

helix-term/src/commands.rs
```diff
@@ -193,13 +193,27 @@ impl MappableCommand {
...
                         editor: cx.editor,
                         jobs: cx.jobs,
                         scroll: None,
                     };
+
+                    let args: Vec<Cow<str>> = match args
+                        .iter()
+                        .map(|arg| {
+                            helix_view::editor::expand_variables(cx.editor, arg).map(Cow::from)
+                        })
```
the-mikedavis, line 207:
```suggestion
                        .map(|arg| helix_view::editor::expand_variables(cx.editor, arg))
```

the-mikedavis (commented):
I think this is looking good. @archseer had the idea of replacing the `%val{varname}` syntax with just `%varname`, for example `%filename`, `%dirname`, etc.. `%sh{..}` would stay the same. I think Kakoune's syntax makes more sense for Kakoune which mostly integrates with scripting / shell. I don't forsee us needing other `%X{..}`s other than maybe one for the plugin system language (though maybe this whole syntax/feature would change with the plugin system). What do you think?

danillos:
@ksdrar Today I updated to use this new branch and this command is not working any more for me.

```
s = ':open %sh{~/.dotfiles/support/pipe.rb switch_to_spec %sh{pwd} %val{filename}}'
```

It opens 4 buffers:

1. `%sh{~/.dotfiles/support/pipe.rb`
2. `switch_to_spec`
3. `~/my-project-pwd`
4. `}`

ksdrar:
> I think this is looking good. @archseer had the idea of replacing the `%val{varname}` syntax with just `%varname`, for example `%filename`, `%dirname`, etc.. `%sh{..}` would stay the same. I think Kakoune's syntax makes more sense for Kakoune which mostly integrates with scripting / shell. I don't forsee us needing other `%X{..}`s other than maybe one for the plugin system language (though maybe this whole syntax/feature would change with the plugin system). What do you think?

I think it's better to leave it be `%val{...}` to avoid issues with something like `%filename.bak` or unintended replacements. Or maybe changing it to `%{varname}` to avoid `val` repetition.

David-Else:
@ksdrar Hope you are well. Is this nearly ready to go? I think it just misses some docs and implementing the-mikedavis's suggested change?

ksdrar:
Hello @David-Else, I just pushed a commit to "cmd-exp-2" branch (it's rebased to latest changes) in my fork changing the expansion method and also uses the proposed %sh{} and %{varname} changes. Can you or someone else try it, please.

David-Else:
@ksdrar I tried the latest pull request here with the `:sh echo %val{filename}` format and it works great. I use this script and the `gh` cli command:

```sh
ghpr() { GH_FORCE_TTY=100% gh pr list --limit 300 |
    fzf --ansi --preview 'GH_FORCE_TTY=100% gh pr view {1}' --preview-window 'down,70%' --header-lines 3 |
    awk '{print $1}' |
    xargs gh pr checkout; }
```

I am afraid I don't know how to pull in your branch `https://github.com/ksdrar/helix/tree/cmd-exp-2` using git commands so it creates a new branch in my helix source code repository on my local machine. If you could tell me I would be happy to try it out... or just push the changes here and I will test them :)

danillos:
@ksdrar I'll try it today. I'm a heavy user of this feature. But I'm using the previous branch, this one had some bugs the last time I tested.

ksdrar:
> @ksdrar I'll try it today. I'm a heavy user of this feature. But I'm using the previous branch, this one had some bugs the last time I tested, reported [here](https://github.com/helix-editor/helix/pull/6979#issuecomment-1603370943)

Thanks!, I tried your command and it works fine (I just tried parsing it as I obviously don't have your ruby script). But if you have any problem, report back please :)

ksdrar:
> @ksdrar I tried the latest pull request here with the `:sh echo %val{filename}` format and it works great. I use this script and the `gh` cli command to audition pull requests:
> 
> ```shell
> ghpr() { GH_FORCE_TTY=100% gh pr list --limit 300 |
>     fzf --ansi --preview 'GH_FORCE_TTY=100% gh pr view {1}' --preview-window 'down,70%' --header-lines 3 |
>     awk '{print $1}' |
>     xargs gh pr checkout; }
> ```
> 
> I am afraid I don't know how to pull in your branch `https://github.com/ksdrar/helix/tree/cmd-exp-2` using git commands so it creates a new branch in my helix source code repository on my local machine. If you could tell me I would be happy to try it out... or just push the changes here and I will test them :)

You may try the branch using the following commands.
`git remote add ksdrar https://github.com/ksdrar/helix`
`git pull ksdrar cmd-exp-2`
`git switch cmd-exp-2`

Or if you don't want to create a branch in your local machine (not recommended).
`git remote add ksdrar https://github.com/ksdrar/helix`
`git merge ksdrar/cmd-exp-2`

David-Else:
@ksdrar I followed your instructions and it works great! For example, this works perfectly:

```
:sh echo %{filename}:%{linenumber}
```
@danillos I think you are better qualified to test the expansions are all correct as per your post https://github.com/helix-editor/helix/pull/3393#issuecomment-1520996787

danillos:
@ksdrar I did some tests and found these issues:

Issue 1:
When executing  `:sh echo %sh{echo a}` it works showing `a`
but the same command in a keybind `t = ':sh echo %sh{echo a}'` it shows `%sh{echo a}`

Issue 2:
`:sh echo %sh{echo %{filename} %sh{pwd}}` it works
`:sh echo %sh{echo %{filename} %sh{pwd} %{filename}}` it doesn't works, it seems that any args after the inner %sh breaks the logic

ksdrar:
@danillos I just pushed a commit for the second issue. I tried and it was caused by not trimming the command output. Can you try it please.

About the first issue, the thing is the keybind arguments are parsed by separating the string and saving it as a vector, so the command arguments end up being ["%sh{echo", "a}"]. We would have to join the vector again and then separate it once more for it to work with the expansion. It was previously spotted by @the-mikedavis in https://github.com/helix-editor/helix/pull/6979#discussion_r1191574086. I think we should join them to be parsed.

danillos:
@ksdrar Confirmed, Issue 2 was fixed.

ksdrar:
> @ksdrar Confirmed, Issue 2 was fixed.

Thanks!. I just pushed a commit addressing the first issue, it should work as expected now.

danillos:
@ksdrar Issue 1 was fixed too.

But I'm not allowed to save changes, when I attempt to `:w`, it shows it:
CleanShot 2023-08-26 at 06 02 52@2x

David-Else:
I have an issue with the latest `cmd-exp-2` version. I can't use my shortcut `Z = { Z = ":wq", Q = ":q!" }`, it just says `quit! takes no arguments`, probably the same thing as https://github.com/helix-editor/helix/pull/6979#issuecomment-1694238823

ksdrar:
Yeah, I just pushed another commit. Now it should work fine

David-Else:
> Yeah, I just pushed another commit. Now it should work fine

Fixed!

David-Else:
Can this be used in `languages.toml`? Something like `formatter = { command = 'prettier', args = [ "%{filename}" ] }
`? I can't really experiment with this until https://github.com/helix-editor/helix/pull/6846 is merged.

ksdrar:
> Can this be used in `languages.toml`? Something like `formatter = { command = 'prettier', args = [ "%{filename}" ] } `? I can't really experiment with this until #6846 is merged.

Not really. There's https://github.com/helix-editor/helix/pull/5626 and it's only being held back because the variable declaration syntax (%val{varname} or %{varname}, etc) is not set in stone yet afaik. I think once this PR is merged the formatter one will follow.

danillos:
@ksdrar, everything is working fine on the branch `cmd-exp-2`.

David-Else:
Currently, `filename` provides the absolute path, while the `dirname` command gives the path up until the filename. Is it possible to obtain just the filename? Wouldn't it be more logical for the `filename` to only provide the file name? We could easily create the absolute path by using `dirname filename`, and get the bonus of having access to just the file name. Just an idea, I am probably missing some context here.

quantonganh:
> Is it possible to obtain just the filename?

You can use [basename](https://linux.die.net/man/1/basename):

```
:sh basename %val{filename}
```

ksdrar:
I just added `basename` and `selection` variables with the latest commit to "cmd-exp-2". But I think renaming `filename` to `filepath` and `basename` to `filename` would make their meaning clearer for some users, what do you think?

David-Else:
> I just added `basename` and `selection` variables with the latest commit to "cmd-exp-2". But I think renaming `filename` to `filepath` and `basename` to `filename` would make their meaning clearer for some users, what do you think?

I agree.

David-Else:
@ksdrar I saw you requested a review, but I think you forgot to push the new `cmd-exp-2` here.

ksdrar:
Yeah, sorry about that. Now the changes are merged with this branch and I added integration tests for variables

jiehuifang:
Helix will merge this pull request in version 23.09 ?

MDeiml:
@ksdrar what's the state of this PR? Just waiting for review?

rcorre:
Looks like the new test is failing on mac: 
https://github.com/helix-editor/helix/actions/runs/6103715977/job/16568045931?pr=6979#step:7:768

zaquestion:
@the-mikedavis is this in a rereviewable state or one where the unittests can be run?

the-mikedavis:
Looks like one of the integration tests fails on windows and the docs need to be updated with `cargo xtask docgen` to add the new command. I'd like to give this another look soon though

zaquestion:
So the windows error (just one) seems to be related to the behavior of [std::fs::canonicalize](https://doc.rust-lang.org/std/fs/fn.canonicalize.html)  which returns windows paths using [extended path syntax](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file).

```
thread 'test::commands::variable_expansion::test_variable_expansion' panicked at 'assertion failed: `(left == right)`
  left: `"C:\\Users\\runneradmin\\AppData\\Local\\Temp\\.tmpr1kjJM"`,
 right: `"\\\\?\\C:\\Users\\runneradmin\\AppData\\Local\\Temp\\.tmpr1kjJM"`', helix-term\tests\test\commands\variable_expansion.rs:29:9
 ```
 The right in this case being what is we are expecting. We can see that the path we actually returned is otherwise correct just not using this syntax. Poking at the code, it seems like we are using paths returned out of helix's Editor crate, which is probably the path type we want to check against (not extended path syntax)? If so, would patching the test to return strip the extended path prefix be suitable here? Perhaps we want to update Editor to return the extended paths? 
 
 If it's just the former I can probably PR this branch with those changes, I might struggle a bit with the later but willing to try with some guidance.

**Comment by pascalkuthe** [comment]
using `std::fs::cannonicalize` is (almost) always a bug you should be using the canonicalization functions from helix core (so canonicalization is always consistent and correct)

**Comment by YurkoHoshko** [comment]
QQ: does this PR require just a doc update and using Helix Core canonicalization functions in test instead of standard library one? If so - I could take a stab at it (not a Rust dev myself, but would love this PR to get merged :))

**Comment by ksdrar** [comment]
Sorry for the delay, I just pushed the docs update and canonicalize changes. Hope it works now.

**Comment by jhugs** [comment]
@the-mikedavis @pascalkuthe can we get some eyes on his changes? Super hyped for this PR! 🎉

**Comment by rcorre** [comment]
Would there be any way to get the primary selection as a line range, as mentioned [in this comment](https://github.com/helix-editor/helix/issues/3134#issuecomment-1426286074)?

I'd really love to use `gh browse` to link to a block of code. Right now it looks like you can get a single line number, or the selection as text, but not the selection as a line range.

File: helix-term/src/ui/mod.rs [hunk 11]
```diff
@@ -382,6 +382,13 @@ pub mod completers {
         })
     }
 
+    pub fn variables(_: &Editor, input: &str) -> Vec<Completion> {
+        fuzzy_match(input, helix_view::editor::VARIABLES, false)
+            .into_iter()
```
the-mikedavis, line 387: Would the available variables ever change depending on the state of Editor? I think this can drop the `&Editor` parameter

helix-term/tests/test/commands/variable_expansion.rs
```diff
@@ -0,0 +1,81 @@
+use super::*;
+use helix_view::editor::expand_variables;
```
the-mikedavis, line 2: Integration tests shouldn't be calling functions like this directly. Instead these should use `test_key_sequence` with `:echo` and `Editor::get_status`

helix-view/src/editor/variable_expansion.rs
```diff
@@ -0,0 +1,149 @@
...
+        }
+    }
+
+    if let Some(o) = output.as_mut() {
+        o.push_str(&input[last_push_end..]);
+    }
+
+    match output {
+        Some(o) => Ok(std::borrow::Cow::Owned(o)),
+        None => Ok(std::borrow::Cow::Borrowed(input)),
```
the-mikedavis, line 147:
```suggestion
        Some(o) => Ok(Cow::Owned(o)),
        None => Ok(Cow::Borrowed(input)),
```

since `Cow` is already in scope

helix-view/src/editor/variable_expansion.rs
```diff
@@ -0,0 +1,149 @@
...
+                                        .and_then(|it| it.to_str())
+                                        .unwrap_or(crate::document::SCRATCH_BUFFER_NAME)
+                                        .to_owned(),
+                                    "dirname" => doc
+                                        .path()
+                                        .and_then(|p| p.parent())
+                                        .and_then(std::path::Path::to_str)
+                                        .unwrap_or(crate::document::SCRATCH_BUFFER_NAME)
+                                        .to_owned(),
+                                    "cwd" => std::env::current_dir()?.to_str().unwrap().to_owned(),
```
the-mikedavis, line 55:
```suggestion
                                    "cwd" => helix_std::env::current_working_dir().to_str().unwrap().to_owned(),
```

that function will be available if you rebase/merge-master

helix-view/src/editor/variable_expansion.rs
```diff
@@ -0,0 +1,149 @@
...
+    "%sh{}",
+    "%{basename}",
+    "%{filename}",
+    "%{dirname}",
+    "%{cwd}",
+    "%{linenumber}",
+    "%{selection}",
+];
+
+pub fn expand_variables<'a>(editor: &Editor, input: &'a str) -> anyhow::Result<Cow<'a, str>> {
```
the-mikedavis, line 15: Let's make this a member function of Editor:

```rust
impl Editor {
    pub fn expand_variables<'a>(&self, input: &'a str) -> anyhow::Result<Cow<'a, str>> {
        // ...
    }
}
```

helix-term/src/commands/typed.rs
```diff
@@ -2471,6 +2471,59 @@ fn move_buffer(
...
+    }
+
+    let args = args.join(" ");
+
+    cx.editor.set_status(args);
+
+    Ok(())
+}
+
+pub fn process_cmd(
```
the-mikedavis, line 2486: Looks like this function is no longer used and can be removed

book/src/generated/typable-cmd.md
the-mikedavis, line 1: Aside from these docs we should document the variable expansion somewhere in the book. I think a new section in `book/src/usage.md` would probably be the right spot

the-mikedavis (commented):
Thanks for continuing to work on this and sorry it's been a while since my last review. This is overall looking pretty good. I left a few comments.

Also \cc @archseer I'm curious what you think of the `%{variable}` syntax over `%variable`. I don't have strong opinions either way and I see how `%variable` might be awkward if you're concatenating the variable with other strings.

helix-view/src/editor/variable_expansion.rs
cole-h: (you probably meant `helix_stdx::env::....`, not `helix_std::env::....`)

ksdrar:
I just pushed the changes, can you review again please?

book/src/usage.md
```diff
@@ -203,6 +204,26 @@ sibling, the selection will move up the syntax tree and select the previous
...
+Helix provides several variables that can be used when typing commands or creating custom shortcuts. These variables are listed below:
+
+| Variable        | Description |
+| ---             | ---                      |
+| `%{basename}`   | The name and extension of the currently focused file. |
+| `%{filename}`   | The absolute path of the currently focused file. |
+| `%{dirname}`    | The absolute path of the parent directory of the currently focused file. |
+| `%{cwd}`        | The absolute path of the current working directory of Helix. |
+| `%{linenumber}` | The line number where the main cursor is positioned. |
+| `%{selection}`  | The text selected by the main cursor. |
```
the-mikedavis, line 217:
```suggestion
| `%{linenumber}` | The line number where the primary cursor is positioned. |
| `%{selection}`  | The text selected by the primary cursor. |
```

helix-term/src/ui/mod.rs
the-mikedavis: It looks like this hasn't been resolved yet

ksdrar: You mean remove the `&Editor` parameter from the function? I can't do it without breaking the `pub type Completer = fn(&Editor, &str) -> Vec<Completion>` requirement

the-mikedavis: Ah right good catch. Let's leave it as-is :+1:

the-mikedavis (approved):

jhugs:
@pascalkuthe @archseer can we get another review? ❤️ (I'm sorry I'm super impatient for this one 😆)

helix-term/src/commands.rs
```diff
@@ -196,16 +196,29 @@ impl MappableCommand {
...
                     };
-                    if let Err(e) = (command.fun)(&mut cx, &args[..], PromptEvent::Validate) {
-                        cx.editor.set_error(format!("{}", e));
-                    }
+
+                    let args = args.join(" ");
+
+                    match cx.editor.expand_variables(&args) {
+                        Ok(args) => {
+                            let args = args.split_whitespace();
```
pascalkuthe, line 210: I am pretty sure this is not correct. We allow quoting strings so arguments can contain whitespace and a simple split like this will not work. This seems like entirely the wrong place to handle command 3xpansion. This probably needs to be integrated into the command parser properly so that it can correctly interact with whitespace and escapes(without requiring quotes everywhere) so into the shellwords parser somehow. Altough that should likely not actually handle the expansion since it's also used for auto completion. This PR also doens't seem to handle completions at all (`:open %{dirname}/` will not offer useful completions). It should be possible to to properly implement that altough it may be tricky (%sh should not be executed for completions tough)

carterols:
What’s the status of this? This is the one feature that is keeping me from using helix, as I can’t bind any key maps to run custom commands involving the current file name.

jhugs:
> What’s the status of this? This is the one feature that is keeping me from using helix, as I can’t bind any key maps to run custom commands involving the current file name.

@carterols I'm also eagerly awaiting this, but depending on what you want to do, there is a "work around" in which you can parse the current file name via terminal if the one you're using has the functionality (I'm using kitty, the OP used [wezterm](https://github.com/quantonganh/helix-wezterm/blob/92e3ffc7f389c4508c4090564d7181b5bc855fd8/helix-wezterm.sh#L6C1-L6C9))

neduard:
Hey all!  I've been looking for this feature as well (thank you all that have put it together so far!!) and I'm wondering if what we have atm is a bit overkill?  I'm thinking we already have the `:sh` command which is great and the only thing we are really missing are some variable expansions when running it.

For example, `git blame` can be run as `:sh git blame -L %{linenumber} {%filepath}`.

One of the things I really like about helix is that it values a clean code base and doesn't "try to be everything for everyone".  As such I'm wondering if we can simplify (or at least consider having a simpler solution and avoid some of the edge cases):
* `%filepath` type variables I think are ok, `%filepath.bak` is actually totally valid and should expand to the filepath with `.bak` extension.  This is in-line with what shells do eg ` f=123; echo $f.bak` outputs `123.bak` (although might be slightly more difficult to implement?)
* we can drop the `:echo` and `%sh{}` as we already have `:sh`

I'm also wondering if `filepath` / `dirpath` are more readable than `filename` / `dirname`?  To me `filename` implies the name, not the full path.

What do people think?  I've only been using helix for a few months (and loving it! it has pretty much replaced like 12 years of vim) but this is my first time here so really sorry if I'm talking complete rubbish haha! :see_no_evil:

postsolar:
> we can drop the `:echo` and `%sh{}` as we already have `:sh`

How would I then do something like `:open %sh{ fzf }`, without resorting to temporary buffers?

carterols:
> > we can drop the `:echo` and `%sh{}` as we already have `:sh`
> 
> How would I then do something like `:open %sh{ fzf }`, without resorting to temporary buffers?

I understand this use case, but maybe this PR can focus on getting basic support for things like `%{filepath}` and we can later tackle the `%sh{}` expansions and other, more complex, situations?

As a Helix user, I would much rather have features now that cover most use cases than trying to lump everything together in one PR that takes a long time to get approved. I really could use this feature because at work, I am forced to use Perforce as my SCM, and you have to manually checkout a file to make it writable via a `p4 edit <filepath>`. Without the command expansion, I have to manually type this in every time with the full file path, whereas in neovim, I’ve bound this action to a key binding that automatically expands the current file path that is open.

danillos:
> we can drop the :echo and %sh{} as we already have :sh

I have a case that that uses `%sh{}` and I use it a lot every day, so I would like to it not be removed.

It uses `%sh{}` to get the current dir, and to run a script that will find the unit test from a file, of the file from a unit test, and after it sends to helix open it.
```
s = ':open %sh{~/.dotfiles/helix_helper.rb switch_to_spec %sh{pwd} %{filename}}'
```

And I have many commands like this, using `%sh{pwd}`

```
B = ':sh ~/.dotfiles/helix_helper.rb rspec_file %sh{pwd} %{filename}:%{linenumber}'
```

David-Else:
`:echo %sh{git blame -L %{linenumber} %{filename}}` vs `:sh git blame -L %{linenumber} {%filepath}` demonstrates that there is a danger of things becoming needlessly complex. Having `sh` and `%sh` is a potential redundancy that really needs more explanation in the docs, especially for people new to using shell commands.

The main question is what is delaying this very important feature getting merged? It is not much code, so maybe the implementation is still not meeting the core maintainers vision as it is?

rcorre:
@danillos could you not get `pwd` within `helix_helper.rb`?

danillos:
> @danillos could you not get `pwd` within `helix_helper.rb`?

Yes, it works thanks, I don't know why I didn't it before, now I can change to:

```
s = ':open %sh{~/.dotfiles/helix_helper.rb switch_to_spec %{filename}}'
```

postsolar:
There is also `%{cwd}` for this.

lizclipse:
I think it's worth keeping the `%sh{}` expansion as it allows a lot of flexibility when writing custom commands and integrating with other tools. At the very least, it's a handy 'get-out-of-jail' tool in that it allows you to drop into a full shell when building more complex arguments.

The main blockers on this seem to be the whitespace handling and completions. My view is that the WS handling is fine as is, especially since the last time I used `:set` it didn't handle whitespace properly (which was a week or 2 back a I think) and both can be fixed up later, and the completions can (and probably should) be done in a separate PR so that this one can get in and better tested.

Both points probably warrant TODOs, but I don't think should hold up and (otherwise) ready feature

noahfraiture:
Hope this will be merge soon, it would be a great workaround for many plugin

tapyu:
I am impatiently waiting for this PR be merged :eyes:

acaloiaro:
I know this PR has quite a few watchers and I don't want to create noise, but since it has been approved by @the-mikedavis for nearing six months and discussion has seemingly reached a stopping point, can someone say what it needs to be merged? 

Is the pending review from @archseer what the merge is pending? Or is it actually ready to go?

ksdrar:
Sorry guys, I've been busy with work but I'll try to find some time to finish this PR this week. I'll leave the completions feature for a future PR

tdaron:
Hey @ksdrar,

Would you mind if I (or someone else) try to finish this PR if you don't have time? It's almost ready to land according to the comments, and a lot of people are waiting for it to be completed. I'd like to help move it forward.

Thanks!

ksdrar:
Yeah @tdaron! That would be awesome, thanks for your help!

tdaron:
Yay !
I rebased your work on master and removed completion stuff (as it would be integrated in another PR) on my personal fork. 

What's currently missing on this to be ready?

ksdrar:
@pascalkuthe wants to integrate the expansion not in the execution phase but in parsing, but I think that it would require a simple AST or something like that to make the expansion lazy but keeping all the command parsing logic (white space handling, quotes, etc) as it is now. I don't have the path right now, but the structure ShellWords is the parsing phase iirc.

David-Else:
If https://github.com/helix-editor/helix/pull/11164 is now the official new version of this PR, then this should be closed and https://github.com/helix-editor/helix/pull/11164 added to the `next` category instead.

quatquatt:
> > What’s the status of this? This is the one feature that is keeping me from using helix, as I can’t bind any key maps to run custom commands involving the current file name.
> 
> @carterols I'm also eagerly awaiting this, but depending on what you want to do, there is a "work around" in which you can parse the current file name via terminal if the one you're using has the functionality (I'm using kitty, the OP used [wezterm](https://github.com/quantonganh/helix-wezterm/blob/92e3ffc7f389c4508c4090564d7181b5bc855fd8/helix-wezterm.sh#L6C1-L6C9))

Could you share your relevant Kitty workaround?

jhugs:
@quatquatt sure, both my own versions of the wezterm / kitty implementations are [here](https://github.com/jhugs/dotfiles-and-stuff/blob/main/bin/hx_ide) (the wezterm one is in a branch).

PR #6985, opened 2023-05-06
Refactor and expand special registers

the-mikedavis:
This implements some special registers from Kakoune:

* `#`: index numbers for the current selections
* `.`: contents of the current selections
* `%`: name of the current file (#5577)

And adds special registers for the system and primary clipboards: `*` and `+` respectively. This also re-uses the trick from #6889 to store the selections yanked to clipboards so they can be re-used on paste if the clipboard hasn't changed.

The strategy here mirrors how Kakoune implements these special registers and should extend to the remaining special registers in Kakoune (`1`..`9` for regex captures), but it's an unfortunately large refactor that adds complexity and allocations and needs updates for all register callsites. The overview is...

* `Register` moves to helix-view
* `read` functions now may return owned values. This is necessary because of the way some special registers produce their values. We use a `Cow` to prevent allocations in places where we don't need them.
* `read`, `first` and `last` functions now take the `Editor` as a parameter. This is the main change that enables special registers and is what mirrors Kakoune.

Closes #5577
Closes https://github.com/helix-editor/helix/pull/5242
Closes https://github.com/helix-editor/helix/issues/5241


helix-view/src/register.rs
```diff
@@ -0,0 +1,355 @@
...
+
+pub trait Register: std::fmt::Debug {
+    fn name(&self) -> char;
+    fn preview(&self) -> &str;
+
+    fn read(&self, editor: &Editor) -> Vec<String>;
+
+    fn write(&mut self, _editor: &mut Editor, _values: Vec<String>) -> Result<()> {
+        Ok(())
+    }
```
pascalkuthe, line 18: you provide a default implementation that returns `Ok(())` and therefore essentially acts like a blackhole register. I think a better default would be to return an error and then overwrite the default implementation for the blackhole register (so that it acts like a noop). Its not super critical but I think register that can't be written to should return an error instead of acting like a blackhole register.

helix-view/src/register.rs
```diff
@@ -0,0 +1,355 @@
...
+
+    fn push(&mut self, _editor: &mut Editor, _value: String) -> Result<()> {
+        Ok(())
+    }
```
pascalkuthe, line 22: (same here)

book/src/usage.md
```diff
@@ -37,19 +37,30 @@ If a register is selected before invoking a change or delete command, the select
...
 | `:`                | Last executed command |
 | `"`                | Last yanked text      |
-| `_`                | Black hole            |
+| `@`                | Last recorded macro   |
+
+### Special registers
 
-The system clipboard is not directly supported by a special register. Instead, special commands and keybindings are provided. Refer to the
-[key map](keymap.md#space-mode) for more details.
+Some registers have special behavior when read from or written to:
```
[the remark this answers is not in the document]
the-mikedavis: Yeah this distinction is cleaner in Kakoune because all special registers are effectively read-only (though the blackhole is meant to be written to). This PR uses `write`/`push` for the clipboard registers. I like the idea of two more columns for reading and writing behaviors 👍

helix-view/src/register.rs
the-mikedavis: This was just being lazy and not wanting to duplicate these 6 LoC for the read-only registers 😅. It's a weird default though so I'll just use these lines in the read-only registers

pascalkuthe: Being explicit is also good :+1: I would remove the default implementation entirely, in that case, though to avoid accidentally forgetting to implement the method

book/src/usage.md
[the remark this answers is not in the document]
the-mikedavis: Hmm actually I think it could make sense to separate out clipboard registers from special registers in the docs since all of the special registers are read-only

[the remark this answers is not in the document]
pascalkuthe: You can write to the blackhole register tough but we could just put that with the system clipboard and make one read-only and one read-write table.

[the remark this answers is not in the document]
the-mikedavis: In Kakoune at least you can write to all of the special registers and they silently discard the values like the blackhole register. We could return an Err though with a message that they're unwritable. What do you think?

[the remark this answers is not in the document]
pascalkuthe: I dont feel to strongly either way but to me it seems better to return an error here. If I want to dicard, I can use the black hole register.

All other registers are really not intended for writing and allowing it is more of a "why not". It kind of makes the black hole register a bit pointless too (you could just use `#` for the same thing mostly). So returning an error feels more consistent to me. 

If somebody used one of these special register for writing it would be more likely a typo instead of intentional (since then they would just use `_`) so an error feels like.better ui.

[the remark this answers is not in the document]
the-mikedavis: Ok I made this refactor. Lmk what you think about the defaults for `write` and `push` in this case. (Is there a nicer way to express this with a ReadOnlyRegister` trait of some sort?)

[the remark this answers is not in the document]
pascalkuthe: you could in theory add a second `ReadOnlyRegister` trait and then add an automatic register implementation that just forwards `Register::read` to `ReadOnlyRegister::read` and returns an error on `write`.

``` rust
impl<T: ReadOnlyRegister> Register for T{
 ...
}
```


The default implementations are a much simpler solution tough and the extra boilerplate might not be worth the extra clarity. I think the default implementations make a lot of sense :+1:

pascalkuthe (dismissed):

edrex:
> #: index numbers for the current selections
> .: contents of the current selections

Is the buffer safe state exposed somehow? That would be useful.  

Also, out of scope for this issue but is there any way to turn this into selection ranges? Maybe at some point a register could be exposed with some sort of Delta representation, like a CRDT or patch.

the-mikedavis:
I can't imagine use-cases for either of these, could you expand on that in a separate discussion? This is specifically about the Kakoune registers & clipboard registers

gibbz00:
When and how is the `#` ever used? I found this line, but I can't find where it set...

https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5480

gibbz00:
Is there a pipe register too? https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5204

pascalkuthe:
> When and how is the `#` ever used? I found this line, but I can't find where it set...
> 
> https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5480

that register was a special case for that command in the past this PR adds it as a general purpose register.



> Is there a pipe register too?
> 
> https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5204



> Is there a pipe register too?
> 
> https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5204

Not in this PR. I don't think that makes too much sense for us given that we don't have an interactive shell yet

the-mikedavis:
Yeah the existing line you linked is used for `"#<C-a>` or `"#<C-x>` so you can increment/decrement by selection number. With this PR you can use `"#P` to paste the selection numbers before each cursor.

gibbz00:
Huh, cool! Where `#` set?

helix-view/src/register.rs
```diff
@@ -0,0 +1,369 @@
...
+
+    match values.next() {
+        Some(first) if contents.starts_with(first) => {
+            contents = &contents[first.len()..];
+        }
+        _ => return false,
+    }
+
+    for value in values {
+        if contents.starts_with(line_ending) == contents[line_ending.len()..].starts_with(value) {
```
gibbz00, line 361: Shouldn't this be:

```suggestion
        if contents.starts_with(line_ending) && contents[line_ending.len()..].starts_with(value) {
```

If document line ending "\r" and content starts with "\n", hen the if statement is true when the remaining content isn't the same right? (false == false => true)

gibbz00: Might be wrong, but the entire function body might be replaced with something similar to this:
 ```rust
cached_value.join(editor.focused_document().line_ending.as_str()) == value
```

Comment (pascalkuthe) [reply helix-view/src/register.rs:361]: `join` alloates a new string so that's pretty inefficient which is why we iterate instead, You are correct about the and tough that looks like a typo

File: helix-view/src/register.rs [hunk 5]
```diff
@@ -0,0 +1,369 @@
...
+
+use crate::{clipboard::ClipboardType, document::SCRATCH_BUFFER_NAME, Editor};
+
+pub const SPECIAL_REGISTERS: [char; 6] = ['_', '#', '.', '%', '*', '+'];
+
+pub trait Register: std::fmt::Debug {
+    fn name(&self) -> char;
+    fn preview(&self) -> &str;
+
+    fn read(&self, editor: &Editor) -> Vec<String>;
```
the-mikedavis, line 14: I think we can get away with fewer allocations if we switch to returning `Cow<str>`:

```rust
fn read<'a>(&'a self, editor: &'a Editor) -> Vec<Cow<'a, str>>;
```

SimpleRegister doesn't need to allocate most of the special registers do.

pascalkuthe: yeah that would be nice, works especially well with the selection content regist (since that is a Cow anyway and we just turn it into a string). I think using a `Box<dyn Iterator>` instead of a Vec. Both aren't super efficient but the dynamic dispatch iterator is probably somewhat faster (and uses less memory)

helix-term/src/application.rs
```diff
@@ -65,6 +66,7 @@ pub struct Application {
     compositor: Compositor,
     terminal: Terminal,
     pub editor: Editor,
+    registers: Registers,
```
archseer, line 69: > Registers moves from a field on Editor to fields on the Context types to avoid borrowing conflicts.

This is because the trait takes `&mut Editor` on write and push right? And that's only needed for `editor.clipboard_provider`? Maybe we could instead move `clipboard_provider` to Application

the-mikedavis: Even if we downgrade to a shared reference there are some borrowing conflicts on write/push:

```
error[E0502]: cannot borrow `cx.editor.registers` as mutable because it is also borrowed as immutable
    --> helix-term/src/commands.rs:2333:27
     |
2333 |         if let Err(err) = cx.editor.registers.write(reg_name, cx.editor, values) {
     |                           ^^^^^^^^^^^^^^^^^^^^-----^^^^^^^^^^^---------^^^^^^^^^
     |                           |                   |               |
     |                           |                   |               immutable borrow occurs here
     |                           |                   immutable borrow later used by call
     |                           mutable borrow occurs here
```

since `write`/`push` take `&mut self` and `&Editor`. We could probably resolve this if we restricted special registers to take only `&self` and `&Editor` but then they become less powerful. For example we couldn't store the last value written to the clipboards which we use here to retain the selections.

Kakoune's `Registers` equivalent "RegisterManager" is unfortunately not very rusty: it's a singleton, see [here](https://github.com/mawww/kakoune/blob/e3122ab2c19248b590cb0143610439ec050766b7/src/register_manager.hh#L118) and [here](https://github.com/mawww/kakoune/blob/e3122ab2c19248b590cb0143610439ec050766b7/src/utils.hh#L11-L48).

the-mikedavis: It would be nice to move the clipboard provider out of Editor and drop the mutable reference anyways though. Putting it on `Application` makes it hard to access within `Register::push`/`Register::write`. We could put it behind a static (like [this](https://github.com/the-mikedavis/helix/commit/3d44c68d27f2029911187c0119af0c0cdd49decb)) which is a little ugly but eventually we can get rid of almost all references of it by reimplementing all clipboard stuff in terms of registers (like [this](https://github.com/the-mikedavis/helix/commit/59833329e09ba7ce971b498c52e4402670189928)).

helix-view/src/register.rs
```diff
@@ -0,0 +1,229 @@
...
+            _ => self
+                .inner
+                .get(&name)
+                .map(|values| Box::new(values.iter().map(Cow::from).rev()) as RegisterValues),
+        }
+    }
+
+    pub fn write(&mut self, name: char, mut values: Vec<String>) -> Result<()> {
+        match name {
+            '_' | '#' | '.' | '%' => Ok(()),
```
pascalkuthe, line 90: I noticed that you traet all special registers as a black hole regist for writing again. I think [previously](https://github.com/helix-editor/helix/pull/6985#discussion_r1193009812) we had discussed returning an error in that case. Like I said I don't feel super strongly about that but would slightly prefer it

helix-view/src/register.rs
```diff
@@ -0,0 +1,235 @@
...
+    pub fn read<'a>(&'a self, name: char, editor: &'a Editor) -> Option<RegisterValues<'a>> {
+        match name {
+            '_' => Some(Box::new(iter::empty())),
+            '#' => {
+                let (view, doc) = current_ref!(editor);
+                let selections = doc.selection(view.id).len();
+                // ExactSizeIterator is implemented for Range<usize> but
+                // not RangeInclusive<usize>.
+                Some(Box::new(
+                    (0..selections).map(|i| i.saturating_add(1).to_string().into()),
```
pascalkuthe, line 53:
```suggestion
                    (1..=selections).map(|i| i.to_string().into()),
```

nit this reads slightly better and there is no need to saturate. Since we can never get a value larger than selections (which it already a valid usize

helix-view/src/register.rs
```diff
@@ -0,0 +1,235 @@
...
+) -> RegisterValues<'a> {
+    match provider.get_contents(clipboard_type) {
+        Ok(contents) => {
+            // If we're pasting the same values that we just yanked, re-use
+            // the saved values. This allows pasting multiple selections
+            // even when yanked to a clipboard.
+            let Some(values) = saved_values else { return Box::new(iter::once(contents.into())) };
+
+            if contents_are_saved(values, &contents) {
+                Box::new(values.iter().map(Cow::from))
```
pascalkuthe, line 196:
```suggestion
                Box::new(values.iter().map(Cow::from).rev())
```

needs to be reversed for consistency (same as the other registers)

helix-view/src/register.rs
```diff
@@ -0,0 +1,235 @@
...
+            );
+
+            Box::new(iter::empty())
+        }
+    }
+}
+
+fn contents_are_saved(saved_values: &[String], mut contents: &str) -> bool {
+    let line_ending = NATIVE_LINE_ENDING.as_str();
+    let mut values = saved_values.iter();
```
pascalkuthe, line 217:
```suggestion
    let mut values = saved_values.iter().rev();
```

(same here)

helix-view/src/register.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    pub fn push(&mut self, name: char, value: String) -> Result<()> {
+        match name {
+            '_' => Ok(()),
+            '#' | '.' | '%' => Err(anyhow::anyhow!("Register {name} does not support pushing")),
+            '*' | '+' => {
+                self.inner.entry(name).or_insert_with(Vec::new).push(value);
+                self.clipboard_provider.set_contents(
+                    self.inner[&name].join(NATIVE_LINE_ENDING.as_str()),
```
pascalkuthe, line 118: This doesn't work correctly if the clipboard has changed outside of helix altough that might be ok. I am not sure how we want to handle that case?

the-mikedavis:
Pascal and I chatted about this and settled on something lighter-weight. The trait is replaced with `match`s on the chars. The `clipboard_provider` moves from Editor to Registers and all clipboard commands switch to using the new clipboard registers. With that change we avoid the borrowing conflict and can keep `registers` on `Editor` plus we delete some clipboard-specific chunks of code out of the commands module.

helix-view/src/register.rs
```diff
@@ -0,0 +1,235 @@
...
+    }
+
+    pub fn write(&mut self, name: char, mut values: Vec<String>) -> Result<()> {
+        match name {
+            '_' => Ok(()),
+            '#' | '.' | '%' => Err(anyhow::anyhow!("Register {name} does not support writing")),
+            '*' | '+' => {
+                self.inner.insert(name, values);
+                self.clipboard_provider.set_contents(
+                    self.inner[&name].join(NATIVE_LINE_ENDING.as_str()),
```
pascalkuthe, line 95: this also needs to be reversesd before being inserted (but not before being set to the
```suggestion
                let clipboard = values.join(NATIVE_LINE_ENDING.as_str());
                values.reverse();
                self.inner.insert(name, values);
                self.clipboard_provider.set_contents(
                    clipboard,
```

pascalkuthe:
maybe we rhould document somewhere that registers are stored in reverse so that we can push at the end? It would be nice to have some docs for that. Its obvious to me now but might be confusing for somebody new to the codebase

helix-view/src/register.rs
the-mikedavis: I would prefer this but ExactSizeIterator isn't implemented for `RangeInclusive<usize>` so we use a regular `Range<usize>` as a workaround. I assume it isn't implemented because the `ExactSizeIterator::len` could hypothetically be `usize::MAX + 1`

helix-view/src/register.rs
the-mikedavis: I opted to use the `read_from_clipboard` helper and then push/prepend to the values returned from that

pascalkuthe (approved):
I think this looks great! I have been daily driving this for a while and found no issues. I did another pass and this lgtm (although it has a couple of conflicts right now). This adds a bunch of (very) useful functionality while only adding 150LOC

archseer (dismissed):

pascalkuthe (dismissed):

archseer (approved):

PR #6996, opened 2023-05-08
Remove `tree-sitter-cabal`

ocharles:
I think I added this prematurely. The upstream parser isn't mature enough, and segfaults on quite a few routine edits of `.cabal` files (e.g., those generated by `hpack`, or files with merge conflicts). I think we should drop this until it's more mature.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2467,10 +2467,6 @@ roots = ["cabal.project", "Setup.hs"]
 indent = { tab-width = 2, unit = "  " }
 comment-token = "--"
 
-[[grammar]]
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
It probably makes sense to remove the query files as well. Those are often broken by updates to the grammar repository so they may not help highlighting when adding cabal back

ocharles:
Good shout, I'll do that.

ocharles:
I'll regen docs when I get a chance soon

the-mikedavis (approved):

PR #7011, opened 2023-05-10
Support use of the SUPER key as a modifier

john-h-k:
This allows creating custom keybindings using the super key (command on macOS, Windows/Linux key otherwise). It uses `X` as the key name as `C` is already taken.

I originally did this on my local helix build and purely for myself, but I decided I might as well open a PR for it in case others find it useful.


john-h-k:
Sorry, didn't see that. Thanks

Issue #6551, opened 2023-04-02
Ctrl + [ doesn't function as Escape in helix 23.03


xcpky:
I think this is because of the new enhanced keyboard protocol enabled in 23.03. Is there a way to disable it? Although I can remap Ctrl+[ to mode change command, escape has more functions than only mode change. And there is no way to remap a key combination to a specific keycode in helix, which is available in vim.

the-mikedavis:
Yeah, the enhanced keyboard protocol change was breaking if you are relying on ambiguous keycodes like `C-[`. IMO you should not rely on these ambiguous keycodes because it's essentially coincidence that they worked in the first place. And I don't like adding a config option for this since I think the use-case is very niche. Ideally this should be fixed by allowing remapping Escape in the remaining places where it is hard-coded (#5505).

On the other hand though, this could be covered by a config option which overrides the detection, `true` to assume the protocol is supported and `false` to assume not. (This has been discussed previously but I can't find the issue.)

wvwwvwwv:
`Ctrl + [` is the de facto standard `esc` alternative on macOS, and I don't think the usage is niche, because `Ctrl + [` is especially useful when your keyboard does not have an `esc` key (which is true for many Mac/iPad users). I'm really looking forward to the upcoming `esc` remapping feature.

mk-f:
I also don't think that using `ctrl+[` as `esc` is a niche use-case, on keyboards with us-layout it is more comfortable to reach (both pinkies) than the actual `esc`.

For what it's worth you can run hx inside tmux for the moment as a workaround.

jaese:
These two other [control character mappings](https://en.wikipedia.org/wiki/Control_character#How_control_characters_map_to_keyboards), backspace `ctrl-h` and line feed `ctrl-j`, are still supported after the enhanced keyboard protocol change, so I don't think it is a stretch to keep supporting `ctrl-[` as escape by default.

n8henrie:
With neither `ctrl-c` and `ctrl-[` working for this purpose, I'm left having to leave my home row to exit insert mode (via `Esc`). Is this by design, or am I missing a recommended alternative?

pascalkuthe:
you can always rebind the keymapping in your own config.

sdemura:
@pascalkuthe can you provide an example?

```
❯ hx
Bad config: TOML parse error at line 19, column 9
   |
19 | "C-[" = "esc"
   |         ^^^^^
No command named 'esc
```

[...]

PR #7020, opened 2023-05-11
Allow configuring enhanced keyboard protocol detection

the-mikedavis:
This can be used to force Helix to enable or disable support for the enhanced (Kitty) keyboard protocol. Force disabling support can be useful if you rely on ambiguous escape codes (C-[ is suprisingly popular). Force enabling this can be useful to avoid any overhead of the automatic detection, though this is usually unnoticeably small.

This fixes https://github.com/helix-editor/helix/issues/6551 but I think the proper way to support that workflow is to enable rebinding Escape in remaining places (#5505). I'll open this so anyone who wants to can use the changes but I'm not sure this should actually be merged.


the-mikedavis:
Superseded by https://github.com/helix-editor/helix/pull/14398

PR #7028, opened 2023-05-12
Fix TS freezing the editor

### LINKED ISSUE #338: Disabling tree-sitter on big files
Opening large files (e.g. `>100mb`) that have an associated syntax highlighting grammar leads to high memory usage and the file taking a while to load: https://github.com/tree-sitter/tree-sitter/issues/222

Ideally opening such a file would prompt the user with disabling syntax highlighting for the file.

### LINKED ISSUE #6697: 100% CPU usage in file picker
### Summary

Hi.
~I don't have a way for you to reproduce this issue~ (**edit:** now I do, see below), but when I type some specific string (in my case `build_`) in [one of my projects](https://github.com/rust-lang/rustc_codegen_gcc), the editor freezes and uses 100% CPU. `perf top` shows that this happens in the function `tree_sitter_rescript_external_scanner_scan`.

~Since I don't have a way to reproduce this issue, I can attempt to debug it, but I'd need some guidance.~

Thanks.

### Reproduction Steps

I tried this:

1. Download this [file](https://github.com/helix-editor/helix/files/11463606/res.txt) in a directory and rename it to `res`.
2. `hx`
3. `<Space>f` to open the file picker
4. Type `res` or whatever is needed to select the above file.

I expected this to happen:

Not freeze and not use 100% CPU.

Instead, this happened:

Helix freezes and uses 100% CPU.

### Helix log

Nothing in the logs.

### Platform

Linux

### Terminal Emulator

alacritty 0.12.0 (5a728195)

### Helix Version

helix 23.03

pascalkuthe:
Fixes #6697
Closes #338

This PR addresses TS syntax highlighting freezing the UI loop in two separate ways:
* It uses the parser timeout mechanism in TS to limit all TS parsing to 500ms. If parsing exceeds 500ms, its canceled and TS is automatically disabled for that buffer. This is preferable to a fixed size limit (as suggested in #338) as it takes different CPU speeds/grammar speeds into account (and future optimizations in the parser/TS). If parsing takes longer than 500ms the editor always feels laggy.
* In the picker syntax highlighting is now performed in a background thread to avoid micro stutters in the UI (if trying to move after a short a delay)

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -765,7 +765,11 @@ fn byte_range_to_str(range: std::ops::Range<usize>, source: RopeSlice) -> Cow<st
 }
 
 impl Syntax {
-    pub fn new(source: &Rope, config: Arc<HighlightConfiguration>, loader: Arc<Loader>) -> Self {
+    pub fn new(
+        source: &Rope,
+        config: Arc<HighlightConfiguration>,
+        loader: Arc<Loader>,
+    ) -> Option<Self> {
         let root_layer = LanguageLayer {
             tree: None,
             config,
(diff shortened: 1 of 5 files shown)


archseer:
> * This is preferable to a fixed size limit

I still think there should be an upper bound but we can set it to something really high (200~500MB?)

alexmo1997:
Is there any way to disable this?
I have some json files I would like to go through with helix (for which syntax highlighting would be great), and I'm fine with waiting a few seconds when opening them.

Issue #5700, opened 2023-01-27
unable to setup "yaml-language-server" configuration on helix


seungwoo-lee-518-at-proton-me:
I'd like to Enable Kubernetes on YAML via [yaml-language-server](https://github.com/redhat-developer/yaml-language-server).

However It's tricky (for me that I'm new to TOML Languages :cry: ) to add configuration on `languages.toml`, how can I add json configuration to enable this feature?

for example

```json
yaml.schemas: {
    "kubernetes": ["*.yaml"]
}
```

seungwoo-lee-518-at-proton-me:
![2023-01-27-233659_screenshot](https://user-images.githubusercontent.com/108667785/215112891-4619bed7-9413-4ff4-8698-2cc3cf614941.png)

Ah! I've found a solution from "nvim-lspconfig" Repository's [Issue](https://github.com/neovim/nvim-lspconfig/issues/1207)!

Here is my configuration to Enable Kubernetes from YAML.

```toml
[[language]]
name = "yaml"
file-types = ["yaml", "yml"]
config = { yaml = { schemas = { kubernetes = "/*.yaml" }, format = { enable = true } } }
```

But, when I type **`:format`** command on helix. It doesn't formatted. and on log (a.k.a `.cache/helix/helix.log`), helix got a malformed notification from LSP Server.

```
2023-01-27T23:48:26.086 helix_term::application [ERROR] received malformed notification from Language Server: Unhandled
2023-01-27T23:48:26.086 helix_term::application [ERROR] received malformed notification from Language Server: Unhandled
```

Anyway, It worked!

PR #7032, opened 2023-05-12
Default yaml-language-server schema example in `languages.toml`

Raunow:
Having an example in the `language.toml` on how to configure the default yaml language server with schemas will save someone else a good bit of tinkering time!
Especially since a major reason for a using the yaml language server is the schemas available.
Going digging through issues/discussions like: #5700 is an option, but a bit far removed from the `language.toml` already on your disk.

I decided on the Github Actions schema as that is unlikely to be polarizing, Helix is hosted on Github and makes use of Actions after all.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -867,6 +867,9 @@ indent = { tab-width = 2, unit = "  " }
 language-server = { command = "yaml-language-server", args = ["--stdio"] }
 injection-regex = "yml|yaml"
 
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Examples should live in the wiki instead of the built-in languages.toml https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

Raunow:
Oh, will add it there then!

PR #7034, opened 2023-05-13
Replace DAP vars popup, instead of adding new

A-Walrus:
Steps to reproduce original issue:
- Add breakpoint
- Start debugging
- Program stops at breakpoint
- List variables
- Step to next
- List variables again
Before:
image


pascalkuthe (approved):
thanks!

the-mikedavis (approved):

PR #7035, opened 2023-05-13
Add linewise-paste config option

zummenix:
Current behaviour of `paste` commands is to detect a line ending and perform the linewise paste. This prevents pasting text (that contains a line ending) in the middle of another line. By toggling `linewise-paste` config option user can control this behaviour.

Example:

I have a line `bc\n` in a default register and I want to paste it in the middle of line `a d`, I place cursor on `d` and type `P` (to paste before)

If `linewise-paste` is set to `true` (current default), the result will be:
```
bc
a d
```

 If `linewise-paste` is set to `false`, the result will be:
```
a bc
d
```

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3855,6 +3855,7 @@ fn paste_impl(
     action: Paste,
     count: usize,
     mode: Mode,
+    linewise: bool,
 ) {
     if values.is_empty() {
         return;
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I personally ave been missing this. While linewise paste can be useful I often prefer the more consistent "newline are not special" behaviour. However, it might be nice to turn this into two separate commands instead of a config option so its easier to use both (bind one to each key).

zummenix:
@pascalkuthe Sorry, I don't understand. Could you please clarify it a little bit more?

pascalkuthe:
Instead of adding a `linewise` config option, instead add new commands:
* `paste_after_no_line`
* `paste_before_no_line`
* ` paste_clipboard_after_no_line`
* `paste_clipboard_before_no_line`
* `paste_primary_clipboard_after_no_line`
* `paste_primary_clipboard_before_no_line`

These commands behave like your PR currently does with `paste-linewise = false`. This allows people to change the behaviour by changing their keymap (instead of adding a new config option). It also allows mixing the two behaviours (by binding to a special key for example or using the command pallet)

zummenix:
Got it, thanks! If someone wants to implement these commands please feel free to pick up this PR or close it in favour of a new one. At this point config option works best for me.

zummenix:
I'm dogfooding this for a couple of month now. I think that having a config option is also not optimal. Closing this and looking forward to view other solutions such as separate commands suggested in the PR by @pascalkuthe

Issue #5446, opened 2023-01-08
Symbol picker could use some extra symbol type indication


nxy7:
Currently symbol picker only prints out symbol name, but I think it could use some extra indication to know if You're dealing with function/struct/variable etc. etc.

Helix:
<img width="1130" alt="image" src="https://user-images.githubusercontent.com/32909532/211178113-02341d8b-f5ca-49af-9fa0-36b580077229.png">

VS Code (screenshot stolen from another issue):
![image](https://user-images.githubusercontent.com/32909532/211178128-f3e80ab5-19c1-4d70-9256-134146553e55.png)

I guess a good start would even be sign indication like
F - Function1
F - Function2
(I'm using class as I think more languages use classes than structs)
C - Struct1

If you think that's good idea I might start getting familliar with Helix codebase and try doing it myself.

PR #7040, opened 2023-05-13
feat: add symbol code prefix in symbol picker

dinhani:
Implements https://github.com/helix-editor/helix/issues/5446.

Not sure if this is the best approach, but I like that it allows me to quickly filter symbols by typing things like `func:` or `struc:`.

It would also be possible to generate the type indicator according to the `language_id`, so for example we could have `fn` for Rust file, `def` for Ruby, `fun` for Kotlin and defaulting to something else when not covered.


the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/5943

PR #7052, opened 2023-05-15
Add LSP parent module command

tarkah:
This adds the `experimental/parentModule` command that is supported by Rust Analyzer. 

I've bound it as a default keybind to `gu` for the moniker "go up". Feel free to suggest changing or removing this keybind since it's not something that will be supported across most LSPs. 

Proposal for this experimental command: https://github.com/microsoft/language-server-protocol/issues/1002

https://github.com/helix-editor/helix/assets/10239377/72bac496-3265-4ab2-9490-c3292dcbb620

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -546,6 +546,7 @@ impl Client {
                     ]),
                     ..Default::default()
                 }),
+                experimental: Some(experimental::capabilities()),
                 ..Default::default()
             },
             trace: None,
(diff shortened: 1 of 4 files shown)


the-mikedavis:
Language-specific or non-standard LSP methods should be implemented in plugins rather than in the core. Maintaining this on a branch on a fork with this change is fine until the plugin system is added but I don't think we want to merge this as-is.

tarkah:
Makes sense 👍 it's a super useful command but I understand the concern for separation. 

I'll keep this forked / rebased on my end for now and circle back when the plugin system is ready. Thanks!

PR #7053, opened 2023-05-15
Add command to merge non-consecutive ranges

hamrik:
Situations may arise in which you want to extend your selection up to a search result, or simply select everything between two or more cursors.

before
before

Currently there is no mechanism for this other than badly abusing `goto_last_modification` as a poor man's mark.[^1]
`merge_selections` combines all current ranges into a single range spanning from the first selection in the buffer to the last.

Now you can search in visual mode for the token you want to select up to, then hit `Alt-minus` (since `Alt-_` could already merge consecutive selections).

[helix_merge_ranges_macro_demo](https://github.com/helix-editor/helix/assets/18398499/728e9fad-2a8a-43b8-8d4f-46f1bf13df28)

[^1]: [How to select "until next occurrence"? - Reddit](https://www.reddit.com/r/HelixEditor/comments/13i527h/how_to_select_until_next_occurence/)

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -522,6 +522,16 @@ impl Selection {
         self
     }
 
+    // Replaces ranges with one spanning from leftmost to rightmost selection
+    pub fn merge_ranges(mut self) -> Self {
+        if !self.ranges.is_empty() {
+            let merged_range = self.ranges.iter().fold(self.ranges[0], |a, b| a.merge(*b));
+            self.ranges = smallvec![merged_range];
+            self.primary_index = 0;
+        }
+        self
+    }
+
     // Merges all ranges that are consecutive
     pub fn merge_consecutive_ranges(mut self) -> Self {
         let mut primary = self.ranges[self.primary_index];
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -522,14 +522,14 @@ impl Selection {
         self
     }
 
-    // Replaces ranges with one spanning from leftmost to rightmost selection
+    /// Replaces ranges with one spanning from first to last range.
     pub fn merge_ranges(self) -> Self {
         let first = self.ranges.first().unwrap();
         let last = self.ranges.last().unwrap();
         Selection::new(smallvec![first.merge(*last)], 0)
     }
 
-    // Merges all ranges that are consecutive
+    /// Merges all ranges that are consecutive.
     pub fn merge_consecutive_ranges(mut self) -> Self {
         let mut primary = self.ranges[self.primary_index];
 
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -523,13 +523,10 @@ impl Selection {
     }
 
     // Replaces ranges with one spanning from leftmost to rightmost selection
-    pub fn merge_ranges(mut self) -> Self {
-        if !self.ranges.is_empty() {
-            let merged_range = self.ranges.iter().fold(self.ranges[0], |a, b| a.merge(*b));
-            self.ranges = smallvec![merged_range];
-            self.primary_index = 0;
-        }
-        self
+    pub fn merge_ranges(self) -> Self {
+        let first = self.ranges.first().unwrap();
+        let last = self.ranges.last().unwrap();
+        Selection::new(smallvec![first.merge(*last)], 0)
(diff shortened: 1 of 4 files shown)


helix-core/src/selection.rs
```diff
@@ -522,6 +522,16 @@ impl Selection {
...
 
+    // Replaces ranges with one spanning from leftmost to rightmost selection
+    pub fn merge_ranges(mut self) -> Self {
+        if !self.ranges.is_empty() {
+            let merged_range = self.ranges.iter().fold(self.ranges[0], |a, b| a.merge(*b));
+            self.ranges = smallvec![merged_range];
+            self.primary_index = 0;
+        }
+        self
+    }
```
the-mikedavis, line 533: This can be simplified a little:

```suggestion
    pub fn merge_ranges(self) -> Self {
        // Selections must have non-empty ranges.
        let first = self.ranges.first().unwrap();
        let last = self.ranges.last().unwrap();
        Self::single(first.from(), last.to())
    }
```

hamrik: Looks good! Didn't even think to make it functional, just blindly followed the existing merge implmentation. My bad.

This change assumes there is always at least one selection and that ranges are always sorted.
I just noticed the invariant in the comment above `Selection`, but it says nothing about the first range being the leftmost. Does that always hold? I see a `sort` in `normalize`, maybe we should use `min` and `max` here?

One other regression I see is that this change no longer keeps the range orientation (see `extend` vs `merge`). Though we could just flip the head and anchor if the first range is flipped.

hamrik:
```suggestion
    pub fn merge_ranges(self) -> Self {
        let (left, right) = self
            .ranges
            .iter()
            .fold((usize::MAX, usize::MIN), |(l, r), &i| {
                (std::cmp::min(l, i.from()), std::cmp::max(r, i.to()))
            });

        match self.ranges.first().unwrap().direction() {
            Direction::Forward => Self::single(left, right),
            Direction::Backward => Self::single(right, left),
        }
    }
```

How about this? It works on any ordering, respects flipped ranges and it's functional. it still does the same number of comparisons as the original but I don't know if we can do better there, as they both visit every range head/anchor exactly once.

the-mikedavis: > This change assumes there is always at least one selection and that ranges are always sorted.

These assumptions should always hold. Ranges can't use `usize::MAX`: they should point to a valid character index in the document. That `fold` should never end up using `MAX` because `self.ranges` should always be non-empty, but it's better to assert that ranges is non-empty with the `unwrap` than to default to using MAX since that could lead to a more confusing behavior or panic.

hamrik:
```suggestion
    pub fn merge_ranges(self) -> Self {
        let first_range = self.ranges.first().unwrap();
        let (left, right) = self
            .ranges
            .iter()
            .fold((first_range.from(), first_range.to()), |(l, r), &i| {
                (std::cmp::min(l, i.from()), std::cmp::max(r, i.to()))
            });
            
        match first_range.direction() {
            Direction::Forward => Self::single(left, right),
            Direction::Backward => Self::single(right, left),
        }
    }
```

This version uses the first range, the existence of which is asserted by `unwrap`, for the initial values to compare against in the `fold`.

the-mikedavis: Iterating over the ranges isn't necessary: you can just pull out the first and last range with `.first()` and `.last()`.

hamrik: Combed through every transformation function of `Selection` and it seems most of them call `normalize`, which sorts the ranges in place. Still not 100% convinced that the first selection will _always_ be the first, but so far I couldn't produce a bug with the following:

```suggestion
    pub fn merge_ranges(self) -> Self {
        let left = self.ranges.first().unwrap();
        let right = self.ranges.last().unwrap();
        Selection::new(smallvec![left.merge(*right)], 0)
    }
```

the-mikedavis: That looks good to me 👍

the-mikedavis (approved):
The `A-minus` keybinding isn't taken in Kakoune AFAIK so I think this looks good 👍

pascalkuthe (approved):
I found myself missing something like this a few times. Thanks for working on this 👍

PR #7059, opened 2023-05-17
Update languages.toml for Nickel

vkleen:
Update the revision of `tree-sitter-nickel` and copy over the highlighting queries. Also, adjust the `auto-pairs` configuration, since `'` isn't a paired delimiter in Nickel.

### DIFF
--- a/runtime/queries/nickel/highlights.scm
+++ b/runtime/queries/nickel/highlights.scm
@@ -20,7 +20,7 @@
 )
 
 (record_operand (atom (ident) @variable))
-(let_expr
+(let_in_block
   "let" @keyword
   "rec"? @keyword
   pat: (pattern
@@ -53,7 +53,7 @@
 (interpolation_end) @punctuation.bracket
 
 ["forall" "default" "doc"] @keyword
-["if" "then" "else" "switch"] @keyword.control.conditional
+["if" "then" "else" "match"] @keyword.control.conditional
 "import" @keyword.control.import
 
 (infix_expr
--- a/languages.toml
+++ b/languages.toml
@@ -577,7 +577,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "nickel"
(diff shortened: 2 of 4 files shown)


ErinvanderVeen (dismissed):

runtime/queries/nickel/highlights.scm
the-mikedavis, line 1: Are these copied over from tree-sitter-nickel? They look like they're used in nvim-treesitter: you can remove the `@spell` captures and some of the other captures need to be renamed to align with the capture names that [Helix uses](https://docs.helix-editor.com/master/themes.html#syntax-highlighting) (for examples, `conditional` should become `keyword.control.conditional`, `boolean` should become `constant.builtin.boolean`, etc.)

the-mikedavis (changes requested):

runtime/queries/nickel/highlights.scm
vkleen: These were indeed copied over from tree-sitter-nickel. I've restored the original ones from helix and tried to update what's necessary for the changes in tree-sitter-nickel.

the-mikedavis (dismissed):
Thanks!

pascalkuthe:
Seems like CI is failing with some identity query related error. Seems like those also need to be updated before we can merge

the-mikedavis:
Ah yeah it looks like the `switch_expr` node doesn't exist in this version (https://github.com/helix-editor/helix/actions/runs/5023053043/jobs/9020326641?pr=7059). It looks like it may be `match_expr` now https://github.com/nickel-lang/tree-sitter-nickel/blob/3a794388773f2424a97b2186828aa3fac4c66ce6/grammar.js#L145-L150

vkleen:
I'll take care of that later today, thanks for diagnosing!

pascalkuthe (approved):

the-mikedavis (approved):

PR #7064, opened 2023-05-18
book: Install helix declaratively from nixpkgs on NixOS

mweinelt:
This is really the most simple way, and helix is well maintained in nixpkgs.

The flake solution is overly complicated for newcomers to NixOS, as I just
witnessed on our support room. Copying/adapting/understanding that
expression is far easier.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -82,7 +82,10 @@ in the AUR, which builds the master branch.
 
 ### NixOS
 
-Helix is available as a [flake](https://nixos.wiki/wiki/Flakes) in the project
+Helix is available in [nixpkgs](https://github.com/nixos/nixpkgs) through the `helix` attribute,
+the unstable channel usually carries the latest release.
+
+Helix is also available as a [flake](https://nixos.wiki/wiki/Flakes) in the project
 root. Use `nix develop` to spin up a reproducible development shell. Outputs are
 cached for each push to master using [Cachix](https://www.cachix.org/). The
 flake is configured to automatically make use of this cache assuming the user


pascalkuthe:
The problem in the chat you saw was actually related to the fact that they installed the nix package exactly as you described here and endedup with an ancient helix version (22.8) which caused issues. Since the flake is officially maintained I think it should definitly stay the recommended installation method. 

We could mentinom this as an alternative but warn that versions could be outdated

mweinelt:
The problem I saw in the chat, was that they thought it was either flakes or the highway. Flakes is considered experimental by upstream, but people are still jumping on the bandwagon left and right.

If it helps, I can also provide instructions on how to install helix from nixos-unstable.

the-mikedavis:
Mentioning that Helix is available as `helix` in [nixpkgs](https://github.com/NixOS/nixpkgs) would be a good improvement as well as mentioning the unstable channel but I don't think we should be giving snippets for NixOS config. There are so many ways to install packages out of nixpkgs:

* the [NixOS way](https://nixos.wiki/wiki/Overview_of_the_NixOS_Linux_distribution) using  `environment.systemPackages` in `/etc/nixos/configuration.nix`
* the other NixOS way, through some `extraPackages` option for a window manager for example
* [home-manager](https://github.com/nix-community/home-manager) in the NixOS config file
* home-manager as an input to a system configuration flake
* `nix-env`
* `nix profile`
* `nix shell`/`nix run`

(and probably more I don't know about.) We mention the flake prominently because we maintain it in-repo as Pascal says, and it's easier to follow the latest changes and to create a custom build for yourself.

I think it's certainly out of scope for Helix docs to be teaching NixOS. Mentioning nixpkgs and the unstable channel would be good though, and we could link out to the NixOS documentation for things like channels (https://nixos.wiki/wiki/Nix_channels) for further reading

archseer:
There is no preferred way to install helix on NixOS though, we simply support all the options. Flakes are the most popular (to keep in sync with master, or nix unstable to get latest release), but flakes are still marked as experimental and some users are still sticking to their old nix-env setups because of this. It's not up to us to teach users how to use their distros -- it's up to users to learn and make an educated choice.

archseer:
e.g. I'd be fine with this change if it dropped any command sequences and instead said "Helix is available in nixpkgs, unstable usually carrying the latest release. Helix is also available as a flake in the project root. "

CobaltCause:
Why does this have to be Helix's responsibility rather than Nix/NixOS/nixpkgs? If you feel uncomfortable editing your Nix things, perhaps you should take the time to learn those tools first, so that you don't need to rely on each individual application you are interested in providing a copy-pastable code snippet for every possible installation method.

Since there are so many ways to make a program available on a system running Nix/NixOS, I think something vague like what was suggested [here](https://github.com/helix-editor/helix/pull/7064#issuecomment-1553932100) is optimal: it states the places the derivation can be sourced from and leaves the user to decide exactly how they want to consume it.

> Telling people "Go read the wiki, educate yourself until you know how to install it by yourself, then come back and install" is not a good idea if you want people to use your app, if you do that many people will say "ok let's use neovim then" and simply leave

I don't think this is a valid argument because, at least hypothetically, you wouldn't know how to do that anyway. It's a completely empty threat.

CobaltCause:
> > Why does this have to be Helix's responsibility rather than Nix/NixOS/nixpkgs?
>
> It simply does not have to be, but it can

It shouldn't be.

> the lack of good docs of nixos/tutorials

Right, this is the crux of the issue.

The way to alleviate this is to improve the Nix/NixOS/nixpkgs documentation. Trying to "compensate" for this situation in your suggested way will continue to result in this exact same problem, except with a different program instead of Helix.

Making direct improvements to Nix/NixOS/nixpkgs documentation will make it easier for knowledgeable parties to keep track of and maintain, and easier for new users to find information and learn from it.

Scattering this sort of information around like this PR currently does is actively detrimental to the class of problem you (and others) are having.

(Nix/NixOS could really use something like [this section](https://wiki.archlinux.org/title/Help:Reading#Installation_of_packages) from the Arch Wiki...)

###### Miscellaneous points

> if we want helix to be used more

The premise of this is invalid because you could apply it to any program. You should want every program's Nix/NixOS installation methodology to be equally well-documented so that this isn't something to compete on in the first place.

> you can find how to install neovim on nixos in its wiki, they simply use nix-env

If you're on NixOS, this is not the right action to take.

> [neovim] is officially mantained unlike helix

This is not true; they are both packaged in nixpkgs.

CobaltCause (approved):

the-mikedavis (approved):

pascalkuthe (approved):

the-mikedavis:
Thanks y'all!

PR #7078, opened 2023-05-19
nix: add overlay for convenient package usage

### LINKED ISSUE #7042: [nix]: add an overlay
@yusdacra when you have a moment, could you add an overlay for helix ?
Since helix is integrated with home-manager, if you want to override helix to use the version for the flake (in case you want to use helix from github master for example), then you have to pass the version all the way to the home manager package.
Having an overlay would make that easier.
Happy to discuss if you think that doesn't make sense.
Thanks for the work on the flake!

happysalada:
This just adds an overlay to the nix flake to make it convenient to use helix.

It's not a big change. I'm adding it since a feature I really want to try was merged yesterday (multiple lsps).

@yusdacra @archseer when you have a moment.

closes #7042

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,7 +64,7 @@
       };
   in
     inp.parts.lib.mkFlake {inputs = inp;} {
-      imports = [inp.nci.flakeModule];
+      imports = [inp.nci.flakeModule inp.parts.flakeModules.easyOverlay];
       systems = [
         "x86_64-linux"
         "x86_64-darwin"
@@ -145,6 +145,10 @@
         packages.helix = makeOverridableHelix config.packages.helix-unwrapped {};
         packages.helix-dev = makeOverridableHelix config.packages.helix-unwrapped-dev {};
         packages.default = config.packages.helix;
+
+        overlayAttrs = {
+          inherit (config.packages) helix;
+        };
 
         devShells.default = config.nci.outputs."helix-project".devShell.overrideAttrs (old: {
           nativeBuildInputs =


yusdacra (approved):
LGTM

the-mikedavis (approved):

MDNLss:
I don't know if this is the right place to put this but, 
What would a sample helix overlay look like using the easyOverlay?

happysalada:
if you are using the helix flake, in your configuration, you can add
```
      nixpkgs = {
        overlays = [
          helix.overlays.default
        ];
      };
```
(this nixpkgs option is available for both nixos and nix-darwin).

(helix comes from your flake inputs
```
    helix.url = "github:helix-editor/helix";
```

That would automatically give you the latest master as your pkgs.helix
hope this helps!

C2Qck:
> 
This helps but what I meant by my question was, What attribute and sets from Helix package would this overlay allow someone to override within an overlay?

happysalada:
This overlay was made to use the helix version on master without having to pass the flake attribute everywhere.

If you want to override attributes, you most likely wont need this overlay. You have to look at the expression in nixpkgs and see which attributes you can override from there.

Im not sure what you want to override in particular, but you might be better off using the flake in the repo if you need a particular build.

C2Qck:
> This overlay was made to use the helix version on master without having to pass the flake attribute everywhere.
> 
> If you want to override attributes, you most likely wont need this overlay. You have to look at the expression in nixpkgs and see which attributes you can override from there.
> 
> Im not sure what you want to override in particular, but you might be better off using the flake in the repo if you need a particular build.

I was planning on overriding in order to add a language setup for several languages configurations into the package and themes similar to the home-manager option. But I'll take a look at the nixpkgs defintion and see what can be done.

happysalada:
You dont need to override for that. You can define your own language.toml file and it will be merged with the default one.
There is an issue where people talk about tailwindcss if you want a reference.

PR #7080, opened 2023-05-19
language-server: add tailwindcss language-server

happysalada:
Just adding the tailwind css language server for convenience.
(now that multiple language servers are enabled, this is going to be great for usage with svelte files for example (or any of the other js frameworks).

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -63,6 +63,7 @@ sourcekit-lsp = { command = "sourcekit-lsp" }
 svelteserver = { command = "svelteserver", args = ["--stdio"] }
 svlangserver = { command = "svlangserver", args = [] }
 swipl = { command = "swipl", args = [ "-g", "use_module(library(lsp_server))", "-g", "lsp_server:main", "-t", "halt", "--", "stdio" ] }
+tailwindcss = { command = "tailwindcss-language-server", args = ["--stdio"], config = { } }
 taplo = { command = "taplo", args = ["lsp", "stdio"] }
 terraform-ls = { command = "terraform-ls", args = ["serve"] }
 texlab = { command = "texlab" }


pascalkuthe:
So all this provides is the definition for the LS but it doesn't actually enable the server by default for any language. I understand that tailwind is just one specific CSS framework (with its own LS?) so I guess that idea is that users then manually enable this server in their (per project) languages.toml. 

I am not sure if that passes the bar for inclusion in core since users would only need to paste this exact line into their per project languages.toml (which they need to modify anyway to enable the LS)

nipuna-perera:
I think that would be helpful to document though. Maybe in an example of
using multiple language servers together?

happysalada:
Here is one of the reason I added it.
Some of use configure the customizations of helix with nix. Interoperability with toml is not amazing, and so when trying to enable multiple language servers through nix, the toml file generated failed. Ultimately I don't think it's a problem that helix should solve, but it would make usage much easier.
The other side of it, is that it seemed to me that the cost wasn't so great, just adding one line in a toml file. I would understand if it was adding a whole tree-sitter grammar or similar that it could be too much.

The other reason would be that tailwind is one of the only sane ways to do css, and so having the editor make it easier to support that would be helping people :-) Jokes aside, I think that what tailwind brings on the table is significant enough to make it a nice experience to use it.

A couple of questions regarding what could go into core and what could not.
I was considering adding surrealql and nickel 
https://github.com/tweag/nickel
https://github.com/surrealdb/surrealdb
Do you think those would make it into core ?

Ultimately I'm fine either way, just let me know how you feel about things.
And thanks again for contributing to this amazing editor! ❤️

happysalada:
@nipuna-perera where do you think it would be best to document ?

the-mikedavis:
Typically LSP examples and docs go in the LSP installation wiki https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

happysalada:
Thanks for the place for the docs!
I'm trying to make the lsp work properly, I'll update the PR once I have used it successfully for a bit.

archseer:
Replaced by #8442

PR #7085, opened 2023-05-19
Change statusline color according to mode if color_modes is true

veranorooz:
The seperator starts to look weird with color modes. This is setting separator `fg` same as the mode, and `bg` same as `ui.statusline` ignoring separator styling.

Before:
image

After:
image
image


CptPotato:
FYI there's some overlap with #5022 and #5465.

Also, I think you should be able to use the separator element anywhere on the statusline. With this change it will only look good when used next to the mode element.

Issue #6382, opened 2023-03-20
Panic when moving between monitors


dr4gut:
### Summary

I had Helix open on a vertical monitor. I pulled the terminal emulator to a larger, horizontal monitor which caused helix to crash.

`thread 'main' panicked at 'index out of bounds: the len is 3180 but the index is 5532', helix-term/src/ui/menu.rs:354:29`

### Reproduction Steps

 

I cannot reproduce the issue

I tried this:

1. Run helix on a vertical monitor and move the window to a horizontal one

I expected this to happen: Helix resizes along with the emulator

Instead, this happened: Helix panicked

### Helix log

No relevant log entry. 

### Platform

Windows (WSL2)

### Terminal Emulator

Windows Terminal

### Helix Version

22.12-382-g34be71f

Lykos153:
This just happened to me without multiple monitors. Just one single laptop monitor, resized Helix from half screen to full screen with a tiling window manager (xmonad) and got:

```
thread 'main' panicked at 'index out of bounds: the len is 2159 but the index is 3084', helix-term/src/ui/menu.rs:367:29
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

Version: helix 23.10 (f6021dd0)

pascalkuthe:
https://github.com/helix-editor/helix/pull/7089#issuecomment-1564990166

PR #7089, opened 2023-05-20
Prevent crash while moving between different sized monitors

veranorooz:
This is preventing the crash described in https://github.com/helix-editor/helix/issues/6382. It doesn't actually fix getting unexpected values. Because I could not reproduce the error consistently, I could not identify the root cause. Instead, this is making sure Helix doesn't panic when overflow happens.


pascalkuthe:
It looks like the reason for the crash is that the scroll value changes when the screen is resized potentially ending up larger than cursor. This fix just prevents the crash but wouldn't fix the weirs visual artifacts associated with that. I would check where the scroll value is updated and maybe adding some sanity vhsvks regarding the cursor there (basically either updating the cursor or the scroll value so the cursor I'd always in bounds)

PR #7090, opened 2023-05-21
term: Add % expansion for relative path to current file

valpackett:
This is especially useful for creating a file next to the current one, and is inspired by vim's expansions (but only handles the basic case of course).

hx cast

I've been missing this from nvim (kakoune too iirc??) ever since I switched, finally got frustrated enough to just implement it :)

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -445,7 +445,7 @@ pub mod completers {
     }
 
     // TODO: we could return an iter/lazy thing so it can fetch as many as it needs.
-    fn filename_impl<F>(_editor: &Editor, input: &str, filter_fn: F) -> Vec<Completion>
+    fn filename_impl<F>(editor: &Editor, input: &str, filter_fn: F) -> Vec<Completion>
     where
         F: Fn(&ignore::DirEntry) -> FileMatch,
     {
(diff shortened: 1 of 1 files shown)


kirawi:
See https://github.com/helix-editor/helix/pull/6979

valpackett:
> See #6979

hm. I'm not sure these should be related. Maybe `%` *could* also be implemented as `%val{filename}` from that PR, but

- that right now seems to expand as the absolute path, and I assume that is very much desired in various places, so there would have to be some `%val{relfilename}` or something
- here I'm changing the *completion* function for files, which results in intuitive and friendly behavior: note how just typing `%` already shows the expanded path in the tab completion list, just like `~` on its own for absolute home dir path. I'm not sure the behavior would be the same with that implementation, as I see that PR touches the general prompt text input code…

philipgiuliani:
In this PR the % char will be added as a register which allows to paste the current filename quickly: https://github.com/helix-editor/helix/pull/6985

pascalkuthe (changes requested):
I dont think we need yet another way to expand to.the current file. Instead we could add `%val{filename}` (and all the other possible placeholders) to the list of completions. This automatically teaches users about the syntax without adding anything new. You couls rebase this PR on https://github.com/helix-editor/helix/pull/6979 and implement that

valpackett:
> add %val{filename} (and all the other possible placeholders) to the list of completions

How would this be done and how would it work? If the *placeholders themselves* would appear in completions, that would be useless for me.

My desired workflow is **exactly as in the screencast**, that is I need to be able to type `:o %<Tab><Ctrl-W>new.file.name` to open a file in the same directory as the current file. I'm not interested in even being aware of placeholders for this use case. One single simple character should expand **to the whole path relative to the current directory** so that I can **edit the path inline**.

kirawi:
This PR is linked to https://github.com/helix-editor/helix/issues/2985

valpackett:
It's literally not a duplicate @kirawi, it's a different approach for a different place.

I guess using the register isn't the worst considering that there apparently *is* [a keybinding in prompts](https://docs.helix-editor.com/keymap.html#prompt) to paste from a register, but right now that `DocumentPathRegister` from #6985 does not make the path relative to the current directory. And… should it? When pasting into the text the absolute path might actually be wanted more often. But in the prompt, the path must be relative to the current directory.

But also I really want to specifically exactly press `%<Tab>` and not `<Ctrl-R>%` 😠

PR #7098, opened 2023-05-21
Add a mappable version of existing :reflow command

jw013:
https://github.com/helix-editor/helix/pull/2128 added the `reflow` command but I found I was not able to bind any keys to it because it was only added in the command (`:`) mode. This change adds a mappable command to call the existing reflow function. When used as a keymap, the count is used as the text width, with a missing or zero count triggering the existing original fallback default logic. This change does not set a default keybinding for the new reflow command to leave that decision up for discussion. I use `Alt-q` myself.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -389,6 +389,7 @@ impl MappableCommand {
         paste_primary_clipboard_before, "Paste primary clipboard before selections",
         indent, "Indent selection",
         unindent, "Unindent selection",
+        reflow, "Hard-wrap selection to given width",
         format_selections, "Format selection",
         join_selections, "Join lines inside selection",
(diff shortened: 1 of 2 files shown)


robinvd:
you should be able to bind it using this syntax
```
[keys.normal]
q = ":reflow"
```

jw013:
Thanks, not sure how I missed that. There is one benefit of a dedicated normal mode command which is that a text-width argument can be specified using the count. So e.g. `99q` would run the equivalent of `:reflow 99` and `80q` would do `:reflow 80`, etc.

the-mikedavis:
With https://github.com/helix-editor/helix/pull/6979 and some extra work we could add a `%val{count}` expansion that would cover that use-case. I don't want to have to add non-typable versions of any typable commands to work around functionality issues with keybinding

PR #7100, opened 2023-05-21
Create new command to open docs

lucascherzer:
# what
This PR introduces a new command (`:docs-open`) which opens the helix documentation at https://docs.helix-editor.com/ in the default web browser
# why
I found it convenient when writing the config file and looking up some commands (still learning the keymappings).
# notes
The implementation is straightforward but introduces a [new dependency](https://docs.rs/open/4.1.0/open/index.html).
I found that the command only works if helix is run on the native OS, and WSL does not work.
The error message it displays under that condition is verbose and could be shortened for simplicity's sake - suggestions are welcome.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1201,6 +1201,7 @@ dependencies = [
  "libc",
  "log",
  "once_cell",
+ "open",
  "pulldown-cmark",
  "serde",
  "serde_json",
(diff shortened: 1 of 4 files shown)


gabydd:
See also https://github.com/helix-editor/helix/pull/5820q

gabydd:
If this is wanted in core something that should be considered is that if someone is running from master this should open the master docs (https://docs.helix-editor.com/master) because usually there are some breaking config changes between a stable release and master.

lucascherzer:
> If this is wanted in core something that should be considered is that if someone is running from master this should open the master docs (https://docs.helix-editor.com/master) because usually there are some breaking config changes between a stable release and master. In my opinion I think most things included in the docs should be included within help docs at some point which would mean they are always synced with whatever version of helix you are using and are available even if you don't have internet. (Also a plus you would be able to use helix keybindings to navigate through them)

I agree. Having the docs come packaged with helix can be a benefit in this regard

pascalkuthe:
Closing this one out as stale. I don't see us adding this to core. Instead I hope that we have a builtin documentation (a `:help` command like in vim)

PR #7111, opened 2023-05-22
Update diagnostics correctly on LSP exit

poliorcetics:
LSPs can produce diagnostics for unopened documents, which previously made Helix increase the
workspace diagnostic count increase on each LSP restart.

There was also a bug in that if a document had diagnostics from two LSPs and one was exited, all
the diagnostics for all LSPs were cleared in `editor.diagnostics`, this has now been fixed.

Initially I was troubleshooting the ever increasing workspace count, that's why my first commit is a
small fix to the workspace diagnostics rendering code

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -964,6 +964,15 @@ impl Application {
                     Notification::Exit => {
                         self.editor.set_status("Language server exited");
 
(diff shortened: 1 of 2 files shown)


archseer (approved):

pascalkuthe (approved):
good catch, thanks!

Issue #7113, opened 2023-05-22
Have an option for picker wrapping


antoyo:
Hi.
I know that the picker started to wrap in [this commit](https://github.com/helix-editor/helix/commit/c7d6e4461f249108189cddf44b487b4c71c5520a).

I find this unconvenient, so it would be nice to have an option to disable this.

Thanks.

PR #7124, opened 2023-05-23
Add option to disable picker list wrap

ksdrar:
Closes https://github.com/helix-editor/helix/issues/7113


pascalkuthe:
we don't like adding config options for small design choices like this. The current behaviour is intentional. This could oneday be solved by component keymaps + plugins but I don't see us adding this to core.

This branch has aslgo gone quite stale so I am closing this out. Thanks for contributing!

PR #7128, opened 2023-05-24
add --readonly flag to enter readonly mode

ds-cbo:
This is a very rough approach, but I was missing something like vim's `-R` where you can open a file without accidentally writing to it

It can probably be adapted to recognise no-write files (r-- or r-x) as well, as suggested in https://github.com/helix-editor/helix/issues/2627 

And it can probably also be adapted to enable readonly for specific buffers instead of editor-wide, as suggested in https://github.com/helix-editor/helix/pull/3709


mcdoker18:
But it's not a readonly mode because I can still modify documents. Maybe would be better to forbid switching to insert mode or show error if user trying to?

LSP also can modify documents.

And there is  the auto save option in the editor config.

the-mikedavis:
I definitely want a concept of read-only in Helix but it should be more extensive, and I think the design of the feature needs some discussion (i.e. an [idea discussion](https://github.com/helix-editor/helix/discussions/categories/ideas)). I'm hesitant about an editor-wide read-only mode and I'm fairly certain a Document should not be locked in a read-only state: it should be possible to write a read-only Document to a path which isn't read-only.

Kakoune has a nice read-only feature including refusal to enter insert-mode that could be inspiration for this in Helix.

ds-cbo:
@3f6a On my part the current status is that I occasionally rebase this branch and use that (plus the patch from #8362) for my custom build. I'm happy with this minimal patch as-is and am not planning to lead the idea discussion or extend it accordingly. I'm keeping this patch relatively up-to-date for my own workflows, and I'm sharing it with the community in case other people would like the same, but I don't expect this PR to be merged any time soon.

pascalkuthe:
Closing this one out as this feature needs design dicussion and major changes and will not be merged in this form (see @the-mikedavis comment above). Thank you for contributing

heaths:
I started https://github.com/helix-editor/helix/discussions/11586 to discuss. I'm interest in seeing this and even working on it.

PR #7132, opened 2023-05-24
Fix off-by-one in select symbol references

mtoohey31:
This merge request fixes an off-by-one error in `select_references_to_symbol_under_cursor`.

When finding the range which contains the head of the current primary selection (so that this range will be the new primary selection), using the `head` field of the primary selection leads to the correct range being missed when `head > anchor` and the cursor is on the final character of the range, since `Range::contains`'s `pos` parameter seems to refer to the position of a character since it is non-inclusive. Using the `cursor` method fixes this since it returns the position of the character under the cursor.


dead10ck (approved):
Looks good to me, thanks for the fix!

PR #7147, opened 2023-05-26
chore: update scala tree-sitter grammar

zetashift:
Update to a new Scala tree-sitter grammar revision, because the current one lags behind by quite a bit:

https://github.com/tree-sitter/tree-sitter-scala/commits/master

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1229,7 +1229,7 @@ language-servers = [ "metals" ]
 
 [[grammar]]
 name = "scala"
-source = { git = "https://github.com/tree-sitter/tree-sitter-scala", rev = "f6bbf35de41653b409ca9a3537a154f2b095ef64" }
+source = { git = "https://github.com/tree-sitter/tree-sitter-scala", rev = "23d21310fe4ab4b3273e7a6810e781224a3e7fe1" }
 
 [[language]]
 name = "dockerfile"
--- a/runtime/queries/scala/locals.scm
+++ b/runtime/queries/scala/locals.scm
@@ -1,0 +1,29 @@
+(template_body) @local.scope
+(lambda_expression) @local.scope
+
+
+(function_declaration
+      name: (identifier) @local.definition) @local.scope
+
+(function_definition
+      name: (identifier) @local.definition)
+
+(parameter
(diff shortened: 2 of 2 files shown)


archseer:
Is there also any highlight query changes? It's usually not enough to just bump the grammar

the-mikedavis:
It's a little hard to tell looking at the diff ([compare](https://github.com/tree-sitter/tree-sitter-scala/compare/f6bbf35de41653b409ca9a3537a154f2b095ef64...23d21310fe4ab4b3273e7a6810e781224a3e7fe1)) but it looks like all of the changes are structural or to unnamed rules so we may not need to update the highlights queries.

We could add the new locals queries though for better highlighting: https://github.com/tree-sitter/tree-sitter-scala/commits/e37d21701bd03d5bb27f5e2e8339ce6afc419784

zetashift:
> We could add the new locals queries though for better highlighting: https://github.com/tree-sitter/tree-sitter-scala/commits/e37d21701bd03d5bb27f5e2e8339ce6afc419784

What are local queries (sorry not a tree-sitter expert).

the-mikedavis:
Ah sorry, I mixed up the permalink. I think we should be able to use these out-of-the-box https://github.com/tree-sitter/tree-sitter-scala/blob/798ec789401de159e0c07ae170c8ae6dfbbfd94c/queries/scala/locals.scm

copying them to runtime/queries/scala/locals.scm. (Also see https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables)

zetashift:
> Ah sorry, I mixed up the permalink. I think we should be able to use these out-of-the-box https://github.com/tree-sitter/tree-sitter-scala/blob/798ec789401de159e0c07ae170c8ae6dfbbfd94c/queries/scala/locals.scm
> 
> copying them to runtime/queries/scala/locals.scm. (Also see https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables)

Oooh, I see, thank you!

the-mikedavis (approved):

pascalkuthe (approved):

Issue #7149, opened 2023-05-26
Config is not passed to the LSP with the new multi LSP support


Marsevil:
I'm trying to make the new multi LSP support works with `ltex-ls` and `texlab`.
My problem is that `ltex-ls` do not use the `en-GB` and do not use my custom dictionary. I guess that the config is not passed to the language server.

Here is my config file.

```toml
[[language]]
name = "latex"
language-servers = [ "ltex-ls-custom" ,"texlab" ]
file-types = ["tex"]
scope = "source.tex"
roots = []

[language-server.ltex-ls-custom]
command = "ltex-ls"
args = ["--log-file=/tmp"]
config = { language = "en-GB", dictionary = { "en-GB" = ["somecustomwords"] } }

[language-server.texlab]
command = "texlab"
```

Did I miss something or is it a bug ??

> **Note** : On the last release `ltex-ls` works as expected.
> I renamed the LSP `ls-ltex-custom` because I already have a default configuration and I would like to be sure that is not a conflict.

gabydd:
Can you try putting all the language-server stuff above ``[[language]]`` that seems to have worked for other people

Marsevil:
I'd already tried but, in doubt, I double checked and it does not work

gabydd:
I got ltex to work with the configuration below, there are two important parts the first is all the regular tables ``[]`` have to go before the arrays of tables ``[[]]`` or else the language-servers wont get defined, The second part is ltex-ls takes the config as ``{ltex = {<actual settings>}`` so that needs to be specified for it use your settings. Checking the logs with: ``:log-open`` is very helpful for debugging something like this if you have issues in the future.

```toml
[language-server.ltex-ls-custom]
command = "ltex-ls"
args = ["--log-file=/tmp"]

[language-server.ltex-ls-custom.config.ltex]
language = "en-GB"
dictionary = { "en-GB" = ["somecustomwords"] } 

[language-server.texlab]
command = "texlab"

[[language]]
name = "latex"
language-servers = [ "ltex-ls-custom" ,"texlab" ]
file-types = ["tex"]
scope = "source.tex"
roots = []
```

Marsevil:
Ok now I feel stupid :sob: 
Thank you very much !!

gabydd:
no worries, we should definitely specify that limitation of toml in the docs cause quite a few people have had this problem

PR #7153, opened 2023-05-27
Docs: specify that language-server table must be before language and grammar array

gabydd:
Due to a limitation in toml, when specifying the new config  for language servers from #2507 because they are just tables they must be put before the arrays of tables (language and grammar). I have seen a couple issues and matrix questions pop up where people didn't know this and put the language server configuration after the language config and then nothing happens, #7149 and https://github.com/helix-editor/helix/discussions/7101#discussioncomment-5962068. If we think there is going to be more arrays of tables in the future we could make the message more general but I think this way is more clear. I couldn't find any official documentation about this behaviour of toml but I am pretty sure it is the case.

### DIFF
--- a/book/src/languages.md
+++ b/book/src/languages.md
@@ -99,8 +99,8 @@ with the following priorities:
 
 Language servers are configured separately in the table `language-server` in the same file as the languages `languages.toml`
 
-> 💡 Due to limitations of toml anything in the ``language-server`` table must
-> be specified before the ``[[language]]`` and ``[[grammar]]`` arrays of tables.
+> 💡 Due to limitations of TOML, anything in the `language-server` table must
+> be specified before the `[[language]]` and `[[grammar]]` arrays of tables.
 
 For example:
 
--- a/book/src/languages.md
+++ b/book/src/languages.md
@@ -99,6 +99,9 @@ with the following priorities:
 
 Language servers are configured separately in the table `language-server` in the same file as the languages `languages.toml`
 
+> 💡 Due to limitations of toml anything in the ``language-server`` table must
+> be specified before the ``[[language]]`` and ``[[grammar]]`` arrays of tables.
+
 For example:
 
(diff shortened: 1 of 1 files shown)


gabydd:
Similarly people have started updating https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers with the new language server config but specificying it below the language table which should probably be fixed as well. It's late here so I will do that in the morning though.

book/src/languages.md
```diff
@@ -99,6 +99,9 @@ with the following priorities:
 
 Language servers are configured separately in the table `language-server` in the same file as the languages `languages.toml`
 
+> 💡 Due to limitations of toml anything in the ``language-server`` table must
+> be specified before the ``[[language]]`` and ``[[grammar]]`` arrays of tables.
```
the-mikedavis, line 103:
```suggestion
> 💡 Due to limitations of TOML, anything in the `language-server` table must
> be specified before the `[[language]]` and `[[grammar]]` arrays of tables.
```

I think the double backticks are not necessary in markdown even in blockquotes

gabydd:
this might not be the best place to ask this but I wonder if we should put a note on the installing language servers wiki page that we are moving to a new configuration format for language servers and then start changing the wiki to the new format. Or wait till the next release to move the wiki over to the new format. We could still have a note but just saying something like: "The configuration on this page is only for the current release here is how to use it with the current master."

gabydd:
been looking at this again and it seems like you don't actually need to put the language-server table above any arrays of tables, I am guessing something else was happening in those issues and I incorrectly assumed that the moving the ``[[language]]`` table fixed it. I am going to close this pr unless someone can reproduce this problem.

gabydd:
ah this was a problem when adding the new language server when running. then reloading config, and then restarting the server. If it seems possible to fix I'll make an issue for it.

PR #7159, opened 2023-05-28
fixed repeat_last_motion extends selection

### LINKED ISSUE #7033: repeat_last_motion extends selection?
If the cursor is at the beginning of a sentence that has several commas, if I do 'f,' the selection gets extended to the first comma. If I manually repeat the command, the selection now consists of the first comma up to and including the second. 

But if instead of manually repeating the command, I use 'alt-.' (repeat_last_motion), the behavior is different -- the selection is *extended* to the second comma. I find this behavior odd because I would expect repeat_last_motion to produce the same behavior that I would get by manually repeating the last motion.

I understand that helix, while influenced by kakoune, is not a clone. But I note that kakoune does what I would expect in the situation discussed above.

roberte777:
Closes #7033


pascalkuthe (approved):
thanks!

PR #7167, opened 2023-05-29
Minor touchups in keymap.rs

gibbz00:
Preparing to take another shot at https://github.com/helix-editor/helix/pull/5635. 

Depends on #7193

PR #7168, opened 2023-05-29
Idiomatic keymap infobox body ordering.

gibbz00:
Depends on: #7167. Motivation is in the commit message.


gibbz00:
Merged with: #7214

PR #7169, opened 2023-05-29
Exclude config no_op bindings in command palette.

gibbz00:
My the top of my command palette currently looks like this 🙈

2023-05-29-222146_screenshot

Depends on #7193


gibbz00:
Merged with: #7214

PR #7178, opened 2023-05-30
Add check for a non-zero value for tab width

### LINKED ISSUE #7094: Add not zero check on tab-width.
If set tab-width to zero, helix will crash (divided by zero). The error might occured from the helix-core/grapheme.rs.
Although it's a bit weird to set tab-width to zero (or even negative numbers), presenting the tab-width as 1 and alerting instead of panic would be better, IMO.

bo1led-owl:
Closes #7094. I think it is weird to set tab width to zero, so I added a custom deserialization function (a small wrapper that checks if the tab width value is zero).

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -48,6 +48,21 @@ where
         .transpose()
 }
 
+fn deserialize_tab_width<'de, D>(deserializer: D) -> Result<usize, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    usize::deserialize(deserializer).and_then(|n| {
+        if n > 0 && n <= 16 {
+            Ok(n)
+        } else {
+            Err(serde::de::Error::custom(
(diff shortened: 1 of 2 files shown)


kirawi:
I think you can use `NonZeroUsize` instead.

bo1led-owl:
Thank you for the advice, I didn't know about this feature. I'll take some time to rewrite it

bo1led-owl:
I've changed it, now it's more clear and also fixes the original issue

pascalkuthe (commented):
While you are at it could you limit the value to something reasonable like 16 at most?I am pretty sure the code doesn't correctly account for tab-width larger than 8 and I can't think of any usecase

pascalkuthe (approved):
Yeah deserialization errors is how we usually handle invalid config. LGTM 👍

kirawi (approved):

Issue #2527, opened 2022-05-21
Panic if CWD gets deleted


zen3ger:
### Summary

If a document is being edited when the current working directory of helix has been deleted then helix will panic.

There's quite a lot of places where `std::env:current_dir()` just has an `.expect()` on it so I'm opening the issue instead of going ahead and patch things up so we can agree on how this should be handled in general.

### Reproduction Steps

 

I tried this:

1. `mkdir temp`
2. `touch myfile.txt`
3. `cd temp; hx myfile.txt`
4.  In a different terminal `rm -rf temp`
5. In gnome-terminal it was enough to click the terminal, but with Alacritty it needed to bring up e.g. the prompt

I expected this to happen:
Pop up that directory no longer exists or something sensible, but not a panic.

Instead, this happened:
```
thread 'main' panicked at 'couldn't determine current directory: Os { code: 2, kind: NotFound, message: "No such file or directory" }', helix-core/src/path.rs:87:45
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::result::unwrap_failed
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1785:5
   3: core::result::Result<T,E>::expect
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/result.rs:1035:23
   4: helix_core::path::get_relative_path
             at /home/zen3ger/Projects/helix/helix-core/src/path.rs:87:21
   5: core::ops::function::FnOnce::call_once
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/ops/function.rs:227:5
   6: core::option::Option<T>::map
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/option.rs:912:29
   7: helix_view::document::Document::relative_path
             at /home/zen3ger/Projects/helix/helix-view/src/document.rs:929:9
   8: helix_term::ui::editor::EditorView::render_statusline
             at /home/zen3ger/Projects/helix/helix-term/src/ui/editor.rs:777:28
   9: helix_term::ui::editor::EditorView::render_view
[...]

PR #7185, opened 2023-05-30
Fix crash when cwd is deleted

yo-main:
Should solve #2527

I didn't figure out how to set an error on the editor from some places.

Where is was crashing before, it now relies on the scratch buffer instead.
In some other places (not related to buffer, like the `open` command or the global search ), I was able to generate a message or rely on the default value of the PathBuffer.

Let me know any improvements/suggestions you can see, I'll be more than happy to correct them :)

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -3,8 +3,11 @@ pub mod grammar;
 
 use etcetera::base_strategy::{choose_base_strategy, BaseStrategy};
 use std::path::{Path, PathBuf};
+use std::sync::RwLock;
 
 pub const VERSION_AND_GIT_HASH: &str = env!("VERSION_AND_GIT_HASH");
+
+static CWD: RwLock<Option<PathBuf>> = RwLock::new(None);
 
 static RUNTIME_DIRS: once_cell::sync::Lazy<Vec<PathBuf>> =
     once_cell::sync::Lazy::new(prioritize_runtime_dirs);
@@ -12,6 +15,31 @@ static RUNTIME_DIRS: once_cell::sync::Lazy<Vec<PathBuf>> =
 static CONFIG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
 
 static LOG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
+
+// Get the current working directory.
+// This information is managed internally as the call to std::env::current_dir
+// might fail if the cwd has been deleted.
+pub fn current_working_dir() -> PathBuf {
+    if let Some(path) = &*CWD.read().unwrap() {
+        return path.clone();
+    }
+
+    let path = std::env::current_dir()
+        .and_then(dunce::canonicalize)
+        .expect("Couldn't determine current working directory");
+    let mut cwd = CWD.write().unwrap();
+    *cwd = Some(path.clone());
+
+    path
+}
+
+pub fn set_current_working_dir(path: PathBuf) -> std::io::Result<()> {
+    let path = dunce::canonicalize(path)?;
+    std::env::set_current_dir(path.clone())?;
+    let mut cwd = CWD.write().unwrap();
+    *cwd = Some(path);
+    Ok(())
+}
 
 pub fn initialize_config_file(specified_file: Option<PathBuf>) {
     let config_file = specified_file.unwrap_or_else(default_config_file);
@@ -217,7 +245,7 @@ pub fn merge_toml_values(left: toml::Value, right: toml::Value, merge_depth: usi
 /// If no workspace was found returns (CWD, true).
 /// Otherwise (workspace, false) is returned
 pub fn find_workspace() -> (PathBuf, bool) {
-    let current_dir = std::env::current_dir().expect("unable to determine current directory");
+    let current_dir = current_working_dir();
     for ancestor in current_dir.ancestors() {
         if ancestor.join(".git").exists() || ancestor.join(".helix").exists() {
             return (ancestor.to_owned(), false);
@@ -243,8 +271,20 @@ fn ensure_parent_dir(path: &Path) {
 mod merge_toml_tests {
     use std::str;
 
-    use super::merge_toml_values;
+    use super::{current_working_dir, merge_toml_values, set_current_working_dir};
     use toml::Value;
+
+    #[test]
+    fn current_dir_is_set() {
+        let new_path = dunce::canonicalize(std::env::temp_dir()).unwrap();
+        let cwd = current_working_dir();
+        assert_ne!(cwd, new_path);
+
+        set_current_working_dir(new_path.clone()).expect("Couldn't set new path");
+
+        let cwd = current_working_dir();
+        assert_eq!(cwd, new_path);
+    }
 
     #[test]
     fn language_toml_map_merges() {
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -163,7 +163,7 @@ impl Application {
         } else if !args.files.is_empty() {
             let first = &args.files[0].0; // we know it's not empty
             if first.is_dir() {
-                std::env::set_current_dir(first).context("set current dir")?;
+                helix_loader::set_current_working_dir(first.clone())?;
                 editor.new_file(Action::VerticalSplit);
                 let picker = ui::file_picker(".".into(), &config.load().editor);
                 compositor.push(Box::new(overlaid(picker)));
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2062,7 +2062,7 @@ fn global_search(cx: &mut Context) {
                 .map(|comp| (0.., std::borrow::Cow::Owned(comp.clone())))
                 .collect()
         },
-        move |_editor, regex, event| {
+        move |editor, regex, event| {
             if event != PromptEvent::Validate {
                 return;
             }
@@ -2076,6 +2076,11 @@ fn global_search(cx: &mut Context) {
                     .build();
 
                 let search_root = helix_loader::current_working_dir();
+                if !search_root.exists() {
+                    editor.set_error("Current working directory does not exist");
+                    return;
+                }
+
                 let dedup_symlinks = file_picker_config.deduplicate_links;
                 let absolute_root = search_root
                     .canonicalize()
@@ -2147,7 +2152,9 @@ fn global_search(cx: &mut Context) {
         let call: job::Callback = Callback::EditorCompositor(Box::new(
             move |editor: &mut Editor, compositor: &mut Compositor| {
                 if all_matches.is_empty() {
-                    editor.set_status("No matches found");
+                    if !editor.is_err() {
+                        editor.set_status("No matches found");
+                    }
                     return;
                 }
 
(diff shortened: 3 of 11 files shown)


helix-term/src/application.rs
```diff
@@ -569,12 +569,15 @@ impl Application {
...
-        ));
+        match get_relative_path(&doc_save_event.path) {
+            Ok(path) => self.editor.set_status(format!(
+                "'{}' written, {}L {}B",
+                path.to_string_lossy(),
+                lines,
+                bytes
+            )),
+            _ => self.editor.set_error("Error trying to save file"),
+        }
```
the-mikedavis, line 580: This isn't actually an error saving though, just an error formatting this path as relative. Instead of erroring here let's use the full path if we can't determine the relative one:

```suggestion
        let path =
            get_relative_path(&doc_save_event.path).unwrap_or_else(|_| doc_save_event.path.into());
        self.editor.set_status(format!(
            "'{}' written, {}L {}B",
            path.to_string_lossy(),
            lines,
            bytes
        ));
```

helix-term/src/commands.rs
```diff
@@ -2021,8 +2021,9 @@ fn global_search(cx: &mut Context) {
 
         fn format(&self, current_path: &Self::Data) -> Row {
             let relative_path = helix_core::path::get_relative_path(&self.path)
-                .to_string_lossy()
-                .into_owned();
+                .map_or(SCRATCH_BUFFER_NAME.to_owned(), |path| {
+                    path.to_string_lossy().into_owned()
+                });
```
the-mikedavis, line 2026: Callers of `get_relative_path` should not use `SCRATCH_BUFFER_NAME` as a fallback value if they can't find the relative path. `SCRATCH_BUFFER_NAME` should only be used when a document doesn't have a path (i.e. is a "scratch buffer"). Let's use the absolute path as a fallback when `get_relative_path` returns an error as above

helix-term/src/commands.rs
```diff
@@ -2067,8 +2068,12 @@ fn global_search(cx: &mut Context) {
...
                     .build();
 
-                let search_root = std::env::current_dir()
-                    .expect("Global search error: Failed to get current dir");
+                let search_root = std::env::current_dir();
+                if search_root.is_err() {
+                    _editor.set_error("Global search error: Failed to get current dir");
+                    return;
+                }
+                let search_root = search_root.unwrap();
```
the-mikedavis, line 2076: This can be cleaned up a little:

```rust
let search_root = match std::env::current_dir() {
    Ok(cwd) => cwd,
    Err(err) => {
        editor.set_error(format!("Global search error: {err}"));
        return;
    }
};
```

`_editor` here should have the underscore removed, and in the closure's parameter list above. It was underscored previously since it was unused.

helix-term/src/commands/lsp.rs
```diff
@@ -116,7 +116,7 @@ impl ui::menu::Item for SymbolInformationItem {
         } else {
             match self.symbol.location.uri.to_file_path() {
                 Ok(path) => {
-                    let get_relative_path = path::get_relative_path(path.as_path());
+                    let get_relative_path = path::get_relative_path(path.as_path()).unwrap();
```
the-mikedavis, line 119: This can use the absolute path rather than panicking

yo-main: Thanks, corrected !

helix-term/src/commands.rs
```diff
@@ -2543,9 +2551,10 @@ fn buffer_picker(cx: &mut Context) {
                 .path
                 .as_deref()
                 .map(helix_core::path::get_relative_path);
-            let path = match path.as_deref().and_then(Path::to_str) {
-                Some(path) => path,
-                None => SCRATCH_BUFFER_NAME,
```
the-mikedavis, line 2548: The existing code here was actually incorrectly returing `SCRATCH_BUFFER_NAME` when the `Path::to_str` call failed to validate the name as UTF-8. It should have been using `Path::to_string_lossy`, like:

```rust
let path = path
    .as_deref()
    .map_or_else(|| SCRATCH_BUFFER_NAME.into(), Path::to_string_lossy);
```

I think we should make a similar fix here after handling the Result from `get_relative_path`. But if `get_relative_path` fails then we should use the full absolute path rather than the `SCRATCH_BUFFER_NAME`.

Same thing for the jumplist picker below.

helix-view/src/document.rs
```diff
@@ -1549,7 +1549,7 @@ impl Document {
     pub fn relative_path(&self) -> Option<PathBuf> {
```
the-mikedavis, line 1549: For this case I think we should either return an `Option<Result<PathBuf>>` or keep the current `Option<PathBuf>` return and return the absolute path when `get_relative_path` fails (and document that behavior in the function docs here).

Or I suppose we could remove this function and have the callers use `Document::path` and handle getting the relative path themselves since the `Option<Result<PathBuf>>` type is so awkward.

helix-term/src/commands.rs
yo-main: Done !
 
I have done the same on the jumplist_picker just below

helix-view/src/document.rs
yo-main: I went for the second option -> keeping the current `Option<PathBuf>` but returning the absolute path as a fallback. It feel more coherent with other places where we apply the same logic :)

the-mikedavis (commented):
This is looking good, just needs the clippy lints fixed

yo-main:
should be fine now :crossed_fingers:

the-mikedavis (approved):

pascalkuthe (changes requested):
Sorry for being a bit late here. I put off reviewing this because I think k this is actually a bit more subtle to solve. IMO it's really unfortunate that get_cwd can fail (if the directory doesn't exist anymore) ideally I would still like to obtain the oath and then fail when attempting any IO.


This PR often uses `unwrap_or_default` or other suboptimal fallback strategies. I would really like to avoid this as that can lead to other issues.

I think a better approach would be to track CWD in helix internally. Just store the path in a static that is intalized on startup (and if we fail to acquire the cwd on startup we can just exit with an error). The only way CWD can change at runtime is by using the command so we only need to update the static in that command.

yo-main:
Thanks ! I tried to implement this by setting the cwd directly on the `Editor` itself. 

It's rather convenient as it's available in all the place I need it (which make me think it's also logical to put it there but I might be wrong ^^).

Probably something different from what you had in mind with a static string, but I'll be happy to do the necessary changes with a little guidance (:pray:) if the current version doesn't do the job.

Also, I kept using the PathBuf default value on the `find_workplace` function. The expect makes helix crash upon reloading the configuration, but having it returns a result do have some heavier impact in other parts of Helix and I couldn't figure out which value to rely on instead on crashing in some case (like when getting the workspace config file).
Please let me know your thoughts on this :)

pascalkuthe:
The idea was to have an `static CWD: RwLock<Path>` and just read from that instead of ever reaquring the cwd. This makes the operation infallible so you won't need to do error handling (or unwrap_or_default) at all. If the CWD doesn't exist we will simply fail if we try to access the pathwhich already has (much more robust) error handling. Especially in the workspace related functions using unwrap_or_default can lead to incorrect behavior

yo-main:
I reworked the whole PR based on that idea ! 
Can you have a look and let me know what you think about it ? :)

helix-loader/src/lib.rs
```diff
@@ -3,14 +3,30 @@ pub mod grammar;
 
 use etcetera::base_strategy::{choose_base_strategy, BaseStrategy};
 use std::path::{Path, PathBuf};
+use std::sync::RwLock;
 
 pub const VERSION_AND_GIT_HASH: &str = env!("VERSION_AND_GIT_HASH");
 
+static CWD: once_cell::sync::Lazy<RwLock<PathBuf>> = once_cell::sync::Lazy::new(|| {
```
pascalkuthe, line 10: You don't need the `Lazy` here just use an RwLock<Option<...>>>` instead . You only access the global in the wrapper function anyway so you can simply fill the option there

pascalkuthe (commented):
This looks great for the most part just one more minor thing

helix-loader/src/lib.rs
yo-main: Ah yeah, that's a nice way to avoid that `Lazy`. Updated the PR, not sure in the best way but let me know !

pascalkuthe:
Sorry about that 😅

helix-loader/src/lib.rs
```diff
@@ -3,14 +3,35 @@ pub mod grammar;
...
 
 static CONFIG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
 
+pub fn current_working_dir() -> PathBuf {
+    if let Some(path) = &*CWD.read().unwrap() {
+        return path.clone();
+    }
+
+    let cwd = std::env::current_dir().expect("Couldn't determine current working directory");
+    set_current_working_dir(cwd.clone()).expect("Couldn't set current working directory const");
```
pascalkuthe, line 23: Not instead of calling the setter functionsi.only accessaccess the global directly here since that avoids an unnecessary call to `std::env::set_current_dir`

pascalkuthe:
If you fix my.last comment and addressthe Lint's so.ci passes this should.be good to go I think

pascalkuthe (approved):
LGTM thanks for sticking with this to the end :+1:

the-mikedavis (commented):
It looks like we will need to do some path expansion. `mkdir tmp` and then either `cargo r tmp` or `cargo r` and then `:cd tmp`: The directory will change correctly but `current_working_dir` will return `tmp` and not the full path. This messes up consumers like `file_picker` that are expecting an expanded path.

I'm not totally sure where this expansion should belong. We could move https://github.com/helix-editor/helix/blob/c33795e770808f11c975f30848067ce1c7ee85be/helix-core/src/path.rs#L88-L97 over to helix-loader and call it from `set_current_working_dir`. That would shift around some dependencies and code between helix-core and helix-loader but might be the cleaner option. Or we could require that each caller of `set_current_working_dir` passes in an absolute path (maybe a debug_assert could work here). What do you think?

pascalkuthe:
> It looks like we will need to do some path expansion. `mkdir tmp` and then either `cargo r tmp` or `cargo r` and then `:cd tmp`: The directory will change correctly but `current_working_dir` will return `tmp` and not the full path. This messes up consumers like `file_picker` that are expecting an expanded path.
> 
> I'm not totally sure where this expansion should belong. We could move
> 
> https://github.com/helix-editor/helix/blob/c33795e770808f11c975f30848067ce1c7ee85be/helix-core/src/path.rs#L88-L97
> over to helix-loader and call it from `set_current_working_dir`. That would shift around some dependencies and code between helix-core and helix-loader but might be the cleaner option. Or we could require that each caller of `set_current_working_dir` passes in an absolute path (maybe a debug_assert could work here). What do you think?

Good catch! .we can actually use `dunce::cannonicolize` derctly here I think. That canonicalization function only makes a difference if the path doesn't currently exits (wich is not the case here since the working directory must exist)

yo-main:
updated ! thanks for spotting :)

pascalkuthe (dismissed):

helix-term/src/commands.rs
the-mikedavis, line 1: For the two changes here we will probably want some feedback:

```rust
if !cwd.exists() {
    editor.set_error("Current working directory does not exist");
    return;
}
```

since the file_picker_in_current_directory won't show any files and global_search won't be able to find any matches if the cwd doesn't exist.

the-mikedavis (approved):
Looks good 👍

I like the "deleted" message in the `show_current_directory` too, that's a nice touch

pascalkuthe (approved):

PR #7189, opened 2023-05-31
Add config option for instant completion entry preview.

luca-schlecker:
This PR adds the config option `preview_completion_insert` which controls the insertion behavior of autocompletion.
Currently, Helix inserts the selected completion entry instantly into the document (reversible with c-c), this behavior is a personal preference though and should be configurable, as I prefer it to only insert on confirmation.

The new option defaults to true, thus the default behavior is not changed.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -52,6 +52,7 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `auto-format` | Enable automatic formatting on save | `true` |
 | `auto-save` | Enable automatic saving on the focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal | `false` |
 | `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. Used for autocompletion, set to 0 for instant | `400` |
+| `preview-completion-insert` | Whether to apply completion item instantly when selected | `true` |
 | `completion-trigger-len` | The min-length of word under cursor to trigger autocompletion | `2` |
 | `completion-replace` | Set to `true` to make completions always replace the entire word and not just the part before the cursor | `false` |
 | `auto-info` | Whether to display info boxes | `true` |
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -110,6 +110,7 @@ impl Completion {
         start_offset: usize,
         trigger_offset: usize,
     ) -> Self {
+        let preview_completion_insert = editor.config().preview_completion_insert;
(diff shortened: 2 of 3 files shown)


CptPotato:
I think somebody mentioned plans that the completion preview should use virtual text in the future, instead of inserting it as actual text until confirmed. I'm not sure if this is actively being worked on, though.

book/src/configuration.md
```diff
@@ -52,6 +52,7 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `auto-format` | Enable automatic formatting on save | `true` |
 | `auto-save` | Enable automatic saving on the focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal | `false` |
 | `idle-timeout` | Time in milliseconds since last keypress before idle timers trigger. Used for autocompletion, set to 0 for instant | `400` |
+| `preview-completion-insert` | Whether to insert the completion suggestion on hover | `true` |
```
pascalkuthe, line 55:
```suggestion
| `preview-completion-insert` | Whether to apply completion item instantly when selected| `true` |
```

helix-term/src/ui/completion.rs
```diff
@@ -230,7 +231,7 @@ impl Completion {
 
             match event {
                 PromptEvent::Abort => {}
-                PromptEvent::Update => {
+                PromptEvent::Update => if preview_completion_insert {
```
pascalkuthe, line 234: this will fail exhaustivenss checks. Have you tested this code?

pascalkuthe (commented):
> I think somebody mentioned plans that the completion preview should use virtual text in the future, instead of inserting it as actual text until confirmed. I'm not sure if this is actively being worked on, though.

that was an idea that some people mentioned in #6594 but I don't think it's feasible. I am not sure if we want a config opinion for this we are usually a bit hesitant to introduce those. The behavior here matcher VSCode/Intellij so this is pretty established behavior. Considering that the changes to get this working are quite small I would lean Towards supporting this. The implementation itself LGTM although I don't think it compiles currently. Please do some testing to ensure this actually works as intended for you

helix-term/src/ui/completion.rs
pascalkuthe: Ah I got  tripped up by the if not being in braces that looked like a match guard 9which it probably should be)

luca-schlecker: An empty arm was need anyway, fixed it now. 👍

pascalkuthe (approved):

dead10ck (approved):
Looks good to me, thanks!

PR #2796, opened 2022-06-17
add prompt history search by input

QiBaobin:

PR #7204, opened 2023-06-01
Use prefix search on the line when navigating prompt history

xlambein:
I've implemented a feature I've been missing in Helix for a while: (prefix) search in the prompt history based on the current line. This is a similar behaviour to what you'll find in many shells and REPLs.

For example, this allows the user to type `:o` and then search up and down through all the files they've opened previously.

### DIFF
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -28,6 +28,7 @@ pub struct Prompt {
     selection: Option<usize>,
     history_register: Option<char>,
     history_pos: Option<usize>,
+    history_search: Option<String>,
     completion_fn: CompletionFn,
     callback_fn: CallbackFn,
     pub doc_fn: DocFn,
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/prompt.rs
```diff
@@ -304,19 +318,29 @@ impl Prompt {
...
+                .find(filter),
+            CompletionDirection::Backward => values
+                .iter()
+                .enumerate()
+                .take(self.history_pos.unwrap_or(values.len()))
+                .rev()
+                .find(filter),
+        }) else {
+            return
+        };
```
poliorcetics, line 337:
```suggestion
       let filtered = match direction {
            CompletionDirection::Forward => values
                .iter()
                .enumerate()
                .skip(self.history_pos.map_or(0, |i| i + 1).min(end))
                .find(filter),
            CompletionDirection::Backward => values
                .iter()
                .enumerate()
                .take(self.history_pos.unwrap_or(values.len()))
                .rev()
                .find(filter),
        };
        let Some((index, _)) = filtered else {
            return
        };
```

is much clearer IMO, else the `return` is easily missed

the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/2796

xlambein:
Ahh shoot, I didn't search properly before writing this :roll_eyes:

PR #7207, opened 2023-06-01
Refactor queries for ecma based languages

yusuphgammut:
## Intro

This PR builds on top of https://github.com/helix-editor/helix/pull/3301, which created the "fake" ecma language to store queries that are common to both javascript and typescript.

## The problem

Ecma-based languages share many traits, but it's not a linear and simple inheritance. Sometimes we want to inherit queries from one of those languages without inheriting the queries that that language inherits. The current state of inheritance is the following:

```
CURRENT INHERITANCE MODEL 
-----------------
javascript <--- ecma
jsx <---------- ecma
typescript <--- ecma
tsx <---------- jsx, typescript
```

There are three main examples why this current model generates limitations and bugs:

- In tsx, the `<` and `>` symbols are being highlighted as operators in the context of generic types like `MyGenericType<GenericArgument>`. I wanted to fix this (change it to punctuation), but I realized that tsx was inheriting highlight queries from both jsx and typescript (check [here](https://github.com/helix-editor/helix/blob/04fbf30488348e417e3922ef23973d4542923731/runtime/queries/tsx/highlights.scm)), and both of them were inheriting ecma. The problem is that the queries in ecma for `<` and `>` (that assigned them to operator highlight, [here](https://github.com/helix-editor/helix/blob/7f5940be80eaa3aec7903903072b7108f41dd97b/runtime/queries/ecma/highlights.scm#L125)) inherited by jsx where overwriting the queries for generic types in typescript ([here](https://github.com/helix-editor/helix/blob/7f5940be80eaa3aec7903903072b7108f41dd97b/runtime/queries/typescript/highlights.scm#LL60C7-L60C7)). If I then changed the order and added typescript first, then those same operator queries in ecma inherited by typescript overwrite the `<` and `>` symbols for jsx elements ([here](https://github.com/helix-editor/helix/blob/7f5940be80eaa3aec7903903072b7108f41dd97b/runtime/queries/jsx/highlights.scm#L9)), highlighting them as operators.
- Aside from the particular problem above, ecma is being inherited twice in tsx, which is not ideal.
- Javascript and typescript treesitter grammars have diverged in some parts, and that's specially evident in relation to parameters, where javascript defines `formal_parameters` and typescript adds `required_parameter` and `optional_parameter`. This affects the way queries should be written in particular for highlights and locals. We then should put those specific queries under javascript and typescript (not under ecma) and those queries should be inherited by jsx and tsx respectively. However, tsx also benefits from inheriting things from jsx, but it shouldn't inherit javascript specific queries, just jsx queries.

We can all agree that the javascript ecosystem is quite messy and that it presents some unique challenges for tooling developers.

So, as stated above, Ecma based languages share many traits, but it's not a linear and simple inheritance. Because of this we want to share as many queries as possible while avoiding nested inheritance in the queries that can make query behavior unpredictable due to unexpected precedence.

## The solution

This PR proposes "public" and "private" versions for javascript, jsx, and typescript, that share the same name, but the "private" version name starts with an underscore (ecma already exists as a sort of base "private" language). __This allows the "private" versions to host the specific queries of the language excluding any `; inherits` statement, in order to make them safe to be inherited by the "public" version of the same language and other languages as well__. This helps fixing the problems above. The tsx language doesn't have a "private" version given it doesn't need to be inherited by other languages.

```
INHERITANCE MODEL 
-----------------
javascript <--- _javascript, ecma
jsx <---------- _jsx, _javascript, ecma
typescript <--- _typescript, ecma
tsx <---------- _jsx, _typescript, ecma
```

In summary, I've... 
1. Created `_javascript`, `_jsx`, and `_typescript` languages (called "private" above).
2. Moved existing queries to those new "private" versions.
3. Changed query files in the `javascript`, `jsx`, `typescript`, and `tsx` languages (called "public" above) to only inherit from "private" languages.

In addition to this, I've adjusted some highlights and locals queries to take advantage of this inheritance and fix existing problems. I've put special attention to queries related to parameters, given it is a place where javascript and typescript treesitter grammars have diverged the most (you can use the following files to test that, just add the correct extensions: 
[test-js.txt](https://github.com/helix-editor/helix/files/11630402/test-js.txt), [test-jsx.txt](https://github.com/helix-editor/helix/files/11630403/test-jsx.txt), [test-ts.txt](https://github.com/helix-editor/helix/files/11630404/test-ts.txt), [test-tsx.txt](https://github.com/helix-editor/helix/files/11630405/test-tsx.txt))

Textobjects can be improved too, but I plan to address that in another PR. Let me know if this seems like a good strategy to tackle this issue.


yusuphgammut:
@the-mikedavis Sorry for tagging you directly. You were the author of https://github.com/helix-editor/helix/pull/3301, so I think it'll be good to know your opinion on this proposal. I know it touches quite many files, but it's mostly moving existing queries to new files. Let me know your thoughts.

yusuphgammut:
This proposal is mostly finished. @kirawi I was wondering if you can add the "S-waiting-on-review" label to this PR so that any of the core team can take a look when they have time available.

runtime/queries/javascript/highlights.scm
```diff
@@ -1,38 +1,14 @@
...
+; query behavior unpredictable due to unexpected precedence. To achieve that,
+; some ecma related languages have "public" and "private" versions that share
+; the same name, but the "private" version name starts with an underscore (with
+; the exception of ecma, that works as the base "private" language). This
+; allows the "private" versions to host the specific queries of the language
+; excluding any "inherits" statement, in order to make them safe to be
+; inherited by the "public" version of the same language and other languages
+; as well.
+; If you plan to add queries to this language, please consider adding them to
+; any of the inherited languages listed below.
```
the-mikedavis, line 12: Rather than duplicating this block everywhere, let's add a file `runtime/queries/ecma/README.txt` that explains the public/private queries. Then each of these files can just say "see runtime/queries/ecma/README.md"

the-mikedavis (commented):
I think this makes sense. It fragments the queries a little but I don't see another way around it without undoing all of the `inherits`

runtime/queries/javascript/highlights.scm
yusuphgammut: Good idea! I've applied this change in the latest commit.

yusuphgammut:
I've rebased this branch with main so we have the latest version. If there are no additional comments, it is safe to merge this with master.

the-mikedavis (approved):

pascalkuthe (approved):

PR #7220, opened 2023-06-03
Maintain the current cursor's position and view in the vsplit/hsplit commands too

bnjbvr:
First of all, thanks for this editor; it's all I've ever wanted from nvim, without the configuration hassle.

I've been quite surprised to discover today that doing `C-w C-v` would keep the current cursor's position and view in document, while the `:vsplit` command would not do that and reset the cursor to the top of the screen. This PR implements a consistent behavior in both cases, which is that of `C-w C-v` which I personally find more pleasant.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4636,17 +4636,19 @@ fn transpose_view(cx: &mut Context) {
     cx.editor.transpose_view()
 }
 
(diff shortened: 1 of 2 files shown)


bnjbvr:
(closing to declutter my PR inbox, feel free to cherry-pick and merge if you still want it :))

PR #7226, opened 2023-06-04
Fix for saving from insert mode

### LINKED ISSUE #3501: Mapped commands that would exit into normal mode do not propagate in history

### LINKED ISSUE #6513: Keybinding for ["normal_mode", ":write"] doesn't remove file-modification-indicator
### Summary

 A keybinding like
```toml
[keys.insert]
k = { j = ["normal_mode", ":write"] }
```
Will exit insert mode and write the file, but the `file-modification-indicator` will remain in the bufferline and statusline (if configured) as if the file were not saved.

This appears to be specific to having the `:write` be a second command in a list, or when going from insert mode to normal mode. Key bindings that save when triggered in normal mode work as expected.

### Reproduction Steps

_No response_

### Helix log

Log shows stuff like this:
```
2023-03-31T11:08:13.875 helix_view::document [DEBUG] id 9 modified - last saved: 0, current: 0
2023-03-31T11:08:13.875 helix_view::document [DEBUG] id 10 modified - last saved: 10, current: 10
2023-03-31T11:08:13.875 helix_view::document [DEBUG] id 7 modified - last saved: 0, current: 4
```

### Platform

macOS

### Terminal Emulator

wezterm 20230326-111934-3666303c

### Helix Version

helix 23.03 (3cf03723)

omentic:
Fixes #6513, fixes #3501, addresses #1583. This somewhat changes the granularity of the undo history: but keeps the preciseness. While adding to the undo history on a save may not seem idiomatic conceptually at first: I think it's the best way to address the issue, because 1) saving a file is somewhat of a "finishing" action and 2) comparing the undo history with the current state of a file is quite a nice way to check if something has been modified and I would prefer not to mess with it.

I also added e2a877432ce7f75eba60d5dd579e5a0ab45b73e8 because I stumbled across it while tracking down this issue and thought that having a separate history entry for a raw paste before any manipulation of it would be desirable. I'd be fine rolling it back but that also propagating to history makes more sense to me.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -834,6 +834,10 @@ impl EditorView {
         cxt.editor.autoinfo = self.keymaps.sticky().map(|node| node.infobox());
 
         let mut execute_command = |command: &commands::MappableCommand| {
+            let doc = doc_mut!(cxt.editor);
+            let view = view_mut!(cxt.editor);
+            doc.append_changes_to_history(view);
+
             command.execute(cxt);
             helix_event::dispatch(PostCommand { command, cx: cxt });
 


archseer:
https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126

j-james:
While committing changes from insert mode is perhaps not idiomatic, this fixes a bug. Any shortcuts to save a file from insert mode are currently broken: they do not update the file modification indicator and Helix thinks the file has not been written and will not let you close it, see #6513.

This shouldn't terribly change how anyone not binding to `:write` uses Helix, and makes Helix more usable for anyone binding to `:write`.

Xalfer:
Looking at the changes you add `doc.append_changes_to_history(view)` into `write_impl` but not into `write_all_impl`.

helix-term/src/ui/editor.rs
```diff
@@ -1362,17 +1362,11 @@ impl Component for EditorView {
...
                     let view = view_mut!(cx.editor, focus);
                     let doc = doc_mut!(cx.editor, &view.doc);
 
                     view.ensure_cursor_in_view(doc, config.scrolloff);
-
-                    // Store a history state if not in insert mode. This also takes care of
-                    // committing changes when leaving insert mode.
-                    if mode != Mode::Insert {
-                        doc.append_changes_to_history(view);
-                    }
```
the-mikedavis, line 1375: This insert-mode check is important: without gating this on `mode != Mode::Insert` we get an undo checkpoint for every modification in insert mode. For example `ifoo<esc>` has an undo checkpoint for each character in `foo`.

helix-term/src/commands/typed.rs
```diff
@@ -732,6 +734,9 @@ pub fn write_all_impl(
             None
         };
 
+        let view = view_mut!(cx.editor);
+        doc.append_changes_to_history(view);
```
the-mikedavis, line 738: I think we should solve this more generically and call `append_changes_to_history` and `ensure_cursor_in_view` before and after executing any typable command, i.e. in this block: https://github.com/helix-editor/helix/blob/64326698225999016300423c4b90f8a8c7b8f38b/helix-term/src/commands.rs#L197-L206

Few typable commands modify the buffer so both `append_changes_to_history`s would be a no-op in most cases. This would also prevent bugs like https://github.com/helix-editor/helix/issues/9619, https://github.com/helix-editor/helix/issues/5733, also see https://github.com/helix-editor/helix/pull/5790

the-mikedavis: Actually this is not totally correct :point_up:

We should save checkpoints before/after each command in sequences too, in this block instead: https://github.com/helix-editor/helix/blob/b7b6f300841bb61d8833fee1c58d0e3670849b61/helix-term/src/ui/editor.rs#L836-L859

This doesn't solve all of the problems of writing in command sequences (like https://github.com/helix-editor/helix/issues/9705) but it should make history commits more consistent

David-Else:
Why not just stop the ability to save in insert mode? `write` could check you are in normal mode, and if not return an error. It would enforce best practice. https://github.com/helix-editor/helix/pull/2883#pullrequestreview-1029452126 explained, and [archseer](https://github.com/archseer) amplified this sentiment by quoting it.

I admit when I first started using Helix I bound ctrl-s to save and saved all the time in insert mode. This was a mistake and led me down the wrong path for using Helix, if it were not possible I never would have wasted my time doing it.

the-mikedavis:
I think it's worth fixing this regardless of the use-case, although I'm not a huge fan of the saving-in-insert-mode use-case. Ideally we shouldn't have to think about making commit checkpoints in typeable commands like `:clipboard-paste-replace` - it's pretty easy to forget to append changes to history there and accidentally introduce a bug. Plus I think it might be possible to trigger the same problems as saving in insert mode with command sequences in normal mode

pascalkuthe:
yeah I don't care for the saving in insertmode usecase that much but I do care about the sequence of commands (in normal mode) usecase

helix-term/src/commands/typed.rs
omentic: Oh fantastic, I was searching for something like this when writing this patch but couldn't find it. That makes for much nicer behavior.

omentic:
@the-mikedavis Feedback addressed, ready for re-review. Seems to work well locally.

Solsticely:
I've been daily driving this for about a week now and it fixes a bunch of stability issues i've been having :)

the-mikedavis:
This improves the situation but I believe we need a larger change to make sure we're treating all `execute_command`s roughly the same. I have a patch that I'm working on locally in https://github.com/the-mikedavis/helix/commit/f7b2f770730cf9a8e0364affe9e78782b830080e that I will turn into a PR if I can iron out the bugs

omentic:
@the-mikedavis any update on your end? I'll be rebasing my fork soon and would like to try out your patch, if it's finished.

Though IMO I think this should just be merged as-is and improved later, because issues related to this (#9705) seem to keep popping up.

Issue #7230, opened 2023-06-04
Instructions to link runtime on Linux/macOS can have a bad result


rseymour:
The docs recommend the command:
```bash
ln -s $PWD/runtime ~/.config/helix/runtime
```

Which works if and only if you don't already have a `~/.config/helix/runtime` So for folks going from a package install to a source install, they get an automatically broken runtime directory. ie they'll end up with something like:

`~/.config/helix/runtime/runtime`

This can be fixed with something like:
`ln -Ts $PWD/runtime ~/.config/helix/runtime` which fails if `~/.config/helix/runtime` already exists, at which point the user could delete it or something.

PR #7231, opened 2023-06-04
Update install.md instructions regarding symlinks

rseymour:
Fixes `ln` command line bug that could hit users moving from packaged to source builds.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -207,11 +207,7 @@ Or, create a symlink in `~/.config/helix` that links to the source code director
 ln -Ts $PWD/runtime ~/.config/helix/runtime
 ```
 
-If the above command notes that the directory exist either move `~/.config/helix/runtime` to a new location or delete it, then run the symlink command above again. For example to move it:
-
-```sh
-mv ~/.config/helix/runtime ~/.config/helix/runtime.old
-```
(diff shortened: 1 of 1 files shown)


rseymour:
fixes #7230

book/src/install.md
```diff
@@ -204,7 +204,13 @@ HELIX_RUNTIME=/home/user-name/src/helix/runtime
...
 ```sh
-ln -s $PWD/runtime ~/.config/helix/runtime
+ln -Ts $PWD/runtime ~/.config/helix/runtime
+```
+
+If the above command notes that the directory exist either move `~/.config/helix/runtime` to a new location or delete it, then run the symlink command above again. For example to move it:
+
+```sh
+mv ~/.config/helix/runtime ~/.config/helix/runtime.old
 ```
```
the-mikedavis, line 214: I don't think we should include a code sample for this: it's common enough knowledge how to move around directories. Some people run every codeblock without reading the associated text so I'm worried the codeblock will cause more installation problems than it solves.

rseymour: That's a good point, I just couldn't believe the 'simplicity' of my broken install when I found it.  I'll edit the PR.

the-mikedavis (approved):

pascalkuthe (approved):

PR #7233, opened 2023-06-05
Jinja language family syntax support

vanarok:
Add jinja grammar for nunjuncks and jinja languages.


languages.toml
```diff
@@ -2577,3 +2577,25 @@ indent = { tab-width = 4, unit = "\t" }
...
+file-types = ["njk"]
+roots = []
+indent = { tab-width = 2, unit = "  " }
+grammar = "jinja2"
+
+[[language]]
+name = "jinja"
+scope = "text.html.jinja"
+injection-regex = "jinja"
+file-types = ["jinja"]
```
the-mikedavis, line 2594:
```suggestion
file-types = ["jinja", "jinja2", "j2"]
```

the-mikedavis (approved):
Looks good. Could you resolve the conflicts?

the-mikedavis (approved):
In the long run I think we'll want a way to detect the language that is being templated. Jinja2 is usually used for HTML but you can use it for any kind of template so usually I see `filename.<ext>.j2` where `<ext>` is `html`, `js`, etc. depending on the language being templated. Ideally we would inject that language but that can be left for future work.

pascalkuthe (approved):

Issue #5177, opened 2022-12-15
Incorrect bracket matching and broken syntax highlighting with this chunk of C text


creikey:
### Summary

Typing `mm` to go to matching bracket on the top bracket goes to the wrong one, and fiddeling with the file changing the grammar of things (e.g `transform_scope` => `transform_scope()`) fixes it. So does removing any line inside of the `pipeline_scope`.

When the highlighting on the outermost scope is colored blue as in the first screenshot:
![image](https://user-images.githubusercontent.com/28424206/207977047-997cbc5d-4aad-404f-91d0-17cecf69261b.png)

That means it's broken as seen in the incorrectly highlighted bracket, and when it's green that means it's working as shown below:
![Uploading image.png…]()

### Reproduction Steps

Open helix and paste this buffer of C code into the file, saving it as `tmp.c` or something so it recognizes it as C:

```c
{
  transform_scope
  {
    pipeline_scope(goodpixel_pipeline)
    {
      sgp_set_image(0, image_rightclick);
      cpVect draw_at = cpvadd(entity_pos(b), cpv(BOX_SIZE, 0));
      rotate_at(-entity_rotation(b) - rotangle(b->compass_rotation), draw_at.x, draw_at.y);
      draw_texture_centered(draw_at, BOX_SIZE + sin(exec_time * 5.0) * BOX_SIZE * 0.1);
      sgp_reset_image(0);
    }
  }
}

```

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-12-15T14:05:58.168 helix_lsp::transport [ERROR] err <- "I[14:05:58.168] --> reply:textDocument/signatureHelp(25) 18 ms\r\n"
2022-12-15T14:05:58.205 helix_lsp::transport [ERROR] err <- "I[14:05:58.205] ASTWorker building file C:\\Users\\Cameron\\Documents\\flight\\tmp.c version 75 with command inferred from C:/Users/Cameron/Documents/flight/thirdparty/enet/callbacks.c\r\n"
2022-12-15T14:05:58.205 helix_lsp::transport [ERROR] err <- "[C:/Users/Cameron/Documents/flight/]\r\n"
[...]

PR #7238, opened 2023-06-05
Fix behaviour of matching when tree-sitter grammar is available

alevinval:
If the current character is a valid bracket, resort to plaintext matching. Otherwise, use regular tree-sitter node traversal.
This is a pragmatic solution that I propose, as addressing the tree-sitter grammars seems way more complex and tedious.
Why is believe this is acceptable? If the user is not on top of a valid bracket, regular operation will take place. It simply deals with the edge cases where the user _is on top_ of a bracket, which brings the unexpected behaviour that users complain about (see https://github.com/helix-editor/helix/issues/3357, https://github.com/helix-editor/helix/issues/5177)

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -50,6 +50,10 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
     let tree = syntax.tree();
     let pos = doc.char_to_byte(pos);
 
+    if is_valid_bracket(doc.char(pos)) {
+        return find_matching_bracket_plaintext(doc, pos);
+    }
+
     let mut node = tree.root_node().named_descendant_for_byte_range(pos, pos)?;
 
     loop {
@@ -85,10 +89,7 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
 ///
 /// If no matching bracket is found, `None` is returned.
 #[must_use]
-pub fn find_matching_bracket_current_line_plaintext(
-    doc: &Rope,
-    cursor_pos: usize,
-) -> Option<usize> {
+pub fn find_matching_bracket_plaintext(doc: &Rope, cursor_pos: usize) -> Option<usize> {
     // Don't do anything when the cursor is not on top of a bracket.
     let bracket = doc.char(cursor_pos);
     if !is_valid_bracket(bracket) {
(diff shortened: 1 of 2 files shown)


helix-core/src/match_brackets.rs
```diff
@@ -85,10 +89,7 @@ fn find_pair(syntax: &Syntax, doc: &Rope, pos: usize, traverse_parents: bool) ->
 ///
 /// If no matching bracket is found, `None` is returned.
 #[must_use]
-pub fn find_matching_bracket_current_line_plaintext(
-    doc: &Rope,
-    cursor_pos: usize,
-) -> Option<usize> {
+pub fn find_matching_bracket_plaintext(doc: &Rope, cursor_pos: usize) -> Option<usize> {
```
alevinval, line 92: 😓 even after having so much looks into it, and fixing the docs, I missed that the function name was left not updated. So I'm addressing it here, removing mention of only working on the current line.

alevinval:
The original proposed solution is quite simple, just look at the first commit: https://github.com/helix-editor/helix/pull/7238/commits/d0eae3382a99da504da4ed9c7d58d255971c9f28

It could be argued that this simple solution is wasteful in case that we are on top of a valid bracket, *and* we are right on the boundary of the node, and this node boundaries are valid bracket pairs. Under such case, we would be resorting to plaintext matching, when we could achieve the matching much faster by looking at the node positions. So... I have pushed a commit that would address those concerns: https://github.com/helix-editor/helix/pull/7238/commits/1fb75bb6835b2717a1efdfd6955e01c7f90b9d1b

Either option works, and I have no strong feelings reverting the second commit, if you think the optimisation is an overkill.

pascalkuthe:
This is actually the opposite direction of what I was looking to do. TS is generally preferable to text matching because it handles edge cases much better. I want to move miX and maX to TS too, see https://github.com/helix-editor/helix/pull/6893#pullrequestreview-1405086580. The TS matching just needed some improvements. I had these in mind but never got around to it but I nave now implemented them in https://github.com/helix-editor/helix/pull/7242. That PR also contains an (intentionally) very limited plaintext fallgack to handle... Mostly horrible plaintext macros https://github.com/helix-editor/helix/issues/3357#issuecomment-1503837732... I might undo that tough we could handle those cases by injecting C into C preprocessor tokens the way nvim does

alevinval:
@pascalkuthe oh, and you have it implemented already. Let me close this one, yours is looking much better and keeping in on TS.

Issue #4505, opened 2022-10-28
Allow indenting when no character other than white space characters are present in a line


Multirious:
For some reason a line must have a non white space character to indent with > or < which is not a problem for vim. If you are going to indent in insert mode, the auto complete will annoy you the **** out with the tab if you're not fast enough or you have to press Ctrl-c every time.

pascalkuthe:
Closing, see https://github.com/helix-editor/helix/pull/7241#issuecomment-1578657785

PR #7241, opened 2023-06-05
fix(commands.rs): remove check for empty line from indent() function

Hessesian:
refs #4505

There is a check for empty line, however this check is probably useless, as one can indent to desired space without any text present and then start typing, having it offers no benefit

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3994,10 +3994,6 @@ fn indent(cx: &mut Context) {
     let transaction = Transaction::change(
         doc.text(),
         lines.into_iter().filter_map(|line| {
-            let is_blank = doc.text().line(line).chunks().all(|s| s.trim().is_empty());
-            if is_blank {
-                return None;
-            }
             let pos = doc.text().line_to_char(line);
             Some((pos, pos, Some(indent.clone())))
         }),


CptPotato:
I wonder if this was intentional. In many languages indents are stripped from empty lines when auto-formatting, so the current behavior does make sense to me.

CptPotato:
> I could see several more approaches:
>
> 1. Only one line is selected and has no characters -> do indent
> 2. More lines are selected and has no characters -> also do indent ?
> 3. More lines selected and has some empty lines and some non-empty -> indent only non-empty

I'm personally not a fan of special handling like this, since that makes the behavior harder to reason about and can lead to odd edge cases. So if the indent is changed it should be kept consistent in my opinion.

Maybe a config option is the right thing here? Though I'm curious about what the maintainers think.

Hessesian:
I have updated the commit to reflect the discussion, now the behaviour will be same as before with exception of all blank selection of lines, which will do the indentation, I'll leave the maintainers opinion on which commit is more desired

the-mikedavis:
Yeah the current behavior is intentional, you can split the selection on newlines with A-s and then indent in insert mode with tab to indent empty lines. I don't like the nuance introduced here where indentation works for empty lines if the selection is all empty lines.

pascalkuthe:
Yeah I think we are not going to go forward with this. I agree with @the-mikedavis points

PR #7261, opened 2023-06-06
Add language support for persistent library syntax

lykahb:
This adds support for the language used by a [persistent](https://hackage.haskell.org/package/persistent) library to define the database models. This syntax is only used by one library, so I chose to prefix the language name with "haskell-" for specificity.

The haskell [injections query](https://github.com/helix-editor/helix/blob/master/runtime/queries/haskell/injections.scm) uses the name of the function that handles that content of the quasi-quote as the name of the language. The persistent library has three of them that I included into the `injection-regex`. 

image

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -59,6 +59,7 @@
 | graphql | ✓ |  |  |  |
 | hare | ✓ |  |  |  |
 | haskell | ✓ | ✓ |  | `haskell-language-server-wrapper` |
+| haskell-persistent | ✓ |  |  |  |
 | hcl | ✓ |  | ✓ | `terraform-ls` |
 | heex | ✓ | ✓ |  | `elixir-ls` |
 | hosts | ✓ |  |  |  |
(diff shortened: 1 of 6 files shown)


lykahb:
I also tried an alternative recognition of the injected language that moves the recognition from injection-regex into the haskell injection query. Let me know which variant you consider cleaner.

``` scheme
; haskell/injections.scm

((comment) @injection.content

(quasiquote
 (quoter) @_quoter
 ((quasiquote_body) @injection.content
  (#match? @_quoter "(persistWith|persistLowerCase|persistUpperCase)")
  (#set! injection.language "haskell-persistent" )
 )
)

(quasiquote
 (quoter) @injection.language
 (quasiquote_body) @injection.content)
```

``` toml
# languages.toml

[[language]]
name = "haskell-persistent"
scope = "source.persistentmodels"
injection-regex = "haskell-persistent"
file-types = ["persistentmodels"]
roots = []
comment-token = "--"
indent = { tab-width = 2, unit = "  " }
```

lykahb:
Moved recognition of language name into the haskell injection queries, added folds, and marked `=` as an operator. Now the language support is closer to the one that got merged into [nvim-treesitter](https://github.com/nvim-treesitter/nvim-treesitter/pull/4990).

the-mikedavis (approved):

pascalkuthe (approved):

PR #7265, opened 2023-06-07
Refactor Picker as a table with multiple columns/filters

the-mikedavis:
The `Menu`-like portion of the `Picker` is replaced with a table. `Picker::new` takes a set of `Column`s which each contain custom logic for rendering and optionally sorting and filtering an option. (Previously `ui::menu::Item` implemented that logic.) This allows filtering a picker by more than one value.

Here's a small example showing changes to the symbol picker (`<space>s`) and the buffer picker (`<space>b`):

demo

Here I open up the symbol picker (`<space>s`) which has two columns: symbol kind (function, enum, module, class, etc.) and name. By default the search works on the primary column which is configured as the symbol name. I can also limit the search to just `%kind:field` though. The buffer picker can filter by buffer ID, modification/focus flags and path. The diagnostics pickers can filter by severity, code, message, and path (for the workspace-wide diagnostics picker).

The workspace symbol picker can filter separately by symbol name, filename, and symbol kind. The symbol name is no longer passed to the fuzzy matcher which fixes some minor issues when using language-server-specific query syntaxes (see #5714). This paves the way for dynamic global search but I have left that off of this PR as the change will be large on its own.

Closes https://github.com/helix-editor/helix/pull/7040
Closes https://github.com/helix-editor/helix/issues/5714
Closes https://github.com/helix-editor/helix/issues/5446
Closes https://github.com/helix-editor/helix/pull/5110
Closes https://github.com/helix-editor/helix/pull/2452
Closes https://github.com/helix-editor/helix/issues/3543

There are a few TODOs:

* Allow quoting in the picker's field syntax
* Figure out if this needs theming changes (adding a theme key for the primary column?)
* Improve width constraint calculations so large columns share the space better

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -17,12 +17,13 @@ use tui::{
     buffer::Buffer as Surface,
     layout::Constraint,
     text::{Span, Spans},
-    widgets::{Block, BorderType, Borders, Cell, Table},
+    widgets::{Block, BorderType, Borders, Cell, Row, Table},
 };
 
 use tui::widgets::Widget;
 
 use std::{
+    borrow::Cow,
     collections::HashMap,
     io::Read,
     path::PathBuf,
@@ -47,7 +48,7 @@ use helix_view::{
 };
 
 pub const ID: &str = "picker";
-use super::{menu::Item, overlay::Overlay};
+use super::overlay::Overlay;
 
 pub const MIN_AREA_WIDTH_FOR_PREVIEW: u16 = 72;
 /// Biggest file size to preview in bytes
@@ -123,38 +124,36 @@ impl Preview<'_, '_> {
     }
 }
 
-fn item_to_nucleo<T: Item>(item: T, editor_data: &T::Data) -> Option<(T, Utf32String)> {
-    let row = item.format(editor_data);
-    let mut cells = row.cells.iter();
-    let mut text = String::with_capacity(row.cell_text().map(|cell| cell.len()).sum());
-    let cell = cells.next()?;
-    if let Some(cell) = cell.content.lines.first() {
-        for span in &cell.0 {
-            text.push_str(&span.content);
+fn inject_nucleo_item<T, D>(
+    injector: &nucleo::Injector<T>,
+    columns: &[Column<T, D>],
+    item: T,
+    editor_data: &D,
+) {
+    let column_texts: Vec<Utf32String> = columns
+        .iter()
+        .filter(|column| column.filter)
+        .map(|column| column.format_text(&item, editor_data).into())
+        .collect();
+    injector.push(item, |dst| {
+        for (i, text) in column_texts.into_iter().enumerate() {
+            dst[i] = text;
         }
-    }
-
-    for cell in cells {
-        text.push(' ');
-        if let Some(cell) = cell.content.lines.first() {
-            for span in &cell.0 {
-                text.push_str(&span.content);
-            }
-        }
-    }
-    Some((item, text.into()))
+    });
 }
 
-pub struct Injector<T: Item> {
+pub struct Injector<T, D> {
     dst: nucleo::Injector<T>,
-    editor_data: Arc<T::Data>,
+    columns: Arc<Vec<Column<T, D>>>,
+    editor_data: Arc<D>,
     shutown: Arc<AtomicBool>,
 }
 
-impl<T: Item> Clone for Injector<T> {
+impl<I, D> Clone for Injector<I, D> {
     fn clone(&self) -> Self {
         Injector {
             dst: self.dst.clone(),
+            columns: self.columns.clone(),
             editor_data: self.editor_data.clone(),
             shutown: self.shutown.clone(),
         }
@@ -163,21 +162,57 @@ impl<T: Item> Clone for Injector<T> {
 
 pub struct InjectorShutdown;
 
-impl<T: Item> Injector<T> {
+impl<T, D> Injector<T, D> {
     pub fn push(&self, item: T) -> Result<(), InjectorShutdown> {
         if self.shutown.load(atomic::Ordering::Relaxed) {
             return Err(InjectorShutdown);
         }
 
-        if let Some((item, matcher_text)) = item_to_nucleo(item, &self.editor_data) {
-            self.dst.push(item, |dst| dst[0] = matcher_text);
-        }
+        inject_nucleo_item(&self.dst, &self.columns, item, &self.editor_data);
         Ok(())
     }
 }
 
-pub struct Picker<T: Item> {
-    editor_data: Arc<T::Data>,
+type ColumnFormatFn<T, D> = for<'a> fn(&'a T, &'a D) -> Cell<'a>;
+
+pub struct Column<T, D> {
+    name: &'static str,
+    format: ColumnFormatFn<T, D>,
+    /// Whether the column should be passed to nucleo for matching and filtering.
+    /// `DynamicPicker` uses this so that the dynamic column (for example regex in
+    /// global search) is not used for filtering twice.
+    filter: bool,
+}
+
+impl<T, D> Column<T, D> {
+    pub fn new(name: &'static str, format: ColumnFormatFn<T, D>) -> Self {
+        Self {
+            name,
+            format,
+            filter: true,
+        }
+    }
+
+    pub fn without_filtering(mut self) -> Self {
+        self.filter = false;
+        self
+    }
+
+    fn format<'a>(&self, item: &'a T, data: &'a D) -> Cell<'a> {
+        (self.format)(item, data)
+    }
+
+    fn format_text<'a>(&self, item: &'a T, data: &'a D) -> Cow<'a, str> {
+        let text: String = self.format(item, data).content.into();
+        text.into()
+    }
+}
+
+pub struct Picker<T: 'static + Send + Sync, D: 'static> {
+    column_names: Vec<&'static str>,
+    columns: Arc<Vec<Column<T, D>>>,
+    primary_column: usize,
+    editor_data: Arc<D>,
     shutdown: Arc<AtomicBool>,
     matcher: Nucleo<T>,
 
@@ -186,7 +221,7 @@ pub struct Picker<T: Item> {
 
     cursor: u32,
     prompt: Prompt,
-    previous_pattern: String,
+    query: HashMap<&'static str, String>,
 
     /// Whether to show the preview panel (default true)
     show_preview: bool,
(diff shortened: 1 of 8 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -124,375 +117,147 @@ impl Preview<'_, '_> {
...
+    // * https://github.com/helix-editor/helix/issues/5714
+    // * https://github.com/helix-editor/helix/pull/4687#issuecomment-1364586418
+    // Then Picker would look for the first column with 'filter: true' and use
+    // that for filtering and same for sorting.
+    sort: bool,
+    filter: bool,
+    format_fn: Box<dyn Fn(&T) -> Cell>,
+    sort_text_fn: Option<Box<dyn Fn(&T) -> Cow<str>>>,
+    filter_text_fn: Option<Box<dyn Fn(&T) -> Cow<str>>>,
+}
```
the-mikedavis, line 138: Column is currently implemented as a struct since it's a nice API to write out your set of columns when creating a Picker: https://github.com/helix-editor/helix/blob/b06482654f59c989ffff53cb8ebebd865d31679c/helix-term/src/commands/lsp.rs#L255-L279

`Item::Data` can be eliminated since it can be captured in these closures.

I think it should be possible to re-implement Column as a trait similar to the current `ui::menu::Item`: https://github.com/helix-editor/helix/blob/78e869542022638af482c0dd85a07086a942907a/helix-term/src/ui/menu.rs#L17-L32

```rust
pub trait Column<T> {
    type Data;
    const NAME: &'static str;
    const FILTER: bool;
    const SORT: bool;

    fn format<'a>(&self, item: &'a T, data: &Self::Data) -> Cell<'a>;

    fn sort_text<'a>(&self, item: &'a T, data: &Self::Data) -> Cow<'a, str> {
        let text: String = self.format(data).contents.into();
        text.into()
    }

    fn filter_text<'a>(&self, item: &'a T, data: &Self::Data) -> Cow<'a, str> {
        let text: String = self.format(data).contents.into();
        text.into()
    }
}
```

But it will be more verbose to define columns and create Pickers (`Picker::new` will still need to take `editor_data`).

I don't have a good intuition about the penalty of dynamic dispatch. Is it worth the trait for static dispatch? These functions may be called very often (per-keypress, per-option).

pascalkuthe:
Oh wow this is awesome, I wanted to implement this sometime too but never got around to it. I think this has a ton of potential 👍

helix-term/src/ui/picker.rs
the-mikedavis: I'll switch to the trait version. Not only do we pay the dynamic dispatch penalty but we need to allocate in more situations because of a limitation of closures: you can't return references of captured variables. So we end up allocating `Cow::Owned` in some places where we previously got away with `Cow::Borrowed`s. The trait version will also allow us to re-use some column definitions like we currently do with the `menu::Item` implementation for `PathBuf`.

the-mikedavis:
Separate prompts for each field feels a little clunky. Instead I think we should have only one prompt, drop the "active column" concept, and use a [lucene](https://lucene.apache.org/core/2_9_4/queryparsersyntax.html)-like syntax for filtering by fields. For example the symbol picker could take some queries:

* `func mysym`: fuzzy-match by "func mysym" across all fields
* `kind:func mysym`: fuzzy-match by "func" in the "type" field and "mysym" in the remaining fields (symbol name)
* `func name:mysym`: fuzzy-match by "mysym" in the name field and "func" in the remaining fields (symbol kind)

(The ordering in these examples shouldn't matter.) Values with spaces could be wrapped in quotes, for example `kind:"enum member"`. The matches would have an implicit AND relationship but we could support `AND`/`OR` as well if necessary.

Other examples:

* `flags:+` in the buffer picker would filter down to modified files
* `severity:error code:34 path:src/ my error message` could be used in the diagnostics picker

pascalkuthe:
I like the general idea of having a single field. The reason I suggested split columns before is that:

* Not all columns may fit on screen at once
* potential false positives (solved by the explicit `kind:` syntax but makes matching for multiple words in the same column (which I do quite a lot) without false positives quite awkward For example: `file:syntax.rs file:helix-core file:src` instead of syntax.rs helix-core src` on master. 
* Most importantly: How does this work with dynamic pickers. Fuzzy matching across all columns doesn't work for regex queries
* one Small thing `:` is too common to require escaping we should use a less common char

My idea to resolve these is the following:

* `%kind` will treat all following words until the next word starting with `%` (or EOL) as queries for the `kind` column. While typing in such an area that column is called `active`
* It the last column name is unknown (or empty) all (non-dynamic} columns are active at once  (like you said fuzzy match across everything)
* When there is not enough space to display all columns, priority is given to the active column
* Since dynamic columns always need to be manually activated we can prefill the picker when it opens to make that more convenient. For example for interactive global search, we prefill `%query foo+bar %` (typing immediately fuzzy matches all columns which would just be path in this case while keeping the qury from the prompt ). For the workspace picker, we just prefill `%query ` and the user can start filtering immediately.

A distant future idea: we could create typable command versions of the pickers that accept a prefill as an argument. This would allow binding: open function picker to a key without adding a dedicated command

the-mikedavis:
For dynamic pickers, I have some local changes that allow disabling filtering/sorting on a column. That will resolve #5714 and allow entering regexes for the global search dyn picker. I also have a change locally that allows specifying an input register for global search to set an initial query.

For prefilling pickers, I think specifying a register would be a nice way to do it. I have some local changes building on #7277 that allow a sequence like `"/<space>/` to open global search using the `/` register contents as the initial query. That could be done for other pickers as well.

I like the active column idea you describe for a priority in rendering. For matches with long lines the global search picker can get very cramped. I think there's room for improvement with truncating columns so that outlier long matches don't push the other columns out, and columns share the available space more evenly.

I'd like to start with a proof-of-concept using the `:` syntax but try it both ways. I think `:` might be sufficient: we may be able to make smart decisions while parsing to ignore `:`s not used in fields. And about the `file:syntax.rs file:helix-core file:src` example, I think that could be the behavior of `file:"syntax.rs helix-core src"`.

pascalkuthe:
> For dynamic pickers, I have some local changes that allow disabling filtering/sorting on a column. That will resolve #5714 and allow entering regexes for the global search dyn picker. I also have a change locally that allows specifying an input register for global search to set an initial query.
> 
> For prefilling pickers, I think specifying a register would be a nice way to do it. I have some local changes building on #7277 that allow a sequence like `"/<space>/` to open global search using the `/` register contents as the initial query. That could be done for other pickers as well.
> 
> I like the active column idea you describe for a priority in rendering. For matches with long lines the global search picker can get very cramped. I think there's room for improvement with truncating columns so that outlier long matches don't push the other columns out, and columns share the available space more evenly.
> 
> I'd like to start with a proof-of-concept using the `:` syntax but try it both ways. I think `:` might be sufficient: we may be able to make smart decisions while parsing to ignore `:`s not used in fields. And about the `file:syntax.rs file:helix-core file:src` example, I think that could be the behavior of `file:"syntax.rs helix-core src"`.

What I mean about the dynamic picker is that when you enter a string without specifying a column it wouldn't work well to fuzzy match that across all column (I would iamgine that works by ocmbining all colums into a single string that is then fuzzyatched on so you can't also modify the query that way). I think for dynamic pickers you always have to specify the column. Hence the idea to have an easy syntax for specifying queries with spaces. I think something more convenient that quotes might be nice for that but we can always bikeshed that later.

Yeah for prototyping the syntax probably doesn't matter that much 👍 It shoukd be easy to change later if it feels clunky while testing. If we use column: we could indeed make that only active if such a column actually exists (removing.the need for.escape.in most cases). That could.be a nice variant too.

Using a register for prefill is also an interesting idea. Do we have a simple way to set a register from a key binding (like :set-register?)` I personally never use that kind of functionality so I always forget :D Otherwise that might be useful for creating bindings like I described

the-mikedavis:
Ah yeah I see what you mean now. For the time being I have made the behavior different in dynamic pickers and regular pickers: in dynamic pickers, the "common" text (text without a field specified) automatically acts as the dynamic column's input any other fields need to be specified explicitly. It feels intuitive but I'm not sure I like having split behaviors.

jhugs:
When you get a chance to rebase again I'd love to test this out

the-mikedavis:
I rebased this on top of master with the nucleo changes. Kudos to @pascalkuthe: the column support in nucleo makes the changes to the fuzzy matching parts very clean! 🎉

I have reduced the scope of this PR a little so that dynamic global search is left for the future. `global_search` now includes the matched line's contents but isn't a DynamicPicker. That change can be done after this PR and I'd like to base it on #8021 so that we have better control over DynamicPicker's debounce behavior. (IME running this branch we want a longer debounce for global search.)

There are a few more changes I'd like to make to this branch but I think it's nearly ready for review.

matuusu:
Are there any plans to move this PR forward?

the-mikedavis:
Yep, I keep this branch up to date locally and build on it for other changes (see https://github.com/helix-editor/helix/issues/196#issuecomment-1712931640). I'm waiting on #8021 for some of the changes to global search but the table refactor could be done independently. I'll rebase this branch and clean up the commits when I get a chance.

There are two small changes this branch needs too:

* The table header needs to be accounted for when paging up/down in the results - currently we ignore the header so we're off by a few lines when switching pages
* @pascalkuthe mentioned a good idea that you should be able to use prefixes of column names and have that select the correct column, so you could say `%p:syntax.rs` instead of `%path:syntax.rs` for example

ItsEthra:
> @pascalkuthe mentioned a good idea that you should be able to use prefixes of column names and have that select the correct column, so you could say %p:syntax.rs instead of %path:syntax.rs for example

 I have a branch that implements it, should I make a pr to this branch for that? I made it so if multiple columns match `%prefix`, then it selects the columns with the shortest name(because it would fit prefix the most) or the first column of all matches.

the-mikedavis:
I will probably recreate this branch from scratch rather than deal with the conflicts. If you link your branch I will pick up the commit from it when I revisit this

ItsEthra:
My branch is [here](https://github.com/ItsEthra/helix/tree/table-picker-prefixes), I also rebased it on master, there was just one conflict and it seems alright.

helix-term/src/ui/picker.rs
```diff
@@ -946,21 +1071,98 @@ impl<T: Item> Drop for Picker<T> {
...
+                in_field = ch == '%';
+            }
+            '%' => in_field = true,
+            ' ' => (),
+            ':' if in_field => {
+                // Go over all columns and their indices, find all that starts with field key,
+                // select a column that fits key the most.
+                field = column_names
+                    .iter()
+                    .filter(|col| col.starts_with(&text))
```
the-mikedavis, line 1133: @pascalkuthe what do you think about using fuzzy matching for this instead? It would be more forgiving, so you could use `%pth:syntax.rs` for `path` for example. We could even highlight the parts of the column name that matched, like we do for the parts of the query that match items

the-mikedavis:
Closing in favor of https://github.com/helix-editor/helix/pull/9647 which includes these commits

helix-term/src/ui/picker.rs
the-mikedavis: We discussed this and fuzzy matching seems like overkill - the prefix is more straightforward. Highlighting the matching parts could also be done for prefix strings as well and would be simpler. (Could be a follow-up to #9647)

PR #7269, opened 2023-06-07
Auto pair deletions with selections

dead10ck:
This PR completes auto pair deletions. Currently, pairs are only deleted on ranges that are a single grapheme in width, since the normal change mapping process results in an incorrect selection.

This is achieved with two new functions:

* `change_by_and_with _selection`
* `delete_by_and_with_selection`

They are helpers which take a transformation function that produces individual changes, one range at a time, and can optionally produce a new range that replaces the input range that it took as an argument. When one is not given, the input range is mapped through the change set as normal. This allows the caller to apply changes to the text and incrementally build a new selection as they go, picking and choosing in which cases new ranges are computed explicitly, and in which they are mapped normally.

Concretely, this solves the problem of needing auto pair deletes to work concurrently with dedenting on backspace; the auto pair hooks were changed to operate on individual ranges and return individual changes, rather than producing a whole complete transaction. Then deleting was enabled to compose changes from all 3 possible scenarios at once, i.e. deleting a single grapheme, dedenting, and auto pair deletions.

[asciicast](https://asciinema.org/a/HlcFfnoFFIpVOnzvtts2Y9OXX)

Additionally, this PR inserts an extra whitespace when the cursor is inside a pair to pad the left and right.

[asciicast](https://asciinema.org/a/iPRzxTo4bLE3acaD5uFIQZo0N)


dead10ck:
This was rebased. I think the new tests had been failing since the integration test refactor, but I fixed those too.

daedroza:
Would it be possible to rebase this branch on new HEAD?

PR #7282, opened 2023-06-07
misc doc fixes/improvements

tshepang:


the-mikedavis (changes requested):
The changes to the markdown files are fine but the changes to comments in src/ should be dropped: they aren't very significant and will cause conflicts for any open PRs that touch that code.

PR #7286, opened 2023-06-07
Bail if no language servers support workspace symbols

the-mikedavis:
The behavior prior to this commit opened the picker and then showed the error. We can check the feature availability before performing the first query to prevent opening up the picker unnecessarily.


pascalkuthe (approved):
Just because I was confused at first and have this written down somewhere: This is only necessary for the workspace symbol picker is a dynamic picker (the normal symbol picker isn't). Thanks!

PR #7290, opened 2023-06-08
Webc language support

vanarok:
Add [webc](https://github.com/11ty/webc) grammar.

### DIFF
--- a/runtime/queries/webc/highlights.scm
+++ b/runtime/queries/webc/highlights.scm
@@ -1,0 +1,1 @@
+; inherits: html
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -162,6 +162,7 @@
 | vue | ✓ |  |  | `vls` |
 | wast | ✓ |  |  |  |
 | wat | ✓ |  |  |  |
+| webc | ✓ |  |  |  |
 | wgsl | ✓ |  |  | `wgsl_analyzer` |
 | wit | ✓ |  | ✓ |  |
 | xit | ✓ |  |  |  |
--- a/languages.toml
+++ b/languages.toml
@@ -2618,3 +2618,12 @@ indent = { tab-width = 2, unit = "  " }
 [[grammar]]
 name = "t32"
 source = { git = "https://codeberg.org/xasc/tree-sitter-t32", rev = "1dd98248b01e4a3933c1b85b58bab0875e2ba437" }
+
+[[language]]
+name = "webc"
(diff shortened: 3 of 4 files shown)


runtime/queries/webc/highlights.scm
```diff
@@ -0,0 +1,19 @@
+(tag_name) @tag
+(erroneous_end_tag_name) @tag.error
```
the-mikedavis, line 2:
```suggestion
(erroneous_end_tag_name) @error
```

runtime/queries/webc/highlights.scm
the-mikedavis, line 1: Are these copies of the ones in `runtime/queries/html`? If so you can include them without duplicating the queries with

```
; inherits: html
```

vanarok: Yep, it is.

the-mikedavis:
This will need a run of `cargo xtask docgen` to fix the CI

pascalkuthe (approved):

johannesrave:
very cool! is there a way for me to use this now without waiting for the next release if i can't rabbithole into getting the buildchain up atm?

johannesrave:
thanks for working on this @vanarok ! i have by now figured out how to compile helix from master, and now the new version is released for packaging too.

i have another question regarding the actual grammar though:
auto-formatting breaks indented frontmatter at the moment, for example with indented arrays:

```
---
layout: "_layouts/document.webc"
title: recent posts
pagination:
# these lines get un-indented on :fmt
  data: collections.post 
  size: 2
  alias: posts
  reverse: true
---
<p>welcome to the rest of the page.</p>
```
looking at it with `:tree-sitter-subtree` the frontmatter is simply parsed as a string by the html-grammar.

i expected that the `injections.scm` in `queries/webc` somehow extends the `html` grammar-injections that `webc` uses as defined in the global `languages.toml`, and that i would be able to copy the yaml injections from the markdown-grammar:

```
; inherits: html
((minus_metadata) @injection.content (#set! injection.language "yaml") (#set! injection.include-unnamed-children))
```

 however, i see no effect after recompiling or rebuilding the grammars on their own. is `queries/webc` even used if the `languages.toml` defines the grammar for `webc` to be `html`?  

i tried working on this but have been unsuccessful, i just don't know enough about the whole build-chain.
i don't fully understand the interplay of the tree-sitter parsers that get pulled on build, the language-definitions and the `queries`-directories (with injections.scm).

Issue #6717, opened 2023-04-12
Hint when execute run shell command inserting


Renkai:
When we use `:sh`, we can get hint like directory and file names, but when we execute `!`(Run shell command, inserting output before each selection), we don't have this feature, can we let it has this feature as also?

the-mikedavis:
The shell commands open a prompt with no completion: https://github.com/helix-editor/helix/blob/d5fec302c9fc6a6e334a0d518c6d10f7dbeb26a7/helix-term/src/commands.rs#L5197-L5202

That could be changed to the filename completer to match `:run-shell-command`: https://github.com/helix-editor/helix/blob/d5fec302c9fc6a6e334a0d518c6d10f7dbeb26a7/helix-term/src/commands/typed.rs#L2707-L2713

PR #7295, opened 2023-06-08
Add filename completion in shell_prompt

infamous55:
This pull request implements the changes mentioned in issue #6717.

Please let me know if I missed something!


infamous55:
I've noticed that the completion runs only until the first space, meaning my changes are incomplete. I'll look into this next!

infamous55:
I've read a bit through helix-term, and I'm definitely missing something about what triggers the completion of every word in other functions, such as `open` or `run_shell_command`, and how I could apply that in my particular case. I should also mention that this is my first attempt to contribute to the project. I'm open to suggestions or instructions from someone more knowledgeable until I figure this out myself.

infamous55:
It has been quite some time, and I haven't been able to dedicate enough time to delve into the problem mentioned above. Consequently, I have decided to close the pull request and allow someone else to take on the task.

PR #7296, opened 2023-06-09
use asdf

windowsrefund:


kirawi:
This doesn't really belong in in the readme. It would go into https://github.com/helix-editor/helix/blob/master/book/src/install.md

pascalkuthe:
And the o ly official installation method is the github release page. The documentation should not suggest that another installation method is preferable

PR #7310, opened 2023-06-10
Consistent statusline spacing

hamrik:
Statusline elements used to have inconsistent spacing due to each element having to pad itself without knowledge of surrounding elements.

Inconsistent spacing

 - `mode`, `selections`, `file-name` and `position` are padded, the rest are not
 - `file-modification-indicator` reserves space for itself, `spinner` and `diagnostics` do not

This PR fixes these spacing inconsistencies.

Consistent spacing

 - No elements pad themselves
 - No elements reserve empty space
 - The statusline pads everything
 - The `spacer` element has been repurposed to join multiple spacers, e.g. to pad the ends of the statusline

The render functions of statusline elements previously used an ad-hoc `write` callback to blit `Spans` directly to the `Surface` of the statusline. This callback has been removed. Now each render function has to return the `Spans` it wants to display, which the renderer of the statusline collects, intersperses with spaces, then blits to the surface.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -92,9 +92,9 @@ The `[editor.statusline]` key takes the following sub-keys:
 
 | Key           | Description | Default |
 | ---           | ---         | ---     |
-| `left`        | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name", "read-only-indicator", "file-modification-indicator"]` |
+| `left`        | A list of elements aligned to the left of the statusline | `["spacer", "mode", "spinner", "file-name", "read-only-indicator", "file-modification-indicator"]` |
 | `center`      | A list of elements aligned to the middle of the statusline | `[]` |
-| `right`       | A list of elements aligned to the right of the statusline | `["diagnostics", "selections", "register", "position", "file-encoding"]` |
+| `right`       | A list of elements aligned to the right of the statusline | `["diagnostics", "selections", "register", "position", "file-encoding", "spacer"]` |
 | `separator`   | The character used to separate elements in the statusline | `"│"` |
 | `mode.normal` | The text shown in the `mode` element for normal mode | `"NOR"` |
 | `mode.insert` | The text shown in the `mode` element for insert mode | `"INS"` |
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -60,7 +60,7 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `undercurl` | Set to `true` to override automatic detection of terminal undercurl support in the event of a false negative | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file | `[]` |
 | `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
-| `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
+| `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `true` |
 | `text-width` | Maximum line length. Used for the `:reflow` command and soft-wrapping if `soft-wrap.wrap-at-text-width` is set | `80` |
 | `workspace-lsp-roots` | Directories relative to the workspace root that are treated as LSP roots. Should only be set in `.helix/config.toml` | `[]` |
 | `default-line-ending` | The line ending to use for new documents. Can be `native`, `lf`, `crlf`, `ff`, `cr` or `nel`. `native` uses the platform's native line ending (`crlf` on Windows, otherwise `lf`). | `native` |
@@ -92,13 +92,13 @@ The `[editor.statusline]` key takes the following sub-keys:
 
 | Key           | Description | Default |
 | ---           | ---         | ---     |
-| `left`        | A list of elements aligned to the left of the statusline | `["spacer", "mode", "spinner", "file-name", "read-only-indicator", "file-modification-indicator"]` |
+| `left`        | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name", "read-only-indicator", "file-modification-indicator"]` |
 | `center`      | A list of elements aligned to the middle of the statusline | `[]` |
 | `right`       | A list of elements aligned to the right of the statusline | `["diagnostics", "selections", "register", "position", "file-encoding", "spacer"]` |
 | `separator`   | The character used to separate elements in the statusline | `"│"` |
-| `mode.normal` | The text shown in the `mode` element for normal mode | `"NOR"` |
-| `mode.insert` | The text shown in the `mode` element for insert mode | `"INS"` |
-| `mode.select` | The text shown in the `mode` element for select mode | `"SEL"` |
+| `mode.normal` | The text shown in the `mode` element for normal mode | `" NOR "` |
+| `mode.insert` | The text shown in the `mode` element for insert mode | `" INS "` |
+| `mode.select` | The text shown in the `mode` element for select mode | `" SEL "` |
 
 The following statusline elements can be configured:
 
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -4,7 +4,6 @@ use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::{
     document::{Mode, SCRATCH_BUFFER_NAME},
     graphics::Rect,
-    theme::Style,
     Document, Editor, View,
 };
 
@@ -58,25 +57,17 @@ pub fn render(context: &mut RenderContext, viewport: Rect, surface: &mut Surface
 
     surface.set_style(viewport.with_height(1), base_style);
 
-    let write_left = |context: &mut RenderContext, text, style| {
-        append(&mut context.parts.left, text, &base_style, style)
-    };
-    let write_center = |context: &mut RenderContext, text, style| {
-        append(&mut context.parts.center, text, &base_style, style)
-    };
-    let write_right = |context: &mut RenderContext, text, style| {
-        append(&mut context.parts.right, text, &base_style, style)
-    };
-
     // Left side of the status line.
 
     let config = context.editor.config();
 
     let element_ids = &config.statusline.left;
-    element_ids
+    context.parts.left = element_ids
         .iter()
         .map(|element_id| get_render_function(*element_id))
-        .for_each(|render| render(context, write_left));
+        .flat_map(|render| render(context).0)
+        .collect::<Vec<Span>>()
+        .into();
 
     surface.set_spans(
         viewport.x,
@@ -88,10 +79,12 @@ pub fn render(context: &mut RenderContext, viewport: Rect, surface: &mut Surface
     // Right side of the status line.
 
     let element_ids = &config.statusline.right;
-    element_ids
+    context.parts.right = element_ids
         .iter()
         .map(|element_id| get_render_function(*element_id))
-        .for_each(|render| render(context, write_right));
+        .flat_map(|render| render(context).0)
+        .collect::<Vec<Span>>()
+        .into();
 
     surface.set_spans(
         viewport.x
@@ -106,10 +99,12 @@ pub fn render(context: &mut RenderContext, viewport: Rect, surface: &mut Surface
     // Center of the status line.
 
     let element_ids = &config.statusline.center;
-    element_ids
+    context.parts.center = element_ids
         .iter()
         .map(|element_id| get_render_function(*element_id))
-        .for_each(|render| render(context, write_center));
+        .flat_map(|render| render(context).0)
+        .collect::<Vec<Span>>()
+        .into();
 
     // Width of the empty space between the left and center area and between the center and right area.
     let spacing = 1u16;
@@ -126,17 +121,9 @@ pub fn render(context: &mut RenderContext, viewport: Rect, surface: &mut Surface
     );
 }
 
-fn append(buffer: &mut Spans, text: String, base_style: &Style, style: Option<Style>) {
-    buffer.0.push(Span::styled(
-        text,
-        style.map_or(*base_style, |s| (*base_style).patch(s)),
-    ));
-}
-
-fn get_render_function<F>(element_id: StatusLineElementID) -> impl Fn(&mut RenderContext, F)
-where
-    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
-{
+fn get_render_function<'a>(
+    element_id: StatusLineElementID,
+) -> impl Fn(&RenderContext) -> Spans<'a> {
     match element_id {
         helix_view::editor::StatusLineElement::Mode => render_mode,
         helix_view::editor::StatusLineElement::Spinner => render_lsp_spinner,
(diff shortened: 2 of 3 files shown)


gabydd:
At first look one thing that this changes that we probably shouldn't is it removes padding around the edges in the default config which imo we should keep especially looks bad when ``color-modes`` is false. This brings up another question if we are going for consistency do we really want ``mode``s size to be dependant on ``coior-modes``? I haven't looked much at the implementation but I think improving consistency is generally good as long as we are not breaking peoples configs.

David-Else:
@gabydd Regarding `color -modes`: https://github.com/helix-editor/helix/discussions/7289#discussioncomment-6124763

gabydd:
Ah right, I only noticed this because I don't use color-modes. Removing the option is probably a good idea in the long run though.

helix-tui/src/text.rs
```diff
@@ -229,6 +229,70 @@ impl<'a> Spans<'a> {
...
+    }
+
+    /// Constructs a single, styled `Span` with the given `string` and `style`
+    /// and wraps it in a `Spans`.
+    pub fn single<T>(string: T, style: Style) -> Spans<'a>
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        Spans(vec![Span::styled(string, style)])
+    }
```
the-mikedavis, line 264: This already exists as a `From`: https://github.com/helix-editor/helix/blob/2a11fb485d33eb33a7743899e9a86f745f6d5e02/helix-tui/src/text.rs#L258-L262

Places where this function is used can either use `Span::styled(string, style).into()` or explicitly call `Spans::from` on that Span.

helix-tui/src/text.rs
```diff
@@ -229,6 +229,70 @@ impl<'a> Spans<'a> {
...
+    /// ## Examples
+    ///
+    /// ```rust
+    /// # use helix_tui::text::{Span, Spans};
+    /// # use helix_view::graphics::{Color, Style};
+    /// let mut spans = Spans::new();
+    /// spans.push("Hello!", Style::default());
+    /// assert_eq!("Hello!", spans.0[0].content);
+    /// ```
+    pub fn push<T>(&mut self, string: T, style: Style)
```
the-mikedavis, line 245: This should take a `Span<'a>` instead

helix-tui/src/text.rs
```diff
@@ -229,6 +229,70 @@ impl<'a> Spans<'a> {
...
+    /// let spans = Spans::join(
+    ///   vec![
+    ///     Spans::single("Hello", Style::default()),
+    ///     Spans::single("World!", Style::default()),
+    ///   ].into_iter(),
+    ///   &Spans::single(" ", Style::default()),
+    /// );
+    /// assert_eq!("Hello World!", String::from(spans));
+    /// ```
+    pub fn join<S>(mut spanses: S, sep: &Spans<'a>) -> Spans<'a>
```
the-mikedavis, line 282:
```suggestion
    pub fn join<S>(mut spans_iter: S, sep: &Spans<'a>) -> Spans<'a>
```

the-mikedavis: Does `sep` need to be a `Spans` here? I think it could just be a `Span`.

helix-tui/src/text.rs
```diff
@@ -229,6 +229,70 @@ impl<'a> Spans<'a> {
...
+        S: Iterator<Item = Spans<'a>>,
+    {
+        let mut spans: Vec<Span> = Vec::new();
+        if let Some(head) = spanses.next() {
+            spans.extend(head.0);
+        }
+        spanses.for_each(|s| {
+            spans.extend(sep.0.clone());
+            spans.extend(s.0);
+        });
```
the-mikedavis, line 293: It's a style nit but `for_each` is generally considered less idiomatic in Rust than regular `for` or `while` loops

```rust
while let Some(spans) = span_iter.next() {
    if !new_spans.is_empty() {
        new_spans.push(sep.clone());
    }
    new_spans.extend(spans);
}
```

(see the note in the rust docs: https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.for_each)

book/src/configuration.md
```diff
@@ -116,7 +118,7 @@ The following statusline elements can be configured:
 | `position` | The cursor position |
 | `position-percentage` | The cursor position as a percentage of the total number of lines |
 | `separator` | The string defined in `editor.statusline.separator` (defaults to `"│"`) |
-| `spacer` | Inserts a space between elements (multiple/contiguous spacers may be specified) |
+| `spacer` | An empty string, used to join multiple spacers |
```
the-mikedavis, line 121:
```suggestion
| `spacer` | A string used as space to separate all statusline elements, defined in `editor.statusline.spacer` (defaults to `" "`) |
```

pascalkuthe:
I am not sure why we need the `editor.statusline.spacer` config. Why would this be ever set to anything but space?

hamrik:
Sorry for disappearing, had a busy couple of weeks. Checking the responses right now.

hamrik:
> At first look one thing that this changes that we probably shouldn't is it removes padding around the edges in the default config which imo we should keep especially looks bad when `color-modes` is false. This brings up another question if we are going for consistency do we really want `mode`s size to be dependant on `coior-modes`? I haven't looked much at the implementation but I think improving consistency is generally good as long as we are not breaking peoples configs.

I originally didn't have padding in `color-modes` either, but then a color cuts right off at the end of the mode name, which didn't look right. I agree that this makes `modes` a bit inconsistent and will have to think of something to address this. 

Padding the statusbar edges could be done by changing the default config to include `spacer`-s on the front or back, or alternatively I could for a unit of padding into the statusline. With the latter option `color-modes` won't look right and the user wouldn't be able to fix it by removing the `spacer`, so I vote for chaging the default config. Those who have customized statuslines shouldn't have a hard time adding a spacer if they don't use `color-modes`.

helix-term/src/ui/statusline.rs
```diff
@@ -164,162 +155,130 @@ where
...
-        context,
-        language_server
-            .and_then(|srv| {
-                context
-                    .spinners
-                    .get(srv.id())
-                    .and_then(|spinner| spinner.frame())
-            })
-            // Even if there's no spinner; reserve its space to avoid elements frequently shifting.
-            .unwrap_or(" ")
```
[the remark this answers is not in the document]
hamrik: But if the behavior is kept then there will be a large empty space between the elements surrounding the spinner when the spinner is not in use.

I guess either situation could be solved by simply placing the spinner as the last item of the left side or the first item of the right side, blending it in with the surrounding spacing. Or we could change the spinner idle frame to always display a faded out placeholder when not in use, e.g. `⣿`.

helix-tui/src/text.rs
```diff
@@ -229,6 +229,70 @@ impl<'a> Spans<'a> {
...
+    where
+        T: Into<Cow<'a, str>>,
+    {
+        self.0.push(Span::styled(string, style));
+    }
+
+    /// Creates a new, empty `Spans`.
+    pub fn new() -> Spans<'a> {
+        Spans(vec![])
+    }
```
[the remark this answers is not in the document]
hamrik: Didn't know, will change. Thanks!

helix-tui/src/text.rs
hamrik: Didn't know, will change. Thanks!

helix-tui/src/text.rs
hamrik: It makes sense, but that will bloat the current usages a bit. I guess I could always add a helper function there instead. Will change and see if a utility function becomes necessary in `statusline.rs`.

helix-tui/src/text.rs
hamrik: Using spans allows for fancier separators, for example one with multiple colors.

I think I will make two more versions of `join`, one taking a `Span` as you suggest, and one taking an interpolator function which takes two neighboring `Spans` and generates a separator for them. This could be useful for powerline-style color transitions.

helix-tui/src/text.rs
hamrik: Will change, thanks!

hamrik:
> I am not sure if we need the `editor.statusline.spacer` config. Why would this be ever set to anything but space?

A couple examples:
  - Two or more spacers for wider spacing without spamming `spacer`: `file.txt   1:1   [+]`
  - Powerline-style (colorless) separators: `file.txt 〉1:1 〉[+]`
    - Altough to make this last one flippable we would have to add separate spacer configs for left, center and right.

helix-tui/src/text.rs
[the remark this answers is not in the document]
hamrik: Fixed in 560fd9aa

helix-tui/src/text.rs
hamrik: Fixed in 560fd9aa

helix-tui/src/text.rs
hamrik: Fixed in 8e8d860b

helix-tui/src/text.rs
hamrik: Variable naming fixed in 8e8d860b

helix-tui/src/text.rs
hamrik: Fixed in 8e8d860b. Boilerplate is minimal, no extra helper added.

pascalkuthe:
> 

Generally I would prefer explicit spacers since it allows different spacing between different elements without adding a ton new config options. I don't think that the spacing inside of elements needs to be configurable. We are shooting for a mininal config file.

pascalkuthe:
> You are free to leave `editor.statusline.spacer` on the default setting and use spacers. It don't feel it's out of place next to the configurable separator and configurable mode names.

we are very conservative with regards to adding new config options and this doens' cross the bar for me. The rest of the PR is probably a nice improvement but I don't see us adding this config option

hamrik:
Alright, I've removed the option to keep things moving along. At least that resolved the documentation wording problem. We can revisit this option in a later PR.

book/src/configuration.md
hamrik: The option has been removed, so this is no longer an issue.

hamrik:
I believe the only two remaining issues are the spacing of the mode indicator and the spinner. 

I don't like the idea of hardcoding spacers on the edges, in my opinion modifying the default config to include spacers would be more flexible.

But regardless of how the padding is handled, it would look weird the color modes option turned on.

If we go with the hardcoded route, we could omit adding spacers where the mode indicator is the first/last item and color modes are enabled. But this feels like a hack.

If we go with the modified default config, we could add/omit the spacer depending on whether we enable color modes by default, then expect people to add/remove the spacer manually if they change the color modes option.

On a similar note, right now padding is automatically added to the mode indicator when color-modes are on, but that could also be removed if we expect people to modify the mode names to include spaces when they toggle color modes.

Thoughts?

hamrik:
I have removed the auto-padding from the `mode` element, regardless of `color-modes`.

 - This makes the `mode` element consistent with the other elements.
 - If `color-modes` is off by default, people who turn it on can add padding via `editor.statusline.mode`.
 - If `color-modes` becomed default, the default values for mode names can also be tweaked to include spaces.

pascalkuthe:
Yeah I agree with that. I think hardcoded padding is never a good idea and changing the modenames is a pretty good workaround IMO

book/src/configuration.md
pascalkuthe: I think I prefer the original docs here. The new docs aren't quite accurate and since the behavior hasn't been changed I see no reason to change the docs either

gabydd: Yeah even if this is what it technically does code wise, the docs are user facing and should explain what is actually important to user which is it adds another space

pascalkuthe (commented):
This PR is very hard to review right now because you changed the order of some.functions (and in general have included some unnecessary changes) sk the diff is pretty clobbered right now.

Could you undo that so it's easier for us to see what's actually changed

pascalkuthe:
Also I am not really sure if we really want the implhct spacer between each element. I actually lprefer having no implicit spacer at all and instead including the necessary specaers in the default config. This offers the most flexibility and is the easiet to implement

hamrik:
> I actually lprefer having no implicit spacer at all and instead including the necessary specaers in the default config.

Since some elements only appear occasionally (such as `file-modification-indicator`, `spinner`, etc) these would leave two spacers behind whenever they are hidden (assuming the config has explicit spacers between every element). Implementing collapsing spacers would be even more complicated IMO.

hamrik:
> This PR is very hard to review right now because you changed the order of some.functions (and in general have included some unnecessary changes) sk the diff is pretty clobbered right now.

I'll clean up the commit history on the weekend.

hamrik:
Okay, I've removed all additions to `Spans`, and consolidated the changes into two commits/"phases":

1) Refactor statusline elements to return `Spans` rather than using a callback. At this point the spacing behavior was identical to the previous implementation.
2) Removed all padding.

Right now there is no spacing interspersed as `Iterator::intersperse()` is in nightly for now. While this matches what @pascalkuthe requested I think spaces should be added because manually interspersed spacers don't collapse when an element hides itself, such as `file-modification-indicator`, leaving two spacers next to each other, which looks inconsistent.

Will address that later.

helix-term/src/ui/statusline.rs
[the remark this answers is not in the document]
twharmon: Items already jump around for other reasons, right? Almost very single element that can go in the status line can cause shifts (diagnostics, selections, position, etc.).

I simply put the elements that cause the most shift or most frequent shift (diagnostics) at the end of the list.

hamrik:
Added back automatic interspersing of spaces. Does not pad the sides of the panel, those should be done in config.

Uses a custom interspersing iterator until `instersperse()` lands in stable Rust, and I'm fairly certain that its implementation should be moved elsewhere, not yet sure where would be best.

As for the spinner, I think simply hiding it would be most consistent with other autohiding elements like `file-modification-indicator`. If the jumpyness is annoying it can be moved to a place where it does not push other elements around, such as the inner edge of the left and right sections.

helix-tui/src/text.rs
hamrik: Removed all custom Spans functions so this issue is now resolved.

book/src/configuration.md
hamrik: Ended up leaving the original description.

hamrik:
The only thing remaining to tweak in my opinion is the default config. After that I'll reopen the PR.

On the left section:
- If `color-modes` becomes the default then the default mode strings should be padded with strings.
- If it does not than an extra spacer needs to be added to the front of the default element list.

On the right section:
- A spacer needs to be appended to the default element list.

hamrik:
Added spacers the default config to pad the statusline, with the assumtion that color modes will remain off by default for now.

People who want to enable it and have it look good will have to:
 1. Enable `color-modes`
 2. Set custom mode strings that include spaces, e.g. `normal = " NOR "`
 3. Set a custom `left` section that does not begin with `spacer`

Also fixed the spacer that added an extra space by accident since the cleanup.

Debating whether this should be documented. Also the intersperse implementation still needs a better place,
but I think it's OK to reopen the PR by this point.

helix-term/src/ui/statusline.rs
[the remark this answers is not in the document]
hamrik: No feedback arrived for a while on this, and since I agree with @twharmon I will mark this as resolved for now. 

As I said, if the spinner absolutely _must_ reserve space we should use a visible idle character, but it would be more elegant to simply place the spinner on the inner edge of the side section where it won't push other elements around.

hamrik:
I was thinking about `color-modes1 lately. Since the goal is to eventually remove the option altogether, maybe this would be a good time to set it to `true` by default and incorporate the rest of the tweaks mentioned above into the default config.

Then it would still look better for less effort for people who turn the feature off, then it does now when people turn it on without the tweaks.

hamrik:
Looks like `intersperse` won't make into the stdlib anytime soon so I introduced the `itertools` dependency to avoid reinventing the wheel.

If this is a problem I can revert the last commit, before which a hand-rolled solution was used add spacers between statusline elements.

the-mikedavis:
I would like to avoid itertools if possible - it's a fair amount of code to add for things that can usually be written inline. Here I think it's cleaner to use a helper function that returns Spans:

```rust
fn join_with_spaces<'a, I: Iterator<Item = Span<'a>>>(iter: I) -> Spans<'a> {
    let mut spans = Vec::new();
    for elem in iter {
        if !spans.is_empty() {
            spans.push(Span::raw(" "));
        }
        spans.push(elem);
    }
    spans.into()
}

// ..

context.parts.left = join_with_spaces(
    element_ids
        .iter()
        .map(|element_id| get_render_function(*element_id))
        .flat_map(|render| render(context).0),
);
```

hamrik:
Okay, I have reversed the inclusion and also replaced the hand-rolled interspersing iterator with your helper function.

hamrik:
Debating whether to increase spacing between elements from one to two spaces.

hamrik:
Increased spacing to two spaces, which is actually looks closer to how spacing looked before the PR.

hamrik:
Double spacing breaks consistency next to `mode` when color-modes is on. Argh! 

This PR became a mess, so I'm closing it and will make a new one with a smaller scope.

PR #7315, opened 2023-06-11
Add support for showing all LSPs in --health

TheRealLorenz:
Hi there, today I had to debug an lsp configuration with multiple servers and came across this missing feature.
I wrote a quick implementation for the `--health <language>` subcommand, but now I'm stuck at `--health languages`.

In my machine `--health languages` doesn't fully show lang-servers' names and if I were to print all names one after another it would be pretty useless.

My idea was to check for the number of lang-servers and if there are more than one than the output could be something like (the names are pretty bad, I'm seeking better ideas):
```
✓ All (in green, it means all servers are ok)
- Some (in yellow, it means some servers are ok and some are not)
✘ None (in red, it means all servers are not ok)
```
Then for a more detailed output the user could use `--health <language>`.

Otherwise we could show every lang-server name in a different line, but I think it could become a little cumbersome to implement.

Let me know your thoughts!

### DIFF
--- a/helix-term/src/health.rs
+++ b/helix-term/src/health.rs
@@ -193,20 +193,17 @@ pub fn languages_all() -> std::io::Result<()> {
         match cmds.len() {
             0 => column("None", Color::Yellow),
             1 => check_binary(cmds.get(0).cloned()),
-            _ => {
-                let mut checks = cmds
-                    .iter()
-                    .map(|cmd| which::which(cmd).is_ok())
-                    .collect::<Vec<bool>>();
-                checks.sort_unstable();
-                checks.dedup();
-
-                if checks.len() == 2 {
-                    column("- Some", Color::Yellow);
-                } else if checks[0] {
-                    column("✓ All", Color::Green);
-                } else {
-                    column("✘ None", Color::Red);
+            n_configured => {
+                let n_available = cmds.iter().filter_map(|cmd| which::which(cmd).ok()).count();
+                match n_available {
+                    0 => column(&format!("✘ 0/{}", n_configured), Color::Red),
+                    n_available if n_available == n_configured => {
+                        column(&format!("✓ {}/{}", n_available, n_configured), Color::Green)
+                    }
+                    n_available => column(
+                        &format!("- {}/{}", n_available, n_configured),
+                        Color::Yellow,
+                    ),
                 }
             }
         };
--- a/helix-term/src/health.rs
+++ b/helix-term/src/health.rs
@@ -189,18 +189,43 @@ pub fn languages_all() -> std::io::Result<()> {
         None => column("None", Color::Yellow),
     };
 
+    let check_binaries = |cmds: Vec<String>| {
+        match cmds.len() {
+            0 => column("None", Color::Yellow),
+            1 => check_binary(cmds.get(0).cloned()),
+            _ => {
+                let mut checks = cmds
+                    .iter()
+                    .map(|cmd| which::which(cmd).is_ok())
+                    .collect::<Vec<bool>>();
+                checks.sort_unstable();
+                checks.dedup();
+
+                if checks.len() == 2 {
+                    column("- Some", Color::Yellow);
+                } else if checks[0] {
+                    column("✓ All", Color::Green);
+                } else {
+                    column("✘ None", Color::Red);
+                }
+            }
+        };
+    };
+
     for lang in &syn_loader_conf.language {
         column(&lang.language_id, Color::Reset);
 
-        // TODO multiple language servers (check binary for each supported language server, not just the first)
-
-        let lsp = lang.language_servers.first().and_then(|ls| {
-            syn_loader_conf
-                .language_server
-                .get(&ls.name)
-                .map(|config| config.command.clone())
-        });
-        check_binary(lsp);
+        let cmds = lang
+            .language_servers
+            .iter()
+            .filter_map(|ls| {
+                syn_loader_conf
+                    .language_server
+                    .get(&ls.name)
+                    .map(|config| config.command.clone())
+            })
+            .collect();
+        check_binaries(cmds);
 
         let dap = lang.debugger.as_ref().map(|dap| dap.command.to_string());
         check_binary(dap);
--- a/helix-term/src/health.rs
+++ b/helix-term/src/health.rs
@@ -189,22 +189,6 @@ pub fn languages_all() -> std::io::Result<()> {
         None => column("None", Color::Yellow),
     };
 
-    let check_binaries = |cmds: Vec<String>| {
-        match cmds.len() {
-            0 => column("None", Color::Yellow),
-            1 => check_binary(cmds.get(0).cloned()),
-            n_configured => {
-                let n_available = cmds.iter().filter_map(|cmd| which::which(cmd).ok()).count();
-                let (icon, color) = match n_available {
-                    0 => ("✘", Color::Red),
-                    n_available if n_available == n_configured => ("✓", Color::Green),
-                    _ => ("-", Color::Yellow),
-                };
-                column(&format!("{} {}/{}", icon, n_available, n_configured), color);
-            }
-        };
-    };
-
     for lang in &syn_loader_conf.language {
         column(&lang.language_id, Color::Reset);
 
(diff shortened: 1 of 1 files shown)


gabydd:
Maybe it would be better then "all, some, none" to represent it will the symbol plus ``<amount of available servers>/<amount of configured servers>``

TheRealLorenz:
# Update

As @gabydd suggested, I updated the code and now `--health languages` looks like this:

```
✓ 2/2 (in green, it means all servers are ok)
- 1/3 (in yellow, it means some servers are ok and some are not)
✘ 0/4 (in red, it means all servers are not ok)
```

I'm still not sured about the simple for the yellow message, I don't like the hyphen very much. Maybe I could just use `✓`?

helix-term/src/health.rs
```diff
@@ -189,18 +189,40 @@ pub fn languages_all() -> std::io::Result<()> {
...
+                    0 => column(&format!("✘ 0/{}", n_configured), Color::Red),
+                    n_available if n_available == n_configured => {
+                        column(&format!("✓ {}/{}", n_available, n_configured), Color::Green)
+                    }
+                    n_available => column(
+                        &format!("- {}/{}", n_available, n_configured),
+                        Color::Yellow,
+                    ),
+                }
+            }
```
AlexanderBrevig, line 208:
```suggestion
            configured => {
                let available = cmds.iter().filter_map(|cmd| which::which(cmd).ok()).count();
                let (icon, color) =  match available {
                    0 => ("✘", Color::Red),
                    available if available == configured =>  ("✓", Color::Green),
                    available => ("-", Color::Yellow),
                }
                column(&format!("{} {}/{}", icon, available, configured), color)
            }

```

Wondering if something like this perhaps reads clearer?

TheRealLorenz: Applied in the newest commit!

CptPotato:
> I'm still not sure about the symbol for the yellow message, I don't like the hyphen very much. Maybe I could just use `✓` or `✘ `?

The way I understand it, multiple language servers for a language are usually alternatives, right?
In that case I think a check mark `✓` is fine.

AlexanderBrevig:
> The way I understand it, multiple language servers for a language are usually alternatives, right? In that case I think a check mark `✓` is fine.

They all run simultaneously.
A typical setup would be `["typescript", "eslint", "prettier"]` for a typescript setup.
Missing one of them would not feel `✓` and as such, I vote for the `-` indicating partial setup.

gyreas:
Will it be possible to specify the languages to display as such: `hx --health lang1 lang2...langn`?

TheRealLorenz:
> Will it be possible to specify the languages to display as such: `hx --health lang1 lang2...langn`?

Well I don't think it would be difficult to implement, but I think it should be done in a different PR. 

Maybe you open an issue and I could start to work on that!

TheRealLorenz:
# Current output of `--health`

`--health languages`:
```
css            ✓ vscode-css-language-server (green)
svelte         ✘ 0/3 (red)
tsx            - 2/3 (yellow)
typescript     ✓ 2/2 (green)
yuck           None (yellow)
zig            ✘ zls (red)
```
----
`--health <language>`:
```
Configured language servers:
  ✓ typescript-language-server: /Users/lollo/Library/pnpm/typescript-language-server
  ✓ vscode-eslint-language-server: /Users/lollo/Library/pnpm/vscode-eslint-language-server
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
```
```
Configured language servers:
  ✓ clangd: /usr/bin/clangd
Configured debug adapter: lldb-vscode
Binary for debug adapter: 'lldb-vscode' not found in $PATH
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
```
```
Configured language servers: None
Highlight queries: ✓
Textobject queries: ✘
Indent queries: ✘
```

archseer:
> svelte         ✘ 0/3 (red)
> tsx            - 2/3 (yellow)
> typescript     ✓ 2/2 (green)

I'm not sure this is the best UI, it makes it look like I need to install all the servers where one should be enough. I'd maybe list the ones that were detected and then a number of those that weren't found.

EpocSquadron:
> I'm not sure this is the best UI, it makes it look like I need to install all the servers where one should be enough.

Technically, if there are multiple lsps configured this is exactly the case (I configured two more, now I have to see if helix sees them). That is unless and until multiple lsps are configured by default for a language specifically to provide fallbacks rather than expecting multiple to be required. In that case, the alternate UX @TheRealLorenz and I independently suggested is better.

TheRealLorenz:
I think this PR should be included in the upcoming release, as it will be the first release with multiple LSP support.

TheRealLorenz:
> > > svelte         ✘ 0/3 (red)
> > > tsx            - 2/3 (yellow)
> > > typescript     ✓ 2/2 (green)
> > 
> > 
> > I'm not sure this is the best UI, it makes it look like I need to install all the servers where one should be enough. I'd maybe list the ones that were detected and then a number of those that weren't found.
> 
> My first idea was actually to make something like:
> 
> ```
> css            ✓ vscode-css-language-server (green)
> svelte         ✘ svelteserver (red)
>                ✘ eslint-language-server (red)
>                ✘ emmet-langugage-server (red)
> tsx            ✓ typescript-language-server (green)
>                ✘ eslint-language-server (red)
>                ✘ emmet-langugage-server (red)
> typescript     ✓ typescript-language-server (green)
>                ✘ eslint-language-server (red)
> yuck           None (yellow)
> zig            ✘ zls (red)
> ```
> 
> But I though it would have been a little cumbersome to implement.
> 
> Let me know if this may be a better solution to the problem.

I've already implemented this (need to push it), I'm just looking for a feedback now

helix-term/src/health.rs
```diff
@@ -291,6 +308,29 @@ pub fn language(lang_str: String) -> std::io::Result<()> {
     Ok(())
 }
 
+/// Display diagnostics about multiple LSPs and DAPs.
+fn probe_protocols(protocol_name: &str, server_cmds: &[&str]) -> std::io::Result<()> {
```
the-mikedavis, line 312: The collect and `as_slice` calls in the caller can be removed if we have this accept an iterator:

```rust
fn probe_protocols<'a, I: Iterator<Item = &'a str> + 'a>(
    protocol_name: &str,
    server_cmds: I,
) -> std::io::Result<()> {
```

helix-term/src/health.rs
```diff
@@ -291,6 +308,29 @@ pub fn language(lang_str: String) -> std::io::Result<()> {
...
+    let stdout = std::io::stdout();
+    let mut stdout = stdout.lock();
+
+    write!(stdout, "Configured {}s:", protocol_name)?;
+    if server_cmds.is_empty() {
```
the-mikedavis, line 317: This would then need to be a peekable instead:

```rust
let mut server_cmds = server_cmds.peekable();
if server_cmds.peek().is_none() {
```

the-mikedavis:
That alternate UI looks clearer to me

TheRealLorenz:
Pushed commits with the alternative UI and applied the requested changes!

helix-term/src/health.rs
```diff
@@ -213,6 +211,12 @@ pub fn languages_all() -> std::io::Result<()> {
         }
 
         writeln!(stdout)?;
+
+        cmds.try_for_each(|cmd| {
+            column("", Color::Reset);
+            check_binary(Some(cmd));
+            writeln!(stdout)
+        })?;
```
the-mikedavis, line 219: Style-wise I think a for loop is considered cleaner than for_each/try_for_each when possible:

```rust
for cmd in cmds {
    column("", Color::Reset);
    check_binary(Some(cmd));
    writeln!(stdout)?;
}
```

the-mikedavis (commented):
Just a small style nit. I'm liking the way this is looking now 👍

helix-term/src/health.rs
TheRealLorenz: Done 👍🏻

David-Else:
I tested this, all seems great! 

Please try to merge it before the big new release as it will be very important to problem shoot the new multiple language server feature.

the-mikedavis (approved):

pascalkuthe (approved):

PR #7316, opened 2023-06-11
Auto indent change if selection is linewise

Tudyx:
This make `c` auto indent if the selection begins by a whole line. 

I inspired from the other functions in `command.rs`.  I'm not sure if the implementation is very good, all the text editing functions in the code base are new to me.

Resolve https://github.com/helix-editor/helix/discussions/4507 https://github.com/helix-editor/helix/discussions/2783

|before|after|
|----|-----|
|before_1|before_1|
|before_2|after_2|

This also works if multiple lines are selected. Otherwise it behaves exactly as before.

## Why not use a custom key bindings for this?

For me if you select a whole line and go to insert mod with `c`, you want the auto indentation behavior all of the time. 

This is also consistent  with the Vim `V c` or `cc`  behavior.

This PR aims to be a QOL improvement, in the same spirit than https://github.com/helix-editor/helix/pull/5837

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2319,10 +2319,11 @@ fn begin_by_a_whole_line(selection: &Selection, text: &Rope) -> bool {
     for range in selection.ranges() {
         // If at least a full line is selected (strange require 2).
         if range.slice(text.slice(..)).len_lines() >= 2 {
-            // If in the begin of the selection is at the begining of a line.
-            let (start_line, _) = range.line_range(text.slice(..));
+            // If the start of the selection is at the start of a line and the end at the end of a line.
+            let (start_line, end_line) = range.line_range(text.slice(..));
             let start = text.line_to_char(start_line);
-            if start == range.anchor {
+            let end = text.line_to_char((end_line + 1).min(text.len_lines()));
+            if start == range.anchor && end == range.head {
                 return true;
             }
         }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2315,14 +2315,14 @@ enum Operation {
     Change,
 }
 
-fn only_whole_lines(selection: &Selection, text: &Rope) -> bool {
+fn selection_is_linewise(selection: &Selection, text: &Rope) -> bool {
     selection.ranges().iter().all(|range| {
-        // If not at least a full line is selected (strange require 2).
-        if range.slice(text.slice(..)).len_lines() < 2 {
+        let text = text.slice(..);
+        if range.slice(text).len_lines() < 2 {
             return false;
         }
         // If the start of the selection is at the start of a line and the end at the end of a line.
-        let (start_line, end_line) = range.line_range(text.slice(..));
+        let (start_line, end_line) = range.line_range(text);
         let start = text.line_to_char(start_line);
         let end = text.line_to_char((end_line + 1).min(text.len_lines()));
         start == range.from() && end == range.to()
@@ -2333,7 +2333,7 @@ fn delete_selection_impl(cx: &mut Context, op: Operation) {
     let (view, doc) = current!(cx.editor);
 
     let selection = doc.selection(view.id);
-    let only_whole_lines = only_whole_lines(selection, doc.text());
+    let only_whole_lines = selection_is_linewise(selection, doc.text());
 
     if cx.register != Some('_') {
         // first yank the selection
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2315,27 +2315,25 @@ enum Operation {
     Change,
 }
 
-fn begin_by_a_whole_line(selection: &Selection, text: &Rope) -> bool {
-    for range in selection.ranges() {
-        // If at least a full line is selected (strange require 2).
-        if range.slice(text.slice(..)).len_lines() >= 2 {
-            // If the start of the selection is at the start of a line and the end at the end of a line.
-            let (start_line, end_line) = range.line_range(text.slice(..));
-            let start = text.line_to_char(start_line);
-            let end = text.line_to_char((end_line + 1).min(text.len_lines()));
-            if start == range.anchor && end == range.head {
-                return true;
-            }
+fn only_whole_lines(selection: &Selection, text: &Rope) -> bool {
+    selection.ranges().into_iter().all(|range| {
+        // If not at least a full line is selected (strange require 2).
+        if range.slice(text.slice(..)).len_lines() < 2 {
+            return false;
         }
-    }
-    false
+        // If the start of the selection is at the start of a line and the end at the end of a line.
+        let (start_line, end_line) = range.line_range(text.slice(..));
+        let start = text.line_to_char(start_line);
+        let end = text.line_to_char((end_line + 1).min(text.len_lines()));
+        start == range.anchor && end == range.head
+    })
 }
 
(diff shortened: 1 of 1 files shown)


pascalkuthe (commented):
I like the general utility and I think @archseer had something similar in mind. However, I think the details should be improved: The current implementation doesn't work with multi cursors. Changing that should be easy (just perform the check for all selection ranges). I also think that the condition should be more strict: We should only insert a new line if the selections exactly covers 1 line (and not just when it starts at a line start)

danillos:
I took some time to be used to the newline char and when it is selected and hit `c` it would change the char. Like it does currently.

If sometimes it doesn't happens, the editor will not following a standard.

I would prefer to add this as default or something like it:

```
m = { l = ["extend_to_line_bounds", "trim_selections"] }
```

Tudyx:
Thanks to your advice, I added  support with multi-cursors.
I've also been stricter about the required condition, the selection may only contain whole lines. I didn't restrict to one line in particular because, in my opinion, we want this behavior even with multiple lines: 

before | after
---- |-----
multi_line_1|multi_line_1
multi_line_2_before | multi_line_2_after

Tudyx:
I found another implementation that is simpler and handles more edge case using `trim_selections`: 

```rust
fn delete_selection_impl(cx: &mut Context, op: Operation) {

    let selection = doc.selection(view.id);

    if cx.register != Some('_') {
        // first yank the selection
        let values: Vec<String> = selection.fragments(text).map(Cow::into_owned).collect();
        let reg_name = cx.register.unwrap_or('"');
        cx.editor.registers.write(reg_name, values);
    };

    if matches!(op, Operation::Change) {
        trim_selections(cx);
    }

    let selection = doc.selection(view.id);
    // then delete
    let transaction =
        Transaction::delete_by_selection(doc.text(), selection, |range| (range.from(), range.to()));

    match op {
        Operation::Delete => {
            // exit select mode, if currently in select mode
            exit_select_mode(cx);
        }
        Operation::Change => {
            enter_insert_mode(cx);
        }
    }
}
```

The current implementation with multiple cursors doesn't work if at least one selection doesn't cover a whole line, but this one does.
I've made sure that the buffer we still store the none-trimmed version.
What do you think?

the-mikedavis:
Isn't that the same as using `trim_selections` (`_`) and then `c`? IMO this should either be done explicitly (i.e. `_c`) or with a custom keybind rather than built in to `c`.

clo4:
I agree in principle, _c is only one more keystroke and there's huge value in the consistency helix currently has. Although I think there's probably more value in making the default keybinds as ergonomic as possible. Most of the time _c *is* what I want to do. Vim makes the same choice with `cc` and `S` for good reason.

Maybe a solution is to implement the default `c` binding in terms of a new command, and keep change_selection as it is to (1) give people the option to use the current behavior and (2) avoid potentially breaking people's keybinds?

pascalkuthe:
I like the idea of introducing a new command. I think while this command is a bit less consistent than the current binding I think this is such a common usecase that it's worth introducing. Regarding the new implementation you posted I think I prefer the current implementation. `xc` acting the same as `xdo` makes sense to me. `x_c` leaves training whitepaces and leading whitespace that were added manually (that the user likely wants to remove, think a weidly formatted macro being replaced with normal code).

Tudyx:
Thanks for your feedback, I will implement it as a new command then.

Tudyx:
I've implemented as a new command. I've also added a "no yank" version, as it also exists for the previous change command.

If you want to try it out, please add this to your config:

```toml
[keys.normal]
"c" = "change_selection_with_indent"
[keys.select]
"c" = "change_selection_with_indent"
```

helix-term/src/commands.rs
```diff
@@ -296,6 +296,8 @@ impl MappableCommand {
         delete_selection, "Delete selection",
         delete_selection_noyank, "Delete selection without yanking",
         change_selection, "Change selection",
+        change_selection_with_indent, "Change selection and place the cursor at the appropriated indent",
```
pascalkuthe, line 299: I am not sure what would be a good name but J think this isn't that clear. Yes this indents but the lrinarmy difference is that a newline is inserted (if lines are aelwcrez)

pascalkuthe:
This looks good to me for the most part but I prefer a better name for the command. I think we can make this the default mapping too

helix-term/src/commands.rs
gabydd: I guess it is sort of  ``change_selection_at_line_start`` to follow the pattern of insert command

Tudyx: Maybe `change_selection_and_auto_indent` would be clearer? Otherwise 
`change_selection_and_open_above_if_lines_selected` or
`change_selection_and_add_new_line_if_lines_selected` might be more explicit but also more verbose

the-mikedavis:
Actually I think just having the one command (`change_selection`) is ok. The current `change_selection` command is effectively equivalent to `di` so restoring the existing behavior is as easy as binding those two commands or typing them out fully in the cases where you don't want the indentation.

I've tinkered around with `c` while thinking about this change. In my own workflows/muscle-memory I almost never use `c` across lines/blocks: I always insert new stuff and then delete the old. Trying out some refactors and intentionally using `c` though, I can see the motivation for this change. In general I don't like adding too much nuanced behavior to commands because I think it can make them unpredictable when you have multiple selections all over a file, but I think I'm ok with it in this scenario 👍

helix-term/src/commands.rs
```diff
@@ -2315,10 +2315,25 @@ enum Operation {
     Change,
 }
 
+fn only_whole_lines(selection: &Selection, text: &Rope) -> bool {
```
the-mikedavis, line 2318: Let's call this `selection_is_linewise` to re-use that terminology from vi

helix-term/src/commands.rs
```diff
@@ -2315,10 +2315,25 @@ enum Operation {
...
+    selection.ranges().iter().all(|range| {
+        // If not at least a full line is selected (strange require 2).
+        if range.slice(text.slice(..)).len_lines() < 2 {
+            return false;
+        }
+        // If the start of the selection is at the start of a line and the end at the end of a line.
+        let (start_line, end_line) = range.line_range(text.slice(..));
+        let start = text.line_to_char(start_line);
+        let end = text.line_to_char((end_line + 1).min(text.len_lines()));
+        start == range.from() && end == range.to()
```
the-mikedavis, line 2328: There are a few places here that turn the rope into a ropeslice. We can slice once at the beginning instead: `let text = text.slice(..);`

pascalkuthe (approved):

helix-term/src/commands.rs
```diff
@@ -2315,10 +2315,25 @@ enum Operation {
     Change,
 }
 
+fn only_whole_lines(selection: &Selection, text: &Rope) -> bool {
+    selection.ranges().iter().all(|range| {
+        // If not at least a full line is selected (strange require 2).
+        if range.slice(text.slice(..)).len_lines() < 2 {
+            return false;
+        }
```
the-mikedavis, line 2323: The "strange require 2" part is a bit confusing here in the comment - is that just commentary on ropey's len_lines?

To clarify about ropey's behavior: `len_lines` returns the count of line breaks in the slice + 1, so even for an empty slice/rope `len_lines` will return `1`. If there are any line breaks then this will be 2 or more.

the-mikedavis (approved):

PR #7322, opened 2023-06-12
Fix underflow when repeating a completion that has a negative shift position

Philipp-M:
While rewriting #2608 I have had a panic, when repeating a completion edit at the beginning of a buffer, with a negative start range, this should fix the issue...


pascalkuthe (approved):
Thanks!

PR #7349, opened 2023-06-15
render completion options by language server priority

estin:
Hi!
On multi ls with completion items rendered without any priority.
This PR groups completion items by language server id.
And in future allows to do some visualization to separate groups 

For my use case (on python) with follow config

```toml
[language-server]
simple-completion-language-server = { command = "simple-completion-language-server" }

[[language]]
name = "python"
language-servers = [ "pylsp", "simple-completion-language-server" ]
```
  

on master, I expects items from `pylsp` on top of the list, for me it primary ls for completion
as-is

on this PR, all as I expected, on top items from `pylsp` and on bottom from `simple-completion-language-server`
with-group

Don't sure about it's nature behavior for other users.
Feel free to close this PR.
Sorry for my poor English.


poliorcetics:
LGTm, though what is `simple-completion-language-server` ? A simple text search doesn't give me much results in duckduckgo or github

estin:
@poliorcetics it's https://github.com/estin/simple-completion-language-server
simple and buggy solution for lexical completion https://github.com/helix-editor/helix/pull/3328

poliorcetics:
Thank you very much, I'll have look!

helix-term/src/ui/completion.rs
```diff
@@ -19,6 +19,12 @@ use helix_lsp::{lsp, util, OffsetEncoding};
 
 impl menu::Item for CompletionItem {
     type Data = ();
+
+    #[inline]
+    fn group(&self, _data: &Self::Data) -> usize {
+        self.language_server_id
+    }
```
the-mikedavis, line 26: I can't think of other places we would want to re-use this `group` concept. Ideally I'd like to see this built in to the `sort_text` function: something like `fn sort_key(&self, data: &Self::Data) -> impl Ord` and then `CompletionItem`'s `Item` implementation would contain the logic for using `self.language_server_id`. That depends on either https://github.com/rust-lang/rust/issues/91611 to return an `impl Ord`, or https://github.com/rust-lang/rust/issues/29661 so that we wouldn't have to explicitly implement `sort_key` in every Item:

```rust
pub trait Item {
    type Data;
    type SortKey: Ord = Cow<str>;

    // ...

    fn sort_key(&self, data: &Self::Data) -> Self::SortKey {
        self.filter_text()
    }

    // ...
}
```

Maybe there's a clever way around the language features here that I'm not seeing though?

There was a similar thread of this idea in https://github.com/helix-editor/helix/pull/2705#discussion_r927234858.

pascalkuthe:
I dont think this needs a config option since it only acts as a tie breaker

helix-term/src/ui/menu.rs
```diff
@@ -100,14 +105,16 @@ impl<T: Item> Menu<T> {
...
                     self.matcher
                         .fuzzy_match(&text, pattern)
-                        .map(|score| (index, score))
+                        .map(|score| (index, group, score))
                 }),
         );
         // Order of equal elements needs to be preserved as LSP preselected items come in order of high to low priority
-        self.matches.sort_by_key(|(_, score)| -score);
+        self.matches
+            .sort_by_key(|(_, group, score)| (*group, -*score));
```
pascalkuthe, line 117: Ah actually if we keep this code as is then it actually shows all items of LSP A over LSP B that does need a cobfig option of sorts. I think it would be the other way around. This should not be the default iMO

helix-term/src/ui/completion.rs
estin: - removed group concept.
- introduced `Item.score` function to control score and sorting

helix-term/src/ui/completion.rs
```diff
@@ -17,6 +17,9 @@ use crate::ui::{menu, Markdown, Menu, Popup, PromptEvent};
 
 use helix_lsp::{lsp, util, OffsetEncoding};
 
+const WEIGHT_SCORE_STEP: u32 = 1_000_000;
```
pascalkuthe, line 20: Messing with the scoring of a fuzzy marcher is never a good idea. Consider them implementation details. The sorting by server should only be done as a tie breaker (like you did previously but instead or raking by server first and then by score it should be the other way around)

helix-term/src/ui/menu.rs
```diff
@@ -18,6 +18,9 @@ pub trait Item: Sync + Send + 'static {
     /// Additional editor state that is used for label calculation.
     type Data: Sync + Send + 'static;
 
+    /// Second ordering rule on equals items score
+    type TieBreaker;
+
```
archseer, line 23: I think you should be able to add a default value here so we don't have to define it in every single implementation.

helix-term/src/ui/menu.rs
```diff
@@ -29,30 +32,36 @@ pub trait Item: Sync + Send + 'static {
         let label: String = self.format(data).cell_text().collect();
         label.into()
     }
+
+    fn tie_breaker(&self, _data: &Self::Data) -> Self::TieBreaker;
```
archseer, line 36: But I don't like having this extra function, it should be unnecessary because this is the intent of `sort_text` (should really be renamed as sort_key).

Let's try modifying the return value of `sort_text` as `K: Ord`, and use a default that uses `Cow<str>`. Then for your one specific implementation you can instead return `(label, language_server_id)` (`(Cow<str>, usize)`)

pascalkuthe: I agree we don't want an extra method. I am not sure what the right API is here though.

A couple points:

* for the picker the concept of sort text does not make sense IMO. In general I think the needs of the picker and of the menu keep diverging (some of mikes colum related work for example) so I think we will likely just endup with two traits (or even something totally different for the picker)
* for the menu I am not sure how much sense all of this actually makes. It feels wrong to me to sort by language server first since fuzzy matching is not very selective so you would see all the "bad" completions from one server before those from a second server. A score bonus is not a good solution because it's hard to implement (how large should the bonuss shpuld be is subjctive and would depend on word length/content) and ultimately not feel predictable. I proposed making it a tie breaker but exactly identicsl fuzzy scores are not as common as you would think fue to the fairly complex bpnus system. It's probably the least bad option right now. 
* all of this doesn't even take into account yet that we do not actually want to sort for a lot of servers. With incomplete completion request support we have to rerequest completions on every keypress and are not supposed to do filtering and scoring ourself. The question then becomes how to mix those servers with other completion options. The server sorts internslly according to their own criteria and scoring system which we are unaware of and also not comparable. This feature has a really large impact and is very commonly used (and the capability is not respected, for example by RA) so I see it as fairly high priority so such an obvious conflict here is a concern for me. 

This makes it hard for me to review since I dont understand the full problem space yet. We may endup needing a very different solution and changing the trait before we have figured that out feels premature. Perhaphs sorting by server first works after all if we filter items with a score below 50% of the top score or something. Ideally we should study what other ides (primarily vscode) do here. There is also a whole can of worms of super long LSP issue trackers where it was discussed how filtering and sorting should work exactly (which can sometimes not have a clear co conclusion,  I need to read trough all that carefully again).

One thing I am fairly sure about is that we want this to be (language_server, label) and not the other way around. Score ties are already rare (which is the only situation where sort_text matters) and when they do occur we want to group by language sever first (otherwise the language sever comparison basically never matters anyway).

archseer: > One thing I am fairly sure about is that we want this to be (language_server, label)

I was assuming that too, but that probably doesn't make sense either since you could have very weak matches from your priority server that would now appear before better matches from a second one.

pascalkuthe: Ah no that actually would not happen (currently). Sort text is not used for fuzzy matching. It's only used as a tie breaker if the fuzzy match score is equal.


So sorting would be `(fuzzy_score, language_server, label)` so the order of these two only takes effect when the score is the same (which is already pretty rare)

Issue #6966, opened 2023-05-04
Prevent Helix from scrolling too far past the EOF


pascalkuthe:
### Discussed in https://github.com/helix-editor/helix/discussions/6965

<div type='discussions-op-text'>

<sup>Originally posted by **dist1ll** May  4, 2023</sup>
Between v22.12 and v23.03, the scrolling behavior at the bottom of a file changed. 
 Now when I scroll to the very bottom, instead of showing only ~8 lines past the end of the file, helix allows me to scroll down almost an entire terminal-height worth of lines. 
</div>

I was able to reproduce this (tested both 22.12 and 23.3 to compare),  this an unintended side-effect of #5420 that we missed. This might be a bit tricky to solve since we only use relative positions with #5420 but I might be able to come up with a heuristic that provides good performance regardless

pascalkuthe:
We decided against reintroducong this. It's actually quite useful in practice to be able to scroll down

rose-bouchayer:
Hi @pascalkuthe, I do understand that it's a useful feature but could this behaviour be configurable please? It would make everyone happy by letting users choose what they prefer.

PR #7350, opened 2023-06-15
Prevent Helix from scrolling too far past the EOF

Xalfer:
Resolves issue #6966.
Also removes redundant variable `doc_text` because it is the same as `text`.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -28,8 +28,8 @@ use helix_core::{
     textobject,
     tree_sitter::Node,
     unicode::width::UnicodeWidthChar,
-    visual_offset_from_block, Deletion, LineEnding, Position, Range, Rope, RopeGraphemes,
-    RopeSlice, Selection, SmallVec, Tendril, Transaction,
+    visual_offset_from_anchor, visual_offset_from_block, Deletion, LineEnding, Position, Range,
+    Rope, RopeGraphemes, RopeSlice, Selection, SmallVec, Tendril, Transaction,
 };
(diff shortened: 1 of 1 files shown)


pascalkuthe (changes requested):
This does not work with soft wrapped text, approximating with document lines doesn't work Jere as it prevents scrolling slk the way down with soft wrapped text

Xalfer:
> This does not work with softwrapping or virtual text, approximating with document lines doesn't workh here as it prevents scrolling slk the way down with soft wrapped text (or virtual text lines). You need to use a relative visual offset from the EOF. However, that would make scrolling quite slow so that computation should only be done if we are past the threshold you are using here (even that might not be enough in the future with conceal but we will have to figure that out when we get to that)

True, after I made the pull request I've been thinking about the virtual lines and how this actually does nothing about them. When I get the time I'll try to come up with something. Thanks for the feedback.

Xalfer:
So I made some changes and now it works. Now the problem is that the code looks really bad. I'd appriciate feedback from someone whose more knowledgable with the codebase than I do.

pascalkuthe (commented):
You just need a single function call, calcutlting the initak position is unneeded and really slow/inefficient. Instead you van transverse from the bottom. There might be an one by off error in the code below you will habe to try but you should get the idea

``` rust 
char_idx_at_visual_offset(
                    text,
                    text.len_chars(),
                    - height as isize + scrolloff as isize,
                    0,
                    &text_fmt,
                    &annotations,
                );
            }
```

PR #7373, opened 2023-06-18
Re-render after a timeout when debouncing LSP messages

the-mikedavis:
The current debounce function has a bug that can prevent rendering when we should: if a language server sends multiple messages within an LSP_DEADLINE we correctly debounce them but do not trigger a render when the language server is done. For example a language server may send many progress messages but we may miss rendering the work done progress end message, preventing the spinner from going away.

The problem is the `editor.language_servers.incoming.is_empty()` check. This is actually [`SelectAll::is_empty`][method] from futures_util: it checks that the set of streams is empty rather than that there are no pending messages in any stream.

~I don't believe there is a way to check that any streams have pending messages, so this commit uses a timer instead which triggers a render when the debounce is done. So we render once per LSP_DEADLINE in cases when a language server sends many messages.~

We can refactor the incoming message stream to be a Peekable stream and then we can poll that once to check if there is another message ready and waiting.

[method]: https://docs.rs/futures-util/0.2.1/futures_util/stream/struct.SelectAll.html#method.is_empty

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -9,7 +9,10 @@ pub use jsonrpc::Call;
 pub use lsp::{Position, Url};
 pub use lsp_types as lsp;
 
-use futures_util::stream::select_all::SelectAll;
+use futures_util::{
+    stream::{select_all::SelectAll, Peekable},
+    StreamExt,
+};
 use helix_core::{
     path,
     syntax::{LanguageConfiguration, LanguageServerConfiguration, LanguageServerFeatures},
@@ -19,6 +22,7 @@ use tokio::sync::mpsc::UnboundedReceiver;
 use std::{
     collections::HashMap,
     path::{Path, PathBuf},
+    pin::Pin,
     sync::Arc,
 };
 
@@ -619,12 +623,14 @@ impl Notification {
     }
 }
 
+type IncomingStreams = Pin<Box<Peekable<SelectAll<UnboundedReceiverStream<(usize, Call)>>>>>;
+
 #[derive(Debug)]
 pub struct Registry {
     inner: HashMap<LanguageServerName, Vec<Arc<Client>>>,
     syn_loader: Arc<helix_core::syntax::Loader>,
     counter: usize,
-    pub incoming: SelectAll<UnboundedReceiverStream<(usize, Call)>>,
+    pub incoming: IncomingStreams,
 }
 
 impl Registry {
@@ -633,7 +639,7 @@ impl Registry {
             inner: HashMap::new(),
             syn_loader,
             counter: 0,
-            incoming: SelectAll::new(),
+            incoming: Box::pin(SelectAll::new().peekable()),
         }
     }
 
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I believe you could simply use a [peekable stream](https://docs.rs/futures-util/latest/futures_util/stream/trait.StreamExt.html#method.peekable)  instead and use `peek().is_some()` to check if there is another element.

the-mikedavis:
Good call, I was hoping there was a way to do this but didn't find Peekable. I've refactored the streams to use Peekable and then we can poll the peekables once to check for outstanding ready messages.

helix-lsp/src/lib.rs
```diff
@@ -619,12 +623,14 @@ impl Notification {
     }
 }
 
+type IncomingStreams = SelectAll<Pin<Box<Peekable<UnboundedReceiverStream<(usize, Call)>>>>>;
```
pascalkuthe, line 626: Instead of `SelectAll<Peekabke` can't wu use `Peekabke<SelectAll`? That's what I had in mind and I thi k might be simpler.  ut maybe I missed something.

the-mikedavis: Ah yeah I think you're right. I ran into some problems with the Pin when I tried that first but it looks like it's possible

pascalkuthe:
I tested this out and using `Peekable` doesn't work at all because a peelable stream is apparently always fused (so if we peek and nothing is available the stream will never yield anything ever again...). which meant we never handled any LSP messages.

Implementing your own stream that handles that is kind of straightforward (and actually cleans up the code a bit)  since we are only dealing with unpin types. I did that in 255605d88ed4d120282074afa9d296c186b51815. It works as intended.

However, the debouncing isn't super reliable since we are essentially using the stream here. This change caused around 120 extra renders when starting RA on helix for me. A proper debounce timeout works better in that regard. I was a bit uncomfortable with the original change because it adds yet another single-purpose timer. 

However, we already have a generic mechanism for (async) debounced rending that is quite robust: the redraw handle (I added that for the diff gutter). It was quite straightforward to change the LSP rendering to use that: 833a5c7647f8977ed42b1916051da21cbb4a4047. The only disadvantage of this approach was that the rendering now has a longer render timeout (96ms instead of 16ms) which felt a bit slow. I changed it to 33ms to emulate 30FPS. The redraw handle works slightly differently and as a result the rendering es more evenly spaced which makes the 33 30 Fps actually feel pretty smooth/less stuttery here (and also reduces performance overhead/battery draw). 

You can give that commit a try. I think the second variant would be my preferred fix (and actually removes code instead of adding new code)

the-mikedavis:
Aha, I thought the progress spinner was working since it disappeared but actually it just never appeared because of the Peekable 😄

I was curious about using the redraw handle for this. I'll take that for a spin but I like it already since it's so many `-` and so few `+`!

the-mikedavis:
https://github.com/helix-editor/helix/commit/833a5c7647f8977ed42b1916051da21cbb4a4047 works perfectly, thanks! Please open it up as a PR when you get a chance

pascalkuthe:
I opened #7538

PR #7376, opened 2023-06-18
Write buffer next/previous commands

johnnyfreeman:
Here are some new commands which would be useful to me:

* Write buffer next
* Write buffer previous


archseer:
I think these should just be a custom user mapping, no need for a built-in

johnnyfreeman:
I don't disagree. If that's the official ruling, I'll live with it.

Just for fun, allow me to try playing devil's advocate as this is my first Rust PR. 😀

1. This exists in other editors (`:wn` in vi) out of the box and might make helix adoption a little easier for these folks.

2. Additionally, these seem to pair well with the already built-in `:wbc` command.

Otherwise, how would I go about adding these mappings myself as you suggest?

Thanks!

gabydd:
We don't currently have support for custom typeable commands so you would have to bind it to key in normal mode

johnnyfreeman:
Thank you.

Issue #7386, opened 2023-06-19
Change MATLAB parser


acristoffers:
Hi,

I developed a new MATLAB parser ([acristoffers/tree-sitter-matlab](https://github.com/acristoffers/tree-sitter-matlab)) which is now being used in neovim too (nvim-treesitter/nvim-treesitter#4944).

However, I'm not an Helix user and don't know exactly what are your requirements, nor how to actually test things out. I have cloned this repo, changed the matlab entry to fetch the right repo and replaced the highlights.scm file for that of emacs and it kinda works, except it's not very colorful (I don't know if it is because of the theme - dracula - or wrong captures) and I noticed this (I don't know what's going on that only 2 of the exactly same things got highlighted):

![image](https://github.com/helix-editor/helix/assets/2191364/81a160ec-24af-492b-8e5a-e8442dcd7615)

Also, I could not get textobjects to work.

So, can someone point me out in the good direction to provide helix-specific queries? Do you have documentation on what is supported, what queries are available, which order the queries match (I guess first in the document gets priority)? I would love to work with you to make the parser support your editor, even if I don't use it.

Thank you.

P.S.: If you change to my parser, this issue will be resolved: #7058.

the-mikedavis:
The scopes we use are here: https://docs.helix-editor.com/master/themes.html#syntax-highlighting. For example when rewriting neovim queries we often need to change `@integer` to `@constant.numeric.integer` and `@parameter` to `@variable.parameter`. I believe we have the same precedence as emacs' tree-sitter: the top-most pattern in the query file that matches gets precedence. Also see https://docs.helix-editor.com/master/guides/adding_languages.html#queries

the-mikedavis:
Yeah we only support highlights (plus injections and locals), indent and textobject queries currently. I think it's fine to include the others since they might be nice starting points when we implement folding for example. The tags.scm queries though I'm not sure we'll ever use, those could probably be dropped from the repo.

gabydd:
Closed by https://github.com/helix-editor/helix/pull/7388

gabydd:
Closed by https://github.com/helix-editor/helix/pull/7388

PR #7388, opened 2023-06-19
Switch MATLAB tree-sitter parser

acristoffers:
Product of #7386.

I believe I got the highlight right, or at least now I get a nice rainbow-vomit when using the `doom_acario_dark` theme. And no more quirks.

Textobjects and indents were tested and are inline with what I found in other grammars. That is to say "could you please take a look at the indents because I think it is broken, but it is defined the same as other languages so I don't know if it's broken or if it's just me who doesn't know how this editor works".

 I simply copied the other queries and did not adapt them or anything. I not even know if helix supports them.

image

### DIFF
--- a/runtime/queries/matlab/highlights.scm
+++ b/runtime/queries/matlab/highlights.scm
@@ -1,97 +1,127 @@
-  ; highlights.scm
+; Constants
 
-function_keyword: (function_keyword) @keyword.function
+(events (identifier) @constant)
+(attribute (identifier) @constant)
+
+"~" @constant.builtin
+
+; Fields/Properties
+
+(superclass "." (identifier) @variable.other.member)
+(property_name "." (identifier) @variable.other.member)
+(property name: (identifier) @variable.other.member)
+
+; Types
+
+(class_definition name: (identifier) @keyword.storage.type)
+(attributes (identifier) @constant)
+(enum . (identifier) @type.enum.variant)
+
+; Functions
 
 (function_definition
-function_name: (identifier) @function
-(end) @function)
+  "function" @keyword.function
+  name: (identifier) @function
+  [ "end" "endfunction" ]? @keyword.function)
 
-(parameter_list (identifier) @variable.parameter)
+(function_signature name: (identifier) @function)
+(function_call name: (identifier) @function)
+(handle_operator (identifier) @function)
+(validation_functions (identifier) @function)
+(command (command_name) @function.macro)
+(command_argument) @string
+(return_statement) @keyword.control.return
 
-[
-    "if"
-    "elseif"
-    "else"
-    "switch"
-    "case"
-    "otherwise"
-] @keyword.control.conditional
+; Assignments
 
-(if_statement (end) @keyword.control.conditional)
-(switch_statement (end) @keyword.control.conditional)
+(assignment left: (_) @variable)
+(multioutput_variable (_) @variable)
 
-["for" "while"] @keyword.control.repeat
-(for_statement (end) @keyword.control.repeat)
-(while_statement (end) @keyword.control.repeat)
+; Parameters
 
-["try" "catch"] @keyword.control.exception
-(try_statement (end) @keyword.control.exception)
+(function_arguments (identifier) @variable.parameter)
 
-(function_definition end: (end) @keyword)
-
-["return" "break" "continue"] @keyword.return
-
-(
-(identifier) @constant.builtin
-(#any-of? @constant.builtin "true" "false")
-)
-
-(
-    (identifier) @constant.builtin
-    (#eq? @constant.builtin "end")
-)
-
-;; Punctuations
-
-[";" ","] @punctuation.special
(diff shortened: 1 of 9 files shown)


runtime/queries/matlab/highlights.scm
```diff
@@ -1,97 +1,127 @@
...
+
+; Punctuation
+
+[ ";" "," "." ] @punctuation.delimiter
+[ "(" ")" "[" "]" "{" "}" ] @punctuation.bracket
+
+; Literals
+
+(escape_sequence) @escape
+(formatting_sequence) @escape
```
the-mikedavis, line 106:
```suggestion
(escape_sequence) @constant.character.escape
(formatting_sequence) @constant.character.escape
```

runtime/queries/matlab/highlights.scm
```diff
@@ -1,97 +1,127 @@
...
 (string) @string
+(number) @constant.numeric.float
+(boolean) @constant.builtin.boolean
 
-;; Comment
-(comment) @comment
+; Comments
+
+[ (comment) (line_continuation) ] @comment.line @spell
```
the-mikedavis, line 113: The `@spell` capture should probably be removed. I think if we do spell-checking with tree-sitter in core then we'll use a separate query file for the spelling queries than the highlight queries.

runtime/queries/matlab/injections.scm
```diff
@@ -0,0 +1 @@
+(comment) @comment
```
the-mikedavis, line 1:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
```

We follow tree-sitter's injection format: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection. I believe this is neovim's style. (I'm not sure if emacs follows nvim here.)

runtime/queries/matlab/locals.scm
the-mikedavis, line 1: For locals we also follow upstream tree-sitter: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

So there should only be `@local.reference`, `@local.definition` and `@local.scope`. We don't differentiate between variable/function/parameter definitions and references.

the-mikedavis:
If you run `cargo xtask docgen` and commit the results that will clear up the X in the docs CI

runtime/queries/matlab/highlights.scm
acristoffers: I skipped them and forgot to go back 🤦‍♂️

runtime/queries/matlab/highlights.scm
acristoffers: I saw it was not in the list, but I wasn't sure whether it was used or not. I'll remove it.

runtime/queries/matlab/locals.scm
acristoffers: I forgot I had to change those for neovim. I may have the original in an old commit, I'll look for it.

runtime/queries/matlab/locals.scm
```diff
@@ -0,0 +1,27 @@
...
+  (function_arguments
+    (identifier)* @local.definition
+    ("," (identifier) @local.definition)*)?) @local.scope
+
+(assignment variable: ((function_call
+                         name: (identifier) @local.definition)))
+(assignment variable: ((struct . [(function_call
+                                    name: (identifier) @local.definition)
+                                  (identifier) @local.definition])))
+(assignment variable: (_) @local.definition)
```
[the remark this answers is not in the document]
acristoffers: I forgot it got changed to `left`, sorry.

runtime/queries/matlab/locals.scm
```diff
@@ -0,0 +1,19 @@
+(function_definition name: (identifier) @local.definition ?) @local.scope
+(function_arguments (identifier)* @local.definition)
+
+(lambda (arguments (identifier) @local.definition)) @local.scope
+
+(assignment left: ((function_call
+                         name: (identifier) @local.definition)))
+(assignment left: ((struct . [(function_call
+                                    name: (identifier) @local.definition)
+                                  (identifier) @local.definition])))
```
the-mikedavis, line 10: It looks like the query analysis is upset about this `struct` node. Was this renamed? I don't see it in the `grammar.js`

acristoffers: I finally had the decency of copying the files over and running `tree-sitter test` before committing, so there should be no more errors like that. I'm sorry about this, it was a busy day and I'm more tired that I thought.

the-mikedavis (approved):
Thank you!

the-mikedavis:
It's not necessary: since it's a single commit we will probably squash+merge which will rename the commit and include the PR number

pascalkuthe (approved):
Thanks!

acristoffers:
One question: do you update the tagged commit automatically or do I have to make pull-requests?

pascalkuthe:
It's done manually since grammar changes usually also require changes to the queries

PR #7408, opened 2023-06-20
map positions through changes in O(N)

### LINKED ISSUE #7396: Hangs when select and edit in a 15MB file.
### Summary

The file is 14.9 MB in size sample data
```
Ok("any-string")
Ok("any-string")
Ok("any-string")
... 388892 lines like that
```

I try "% + s + Ok\(" and it hangs

### Reproduction Steps

 

I tried this:

1. `hx 15mb_file_name`
2. % + s + Ok\( + enter 

I expected this to happen:
I should be able to edit the text.

Instead, this happened:
It hangs and does not respond.

### Helix log

Unable to get logs since it is not responding at all.

### Platform

Linux (Ubuntu 23.04)

### Terminal Emulator

alacrity + zellij

### Helix Version

master

image

pascalkuthe:
Closes #7396

Previously we iterated the entire changset for every position we mapped through the changes. That made the mapping `O(MN)` when mapping multiple positions  and `O(N^2)` for most mutlicursor operations but since we are basically always dealing with sorted lists we can do the two iterations in lockstep which allows  `O(M+N)` mapping

With this PR the example from #7396 now is pretty usable and the mapping doesn't show up in the flamegraph anymore

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -451,15 +451,8 @@ impl Selection {
 
     /// Map selections over a set of changes. Useful for adjusting the selection position after
     /// applying changes to a document.
-    pub fn map(mut self, changes: &ChangeSet) -> Self {
-        if changes.is_empty() {
-            return self;
-        }
-        self = self.map_no_normalize(changes);
-        // TODO: only normalize if needed (any ranges out of order)
-        self = self.normalize();
-
-        self
+    pub fn map(self, changes: &ChangeSet) -> Self {
+        self.map_no_normalize(changes).normalize()
     }
 
     /// Map selections over a set of changes. Useful for adjusting the selection position after
@@ -524,6 +517,9 @@ impl Selection {
 
     /// Normalizes a `Selection`.
     fn normalize(mut self) -> Self {
+        if self.len() < 2 {
+            return self;
+        }
         let mut primary = self.ranges[self.primary_index];
         self.ranges.sort_unstable_by_key(Range::from);
 
@@ -588,17 +584,12 @@ impl Selection {
         assert!(!ranges.is_empty());
         debug_assert!(primary_index < ranges.len());
 
-        let mut selection = Self {
+        let selection = Self {
             ranges,
             primary_index,
         };
 
-        if selection.ranges.len() > 1 {
-            // TODO: only normalize if needed (any ranges out of order)
-            selection = selection.normalize();
-        }
-
-        selection
+        selection.normalize()
     }
 
     /// Takes a closure and maps each `Range` over the closure.
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -161,34 +161,35 @@ impl Range {
         self.from() <= pos && pos < self.to()
     }
 
-    /// Map a range through a set of changes. Returns a new range representing the same position
-    /// after the changes are applied.
-    pub fn map(self, changes: &ChangeSet) -> Self {
+    /// Map a range through a set of changes. Returns a new range representing
+    /// the same position after the changes are applied. Note that this
+    /// function runs in O(N) (N is number of changes) and can therefore
+    /// cause performance problems if run for a large number of ranges as the
+    /// complexity is then O(MN) (for multicuror M=N usually). Instead use
+    /// [Selection::map] or [ChangeSet::update_positions] instead
+    pub fn map(mut self, changes: &ChangeSet) -> Self {
         use std::cmp::Ordering;
-        let (anchor, head) = match self.anchor.cmp(&self.head) {
-            Ordering::Equal => (
-                changes.map_pos(self.anchor, Assoc::After),
-                changes.map_pos(self.head, Assoc::After),
-            ),
-            Ordering::Less => (
-                changes.map_pos(self.anchor, Assoc::After),
-                changes.map_pos(self.head, Assoc::Before),
-            ),
-            Ordering::Greater => (
-                changes.map_pos(self.anchor, Assoc::Before),
-                changes.map_pos(self.head, Assoc::After),
-            ),
-        };
-
-        // We want to return a new `Range` with `horiz == None` every time,
-        // even if the anchor and head haven't changed, because we don't
-        // know if the *visual* position hasn't changed due to
-        // character-width or grapheme changes earlier in the text.
-        Self {
-            anchor,
-            head,
-            old_visual_position: None,
+        if changes.is_empty() {
+            return self;
         }
+
+        let positions_to_map = match self.anchor.cmp(&self.head) {
+            Ordering::Equal => [
+                (&mut self.anchor, Assoc::After),
+                (&mut self.head, Assoc::After),
+            ],
+            Ordering::Less => [
+                (&mut self.anchor, Assoc::After),
+                (&mut self.head, Assoc::Before),
+            ],
+            Ordering::Greater => [
+                (&mut self.head, Assoc::After),
+                (&mut self.anchor, Assoc::Before),
+            ],
+        };
+        changes.update_positions(positions_to_map.into_iter());
+        self.old_visual_position = None;
+        self
     }
 
     /// Extend the range to cover at least `from` `to`.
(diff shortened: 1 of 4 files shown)


helix-core/src/selection.rs
```diff
@@ -450,18 +420,44 @@ impl Selection {
 
     /// Map selections over a set of changes. Useful for adjusting the selection position after
     /// applying changes to a document.
-    pub fn map(self, changes: &ChangeSet) -> Self {
+    pub fn map(mut self, changes: &ChangeSet) -> Self {
         if changes.is_empty() {
             return self;
         }
+        self = self.map_no_normalize(changes);
+        if self.ranges.len() > 1 {
```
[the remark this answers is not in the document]
pascalkuthe: This is existing code inclined from `Selection:;new` but I could change that in a seperate commit

dead10ck (commented):
This is awesome! I haven't taken a close look yet, but at first glance, it unfortunately seems to squarely conflict with #7269, which relies on the fact that each range is mapped individually. 🤔 Hopefully there's an alternative I can do.

pascalkuthe:
> This is awesome! I haven't taken a close look yet, but at first glance, it unfortunately seems to squarely conflict with #7269, which relies on the fact that each range is mapped individually. 🤔 Hopefully there's an alternative I can do.

I hadn't reviewed that PR yet but I took a look and don't really think it conflicts. You are only mapping ranges trough a changeset with a single change (and do the offset tracking yourself so it's already O(N)). I deleted `Range::map` function because there were no uses left but if that PR still needs it then I can keep it for your used case with a caveat that it's `O(N)` and therefore `O(MN)` in loops.

This OR doesn't really remove anything. It basically just changes `map_pos` to operate on a sorted list instead. Mapping a single position is still possible (in fact `mal_pos` still exists since it's used in a few placea, its trivial to implement using `once` and behaves the exactly the same as on master)

dead10ck:
Hm, yeah, I think you're right, other than `Range::map` getting removed, maybe nothing really changes. Even now, `Selection::map` isn't used in the code path of `insert_char` or `delete_char_backwards`—both still compute the resulting selection by iterating linearly over the ranges. So if I'm understanding this right, this change mostly affects the application of a change set to the document's text, rather than the computation of the transactions to begin with.

pascalkuthe:
Yeah this only affect comments which don't manually update the transaction but also diagnostic mapping and virtual text mapping. It does howbwr affect the "default" codepath of insert_char since that relies on the automatic range mapping instead of a manual implementation. 

Our two PRs are kind of similar in that you have also included a more efficent implementation of mapping changes trough transaction by accident (by storing an offset and calling map on a single element change set). They just affect different cosepaths

dead10ck:
This has made me realize that I think we're unnecessarily mapping the whole selection through the change set in the cases where the transaction explicitly specifies a new one.

https://github.com/helix-editor/helix/blob/18160a667be6ce30eb644dac2604d75fab9d63dc/helix-view/src/document.rs#L1070-L1085

We essentially do that whole mapping and then immediately throw all that work away.

helix-core/src/selection.rs
```diff
@@ -161,36 +161,6 @@ impl Range {
...
-        // even if the anchor and head haven't changed, because we don't
-        // know if the *visual* position hasn't changed due to
-        // character-width or grapheme changes earlier in the text.
-        Self {
-            anchor,
-            head,
-            old_visual_position: None,
-        }
-    }
-
```
dead10ck, line 193: This will need to be kept for #7269, but otherwise I think this looks good 👍

dead10ck (changes requested):

archseer:
> This has made me realize that I think we're unnecessarily mapping the whole selection through the change set in the cases where the transaction explicitly specifies a new one.

The mapping still needs to happen for other views than the current one. We could speed things up slightly by removing the current view's selection from the selection set before doing the mapping if the transaction contains a selection

helix-core/src/selection.rs
pascalkuthe: I restored the function, its actually benefits a tiny bit two (just transverse the changeset twice instead of once)

helix-core/src/selection.rs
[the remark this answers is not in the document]
pascalkuthe: done in 85dd01b9405b3ddbc39575e1a3b0a2da7c6d9347, this does feels like an improvement in general

pascalkuthe:
I fixed a small oversight that was causing test failures and addressed the comments. Should be ready now

dead10ck (dismissed):
🍓🫐🥞

helix-core/src/selection.rs
```diff
@@ -161,34 +161,35 @@ impl Range {
         self.from() <= pos && pos < self.to()
     }
 
-    /// Map a range through a set of changes. Returns a new range representing the same position
-    /// after the changes are applied.
-    pub fn map(self, changes: &ChangeSet) -> Self {
+    /// Map a range through a set of changes. Returns a new range representing
+    /// the same position after the changes are applied. Note that this
+    /// function runs in O(N) (N is number of changes) and can therefore
+    /// cause performance problems if run for a large number of rangs as the
```
A-Walrus, line 167:
```suggestion
    /// cause performance problems if run for a large number of ranges as the
```

helix-core/src/transaction.rs
```diff
@@ -326,17 +326,33 @@ impl ChangeSet {
         self.changes.is_empty() || self.changes == [Operation::Retain(self.len)]
     }
 
-    /// Map a position through the changes.
+    /// Map a *sorted* list of position through the changes.
```
A-Walrus, line 329:
```suggestion
    /// Map a *sorted* list of positions through the changes.
```

helix-core/src/transaction.rs
```diff
@@ -368,41 +380,48 @@ impl ChangeSet {
...
-                pos, old_pos
-            )
-        }
-        new_pos
+        panic!("Positions {out_of_bounds:?} are out of range for changeset len {old_pos}!",)
+    }
+
+    /// Map a position through the changes.
+    ///
+    /// `assoc` indicates which size to associate the position with. `Before` will keep the
```
A-Walrus, line 418:
```suggestion
    /// `assoc` indicates which side to associate the position with. `Before` will keep the
```

A-Walrus (commented):
Just some typos :)

helix-core/src/transaction.rs
```diff
@@ -326,17 +326,33 @@ impl ChangeSet {
...
     ///
-    /// `assoc` indicates which size to associate the position with. `Before` will keep the
-    /// position close to the character before, and will place it before insertions over that
-    /// range, or at that point. `After` will move it forward, placing it at the end of such
-    /// insertions.
-    pub fn map_pos(&self, pos: usize, assoc: Assoc) -> usize {
+    /// This is equivalent to updating each position with `map_pos`:
+    ///
+    /// ``` no-compile
+    /// for (assoc, pos) in positions {
```
the-mikedavis, line 334:
```suggestion
    /// for (pos, assoc) in positions {
```

to match `positions`

helix-core/src/transaction.rs
```diff
@@ -326,17 +326,33 @@ impl ChangeSet {
...
+    /// However this function is significantly faster running in `O(N+M)` instead of `O(NM)`
+    pub fn update_positions<'a>(&self, positions: impl Iterator<Item = (&'a mut usize, Assoc)>) {
         use Operation::*;
+
+        let mut positions = positions.peekable();
+        macro_rules! map {
+            ($map: expr) => {
+                loop {
+                    let Some((pos, assoc)) = positions.peek_mut() else { return; };
+                    let Some(new_pos) = $map(*assoc, **pos) else { break; };
```
the-mikedavis, line 347: very minor nit but I think that the order args of the function here would be more intuitive if they matched the tuple in `positions`:

```suggestion
                    let Some(new_pos) = $map(**pos, *assoc) else { break; };
```

just a small thing that tripped me up while reading

the-mikedavis (commented):
some very minor comments but otherwise this looks great, nice work!

helix-lsp/src/lib.rs
```diff
@@ -420,6 +420,7 @@ pub mod util {
             return transaction;
         }
```
archseer, line 422: Can you explain/add a comment why normalize isn't needed here? This is an additional invariant that we should document in places where no_normalize is used

helix-core/src/transaction.rs
pascalkuthe: done in force push to keep history clean

helix-core/src/transaction.rs
pascalkuthe: done in force push to keep history clean

helix-lsp/src/lib.rs
pascalkuthe: I added a comment but this was actually exiting behaviour. Really this isn't a selection we produce here at all, we need the mapped ranges for calculating the actual selecton below (which is normalized later). In the past I achieved that by calling `map_pos' on the ranges directly (below). Normalizing here not only has no point it would also break the association between tabstops and selections (if selections are merged/sorted). I just switched to the selection function here to also take advantage of the optimization here but really this is not a selection at all we are just updating a list of ranges.

the-mikedavis (approved):

the-mikedavis:
Github is acting strange but this is now merged onto master as d491e234f4eb4d8c3869f44ab71fedf022dc463e and b33516fb16df93572ac4ddec1d4079282a2b4b5a.

Issue #7355, opened 2023-06-17
Option to not centre the cursor when switching between buffers


DerekLeach:
When I switch between buffers the view is always moved up or down so that the cursor is vertically centred on the screen. When I rapidly flip back and forward between two buffers using gn and gp my eyes are anticipating a certain shape to the text on the screen but this is confused when the text keeps moving up or down a few lines so that it takes me a moment to reorient myself. This also happens when using the buffer picker.

It would be great if there were an option to keep the previous scroll location and not automatically centre the cursor vertically on the screen each time.

t-monaghan:
I've had an attempt at this issue with #7414 

I just noticed that @DerekLeach mentions having this as an option. My experience is that this ViewPosition centering adds friction to my code comprehenesion, and I would have the preference that restoring the ViewPosition is the default behaviour. 

Again, I'm very green in this area, so I would appreciate some guidance.

PR #7414, opened 2023-06-21
Replace_document_in_view() applies the last known view for documents 

t-monaghan:
This PR is to address issue #7355


helix-view/src/editor.rs
```diff
@@ -1249,17 +1247,27 @@ impl Editor {
         }
     }
 
-    fn replace_document_in_view(&mut self, current_view: ViewId, doc_id: DocumentId) {
+    fn replace_document_in_view(
+        &mut self,
+        current_view: ViewId,
+        doc_id: DocumentId,
+        from_scratch: bool,
```
t-monaghan, line 1254: I struggled to find a way to query if the current document was a scratch buffer.
However, the contexts in which this method is called have this knowledge.

the-mikedavis: Detecting if it is a scratch buffer could be done with `doc!(self).path().is_none()` but not all scratch buffers are removed, only ones that aren't modified. We just want to protect against using `doc_mut!` with a DocumentId that doesn't exist:

```rust
// If swapping from a document, record the last view position
if let Some(old_doc) = self.documents.get_mut(&view.doc) {
    old_doc.last_view = view.offset;
}
```

helix-view/src/document.rs
```diff
@@ -185,6 +186,8 @@ pub struct Document {
 
     // when document was used for most-recent-used buffer picker
     pub focused_at: std::time::Instant,
+
+    pub last_view: ViewPosition,
```
the-mikedavis, line 190: Minor nit but let's call this `last_view_position` since it's easy to confuse this as being the last View that focused the doc

dead10ck (dismissed):

helix-view/src/editor.rs
```diff
@@ -1251,15 +1249,18 @@ impl Editor {
...
+        }
+
+        view.doc = doc_id;
         let doc = doc_mut!(self, &doc_id);
+        view.offset = doc.last_view_position;
         doc.ensure_view_init(view.id);
         view.sync_changes(doc);
         doc.mark_as_focused();
-
-        align_view(doc, view, Align::Center);
```
the-mikedavis, line 1262: The align_view call here was previously ensuring that the cursor was in-view and we will still need to do that with this change:

```rust
// at the beginning of the function:
let scrolloff = self.config().scrolloff;
// ...
// at the end:
```

Updating the `view.offset` will adjust the scroll and that may leave the cursor outside the visible area. See https://asciinema.org/a/Uxh9FvcWFi2B26yYE2yb6Kfun: I leave the old view position of `README.md` at the end of the buffer, create a new split View and focus `README.md`. The scroll of the buffer is updated to the last view position but the cursor is at the beginning, so any movements (hjkl) or commands will move the viewport to the cursor.

pascalkuthe (changes requested):
This only works with a single view, not multiple views. Not only would they smudge `last_view_position` but the view offset can shift by various edits so that the position doesn't shift when the file is edited in another view. So the right way to fix this would be to store view offsets on the document per view and map them through changes. At that point, it doesn't make sense to differentiate between the last view position and the current view position. Instead, just store them all on the document. This is a general pattern we have been seeing (storing stuff on the document instead of view to map them through changes). Its sort of starting to bloat the document tough so we likely want some kind of shared `ViewData` struct for that. There was some discussion about that in #6198

t-monaghan:
@pascalkuthe I'm not quite across the intended behaviour for multiple views. Could I ask you to have a look at this recording and let me know if it demonstrates the smudging/shifting that you've raised? https://asciinema.org/a/8F4fvSJB4rY0tgtwJvXuJXUCy

pascalkuthe:
the intended behavoiur is that every view has its own position in the document. Basically views should not change others positions. so if I have file X open in views A and B at positions F and G respectively and switch to file Y in both views and then bach to X  then A should endup at position F and B at position G again.

the-mikedavis:
Let's close this in favor of #7568

PR #7417, opened 2023-06-21
Fix tree sitter chunking

A-Walrus:


pascalkuthe (commented):
The fix for TSchunk access  is correct and needed 👍  I am not sure sure about the byte alignment in the grapheme functions. Where would that be needed? It's pretty slow and I can't think of a scenario where we would want to align invalid bytes?

A-Walrus:
>Where would that be needed? It's pretty slow and I can't think of a scenario where we would want to align invalid bytes?

Currently they're necessary to stop helix from crashing in certain situations, but I think they might just be a bandaid solution and not a fix for the actual bug...
Marking as draft while I try to dig a bit deeper.

pascalkuthe:
If you have a reproduction case with a tracebacj that would be great. Maybe a buggy TS grammar could cause this. I actually want to remove the manual grapheme alignment we currently do for TS anyway. The doc formatter can do the byte/char/grapheme alignment on the fly which is a nice perf bump for long lines

pascalkuthe:
A thanks yeah I saw the issue now. Let ne see if I can comeuo with a quick fix.

pascalkuthe:
Ah yeah it's related to what I meant. It shouldn't be too hard to track the byte pos while rendering too and simply use byte indecies in the entire syntax highlighting infrastructure. That would fix some inefficencies jn the rendering. I have been plannjng that for a while but didnt get around to it yet. If you are interested you can take a crack at this but it would be a bit more invokved.

IMO it's still a bug in the C TS grammar the the byte index is not byte aligned tough 🤔 Unaligned bytes make sense as an input (since TS doesn't know ahead if time where char boundaries are) but not as an output (a highlighting starting in the middle of a Unicode codepoint makes no sense)

A-Walrus:
I got rid of the bandaid and left only the relevant fix. Will file a bug report with the language servers, and see if I can rework syntax highlighting to use bytes everywhere

the-mikedavis (approved):

Issue #7424, opened 2023-06-22
Default configuration file?


qtfkwk:
* Searched and did not find any default configuration file in the repository nor is one created when helix runs the first time; did not find any issue mentioning a default configuration file; afaik, there is no option to dump one, etc
* A default configuration file is useful because it provides a starting point to customize the editor, exposes all options upfront, is offline, and so on
* The alternative to providing a default configuration file is to consult / search the multi-page [Configuration section of the documentation](https://docs.helix-editor.com/configuration.html) and cherry-pick the ones I'm interested in right now, then repeat as necessary
* I do love helix and appreciate the work you've done to make such an awesome editor, also specifically the many sane default settings as well as helix's ability to configure within a session!

Until an official default configuration file can be provided somewhere better like in the repository or perhaps in or linked to from the documentation... I hope this issue can serve as a good place to start.

I've cobbled the following from the current documentation as of yesterday. It is likely not 100%. It does not cover key remapping for example. It also does not provide a list of themes (which can be easily found via `ls ~/.config/helix/runtime/themes` btw). I've tried to capture the range of possible options and annotate the default option with an asterisk in comments, except where the value is an integer or boolean. If you want to capture your specific diffs, I'd recommend commenting the default line and adding yours, or just change the values and diff your file with a copy of the original. Also *theoretically*, adding the below as your `~/.config/helix/config.toml` *should* provide *zero changes* from helix's default behavior but this has not yet been officially confirmed.

```toml
theme = "default"

[editor]
scrolloff = 5
mouse = true
middle-click-paste = true
scroll-lines = 3
shell = ["sh", "-c"] # Unix
#shell = ["cmd", "/C"] # Windows
line-number = "absolute" # absolute*, relative
cursorline = false
cursorcolumn = false
auto-completion = true
auto-format = true
auto-save = false
idle-timeout = 400
completion-trigger-len = 2
completion-replace = false
auto-info = true
true-color = false
undercurl = false
rulers = []
bufferline = "never" # never*, always, multiple
auto-pairs = true

[editor.statusline]
left = ["mode", "spinner", "file-name"]
center = []
right = ["diagnostics", "selections", "position", "file-encoding"]
separator = "│"
mode.normal = "NOR"
mode.insert = "INS"
mode.select = "SEL"

[editor.lsp]
enable = true
display-messages = false
auto-signature-help = true
display-inlay-hints = false
display-signature-help-docs = true
snippets = true
goto-reference-include-declaration = true

[...]

PR #7425, opened 2023-06-22
add default configuration file

qtfkwk:
See https://github.com/helix-editor/helix/issues/7424


CptPotato:
I think a command cli argument to export the default/current configuration could be a better alternative.

With a plain file there is no guarantee that the default config actually is the same as the file and I think it's easy to miss/break.

archseer:
https://github.com/helix-editor/helix/issues/1210#issuecomment-983629067

PR #7429, opened 2023-06-22
Allow any indent size from 1 to 16

msdrigg:
Previously, helix only allowed indent sizes up to 8. This commit changes to allow any indent sizes that fit within a u8.

This pr is motivated by a hackernews user with a ~deranged~ normal desire to have a 12 space indent.

See request here: https://news.ycombinator.com/item?id=36429221

Code motivated by suggestions here: https://news.ycombinator.com/item?id=36432996

I used a single line of 256 spaces because it's super simple and easy to validate rather than using an additional dependency like const_format or writing a macro myself.

### DIFF
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -19,6 +19,9 @@ pub enum IndentStyle {
     Spaces(u8),
 }
 
+// 256 spaces
+const INDENTS: &str = "                                                                                                                                                                                                                                                                ";
+
 impl IndentStyle {
     /// Creates an `IndentStyle` from an indentation string.
     ///
@@ -27,7 +30,7 @@ impl IndentStyle {
     #[inline]
     pub fn from_str(indent: &str) -> Self {
         // XXX: do we care about validating the input more than this?  Probably not...?
-        debug_assert!(!indent.is_empty() && indent.len() <= 8);
+        debug_assert!(!indent.is_empty() && indent.len() <= 256);
 
         if indent.starts_with(' ') {
             IndentStyle::Spaces(indent.len() as u8)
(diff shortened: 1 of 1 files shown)


kirawi:
https://github.com/helix-editor/helix/pull/7178 the limit is 16, so the static string should be limited to that.

pascalkuthe:
That's tab width and not indent width. That said I still think that a value of 16 is large enough.

helix-core/src/indent.rs
```diff
@@ -40,20 +43,13 @@ impl IndentStyle {
...
-            IndentStyle::Spaces(1) => " ",
-            IndentStyle::Spaces(2) => "  ",
-            IndentStyle::Spaces(3) => "   ",
-            IndentStyle::Spaces(4) => "    ",
-            IndentStyle::Spaces(5) => "     ",
-            IndentStyle::Spaces(6) => "      ",
-            IndentStyle::Spaces(7) => "       ",
-            IndentStyle::Spaces(8) => "        ",
-
+            IndentStyle::Spaces(n) if n <= 16 && n >= 1 => &INDENTS[0..n as usize],
```
pascalkuthe, line 46: The last two branches can be merged by just unconditionally clamping. That fixes the clippy libt too that makes CI fails

msdrigg:
Okay, all suggestions fixed and clippy should be happy. I also added a constant so it's clear why the literals were chosen, but I can put it back to literals if it's preferred. 

Also, I didn't see a good way to test the failing case where indent > 16 specifically because behavior could vary between debug and release builds (debug will panic for indent > 16 while release will pass with indent = 16)

PR #7430, opened 2023-06-22
allow HELIX_CONFIG override

dis-da-moe:
- as the name says


kirawi:
Conflicts with https://github.com/helix-editor/helix/pull/2135

PR #7431, opened 2023-06-23
Mark buffers created from stdin as modified

### LINKED ISSUE #3305: Initial scratch buffer is discarded when non-empty and unmodified

the-mikedavis:
This resolves some confusing behavior where a scratch document created by piping into hx is discarded when navigating away from that document.

We discard any scratch documents that are not modified and the original `Editor::new_file_from_stdin` would create unmodified documents. We refactor this function to create an empty document first and then to apply the text from stdin as a change.

This comes out of the discussion in #3306
Closes #3306 
Closes #3305

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -397,33 +397,11 @@ pub fn from_reader<R: std::io::Read + ?Sized>(
     let mut buf_out = [0u8; BUF_SIZE];
     let mut builder = RopeBuilder::new();
 
-    // By default, the encoding of the text is auto-detected by
-    // `encoding_rs` for_bom, and if it fails, from `chardetng`
-    // crate which requires sample data from the reader.
-    // As a manual override to this auto-detection is possible, the
-    // same data is read into `buf` to ensure symmetry in the upcoming
-    // loop.
-    let (encoding, has_bom, mut decoder, mut slice, mut is_empty) = {
-        let read = reader.read(&mut buf)?;
-        let is_empty = read == 0;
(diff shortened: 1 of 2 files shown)


helix-view/src/document.rs
```diff
@@ -493,6 +471,81 @@ pub fn from_reader<R: std::io::Read + ?Sized>(
     Ok((rope, encoding, has_bom))
 }
 
+pub fn read_to_string<R: std::io::Read + ?Sized>(
```
the-mikedavis, line 474: This would also be useful for #3966

PR #7435, opened 2023-06-23
Cursor color is changed according to editor mode.

fvacek:
The same color as status-line are used for normal and insert modes


the-mikedavis:
What do you think about these colors (and cursor colors in general) @enkodr?

These changes could probably be applied to the other ayu themes as well.

the-mikedavis:
I will close this out as it's gone stale. Feel free to open another PR if you come back around to this change

Issue #3007, opened 2022-07-08
Add a variant of sub-selection (`s`) that uses textobjects instead of regex


EpocSquadron:
I think it could be very ergonomic to be able to have a selection broken down by tree-sitter driven textobjects that can give us commonly desired selections.

Consider a selection like the following:

```php

//                  [                   selection                           |]
function doSomething($firstParam, MyTypeHint $secondParam, $thirdParam = null) {
    // do something
}
```

If you want to select all parameters to the function, say so you could reorder them with `Alt-)` or append a new line to each to begin reformatting them, you would need to split on spaces with `S <enter>`, but this will incorrectly split the selection where the default value is defined, and where the typehint is included.

With this proposal, you could enter the new subselection mode (keybind to be determined), and enter `a` meaning the textobject for parameters/arguements, resulting in the following selections:

```php

//                   [   sel1  |] [         sel2        |] [      sel3     |]
function doSomething($firstParam, MyTypeHint $secondParam, $thirdParam = null) {
    // do something
}
```

Similar to the `mi`/`ma` and `]`/`[` binds, this new bind would show an autoinfo of possible textobjects, rather than prompting for a regex as the current `s` and `S` keybind does.

CBenoit:
Hi @the-mikedavis 
FYI I was thinking about implementing this if it’s okay (I see you assigned yourself for #3263)

the-mikedavis:
Sweet! I think the changes for #3263 will end up being separate from this so I don't think we'll step on each others' toes 🙂

EpocSquadron:
For keybinds, I wonder if the existing behavior of `s`/`S` can be moved to the autoinfo alongside the textobjects, with `/` being the triggering option.

pascalkuthe:
I would love to see this, @CBenoit did you make any progress on this otherwise I might take this on soonish (altough I quite a few other things I want to finish first)

> For keybinds, I wonder if the existing behavior of `s`/`S` can be moved to the autoinfo alongside the textobjects, with `/` being the triggering option.

I would rather see this as a seperate keybinding. `s` is likely one of the most common keybindings I use in helix, it's so insanely useful and I would want to keep it as a single-key hotkey. Perhaphs an uppercase version might be an option: `mI`/`mA`  uppercase characters for alternative behaviour seems to be a pattern we are moving towrads in general it sems. What do you think @the-mikedavis?

EpocSquadron:
[...]

PR #7445, opened 2023-06-25
feat(command): select all siblings and children

dead10ck:
This adds two new related commands:

### `select_all_siblings`

Selects all siblings of the node under the cursor.

[asciicast](https://asciinema.org/a/yt7SserYwGL7QJ3UGVAmhe60t)

### `select_all_children`

Selects all children of the node under the cursor.

[asciicast](https://asciinema.org/a/GiMx8a5uZMtbyXI0Zh8Z6CoDh)

### `select_all_children_in_selection`

~~Selects all children of the node under the cursor, but only if the children are contained inside the starting selection.~~

Edit: saving this one for a future PR

### DIFF
--- a/helix-core/src/object.rs
+++ b/helix-core/src/object.rs
@@ -1,4 +1,5 @@
-use crate::{syntax::TreeCursor, Range, RopeSlice, Selection, Syntax};
+use crate::{movement::Direction, syntax::TreeCursor, Range, RopeSlice, Selection, Syntax};
+use tree_sitter::{Node, Tree};
 
 pub fn expand_selection(syntax: &Syntax, text: RopeSlice, selection: Selection) -> Selection {
     let cursor = &mut syntax.walk();
@@ -37,6 +38,65 @@ pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selectio
                 break;
             }
         }
+    })
+}
+
+fn find_parent_with_more_children(mut node: Node) -> Option<Node> {
+    while let Some(parent) = node.parent() {
+        if parent.child_count() > 1 {
+            return Some(parent);
+        }
+
+        node = parent;
+    }
+
+    None
+}
+
+pub fn select_all_siblings(tree: &Tree, text: RopeSlice, selection: Selection) -> Selection {
+    let root_node = &tree.root_node();
+
+    selection.transform_iter(|range| {
+        let from = text.char_to_byte(range.from());
+        let to = text.char_to_byte(range.to());
+
+        root_node
+            .descendant_for_byte_range(from, to)
+            .and_then(find_parent_with_more_children)
+            .map(|parent| select_children(parent, text, range.direction()))
+            .unwrap_or_else(|| vec![range].into_iter())
+    })
+}
+
+fn select_children(
+    node: Node,
+    text: RopeSlice,
+    direction: Direction,
+) -> <Vec<Range> as std::iter::IntoIterator>::IntoIter {
+    let mut cursor = node.walk();
+
+    node.named_children(&mut cursor)
+        .map(|child| {
+            let from = text.byte_to_char(child.start_byte());
+            let to = text.byte_to_char(child.end_byte());
+
+            if direction == Direction::Backward {
+                Range::new(to, from)
+            } else {
+                Range::new(from, to)
+            }
+        })
+        .collect::<Vec<_>>()
+        .into_iter()
+}
+
+fn find_sibling_recursive<F>(node: Node, sibling_fn: F) -> Option<Node>
+where
+    F: Fn(Node) -> Option<Node>,
+{
+    sibling_fn(node).or_else(|| {
+        node.parent()
+            .and_then(|node| find_sibling_recursive(node, sibling_fn))
     })
 }
 
--- a/helix-core/src/object.rs
+++ b/helix-core/src/object.rs
@@ -1,5 +1,4 @@
-use crate::{movement::Direction, syntax::TreeCursor, Range, RopeSlice, Selection, Syntax};
-use tree_sitter::{Node, Tree};
+use crate::{syntax::TreeCursor, Range, RopeSlice, Selection, Syntax};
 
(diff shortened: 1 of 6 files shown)


pascalkuthe:
after trying this out for a bit I think `select_all_children` would benefit a lot from only selecting children that are actually contained in the selection. The current implementation is also useful but i found that quite a lof of nodes have children I don't care about. If someone really wants all children inside the node they can do a-o first. That change would also make the command more predictable I think

dead10ck:
> after trying this out for a bit I think `select_all_children` would benefit a lot from only selecting children that are actually contained in the selection.

Hmm, but in this case, the command would basically never do anything when the selection is just a cursor, or really any time you don't happen to have an entire node contained in the selection.

But on the other hand, I could see this being useful for if you only want a subset of the children. I'll add a third command that does this.

dead10ck:
Just added `select_all_children_in_selection` that only selects if all resulting ranges are contained within the starting selection.

helix-core/src/object.rs
```diff
@@ -40,6 +41,85 @@ pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selectio
     })
 }
 
+fn find_parent_with_more_children(mut node: Node) -> Option<Node> {
```
pascalkuthe, line 44: I think thus also needs to be able to cross injection layer so I would use the new tree cursor here

helix-core/src/object.rs
```diff
@@ -40,6 +41,85 @@ pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selectio
...
+            } else {
+                Range::new(from, to)
+            }
+        })
+        .collect::<Vec<_>>();
+
+    if !children.is_empty() {
+        Some(children.into_iter())
+    } else {
+        None
```
pascalkuthe, line 109: Dame here this could be an injection that aus names children that are inside the injection

pascalkuthe (commented):
This should be Adapter to cross tree sitter injection layer like the either motions now do rest lgtm

dead10ck:
Sounds good, will do

helix-term/src/commands.rs
```diff
@@ -438,8 +438,11 @@ impl MappableCommand {
         reverse_selection_contents, "Reverse selections contents",
         expand_selection, "Expand selection to parent syntax node",
         shrink_selection, "Shrink selection to previously expanded syntax node",
-        select_next_sibling, "Select next sibling in syntax tree",
-        select_prev_sibling, "Select previous sibling in syntax tree",
+        select_next_sibling, "Select next sibling in the syntax tree",
+        select_prev_sibling, "Select previous sibling the in syntax tree",
+        select_all_siblings, "Select all siblings of the current node",
+        select_all_children, "Select all children of the current node",
+        select_all_children_in_selection, "Select all children of the current node that are contained in the current selection",
```
pascalkuthe, line 445: this reminds me of #3007, I would like to potentially see this as somekind of mode (potentially under `ms..`, `mk..`, `mr..`, `mS..` as @the-mikedavis suggested). Then would could have a default binding for this in the future (doesn't need to happen in this pr )

helix-term/src/commands.rs
```diff
@@ -4974,6 +4977,52 @@ pub fn extend_parent_node_start(cx: &mut Context) {
...
+fn select_all_children_in_selection(cx: &mut Context) {
+    let motion = |editor: &mut Editor| {
+        select_all_impl(editor, |syntax, text, selection| {
+            let all_children = object::select_all_children(syntax, text, selection.clone());
+
+            if selection.contains(&all_children) {
+                all_children
+            } else {
+                selection
+            }
```
pascalkuthe, line 5011: Hmm I am not sure this is quite what I had in  mind here. If I can tell this right this would just not do anything as soon as any child is not in the  selection. Instead I would like to see this be sort of like anl intersection. We should select the children covered by a selection.

In pseudo code `all_chidren.retain(|range| selection.contains(range)` (just better implemented so its not `O(n^2)`.

If this is too complex just omit this and I think I can implement this when solving #3007

helix-core/src/object.rs
```diff
@@ -40,6 +40,48 @@ pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selectio
...
+    let children = cursor
+        .named_children()
+        .map(|child| Range::from_node(child, text, direction))
+        .collect::<Vec<_>>();
+
+    if !children.is_empty() {
+        Some(children.into_iter())
+    } else {
+        None
+    }
```
pascalkuthe, line 82: Every callsite has a fallback so lets just inline that code here to avoid duplication

```suggestion
    if children.is_empty() {
        vec![fallback]
    } else {
        chidren
    }
```

helix-core/src/object.rs
```diff
@@ -40,6 +40,48 @@ pub fn select_next_sibling(syntax: &Syntax, text: RopeSlice, selection: Selectio
...
+        select_children(&mut cursor, text, range.direction())
+            .unwrap_or_else(|| vec![range].into_iter())
+    })
+}
+
+fn select_children<'n>(
+    cursor: &'n mut TreeCursor<'n>,
+    text: RopeSlice,
+    direction: Direction,
+) -> Option<<Vec<Range> as std::iter::IntoIterator>::IntoIter> {
```
pascalkuthe, line 72: I would just return a Vec and let the callsite call `into_iter` or if you really want to return an iterator use an anonymous `impl Iterator` instead this is a bit hard to read

helix-term/src/commands.rs
dead10ck: Hmm, maybe I'll just leave it out for now

dead10ck:
Good suggestions, I took care of them.

pascalkuthe (approved):

the-mikedavis (approved):
Looks great! Sorry for throwing a wrench into this with the TreeCursor changes - thanks for rebasing and improving the TreeCursor while you were at it

sudormrfbin:
I can finally stop splitting by commas inside a function call to select the arguments :) But it seems the docs weren't updated with the new `alt-a` and `alt-I` keybinds.

PR #7457, opened 2023-06-26
add:translate

ptechen:


the-mikedavis:
See https://github.com/helix-editor/helix/discussions/7336. This is better suited as a plugin.

ptechen:
> See #7336. This is better suited as a plugin.

pipe is inconvenient!

pascalkuthe:
you can bind `:pipe` to a key. You can achieve the exact same thing as this PR already for example using `curl`. This kind of thing is just entirely out of scope for an editor. Not even large ides like vscode or intellij include something like this in core. You would need to make the language configurable too. What you implemented is hardcoded for a single language. All that adds a ton of complexity that just doesn' t need to exist in core

PR #7467, opened 2023-06-27
LSP: Discard publishDiagnostic from uninitialized servers

the-mikedavis:
The spec explicitly disallows publishDiagnostic to be sent before the initialize response:

> ... the server is not allowed to send any requests or notifications to the client until it has responded with an InitializeResult ...

(https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#initialize)

But if a non-compliant server sends this we currently panic because we '.expect()' the server capabilities to be known to fetch the position encoding. Instead of panicking we can discard the notification and log the non-compliant behavior.

See #7466

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -746,7 +746,12 @@ impl Application {
                                 return;
                             }
                         };
-                        let offset_encoding = language_server!().offset_encoding();
+                        let language_server = language_server!();
+                        if !language_server.is_initialized() {
+                            log::error!("Discarding publishDiagnostic notification sent by an uninitialized server: {}", language_server.name());
+                            return;
+                        }
+                        let offset_encoding = language_server.offset_encoding();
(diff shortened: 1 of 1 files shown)


pascalkuthe:
Are there any other cases where we expect the capabilities? I think k I saw that a few times and that could cause si.klar bugs (diagnostics is likely just the most common)

the-mikedavis:
It looks like this is actually the only place in `handle_language_server_message` that we assume we have the capabilities. Though it looks like we should be checking the capabilities when applying workspace edits. Currently it's hard-coded: https://github.com/helix-editor/helix/blob/d8f9b901ddb1d91bf3ade66cf091e93b5cab6e49/helix-term/src/application.rs#L1045-L1049

I'll make a PR for that too

pascalkuthe:
oh yeah, that should definitely be fixed. I remember fixing this elsewhere too but I must have missed that. Good catch!

pascalkuthe (approved):
thanks!

PR #7469, opened 2023-06-27
LSP: Use negotiated position encoding for workspace edits

the-mikedavis:
Previously this was hard-coded to UTF-8 but we might have negotiated another position encoding.


pascalkuthe (approved):
thanks!

PR #7471, opened 2023-06-27
correctly map unsorted positions

### LINKED ISSUE #7468: Panic on weird n" in a Rust file
### Summary

I was writing the following line in Rust:

```
writeln!(writer, "{}\n", positions.len())?;
```

After entering `\n`, the editor crashed!

The panic is from the following line:

https://github.com/helix-editor/helix/blob/d8f9b901ddb1d91bf3ade66cf091e93b5cab6e49/helix-core/src/transaction.rs#L413

### Reproduction Steps

I found a minimal (and very weird) example to reproduce this bug:

1. Run the commands
    1. `cargo new project`
    2. `cd project`
    3. `echo 'n"' > src/main.rs`
    4. `hx src/main.rs`
2. Wait for the language server to start.
3. Press `c` over the character `n`  to change it.
4. Helix crashes with the following error:
    ```
    thread 'main' panicked at 'Positions [(1, After), (1, After)] are out of range for changeset len 2!', /var/home/mo/code/compile/helix/helix-core/src/transaction.rs:413:9
    ```

### Helix log

```
2023-06-27T16:20:03.976 helix_view::document [WARN] LSP formatting failed: protocol error: ServerError(-32900): rustfmt exited with:
                           Status: exit status: 101
                           stdout: 
                           stderr: error: prefix `n` is unknown
 --> <stdin>:1:1
  |
1 | n"
  | ^ unknown prefix
  |
  = note: prefixed identifiers and literals are reserved since Rust 2021
help: consider inserting whitespace here
  |
1 | n "
  |  +

error[E0765]: unterminated double quote string
 --> <stdin>:1:2
  |
1 | n"
  |  ^^
```

</details>

### Platform

Linux

### Terminal Emulator

Alacritty 0.12.1

### Helix Version

helix 23.05 (d5111222)

### LINKED ISSUE #7470: Panic with Positions [..] are out of range ...
### Summary

```
thread 'main' panicked at 'Positions [(6315, After), (6348, After), (6481, After), (6481, After), (6496, After), (6522, After), (7903, After), (7907, After), (8067, After), (8067, After), (9463, After), (9467, After), (9523, After), (9523, After), (9601, After), (9605, After), (10170, After), (10170, After), (10942, After), (11508, After), (11508, After), (11508, After), (11508, After), (12630, After), (12634, After), (12898, After), (12898, After), (12968, After), (13344, After), (13348, After), (13380, After), (13380, After), (14390, After), (14394, After), (14657, After), (14657, After), (14727, After), (15289, After), (15293, After), (15355, After), (15359, After), (15727, After), (15727, After), (15974, After), (15974, After), (16280, After), (16284, After), (16340, After), (16344, After)] are out of range for changeset len 23220!', /home/xguerin/Sandbox/helix/helix-core/src/transaction.rs:413:9
```

Since: d491e234, still there after 8d39a81a.

### Reproduction Steps

I wish I knew!

### Helix log

No relevant entries in the log.

### Platform

Linux

### Terminal Emulator

iTerm2 stable over tmux

### Helix Version

master at d491e234, still there at d8f9b901

pascalkuthe:
Fixes #7468
Fixes #7470

#7408 caused cashes because it assumed that all positions that are mapped with `update_position` are sorted. This is almost always the case. Except for the **end** of diagnostics. Diagnostics are sorted by their start position (and only if start positions are equal by end position). That caused various crashes. This is not a problem for selections since they are non-overlapping so ends re always sorted. 

This PR addresses that by allowing `update_positions` to backtrack in case if encounters an unsorted position. That does mean that the position mapping isn't quite linear anymore for diagnostics ends. However, this shouldn't really be an issue in practice as most diagnostics don't overlap, so the performance impact should be quite small in practice.

### DIFF
--- a/helix-core/src/transaction.rs
+++ b/helix-core/src/transaction.rs
@@ -326,7 +326,7 @@ impl ChangeSet {
         self.changes.is_empty() || self.changes == [Operation::Retain(self.len)]
     }
 
-    /// Map a *sorted* list of positions through the changes.
+    /// Map a (mostly) *sorted* list of positions through the changes.
     ///
     /// This is equivalent to updating each position with `map_pos`:
     ///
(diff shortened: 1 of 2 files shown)


helix-core/src/transaction.rs
```diff
@@ -335,27 +335,63 @@ impl ChangeSet {
     ///     *pos = changes.map_pos(*pos, assoc);
     /// }
     /// ```
-    /// However this function is significantly faster running in `O(N+M)` instead of `O(NM)`
+    /// However this function is significantly faster for sorted lists running
+    /// in `O(N+M)` instead of `O(NM)`. This function also handles unsorted/
+    /// partially sorted lists. However, in that case worst case complexity is
+    /// again `O(MN)`.  For lists that are often/mostly sorted (like the end of diagnostic ranges)
+    /// performanc is usally close to `O(N + M)`
```
the-mikedavis, line 342:
```suggestion
    /// performance is usally close to `O(N + M)`
```

Issue #7453, opened 2023-06-26
sh command doesn't handle arguments with spaces


dlip:
If i try to run a command with spaces, even when quoted correctly eg. `:sh cat 'test file.md'`

I get an error as it was received as 2 arguments 
```
Async job failed: Shell error: cat: test: No such file or directorycat: file.md: No such file or directory
```

cgahr:
Possibly duplicate of #7005, #5779

PR #7473, opened 2023-06-28
Fix unquoting arguments before passing to shell

nuid64:
#7453
Now quotes all arguments passed to the shell

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2154,10 +2154,7 @@ fn run_shell_command(
     }
 
     let shell = cx.editor.config().shell.clone();
-    let args = args.into_iter()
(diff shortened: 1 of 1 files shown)


nuid64:
Do we want " act same as '?

pascalkuthe:
Mike was just giving an example. There are lots of similar edge cases. I was about to make a similar comment. Simply forwarding the arguments to the shell unchanged (with the minor exception I mentioned above) is the only consistent and intuitive solution IMO

nuid64:
We could just not trim the quotes when parsing shellwords. Is it a good solution?

the-mikedavis:
No, the quotes should be dropped where shellwords are actually used. For example `:open "a b.txt"` should open a file `a b.txt` but with that change it would open `"a b.txt"` (where the quotes would be part of the filename).

Issue #7045, opened 2023-05-14
Configurable placeholder for snippets


kirawi:
Currently, placeholders aren't included when a completion is confirmed. It feels a little awkward, at least In Rust, since you can't use `]a` or `[a` to write out the arguments.

pascalkuthe:
snippet placeholders are never meant to be inserted into the document.  In fact the plan is to replace them with virtual text in the future. Snippet placeholders are intended as text that gives you a hint what text should be inserted there, not a form of autocompletion.

Placeholders are act as jump points that you can reach while pressing `tab`(and `s-tab`) so that will cover the jump between usecase once implemented. At that point having this option doesn't make sense IMO and I don't think we should add a new config option that would be useless once we have better snippet support

kirawi:
That's interesting. But at least locally, VSCode does keep the placeholders. Would you have any idea why?

PR #7480, opened 2023-06-29
Add an editor option for keeping completion placeholders.

gibbz00:
I keep finding myself having to retype them again.

Possibly relevant commit: https://github.com/helix-editor/helix/commit/ba24cfe9125eda97346e3ceee42686fb9f46046f

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -55,6 +55,7 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `preview-completion-insert` | Whether to apply completion item instantly when selected | `true` |
 | `completion-trigger-len` | The min-length of word under cursor to trigger autocompletion | `2` |
 | `completion-replace` | Set to `true` to make completions always replace the entire word and not just the part before the cursor | `false` |
+| `completion-keep-placeholders` | Set to `true` to keep snippet placeholders when accepting completions. | `false` |
 | `auto-info` | Whether to display info boxes | `true` |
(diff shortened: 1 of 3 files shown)


pascalkuthe:
see https://github.com/helix-editor/helix/issues/7045

gibbz00:
I did see some PR/issues regarding completion handling, and I do understand that support isn't entirely there yet. Ideally, I would love to be able to have a keybinding that it inserts the virtual placeholder text before "tabbing" to the next placeholder.

Right now, however; all placeholders are removed as soon as the first argument is being filled in, which kind of defeats the purpose of having them in the first place.

It is also a lot more efficient for me to retype the arguments that don't have variables of the same name already in scope, than to retype those that do. This was the primary motivation for this PR.

https://github.com/helix-editor/helix/assets/31696304/a9e3b21f-68aa-4b22-9bf3-eae0e05f73df

Now, I don't want to say that one way is better than the other, tut this gives users at the very least a choice until the underlying completion system is fledged out.

AloeareV:
I am enthusiastically in favor if this feature!

So1aric:
Yep, this is also helpful when writing C code. For example, when you define a prototype in a header file and go to a C file to implement it, Clang will give a snippet which is pretty helpful. However, Helix will then delete it 😢

gibbz00:
@pascalkuthe what's the status on this PR?

the-mikedavis:
#9801 is the proper fix for this

PR #7488, opened 2023-06-30
docs: align content with parent paragraph

tshepang:


pascalkuthe:
Markdown section content is typically not indented. Especially because this indicates a quote in some markdown flavours

the-mikedavis:
It looks like mdbook does the right thing with the indentation:

| Before | After |
|---|---|
| before | after |

I'm pretty sure the github markdown renderer has the same indentation behavior here

the-mikedavis (approved):

pascalkuthe (approved):

PR #7491, opened 2023-06-30
Update tree-sitter-matlab

acristoffers:
In matlab, it seems, `true` and `false` are also functions and `true(2)` is something valid. Fixed that.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2285,7 +2285,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "matlab"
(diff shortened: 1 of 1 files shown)


acristoffers:
Also fixed missing token in switch statement.

acristoffers:
Sorry for force-pushing twice. I found 3 bugs when trying the parser in an external project. I'm done for now, however.

pascalkuthe (approved):

the-mikedavis (approved):

PR #7492, opened 2023-06-30
Add .cppm file type to cpp language configuration

hazeycode:
The .cppm extension is used for C++2x modules


the-mikedavis:
More context: https://gcc.gnu.org/onlinedocs/gcc/C_002b_002b-Modules.html

the-mikedavis (approved):

pascalkuthe (approved):

Issue #7502, opened 2023-07-01
Register `infobox` value only updated once.


gibbz00:
### Summary

[Kooha-2023-07-01-20-45-03.webm](https://github.com/helix-editor/helix/assets/31696304/ec0e5f6e-c79f-42a1-9e40-cd4e796df12c)

### Platform

Linux

### Terminal Emulator

foot

### Helix Version

master

gabydd:
Duplicate of https://github.com/helix-editor/helix/issues/5241

PR #7503, opened 2023-07-01
Fix: #7502

gibbz00:
The fix itself was a one-liner, but I added a test for it anyway. I'm interested in working on register histories, so I might as well get started with setting up the necessary test scaffolding here.

The test was placed in `info.rs` because `Register` is in core, which can't have `helix-view` as a dependency. Not being able to place it in `registers.rs` forced me to make the `inner` field public. Maybe we should discuss if it would be appropriate to move `Register`(s) to `helix-view`.

### DIFF
--- a/helix-core/src/register.rs
+++ b/helix-core/src/register.rs
@@ -1,6 +1,6 @@
 use std::collections::HashMap;
 
-#[derive(Debug)]
+#[derive(Debug, Clone)]
 pub struct Register {
     name: char,
     values: Vec<String>,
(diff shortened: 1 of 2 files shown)


gabydd:
Duplicate of https://github.com/helix-editor/helix/pull/5242 also see Mike's comments

gibbz00:
Ahh, my mistake, I misinterpreted what he meant at first. Closing in favor of sending in another PR that can leverage the other register PRs I've sent in.

PR #7506, opened 2023-07-01
Move `Registers` to `helix-view`

gibbz00:
Depends on: https://github.com/helix-editor/helix/pull/7505

Motivation: https://github.com/helix-editor/helix/pull/7503#issue-1784173761

PR #7512, opened 2023-07-02
Typeset registers

gibbz00:
Depends on: https://github.com/helix-editor/helix/pull/7506

I know https://github.com/helix-editor/helix/pull/6985 exists, and I open to integrating it into the PRs I'm planning to send for multiple registers functionality. Interesting things that I found out from doing this alone: 

* Undocumented/unused: 
https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5480
* Undocumented: 
https://github.com/helix-editor/helix/blob/a9849ebee41a8cb884a9351d36ebd399adb4111e/helix-term/src/commands.rs#L5204


the-mikedavis (changes requested):
Chars are the correct representation for registers, see https://github.com/helix-editor/helix/pull/6820. The registers mentioned in the source are just defaults: there isn't really a concept of a "search register" for example, just a default char we use in search commands.

helix-view/Cargo.toml
```diff
@@ -27,6 +27,10 @@ helix-vcs = { version = "0.6", path = "../helix-vcs" }
 once_cell = "1.18"
 url = "2"
 
+strum = "0.25"
+strum_macros = "0.25"
+derive_more = "0.99"
```
archseer, line 32: Agree with @the-mikedavis, the benefits here are dubious and this just inflates the code size, adding to build time with three compile time proc macro dependencies (that import yet more)

archseer (changes requested):

Issue #6752, opened 2023-04-14
'Position ... is out of range for changeset len ....!' panic on ~23.03 when saving a file


bcspragu:
### Summary

Helix panicked on a recent build while saving a file, with error:

```
thread 'main' panicked at 'Position 6234 is out of range for changeset len 5108!', helix-core/src/transaction.rs:399:13
```

I was saving some combination of Go + Vue 2 files at the time

### Reproduction Steps

Unfortunately I don't have a reproducible use case, this is the first panic I've had since updating to something around 23.03

### Helix log

The `helix.log` didn't have anything useful, just LSP-related errors that aren't actually problems.

### Platform

Linux

### Terminal Emulator

alacritty / tmux

### Helix Version

helix 23.03 (3cf03723)

trink:
I am working on the steps to reproduce, for now here is a backtrace of a similar case when changing views.

```
thread 'main' panicked at 'Position 2 is out of range for changeset len 1!', helix-core/src/transaction.rs:399:13
stack backtrace:
   0: rust_begin_unwind
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/897e37553bba8b42751c67658967889d11ecd120/library/core/src/panicking.rs:142:14
   2: helix_core::transaction::ChangeSet::map_pos
   3: <smallvec::SmallVec<A> as core::iter::traits::collect::Extend<<A as smallvec::Array>::Item>>::extend
   4: helix_core::selection::Selection::map
   5: helix_view::view::View::apply
   6: helix_view::view::View::sync_changes
   7: helix_view::editor::Editor::replace_document_in_view
   8: helix_view::editor::Editor::switch
   9: helix_term::ui::editor::EditorView::handle_keymap_event::{{closure}}
  10: helix_term::ui::editor::EditorView::handle_keymap_event
  11: <helix_term::ui::editor::EditorView as helix_term::compositor::Component>::handle_event
  12: helix_term::compositor::Compositor::handle_event
  13: <core::future::from_generator::GenFuture<T> as core::future::future::Future>::poll
  14: tokio::runtime::park::CachedParkThread::block_on
  15: tokio::runtime::scheduler::multi_thread::MultiThread::block_on
  16: tokio::runtime::runtime::Runtime::block_on
  17: hx::main
```

trink:
[...]

PR #7519, opened 2023-07-03
feat(editor): Allow Ctrl-[ to function as Escape

harritaylor:
Issue:  https://github.com/helix-editor/helix/issues/6551

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -488,6 +488,16 @@ impl From<crossterm::event::KeyEvent> for KeyEvent {
                 code: KeyCode::Tab,
                 modifiers,
             }
+        } else if code == crossterm::event::KeyCode::Char('[')
+            && modifiers.contains(crossterm::event::KeyModifiers::CONTROL)
+        {
+            // Support Ctrl-[ as ESC for terminals that support sending Ctrl-[
+            let mut modifiers = modifiers;
+            modifiers.remove(crossterm::event::KeyModifiers::CONTROL);
+            Self {
+                code: KeyCode::Esc,
+                modifiers: modifiers.into(),
+            }
         } else {
             Self {
                 code: code.into(),


pascalkuthe:
Its on purpose that these do not always act the same and sglukd be covered by keymaps not by patching krycodes: https://github.com/helix-editor/helix/issues/6551#issuecomment-1493359754

harritaylor:
@pascalkuthe Sure, I understand - thanks for the reply. What is the recommended approach for "C-[ behaves like Esc" in this case? It is a core part of my muscle memory for terminal applications ([and I'm assuming it is for a few others too](https://github.com/helix-editor/helix/issues/6551#issuecomment-1502574646)).

I tried remapping in configs but there are still instances in `hx` where the `C-[` key is sent instead and I would prefer that particular key combo to always "do what I mean". I have seen https://github.com/helix-editor/helix/issues/5505 which I would be happy to play around with a solution for. I have also seen https://github.com/helix-editor/helix/pull/7020 but I agree with the author that opting out of kitty protocol support should probably not be the way to go.

pascalkuthe:
Solving https://github.com/helix-editor/helix/issues/5505 is the way to fix that issue but that will take quite a while and should probably be done by a maintainer/in close coordination with a maintainer. Such fundamental large changes are hard to coordinate otherwise

harritaylor:
@pascalkuthe Sure. Either way I'd like to contribute - anything tangentially related I can work on as a "good first issue"?

pascalkuthe:
The following could be good things to work on:

* https://github.com/helix-editor/helix/issues/4274: prior work in https://github.com/helix-editor/helix/pull/5435, the author didn't have time to finish, you can use that PR as a starting point and address the remaining review comments. Past  review comments can also teach you more about the codebase)
* https://github.com/helix-editor/helix/issues/4153: add a variant of `:sort` that sorts numerically instead of lexigrahically (so `10` is sorted after `5` for example). Should be easy to just add an additional `numeric: bool` arg to `sort_impl` and add a `:sort-numeric` command that sets that arg to true. When that option is set simply sort with https://crates.io/crates/alphanumeric-sort
* https://github.com/helix-editor/helix/issues/6752#issuecomment-1608816732: a crash that is a bit more challenging to fix but would provide good onsight into the codebase. I left instructions in the comment.

harritaylor:
@pascalkuthe great, thanks!

PR #7520, opened 2023-07-03
Show newest register values in infobox and prompt.

gibbz00:
Fixes: https://github.com/helix-editor/helix/issues/7502, but also how "C-r" in prompt only shows the oldest register value.

I've closed the other PRs (https://github.com/helix-editor/helix/pull/7503, https://github.com/helix-editor/helix/pull/7504, https://github.com/helix-editor/helix/pull/7505, https://github.com/helix-editor/helix/pull/7506) as I can see how they didn't make much sense in the broader context of this register fix and the preparation for implementing "register histories". 

As for the comment in: https://github.com/helix-editor/helix/pull/7512#discussion_r1250000269; release builds proceeding a cargo clean, all other applications closed:

| | Bytes | Compile time |
|---|---|---| 
master |  26 492 | 2:01.85
register_newest | 26 532 |2:02.53
diff | 40 bytes | ~1 second |

It is also misleading to say the creates increase code size as there are plenty of [locations.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/input.rs#L178)  [that.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-core/src/line_ending.rs#L82) [could.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/document.rs#L65)  [be.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/document.rs#L55)  [made.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/editor.rs#L605) [more.](
https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-core/src/line_ending.rs#L57)  [consise.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-core/src/line_ending.rs#L37)  [using.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/graphics.rs#L315)  [them.](https://github.com/helix-editor/helix/blob/457b38939526434fa2ec07ceb2264bac3e4575af/helix-view/src/graphics.rs#L368)


archseer:
As I've stated, I do not want to add 3 new dependencies (8 total with subdependencies). They add additional surface area (we vet every dependency, and if a dependency becomes unmaintained, it's on us to take over), it's much more beneficial to expand them once then embed the expanded code in the project tree. (There was a good burntsushi comment to the same point but I can't find it now)

helix-view/src/register.rs
```diff
@@ -0,0 +1,191 @@
+use crate::info::Info;
+use std::{collections::HashMap, fmt::Display, str::FromStr};
+use strum_macros::{AsRefStr, EnumString};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Hash, EnumString, AsRefStr)]
+pub enum Register {
```
archseer, line 6: Let's keep a `char` as the register identifier, since that was the intent of the original implementation. There is no special "yank" register, just a default slot that's usually used for yanking but can be overriden. In fact all these extra dependencies would be shaken out if we just used a `const Register::YANK = "\""` constant

gibbz00:
@archseer, https://github.com/helix-editor/helix/pull/7520/commits/f69900fa79bdb3bbe8a31ad485759cabdc14ab0a better?

pascalkuthe:
Closing as stale

PR #7523, opened 2023-07-03
Add ability to customize colors for inactive, active modified, and inactive modified buffers on the bufferline

### LINKED ISSUE #7481: Add new highlights to style modified and inactive buffers in the bufferline
A highlight called `ui.bufferline.inactive` could be added which behaves like `ui.bufferline.active`, but only targets inactive buffers. Also add `ui.bufferline.active.modified` and `ui.bufferline.inactive.modified` for styling modified buffers. This will let themes create "tabs" that convey more information to the user using color and font styles.

It would also be nice to make the modified indicator (which is currently `[+]`) configurable, but this might belong in another issue.

cmattinson:
Resolves #7481. Preserve default behavior for ui.bufferline.inactive but add ability to customize. Add ui.bufferline.* entries to the themes book.

Tested with github_dark.toml.
image

editor.rs active
image

Both modified, editor.rs active
image

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -269,6 +269,10 @@ These scopes are used for theming the editor interface:
 | ---                               | ---                                                                                            |
 | `ui.background`                   |                                                                                                |
 | `ui.background.separator`         | Picker separator below input line                                                              |
+| `ui.bufferline.active`            |                                                                                                |
+| `ui.bufferline.active.modified`   |                                                                                                |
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I think abusing the bufferline as a sort of statusline like is something we really want to be doing. In general the bufferline is a focus for us. So I don't think we will be going forward with this. Thank you for contributing!

PR #7524, opened 2023-07-03
docs: Update mdBook theme and improve maintainability

zqianem:
Previously, theming Helix's documentation involved making edits to files copied from the default mdBook theme, which would fully override the upstream theme files. This made it difficult to see which theme changes were specific to Helix and prevented any upstream theme fixes from being applied when mdBook updated.

This PR addresses the issue by extracting all of Helix's custom CSS into a dedicated file and removing as many theme files as possible, as recommended by the [mdBook docs](https://rust-lang.github.io/mdBook/format/theme/index.html):

> Note: When you override a file, it is possible that you break some functionality. Therefore I recommend to use the file from the default theme as template and only add / modify what you need. You can copy the default theme into your source directory automatically by using mdbook init --theme and just remove the files you don't want to override.

The `index.hbs` file could not be removed since Helix adds a custom theme; once https://github.com/rust-lang/mdBook/pull/2059 is merged it may be possible. Upstream changes were manually merged instead.

Tested by running `mdbook serve` under `book/` — pages should be pixel-identical to the current docs except for some upstream changes and a fix for layout differences between themes.


the-mikedavis (approved):
Thanks! I tried serving this and master side-by-side and couldn't tell the difference 👍

pascalkuthe (approved):

PR #7526, opened 2023-07-03
Add ability to configure the file modification indicator

cmattinson:
Relates to the second request in #7481.

Previously the file modification indicator was documented under [[editor.statusline]](https://docs.helix-editor.com/configuration.html#editorstatusline-section), but the way I have it currently, the configuration will change the file modification indicator on both the buffer line AND the status line. I've put it there as it's more of a general editor change. Wondering if there are any opinions on whether:

A - The configurable indicator should control both the buffer and status line indicator (current)
B - The configurable indicator should only be for the buffer line and the status line default [+] should remain
C - They should be independently configurable 

I've also made the assumption that someone may the entire indicator configurable, rather than just the character(s) inside the [].

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -432,8 +432,9 @@ fn render_file_modification_indicator<F>(context: &mut RenderContext, write: F)
 where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
+    let indicator = &context.editor.config().file_modification_indicator;
     let title = (if context.doc.is_modified() {
-        "[+]"
+        indicator
     } else {
         "   "
     })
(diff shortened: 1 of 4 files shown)


sharpenedblade:
I think the indicator being independently configurable is the best way to go about this. The indicators could be  different because the bufferline needs to fit more buffer names but the statusline only needs to fit one (example: `[m]` in the bufferline and `[modified]` in the statusline).

helix-term/src/ui/statusline.rs
```diff
@@ -432,8 +432,9 @@ fn render_file_modification_indicator<F>(context: &mut RenderContext, write: F)
 where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
+    let indicator = &context.editor.config().file_modification_indicator;
     let title = (if context.doc.is_modified() {
-        "[+]"
+        indicator
     } else {
         "   "
```
cunha, line 439: Personal preference, but `indicator.len()` spaces will likely look better for shorter indicators.

PR #7534, opened 2023-07-04
Fix incorrect gutter bail message

t-taylor:
`"spacer"` was listed twice in the `editors.gutter` example and the error message for the gutter config parsing was missing newer options.

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -611,7 +611,9 @@ impl std::str::FromStr for GutterType {
             "spacer" => Ok(Self::Spacer),
             "line-numbers" => Ok(Self::LineNumbers),
             "diff" => Ok(Self::Diff),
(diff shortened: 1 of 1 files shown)


pascalkuthe (dismissed):

pascalkuthe:
Please run rustfmt to make CI pass

pascalkuthe (approved):

the-mikedavis (commented):
The change in the `bail!` message looks good but I believe the configuration docs are correct: https://github.com/helix-editor/helix/blob/57e538d07b15c7ac33a0469f1ca05981704ee5af/helix-view/src/editor.rs#L86-L99

Gutter elements are repeatable and spacer is used to add some space between other elements.

pascalkuthe:
ah right I misserad the change. The docs show the default configureitaton (which contains two spacers)  not the list of available options

pascalkuthe (approved):

the-mikedavis (approved):

PR #7538, opened 2023-07-04
use redraw handle for debouncing LSP messages

pascalkuthe:
See https://github.com/helix-editor/helix/pull/7373#issuecomment-1616716678 for prior discussion

Fixes helix not redrawing after receiving messages from the LSP by replacing the existing (buggy) special-purpose deboucing with using the redraw handle that was added as part of #3890

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -34,7 +34,6 @@ use std::{
     io::{stdin, stdout},
     path::Path,
     sync::Arc,
-    time::{Duration, Instant},
 };
 
 use anyhow::{Context, Error};
@@ -44,8 +43,6 @@ use crossterm::{event::Event as CrosstermEvent, tty::IsTty};
 use {signal_hook::consts::signal, signal_hook_tokio::Signals};
 #[cfg(windows)]
 type Signals = futures_util::stream::Empty<()>;
-
-const LSP_DEADLINE: Duration = Duration::from_millis(16);
 
 #[cfg(not(feature = "integration"))]
 use tui::backend::CrosstermBackend;
@@ -76,7 +73,6 @@ pub struct Application {
     signals: Signals,
     jobs: Jobs,
     lsp_progress: LspProgressMap,
-    last_render: Instant,
 }
 
 #[cfg(feature = "integration")]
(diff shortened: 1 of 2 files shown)


helix-term/src/application.rs
```diff
@@ -609,12 +603,7 @@ impl Application {
             EditorEvent::LanguageServerMessage((id, call)) => {
                 self.handle_language_server_message(call, id).await;
                 // limit render calls for fast language server messages
-                let last = self.editor.language_servers.incoming.is_empty();
```
the-mikedavis, line 612: For context: the `is_empty` here checks that the set of streams is empty rather than that there is no pending message ready to be handled. Detecting that case is a little involved and not very effective for debouncing (see https://github.com/helix-editor/helix/pull/7373#issuecomment-1616716678)

the-mikedavis (approved):

helix-view/src/editor.rs
```diff
@@ -1703,7 +1703,7 @@ impl Editor {
                 _ = self.redraw_handle.0.notified() => {
                     if  !self.needs_redraw{
                         self.needs_redraw = true;
-                        let timeout = Instant::now() + Duration::from_millis(96);
+                        let timeout = Instant::now() + Duration::from_millis(33);
```
archseer, line 1706: Why the change here?

pascalkuthe: > The only disadvantage of this approach was that the rendering now has a longer render timeout (96ms instead of 16ms) which felt a bit slow. I changed it to 33ms to emulate 30FPS. It doesn't matter too much for diff gutters what that value is since diff gutters only use this as a fallback mechanism but I would prefer to not lower this too much more to avoid adding too many renders. 

https://github.com/helix-editor/helix/pull/7373#issuecomment-1616716678

We were rendering with approximately 60fps in the past (the debouncer wasn't exact but we were using 16 FPS). The redraw handle used to have a longer timeout since its only used as a fallback for the diff gutters (basically if we get here its already slow) but the exact value doesn't matter too much. I lowered it so we at least keep rending with 30 FPS. I didn't want to go quite as low as 16 here, if we really care we could allow different users of the redraw handle to specify different timeouts. That seemed overkill tough but if you think that's worth it, it shouldn't be too hard

PR #7541, opened 2023-07-04
Use theme based on color scheme

Siilwyn:
Start at solving #2158, I hope this can eventually land in master, otherwise I'm happy to share this branch for others to use.

- This uses the freedesktop portal to read the color-scheme preference.
- At the moment this only reads the color scheme on startup.
- zbus supports watching a property but I'm **unsure where in the codebase to start this process, I'd love help with this**: any pointers welcome!
- Thinking of removing the `theme_light` in favor of using the "normal" `theme` when the color scheme is: not found, has no preference or is set to prefer light. This simplifies things a bit and matches more with most distros that don't communicate prefer light when dark mode is off.

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -61,6 +61,8 @@ toml = "0.7"
 
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+zbus = { version = "3.13" }
+zvariant = { version = "3.13" }
 
 # ripgrep for global search
 grep-regex = "0.1.11"
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -92,6 +92,134 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "bddcadddf5e9015d310179a59bb28c4d4b9920ad0f11e8e14dbadf654890c9a6"
 
 [[package]]
+name = "async-broadcast"
+version = "0.5.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7c48ccdbf6ca6b121e0f586cbc0e73ae440e56c67c30fa0873b4e110d9c26d2b"
+dependencies = [
+ "event-listener",
+ "futures-core",
+]
+
+[[package]]
+name = "async-channel"
+version = "1.8.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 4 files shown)


helix-term/Cargo.toml
```diff
@@ -61,6 +61,8 @@ toml = "0.7"
 
 serde_json = "1.0"
 serde = { version = "1.0", features = ["derive"] }
+zbus = { version = "3.13" }
+zvariant = { version = "3.13" }
```
archseer, line 65: This pulls in hundreds of dependencies just to read a single value. Could this be done by a wrapper script on the user side that's executed when the editor starts?

Siilwyn: Yes that's quite unfortunate, though I think they don't have a big impact on the binary size I'll test that. I'm going to add functionality that watches the dbus as well so a wrapper script won't cut it.
I am thinking though maybe there's a 'cheaper' way to interact with the dbus by using cli tools like `dbus-send`

jesse-c:
As someone who's invested in this sort of behaviour [1], I'm curious about this feature! I primarily use macOS, and as far as I can see, the proposed solution so far wouldn't support macOS? If that's true, I'm happy to do some thinking about how this could feature could be supported for macOS.

[1] https://github.com/jesse-c/AppearanceNotifier

pascalkuthe:
Concenrs like this is why I personally belive something like this does not belong in core. There is no standard solution/its highly platform specific and all that platform-specific code pulls in a huge number of dependencies. Something like this could be a plugin (which could reasonably be a platform-specific so you have one plugin for mac and one for linux) once we have ar plugin system until then you could use a deamon to symlink a different config file depending on whether dark/light theme is active and send  a signal (pkill -USR1 hx) to reload the config

jesse-c:
That makes sense, @pascalkuthe!

>  If this could be implemented in a couple lines of code plus a simple wrapper script that could be ok but I don't think that's the case.

For macOS, if with performance, and other points, allowed, it can be done with a timer that runs every `n` milliseconds (or even every 1-3 seconds tbh), that checks the output of the terminal command `defaults read -g AppleInterfaceStyle`, which will tell you what the current macOS theme setting is (light vs dark).

> Until then you could implement a rimple daemon to symlinks a different config file depending on whether dark/light theme is active and send a signal (pkill -USR1 hx) to reload the config.

This is what I have at the moment! I was happy when I discovered this just last week, haha.

pascalkuthe:
closing this one out as I don't see us adding this to core (see discussion above). Thank you for contributing!

PR #7542, opened 2023-07-04
Register histories

gibbz00:
Builds upon: https://github.com/helix-editor/helix/pull/7520

Fixes to some extent: https://github.com/helix-editor/helix/issues/4952

Future improvements:

* Pagination (Only the top 10 most recent values are currently shown.)
* Incorporate: https://github.com/helix-editor/helix/pull/6985
* Indicator for showing number of selections in "yank" register.

Demo:

https://github.com/helix-editor/helix/assets/31696304/fa203b35-2b6d-458d-9d8a-70235981a0f8

PR #7565, opened 2023-07-07
Register histories + special registers 

gibbz00:
Incorporates https://github.com/helix-editor/helix/pull/7542 with https://github.com/helix-editor/helix/pull/6985

Some differences however:

* Doesn't use dynamic dispatching.
* Haven't implemented the system selection caching feature. (Can do so if requested.)

Haven't been too nitpicky about writing highly optimized code from the get go, partly because I felt the added complexity would be premature. Nevertheless open to implementing any optimization suggestions.

Register history selection by index is only implemented for the "non-context aware" registers. The only context register where I could see it make sense was the system clipboard registers, but that felt a bit out of my reach right now.

I also noticed that clipboard error handling is done on each call to `ClipboardProvider::get/set_ contents()`. It might be a good idea to make trait methods return an `Option` and centralize the error logging in the `Provider` implementation.

2023-07-07-223107_screenshot


pascalkuthe:
closing as this is very stale and we merged https://github.com/helix-editor/helix/pull/6985

PR #7568, opened 2023-07-08
Restore view offset when switching buffers with multiple windows

t-monaghan:
This pull request is in response to changes requested by @pascalkuthe in #7414 looking to solve the issue #7355.

The changes made are to ensure views can store and access their own positioning for documents by way of including the field `view_data: HashMap<ViewId, ViewData>` in `Document`. 
 
I haven't yet addressed the point of not differentiating between last view position and current view position. I'm unsure how to proceed with this goal.

Fixes https://github.com/helix-editor/helix/issues/7355

### DIFF
--- a/helix-view/src/document.rs
+++ b/helix-view/src/document.rs
@@ -34,6 +34,7 @@ use helix_core::{
 };
 
 use crate::editor::{Config, RedrawHandle};
+use crate::view::ViewPosition;
 use crate::{DocumentId, Editor, Theme, View, ViewId};
 
 /// 8kB of buffer space for encoding and decoding `Rope`s.
@@ -131,6 +132,7 @@ pub struct Document {
     pub(crate) id: DocumentId,
     text: Rope,
     selections: HashMap<ViewId, Selection>,
+    view_data: HashMap<ViewId, ViewData>,
 
     /// Inlay hints annotations for the document, by view.
     ///
@@ -263,6 +265,7 @@ impl fmt::Debug for Document {
             .field("selections", &self.selections)
             .field("inlay_hints_oudated", &self.inlay_hints_oudated)
             .field("text_annotations", &self.inlay_hints)
+            .field("view_data", &self.view_data)
             .field("path", &self.path)
             .field("encoding", &self.encoding)
             .field("restore_cursor", &self.restore_cursor)
@@ -654,6 +657,7 @@ impl Document {
             selections: HashMap::default(),
             inlay_hints: HashMap::default(),
             inlay_hints_oudated: false,
+            view_data: Default::default(),
             indent_style: DEFAULT_INDENT,
             line_ending,
             restore_cursor: false,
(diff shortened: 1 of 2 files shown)


helix-view/src/document.rs
```diff
@@ -1614,6 +1618,14 @@ impl Document {
         &self.selections
     }
 
+    pub fn view_data(&mut self, view_id: ViewId) -> &ViewData {
+        self.view_data.entry(view_id).or_default()
+    }
+
+    pub fn view_data_mut(&mut self, view_id: ViewId) -> &mut ViewData {
+        self.view_data.entry(view_id).or_default()
+    }
```
the-mikedavis, line 1627: I wonder if these should return `Option<&[mut ]Viewdata>` instead? That would allow us to drop the Default requirement for `ViewData`. It's not currently useful to do that between this and https://github.com/helix-editor/helix/pull/6198 though so I don't have a strong opinion either way

the-mikedavis (commented):
This is looking good to me. I'll test it for a while to make sure the behavior feels right

pascalkuthe (changes requested):
Thus still hast Adressen the issue about mapping view position trough changes. If you edit the buffer in another split the view offset would change. So deleting text in another view causes the position to slide up.

This idea was to entirely replace view.offset with the ViewPosition introduced in this PR (since that fixes other oddities that I ran into) but it's a larger change that can be kept separate.

the-mikedavis:
For reference see where we update selections when there are changes to a document: https://github.com/helix-editor/helix/blob/d570c29ce37ffbb46a9c49708c31dfd81daa27cf/helix-view/src/document.rs#L1160-L1167

diagnostics: https://github.com/helix-editor/helix/blob/d570c29ce37ffbb46a9c49708c31dfd81daa27cf/helix-view/src/document.rs#L1222-L1252

and inlay hints: https://github.com/helix-editor/helix/blob/d570c29ce37ffbb46a9c49708c31dfd81daa27cf/helix-view/src/document.rs#L1262-L1270

pascalkuthe:
>  we should still be able to map through changes by finding views relevant to a document from Document.selections, and updating the relevant view position history value on them.

this is not possible you can't access views from `apply`


I plan to remove document and view as separate structs entirely in the future. Separating them to begin with was fairly pointless since nearly all fields are public so there is little to no encapsulation to begin with (for example the jumplist is a gift that keeps on giving more and more crashes because the laz4y mapping is just not robusta and it can't be encapsulated because it needs to be accessed from document).

I want to refactor that so `Document` and `View` only become ids and all their data is stored in separate containers (slotmap or hashmap) instead of one big sprawling struct (so a more entity component-like system which in my experience is a better data model for large rust applications). The data would be grouped into files/substructs where encapsulation actually makes sense/is possible (litmus test: all fields can and should be private). These changes will be required anyway for plugins/making the event system more useful (and the config refactor). But it's a really large set of changes that I will take on down the line.

I am not concerned about the extra hashmap lookup, that is basically irrelevant. HasMaps are really really fast.

the-mikedavis:
Closing in favor of #10559

PR #7573, opened 2023-07-08
Fix `:log-open` when `--log` is specified

alevinval:
The command always opened the default log file path. The path can be customized passing the `--log` optional argument. When this is the case the `open-log` command should respect it.

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -11,19 +11,23 @@ static RUNTIME_DIRS: once_cell::sync::Lazy<Vec<PathBuf>> =
 
 static CONFIG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
 
+static LOG_FILE: once_cell::sync::OnceCell<PathBuf> = once_cell::sync::OnceCell::new();
+
 pub fn initialize_config_file(specified_file: Option<PathBuf>) {
-    let config_file = specified_file.unwrap_or_else(|| {
-        let config_dir = config_dir();
-
-        if !config_dir.exists() {
-            std::fs::create_dir_all(&config_dir).ok();
-        }
-
-        config_dir.join("config.toml")
-    });
-
-    // We should only initialize this value once.
+    let config_file = specified_file.unwrap_or_else(default_config_file);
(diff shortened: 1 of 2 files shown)


pascalkuthe (dismissed):
thanks!

the-mikedavis (commented):
I think we should use the same strategy as the config file so we don't need to tote around this log_file parameter on more structs and constructors: https://github.com/helix-editor/helix/blob/618620b36907bb2cbe29cd0e737efad9f651a00d/helix-loader/src/lib.rs#L12-L27

https://github.com/helix-editor/helix/blob/618620b36907bb2cbe29cd0e737efad9f651a00d/helix-loader/src/lib.rs#L124-L129

alevinval:
@the-mikedavis really nice! much cleaner, I've also done some boy-scouting around that code, let me know if you are cool with it.

the-mikedavis (commented):
I believe we also need an `ensure_log_dir` since that may be in a different directory than config (by deafult it is for example). I think it would be fine for those functions to be called in the initialize_ functions since those are only meant to be called once at initialization anyways

helix-loader/src/lib.rs
```diff
@@ -122,10 +121,11 @@ pub fn cache_dir() -> PathBuf {
...
 pub fn config_file() -> PathBuf {
-    CONFIG_FILE
-        .get()
-        .map(|path| path.to_path_buf())
-        .unwrap_or_else(|| config_dir().join("config.toml"))
+    CONFIG_FILE.get().map(|path| path.to_path_buf()).unwrap()
+}
+
+pub fn log_file() -> PathBuf {
+    LOG_FILE.get().map(|path| path.to_path_buf()).unwrap()
```
alevinval, line 128: I was about to write the following here (and the equivalent for the `config_file()`)

```rust
unwrap_or_else(|| {
   initialize_log_file(None)
   log_file()
})
```

But then I really think we should `unwrap()` and potentially fail. A failure here means we have seriously broken the initialization steps in the `main.rs`, it seems less misleading than re-replicating the initialization of the cell in two places.

the-mikedavis (approved):

pascalkuthe (approved):

PR #7576, opened 2023-07-08
Update dart commit in languages.toml

Queyrouzec:
This fixes syntax highlighting problems with major Dart 3 update on classes, enums, and switch statements

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1224,7 +1224,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "dart"
-source = { git = "https://github.com/UserNobody14/tree-sitter-dart", rev = "2d7f66651c9319c1a0e4dda226cc2628fbb66528" }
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Looks like dispite the big changes to the grammar, the old queries are still compatible https://github.com/UserNobody14/tree-sitter-dart/compare/2d7f66651c9319c1a0e4dda226cc2628fbb66528...e398400a0b785af3cf571f5a57eccab242f0cdf9#diff-919ac210accac9ecc55a76d10a7590e3d85ca3f0e165b52d30f08faee486d0cb

It looks to me like mostly like refactoring and bugfixes but if we see some highlights change we can address those in follow-ups 👍

Queyrouzec:
Most of the commits are bug fixes but the addition of the new syntax was added under the commits for May 3rd, 2023 and the current commit is Aug 11, 2022.

PR #7585, opened 2023-07-09
Initialize log and config files right after parsing arguments

alevinval:
Either after parsing arguments or before running the `--health` logic, since that one already requires access to the `config` dir. I'd say it's better to play it safe and put it right after parsing arguments.

### DIFF
--- a/helix-term/src/main.rs
+++ b/helix-term/src/main.rs
@@ -76,6 +76,9 @@ FLAGS:
 
     let args = Args::parse_args().context("could not parse arguments")?;
 
+    helix_loader::initialize_config_file(args.config_file.clone());
+    helix_loader::initialize_log_file(args.log_file.clone());
+
     // Help has a higher priority and should be handled separately.
     if args.display_help {
         print!("{}", help);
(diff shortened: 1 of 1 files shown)


helix-term/src/main.rs
```diff
@@ -76,6 +76,11 @@ FLAGS:
 
     let args = Args::parse_args().context("could not parse arguments")?;
 
+    helix_loader::initialize_config_file(args.config_file.clone());
+    helix_loader::initialize_log_file(args.log_file.clone());
+
+    setup_logging(args.verbosity).context("failed to initialize logging")?;
```
alevinval, line 82: Technically it could be moved right before `--health`, I can do that. In terms of future-proofing, I think doing it as soon as possible is best.

alevinval: I see that fetch/build grammars and health do not use logging, but I still moved `setup_logging` to the top. I can leave it where it was if you disagree.

pascalkuthe (approved):

the-mikedavis (approved):

PR #7588, opened 2023-07-10
enhanced `surround_replace` to provide visual feedback

alevinval:
When the target character is pressed, the selection is temporarily updated to display the affected positions. It restores the original selection after the operation is cancelled or completed.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -5115,33 +5115,26 @@ fn surround_replace(cx: &mut Context) {
             }
         };
 
-        // Visual feedback
-        let selection = selection.clone();
-        let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
-        // TODO: Use [`array_chunks`] once stabilized
-        for p in change_pos.chunks_exact(2) {
-            let [from, to] = *p else { unreachable!() };
-            ranges.push(Range::point(from));
-            ranges.push(Range::point(to));
-        }
+        let original_selection = selection.clone();
+        let ranges: SmallVec<_> = change_pos.iter().map(|&p| Range::new(p, p + 1)).collect();
         doc.set_selection(view.id, Selection::new(ranges, 0));
 
         cx.on_next_key(move |cx, event| {
             let (view, doc) = current!(cx.editor);
             let to = match event.char() {
                 Some(to) => to,
-                None => return doc.set_selection(view.id, selection),
+                None => return doc.set_selection(view.id, original_selection),
             };
             let (open, close) = surround::get_pair(to);
-            let transaction = Transaction::change(
-                doc.text(),
-                change_pos.iter().enumerate().map(|(i, &pos)| {
+            let mut i = 0;
+            let transaction =
+                Transaction::change_by_selection(doc.text(), doc.selection(view.id), |range| {
                     let mut t = Tendril::new();
                     t.push(if i % 2 == 0 { open } else { close });
-                    (pos, pos + 1, Some(t))
-                }),
-            );
-            doc.set_selection(view.id, selection);
+                    i += 1;
+                    (range.from(), range.to(), Some(t))
+                });
+            doc.set_selection(view.id, original_selection);
             doc.apply(&transaction, view.id);
             exit_select_mode(cx);
         });
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -5119,11 +5119,11 @@ fn surround_replace(cx: &mut Context) {
         let selection = selection.clone();
         let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
         // TODO: Use [`array_chunks`] once stabilized
-        change_pos.chunks_exact(2).for_each(|p| {
+        for p in change_pos.chunks_exact(2) {
             let [from, to] = *p else { unreachable!() };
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -5115,11 +5115,25 @@ fn surround_replace(cx: &mut Context) {
             }
         };
 
+        // Visual feedback
+        let selection = selection.clone();
+        let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
+        change_pos.chunks(2).for_each(|p| {
+            if p.len() == 2 {
```
archseer, line 5122: With `chunks_exact` you can then avoid this `len` check

I'd also leave a note to use `array_chunks` in the future once stabilized

helix-term/src/commands.rs
```diff
@@ -5115,11 +5115,25 @@ fn surround_replace(cx: &mut Context) {
...
+                let from = *p.first().unwrap();
+                let to = *p.get(1).unwrap();
+                ranges.push(Range::new(from, from + 1));
+                ranges.push(Range::new(to, to + 1));
+            }
+        });
+        let feedback = Selection::new(ranges, 0);
+        doc.set_selection(view.id, feedback);
+
         cx.on_next_key(move |cx, event| {
```
archseer, line 5132: What happens if esc is pressed? I think this next key operation then doesn't run

alevinval: Yes, it goes via the `None` branch, which is where I restore the original selection.

helix-term/src/commands.rs
```diff
@@ -5115,11 +5115,22 @@ fn surround_replace(cx: &mut Context) {
             }
         };
 
+        // Visual feedback
+        let selection = selection.clone();
+        let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
+        // TODO: Use [`array_chunks`] once stabilized
+        change_pos.chunks_exact(2).for_each(|p| {
```
the-mikedavis, line 5122: `for` is considered more idiomatic than `Iterator::for_each` when not many methods are chained like in this case. See https://doc.rust-lang.org/core/iter/trait.Iterator.html#method.for_each

```rust
for point in change_pos.chunks_exact(2) {
    let [from, to] = *point else { unreachable!() };
    ranges.push(Range::point(from));
    ranges.push(Range::point(to));
}
```

alevinval: Learning a lot! Thx

helix-term/src/commands.rs
```diff
@@ -5115,11 +5115,22 @@ fn surround_replace(cx: &mut Context) {
             }
         };
 
+        // Visual feedback
+        let selection = selection.clone();
+        let mut ranges: SmallVec<[Range; 1]> = SmallVec::new();
+        // TODO: Use [`array_chunks`] once stabilized
+        for p in change_pos.chunks_exact(2) {
```
archseer, line 5122: With the latest code, do wee even need to chunk? It's just mapping all the values to point ranges now

alevinval: I wonder why I chunked in the first place... I've revisited and gone the extra mile now, making the transaction work on the selection, since we're updating it anyway to show the visual feedback. Hopefully that should be it.

helix-term/src/commands.rs
```diff
@@ -5115,21 +5115,26 @@ fn surround_replace(cx: &mut Context) {
...
+            let transaction =
+                Transaction::change_by_selection(doc.text(), doc.selection(view.id), |range| {
                     let mut t = Tendril::new();
                     t.push(if i % 2 == 0 { open } else { close });
-                    (pos, pos + 1, Some(t))
-                }),
-            );
+                    i += 1;
+                    (range.from(), range.to(), Some(t))
+                });
```
archseer, line 5136: Why is the change here necessary? We could just keep the original version that uses enumerate()

alevinval: I thought it would make things more consistent to follow, since the selection has already been updated to show the affected surround positions, the transaction just needs to apply the change function to each selected position.

helix-term/src/commands.rs
```diff
@@ -5115,11 +5115,15 @@ fn surround_replace(cx: &mut Context) {
             }
         };
 
+        let selection = selection.clone();
+        let ranges: SmallVec<_> = change_pos.iter().map(|&pos| Range::point(pos)).collect();
+        doc.set_selection(view.id, Selection::new(ranges, 0));
```
archseer, line 5120: Okay, one last thing: you need to compute the correct primary_idx here or the view will be moved if you have a lot of selections: if you selected all the lines in a very big file and were viewing somewhere further down, an index of 0 would move the view all the way to the top

Something like

```rust
let primary = selection.primary();
let (i, _) = ranges.iter().enumerate().find(|_i, range| range.overlaps(primary))...

Selection::new(ranges, i)
```

alevinval: Good point. I've pushed [this change](https://github.com/helix-editor/helix/pull/7588/commits/6dd6347f732431fc78f5dd338572281ac4de3f9a) and so far so good during smoke testing,  but I wonder if there's a better way.

pascalkuthe (approved):
this looks good to me, nice visual improvement with very little additional complexity added. Great job!

PR #7591, opened 2023-07-10
Unify logging and expand its documentation in CONTRIBUTING.md

gibbz00:
Makes it also possible to provide the `HELIX_LOG_LEVEL` env var that was previously only available in the integration test suite. Also did some minor clean up in main.rs if that's ok.


pascalkuthe:
Closing as stale

PR #7592, opened 2023-07-10
Register systemd files as ini

Arian-D:
Howdy yall,
I edit a lot of systemd files, which are the INI format, and upon opening helix I constantly have to `:set-language`. I was wondering if it would be okay to add them to the ini file types.
- [Systemd unit files](https://www.man7.org/linux/man-pages/man5/systemd.unit.5.html)
- [Podman quadlet systemd units](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html)

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2233,7 +2233,26 @@ grammar = "bash"
 [[language]]
 name = "ini"
 scope = "source.ini"
-file-types = ["ini"]
+file-types = [
+  "ini",
+  # Systemd unit files
+  "service",
+  "automount",
+  "device",
+  "mount",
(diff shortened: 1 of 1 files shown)


languages.toml
```diff
@@ -2233,7 +2233,7 @@ grammar = "bash"
 [[language]]
 name = "ini"
 scope = "source.ini"
-file-types = ["ini"]
+file-types = ["ini", ".service", ".container", ".volume", ".kube", ".network"]
```
the-mikedavis, line 2236: The leading `.`s can all be removed here

PR #7607, opened 2023-07-12
Add cargo feature that enables piping on macos

yyogo:
Follow up to #5468. 

Since there have been understandable performance concerns around the tty polling implementation in the `crossterm` feature that enables piping on macos (https://github.com/crossterm-rs/crossterm/pull/735), this PR adds a feature flag to selectively enable this, so that we can conditionally compile it only for macos targets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -263,6 +263,7 @@ checksum = "a84cda67535339806297f1b331d6dd6320470d2a0fe65381e79ee9e156dd3d13"
 dependencies = [
  "bitflags 1.3.2",
  "crossterm_winapi",
+ "filedescriptor",
  "futures-core",
  "libc",
  "mio",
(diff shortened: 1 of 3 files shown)


pascalkuthe:
I don't think a feature is the right way to handle this. On macos we likely never want to turn this off and requiring an extra argument to build on macos will be missed by many.

[Since resolver version 2 you can use a target specific dependency to handle this](https://doc.rust-lang.org/cargo/reference/resolver.html#feature-resolver-version-2):

``` toml
[target.'cfg(target_os = "macos")'.dependencies]  # https://github.com/vorner/signal-hook/issues/100
crossterm = { version = "0.26", features = ["use-dev-tty"] }
```

yyogo:
@pascalkuthe  totally missed that this was possible! will add this in a jiffy

yyogo:
Pushed to the original PR https://github.com/helix-editor/helix/pull/5468

PR #7611, opened 2023-07-12
Change dark_plus inlay-hints colors to more pleasant colors

mkblast:
A little change to the `dark_plus` theme.

- Before:
2023-07-12T17:48:17,965895541+01:00

- After:
2023-07-12T17:48:02,745411692+01:00

### DIFF
--- a/runtime/themes/dark_plus.toml
+++ b/runtime/themes/dark_plus.toml
@@ -89,7 +89,7 @@
 "ui.virtual.whitespace" = { fg = "dark_gray" }
 "ui.virtual.ruler" = { bg = "borders" }
 "ui.virtual.indent-guide" = { fg = "dark_gray4" }
-"ui.virtual.inlay-hint" = { fg = "dark_gray"}
+"ui.virtual.inlay-hint" = { fg = "dark_gray5"}
 
 "warning" = { fg = "gold2" }
 "error" = { fg = "red" }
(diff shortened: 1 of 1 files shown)


David-Else:
@mkblast Did you check out https://github.com/helix-editor/helix/discussions/6270 ?

I agree that they look dreadful at the moment, I think they were hastily done to copy the VS Code default.. but in VS Code there are also other ways it is styled so it looks better. If a direct copy of the colors from the official theme are needed, maybe better to take them from the [dark+ experimental version](https://github.com/microsoft/vscode/blob/cd1d31179d612c380262746f00d35ed072dc7728/extensions/theme-defaults/themes/dark_plus_experimental.json#L43-L46) as discussed in the discussion? What do you think? Your colors look OK to me, but I think some others will want to input their opinion on this.

mkblast:
I think a background colors is not needed. for me its too distracting since it has more contrast.

However, I like the foreground color of [dark+ experimental version](https://github.com/microsoft/vscode/blob/cd1d31179d612c380262746f00d35ed072dc7728/extensions/theme-defaults/themes/dark_plus_experimental.json#L43-L46) way better than what i choose.

David-Else:
> However, I like the foreground color of [dark+ experimental version](https://github.com/microsoft/vscode/blob/cd1d31179d612c380262746f00d35ed072dc7728/extensions/theme-defaults/themes/dark_plus_experimental.json#L43-L46) way better than what i choose.

So do I, it would be cool to change your PR to that colour, it is then an official VS Code colour too :)

mkblast:
Are we good with no background color or we wait for more input?

kirawi (approved):

the-mikedavis (approved):

pascalkuthe (approved):

PR #7622, opened 2023-07-14
Support inlay-hints for svelteserver.

erasin:
- Support inlay-hints
- Add indents.scm for svelte


pascalkuthe (approved):
I am not an expert on the indent queries but this LGTM

Issue #6265, opened 2023-03-11
Helix unable to parse stackTrace response from debugpy (Python DAP)


rkshthrmsh:
While trying to integrate [debugpy](https://github.com/microsoft/debugpy/), it seems that Helix is unable to parse the `stackTrace` response. This is possibly because Helix is attempting to interpret `Module.id` and / or `StackFrame.moduleId` as a string instead of a number, which is what debugpy returns. The DAP spec recommends either of these types for those fields. Here is a log from Helix:

```
2023-03-05T18:42:12.778 mio::poll [TRACE] registering event source with poller: token=Token(5), interests=READABLE | WRITABLE
2023-03-05T18:42:12.778 mio::poll [TRACE] registering event source with poller: token=Token(6), interests=READABLE | WRITABLE
2023-03-05T18:42:12.778 helix_dap::transport [INFO] -> DAP {"type":"request","seq":0,"command":"initialize","arguments":{"adapterID":"debugpy","clientID":"hx","clientName":"helix","columnsStartAt1":true,"linesStartAt1":true,"locale":"en-us","pathFormat":"path","supportsInvalidatedEvent":false,"supportsMemoryReferences":false,"supportsProgressReporting":false,"supportsRunInTerminalRequest":true,"supportsVariablePaging":false,"supportsVariableType":true}}
2023-03-05T18:42:12.850 helix_dap::transport [INFO] <- DAP {"seq": 1, "type": "event", "event": "output", "body": {"category": "telemetry", "output": "ptvsd", "data": {"packageVersion": "1.6.6"}}}
2023-03-05T18:42:12.851 helix_dap::transport [INFO] <- DAP event Output(Output { output: "ptvsd", category: Some("telemetry"), group: None, line: None, column: None, variables_reference: None, source: None, data: Some(Object {"packageVersion": String("1.6.6")}) })
2023-03-05T18:42:12.851 helix_dap::transport [INFO] <- DAP {"seq": 2, "type": "event", "event": "output", "body": {"category": "telemetry", "output": "debugpy", "data": {"packageVersion": "1.6.6"}}}
2023-03-05T18:42:12.851 helix_dap::transport [INFO] <- DAP event Output(Output { output: "debugpy", category: Some("telemetry"), group: None, line: None, column: None, variables_reference: None, source: None, data: Some(Object {"packageVersion": String("1.6.6")}) })
[...]

PR #7627, opened 2023-07-14
Use skip_serializing_none macro to reduce clutter in helix-dap

thomasaarholt:
New rust developer here, please be gentle. 

I decided to take a look at #6265 to see if there was anything I could help with, and noticed that there were about 200 lines of `#[serde(skip_serializing_if = "Option::is_none")]` that could be removed using the `#[skip_serializing_none]` macro. It's now a lot easier to parse it, at least to my eyes.

I wasn't sure whether to add cargo.lock or not, so I added it in a separate commit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -113,6 +113,12 @@ dependencies = [
 ]
 
 [[package]]
+name = "base64"
+version = "0.21.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "604178f6c5c21f02dc555784810edfb88d34ac2c73b2eae109655649ee73ce3d"
+
+[[package]]
 name = "bitflags"
 version = "1.3.2"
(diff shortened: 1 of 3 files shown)


pascalkuthe (changes requested):
From a functional perspective, this seems fine. However, we usually don't want to add new dependencies if it can be avoided. Both to avoid additional compile time hits and to reduce the need to maintain and audit them. There is of course no hard rule but generally the rule of thumb is: If a dependency just saves us writing a couple lines of trivial code its not worth it. 

I think both for DAP we never actually want to send `None' values so one way to solve this without requiring additional proc macros/dependencies would be to simply write our own serializer to wraps the serde_json serializer and always ignore `None` values. So that could be way to reduce the noise here and avoid accidently forgetting there annotations.

thomasaarholt:
I understand, that makes sense. I’m not sure I can commit to that sort of thing right now, but I will take a look.

Issue #5951, opened 2023-02-12
Add DBG (Debug) editor mode


filipdutescu:
# Description

Add a new editor mode, alongside `NOR`, `INS`, `SEL`: the `DBG` mode, which would be basically `NOR` with extra configurable keybindings for the debugging commands. This would also remove the need for the sticky popup, which could be turned on by a new keybinding, such as `?`, for example;

# Proposal

Add a new editor mode which has the same bindings as normal, but with extra bindings for debugging actions, such as launch/attach, step into/over/out etc. For more details and reasoning see #5950.

# UI mockup

![image](https://user-images.githubusercontent.com/46979062/218331354-a631c9fa-2536-4da8-8216-0009a18b3778.png)

archseer:
The sticky popup **is** essentially a separate mode (just the statusline doesn't update). I'd prefer to keep it that way rather than add another state to the mode enum. (In fact the select mode could probably be rewritten this way too -- we only want to internally distinguish between insert mode and non-insert mode)

pascalkuthe:
> The sticky popup **is** essentially a separate mode (just the statusline doesn't update). I'd prefer to keep it that way rather than add another state to the mode enum. (In fact the select mode could probably be rewritten this way too -- we only want to internally distinguish between insert mode and non-insert mode)

I suggested a new mode in mateix. It seemed like the easiest solution:

* Its super frustrating that you can't use normal keybindgs like movements while debugging and entering/exiting
* the popup takes up a lot of screen real estate so it would be nice to make that toggabke rather than keep it around all the time. Especially if we ads more bindings this would become quite large 
* But if the popup is gone there is no easy way to tell if we the debug bindings are active => hence showing something in the ststusline helps

Really these are more or less the same reasons it's convenient to have selectmode as a mode so it seemed a lot easier to use the existing functionality rather than add something new just for debug mode.

I guess you worry about the need to modify commands that match on the mode enum? Perhaps we could just add a `is_insert` function to the mode enum that returns true for insert mode and false otherwise and just forbid matching comparimg the mode enum directly?

filipdutescu:
@pascalkuthe summed up perfectly both the advantages and the reasoning behind this proposal. I would like to add the fact that it is, currently, also very jarring, in my opinion. There is no other sticky popup that I have seen in the editor, besides the debug one. This, for me, makes it feel unnatural and somewhat unfinished or experimental (which debugging is _*right now*_) compared to the rest of the UX I get.

[...]

PR #7632, opened 2023-07-14
Dedicated mode for sticky and debug

Marsevil:
- [ ] Rework sticky mode
- [ ] Rework debug mode


the-mikedavis (changes requested):
Adding items to the mode enum is not the direction we want to go: https://github.com/helix-editor/helix/issues/5951#issuecomment-1427226973

Marsevil:
That way, I don't understand this one https://github.com/helix-editor/helix/issues/5951#issuecomment-1437983827
May we want something like "submodes" which are trait object and so add keybinds and so on ?

PR #7633, opened 2023-07-14
Add a more file types for R

dgkf:
Registers extra R ecosystem files

- `.Rprofile` & `Rprofile.site` files, which are configured in R
- `Renviron` & `.Renviron` files, which are [GNU configuration files](https://www.gnu.org/prep/standards/html_node/Configuration.html), a syntax that is a subset of the existing `bash` syntax
- `Rd` documentation files, which use `infotex`, a subset of LaTeX
- `.Rbuildignore` (sort of like a `.gitignore` for R packages), which is defined as `regex`s

Quite a few files in the R ecosystem use the Debian Control File (DCF) format, but there isn't an existing language to couple that to, so I'll leave that for a separate PR.

(using [R administration guide as reference](https://cran.r-project.org/doc/manuals/R-admin.html))

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1324,7 +1324,7 @@ source = { git = "https://github.com/the-mikedavis/tree-sitter-git-rebase", rev 
 name = "regex"
 scope = "source.regex"
 injection-regex = "regex"
-file-types = ["regex"]
+file-types = ["regex", ".Rbuildignore"]
 roots = []
 
 [[grammar]]
--- a/languages.toml
+++ b/languages.toml
@@ -715,7 +715,7 @@ source = { git = "https://github.com/tree-sitter/tree-sitter-ruby", rev = "206c7
 name = "bash"
 scope = "source.bash"
 injection-regex = "(shell|bash|zsh|sh)"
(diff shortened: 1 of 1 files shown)


languages.toml
```diff
@@ -1324,7 +1324,7 @@ source = { git = "https://github.com/the-mikedavis/tree-sitter-git-rebase", rev
 name = "regex"
 scope = "source.regex"
 injection-regex = "regex"
-file-types = ["regex"]
+file-types = ["regex", ".Rbuildignore"]
```
the-mikedavis, line 1327: Regex is almost entirely used for injections, so `/foo/` from javascript can inject the language for example. `regex` as a file-type is just here for testing and could probably be removed. This file-type probably belongs under git-ignore instead

dgkf: Ah, I see. Unfortunately the `.Rbuildignore` syntax isn't interchangeable with the `.gitignore` syntax - it's just simply one-regex-per-line and doesn't support the `.gitignore`-style `**/*` path wildcards or the `!` negation syntax.

Even if the `regex` config wasn't initially intended to support files, it seems to work quite well for this case. If there isn't a more appropriate option, I'd be in favor of leaving it as is for now even if it is eventually removed. 

If you disagree, though, I'd be fine removing it also. Your call.

the-mikedavis: I thought this grammar might struggle with newlines and multiple regexes but if it works well then I think it's ok 👍

the-mikedavis (approved):

pascalkuthe (approved):

PR #7635, opened 2023-07-15
Always visible statusline (under prompt)

### LINKED ISSUE #5758: Allow to position status line on the top

woojiq:
Closes #5758
This PR is a duplicate of #5882 , but not stale.

For now, I quickly fixed it with a non-ideal config name and a strange code style that might be better commented out or simplified.
In #5882 I read that some want it to be a configurable option, some want a permanent layout. So I need to hear the "final" decision from the boss :smiley: 

Demo: https://asciinema.org/a/M3np9Q5K6c9OtShFhadtPASjT

TODO:
- [ ] Choose a better name (always-visible?)
- [ ] Default option true/false?
- [ ] More understandable code (is it understandable without comments? should I add comments? )
- [ ] Rework tests if these changes affect some

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -84,18 +84,20 @@ separator = "│"
 mode.normal = "NORMAL"
 mode.insert = "INSERT"
 mode.select = "SELECT"
+under-prompt = true
 ```
 The `[editor.statusline]` key takes the following sub-keys:
 
-| Key           | Description | Default |
-| ---           | ---         | ---     |
-| `left`        | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name"]` |
-| `center`      | A list of elements aligned to the middle of the statusline | `[]` |
-| `right`       | A list of elements aligned to the right of the statusline | `["diagnostics", "selections", "position", "file-encoding"]` |
-| `separator`   | The character used to separate elements in the statusline | `"│"` |
-| `mode.normal` | The text shown in the `mode` element for normal mode | `"NOR"` |
-| `mode.insert` | The text shown in the `mode` element for insert mode | `"INS"` |
-| `mode.select` | The text shown in the `mode` element for select mode | `"SEL"` |
+| Key            | Description | Default |
+| ---            | ---         | ---     |
+| `left`         | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name"]` |
+| `center`       | A list of elements aligned to the middle of the statusline | `[]` |
+| `right`        | A list of elements aligned to the right of the statusline | `["diagnostics", "selections", "position", "file-encoding"]` |
+| `separator`    | The character used to separate elements in the statusline | `"│"` |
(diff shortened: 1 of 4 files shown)


the-mikedavis:
Could we solve #5758 by rendering the prompt's completion options one higher so that it doesn't overlap with the statusline instead?

| Before | After |
|---|---|
| before | after |

```diff
diff --git a/helix-term/src/ui/prompt.rs b/helix-term/src/ui/prompt.rs
index a9ccfb73..2f98bfbd 100644
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -375,7 +375,7 @@ pub fn render_prompt(&self, area: Rect, surface: &mut Surface, cx: &mut Context)
 
         let completion_area = Rect::new(
             area.x,
-            (area.height - height).saturating_sub(1),
+            (area.height - height).saturating_sub(2),
             area.width,
             height,
         );
```

I agree with the comments in #5882 about not wanting a configuration option to change the UI like this and I think prompt above statusline is an odd look.

pascalkuthe:
I peronally also don't think we should make the UI configurable like this but I don't love the look above the statusline either. I prefer the UI to remain unchanged here. I think the usecase in https://github.com/helix-editor/helix/issues/5758 is a bit niche (being able to see the filename). I think its reasonable to just remember the buffername.

If the buffer is complex then the current file regist (#6985) or placehoders like `#filename` are a better fit

woojiq:
I fully agree with @pascalkuthe . In case it is still necessary, let me know. I gonna close PR.

quomaz:
> I peronally also don't think we should make the UI configurable like this but I don't love the look above the statusline either. I prefer the UI to remain unchanged here. 

@pascalkuthe When speaking of "we shouldn't make UI configurable like this" you mean that the proposed configuration option here specifically or that having ui configurable at all is bad?

I could imagine the whole layout customizable, allowing things like having the command line on the top for example. I personally really like the look proposed here, with upper command line, so I believe it would be useful to users.

pascalkuthe:
https://github.com/helix-editor/helix/pull/5882#issuecomment-1424007848)

We just don't want to make the UI that configurable and in general prefer to add as few config options as possible since that always means more maintenance effort

zetashift:
> We just don't want to make the UI that configurable and in general prefer to add as few config options as possible since that always means more maintenance effort

Completely understand the motivation for this and I agree.

But just out of curiosity: is the "UI being more configurable through a (potential) GUI or through a plugin system" an idea that is plausible?

pascalkuthe:
theoretically sure but that is something so distant in the future its not even worth talking about yet.

PR #7644, opened 2023-07-16
switch to https://github.com/tree-sitter-perl/tree-sitter-perl

### LINKED ISSUE #7201: Perl grammar is old and unreliable
### Summary

The default configuration points to using a tree-sitter grammar for Perl which in my experience crashes frequently, and [is discouraged by related programmers](https://github.com/ganezdragon/tree-sitter-perl/pull/12#issuecomment-1481754928).

Please consider switching to https://github.com/tree-sitter-perl/tree-sitter-perl as recommended in about comment. It has recent(!) commits from core Perl developers (oalders, leonerd).

### Reproduction Steps

I have (briefly) tested that the proposed grammar works succesfully - after patching Helix to support it (see ~#7198~ #7202).

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty 0.11.0

### Helix Version

helix 23.05

doy:
fixes #7201


runtime/queries/perl/highlights.scm
the-mikedavis, line 1: We use these scopes for highlights: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

So some of these will need to be adjusted, for example `@number` becomes `@constant.numeric`

runtime/queries/perl/highlights.scm
```diff
@@ -1,181 +1,87 @@
...
+(method_call_expression invocant: (bareword) @type)
 
-(function_attribute) @label
+(func0op_call_expression function: _ @function.builtin)
+(func1op_call_expression function: _ @function.builtin)
 
-(function_signature) @type
+(function) @function
 
+(ERROR) @error
```
the-mikedavis, line 87:
```suggestion
```

In the future we'd like to add an option for error highlighting but by default we leave it off because it can be visually distracting

runtime/queries/perl/injections.scm
the-mikedavis, line 1: The existing injection for `comment` is still correct if you change the node from `(comments)` to `(comment)`. The injection for pod can be removed since we don't have a pod language defined. If there was a pod language though it would take this form:

```scm
((pod) @injection.content
 (#set! injection.language "pod"))
```

following tree-sitter's way of doing injections: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection

doy:
how's this?

runtime/queries/perl/highlights.scm
```diff
@@ -1,181 +1,100 @@
...
+(hash) @variable.hash
+(hash_deref_expression ["->" "%" "*"] @variable.hash)
+
+(array_element_expression [array:(_) "->" "[" "]"] @variable.array)
+(slice_expression [array:(_) "->" "[" "]"] @variable.array)
+(keyval_expression [array:(_) "->" "[" "]"] @variable.array)
+
+(hash_element_expression [hash:(_) "->" "{" "}"] @variable.hash)
+(slice_expression [hash:(_) "->" "[" "]"] @variable.hash)
+(keyval_expression [hash:(_) "->" "[" "]"] @variable.hash)
```
the-mikedavis, line 75: Each of these should just be `@variable`

the-mikedavis (commented):
You will need to run `cargo xtask docgen` and commit the changes to fix the CI

doy:
looks like it's passing now!

the-mikedavis (approved):

archseer:
Hmm, this removed all the indentation queries though, so perl now doesn't support formatting anymore. I'll leave it for a follow-up

PR #7656, opened 2023-07-17
Update my name README.md

jakenvac:
I changed my username.

Removing it entirely is also fine if that's preferable. Credit isn't necessary - just making sure there are no dead links anywhere.

Thanks! ❤️


jakenvac:
Thanks @archseer <3

PR #7658, opened 2023-07-17
fix(path): change normalization strategy

### LINKED ISSUE #7458: Bug: A symlink that points to a file with with an unrecognizable file extension should fallback to the symlinks file extension
### Summary

The file extension of a symlink should be used as a file type hint, when the one on the destination file does not help.
This is in regards to syntax highlighting.

### Reproduction Steps

Minimum reproducible example:
I have two files, one real one name `test.c.1` with my c source code
and a symlink to that previous file named `test.c`.

When I open test.c.1 there is no syntax highlighting which is expected since there is not hint for the editor, however when I open test.c (symlink to test.c.1), I would expect it to open with c syntax highlighting (at least as a fallback).

This expected behavior is the current behavior of other popular editors (Ex: neovim)

### Helix log

```
```

</details>

### Platform

Linux + MacOs

### Terminal Emulator

Iterm2 Build 3.4.19

### Helix Version

helix 23.05 (7f5940be)

woojiq:
Closes #7458 
If the file does not have a language detected but has a symbolic link, try using symlink file extension to set the language. This applies to opening new files as well as previewing them in the file picker.

Demo: https://asciinema.org/a/KISAzgMSh7ToTcZ3yKme5f0hA
I run the patched helix for the first time, where you can see at the bottom of the screen that the canonicalized path is a file with no extension (+ syntax hightlighting).
I am using the master helix for the second time (- syntax hightlighting)

TODO:
- [x] Add tests with symlinks

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1,6 +1,8 @@
 # Language support configuration.
 # See the languages documentation: https://docs.helix-editor.com/master/languages.html
 
+use-grammars = { except = [ "hare", "wren", "gemini" ] }
+
 [language-server]
 
 als = { command = "als" }


helix-core/src/path.rs
```diff
@@ -30,15 +30,7 @@ pub fn expand_tilde(path: &Path) -> PathBuf {
...
-/// Normalize a path, removing things like `.` and `..`.
-///
-/// CAUTION: This does not resolve symlinks (unlike
-/// [`std::fs::canonicalize`]). This may cause incorrect or surprising
-/// behavior at times. This should be used carefully. Unfortunately,
-/// [`std::fs::canonicalize`] can be hard to use correctly, since it can often
-/// fail, or on Windows returns annoying device paths. This is a problem Cargo
-/// needs to improve on.
-/// Copied from cargo: <https://github.com/rust-lang/cargo/blob/070e459c2d8b79c5b2ac5218064e7603329c92ae/crates/cargo-util/src/paths.rs#L81>
+/// Normalize a path (removing things like `.` and `..`) and resolving symlinks.
```
woojiq, line 33: I removed the `caution` because under the hood it still recognizes symlinks and uses `std::fs::canonicalize`.
https://github.com/helix-editor/helix/blob/2ceef99a8dfea19283a5bdb98d768a28136ef264/helix-core/src/path.rs#L40-L40

helix-view/src/document.rs
```diff
@@ -141,6 +141,8 @@ pub struct Document {
     pub inlay_hints_oudated: bool,
 
     path: Option<PathBuf>,
+    /// Path without normalization or canonicalization
+    raw_path: Option<PathBuf>,
```
woojiq, line 145: I could have done without changing the existing signatures, but it looked terrible.

helix-view/src/document.rs
```diff
@@ -939,6 +945,11 @@ impl Document {
         config_loader
             .language_config_for_file_name(self.path.as_ref()?)
             .or_else(|| config_loader.language_config_for_shebang(self.text()))
+            .or_else(|| {
+                self.raw_path
+                    .as_ref()
+                    .and_then(|p| config_loader.language_config_for_file_name(p))
+            })
```
woojiq, line 952: If the canonicalized path and shebang don't recognize any language, we can try to guess with the initial path.

pascalkuthe:
I think this is actually an accidental/unintentional side effect introduced by  #6290. I think the real crux is that we don't actually want to resolve symbolic link in normalize_paths.

I think we can still keep the cannonicalization (which is the only way to correctly compare two paths) while making that work. We probably want to iterate the ancestors in reverse instead and check:
* if its a symlink => cannonicalize parent,  cannonicalize this path and rember the pair for later
* at the last ancestor that exists (so usually the full path)  and when encountering another symlink cannonicalize the path, strip the last symlink, strip the last symlink we encountered (which could be this path) and then join the result on top of the last saved parent 

It's a bit more subtle but definitly the right fix IMO

woojiq:
I'm not sure I understand you. Should we change canonicalization strategy?
For example we have `~/1/2/a.cpp` which is a symlink to `/3/4/b` what are steps? I can't going step by step with your explanation. 
  Maybe I overthink.

pascalkuthe:
Hmm let me show some exanoles in your case we would cannicolize  `~/1/2/` and then must houn the link name.


What I mean twith my description is that we basically want to cannonicalize the parent directory and then just join the link name on top...

Howver the parent directory (and any of its parents) could also be a symlink. So you would need to start at the bottom.

For example `/foo/link1/bar/link2/baz/link3.cpp` we would cannicolize `/foo` 

Then join `link1`, then cannonicalize the resukting oarh and remeber the re

Next we cannonicalize `/too/link1/bar` and strip_prefix the oath we remembered (so we just endup with `bar` again but cannonicalizd) and join that in too of our existing paths.

We repeat the sane thing for `link2` and `baz` (cannonicalize link parent, join link name, cannonicalize path, caboonicalize next link parent and strip prefix to join again).

Thks way we basically don't resolve symlinks while still normalizing filenames/paths correctly in every case

woojiq:
If I understand correctly, if we implement this, we should call `get_normalize_path` when opening files instead of `get_canonicalize_path`, but use the second one to check if we have already opened that file.

pascalkuthe:
I don't think that makes too much of a difference what functions are called when should not change, only  `get_normalized_path` should be changed as I described above

woojiq:
If we implement this and try to open the original file (not the symlink), we will have two buffers. VsCode and Vim behave this way, it seems. Is this the desired result or am I wrong?

pascalkuthe:
> If we implement this and try to open the original file (after symlink), we will have two buffers. VsCode and Vim behave this way, it seems. Is this the desired result or am I wrong?

that's at least how helix used to behave before the (unintentional) change to cannoncialize symlinks. Even in the past the cannoncialize function has never not resolved symlinks.

 A buffer should have a single path unique path. We should only ever use that path to determine the language of the file and interact with it in other way. Otherwise, you can symlink (`foo.c`, `foo.rs` and `foo.sh` all to `foo` and the language would be what? We also need to provide a single path to the LSP and the LSP can't backtrack that the file comes from a link so for symlinks to work with LSP we need to treat them as separate buffers.

So either we can always normalize symlinks so we only have one buffer per actual path or we can solve https://github.com/helix-editor/helix/issues/7458. We can't do both. And IMO the way helix used to behave was correct. The problem of potentially having multiple links open at once exists anyway (hardlinks) so I don't think that's that big of an issue (especially because we hide symlinks that point into the current workspace from the file picker so you have to put in some effort to open the same file twice)

woojiq:
Yeah, that's exactly what I was thinking. Very clear explanation. I just didn't know what the previous behavior was when opening a symlink and original file, so I assumed that maybe current behavior was desired. Now I got it, ty.

woojiq:
While changing the implementation, I found one interesting case.
E.g. we have `foo/bar` which is a symbolic link to `baz/`. And we have the path `smth/foo/bar/..`. What do we want to see as a result: `smth/foo/` or `smth/`? We don't want to resolve symlinks, so I suspect `smth/foo/` is the desired result?

I believe Pr is now ready for reviewing, but in a few days I will add tests with creating files and links to catch more pitfalls.

pascalkuthe:
> While changing the implementation, I found one interesting case. E.g. we have `foo/bar` which is a symbolic link to `baz/`. And we have the path `smth/foo/bar/..`. What do we want to see as a result: `smth/foo/` or `smth/`? We don't want to resolve symlinks, so I suspect `smth/foo/` is the desired result?
> 
> I believe Pr is now ready for reviewing, but in a few days I will add tests with creating files and links to catch more pitfalls.

I think in that case we actually want to keep the colons. `smth/` would be wrong (that's not how symlik resolution works) and whatever the parent of `baz` is would also be incorrect (since that would resolve the link). I think if you encounter colons you should manually handle them (instead of canonicalizing/calling dunce) by calling `path.pop`. If you reach a symlink that way you just keep the component instead (also you need to check if `path.pop` returns double colons in that case you had previously decided to keep these colons and need to keep the previous colons and the new colons (so that `smth/foo/../...` also canonicalizes correctly).

pascalkuthe (commented):
I think this lookgs good to me. Ideally we would reduce the number of cannonivalize calls a bit and try to sk.how do case folding on windows for symlinks too but that would involve pulling in the windows API and make the code more complex.


I think the edgecase of case folding symlinks on windows is rare enough to ignore (symlinks are rare on windows anyway) and the couple extra realpath syscalls shouldnt matter much

helix-core/tests/path.rs
```diff
@@ -0,0 +1,114 @@
...
+#[cfg(unix)]
+use std::os::unix;
+#[cfg(windows)]
+use std::os::windows;
+
+use helix_core::path::get_normalized_path;
+use tempfile::Builder;
+
+#[test]
+fn test_normalize_path() -> Result<(), Box<dyn Error>> {
```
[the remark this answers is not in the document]
woojiq: * Added in d9da1026bdd3ca172b78ee696dbefca1af0e06d4
* Modified my own test c20d041ab612773c8f44d701f9d717c31c3bb08b for `alt-ret` a bit, because in this test I was using a global search picker that uses `WalkDir` with `following_symlinks` under the hood, now I compare by filename instead of full path.

I'm also wondering is it bad at adding a crate with a different version than the version in the other workspace packages. I added the temporary file "3.7.0" to `helix-core`, but `helix-term` uses "3.7.1". Will dependabot equates them?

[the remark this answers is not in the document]
pascalkuthe: I had to drop the the for alt-ret in #7814 anyway because we will stream in picker results there so that test is always flakey (its essentially a race condition whether testcode or the streaming is faster). I can't really thing of a way to solve that so we decided to drop the test. I don't think you need to fix it here.

The version missmatche doesn't matter as as long as they are semver compatible Cargo will unify them anyway (if you check the `Cargo.lock` changes there is no actual new dependency)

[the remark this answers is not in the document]
woojiq: If I revert the test fix, the action will fail. If your PR merges first, I'll need just resolve one merge conflict.

woojiq:
Rebased on master.

helix-core/src/path.rs
```diff
@@ -30,30 +30,11 @@ pub fn expand_tilde(path: &Path) -> PathBuf {
...
-            let canonicalized_base = dunce::canonicalize(base).ok()?;
-            let remainder = path.strip_prefix(base).ok()?.into();
-            Some((canonicalized_base, remainder))
-        })
-        .unwrap_or_else(|| (PathBuf::new(), PathBuf::from(path)));
-
-    if path.as_os_str().is_empty() {
-        return base;
-    }
+    log::debug!("path before normalization: {}", path.to_string_lossy());
```
the-mikedavis, line 37: Should these debug logs be removed? I would guess that this could be quite noisy while running in verbose mode

woojiq: I don't remember how often this function is called, but if it's just on opening/saving, I think logs would be useful. Anyway, verbose mode is only run when you want to reproduce a bug. Normalization is a tricky thing, especially in windows.

woojiq: I see that lsp also calls this function. I can check logs in verbose mode to see how it bloats logs.

woojiq: You were right, they are very noisy. I opened one file and almost all logs were about normalization, xd. I removed them and rebased branch on master🤓

the-mikedavis (dismissed):

pascalkuthe:
One more edgecase for windows: You must use std cannonicalize instead of dunce and then only simplify the final path with https://docs.rs/dunce/latest/dunce/fn.simplified.html otherwise stripping the prefix may fail.

I think for unix I would actually prefer if we just reverted to the old handling (no cannonicalization at all) to avoid the (on slow FS potentially significant) overhead. Other unix programs (like vim, nvim or emacs) don't handle this either. I think on unix the cases where this would be useful are extremely nieche. This is mainly important for windows (casefolding)

woojiq:
Damn, Windows... I suggest adding `#[cfg(windows)]` for this block https://github.com/helix-editor/helix/blob/master/helix-core/src/path.rs#L43 and be happy 🙂
This PR looks like it shouldn't be solved like that, there may be something simpler.

pascalkuthe:
I considered that but I do think the points I made to motivate this PR still apply for windows. We want helix to correctly handle symlinks on windows too (or more accurately similarly as windows). I do think what you implemented here is good enough for windows (except the detail I mentioned). You just need to feature gate the new parent and normal match arms and add the old implementation back (plus the use of different function I mentioned above)

woojiq:
Wait what. How did I do it? I just pull from master, as usual 😅

PR #7665, opened 2023-07-18
Add initial support for LSP DidChangeWatchedFiles

ryanfowler:
Context: https://github.com/helix-editor/helix/issues/1125#issuecomment-1636831867

This PR implements initial support for LSP [DidChangeWatchedFiles](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_didChangeWatchedFiles) by forwarding a notification for all file writes or reloads to any LSP that has registered with a matching glob. This  work will enable LSPs to respond accordingly to changes made to any metadata files (e.g. `Cargo.toml`) that they have registered for notifications on. For now it can only forward notifications for changes that occur within Helix, however this code can continue to be used when a proper file watcher is added at some point in the future. 

For the implementation, a new file event `Handler` struct spawns a dedicated tokio task to listen for any registrations/unregistrations by LSPs as well as any file change events. Since the `Editor` struct owns the `helix-lsp::Registry` instance, I saw a few potential designs for the `Handler` to be able to access LSP clients. Ultimately, I chose to pass the `Handler` a `Weak` reference to the LSP clients as they register for file watching. This is done so that the `Handler` can have direct access to the client without preventing the client from being dropped if it is closed and isn't properly notified. Alternatives could include synchronizing references to clients between the `Editor` and `Handler`, synchronizing _access_ to the LSP clients, or notifying the `Editor` of file change notifications that need to be sent to LSPs via a new channel. My opinion is that all of those alternatives seem more complex or bug-prone.

There are certainly optimizations that can be made (e.g. batching file changes) but I think those can be added in future PRs, as necessary. 

I've tested these changes successfully with the rust-analyzer and gopls LSPs on both macos and Linux. Changes to the `Cargo.toml` of a crate are immediately recognized by rust-analyzer, instead of having to restart the LSP.

Thanks!


pascalkuthe (changes requested):
This looks really promising, thanks for taking a stab at this! I will do a more thorough review later. One thing that I noticed immidietly us that I would prefer the filer handlr to be moved to `helix-lsp`. You can store the handle in the registry (that's stored on the editor so you can still keep most of the rest the same)

pascalkuthe:
Also `:reload ` and `:reload-all` should send events too, they don't seem to at the moment.

ryanfowler:
Thanks for taking a look! I've moved the file event `Handler` to helix-lsp, as you've suggested.

> Also :reload  and :reload-all should send events too, they don't seem to at the moment.

Those should already be in place for :reload [here](https://github.com/helix-editor/helix/pull/7665/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1287-R1289) and :reload-all [here](https://github.com/helix-editor/helix/pull/7665/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1331-R1333).

pascalkuthe (changes requested):
some minor technical commints the rest looks good

pascalkuthe:
> Thanks for taking a look! I've moved the file event `Handler` to helix-lsp, as you've suggested.
> 
> > Also :reload  and :reload-all should send events too, they don't seem to at the moment.
> 
> Those should already be in place for :reload [here](https://github.com/helix-editor/helix/pull/7665/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1287-R1289) and :reload-all [here](https://github.com/helix-editor/helix/pull/7665/files#diff-8f2b243886fc16c4a33efa5bd4ef54824301ed882ef62878ad9d4f6188d3d6a1R1331-R1333).

ah sorry about that I somehow missed that in GH

helix-view/src/file_event.rs
```diff
@@ -0,0 +1,206 @@
...
+    async fn run(mut rx: mpsc::UnboundedReceiver<Event>) {
+        let mut state: HashMap<usize, ClientState> = HashMap::new();
+        while let Some(event) = rx.recv().await {
+            match event {
+                Event::FileChanged { path } => {
+                    log::debug!("Received file event for {:?}", &path);
+
+                    let mut dropped_clients = Vec::new();
+
+                    let futures = state
```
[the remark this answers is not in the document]
ryanfowler: Just pushed some some changes that are hopefully in-line with your ask here 🙂

helix-lsp/src/file_event.rs
```diff
@@ -0,0 +1,197 @@
...
+                Event::FileChanged { path } => {
+                    log::debug!("Received file event for {:?}", &path);
+
+                    let mut dropped_clients = Vec::new();
+                    for (id, client_state) in state.iter() {
+                        if client_state
+                            .registered
+                            .values()
+                            .any(|glob| glob.is_match(&path))
+                        {
```
pascalkuthe, line 103: small nit instead of nesting everything inside an if you can simply continue if the condition doesn't match

helix-view/src/file_event.rs
```diff
@@ -0,0 +1,206 @@
...
+                    }
+
+                    // Remove any clients that've been dropped but we still
+                    // have a weak reference to.
+                    if !dropped_clients.is_empty() {
+                        log::warn!("LSP clients were dropped: {:?}", dropped_clients);
+                        for id in dropped_clients {
+                            state.remove(&id);
+                        }
+                    }
```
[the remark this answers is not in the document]
pascalkuthe: you can achieve that using `HashMap::retain` (and returning false in that case and true otherwise)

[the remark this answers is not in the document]
ryanfowler: If I attempt to use `HashMap::retain` inside of the for loop I get the same `cannot borrow state as mutable because it is also borrowed as immutable` error, since it also requires a mutable reference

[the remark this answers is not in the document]
pascalkuthe: you can use `HashMap::retain` *instead* of the for loop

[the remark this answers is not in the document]
ryanfowler: Ahh okay, sorry I misunderstood! The problem I'm running into now is that the closure that `HashMap::retain` takes cannot be `async` (required to await the LSP notification) 😞

helix-lsp/src/file_event.rs
ryanfowler: Makes sense, code updated

helix-lsp/src/file_event.rs
```diff
@@ -0,0 +1,197 @@
...
+                            let Some(client) = client_state.client.upgrade() else {
+                                log::warn!("LSP client was dropped: {id}");
+                                dropped_clients.push(*id);
+                                continue;
+                            };
+                            if let Ok(uri) = lsp::Url::from_file_path(&path) {
+                                log::debug!(
+                                    "Sending didChangeWatchedFiles notification to client '{}'",
+                                    client.name()
+                                );
```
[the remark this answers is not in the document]
ryanfowler: I agree, code updated!

helix-view/src/file_event.rs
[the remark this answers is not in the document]
ryanfowler: 👍 that makes sense, just pushed that change!

pascalkuthe (dismissed):

helix-lsp/src/client.rs
```diff
@@ -543,6 +543,10 @@ impl Client {
                         normalizes_line_endings: Some(false),
                         change_annotation_support: None,
                     }),
+                    did_change_watched_files: Some(lsp::DidChangeWatchedFilesClientCapabilities {
+                        dynamic_registration: Some(true),
```
archseer, line 547: I think this should be false? We don't support dynamic registration at all

helix-lsp/src/lib.rs
```diff
@@ -1,9 +1,11 @@
 mod client;
+pub mod file_event;
 pub mod jsonrpc;
 pub mod snippet;
 mod transport;
 
 pub use client::Client;
+use file_event::Handler;
```
the-mikedavis, line 8: Small readability nit: rather than `use`ing these here, let's qualify the uses of `Handler` since its name is so generic.

helix-lsp/src/file_event.rs
```diff
@@ -0,0 +1,193 @@
...
+
+enum Event {
+    FileChanged {
+        path: PathBuf,
+    },
+    Register {
+        client_id: usize,
+        client: Weak<Client>,
+        registration_id: String,
+        ops: lsp::DidChangeWatchedFilesRegistrationOptions,
```
the-mikedavis, line 16:
```suggestion
        options: lsp::DidChangeWatchedFilesRegistrationOptions,
```

helix-view/src/editor.rs
```diff
@@ -1671,6 +1682,14 @@ impl Editor {
         &self,
         timeout: Option<u64>,
     ) -> Result<(), tokio::time::error::Elapsed> {
+        // Remove all language servers from the file event handler.
+        // Note: this is non-blocking.
+        for client in self.language_servers.iter_clients() {
+            self.language_servers
+                .file_event_handler
+                .remove_client(client.id());
```
the-mikedavis, line 1690: Does this clean-up need to be run in other places where clients are shutdown/removed like `Registry::shutdown`, `remove_by_id` and `restart`?

helix-view/Cargo.toml
```diff
@@ -45,7 +45,6 @@ log = "~0.4"
 which = "4.4"
 parking_lot = "0.12.1"
 
-
```
the-mikedavis, line 48: this change can be reset

the-mikedavis (commented):
I have a few style nits and a question but overall I think this looks good, I like the idea 👍

helix-view/src/editor.rs
ryanfowler: Great point! Originally the `file_event::Handler` lived on the `Editor` struct, but now that it's part of the `Registry`, I think it's best to do the cleanup from those `Registry` methods themselves. I've added `remove_client` calls to the methods: "remove_by_id", "restart", and "stop".

helix-lsp/src/lib.rs
ryanfowler: I agree, fixed!

ryanfowler:
@the-mikedavis thanks for the review! I think I've addressed all feedback, but let me know if there's anything else you'd like me to change 🙂

the-mikedavis (approved):

pascalkuthe (approved):

Issue #7669, opened 2023-07-18
Support more key bindings in windows, including f13-f24


AsafMah:
Windows has plenty of extra keys available in the virtual keycodes:
https://learn.microsoft.com/en-us/windows/win32/inputdev/virtual-key-codes

Which is useful for extended keyboards, or people like me who map CapsLock to F13 to avoid it.. capslocking.

Currently helix explicitly forbids F13 (there is a test for that).
It would be nice if it could support any virtual key, or at least F13-F24.

christianholman:
submitted a PR for this and updated the test as well, saw you removed good first issue so let me know if there's anything more involved there  - mostly doing this to get setup with the repo:>

christianholman:
> This should be a very small change to support. This bounds check just needs to be expanded to allow up to 24:
> 
> https://github.com/helix-editor/helix/blob/b47519ab11591b525132c5a13e464875b0c4f796/helix-view/src/input.rs#L382
> 
> I'm not sure the history of that line but supporting up to 24 seems fine to me.

submitted a PR for this and updated the test as well, saw you removed good first issue so let me know if there's anything more involved there - mostly doing this to get setup with the repo:>

archseer:
Closed by #7672

PR #7672, opened 2023-07-18
allow for higher F keys to be used

christianholman:
As mentioned in #7669, windows allows for higher function keys than F12.
Additionally, custom keyboard firmwares like ZMK support up to F24 https://zmk.dev/docs/codes/keyboard-keypad#f-keys


pascalkuthe (approved):
I can't test this since I don't use these keys but this seems reasonable to me. I think in the past `crossterm` didn't support these keys (its in the changelog as new) so that's probably why it fas forbidden in the past

Issue #5118, opened 2022-12-11
Support lsp "codeAction/resolve"


sigmaSd:
### Summary

What' supposed to happen:

- client -> "textDocument/codeAction" -> server
- server -> Available Actions -> client
- when user hits enter client -> "codeAction/resolve" -> server

The third step seems to be not made in helix

deno lsp uses this code action for refactoring (going from arrow function to normal function for example) 

It seems like this resolve method is used to improve performance, so by default the codeAction edits will be set to None unless the client try to resolve it https://github.com/denoland/deno/pull/9405

### Platform

linux

### Terminal Emulator

wezterm

### Helix Version

helix 22.12

sigmaSd:
https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve

PR #7677, opened 2023-07-19
feat: resolve code action

sigmaSd:
fix https://github.com/helix-editor/helix/issues/5118

tested with deno (works)

I imagine the code is not ideal since I'm blocking for a future inside the callback to wait for the resolver also I had to clone the code action

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -954,6 +954,24 @@ impl Client {
         Some(self.call::<lsp::request::ResolveCompletionItem>(completion_item))
     }
 
+    pub fn resolve_code_action(
+        &self,
+        code_action: lsp::CodeAction,
+    ) -> Option<impl Future<Output = Result<Value>>> {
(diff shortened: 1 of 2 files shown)


sigmaSd:
As far as I can tell, I can't nest jobs callback, I searched the code base and there is no other instance of nesting callbacks

There seems to be a similar lsp resolve call, but its registered to be triggered with idle timeout https://github.com/helix-editor/helix/blob/master/helix-term/src/ui/editor.rs#L1010

sigmaSd:
Another similar thing is this https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs#L4513

set_completion does block internally (it uses tokio_blockon internally in Completion::new to resolve items)

pascalkuthe:
hmm right I forgot this was already inside a box, I have a patch locally to allow that but its a bit more involved so that's probably ok for now

pascalkuthe (approved):

the-mikedavis (approved):

PR #7687, opened 2023-07-20
Popups UI: add hint about how to scroll

ilyagr:
For popups that have a scrollbar, this adds a hint to
`Use C-u & C-d to scroll`. The hint disappears after
the user's first `C-u` or `C-d`.

Without a hint, I found it difficult and frustrating to discover how
to scroll, since any wrong attempt caused the popup to disappear.

The hint uses the theme `ui.virtual.inlay-hint` to look like a hint.
It looks like this (the whitespace on the right is unrelated to the hint,
there's a long line in the popup if it's scrolled down): 
Screenshot of hint

I tried using markdown for the hint, but it looked too similar to the
text inside the popup and was harder to position correctly for some
reason.

**Update:** I noticed that some themes don't support inlay_hints very well. I'm considering adding inlay_hints to a few more themes (especially the base16 ones). Alternatively, we could switch to the `comment` theme, but it looks less good in many fancy themes (like the default one, shown in the screenshot).

### DIFF
--- a/helix-term/src/ui/popup.rs
+++ b/helix-term/src/ui/popup.rs
@@ -3,7 +3,10 @@ use crate::{
     compositor::{Callback, Component, Context, Event, EventResult},
     ctrl, key,
 };
-use tui::buffer::Buffer as Surface;
+use tui::{
+    buffer::Buffer as Surface,
+    widgets::{Paragraph, Widget},
+};
 
 use helix_core::Position;
 use helix_view::{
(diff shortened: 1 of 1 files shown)


ilyagr:
I noticed that some themes don't support inlay_hints very well. I'm considering adding inlay_hints to a few more themes (especially the base16 ones). Alternatively, we could switch to the `comment` theme, but it looks less good in many fancy themes (like the default one, shown in the screenshot).

ilyagr:
It's been a bit more than a month and it seems like there is some demand for this feature, so I thought I'd ping the thread. Let me know if there is anything I can do to make this easier to review. Alternatively, feel free to treat this as a feature request.

pascalkuthe:
closing this as stale, I don't think we will be adding this level of help text to the editor. That kind of thing belongs into the documentation.

Thank you for contributing!

PR #7692, opened 2023-07-20
Add asm (assembler) language support based on tree-sitter-asm

devyn:
The tree-sitter-asm syntax is more suitable for generic assembly language, including GNU assembler syntax, than the NASM-specific one.

This changes extensions for NASM so that asm syntax is used for ".s" and ".S" files, which are commonly used with the GNU assembler, and ".asm" and ".nasm" for NASM which are more commonly used with that.

---

The only thing I'm not sure about is tree-sitter-asm had highlight queries in the package already, so I just copied those but appended the LICENSE file to the beginning. Since it's MIT license I hope this is not a problem, but unfortunately I'm not the author of tree-sitter-asm so I can't help you with that. It would probably be more ideal if helix could also include highlight queries from git while building the grammars rather than having to be in the main helix repo.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2474,10 +2474,23 @@ name = "po"
 source = { git = "https://github.com/erasin/tree-sitter-po", rev = "417cee9abb2053ed26b19e7de972398f2da9b29e" }
 
 [[language]]
+name = "asm"
+scope = "source.asm"
+injection-regex = "asm"
+file-types = ["s", "S"]
+comment-token = "#"
+indent = { tab-width = 8, unit = "        " }
+roots = []
+
+[[grammar]]
+name = "asm"
+source = { git = "https://github.com/rush-rs/tree-sitter-asm.git", rev = "36dc26acc7818920de2e103e20a9f42358caf926" }
+
+[[language]]
 name = "nasm"
 scope = "source.nasm"
-file-types = ["asm", "s", "S", "nasm"]
-injection-regex = "n?asm"
+file-types = ["asm", "nasm"]
+injection-regex = "nasm"
 roots = []
 comment-token = ";"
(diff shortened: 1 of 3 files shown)


languages.toml
```diff
@@ -2473,11 +2473,24 @@ comment-token = "#"
...
+scope = "source.asm"
+injection-regex = "asm"
+file-types = ["s", "S"]
+comment-token = "#"
+indent = { tab-width = 8, unit = "        " }
+roots = []
+
+[[grammar]]
+name = "asm"
+source = { git = "https://github.com/rush-rs/tree-sitter-asm.git", rev = "36dc26acc7818920de2e103e20a9f42358caf926" }
```
the-mikedavis, line 2487:
```suggestion
source = { git = "https://github.com/rush-rs/tree-sitter-asm", rev = "36dc26acc7818920de2e103e20a9f42358caf926" }
```

runtime/queries/asm/highlights.scm
the-mikedavis, line 1: The highlights should use helix's capture names: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

It looks like they're currently using nvim-like captures. For example `(int) @number` should be `(int) @constant.numeric.integer`.

runtime/queries/asm/highlights.scm
```diff
@@ -0,0 +1,54 @@
...
+; General
+(label (ident) @label)
+(reg) @variable.builtin
+(meta
+  kind: (_) @function.builtin)
+(instruction
+  kind: (_) @function.call)
+
+; Comments
+(line_comment) @comment @spell
```
the-mikedavis, line 32:
```suggestion
(line_comment) @comment
```

PR #7693, opened 2023-07-21
chores: run `cargo fmt`, fix Clippy warnings

ilyagr:
`cargo fmt` recently learned to format `let else` blocks, and reformats a few of them.

Two of the clippy warnings look like false positives, most of the rest were fixed with `cargo clippy --fix`.

I ran everything with `+nightly`, but am not sure which version actually ran because of `rust-toolchain.toml`.

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -106,12 +106,16 @@ fn find_pair(
         for close in
             iter::successors(node.next_sibling(), |node| node.next_sibling()).take(MATCH_LIMIT)
         {
-            let Some(open) = as_close_pair(doc, &close) else { continue; };
+            let Some(open) = as_close_pair(doc, &close) else {
+                continue;
+            };
             if find_pair_end(doc, Some(node), open, Backward).is_some() {
                 return doc.try_byte_to_char(close.start_byte()).ok();
             }
         }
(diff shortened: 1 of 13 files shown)


archseer:
Nightly definitely ran because you have errors on our MSRV.

ilyagr:
Good point, fixed.

Note that the [formatting changes](https://github.com/helix-editor/helix/compare/970f35a3b732396494e85161b57ca11bac4edf21..d4322e4c0d008ad38e972d4df904ad06c03b7023) that the older `cargo fmt` does are awful. I wonder if it would be better to run a newer version of `cargo fmt` than the MSRV by default, but it's up to you.

Fortunately, all the `let-else` changes are accepted by both versions of `fmt`.

If you prefer to make these changes when you update the MSRV, we don't have to merge this.

the-mikedavis:
Yeah I think it makes sense to only make these changes as we update the MSRV. Fixing future clippy lints is usually fine to do ahead of time but some of the formatting changes here will change again when we update the MSRV.

Running a formatter that is different than the MSRV seems unnecessarily complicated to me. Needing to install multiple versions of rust just to be able to contribute is not worth the improved formatting IMO.

ilyagr:
OK, sounds good.

Issue #7715, opened 2023-07-22
Disable searching for git/ignore files for path completion


sigmaSd:
Currently helix uses default git/ignore searching options for file walker https://github.com/helix-editor/helix/blob/master/helix-term/src/ui/mod.rs#L484-L488

This searching for git/ignore files issues 14 syscalls total per directory instead of 3 without them 
disable with
```rs
        .git_ignore(false)
        .git_exclude(false)
        .ignore(false)
```

As an example if a folder contains 287 files it issues **4110** syscalls instead of **926** , this in my testing with cold cache change the time from **15 second** to **100 ms** (in an old pc with hdd)

Also note that this searching is useless because we're already searching in depth ==1, so the only thing needed is the git/ignore files inside the current directory and *not* the sub-directories, (so 9 extra syscalls instead of 3186)

This is currently configurable in the file picker, but not i the path completer

What do I propose:
- make git/ignore also configurable in path completer (seems reasonable)
- change the defaults in file picker and path completer to not do this extra search by default (debatable)

Another idea is since the search is for depth 1 in case of completer, we can keep the same functionality while gaining all the performance, by disabling the git/ingore search, and manually filtering the git/ignore entries only in the current directory

sigmaSd:
ref https://github.com/helix-editor/helix/issues/6867 https://github.com/helix-editor/helix/issues/6114 https://github.com/helix-editor/helix/issues/5871

pascalkuthe:
I am strongly opposed to changing the default for the file picker. Gitignore is a very useful feature and the overhead is negligebale since its a recursive crawl anyway. All the issues you linked regarding the file licker are unrelated

Something you left out here but mentioned on latex. This was a system with an old HDD and a cold cache. On a modern system the overhead is not very significant

norcalli:
@pascalkuthe You're forgetting about NFS, which many of us have to use for work reasons and is definitely *not* negligible.

For me, every keystroke is slow and I checked via strace that it is in fact invoking this recursive search every time the path changes, not caching the information despite the path being appended to in 99% of cases, meaning that all of the ancestor git ignore information should probably still be valid.

Yutsuten:
I'd love to have `:open` not ignoring files in `.gitignore` for a different reason.

[...]

PR #7716, opened 2023-07-22
feat: make filename_impl walker configurable

sigmaSd:
ref https://github.com/helix-editor/helix/issues/7715

couple of notes:
- I'm using file_picker config , is that a good idea or should filename have its own config (which will need duplicating FileConfig struct)
- Should I copy all the configs from https://github.com/helix-editor/helix/blob/master/helix-term/src/commands.rs#L2094 I currently only added the options that impact performance

### DIFF
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -441,7 +441,7 @@ pub mod completers {
     }
 
     // TODO: we could return an iter/lazy thing so it can fetch as many as it needs.
-    fn filename_impl<F>(_editor: &Editor, input: &str, filter_fn: F) -> Vec<Completion>
+    fn filename_impl<F>(editor: &Editor, input: &str, filter_fn: F) -> Vec<Completion>
     where
         F: Fn(&ignore::DirEntry) -> FileMatch,
     {
@@ -481,10 +481,14 @@ pub mod completers {
 
         let end = input.len()..;
 
+        let config = editor.config();
         let mut files: Vec<_> = WalkBuilder::new(&dir)
             .hidden(false)
             .follow_links(false) // We're scanning over depth 1
             .max_depth(Some(1))
+            .ignore(config.file_picker.ignore)
+            .git_ignore(config.file_picker.git_ignore)
+            .git_exclude(config.file_picker.git_exclude)
             .build()
             .filter_map(|file| {
                 file.ok().and_then(|entry| {


sigmaSd:
adding another config option might look like this
```rs
diff --git a/helix-term/src/ui/mod.rs b/helix-term/src/ui/mod.rs
index 1dc3ddac..897ad4aa 100644
--- a/helix-term/src/ui/mod.rs
+++ b/helix-term/src/ui/mod.rs
@@ -486,9 +486,9 @@ fn filename_impl<F>(editor: &Editor, input: &str, filter_fn: F) -> Vec<Completio
             .hidden(false)
             .follow_links(false) // We're scanning over depth 1
             .max_depth(Some(1))
-            .ignore(config.file_picker.ignore)
-            .git_ignore(config.file_picker.git_ignore)
-            .git_exclude(config.file_picker.git_exclude)
+            .ignore(config.path_completer.ignore)
+            .git_ignore(config.path_completer.git_ignore)
+            .git_exclude(config.path_completer.git_exclude)
             .build()
             .filter_map(|file| {
                 file.ok().and_then(|entry| {
diff --git a/helix-view/src/editor.rs b/helix-view/src/editor.rs
index 20469ae9..94f16b5b 100644
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -260,6 +260,8 @@ pub struct Config {
     /// Whether to display infoboxes. Defaults to true.
     pub auto_info: bool,
     pub file_picker: FilePickerConfig,
+    // we don't use all the fields of FilePickerConfig
+    pub path_completer: FilePickerConfig,
     /// Configuration of the statusline elements
     pub statusline: StatusLineConfig,
     /// Shape for cursor in each mode
@@ -801,6 +803,12 @@ fn default() -> Self {
             completion_trigger_len: 2,
             auto_info: true,
             file_picker: FilePickerConfig::default(),
+            path_completer: FilePickerConfig {
+                ignore: false,
+                git_ignore: false,
+                git_exclude: false,
+                ..Default::default()
+            },
             statusline: StatusLineConfig::default(),
             cursor_shape: CursorShapeConfig::default(),
             true_color: false,
```

sigmaSd:
If  https://github.com/BurntSushi/ripgrep/pull/2566 lands, this PR wont be needed

sigmaSd:
It would be great if this get reviewd, the PR on ripgrep seems to take a while

The motivation is with nucleo space f have no lag but using `:o` is basically impossible an my old PC, at least with this option I can fix it

PR #7718, opened 2023-07-22
Change operator color to brown in gruvbox theme

rsjethani:
This would help differentiating operators from constants and builtins.
We do this by adding new `brown` color to the palette and using it as
foreground color for operators.

Go based example:

Before:

Screenshot from 2023-07-23 00-33-47

After:

Screenshot from 2023-07-23 00-33-16

### DIFF
--- a/runtime/themes/gruvbox.toml
+++ b/runtime/themes/gruvbox.toml
@@ -7,7 +7,7 @@
 "namespace" = "aqua1"
 "punctuation" = "orange1"
 "punctuation.delimiter" = "orange1"
-"operator" = "purple1"
+"operator" = "brown"
 "special" = "purple0"
 "variable.other.member" = "blue1"
 "variable" = "fg1"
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
This `gruvbox` tries to follow the upstream gruvbox as closely as possible: https://github.com/morhetz/gruvbox

The new brown isn't in the palette as far as I can see https://github.com/morhetz/gruvbox/blob/bf2885a95efdad7bd5e4794dd0213917770d79b7/colors/gruvbox.vim#L89-L132. It looks like upstream highlights operators as fg1: https://github.com/morhetz/gruvbox/blob/bf2885a95efdad7bd5e4794dd0213917770d79b7/colors/gruvbox.vim#L590-L591, https://github.com/morhetz/gruvbox/blob/bf2885a95efdad7bd5e4794dd0213917770d79b7/colors/gruvbox.vim#L465.

PR #7721, opened 2023-07-23
Reformat with nightly rustfmt for better let-else formatting

Philipp-M:
As discussed in https://github.com/helix-editor/helix/pull/4443#discussion_r1271457803

I ran a format with rust nightly over the code, and fixed all issues where MSRV rustfmt would do something different (so that CI runs through).

I think this makes sense, since I often (and I'm sure I'm not the only one) reformat code with whatever rustfmt version I'm currently using, and this introduces a lot of changes that are unwanted compared to the actual change I was doing. This should at least minify this issue (two changes are still different with nightly rustfmt, the infamous comment in `helix-view/src/editor.rs` and a let else in a macro (which I formatted manually)).

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -106,12 +106,16 @@ fn find_pair(
         for close in
             iter::successors(node.next_sibling(), |node| node.next_sibling()).take(MATCH_LIMIT)
         {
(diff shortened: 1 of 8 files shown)


pascalkuthe (approved):

ilyagr:
+1, this was my main motivation for https://github.com/helix-editor/helix/pull/7693.

PR #7732, opened 2023-07-25
fix(comment): fix incorrect predicate

lucario387:
These 2 captures were missing `#` so it's not working as intended

Was there any particular reason they were done as such?


the-mikedavis (approved):
Yep that's just an oversight. Thanks for the fix!

pascalkuthe (approved):

PR #7736, opened 2023-07-25
Update soft-wrap indicator in gruvbox.toml and themes derived from it

FedericoStra:
Hi!

I find the default brightness of the soft-wrap indicator to be too distracting. This dims it quite a bit, bringing it down almost to the background level.

Could such a change be considered for merging? I personally think it has a positive impact on the quality of gruvbox-derived themes.


the-mikedavis (approved):
This seems good to me 👍

pascalkuthe (approved):

PR #7738, opened 2023-07-25
Syntax-highlight regex prompts

the-mikedavis:
We can re-use the code-block highlighting function from the markdown component to highlight regex prompts with tree-sitter-regex. To make this change a little cheaper, we switch some functions from taking `&Rope`s to `RopeSlice`s (see 50557c62747e076e0c35f72c396e20d48d4e9052's message).

### DIFF
--- a/helix-term/src/ui/markdown.rs
+++ b/helix-term/src/ui/markdown.rs
@@ -10,7 +10,7 @@ use pulldown_cmark::{CodeBlockKind, Event, HeadingLevel, Options, Parser, Tag};
 
 use helix_core::{
     syntax::{self, HighlightEvent, InjectionLanguageMarker, Syntax},
-    Rope,
+    RopeSlice,
 };
 use helix_view::{
     graphics::{Margin, Rect, Style},
@@ -45,13 +45,13 @@ pub fn highlighted_code_block<'a>(
         None => return styled_multiline_text(text, code_style),
     };
 
-    let rope = Rope::from(text.as_ref());
+    let ropeslice = RopeSlice::from(text);
     let syntax = config_loader
         .language_configuration_for_injection_string(&InjectionLanguageMarker::Name(
             language.into(),
         ))
         .and_then(|config| config.highlight_config(theme.scopes()))
-        .and_then(|config| Syntax::new(&rope, config, Arc::clone(&config_loader)));
+        .and_then(|config| Syntax::new(ropeslice, config, Arc::clone(&config_loader)));
 
     let syntax = match syntax {
         Some(s) => s,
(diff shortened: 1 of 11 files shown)


helix-term/src/ui/prompt.rs
```diff
@@ -456,30 +465,28 @@ impl Prompt {
...
                 .and_then(|reg| cx.editor.registers.last(reg))
-                .map(|entry| entry.into())
             {
-                Some(value) => (value, true),
-                None => (Cow::from(""), false),
+                surface.set_string(line_area.x, line_area.y, suggestion, suggestion_color);
             }
+        } else if let Some((language, loader)) = self.language.as_ref() {
+            let mut text: ui::text::Text = crate::ui::markdown::highlighted_code_block(
+                self.line.clone(),
```
pascalkuthe, line 479: I think it should be possible to pass a `&str` instead of a `String` here (would require some additional changes but the pref concern is pretty similar to the Rope -> RopeSlice case. Its not a huge deal ofcourse but from a quick glimpse just changing the argument type should be enough

helix-term/src/ui/markdown.rs
```diff
@@ -45,22 +42,20 @@ pub fn highlighted_code_block<'a>(
         None => return styled_multiline_text(text, code_style),
     };
 
-    let rope = Rope::from(text.as_ref());
+    let rope = text.as_str().into();
```
archseer, line 45: What's the type here? I think this still ends up allocating a rope

runtime/queries/regex/highlights.scm
```diff
@@ -50,4 +50,3 @@
   ])
 
 (class_character) @constant.character
-(pattern_character) @string
```
archseer, line 53: Can we add the `(ERROR)` class so that invalid regexes are highlighted? That's something that has been a TODO in the prompt for a while, the idea was to highlight the whole line as red but I guess we can actually do better and use `(ERROR)` nodes

archseer: (This TODO: https://github.com/helix-editor/helix/blob/fb103eef3ef448c0983e1534d1a5dfae405c0deb/helix-term/src/ui/mod.rs#L147 kakoune just colors the whole line as red)

the-mikedavis: I think we could overlay some error highlights for regex syntax errors according to rust's regex without too much extra work. We could check the syntax with the `regex-syntax` dep and using the [location span](https://docs.rs/regex-syntax/0.7.4/regex_syntax/ast/struct.Error.html#method.span) in a syntax error, then merging those as the `additional_highlight_spans`: https://github.com/helix-editor/helix/blob/5a5289701429c2d7019d11d2184833745865063f/helix-term/src/ui/markdown.rs#L29-L35

pascalkuthe: Ah that's a really neat idea 👍 Although you need to be sure to convert byte/utf8 indeces to char/utf32 indices

PR #7740, opened 2023-07-26
Readonly status

Connortsui20:
These changes address issue [#2627](https://github.com/helix-editor/helix/issues/2627).

(This is my first PR ever so apologies if I messed anything up...)

Reading through the issues and the previous PRs that reference a read-only flag on the statusline, it seems like this was supposed to be implemented, but other things (related to the statusline) took priority and this ended up getting dropped.

I am well aware that this is probably not the best way to do this (as this does not let users choose what color they want the text to be), but as a quick solution this works well enough. I feel like people would rather be able to see "[readonly]" than not at all, especially newcomers to helix who are used to vim that might make a bunch of changes not realizing they were editing a read only file (this was me pretty recently).

(As a side note I don't know how to undo that first gitignore commit, so sorry about that!)


pascalkuthe (changes requested):
Your PR queries the FS on every single frame which is unacceptable. The FS access must be cached (when opening, reloading and writing the file) at a minimum.

Connortsui20:
Would storing it in the Document struct be okay?

pascalkuthe:
> https://github.com/helix-editor/helix/pull/7128#issuecomment-1581560237

yeah that's the idea conderidering we are storing mtime too it might make sense to have a dedicated struct for metada (essentially our own reduced `std::fs::Metadata`) that's how a lot of c code usually handles that. But for now adding an extra field would probably do.
 
Ideally, we would have more extensive read-only handling, see https://github.com/helix-editor/helix/pull/7128#issuecomment-1581560237 although a status line indicator that indicates the file is not writable might be a reasonable first step.

CC @the-mikedavis

Connortsui20:
In the future, instead of having a `readonly: Option<bool>` field, it could be a `metadata: Option<helix::Metadata>` where the struct contains the `readonly` field, and it might also be beneficial to have other fields like time of access, time of last modification, or even length of file and file type. These are all things that `std::fs::Metadata` gives out (documentation [here](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.permissions)), and I can imagine it being nice to have on the statusline as an user-configured option.

sigmaSd:
Vim also shows readonly flag for files that the user can't modify, like root files, which I think is actually the more common case then editing an actual readonly file.

sigmaSd:
vim uses `access` syscall on linux for this (not exposed by rust std), don't have windows to test but maybe it uses https://learn.microsoft.com/en-us/windows/win32/api/securitybaseapi/nf-securitybaseapi-accesscheck

pascalkuthe:
Using `access` would indeed be nicer, we could use `rustix` to execute that function (without unsafe code) altough it would only work on `[cfg(Unix)`. On windows we can use the windows-sys crate altoufh that may require unsafe and a utf16 conversion? We already depend on both transitively and these are extenmenly widely used projects with significant institutional backing so it's reaosnabke to pull them in here IMO

helix-term/src/ui/statusline.rs
```diff
@@ -432,12 +432,14 @@ fn render_file_modification_indicator<F>(context: &mut RenderContext, write: F)
 where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
-    let title = (if context.doc.is_modified() {
+    debug_assert!(!(context.doc.readonly.unwrap_or(false) && context.doc.is_modified()));
```
pascalkuthe, line 435: This debug assert is wrong. We have no eadonky editor mode right now. The document can definitly be modified and read-only (even editors that have a read-only mode allow that)

helix-term/src/ui/statusline.rs
```diff
@@ -432,12 +432,14 @@ fn render_file_modification_indicator<F>(context: &mut RenderContext, write: F)
...
+    let title = (if context.doc.readonly.unwrap_or(false) {
+        "[readonly]"
+    } else if context.doc.is_modified() {
```
pascalkuthe, line 438: Thus should be an else (see abovw)

helix-view/src/document.rs
```diff
@@ -185,6 +185,9 @@ pub struct Document {
 
     // when document was used for most-recent-used buffer picker
     pub focused_at: std::time::Instant,
+
+    // logically true when equal to Some(true); None implies we haven't checked yet
+    pub readonly: Option<bool>,
```
pascalkuthe, line 190: I don't think `None` should exist here. We should always recheck during certain actions since the FSA could change at any point. So we should just init to false here and the always rechdck

.gitignore
```diff
@@ -3,3 +3,4 @@ target
 helix-term/rustfmt.toml
 result
 runtime/grammars
+runtime
```
pascalkuthe, line 6: This change is unrelated and should be removed. The runtime files are part of the repo and should not be gitignored

pascalkuthe (changes requested):
Read-only status should be checked when sabing and reloading. Th disk can change at any time. Also the caching scheme is not necessary (see comments).


I also think that going the access route is likely nicer (see above)

.gitignore
Connortsui20: How do I undo this commit? Should I just delete that in a new commit have it be squashed later?

helix-term/src/ui/statusline.rs
Connortsui20: Should there be a warning message if a user is modified a readonly file? Vim has: `Warning: Changing a readonly file`

.gitignore
pascalkuthe: We always squash snakmer ORs kune this anyway so don't worry about it

helix-term/src/ui/statusline.rs
pascalkuthe: Yeah that what I mean with:

>  Ideally, we would have more extensive read-only handling, see https://github.com/helix-editor/helix/pull/7128#issuecomment-1581560237 although a status line indicator that indicates the file is not writable might be a reasonable first step.

But that would be a larger chabge. You can work in that jf you want ti but it would require some design discussion (look at what kakune does)  and would require more substantial changes si it might be better to keep that for future work

Connortsui20: I was thinking along the lines of this:
```
let title = match (context.doc.readonly, context.doc.is_modified()) {
    (false, false) => "   ",
    (false, true) => "[+]",
    (true, false) => "[readonly]",
    (true, true) => "Warning: Changing a readonly file",
    }.to_string();
```
If it's too ugly then I can leave it out but ugly is probably better here since the user probably wants to know this.

the-mikedavis:
I think a read-only indicator based on the fs metadata is a good idea but it should be its own statusline element rather than shared with the modification indicator. I think there should be two concepts for read-only: read-only files (fs-metadata-wise) which could just be a statusline element like this and a read-only way of opening files where you aren't allowed to modify them. Even if a file is read-only, you still might want to make changes and save those changes to a different writable path.

So for this PR I think a new statusline element is a nice change but the read-only way of interacting with files will need much larger changes.

sigmaSd:
Here is the code for vim windows access implementation https://github.com/vim/vim/blob/4c0089d696b8d1d5dc40568f25ea5738fa5bbffb/src/os_win32.c#L7665
It seems to boil down to `CreateFileW` + some tricks to handle directories + network shares

Connortsui20:
I'm not going to lie I have no clue what is going on with what vim is doing [here](https://github.com/vim/vim/blob/4c0089d696b8d1d5dc40568f25ea5738fa5bbffb/src/os_win32.c#L7665). However the unix access syscall seems to work well.

One  other thing that I don't know how to approach is what the behavior should be when `write-all` is called. In vim it writes everything that can be written, but then shows a red error message `'readonly' option is set (add ! to overwrite)`. 
The error message doesn't specify which file is readonly, only that there is one among the current buffers. 

It would be nice if it showed which specific file or files are readonly, as the above message is kind of vague. Also, while testing this in helix sometimes the error message would get covered by the `"file_name" written ...`, but I can't consistently reproduce that.

pascalkuthe:
> Here is the code for vim windows access implementation [vim/vim@`4c0089d`/src/os_win32.c#L7665](https://github.com/vim/vim/blob/4c0089d696b8d1d5dc40568f25ea5738fa5bbffb/src/os_win32.c#L7665) It seems to boil down to `CreateFileW` + some tricks to handle directories + network shares

yeah I don't think its fair to require that for merging. If you feel interested/up to it @Connortsui20 then go for it but honestly just using `access` from rustix on UNIX and using the standard metadata output on Windows is enough for me (leave a todo note for the future with a like tough). The access syscall is a very standard part of POSIX and widely used/avialable (and easy to use). The windows variant seems to me like POSIX emulation and not exactly standard. It could be nice to have in the future but it doesn't seem to be standard practice on windows (I am not even sure fi this kind of thing should live in helix or standalone crate).

>  One other thing that I don't know how to approach is what the behavior should be when write-all is called. In vim it writes everything that can be written, but then shows a red error message 'readonly' option is set (add ! to overwrite).
The error message doesn't specify which file is readonly, only that there is one among the current buffers.

Hmm I guess the issue with that is that there could be many readonly files but there is only so much space in the statusline.  It's probably fine to just print a list of butters here (that would be truncated if there is not enough space). Usually there are only few readonly files. 

> It would be nice if it showed which specific file or files are readonly, as the above message is kind of vague. Also, while testing this in helix sometimes the error message would get covered by the "file_name" written ..., but I can't consistently reproduce that.

Yeah that's probably expected since writing happens asycn so that message could come trough later or earlier (non-deteministically). That's a problem in general, we should likely improve the way we display status messages. You don't need to worry about that here tough, that can be left to future work since il likely needs a systematic change

pascalkuthe:
one thing that could also be nice is to show whether a file is readonly in the buffer picker. What do you think about that @the-mikedavis?

Connortsui20:
The readonly error messages are kind of verbose for `write-all`, but it sort of doesn't matter because you can't see it anyways if there is even a single writeable file among the buffers

Connortsui20:
Also I'm just going to mention #3709 because I don't really know what happened over there but it seems like similar changes were made then dropped

pascalkuthe:
> Also I'm just going to mention #3709 because I don't really know what happened over there but it seems like similar changes were made then dropped

we don't know why that author closed they PR we gave them feedback in a timely manner and weren't opposed to the changes. Some things will always remain a mystery I guess

helix-term/src/ui/statusline.rs
```diff
@@ -137,30 +137,28 @@ fn get_render_function<F>(element_id: StatusLineElementID) -> impl Fn(&mut Rende
...
+        E::WorkspaceDiagnostics => render_workspace_diagnostics,
+        E::Selections => render_selections,
+        E::PrimarySelectionLength => render_primary_selection_length,
+        E::Position => render_position,
+        E::PositionPercentage => render_position_percentage,
+        E::TotalLineNumbers => render_total_line_numbers,
+        E::Separator => render_separator,
+        E::Spacer => render_spacer,
+        E::VersionControl => render_version_control,
+        E::Register => render_register,
```
pascalkuthe, line 161: this change is unrelated and not nesccary for the PS, please don't include unrelated changes

helix-term/src/commands/typed.rs
```diff
@@ -334,6 +334,10 @@ fn write_impl(
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
 
+    if doc.detect_readonly() && !force {
+        bail!("'readonly' option is set (add ! to overwrite)")
```
pascalkuthe, line 338:
```suggestion
        bail!("'file is readonly' option use ! to overwrite")
```

helix-term/src/commands/typed.rs
```diff
@@ -334,6 +334,10 @@ fn write_impl(
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
 
+    if doc.detect_readonly() && !force {
```
pascalkuthe, line 337: this check should only fire if `path.is_none()` 
```suggestion
    if doc.detect_readonly() && !force $$path.is_none() {
```

That means that `:w` will fail but `:w <path>` won't fail (as that path is likely not read-only)

helix-term/src/commands/typed.rs
```diff
@@ -720,8 +735,19 @@ pub fn write_all_impl(
...
+    }
+
+    let mut error_msg = errors.join(" ");
+    if !readonly_errors.is_empty() {
+        // statusline text will have the form:
+        // ["'readonly' option is set for /home/user/readonly.txt", ...] (add ! to overwrite)
+        error_msg = format!("{:?} (add ! to overwrite)", readonly_errors);
+    }
+
+    if !error_msg.is_empty() && !force {
```
pascalkuthe, line 749:
```suggestion
    if !error_msg.is_empty() {
```
not needed herer since you bail early above

helix-term/src/commands/typed.rs
Connortsui20: I'll change it to `file is readonly, use ! to overwrite`, I assume that's what you meant

helix-term/src/ui/statusline.rs
Connortsui20: Should this be a separate PR? I did this to be consistent with [this](https://github.com/helix-editor/helix/blob/d5571968faf498f2bf12e18835755377d22e13a4/helix-view/src/editor.rs#L400) part of the code

helix-view/Cargo.toml
```diff
@@ -16,6 +16,7 @@ term = ["crossterm"]
 [dependencies]
 bitflags = "2.3"
 anyhow = "1"
+rustix = { version = "0.38", features = ["fs"] }
```
pascalkuthe, line 19: I think this should only be a `cfg(unix)' dependency (its possible to make dependencies target specific)

pascalkuthe:
I think the readonly statusline element should be added to the docs too and I haven't checked but IMO it should also be part of the default statusline so if its not part of the default yet you can add it

.gitignore
```diff
@@ -2,4 +2,4 @@ target
 .direnv
 helix-term/rustfmt.toml
 result
-runtime/grammars
+runtime/grammars
```
gabydd, line 5: Could you add back the newline here as well

Connortsui20:
Do I just need to change the docs in `configuration.md`? I'm not actually sure how the configuration in the `config.toml` pulls settings in, I'm assuming the logic is here but I'm not sure if I need to add something to translate a `read-only-indicator` to the enum datatype `StatusLineElement::ReadOnlyIndicator`... or is that being handled by serde?

pascalkuthe:
> Do I just need to change the docs in `configuration.md`? I'm not actually sure how the configuration in the `config.toml` pulls settings in, I'm assuming the logic is [here](https://github.com/helix-editor/helix/blob/d5571968faf498f2bf12e18835755377d22e13a4/helix-term/src/config.rs#L59) but I'm not sure if I need to add something to translate a `read-only-indicator` to the enum datatype `StatusLineElement::ReadOnlyIndicator` ... or is that being handled by `serde`?

default config is in the various `Default` implementations for the config structs

Connortsui20:
Okay then that is it, I had already added the element right under [here](https://github.com/helix-editor/helix/blob/d5571968faf498f2bf12e18835755377d22e13a4/helix-view/src/editor.rs#L407)

pascalkuthe (approved):

helix-term/src/ui/statusline.rs
```diff
@@ -433,15 +434,30 @@ where
     F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
 {
     let title = (if context.doc.is_modified() {
-        "[+]"
+        " [+] "
     } else {
-        "   "
+        ""
```
gabydd, line 439: What is the reason for the change here? Usually we want status line elements to always take up the same amount of space so they isn't any shifting when they change

gabydd (changes requested):

Connortsui20:
Something somewhat related to this is that the spacing between the statusline elements text is not really consistent, the default one is fine but with some mixing and matching the spaces become weird. I might file an issue for that if I get a chance.

helix-term/src/ui/statusline.rs
Connortsui20: Should the else case be 5 spaces then? If that is the case, does the readonly statusline element also need to take a consistent amount of space?

gabydd:
Yeah there is some discussion and implementation of consistent spacing in this pr https://github.com/helix-editor/helix/pull/7310 I would aim to keep things as consistent as possible but only within the scope of the read-only status indicator. In my opinion the message for having edited a read-only file is probably a little to long (though I gotten around testing how it looks yet because I am not at my computer).

helix-term/src/ui/statusline.rs
gabydd: If the padding here was just added to bring more consistentency my preference would be for doing it in another pr like the one I linked in my other comment

Connortsui20: Ig this links back to the comment I made, the spacing isn't really consistent at all among the different elements: some of them have spaces only after, some both before and after, some not at all. 
I agree that the file modification should be consistent, but I don't really want the ` [readonly] ` element to take up so much space when it isn't readonly: `          `

Connortsui20:
Personally I would prefer to have the file modification not take up any space as it leads to this: 
image

But for the scope of this PR this is probably good enough. Maybe after #7310 is done it would be nice to implement some sort of flag for each element that locks the width.

helix-term/src/ui/statusline.rs
gabydd: That's true it could probably shouldn't take up space for non read-only files cause that can't change while editing, but the two messages that can show when a file is read-only should take up the same or similar space

Connortsui20: What do you mean by the two messages? The second message is really a warning and I think it is fine for the warning to change the size

gabydd: Ah maybe I am mistaken could you share a screenshot of how the warning message looks?

Connortsui20: image

gabydd:
Whe could also switch where they are positioned in the default config and then keep the modified one exactly the same, not sure about how others feel about that

Connortsui20:
That could probably work actually since the warning message can only show up when ` [+] ` is also there

Connortsui20:
Yeah this is much nicer:
image
image

pascalkuthe (dismissed):
I like @gabydd suggestion. I think eventually we want to s streamline statusline spacing but for bow this seems reasonable. I don't really feel strongly on this as long as the default acts reasonably.

gabydd:
I think we can also remove the changes to the modified indicator and we will still have   ok spacing

Connortsui20:
Only reason I made the modified indicator changes was to be consistent with *most* of the other statusline elements, and also because I think it's better to have something like `file.txt main-branch  [+]  [readonly]` rather than `file.txt main-branch[+] [readonly]`, really I think this is an issue of cleaning up the ones that do not have the extra spaces

pascalkuthe:
> Only reason I made the modified indicator changes was to be consistent with _most_ of the other statusline elements, and also because I think it's better to have something like `file.txt main-branch [+] [readonly]` rather than `file.txt main-branch[+] [readonly]`, really I think this is an issue of cleaning up the ones that do not have the extra spaces

We prefer explicit spacer elements for cases lime this. I don't think you should touch existing spacing here. Any spacing changes to existing elements should be a seperate PR

gabydd (approved):

pascalkuthe (approved):

book/src/configuration.md
```diff
@@ -89,7 +89,7 @@ The `[editor.statusline]` key takes the following sub-keys:
 
 | Key           | Description | Default |
 | ---           | ---         | ---     |
-| `left`        | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name", "file-modification-indicator"]` |
+| `left`        | A list of elements aligned to the left of the statusline | `["mode", "spinner", "file-name", "file-modification-indicator", "read-only-indicator"]` |
```
the-mikedavis, line 92: The read-only and modification indicators should be swapped here to match the `Default for StatusLineConfig`

helix-term/src/commands/typed.rs
```diff
@@ -334,6 +334,11 @@ fn write_impl(
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
 
+    // `path.is_none()` allows :w <path> for other files
+    if path.is_none() && doc.detect_readonly() && !force {
+        bail!("file is readonly, use ! to overwrite")
```
the-mikedavis, line 339: `:w!` won't behave any different than `:w`: if the underlying file is read-only then any write will fail. I don't think we need any extra handling in the write functions for this. The errors when attempting to write to the file will already show up as errors on the statusline.

helix-term/src/commands/typed.rs
```diff
@@ -720,8 +739,19 @@ pub fn write_all_impl(
...
-        bail!("{:?}", errors);
+    if force {
+        return Ok(());
+    }
+
+    let mut error_msg = errors.join(" ");
+    if !readonly_errors.is_empty() {
+        // statusline text will have the form:
+        // [/home/user/readonly.txt is readonly, ..., /readonly2.txt is readonly] add ! to overwrite
+        error_msg = format!("{:?} use ! to overwrite", readonly_errors).replace('"', "");
```
the-mikedavis, line 750: Similar here: adding `!` won't make a difference for read-only files. I don't think we need extra handling for these errors either: the errors from the writes failing will already show up in the statusline. Improving the display of that when multiple writes fail could be nice but that could be a separate PR since writes can fail for other reasons than being read-only.

helix-term/src/ui/statusline.rs
```diff
@@ -442,6 +443,21 @@ where
...
 }
 
+fn render_read_only_indicator<F>(context: &mut RenderContext, write: F)
+where
+    F: Fn(&mut RenderContext, String, Option<Style>) + Copy,
+{
+    let (title, style) = match (context.doc.readonly, context.doc.is_modified()) {
+        (true, false) => (" [readonly] ", None),
+        (true, true) => (
+            " Warning: Changing a readonly file ",
```
the-mikedavis, line 453: I don't like the warning text here, I think it's too noisy to have help text like this in the part of the statusline with all the other widgets. I would prefer that this element only paid attention to `context.doc.readonly` and not `is_modified`

helix-view/src/document.rs
```diff
@@ -955,6 +958,35 @@ impl Document {
         }
     }
 
+    #[cfg(unix)]
+    // Detect if the file is readonly and change the readonly field if necessary (unix only)
+    pub fn detect_readonly(&mut self) -> bool {
```
the-mikedavis, line 963: These functions don't need to return `self.readonly`. The return value isn't used anywhere and the field is already public on the doc so callers can check there.

the-mikedavis: I believe this function also needs to be called in `Document::open` with the other auto-detection functions. That way the read-only indicator will be accurate if you `:open some-read-only-file.txt`, choose one from the picker, etc..

helix-view/src/editor.rs
```diff
@@ -456,6 +457,9 @@ pub enum StatusLineElement {
     // The file modification indicator
     FileModificationIndicator,
 
+    /// A read only file indicator
```
the-mikedavis, line 460:
```suggestion
    /// An indicator that shows `"[readonly]"` when a file cannot be written
```

the-mikedavis (changes requested):

Connortsui20:
I've seen the review, ill get to it by the end of this wekk

helix-view/src/document.rs
Connortsui20: I have `detect_readonly()` running every time `set_path()` is called, and `set_path()` is called inside `Document::open`, so I don't think we need to.

helix-term/src/commands/typed.rs
```diff
@@ -656,7 +656,7 @@ pub fn write_all_impl(
     force: bool,
     write_scratch: bool,
 ) -> anyhow::Result<()> {
-    let mut errors: Vec<&'static str> = Vec::new();
+    let mut errors: Vec<String> = Vec::new();
```
Connortsui20, line 659: I kept this as String since it will allow changes to the error messages easier and there is low overhead here

helix-term/src/commands/typed.rs
```diff
@@ -656,7 +656,7 @@ pub fn write_all_impl(
     force: bool,
     write_scratch: bool,
 ) -> anyhow::Result<()> {
```
the-mikedavis, line 658: The changes to write_all_impl should be reset: changing `&'static str` to `String` will cause a needless allocation and we shouldn't be discarding errors when the `force` flag is passed.

Connortsui20: > we shouldn't be discarding errors when the `force` flag is passed.

It was like this before: https://github.com/helix-editor/helix/blob/c7e9e94f007e04836d134cd3e53f71c38b7ba1ec/helix-term/src/commands/typed.rs#L723

Should I just revert it?

Connortsui20: im just going to revert it since in the end no logic was changed in this function

the-mikedavis (approved):

helix-term/src/commands/typed.rs
```diff
@@ -672,7 +672,7 @@ pub fn write_all_impl(
             }
             if doc.path().is_none() {
                 if write_scratch {
-                    errors.push("cannot write a buffer without a filename\n");
+                    errors.push("cannot write a buffer without a filename");
```
gabydd, line 675: it doesn't seem like this change is needed anymore?

gabydd: oh I see I guess this is fine

Connortsui20: I just reverted it back

pascalkuthe (approved):

.gitignore
goyalyashpal: > We always squash snakmer ORs kune this anyway

hmm??

kirawi:
> > Here is the code for vim windows access implementation [vim/vim@`4c0089d`/src/os_win32.c#L7665](https://github.com/vim/vim/blob/4c0089d696b8d1d5dc40568f25ea5738fa5bbffb/src/os_win32.c#L7665) It seems to boil down to `CreateFileW` + some tricks to handle directories + network shares
> 
> yeah I don't think its fair to require that for merging. If you feel interested/up to it @Connortsui20 then go for it but honestly just using `access` from rustix on UNIX and using the standard metadata.readonly field on Windows is enough for me (leave a todo note for the future with a link tough). The access syscall is a very standard part of POSIX and widely used/avialable (and easy to use). The windows variant seems to me like POSIX emulation and not exactly standard (and pretty complex to implement). It could be nice to have in the future but it doesn't seem to be standard practice on Windows (I am not even sure if this kind of low level thing should live in helix or a standalone crate).
> 
> > One other thing that I don't know how to approach is what the behavior should be when write-all is called. In vim it writes everything that can be written, but then shows a red error message 'readonly' option is set (add ! to overwrite).
> > The error message doesn't specify which file is readonly, only that there is one among the current buffers.
> 
> Hmm I guess the issue with that is that there could be many readonly files but there is only so much space in the statusline. It's probably fine to just print a list of butters here (that would be truncated if there is not enough space). Usually there are only few readonly files.
> 
> > It would be nice if it showed which specific file or files are readonly, as the above message is kind of vague. Also, while testing this in helix sometimes the error message would get covered by the "file_name" written ..., but I can't consistently reproduce that.
> 
> Yeah that's probably expected since writing happens asycn so that message could come trough later or earlier (non-deteministically). That's a problem in general, we should likely improve the way we display status messages. You don't need to worry about that here tough, that can be left to future work since il likely needs a systematic change

Neovim uses `libuv`'s [`uv_fs_access`](: https://github.com/neovim/neovim/blob/8df37423781493f58de060e1c9219cd1c3768130/src/nvim/os/fs.c#L933-L939

It uses `GetFileAttributesW` which is exposed in Rust via `std::os::windows::fs::MetadataExt::file_attributes()` which is consequently exposed by `Permissions::readonly()`. I think we could either copy vim or try using `AccessCheck`.

pascalkuthe:
sounds like we can just use access via rustix on UNIX and std everywhere else (on windows)

kirawi:
Yeah, but unfortunately it would not correctly report if the client has write privileges for that file.

PR #4252, opened 2022-10-13
Dedupe goto reference results

abseee:
We could dedupe more simply with just `locations.sort(); locations.dedupe()` but presumably language servers order results by relevance. Two lingering questions:  
1) How to test this
2) This belongs in helix-lsp rather than helix-term, but I couldn't figure a way to attach a callback to the Future directly

Fixes #3351
<details>
  <summary>Screenshots</summary>

Before: 
ksnip_20221013-053509
After: 
ksnip_20221013-092628
</details>

PR #7745, opened 2023-07-26
fix(picker): duplicate entries in pickers

### LINKED ISSUE #2286: Goto references returns duplicates
### Summary

Goto references (LSP) are currently returning duplicates.

I need to do some digging if this bug is originating from the LSP, but I think we should be able to remove duplicates from the the `FilePickers` options when created. 
Screen Shot 2022-04-26 at 8 24 55 AM
 

### Reproduction Steps

I found this when using goto ref on the `Modifier` bitmask in helix's repo 

### Helix log

Nothing relevant logged for LSP 

### Platform

MacOs

### Terminal Emulator

N/A

### Helix Version

22.05-dev (6de2e763)

woojiq:
Closes #2286

Added `PartialEq` subtrait for `Item` trait. Wrote a method in `Picker impl` that deduplicates picker options using only the `PartialEq` trait. That's why the complexity is O(n^2), but I don't think it's a problem since pickers probably won't have more than 10^4 elements.

<details><summary>Note: I didn't do a `--force` push, so the first two commits in this pr are also valid solutions though quite different, but I think it's better.</summary>

1. 93303a825946d4c01583e4d7ce659768afa80853 _Only fixes duplicates in lsp pickers_. Removes duplicates from a list of items if they have the same filename and range. So if you have a line: 
``` rust
let mut q = 1; q = 2; q = 3;
```
it will show 3 entries in `goto_reference` picker anyway because these references have different `Range`.

2. 37f7b6d13ee527598d7eb3bb1d7231a3ced42512 Add `as_partial_eq` method to the `Item` trait with the default implementation. So we can deduplicate inside the picker initialization. With this approach, we simulate optional `PartialEq` trait for `Item` so that anyone who wants can make items comparable. This is both a terrible and a beautiful solution.

The first solution is still valid because right now only lsp pickers can return duplicates (e.g `gr` in rust macros). Though If you satisfy with second approach I can add `as_partial_eq` implementation for other items (just in case).
</details>

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2008,7 +2008,7 @@ fn make_search_word_bounded(cx: &mut Context) {
 }
 
 fn global_search(cx: &mut Context) {
-    #[derive(Debug)]
+    #[derive(Debug, PartialEq)]
     struct FileResult {
         path: PathBuf,
         /// 0 indexed lines
@@ -2581,6 +2581,7 @@ fn file_picker_in_current_directory(cx: &mut Context) {
 fn buffer_picker(cx: &mut Context) {
     let current = view!(cx.editor).doc;
 
+    #[derive(PartialEq)]
     struct BufferMeta {
         id: DocumentId,
         path: Option<PathBuf>,
@@ -2648,6 +2649,7 @@ fn buffer_picker(cx: &mut Context) {
 }
 
 fn jumplist_picker(cx: &mut Context) {
+    #[derive(PartialEq)]
     struct JumpMeta {
         id: DocumentId,
         path: Option<PathBuf>,
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -99,6 +99,10 @@ impl ui::menu::Item for lsp::Location {
             .expect("Will only failed if allocating fail");
         res.into()
     }
+
+    fn as_partial_eq(&self) -> Option<&dyn PartialEq<Self>> {
+        Some(self as &dyn PartialEq<Self>)
+    }
 }
 
 struct SymbolInformationItem {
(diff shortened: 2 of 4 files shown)


pascalkuthe (changes requested):
>    /// Removes duplicates from picker options, complexity - `O(n^2)`.
>    /// Pickers usually have no more than 10^5 entries, so it doesn't matter.th

this is absolutely not true! quadratic time complexity like this is definitely a problem. For example a file picker started from the home directory. The only reason this is limited to 10k elements right now is that we manually limit picker results. But I will lift that restriction soon which means we will have a picker with millions of entries (streamed in which would require constant deduplication). 

Duplicate of:

*  #4252
* #4097

In general, our stance on this is: This is an upstream bug that should be solved in RA not in helix: https://github.com/helix-editor/helix/pull/4097#issuecomment-1267540506. If you want to see a fix for this you can submit a PR to RA. 

If we do want to deduplicate it should only be done for LSP requests and not in the Picker.  As the overhead for FS transversal and global search is simply unacceptable.

archseer:
https://github.com/helix-editor/helix/pull/4097#issuecomment-1267540506

pascalkuthe:
ah for RA in particular this was actually already fixed upstream: https://github.com/rust-lang/rust-analyzer/pull/13571

pascalkuthe:
Doing `gr` on the exact same struct does not how any duplicates for me. I am using nightly RA tough so its possible this hasn't landed on stable yet. Stable RA versions are weirdly old. In general it makes no sense to use stable RA since it doesn't follow the release cadence of the rest of the project. Every release on every Monday is just as stable/unstable as any other and stable just ships with the last Monday release that was released at the start of the nightly period (I think?)

woojiq:
Strange, even with nightly I have duplicates. Are you sure you press `gr` on `Modifier` inside `bitflags!`? How did you run? Me: 
```RUSTUP_TOOLCHAIN=nightly hx helix-view/src/graphics.rs```
`sh rust-analyzer --version` returns 1.73-nighty

pascalkuthe:
yeah I executed `gr` on the exact same line you did inside the bitflags. I have a built from source RA version installed under a different name because rustup keept replacing it with a stable version.

woojiq:
One last question (sorry for disturbing you) to see if the problem is on my end. Did you enter "356" in the picker? Because duplicates don't go in a row. If you have one entry anyway, I messed up rustup somehow :disappointed:

PR #7763, opened 2023-07-27
Refactor Nix flake to use crane

### LINKED ISSUE #7761: Cachix / Publish Flake (push) Failure, Unable to Build Flake
### Summary

After #7737, the "Cachix / Publish Flake (push)" action has failed with the following error which is propagated to others who try to build the latest `master`

```
error: builder for '/nix/store/klr0n42bzh6n8qhlf40ii7jca5vpnwnr-helix-term-deps-0.6.0.drv' failed with exit code 101;
       last 10 log lines:
       >   failed to update replaced source registry `crates-io`
       >
       > Caused by:
       >   failed to parse manifest at `/build/nix-vendor/tree-sitter-0.20.10/Cargo.toml`
       >
       > Caused by:
       >   error inheriting `rust_version` from workspace root manifest's `workspace.package.rust_version`
       >
       > Caused by:
       >   failed to find a workspace root
       For full logs, run 'nix log /nix/store/klr0n42bzh6n8qhlf40ii7jca5vpnwnr-helix-term-deps-0.6.0.drv'.
error: 1 dependencies of derivation '/nix/store/4pkv68m8bpijv8p97q5hw0x0yv9wf1vk-helix-term-0.6.0.drv' failed to build
error: 1 dependencies of derivation '/nix/store/3gv3cglh8458d8ma4yd647mxpnhpdqsn-helix-term-0.6.0.drv' failed to build
Error: Process completed with exit code 1.
```

### Reproduction Steps

 

I tried this:

Update flake input, attempt to build

I expected this to happen:

Successful build

### Helix log

N/A

### Platform

Linux

### Terminal Emulator

Kitty

### Helix Version

8a28f3059398eec7b5496bb1088b4d9870e6eb7e

the-mikedavis:
This resolves a build issue with nci/dream2nix and git dependencies. We can keep most of the helix-specific parts of the flake, we just need to switch from configuring nci to calling craneLib functions.

We also switch to flake-utils from flake-parts:

* Using rust-overlay with flake-parts directly is unergonomic (see https://github.com/hercules-ci/flake-parts/discussions/83).
* Removing flake-parts reduces the overall dependencies: rust-overlay already depends on flake-utils.

Closes #7761

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,166 +3,190 @@
 
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+    flake-utils.url = "github:numtide/flake-utils";
     rust-overlay = {
       url = "github:oxalica/rust-overlay";
-      inputs.nixpkgs.follows = "nixpkgs";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
     };
-    nci = {
-      url = "github:yusdacra/nix-cargo-integration";
-      inputs.nixpkgs.follows = "nixpkgs";
+    crane = {
+      url = "github:ipetkov/crane";
       inputs.rust-overlay.follows = "rust-overlay";
+      inputs.flake-utils.follows = "flake-utils";
+      inputs.nixpkgs.follows = "nixpkgs";
     };
-    parts.url = "github:hercules-ci/flake-parts";
   };
 
-  outputs = inp: let
-    mkRootPath = rel:
-      builtins.path {
-        path = "${toString ./.}/${rel}";
-        name = rel;
+  outputs = {
+    self,
+    nixpkgs,
+    crane,
+    flake-utils,
+    rust-overlay,
+    ...
+  }:
+    flake-utils.lib.eachDefaultSystem (system: let
+      pkgs = import nixpkgs {
+        inherit system;
+        overlays = [(import rust-overlay)];
       };
-    filteredSource = let
-      pathsToIgnore = [
-        ".envrc"
-        ".ignore"
-        ".github"
-        ".gitignore"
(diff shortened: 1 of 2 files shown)


pascalkuthe (approved):
I can't really review this since I don't use nix anymore but from what I remember from my days this seems reasonable :+1:  Thanks!

flake.nix
```diff
@@ -3,166 +3,188 @@
 
   inputs = {
     nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
-    rust-overlay = {
-      url = "github:oxalica/rust-overlay";
+    crane = {
+      url = "github:ipetkov/crane";
       inputs.nixpkgs.follows = "nixpkgs";
```
archseer, line 8: Should we also add `inputs.rust-overlay.follows = "rust-overlay";`?

archseer: Same for `flake-utils`, it's defined twice in `flake.lock` otherwise

flake.nix
```diff
@@ -3,166 +3,190 @@
...
-      inputs.nixpkgs.follows = "nixpkgs";
+      inputs = {
+        nixpkgs.follows = "nixpkgs";
+        flake-utils.follows = "flake-utils";
+      };
     };
-    nci = {
-      url = "github:yusdacra/nix-cargo-integration";
-      inputs.nixpkgs.follows = "nixpkgs";
+    crane = {
```
the-mikedavis, line 14: We could also remove crane and use only rust-overlay like [nil does](https://github.com/oxalica/nil/blob/main/flake.nix). Crane has some nice helpers for the `checks` and it seems to be very popular and stable though so I think it's fine to depend on it for now. If it becomes a problem in the future though, pure `rust-overlay` is an option

Issue #3001, opened 2022-07-07
Remove yank behavior of `c` and `d`


samholmes:
I think Vim made a mistake by letting `c` and `d` pollute the default register by "yanking" the contents of the change or deletion. I'll make a short case as to why I think Helix should make a bold decision to not do this like vim (and maybe even kak).

# Why not?

The reason why `d` and `c` should not also yank is simple, because you can yank before `d` and `c` easily with `y`. You cannot as easily "prevent yank" without knowing about registers and particularly the black-hole register `"_`; so you're left with `"_d` or `"_c` which is too often a use-case to make it not the default behavior. The times when you want to yank before a modification is easily opt-in with `yd` and `yc`. Simply put, the answer why the default behavior should be inverted: it's the correct UX.

Note, in View mode, `y` does not remove the selection like it does in vim. This positions Helix to be able to even consider the correct behavior.

With the correct default behavior in place, users can edit happily knowing that when they copy something to their default register many modifications back in their history, they will not accidentally overwrite the default register with `d` and `c`. The foot-gun has been holstered.

# How do we migrate?

Some users (vim and kak alike) will not expect this change and will perhaps not like to modify their habit. However, this could easily be mitigated with a config to change the default behavior back (which is what user's do already). If a user doesn't like the position Helix makes, they can modify the behavior or change their habits.

Please.

the-mikedavis:
This is currently implemented in the `delete_selection_noyank` (A-d) and `change_selection_noyank` (A-c). You can switch the keybinds in your `config.toml`:

```toml
# ~/.config/helix/config.toml
[keys.normal]
A-d = "delete_selection"
d = "delete_selection_noyank"
A-c = "change_selection"
c = "change_selection_noyank"
```

I'm not sure it makes sense for this to be the default keymap though - `yd` or `yc` is the same number of keystrokes as `A-d` or `A-c` so if anything it would make sense to remove the `A-d` and `A-c` keybinds. Without anything to replace those keybinds I'm not sure it makes much sense to reclaim them. Plus in my experience, if you're running around a code-base selecting lines/blocks/etc and deleting them with `d` and then placing something new with `y`, you might be better off using `R` which doesn't affect the yank register.

On the other hand I agree with your reasoning:

> you can yank before `d` and `c` easily with `y`.

samholmes:
@the-mikedavis `A-d` being Alt+d or Option+d on macOS, does not appear to delete without yank.

[...]

PR #7768, opened 2023-07-28
Make pasting replace the entire selection

Resonious:
Hey folks.

I don't think anyone asked for this, but it's an idea that would improve my productivity at least.

In vim, you can make a selection and then "p" to replace the selection with whatever you have yanked. I used to do this a ton in vim. It's one way to work around https://github.com/helix-editor/helix/issues/3001.

This PR makes non-linewise selections _replaced_ when pasting onto them.

[Kooha-2023-07-28-16-29-38.webm](https://github.com/helix-editor/helix/assets/2793160/e6bdd784-7d85-482c-90d1-32d9a5cfeba3)

This is a breaking change so I don't really expect a merge as-is. Just wanted to bring up the idea and do some legwork as well. One alternative could be to implement this as a separate command and bind it to, say, `a-p`.

Anyway let me know what you think. If you have time.

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -113,7 +113,7 @@ impl Range {
     }
 
     /// `Direction::Backward` when head < anchor.
-    /// `Direction::Backward` otherwise.
+    /// `Direction::Forward` otherwise.
     #[inline]
     #[must_use]
     pub fn direction(&self) -> Direction {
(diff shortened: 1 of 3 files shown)


gabydd:
there are already separate commands for this: ``R`` and ``<space>R``

pascalkuthe (changes requested):
We already have `R` for this and I don't see us changing the `p` command

Resonious:
Oh wow how embarrassing 😂. Carry on then!

PR #7772, opened 2023-07-28
feat: display invisible chars inside selection

### LINKED ISSUE #1068: Configurable rendering of invisible characters like space, tab, and line feed
Some config option for whether to display invisible characters like rendering spaces as dots, tabs as arrows, etc.

The config option would have 3 possible values:
* `none`: Don't render invisible characters.
* `selection`: Render invisible characters that fall into a selection.
* `all`: Render all invisible characters.

I presume this would also add a field to themes?

woojiq:
Closes #1068
It appeared that #2208 is stale and would not be rebased on top of #5420. Though, if the author wakes up, you can close this PR without hesitation.

In general now it works fine: https://asciinema.org/a/vVQptp5dofjy3nz7CckwxjVFg

But I found one ~~bug~~ feature: I can't `set whitespace.render.<smth> <value>` only `set whitespace.render <value>`. This is related to master helix too. I don't know how to fix this at the moment (probably something with serde), but I would like to add this fix to the PR also

Code isn't the cleanest in the world, and I need to remove a lot of `clone` calls, maybe replace `String` with generic.

### DIFF
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -43,9 +43,25 @@ struct StyleIter<'a, H: Iterator<Item = HighlightEvent>> {
     theme: &'a Theme,
 }
 
+struct StyleIterItem {
+    style: Style,
+    end: usize,
+    is_selection: bool,
+}
+
+impl Default for StyleIterItem {
+    fn default() -> Self {
+        Self {
+            style: Style::default(),
+            end: usize::MAX,
+            is_selection: false,
+        }
+    }
+}
+
 impl<H: Iterator<Item = HighlightEvent>> Iterator for StyleIter<'_, H> {
-    type Item = (Style, usize);
-    fn next(&mut self) -> Option<(Style, usize)> {
+    type Item = StyleIterItem;
+    fn next(&mut self) -> Option<Self::Item> {
         while let Some(event) = self.highlight_iter.next() {
             match event {
                 HighlightEvent::HighlightStart(highlights) => {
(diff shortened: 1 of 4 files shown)


helix-term/src/ui/document.rs
```diff
@@ -58,13 +74,22 @@ impl<H: Iterator<Item = HighlightEvent>> Iterator for StyleIter<'_, H> {
                     if start == end {
                         continue;
                     }
+                    let mut is_selection = false;
                     let style = self
                         .active_highlights
                         .iter()
                         .fold(self.text_style, |acc, span| {
+                            // find scoped used in the selection to later determine if text is in a selection
+                            if self.theme.selection_scopes().binary_search(&span.0).is_ok() {
```
pascalkuthe, line 83: This is pretty hacky and would be prone tincause problems if we reuse theme scopes somewhere.  This also basically hardcoded yet another function in the rendering code. One of my big goals with #5420 was to avoid exactly that problem. 


Yo that end the `TextDecoration` API exists. It's currently not quite fleshed out enough but once #6417 lands the API will be expanded/refactored so you can draw decorations as specific char anchors. Virtual text is special cased here because this is genuinely the only place we know whether a grapheme is virtual text (since it's nitnpartnof the normakd document) but selections are a high keven concept that the rendering shouldn't be aware of

woojiq: So should I wait merge of your PRs or overhaul mine? Tbh, if it's better to wait, I guess I'll close it for now, because I'm not sure if I'll have time in six months or a year to maintain it. Hope you are not tired of my attempts :smiley:

pascalkuthe: no worries. Unfortunately, I do think it would be better to wait here. I hope you understand. In this case if we were to merge your PR as is I would have to undo the entire change in the future (and then reimplement it with the new API). Thanks for contributing tough!

woojiq: np, I understand

PR #7784, opened 2023-07-30
Bump the version of Hare's grammar

atomcult:
Exactly what it says on the tin. 

Syntax highlighting has been moderately broken since Hare's introduction of the new `yield` keyword.
The new version of Hare's grammar fixes this, and updates the `match`/`switch` definitions.


the-mikedavis (commented):
The new yield node should be added to the highlight queries as a keyword

the-mikedavis (approved):

pascalkuthe (approved):

PR #7788, opened 2023-07-30
add basic modeline support

### LINKED ISSUE #729: Vim modeline parsing

In addition to EditorConfig #279 and auto-detection #245 (amazing that it's already done!), an extra source of indentation settings is modelines in the file.

e.g. Kakoune includes [a script by default](https://github.com/mawww/kakoune/blob/master/rc/detection/modeline.kak) that parses vi(m) style modelines using sed and shell.

doy:
fixes #729 

currently only supports setting language, indent style, and line endings, but maybe this is enough?


helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+use helix_core::regex::Regex;
+use helix_core::{LineEnding, Rope};
+use once_cell::sync::Lazy;
+
+// 5 is the vim default
+const LINES_TO_CHECK: usize = 5;
+const LENGTH_TO_CHECK: usize = 256;
+
+static MODELINE_REGEX: Lazy<Regex> =
+    Lazy::new(|| Regex::new(r"(?:\S+\s+)?(?:vi|vim|Vim|ex):\s*(?:set? )?(.*)(?::[^:]*)?").unwrap());
```
pascalkuthe, line 13: The regec should be anchored (that makes it both faster and correct since the modeline is specific about how it may be prefixed We also don't need capture groups here. Since modeline support escapes (with `\:` we can't phrase the option with regec anyway (it's technically possible but not worth the effort. You can simply find the regec match and then use the rest of the line (from the start if the regec). That also removes the need for non-capturing groups (since you will not compute captures anywya which sinplfies the regec a bit further

```suggestion
    Lazy::new(|| Regex::new(r"^(\S*\s+)?(vi|[vV]im[<=>]?\d*|ex):\s*(set?\s+)?").unwrap());
```

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+    pub fn indent_style(&self) -> Option<IndentStyle> {
+        self.indent_style
+    }
+
+    pub fn line_ending(&self) -> Option<LineEnding> {
+        self.line_ending
+    }
+
+    fn parse_line(line: &str) -> Option<Self> {
+        if let Some(captures) = MODELINE_REGEX.captures(line) {
```
pascalkuthe, line 57: As described above simply use find instead of captures here

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+            let mut language = None;
+            let mut indent_style = None;
+            let mut line_ending = None;
+
+            let options = &captures[1];
```
pascalkuthe, line 62: And then do `line[march.end..]`

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+            for option in options.split([' ', '\t', ':']) {
```
pascalkuthe, line 63: Using a simple split here doesn't work since mode lines allow escaping `with \:` you need to handle that here. You can take a look at the `query_atom` function to see how you can use split to split on a char that can be escaped (in that case it was spaces)

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+            return Some(Self {
+                language,
+                indent_style,
+                line_ending,
+            });
+        }
+        None
+    }
+
+    fn merge(&mut self, other: Self) {
```
pascalkuthe, line 99: I think this is unnecessary. Instead of Creating a new `ModeLine` for each line that is parsed simply make the parsing accept a mutable reference to self and write the results directly into self

pascalkuthe (commented):
This looks pretty good, thanks for taking this on!

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+        for line in text.lines().take(LINES_TO_CHECK).chain(
+            text.lines_at(text.len_lines())
+                .reversed()
+                .take(LINES_TO_CHECK),
+        ) {
+            // can't guarantee no extra copies, since we need to regex and
+            // regexes can't operate over chunks yet, but we can at least
+            // limit how much we potentially need to copy because modelines
+            // are typically quite short.
+            let line = Cow::<str>::from(line.slice(..(LENGTH_TO_CHECK.min(line.len_chars()))));
```
pascalkuthe, line 35: I think we can just ignore lines longer than `LENGTH_TOK_CJECK` since otherwise we cut off mode lines in the middle and get a wrong result. I don't think this occurs in practice anyway

pascalkuthe:
Also one more thing: I think the `modeline.rs` file can/should be moved to `helix-core`

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+
+#[derive(Default, Debug, Eq, PartialEq)]
+pub struct Modeline {
+    language: Option<String>,
+    indent_style: Option<IndentStyle>,
+    line_ending: Option<LineEnding>,
+}
+
+impl Modeline {
+    pub fn parse(text: &Rope) -> Self {
```
the-mikedavis, line 23:
```suggestion
    pub fn parse(text: RopeSlice) -> Self {
```

helix-view/src/modeline.rs
```diff
@@ -0,0 +1,258 @@
...
+        let mut modeline = Self::default();
+
+        for line in text.lines().take(LINES_TO_CHECK).chain(
+            text.lines_at(text.len_lines())
+                .reversed()
+                .take(LINES_TO_CHECK),
```
the-mikedavis, line 29: Should this iterator be reversed again after the `take`? The lines at the end of the file would have the opposite merging behavior as the lines at the beginning otherwise

the-mikedavis (commented):
To me this feels like plugin territory rather than something that should be built into the core. I believe only each editor supports their own modeline natively: vim variants are the only ones to support vim modelines, only emacs supports emacs modelines natively, vscode doesn't support any modelines natively AFAIK. The translation between filetypes and line ending values from Vi to Helix feels a little hacky to me here: we're basically parsing and trying to intepret Vi `set` commands. If we do support this in core, I'm not sure where we draw the lines for which Vi settings should be supported. And do we support interpreting emacs modelines as well?

pascalkuthe:
vim modelines are pretty standard even in editors other than vim. Kakoune also parses these by default (https://github.com/mawww/kakoune/blob/master/rc/detection/modeline.kak). I think it would be reasonable to mirror kakoune here. It supports the following options currently:

* `scrolloff`
* `sidescrolloff` (do we even have a sepreat setting for this?
* `tabstop`
*  `shiftwidth`
*  `textwidth`
*   `fileformat`
*  `filetype`
*  `spelllang`
* `bom`
* `nobom`

helix-view/src/modeline.rs
doy: fascinatingly enough no, the vim implementation also reads from the end of the file up

doy:
yeah, i don't think this belongs in a plugin. i'm not tied to vim syntax specifically (we could also totally just do our own thing) but it feels pretty important to be able to have some way to override settings like file type that heuristics aren't always going to correctly identify (i have quite a few files that aren't identifiable by filename or shebang).

doy:
this should resolve the issues mentioned so far (other than the extended list of options that kakoune supports - should i implement those now or can it wait?). let me know if there are any other issues!

helix-core/src/modeline.rs
```diff
@@ -0,0 +1,247 @@
...
+        "dos" => Some(LineEnding::Crlf),
+        "unix" => Some(LineEnding::LF),
+        #[cfg(feature = "unicode-lines")]
+        "mac" => Some(LineEnding::CR),
+        _ => None,
+    }
+}
+
+// if vim and helix use different names for a language, put that mapping here
+fn vim_ft_to_helix_language(val: &str) -> String {
```
pascalkuthe, line 110: @the-mikedavis was concerned about too much vim emulation here. I tend to agree, kakoune doesn't do this either. We don't need to translate file types

doy: fair enough - i do kind of wonder what the right answer is for files that are intended to be shared between people using vim and helix though. maybe we do want our own format in addition?

doy:
i added a second commit which adds support for parsing a helix-specific modeline, since that is probably going to be more useful going forward. i think parsing some amount of vim modelines is useful because they are so widespread, but for configuration that doesn't exist in vim (or is different in vim, such as language names), we are going to want a different method for configuring that.

this currently uses the same toml configuration syntax we use for the configuration file, but not sure if that's what we actually want (given that my understanding is that the toml configuration file is eventually intended to go away in favor of some kind of scripted configuration - telling people to rewrite their local configuration file is different from telling them to fix modelines in every file they want to use). let me know if you think this is fine or if we can come up with something better here.

helix-core/src/modeline.rs
```diff
@@ -0,0 +1,292 @@
...
+                                self.indent_style = Some(IndentStyle::Spaces(val));
+                            }
+                        }
+                    }
+                    "ff" | "fileformat" => {
+                        if let Some(val) = parts.get(1) {
+                            self.line_ending = vim_ff_to_helix_line_ending(val);
+                        }
+                    }
+                    "noet" | "noexpandtab" => {
```
pascalkuthe, line 92: I wonder if we should habdle `expandtab` without an explicit shiftwidth. I think that isn't too rare. We could simply set the indentation to the tabwidth currently set for that document.  This would match `sw=0` (which we may also want to handle) and is generally what a user would expect.

One minor csbeat in both cases: If the document ks already configured to use spaces then expandtab should do nothing. That will need special representation in the ModeLine struct but IMO it's worth it since that is pretty common.

helix-core/src/modeline.rs
```diff
@@ -0,0 +1,292 @@
...
+                        self.indent_style = Some(IndentStyle::Tabs);
+                    }
+                    _ => {}
+                }
+            }
+        }
+
+        if let Some(pos) = HELIX_MODELINE_REGEX.find(line) {
+            let config = &line[pos.end()..];
+            match toml::from_str::<ModelineConfig>(config) {
```
pascalkuthe, line 102: I dont think tonl is they syntax I would go for here. We are removing toml hopefzlly soon and generally it makes more sense to me what vim does and emulate typable commands (everything you allow users to set here has an interactive typable command that allows setting that config per document interactively). The parsing should be pretty simple. So for example it would be `// helix: line-ending=lf indent=4 lang=rust`

In terms of aliases you should also just support the aliases these commands support.

helix-view/src/document.rs
```diff
@@ -668,6 +672,7 @@ impl Document {
             focused_at: std::time::Instant::now(),
             readonly: false,
             jump_labels: HashMap::new(),
+            modeline,
```
pascalkuthe, line 675: I am not a fan of adding a field for this. Mofeline detection should only run when reading a file for the first time and set the buffer setting. In the future with the new config system there will be a dedicated buffer config. For now all fields that you allow to overwrite already have fields on document and you should only set these fields automatically after parsing the ModeLine/bypass the existing autodetection

helix-core/src/modeline.rs
```diff
@@ -0,0 +1,292 @@
...
+            ),
+            (
+                "// vim: noai:ts=4:sw=4",
+                Modeline {
+                    indent_style: Some(IndentStyle::Spaces(4)),
+                    ..Default::default()
+                },
+            ),
+            (
+                "/* vim: set noai ts=4 sw=4: */",
```
lucasew, line 175: Funny

This `:` at the end of the directive is not required in Vim

lucasew:
There is some exoteric examples that I tested if Vim can properly detect the file type. Vim got it right in all but the FSharp example, that is expected because Vim doesn't have by default the file type definition for FSharp. In my tests the only thing that seems to break detection here is that requirement of a : at the end of the directive.

https://github.com/lucasew/nix-flake-shell/tree/c896400199f12a293ec6ee7d54b5f2e19f0f0a99/tests

markstos:
> vim modelines are pretty standard even in editors other than vim.

I don't know common it is parse vim's modelines. Gedit can parse expandtab, tabstop, shiftwidth, wrap and textwidth: https://help.gnome.org/users/gedit/stable/gedit-plugins-modelines.html.en

The Kate editor supports their own Kate-specific modelines:
https://kate-editor.org/2006/02/09/kate-modelines/#:~:text=Kate%20Part's%20modelines%20%E2%80%93%20also%20called,to%20Emacs%20and%20vim%20modelines.

But I imagine a lot of Helix's users will come from vim/neovim, which parse them. 

I agree parsing *vim's* modelines would be a good fit for a plugin. Emacs can parse Vim's modelines using a plugin:
https://www.emacswiki.org/emacs/vim-modeline

Regarding the trailing colon. I'm one of the vim users who waw aware that's not required, so I don't put in in my own files. I'm not sure how common that is.

doy:
okay, i think this should address all of the remaining comments - let me know what you think!

eikenb:
Looks like this could use a rebase.

doy:
i'll probably wait for further review before doing any more rebasing.

the-mikedavis:
I chatted with @pascalkuthe about this a while ago and I think this could fit well into core. Even GitHub has some basic modeline parsing for language detection https://github.com/github-linguist/linguist/blob/main/lib/linguist/strategy/modeline.rb

I'm hesitant about adding a new Helix-specific modeline style though. Was this added with a specific use-case in mind? Can Vim/Emacs style modelines work instead?

markstos:
I see adopting Vim's modeline as a feature. Consider a team with both Vim and Helix users. Maintaining mode lines for two different editors would be a pain.

Also, there probably a lot of users who switching from Vim/Neovim to Helix, so it's also a feature to keep all those old modelines working without modification.

Only issue would be if Helix wanted to add a new file type to be supported before vim/neovim did, so there would need be some coordination to not unintentionally fork compatibility in the future.

chtenb:
I would like to add this to my fork to test this, but it seems to have a bunch of merge conflicts

doy:

PR #7791, opened 2023-07-31
Initial Ada language support

tomekw:
Tree-Sitter grammar and queries from: https://github.com/briot/tree-sitter-ada

Language server from: https://github.com/AdaCore/ada_language_server

Ada build tool: https://alire.ada.dev

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -3,6 +3,7 @@
 
 [language-server]
 
+ada-language-server = { command = "ada_language_server" }
 awk-language-server = { command = "awk-language-server" }
 bash-language-server = { command = "bash-language-server", args = ["start"] }
 bass = { command = "bass", args = ["--lsp"] }
@@ -2099,7 +2100,7 @@ file-types = ["adb", "ads"]
 roots = ["alire.toml"]
 comment-token = "--"
 indent = { tab-width = 3, unit = " " }
-language-server = { command = "ada_language_server" }
+language-servers = [ "ada-language-server" ]
 
 [[grammar]]
 name = "ada"
--- a/languages.toml
+++ b/languages.toml
@@ -2092,6 +2092,20 @@ name = "jsonnet"
 source = { git = "https://github.com/sourcegraph/tree-sitter-jsonnet", rev = "0475a5017ad7dc84845d1d33187f2321abcb261d" }
 
 [[language]]
+name = "ada"
+scope = "source.ada"
+injection-regex = "ada"
+file-types = ["adb", "ads"]
+roots = ["alire.toml"]
+comment-token = "--"
+indent = { tab-width = 3, unit = " " }
+language-server = { command = "ada_language_server" }
+
+[[grammar]]
+name = "ada"
+source = { git = "https://github.com/briot/tree-sitter-ada", rev = "ba7951a8f3fb08f9ea923625153e7670c89f30b4" }
+
+[[language]]
 name = "astro"
 scope = "source.astro"
 injection-regex = "astro"
--- a/runtime/queries/ada/highlights.scm
+++ b/runtime/queries/ada/highlights.scm
@@ -1,0 +1,180 @@
+;; highlight queries.
+;; See the syntax at https://tree-sitter.github.io/tree-sitter/using-parsers#pattern-matching-with-queries
+;; See also https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md#parser-configurations
+;; for a list of recommended @ tags, though not all of them have matching
+;; highlights in neovim.
+
+[
+   "abort"
+   "abs"
+   "abstract"
+   "accept"
+   "access"
+   "all"
+   "array"
+   "at"
+   "begin"
+   "declare"
+   "delay"
+   "delta"
+   "digits"
+   "do"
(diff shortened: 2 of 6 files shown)


languages.toml
```diff
@@ -2091,6 +2091,20 @@ language-servers = [ "jsonnet-language-server" ]
 name = "jsonnet"
 source = { git = "https://github.com/sourcegraph/tree-sitter-jsonnet", rev = "0475a5017ad7dc84845d1d33187f2321abcb261d" }
 
+[[language]]
+name = "ada"
+scope = "source.ada"
+injection-regex = "ada"
+file-types = ["adb", "ads"]
+roots = ["alire.toml"]
```
tomekw, line 2099: Is it possible to add wildcards here? Ada supports `*.gpr` files in a form of `PROJECT_NAME.gpr` (`gpr` = GNAT Project).

tomekw:
:point_up: @briot ty for the grammar! Any ideas how to add the indent queries as well?

archseer:
What's the compile time on this? We had to drop the old tree-sitter-ada grammar that was under the official tree-sitter org because it took too long to compile and had various issues.

tomekw:
@archseer I’ve no idea TBH. How can I check this?

briot:
I just did a quick check:

 > cargo clean
❯ time cargo build
    Compiling memchr v2.5.0
    Compiling cc v1.0.73
    Compiling regex-syntax v0.6.27
    Compiling tree-sitter v0.20.9
    Compiling tree-sitter-ada v0.0.1 (/home/briot/tree-sitter-ada)
    Compiling aho-corasick v0.7.19
    Compiling regex v1.6.0
     Finished dev [unoptimized + debuginfo] target(s) in 3.14s
cargo build  10.27s user 1.09s system 357% cpu 3.179 total

So 3s total, when all packages have been downloaded already (and these 
are dependencies
from tree-sitter itself, so would apply to all languages that you want 
to support).


I have not investigated the indentation queries (not sure that neovim, 
which is my editor
of choice) supports them properly yet.

Emmanuel

languages.toml
the-mikedavis: Not currently, but this would be useful for other languages (dotnet `PROJECT_NAME.sln` comes to mind)

runtime/queries/ada/highlights.scm
the-mikedavis, line 1: The captures here should be adjusted to the ones that Helix uses: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example `@conditional` should become `@keyword.control.conditional`, `@number` should become `@constant.numeric`, etc.

runtime/queries/ada/highlights.scm
```diff
@@ -0,0 +1,180 @@
...
+;    [
+;       (procedure_specification)
+;       (function_specification)
+;    ] @function.spec
+;)
+
+;; Highlight errors in red. This is not very useful in practice, as text will
+;; be highlighted as user types, and the error could be elsewhere in the code.
+;; This also requires defining    :hi @error guifg=Red    for instance.
+(ERROR) @error
```
the-mikedavis, line 179:
```suggestion
```

Eventually we want to add a language config option to show syntax errors but this is usually just noisy by default so we disable it (remove the `(ERROR) @error` pattern)

runtime/queries/ada/locals.scm
the-mikedavis, line 1: The captures here should follow tree-sitter's locals capture names: `@local.scope`, `@local.reference` and `@local.definition` https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

runtime/queries/ada/textobjects.scm
the-mikedavis, line 1: See the textobjects docs, Helix has a different set of textobject captures than nvim: https://docs.helix-editor.com/master/guides/textobject.html

the-mikedavis (commented):
The links to nvim documentation should also be removed, that could be confusing for anyone reading the files

the-mikedavis:
Looking at the state counts in the parser.c, this grammar seems larger than average but not as huge as something like tree-sitter-elixir. On-disk I see the so as 327kb and the compilation time of `hx -g build` of just the ada grammar seems very normal to me.

languages.toml
```diff
@@ -2091,6 +2091,20 @@ language-servers = [ "jsonnet-language-server" ]
...
+comment-token = "--"
+indent = { tab-width = 3, unit = " " }
+language-server = { command = "ada_language_server" }
```
the-mikedavis, line 2102: The format for declaring language servers changed after the 23.05 release. An `ada-language-server` key should be added under the `[language-server]` table at the top:

```toml
[language-server]
# ...
ada-language-server = { command = "ada_language_server" }
# ...

[[language]]
name = "ada"
language-servers = [ "ada-language-server" ]
# ...
```

the-mikedavis (commented):
Running `cargo xtask docgen` and committing the changes will fix the docs CI

KhazAkar:
Would this PR also take into consideration .gpr file support with ada language-server running with `--language-gpr` flag?

pascalkuthe:
closed by https://github.com/helix-editor/helix/pull/9908

PR #7796, opened 2023-08-01
languages: add protobuf language servers

jzelinskie:


gabydd (changes requested):
Please run `cargo xtask docgen` to add the changes to the language support docs

gabydd:
The language server here https://github.com/pbkit/pbkit/ apparently also supports completion haven't tested it out though so the other language server might be better

jzelinskie:
I added pbkit -- neither projects are extremely mature, but i'll take limited functionality over none!
bufls is an experiment, but backed by a company that sells protobuf tool.  If I had to place my bet somewhere, it'd be there.

the-mikedavis (approved):

pascalkuthe (approved):

softkot:
https://github.com/hudbrog/protobuf-lsp is more usable 

```
go install github.com/hudbrog/protobuf-lsp@latest && ln -s protobuf-lsp bufls 
```

gives better UX

PR #7800, opened 2023-08-01
update yanked dependency

Zoybean:
I noticed that [hermit-abi 0.3.1](https://crates.io/crates/hermit-abi/0.3.1) is yanked, so this just updates that to the next version in Cargo.lock


pascalkuthe (approved):
realistically we don't support hermit anyway so I wish there was just a way to not have it in the the dependency graph at all but there is not much we can do

PR #7805, opened 2023-08-02
support diff gutters for non-vcs files

### LINKED ISSUE #6571: Show gutter diff for files outside a git repo
Sublime Text has this nice feature showing all the git diffs in the gutter even for files that aren't in a git repo (it shows changes from the last time you've opened a file).

Would be nice if Helix's recent feature of git diffs could be extended to also work for standalone files

doy:
fixes #6571 

i'm not sure if the file size limitation is necessary (it doesn't exist for git diffs) but i added it because files in git repositories tend to be relatively small (git in general doesn't handle large files particularly well), and using a text editor to open enormous log files is a pretty common operation.

### DIFF
--- a/helix-vcs/src/file.rs
+++ b/helix-vcs/src/file.rs
@@ -1,0 +1,39 @@
+use std::{io::Read, sync::Arc};
+
+use anyhow::bail;
+use arc_swap::ArcSwap;
+
+use crate::DiffProvider;
+
+pub struct File;
+
+const MAX_SIZE: usize = 10 * 1024 * 1024;
+const READ_SIZE: usize = 8 * 1024;
+
+impl DiffProvider for File {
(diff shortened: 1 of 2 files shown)


pascalkuthe:
I am personally not much of a fan of this. I save frequently so this would inky be confusing. I could see some use in enabling this manually (for example when there is a read conflict). I want to add an option to configure diff sources in the future (both o'er buffer and globally) that could be used for this.

I don't think it should exist by default tough (and I actually want to get rid of the cascading fallback for diff providers). The implementation here is also pretty inefficient/not quite correct because it reads the file from disk. Considering we currently only reload the diff provider when opening a file thus should simply reuse the file when opened. This also doesn't handle encoding an line endings properly (something that git handles for UA with git).

doy:
ah, yeah, non-cascading diff sources that can be dynamically changed per buffer is actually the thing i want here. let me see if i can come up with something more in that direction.

PR #7819, opened 2023-08-03
Improve wren support

Jummit:
This uses the improved grammar and adds queries for indenting, injections and locals, plus overall improvements.

Maybe this isn't the right place to ask this, but is there a reason why only parameters are highlighted a different color using the locals queries?

Example:

```javascript
function test(a) {
  a // highlighted
  const b = 1
  b // not highlighted
}
```

I guess it's because it's less confusing?

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -167,7 +167,7 @@
 | webc | ✓ |  |  |  |
 | wgsl | ✓ |  |  | `wgsl_analyzer` |
 | wit | ✓ |  | ✓ |  |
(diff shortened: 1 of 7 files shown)


the-mikedavis:
Locals queries allow the highlighting to track between definitions and references. `a` is `@variable.parameter` because when `a` is defined it's tagged as `@variable.parameter`. `b` is defined just as `@variable` on line 3 so all usages of `b` will also be `@variable`. Both are affected by the locals query but `a` is parameter.

the-mikedavis (approved):

pascalkuthe (approved):

Issue #7154, opened 2023-05-27
Editor crashes when gutters.line-numbers.min-width is too large


gyreas:
### Problem
When I set the gutters' line width using `:set gutters.line-numbers.min-width 477`, the editor crashes (or more correctly, Rust panics) spitting out `index out of bounds` error.

### Reproduce
I'm guessing the array's, or whatever is used, size is dependent upon the host system, so just try some absurd number, say 123456.

### Outcome
Same panic message as above

### Potential fix
- the input can be truncated to less than the maximum allowed size should it be beyond that threshold
- then, the user is alerted to such illegal use (and possibly given a knock on the head)*

### Version
hx 23.05 (multi-lsp incl.)

### Host OS
- Fedora 37
- I don't remember the kernel version but it's >6

kirawi:
https://github.com/helix-editor/helix/blob/3a8592abdb9f30c7ebb74f10fa5127e754d18bc6/helix-view/src/gutter.rs#L216

This needs to be bounded to the width of the terminal (`view.rect.width`).

gyreas:
How do I invoke a red message at cmdline for complaints to the user?

gyreas:
How do I invoke a red message at cmdline for complaints to the user?

kirawi:
I'm not sure if a message should be returned because I could imagine the bounding also happening when the terminal is being resized.

gyreas:
Command line toggling, yes. It'll provide a good UX.

kirawi:
https://github.com/helix-editor/helix/blob/3a8592abdb9f30c7ebb74f10fa5127e754d18bc6/helix-term/src/commands/typed.rs#L1745
via `cx.set_error()`

roberte777:
> https://github.com/helix-editor/helix/blob/3a8592abdb9f30c7ebb74f10fa5127e754d18bc6/helix-term/src/commands/typed.rs#L1745
> 
> via `cx.set_error()` or returning an error.

I don't think this function really provided an opportunity to return an error, because it's just a general lookup and there are no checks for setting any of the other values. 

I've got a basic PR open, let me know if we should start looking to add errors or if what I've implemented is enough.

gyreas:
I was actually working on it but it's good that you're on it since I don't have the resources to do a PR. 

I'll keep an eye on this

PR #7821, opened 2023-08-03
Skip rendering gutters when gutter width exceeds view width

ArchUsr64:
Closes https://github.com/helix-editor/helix/issues/7154
Setting `gutters.line-numbers.min-width` to a value higher than `terminal_width * terminal_height` led to index out of bounds.  
My approach was to have the width of `LineNumbers` be capped at 90% of the viewport size. Everything works as expected until the threshold is reached, when `min_width` is set to default value.

https://github.com/helix-editor/helix/assets/83179501/6f4b82a2-3c02-49c9-adf0-24683ccedbe6

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -163,15 +163,18 @@ impl EditorView {
             Box::new(highlights)
         };
 
-        Self::render_gutter(
-            editor,
-            doc,
-            view,
-            view.area,
-            theme,
-            is_focused,
-            &mut line_decorations,
-        );
+        let gutter_overflow = view.gutter_offset(doc) == 0;
+        if !gutter_overflow {
+            Self::render_gutter(
+                editor,
+                doc,
+                view,
+                view.area,
+                theme,
+                is_focused,
+                &mut line_decorations,
+            );
+        }
 
         if is_focused {
             let cursor = doc
--- a/helix-view/src/view.rs
+++ b/helix-view/src/view.rs
@@ -187,11 +187,17 @@ impl View {
     }
 
     pub fn gutter_offset(&self, doc: &Document) -> u16 {
-        self.gutters
+        let total_width = self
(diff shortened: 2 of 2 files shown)


pascalkuthe:
As I already mentioned in https://github.com/helix-editor/helix/pull/7156#discussion_r1211604930 fixing this by just limiting the line number width is just a hack. You cos have a gutter with 20 spacers and a terminal width of 20 (the terminal width can even be 0).

So instead of just limiting this setting the right way to fix this is ti perform bkubds check before deciding to even draw gutter elements

ArchUsr64:
> As I already mentioned in [#7156 (comment)](https://github.com/helix-editor/helix/pull/7156#discussion_r1211604930) fixing this by just limiting the line number width is just a hack. You cos have a gutter with 20 spacers and a terminal width of 20 (the terminal width can even be 0).  

My apologies, I misunderstood as it just being an issue with `line-number`. The most recent commits address this issue by disabling gutter rendering if gutters overflow the width.  

The video showcased has 47 `spacer` defined in the `[editor.gutters]` section of the config.

https://github.com/helix-editor/helix/assets/83179501/1fcac2c7-00d4-490e-b721-07bb6eef54d7

helix-term/src/ui/editor.rs
```diff
@@ -163,15 +163,18 @@ impl EditorView {
...
-        Self::render_gutter(
-            editor,
-            doc,
-            view,
-            view.area,
-            theme,
-            is_focused,
-            &mut line_decorations,
-        );
+        let gutter_overflow = view.gutter_offset(doc) == 0;
```
ArchUsr64, line 166: Not sure if this is the right way to go about it or if `gutter_offset` should return an `Option<u16>` instead with `None` in case of overflow

pascalkuthe (approved):

the-mikedavis (approved):

PR #7829, opened 2023-08-04
Add new moon theme to default themes

mohamedimrane:
Here is the theme I ported https://github.com/taniarascia/new-moon


runtime/themes/.DS_Store
the-mikedavis, line 1: This change should be dropped

PR #7832, opened 2023-08-04
CI: Remove the aarch64 appimage build steps from release

### LINKED ISSUE #5237: The helix-22.12-aarch64.AppImage is actually for x86
From the 22.12 release page at https://github.com/helix-editor/helix/releases/tag/22.12 

```
$ file helix-22.12-aarch64.AppImage
helix-22.12-aarch64.AppImage: ELF 64-bit LSB executable, x86-64, version 1 (SYSV),
  dynamically linked, interpreter /lib64/ld-linux-x86-64.so.2,
  BuildID[sha1]=9b0b692e9674b12a987463ee5486fdf2a2d3ff1b,
  for GNU/Linux 2.6.32, stripped
```

the-mikedavis:
The steps mistakenly produce a x86_64 appimage and call it aarch64. linuxdeploy doesn't currently support producing aarch64 appimages so we should just remove these steps for aarch64-linux.

Closes #5237

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -160,7 +160,7 @@ jobs:
 
       - name: Build AppImage
         shell: bash
-        if: matrix.build == 'aarch64-linux' || matrix.build == 'x86_64-linux'
+        if: matrix.build == 'x86_64-linux'
         run: |
           # Required as of 22.x https://github.com/AppImage/AppImageKit/wiki/FUSE
           sudo add-apt-repository universe
@@ -263,7 +263,7 @@ jobs:
               mv bins-$platform/hx$exe $pkgname
               chmod +x $pkgname/hx$exe
 
-              if [[ "$platform" = "aarch64-linux" || "$platform" = "x86_64-linux" ]]; then
+              if [[ "$platform" = "x86_64-linux" ]]; then
                   mv bins-$platform/helix-*.AppImage* dist/
               fi
 


pascalkuthe (approved):

gabydd:
for anyone confused like I was, the ci is failing because ``GITHUB_REF_NAME`` becomes ``7832/merge``(see https://github.com/github/docs/issues/15319) which has a slash in it so we refer to an uncreated directory which doesn't happen in the normal release process

agatekartik:
Appears there is a aarch64 now https://github.com/linuxdeploy/linuxdeploy/releases/download/1-alpha-20240109-1/linuxdeploy-aarch64.AppImage - will we see a corresponding helix AppImage soon?

agatekartik:
> Feel free to submit a PR! (Once a linuxdeploy version with aarch64 support is released)

Thank you - yes might be a good idea, I tried a few things off the cuff (like undoing this commit and then ensuring aarch64 was used to create appimage) - however, from the runner output, looks like the binaries are not of the correct format - which means, while the linuxdeploy tool maybe aarch64, the binaries produced by the build are likely not (note: I haven't done a deep dive yet).
I shall fork and try locally with self hosted runners as time permits and see how to get things done, if they are indeed possible :-)

agatekartik:
I tried getting the linuxdeploy-aarch64 tool to run on my raspberry pi 4. However, I keep getting a file format error of the linuxdeploy binary too. I guess it is best we remain patient till linuxdeploy announces release for aarch64

PR #7833, opened 2023-08-04
feat(health): show info about multiple lsp/dap binaries

woojiq:
This PR introduces two changes:
1. Add support for multiple health arguments. E.g, `hx --health rust ruby` (05d1f94)
2. Show information about all lsp/dap binaries when `hx --health` or `hx --health <lang>` (5c4b873)

It doesn't close any existing issues afaik, but removes some TODOs introduced in #2507.
If you are not happy with the first change, I can rebase the branch to contain only the second.

[asciinema](https://asciinema.org/a/ZMlHi5XpJ8iHpEfgzdRxOL4zp)

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -8,7 +8,7 @@ pub struct Args {
     pub display_help: bool,
     pub display_version: bool,
     pub health: bool,
-    pub health_arg: Option<String>,
+    pub health_arg: Vec<String>,
     pub load_tutor: bool,
     pub fetch_grammars: bool,
     pub build_grammars: bool,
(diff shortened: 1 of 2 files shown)


woojiq:
Ty, I haven't seen this PR, only https://github.com/helix-editor/helix/pull/5692.

> Will it be possible to specify the languages to display as such: hx --health lang1 lang2...langn?
>> Well I don't think it would be difficult to implement, but I think it should be done in a different PR.

So yeah, I can rebase this PR to only include the first change, while the second is already implemented in mentioned PR.

woojiq:
Well, that issue also has a pr. No problem, gonna watch netflix :crying_cat_face:

PR #7835, opened 2023-08-04
feat: add todo.txt tree-sitter

Jan9103:
Add support for todo.txt ([docs](https://github.com/todotxt/todo.txt), [website](http://todotxt.org/)).

Since i often use todo.txt for organizing and markdown is already included i wanted to offer upstreaming my config.

The official file extension is `.todo.txt`, but that doesn't work with helix (and a bunch of other editors) due to the recognition prioritization, therefore i also added the alternative `.todotxt` extension.

This uses the same tree-sitter as [neovim](https://github.com/nvim-treesitter/nvim-treesitter/pull/2594).

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2736,3 +2736,14 @@ indent = { tab-width = 4, unit = "    " }
 name = "unison"
 source = { git = "https://github.com/kylegoetz/tree-sitter-unison", rev = "98c4e8bc5c9f5989814a720457cf36963cf4043d" }
 
+[[language]]
+name = "todotxt"
+scope = "text.todotxt"
+file-types = [{ suffix = ".todo.txt" }, "todotxt"]
+roots = []
+formatter = { command = "sort" }
+auto-format = true
+
+[[grammar]]
+name = "todotxt"
+source = { git = "https://github.com/arnarg/tree-sitter-todotxt", rev = "0207f6a4ab6aeafc4b091914d31d8235049a2578" }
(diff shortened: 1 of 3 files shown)


languages.toml
```diff
@@ -2713,3 +2713,15 @@ injection-regex = "wren"
...
+name = "todotxt"
+scope = "text.todotxt"
+file-types = ["todo.txt", "todotxt"]
+roots = []
+formatter = { command = "sort" }
+auto-format = true
+
+[[grammar]]
+name = "todotxt"
+source = { git = "https://github.com/arnarg/tree-sitter-todotxt.git", rev = "0207f6a4ab6aeafc4b091914d31d8235049a2578" }
```
the-mikedavis, line 2727:
```suggestion
source = { git = "https://github.com/arnarg/tree-sitter-todotxt", rev = "0207f6a4ab6aeafc4b091914d31d8235049a2578" }
```

languages.toml
```diff
@@ -2713,3 +2713,15 @@ injection-regex = "wren"
 file-types = ["wren"]
 roots = []
 indent = { tab-width = 2, unit = "  "}
+
+[[language]]
+name = "todotxt"
+scope = "text.todotxt"
+file-types = ["todo.txt", "todotxt"]
```
the-mikedavis, line 2720:
```suggestion
file-types = [".todo.txt"]
```

This should work to catch `.todo.txt` files

the-mikedavis: Oh I see, I was thinking that `.todo.txt` was the full filename

the-mikedavis: You could use `suffix` here, I think that should catch `test.todo.txt`:

```toml
file-types = [{ suffix = ".todo.txt"}, "todotxt"]
```

the-mikedavis (approved):

PR #7837, opened 2023-08-05
Center the picker preview selection using visual lines

### LINKED ISSUE #7836: global search preview doesn't correctly display the found line
### Summary

i believe this is an issue with soft wrapping - the line that the preview window puts at the top would be correct to center the selected line if lines were not wrapped, but if lines were wrapped, it pushes the selection down, eventually past the bottom of the preview window.

### Reproduction Steps

if i do a global search for `preview` in the helix repository, a file that doesn't require line wrapping displays the found line properly centered:
preview1
but if the file has longer lines, the found line gets pushed downward:
preview2
sometimes off the visible area entirely:
preview3

### Helix log

_No response_

### Platform

linux

### Terminal Emulator

alacritty 0.12.2

### Helix Version

helix 23.05

doy:
this way the preview always shows the selection even if lines were wrapped

fixes #7836

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -27,7 +27,7 @@ use std::{collections::HashMap, io::Read, path::PathBuf};
 
 use crate::ui::{Prompt, PromptEvent};
 use helix_core::{
-    movement::Direction, text_annotations::TextAnnotations,
+    char_idx_at_visual_offset, movement::Direction, text_annotations::TextAnnotations,
     unicode::segmentation::UnicodeSegmentation, Position, Syntax,
 };
 use helix_view::{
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -690,20 +690,19 @@ impl<T: Item + 'static> Picker<T> {
...
-
-            let offset = ViewPosition {
-                anchor: doc.text().line_to_char(first_line),
-                horizontal_offset: 0,
-                vertical_offset: 0,
-            };
+            let text_fmt = doc.text_format(inner.width, None);
+            let annotations = TextAnnotations::default();
+            let mut offset = ViewPosition::default();
+            (offset.anchor, offset.vertical_offset) = char_idx_at_visual_offset(
```
pascalkuthe, line 696: If `range` is `None` we don't need need this. Since scrolling upwards from 0 is not possible anyway so I would siimoly change this to 

``` rust
if let Some(range) = range {
 (offset.anchor, offset.vertical_offset) = ...
}
```

pascalkuthe (dismissed):
One small not, otherwise thus liims good. This was a good catch and something I missed in #5420

doy:
fixed

pascalkuthe (approved):

the-mikedavis (approved):

PR #7838, opened 2023-08-05
Add ltex-ls language server

David-Else:
`ltex-ls` now supports spelling and grammar correction for git-commit files, so I think it makes sense to support it by default. Everyone could benefit from it as Helix has no built-in spelling/grammar checker.

Screenshot from 2023-08-05 13-50-58

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -46,7 +46,7 @@
 | fsharp | ✓ |  |  | `fsautocomplete` |
 | gdscript | ✓ | ✓ | ✓ |  |
 | git-attributes | ✓ |  |  |  |
-| git-commit | ✓ | ✓ |  | `ltex-ls` |
+| git-commit | ✓ | ✓ |  |  |
 | git-config | ✓ |  |  |  |
 | git-ignore | ✓ |  |  |  |
 | git-rebase | ✓ |  |  |  |
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -46,7 +46,7 @@
 | fsharp | ✓ |  |  | `fsautocomplete` |
 | gdscript | ✓ | ✓ | ✓ |  |
 | git-attributes | ✓ |  |  |  |
-| git-commit | ✓ | ✓ |  |  |
+| git-commit | ✓ | ✓ |  | `ltex-ls` |
 | git-config | ✓ |  |  |  |
 | git-ignore | ✓ |  |  |  |
 | git-rebase | ✓ |  |  |  |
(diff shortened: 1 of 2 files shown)


pascalkuthe (dismissed):
I have this in my config. I think it makes sense to have this for git commits by default

dead10ck:
Sorry, I disagree. I don't want an LS to spin up every time I do a git commit. Besides that, there are specific contexts where spell checking by default makes sense, but git is used for all kinds of technical projects where a spell checker would just be way too noisy. Even in the screenshot here there's a false positive on the LS's own name.

David-Else:
@dead10ck You are free to ignore the results, and there is no delay while the language server loads, so how come you don't want it spinning up in the background?

You can easily add words to the spellchecker in the config, even in multiple languages:

```toml
ltex.dictionary = { "en-US" = [
  "ltex-ls",
], "en-GB" = [
  "ltex-ls",
] } 
```

I also assume you can easily disable it if you want.

dead10ck:
> @dead10ck You are free to ignore the results, and there is no delay while the language server loads, so how come you don't want it spinning up in the background?

It's wasteful to spin up a LS for editing a tiny document that's rarely open for more than a minute max, especially when it's unwanted in the first place.

>You can easily add words to the spellchecker in the config, even in multiple languages:
> 
> ```toml
> ltex.dictionary = { "en-US" = [
>   "ltex-ls",
> ], "en-GB" = [
>   "ltex-ls",
> ] } 
> ```

Git is used for a wide variety of technical contexts using language that a spell checker will flag erroneously. Every time you use the name of a class, tool, abbreviation, variable, module, keyword, etc, etc, it will add noise to the document. They can't all go in an ignore list.

In my opinion, git is used in too many disparate contexts to justify turning on a spell checker by default. It's very visually distracting and difficult to ignore.

> I also assume you can easily disable it in your config if you want.

This goes both ways. 🙂

Also, this is just my opinion. I don't have final say.

clo4:
Just my 2c as a user: I think @dead10ck is probably right, this shouldn't be a default, but it *should* be easy to enable — if there's a place to put "recipes" like enabling this (wiki or docs), that would be a great home for it!

David-Else:
People seem split on this, shall I just keep the general `ltex-ls = { command = "ltex-ls" }` and remove `language-servers = [ "ltex-ls" ]` from `git-commit` so it is slightly easier for people to make the change themselves?

Maybe @archseer  needs to make an executive decision on this small but seemingly controversial change?

the-mikedavis:
I'm also hesitant to have this set for git-commits by default but adding a default config for ltex-ls so it's easy to reference in `~/.config/helix/languages.toml` seems like a nice win to me

dead10ck (approved):
lgtm, thanks!

David-Else:
I have done as requested.

pascalkuthe (approved):

Issue #3850, opened 2022-09-15
Split Bug at Start of Line


joegm:
### Summary

When <kbd>S</kbd>plit is used with a pattern that matches the start of a line, a selection is placed on the match even when there aren't any characters before it:

https://asciinema.org/a/NDCiZr7kfsMiUEuI88K37vybT

### Reproduction Steps

 

I tried this:

1. `hx`
2. <kbd>i| | | |(esc)S\\|(ret)</kbd>

I expected this to happen:
The spaces in between and the newline would be selected

Instead, this happened:
There was an additional selection at the start of the line.

### Helix log

N/A

### Platform

Linux

### Terminal Emulator

Alacritty

### Helix Version

22.08.1-61-gb191ecf8

the-mikedavis:
This looks like https://github.com/helix-editor/helix/issues/3544. The same fix from https://github.com/helix-editor/helix/pull/3598/files can probably be applied to `split_on_matches`

palango:
I think there are some edge cases that need to be decided. For example, on the input `|a|b|c|`, and a split on `|` what would be the expected outcome: three or four selections?

For me it feels more natural to have three, but have the special case of having the first selection remaining if there is no match. I implemented this in https://github.com/palango/helix/commit/17d1cd3d7be4677f635632d68af2c4612d6e4d31#diff-f218cebd639a867087e2a7abc84160d45888ad5bcff4ab10fcf7217cb16586e9R1303 , however the code isn't optimized yet.

dorianmeric:
I guess the question is: if the split results in a zero-length selection, should you have a selection or not? I would say so.

If you have a CSV file with this text:
```
name,size,orderid
ab,12,1
,11,2
,,,
da,20,4
```
And then you split the selection with `% S , <Enter>` , you would want a selection at the beginning of each row (so that you have 3 selections per row, so that you could insert a prefix in each cell with `iprefix` for example.


Strangely enough, as of v25.07.1, on the very first line of the selection, the split behaves as I suggested (meaning it does create a selection at the first position), but if you have 2 or more lines and do S, the first line has a selection in the first position but not the following lines. So the behaviour changes depending on whether you are splitting one line or multiple.

So doing `% S , <ENT>` on this gives 5 selections:
```
,,,,
```
but doing `% S , <ENT>` on this gives 5 selections then 4 selections on the other 2 lines:
```
,,,,
,,,,
,,,,
```
[...]

PR #7854, opened 2023-08-07
fix unexpected first selection when line splitting

brendonh8:
fixes [https://github.com/helix-editor/helix/issues/3850](3850)

Original logic was intentionally following inclusivity similar to `Range` for returned selections after splitting, so inclusive on left and exclusive on right. However, when you have a match right at the start of a selection, you will get a returned selection before that initial match. This doesn't make sense when the rest of the selections will always come after a match.

### DIFF
--- a/helix-core/src/selection.rs
+++ b/helix-core/src/selection.rs
@@ -770,12 +770,6 @@ pub fn split_on_matches(
     let mut result = SmallVec::with_capacity(selection.len());
 
     for sel in selection {
-        // Special case: zero-width selection.
-        if sel.from() == sel.to() {
-            result.push(*sel);
-            continue;
-        }
-
         // TODO: can't avoid occasional allocations since Regex can't operate on chunks yet
         let fragment = sel.fragment(text);
 
@@ -789,11 +783,15 @@ pub fn split_on_matches(
         for mat in regex.find_iter(&fragment) {
             // TODO: retain range direction
             let end = text.byte_to_char(start_byte + mat.start());
-            result.push(Range::new(start, end));
+            let range = Range::new(start, end);
+            // Don't add a zero-width range if a match is at the start of selection
+            if range != Range::point(sel.from()) {
+                result.push(range);
+            }
             start = text.byte_to_char(start_byte + mat.end());
         }
 
-        if start < sel_end {
+        if start <= sel_end {
             result.push(Range::new(start, sel_end));
         }
     }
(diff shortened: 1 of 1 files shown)


archseer (approved):

helix-core/src/selection.rs
```diff
@@ -1185,25 +1183,22 @@ mod test {
...
+                // the previous range is the start of the current match. Any anchors should
+                // come directly after a match
                 Range::new(1, 5),
                 Range::new(6, 9),
                 Range::new(11, 13),
                 Range::new(16, 19),
-                // In contrast to the comment above, there is no
-                // _trailing_ zero-width range despite the trailing
-                // match, because ranges are exclusive on the right.
+                Range::new(20, 20),
```
the-mikedavis, line 1195: I believe we don't want this trailing point either. With (essentially) the same reproduction case as #3850: `i| | | |<esc>%_S\|<ret>`: we have a trailing selection created at the end which lives outside of the selection at the start of `S` (when the selection is over `| | | |`). I believe that case should leave only the spaces between the `|`s selected.

helix-core/src/selection.rs
```diff
@@ -789,11 +783,15 @@ pub fn split_on_matches(
         for mat in regex.find_iter(&fragment) {
             // TODO: retain range direction
             let end = text.byte_to_char(start_byte + mat.start());
-            result.push(Range::new(start, end));
+            let range = Range::new(start, end);
+            // Don't add a zero-width range if a match is at the start of selection
+            if range != Range::point(sel.from()) {
```
the-mikedavis, line 788: I wonder if it is enough to test that the range is not a point? Is there any case where we want a selection over a zero-width match?

pascalkuthe: yeah I actually think helix currently handles zero width matches weirdly. For example you can currently use an emply anchor match `^$`  to match empty lines (or `|` to match every char index) in helix. This works for forward search but not reverse search. This behavior is incorrect. I think we should always ignore zero-width matches so that there is always some unambiguous text to select.

There may be an argument that you want to keep search working in that case (since you could view that as selection being a secondary instead of primary effect) but for `s` in particular the primary goal is to select the text which matches the regex pattern and `^$` matches no text (and for consistency and avoiding weird edgecases I think we should do the same for search)

ViktorKarpilov (approved):

PR #7869, opened 2023-08-08
Add text objects for protobuf

john-h-k:
* `service` and `message` are type objects
* `rpc`s are functions, both `around` and `inside` selects entire line (as they have no body)
* identifiers within `rpc`s are parameters, although pretty useless as there can only be one


the-mikedavis (commented):
If you run `cargo xtask docgen` and commit the results that will fix the docs CI

the-mikedavis (approved):

pascalkuthe:
Seems #9184 was already merged. Can this be closed then?

the-mikedavis:
Yep it looks to me like this and #9184 have almost the same patterns, just written out differently

PR #7875, opened 2023-08-08
Detect non-existent files as non-readonly

the-mikedavis:
Previously all files not backed by an existing file (`:open does-not-exist.txt`) would be marked as readonly in the new read-only-indicator statusline element (#7740).


pascalkuthe (approved):
thanks!

PR #7878, opened 2023-08-08
feat: pest language server support

Jamalam360:
support for the [pest](https://pest.rs) language server: https://github.com/pest-parser/pest-ide-tools 

tested and working. i'll be back when we have a tree sitter grammar, we only have textmate for now unfortunately.


the-mikedavis:
Generally we don't add language-server-only configs. Let's wait until there's a tree-sitter grammar available for pest

Jamalam360:
Sure :)

elenakrittik:
Maybe convert this to draft PR instead?

PR #7884, opened 2023-08-09
Document Alt-Enter for the picker

bjorn-ove:
I found out about the existence of `Alt-Enter` when I noticed #7691 being merged. I then tested it in the `space-f` picker and found it worked there as well. Since this seems to be consistent behavior, I figured it should be documented.

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -411,19 +411,20 @@ you to selectively add search terms to your selections.
 
 Keys to use within picker. Remapping currently not supported.
 
-| Key                          | Description       |
-| -----                        | -------------     |
-| `Shift-Tab`, `Up`, `Ctrl-p`  | Previous entry    |
-| `Tab`, `Down`, `Ctrl-n`      | Next entry        |
-| `PageUp`, `Ctrl-u`           | Page up           |
-| `PageDown`, `Ctrl-d`         | Page down         |
-| `Home`                       | Go to first entry |
-| `End`                        | Go to last entry  |
-| `Enter`                      | Open selected     |
-| `Ctrl-s`                     | Open horizontally |
-| `Ctrl-v`                     | Open vertically   |
-| `Ctrl-t`                     | Toggle preview    |
-| `Escape`, `Ctrl-c`           | Close picker      |
+| Key                          | Description                                                |
+| -----                        | -------------                                              |
+| `Shift-Tab`, `Up`, `Ctrl-p`  | Previous entry                                             |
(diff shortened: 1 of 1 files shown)


poliorcetics (approved):

woojiq:
I wanted to add the docs while working on that PR, but it completely slipped my mind.

pascalkuthe (dismissed):

book/src/keymap.md
```diff
@@ -411,19 +411,20 @@ you to selectively add search terms to your selections.
...
+| Key                          | Description                     |
+| -----                        | -------------                   |
+| `Shift-Tab`, `Up`, `Ctrl-p`  | Previous entry                  |
+| `Tab`, `Down`, `Ctrl-n`      | Next entry                      |
+| `PageUp`, `Ctrl-u`           | Page up                         |
+| `PageDown`, `Ctrl-d`         | Page down                       |
+| `Home`                       | Go to first entry               |
+| `End`                        | Go to last entry                |
+| `Enter`                      | Open selected                   |
+| `Alt-Enter`                  | Open selected in the background |
```
the-mikedavis, line 423:
```suggestion
| `Alt-Enter`                  | Open selected without closing the picker |
```

I agree with woojiq: the "background" part of this might be confusing

bjorn-ove: But it doesn't change the current document, it opens hidden, and then you switch to it. Not saying "in the background" is better, but I think "without closing the picker" is more confusing as nothing appears to be happening behind the picker

bjorn-ove: I see the point now. It is confusing that the picker doesn't close. But I still think it is also confusing that it doesn't open the focused window, so there is something missing. Perhaps what we would want is `Opens selected in the background without closing the picker`, however I guess that is too long?

bjorn-ove: How about `Open in background & keep picker open`

woojiq: `Open selected in the background without closing the picker` sounds good. It isn't too long, because if you look at [docs](https://docs.helix-editor.com/keymap.html#prompt) you will find even longer description.

bjorn-ove: I'm away from my computer until Monday, but will make the change as soon as possible. Feel free to close this PR and make a new one if necessary.

bjorn-ove:
Updated

the-mikedavis (approved):

pascalkuthe (approved):

PR #7814, opened 2023-08-03
transition to nucleo for fuzzy matching

pascalkuthe:
Closes #1707
Closes #1987
Closes #1543
Closes #5871
Closes #7645
Closes #7652

~**NOTE** This PR is currently still experimental I want to tweak the fuzzy matching scoring calculations in nucleo a bit more, add more tests, cleanup the nucleo codebase. There might also be some issues with the completers~

This PR fully replaces the skim matcher with used for all fuzzy matching in helix currently (completions, picker, ..) with `nucleo`. I have written nucleo to be as performant as possible (heavily referenced fzf). The matching algorithm itself can often be 8 times faster than skim.

https://github.com/helix-editor/nucleo

Furthermore, `nucleo` offers a high-level API that essentially replaces the entire picker logic in helix. Compared to the current helix `nucleo`s' high-level API comes with some nice features:

* matching in a background thread pool (no freezing the UI thread)
* match items can be streamed into the matches (lock-free)

I think the following demonstrates the practical use case well:

I am able to fuzzy match to the helix changelog very quickly.. From my root directory:

https://asciinema.org/a/MnRf8JGPFll1jw1GpHwNGGCXU

On the current helix version I have to wait for all (3 million) files to be crawled (with the editor frozen in the meantime) and matching is slowed to a crawl afterwards.

Just for completeness another capture of actually waiting for all files to stream in and fuzzy matching on the result (still just as fast/no noticeable lag)

https://asciinema.org/a/v1RZmq5wV80hf2BWSM1RUqaod

Nulceo is **faster than fzf** (see the comparison in the README) while yielding the same results (and even better results in some cases)

PR #7891, opened 2023-08-09
use rust 1.68 in rust-toolchain.toml

gabydd:
fixes the problem found in https://github.com/helix-editor/helix/pull/7814#issuecomment-1666514270 which was an existing problem in helix that could occur while file-walking, this occured due to a bug in rust which was fixed and in included in the 1.68 release. This pr only updates the rust-toolchain.toml so that users building from source won't encounter this bug, but our msrv is still 1.65 which is what we check in CI. As before releases are still built with the latest rust stable so nothing needs to change there.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -15,12 +15,12 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
-      - name: Install stable toolchain
-        uses: helix-editor/rust-toolchain@v1
-        with:
-          profile: minimal
-          override: true
 
+      - name: Remove rust-toolchain.toml file
+        run: rm rust-toolchain.toml
+
+      - name: Install stable toolchain
+        uses: dtolnay/rust-toolchain@1.65
       - uses: Swatinem/rust-cache@v2
 
       - name: Run cargo check
@@ -35,6 +35,9 @@ jobs:
     steps:
       - name: Checkout sources
         uses: actions/checkout@v3
+
+      - name: Remove rust-toolchain.toml file
+        run: rm rust-toolchain.toml
 
       - name: Install stable toolchain
         uses: dtolnay/rust-toolchain@1.65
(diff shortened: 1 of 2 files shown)


pascalkuthe (approved):
It's kind of annoying that we have to use this workaround. Maybe upstream can deal with rust-toolchain.tonl field better in the future.

There are a bunch of closed issues/PRs about automatically detecting the rust version from the rust-toolchain.tonl file but no issues about ignoring/overwriting that file. Perhaps they will be open to that 


Regardless thanks for investigating!

gabydd:
I was looking into that as well we could override the tool chain maybe: https://rust-lang.github.io/rustup/overrides.html

pascalkuthe:
Yeah an directory overwrite would again have to be done during every CI run so it's not better than just deleting the file temporarily

pascalkuthe:
Btw. for the release CI' doesn't that also use rustup? I think the rust-toolchain.toml file always takes precedence no matter his we configure CI.

The rust-toolchain workflow works by installing the toolchain and setting the rustup default toolchain. It does nothing soecicial so just like locally the rust-toolchain.toml file always takes priority

pascalkuthe:
Ah nvm I mussed that the release CI dies the exact same thing as this PR. So it's reasonable to do the same here I think

gabydd:
yeah we remove the rust-toolchain.toml in release, also I think we can set the ``RUSTUP_TOOLCHAIN`` environment variable which takes precedence over the ``rust-toolchain.toml`` but I guess it's also fine to remove it as well

pascalkuthe:
The env variabke would be yet another place where we would have to set the MSRV tough. I think J would prefer to just remove the file in CI.

I missed ut before but release CI already does this too :D

the-mikedavis:
I think we actually want to keep the in-tree `rust-toolchain.toml` at the MSRV. It's better for local development since `check`/`clippy` will catch MSRV violations. It's a rough edge for those `cargo install`ing from source but the bug fixed in 1.68 is an unlikely edge-case to be hit while using Helix normally: I suspect that opening a file picker at `/` is very rarely done. Anyways we will probably be bumping the MSRV to 1.68 soon enough: the [Mozilla docs](https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html) seem maybe out of date but I believe the ubuntu packages are already caught up to 1.67.

PR #7899, opened 2023-08-10
Align space indentation to the nearest level when inserting

### LINKED ISSUE #2124: Snap tab to nth column
# Current functionality
assuming tab size is 4, the tab key inserts 4 spaces.
`a<Tab>b` results in `a _ _ _ _ b`
# Suggested functionality
assuming tab size is 4, the tab key inserts spaces until the cursor is at a column that is a multiple of 4, like in Vim, NeoVim, and most terminal shells.
`a<Tab>b` results in `a _ _ _ b`

manokara:
Currently, when executing `insert_tab` with `indent-style` set to a spaces multiplier the editor always adds a fixed amount of spaces regardless of where the cursor is, which is not the expected UX. Fixes #2124.

- [x] Insert necessary amount of spaces for the next indentation level
- [ ] (Possibly) fix deidentation bug with `delete_char_backward` when there are multiple selections
- [ ] Add tests

## Sample

```
assuming `indent-style 4` and `set smart-tab.enable false`

==========================
Initial state
--------------------------
0   1   2   3   4   5   6
*---*---*---*---*---*---*-
foobar
 barfoo
   foobar
==========================

==========================
Press tab on the first char of each line
--------------------------
0   1   2   3   4   5   6
*---*---*---*---*---*---*-
    foobar
    barfoo
    foobar
==========================

==========================
Filtering 'bar' from the selection of the three lines, then pressing tab once
--------------------------
0   1   2   3   4   5   6
*---*---*---*---*---*---*-
foo bar
    barfoo
   foo  bar
==========================

==========================
Filtering 'bar' from the selection of the three lines, then pressing tab twice
--------------------------
0   1   2   3   4   5   6
*---*---*---*---*---*---*-
foo     bar
        barfoo
   foo      bar
==========================

==========================
Filtering 'bar' from the selection of the three lines, then pressing tab twice, then backspace once
--------------------------
0   1   2   3   4   5   6
*---*---*---*---*---*---*-
     -- EXPECTED --
foo bar
    barfoo
   foo  bar

     -- RESULT --
foo    bar
    barfoo
   foo     bar
==========================
```

For some reason, `delete_char_backward` only correctly deindents the middle line regardless of which one is the primary selection. This might be out of scope for this PR, though, as it would be a bug fix for the `delete_char_backward` command.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -10,7 +10,7 @@ use tui::widgets::Row;
 pub use typed::*;
 
 use helix_core::{
-    char_idx_at_visual_offset, comment,
+    char_idx_at_visual_offset, comment, coords_at_pos,
     doc_formatter::TextFormat,
     encoding, find_first_non_whitespace_char, find_workspace, graphemes,
     history::UndoKind,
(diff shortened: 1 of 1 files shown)


manokara:
There's a bug currently with this implementation (that I'll get back to when I have more free time) that if you tab after inserting a new line, it will indent the line below instead of the current one. I think it's because `insert_newline` is implicity extending the selection to the next non-whitespace character:

https://github.com/helix-editor/helix/blob/57f093d83641642ad5d4ba42ae59f03272efcfcc/helix-term/src/commands.rs#L3549-L3557

pascalkuthe (commented):
I didn't finish my review yet but some comments already. What makes this tricky to implement (and what you don't handle yet) is multicursors. If you have multiple multicurosors on the same line then the code here would be buggy. To fix that you need to account for previous cursors on the same line in later cursors. This is a bit tricky to do since `visual_coords_at_pos` always assumes the slice it receives to be at these rtart of the line. you likely want to add ann `offest parameter that can be used to supply an offset (which would be the width of the line up to the cursor plus any extra text already inserted into that line)

pascalkuthe:
a feature I would definitly like to see but this has unfortunly gone stale/been inactive for  long while so I am closing this. Thank you for contributing!

PR #7900, opened 2023-08-10
feat(lsp): support textDocument/diagnostic specification

### LINKED ISSUE #7757: Add pullDiagnostics
Some of LSP support only [pull diagnostics](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#textDocument_pullDiagnostics)
source here https://github.com/helix-editor/helix/discussions/7725

woojiq:
Closes #7757 

Explanation for easier review, see https://github.com/helix-editor/helix/pull/7900#issuecomment-1675911104.

Tested manually with `ruby-lsp`.

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -235,22 +235,26 @@ impl<'de> Deserialize<'de> for FileType {
 #[serde(rename_all = "kebab-case")]
 pub enum LanguageServerFeature {
     Format,
+    // Goto, use bitflags, combining previous Goto members?
     GotoDeclaration,
     GotoDefinition,
     GotoTypeDefinition,
     GotoReference,
     GotoImplementation,
-    // Goto, use bitflags, combining previous Goto members?
+
     SignatureHelp,
     Hover,
     DocumentHighlight,
     Completion,
     CodeAction,
     WorkspaceCommand,
+    // Symbols, use bitflags, see above?
     DocumentSymbols,
     WorkspaceSymbols,
-    // Symbols, use bitflags, see above?
+
+    // Diagnostic in any form that is displayed on the gutter and screen
     Diagnostics,
+    PullDiagnostics,
     RenameSymbol,
     InlayHints,
 }
@@ -274,6 +278,7 @@ impl Display for LanguageServerFeature {
             DocumentSymbols => "document-symbols",
             WorkspaceSymbols => "workspace-symbols",
             Diagnostics => "diagnostics",
+            PullDiagnostics => "pull-diagnostics",
             RenameSymbol => "rename-symbol",
             InlayHints => "inlay-hints",
         };
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -15,13 +15,13 @@ use lsp_types as lsp;
 use parking_lot::Mutex;
 use serde::Deserialize;
 use serde_json::Value;
-use std::future::Future;
-use std::process::Stdio;
 use std::sync::{
     atomic::{AtomicU64, Ordering},
     Arc,
 };
 use std::{collections::HashMap, path::PathBuf};
+use std::{future::Future, sync::atomic::AtomicBool};
+use std::{process::Stdio, sync::atomic};
 use tokio::{
     io::{BufReader, BufWriter},
     process::{Child, Command},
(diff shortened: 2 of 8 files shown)


woojiq:
Few notes for easier reviewers. Two commits. First one fixes when you have several servers in the language configuration, and one of them is unable to start, then everyone else also fails. This bug was introduced in https://github.com/helix-editor/helix/pull/2507.

The second commit is introducing Diagnostic support. I took a large piece of code from `PublishDiaagnostic` notification and splited in two methods:
* [`lsp_diagnostic_to_diagnostic`](https://github.com/helix-editor/helix/pull/7900/files#diff-e6ed08649c7d2a17eb2784c889bde112e25d7c717b152ea584257177fca9bd63R124)
* [`add_diagnostics`](https://github.com/helix-editor/helix/pull/7900/files#diff-ca59dc0cbfb3ce30bc37926842acf5b65898146ac8c78c9e29229c933a480247R1742)

Added [`text_document_diagnostic`](https://github.com/helix-editor/helix/pull/7900/files#diff-c91126f95089071bac8842b1e47f3983b2a3ba223f572dbd7f26e0f15e3dd93dR1168) method to send actuall json-rpc request, it looks like any other `text_document_*` method (hover, definition etc). And the actual thing you want to review is [`pull_diagnostic_for_current_doc`](https://github.com/helix-editor/helix/pull/7900/files#diff-abe37ebb421b375471daa248c8de6541335e61dd3dfc93ee999eb108876e5837R1500) which is called later in [`handle_idle_timeout` ](https://github.com/helix-editor/helix/pull/7900/files#diff-408ac3559e5a464e3cce528ed57e0bf48453d28c95212437790ac5dec0f450beR1011). There is nothing hard tbh. Just send a server request, then create callback for extracting info from response (aforementioned method `lsp_diagnostic_to_diagnostic` and `add_diagnostics` are reused).
And the last thing to mention is I added field [`supports_publish_diagnostic`](https://github.com/helix-editor/helix/pull/7900/files#diff-c91126f95089071bac8842b1e47f3983b2a3ba223f572dbd7f26e0f15e3dd93dR62) for `Client`, during first "PublishDiagnostic` notification from a server set this field to true. And then, if it is true, we do not run PullDiagnostic. I think we should prefere Publish.

pascalkuthe:
I think it makes sense to delay this until the event system (#8021) lands. We really only want to pull diagnostics of all *visible* documents (and only when their contents actually changed, likely with a debounce). This kind of use case is precisely what I intended the event system for.

Additionally, the support implemented here is still lacking some features that are need to make pull diagnostics really work:

* have the capability to ask the client to refresh diagnostics that we need to respond to: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#diagnostic_refresh
* have a separate request for workspace diagnostics (which should be streamed into the picker using nulceo): https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#workspace_diagnostic
* allow incremental updates (so the server doesn't resend diagnostics that didn't actually change)
* VScode allows language plugins to configure whether to only pull diagnostics on save. This is somewhat related to #6447. It may make sense to offer a per language config option for that.

I think the last two points are pretty optional but the first tow capabilities seems pretty essential to me.

woojiq:
@pascalkuthe thanks for the extensive answer. I agree with you. I'm going to rework this PR after the event system is deployed. If for some reason I miss this part, feel free to ping.✌️

4ndv:
@woojiq looks like the event system was merged recently

woojiq:
> @woojiq looks like the event system was merged recently

I will have a look. What language server do you use that you need pullDiagnostic, ruby-lsp?

4ndv:
> > @woojiq looks like the event system was merged recently
> 
> I will have a look. What language server do you use that you need pullDiagnostic, ruby-lsp?

Planning to, yes. Right now I'm using lsp mode in rubocop, but it lacks some stuff

woojiq:
Unfortunately, I am no longer interested in it for two reasons:
1. none of language servers I use (clangd, rust-analyzer) support this;
2. much more work need to be done than I thought

Although implementing this probably shouldn't be that difficult since almost all the building blocks for working with LSP are already in the codebase.

I might come back to it in the summer, but point 1 is crucial. Sorry for the false expectations.

David-Else:
Seems from version 3.0.1 `vscode-eslint` uses textDocument/diagnostic: https://github.com/hrsh7th/vscode-langservers-extracted/commit/859ca87fd778a862ee2c9f4c03017775208d033a#commitcomment-142170236

PR #7907, opened 2023-08-11
Add pod highlighting

doy:


runtime/queries/pod/highlights.scm
the-mikedavis, line 1: The `#match?` predicates here can be replaced with `#eq?`. `#eq?` is an exact match while `#match?` compiles the pattern as a regex

doy: fixed

runtime/queries/pod/highlights.scm
```diff
@@ -0,0 +1,61 @@
...
+  (sequence_letter) @letter
+  (#match? @letter "F")
+  (content) @markup.italic)
+(interior_sequence
+  (sequence_letter) @letter
+  (#match? @letter "I")
+  (content) @markup.italic)
+(interior_sequence
+  (sequence_letter) @letter
+  (#match? @letter "L")
```
pascalkuthe, line 60: i think the same comment from @the-mikedavis should also be applied here

doy: fixed

pascalkuthe (approved):

the-mikedavis (approved):

PR #7909, opened 2023-08-11
feat: open file picker on directories using goto_file (gf)

### LINKED ISSUE #7906: Support opening a folder with goto_file
When the selection points to a folder and the user uses `goto_file` (gf), the editor will raise an error.

I would propose that it opens a file picker with the elements of the folder, similar to when you launch helix by passing it a folder name.

Thoughts?

(I could work on a PR, with some guidance)

cbr9:
Closes #7906.

Potential problems:
- Relative paths are resolved with reference to the current working directory, instead of the file they are selected from (this was already the behavior, I tried to fix it but couldn't)

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1191,12 +1191,11 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
         let p = sel.trim();
         if !p.is_empty() {
             let path = Path::new(p);
-            if let Ok(path) = helix_core::path::get_canonicalized_path(path) {
-                if path.is_dir() {
-                    file_picker_in(cx, path);
-                } else if let Err(e) = cx.editor.open(&path, action) {
-                    cx.editor.set_error(format!("Open file failed: {:?}", e));
-                }
+            if path.is_dir() {
+                let picker = ui::file_picker(path.into(), &cx.editor.config());
+                cx.push_layer(Box::new(overlaid(picker)));
+            } else if let Err(e) = cx.editor.open(path, action) {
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -1190,8 +1190,13 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     for sel in paths {
         let p = sel.trim();
         if !p.is_empty() {
-            if let Err(e) = cx.editor.open(&PathBuf::from(p), action) {
-                cx.editor.set_error(format!("Open file failed: {:?}", e));
+            let path = Path::new(p);
+            if let Ok(path) = helix_core::path::get_canonicalized_path(path) {
```
pascalkuthe, line 1194: you don't need to cannonicalize here

helix-term/src/commands.rs
```diff
@@ -2561,14 +2566,18 @@ fn append_mode(cx: &mut Context) {
     doc.set_selection(view.id, selection);
 }
 
+fn file_picker_in(cx: &mut Context, root: PathBuf) {
+    let picker = ui::file_picker(root, &cx.editor.config());
+    cx.push_layer(Box::new(overlaid(picker)));
+}
+
```
pascalkuthe, line 2573: adding a helper for two lines of code is really nod necessary and will cause merge conflicts with other changes to the picker

pascalkuthe (changes requested):
I think this makes sense. This makes `gf` matches `:open`

the-mikedavis (approved):

pascalkuthe (approved):

PR #7920, opened 2023-08-12
Prevent GraphemeStrs created from Strings from leaking

dilr:
I noticed that the `From<String>` impl of `GraphemeStr` doesn't set the high bit on `self.len`.  My understanding is that this would mean that the backing memory for the string will not get dropped when the `GraphemeStr` is dropped. This PR fixes the that by setting the high bit during conversion.

### DIFF
--- a/helix-core/src/graphemes.rs
+++ b/helix-core/src/graphemes.rs
@@ -481,7 +481,7 @@ impl<'a> From<String> for GraphemeStr<'a> {
         let ptr = Box::into_raw(g.into_bytes().into_boxed_slice()) as *mut u8;
         GraphemeStr {
             ptr: unsafe { NonNull::new_unchecked(ptr) },
-            len: i32::try_from(len).unwrap() as u32,
+            len: (i32::try_from(len).unwrap() as u32) | Self::MASK_OWNED,
             phantom: PhantomData,
         }
     }


pascalkuthe (approved):
That's a good catch! In practice multi codepoint graphemes (that are not CRLF since ropey doesn't split those) that span across chunk boundaries are very rare so this doesn't usually get bkriced in practice but it's good to fix nonetheless 👍

dilr:
This isn't part of the PR, but if a `1<<31` sized `&str` will currently convert to an "owned" `GraphemeStr`, and get double freed. Helix should never do this if `GraphemeStr` is only supposed to represent one grapheme. However, if a bad actor can trick Helix into creating such a grapheme (for instance, by putting invalid utf-8 in a file which a helix user edits), then they might be able to use the double free to start a ROP gadget chain or something. I don't know the code base well enough to know if it would actually be possible to do so though.

pascalkuthe:
they can't because `i32::try_from(len).unwrap()` will painc in that case

dilr:
I think that would be the case if we had a `1<< 32` sized string, but a `1 << 31` sized string should pass the try_from test. The entire point of the module is that we are using the biggest bit of `len` to store ownedness. However, `u32::try_from` doesn't know about that.

pascalkuthe:
[that's why ](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=9ca08215c2127c84bd9335a434810ca8)

dilr:
Ah, I see what you mean. That's clever!

the-mikedavis (approved):

Issue #1017, opened 2021-11-08
Unimpaired pick item


QiBaobin:
#### Describe your feature request

As we already remember last picker status, would it be good to add a pair of unimpaired keys (something like `[p, [P,  ]p, ]P`) to navigate to next/previous item in last picker? This would reduce the effort for navigation among workspace searching, buffers, symbols and all other supported pickers.

pickfire:
Note, `]p` and `[p` I am planning to use it to move between paragraph. I wonder if this should be under space buffer or something.

PR #7921, opened 2023-08-12
Goto next/prev in last picker

koalefant:
Hi, I am long time Vim user. I am tired of managing Vim plugins and configuration. Helix has a fantastic feature set out of the box and I would be happy to transition to it, however there are some features missing that are critical to my daily workflow.

One of them is being able to jump between global search results quickly. I discovered "last picker" but it  takes five key strokes to use it the way I would like to (`Space`, `'`, `Ctrl+n`, `Enter`). I would be happy to get it down to two.

I noticed in #1017  there were a suggestion to use `[p`, `]p`, I am proposing to use `[[` and `]]` as it is easy to type and would be useful for all picker types. 

I realize that proposed implementation is ugly and fragile. I am looking for guidance to improve it.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -35,7 +35,7 @@ use helix_view::{
     document::{FormatterError, Mode, SCRATCH_BUFFER_NAME},
     editor::{Action, CompleteAction},
     info::Info,
-    input::KeyEvent,
+    input::{Event,  KeyEvent, KeyModifiers},
     keyboard::KeyCode,
     tree,
     view::View,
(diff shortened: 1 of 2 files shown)


the-mikedavis:
See https://github.com/helix-editor/helix/pull/3360, I would prefer this to be solved with #1383

koalefant:
I was not aware of macros: thank you! :pray: 

There is, however, an opportunity to give more context being a native command, for example one could show a match index and total number of matches without popping up the picker.

Perhaps this desires stems from the fact that the current implementation of picker provides little context when number of matches is large and and is actually less ergonomic than jumping directly to the location as it requires extra step to do so.

PR #7922, opened 2023-08-13
Add .star as starlark file extension

dastbe:
In addition to the other defined extensions, `.star` is a frequently used extension for starlark files. This can be demonstrated through a cursory search of github for files ending in `.star` here: https://github.com/search?q=path%3A%2F.star%24%2F&type=code

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1915,7 +1915,7 @@ source = { git = "https://github.com/sogaiu/tree-sitter-clojure", rev = "e57c569
 name = "starlark"
 scope = "source.starlark"
 injection-regex = "(starlark|bzl|bazel)"
-file-types = ["bzl", "bazel", "BUILD"]
+file-types = ["bzl", "bazel", "BUILD", "star"]
 roots = []
 comment-token = "#"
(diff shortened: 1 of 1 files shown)


languages.toml
```diff
@@ -1914,7 +1914,7 @@ source = { git = "https://github.com/sogaiu/tree-sitter-clojure", rev = "e57c569
 [[language]]
 name = "starlark"
 scope = "source.starlark"
-injection-regex = "(starlark|bzl|bazel)"
+injection-regex = "(starlark|star|bzl|bazel)"
 file-types = ["bzl", "bazel", "BUILD"]
```
the-mikedavis, line 1918: You can add `"star" to the `file-types` list here so that the extension is recognized. `injection-regex` is used for features like markdown codeblocks, so triple backticks and `star` would be a starlark codeblock with this change.

If you have `cargo` set up locally you can try the changes out with `cargo run file.star`

the-mikedavis (approved):

pascalkuthe (approved):

PR #7929, opened 2023-08-13
Feature/match inner whitespace

### LINKED ISSUE #4509: mis to match inner whitespace

bluthej:
Closes #4509.

I picked up where @Stonks3141 left off, refactored a little bit and fixed a corner case (used to panic because of integer underflow when the cursor was at `pos = slice.len_chars()`). I also added some unit tests for validation.

I used @Stonks3141's documentation proposal as is (just reapplied in the right spot).

If there is anything else I should do, please do tell! Any feedback is more than welcome :)


bluthej:
Hey, I just rebased on master, is there someone who would like to review this PR? Otherwise I guess I'm going to close it since it's been a while.

Issue #7404, opened 2023-06-20
Using a keypress to :quit fails to restore terminals cursor style


Joe-Zer0:
### Summary

Executing the command ":quit" to exit helix successfully restores my terminals cursor style (blinking bar).  But if I map quit to a key in the config 'q = ":quit"'.  If I press q to exit helix, then my terminal cursor becomes a block style cursor.

Environment Details:
Windows 10
Windows Terminal Version: 1.15.3465.0
PowerShell Version: 7.3.4

### Reproduction Steps

Add the following to the helix config:
[keys.normal]
q = ":quit"

Open any file with helix.
Press 'q' to exit helix.
See the cursor style is now a block.

### Helix log

I cleared the log and performed the above actions.  I was able to replicate the error, but no logs were created.

### Platform

Windows

### Terminal Emulator

Windows Terminal

### Helix Version

helix 23.03 (3cf03723)

Joe-Zer0:
Nice work, thanks!

PR #7931, opened 2023-08-13
Always reset cursor correctly on exit.

dilr:
This is intended to be a fix for #7404. I was able to replicate on MacOS by setting Terminal.app's default cursor to something other than Block, and then binding a key to `:quit` as described in the issue. Via debugging, I found that `Terminal::restore` actually does reset the cursor correctly regardless of if a keybind is used or if `:quit` is used.
However, the cursor gets set back to `CursorKind::Block` when the process exits if a keybinding was used.

It turns out that the `Drop` impl for `Terminal` will put the cursor back in block mode if `terminal.cursor_kind` is `Hidden`. Via debugging, I found that this was not the case if `:quit` was used but was the case if the keybind was used.

My understanding is that either `Terminal::restore` or `TerminalBackend::force_restore` will be called before exit, and both reset the cursor. Thus it seems to me that the `Drop` impl for `Terminal` is redundant and does not have the information needed to do the reset properly. Thus this PR removes the `Drop` impl.

### DIFF
--- a/helix-tui/src/terminal.rs
+++ b/helix-tui/src/terminal.rs
@@ -65,20 +65,6 @@ where
     viewport: Viewport,
 }
 
-impl<B> Drop for Terminal<B>
-where
-    B: Backend,
-{
-    fn drop(&mut self) {
-        // Attempt to restore the cursor state
-        if self.cursor_kind == CursorKind::Hidden {
-            if let Err(err) = self.show_cursor(CursorKind::Block) {
-                eprintln!("Failed to show the cursor: {}", err);
-            }
-        }
-    }
-}
-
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Good catch!

For context: we use the terminal's (block) cursor in the Prompt component used to enter typable commands, so when you use `:quit` you're exiting from the prompt with `Terminal.cursor_kind == Block`. With a keybind you can exit from normal mode where the cursor is hidden (done by default, so we can color the cursor), so the Drop calls `show_cursor`. `restore` and `force_restore` should always be called even if we panic so no need to reset the cursor in the Drop anyways.

PR #7940, opened 2023-08-14
feat: add commands to yank current buffer filepath

### LINKED ISSUE #2985: Add a command to copy the filename and cursor line number in the clipboard.

Regaez:
Closes #2985

## What

This PR adds a series of commands to enable the user to yank the current buffer's filepath to a specified register, or the clipboard. Per the referenced issue thread, there have been distinct commands created to allow the user to retrieve the following combinations:

- absolute path
- absolute path + line number
- absolute path + line and column numbers
- relative path
- relative path + line number
- relative path + line and column numbers

Each of the above combinations have been created as `Static` and `Typable` commands, so they can be mapped to a key by the user, or invoked with the `:` command mode. 

The general `yank_*_filepath` implementations will write to the current selected register, whereas the more specific `*_to_clipboard` commands will write to the system clipboard directly. Having both retains the flexibility for these commands to be used within Helix, while also enabling easier export of the values outside of Helix, i.e. some examples from issue thread:

- run `git blame` externally on current file
- share specific line/column reference with colleague, e.g. better leverage this functionality: 
```
hx some/file/path.ext:row:col
```

### Naming

I welcome any suggestions for improvements to the naming. I tried to keep it verbose enough to differentiate between each combination easily, yet concise enough to not be _too_ long... but naming is hard. 

As for `TypableCommand` aliases, I considered a few, but was not sufficiently convinced of their value to add any of them. Some ideas I had:

- `abs`: yank absolute filepath
- `rel`: yank relative filepath
- `line`: yank relative filepath with line
- `col`: yank relative filepath with line and column numbers
- `cpa`: yank absolute path to clipboard
- `yfa`: yank absolute filepath to clipboard
- etc

In most of these cases, it seemed easier to me to just start typing the full command and then `<tab>` to the desired command, than to remember all the shorthand variants.

### Primary clipboard

I noticed that the `yank` implementations have dedicated functions for yanking to the primary clipboard in addition to the system clipboard. As there were already quite a few command combinations being added, I skipped adding these `primary-` specific ones, mainly for 2 reasons:

- I believe the general `yank_*_filepath` static commands could be user-mapped to a key and used in combination with setting the `+` register in advance to yank them to the primary clipboard. Although, it doesn't appear possible to do that with a `TypedCommand`.
- I'm not entirely sure when this primary clipboard would be used/want to be used for this feature vs using the system clipboard...

However, I am also happy to add these additional `primary-` command combos, if desired.

### Keymappings

I considered adding some default keymappings (as suggested in the original issue thread) to make these commands more readily accessible to users. I decided against it for now as these can always be configured by each user per their needs/preference and I am not sure how strict the project is in general about extending the default keymapping.

For those curious/wanting this feature, I was thinking of adding something like this to the "space" category:

```
"." => { "Yank filepath to clipboard"
  "a" => yank_absolute_filepath_to_clipboard,
  "r" => yank_relative_filepath_to_clipboard,
  "l" => yank_relative_filepath_with_line_to_clipboard,
  "c" => yank_relative_filepath_with_line_column_to_clipboard,
},
```
I can add this to the PR if there's a general consensus that they would be good additions, but can also be done as a follow up PR.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -10,7 +10,7 @@ use tui::widgets::Row;
 pub use typed::*;
 
 use helix_core::{
-    char_idx_at_visual_offset, comment,
+    char_idx_at_visual_offset, comment, coords_at_pos,
     doc_formatter::TextFormat,
     encoding, find_first_non_whitespace_char, find_workspace, graphemes,
     history::UndoKind,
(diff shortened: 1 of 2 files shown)


pascalkuthe:
We already have the current file register `%` I don't think we need additional commands beyond that.

Regaez:
> We already have the current file register `%` I don't think we need additional commands beyond that.

I don't think the current file register `%` meets users' needs right now, as mentioned in #2985.

1. The `%` file register only allows you to read from it and to get this value into the system clipboard for use outside of Helix is quite cumbersome (unless there are some nice register shortcuts I'm not aware of?). 
2. The `%` register also only keeps a reference to the absolute path.

I believe these commands will add more user-friendly functionality, such as getting relative paths, with line and column information, and writing to the system clipboard, similar to that of other text editors, such as VSCode and Sublime.

gabydd:
Instead of having seperate commands for this maybe we want to have the seperate components here as expansions: https://github.com/helix-editor/helix/pull/6979 this would allow for composing them and using them in not only yanking contexts

Sammo98:
> > We already have the current file register `%` I don't think we need additional commands beyond that.
> 
> I disagree; I don't think the current file register `%` meets users' needs right now, as mentioned in #2985.
> 
> 1. The `%` file register only allows you to read from it and to get this value into the system clipboard for use outside of Helix is a bit cumbersome (unless there are some nice register shortcuts I'm not aware of? I'm doing: `"%p<space>yd` pasting it _into_ the document and having to delete it after yanking to system clipboard... feels yucky).
> 2. The `%` register also only keeps a reference to the absolute path.
> 
> I believe these commands will add more user-friendly functionality, such as:
> 
> 1. getting relative file paths,
> 2. enriching paths with line and column information,
> 3. easily writing to the system clipboard, similar to that of other text editors, such as VSCode and Sublime,
> 4. enabling shortcut menus via keymapping (such as extending the "space" command) to expose this functionality to the user via the UI

I would agree with this. Having functionality to get current buffer filepath into system clipboard would be of use.

Out of interest, is it possible to get this into a keybinding? `"%p<space>yd`

Regaez:
Closing this PR as it's clear that command expansions is the preferred solution.

PR #7955, opened 2023-08-15
fix: line numbers remain relative when helix loses focus

woojiq:
If `line number = relative` and a new window is opened in helix, lines inside unfocused windows will be `absolute`. This commit adds the same thing when helix becomes unfocused in a terminal emulator (or on screen in general).

I thought about introducing a new `set_focus(self, bool)` editor method, but saw that almost every `Editor` field was public so I decided to change it manually.

Small discussion: https://github.com/helix-editor/helix/discussions/7954

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -43,6 +43,8 @@ pub struct EditorView {
     pub(crate) last_insert: (commands::MappableCommand, Vec<InsertEvent>),
     pub(crate) completion: Option<Completion>,
     spinners: ProgressSpinners,
+    /// Tracks if the terminal window is focused by reaction to terminal focus events
+    terminal_focused: bool,
 }
 
 #[derive(Debug, Clone)]
@@ -71,6 +73,7 @@ impl EditorView {
             last_insert: (commands::MappableCommand::normal_mode, Vec::new()),
             completion: None,
             spinners: ProgressSpinners::default(),
+            terminal_focused: true,
         }
     }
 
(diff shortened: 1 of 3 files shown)


woojiq:
Yes, it overlaps almost completly. In fact, if I rebase this PR on top of the mentioned, it will be a one-line solution? Should I rebase or maybe ask the author of the mentioned pr to add it to his pr?

woojiq:
I changed the name to match the mentioned pr. This is a bit like crutch, but I don't want to change `is_focus` because it represents the focus among views, not focus in desktop/terminal. Therefore, I changed it only in gutters.

gabydd:
That looks good to me now👍

helix-term/src/ui/editor.rs
```diff
@@ -171,7 +174,7 @@ impl EditorView {
                 view,
                 view.area,
                 theme,
-                is_focused,
+                is_focused & self.terminal_focused,
```
tqwewe, line 177: The use of a single `&` here is a little confusing. Wouldn't it have been better to just use `&&` here for clarity?

woojiq: It's already merged, I'm not going to create a PR for this small change.
Anyway, it is not difficult to learn bitwise operators and improve your programming knowledge a little.

tqwewe: There's just no reason to add complexity by making this a bit-wise operation rather than a logical AND operation, especially in an open source project with this many moving pieces. I didn't expect a PR be made for this small change, was just pointing it out.

archseer: A bitwise operation on bool is no less clearer than a logical operation. I prefer the code as is.

Issue #3389, opened 2022-08-11
Clojure crashing the editor


HiiGHoVuTi:
### Summary

The bug is really simple, the helix-lsp-client seems to crash when editing clojure code.
At first the editing works and at some point it crashes and from then on, it is impossible to re-open the file without the editor crashing.

### Reproduction Steps

I'm in a default `lein` project.

I tried this:

1. `hx src/project/core.clj`
2. go around changing some code

I expected this to happen:

being able to use helix to edit clojure files

Instead, this happened:

after some time editing the file, the editor crashes. When trying to re-open the file, it crashes immediately.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
Caused by:
    channel closed)
2022-08-11T10:46:32.297 helix_term::application [ERROR] Language Server: Method not found client/registerCapability
2022-08-11T10:53:21.546 helix_lsp::client [WARN] language server failed to terminate gracefully - failed to parse: invalid type: map, expected unit
2022-08-11T10:54:14.728 helix_lsp::transport [ERROR] err: <- Other(failed to send a message to server

Caused by:
    channel closed)
2022-08-11T10:55:26.963 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-08-11T10:55:26.963 helix_lsp::transport [ERROR] err: <- StreamClosed
2022-08-11T10:56:30.952 helix_lsp::transport [ERROR] err: <- Other(failed to send a message to server

Caused by:
    channel closed)
2022-08-11T10:57:14.206 helix_lsp::transport [ERROR] err: <- Other(failed to send a message to server

Caused by:
    channel closed)
```

or sometimes

```
Caused by:
    channel closed)
2022-08-11T11:11:55.481 helix_view::editor [ERROR] Failed to initialize the LSP for `source.clojure` { cannot find binary path }
2022-08-11T11:12:00.294 helix_view::editor [ERROR] Failed to initialize the LSP for `source.clojure` { cannot find binary path }
```
(`clojure-lsp` is in my path and it's found sometimes)

on top of that, the instantaneous crash when I open the file doesn't produce any logs !
</details>

<details><summary> Rust stack trace </summary>

Here are the rust logs, since they were provided too:
```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', /build/81fl2hx743wzi1fr8zmak81ggyahvz9m-source/helix-term/src/application.rs:513:81
stack backtrace:
[...]

PR #7958, opened 2023-08-15
fix(helix-lsp): don't panic when lsp capabilities aren't initialised

rslabbert:
Some language servers don't 100% align with the spec and might start to send commands before the initialization has fully completed. In these cases, it's probably best for Helix not to panic entirely.

I'm not sure if this is the best version of this change (maybe logging a warning would be good), but it converts an outright panic into a no-op, similar to what a lot of the code does when the capability required isn't supported. Happy to work on this if needed.

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -838,7 +838,7 @@ impl Client {
         new_text: &Rope,
         changes: &ChangeSet,
     ) -> Option<impl Future<Output = Result<()>>> {
-        let capabilities = self.capabilities.get().unwrap();
+        let capabilities = self.capabilities.get()?;
 
         // Return early if the server does not support document sync.
         let sync_capabilities = match capabilities.text_document_sync {
@@ -893,7 +893,7 @@ impl Client {
         text_document: lsp::TextDocumentIdentifier,
         text: &Rope,
     ) -> Option<impl Future<Output = Result<()>>> {
-        let capabilities = self.capabilities.get().unwrap();
+        let capabilities = self.capabilities.get()?;
 
         let include_text = match &capabilities.text_document_sync {
             Some(lsp::TextDocumentSyncCapability::Options(lsp::TextDocumentSyncOptions {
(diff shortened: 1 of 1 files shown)


pascalkuthe (changes requested):
Do you have a reproduction case?

I am fairly certain that this is not the right fix. You are changing notifications and requests sent from the client to server here. It's ok to unwrap here because helix will ignore any LSP server until it's finished initializing (capabilities are set by that point).

Any code that somehow obtains a reference to the server regardless and sends a notification/requests to an uninitialized server is buggy/a programmer error and should lead to a panic.

There might be some edgecase where a notification leads to us sending a notification/request back to the server altough I can't really think of a case where this would occur (hence the request for reproduction case). In that case this requests likely need special treatment

rslabbert:
It's quite hard to reproduce reliably, but I've mainly encountered it in the context of clojure-lsp. A similar type of crash was noted on #3389. I can see about working on a reproduction case (maybe with a mock LSP).

My suspicion (given that the cases that I've run into seems to be primarily around starting to edit before the LSP has initialized), is that if the LSP start up times out somehow, another command that was waiting on the initialization notification will proceed. To your point, this is an unexpected state to be in and most likely a bug.

That being said, my logic for this change was to ensure that if something like this happens, the editor won't crash, especially given that the times I've run into it, I've probably been a minute or two into a larger edit. Because of that, I sometimes wait until the LSP is initialised before doing anything to prevent helix panicking.

So I can look for the root cause and a reproduction, but I still think having a non-panic approach to these situations would be good.

pascalkuthe:
https://github.com/helix-editor/helix/issues/3389 was different because the crash occurred in an *event handler* so in core that is run when the server sends a message. This is fundamentally different to the code you are changing here which is used to *send* a message yo the server.

#7467 should have caught the last cases of panics in the LSP message handles for unitalized servers. If you are still using 23.5 (or were using an older master commit when you last experiences thag) then that could already be fixed. 

In general we purposefully panic in codepaths which can onky be reached if invariants are broken since it makes bugs inifunitely easier to track down. Any code that sends a request to an unintalized server is buggy (and it's a very hard to actually do) and should cause a panic so it can be fixed properly.

rslabbert:
I missed #7467 when I was looking through the issues. I suspect that's what I was running into.

It's good to know about the use of unwrap as a convention here. Thanks for taking the time to review.

PR #7962, opened 2023-08-16
Make '<C-j>' work like '<enter>' in most cases.

### LINKED ISSUE #7076: Numeric keyboard is printing "j" when "enter" was pressed in pickers
### Summary

The numeric keyboard is printing the "j" character when "enter" key was pressed.

image

### Reproduction Steps

1 - Open helix
2 - press space + s or space + f to get a file or symbol picker
3 - select a symbol or file and press "enter" in numeric keyboard

expected behavior:
open the file or select the symbol and return to edition

happens:
the key "j" is printed instead.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

alacritty

### Helix Version

23.05

dilr:
Meant to partially fix #7076. This adds `<C-j>` as a substitute for `<enter>` in all non keybindable places.

The issue seems to be that crossterm expects the enter key to be presented to it as carriage return or a CSI function code when in raw mode (see [here](https://github.com/crossterm-rs/crossterm/blob/master/src/event/sys/unix/parse.rs#L99) for details). This seems like a reasonable assumption, as turning on raw mode turns ICRNL off. However, it seems like under some circumstances, some people's terminal drivers ends up sending newlines anyways.

I'll see if I can track down why this happens in my case, and submit a patch to crossterm if it is fixable. What is happening in my case may not be what is happening for other people though, and this patch should make helix usable even if crossterm fails to properly detect enter.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -36,7 +36,7 @@ use helix_view::{
     editor::{Action, CompleteAction},
     info::Info,
     input::KeyEvent,
-    keyboard::KeyCode,
+    keyboard::{KeyCode,KeyModifiers},
     tree,
     view::View,
     Document, DocumentId, Editor, ViewId,
@@ -1262,6 +1262,9 @@ fn find_char(cx: &mut Context, direction: Direction, inclusive: bool, extend: bo
             KeyEvent {
                 code: KeyCode::Enter,
                 ..
+            } | KeyEvent {
+                code: KeyCode::Char('j'),
+                modifiers: KeyModifiers::CONTROL
             } =>
             // TODO: this isn't quite correct when CRLF is involved.
             // This hack will work in most cases, since documents don't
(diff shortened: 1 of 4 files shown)


dilr:
Also, I did manage to get my terminal to send `<C-j>` in insert mode by switching the cursor style to "bar" (which doesn't translate to CursorStyle::Hidden, unlike the "block" style), and insert mode already seems to insert newlines when given `<C-j>`.

Thus, I think the only remaining issue is that binding `<enter>` to commands won't work for people who are affected by the crossterm bug. Fixing this would require that people who aren't affected by the bug to consider `<C-j>` and `<enter>` to be the same key, so I'll leave the key bindings alone.

the-mikedavis (changes requested):
This should be fixed upstream in crosssterm or in the offending terminal emulators instead of patched around here. This change disallows C-j for everyone even though it's a valid keybinding for terminals that support Kitty's enhanced keyboard protocol.

dilr:
Indeed, it should be patched upstream. And I do intend to work on that when I have more time. However, the changes I made are only for the sections of the editor who's keys cannot be rebound. Thus people can still bind separate command for `<C-j>` and `<enter>`. Also, none of the non-rebindable sections of the editor had anything bound to `<C-j>` before this PR, so no functionality is lost.

dilr:
Also, note that `<C-j>`, `<C-h>`, and `<C-i>` already do the same thing as `<enter>`, `<backspace>`, and `<tab>` in insert mode. Similarly, `<C-h>` and `<C-i>` do the same thing as `<backspace>` and `<tab>` in prompt mode. Thus people who know about the correspondence between control characters and keys might expect `<C-j>` to also work in prompt mode.

pascalkuthe:
I don't think we will be going forward with this so I am closing this. Thank you for contributing!

PR #7986, opened 2023-08-17
health - add formatter to display

### LINKED ISSUE #6992: Add `formatter` to `hx --health`
https://github.com/helix-editor/helix/blob/5ae30f1993180b2c043d13f83e16f8647f7ff129/helix-term/src/health.rs needs to be modified.

Sammo98:
First PR here, hopefully this should be able to close #6992 by adding formatter to both the `--health` and `--health <lang>` options.


Sammo98:
Ah I've just noticed there is another PR for this issue here https://github.com/helix-editor/helix/pull/5692

Although it seems to me that would only work for the `hx --health <lang>` option and not add the column for all langs.

Sammo98:
Some screenshots here to show the changes in effect:

Screenshot 2023-08-17 at 21 17 12
Screenshot 2023-08-17 at 21 16 51

helix-term/src/health.rs
```diff
@@ -279,6 +285,13 @@ pub fn language(lang_str: String) -> std::io::Result<()> {
         }),
     )?;
 
+    probe_protocol(
+        "formatter",
+        lang.formatter
+            .as_ref()
+            .map(|formatter| formatter.command.to_string()),
+    )?;
```
the-mikedavis, line 293: Let's move this after the debug adapter so the order matches the overall `--health` display

the-mikedavis (commented):
This looks good, I just have a minor nit about the order we print the health checks in `--health <lang>`

helix-term/src/health.rs
Sammo98: Thanks for reviewing @the-mikedavis , just pushed to move this as you have suggested.

```
Configured language server: pylsp
Binary for language server: /opt/homebrew/bin/pylsp
Configured formatter: black
Binary for formatter: /opt/homebrew/bin/black
Highlight queries: ✓
Textobject queries: ✓
```

PR #7988, opened 2023-08-18
Add different background color for inactive panes

### LINKED ISSUE #1187: Helix in unfocused window should appear differently than in a focused window

lefevau:
Adds a new theme key `ui.background.inactive` to control the background color of inactive panes.
This PR will not have any effect until the inactive color is defined, which I will leave to the theme maintainers.

Closes: https://github.com/helix-editor/helix/issues/1187

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -269,6 +269,7 @@ These scopes are used for theming the editor interface:
 | Key                               | Notes                                                                                          |
 | ---                               | ---                                                                                            |
 | `ui.background`                   |                                                                                                |
+| `ui.background.inactive`          | Background of unfocused document                                                               |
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):

freepicheep:
Just found this PR. I hope this gets merged!

Issue #7971, opened 2023-08-16
Helix Hard Crash (tokio-runtime-worker)


binarypie:
### Summary

Helix will hard crash on malformed LSP response.

### Reproduction Steps

 

I tried this:

1. `helix`
2. Open several files in your workspace
3. Open `.json` `.yaml` or `.ts` files
4. Make edits, switch files, and eventually helix will hard crash.

I expected this to happen:
To be able to edit files.

Instead, this happened:
helix hard crashes
```
❱ helix
thread 'tokio-runtime-worker' panicked at 'byte index 980 is out of bounds of ``', /home/abuild/rpmbuild/BUILD/helix-23.05/vendor/ropey/src/iter.rs:1371:25
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-08-16T09:20:59.760 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:21:28.175 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:21:40.703 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:23:37.254 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:24:12.952 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:36:41.880 helix_term::application [ERROR] received malformed notification from Language Server: Unhandled
2023-08-16T09:39:22.295 helix_view::editor [ERROR] Failed to initialize the LSP for `source.yaml` { cannot find binary path }
2023-08-16T09:39:45.704 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
2023-08-16T09:40:57.241 helix_term::application [ERROR] received malformed notification from Language Server: Unhandled
2023-08-16T09:43:50.712 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
2023-08-16T09:45:32.878 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
[...]

PR #7996, opened 2023-08-18
Update install.md to include packaging disclaimer

binarypie:
As part of https://github.com/helix-editor/helix/issues/7971 it was determined that a version of helix was compiled with a newer rust compiler version (which did not exist when 23.5 was released so the bug went unnoticed).

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -45,6 +45,7 @@ Download pre-built binaries from the
 line.
 
 ## Linux, macOS, Windows and OpenBSD packaging status
+Packages found outside the [GitHub Releases page](https://github.com/helix-editor/helix/releases) are outside the control of the Helix team. If you have an issue with a specific package, it may be necessary to contact the package maintainers.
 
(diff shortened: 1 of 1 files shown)


archseer:
This should be common sense though? Let's say the `git` package on Arch Linux etc. clearly isn't maintained by the git developers. These packages are still official, they're just official downstream packages (https://www.redhat.com/en/blog/what-open-source-upstream).

pascalkuthe:
Yeah I also think this is common sense and what I was implying in #7971. Packages are almost never maintained by upstream directly

binarypie:
> Yeah I also think this is common sense and what I was implying in #7971. Packages are almost never maintained by upstream directly

This is a pattern I've seen called out on several other open source applications with wide variety of downstream packages. Common sense or not it helps to over communicate responsibilities.

pascalkuthe:
I don't think it makes sense to add this to the docs. We can't document everything that would bloat the docs too much

PR #7997, opened 2023-08-18
Cursor’s position custom format support

### LINKED ISSUE #5049: Improve display of total-line-numbers in the statusline
Hi.
Currently, the display for current line and total line numbers in the statusline is separate.
I'd prefer if it could be combined.
For instance, to have something like this:
```
5/15:2

(current line)/(total line numbers):(column)
```
instead of:
```
5:2 15
```
I'm also open to other ideas, but I like having `(current line)/(total line numbers)`.
Thanks.

VincentBerthier:
# Objective

Fixes #5049
The goal of this PR is to allow users to use a custom format to display the cursor’s position in a buffer.

## Solution

The solution used was mostly similar to the one proposed in the issue by @austinletson, except for one variation:

if a user specifies a format, the format string is, by definition, dynamic which prevents the use of the format! macro and can lead to a small hit in performance.
To prevent that, if no custom format is specified, the position of the cursor will be presented exactly as it was previously.

When a custom format has been specified, then the first instances (again, to limit any performance hit) of the format markers {column}, {line} and {lines_count}
are replaced with the appropriate values.

---

## Changelog

- Statusline: added custom cursor position format support

--- 
(First PR, I hope I did everything right!)

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -81,21 +81,23 @@ left = ["mode", "spinner"]
 center = ["file-name"]
 right = ["diagnostics", "selections", "position", "file-encoding", "file-line-ending", "file-type"]
 separator = "│"
+position-format = "{column}:{line}/{lines_count}"
 mode.normal = "NORMAL"
 mode.insert = "INSERT"
 mode.select = "SELECT"
 ```
 The `[editor.statusline]` key takes the following sub-keys:
 
-| Key           | Description | Default |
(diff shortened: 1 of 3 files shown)


pascalkuthe:
This PR has gone stale, we are also not looking to add this level of custamizability to the statuline. Something like this would be covered by plugins or a sciptable config.

Thank you for contributing!

PR #8001, opened 2023-08-19
Aligned the recap for chapter 7.

qcerasmus:
while I do find it hilarious that the recap for the chapter regarding indentation is misaligned, I figured this could be a good first PR :)

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -845,7 +845,7 @@ lines.
  * Type < and > to indent / outdent lines.
 
  * Press Ctrl-a to increment the selected number.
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
I believe the spacing is intentional like in chapter 3's recap: https://github.com/helix-editor/helix/blob/e4c95f65a64a0553e1dafa365a40dcd5c328fa69/runtime/tutor#L428-L429

since the two keybinds are very closely related

qcerasmus:
Fair enough, I'll close the PR

PR #8003, opened 2023-08-19
Auto-format specified by language is now opt-in [#7868]

### LINKED ISSUE #7868: Provide an option to enable `auto_format` for individual languages, keeping global disabled
Right now the way it works is that you have a global `auto_format` to toggle whether it's enabled or not, and then you have each individual language server doing auto-format
on its own by default.

However when auto format is generally not desired and you want to opt-into it only for a small amount of languages this won't work, because global `auto-format = false` and then
having `auto-format = true` in `[[language]]` section won't do anything.

This is pretty much what the code does.

```rust
    let fmt = if editor_auto_fmt {
        doc.auto_format().map(|fmt| {
```

Given that other default is `true` it's tricky to get it right here, I guess the way it should be implemented is that the config value of each language is populated with the `Option<bool>` and then you `unwrap_or` to global default or global config is populated
during the deserialization(more complex, but you'll remove the need for option and the config global value wouldn't be used at all), that way it'll work automatically without nested bool checking complexity and will meat common expectations,
because right now what `auto-format` does in `[[language]]` is `opt-out`.

VincentBerthier:
# Objective

Fixes #7868
The auto-format parameter when specified by language could disable the auto-formatting when it was globally activated. However when the auto-formatting was disabled globally there was no way to override that parameter for an individual language.

## Solution

This PR changes that, such that if auto-format is turned off globally *but* turned on for the language, the auto-formatting will be done on save.

| Global | Language | Current Behaviour | New Behaviour |
| ------ | -------- | ----------------- | ------------- |
| True   | True     | True              | True          |
| True   | False    | False             | False         |
| False  | True     | **False**         | **True**      |
| False  | False    | False             | False         |

### Changed

- Auto-format: can now be opted-in per language when globally deactivated

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -333,8 +333,13 @@ fn write_impl(
     let jobs = &mut cx.jobs;
     let (view, doc) = current!(cx.editor);
     let path = path.map(AsRef::as_ref);
+    let language_auto_fmt = if let Some(language_config) = doc.language_config() {
+        language_config.auto_format
+    } else {
+        editor_auto_fmt
+    };
 
-    let fmt = if editor_auto_fmt {
+    let fmt = if language_auto_fmt {
         doc.auto_format().map(|fmt| {
             let callback = make_format_callback(
                 doc.id(),
@@ -692,7 +697,12 @@ pub fn write_all_impl(
                 current_view.id
             };
 
-            let fmt = if auto_format {
+            let language_auto_fmt = if let Some(language_config) = doc.language_config() {
+                language_config.auto_format
+            } else {
+                auto_format
+            };
+            let fmt = if language_auto_fmt {
                 doc.auto_format().map(|fmt| {
                     let callback = make_format_callback(
                         doc.id(),


dead10ck:
This is actually desired behavior; the global setting should take precedence over per language settings. Languages themselves have auto formatting on by default, e.g. Rust, and this would change behavior such that users that don't want auto formatting in Rust would be forced to create a custom `languages.toml`, in addition to the global setting, to "really" turn it off.

VincentBerthier:
Oh, that’s fine then.

But that means the issue should be closed, no?

VincentBerthier:
Oh right.

Well… Nevermind then. Although…the issue should still be closed?

dead10ck:
Oh sorry, I didn't see the issue. I'll comment in there and let other maintainers weigh in

VincentBerthier:
Though now that I think about it, it’s likely 'just' that my PR is lacking something:
if we make the 'global' language auto-format be false if the global editor auto-format is false, it should still work as intended, right?

A global editor auto-format to false would still disable auto-format, with the option to opt-in *via* language overrides.

Could that be a thing?

pascalkuthe:
Yeah I don't think we want to change the behavior here. As long as there is two options it can be either `global && local` or `globsl || local`. With some settings we work around this by using `Option<bool>`.

I am not really sure what you mean with your latest comment. You can already overwrite the default autoformat settings for individual languages in your languages.toml.

I agree with @dead10ck that the current behaviour is preferable since you usualy want to turn formatting off rather than on.

VincentBerthier:
What I meant was that it could be:
1) If auto-format is defined in a user language configuration, use that no matter what the value is
2) else use the editor global configuration as needed

But that’s fine. I found a way to make it work but the code was…meh. I’m closing this then!

dead10ck:
> Though now that I think about it, it’s likely 'just' that my PR is lacking something:
> if we make the 'global' language auto-format be false if the global editor auto-format is false, it should still work as intended, right?

That is the current behavior. Global and language settings have to be combined somehow to come out with one boolean to indicate whether or not to auto format. Right now they're logically combined with an `and`, i.e. global must be true and language must be true.

> A global editor auto-format to false would still disable auto-format, with the option to opt-in *via* language overrides.
> 
> Could that be a thing?

The problem is the language setting is true by default for some languages, like Rust. So with no other configuration, global && language == true. If a user doesn't like auto formatting, they only need to add one change to their global editor config to turn the feature off.

With the change in this PR, only the language setting must be on for the feature to be on. So if a user doesn't want auto formatting, they would have to turn it off in their own language settings file that overrides the shipped default.

Issue #7029, opened 2023-05-12
Highlighting is done for file named exactly a language file name extension (rs, py, …)


antoyo:
### Summary

Hi.
Helix will enable the Rust syntax highlighting for a file named `rs`, the Python one for `py`, etc.
The config should include a leading dot to avoid this.
Thanks.

### Reproduction Steps

I tried this:

1. `hx rs`
2. Type `fn`.

I expected this to happen:

No syntax highlighting.

Instead, this happened:

`fn` got highlighted because helix loads the Rust syntax highlighter.

### Helix log

n/a

### Platform

Linux

### Terminal Emulator

alacritty 0.12.0 (5a728195)

### Helix Version

helix 23.03

pascalkuthe:
This is caused by this check: https://github.com/helix-editor/helix/blame/06d7f5d100fdcc99f4cdfda879898b2d488d8d7c/helix-core/src/syntax.rs#L643.

For the most part file types are already treated as extensions. We have the `{ suffix = "..." }` syntax for matching the rest of the filename. I wonder if we should remove this check and simply convert all files that check the entire file-name (like `poetry.lock` for toml) to use suffix matching.


What do you think @the-mikedavis? I saw in the blame that you introduced that syntax.

the-mikedavis:
There is some more discussion of this in https://github.com/helix-editor/helix/issues/5869. Using suffix for full filenames could be a nice workaround to avoid introducing breaking changes or new file-type formats but it seems a little hacky to me: I would prefer having a `{ filename = "poetry.lock" }` over using suffix

pascalkuthe:
Ah I wasnt aware of the discussion there. From the issue title I wouldn't have looked there.

Using the `filename` syntax looks quite nice. If we want to avoid the breaking change an alternative could be to just add a swcond`{extension = "rs" }`  and keep the current syntax unchanged.

It's not a huge deal but can cause weird bugs when a language uses a common filename (like `res`). Especially if the TS grammar isn't super robust. So I would Kean towards trying to fix this somehow.

the-mikedavis:
This was fixed in https://github.com/helix-editor/helix/pull/8006

PR #8006, opened 2023-08-19
Add glob file type support

### LINKED ISSUE #7422: Ability to specify file type prefixes for language detection
A pattern I've seen or implemented in many repos is to have `Containerfile` alongside a `Containerfile.base` - the former bundles up the app code (which changes frequently), the latter installs "system level dependencies" which generally changes far less frequently (this would be where, say, `libfoo-devel` gets installed for `foo-gem` to build against).

My Neovim config appears to auto-detect `Containerfile.base` as a Dockerfile (not sure if that's by way of file name or by way of the `FROM blah` line at the top), but I can't so far figure out how to emulate this in Helix, and further, Vim modelines don't flip Helix into Dockerfile mode, so I can't even hint at "hey, here's what you should use to highlight this!".

I'd propose something like the following, though it's probably imperfect and open to all sorts of discussion:

```toml
file-types = ["Dockerfile", "dockerfile", "Containerfile", "containerfile", { prefix = "Containerfile." }]
```

or maybe `Containerfile.*`?

gjabell:
One of the issues I've run into with Helix (coming from nvim) is the ability to specify file types based on a full path pattern and not just a suffix. This is necessary for file types like YAML which might need different tooling based on their filesystem context (examples include GitHub Actions and Helm charts) or things like Dockerfile (a common paradigm is to add a suffix for different use cases, e.g. `Dockerfile.dev`).

It seems that this issue was already discussed in #2455; not sure if there are plans to introduce a different paradigm for file type matching which solves it without globs but I think it's a rather necessary feature either way. Other options would include:
- adding a separate prefix option => this makes the file-types even more complicated and also might end up with conflicting matches
- using regex instead of globbing => possibly a better solution (allows for more fine-grained matching), but it feels a bit unnecessary to me; globbing is much simpler and I think still covers all the cases where you'd want to match files based on path

I'm new to Rust and to Helix, so please let me know where improvements can be made or where I'm missing something :slightly_smiling_face: 

Fixes #7422?

### DIFF
--- a/helix-core/Cargo.toml
+++ b/helix-core/Cargo.toml
@@ -49,6 +49,7 @@ chrono = { version = "0.4", default-features = false, features = ["alloc", "std"
 
 etcetera = "0.8"
 textwrap = "0.16.0"
+globset = "0.4.14"
 
 nucleo.workspace = true
 parking_lot = "0.12"
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -82,12 +82,6 @@ pub struct Configuration {
     pub language_server: HashMap<String, LanguageServerConfiguration>,
 }
 
-impl Default for Configuration {
-    fn default() -> Self {
-        crate::config::default_syntax_loader()
-    }
-}
-
 // largely based on tree-sitter/cli/src/loader.rs
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(rename_all = "kebab-case", deny_unknown_fields)]
@@ -821,8 +815,10 @@ pub struct Loader {
     scopes: ArcSwap<Vec<String>>,
 }
 
+pub type LoaderError = globset::Error;
+
 impl Loader {
-    pub fn new(config: Configuration) -> Result<Self, globset::Error> {
+    pub fn new(config: Configuration) -> Result<Self, LoaderError> {
         let mut language_configs = Vec::new();
         let mut language_config_ids_by_extension = HashMap::new();
         let mut language_config_ids_by_shebang = HashMap::new();
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -164,9 +164,11 @@ pub enum FileType {
     /// The extension of the file, either the `Path::extension` or the full
     /// filename if the file does not have an extension.
     Extension(String),
-    /// The suffix of a file. This is compared to a given file's absolute
-    /// path, so it can be used to detect files based on their directories.
-    Suffix(String),
+    /// A Unix-style path glob. This is compared to the file's absolute path, so
+    /// it can be used to detect files based on their directories. If the glob
+    /// is not an absolute path and does not already start with a glob pattern,
+    /// a glob pattern will be prepended to it.
+    Glob(globset::Glob),
 }
 
 impl Serialize for FileType {
@@ -178,9 +180,9 @@ impl Serialize for FileType {
 
         match self {
             FileType::Extension(extension) => serializer.serialize_str(extension),
-            FileType::Suffix(suffix) => {
+            FileType::Glob(glob) => {
                 let mut map = serializer.serialize_map(Some(1))?;
-                map.serialize_entry("suffix", &suffix.replace(std::path::MAIN_SEPARATOR, "/"))?;
+                map.serialize_entry("glob", glob.glob())?;
                 map.end()
             }
         }
@@ -213,9 +215,20 @@ impl<'de> Deserialize<'de> for FileType {
                 M: serde::de::MapAccess<'de>,
             {
                 match map.next_entry::<String, String>()? {
-                    Some((key, suffix)) if key == "suffix" => Ok(FileType::Suffix({
-                        suffix.replace('/', std::path::MAIN_SEPARATOR_STR)
-                    })),
+                    Some((key, mut glob)) if key == "glob" => {
+                        // If the glob isn't an absolute path or already starts
+                        // with a glob pattern, add a leading glob so we
+                        // properly match relative paths.
+                        if !glob.starts_with('/') && !glob.starts_with("*/") {
+                            glob.insert_str(0, "*/");
+                        }
+
+                        globset::Glob::new(glob.as_str())
+                            .map(FileType::Glob)
+                            .map_err(|err| {
+                                serde::de::Error::custom(format!("invalid `glob` pattern: {}", err))
+                            })
+                    }
                     Some((key, _value)) => Err(serde::de::Error::custom(format!(
                         "unknown key in `file-types` list: {}",
                         key
@@ -759,7 +772,7 @@ pub struct Loader {
     // highlight_names ?
     language_configs: Vec<Arc<LanguageConfiguration>>,
     language_config_ids_by_extension: HashMap<String, usize>, // Vec<usize>
-    language_config_ids_by_suffix: HashMap<String, usize>,
+    language_config_ids_by_glob: HashMap<globset::Glob, usize>,
     language_config_ids_by_shebang: HashMap<String, usize>,
 
     language_server_configs: HashMap<String, LanguageServerConfiguration>,
@@ -773,7 +786,7 @@ impl Loader {
             language_configs: Vec::new(),
             language_server_configs: config.language_server,
             language_config_ids_by_extension: HashMap::new(),
-            language_config_ids_by_suffix: HashMap::new(),
+            language_config_ids_by_glob: HashMap::new(),
             language_config_ids_by_shebang: HashMap::new(),
             scopes: ArcSwap::from_pointee(Vec::new()),
         };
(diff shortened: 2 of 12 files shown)


pascalkuthe:
We can avoid adding a new deoendy by using globset (which we already depend on)

pascalkuthe:
I still want to solve https://github.com/helix-editor/helix/issues/7029. I think if we do go ahead with this then that could be a good opportunity for fixing that aswell. `{ glob = "foo" }` already behaves the same as the proposed filename matcher there and can be changed to behave the same as suffix by using `{ glob = "*foo" }` so we could get rid of that special cases.

Technically globs can also match file extensions so we  could theoretically replace all matching modes with globs for maximum consistency but that's sadly too much of a breaking change. Maybe one day when we migrate to script based config

pascalkuthe:
Yeah that would be too large of a breaking change. Breaking the suffix config is fine IMO, since it's pretty rarely used but for file extensions it would break a ton of user config.

languages.toml
```diff
@@ -2775,7 +2775,7 @@ source = { git = "https://github.com/kylegoetz/tree-sitter-unison", rev = "98c4e
 [[language]]
 name = "todotxt"
 scope = "text.todotxt"
-file-types = [{ suffix = ".todo.txt" }, "todotxt"]
+file-types = [{ glob = ".todo.txt" }, "todotxt"]
```
pascalkuthe, line 2778: I am not sure this is quite the same thing i believe the indention was to allow `*.todo.txt` although I never use that format so I am not sure

helix-core/src/syntax.rs
```diff
@@ -797,20 +807,20 @@ impl Loader {
             .and_then(|n| n.to_str())
             .and_then(|file_name| self.language_config_ids_by_extension.get(file_name))
```
pascalkuthe, line 808: I believe we should remove this special case and handle all existing cases that use this as a glob

pascalkuthe: no those are actual fie extensions. As you can see here this  lookup doesn't look at the file extension but the entires filename instead. So this matches `../rs` but not `../foo.rs`. The file extension check is done later. The check here if for historical reasons to also support specifying full filnames like `Cargo.toml`. This hack has the unfortunate consequence of common file extensions matching the full filename. For example so weird scripting language uses `.res` as a file extension and I commonly use temporary files named just `res`  which  then have the wrong filetype.

So to fix this we should remove this hack (so that a plain string really only matches file extensions) and full filnames should use a glob pattern. So for example `file-types = ["toml", "poetry.lock", "Cargo.lock"]` would become `file-types = ["toml", { glob = "poetry.lock" }, { glob = "Cargo.lock" }]`

languages.toml
David-Else: @Jan9103 As you added todotxt, what do you think?

helix-core/src/syntax.rs
```diff
@@ -792,25 +802,20 @@ impl Loader {
...
         // or a suffix of the file name.
-        let configuration_id = path
-            .file_name()
-            .and_then(|n| n.to_str())
-            .and_then(|file_name| self.language_config_ids_by_extension.get(file_name))
+        let configuration_id = self
+            .language_config_ids_by_glob
+            .iter()
+            .find_map(|(glob, id)| {
+                if glob.compile_matcher().is_match(path) {
```
pascalkuthe, line 809: This is fairly inefficient, I the very least we should precompile the glob here but even if we do that matching every glob in sequence is fairly slow.
 
Instead you cab jse a GlobSet. Matching a globset is not much slow than matching a single Glob but matches all globs at once. You can use  https://docs.rs/globset/latest/globset/struct.GlobSet.html#method.matches)

languages.toml
pascalkuthe: Thanks for filling jn the details. It should be `[{ glob = "todo.txt" },  { glob = "*.todo.txt" }, "todotxt", "todo.txt"]` with the syntax from this PR. Plaintext now only matches file extensions with this PR, so you need a glob for the " only filename case". By default globs match relatively paths but if your glib starts with `/` or `*` it matches an absolute path instead

helix-core/src/syntax.rs
pascalkuthe: I think using the longest match makes sense as a tiebreaker. For example, if we switch to using globs for filextensions in the future then `*.txt` and `*.todo.txt` (example from above) would both match the same baths but `.todo.txt` should be used because its longer (its matches are more specific). I can imagine something similar for other paths. I can't really think of a way where would would want to use the more general pattern over the more specific one.

David-Else:
@gjabell I don't think my Rust is good enough for me to review your code, but thanks for asking :)

pascalkuthe (dismissed):
I am gonna take this for a spin soonish but looking trough the source code this lgtm

pascalkuthe:
The label was an accident/just forgot about it.

we are currently in a soft feature freeze period leading up to the next release so this won't be merged until the release happens.

eliliam:
Could we get some authorized reviewers to check this out? 

If this is still a ways out then I can think of another way to accomplish this, very similar to how vim does it. This is by including a comment snippet at the top or bottom of a file defining what language the file is, which is especially useful in cases where there is no file extension, especially when there can be different files of the same filename but have different language types(Think of `config` or `config.conf`, where the `.conf` extension is rather useless). 

An example of how vim does it is via a comment such as the one below:
``` 
# vi: ft=sshconfig
```

This is particularly useful for me since I have multiple SSH configs aside from the main `.ssh/config` file. Does anyone know of a way to do something like this?

helix-core/src/syntax.rs
```diff
@@ -752,65 +803,54 @@ pub struct Loader {
...
-    pub fn new(config: Configuration) -> Self {
-        let mut loader = Self {
-            language_configs: Vec::new(),
-            language_server_configs: config.language_server,
-            language_config_ids_by_extension: HashMap::new(),
-            language_config_ids_by_suffix: HashMap::new(),
-            language_config_ids_by_shebang: HashMap::new(),
-            scopes: ArcSwap::from_pointee(Vec::new()),
-        };
+    pub fn new(config: Configuration) -> Result<Self, globset::Error> {
```
the-mikedavis, line 806: Switching this to return a Result makes the error handling for config (re)loading a little bit messier. I think it's revealing an odd separation between `Configuration` and `Loader` that makes error handling awkward, and it might be a nice chance for a refactor.

I pushed up some ideas in this commit: https://github.com/the-mikedavis/helix/commit/6c455ec4c4b250c8170d632c6a3436b51c0e8f85

The main thing I want to avoid is handling the `Loader::new` error in `Application::new`. If either the Configuration and Loader creation fails we should report on it in this part of `main`: https://github.com/helix-editor/helix/blob/bbcc89241fd63b460ea7b1661a550f5b9eacba4c/helix-term/src/main.rs#L148-L155

To do that we basically handle `Loader` everywhere other than `helix_core` (or the health-check part of `xtask`) rather than `Configuration`, which I think is cleaner anyways.

the-mikedavis (commented):
This looks good, I just have a concern about the new Result making error handling a little messier when we're (re)loading configuration.

(Sorry for the delay in review, I've been meaning to give this a proper look for a while!)

Agilan989:
Still some one hacked to used and login to my account same devices

Thank you
Best regards
Agilanberan

Agilan989:
Please do some thing .. login to iPhone 11 s same device was used for
modified version and it

Thank you
Best regards
Agilanberan

helix-core/Cargo.toml
```diff
@@ -49,6 +49,7 @@ chrono = { version = "0.4", default-features = false, features = ["alloc", "std"
 
 etcetera = "0.8"
 textwrap = "0.16.0"
+globset = "0.4.13"
```
the-mikedavis, line 52: I believe this was bumped to `0.4.14` in `helix-lsp` by dependabot a while ago, might as well keep this and the helix-lsp dep in-sync

```suggestion
globset = "0.4.14"
```

languages.toml
```diff
@@ -2045,7 +2045,7 @@ source = { git = "https://github.com/sogaiu/tree-sitter-clojure", rev = "e57c569
 name = "starlark"
 scope = "source.starlark"
 injection-regex = "(starlark|bzl|bazel)"
-file-types = ["bzl", "bazel", "BUILD", "star"]
+file-types = ["bzl", "bazel", "star", { glob = "BUILD" }]
```
the-mikedavis, line 2048:
```suggestion
file-types = ["bzl", "bazel", "star", { glob = "BUILD" }, { glob = "BUILD.*" }]
```

globs work well for bazel since there is a convention of `BUILD.<app>` filenames

the-mikedavis (commented):
I took a look through the languages.toml and I think everything that is a full filename has been converted to a `{ glob = "<filename>" }` :+1:

I just have a suggestion for starlark file-types but otherwise this looks good to me

the-mikedavis (approved):
Looks great, thanks for working on this!

pascalkuthe (approved):
great to see this finished! I think moving fully to globs for all ft detection in the future would be great so we don't have to keep different ft variants around (potentially once we switch away from toml config)

PR #8009, opened 2023-08-20
[chore] Remove unnecessary `Err` from `get_canonicalized_path`

nkitsaini:
Following function never fails but returns a `Result`.
https://github.com/helix-editor/helix/blob/01a1e5ec2a8c5b81da1360746dad0a96cd88ae8e/helix-core/src/path.rs#L88-L97

This change just fixes this and all the callers of this function.

History:
Previously the function returned a `Result` as it had to rely on `std::env::current_dir` which could fail. But in https://github.com/helix-editor/helix/commit/8afc0282f28e73cf78d1bd7b11d78fd853ae2036 we started relying on `helix_loader::current_working_dir` which consumes the `Err` already.

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -51,12 +51,12 @@ pub enum PathOrId {
 }
 
 impl PathOrId {
-    fn get_canonicalized(self) -> std::io::Result<Self> {
+    fn get_canonicalized(self) -> Self {
         use PathOrId::*;
-        Ok(match self {
-            Path(path) => Path(helix_core::path::get_canonicalized_path(&path)?),
+        match self {
+            Path(path) => Path(helix_core::path::get_canonicalized_path(&path)),
             Id(id) => Id(id),
-        })
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


pascalkuthe:
I think this js ok, we likely don't want path normalization to be fallible. Even if we do introduce failure we likely want to transparently handle tat since canonicalization is just too fundementsl to fail

pascalkuthe (approved):

helix-term/src/ui/picker.rs
```diff
@@ -375,7 +375,7 @@ impl<T: Item + 'static> Picker<T> {
     fn current_file(&self, editor: &Editor) -> Option<FileLocation> {
         self.selection()
             .and_then(|current| (self.file_fn.as_ref()?)(editor, current))
-            .and_then(|(path_or_id, line)| path_or_id.get_canonicalized().ok().zip(Some(line)))
+            .and_then(|(path_or_id, line)| Some(path_or_id.get_canonicalized()).zip(Some(line)))
```
the-mikedavis, line 378:
```suggestion
            .map(|(path_or_id, line)| (path_or_id.get_canonicalized(), line))
```

the-mikedavis (approved):

pascalkuthe (approved):

PR #8010, opened 2023-08-20
Remove path completions for `:new` command

nkitsaini:
Since the argument of `:new` is never used, the completion might be confusing for user.


pascalkuthe:
The current behaviour is intentional, there was a previous PR (can't find it tight now) with the same change. `:new` is intended for creating a new scratch buffer. If you want to open a new buffer with a path use `:open`

nkitsaini:
@pascalkuthe That makes sense. In that case I'll revert this change but also remove the autocompletion that we get for `:new`.

pascalkuthe (approved):

the-mikedavis (approved):

PR #8021, opened 2023-08-20
Add an event system

pascalkuthe:
Closes #6910
Closes #5803
Closes #2027
Closes #2581 (moving in insert mode will always close the completion popup now)
Closes #5906 (signature help will be requested regularly on edits and movement like suggested in the LSP standard and therefore closes reliably)
Closes #5962 (I think I am having trouble reproducing but i am pretty aggressive about closing the popups now)
Closes #7497 

Also helps with #6248 (now the buggy completion won't be shown automatically although you can still get them manually) 

This PR introduces an event/hook system to helix with the goal of decoupling logic from the core ui and editing code. This PR reimplements the completion and signature help features using that event system to showoff how the system works in practice.

I specifically choose these two subsystems since I have mentally accumulated a large list of issues with these systems that would both be fixed by treating them as an async debounced event handler. So the refactor here was necessary anyway but thanks to the event system this doesn't spread a ton of extra code everywhere.

I noticed that there are actually many more opportunities where the event system could help to move subsystem data out of the core editor code. For example:

* diff gutter could be moved to `DocumentDidChange`
* diff gutter could be turned into an async hook
* lsp document change notification could be moved to `DocumentDidChange`
* LSP inlay hint requests (fixing many of the issues described in the original PR around better timeouts)
* both diff gutter and lsp subscribe to `:reload` (and lsp also to `:write`) that can also be handled by the event system

I refrained from implementing all of these to make this already large PR completely unmanageable. Once this PR lands I will send followup PRs to update systems where it makes sense (again with the primary focus on systems that see functional improvements such as inlay hints).

The event system also opens up the possibility to subscribe to events by name. This could allow us to run shell scripts on save (for example) and is also an important step towards a plugin system IMO. While the basic infrastructure exists it's not hooked up to a config option currently to keep this PR manageable.

~Note that this PR currently includes #7814 to avoid introducing conflicts with the code there so the diff looks inflated.~

~The completion timeout is currently set to a very low value. This is intentional for testing (since lower values are more interesting for triggering edgecases) but will be raised before merge likely back up to 400ms. Note that the idle timeout-based completion timeout is currently broken on master anyway (always 33ms timeout) due to a regression so comparisons may seem off I am working on a fix for that too (since some stuff still uses the idle timeout) but that's orthogonal~

PR #8023, opened 2023-08-20
create separate timer for redraw requests

pascalkuthe:
The redraw handle was using the idle timeout to denounce redraw requests so far. I will openly admit that I didn't quite appreciate the importance of the idle timeout back when I wrote the diff gutter PR (as I was still new to the codebase). It didn't really matter back then because the diff gutter timing out is a very rare occurrence but with #7538 the redraw handle is now triggered quite commonly which essentially makes the idle-timeout setting useless.

This PR fixes that by simply adding a separate timer for the redraw handle. In the long term we may move away from using the idle-timeout for debouncing data model update (#8021) at which point it would be fine to switch back to the idle timeout. Until all of that (substantial) work is finished I think its a good idea to fix this in the meantime (especially looking toward the next release).

### DIFF
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -918,7 +918,7 @@ pub struct Editor {
     pub auto_pairs: Option<AutoPairs>,
 
     pub idle_timer: Pin<Box<Sleep>>,
-    pub redraw_timer: Pin<Box<Sleep>>,
+    redraw_timer: Pin<Box<Sleep>>,
     last_motion: Option<Motion>,
     pub last_completion: Option<CompleteAction>,
 
--- a/helix-view/src/editor.rs
+++ b/helix-view/src/editor.rs
@@ -918,6 +918,7 @@ pub struct Editor {
     pub auto_pairs: Option<AutoPairs>,
 
     pub idle_timer: Pin<Box<Sleep>>,
+    pub redraw_timer: Pin<Box<Sleep>>,
     last_motion: Option<Motion>,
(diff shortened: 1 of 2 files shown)


pascalkuthe:
(Added to next because this is a regression)

helix-view/src/editor.rs
```diff
@@ -918,6 +918,7 @@ pub struct Editor {
     pub auto_pairs: Option<AutoPairs>,
 
     pub idle_timer: Pin<Box<Sleep>>,
+    pub redraw_timer: Pin<Box<Sleep>>,
```
the-mikedavis, line 921:
```suggestion
    redraw_timer: Pin<Box<Sleep>>,
```

I think this one can be restricted to just this module, right? We wouldn't want anything else modifying the timer since it might mess up the debounce

helix-view/src/editor.rs
```diff
@@ -1753,12 +1756,16 @@ impl Editor {
                     if  !self.needs_redraw{
                         self.needs_redraw = true;
                         let timeout = Instant::now() + Duration::from_millis(33);
-                        if timeout < self.idle_timer.deadline(){
-                            self.idle_timer.as_mut().reset(timeout)
+                        if timeout < self.idle_timer.deadline() && timeout < self.redraw_timer.deadline(){
```
the-mikedavis, line 1759: Do we still need to check the idle_timer here or can we switch this to only look at the redraw_timer?

the-mikedavis (approved):

PR #8047, opened 2023-08-23
Add scrolling to completion doc popup

1nwf:
allows scrolling in the documentation popup from the completion menu. Previously, the extra text would be just cutoff.

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -94,6 +94,8 @@ pub struct CompletionItem {
 /// Wraps a Menu.
 pub struct Completion {
     popup: Popup<Menu<CompletionItem>>,
+    doc_popup: Option<Popup<Markdown>>,
+    rerender_doc_popup: bool,
     start_offset: usize,
     #[allow(dead_code)]
     trigger_offset: usize,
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Duplicate of https://github.com/helix-editor/helix/pull/4364

1nwf:
> Duplicate of #4364

I was not aware of that. thanks for pointing it out. However, that PR is a almost a year old and seems like the author abandoned it.

kirawi:
I'm not questioning your solution, but what are you trying to achieve with `allow_modify_viewport`?

1nwf:
> I'm not questioning your solution, but what are you trying to achieve with `allow_modify_viewport`?

Feel free to give feedback `allow_modify_viewport` set to true maintains the original behavior of how the Popup size and position are calculated on render. By default, this will modify those values based on cursor position, and other things such as min_width and max_width of the popup.  If it is set to false, it will not modify them and strictly use the viewport that is passed in.

andradei:
Would love to see progress on this.

bfedie518:
Still no updates on this? It would be a surprisingly wonderful fix.

PR #8048, opened 2023-08-23
Add config for diagnostics messages

twharmon:
I think we need a way to make the diagnostics display configurable. By default, the messages are right on the edge of the screen and can be hard to read when the messages get too long.

Here are the defaults (current behavior):
```toml
[editor.diagnostics]
top-margin = 1
right-margin = 0
width = 100
height = 15
```

Which will look something like this:
image

Here is how it can be adjusted:
```toml
[editor.diagnostics]
top-margin = 1
right-margin = 2
width = 100
height = 15
```

which will look like this:
image

Further possibilities could be an `align` property to align left if desired, or perhaps a way to truncate each of the messages to a max of N characters for extremely long messages that would be intrusive.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -29,7 +29,6 @@ use helix_view::{
     keyboard::{KeyCode, KeyModifiers},
     Document, Editor, Theme, View,
 };
-// use helix_tui::layout::Alignment;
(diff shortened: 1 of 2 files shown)


David-Else:
This looks cool, but will be made redundant by https://github.com/helix-editor/helix/pull/6417

PR #8049, opened 2023-08-24
Allow new_file command to optionally take a path 

cbr9:
I usually run into the following problem: create a new scratch buffer with `:n`, then save it somewhere with `:w`. I would like to be able to pass the path directly to `:n`, to save one step.

This PR does just this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,9 +81,9 @@ dependencies = [
 
 [[package]]
 name = "anyhow"
-version = "1.0.72"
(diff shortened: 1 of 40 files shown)


pascalkuthe:
You can use `:open <path>` for creating a new buffer with a path (even if that oath doesn't exist), `:new` is intended for creating a scratch buffer. We have rejected similar PRs a couple times now.

cbr9:
My bad. Sorry.

PR #8055, opened 2023-08-24
Allow indent-guides to be enable in languages.toml

### LINKED ISSUE #8054: Allow indent-guides to be enable in `languages.toml`
At the moment trying to set  `editor.indent-guides.render` to `true` but only for specific languages.  yaml and maybe python in my case.  Looking at the docs, and the helix src, I can't see a way for me to reach `indent-guides.render` in a `[[languages]]` table group. 

it would be great if it was possible to enable this at the language level instead of just globally so that I could enable indent guides only for languages where indentation is significant

twharmon:
Closes #8054

Allow configuring indent guides at the language level to override the global setting:
```toml
[[language]]
name = "python"
indent-guides = { render = true, character = "▏", skip-levels = 1 }
```

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -117,6 +117,9 @@ pub struct LanguageConfiguration {
 
     pub grammar: Option<String>, // tree-sitter grammar name, defaults to language_id
 
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub indent_guides: Option<IndentGuidesConfig>, // override global indent_guides
+
     // content_regex
     #[serde(default, skip_serializing, deserialize_with = "deserialize_regex")]
     pub injection_regex: Option<Regex>,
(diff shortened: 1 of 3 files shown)


pascalkuthe:
There is a bit of scope creep with config options right now. There is a large (and ever-growing) set of feature requests to configure XYZ per language/per buffer. I think a more generic approach would be better here instead. Enabling indent guides per language feels pretty niche to me. It does seem like something that makes some sense in some situations but it feels too niche to have a dedicated codepath. 

IMO we should allow overwriting arbitrary config options by language to solve this generally (and in the future, I think we likely want to add the concept of a per buffer config too (which the language config would initialize but can pe overwritten afterward)

PaulOlteanu:
Hey @pascalkuthe is there an issue or discussion for the larger scale config changes you'd like to see? I could take a look at changing the way configuration is handled if there isn't already an effort underway

pascalkuthe:
Let's close this on favor of the more general config rework derived above, see #8853

PR #8056, opened 2023-08-25
allow configuring diff gutter source

doy:
following up from #7805 

this implementation removes the diff provider cascading fallback behavior in favor of allowing the user to configure which diff provider to use (currently just `vcs`, `file`, or `none`). this should allow maintaining the current behavior in the normal case, but also allow using `:set diff-source file` to see a diff against the state of the file on disk.

this does not yet support per-buffer configuration of diff sources because helix doesn't really have a good mechanism for per-buffer configuration in general at the moment - it feels like something that would be easier to add once the more general feature exists.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -13,7 +13,7 @@ use helix_lsp::{
 use helix_view::{
     align_view,
     document::DocumentSavedEventResult,
-    editor::{ConfigEvent, EditorEvent},
+    editor::{ConfigEvent, DiffSource, EditorEvent},
     graphics::Rect,
     theme,
     tree::Layout,
@@ -362,11 +362,15 @@ impl Application {
             // the Application can apply it.
             ConfigEvent::Update(editor_config) => {
                 let mut app_config = (*self.config.load().clone()).clone();
+                let update_diff_base = app_config.editor.diff_source != editor_config.diff_source;
                 app_config.editor = *editor_config;
                 if let Err(err) = self.terminal.reconfigure(app_config.editor.clone().into()) {
                     self.editor.set_error(err.to_string());
                 };
                 self.config.store(Arc::new(app_config));
+                if update_diff_base {
+                    self.editor.update_diff_base();
+                }
             }
         }
 
@@ -566,6 +570,14 @@ impl Application {
             let id = doc.id();
             doc.detect_language(loader);
             self.editor.refresh_language_servers(id);
+        }
+
+        let diff_source = self.editor.config().diff_source;
+        let doc = doc_mut!(self.editor, &doc_save_event.doc_id);
+        if diff_source == DiffSource::File {
+            if let Some(path) = doc.path().cloned() {
+                doc.update_diff_base(&path, &self.editor.diff_provider, diff_source);
+            }
         }
 
         // TODO: fix being overwritten by lsp
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1283,10 +1283,12 @@ fn reload(
     }
 
     let scrolloff = cx.editor.config().scrolloff;
+    let diff_source = cx.editor.config().diff_source;
     let (view, doc) = current!(cx.editor);
-    doc.reload(view, &cx.editor.diff_provider).map(|_| {
-        view.ensure_cursor_in_view(doc, scrolloff);
-    })?;
+    doc.reload(view, &cx.editor.diff_provider, diff_source)
+        .map(|_| {
+            view.ensure_cursor_in_view(doc, scrolloff);
+        })?;
     if let Some(path) = doc.path() {
         cx.editor
             .language_servers
@@ -1324,6 +1326,7 @@ fn reload_all(
         .collect();
 
     for (doc_id, view_ids) in docs_view_ids {
+        let diff_source = cx.editor.config().diff_source;
         let doc = doc_mut!(cx.editor, &doc_id);
 
         // Every doc is guaranteed to have at least 1 view at this point.
@@ -1332,7 +1335,7 @@ fn reload_all(
         // Ensure that the view is synced with the document's history.
         view.sync_changes(doc);
 
-        doc.reload(view, &cx.editor.diff_provider)?;
+        doc.reload(view, &cx.editor.diff_provider, diff_source)?;
         if let Some(path) = doc.path() {
             cx.editor
                 .language_servers
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1284,7 +1284,7 @@ fn reload(
 
     let scrolloff = cx.editor.config().scrolloff;
     let (view, doc) = current!(cx.editor);
-    doc.reload(view, &cx.editor.diff_providers).map(|_| {
+    doc.reload(view, &cx.editor.diff_provider).map(|_| {
         view.ensure_cursor_in_view(doc, scrolloff);
     })?;
     if let Some(path) = doc.path() {
@@ -1332,7 +1332,7 @@ fn reload_all(
         // Ensure that the view is synced with the document's history.
         view.sync_changes(doc);
 
-        doc.reload(view, &cx.editor.diff_providers)?;
+        doc.reload(view, &cx.editor.diff_provider)?;
         if let Some(path) = doc.path() {
             cx.editor
                 .language_servers
--- a/helix-vcs/src/git.rs
+++ b/helix-vcs/src/git.rs
@@ -7,8 +7,6 @@ use gix::objs::tree::EntryMode;
 use gix::sec::trust::DefaultForLevel;
 use gix::{Commit, ObjectId, Repository, ThreadSafeRepository};
 
-use crate::DiffProvider;
-
 #[cfg(test)]
 mod test;
 
@@ -59,10 +57,8 @@ impl Git {
 
         Ok(res)
     }
-}
 
-impl DiffProvider for Git {
-    fn get_diff_base(&self, file: &Path) -> Result<Vec<u8>> {
+    pub fn get_diff_base(&self, file: &Path) -> Result<Vec<u8>> {
         debug_assert!(!file.exists() || file.is_file());
         debug_assert!(file.is_absolute());
 
@@ -101,7 +97,7 @@ impl DiffProvider for Git {
         Ok(data)
     }
 
-    fn get_current_head_name(&self, file: &Path) -> Result<Arc<ArcSwap<Box<str>>>> {
+    pub fn get_current_head_name(&self, file: &Path) -> Result<Arc<ArcSwap<Box<str>>>> {
         debug_assert!(!file.exists() || file.is_file());
         debug_assert!(file.is_absolute());
         let repo_dir = file.parent().context("file has no parent directory")?;
--- a/helix-vcs/src/lib.rs
+++ b/helix-vcs/src/lib.rs
@@ -14,65 +14,14 @@ mod diff;
 
 pub use diff::{DiffHandle, Hunk};
 
-pub trait DiffProvider {
-    /// Returns the data that a diff should be computed against
-    /// if this provider is used.
-    /// The data is returned as raw byte without any decoding or encoding performed
-    /// to ensure all file encodings are handled correctly.
-    fn get_diff_base(&self, file: &Path) -> Result<Vec<u8>>;
-    fn get_current_head_name(&self, file: &Path) -> Result<Arc<ArcSwap<Box<str>>>>;
-}
-
 #[doc(hidden)]
 pub struct Dummy;
-impl DiffProvider for Dummy {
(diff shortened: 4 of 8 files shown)


helix-vcs/src/git.rs
```diff
@@ -61,24 +61,46 @@ impl Git {
...
 }
 
-impl DiffProvider for Git {
-    fn get_diff_base(&self, file: &Path) -> Result<Vec<u8>> {
+impl VcsProvider for Git {
+    fn root_for_file(&self, file: &Path) -> Result<Box<dyn VcsRoot>> {
         debug_assert!(!file.exists() || file.is_file());
         debug_assert!(file.is_absolute());
-
-        // TODO cache repository lookup
```
pascalkuthe, line 69: This should be addressed in a sperate PR. I prefer a different interface (this kind of disjoint dynamic dispatch based approach will not work once we add filewatchjng to the mix). Repo caching is also sufficiently tricky and evolves resource based eviction and should just be it's ownPR (which also makes this one easier to review).

helix-term/src/commands/typed.rs
```diff
@@ -1317,7 +1322,12 @@ fn reload_all(
         view.sync_changes(doc);
 
         let redraw_handle = cx.editor.redraw_handle.clone();
-        doc.reload(view, &cx.editor.diff_providers, redraw_handle)?;
+        doc.reload(
+            view,
+            diff_provider!(cx.editor),
+            &cx.editor.vcs_providers,
```
pascalkuthe, line 1328: This conflicts with the changes in #7814

helix-term/src/application.rs
```diff
@@ -565,6 +569,18 @@ impl Application {
             self.editor.refresh_language_servers(id);
         }
 
+        if self.editor.config().diff_source == DiffSource::File {
```
pascalkuthe, line 572: Yeah this should not be handled by reading the file again from disk (on the IO thread too!). Instead you should sent the current rope directly to the diff worker (altough this is pretty hacky to begin with)

pascalkuthe (changes requested):
Thabks fir working on this! Please try to keep your PRs as minimal as possible that nkes them easier to review and merge. This contains multie changes at once.

Some if them (see my review comments) are entirely orthogonal and also not the direction J want to got.

helix-view/src/macros.rs
```diff
@@ -73,3 +73,18 @@ macro_rules! doc {
         $crate::current_ref!($editor).1
     }};
 }
+
+#[macro_export]
+macro_rules! diff_provider {
```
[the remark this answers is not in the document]
doy: for similar reasons as the other macros - it allows just borrowing `self.vcs_providers` disjointly, which is necessary in some cases (borrowing all of `self` was pretty awkward)

helix-view/src/editor.rs
```diff
@@ -891,7 +907,7 @@ pub struct Editor {
     pub macro_replaying: Vec<char>,
     pub language_servers: helix_lsp::Registry,
     pub diagnostics: BTreeMap<lsp::Url, Vec<(lsp::Diagnostic, usize)>>,
-    pub diff_providers: DiffProviderRegistry,
+    pub vcs_providers: VcsProviderRegistry,
```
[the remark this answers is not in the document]
doy: this isn't renaming - it's actually replacing it with something different? i split up the existing `DiffProvider` trait into two separate traits - `DiffProvider` and `VcsProvider` which do different things, and this only handles the `VcsProvider` part (the `DiffProvider` part is handled elsewhere)

helix-vcs/src/lib.rs
```diff
@@ -20,44 +21,26 @@ pub trait DiffProvider {
...
-                }
-            })
-    }
+pub struct VcsProviderRegistry {
+    providers: Vec<Box<dyn VcsProvider>>,
+}
 
-    pub fn get_current_head_name(&self, file: &Path) -> Option<Arc<ArcSwap<Box<str>>>> {
+impl VcsProviderRegistry {
+    pub fn root_for_file(&self, file: &Path) -> Option<Box<dyn VcsRoot>> {
```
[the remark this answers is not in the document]
doy: should the `DiffProviderRegistry` just be removed entirely in that case? the reason i structured it this way is that i had thought that it would be useful to be able to try multiple vcs providers to figure out which one applied to the current directory, but then be able to diff against the one that it found in various different ways, but if you want to remove the fallback behavior entirely, i can also do that

helix-term/src/application.rs
doy: does this mean that we should always read an extra copy of the file into memory for every file we open? reading from disk means that we can avoid duplicating the file contents in memory until the user actually switches to the file-based diff, but if we only read the diff source when a file is opened, there wouldn't be any way to switch to the file-based diff source after changes have already been made to the buffer.

helix-vcs/src/git.rs
doy: oh, i didn't mean to actually remove this comment - the changes as they exist don't actually do any caching at all, they just always call `root_for_file` to recalculate the root (because yeah, doing the caching correctly would be tricky)

helix-vcs/src/lib.rs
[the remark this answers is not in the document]
pascalkuthe: yeah exactly I wnat to remove the fallback behaviour entirely

helix-vcs/src/git.rs
pascalkuthe: Ah I see I just kind of skimmed it and it looked lime you were doing caching with the root behavior ut now I see that you still tried to keep the fallback behavior. I think we should just remove all.fallback behavior (and so t add this extra notion of a root)

helix-term/src/application.rs
pascalkuthe: I don't think k I understand what you mean here.

Ofcourse this condition should still only trigger if the config option is set appropriaietly (the check is already there) but you can simply send the  document rope to the diff worker then instead of doing IO here.

Cloneing a rope does not cause a copy of the file to be created in memory (ropes are refernebced counted and copy on write)

doy:
i think this addresses the comments you had so far, let me know if there's anything else!

helix-view/src/document.rs
```diff
@@ -652,7 +653,8 @@ impl Document {
             path: None,
             encoding,
             has_bom,
-            text,
+            text: text.clone(),
+            original_text: text,
```
pascalkuthe, line 657: I think I get your easlier comment now. I didn't mean adding an extra field here. 

I was only talking about situations where `text` is known to be equal to the on-disk version these situations would be:

* when the file is first opened
*  after `:reload`
* after `:write`

in theses cases you should just send `doc.text().clone()` to the diff worker as a diffbase.

But when the diff provider gets changed (lets say by a seggints change) its indeed fine to read the file from disk again.

helix-view/src/document.rs
```diff
@@ -1584,16 +1584,48 @@ impl Document {
...
+                    Err(err) => {
+                        log::info!("{err:#?}");
+                        log::info!("failed to obtain current head name for {}", path.display());
+                        None
+                    }
+                };
+            }
+            DiffSource::File => {
+                self.set_diff_base_raw(self.original_text.clone());
+                self.version_control_head = None;
```
pascalkuthe, line 1624: I don't think version_control_head should be controlled by this setting, I think it should always use (git head). One we support multiple vcs this needs to be configured somehow but that should actually be automatic (a problem for the future). For now you can just hardcode git tough

helix-view/src/editor.rs
```diff
@@ -73,6 +73,14 @@ where
     )
 }
 
+#[derive(Debug, Default, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "kebab-case")]
+pub enum DiffSource {
+    #[default]
```
pascalkuthe, line 79: why did you remove the `None` vraint here?

I think its quite reasonable to allow people to turn the diff gutter off. 
Currently hiding the statusbar element will not actually disable the background diff computation but that may be desirable for better perf/battery life (especially if you have some random strang large file where you just don't case for the diff). Its a bit niche but I would be fine with adding it.

helix-view/src/editor.rs
```diff
@@ -289,6 +297,9 @@ pub struct Config {
     pub default_line_ending: LineEndingConfig,
     /// Enables smart tab
     pub smart_tab: Option<SmartTabConfig>,
+    #[serde(default)]
+    /// What the diff gutter should diff against
+    pub diff_source: DiffSource,
```
pascalkuthe, line 302: I think this should be just  `editor.diff-source` since the diff is also used for the `reset-diff-change` command and `]g` and also for a side by side diff in the future so it doesn't just pertain to the gutter

doy: this actually is just `editor.diff-source`, i guess the comment is a bit misleading. i'll clean up the docs etc once the feature itself is more finalized.

doy:
alright, how's this?

helix-view/src/editor.rs
pascalkuthe: ah sorry github collapsed the code so that it looked like it was a gutter config field and the comment reinforced that idea :sweat_smile: 

image

helix-view/src/document.rs
```diff
@@ -1584,25 +1580,67 @@ impl Document {
...
+                Err(err) => {
+                    log::info!("{err:#?}");
+                    log::info!("failed to open diff base for for {}", path.display());
+                    self.diff_handle = None;
+                }
+            },
+            DiffSource::File => {
+                if let Some(text) = text {
+                    self.set_diff_base_raw(text);
+                } else {
```
pascalkuthe, line 1616: sorry I modified the originial review comment but you were too fast.

Instead of making this a parameter you can simply check this from within the document:


```suggestion
                if !self.is_modified() {
                    self.set_diff_base_raw(self.text.clone());
                } else {
```

helix-view/src/document.rs
```diff
@@ -1584,25 +1580,67 @@ impl Document {
...
-            self.diff_handle = Some(DiffHandle::new(diff_base, self.text.clone()))
+            self.set_diff_base_raw(diff_base);
         } else {
             self.diff_handle = None;
         }
     }
 
-    pub fn version_control_head(&self) -> Option<Arc<Box<str>>> {
-        self.version_control_head.as_ref().map(|a| a.load_full())
+    pub fn set_diff_base_raw(&mut self, diff_base: Rope) {
```
pascalkuthe, line 1589: not really sure I love the naming here. How about just `sef_diff_base` for this function and `set_diff_base_bytes` for the other function?

Rope is the standard text data type in helix not bytes so the other function should be the one with a special neame

pascalkuthe:
I think this is getting pretty good now. I want the git diff provider to be configurable (so you can diff against any revision like `upstream/master` or `v23.5.0` and even `stage`. You don' t necessarily have to implement that in this PR (you can if you want to) but I think we should think about how that will be handled in the config to avoid future breaking changes

doy:
i thiiiink the current config should be extensible? like right now we have

```toml
[editor]
diff-source = "git" # or "file" etc
```

but this would probably still allow

```toml
[editor.diff-source]
type = "git"
rev = "main"
staged = true
```

or whatever in the future? i think it'll be easier to do that in a followup pr, but it seems pretty doable to me.

doy:
(the test failures look like upstream repositories having problems)

pascalkuthe (dismissed):
LGTM now 👍

poliorcetics:
I strongly disagree with forcing people to have a repo-local config each time they want to change the diff provider.

I want to configure helix once and have that config be enough for 99.99% of my usage, not have every repo on which I use JJ/git/no VCS (file) have a different configuration file and have to remember to copy some cargo-culted file from somewhere else.

Not only is it repetitive for very little gain, it also pollutes repos with local files, which mean adding git worktrees will not work the same for example, because the ignored files won't be copied over. Plus, Helix doesn't respect the `.config/` format for local configs, which also pollutes listings with "ls -la" for example

poliorcetics:
Some stats for example:

Currently on my personal machine, I have 14 repos using JJ, two using no VCS (playgrounds to try things out in Python and Rust) and 22 using git (I think I counted them all, I may have missed some in some deep folder tree)

By this PR, if I want to minimize configs, I have to write 16 different files for JJ/no VCS and set git at the default. That's not a good state

pascalkuthe:
the assumption is that you very much will be working with a single vcs system 99% of the time (really always git) and if you are forced to use something niche like mecurial adding a config file isn't too bad.

The fact that you seem evenly split between two different vcs is incredibly niche.

If anything I would not allow this to be configured at all and instead just hardcode the detection with the option to explicitly overwrite (which would be the same as this PR. We weren't planning to support vcs other than git when this was created which would mean the autodetectin would always return git)

poliorcetics:
If autodetection is a thing then yes, this PR will work well. Is there an issue for it ?

pascalkuthe:
clsoign thi onein favor of https://github.com/helix-editor/helix/pull/9951, thanks for working on this on!

PR #8064, opened 2023-08-25
feat: use which on formatter command

zaucy:
This allows my formatter command to work on Windows and Linux without having to edit my config between machines. Example:

```toml
[[language]]
name = "typescript"
formatter = { command = './node_modules/.bin/prettier', args = ["--parser", "typescript"] }
```


Masynchin:
You can get rid of `.unwrap()`'s by converting `Result` to `Option`. I wrote this code sample as an example of it, haven't test it though. You can use it if it is improves readability.

~~~rust
let fmt_cmd = |formatter| which::which(&formatter.command)
    .ok()
    .and_then(|cmd| cmd.into_os_string().into_string())

if let Some((fmt_cmd, fmt_args)) = self
    .language_config()
    .and_then(|c| c.formatter.clone())
    .and_then(|formatter| fmt_cmd(formatter).map(|cmd| (cmd, formatter.args.clone())))
~~~

zaucy:
Thanks! I removed the unwrap calls.

kirawi:
I think https://github.com/helix-editor/helix/pull/5832 addresses this.

zaucy:
I personally would prefer to not run through my shell (especially on Windows), but I can see the advantage with environment variables etc.

clo4:
> I personally would prefer to not run through my shell (especially on Windows)

Hmm, yeah I feel the same way - I'd rather not launch my shell just to run the formatter, that seems like unnecessary overhead. I keep my shell launching fast so it's no big deal to me, but there's no reason to run it just for this, and for people that have slower setups it'll definitely be noticeable.

The other issue with launching a shell is that it will very likely cause issues with other people's shell setups. Some people have interactive prompts when they launch a shell regardless of whether it's interactive or not (while working for Fig I saw a surprising amount of people with this 😔). Also seems like it might (?) cause issues with ZSH users that unconditionally prepend their path in `~/.zshenv`, for example with Python virtualenvs.

IIRC Helix now has its own reusable variable expansion, right? Could we use that here?

kirawi:
There is, but it only takes in a `Path`. You would also have to parse the arguments being passed, and correctly replace them with the expanded paths.

archseer:
> I personally would prefer to not run through my shell

I agree here, if shell expansion is desired the end user can always manually invoke the shell: `zsh -c 'formatter $VAR'` etc

archseer:
Does this also address #5538?

pascalkuthe:
> Does this also address #5538?

I don't think so arguments are not shell expanded here. We doul check if an argument starts with a tilde  and expand appropriately (or full shell expansion ala https://crates.io/crates/shellexpand).  

That seems orthogonal to me tough, this PR makes formatters behave the same as LSP servers which I think is the right thing for now. If we do endup adding such a feature I thing we should do the same thing for LSP servers.

pascalkuthe (approved):

helix-view/src/document.rs
```diff
@@ -733,16 +733,25 @@ impl Document {
...
             .language_config()
             .and_then(|c| c.formatter.clone())
-            .filter(|formatter| which::which(&formatter.command).is_ok())
+            .and_then(|formatter| {
+                Some((
+                    which::which(&formatter.command)
+                        .ok()?
+                        .into_os_string()
+                        .into_string()
+                        .ok()?,
```
pascalkuthe, line 745: Ah one thing I missed here, paths are not necessarily utf-8 you should just return a path here instead of converting to string

pascalkuthe (changes requested):
ah missed somethinmg

zaucy:
cargo check is passing locally for me, but the CI is failing. I don't think it's related to my PRs contents. Rebasing against master to see if it is fixed.

pascalkuthe (dismissed):

helix-view/src/document.rs
```diff
@@ -733,16 +733,16 @@ impl Document {
...
-            .filter(|formatter| which::which(&formatter.command).is_ok())
+            .and_then(|formatter| Some((which::which(&formatter.command).ok()?, formatter.args)))
         {
             use std::process::Stdio;
             let text = self.text().clone();
-            let mut process = tokio::process::Command::new(&formatter.command);
+            let mut process = tokio::process::Command::new(&fmt_cmd);
             process
-                .args(&formatter.args)
+                .args(&fmt_args)
```
the-mikedavis, line 745: We can remove a clone here:

```rust
if let Some((fmt_cmd, fmt_args)) = self
    .language_config()
    .and_then(|c| c.formatter.as_ref())
    .and_then(|formatter| Some((which::which(&formatter.command).ok()?, &formatter.args)))
{
    use std::process::Stdio;
    let text = self.text().clone();
    let mut process = tokio::process::Command::new(&fmt_cmd);
    process
        .args(fmt_args)
```

zaucy: thanks!

the-mikedavis (approved):

PR #8065, opened 2023-08-26
add command line option to execute commands on editor startup

doy:
i'd like to be able to do things like alias a command to `helix --command file_picker` to immediately jump to picking a file, or set `GIT_EDITOR="helix --command insert_mode"` to simplify the process of writing commit messages. this isn't a hugely impactful feature, but it's pretty hard to do the things it allows without it. it is also a fairly common editor feature (basically the same as `vim -c` or `kak -e`).

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -24,7 +24,7 @@ use tui::backend::Backend;
 
 use crate::{
     args::Args,
-    commands::apply_workspace_edit,
+    commands::{self, apply_workspace_edit},
     compositor::{Compositor, Event},
     config::Config,
     job::Jobs,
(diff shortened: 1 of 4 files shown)


HungryJoe:
When I used neovim, I had a little [script](https://github.com/HungryJoe/fish_config/blob/main/functions/select-files-to-edit.fish) that would search a directory for a given string, open files with matches, and search them for that string. I only needed to type the query once to see and edit all matches in the directory.

Coming to helix, I don't see a way to adapt this script since I can't run a command or populate a register on startup. This PR would help me greatly. Thanks for making it!

pascalkuthe:
we want to keep the command line inteerface minimal so I don't think we will be adding this to core. This also has conflicts so I am clsoing this out as stale. Thank you for contributing!

PR #8066, opened 2023-08-26
allow optionally highlighting the area past the end of the file

doy:
this makes it easier to see when the file is actually ended, rather than just having a bunch of whitespace near the bottom of the terminal

corresponds to the `NonText` highlight group in vim, or the `BufferPadding` face in kakoune.


doy:
anything i could do to get another review here? or is this not a feature people are interested in?

doy:

Issue #279, opened 2021-06-16
Support for EditorConfig


cessen:
In #239 it was mentioned that support for [EditorConfig](https://editorconfig.org) would be nice, and it seemed like there was general agreement that we want to do that at some point.

I'm opening this issue just so it stays on our radar.

pickfire:
While we do that, I think we want to have a logo first to be able to put our editor in editorconfig.org.

sudormrfbin:
It doesn't [seem to be maintained](https://github.com/mrandri19/rust-editorconfig/issues/9) so we might have to pull it in and make some changes.

pickfire:
Maybe we can take over the maintainership of the crate?

sudormrfbin:
There's an open issue for that too (offering maintainership): https://github.com/mrandri19/rust-editorconfig/issues/13

archseer:
I wonder if we could just attempt to parse it via TOML, it shares some similarities with INI.

cessen:
There's also [serde_ini](https://crates.io/crates/serde_ini), which seems like a reasonable way to go since we already have the serde dependency anyway.

kirawi:
It might just make sense to adapt https://github.com/zonyitoo/rust-ini. It seems like the only difference between `.ini` and `.editorconfig` are wildcard patterns in the section names.

sudormrfbin:
Dropping this here for future reference if we go for the manual implementation using an ini crate route, there is an [`editorconfig-plugin-tests` repository](https://github.com/editorconfig/editorconfig-plugin-tests) containing instructions and sample files for testing.

kirawi:
Worst case scenario, we can just bind to https://github.com/editorconfig/editorconfig-core-c. But I don't think we'll need to.

TheDaemoness:
What is the status of this feature's implementation? If there is a lack of manpower for it, then I'd be happy to contribute some.

kirawi:
Feel free :) I had some IRL things happen that prevented me from tackling hits.

TheDaemoness:
Is there a preference for wrapping editorconfig-core-c or writing something new in pure Rust? [The spec](https://editorconfig-specification.readthedocs.io/) seems simple enough for now.

kirawi:
https://github.com/zonyitoo/rust-ini seems to be able to parse it just fine.

TheDaemoness:
I'll take that as a preference for writing a pure Rust implementation. That was my preference as well. :+1:

[...]

PR #8076, opened 2023-08-27
languages: add editorconfig

solfisher:
note that this is unrelated to #279; this is simply a highlighting grammar for .editorconfig files, which i felt was useful regardless.

[homepage](https://editorconfig.org), [specification](https://spec.editorconfig.org/), [tree-sitter grammar i wrote](https://git.sr.ht/~sfr/tree-sitter-editorconfig)

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -30,6 +30,7 @@
 | dockerfile | ✓ |  |  | `docker-langserver` |
 | dot | ✓ |  |  | `dot-language-server` |
 | dtd | ✓ |  |  |  |
+| editorconfig | ✓ |  |  |  |
 | edoc | ✓ |  |  |  |
 | eex | ✓ |  |  |  |
 | ejs | ✓ |  |  |  |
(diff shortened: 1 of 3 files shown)


the-mikedavis (commented):
Being able to highlight editorconfig syntax without support for editorconfig seems needless to me. In a pinch you can use the `ini` language, so I don't think we need to add another language just for this use case

solfisher:
fair enough. should i turn this into a draft so we could look at it again when editorconfig support gets implemented?

yoav-lavi:
@the-mikedavis I think adding `.editorconfig` to https://github.com/helix-editor/helix/blob/941dc6c614d684804c3580c4ae8b3c4368aab217/languages.toml#L2275 in the meantime, until we have support would be a net benefit for everyone. I can open a PR if needed

yoav-lavi:
RE: Why you'd want this without support - Shared projects use EditorConfig files often, and being able to read them highlighted is a nicer experience even if the editor doesn't yet have the functionality to support them yet IMO

the-mikedavis:
INI seems like a nice compromise to me 👍. I'd like to avoid pulling in a special parser for the syntax if possible it anyways

the-mikedavis:
Superseded by https://github.com/helix-editor/helix/pull/8308

PR #8083, opened 2023-08-27
Expose "reset" color value in themes

### LINKED ISSUE #8075: Expose "reset" color value in themes
[The Color struct has a Reset value](https://github.com/helix-editor/helix/blob/e216e9621e73cda1968632cd20595231af5e07be/helix-view/src/graphics.rs#L215-L237), which should reset the color to the default terminal foreground or background. But this doesn't seem to be exposed to be used in themes, as far as I could find anyway.

Being able to access the default fg/bg colors in a theme is useful, because they may differ from the 16 colors/256 colors in the terminal color palette. In curses, the default color can be called with the number `-1`, while the colors are called with numbers `0..255`. I have no preference if helix should expose it like this or rather something else.

wildwestrom:
Seems like this was a pretty trivial fix, but please let me know if you spot anything obviously wrong here.

I tested it and it seemed like you could pretty easily modify `reset` to whatever you want and have it show up in your theme.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -70,6 +70,7 @@ over it and is merged into the default palette.
 
 | Color Name      |
 | ---             |
+| `reset`         |
 | `black`         |
 | `red`           |
 | `green`         |
--- a/helix-view/src/theme.rs
+++ b/helix-view/src/theme.rs
@@ -359,6 +359,7 @@ impl Default for ThemePalette {
     fn default() -> Self {
         Self {
             palette: hashmap! {
+                "reset".to_string() => Color::Reset,
                 "black".to_string() => Color::Black,
                 "red".to_string() => Color::Red,
                 "green".to_string() => Color::Green,


wildwestrom:
Since I can't put this in the title, this fixes #8075 to be clear.

book/src/themes.md
```diff
@@ -70,6 +70,7 @@ over it and is merged into the default palette.
 
 | Color Name      |
 | ---             |
+| `reset`         |
```
chtenb, line 73: Perhaps the documentation could use an update. 
> The default palette uses the terminal's default 16 colors
I might send a PR at some point making some clarifications to this documentation chapter in general.

chtenb:
I had a look at how the `Reset` value is used in `crossterm`, and I noticed that it's used for `default` colors, and does not correspond to the `reset` code as I assumed.

https://github.com/crossterm-rs/crossterm/blob/08762b3ef4519e7f834453bf91e3fe36f4c63fe7/src/style/types/colored.rs#L59

As you see, it's translated to code 39 and 49 which correspond to the [default fg and bg colors](https://gist.github.com/fnky/458719343aabd01cfb17a3a4f7296797#color-codes).

In this light, shouldn't we rather expose this as "default" instead of "reset"? The phrase "reset" could then be reserved for the actual reset code (`0`), which resets all modifiers as well as the colors (bg _and_ fg) to their defaults.

PR #8099, opened 2023-08-28
Add helix-specific ignore files

### LINKED ISSUE #7423: Helix-specific ignore file
I'm working in a project with lots of `.blend` files. These are committed to version control (so they're not in `.gitignore`). I also find it useful to type `blender <ctrl+t>`, using `fzf` to select a blend file to open, so I don't want to add `.blend` to `.ignore`. However, I don't want these files to show up in the helix file picker. Could we have helix respect an additional ignore file, `.helix/ignore`? This would contain patterns that should be ignored by helix, but not other tools that also use `.ignore`.

PaulOlteanu:
Fixes #7423

I chose `.hxignore` as the file name to match other tools such as ripgrep's `.rgignore`, but this is open to suggestions


poliorcetics:
I would much prefer `.helix/ignore` for two reasons:

1. We could have one in the global config directory that always apply, making it easy to ignore common but useless files
2. It would integrate with the existing `.helix` machinery, making less clutter (there is now `.helix` and `.hxignore`, when it already should always be `.config/hx/...`)

helix-view/src/editor.rs
```diff
@@ -177,6 +177,9 @@ pub struct FilePickerConfig {
     pub deduplicate_links: bool,
     /// Enables reading ignore files from parent directories. Defaults to true.
     pub parents: bool,
+    /// Enables reading an ignore file from `.helix/ignore`.
+    /// Whether to hide files listed in .helix/ignore in file picker and global search results. Defaults to true.
+    pub helix_ignore: bool,
     /// Enables reading `.ignore` files.
```
pascalkuthe, line 183: this doesn't need a config options. This file is helix specific, if you don't want to read it then delete it

pascalkuthe (changes requested):

pascalkuthe:
I think you could add the gloabl ignore file (see https://github.com/helix-editor/helix/pull/8300) right in this PR too, so we can get this done in one go

helix-term/src/commands.rs
```diff
@@ -2140,73 +2143,77 @@ fn global_search(cx: &mut Context) {
...
-                                        "Global search error: {}, {}",
-                                        entry.path().display(),
-                                        err
-                                    );
-                                }
-                                if stop {
-                                    WalkState::Quit
+                    walk_builder.add_custom_ignore_filename(".helix/ignore");
+                    walk_builder
+                        .add_custom_ignore_filename(helix_loader::config_dir().join(".ignore"));
```
the-mikedavis, line 2150:
```suggestion
                        .add_custom_ignore_filename(helix_loader::config_dir().join("ignore"));
```

to match the one under `.helix`

SecretPocketCat:
> I think you could add the gloabl ignore file (see #8300) right in this PR too, so we can get this done in one go

I thought these were orthogonal, so I created a new PR for the global file. Either way I'll be happy. Should I close #8300 then?

helix-term/src/ui/mod.rs
```diff
@@ -177,6 +177,9 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
         .max_depth(config.file_picker.max_depth)
         .filter_entry(move |entry| filter_picker_entry(entry, &absolute_root, dedup_symlinks));
 
+    walk_builder.add_custom_ignore_filename(".helix/ignore");
+    walk_builder.add_custom_ignore_filename(helix_loader::config_dir().join(".ignore"));
```
SecretPocketCat, line 181: The docs for [add_custom_ignore_filename](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.add_custom_ignore_filename) mention this takes precedence, which I'd argue doesn't make sense for a global ignore file and that's why I have opted for [add_ignore](https://docs.rs/ignore/latest/ignore/struct.WalkBuilder.html#method.add_ignore) in #8300 .

helix-term/src/commands.rs
PaulOlteanu: I opted to make them both `.ignore` instead because `.ignore` seems to be more common

helix-term/src/ui/mod.rs
pascalkuthe: I actually disagree. The ignore file is more specficic because it's specific to helix and should take priority. 

By making the global helic specific ignore file take priority that would allow people to always show some folder (like `node_modules`) in helix but still add it to gitignore.

I think the case where somebody whitelists a glob in their global helix ignorefile and still wants to ignore that file for a specific project is pretty niche and can be handled by adding that file to the helix specific ignore file for that project.


Priority should be 

* local helix ignore file 
* global high ignore file
* all other ignore files

This is what using `add_custom_ignorefile` accomplishes. The problem here was just the order (you should add the global ignore file first and then the local one so the local one has higher priority)

helix-term/src/commands.rs
the-mikedavis: I would prefer `ignore` - there's no need for the file to be hidden since it will always be under a hidden directory (`.config` or `.helix`)

pascalkuthe (approved):

the-mikedavis (approved):

gamma-delta:
Hello, just like to add my two cents here -- I'm using Helix as my editor for C# in Godot and the ability to mask out all the `.import` and `.tscn` files and so on would be *amazing*, especially because for LSP reasons I need to have Helix's origin directory be the root of the project. :)

David-Else:
@PaulOlteanu Hi, I have read the docs at https://docs.helix-editor.com/master/configuration.html?highlight=ignore#editorfile-picker-section, and it mentions a `.ignore` file, but if I want a global ignore file in my config dir then I need to use a `ignore` file, am i misunderstanding something or is this some kind of error? There seems to be no mention in the docs of a `ignore` file, and a `.ignore` file works in my local directory. Should the global ignore file not also have a `.` in front of it?

the-mikedavis:
It looks like we don't document these new files anywhere yet. That section in the docs could mention that you can use ignore files with a local `.ignore`, `~/.ignore` or now `.helix/ignore` and `~/.config/helix/ignore`

PR #8107, opened 2023-08-29
Adds ability to highlight surrounding brace pair

LukeHalasy:
Hello. I am excited to try and contribute to this great editor 😸 (and make my first open-source contrib).

This PR adds a new highlight key, `ui.color.surround_pair`, for specifying how you would like to highlight the current surrounding brace pair.


helix-term/src/ui/editor.rs
```diff
@@ -509,11 +511,25 @@ impl EditorView {
             {
                 // ensure col is on screen
                 if let Some(highlight) = theme.find_scope_index_exact("ui.cursor.match") {
-                    return vec![(highlight, pos..pos + 1)];
+                    elements.push((highlight, pos..pos + 1));
+                }
+            }
+
+            // Highlight surrounding brace
+            if let Some(highlight) = theme.find_scope_index_exact("ui.cursor.surround") {
```
LukeHalasy, line 519: I put this check first because I assume it would be faster than `find_matching_bracket_fuzzy`.

helix-term/src/ui/editor.rs
```diff
@@ -509,11 +511,25 @@ impl EditorView {
             {
                 // ensure col is on screen
```
LukeHalasy, line 512: is this comment relevant to the code ?

LukeHalasy:
Would this be better as a [editor] config option (and if the option is set to true then the surrounding braces use `ui.cursor.match` ?

LukeHalasy:
need to figure out the git mess I made. Only want to keep the three commits I made. Will fix this later

LukeHalasy:
> Would this be better as a [editor] config option (and if the option is set to true then the surrounding braces use `ui.cursor.match` ) ? Highlighting surrounding pairs can be a bit noisy and might not be everyone's cup of tea.

Addressed this comment. Now a config option

LukeHalasy:
@the-mikedavis 
Could I get a review on this?
Sorry to ping you directly if that's not the preferred way to let y'all know a PR is ready for review (let me know).

clo4:
This is a pretty useful feature! My thoughts though, I'd prefer having an underline on the matching pair instead of a background, because I know *for sure* that if it's always highlighted I'm going to think it's my cursor and make mistakes because of that 😅

In this case, I think it would make sense to maybe split this into two options:

```toml
[editor.highlight-surround]
show = "always"  # always | cursor | never
style = "block"  # block | underline
```

I'm not sold on either of the names but I think you see what I mean by them. One to change *when* it's displayed, the other to change *how* it's displayed.

For `editor.highlight-surround.show`, `cursor` would be the current behavior (highlight when cursor is on a paired character), `always` would be this PR's behavior, and `never` would allow people to disable the match highlighting altogether.

For `editor.highlight-surround.style`, using `block` would be the current default, `underline` would underline the match character.

That's just my thoughts as a normal user that wants the configurability for this feature :)

poliorcetics:
Style in helix works with themes, not in the `config.toml`. What you’re proposing already works in this PR by setting `ui.cursor.match = { modifiers = ["underline"] }` in your theme :)

clo4:
Oh, sweet, thank you for letting me know @poliorcetics! Ignore my suggestion 😶‍🌫️

PR #8109, opened 2023-08-29
Maintain the current cursor's position and view in the vsplit/hsplit commands too

TheRealLorenz:
Supersedes #7220.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4848,7 +4848,7 @@ fn split_helper(editor: &mut Editor, action: Action) {
 }
 
 fn hsplit(cx: &mut Context) {
-    split_helper(&mut cx.editor, Action::HorizontalSplit);
+    split_helper(cx.editor, Action::HorizontalSplit);
 }
 
 fn hsplit_new(cx: &mut Context) {
@@ -4856,7 +4856,7 @@ fn hsplit_new(cx: &mut Context) {
 }
 
 fn vsplit(cx: &mut Context) {
-    split_helper(&mut cx.editor, Action::VerticalSplit);
+    split_helper(cx.editor, Action::VerticalSplit);
 }
 
 fn vsplit_new(cx: &mut Context) {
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -4828,33 +4828,35 @@ fn transpose_view(cx: &mut Context) {
     cx.editor.transpose_view()
 }
 
-// split helper, clear it later
-fn split(cx: &mut Context, action: Action) {
-    let (view, doc) = current!(cx.editor);
+/// Open a new split in the given direction specified by the action.
+///
+/// Maintain the current view (both the cursor's position and view in document).
+fn split_helper(editor: &mut Editor, action: Action) {
```
pascalkuthe, line 4834: Avoid renames, they pollute diffs and and in RBIs case I also preferred the old naming. Similailry the argument should stay a context (same as for all.functions in commands.rs)

pascalkuthe (changes requested):
The core change lookms.gone to me but the diff should be +2-2, this contains a bunch unnecessary change that should be removed. 

The extra docs can stay jf you are inclines but the changes to the sifnsfjre should be reverted.

helix-term/src/commands.rs
gabydd: I thought this too when I was initially review this pr but the context for type-able commands and regular commands is different so the editor needs to be passed in instead of context so they can all use the same helper function

TheRealLorenz: Let me know if there's a way to maybe convert between the two context types

TheRealLorenz: I don't know if something like this could be used: 

https://github.com/helix-editor/helix/blob/48373d4a2b6402dce91ebcdf29b6341fcf0ba302/helix-term/src/commands.rs#L2838-L2845

gabydd: Ah sorry for not replying, if we only need the editor here it's better to just pass that instead of constructing a new context.

pascalkuthe: yeah I missed that in that case changing the argument is fine I suppose but the function still shouldnt' be renamed

pascalkuthe (approved):

the-mikedavis (approved):

PR #8111, opened 2023-08-29
Fix find commands for buffers with non-LF line-endings

### LINKED ISSUE #8057: T<ret> on windows selects the previous newline
### Summary

Pressing `T<ret>` on a line of text has the same effect as pressing `F<ret>`, the previous newline is selected:

image

Note that setting line-ending to `lf` makes `T` behave correctly.

### Reproduction Steps

1. Put cursor on a line of text.
2. Press `T<ret>`

### Platform

Windows

### Terminal Emulator

Windows Terminal Version: 1.17.11461.0

### Helix Version

helix 23.05 (7f5940be)

woojiq:
Adds separate method for finding `<ret>` char.

Closes #8057 

Resolves TODO: https://github.com/helix-editor/helix/blob/79c04251547e6c0c21366d322e268a6581c0f5dd/helix-term/src/commands.rs#L1270-L1273

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1256,7 +1256,7 @@ fn extend_next_long_word_end(cx: &mut Context) {
 // This is necessary because the one document can have different line endings inside. And we
 // cannot predict what character to find when <ret> is pressed. On the current line it can be `lf`
 // but on the next line it can be `crlf`. That's why [`find_char_impl`] cannot be applied here.
-fn find_char_ending(
+fn find_char_line_ending(
     cx: &mut Context,
     count: usize,
     direction: Direction,
@@ -1324,7 +1324,7 @@ fn find_char(cx: &mut Context, direction: Direction, inclusive: bool, extend: bo
                 code: KeyCode::Enter,
                 ..
             } => {
-                find_char_ending(cx, count, direction, inclusive, extend);
+                find_char_line_ending(cx, count, direction, inclusive, extend);
                 return;
             }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1251,6 +1251,65 @@ fn extend_next_long_word_end(cx: &mut Context) {
     extend_word_impl(cx, movement::move_next_long_word_end)
 }
 
+/// Separate branch to find_char designed only for <ret> char.
+//
+// This is necessary because the one document can have different line endings inside. And we
+// cannot predict what character to find when <ret> is pressed. On the current line it can be `lf`
+// but on the next line it can be `crlf`. That's why [`find_char_impl`] cannot be applied here.
+fn find_char_ending(
+    cx: &mut Context,
+    count: usize,
+    direction: Direction,
+    inclusive: bool,
+    extend: bool,
+) {
+    let (view, doc) = current!(cx.editor);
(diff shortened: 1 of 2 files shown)


pascalkuthe (commented):
I think this is an ok fix but I think a better solution would be to special case this branch. And just use ropey to find the next line ending/start from the cursor (look at the implementation of gl/gh). That should be faster and correctly handle files with mixed line endings (resolving the TODO comment here)

woojiq:
Done. Thank you for the suggestion. There are some edge cases thought. Comments explain them more clearly.

helix-term/src/commands.rs
```diff
@@ -1253,6 +1253,69 @@ fn extend_next_long_word_end(cx: &mut Context) {
...
+                } else {
+                    line as usize
+                }
+            }
+        };
+
+        let pos = match (direction, inclusive) {
+            (Direction::Forward, true) => line_end_char_index(&text, find_on_line),
+            (Direction::Forward, false) => {
+                graphemes::prev_grapheme_boundary(text, line_end_char_index(&text, find_on_line))
```
pascalkuthe, line 1302: nit: I know other parts of the codebase do the same thing but selections will be normalized to grapheme boundaries anyway so you can just do `line_end_char_index(...) - 1` instead and it will be a bit more efficient

helix-term/src/commands.rs
```diff
@@ -1253,6 +1253,69 @@ fn extend_next_long_word_end(cx: &mut Context) {
...
+            }
+            (Direction::Backward, true) => {
+                graphemes::prev_grapheme_boundary(text, text.line_to_char(find_on_line))
```
pascalkuthe, line 1305:
```suggestion
                line_end_char_index(&text, find_on_line)
```

this is more consistent with the line above (and also a tiny bit faster)

pascalkuthe (commented):
this looks very nice, some style nits but it seems like you covered all edgecases :+1:

pascalkuthe (approved):

helix-term/src/commands.rs
```diff
@@ -1266,13 +1325,9 @@ fn find_char(cx: &mut Context, direction: Direction, inclusive: bool, extend: bo
...
                 ..
-            } =>
-            // TODO: this isn't quite correct when CRLF is involved.
-            // This hack will work in most cases, since documents don't
-            // usually mix line endings.  But we should fix it eventually
-            // anyway.
-            {
-                doc!(cx.editor).line_ending.as_str().chars().next().unwrap()
+            } => {
+                find_char_ret(cx, count, direction, inclusive, extend);
```
the-mikedavis, line 1329: minor nit: let's call this `find_line_ending` instead

pascalkuthe (dismissed):

helix-term/src/commands.rs
the-mikedavis: Specifically I'm looking for "line ending" to be in the name since we use that language consistently throughout the codebase & docs. `find_char_ending` I think is ambiguous but `find_char_line_ending` / `find_line_ending_char` etc would be good

woojiq: Oops 😅

the-mikedavis (approved):

pascalkuthe (approved):

Issue #5632, opened 2023-01-22
Panic when exiting search


par5er:
### Summary

Editor panics when exiting search

### Reproduction Steps

 

I tried this:

1. `hx .`
2. Open a small file
3. Make a vertical split and open a larger file in the second pane
4. Search for something in the larger file and hit ENTER
5. Search for something again, and this time switch to the first pane without closing the search dialog
6. Hit ESCAPE and the program crashes

I expected this to happen:
The search dialog to close

Instead, this happened:
https://asciinema.org/a/eLc59nJruhSfKI2qW7whCK5ig

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

Linux (Debian 11)

### Terminal Emulator

st

### Helix Version

helix 22.12 (7ac72a39)

archseer:
```
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace    
```

the-mikedavis:
I can reproduce this. On a debug build the backtrace is:

```
thread 'main' panicked at 'assertion failed: char_idx <= slice.len_chars()', helix-core/src/graphemes.rs:37:5
stack backtrace:
   0: rust_begin_unwind
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/std/src/panicking.rs:584:5
   1: core::panicking::panic_fmt
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:143:14
   2: core::panicking::panic
             at /rustc/fe5b13d681f25ee6474be29d748c65adcd91f69e/library/core/src/panicking.rs:48:5
   3: helix_core::graphemes::nth_prev_grapheme_boundary
             at ./helix-core/src/graphemes.rs:37:5
   4: helix_core::graphemes::prev_grapheme_boundary
             at ./helix-core/src/graphemes.rs:79:5
   5: helix_core::graphemes::ensure_grapheme_boundary_prev
             at ./helix-core/src/graphemes.rs:194:9
   6: helix_core::selection::Range::grapheme_aligned
             at ./helix-core/src/selection.rs:271:17
   7: helix_core::selection::Selection::ensure_invariants::{{closure}}
             at ./helix-core/src/selection.rs:586:28
   8: helix_core::selection::Selection::transform
[...]

PR #8118, opened 2023-08-30
Clear completion when switching windows via click

the-mikedavis:
The completion component assumes that it operates on the same View but it's possible to break this assumption by switching windows through left-clicking. I believe we should clear the completion menu when switching windows to fix this.

This change fixes a panic for this scenario:

* Open a buffer with LSP completion available
* Split the window (for example `<C-w>v`)
* Enter insert mode and trigger the completion menu
* Select a completion candidate (for example with `<C-n>`)
* Switch to the original window by left-clicking in its area
* Enter insert mode and make edits (for example `o<backspace>`)

This will trip the `assert_eq` in Document::restore.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -1073,6 +1073,7 @@ impl EditorView {
                 let editor = &mut cxt.editor;
 
                 if let Some((pos, view_id)) = pos_and_view(editor, row, column, true) {
+                    let prev_view_id = view!(editor).id;
                     let doc = doc_mut!(editor, &view!(editor, view_id).doc);
 
                     if modifiers == KeyModifiers::ALT {
@@ -1080,6 +1081,10 @@ impl EditorView {
                         doc.set_selection(view_id, selection.push(Range::point(pos)));
                     } else {
                         doc.set_selection(view_id, Selection::point(pos));
+                    }
+
+                    if view_id != prev_view_id {
+                        self.clear_completion(editor);
                     }
 
                     editor.focus(view_id);


the-mikedavis:
Ideally we would close completion on any window switch from within `Editor::focus` but I believe that would need the move-compositor-to-helix-view change (https://github.com/helix-editor/helix/issues/5632#issuecomment-1399662726)

pascalkuthe (approved):
~We can merge this as is to get the panic out of tree but I actually included the same fix  in #8021 :D~


Sorry I think I confused something. This is a good catch!

pascalkuthe:
> Ideally we would close completion on any window switch from within `Editor::focus` but I believe that would need the move-compositor-to-helix-view change ([#5632 (comment)](https://github.com/helix-editor/helix/issues/5632#issuecomment-1399662726))

That issue was about the search prompt not completion. Could we do the same there (close the current prompt)?

the-mikedavis:
Yeah, I just mean that for both prompts and completion that we need access to the compositor from within (helix-view's) Editor to solve it really nicely as you say in #5632.

It should be possible to also close other components here:

```rust
let callback = if view_id != prev_view_id {
    self.clear_completion(editor);
    let callback: crate::compositor::Callback = Box::new(|compositor, _| {
        // Close any prompts or other components.
    });
    Some(callback)
} else {
    None
};

editor.focus(view_id);
editor.ensure_cursor_in_view(view_id);

return EventResult::Consumed(callback);
```

But that may need larger changes to the prompt and maybe other components. We would want to cancel whatever the prompt had selected, for example if you had `:theme <sometheme>` previewed and clicked another pane then it would essentially accept the preview but I would expect the prompt to cancel (`PromptEvent::Abort`).

I think it's possible with some more changes to the components that could be closed but it isn't quite trivial so I'd like to leave it off of this PR if possible.

PR #8121, opened 2023-08-30
Save an undo checkpoint before paste in insert mode

the-mikedavis:
I often paste and find out that my clipboard didn't have what I expected on it. Previously, undoing the change (`u` in normal mode) would also undo everything else I added in insert mode which is usually too much.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3948,6 +3948,10 @@ fn paste_impl(
         return;
     }
 
+    if mode == Mode::Insert {
+        doc.append_changes_to_history(view);
+    }
+
     let repeat = std::iter::repeat(
         // `values` is asserted to have at least one entry above.
         values


pascalkuthe (approved):
I agree with this :+1:  that said should we also do the same thing for the various paste commands in insert mode. It would seem odd if bracketed paste does this but a cumst binding like `a-v` wouldn't

the-mikedavis:
Yeah we could do this in `helix_term::commands::paste_impl` so it works for all paste commands in insert mode 👍. Pastes from other registers should be fine to have undo checkpoints for too (maybe you hit the wrong register before pasting with C-r for example).

pascalkuthe (approved):
A I keep forgetting about `c-r`, even tough it's such a nice binding 😅 Considering that c-r is a default binding it makes even Kore sense to create a checkpoint here 👍

archseer:
Somewhat related to https://github.com/helix-editor/helix/pull/7226 ? Should merge both in tandem

the-mikedavis:
Not exactly, this one only has to do with insert mode behavior rather than mode switching, similar to https://github.com/helix-editor/helix/pull/7747.

I've been meaning to take a look at #7226 and figure out what's going on in the mode-switching though, I'll give that a look soonish

pascalkuthe:
https://github.com/helix-editor/helix/pull/7226/commits/e2a877432ce7f75eba60d5dd579e5a0ab45b73e8 from that PR is actually almost the same as the original version of this PR I think?

The changes to :write are probably orthogonal

the-mikedavis:
Ah yeah actually I didn't notice that part. That part could be dropped after this change. It's saving after the paste which is a little different behavior-wise. I believe we want to save before the paste though not after.

PR #8126, opened 2023-08-31
goto_file: file path resolution relative to either the workspace or the current file

cbr9:
I'm currently working on a LaTeX project with the following structure:

The LaTeX compiler (I'm using `xelatex`) interprets paths relative to the main.tex file (which is at the root of the folder in this case). So even in `background/chapters/main.tex`, I can't just type `./ssl.tex`. Since the PR that implement relative paths in `goto_file_impl`, if I type `mim` (match inside the brackets) and then `gf` in one of the input statements on the left, helix will open the file `chapters/background/chapters/background/file.tex`. So to select the file I need now takes extra keystrokes.

This PR makes `goto_file_impl` first check if it can find the file relative to the current working directory, and if it doesn't exist, open it relative to the currently opened file.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1192,11 +1192,15 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     for sel in paths {
         let p = sel.trim();
         if !p.is_empty() {
-            let path = &rel_path.join(p);
+            let mut path = PathBuf::from(p);
+            if !path.exists() {
+                path = rel_path.join(p);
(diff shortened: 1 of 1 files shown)


cbr9:
Anything against this?

pascalkuthe:
I think the path relative to the current file should definitely take priority. Also the CWD is not the workspace root (which is what you are currently checking against).  Apart from that it's probably fine, We have lots of open PRs, it can take a while until we get around to a review.

PR #8127, opened 2023-08-31
avoid excessive memory consumption in picker

pascalkuthe:
We currently keep the picker around after it's closed to allow it to be reopened. This can lead to excessive memory consumption for very large pickers (like when crawling home with nucleo). This PR ensures that we drop the last picker immediately when a new picker is opened, that we stop streaming in new matches into the closed picker and the background and does not store pickers with more than 100_000 items. 

In the long term a better solution would be to restart the stream (recrawl the FS) for such large pickers instead, but that requires larger changes.

### DIFF
--- a/helix-term/src/compositor.rs
+++ b/helix-term/src/compositor.rs
@@ -101,7 +101,7 @@ impl Compositor {
 
     /// Add a layer to be rendered in front of all existing layers.
     pub fn push(&mut self, mut layer: Box<dyn Component>) {
-        // immidietly clear last_picker field to avoid excessive memory
+        // immediately clear last_picker field to avoid excessive memory
         // consumption for picker with many items
         if layer.id() == Some(picker::ID) {
             self.last_picker = None;
--- a/helix-term/src/compositor.rs
+++ b/helix-term/src/compositor.rs
@@ -16,6 +16,7 @@ pub enum EventResult {
 }
 
 use crate::job::Jobs;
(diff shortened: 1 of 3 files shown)


helix-term/src/compositor.rs
```diff
@@ -100,6 +101,11 @@ impl Compositor {
 
     /// Add a layer to be rendered in front of all existing layers.
     pub fn push(&mut self, mut layer: Box<dyn Component>) {
+        // immidietly clear last_picker field to avoid excessive memory
```
cd-a, line 104:
```suggestion
        // immediately clear last_picker field to avoid excessive memory
```

helix-term/src/ui/picker.rs
```diff
@@ -802,11 +803,28 @@ impl<T: Item + 'static + Send + Sync> Component for Picker<T> {
...
+            {
+                Box::new(|compositor: &mut Compositor, _ctx| {
+                    // remove the layer
+                    compositor.pop();
+                })
+            } else {
+                // stop streaming in new items in the background, really we should
+                // be restarting the stream somehow once the picker gets
+                // reopened instead (like for an FS crawl) that would also remove the
+                // need for the specical case above but that is pretty tricky
```
cd-a, line 819:
```suggestion
                // need for the special case above but that is pretty tricky
```

the-mikedavis (approved):

PR #2168, opened 2022-04-19
G goes to the last line if no count is provided

adaliaramon:

PR #8135, opened 2023-09-01
goto end of file if count is not present

gim913:
Hello,

I'm not quite sure if this change is something wanted, but coming from vim, I'm _terribly_ missing this.

`G` is bound to `goto_line`. The change allows `goto_line` without count, therefore `G` without count results in going to the end of file.

The only thing I wasn't sure about is if there should be `push_jump` or not, but it seems more natural to always add current location to jumplist.

From vim docs:

> With a `count`, this command positions you at the given line number.  For example, "33G" puts you on line 33.  (...)
> With no argument, `G` positions you at the end of the file.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3105,33 +3105,35 @@ fn push_jump(view: &mut View, doc: &Document) {
 }
 
 fn goto_line(cx: &mut Context) {
-    if cx.count.is_some() {
-        let (view, doc) = current!(cx.editor);
-        push_jump(view, doc);
+	let (view, doc) = current!(cx.editor);
+	push_jump(view, doc);
 
-        goto_line_without_jumplist(cx.editor, cx.count);
-    }
+	goto_line_without_jumplist(cx.editor, cx.count);
 }
 
 fn goto_line_without_jumplist(editor: &mut Editor, count: Option<NonZeroUsize>) {
-    if let Some(count) = count {
-        let (view, doc) = current!(editor);
-        let text = doc.text().slice(..);
-        let max_line = if text.line(text.len_lines() - 1).len_chars() == 0 {
-            // If the last line is blank, don't jump to it.
-            text.len_lines().saturating_sub(2)
-        } else {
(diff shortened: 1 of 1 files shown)


baldwindavid:
Thanks for this; it continually tripped me up. Most other tools I use such as Vimium support `G` to go to the end of the file. At first I thought `G` wasn't a thing at all in Helix, but then found that only the count works with it. And then I found that this sort of dual functionality DOES exist in Helix, but for `goto_file_start` rather than end. In any case, it would be nice to at least support getting that dual behavior via keybinding (or maybe you can and I don't know how) if not with the change in this PR.

nikitarevenco:
How about making this also extend selection until the end of the file? 

Would be really useful since we would have both G to extend selection and ge to jump to the end of the file without extending the selection

gim913:
I've rebased the change on top of current master.

@baldwindavid this change should already handle `G` keybinding properly (as it should already be bound to `goto_file` as per https://docs.helix-editor.com/keymap.html)

NikitaRevenco (approved):

NikitaRevenco:
honestly is there really a reason to have 2 keys doing the exact same thing? `ge` and `G` take the same amount of keystrokes. I like **ge** more for 2 reasons:
- mnemonic "goto end"
- faster to type

This pr was made before and it was rejected https://github.com/helix-editor/helix/pull/2168

gim913:
> honestly is there really a reason to have 2 keys doing the exact same thing? `ge` and `G` take the same amount of keystrokes. I like **ge** more for 2 reasons:
> 
> * mnemonic "goto end"
> * faster to type
> 
> This pr was made before and it was rejected #2168

reason for "G" - muscle memory ?
TBF/TBH, I wasn't aware there's `ge`. 

but this change isn't really about binding G,
I simply was wondering why `goto_line_without_jumplist` (and `goto_line`) are taking `count`  as an Option, when clearly they _always_ require count to actually do anything? (I mean I guess the reason is it's an Option inside the context, but `if` spanning whole functions feels weird)

PR #8136, opened 2023-09-01
Simplify README.md

kacperwyczawski:
Remove unnecessary text and add helix-themed buttons that redirect users to website, docs, wiki or matrix. You can see how it looks [here](https://github.com/kacperwyczawski/helix) or by using rich diff on github.

`[!NOTE]` might not render in preview, but it's supported by github (see https://github.com/orgs/community/discussions/16925) and should display properly in repo's readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,17 +18,17 @@
 
 ![Screenshot](./screenshot.png)
 
-A Kakoune / Neovim inspired editor, written in Rust.
-
-The editing model is very heavily based on Kakoune; during development I found
-myself agreeing with most of Kakoune's design decisions.
-
-For more information, see the [website](https://helix-editor.com) or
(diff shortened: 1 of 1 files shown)


David-Else:
Nice :)

How about a 4k screen shot with the right aspect ratio for a modern monitor? The current screen shot bugs me as it does not look like an actual screen. Added bonus it has 2 windows and the new git feature on display showing the branch :)

Screenshot from 2023-09-01 14-24-32

kacperwyczawski:
1. Good idea, thanks. I've added new screenshot and co-authored you.
2. I've messed up with git again, sorry. I'm closing this PR and [here](https://github.com/helix-editor/helix/pull/8138) is link to new one, with all commits squashed into one.

PR #8138, opened 2023-09-01
Update README.md and screenshot in it.

kacperwyczawski:
I've simplified the README a bit. Please visit [my fork](https://github.com/kacperwyczawski/helix/blob/master/README.md) for nice preview :)

Thanks to @David-Else for the new screenshot.


David-Else:
I think this whole section is pretty redundant, it is just not important enough to take space:
```
It's a terminal-based editor first, but I'd like to explore a custom renderer (similar to Emacs) in wgpu or skulpin.

Note
Only certain languages have indentation definitions at the moment. Please check runtime/queries/<lang>/ for indents.scm file.
```
How about instead a call to action button like 'Install Now' which links to the installation docs, like it does on the current main page.

It could have something cute like a list of contributors to emphasize the open source aspect of the project?

I think what you have done is cool for the design, but maybe some more and better content would be good. I am sure the core maintainers will have some input on that.

kacperwyczawski:
I agree with everything you said, but the changes are already quite *radical*, so I think that waiting for people more involved in this project is a good idea.

archseer:
I personally prefer the current version, new one has less information and replaces regular links with images

the-mikedavis:
I agree. Also the larger screenshot makes it harder to see details - I prefer what we have now

kacperwyczawski:
Alrighty then. I'm closing this PR, thanks for your input.

PR #8148, opened 2023-09-02
Do not show (running) when opening picker

### LINKED ISSUE #8142: Stop `(running)` flickering on and off when opening the new picker
The new nucleo picker is amazing, but it is annoying that the word `(running)` flickers on and off when opening the new picker however small the amount of items to pick is. It really catches the eye.

Is there a mechanism possible so that it won't display for a fraction of a second, and then only display if the search takes any longer? Maybe if it has displayed before it doesn't display again?

pascalkuthe:
Closes #8142

There were actually two issues here:
* `nucleo` had a bug where it would start the matcher but not wait for it to finish on the first call to `tick`. I fixed this upstream and also updated the `Cargo.lock` file a bit (using cargo update) to remove some duplicate dependencies.
* For the file picker the `(running)` indicator would still always appear because starting a background thread takes a while causing a timeout. I fixed this by running the FS transversal on the UI thread for up to 30ms. This is short enough to not cause UI lag but long enough for most reasonable directories to finish.

 For the global search picker the problem persists, but since the global search runs eagerly there is no way to fix that (one we start the search it will always block the main thread until its complete). I think its ultimately not a huge issue.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -19,17 +19,6 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "ahash"
-version = "0.7.6"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "fcb51a0695d8f838b1ee009b3fbf66bda078cd64590202a864a8f3e8c4315c47"
-dependencies = [
- "getrandom",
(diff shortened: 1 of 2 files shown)


poliorcetics (approved):

the-mikedavis (approved):

gabydd:
can confirm this works beautifully (probably should have mentioned this in the initial pr whoops)

PR #8155, opened 2023-09-03
Add title box to `Picker`

### LINKED ISSUE #8139: Add title to fuzzy picker box
The picker box in it's current implementation doesn't provide any context about what is being picked for.  
Below is a screenshot of the picker on global-search for the term `Viewport`, there is no way to find what search term the picker is referring to once you hit Enter on the search option.  

image

The above issue persists for other picker related things like the references picker etc. A helpful title such as "Global search: Viewport" or "LSP References: SomeStruct" would be nice.

ArchUsr64:
Fixes #8139  
The current implementation of the picker doesn't display any context information about what it's displaying.  
This PR adds a `title` field to the `Picker` struct and implements the necessary logic for proper rendering of the title box based on the config option `PickerTitle`.  
```rust
pub struct Picker<T: Item> {
...
    /// Title for the picker box
    title: String,
}
```
```rust
pub enum PickerTitle {
    /// Don't render the title
    #[default]
    Never,
    /// Render on top of picker box
    Center,
    /// Render inline with the picker's border
    Inline,
    /// Render inline with the picker's border
    InlineBorder,
    /// Render title in the prompt line
    Prompt,
}
```
This changes the rendering of a picker as follows:
```md

                `inline`                 `inline-border`           `center`

                                                                  +-------+
                                  +-------+                       | title |
         +-title-+-+ +---------+  | title +-+ +---------+  +------+--+ +--+------+
         | prompt  | | preview |  | prompt  | | preview |  | prompt  | | preview |
         |---------| |         |  |---------| |         |  |---------| |         |
         | picker  | |         |  | picker  | |         |  | picker  | |         |
         |         | |         |  |         | |         |  |         | |         |
         |         | |         |  |         | |         |  |         | |         |
         |         | |         |  |         | |         |  |         | |         |
         |         | |         |  |         | |         |  |         | |         |
         |         | |         |  |         | |         |  |         | |         |
         +---------+ +---------+  +---------+ +---------+  +---------+ +---------+
```
The `title` is exposed to the caller via `Picker::new()` and `Picker::with_stream()` functions.
I have added the following as the titles for now:
| Caller | Title | Dynamic |
| -| :-: | :---: |
|`global_search`|**Global Search: '`user_entered_regex`' [`case_sensitive`]** | :heavy_check_mark:|
|`buffer_picker`|**Buffer List** | |
|`jumplist_picker`|**Jumplist List** | |
|`command_palette`|**Command Palette** | |
|`file_picker`|**Files in: `current working directory`** | :heavy_check_mark: |
|`sym_picker`|**Symbol Picker** | |
|`diag_picker`|**Diagnostics: [`errors` E] [`warnings` W] [`informations` I] [`hints` H]** | :heavy_check_mark: |
|`lsp_workspace_command`|**LSP Workspace Commands**| |
|`goto_impl`|**Goto Picker**| |
|`thread_picker`|**Thread Picker**| |
|`dap_launch`|**DAP Picker**| |
|`dap_switch_stack_frame`|**DAP Stack Frame Picker**| |

### Changes to Picker:
image
### File Picker:
image
### Global Search Picker:
image
### Diagnostics Picker:
image

### Dynamic Titles:
https://github.com/helix-editor/helix/assets/83179501/ea3c3441-feb1-4d2c-ad80-739b8eb99187
### Rendering Options:
https://github.com/helix-editor/helix/assets/83179501/6ae20ae0-bbdf-499b-9f34-0f393d6873fd

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -228,6 +228,15 @@ Search specific options.
 | `smart-case` | Enable smart case regex searching (case-insensitive unless pattern contains upper case characters) | `true` |
 | `wrap-around`| Whether the search should wrap after depleting the matches | `true` |
 
+
+### `[editor.picker]` Section
+
+Picker specific options.
+
+| Key | Description | Default |
+|--|--|---------|
+| `title` | Weather to draw picker's titlebox | `true` |
+
 ### `[editor.whitespace]` Section
 
 Options for rendering whitespace with visible characters. Use `:set whitespace.render all` to temporarily enable visible whitespace.
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -235,7 +235,7 @@ Picker specific options.
 
 | Key | Description | Default |
 |--|--|---------|
-| `title` | Weather to draw picker's titlebox | `true` |
+| `picker-title` | Renders a title for the opening picker. Can be `never`, `center`, `inline`, `inline-border` or `prompt` | `never` |
 
 ### `[editor.whitespace]` Section
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -2211,7 +2211,9 @@ fn global_search(cx: &mut Context) {
 
                 cx.jobs.callback(async move {
                     let call = move |_: &mut Editor, compositor: &mut Compositor| {
+                        let picker_title = String::from("Global Search");
                         let picker = Picker::with_stream(
+                            picker_title,
                             picker,
                             injector,
                             move |cx, FileResult { path, line_num }, action| {
@@ -2680,7 +2682,8 @@ fn buffer_picker(cx: &mut Context) {
     // mru
     items.sort_unstable_by_key(|item| std::cmp::Reverse(item.focused_at));
 
-    let picker = Picker::new(items, (), |cx, meta, action| {
+    let picker_title = String::from("Buffer List");
+    let picker = Picker::new(picker_title, items, (), |cx, meta, action| {
         cx.editor.switch(meta.id, action);
     })
     .with_preview(|editor, meta| {
@@ -2757,7 +2760,9 @@ fn jumplist_picker(cx: &mut Context) {
         }
     };
 
+    let picker_title = String::from("Jumplist Picker");
     let picker = Picker::new(
+        picker_title,
         cx.editor
             .tree
             .views()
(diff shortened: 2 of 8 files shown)


pascalkuthe:
I am not a fan of adding a title it feels unnecessary and tahkesup more vertical space. With #7265 there will be a heading for each column so at that point it gets too much

ArchUsr64:
> I am not a fan of adding a title it feels unnecessary and tahkesup more vertical space. With #7265 there will be a heading for each column so at that point it gets too much

I would agree with you that the title would be redundant if it just displayed some static information but for something like global search, the additional context the title provides is quite helpful, have a look:

image

As for the concerns regarding limited vertical space, I'm working on an implementation that just adds title to the top of existing picker instead of shrinking the picker to make room for the title box.

EpocSquadron:
Could we not display the title inline with the bounding box at top, similar to what we do with the title for keymap modes?

ArchUsr64:
> Could we not display the title inline with the bounding box at top, similar to what we do with the title for keymap modes?

This is how that could look like, I'm not sure I prefer either of them over the title box being at the center, though not wasting extra space is a plus.
image
image

EpocSquadron:
I quite like the look of the second option, but from a minimal point of view the first is best. Curious what others think.

ArchUsr64:
> I quite like the look of the second option, but from a minimal point of view the first is best. Curious what others think.  

Me too though you might wanna have a look at my original [post](https://github.com/helix-editor/helix/pull/8155#issue-1879099463) which I have since updated with the new changes to title rendering.  

Would a configuration option be better? Something like `picker.title_style` with options of `center`, `inline`, `inline_border` and `none`?

zen3ger:
Since the promp/commandline at the bottom is always empty when a picker is opened, can't we just put this information there? It wouldn't take up any extra space and would still be there for those that need it.

EpocSquadron:
It’s definitely not intuitive to put it in the prompt, and I wouldn’t want it inconsistent from picker to picker . Global search will eventually not use the prompt anyway, once interactive global search gets done.

I say (without much authority) we go for the configurable but on by default approach, similar to how autoinfo is there by default but can be configured off by those who prefer not to have the noise. A big draw of the editor is contextual help or intuitive/unsurprising behavior. 

As for which style to use, since the works already done I don’t see any harm in keeping both versions as part of configuration. My bias is again to use the prettier version as the default.

ArchUsr64:
> I say (without much authority) we go for the configurable but on by default approach, similar to how autoinfo is there by default but can be configured off by those who prefer not to have the noise. A big draw of the editor is contextual help or intuitive/unsurprising behavior.  

I agree, it's probably the best to let users decide for themselves. For now, I have implemented the configuration options for all the suggestions presented above namely: `center`, `inline`, `inline-border` and `prompt`. Though I have set the default to `never` for now.  
The top post has been updated with all the relevant details if you wanna have a look.

book/src/configuration.md
```diff
@@ -228,6 +228,15 @@ Search specific options.
...
 | `wrap-around`| Whether the search should wrap after depleting the matches | `true` |
 
+
+### `[editor.picker]` Section
+
+Picker specific options.
+
+| Key | Description | Default |
+|--|--|---------|
+| `picker-title` | Renders a title for the opening picker. Can be `never`, `center`, `inline`, `inline-border` or `prompt` | `never` |
```
poliorcetics, line 238: The formulation here can make a non-fluent reader think that the title will be `never`, `center`, ...

What do you think about:

```suggestion
| `title-format` | How the picker's title is displayed. Can be `never`, `center`, `inline`, `inline-border` or `prompt` | `never` |
```

helix-term/src/commands.rs
```diff
@@ -2209,9 +2209,27 @@ fn global_search(cx: &mut Context) {
...
                     let call = move |_: &mut Editor, compositor: &mut Compositor| {
+                        let picker_title = format!(
+                            "Global Search: '{}' {}",
+                            regex,
+                            if case_sensitive {
+                                "[Case Sensitive]"
+                            } else {
+                                "[Case Insensitive]"
+                            }
+                        );
```
poliorcetics, line 2230: With a long regex, this will be quite unreadable, the length of the displayed regex should be limited to the N (to determine) first chars of it.

helix-term/src/commands/lsp.rs
```diff
@@ -290,8 +296,28 @@ fn diag_picker(
...
+        .for_each(|severity| match severity {
+            DiagnosticSeverity::INFORMATION => informations += 1,
+            DiagnosticSeverity::WARNING => warnings += 1,
+            DiagnosticSeverity::ERROR => errors += 1,
+            DiagnosticSeverity::HINT => hints += 1,
+            _ => (),
+        });
 
+    let picker_title = format!(
+        "Diagnostics: [{} E] [{} W] [{} I] [{} H]",
```
poliorcetics, line 316: I think just: `{} E, {} W, {} I, {} H` will be enough and have less sigils to read. Commas are easier to skip than `[]` too

helix-term/src/ui/mod.rs
```diff
@@ -199,16 +199,22 @@ pub fn file_picker(root: PathBuf, config: &helix_view::editor::Config) -> Picker
...
-        if let Err(e) = cx.editor.open(path, action) {
-            let err = if let Some(err) = e.source() {
-                format!("{}", err)
-            } else {
-                format!("unable to open \"{}\"", path.display())
-            };
-            cx.editor.set_error(err);
-        }
-    })
+    let picker_title = format!("Files in {:?}", root.as_os_str());
```
poliorcetics, line 202: This will also have the length problem and should probably be collapsed, only expand the last N directories, if it reaches the width limit of the windows

helix-term/src/ui/picker.rs
```diff
@@ -762,13 +776,96 @@ impl<T: Item + 'static> Picker<T> {
...
+                block.render(area, surface);
+
+                // render the title text
+                surface.set_string(
+                    // Add two for spacing for border box and margin
+                    area.x + 2,
+                    area.y + 1,
+                    &self.title,
+                    text_style,
+                );
```
poliorcetics, line 823: Yes I think the picker title will overflow if it is too long, I remember hitting this with inline diagnostics

pascalkuthe:
I am going to close this PR. This also  already has conflicts and will become very hard to rebase once #9647 is merged.  I don't think we want to add this feature in this form.  This is a mimor visual feature where I just don't think it's worth or necessary to invest the effort.

Thank you for contributing.

PR #8170, opened 2023-09-04
Add tree-sitter-highlight-name command

### LINKED ISSUE #8123: Add tree-sitter-highlight-name command for viewing highlight name of node at the cursor's location
Add a command, `tree-sitter-highlight-name`, that will output the name of the theming scope needed to modify the look of the node at the cursor's location.

The functionality would be the same as `tree-sitter-scopes`.

So for instance in the following screenshot, the `tree-sitter-highlight-name` command _would_ show `keyword.function`.
Is their any interest in this command? If so I would love to implement it.

tree-sitter-scopes_command_output
tree-sitter-highlight-scope-fn

LukeHalasy:
Closes #8123.

Add a command, `tree-sitter-highlight-name`, that outputs the name of the theming scope needed to modify the look of the node at the cursor's location.

Note: The ``json part that appears in the screenshots below has been addressed
result_1
result_2
result_3

### DIFF
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -5,7 +5,7 @@ use tree_sitter::{Query, QueryCursor, QueryPredicateArg};
 use crate::{
     chars::{char_is_line_ending, char_is_whitespace},
     graphemes::{grapheme_width, tab_width_at},
-    syntax::{LanguageConfiguration, RopeProvider, Syntax},
+    syntax::{Highlight, HighlightEvent, LanguageConfiguration, RopeProvider, Syntax},
     tree_sitter::Node,
     Rope, RopeGraphemes, RopeSlice,
 };
@@ -894,6 +894,44 @@ pub fn get_scopes(syntax: Option<&Syntax>, text: RopeSlice, pos: usize) -> Vec<&
 
     scopes.reverse();
     scopes
+}
+
+pub fn get_highlight_name(
+    syntax: Option<&Syntax>,
+    text: RopeSlice,
+    pos: usize,
+) -> Option<Highlight> {
+    let mut highlight: Option<Highlight> = None;
+    if let Some(syntax) = syntax {
+        let pos = text.char_to_byte(pos);
+
+        let node: Node = match syntax
+            .tree()
+            .root_node()
+            .descendant_for_byte_range(pos, pos)
+        {
+            Some(node) => node,
+            None => return None,
+        };
+
+        for event in syntax
+            .highlight_iter(text.slice(..), Some(node.byte_range()), None)
+            .map(|event| event.unwrap())
+        {
+            match event {
+                HighlightEvent::Source { start, end } => {
+                    if start == node.start_byte() && end == node.end_byte() {
+                        break;
+                    }
+                }
+                HighlightEvent::HighlightStart(hl) => {
+                    highlight = Some(hl);
+                }
+                HighlightEvent::HighlightEnd => {}
+            }
+        }
+    }
+    highlight
 }
 
 #[cfg(test)]
--- a/helix-core/src/indent.rs
+++ b/helix-core/src/indent.rs
@@ -5,7 +5,7 @@ use tree_sitter::{Query, QueryCursor, QueryPredicateArg};
 use crate::{
     chars::{char_is_line_ending, char_is_whitespace},
     graphemes::{grapheme_width, tab_width_at},
-    syntax::{Highlight, HighlightEvent, LanguageConfiguration, RopeProvider, Syntax},
+    syntax::{LanguageConfiguration, RopeProvider, Syntax},
     tree_sitter::Node,
     Rope, RopeGraphemes, RopeSlice,
 };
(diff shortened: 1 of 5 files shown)


helix-core/src/indent.rs
```diff
@@ -896,6 +896,44 @@ pub fn get_scopes(syntax: Option<&Syntax>, text: RopeSlice, pos: usize) -> Vec<&
     scopes
 }
 
+pub fn get_highlight_name(
```
LukeHalasy, line 899: I feel like this shouldn't be in this file. What would be a better file for this ? I put this in here initially because the treesitter-scopes function was in here

helix-term/src/commands/typed.rs
```diff
@@ -1527,6 +1527,46 @@ fn tree_sitter_scopes(
...
+    }
+
+    let (view, doc) = current!(cx.editor);
+    let text = doc.text().slice(..);
+
+    let pos = doc.selection(view.id).primary().cursor(text);
+
+    let mut highlight_name: &str = "";
+    if let Some(highlight) =
+        helix_core::syntax::get_highlight_for_node_at_position(doc.syntax(), text, pos)
```
the-mikedavis, line 1546: I don't think we want this function to live in helix-core since I don't think we'll re-use it in the future. Instead let's have a helper function within this function for finding the highlight at the cursor. That will also help us handle the `Option`s a little cleaner via the `?` operator:

```rust
fn find_highlight_at_cursor(
    cx: &mut compositor::Context<'_>,
) -> Option<helix_core::syntax::Highlight> {
    use helix_core::syntax::HighlightEvent;

    let syntax = doc.syntax()?;
    let cursor = doc.selection(view.id).primary().cursor(text);
    let byte = text.char_to_byte(cursor);
    let node = syntax
        .tree()
        .root_node()
        .descendant_for_byte_range(byte, byte)?;
    // Query the same range as the one used in syntax highlighting.
    let range = {
        // Calculate viewport byte ranges:
        let row = text.char_to_line(view.offset.anchor.min(text.len_chars()));
        // Saturating subs to make it inclusive zero indexing.
        let last_line = text.len_lines().saturating_sub(1);
        let height = view.inner_area(doc).height;
        let last_visible_line = (row + height as usize).saturating_sub(1).min(last_line);
        let start = text.line_to_byte(row.min(last_line));
        let end = text.line_to_byte(last_visible_line + 1);

        start..end
    };

    let mut highlight = None;

    for event in syntax.highlight_iter(text, Some(range), None) {
        match event.unwrap() {
            HighlightEvent::Source { start, end }
                if start == node.start_byte() && end == node.end_byte() =>
            {
                return highlight;
            }
            HighlightEvent::HighlightStart(hl) => {
                highlight = Some(hl);
            }
            _ => (),
        }
    }

    None
}
```

Then we can call that below and return early if there isn't a highlight for the cursor:

```rust
    return Ok(());
}

let Some(highlight) = find_highlight_at_cursor(cx) else {
    return Ok(());
};
```

helix-core/src/syntax.rs
```diff
@@ -2456,6 +2456,44 @@ pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node) -> fmt::Result
...
+                    }
+                }
+                HighlightEvent::HighlightStart(hl) => {
+                    highlight = Some(hl);
+                }
+                HighlightEvent::HighlightEnd => {}
+            }
+        }
+    }
+    highlight
```
the-mikedavis, line 2494: This should be `None` instead as in the snippet I left above: if there isn't a highlight for the node then this function would return the last highlight given by the iterator instead when it should be `None`

helix-core/src/syntax.rs
```diff
@@ -2456,6 +2456,44 @@ pub fn pretty_print_tree<W: fmt::Write>(fmt: &mut W, node: Node) -> fmt::Result
...
+            .tree()
+            .root_node()
+            .descendant_for_byte_range(pos, pos)
+        {
+            Some(node) => node,
+            None => return None,
+        };
+
+        for event in syntax
+            .highlight_iter(text.slice(..), Some(node.byte_range()), None)
```
the-mikedavis, line 2478: The `range` argument here should be the same as we use [in rendering](https://github.com/helix-editor/helix/blob/61814fea7f779267bd0c6bc9428c1f53142ee7f1/helix-term/src/ui/editor.rs#L295-L304) so that the command always returns the highlight as you see from syntax highlighting

helix-term/src/commands/typed.rs
```diff
@@ -1527,6 +1527,46 @@ fn tree_sitter_scopes(
...
+    {
+        let theme = &cx.editor.theme;
+        highlight_name = theme.scope(highlight.0);
+    }
+
+    let content = format!("``json\n{:?}\n````", highlight_name);
```
the-mikedavis, line 1552: Looks like the bacticks are unbalanced here which is causing the <code>``json</code> in the screenshots. I don't think we need any fancy formatting here though, let's just turn the scope into a String and use that without syntax highlighting:

```rust
let content = cx.editor.theme.scope(highlight.0);
```

helix-term/src/commands/typed.rs
```diff
@@ -2684,6 +2724,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
         fun: tree_sitter_scopes,
         signature: CommandSignature::none(),
     },
+    TypableCommand {
+        name: "tree-sitter-highlight-name",
+        aliases: &[],
+        doc: "Display tree-sitter highlight name.",
```
the-mikedavis, line 2730:
```suggestion
        doc: "Display name of tree-sitter highlight scope under the cursor.",
```

the-mikedavis (commented):
I'm liking the approach here, it just needs some tweaks to make it a little more idiomatic

LukeHalasy:
> I'm liking the approach here, it just needs a few tweaks

Appreciate the review! Have implemented all suggested tweaks (The ? operator is awesome).

helix-view/src/theme.rs
```diff
@@ -297,6 +297,11 @@ impl Theme {
         self.highlights[index]
     }
 
+    #[inline]
+    pub fn scope(&self, index: usize) -> String {
+        self.scopes[index].clone()
```
the-mikedavis, line 302:
```suggestion
    pub fn scope(&self, index: usize) -> &str {
        &self.scopes[index]
```

it's usually a nicer API to return borrowed data (`&str`) rather than owned (`String`) so that the caller can decide when/if to own the data. For example a caller of this function might only need to check `scope.starts_with("keyword")` which it can do without allocating a `String`.

Callers that need to allocate though (like `tree_sitter_highlight_name`) can then make the call themselves:

```rust
let content = cx.editor.theme.scope(highlight.0).to_string();
```

LukeHalasy: That makes sense. Appreciate the explanation. Addressed in most recent commit.

frou:
Thanks so much for implementing this. Without it, it is really laborious to tweak one's theme to get a specific language looking just right.

Issue #7224, opened 2023-06-04
Failure to Copy to System Clipboard in tmux in Specific Directory


tqwewe:
### Summary

While operating in the `$HOME/dev` directory (including its subdirectories) and using the yank command `Space + y` to copy text to the system clipboard, the operation fails to execute as expected. However, when I navigate up to the `$HOME` directory, the same operation functions as expected.

This issue seems to be specifically tied to operating in an SSH session from a Mac into a PC. Using a terminal directly from the PC does not reproduce the issue; all clipboard functions work correctly in that scenario.

### Reproduction Steps

 

**Steps to Reproduce:**

1. SSH into the PC from a Mac.
2. Launch tmux and navigate to `$HOME/dev` directory (or any subdirectory).
3. Attempt to yank text to system clipboard using `Space + y`.

**Expected Behavior:**

The yanked text should be copied to the system clipboard, regardless of the current directory.

**Actual Behavior:**

No visible error upon pressing the keyboard shortcut, but an error message appears when running the `:clipboard-yank` command:
> Couldn't set the system clipboard content: clipboard provider tmux failed

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-06-04T13:01:47.888 helix_view::editor [ERROR] editor error: Couldn't set system clipboard content: clipboard provider tmux failed
```

</details>

### Platform

Linux & macOS

### Terminal Emulator

Alacritty on Mac

### Helix Version

helix 22.12

Jackojc:
Same issue for me however seems to be present in any directory. My setup looks like WSL2 -> SSH -> Tmux -> Helix. Clipboard works fine from Tmux but not within Helix.

logan272:
same issue, helix system clipboard(Space + y) is not working inside tmux.  

**Platform**
macOS

**Terminal Emulator**
iterm2

**Helix Version**
23.05(built from master branch)

adriangalilea:
Having this issue, but it's very erratic, sometimes it works, sometimes it doesn't.

JosiahBLB:
Similar issue, Space + p does not work.

To fix Space + y not working, I did the following:
In iterm2, enabling `Applications in terminal may access clipboard` in `General > Selection` [source](https://github.com/helix-editor/helix/issues/8715)

As a crude fix for the clipboard paste I set the following keymaps:
```toml
[keys.normal.space]
P = ":insert-output tmux send-keys i; pbpaste; tmux send-keys Escape"
[...]

PR #8182, opened 2023-09-05
Detect tmux clipboard provider on macOS

xJonathanLEI:
Currently, only `pbcopy` and `pbpaste` are allowed as clipboard provider on macOS, seemingly without good reasons.

When SSH-ing into a macOS server, there's no way to yank to the client clipboard, even inside a tmux session, as Helix refuses to use tmux as clipboard provider on macOS, and the default provider (`pbcopy`) does not pass the content to SSH clients.

This PR adds the same tmux test to macOS.


dead10ck (approved):
This seems fine to me 👍

tqwewe:
This might fix https://github.com/helix-editor/helix/issues/7224 , though I haven't tested it yet.

PR #8194, opened 2023-09-06
Fix Clone definition for Injector

the-mikedavis:
Injectors detect that a Picker has been shut down / removed by checking the shutdown atomic bool (which is shared with the Picker instance and is set to `true` on Picker's Drop). Cloning Injector should create more references to the Picker's atomic bool rather than creating new ones. This should fix global search so that the WalkBuilder halts when the picker is dropped.


pascalkuthe (approved):

archseer:
Note: that field is typoed as "shutown" but we can fix that later

PR #8198, opened 2023-09-07
Goto buffer by num

DreckSallow:
Implement the select buffer by index :

- Command: `:buffer-index <number>`
- Alias: `:bi <number>`
- Command: `:buffer-first`
- Command: `:buffer-last`

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -9,6 +9,7 @@
 | `:buffer-close-others!`, `:bco!`, `:bcloseother!` | Force close all buffers but the currently focused one. |
 | `:buffer-close-all`, `:bca`, `:bcloseall` | Close all buffers without quitting. |
 | `:buffer-close-all!`, `:bca!`, `:bcloseall!` | Force close all buffers ignoring unsaved changes without quitting. |
+| `:buffer-index`, `:bi` | Goto buffer by number [1-9].|
 | `:buffer-next`, `:bn`, `:bnext` | Goto next buffer. |
 | `:buffer-previous`, `:bp`, `:bprev` | Goto previous buffer. |
 | `:write`, `:w` | Write changes to disk. Accepts an optional path (:write some/path.txt) |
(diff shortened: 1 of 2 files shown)


tmke8:
Thanks for implementing this! It looks to me like `:buffer-index 9` brings you to the last buffer but I think that might be better handled with a separate command like `:buffer-last` (and there could also be `:buffer-first` though that is of course equivalent to `: buffer-index 1`). Then people can map whatever keys they want to `:buffer-last`.

And I think it would be okay if `:buffer-index` also worked with higher numbers? Like, `:buffer-index 56`. Not sure what exactly the use case would be but there might be one.

gabydd:
I think for larger numbers we could just min it with the amount of buffers and then any higher number would just bring you to the last one, then we don't have to special case 9 you can just use something like 999 to go to the last buffer. Also mentioning this pr: https://github.com/helix-editor/helix/pull/4756 so they are linked

DreckSallow:
I wanted to apply the same behavior of Chrome and Firefox tabs. But I think your approaches are good.
Now you can:
- Access the first buffer using `:buffer-first` or `:bf`
- Access the last buffer using `:buffer-last` or `:bl`
- Access any buffer by its index with `:buffer-index <n>` or `:bi <n>` . If the index exceeds the number of buffers, it will automatically switch to the last buffer.

tmke8:
#8471 has been merged, so it should now be possible to use `space = {1 = ":buffer-index 1"}` as a mapping.

pascalkuthe:
I don't think we will be going forward with this one. The bufferline and the "gui like ordered tabs" workflow is not a priorty. We focus on picker baswd navigation instead.

The current buffer ordering is an implementation detail and I would like to not make users rely on it with commands or introsuce somekind of manual order (buffer picker is sorted by most recently used by contrast which I much prefer).

Sorry we didn't give you that kind of feedback sooner. Thank you for contributing!

obedm503:
The buffer picker is exactly the issue because it changes so much as you work through multiple files. Being able to jump around without reading or typing the buffer name saves time and maintains focus.

Right now I open the files I'll work on in some specific order (there's no way to order the bufferline manually) and `g-n`/`g-p` my way around. The buffer picker sorted by recency is impractical.

markstos:
@obedm503 On PR #8875 I proposed that the "amp-like" jumping to other locations should extended other windows: https://github.com/helix-editor/helix/pull/8875#issuecomment-2120809262

If implemented, it would also directly jumping to  other buffers, at the ones that have some of their content or name visible.

vaknin:
@pascalkuthe 

> I don't think we will be going forward with this one. The bufferline and the "gui like ordered tabs" workflow is not a priorty. We focus on picker baswd navigation instead.

What if someone doesn't want to hit 7 keys in order to simply switch a buffer?  
Why not make it configurable, and only allow it if someone wants it?

PR #8217, opened 2023-09-09
Add additional YAML injections

gjabell:
Adds injections for Github Actions & Gitlab CI.

Taken from https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/yaml/injections.scm, which is licensed under Apache-2.0. Let me know if I need to change the license text here :)

### DIFF
--- a/runtime/queries/yaml/injections.scm
+++ b/runtime/queries/yaml/injections.scm
@@ -1,2 +1,54 @@
 ((comment) @injection.content
  (#set! injection.language "comment"))
+
+; The remaining code in this file incorporates work covered by the following
+; copyright and permission notice:
+;
+;   Copyright 2023 the nvim-treesitter authors
+;
+;   Licensed under the Apache License, Version 2.0 (the "License");
+;   you may not use this file except in compliance with the License.
+;   You may obtain a copy of the License at
+;
+;       http://www.apache.org/licenses/LICENSE-2.0
+;
+;   Unless required by applicable law or agreed to in writing, software
+;   distributed under the License is distributed on an "AS IS" BASIS,
+;   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;   See the License for the specific language governing permissions and
+;   limitations under the License.
+
+; Modified for Helix from https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/yaml/injections.scm
+
+;; Github actions ("run") / Gitlab CI ("scripts")
+(block_mapping_pair
+  key: (flow_node) @_run (#match? @_run "^(run|script|before_script|after_script)$")
+  value: (flow_node
+           (plain_scalar
+             (string_scalar) @injection.content)
+           (#set! injection.language "bash")))
+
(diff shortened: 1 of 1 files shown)


runtime/queries/yaml/injections.scm
```diff
@@ -1,2 +1,91 @@
...
+                  (#set! injection.language "bash")
+                  (#offset! @injection.content 0 1 0 0))))))
+
+;; Prometheus Alertmanager ("expr")
+(block_mapping_pair
+  key: (flow_node) @_expr (#eq? @_expr "expr")
+  value: (flow_node
+           (plain_scalar
+             (string_scalar) @injection.content)
+           (#set! injection.language "promql")))
```
archseer, line 64: We don't actually have a promql grammar so these will fail

runtime/queries/yaml/injections.scm
```diff
@@ -1,2 +1,91 @@
...
+;   distributed under the License is distributed on an "AS IS" BASIS,
+;   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;   See the License for the specific language governing permissions and
+;   limitations under the License.
+
+; See https://github.com/nvim-treesitter/nvim-treesitter/blob/master/queries/yaml/injections.scm
+
+;; Github actions ("run") / Gitlab CI ("scripts")
+(block_mapping_pair
+  key: (flow_node) @_run (#any-of? @_run "run" "script" "before_script" "after_script")
```
the-mikedavis, line 25: We don't support `#any-of?`. This will need to be rewritten as a `#match?`: `(#match? @_run "^(run|script|before_script|after_script)$")`

runtime/queries/yaml/injections.scm
```diff
@@ -1,2 +1,91 @@
...
+           (plain_scalar
+             (string_scalar) @injection.content)
+           (#set! injection.language "bash")))
+
+(block_mapping_pair
+  key: (flow_node) @_run (#any-of? @_run "run" "script" "before_script" "after_script")
+  value: (block_node
+           (block_scalar) @injection.content
+           (#set! injection.language "bash")
+           (#offset! @injection.content 0 1 0 0)))
```
the-mikedavis, line 36: We don't support `#offset!` - this line can be removed

runtime/queries/yaml/injections.scm
archseer: Note: any-of? Was added upstream to tree-sitter now

the-mikedavis: Ah yep, once we update tree-sitter we'll have support for `#any-of?`. The commit we are currently pinned to came right before that change

the-mikedavis (approved):

runtime/queries/yaml/injections.scm
the-mikedavis: We're looking to update again when the next release of tree-sitter gets cut but we're not sure when that will be yet

pascalkuthe (approved):

PR #8222, opened 2023-09-09
Update EdgedDB (ESDL) grammar

greym0uth:
This updates the esdl tree sitter grammar to the most recent version.

For context: Edgedb v3 came with some syntax changes which required an update to the tree sitter along with some other fixes.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2069,7 +2069,7 @@ roots = ["edgedb.toml"]
 
 [[grammar]]
 name ="esdl"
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Do the highlight queries need any changes? Looking at the diff for the grammar it looks like we might be ok without query changes

greym0uth:
@the-mikedavis Nope, I didn't make any changes to any of the syntax tokens, so current highlights work all the same.

the-mikedavis (approved):

pascalkuthe (approved):

PR #8162, opened 2023-09-04
Cli --working-path option added to helix-term

lloydbond:
You can now use -w or --working-path to set the current working path of the hx editor from command line.

Similar to these other editors [here](https://docs.godotengine.org/en/stable/tutorials/editor/external_editor.html)
example:
`hx -w ~/helix/helix-term/ -- src/args.rs src/application.rs`
confirm it works by checking your current working directory from inside the editor with `:pwd`

PR #8223, opened 2023-09-09
enable starting hx with a working directory

lloydbond:
You can now use -w or --working-dir to set the current working directory of the hx editor from command line launch.

Similar to these other editors [here](https://docs.godotengine.org/en/stable/tutorials/editor/external_editor.html)
example:
`hx -w <path> --log hx.log -- <files>...`
`hx --working-dir <path> --log hx.log -- <files>...`
`hx -w <path> --log hx.log`
`hx --log hx.log <path>`
confirm it works by checking your current working directory from inside the editor with `:pwd`
confirm the files are opened with `:bn`


pascalkuthe:
I don't think this need to be builtin, you could just do `cd {project} && hx {file}`. I don't think this needs to be builtin, we try to keep the argument parsing minimal

lloydbond:
Yes, That is correct. This improves `hx`'s ability to be triggered as an external editor.

Also I can close this [PR](https://github.com/helix-editor/helix/pull/8162)

helix-term/src/application.rs
```diff
@@ -162,15 +162,19 @@ impl Application {
             // Unset path to prevent accidentally saving to the original tutor file.
             doc_mut!(editor).set_path(None);
         } else if !args.files.is_empty() {
-            let first = &args.files[0].0; // we know it's not empty
+            let mut files = args.files.clone();
```
pascalkuthe, line 165: Instead of cloning here (which is a bit inefficient you can just use:


```suggestion
            let mut files = &args.files;
```

and `files = &files[1..]` instead of `remove`

archseer:
Using a regular parameter and making this order dependent seems wrong for me and potentially confusing for ens users. Why not just use a proper flag?

helix-term/src/application.rs
lloydbond: I switched to a reference to a slice. I think that gets what you are looking for with `files = &files[1..]` Then I also had to deference a variable (`pos`) in one spot further down. I'm new to Rust so please let me know if I am jumbling and mixing syntax.

lloydbond:
Hi @archseer, it was mentioned in the matrix chat that the developers of Helix-Term are trying to avoid adding flags. You can see in a comment I made above that I mention closing a [PR](https://github.com/helix-editor/helix/pull/8162) in favor of this PR, that PR does use a flag for setting the initial workspace directory. Either or both approaches is fine with me.

archseer:
On Matrix I was specifically talking about configuration flags that would be put in config.toml, not CLI flags.

I'd prefer this be a flag rather than a parameter

gabydd:
I think that was me that said we try not too add many command line options, I think based on some comments by mike in other issues, though from what I just read know those where about not having command line options that could be config options. From what I remember from matrix it seems like the reason for this was because there was use case where ``cd`` couldn't be used? If cli flag is preferred that is good with me :+1:

lloydbond:
Okay, do we want to close this PR and reopen this [PR](https://github.com/helix-editor/helix/pull/8162) with the `-w` flag? I'm good with either.

archseer:
Why not just update the current PR by force pushing? No need to open a PR each time the approach is changed, that way the conversation isn't fragmented

lloydbond:
Sure, that works. Let me make the updates.

helix-term/src/application.rs
lloydbond: Resolving since no longer relevent.

helix-term/src/main.rs
```diff
@@ -66,6 +66,7 @@ FLAGS:
     -V, --version                  Prints version information
     --vsplit                       Splits all given files vertically into different windows
     --hsplit                       Splits all given files horizontally into different windows
+    -w, --working-path <path>      Specify an initial working directory
```
pascalkuthe, line 69:
```suggestion
    -w, --working-dir <path>      Specify an initial working directory
```

I think calling working-diretocry is the more common naming convention (your documentation even calls it that) and more concise since that path can not be a a directory)

helix-term/src/args.rs
```diff
@@ -17,6 +17,7 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub working_path: Option<PathBuf>,
```
pascalkuthe, line 20: similiary this should be callid `working_directory`

pascalkuthe (changes requested):

lloydbond:
@pascalkuthe I updated the code on `line 172` to have 
```
if let Some(path) = args.working_directory {
            helix_loader::set_current_working_dir(path)?
        }
```
Then I realized, If there are no files passed, then the -w flag would get completely ignored and no working directory would be set. This could be okay, but I think not an expected behavior from a normal user's perspective. For now, I have moved the code snippet above to `line 159`. 

This means that `hx -w <path>` and `hx <path>` are functionally the same. 
Also `hx -w <path1> -- <path2> <file1> <file2>`
would follow these steps
1. set the working directory to `path1`
2. set the working directory to `path2` like normal
3. open the fuzzy file finder window like normal
4. ignore `file1` and `file2` like normal
Unless there is some other intended behavior that is preferred, I think this is the better option for now.

Other options could be: 
1. original solution with the code on `line 172` and an additional `else if` check to test if we have `-w` set but `args.files` is empty (not my favorite and complicates things).
2. we could go with the original solution and a note in the `--help` message for `-w` flag that mentions, `-w` does nothing if no files are passed with it.
3. original solution only, and when `-w` silently does nothing, let the user figure out they should have gone with `hx <path>` and not `hx -w <path>`

Let me know what you think. Hopefully this doesn't overly complicate things.

pascalkuthe (dismissed):
I think k the way you implemented it now makes sense and is quite simple. I prefer to keep the cli simple 👍

helix-term/src/main.rs
```diff
@@ -66,6 +66,7 @@ FLAGS:
     -V, --version                  Prints version information
     --vsplit                       Splits all given files vertically into different windows
     --hsplit                       Splits all given files horizontally into different windows
+    -w, --working-dir <path>      Specify an initial working directory
```
the-mikedavis, line 69:
```suggestion
    -w, --working-dir <path>       Specify an initial working directory
```

pascalkuthe (approved):

the-mikedavis (approved):

PR #8233, opened 2023-09-10
Fix various typos

amousset:
Various typos, found using [typos-cli](https://github.com/crate-ci/typos). This actually modifies the molokai and jellybeans themes as some keys were incorrect (but the changes "look" correct).

### DIFF
--- a/helix-core/src/syntax.rs
+++ b/helix-core/src/syntax.rs
@@ -962,7 +962,7 @@ impl Syntax {
         let res = syntax.update(source, source, &ChangeSet::new(source));
 
         if res.is_err() {
-            log::error!("TS parser failed, disabeling TS for the current buffer: {res:?}");
+            log::error!("TS parser failed, disabling TS for the current buffer: {res:?}");
             return None;
         }
         Some(syntax)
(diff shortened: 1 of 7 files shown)


helix-loader/build.rs
```diff
@@ -32,7 +32,7 @@ fn main() {
     // we need to revparse because the git dir could be anywhere if you are
     // using detached worktrees but there is no good way to obtain an OsString
     // from command output so for now we can't accept non-utf8 paths here
-    // probably rare enouch where it doesn't matter tough we could use gitoxide
+    // probably rare enough where it doesn't matter tough we could use gitoxide
```
gabydd, line 35:
```suggestion
    // probably rare enough where it doesn't matter, though we could use gitoxide
```

gabydd (changes requested):

the-mikedavis (commented):
We try to avoid changes like this to the Rust files since they cause conflicts for any open PRs that also modify those lines. For example I have some changes to the Picker that will conflict here and the conflict resolution will be tricky because the blocks/indentation change.

So I think it's fine to fix all the typos in the themes but let's leave the changes to Picker and global_search out - I'll fix those in later PRs. The other changes to the source files I don't think have any open conflicts so those are probably ok to fix in this PR along with the theme fixes

amousset:
Makes sense, especially for typos in comments. I only kept the fixes in error logs and themes.

pascalkuthe (approved):

the-mikedavis (approved):

Issue #5921, opened 2023-02-11
`(os error 123)` with `:cd` to name containing quotes


yashpalgoyal1304:
### Summary

- When the subdirectory (or a file too?) contianing a space is tab selected in helix's command line
- only that subdirs name is enclosed in quotes (this is also standard behaviour in bash too)

i suspect that this (quotes in middle of name) is what causes the error 
but since, those quotes were put by helix itself, this is bug.

This same error is raised when the name contain any naked spaces 
but that is not a bug, that's a user error.

_Originally posted by @yashpalgoyal1304 in https://github.com/helix-editor/helix/issues/5882#issuecomment-1426836599_

### Reproduction Steps

 

I tried this:

1. `hx`
2. try typing :cd C:\Prog <tab> <ret>
i.e. execute command :cd C:\"Program Files"
or essentially any root folder which contains subfolders with spaces in their names.

I expected this to happen:

```
Current working directory is now C:\Program Files
```

Instead, this happened:

```
Couldn't change the current working directory: The filename, directory name, or volume label syntax is incorrect. (os error 123)
```

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
I did not file any helix directory under `%userprofile%/.cache`
```

</details>

### Platform

Windows

### Terminal Emulator

wezterm 20221119-145034-49b9839f

### Helix Version

helix 22.12 (96ff64a8)

yashpalgoyal1304:
this issue reduces to filepath and autocompletion handling in helix command line....

Following are some ways to solve
- not supporting DOS or win32 styled naked backslashes for sublevel (not ideal since copy pasting paths won't work)
    - the unix style path of forward slashes `/` for sublevel provides escape sequences using backward slashes `\` so `\ ` to avoid quotes
- autocompletion can be made to put the whole path in quotations
- the command can be made to deal with in between quotations

vgwidt:
I think either way `C:\"Program Files"` should be an acceptable input as that works in Windows, though `"C:\Program Files"` is more common.  Typing out `:cd "C:\Program Files"` does work, typing `:cd "C:\` then autocompleting results in `:cd "C:\"Program Files"` being typed.

I thought maybe I could get away with just adding the following to the Unqouted match in Shellwords::from to handle quotes in the middle of the word:
```
'"' => {
    if cfg!(windows) {
        Quoted
    } else {
[...]

PR #8279, opened 2023-09-14
Fix open windows path with spaces in prompt

vgwidt:
This allows Windows paths to be opened using the prompt if the path has a space in it (#5921).  Right now for `:cd` or `:open`, if a path has spaces you must type the entire path manually between quotes.  The prompt autocomplete currently puts quotes between folders or files with spaces, but it 1. is not separated properly in shellwords, 2. cannot be opened  with quotes in the middle of a path, and 3. puts the separator before the closing quote which breaks the file list builder.

The " characters need to be stripped out from the string if they are in the middle of the path. It is possible to use a path surrounded with doublequotes, and I originally wanted to rewrite the argument to add a " at the beginning of the argument when a file or folder with spaces is selected, but I ended up with too hacky of a solution using shellwords in the change_completion_selection function.  I think if there were an ergonomic way to do this, I'd much prefer to do it that way. 

I was tempted to add the bit that strips out the quotes in expand_tilde 😅. Not sure if there's a better way to do this.

The caveat with this fix is it still will not work if you start the path with " (i.e. typing :cd "C:\"Program Files"" will not work). Entering :cd "C:\Program Files" does work but the autocomplete prompts will not work. I had issues maintainly compatibility with other tests and #4098 when trying to get this to work in shellwords.  Maybe it could be worked around by checking if the argument starts with with " but then we run into the same issue above of determining where the argument begins.

### DIFF
--- a/helix-core/src/path.rs
+++ b/helix-core/src/path.rs
@@ -165,7 +165,7 @@ pub fn get_truncated_path<P: AsRef<Path>>(path: P) -> PathBuf {
 pub fn remove_quotes_from_windows_path(path: &Path) -> PathBuf {
     if cfg!(windows) {
         // remove double quotes from paths from Windows
-        PathBuf::from(path.to_string_lossy().replace("\"", ""))
+        PathBuf::from(path.to_string_lossy().replace('\"', ""))
     } else {
         path.to_path_buf()
     }
--- a/helix-core/src/path.rs
+++ b/helix-core/src/path.rs
@@ -160,3 +160,13 @@ pub fn get_truncated_path<P: AsRef<Path>>(path: P) -> PathBuf {
     ret.push(file);
     ret
 }
+
+/// Removes double quotes from Windows paths, otherwise returns path unchanged.
+pub fn remove_quotes_from_windows_path(path: &Path) -> PathBuf {
+    if cfg!(windows) {
+        // remove double quotes from paths from Windows
+        PathBuf::from(path.to_string_lossy().replace("\"", ""))
+    } else {
+        path.to_path_buf()
+    }
+}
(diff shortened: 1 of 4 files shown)


pascalkuthe:
Hmm so you seemed to have solved this very generically by allowing arbitrary quoted oath components.

Foes windows actually support those (like in a shell). I was under the impression you need to always quite the entire path (which should already work) and the issue here was that auto completions currently don't do that and instead only quote the last component.

vgwidt:
> Hmm so you seemed to have solved this very generically by allowing arbitrary quoted oath components.
> 
> Foes windows actually support those (like in a shell). I was under the impression you need to always quite the entire path (which should already work) and the issue here was that auto completions currently don't do that and instead only quote the last component.

The Windows shell does actually support this.  It is very forgiving; even `cd C:\"Program Files\"Something' works.  But it is less known, and quoting the whole path is probably preferred as it is more standard.

In #5921 it is mentioned that maybe both ways should be supported, and I guess this PR would at least cover allowing quotes around the individual components.  I wonder if there is a way for Rust to handle these paths for Windows without having to explicity remove the quotes...

What could alternatively be done is to remove the automatic inserts of " in the components with spaces as this just doesn't work at all in its current state.  Then it would be up to the user to add " at the front if they anticipate using paths with spaces　as autocomplete will work properly as long as the components aren't quoted.  They don't even have to add a " at the end for it to work :).  Then later if there is a more ergonomic way to track the beginning of an argument in the prompt, the " could be automatically added to the beginning when selecting an autocompletion with spaces.  I don't think there is a strong need to support individual components being quoted so I'm almost leaning towards this as being the better option now.  And the tradeoff with the PR in its current state is that adding a quote at the beginning wouldn't work with autocomplete...

vgwidt:
Shellwords is being reworked so this can probably be closed.

PR #3092, opened 2022-07-18
add agda language support (and als language server)

pretentious7:
Hello, this is just a draft pr for adding agda language server support. I'll be figuring out the treesitter highlighting queries and code actions next weeked  hopefully.

PR #8285, opened 2023-09-14
Add support for Agda

omentic:
supersedes #3092 

todo:
- [x] try to get the updated tree-sitter code upstreamed (instead of relying on my fork)
  - https://github.com/tree-sitter/tree-sitter-agda/pull/19
- [x] make queries better
  - [ ] properly match functions
  - [x] more fine-grained matching of types, ex. `(expr (expr (expr... (atom) @type)))`
    - lost cause, see https://github.com/tree-sitter/tree-sitter/issues/880
- [ ] indentation queries?


omentic:
Syntax highlighting just about working. Trying to get functions working, then I think this'll be ready to merge.

The language server does not appear to work at all, unfortunately. I might remove it or leave it in with a note.

image

omentic:
This PR is ready for review. I may have a follow-up sometime adding proper function highlighting & als support, the former appears to be a legitimate upstream bug and the latter still a mystery. 

@the-mikedavis can you squash merge this or would you prefer me to rebase?

the-mikedavis (dismissed):

the-mikedavis:
We can squash + merge this, no need to rebase

the-mikedavis (dismissed):

Septias:
Is this stalling?

omentic:
It's done on my end, I've been running it locally for a while. The syntax highlighting is unfortunately pretty terrible - but it's nothing that can be addressed until https://github.com/tree-sitter/tree-sitter/issues/880 is addressed. Any more specificity with regard to types kills performance exponentially (> 3 ish is unusable) and breaks highlighting for nested types above a certain level.

Septias:
Do you use it regularly and would you advise others to do so aswell? @omentic

omentic:
I mean, it's fine, it looks just like the above picture. I do not write Agda regularly in Helix and I would not advise others to do so as well because Helix lacks support for Unicode input, and Unicode input is required to write Agda.

When a snippets system lands, it'd be fine, I guess: but without a "proof assistant mode" or similar not great.

pascalkuthe (dismissed):

pascalkuthe:
This needs co flicts resolved before being g merged

the-mikedavis (approved):

omentic:
(tangential re: a proof assistant mode: it's very likely it could piggyback off of debug mode, before or after a redesign. i've had no time to poke around and prototype something, unfortunately, but might in the far future...)

pascalkuthe (approved):

PR #8300, opened 2023-09-16
Global .ignore file

SecretPocketCat:
I was missing an option to use a global .ignore file to globally ignore (some) binary files. Some of those (e.g. images) might be checked into VCS, so the global git ignore is not helpful in that case.  
I considered just checking whether the file exists and adding it but then switched to adding it to the config instead. The option could also be a path to the global .ignore.

A command to create such a default global .ignore could be added, but a follow-up PR would probably make more sense for that.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -86,6 +86,7 @@ mode.normal = "NORMAL"
 mode.insert = "INSERT"
 mode.select = "SELECT"
 ```
+
 The `[editor.statusline]` key takes the following sub-keys:
 
 | Key           | Description | Default |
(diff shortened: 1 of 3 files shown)


book/src/configuration.md
```diff
@@ -171,6 +171,7 @@ All git related options are only enabled in a git repository.
 |`deduplicate-links` | Ignore symlinks that point at files already shown in the picker | true
 |`parents` | Enables reading ignore files from parent directories | true
 |`ignore` | Enables reading `.ignore` files | true
+|`global-ignore` | Enables reading a global `.ignore` file from the same directory as the config.toml configuration | true
```
archseer, line 174: Why do we need a feature flag? If the user doesn't have an ignore file then this feature is effectively disabled anyway

SecretPocketCat: It's mentioned in the OP.
> I considered just checking whether the file exists and adding it [...]

Should I switch to that approach?
Where should I put the info about the global .ignore? configuration.md or some place else?

SecretPocketCat: How's the current version?

SecretPocketCat:
I somehow overlooked this comment in the docs
> Set options for file picker and global search. Ignoring a file means it is not visible in the Helix file picker and global search.

I should probably add the global ignore to the global search too then.

An alternative approach would be adding a global ignore for each (e.g. `file_picker.ignore` & `global_search.ignore`) to make it more granular. Not sure if that would be all that useful though and not just muddy the configuration waters.

SecretPocketCat:
Closing in favour of #8099

PR #8306, opened 2023-09-16
Use Maskhjarnas tree-sitter-purescript

NomisIV:
Currently PureScript piggybacks on the haskell tree-sitter, which is less than optimal. I use @Maskhjarna's tree-sitter-purescript, which currently is the most feature complete implementation. I use this daily in my work, and while it has a couple bugs, it's far better than the haskell tree-sitter.

### DIFF
--- a/runtime/queries/purescript/textobjects.scm
+++ b/runtime/queries/purescript/textobjects.scm
@@ -1,13 +1,13 @@
 (comment) @comment.inside
 
 [
-  (decl_adt)
-  (decl_type)
+  (data)
+  (type)
   (newtype)
 ] @class.around
 
 ((signature)? (function rhs:(_) @function.inside)) @function.around 
 (exp_lambda) @function.around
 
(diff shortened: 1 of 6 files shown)


runtime/queries/purescript/textobjects.scm
```diff
@@ -0,0 +1,13 @@
+(comment) @comment.inside
+
+[
+  (decl_adt)
```
the-mikedavis, line 4: It looks like this is failing the query check since the node isn't in the grammar

the-mikedavis (commented):
To pass the docs CI you'll also need to run `cargo xtask docgen` and commit the results

NomisIV:
Thank you for the review! I have now fixed the issues, and CI should pass. :)

the-mikedavis (approved):

pascalkuthe (approved):

PR #8308, opened 2023-09-16
Add `.editorconfig` highlighting as INI

yoav-lavi:
As discussed in #8076 (https://github.com/helix-editor/helix/pull/8076#issuecomment-1722254250), this PR adds syntax highlighting for `.editorconfig` files as INI temporarily until Helix supports them directly.

While there is a difference between `.editorconfig` and INI syntax, it is very minor (see below) and most files are likely to highlight correctly, which should be a better compromise that not highlighting at all.

> EditorConfig files use an INI format that is compatible with the format used by Python ConfigParser Library, but [ and ] are allowed in the section names. The section names are filepath globs, similar to the format accepted by gitignore. Forward slashes (/) are used as path separators and semicolons (;) or octothorpes (#) are used for comments. Comments should go individual lines. EditorConfig files should be UTF-8 encoded, with either CRLF or LF line separators.

cc @the-mikedavis

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2290,7 +2290,8 @@ file-types = [
   "container",
   "volume",
   "kube",
-  "network"
+  "network",
+  ".editorconfig"
 ]
 injection-regex = "ini"
 comment-token = "#"


the-mikedavis (approved):

pascalkuthe (approved):

yoav-lavi:
As far as I can tell these are the two cases in which `[]` will be used:

> [name] | Matches any single character in name
>  [!name] | Matches any single character not in name

It appears that in these cases string highlighting will break for the specific section that used this (they'll be the same color as keys) but otherwise the result is still coherent and much better IMO than no highlighting (see images below)
 
Screenshot 2023-09-16 at 23 16 48
Screenshot 2023-09-16 at 23 16 42

PR #8320, opened 2023-09-18
Picker movements

davxy:
`CTRL+j` and `CTRL+k` to move up and down are - IMO - a bit less surprising for users coming from other modal editors or tools like fzf.

Any drawback?


archseer:
#1792

PR #8349, opened 2023-09-21
Fix C++ highlight not working

nuid64:
Everything works on my laptop, but, for some reason, on my PC there is treesitter error because of `nullptr` is not in `""`

### DIFF
--- a/runtime/queries/cpp/highlights.scm
+++ b/runtime/queries/cpp/highlights.scm
@@ -47,7 +47,7 @@
 ; Constants
 
 (this) @variable.builtin
-(nullptr) @constant.builtin
+("nullptr") @constant.builtin
 
 ; Keywords
 


nuid64:
Well, check failed, ***but on my machine*** it passed the test. Probably it has something to do with the fact, that highlight only works on one of my machines

the-mikedavis:
This is an issue with your installation rather than the queries in master: the CI tests the queries against the parsers as locked in `languages.toml` and your machine must have an older version of the C++ parser. That can happen if you have a stale `[[grammar]]` section specified in a custom `languages.toml` (`~/.config/helix/languages.toml` or a local `.helix/languages.toml`), outdated queries in your runtime folder, or you may need to run `hx -g fetch` and `hx -g build`.

PR #8354, opened 2023-09-22
add command history picker

jrvidal:
Adds a simple command history picker:

image

cf. https://github.com/helix-editor/helix/discussions/7617

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -310,6 +310,7 @@ impl MappableCommand {
         code_action, "Perform code action",
         buffer_picker, "Open buffer picker",
         jumplist_picker, "Open jumplist picker",
+        command_history_picker, "Open command history picker",
         symbol_picker, "Open symbol picker",
         select_references_to_symbol_under_cursor, "Select symbol references",
         workspace_symbol_picker, "Open workspace symbol picker",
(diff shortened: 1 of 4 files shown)


rockboynton:
I like this, although I wish it didn't duplicate commands

helix-term/src/commands.rs
```diff
@@ -2926,6 +2927,42 @@ fn buffer_picker(cx: &mut Context) {
     cx.push_layer(Box::new(overlaid(picker)));
 }
 
+fn command_history_picker(cx: &mut Context) {
+    let command_history = cx.editor.registers.read(':', cx.editor);
+
+    let items: Vec<Command> = command_history
```
rockboynton, line 2933: Perhaps use a set? A [`BTreeSet`](https://doc.rust-lang.org/stable/std/collections/struct.BTreeSet.html) to rather than HashSet to maintain ordering

jrvidal: I think a BTreeSet wouldn't help, b/c we need insertion order, not lexicographical order.

jrvidal:
@the-mikedavis Sorry for the ping, I know y'all are busy. But this is a pretty self-contained, small addition. I'd be happy to withdraw it if it is not something you deem appropriate for core.

the-mikedavis:
I think this could be covered more generally by a register history feature like #11184. You could pick from the history of the `:` register and then use `:<ret>` to run that command

PR #8362, opened 2023-09-23
Better bufferline

CedricMeu:
This PR ads a way to change the behaviour of the bufferline on overflow.
This introduces a breaking change to the bufferline configuration, which now looks like

```toml
[editor.bufferline]
show = "always"
style = "scroll"
```

`editor.bufferline.show` accepts any of `always`, `multiple` or `never`, and `editor.bufferline.style` any of 

- `overflow`, original behaviour, let it overflow to the right and don't care.
- `wrap`, on overflow, expand downwards and wrap.
- `scroll`, on overflow, overflow and scroll the active buffer's tab as centred as possible, essentially scrolling the bufferline.

For now `overflow` is the default behaviour, but I'm unsure if this is a good default setting.
as alternative I'd propose `scroll` as this handles the bufferline similar to overflow, having a fixed height of one but adding the benefit of always showing what your active buffer is.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -59,12 +59,26 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `true-color` | Set to `true` to override automatic detection of terminal truecolor support in the event of a false negative | `false` |
 | `undercurl` | Set to `true` to override automatic detection of terminal undercurl support in the event of a false negative | `false` |
 | `rulers` | List of column positions at which to display the rulers. Can be overridden by language specific `rulers` in `languages.toml` file | `[]` |
-| `bufferline` | Renders a line at the top of the editor displaying open buffers. Can be `always`, `never` or `multiple` (only shown if more than one buffer is in use) | `never` |
 | `color-modes` | Whether to color the mode indicator with different colors depending on the mode itself | `false` |
 | `text-width` | Maximum line length. Used for the `:reflow` command and soft-wrapping if `soft-wrap.wrap-at-text-width` is set | `80` |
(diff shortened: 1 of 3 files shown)


the-mikedavis:
Rather than making the rendering configurable I would prefer we be opinionated and pick a sane default and have no config. IMO `scroll` is the best presentation, plus it aligns with other UIs like browsers.

CedricMeu:
> Rather than making the rendering configurable I would prefer we be opinionated and pick a sane default and have no config. IMO `scroll` is the best presentation, plus it aligns with other UIs like browsers.

Like I mentioned, I am in favour of using `scroll` as the default. 
Also `overflow` could be removed as no sane person would favour that over `scroll` nor `wrap`.
I am however hesitant on removing the configuration for the behaviour as other people might like wrapping over scrolling.
If we remove the option, someone down the line will probably reimplement it.

CedricMeu:
I added markers on the left and right to indicate under and overflow as per #10868.
I used the `ui.bufferline` style as default for these, but am considering to add a style for these specifically. That should help differentiate them from the actual buffer tabs.

dvdvgt:
Thanks for your continued work on this. I just tested this locally, but I cannot observe any scrolling behaviour by default. By the previous [comments](https://github.com/helix-editor/helix/pull/8362#issuecomment-1732591390), I would have expected that this just works.

CedricMeu:
> Thanks for your continued work on this. I just tested this locally, but I cannot observe any scrolling behaviour by default. By the previous [comments](https://github.com/helix-editor/helix/pull/8362#issuecomment-1732591390), I would have expected that this just works.

Scrolling behaviour is default and non-configurable.
In order to develop locally, I just use `cargo run` and that seems to work for me.

PR #8366, opened 2023-09-24
`:trim` to remove trailing whitespace

### LINKED ISSUE #1481: Add `trim_file` command and `trim_file_on_save` configuration option

I know there is an `trim_selection` command but using it breaks the current selection, which is a not very helpful. It's also not very practical to do it for each opened file when ending a somewhat long session.

kirawi:
Supersedes #3674 and resolves #1481

`:trim` will remove both trailing empty (lines consisting of only whitespace characters) and trailing spaces on lines. The superseded PR did this as well, but the code became harder to read/reason about (for me) while trying to keep everything in a single transaction, as it allowed you to choose to trim either spaces, lines, or both. This implementation will always trim both.

Edit: I just noticed that it would actually be trivial to reintroduce those options because I ended up implementing this differently than before. It would just be adding `lines` (would replace var `trailing`) and `spaces` as arguments and conditionals to the existing if statements. I won't be doing that though.

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -88,3 +88,4 @@
 | `:move`, `:mv` | Move the current buffer and its corresponding file to a different path |
 | `:yank-diagnostic` | Yank diagnostic(s) under primary cursor to register, or clipboard by default |
 | `:read`, `:r` | Load a file into buffer |
+| `:trim-trailing-whitespace`, `:trim` | Delete whitespace |
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2520,6 +2520,97 @@ fn read(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> 
     Ok(())
 }
 
+fn trim_whitespace_impl(doc: &Document, selection: &Selection) -> Transaction {
+    /// Find the last non-whitespace char of a line
+    fn find_trailing(l: RopeSlice) -> Option<usize> {
+        // Handle empty docs
+        if l.len_chars() == 0 {
+            return None;
+        }
+
+        // Returns the left-wise beginning of the trailing whitespace
+        // It is +1 the index of that char so that char is not deleted
+        l.chars_at(l.len_chars())
+            .reversed()
+            .position(|ch| !ch.is_whitespace())
+            .map(|n| l.len_chars() - n)
+            .or(Some(0))
+    }
+
+    let mut deletions: Vec<helix_core::Deletion> = Vec::new();
+    let mut delete = |start, end| {
+        // Don't push empty changes
+        if start != end {
+            deletions.push((start, end));
+        }
+    };
+
+    // Assume ranges are in order and not overlapping
+    for range in selection.ranges().iter().rev() {
+        let slice = range.slice(doc.text().slice(..));
+        let lines = slice.lines_at(slice.len_lines()).reversed();
+
+        // Cap the `end` to not delete the line ending
+        let end_account_le = |line: RopeSlice, n: usize| {
+            let le_len = helix_core::line_ending::get_line_ending(&line)
+                .map(|le| le.len_chars())
+                .unwrap_or(0);
+            // Map `end` with respect to the whole doc
+            range.from() + n - le_len
+        };
+
+        // Ignore empty lines if `trailing` is true.
+        // If not `trailing`, delete trailing whitespace on lines.
+        let mut trailing = true;
+        for (idx, line) in lines
+            .enumerate()
+        
+        {
+            // This subtraction cannot underflow because len_lines must be `> 0` to enter
+            // this loop and `idx` is always `< len_lines`.
+            let line_number = slice.len_lines() - 1 - idx;
+            if trailing {
+                // `n @ 1..` will ignore `Some(0)` from empty lines
+                if let Some(n @ 1..) = find_trailing(line) {
+                    let start = range.from() + slice.line_to_char(line_number) + n;
+                    // Needed to retain the last EOL of the selection, which would be selected. e.g. in `%:trim`
+                    let end = end_account_le(slice, slice.len_chars());
+                    delete(start, end);
+                    trailing = false;
+                }
+            } else if let Some(n) = find_trailing(line) {
+                let start = range.from() + slice.line_to_char(line_number) + n;
+                let end = end_account_le(line, slice.line_to_char(line_number + 1));
+                delete(start, end);
+            }
+        }
+
+        // Delete empty selections
+        if trailing {
+            let start = range.from();
+            let end = end_account_le(slice, slice.len_chars());
+            delete(start, end);
+        }
+    }
+    Transaction::delete(doc.text(), deletions.into_iter().rev())
+}
+
+fn trim_whitespace(
+    cx: &mut compositor::Context,
+    _args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+    let (view, doc) = current!(cx.editor);
+    let selection = doc.selection(view.id);
+    let tx = trim_whitespace_impl(doc, selection);
+    doc.apply(&tx, view.id);
+    doc.append_changes_to_history(view);
+    Ok(())
+}
+
 pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
     TypableCommand {
         name: "quit",
@@ -3140,6 +3231,14 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
         doc: "Load a file into buffer",
         fun: read,
         signature: CommandSignature::positional(&[completers::filename]),
+    },
+    TypableCommand {
+        name: "trim-trailing-whitespace",
+        aliases: &["trim"],
+        doc: "Delete trailing whitespace from the current selections",
+        fun: trim_whitespace,
+        signature: CommandSignature::none(),
+        
     },
 ];
 


helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2408,82 @@ fn redraw(
...
+                    let end = range.from() + r.len_chars() - le_len;
+                    changes.push((start, end, None));
+                    trailing = false;
+                }
+            } else if let Some(n) = find_trailing(l) {
+                let start = range.from() + r.line_to_char(i) + n;
+                let le_len = helix_core::line_ending::get_line_ending(&l)
+                    .unwrap()
+                    .len_chars();
+                // TODO: prove this is safe
```
kirawi, line 2461: To elaborate on this comment, I wasn't sure if the line could underflow.

Joe-Zer0:
I would very much like a config flag to run this command on write. Is that something you would consider adding?

kirawi:
I'll think about it, but at the moment I chose not to so the PR would be simple.

pascalkuthe:
That should be a seperate PR and likely based on #8021. Maybe it doesn't even need a special config flag and instead we can allow users to specify a list of commands to run pre-write

Joe-Zer0:
Sounds good. Thanks for the PR by the way. It's a feature I'm really looking forward to. I was actually attempting to do the same thing based on the code from your last PR. But it was slow going as I was trying to learn Rust along the way :)

lelgenio:
What about removing leading lines? When working with bigger expressions I tend to add a few empty lines around it to better focus on it, `:trim` could be useful to clean-up in such cases if it was able to remove leading lines.

Before:

```rust
Animals::find().filter(AnimalType::Horse).for_each(|horse|{
    horse.feed(
// ⬅️ selection-start

    
      CarrotBuilder::new()
        .color(Color::Orange)
        .size(Size::Big)
        .taste(Taste::Good)
        .build()


// ⬅️ selection-end
    );
});

```

After `:trim`:

```rust
Animals::find().filter(AnimalType::Horse).for_each(|horse|{
    horse.feed(
      CarrotBuilder::new()
        .color(Color::Orange)
        .size(Size::Big)
        .taste(Taste::Good)
        .build()
    );
});

```

To me it also makes sense to remove leading spaces:


Before:

```rust
horse.feed(⬇️      Carrot::new()  ⬇️);
```

After `:trim`:

```rust
horse.feed(Carrot::new());
```
image

helix-term/src/commands/typed.rs
```diff
@@ -3008,6 +3105,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
         fun: redraw,
         signature: CommandSignature::none(),
     },
+    TypableCommand {
+        name: "trim",
+        aliases: &[],
+        doc: "Trim whitespace",
```
lelgenio, line 3111: The description of the command could be more specific(longer), until you try to use the command it's not clear whether it trims the buffers' content, or just the selection.

Also, the word "trim" is used with another meaning than in `trim_selection`, how could we differentiate these two?
* trim: to move the selection anchors so that the selection no longer includes undesirable characters, but these characters still exist.
* trim: to delete characters from the buffer, if they are undesirable.

kirawi:
That's more or less the job of a formatter. The cases you describe are going to be affected by how the language is defined. For instance, how can you determine whether to trim leading spaces or not when the user might want to keep leading spaces in strings, e.g. `"  Hello, world!"` Or in the case of leading empty lines, whether it fits the coding style? At the point you start going down a rabbit hole of exceptions and cases until you essentially end up with a formatter.

If you really want to do the cases you're talking about, you can split your selection to empty lines and leading spaces and then `d`. `:trim` is not necessary in that case.

book/src/generated/typable-cmd.md
```diff
@@ -85,3 +85,4 @@
 | `:reset-diff-change`, `:diffget`, `:diffg` | Reset the diff change at the cursor position. |
 | `:clear-register` | Clear given register. If no argument is provided, clear all registers. |
 | `:redraw` | Clear and re-render the whole UI |
+| `:trim` | Trim whitespace |
```
lelgenio, line 88: This line does not match what's in `helix-term/src/commands/typed.rs`

leandrobbraga:
> That's more or less the job of a formatter. The cases you describe are going to be affected by how the language is defined. For instance, how can you determine whether to trim leading spaces or not when the user might want to keep leading spaces in strings, e.g. `" Hello, world!"` Or in the case of leading empty lines, whether it fits the coding style? At the point you start going down a rabbit hole of exceptions and cases until you essentially end up with a formatter.
> 
> If you really want to do the cases you're talking about without `:fmt`, you can split your selection to empty lines and leading spaces and then `d`. `:trim` is not necessary in that case.

Since we want to remove only trailling whitespaces, shouldn't we call this :rtrim:? It would be less surprising since most languages I know contains a triplet of funtions: `trim`, `ltrim` and `rtrim`. The `trim` usually strips from both directions.

leandrobbraga:
> That's more or less the job of a formatter. The cases you describe are going to be affected by how the language is defined. For instance, how can you determine whether to trim leading spaces or not when the user might want to keep leading spaces in strings, e.g. `" Hello, world!"` Or in the case of leading empty lines, whether it fits the coding style? At the point you start going down a rabbit hole of exceptions and cases until you essentially end up with a formatter.
> 
> If you really want to do the cases you're talking about without `:fmt`, you can split your selection to empty lines and leading spaces and then `d`. `:trim` is not necessary in that case.

Since we want to remove only trailling whitespaces, shouldn't we call this :rtrim:? It would be less surprising since most languages I know contains a triplet of funtions: `trim`, `ltrim` and `rtrim`. The `trim` usually strips from both directions. 

See:
rust: https://doc.rust-lang.org/std/primitive.str.html#method.trim
javascript: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim
python: https://docs.python.org/3/library/stdtypes.html (called strip)

the-mikedavis:
I don't forsee us adding a trim-leading-whitespace feature so "rtrim"/"rstrip" seems unnecessary to me. If we want to be really clear let's call the command `:trim-trailing-whitespace` and alias it to `:trim`.

kirawi:
I've been running this PR for a few months now. While I don't use it very often, it hasn't caused any issues for me so far.

kirawi:
I'm wondering if we want this or a way to select all trailing whitespace.

tgross35:
> I'm wondering if we want this or a way to select all trailing whitespace.

Deleting trailing whitespace seems significantly more common than needing to otherwise manipulate trailing whitespace - for anyone who does need to manipulate, I would think that just using `\W+$` is suitable.

Could `:trim` be added and, if it turns out there is a valid usecase for selecting trailing whitespace, just turn `:trim` into an alias for select+delete?

the-mikedavis:
Yeah I wonder if we really need a typable command for this given you can already select it with `%s[ \t]+$` (could be bound to a macro with #4709) and then delete. Automatically trimming whitespace on save could be a nice addition though I think, similar to #8157

kirawi:
Regex that approaches the behavior of this command is something like: `(?-m)[ \t]+\n|\s+$`. It's not perfectly accurate though since it will always remove the trailing line (EOF).

ArthaTi:
Is there any particular reason why this hasn't been merged yet?

tgross35:
@the-mikedavis (or other maintainers) is this something that could be added to the `next` milestone? Even though there are workarounds (e.g. what is possible with https://github.com/helix-editor/helix/pull/4709), having this out of the box is such a nice quality of life thing. Especially working with any file types that don't have autoformatters.

helix-term/src/commands/typed.rs
```diff
@@ -2483,6 +2483,101 @@ fn move_buffer(
...
+            // Map `end` with respect to the whole doc
+            range.from() + n - le_len
+        };
+
+        // Ignore empty lines if `trailing` is true.
+        // If not `trailing`, delete trailing whitespace on lines.
+        let mut trailing = true;
+        for (idx, line) in lines
+            .enumerate()
+            .map(|(idx, line)| (map_reverse(slice.len_lines(), idx), line))
```
the-mikedavis, line 2537: This is a bit confusing. Instead of the `map` here to adjust the `idx` let's use a separate binding:

```rust
for (idx, line) in lines.enumerate() {
    // This subtraction cannot underflow because len_lines must be `> 0` to enter
    // this loop and `idx` is always `< len_lines`.
    let line_number = len_lines - 1 - idx;
```

Also we might as well bind `len_lines = slice.len_lines()` above since we use it a few times in this block

helix-term/src/commands/typed.rs
```diff
@@ -2483,6 +2483,101 @@ fn move_buffer(
...
+    let mut deletions: Vec<helix_core::Deletion> = Vec::new();
+    let mut delete = |start, end| {
+        // Don't push empty changes
+        if start != end {
+            deletions.push((start, end));
+        }
+    };
+
+    // Assume ranges are in order and not overlapping
+    for range in selection.ranges().iter().rev() {
```
the-mikedavis, line 2519: I wonder if instead of operating on each selection's range we should use `Selection::line_ranges` like we do for `:reset-diff-change` now. I don't think it ever makes sense to not operate on full lines.

helix-term/src/commands/typed.rs
```diff
@@ -2483,6 +2483,101 @@ fn move_buffer(
     Ok(())
 }
 
+fn trim_whitespace_impl(doc: &Document, selection: &Selection) -> Transaction {
+    /// Maps reverse index to non-reverse index
+    fn map_reverse(len: usize, rev: usize) -> usize {
```
Rudxain, line 2488: This looks like it could be a useful utility/helper somewhere else. Why was it defined locally? Is it to make the `fn` easier to call, as `len: usize` is more convenient than `len: NonZero<usize>`?

helix-term/src/commands/typed.rs
Rudxain: I might have misunderstood the code or this review, but here are the pros & cons that come to mind right now:
- Trimming selections is much more versatile, but comes at the cost of hitting `x` before the cmd (if we want to trim the full line)
- Full-line-trim seems more common (in practice), so it should have a shorter key-sequence. But (mathematically) it's just 1 special case among an `usize::MAX` ("pseudo-Infinity") of potential selection-trims. Therefore, full-line-trim is rare (in theory)

Although, it seems fallacious to believe that FLT is common, simply because practically-all editors only support trimming all lines in a buffer, thereby forcing users to get used to this limitation

the-mikedavis: For a bit of a contrived example, say we have a file like:

```
····[········ffffffffffff···············
············ffffffffffff···············
············ffffffffffff·······]········
```

(with "·" being a space character and the selection being represented by the brackets)

I think `:trim` here should operate on all lines covered by the selection, resulting in

```
············ffffffffffff
············ffffffffffff
············ffffffffffff
```

where I think currently this PR would end up with

```
············ffffffffffff
············ffffffffffff
············ffffffffffff·······
```

I'm not sure it's ever useful to use the end of the selection range as the end of the area you're trimming

Rudxain: I agree. Now I see why it doesn't seem useful.

> ```
> ············ffffffffffff
> ············ffffffffffff
> ············ffffffffffff·······
> ```

In my mind, I thought it would be:
```
············ffffffffffff············
············ffffffffffff············
············ffffffffffff·······
```
Which makes less sense, I guess?

Then I thought "what if there were non-space code-points somewhere between the consecutive spaces?", then I answered my own question:

In practice, typically, users want 0, 1, or (rare) 2 spaces between non-space chars (let's ignore indentation, for simplicity). So, why would users need to specify how many spaces they want to delete? It'd be [much simpler](https://en.wikipedia.org/wiki/Zero_one_infinity_rule) to:
- delete all spaces from all lines
- delete all spaces from all lines, except 1 space from the last line
- preserve some lines as-is

omentic:
Does anyone here know how to make a binding to trim trailing whitespace on write? The best I've got is `["select_all", ":trim", "collapse_selection", ":write"] `, but this moves the cursor position to the end of the file.

pluiedev:
Any updates? It's really quite annoying to have CI complain about trailing whitespace and having to manually fix it line-by-line

kirawi:
A little busy right now. The comment I still need to address wasn't trivial since there isn't an equivalent of `Selection::line_ranges` for going backwards in the way I needed it to. Or at least I think so, it's been a while so I don't exactly recall what I encountered. It's not difficult but it was enough to have me put this to the side for now. I wouldn't be opposed if anyone wants to take this PR over the finish line themselves though.

senekor:
In case anyone's interested, I've been working on [ec2hx](https://github.com/senekor/ec2hx) to add the semblance of EditorConfig support to Helix. As part of that, `ec2hx` now ships with a builtin formatter that just trims trailing whitespace. Even if you are not interested in EditorConfig support, you may want to use this formatter to get just the trimming functionality.

There are [pre-built binaries](https://github.com/senekor/ec2hx?tab=readme-ov-file#usage) for `ec2hx`. Once installed, you can add the following formatter config to a language of your choice:

```toml
formatter = { command = "ec2hx", args = ["trim-trailing-whitespace"] }
auto-format = true
```

You can also run `ec2hx` in a directory with a `.editorconfig` file with the following content:
```ini
[*]
trim_trailing_whitespace = true
```

That will generate the above formatter configuration for all languages which don't already have a formatter or LSP configured.

---

I hope this is useful to someone.
It works well for me on Linux and I have thought about Windows line breaks, but that part isn't tested.
It essentially just tries to preserve whatever line breaks the file already has.

kirawi:
Replaced by #15907

PR #8371, opened 2023-09-24
Use is_ascii_graphic for checking auto completion words

ItsEthra:
Before, auto completion would reset if one types a character that is not `[a-zA-Z_]` which is really annoying especially when working with html,css and tailwindcss-language-server.
Example:
```css
a {
}
```
Put cursor between braces and type `background-color`. When you type in the hyphen, completion results would reset and it will only show completion for `color`. This pr fixes this behavior.

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -61,8 +61,8 @@ pub enum OffsetEncoding {
 pub mod util {
     use super::*;
     use helix_core::line_ending::{line_end_byte_index, line_end_char_index};
-    use helix_core::{chars, RopeSlice, SmallVec};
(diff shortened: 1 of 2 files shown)


gabydd:
This will be fixed by https://github.com/helix-editor/helix/pull/8021 I believe (cc: @pascalkuthe )

pascalkuthe:
it will help wi(th that but not entirely fix it. The way we filter completion is currently not ideal. IIRC correctly the discussion in the LSP issue  tracker concluded that you should be using the replacement range from the LSP transaction (plus any new adjacent text).

Changing the definition of word characters for the entire editor is definitely not the right approach tough. We will not change that.

ItsEthra:
I see, I'll close this pr then until I have a better implementation.

PR #8374, opened 2023-09-24
Filter out language servers which fail to spawn

### LINKED ISSUE #8361: Language servers fail if one is missing
### Summary

If one of the language server from `language-servers` list is missing, none of the language servers work.

### Reproduction Steps

Configure multiple language servers

<details>
<summary>~/.config/helix/languages.toml</summary>

```toml
[[language]]
name = "jsx"
auto-format = true
language-servers = ["typescript-language-server", "biome", "emmet"]
formatter = { command = "biome", args = ["format", "--stdin-file-path", "test.jsx"]}
```
</details>

Expected behavior:
The existing language server should work even if other language servers are missing.

What happened:
The language server process is spawned but it doesn't work.

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

foot version: 1.15.3

### Helix Version

helix 23.05 (1c88432e)

woojiq:
Return `Result<Arc<Client>>` instead of `Arc<Client>` so that we can then filter only successfully started language servers and create an error log otherwise.

Closes #8361

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -765,19 +765,22 @@ impl Registry {
                         return (name.to_owned(), Ok(client.clone()));
                     }
                 }
-                let client = match self.start_client(
+                match self.start_client(
                     name.clone(),
                     language_config,
                     doc_path,
                     root_dirs,
                     enable_snippets,
                 ) {
-                    Ok(client) => client,
-                    Err(err) => return (name.to_owned(), Err(err)),
-                };
-                let clients = self.inner.entry(name.clone()).or_default();
-                clients.push(client.clone());
-                (name.clone(), Ok(client))
+                    Ok(client) => {
+                        self.inner
+                            .entry(name.to_owned())
+                            .or_default()
+                            .push(client.clone());
+                        (name.clone(), Ok(client))
+                    }
+                    Err(err) => (name.to_owned(), Err(err)),
+                }
             },
         )
     }
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -755,7 +755,7 @@ impl Registry {
         doc_path: Option<&std::path::PathBuf>,
         root_dirs: &[PathBuf],
         enable_snippets: bool,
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
This is currently nice and minimal but I think we should do a slightly larger refactor here since `Registry::get` is forcing us to collect twice and it's making the call site in Editor awkward. (Although the Option handling there is also awkward, so I say let's fix that in this patch too.)

Let's make `Registry::get` return an `impl Iterator<Item = (LanguageServerName, Result<Arc<Client>>)>`. Then when launching the language servers we can filter out the Errs as you already do in this patch, collect into the HashMap, and return early when `language_servers` is empty.

That `launch_language_servers` function is returning `Option<()>` and bubbling it up to `refresh_language_servers` but the return value is never used. I think the option is very confusing - I don't really understand the distinction between `Some(())` and None - so let's remove those return types on `refresh_language_servers` and `launch_language_servers`. Each of the `?` operators can be replaced with `let Some(..) = .. else { return };`s

woojiq:
Done, thanks for the suggestion to return an iterator instead of collecting twice) The diff is a bit crazy because I moved the `if` content outside, so it looks like I overwrote everything.
We needed to unify all lifetimes in `get` so that we can use `move` in the closure (without `move` - error that closure will outlive borrowed `enable_snippets`) and also to return the iterator.

the-mikedavis (dismissed):

pascalkuthe (commented):
this looks great, thnaks for working on this! I agree this should land before the release. This is also a nice cleanup.

I found one minor nit, otherwise this LGTM

woojiq:
Done. 

Off-topic: _Do I need to write comments like "Done" after the suggested changes? Do you get notified about new commits of reviewed code?_

pascalkuthe (approved):
you don't have to write it, I get GitHub notifications when somebody pushes to a PR. In cases like this where its just a small change missing before merge I would check myself (but it doesn't hurt if you leave the comment either). For larger PRs ts nice if you do leave some comments, a push may not necessarily indicate they are done and I often skip those notifications (since I get a lot of those).

the-mikedavis (approved):

PR #8385, opened 2023-09-25
Style Bold/Italic/Strikethrough markdown in docs

A-Walrus:
Style doc comment markdown correctly with Bold / Strikethrough / Italics.
**Before:**
image

**After:**
image

Note the Italics of the word `do`


the-mikedavis (dismissed):

pascalkuthe (commented):
Changes look fine but I think the two nested match blocks are a bit hard to read. I think having one match arm for each recognized tag is fine (so simply text_style.add_modified(..) in each arm of the outer match)

the-mikedavis (approved):

pascalkuthe (approved):

Issue #7689, opened 2023-07-20
Helix 23.05 (d52b7903) freezes editing markdown


davidthewatson:
### Summary

I've been running the git release of Helix on arch for a couple years now. My current install is on Garuda Linux:
        
         ╭─watson@acer in ~ took 8ms
         ╰─λ pamac info helix-git
        Name                  : helix-git
        Version               : 23.05.242.gd52b79037-1
        Description           : A text editor written in rust
        URL                   : https://helix-editor.com
        Licenses              : Unknown
        Repository            : chaotic-aur
        Installed Size        : 144.7 MB
        Groups                : --
        Depends On            : --
        Optional Dependencies : --
        Required By           : --
        Optional For          : --
        Provides              : hx
        Replaces              : --
        Conflicts With        : helix
        Packager              : UFSCar HPC Builder <hpc.ufscar@chaotic.cx>
        Build Date            : Wed 19 Jul 2023 08:09:51 PM EDT
        Install Date          : Thu 20 Jul 2023 05:07:40 AM EDT
        Install Reason        : Explicitly installed
        Validated By          : Signature
        Backup files          : --

I've been editing markdown all day, every day for a while and helix started freezing recently where all keystroke processing ceases. Nothing seems to release it, esc, colon, etc.

I've taken to `pgrep hx` follwed by `pkill hx` at the command line to resolve and restart helix.

### Reproduction Steps

 

I tried this:

1. `hx`
2. edit files with md extension
3. wait for freeze within a few minutes

I expected this to happen:

No frezee.

Instead, this happened:

Freeze.

### Helix log

I'll provide the verbose log when I get it.

<details><summary>~/.cache/helix/helix.log</summary>

```
 ╭─watson@acer in ~ took 16ms
 ╰─λ cat ~/.cache/helix/helix.log
/home/watson/.cache/helix/helix.log
2023-07-13T08:32:07.058 helix_view::editor [ERROR] Failed to initialize the language servers for `source.md` { cannot find binary path }
2023-07-13T08:32:56.575 helix_view::editor [ERROR] Failed to initialize the language servers for `source.md` { cannot find binary path }
2023-07-13T08:36:45.385 helix_view::editor [ERROR] Failed to initialize the language servers for `source.md` { cannot find binary path }
2023-07-19T12:19:55.671 helix_view::editor [ERROR] Failed to initialize the language servers for `source.bash` { cannot find binary path }
[...]

PR #8396, opened 2023-09-26
Pin tree-sitter to the 0.20.10 release

the-mikedavis:
Packagers downstream sometimes struggle with git dependencies so we want to switch to the published crate for the 23.09 (Helix) release. We can revert this and update the pinned commit after the release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2483,7 +2483,8 @@ dependencies = [
 [[package]]
 name = "tree-sitter"
 version = "0.20.10"
-source = "git+https://github.com/tree-sitter/tree-sitter?rev=ab09ae20d640711174b8da8a654f6b3dec93da1a#ab09ae20d640711174b8da8a654f6b3dec93da1a"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


David-Else:
Sorry for my ignorance, but doesn't this mean the new Helix release will have an inferior 6 month old version of tree sitter? This Helix version will be the version the vast majority of people will be using for the next 3 or so months, so it would be quite a big deal?

the-mikedavis:
The only issue I know of with 0.20.10 that affects us is https://github.com/helix-editor/helix/issues/7689 so I don't see this being a big deal

pascalkuthe:
And we have been blocking the helix release on TS releasing for close to two months now. At some point, we just have to stop waiting. Once TS releases its next version we can do a patch release.

pascalkuthe (approved):

David-Else:
https://github.com/tree-sitter/tree-sitter/issues/2669#issuecomment-1736529430

> at the end of this week there will be a new tree-sitter release so I'll include this request in it.

archseer:
Now a month ago. We can't wait any longer so we'll merge this revert

PR #8400, opened 2023-09-26
Update tutor to show that helix does not have any recovery method yet

ardijanr:
Lost quite a bit of work, even though I went through the tutor. It would have been nice to be warned of this ahead of time. 

Up to you guys to determine if it is necessary, imo is that it would be nice to have some kind of warning somewhere.

### DIFF
--- a/runtime/tutor
+++ b/runtime/tutor
@@ -35,8 +35,10 @@
  CapsLock key is not pressed and hold the j key until you reach
  the first lesson.
 
-
-
+ PS:
+ Please note that helix does not have autosave enabled by default,
+ nor does helix have any method of recovering unsaved changes at
+ the moment. Please be aware of this and remember to write often.
 
 
 


the-mikedavis:
The tutor isn't a place to discuss features that aren't implemented yet

the-mikedavis:
IMO it's unfair to assume that Helix has a feature like this and to rely on it just because other tools in the same vein do

ardijanr:
What is fair has very little to do with what people expect from a tool. 
It is seriously the first thing i would expect to be implemented, much more so than multi cursors. 

I can safely say I am not the only one bitten by this. 

I suggested that it would be informed somewhere if not in the tutor, either in a todo list in the readme or something somewhere. Idk, omitting it seems dishonest imo. 

I somehow knew this would just be ignored and closed though.

Issue #8413, opened 2023-09-27
Always shown ruff-lsp code actions do nothing


bluthej:
### Summary

There are two  code actions provided by ruff-lsp that are always shown: 
- `Ruff: Organize Imports`
- `Ruff: Fix All`

Both of them do nothing when they are selected.

Note that when I select the `I` rule I have another code action that pops which is `Ruff (I001): Organize imports` which **does** work. It's the two code actions that are always shown that do nothing.

### Reproduction Steps

- I am building helix from source on the master branch (updated on the 27th of september 2023).

> Note that in my config I have both `pylsp` and `ruff-lsp` activated, but I tried running only `ruff-lsp` and it did not help.

<details><summary>Python lsp config</summary>
<p>

```toml
[language-server.ruff-lsp]
command = "ruff-lsp"
args = []

[[language]]
name = "python"
formatter = { command = "black", args = ["--quiet", "-"] }
auto-format = true
language-servers = ["ruff-lsp", "pylsp"]
```

</p>
</details> 

- I open a Python file with some unordered imports
- I open the code actions menu (`<space-a>`) and select the organize imports or the fix all code action
- Nothing happens

I opened an [issue](https://github.com/astral-sh/ruff-lsp/issues/249#issuecomment-1737193677) on the ruff-lsp repo and I tried the config kindly provided by [Charlie Marsch](https://github.com/charliermarsh) on the latest stable release. On this release, both code actions work, but on the master branch I cannot get them to work.

### Helix log

<details><summary>~/.cache/helix/helix.log (master branch)</summary>

```
2023-09-27T22:27:59.858 helix_lsp::transport [INFO] ruff-lsp -> {"jsonrpc":"2.0","method":"textDocument/codeAction","params":{"context":{"diagnostics":[{"code":"I001","data":{"fix":{"applicability":"Automatic","edits":[{"content":"import os\nimport sys\n\n\n","end_location":{"column":0,"row":3},"location":{"column":0,"row":1}}],"message":"Organize imports"},"noqa_row":1},"message":"Import block is un-sorted or un-formatted","range":{"end":{"character":0,"line":2},"start":{"character":0,"line":0}},"severity":2,"source":"Ruff"}],"triggerKind":1},"range":{"end":{"character":1,"line":0},"start":{"character":0,"line":0}},"textDocument":{"uri":"file:///tmp/ruff/test.py"}},"id":1}
[...]

PR #8421, opened 2023-09-29
LSP: Fix codeAction/resolve server capability check

the-mikedavis:
Previously we accidentally checked the server's _completion_ resolve capability rather than the code action resolve capability. It looks like this was an oversight introduced in #7677.

Fixes https://github.com/helix-editor/helix/issues/8413
Fixes https://github.com/astral-sh/ruff-lsp/issues/249


pascalkuthe (approved):
great catch!

Issue #8321, opened 2023-09-18
Haskell highlighting breaks when using syntax for injective type families


bramvdbogaerde:
### Summary

Note: I'm not sure if this is the right place to create this issue, since it is probably also tree-sitter related.

Helix does not highlight the syntax used for injective type families correctly. In fact, it causes the syntax highlighting for the remainder of the document to break.

Helix highlighting (incorrect):

<img width="391" alt="afbeelding" src="https://github.com/helix-editor/helix/assets/2016763/706f271c-a3e3-4071-a715-237c1d9029dc">

(note that everything starting from line 4 is highlighted incorrectly)

Neovim highlighting (correct):
<img width="362" alt="afbeelding" src="https://github.com/helix-editor/helix/assets/2016763/8b20481f-9caf-4110-af59-8667833ac30c">

### Reproduction Steps

N/A

### Helix log

N/A

### Platform

macOS

### Terminal Emulator

iTerm

### Helix Version

helix 23.05 (7f5940be)

PR #8435, opened 2023-10-01
fix: update haskell queries

nuid64:
#8321


archseer:
Updating the grammar isn't enough, you need to also update the queries as mentioned in the linked issue

nuid64:
Sorry, not competent in queries. Thought it would be a quick little change

PR #8436, opened 2023-10-01
Add alphanumerically sorting functions

### LINKED ISSUE #4153: Add numeric :sort and :rsort

leandrobbraga:
Closes #4153 

https://asciinema.org/a/e5E1jGM0mF6FMpVGPCE8B7qVo


leandrobbraga:
I didn't see that there was already an PR with an implementation

PR #8445, opened 2023-10-02
update with new mdbook index.hbs to fix sidebar

gabydd:
fixes the sidebar problems after we updated to the latest release of mdbook cc: @zqianem who was the last one to work on this part of the book

### DIFF
--- a/book/theme/index.hbs
+++ b/book/theme/index.hbs
@@ -1,11 +1,11 @@
 <!DOCTYPE HTML>
-<html lang="{{ language }}" class="sidebar-visible no-js {{ default_theme }}">
+<html lang="{{ language }}" class="{{ default_theme }}" dir="{{ text_direction }}">
     <head>
         <!-- Book generated using mdBook -->
         <meta charset="UTF-8">
         <title>{{ title }}</title>
         {{#if is_print }}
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
Odd, I see the odd sidebar behavior on https://docs.helix-editor.com/master/ but not when running locally 🤔

Also see https://github.com/rust-lang/mdBook/blob/master/src/theme/index.hbs

gabydd:
Hmm I was able to see it locally maybe you have an older version of mdbook locally I tested with the binary from the latest release. In terms of the upstream template this is just the upstream template with the colibri theme added as one of the theme options, which I believe was the only change for the older template as well

zqianem:
The changes look good to me.

It might be a good idea to [pin the mdBook version](https://github.com/helix-editor/helix/blob/7fbfec766c2ce2570ca4160744723813dbdc3019/.github/workflows/gh-pages.yml#L19) and document that `index.hbs` needs to be merged from upstream when updating the pin, to avoid breakages like this in the future. We would be able to unpin after https://github.com/rust-lang/mdBook/pull/2059 allows us to avoid overriding `index.hbs`.

PR #7387, opened 2023-06-19
highlight(sql): Update tree-sitter-sql to `98a7fc9`

LeoniePhiline:
# Changes:

Updates `tree-sitter-sql` [from `3a3f92b29c880488a08bc2baaf1aca6432ec3380` to  `c5b1548ec95956c8da6393dceacf30aa349a6e9f`](https://github.com/DerekStride/tree-sitter-sql/compare/3a3f92b29c880488a08bc2baaf1aca6432ec3380...c5b1548ec95956c8da6393dceacf30aa349a6e9f#commits_bucket).

[Full `grammar.js`](https://github.com/DerekStride/tree-sitter-sql/blob/c5b1548ec95956c8da6393dceacf30aa349a6e9f/grammar.js).

Note that to avoid further diff noise, the tree sitter JSON and C sources were moved onto the [`gh-pages`  branch](https://github.com/DerekStride/tree-sitter-sql/commits/gh-pages).

This is why `languages.toml` references https://github.com/DerekStride/tree-sitter-sql/commit/7cbac0472e5b8f8486ce64ffbcf1982d5cd5fc8d on the `gh-pages` branch, which contains the sources for https://github.com/DerekStride/tree-sitter-sql/commit/98a7fc965e455c38e508c1424671685dd6c610ab on the `main` branch ([the latest grammar change on `main`](https://github.com/DerekStride/tree-sitter-sql/compare/98a7fc965e455c38e508c1424671685dd6c610ab...c5b1548ec95956c8da6393dceacf30aa349a6e9f)).

## Merged PRs:

- https://github.com/DerekStride/tree-sitter-sql/pull/105
- https://github.com/DerekStride/tree-sitter-sql/pull/109
- https://github.com/DerekStride/tree-sitter-sql/pull/112
- https://github.com/DerekStride/tree-sitter-sql/pull/110
- https://github.com/DerekStride/tree-sitter-sql/pull/113
- https://github.com/DerekStride/tree-sitter-sql/pull/114
- https://github.com/DerekStride/tree-sitter-sql/pull/118
- https://github.com/DerekStride/tree-sitter-sql/pull/116
- https://github.com/DerekStride/tree-sitter-sql/pull/100
- https://github.com/DerekStride/tree-sitter-sql/pull/119
- https://github.com/DerekStride/tree-sitter-sql/pull/117
- https://github.com/DerekStride/tree-sitter-sql/pull/121
- https://github.com/DerekStride/tree-sitter-sql/pull/123
[...]

PR #8464, opened 2023-10-04
bump tree-sitter-sql

ds-cbo:
Bumps to https://github.com/DerekStride/tree-sitter-sql/commit/eeab7240a11098724e6f95bc57cc3ceaf5487d5f (the commit with the compiled grammar corresponding to `main` commit https://github.com/DerekStride/tree-sitter-sql/commit/385aff480f27bf69cce53453eede371ad30fff58)

Changes since previous revision (https://github.com/DerekStride/tree-sitter-sql/commit/7cbac0472e5b8f8486ce64ffbcf1982d5cd5fc8d, 19 June):

* https://github.com/DerekStride/tree-sitter-sql/pull/150
* https://github.com/DerekStride/tree-sitter-sql/pull/162
* https://github.com/DerekStride/tree-sitter-sql/pull/163
* https://github.com/DerekStride/tree-sitter-sql/pull/164
* https://github.com/DerekStride/tree-sitter-sql/pull/168
* https://github.com/DerekStride/tree-sitter-sql/pull/169
* https://github.com/DerekStride/tree-sitter-sql/pull/170
* https://github.com/DerekStride/tree-sitter-sql/pull/173
* https://github.com/DerekStride/tree-sitter-sql/pull/175
* https://github.com/DerekStride/tree-sitter-sql/pull/177
* https://github.com/DerekStride/tree-sitter-sql/pull/179
* https://github.com/DerekStride/tree-sitter-sql/pull/178
* https://github.com/DerekStride/tree-sitter-sql/pull/181
* https://github.com/DerekStride/tree-sitter-sql/pull/180
* https://github.com/DerekStride/tree-sitter-sql/pull/182
* https://github.com/DerekStride/tree-sitter-sql/pull/185
* https://github.com/DerekStride/tree-sitter-sql/pull/186
* https://github.com/DerekStride/tree-sitter-sql/pull/187
* https://github.com/DerekStride/tree-sitter-sql/pull/188
* https://github.com/DerekStride/tree-sitter-sql/pull/190
* https://github.com/DerekStride/tree-sitter-sql/pull/192
* https://github.com/DerekStride/tree-sitter-sql/pull/194
* https://github.com/DerekStride/tree-sitter-sql/pull/196
* https://github.com/DerekStride/tree-sitter-sql/pull/197
* https://github.com/DerekStride/tree-sitter-sql/pull/199

Also see: https://github.com/helix-editor/helix/pull/7387


runtime/queries/sql/highlights.scm
the-mikedavis, line 1: Were these copied from https://github.com/DerekStride/tree-sitter-sql/blob/main/queries/highlights.scm? We should update the existing queries incrementally instead, or adapt these queries to the Helix captures. These are in nvim-treesitter format so they use captures like `@boolean` rather than the Helix captures like `@constant.builtin.boolean`, `@parameter` rather than `@variable.parameter`, etc.. See the Helix set of highlight captures here: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

the-mikedavis (changes requested):

runtime/queries/sql/highlights.scm
ds-cbo: Ah, I indeed copied them blindly. I'll have a look at the Helix naming scheme, thanks

ds-cbo, line 1: Would you prefer me to make the diff easier by moving things around in the file to make the keywords align with the current highlights file, or would it be nicer to keep the order of upstream so that future changes upstream are easier to follow?

the-mikedavis: I think it would be nicer to follow the upstream highlights closely so that when they make changes to the highlights upstream it's clear where to make changes when we update the grammar. But I don't look at SQL much so I'm not sure if the queries we have in master are significantly better than the new ones in the repo (- looks like they were added since the last time we updated). If you think the new ones look good then let's use those (with the adjustments to the Helix capture names)

runtime/queries/sql/highlights.scm
```diff
@@ -1,343 +1,391 @@
...
-  "."
-] @punctuation.delimiter
+(literal) @string
+(comment) @comment.line
+(marginalia) @comment.block
 
-(binary_expression
-  operator: _ @operator)
+((literal) @constant.numeric.integer
+ (#lua-match? @constant.numeric.integer "^%d+$"))
```
the-mikedavis, line 37:
```suggestion
 (#match? @constant.numeric.integer "^\d+$"))
```

`#lua-match?` is nvim-specific. We use just the ones tree-sitter provides in the rust bindings https://tree-sitter.github.io/tree-sitter/using-parsers#predicates (though there are some there like `#any-of?` that we don't implement yet because we're waiting on a tree-sitter release that adds them)

runtime/queries/sql/highlights.scm
```diff
@@ -1,343 +1,391 @@
...
 
-(unary_expression
-  operator: _ @operator)
+((literal) @constant.numeric.float
+(#lua-match? @constant.numeric.float "^[-]?%d*\.%d*$"))
```
the-mikedavis, line 40: Same here - I think the regex from the old highlights should work in place of this one (I think the `%d` is lua-regex-specific)

runtime/queries/sql/highlights.scm
ds-cbo: Also resolved. Thanks for your patience, I'm new to this tree-sitter world so don't know many of the keywords/possibilities yet, but it looks interesting

pascalkuthe (approved):

PR #8471, opened 2023-10-06
Allow numbers as second input event

### LINKED ISSUE #8113: Allow numbers as second input event
I'm a long time emacs user and I'm used to split window with `C-x 3` (vertical) and `C-x 2` (horizontal).
Right now this config doesn't work:
``` toml
"C-x" = { "3" = "vsplit", "2" = "hsplit", "0" = "wclose" }
```
this is my setup for neovim:
``` lua
local map = function(mode, key, cmd, opts)
  vim.api.nvim_set_keymap(mode, key, cmd, opts or { noremap = true, silent = true })
end
map('n', '<C-x>0', '<C-w>q')
map('n', '<C-x>2', '<C-w>s')
map('n', '<C-x>3', '<C-w>v')
```
I know numbers are reserved for movements but being inspired by vim, I can't see why helix shouldn't be able to do the same.

areux:
This will allow using numbers as second key event. Closes #8113

### DIFF
--- a/helix-term/src/keymap.rs
+++ b/helix-term/src/keymap.rs
@@ -303,12 +303,13 @@ impl Keymaps {
         self.sticky.as_ref()
     }
 
-    pub fn key_exists(&self, mode: Mode, key: KeyEvent) -> bool {
+    pub fn contains_key(&self, mode: Mode, key: KeyEvent) -> bool {
         let keymaps = &*self.map();
         let keymap = &keymaps[&mode];
-        let mut ks = self.pending().to_vec();
-        ks.push(key);
-        keymap.search(&ks).is_some()
+        keymap
+            .search(self.pending())
+            .and_then(KeyTrie::node)
+            .is_some_and(|node| node.contains_key(&key))
     }
 
     /// Lookup `key` in the keymap to try and find a command to execute. Escape
--- a/helix-term/src/keymap.rs
+++ b/helix-term/src/keymap.rs
@@ -303,6 +303,14 @@ impl Keymaps {
         self.sticky.as_ref()
     }
 
+    pub fn key_exists(&self, mode: Mode, key: KeyEvent) -> bool {
+        let keymaps = &*self.map();
+        let keymap = &keymaps[&mode];
+        let mut ks = self.pending().to_vec();
+        ks.push(key);
+        keymap.search(&ks).is_some()
+    }
+
     /// Lookup `key` in the keymap to try and find a command to execute. Escape
     /// key cancels pending keystrokes. If there are no pending keystrokes but a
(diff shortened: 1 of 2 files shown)


helix-term/src/ui/editor.rs
```diff
@@ -886,7 +886,14 @@ impl EditorView {
...
         match (event, cxt.editor.count) {
             // count handling
-            (key!(i @ '0'), Some(_)) | (key!(i @ '1'..='9'), _) => {
+            (key!(i @ '0'), Some(_)) | (key!(i @ '1'..='9'), _)
+                if self.keymaps.pending().is_empty()
+                    || self.keymaps.pending().last()
+                        == Some(&KeyEvent {
+                            code: KeyCode::Char('g'),
+                            modifiers: KeyModifiers::NONE,
+                        }) =>
```
the-mikedavis, line 895: `g<n>g` can be remapped so this won't work robustly

the-mikedavis (changes requested):
I believe the solution for #8113 should be to look up the key event in the keymap before considering it for a count. If it's a possible keybinding (pending or command(s)) then it can just be used for the keybinding and if it isn't then it can be considered part of a count.

helix-term/src/keymap.rs
```diff
@@ -303,6 +303,14 @@ impl Keymaps {
         self.sticky.as_ref()
     }
 
+    pub fn key_exists(&self, mode: Mode, key: KeyEvent) -> bool {
```
the-mikedavis, line 306: To fit with standard data structures like hashmap let's call this `contains_key`

helix-term/src/keymap.rs
```diff
@@ -303,6 +303,14 @@ impl Keymaps {
...
+        let keymaps = &*self.map();
+        let keymap = &keymaps[&mode];
+        let mut ks = self.pending().to_vec();
+        ks.push(key);
+        keymap.search(&ks).is_some()
```
the-mikedavis, line 311: We can avoid the allocation of another vec for the pending keys here:

```suggestion
        keymap
            .search(self.pending())
            .and_then(KeyTrie::node)
            .is_some_and(|node| node.contains_key(&key))
```

the-mikedavis (changes requested):
Sorry for the delay, this fell of my radar for a while. I have only some minor nits.

areux:
Thanks for helping with this, I renamed the function and fixed the allocation issue.
I'm not familiar with helix source code, just followed what you said in your first comment and made these changes. I hope they're in the right places!

the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

PR #8475, opened 2023-10-06
Make parse_macro work for "-" outside "<..>"

bjorn-ove:
When running the integration tests, I ended up with temp file paths containing "-" in the name. This in turn triggered an issue where the key "" could not be resolved. This is one way of fixing this issue, another is to check for "-" in parse_macro() and replace it with "minus".

### DIFF
--- a/helix-view/src/input.rs
+++ b/helix-view/src/input.rs
@@ -549,7 +549,11 @@ pub fn parse_macro(keys_str: &str) -> anyhow::Result<Vec<KeyEvent>> {
         if c == ">" {
             keys_res = Err(anyhow!("Unmatched '>'"));
         } else if c != "<" {
-            keys.push(c);
+            if c != "-" {
+                keys.push(c);
+            } else {
+                keys.push("minus");
+            }
             i += end_i;
         } else {
             match s.find('>').context("'>' expected") {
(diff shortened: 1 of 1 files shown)


the-mikedavis (changes requested):
`-` isn't a valid part of a key sequence though, it's only used for connecting multiple keys into a chord like `C-c`, and `minus` stands for the `-` key. So parsing it into a KeyEvent or in parse_macro isn't correct.

Instead we could encode filenames as key sequences (so `a-b.txt` would become `a<minus>b.txt`). It might be possible to use `<C-r>%` to enter the buffer name into the prompt in the test cases instead though which would be simpler.

bjorn-ove:
I understand why `-` is not valid for KeyEvent::from_str, but not why it also applies to `parse_macro()`. In the latter case, `C-a` and similar is always inside `<..>`. So, I would passing `-` to `parse_macro()` is not ambiguous or anything like that.

For example `parse_macro("BigC-world")` in my head should be the letters as they are, while `parse_macro("Big<C-w>orld")` would be the way when `-` matters.

So, because of this it feels to me that the correct solution would be to check for `-` in `parse_macro()` and translate it to `minus`.

What am I missing here?

helix-view/src/input.rs
```diff
@@ -549,7 +549,11 @@ pub fn parse_macro(keys_str: &str) -> anyhow::Result<Vec<KeyEvent>> {
         if c == ">" {
             keys_res = Err(anyhow!("Unmatched '>'"));
         } else if c != "<" {
-            keys.push(c);
+            if c != "-" {
+                keys.push(c);
+            } else {
+                keys.push("minus");
```
the-mikedavis, line 555: Minor nit but let's swap these branches:

```rust
keys.push(if c == "-" { "minus" } else { c });
```

the-mikedavis (commented):
I was mostly worried about KeyEvent, it seems ok for parse_macro. I would prefer that converting to/from the key sequence format be round-trip, so parsing then encoding a key sequence would give the input string. Being able to use `-` in a macro should be very convenient though so I'm ok with the change here.

The solution I had in mind was to encode the path in terms of a key sequence, so `a-b.txt` would become `a<minus>b.txt`. Other than minus I don't think we have any special characters that are also valid in path names though so I don't think we really need that solution.

pascalkuthe (approved):

PR #8486, opened 2023-10-08
Handle suffixes of cancelled keymaps in insert mode

### LINKED ISSUE #2612: Issue with `[keys.insert]` mappings such as `j = { f = "normal_mode" }`
### Summary

I'm using `jf` to exit insert mode. When I insert anything that ends with `j` (not `jj`) and try to exit insert mode with `jf` I end up inserting `jf` and remain in `insert` mode. To overcome this I usually end up pressing escape key twice to exit insert mode _which is not ideal._

### Reproduction Steps

context:
I'm using `jf` to exit insert mode.

I tried this:

- tried to inset `hij` and exit insert mode with `jf`

I expected this to happen:

- insert `hij` and end up in normal mode

Instead, this happened:

- inset `hijjf` and end up in `insert` mode

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

gnome terminal

### Helix Version

helix 22.03 (d4e45fd4)

kmicklas:
Fixes #2612.

TODO: tests


helix-term/src/ui/editor.rs
```diff
@@ -857,28 +857,40 @@ impl EditorView {
     }
 
     fn insert_mode(&mut self, cx: &mut commands::Context, event: KeyEvent) {
-        if let Some(keyresult) = self.handle_keymap_event(Mode::Insert, cx, event) {
-            match keyresult {
-                KeymapResult::NotFound => {
-                    if let Some(ch) = event.char() {
-                        commands::insert::insert_char(cx, ch)
+        let mut queue = VecDeque::from([event]);
```
the-mikedavis, line 860: The VecDeque can be replaced with a regular Vec and you can just `Vec::pop`/`Vec::extend` to the end

pascalkuthe:
See also #6960 thus issue is not specific to insertmode and #6878

kmicklas:
> See also #6960 and #6878

Looking just at insert mode, I think the approach in this PR is more general than #6878. The problem is not specific to the last character in a sequence. That is only equivalent when the sequence only has length two.

> this issue is not specific to insertmode

"Fixing" this in normal mode seems like it could cause problems. If I cancel a key sequence in normal mode, I don't want it to try to execute some suffix of the key sequence, which could have arbitrary effect that is hard to predict.

Technically this concern exists in insert mode too, but there we have to balance it against the desire to support sequences like jk/fd. In insert mode we also have good reason to expect that the user wants to execute a "keymap suffix" - they just wanted to type the prefix as regular characters. This wouldn't make sense in normal mode.

helix-term/src/ui/editor.rs
kmicklas: Good point, in fact this is a stack and not sure why I thought it was going to be a queue.

PR #8494, opened 2023-10-09
Reword on goto_change shortcut keys

hongquan:
The old wording is confusing.


leandrobbraga:
I do believe that the idea of the previous naming is to be VCS agnostic

hongquan:
At first, I didn't understand what it means, until I found the blog post about Helix's Git support.
I think, in the future, when Helix supports more VCS, it can be named "Goto VCS change/hunk". For now, Helix only supports Git and it is more obvious to say "Go to Git hunk".

pascalkuthe (changes requested):
* we Plan to support other VCS in the future so Git should not be mentioned explicitl
* we will allow diffing against arbitrary files and Buffets (a PR is already open) so it should not mention VCS either

These were designed goals from the start which is why I choose current description. The most I would co sidereal is to use `diff change` instead of `change`.

These also aren't the rendered docs. Those should be adjusted to match

hongquan:
Thanks. I will close this PR and wait for Helix update.

Issue #8469, opened 2023-10-05
Opening helix with a directory as the first argument does not load .helix/config.toml correctly


bjorn-ove:
### Summary

The config is loaded before the working directory is changed during load, hence the wrong directory is checked for .helix/config.toml

https://github.com/helix-editor/helix/blob/68fce3e160c64c488567f0e80f8d57bbbdd9dd82/helix-term/src/main.rs#L118

I can fix this, but wanted to know if there are some caveats first.

### Reproduction Steps

I tried this:

1. `hx <dir>` where the directory contains .helix/config.toml

I expected this to happen:

The config is loaded

Instead, this happened:

The config is not loaded

### Helix log

Link to where the bug is in the source code is provided in the description

### Platform

All

### Terminal Emulator

All

### Helix Version

master (68fce3e160c64c488567f0e80f8d57bbbdd9dd82)

the-mikedavis:
Closed by https://github.com/helix-editor/helix/pull/8498

PR #8498, opened 2023-10-09
Set the working directory before loading the config

bjorn-ove:
When you start the editor with a directory as the first argument, or specify `-w, --working-dir <path>` it loads `.helix/config.toml` as if you didn't specify it.

This fixes issue #8469

### DIFF
--- a/helix-loader/src/lib.rs
+++ b/helix-loader/src/lib.rs
@@ -33,9 +33,9 @@ pub fn current_working_dir() -> PathBuf {
     path
 }
 
-pub fn set_current_working_dir(path: PathBuf) -> std::io::Result<()> {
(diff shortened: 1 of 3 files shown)


pascalkuthe (approved):

the-mikedavis (approved):

boofexxx:
seems like the merge has introduced a bug. if you pass a directory as the first argument then it changes the current directory to it but it still tries to open the directory which causes opening a new file as the new of the directory

e.g.: `hx src/` makes it change the `CWD` to `src` and opens `src` inside of the `src` so we have `src/src` opened

bjorn-ove:
This is a good catch. I don't know how I missed it 😟. When the directory is changed, the later logic checks if the relative path is a directory, which now is false because the current directory has changed.

Will fix ASAP

bjorn-ove:
Fixed in #8520

PR #8511, opened 2023-10-10
Handle absolute LSP clients paths

zoedsoupe:
For most users the LSP client would be installed and available to `$PATH` env var. However, on declarative build systems like `nix` or `guix` the most common is to make a derivation with aboslute paths set up to avoid cluttering the user environment.

So if I would like to set up `elixir-ls` from `nix` to the helix, using the `home-manager` module: https://nix-community.github.io/home-manager/options.html#opt-programs.helix.enable it should be like that:
```nix
{pkgs, ...}:

{
  programs.helix = {
    enable = true;
    language = {
      language-server = {
        elixir-ls = {
          command = "${pkgs.elixir-ls}/bin/elixir-ls";
        };
      };
    };
  };
}
```

Which resolves to a `languages.toml` file like that:
```toml
[language-server.elixir-ls]
command = "/nix/store/d77wwh0dcc066cq7x8jmxldmskf44773-elixir-ls-0.16.0/bin/elixir-ls"  
```

Normally it would fail, as it would try to execute `which` crate on an aboslute path, making the LSP client unavailable.

With this patch, it's possible to give an absolute path or simply the base name for the executable.

Now, running `cargo -- --health elixir` gives the right output:
```sh
Configured language server: /nix/store/d77wwh0dcc066cq7x8jmxldmskf44773-elixir-ls-0.16.0/bin/elixir-ls
Binary for language server: /nix/store/d77wwh0dcc066cq7x8jmxldmskf44773-elixir-ls-0.16.0/bin/elixir-ls
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
```

This also solves this https://github.com/helix-editor/helix/discussions/8457 and also this issue: https://github.com/nix-community/home-manager/issues/4525

https://github.com/helix-editor/helix/assets/44469426/ce46b627-c0a8-47b3-9975-e65491353055

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -21,7 +21,10 @@ use std::sync::{
     atomic::{AtomicU64, Ordering},
     Arc,
 };
-use std::{collections::HashMap, path::PathBuf};
+use std::{
+    collections::HashMap,
+    path::{Path, PathBuf},
+};
 use tokio::{
     io::{BufReader, BufWriter},
     process::{Child, Command},
@@ -181,8 +184,8 @@ impl Client {
         req_timeout: u64,
         doc_path: Option<&std::path::PathBuf>,
     ) -> Result<(Self, UnboundedReceiver<(usize, Call)>, Arc<Notify>)> {
-        // Resolve path to the binary
-        let cmd = which::which(cmd).map_err(|err| anyhow::anyhow!(err))?;
+        // Reads the absolute path otherwise resolves path to the binary
+        let cmd = Self::resolve_cmd_path(cmd)?;
 
         let process = Command::new(cmd)
             .envs(server_environment)
@@ -241,6 +244,15 @@ impl Client {
         };
 
         Ok((client, server_rx, initialize_notify))
+    }
+
+    fn resolve_cmd_path(cmd: &str) -> Result<PathBuf> {
+        if Path::new(cmd).exists() {
+            return Ok(PathBuf::from(cmd));
+        }
+
+        let path = which::which(cmd).map_err(|err| anyhow::anyhow!(err))?;
+        return Ok(path);
     }
 
     pub fn name(&self) -> &str {


zoedsoupe:
I’m willing to write integrations tests for this feature-patch, would anyone give me some initial instructions for that?

kirawi:
It should already handle absolute paths. See the documentation for [`which`:](https://docs.rs/which/4.4.2/which/fn.which.html)

> Find an executable binary’s path by name.
> 
> If given an absolute path, returns it if the file exists and is executable.
> 
> If given a relative path, returns an absolute path to the file if it exists and is executable.
> 
> If given a string without path separators, looks for a file named binary_name at each directory in $PATH and if it finds an executable file there, returns it.

helix-lsp/src/client.rs
```diff
@@ -243,6 +246,15 @@ impl Client {
         Ok((client, server_rx, initialize_notify))
     }
 
+    fn resolve_cmd_path(cmd: &str) -> Result<PathBuf> {
+        if Path::new(cmd).exists() {
```
pascalkuthe, line 250: This check should be whether the path is absolute rather than whether the file exists. Otherwise files from the CWD will be prioritized over those from the path.

zoedsoupe:
> It should already handle absolute paths. See the documentation for [`which`:](https://docs.rs/which/4.4.2/which/fn.which.html)
> 
> > Find an executable binary’s path by name.
> > If given an absolute path, returns it if the file exists and is executable.
> > If given a relative path, returns an absolute path to the file if it exists and is executable.
> > If given a string without path separators, looks for a file named binary_name at each directory in $PATH and if it finds an executable file there, returns it.

ok, that’s seen strange though… As I wrote on the discussion, and in the home-manager issue, giving abosulute paths to the language-server config fails on `elixir-ls` is not avaialable on `$PATH`

pascalkuthe:
which has some extra checks around executable flags I think.

Does running `which <path>` in your shell work for your example?

I tested replacing rust-analyzer with `/home/pascalkuthe/.cargo/bin/rust-analyzer-nightly` in my config and it works just fine so it doens't seem broken upstream in general

zoedsoupe:
> which has some extra checks around executable flags I think.
> 
> Does running `which <path>` in your shell work for your example?
> 
> I tested replacing rust-analyzer with `/home/pascalkuthe/.cargo/bin/rust-analyzer-nightly` in my config and it works just fine so it doens't seem broken upstream in general

So there’s something wrong with the `which` crate (:

Trying to execute `which <nix-elixir-ls-path` indeed returns the actual aboslute path, however helix seems unable to find it on `$PATH`, as caould be seen in the video.

https://github.com/helix-editor/helix/assets/44469426/984d6fb9-bf39-4fa7-9e2d-eabcc389c27b

pascalkuthe:
I don't actually think so.


The error you djow says `elixer-ls` not found, it should says <PATH> not found: https://github.com/helix-editor/helix/blob/1852292451d984de33ea526b3b7dc7f51c6130eb/helix-term/src/health.rs#L305


It looks like your co ig file is not being picked up by your helix install but is being picked up by your from source builder. Executing Cargo run with an unmodified main branch will likely work.

zoedsoupe:
Ok so now I have some doubts… Do I need to “import" `languages.toml` into `config.toml`?

My config folder seems like this:
```sh
/Users/zoedsoupe/.config/helix
├── config.toml -> /nix/store/f1al2b4yn0ms9ax2ys5kzkbj221kbvja-home-manager-files/.config/helix/config.toml
└── languages.toml -> /nix/store/f1al2b4yn0ms9ax2ys5kzkbj221kbvja-home-manager-files/.config/helix/languages.toml

1 directory, 2 files
```

Executing `:config-open` on helix it correctly opens the `config.toml`. So what would be incorrect?

pascalkuthe:
languages.toml being a dedicated fine is file. You don't need to import it. It's hard for me to debug it here but the health check should say `/path/to/nix/store/elixr-ls not found` and not `elixr-ls not found`.

Like I said since the from source build does eem to pickup the co fig you can try building and running the unomdified master branch and checking the health output

pascalkuthe:
Notice in your own Screencast how it prints the full path twice.

The first thing that is printed `Configured language server ...` is always just a printout of whatever was loaded from the co fig file. So no matter what which does the first path should be the same thing as in the config.

zoedsoupe:
> languages.toml being a dedicated fine is file. You don't need to import it. It's hard for me to debug it here but the health check should say `/path/to/nix/store/elixr-ls not found` and not `elixr-ls not found`.
> 
> Like I said since the from source build does eem to pickup the co fig you can try building and running the unomdified master branch and checking the health output

In fact building master from the upstream and running `cargo run` seems to loads the right config. So how I could debug this problem to discover the problem root? I know it’s possible to use custom config with `-c` option but how about `languages.toml`?

zoedsoupe:
Running `nix shell nixpkgs#elixir-ls` puts `elixir-ls` directly on $PATH and o running `hx —health elixir` gives the right config:
```sh
Configured language server: elixir-ls
Binary for language server: /nix/store/a8p0jqx800k1hx1pd86ncwaydhskgg6i-elixir-ls-0.16.0/bin/elixir-ls
Highlight queries: ✓
Textobject queries: ✓
Indent queries: ✓
```

So it seems that helix always search on the `$PATH` env var, which it isn’t filled by `nix`, because it isolates the binary into its own path `/nix/store/<hash>`.

pascalkuthe:
Yes ofcourse but as you discovered you can indeed use absolute paths too (atloguh I wouldn't recommend.endedsince you manually need to stup each LSP).

It's just that bommanaher is setting some environmental.ent variable or passing some flags (or doing something else weird) to make your installed binary look for its languages.toml elsewhere. I can't tell you where or how that's happening but this seems pretty clearly like a bug with homemanager to me since the issue is not reproducible when building from master.

gabydd:
home-manager doesn't do anything specific with runtime paths but the nixpkgs pkg does: https://github.com/NixOS/nixpkgs/blob/36aa2d9072233ff2a27f8643a3b64b42a7a079a3/pkgs/applications/editors/helix/default.nix#L37-L38 I think this will be fixed in the next release though because we will now look in .config no matter what, as well look in the nix specific runtime directory

zoedsoupe:
> home-manager doesn't do anything specific with runtime paths but the nixpkgs pkg does: https://github.com/NixOS/nixpkgs/blob/36aa2d9072233ff2a27f8643a3b64b42a7a079a3/pkgs/applications/editors/helix/default.nix#L37-L38 I think this will be fixed in the next release though because we will now look in .config no matter what, as well look in the nix specific runtime directory

This doesn’t seems to be the root cause. the `runtime` path set up for helix contains only those themes and tressitter gramars. a literal copy of the `runtime` path from the repository. There isn’t any other user config (aka `config.toml` and `languages.toml`) defined by nix

PR #8514, opened 2023-10-11
Update PureScript grammar and queries

postsolar:


the-mikedavis:
The changes to `src` from `tree-sitter generate` need to be committed. It looks like https://github.com/Maskhjarna/tree-sitter-purescript/pull/21 made the changes in grammar.js but didn't generate & check in the changes

Issue #8515, opened 2023-10-11
File picker not showing on directory open


bbrodriges:
### Summary

File picker behaviour has been broken somewhere between tag `23.05` and commit `1852292451d984de33ea526b3b7dc7f51c6130eb`. On the most recent commit file picker does not show when invoke Helix with directory path as argument.

### Reproduction Steps

1. Build Helix from source at `1852292451d984de33ea526b3b7dc7f51c6130eb`
2. Invoke `hx /somedir/someproject`, no file picker will be shown
3. Invoke `cd /somedir/someproject && hx .`, file picker will be shown

On `23.05` release both invocations would show file picker.

### Helix log

```
2023-10-11T12:43:10.371 helix_vcs [INFO] Error {
    context: "failed to open git repo",
    source: Discover(
        InaccessibleDirectory {
            path: "<path_to_non-git_project>",
        },
    ),
}
2023-10-11T12:43:10.372 helix_vcs [INFO] failed to open diff base for <path_to_non-git_project>
2023-10-11T12:43:10.372 helix_vcs [INFO] Error {
    context: "failed to open git repo",
    source: Discover(
        InaccessibleDirectory {
            path: "<path_to_non-git_project>",
        },
    ),
}
2023-10-11T12:43:10.372 helix_vcs [INFO] failed to obtain current head name for <path_to_non-git_project>
```

### Platform

macOS

### Terminal Emulator

iTerm2 Build 3.4.21

### Helix Version

23.05-473-g18522924

the-mikedavis:
Looks like a regression from https://github.com/helix-editor/helix/pull/8498 (https://github.com/helix-editor/helix/pull/8498#issuecomment-1756997071)

woojiq:
Off topic: @bbrodriges you don't need to format the commit hash with `.
Just paste as is and github will generate a nice formatted hash with the link for you. 
E.g. 5cb76e74f9bd226f1b4757f478bf49873d9ecfe6

bjorn-ove:
It is fixed in #8520

PR #8520, opened 2023-10-11
Fix issue with opening a relative directory with `hx <dir>` introduced in #8498

bjorn-ove:
When opening the editor using `hx <dir>` where `<dir>` is relative it would either open the wrong directory in the picker, or a new file.

This was introduced by #8498 

My solution is to replace the first argument by "." when it is a directory and override `args.working_directory`. This should effectively leave the original logic in place while still solve the config issue.

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -17,6 +17,7 @@ pub struct Args {
     pub log_file: Option<PathBuf>,
     pub config_file: Option<PathBuf>,
     pub files: Vec<(PathBuf, Position)>,
+    pub open_cwd: bool,
     pub working_directory: Option<PathBuf>,
 }
 
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -162,8 +162,7 @@ impl Application {
             // Unset path to prevent accidentally saving to the original tutor file.
             doc_mut!(editor).set_path(None);
         } else if !args.files.is_empty() {
-            let first = &args.files[0].0; // we know it's not empty
-            if first.is_dir() {
+            if args.open_cwd {
                 // NOTE: The working directory is already set to args.files[0] in main()
                 editor.new_file(Action::VerticalSplit);
                 let picker = ui::file_picker(".".into(), &config.load().editor);
--- a/helix-term/src/main.rs
+++ b/helix-term/src/main.rs
@@ -76,7 +76,7 @@ FLAGS:
         helix_loader::default_log_file().display(),
     );
 
-    let args = Args::parse_args().context("could not parse arguments")?;
+    let mut args = Args::parse_args().context("could not parse arguments")?;
 
     helix_loader::initialize_config_file(args.config_file.clone());
     helix_loader::initialize_log_file(args.log_file.clone());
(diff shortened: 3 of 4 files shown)


woojiq:
Does this fix https://github.com/helix-editor/helix/issues/8515?

bjorn-ove:
> Does this fix #8515?

Yes, it should

helix-term/src/args.rs
```diff
@@ -97,6 +97,14 @@ impl Args {
             args.files.push(parse_file(&arg));
         }
 
+        // Set args.working_directory to the first file if it is a directory + replace it with "."
+        if let Some((path, _)) = args.files.first_mut() {
```
the-mikedavis, line 101: Should we also check that `args.working_directory.is_none()` here? I think that flag should take precedence over the first file/dir

bjorn-ove: In the original code flow, the directory in the first file took precedence. This is why I kept it that way.

We could change the old behavior, and leave files[0] untouched when the working dir is set, and I think the behaviour would make sense logically (like `tar -C <dir>`)

bjorn-ove: Sorry, if the path was relative before, I guess it would work as if working dir took precedence. I just thought about it wrong.

Will fix

bjorn-ove: I think the way to fix this is to store the directory from `args.files[0]` in a new member such as `args.extra_working_directory`, and then leave the existing logic. Then, in `main()` set the working directory twice, in order.

I will make this change later today unless I'm told otherwise.

bjorn-ove: My last suggestion also has an edge case that doesn't work, will need to consider it more

bjorn-ove:
The new solution should work well together with working_directory. I have also rebased against master.

pascalkuthe (dismissed):

the-mikedavis (commented):
I'd like to avoid passing around a flag like this, it seems leaky / ad-hoc to me. I think it should work to turn all relative paths we get in the args to absolute, relative to the current working directory at start-up time (before we switch the working directory to the one specified with `-w`). That should also have less confusing behavior when using the `-w` flag: `hx -w path/to/helix Cargo.toml` should open the `Cargo.toml` in the current directory rather than Helix's `Cargo.toml` IMO

bjorn-ove:
I also didn't like that flag, but it was the cleanest I could come up with while keeping the original behaviour. Since that was to be relative from `-w <dir>`, I tried to keep it.

This is also how all other tools I know work. If you run `tar -C <dir> hello.txy` it changes the directory before opening hello.txt. Usually it is basically an alias for `(cd <dir> && tool arg1 arg2 arg3)`

I also thought about joining `working_directory` and `files[0]`, but to make that behave correctly on all platforms is complex.

bjorn-ove:
Another issue with your suggestion is non existing paths. As canonicalize needs the path to exist.

the-mikedavis:
Hmm yeah, let's follow other tools then and keep those files relative to the working path rather than cwd at invocation.

The behavior here seems ok then but I'd still like to clean up that rogue flag - it seems like that's leaking out of `args`. Let's make it a part of the Args struct, something like `open_cwd`

bjorn-ove:
> it seems like that's leaking out of `args`. Let's make it a part of the Args struct, something like `open_cwd`

The reason I didn't do this is that I wanted to leave the args parsing without side-effects. And since I depend on setting the working directory before I check if it is a directory, that wouldn't work. I could make it mutable in `main()` and set it there, if you think that is acceptable?

the-mikedavis (approved):

pascalkuthe (approved):

bjorn-ove:
> Hmm yeah, let's follow other tools then and keep those files relative to the working path rather than cwd at invocation.

I may have unintentionally lied about other tools. At least for tar, it will open the file from -f from the current directory, then change, and create all files in the new location.

So it might actually make sense to iterate trough all files in main() and resolve to full path before altering the directory.

Should I look into this, or just leave it as it is now?

the-mikedavis:
Hmm yeah, let's resolve the paths before altering the directory. IMO that behavior is more intuitive and I believe shell completion should help with that approach (it should assume files are relative to the shell's cwd)

PR #8524, opened 2023-10-12
runtime/themes: adding "ttox" theme

tomasruud:
I've made a small theme that should play well with most terminals, including Terminal.app on macOS. This theme has no explicit `ui.background` which means that if you were to use something like Terminal.app, it automatically handles dark and light mode for you.

Here are a couple of screenshots

light mode
dark mode

### DIFF
--- a/runtime/themes/ttox.toml
+++ b/runtime/themes/ttox.toml
@@ -1,6 +1,9 @@
 # Author : Tomas Ruud <git@me.2ms.no>
 
 "ui.selection" = { fg = "white", bg = "gray" }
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):

boofexxx:
like it but it's not possible to see the primary selection and the cursors on the selections

image

tomasruud:
You're right @boofexxx, I didn't think about that. I added a commit now with some improvements.

the-mikedavis (approved):

pascalkuthe (approved):

PR #8528, opened 2023-10-13
Add 'while_statement' to bash indents

gajdusek:


David-Else:
Bash tree sitter has also had some major updates and could do with updating in Helix: https://github.com/tree-sitter/tree-sitter-bash/releases

pascalkuthe (approved):

PR #8535, opened 2023-10-13
Highlight types and enum members in the rust prelude

TheZoq2:
A relatively simple change, but I like having the things in the prelude highlighted as builtin types

### DIFF
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -49,43 +49,51 @@
   (identifier) @label)
 
 ; ---
-; Standard library
+; Prelude
 ; ---
 
-((identifier) @constant.builtin
+((identifier) @type.enum.variant.builtin
  (#any-of? @constant.builtin "Some" "None" "Ok" "Err"))
 
-((identifier) @type.builtin
- (#any-of?
-    @type.builtin
-    "Result"
-    "Option"
-    "Vec"
-    "Box"
-    "Clone"
-    "Iterator"
-    "IntoIterator"
-    "DoubleEndedIterator"
-    "ExactSizeIterator"
-    "String"
-    "ToString"
-    "TryFrom"
-    "TryInto"))
+
 
 ((type_identifier) @type.builtin
  (#any-of?
     @type.builtin
-    "Result"
-    "Option"
-    "Vec"
-    "Box"
-    "Clone"
-    "Iterator"
-    "IntoIterator"
+    "Send"
+    "Sized"
+    "Sync"
+    "Unpin"
+    "Drop"
+    "Fn"
+    "FnMut"
+    "FnOnce"
+    "AsMut"
+    "AsRef"
+    "From"
+    "Into"
     "DoubleEndedIterator"
     "ExactSizeIterator"
+    "Extend"
+    "IntoIterator"
+    "Iterator"
+    "Option"
+    "Result"
+    "Clone"
+    "Copy"
+    "Debug"
+    "Default"
+    "Eq"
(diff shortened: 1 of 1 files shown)


kirawi:
This isn't exhaustive (and we don't support `#any-of` I'm pretty sure). IMO this shouldn't merged and in the future would be better addressed with LSP semantic highlighting.

archseer:
any-of will soon be supported by upstream tree-sitter

the-mikedavis (dismissed):
Since https://github.com/helix-editor/helix/pull/9317 we now support `#any-of?`

runtime/queries/rust/highlights.scm
```diff
@@ -48,6 +48,57 @@
...
+    @type.builtin
+    "Send"
+    "Sized"
+    "Sync"
+    "Unpin"
+    "Drop"
+    "Fn"
+    "FnMut"
+    "FnOnce"
+    "drop"
```
pascalkuthe, line 71:
```suggestion
```

This is not a type it's a function

runtime/queries/rust/highlights.scm
```diff
@@ -48,6 +48,57 @@
...
+    "Option"
+    "Result"
+    "Clone"
+    "Copy"
+    "Debug"
+    "Default"
+    "Eq"
+    "Hash"
+    "Ord"
+    "Ord"
```
pascalkuthe, line 90:
```suggestion
```

Duplicate

TheZoq2:
Sorry, I switched away from helix for a while so this got stale. Updated with your comments now

the-mikedavis (approved):

runtime/queries/rust/highlights.scm
```diff
@@ -48,6 +48,55 @@
   "'" @label
   (identifier) @label)
 
+; ---
+; Prelude
+; ---
+
+((identifier) @constant.builtin
```
pascalkuthe, line 55: I am not sure if these are really constants. These are normally highlighted as enum variants. I think these prelude highlights should only specialize the highlight scope (like `type.builtin` instead of `type`) not change it entirely.

pascalkuthe: In theory the scoppe is supposed to match here. So I guess it would be `type.enum.variant.builtin`?

I don't see that pattern actually matching for me tough (instead `Ok()` seems to match `constructor`). Maybe there is also a precedance issue? @the-mikedavis I think you know more than me here

the-mikedavis: With the interpretation that these are values at runtime, any type is a constant. As enum members these are certainly types and I would consider neovim's queries to be incorrect. Pascal's suggestion of `type.enum.variant.builtin` is probably the most accurate here although I don't believe any themes define that key yet

TheZoq2: That's fair, pushed an update with that change

runtime/queries/rust/highlights.scm
```diff
@@ -48,6 +48,55 @@
   "'" @label
   (identifier) @label)
 
+; ---
+; Prelude
+; ---
+
+((identifier) @type.enum.variant.builtin
+ (#any-of? @constant.builtin "Some" "None" "Ok" "Err"))
```
the-mikedavis, line 56:
```suggestion
 (#any-of? @type.enum.variant.builtin "Some" "None" "Ok" "Err"))
```

the-mikedavis (approved):

pascalkuthe (approved):

Issue #2797, opened 2022-06-17
A way to insert text with (insert mode) keymapping


thomkeh:
For markdown files, I like to have a keybinding where when I type `` `p`` it's expanded to `` ```python`` (similarly, `` `r`` to `` ```rust`` and so on).

Currently, I do this with
```toml
[keys.insert]
"`" = { p = ["normal_mode", ":pipe echo '```python'", "collapse_selection", "insert_mode"] }
```
This does work, but it seems a bit convoluted.

As far as I know there is currently no other way. (Correct me if I'm wrong.)

I would imagine a solution like this:
```toml
[keys.insert]
"`" = { p = ":insert_text ```python" }
```

the-mikedavis:
You could simplify that keymap with `:insert-output`/`:append-output` https://github.com/helix-editor/helix/pull/2589

It's probably best to solve this with #1383 in the long run.

thomkeh:
Ah, I was using release 22.05 so I didn't have those commands. Sorry!

thomkeh:
On closer inspection, those commands are also based on calling shell commands. So, it's basically the same issue with them.

sudormrfbin:
I would say snippet support (#1178) is the best way to solve this since you want some text to expand to an arbitrary longer text.

PR #2116, opened 2022-04-15
Add command to insert a space character

AntonioLucibello:
this, alongside the command to commit undo state, would allow a user to remap space in insert mode to add a space and create an undo checkpoint, increasing undo granularity.

PR #8542, opened 2023-10-15
Add insert_space command

yerlaser:
Add command to insert whitespace.
Motivation: I would like to add a special binding in normal mode to Shift-Space shortcut.
However, in insert mode it is very easy to hit this combination by accident as Shift often remains depressed after a previous character was inserted.
Therefore, I would like to assign Shift-Space to insert space, just like Space does, but only in insert mode.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -366,6 +366,7 @@ impl MappableCommand {
         smart_tab, "Insert tab if all cursors have all whitespace to their left; otherwise, run a separate command.",
         insert_tab, "Insert tab char",
         insert_newline, "Insert newline char",
+        insert_space, "Insert whitespace",
         delete_char_backward, "Delete previous char",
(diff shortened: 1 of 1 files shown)


the-mikedavis:
See https://github.com/helix-editor/helix/pull/2116

yerlaser:
> See #2116

Thanks, that's another use case that I didn't think of.
In general, though, this command can have a few other uses.

In that other PR you refer to the Macro Style Bindings, but that feature seems to be stalled since 2021.
May I suggest to have this now and once that arrives all three _insert_ commands can be replaced by one.

tmke8:
This could also be generalized to an `:insert-text` typeable command, and then you'd use

```toml
"S-ret" = ":insert-text ' '"
```

There is an open issue for that: #2797

yerlaser:
OK. I see.

PR #8561, opened 2023-10-17
add picker config

zaucy:
This adds some general/global configurability for the picker. I've been running a fork and I was using different styles than the default. I figured others might like to as well.

Additionally I've noticed there are some PRs that maybe want to change how the picker works that might also like to put those options in the new `PickerConfig`. Namely #8155, #7124

<details>
  <summary>Old Example see <a href="#pullrequestreview-1683190609">comment</a></summary>
  
helix-picker-config
  
</details>

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -522,12 +522,8 @@ impl<T: Item + 'static> Picker<T> {
         }
 
         let text_style = cx.editor.theme.get("ui.text");
-        let selected = cx.editor.theme.get(&picker_config.selected_style);
-        let highlight_style = cx
-            .editor
-            .theme
-            .get(&picker_config.highlight_style)
-            .add_modifier(Modifier::BOLD);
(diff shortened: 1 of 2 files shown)


pascalkuthe (changes requested):
Config options should never be used to configure themen keys. If you want to change the way the picker looks change theme.

zaucy:
@pascalkuthe understood. I removed the style options and kept `picker.highlight-symbol`.

the-mikedavis:
I would prefer that we be opinionated rather than configurable for small display items like this - this seems unnecessary to me

zaucy:
I understand. My motiviation is that on my fork of helix I have icons and a different look for selection that the " > " doesn't really fit well with. I want it to be empty ("").

image

pascalkuthe:
If you have your own fork anyway then you can simply patch the indicator there. I don't think it makes sense as an option in core. We try to keep co fig options to a minimum

PR #8569, opened 2023-10-19
don't break on hyphen with :reflow

Stephen-Seo:
This PR adds a new config option `editor.no-break-on-hyphen` that if set to `true`, will cause `:reflow` to not line-break on a hyphen. (By default this is `false`, which is the previous behavior.)

A use case for this behavior is in LaTeX, where if a line-break exists between a word, a hyphen, and another word, the hyphen'd words will not be properly displayed in the generated .pdf . bad linebreak in hyphen in LaTeX

Please let me know if I made a mistake.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -65,7 +65,6 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `workspace-lsp-roots` | Directories relative to the workspace root that are treated as LSP roots. Should only be set in `.helix/config.toml` | `[]` |
 | `default-line-ending` | The line ending to use for new documents. Can be `native`, `lf`, `crlf`, `ff`, `cr` or `nel`. `native` uses the platform's native line ending (`crlf` on Windows, otherwise `lf`). | `native` |
 | `insert-final-newline` | Whether to automatically insert a trailing line-ending on write if missing | `true` |
-| `no-break-on-hyphen` | Disable linebreak on hyphen when using `:reflow` | `true` |
 
 ### `[editor.statusline]` Section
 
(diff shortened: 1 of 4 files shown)


Stephen-Seo:
I changed the default to adopt the new behavior, if that's ok with everyone.

book/src/configuration.md
```diff
@@ -65,6 +65,7 @@ Its settings will be merged with the configuration directory `config.toml` and t
 | `workspace-lsp-roots` | Directories relative to the workspace root that are treated as LSP roots. Should only be set in `.helix/config.toml` | `[]` |
 | `default-line-ending` | The line ending to use for new documents. Can be `native`, `lf`, `crlf`, `ff`, `cr` or `nel`. `native` uses the platform's native line ending (`crlf` on Windows, otherwise `lf`). | `native` |
 | `insert-final-newline` | Whether to automatically insert a trailing line-ending on write if missing | `true` |
+| `no-break-on-hyphen` | Disable linebreak on hyphen when using `:reflow` | `true` |
```
the-mikedavis, line 68: If we do introduce this option it should be named so that it's clear that it's for the reflow command, like `reflow-break-on-hyphen`

the-mikedavis (commented):
Do we actually need a config option for this? If it's sane behavior in all cases then I think we should just hardcode it and get rid of the config option. If it makes sense to break on hyphens only for some languages then this should be language-level configuration instead of editor-wide

pascalkuthe:
Yeah I meant that we can hard-coded it and wouldn't need a config

Stephen-Seo:
Ok, I made a commit that reverted all changes and only has the behavior to not break in a hyphen. I guess the branch name is inaccurate now since there is no config option that changes this behavior. Is this ok with everyone?

pascalkuthe:
brnahc names don't really matter, don't worry about it

pascalkuthe (approved):

the-mikedavis (approved):

PR #8584, opened 2023-10-21
Add command to move files with LSP support

### LINKED ISSUE #4393: Rename buffers

yo-main:
close #4393

Allows to rename the active buffer by moving it to another path.

```
:move my_new_file.rs
```

Heavily based on all the work done in https://github.com/helix-editor/helix/pull/5531 by @ontley and apply the suggestions

The PR seems dead and I would really like to see that feature come into helix :)

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -548,6 +548,11 @@ impl Client {
                         dynamic_registration: Some(true),
                         relative_pattern_support: Some(false),
                     }),
+                    file_operations: Some(lsp::WorkspaceFileOperationsClientCapabilities {
+                        will_rename: Some(true),
+                        did_rename: Some(true),
+                        ..Default::default()
+                    }),
                     ..Default::default()
                 }),
                 text_document: Some(lsp::TextDocumentClientCapabilities {
@@ -698,6 +703,28 @@ impl Client {
         self.notify::<DidChangeWorkspaceFolders>(DidChangeWorkspaceFoldersParams {
             event: WorkspaceFoldersChangeEvent { added, removed },
         })
+    }
+
+    pub fn will_rename_files(
+        &self,
+        params: &Vec<lsp::FileRename>,
+    ) -> impl Future<Output = Result<Option<lsp::WorkspaceEdit>>> {
+        let files = params.clone();
+        let request = self.call::<lsp::request::WillRenameFiles>(lsp::RenameFilesParams { files });
+
+        async move {
+            let json = request.await?;
+            let response: Option<lsp::WorkspaceEdit> = serde_json::from_value(json)?;
+            Ok(response)
+        }
+    }
+
+    pub fn did_rename_files(
+        &self,
+        params: &Vec<lsp::FileRename>,
+    ) -> impl Future<Output = Result<()>> {
+        let files = params.clone();
+        self.notify::<lsp::notification::DidRenameFiles>(lsp::RenameFilesParams { files })
     }
 
     // -------------------------------------------------------------------------------------------
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -404,9 +404,19 @@ impl Client {
     where
         R::Params: serde::Serialize,
     {
+        self.call_with_timeout::<R>(params, self.req_timeout)
+    }
+
+    fn call_with_timeout<R: lsp::request::Request>(
+        &self,
+        params: R::Params,
+        timeout_secs: u64,
+    ) -> impl Future<Output = Result<Value>>
+    where
+        R::Params: serde::Serialize,
+    {
         let server_tx = self.server_tx.clone();
         let id = self.next_request_id();
-        let timeout_secs = self.req_timeout;
 
         async move {
             use std::time::Duration;
@@ -727,7 +737,10 @@ impl Client {
             old_uri: old_uri.to_string(),
             new_uri: new_uri.to_string(),
         }];
-        let request = self.call::<lsp::request::WillRenameFiles>(lsp::RenameFilesParams { files });
+        let request = self.call_with_timeout::<lsp::request::WillRenameFiles>(
+            lsp::RenameFilesParams { files },
+            5,
+        );
 
         Some(async move {
             let json = request.await?;
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -705,23 +705,60 @@ impl Client {
         })
     }
 
-    pub fn will_rename_files(
+    pub fn prepare_file_rename(
         &self,
-        params: &Vec<lsp::FileRename>,
-    ) -> impl Future<Output = Result<Option<lsp::WorkspaceEdit>>> {
-        let files = params.to_owned();
+        old_uri: &lsp::Url,
+        new_uri: &lsp::Url,
+    ) -> Option<impl Future<Output = Result<lsp::WorkspaceEdit>>> {
+        let capabilities = self.capabilities.get().unwrap();
+
+        // Return early if the server does not support willRename feature
+        match &capabilities.workspace {
+            Some(workspace) => match &workspace.file_operations {
+                Some(op) => {
+                    op.will_rename.as_ref()?;
+                }
+                _ => return None,
+            },
+            _ => return None,
+        }
+
+        let files = vec![lsp::FileRename {
+            old_uri: old_uri.to_string(),
+            new_uri: new_uri.to_string(),
+        }];
         let request = self.call::<lsp::request::WillRenameFiles>(lsp::RenameFilesParams { files });
 
-        async move {
+        Some(async move {
             let json = request.await?;
             let response: Option<lsp::WorkspaceEdit> = serde_json::from_value(json)?;
-            Ok(response)
-        }
+            Ok(response.unwrap_or_default())
+        })
     }
 
-    pub fn did_rename_files(&self, params: &[lsp::FileRename]) -> impl Future<Output = Result<()>> {
-        let files = params.to_owned();
-        self.notify::<lsp::notification::DidRenameFiles>(lsp::RenameFilesParams { files })
+    pub fn did_file_rename(
+        &self,
+        old_uri: &lsp::Url,
+        new_uri: &lsp::Url,
+    ) -> Option<impl Future<Output = std::result::Result<(), Error>>> {
+        let capabilities = self.capabilities.get().unwrap();
+
+        // Return early if the server does not support DidRename feature
+        match &capabilities.workspace {
+            Some(workspace) => match &workspace.file_operations {
+                Some(op) => {
+                    op.did_rename.as_ref()?;
+                }
+                _ => return None,
+            },
+            _ => return None,
+        }
+
+        let files = vec![lsp::FileRename {
+            old_uri: old_uri.to_string(),
+            new_uri: new_uri.to_string(),
+        }];
+        Some(self.notify::<lsp::notification::DidRenameFiles>(lsp::RenameFilesParams { files }))
     }
 
     // -------------------------------------------------------------------------------------------
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -709,7 +709,7 @@ impl Client {
         &self,
         params: &Vec<lsp::FileRename>,
     ) -> impl Future<Output = Result<Option<lsp::WorkspaceEdit>>> {
-        let files = params.clone();
+        let files = params.to_owned();
         let request = self.call::<lsp::request::WillRenameFiles>(lsp::RenameFilesParams { files });
 
         async move {
@@ -719,11 +719,8 @@ impl Client {
         }
     }
 
-    pub fn did_rename_files(
-        &self,
-        params: &Vec<lsp::FileRename>,
-    ) -> impl Future<Output = Result<()>> {
-        let files = params.clone();
+    pub fn did_rename_files(&self, params: &[lsp::FileRename]) -> impl Future<Output = Result<()>> {
+        let files = params.to_owned();
         self.notify::<lsp::notification::DidRenameFiles>(lsp::RenameFilesParams { files })
     }
 
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -2411,7 +2411,7 @@ fn redraw(
     Ok(())
 }
 
-fn rename_buffer(
+fn move_buffer(
     cx: &mut compositor::Context,
     args: &[Cow<str>],
     event: PromptEvent,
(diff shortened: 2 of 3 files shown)


archseer:
Can you keep the commit history and retain the original author then? If the PR is simply rebased or patched up I'd like to retain the original author so we properly attribute the work they've put in.

ontley:
> Can you keep the commit history and retain the original author then? If the PR is simply rebased or patched up I'd like to retain the original author so we properly attribute the work they've put in.

It's a big change compared to the code I had, this adds multiple lsp server support and it handles it properly unlike mine

helix-lsp/src/client.rs
```diff
@@ -1341,6 +1341,56 @@ impl Client {
         Some(self.call::<lsp::request::DocumentSymbolRequest>(params))
     }
 
+    pub fn prepare_file_rename(
+        &self,
+        old_uri: String,
+        new_uri: String,
```
pascalkuthe, line 1347: these shuold be URLs instead of strings, ints unfortunate that the LSP prototocoal stores these as strings instead of URIs (which they are required to be according to the protocol). You can then easily convert to string below

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2408,57 @@ fn redraw(
...
+    }
+
+    ensure!(args.len() == 1, format!(":move takes one argument"));
+
+    let new_path = args.first().unwrap().to_string();
+    let doc = doc!(cx.editor);
+    let old_path = doc
+        .path()
+        .ok_or_else(|| anyhow!("Scratch buffer cannot be moved. Use :write instead"))?
+        .to_string_lossy()
```
pascalkuthe, line 2427: Lossy conversion is incorrect. If there was no way to recover here, we should report an error to the user.

Luckily its perfectly possible to construct URLs for invalid utf8 (by percent-encoding non-ascii bytes). So if you use `document.url` (and accept that as arguments for will/did_rename) then that will cover all cases.

This will also prefix the path with `file://` (the fact that it doesn't currently do that is another bug here, the spec specifies that this should be a file:// URL and rust-analyzer also implements it like that)

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2408,57 @@ fn redraw(
...
+                Err(e) => {
+                    log::error!("LSP willRename request failed: {:?}", e);
+                }
+            }
+        };
+        None
+    });
+
+    if let Some(edit) = lsp_reply {
+        if let Err(e) = apply_workspace_edit(cx.editor, helix_lsp::OffsetEncoding::Utf8, &edit) {
```
pascalkuthe, line 2443: hardcoding the offset encodign to utf8 is always incorrect. You need to use the correct offset encoding for the server

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2408,57 @@ fn redraw(
...
+        if let Some(future) = lsp.prepare_file_rename(old_path.clone(), new_path.clone()) {
+            match helix_lsp::block_on(future) {
+                Ok(edit) => return Some(edit),
+                Err(e) => {
+                    log::error!("LSP willRename request failed: {:?}", e);
+                }
+            }
+        };
+        None
+    });
```
pascalkuthe, line 2440: I am not really sure if this is the correct approach for handling `will_rename`. You likely want to call `will_rename` for all servers. Workspace edits will not be applied if they conflict with previous changes anyway.

I think some servers would not handle it well if we called `did_rename` without `will_rename`.

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2408,57 @@ fn redraw(
...
+    let doc = doc!(cx.editor);
+    let old_path = doc
+        .path()
+        .ok_or_else(|| anyhow!("Scratch buffer cannot be moved. Use :write instead"))?
+        .to_string_lossy()
+        .to_string();
+
+    let lsp_reply = doc.language_servers().find_map(|lsp| {
+        if let Some(future) = lsp.prepare_file_rename(old_path.clone(), new_path.clone()) {
+            match helix_lsp::block_on(future) {
```
pascalkuthe, line 2432: This should have a 5s timeout I think. The current timeout is in minutes which is way too long.

pascalkuthe (changes requested):

yo-main:
> > Can you keep the commit history and retain the original author then? If the PR is simply rebased or patched up I'd like to retain the original author so we properly attribute the work they've put in.
> 
> It's a big change compared to the code I had, this adds multiple lsp server support and it handles it properly unlike mine

Couldn't have manage the lsp thing without seeing how it was implemented in your PR :)
I managed to rebase my branch using your fork (that wasn't easy indeed :smile: I hope I didn't mess anything up)

helix-lsp/src/client.rs
yo-main: Updated !

helix-term/src/commands/typed.rs
yo-main: Updated :) I struggled while dealing between Url and Path (as I need them on on the other side) but I think I got something okay.
Let me know if you see any improvements !

helix-term/src/commands/typed.rs
yo-main: Done !

helix-term/src/commands/typed.rs
yo-main: Yep, I would tend to agree as well, but I followed that past comment of yours :smile: 
https://github.com/helix-editor/helix/pull/5531/files#r1255798844

Code is updated, let me know if there's way to make it more readable (I'm not that good with rust idoms) :pray:

helix-term/src/commands/typed.rs
yo-main: updated :)

helix-term/src/commands/typed.rs
pascalkuthe: yeah back then the idea was that we would only send both `willRename` and `didRename` to a single server but I support sending it to all of them should really work aswell.

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2409,78 @@ fn redraw(
...
+                log::error!("LSP {} failed to treat willRename request: {:?}", name, e);
+            }
+        };
+    }
+
+    let doc = doc_mut!(cx.editor);
+    doc.set_path(Some(new_path_as_url.to_file_path().unwrap().as_path()));
+    if let Err(e) = std::fs::rename(
+        &old_path_as_url.to_file_path().unwrap(),
+        &new_path_as_url.to_file_path().unwrap(),
```
pascalkuthe, line 2471: a bit of a code cleanliness here: You convert path -> URL -> path with a n unwrap.

Its mostly fine but a bit inefficient/nonideomatic (avoid unwraps whenever so I don't need to think about whether this could panic or not).

Just store the paths in separate variables.

helix-term/src/commands/typed.rs
pascalkuthe: no, worries it looks good now

helix-term/src/commands/typed.rs
yo-main: Done !

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2409,82 @@ fn redraw(
...
+                log::error!("LSP {} failed to treat willRename request: {:?}", name, e);
+            }
+        };
+    }
+
+    let doc = doc_mut!(cx.editor);
+    let old_path = old_path_as_url.to_file_path().unwrap();
+    let new_path = new_path_as_url.to_file_path().unwrap();
+
+    doc.set_path(Some(new_path.as_path()));
```
ontley, line 2470: I think the `Document`s path should only be set if the file rename succeeds, and the same for LSP DidRenameFiles.

Unfortunately, when I was testing my own version of the rename feature, rust-analyzer accepted the rename even without DidRenameFiles, so I'm not sure what would happen if the file rename failed

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2409,82 @@ fn redraw(
...
+    if let Err(e) = std::fs::rename(&old_path, &new_path) {
+        doc.set_path(Some(old_path.as_path()));
+        bail!("Could not move file: {}", e);
+    };
+
+    doc.language_servers().for_each(|lsp| {
+        lsp.did_file_rename(&old_path_as_url, &new_path_as_url);
+    });
```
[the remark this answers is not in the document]
yo-main: > I think the Document path should only be set if the file rename succeeds.

Hmm that's the case already, isn't it ? There's an error returned if the call to `std::fs:rename` fails.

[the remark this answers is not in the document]
pascalkuthe: while the current code is technically correct it would be clearer/more correct to just set the path once after the rename succeded.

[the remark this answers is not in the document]
pascalkuthe: > Also, when I was testing my own version of the rename feature, rust-analyzer seemed to accept the rename even without DidRenameFiles, so I'm not sure what would happen if the file rename failed

we probably have to rollback the edit if the rename fails. we currently don't implement that but we will want transactional lsp edits eventually which will allow that.

helix-term/src/commands/typed.rs
```diff
@@ -2408,6 +2409,82 @@ fn redraw(
...
+            }
+            Err(e) => {
+                log::error!("LSP {} failed to treat willRename request: {:?}", name, e);
+            }
+        };
+    }
+
+    let doc = doc_mut!(cx.editor);
+    let old_path = old_path_as_url.to_file_path().unwrap();
+    let new_path = new_path_as_url.to_file_path().unwrap();
```
pascalkuthe, line 2468: I didn't just mean that you should create two variables here but instead that you store the paths into variables before converting[ to a nurl to avoid the URL roundtrip

helix-term/src/commands/typed.rs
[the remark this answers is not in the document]
yo-main: > while the current code is technically correct it would be clearer/more correct to just set the path once after the rename succeded instead of setting it immidiedly and then changing it back.

Oh ok, we are talking about the `doc.set_path`, sorry :sweat_smile: 

I can do that, I only implemented in that way following [that](https://github.com/helix-editor/helix/pull/5531/files#r1255790329) comment on the previous PR :smile: 

But I do tend to agree with that comment though. Meaning that if the rename on the FS goes through, but not for the doc in helix, it's indeed a bit more touchy to rollback the change on the FS.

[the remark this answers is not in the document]
ontley: Yeah the rollbacks are a bit of a weird situation, I couldn't find anything about failing to rename a file on the LSP docs, or anywhere actually. I think rolling back with both WillRenameFiles and DidRenameFiles is the best option.

As for `doc.set_path`, I think that's something that can be relied upon, since there shouldn't be a reason that setting the `Document`'s path should fail, other than a bug with the method itself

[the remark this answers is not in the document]
ontley: I forgot to mention, if I remember correctly, rust-analyzer doesn't even recognize DidRenameFiles, though my memory is a bit foggy there, I'll get around to testing that tmr and I'll drop what I find here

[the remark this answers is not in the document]
pascalkuthe: yeah rust-analyzer does not have the didRenameFile capability.
WillReanmeFile does not affect any RA state tough only once we actually send the corresponding transcation to the server (which is why we could rollback easily). 


The whole rollback thing is a can of worms I don't want to open in this PR that is an existing issue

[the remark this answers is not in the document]
yo-main: well maybe rust-analyser don't care about it but other lsp might ^^

I don't think though we should rollback things on the lsp if anything bad happens. Probably the edits but I haven't been able to reproduce a case where a LSP provides edit to apply while having the rename on filesystem to fail.

Somehow I think rust-analyzer is able to forecast that we won't be able to rename the file as we wish to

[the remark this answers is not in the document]
pascalkuthe: > Somehow I think rust-analyzer is able to forecast that we won't be able to rename the file as we wish to

rust-analyzer can't forecast that. IO is inherently non-deterministic. You are right its not a huge issue in practice but it definitely is in edgecases and  vscode does implement that. Ubt like I said rolling back the edits is a bit hard and doesn't need to be done in this PR.

pascalkuthe (approved):

pascalkuthe:
this will probably need to wait for the next release, the current one is pretty close and I doubt we will merge new features before it

ontley (approved):

alirezaayande92 (approved):
https://github.com/users/Mia-Tron/achievements/starstruck

the-mikedavis (approved):

cbr9:
Shouldn't this command allow to move files into directories by automatically adding the filename? Currently you have to manually type the filename if you want to move the buffer into a directory

David-Else:
I just tried this with ltex-ls and immediate problems :(

`hx hello.md`

(create some text and save)

`:move bye.md`

if you use move too fast (before the language server has finished working I assume) you get an error like:
```
2023-11-10T16:12:02.351 helix_lsp::transport [ERROR] ltex-ls err <- "WARNING: Could not find document with URI 'file:///home/david/Downloads/bye.md'\n"
2023-11-10T16:12:02.353 helix_lsp::transport [ERROR] marksman err <- "[16:12:02 WRN] <MarksmanServer> Document not found: {\"uri\": \"{ uri = \\\"file:///home/david/Downloads/bye.md\\\"\\n  data = AbsPath \\\"/home/david/Downloads/bye.md\\\" }\", \"method\": \"textDocumentDidChange\"}\n"
2023-11-10T16:12:02.460 helix_lsp::transport [ERROR] marksman err <- "[16:12:02 WRN] <MarksmanServer> Document not found: {\"uri\": \"{ uri = \\\"file:///home/david/Downloads/bye.md\\\"\\n  data = AbsPath \\\"/home/david/Downloads/bye.md\\\" }\", \"method\": \"textDocumentDidChange\"}\n"
2023-11-10T16:12:02.460 helix_lsp::transport [ERROR] ltex-ls err <- "Nov 10, 2023 4:12:02 PM org.bsplines.ltexls.server.LtexTextDocumentService getDocument\n"
2023-11-10T16:12:02.460 helix_lsp::transport [ERROR] ltex-ls err <- "WARNING: Could not find document with URI 'file:///home/david/Downloads/bye.md'\n"
```
If you wait for ltex-ls to fully load before moving then the new file then you get 'no code actions available' but the text seems to have inherited the old underline errors?

Anyway, something funny is going on with the language server in this case. I am running ltex-ls and marksman. Cheers.

pascalkuthe:
That is a language server problem. What we do on our side is quite trivial

Issue #3741, opened 2022-09-07
Helix leaves the cursor blinking


iio7:
I have my xterm cursor setup to not blink:

```
xterm*cursorBlink: false
```

In Helix, the cursor doesn't blink. But when I exit a file and returns to xterm, the cursor is blinking.

This only happens with Helix in xterm, not with Helix in st, Alacritty or kitty.

I am running Arch Linux 5.19.7-arch1-1 with Helix version 22.08.1.

the-mikedavis:
Partially a duplicate of https://github.com/helix-editor/helix/issues/2684

The PR against that issue fixed the problem on most terminals but some may interpret the `\x1B[0 q` code differently. Once we have terminfo knowledge we should be able to restore the cursor more consistently across terminals

PR #8591, opened 2023-10-23
Use terminfo to reset terminal cursor style

### LINKED ISSUE #2684: Helix changes terminal cursor style

rmehri01:
Uses the `cnorm` capability from `terminfo` if available to reset the terminal cursor style. I tested this with `xterm` and it seems to fix the issue with making the cursor blink like in #3741.

Closes #2684

### DIFF
--- a/helix-tui/src/backend/crossterm.rs
+++ b/helix-tui/src/backend/crossterm.rs
@@ -32,10 +32,21 @@ fn vte_version() -> Option<usize> {
 }
 
 /// Describes terminal capabilities like extended underline, truecolor, etc.
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 struct Capabilities {
     /// Support for undercurled, underdashed, etc.
     has_extended_underlines: bool,
+    /// Support for resetting the cursor style back to normal.
+    reset_cursor_command: String,
+}
+
+impl Default for Capabilities {
+    fn default() -> Self {
+        Self {
+            has_extended_underlines: false,
+            reset_cursor_command: "\x1B[0 q".to_string(),
+        }
+    }
 }
 
 impl Capabilities {
@@ -54,6 +65,10 @@ impl Capabilities {
                     || t.extended_cap("Su").is_some()
                     || vte_version() >= Some(5102)
                     || matches!(term_program().as_deref(), Some("WezTerm")),
+                reset_cursor_command: t
+                    .utf8_string_cap(termini::StringCapability::CursorNormal)
+                    .unwrap_or("\x1B[0 q")
+                    .to_string(),
             },
         }
     }
@@ -154,7 +169,8 @@ where
 
     fn restore(&mut self, config: Config) -> io::Result<()> {
         // reset cursor shape
-        write!(self.buffer, "\x1B[0 q")?;
+        self.buffer
+            .write_all(self.capabilities.reset_cursor_command.as_bytes())?;
         if config.enable_mouse_capture {
             execute!(self.buffer, DisableMouseCapture)?;
         }


helix-tui/src/backend/crossterm.rs
```diff
@@ -154,7 +159,12 @@ where
 
     fn restore(&mut self, config: Config) -> io::Result<()> {
         // reset cursor shape
-        write!(self.buffer, "\x1B[0 q")?;
+        let cursor_normal = self
+            .capabilities
+            .cursor_normal
+            .as_deref()
+            .unwrap_or("\x1B[0 q");
+        write!(self.buffer, "{cursor_normal}")?;
```
the-mikedavis, line 167:
```suggestion
        self.buffer.write_all(cursor_normal.as_bytes())?;
```

here the formatting is just overhead, we could convert to a byte slice instead directly and avoid all that

helix-tui/src/backend/crossterm.rs
```diff
@@ -32,10 +32,12 @@ fn vte_version() -> Option<usize> {
 }
 
 /// Describes terminal capabilities like extended underline, truecolor, etc.
-#[derive(Copy, Clone, Debug, Default)]
+#[derive(Clone, Debug, Default)]
 struct Capabilities {
     /// Support for undercurled, underdashed, etc.
     has_extended_underlines: bool,
+    /// Support for resetting the cursor style back to normal.
+    cursor_normal: Option<String>,
```
the-mikedavis, line 40:
```suggestion
    reset_cursor_command: Option<String>,
```

the-mikedavis: This could just be a `String` and we would default to `\x1B[0 q` in `from_env_or_default`

the-mikedavis (commented):
Just some minor style nits

the-mikedavis (approved):

helix-tui/src/backend/crossterm.rs
```diff
@@ -54,6 +56,10 @@ impl Capabilities {
                     || t.extended_cap("Su").is_some()
                     || vte_version() >= Some(5102)
                     || matches!(term_program().as_deref(), Some("WezTerm")),
+                reset_cursor_command: t
+                    .utf8_string_cap(termini::StringCapability::CursorNormal)
+                    .unwrap_or("\x1B[0 q")
+                    .to_string(),
```
pascalkuthe, line 62: On unic thus seems fine but does this work on windows?

rmehri01: Ah good catch, I just tried it out and it doesn't work on windows since the capability will be set to an empty string instead.

helix-tui/src/backend/crossterm.rs
```diff
@@ -31,11 +31,22 @@ fn vte_version() -> Option<usize> {
     std::env::var("VTE_VERSION").ok()?.parse().ok()
 }
 
+#[derive(Clone, Debug)]
+struct ResetCursorCommand(String);
```
the-mikedavis, line 35: The newtype wrapper for the string here seems unnecessary to me. It's fine just to default with `unwrap_or`/`unwrap_or_else` rather than `unwrap_or_default`

rmehri01: I added this since we also want the default `"\x1B[0 q"` to be used when `terminfo` isn't available, for example on Windows, or else it just defaults to an empty string. Or should I just explicitly make the struct in the other case instead of keeping the `Capabilities::default()`?

the-mikedavis: Ah I see. Let's define `impl Default for Capabilities` instead so Windows can keep calling `Capabilities::default`

the-mikedavis (dismissed):

helix-tui/src/backend/crossterm.rs
```diff
@@ -32,10 +32,21 @@ fn vte_version() -> Option<usize> {
...
     /// Support for undercurled, underdashed, etc.
     has_extended_underlines: bool,
+    /// Support for resetting the cursor style back to normal.
+    reset_cursor_command: String,
+}
+
+impl Default for Capabilities {
+    fn default() -> Self {
+        Self {
+            has_extended_underlines: bool::default(),
```
pascalkuthe, line 46:
```suggestion
            has_extended_underlines: false,
```

pascalkuthe (changes requested):
Small style nit: I dislike using default instead of false. Apart from that this LGTM

pascalkuthe (approved):

the-mikedavis (approved):

kladd:
The terminfos of the terminals on my machine have `cnorm` as `\E[?12l\E25h`, "Stop blinking cursor (AT&T 610)" and "Show cursor (DECTCEM), VT220", which has been my experience since this change.

It seems like there's some adoption of `DECSCUSR 0`/ `\E[0 q` resetting the user default cursor though. Alacritty and the default terminal on macOS seem to—and Windows I guess https://github.com/microsoft/terminal/pull/7379 (haven't tried).

PR #8599, opened 2023-10-24
we hate vscode users

nanodr0ne:
vscode users are so bad and this is the helix repository so merge this so we can fend off all the poor vscode users who don't know how to use helix


the-mikedavis:
?

PR #8607, opened 2023-10-24
[draft] port to `wasm32-unknown-unknown` & web

makemeunsee:
Hi Helix team,

As mentioned in [a discussion](https://github.com/helix-editor/helix/discussions/8588#discussioncomment-7366601), here's a draft PR for my work to have (a subset of) Helix fully bundled as web app.

The [README of the new crate `helix-web`](https://github.com/makemeunsee/helix/blob/wasm32/helix-web/README.md) lists current limitations and known issues.

Let me know if you'd like me to point out things here, or if you prefer to have a look first.

Also I'm looking now at re-enabling parsing with `tree-sitter`, I could compile some parsers to wasm so I'll try to embed them, instead of hot loading them, and use them. Not sure if I should push that work to the same branch though, it would make this PR even messier.


archseer:
@pascalkuthe continuing from https://github.com/helix-editor/helix/discussions/8588#discussioncomment-7371653, I think tokio added wasm support over the last year, running in a single thread in a service worker

helix-core/src/history.rs
```diff
@@ -2,7 +2,12 @@ use crate::{Assoc, ChangeSet, Range, Rope, Selection, Transaction};
 use once_cell::sync::Lazy;
 use regex::Regex;
 use std::num::NonZeroUsize;
-use std::time::{Duration, Instant};
+use std::time::Duration;
+
+#[cfg(target_arch = "wasm32")]
+use instant::Instant;
+#[cfg(not(target_arch = "wasm32"))]
+use std::time::Instant;
```
archseer, line 10: Rather than adding this conditional everywhere, let's re-export instant in the root of the `helix-core` crate, then use that everywhere else

helix-loader/src/grammar.rs
```diff
@@ -85,13 +86,15 @@ pub fn get_language(name: &str) -> Result<Language> {
     Ok(language)
 }
 
+#[cfg(not(target_arch = "wasm32"))]
```
archseer, line 89: All of this git/build handling seems like it could be split into a separate file/module then gated behind a single conditional

archseer: I think this is how you could add a WASM entrypoint for grammars too: add an alternative loader implementation that just uses an internal list of compiled-in grammars

archseer:
Testing this out, `xterm.js` actually supports truecolor (but I guess doesn't advertise correctly). Setting `:set true-color true` I'm able to use the default theme. You probably want to add another override similar to windows for wasm: https://github.com/helix-editor/helix/blob/5b74eaadb47263a62c64be451a3900108ede8d8f/helix-term/src/lib.rs#L25-L28

.github/workflows/web_demo.yml
```diff
@@ -0,0 +1,45 @@
...
+          node-version: 20
+
+      - name: Set up cache
+        uses: Swatinem/rust-cache@v2
+
+      - name: Install wasm-pack
+        uses: jetli/wasm-pack-action@v0.4.0
+        with:
+          version: 'latest'
+
```
archseer, line 25: The action does so little it would be nice to just replace with the official invocation: https://rustwasm.github.io/docs/wasm-bindgen/wasm-bindgen-test/continuous-integration.html#github-actions

helix-term/src/commands/typed.rs
```diff
@@ -339,6 +340,7 @@ fn write_impl(
         insert_final_newline(doc, view);
     }
 
+    #[cfg(feature = "dap_lsp")]
```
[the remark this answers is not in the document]
makemeunsee: I originally tried to rebase your `gui` branch but it was too much outdated. So I redid some of the work you did there in the first, and took a shortcut for these 2 features, knowing neither would work any time soon on wasm. But yeah, they should be distinct, I kinda regret not splitting them in the first place...

helix-loader/src/grammar.rs
makemeunsee: yes that's exactly what I'm attempting now; on the Helix side it looks quite doable, it's all the `tree-sitter-*` libs which will be more tedious to bring to wasm.

.github/workflows/web_demo.yml
makemeunsee: I picked this one as they advertise much faster (seconds vs. minutes) set up times, but we can use the official one sure.

makemeunsee:
I've just rebased onto master, while it's still manageable.

I'll try to ack the various comments over the next days, and point out of few other things to discuss.

helix-loader/src/grammar.rs
makemeunsee: I got syntax highlighting for `rust`, `toml` and `regex` to work, but it's very hacky:
* hacked `tree-sitter-c2rust` so it uses [`instant`](https://github.com/sebcrozet/instant/) instead of `std::time`
* hacked C code of `tree-sitter-*` crates, for `wasm` compilation + dependency on hacked `tree-sitter-c2rust`; not big changes but still tedious and I dont see a way to automate the process. Maybe `c2rust`ing the language libs, but I have no experience with it.

It's nice to see it working but it's quite a mess, so I'll keep it on the side for now.

archseer: Why do we need to use `tree-sitter-c2rust` rather than https://github.com/tree-sitter/tree-sitter/blob/master/lib/binding_web/README.md ? I think we should look at how the tree-sitter playground functions since it runs grammars via wasm (and it lazy loads them instead of compiling them into the main binary)

https://tree-sitter.github.io/tree-sitter/playground

archseer:
To get this merged I'd split the PR into parts:

- Adding feature flags for lsp and dap, enabling by default
- Adding wasm conditionals, and adding a wasm target to the CI build
- `helix-web`, loader, tree-sitter changes etc.

With first two out of the way it would make the PR easier to rebase since the changes would be mostly scoped to helix-web

.github/workflows/web_demo.yml
archseer:
```yaml
      - name: Install
        run: curl https://rustwasm.github.io/wasm-pack/installer/init.sh -sSf | sh
```

From the docs should be equally as fast, it figures out the target architecture and downloads the release (rather than cargo-install-ing and compiling from source)

helix-loader/src/grammar.rs
makemeunsee: Oh that's interesting, I hadnt spotted they had the playground in Rust on wasm too.

It's puzzling, as `tree-sitter` actually compiles to `wasm32-unknown-emscripten`, but not to `wasm32-unknown-unknown`. And `wasm32-unknown-emscripten` is not compatible with `wasm-bindgen`: https://rustwasm.github.io/docs/wasm-bindgen/reference/rust-targets.html?highlight=emscripten#other-web-targets.

I'll take a look, it would be so much better this way.

makemeunsee: Ok so the playground is a Rust web server, serving `emscripten` wasm binaries and regular JS code. The hot loading is js/emscripten code, see https://github.com/tree-sitter/tree-sitter/blob/7c0cee70f55fdfe1667890d372441a85f3127ee6/lib/binding_web/binding.js#L1024-L1042

The Rust bindings compile to linux and wasm-emscripten, but not to wasm-unknown.

I dont see an easy way to interop between a `wasm32-unknown` app and `wasm32-emscripten` libs. Creating Rust bindings for tree-sitter-js and then writing a facade/adapter so Helix code can abstract over the native and web bindings? I got a headache just thinking about it :sweat_smile: 

Open to any ideas!

makemeunsee:
> To get this merged I'd split the PR into parts:
> 
>     * Adding feature flags for lsp and dap, enabling by default
> 
>     * Adding wasm conditionals, and adding a wasm target to the CI build
> 
>     * `helix-web`, loader, tree-sitter changes etc.
> 
> 
> With first two out of the way it would make the PR easier to rebase since the changes would be mostly scoped to helix-web

Sounds like a plan!

makemeunsee:
Created https://github.com/helix-editor/helix/pull/8638

PR #8609, opened 2023-10-25
PoC for printing config location

3tilley:
Hello,

Please don't worry too much about the code quality at the moment, I just wanted to get a feel for whether the functionality below would be considered useful.

I can find it a bit fiddly to find where Helix is loading config files from - especially on Windows. For example lots of software looks for config in `~`, `~/.config` which maps to `C:\Users\$USERNAME\` on modern Windows, Helix (or `etcetera`) puts things in `C:\Users\$USERNAME\AppData\Roaming\Helix` which might be more "technically correct" but doesn't help my config file woes.

It would help if I could quickly see where Helix was looking on a given system. Ideally through a CLI command so it could be semi-scripted. I've attempted to implement that below, where `hx --print-config-location` gives the output below.

```
$ cargo run -- --print-config-location
   Compiling helix-term v0.6.0 (helix\helix-term)
    Finished dev [unoptimized + debuginfo] target(s) in 12.54s
     Running `target\debug\hx.exe --print-config-location`
Workspace config loaded: false
Location:
C:\<snipped>\helix\.helix\config.toml
Global config loaded: false
Location:
C:\<snipped>\AppData\Roaming\helix\config.toml
```

If you do like the idea of this, I'm happy to polish it up and document it. I'd be interested in thoughts on:
1. Command name - is there something shorter than `--print-config-location` that makes it clear it's the path, not the content?
2. Thoughts on structure - should this logic be put into `helix-loader`?
3. Output format. Bear in mind someone is likely to want to do `cat "$(hx --print-config-location | tail -n 1)"` or similar, and especially on Windows that would likely be a path with spaces. This suggests to me the location would be on it's own on one line
4. Does `languages.toml` or anything else belong in the output as well?

Can anyone think of similar commands in other software? `git config --show-origin --list` is the only one that springs to mind. (As an aside in the future if Helix gets more complex I do wonder whether tagging the origin of each piece of config may be useful)

### DIFF
--- a/helix-term/src/args.rs
+++ b/helix-term/src/args.rs
@@ -19,6 +19,7 @@ pub struct Args {
     pub files: Vec<(PathBuf, Position)>,
     pub open_cwd: bool,
     pub working_directory: Option<PathBuf>,
+    pub print_config_location: bool,
 }
 
 impl Args {
@@ -76,6 +77,7 @@ impl Args {
                         anyhow::bail!("--working-dir must specify an initial working directory")
                     }
                 },
+                "--print-config-location" => args.print_config_location = true,
                 arg if arg.starts_with("--") => {
                     anyhow::bail!("unexpected double dash argument: {}", arg)
                 }
--- a/helix-term/src/config.rs
+++ b/helix-term/src/config.rs
@@ -7,13 +7,17 @@ use std::collections::HashMap;
 use std::fmt::Display;
 use std::fs;
 use std::io::Error as IOError;
+use std::path::PathBuf;
 use toml::de::Error as TomlError;
+use crate::ui::picker::PathOrId::Path;
 
 #[derive(Debug, Clone, PartialEq)]
 pub struct Config {
     pub theme: Option<String>,
     pub keys: HashMap<Mode, KeyTrie>,
     pub editor: helix_view::editor::Config,
+    pub global_location: (PathBuf, bool),
+    pub local_location: (PathBuf, bool),
 }
 
 #[derive(Debug, Clone, PartialEq, Deserialize)]
@@ -30,6 +34,8 @@ impl Default for Config {
             theme: None,
             keys: keymap::default(),
             editor: helix_view::editor::Config::default(),
+            global_location: (helix_loader::config_file(), false),
+            local_location: (helix_loader::workspace_config_file(), false),
         }
     }
 }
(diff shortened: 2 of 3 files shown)


the-mikedavis:
Rather than a new flag, let's list all of the config files in `hx --health` where we currently only mention the global config

3tilley:
Ok, I'm happy with that, I'll update.

3tilley:
On this topic, is there:
1. A way to specify the language.toml location?
2. A way to point hx at a config dir that contains both config file and languages? Detect if `--config ~/helix` is a directory for example and take that for both?

EphemeralDev:
I didn't want to comment/clog up the discussion but 3tilley brought up some points I may have had an issue with. I installed helix via scoop, I think i used the -c flag to specify a config location in scoops install directory(scoop may have done it, I can't remember). Later down the line I was having trouble getting a formatter to work. 

I was placing languages.toml in the location of my config file while in reality helix was still looking for it in %AppData%\helix\ . I don't know if it's considered a bug or intentional that languages.toml location didn't change to match the config location. If it is intentional I think the docs should maybe have an additional note about that. This PR would have helped me diagnose the issue faster though.

the-mikedavis:
`-c` only applies to the editor configuration (`config.toml`). I'm not sure we need a way to specify language configuration on the fly since we support per-project language configuration with `.helix/languages.toml` now. We want to keep the number of CLI arguments to a minimum since Helix is interactive so if we do support this, I'd like to see it as a config variable rather than a CLI flag, like in https://github.com/helix-editor/helix/pull/2135.

3tilley:
> `-c` only applies to the editor configuration (`config.toml`). I'm not sure we need a way to specify language configuration on the fly since we support per-project language configuration with `.helix/languages.toml` now. We want to keep the number of CLI arguments to a minimum since Helix is interactive so if we do support this, I'd like to see it as a config variable rather than a CLI flag, like in #2135.

Just to help me understand this point, if I had `~/myconfigs/config.toml` and `~/local_config/languages.toml`, how do I run Helix to point at those?

the-mikedavis:
You can use the `-c` CLI argument for the config file but there isn't currently a way to point to a language config file that isn't in a standard location like `~/.config/helix/languages.toml` or `.helix/languages.toml`

PR #8616, opened 2023-10-25
Update MSRV to reduce install times

NukeManDan:
As I didn't have the prior rust toolchain version installed, the install from source of helix was very very long, in large part as I needed to do a full crates.io registry update. 

[In 1.68.0](https://blog.rust-lang.org/2023/03/09/Rust-1.68.0.html#cargos-sparse-protocol) sparse registry protocol was introduced, and this _radically_ reduces the build time (order of a few minuets for me) including the registry update. Here I propose the MSRV is bumped to at least this release, inclusive of patches to 1.68.2 .

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.65.0"
+channel = "1.68.0"
 components = ["rustfmt", "rust-src"]
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.68.0"
+channel = "1.68.2"
 components = ["rustfmt", "rust-src"]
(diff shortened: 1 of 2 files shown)


rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.65.0"
+channel = "1.68.0"
```
NukeManDan, line 2:
```suggestion
channel = "1.68.2"
```

pascalkuthe:
we have a fixed MSRV policy to never exceed the MSRV required by firefox. That allows any distro that can ship a secure browser to also ship helix. Firefox currently has an MSRV of 1.66 so bumping this far is not acceptable for us.

Bumping to 1.65 from 1.66 doesn't really improve much so its not worth the effort. We are currently waiting for fixefox/ubuntu to bump to 1.67

Issue #8615, opened 2023-10-25
Master branch docs are a 404


NukeManDan:
### Summary

https://docs.helix-editor.com/master

![image](https://github.com/helix-editor/helix/assets/35669742/9553e291-633f-4ae0-bd37-53df28336f0b)

### Platform

All

### Terminal Emulator

Any

### Helix Version

N/A

the-mikedavis:
I believe this happens every time we cut a release: the new release wipes out the master directory. In the short term, pushing a new commit to master should fix this but long term we should preserve that master directory

archseer:
Resolved on latest master

PR #8617, opened 2023-10-26
Update the url of documentation

lens0021:
Replaces `https://docs.helix-editor.com/master/languages.html` which is 404


kirawi:
See https://github.com/helix-editor/helix/issues/8615 but this change looks good anyway now that multi-lsp is in a release.

PR #8624, opened 2023-10-26
switch to tree-sitter-ron

### LINKED ISSUE #8578: Usage of `rust` grammar for `ron` files causes parsing inaccuracies
Currently, Helix uses `grammar = "rust"` for `ron` files. `ron` syntax is slightly different from Rust's so this creates fundamental errors in parsing of ron files and messes up things like auto-indentation and syntax node selection. This has become a major annoyance for me when working with ron files.

Looks like https://github.com/zee-editor/tree-sitter-ron exists, so that would be worth considering.

gabydd:
fixes #8578 

here is the pr @Omnikar its based on the queries from https://github.com/nvim-treesitter/nvim-treesitter/tree/master/queries/ron and https://github.com/zee-editor/tree-sitter-ron/tree/main/queries changed to work with helix

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1676,7 +1676,10 @@ file-types = ["ron"]
 roots = []
 comment-token = "//"
 indent = { tab-width = 4, unit = "    " }
-grammar = "rust"
+
+[[grammar]]
+name = "ron"
+source = { git = "https://github.com/zee-editor/tree-sitter-ron", rev = "7762d709a0f7c1f9e269d0125a2e8a7a69006146" }
 
 [[language]]
 name = "robot"
--- a/runtime/queries/ron/highlights.scm
+++ b/runtime/queries/ron/highlights.scm
@@ -1,1 +1,43 @@
-; inherits: rust
+; Literals
+;------------
+
+(string) @string
+(boolean) @constant.builtin.boolean
+(integer) @constant.numeric.integer
+(float) @constant.numeric.float
+(char) @constant.character
+
+; Structs
+;------------
+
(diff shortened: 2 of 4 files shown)


languages.toml
```diff
@@ -1676,7 +1676,11 @@ file-types = ["ron"]
 roots = []
 comment-token = "//"
 indent = { tab-width = 4, unit = "    " }
-grammar = "rust"
+grammar = "ron"
```
the-mikedavis, line 1679: since the name matches the language name we don't need to specify this key - it will default to this value

runtime/queries/ron/highlights.scm
```diff
@@ -1 +1,44 @@
...
+"[" @punctuation.bracket
+"]" @punctuation.bracket
+"{" @punctuation.bracket
+"}" @punctuation.bracket
+
+
+; Special
+;------------
+(escape_sequence) @constant.character.escape
+(ERROR) @error
```
the-mikedavis, line 44: In general we have avoided adding this to queries. Eventually we want to make this a config option, so when the option is enabled this is automatically appended to the query text and when the config option is off we just use the queries untouched. For now let's remove this line since the error highlights tend to be really noisy when typing

gabydd: thats fair, I kept it in there because ron doesn't have an lsp so this will show places where there is an error but I can see how this adds extra noise

the-mikedavis (approved):

Omnikar (approved):
LGTM

pascalkuthe (approved):

PR #8635, opened 2023-10-27
Import obsd patch

daytime-brick:
not sure why not given before but here are some openbsd patches. there is one more but not sure if needed. dont know why fetch disabled. maybe takes too long or ulimit hit? may need to contact author about it or questions.

ref: https://github.com/openbsd/ports/tree/master/editors/helix/patches

feel free to edit

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,8 @@ members = [
   "xtask",
 ]
 
+resolver = "2"
+
 default-members = [
   "helix-term"
 ]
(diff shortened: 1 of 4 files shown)


the-mikedavis (changes requested):
I don't think we need either of these patches. They're disabling grammar fetching with patches although there is already a way to do this by setting the `HELIX_DISABLE_AUTO_GRAMMAR_BUILD` environment variable. Packagers typically want to disable grammar fetching if they build them separately. For example, the Nix flake uses the same idea:

https://github.com/helix-editor/helix/blob/f992c3b5972dbe2432ceb55bc8d47fed912f88bf/flake.nix#L130-L131

pascalkuthe:
Closing as these patches don't make sense to upstream

Issue #312, opened 2021-06-19
RFC: Remove window management features in favor of Kakoune-style client-server model?


malte-v:
One thing I really like about Kakoune is that it lets me use my X11 window manager for window management (duh) while still sharing buffers across windows. Kakoune achieves this using a sort of client-server model where one Kakoune process can attach to another one running in a different terminal, so effectively multiple terminals are controlled by a single editor instance. You could alternatively use tmux if you want the "vim experiece", but I really enjoy not having to think about two layers of window management (editor windows/desktop windows). One other nice consequence is that you can arbitrarily mix windows containing editor buffers and windows belonging to other (graphical) applications.

What are your thoughts?

malte-v:
I guess you could also keep the built-in window splitting features if people want to have that. It would just make things more complicated since one has to distinguish between windows from different terminals and window splits.

pickfire:
I think kakoune approach is nice for people who can do many terminals (like the time when I was using a tiling vm). Otherwise it doesn't help much. I still think it would be nice to have it, not sure how to implement it though.

cessen:
I'm against removing the built-in splitting features, as I expect that to be my primary workflow.  I don't really view it as two layers of window management.  (And come to think of it, I'm now remembering that this is one of the reasons I didn't dive more into kakoune, despite it otherwise seeming really cool.)

As for client/server capabilities, I'm pretty indifferent.  I certainly wouldn't try to stand in the way of such a feature, but don't consider it important myself.

CBenoit:
I don't really use the splitting feature because i3 does the job for me, but I think it can be useful to some people. However, I really like the client/server architecture of kakoune allowing me to open multiple windows when working on the same instance (very useful when working using multiple screens). That's not the only approach though: in (doom) emacs I used multiple windows but I'm not sure it was client/server architecture.

wojciechkepka:
> I'm against removing the built-in splitting features, as I expect that to be my primary workflow. I don't really view it as two layers of window management. 

Same for me, I use both tmux and neovim splits + tabs together and it works perfectly. I use tmux to launch a terminal next to the editor or tail logs while I use neovim splits and tabs to open separate documents.

CBenoit:
So consensus so far appears to be in favor of keeping the built-in splits.
However, I'm strongly in favor of the ability to share buffers across windows.

simias:
[...]

PR #8636, opened 2023-10-27
Dynamic command

SilasMarvin:
This pull request is meant to pair with: https://github.com/helix-editor/helix/pull/6979

It adds the `dynamic` command which is basically a placeholder command. When paired with shell expansion in the above pull request it allows for commands like the following:
```
:dynamic %sh{if [ -f Cargo.toml ]; then echo vsplit Cargo.toml; fi}
```
The above command uses a simple bash if else to check if Cargo.toml exists, before issuing the `vsplit` command with `Cargo.toml` as its argument. If `Cargo.toml` does not exist it does nothing. 

I personally use this command in concert with Wezterm to integrate tools like ripgrep for live fuzzy searching and ranger for an awesome file picking experience right from helix. Specifically, I dynamically generate `vsplit`, `hsplit`, and `open` commands depending on keystrokes that I issue, but with this dynamic command and the shell expansion in https://github.com/helix-editor/helix/pull/6979 anything becomes possible!

This is my first pull request so I apologize if I am missing some necessary tests or anything. Please let me know what you all think. Thank you!

### DIFF
--- a/book/src/generated/typable-cmd.md
+++ b/book/src/generated/typable-cmd.md
@@ -86,3 +86,4 @@
 | `:clear-register` | Clear given register. If no argument is provided, clear all registers. |
 | `:redraw` | Clear and re-render the whole UI |
 | `:move` | Move the current buffer and its corresponding file to a different path |
+| `:dynamic` | Run a potentially dynamically generated command |
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -1,5 +1,6 @@
 use std::fmt::Write;
 use std::ops::Deref;
+use std::borrow::Borrow;
 
 use crate::job::Job;
 
@@ -1047,8 +1048,7 @@ fn replace_selections_with_primary_clipboard(
     cx: &mut compositor::Context,
     _args: &[Cow<str>],
     event: PromptEvent,
-) -> anyhow::Result<()> {
-    if event != PromptEvent::Validate {
+) -> anyhow::Result<()> { if event != PromptEvent::Validate {
         return Ok(());
     }
 
@@ -2401,10 +2401,6 @@ fn move_buffer(
     args: &[Cow<str>],
     event: PromptEvent,
 ) -> anyhow::Result<()> {
-    if event != PromptEvent::Validate {
-        return Ok(());
-    }
-
     ensure!(args.len() == 1, format!(":move takes one argument"));
     let doc = doc!(cx.editor);
 
@@ -2469,6 +2465,27 @@ fn move_buffer(
         .file_changed(new_path);
 
     Ok(())
+}
+
+fn dynamic(
+    cx: &mut compositor::Context,
+    args: &[Cow<str>],
+    event: PromptEvent,
+) -> anyhow::Result<()> {
+    if event != PromptEvent::Validate {
+        return Ok(());
+    }
+
+    if args.is_empty() {
+        return Ok(());
+    }
+
+    let command_name: &str = args[0].borrow();
+    if let Some(command) = typed::TYPABLE_COMMAND_MAP.get(command_name) {
+        (command.fun)(cx, &args[1..], event)
+    } else {
+        Err(anyhow::anyhow!("no such command: '{}'", args[0]))
+    }
 }
 
 pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
@@ -3078,6 +3095,13 @@ pub const TYPABLE_COMMAND_LIST: &[TypableCommand] = &[
         fun: move_buffer,
         signature: CommandSignature::positional(&[completers::filename]),
     },
+    TypableCommand {
+        name: "dynamic",
+        aliases: &[],
+        doc: "Run a potentially dynamically generated command",
+        fun: dynamic,
+        signature: CommandSignature::all(completers::filename)
+    }
 ];
 
 pub static TYPABLE_COMMAND_MAP: Lazy<HashMap<&'static str, &'static TypableCommand>> =


postsolar:
Currently I achieve something like this by using ydotool, because I don't use Wez or Kitty or multiplexers. The interface you propose is much cleaner which is great but it doesn't expand on the current capabilities and it still suffers from the same problem: only being able to run commands that act on a Helix process from this very process. I think Helix should just be able to be launched as a server and each Helix process would be assigned an ID upon launch, and a user would be able to send arbitrary commands to these processes. Sorry it's not directly related to your PR, but I felt like sharing my general thoughts around this situation (as well as the tricks I currently use).

kirawi:
That is part of https://github.com/helix-editor/helix/issues/312

SilasMarvin:
> Currently I achieve something like this by using ydotool, because I don't use Wez or Kitty or multiplexers. The interface you propose is much cleaner which is great but it doesn't expand on the current capabilities and it still suffers from the same problem: only being able to run commands that act on a Helix process from this very process. I think Helix should just be able to be launched as a server and each Helix process would be assigned an ID upon launch, and a user would be able to send arbitrary commands to these processes. Sorry it's not directly related to your PR, but I felt like sharing my general thoughts around this situation (as well as the tricks I currently use).

I haven't heard of ydotool I'll check it out!

I agree a client server architecture would be ideal, but I've read through the issues and talked in the matrix and it unfortunately does not seem like something that will happen anytime soon. If you are looking for something that will spawn processes that are not sub-processes of helix I would suggest using Wezterm (or another terminal multiplexer/multiplexer in general) and configuring specific keybindings for it. These can still be used while you are editing in helix.

I do disagree that it does not expand on the current capabilities. Without the `dynamic` command if I wanted to spawn a file picker to browse and open a new file I could do something like the following:
```
:open %sh{./SOME_FILE_PICKER_SCRIPT}
```
However this will only ever open it in the current view. This means if I also wanted the option to open it in a vertical and horizontal split I would need to add a total of three new key mappings in my configuration file. It also does not give the option to quite the file picker without opening something. The `dynamic` command completely solves this issue. It lets shell scripting entirely decide which action, if any, action to take. This is not a current feature of helix and one that I think is definitely missing.

Don't apologize I appreciate the thoughts!

postsolar:
> However this will only ever open it in the current view. This means if I also wanted the option to open it in a vertical and horizontal split I would need to add a total of three new key mappings in my configuration file. It also does not give the option to quite the file picker without opening something.

All of this is achievable currently. Your script can just send no key events in case you quit the picker without picking anything. It [also works](https://gist.github.com/lukepighetti/8e4a13db5bdcd68a7d83eee19051ab14) for scripts which operate not on key events but via `:insert_output` + `goto_file`. But I don't insist on it remaining the only way to do it, I too would prefer something cleaner. And wrt adding three new keybindings: how does dynamic address this?

SilasMarvin:
> > However this will only ever open it in the current view. This means if I also wanted the option to open it in a vertical and horizontal split I would need to add a total of three new key mappings in my configuration file. It also does not give the option to quite the file picker without opening something.
> 
> All of this is achievable currently. Your script can just send no key events in case you quit the picker without picking anything. It [also works](https://gist.github.com/lukepighetti/8e4a13db5bdcd68a7d83eee19051ab14) for scripts which operate not on key events but via `:insert_output` + `goto_file`. But I don't insist on it remaining the only way to do it, I too would prefer something cleaner. And wrt adding three new keybindings: how does dynamic address this?

Thanks for the feedback, let me clarify because I don't think all of this is achievable currently. 

When binding keys to commands in helix in the configuration file, the command must be specified:
```
"C-f" = ":open %sh{./SOME_FILE_PICKER_SCRIPT}"
```
No matter what the expansion `%sh{./SOME_FILE_PICKER_SCRIPT}` evaluates to, it will be opened in the current view. Not specifying a command before shell expansion causes a config error. There is no way for the expansion of `%sh{./SOME_FILE_PICKER_SCRIPT}` to specify whether it should be opened in the current view, vertically split, horizontally split, or do nothing.

The `dynamic` command fixes this by acting as a placeholder command, not causing a config error, expanding  `%sh{./SOME_FILE_PICKER_SCRIPT}` and then executing the resulting command.

For example:
```
"C-f" = ":dynamic %sh{./SOME_FILE_PICKER_SCRIPT}"
```
Where `%sh{./SOME_FILE_PICKER_SCRIPT}` might evaluate to:
- `open test.xt`
- `vsplit test.txt`
- `hsplit test.txt`
- `''`
and the `dynamic` command would execute it. In the case of `''`, the `dynamic` command would do nothing.

The `dynamic` command can be used for much more than just file picking. This is just a simple example to outline its use case.

postsolar:
> let me clarify because I don't think all of this is achievable currently, although I may just be misunderstanding the example you linked

You are right, with the linked example it is not achievable. The other disadvantage of the linked method is that it can't open files on specific lines and columns. But it's all achievable with any script which is able to send key events, be it ydotool, or Kitty/Wezterm, or a standalone multiplexer. Here is how I do it with fish and ydotool:
```fish
  …
  set -l pick (open-picker-in-new-window | string join ' ') 
  sleep 0.05
  if test $pick != "" # no-op if nothing was picked
    ydotool type ":o $pick" -d 1 -H 1 # 1ms delay/hold instead of 20ms
    ydotool key 28:1 28:0 # -- Enter
  end
  …
```

> The dynamic command can be used for much more than just file picking. This is just a simple example to outline its use case
I would like to hear about your ideas. What else do you use or intend to use it for?

SilasMarvin:
> > let me clarify because I don't think all of this is achievable currently, although I may just be misunderstanding the example you linked
> 
> You are right, with the linked example it is not achievable. The other disadvantage of the linked method is that it can't open files on specific lines and columns. But it's all achievable with any script which is able to send key events, be it ydotool, or Kitty/Wezterm, or a standalone multiplexer. Here is how I do it with fish and ydotool:
> 
> ```fish
>   …
>   set -l pick (open-picker-in-new-window | string join ' ') 
>   sleep 0.05
>   if test $pick != "" # no-op if nothing was picked
>     ydotool type ":o $pick" -d 1 -H 1 # 1ms delay/hold instead of 20ms
>     ydotool key 28:1 28:0 # -- Enter
>   end
>   …
> ```
> 
> > The dynamic command can be used for much more than just file picking. This is just a simple example to outline its use case
> 
> I would like to hear about your ideas. What else do you use or intend to use it for?

Ah I see now how it is achievable. I'm sorry I misunderstood before. Sending key events back is very interesting. I didn't think of that method before. I agree though, we should have multiple ways to run helix commands produced by shell commands.

After further thought I think its main use case would be conditionally opening files (although I'm sure others could find at least some other command we want to run conditionally). I currently only use it to determine which method (if any) to open files with from my various fuzzy searches and file picker. Specifically, I use this tool along with the expect, fzf, and ranger programs to integrate really nice fuzzy search, fuzzy file finding (both with live previews thanks to bat), and a great file tree explorer and manipulator.

I think its real value is in providing flexibility. It gives users more options and control over how they want to integrate commands generated from bash scripts.

postsolar:
> Sending key events back is very interesting. I didn't think of that method before.

Sorry for the confusion, I thought this is what you were implying as well when mentioning Wez/Kitty/multiplexers.

I will be using your PR in my builds as, as I mentioned, it's a cleaner way to do these things. I will also watch for further developments. I will report back if I see any problems with this or have suggestions.

SilasMarvin:
> Sorry for the confusion, I thought this is what you were implying as well when mentioning Wez/Kitty/multiplexers.
> 
> I will be using your PR in my builds as, as I mentioned, it's a cleaner way to do these things. I will also watch for further developments. I will report back if I see any problems with this or have suggestions.

Don't apologize at all. Thank you for the feedback and new ideas. Let me know if anything doesn't work for you, or if I can alter my pr at all!

the-mikedavis:
This reminds me of Kakoune's exec / eval: https://github.com/mawww/kakoune/blob/6a39ac224bf0407b0ec5d39ffee72d6308b73cfc/doc/pages/execeval.asciidoc

For Kakoune that kind of scripting makes sense since it integrates with most everything through the shell. For Helix I would prefer that functionality like this eventually be provided by Scheme rather than shell - see #8675.

SilasMarvin:
> For Kakoune that kind of scripting makes sense since it integrates with most everything through the shell. For Helix I would prefer that functionality like this eventually be provided by Scheme rather than shell - see #8675.

Wow that is a super cool pr!

I agree that is probably a better system overall to move towards, and I would also prefer to work in Scheme than scripting in bash. I honestly do think there is some value in letting users choose which method they want to implement. If someone wants to script in bash, doesn't want to mess around with plugins, and it only requires us adding a command like the one here, I think that is probably worth it.

cammm:
I like how simple this solution is, it could really solve a lot of integrations in the meantime while plugins are being sorted out. I could also see a lot of smaller ad-hoc key bindings get implemented with this that building a plugin might be too much.

Kaporos:
Hi!

Would it be possible to make dynamic support commands chaining to allow a script to return a list of commands to execute 🤔

SilasMarvin:
> Hi!
> 
> Would it be possible to make dynamic support commands chaining to allow a script to return a list of commands to execute 🤔

This is an interesting idea. I don't see why not, I will look into it! 

I should say that before making more changes I am hoping to see some more activity from maintainers on whether this pull request is one they have interest in merging in. If there is no interest to merge it in, I do understand, but can't make any promises about updates being made.

the-mikedavis:
I don't intend on merging this: scripting should be done through the plugin system / scheme and I don't want to introduce competing tools to do the same thing (shell vs. scheme)

SilasMarvin:
> I don't intend on merging this: scripting should be done through the plugin system / scheme and I don't want to introduce competing tools to do the same thing (shell vs. scheme)

No worries I can close this pr. Is the plan to merge in: https://github.com/helix-editor/helix/pull/8675 

The scripts I have that use this dynamic command are essential to me working with helix and I would like to migrate them to the plugin system / scheme.

the-mikedavis:
Yep we want to go with #8675. I haven't looked at that branch in depth yet and it's a large change so I'm not sure how close it is to being mergeable. In the meantime you can keep this branch open or even implement whatever commands you need directly in rust in a fork.

PR #8644, opened 2023-10-27
Add a json schema for the configuration

### LINKED ISSUE #2894: config.toml autocomplete
Would be convenient to have autocomplete built-in for the helix config.toml file, suggest keys, valid values, sections, etc.

jneem:
Automatically derives a json schema from the configuration struct. This uses the `schemars` crate which checks the `serde` annotations to make sure that, e.g., names are correctly renamed to kebab-case. 

I added a little binary to generate the schema as a text file. I'm not sure what to *do* with the result (e.g. maybe it should be automatically built and uploaded somewhere as part of a release).

This should basically fix #2894, as taplo apparently supports autocompleting from json schemas.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -395,6 +395,12 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "56ce8c6da7551ec6c462cbaf3bfbc75131ebbfa1c944aeaa9dab51ca1c5f0c3b"
 
 [[package]]
+name = "dyn-clone"
+version = "1.0.14"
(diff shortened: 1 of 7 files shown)


pascalkuthe:
there was a past PR for this where it turned out that achieving generally correct schemas requires a bunch of effort.

Since the TOML config will be removed in the future I don't think it makes sense to invest that effort: https://github.com/helix-editor/helix/issues/3901#issuecomment-1737898373

pascalkuthe:
https://github.com/helix-editor/helix/pull/6174 for reference

PR #8645, opened 2023-10-27
Bothersome softwrapping of only a newline character

Omnikar:
When a line perfectly fills the viewport width and softwrap is enabled, the newline character will be softwrapped, leading to an extra blank line below.

Screenshot 2023-10-27 at 17 47 29

This behavior can be mitigated by simply changing the inequality that checks for softwraps to a strict inequality, which this PR does. However, this change has the side effect that when the cursor is on the newline, it will be off-screen. To be clear, these are both problems exclusively experienced when a line has the exact number of characters to fill the width of the viewport. These two behaviors are both rather bothersome, so I am not sure which is preferable, so it's up to the maintainers whether to accept this PR or not.

### DIFF
--- a/helix-core/src/doc_formatter.rs
+++ b/helix-core/src/doc_formatter.rs
@@ -293,7 +293,7 @@ impl<'t> DocumentFormatter<'t> {
 
         loop {
             // softwrap word if necessary
-            if word_width + self.visual_pos.col >= self.text_fmt.viewport_width as usize {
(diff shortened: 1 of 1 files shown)


pascalkuthe:
Newline characters are supposed to be always selectable.  That's a core part of the helix editing model. Hiding them is not acceptable. The current behavior is intentional and was a deliberate design choice in #5420

Omnikar:
Alright.

PR #8652, opened 2023-10-28
fix: replace kdl tree-sitter to fix highlighting

cgahr:
The tree-sitter for the kdl language currently being used contains bugs.

Consider the following kdl config:
```kdl
bigFlatItems "multiline \" \" \"\nstring\n" null
c "multiline string\nwith special characters:\n\t \n \" \"\n"
```

With the current tree-sitter implementation, escaped double quotes `\"` are being detected as characters ending a string. As such, they are improperly highlighted.

I propose to change the used implementation to https://github.com/amaanq/tree-sitter-kdl. It seems to be more up to date, more activate and highlights the previous example correctly.

grafik

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -81,7 +81,7 @@
 | jsx | ✓ | ✓ | ✓ | `typescript-language-server` |
 | julia | ✓ | ✓ | ✓ | `julia` |
 | just | ✓ | ✓ | ✓ |  |
-| kdl | ✓ |  |  |  |
+| kdl | ✓ | ✓ | ✓ |  |
 | kotlin | ✓ |  |  | `kotlin-language-server` |
 | latex | ✓ | ✓ |  | `texlab` |
 | lean | ✓ |  |  | `lean` |
--- a/languages.toml
+++ b/languages.toml
@@ -2334,7 +2334,7 @@ injection-regex = "kdl"
 
 [[grammar]]
 name = "kdl"
-source = { git = "https://github.com/Unoqwy/tree-sitter-kdl", rev = "e1cd292c6d15df6610484e1d4b5c987ecad52373" }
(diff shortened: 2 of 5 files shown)


the-mikedavis (changes requested):
The queries in `runtime/queries/kdl/*.scm` will need to be updated to match the new grammar. For example `cargo xtask query-check` is currently failing because the queries match a `comment` node which doesn't exist in the new grammar (it has a `single_line_comment` node instead https://github.com/amaanq/tree-sitter-kdl/blob/e180e05132c4cb229a8ba679b298790ef1eca77c/grammar.js#L291)

cgahr:
Thanks for your comments. I updated `highlights.scm` to match the new tree-sitter. I had to remove `bare_identifier` since it does not exist (is not public) in the new tree-sitter.

Also, I added indent queries.

cgahr:
I added text object queries for tree sitter navigation:
- `class.*` and `function.*` behave the same but the movement
- both match the full node, `inside` the inside of the brackets, `around` the full surrounding node
- `class.movement` moves to the next node
- `function.movement` moves to the next identifier, I think this is very useful to cycle identifiers and is not bound to the hierarchy (in contrast to `class.movement`)
- parameter work as well. `inside` selects the parameter, `around` the parameter with type if
- `parameter.movement` moves to the next parameter without type, I think this matches the most common use-case of only editing the parameter without type
- comment matching works too, however, `comment.around` for single line comments only selects the same and previous lines, not the following ones too. I don't know how to fix that.
- lastly, `test.*` matches the types of values; tests don't exist in kdl and this type of movement might be useful but probably not very common.

I also added indents, they just work.


Lastly, I slightly improved the syntax highlighting. node identifiers are now variable colored, which afaik is normally the most common color, white. This removes colors, with them being the most common element of kdl documents, and removes color clashes with type annotations. Also, types and their brackets are differently colored.

the-mikedavis (approved):

pascalkuthe (approved):

PR #8658, opened 2023-10-29
feat: add variable segment movements and textobject

gertqin:
Hi,

I only discovered helix a few days ago, but I've already (more or less) switched over from nvim - so thanks for an awesome editor! 🙂 

However, one plugin I really like in nvim is [vim-textobj-variable-segment](https://github.com/Julian/vim-textobj-variable-segment) which makes it a lot easier to navigate variables in title/camel/snake-case and edit parts of it.

In this draft PR I have made a simple implementation of it in helix, i.e. added
- `move/extend_prev_partial_word_start` (keybind `H`)
  - Ex `let |VarInTitleCase` => `let |Var|InTitleCase` => `let Var|In|TitleCase`
- `move/extend_next_partial_word_end` (keybind `L`)
  - Ex `let var_in_snake_case| ` => `let var_in_snake|_case|`  => `let var_in|_snake|_case` 
- textobject `partial word` (keybind `v` in match inside/outside mode)
  - Ex `let var_in_sn|ake_case ` => mav => `let var_in|_snake|_case `
  - Ex `let var_in_sn|ake_case ` => miv => `let var_in_|snake|_case `
 
(I named everything `partial_word` in the code, because that's what I thought the vim plugin was called until I looked it up...)

Do you think these commands belong to the helix core? If that's the case then I'll add unit tests to the PR. 

I'll also gladly rename `partial_word` to something better if you can think of something, and you probably also want to reserve the keys `H` and `L` for something else.. (I think the movements could be added without default keybindings).

### DIFF
--- a/helix-core/src/chars.rs
+++ b/helix-core/src/chars.rs
@@ -11,6 +11,15 @@ pub enum CharCategory {
     Unknown,
 }
 
+#[derive(Debug, Eq, PartialEq)]
+pub enum WordCharCategory {
+    Underscore,
+    Numeric,
+    UpperCase,
+    LowerCase,
+    Unknown,
+}
+
 #[inline]
 pub fn categorize_char(ch: char) -> CharCategory {
     if char_is_line_ending(ch) {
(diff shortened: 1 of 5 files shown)


kirawi:
Duplicate of https://github.com/helix-editor/helix/pull/8147 (leave this open for the maintainers to see which solution is preferred)

RoloEdits:
#8147 was merged, this can now be closed

PR #8660, opened 2023-10-29
rust: add highlight scope for type parameters

jmhain:
Highly generic code can be challenging to read when type parameters are highlighted the same as other types. This is especially true when the type parameters are not single letters, e.g. `Fut` is somewhat common in Rust. 

So this PR adds a new highlighting scope for type parameters. I added it only for Rust as a proof-of-concept, but if accepted I'll update the queries for any other languages I'm familiar with (either in this PR or in a follow-up, whichever is preferred).

Here is a before and after with my in-progress port of the [Codely theme](https://plugins.jetbrains.com/plugin/12891-codely-theme).

image

image

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -155,6 +155,7 @@ We use a similar set of scopes as
 
 - `type` - Types
   - `builtin` - Primitive types provided by the language (`int`, `usize`)
+  - `parameter` - Generic type parameters (`T`)
   - `enum`
     - `variant`
 - `constructor`
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -13,6 +13,8 @@
   (type_identifier) @type.parameter)
 (constrained_type_parameter
   left: (type_identifier) @type.parameter)
+(optional_type_parameter
+  name: (type_identifier) @type.parameter)
 
 ; ---
 ; Primitives
--- a/runtime/queries/rust/highlights.scm
+++ b/runtime/queries/rust/highlights.scm
@@ -9,6 +9,11 @@
 ; Types
 ; -------
 
+(type_parameters
+  (type_identifier) @type.parameter)
+(constrained_type_parameter
+  left: (type_identifier) @type.parameter)
+
 ; ---
 ; Primitives
 ; ---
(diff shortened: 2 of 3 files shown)


kirawi (approved):
Looks good.

kirawi:
There is also [`optional_type_parameter`](https://github.com/tree-sitter/tree-sitter-rust/blob/48e053397b587de97790b055a1097b7c8a4ef846/grammar.js#L562C34-L569) (which I didn't know was a thing until now) seen in
```rs
fn foo<T = Roo>() {
}
```
with an AST of
```
[source_file](https://tree-sitter.github.io/tree-sitter/playground#) [0, 0] - [2, 0]
  [function_item](https://tree-sitter.github.io/tree-sitter/playground#) [0, 0] - [1, 1]
    name: [identifier](https://tree-sitter.github.io/tree-sitter/playground#) [0, 3] - [0, 6]
    type_parameters: [type_parameters](https://tree-sitter.github.io/tree-sitter/playground#) [0, 6] - [0, 15]
      [optional_type_parameter](https://tree-sitter.github.io/tree-sitter/playground#) [0, 7] - [0, 14]
        name: [type_identifier](https://tree-sitter.github.io/tree-sitter/playground#) [0, 7] - [0, 8]
        default_type: [type_identifier](https://tree-sitter.github.io/tree-sitter/playground#) [0, 11] - [0, 14]
    parameters: [parameters](https://tree-sitter.github.io/tree-sitter/playground#) [0, 15] - [0, 17]
    body: [block](https://tree-sitter.github.io/tree-sitter/playground#) [0, 18] - [1, 1]
```

jmhain:
Thanks for catching that @kirawi, I pushed a commit to handle that case

cd-a:
Love it. Thanks!

As you mentioned, that would be awesome for other languages like TypeScript as well.

jmhain:
@cd-a I was planning on doing TypeScript next :slightly_smiling_face:

the-mikedavis (approved):
This makes sense to me 👍

pascalkuthe (approved):
I think this is generally a nice feature. Note that #1151 will cause issues here more frequently since generieren parameters often have larger scopes than function Parameters (impl Block in rust for example)

Issue #1730, opened 2022-03-03
Continue comments


antoyo:
Hi.

#### Describe your feature request

I'd like a feature in vim with the setting `formatoptions=ro`, that is something that will continue a comment when you press `Enter` in insert mode or open a new line inside a comment.

Thanks!

seanchen1991:
This is a feature that I'm quite sorely missing. I'd love to see it get merged soon!

antoyo:
I'll see if I can cleanup this PR in January. I've been quite busy lately.

icefoxen:
Not the same, but related: using `J` to join lines in comments leaves the stray comment markers in the line.  So if I have:

```
// foo
// bar
```

and I hit `J` in normal mode I get:


```
// foo// bar
```

Interestingly, `:reflow` handles both cases correctly.

dist1ll:
Note that vim also carries over the preceding whitespace of the *current comment line*.

```
/* Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do 
 * eiusmod tempor incididunt ut labore et dolore magna aliqua.
 *
 *       push   %rax
 *       jmpq   *%rax▎
 *       ▎           ^ 
         ^       if you press enter here
       cursor
     lands here 

...
 */
```

This is very useful when writing comments with varying levels of indentation.

johnbchron:
Hey this really feels like it should be a config-switched feature, and probably disabled by default. This is a "breaking change" experience-wise, especially for people who don't want to use this feature.

(For reference, my specific use case is writing doc comments for Rust struct fields. Most of the time they're just 1 line, and then when I hit newline I have to delete the comment characters before I write the field)

johnbchron:
@TornaxO7 The point is not that I don't like it, but that it's a _breaking change_ for user experience. For people who don't want or expect this (not just me), there's no way to go back to the previous behavior unless we add a config option. Whether or not it's on by default is a different matter.

inferiorhumanorgans:
> @johnbchron I could try in the future to add a config option or this. But may I ask why you aren't satisfied by pressing `Ctrl` + `w` instead to remove the lien comment?

Because that's an extra keystroke.  I wasn't thrilled at the surprise UX change, but I'd much rather have the ability to start a new line on demand without having to clear it.  E.g. modifier+return = new line, return = new line continue comment.

PR #1937, opened 2022-04-03
Continue comments when opening a new line

antoyo:
Fix #1730.

We probably want to be able to configure whether we want this or not. If so, what option name do you propose?

This does not support continuing block comment, i.e. adding a `*` when a comment starts with `/*`. I propose we do this in a future PR since this would require some more thought regarding the block comment token (i.e. do we want to add it to `languages.toml`?).

PR #8664, opened 2023-10-30
feat: continue single-line comments

seanchen1991:
Closes: #1730 

This is a re-attempt at implementing continue-comment functionality, following the earlier attempt in #1937; I lifted most of @antoyo's logic from his initial PR. 

This PR only implements continuation of single-line comments; block comments are left as a TODO. Additionally, I attempted to implement proper indentation in new comment lines as per [this](https://github.com/helix-editor/helix/issues/1730#issuecomment-1651698674) comment, but was finding it really finicky to get right, and so I'm opening this PR without proper handling of indentation in continued comments for now. I'll try to open a subsequent PR that adds that logic at a later time. 

The functionality is behind a `continue-comments` configuration option that defaults to enabled. Additionally, a `comment-tokens` field was added for each language in `languages.toml` that accepts an array of strings in order to specify which comment tokens to look for. I think there's two ways to handle this: introduce the new `comment-tokens` field and deprecate the old `comment-token` field, or implement some logic that parses a string of comma-separated tokens in order to not introduce a breaking change. Which would be the preferred route here?


helix-core/src/comment.rs
```diff
@@ -94,12 +94,56 @@ pub fn toggle_line_comments(doc: &Rope, selection: &Selection, token: Option<&st
     Transaction::change(doc, changes.into_iter())
 }
 
+/// Return the comment token of the current line if it is commented.
+///
+/// Return None otherwise.
+pub fn get_comment_token<'a>(line: &RopeSlice, tokens: &'a [String]) -> Option<&'a str> {
+    // TODO: don't continue shebangs
```
seanchen1991, line 101: This TODO was carried over from #1937, but I'm not sure what it means.

the-mikedavis: About shebangs in general: https://en.wikipedia.org/wiki/Shebang_(Unix)

We shouldn't continue shebangs as if they were regular comments since they only take up one line

the-mikedavis:
#4718 also adds support for multiple tokens. We will most likely merge that PR before we can give this a proper look.

If you're resubmitting work from another PR it's customary to give the original author credit, either by re-using their commits without changes or by giving them co-author on the commits where you re-use their code: https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors

seanchen1991:
> If you're resubmitting work from another PR it's customary to give the original author credit

Is there a way to apply these changes to commits retroactively?

the-mikedavis:
See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/changing-a-commit-message#amending-older-or-multiple-commit-messages

Rebasing / ammending old commits can be complicated though. We could instead add a co-authored-by tag at merge-time if we squash + merge

seanchen1991:
> We could instead add a co-authored-by tag at merge-time if we squash + merge

Sounds good to me🙂

TornaxO7:
@seanchen1991 just a little reminder that the [toogle-block-comments PR](https://github.com/helix-editor/helix/pull/4718) if you haven't noticed it yet :)

djugei:
This change seems to be quite far along, what additional work is necessary to get it merged?

PR #8667, opened 2023-10-30
Only render preview if picker has a preview function

### LINKED ISSUE #8665: Command Palette Shows Preview
Small UX enhancement request.
The default for pickers is to show preview, however this isn't a sensible default for the command palette picker.
Is there a way to distinguish between different pickers to change the default for showing the preview split for command palette picker and file picker?

rmehri01:
Fixes #8665


the-mikedavis (approved):

pascalkuthe (approved):
Thanks!

PR #8676, opened 2023-10-31
Resolve args.files before changing directory

bjorn-ove:
As discussed with @the-mikedavis in https://github.com/helix-editor/helix/issues/8520#issuecomment-1784158957

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -161,20 +161,20 @@ impl Application {
             editor.open(&path, Action::VerticalSplit)?;
             // Unset path to prevent accidentally saving to the original tutor file.
             doc_mut!(editor).set_path(None);
-        } else if let Some((first, _)) = args.files.first() {
-            let mut skip_files = 0;
+        } else if !args.files.is_empty() {
+            let mut files_it = args.files.into_iter().peekable();
 
             // If the first file is a directory, skip it and open a picker
-            if first.is_dir() {
-                let picker = ui::file_picker(first.into(), &config.load().editor);
+            if let Some((first, _)) = files_it.next_if(|(p, _)| p.is_dir()) {
+                let picker = ui::file_picker(first, &config.load().editor);
                 compositor.push(Box::new(overlaid(picker)));
-                skip_files = 1;
             }
 
             // If there are any more files specified, open them
-            if args.files.len() > skip_files {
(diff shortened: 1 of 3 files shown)


pascalkuthe:
I think you can now simplify the logic to open the CWD again (get rid of the `open_cwd` flags and instead simply restore the logic to how it was before #8520)

bjorn-ove:
Good point

bjorn-ove:
One question that may be worth asking now is what is the logical behaviour of running `helix -w ~/.config ~/.ssh` or `helix -w ~/.config ~/.config/helix`

Currently it will effectively ignore `-w`, but perhaps it is better to change this, so it honors `-w` for working directory, and opens a picker at the first argument?

the-mikedavis:
Yeah, I would prefer `-w` to take precedence, so we would set the cwd to `~/.config` in that first example and open a picker for `~/.ssh`

bjorn-ove:
Ok, I made two changes now.

The first as discussed above, the second to open the remaining files in args.files when the first one is a directory.

helix-term/src/application.rs
```diff
@@ -161,15 +161,20 @@ impl Application {
             editor.open(&path, Action::VerticalSplit)?;
             // Unset path to prevent accidentally saving to the original tutor file.
             doc_mut!(editor).set_path(None);
-        } else if !args.files.is_empty() {
-            if args.open_cwd {
-                // NOTE: The working directory is already set to args.files[0] in main()
-                editor.new_file(Action::VerticalSplit);
-                let picker = ui::file_picker(".".into(), &config.load().editor);
+        } else if let Some((first, _)) = args.files.first() {
+            let mut skip_files = 0;
```
the-mikedavis, line 165: It should be a little cleaner here to make an iterator over the files so we don't need to keep track of indices:

```rust
let files = args.files.iter(); // maybe `.peekable()` too if we want to check if the iterator is empty
// ..
if let Some((first, _)) = files.next() {
    // ..
}
```

bjorn-ove: I agree, just tried to keep refactoring to a minimum.

bjorn-ove:
Updated after feedback

the-mikedavis (approved):

pascalkuthe (approved):

PR #8678, opened 2023-10-31
Recognize `.ispc`  extension as CPP

MarijnS95:
Intel's Implicit SIMD Program Compiler (https://ispc.github.io/) files are "a variant of the C programming language" with support for templates and references.

Unfortunately they also use some custom keywords and constructs that make this language unsuitable for `clangd`, but the CPP grammar works well enough for most highlighting.  There is supposedly a user-made tree-sitter grammar: https://github.com/tree-sitter/tree-sitter/pull 2282

---

Creating this as draft as we should probably not use `clangd` as LSP here, and perhaps switch to https://github.com/fab4100/tree-sitter-ispc for grammar.


cloudhan:
ISPC has language server is pending a release https://togithub.com/ispc/ispc/issues/1605#issuecomment-1771833547

the-mikedavis:
Let's focus this PR on using tree-sitter-ispc and the ispc language server. For those looking for some basic syntax highlighting for ispc in the mean time, the current change can be added to your custom language config:

```toml
[[language]]
name = "cpp"
file-types = ["cc", "hh", "c++", "cpp", "hpp", "h", "ipp", "tpp", "cxx", "hxx", "ixx", "txx", "ino", "C", "H", "cu", "cuh", "cppm", "h++", "ii", "inl", "ispc", { suffix = ".hpp.in" }, { suffix = ".h.in" }]
```

MarijnS95:
Imported the grammar from https://github.com/fab4100/tree-sitter-ispc.

Somewhat strangely some of the queries were removed in https://github.com/fab4100/tree-sitter-ispc/commit/0197076648b8f31aab85c447015f29dadb53a8b3 and have been moved into https://github.com/nvim-treesitter/nvim-treesitter/tree/7da3cb6323d1e1fbfc4d7dc05cac74946458b8a0/queries/ispc.

runtime/queries/ispc/highlights.scm
the-mikedavis, line 1: The queries will need to be adjusted - neovim queries don't work out of the box for Helix. For example some captures need to be adjusted to match the ones we use (https://docs.helix-editor.com/master/themes.html#syntax-highlighting) like `@repeat` should be `@keyword.control.repeat`. `#lua-match?` will need to be replaced with an equivalent `#match?`

runtime/queries/ispc/indents.scm
the-mikedavis, line 1: For indents see https://docs.helix-editor.com/master/guides/indent.html

runtime/queries/ispc/locals.scm
the-mikedavis, line 1: For locals we follow tree-sitter: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

So these should be changed to only use the `@local.definition`, `@local.reference` and `@local.scope` captures

MarijnS95: Thanks for these pointers! I've quickly cobbled some changes together and now understand a lot better how these are supposed to work.

There's still more to do, and I should likely also contribute the C changes separately, which enable parameter highlights for `int x[10]` and `int **x` (only `int *x` was supported).

runtime/queries/ispc/highlights.scm
fab4100: The two sets of queries exist because neovim has its own query engine with some extensions as mentioned above.  Unfortunately there is no common ground for query syntax at the moment.  The queries in the ISPC parser repo are very basic and minimal (based off the C language parser).  As mentioned, you will need to adjust and modify for your needs.  If you have additions that will work for general queries (without custom engine syntax) I am happy to add them in the ISPC parser repo.  Both query sets (in the ISPC parser repo and nvim-treesitter) are currently maintained (the one in nvim-treesitter more actively however).

runtime/queries/ispc/locals.scm
the-mikedavis: Yep feel free to add some locals for C and re-use them if possible, that should be a nice improvement to C highlighting

MarijnS95: Factored out to a separate file :+1:

runtime/queries/ispc/highlights.scm
MarijnS95: I'll unfortunately be unavailable for the next month but hope to pick this back up after that and finalize this PR!

PR #8684, opened 2023-10-31
Add missing license files for themes

blinxen:
Some themes are based on `MIT` licensed themes / code. For this reason we need to add  the original license text to be able to redistribute them. This is required for Fedora. I can add them myself in Fedora but having the files upstream makes it easier :D.

### DIFF
--- a/runtime/themes/licenses/everforest.LICENSE
+++ b/runtime/themes/licenses/everforest.LICENSE
@@ -1,0 +1,21 @@
+MIT License
+
+Copyright (c) 2019 sainnhe
+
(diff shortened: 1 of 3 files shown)


blinxen:
One theme does not have a license text (yet), so I also opened a PR there --> https://github.com/arturoalviar/serika-syntax/pull/2

pascalkuthe (approved):
thanks for taking care of these things, I would be alright with keeping these files upstream

the-mikedavis (approved):

PR #8689, opened 2023-11-01
feat: make more commands exit select mode

### LINKED ISSUE #8687: Indent/unindent should exit select mode
Just like all other operations (paste, yank, etc) indent/unindent should exit select mode. This can already be achieved with a custom mapping, but I think it should be the default.

```
[keys.select]
">" = ["indent", "normal_mode"]
"<" = ["unindent", "normal_mode"]
```

rmehri01:
Fixes #8687 

The issue also mentioned that paste exits select mode but I don't think that is the case, only replace seems to. Should we also add exiting select mode to paste or any other commands?

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -984,6 +984,7 @@ fn align_selections(cx: &mut Context) {
 
     let transaction = Transaction::change(doc.text(), changes.into_iter());
     doc.apply(&transaction, view.id);
+    exit_select_mode(cx);
 }
 
 fn goto_window(cx: &mut Context, align: Align) {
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Yeah, I think in general if we're making edits then we should exit select mode (also see https://github.com/helix-editor/helix/pull/3932#issuecomment-1257280145, and never mind https://github.com/helix-editor/helix/pull/3932#issuecomment-1256443734 - I've changed my mind since then). Having paste exit select mode definitely makes sense since we already exit select mode on replace.

rmehri01:
I see, that makes sense, thanks! In that case, I think some other good candidates along with paste could be `replace_selections_with_clipboard` (seems that `replace_with_yanked` exits select but not replace with clipboard), `switch_case`, `increment`/`decrement`, and maybe also `align_selections`?

archseer (approved):

the-mikedavis (approved):

PR #8696, opened 2023-11-02
Add required-root-patterns for situational lsp activation

### LINKED ISSUE #8505: Situational LSP on same file extension
## Intro
Some languages have multiple possible language servers, in my case the (in)famous Javascript/Typescript:
- `typescript-language-server`
- `deno lsp`
- even `eslint`, `tailwindcss` or any other situational language support, all for the same file types (`jsx?`/`tsx?`)

## Formal issue
Language support may depend on configured tooling in the workspace rather than just file extension. `roots` isn't requiring the existence of the files.

## Possible enhancements
1. Adding a `language.toml` flag key to actually require **one of** the `roots` files to be found to in order to enable the LSP
_Eg. `require-root = true`_
2. Adding a `language.toml` configuration key of **possible marker files** in found root that enables the LSP.
_Eg. `required-roots = ["deno.json", "deno.jsonc", "deno.lock"]`_

Option 1 seems relevant and sufficient to cover multiple LSP per file type while keeping things simple.

## Current workaround
Dropping an `.helix/languages.toml` situationally in workspaces.

ontley:
Closes #8505 

Adds the optional `required-root-patterns` config option to `language-server`s,
and only activates the language server if one or more glob patterns from `required-root-patterns` are found inside the language server's root directory


pascalkuthe (dismissed):

pascalkuthe:
I don't love having to add another config option but I think the usecase is clearly there in more fragmented ecosystems (Javascript)

helix-core/src/syntax.rs
```diff
@@ -371,6 +371,9 @@ pub struct LanguageServerConfiguration {
     pub config: Option<serde_json::Value>,
     #[serde(default = "default_timeout")]
     pub timeout: u64,
+    #[serde(default)]
+    #[serde(skip_serializing_if = "Vec::is_empty")]
+    pub required_root_patterns: Vec<String>,
```
the-mikedavis, line 376: Let's do the compilation into a globset at config parsing time instead, adding `globset` to `helix-core` (we want to do this anyways for #8006):

```rust
pub struct LanguageServerConfiguration {
    ..
    #[serde(
        default,
        skip_serializing,
        deserialize_with = "deserialize_required_root_patterns"
    )]
    pub required_root_patterns: Option<GlobSet>,
}
```

```rust
fn deserialize_required_root_patterns<'de, D>(deserializer: D) -> Result<Option<GlobSet>, D::Error>
where
    D: serde::Deserializer<'de>,
{
    let patterns = Vec::<String>::deserialize(deserializer)?;
    if patterns.is_empty() {
        return Ok(None);
    }
    let mut builder = globset::GlobSetBuilder::new();
    for pattern in patterns {
        let glob = globset::Glob::new(&pattern).map_err(serde::de::Error::custom)?;
        builder.add(glob);
    }
    builder.build().map(Some).map_err(serde::de::Error::custom)
}
```

Then the check in `start_client` becomes really simple:

```rust
if let Some(globset) = &ls_config.required_root_patterns {
    if !root_path
        .read_dir()?
        .flatten()
        .map(|entry| entry.file_name())
        .any(|entry| globset.is_match(entry))
    {
        return Ok(None);
    }
}
```

helix-lsp/src/lib.rs
```diff
@@ -888,18 +892,70 @@ fn start_client(
     doc_path: Option<&std::path::PathBuf>,
     root_dirs: &[PathBuf],
     enable_snippets: bool,
-) -> Result<NewClient> {
+) -> Result<Option<NewClient>> {
+    let (workspace, workspace_is_cwd) = helix_loader::find_workspace();
+    let workspace = path::normalize(&workspace);
```
the-mikedavis, line 897:
```suggestion
    let workspace = path::normalize(workspace);
```

I'm surprised that clippy doesn't catch this: `workspace` here already implements `AsRef<Path>` so the borrow is needless

helix-core/src/syntax.rs
ontley:
```rs
let glob = globset::Glob::new(&pattern).map_err(serde::de::Error::custom)?;
```
Would this drop the entire required roots list if any glob pattern couldn't be built? My intention was to log an error on bad patterns, but still check the ones that could be built.

the-mikedavis: The whole language configuration parse would fail, so you would see the Bad config error on startup: https://github.com/helix-editor/helix/blob/d570c29ce37ffbb46a9c49708c31dfd81daa27cf/helix-term/src/main.rs#L149-L150

I would prefer to fail early and loudly in this case since it can be confusing when the config successfully parses and the editor starts but not everything you wanted to configure works

ontley: Ah ok, makes sense then.

ontley:
Why is `cargo fmt --all --check` failing? Running locally doesn't output anything, and I ran `cargo fmt` before the last commit.

the-mikedavis:
Are you using the rustfmt from the 1.70.0 toolchain? We have rustfmt and clippy locked on that toolchain: https://github.com/helix-editor/helix/blob/master/rust-toolchain.toml

I believe using rustfmt through rustup should fix the version problem

helix-core/src/transaction.rs
```diff
@@ -378,7 +378,9 @@ impl ChangeSet {
             macro_rules! map {
                 ($map: expr, $i: expr) => {
                     loop {
-                        let Some((pos, assoc)) = positions.peek_mut() else { return; };
+                        let Some((pos, assoc)) = positions.peek_mut() else {
+                            return;
```
the-mikedavis, line 382: Newer versions of rustfmt can handle let-else statements, so these unrelated blocks of code are modified. Could you reset these changes? I would prefer to update the formatting of these when we bump the MSRV

ontley: Should be reset now.
Sorry I'm pretty new to open-source.

the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

PR #8698, opened 2023-11-02
Info on how to skip grammar build when building from source

NewtonChutney:
## Situation:
Currently, the documentation doesn't indicate users can compile Helix without building grammars

## Proposal:
Add instructions on how to compile without compiling grammars to the install doc

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -207,7 +207,7 @@ RUSTFLAGS="-C target-feature=-crt-static"
    This command will create the `hx` executable and construct the tree-sitter
    grammars in the local `runtime` folder.
 
-> ❗💡 In case you do not want to build grammars, set an environment variable `HELIX_DISABLE_AUTO_GRAMMAR_BUILD` = 1
+> 💡 If you do not want to fetch or build grammars, set an environment variable `HELIX_DISABLE_AUTO_GRAMMAR_BUILD`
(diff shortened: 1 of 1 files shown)


book/src/install.md
```diff
@@ -207,6 +207,8 @@ RUSTFLAGS="-C target-feature=-crt-static"
    This command will create the `hx` executable and construct the tree-sitter
    grammars in the local `runtime` folder.
 
+> ❗💡 In case you do not want to build grammars, set an environment variable `HELIX_DISABLE_AUTO_GRAMMAR_BUILD` = 1
```
the-mikedavis, line 210:
```suggestion
> 💡 If you do not want to fetch or build grammars, set an environment variable `HELIX_DISABLE_AUTO_GRAMMAR_BUILD`
```

A few small nits here:

* most people building from source will want to fetch & build grammars as part of the build so let's remove the `!`
* the value of the environment variable doesn't matter, just that it is defined
* that environment variable also stops the build script from fetching the grammar repositories as well as compiling them

NewtonChutney: Ahh, yess
Thanks @the-mikedavis for the review

the-mikedavis (approved):

pascalkuthe (approved):

PR #8703, opened 2023-11-02
Swap system and primary clipboard registers

### LINKED ISSUE #8702: Swapped special register keys
Hi, I've been enjoying the enhanced register support in release 23.10, but there's a minor issue that's bothering me. Coming from Vim, `+` was used to refer the system clipboard, and `*` was used to refer to the primary selection, but in Helix these are currently swapped with each other. This isn't a big deal, but it works against my muscle memory, and I imagine that might also be true for other ex-vim users with a habit of using those register keys.

Is this an intentional difference? If it makes sense to swap these keys to be consistent with Vim's, and if it would help, I could open a pull request to change the relevant code and documentation. I suppose there's also the possibility of making this behavior configurable, but I don't have a good sense of what that might take.

Omnikar:
Closes #8702

### DIFF
--- a/helix-view/src/register.rs
+++ b/helix-view/src/register.rs
@@ -75,8 +75,8 @@ impl Registers {
                 self.clipboard_provider.as_ref(),
                 self.inner.get(&name),
                 match name {
-                    '*' => ClipboardType::Clipboard,
-                    '+' => ClipboardType::Selection,
+                    '+' => ClipboardType::Clipboard,
+                    '*' => ClipboardType::Selection,
                     _ => unreachable!(),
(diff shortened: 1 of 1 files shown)


pascalkuthe (approved):

pascalkuthe:
thanks!

David-Else:
You forgot to update the documentation: https://docs.helix-editor.com/usage.html?highlight=registers#special-registers

dead10ck:
Thanks @David-Else , I have that fixed in https://github.com/helix-editor/helix/pull/8708 now

PR #8708, opened 2023-11-03
swap yank command registers

dead10ck:
#8703 swapped the `+` and `*` registers, but did not swap them in the corresponding yank commands.


pascalkuthe (dismissed):

dead10ck:
shout out to `calisti` from the Matrix chat room for finding more places that needed to be swapped

mo8it:
That should fix `yank_joined_to_clipboard` not yanking to the system clipboard since I compiled from source today 😅
I didn't test it though. I will test it after this PR gets merged.

Thanks for the fix 🥰

the-mikedavis (approved):
I looked around for a while and I'm pretty sure this is all the places that need updating. Thanks for cleaning this up!

PR #8711, opened 2023-11-03
Add an installation method field to the bug report template

the-mikedavis:
We can guess the installation method from the version tag and platform in some cases but it would be useful to have this be explicit for the sake of debugging packager-specific problems.


pascalkuthe (approved):
Good idea! It would have made it much easier to trigge the diff futter sigfault reports we got in the last release (instead of asking "are you using XYZ which is k own to be broken"

PR #8717, opened 2023-11-04
implement `goto_[first|prev|next|last]_error`

cgahr:
This PR implements the commands `goto_[first|prev|next|last]_error`. They are basically equivalent to `goto_[first|prev|next|last]_diag` but ignore all diagnostics that aren't severe enough.

It closes issue #3405 and supersedes PR #6331. The latter is very similar (and was used as an inspiration) but doesn't implement `goto_[first|last]_error`.

The four commands are mapped to `[|] + e|E` similar to `goto_[first|prev|next|last]_diag`. The latter `e` is IMO the obvious choice, `e` for error and also its proximity to `d` is nice, enhancing the existing muscle memory.

In theory, the severity level for which diagnostics are valid targets for `goto_[first|prev|next|last]_error` could be a config flag. For now, it is hard-coded to `Severity::Error`.

What do you think?


kirawi:
I feel like `first` and `last` are unnecessary because they can be achieved very easily by, respectively: `1G<goto_next_error>` and `ge<goto_prev_error>

cgahr:
Tbh, I never thought about using `gg]e` (or `gg]d` for that matter) to get the next error/diagnostic. 

In any case, I think binding the capitalized letter to first and last doesn't really waste "keymap realestate", so why not have it?

archseer:
> Edit: Actually, I see that there is precedence for this behavior, but it feels odd to me for that too.

Yes, this is a pretty well established pattern.

georgemessiha22 (approved):

aral:
Was just coming over to file an issue to request this very thing.

Any thoughts on when this might be merged? :)

CC @archseer

altanozlu:
Any news on this ?

dunxen:
> In any case, I think binding the capitalized letter to first and last doesn't really waste "keymap realestate", so why not have it?

This also aligns with first/last diagnostic behaviour so it makes sense.

cgahr:
What's new:
- I fixed the merge conflicts.
- find * error is now on the `q` key since `e` is already in use
- I fixed the behavior of `]d` to match `[d` when the cursor is in the middle of a diagnostic. That is, both match the diagnostic surrounding the cursor, and only on the next press select the prev/next diagnostic.

For the maintainers: `goto_next_or_prev_diag_impl` and `goto_first_or_last_diag_impl` contain a lot of duplicated code, but as far as I can see, there is no easy way to get rid of it. Maybe you have an idea.

beeb:
Hey! I found this PR while searching if someone already suggested adding these. What's the status, can we try to revive this?

PR #8719, opened 2023-11-04
theme: Make dark_plus.toml more accurate to VSCode

saccarosium:
The current implementation of the Dark+ has some inconsistencies with the original. I've made some modification, witch the most importants are:
1. Statusbar is now gray instead of blue, to make it more readable;
2. Whitespace are a more subtle color;
3. Indent Character is now distinguishable from the whitespaces.

## C

| Current | Updated | VSCode |
| --- | --- | --- |
| Screenshot 2023-11-04 at 21 30 00 | Screenshot 2023-11-04 at 21 28 33 | Screenshot 2023-11-04 at 21 29 43 |

## CPP

| Current | Updated | VSCode |
| --- | --- | --- |
| Screenshot 2023-11-04 at 21 41 31 | Screenshot 2023-11-04 at 21 41 19 | Screenshot 2023-11-04 at 21 41 14 |

### DIFF
--- a/runtime/themes/dark_plus.toml
+++ b/runtime/themes/dark_plus.toml
@@ -1,103 +1,87 @@
 # Author: Shafkath Shuhan <shafkathshuhannyc@gmail.com>
 
-"namespace" = { fg = "type" }
-"module" = { fg = "type" }
+# SYNTAX
+"attribute"                 = "fn_declaration"
+"comment"                   = "dark_green"
+"constant"                  = "constant"
+"constant.builtin"          = "blue2"
+"constant.character"        = "orange"
+"constant.character.escape" = "gold"
+"constant.numeric"          = "pale_green"
+"constructor"               = "type"
+"diff.delta"                = "blue4"
+"diff.minus"                = "orange_red"
+"diff.plus"                 = "dark_green2"
+"function"                  = "fn_declaration"
+"function.builtin"          = "fn_declaration"
+"function.macro"            = "blue2"
+"keyword"                   = "blue2"
+"keyword.control"           = "special"
+"keyword.directive"         = "special"
+"label"                     = "blue2"
+"module"                    = "type"
+"namespace"                 = "type"
+"operator"                  = "text"
+"punctuation"               = "text"
+"punctuation.delimiter"     = "text"
+"special"                   = "text"
+"string"                    = "orange"
+"string.regexp"             = "gold"
+"tag"                       = "blue2"
+"type"                      = "type"
+"type.builtin"              = "blue2"
+"type.enum.variant"         = "constant"
+"variable"                  = "variable"
+"variable.builtin"          = "blue2"
+"variable.other.member"     = "variable"
+"variable.parameter"        = "variable"
 
-"type" = { fg = "type" }
-"type.builtin" = { fg = "type" }
-"type.enum.variant" = { fg = "constant" }
-"constructor" = { fg = "type" }
-"variable.other.member" = { fg = "variable" }
-
-"keyword" = { fg = "blue2" }
-"keyword.directive" = { fg = "blue2" }
-"keyword.control" = { fg = "special" }
-"label" = { fg = "blue2" }
-"tag" = "blue2"
-
-"special" = { fg = "text" }
-"operator" = { fg = "text" }
-
-"punctuation" = { fg = "text" }
-"punctuation.delimiter" = { fg = "text" }
-
-"variable" = { fg = "variable" }
-"variable.parameter" = { fg = "variable" }
-"variable.builtin" = { fg = "blue2" }
-"constant" = { fg = "constant" }
-"constant.builtin" = { fg = "blue2" }
-
-"function" = { fg = "fn_declaration" }
-"function.builtin" = { fg = "fn_declaration" }
-"function.macro" = { fg = "blue2" }
-"attribute" = { fg = "fn_declaration" }
-
-"comment" = { fg = "dark_green" }
-
-"string" = { fg = "orange" }
-"constant.character" = { fg = "orange" }
-"string.regexp" = { fg = "gold" }
-"constant.numeric" = { fg = "pale_green" }
-"constant.character.escape" = { fg = "gold" }
-
-"markup.heading" = { fg = "blue2", modifiers = ["bold"] }
-"markup.list" = "blue3"
-"markup.bold" = { fg = "blue2", modifiers = ["bold"] }
-"markup.italic" = { modifiers = ["italic"] }
+# MARKUP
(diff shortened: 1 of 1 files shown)


kirawi:
If I'm reading the changes correctly, it looks like the colors are actually inaccurate to the original theme. All the colors were taken directly from the definition. I'm also against changing the statusline color since it the theme is supposed to be a faithful port of the original.

saccarosium:
Hi @kirawi,
You are right! Actually the diffs mismatch slide throw my radar. It is probably because I've crossed referenced also two vim colorschemes: [codedark.vim](https://github.com/tomasiser/vim-code-dark) and [vscode.nvim](https://github.com/Mofiqul/vscode.nvim). 

As for VSCode's source I checked this two: [dark+](https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_plus.json) and [vsdark](https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_vs.json)

>  I'm also against changing the statusline color since it the theme is supposed to be a faithful port of the original.

I think the statusline's color can be discussed, some people could prefer that.

kirawi:
Neither contain the full definition of the theme. What I had to do was open up the command palette and run `Generate color theme from current settings`.

runtime/themes/dark_plus.toml
```diff
@@ -1,138 +1,107 @@
...
+DiffGreen = '#4B5632'
+DiffLightBlue = '#87d7ff'
+DiffBlue = '#005f87'
+SearchCurrent = '#515c6a'
+Search = '#613315'
+PopupFront = '#BBBBBB'
+PopupBack  = '#2D2D30'
+Selection = '#264F78'
+Cursor = '#A6A6A6'
+WhiteSpace = '#4D4D4D'
```
kirawi, line 103: Should be `Whitespace` set to `3e3e3d`

kirawi:
Inlay hint colors were changed to the colors of Dark+ experimental in https://github.com/helix-editor/helix/pull/7611, but they are now defined as
```json
"editorInlayHint.background": "#6161611a",
"editorInlayHint.foreground": "#969696",
```

runtime/themes/dark_plus.toml
kirawi: Actually, I prefer to keep the names in terms of colors. That was originally my intention with this theme, but I never found the time to finish that. Trust me, its less of a headache trying to find a semantic name for things.

kirawi:
Actually, I see that they changed many colors in the theme. So the previous values will have to be rechecked anyway. The statusline change is okay then since they changed it.

kirawi:
I'll get back to you once I've checked everything.

saccarosium:
> What I had to do was open up the command palette and run Generate color theme from current settings

Ok, when I get home I will check it myself also.

For the change a think we stick to the original `dark+` theme instead of the new `dark modern`.

saccarosium:
I've came back to the original structure and only made some modifications.

kirawi:
Sorry, I didn't mean to indicate that all of your changes were unwanted. I liked how you previously formatted it, but it's fine if you don't want to do that work again. I agree that VSCode Dark Modern should be a separate theme and you are free to open up a new PR with the original changes for that. You would have more creative liberty that route since I only maintain Dark+ :P

saccarosium:
> Sorry, I didn't mean to indicate that all of your changes were unwanted.

No problem I did realize a lot of the colors I was used to were influenced by the vim colorschemes I previously used. So there was a lot of personal taste I decided to rollback to your version. After all you are the maintainer.

> I liked how you previously formatted it, but it's fine if you don't want to do that work again.

Do you mean separated the names with comments and aligned everything?

kirawi:
> So there was a lot of personal taste I decided to rollback to your version.

If it is close enough to the Vim version, then we could probably offer a Vim variant of Dark+. We already offer a Zed variant of onelight.

> Do you mean separated the names with comments and aligned everything?

Yeah, it looked nice.

saccarosium:
> Did the formatting. Yeah, it looked nice.

kirawi:
I think it would be acceptable, but I'll leave that to the discretion of the maintainers.

David-Else:
Please note line 10 and 11, usize has now gone blue, is this a mistake? Please don't make any changes to the inlay hint colours or statusbar, thanks.

## This PR
Screenshot from 2023-11-06 22-02-43
## How it was
Screenshot from 2023-11-06 21-58-50

saccarosium:
> Please note line 10 and 11, usize has now gone blue, is this a mistake?

In VSCode are green but it is inconsistent because it uses a regex highlight system. Since we are using TreeSitter, it does what we tell it to do, I've changed how it displays `built-in` types in blue and the rest of the types in green, witch does the right thing for C and C++.

kirawi:
So, I think reverting that change in highlighting would be the more correct choice. For some reason (maybe historical), primitive types are not treated like any other type. In VSCode, they're actually under `storage` rather than `type`. I think the more correct change would be to modify the tree-sitter queries to highlight them under that scope.

archseer:
> In VSCode, they're actually under `storage` rather than `type`. I think the more correct change would be to modify the tree-sitter queries to highlight them under an equivalent scope.

From what I remember I deliberately omitted `storage` as a toplevel member. There's `keyword.storage.type` for modifiers but built in types should just be `type.builtin`.

Highlighting doesn't need to be identical across editors, and given we're using a different engine (tree-sitter vs regex-based textmate grammar) it's unlikely we can get them 1:1 anyway.

saccarosium:
For recap. Do I revert the built-in types to green?

David-Else:
Is it possible to add the highlights for debugging that were introduced in https://github.com/helix-editor/helix/pull/5957? 
```
| Key                               | Notes                                                                                          |
| ---                               | ---                                                                                            |
| `ui.debug.breakpoint`             | Breakpoint indicator, found in the gutter                                                      |
| `ui.debug.active`                 | Indicator for the line at which debugging execution is paused at, found in the gutter          |
| `ui.highlight.frameline`          | Line at which debugging execution is paused at                                                 |

```

I am not sure exactly what they are, I looked at https://github.com/microsoft/vscode/blob/main/extensions/theme-defaults/themes/dark_plus.json. 

image

kirawi:
Hey, sorry for the delay. I'll be looking at this within a few days.

kirawi:
@David-Else I haven't used dark plus in a year or so. I'll be removing myself from maintainer of the theme after this PR. I think that your suggestions thus far have been consistent with VSCode, so if you're interested in becoming a new maintainer then I'd be fine with that.

saccarosium:
At this point, I think we can close this. I don't really use helix anymore.

Issue #5675, opened 2023-01-25
Cursor changes color on whitespaces and indent guides


imsuck:
### Summary

Even if you set the cursor's foreground and the reverse modifier, it would still change its color if it was on a virtual whitespace or indent guide. Which is not expected since on normal text it stays white.

<details><summary>theme.toml to replicate</summary>

```toml
"ui.selection" = { bg = "gray" }

"ui.cursor" = { fg = "light-gray", modifiers = ["reversed"] }

"ui.virtual.indent-guide" = { fg = "blue" }
"ui.virtual.whitespace" = { fg = "blue" }
```

</details>

### Reproduction Steps

 

I tried this:
[![asciicast](https://asciinema.org/a/79vFx0OJmmdrp6qEDhalNGeKL.svg)](https://asciinema.org/a/79vFx0OJmmdrp6qEDhalNGeKL)

I expected this to happen:
Whitespaces and indent guides should respect the cursor's foreground, and the cursor should stay white.

Instead, this happened:
The cursor changes its color.

Note: I use gray `▏` indent guides. And my selection bg has the same color. Which makes it hard to tell which side is the selection anchor on is if it was on the `▏`. I used blue for the asciinema just to make it stand out.

### Helix log

Since this is a theme related problem logs wouldn't be useful. I still left it here just in case.

<details><summary>~/.cache/helix/helix.log</summary>

```
2023-01-24T12:29:59.292 helix_view::editor [ERROR] Failed to initialize the LSP for `source.toml` { cannot find binary path }
2023-01-24T12:40:28.317 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
2023-01-24T13:35:50.586 helix_view::editor [ERROR] editor error: 1 unsaved buffer(s) remaining: ["hyprland.conf"]
2023-01-24T20:15:58.854 helix_view::editor [ERROR] editor error: No definition found.
2023-01-24T20:19:57.833 helix_view::editor [ERROR] editor error: No definition found.
2023-01-25T05:57:24.824 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
2023-01-25T05:59:36.373 helix_view::editor [ERROR] Failed to initialize the LSP for `source.bash` { cannot find binary path }
2023-01-25T06:46:01.835 helix_view::editor [ERROR] editor error: No more matches
[...]

PR #8733, opened 2023-11-06
Fix precedence of ui.virtual.whitespace

chtenb:
This fixes the first half of https://github.com/helix-editor/helix/issues/5675, namely the whitespace case.
The idea is to move the whitespace highlighting logic to the same place as the other highlighting logic, such that it's precedence can be precisely controlled.

### DIFF
--- a/helix-term/src/ui/document.rs
+++ b/helix-term/src/ui/document.rs
@@ -307,7 +307,7 @@ pub fn render_text<'t>(
 pub struct TextRenderer<'a> {
     pub surface: &'a mut Surface,
     pub text_style: Style,
-    pub whitespace_style: Style,
+    pub whitespace_style: Style, // TODO: this prop can be removed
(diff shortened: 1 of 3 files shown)


pascalkuthe (changes requested):
I donüt think we should implement it this way. That kind of highlighting is fairly inefficient since it requires collection to a vec (its fine for something rare like the cursor but not something common like whitespace). Even worse it requires a secondary transversal of all visible text. The text transversal is already the rendering bottleneck so doing it twice is not a good idea. 

I am still not sure what the best way to handle this would be without causing too much overhead. One way could be to maintain two highlgt iterators. One for sytnax highlighting and one for overlays. The we could have the option to patch highlights before the syntax  highlts, after the syntax highlights but before the overlay (like whitespace and indent guides) and then finally after the overlays.

That also should be pretty easy to implement.

Issue #4106, opened 2022-10-05
Highlights wrong part in LSP signature help feature.


nyeong:
### Summary

<img width="658" alt="스크린샷 2022-10-05 17 02 45" src="https://user-images.githubusercontent.com/12120034/194010751-bdcea589-e253-4c6c-8465-47156d43bf80.png">

After 22.08, *LSP signature help* has come, which highlights the function parameters which is currently being entered.
*Signature help* highlights the words in the function name if there are the same words in the parameters.

### Reproduction Steps

 

It can be reproduced by defining a function named with the words in parameters.

https://user-images.githubusercontent.com/12120034/194015259-836612af-bfa8-49ac-8412-ec0a541daa7b.mov

### Helix log

There is no log about this.

### Platform

macOS 12.6 (Apple Silicon)

### Terminal Emulator

iTerm 3.4.16

### Helix Version

helix 22.08.1 (66276ce6)

nyeong:
https://github.com/helix-editor/helix/blob/4a3b776b78c68f21017b1900eafb7fefeadeefa8/helix-term/src/commands/lsp.rs#L891-L912

`signature_help_impl` function is related.

A-Walrus:
Played around with this in VS Code and a similar thing occurs, but only when the name of the argument is the entire name of the function (which may or may not be valid code depending on your language but whatever).
![image](https://user-images.githubusercontent.com/58790821/194269670-2d7c821a-4cbc-4722-8440-e29f875478cb.png)

I believe that the LSP specification is flawed. 
According to the spec [specification#parameterInformation](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#parameterInformation), language servers can pass either the string label of the parameter, or the range of characters for it. If they pass the label as a string the client is expected to find the substring within the full label for the signature help and highlight it. This is problematic as the editor could mistakenly find substrings within the name of the function, within another parameter which contains that parameter name, within keywords such as `fn` `def` `do` `func` ...

We could try to make better guesses as to where and how we search for the substring, but ultimately I think the specification needs to be changed so that passing the parameter label as a string is deprecated and instead all language servers pass the character range.

PR #8734, opened 2023-11-06
Fix the parameter highlighting in the signature help overlay

marc-thieme:
Bug occured when:
* LSP sends simple ParameterLabel (just the name of the parameter instead of its location)
* the highlight is for a parameter which is a substring of the function name

Explanation: The heuristic for determining the byte offset of the parameter just does a `find` on the signature string. However, when the parameter name is a substring of the function name, then this hightlights this part of the function name instead of the parameter.

This PR changes the find to an rfind. The bug will still come up if the parameter is a substring a following parameter. However, it at least takes the function name out of the equation as this would be mathed at the very end.

An alternative approach might be to make a regex match of some sort to make sure that the highlighted region is a complete word/identifier.

hx_bug_wrong_param_highlight

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -18,7 +18,7 @@ use tui::{
 use super::{align_view, push_jump, Align, Context, Editor, Open};
 
 use helix_core::{
-    path, syntax::LanguageServerFeature, text_annotations::InlineAnnotation, Selection,
+    path, regex, syntax::LanguageServerFeature, text_annotations::InlineAnnotation, Selection,
 };
 use helix_view::{
     document::{DocumentInlayHints, DocumentInlayHintsId, Mode},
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Using regex word boundaries is a clever idea. It would still be possible to highlight incorrectly no matter how we detect the variable name in the label though. See https://github.com/helix-editor/helix/issues/4106 and https://github.com/microsoft/language-server-protocol/issues/1559: ideally language servers should only ever send `lsp::ParameterLabel::LabelOffsets` because sending just the label makes it impossible for the client to highlight the correct parameter without having language-specific knowledge (and even then, it wouldn't work properly in some cases.)

For Python I would recommend opening up an issue upstream in the language server's repo for switching to label offsets.

marc-thieme:
I just did the word boundary implementation as well. No pressure to follow up on this though, I had in mind that it's probably not gonna be merged when the option is deprecated anyway. 
Nevertheless, as python does still implements it this way, maybe it is relevant afterall.

the-mikedavis:
I doubt the performance impact would be noticeable since the regex would usually be very simple and the search string would usually be very short. It's calculated every time we switch between parameters so it's not as performance-sensitive as if it were run per-render but I think we should just stick with the `find` since this codepath is essentially deprecated. I imagine most language servers that send just the label do it for simplicity. IMO the range should be required

pascalkuthe:
I am going to close this since this PR has gone stale (conflicts and unadressed review comments).

I also don't want to invest much effort into an essentially deprecated codepath/fundementally flawed part specification

PR #8751, opened 2023-11-07
Add tutor for match mode

jerabaul29:
Add a section in the tutor to introduce the match mode. This simply walks the user through the different options, through a few examples.


runtime/tutor
the-mikedavis, line 1: Each section has a constant width, it looks like this new section is too tall

runtime/tutor
```diff
@@ -1184,6 +1184,46 @@ multiple cursors, they won't be uncommented but commented again.
...
+Select the "surround me" words, and surround them with "" using
+ms"
+
+Play around matching the different levels of surrouding in the
+following lines to learn to work with a superposition of pairs.
+
+ --> "some stuff" (inside parenthesis), [inside brackets]
+ --> surround me with something!
+ --> and some more stuff (outside the [inside])
+ --> it is also possible to surround THIS word
```
the-mikedavis, line 1216: This seems a little too free-form / exploratory to me - sections usually have concrete steps to try, for example: https://github.com/helix-editor/helix/blob/7bc564d3dcdee69555578bbec75d08e6ce227a9e/runtime/tutor#L253-L260

It may make sense to split this chapter into multiple sections, introducing `mi` with some examples for selecting within `"`, `(`, `[`, then another section for `ms` and then `md` and `mr`.

runtime/tutor
```diff
@@ -1186,6 +1186,160 @@ multiple cursors, they won't be uncommented but commented again.
```
David-Else, line 1188: I think the first section should begin with `CHAPTER 12 MATCH MODE`. It seems to begin on 12.1, maybe this was an oversight?

runtime/tutor
```diff
@@ -1186,6 +1186,160 @@ multiple cursors, they won't be uncommented but commented again.
...
+From normal mode, type "m" to enter the match mode. This is
+convenient to work with bracket pairs and the likes, and with
+their content.
+
+There are several actions that can be performed in match mode,
+as indicated by the help pop up. To jump to a matching ([{ pair,
+simply press "mm". For example on the lines below (starting with
+-->), move the cursor in normal mode to (, and press "mm" to jump
+to the matching ). You can do the same on the line below: for example
+move to ], and press "mm" to jump to [ .
```
David-Else, line 1202: Should we be talking about `matching bracket pair` instead of `matching ([{ pair`? Helix could have other settings for the actual brackets it is counting as a matching pair?

```suggestion
There are several actions that can be performed in match mode,
as indicated by the help pop-up. To jump to a matching ([{ pair,
simply press mm. For example on the lines below (starting with
-->), move the cursor in normal mode to (, and then press mm to jump
to the matching ). You can do the same on the line below: for example
move to ], and press mm to jump to [ .
```

runtime/tutor
```diff
@@ -1186,6 +1186,160 @@ multiple cursors, they won't be uncommented but commented again.
...
+
+--> some (nested surroundings [can be replaced])
+--> this "works with 'other surroundings' too"
+
+You can try to replace a non existing pair: this will show
+an error warning at the bottom bar and do nothing.
+
+
+=================================================================
+=                   12 MATCH MODE SUMMARY                       =
```
David-Else, line 1322: The other chapters end in this format using the phrase 'recap'.
```suggestion
=                     CHAPTER 12 RECAP                       =
```

runtime/tutor
```diff
@@ -1186,6 +1186,160 @@ multiple cursors, they won't be uncommented but commented again.
 
 
 
+=================================================================
+=                   12.1 USING MATCH MODE JUMP                  =
+=================================================================
+
+From normal mode, type "m" to enter the match mode. This is
+convenient to work with bracket pairs and the likes, and with
+their content.
```
David-Else, line 1195: I am not a fan of the phrase _and the likes_, it pops up quite a bit but it doesn't really make sense to me, and maybe non native speakers will get confused by it.

Looking at the rest of the tutor it seems the done thing is to use the name of the key alone rather than in quotes, so `m` rather than `"m"`. I think it would be a good idea to change it everywhere in the PR.

```suggestion
To switch to match mode from normal mode, type m. This feature is particularly useful for handling bracket pairs and their contents.
```

David-Else (commented):
Thanks for doing this, great work! I have only had time for a very quick look over. Let me know what you think of these suggestions. Cheers.

the-mikedavis (approved):
Thank you!

Issue #8757, opened 2023-11-08
Increase max tab-width


chtenb:
Tab-widths larger than 16 are incredibly useful for viewing TSV files, something that I've been doing for years using Vim.
Unfortunately Helix forbids any tab-width larger than 16, but I presume this would be fixed quite easily.

PR #8758, opened 2023-11-08
Extend max tab-width to 64

chtenb:
Fixes https://github.com/helix-editor/helix/issues/8757.
I chose 64 somewhat arbitrarily, but I think it is large enough we won't have complaints in the forseeable future.


helix-core/src/indent.rs
```diff
@@ -20,8 +20,9 @@ pub enum IndentStyle {
 }
 
 // 16 spaces
-const INDENTS: &str = "                ";
-const MAX_INDENT: u8 = 16;
+// TODO: up this to 64 or something
```
the-mikedavis, line 23:
```suggestion
// 64 spaces
```

the-mikedavis (approved):
Extending this up to 64 seems ok to me. Hopefully 64 will be enough :P

helix-core/src/indent.rs
```diff
@@ -20,8 +20,9 @@ pub enum IndentStyle {
 }
 
 // 16 spaces
```
woojiq, line 22:
```suggestion
```
Probably you need to delete this mention.

the-mikedavis (approved):

chtenb:
> Ref [#7178 (review)](https://github.com/helix-editor/helix/pull/7178#pullrequestreview-1455267920)

Any idea what problems @pascalkuthe was alluding to? I tried it out a with a TSV file with tab-width set to 64 and found no unexpected behavior thus far.

pascalkuthe:
the rendering code will infine loop if softwrab is enabled and a grapheme has a width larger than the viewport. Fundementally you just can't properly softrwap in that case I handled that by hardcoding a minimum width for which to enable softwrapping and a hard limit on the tabwidth. You can probably get more clever there, I just never bothered because I assumed that nobody really uses anything above 8

chtenb:
For TSV files softwrap does not make sense, and I had it turned off.  It sounds like there are two options, correct me if I'm wrong
1. Forbid softwrap and tabwidth>8
2. Increase the hardcoded minimum width to match the configured tabwidth

Would these solutions work, and which one would be preferred?

chtenb:
> the rendering code will infine loop if softwrab is enabled and a grapheme has a width larger than the viewport. Fundementally you just can't properly softrwap in that case I handled that by hardcoding a minimum width for which to enable softwrapping and a hard limit on the tabwidth. You can probably get more clever there, I just never bothered because I assumed that nobody really uses anything above 8

@pascalkuthe  I think I misunderstood this as in that soft-wrap would actually cause helix to hang currently. But since I cannot reproduce this, I think you meant that softwrapping is only accurately computed for graphemes smaller than 8 characters.

I think this is fine for now, and the problem does not become bigger with increasing the maximum tab-width, and it does make it possible to conveniently read and edit TSV files in Helix.

helix-core/src/syntax.rs
```diff
@@ -59,11 +59,11 @@ where
     D: serde::Deserializer<'de>,
 {
     usize::deserialize(deserializer).and_then(|n| {
-        if n > 0 && n <= 16 {
+        if n > 0 && n <= 64 {
```
TornaxO7, line 62:
```suggestion
        if n > 0 && n <= MAX_INDENT {
```

helix-core/src/syntax.rs
```diff
@@ -59,11 +59,11 @@ where
...
             Ok(n)
         } else {
             Err(serde::de::Error::custom(
-                "tab width must be a value from 1 to 16 inclusive",
+                "tab width must be a value from 1 to 64 inclusive",
```
TornaxO7, line 66:
```suggestion
                format!("tab width must be a value from 1 to {} inclusive", MAX_INDENT),
```

TornaxO7 (commented):
I'd prefer to use the constant `MAX_INDENT`, if we already have it to avoid hard-coded integers :D

TornaxO7 (approved):

chtenb:
@pascalkuthe @the-mikedavis Sorry to ping, but can we have another look this PR? It seems like a simple improvement to me, and I'm not clear on the reasons for letting it hang.

Regarding the worry with long graphemes and softwrap, there is no infinite loop, the long graphemes just do not show when the viewport falls below the grapheme width:

image
image

I think this is more than acceptable, especially compared to not being able to use long tabwidths at all.

PR #8760, opened 2023-11-08
Fix outdated diagnostics detection

dvic:
context: https://github.com/helix-editor/helix/discussions/8754

copying from discussion:

Question: it seems like (maybe only sometimes) diagnostics are not updating after fixing an issue in my Elixir projects. I'm seeing the following logs:

```
2023-11-08T09:41:48.358 helix_lsp::transport [INFO] elixir-ls <- {"jsonrpc":"2.0","method":"textDocument/publishDiagnostics","params":{"diagnostics":[],"uri":"file:///Users/dvic/Development/pluto/lib/sentry_filter.ex","version":5}}
2023-11-08T09:41:48.358 helix_term::application [ERROR] received malformed notification from Language Server: Unhandled
2023-11-08T09:41:48.358 helix_term::application [INFO] Version (5) is out of date for "/Users/dvic/Development/pluto/lib/sentry_filter.ex" (expected (4), dropping PublishDiagnostic notification
```

This PR fixes the issues, but I'm not 100% sure it's the right fix?

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -726,7 +726,7 @@ impl Application {
                         let offset_encoding = language_server.offset_encoding();
                         let doc = self.editor.document_by_path_mut(&path).filter(|doc| {
                             if let Some(version) = params.version {
-                                if version != doc.version() {
+                                if doc.version() > version {
                                     log::info!("Version ({version}) is out of date for {path:?} (expected ({}), dropping PublishDiagnostic notification", doc.version());
                                     return false;
                                 }


pascalkuthe:
this is not correct if document versions do not match then diagnostic should always be discarded as the content could be drastically different. If the versions send by the LSP do not match that means either:

* that the LSP is too slow (waiting a bit should fix this)
* that the LSP is not correctly following the LSP standard or has some other bug

The version checking here follows the LSP standard and is working as intended

dvic:
> this is not correct if document versions do not match then diagnostic should always be discarded as the content could be drastically different. If the versions send by the LSP do not match that means either:
> 
> * that the LSP is too slow (waiting a bit should fix this)
> * that the LSP is not correctly following the LSP standard or has some other bug
> 
> The version checking here follows the LSP standard and is working as intended

Thanks for the quick reply 👍!

I got this back from the LS author:

*ElixirLS pushes diagnostic with current document version and not the version from when the diagnostic was computed. This may not be optimal as the diagnostic may not be relevant to updated document version*
https://github.com/elixir-lsp/elixir-ls/issues/1024#issuecomment-1801910555

So if I understand correctly, then this is the reason for the mismatch, right?

and 

*Version documentation is rather scarce https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#publishDiagnosticsParams*

*... the spec does not mention anything about client rejecting diagnostics. I wonder if helix validation isn't too strict. Otherwise race conditions are inherent to diagnostic publish. Code parsing and compilation happens asynchronously and document updates can happen in between start of the parsing and document publish.*

Do you have any doc pointers for how the language server should behave?

cc @lukaszsamson

pascalkuthe:
the problem is quite simply that the diagnostics ranges may or may not be accurate anymore if thvresion number doesn't match. The point of the document version is to prevent the client from using outdated/incorrect data that is sent due to inherent race conditions of an async LSP server. This may lead to visual artifacts in the best case and crashes in the worst case or if these notifications are sent/processed out of order even an older diagnostic notification overwriting a new one.

The expectation is that the server would resend the diagnostics on documentDidChange anyway. While we can map diagnostic positions on the client side most editors (like vscode) don't do that so all diagnostics need to be republished whenever the document changes anyway. if we discard a request with an old version number that just means that there will be a notification with new data soon as soon as server is done processing.

The change here essentially makes that check entirely useless. It's impossible for the server to send a newer version number since the version number is only ever increased by the client (even on undo).

dvic:
Thank you for the lengthy response! Great explanation, I think I understand now. We'll investigate further on ElixirLS side of things.

PR #8795, opened 2023-11-12
Add smooth jumping between Helix and Tmux panes.

fiodorbaczynski:
Hi,

I've recently switched over from Neovim, and I love Helix. However, one crucial thing I'm missing is smooth navigation between editor panes and Tmux panes.

I created a PoC adding this functionality to Helix. I based the implementation on [nvim-tmux-navigator](https://github.com/alexghergh/nvim-tmux-navigation/blob/543f090a45cef28156162883d2412fffecb6b750/lua/nvim-tmux-navigation/tmux_util.lua#L10).

It would probably be better to have it as a plugin, but the plugin system is not available yet, and I believe using TUI editors with Tmux is a very common case.

This PR adds an editor option `jump_tmux_panes: bool`. If enabled, commands `jump_view_(right|left|up|down)` will send the appropriate command to Tmux whenever there is no Helix pane to jump to, i.e. the focus is already at the edge of the editor.

Please let me know if this is something you'd be willing to add to Helix (in this or another form).

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4858,19 +4858,46 @@ fn rotate_view_reverse(cx: &mut Context) {
 }
 
 fn jump_view_right(cx: &mut Context) {
-    cx.editor.focus_direction(tree::Direction::Right)
+    jump_view(cx, tree::Direction::Right);
 }
 
 fn jump_view_left(cx: &mut Context) {
-    cx.editor.focus_direction(tree::Direction::Left)
+    jump_view(cx, tree::Direction::Left);
 }
 
 fn jump_view_up(cx: &mut Context) {
-    cx.editor.focus_direction(tree::Direction::Up)
+    jump_view(cx, tree::Direction::Up);
 }
 
 fn jump_view_down(cx: &mut Context) {
-    cx.editor.focus_direction(tree::Direction::Down)
+    jump_view(cx, tree::Direction::Down);
+}
+
+fn jump_view(cx: &mut Context, direction: tree::Direction) {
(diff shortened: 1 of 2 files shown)


fiodorbaczynski:
To make practical use of this feature you need to set the same keybindings for pane navigation in Helix and Tmux.

`helix/config.toml`

```
[keys.normal]
"C-h" = ["jump_view_left"]
"C-j" = ["jump_view_down"]
"C-k" = ["jump_view_up"]
"C-l" = ["jump_view_right"]
```

`tmux/tmux.conf`

```
is_vim_or_helix="ps -o state= -o comm= -t '#{pane_tty}' | grep -iqE '^[^TXZ ]+ +(\\S+\\/)?(g?\.?(view|n?vim?x?)|hx(-wrapped)?(diff)?)$'"

bind-key -n 'C-h' if-shell "$is_vim_or_helix" 'send-keys C-h' 'select-pane -L'
bind-key -n 'C-j' if-shell "$is_vim_or_helix" 'send-keys C-j' 'select-pane -D'
bind-key -n 'C-k' if-shell "$is_vim_or_helix" 'send-keys C-k' 'select-pane -U'
bind-key -n 'C-l' if-shell "$is_vim_or_helix" 'send-keys C-l' 'select-pane -R'
```

pascalkuthe:
It's already possible to create such keybindings with `:sh`. Such specific integration with other tools is not suited to core generally

fiodorbaczynski:
Could you please point me in the right direction on how to create such keybindings? I was under the impression that you could only give a list of commands to a keybinding without arguments.

PR #8803, opened 2023-11-13
Default 'roots' field of language config

the-mikedavis:
Previously roots needed to be specified by every language and `[]` was used as an explicit default. Root files don't make sense for every language (for example TOML) so I think we should allow languages to not explicitly mention the key and have the `[]` default automatically.


pascalkuthe (approved):

archseer:
Should we also remove the empty key everywhere to shrink the file a bit?

pascalkuthe (approved):

PR #8865, opened 2023-11-19
feat(lsp): implement show document request

matoous:
Implement [window.showDocument](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#window_showDocument)
LSP server-sent request.

This PR builds on top of https://github.com/helix-editor/helix/pull/5820,
moves the external-URL opening functionality into shared crate-level
function that returns a callback that is now used by both the
`open_file` command as well as the window.showDocument handler if
the URL is marked as external.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -7,7 +7,7 @@ use helix_core::{
 };
 use helix_lsp::{
     lsp::{self, notification::Notification},
-    util::lsp_pos_to_pos,
+    util::{lsp_pos_to_pos, lsp_range_to_range},
     LspProgressMap,
 };
 use helix_view::{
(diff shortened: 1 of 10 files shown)


helix-term/src/lib.rs
```diff
@@ -47,3 +51,22 @@ fn filter_picker_entry(entry: &DirEntry, root: &Path, dedup_symlinks: bool) -> b
 
     true
 }
+
+/// Opens URL in external program.
+fn open_external_url_callback(
```
matoous, line 56: **question**: any ideas about better location for this?

helix-term/src/application.rs
```diff
@@ -1134,6 +1134,63 @@ impl Application {
...
+                            Ok(path) => path,
+                            Err(_) => {
+                                log::error!("Unsupported file URI: {}", params.uri);
+                                return;
+                            }
+                        };
+
+                        let action = match params.take_focus {
+                            Some(true) => helix_view::editor::Action::Replace,
+                            _ => helix_view::editor::Action::HorizontalSplit,
```
pascalkuthe, line 1157: hmm I am not sure a hsplit is always preferable that will depend on the exact screen layout and also personal preference.

I would expect a vsplit to be the more common choice (since it would allow side by side comparisons of files)?

helix-term/src/application.rs
```diff
@@ -1134,6 +1134,63 @@ impl Application {
...
+                            doc.set_selection(
+                                view.id,
+                                Selection::single(new_range.head, new_range.anchor),
+                            );
+                            if action.align_view(view, doc.id()) {
+                                align_view(doc, view, Align::Center);
+                            }
+                        }
+
+                        Ok(serde_json::Value::Null)
```
pascalkuthe, line 1192: you need to return an appropriate result here: https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#showDocumentResult

matoous: Added. It made the code way more verbose than I would like it to but I didn't find easy way to simplify it, happy to hear suggestions.

helix-term/src/application.rs
```diff
@@ -1134,6 +1134,84 @@ impl Application {
                         }
                         Ok(serde_json::Value::Null)
                     }
+                    Ok(MethodCall::ShowDocument(params)) => {
+                        let language_server = language_server!();
+                        let offset_encoding = language_server.offset_encoding();
+
+                        let success = match params {
```
pascalkuthe, line 1141: a nested match/if assignment like this is pretty verbose. I would just use `let mut success = false` and only set it to true if the command succeded (two places).

Also using a match is pretty verbose here. I would just use early return liberally  (`return Ok(json!(lsp::ShowDocumentResult { success: true/false }))` is simple enough to repeat a few times).

matoous: Ended up extracting the handling logic into separate method which ended up being way cleaner.

helix-term/src/lib.rs
pascalkuthe: This can be moved to helix-event once #8021 lands (will.need to be refactoring then anyway)

pascalkuthe (approved):

helix-term/src/application.rs
pascalkuthe: Yeah that is better

pascalkuthe (approved):

the-mikedavis (approved):

pascalkuthe (approved):

David-Else:
@matoous Thanks for your great work! Does this unlock any features in any particular language server? For example, the ability to jump to documentation when we get a lint error? Cheers!

matoous:
@David-Else I honestly haven't check the support from various LSPs.

Some references I found:

- `gopls`: https://github.com/golang/tools/commit/8f0778217741cbba088ee8426457809bfff3b3f5
- `typeprof` (ruby): https://github.com/ruby/typeprof/blob/90c1b94d0206abd66a9bace6222a0bc7cd24e0ef/lib/typeprof/lsp.rb#L540
- `closure.lsp`: https://github.com/clojure-lsp/clojure-lsp (used to open generated test file)

I don't think the adoption is wide.

tmistele:
@David-Else It unlocked this for me ([typst](https://github.com/typst/typst/) live preview with "inverse search")

[typst-lsp-test.webm](https://github.com/helix-editor/helix/assets/29761456/5c8509d2-349e-4cbf-82b4-8cf58dc03124)

which I find pretty cool, thanks @matoous for the implementation in helix!

What you're seeing here is [typst-lsp](https://github.com/nvarner/typst-lsp) with some [experimental patches for live preview](https://github.com/tmistele/typst-lsp) on top.

knuesel:
Thanks for the inspiration @tmistele, I've suggested enabling something similar in texlab and the devs have been super responsive (see texlab 5.16.0 released today). So Latex inverse search now works in Helix :)

PR #8873, opened 2023-11-20
Initialize diagnostics when opening a document

Philipp-M:
I've noticed, that diagnostics that are sent from the language server aren't propagated when opening a document (via `helix_view::Document::open`). So that the language server needs to send a `PublishDiagnostics` again to populate the document with diagnostics. This can take some time (noticed with rust-analyzer), when opening a file via the workspace diagnostic picker (if it is sent at all). 

This PR should ensure that diagnostics that are already in the editor instance are propagated when opening a document.
This has the additional advantage, that diagnostics can be rendered directly in the workspace diagnostic picker, when the file/buffer isn't open yet.

### DIFF
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -1,14 +1,8 @@
 use arc_swap::{access::Map, ArcSwap};
 use futures_util::Stream;
-use helix_core::{
-    chars::char_is_word,
-    diagnostic::{DiagnosticTag, NumberOrString},
-    path::get_relative_path,
-    pos_at_coords, syntax, Selection,
-};
+use helix_core::{path::get_relative_path, pos_at_coords, syntax, Selection};
 use helix_lsp::{
     lsp::{self, notification::Notification},
-    util::lsp_pos_to_pos,
     LspProgressMap,
 };
 use helix_view::{
@@ -731,7 +725,6 @@ impl Application {
                             log::error!("Discarding publishDiagnostic notification sent by an uninitialized server: {}", language_server.name());
                             return;
                         }
-                        let offset_encoding = language_server.offset_encoding();
                         // have to inline the function because of borrow checking...
                         let doc = self.editor.documents.values_mut()
                             .find(|doc| doc.path().map(|p| p == &path).unwrap_or(false))
@@ -745,11 +738,10 @@ impl Application {
                                 true
                             });
 
-                        if let Some(doc) = doc {
+                        let mut unchanged_diag_sources = Vec::new();
+                        if let Some(doc) = &doc {
                             let lang_conf = doc.language.clone();
-                            let text = doc.text().clone();
 
-                            let mut unchaged_diag_sources_ = Vec::new();
                             if let Some(lang_conf) = &lang_conf {
                                 if let Some(old_diagnostics) =
                                     self.editor.diagnostics.get(&params.uri)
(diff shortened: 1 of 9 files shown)


helix-view/src/document.rs
```diff
@@ -1692,15 +1711,97 @@ impl Document {
...
     #[inline]
     pub fn diagnostics(&self) -> &[Diagnostic] {
         &self.diagnostics
     }
 
     pub fn shown_diagnostics(&self) -> impl Iterator<Item = &Diagnostic> + DoubleEndedIterator {
         self.diagnostics.iter().filter(|d| {
-            self.language_servers_with_feature(LanguageServerFeature::Diagnostics)
-                .any(|ls| ls.id() == d.language_server_id)
+            self.language_servers.len() == 0
```
Philipp-M, line 1801: This is slightly hacky, there's likely a cleaner and more correct way to do this, but this is probably much simpler (I'm very open to better solutions though).

The issue is, that language servers are initialized and assigned somewhere else (in `helix_view::Editor::launch_language_servers`) after the Document is opened and only, when it is really opened. In the pickers, this isn't the case, but the language server is likely loaded...

pascalkuthe: hmm I am not sure how correct this is, if a language server doesn't have `LanguageServerFeature::Diagnostic` then we shounldn't show the diagnosts in the preview either I think? 

A less hacky (and generally faster) solution could be to remove the `shown_diagnostics` function and instead simply only copy the diagnostics from the editor for the language servers that have the appropiate feature.

That would make code less error prone in general (remiving the common pitfall of using `shown_diagnostics instead of normal diagnostics).

This would necessitate repopulating all document diagnostics on config reload but I think I would be ok with that. Right now it would be a bit messy (you would have to add it to refresh_config in application.rs) but with the event system we could use that meachanisim to move that behaviour to helix_view where it belongs

helix-view/src/document.rs
```diff
@@ -692,20 +693,20 @@ impl Document {
         path: &Path,
         encoding: Option<&'static Encoding>,
         config_loader: Option<Arc<syntax::Loader>>,
-        config: Arc<dyn DynAccess<Config>>,
```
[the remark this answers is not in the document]
Philipp-M: I'd agree, and I think I have thought about this, but the issue is, that `Diagnostic` needs the language server id (and offset_encoding).

So maybe instead keep using the (non-lsp) `Diagnostic`, do the conversion in `Editor` and make 
 `language_server_id` optional (as you described).
Which would probably require the current conversion stuff below as a method in `Editor` instead, something like `fn doc_diagnostics(&self, &Document) -> impl Iterator<Diagnostic>`.

helix-view/src/document.rs
Philipp-M: Yeah this is ugly and incorrect, I'm not happy about this either.

And I agree, but the issue is (why `shown_diagnostics` exists at all), that the command `code_action` relies on non-visible diagnostics (I had a case when I implemented that, not sure which language server it was, where filtered diagnostics would hide some code-actions because of this). This can probably be solved by using the editor diagnostics in `code_action` directly though (maybe via the `doc_diagnostics` as described above).

pascalkuthe: you will not be able to use the `doc_diagnostecs` function described above because diagnostics are mapped rough changes in the document while those on `Editor` are not. I think it could be ok if we just don't show codeactions that pertain only to diagnostics if diagnostics are disabled for a LS (I think that wouldn't be surprising UI wise that ifyoou disable diagnistoces that the associated codeactions also go away).

helix-view/src/document.rs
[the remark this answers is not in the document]
pascalkuthe: That could work but you will have to add a way to filter these diagnostics (before conversion) parameter. Publish diagnostic only wants to replace diagnostics from one language server (and in #6447 I also do some additional filtering). A closure `filter: impl Fn(&lsp::Diagnostic) -> bool` that you then pass to `Iterator::filter` would work well I think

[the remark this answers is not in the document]
Philipp-M: I've added the `doc_diagnostics` method. It got a little bit bigger and pyramidic than anticipated though... And I couldn't easily use it in `application.rs` due to borrow-checking issues, not sure whether it makes sense to work around them just for this function.

helix-view/src/document.rs
Philipp-M: Right... thanks for the hint. I guess then it's indeed better to just disable code-actions that are dependent on the diagnostics and remove the `shown_diagnostics`, I've done that in the last commit.

helix-view/src/document.rs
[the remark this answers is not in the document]
pascalkuthe: I actually added work around for the borrow checker issues in #6447. It's not too hard since everything relevant is an Arc so you can simply clone everything (mostly document text and config IIRC)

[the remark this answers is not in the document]
Philipp-M: Yes I know (as I tried that), it's dependent only on `text`, `path` and because we want to filter for enabled diagnostics `language_config`, but I think it's more convoluted that way and it should also be a little bit less efficient (not that that matters with these comparatively small `n`).

Philipp-M:
Since there were quite some merge conflicts to resolve, I tried again with reusing `Editor::doc_diagnostics` (by not acting directly on the editor reference, but instead the registry and diagnostics), so in that form it probably improves it.

Can you check again @pascalkuthe?

helix-term/src/application.rs
```diff
@@ -910,6 +795,24 @@ impl Application {
...
+                                    &self.editor.language_servers,
+                                    &self.editor.diagnostics,
+                                    doc,
+                                    |diagnostic, ls_id| {
+                                        ls_id == server_id
+                                            && diagnostic.source.as_ref().map_or(true, |source| {
+                                                !unchanged_diag_sources.contains(source)
+                                            })
+                                    },
+                                ),
```
pascalkuthe, line 811: nit: make a seperate variable for this. It's a bit unreadable with so much nesting

pascalkuthe (commented):
This looks pretty good to me now! Mostly just some style nits. I will test this locally for a few days.

One thing that is still missing is that you need to repopulate diagnostics on config reload in case the diagnostics feature was enabled/disabled for some LS.

helix-term/src/application.rs
Philipp-M: I've made it slightly less nested in the latest changes...

Philipp-M:
> One thing that is still missing is that you need to repopulate diagnostics on config reload in case the diagnostics feature was enabled/disabled for some LS.

I think I have covered this now everywhere where the language configuration can be changed, let me know if I missed something (or was to thorough).

helix-view/src/document.rs
```diff
@@ -1075,14 +1077,6 @@ impl Document {
...
     }
 
-    /// Set the programming language for the file if you know the name (scope) but don't have the
-    /// [`syntax::LanguageConfiguration`] for it.
-    pub fn set_language2(&mut self, scope: &str, config_loader: Arc<syntax::Loader>) {
-        let language_config = config_loader.language_config_for_scope(scope);
-
-        self.set_language(language_config, Some(config_loader));
-    }
-
```
Philipp-M, line 1085: I removed this, since I haven't found any occurrence anywhere in the codebase

helix-term/src/commands/typed.rs
```diff
@@ -2008,6 +2008,10 @@ fn language(
 
     let id = doc.id();
     cx.editor.refresh_language_servers(id);
+    let doc = doc_mut!(cx.editor);
+    let diagnostics =
+        Editor::doc_diagnostics(&cx.editor.language_servers, &cx.editor.diagnostics, doc);
+    doc.replace_diagnostics(diagnostics, &[], None);
```
pascalkuthe, line 2014: I am afraid this is not sufficient since any document in a split (or in general any buffer open in the background) also need to be updated. It probably makes more sense to create an editor function that handles this for all documents (similar to `refresh_language_server`).

Philipp-M: Oh, ok, I see, so there can be multiple documents with different ids but the same path? Should such function have the signature `Editor::reinitialize_diagnostics(&mut self, document_path: &Path)` or something like that?

Isn't the rest of the code above this incorrect as well (e.g. `doc.set_language()`)?

pascalkuthe: ah sorry I was tired and missrethis is for language and not general config reload in that case only refreshing a single document is fine. never mind

pascalkuthe (approved):

helix-term/src/commands/typed.rs
Philipp-M: Ok, that doesn't seem to be the case. So maybe I don't get what you mean. Do you mean that diagnostics need to be resent from the language server?

Philipp-M: Ah ok, had this still open, thanks for clearing it up :)

Issue #5672, opened 2023-01-24
Search from each selection


the-mikedavis:
The current search (`/`, `n` and `N`) only searches from the primary cursor. In select mode (`v/`) this can be used to add matches of the search pattern as new cursors.

It would also be useful to be able to search from every cursor. Instead of adding new selections for the next match from the primary cursor, this search would replace or extend the current selection(s) (depending on normal or select mode).

This behavior would be the same as textobjects like `]p` or `]f`. This could be used together with #1383 to implement regex-based textobjects like searching for numbers, dates, version control markers (https://github.com/helix-editor/helix/issues/4647), etc.. And I think it could be useful on its own in some multi-cursor scenarios and some single-cursor scenarios like #5668.

There might be some overlap with this and https://github.com/helix-editor/helix/pull/4635 with how search matches would be computed.

fennewald:
Ran into a case were running a macro with multiple cursors will cause the macro to behave differently for primary vs the others. Would definitely appreciate this!

xxxserxxx:
I want to comment here a work-around, for anyone (e.g., future me) searching for how to accomplish this (until such time as something better is implemented):
```
v<Alt-/>[search term]<Alt-->
```
and then trim the selection as necessary.

sibouras:
you could `/ccc<Enter>C/fff` then `<alt>-`

the-mikedavis:
`A-/` has never been bound in the keymap - you can search for it with `git log -S 'A-/'` and `git log -S "alt!('/"`.

For that use-case we could respect the mode in `/`/`?` so that it behaves like `n`/`N` in select mode. Currently it's hard-coded to move the primary cursor rather than add a cursor like you would with `n`/`N` in select mode. (That's why @sibouras's workaround works - searching in normal mode moves the primary cursor and leaves the rest unchanged.) Then you could merge selections with `A-minus` to select the whole range.

rcorre:
> but how do you get to fff extending your current selection?

`vgw` works, using the new `extend_to_word` (not saying that fixes the issue, but just another workaround to be aware of).

xxxserxxx:
> `A-/` has never been bound in the keymap - you can search for it with `git log -S 'A-/'` and `git log -S "alt!('/"`.

Yeah, I don't remember what I meant to write; I had A-/ bound to something, or I just typed it in from a faulty memory. The entire work-around was so klugy I rarely used it. @sibouras' solution does the same thing and is cleaner.

[...]

PR #8875, opened 2023-11-21
Add a simple amp-like jump command

### LINKED ISSUE #274: Better `f` à la clever-f and vim-sneak like jump motion

### LINKED ISSUE #510: AMP like Jump Mode

pascalkuthe:
Closes #274
Closes #510
Closes #5340

This is a simple amp-like jumpmode that @the-mikedavis and I have been toying around with for a while. When discussing/starting to review #5340 we found the PR more complex than we would like for this feature (a diff above 1k loc). We were uncomfortable with how closely it follows hop and the subsequent debate about emulating the million other jumpmode plugins out there. There are many different approaches to jump commands, and we cannot maintain them all in core. Instead, we want to pick a single opinionated choice that we feel fits well with the rest of the editor.

While testing that PR we made the following observations:

* character based jumps (which require a character before tying a label) don't really hold their water, you can use `f` for simple cases and `/` (which we want to improve further) for complex cases. Compared to a jumpmode these commands work with multicursors.
* word base jumps can be useful for quickly repositioning the primary cursor.
* a simple implementation like in AMP without dimming, color changes and complex letter assignment based on distance, ... was not only a lot simpler to implement but also more predictable/less disruptive

Based on these observations we decided to mostly follow amps implementation. This PR also has a more correct implementation fixing issues I found with the original PR (labels would appear on single char words, punction and would include leading whitespace, ..).

This PR contains a cherry pick from #6417, depending on which PR is merged first I will rebase the other PR and drop that commit

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -224,6 +224,7 @@ Jumps to various locations.
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
 | `j`   | Move down textual (instead of visual) line       | `move_line_down`           |
 | `k`   | Move up textual (instead of visual) line         | `move_line_up`             |
+| `w`   | Show labels at each word and select the word that belongs to the entered labels | `goto_word` |
 
 #### Match mode
 
--- a/helix-core/src/graphemes.rs
+++ b/helix-core/src/graphemes.rs
@@ -425,6 +425,85 @@ impl<'a> Iterator for RopeGraphemes<'a> {
     }
 }
 
+/// An iterator over the graphemes of a `RopeSlice` in reverse.
+#[derive(Clone)]
+pub struct RevRopeGraphemes<'a> {
+    text: RopeSlice<'a>,
+    chunks: Chunks<'a>,
+    cur_chunk: &'a str,
+    cur_chunk_start: usize,
+    cursor: GraphemeCursor,
+}
+
+impl<'a> fmt::Debug for RevRopeGraphemes<'a> {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("RevRopeGraphemes")
+            .field("text", &self.text)
+            .field("chunks", &self.chunks)
+            .field("cur_chunk", &self.cur_chunk)
+            .field("cur_chunk_start", &self.cur_chunk_start)
+            // .field("cursor", &self.cursor)
+            .finish()
+    }
+}
+
+impl<'a> RevRopeGraphemes<'a> {
+    #[must_use]
+    pub fn new(slice: RopeSlice) -> RevRopeGraphemes {
+        let (mut chunks, mut cur_chunk_start, _, _) = slice.chunks_at_byte(slice.len_bytes());
+        chunks.reverse();
+        let first_chunk = chunks.next().unwrap_or("");
+        cur_chunk_start -= first_chunk.len();
+        RevRopeGraphemes {
+            text: slice,
+            chunks,
+            cur_chunk: first_chunk,
+            cur_chunk_start,
+            cursor: GraphemeCursor::new(slice.len_bytes(), slice.len_bytes(), true),
+        }
+    }
+}
+
+impl<'a> Iterator for RevRopeGraphemes<'a> {
+    type Item = RopeSlice<'a>;
+
+    fn next(&mut self) -> Option<RopeSlice<'a>> {
+        let a = self.cursor.cur_cursor();
+        let b;
+        loop {
+            match self
+                .cursor
+                .prev_boundary(self.cur_chunk, self.cur_chunk_start)
+            {
+                Ok(None) => {
+                    return None;
+                }
+                Ok(Some(n)) => {
+                    b = n;
+                    break;
+                }
+                Err(GraphemeIncomplete::PrevChunk) => {
+                    self.cur_chunk = self.chunks.next().unwrap_or("");
+                    self.cur_chunk_start -= self.cur_chunk.len();
+                }
+                Err(GraphemeIncomplete::PreContext(idx)) => {
+                    let (chunk, byte_idx, _, _) = self.text.chunk_at_byte(idx.saturating_sub(1));
+                    self.cursor.provide_context(chunk, byte_idx);
+                }
+                _ => unreachable!(),
+            }
+        }
+
+        if a >= self.cur_chunk_start + self.cur_chunk.len() {
+            Some(self.text.byte_slice(b..a))
+        } else {
+            let a2 = a - self.cur_chunk_start;
+            let b2 = b - self.cur_chunk_start;
+            Some((&self.cur_chunk[b2..a2]).into())
+        }
+    }
+}
+
 /// A highly compressed Cow<'a, str> that holds
 /// atmost u31::MAX bytes and is readonly
 pub struct GraphemeStr<'a> {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -10,9 +10,12 @@ use tui::widgets::Row;
 pub use typed::*;
 
 use helix_core::{
-    char_idx_at_visual_offset, comment,
+    char_idx_at_visual_offset,
+    chars::char_is_word,
+    comment,
     doc_formatter::TextFormat,
-    encoding, find_workspace, graphemes,
+    encoding, find_workspace,
+    graphemes::{self, next_grapheme_boundary, RevRopeGraphemes},
     history::UndoKind,
     increment, indent,
     indent::IndentStyle,
@@ -24,7 +27,7 @@ use helix_core::{
     search::{self, CharMatcher},
     selection, shellwords, surround,
     syntax::{BlockCommentToken, LanguageServerFeature},
-    text_annotations::TextAnnotations,
+    text_annotations::{Overlay, TextAnnotations},
     textobject,
     tree_sitter::Node,
     unicode::width::UnicodeWidthChar,
@@ -503,6 +506,8 @@ impl MappableCommand {
         record_macro, "Record macro",
         replay_macro, "Replay macro",
         command_palette, "Open command palette",
+        goto_word, "Jump to a two-character label",
+        extend_to_word, "Extend to a two-character label",
     );
 }
 
@@ -5808,4 +5813,183 @@ fn replay_macro(cx: &mut Context) {
         // replaying recursively.
         cx.editor.macro_replaying.pop();
     }));
+}
+
+fn goto_word(cx: &mut Context) {
+    jump_to_word(cx, Movement::Move)
+}
+
+fn extend_to_word(cx: &mut Context) {
+    jump_to_word(cx, Movement::Extend)
+}
+
+fn jump_to_label(cx: &mut Context, labels: Vec<Range>, behaviour: Movement) {
+    let doc = doc!(cx.editor);
+    let alphabet = &cx.editor.config().jump_label_alphabet;
+    if labels.is_empty() {
+        return;
+    }
+    let alphabet_char = |i| {
+        let mut res = Tendril::new();
+        res.push(alphabet[i]);
+        res
+    };
+
+    // Add label for each jump candidate to the View as virtual text.
+    let text = doc.text().slice(..);
+    let mut overlays: Vec<_> = labels
+        .iter()
+        .enumerate()
+        .flat_map(|(i, range)| {
+            [
+                Overlay::new(range.from(), alphabet_char(i / alphabet.len())),
+                Overlay::new(
+                    graphemes::next_grapheme_boundary(text, range.from()),
+                    alphabet_char(i % alphabet.len()),
+                ),
+            ]
+        })
+        .collect();
+    overlays.sort_unstable_by_key(|overlay| overlay.char_idx);
+    let (view, doc) = current!(cx.editor);
+    doc.set_jump_labels(view.id, overlays);
+
+    // Accept two characters matching a visible label. Jump to the candidate
+    // for that label if it exists.
+    let primary_selection = doc.selection(view.id).primary();
+    let view = view.id;
+    let doc = doc.id();
+    cx.on_next_key(move |cx, event| {
+        let alphabet = &cx.editor.config().jump_label_alphabet;
+        let Some(i ) = event.char().and_then(|ch| alphabet.iter().position(|&it| it == ch)) else {
+            doc_mut!(cx.editor, &doc).remove_jump_labels(view);
+            return;
+        };
+        let outer = i * alphabet.len();
+        // Bail if the given character cannot be a jump label.
+        if outer > labels.len() {
+            doc_mut!(cx.editor, &doc).remove_jump_labels(view);
+            return;
+        }
+        cx.on_next_key(move |cx, event| {
+            doc_mut!(cx.editor, &doc).remove_jump_labels(view);
+            let alphabet = &cx.editor.config().jump_label_alphabet;
+            let Some(inner ) = event.char().and_then(|ch| alphabet.iter().position(|&it| it == ch)) else {
+                return;
+            };
+            if let Some(mut range) = labels.get(outer + inner).copied() {
+                range = if behaviour == Movement::Extend {
+                    let anchor = if range.anchor < range.head {
+                        let from = primary_selection.from();
+                        if range.anchor < from {
+                            range.anchor
+                        } else {
+                            from
+                        }
+                    } else {
+                        let to = primary_selection.to();
+                        if range.anchor > to {
+                            range.anchor
+                        } else {
+                            to
+                        }
+                    };
+                     Range::new(anchor, range.head)
+                }else{
+                    range.with_direction(Direction::Forward)
+                };
+                doc_mut!(cx.editor, &doc).set_selection(view, range.into());
+            }
+        });
+    });
+}
+
+fn jump_to_word(cx: &mut Context, behaviour: Movement) {
+    // Calculate the jump candidates: ranges for any visible words with two or
+    // more characters.
+    let alphabet = &cx.editor.config().jump_label_alphabet;
+    let jump_label_limit = alphabet.len() * alphabet.len();
+    let mut words = Vec::with_capacity(jump_label_limit);
+    let (view, doc) = current_ref!(cx.editor);
+    let text = doc.text().slice(..);
+
+    // This is not necessarily exact if there is virtual text like soft wrap.
+    // It's ok though because the extra jump labels will not be rendered.
+    let start = text.line_to_char(text.char_to_line(view.offset.anchor));
+    let end = text.line_to_char(view.estimate_last_doc_line(doc) + 1);
+
+    let primary_selection = doc.selection(view.id).primary();
+    let cursor = primary_selection.cursor(text);
+    let mut cursor_fwd = Range::point(cursor);
+    let mut cursor_rev = Range::point(cursor);
+    if text.get_char(cursor).is_some_and(|c| !c.is_whitespace()) {
+        let cursor_word_end = movement::move_next_word_end(text, cursor_fwd, 1);
+        //  single grapheme words need a specical case
+        if cursor_word_end.anchor == cursor {
+            cursor_fwd = cursor_word_end;
+        }
+        let cursor_word_start = movement::move_prev_word_start(text, cursor_rev, 1);
+        if cursor_word_start.anchor == next_grapheme_boundary(text, cursor) {
+            cursor_rev = cursor_word_start;
+        }
+    }
+    'outer: loop {
+        let mut changed = false;
+        while cursor_fwd.head < end {
+            cursor_fwd = movement::move_next_word_end(text, cursor_fwd, 1);
+            // The cursor is on a word that is atleast two graphemes long and
+            // madeup of word characters. The latter condition is needed because
+            // move_next_word_end simply treats a sequence of characters from
+            // the same char class as a word so `=<` would also count as a word.
+            let add_label = RevRopeGraphemes::new(text.slice(..cursor_fwd.head))
+                .take(2)
+                .take_while(|g| g.chars().all(char_is_word))
+                .count()
+                == 2;
+            if !add_label {
+                continue;
+            }
+            changed = true;
+            // skip any leading whitespace
+            cursor_fwd.anchor += text
+                .chars_at(cursor_fwd.anchor)
+                .take_while(|&c| !char_is_word(c))
+                .count();
+            words.push(cursor_fwd);
+            if words.len() == jump_label_limit {
+                break 'outer;
+            }
+            break;
+        }
+        while cursor_rev.head > start {
+            cursor_rev = movement::move_prev_word_start(text, cursor_rev, 1);
+            // The cursor is on a word that is atleast two graphemes long and
+            // madeup of word characters. The latter condition is needed because
+            // move_prev_word_start simply treats a sequence of characters from
+            // the same char class as a word so `=<` would also count as a word.
+            let add_label = RopeGraphemes::new(text.slice(cursor_rev.head..))
+                .take(2)
+                .take_while(|g| g.chars().all(char_is_word))
+                .count()
+                == 2;
+            if !add_label {
+                continue;
+            }
+            changed = true;
+            cursor_rev.anchor -= text
+                .chars_at(cursor_rev.anchor)
+                .reversed()
+                .take_while(|&c| !char_is_word(c))
+                .count();
+            words.push(cursor_rev);
+            if words.len() == jump_label_limit {
+                break 'outer;
+            }
+            break;
+        }
+        if !changed {
+            break;
+        }
+    }
+    jump_to_label(cx, words, behaviour)
 }
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -618,6 +618,7 @@ fn move_impl(cx: &mut Context, move_fn: MoveFn, dir: Direction, behaviour: Movem
             &mut annotations,
         )
     });
+    drop(annotations);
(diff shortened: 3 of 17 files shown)


book/src/keymap.md
```diff
@@ -223,7 +223,7 @@ Jumps to various locations.
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
 | `j`   | Move down textual (instead of visual) line       | `move_line_down`           |
 | `k`   | Move up textual (instead of visual) line         | `move_line_up`             |
-
+| `w`   | Show labels at each word and select the word that belongs to the entered labels | `goto_word`
 #### Match mode
```
the-mikedavis, line 227:
```suggestion
| `w`   | Show labels at each word and select the word that belongs to the entered labels | `goto_word` |

#### Match mode
```

this probably isn't an issue while rendering the markdown but it looks like it's breaking github's markdown syntax highlighting a little

helix-term/src/commands.rs
```diff
@@ -5750,3 +5755,156 @@ fn replay_macro(cx: &mut Context) {
...
+
+fn goto_label(cx: &mut Context) {
+    jump_to_word(cx, false)
+}
+
+fn extend_to_label(cx: &mut Context) {
+    jump_to_word(cx, true)
+}
+
+fn jump_to(cx: &mut Context, labels: Vec<Range>, extend: bool) {
```
the-mikedavis, line 5767:
```suggestion
fn jump_to_label(cx: &mut Context, labels: Vec<Range>, extend: bool) {
```

what do you think? I'm thinking that just "jump_to" is too generic

helix-term/src/commands.rs
```diff
@@ -5750,3 +5755,156 @@ fn replay_macro(cx: &mut Context) {
...
+    });
+}
+
+fn jump_to_word(cx: &mut Context, extend: bool) {
+    // Calculate the jump candidates: ranges for any visible words with two or
+    // more characters.
+    let alphabet = &cx.editor.config().jump_label_alphabet;
+    let jump_label_limit = alphabet.len() * alphabet.len();
+    let mut words = Vec::with_capacity(jump_label_limit);
+    let (view, doc) = current!(cx.editor);
```
the-mikedavis, line 5857:
```suggestion
    let (view, doc) = current_ref!(cx.editor);
```

helix-term/src/commands.rs
```diff
@@ -5750,3 +5755,156 @@ fn replay_macro(cx: &mut Context) {
...
+fn goto_label(cx: &mut Context) {
+    jump_to_word(cx, false)
+}
+
+fn extend_to_label(cx: &mut Context) {
+    jump_to_word(cx, true)
+}
+
+fn jump_to(cx: &mut Context, labels: Vec<Range>, extend: bool) {
+    let (_, doc) = current_ref!(cx.editor);
```
the-mikedavis, line 5768:
```suggestion
    let doc = doc!(cx.editor);
```

helix-term/src/commands.rs
```diff
@@ -5750,3 +5755,156 @@ fn replay_macro(cx: &mut Context) {
         cx.editor.macro_replaying.pop();
     }));
 }
+
+fn goto_label(cx: &mut Context) {
+    jump_to_word(cx, false)
+}
+
+fn extend_to_label(cx: &mut Context) {
+    jump_to_word(cx, true)
```
the-mikedavis, line 5764: We could re-use the `Movement` enum here instead of the bool

helix-term/src/commands.rs
```diff
@@ -5750,3 +5755,156 @@ fn replay_macro(cx: &mut Context) {
...
+            }
+            changed = true;
+            if words.len() == jump_label_limit {
+                break;
+            }
+        }
+        if !changed {
+            break;
+        }
+    }
```
matoous, line 5908: **thought**: this could reduce the nesting and (subjectively) make it easier to read:
```suggestion
    let mut cursor_rev = cursor_fwd;

    // collect words after the cursor
    while cursor_fwd.head < end && words.len() < jump_label_limit {
        cursor_fwd = movement::move_next_word_end(text, cursor_fwd, 1);
        let range = movement::move_prev_word_start(text, cursor_fwd, 1);
        // The cursor is on a word longer than 2 characters.
        if RopeGraphemes::new(text.slice(range.head..))
            .take(2)
            .all(|g| g.chars().all(char_is_word))
        {
            words.push(range.flip());
        }
    }

    // collect words before the cursor
    while cursor_rev.head > start && words.len() < jump_label_limit {
        cursor_rev = movement::move_prev_word_start(text, cursor_rev, 1);
        // The cursor is on a word longer than 2 characters.
        if RopeGraphemes::new(text.slice(cursor_rev.head..))
            .take(2)
            .all(|g| g.chars().all(char_is_word))
        {
            words.push(movement::move_next_word_end(text, cursor_rev, 1).flip());
        }
    }
```
I hope I didn't miss something in regards to the logic.

pascalkuthe: that is not equivalent tough. I purposefully iterate forward and reverse in lockstep so that labels are assigned in order of the textual distance.

matoous: Ah, gotcha. I see now. Thank you.

helix-term/src/commands.rs
pascalkuthe: yeah that seems reasonable , originally `jump_to_word` was actualy named `jump_to_label` so I couldn't call it that originally but I renamed that function after so now that name makes sense

book/src/keymap.md
```diff
@@ -223,6 +223,7 @@ Jumps to various locations.
 | `.`   | Go to last modification in current file          | `goto_last_modification`   |
 | `j`   | Move down textual (instead of visual) line       | `move_line_down`           |
 | `k`   | Move up textual (instead of visual) line         | `move_line_up`             |
+| `w`   | Show labels at each word and select the word that belongs to the entered labels | `goto_word` |
```
JeftavanderHorst, line 226: The command is documented as `goto_word` but is actually called `goto_label`

pascalkuthe: its actually the documentation that is right here, I renamed the command in the codebase to match

danillos:
I was testing it and I didn't saw any issue so far on the current implementation. Nice work.

I have a question, would it be possible to work in all visible buffers? So we could go to a word in another split window?

David-Else:
I have been testing it and it works brilliantly, thanks!

Omnikar:
Should the default value for `jump-label-alphabet` be reordered to prefer QWERTY home-row characters first?

archseer (approved):
Looks fine to me but I'm not a jump mode user. I'd be interested in @semin-park's feedback as the original PR owner

David-Else:
> > Should the default value for `jump-label-alphabet` be reordered to prefer QWERTY home-row characters first?
> 
> We generally don't try to tailor our keymap to qwerty (many people use different keymaps). Using the alphabet is a pretty reasonal default IMO.
> 
> I have set the setting to my (colmak) homerow plus some keys that are easily reachabke. What exactly those are is highly individual.

Alphabet is certainly a good default, but having a config option to prefer the home row keys of QUERTY / Colmeak / Dvorak  would be next level awesome :)

> I have set the setting to my (colmak) homerow plus some keys

How is this possible? Cheers!

pascalkuthe:
just add this to your config (under editor, its also documented in this PR):

jump-label-alphabet = "eirsaotndhfuywgm(cplbjvkx)z_"

(that example is for my personal layout where brackets and underscore are on the base layer)

David-Else:
@pascalkuthe You are amazing, you thought of everything. Sorry I didn't see that in the PR and wasted your time.

Omnikar:
> > Should the default value for `jump-label-alphabet` be reordered to prefer QWERTY home-row characters first?
> 
> We generally don't try to tailor our keymap to qwerty (many people use different keymaps).

Doesn't having HJKL as default directional controls already do that?

Omnikar:
@qwerty01 I think the recommendation for that is to just use search (`/` and `?`)

qwerty01:
`/` and `f` would work well for spots that are easily searchable, but braces and other non-word characters are usually not very searchable since they're 1 character and would have lots of matches. With the current solution, it would probably be finding some other spot on the line and then doing some other key combinations to get to the spot you want to, but that kinda defeats the purpose of amp jumps, which is looking at the spot you want to go and typing the keys that pop up.

Omnikar:
Well you could also type characters before and/or after the brace you're trying to jump to when using `/`.

qwerty01:
Oh quick aside, found an issue when scrolling while jump mode is active, only the visible screen at the time of enabling jump mode gives jump options, so if you scroll while jump mode is active, there aren't more options to jump:

scrolling

the-mikedavis:
Yeah, I think we should clear the jump labels / exit jumping mode when the viewport changes (scrolling, resizing, etc.). It's only meant to jump to locations that are in-view.

> I have a question, would it be possible to work in all visible buffers? So we could go to a word in another split window?

Currently it's a regular motion so like `w`/`b`/`e` or text-objects it only works on the current view (window). IMO jumping across views is not necessary to optimize for - I think switching views and then jumping within the view would work well enough.

pascalkuthe:
> Yeah, I think we should clear the jump labels / exit jumping mode when the viewport changes (scrolling, resizing, etc.). It's only meant to jump to locations that are visible/in-view.
> 
> > I have a question, would it be possible to work in all visible buffers? So we could go to a word in another split window?
> 
> Currently it's a regular motion so like `w`/`b`/`e` or text-objects it only works on the current view (window). IMO jumping across views is not necessary to optimize for - I think switching views and then jumping within the view would work well enough.

yeah I fixed this by simply dismissing any pending (or pseudo pending) keys whenever a mouse button is pressed (also during bracketed paste). I think mouse buttons behaving the same as other keys in that regard just makes sense. I think this will fix quite a few similar bugs.

archseer:
> We generally don't try to tailor our keymap to qwerty (many people use different keymaps).

I don't have strong opinions here, but I wouldn't mind ordering the default to use the qwerty homerow. I also don't use qwerty but the majority of the users do and it would make the default config slightly more efficient for them (the rest of us will remap anyway).

helix-view/src/editor.rs
```diff
@@ -291,6 +299,9 @@ pub struct Config {
     pub insert_final_newline: bool,
     /// Enables smart tab
     pub smart_tab: Option<SmartTabConfig>,
+    /// labels characters used in jumpmode
+    #[serde(skip_serializing, deserialize_with = "string_as_chars")]
+    pub jump_label_alphabet: Vec<char>,
```
archseer, line 304: We probably want to validate that the chars are unique with no duplicates

postsolar:
I think a feature worth considering is jumping while adding the target word as a new selection, similarly to what `n` in select mode does.

the-mikedavis:
We might want to consider that as a larger effort of https://github.com/helix-editor/helix/issues/5672. It might be useful to add selections rather than extend in some other cases like textobjects and tree-sitter motions. We could have a separate mode for it so that select mode always extends and the new mode always adds.

In any case though that should be left for future work rather than added to this PR

brunoherrick:
This seems to implement a "hop.nvim" jump mechanism (2 random chars). Will this also include a "leap.nvim"/ "lightspeed.nvim" jump mechanism (2 chars + eventual 3rd)? Gauging from github and forums such as reddit, the "leap" strategy seems more popular and is implemented in the previous PR #5340 . If this is intended to be part of the core, I would suggest to also include the seemingly more popular approach.

the-mikedavis:
Could you elaborate on the behavior you're looking for? I'm not sure anything like that is implemented in #5340. The character based jumps like with `g/` from #5340 I think would be ok to add as a follow-up to this PR as long as it only operates on the current view and re-uses most of the code introduced here.

brunoherrick:
@the-mikedavis exactly, the behavior you referenced via `g/` from the previous PR is the one I tried to describe. Hopefully it is possible to reuse the code from this PR and implement both types of jumps, as these two seem to be the most popular approaches for blazing fast movement.

postsolar:
Re: how it acts with regards to selections. An interesting implementation has been recently tried out in https://github.com/phaazon/hop.kak. What I found really elegant and interesting is that it's extensible as in it only acts as a "selections filter" so by using selections registers the user is free to build their own command, be it one adding a new cursor, or extending current selection, or something else

the-mikedavis:
Ah right, I had forgotten that the number of characters to search is configurable in #5340. That reminds me why I'm not enthusiastic about adding character based jumping or jump features in general to core.

What I don't like about #5340 is the size of the feature: it's adds a lot of code (four new modules, state-machine-like code, cosmetics like dimming), a new configuration section under `editor.jump-mode`, and basically adds a new "mode" to the editor (where this PR uses established machinery: just `cx.on_next_key`). I'm hesitant to add jumping to core at all because there are so many methods of doing it: there are many existing plugins and each plugin has a opinions on how to jump. I'm imagining us ending up with a lot of code in core covering multiple methods of jumping and/or behavior configuration like number of characters because we're trying to satisfy all the ways people think a jump feature should work.

So if we add jumping at all I'd like to keep it very simple like this Amp-like jump and leave everything else to plugins.

helix-term/src/commands.rs
```diff
@@ -5836,3 +5841,151 @@ fn replay_macro(cx: &mut Context) {
...
+            return;
+        }
+        cx.on_next_key(move |cx, event| {
+            doc_mut!(cx.editor, &doc).remove_jump_labels(view);
+            let alphabet = &cx.editor.config().jump_label_alphabet;
+            let Some(inner ) = event.char().and_then(|ch| alphabet.iter().position(|&it| it == ch)) else {
+                return;
+            };
+            if let Some(mut range) = labels.get(outer + inner).copied() {
+                if behaviour == Movement::Extend {
```
the-mikedavis, line 5908: When jumping in normal mode, should we normalize the direction of the selection like we do for LSP jumping (like we discussed in https://github.com/helix-editor/helix/discussions/6500#discussioncomment-5492299)?

Currently if you're jumping ahead the direction is forward and behind is backwards which makes sense for extending but I think we could always use backwards for normal mode jumping.

pascalkuthe: yeah good point

helix-view/src/editor.rs
pascalkuthe: done, I didn't change the default because I felt there was no good default choice for qwerty its subjective defending on handsize and so forth.

yerlaser:
Tested it and it works beautifully.
Do you, guys, have a document/wiki that describes what a proper reviewer should do and how to become one?

the-mikedavis:
You don't need any permissions to review PRs, feel free to leave review comments on open PRs. We don't have an exact procedure for becoming a maintainer / gaining triage or write priviledges if that's what you mean. This PR isn't really waiting on review though: the changes look good but it'd be nice to get some input from @semin-park (see https://github.com/helix-editor/helix/pull/8875#pullrequestreview-1743457571). He's been inactive for quite some time though so I think it would be ok to go ahead with this (\cc @archseer)

dvic:
For what it's worth: I've been using this for a few months and have not encountered any issues with it.

wontem:
One minor bug I encountered immediately: the last line in the buffer has an invalid number of anchors unless it's new-line-terminated:

rust, `Ok` is not reachable if there's no new line (not formatted correctly just for the demo purpose):





toml, `"all"` is not reachable if there's no new line:




```
[editor.whitespace]
render = "all"
```

pascalkuthe:
I can't reproduce that

wontem:
I apologize, my fault, pulled from the `helix:amp-jump` instead of `pascalkuthe:amp-jump`. It works like a charm, I will never leave this branch now 😄

wontem:
Now I'm on the correct version. I wondered what if I mistyped a command, so I typed `gw` and then `bw` (which is not present across the variants). I expected I stay in the same place, but I jumped into a random position. Then I repeated and I jumped somewhere else again. I checked with other absent combinations (`az, aw etc`) - the result is the same

the-mikedavis:
I can't reproduce that. If you hit a key that can't be a jump label then we bail out:

https://github.com/pascalkuthe/helix/blob/48d2638c031b3e17f60fbf82d1388584d43fc31d/helix-term/src/commands.rs#L5656-L5660

https://github.com/pascalkuthe/helix/blob/48d2638c031b3e17f60fbf82d1388584d43fc31d/helix-term/src/commands.rs#L5664-L5666

Can you make an https://asciinema.org/ recording / video / gif of what you see?

wontem:
I have no `asciinema` configured, going to do it a bit later. For now, hopefully, this video will suffice:

https://github.com/helix-editor/helix/assets/7056524/7220d08c-e9ea-4c84-8da5-9c557b289995

toml
```toml
theme = "custom"

[editor]
line-number = "relative"
mouse = false
bufferline = "always"
true-color = true
auto-format = true
auto-save = true
```

helix-term/src/commands.rs
```diff
@@ -5595,3 +5601,153 @@ fn replay_macro(cx: &mut Context) {
...
+    let mut cursor_rev = cursor_fwd;
+    loop {
+        let mut changed = false;
+        if cursor_fwd.head < end {
+            cursor_fwd = movement::move_next_word_end(text, cursor_fwd, 1);
+            let range = movement::move_prev_word_start(text, cursor_fwd, 1);
+            // The cursor is on a word longer than 2 characters.
+            if RopeGraphemes::new(text.slice(range.head..))
+                .take(2)
+                .all(|g| g.chars().all(char_is_word))
```
the-mikedavis, line 5725: Ah looks like we have some duplicates in the ranges:

<details><summary>Ranges from the reproduction case...</summary>

```rust
[
	Range { anchor: 19, head: 25, .. },
	Range { anchor: 15, head: 9, .. },
	Range { anchor: 19, head: 25, .. },
	Range { anchor: 27, head: 31, .. },
	Range { anchor: 27, head: 31, .. },
	Range { anchor: 5, head: 0, .. },
	Range { anchor: 32, head: 38, .. },
	Range { anchor: 32, head: 39, .. },
	Range { anchor: 42, head: 50, .. },
	Range { anchor: 42, head: 50, .. },
	Range { anchor: 52, head: 57, .. },
	Range { anchor: 52, head: 58, .. },
	Range { anchor: 60, head: 65, .. },
	Range { anchor: 66, head: 76, .. },
	Range { anchor: 66, head: 77, .. },
	Range { anchor: 80, head: 86, .. },
	Range { anchor: 80, head: 86, .. },
	Range { anchor: 88, head: 92, .. },
	Range { anchor: 88, head: 92, .. },
	Range { anchor: 93, head: 98, .. },
	Range { anchor: 93, head: 99, .. },
	Range { anchor: 101, head: 105, .. },
	Range { anchor: 106, head: 110, .. },
	Range { anchor: 106, head: 110, .. },
	Range { anchor: 111, head: 117, .. },
	Range { anchor: 111, head: 118, .. },
	Range { anchor: 120, head: 124, .. },
	Range { anchor: 125, head: 129, .. },
	Range { anchor: 125, head: 129, .. },
	Range { anchor: 130, head: 134, .. },
	Range { anchor: 130, head: 135, .. },
	Range { anchor: 137, head: 141, .. },
	Range { anchor: 137, head: 141, .. }
]
```

</details>

`gwaw` there is going to a valid label, it's just not visible since that one is overlapped by another label.

It looks like the `eb`-like movement trips over punctuation and can re-select the last word

wontem:
Another quick question, is it necessary to have an anchor in the current cursor position (if you're already in the beginning of a word)?

runtime/themes/everforest_dark.toml
James-Firth, line 1: I'm not sure what's common for the scope of theme changes like this but is it worth adding to more themes during this PR if it's going to be accepted?

Looking at Solarized Dark
image

and Dracula (my personal favourite)
image

The text doesn't end up changing. That said not all of them stay the same, github dark changes (though hard to read)
image
or Adwaita looks fine
image

James-Firth (commented):
Docs are clear and the feature works great! (Just did my first build to test this out!)

Just have one comment/question about themes

runtime/themes/everforest_dark.toml
david-crespo: While most people will probably want the color to change, it's still usable without that, and I think it's the most reasonable default when no color is configured because there's no way to generically choose the right color for all themes. Not sure why some would change without a color configured, though.

James-Firth: @david-crespo I just did some poking around, adding `"ui.virtual" = "colourname"` adds the functionality since the jump mode is using [virtual text, noted here](https://github.com/helix-editor/helix/pull/8875/files#diff-d782c07f22492bcae17333aba5dc5b840fb30317727e71267ebf4863f84787fdR5625).

Example adding this line to Solarized Dark and then activating jump mode
image
image

Seeing this, I actually no longer have any comment about the colours. My original point was wondering if more themes should be updated in advance for broader support, but really this just highlighted to me which themes don't have colour support for virtual text.

helix-term/src/commands.rs
pascalkuthe: It took some time to dig trough this but I fixed this and cleaned up some edgecases

runtime/themes/everforest_dark.toml
pascalkuthe: yeah we don't require new prs to edit themes. I changed the default theme other themes can be adopted in separate prs once this is merged

postsolar:
I noticed that, unlike `search_next` and similar commands, `goto_word` does not respect cursor orientation.

pascalkuthe:
It does in visual Mode just not on normal mode. It's intentional.

postsolar:
Got it, but what is the reasoning for this?

the-mikedavis:
See my comment above: https://github.com/helix-editor/helix/pull/8875#discussion_r1428152259

It matches the LSP location jumping features

helix-core/src/graphemes.rs
```diff
@@ -425,6 +425,85 @@ impl<'a> Iterator for RopeGraphemes<'a> {
     }
 }
 
+/// An iterator over the graphemes of a `RopeSlice`.
```
the-mikedavis, line 428:
```suggestion
/// An iterator over the graphemes of a `RopeSlice` in reverse.
```

helix-term/src/commands.rs
```diff
@@ -5755,3 +5762,182 @@ fn replay_macro(cx: &mut Context) {
...
+        }
+    }
+    'outer: loop {
+        let mut changed = false;
+        while cursor_fwd.head < end {
+            cursor_fwd = movement::move_next_word_end(text, cursor_fwd, 1);
+            // The cursor is on a word that is atleast two graphemes long and
+            // madeup of word characters. The latter condition is needed because
+            // move_next_word_end simply treats a sequence of characters from
+            // the same char class as a word so `=<` would also count as a word.            cursor_fwd = movement::move_next_word_end(text, cursor_fwd, 1);
```
the-mikedavis, line 5891: Looks like maybe a stray comment here

the-mikedavis (approved):

archseer (approved):

lucmos:
Thank you for this feature! It's working great :smile: 

Just a quick question -- Is it possible to "add" a selection to the word I am jumping to? Without extending the current selection as in visual mode, i.e., an easy way to add multiple cursors.

the-mikedavis:
There's some discussion about that above: https://github.com/helix-editor/helix/pull/8875#issuecomment-1845500723

pblkt:
It's likely that your theme doesn't define the [`ui.virtual.jump-label`](https://docs.helix-editor.com/themes.html#scopes:~:text=ui.virtual.jump,virtual%20jump%20labels) or the fallback `ui.virtual`. 

To verify that this is indeed the issue, check that the labels in `:theme default` are highlighted as expected .

To fix, make sure your theme defines it by either creating and using a local one or opening a PR.

giann:
That was it thank you

markstos:
> danillos I have a question, would it be possible to work in all visible buffers? So we could go to a word in another split window?

> @the-mikedavis  jumping across views is not necessary to optimize for - I think switching views and then jumping within the view would work well enough.

Allowing this to work across all visible windows could be an elegant way to directly switch to another window to reach ones that  can't be directly access by simply going "left/right/up/down" or to the last window.

wmfocus, sway-easyfocus, tmux, and some vim plugins all offer a way to switch to another window by selecting a letter. Here's the concept implemented in wmfocus (if the gif animates):

image

This allows navigating directly to any visible window-- helpful on 4k screens where 6 or Helix windows can fit in a terminal. 

By extending it to every visible window, the feature would be useful for both "amp-like jumping to a location in the current window" as well "switching to another window by label".

pickfire:
I noticed there is a bug where jump label switches positions with the text when cjk characters is involved, which causes layout shift, making it hard to know where to jump to.

I think it should preserves the existing positions if cjk characters is involved, otherwise the user need to read the whole thing to know where to jump to again.

For example

截屏2024-07-19 上午9 38 56

After `gw`, it becomes

截屏2024-07-19 上午9 39 13

In the second line, it should pad the first cjk character so it doesn't do layout shift either. The same goes for emoji and double width characters.

PR #8877, opened 2023-11-21
rust-toolchain.toml: bump MSRV to 1.70.0

cole-h:
With Firefox 120 released on 21 November 2023, the MSRV is now 1.70.0.

---

According to https://firefox-source-docs.mozilla.org/writing-rust-code/update-policy.html (and after attempting to download Firefox myself), Firefox 120 has been released. Since helix follows Firefox's MSRV (https://github.com/helix-editor/helix/blob/47b6c4bc78d449e3586f28d26507bfc8fb8608e2/docs/CONTRIBUTING.md#minimum-stable-rust-version-msrv-policy), I bumped the MSRV to 1.70.0 in line with Firefox!

(The main thing I'm looking forward to is the sparse crates registry support added in 1.68.0.)

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -49,7 +49,7 @@ use std::{
 /// If there is no configured language server that supports the feature, this displays a status message.
 /// Using this macro in a context where the editor automatically queries the LSP
 /// (instead of when the user explicitly does so via a keybind like `gd`)
-/// will spam the "No configured language server supports <feature>" status message confusingly.
+/// will spam the "No configured language server supports \<feature>" status message confusingly.
 #[macro_export]
 macro_rules! language_server_with_feature {
     ($editor:expr, $doc:expr, $feature:expr) => {{
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -39,7 +39,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@1.70
 
       - uses: Swatinem/rust-cache@v2
         with:
@@ -70,7 +70,7 @@ jobs:
         uses: actions/checkout@v4
 
       - name: Install stable toolchain
-        uses: dtolnay/rust-toolchain@1.65
+        uses: dtolnay/rust-toolchain@1.70
         with:
           components: rustfmt, clippy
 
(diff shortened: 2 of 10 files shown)


cole-h:
I also wonder if you'd be amenable to adding the `rust-analyzer` component to the toolchain components list... I've had to add it to the list locally in the past to get a rust-analyzer working, but maybe that's a problem with my system configuration.

pascalkuthe:
finally! We still want to wait till ubuntu updates but that is just a matter of time now.

I don't think we should include the rust-analyzer component, RA is backwards compatible (and using new version is preferable in every case, I always use nightly)

helix-term/src/commands/lsp.rs
```diff
@@ -49,7 +49,7 @@ use std::{
 /// If there is no configured language server that supports the feature, this displays a status message.
 /// Using this macro in a context where the editor automatically queries the LSP
 /// (instead of when the user explicitly does so via a keybind like `gd`)
-/// will spam the "No configured language server supports <feature>" status message confusingly.
+/// will spam the "No configured language server supports \<feature>" status message confusingly.
```
cole-h, line 52: I couldn't decide whether to backquote `<feature>` (so it appears as "code") or just escape the HTML tag rustdoc was complaining about. Happy to go either way.

the-mikedavis (commented):
For CI you'll want to update the old references to `dtolnay/rust-toolchain@1.65` in `.github/workflows/*` to use `1.70` instead

cole-h:
Good catch. I also notice that there's a `rust-version` in `helix-term/Cargo.toml` -- should I bump that to 1.70 as well? It's currently 1.65.

pascalkuthe:
Yes

archseer:
There's also a couple of TODOs in the code to simplify a couple things once MSRV is high enough, I think a couple can be tackled now

cole-h:
Is that something you'd like me to handle in this PR, or do you want that to be a separate PR?

pascalkuthe:
We usually handle that in the MSRV bump PR

cole-h:
OK, I think I got everything in my latest commits.

I didn't know what you would prefer for the checked math error handling in 44efc0333432e16957a7e2a1687c2de05099ec25, so I just used  `expect` with the same message of `Selection ranges cannot overlap` (going off the comment above that block). I'll gladly implement a better approach, or even feel free to push directly to my branch if you don't want to deal with the back-and-forth.

the-mikedavis:
Looks like ubuntu has bumped to 1.70 if I'm reading their packages correctly: https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=rust-all&searchon=names

helix-term/src/commands/lsp.rs
the-mikedavis: Escaping here seems proper to me 👍

the-mikedavis (approved):

pascalkuthe (approved):

PR #8879, opened 2023-11-22
Fix precedence of ui.virtual.whitespace 

chtenb:
This fixes the first half of https://github.com/helix-editor/helix/issues/5675, namely the whitespace case.
This is a patch up from https://github.com/helix-editor/helix/pull/8750, which was reverted. Sorry for that.

### DIFF
--- a/book/src/themes.md
+++ b/book/src/themes.md
@@ -300,7 +300,7 @@ These scopes are used for theming the editor interface:
 | `ui.popup.info`                   | Prompt for multiple key options                                                                |
 | `ui.window`                       | Borderlines separating splits                                                                  |
 | `ui.help`                         | Description box for commands                                                                   |
-| `ui.text`                         | Command prompts, popup text, etc.                                                              |
+| `ui.text`                         | Default text style, command prompts, popup text, etc.                                          |
 | `ui.text.focus`                   | The currently selected line in the picker                                                      |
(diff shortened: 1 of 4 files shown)


book/src/themes.md
```diff
@@ -300,7 +300,7 @@ These scopes are used for theming the editor interface:
 | `ui.popup.info`                   | Prompt for multiple key options                                                                |
 | `ui.window`                       | Borderlines separating splits                                                                  |
 | `ui.help`                         | Description box for commands                                                                   |
-| `ui.text`                         | Command prompts, popup text, etc.                                                              |
+| `ui.text`                         | Default text style, command prompts, popup text, etc.                                          |
```
atahrijouti, line 303: Is it the default text for any text type? even editor text? or just the UI?
if it's only the UI, I think it might be best to use `UI text style` instead of `Default ...`

chtenb: Also for the editor text

the-mikedavis (approved):

pascalkuthe (approved):

mandx:
I was expecting this PR to fix my issue with indent guides coloring, but looks like it doesn't? I've created #8909 to discuss more.

the-mikedavis:
From the description: "This fixes the first half of https://github.com/helix-editor/helix/issues/5675, namely the whitespace case."

So indent guides are left for future work. Now with #8909 we can close #5675 though.

PR #8895, opened 2023-11-23
added languages-open command

NoSpawnn:
added `languages-open` and `languages-open-workspace` commands to open the user and workspace languages.toml files respectively (both have aliases with `languages` -> `langs`)


pascalkuthe:
Considering that we are making progress towards integrating scheme/a scheme based config (which would nit have a separate language config file). I am not sure this is worth adding at this point

pascalkuthe:
I don't think we will be going forward with this, thank you for contributing!

Issue #1354, opened 2021-12-24
Highlighting for log files


kirawi:
[VSCode already does this natively,](https://github.com/microsoft/vscode/tree/94c9ea46838a9a619aeafb7e8afd1170c967bb55/extensions/log) and I find it's helpful when looking through huge logs.

Tudyx:
What would be the approach to tackle this? Write a grammar general enough that will work for most of the log files?

pascalkuthe:
You would need to write a Grammer. We don't support regen based highlighting (and don't really plan to in the future either).

Tudyx:
Ok I got something working locally, hopefully I will push an MR soon

![Screenshot from 2023-10-12 15-43-39](https://github.com/helix-editor/helix/assets/56633664/9cd8babd-8307-431c-9462-416cab24a67f)

PR #8916, opened 2023-11-26
add log tree-sitter

Tudyx:
Fix https://github.com/helix-editor/helix/issues/1354

Highlight examples:

dark_plus:
image
dracula:
image

### DIFF
--- a/runtime/queries/log/highlights.scm
+++ b/runtime/queries/log/highlights.scm
@@ -1,0 +1,10 @@
+(trace) @ui.virtual.inlay-hint
+(debug) @diff.plus
+(info) @info
+(warn) @warning
+(error) @error
+(year_month_day) @keyword
+(time) @constant
+(string_literal) @string
+(number) @constant.numeric
+(constant) @constant.builtin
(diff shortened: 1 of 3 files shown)


runtime/queries/log/highlights.scm
```diff
@@ -0,0 +1,10 @@
+(trace) @ui.virtual.inlay-hint
```
the-mikedavis, line 1: I think highlight queries should avoid using UI scopes. Could this just be `@comment` instead?

runtime/queries/log/highlights.scm
```diff
@@ -0,0 +1,10 @@
...
+(debug) @diff.plus
```
the-mikedavis, line 2:
```suggestion
(debug) @hint
```

to fit with the other info/warn/error highlights

Tudyx:
@the-mikedavis thanks for the suggestions, I updated the queries :wink:

the-mikedavis (approved):

pascalkuthe (approved):

PR #8924, opened 2023-11-27
LSP: Update doc paths renamed by TextDocumentEdit

the-mikedavis:
A language server might tell us to rename a file via a TextDocumentEdit when renaming a module for example. Previously if you had a buffer open for that file it would retain the old filename. With this change, the path for that document is updated if the file rename succeeds.

The actual change is quite small - we just need to do some shuffling of how the `&mut Editor` is passed around to satisfy the borrow checker.

Noticed this while looking into https://github.com/helix-editor/helix/discussions/8923

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -782,7 +782,10 @@ pub fn execute_lsp_command(editor: &mut Editor, language_server_id: usize, cmd: 
     });
 }
 
-pub fn apply_document_resource_op(op: &lsp::ResourceOp) -> std::io::Result<()> {
+pub fn apply_document_resource_op(
+    editor: &mut Editor,
+    op: &lsp::ResourceOp,
+) -> std::io::Result<()> {
     use lsp::ResourceOp;
     use std::fs;
     match op {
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I have the exact same patch lying around s locally (that I was still testing). I noticed some related RA bugs while working on this that I got fixed upstream first. 

We also need to trigger file watcher notifications here which I also added.

the-mikedavis:
Ah ok, I will wait for your patch then!

pascalkuthe:
I am still not quite sure if just renaming internally is enough. We may have to send a close and open notification for the old and new path too (that is why I hadn't submitted yet). I doesn't seem to work as well as I would like yet and I suspect its related to that

the-mikedavis:
I'm also curious what we should do about unsaved modifications. `Document::set_path` is nice for that since it will preserve everything but maybe if there are unsaved modifications we should just open the newly renamed file as a new buffer and keep the old doc around in scratch?

pascalkuthe:
I think setting the path is fine/exactly the right thing to do. Its pretty annoying renaming a module and having 20 buffers left open in incorrect paths. I am mostly unsure if we need to communicate `set_path` to the LSP somehow (which I think we do)

PR #8925, opened 2023-11-27
use workspace inheritance for common version

dead10ck:
By using the top-level `Cargo.toml` and workspace inheritance, we can eliminate all the duplication of the version number, and make `cargo build` report the correct version (rather than the old 0.6.0). This does require adding a patch version to make it semver compatible.


pascalkuthe (approved):
I would like to see the same thing being done for dependencies too (both so we get a better overview of all external dependencies and avoid duplicate versions)

helix-loader/build.rs
```diff
@@ -2,7 +2,7 @@ use std::borrow::Cow;
 use std::path::Path;
 use std::process::Command;
 
-const VERSION: &str = include_str!("../VERSION");
```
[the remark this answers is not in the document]
dead10ck: Yep, good idea. I updated the wording, and deleted the `VERSION` file.

the-mikedavis (approved):

docs/releases.md
```diff
@@ -1,16 +1,18 @@
 ## Checklist
 
 Helix releases are versioned in the Calendar Versioning scheme:
-`YY.0M(.MICRO)`, for example, `22.05` for May of 2022. In these instructions
+`YY.MM.MICRO`, for example, `22.05.0` for May of 2022. In these instructions
```
archseer, line 4: I personally don't like this notational change and would prefer we keep the original format.

Note that `MM` should also be changed back to `0M` regardless 

MM - Short month - 1, 2 ... 11, 12
0M - Zero-padded month - 01, 02 ... 11, 12

https://calver.org/#scheme

dead10ck: >Note that MM should also be changed back to 0M regardless

Ah sorry, I was not familiar with CalVer notation until now. I had assumed it was a typo.

>I personally don't like this notational change and would prefer we keep the original format.

So just to clarify, are you talking about just the 0M thing, or having a patch version at all? Because Cargo throws an error without one.

Cargo.toml
```diff
@@ -40,6 +40,11 @@ tree-sitter = { version = "0.20", git = "https://github.com/tree-sitter/tree-sit
 nucleo = "0.2.0"
 
 [workspace.package]
+version = "23.10.0"
```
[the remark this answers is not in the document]
dead10ck: Yes, it throws an error if the version is not semver. But it looks like having a patch version is a thing in CalVer, so it can be compatible with both.

dead10ck:
Ok this was changed to still use the `Cargo.toml`, but it drops the micro version if it's equal to 0. This will let us track the project version with Cargo, while the user facing version format remains unchanged.

docs/releases.md
```diff
@@ -1,16 +1,23 @@
 ## Checklist
 
 Helix releases are versioned in the Calendar Versioning scheme:
-`YY.0M(.MICRO)`, for example, `22.05` for May of 2022. In these instructions
-we'll use `<tag>` as a placeholder for the tag being published.
+`YY.0M(.MICRO)`, for example, `22.05` for May of 2022, or in a patch release,
```
pascalkuthe, line 4: Is this change intentional?

dead10ck: Yes. Is there an issue with it?

pascalkuthe (approved):

PR #8932, opened 2023-11-28
Update git source of Gemini grammar

nimr0d:
The author's username changed from "sfr" to "nbsp"

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -2916,7 +2916,7 @@ file-types = ["gmi"]
 
 [[grammar]]
 name = "gemini"
-source = { git = "https://git.sr.ht/~sfr/tree-sitter-gemini", rev = "3cc5e4bdf572d5df4277fc2e54d6299bd59a54b3" }
+source = { git = "https://git.sr.ht/~nbsp/tree-sitter-gemini", rev = "3cc5e4bdf572d5df4277fc2e54d6299bd59a54b3" }
 
 [[language]]
 name = "templ"


gabydd (approved):

archseer:
Fixed in master

mway:
Can a new release be cut to include this so that the latest release tag is buildable?

the-mikedavis:
We generally don't cut new releases for small bugs like this. You can work around it by adding:

```toml
use-grammars = { except = ["gemini"] }
```

to the top of your `~/.config/helix/languages.toml`.

Or you could download the source tarball `helix-23.10-source.tar.xz` from the releases page https://github.com/helix-editor/helix/releases/tag/23.10 which contains all of the grammars at the proper revisions.

the-mikedavis:
Github automatically attaches the source as it appears in-tree in the "Source files" artifacts on every release. No need to duplicate it

carlocaione:
Indeed but if you now take the "Source files" artifacts on for the `23.10` release and you try to compile those, you are 100% going to fail and not everybody wants (like macports) to get the pre-compiled version in the tarball.

So right now it is impossible to compile `helix` from scratch using any git tag (or tagged "source files") whatsoever.

vpayno:
It would be nice if a new tag was released since `23.10` isn't buildable without cherry-picking this fix. Or maybe a `23.10.01` hot-fix?

jordan-tuinman:
This is an annoying workaround - really feels like there should be a hotfix

PR #8934, opened 2023-11-28
fix(ui): use crossterm cursor when at the end of the rope

hnorkowski:
I found a missing special case for the out-of-focus crossterm cursor: When the block cursor is at the end of the rope it is not hidden.


the-mikedavis (approved):
Ah nice catch, I forgot about this edge-case

PR #8938, opened 2023-11-28
Make diagnostic picker also obey ignore files.

### LINKED ISSUE #8937: `.helix/ignore` should also mask out files from diagnostic pickers
Hello,

I'm using the `ignore` file feature for Godot/C# development and it's really handy!

But, when I go `space D` to search the workspace for errors, it also displays errors in autogenerated C# files in `.godot/...`, even though I have `.godot/**` in my ignore file.

I think it would make sense if this picker also obeyed the ignore file.

gamma-delta:
This does not completely obey all ignore files to the extent that the file picker does, because the `ignore` crate does a lot of background processing for walking *directory trees* with many ignore files that it doesn't expose for just querying arbitrary paths.

Fixes #8937.

My first PR to Helix! I use this editor every day so I'm really proud and happy I could contribute! `\[^_^]/`

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -266,11 +266,37 @@ fn diag_picker(
     _current_path: Option<lsp::Url>,
     format: DiagnosticsFormat,
 ) -> Picker<PickerDiagnostic> {
+    use ignore::gitignore::{Gitignore, GitignoreBuilder};
     // TODO: drop current_path comparison and instead use workspace: bool flag?
+
+    let ignore = {
+        let root = helix_core::find_workspace().0;
+        let absolute_root = root.canonicalize().unwrap_or_else(|_| root.clone());
+        let mut builder = GitignoreBuilder::new(absolute_root);
+
+        builder.add(helix_loader::config_dir().join("ignore"));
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I don't think  we should be filtering the diagnostics provided by the LSP in any way. If a file is not of interest the LSP should not be sending diagnostics for them. 

If an LSP is sending diagnostics you are not onterested in that is either issue you need to raise with the LSP or there may already be a way to ignore certain modules/files that you may not know about.

The specifics are language specific and I feel strongly that they should not be be handled on the editor side.  Using an ignore file is a hack that happens to work in this specific case and is undesirable in many situations. 

If you want to quickly filter a directory on the editor side you can also use negative matches in the picker.

gamma-delta:
My use case here is with Godot and C#. Godot autogenerates a lot of code
behind the scenes and stores it in the .godot folder... and those generate
diagnostics.

I'm not sure why it never occurred to me to just make OmniSharp ignore that
folder though. Thanks!

pascalkuthe:
I am closing this as per my comment above. I don't think it makes sense to add this. Sounds like handling this on the LS side was also the solution in your case.

PR #8941, opened 2023-11-29
Add instructions for building a specific version of helix

nbfritch:
This PR adds a small note in the book in the section on installing from source reminding users to checkout a version specific tag if they want to build a particular version from source.

I have tripped over this a few times and accidentally installed the master version.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -195,6 +195,15 @@ RUSTFLAGS="-C target-feature=-crt-static"
    cd helix
    ```
 
+   By default, git will checkout the master branch, which is the latest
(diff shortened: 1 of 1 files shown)


the-mikedavis:
Running master is the main draw of building from source though - there isn't much reason not to run a version from the releases page or a package manager if you want a tagged release

pascalkuthe:
Yeah I would generally want to steer people towards building master rather than a tag. If someone really wants to do that then it's pretty standard to checkout a tag. That's just how git works. I don't think we need to explain that

PR #8943, opened 2023-11-29
nix: update flake inputs

evanrichter:
also removed non-existent crane flake input overrides

This was just a simple `nix flake update`

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,44 +2,21 @@
   "nodes": {
     "crane": {
       "inputs": {
-        "flake-compat": "flake-compat",
-        "flake-utils": [
-          "flake-utils"
-        ],
         "nixpkgs": [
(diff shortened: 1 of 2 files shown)


the-mikedavis (approved):
Thanks!

pascalkuthe:
we ofcourse have a ci step that builds helix (including grammars) but that runs on the distros that GH runner provide so we can't catch "grammar won't work with distro/compiler configuration X" this way (and we can't change that without making CI really complicated an slow)

pascalkuthe:
Ah no I didn't mean nix specifically.  I meant we build helix in CI. The official build process for development uses cargo and that is what we test (on the GJ runner). To me nix is just another distro/package manger and shouldn't affect whether we merge PRs.

PR #8953, opened 2023-11-30
Do not disable GNU extensions when building grammars

fweimer-rh:
Commit ca65d312e715413cd5675cafd599d74534900d3f ("always build grammars with c++14 and c11 (#6792)") turned on semi-strict standard mode for C and C++ by accident.  Some grammars use non-standard functions such as isascii (from <ctype.h>) and fail to build as a result, particularly with future compilers which do not support implicit function declarations.

(Ideally, the build environment would only raise the standard versions relative to the toolchain default, not lower them, but this change does not implement that.)

Related to:

* https://fedoraproject.org/wiki/Changes/PortingToModernC
* https://fedoraproject.org/wiki/Toolchain/PortingToModernC

### DIFF
--- a/helix-loader/src/grammar.rs
+++ b/helix-loader/src/grammar.rs
@@ -509,7 +509,7 @@ fn build_tree_sitter_library(
 
         if let Some(scanner_path) = scanner_path.as_ref() {
             if scanner_path.extension() == Some("c".as_ref()) {
-                command.arg("-xc").arg("-std=c11").arg(scanner_path);
+                command.arg("-xc").arg("-std=gnu11").arg(scanner_path);
             } else {
                 let mut cpp_command = Command::new(compiler.path());
                 cpp_command.current_dir(src_path);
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I would rather see this fixed on the grammar side. Especially because we want grammars to build on windows too where gnu extensions are not available anyway. In this case there also seems to be a Microsoft specific isascii function but I am not actually sure it's equivalent.

Defining isascii is pretty trivial (it's even a macro on windows) so I would rather go that route

pascalkuthe:
sure but those are non-standard microsoft extensions. In this case its not a huge issue but there are many gnu extensions without equivalents on windows (and the reverse) or functions that work subtely differently. The function in question is even deprecated so its probably a good idea to move away from using it regardless.

Grammars should not be relying on non-standard C extensions that will just lead to portability issues.

pascalkuthe:
Closing as #9021 was merged which addresses the treesitter-d issue. Upstream made the same call as we did here: We should not be using POSIX/GNU extensions as that is not portable: 

https://github.com/gdamore/tree-sitter-d/pull/20#pullrequestreview-1758698117

PR #8957, opened 2023-11-30
highlights: rust: adjust enum variant recognition

rosefromthedead:
Now, enum variants with no arguments are recognised:
```
match x {
    Here => {},
}
ErrorKind::Here.clone();
module::Enum::AndHere.into();
```
and have been tested to not be falsely recognised:
```
Here.clone();
module::OrHere.into(); // probably unit structs, not variants
```
but still aren't recognised in the absence of a field access, method call or match arm because I think we'd need to put a `PascalCase::PascalCase` rule for that everywhere where expressions can appear (to rule out unit structs) which is likely possible but definitely annoying.


kirawi:
I don't feel like this is quite the right away to approach this because now it creates further ambiguity between enums and structs outside of pattern matching. I think the most correct way to address this is to have `locals.scm` track enums in a file (and potentially use stackgraphs in the future) and use semantic highlighting from the LSP, the latter of which has an implementation in an old PR: https://github.com/helix-editor/helix/pull/6102

I have been meaning to get to it, but life has been busy. But, it should fix your issues with the inconsistent highlighting once I or someone else continues that work.

runtime/queries/rust/highlights.scm
```diff
@@ -40,7 +40,7 @@
 ; ---
 
 (self) @variable.builtin
-(enum_variant (identifier) @type.enum.variant)
+(enum_variant (identifier) @constructor)
```
kirawi, line 43: Except for this change, I am fine with the rest of the PR as long as tracked to be at least partially reverted once semantic highlighting is implemented.

kirawi:
Are you still interested in this PR? The comment before my last one may have given the idea I was rejecting the change all together, which is not the case.

rosefromthedead:
Thanks for the poke, I had just forgotten :)

kirawi (approved):

the-mikedavis (approved):

pascalkuthe (approved):

PR #8976, opened 2023-12-03
upgrade tree-sitter-python

### LINKED ISSUE #8974: Broken syntax highlighting and auto-indentation when using python3.12 generics syntax
### Summary

image

Minor bug,

When using the `class Foo[T]:` syntax in python3.12, syntax highlighting is broken on the first method, and auto-indentation is broken on every method.

I'm guessing this has to do with grammars

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

Linux

### Terminal Emulator

kitty 0.31.0

### Installation Method

pacman

### Helix Version

helix 23.10

dead10ck:
supports new syntaxes from Python 3.12

Fixes #8974


archseer:
This should also pull in any highlights query changes upstream since the last bump (there's been a few since the original revision is from 2022)

Issue #8977, opened 2023-12-03
Joining lines with empty line adds 2 spaces when joined


erf:
### Summary

Joining two lines with an empty line between adds 2 spaces when joined. In `vim` only a single space is added. No space should be added on empty lines.

A weird features is that if i move the cursor to the middle empty line and press `J` then move the cursor to the first line and press `J` there is only a single space.

### Reproduction Steps

 

Given the following text:

```
abc

def
```

I tried this:

1. `hx`
2. `JJ`

I expected this to happen:
`abc def`

Instead, this happened (2 spaces between is not shown in markdown):
`abc  def`

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
please provide a copy of `~/.cache/helix/helix.log` here if possible, you may need to redact some of the lines
```

</details>

### Platform

macOS

### Terminal Emulator

Wezterm 20231122-174138-6a58a5ce

### Installation Method

brew

### Helix Version

helix 23.10 (f992c3b5)

PR #8989, opened 2023-12-04
Join empty lines with only one space in `join_selections`

TornaxO7:
fixes the issue that lines with only spaces are getting joined as well

fixes https://github.com/helix-editor/helix/issues/8977

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4412,7 +4412,8 @@ fn join_selections_impl(cx: &mut Context, select_space: bool) {
             // need to skip from start, not end
             let change = {
                 let separator = {
-                    let line_contains_only_space = LineEnding::from_char(text.char(end)).is_some();
+                    let line_end_index = line_end_char_index(&slice, line + 1);
+                    let line_contains_only_space = end == line_end_index;
                     if line_contains_only_space {
                         None
                     } else {
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -4390,24 +4390,24 @@ fn format_selections(cx: &mut Context) {
 fn join_selections_impl(cx: &mut Context, select_space: bool) {
     use movement::skip_while;
     let (view, doc) = current!(cx.editor);
+    let text = doc.text();
+    let slice = text.slice(..);
 
     let mut changes = Vec::new();
 
     for selection in doc.selection(view.id) {
-        let slice = doc.text().slice(..);
-        let text = doc.text();
-
-        let selected_lines = {
+        let lines = {
             let (start, mut end) = selection.line_range(slice);
             if start == end {
                 end = (end + 1).min(text.len_lines() - 1);
             }
             start..end
         };
+        changes.reserve(lines.len());
 
-        for selected_line in selected_lines {
-            let start = line_end_char_index(&slice, selected_line);
-            let mut end = text.line_to_char(selected_line + 1);
+        for line in lines {
+            let start = line_end_char_index(&slice, line);
+            let mut end = text.line_to_char(line + 1);
             end = skip_while(slice, end, |ch| matches!(ch, ' ' | '\t')).unwrap_or(end);
 
             // need to skip from start, not end
@@ -4446,9 +4446,9 @@ fn join_selections_impl(cx: &mut Context, select_space: bool) {
             })
             .collect();
         let selection = Selection::new(ranges, 0);
-        Transaction::change(doc.text(), changes.into_iter()).with_selection(selection)
+        Transaction::change(text, changes.into_iter()).with_selection(selection)
     } else {
-        Transaction::change(doc.text(), changes.into_iter())
+        Transaction::change(text, changes.into_iter())
     };
 
     doc.apply(&transaction, view.id);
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -4390,28 +4390,39 @@ fn format_selections(cx: &mut Context) {
...
+        let text = doc.text();
 
-        changes.reserve(lines.len());
+        let selected_lines = {
+            let (start, mut end) = selection.line_range(slice);
+            if start == end {
+                end = (end + 1).min(text.len_lines() - 1);
+            }
+            start..end
+        };
```
the-mikedavis, line 4407: Could you reset the unnecessary changes like variable renames and moving the `slice` and `text` bindings? Inlining `Tendril::from(" ")` is ok since that will avoid some allocations but the rest is just cosmetic

helix-term/src/commands.rs
```diff
@@ -4390,28 +4390,39 @@ fn format_selections(cx: &mut Context) {
...
+        for selected_line in selected_lines {
+            let start = line_end_char_index(&slice, selected_line);
+            let mut end = text.line_to_char(selected_line + 1);
             end = skip_while(slice, end, |ch| matches!(ch, ' ' | '\t')).unwrap_or(end);
 
             // need to skip from start, not end
-            let change = (start, end, Some(fragment.clone()));
+            let change = {
+                let separator = {
+                    let line_contains_only_space = text.char(end) == '\n';
```
the-mikedavis, line 4416: This check won't work for non-LF line ending documents. For example if you try the reproduction case from the issue with `:line-ending crlf` and then `JJ` it will still have the empty space.

the-mikedavis:
This would be a good opportunity to add a test case for `J` to the integration tests in `helix-term/tests/test/commands.rs`

helix-term/src/commands.rs
TornaxO7: updated. I'd like to have this block at least:
```rust
        let lines = {
            let (start, mut end) = selection.line_range(slice);
            if start == end {
                end = (end + 1).min(text.len_lines() - 1);
            }
            start..end
        };
```
makes it more readable, in my opinion, since we are seeing that we just need `stard` and `end` for `lines` but I can remove it if you insist.

helix-term/src/commands.rs
TornaxO7: updated

TornaxO7:
Damn, hm... may I ask if there's a possible way to start the tests for the other platforms?

TornaxO7:
ok, just found [act](https://github.com/nektos/act) and I'll try with it.

helix-term/src/commands.rs
pascalkuthe: I think this kind of extra indention ist actually very unreadable and clutterwd. Please reverse the Extra nestong jere and also remove the unnecssary nestong in the cide you added.

the-mikedavis: Yeah in general I'm not a big fan of the `let var = { .. }` syntax unless we're avoiding some really nasty scoping issues. Readability / cosmetic changes like this are always debatable and they inflate the diff (make review harder and might introduce unnecessary conflicts with other PRs) so we try to avoid them

TornaxO7: reverted

TornaxO7:
@the-mikedavis ok, could you please restart the workflow please? Hopefully it will work this time.

helix-term/src/commands.rs
```diff
@@ -4411,7 +4410,19 @@ fn join_selections_impl(cx: &mut Context, select_space: bool) {
...
+                    if line_contains_only_space {
+                        None
+                    } else {
+                        Some(Tendril::from(" "))
+                    }
+                };
+
+                (start, end, separator)
+            };
             changes.push(change);
```
the-mikedavis, line 4426: Just some style nits:

* I'd like to avoid the extra nesting here too
* Turning an `if` condition into a var with a good name is a nice tool for long `if` conditions but IMO it's always clearer to just have a comment explaining what the `if` is branching on:

```rust
let separator = if end == line_end_char_index(&slice, line + 1) {
    // The line only contains whitespace so skip adding a space.
    None
} else {
    Some(Tendril::from(" "))
};
// need to skip from start, not end
changes.push((start, end, separator));
```

the-mikedavis (commented):
This is looking good :+1:

Just some small style nits

the-mikedavis (approved):
Thanks for fixing this!

pascalkuthe (approved):

cezhunter:
is there a way to join lines *without* including spaces? Say you have something like
```
func
    .test()
    .evaluate()
    .whatever()
```
I want to join the lines into one but not have it be `func .test() .evaluate() .whatever()`

TornaxO7:
Hm... we could introduce a special case: If the current line starts with a punctuation, the whitespace of the previous line should be omitted. The other question is: Is there a case where the whitespace shouldn't be removed if the current line starts with a punctuation?

the-mikedavis:
For that case you can use `A-J` (<kbd>Alt</kbd>+<kbd>Shift</kbd>+<kbd>j</kbd>) which joins lines but selects the separator. Then you can use `d` to delete the spaces.

cezhunter:
Yup - actually figured it out shortly after I commented (`Alt-Shift-J` and `d`). That is something I like about Helix - if what you expect doesn't work try it with `shift` or `alt` :) same thing with paste (`shift` lets you paste without the newline)

PR #9001, opened 2023-12-05
Update goto_file docs

jw013:
Make the pluralization of files and selections consistent to emphasize the 1-to-1 relation between files and selections. The prior wording with plural "files" and singular "selection" can mislead users into thinking the command can open multiple files from a single selection.


the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

PR #9002, opened 2023-12-05
Fix cursor dissapearing when over indent lines in Darcula themes

jesusmgg:
Before:
before

After:
after

The cursor is in the exact same position in both screenshots.


the-mikedavis (approved):

getreu:
#8234 is related

PR #9003, opened 2023-12-05
Update gleam grammar and queries

fschwalbe:
Updates tree-sitter-gleam to 0.32.4.

I also updated the queries according to the ones [from tree-sitter-gleam itself](https://github.com/gleam-lang/tree-sitter-gleam/tree/v0.32.4/queries) and from a quick test everything seems to work fine.


the-mikedavis (approved):
Thank you!

PR #1998, opened 2022-04-07
add languages `r` and `rmarkdown`

mtoohey31:
A few notes:

- The queries for r files are copied from the grammar [here](https://github.com/r-lib/tree-sitter-r).
- All the queries for rmarkdown inherit from markdown, though rmarkdown is basically pandoc markdown, so it has more syntax features that aren't supported by the plain markdown grammar. The plain markdown grammar is the best option we have though because pandoc markdown supports inline latex which makes it really complicated, so there's no complete tree-sitter grammar.
- rmarkdown needs to be a separate language from markdown because the language server can analyze the contents of `{r}` code blocks but should only be enabled in `rmd`/`Rmd` files.
- Both cases are included for file extensions because I've seen both used quite frequently.

And two questions: 

- I included a shebang definition because it technically works, since `R` is an interpreted language, but I've never seen someone use a shebang in an R file. Any opinion on whether I should leave it or remove it?
- The generated `lang-support.md` file appears to use the name of the language server command for the default language server column. However, the name of the command for the language server in use here ([this one](https://github.com/REditorSupport/languageserver/), it's the only one I'm aware of) is just the `R` interpreter itself, and the arguments instruct it to invoke the `languageserver` package. Do we want to add in a workaround for this kind of situation so we can name the `languageserver` package specifically, or should we just leave it?

PR #9012, opened 2023-12-05
Change R markdown language id to fix language server filetype detection

mtoohey31:
This merge request renames the R markdown language from `rmarkdown` to `rmd`. When I added R and R markdown support back in #1998, I mistakenly chose the `rmarkdown` name because I didn't realize it was used for anything other than the display name in Helix's statusline. Switching the name to `rmd` is necessary to fix the R language server's filetype detection, which relies on the language name (see [this line of code](https://github.com/REditorSupport/languageserver/blob/7f7605b55a6cbe05e0033b40f21e87408c06e102/R/document.R#L18)).

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1668,6 +1668,7 @@ source = { git = "https://github.com/r-lib/tree-sitter-r", rev = "cc04302e1bff76
 [[language]]
 name = "rmarkdown"
 scope = "source.rmd"
+language-id = "rmd"
 injection-regex = "(r|R)md"
 file-types = ["rmd", "Rmd"]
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
I think we should fix this by adding the `language-id = "rmd"` key instead: that will send `rmd` to the language server but still keep `rmarkdown` as the name for `:set-language` etc.

mtoohey31:
Cool, didn't know that was a thing, thanks @the-mikedavis. I've confirmed that the language server works if we change that instead and pushed the simplified changes.

the-mikedavis (approved):
Thanks!

PR #9013, opened 2023-12-06
Fix version of Nix package

Dragonink:
The version of the Nix package is using a placeholder (probably since bf7c4e1).

This PR fixes this issue by taking the package's version from the workspace `Cargo.toml`.


the-mikedavis (approved):
Thanks!

Issue #8853, opened 2023-11-18
RFC: Revamp Config System


pascalkuthe:
> We currently don't have an RFC system, this is an experiment of starting a more RFC-like process before a larger change. I am also fairly busy right now so I am also creating this issue in the hope that somebody else may pick this issue up

# Problem Statement

Currently, configuration in helix is closely coupled to serde/toml. Specifically, configuration is handled by derive(Desirilaize) on a struct. This approach was easy to get started with but doesn't scale well:

* Every config option requires a handwritten struct field.
	* If we want to overwrite a config option in the language config we need to add that config to the language config aswell. This doesn't scale, its impractical to duplicate the entire config
	* This won't work for additional configuration added by plugins
* Similarly if we want to overwrite a config option per buffer that requires yet another handwritten field on `Document` (which already polluted by too many fields). 
* The current config is not well suited for mutation. The `:set` and `:toggle` commands essentially serialize the current config to json, edit them and desiralize again. Serde however (especially with our many custom deserilaizer) doesn't necessarily roundtrip correctly through json. 
	* This becomes especially relevant with scheme plugins/scripts where the config should be both easily (and efficiently) readable and writable
* Overwrite/Inheritance/merging is inconsistent (we have an open issue about that) with regard to depth
* inspecting config options (and including documentation like we do for commands) and auto-generating docs is not possible/very hard

# Prior work

The solution I would like to see is something that mirrors kakoune. I quite liked how simple yet flexible the config is. Especially the concept of scopes works quite well IMO. However, our needs are slightly more complex since kakoune doesn't have language config/LSP (which I would like to stay effortless). I cameup with a way to also use the scope system for these use-cases.

# Proposed Solution

* completely separate our config model from serde
	* we plan to eventually move to scheme and abandon the toml config.
	* until then we will have to implement (a backwards compatible) compatibility layer.

* All options will have a simple canonical in-memory representation that is easy to read and modify. Convenience syntax should be handled during parsing.
	* example: instead of `auto-pairs` accepting either a map of autopairs (like `{ "(" = ")", "{" = "}"`) or a bool to indicate whether they are enabled simply add two separate options: `auto-pairs.enable` and `auto-pairs.pairs`. This allows consistent merging/inheritance/defaulting and doesn't require weird special cases in the config system. 
	
* config options are represented as a flat called `OptionManager`.
	* Example: `softwrap.enable` and `softwrap.max-wrap` are two entirely separate config options. The only thing that links them is a naming convention.
[...]

PR #9018, opened 2023-12-06
set soft-wrap enabled by default in markdown files

### LINKED ISSUE #8331: `set soft-wrap.enable true` should be set by default in `.md` files
I suppose that it is sensible to set soft-wrap feature to true in the default settings for [markdown][1] files as it often involves long paragraphs....

[1]: https://github.com/helix-editor/helix/blob/master/languages.toml#L1239

arcstur:
closes #8331

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1333,6 +1333,7 @@ file-types = ["md", "markdown", "PULLREQ_EDITMSG", "mkd", "mdwn", "mdown", "mark
 roots = [".marksman.toml"]
 language-servers = [ "marksman" ]
 indent = { tab-width = 2, unit = "  " }
+soft-wrap.enable = true
 
 [[grammar]]
 name = "markdown"


dead10ck (approved):

dead10ck:
What is the behavior of overrides with the language setting? If this is set to true by default for markdown, and you toggle the global option off, what wins? I'd still like to be able to toggle it off at runtime.

dead10ck:
Just tried it out locally. I can confirm that the language setting always wins. And it goes both ways: even if you set it to false in your own `languages.toml`, you can't turn it on at runtime either.

This seems like a broader issue that's tangential to this PR, but I'd consider it a blocker for this, unfortunately.

dead10ck:
Ah, thanks Pascal, somehow that issue flew under my radar. I'll take a read through it when I get some time.

arcstur:
Glad you catched the misbehaviour! I think we can close this, since it's a simple PR, and discuss the config overhaul.

pascalkuthe:
Just to expand: You would be able to set config in different scipes lik in kakoune so you could do:

* `:set markdown softwrap.enable false`
* `:set buffer softwrap.enable false`
* `:set buffer-language softwrap.enable false`
* `:set global softwrap.enable false`

Each lf these scope are essentially their own full comfig (with the same inheritance rules as we have currently) so even all the first 3 options could be used to disable softwrap for markdown with this PR merged (last option is the same the current set does, the second last option would overwrite the setting for the current buffer and the first two would change the language settings)

the-mikedavis:
Yeah let's close this out for now pending the config changes in #8853

Agilan989:
Msg noted 


Sent from Yahoo Mail for iPhone


On Wednesday, December 6, 2023, 9:59 PM, Artur Corrêa Souza ***@***.***> wrote:




Glad you catched the misbehaviour! I think we can close this, since it's a simple PR, and discuss the config overhaul.

PR #9024, opened 2023-12-07
Add Hoon

ilyakooo0:


languages.toml
```diff
@@ -2952,3 +2952,18 @@ file-types = ["log"]
...
 source = { git = "https://github.com/Tudyx/tree-sitter-log", rev = "62cfe307e942af3417171243b599cc7deac5eab9" }
+
+[[grammar]]
+name = "hoon"
+source = { git = "https://github.com/urbit-pilled/tree-sitter-hoon", rev = "1d5df35af3e0afe592832a67b9fb3feeeba1f7b6" }
+
+[[language]]
+name = "hoon"
+scope = "source.hoon"
+grammar = "hoon"
```
the-mikedavis, line 2963: This key can be omitted since the `[[grammar]]` entry uses the same name as the `[[language]]` name

languages.toml
```diff
@@ -2952,3 +2952,18 @@ file-types = ["log"]
...
+injection-regex = "hoon"
+file-types = ["hoon"]
+comment-token = "::"
+indent = {tab-width = 2, unit = "  "}
+roots = []
```
the-mikedavis, line 2968: This line can be removed now that we default this field: https://github.com/helix-editor/helix/pull/8803

the-mikedavis (commented):
Helix needs queries under `runtime/queries/hoon/highlights.scm` in order to provider syntax highlighting: https://docs.helix-editor.com/master/guides/adding_languages.html#queries

runtime/queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
+(number) @number
```
the-mikedavis, line 1:
```suggestion
(number) @constant.numeric
```

We use some custom capture names: https://docs.helix-editor.com/master/themes.html#syntax-highlighting

runtime/queries/hoon/highlights.scm
```diff
@@ -0,0 +1,32 @@
...
+
+(rune) @keyword
+
+(term) @constant
+
+(aura) @constant.builtin
+
+(Gap) @comment
+
+(boolean) @constant.builtin
```
the-mikedavis, line 26:
```suggestion
(boolean) @constant.builtin.boolean
```

languages.toml
```diff
@@ -2952,3 +2952,16 @@ file-types = ["log"]
 [[grammar]]
 name = "log"
 source = { git = "https://github.com/Tudyx/tree-sitter-log", rev = "62cfe307e942af3417171243b599cc7deac5eab9" }
+
+[[grammar]]
+name = "hoon"
+source = { git = "https://github.com/urbit-pilled/tree-sitter-hoon", rev = "1d5df35af3e0afe592832a67b9fb3feeeba1f7b6" }
```
gabydd, line 2958: Can you move this below the language definition so it's formatted the same as the other languages

gabydd (changes requested):

gabydd:
Looks like there is a language server: https://github.com/urbit/hoon-language-server might be nice to add the config for it if it works and add it to the wiki. If not that's not a blocker

ilyakooo0:
Running the language server isn't as simple as running a binary at the root of the project. There's more setup involved. 

From what I heard the language server doesn't work that well and nobody actually uses it.

gabydd (approved):

the-mikedavis (approved):

languages.toml
```diff
@@ -2999,6 +2999,17 @@ name = "log"
...
+injection-regex = "hoon"
+file-types = ["hoon"]
+comment-token = "::"
+indent = {tab-width = 2, unit = "  "}
+
+[[grammar]]
+name = "hoon"
+source = { git = "https://github.com/urbit-pilled/tree-sitter-hoon", rev = "1d5df35af3e0afe592832a67b9fb3feeeba1f7b6" }
+
 name = "janet"
```
the-mikedavis, line 3013: It looks like the merge cut off some lines. The `Hoon` line at the top here can be removed and the `name = "janet"` line should have a line that says `[[language]]` above it

Issue #8822, opened 2023-11-15
goto_file path with `~`


ModProg:
`goto_file` is probably most useful in non lsp files like scripts or configurations, in these `~` usually expands to `$HOME` it would make sense to honor this.

TornaxO7:
I'd like to give this a try.

Issue #8824, opened 2023-11-15
more flexible `goto_file` 


ModProg:
After opening #8822 I noticed that there are more things I miss from `vim` in regard to `gf`.

1. Allow non `.` punctuation after a path, e.g., `path,` or `"path";` should open `path` and not `path,` and `path";` respectively, if you want the latter behavior it is easy enough to just do a `f,gf`.  (maybe we could have a config option for what characters are path, like `isfname` in VIM)
2. Support paths that omit the file extension. Simple behavior could be “if file not exist, try appending current extension”, but a more flexible tool could be a file type based `suffixesadd` (VIM) like list of extensions to try.
3. VIM also as a `includeexpr` which allows transforming a path by replacing characters, i.e., `.` with `/` in java imports. (This is applied after the normal behavior failed.) This might be best implemented when the scripting support lands, though we could also just have a `regex, replacements` tuple that would probably already cover most things.

Not every file has an LSP, and this just allows navigating these almost as if one had one.

ModProg:
I feel like at least 1. is very actionable, and I'd look into implementing this and #8822. 

If you agree with the other additions, I'd look into those as well, though I'd need some input on how the config options should look.

daedroza:
I am unsure about the direction of VIM about such this feature but can we instead have file picker show up with the relevant text in search box?

Until it is implemented is there a way to map a new keyboard shortcut to search for file using file picker with selected text under cursor?

TornaxO7:
for anyone who subscribed to this: Could you try out https://github.com/helix-editor/helix/pull/9065 pls? It should basically behave the same as in neovim.

PR #9038, opened 2023-12-09
Feat/extend goto file

TornaxO7:
Closes https://github.com/helix-editor/helix/issues/8822 https://github.com/helix-editor/helix/issues/8824

I'm unsure how to write a test with:
- environment-variables
- `~`

Also executing `gf` with:
```rust
let a = "some/path".to_string();
```
will interpret it as `some/path".to_string`. The only solution which I can think of, is manually collecting all "valid" characters on the right and on the left of the cursors, should I do it like that?


helix-term/tests/test/commands.rs
```diff
@@ -153,6 +153,17 @@ async fn test_goto_file_impl() -> anyhow::Result<()> {
     )
     .await?;
 
+    // path ends with ","
+    test_key_sequence(
+        &mut AppBuilder::new().with_file(file.path(), None).build()?,
+        Some("i/tmp/test.txt,<esc>bgf"),
```
ModProg, line 159: This test seems a bit strange, the cursor is on the `,` when triggering `gf`, meaning a character that is excluded from the output path.

ModProg:
> will interpret it as `some/path".to_string`. The only solution which I can think of, is manually collecting all "valid" characters on the right and on the left of the cursor, should I do it like that?

On non selection `gfs`, there are two approaches, one would be to have a constant that defines a set of characters that are allowed/disallowed in paths, and you just go left and right until you find one that matches. This might result in poor performance in some cases, not allowing paths to contain `",;'` … at all, so it might make sense to have a more advanced matcher that allows them under certain conditions, e.g., allows `,;` if they are not followed by whitespace, e.g., `hello,world.txt` would be one path, but `hello, world.txt` would be two, `hello` and `world.txt`.

One thing one might consider, is using treesitter if available, this would allow actual string support, and could even support spaces. This is probably more of a future extension of this, as it would only work when treesitter is supported for the file at hand.

TornaxO7:
Hm... good point, I'll try to make use of tree-sitter here

kirawi:
You can see how tree-sitter is used for matching brackets as a stepping stone: https://github.com/helix-editor/helix/blob/510928618dfcddcd6d3086b278332fa93ee32998/helix-core/src/match_brackets.rs#L54

TornaxO7:
eh, I don't remember that I closed this O.O

the-mikedavis:
Github will automatically close PRs if it doesn't see any difference between the PR's branch and the target branch. It looks like you force-pushed the branch to 5109286 which is a commit from master

TornaxO7:
oooh ok, thank you for the information!

PR #9065, opened 2023-12-12
Improve `goto_file_impl`

TornaxO7:
continues https://github.com/helix-editor/helix/pull/9038

Closes https://github.com/helix-editor/helix/issues/8822 https://github.com/helix-editor/helix/issues/8824


TornaxO7:
@ModProg

may I ask if you can try it out?

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            let valid_chars: &[char] = if cfg!(target_os = "windows") {
+                &[
+                    '@', '/', '\\', '.', '-', '_', '+', ',', '#', '$', '%', '{', '}', '[', ']',
+                    ':', '!', '~', '=',
+                ]
+            } else {
+                &['@', '/', '.', '-', '_', '+', ',', '#', '$', '%', '~', '=']
+            };
+            valid_chars.contains(c) || c.is_alphabetic()
+        };
```
TornaxO7, line 1191: I looked which characters neovim checks. See `:h isf`.

TornaxO7:
# File path detection strategy
I decided to read the chars next to the cursor one by one because:

1. using treesitter requires a parser which doesn't exist for text or log files
   although paths could occur there
2. paths are short enough to be read char by char in my opinion without waiting too
   long

# Other information
So I tested it on `$HOME/test.txt` and `~/test.txt`. Both opened up `~/test.txt` which is expected
but I don't know how to write tests for them.

ModProg:
I'd consider adding `,` to the characters terminating a path (like `;`).

TornaxO7:
> I'd consider adding `,` to the characters terminating a path (like `;`).

added

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+        let is_valid_path_char = |c: &char| {
+            let valid_chars: &[char] = if cfg!(target_os = "windows") {
+                &[
+                    '@', '/', '\\', '.', '-', '_', '+', '#', '$', '%', '{', '}', '[', ']', ':',
+                    '!', '~', '=',
+                ]
+            } else {
+                &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '=']
+            };
+            valid_chars.contains(c) || c.is_alphabetic()
```
TornaxO7, line 1190: I looked that up from `neovim`. See `:h isf` in `neovim` for the list of chars.

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+        };
+
+        let path = {
+            let start = {
```
pascalkuthe, line 1194: The logic here is unnecessarily complicated. Simply start counting (see other comment) at the next/prev char (that will be required to address the other comment anyway)

pascalkuthe (changes requested):

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+
+            let head = start
+                .clone()
+                .reversed()
+                .take_while(is_valid_path_char)
+                .collect::<String>()
+                .chars()
+                .rev()
+                .collect::<String>();
+
```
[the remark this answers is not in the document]
TornaxO7: good point!

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                    '@', '/', '\\', '.', '-', '_', '+', '#', '$', '%', '{', '}', '[', ']', ':',
+                    '!', '~', '=',
+                ]
+            } else {
+                &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '=']
+            };
+            valid_chars.contains(c) || c.is_alphabetic()
+        };
+
+        let path = {
```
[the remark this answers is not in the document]
TornaxO7: Sorry! I'm still unfamiliar with the codestyle. It's removed now :+1:

helix-term/src/commands.rs
TornaxO7: I hope that it's better now.

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,57 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                    '@', '/', '\\', '.', '-', '_', '+', '#', '$', '%', '{', '}', '[', ']', ':',
+                    '!', '~', '=',
+                ]
+            } else {
+                &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '=']
+            };
+            valid_chars.contains(c) || c.is_alphabetic()
+        };
+
+        let start_pos = {
```
the-mikedavis, line 1193: Same here about nested assignments

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,57 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            } else if is_valid_path_char(&text.char(pre_cursor_pos)) {
+                pre_cursor_pos
+            } else {
+                post_cursor_pos
+            }
+        };
+
+        let prefix_len = text
+            .chars_at(start_pos)
+            .clone()
```
the-mikedavis, line 1209: I think this clone is unnecessary, no? I'm surprised that clippy doesn't catch it

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,57 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+        let postfix_len = text
+            .chars_at(start_pos)
+            .take_while(is_valid_path_char)
+            .count();
+
+        let path = Cow::Borrowed(
+            text.slice((start_pos - prefix_len)..(start_pos + postfix_len))
+                .as_str()
+                .unwrap(),
         );
```
the-mikedavis, line 1223:
```suggestion
        let path: Cow<str> = text
            .slice((start_pos - prefix_len)..(start_pos + postfix_len))
            .into();
```

helix-term/src/commands.rs
```diff
@@ -1176,25 +1176,55 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                .rev()
+                .collect::<String>();
+
+            let tail = start.take_while(is_valid_path_char).collect::<String>();
+
+            format!("{}{}", head, tail)
+        };
+        log::debug!("Goto file path: {}", path);
+
+        match shellexpand::full(&path) {
```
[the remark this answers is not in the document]
the-mikedavis: I'm wondering if need to do shell expansion at all? We could pass the path through `expand_tilde` to take care of that common case. Full shell expansion seems like it might be overkill to me

the-mikedavis (changes requested):

helix-term/src/commands.rs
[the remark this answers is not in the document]
TornaxO7: I didn't knew that there's `expand_tilde`. I removed the dependency and using `expand_tilde` now.

helix-term/src/commands.rs
TornaxO7: good catch. Indeed, removing the clone doesn't create an error.

TornaxO7:
I can't execute the integration tests. Executing `cargo integration-test` gives me:
```
error[E0412]: cannot find type `Id` in module `tokio::runtime`
  --> helix-event/src/runtime.rs:45:64
   |
45 |         parking_lot::RwLock<hashbrown::HashMap<tokio::runtime::Id, &'static T, ahash::RandomState>>,
   |                                                                ^^ not found in `tokio::runtime`

   Compiling gix-tempfile v13.0.0
error[E0599]: no method named `id` found for struct `Handle` in the current scope
  --> helix-event/src/runtime.rs:67:52
   |
67 |         let id = tokio::runtime::Handle::current().id();
   |                                                    ^^ method not found in `Handle`

Some errors have detailed explanations: E0412, E0599.
For more information about an error, try `rustc --explain E0412`.
error: could not compile `helix-event` (lib) due to 2 previous errors
warning: build failed, waiting for other jobs to finish...
```

helix-term/src/commands.rs
```diff
@@ -1180,25 +1183,51 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+                '~', '=',
+            ];
+            #[cfg(not(target_os = "windows"))]
+            let valid_chars: &[char] = &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '='];
+
+            valid_chars.contains(c) || c.is_alphabetic() || c.is_numeric()
+        };
+
+        let cursor_pos = primary.cursor(text.slice(..));
+        let pre_cursor_pos = cursor_pos.saturating_sub(1);
```
pascalkuthe, line 1201: I think I also addressed this in a previous comment but saturing ad dis not necessary here. We will never exhaust 64bit adresses.

TornaxO7: removed :+1:

helix-term/src/commands.rs
```diff
@@ -1180,25 +1183,51 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
-        );
+
+        let is_valid_path_char = |c: &char| {
+            #[cfg(target_os = "windows")]
+            let valid_chars: &[char] = &[
+                '@', '/', '\\', '.', '-', '_', '+', '#', '$', '%', '{', '}', '[', ']', ':', '!',
+                '~', '=',
+            ];
+            #[cfg(not(target_os = "windows"))]
+            let valid_chars: &[char] = &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '='];
```
the-mikedavis, line 1195: We should also include `:` here so that `gf` can work on links. (since https://github.com/helix-editor/helix/pull/5820)

the-mikedavis: Also I believe the explicit types are not necessary

```suggestion
            let valid_chars = &['@', '/', '.', '-', '_', '+', '#', '$', '%', '~', '='];
```

TornaxO7: applied

pascalkuthe (approved):

helix-term/src/commands.rs
```diff
@@ -1196,25 +1199,51 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
...
+            .take_while(is_valid_path_char)
+            .count();
+
+        let postfix_len = text
+            .chars_at(start_pos)
+            .take_while(is_valid_path_char)
+            .count();
+
+        let path: Cow<str> = text
+            .slice((start_pos - prefix_len)..(start_pos + postfix_len))
```
the-mikedavis, line 1239: For readers' context: rope slicing is by character index rather than byte index (like `&str`) which is why this uses `chars_at(..).count()` rather than a len of a substring

the-mikedavis (approved):
Thanks!

ModProg:
Has there been a regression here? I just noticed `goto_file` opening files ending in not only `,` and `;` but also `.` without a selection?

ModProg:
Just double checked and it works in 24.7 but is broken in 25.1.1.

PR #9069, opened 2023-12-13
Update keymap.md: improve grammar

valerii15298:


book/src/keymap.md
```diff
@@ -25,7 +25,7 @@
 
 ## Normal mode
 
-Normal mode is the default mode when you launch helix. Return to it from other modes by typing `Escape`.
+Normal mode is the default mode when you launch helix. Return to it from other modes by pressing the `Escape` key.
```
David-Else, line 28:
```suggestion
Normal mode is the default mode when you launch helix. You can return to it from other modes by pressing the `Escape` key.
```

David-Else (commented):
I think this is even better.

the-mikedavis (approved):

pascalkuthe (approved):

Issue #9082, opened 2023-12-14
Markdown file indent-style detection and crash


sicher:
### Summary

The tab width auto detection seems to bug on some markdown files and cause a crash when trying to probe the indent-style setting.

### Reproduction Steps

 

I tried this:

1. Create a markdown file `test.md` with the following content:
```
# A test file

          TEST
  Here is some text.
```

2. `hx test.md`
3. type `:indent-style`

I expected this to happen:

Helix to report the current indent-style setting.

Instead, this happened:

Helix crashes with `internal error: entered unreachable code
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace`

### Platform

Windows

### Terminal Emulator

Windows Terminal Version: 1.17.11461.0

### Installation Method

scoop

### Helix Version

helix 23.10 (f6021dd0)

TornaxO7:
> type :indent-type

I think you mean `:indent-style` but yeah. I can reproduce this and I'd like to claim this.

sicher:
> > type :indent-type
> 
> I think you mean `:indent-style` but yeah. I can reproduce this and I'd like to claim this.

Oh, yeah... sorry about that... :)

Also, maybe a different issue but I wonder if the indent auto detection is working as expected. I get 10 space indents set in that file.

TornaxO7:
> Also, maybe a different issue but I wonder if the indent auto detection is working as expected. I get 10 space indents set in that file.

Indeed, I just found that out when you posted your answer xD

pascalkuthe:
there are two things going on here which I was kind of aware of before:

* indent style detection for very short files (like this one) doens't work well. Test is indented with 10 spaces and that is what it uses. I think the exact ratios need to be adjusted in our detection heuristic (potentially we need to add some absolute requirements in addition to values, this may be a case where it hits 50%).

* the check in `:indent-style` assumes that indent can have at most 8 spaces but 16 is possible (this was adjusted in a semi-recent PR). The check should just be removed (and the _=> unreachable branches removed)

TornaxO7:
@pascalkuthe so may I ask what you suggest for a PR for this? Should the PR just remove the unreachable branch?

pascalkuthe:
yeah mostly just remove the unreachable branch altough fixing the ident heuristic would be a nice bonus

TornaxO7:
aye aye captain

PR #9087, opened 2023-12-14
fix `:indent-style` crash

TornaxO7:
Closes https://github.com/helix-editor/helix/issues/9082


TornaxO7:
I couldn't think of a better solution for the indent-auto-detecting function because I agree its implementation.

the-mikedavis (approved):

pascalkuthe:
I do still want to investigate the heuristic. I do think it can be I.proved and that this case was a bug

pascalkuthe:
I will merge this for now but we should investigate further

pascalkuthe (approved):

PR #9089, opened 2023-12-15
add smali language support

evanrichter:


runtime/queries/smali/highlights.scm
```diff
@@ -0,0 +1,218 @@
...
+   . (identifier) @_first @type.builtin
+   .  (identifier) @_second @type.builtin
+   (identifier) @type.builtin)
+  (#eq? @_first "com")
+  (#any-of? @_second "android" "google"))
+
+; Methods
+
+(method_definition
+  (method_signature (method_identifier) @method))
```
archseer, line 23: Quite a few scopes in this file don't match our expected scopes so they won't highlight correctly: https://docs.helix-editor.com/themes.html#syntax-highlighting

evanrichter: Ok I'll work on it!

I just copied these scm files from upstream and it looked good enough to me in the smali files I am editing

evanrichter: Should be much better now! I left my changes as a separate commit for review, but I can easily rebase this PR into one commit if you prefer

runtime/queries/smali/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+  (#any-of? @constructor "<init>" "<clinit>"))
+
+"constructor" @constructor
+
+; Fields
+
+[
+  (field_identifier)
+  (annotation_key)
+] @field
```
the-mikedavis, line 74:
```suggestion
] @variable.other.member
```

runtime/queries/smali/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+  ".catchall"
+] @keyword.control.exception
+
+; Literals
+
+(string) @string
+(source_directive (string "\"" _ @string.special.url "\""))
+(escape_sequence) @constant.character.escape
+
+(character) @character
```
the-mikedavis, line 170:
```suggestion
(character) @constant.character
```

runtime/queries/smali/highlights.scm
```diff
@@ -0,0 +1,216 @@
...
+   . (identifier) @_first @type.builtin
+   (identifier) @type.builtin)
+  (#any-of? @_first "android" "dalvik" "java" "kotlinx"))
+
+((class_identifier
+   . (identifier) @_first @type.builtin
+   . (identifier) @_second @type.builtin
+   (identifier) @type.builtin)
+  (#eq? @_first "com")
+  (#any-of? @_second "android" "google"))
```
the-mikedavis, line 18: We don't currently support `#any-of?`. These will need to be rewritten to use `#match?` instead, for example `(#match? @_second "^(android|google)$")`

runtime/queries/smali/injections.scm
```diff
@@ -0,0 +1 @@
+(comment) @comment
```
the-mikedavis, line 1:
```suggestion
((comment) @injection.content
 (#set! injection.language "comment"))
```

we follow tree-sitter for injections https://tree-sitter.github.io/tree-sitter/syntax-highlighting#language-injection, nvim has its own injection format

runtime/queries/smali/locals.scm
the-mikedavis, line 1: We also follow tree-sitter for locals: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

So these need to be rewritten to only use the captures `@local.scope`, `@local.reference` and `@local.definition`

runtime/queries/smali/indents.scm
the-mikedavis, line 1: These look like nvim indent queries. See https://docs.helix-editor.com/master/guides/indent.html for writing Helix ones

languages.toml
```diff
@@ -843,6 +843,19 @@ indent = { tab-width = 2, unit = "  " }
 name = "java"
 source = { git = "https://github.com/tree-sitter/tree-sitter-java", rev = "09d650def6cdf7f479f4b78f595e9ef5b58ce31e" }
 
+[[language]]
+name = "smali"
+scope = "source.smali"
+injection-regex = "smali"
+file-types = ["smali"]
+comment-token = "#"
+language-servers = []
```
the-mikedavis, line 852: This is the default so this key can be removed

runtime/queries/smali/indents.scm
evanrichter: ok I did my best! smali does not use indentation too much (it's basically just a text representation of dalvik bytecode with gotos for control flow). But I checked the behavior with this example with the most indentation I could find and it seems correct:

```smali
.method public static updateDigest(Ljava/security/MessageDigest;Ljava/io/File;)V
    .locals 2
    .annotation system Ldalvik/annotation/Throws;
        value = {
            Ljava/io/IOException;
        }
    .end annotation
```

runtime/queries/smali/injections.scm
evanrichter: I made the change as suggested, and this reminded me to check `goto_next_comment` which I found doesn't work.

tree-sitter-subtree does report `(comment)` for comment lines, so I don't know what to check beyond that

the-mikedavis (approved):

pascalkuthe (approved):

PR #9097, opened 2023-12-15
Update keymap.md

Ozymandias42:
Added key-binds for add cursor `Shift-C` and collapse cursors into one `,` to Normal Mode section.

Had to search for this far too often in issues as a vim user used to visual block mode editing


gabydd:
These are already documented under selection manipulation

PR #9099, opened 2023-12-16
Update tokyonight themes

UntimelyCreation:
Hey there!

After seeing some [auto-generated Helix themes](https://github.com/folke/tokyonight.nvim/tree/main/extras/helix) by folke in his `neovim` tokyonight repository, I decided to update Helix's themes accordingly.

I carefully went through everything and landed on a clean, best-of-both-worlds result which I think is a nice improvement from the previous versions. I've added the two missing themes, `tokyonight_moon` and `tokyonight_day`, and updated my email address.

Here are some screenshots below.

If you see any mistake or improvement to make, feel free to contact me or make any changes!

Cheers!

### DIFF
--- a/runtime/themes/tokyonight.toml
+++ b/runtime/themes/tokyonight.toml
@@ -1,75 +1,96 @@
-# Author: Paul Graydon <p.y.graydon@gmail.com>
+# Author: Paul Graydon <untimely.creation97@proton.me>
 
-"comment" = { fg = "comment", modifiers = ["italic"] }
-"constant" = { fg = "orange" }
+attribute = { fg = "cyan" }
+comment = { fg = "comment", modifiers = ["italic"] }
+"comment.block.documentation" = { fg = "yellow" }
+constant = { fg = "orange" }
+"constant.builtin" = { fg = "aqua" }
+"constant.character" = { fg = "light-green" }
 "constant.character.escape" = { fg = "magenta" }
-"function" = { fg = "blue", modifiers = ["italic"] }
+constructor = { fg = "aqua" }
+function = { fg = "blue", modifiers = ["italic"] }
+"function.builtin" = { fg = "aqua" }
 "function.macro" = { fg = "cyan" }
-"keyword" = { fg = "cyan", modifiers = ["italic"] }
+"function.special" = { fg = "cyan" }
+keyword = { fg = "purple", modifiers = ["italic"] }
 "keyword.control" = { fg = "magenta" }
 "keyword.control.import" = { fg = "cyan" }
-"keyword.operator" = { fg = "turquoise" }
-"keyword.function" = { fg = "magenta", modifiers = ["italic"] }
-"operator" = { fg = "turquoise" }
-"punctuation" = { fg = "turquoise" }
-"string" = { fg = "light-green" }
-"string.regexp" = { fg = "light-blue" }
-"tag" = { fg = "red" }
-"type" = { fg = "teal" }
-"namespace" = { fg = "blue" }
-"variable" = { fg = "white" }
+"keyword.control.return" = { fg = "purple", modifiers = ["italic"] }
+"keyword.directive" = { fg = "cyan" }
+"keyword.function" = { fg = "magenta" }
+"keyword.operator" = { fg = "magenta" }
+label = { fg = "blue" }
+namespace = { fg = "cyan" }
+operator = { fg = "turquoise" }
+punctuation = { fg = "turquoise" }
+special = { fg = "aqua" }
+string = { fg = "light-green" }
+"string.regexp" = { fg = "light-cyan" }
+"string.special" = { fg = "aqua" }
+tag = { fg = "magenta" }
+type = { fg = "aqua" }
+"type.builtin" = { fg = "aqua" }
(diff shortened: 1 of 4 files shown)


the-mikedavis (commented):
Could we keep tokyonight_night as tokyonight? This would be a breaking change for anyone using `theme = "tokyonight"` in their config

UntimelyCreation:
Very fair point, I just changed it back!

the-mikedavis (approved):

archseer:
What were the changes from the upstreamed theme and could we merge some of the improvements?

\cc @Finistere who has contributed this upstream (https://github.com/folke/tokyonight.nvim/pull/409 -- wish we knew about it sooner!)

UntimelyCreation:
What I did was use the upstream themes as a new baseline, and manually edit the file into something cleaner, mostly putting all colors into variables and dividing the file into sections in the spirit of the documentation and my previous themes.
I then corrected and added a few elements by hand to make it look nicer, as the upstream auto-generation seemed to miss a few thing, such as the ruler, indent guides, etc

runtime/themes/tokyonight.toml
```diff
@@ -1,96 +1,131 @@
...
+"ui.statusline.normal" = { bg = "blue", fg = "bg", modifiers = ["bold"] }
+"ui.statusline.insert" = { bg = "light-green", fg = "bg", modifiers = ["bold"] }
+"ui.statusline.select" = { bg = "magenta", fg = "bg", modifiers = ["bold"] }
+"ui.text" = { bg = "bg", fg = "fg" }
+"ui.text.focus" = { bg = "bg-visual" }
+"ui.text.inactive" = { fg = "comment", modifiers = ["italic"] }
+"ui.text.info" = { bg = "bg-menu", fg = "fg" }
+"ui.virtual.ruler" = { bg = "fg-gutter" }
+"ui.virtual.whitespace" = { fg = "fg-gutter" }
+"ui.virtual.inlay-hint" = {  bg = "bg-inlay", fg = "teal" }
```
SoraTenshi, line 92: As someone that uses `tokyo-night-storm` and inlay-hints, i'd be relatively unhappy in having so strong colours for `inlay-hints` they should imo be as unnoticeable as possible. That's why i also made them have no bg as well as its fg-color to be the same as comments (which imo is the best way to handle it).

Any insight as to why make them so bold in colour?

runtime/themes/tokyonight.toml
```diff
@@ -1,96 +1,131 @@
...
+"constant.character" = { fg = "light-green" }
 "constant.character.escape" = { fg = "magenta" }
-"function" = { fg = "blue", modifiers = ["italic"] }
+constructor = { fg = "aqua" }
+function = { fg = "blue", modifiers = ["italic"] }
+"function.builtin" = { fg = "aqua" }
 "function.macro" = { fg = "cyan" }
-"keyword" = { fg = "cyan", modifiers = ["italic"] }
+"function.special" = { fg = "cyan" }
+keyword = { fg = "purple", modifiers = ["italic"] }
```
SoraTenshi, line 15: Why change the normal keyword to purple?
Just a question, no particular opinion on it as of right now.

SoraTenshi (commented):
Generally this seems to refresh the theme quite a bit.
I added some comments, as i daily drive `tokyo-night-storm` and want to know what's going on with the theme i love so much.

runtime/themes/tokyonight.toml
UntimelyCreation: I just think of it the other way around. I like to make them stand out a bit from the base code as it is not part of it and is useful hinting, so I kept the upstream colors.
I also completely understand your point of view, and would be happy to leave them as they were before, which is `"ui.virtual.inlay-hint" = { fg = "comment" }` as you described.

runtime/themes/tokyonight.toml
UntimelyCreation: Coming from neovim, I try to stay as close as possible to folke's colors, as I intended when I first created the helix ones. You can see this as a long due, small correction. I just didn't want to make a PR just for this and waited for a more general update

runtime/themes/tokyonight.toml
SoraTenshi: I wouldn't mind giving your idea a test-drive, there's always the ability to revert it anyway in the future :)
Maybe i would base the decision on whether it should be colorful / plain based on how it is handled in other themes?

runtime/themes/tokyonight.toml
SoraTenshi: Sounds good to me!

SoraTenshi (approved):
Generally i think i am OK with how things end up either way in terms of the inlay-hints.

runtime/themes/tokyonight.toml
UntimelyCreation: After looking at the other themes, the inlay hints do seem to follow the more discreet approach, though there are both.
I'd say it's just a matter of personal preference.
As you seem OK with the change, let's leave it like this for now. There's always the option to change it back afterwards as you said!

pascalkuthe (approved):

PR #9101, opened 2023-12-16
when text document sync capability is only kind send didSave without text

### LINKED ISSUE #9034: Helix doesn't trigger `onDidSave` event in diagnostic-languageserver
### Summary

I have got [diagnostic-languageserver](https://github.com/iamcco/diagnostic-languageserver) working with Helix.
Everything works fine except linters with the `onSaveOnly` option set to `true`.

I did some research and found that the [`onDidSave`](https://github.com/microsoft/vscode-languageserver-node/blob/4812428a8621dff4153282407fc71494b994eda3/server/src/common/textDocuments.ts#L131) handler [right here](https://github.com/iamcco/diagnostic-languageserver/blob/a5c9565934d12c64d19116c1980576d89680c120/src/index.ts#L87) is never being called. 

In NeoVIM `onSaveOnly` works as expected and `onDidSave` is being called.

### Reproduction Steps

 

I tried this:
1. Install `diagnostic-languageserver`;
2. Modify `languages.toml` (see below);
3. Run `helix test.md`;
4. Add some text;
5. Hit `:w` to start linter;

No warning is reported. But it should.

6. Comment `onSaveOnly` in `languages.toml`;
7. Run `helix test.md`.

Warning will appear on first line.

Same with NeoVIM:
1. Install `diagnostic-languageserver` if not yet installed;
2. Install `lspconfig` module;
5. Modify `~/.config/neovim/init.lua` (see below);
6. Run `nvim test.md`;
7. Add some text;
8. Hit `:w` to start linter.

Warning will appear on first line.

Minimal `languages.toml`:
```toml
[[language]]
name = "markdown"
language-servers = ["markman", "dls"]

[language-server.dls]
command = "diagnostic-languageserver"
args = ["--stdio", "--log-level=4"]

[language-server.dls.config.filetypes]
"*" = "itworks"

[language-server.dls.config.linters.itworks]
command = "echo"
args = [":1:1: warning: It works!"]
sourceName = "itworks"
debounce = 1000
onSaveOnly = true  # Does not work with this.
formatLines = 1
formatPattern = [
  "^:(\\d+):(\\d+): ([^:]+): (.*)$",
  { line = 1, column = 2, security = 3, message = [4] }
]
```

Minimal nvim's `init.lua`:
```lua
require("lspconfig").diagnosticls.setup {
  filetypes = { "markdown" },
  cmd = { "diagnostic-languageserver", "--stdio", "--log-level=4" },
  init_options = {
    filetypes = {
      ["*"] = "itworks",
    },
    linters = {
      itworks = {
        command = "echo",
        args = { ":1:1: warning: It works!" },
        sourceName = "itworks",
        debounce = 1000,
        onSaveOnly = true,
        formatLines = 1,
        formatPattern = {
          "^:(\\d+):(\\d+): ([^:]+): (.*)$",
          { line = 1, column = 2, security = 3, message = { 4 } }
        }
      }
    }
  }
}
```

### Helix log

```
2023-12-08T11:23:05.434 helix_view::editor [ERROR] Failed to initialize the language servers for `source.md` - `markman` { Language server 'markman' not defined }
```

</details>

### Platform

Linux / Manjaro and OpenSUSE

### Terminal Emulator

gnome-console / tmux

### Installation Method

pacman / zypper and buit from source

### Helix Version

helix 23.10 (f6021dd0) / helix 23.10 (b81aacc5)

gabydd:
see https://github.com/microsoft/language-server-protocol/issues/288 for details

fixes #9034 and https://github.com/helix-editor/helix/discussions/8305

I tested that this does fixes the issues but it would be great if @akhilman could also confirm

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -969,20 +969,19 @@ impl Client {
     ) -> Option<impl Future<Output = Result<()>>> {
         let capabilities = self.capabilities.get().unwrap();
 
-        let include_text = match &capabilities.text_document_sync {
-            Some(lsp::TextDocumentSyncCapability::Options(lsp::TextDocumentSyncOptions {
-                save: Some(options),
+        let include_text = match &capabilities.text_document_sync.as_ref()? {
+            lsp::TextDocumentSyncCapability::Options(lsp::TextDocumentSyncOptions {
+                save: options,
                 ..
-            })) => match options {
+            }) => match options.as_ref()? {
                 lsp::TextDocumentSyncSaveOptions::Supported(true) => false,
                 lsp::TextDocumentSyncSaveOptions::SaveOptions(lsp_types::SaveOptions {
                     include_text,
                 }) => include_text.unwrap_or(false),
-                // Supported(false)
-                _ => return None,
+                lsp::TextDocumentSyncSaveOptions::Supported(false) => return None,
             },
-            // unsupported
-            _ => return None,
+            // see: https://github.com/microsoft/language-server-protocol/issues/288
+            lsp::TextDocumentSyncCapability::Kind(..) => false,
         };
 
         Some(self.notify::<lsp::notification::DidSaveTextDocument>(


akhilman:
Seems to be working fine. I can see `onDidSave` in the log and diagnostics in the editor.

helix-lsp/src/client.rs
```diff
@@ -981,6 +981,8 @@ impl Client {
                 // Supported(false)
                 _ => return None,
             },
+            // see: https://github.com/microsoft/language-server-protocol/issues/288
+            Some(lsp::TextDocumentSyncCapability::Kind(..)) => false,
             // unsupported
             _ => return None,
```
pascalkuthe, line 987: can we remove the blanket case now or turn it into an explicit case. Consedireng that this is pretty subtle (as evident by us getting this wrong) I think explicitly spelling out the other variants would be nicer

gabydd: Ah yeah I think the other variant is just None so that should be fine

pascalkuthe: if the other variant is really just `None` then you could use `.as_ref()?` instead of matching on the option

gabydd: Ah the other variant was not only none it was also when the first match arm was not matched for None for save of TextDocumentSaveOptions, I need to figure out what that should do

gabydd: That's still a good point about the as_ref though

gabydd: refactored it a little bit to use as_ref in two places where None should just return, and then also make sure we are explicitly covering each variant now

pascalkuthe (approved):

pascalkuthe:
thanks for looking into this!

the-mikedavis (approved):
Nice catch!

PR #9104, opened 2023-12-17
Add icon to Windows executable

### LINKED ISSUE #5940: Use helix logo as icon for the `hx` executable 😋
### Helix logo as icon file
* Is the icon file (not png image) for the new helix logo shared anywhere here on github repo or in the binary?
* as i found that [shortcut suggestion] to be _really_ effective, 
* and having the nice helix logo as the icon will be aweome.

[shortcut suggestion]: https://github.com/helix-editor/helix/issues/39#issuecomment-1418562617

### Proposal
* Can that icon be put on the `hx` executable file too? 
* I know that TUI exe files don't have any icon (and understandably too) but i'd say it doesn't hurt to have it there or does it?

in hx 22.12 the icon for exe is this:  doublecmd_VPMBhqyHf3

NewtonChutney:
This PR changes the build to include an icon for the Windows executable

fixes #5940

Icon [contributed](https://github.com/helix-editor/helix/issues/5940#issuecomment-1426958717) by @goyalyashpal


NewtonChutney:
Build artifact with icon:
image

NewtonChutney:
And somebody needs to clarify the cargo.lock changes..
I'm unable to `cargo build --locked`

assets/logo.svg
```diff
@@ -0,0 +1,29 @@
+<svg version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve"
```
archseer, line 1: Why were the assets moved around and modified?

NewtonChutney: Adding the icon to the root directory didn't look good.. Too many assets. So, I intended to move them into an assets directory similar to nushell's structure.. 

The logo.svg was a one liner while the logo-dark/light were multi line files.. I had run a formatter on all of them.

NewtonChutney: But yes, I could limit the scope of this PR to the main build file and the required dependencies..
Shall I revert the 2nd commit?

archseer: The file is a one liner since the newlines aren't necessary and only increase the file size (the file was minified)

icon.ico can be placed into contrib/, alongside AppImage files that are already there

NewtonChutney: Ahh I see, will make the changes

pascalkuthe:
I am personally not a fan of pulling in an extra dependency for something that IMO has very little practical use. Nobody start a TUI application from clicking on it, especially because you would need to run it inside a terminal emultor (so the shortcut would have the icon of the terminal emulator).

NewtonChutney:
> I am personally not a fan of pulling in an extra dependency for something that IMO has very little practical use. Nobody start a TUI application from clicking on it, especially because you would need to run it inside a terminal emultor (so the shortcut would have the icon of the terminal emulator).

Creating a shortcut that launches another terminal emulator (eg wezterm), might be the only way to launch a tui app in linux.. (the Linux desktop shortcut icon strategy)

But in Windows, we _can_ launch an tui app by double clicking it.. and it would launch in either conhost/WT (as per the user's choice).

Tabs: conhost (build with icon), conhost (stable release), windows terminal
image

NewtonChutney:
There's another approach that doesn't use external crates.. have to check: https://stackoverflow.com/a/76500197

NewtonChutney:
With winres, I see the binary's details are being injected too from the cargo config..
image

goyalyashpal:
> With winres, I see the binary's details are being injected too from the cargo config..

that's awesome @NewtonChutney - and if you urself hid that comment as OT, then i think u should unhide it.

OT: what window manager are you using which shows those "window tools" like move up, pin, size info, resize, etc above the window title bar? it looks useful as well as nice.

goyalyashpal:
_self hiding at offtopic_

----
OT: 

1. what window manager are you using which shows those "window tools" like move up, pin, size info, resize, etc above the window title bar? it looks useful as well as nice.

> The logo.svg was a one liner while the logo-dark/light were multi line files.. I had run a formatter on all of them.

2. lol, i tried same here: https://github.com/helix-editor/helix/pull/5941 . apparently, 96bits (`CLR\t` * 32 times) is a bit too much.

3. similar attempt for icon/desktop files for another distribution: https://github.com/helix-editor/helix/pull/7979

goyalyashpal:
> The logo.svg was a one liner while the logo-dark/light were multi line files.. I had run a formatter on all of them.

2. lol, i tried same here: https://github.com/helix-editor/helix/pull/5941 . apparently, 96bytes (`CLR\t` * 32 times) is a bit too much.

NewtonChutney:
_self hiding at offtopic_

> ----
> OT: 
> 
> 1. what window manager are you using which shows those "window tools" like move up, pin, size info, resize, etc above the window title bar? it looks useful as well as nice.
> 
> 2. similar attempt for icon/desktop files for another distribution: https://github.com/helix-editor/helix/pull/7979

1. Actually, those were screenshot files with names opened in QL-Win/QuickLook.. 😅
Awesome previewer..
The buttons are to not close the preview if focus is lost (pin), stay on top (up arrow)

2. Yes, I had gone through that, and even the instructions in the building from source page, where there are instructions on how to create a desktop shortcut in Linux if I'm not wrong
But Windows' shortcuts are different.. 😅
tui apps _can_ be launched by double clicking, and will launch automatically in conhost, or WT if that's set to the default..

goyalyashpal:
> 3\. So ppl don't waste their time in formatting the SVG files..

https://github.com/helix-editor/helix/pull/9106

seems i had shot 2 birds with one bow this time lol.

NewtonChutney:
_self hiding as offtopic_

> * https://github.com/helix-editor/helix/pull/9106
> 
> seems i had shot 2 birds with one bow this time lol.

Ooh nice!, also ask the members if they accept to move the files into another directory?

NewtonChutney:
By the logs I see the CI pipeline couldn't take the file..
Lemme try another take..
Or could we keep that file in the helix-term directory beside the build.rs script? This was what was suggested in the post

NewtonChutney:
The relative upper directory `../` doesn't work locally:
image

but `./` does work:
image

The attempt with subpath from the root path:
image

NewtonChutney:
No luck.. I'd need help fixing this!

pascalkuthe:
This is just how rust buildscripts work. Depending on how cargo is I.voked the cwd is different you cant make assumptions about that.


You will need to use the environment variables set by cargo

NewtonChutney:
With this line in `helix-term/build.rs`, 
```rust
    println!("cargo:warning=CWD: {:?}", std::env::current_dir());
```

The cwd is the same irrespective of where I launch the cargo build command from
image

image

So, why do the CI pipelines act differently?

NewtonChutney:
> You will need to use the environment variables set by cargo

Thanks for the tip @pascalkuthe
You inspired me to dig inside the winres crate to look for cargo vars

I hope this works.. @the-mikedavis one more test..? :-)
ᓚᘏᗢ

NewtonChutney:
I have pushed again after formatting, could you run the workflow again and do a final review? @the-mikedavis @pascalkuthe

pascalkuthe:
I am not quite comfortable committing an opaque binary to the repo. Could we build the .lib file by invoking rc from build.rs?

NewtonChutney:
Yes, but we'll have to do similarly to the winres crate.. Parse regedit for entries, and check if the binary exists..
Working on it

pascalkuthe:
> > I am not quite comfortable committing an opaque binary to the repo. Could we build the .lib file by invoking rc from build.rs?
> 
> Yes, but we'll have to do similarly to the winres crate.. Parse regedit for entries, and check if the binary exists.. Working on it

I think you can just use `cc` for that. The `cc` crat3 provides a function that finds the toolchain in the registry for you and we depend on that already

NewtonChutney:
So far, with the cc crate I've been able to get link.exe but not rc.exe
I'll have to check if there's another function..
image

NewtonChutney:
I'm having difficulty utilizing this function: get_sdk10_dir() which is in the source code of `cc`
https://github.com/rust-lang/cc-rs/blob/f17047de579adbe1c3a562b87cf9c0376a8e66cc/src/windows_registry.rs#L735

It appears to be part of this module.. But there's no way to instantiate/utilize it?
image

pascalkuthe:
yeah that is private API I assume that you could just use `windows_registry::find` directly. But that doesn't seem to be the case, unfortunately. In that case just copying the `get_sdk` function from `winres` is probably the best solution as `cc` doesn't expose the relevant internals.

NewtonChutney:
Yepp, I'll copy the minimum required from winres
To clarify, windows_registry::find isn't private, but it only returns VS installations and related tools.. 
The other get_10sdk is private

NewtonChutney:
@pascalkuthe, please do take a look now.. 😅

helix-term/build.rs
```diff
@@ -1,9 +1,143 @@
...
+
+        let resource_file = PathBuf::from(&output_dir).join("resource.lib");
+        compile_with_toolkit_msvc(rc_exe, resource_file, rc_path);
+
+        println!("cargo:rustc-link-search=native={}", output_dir.display());
+        println!("cargo:rustc-link-lib=dylib=resource");
+    }
+}
+
+fn compile_with_toolkit_msvc(rc_exe: PathBuf, output: PathBuf, input: PathBuf) {
```
pascalkuthe, line 34: all of these functions should be behind #[cfg(windows)] feature flags.

helix-term/build.rs
```diff
@@ -1,9 +1,143 @@
...
+        let rc_path = output_dir.join("resource.rc");
+        let icon_path = PathBuf::from("../contrib/helix-256p.ico");
+        write_resource_file(&rc_path, icon_path.to_str().unwrap()).unwrap();
+
+        let resource_file = PathBuf::from(&output_dir).join("resource.lib");
+        compile_with_toolkit_msvc(rc_exe, resource_file, rc_path);
+
+        println!("cargo:rustc-link-search=native={}", output_dir.display());
+        println!("cargo:rustc-link-lib=dylib=resource");
+    }
```
pascalkuthe, line 31: this should be moved to a sepreate cfg(windows) windows function (then you can define the function on cfg(not(windows)) to do nothing)

helix-term/build.rs
```diff
@@ -1,9 +1,143 @@
...
+        .unwrap();
+
+    if find_reg_key.status.code().unwrap() != 0 {
+        return Err(io::Error::new(
+            io::ErrorKind::Other,
+            "Can not find Windows SDK",
+        ));
+    }
+
+    let lines = String::from_utf8(find_reg_key.stdout).expect("Should be able to parse the output");
```
pascalkuthe, line 74: IMO Rfailing should never cause a build to fail so please don't panic here and instead return a result that you print in case this function fails

pascalkuthe (commented):
this looks mostly good to me the amount of code would be ok as long as you feature gate it so other platforms are not slowed down (see comments)

helix-term/build.rs
NewtonChutney: Would it be better off like this.. in a separate module with a single entry point and imports also conditionally imported?

NewtonChutney:
It appears the Cachix workflow is failing..
https://github.com/NewtonChutney/helix/actions/runs/7469536530/job/20326793896#step:5:1

pascalkuthe (approved):

helix-term/build.rs
NewtonChutney: Thanks!

the-mikedavis (approved):

PR #9106, opened 2023-12-18
Reduce logo.svg size (-93B)

goyalyashpal:
#### pretty logo chore(logo): PrettyPrint to reduce size by 93B ...
  * Current size:  2,755 B
  * Original size: 2,848 B
  * Add file's final newline (linux convention)
  * Remove tag separator spaces
  * Add newlines
  * Add tab indentation (instead of 2/4 spaces)
  * Prettify root svg's attribs

PR #9108, opened 2023-12-18
Reduce logos-with-text sizes (-323B -322B)

goyalyashpal:
#### chore(logo_dark): Pretty to reduce size by 323B ...
  * Current size:  6,930 B
  * Original size: 7,253 B
  * Add tab indentation (instead of 2/4 spaces)
  * Pretty Print by hand

#### chore(logo_light): Pretty to reduce size by 322B ...
  * Current size:  6,599 B
  * Original size: 6,922 B
  * Add tab indentation (instead of 2/4 spaces)
  * Pretty Print by hand

### DIFF
--- a/logo_dark.svg
+++ b/logo_dark.svg
@@ -1,115 +1,151 @@
 <?xml version="1.0" encoding="UTF-8" standalone="no"?>
 <svg
-   version="1.1"
-   xml:space="preserve"
-   style="clip-rule:evenodd;fill-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2"
-   viewBox="663.38 37.57 2087.006 903.71997"
-   id="svg22"
-   sodipodi:docname="logo_dark.svg"
-   width="2087.0059"
-   height="903.71997"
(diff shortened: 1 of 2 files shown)


the-mikedavis:
Same as #9106, it'd be fine to minify these by stripping all unnecessary whitespace, the trailing newline and maybe stripping the inkscape attributes but I don't see any value in pretty printing. These files are formatted right now only because that's how inkscape exported them

goyalyashpal:
_"minify"_

after 2-3 days.

_"strip ink's attribs"_

stripping ink's attribs would be a destructive operation. rendering these
equivalent to a useless build adtifact, unable to be modified later. all
the work put into this will be gone. are u _sure_ u want to do that?

goyalyashpal:
> Same as #9106, it'd be fine to minify these by stripping ...

also, the case with logo and logo-with-text is not same for minification - as here, there are 2 versions. the prettification makes it trivial to compare both the themes.



rest the decision is onto u.

the-mikedavis:
> \_"minify"\_ after 2-3 days

Yeah this is not exactly high priority. Let's leave the files as they are, they're just used for display purposes for the README.

PR #9109, opened 2023-12-18
(feat) Vim-like EOL Inlay Hints

Koranir:
Adds an option that makes inlay hints appear on the end of the line, as well as some configuration option for the prefix before the hints.

https://github.com/helix-editor/helix/assets/101683475/f0d97568-6ae3-44b7-847d-9b407ae838a2

Need to figure out a way to get rid of all the `: `s, which is why this PR is a draft.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1690,7 +1690,7 @@ fn compute_inlay_hints_for_view(
                         }
                         type_built_string.push_str(type_hint_suffix);
 
-                        InlineAnnotation::new(char_index, type_built_string)
+                        InlineAnnotation::new(0, type_built_string)
                     })
                     .collect()
             }
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1636,79 +1636,29 @@ fn compute_inlay_hints_for_view(
                 }
             }
 
-            fn collect_to_map<'a>(
-                doc_text: &helix_core::Rope,
-                hints: &'a Vec<InlineAnnotation>,
-            ) -> HashMap<usize, helix_core::SmallVec<[&'a str; 1]>> {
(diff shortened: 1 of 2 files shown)


kirawi:
See https://github.com/helix-editor/helix/discussions/7690#discussioncomment-6499408

Koranir:
> See [#7690 (comment)](https://github.com/helix-editor/helix/discussions/7690#discussioncomment-6499408) I'm pretty sure the colon is an example of what Pascal meant about EOL inlay hints not working well -- it's sent from rust-analyzer, not done in Helix.

Yeah. I'm thinking of having user-set regex patterns to match and keep only the wanted bits.

pascalkuthe:
This is not the right way to implement this feature. This not only doesn't properly handle mapping (if you join two lines the hints are now at the start of the line) it also unnecessarily moves the EOL character. #6417 adds functionality to allow you to truely render behind text (so behind the eol character) but since we now render diagnostics there it doesn't seem like a good idea to also render inlay hints there. I am not a fan of adding such an option to core. The standard is pretty explicit how and where inlay hints are rendered.

The only reason it was implememted like this in vim in the first place is because vim doesn't support displaying virtual text inline.

A hack like this is something I would relegate to plugins once that lands.

This just overly complicates the feature for questionable gain. If you don't like inlay hints turn them off and use hover to find out a type when you need it.

The only issues I ever had with inlay hints stem from the fact that our implementation is not as good as it could be (it's still a prototype feature) and inlay hints are a bit jumpy as a result. I am orking on improving yyat so they work more more like in vscode (where I never found them distracting)

Koranir:
I think having type data always visible on-screen is a valuable thing to have but having inlay hints in the middle of a line creates issues with code alignment and such, which is where having them EOL helps lot.

I'll keep this on the backburner, then, until plugins land.

PR #9120, opened 2023-12-19
Update upload/download-artifact actions to v4

the-mikedavis:


pascalkuthe (approved):

the-mikedavis:
Ah, the only problem with #9114 / #9115 was that v3 and v4 are incompatible so both upload and download need to be updated at the same time

PR #9122, opened 2023-12-19
Refactor statusline elements to build `Spans`

hamrik:
The current statusline implementation uses ad-hoc write callbacks to render statusline elements. This make statusline elements hard to work with, because we get no useful information about their output.

This PR refactors these elements to work more like widgets: each `render_*` function returns a `Spans`, which can be combined into a complete statusline. This has remained the responsibility of the `render` function.

This PR also separates the `render` function into two parts: `render_statusline` combines the aforementioned elements into one large `Spans` with given size constraints, while `render` blits it to a given `Surface` as it has done so previously. 

The rationale behind this is to allow post-processing the widget outputs and to make decisions based on them. 
For example, in the future we can now check if the statusline is empty and autohide it. Or revisit #7310 and automatically space the elements.

*This is merely a refactor; the current apparent behavior of each statusline element should be identical to previous versions.*

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -4,7 +4,6 @@ use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::{
     document::{Mode, SCRATCH_BUFFER_NAME},
     graphics::Rect,
-    theme::Style,
     Document, Editor, View,
 };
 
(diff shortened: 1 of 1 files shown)


archseer (approved):

the-mikedavis (dismissed):

the-mikedavis:
This just needs a rebase / master merge to resolve the conflicts

hamrik:
Rebase done!

the-mikedavis (approved):

pascalkuthe (approved):

PR #9143, opened 2023-12-22
Persistent state

intarga:
Resolves part of #401, along with #9154

TODO LIST:

- [x] persist command history
- [x] persist search history
- [x] persist file history
- [x] persist clipboard contents
~~- [ ] persist splits~~
~~- [ ] load in background tasks~~
- [x] command to manually reload history
- [x] handle errors nicely
- [x] trim history files
- [x] config options for persistence (on/off)
- [x] config options for persistence (limits)
- [x] config options for persistence (excluded files)
- [x] testing
- [x] documentation


kirawi:
Are you interested in working on buffer state, history, or both? I was going to get back to persistent undo this week, but if you intend to tackle it then I'll leave it to you.

kirawi:
However, if this is tackling persistent undo as well, then I believe @pascalkuthe wanted to avoid using serde.

intarga:
Hey @kirawi! This is not tackling undo, I looked at your PR and I don’t think there’s any overlap. I’m going to make a detailed post for discussion on the issue once I have a working protype, but I’ll write up some of my findings here now in case they’re helpful:

I looked into Neovim’s ShaDa (one single file in MessagePack format), following archseer’s suggestion. It’s used to persist most of what we want _except_ notably undo history, and there seem to be some good reasons for that. For one thing, undo histories get big very quickly, and if we want to save a significant amount of history for a significant number of files, they will bloat the shada file probably more than we want. Neovim instead seems to have one undo file per edited file, which seems like a better approach, though I couldn’t find any info on the details of the format for their undofiles.

From this though, I think it makes sense for persistent undo to be implemented separately from the rest.

pascalkuthe:
So the problem I have with serde is that it doesn't allow incremental or streaming serialization and deserialization. 

I think for undofile having that would be nice but not a hard requirement. (N)vim has a pretty simple file format and simply fully writes/reads the undofile whenever it writes/reads the file. 

I bieve we could do better by changing the format a bit so we only append whenever we save (and make it easy to evict old revsions). But I am not set on that. If we comeup with a scheme to instead evict in memory, keep the undofile fairly small writing on each save (just like vim) may be ok and then we could use serde/bincode (very fast and mature serialization format).

For undofiles the ultimate question is probably garbage collection. How do we keep the u dofile from growing forever. 

For session data/something like shada in vim I do like the idea of doing something similar vht here again append only serialization and streaming deserialization are important.


That wknt work well with serde. The pvject headers are simple e ought to write your own parser (no need for msgpack).


They are fully seldescibing and describe the size of some arbitrary data that follows the header. We can read that data and deserialize with serde (And don't need msgpack I prefer bincode).

intarga:
I have no particular attachment to serde or messagepack, so I'll happily switch to bincode.

> For session data/something like shada in vim I do like the idea of doing something similar vht here again append only serialization and streaming deserialization are important.

I'm not sure I agree about "append only". I see the benefit that it would make simple writes faster, but at the cost of flexibility and ease in reading and trimming. The design of shada in nvim seems to at least initially have been motivated by append-only writes, which is why it's a concat of msgpack objects instead of an array, but it seems like they ultimately decided not to follow through with this and instead they [merge](https://neovim.io/doc/user/starting.html#shada-merging) with the existing file when writing. While merging is presumably more expensive than appending, it has some notable advantages:

1. Entries from different contemporary sessions can be shown in chronological order, instead of the order their sessions quit.
2. Entry types where it only makes sense to have a single entry (clipboard contents and split layout, perhaps) can avoid duplication, and can choose the last entry chronologically.
3. Length limits for different entry types can be enforced separately (i.e. limit oldfiles to 500 entries, but command history to 1000).
4. Spamming entries of one type does not risk evicting all entries of another (i.e. if I send a ton of commands, I don't risk losing all my oldfiles).
5. Entries of the same type can stay clustered, which makes reading easier and more flexible.
6. There's a straight forward answer to when and how to trim the file. If we go append-only, it's not clear what to do. If we're enforcing a limit of X entries, then surely that means we have to trim every time we write. Unless I'm missing something, trimming will have to involve at least partly reading and deserialising the file so we can figure out what to remove. If we're reading the file every time we write, I'm not sure I see a meaningful benefit to append-only over merging.

Some, but not all, of these problems could be solved by having dedicated files for different entry types, but then you have to write a bunch of files instead of just one 🤷‍♀️

pascalkuthe:
I personally always thought that having multiple files would be the way to go. Especially for command history that should work well. That could work more or less the same as zsh history (it's 3xactly the same concept). I don't really think it's a problem having multiple datafiles.

I think a big advantage of appendonly is that you avoid frequently writing large amounts of data to disk. It's not a huge deal but reducing background io is nice. 

For other files like registers I agree that it doesn't make sense to have appendonly files. But I would just keep these entirely separate.

I never understood why nvim went with a single file model it seems to make everything (including the merging) much more complicated without much tangiböe benefit.

intarga:
Ok, I'll have a go at the multi-file approach then 👍

gyreas:
If you need an alternative name to ShaDa, can you consider Hexion (Helix session)?

It's tongue-in-check tho

helix-loader/src/session.rs
```diff
@@ -0,0 +1,49 @@
...
+
+    let mut file = OpenOptions::new()
+        .append(true)
+        .create(true)
+        .open(filepath)
+        // TODO: do something about this unwrap
+        .unwrap();
+
+    // TODO: do something about this unwrap
+    writeln!(file, "{}", line).unwrap();
```
the-mikedavis, line 23: Register contents will probably need to be encoded somehow, for example base64. You can copy line endings into the registers which will break reading below

helix-view/src/register.rs
the-mikedavis, line 1: Registers being reversed should be an implementation detail that doesn't escape this module. Instead of adding a function here let's reverse the collection in `helix_loader::session` in the read functions.

helix-loader/src/session.rs
intarga: Good point. You ok with bincode? Since that's what I'm using anyway for files and splits

helix-view/src/register.rs
intarga: Ok, just seemed wasteful to reverse twice... 

Would it be an option to change the arg in write to Iterator instead of Vec? Then we could avoid the wasted work while still keeping the reversal as an implementation detail

helix-loader/src/session.rs
```diff
@@ -0,0 +1,90 @@
...
+// TODO: should this contain a ViewPosition?
+// it would require exposing that type in a new crate, re-exporting in helix-view,
+// since this crate is a dependency of helix-view
+#[derive(Debug, Serialize, Deserialize)]
+pub struct FileHistoryEntry {
+    path: PathBuf,
+    anchor: usize,
+    vertical_offset: usize,
+    horizontal_offset: usize,
+}
```
intarga, line 19: @the-mikedavis Is there currently a place to put types shared between crates? If not, is it ok if I make a helix-types crate for this purpose?

helix-view/src/register.rs
the-mikedavis: Most places that need to write to a register need to collect first to get the length of the vec for display purposes so switching to taking an iterator will be wasteful for all of those callsites

helix-loader/src/session.rs
the-mikedavis: `_-types` crates can allow types to leak all over a project when they should be local to a specific crate. Here I think it's a good indicator that `helix-loader` is currently trying to do too much. The types should be moved into `helix-view` in a new `session` module. `helix-loader` can expose functions that give the file history file (and any other files) and then `helix-view` can read/write from those. `helix-loader` is already doing things 'above its paygrade' here by deciding what to do based on register characters: that decision should be made by a higher up crate like `helix-view`

helix-loader/src/session.rs
the-mikedavis: Yep I think bincode should be ok for this too

helix-loader/src/session.rs
intarga: > `helix-loader` is already doing things 'above its paygrade' here by deciding what to do based on register characters: that decision should be made by a higher up crate like `helix-view`

I'll make this change.

About the types, my thought was that abstractions are going to leak anyway, either `-loader` needs to know about `-view` and `-term` types, or `-view` and `-term` need to know about serialisation formats. IMO it's better to leak into the smaller crate, and better to leak types than functionality (especially if that type is already public, as is the case for ViewPosition, which seems to be used extensively in `-term`), but if that's not norm for helix I'll adapt. 

What's the more relevant guideline when something conceptually spans two crates in your opinion:
- it should go in the crate higher in the dependency graph
- it should go in the crate with the revelevant types, splitting if necessary

helix-view/src/register.rs
intarga: Understood

helix-loader/src/session.rs
the-mikedavis: I think it's fine for `helix-view` and `helix-term` to be aware of serialization. `helix-loader` could re-export or wrap the necessary parts of bincode if we want some separation. The closest comparable code we have is probably config (see `Config` in `helix-view/src/editor.rs`). For config we also need to read from files and (de)serialize types that have some editor specifics like the statusline config. For that, `helix-loader` is only providing the files while the higher level crates take care of (de)serialization.

Generally I would prefer to see things in the higher crate in the dependency graph. I don't want to see editor specifics like `ViewPosition` in low-level crates like `helix-loader`.

intarga:
@the-mikedavis When opening files with their saved position, should the view be aligned to centre?

Although I would personally prefer to not align, I noticed that helix aligns to centre even on buffer switches, so I'm assuming that's the convention, and in that case we only need to persist the selection, not the view. Should I stick with that? Should alignment be configurable?

gabydd:
I think the aligning when switching might actually be a bug, cause it can cause a lot of shifting when doing something like ``ga`` (going back and forth to the last accessed buffer)

intarga:
@gabydd I also don't like that behaviour, though I don't think it's a bug. The code for it looks pretty intentional, and vim notably seems to behave the same

gabydd:
Ah okay I'll try to take a better look when I have the time

intarga:
```
    fn replace_document_in_view(&mut self, current_view: ViewId, doc_id: DocumentId) {
        let view = self.tree.get_mut(current_view);
        view.doc = doc_id;
        view.offset = ViewPosition::default();

        doc.ensure_view_init(view.id);
        view.sync_changes(doc);
        doc.mark_as_focused();

        align_view(doc, view, Align::Center);
    }
```

I'm not completely sure, but I think this is the relevant function, and that alignment looks pretty intentional. 

If I get approval I would very much like to change this behaviour. Though I think to remove the shifting on buffer switched we might have to persist ViewPosition information in Document, since it currently only seems to have information about the selections.

the-mikedavis:
There was a PR about this that I think is not yet finished and could probably be picked up and brought across the finish line if you're interested: https://github.com/helix-editor/helix/pull/7414

I would prefer that we save the View's offset (ViewPosition) and use it if possible, centering only if the cursor wouldn't be visible.

intarga:
I would gladly take that on! It seems though that the work might already be done. I found this PR #7568 by the same author which seems to address the issues raised in that thread, and is waiting on review. The author mentions an unresolved issue, but looking at the code, I think that was just a misunderstanding.

helix-view/src/editor.rs
```diff
@@ -913,6 +924,18 @@ impl Default for Config {
             popup_border: PopupBorderConfig::None,
             indent_heuristic: IndentationHeuristic::default(),
             jump_label_alphabet: ('a'..='z').collect(),
+            persist_old_files: false,
+            persist_commands: false,
+            persist_search: false,
+            persist_clipboard: false,
+            // TODO: any more defaults we should add here?
+            persistence_file_exclusions: [r".*/\.git/.*"]
```
Ordoviz, line 932: Please add `r".*/COMMIT_EDITMSG"` here. When using bare git repos, the git dir is not always named `.git` (see https://news.ycombinator.com/item?id=11070797).

sibouras:
hi got a panic, hx doesn't work even when doing `hx -h`

```
thread 'main' panicked at helix-loader\src\lib.rs:171:41:
called `Option::unwrap()` on a `None` value
stack backtrace:
   0: std::panicking::begin_panic_handler
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library\std\src\panicking.rs:597
   1: core::panicking::panic_fmt
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library\core\src\panicking.rs:72
   2: core::panicking::panic
             at /rustc/79e9716c980570bfd1f666e3b16ac583f0168962/library\core\src\panicking.rs:127
   3: helix_loader::initialize_command_histfile
   4: tokio::runtime::park::CachedParkThread::block_on
   5: tokio::runtime::runtime::Runtime::block_on
   6: hx::setup_logging
```

intarga:
@sibouras Thanks for the report. This crash is due to a missing state dir, it seems that state dir does not exist on Windows.

I already planned to replace that unwrap with an error message, and to let users override the state dir with an environment variable, which would make this usable on Windows, but that alone doesn't seem like an ideal solution. I'm wondering if there's a convention or good default for a directory I could pick to use for storing state on Windows? I would welcome any input on this front, as I'm not very familiar with Windows.

Guekka:
> @sibouras Thanks for the report. This crash is due to a missing state directory, it seems that state directory does not exist on Windows.
> 
> I already planned to replace that unwrap with an error message, and to let users override the state dir with an environment variable, which would make this usable on Windows, but that alone doesn't seem like an ideal solution. I'm wondering if there's a convention or good default for a directory I could pick to use for storing state on Windows? I would welcome any input on this front, as I'm not very familiar with Windows. Perhaps a fallback to <cache dir>/helix/state if no state dir is found?

I usually refer to [this library](https://github.com/sago007/PlatformFolders) (or [a rust equivalent](https://github.com/cjbassi/platform-dirs-rs))

So in this case, I think it should be `C:\Users\%USERNAME%\AppData\Local\helix`

intarga:
Thanks for the tip! 

It seems like that convention picks the same directory we already use for cache, which in helix we treat as a distinct directory. So I'll go with `<cache dir>/helix/state` which should resolve to `~\AppData\Local\helix\state` to avoid any conflicts

intarga:
I've pushed the fix.

@sibouras let me know if it works for you

sibouras:
> I've pushed the fix.
> 
> @sibouras let me know if it works for you

yayy it works great!! thanks

agieocean:
I'm trying to use this I've compiled it without issue on Ubuntu 22.04 and it accepts these options:
```
persist-old-files = true                     
persist-commands = true
persist-search = true
```
But I don't understand what to do in the session to make it persist, is there a command I run or is it supposed to do it automatically and it's not for some reason? I have the reload-history command but I can't seem to find another one

intarga:
@agieocean It is intended to work automatically. What are you expecting to happen that doesn't?

agieocean:
> @agieocean It is intended to work automatically. What are you expecting to happen that doesn't?

My apologies I read back through the work you did and realized I misunderstood something, good work

davidrios:
Hello. Would it be possible to add a bit more info to the missing items in the to-do list? I mean if you need help with that, so that potentially interested people have an easier starting point. Thanks

olxandr:
It would be nice if the command history would match the text typed, for example:
when the user type ":cd" and press arrow up, only "cd*" commands are shown, not every single command from the command history.

Also it would be cool to have something like nvim sessions. When you can restore all of your buffers from the last session/previous sessions.

Thanks!

intarga:
@davidrios Are you offering to submit a patch?

intarga:
@olxandr 

> It would be nice if the command history would match the text typed

That sounds like a useful feature, but it's unrelated to this PR. command history actually already existed before, and I haven't changed the way it works, just persisted it between sessions. If there doesn't already exist an issue for this, you should make one. Better yet, if you're feeling brave, I would encourage you to have a go at implementing it 🙂 

> When you can restore all of your buffers from the last session/previous sessions

That was my intention with the "persist splits" task, but I decided not to implement it in this PR, because the PR is already quite big, and that feature is different enough from the rest I think it deserves its own treatment. I also think it would be better implemented by someone who actually uses splits and would use this feature, which I don't and wouldn't. I would be happy to offer my help/context/advice to anyone who wants to have a go at implementing this.

> And the list of the last opened files too. The recent files feature from nvim's telescope plugin.

The work I've done in this PR makes this pretty low-hanging fruit, but again, I think it deserves it's own PR. I might pick this one up myself after merging the current PR though

olxandr:
I didn't realize my feature requests was unrelated to this PR, sorry. Just wanted to share my thoughts about editor's persistent state so maybe in the future someone will implement these features. Thank you for your work.

davidrios:
@intarga Yeah pretty much, if you don't mind. Doesn't need to be anything detailed, just the gist, something like "testing: add 100% unit test coverage" and "handle errors nicely: remove all unwraps".

intarga:
@davidrios I would be glad to receive a patch!

With the errors you're mostly on the money; there are a lot of unwraps/panics that should be recoverable, and either raise an error message to the user, and/or put something in the log.

In terms of tests I'm not sure there's much value in unit tests for this (though it never hurts to have them), my intention was to write an integration test. In [this commit](https://github.com/helix-editor/helix/pull/9143/commits/ca16809d753bc034c60aa11298442bac1f524863) I fixed the existing integration tests to work with persistence using tempfiles. This was useful to catch panics, but it does not actually verify the correctness of the persisted state. Ideally we should add an integration test that opens and closes some files, sends some commands, does some searches, ends the session, then starts a new session and checks the state is all loaded correctly.

intarga:
@davidrios I had some time and energy to write the integration test, so that no longer needs doing

intarga:
@davidrios Aaaaand I fixed the error handling too 😅 (though I actually decided to keep pretty much all the unwraps)

intarga:
This is finally ready for review 🎉 

A couple of notes for the reviewers:
- This PR tackles persisting file locations (view offsets & selections), command history, search history, and clipboard, but not undo trees or splits.
- I considered loading the persisted state in background tasks, as its on the critical path for startup time, but ultimately decided it's a premature optimisation, since it doesn't have a noticeable impact on startup times for me, and anyway it only affects people who enable the feature. If you disagree, I'm happy to change this.

neutralinsomniac:
I've been using this for about a month now with no issues. *huge* help to my development. Thank you and I hope this gets merged!

commandline-be:
> I've been using this for about a month now with no issues. _huge_ help to my development. Thank you and I hope this gets merged!

if you mean the feature where closing and opening the same file does bring back the screen to the last known cursor position, please let us know how to use this feature

intarga:
@commandline-be add this to your config and it should work automatically:

```
[editor.persistence]
old-files = true
```

commandline-be:
> @commandline-be if you’re running this branch, add this to your config and it should work automatically:
> 
> ```
> [editor.persistence]
> old-files = true
> ```

thanks for sharing, did not work since "Bad config: unknown field `persistence`"
guess i'll have to build Helix myself or wait for an update ?

intarga:
@commandline-be Of course you will need to build it, this PR is not merged

useche:
I have a question about this patch. I usually have many helix instances to work on different things. In this patch, I noticed that the state of all these helix instances would be funneled to the same file. My question is: how does `serialize_into` handles simultaneous writes to the same file? Will it handle that case correctly? My guess is that it won't but I couldn't find the answer in a couple of minutes search. If so, how are we handling this case?

helix-view/src/editor.rs
```diff
@@ -1033,6 +1072,8 @@ pub struct Editor {
     pub debugger_events: SelectAll<UnboundedReceiverStream<dap::Payload>>,
     pub breakpoints: HashMap<PathBuf, Vec<Breakpoint>>,
 
+    pub old_file_locs: HashMap<PathBuf, (ViewPosition, Selection)>,
```
useche, line 1075: If I understand correctly, this `HashMap` will store the information of all the files opened by any instance of Helix in the system. Isn't that too much? Shouldn't a particular instance care only about its current working directory?

intarga:
> I have a question about this patch. I usually have many helix instances to work on different things. In this patch, I noticed that the state of all these helix instances would be funneled to the same file. My question is: how does `serialize_into` handles simultaneous writes to the same file? Will it handle that case correctly? My guess is that it won't but I couldn't find the answer in a couple of minutes search. If so, how are we handling this case?

The entries are independent of each other, and the OS ensures that only one append operation happens at once. So if 2 instances try to write at the same time, the OS will block one until the other is done, and it doesn't really matter which one blocks.

helix-view/src/editor.rs
intarga: Too much in terms of what? Memory footprint? With the default settings running for a few months I just checked my `file_history` file and it's 12KiB, when deserialised to a hash map it should be about the same size in RAM. That seems pretty negligible.

Personally I find this useful, as I do sometimes open files outside the working directory, and I don't see that there's really a tradeoff for it.

useche:
> > I have a question about this patch. I usually have many helix instances to work on different things. In this patch, I noticed that the state of all these helix instances would be funneled to the same file. My question is: how does `serialize_into` handles simultaneous writes to the same file? Will it handle that case correctly? My guess is that it won't but I couldn't find the answer in a couple of minutes search. If so, how are we handling this case?
> 
> The entries are independent of each other, and the OS ensures that only one append operation happens at once. So if 2 instances try to write at the same time, the OS will block one until the other is done, and it doesn't really matter which one blocks.

Thanks for the answer. I didn't know that. From the `OpenOptions::append` documentation:
```
Append mode guarantees that writes will be positioned at the current end of file, even when there are other processes or threads appending to the same file.
```

helix-view/src/editor.rs
useche: Thanks for your answer. Yes, I meant memory footprint.

And it makes sense for it to be that low. After reading your answer I did some back of the envelop calculation and assuming a 64 bytes entry per file, you'll need 16K files in the history to crate a 1MiB footprint. Since we are pruning that history from time to time (IIUC), then it's unlikely we get to that extreme anyway.

useche:
Another question: why isn't there a `clipboard-trim` option?

intarga:
> Another question: why isn't there a `clipboard-trim` option?

There isn’t any clipboard history, so your clipboard (and the persisted file) only contain 1 entry.

useche:
> > Another question: why isn't there a `clipboard-trim` option?
> 
> There isn’t any clipboard history, so your clipboard (and the persisted file) only contain 1 entry.

That means that this entry will be replaced back and forth between difference instances of Helix, right? Then, when you open and close multiple instances of Helix in different working directories for different projects, a Helix instance will end up loading an unrelated clipboard entry.

From my point of view, ideally, you would load the exact same state you had in that working directory undisturbed by instances in unrelated working directories. This is much easier to reason and you will end up with a more predictable workspace.

To solve this problem, we can simply store the state file in the working directory `.helix`. What do you think? This would also address my question earlier where all the old file states of all helix instances are mixed.

intarga:
@useche if you want to implement this, I'll accept a patch.

It's a good idea, especially since with your extra comment 1 in mind, you should be able to implement it without affecting users that don't have this kind of workspace-style workflow.

useche:
> @useche if you want to implement this, I'll accept a patch.
> 
> It's a good idea, especially since with your extra comment 1 in mind, you should be able to implement it without affecting users that don't have this kind of workspace-style workflow.
> 
> About 2, my original thought was to persist splits only through a particular command and flag, I decided not to implement that in this PR because I don't use splits, so I didn't feel I can really understand the needs of those users and design something that fits them well.

Sure, I could give this a try. However, I think that, in this scenario, the default option (all state files in a single directory) will be awkward for split persist. If someone enable split persist, and we open the last split at startup (something I want), people will find a weird split with weird files from the last entry that was created in the file. Having the states on a per project basis would be way easier [https://github.com/useche/helix/commit/97178e8fac3317fadd9a864f4b4c5e19b047e434].

What do you think? From your point of view, what's the advantage of having a single persist file? It would be good to know what other people think.

intarga:
> will be awkward for split persist

Split persistence should definitely be off by default. A note in the config docs like so:

> you probably only want to enable this in `.helix` for specific projects along with specifying `persistence.peristent_state_directory` as `.helix`

Should suffice I think.

> From your point of view, what's the advantage of having a single persist file? 

Many people don't have a workspace based workflow, but even if they do:
- They might often want to open files outside their workspace, like files from a project that is a dependency of their current one, or config files in `~/.config` or `/etc`. In these cases they probably want history relevant to those files that didn't come from their current workspace.
- When starting a new workspace, they might like to have their command and search history from other projects rather than starting from scratch.
- They might just not be bothered to configure their workspace "properly", or not want a .helix directory in it.

I do see the utility of what you're proposing, but I personally wouldn't use it.

It's also worth noting that if storing state in `<workspace_dir>/.helix` was the default, you'd have to decide how to handle what to do if the detected workspace dir doesn't contain a `.helix` dir. You'd have to either:
1. Not offer persistence in these cases
2. Create `.helix` directories in places the user probably doesn't expect and may not want

Neither seem like good options. From a cursory look, the commit you linked seems to do 2. 

Consider what would happen if a user has an immutable root filesystem, is hanging out in `/etc` and opens helix. In your commit it looks like it would try to create `/etc.helix` and crash. Even if the root filesystem isn't immutable, the user probably doesn't want you creating `/etc/.helix`.

> It would be good to know what other people think.

In particular the maintainers if/when they get around to reviewing this, since it's ultimately their call to make.

useche:
@intarga thanks for the discussion.

Thinking more about this problem and the discussion I think I reached a pretty good solution. The idea is two fold:
1. Add an option that allows to configure persistence to save it "all-in-one" (the default option) or "per-workspace".
2. Instead of creating a per-workspace `.helix` directory, we can save per-workspace state in the `.local/state` directory. Example:
```
/home/luis/.local/state/helix/
└── home
    └── luis
        ├── code
        │   └── helix
        │       ├── command_history
        │       └── file_history
        └── .config
            └── home-manager
                ├── command_history
                └── file_history
```

I implemented this [https://github.com/useche/helix/commit/99fdec230363ae1ea68d929b17cd5e172bd61251]. I may have gone overboard with the refactoring, but I did need some changes to use the config to decide what directory to use. If I went too far, let me know. The documentation is still missing, and it fails to compile for mac-os, but the main idea is there.

intarga:
@useche I like this design!

Some notes:

> Add an option that allows to configure persistence to save it "all-in-one" (the default option) or "per-workspace".

Rather than adding an extra config option, perhaps it would be simpler to modify the existing `persistence.old_files`, `persistence.commands` etc. options to be enums with "global", "workspace", and "off" (default) variants? This would make it more configurable (the user can have commands per workspace, but files global for example) with fewer options, and also avoids some meaningless config states like having `persistence.scope` set while all the bool options are false. What do you think?

> Example:

Consider flattening the directory structure, to avoid replicating the whole systems file hierarchy, and to make it easier for users to find and manage the state files if they wish. Modified version of your example:

```
/home/luis/.local/state/helix/
└── home%luis%code%helix
    ├── command_history
    └── file_history
└── home%Luis%.config%home-manager
    ├── command_history
    └── file_history
```

> I may have gone overboard with the refactoring

From a cursory look, is there any particular reason you ditched the `once_cell`s in `helix-loader`? I'll do a more thorough review when you have it passing CI

> I tried to create a pull request against your repository

Most convenient option for me would be if you email me a patch.

useche:
> @useche I like this design!
> 
> Some notes:
> 
> > Add an option that allows to configure persistence to save it "all-in-one" (the default option) or "per-workspace".
> 
> Rather than adding an extra config option, perhaps it would be simpler to modify the existing `persistence.old_files`, `persistence.commands` etc. options to be enums with "global", "workspace", and "off" (default) variants? This would make it more configurable (the user can have commands per workspace, but files global for example) with fewer options, and also avoids some meaningless config states like having `persistence.scope` set while all the bool options are false. What do you think?

I think your idea makes sense. I was also thinking that we have way too many options, but I was thinking on something like:
```
persistence.enable_on = [ "commands", "old_files", "etc" ]
```

What's nice about your idea is that it has pretty good flexibility. I wonder if we could do something like:
```
persistence.all = "global"
```

So that we don't have to repeat the same option everywhere. I'll give this some thought.

> 
> > Example:
> 
> Consider flattening the directory structure, to avoid replicating the whole systems file hierarchy, and to make it easier for users to find and manage the state files if they wish. Modified version of your example:
> 
> ```
> /home/luis/.local/state/helix/
> └── home%luis%code%helix
>     ├── command_history
>     └── file_history
> └── home%Luis%.config%home-manager
>     ├── command_history
>     └── file_history
> ```
> 

This sounds good. I'll make the change.

> > I may have gone overboard with the refactoring
> 
> From a cursory look, is there any particular reason you ditched the `once_cell`s in `helix-loader`? I'll do a more thorough review when you have it passing CI

I was having an issue (which I don't remember) with the fact that they were initialized so early. In any case, even now some reasons come to mind: (1) I'm not a fan of global variables and the persistent state location is something that only the persistent code cares about, (2) it's nice to have related code close by, (3) the persistent state location might now change with `:reload-config` and I didn't feel that comfortable changing the global variables at this point. I see that they should be synchronized, though, so that technically shouldn't be a problem. Let me know your thoughts.

> 
> > I tried to create a pull request against your repository
> 
> Most convenient option for me would be if you email me a patch.

I will once it's ready.

BTW, I also wrote a few patches over the weekend that add persistent splits. They are in my main branch in `useche/helix`. I'm not sure whether we should eventually mix those with this pull request or rather wait until this is checked in and then create a new pull request with those. Let me know if you have an opinion about that.

Axlefublr:
thank you very much for your work! have been using this PR for a while, and it works perfectly

some (very) recent commits in core introduce conflicts, though, because of which helix doesn't build with your PR included

I would very much appreciate if you rebased onto master to fix those conflicts :3

no rush, of course; just wanted to let you know as I'm sure I'm not the only one who's using helix with your PR included!

useche:
Continuing our conversation, I implemented a new way to set the persistence options. I wanted to simplify it based on the conversations we had. Now my configuration looks like:

```
[editor.persistence]
all.enabled = true
all.max-entries = 1000
all.scope = "per-workspace"
autostart-splits = true
```

The option `all` will be the default for all the specific options. Specific options (like "search" or "commands") can be changed for more specific options. The patch is in the top of my github tree useche/helix. Let me know what you think.

intarga:
@useche Apologies for the late response.

> I was having an issue (which I don't remember) with the fact that they were initialized so early. In any case, even now some reasons come to mind: (1) I'm not a fan of global variables and the persistent state location is something that only the persistent code cares about, (2) it's nice to have related code close by, (3) the persistent state location might now change with :reload-config and I didn't feel that comfortable changing the global variables at this point. I see that they should be synchronized, though, so that technically shouldn't be a problem. Let me know your thoughts.

My opposition to this was only that I think we should try to stick to existing conventions where possible. (3) seems like a good reason to break the convention though.

>  I'm not sure whether we should eventually mix those with this pull request or rather wait until this is checked in and then create a new pull request with those. Let me know if you have an opinion about that.

I think it would be best to have split persistence in a follow on PR to make things easier for reviewers, as this one is already quite big.

>Continuing our conversation, I implemented a new way to set the persistence options. I wanted to simplify it based on the conversations we had. Now my configuration looks like:
>```
>[editor.persistence]
>all.enabled = true
>all.max-entries = 1000
>all.scope = "per-workspace"
>autostart-splits = true
>```
>The option all will be the default for all the specific options. Specific options (like "search" or "commands") can be changed for more specific options. The patch is in the top of my github tree useche/helix. Let me know what you think.

I'm not sure I see the point in `all.max-entries` since the content of the entries is so different. I also don't think `all.enabled` and `all.scope` should be separate; I would prefer `all = "global"`|`all = "workspace"`|`all = "off"`

On another note, it would be good to have integration testing of per-workspace persistence

intarga:
@Axlefublr It's rebased on master now

Axlefublr:
@intarga thank you so much! and what wonderful timing, I was just in the process of git magicking my fork to add a new feature while being on an older commit; now I can just rebase on master like normal :D

ThanHenderson:
Thanks a lot for this. Is there a timeline to merge here?

intarga:
> Thanks a lot for this. Is there a timeline to merge here?

It's up to the maintainers now to review it. I have no idea where this fits on their priority list, I imagine they have a lot on their plates.

Axlefublr:
can you please rebase on master? there are quite a few merge conflicts, that I couldn't figure out how to handle myself

gabydd:
I have it merged with master here https://github.com/gabydd/helix/commit/eca36b7fa7b8c9049f5a17629ab28a77e455cb62

lucazffz:
Is this going to be merged any time soon?

NemesisRE:
> @DawnMagnet @lucazffz Do you really think it’s constructive to add one more comment complaining on a thread the maintainers clearly aren’t paying attention to?
> 
> If you need it that badly you can build the branch like the rest of us. I’m tempted to close this PR because I don’t want to be pinged by these spam comments

Can you lock the conversation or can only the maintainer/collaborators do this?

intarga:
> Can you lock the conversation or can only the maintainer/collaborators do this?

That option is only available to collaborators 🙁

intarga:
Willing to reopen and rebase if/when there's a maintainer who wants to review and merge this

linsomniac:
FYI: I've had Claude implement save/restore position off master at https://github.com/linsomniac/helix-resume

It does not implement any of the other functionality discussed here, just position saving, as that is a deal-breaker for me for using helix.

iByteABit256:
Leaving [this comment](https://github.com/helix-editor/helix/issues/401#issuecomment-3582758515) here, in case you are willing and have the time to transform this into a plugin instead.

The plugin system still hasn't been merged as far as I see but it's probably going to get a much higher priority than this issue.

PR #9154, opened 2023-12-23
Persistent Undo

kirawi:
Resolves persistent undo as part of #401 

TODO


kirawi:
It has been a while since I touched this branch, so I'll be working out where I was over the next few days. As far as I recall, these were some questions I wanted to answer:

- Should (de)serialization be implemented in the same file as all these types, or should they be defined in a different module (but that would mean setting the scope of many fields to `crate`-level)?
- Should there be a dedicated `State` struct to abstract undo files?

kirawi:
I am rethinking utilizing `blake3` as the hasher. We don't need it to be cryptographically secure and `sha256`'s speed has worked well enough for Vim. The issue arises in that I intend to use the hash as a checksum to verify the undofile's integrity. If, as I suspect, these undofiles grow big enough, then the hashing will take noticably more time. This could end up not being a problem at all in practice or there may be better solutions (such as pruning old branches).

https://lore.kernel.org/git/20180609224913.GC38834@genre.crustytoothpaste.net/ compared various algorithms for `git` (Blake3 did not exist yet) on a single-threaded workload. Torvalds said, "The performance difference between different hashing functions is likely not all that noticeable in most common cases as long as we're not talking orders of magnitude," which is more or less what I think would be the case here too.

Thoughts on using blake3 or reverting to sha1?

cessen:
Regarding hashes and persistent undo:  I agree that the undo file getting corrupted is unlikely, and probably not worth considering.

However, as you've noted in the PR description, there *is* a high probability of the *text file* getting modified outside of Helix (checking out git branches, running auto formatters, making quick edits in other editors, etc.).  So I think storing the hash for the text contents that the head of the undo file corresponds to does make sense, since time stamps can be fickle on some systems.

And with that in mind, I think using a faster hash still makes sense, since you'll be hashing text file contents on load to verify compatibility with the on-disk persistent undo.  IIRC blake3 is notably faster than any of the SHA hashes, even single-threaded.

You might also consider tagging the hash value with what hash it is (Blake2, SHA1, whatever) so that it's easier to change it in the future if desired.

kirawi:
Yeah it already was using a hash (sha1), but I've switched it to xxhash since it seems more geared towards this kind of thing (e.g. `rsync`). I don't think it's necessary to store the hasher since we already have a `VERSION` tag in the header. I also got rid of mtime now since it's not really useful (and Vim seemed to be using it for its history rather than any validation).

David-Else:
I had a quick look at the code and can't find the config options, I assume I can turn this feature off?

kirawi:
Ah, I forgot to document it. It's `editor.undofile` and it's off by default.

helix-core/src/history.rs
```diff
@@ -630,4 +875,151 @@ mod test {
...
+        use rand::SeedableRng;
+
+        let dist = Uniform::new_inclusive(0, 2);
+        let mut rng = rand::rngs::SmallRng::from_entropy();
+
+        let mut hist = History::default();
+        let mut doc = Rope::default();
+        let sel = Selection::point(0);
+
+        while !inserts.is_empty() {
```
kirawi, line 933: Self-note: This is a bad test implementation because it commits a new revision more often than it performs some other action.

fastfading:
can we proritize this ?    a new version released .  still no persistent undo 
many people use helix to replace simple vim tool on server  .   
not many people will use this as a serious ide.  we don't need any other fancy feature. 
we just need the basic function like this feature.

kirawi:
I'm not currently working on major features for Helix for at least a few months. For the past few years, all my contributions have been to Helix, and right now I'm eager to focus on my own projects.

EricHenry:
@kirawi want some help finishing this one up?

kirawi:
I restarted work on this on a private branch already, and I'll try to get that over the finish line. If I'm not available for it, then I'll make that branch public and allow someone to adopt the PR.

kirawi:
Just a minor status update: We'll be using an SQLite DB to host the persisted file histories instead of a bespoke file format. The underlying architecture will have to be significantly changed since revisions won't be serialized in the same way. Otherwise, I have addressed some bugs with the architecture.

kirawi:
I'm fairly busy right now, so if anyone is interested and has the experience to tackle this, please message me on Matrix. Otherwise, I will continue to work on it as I am available.

Smithx10:
Anyone rebase this on master lately?

daedroza:
I have rebased the patch but compilation fails due to 
   Compiling helix-view v25.1.1 (/home/kalpaj/base/helix/helix-view)
error[E0061]: this method takes 2 arguments but 1 argument was supplied
    --> helix-view/src/document.rs:1354:26
     |
1354 |                     self.set_last_saved_revision(last_saved_revision);
     |                          ^^^^^^^^^^^^^^^^^^^^^^^--------------------- an argument of type `SystemTime` is missing
     |
note: method defined here
    --> helix-view/src/document.rs:1872:12
     |
1872 |     pub fn set_last_saved_revision(&mut self, rev: usize, save_time: SystemTime) {
     |            ^^^^^^^^^^^^^^^^^^^^^^^            ----------  ---------------------
help: provide the argument
     |
1354 |                     self.set_last_saved_revision(last_saved_revision, /* SystemTime */);
     |                                                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

For more information about this error, try `rustc --explain E0061`.

daedroza:
Passed SystemTime::now() to that function, it compiles...

fastfading:
is this issue finished ? 

I saw the home pages says that persistent undo feature is supported .  But it does not work actually
https://github.com/helix-editor/helix/discussions/14321

con-f-use:
That "home page" is a fake. It's not supported, yet.

eliliam:
> I saw the home pages says that persistent undo feature is supported

@fastfading see https://github.com/helix-editor/helix/discussions/14321#discussioncomment-14231086:
> That's not the official website. The official website is helix-editor.com

hwamil:
damn this feature seemed so basic when I was using nvim.
lacking this hits hard when I make a mistake or want to try different settings on stuff. shame.

kirawi:
Closing as stale.

Smithx10:
What is going to happen instead?  Are we not going to get this in helix but have it implemented via plugin?

Would be nice if there was a table that stated what should be implemented via a plugin vs in the main bin.

kirawi:
It's still planned for core AFAIK.

PR #9173, opened 2023-12-26
make sure to sync views when applying edits to unfocused views

### LINKED ISSUE #6752: 'Position ... is out of range for changeset len ....!' panic on ~23.03 when saving a file

gabydd:
this fixes bugs when applying workspace edits with lsp to a view where the most recent selection range is outside the current document or when adding a newline/formatting during write all in the same situation

fixes https://github.com/helix-editor/helix/discussions/9157 and fixes #6752

I decided to sync the view when getting the id instead of calling sync when applying the transaction. I am not sure if I put the utility function in the correct place, I did have to take the tree instead of the editor because rust was giving me error for passing the editor within the closure in write-all, which I wasnt able to figure out

(I think we might be able to use this function in reload-all as well, though it already properly handles this)

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -886,7 +886,6 @@ pub fn apply_workspace_edit(
             }
         };
 
-        let current_view_id = view!(editor).id;
         let doc_id = match editor.open(&path, Action::Load) {
             Ok(doc_id) => doc_id,
             Err(err) => {
@@ -897,7 +896,7 @@ pub fn apply_workspace_edit(
             }
         };
 
-        let doc = doc_mut!(editor, &doc_id);
+        let doc = doc!(editor, &doc_id);
         if let Some(version) = version {
             if version != doc.version() {
                 let err = format!("outdated workspace edit for {path:?}");
@@ -908,18 +907,8 @@ pub fn apply_workspace_edit(
         }
 
         // Need to determine a view for apply/append_changes_to_history
-        let selections = doc.selections();
-        let view_id = if selections.contains_key(&current_view_id) {
-            // use current if possible
-            current_view_id
-        } else {
-            // Hack: we take the first available view_id
-            selections
-                .keys()
-                .next()
-                .copied()
-                .expect("No view_id available")
-        };
+        let view_id = editor.get_synced_view_id(doc_id);
+        let doc = doc_mut!(editor, &doc_id);
 
         let transaction = helix_lsp::util::generate_transaction_from_edits(
             doc.text(),
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -675,13 +675,15 @@ pub fn write_all_impl(
     let mut errors: Vec<&'static str> = Vec::new();
     let config = cx.editor.config();
     let jobs = &mut cx.jobs;
-    let current_view = view!(cx.editor);
-
     let saves: Vec<_> = cx
         .editor
         .documents
-        .values_mut()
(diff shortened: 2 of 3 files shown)


gabydd:
I can also confirm it fixes a bug similar to the one in the discussion but with renaming.where you open a document (a) then create a split and go to near the end of the file there, then change to a different document(b) that has a function you can rename that is referenced in file (a), then go to file (a) in the other split find the reference to the function you are going to rename and delete from the line after there to the end of the file, switch back to the other split with file (b) and rename the function, before this pr helix crashes there

helix-view/src/document.rs
```diff
@@ -1141,6 +1142,29 @@ impl Document {
         self.inlay_hints.remove(&view_id);
     }
 
+    /// Returns the id of a view that this doc contains a selection for,
+    /// making sure it is synced with the current changes
+    /// if possible or there are no selections returns current_view
+    /// otherwise uses an arbitrary view
+    pub fn get_synced_view_id(&mut self, current_view: ViewId, tree: &mut Tree) -> ViewId {
```
pascalkuthe, line 1149: I think this should be a function on editor that takes a `DocumentId` as argument.

helix-view/src/document.rs
```diff
@@ -1141,6 +1142,29 @@ impl Document {
...
+        if self.selections.contains_key(&current_view) {
+            let view = tree.get_mut(current_view);
+            // only need to sync current view if this is not the current doc
+            if view.doc != self.id {
```
pascalkuthe, line 1153: This is not necessary.  The current view is always sycned. We don't sync when switching buffer either, only when switching Views.

gabydd: The reason for this being here is the document you might be fetching a view id for might not be the document but might contain the current view in its selection, maybe we should only take this path if we are getting for the current document I'll see when I move it too Editor

pascalkuthe: Hmm I am not sure I understand. The currently selected view (whoch should be current_view) never needs to be synchronized. You only need to synchronize views that are not visible. This applys to all documwnts no matter whether they are currently shown in the view.

gabydd: coming back to this the reason why this is needed is if 
1. you have two views ``a`` and ``b``
2. you open the same document (``z``) in them
3. then in ``b`` open a different document (``x``) (this will be view.doc)
4. then go to ``a`` and make a change to ``z`` that won't be synced with b
5. so if you change to ``b`` and write-all, when looking for a view for ''z'' , ``b`` the current_view will work and need to be synced because the change was done in another view and we never synced that change when switching to this view
I hope that makes sense

gabydd:
I moved it to editor, needed to change some things to appease the borrow checker stuff

helix-view/src/document.rs
pascalkuthe: indeed you are right I was misremembering the details

pascalkuthe (approved):

PR #9183, opened 2023-12-28
added local search functionality (like global search only in the current document)

Kim-Dewelski:
This code adds the equivalent of global-search but only within a single buffer. It's useful because using the global-search functionality as a ctrl + f equivalent for a single document is a bit tedious and something like this will make the processes of finding instances of something within a single document a lot easier.

### DIFF
--- a/book/src/keymap.md
+++ b/book/src/keymap.md
@@ -295,6 +295,7 @@ This layer is a kludge of mappings, mostly pickers.
 | `Y`     | Yank main selection to clipboard                                        | `yank_main_selection_to_clipboard`         |
 | `R`     | Replace selections by clipboard contents                                | `replace_selections_with_clipboard`        |
 | `/`     | Global search in workspace folder                                       | `global_search`                            |
+| '&'     | Local search in buffer contents                                         | `local_search`                             |
 | `?`     | Open command palette                                                    | `command_palette`                          |
 
 > 💡 Global search displays results in a fuzzy picker, use `Space + '` to bring it back up after opening a file.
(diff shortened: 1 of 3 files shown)


pascalkuthe:
We already have `/` for that I don't see why this is needed

Kim-Dewelski:
because using '/' is really tedious, I need to write / multiple times as it will only step forward to the next instance of the match. This will open a menu (or picker) similar to global-search which gives a much better overview of the current matches in the document, imo.

Kim-Dewelski:
The workaround solution is using global search then just filtering after the file you care about but it's not exactly the best solution compared to just making this feature

pascalkuthe:
You can jump to the next and previous search Position with `n` and `p`. I don't see additional value in using a picker here. The picker is not useful here. The picker primary purposefully is fuzzy filtering but since the filepath is the same for all entrys so filtering is not useful here.

Kim-Dewelski:
I guess I'm coming from the angle I like to not jump around constantly and like the preview I can get with the picker. My questions would be what the issue is merging this feature? If I do use the n/p thing will that add onto the jump list for every time I press them? I like this feature because with the preview I can see where I go before I commit to jump my cursor to that position.

Kim-Dewelski:
The jump list is relevant because I use ctrl-o quite a lot

kirawi:
They don't add it to the jumplist.

Kim-Dewelski:
@pascalkuthe I personally feel it would be useful to merge this but if it doesn't follow the philosophy of Helix then perhaps we should close this?

pascalkuthe:
I don't think we will merge this. As we said `/` and `n`/`N` cover this. Including the jumplist behavior.

PR #9198, opened 2024-01-01
update remapping.md: clarify difference in modifiers between non-character and character bindings

DrymarchonShaun:
Documenting the difference between using `shift` as a modifier for a special key and a character key. See #8831 and #9197


the-mikedavis:
Let's fix this rather than document it as you said in https://github.com/helix-editor/helix/discussions/8831#discussioncomment-7989450

DrymarchonShaun:
That's fine by me. Actually implementing it was beyond my skill set, the best I could do personally was document it.

Should I close this then?

the-mikedavis:
Yep let's close this in favor of #9213

PR #9214, opened 2024-01-02
Add .prettierrc as json file type

jpaju:


the-mikedavis (approved):

pascalkuthe (approved):

leungbk:
FYI, `.prettierrc` is not always a JSON file, and can be a YAML file: https://prettier.io/docs/en/configuration.html

> You can configure Prettier via (in order of precedence):
>
>    A `"prettier"` key in your `package.json` file.
>    A `.prettierrc` file written in JSON or YAML.

PR #9232, opened 2024-01-04
Add auto-pairs to scheme language support

gamma-delta:
Currently, typing a single quote in a `.scm` file "helpfully" auto- completes a closing quote. This is because there is no auto-pairs section in the languages.toml. This commit adds that.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1914,6 +1914,12 @@ shebangs = ["scheme", "guile", "chicken"]
 comment-token = ";"
 indent = { tab-width = 2, unit = "  " }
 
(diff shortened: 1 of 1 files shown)


dead10ck (approved):
This seems fine to me, though someone with more experience in the lisps might want to check too.

zetashift:
Steel user here, LGTM as well! Thank you for this 🎉

the-mikedavis (approved):

gamma-delta:
Wooh! I use Helix pretty much every day so it's nice that I could contribute. \\[^.^]/

PR #9081, opened 2023-12-14
Add initial support for janet-lang

pnelson:
Resolves https://github.com/helix-editor/helix/issues/4617 https://github.com/helix-editor/helix/issues/4574

PR #9247, opened 2024-01-05
Use janet-simple grammar for Janet

sogaiu:
Following up on [this comment](https://github.com/helix-editor/helix/pull/9081#issuecomment-1878091306), this PR is mostly about using the [tree-sitter-janet-simple grammar](https://github.com/sogaiu/tree-sitter-janet-simple) for Janet support along with a syntax highligting query file.

It also adds some bits for `janet-format` -- a formatter for Janet that comes with the [spork library](https://github.com/janet-lang/spork/).

---

I am aware of 2 grammars [1] for Janet and each took a slightly different approach from the other.  The one that's part of this PR is used in nvim-treesitter, difftastic, and there is code to use it from Emacs as well.

Due to historical reasons, this one is named "tree-sitter-janet-simple" (internally `janet_simple` -- to avoid name collisions) and IIUC that means the value for `name` in `languages.toml` needs to be `janet-simple` (under both `[[language]]` and `[[grammar]]`).  Or, at least, that's what made it work for me.  Please point out if that should be changed and if so how :)

[LSP support for Janet is still maturing](https://github.com/CFiggers/janet-lsp) and although I did get it to work with some degree of success, I didn't feel the time was right yet.

---

[1] [Here](https://github.com/GrayJack/tree-sitter-janet) is the other one for reference.


languages.toml
```diff
@@ -324,6 +324,28 @@ comment-token = "//"
...
+
+[language.auto-pairs]
+'"' = '"'
+'(' = ')'
+'[' = ']'
+'{' = '}'
+"`" = "`"
+
+[[grammar]]
+name = "janet-simple"
```
the-mikedavis, line 346: The name of the `[[grammar]]` section needs to match the name defined in `grammar.js` (replacing hyphens with underscores) but the language definition doesn't need to be called `janet-simple` too:

```toml
[[language]]
name = "janet"
# ..
grammar = "janet-simple"

[[grammar]] # this part is unchanged
name = "janet-simple"
# ..
```

then the queries need to move to `runtime/queries/janet/*.scm`

sogaiu: Thanks, I think I understand how to change things.

sogaiu:
Does squashing to a single commit and force-pushing sound ok?

book/src/generated/lang-support.md
```diff
@@ -75,7 +75,7 @@
 | idris |  |  |  | `idris2-lsp` |
 | iex | ✓ |  |  |  |
 | ini | ✓ |  |  |  |
-| janet | ✓ |  |  |  |
+| janet-simple | ✓ |  |  |  |
```
archseer, line 78: I think this now needs to be `janet`?

archseer:
No need to manually squash, we'll squash when merging

book/src/generated/lang-support.md
sogaiu: Ah good point.

sogaiu:
I don't understand [this error](https://github.com/helix-editor/helix/actions/runs/7455204292/job/20306191494?pr=9247#step:5:193):

```
 Error: "Failed to parse highlights.scm queries for janet: Query error at 5:2. Invalid node type long_str_lit"
Error: Process completed with exit code 1.
```

I guess the line in question is [this one](https://github.com/helix-editor/helix/pull/9247/files#diff-08a19aa699c008219045e25572824d4b38e5606181ec9cf4b799ed891d2f523dR5).

I think [`long_str_lit` is a valid node in the grammar](https://github.com/sogaiu/tree-sitter-janet-simple/blob/51271e260346878e1a1aa6c506ce6a797b7c25e2/grammar.js#L62).

I don't see any error via `:log-open` when I view a `.janet` file -- would an error show up there if there were a problem with `highlights.scm`?

Any hints as to what I might be missing?

the-mikedavis (commented):
You need to remove the original `[[language]]` block for janet. It's overriding `grammar` to use the `clojure` one which is causing the queries to fail (since they aren't valid for the clojure grammar)

the-mikedavis:
You should see the same error printed to the log when opening a janet file. You need to run helix via `cargo run` to build these changes though - they won't affect any instances of Helix you installed from a package manager or from source

sogaiu:
Thanks for the hints.

I guess a duplicate section got added back [here](https://github.com/helix-editor/helix/pull/9247/commits/794b122bb5e8d7a5bca3679d182586c1183152f1#diff-f3ea26d68561619ec2cb9ee551ef03916b667bd5f55b2ae2c54cae96e8821b2eR3031-R3038).

I guess my bad for putting the new section elsewhere and [removing the old one before the merge](https://github.com/helix-editor/helix/pull/9247/commits/2be96f7bfd0996e33f0e1ac5f16902fc8477e650#diff-f3ea26d68561619ec2cb9ee551ef03916b667bd5f55b2ae2c54cae96e8821b2eL3008-L3015) (^^;  Sorry about that.

runtime/queries/janet/highlights.scm
```diff
@@ -1 +1,72 @@
...
+
+[(bool_lit) (nil_lit)] @constant.builtin
+
+(comment) @comment
+
+(quote_lit
+ (sym_lit) @symbol)
+
+(qq_lit
+ (sym_lit) @symbol)
```
the-mikedavis, line 21:
```suggestion
 (sym_lit) @string.special.symbol)

(qq_lit
 (sym_lit) @string.special.symbol)
```

sogaiu: I guess what's being suggested is that `symbol` doesn't exist so use something else?

Assuming so, that makes sense to me, though `string.special.symbol` is used for `kwd_lit` (and according the [syntax highlighting section](https://github.com/helix-editor/helix/blob/84e24b33dcda16d1d64805f34dcc02d82d0de8f1/book/src/themes.md#syntax-highlighting) something similar is Clojure keywords) and a distinction might be nice.

May be just dropping highlighting for `quote_lit` and `qq_lit` for now is another path.  Possibly there isn't much value in highlighting quoted symbols in these types of situations.

the-mikedavis (approved):

the-mikedavis:
(The CI failures are unrelated - it's caused by a sourcehut outage)

PR #9248, opened 2024-01-05
Fix minor typo: anntoation -> annotation

jerabaul29:
Just a minor typo fix I found reading the source :) . Fixed using the LSP ("space" + "r" to replace symbol in the whole project).


pascalkuthe (commented):
We generally don't accept typo fixes for non-commemts since that creates a lot of merge conflicts. Particularly this code has a large PR open where I already have a huge amount of conflicts

Issue #5190, opened 2022-12-17
Configure what character is used to draw the ruler


aspizu:
Instead of drawing a filled column, optionally
draw a column with the configured character.
This then could be set to a box-drawing character
to get a line as a ruler.

archseer:
It's deliberately not a character so it doesn't obscure the text though

aspizu:
this would be optional though

dgkf:
I just wanted to chime in to say that I see the value here. I've been using TUI-based text editors for years and I still second-guess whether my ruler markings are the last column of my preferred width or the first column of overflow.

Using a box-drawing character helps make this less ambiguous.

But I also agree with @archseer that this loses some information, for example in cases where almost all lines overflow a ruler.

One alternative I had in mind would be to permit ranges in the ruler config. If 80 is a preferred linewidth, then this makes it more obvious that 81 is the start of a region of overflow, not the end of acceptable widths. The example below would use the ruler style for all columns >=81. 

```toml
[editor]
rulers = [ 81.. ]
```

Granted, this isn't valid `toml` so the syntax would need to be workshopped.

What do you think @aspizu?

dylangiles:
Howdy all, I have opened PR #9256 that adds this functionality using the config syntax @aspizu suggested. It is opt-in - if the config key is not provided the original behaviour for rulers is used

Issue #5721, opened 2023-01-29
Not all themes display rulers


non-descriptive:
### Summary

themes which do not display rulers

- [ ] default
- [ ] kanagawa
- [ ] rasmus
- [ ] solarized_light
- [ ] solarized_dark

Basically theme doesn't have proper bg color for `ui.virtual.ruler` or  fg color for `ui.virtual.rulers`. Usually they should have either `gray` color or the same color as `ui.cursorline.primary`

### Reproduction Steps

1. setup [`editor.rulers`](https://docs.helix-editor.com/configuration.html#editor-section) value.
2. set theme to one in the list above

### Platform

Windows

### Terminal Emulator

Windows Terminal

### Helix Version

helix 22.12 (4d548a0e)

the-mikedavis:
Except for kanagawa, these all use foreground rulers. The default and rasmus set `ui.virtual` to a foreground color which makes sense for some scopes like `ui.virtual.indent-guide` or `ui.virtual.whitespace`. I think those are probably oversights and could be updated.

Solarized light and dark set `ui.virtual.ruler` to a foreground color explicitly which I think is a stylistic choice: foreground rulers are visible if you have lines that cross them. Foreground rulers can be useful because you can see when lines are too long without the ruler constantly being visible.

Kanagawa has a background ruler though. It looks to me like it's displaying a ruler.

non-descriptive:
I fixed some themes locally and didn't test for others, just noticed there's no rulers were displayed. Alsoit's kinda weird way to separate rulers - themes differentiate `ruler` and `rulers` keywords. One of it works with background and one woth foreground color.

pieqq:
I think at the very least, the default theme should display both foreground and background rulers. I wanted to display rulers the way I was used to in vim (what is called background rulers in Helix apparently), so I added

```
[editor]
rulers = [80]
```

in my `config.toml`, and yet nothing showed up. Sure, if lines go beyond 80 characters, the character placed at column 80 will be in a different color, but:

- this does not work if the character is a space
- this only shows for lines that go through the ruler limit

Only this morning did I figure this out when trying other themes! :)

(by the way, the live preview when running `:theme <tab><tab>` is amazing!!)

ardrigh:
I just noticed this on the default theme in Helix. I had set a ruler for 100 characters, and when I was typing the character changed colour, but the vertical bar isn't visible.

There is no indication the ruler has been set correctly based on the configuration.

[...]

PR #9256, opened 2024-01-06
feat: Use custom character to draw editor rulers

dylangiles:
This pull request adds functionality requested in #5190.

The user may specify a character that is used when rendering rulers, instead of colouring a line in the editor. The default behaviour is retained - if the user does not specify the ruler character config, the original behaviour is used. This functionality will be useful for users that want to, for example, use a box-drawing character to draw the ruler.

Please feel free to make suggestions and comments on my pull request, as it is my first time contributing to Helix. Additionally, I would like to thank the Helix project for making an awesome piece of software that has helped with my daily tasks.

## Example config:
```
[editor]
ruler = [70]
ruler-char = '│'
```

## Screenshot:
Screen Shot 2024-01-06 at 12 12 08 pm

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -37,36 +37,37 @@ Its settings will be merged with the configuration directory `config.toml` and t
 
 ### `[editor]` Section
 
-| Key | Description | Default |
-|--|--|---------|
-| `scrolloff` | Number of lines of padding around the edge of the screen when scrolling | `5` |
-| `mouse` | Enable mouse mode | `true` |
-| `middle-click-paste` | Middle click paste support | `true` |
-| `scroll-lines` | Number of lines to scroll per scroll wheel step | `3` |
-| `shell` | Shell to use when running external commands | Unix: `["sh", "-c"]`<br/>Windows: `["cmd", "/C"]` |
-| `line-number` | Line number display: `absolute` simply shows each line's number, while `relative` shows the distance from the current line. When unfocused or in insert mode, `relative` will still show absolute line numbers | `absolute` |
-| `cursorline` | Highlight all lines with a cursor | `false` |
-| `cursorcolumn` | Highlight all columns with a cursor | `false` |
-| `gutters` | Gutters to display: Available are `diagnostics` and `diff` and `line-numbers` and `spacer`, note that `diagnostics` also includes other features like breakpoints, 1-width padding will be inserted if gutters is non-empty | `["diagnostics", "spacer", "line-numbers", "spacer", "diff"]` |
-| `auto-completion` | Enable automatic pop up of auto-completion | `true` |
-| `auto-format` | Enable automatic formatting on save | `true` |
-| `auto-save` | Enable automatic saving on the focus moving away from Helix. Requires [focus event support](https://github.com/helix-editor/helix/wiki/Terminal-Support) from your terminal | `false` |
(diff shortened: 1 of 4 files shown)


archseer (approved):

pascalkuthe:
hmm I am not sure about this one. This will draw over any rendered text (so if you write to the ruler the text there is hidden). A similar issues already occurs with rulers hiding cursors.

IMO the right solution is to use the new infrastructure in #6417 and render the rulers before the text and the ruler highlights after the text highlights but before the selection highlights

helix-view/src/editor.rs
```diff
@@ -273,6 +273,11 @@ pub struct Config {
     pub terminal: Option<TerminalConfig>,
     /// Column numbers at which to draw the rulers. Defaults to `[]`, meaning no rulers.
     pub rulers: Vec<u16>,
+
+    /// If set, use this character to draw the editor's rulers.
+    #[serde(rename = "ruler-char")]
```
the-mikedavis, line 278: This is unnecessary because the struct is tagged with `#[serde(rename_all = "kebab-case")]`

helix-core/src/syntax.rs
```diff
@@ -149,7 +149,6 @@ pub struct LanguageConfiguration {
     /// global setting.
     #[serde(default, skip_serializing, deserialize_with = "deserialize_auto_pairs")]
     pub auto_pairs: Option<AutoPairs>,
-
```
the-mikedavis, line 152: This should be reset

book/src/configuration.md
the-mikedavis, line 1: No need to reformat the tables, let's leave them as-is. Reformatting them inflates the diff and may introduce conflicts with other PRs that touch these docs

the-mikedavis (commented):
I'm also unsure of this but for a different reason. This will introduce some issues with the theming for rulers, sort of opposite of the problem we have now in https://github.com/helix-editor/helix/issues/5721. Themes should define rulers with background colors rather than foreground colors. If you configure a `ruler-char` with this change though you want the ruler's foreground to be themed rather than the background though. So themes will either look wrong if you have `ruler-char` set or if you don't.

codingjerk:
Can I continue working on this PR? I want to use character rulers myself a lot.

@the-mikedavis, what are the problems right now?

1. Styling issues
  - Reformatted table @ `book/src/configuration.md`
  - Unnecessary removal of newline @ `helix-core/src/syntax.rs`
  - Unnecessary `rename` annotation @ `helix-view/src/editor.rs`
2. Problem with themes: background vs foreground color
3. Drawing ruler under the text

Anything else?

codingjerk:
I played with code a bit and I have some points here, except styling related, since they are trivial:

1. For themes, I think it will be better to introduce another theme in addition to `ui.virtual.ruler`, like `ui.virtual.ruler.char` or `ui.virtual.ruler-char` when `editor.ruler-char` is set.
2. To draw ruler under text I simply move `render_rulers` call before `render_document`. This change will be required if we will use `DecorationManager` for rulers in future anyway. Cause `render_document` takes ownership on it.

For now it looks great on my setup, I would like to send PR.

image

I'm not completely happy with rendering code and introducing two separate theme keys tho. So I would appreciate any help.

codingjerk:
I created a PR https://github.com/helix-editor/helix/pull/11798 and I suggest moving further work there

PR #9278, opened 2024-01-08
Add support for `matlab-language-server`

### LINKED ISSUE #9276: Use `matlab-language-server` for MATLAB code
MathWorks released an open-source language server for MATLAB early last year, which should probably be part of the default language settings for MATLAB.

Link: [https://github.com/mathworks/MATLAB-language-server](https://github.com/mathworks/MATLAB-language-server)

colin-heffernan:
This is currently **UNTESTED**, as I am currently away from my Linux tower, and I cannot find an easy way to install `matlab-language-server` on Windows. I can (and will) test this when I get back to my tower, but I'm opening the PR now in case others would like to test.

Closes #9276.

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -101,7 +101,7 @@
 | markdoc | ✓ |  |  | `markdoc-ls` |
 | markdown | ✓ |  |  | `marksman` |
 | markdown.inline | ✓ |  |  |  |
-| matlab | ✓ | ✓ | ✓ |  |
+| matlab | ✓ | ✓ | ✓ | `matlab-language-server` |
 | mermaid | ✓ |  |  |  |
 | meson | ✓ |  | ✓ |  |
 | mint |  |  |  | `mint` |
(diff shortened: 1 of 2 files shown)


colin-heffernan:
Forgot to generate docs, one moment.

colin-heffernan:
Getting issues, the server is exiting on startup. It may be due to the fact that I do not have MATLAB installed, but I would think the language server would be independent. However, I cannot get MATLAB, so I believe that I am unable to test this myself. Any help would be appreciated.

gabydd:
this should work fine(got a similar config working here: https://github.com/helix-editor/helix/discussions/8956#discussioncomment-7724549) just need to put a note in the install wiki that people have to configure the matlab install path. I also am not sure if we should disable indexing, I would be ok with disabling telemetry by default but indexing seems quite useful see https://github.com/mathworks/MATLAB-extension-for-vscode configuration section

colin-heffernan:
Honestly, since one would need to configure the install path for MATLAB to use it, would this best be left as a non-default configuration?

the-mikedavis:
Yeah if it takes some machine-specific configuration then let's add it to the language server wiki rather than `languages.toml`: https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

colin-heffernan:
Added to the wiki, though a double-check to ensure that my formatting is correct would be appreciated. In the meantime, I'll close the issue and PR.

PR #9296, opened 2024-01-10
languages: add CEL, SpiceDB schema language

jzelinskie:
CEL (Common Expression Language) is a non-Turing-Complete language created at Google for embedding into programs. It's used in a variety of projects such as Kubernetes and SpiceDB.

SpiceDB is a popular, open source authorization system inspired by the system internal at Google, Zanzibar.

This PR adds language/grammar support for both of these projects.

I'm not sure the CEL highlighting is working -- I might need some help with writing custom queries for it. The SpiceDB grammar includes queries that target Helix.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -504,6 +504,18 @@ name = "cel"
 source = { git = "https://github.com/bufbuild/tree-sitter-cel", rev = "9f2b65da14c216df53933748e489db0f11121464" }
 
 [[language]]
+name = "spicedb"
+scope = "source.zed"
+injection-regex = "spicedb"
+file-types = ["zed"]
+comment-token = "//"
+indent = { tab-width = 2, unit = "  " }
+
+[[grammar]]
+name = "spicedb"
+source = { git = "https://github.com/jzelinskie/tree-sitter-spicedb", rev = "04ea9c294b3e20f9e3433114fa138cc20dc92770" }
+
+[[language]]
 name = "go"
 scope = "source.go"
 injection-regex = "go"
--- a/languages.toml
+++ b/languages.toml
@@ -513,7 +513,7 @@ indent = { tab-width = 2, unit = "  " }
 
 [[grammar]]
 name = "spicedb"
-source = { git = "https://github.com/jzelinskie/tree-sitter-spicedb", rev = "04ea9c294b3e20f9e3433114fa138cc20dc92770" }
+source = { git = "https://github.com/jzelinskie/tree-sitter-spicedb", rev = "a4e4645651f86d6684c15dfa9931b7841dc52a66" }
 
 [[language]]
 name = "go"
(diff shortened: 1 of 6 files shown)


jzelinskie:
I seem to be getting the following error when I run `cargo xtask docgen`:

```error[E0658]: use of unstable library feature 'stdsimd'```

**Comment by the-mikedavis** [comment]
Are you running on aarch64? That may be https://github.com/helix-editor/helix/pull/9294. You can apply the diff suggested in the Docs CI or revert the dep locally temporarily to fix the build

File: languages.toml [hunk 1]
```diff
@@ -489,6 +489,34 @@ args = { processId = "{0}" }
 name = "c-sharp"
 source = { git = "https://github.com/tree-sitter/tree-sitter-c-sharp", rev = "5b60f99545fea00a33bbfae5be956f684c4c69e2" }
 
+[[language]]
+name = "cel"
+scope = "source.cel"
+injection-regex = "cel"
+file-types = ["cel"]
+comment-token = "//"
+roots = []
```
the-mikedavis, line 498: The empty `roots` keys can be removed since they're defaulted now (https://github.com/helix-editor/helix/pull/8803)

languages.toml
```diff
@@ -489,6 +489,34 @@ args = { processId = "{0}" }
...
+indent = { tab-width = 2, unit = "  " }
+
+[[grammar]]
+name = "cel"
+source.git = "https://github.com/bufbuild/tree-sitter-cel"
+source.rev = "9f2b65da14c216df53933748e489db0f11121464"
```
the-mikedavis, line 504:
```suggestion
source = { git = "https://github.com/bufbuild/tree-sitter-cel", rev = "9f2b65da14c216df53933748e489db0f11121464" }
```

and same below, just to match the style we use for the rest of languages.toml

the-mikedavis (commented):
You'll need to add the queries to `runtime/queries/cel/highlights.scm` and `runtime/queries/spicedb/highlights.scm`. You can use the queries in the repos if they exist (`queries/highlights.scm`) as guides or copy them if they already use the Helix captures (https://docs.helix-editor.com/master/themes.html#syntax-highlighting) - we don't automatically use the ones in the repos.

jzelinskie:
Thanks for the feedback!

I am using an M1, so rebasing on master fixed my issue.

I copied over the queries. I'd love a sanity check from someone better at tree-sitter because I'm not sure CEL is being properly injected into the `(caveat_expr)` .

jzelinskie:
Thanks very much for the feedback! I've commit your changes.

What's a good way to test the injecting of CEL into the SpiceDB schema?

the-mikedavis:
You'll have to test the injection manually by finding a `caveat_expr` node in a SpiceDB file and checking its highlighting

jzelinskie:
Thanks a ton!
I'm not sure how I didn't discover that command before, but it was vital for debugging.

I've got it all working now!

pascalkuthe (approved):

PR #9304, opened 2024-01-10
Add Sourcepawn language support

Sarrus1:
This PR adds a tree-sitter parser and queries for the [Sourcepawn programming language](https://wiki.alliedmods.net/Introduction_to_SourcePawn_1.7).

### DIFF
--- a/runtime/queries/sourcepawn/highlights.scm
+++ b/runtime/queries/sourcepawn/highlights.scm
@@ -1,0 +1,308 @@
+(identifier) @variable
+; Assume all-caps names are constants
+((identifier) @constant
+  (#lua-match? @constant "^[A-Z][A-Z0-9_]+$"))
+
+; Function definitions/declarations
+(function_definition
+  name: (identifier) @function)
+
+(function_declaration
+  name: (identifier) @function)
+
+(parameter_declaration
+  name: (identifier) @variable.parameter)
+
+; Methods / Properties
+(field_access
+  field: (identifier) @variable.member)
+
+; Function calls
+(call_expression
+  function: (identifier) @function)
+
+(call_expression
+  function:
+    (field_access
+      field: (identifier) @function.method.call)) ; Must be after field_access
+
+; Types
+(builtin_type) @type.builtin
+
+(type
+  (identifier) @type)
+
+(any_type) @type
+
+; Variables
+(variable_storage_class) @keyword.storage
+
+(variable_declaration
+  name: (identifier) @variable)
+
+(old_variable_declaration
+  name: (identifier) @variable)
+
+; Preprocessor
+(preproc_include) @keyword.import
+
+(preproc_tryinclude) @keyword.import
+
+(system_lib_string) @string
+
+(string_literal) @string
+
+(preproc_assert) @keyword.directive
+
+(preproc_pragma) @keyword.directive
+
+(preproc_arg) @constant
+
+(preproc_macro) @function.macro
+
+(macro_param) @variable.parameter
+
+(preproc_if) @keyword.directive
+
+(preproc_else) @keyword.directive
+
+(preproc_elseif) @keyword.directive
+
+(preproc_endif) @keyword.directive
+
+(preproc_endinput) @keyword.directive
+
+(preproc_define) @keyword.directive.define
+
+(preproc_define
+  name: (identifier) @constant)
+
+(preproc_undefine) @keyword.directive.define
+
+(preproc_undefine
+  name: (identifier) @constant)
+
+(preproc_error) @function.macro
+
+(preproc_warning) @function.macro
+
+; Expressions
+(view_as) @function.builtin
+
+(sizeof_expression) @function.macro
+
+(this) @variable.builtin
+
+; https://github.com/alliedmodders/sourcemod/blob/5c0ae11a4619e9cba93478683c7737253ea93ba6/plugins/include/handles.inc#L78
+(hardcoded_symbol) @variable.builtin
+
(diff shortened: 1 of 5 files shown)


runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+(parameter_declaration
+  defaultValue: (identifier) @constant)
+
+(fixed_dimension) @punctuation.bracket
+
+(dimension) @punctuation.bracket
+
+(array_indexed_access) @punctuation.bracket
+
+(escape_sequence) @string.escape
```
the-mikedavis, line 112:
```suggestion
(escape_sequence) @constant.character.escape
```

the captures should line up with https://docs.helix-editor.com/master/themes.html#syntax-highlighting

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+(methodmap_property_setter) @function.method
+
+; Enum structs
+(enum_struct) @type.definition
+
+(enum_struct
+  name: (identifier) @type)
+
+(enum_struct_field
+  name: (identifier) @variable.member)
```
the-mikedavis, line 151:
```suggestion
  name: (identifier) @variable.other.member)
```

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+
+(enum_entry
+  value: (_) @constant)
+
+; Literals
+(int_literal) @number
+
+(char_literal) @character
+
+(float_literal) @number.float
```
the-mikedavis, line 253:
```suggestion
(int_literal) @constant.numeric.integer

(char_literal) @constant.character

(float_literal) @constant.numeric.float
```

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+
+(string_literal) @string
+
+(array_literal) @punctuation.bracket
+
+(concatenated_string) @punctuation.delimiter
+
+(bool_literal) @boolean
```
the-mikedavis, line 261:
```suggestion
(bool_literal) @constant.builtin.boolean
```

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+  "default"
+  "switch"
+] @keyword.conditional
+[
+  "do"
+  "while"
+  "for"
+  "continue"
+  "break"
+] @keyword.repeat
```
the-mikedavis, line 283:
```suggestion
"return" @keyword.control.return
[
  "if"
  "else"
  "case"
  "default"
  "switch"
] @keyword.control.conditional
[
  "do"
  "while"
  "for"
  "continue"
  "break"
] @keyword.control.repeat
```

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+  "void"
+] @keyword
+
+[
+  "const"
+  "native"
+  "static"
+  "stock"
+  "forward"
+] @type.qualifier
```
the-mikedavis, line 308: The trailing newline is missing on a few files: `highlights.scm`, `injections.scm` and `languages.toml`

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
+(identifier) @variable
+; Assume all-caps names are constants
+((identifier) @constant
+  (#lua-match? @constant "^[A-Z][A-Z0-9_]+$"))
```
the-mikedavis, line 4:
```suggestion
  (#match? @constant "^[A-Z][A-Z0-9_]+$"))
```

`#lua-match?` is specific to neovim

the-mikedavis, line 4: These two matches on `(identifier)` should be at the bottom of the file rather than the top. In neovim the precedence of the queries is the opposite of what it is in Helix (and tree-sitter-cli): in Helix the higher patterns in the file take precedence over the lower ones. So as-written these `(identifier)` matches will match before any of the patterns below that also match `(identifier)` like on L7 or L10. You will also want to swap the order of these two matches since the one with the `#match?` is more specific (so it should be higher precedence) than the one on L1.

runtime/queries/sourcepawn/highlights.scm
```diff
@@ -0,0 +1,308 @@
...
+(variable_declaration
+  name: (identifier) @variable)
+
+(old_variable_declaration
+  name: (identifier) @variable)
+
+; Preprocessor
+(preproc_include) @keyword.import
+
+(preproc_tryinclude) @keyword.import
```
the-mikedavis, line 49:
```suggestion
(preproc_include) @keyword.control.import

(preproc_tryinclude) @keyword.control.import
```

the-mikedavis (changes requested):

Sarrus1:
Hello, thank you very much for the feedback.

I should have tested this PR before submitting it, I naively thought that helix queries were fully compatible with the new nvim-tree-sitter queries. Sorry about that.

I will convert it to draft while I fix it.

Is there a way to "easily" test queries?

the-mikedavis:
We don't have test cases but you can `cargo run` in the Helix directory on an example file to check your work

archseer:
Replaced by https://github.com/helix-editor/helix/pull/13028

PR #9306, opened 2024-01-10
Fix export instructions in installation doc

jerabaul29:
I have tried to install helix from source, including following the instructions at: https://docs.helix-editor.com/install.html#configuring-helixs-runtime-files . I am on Ubuntu 22.04, and I tried to install the latest master branch.

I hit 2 issues:

- 1: if I understand well, in bash, in order to export an env variable, the ```export``` keyword is needed; otherwise, the variable is not exported to child processes, and helix does not seem to receive the env variable (inspecting the ```hx --health``` output without the ```export``` keyword, I cannot see the added path from ```HELIX_RUNTIME``` listed).

- 2: it seems that helix is not able to expand ```~```. I.e., if I use in my ```.bashrc```: ```export HELIX_RUNTIME="~/src/helix/runtime"```, I then get a not recognized path:

```sh
kzm@kzm-bpq:~$ hx --health | head -8
Config file: /home/kzm/.config/helix/config.toml
Language file: default
Log file: /home/kzm/.cache/helix/helix.log
Runtime directories: /home/kzm/.config/helix/runtime;~/src/helix/runtime;/home/kzm/.cargo/bin/runtime
Runtime directory does not exist: ~/src/helix/runtime
Runtime directory is empty: /home/kzm/.cargo/bin/runtime
Clipboard provider: wl-paste+wl-copy
System clipboard provider: wl-paste+wl-copy
```

while if I use in my ```.bashrc```: ```export HELIX_RUNTIME="${HOME}"/src/helix/runtime```, then the path is well recognized:

```
kzm@kzm-bpq:~$ hx --health | head -8
Config file: /home/kzm/.config/helix/config.toml
Language file: default
Log file: /home/kzm/.cache/helix/helix.log
Runtime directories: /home/kzm/.config/helix/runtime;/home/kzm/src/helix/runtime;/home/kzm/.cargo/bin/runtime
Runtime directory is empty: /home/kzm/.cargo/bin/runtime
Clipboard provider: wl-paste+wl-copy
System clipboard provider: wl-paste+wl-copy
```

This PR updates these 2 points.

### DIFF
--- a/book/src/install.md
+++ b/book/src/install.md
@@ -216,12 +216,12 @@ RUSTFLAGS="-C target-feature=-crt-static"
 
 #### Linux and macOS
 
-The **runtime** directory is one below the Helix source, so either set a
+The **runtime** directory is one below the Helix source, so either export a
 `HELIX_RUNTIME` environment variable to point to that directory and add it to
 your `~/.bashrc` or equivalent:
 
 ```sh
-HELIX_RUNTIME=~/src/helix/runtime
+export HELIX_RUNTIME=~/src/helix/runtime
 ```
 
 Or, create a symbolic link:


book/src/install.md
```diff
@@ -218,10 +218,11 @@ RUSTFLAGS="-C target-feature=-crt-static"
 
 The **runtime** directory is one below the Helix source, so either set a
 `HELIX_RUNTIME` environment variable to point to that directory and add it to
-your `~/.bashrc` or equivalent:
+your `~/.bashrc` or equivalent (you need to expand the full path, using ```~```
+in this configuration is not parsed by helix):
```
the-mikedavis, line 222: As far as I know this is usually handled by your shell:

```bash
$ FOO=~/src/helix
$ echo $FOO
/home/michael/src/helix
```

Maybe when you were doing this you were quoting the value? That will stop the shell from expanding the path:

```bash
$ FOO="~/src/helix"
$ echo $FOO
~/src/helix
```

the-mikedavis (commented):
Adding `export` I believe is right but I don't think we need to change the `~/` part

book/src/install.md
the-mikedavis: I think it would be ok to pass the value of `HELIX_RUNTIME` through `expand_tilde`: as you say it looks pretty awkward when `--health` says that `~/src/helix/runtime` doesn't exist

the-mikedavis: Ah yeah, we want to move the path utilities in helix-core into a lower "stdx" crate that has extensions to the standard library. Let's merge this with just the change to add `export` and I'll add a follow-up issue to expand the tilde once we create the stdx crate (which I'm working on a PR for now)

the-mikedavis (approved):

PR #9311, opened 2024-01-11
Add  command to open a popup as a new buffer

### LINKED ISSUE #8748: Provide ways to open pop ups in a new buffer / split
Pop ups are very convenient for small items (a few lines), however I found them a bit inconvenient for large items (for example, a long manual page for a function in numpy or scipy).

Some examples in my workflow when pop ups are not so convenient:

1. When using pop ups to get access to long documentation that I have to read through in details, I end up spending quite a lot of time reading and moving from inside the pop up itself to read it, which is not very convenient as the possibilities for moving around are much more limited than on a normal buffer: I would much better enjoy to move around in the pop up with the full power of helix commands.

2. When displaying the documentation for a slightly advanced function through a pop up, I may want to be able to look both, at the same time, at my code and the pop up documentation, as I may need to think a bit, compare my use with the different documentation examples, etc. I may also want to copy paste some lines from the pop up examples to my code and to tweak it in my code. This is not achievable as far as I know with the pop ups as these are now.

So after a bit of use, it feels like the pop ups are not too adapted to display more than a few lines of information: when more information is displayed, such as a long manpage with a lot of explanations and examples, I would really prefer to have the possibility to get the pop up content into a (temporary) buffer for moving around / copying examples to my code, etc.

Do you think / any idea how a good UX could be for that? For example, when inside a pop up, should we add a new pop up mode, with a few "popup mode commands" that would allow to, for example:

- open the pop up in a new temporary buffer displayed as a full screen buffer ("b" key in pop up mode?)
- open the pop up in a new temporary buffer displayed as a split (with different shortcuts to open the split on the top, bottom, right, left of the current buffer / split, for example by using the "hjkl" keys in pop up mode?).

Could something like this make sense? I do not think something like this is available at the moment, or am I missing something? :)

jerabaul29:
closes #8748 

The idea is to make it possible to press ```ctrl-o``` to open the popup content in a new buffer. Once this work, I can easily add ```ctrl-h``` and ```ctrl-v``` to open in horizontal and vertical splits :) .

This is still an early WIP; what remains to be done:

- create the new buffer as read only to avoid the need to use ! to exit etc
- find how to grab the content of the popup to put it into the new buffer
- find how to grab a meaningful title for the popup to put it as the new buffer title
- find how to enable the correct coloring of the new buffer

I will look into this more in the days to come :) .

### DIFF
--- a/helix-term/src/ui/popup.rs
+++ b/helix-term/src/ui/popup.rs
@@ -22,6 +22,12 @@ use std::path::Path;
 // TODO: share logic with Menu, it's essentially Popup(render_fn), but render fn needs to return
 // a width/height hint. maybe Popup(Box<Component>)
 
+pub struct StringsContent {
+    pub title: String,
+    pub body: String,
+    pub language: String,
+}
+
 pub struct Popup<T: Component> {
     contents: T,
     position: Option<Position>,
@@ -34,6 +40,7 @@ pub struct Popup<T: Component> {
     ignore_escape_key: bool,
     id: &'static str,
     has_scrollbar: bool,
+    contents_as_strings: Option<StringsContent>,
 }
 
 impl<T: Component> Popup<T> {
@@ -50,6 +57,28 @@ impl<T: Component> Popup<T> {
             ignore_escape_key: false,
             id,
             has_scrollbar: true,
+            contents_as_strings: None,
+        }
+    }
+
+    pub fn new_with_contents_as_strings(
+        id: &'static str,
+        contents: T,
+        contents_as_strings: StringsContent,
+    ) -> Self {
+        Self {
+            contents,
+            position: None,
+            margin: Margin::none(),
+            size: (0, 0),
+            position_bias: Open::Below,
+            child_size: (0, 0),
+            scroll: 0,
+            auto_close: false,
+            ignore_escape_key: false,
+            id,
+            has_scrollbar: true,
+            contents_as_strings: Some(contents_as_strings),
         }
     }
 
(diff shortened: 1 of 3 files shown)


helix-term/src/ui/popup.rs
```diff
@@ -207,6 +243,67 @@ impl<T: Component> Component for Popup<T> {
...
+                    .selection(view.id)
+                    .clone()
+                    .transform(|range| range.put_cursor(text, 0, cx.editor.mode == Mode::Select));
+                push_jump(view, doc);
+                doc.set_selection(view.id, selection);
+
+                // put a meaningful title
+                // TODO: get a meaningful buffer title
+                let stringified_buffername = Some(Path::new(&stringified_buffername));
+                doc.set_path(stringified_buffername);
```
the-mikedavis, line 294: The path of a buffer isn't meant to be a name - this should be left as a scratch buffer instead

helix-term/src/ui/popup.rs
```diff
@@ -207,6 +243,67 @@ impl<T: Component> Component for Popup<T> {
...
+                // put the content in the new buffer
+                let (view, doc) = current!(cx.editor);
+                paste_impl(
+                    &[stringified_body],
+                    doc,
+                    view,
+                    Paste::Before,
+                    1,
+                    cx.editor.mode,
+                );
```
the-mikedavis, line 275: The new document should be created with `Document::from` and then added to the editor with `Editor::new_file_from_document` (which is currently private but could be `pub`). That will remove the need to modify the selection, jumplist and last-saved-revision.

helix-term/src/ui/popup.rs
```diff
@@ -207,6 +243,67 @@ impl<T: Component> Component for Popup<T> {
...
+                // put a meaningful title
+                // TODO: get a meaningful buffer title
+                let stringified_buffername = Some(Path::new(&stringified_buffername));
+                doc.set_path(stringified_buffername);
+
+                // pretend that this buffer is up to date and has no changes: allow to just :bc when finished exporing, no need to :bc!, and no saving to disk happening after browsing around
+                doc.set_last_saved_revision(1);
+                // make readonly
+                // TODO: this does not make really readonly from helix viewpoint, why?
+                doc.readonly = true;
```
the-mikedavis, line 300: This is set for files that have readonly permissions. A readonly "mode" for buffers that prevents editing isn't implemented yet

helix-term/src/ui/popup.rs
```diff
@@ -207,6 +243,67 @@ impl<T: Component> Component for Popup<T> {
...
+
+                // close the popup since we just opened it in a new buffer
+                let _ = self.contents.handle_event(event, cx);
```
the-mikedavis, line 303: This passes the `C-o` event to the contained component (a `Markdown` in hover's case) which isn't necessary. The popup is handling the event so it shouldn't pass it down. Closing the popup is done below by returning `Some(close_fun)`

the-mikedavis (changes requested):
I don't intend on properly looking at or merging this feature until #5505 is possible. With #5505 we could make this possible only on popups where this makes sense, like hover, and we could cleanly downcast the type to a concrete `Popup<Markdown>`, where we would be able to read the contents. Without it we need to duplicate the contents and store it on the Popup as you have here with `StringsContent` which IMO is too hacky.

helix-term/src/ui/popup.rs
the-mikedavis: No, these docs shouldn't be opened for very long if they're just being used for reference anyways so I think scratch is fine

helix-term/src/ui/popup.rs
the-mikedavis: See https://github.com/helix-editor/helix/pull/7128. I suspect that it would be a large change and it needs some design before anyone goes about making a PR that attempts it

the-mikedavis:
Here's a block that finds and downcasts a component in the compositor into a `Overlay<Picker<T>>` or `Overlay<DynamicPicker<T>>`: https://github.com/helix-editor/helix/blob/2661e05b3405c216a2303c20e71830b4bde7ac35/helix-term/src/ui/picker.rs#L477-L482

The compositor uses [`std::any::Any`](https://doc.rust-lang.org/std/any/trait.Any.html) for dynamically typing the components.

jerabaul29:
Close in favor of #15187 .

PR #9316, opened 2024-01-11
Remove sourcehut tree-sitter grammars from default build

the-mikedavis:
Sourcehut has outages occasionally that cause the CI and from-source builds to fail. It also doesn't setup redirects when a user renames themselves, so if a user that publishes a tree-sitter grammar we use changes their sourcehut name then it breaks the build and any prior builds using that grammar.

For now let's remove them from the default build. It's a bandaid over a larger reliability and trust problem with the grammar repositories but it should fix the build for now.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1,6 +1,8 @@
 # Language support configuration.
 # See the languages documentation: https://docs.helix-editor.com/master/languages.html
 
+use-grammars = { except = [ "hare", "wren", "gemini" ] }
+
 [language-server]
 
 als = { command = "als" }


archseer (approved):

pascalkuthe (approved):
these are really niche languages anyway and it is not hard to reenable them so this seems like a good solution. I would be in favor of some sort of mirror eventually

ddevault:
Hi there! I'm both a maintainer of Hare and a sysop at SourceHut. I would appreciate it if you would consider restoring this to the default build.

The SourceHut outage that caused this problem was an extraordinary event. Ordinarily, SourceHut has substantially better uptime and reliability than even GitHub. If you'd like to read our post-mortem on events to judge if it instills greater confidence, we have published it here:

https://sourcehut.org/blog/2024-01-19-outage-post-mortem/

Thanks for your consideration.

the-mikedavis:
Thank you for writing that up, it's an excellent read! I was following along at the time and I really appreciate your transparency especially during a bad situation.

I'm not so worried about the reliability now - particularly after reading the postmortem. The tougher issue for us to work around is user renaming. I can't find the original issue about it but a few releases ago one of the SourceHut-hosted grammar repositories' authors changed their username. That caused the clone of their repository to fail which broke the build for that release. That's what makes me hesitant to re-enable these by default: if a repository owner changes their name then we're forced to cut a new release to fix the build for packagers.

For the future our plan is to change the way that we download the parser/scanner and query files. (Currently we're looking at something like a Cargo v1 style index, i.e. a git repository.) There are a number of drawbacks of cloning so many repositories that we can solve by setting up our own index, and it would pave the way for interactively installing parsers and queries in the future - something we've wanted for a while. With that in place we could comfortably re-enable these languages by default.

ddevault:
>The tougher issue for us to work around is user renaming.

I'm not sure what happened here but it definitely wasn't that -- users cannot be renamed on SourceHut right now. What's more likely is that they deleted the repository and/or their account, which can happen on any platform and isn't unique to SourceHut.

>For the future our plan is to change the way that we download the parser/scanner and query files.

:+1: 

One of the Hare maintainers is maintaining a tree sitter for Hare here, which is the current recommended option from Hare upstream:

https://git.sr.ht/~ecs/tree-sitter-hare

This can be reasonably expected to have longevity and availability. Perhaps you're open to using this for now?

the-mikedavis:
Ah ok that makes sense, they must've deleted their account and moved their repositories over to a new one. I would be open to re-enabling Hare by default then, I'll add a PR and see what the others think. Pascal and I worked on the future plan I mentioned this last week and got pretty far so it may be a moot point soon enough anyways (i.e. all of the sourcehut based languages would be enabled by default).

ddevault:
Thanks! Can you also update it to point to the upstream-maintained tree sitter?

https://git.sr.ht/~ecs/tree-sitter-hare

the-mikedavis:
Yep actually that's what we already use: https://github.com/helix-editor/helix/blob/e2904794749ccc709a238a10e6ee364411c2b820/languages.toml#L2079-L2081

ddevault:
Thanks :)

PR #9317, opened 2024-01-11
Bump tree-sitter to latest master

the-mikedavis:
* query capture names now return `&str`s rather than `String`s
* the `#any-of?` predicate is now supported

I've also switched to our fork after fast-forwarding it to the latest master. The plan will be to publish a release to Cargo of our tree-sitter fork (with no changes to the library) right before the next Helix release.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2177,7 +2177,7 @@ dependencies = [
 [[package]]
 name = "tree-sitter"
 version = "0.20.10"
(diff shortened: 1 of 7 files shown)


pascalkuthe (approved):

woojiq:
Looks like v22.0 was released several days ago: https://github.com/tree-sitter/tree-sitter/releases
I haven't read the entire changelog, but I think we can use the stable version again.

the-mikedavis:
We should be able to switch back to upstream. I updated locally but the new msrv is 1.74.1 :/

PR #9318, opened 2024-01-11
config system prototype

pascalkuthe:
not anywhere near ready to merge. The general system is mostly complete.
What is missing is porting the entire editor to use the new config system and to add an adapter for the legacy serde config.

I already ported the language server configuration in 49dc5b8356fbb5de9ddfe6de1fd41b1aeffb0ef1 which should help showoff how this works out in practice.


helix-config/src/any.rs
```diff
@@ -0,0 +1,76 @@
+/// this is a reimplementation of dynamic dispatch that only stores the
+/// information we need and stores everythin inline. Values that are smaller or
+/// the same size as a slice (2 usize) are also stored inline. This avoids
+/// significant overallocation when setting lots of simple config
+/// options (integers, strings, lists, enums)
```
dead10ck, line 5: Is avoiding an allocation for every config item the only reason we're doing this? Has there been a known performance bottleneck on config memory access?

Trying to understand where this is coming from. This is much harder to understand than just boxing everything, and I was not under the impression that config was in a hot path.

pascalkuthe: I actually do kind of think its pretty hot. We read config fields on every single frame for a large amount of config options (hundreds). It also reduces memory consumption quite a bit which is more relevant if each config option is set as default, by the user per languages and potentially on the document (so in total up to 4 times). Really this sort of thing ought to be built into rust. There are some efforts around that.

You don't really need to be aware of how this work tough just treat like `Box<dyn Any>` when reading the rest of the code its really just optimizing access

dead10ck: Ah that's true, I didn't consider they are read in the rendering hot loop, good point 👍

Issue #7364, opened 2023-06-17
Determine comment tokens from injection layers


the-mikedavis:
`toggle_comments` should take language injections into account. For example in HTML:

```html
<p>
  C-c on this line should use the HTML comment token(s).
</p>
<script type="text/javascript">
  // C-c on this line should use the javascript comment token(s).
  foo();
</script>
```

To do this we should look at the byte range for each Range in the current Selection and find the LanguageLayer in `doc.syntax()` (`Syntax`) with the smallest/deepest range that completely covers the selection range. We will then need a way to find the LanguageConfiguration for a given LanguageLayer and then we can use that LanguageConfiguration's `comment_token` for that range in the selection.

The changes for this will conflict with #1505 / #4718 so this feature may be better off waiting for those changes to land.

gabydd:
Thanks for making this issue especially the implementation notes, I meant to make it this morning but got caught up with work. This does conflict a tiny bit with my pr but if someone does implement this it shouldn't be too hard to rebase.

I might pick this up after #4718 as well

pascalkuthe:
One complication will be that we have a language for comments that is injected into comments. We don't want to change the kind of comment tokens inside that obviously.

I think we want to do something similar for indentation (see https://github.com/helix-editor/helix/discussions/7381?notification_referrer_id=NT_kwDOA6_EWrM2ODA2NzgxOTM3OjYxODUwNzE0#discussioncomment-6230647)  so having shared infrastructure for detecting language inside injections would be useful

the-mikedavis:
For indentation and textobjects I think we will want to switch to running queries across injections using something like: https://github.com/the-mikedavis/helix/blob/8483de78624be0b52dcd4144bc54a5b918d369cd/helix-core/src/syntax.rs#L1271-L1309 (currently that lives on #2857 but could be ported out). That QueryItem gives the LayerId of the current match, so maybe the function in Syntax that gets a LayerConfiguration by byte range should actually be two functions: find the LayerId for a given byte range and then find a LanguageConfiguration for a LayerId.

pascalkuthe:
Note that I didn't mean indent queries (altough those should also be run across injections) but determining what kind of indent to use (tab, single space, four spaces,...). We could use th layerid from the query for autoindent but not for manual indent with `>` and `<` as we don't run a query there afaik

Splitting the functions in two sounds like a good idea tough 👍

the-mikedavis:
[...]

PR #9320, opened 2024-01-12
Run textobject queries across injections

### LINKED ISSUE #4618: Multi-language documents don't benefit from navigation by tree-sitter
### Summary

In documents with multiple languages such as vue, svelte or html files, contents of `<script>` or `<style>` represented by a single node (most often `raw_text`).

Here's example of what I see when calling `:tree-sitter-subtree` command (I deleted `<template>` contents so that whole subtree is visible, but everything inside template seems to work as it should):
Screenshot_20221106_202029

Similar story with template function invocations in javascript, in places where additional syntax highlighting is supported (to show that it's not just vue issue) 
Screenshot_20221106_203051

#### I tried this:

1. `hx [any vue, svelte or html file]`
2. Start using unimpaired-style navigation or Syntax-tree Motions in script or style sections

#### I expected this to happen:
Navigation works just as in javascript/css files 

#### Instead, this happened:
Unimpaired-style motions have no effect, `A-left` or `A-right` takes me to a sibling tag as opposed to a javascript textobject

### Helix log

Probably nothing useful

```
2022-11-06T20:47:39.351 helix_term::application [ERROR] Language Server: Method not found client/registerCapability
2022-11-06T20:47:40.244 helix_lsp::transport [ERROR] err <- "DeprecationWarning: 'createIdentifier' has been deprecated since v4.0.0. Use the appropriate method on 'ts.factory' or the 'factory' supplied by your transformation context instead.\n"
2022-11-06T20:47:40.244 helix_lsp::transport [ERROR] err <- "DeprecationWarning: 'createImportClause' has been deprecated since v4.0.0. Use the appropriate method on 'ts.factory' or the 'factory' supplied by your transformation context instead.\n"
2022-11-06T20:47:40.244 helix_lsp::transport [ERROR] err <- "DeprecationWarning: 'createLiteral' has been deprecated since v4.0.0. Use `factory.createStringLiteral`, `factory.createStringLiteralFromNode`, `factory.createNumericLiteral`, `factory.createBigIntLiteral`, `factory.createTrue`, `factory.createFalse`, or the factory supplied by your transformation context instead.\n"
2022-11-06T20:47:40.244 helix_lsp::transport [ERROR] err <- "DeprecationWarning: 'createImportDeclaration' has been deprecated since v4.0.0. Use the appropriate method on 'ts.factory' or the 'factory' supplied by your transformation context instead.\n"
2022-11-06T20:47:40.245 helix_lsp::transport [ERROR] err <- "DeprecationWarning: 'createCallExpression' has been deprecated since v4.0.0. Use the appropriate method on 'ts.factory' or the 'factory' supplied by your transformation context instead.\n"
```

</details>

### Platform

Manjaro Linux

### Terminal Emulator

Konsole, Kitty (probably irrelevant) 

### Helix Version

helix 22.08.1 (3881fef3)

the-mikedavis:
Textobjects currently only work on the root layer which can be limiting for languages that use injections extensively like Vue or HTML. We can run textobject queries across all injection layers to be able to use things like `]f`/`[f`/`mif`/`maf`/etc, for example in HTML:

#2857 has a few commits that add a function to `Syntax` that creates an iterator similar to `tree_sitter::QueryCaptures` that works across injection layers. This PR cherry-picks those commits and adds another that refactors textobject queries to use that function instead. The novel part of this PR is only the last commit. Textobject queries are relocated from `LanguageConfiguration` to `HighlightConfiguration` so they are accessible from `LanguageLayer` within queries. In the future, the indentation query should also be refactored like this but that has some extra work involved, see https://github.com/helix-editor/helix/issues/7364.

Closes #4618 (the other half of that issue is covered by issue #2311)

### DIFF
--- a/helix-core/src/movement.rs
+++ b/helix-core/src/movement.rs
@@ -1,7 +1,7 @@
 use std::{cmp::Reverse, iter};
 
 use ropey::iter::Chars;
-use tree_sitter::{Node, QueryCursor};
+use tree_sitter::Node;
 
 use crate::{
     char_idx_at_visual_offset,
@@ -13,7 +13,6 @@ use crate::{
     },
     line_ending::rope_is_line_ending,
     position::char_idx_at_visual_block_offset,
-    syntax::LanguageConfiguration,
     text_annotations::TextAnnotations,
     textobject::TextObject,
     visual_offset_from_block, Range, RopeSlice, Selection, Syntax,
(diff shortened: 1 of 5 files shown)


pascalkuthe:
I wonder if we want to create a way to turn this back off somehow. For HTML it makes a lot of sense but if we have doc comment injections in the future I could see this making text objects less useful. For example if navigating crates with detailed documentation each function may have multiple functions in its doc comments so `]f` becomes a lot less useful

the-mikedavis:
Ah I see what you mean, I wouldn't want to select a closure in a doc comment that's a rust codefence inside markdown for example. I think that wouldn't be too hard to add - we could add a property like `(#set! injection.no-textobjects-within)` to exclude those layers and all of their children recursively from the query iterator.

pascalkuthe:
Yeah exactlt. I didnt event think about closures (I always forgot ]f junps to those :D) that is an even better exampl.

A property like that sounds like a grear solution

the-mikedavis:
This will need to be largely rewritten after #12972. I will take another swing at this after that PR lands

PR #9326, opened 2024-01-12
Make nix flake respect unused grammars

loewenheim:
https://github.com/helix-editor/helix/commit/17dd102e5cccbb2a9a0f0224af63e52f3dab846b marked some grammars as "do not use", but the nix flake doesn't respect this setting. This is my attempt to rectify this situation.

### DIFF
--- a/grammars.nix
+++ b/grammars.nix
@@ -28,7 +28,8 @@
     owner = builtins.elemAt match 0;
     repo = builtins.elemAt match 1;
   };
-  grammarsToUse = builtins.filter (grammar: !(builtins.elem grammar.name (languagesConfig.use-grammars.except or []))) languagesConfig.grammar;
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Other than `except`, `only` is also possible and I think we should have `grammars.nix` respect it even if we don't currently use it in `languages.toml`: https://docs.helix-editor.com/languages.html#choosing-grammars

Here's where we do it in the source: https://github.com/helix-editor/helix/blob/17dd102e5cccbb2a9a0f0224af63e52f3dab846b/helix-loader/src/grammar.rs#L208-L220

I believe if both `only` and `except` are defined then we prefer `only` (ignoring `except`).

loewenheim:
Thanks for the heads-up, will refactor.

the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

Issue #7368, opened 2023-06-18
Supercollider support


mitchmindtree:
Just opening a tracking issue for supporting [SuperCollider's sclang](https://en.wikipedia.org/wiki/SuperCollider).

It looks like there's a tree-sitter implementation [here](https://github.com/madskjeldgaard/tree-sitter-supercollider).

There's some ongoing discussion around an LSP implementation [here](https://scsynth.org/t/supercollider-vscode-language-server-protocol-support/5574).

It'd be great to also support the SuperCollider IDE shortcuts for evaluating "paragraph"s of code for live-coding, though I haven't looked into how helix handles these sorts of lang-specific features just yet.

PR #9329, opened 2024-01-13
Add initial support for SuperCollider

pnelson:
Implements syntax highlighting for SuperCollider. See also #7368.

### DIFF
--- a/runtime/queries/supercollider/highlights.scm
+++ b/runtime/queries/supercollider/highlights.scm
@@ -1,0 +1,76 @@
+(line_comment) @comment.line
+(block_comment) @comment.block
+
+(argument name: (identifier) @variable.parameter)
+
+(local_var name: (identifier) @variable)
+(environment_var name:(identifier) @variable.builtin)
+(builtin_var) @constant.builtin
+
+(function_definition name: (variable) @function)
+
+(named_argument name: (identifier) @variable.other.member)
+
+(method_call name: (method_name) @function.method)
+
+(class) @keyword.storage.type
+
+(number) @constant.numeric
+(float) @constant.numeric.float
+
+(string) @string
+(symbol) @string.special.symbol
+
+[
+"&&"
+"||"
+"&"
+"|"
+"^"
+"=="
+"!="
+"<"
+"<="
+">"
+">="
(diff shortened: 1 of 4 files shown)


runtime/queries/supercollider/highlights.scm
```diff
@@ -0,0 +1,101 @@
+; highlights.scm
+; See this for full list: https://github.com/nvim-treesitter/nvim-treesitter/blob/master/CONTRIBUTING.md
```
the-mikedavis, line 2: Helix uses different capture names than neovim. See https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example captures like `parameter` need to become `variable.parameter`, `property` becomes `variable.other.member`, etc.

runtime/queries/supercollider/indents.scm
the-mikedavis, line 1: Helix also has a custom indentation system that these queries will need to be rewritten for, see https://docs.helix-editor.com/master/guides/indent.html

runtime/queries/supercollider/locals.scm
the-mikedavis, line 1: For locals we follow tree-sitter: https://tree-sitter.github.io/tree-sitter/syntax-highlighting#local-variables

so there should only be the captures `local.scope`, `local.definition` and `local.reference`

the-mikedavis (changes requested):

pnelson:
I touched up the `highlights.scm` and it's looking good when I run it on [tutorial.scd](https://raw.githubusercontent.com/igramjp/supercollider/master/tutorial.scd) but really struggled with the other tree-sitter files. Are highlights-only an acceptable breakpoint for this PR?

the-mikedavis (dismissed):

the-mikedavis:
Yep highlights-only is a good start :+1:. The other queries can be contributed as follow-ups if anyone is interested in working on them

the-mikedavis (dismissed):

pascalkuthe (dismissed):

pascalkuthe:
needs a rebase

the-mikedavis (approved):

PR #9330, opened 2024-01-13
Change path normalization strategy to not resolve symlinks

woojiq:
Closes https://github.com/helix-editor/helix/issues/7458

Successor to #7658, which I messed up somehow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1074,6 +1074,7 @@ dependencies = [
  "slotmap",
  "smallvec",
  "smartstring",
+ "tempfile",
  "textwrap",
  "toml",
  "tree-sitter",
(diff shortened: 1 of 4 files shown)


woojiq:
@pascalkuthe I added feature flags as you suggested. I don't know what to do with tests. It's bad to not write them for this non-trivial change, but testing is not easy too because we need to create a directory structure to test properly (it will do almost nothing on non-existent paths). What do you think?

pascalkuthe (approved):
I think yhe test cases you have are good enough. Great you added those!

Thanks for sticking with this trough the back and forth. This is a really dofficult problem and fiding a good solution that habdeles all constrains adequatly on all platforms isn't easy. It will be great to have symlinks working again!

the-mikedavis (approved):

PR #9340, opened 2024-01-15
highlight(scala): highlight abstract methods in traits and classes

jpaju:
Updates Scala queries to highlight abstract methods in traits and abstract classes.

Currently they are highlighted differently:
before

Updated queries highlight them equally:
after


the-mikedavis (approved):

jpaju:
If there are no blockers, could this be merged?

PR #9355, opened 2024-01-16
Add a helix-stdx crate for stdlib extensions

the-mikedavis:
We've talked internally about introducing a stdx crate similar to [rust-analyzer's](https://github.com/rust-lang/rust-analyzer/tree/master/crates/stdx). It should contain some of the functionality that extends Rust's standard library and can be a dependency of any other crate.

I started small in this PR with the path operations from `helix-core`. Canonicalizing and normalizing paths is useful in `helix-loader` as well, for example to expand tildes and normalize the path optionally provided in the `HELIX_RUNTIME` environment variable. This PR is mostly moved lines except for adding canonicalization/normalization/tilde-expansion to helix-loader (see the second commit).

In the future we could move other things like `helix-view/src/base64.rs`. @pascalkuthe also mentioned moving `helix-view/src/env.rs`, `helix-core/src/fuzzy.rs` and some utilities on top of Ropey like `rope_reader` and `find_first_non_whitespace_char`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1114,6 +1114,7 @@ dependencies = [
  "cc",
  "dunce",
  "etcetera",
+ "helix-stdx",
  "libloading",
  "log",
  "once_cell",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1060,6 +1060,7 @@ dependencies = [
  "etcetera",
  "hashbrown 0.14.3",
  "helix-loader",
+ "helix-stdx",
  "imara-diff",
  "indoc",
  "log",
@@ -1074,7 +1075,6 @@ dependencies = [
  "slotmap",
  "smallvec",
  "smartstring",
- "tempfile",
  "textwrap",
  "toml",
  "tree-sitter",
@@ -1136,6 +1136,7 @@ dependencies = [
  "helix-core",
  "helix-loader",
  "helix-parsec",
+ "helix-stdx",
  "log",
  "lsp-types",
  "parking_lot",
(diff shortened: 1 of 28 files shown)


helix-core/src/path.rs
```diff
@@ -1,125 +1,16 @@
-use etcetera::home_dir;
-use std::path::{Component, Path, PathBuf};
+use helix_stdx::path::{canonicalize, fold_home_dir, normalize};
```
pascalkuthe, line 1: IMO path hbdeling doesn't belong on helix-core at all. I would like helix-core to just be about "core editing/editor components".

I see that you introduce this to handle the dependency on the CWD handling in helix-loader.

I think there are two alternatives here:
* move the cwd handling to stdx::path. While its somewhat in spirit of helix-loader I think ultimately what we are doing is essentially wrapping `std::env::current_dir` to make preventing it from crashing.
* move this Module to hwlix-loader

I would prefer the first variant since it really just feels like the cwd stuff is making std more robust rather than being any actual loading.

pascalkuthe:
seems like some doctests are still failing

pascalkuthe:
otherwise lgtm

pascalkuthe (approved):
I think its good to establish this. [This article](https://matklad.github.io/2021/05/12/design-pattern-dumping-ground.html) really resonated with me about having a dedicated kitchen sync (and infact that is just what stdx is in ra).  i do have a lot more I want to move there (there is especially a lot of stuff in open PRs).

Having the crates established allows moving stuff in other open PRs there so its good to start small :+1:

archseer (approved):

Issue #1115, opened 2021-11-16
Display the color of color-hex-codes, color functions


NNBnh:
Display the color of color-hex-codes and color functions is pretty helpful for writing CSS and UI design... VScode support this feature by default, while NeoVim have an acceptable plugin [`vim-hexokinase`](https://github.com/RRethy/vim-hexokinase) that do the job:

> ![gif](https://media.giphy.com/media/PikQiakj2NYxFDmcdl/giphy.gif)
>
> <img src="https://user-images.githubusercontent.com/21000943/64057161-91d16500-cb67-11e9-83ab-535ad2489c5a.png" width="520" height="342">

I hope Helix could support for coloring:
- 6 digit hex codes (`#FFFFFF`).
- 3 digit hex codes (`#FFF`).
- RGB functions (`rgb(100%, 100%, 100%)`,  `rgb(255, 255, 255)`).
- RGBA functions (`rgba( ... )`).
- HSL functions (`hsl( ... )`).
- HSLA functions (`hsla( ... )`).

With styling options:
- Background.
- Foreground.
- Virtual (display a small colored square on the right).
- Sign column (a column next to the line numbers).

cole-h:
IMHO, I feel like this is more suited to a plugin, once that's become functional, and not something that should be built-in.

matoous:
So I mostly agree with @cole-h about this being suited for a plugin. On the other hand, I was very surprised to find this in LSP specs today: https://microsoft.github.io/language-server-protocol/specifications/specification-current/#textDocument_documentColor.

VuiMuich:
Despite the discussion if this should be a native option I would like to add a few wishes for config options:
- for the virtual display and the gutter/sign column display it would be cool if there where three options: `Square`, `Circle` and `Unicode` with an additional variable that takes the Unicode character as a String.
- if it is possible to control the thickness of an `underline` it would also be nice to have that as an option as well
- I guess an `outline` option is not feasibly achievable for a terminal editor, right?

Would it possible to implement this as a tree-sitter? I have no idea if it is possible to have to tree-sitter modules running on a single file..

the-mikedavis:
[...]

PR #9360, opened 2024-01-17
Semantic tokens

kirawi:
Based on #8021 

TODO

## Notes

- Semantic highlighting will be combined with tree-sitter by asynchronously extrapolating definitions per `locals.scm` from the semantic highlighting. I will be taking a look at this soon, but this is something that would also need to be addressed for #1252 in the future.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,3 +1,17 @@
+# we use tokio_unstable to enable runtime::Handle::id so we can separate
+# globals from multiple parallel tests. If that function ever does get removed
+# its possible to replace (with some additional overhead and effort)
+# Annoyingly build.rustflags doesn't work here because it gets overwritten
+# if people have their own global target.<..> config (for example to enable mold)
+# specifying flags this way is more robust as they get merged
+# This still gets overwritten by RUST_FLAGS though, luckily it shouldn't be necessary
+# to set those most of the time. If downstream does overwrite this its not a huge
+# deal since it will only break tests anyway
+[target."cfg(all())"]
+rustflags = ["--cfg", "tokio_unstable", "-C", "target-feature=-crt-static"]
+
+
 [alias]
 xtask = "run --package xtask --"
 integration-test = "test --features integration --profile integration --workspace --test integration"
+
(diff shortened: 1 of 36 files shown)


kirawi:
My main focus has been trying to workout how to use semantic highlighting to define new local definitions. It took a while since I went through some overly complicated designs before arriving at the one I'm proposing now:

- `Syntax`'s API would be extended to allow callers to pass a function that would define locals per each tree-sitter layer and each `local.scope` defined in `locals.scm`. The difference between each layer and each scope is significant from `Syntax`'s perspective since it would have different `local.scope`s if it references a different language. I could see this as a function that returns a vector of 3-tuples of the identifier itself, the token_type and the token_modifier. Ideally this API would also allow us to only work on a range of the text.
- However, it's important that we also remove outdated `local.definitions`. I think this should also be an extension of the API where a function is passed and can mutate the locals per each layer/scope (only those defined by the method above). In the case of semantic tokens, I think it would just be removed if either there are no more `local.references` with the definition's corresponding identifier or if there is now a `local.definition`. The reason I think a function is necessary is to accommodate stack-graphs where the existence of references does not matter in terms of correctness. I don't think this makes the API more complicated than if this change was scoped to just semantic highlighting.

kirawi:
@the-mikedavis I'm actually not sure that the approach I outlined above is optimal because a `LocalDef`'s range would have to be mapped on every change to the document, which I think is out-of-scope for `Syntax`. Since you have a lot of experience with tree-sitter, do you have any suggestions for how I can add "external" local definitions for a document?

kirawi:
In the future, I think https://github.com/helix-editor/helix/issues/1115 can be handled through the same code path as semantic tokens.

the-mikedavis:
Yeah extra information from the LSP that needs to be mapped through changes should live on the Document. `LocalDef` should still be private to `syntax.rs` though, I think it should be possible just to pass in a `std::ops::Range<usize>` since that's really the only info the language server gives you.

#1115 is more complicated and I don't think it will overlap with this. Knowing all of the definitions in the document (which you could figure out when running injections queries) doesn't help since you also need to run highlights across potentially the entire text in order to determine the `LocalDef`'s `highlight`.

PR #9371, opened 2024-01-17
Address clippy lints

wutchzone:
I ran the Clippy against the codebase and found some things that can be simplified so the code can be more readable.


the-mikedavis (approved):

wutchzone:
My apologies. Now, it should be correctly formatted.

pascalkuthe (approved):

PR #9372, opened 2024-01-17
Add diff support for mercurial (hg) repos

zegervdv:
Adds diff and branch name support for mercurial (hg) repositories.

I'm using the [rhg](https://foss.heptapod.net/mercurial/mercurial-devel/-/tree/branch/default/rust/rhg) command for faster execution, but that is currently not installed by default.
The speed-up brings it down from ~750ms overhead on startup to ~50ms, so definitely noticeable. 
But it breaks the portability for users or systems that only have the commonly used `hg` command installed.

Perhaps a config option might allow the user to decide to use the faster `rhg` or fall back to `hg`?

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -13,10 +13,11 @@ repository.workspace = true
 homepage.workspace = true
 
 [features]
-default = ["git"]
+default = ["git", "hg"]
 unicode-lines = ["helix-core/unicode-lines"]
 integration = []
 git = ["helix-vcs/git"]
+hg = ["helix-vcs/hg"]
 
 [[bin]]
 name = "hx"
--- a/helix-vcs/Cargo.toml
+++ b/helix-vcs/Cargo.toml
@@ -27,6 +27,7 @@ log = "0.4"
 
 [features]
 git = ["gix"]
+hg = []
 
 [dev-dependencies]
 tempfile = "3.9"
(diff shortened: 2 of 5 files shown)


kirawi:
We don't want to support any VCS other than `git` in core. Instead, we want support for other systems to be implemented via plugins.

zegervdv:
Oh, I see. In the meantime I'll keep using my fork then :)

archseer:
I think we should at least take a look, @kirawi's tone was too strong here. I like how small the implementation is but I wonder about performance

pascalkuthe:
I am sceptic about merging support for other VCS into core

* the testcases in this PR fail CI (and will also fail for almost all developers locally) because they require mecurial binaries to be present.
* I have a lot of plans regarding the evolution of diff support (diff against stash, diff against commits and other buffers) having a second VCS would require me to apply that same effort to that second VCS too which would increase the required effort a lot
* there are some subtelties here that aren't properly considered (encodings and line endings). I know a lot about git to maintain and handle these edgecases there but the same is not true for mecurial
* this is an API where its very easy to expose a plugin interface and this implemention shells out. That is just as easy to do in scheme as in rust
* adding support for everything in core blots compiletime. With firefox moving to git (and the mecurial project itself slowing donw). I think mecurial (and generally any no-git VCS) is niche enough to relegate to plugins (especially because it would be quite easy to implement as a plugin)

PR #9382, opened 2024-01-19
Fix error message shown for goto references

theteachr:
When a symbol with no references is <kbd>g</kbd><kbd>r</kbd>'d on, the message shown is `No definition found.` Make it `No references found.`

Achieves this by introducing empty container checks at call sites rendering one of the patterns redundant. Another possible solution is to pass the error message as another argument.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -1066,11 +1066,13 @@ where
 
     cx.callback(
         future,
-        move |editor, compositor, response: Option<lsp::GotoDefinitionResponse>| match to_locations(
-            response,
-        ) {
-            items if items.is_empty() => editor.set_error("No definition found."),
-            items => goto_impl(editor, compositor, items, offset_encoding),
+        move |editor, compositor, response: Option<lsp::GotoDefinitionResponse>| {
+            let items = to_locations(response);
+            if items.is_empty() {
+                editor.set_error("No definition found.");
+            } else {
+                goto_impl(editor, compositor, items, offset_encoding);
+            }
         },
     );
 }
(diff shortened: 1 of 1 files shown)


helix-term/src/commands/lsp.rs
```diff
@@ -1023,8 +1023,9 @@ fn goto_impl(
         [location] => {
             jump_to_location(editor, location, offset_encoding, Action::Replace);
         }
+        // XXX: All call sites pass non empty containers. This pattern is redundant.
```
the-mikedavis, line 1026: Let's have this be `[] => unreachable!("goto_impl must be passed a non-empty locations vec")` and add a line of documentation for the function that says that `locations` must be non-empty

helix-term/src/commands/lsp.rs
```diff
@@ -1065,9 +1066,11 @@ where
 
     cx.callback(
         future,
-        move |editor, compositor, response: Option<lsp::GotoDefinitionResponse>| {
-            let items = to_locations(response);
-            goto_impl(editor, compositor, items, offset_encoding);
+        move |editor, compositor, response: Option<lsp::GotoDefinitionResponse>| match to_locations(
```
the-mikedavis, line 1069: The `match` is concise but it looks a bit awkward to me. Style-wise a regular `if` seems clearer to me:

```rust
let items = to_locations(response);
if items.is_empty() {
    editor.set_error("No definitions found.");
} else {
    goto_impl(editor, compositor, items, offset_encoding);
}
```

and same for the change for references below

theteachr: I had this earlier, but switched to using a `match`. But I agree. Will revert to using a plain `if-else`.

helix-term/src/commands/lsp.rs
theteachr: Addressed in https://github.com/helix-editor/helix/pull/9382/commits/e6ba42934868f9cef9d55e2a37534e7f1c30fd31.

the-mikedavis (approved):
Thanks!

PR #9384, opened 2024-01-19
Add `unbind_default_keys` config option

### LINKED ISSUE #2720: Unbind all default keybindings config option

postsolar:
Based on https://github.com/helix-editor/helix/pull/2733/

Closes #2720


postsolar:
There's a bug currently: default keymap still gets applied if there's a `.helix` directory with other bindings in the project root.

the-mikedavis:
See https://github.com/helix-editor/helix/issues/2720#issuecomment-2065533370, we want to solve this when we switch to scheme rather than adding a new config option as a stopgap

Issue #7982, opened 2023-08-17
Map Mouse Buttons in Config


Joe-Zer0:
I would like to propose allowing users to set the mappings for the mouse buttons in the config file.  Here is an 

```
[mouse.normal.editor]
# right click copies text to system clipboard
mouse2 = ":clipboard-yank"
# something like the following would be useful too
# double-click = "miw"

[mouse.normal.gutter]
# disable setting breakpoints with mouse
mouse1 = "no_op"
```

willhansen:
I've been trying to figure out how to bind middle-click to "gd" and the thumb button to "C-o".  Has it just not been implemented as possible yet?

pascalkuthe:
no its not implemented yet, its probably reasonable to add some form of remap capability but I see helix as a primarily keyboard driven editor so I  see mouse-based mapping as pretty low priority

Joe-Zer0:
That's fair.  Thanks for the response!

tmpm697:
I aslo have this interest, I want to select text with left mouse and then copy them to `system` clipboard with right mouse.

0x61nas:
I think that I can work on this, please assign this issue to me if there's no one working on

Nilsia:
0x61nas I do not know if you implemented this feature yet, however if you are not motivated or don't want to do it anymore.

I would be pleased to take this issue.

I think a bit about it, and I believe that the best way to implement it is to pass into selection mode when double-click / triple-click, or even pressing continuously the left click.
For yanking the solution would be to use the command 'yank' instead of the right click (I do not think that is a good idea to overwrite it, as well as the right click open a menu provided by the OS)

0x61nas:
> however if you are not motivated or don't want to do it anymore

I don't if I'll have the space to work on this anytime soon

> I would be pleased to take this issue.

it's yours now, go ahead

Joe-Zer0:
In Windows Terminal (not in Helix) the right click behavior is to copy to clipboard if there is a selection and the paste if there is no selection.  There are no menu's inside the terminal.  I think it would be intuitive to mimic this behavior inside of Helix.

Nilsia:
Yes, that right, when the mouse is enabled, there is no menu.
I will do what I can do and give a try to Windows when finished.

Joe-Zer0:
That would be amazing! Right click yank is my most desired feature.

Nilsia:
Hi, I have mostly finished the configuration for mouses.
[...]

PR #9386, opened 2024-01-19
Feature: mouse keymaps configuration

Nilsia:
Feature: enable configuration for mouse keys

Extended KeyMaps for MouseEvents.
Added a trait for KeyEvent and MouseEvent, so Keymaps depends on generics.

Updated / Added tests for mouses in configuration.

This is a pull request for this issue : https://github.com/helix-editor/helix/issues/7982

### DIFF
--- a/helix-lsp/src/client.rs
+++ b/helix-lsp/src/client.rs
@@ -1,4 +1,5 @@
 use crate::{
+    file_operations::FileOperationsInterest,
     find_lsp_workspace, jsonrpc,
     transport::{Payload, Transport},
     Call, Error, OffsetEncoding, Result,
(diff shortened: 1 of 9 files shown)


the-mikedavis (changes requested):
I agree with @pascalkuthe that #7982 would be a nice-to-have feature but Helix is keyboard-centric (ref https://github.com/helix-editor/helix/issues/7982#issuecomment-1686723233). This is a really massive change that we wouldn't be comfortable merging for a low priority issue.

I scanned through the changes though and it looks like the bulk of the changes come from trying to make mouse config and commands roughly equivalent to keyboard config and commands via the trait. The trait is forcing an abstraction that shouldn't exist. You don't need to be able to execute regular commands with the mouse and definitely not execute mouse commands via the keyboard. Instead mouse commands can be separate (for example under a `helix-term/src/commands/mouse.rs` module) and take additional parameters like the mouse position. The existing commands and keymap structs can be left untouched.

Nilsia:
Well, I all agree.
I will rework on this, to make these changes:
 - Add a new type of commands that are only made for the mouse (some of them will point out to key commands like goto_line_end or move_next_word_end (using the last one makes easier selecting a word)),
 - differ Keymaps and Mousemaps where Mousemaps will only accept 2-left = "" and not 2-left = { 1-right = "" }
 - the code will maybe be in some ways a bit duplicated, as so I will try to add traits without generics

PR #9397, opened 2024-01-21
Add bufferline config for onedark & onedarker themes

jpaju:
Highlight the active buffer more clearly in onedark and onedarker themes.

Before:
before

After:
after


sudormrfbin:
LGTM :+1:

the-mikedavis (approved):

PR #9411, opened 2024-01-23
Add text object queries for dart

melted-brownie:
I added queries for text objects for dart.

To do it, I looked at what was done in queries for rust. And I looked with `:tree-sitter-scopes` and `:tree-sitter-subtree` in my dart code. I works quite well for me (and should work well for others too, unless there is parts of the grammar not covered in my dart code)

For the "test" object, I used an alternation with `#eq?` predicates, because I could not make it work with one `#any-of?` predicate (it matched any name). I don't know if I missed something and it could work, or if I missed something in my understanding and that it cannot work

### DIFF
--- a/runtime/queries/dart/textobjects.scm
+++ b/runtime/queries/dart/textobjects.scm
@@ -61,7 +61,7 @@
 (formal_parameter_list) @parameter.around
 
 (expression_statement
-  ((identifier) @_name1 (#any-of? @_name1 "test" "testWidgets"))
+  ((identifier) @_name (#any-of? @_name "test" "testWidgets"))
   .
   (selector (argument_part (arguments . (_) . (argument) @test.inside)))
 ) @test.around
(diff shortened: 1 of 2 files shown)


runtime/queries/dart/textobjects.scm
```diff
@@ -0,0 +1,71 @@
...
+
+(formal_parameter) @parameter.inside
+
+(formal_parameter_list) @parameter.around
+
+(expression_statement
+  [
+    ((identifier) @_name1 (#eq? @_name1 "test"))
+    ((identifier) @_name2 (#eq? @_name2 "testWidgets"))
+  ]
```
the-mikedavis, line 67:
```suggestion
  ((identifier) @_name1 (#any-of? @_name1 "test" "testWidgets"))
```

Is this what you were referring to in the PR description? This works for me locally

pascalkuthe (approved):

PR #9412, opened 2024-01-23
Re-export `which` from `helix-stdx::env`

the-mikedavis:
We use `which::which` in many crates, so `which` was a separate dependency across all of them. We can centralize `which` into the stdx crate so it's easy for all crates to depend on it.

I also moved the rest of `helix-view/src/env.rs` into helix-stdx's `env` module since it only contained a thin wrapper around `which` and `std::env`.

This also updates `which` to 6.0.0, so it closes #9406

### DIFF
--- a/helix-stdx/src/env.rs
+++ b/helix-stdx/src/env.rs
@@ -1,6 +1,5 @@
-pub use which::which;
-
 use std::{
+    ffi::OsStr,
     path::{Path, PathBuf},
     sync::RwLock,
 };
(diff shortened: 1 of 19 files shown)


pascalkuthe:
Looking at the callsite the error handling is kind of ad-hoc in most placed we just create an anyhow error. We never really care about the original error type.

Could we wrap the which function here and provide a reasonable error message in a single place so we don't have to repeat that everywhere (and good the same quality and consistent error messages everywhere)

helix-stdx/src/env.rs
```diff
@@ -30,6 +31,41 @@ pub fn set_current_working_dir(path: impl AsRef<Path>) -> std::io::Result<()> {
...
+pub struct ExecutableNotFoundError {
+    command: String,
+    inner: which::Error,
+}
+
+impl std::fmt::Display for ExecutableNotFoundError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(
+            f,
+            "could not find '{}' executable: {}",
```
pascalkuthe, line 61:
```suggestion
            "command '{}' not found: {}",
```

nit: I think this sounds better (closer to the error messages in shells) :D

pascalkuthe (approved):

PR #9414, opened 2024-01-24
RFC: Implement log compression

StratusFearMe21:
I was checking my helix log today and it's gotten to 372.5 MB, and last year I let it get to over 8 GB. I was thinking that maybe to prevent log files from getting so big, the editor could have an option to compress them when the user enables an option. This PR implements a system that does that.

If the user enables the `editor.compress-log` option in `config.toml`, the log file from then on would be compressed using a custom `zstd` based format. The compression works by compressing each individual log line separately using `zstd` and a custom dictionary, and separating each individual compressed line using the `cobs` format. Log files can be decompressed by piping a program into helix and running it with `--decompress-log`. This allows for parts of the file to be decompressed without decompressing the whole file, which allows for stuff like this to still work
```
tail -f ~/.cache/helix/helix.log.cmp | hx --decompress-log
```
I've found the compression to be pretty solid, it can get a 372.5 MB log file down to 145.1 MB, a 39% compression ratio.

Personally, I'm pretty happy with what I've come up with, but I'd love to hear your ideas. As far as I know there's no other system that compresses logs like this, so any input from the community and you guys would be stellar


pascalkuthe:
This was discussed before on Matrix. IIRC we agreed this wasn't really a concern. If we do want to address this we should just do some simple log rotation (if log file is larger than x on startup move it to log.old, next time that happens log.old is overwriten)

Also this is not an RFC. We don't have Ana rfc process and an RFC is a proposal aimed to gather feedback from a detailed writeup before putting effort into an implementation.

StratusFearMe21:
Gotcha, thanks for responding so quickly (and thanks for educating me on what an RFC is :grimacing:)

archseer:
There's quite a few common errors that are getting logged to the log file that I think should be silenced. Namely, the editor status line displaying an error writes to the error log (seems like a leftover from integration tests?) and not finding a LSP binary. If we silence the two it should stop the growth in most cases (where it's not caused by a faulty LSP)

archseer:
https://github.com/helix-editor/helix/commit/6bfe1ddc53f542d62e242fd4aaf6748dda1b0e71

archseer:
https://github.com/helix-editor/helix/commit/83f09ecbff5160e4350c8099be1ad4c64513f665

dead10ck:
Yeah I agree, there is room for adjusting our logs to not be quite as noisy, which will help, but the overall problem of log management should not be something helix concerns itself with. This is standard faire for system logs.

Applications usually shouldn't have to know or care how logs are handled once they're written, as this is usually handled externally, either by the OS's facilities, like SystemD, or another application meant for handling logs, like logrotate.

PR #9424, opened 2024-01-25
Respect injections when using the syntax tree's root node

### LINKED ISSUE #9423: match bracket wrong in `<style>` tag of html
### Summary

match bracket position wrong for style tag in html file.
```html
<style>
.example { // <- I want to move to here by `mm` cmd.
   box-sizing: border-box;
} // <- cursor here
</style> 
```
But position was at <kbd>></kbd> of `</style>` when I press `mm`.

### Reproduction Steps

图片

https://github.com/helix-editor/helix/assets/23305067/a822e7bd-45d1-4ab2-a156-c1951f4c5447

### Helix log

_No response_

### Platform

macOS

### Terminal Emulator

wezterm-20230712-072601-f4abf8fd

### Installation Method

brew

### Helix Version

23.10

the-mikedavis:
Previously we used the root syntax tree for bracket matching. We can use the new functionality in `Syntax` for finding the correct syntax tree for a given byte range though so we use the correct syntax tree within injections. This improves bracket matching behavior within HTML injections like script or style tags for example.

Fixes #9423

### DIFF
--- a/helix-core/src/match_brackets.rs
+++ b/helix-core/src/match_brackets.rs
@@ -57,10 +57,10 @@ fn find_pair(
     pos_: usize,
     traverse_parents: bool,
 ) -> Option<usize> {
-    let tree = syntax.tree();
     let pos = doc.char_to_byte(pos_);
 
-    let mut node = tree.root_node().descendant_for_byte_range(pos, pos + 1)?;
+    let root = syntax.tree_for_byte_range(pos, pos + 1).root_node();
+    let mut node = root.descendant_for_byte_range(pos, pos + 1)?;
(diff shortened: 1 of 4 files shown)


the-mikedavis:
There are a few other places where we can replace `Syntax::tree` with functions that take injections into account. I'll add some commits for fixing `:tree-sitter-highlight-name` and `helix_core::movement::move_parent_node_end` to this PR too since they're small. In the long run we should treat `Syntax::tree` as deprecated and remove it once we respect injection layers in the indent system

archseer (approved):

PR #9435, opened 2024-01-26
Make status line modes bold for theme onedark

twharmon:
This is just my opinion, but with the colored backgrounds, the mode is much easier to read when bold (like the catppuccin themes).

before:
regular

after:
bold


the-mikedavis:
What do you think @sudormrfbin?

PR #9448, opened 2024-01-27
Use range positions to determine insert_newline motion

### LINKED ISSUE #9443: Odd behaviour with `append_mode` `collapse_selection`
### Summary

I use helix, and I map <kbd> a </kbd> to `append_mode` `collapse_selection`. There is an odd bug where the selection comes up again after pressing enter even though the last command collapsed the selection.

### Reproduction Steps

```toml
# ~/.config/helix/config.toml
keys.normal.a = ["append_mode", "collapse_selection"]
```

image

Write some text and then add a left parenthesis to generate a pair
image

Then press <kbd> esc </kbd> then <kbd> a </kbd> (which is mapped as above)
image

Now press <kbd> enter </kbd>
image

I expected this to happen:
image

Instead, this happened:
image

### Helix log

```
```

</details>

### Platform

Windows

### Terminal Emulator

Windows Terminal

### Installation Method

Scoop

### Helix Version

helix 23.10 (f6021dd0)

wmstack:
Fixes #9443

Uses the relative postion of the anchor and the head of a selection to determine whether to shift the selection or to extend it.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3650,8 +3650,7 @@ pub mod insert {
 
                 (pos, pos, local_offs)
             };
-
-            let new_range = if doc.restore_cursor {
+            let new_range = if range.head > range.anchor {
                 // when appending, extend the range by local_offs
                 Range::new(
                     range.anchor + global_offs,
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -3650,7 +3650,8 @@ pub mod insert {
 
                 (pos, pos, local_offs)
             };
-            let new_range = if range.head > range.anchor {
+
+            let new_range = if range.cursor(text) > range.anchor {
                 // when appending, extend the range by local_offs
                 Range::new(
                     range.anchor + global_offs,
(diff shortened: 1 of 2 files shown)


helix-term/src/commands.rs
```diff
@@ -3650,8 +3650,7 @@ pub mod insert {
 
                 (pos, pos, local_offs)
             };
-
-            let new_range = if doc.restore_cursor {
+            let new_range = if range.head > range.anchor {
```
the-mikedavis, line 3653: This will also extend if you enter insert mode via `o` (`open_below`), `O` (`open_above`), or mouse click when in insert mode and then hit enter. Instead I believe we want to check `range.cursor(text) > range.anchor` to avoid cases where the range is a point (`Range::point`)

wmstack: Done. However, I am not too familiar with `range.cursor(text)`

the-mikedavis (dismissed):
I'm pretty sure that this is the correct condition. I will test this out locally for a while to see if I can find any cases where we extend unexpectedly

pascalkuthe:
I agree this is correct and inline with how other commands work. I would actually go further: There is only one other place where we use `restore_cursor` (actually restoring the cursor) I think there we should check `range.head > range.anchor && self. restore.cursor` for each selection and only reduce those selections by one where that holds.

the-mikedavis:
:+1: That would fix the other behavior mentioned in #9443

the-mikedavis (approved):

the-mikedavis:
Looks like this needs a run of `cargo fmt` to pass the lint CI

Issue #9436, opened 2024-01-26
Reverse tree-sitter query precedence ordering


the-mikedavis:
Currently our query precedence order follows the behavior of `tree-sitter-cli` (the `tree-sitter-highlight` crate) and is the reverse of the order followed by Neovim and now Zed. For example with a query like so:

```scm
((identifier) @constant
 (#match? @constant "^[A-Z][A-Z\\d_]*$"))
(identifier) @variable
```

Helix will currently treat the first pattern as a higher precedence than the second, so identifiers starting with capital letters will be highlighted as constants. In Neovim for example though, all identifiers would always be highlighted as variables since the later pattern is higher precedence.

In the next release, `tree-sitter-highlight` will reverse its ordering so that it follows Neovim / Zed. We should reverse our order as well so that queries are closer to being interoperable between editors. The ordering has always been a snag for language support PRs when translating queries from neovim anyways.

The actual change to `helix-core/src/syntax.rs` in order to reverse the precedence should be very small, and will probably be almost identical to the changes necessary upstream in tree-sitter (https://github.com/tree-sitter/tree-sitter/pull/2412/files#diff-538bdc046297d41717e7a863a54456d47829fb2e298963198cdfb32684500cdd), but we will need to modify all existing tree-sitter queries (`runtime/queries/*/*.scm`) to respect the new ordering.

Also see https://github.com/nvim-treesitter/nvim-treesitter/issues/3944#issuecomment-1458782497

postsolar:
If someone is willing to submit a patch for a grammar they're specifically accustomed with, when should that happen?

the-mikedavis:
I pushed up a branch in https://github.com/helix-editor/helix/pull/9458: `reverse-query-precedence-ordering`. Feel free to PR against that branch for any languages you'd like to update

postsolar:
Thank you, I sent a PR for PureScript and might look into Haskell a bit later.
Another question though: could you give a link for this bit:
> In the next release, tree-sitter-highlight will reverse its ordering so that it follows Neovim / Zed

I couldn't find that announcement in TS repo.

the-mikedavis:
I'm not sure if there's an announcement in the tree-sitter repo yet. I heard it in a matrix chat we have with some tree-sitter/neovim people

tgross35:
Upstream TS did the release with reversed precedence order this week, at least that is my understanding of https://github.com/tree-sitter/tree-sitter/releases/tag/v0.21.0.

[...]

PR #9458, opened 2024-01-28
Reverse highlight precedence ordering

the-mikedavis:
Closes https://github.com/helix-editor/helix/issues/9436


pascalkuthe:
dump question: couldn't we simply write a script that readsc all querys and reverses all lines?

the-mikedavis:
It would probably be a little tricky to write so that it handles comments correctly but yeah that would be the most efficient way

archseer:
There's also files that were copied and weren't properly reversed in the first place

archseer:
I added a progress checklist to the PR description. I'm going through these one by one and comparing them to the nvim-treesitter ordering: in many cases there's nothing to do or the file wasn't correctly reordered in the first place.

the-mikedavis:
I went through the remaining languages and the new ones after this PR was last updated. I spent some extra time on the common programming languages like Python, C++, etc., but it's hard to be confident about all of these changes without rigorous testing against example files, and there are too many languages for that to be feasible. In the future we could consider having highlight tests like the tree-sitter-cli does. (@pascalkuthe and I also talked about the possibility of "code coverage" for queries in the past, for an extra step in this direction.)

I'm going to merge this down but I expect some level of breakage. Since we're early in the release cycle this is a good time to make a change like this that may need follow-ups/fixes.

PR #9463, opened 2024-01-29
Update PureScript highlight queries in light of reversing queries pre…

postsolar:
…cedence order

This commit adapts PureScript highlight queries to match upcoming changes in Helix which will make queries matching order go from more general to more specific, the opposite of previous order.

### DIFF
--- a/runtime/queries/purescript/highlights.scm
+++ b/runtime/queries/purescript/highlights.scm
@@ -1,33 +1,17 @@
 ; ----------------------------------------------------------------------------
-; Record fields would need to come before literal strings in order to be captured correctly
-
- (record_accessor
-    field: [ (variable)
-             (string)
-             (triple_quote_string)
-           ] @variable.other.member)
-
- (exp_record_access
-    field: [ (variable)
-             (string)
-             (triple_quote_string)
-           ] @variable.other.member)
-
-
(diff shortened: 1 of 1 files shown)


runtime/queries/purescript/highlights.scm
```diff
@@ -151,14 +157,5 @@
...
-
- ("@" @namespace)  ; "as" pattern operator, e.g. x@Constructor
-
-; ----------------------------------------------------------------------------
-; Types
-
- (type) @type
-
- (constructor) @constructor
+ (patterns (pat_as "@" @namespace))
```
postsolar, line 160: The change here is off-topic but fixes a bug (conflict with the other use of `@`).

runtime/queries/purescript/highlights.scm
```diff
@@ -41,18 +25,19 @@
    "]"
  ] @punctuation.bracket
 
- [
- (comma)
- ";"
- ] @punctuation.delimiter
+ (comma) @punctuation.delimiter
```
postsolar, line 28: Removed `;` is also off-topic but it's not a token we parse in the upstream grammar. It was present here because this is a fork of Haskell's grammar, and I forgot to remove it previously.

the-mikedavis (approved):

PR #9469, opened 2024-01-29
Key remappings for mouse configuration

Nilsia:
Feature: from the configuration file, it is now possible to map mouse buttons on some helix functionality.

Code:
Added new simple structures for Mousemaps,
and cloned only few key maps functions into mouse functions, do not hesitate if more are required.

### DIFF
--- a/helix-term/src/commands/mouse.rs
+++ b/helix-term/src/commands/mouse.rs
@@ -1,0 +1,283 @@
+use crate::ui::{
+    editor::{gutter_coords_and_view, pos_and_view},
+    EditorView,
+};
+use anyhow::anyhow;
+use helix_core::{movement::Direction, Range, Selection};
+use helix_view::{input::MouseEvent, Document, ViewId};
+
+use super::{
+    move_next_long_word_end, move_next_word_end, move_prev_long_word_start, move_prev_word_start,
+    paste_primary_clipboard_before, replace_selections_with_primary_clipboard,
+    yank_primary_selection_impl, Context,
+};
+
+#[derive(Eq, PartialEq, PartialOrd, Ord, Clone, Debug, Copy)]
+pub struct StaticMouseCommand {
+    name: &'static str,
+    pub(crate) fun: fn(&mut Context, &MouseEvent, &mut EditorView),
+    doc: &'static str,
+}
+
+impl std::str::FromStr for StaticMouseCommand {
+    type Err = anyhow::Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        StaticMouseCommand::STATIC_MOUSE_COMMAND_LIST
+            .iter()
+            .find(|cmd| cmd.name() == s)
+            .cloned()
+            .ok_or_else(|| anyhow!("No command named '{}'", s))
+    }
+}
+
+macro_rules! static_mouse_commands {
+    ( $($name:ident, $doc:literal,)* ) => {
+        $(
+            #[allow(non_upper_case_globals)]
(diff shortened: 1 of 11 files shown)


helix-view/src/input.rs
```diff
@@ -27,6 +50,106 @@ pub struct MouseEvent {
...
+    type Err = Error;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        let mut tokens: Vec<_> = s.split('-').collect();
+        let kind = match tokens.pop().ok_or_else(|| anyhow!("Missing mouse code"))? {
+            mouse_keys::LEFT => MouseEventKind::Down(MouseButton::Left),
+            mouse_keys::MIDDLE => MouseEventKind::Down(MouseButton::Middle),
+            mouse_keys::RIGHT => MouseEventKind::Down(MouseButton::Right),
+            mouse_keys::SCROLL_DOWN => MouseEventKind::ScrollDown,
+            mouse_keys::SCROLL_UP => MouseEventKind::ScrollUp,
```
thecaralice, line 117: I think you forgot left and right scroll here

thecaralice:
Looks like only a dozen or so commands are available for mapping to mouse events, suffixed with `_mouse` and non-discoverable (i.e. not listed in ` ?` picker). I feel like it should be possible to bind other commands to mouse events without duplicating everything.

thecaralice:
What I expected to work (aka "two-finger scroll moves through time instead of space" aka [xkcd:1806](https://xkcd.com/1806/)):

```toml
[mouse.normal]
scroll_right = "undo"
scroll_left = "redo"
```

Vim supports that as
```vim
nnoremap <ScrollWheelRight> :earlier<CR>
nnoremap <ScrollWheelLeft>  :later<CR>
```
(source: [r/vim](https://old.reddit.com/r/vim/comments/5xkh2j/xkcd_1806_would_it_be_possible_to_implement_the/dej7a8j/))

helix-term/src/commands/mouse.rs
```diff
@@ -0,0 +1,283 @@
...
+    move_prev_long_word_start(cx);
+    move_next_long_word_end(cx);
+}
+
+fn select_all_mouse(cx: &mut Context, _: &MouseEvent, _: &mut EditorView) {
+    super::select_all(cx)
+}
+
+fn scroll_mouse_impl(cx: &mut Context, evt: &MouseEvent, dir: Direction, _: &mut EditorView) {
+    let current_view = cx.editor.tree.focus;
```
isaldarriaga, line 151: Should we check if a picker has the focus?

currently the mouse wheel does not scroll the code preview in the picker, but the code in the current buffer.

Nilsia:
I am trying to concatenate both STATIC_COMMANDS list through a macros without repeating everything, however, the macro does not seem nice.
Well you'll see, I do not know if a procedural macros is going to be better for this usage.

Nilsia:
I have pushed a new macro which generate the commands, avoiding repetition of code when generation them.

However there are some things that I do not really like : 
 - the second state of the macro is a bit big, I would have wanted to to something like the generation of the keyboard, but I cannot call macros inside my own macro calls.
 - all functions are duplicated, however, this could be avoided with the @both pattern, with converting MouseTrie into an enum that can take both mouse commands and key command. What do you think about that ?
 - And the third one is: when generation commands that way, some mouse commands are not suffixed by _mouse, but I cannot find out examples.

I have not handled yet the current buffer focus.
If you have any modifications to suggest about the @key/mouse/both, do not hesitate to tell me.

PR #9477, opened 2024-01-30
Adding two themes using only colors from 16-color terminal themes

dgkf:
Adds two themes, using only the 16 color palette used for terminal themes. 

image (Tango Light)

This allows for the helix theme to match the terminal theme more consistently for those of us that like to have a helix pane open in a terminal multiplexer. 

The theme style choices are meant to look best for palettes that reasonably adhere to the 8-color palette + 8 lighter variants. At least in kitty, many of the themes do not adhere to this style - the "lighter" color is often darker than the base color, or the "white" and "black" colors are swapped. I prioritized accommodating the themes that adhere more closely to the spec, so some of these deviating themes sometimes have quirks.

Here are some screenshots of some themes that adhere to the general ansi palette: 

| `kitty` theme (light) | `term16_dark` | `term16_light` |
| :-- | :-: | :-: |
| Tango (Tango Light) |  |  |
| Ayu (Ayu Light) | image |  image |
| Solarized Dark* (Solarized Light) | image | image |
| Pencil Dark (Pencil Light) |image | image |

<sup>*Used "Higher Contrast" variant because the kitty default Solarized Dark uses grey for all the "light" variants</sup>


runtime/themes/term16_light.toml
```diff
@@ -0,0 +1,85 @@
+# Author: dgkf <doug.kelkhoff@gmail.com>
+# Modified from base16_terminal, Author: NNB <nnbnh@protonmail.com>
+
+inherits = "term16-dark"
```
ascarter, line 4: The name here should be `term16_dark`. Inherit is not working.

dgkf: Ah good catch - that was a late change when I realized the more common convention is to use underscores. Will make a quick fix.

ascarter (approved):

the-mikedavis (approved):

pascalkuthe (approved):

PR #9499, opened 2024-02-02
Fix rendering of popup scrollbar

### LINKED ISSUE #9498: Popup scroll track (`ui.menu.scroll.bg`) no longer being colored properly
### Summary

helix master (dbac78b):
image

helix 23.10 (f6021dd0):
image

### Platform

macOS

### Terminal Emulator

wezterm 20240128-202157-1e552d76

### Installation Method

source

### Helix Version

helix master (dbac78b)

flinesse:
Fixes #9498 and keeps #4313's intention (I believe) of repurposing the border as scrollbar track.

image

### DIFF
--- a/helix-term/src/ui/menu.rs
+++ b/helix-term/src/ui/menu.rs
@@ -419,14 +419,19 @@ impl<T: Item + 'static> Component for Menu<T> {
             for i in 0..win_height {
                 cell = &mut surface[(area.right() - 1, area.top() + i as u16)];
 
-                let half_block = if render_borders { "▌" } else { "▐" };
+                let (scroll_thumb, scroll_track) = if render_borders {
+                    ("▌", "│")
+                } else {
+                    ("▐", "▐")
+                };
 
                 if scroll_line <= i && i < scroll_line + scroll_height {
                     // Draw scroll thumb
-                    cell.set_symbol(half_block);
+                    cell.set_symbol(scroll_thumb);
                     cell.set_fg(scroll_style.fg.unwrap_or(helix_view::theme::Color::Reset));
-                } else if !render_borders {
+                } else {
                     // Draw scroll track
+                    cell.set_symbol(scroll_track);
                     cell.set_fg(scroll_style.bg.unwrap_or(helix_view::theme::Color::Reset));
                 }
             }
--- a/helix-term/src/ui/popup.rs
+++ b/helix-term/src/ui/popup.rs
@@ -295,14 +295,19 @@ impl<T: Component> Component for Popup<T> {
                 for i in 0..win_height {
                     cell = &mut surface[(inner.right() - 1, inner.top() + (border + i) as u16)];
 
-                    let half_block = if render_borders { "▌" } else { "▐" };
+                    let (scroll_thumb, scroll_track) = if render_borders {
+                        ("▌", "│")
+                    } else {
+                        ("▐", "▐")
+                    };
 
                     if scroll_line <= i && i < scroll_line + scroll_height {
                         // Draw scroll thumb
-                        cell.set_symbol(half_block);
+                        cell.set_symbol(scroll_thumb);
                         cell.set_fg(scroll_style.fg.unwrap_or(helix_view::theme::Color::Reset));
-                    } else if !render_borders {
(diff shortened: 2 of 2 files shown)


ath3:
If i see correctly this breaks the colors of borders on the right side of popups.
Also it hardcodes the right border to a specific border style.
I think i know where the issue is and how to fix it, let me make a pr and see if it works for you.

ath3:
PR is #9508

flinesse:
> If i see correctly this breaks the colors of borders on the right side of popups.

It does not break colors. I had chosen a darker .bg to illustrate the fixed behavior with a colorable scroll track. Matching `ui.menu.scroll.bg` to the popup fg text color would bring that back:

image

> Also it hardcodes the right border to a specific border style.

I'm not sure what you mean by "hardcodes"—it's simply the nature of working with the `Cell` tui primitive. If you're referring to [defining the scroll thumb and scroll track](https://github.com/helix-editor/helix/pull/9499/files#diff-503491a7f89629a9e0b81f33fb1ec6c5adaf97476f24e4d661dbd86d22cfcf59R422-R426), I think distinguishing the two elements makes sense in this context and avoids the kind of confusion mentioned [here](https://github.com/helix-editor/helix/pull/9508#issuecomment-1925508612).
<br></br>
I presumed you had intended the border-enabled scrollbar design to be this way, which is what I meant by "keeping [your] intention of **repurposing the border as scrollbar track**".

In any case, I chose to keep it because of a.) the visual harmony with borders and b.) the design had already found a certain degree of acceptance (it's what people would have been testing with so far in your original PR and master).

For reference, this is what a (bordered) scrollable popup would look like using the original style:

image

cc: @the-mikedavis

ath3:
> It does not break colors. I had chosen a darker .bg to illustrate the fixed behavior with a colorable scroll track. Matching ui.menu.scroll.bg to the popup fg text color would bring that back:

So it does break colors.
Idk which theme are you using but every theme i tried is having the same color problems as seen in your original screenshot.

> I'm not sure what you mean by "hardcodes"—it's simply the nature of working with the Cell tui primitive. If you're referring to [defining the scroll thumb and scroll track](https://github.com/helix-editor/helix/pull/9499/files#diff-503491a7f89629a9e0b81f33fb1ec6c5adaf97476f24e4d661dbd86d22cfcf59R422-R426), I think distinguishing the two elements makes sense in this context and avoids the kind of confusion mentioned https://github.com/helix-editor/helix/pull/9508#issuecomment-1925508612.

We have a default border type which is currently plain, but has four possible values.

My goal was *not* to have scroll track when borders are enabled, because its too much visually.
Coloring part of the border as scroll track looks very odd.
If for some reason we decide to have scroll tracks then it should look like it does without borders, imo.

But like i said i intentionally didnt want scroll tracks when popup borders are enabled.

flinesse:
Colors "break" because most themes do not have `ui.menu.scroll.bg` explicitly set. Try kanagawa, modus_vivendi, or nightfox, or setting your own `ui.menu.scroll.bg`. An unset theme config can typically rely on setting inheritance to give you a discernible default, but in this case we have an atypical (though understandable) definition for [`ui.menu.scroll`](https://github.com/helix-editor/helix/blame/a1272bdb17a63361342a318982e46129d558743c/book/src/themes.md#L319).

To reiterate, this PR:

- Fixes the scroll track not showing when borders are off
- Retains the original behavior of turning the wide scroll track _off_ when borders are enabled
- Allows coloring the section of border that the scroll track occupies by setting (`ui.menu.scroll.bg`)

| `popup-border` on | `popup-border` off |
| --- | --- |
| image | image |
| image | image |
| image | image |

_rose_pine with custom overrides_

pascalkuthe:
I agree with @the-mikedavis. Closing as https://github.com/helix-editor/helix/pull/9508 was merged.

the-mikedavis:
> this PR... Allows coloring the section of border that the scroll track occupies by setting `ui.menu.scroll.bg`

This is the part that I don't think is necessary. If we're not drawing the scroll track then theming it seems needless and I would prefer to see the border have a consistent color whether it's part of the scroll track or not

PR #9503, opened 2024-02-02
Use gix pipeline filter instead of manual crlf implementation

### LINKED ISSUE #8145: Git gutter shows everything changed with automatic end-of-line conversion (.gitattributes contains * text=auto)
### Summary

[Git has a feature to check out text files with native line endings](https://www.git-scm.com/docs/gitattributes#_end_of_line_conversion), something that is often used for crossplatform development.
When viewing a file on windows which has been converted to crlf this way, the git gutter shows everything changed:

image

However, when setting `:line-ending lf` it doesn't anymore:

image

This is no workaround, because this actually changed the line endings of the file to lf, where crlf is desired.
Note that `git diff` works correctly, and does not report any changes.

### Reproduction Steps

_No response_

### Helix log

_No response_

### Platform

Windows

### Terminal Emulator

wezterm 20230828-220414-99c96139

### Helix Version

helix 23.05 (19b73bc7)

ShoyuVanilla:
Fixes #8145

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -449,6 +449,18 @@ dependencies = [
 ]
 
 [[package]]
+name = "filetime"
+version = "0.2.23"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1ee447700ac8aa0b2f2bd7bc4462ad686ba06baa6727ac149a2d6277f0d240fd"
+dependencies = [
+ "cfg-if",
+ "libc",
+ "redox_syscall 0.4.1",
+ "windows-sys 0.52.0",
+]
+
+[[package]]
 name = "flate2"
 version = "1.0.27"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -527,33 +539,41 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "31887c304d9a935f3e5494fb5d6a0106c34e965168ec0db9b457424eedd0c741"
 dependencies = [
  "gix-actor",
+ "gix-attributes",
+ "gix-command",
  "gix-commitgraph",
  "gix-config",
  "gix-date",
  "gix-diff",
  "gix-discover",
  "gix-features",
+ "gix-filter",
  "gix-fs",
  "gix-glob",
  "gix-hash",
  "gix-hashtable",
+ "gix-ignore",
+ "gix-index",
  "gix-lock",
  "gix-macros",
  "gix-object",
  "gix-odb",
  "gix-pack",
  "gix-path",
+ "gix-pathspec",
  "gix-ref",
  "gix-refspec",
  "gix-revision",
  "gix-revwalk",
  "gix-sec",
+ "gix-submodule",
  "gix-tempfile",
  "gix-trace",
  "gix-traverse",
  "gix-url",
  "gix-utils",
  "gix-validate",
+ "gix-worktree",
  "once_cell",
  "parking_lot",
  "smallvec",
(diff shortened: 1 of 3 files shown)


helix-vcs/src/git.rs
```diff
@@ -76,29 +78,20 @@ impl DiffProvider for Git {
...
-            }
-            data = normalized_file
+        let data = file_object.detach().data;
+        // Get the actual data that git would make out of the git object.
+        // This will apply the user's git config or attributes like crlf conversions.
+        if let Some(work_dir) = repo.work_dir() {
+            let rela_path = file.strip_prefix(work_dir)?.to_string_lossy();
+            let (mut pipeline, _) = repo.filter_pipeline(None)?;
+            let mut worktree_outcome =
+                pipeline.convert_to_worktree(&data, rela_path.as_ref().into(), Delay::Forbid)?;
```
pascalkuthe, line 88: @Byron how does this handle encoding? Git stores everything as utf-8 internally IIRC and only converts back on checkout.

I guess that this function would do that? Could we instead checkout utf-8 somehow. (My memory is very cloudy here so I may be missremebering)

Cargo.lock
```diff
@@ -448,6 +448,18 @@ dependencies = [
  "winapi",
 ]
 
+[[package]]
+name = "filetime"
```
archseer, line 452: A lot of new dependencies :/

ShoyuVanilla: It comes from new feature flag `gix = { features = ["attributes"], .. }`, which is needed for https://docs.rs/gix/0.58.0/gix/struct.Repository.html#method.filter_pipeline 😢

helix-vcs/src/git.rs
```diff
@@ -76,29 +78,21 @@ impl DiffProvider for Git {
...
-                normalized_file.push(byte)
-            }
-            data = normalized_file
+        let data = file_object.detach().data;
+        // Get the actual data that git would make out of the git object.
+        // This will apply the user's git config or attributes like crlf conversions.
+        if let Some(work_dir) = repo.work_dir() {
+            let rela_path = file.strip_prefix(work_dir)?;
+            let rela_path = gix::path::try_into_bstr(rela_path)?;
+            let (mut pipeline, _) = repo.filter_pipeline(None)?;
```
Byron, line 87: Just as a note: It's a bit sad that the filter pipeline can't be reused under the circumstances. if this should ever be an issue that can be achieved by dropping down to the [internal plumbing types](https://github.com/Byron/gitoxide/blob/17638628586900d43d730e6ed2a0862d8e408f29/gix/src/filter.rs#L70-L72).

However, I wouldn't go there as this will seriously complicate the code - I just wanted you to know that you can to shave off some cycles.

helix-vcs/src/git.rs
Byron: That's not the case, but it's true that Git has an 'internal' form that it wants to store in the object database. This is merely a form that passed through the filter pipeline though, which on Unix is most often a no-op.

It's true though that this method takes care of transforming `data` which is supposed to be directly from the object database and turn it into what would be checked out, applying all necessary transformations.

Byron (approved):
I took my time to study the PR and believe that't the way to do it! Good work :)!

Additionally I was thinking that using the worktree-version as base for a diff is also exactly would should be done given how the user opens and edits files in the worktree.

For posterity, let me mention that by now there is also the [`repo.diff_recource_cache()`](https://github.com/Byron/gitoxide/blob/17638628586900d43d730e6ed2a0862d8e408f29/gix/src/filter.rs#L70-L72) method which allows to perform diffs just like diff does which, you guessed it, may apply even more transformations 😁. Doing so would be wrong here, but if there ever is a 'show me the git diff' function in `helix`, that would be the way to do it (*and that's particularly interesting if the user configured diff-filters to turn binary files into something diffable*).

Cargo.lock
Byron: I reviewed how this dependency is used and it seems 'convenience' is the answer. No feature is used that wouldn't be covered by the doing what the crate does.

I added a task to [the `helix` tracking issue](https://github.com/Byron/gitoxide/issues/637) to represent that possibility, and help would definitely be welcome.

pascalkuthe (approved):

the-mikedavis (approved):
Thanks!

This fixes the diff gutter for me in a repo where I use [`git-crypt`](https://github.com/AGWA/git-crypt) :rocket:

PR #9508, opened 2024-02-03
Fix scroll track

ath3:
Fix for issue #9498


the-mikedavis (approved):

pascalkuthe:
Hmm so if borders are enabled the scroll line is still hidden?

Also it seems we now have a branch ere we're the if and else if condition do the exact same thing here. That seems like codesmell to me at best a sign of a potential trial bug at worst

ath3:
Yeah, if we have borders the scroll line is hidden.

The if condition is the actual scrollbar position and is painted with scroll_style.fg.
The else condition is when the position is not the actual scrollbar position and we dont have borders enabled, and paints with scroll_style.bg.

pascalkuthe (approved):

PR #9515, opened 2024-02-03
feat(languages): pkl

matoous:
Add [pkl](https://github.com/apple/pkl) language. Official documentation: https://pkl-lang.org/. Mostly adopted from https://github.com/apple/pkl-neovim and https://github.com/apple/tree-sitter-pkl plus resolved some injections for comments.

### DIFF
--- a/runtime/queries/pkl/indents.scm
+++ b/runtime/queries/pkl/indents.scm
@@ -1,0 +1,32 @@
+; Copyright © 2024 Apple Inc. and the Pkl project authors. All rights reserved.
+;
+; Licensed under the Apache License, Version 2.0 (the "License");
+; you may not use this file except in compliance with the License.
+; You may obtain a copy of the License at
+;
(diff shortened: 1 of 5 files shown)


runtime/queries/pkl/indents.scm
```diff
@@ -0,0 +1,32 @@
...
+] @indent
+
+[
+  "("
+  ")"
+  "{"
+  "}"
+  "if"
+  "else"
+] @branch ; TODO: what does this query do?
```
the-mikedavis, line 32: I believe that `@branch` is specific to nvim's indent system. We could remove the indent query for now and rely on the default behavior until someone wants to write Helix-specific ones

matoous: Done, thanks for review.

the-mikedavis (approved):

pascalkuthe (approved):

PR #9519, opened 2024-02-04
Let gf go to file with line number

### LINKED ISSUE #9440: Let gf go to file with line number
The goto_file function doesn't handle file paths with line numbers unlike :open
On a path:
'path/to/file.txt:55'
goto_file will open an empty buffer path/to/file.txt:55, which is somewhat nonsensical. (shouldn't this just error out?)

:open path/to/file.txt:55
will instead open the correct file on the expected line.
This is the behavior I would expect from marking a path and pressing gf

scdailey:
Fixes #9440

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1215,8 +1215,8 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
                 Err(e) => {
                     cx.editor.set_error(format!("Open file failed: {:?}", e));
                 }
+            }
         }
-    }
     }
 }
 
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -1162,10 +1162,6 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     let (view, doc) = current_ref!(cx.editor);
     let text = doc.text();
     let selections = doc.selection(view.id);
-    let rel_path = doc
-        .relative_path()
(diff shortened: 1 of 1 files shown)


helix-term/src/commands.rs
```diff
@@ -1162,10 +1162,6 @@ fn goto_file_impl(cx: &mut Context, action: Action) {
     let (view, doc) = current_ref!(cx.editor);
     let text = doc.text();
     let selections = doc.selection(view.id);
-    let rel_path = doc
```
the-mikedavis, line 1165: I believe this changes the behavior of `gf` a little: the path should be interpreted as relative to the current document. Also see https://github.com/helix-editor/helix/pull/8126 which aims to change that behavior

the-mikedavis (changes requested):

helix-term/src/commands.rs
scdailey: Okay I can update it. Just to clarify, are we going with the original wording of that issue (not sure if it even applies still) or the clarification here:

https://github.com/helix-editor/helix/pull/8126#issuecomment-1712944308

Just resolve relative to the current file path no matter what?

the-mikedavis: I would prefer the behavior from Pascal's comment. I think this makes sense:

1. Check if the file exists relative to the current doc
2. Check if the file exists relative to the CWD
3. Otherwise open a new file relative to the current doc

PR #9524, opened 2024-02-04
fix division by zero when prompt completion area is too small

### LINKED ISSUE #9523: attempt to divide by zero when resizing the terminal
### Summary

similarly to #1403, the program crashes when resizing the terminal while the preview/info box is displayed
see the reproduction steps below

### Reproduction Steps

 

I tried this:

1. open a terminal
2. run helix
3. press colon to open the prompt
4. press tab to show the preview/info box above the items
6. make the terminal really small (0 x 0)

I expected this to happen:

- helix resizes accordingly

Instead, this happened:

- helix crashed

### Helix log

thread 'main' panicked at helix-term/src/ui/prompt.rs:404:34:
attempt to divide by zero

### Platform

arch linux

### Terminal Emulator

alacritty

### Installation Method

pacman

### Helix Version

helix 23.10

diegostafa:
fixes #9523

### DIFF
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -401,7 +401,7 @@ impl Prompt {
 
             let offset = self
                 .selection
-                .map(|selection| selection / items * items)
+                .map(|selection| selection / std::cmp::max(1, items * items))
                 .unwrap_or_default();
 
             surface.clear_with(area, background);
--- a/helix-term/src/ui/prompt.rs
+++ b/helix-term/src/ui/prompt.rs
@@ -393,7 +393,7 @@ impl Prompt {
             height,
         );
 
-        if !self.completion.is_empty() {
+        if completion_area.height > 0 && !self.completion.is_empty() {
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/prompt.rs
```diff
@@ -401,7 +401,7 @@ impl Prompt {
 
             let offset = self
                 .selection
-                .map(|selection| selection / items * items)
+                .map(|selection| selection / std::cmp::max(1, items * items))
```
the-mikedavis, line 404: `/` and `*` are left-to-right associative in Rust but this is grouping them right-to-left: before it was `(selection / items) * items` but now it's `selection / (items * items)`.

Let's leave this expression untouched instead and check that `completion_area.height > 0` in the `if` condition for this block. Figuring out the offset doesn't make sense when there are no items and there's no need to try to render the completion popup if we don't have space for it

the-mikedavis (changes requested):

the-mikedavis (approved):

PR #9536, opened 2024-02-05
adding asynchronous job tracking support for statusline

Kim-Dewelski:
Adding a statusline option for displaying the amount of asynchronous jobs which are scheduled. I find myself wanting this information because currently, when working on large workspaces, certain operations take a long time to process and therefore being able to see whether or not the job was successfully scheduled could help so that you're not spending time thinking about how many unfinished write commands you have ongoing, or if you correctly opened the workspace symbol picker because it takes so long to open.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -125,6 +125,7 @@ The following statusline elements can be configured:
 | `spacer` | Inserts a space between elements (multiple/contiguous spacers may be specified) |
 | `version-control` | The current branch name or detached commit hash of the opened workspace |
 | `register` | The current selected register |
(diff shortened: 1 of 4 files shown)


pascalkuthe:
This is not feasible anymkee because with the event system jobs now get spawned as tokio tasks instead of being put in a futures unordered. You can't count them anymore.

Background tasks are an implementation detail that often have nothing to do with system lands since tokio can easily hane millions of (sleeping) background tasks. Particularly with the new event system the number of background tasks increases a lot (and can cahnge based kn vsrious factors) but that has basically no impact on performance.

The way I want to handle waiting on tasks is with something like #6436

PR #9537, opened 2024-02-05
Add support for Odin's formatter

Chickenkeeper:
Odin's language server also comes with a formatter, so it would be nice to make use of it.

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -1897,13 +1897,14 @@ source = { git = "https://github.com/fvacek/tree-sitter-cpon", rev = "0d01fcdae5
 
 [[language]]
 name = "odin"
-auto-format = false
+auto-format = true
 scope = "source.odin"
 file-types = ["odin"]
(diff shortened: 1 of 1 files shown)


the-mikedavis (commented):
Does the language server support auto-formatting? If so then let's just enable `auto-format` rather than adding an external formatter config

Chickenkeeper:
The language server does support auto-formatting, however `odinfmt` is also provided as a seperate executable, so I added the external formatter config so people could use the `odinfmt` on its own if they don't want to use the full language server

the-mikedavis:
If you have an external formatter and the language server installed, we currently prefer the formatter even though it's likely less efficient than the language server computing and sending the edits (since we need to shell out and then generate the edits between the old and new text).

I think this comes up with other languages though (Zig IIRC) so I'll make a separate issue and let's merge this as-is :+1:

the-mikedavis (approved):

Issue #8820, opened 2023-11-15
Compilation in a macOS GitHub runner fails with cc v1.0.84


zydou:
https://github.com/helix-editor/helix/commit/40959bb449ec9dfbff9249332853a02b50f483fe bumps `cc` to `v1.0.84`. However, this release has a problem that causes the build on macOS to fail.

upstream issue: https://github.com/rust-lang/cc-rs/issues/900
fail logs: https://github.com/zydou/helix/actions/runs/6870927003/job/18689481341

This release has been yanked on the [release page](https://github.com/rust-lang/cc-rs/releases/tag/1.0.84):

> This release has been yanked, due to the Apple deployment target changes causing unintentional widespread breakage for building C++ code on the target (for projects without an explicit deployment target set). A release without these issues will be made in the near future.
> Sorry, and thanks.

PR #9548, opened 2024-02-06
Revert "build(deps): bump cc from 1.0.83 to 1.0.84 (#8809)"

pashadia:
The dependency `cc` version `1.0.84` was yanked. Here is the [explanation](https://github.com/rust-lang/cc-rs/releases/tag/1.0.84) on the release page.

This PR simply reverts back to the previous version of the dependency. It will also fix issue #8820 .

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -145,9 +145,9 @@ checksum = "df8670b8c7b9dae1793364eafadf7239c40d669904660c5960d74cfd80b46a53"
 
 [[package]]
 name = "cc"
-version = "1.0.84"
+version = "1.0.83"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
I was thinking that CC would cut a 1.0.85 release reverting the change and dependabot would clean this up but it's been a while since 1.0.84 was released, so let's downgrade :+1:

Agilan989 (commented):
Need to asking about for transportation to when from platform they are connecting with  account how??? Because notifications never get to if not I stopped using my account

pashadia:
> Need to asking about for transportation to when from platform they are connecting with account how??? Because notifications never get to if not I stopped using my account

I'm sorry, I don't understand your reply. @Agilan989 is this where you intended to write it?

archseer (approved):

PR #9551, opened 2024-02-06
slint: Update treesitter parser and queries

hunger:
This PR switches the Slint language tree-sitter plugin over to the official one hosted in the slint-ui organiazation. It follows closely what we are doing in https://github.com/nvim-treesitter/nvim-treesitter/pull/6027

@jrmoulton is aware and OK with this switch and started the transition on nvim-treesitter in https://github.com/nvim-treesitter/nvim-treesitter/pull/5985

Unfortunately Jared did not have the time to see this one through, so I created a fresh PR to get hings into shape faster.

This PR also registers the `slint!` macro with the rust language as an injection.

### DIFF
--- a/runtime/queries/slint/highlights.scm
+++ b/runtime/queries/slint/highlights.scm
@@ -2,6 +2,9 @@
 
 ; Different types:
 (string_value) @string
+(bool_value) @constant.builtin.boolean
+
+; Constants
 
 (escape_sequence) @constant.character.escape
 
@@ -12,29 +15,25 @@
   (easing_kind_identifier)
 ] @constant.builtin
 
-(bool_value) @constant.builtin.boolean
-
-(int_value) @constant.numeric.integer
+[
+  (int_value)
+  (physical_length_value)
+] @constant.numeric.integer
 
 [
   (float_value)
   (percent_value)
   (length_value)
-  (physical_length_value)
   (duration_value)
   (angle_value)
   (relative_font_size_value)
 ] @constant.numeric.float
-
-(simple_identifier) @variable.other
 
 (purity) @keyword.storage.modifier
 
 (function_visibility) @keyword.storage.modifier
 
 (property_visibility) @keyword.storage.modifier
-
-(animate_option_identifier) @keyword
 
 (builtin_type_identifier) @type.builtin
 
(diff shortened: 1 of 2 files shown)


runtime/queries/slint/highlights.scm
```diff
@@ -1,122 +1,125 @@
...
+  [
+    "if"
+    "else"
+  ] @keyword.conditional)
+
+(ternary_expression
+  [
+    "?"
+    ":"
+  ] @keyword.conditional)
```
the-mikedavis, line 114:
```suggestion
  "if" @keyword.control.conditional)

(if_statement
  ":" @punctuation.delimiter)

(if_expr
  [
    "if"
    "else"
  ] @keyword.control.conditional)

(ternary_expression
  [
    "?"
    ":"
  ] @keyword.control.conditional)
```

(to align with https://docs.helix-editor.com/master/themes.html#syntax-highlighting)

runtime/queries/slint/highlights.scm
```diff
@@ -126,46 +129,90 @@
...
+  ] @keyword.storage.type)
+
+(enum_definition
+  "enum" @keyword.storage.type)
+
+(for_loop
+  [
+    "for"
+    "in"
+  ] @keyword.repeat)
```
the-mikedavis, line 165:
```suggestion
  ] @keyword.control.repeat)
```

runtime/queries/slint/highlights.scm
```diff
@@ -126,46 +129,90 @@
...
+  "return" @keyword.return)
+
+(import_statement
+  [
+    "import"
+    "from"
+  ] @keyword.import)
+
+(import_type
+  "as" @keyword.import)
```
the-mikedavis, line 192:
```suggestion
  "return" @keyword.control.return)

(import_statement
  [
    "import"
    "from"
  ] @keyword.control.import)

(import_type
  "as" @keyword.control.import)
```

the-mikedavis (commented):
Thanks for updating this! I left some minor comments about the `keyword.control` captures but otherwise this looks good

the-mikedavis (dismissed):
Thanks!

the-mikedavis (approved):

Issue #9239, opened 2024-01-05
Undocumented theme property: `ui.statusline.active`


tryoxiss:
The theme property `"ui.statusline.active"` does not appear anywhere on the [themeing documentation](https://docs.helix-editor.com/themes.html), I found out it exists by guessing based off the inactive property (I was going to suggest it if it didnt exist). 

Every case of `ui.statusline`
![image](https://github.com/helix-editor/helix/assets/94419893/738e66ef-6101-433a-97da-5b378905215a)

but adding a `.a` leads to zero results
![image](https://github.com/helix-editor/helix/assets/94419893/5bafc6e0-402b-4669-a9a9-254b8d9e8077)

Netetra:
https://github.com/helix-editor/helix/blob/dbac78bb3c06717f1f83984da3035e28f39d49a5/helix-term/src/ui/statusline.rs#L52-L57
https://github.com/helix-editor/helix/blob/dbac78bb3c06717f1f83984da3035e28f39d49a5/helix-term/src/ui/editor.rs#L559-L567
Looking at these two codes, I think `ui.statusline.active` in `editor.rs` is written incorrectly as the opposite of `ui.statusline.inactive`.
(I am a newbie and have not read all the code yet...)

Netetra:
Is anyone working on this project right now? I'd like to make my first contribution.

tryoxiss:
I think since nobody has responded yet its safe to say nobody is working on it !

PR #9567, opened 2024-02-08
add ui.statusline.active documentation

Netetra:
#9239 `ui.statusline.active` added to documentation
This is my first contribution, so sorry if I'm doing something wrong.


the-mikedavis:
Actually I think the usage of this key is an oversight: https://github.com/helix-editor/helix/blob/a1272bdb17a63361342a318982e46129d558743c/helix-term/src/ui/editor.rs#L551-L567

Instead I think the fallback should work like this:

* `ui.bufferline.background` falls back to `ui.statusline.inactive`
* `ui.bufferline.active` falls back to `ui.statusline`
* `ui.bufferline` falls back to `ui.statusline.inactive` (current behavior)

The statusline itself doesn't use `ui.statusline.active` so I think we should remove that key.

PR #9587, opened 2024-02-10
docs: Document what the `diff` gutter symbol does

markstos:
Before there was no document about what the `diff` gutter is displaying
or what the colors mean.

These docs clarify it's a `git` diff and makes it easier to
cross-reference the theme if you aren't sure what the colors mean or
want to change them.

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -340,7 +340,12 @@ Currently unused
 
 #### `[editor.gutters.diff]` Section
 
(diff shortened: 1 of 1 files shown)


the-mikedavis (approved):
In the long-run I'd like to see docs for features like this live in a help command like https://github.com/helix-editor/helix/pull/997. For now though this seems like the best spot :+1:

PR #9590, opened 2024-02-10
Respect lsp definition order for code actions

estin:
Hi!

It's small changes to render completions/actions from lsp by they definition order on languages.toml

### DIFF
--- a/helix-term/src/handlers/completion.rs
+++ b/helix-term/src/handlers/completion.rs
@@ -199,7 +199,7 @@ fn request_completion(
     let mut futures: FuturesOrdered<_> = doc
         .language_servers_with_feature(LanguageServerFeature::Completion)
         .filter(|ls| seen_language_servers.insert(ls.id()))
-        .filter_map(|ls| {
+        .map(|ls| {
             let language_server_id = ls.id();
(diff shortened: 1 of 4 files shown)


poliorcetics:
Probably needs a documentation update so people know about that behavior and can rely on it

helix-term/src/handlers/completion.rs
```diff
@@ -227,8 +227,8 @@ fn request_completion(
                 }
             };
 
-            let completion_response = ls.completion(doc_id, pos, None, context).unwrap();
-            async move {
+            let completion_response = ls.completion(doc_id, pos, None, context)?;
+            Some(async move {
```
the-mikedavis, line 231: The `unwrap` is intentional here because `language_servers_with_feature(LanguageServerFeature::Completion)` should only give the language servers that will return Some from `completion`

pascalkuthe:
I have larger changes for completion in the works that will (among many other thigns) handle multi lsp completion ordering (somewhat similarly but note quite the same). 

I think the code actions changes look good so if you remove the completion changes I would be happy to merge this one.

pascalkuthe (approved):

the-mikedavis (approved):

PR #9594, opened 2024-02-11
Statusline indicator showing how many lines across all selections

mandx:
This indicator shows how many lines total are used by all selections in the current view.
* One or more selections on the same line only counts that line once
* A single cursor counts that line as well
* A selection that spans several lines will count all those lines as well.

Notes:
* The idea is, after making selections, if we run the `extend_to_line_bounds` command, how many lines do we get? That's the count I would like shown here.
* Maybe a zero width selection shouldn't be counted?
* Current implementation uses `HashSet`, maybe there's a better implementation that can use ranges only?

### DIFF
--- a/book/src/configuration.md
+++ b/book/src/configuration.md
@@ -121,6 +121,7 @@ The following statusline elements can be configured:
 | `workspace-diagnostics` | The number of warnings and/or errors on workspace |
 | `selections` | The number of active selections |
 | `primary-selection-length` | The number of characters currently in primary selection |
+| `selections-line-count` | Number of lines across all selections |
 | `position` | The cursor position |
 | `position-percentage` | The cursor position as a percentage of the total number of lines |
 | `separator` | The string defined in `editor.statusline.separator` (defaults to `"│"`) |
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -145,6 +145,7 @@ fn get_render_function<'a>(
         helix_view::editor::StatusLineElement::Diagnostics => render_diagnostics,
         helix_view::editor::StatusLineElement::WorkspaceDiagnostics => render_workspace_diagnostics,
         helix_view::editor::StatusLineElement::Selections => render_selections,
+        helix_view::editor::StatusLineElement::SelectionsLineCount => render_selections_line_count,
         helix_view::editor::StatusLineElement::PrimarySelectionLength => {
             render_primary_selection_length
         }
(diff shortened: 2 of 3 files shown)


pascalkuthe:
this feels pretty niche to be honest. I think this can be deferred to a plugin.

mandx:
Maybe others would use this differently; my most frequently case for this is:

* Open a new, blank buffer
* Paste some big text dump (usually JSON which in that case I would `:format` first)
* Select the whole buffer (`%`)
* Select pattern (hit `s` then type the pattern)
* Glance at the statusline to see how many _lines_ (not selections) match the pattern

There are workarounds of course:
* Select all; then select pattern, then hit `x` to extend selections to line boundaries; this will show how many lines are selected with the current `selections` statusline indicator.
  * This has the downside of discarding the original selections; to get them back I would have to start over the pattern selection again.
* Select all, then `:pipe` it to some command like `grep -i 'some-pattern' | wc -l`
  * The downside here is that the entire contents of the buffer is lost and replaced with just the line count; so if I want the original selections that match the pattern I have to `u`ndo then start over the pattern selection again.
* Pipe the file to the same `grep | wl` command in a separate shell
  * Besides having to shell-switch back and forth; this requires the file to be saved to disk first to work properly.

> I think this can be deferred to a plugin.

Probably, although, as I understand it, there's no telling when the plugin system will land (which it's preferable IMO, better done right than rushed) and it's not clear (to me at least) if the plugin system will allow modifications like this.

pascalkuthe:
> Select all; then select pattern, then hit x to extend selections to line boundaries; this will show how many lines are selected with the current selections statusline indicator

I don't see why you would care how many lines as opposed to selections match. I can't think of a usecase for that.

> This has the downside of discarding the original selections; to get them back I would have to start over the pattern selection again.

`s<ret>` will instantly restore the old selections if you previously created it with `s`. There is also some work on a selection history so you could undo the selection change which would cover cases like yhsi more generally. 

So your usecase is covered by doing three extra keypresses `xs<ret>`. I don't think it's worth adding an extra statuskine component for that to core.

> Probably, although, as I understand it, there's no telling when the plugin system will land (which it's preferable IMO, better done right than rushed) and it's not clear (to me at least) if the plugin system will allow modifications like this.

We are actively working towards the plugin system. It's not something hypothetical. And yes writing custom statusline components will be possible. 

I am going to close this since I don't think this is something that we will include in core. You can either use the workaround I described or keep this in your personal branch. Thank you for contributing!

Issue #6640, opened 2023-04-07
Add narrow no-break space to `editor.whitespace`


Iorvethe:
On the bépo keyboard layout, the narrow no-break space [1] is very easy to inadvertently type and having it as highlighted would be very helpful to avoid hard-to-track compile errors.

[1] https://symbl.cc/en/202F/

archseer:
Is this is the same character as `&nbsp` in HTML?

Iorvethe:
I believe it’s a different character, namely the `&nnbsp;` one.

quentin-bettoum:
I would also like to have it highlighted. I will attempt a PR.

PR #9604, opened 2024-02-11
Add narrow no-break space support

quentin-bettoum:
The goal of this PR is to support the rendering of narrow no-break spaces as asked in #6640.

Note that I'm not proficient in Rust, but since it looked simple, I gave it a try by looking at what have been done for regular `nbsp`.

The tests passes and the feature works using `cargo run`.

I don't know what's better between having a dedicated `editor.whitespace.render` option for `nnbsp` or including it in the existing `nbsp` parameter.

Currently, the new config parameters would look like this:
```toml
[editor.whitespace.render]
space = "all"
nbsp = "all"
nnbsp = "all"
tab = "all"
newline = "none"

[editor.whitespace.characters]
space = "·"
nbsp = "⍽"
nnbsp = "␣"
tab = "→"
newline = "⏎"
tabpad = "·"

```

### DIFF
--- a/helix-core/src/chars.rs
+++ b/helix-core/src/chars.rs
@@ -45,8 +45,8 @@ pub fn char_is_whitespace(ch: char) -> bool {
         '\u{0009}' | // Character Tabulation
         '\u{0020}' | // Space
         '\u{00A0}' | // No-break Space
-        '\u{202F}' | // Narrow No-break Space
         '\u{180E}' | // Mongolian Vowel Separator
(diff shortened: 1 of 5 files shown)


helix-core/src/chars.rs
the-mikedavis, line 1: Looks like this change is unrelated - could you reset this?

the-mikedavis (approved):

archseer (approved):

PR #9613, opened 2024-02-12
Add change_directory command

godalming123:
Adds a command to change the working directory to the selected text. This would make it easier to (for example) embed a file-manager in helix (https://github.com/helix-editor/helix/discussions/9607) with the following code:
```toml
[keys.normal]
C-f = [":new", ":insert-output pwd; ls -F1", ":goto 2"]
# Mode for the file picker, TODO activate when the user presses control-f in normal mode
# 3 roadblocks:
# - activate modes from keymaps
# - if statements so that their can be 1 key to open the directory *or* the file (out of scope for the config file maybe)
# - using the escape key in custom modes
[keys.normal."_"]
esc = [":buffer-close!"]
f = ["extend_to_line_bounds", "goto_file", ":buffer-previous", ":buffer-close!"]
h = [":change-current-directory ../", "select_all", "delete_selection", ":insert-output pwd; ls -F1 --indicator-style=none", ":goto 2"]
d = ["extend_to_line_bounds", "change_directory", "select_all", "delete_selection", ":insert-output pwd; ls -F1", ":goto 2"] # Changes the current working directory to the directory under the cursor
j = "move_line_down"
k = "move_line_up"
```
**Note** that this PR only has the code, documentation still needs to be written

I plan to add a command to change the current mode in a subsequent PR so that pressing control-f could automatically open the `normal."_"` mode, and from that mode when a key is pressed the mode does not have to exit because it can change the mode to itself againt (tommorow maybe)

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -5,6 +5,7 @@ pub(crate) mod typed;
 pub use dap::*;
 use helix_vcs::Hunk;
 pub use lsp::*;
+use std::fs;
 use tui::widgets::Row;
 pub use typed::*;
 
@@ -225,6 +226,7 @@ impl MappableCommand {
 
     #[rustfmt::skip]
     static_commands!(
+        change_directory, "Change the working directory to the directory in the selection",
         no_op, "Do nothing",
         move_char_left, "Move left",
         move_char_right, "Move right",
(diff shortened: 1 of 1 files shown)


kirawi:
If I'm understanding this correctly, you would be able to achieve this with https://github.com/helix-editor/helix/pull/4709 using existing commands.

the-mikedavis (changes requested):
This seems like a hack around not having macro keybindings #1383 - you would use `:cd <C-r>"<ret>` instead. We don't generally accept 'bandaid' fixes around missing features and I don't see this command being useful once macro keybindings or scripting is possible: you would use `:cd` and registers interactively (i.e. outside of keybindings).

godalming123:
See https://github.com/helix-editor/helix/discussions/9607#discussioncomment-8452490

PR #9633, opened 2024-02-15
improve handling of different URI schemes over LSP

soqb:
## Objective

[`url::Url::to_file_path`](https://docs.rs/url/latest/url/struct.Url.html#method.to_file_path) does not check which scheme the URI adheres to, meaning currently when a language server sends us a URI containing an unrecognized scheme, we blindly treat it as if it were a file path, leading to strange results.

This was causing issues for me using [csharp-ls](https://github.com/razzmatazz/csharp-language-server) since when it resolves definitions over assembly boundaries, it will return a `csharp:/metadata/foo/bar/Baz.cs` URI, which a willing client can use to get readonly, decompiled definition info. In the current version, helix opens a blank buffer at `/metadata/foo/bar/Baz.cs`.

## Solution

This PR introduces a `uri_to_file_path` method which performs the expected behaviour, first checking the scheme, and changes all relevant uses of `Url::to_file_path` to go via the new method instead and error without panicking when an unexpected scheme is found.

## Details

It appears the [LSP spec](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#uri) says nothing about which schemes are valid, so I don't think csharp-ls is at fault here (though I wouldn't expect us to support the "csharp" scheme without plugins).

> [!NOTE]
> Strangely, while csharp-ls uses its own scheme, omnisharp's URIs are of the form `file:///$metadata$/foo/bar/Baz.cs`. It seems they adhere strongly to VSCode's lsp model which lets an extension act as middleware to intercept that URI to use other behaviour. Ergo, this PR does not fix the issue for omnisharp.

### DIFF
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -68,6 +68,27 @@ pub mod util {
     use helix_core::{chars, RopeSlice, SmallVec};
     use helix_core::{diagnostic::NumberOrString, Range, Rope, Selection, Tendril, Transaction};
 
+    #[derive(Debug, Error)]
+    pub enum FilePathError {
+        #[error("unsupported scheme in URI")]
+        UnsupportedScheme,
+        #[error("unable to convert URI to file path")]
+        UnableToConvert,
+    }
+
+    /// Converts a [`Url`] into a [`PathBuf`].
+    ///
+    /// Unlike [`Url::to_file_path`], this method respects the uri's scheme
+    /// and returns `Ok(None)` if the scheme was not "file".
+    pub fn uri_to_file_path(uri: &Url) -> ::core::result::Result<PathBuf, FilePathError> {
+        if uri.scheme() == "file" {
+            uri.to_file_path()
+                .map_err(|_| FilePathError::UnableToConvert)
+        } else {
+            Err(FilePathError::UnsupportedScheme)
+        }
+    }
+
     /// Converts a diagnostic in the document to [`lsp::Diagnostic`].
     ///
     /// Panics when [`pos_to_lsp_pos`] would for an invalid range on the diagnostic.
--- a/helix-lsp/src/lib.rs
+++ b/helix-lsp/src/lib.rs
@@ -68,27 +68,6 @@ pub mod util {
     use helix_core::{chars, RopeSlice, SmallVec};
     use helix_core::{diagnostic::NumberOrString, Range, Rope, Selection, Tendril, Transaction};
 
-    #[derive(Debug, Error)]
-    pub enum FilePathError {
-        #[error("unsupported scheme in URI")]
-        UnsupportedScheme,
-        #[error("unable to convert URI to file path")]
-        UnableToConvert,
-    }
-
-    /// Converts a [`Url`] into a [`PathBuf`].
-    ///
-    /// Unlike [`Url::to_file_path`], this method respects the uri's scheme
-    /// and returns `Ok(None)` if the scheme was not "file".
-    pub fn uri_to_file_path(uri: &Url) -> ::core::result::Result<PathBuf, FilePathError> {
-        if uri.scheme() == "file" {
-            uri.to_file_path()
-                .map_err(|_| FilePathError::UnableToConvert)
-        } else {
-            Err(FilePathError::UnsupportedScheme)
-        }
-    }
-
     /// Converts a diagnostic in the document to [`lsp::Diagnostic`].
     ///
     /// Panics when [`pos_to_lsp_pos`] would for an invalid range on the diagnostic.
--- a/helix-term/src/application.rs
+++ b/helix-term/src/application.rs
@@ -3,10 +3,10 @@ use futures_util::Stream;
 use helix_core::{diagnostic::Severity, pos_at_coords, syntax, Selection};
 use helix_lsp::{
     lsp::{self, notification::Notification},
-    util::{lsp_range_to_range, uri_to_file_path},
+    util::lsp_range_to_range,
     LspProgressMap,
 };
-use helix_stdx::path::get_relative_path;
+use helix_stdx::{path::get_relative_path, uri::uri_to_file_path};
 use helix_view::{
     align_view,
     document::DocumentSavedEventResult,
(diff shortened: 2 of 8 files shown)


pascalkuthe:
Oh I always thought that method did that check automatically but apparently it does not according to the docs. Good catch

helix-lsp/src/lib.rs
```diff
@@ -68,6 +68,27 @@ pub mod util {
...
+    /// Unlike [`Url::to_file_path`], this method respects the uri's scheme
+    /// and returns `Ok(None)` if the scheme was not "file".
+    pub fn uri_to_file_path(uri: &Url) -> ::core::result::Result<PathBuf, FilePathError> {
+        if uri.scheme() == "file" {
+            uri.to_file_path()
+                .map_err(|_| FilePathError::UnableToConvert)
+        } else {
+            Err(FilePathError::UnsupportedScheme)
+        }
+    }
```
pascalkuthe, line 90: lets move this to `helix-stdx::uri`

helix-view/src/handlers/lsp.rs
```diff
@@ -72,22 +74,26 @@ impl ToString for ApplyEditErrorKind {
 }
 
 impl Editor {
+    fn uri_to_file_path(&mut self, uri: &helix_lsp::Url) -> Result<PathBuf, ApplyEditErrorKind> {
+        match uri_to_file_path(uri) {
+            Ok(path) => Ok(path),
+            Err(err) => {
+                let err = format!("{err}: {uri}");
+                log::error!("{err}");
+                self.set_error(err);
```
the-mikedavis, line 83: I'm not sure we need this error on the statusline - I think the error line in the log is enough

helix-term/src/application.rs
```diff
@@ -1127,10 +1127,10 @@ impl Application {
             ..
         } = params;
 
-        let path = match uri.to_file_path() {
+        let path = match uri_to_file_path(&uri) {
             Ok(path) => path,
             Err(err) => {
-                log::error!("unsupported file URI: {}: {:?}", uri, err);
+                log::error!("{err}: {uri}");
```
pascalkuthe, line 1133: if we are going to create a custom error method just include the URI in the error straight away so we don't have to do custom error formatting everywhere

helix-term/src/commands/lsp.rs
```diff
@@ -182,13 +186,24 @@ impl ui::menu::Item for PickerDiagnostic {
...
+                location.range.end.line as usize,
+            ));
+            Some((path.into(), line))
+        }
+        Err(FilePathError::UnsupportedScheme) => None,
+        Err(FilePathError::UnableToConvert) => unreachable!(
+            "file path should be able to be acquired from URI: {}",
+            location.uri
+        ),
+    }
```
pascalkuthe, line 206: I don't really think the panic is correct here this actually caused crashes in the past. The real fix is #7367 not sure what to do with this until then.

helix-view/src/handlers/lsp.rs
```diff
@@ -214,29 +220,36 @@ impl Editor {
...
     }
 
-    fn apply_document_resource_op(&mut self, op: &lsp::ResourceOp) -> std::io::Result<()> {
+    fn apply_document_resource_op(
+        &mut self,
+        op: &lsp::ResourceOp,
+    ) -> Result<(), ApplyEditErrorKind> {
+        fn wrap_io(err: std::io::Error) -> ApplyEditErrorKind {
+            ApplyEditErrorKind::IoError(err)
+        }
```
pascalkuthe, line 229: this should not exist. We should implement `From<io::Error> for ApplyEditErrorKind` then `?` would handle the conversion automatically

helix-view/src/handlers/lsp.rs
```diff
@@ -214,29 +220,36 @@ impl Editor {
         Ok(())
     }
 
-    fn apply_document_resource_op(&mut self, op: &lsp::ResourceOp) -> std::io::Result<()> {
+    fn apply_document_resource_op(
+        &mut self,
+        op: &lsp::ResourceOp,
+    ) -> Result<(), ApplyEditErrorKind> {
+        fn wrap_io(err: std::io::Error) -> ApplyEditErrorKind {
```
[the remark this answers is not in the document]
soqb: you're absolutely right. for some reason i was under the impression `ApplyEditErrorKind` was owned by `lsp-types`.

helix-view/src/handlers/lsp.rs
soqb: agreed although we do statusline report for all the related errors in the file, and did for the `FilePathError::UnableToConvert` case before i refactored it out.

helix-term/src/commands/lsp.rs
```diff
@@ -246,7 +261,7 @@ fn sym_picker(symbols: Vec<SymbolInformationItem>, current_path: Option<lsp::Url
             action,
         );
     })
-    .with_preview(move |_editor, item| Some(location_to_file_location(&item.symbol.location)))
+    .with_preview(move |_editor, item| location_to_file_location(&item.symbol.location))
```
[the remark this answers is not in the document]
soqb: i suppose, though i don't know which behavior is more desirable. i feel if we don't include them in the pickers people might potentially be confused about why certain (albeit inaccessible) references are hidden.

helix-term/src/commands/lsp.rs
soqb: i've not to panic here to air on the safe side, so both simply return `None`. note that if this crash is pervasive enough, #7367 isn't a complete fix since this function is also used for goto defintion and friends.

gzmorell:
I can confirm that this patch, at this moment, does not solve the problem with helix not showing lsp detected errors under Windows, while [#7367](https://github.com/helix-editor/helix/pull/7367) solves the problem.

the-mikedavis:
This handles a different kind of error than #7367 - this covers the language server sending us URIs that can't be understood as file paths while #7367 handles language servers sending us valid URIs that correspond to real paths but aren't normalized (e.g. URL encoding characters like `+`/`&`)

the-mikedavis:
There is a little overlap between how we solve the problems actually. We talked about this internally and we'd prefer to introduce a custom URI type and take care of invalid LSP URLs in `TryFrom<url::Url> for Uri`. In the long run we might want to have custom schemes (for example accessing files over networks) and that would mean throwing out the changes in #7367. But we can solve both problems with a custom URI type.

I've pushed up https://github.com/helix-editor/helix/pull/9652 re-using most of this work

soqb:
closing in favour of #9652.

PR #9634, opened 2024-02-15
Bump tree-sitter-sql

### LINKED ISSUE #9550: Crash when rendering certain SQL files
### Summary

Helix crashes with a "buffer overflow detected" when rendering certain SQL files. This happens on the latest master, but was not an issue in the previous version of helix I was using, which is a custom branch based off of commit `c0fd8bc61b4c1611a48312938aaf0e3121f393b1` from September 6th.

### Reproduction Steps

I created a directory with the following sql file:

```sql
CREATE OR REPLACE FUNCTION trigger_updated_at ()
    RETURNS TRIGGER
    AS $$
BEGIN
    NEW.updated_at = NOW();
    RETURN NEW;
END;
$$
LANGUAGE plpgsql;
```

I opened `hx` in this directory, and opened the file picker. As soon as that sql file was highlighted, helix crashed:
2024-02-06-073046-s

### Helix log

No log entries are created.

### Platform

Linux ubuntu 6.2.0-39-generic #40~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Thu Nov 16 10:53:04 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Terminal Emulator

alacritty 0.13.1

### Installation Method

compiled nix flake from master

### Helix Version

helix 23.10 -- commit 5c567f31e236134c0e27dc689c8ad07ef5f9b5d1

ds-cbo:
Fixes #9550


runtime/queries/sql/highlights.scm
```diff
@@ -24,20 +24,20 @@
 (term
   alias: (identifier) @variable.parameter)
 
-(term
+((term
    value: (cast
     name: (keyword_cast) @function.builtin
-    parameter: [(literal)]?))
+    parameter: [(literal)]?)))
```
ds-cbo, line 30: I don't think these parentheses have any function, but I copied them from upstream to have as little difference as possible, makes `diff` updates easier

(same for the extra leading space a few lines down)

languages.toml
```diff
@@ -1813,7 +1813,7 @@ injection-regex = "sql"
 
 [[grammar]]
 name = "sql"
-source = { git = "https://github.com/DerekStride/tree-sitter-sql", rev = "25be0b8f17e9189ad9e1b875869d025c5aec1286" }
+source = { git = "https://github.com/DerekStride/tree-sitter-sql", rev = "4ffda1fe6841bcb2bfedb4a465c36ef8db0079b3" }
```
the-mikedavis, line 1816: tree-sitter-sql doesn't check in the generated files like `parser.c` into its main branch. You can use the latest revision from the `gh-pages` branch instead https://github.com/DerekStride/tree-sitter-sql/tree/gh-pages

ds-cbo: ah, right, I keep forgetting that

the-mikedavis (approved):
Thanks!

archseer (approved):

Issue #4956, opened 2022-12-01
No syntax highlighting in preview on global search


xJonathanLEI:
### Summary

The global search picker fails to show syntax highlighting when it's first shown. It works fine after navigating through the items.

### Reproduction Steps

 

I tried this:

1. Set up the file and open Helix:
    ```
    $ mkdir preview-bug && cd preview-bug
    $ printf "const a = 1;\nconst b = 2;\n" > test.ts
    $ hx -c /dev/null .
    ```
2. Quit the initial file picker: `<ESC>`
3. Do a global search: `<SPACE>/const<ENTER>`

I expected this to happen:

The file content on the preview is properly highlighted (after the idle timeout).

Instead, this happened:

The preview is not highlighted.

### Helix log

<details><summary>~/.cache/helix/helix.log</summary>

```
2022-12-01T16:00:47.682 helix_loader [DEBUG] Located configuration folders: []
2022-12-01T16:00:47.690 helix_view::clipboard [DEBUG] Using tmux to interact with the system clipboard
[...]

Issue #9629, opened 2024-02-14
Refactor ad-hoc hooks to use the event system


the-mikedavis:
This is meant to be a tracking issue for refactoring existing ad-hoc hooks to use the event system introduced in #8021.

- [x] LSP completion (#8021)
- [x] LSP signature help (#8021)
- [ ] LSP inlay hints
- [ ] Diff gutter
- [ ] LSP document change notifications
- [ ] Picker preview highlighting
- [ ] DynamicPicker callback

pascalkuthe:
we may also want to update mapping diagnostic change (and inlay hint mapping and so forth) to use the event system. Really most of the custom code in `apply_impl` could be moved there.

PR #9638, opened 2024-02-15
Use an AsyncHook for picker preview highlighting

the-mikedavis:
The picker previously used the IdleTimeout event as a trigger for syntax-highlighting the currently selected document in the preview pane. This is a bit ad-hoc now that the event system has landed and we can refactor towards an AsyncHook (like those used for LSP completion and signature-help). This should resolve some odd scenarios where the preview did not highlight because of a race between the idle timeout and items appearing in the picker.

Connects https://github.com/helix-editor/helix/issues/9629
Fixes https://github.com/helix-editor/helix/issues/4956

### DIFF
--- a/helix-term/src/ui/picker.rs
+++ b/helix-term/src/ui/picker.rs
@@ -10,9 +10,11 @@ use crate::{
         EditorView,
     },
 };
-use futures_util::{future::BoxFuture, FutureExt};
+use futures_util::future::BoxFuture;
+use helix_event::AsyncHook;
 use nucleo::pattern::CaseMatching;
 use nucleo::{Config, Nucleo, Utf32String};
+use tokio::time::Instant;
 use tui::{
     buffer::Buffer as Surface,
     layout::Constraint,
@@ -30,6 +32,7 @@ use std::{
         atomic::{self, AtomicBool},
         Arc,
     },
+    time::Duration,
 };
 
 use crate::ui::{Prompt, PromptEvent};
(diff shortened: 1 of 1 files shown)


helix-term/src/ui/picker.rs
```diff
@@ -917,6 +843,8 @@ impl<T: Item + 'static + Send + Sync> Component for Picker<T> {
             }
         }
 
+        helix_event::send_blocking(&self.preview_highlight_handler, ());
```
pascalkuthe, line 846: I am not too sure how we handle this currently but this may actually not be sufficient. If the matcher is still running in the background we may stream in new files. I didn't realize this when working on nucleo. What do you think of sending an event every time we render and don't have a preview instead?

pascalkuthe: we only did this here in the past because we only had access to the compositor here to reset the idle timeout

helix-term/src/ui/picker.rs
```diff
@@ -949,6 +877,108 @@ impl<T: Item> Drop for Picker<T> {
...
+            if doc.language_config().is_some() {
+                return;
+            }
+
+            let Some(language_config) = doc.detect_language_config(&editor.syn_loader.load()) else { return };
+            doc.language = Some(language_config.clone());
+            let text = doc.text().clone();
+            let loader = editor.syn_loader.clone();
+
+            tokio::spawn(async move {
```
pascalkuthe, line 932: this still needs to be spawn_blocking. We shouldn't be CPU intense stuff (which parsing definitely is) in an async job. Using job::dispatch_blocking isn't much of an issue. In theory you could also use block in place but that probably has more overhead than just using dispatch_blocking (dispatch almost never blocks anyway)

the-mikedavis:
I've included this in https://github.com/helix-editor/helix/pull/9647 so I will close in favor of that. (However this change could be ported out very easily if desired.) The only change on https://github.com/helix-editor/helix/pull/9647 is that it's moved to a `helix-term/src/ui/picker/handlers.rs` file shared by the dynamic picker handler

PR #9639, opened 2024-02-15
ci: add labeler action to automate labeling of common cases

dundargoc:
This is purely for convenience to avoid needing to label common file
changes. The labeling is only done when opening and not when updating
the PR as empirically it's more annoying to keep removing an unwanted
label each time a PR is updated, rather than adding a wanted label once.


the-mikedavis:
Thanks for opening this but I'd prefer we do this manually. The location of the changes is not always a good indicator for the area of the PR and IMO it's more personal to attach labels manually

PR #9640, opened 2024-02-16
Add a yank diagnostic command

### LINKED ISSUE #9527: Add command to yank diagnostic under cursor
https://github.com/helix-editor/helix/pull/3786, adds a yank shortcut to the picker, but if my cursor is already over the diagnostic in question, I'd like a shortcut to just yank that without finding it in the picker.

I suppose the cursor could be over multiple diagnostics, but this is rare. In that case I think any of the following would be fine:

1. Yank all diagnostics joined
2. Open a picker
3. Open a dropdown

kirawi:
Resolves #9527

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -388,6 +388,7 @@ impl MappableCommand {
         yank_main_selection_to_clipboard, "Yank main selection to clipboard",
         yank_joined_to_primary_clipboard, "Join and yank selections to primary clipboard",
         yank_main_selection_to_primary_clipboard, "Yank main selection to primary clipboard",
+        yank_diagnostic, "Yank diagnostic(s) under primary cursor to register, or clipboard by default",
         replace_with_yanked, "Replace with yanked text",
         replace_selections_with_clipboard, "Replace selections by clipboard content",
         replace_selections_with_primary_clipboard, "Replace selections by primary clipboard",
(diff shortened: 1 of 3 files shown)


helix-term/src/commands.rs
```diff
@@ -5519,6 +5520,32 @@ fn increment_impl(cx: &mut Context, increment_direction: IncrementDirection) {
...
+    let (view, doc) = current_ref!(cx.editor);
+    let primary = doc.selection(view.id).primary();
+
+    let default_sep = doc.line_ending.as_str();
+    let diag = doc
+        .diagnostics()
+        .iter()
+        .filter(|d| primary.overlaps(&helix_core::Range::new(d.range.start, d.range.end)))
+        .map(|d| d.message.clone())
+        .fold(String::new(), |mut acc, fragment| {
```
the-mikedavis, line 5533: I wonder if we should just collect this into a Vec instead: the `+` register will join the values before copying to the clipboard and it could be useful to have each diagnostic be a separate entry in the register (for example you could paste just the first value if there are duplicated diagnostics)

kirawi: I've made the change, but I don't think we currently have a way to paste a value by ordinal count.

helix-term/src/commands.rs
```diff
@@ -5519,6 +5520,24 @@ fn increment_impl(cx: &mut Context, increment_direction: IncrementDirection) {
...
+    let (view, doc) = current_ref!(cx.editor);
+    let primary = doc.selection(view.id).primary();
+
+    let diag = doc
+        .diagnostics()
+        .iter()
+        .filter(|d| primary.overlaps(&helix_core::Range::new(d.range.start, d.range.end)))
+        .map(|d| d.message.clone());
+    let reg = cx.register.unwrap_or('+');
+    match cx.editor.registers.write(reg, diag.collect()) {
```
the-mikedavis, line 5533: Let's collect in the `let diag` above and record the length of the vec before `write` so we can improve the success message, like we have for `yank_impl`:

```rust
Ok(_) => cx.editor.set_status(format!(
    "Yanked {n} diagnostic{} to register {reg}",
    if n == 1 { "" } else { "s" }
)),
```

That should make it clearer when you're copying multiple diagnostics

the-mikedavis (dismissed):

zetashift (approved):
My Rust knowledge is limited, but from I've had no problems running this branch and the codechange is small but readable.

Easy diagnostics sharing is essential for mewhen communicating errors and filing tickets, so thank you very much for this!

pascalkuthe (dismissed):

kirawi:
Sorry about that, I forgot to change some uses of `set_error` to `bail`. It's finished now.

pascalkuthe:
looks like you need to commit docgen

the-mikedavis (approved):

pascalkuthe (approved):

adriangalilea:
Sorry to ask here, but how is the yank diagnostic command performed?

the-mikedavis:
It's a typable command so you can use it command mode, similar to `:write`. In normal mode use `:` to enter command mode then type `yank-diagnostic` and hit Enter

adriangalilea:
I could not find it in the [docs](https://docs.helix-editor.com/commands.html), this is why I ask :)

kirawi:
This will be the in the upcoming release.

PR #9643, opened 2024-02-16
feat: Add Jujutsu diff handling to Helix

poliorcetics:
[Jujutsu (`jj`)](https://github.com/martinvonz/jj) is a new change-based VCS (whereas `git` is branch-based).

In this PR, I add the ability for helix to get the diffs and current head, behind a feature called `jujutsu` that is active by default. That makes it the first editor I know off that handles such that VCS!

To handle all current, future and private backends (Google already has one I believe), I instead made it so Helix can behave as a diff tool for Jujutsu and then use that as a subcommand to get the diff base.

For the head, I simply used the templating system to extract all relevant informations.

Jujutsu has a library, called [`jj-lib`](https://docs.rs/jj-lib) but it's not ready for use in third party programs and wouldn't fix the issue of custom backends anyway.

# Testing

Since testing needs `jj` installed, I haven't written any for Helix yet to discuss how to do it:

1. Ignore the issue. I don't like it much, but it doesn't cost CI time and doesn't need any change. For a feature that probably won't be used much yet, that's probably ok for a time
2. Install `jj` in CI and use it to test the new feature

### DIFF
--- a/helix-term/Cargo.toml
+++ b/helix-term/Cargo.toml
@@ -13,10 +13,14 @@ repository.workspace = true
 homepage.workspace = true
 
 [features]
-default = ["git"]
+default = ["vcs"]
 unicode-lines = ["helix-core/unicode-lines", "helix-view/unicode-lines"]
 integration = ["helix-event/integration_test"]
+
+# All VCSes available for diffs in Helix
+vcs = ["git", "jujutsu"]
 git = ["helix-vcs/git"]
+jujutsu = ["helix-vcs/jujutsu"]
 
 [[bin]]
 name = "hx"
--- a/helix-vcs/Cargo.toml
+++ b/helix-vcs/Cargo.toml
@@ -24,9 +24,13 @@ imara-diff = "0.1.7"
 anyhow = "1"
 
 log = "0.4"
+# For `jujutsu`
+tempfile = { version = "3.10", optional = true }
 
 [features]
 git = ["gix"]
+jujutsu = ["tempfile"]
+
 
 [dev-dependencies]
 tempfile = "3.13"
--- a/helix-vcs/src/jujutsu.rs
+++ b/helix-vcs/src/jujutsu.rs
@@ -1,0 +1,222 @@
+//! Jujutsu works with several backends and could add new ones in the future. Private builds of
+//! it could also have private backends. Those make it hard to use `jj-lib` since it won't have
+//! access to newer or private backends and fail to compute the diffs for them.
+//!
+//! Instead in case there *is* a diff to base ourselves on, we copy it to a tempfile or just use the
+//! current file if not.
+
+use std::path::Path;
+use std::process::Command;
+use std::sync::Arc;
+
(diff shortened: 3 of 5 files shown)


helix-term/src/main.rs
```diff
@@ -114,6 +114,18 @@ FLAGS:
...
+        match std::fs::copy(diff_base.base, &diff_base.target) {
+            Ok(_) => (),
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+                std::fs::copy(diff_base.current, diff_base.target)?;
+            }
+            Err(err) => return Err(err.into()),
+        }
+        return Ok(0);
+    }
+
```
pascalkuthe, line 128: this could be done in a wrapper script I don't think that belongs in core

poliorcetics: You mean in a personal script for it ? it would work yes, I made my tests simply using `cp` but it means relying on another third-party binary.

I can change it to use `cp` on Unix and `copy.exe` on windows if wanted

pascalkuthe: yeah I mant as in a personal script not upstream in this repo. `cp`/`copy.exe` is available everywhere I don't think its an issue relying on that.

poliorcetics: That'll simply the diff, will do!

poliorcetics: Done

maan2003:
would be cool to automatically run some `jj` command every save, so jj records the working copy. then we can run `jj undo` to revert a save

poliorcetics:
I like the idea, but probably for another MR ? It's more involved and would need some configuration changes to ensure people not using `jj` are not paying the cost on each save 

A more general "run X on save" could be very nice.

---

Note: `jj` already supports `watchman` (https://martinvonz.github.io/jj/v0.14.0/config/#watchman), you can use that instead for now

mjm:
It would be nice if the Jujutsu provider was tried before Git, so that it was used instead of Git when using a colocated repository. Right now, I find the Jujutsu provider is only used for non-colocated repos.

poliorcetics:
> It would be nice if the Jujutsu provider was tried before Git, so that it was used instead of Git when using a colocated repository. Right now, I find the Jujutsu provider is only used for non-colocated repos.

IMO no, at least no by default, the idea being people not using JJ shouldn't be paying the cost for it on each opened file.

I have https://github.com/helix-editor/helix/pull/9951 open instead and the plan is to make the diff source discovery order configurable, so that people using JJ can put it first without impacting anyone else :)

ShaddyDC:
May I ask why this was closed?

poliorcetics:
I don't like maintaining MRs without response for months.

I wanted to instead make an MR that builds on my other one in https://github.com/helix-editor/helix/pull/9951, but I never got around to it, I'll post a message here when I publish it

poliorcetics:
Done in https://github.com/helix-editor/helix/pull/12022 :)

PR #9651, opened 2024-02-17
themes: add monokai soda

jzelinskie:
I've been using this theme for about a year and haven't really touched it in a while, so I consider it pretty stable.

Monokai Soda is a fairly old theme around from the TextMate era. It's not as popular, but exists for every text editor, shell, etc that I've used.

Some screenshots:

Screenshot 2024-02-17 at 1 28 43 PM
Screenshot 2024-02-17 at 1 28 21 PM
Screenshot 2024-02-17 at 1 29 39 PM

### DIFF
--- a/runtime/themes/monokai_soda.toml
+++ b/runtime/themes/monokai_soda.toml
@@ -1,0 +1,120 @@
+# Monokai Soda port for Helix (https://helix-editor.com)
+# Author : Jimmy Zelinskie <jimmy@zelinskie.com>
+
+# Syntax
+
+## Constants
+"constant" = "white"
+"constant.builtin" = "pink"
+"constant.character.escape" = "blue"
+"constant.numeric" = "purple"
+
+## Diagnostics
+"diagnostic" = { modifiers = ["underlined"] }
+"diagnostic.error" = { underline = { style = "curl", color = "pink" } }
+"diagnostic.warning" = { underline = { style = "curl", color = "orange" } }
+"diagnostic.info" = { underline = { style = "curl", color = "white" } }
+
+## Diffs
+"diff.plus" = "green"
+"diff.delta" = "orange"
+"diff.minus" = "pink"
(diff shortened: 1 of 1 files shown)


runtime/themes/monokai_soda.toml
```diff
@@ -0,0 +1,122 @@
...
+
+## Statusline
+"ui.statusline" = { bg = "darkgray" }
+"ui.statusline.inactive" = { fg = "white", bg = "darkgray" }
+"ui.statusline.normal" = { fg = "white", bg = "blue" }
+"ui.statusline.insert" = { fg = "white", bg = "green" }
+"ui.statusline.select" = { fg = "white", bg = "purple" }
+
+"ui.text.focus" = { fg = "yellow", modifiers = ["bold"] }
+"ui.virtual.whitespace" = "darkgray"
```
the-mikedavis, line 105:
```suggestion
"ui.virtual" = "darkgray"
```

Other than rulers, the other `ui.virtual` scopes all benefit from faint foreground colors like this. I'm suggesting this just so that others with inlay-hints for example get sane colors for those (though you might not use inlay hints yourself)

runtime/themes/monokai_soda.toml
```diff
@@ -0,0 +1,122 @@
...
+"diff.minus" = "pink"
+
+## Functions
+"function" = "green"
+"function.macro" = "blue"
+"function.builtin" = "pink"
+"constructor" = "blue"
+
+## Git
+"git.delta.moved" = "orange"
```
the-mikedavis, line 30:
```suggestion
"diff.delta.moved" = "orange"
```

we don't have anything namespaced under `git` at the moment. I see this key in the nord theme too but I think it's a type for this key. `diff.delta.moved` is used in git commit message highlighting for files that were renamed

the-mikedavis (approved):

pascalkuthe (approved):

PR #9653, opened 2024-02-17
Add tmux.conf glob as a bash file type

vincevargadev:
Tmux and tmux.conf are used widely among Helix users. Having the tmux.conf file not have any syntax highlighting by default is (IMO) not ideal for an editor that otherwise "just works". Feel free to close without merge if you disagree, but I thought it's better I bring this up and give this change a chance to be merged.

## Before

Screenshot 2024-02-17 at 21 10 13

## After

Screenshot 2024-02-17 at 21 11 12

### DIFF
--- a/languages.toml
+++ b/languages.toml
@@ -860,7 +860,7 @@ file-types = [
   "tcshrc",
   "bashrc_Apple_Terminal",
   "zshrc_Apple_Terminal",
-  "conf",
+  { glob = "tmux.conf" },
   { glob = ".bash_history" },
   { glob = ".bash_login" },
   { glob = ".bash_logout" },
--- a/languages.toml
+++ b/languages.toml
@@ -860,6 +860,7 @@ file-types = [
   "tcshrc",
   "bashrc_Apple_Terminal",
   "zshrc_Apple_Terminal",
+  "conf",
   { glob = ".bash_history" },
   { glob = ".bash_login" },
   { glob = ".bash_logout" },


filipwiech:
Won't this break the support for the HOCON highlighting, which is defined later in the `languages.toml` configuration and already uses the `*.conf` file extension? Or will it somehow work together? :thinking:

languages.toml
```diff
@@ -860,6 +860,7 @@ file-types = [
   "tcshrc",
   "bashrc_Apple_Terminal",
   "zshrc_Apple_Terminal",
+  "conf",
```
the-mikedavis, line 863:
```suggestion
  { glob = "tmux.conf" },
```

`conf` as a file-type is already taken by another language but we can specifically match `tmux.conf` as a full filename here instead for the same effect

dead10ck:
`conf` is unfortunately used by many, many applications. Just off the top of my head, some others are:

* SystemD for ini
* nginx for its own language
* Salt for yaml
* logrotate for its own language

I don't really think it should be the default for any language, since it is used in so many different contexts. If you folks would like it to be the default for their config language of choice, they should just make a local `languages.toml` that does this.

pascalkuthe:
Yeah I was going to say something similar. If we already have a language for .conf it should be removed.  Adding a specific glob pattern like tmux.conf as Mike said is fine tough I think

webdev23:
In the particular case of tmux.conf, what we can do is name it `tmux.conf.sh`, so we get the bash syntax highlighting.
Then, tmux will understand just fine: `tmux source myconfig.conf.sh` without the need to be an executable file.

It's the same for many tools that use the .conf format, in reality it does not matters.

Now, what i read in this title is not that. What could be great is the ability to source Helix configs, just as shown here, as a regular Linux environment. That would allow to do cool things, without going in details, but for example to source text utilities and their shortcuts as a minor mode. It's another approach of a plugin system. Of course that is a Unix thing only.

the-mikedavis (approved):
Thanks!

pascalkuthe (approved):

Issue #9656, opened 2024-02-18
`TriggerKind::Auto` violates `lsp` spec 


nkitsaini:
For autocomplete triggers due to actions like Change mode to Insert, Idle Timeout etc, we send wrong [CompletionTriggerKind](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionTriggerKind) to language servers. And following that [CompletionContext](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#completionContext) is also invalid.

### According to spec, these should be the values of `CompletionContext`:
```js
// 24x7 trigger, Manual Trigger (Ctrl-X) etc.
{triggerKind: 1} // Invoked

// Trigger due to trigger char
{triggerKind: 2, triggerCharacter: ">"} //  TriggerCharacter

// Trigger due to previously incomplete completion list
{triggerKind: 3}
```

### Violation by `helix`
For Auto Triggers (like after typing non-trigger characters) we send `triggerKind: 2` instead of `triggerKind: 1`. Due to this we also end up sending `triggerKind: 2` without a `triggerCharacter`.

Sidenote: We don't seem to use `triggerKind: 3`. Ideally this should be used for incomplete completions but I don't understand this part quiet well so not sure if this a violation or not.
 
### Asciinema
- https://asciinema.org/a/naZB2MlsH6MrERB4e5cEIsD99
	- First, I try to use current implementation of helix with `svelteserver`, resulting in no completion. And logs showing that we send invalid `CompletionContext = {triggerKind: 2}`
	- Then I try to use a modified implementation by sending `{triggerKind: 1}`, which results in completion list by `svelteserver`.
	    - Note that `{` is not a trigger character of `svelteserver`, the autocomplete was triggered after I typed 2 characters.
    - Please ignore the `copilot` completions.

the-mikedavis:
Incomplete completions are not currently implemented but I believe @pascalkuthe was interested in looking at adding that since it's widely used.

the-mikedavis:
From the spec:

```ts
/**
 * Contains additional information about the context in which a completion
 * request is triggered.
 */
export interface CompletionContext {
	/**
	 * How the completion was triggered.
	 */
	triggerKind: CompletionTriggerKind;

	/**
	 * The trigger character (a single character) that has trigger code
	 * complete. Is undefined if
	 * `triggerKind !== CompletionTriggerKind.TriggerCharacter`
	 */
[...]

PR #9660, opened 2024-02-18
Fix LSP `ComplitionTriggerKind` value for `TriggerKind::Auto`

nkitsaini:
Fixes: https://github.com/helix-editor/helix/issues/9656


the-mikedavis (dismissed):

pascalkuthe:
https://github.com/helix-editor/helix/issues/9656#issuecomment-1951383471

pascalkuthe (changes requested):
ok the vscode code is pretty opaque but you are right that indeed we should be sending invoked for word completions. I find that a bit odd but oh well. The implementation is not correct tough. The completion debouncer runs in a background task/thread. Even if the completion was originally caused by autocompletion we could now be at a trigger char (or not anymore). 

So we should not be looking at `trigger.kind` beyond checking if it is set to manual. You should determine whether to use `INVOKED` or `TRIGGER_CHARACTER` by checking if `trigger_char.is_some()`

nkitsaini:
> ok the vscode code is pretty opaque

Yeah, I tried to go through it yesterday night but wasn't very confident. But I was to able to confirm the behavior by enabling debug logs for both `svelte` and `rust-analyzer` plugin in vscode and looking at LSP message logs.

> So we should not be looking at trigger.kind beyond checking if it is set to manual. You should determine whether to use INVOKED or TRIGGER_CHARACTER by checking if trigger_char.is_some()

Makes sense, updated the PR.

the-mikedavis (approved):

pascalkuthe (approved):

PR #9664, opened 2024-02-19
Enable backtraces by default if RUST_BACKTRACE is not set

tgross35:
This was inspired because I occasionally run across a pretty rare error that unfortunately isn't very useful on its own:

```
thread 'main' panicked at 'Positions [(9997, After), (9998, Before)] are out of range for changeset len 8910!', helix-core/src/transaction.rs:461:9
```

Having backtraces enabled by default should make it easier to get more information about a bug that cannot be easily reproduced, without requiring users to configure the `RUST_BACKTRACE` environment variable themselves.

### DIFF
--- a/helix-term/src/main.rs
+++ b/helix-term/src/main.rs
@@ -34,6 +34,11 @@ fn setup_logging(verbosity: u64) -> Result<()> {
 }
 
 fn main() -> Result<()> {
+    // Print backtraces by default
+    if std::env::var_os("RUST_BACKTRACE").is_none() {
+        std::env::set_var("RUST_BACKTRACE", "1");
+    }
+
     let exit_code = main_impl()?;
     std::process::exit(exit_code);
 }


pascalkuthe:
This feels quite hacks. Generally setting environment variables is not a good idea since it's not really thread save. It's probably OK here but I would still rather see a proper crash handler like [human-panic](https://github.com/rust-cli/human-panic).

That crate has too many dependencies for my liking and I don't love that is serialize as toml instead of priting a normal file (I don't my crash reports to be machine readable).

I forked human panic in another project of mine a while ago to arrest these things. Adding something like that to helix could be good

tgross35:
My understanding of the thread safety concern is is only unsafe to write environment variables when you have observers, meaning there should be no problems with writing before any new threads are created (based on https://doc.rust-lang.org/std/env/fn.set_var.html#safety). I believe this is a known pattern for self-enabling backtraces as well.

Another option is to use `catch_unwind` then do a backtrace capture manually. This wouldn't do anything for panics in other threads that aren't synchronized.

The tree for `human-panic` doesn't look _terrible_, the only non-optional dependencies are:

- `uuid` + v4 (new), depends on `getrandom` (already in Cargo.lock)
- `serde` + derive (already in Cargo.lock)
- `os_info`, depends on `serde` and `log` (both already in Cargo.lock)
- `backtrace`, which has a deeper tree but is used by `std` 

So that isn't too bad - the delta is just `uuid` and `os_info`, which could be useful elsewhere, and backtrace which is already built in somehow..

pascalkuthe:
I don't think we will go forward with this and instead should look into a panic handler that generatse a crash report. Thank you for contributing!

PR #9668, opened 2024-02-19
Use a hook for resolving completion items

the-mikedavis:
Previously we used the IdleTimeout event to trigger LSP `completion/resolveItem` requests. We can now refactor this to use an event system hook instead and lower the timeout.

Connects #9629

### DIFF
--- a/helix-term/src/ui/completion.rs
+++ b/helix-term/src/ui/completion.rs
@@ -1,7 +1,9 @@
 use crate::{
     compositor::{Component, Context, Event, EventResult},
     handlers::trigger_auto_completion,
+    job,
 };
+use helix_event::AsyncHook;
 use helix_view::{
     document::SavePoint,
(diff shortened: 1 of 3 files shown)


helix-term/src/ui/completion.rs
```diff
@@ -599,3 +552,88 @@ impl Component for Completion {
...
+        if self
+            .trigger
+            .as_ref()
+            .is_some_and(|trigger| trigger == &item)
+        {
+            timeout
+        } else {
+            self.trigger = Some(item);
+            self.request = None;
+            Some(Instant::now() + Duration::from_millis(150))
```
univerz, line 590: I understand that you don't want to introduce 10 different settings, but what about `low_latency = true` which will set all these annoying debounces to zero?

dead10ck (approved):
Looks good to me, thanks!

PR #9677, opened 2024-02-20
Extend groovy support

britter:
- Use more complete parser introduced in nvm-treesitter in
  https://github.com/nvim-treesitter/nvim-treesitter/commit/d4dac523d2546afc266eb9b5a7986690b5319c41
- Use groovy support when working with *.gradle files


runtime/queries/groovy/locals.scm
```diff
@@ -0,0 +1,6 @@
+(function_definition) @local.scope
+
+(parameter
+  name: (identifier) @local.definition.parameter)
```
the-mikedavis, line 4:
```suggestion
  name: (identifier) @local.definition)
```

we only support `local.scope`, `local.definition` and `local.reference` in locals queries

runtime/queries/groovy/indents.scm
the-mikedavis, line 1: These look like nvim indent queries. They will need to be rewritten to fit our indentation system https://docs.helix-editor.com/master/guides/indent.html, or you can remove them

britter: Yes, I just copied this from the nvim-treesitter repository. I'd love to have good support for groovy, but I'm not sure I can pull this off. Theoretically the indent queries should be close to the java queries since Groovy is a superset of Java. Maybe we can try just using the queries from Java and let the community improve from there?

the-mikedavis: Typically you can't re-use queries from another language since the depend on how the parser is written (i.e. which nodes exist and which nodes are possible to be children of others). For now we can just remove the `indents.scm` and rely on the built-in indentation. Then indents can be contributed later if anyone is interested in working on them

runtime/queries/groovy/highlights.scm
the-mikedavis, line 1: I meant to leave a comment on this file but I think it got lost: the captures here need to be updated to match the ones we use https://docs.helix-editor.com/master/themes.html#syntax-highlighting

For example `@boolean` becomes `@constant.builtin.boolean`. The `@spell` and `@nospell` captures can be removed

runtime/queries/groovy/indents.scm
britter: I removed the file.

runtime/queries/groovy/highlights.scm
britter: I've tried my best to get this file right, PTAL.

the-mikedavis (commented):
The queries look good but you will need to merge master / rebase and resolve the conflict and also run `cargo xtask docgen` and commit the changes

britter:
@archseer do you have some time to check this? I'd really like to get this merged 💚

pascalkuthe (approved):

PR #9691, opened 2024-02-21
Consistent statusline padding

### LINKED ISSUE #8061: Are the status line elements spaced inconsistently?
Issue #6408 was closed but I think it needs to be addressed.

Most of the status line elements are rendered with a space on either side, which eliminates the need for the spacer element, since two spaces will separate each item. This is a good design, since some elements can me missing (version-control, for example). If elements can be missing, most people probably don't want multiple spacers stacking up.

Is there a way this can be made more consistent? Perhaps add the built-in spacing to the elements that don't include it yet:
- version-control
- spinner
- file-modification-indicator

The mode element is a special case since it has a background color. Adjacent to the mode element would be a good place to put the spacer element.

hamrik:
Closes #8061.

Statusline elements were inconsistent in their use of spacing: some had margins, some didn't, some reserved space for when they were not in use while others did not.

This PR adjusts each element to act consistently, without involving the statusline renderer:
 - Every statusline element hides when not in use.
 - Every statusline element has a single space margin.

This creates a 2 space gap between every element, avoiding confusion with spaces between words, and a 1 space padding from the statusline edges.

Since elements handle their own padding, special cases like `color-modes`, `spacer` and `separator` work correctly.

### DIFF
--- a/helix-term/src/ui/statusline.rs
+++ b/helix-term/src/ui/statusline.rs
@@ -75,20 +75,6 @@ pub fn render_statusline<'a>(context: &mut RenderContext, width: usize) -> Spans
         .flat_map(|render| render(context).0)
         .collect::<Vec<Span>>();
 
-    let element_ids = &config.statusline.left;
-    let mut left = element_ids
-        .iter()
-        .map(|element_id| get_render_function(*element_id))
-        .flat_map(|render| render(context).0)
-        .collect::<Vec<Span>>();
-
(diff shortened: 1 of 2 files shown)


hamrik:
The only potential regression is that `spinner` no longer reserving space may cause neighboring elements to jump around. I think the most elegant solution would be to simply move it towards the inner end of either `left` or `right` in the default config, where it couldn't shift other elements.

Issue #5706, opened 2023-01-27
`:open` should support globbing / wildcards


LeoniePhiline:
I haven't found a bug report for this yet - sorry if I missed one!

I had just now wanted to open a whole range of files via `:open ~/path/to/prefix_*`.

Helix created a new buffer with that name instead of expanding the `*` and opening all matching files. It would be real neat if globbing was supported!

mslot:
Should we have an upper recursive and directory limit that can be overridden? Imagine opening /**/*, or just a directory with one million files in it.

askreet:
The behavior in (n)vim seems to be:
* If something looks like a glob pattern, try to match it into a file list.
* If that works, check to see if there are "too many" files and error out.
* If that doesn't work, open a buffer with the filename including glob-like characters.

Would we want the mirror that behavior? Seems generally OK to me. The alternative to #3 would be something like error out if no files match and provide a way to escape the glob pattern should you want to edit a file called "file*".

mslot:
> The behavior in (n)vim seems to be:
> 
>     * If something looks like a glob pattern, try to match it into a file list.
> 
>     * If that works, check to see if there are "too many" files and error out.
> 
>     * If that doesn't work, open a buffer with the filename including glob-like characters.
> 
> 
> Would we want the mirror that behavior? Seems generally OK to me. The alternative to #3 would be something like error out if no files match and provide a way to escape the glob pattern should you want to edit a file called "file*".

I think that is a good algorithm. I can accept that.

Pablu23:
Hi I would like to try and implement this. Its my first time collaborating, so I was curious if it is allowed to use crates, I would try to implement this with the glob crate.

the-mikedavis:
I'm not sure we need globbing support. IMO this overlaps with the file picker and I would rather see multiple selections in the file-picker (https://github.com/helix-editor/helix/pull/4837) than extending `:open`. We try to keep the dependency count small so pulling in a new crate just for this would not be ideal.

mslot:
> I'm not sure we need globbing support. IMO this overlaps with the file picker and I would rather see multiple selections in the file-picker (#4837) than extending `:open`. We try to keep the dependency count small so pulling in a new crate just for this would not be ideal.

I can't see why this shouldn't be supported with :open, because this is one of those things where you would want to quickly point out what to open, and globbing is perfect for that.

I do agree that pulling in another crate is not what we want. 

[...]

PR #9709, opened 2024-02-23
Optimization of tilde expansion

mo8it:
I was reading some code to fix https://github.com/helix-editor/helix/issues/5706 but then I found that `expand_tilde` allocates even when it is not necessary. Therefore, I added some 🐄

Please read the commit messages and ask me if anything is not clear.

This is my first contribution, so please tell me if I missed some contribution guide point 😇


helix-stdx/src/path.rs
```diff
@@ -19,19 +22,18 @@ pub fn fold_home_dir(path: &Path) -> PathBuf {
...
 /// and only slash follows it.
-pub fn expand_tilde(path: impl AsRef<Path>) -> PathBuf {
-    let path = path.as_ref();
-    let mut components = path.components().peekable();
-    if let Some(Component::Normal(c)) = components.peek() {
-        if c == &"~" {
-            if let Ok(home) = home_dir() {
-                // it's ok to unwrap, the path starts with `~`
-                return home.join(path.strip_prefix("~").unwrap());
+pub fn expand_tilde(path: Cow<'_, Path>) -> Cow<'_, Path> {
```
the-mikedavis, line 25: Let's use `Into` here so most callers don't need to change:

```rust
fn expand_tilde<'a, P: Into<Cow<'a, Path>>>(path: P) -> Cow<'a, Path>
```

the-mikedavis (changes requested):
We generally try to avoid refactors unless the new code is a big improvement. Avoiding the allocation here is nice-to-have but not worth a medium/large change. With some small tweaks we can get the diff to be very small though

the-mikedavis (approved):
Thanks!

PR #9712, opened 2024-02-24
re-detect language after file modifications

rikushoney:
While working on shell scripts without file extensions I noticed that the language is not detected after adding a shebang to the top of the file. I had to close the file using `buffer-close` and `open` it again for the changes to be detected.

I suggest that the language should be re-detected after the following commands:

- write
- write-all
- reload
- reload-all

Please let me know what you think and whether there are more commands that should have this behaviour.

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -340,6 +340,8 @@ fn write_impl(
         insert_final_newline(doc, view);
     }
 
+    doc.detect_language(cx.editor.syn_loader.clone());
+
     let fmt = if config.auto_format {
         doc.auto_format().map(|fmt| {
             let callback = make_format_callback(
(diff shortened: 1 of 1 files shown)


pascalkuthe (changes requested):
The way you refresh document languages is wrong. You would need to use `Editor::refresh_doc_language`.

That said I am against this change. We allow manually setting filetype (and indentstion setting) and redetcting those everytime you save/reload would make that mechanism useless.

Realistically the language if a file isn't changing often and in that case using `:lang` seems perfectly fine to me

dead10ck:
Yeah this would be extremely annoying to the point of breaking the editor for many use cases, especially templated files. It's not that hard to `:set-language bash`

rikushoney:
All good, thanks for the feedback!

PR #7494, opened 2023-07-01
Add set-register command

ElSargo:
Adds a typeable command to push a value onto a register. Useful for keybinds.

PR #9717, opened 2024-02-24
Added command to copy between registers

### LINKED ISSUE #9552: Add command to copy values between registers
Inspired by @the-mikedavis's suggestion here https://github.com/helix-editor/helix/discussions/8012#discussioncomment-6774364

> I think it could be useful to have a command that copies values between registers though since pasting and deleting adds an undo checkpoint.

uek-1:
Added a command to copy the contents of the selected register to another. Closes #9552.

### DIFF
--- a/helix-term/src/commands.rs
+++ b/helix-term/src/commands.rs
@@ -440,6 +440,7 @@ impl MappableCommand {
         wclose, "Close window",
         wonly, "Close windows except current",
         select_register, "Select register",
+        copy_between_registers, "Copy contents of selected register to another",
         insert_register, "Insert register",
         align_view_middle, "Align view middle",
         align_view_top, "Align view top",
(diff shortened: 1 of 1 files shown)


pascalkuthe:
I wonder if this is quite the right command tough. kacuone has a more generic `:set-register` typable command. That could be combined with command expansion (where I think being able to expand to registers would be quite useful) to implement the same thing

uek-1:
> I wonder if this is quite the right command tough. kacuone has a more generic `:set-register` typable command. That could be combined with command expansion (where I think being able to expand to registers would be quite useful) to implement the same thing

To be clear, one would do this in kakoune by typing `:set_register <destination_register> %reg{source_register}` right? If so, I think this is super versatile but I think it may be less intuitive than an explicit copy command.

the-mikedavis:
I think a `:set-register` command could also be useful but it might be complicated to use it to copy between registers, for example if you're copying a set of selections from one register to another

pascalkuthe:
We discussed this briefly: I don't think a typable`:copy-register` command carrys its weight over `:set-register %reg{..}` but an interactive version bound to a key that would prompt the user which registers to copy from/to would.be worth adding

the-mikedavis (commented):
To make this a little more interactive let's show the autoinfo popup for the registers like in `select_register`: https://github.com/helix-editor/helix/blob/f46a09ab4f945273c7baf32e58438b501914fabb/helix-term/src/commands.rs#L4836-L4844

and prompt for both register with two `on_next_key` callbacks

kirawi:
Should we close https://github.com/helix-editor/helix/pull/7494 in favor of this?

pascalkuthe:
> Should we close #7494 in favor of this?

No we want both set-register so you can combine it with commands expansion (and use it in keymaps) and an interactive command to copy between registers.

the-mikedavis:
Superseded by 066e938ba083c0259ff411b681eca7bad30980df

PR #9723, opened 2024-02-25
Globbing support in `:open`

mo8it:
Closes https://github.com/helix-editor/helix/issues/5706

The functionality is done, but I would like to have feedback on how to improve the error handling.

Please tell me if you prefer moving the two inner functions to somewhere else.

Should I add tests for the globing functionality? If yes, where?

### DIFF
--- a/helix-term/src/commands/typed.rs
+++ b/helix-term/src/commands/typed.rs
@@ -5,9 +5,11 @@ use crate::job::Job;
 
 use super::*;
 
+use globset::{Glob, GlobBuilder, GlobSet};
 use helix_core::fuzzy::fuzzy_match;
 use helix_core::indent::MAX_INDENT;
 use helix_core::{encoding, line_ending, shellwords::Shellwords};
+use helix_stdx::env::current_working_dir;
 use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::editor::{Action, CloseError, ConfigEvent};
 use serde_json::Value;
@@ -108,16 +110,16 @@ fn open(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) -> 
     }
 
     ensure!(!args.is_empty(), "wrong argument count");
-    for arg in args {
-        let (path, pos) = args::parse_file(arg);
-        let path = helix_stdx::path::expand_tilde(path);
-        // If the path is a directory, open a file picker on that directory and update the status
-        // message
-        if let Ok(true) = std::fs::canonicalize(&path).map(|p| p.is_dir()) {
+
+    fn open(cx: &mut compositor::Context, path: Cow<Path>, pos: Position) -> anyhow::Result<()> {
+        // If the path is a directory, open a file picker on that directory and update
+        // the status message
+        if path.is_dir() {
+            let path = path.into_owned();
             let callback = async move {
                 let call: job::Callback = job::Callback::EditorCompositor(Box::new(
                     move |editor: &mut Editor, compositor: &mut Compositor| {
-                        let picker = ui::file_picker(path.into_owned(), &editor.config());
+                        let picker = ui::file_picker(path, &editor.config());
                         compositor.push(Box::new(overlaid(picker)));
                     },
                 ));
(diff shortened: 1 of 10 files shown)


daedroza:
Great functionality, testing here and works good!

mo8it:
I added an upper limit of files to open with a glob and implemented ignoring directories when globbing. The PR is ready to be merged after a review 😃

postsolar:
Currently this breaks opening non-existent file, e.g. `:o file_which_doesnt_exist.txt`.

mo8it:
@postsolar You are right. I will fix it in the next days.

mo8it:
Fixed opening non-existent files and merge conflicts.

helix-term/src/commands/typed.rs
```diff
@@ -5,14 +5,20 @@ use crate::job::Job;
...
 use helix_core::{encoding, line_ending, shellwords::Shellwords};
+use helix_stdx::env::current_working_dir;
 use helix_view::document::DEFAULT_LANGUAGE_NAME;
 use helix_view::editor::{Action, CloseError, ConfigEvent};
 use serde_json::Value;
 use ui::completers::{self, Completer};
 
+// The maximum number of files to open with globbing to avoid freezing while trying to open too many files.
+// It is kind of arbitrary and can be raised or offered as a configuration option.
+const GLOBBING_MAX_N_FILES: usize = 32;
```
pascalkuthe, line 20: I think this limit should be higher. Maybe 64. I doubt that would freeze to ui much and I had some niche cases where I really did want to open that many files. Wouldn't want to add a confog for this

helix-term/src/commands/typed.rs
```diff
@@ -108,30 +114,115 @@ fn open(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) ->
     }
 
     ensure!(!args.is_empty(), "wrong argument count");
+
+    fn open_file(cx: &mut compositor::Context, path: &Path, pos: Position) -> anyhow::Result<()> {
```
pascalkuthe, line 118: Please don't make these inner functions. I find those quite hard to read with all the nesting

helix-term/src/commands/typed.rs
```diff
@@ -108,30 +114,115 @@ fn open(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) ->
...
+                    Ok(call)
+                };
+                cx.jobs.callback(callback);
+            } else {
+                open_file(cx, &path, pos)?;
+            }
+            continue;
+        }
+
+        let mut glob_set_builder = GlobSet::builder();
```
pascalkuthe, line 161: This entire logic seems both confusing/buggy and backwards to me. I would think you should just:

* check if the oath with the given name exists (open it if it does)
* use the override matcher of ignore to ignore everything (just an ignore for `*`) and then whiteliste the specified glob pattern

helix-term/src/commands/typed.rs
```diff
@@ -108,30 +114,115 @@ fn open(cx: &mut compositor::Context, args: &[Cow<str>], event: PromptEvent) ->
...
+        }
+
+        let glob_set = glob_set_builder.build().context("invalid glob")?;
+        let root = root.unwrap_or_else(current_working_dir);
+
+        let mut walk = cx
+            .editor
+            .config()
+            .file_picker
+            .walk_builder(root)
```
pascalkuthe, line 191: Don't reuse the walk idler config of the file picker here. We do not want to be ignoring files here, in fact make sure to turn all default options for ignore entirely. If you write a glib pattern that matches a hidden file then that hidden file should be opened

helix-term/src/commands/typed.rs
mo8it: I don't see how I could use `ignore` for this. Maybe I am missing something though.

I tried something like the following code, but it goes through the whole file system and takes 15 seconds on my machine:

```rust
fn main() {
    let walker = ignore::WalkBuilder::new("/")
        .overrides(
            ignore::overrides::OverrideBuilder::new("")
                .add("/bacon.toml")
                .unwrap()
                .build()
                .unwrap(),
        )
        .build();

    for entry in walker {
        let Ok(entry) = entry else {
            continue;
        };
        if entry.metadata().unwrap().is_file() {
            println!("{}", entry.path().display());
        }
    }
}
```

It should only go through the root directory and stop because the glob starts with `/`. I tried to add `!*` but then it doesn't match anything anymore (also not when `bacon.toml` exists).

Anyway, I don't think that we need `ignore`. I wrote an optimized walker without `ignore` and ask for a new review :)

mo8it:
I was too focused on skipping non-relevant parents 😅 I just use the parallel `ignore` walker now. Ready for a review :)

gyreas:
Your work is a prerequisite for a similar one that I have in mind which requires file globbing. Can I trouble you to factor out the code that retrieves matching paths, so I don't have to do it later on?

Issue #9726, opened 2024-02-25
Consider update from old Lua grammar


briandorsey:
Helix appears to be referencing a two year old version of the Lua grammar:
https://github.com/helix-editor/helix/blob/928bf80d9a1d6206f864e9b375f67662a49a6265/languages.toml#L1104C75-L1104C115
```toml
[[grammar]]
name = "lua"
source = { git = "https://github.com/MunifTanjim/tree-sitter-lua", rev = "887dfd4e83c469300c279314ff1619b1d0b85b91" }
```

There is a more recent release (and new repo url, which appears to be the same repo, just moved):
https://github.com/tree-sitter-grammars/tree-sitter-lua/commit/88e446476a1e97a8724dff7a23e2d709855077f2

Is it possible/easy to update this? Is it just a matter of changing the values? eg, something like this: 
```toml
[[grammar]]
name = "lua"
source = { git = "https://github.com/tree-sitter-grammars/tree-sitter-lua", rev = "88e446476a1e97a8724dff7a23e2d709855077f2" }
```

briandorsey:
Follow up. I tested locally by adding a `.config/helix/languages.toml` file with the following content:
```toml
use-grammars = { only = [ "lua" ] }

[[grammar]]
name = "lua"
source = { git = "https://github.com/tree-sitter-grammars/tree-sitter-lua", rev = "88e446476a1e97a8724dff7a23e2d709855077f2" }
```

Then updated the grammar:
```zsh
% hx --grammar fetch
Fetching 1 grammars
1 updated grammars
	lua now on 88e446476a1e97a8724dff7a23e2d709855077f2
% hx --grammar build
Building 1 grammars
1 grammars built now
	["lua"]
```

Then confirmed that the new grammar is being used. It is, yea! 

This is also a work-around for anyone who needs updated Lua parsing before helix itself is updated.

archseer:
Can you open a PR?

briandorsey:
Just created a PR. Please take a look when it's convenient.

PR #9727, opened 2024-02-26
update languages.toml: tree-sitter-lua grammar

briandorsey:
repo has moved, use new URL and the rev of the latest release (v0.0.19)

There is a more recent release of the tree-sitter-lua grammar (and new repo url, which appears to be the same repo, just moved):
https://github.com/tree-sitter-grammars/tree-sitter-lua/commit/88e446476a1e97a8724dff7a23e2d709855077f2

I'm a first-time contributor, opening this PR per request from @archseer on #9726. 

I tested locally by adding a `.config/helix/languages.toml` file with the following content (the `source =` line is the same content as this PR):
```toml
use-grammars = { only = [ "lua" ] }

[[grammar]]
name = "lua"
source = { git = "https://github.com/tree-sitter-grammars/tree-sitter-lua", rev = "88e446476a1e97a8724dff7a23e2d709855077f2" }
```

Then updated the grammar:
```zsh
% hx --grammar fetch
Fetching 1 grammars
1 updated grammars
	lua now on 88e446476a1e97a8724dff7a23e2d709855077f2
% hx --grammar build
Building 1 grammars
1 grammars built now
	["lua"]
```

Then confirmed that the new grammar is being used. It is, yea! 

fixes: #9726

### DIFF
--- a/runtime/queries/lua/highlights.scm
+++ b/runtime/queries/lua/highlights.scm
@@ -128,9 +128,9 @@
 
 ;; Tables
 
-(field name: (identifier) @field)
+(field name: (identifier) @variable.other.member)
 
-(dot_index_expression field: (identifier) @field)
+(dot_index_expression field: (identifier) @variable.other.member)
 
 (table_constructor
 [
@@ -198,7 +198,7 @@
 ;; Nodes
 (comment) @comment
 (string) @string
-(escape_sequence) @string.escape
+(escape_sequence) @constant.character.escape
 (number) @constant.numeric.integer
 (label_statement) @label
 ; A bit of a tricky one, this will only match field names
--- a/runtime/queries/lua/highlights.scm
+++ b/runtime/queries/lua/highlights.scm
@@ -1,9 +1,5 @@
 ;;; Highlighting for lua
 
-;;; Builtins
-((identifier) @variable.builtin
- (#eq? @variable.builtin "self"))
-
 ;; Keywords
 
 (if_statement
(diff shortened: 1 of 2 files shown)


woojiq:
Very often, when you update a version of the Tree-sitter grammar, the node names change, causing the current helix queries to become inaccurate. Check the changelog and update our queries accordingly.
For example, I see that highlight queries have changed several times in the last 2 years: https://github.com/tree-sitter-grammars/tree-sitter-lua/commits/main/queries/highlights.scm

briandorsey:
Are there any tools or docs to help validate that changes to these queries correctly map to helix expectations? I assume it's not ok to just copy the query files over from the `tree-sitter-lua` repo, correct?

I spent about an hour trying to learn about this. I started by poking around at the files in the `tree-sitter-lua` repo vs. what is in the `helix` repo... they are very different, and I don't really know where to start changing them with any confidence that it wouldn't break something. Later, started looking at tree-sitter docs and trying to run tree-sitter query (some progress) or playground (blocked on wasm/Docker troubleshooting)... and then ran out of the time I have available to look at it today. 

At the moment, this looks like it'll take a while to learn enough to update the queries with any confidence.

briandorsey:
Thanks for the guidance! I attempted to follow it. I'm not very confident that I got the ordering or captures correct.

The `:tree-sitter-*` commands were very helpful for visualizing/understanding what's happening with the mapping. 

Aside: I was a bit confused how the `:tree-sitter-*` commands decided what to show. `-subtree` seemed to perhaps parse the whole file when a full line was selected (with `x`)

runtime/queries/lua/highlights.scm
```diff
@@ -153,6 +198,7 @@
 ;; Nodes
 (comment) @comment
 (string) @string
+(escape_sequence) @string.escape
```
the-mikedavis, line 201:
```suggestion
(escape_sequence) @constant.character.escape
```

runtime/queries/lua/highlights.scm
```diff
@@ -130,16 +126,65 @@
...
  (#match? @constant "^[A-Z][A-Z_0-9]*$"))
 
-;; Parameters
-(parameters
-  (identifier) @variable.parameter)
+;; Tables
+
+(field name: (identifier) @field)
+
+(dot_index_expression field: (identifier) @field)
```
the-mikedavis, line 133:
```suggestion
(field name: (identifier) @variable.other.member)

(dot_index_expression field: (identifier) @variable.other.member)
```

the-mikedavis (approved):
Thanks!

the-mikedavis:
> `-subtree` seemed to perhaps parse the whole file when a full line was selected (with `x`)

`:tree-sitter-subtree` tries to find the largest node that fully contains the current selection. Usually newlines are part of the root of the syntax tree so `:tree-sitter-subtree` on a full line gives the root. You can use `trim_selections` (`_`) before to trim the leading/trailing whitespace of the selection

pascalkuthe (approved):

PR #9732, opened 2024-02-26
Auto Save All Buffers After A Delay

miggs597:
The feature I would like to introduce with this pull request is the ability to auto save one's work after idling. I've found it to be a great editing experience, especially while working with a stack that offers hot module reloading.

My implementation works similarly to `idle_timer`, that is once my `save_delay_timer` has reached zero an `EditorEvent::SaveDelayTimer` is returned causing `EditorView::handle_event` to handle that timeout and save the buffers.

Any feedback is greatly appreciated.

### DIFF
--- a/helix-term/src/handlers/auto_save.rs
+++ b/helix-term/src/handlers/auto_save.rs
@@ -16,7 +16,8 @@ use helix_view::{
 use tokio::time::Instant;
 
 use crate::{
-    commands, compositor, job::{self, Jobs}
+    commands, compositor,
+    job::{self, Jobs},
 };
 
 #[derive(Debug)]
@@ -44,25 +45,25 @@ impl helix_event::AsyncHook for AutoSaveHandler {
 
     fn handle_event(
         &mut self,
-        _: Self::Event,
+        event: Self::Event,
         _: Option<tokio::time::Instant>,
     ) -> Option<Instant> {
-        // match event {
-        //     AutoSaveEvent::Trigger => {
-        //         if matches!(self.state, State::Closed) {
-        //             self.trigger = Some(AutoSaveInvoked::Automatic);
-        //             return Some(Instant::now() + Duration::from_millis(1000));
-        //         }
-        //     }
-        //     AutoSaveEvent::Cancel => {
-        //         self.state = State::Closed;
-        //         return None;
-        //     }
-        // }
+        match event {
+            AutoSaveEvent::Trigger => {
+                if matches!(self.state, State::Closed) {
+                    self.trigger = Some(AutoSaveInvoked::Automatic);
+                    return Some(Instant::now() + Duration::from_millis(1000));
+                }
+            }
+            AutoSaveEvent::Cancel => {
+                self.state = State::Closed;
+                return None;
+            }
+        }
 
-        // if self.trigger.is_none() {
-        //     self.trigger = Some(AutoSaveInvoked::Automatic)
-        // }
+        if self.trigger.is_none() {
+            self.trigger = Some(AutoSaveInvoked::Automatic)
+        }
 
         Some(Instant::now() + Duration::from_millis(1000))
     }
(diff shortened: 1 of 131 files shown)


woojiq:
Sorry, I don't understand the use case if helix already has `auto-save`:
> Enable automatic saving on the focus moving away from Helix.

Can you explain more about  hot reloading usage? Do you want to automatically save every second or what?

kirawi:
Resolves https://github.com/helix-editor/helix/issues/3451

miggs597:
> Sorry, I don't understand the use case if helix already has `auto-save`:
> 
> > Enable automatic saving on the focus moving away from Helix.
> 
> Can you explain more about hot reloading usage? Do you want to automatically save every second or what?

Yes, helix does already have auto-save. However if enabled it's only triggered by changing focus away from helix. This change would allow for buffers to get saved after a certain amount of time has elapsed without any new input. 

Before I was using vscode as my text editor and got used to it there. https://code.visualstudio.com/docs/editor/codebasics#_save-auto-save

miggs597:
> This is definitely useful but it shouldn't be implemented with the idle timeout we will remove that soonish (once all items in the tracking issue #9629 are done)

Can you point me in the direction of some code samples that use the event system. I can work on updating my pr to use it instead of the current ad-hoc method.

kirawi:
https://github.com/helix-editor/helix/tree/master/helix-term/src/handlers contain them

miggs597:
Moving to a draft for now. I've started working on porting the idea over to the event system.

miggs597:
> This is definitely useful but it shouldn't be implemented with the idle timeout we will remove that soonish (once all items in the tracking issue #9629 are done)

@pascalkuthe I realized that my initial pr description wasn't clear. I am not piggybacking off of idle timeout, however I do depend on the `event_loop_until_idle`. That's where it is being checked if the save timeout has been reached. 

Is `Application::event_loop_until_idle` also being ported over to the event system, or is that code going to stay as is?
I'm trying to figure out from where it would be most appropriate to dispatch an auto save event after a duration has expired.

pascalkuthe:
All of that is going away. Eberything is getting its own async handler with its own timout.

Moving around should not prevent autosave. You should be registering documentdidchange handler and start a (fairly long like 5s) timeout there that will trigger autosave when finished. Like kirawi said look at how the signature help handler is implemented (just that you don't care about mode and selection changes)

miggs597:
Initial port to the event system, implementation is currently half-baked. I need to iron out the user experience, and to get it working consistently.

helix-term/src/handlers/auto_save.rs
```diff
@@ -0,0 +1,101 @@
...
+};
+
+#[derive(Debug)]
+enum State {
+    Closed,
+}
+
+#[derive(Debug)]
+pub(super) struct AutoSaveHandler {
+    trigger: Option<AutoSaveInvoked>,
```
pascalkuthe, line 30: There is no need to keep the trigger around , you don't actually use this

helix-term/src/handlers/auto_save.rs
```diff
@@ -0,0 +1,101 @@
...
+        editor,
+        scroll: Some(0),
+        jobs: &mut Jobs::new(),
+    };
+
+    if let Err(e) = commands::typed::write_all_impl(context, false, false) {
+        context.editor.set_error(format!("{}", e));
+    }
+
+    if let Err(e) = context.block_try_flush_writes() {
```
pascalkuthe, line 87: I don't think thsi should be here, this is only intended for shutdown. We don't want to freeze the editor while writing to disk

pascalkuthe (commented):
This is actually already a really good implementation. I have some minor comments but apart from that this seems reads to merge actually

helix-view/src/editor.rs
```diff
@@ -840,7 +856,9 @@ impl Default for Config {
             auto_completion: true,
             auto_format: true,
             auto_save: false,
+            save_style: SaveStyle::FocusLost,
             idle_timeout: Duration::from_millis(250),
+            save_delay_timeout: Duration::from_millis(1000),
```
pascalkuthe, line 861: Maybe the default should be longer like 5s?

Banzobotic: For me the primary benefit of auto saving after a delay is updating lints from LSPs such as rust-analyzer. IMO one second is the best default for this use case. This also seems to be main use case mentioned on #3451. 1s is also the default in vscode and it works well there.

pascalkuthe:
closing in favor of #10899

PR #9739, opened 2024-02-27
Avoid cloning the whole paragraph content just for rendering

mo8it:
No corresponding issue, just found it odd while reading. Feel free to reject it, but I think that cloning the whole content of a paragraph just to render it can be significant. This is not just one allocation, these are multiple allocations because `Text` is a vector of vectors of possibly owned strings.

### DIFF
--- a/helix-term/src/ui/editor.rs
+++ b/helix-term/src/ui/editor.rs
@@ -716,7 +716,8 @@ impl EditorView {
             }
         }
 
-        let paragraph = Paragraph::new(lines)
+        let text = Text::from(lines);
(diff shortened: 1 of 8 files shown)


pascalkuthe (approved):
I want to get rid of the utilities in helix-tui anyway and use the wrapping from helix-core anyway. This code doesn't get much attention. But I would be fine with this

mo8it:
Fixed the tests. I couldn't run them locally because of https://github.com/helix-editor/helix/discussions/9710

the-mikedavis (approved):

pascalkuthe (approved):

Issue #8985, opened 2023-12-04
Request textDocument/hover from all configured language servers


tcoopman:
I couldn't find if this is an issue with the event system (#8021), and can't seem to find a similar issue.

`onHover` docs doesn't work well with multiple LSPs. For example, I've got both `["tailwind", "elixir"]` configured.
If I hover on something from elixir, I don't get any docs. If I remove the tailwind lsp, I do get docs for elixir, but of course none for the tailwind classes.
If I reverse the order, hover on `elixir` works but not on `tailwind`.

So the issue is that hover should work for both configured lsps.

### Discussed in https://github.com/helix-editor/helix/discussions/8872

kirawi:
https://github.com/helix-editor/helix/blob/510928618dfcddcd6d3086b278332fa93ee32998/helix-term/src/commands/lsp.rs#L1279-L1328

PR #9744, opened 2024-02-28
Show hover results for multiple LSPs

kyfanc:
This PR attempts to address #8985.

Currently, only the first configured LSP is used for handling hover feature. This PR supports multiple configured LSPs and merged the results by appending in the popup markdown.

# Result:

## with single LSP configured

- effectively same behaviour as before this PR.
single-lsp

## with multiple LSP configured

- results are joined in LSP response order, with LSP name prefixed to identify which LSP the result came from.
multiple-lsp

Testing:

1. modify `languages.toml` to add multiple LSPs
```toml
# rust
[[language]]
name = "rust"
# multiple lsp
language-servers = [ "rust-analyzer", "rust-analyzer-dup"]

# secondary rust-analyzer
[language-server.rust-analyzer-dup]
command = "rust-analyzer"
```
1. open a codebase, and trigger LSP hover commands

### Ordering of results

- Since the LSP hover requests are dispatched in async callback style, the ordering of results is not guaranteed. So the ordering in popup markdown will follow to the LSP response order. 
- The LSP name prefix would help identify which LSP the response is coming from.

### DIFF
--- a/helix-term/src/commands/lsp.rs
+++ b/helix-term/src/commands/lsp.rs
@@ -955,9 +955,11 @@ pub fn hover(cx: &mut Context) {
         .map(|language_server| {
             // TODO: factor out a doc.position_identifier() that returns lsp::TextDocumentPositionIdentifier
             let pos = doc.position(view.id, language_server.offset_encoding());
-            language_server
+            let name = language_server.name().to_string();
+            let request = language_server
                 .text_document_hover(doc.identifier(), pos, None)
-                .unwrap()
+                .unwrap();
+            (name, request)
         })
         .collect();
 
@@ -970,7 +972,9 @@ pub fn hover(cx: &mut Context) {
         return;
     }
 
-    for future in requests {
+    let lsp_count = requests.len();
+
+    for (name, future) in requests {
         cx.callback(
             future,
             move |editor, compositor, response: Option<lsp::Hover>| {
@@ -990,7 +994,7 @@ pub fn hover(cx: &mut Context) {
                         }
                     }
 
-                    let contents = match hover.contents {
+                    let mut contents = match hover.contents {
                         lsp::HoverContents::Scalar(contents) => marked_string_to_markdown(contents),
                         lsp::HoverContents::Array(contents) => contents
                             .into_iter()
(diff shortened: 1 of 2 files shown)


kyfanc:
Updated to address some issues in previous commits.
- now hover contents shall follow the lsp configure order more reliably
- deduplication is performed for same LSP configured repeatedly
- fix async race condition of slow hover response and hover popup interactions

woojiq:
suggestion (ui): a more visible separator between multiple results. Maybe some kind of bold horizontal line or even a double horizontal line.

kyfanc:
Thank you for the suggestion.

Since the hover contents are rendering in markdown, I think it is better to stay with markdown horizontal rule `---` as separator between LSPs results, so that future changes to markdown rendering will be compatible. 

And I noticed that depending on the LSP and languages, the hover result from LSP may include `---` as well, making it harder to identify our separators from a glance. And the LSP names varies, making it hard to spot the names especially jumping between languages.

I think adding a prefix to LSP name may help, as it provides a more consistent visual "anchor".

# rust
Screenshot 2024-03-06 at 11 14 12 AM

- rust-analyser may include `---` 
Screenshot 2024-03-06 at 11 12 57 AM

- default theme
Screenshot 2024-03-06 at 11 10 42 AM

# typescript
- `typescript-language-server` looks cleaner
Screenshot 2024-03-06 at 11 24 09 AM


Any thoughts?

gabydd:
I wonder if we want to use the ui from https://github.com/helix-editor/helix/pull/9974 instead

kyfanc:
The UI from in that PR looks cool! 

With a brief look, it seems that the code for signature help feature has been refactored with the new event system recently, while the hover feature is still using the old way. The event system helps dealing with the interactivity of cycling through async results. 

For the key binding for cycling, intuitively, I think it would be great to share the same keybindings with any popup feature, like in that PR, but I am not sure if it is possible under the current architecture.

Considering the scale of refactoring and discussion of keybinding for cycling through "popup pages", maybe that could be another PR? 

Anyway, I have pushed the prefix improvement mentioned previously for now, so it is ready to merged as is. 
And I will look into the event system/ cycle UI if I have time.

kyfanc:
I opened another PR #10122 for the new approach with event system and cycling pages UX. 
This PR can be closed if that is more promising.

pascalkuthe:
I like the cycle approach better, let's close tgjsn in favor of #9744 👍

PR #9758, opened 2024-02-29
feat: Add markdown-oxide language server

Feel-ix-343:
I would like to add this language server. How do you all handle this? https://github.com/Feel-ix-343/markdown-oxide

### DIFF
--- a/book/src/generated/lang-support.md
+++ b/book/src/generated/lang-support.md
@@ -104,7 +104,7 @@
 | lua | ✓ | ✓ | ✓ | `lua-language-server` |
 | make | ✓ |  | ✓ |  |
 | markdoc | ✓ |  |  | `markdoc-ls` |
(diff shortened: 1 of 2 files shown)


the-mikedavis (commented):
If markdown-oxide is specific to obsidian markdown it would probably be best to only add this in the language server wiki: https://github.com/helix-editor/helix/wiki/How-to-install-the-default-language-servers

But if it's generally useful for markdown then this patch looks good. You'll just need to run `cargo xtask docgen` and commit the changes to fix the docs CI

Feel-ix-343:
It is compatible with obsidian and takes influence from it but is primarily an advanced markdown language server.

There are a few people who are migrating to it not from obsidian but from marksman and other non obsidian systems. 

I think it is general enough to be built in.

pascalkuthe (approved):

the-mikedavis (approved):

